PR #1322, opened 2025-07-24
Add k_scale and v_scale to persistent attention 

Edenzzzz:
## 📌 Description

For consistency with the general call in SGLang
## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/flashinfer/attention.py
+++ b/flashinfer/attention.py
@@ -130,6 +130,8 @@ class BatchAttention:
         kv_cache: Union[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]],
         out: Optional[torch.Tensor] = None,
         lse: Optional[torch.Tensor] = None,
+        k_scale: Optional[torch.Tensor] = None,
+        v_scale: Optional[torch.Tensor] = None,
         profiler_buffer: Optional[torch.Tensor] = None,
     ) -> Tuple[torch.Tensor, torch.Tensor]:
         if profiler_buffer is None:
@@ -147,6 +149,8 @@ class BatchAttention:
         head_dim_qk = q.shape[2]
         if self._sm_scale is None:
             self._sm_scale = 1.0 / math.sqrt(head_dim_qk)
+        if k_scale is not None:
+            self._sm_scale *= k_scale
 
         # profiler_buffer is optional
         profiler_args = (profiler_buffer,) if self._use_profiler else ()
@@ -169,5 +173,11 @@ class BatchAttention:
             self._sm_scale,
             *profiler_args,
         )
+        if v_scale is not None:
+            # TODO(Zihao): fused into kernel
+            if out.itemsize == 1:
+                out = (out.to(float) * v_scale).to(out.dtype)
+            else:
+                out *= v_scale
 
         return out, lse


yyihuang:
I think it's better to keep only one pre-fused param as attn softmax scale, with some examples here:
https://github.com/flashinfer-ai/flashinfer/blob/main/flashinfer/decode.py#L1799-L1800
https://github.com/flashinfer-ai/flashinfer/blob/main/flashinfer/decode.py#L2052-L2053

Edenzzzz:
However, the interface has been like this in SGLang

yyihuang:
It's better to keep the fusion overhead at inference engine side, like https://github.com/sgl-project/sglang/pull/7938

https://github.com/farazkh80/sglang/blob/fkhoubsirat-trtllm_gen_mla_sglang/python/sglang/srt/layers/attention/trtllm_mla_backend.py#L352

flashinfer/attention.py
```diff
@@ -169,5 +173,11 @@ def run(
             self._sm_scale,
             *profiler_args,
         )
+        if v_scale is not None:
+            # TODO(Zihao): fused into kernel
+            if out.itemsize == 1:
+                out = (out.to(float) * v_scale).to(out.dtype)
```
yzh119: Can you also implement fusing the multiply inside the kernel in the persistent kernel template?

Also, it will be good to specify an out_dtype, for quantized input such as fp8, we usually use bf16 as output data type. You can check https://github.com/flashinfer-ai/flashinfer/blob/435f34175c5967e88631a31325a4bce46fefe47e/flashinfer/gemm.py#L1198-L1203 for reference.

Edenzzzz: In that case we need to pass in `out_dtype` during plan? Which can be troublesome

tests/test_batch_attention.py
```diff
@@ -140,7 +141,7 @@ def _run_attention(
 
     # --------- old scheduler --------- #
     wrapper_old = flashinfer.BatchPrefillWithPagedKVCacheWrapper(
-        torch.empty(128 * 1024 * 1024, dtype=torch.uint8, device=dev),
+        torch.empty(256 * 1024 * 1024, dtype=torch.uint8, device=dev),
```
Edenzzzz, line 144: when running the test, the prefill kernel allocator hit OOM, which seems unrelated to this PR (also hit on main), so I increased the buffer size.

Edenzzzz:
Hi @yzh119 I fused `v_scale` into the kernel and tests pass on b200

tests/test_batch_attention.py
yzh119: Np, please go ahead with it.

Edenzzzz:
@yzh119 Seems ready to merge

csrc/batch_attention.cu
```diff
@@ -68,7 +68,9 @@ void BatchPagedAttentionRun(at::Tensor float_workspace_buffer, at::Tensor int_wo
                             at::Tensor v_cache, at::Tensor kv_indices, at::Tensor o,
                             std::optional<at::Tensor> maybe_lse, int64_t mask_mode_code,
                             int64_t layout_code, int64_t num_qo_heads, int64_t num_kv_heads,
-                            int64_t page_size, double sm_scale,
+                            int64_t page_size,
+                            double v_scale,  // must use double due to pytorch binding
```
yzh119, line 72: We should add the option of making this value (`v_scale`) and `sm_scale` on device to make sure we support dynamic scale in CUDAGraph mode (e.g. in #1630 ), we can leave it for future PRs anyways.

PR #1324, opened 2025-07-25
feat: Support logits_soft_cap for Persistent attn; fix kv split limit

Edenzzzz:
## 📌 Description

When integrating this kernel into SGLang, I quickly hit an assertion error with input len 4000, output len 200 and 8 request/s due to the hard limit of 4 kv splits per tile size per SM. This PR fixes the constraint.
image

It also adds support for logits_soft_cap, which is used by Gemma model in sgl
cc @happierpig @yzh119 
## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests


include/flashinfer/attention/persistent.cuh
```diff
@@ -298,6 +298,11 @@ struct BlockBatchPagedAttentionPersistent {
             __syncthreads();
 
             compute_qk<KTraits>(&q_smem, &q_smem_offset_r, &k_smem, &k_smem_offset_r, s_frag);
+            logits_transform<KTraits>(
+                params, variant, /*batch_idx=*/0, qo_packed_idx_base,
+                kv_start + (kv_tile_idx * NUM_WARPS_KV + get_warp_idx_kv<KTraits>(tid.z)) *
+                               NUM_MMA_KV * 16,
+                q_len, kv_len, gqa_group_size, s_frag, tid, kv_head_idx);
```
Edenzzzz: Seems no need to guard with something like `LOOP_SPLIT_MASK` because the func uses static if else branch? Feel free to suggest

happierpig:
@Edenzzzz IMO, the constant buffer size, starting with `max_*` is for cuda graph-compatiable. Only by keeping the same offsets, the same pointer can be fed into the later kernel.

Edenzzzz:
> @Edenzzzz IMO, the constant buffer size, starting with `max_*` is for cuda graph-compatiable. Only by keeping the same offsets, the same pointer can be fed into the later kernel.

Oh I see, then will just use a fixed size. (Though the true array length should be `total_num_rows`)

Edenzzzz:
Now the kernel compiles and runs fine. `ADDITIONAL_FUNC_PARAMS` was buggy due to the use of `params[i]` in persistent launch

yzh119 (approved):

yzh119:
> ADDITIONAL_FUNC_PARAMS was buggy due to the use of params[i] in persistent launch

We should figure out a more elegant solution for additional function parameters as they are shared by all tasks.
Maybe we should add another common_params to the persistent kernel template.

Edenzzzz:
Precision is failing for logits soft cap, still investigating

Edenzzzz:
@yzh119 `tests/test_batch_attention.py` now passes. Should we add tests to the CI build at some point?

Edenzzzz:
I increased `max_total_num_works` to 65536 because under input len 8192, output len 400 and 8 reqs/s, this breaks in SGlang. Maybe we should increase it more because buffer size is not a big issue compared to server crashing

Edenzzzz:
Even if we assume int64 `IdType`, 11 * 65536 * 4 / 1024 ** 2 = 2.75 MB, so I will increase it to 131072

yzh119:
> @yzh119 tests/test_batch_attention.py now passes. Should we add tests to the CI build at some point?
Sure I'll create a PR for that.

PR #1331, opened 2025-07-25
feat: masked layout fp4 gemm using cute-dsl

yzh119:
## 📌 Description

Implement fp4 gemm (w/ masked layout) requested in https://github.com/sgl-project/sglang/issues/7994
Adapted from cutlass's [dense_blockscaled_gemm_persistent](https://github.com/NVIDIA/cutlass/blob/main/examples/python/CuTeDSL/blackwell/dense_blockscaled_gemm_persistent.py) example, with [DeepGEMM style tile-scheduler](https://github.com/deepseek-ai/DeepGEMM/blob/187656694f7f69e3e7975617a68bc3387680a7e1/deep_gemm/include/deep_gemm/common/scheduler.cuh)

## 🔍 Related Issues

https://github.com/sgl-project/sglang/issues/7994

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Reviewer Notes

cc @fzyzcjy
Co-authored-by: Avery Huang <averyh@nvidia.com>


yyihuang (approved):

yzh119:
There are still a lot of work to be done:
1. AOT compile.
2. API design.

Left them for future PRs, let's unblock users and test functionality first.

flashinfer/cute_dsl/blockscaled_gemm.py
```diff
@@ -0,0 +1,2704 @@
...
+            self.num_persistent_clusters
+        )
+        self._num_tiles_executed += Int32(1)
+
+    @property
+    def num_tiles_executed(self) -> Int32:
+        return self._num_tiles_executed
+
+
+"""
```
fengxie, line 312: I think it's safe to delete this docstring?

flashinfer/cute_dsl/blockscaled_gemm.py
```diff
@@ -0,0 +1,2704 @@
...
+        sfa_ptr: cute.Pointer,
+        sfb_ptr: cute.Pointer,
+        c_ptr: cute.Pointer,
+        masked_mptr: cute.Pointer,
+        current_stream: cuda.CUstream,
+    ):
+        a_tensor = cute.make_tensor(
+            a_ptr,
+            layout=cute.make_ordered_layout(
+                (self._m, self._k, self._l),
```
fengxie, line 2454: It's a non-blocking comments. 

This assumes static shape if it's passed by members. Just double check it's what we are expecting here? To support dynamic shape, m/k/l must be passed via `run_cute_ptr`'s argument list as `Int32` type I believe.

yzh119: I think currently it's okay to assume we have static shapes, the number of groups should depend on the TP/EP size and N/K are fixed, we can compile one for each cudagraph configuration. For M we can just set a maximum possible value and the kernel execution time will only depend on the value of `mask_m` tensor, not `M`.

cc @kaixih for confirmation.

fengxie: cool. I think it's also one of the advantage of using `jit` here. You can also selectively choose static shape which usually end-up with better SASS.

PR #1333, opened 2025-07-25
add torch float4_e2m1fn_x2 check for cudnn fp4 backend

ttyio:
add torch float4_e2m1fn_x2 check for cudnn fp4 backend

## 📌 Description

cudnn need this datatype to build graph.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/flashinfer/gemm.py
+++ b/flashinfer/gemm.py
@@ -776,6 +776,11 @@ def _check_cudnn_fp4_availability():
             "Unable to determine cuDNN backend version. FP4 requires backend >= 91002."
         )
 
+    if not hasattr(torch, "float4_e2m1fn_x2"):
+        raise RuntimeError(
+            "Torch version is outdated, the float4_e2m1fn_x2 data type is required to enabled cudnn FP4 backend. Please upgrade to Torch 2.8 or later."
+        )
+
 
 def _get_native_fp4_dtype():
     """get native fp4 datatype if supported in the torch, otherwise return uint8."""


flashinfer/gemm.py
```diff
@@ -776,6 +776,11 @@ def _check_cudnn_fp4_availability():
             "Unable to determine cuDNN backend version. FP4 requires backend >= 91002."
         )
 
+    if not hasattr(torch, "float4_e2m1fn_x2"):
```
yzh119, line 779: Per our discussion we can fallback to uint8/int8 as container for torch 2.7.1 right? Why is it a hard constraint here?

ttyio: We hit issue that in torch 2.7, even upgrade cudnn to latest version, we still cannot generate engine from cudnn api.
from @Anerudhan's comment: 
```
The torch in 25.04 container does not recognize types equivalent to CUDNN_DATA_FP4_E2M1 which means this is passed as a default UINT8 type to cudnn.
```
So the hack we made previous not working here. @Anerudhan  to correct me if I am wrong. Thanks!

yzh119: I see, so this check is designed specifically only for cudnn backend of fp4 gemm?

ttyio: BTW, we tested: 
- cudnn 9.11 + torch 2.7 not works
- cudnn 9.11 + torch 2.8 works

ttyio: > I see, so this check is designed specifically only for cudnn backend of fp4 gemm?

Yes, this check is added inside `_check_cudnn_fp4_availability`

yzh119 (approved):
LGTM, thanks for the hotfix!

PR #1336, opened 2025-07-28
draft: repair C++ build and improve dev experience in native code

hypdeb:
- [ ] Replace `config.cmake` with a `CMakePresets.json`
- [ ] Added `Makefile` with commonly used commands for building and testing the code
- [ ] Fixed a multitude of build and linking errors

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,39 @@
 cmake_minimum_required(VERSION 3.23.1)
 project(flashinfer CUDA CXX)
 
+option(ENABLE_TESTING "Enable testing" ON)
+
+# Everything in here should NOT happen when building the package for distribution.
+if (ENABLE_TESTING)
+
+  # Generates the compile_commands.json file, which is used by the language server (clangd) to do static analysis.
+  set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
+  # By default, CUDA uses response files to pass arguments to the compiler. This is not well supported by tooling, such as clangd.
+  set(CMAKE_CUDA_USE_RESPONSE_FILE_FOR_INCLUDES 0)
+  set(CMAKE_CUDA_USE_RESPONSE_FILE_FOR_LIBRARIES 0)
+  set(CMAKE_CUDA_USE_RESPONSE_FILE_FOR_OBJECTS 0)
+
+  enable_testing()
+
+  # Add lineinfo to the nvcc flags only in release builds.
+  if(CMAKE_BUILD_TYPE STREQUAL "Release")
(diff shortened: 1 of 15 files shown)


CMakeLists.txt
```diff
@@ -1,6 +1,39 @@
...
+  set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
+  # By default, CUDA uses response files to pass arguments to the compiler. This is not well supported by tooling, such as clangd.
+  set(CMAKE_CUDA_USE_RESPONSE_FILE_FOR_INCLUDES 0)
+  set(CMAKE_CUDA_USE_RESPONSE_FILE_FOR_LIBRARIES 0)
+  set(CMAKE_CUDA_USE_RESPONSE_FILE_FOR_OBJECTS 0)
+
+  enable_testing()
+
+  # Add lineinfo to the nvcc flags only in release builds.
```
joker-eph: Why do we do that?
(the comment is basically describing what I can already read on the next line, but does not provide more information here.)

CMakePresets.json
```diff
@@ -0,0 +1,151 @@
...
+                "FLASHINFER_SAMPLING": {
+                    "value": "ON",
+                    "type": "STRING"
+                },
+                "FLASHINFER_NORM": {
+                    "value": "ON",
+                    "type": "STRING"
+                },
+                "FLASHINFER_CUDA_ARCHITECTURES": {
+                    "value": "90a",
```
joker-eph, line 41: Hopper only is a good default?

CMakeLists.txt
jdebache: Added more detailed doc. This allows source mapping in ncu to map assembly back to code and identify bottlenecks.

CMakeLists.txt
```diff
@@ -177,13 +209,15 @@ set(AOT_GENERATE_COMMAND
     --enable_bf16 ${FLASHINFER_ENABLE_BF16} --enable_fp8_e4m3
     ${FLASHINFER_ENABLE_FP8_E4M3} --enable_fp8_e5m2
     ${FLASHINFER_ENABLE_FP8_E5M2})
+MESSAGE(STATUS "AOT_GENERATE_COMMAND=${AOT_GENERATE_COMMAND}")
```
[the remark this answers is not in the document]
jdebache: Done

CMakePresets.json
jdebache: Not necessarily. I would default to whichever compute capability can be detected on the host. In theory, the string `native` should be recognized by cmake, but I need to track all the usages of this variable.

jdebache:
Closing as @yzh119 confirmed that these tests and benchmarks, and the usage of CMake for building is deprecated. Will open another PR to remove all of this instead.

PR #1345, opened 2025-07-29
chore: remove cpp benchmarks, tests, cmake path, as they are deprecated

hypdeb:
After discovering that these tests and benchmarks do not build, I attempted to repair them [here](https://github.com/flashinfer-ai/flashinfer/pull/1336). I was informed that this part of the code is deprecated, and that CMake is not used to perform builds anymore.

To avoid anyone else attempting to use this code and spending time trying to repair it, I suggest that it be removed.


yzh119 (approved):
LGTM. Since we can't guarantee ABI stability for the C++ APIs, I think it's a good decision to retire these obsolete C++ tests and benchmarks--they may otherwise confuse users.

Thank you @hypdeb

PR #1347, opened 2025-07-29
bugfix: Change default index in routingTopKExperts

amirkl94:
## 📌 Description

When all expert scores are `NaN` or `-INFINITY` the default expert index is returned which is currently -1, and this causes an illegal memory access, changed the default index to 0.

### DIFF
--- a/csrc/trtllm_fused_moe_routing_kernel.cu
+++ b/csrc/trtllm_fused_moe_routing_kernel.cu
@@ -561,7 +561,7 @@ __forceinline__ __device__ void routingTopKExperts(cg::thread_block_tile<WarpSiz
                                                    DataType const* ptrScores) {
   DataType minScore = DataType{-INFINITY};
   DataType maxScore = minScore;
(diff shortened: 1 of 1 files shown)


nvpohanh:
@yzh119 please help us to merge this tiny yet important fix. thanks!

yzh119 (approved):
Thanks for the timely fix, also cc @aleozlx @azhurkevich for visibility

nvpohanh:
The team prefers #1351 . We may choose that. thanks!

nvpohanh:
Actually, the team thinks this PR is okay as well.

nvpohanh:
To add context why this is needed:

When all scores are -INF or NaNs, the resulting warpMaxExpertIdx[0] will become 65535, and that causes illegal mem access at this line:

smemExpertTokenCountFull[tokenIdx][warpMaxExpertIdx[0] / ExpertsPerThread] =
            expertTokenCount;

Therefore, we should just route to expert 0 when all scores are just garbage value.

PR #1368, opened 2025-08-03
feat: Support sliding window for persistent kernel 

Edenzzzz:
## 📌 Description
Precision still seems buggy, will try to fix tmr

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests


Edenzzzz:
Only wrong precision when split kv

yzh119:
Sliding window is a little bit tricky with split-kv in terms of boundary computation.

Edenzzzz:
Even when I set `window_tile_idx = kv_tile_idx`, there's still a bit of mismatch in the second case, so other indices could also be problematic
image

PR #1369, opened 2025-08-03
Artifact downloading and single sourced artifact path

cyx-6:
## 📌 Description

This PR adds the features of downloading complete artifacts and makes artifacts path single sourced in python.

cc: @yyihuang @zhyncs 

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/csrc/cudnn_sdpa_kernel_launcher.cu
+++ b/csrc/cudnn_sdpa_kernel_launcher.cu
@@ -29,6 +29,12 @@
 #include "cudnn_sdpa_utils.h"
 #include "pytorch_extension_utils.h"
 
+#ifdef CUDNN_SDPA_CUBIN_PATH
+static const std::string cudnn_sdpa_cubin_path = std::string(CUDNN_SDPA_CUBIN_PATH);
+#else
+static_assert(false, "CUDNN_SDPA_CUBIN_PATH macro is not defined when compiling");
+#endif
+
 namespace flashinfer {
 
 namespace cudnn_sdpa_kernel_launcher {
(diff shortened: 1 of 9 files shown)


csrc/cudnn_sdpa_kernel_launcher.cu
```diff
@@ -77,19 +83,15 @@ enum PrefillType {
...
-  cubin_map[DECODE] = getCubin(
-      "4c623163877c8fef5751c9c7a59940cd2baae02e/fmha/cudnn/cudnn_sm100_fprop_sdpa_decode_d128_bf16",
-      "e7ce0408b4c3a36c42616498228534ee64cab785ef570af5741deaf9dd1b475c");
-
-  cubin_map[PREFILL_DEEPSEEK] = getCubin(
-      "4c623163877c8fef5751c9c7a59940cd2baae02e/fmha/cudnn/"
-      "cudnn_sm100_fprop_sdpa_prefill_d192_bf16",
-      "2190967b8733e193cdcecc054eeb7c2907080a158a33fe7ba2004523a4aff6f9");
+  cubin_map[PREFILL] = getCubin(cudnn_sdpa_cubin_path + "cudnn_sm100_fprop_sdpa_prefill_d128_bf16",
+                                "ff14e8dcfc04d9b3a912dd44056be37d9aa8a85976e0070494ca0cce0524f2a1");
```
yzh119, line 87: Could the sha256 value also be read from python?

flashinfer/artifacts.py
```diff
@@ -0,0 +1,95 @@
...
+
+class MetaInfoHash:
+    TRTLLM_GEN_FMHA: str = (
+        "8c5630020c0452fb1cd1ea7e3b8fdbb7bf94f71bd899ed5b704a490bdb4f7368"
+    )
+    DEEPGEMM: str = "69aa277b7f3663ed929e73f9c57301792b8c594dac15a465b44a5d151b6a1d50"
+
+
+def download_artifacts():
+    env_backup = os.environ.get("FLASHINFER_CUBIN_CHECKSUM_DISABLED", None)
```
yzh119, line 70: Please add file lock for multi-process safety

yzh119 (approved):
With this PR we can pre-download all the cubins with
```
python -m flashinfer --download-cubin
```

In the future we should extend it with:
1. more reliable download that supports resumable download
2. sha256 check for each cubin
3. more user friendly (with progress bar, etc)

flashinfer/artifacts.py
```diff
@@ -0,0 +1,100 @@
...
+class MetaInfoHash:
+    TRTLLM_GEN_FMHA: str = (
+        "8c5630020c0452fb1cd1ea7e3b8fdbb7bf94f71bd899ed5b704a490bdb4f7368"
+    )
+    DEEPGEMM: str = "69aa277b7f3663ed929e73f9c57301792b8c594dac15a465b44a5d151b6a1d50"
+
+
+def download_artifacts() -> bool:
+    env_backup = os.environ.get("FLASHINFER_CUBIN_CHECKSUM_DISABLED", None)
+    os.environ["FLASHINFER_CUBIN_CHECKSUM_DISABLED"] = "1"
```
joker-eph, line 71: Can you document why you're doing this kind of manipulations?

flashinfer/artifacts.py
```diff
@@ -0,0 +1,100 @@
...
+        ArtifactPath.TRTLLM_GEN_GEMM,
+        ArtifactPath.DEEPGEMM,
+    ]:
+        cubin_files += [
+            (kernel + name, extension)
+            for name, extension in get_available_cubin_files(
+                FLASHINFER_CUBINS_REPOSITORY + "/" + kernel
+            )
+        ]
+    pool = ThreadPoolExecutor(4)
```
joker-eph, line 85: Can this "4" be a command line argument with a default value instead?

Also how did you pick 4? Ran some local tuning experiment?

flashinfer/artifacts.py
```diff
@@ -0,0 +1,100 @@
...
+        futures.append(ret)
+    results = []
+    for ret in as_completed(futures):
+        result = ret.result()
+        results.append(result)
+    all_success = all(results)
+    if not env_backup:
+        os.environ.pop("FLASHINFER_CUBIN_CHECKSUM_DISABLED")
+    else:
+        os.environ["FLASHINFER_CUBIN_CHECKSUM_DISABLED"] = env_backup
```
joker-eph, line 98: If anything throws an exception, I suspect the env var won't be reset, is this the most pythonic pattern to handle this? Would some `with` scope handle this better?

flashinfer/artifacts.py
cyx-6: sure, since our sha256 per cubin is now in metainfo.h, we do not have it ready at python side. We have to set `FLASHINFER_CUBIN_CHECKSUM_DISABLED` env var to get cubin without sha256. Otherwise, we should parse the metainfo.h, which is not an efficient method. And here we backup the env var before change, and recover after downloading. I will document this in following pr.

flashinfer/artifacts.py
cyx-6: yes, this `4` is a magic number now. We should have it set with argument `--parallel-task` in command line.

flashinfer/artifacts.py
cyx-6: yes, a `with` scope should be better. We will improve it.

flashinfer/artifacts.py
joker-eph: There is a risk here that there is some issue in the download and we produce a container which wouldn't have the right cubins and need to redownload later.

PR #1289, opened 2025-07-18
refactor: refactor trtllm-gen attention kernel integration code

yzh119:
## 📌 Description

Simplify and unify the interface for trtllm-gen decode/prefill/mla kernels, and add support for shared-kv (in MLA, https://github.com/flashinfer-ai/flashinfer/pull/1273).

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [x] Tests have been added or updated as needed.
- [x] All tests are passing (`unittest`, etc.).

PR #1375, opened 2025-08-04
Unify and modularize decode and prefill test.

weireweire:
## 📌 Description

Unify decode and prefill attention test for trtllm-gen by extract common function. Support fp8/fp4 qkvo scale, o_sf, o_sf offset test.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests


tests/test_trtllm_gen_attention.py
```diff
@@ -0,0 +1,522 @@
...
+        output, output_ref = unpack_compare_nvfp4(
+            output, output_ref, o_sf_scale, o_sf_vec_size
+        )
+        assert o_scale == 1.0
+        rtol, atol = 3e-1, 1e0
+    elif o_dtype == "fp8":
+        rtol, atol = 5e-2, 7e-2
+    else:
+        rtol, atol = 1e-2, 1e-2
+
```
weireweire, line 508: @yzh119 here used to have a `output = output.squeeze(1)` but I don't know why and it has no effect. removed.

tests/test_trtllm_gen_attention.py
```diff
@@ -0,0 +1,522 @@
...
+    if o_dtype != "nvfp4":  # wrapper api does not support fp4 output yet.
+        # test wrapper with trtllm-gen backend
+        wrapper_trtllm_gen = flashinfer.decode.BatchDecodeWithPagedKVCacheWrapper(
+            workspace_buffer, kv_layout, backend="trtllm-gen"
+        )
+        plan_params["q_data_type"] = q.dtype
+        plan_params["kv_data_type"] = kv_cache.dtype
+        wrapper_trtllm_gen.plan(**plan_params)
+        output_wrapper = wrapper_trtllm_gen.run(
+            q.contiguous(),
```
weireweire, line 523: Don't know why need these contiguous too. I kept them, but even remove them can pass the test.

yzh119: This contiguous can not be removed. trtllm requires query tensor to be contiguous to apply gqa packing.

> but even remove them can pass the test.

Because the `q` tensor we created is already contiguous, but if `q` is sliced from another tensor, an explicit `contiguous` is necessary.

tests/test_trtllm_gen_attention.py
yzh119: It was introduced in:

https://github.com/flashinfer-ai/flashinfer/commit/8608c3755024d617566cd2e9f4617f30ca03f8ff

I don't think it's necessary here either.

tests/test_trtllm_gen_decode.py
yzh119, line 1: Seems only mla is checked in this file now, how about renaming it to test_trtllm_gen_mla?

tests/test_trtllm_gen_attention.py
weireweire: yes, I mean in our test it is contiguous. but I'm ok to keep it here to make it safer.

tests/test_trtllm_gen_attention.py
weireweire: actually it's introduced in https://github.com/flashinfer-ai/flashinfer/pull/1289/files#diff-6dda01a6d2a12119b447b6f057a03661f6dd2fd49e9af835cc4cc3e6b402f605R200

if not necessary I'll drop it.

tests/test_trtllm_gen_decode.py
weireweire: renamed. Tried to keep the git history. @yyihuang to move the test to test_trtllm_gen_attention.py later as it's also attention.

weireweire:
It's better to reuse this test for recently added sink support. please consider the conflict.

yzh119 (approved):
LGTM, thank you @weireweire for the refactor.

> It's better to reuse this test for recently added sink support. please consider the conflict.

Yes let's leave that for future work.

weireweire:
We'd better merge this soon, more conflict will coming

weireweire:
rebased and fixed a prefill blocker. please merge soon, @yzh119 
And I didn't see sink test in old attention test. is it added?

flashinfer/prefill.py
```diff
@@ -2096,7 +2096,8 @@ def run(
                 lse=lse,
             )
         else:
-            assert self._plan_info is not None, "plan info is not initialized"
+            if self._backend == "cudnn":
```
Anerudhan: This check is not needed for cudnn. See Line 2073. This is already in else part. I added this since run_args has _plan_info and mypy was complaining.

weireweire: ok, filter out trtllm-gen check then.

yzh119 (approved):

yzh119:
> And I didn't see sink test in old attention test. is it added?

They are tested in https://github.com/flashinfer-ai/flashinfer/blob/main/tests/test_attention_sink.py and https://github.com/flashinfer-ai/flashinfer/blob/main/tests/test_attention_sink_blackwell.py, we can unify them later.

PR #1379, opened 2025-08-04
Removes MPI dependency from MNNVL AllReduce

pranavm-nvidia:
## 📌 Description

This change removes the MPI dependency from the MNNVL AllReduce kernel, using `torch.distributed` instead. 

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Reviewer Notes

There are two hacks here I want to call out:

1. I needed to change the paths in the JIT environment since they were pointing to non-existent directories. I'm not sure if this is a real bug or whether I missed a step when setting something up.
2. The unit test is now tied to SLURM, which might not be desirable, but I am unsure how to generalize it or how the existing CI works with multi-node.

### DIFF
--- a/flashinfer/jit/env.py
+++ b/flashinfer/jit/env.py
@@ -51,18 +51,20 @@ def _get_workspace_dir_name() -> pathlib.Path:
 FLASHINFER_WORKSPACE_DIR = _get_workspace_dir_name()
 FLASHINFER_JIT_DIR = FLASHINFER_WORKSPACE_DIR / "cached_ops"
 FLASHINFER_GEN_SRC_DIR = FLASHINFER_WORKSPACE_DIR / "generated"
-_package_root = pathlib.Path(__file__).resolve().parents[1]
-FLASHINFER_DATA = _package_root / "data"
-FLASHINFER_INCLUDE_DIR = _package_root / "data" / "include"
-FLASHINFER_CSRC_DIR = _package_root / "data" / "csrc"
-# FLASHINFER_SRC_DIR = _package_root / "data" / "src"
-FLASHINFER_TVM_BINDING_DIR = _package_root / "data" / "tvm_binding"
-FLASHINFER_AOT_DIR = _package_root / "data" / "aot"
+# TODO (pranavm): Check if this is right?
+# Why were these pointing to non-existent directories? Must be a copy missing somewhere?
+_package_root = pathlib.Path(__file__).resolve().parents[2]
+FLASHINFER_DATA = _package_root
+FLASHINFER_INCLUDE_DIR = _package_root  / "include"
+FLASHINFER_CSRC_DIR = _package_root  / "csrc"
+# FLASHINFER_SRC_DIR = _package_root  / "src"
+FLASHINFER_TVM_BINDING_DIR = _package_root  / "tvm_binding"
+FLASHINFER_AOT_DIR = _package_root  / "aot"
 CUTLASS_INCLUDE_DIRS = [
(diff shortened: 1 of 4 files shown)


flashinfer/jit/env.py
```diff
@@ -51,18 +51,20 @@ def _get_workspace_dir_name() -> pathlib.Path:
...
-FLASHINFER_DATA = _package_root / "data"
-FLASHINFER_INCLUDE_DIR = _package_root / "data" / "include"
-FLASHINFER_CSRC_DIR = _package_root / "data" / "csrc"
-# FLASHINFER_SRC_DIR = _package_root / "data" / "src"
-FLASHINFER_TVM_BINDING_DIR = _package_root / "data" / "tvm_binding"
-FLASHINFER_AOT_DIR = _package_root / "data" / "aot"
+# TODO (pranavm): Check if this is right?
+# Why were these pointing to non-existent directories? Must be a copy missing somewhere?
+_package_root = pathlib.Path(__file__).resolve().parents[2]
+FLASHINFER_DATA = _package_root
```
yongwww, line 57: cc: @yzh119

flashinfer/jit/env.py
```diff
@@ -51,18 +51,20 @@ def _get_workspace_dir_name() -> pathlib.Path:
...
 FLASHINFER_JIT_DIR = FLASHINFER_WORKSPACE_DIR / "cached_ops"
 FLASHINFER_GEN_SRC_DIR = FLASHINFER_WORKSPACE_DIR / "generated"
-_package_root = pathlib.Path(__file__).resolve().parents[1]
-FLASHINFER_DATA = _package_root / "data"
-FLASHINFER_INCLUDE_DIR = _package_root / "data" / "include"
-FLASHINFER_CSRC_DIR = _package_root / "data" / "csrc"
-# FLASHINFER_SRC_DIR = _package_root / "data" / "src"
-FLASHINFER_TVM_BINDING_DIR = _package_root / "data" / "tvm_binding"
-FLASHINFER_AOT_DIR = _package_root / "data" / "aot"
+# TODO (pranavm): Check if this is right?
```
yzh119, line 54: Hi @pranavm-nvidia , we shouldn't change the logic here.

When you install flashinfer (using pip install), the data folder will be created:
https://github.com/flashinfer-ai/flashinfer/blob/40471bf5db5dff25376aac3357b3842e4bd487aa/pyproject.toml#L64-L86

all of the original paths such as `_package_root / "include"` or `_package_root / "3rdparty"` will be gone when you build the packages.

If you develop locally in editable mode, you will not see the difference because your package root is the repository path, but if you install flashinfer package on pypi on a new environment, you will find there is no `include`/`3rdparty` path in flashinfer package directory, and that why there is a `data` path designed to handle these source code dependencies.

pranavm-nvidia: Thanks, I had a feeling it was a setup issue on my end. What's the correct way to test local changes then? Do I always need to reinstall the package when I make a change? I was so far just setting `PYTHONPATH`

yzh119: > What's the correct way to test local changes then? Do I always need to reinstall the package when I make a change?

You can make editable installations:
```
pip install -e . -v
```
and you don't have to reinstall the package for source code changes (because it's editable).

hjjq:
Are there plans to merge this PR? It's needed for MNNVL-AllReduce integration into vLLM and sglang. @pranavm-nvidia @yzh119

PR #1385, opened 2025-08-05
[DO NOT MERGE] Separating the k_cache and v_cache for trtllm-gen attention interface.

yzh119:
## 📌 Description

Some frameworks separate the storage of `k_cache` and `v_cache`, and the coupled kv-cache interface `[num_pages, 2, num_heads, page_size, head_dim]` will be hard to use for them.

This PR adds support for decoupled `k_cache`/`v_cache`, while keeping backward compatibility with original interface.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/csrc/trtllm_fmha_kernel_launcher.cu
+++ b/csrc/trtllm_fmha_kernel_launcher.cu
@@ -186,14 +186,18 @@ inline Data_type torch_dtype_to_tllm_data_type(at::ScalarType dtype) {
 inline bool is_4bit(Data_type data_type) { return data_type == Data_type::DATA_TYPE_E2M1; }
 
 void trtllm_paged_attention_decode(at::Tensor out, std::optional<at::Tensor> out_scale_factor,
-                                   at::Tensor query, at::Tensor key_value_cache,
+                                   at::Tensor query, at::Tensor key_cache, at::Tensor value_cache,
                                    at::Tensor workspace_buffer, at::Tensor block_tables,
                                    at::Tensor seq_lens, int64_t max_kv_len, double bmm1_scale,
                                    double bmm2_scale, double o_sf_scale, int64_t o_sf_vec_size,
                                    int64_t o_sf_start_index, int64_t window_left,
                                    int64_t sm_count) {
   auto q_data_type = torch_dtype_to_tllm_data_type(query.scalar_type());
-  auto kv_data_type = torch_dtype_to_tllm_data_type(key_value_cache.scalar_type());
+  auto kv_data_type = torch_dtype_to_tllm_data_type(key_cache.scalar_type());
+  TORCH_CHECK_EQ(key_cache.dim(), value_cache.dim());
+  for (int i = 0; i < key_cache.dim(); i++) {
+    TORCH_CHECK_EQ(key_cache.size(i), value_cache.size(i));
+  }
   auto o_data_type = torch_dtype_to_tllm_data_type(out.scalar_type());
   // NOTE(Zihao): query is [B, Q, H, D]
   // where Q is the number of query tokens per request, used in MTP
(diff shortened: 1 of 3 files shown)


csrc/trtllm_fmha_kernel_launcher.cu
```diff
@@ -204,7 +204,7 @@ void trtllm_paged_attention_decode(at::Tensor out, std::optional<at::Tensor> out
   int sum_seq_q = batch_size * q_len_per_request;
   int num_qo_heads = query.size(2);
   // Multiply by two for FP4 tensor as it is stored as UINT8 dtype. Assume the dim is even.
-  int head_dim_kv = is_4bit(kv_data_type) ? key_value_cache.size(-1) * 2 : key_value_cache.size(-1);
+  int head_dim_kv = is_4bit(kv_data_type) ? key_cache.size(-1) * 2 : key_cache.size(-1);
   int head_dim_qk = is_4bit(q_data_type) ? query.size(-1) * 2 : query.size(-1);
```
weireweire: should check head_dim_q, head_dim_k, head_dim_v then

csrc/trtllm_fmha_kernel_launcher.cu
```diff
@@ -213,30 +213,23 @@ void trtllm_paged_attention_decode(at::Tensor out, std::optional<at::Tensor> out
...
-                  std::to_string(key_value_cache.size(1)));
-  bool share_kv_cache = key_value_cache.size(1) == 1;
-  int page_size = key_value_cache.size(-2);
-  int num_kv_heads = key_value_cache.size(-3);
+  int page_size = key_cache.size(-2);
+  int num_kv_heads = key_cache.size(-3);
   int max_num_blocks_per_seq = block_tables.size(-1);
-  int num_pages_in_mem_pool = key_value_cache.size(0) * key_value_cache.size(1);
+  int num_pages_in_mem_pool =
+      (key_cache.data_ptr() == value_cache.data_ptr()) ? key_cache.size(0) :  // shared key & value
```
weireweire: k_cache and v_cache is always two tensor in python API impl, why they can be equal? is there some feature can make k and v same tensor?

csrc/trtllm_fmha_kernel_launcher.cu
```diff
@@ -186,14 +186,14 @@ inline Data_type torch_dtype_to_tllm_data_type(at::ScalarType dtype) {
...
 void trtllm_paged_attention_decode(at::Tensor out, std::optional<at::Tensor> out_scale_factor,
-                                   at::Tensor query, at::Tensor key_value_cache,
+                                   at::Tensor query, at::Tensor key_cache, at::Tensor value_cache,
                                    at::Tensor workspace_buffer, at::Tensor block_tables,
                                    at::Tensor seq_lens, int64_t max_kv_len, double bmm1_scale,
                                    double bmm2_scale, double o_sf_scale, int64_t o_sf_vec_size,
                                    int64_t o_sf_start_index, int64_t window_left,
                                    int64_t sm_count) {
   auto q_data_type = torch_dtype_to_tllm_data_type(query.scalar_type());
-  auto kv_data_type = torch_dtype_to_tllm_data_type(key_value_cache.scalar_type());
```
weireweire, line 196: if we only use key shape, better check k_cache and v_cache have the same shape

csrc/trtllm_fmha_kernel_launcher.cu
weireweire: for original share_kv_cache flag and only use in decode?

csrc/trtllm_fmha_kernel_launcher.cu
```diff
@@ -273,29 +266,21 @@ void trtllm_paged_attention_context(at::Tensor out, std::optional<at::Tensor> ou
                                               std::to_string(head_dim_kv) + " and " +
                                               std::to_string(head_dim_vo));
   int max_num_blocks_per_seq = block_tables.size(-1);
-  int num_pages_in_mem_pool = key_value_cache.size(0) * key_value_cache.size(1);
```
weireweire, line 276: seems now the case that kv-cache sharing is gone.

flashinfer/prefill.py
```diff
@@ -3018,6 +3018,12 @@ def trtllm_batch_context_with_kv_cache(
     out: Union[torch.Tensor, FP4Tensor]
         output torch.Tensor or FP4Tensor.
     """
+
+    if isinstance(kv_cache, tuple):
+        k_cache, v_cache = kv_cache
+    else:
+        k_cache, v_cache = kv_cache.unbind(dim=1)
```
weireweire: is this valid when shape[1] ==1 , I see that in docstring

csrc/trtllm_fmha_kernel_launcher.cu
yzh119: > k_cache and v_cache is always two tensor in python API impl, why they can be equal

For MLA, they are equal

> for original share_kv_cache flag and only use in decode?

Yes `share_kv_cache` means mla decode.

csrc/trtllm_fmha_kernel_launcher.cu
yzh119: The prefill(context), we don't expect kv-cache sharing (which is only expected to be used in MLA).

weireweire: OK，then why you added is_shared_kv here.

yzh119:
This PR is now part of https://github.com/flashinfer-ai/flashinfer/pull/1389

PR #1394, opened 2025-08-05
refactor: reuse workspace for bmm_fp8

ttyio:
## 📌 Description

refactor: reuse workspace for bmm_fp8

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/flashinfer/gemm.py
+++ b/flashinfer/gemm.py
@@ -1104,7 +1104,7 @@ def execute_cudnn_gemm_fp4_graph(
         UIDs.O_UID.value: c_final,
     }
 
-    if graph.get_workspace_size() > DEFAULT_WORKSPACE_SIZE:
+    if workspace_buffer.numel() < graph.get_workspace_size():
         workspace_buffer = torch.empty(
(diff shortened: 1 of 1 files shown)


flashinfer/gemm.py
```diff
@@ -1550,12 +1547,12 @@ def bmm_fp8(
             dtype=dtype,
         )
 
+    workspace_buffer = _get_cache_buf(
```
yzh119, line 1550: Why not separating the cache of different backends and allocate cudnn workspace inside `execute_cudnn_gemm_with_per_tensor_q_graph` after `graph.get_workspace_size()` is calculated.

ttyio: I noticed that all the methods in `gemm.py` use `DEFAULT_WORKSPACE_SIZE`. I assumed the design choice was because different tactics might require different workspace sizes, and using a sufficiently large default helps avoid frequent memory reallocations when the workspace changes. If that’s incorrect, I can update all the methods to request a workspace size matching the current tactic’s requirement instead. Thanks!

Issue #1064, opened 2025-05-19
AOT Refactor


abcdabcd987:
Tracking ticket for the implementation of unifying JIT and AOT code path (#791).

High level approach:

1. Capture the building instructions for each torch cpp extension in a `JitSpec` object.
2. Generate ninja file for each `JitSpec`. Notably, the output target should be `$name/$name.so`, i.e., including the name of the torch cpp extension.
3. Generate a wrapper ninja file that use `subninja` to include all ninja files.
4. After building all torch cpp extensions, copy the `.so` files to `data/aot/` and ship the package.

Tasks:

- [x] JitSpec and Generate ninja file. https://github.com/flashinfer-ai/flashinfer/pull/1065
    - [x] bugfix for _get_glibcxx_abi_build_flags https://github.com/flashinfer-ai/flashinfer/pull/1080
- [x] Remove `load_cuda_ops()`. https://github.com/flashinfer-ai/flashinfer/pull/1066
- [x] Refactor to let all modules to return JitSpec. https://github.com/flashinfer-ai/flashinfer/pull/1069
- [x] Replace `parallel_load_modules()` with `build_jit_specs()`. https://github.com/flashinfer-ai/flashinfer/pull/1070
- [x] Override workspace dir. https://github.com/flashinfer-ai/flashinfer/pull/1073
- [x] Script to build all AOT ops. https://github.com/flashinfer-ai/flashinfer/pull/1067
- [x] Packaging. https://github.com/flashinfer-ai/flashinfer/pull/1075
- [x] Remove has_prebuilt_ops. https://github.com/flashinfer-ai/flashinfer/pull/1076
- [ ] Use `/dev/shm` for lock.

yzh119:
Some remaining tasks:
- [x] fix the fp8 flashattention-3 template aot compilation #1087

yzh119:
cc @MasterJH5574 , it's also worth checking whether tvm jit is broken or not.

MasterJH5574:
> cc [@MasterJH5574](https://github.com/MasterJH5574) , it's also worth checking whether tvm jit is broken or not.

@yzh119 TVM JIT is good for now.  I'll send a patch later to update with the latest TTM FFI refactor.

PR #1404, opened 2025-08-07
Fix signature mismatch in flashinfer_ops.cpp

842974287:
## 📌 Description

Make the function signature in flashinfer_ops.cpp the same as https://github.com/flashinfer-ai/flashinfer/blob/4c4276a71c99e8be0f871c7b619ac9fbdd54ffa2/csrc/batch_prefill.cu#L200.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests


yzh119:
flashinfer_ops.cpp should no longer required after #1064 , did you see it being used anywhere?

If not, we should remove this file instead.

842974287:
@yzh119 I think it's because we are using some older versions previously and still using flashinfer_ops.cpp to build. Going to upgrade to 0.2.10 release. Thanks for taking a look!

PR #1388, opened 2025-08-05
feat(install): allow skipping nvshmem pip install via env var

EmilienM:
## 📌 Description

Some systems, particularly RHEL-based distributions using the NVIDIA CUDA repository, provide `nvshmem` as a system-level package (e.g., RPM).
In such cases, installing the `nvidia-nvshmem-cu12` package from pip is unnecessary and can cause conflicts.

This commit introduces the `SKIP_NVSHMEM_PIP` environment variable.
When this variable is set to "1", the `setup.py` script will remove `nvidia-nvshmem-cu12` from the list of dependencies, allowing the system-provided version to be used instead.

## 🔍 Related Issues

Fixes #1387

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [X] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [X] I have installed the hooks with `pre-commit install`.
- [X] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [X] Tests have been added or updated as needed.
- [X] All tests are passing (`unittest`, etc.).

PR #1409, opened 2025-08-07
install: remove nvidia-cudnn-12 from package dependency

yzh119:
## 📌 Description

The dependency is not required because:
1. For x86_64 user environments that installs pytorch from pypi, `nvidia-cudnn-12` is already a torch dependency (https://github.com/pytorch/pytorch/blob/3f1636ebef9b45e8a3cb0eb20d327ee6acb74be0/.github/scripts/generate_binary_build_matrix.py#L44-L95), and flashinfer depends on torch, there is no need to specify another `nvidia-cudnn-12` dependency in flashinfer.
2. For user environments where `nvidia-cudnn-12` is not a torch dependency (like in nvidia's pytorch container), cudnn is installed at system wide and cudnn-frontend will rely on cudnn system installation instead of through `nvidia-cudnn-12`.

## 🔍 Related Issues

`nvidia-nvshmem-12` package is a similar redundant dependency: https://github.com/flashinfer-ai/flashinfer/pull/1388

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,7 +61,6 @@ install_requires = [
     "pynvml",
     "einops",
     "nvidia-nvshmem-cu12",
-    "nvidia-cudnn-cu12",
     "nvidia-cudnn-frontend>=1.13.0",
 ]
 generate_build_meta({})


EmilienM:
thanks!

yongwww (approved):

PR #1416, opened 2025-08-08
Fix missing v_scale for prefill wrapper.

weireweire:
## 📌 Description

We do need k,v scale !=1 for llama3 fp4 model.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/flashinfer/prefill.py
+++ b/flashinfer/prefill.py
@@ -2002,7 +2002,12 @@ class BatchPrefillWithPagedKVCacheWrapper:
             ]
 
         self._cached_module.paged_run(*run_args)
-
+        if v_scale is not None:
+            # TODO(Zihao): fused into kernel
+            if out.itemsize == 1:
+                out = (out.to(float) * v_scale).to(out.dtype)
+            else:
+                out *= v_scale
         return (out, lse) if return_lse else out
 
     run_return_lse = functools.partialmethod(run, return_lse=True)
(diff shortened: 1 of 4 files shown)


flashinfer/prefill.py
```diff
@@ -2002,7 +2002,12 @@ def run(
             ]
 
         self._cached_module.paged_run(*run_args)
-
+        if v_scale is not None:
+            # TODO(Zihao): fused into kernel
+            if out.itemsize == 1:
```
nvpohanh: Is this to detect FP8 and FP4 output? If so, I think we should be more explicit (checking FP8 and UINT8) or at least add a comment.

flashinfer/prefill.py
```diff
@@ -2002,7 +2002,12 @@ def run(
...
+                out = (out.to(float) * v_scale).to(out.dtype)
```
nvpohanh: `out` is a torch tensor, so I think we should use `out.to(torch.float32)`. Your code works but that relies on an implicit conversion from python type to torch dtype.

flashinfer/prefill.py
weireweire: actually only fp8. fp4 is not supported by wrapper and cannot do the convert here. changed to fp8 check.

flashinfer/prefill.py
weireweire: changed

yzh119 (approved):
LGTM

PR #1427, opened 2025-08-08
Sink attention AoT

nandor:
## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests


flashinfer/aot.py
```diff
@@ -371,24 +338,33 @@ def gen_all_modules(
...
-        from .comm import gen_trtllm_comm_module, gen_vllm_comm_module
+        try:
+            from .comm.nvshmem import gen_nvshmem_module
 
-        if has_sm100:
-            jit_specs.append(gen_trtllm_comm_module())
-        jit_specs.append(gen_vllm_comm_module())
+            jit_specs.append(gen_nvshmem_module())
+        except ImportError:
+            pass
```
abcdabcd987: I guess we should raise here. Because `if add_comm` explicitly request for comm.

flashinfer/aot.py
```diff
@@ -371,24 +338,33 @@ def gen_all_modules(
...
+
+        try:
+            from .comm import gen_trtllm_comm_module, gen_vllm_comm_module
+
+            if has_sm100:
+                jit_specs.append(gen_trtllm_comm_module())
+            jit_specs.append(gen_vllm_comm_module())
+        except ImportError:
+            pass
```
abcdabcd987: Same here.

I guess maybe a reasonable approach would be split the `add_comm` into two flags.

flashinfer/aot.py
```diff
@@ -109,47 +107,31 @@ def gen_fa3(
...
-            dtype_o=dtype_o,
-            head_dim_qk=head_dim_qk,
-            head_dim_vo=head_dim_vo,
-            pos_encoding_mode=0,
-            use_sliding_window=use_sliding_window,
-            use_logits_soft_cap=use_logits_soft_cap,
-            use_fp16_qk_reduction=False,
-        ),
-        gen_batch_prefill_module(
+    for dtype_idx in [torch.int32, torch.int64]:
```
abcdabcd987: is int64 index used?

yzh119 (approved):

flashinfer/aot.py
```diff
@@ -539,10 +505,10 @@ def main():
         # True,
     ]
     add_comm = False
-    add_gemma = True
```
DevashishLal-CB, line 542: Any reason why act and gemma modules are disabled by default ?

PR #1444, opened 2025-08-10
fix: remote redundant zero_init from trtllm-gen attn

yyihuang:
## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/tests/test_trtllm_gen_decode.py
+++ b/tests/test_trtllm_gen_decode.py
@@ -465,7 +465,7 @@ def test_trtllm_batch_decode_mla(
 
     # Allocate workspace buffer
     # todo(Yingyi): calculate the actual size of workspace buffer
-    workspace_buffer = torch.empty(128 * 1024 * 1024, dtype=torch.int8, device=device)
+    workspace_buffer = torch.zeros(128 * 1024 * 1024, dtype=torch.int8, device=device)
 
     bmm1_log2_scale_tensor = (
         torch.tensor(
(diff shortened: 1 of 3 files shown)


tests/test_trtllm_gen_decode.py
```diff
@@ -483,21 +483,22 @@ def test_trtllm_batch_decode_mla(
...
-        qk_rope_head_dim=qk_rope_head_dim,
-        block_tables=block_tables,
-        seq_lens=seq_lens_tensor,
-        max_seq_len=max_seq_len,
-        bmm1_scale=scale / ((128 + 64) ** 0.5),
-        bmm2_scale=1.0,
-        bmm1_scale_log2_tensor=bmm1_log2_scale_tensor,
-        bmm2_scale_tensor=bmm2_scale_tensor,
-    )
+    for _ in range(3):
```
fzyzcjy, line 486: optional nit: shall we provide diff input for each run, i.e. do the for loop for the whole testing logic. then we can avoid the case that, maybe the attn does not work on such case, and it seems to pass test only b/c we provide the exact same input

fzyzcjy (approved):

tests/test_trtllm_gen_decode.py
yyihuang: We could test if one workspace buffer init would work , even for multiple attention runs by this for loop.

fzyzcjy: that looks pretty reasonable

yzh119 (approved):
For more context:
We only need setting semaphores before the first run, and they will be reset to zeros in each trtllm-gen attention kernel run. No need to reset it each time we call the trtllm-gen attention kernel.

tests/test_trtllm_gen_decode.py
```diff
@@ -465,7 +465,7 @@ def test_trtllm_batch_decode_mla(
 
     # Allocate workspace buffer
     # todo(Yingyi): calculate the actual size of workspace buffer
-    workspace_buffer = torch.empty(128 * 1024 * 1024, dtype=torch.int8, device=device)
+    workspace_buffer = torch.zeros(128 * 1024 * 1024, dtype=torch.int8, device=device)
```
yzh119, line 468: After second thought, I think the best solution is to explicitly call a `empty_` function inside the __init__ function (if it's a wrapper), for standalone function call, we need to clearly explain the assumption in the docstring.

yyihuang: Will do in the next PR👌

PR #1453, opened 2025-08-11
feat: enable trtllm-gen attn speculative decoding verify by decode

yyihuang:
## 📌 Description

decode with q_len > 1

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests


tests/test_trtllm_gen_attention.py
```diff
@@ -196,6 +204,51 @@ def get_last_page_len(seq_lens, page_size):
     return kv_last_page_len
 
 
+def assert_close_with_mismatch_tolerance(
```
yzh119: consider moving this to conftests

yyihuang: fixed

tests/test_trtllm_gen_attention.py
```diff
@@ -570,14 +617,37 @@ def test_trtllm_batch_decode(
...
         )
         # v_scale, o_scale in wrapper is emulated by multiplying output by v_scale instead of fused into kernel.
         if v_scale == o_scale == 1.0:
             assert (output_wrapper == output).all()
         else:
-            torch.testing.assert_close(
-                output.float(), output_wrapper.float(), rtol=1e-1, atol=1e-1
-            )
+            # todo(Yingyi): fix precision issue with this test
+            if not (
```
yzh119, line 627: I still have deep concern about the special handling of precision here, a more fundamental solution could be a PrecisionManager to centrialize the handling of tolerance for different data types.

Also, let's add a flashinfer.testing.assert_allclose() function where there could be arguments such as max mismatched elements, it will be fallback to torch.testing.assert_allclose when mismatched elements is not specified.

yzh119 (approved):
Would be great if you can work on the remaining item I mentioned in later PRs.

yyihuang:
should unblock https://github.com/sgl-project/sglang/issues/9471 now

tests/test_trtllm_gen_attention.py
```diff
@@ -517,6 +532,30 @@ def test_trtllm_batch_decode(
     wrapper_ref.plan(**plan_params)
     output_ref = wrapper_ref.run(ref_q, ref_kv_cache)
 
+    if q_len_per_req > 1:
```
weireweire, line 535: we should use if else right? otherwise the above wrapper run is useless when q_len_per_req>1

Besides, the params is very similar, we should able to reuse.

please confirm this so I can change it.

PR #1463, opened 2025-08-11
fix: remove redundant zero_init reverted by #1459

yyihuang:
## 📌 Description

The duplicate zero_init should be fixed. But we got some crash reported from DLFW. So we revert it in #1459 and make 0.2.11.post1. 

After this fix, **workspace buffer passed into any trtllm-gen attn interface must be zero-initialized**.

This PR is to enable this optimization. It should be merged and released only after these two are tested.
- https://github.com/sgl-project/sglang/pull/9065
- https://github.com/vllm-project/vllm/pull/22603

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests


nvpohanh:
cc @weireweire @elvischenv we may need to do the same in vLLM to zero-init the workspace

yyihuang:
@nvpohanh Vllm impl is in https://github.com/vllm-project/vllm/pull/22603, with our tests passed locally. Could you help review and test it?

I think we could merge DLFW update before merging at FI otherwise it would introduce some errors.

csrc/trtllm_fmha_kernel_launcher.cu
```diff
@@ -146,13 +146,18 @@ void trtllm_paged_attention_launcher(
         use_multi_block ? TileScheduler::Static : TileScheduler::Persistent;
     runner_params.mMultiCtasKvMode = use_multi_block;
 
-    size_t num_semaphores =
-        round_up(batch_size * num_qo_heads, 8);  // align multiCtasKvScratchPtr to 16 bytes
+    size_t max_batch_size = 8192;   // todo(Yingyi): get from dlfw
+    size_t max_num_qo_heads = 256;  // todo(Yingyi): get from dlfw, in total 8MB
+    size_t num_semaphores = round_up(max_batch_size * max_num_qo_heads, 8);  // max 8MB
+    // size_t num_semaphores =
```
nvpohanh: Can we delete the commented out code?

include/flashinfer/semaphore_utils.cuh
```diff
@@ -48,6 +48,34 @@ cudaError_t zero_gmem_semaphore_launcher(T* semaphore, int size, bool enable_pdl
   return cudaSuccess;
 }
 
+template <typename T>
```
nvpohanh: should we remove these if we no longer use them?

tests/test_trtllm_gen_decode.py
```diff
@@ -583,3 +595,46 @@ def test_trtllm_batch_decode_mla(
             print("output:", output)
             print("o_ref:", o_ref)
             raise e
+
+
+if __name__ == "__main__":
```
yzh119: Can we remove them (if any of the combination is missing from pytest, just adding them to pytest parameters).

yyihuang: removed.

yzh119 (approved):
LGTM.

PR #1472, opened 2025-08-12
feat: Enable multiple fused-moe backends

amirkl94:
## 📌 Description

Change CUTLASS kernels JitSpec name as to not collide with `trtllm` FusedMoE name.
Currently, in a python session we can't both `trtllm` FusedMoE backend, and CUTLASS backend at the same time, this PR enables this support.

### DIFF
--- a/flashinfer/fused_moe/core.py
+++ b/flashinfer/fused_moe/core.py
@@ -236,7 +236,7 @@ def gen_cutlass_fused_moe_sm100_module(use_fast_build: bool = False) -> JitSpec:
         raise RuntimeError(f"Failed to generate Cutlass kernels: {e}") from e
 
     return gen_jit_spec(
-        "fused_moe_sm100",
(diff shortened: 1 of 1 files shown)


flashinfer/fused_moe/core.py
```diff
@@ -236,7 +236,7 @@ def gen_cutlass_fused_moe_sm100_module(use_fast_build: bool = False) -> JitSpec:
         raise RuntimeError(f"Failed to generate Cutlass kernels: {e}") from e
 
     return gen_jit_spec(
-        "fused_moe_sm100",
+        "fused_moe_cutlass_sm100",
```
yzh119, line 239: Yes we should never collide for any two extensions, can you also rename the other module as `fused_moe_trtllm_sm100`?

yzh119 (approved):

PR #1396, opened 2025-08-06
gpt-oss: Add MXFP8 x MXFP4 CUTLASS MOE for SM100 and BF16 x MXFP4 CUTLASS for SM90 + SwigluBias Activation

djmmoss:
## 📌 Description

This PR adds MXFP8 x MXFP4 CUTLASS MOE with SwigluBias for SM100 GPUs.

It also adds BF16 x MXFP4 CUTLASS MOE with SwigluBias for SM 90

## 🔍 Related Issues

N/A

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [x] Tests have been added or updated as needed.
- [ ] All tests are passing (`unittest`, etc.).

PR #1475, opened 2025-08-12
tuner: Trtllm-gen Fp4 MoE Autotunner

IwakuraRein:
## 📌 Description

- Update the `AutoTuner` , `OptimizationProfile`, and `DynamicTensorSpec`. 
  - `DynamicTensorSpec` can take multiple input tensors.
  - The `tensor_initializers` in `DynamicTensorSpec` defines the initialization method for dynamic tensors. Before they were all zero-initialized and this will cause IMA in trtllm-gen's routing kernels.
- Add `DtypeTrtllmGen` in `flashinfer/fused_moe/core.py`
- Add autotuner to trtllm-gen fp4 moe.
- Relax the check on `hidden_states_scales` in trtllm-gen fp4 moe.  It doesn't need to be 1D.
  - **When autotuning, it must be 2D.**

### TODOs

- Unify the launcher for both trtllm-gen fp8 and fp4 moe.
- After unifying the launchers, add auto tuner to fp8 moe.
- If routing is DeepSeek V3, there will be illegal memory access. A WAR is to limit the search space to [8 ... 1024] 

### Performance

B200, clock speed locked at 1500mhz, 1000 warmups, 1000 iterations, `mxfp4 x mxfp8`

| num_tokens | wo tuner  | with tuner | diff   |
|------------|-----------|------------|--------|
|          1 |     0.042 |      0.042 |  0.00% |
|          2 |     0.057 |      0.057 |  0.00% |
|          4 |     0.081 |      0.081 |  0.00% |
|          8 |     0.114 |      0.105 |  7.89% |
|         16 |     0.201 |      0.183 |  8.96% |
|         32 |     0.274 |      0.246 | 10.22% |
|         64 |     0.348 |      0.308 | 11.49% |
|        128 |      0.41 |      0.365 | 10.98% |
|        256 |     0.548 |      0.429 | 21.72% |
|        512 |     0.576 |      0.453 | 21.35% |
|       1024 |     0.651 |      0.578 | 11.21% |

For `nvfp4 x nvfp4` and `mxfp4 x bf16`, there is no significant perf gain.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests


nvpohanh:
cc @amirkl94

csrc/trtllm_fused_moe_kernel_launcher.cu
```diff
@@ -1122,7 +1124,84 @@ std::vector<at::Tensor> trtllm_fp4_block_scale_moe(
       output1_scales_gate_scalar, output2_scales_scalar, num_experts, top_k, n_group, topk_group,
       intermediate_size, local_expert_offset, local_num_experts, routed_scaling_factor,
       tile_tokens_dim, routing_method_type, do_finalize, *mRunner, mDtypeAct, mDtypeWeights,
-      moeConfigIndex, enable_pdl, output);
+      config_index, enable_pdl, output);
+}
+
+inline btg::Dtype get_dtype(int64_t const dtype) {
+  switch (dtype) {
+    case 0:
```
aleozlx: this currenly finds the type by Dtype's Uid bits.
would it be more maintainable if we simply pass strings (so in code review, the case will more obviously match the type name) and document at the binding interface that the available options are in DtypeDecl.h Dtype

aleozlx: the current approach has the advantage of being checked early using class DtypeTrtllmGen(IntEnum). so no objections, just raising options

flashinfer/fused_moe/core.py
```diff
@@ -65,6 +66,43 @@ class RoutingMethodType(IntEnum):
     Unspecified = 5
 
 
+# NOTE(siyuan): Need to keep this in sync with the counterpart defined in include/flashinfer/trtllm/batched_gemm/trtllmGen_bmm_export/trtllm/gen/DtypeDecl.h
+class DtypeTrtllmGen(IntEnum):
+    Bfloat16 = (0,)
```
aleozlx: or we can borrow the bits formation from TLLM_ENCODE_DTYPE, and allowing the deletion of get_dtype() conversion

e.g.

```python
class Dtype(IntEnum):
    def __new__(cls, block_format_bit, signed_bit, integer_bit, num_bits, uid):
        value = (block_format_bit << 24) | (signed_bit << 20) | (integer_bit << 16) | (num_bits << 8) | uid
        obj = int.__new__(cls, value)
        obj._value_ = value
        return obj
    
    # keep the values in sync with include/flashinfer/trtllm/batched_gemm/trtllmGen_bmm_export/trtllm/gen/DtypeDecl.h
    Bfloat16 = (0, 1, 0, 16, 0)
    Bool     = (0, 0, 1, 1,  1)
    E2m1     = (1, 1, 0, 4,  2)
    E2m3     = (1, 1, 0, 6,  3)
    E3m2     = (1, 1, 0, 6,  4)
    E4m3     = (0, 1, 0, 8,  5)
    E5m2     = (0, 1, 0, 8,  6)
    Fp16     = (0, 1, 0, 16, 7)
    Fp32     = (0, 1, 0, 32, 8)
    Int8     = (0, 1, 1, 8,  9)
    Int32    = (0, 1, 1, 32, 10)
    Int64    = (0, 1, 1, 64, 11)
    MxE2m1   = (1, 1, 0, 4,  12)
    MxE4m3   = (1, 1, 0, 8,  13)
    UE8m0    = (0, 0, 0, 8,  14)
    UInt8    = (0, 0, 1, 8,  15)
    UInt16   = (0, 0, 1, 16, 16)
    UInt32   = (0, 0, 1, 32, 17)
    UInt64   = (0, 0, 1, 64, 18)
    UInt128  = (0, 0, 1, 128, 19)
    Void     = (0, 1, 0, 0,  20)
```

csrc/trtllm_fused_moe_kernel_launcher.cu
IwakuraRein: @aleozlx Thanks for the suggestion! What about defining a new enum in `trtllm_fused_moe_kernel_launcher.cu` and use macros to map it to `btg::Dtype`. Then expose to python.

aleozlx: sure . that improves readability. or we can use `class Dtype(IntEnum)` i proposed in the other thread and remove another conversion.  and hopefully `__new__` serves to hide impl details in a way that user won't rely on the value (compared to a separate function). although the value changes can cause breakage thru the cpp interface, so we just have to align on something early.

flashinfer/fused_moe/core.py
```diff
@@ -1481,6 +1840,7 @@ def trtllm_fp4_block_scale_routed_moe(
     routing_method_type: int = 0,
     do_finalize: bool = True,
     enable_pdl: Optional[bool] = None,
+    tune_max_num_tokens: int = 1024,
```
aleozlx: shall we always append? so if users are using the optional arguments positionally, they won't break. 

or i guess we should have put optionals after `, /, ...` in hind sight to prevent them from being used positionally, which is easier to break

tests/bench_trtllm_gen_fused_moe_autotuner.py
```diff
@@ -0,0 +1,216 @@
+import argparse
```
yzh119: We should move it to `benchmarks` instead

flashinfer/fused_moe/core.py
```diff
@@ -868,12 +918,268 @@ def trtllm_gen_fused_moe_sm100_module() -> JitSpec:
...
+    x: torch.Tensor, scale: Optional[torch.Tensor]
+) -> DtypeTrtllmGen:
+    hidden_size = x.shape[-1]
+    if x.dtype == torch.uint8:
+        hidden_size *= 2
+    if x.dtype == torch.bfloat16:
+        dtype = DtypeTrtllmGen.Bfloat16
+    elif x.dtype == torch.float8_e4m3fn:
+        dtype = DtypeTrtllmGen.E4m3 if scale is None else DtypeTrtllmGen.MxE4m3
+    elif x.dtype == torch.uint8:
```
yzh119: Just a minor note: we should also take care of `torch.float4_e2m1x2` for torch 2.8+

IwakuraRein: Thanks. But I didn't see `float4_e2m1fn_x2` is used anywhere in the flashinfer? Will we add it altogether in the future?

yzh119: Not yet, we need to prepare for it when frameworks are all upgrading to torch 2.8.

It could be done in later PRs.

yzh119:
There are some conflicts with main branch after #1396 got merged, would you mind rebasing?

flashinfer/fused_moe/core.py
```diff
@@ -879,6 +952,241 @@ def get_trtllm_moe_sm100_module():
...
+            assert topk_ids.shape[0] == num_tokens, (
+                "topk_ids's first dimension must be batch size."
+            )
+            assert expert_weights.shape[0] == num_tokens, (
+                "expert_weights's first dimension must be batch size."
+            )
+            assert hidden_states.shape[0] == num_tokens, (
+                "hidden_states's first dimension must be batch size."
+            )
+            assert hidden_states_scale is None or (
```
weireweire, line 1123: This assertion make vllm gpt-oss fail. In that case the scale is `tensor([1., 1., 1.,  ..., 1., 1., 1.], device='cuda:0',dtype=torch.float8_e4m3fn)`

IwakuraRein: @weireweire I created [this pr](https://github.com/vllm-project/vllm/pull/23209) to fix this. Currently [vllm's flashinfer tag](https://github.com/vllm-project/vllm/blob/31282401b6c5c4271eb558e6af3f1d42b1d157c8/setup.py#L688) is 0.2.12. I can mark this pr as ready and increment the flashinfer tag to 0.2.13

flashinfer/fused_moe/core.py
```diff
@@ -879,6 +952,241 @@ def get_trtllm_moe_sm100_module():
...
+                hidden_states_scale.dim() == 2
+                and hidden_states_scale.shape[0] == num_tokens
+            ), "hidden_states_scale's first dimension must be batch size"
+
+            # TODO(siyuan): support fp8
+            moe_op.trtllm_fp4_block_scale_moe(
+                routing_logits.to(torch.bfloat16),
```
IwakuraRein, line 1130: This line actually breaks DeepSeek v3 routing. I have left suggest changes in #1494

PR #1481, opened 2025-08-13
Add python API for masked grouped gemm

kaixih:
This PR introduces `grouped_gemm_nt_masked`, which simplifies the usage of masked grouped GEMM. It calls into `MaskedBatchedMatmulCuteDSL` but requires only the minimal set of parameters, with all other parameters deduced from the provided tensors.

The new function includes a docstring that explains the expected input shapes and behavior.

cc. @yzh119 @wenscarl @kushanam

### DIFF
--- a/flashinfer/cute_dsl/utils.py
+++ b/flashinfer/cute_dsl/utils.py
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2023 by FlashInfer team.
+Copyright (c) 2025 by FlashInfer team.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -23,6 +23,7 @@ def is_cute_dsl_available() -> bool:
         importlib.util.find_spec("cutlass") is not None
         and importlib.util.find_spec("cutlass.cute") is not None
     )
+
 
 def get_cutlass_dtype(dtype: str) -> cutlass.dtype:
     dtype_map = {


flashinfer/cute_dsl/blockscaled_gemm.py
```diff
@@ -2702,3 +2702,91 @@ def dtype(cutlass_dtype):
         )
 
         return c_tensor_gpu
+
+
+def _dtype_to_cutlass_dtype(dtype: str) -> cutlass.dtype:
```
yongwww: Probably we could consider moving the helper func to `flashinfer/flashinfer/utils.py`

fzyzcjy (approved):
seems the new API is more aligned with DeepGEMM, then lgtm (as long as test pass, code details are clean, etc) since this may make integration more unified

flashinfer/cute_dsl/blockscaled_gemm.py
kaixih: Done. PTAL.

kaixih:
Yes, the api is motivated by the masked deepgemm api, e.g `grouped_gemm_nt_f8f8bf16_masked` from sgl.

flashinfer/utils.py
```diff
@@ -21,6 +21,7 @@
 from enum import Enum
 from typing import Callable, Dict, Iterable, Optional, Sequence, Tuple, Union
 
+import cutlass
```
yzh119: cutlass is not package level dependency yet (`nvidia-cutlass-dsl` only supports python 3.12 at this moment so we can not specify it as install_requires in setup.py

flashinfer/utils.py
```diff
@@ -666,6 +667,19 @@ def get_shuffle_matrix_sf_a_row_indices(
     return row_indices
 
 
+def get_cutlass_dtype(dtype: str) -> cutlass.dtype:
```
yzh119: Consider putting it under `flashinfer/cute_dsl/utils.py` instead (we assume user have installed `nvidia-cutlass-dsl` by their own to use this module), `utils.py` is too general (before we make `nvidia-cutlass-dsl` part of our project dependencies).

cc @yongwww

kaixih: Done. PTAL.

flashinfer/utils.py
kaixih: Done. PTAL.

kaixih:
> We should also add `flashinfer.cute_dsl` to
> 
> https://github.com/flashinfer-ai/flashinfer/blob/1d29426da1a4a6e98cd0978cb03408265f94e2aa/pyproject.toml#L40
> 
> , we forgot to do so in #1331

Done.

yyihuang (approved):

PR #1487, opened 2025-08-14
feat: Centralize env vars in flashinfer.json with caching

yongwww:
## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Reviewer Notes

cc: @yzh119


flashinfer.json
yzh119, line 1: This file should not exist in codebase, it should be created (the first time user import flashinfer) and placed under `~/.config/flashinfer.json` (or user specified path) instead.

flashinfer/__init__.py
```diff
@@ -14,6 +14,120 @@
...
+
+        # Write the filled configuration
+        with open(cached_config_path, "w") as f:
+            json.dump(filled_config, f, indent=2)
+
+    except Exception as e:
+        print(f"Warning: Could not save filled config to {cached_config_path}: {e}")
+
+
+def _set_env_var(var_name, var_info):
```
yzh119: We should not pollute environment variable like this, once the config is loaded from disk, we should be able to access them like `flashinfer.env.CUDA_HOME` etc.

yongwww: glad you brought it up; I totally agree.

joker-eph:
Can you fill the description with all the context and motivation for the feature?

yongwww:
> Can you fill the description with all the context and motivation for the feature?

I’ve added the description; I should have done it earlier. Thanks for the comment!

PR #1488, opened 2025-08-14
fix: update cutedsl masked moe gemm

yyihuang:
## 📌 Description

initial PR: https://github.com/flashinfer-ai/flashinfer/pull/1331
fix following PR #1481

For problem shape (l, m, k), k is leading:
- We assume DLFW passed in packed tensors with 8-bit storage (eg, int8, fp4x2, fp8) with shape (l, m, k//2), which takes half as the shape at leading dim.
- Therefore we introduce is_packed params (default True) to indicate if it's packed or not. If it's packed, we deduce the leading dim by *2.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/flashinfer/cute_dsl/blockscaled_gemm.py
+++ b/flashinfer/cute_dsl/blockscaled_gemm.py
@@ -2753,7 +2753,9 @@ def grouped_gemm_nt_masked(
     n, _, _ = b_torch.shape
 
     if ab_dtype == "float4_e2m1fn":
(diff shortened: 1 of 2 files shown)


kaixih (approved):
Chatted offline a bit. It looks good to me. The only thing is I think we no longer need the is_packed, since the kernel behavior is always assuming packed inputs for nvfp4.

kaixih (approved):

PR #1497, opened 2025-08-15
Update moe to benchmark

nv-yunzheq:
## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/benchmarks/routines/flashinfer_benchmark_utils.py
+++ b/benchmarks/routines/flashinfer_benchmark_utils.py
@@ -49,7 +49,7 @@ output_column_dict = {
         "top_k",
         "n_group",
         "topk_group",
-        "routing_method_type",
+        "routing_method",
         "use_shuffled_weight",
         "weight_layout",
         "use_routing_scales_on_input",
--- a/benchmarks/routines/flashinfer_benchmark_utils.py
+++ b/benchmarks/routines/flashinfer_benchmark_utils.py
@@ -89,6 +89,7 @@ benchmark_apis = {
         "trtllm_fp4_block_scale_moe",
         "trtllm_fp8_block_scale_moe",
         "trtllm_fp8_per_tensor_scale_moe",
+        "cutlass_fused_moe",
     ],
 }
 
--- a/benchmarks/routines/flashinfer_benchmark_utils.py
+++ b/benchmarks/routines/flashinfer_benchmark_utils.py
@@ -40,6 +40,22 @@ output_column_dict = {
         "mma_sm",
         "use_128x4_sf_layout",
     ],
+    "moe": [
+        "backend",
+        "num_tokens",
+        "hidden_size",
+        "intermediate_size",
+        "num_experts",
+        "top_k",
+        "n_group",
+        "topk_group",
+        "routing_method_type",
+        "use_shuffled_weight",
+        "weight_layout",
+        "use_routing_scales_on_input",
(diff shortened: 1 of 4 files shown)


benchmarks/routines/flashinfer_benchmark_utils.py
```diff
@@ -40,6 +40,22 @@
         "mma_sm",
         "use_128x4_sf_layout",
     ],
+    "moe": [
+        "backend",
```
bkryu: Can you actually move the backend in "moe" and "attention" to "general" to avoid redundancy?

benchmarks/routines/moe.py
```diff
@@ -0,0 +1,1378 @@
...
+        global_sfs.append(a_global_sf)
+
+    result_quant_a = torch.stack(quant_a)
+    result_sfs = torch.stack(sfs)
+    result_global_sfs = torch.stack(global_sfs)
+
+    return result_quant_a, result_sfs, result_global_sfs
+
+
+def calculate_moe_tflops(
```
bkryu, line 370: I'm generally not familiar with the MOE operation. 

Were the FLOPs and TB/s calculations brought over from existing benchmark files? If not, let's ask someone who might be familiar.

benchmarks/routines/moe.py
```diff
@@ -0,0 +1,1378 @@
...
+        cur_res["backend"] = backend
+        cur_res["num_tokens"] = num_tokens
+        cur_res["hidden_size"] = hidden_size
+        cur_res["intermediate_size"] = intermediate_size
+        cur_res["num_experts"] = num_experts
+        cur_res["top_k"] = top_k
+        # Routing method/weight layout not applicable; leave defaults
+        cur_res["n_group"] = None
+        cur_res["topk_group"] = None
+        cur_res["routing_method"] = None
```
bkryu: Since `cur_res` is a defaultdict, I'm finding that it is okay to remove these lines for brevity. I'll also be making that change in a PR I am preparing

benchmarks/routines/moe.py
yzh119: The moe tflops calculation looks correct to me.

benchmarks/routines/moe.py
```diff
@@ -0,0 +1,1510 @@
...
+        0 if routing_logits_dtype is None else routing_logits_dtype.itemsize
+    )
+    input_bytes = (
+        # Count hidden states once; kernels typically reuse inputs for multiple experts
+        num_tokens * hidden_size * input_bytes_per_element
+        + num_tokens * num_experts * routing_logits_bytes
+    )
+
+    # Weight memory (reuse weights across tokens by grouping tokens per expert)
+    # Assume each active expert's weights are read once per run.
```
yzh119, line 446: This means we will not skip expert weights if they are not activated (in small batch size case), is that correct?

nv-yunzheq: No. We will skip expert weights if they are not activated.
The code below uses `num_active_experts * weight_bytes_per_expert`. So, only experts being activated is considered. 
For trtllm-gem moe, since the routing is inside the function, we don't know which experts being activated, we just return -1 for now.

yzh119: Got it, so the `num_active_experts` need to be specified explicitly.

PR #1503, opened 2025-08-18
feat: integrate xqa attention backend

qsang-nv:
## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests


flashinfer/__init__.py
```diff
@@ -125,3 +125,4 @@
     VariableBlockSparseAttentionWrapper as VariableBlockSparseAttentionWrapper,
 )
 from .utils import next_positive_power_of_2 as next_positive_power_of_2
+from .xqa import get_xqa_module as get_xqa_module
```
jiahanc: Shall we also do `from .xqa_decode import xqa_decode as xqa_decode` so user can use `flashinfer.xqa_decode` instead of `flashinfer.xqa_decode.xqa_decode` given there is only 1 function in `xqa_decode.py`

qsang-nv: Done.

yzh119: CI failed at: https://ci.tlcpack.ai/blue/organizations/jenkins/flashinfer-ci/detail/PR-1503/4/pipeline

flashinfer/xqa.py
yzh119, line 1: Can we merge this file with the `xqa_decode` module?

Personally I would prefer renaming the `xqa_decode` module to `xqa` and put all functions related to xqa there.

yzh119 (commented):
Please add modules to https://github.com/flashinfer-ai/flashinfer/blob/cc5ab77370dd9a489357a47e34315d9a8f3ad5fb/pyproject.toml#L41 otherwise they will not be packaged.

flashinfer/xqa.py
qsang-nv: Done, flashinfer/xqa_decode is deleted and the decode function is in xqa.py.

tests/test_xqa.py
```diff
@@ -0,0 +1,349 @@
...
+    seq_len_list.fill_(seq_len)
+
+    kv_cache_scale = torch.ones(1, dtype=torch.float32, device="cuda")
+
+    nb_seq = nb_k_heads * batch_size
+    nb_semaphores = round_up(nb_seq, 2) + 2 + nb_seq + 2
+
+    semaphores = torch.zeros(nb_semaphores, dtype=torch.uint32, device="cuda")
+
+    scratch_size = 256 << 20
```
jdebache, line 286: Could we expose a function to get the size of the workspace required for a given set of max problem dimensions? The size but also the alignment of it are delicate:

flashinfer/xqa.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+def gen_xqa_module(
+    use_fp16: bool,
+    token_per_page: int,
+    head_size: int,
+    head_grp_size: int,
+    use_sliding_window: bool,
+) -> JitSpec:
+    if use_fp16:
```
jdebache, line 47: Could we also add support for fp8 arithmetic / kv-cache?

tests/test_xqa.py
qsang-nv: This size setup comes from xqa repo's test file https://github.com/NVIDIA/TensorRT-LLM/blob/main/cpp/kernels/xqa/test/test.cpp#L268. There is no such function for now in xqa repo and I will talk to xqa team to see how to get one and add in following PRs.

flashinfer/xqa.py
qsang-nv: Will add fp8 support in following PRs.

tests/test_xqa.py
jdebache: I think you could use the logic from TensorRT-LLM: https://github.com/NVIDIA/TensorRT-LLM/blob/4541655e5f10a2da034fd2e8bd484a7373914bce/cpp/tensorrt_llm/kernels/xqaDispatcher.cpp#L167

yzh119 (approved):
For the next steps, let's add:
1. more flexible layouts (reading strides from tensors), and support both HND and NHD.
2. fp8 xqa attention

tests/test_xqa.py
yzh119: yes it can be exposed as a `get_xqa_workspace` function to python.
But in general I think having a upperbound (e.g. 256mb here) for the workspace size is the best strategy as frameworks usually don't change the space they allocated for specific kernels.

PR #1508, opened 2025-08-18
Support cuda<12.8 built for trtllm_allreduce_fusion.

strgrb:
## 📌 Description

I want to use trtllm_allreduce_fusion with cuda<12.8 for hopper gpu in sglang, so I wrap fp4 code with cuda version check to make it compiled.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/include/flashinfer/comm/trtllm_allreduce_fusion.cuh
+++ b/include/flashinfer/comm/trtllm_allreduce_fusion.cuh
@@ -1,4 +1,5 @@
 #include <cooperative_groups.h>
+#include <cuda.h>
 #include <cuda_bf16.h>
 #include <cuda_fp16.h>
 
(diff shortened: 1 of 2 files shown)


yuan-luo:
This PR can make https://github.com/sgl-project/sglang/pull/7946/ to support hopper.

yongwww:
> This PR can make sgl-project/sglang#7946 to support hopper.

out of curiosity — this change looks like it’s for CUDA version compatibility. Why does it specifically help Hopper here? if fp4 is the reason, then I think checking SM instead of cuda version would make more sense

yuan-luo:
> > This PR can make [sgl-project/sglang#7946](https://github.com/sgl-project/sglang/pull/7946) to support hopper.
> 
> out of curiosity — this change looks like it’s for CUDA version compatibility. Why does it specifically help Hopper here? if fp4 is the reason, then I think checking SM instead of cuda version would make more sense

I didn't express clearly. Currently sglang only supports trt-llm-allreduce_fusion for sm100 and this PR makes it support sm90. With the similar fix like this PR, FlashInfer can support trtllm_allreduce for sm90, so as sglang can be modified to support both sm90 and sm100 for trt_llm_allreduce(sglang #7946).

yzh119:
More precisely, this PR adds guard for cuda<12.8.

trtllm allreduce should work with hopper + cuda 12.8 before this PR, but there are still users that relies on Hopper + cuda 12.6.

PR #1512, opened 2025-08-18
flashinfer_benchmark QoL Improvements and Attention FP8 Support

bkryu:
## 📌 Description
Current PR:
- Adds `case_tag` and `generate_repro_command` flags to flashinfer_benchmark.py for annotating outputs and generating reproducer commands for a test case. See example use case at [sample_testlist_output.csv](https://github.com/flashinfer-ai/flashinfer/blob/83e25fb135aa32519f0e089e1cff183347b34089/benchmarks/samples/sample_testlist_output.csv)
- Adds support for benchmarking "out-of-wrapper" trtllm-gen attention kernels via backend "trtllm-gen-native". Perf can be compared to "in-wrapper" kernels from e.g. fa2, trtllm-gen.
- Adds support for benchmarking FP8 attention in decode, prefill, and MLA attention variants.
- Adds cutlass backend for benchmarking bmm_fp8
- Updates README and sample testlist & output in `benchmark/samples`

**Changes only in the `benchmark` directory. No library code changes.**

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests


benchmarks/routines/attention.py
```diff
@@ -57,7 +57,15 @@ def parse_attention_args(line, parser):
         required=False,
         nargs="+",
         default=["fa2"],
-        choices=["fa2", "fa2_tc", "fa3", "cudnn", "cutlass", "trtllm-gen"],
+        choices=[
+            "fa2",
+            "fa2_tc",
+            "fa3",
+            "cudnn",
+            "cutlass",
```
yzh119, line 65:
```suggestion
            "cutlass_fmha",
```

fa3 implementation relies on cutlass, to make it less confusing we can call it `cutlass_fmha`, means fmha implementation from cutlass.

bkryu: Thanks Zihao, for the feedback. The reason why I named the cutlass backend is because "cutlass" is a backend option for [Ragged Prefill](https://github.com/flashinfer-ai/flashinfer/blob/main/flashinfer/prefill.py#L2675-L2680) and [MLA](https://github.com/flashinfer-ai/flashinfer/blob/main/flashinfer/mla.py#L181-L186). As such, I lean towards keeping the backend as cutlass.

However I do understand the potential source of confusion. I have pushed [another commit](https://github.com/flashinfer-ai/flashinfer/pull/1512/commits/35e85d16a5c03543dd75623abefd2309157edbe6) that does some cleanup of the README file and [states in the support matrix](https://github.com/flashinfer-ai/flashinfer/blob/46eaf1a935bd86a21242450b1154435e0e0404bf/benchmarks/README.md#tester-attention-backend-support-matrix) that `cutlass` means _FMHA implemented with CUTLASS_ while `fa3` means _FlashAttention-3 implemented with CUTLASS_.

Would this help clear out the confusion a bit?

(Side note: I'm noticing that the docstring for ragged prefill is missing info that cutlass could be supported.)

benchmarks/README.md
```diff
@@ -7,260 +7,90 @@ The aim of `flashinfer_benchmark.py` is to provide a single framework for benchm
...
 | `--num_iters`            | Number of iterations for performance measurement                                                           |
 | `--dry_run_iters`        | Number of warmup iterations                                                                                |
 | `--no_cuda_graph`        | Disable CUDA graph to execute kernels outside of the graph.                                                |
-| `--allow_output_mismatch`| Continue testing even if outputs don't match between backends                                              |
 | `--refcheck`             | Verify outputs match between different backends                                                            |
+| `--allow_output_mismatch`| Continue testing even if outputs don't pass refcheck                                              |
```
yzh119, line 82: Just curious which benchmark relies on this flag?

yzh119 (approved):
LGTM in general, thanks for the update!

benchmarks/routines/attention.py
yzh119: Yes I think the fundamental issue is to view cutlass as a kernel provider or a framework to implement kernels. We should have a more standard way of backend naming and your changes look good to me so far.

benchmarks/README.md
bkryu: By default, when`--refcheck` is provided and there is a mismatch to reference, `flashinfer_benchmark.py` will raise an error and terminate without reporting any perf numbers. 

If `--allow_output_mismatch` is provided, an error message will be printed, and perf numbers will be printed as well without raising an error. 

The flag becomes more useful in batch testing situations where a testlist containing multiple test cases is provided: adding the allow output mismatch flag will allow the test to continue on instead of exiting upon seeing a mismatch.

PR #1529, opened 2025-08-21
feat: pass sm_count as param for fp4_masked_gemm

yyihuang:
## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests


flashinfer/cute_dsl/blockscaled_gemm.py
```diff
@@ -2456,6 +2457,9 @@ def __init__(
         self._max_active_clusters = hardware_info.get_max_active_clusters(
             self._cluster_shape_mn[0] * self._cluster_shape_mn[1]
         )
+        # if sm_count is provided, use it instead of the default max value
+        if sm_count > 0:
+            self._max_active_clusters = sm_count
```
fzyzcjy: qq: will we meet 2sm per cluster and thus need half

fzyzcjy (approved):

flashinfer/cute_dsl/blockscaled_gemm.py
```diff
@@ -2456,6 +2457,9 @@ def __init__(
         self._max_active_clusters = hardware_info.get_max_active_clusters(
             self._cluster_shape_mn[0] * self._cluster_shape_mn[1]
         )
+        # if sm_count is provided, use it instead of the default max value
+        if sm_count > 0:
+            self._max_active_clusters = min(self._max_active_clusters, sm_count)
```
fzyzcjy: my guess is maybe it is `sm_count / num_sm_per_cluster`, but not sure

yyihuang: will try to figure out 2cta optimization at https://github.com/flashinfer-ai/flashinfer/blob/62fcef5669fe0c224f4fe5535a9f8a2d52a90ff1/flashinfer/cute_dsl/blockscaled_gemm.py#L933

benchmarks/bench_cute_dsl_blockscaled_gemm.py
```diff
@@ -120,6 +120,7 @@ def enumerate_m_grouped_masked():
 
 # Copy and modified from test_cute_dsl_blockscaled_gemm.py, may extract common logic later if needed
 def create_data(num_groups, max_m, expected_m_per_group, n, k):
```
yzh119: device could be part of input arguments

flashinfer/cute_dsl/blockscaled_gemm.py
```diff
@@ -2416,6 +2416,7 @@ def __init__(
         sf_vec_size: int,
         mma_tiler_mn: Tuple[int, int],
         cluster_shape_mn: Tuple[int, int],
+        sm_count: int = 0,
```
yzh119: Don't set a default value here.

flashinfer/cute_dsl/blockscaled_gemm.py
yzh119: It might change `self._max_active_clusters` to some value not compatible with current cluster configs.

Suggested changes:
```python
        self._max_active_clusters = hardware_info.get_max_active_clusters(
            min(self._cluster_shape_mn[0] * self._cluster_shape_mn[1], sm_count)
        )
```

flashinfer/cute_dsl/blockscaled_gemm.py
```diff
@@ -2582,6 +2586,7 @@ def get_cute_dsl_compiled_masked_gemm_kernel(
     sf_vec_size: int,
     mma_tiler_mn: Tuple[int, int],
     cluster_shape_mn: Tuple[int, int],
+    sm_count: int = 0,
```
yzh119: remove default value

flashinfer/cute_dsl/blockscaled_gemm.py
```diff
@@ -2767,6 +2773,7 @@ def grouped_gemm_nt_masked(
         cluster_shape_mn (Tuple[int, int], optional): Shape of the CTA cluster (ClusterM, ClusterN). Default: (1, 1).
         alpha_dtype (str, optional): Data type for alpha scaling factors.
         alpha (torch.Tensor, optional): Optional 1D tensor of shape (l,) containing per-batch scaling factors. Perform per-batch scaling out = alpha * out.
+        sm_count (int, optional): Number of SMs to use. Default: 0.
```
yzh119: Remove default value.

flashinfer/cute_dsl/blockscaled_gemm.py
```diff
@@ -2794,6 +2801,7 @@ def grouped_gemm_nt_masked(
 
     mma_tiler_mn = kwargs.get("mma_tiler_mm", (128, 128))
     cluster_shape_mn = kwargs.get("cluster_shape_mm", (1, 1))
+    sm_count = kwargs.get("sm_count", 0)
```
yzh119:
```suggestion
    sm_count = kwargs.get("sm_count", get_num_sm(a_torch.device))
```

yzh119 (commented):
Add another function `get_num_sm` under flashinfer.utils:

```python
@functools.cache
def get_num_sm(device: torch.device) -> int:
    return torch.cuda.get_device_properties(device).multi_processor_count
```

It must be cached because `get_device_properties` will call a time-consuming CUDA API.

flashinfer/cute_dsl/blockscaled_gemm.py
yzh119: Also it's better to make `num_sm` an explicit argument of this function, with type `Optional[int]` and default value None.

flashinfer/cute_dsl/blockscaled_gemm.py
yyihuang: updated.

fzyzcjy:
after my fix above, I can confirm that the SM is reduced as expected

image

zhyncs (approved):

flashinfer/cute_dsl/blockscaled_gemm.py
```diff
@@ -2454,8 +2454,11 @@ def __init__(
 
         # Compute max active clusters on current device
         hardware_info = cutlass.utils.HardwareInfo()
-        self._max_active_clusters = hardware_info.get_max_active_clusters(
-            min(self._cluster_shape_mn[0] * self._cluster_shape_mn[1], sm_count)
+        self._max_active_clusters = min(
+            hardware_info.get_max_active_clusters(
+                self._cluster_shape_mn[0] * self._cluster_shape_mn[1]
+            ),
+            sm_count,
```
yzh119, line 2461: Does this work when cluster size is greater than 1?

yzh119 (commented):
Also `num_sm` should not be a compile time args, we can fix them in later PRs as this one have already been merged..

flashinfer/cute_dsl/blockscaled_gemm.py
fzyzcjy: I think no, that is what I am worried above https://github.com/flashinfer-ai/flashinfer/pull/1529#discussion_r2289901605

yzh119: Will create a followup PR in the daytime

Issue #1425, opened 2025-08-08
Perf drop again of norm


xiaoqi35:
The end-2-end time of norm cpu API droped between flashinfer-python-0.2.7.post1 and flashinfer-python-0.2.6.post1.
Because prefill stage executed with disabled cuda-graph, so we also do care about cpu cost, especially in classfiy model.

Seems the most same with closed issues https://github.com/flashinfer-ai/flashinfer/issues/960

xiaoqi35:
test on A100:
[test_norm_end2end.json](https://github.com/user-attachments/files/21681024/test_norm_end2end.json)

left col is 0.2.6 , right col is 0.2.7 
![Image](https://github.com/user-attachments/assets/2f4fabe5-1c50-49ba-89f8-f596fc8b0a41)

yzh119:
I checked the [changelog](https://github.com/flashinfer-ai/flashinfer/compare/v0.2.6.post1...v0.2.7), and it seems `device_support_pdl` might increase cpu-side execution time (because it calls `get_compute_capability`, which calls a cuda runtime api internally).

This value should be cached to avoid calling cuda runtime apis.

@xiaoqi35 do you have interest in trying this change:

Adding a `@functools.cache` to this function:
https://github.com/flashinfer-ai/flashinfer/blob/7c4541228d1255311f472d882182b841ef8fd4f2/flashinfer/utils.py#L209-L213

```python
@functools.cache
def get_compute_capability(device: torch.device) -> Tuple[int, int]:
    if device.type != "cuda":
        raise ValueError("device must be a cuda device")
    return torch.cuda.get_device_capability(device.index)
```

cc @cyx-6 in case you have time double checking this issue.

yzh119:
@xiaoqi35 feel free to re-open it if the issue persists.

PR #1535, opened 2025-08-21
Add sm check for sm100 only cutlass/trtllm kernel

ttyio:
## 📌 Description

Add sm check for sm100 only cutlass/trtllm kernel

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/flashinfer/gemm.py
+++ b/flashinfer/gemm.py
@@ -65,13 +65,14 @@ from .utils import (
     is_float8,
     register_custom_op,
     register_fake_op,
+    get_compute_capability,
 )
 
 DEFAULT_WORKSPACE_SIZE = 32 * 1024 * 1024
 
 
 def _match_sm_version(sm_version: str):
-    major, minor = torch.cuda.get_device_capability()
+    major, minor = get_compute_capability(torch.device("cuda"))
     device_arch = f"{major * 10 + minor}"
     return device_arch == sm_version
 
@@ -449,7 +450,7 @@ def fp8_gemm_sm100(
         runners.append(_cudnn_gemm_fp8_runner())
 
     if len(runners) == 0:
-        major, minor = torch.cuda.get_device_capability()
+        major, minor = get_compute_capability(torch.device("cuda"))
         raise ValueError(f"No valid runner found for current device sm{major}{minor}")
 
     tuner = AutoTuner.get()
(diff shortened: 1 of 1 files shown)


flashinfer/gemm.py
```diff
@@ -70,6 +70,12 @@
 DEFAULT_WORKSPACE_SIZE = 32 * 1024 * 1024
 
 
+def _match_sm_version(sm_version: str):
+    major, minor = torch.cuda.get_device_capability()
```
yzh119: Consider calling https://github.com/flashinfer-ai/flashinfer/blob/669ff33409d721a111b1b3c2ebca68ffa21e68de/flashinfer/utils.py#L212 instead, which cached this value and should be more efficient.

We observe significant CPU side overhead of the cuda runtime api to query device properties, see #1425 and #1456

ttyio: updated, thanks!

flashinfer/gemm.py
```diff
@@ -65,11 +65,18 @@
     is_float8,
     register_custom_op,
     register_fake_op,
+    get_compute_capability,
 )
 
 DEFAULT_WORKSPACE_SIZE = 32 * 1024 * 1024
 
 
+def _match_sm_version(sm_version: str):
```
yzh119: I would encourage adding another device argument to this function.

flashinfer/gemm.py
```diff
@@ -65,11 +65,18 @@
...
+    major, minor = get_compute_capability(torch.device("cuda"))
```
yzh119: Get compute capability of the given device.

flashinfer/gemm.py
```diff
@@ -434,13 +441,18 @@ def fp8_gemm_sm100(
     runners = []
     # No e5m2 for cutlass
     is_e5m2 = a.dtype == torch.float8_e5m2 or b.dtype == torch.float8_e5m2
-    if "cutlass" in runner_names and not is_e5m2:
+    is_sm100 = _match_sm_version("100")
```
yzh119: Pass a's device to this functions

ttyio: Ah, make sense, thanks for the catch!

flashinfer/gemm.py
ttyio: updated, thanks!

yzh119 (approved):

PR #1559, opened 2025-08-24
bugfix: fix persistent attention kernel correctness on blackwell

yzh119:
## 📌 Description

An explicit `__syncthreads` is required for guarantee correctness on blackwell, otherwise unittests will fail.

Also add `test_batch_attention` unittests to CI.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Reviewer Notes

cc @Edenzzzz @happierpig

### DIFF
--- a/include/flashinfer/attention/persistent.cuh
+++ b/include/flashinfer/attention/persistent.cuh
@@ -483,7 +483,7 @@ struct BlockBatchReductionPersistent {
       const typename KTraits::IdType num_packed_qo_len, const uint_fastdiv gqa_group_size,
       const uint32_t num_kv_heads, const typename KTraits::IdType* indptr,
       const typename KTraits::IdType* o_indices, uint8_t* smem PROFILER_CLOSURE_FUNC_PARAMS) {
-    __syncthreads();
+    __syncthreads();  // NOTE(Zihao): required for guarantee correctness on blackwell
     using DTypeIn = typename KTraits::DTypeIn;
     using DTypeO = typename KTraits::DTypeO;
     using IdType = typename KTraits::IdType;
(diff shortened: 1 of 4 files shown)


include/flashinfer/attention/persistent.cuh
```diff
@@ -483,6 +483,7 @@ struct BlockBatchReductionPersistent {
       const typename KTraits::IdType num_packed_qo_len, const uint_fastdiv gqa_group_size,
       const uint32_t num_kv_heads, const typename KTraits::IdType* indptr,
       const typename KTraits::IdType* o_indices, uint8_t* smem PROFILER_CLOSURE_FUNC_PARAMS) {
+    __syncthreads();
```
Edenzzzz: guard with >= SM100 flag?

yzh119: Does it hurt performance on earlier architectures?

Edenzzzz: may be negligible but makes it more clear?

yzh119: It's still not clear why it only happens on blackwell (it might also exist hopper and ampere, but not found because of different hardware latency etc). I tend to leave a note here instead.

spectrometerHBH:
IMO, a warpsync is required at the start of the loop
https://github.com/flashinfer-ai/flashinfer/blob/main/include/flashinfer/attention/persistent.cuh#L514

warp_sync_state loads from s_smem v_smem at the end of the loop
pipeline prelogue writes to v_smem at the beginning of the loop
They don't have any warpsync in between to guarantee loop-carried dependency across works

yongwww (approved):

Issue #1243, opened 2025-07-11
Beginner Getting Started Tasks


skylee-01:
Please assign me a beginner-friendly task to start as a contributor.

Edenzzzz:
https://github.com/flashinfer-ai/flashinfer/issues/675

yzh119:
Hi @skylee-01 , yes as @Edenzzzz mentioned you can check #675  for the a list of features we are calling from community.
The blackwell related kernel development are mostly driven by teams at NVIDIA, feel free to work on other items if you have interest.

One of the todo item is to accelerate top-k with radix sort (our current top-k sampling is binary-search based and optimized for large-k, for smaller-k, radix top-k is faster), you can check this paper https://arxiv.org/html/2501.14336v1 for reference.

skylee-01:
Thank you for the guidance from both teachers. @yzh119 @Edenzzzz

PrithviElancherran:
I opened a CUDA test PR for short-sequence prefill (qo_len=1) to improve edge-case coverage. 
PR: https://github.com/flashinfer-ai/flashinfer/pull/1515 - @PrithviElancherran

sricketts:
Moving forward, feel free to search for the `good first issue` label to find beginner tasks. Closing this ticket.

PR #1561, opened 2025-08-24
[DRAFT] feat: support radix-based top-k sampling algorithm

JasonJ2021:
## 📌 Description
This commit introduces a new radix sort-based top-k sampling algorithm,`radik_sampling_from_probs()`, which is ported from https://github.com/leefige/radik.git. Following the design of https://arxiv.org/abs/2501.14336.

## 🔍 Related Issues

https://github.com/flashinfer-ai/flashinfer/issues/1243

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

   - Add tests to prove correctness of radix-sorting based top-k sampling algorithms

## Reviewer Notes
The current radix-sort based top-k implementation only supports cases where k <= 1024. I'm still looking for a way to integrate it into the existing top-k sampling. An ideal scenario would be to use the radix implementation when k <= 1024 and the batch size is small, and use the existing implementation for all other cases.

The radix implementation requires an additional buffer to store intermediate results for sorting. For example, for top-k sampling task(batch_size=1024, vocab_size=128512, k=10), it require ~1GB workspacebuffer. To accelerate memory allocation, the current implementation pre-allocates a 64MB buffer. 

Compared to the existing top-k sampling implementation, the radix-sort based approach significantly reduces end-to-end latency, especially with smaller batches. I selected some benchmark data with a relatively small batch size; the complete benchmark data is in the attachment. Benchmarks were conducted on H20 GPU. 

Original Implementation of **top_k_sampling_from_probs:**
> vocab_size: 128512, batch_size: 1, distrib: normal_distribution(std=1), deterministic: False, k: 10, duration: 523.30 us, effective bandwidth: 0.98 GB/s
vocab_size: 128512, batch_size: 1, distrib: normal_distribution(std=1), deterministic: False, k: 100, duration: 340.35 us, effective bandwidth: 1.51 GB/s
vocab_size: 128512, batch_size: 1, distrib: normal_distribution(std=1), deterministic: False, k: 1000, duration: 195.71 us, effective bandwidth: 2.63 GB/s
vocab_size: 128512, batch_size: 16, distrib: normal_distribution(std=1), deterministic: False, k: 10, duration: 1412.06 us, effective bandwidth: 5.82 GB/s
vocab_size: 128512, batch_size: 16, distrib: normal_distribution(std=1), deterministic: False, k: 100, duration: 961.15 us, effective bandwidth: 8.56 GB/s
vocab_size: 128512, batch_size: 16, distrib: normal_distribution(std=1), deterministic: False, k: 1000, duration: 696.38 us, effective bandwidth: 11.81 GB/s
vocab_size: 128512, batch_size: 32, distrib: normal_distribution(std=1), deterministic: False, k: 10, duration: 1028.06 us, effective bandwidth: 16.00 GB/s
vocab_size: 128512, batch_size: 32, distrib: normal_distribution(std=1), deterministic: False, k: 100, duration: 839.38 us, effective bandwidth: 19.60 GB/s
vocab_size: 128512, batch_size: 32, distrib: normal_distribution(std=1), deterministic: False, k: 1000, duration: 655.28 us, effective bandwidth: 25.10 GB/s

**radik_sampling_from_probs**:
> vocab_size: 128512, batch_size: 1, distrib: normal_distribution(std=1), deterministic: False, k: 10, duration: 101.28 us, effective bandwidth: 5.08 GB/s
vocab_size: 128512, batch_size: 1, distrib: normal_distribution(std=1), deterministic: False, k: 100, duration: 102.05 us, effective bandwidth: 5.04 GB/s
vocab_size: 128512, batch_size: 1, distrib: normal_distribution(std=1), deterministic: False, k: 1000, duration: 104.32 us, effective bandwidth: 4.93 GB/s
vocab_size: 128512, batch_size: 16, distrib: normal_distribution(std=1), deterministic: False, k: 10, duration: 111.55 us, effective bandwidth: 73.73 GB/s
vocab_size: 128512, batch_size: 16, distrib: normal_distribution(std=1), deterministic: False, k: 100, duration: 121.98 us, effective bandwidth: 67.43 GB/s
vocab_size: 128512, batch_size: 16, distrib: normal_distribution(std=1), deterministic: False, k: 1000, duration: 114.82 us, effective bandwidth: 71.63 GB/s
vocab_size: 128512, batch_size: 32, distrib: normal_distribution(std=1), deterministic: False, k: 10, duration: 120.45 us, effective bandwidth: 136.57 GB/s
vocab_size: 128512, batch_size: 32, distrib: normal_distribution(std=1), deterministic: False, k: 100, duration: 121.34 us, effective bandwidth: 135.56 GB/s
vocab_size: 128512, batch_size: 32, distrib: normal_distribution(std=1), deterministic: False, k: 1000, duration: 138.34 us, effective bandwidth: 118.91 GB/s

[benchmark.log](https://github.com/user-attachments/files/21955916/benchmark.log)

### DIFF
--- a/flashinfer/sampling.py
+++ b/flashinfer/sampling.py
@@ -214,6 +214,37 @@ def get_sampling_module():
         )
         return samples
 
+    @register_custom_op(
+        "flashinfer::radik_sampling_from_probs", mutates_args=("workspace_buffer",)
+    )
+    def radik_sampling_from_probs(
+        workspace_buffer: torch.Tensor,
+        probs: torch.Tensor,
+        indices: Optional[torch.Tensor],
+        maybe_top_k_arr: Optional[torch.Tensor],
+        top_k_val: int,
+        deterministic: bool,
+        selected_probs: Optional[torch.Tensor],
+        generator: Optional[torch.Generator],
+    ) -> torch.Tensor:
+        device = probs.device
+        probs = probs.float()
+        batch_size = indices.size(0) if indices is not None else probs.size(0)
+        maybe_top_k_arr = maybe_top_k_arr.int() if maybe_top_k_arr is not None else None
+        samples = torch.empty(batch_size, dtype=torch.int32, device=device)
+        module.radik_sampling_from_probs.default(
+            workspace_buffer,
+            probs,
+            samples,
+            indices,
+            maybe_top_k_arr,
+            top_k_val,
+            deterministic,
+            selected_probs,
+            generator,
+        )
+        return samples
+
     @register_fake_op("flashinfer::top_k_sampling_from_probs")
     def _fake_top_k_sampling_from_probs(
         probs: torch.Tensor,
@@ -453,6 +484,7 @@ def get_sampling_module():
         top_k_renorm_probs=top_k_renorm_probs,
         top_k_mask_logits=top_k_mask_logits,
         chain_speculative_sampling=chain_speculative_sampling,
+        radik_sampling_from_probs=radik_sampling_from_probs,
     )
 
 
@@ -803,6 +835,66 @@ def top_k_sampling_from_probs(
             raise ValueError("Input probs contains NaN.")
     return get_sampling_module().top_k_sampling_from_probs(
         probs, indices, *_to_tensor_scalar_tuple(top_k), deterministic, generator
+    )
+
+
+def radik_sampling_from_probs(
+    probs: torch.Tensor,
+    top_k: Union[torch.Tensor, int],
+    indices: Optional[torch.Tensor] = None,
+    deterministic: bool = True,
+    generator: Optional[torch.Generator] = None,
+    selected_probs: Optional[torch.Tensor] = None,
+    check_nan: bool = False,
+) -> torch.Tensor:
+    r"""Fused GPU kernel for top-k sampling from probabilities,
+    this operator implements GPU-based rejection sampling without explicit sorting.
+    Check the `blog post <https://flashinfer.ai/2025/03/10/sampling.html>`_ for more details.
+
+    The multiple rounds of rejection sampling are implemented in a single CUDA kernel,
+    which is more efficient than the naive implementation that launches a series of kernels.
+
+    Parameters
+    ----------
+    probs: torch.Tensor
+        Probabilities for sampling. When indices is not provided, shape should be ``(batch_size, num_classes)``
+        and the i-th output will be sampled from the i-th row of probabilities. When indices is provided,
+        shape should be ``(unique_batch_size, num_classes)`` where unique_batch_size is the number of unique
+        probability distributions.
+    top_k: Union[torch.Tensor, int]
+        Either a scalar or a tensor of shape ``(batch_size,)``, representing the threshold for top-k sampling.
+        If a scalar, the same threshold is used for all requests.
+        If a tensor, each request has its own threshold.
+    indices: Optional[torch.Tensor]
+        Optional indices tensor of shape ``(batch_size,)`` that maps each output to a row in probs.
+        For example, if indices[i] = j, then the i-th output will be sampled from probs[j].
+        This allows reusing the same probability distribution for multiple outputs.
+        If indices is not provided, the i-th output will be sampled from the i-th row of probs.
+    deterministic: bool
+        Whether to use deterministic kernel implementation, default is ``True``.
+    generator: Optional[torch.Generator]
+        A random number generator for the operation.
+    selected_probs: Optional[torch.Tensor]
+        Optional tensor of shape ``(batch_size, top_k)`` that stores the top-k selected probabilities.
+    check_nan: bool
+        Whether to check nan in :attr:`probs`, default is ``False``.
+    """
+    if check_nan:
+        if torch.any(torch.isnan(probs)):
+            raise ValueError("Input probs contains NaN.")
+
+    workspace_buffer = _get_cache_buf(
+        "radik_sampling_from_probs_workspace", 64 * 1024 * 1024, probs.device
+    )
+
+    return get_sampling_module().radik_sampling_from_probs(
+        workspace_buffer,
+        probs,
+        indices,
+        *_to_tensor_scalar_tuple(top_k),
+        deterministic,
+        selected_probs,
+        generator,
     )
 
 
--- a/include/flashinfer/sampling.cuh
+++ b/include/flashinfer/sampling.cuh
@@ -28,9 +28,11 @@
 #include <limits>
 #include <numeric>
 #include <tuple>
+#include <vector>
 
 #include "allocator.h"
 #include "math.cuh"
+#include "pytorch_extension_utils.h"
 #include "utils.cuh"
 #include "vec_dtypes.cuh"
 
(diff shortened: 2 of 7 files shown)


include/flashinfer/sampling.cuh
```diff
@@ -2211,6 +2213,704 @@ cudaError_t ChainSpeculativeSampling(DType* draft_probs, IdType* draft_token_ids
...
+  T* top_k_select_result = reinterpret_cast<T*>(binIdPtr + batch_size);
+  IdType* top_k_select_idx =
+      reinterpret_cast<IdType*>(top_k_select_result + batch_size * max_top_k_val);
+  std::vector<int> taskLenHost(batch_size);
+
+  // clear hist and globalCount
+  FLASHINFER_CUDA_CALL(
+      cudaMemsetAsync(histPtr, 0, sizeof(int) * batch_size * ((1 << 12) + 1), stream));
+
+  DISPATCH_COMPUTE_CAP_NUM_THREADS(compute_capacity, BLOCK_THREADS, {
```
yzh119: CI failed because of too many shared memory usage (the public CI was executed on a machine with small shared memory size ~98k/sm), need to tweak `BLOCK_THREADS` to make sure shared memory usage is bounded.

We can create a new macro, the heuristics defined in `DISPATCH_COMPUTE_CAP_NUM_THREADS` is designed specifically for certain kernels and might not work for radix top-k.

JasonJ2021: ok, I'll try to fix it

JasonJ2021: The shared memory issue has been resolved.

A new macro is used to determine `BLOCK_THREADS` for the `selectCandidateEx` kernel, while all other kernel launches use the existing macro. I am not sure whether this is the correct approach and if it is consistent with the existing flashinfer codestyle.

yzh119:
Hi @JasonJ2021 is this PR ready for review?

JasonJ2021:
Yes.

include/flashinfer/sampling.cuh
```diff
@@ -2213,6 +2224,712 @@ cudaError_t ChainSpeculativeSampling(DType* draft_probs, IdType* draft_token_ids
...
+
+template <uint32_t BLOCK_THREADS>
+struct SelectBinTempStorage {
+  union {
+    typename cub::BlockScan<int, BLOCK_THREADS>::TempStorage scan;
+  } block_prim;
+};
+
+template <int BLOCK_THREADS, int HISTLEN>
+__global__ void __launch_bounds__(1024)
```
yzh119:
```suggestion
__global__ void __launch_bounds__(BLOCK_THREADS)
```

include/flashinfer/sampling.cuh
```diff
@@ -2213,6 +2224,712 @@ cudaError_t ChainSpeculativeSampling(DType* draft_probs, IdType* draft_token_ids
...
+        break;
+      }
+      oldK -= count[i];
+    }
+  }
+  return;
+}
+
+template <int BLOCK_THREADS, int LEFT, int RIGHT>
+__global__ void __launch_bounds__(1024)
```
yzh119:
```suggestion
__global__ void __launch_bounds__(BLOCK_THREADS)
```

include/flashinfer/sampling.cuh
```diff
@@ -2213,6 +2224,712 @@ cudaError_t ChainSpeculativeSampling(DType* draft_probs, IdType* draft_token_ids
...
+  const int taskId = blockIdx.y;
+  const int taskLen = taskLenPtr[taskId];
+  const int mask = binIdPtr[taskId];
+  const int row_idx = indices == nullptr ? taskId : indices[taskId];
+  vec_t<T, VEC_SIZE> dataIn_vec;
+
+  __shared__ int blockCount[1];
+  __shared__ T blockCache[BLOCK_THREADS * VEC_SIZE];
+
+#pragma unroll 2
```
yzh119: why decorating "if" with loop unroll?

include/flashinfer/sampling.cuh
```diff
@@ -2213,6 +2224,712 @@ cudaError_t ChainSpeculativeSampling(DType* draft_probs, IdType* draft_token_ids
...
+  for (uint32_t j = 0; j < VEC_SIZE; j++) {
+    if (tx * VEC_SIZE + j < count) {
+      dataOut[taskId * stride + blockCount[0] + tx * VEC_SIZE + j] = blockCache[tx * VEC_SIZE + j];
+    }
+  }
+  return;
+}
+
+template <int BLOCK_THREADS, int VEC_SIZE, int CACHE_SIZE, typename IdxType, typename ValType>
+__global__ void __launch_bounds__(1024)
```
yzh119:
```suggestion
__global__ void __launch_bounds__(BLOCK_THREADS)
```

include/flashinfer/sampling.cuh
```diff
@@ -2213,6 +2224,712 @@ cudaError_t ChainSpeculativeSampling(DType* draft_probs, IdType* draft_token_ids
...
+          valBlockCache[pos] = dataIn_vec[j];
+          idxBlockCache[pos] = (i * BLOCK_THREADS + tx) * VEC_SIZE + j;
+        }
+      }
+    }
+  } else {
+    // N > K, filter by k-th element
+    const CompT kThElem = *(kThElePtr + taskId * stride);
+    // if (tx == 0) {
+    //   printf("task%d kThElem: %.14f\n", taskId, kThElem);
```
yzh119: Remove debug code if no longer required

include/flashinfer/sampling.cuh
```diff
@@ -2213,6 +2224,712 @@ cudaError_t ChainSpeculativeSampling(DType* draft_probs, IdType* draft_token_ids
...
+      top_k_arr == nullptr ? top_k_val : *std::max_element(tmpK.begin(), tmpK.end());
+
+  size_t workSpaceSize = 0;
+  getRadixSelectWorkSpaceSize<T, IdType>(max_top_k_val, d, batch_size, &workSpaceSize);
+  //   printf("workSpaceSize: %zuGB\n", workSpaceSize / 1024 / 1024 / 1024);
+  //   printf("batch_size: %d\n", batch_size);
+  void* workSpace = 0;
+  bool workspace_allocated = false;
+  if (workSpaceSize > workspace_buffer_size_in_bytes) {
+    FLASHINFER_CUDA_CALL(cudaMalloc(&workSpace, workSpaceSize));
```
yzh119: I would encourage pre-allocating these buffer in pytorch and pass the allocated buffer to C++ instead (and reuse them).
The cuda malloc time could be non-neglibile.

include/flashinfer/sampling.cuh
```diff
@@ -2213,6 +2224,712 @@ cudaError_t ChainSpeculativeSampling(DType* draft_probs, IdType* draft_token_ids
...
+
+  CompT* valBuffer[2]{static_cast<CompT*>(workSpace),
+                      static_cast<CompT*>(workSpace) + batch_size * d};
+  int* histPtr = reinterpret_cast<int*>(valBuffer[1] + batch_size * d);
+  int* globalCountPtr = histPtr + (1 << 12) * batch_size;
+
+  int* taskLenPtr[2]{globalCountPtr + batch_size, globalCountPtr + 2 * batch_size};
+  std::vector<int> tmpTaskLen(2 * batch_size, d);
+
+  FLASHINFER_CUDA_CALL(cudaMemcpyAsync(taskLenPtr[0], tmpTaskLen.data(),
```
yzh119: Haven't profiled the performance yet, but could be use a standalone cuda kernel (and enable PDL) to initialize these global buffers (e.g. taskLenPtr, kPtr and histPtr) instead of calling multiple MemcpyAsync like this?

include/flashinfer/sampling.cuh
```diff
@@ -2213,6 +2224,712 @@ cudaError_t ChainSpeculativeSampling(DType* draft_probs, IdType* draft_token_ids
...
+                                      &binIdPtr,        &taskLenPtr[flag],    &d};
+
+      FLASHINFER_CUDA_CALL(cudaLaunchKernel((void*)selectcan_iter3_kernel, selectcan_iter3_nblks,
+                                            selectcan_iter3_nthrs, selectcan_iter3_args, 0,
+                                            stream));
+
+      flag ^= 1;
+    }
+    // clear globalCount
+    FLASHINFER_CUDA_CALL(cudaMemsetAsync(globalCountPtr, 0, sizeof(int) * batch_size, stream));
```
yzh119, line 2875: ditto

include/flashinfer/sampling.cuh
```diff
@@ -2213,6 +2224,712 @@ cudaError_t ChainSpeculativeSampling(DType* draft_probs, IdType* draft_token_ids
...
+                                            stream));
+
+      // update taskLen
+      FLASHINFER_CUDA_CALL(cudaMemcpyAsync(taskLenHost.data(), taskLenPtr[flag ^ 1],
+                                           sizeof(int) * batch_size, cudaMemcpyDefault, stream));
+      flag ^= 1;
+    }
+
+    // === Iter 3 ===
+    FLASHINFER_CUDA_CALL(cudaStreamSynchronize(stream));
```
yzh119, line 2827: Is it possible to get rid of cudaStreamSynchronization like this?
Seems we rely on host-side maxTaskLen as kernel configuration of `countbin_iter3_kernel`, but is it possible to rewrite `countbin_iter3_kernel` as a persistent kernel (fix grid configuration), but inside the kernel we depend on device-side `maxTaskLen` value (so that we can reduce synchronization overhead here).

include/flashinfer/sampling.cuh
```diff
@@ -2213,6 +2224,712 @@ cudaError_t ChainSpeculativeSampling(DType* draft_probs, IdType* draft_token_ids
...
+                               &philox_seed,         &philox_offset};
+
+        FLASHINFER_CUDA_CALL(cudaLaunchKernel((void*)sample_kernel, sample_nblks, sample_nthrs,
+                                              sample_args, sample_smem_size, stream));
+      });
+    });
+  });
+
+  // check global histgram
+  FLASHINFER_CUDA_CALL(cudaStreamSynchronize(stream));
```
yzh119, line 2933: ditto

flashinfer/sampling.py
```diff
@@ -806,6 +838,66 @@ def top_k_sampling_from_probs(
     )
 
 
+def radik_sampling_from_probs(
```
yzh119, line 874: I would also encourage dispatching topk sampling to this function for small K's.

yzh119:
Hi @JasonJ2021 do you have time to address the comments?

JasonJ2021:
> Hi @JasonJ2021 do you have time to address the comments?

Yes, I will start to address these issues in the next two to three days.

include/flashinfer/sampling.cuh
JasonJ2021:
```python
    workspace_buffer = _get_cache_buf(
        "radik_sampling_from_probs_workspace", 64 * 1024 * 1024, probs.device
    )
```

I have pre-allocated a 64MB buffer in `radik_sampling_from_probs` to handle small batch & top-k sampling; this buffer is sufficient for a request with maximum parameters of batch_size=64, top_k=10, and vocab_size=128512. I'm not sure if I should choose a larger buffer size to handle requests with a larger batch_size.

include/flashinfer/sampling.cuh
JasonJ2021: Sorry, i'm not experienced in cuda programming. 
My understanding is to first use a standalone kernel (e.g., `initialize_buffer()`) to initialize these global buffers, and then leverage PDL to allow this initialization kernel to execute concurrently with a subsequent compute kernel (e.g., countBinExKernel) in a single stream. Is my understanding correct?

JasonJ2021:
```python
def top_k_sampling_from_probs(
    probs: torch.Tensor,
    top_k: Union[torch.Tensor, int],
    indices: Optional[torch.Tensor] = None,
    deterministic: bool = True,
    generator: Optional[torch.Generator] = None,
    check_nan: bool = False,
) -> torch.Tensor:
    ...
    # NOTE: radik_sampling_from_probs is non-deterministic
    use_radik_impl = not deterministic and (
        (isinstance(top_k, int) and top_k <= 100)
        or (isinstance(top_k, torch.Tensor) and top_k.max() <= 100)
    )

    if use_radik_impl:
        return radik_sampling_from_probs(
            probs,
            top_k,
            indices,
            deterministic,
            generator,
            selected_probs=None,
            check_nan=check_nan,
        )

    return get_sampling_module().top_k_sampling_from_probs(
        probs, indices, *_to_tensor_scalar_tuple(top_k), deterministic, generator
    )
```

i found that the Radik implementation for `top_k_sampling_from_probs` should only be dispatched when deterministic output is not required.

The root cause is that Radik's parallel algorithm for selecting the top-k elements is non-deterministic. It uses multiple threadblocks with atomic operations to populate an top-k-select array, but the final order of elements in that array is not guaranteed between runs.

This unstable ordering directly affects the subsequent Cumulative Distribution Function (CDF) sampling step, leading to different results from the same input probabilities. 

For example, for the same input, the Radik top-k selection might produce [0.4, 0.3, 0.2, 0.1] in one run and [0.1, 0.2, 0.3, 0.4] in another. Given the same random number for sampling (e.g., 0.35), the CDF-based logic would select index 0 (prob=0.4) in the first case, but index 2 (prob=0.3) in the second, leading to a different result.

ECMGit:
Hi experts, When will the commit be merged? Looks like we are facing a under performing issue due to the sampling strategy, if this commit can be merged, we can get rid of [this change](https://github.com/vllm-project/vllm/pull/28099) from replacing `flashinfer.sampling.top_k_top_p_sampling_from_logits` to `flashinfer.sampling.top_k_top_p_sampling_from_probs`.

yzh119:
Hi @ECMGit @JasonJ2021 can you try #2119 ?

JasonJ2021:
> Hi @ECMGit @JasonJ2021 can you try #2119 ?

Thanks, Zihao! I learned a lot from your kernel. Sorry I didn't have the time to merge my code into FlashInfer yet, but I'll definitely find a way to contribute to the project later!

PR #1576, opened 2025-08-26
re-enable torch.compile

guilhermeleobas:
List of changes:
* Use a faster version of `register_custom_op` / `register_fake_op`
* Fix `top_p_sampling_from_probs` fake impl
* (hack) Replace `torch._C.Generator` with `bool` as the former cannot be parsed in a schema
* Raise `ValueError(...)` if generator is specified when torch.compiling

Test plan:
* Create test file `test_torch_compile.py` with the content from `test_sampling.py`
* Test each exported sampling method with `torch.library.opcheck(...)`

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/flashinfer/sampling.py
+++ b/flashinfer/sampling.py
@@ -477,7 +477,7 @@ def _get_sampling_module():
     )
 
 
-@torch._dynamo.disable(reason="Dynamo cannot trace 'get_sampling_module'. Set fullgraph=False")
+@torch._dynamo.disable
 def get_sampling_module():
     return _get_sampling_module()
 


guilhermeleobas:
@yzh119, I'd just like to get your approval before I continue working on this. Is this something the flashinfer community would be interested?

Edit: I'll restrict this PR to just sampling.py as the approach can be easily replicated to the other files. Also, the `generator` argument in some of the functions is problematic as `torch._C.Generator` is not valid in a schema. Is this argument used? If not, can I remove it?

yzh119:
@abcdabcd987 can you help with reviewing this?

guilhermeleobas:
Hi @abcdabcd987, could you take a look at this PR once you have a chance? I'm happy to answer any question you have regarding the changes I made.

PR #1588, opened 2025-08-27
refactor: use allocator class for workspace buffer allocation

yyihuang:
## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/csrc/trtllm_fmha_kernel_launcher.cu
+++ b/csrc/trtllm_fmha_kernel_launcher.cu
@@ -330,7 +330,7 @@ void trtllm_ragged_attention_launcher(
     double o_sf_scale, int64_t batch_size, int64_t window_left, int64_t sm_count, bool enable_pdl,
     bool is_causal, int64_t k_stride_keys_values, int64_t k_stride_heads, int64_t k_stride_batch,
     int64_t v_stride_keys_values, int64_t v_stride_heads, int64_t v_stride_batch,
-    cudaStream_t stream) {
+    int64_t workspace_size, cudaStream_t stream) {
   if (num_qo_heads % num_kv_heads != 0) {
     std::ostringstream err_msg;
     err_msg << "num_qo_heads must be a multiple of num_kv_heads, got num_kv_heads: " << num_kv_heads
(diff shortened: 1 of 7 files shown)


csrc/trtllm_fmha_kernel_launcher.cu
```diff
@@ -153,9 +156,10 @@ void trtllm_paged_attention_launcher(
     size_t num_semaphores =
         round_up(max_batch_size * max_num_qo_heads, 8);  // max 8MB, should align to 16 bytes
     // semaphores be at the first 8MB of workspace buffer: counter | scratch
-    runner_params.multiCtasKvScratchPtr = reinterpret_cast<void*>(
-        static_cast<char*>(workspace_buffer) + num_semaphores * sizeof(uint32_t));
-    runner_params.multiCtasKvCounterPtr = reinterpret_cast<int32_t*>(workspace_buffer);
+    runner_params.multiCtasKvCounterPtr = float_allocator.aligned_alloc<int32_t>(
+        num_semaphores * sizeof(uint32_t), 16, "trtllm_gen_counter_workspace");
+    runner_params.multiCtasKvScratchPtr =
+        float_allocator.aligned_alloc<void>(0, 16, "trtllm_gen_scratch_workspace");
```
yzh119, line 164: better to annotate the reason why we specify 0 here.

yyihuang: updated.

flashinfer/decode.py
```diff
@@ -2229,6 +2246,7 @@ def trtllm_batch_decode_with_kv_cache_mla(
     bmm2_scale_tensor: Optional[torch.Tensor] = None,
     sinks: Optional[List[torch.Tensor]] = None,
     enable_pdl: bool = None,
+    workspace_size: int = DEFAULT_WORKSPACE_SIZE,
```
yzh119: This value could be directly inferred from `workspace_buffer` tensor so I don't think it's necessary to change interface at python side.

yyihuang: fixed

flashinfer/decode.py
```diff
@@ -887,6 +892,8 @@ def plan(
 
         The :meth:`plan` method cannot be used in Cuda Graph or in ``torch.compile``.
         """
+        self._workspace_size = workspace_size
```
yzh119: I don't feel like this is required as we can infered it from `self.float_workspace_buffer`

yyihuang: fixed

yzh119 (approved):
cc @weireweire for the interface change, now workspace_size is passed to C++ side as an argument

PR #1578, opened 2025-08-26
feat: Support for inferring out_dtype from out.dtype for TRTLLM attention kernel

elvischenv:
## 📌 Description

Instead of directly use `query.dtype` as `out_dtype`, we could infer the `out_dtype` from `out.dtype` if `out` is passed into the API.

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [ ] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [ ] I have installed the hooks with `pre-commit install`.
- [ ] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [ ] Tests have been added or updated as needed.
- [ ] All tests are passing (`unittest`, etc.).

tests/test_trtllm_gen_attention.py::test_trtllm_batch_prefill:
`==== 972 passed, 4 warnings in 47.94s ====`
tests/test_trtllm_gen_attention.py::test_trtllm_batch_decode:
`==== 2592 passed, 15 warnings in 88.27s (0:01:28) ====`

PR #1590, opened 2025-08-27
fix: Improve TRTLLM attention kernel out_dtype unit test

elvischenv:
## 📌 Description

- Added follow-up unit tests for #1578: passed or not passed `out_dtype` to the attn api.
- fixed the issue when only passing `out` for NVFP4 cases
`tests/test_trtllm_gen_attention.py::test_trtllm_batch_prefill`
`===== 972 passed in 30.55s =====`
`tests/test_trtllm_gen_attention.py::test_trtllm_batch_decode`
`===== 2592 passed in 74.97s (0:01:14) ====`

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/flashinfer/decode.py
+++ b/flashinfer/decode.py
@@ -2092,7 +2092,7 @@ def trtllm_batch_decode_with_kv_cache(
             o_sf_start_index = out.scale_start_index
             out = out.data
             # out_dtype may be None
-            out_dtype = "nvfp4"
+            out_dtype = out_dtype or "nvfp4"
         elif out is None:
             fp4_out_scale_shape = (
                 round_up(query.shape[0], 128),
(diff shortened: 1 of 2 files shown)


flashinfer/decode.py
```diff
@@ -2091,6 +2091,8 @@ def trtllm_batch_decode_with_kv_cache(
             out_scale_factor = out.scale
             o_sf_start_index = out.scale_start_index
             out = out.data
+            # out_dtype may be None
+            out_dtype = "nvfp4"
```
weireweire: let's also add check:
```
out_dtype = out_dtype or "nvfp4"
assert out_dtype == "nvfp4"
```

elvischenv: Inside this `if out_dtype == "nvfp4" or (out_dtype is None and isinstance(out, FP4Tensor)):`, `out_dtype` could only be `nvfp4` or `None`. The assert seems unnecessary. Do we still want to add it?

weireweire (approved):

flashinfer/decode.py
weireweire: I'd like to add for safety. code may change, and the overwrite may be issue in the future.

elvischenv: Added and thanks for the review.

PR #1608, opened 2025-08-30
Feature/sm103 bringup

aleozlx:
## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [x] Tests have been added or updated as needed.
- [ ] All tests are passing (`unittest`, etc.).

PR #1610, opened 2025-08-30
feat: cutlass fp8 gemm bringup for SM120 & SM121

yongwww:
## 📌 Description

It depends on https://github.com/flashinfer-ai/flashinfer/pull/1608, mainly the cutlass fp8 gemm support for sm120/121, will rebase after #1608 lands.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/csrc/gemm_groupwise_sm120.cu
+++ b/csrc/gemm_groupwise_sm120.cu
@@ -31,43 +31,45 @@ using namespace flashinfer;
                                    SCALE_GRANULARITY_M, SCALE_GRANULARITY_N, SCALE_GRANULARITY_K, \
                                    ...)                                                           \
   [&]() -> bool {                                                                                 \
-    constexpr int SCALE_GRANULARITY_M = 1;  /* Always 1 for SM120 */                             \
-    constexpr int SCALE_GRANULARITY_K = 128; /* Always 128 for SM120 per CUTLASS requirement */  \
-    if (scale_granularity_m != 1) {                                                              \
-      TORCH_CHECK(false, "SM120 only supports scale_granularity_m=1");                           \
+    constexpr int SCALE_GRANULARITY_M = 1;   /* Always 1 for SM120 */                             \
+    constexpr int SCALE_GRANULARITY_K = 128; /* Always 128 for SM120 per CUTLASS requirement */   \
+    if (scale_granularity_m != 1) {                                                               \
+      TORCH_CHECK(false, "SM120 only supports scale_granularity_m=1");                            \
       return false;                                                                               \
     }                                                                                             \
-    if (scale_granularity_k != 128) {                                                            \
-      TORCH_CHECK(false, "SM120 only supports scale_granularity_k=128");                         \
+    if (scale_granularity_k != 128) {                                                             \
+      TORCH_CHECK(false, "SM120 only supports scale_granularity_k=128");                          \
       return false;                                                                               \
     }                                                                                             \
-    /* Dispatch based on n granularity only */                                                   \
-    if (scale_granularity_n == 128) {                                                            \
-      constexpr int SCALE_GRANULARITY_N = 128;                                                   \
+    /* Dispatch based on n granularity only */                                                    \
+    if (scale_granularity_n == 128) {                                                             \
+      constexpr int SCALE_GRANULARITY_N = 128;                                                    \
       return __VA_ARGS__();                                                                       \
-    } else if (scale_granularity_n == 64) {                                                      \
-      constexpr int SCALE_GRANULARITY_N = 64;                                                    \
+    } else if (scale_granularity_n == 64) {                                                       \
+      constexpr int SCALE_GRANULARITY_N = 64;                                                     \
       return __VA_ARGS__();                                                                       \
-    } else if (scale_granularity_n == 32) {                                                      \
-      constexpr int SCALE_GRANULARITY_N = 32;                                                    \
+    } else if (scale_granularity_n == 32) {                                                       \
+      constexpr int SCALE_GRANULARITY_N = 32;                                                     \
       return __VA_ARGS__();                                                                       \
-    } else if (scale_granularity_n == 16) {                                                      \
-      constexpr int SCALE_GRANULARITY_N = 16;                                                    \
+    } else if (scale_granularity_n == 16) {                                                       \
+      constexpr int SCALE_GRANULARITY_N = 16;                                                     \
       return __VA_ARGS__();                                                                       \
     }                                                                                             \
     TORCH_CHECK(false, "SM120: Unsupported scale granularity combination (", scale_granularity_m, \
-                ",", scale_granularity_n, ",", scale_granularity_k, ")");                        \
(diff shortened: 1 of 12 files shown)


flashinfer/gemm.py
```diff
@@ -349,6 +358,216 @@ def get_gemm_sm100_module():
...
+                profile: OptimizationProfile,
+            ) -> List[int]:
+                # For now, return a single default tactic
+                return [-1]
+            
+            def forward(
+                self,
+                inputs: List[torch.Tensor],
+                tactic: int = -1,
+                do_preparation: bool = False,
```
nvmbreughe, line 526: seems like this parameter is unused?

yongwww: Good catch. They’re part of TunableRunner, keeping them for consistency with the others.

aleozlx:
looks good no further comments from me

csrc/gemm_groupwise_sm120.cu
```diff
@@ -0,0 +1,158 @@
...
+  }()
+
+#define DISPATCH_MMA_SM(mma_sm, MMA_SM, ...)                                                  \
+  [&]() -> bool {                                                                             \
+    if (mma_sm == 1) {                                                                        \
+      constexpr int MMA_SM = 1;                                                               \
+      return __VA_ARGS__();                                                                   \
+    }                                                                                         \
+    TORCH_CHECK(false,                                                                        \
+                "SM120 only supports MmaSM=1 (MmaSM=2 would violate Cooperative kernel tile " \
```
yzh119: I don't understand this error message here, why `MmaSM=2` will make `M < 128`?

csrc/gemm_groupwise_sm120.cu
```diff
@@ -0,0 +1,158 @@
...
+  }()
+
+#define DISPATCH_SCALE_GRANULARITY(scale_granularity_m, scale_granularity_n, scale_granularity_k, \
+                                   SCALE_GRANULARITY_M, SCALE_GRANULARITY_N, SCALE_GRANULARITY_K, \
+                                   ...)                                                           \
+  [&]() -> bool {                                                                                 \
+    constexpr int SCALE_GRANULARITY_M = 1;   /* Always 1 for SM120 */                             \
+    constexpr int SCALE_GRANULARITY_K = 128; /* Always 128 for SM120 per CUTLASS requirement */   \
+    if (scale_granularity_m != 1) {                                                               \
+      TORCH_CHECK(false, "SM120 only supports scale_granularity_m=1");                            \
```
yzh119: Didn't see this constraint in https://github.com/NVIDIA/cutlass/blob/b2dd65dc864e09688245b316ac46c4a6cd07e15c/include/cutlass/gemm/collective/sm120_mma_tma_blockwise_scaling.hpp, what's the error message if you set it to 128?

yongwww: will run into `static assertion failed with "Scale Granularity M must evenly divide the tile shape M."`

yzh119: This error should only happend when `ScaleGranularityM == 128` and `TileShapeM = 64`:

https://github.com/NVIDIA/cutlass/blob/b2dd65dc864e09688245b316ac46c4a6cd07e15c/include/cutlass/gemm/collective/sm120_mma_tma_blockwise_scaling.hpp#L134

which is the case when `MmaSM == 1` here: https://github.com/flashinfer-ai/flashinfer/pull/1610/files#diff-0977093a8d2429e66dab4cc40f31563717098cb5aca4354a814e4208f58f068bR78, which you disabled in https://github.com/flashinfer-ai/flashinfer/pull/1610/files#diff-68929275a79ec730031c1d5bec894f35ba6e932a08841fdae63087a6937c0f4fR70

yzh119:
There might be some misunderstanding of MmaSM here: we use it in sm100 gemm because sm100 supports tcgen05 and 2-cta mode (where 2 ctas cooperatively perform a mma computation).

However, sm120 do not have tcgen05 and 2-cta mma, so MmaSM doesn't make sense here, it should always be 1 in sm120.

The condition used separate to the two cases in https://github.com/flashinfer-ai/flashinfer/pull/1610/files#diff-0977093a8d2429e66dab4cc40f31563717098cb5aca4354a814e4208f58f068bR78 should not be `MmaSM == 1`, but Cooperative/PingPong schedule: https://github.com/NVIDIA/cutlass/blob/b2dd65dc864e09688245b316ac46c4a6cd07e15c/examples/87_blackwell_geforce_gemm_blockwise/87b_blackwell_geforce_fp8_bf16_gemm_groupwise.cu#L120-L123.

Please consider the following changes:
1. remove the confusing `MmaSM` argument, there is no concept of 2SM in sm120.
2. if we want to support both PingPong and Cooperative GEMM, please refer to https://github.com/NVIDIA/cutlass/blob/b2dd65dc864e09688245b316ac46c4a6cd07e15c/examples/87_blackwell_geforce_gemm_blockwise/87b_blackwell_geforce_fp8_bf16_gemm_groupwise.cu#L166-L171

csrc/gemm_groupwise_sm120.cu
yongwww: right. I used a standalone test (not in this pr) to trigger that error message. will go with the https://github.com/flashinfer-ai/flashinfer/pull/1610#issuecomment-3251193531

yongwww:
Thanks, @yzh119, @nvmbreughe , @aleozlx  for the helpful and insightful comments! I’ve incorporated them. Please take a look. For the PingPong gemm, I left it as a todo for now; the current default in the cutlass examples is cooperative gemm.

csrc/gemm_groupwise_sm120.cu
```diff
@@ -0,0 +1,149 @@
...
+
+#define DISPATCH_SCALE_GRANULARITY(scale_granularity_m, scale_granularity_n, scale_granularity_k,  \
+                                   SCALE_GRANULARITY_M, SCALE_GRANULARITY_N, SCALE_GRANULARITY_K,  \
+                                   ...)                                                            \
+  [&]() -> bool {                                                                                  \
+    constexpr int SCALE_GRANULARITY_M = 1;   /* Always 1 for SM120 */                              \
+    constexpr int SCALE_GRANULARITY_K = 128; /* equal tile K dimension*/                           \
+    if (scale_granularity_m != 1) {                                                                \
+      TORCH_CHECK(false,                                                                           \
+                  "SM120 only supports scale_granularity_m=1 to ensure compatibility with all "    \
```
yzh119: Is this still the case after your changes? If not, let's add 128 back.

yongwww: Good catch! A divisor of 128 should be valid. I added scale_granularity_m=1  and scale_granularity_m=128, the change was:  https://github.com/flashinfer-ai/flashinfer/pull/1610/files#diff-68929275a79ec730031c1d5bec894f35ba6e932a08841fdae63087a6937c0f4fR44-R52

PR #1625, opened 2025-09-02
bugfix: fix flashinfer_benchmark.py IMA when running a test list

bkryu:
## 📌 Description

The current `flashinfer_benchmark.py` script can trigger an IMA when a testlist is provided to batch-benchmark multiple test cases.

Current PR:
* Fixes the bug by clearing torch's cache and synchronizing the device at the beginning of each test.
     - IMA occurs between test cases and can only be reproduced when running a testlist.
     - Fix is to `torch.cuda.empty_cache()` and `torch.cuda.synchronize()` at the beginning of each test.
* Miscellaneous improvements to flashinfer_benchmark.py:
     - Attention benchmarks:
         - Reduces unnecessary reference calculations.
         - Prints statistics upon encountering reference check failures
     - GEMM benchmarks:
         - Allows testing `trtllm`backend in `testGemmFp8NtGroupwise` API

*No changes to the library code*

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/benchmarks/routines/attention.py
+++ b/benchmarks/routines/attention.py
@@ -501,8 +501,6 @@ def testBatchDecodeWithPagedKVCacheWrapper(args):
     has_reference_output = False
     # Iterate over each backend:
     for cur_backend in backends:
-        torch.cuda.empty_cache()
-        torch.cuda.synchronize()
         if run_refcheck:
             outputs[cur_backend] = run_backend_wrapper(cur_backend).detach().clone()
             if cur_backend == "fa2":
@@ -963,8 +961,6 @@ def testBatchPrefillWithPagedKVCacheWrapper(args):
     has_reference_output = False
     # Iterate over each backend:
     for cur_backend in backends:
-        torch.cuda.empty_cache()
-        torch.cuda.synchronize()
         if run_refcheck:
             outputs[cur_backend] = run_backend_wrapper(cur_backend).detach().clone()
             if cur_backend == "fa2":
@@ -1362,8 +1358,6 @@ def testBatchPrefillWithRaggedKVCacheWrapper(args):
     has_reference_output = False
     # Iterate over each backend:
     for cur_backend in backends:
-        torch.cuda.empty_cache()
-        torch.cuda.synchronize()
         if run_refcheck:
             outputs[cur_backend] = run_backend_wrapper(cur_backend).detach().clone()
             if cur_backend == "fa2":
@@ -1713,8 +1707,6 @@ def testBatchMLAPagedAttentionWrapper(args):
     has_reference_output = False
     # Iterate over each backend:
     for cur_backend in backends:
-        torch.cuda.empty_cache()
-        torch.cuda.synchronize()
         if run_refcheck:
             outputs[cur_backend] = run_backend_wrapper(cur_backend).detach().clone()
             if cur_backend == "fa2":


yongwww:
seems the overall duration of this unittest suite increased significantly (from 21 minutes to >2 hours) :https://ci.tlcpack.ai/blue/organizations/jenkins/flashinfer-ci/detail/PR-1625/1/pipeline/23. Please take a look.

benchmarks/routines/attention.py
```diff
@@ -498,19 +499,16 @@ def run_backend_wrapper(backend):
...
-        reference_output = (
-            backend_wrappers["fa2"]
-            .run(q, kv_cache, k_scale=k_scale, v_scale=v_scale)
-            .detach()
-        )
-        has_reference_output = True
-
     # Iterate over each backend:
     for cur_backend in backends:
+        torch.cuda.empty_cache()
```
yongwww: need to double-check if this introduced a regression. Curious about the reason why we need to do this

bkryu:
> seems the overall duration of this unittest suite increased significantly (from 21 minutes to >2 hours) :https://ci.tlcpack.ai/blue/organizations/jenkins/flashinfer-ci/detail/PR-1625/1/pipeline/23. Please take a look.

Interesting because none of the CI scripts should touch the files affected by this PR. The pipeline that took 2 hours ran  [task_jit_run_tests_part1.sh](https://github.com/flashinfer-ai/flashinfer/blob/main/scripts/task_jit_run_tests_part1.sh). The [output raw log](https://ci.tlcpack.ai/blue/rest/organizations/jenkins/pipelines/flashinfer-ci/branches/PR-1625/runs/1/nodes/23/steps/317/log/?start=0) shows that the pytest times add up to less than 20 minutes so I am fairly confident that the longer duration is due to some external factors.

The same pipeline after the latest commit finished in 21 minutes total

benchmarks/routines/attention.py
bkryu: As per my other comment, highly unlikely that it introduced a regression. 

Separately, thanks for pointing out that this empty_cache() is unnecessary. I added it when I was looking for the source of IMAs and forgot to remove it. I have removed it in my latest commit

benchmarks/routines/attention.py
```diff
@@ -485,7 +486,7 @@ def run_backend_wrapper(backend):
             )
         elif backend == "trtllm-gen-native":
             return flashinfer.decode.trtllm_batch_decode_with_kv_cache(
-                query=q,
+                query=q.contiguous(),
```
yzh119, line 489: I think contiguous is already applied in https://github.com/flashinfer-ai/flashinfer/blob/75df649d2ec4bac44955757ce72fb5b192374b17/flashinfer/decode.py#L1933

benchmarks/routines/flashinfer_benchmark_utils.py
```diff
@@ -117,6 +117,9 @@ def print_perf_metrics(backend, median_time, std_time, tflops, tb_per_sec):
 
 
 def get_device(args):
+    # Synchronize to ensure that the device is ready after previous tests
```
yzh119, line 120: Thanks for the fix but I'm confused why it works here, can you provide more insights?

yzh119:
> The [output raw log](https://ci.tlcpack.ai/blue/rest/organizations/jenkins/pipelines/flashinfer-ci/branches/PR-1625/runs/1/nodes/23/steps/317/log/?start=0) shows that the pytest times add up to less than 20 minutes so I am fairly confident that the longer duration is due to some external factors.

It should be due to the waiting time to get next available runner.

benchmarks/routines/flashinfer_benchmark_utils.py
bkryu: Sure let me try to bottom out on this. What I observe is:

1. When I run a testlist, I hit an error looking like 
`[ERROR] Error: CUDA error: an illegal memory access was encountered
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1
Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.`
2. Error does not reproduce when:
      - I use `compute-sanitizer`.
      - I set `CUDA_LAUNCH_BLOCKING=1`
      - I test the individual test cases one by one in the test list.

The tester synchronizes after kernel launches to collect measured times, but not after (or before) each test case; and this PR empirically works. My current guess is some unanticipated behavior from PyTorch's caching allocator and I'm looking into this

Issue #1638, opened 2025-09-05
Possible FLOPS mis-calculation in bench_trtllm_gen_mla.py


RayWang96:
When the data are generated, the sequence lengths actually vary from 1 up to seq_len.

[https://github.com/flashinfer-ai/flashinfer/blob/main/benchmarks/bench_trtllm_gen_mla.py#L31](url)

However, the TFLOPS computation assumes every sequence has the [full length seq_len](https://github.com/flashinfer-ai/flashinfer/blob/main/benchmarks/bench_trtllm_gen_mla.py#L109-L116), which inflates the reported TFLOPS— in some configurations it even exceeds 2000 TFLOPS.

yzh119:
yes the calculation of tflops here is wrong, would you mind creating a PR to fix this?

cc @yyihuang

RayWang96:
> yes the calculation of tflops here is wrong, would you mind creating a PR to fix this?
> 
> cc [@yyihuang](https://github.com/yyihuang)

Sure, created one.

PR #1640, opened 2025-09-05
bugfix: Fix FLOPS calculation for bench_trtllm_gen_mla.py

RayWang96:
Fix FLOPS calculation for bench_trtllm_gen_mla.py

Details in https://github.com/flashinfer-ai/flashinfer/issues/1638


benchmarks/bench_trtllm_gen_mla.py
```diff
@@ -111,7 +112,7 @@ def bench_trtllm_mla(batch_size, q_len_per_request, seq_len, page_size, dtype):
         * batch_size
```
yzh119: Or removing the `batch_size` here and multiply `sum(seq_lens)` instead?

RayWang96: You're right. It is simpler.

yzh119 (approved):
LGTM, thanks for the fix!

PR #1641, opened 2025-09-05
refactor: using tvm-ffi for multi-platform bindings

yzh119:
## 📌 Description

This PR refactors the codebase to use tvm-ffi for python bindings. The goal is as following:
1. Supporting multiple backends
2. PyTorch version agnostic aot wheels
3. Accelerate compilation time and reduce binary bloat by minimizing dependencies

Co-authored-by: Yaxing Cai<yaxingc@nvidia.com>
Co-authored-by: Kathryn Chen<kathrync@nvidia.com>

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [x] Tests have been added or updated as needed.
- [ ] All tests are passing (`unittest`, etc.).

## Reviewer Notes

cc @tqchen

PR #1642, opened 2025-09-05
Support PaddlePaddle with compatible API

SigureMo:
We are PaddlePaddle contributors working on a PyTorch compatibility layer aimed at making it significantly easier for PyTorch ecosystem libraries to run on Paddle. See context: https://github.com/flashinfer-ai/flashinfer/issues/1563

## Summary

- This PR introduces a minimal, opt-in compatibility path so third-party projects such as flashinfer can be used with Paddle with very small changes.
- The approach is intentionally minimal and opt-in to avoid breaking upstream behavior for existing PyTorch users.

## Design

- C++ / CUDA layer: provide an adapter that is fully compatible with the PyTorch C API surface (ATen / c10 / torch)[^1]. This allows third-party libraries that call into PyTorch C++/CUDA APIs to instead invoke Paddle's C++/CUDA implementation via the adapter.
- Python layer: reorganize a small compatibility layer so that Paddle's Python API matches PyTorch's API shape as closely as possible (we avoid reproducing PyTorch-specific internals like TorchVersion). The goal is that Python code can do `import paddle as torch` and run with minimal or no source changes.
- Import proxy: provide `paddle.compat.enable_torch_proxy()`[^2] which makes `import torch` actually load `paddle`. This removes the need for `import paddle as torch` in most cases and keeps changes non-invasive.

## Usage (example)

- Install (build with compatibility enabled)

    ```bash
    PADDLE_COMPATIBLE_API=1 pip install -v --no-build-isolation .
    ```

- Runtime example

    ```python
    # example.py
    import paddle

    paddle.compat.enable_torch_proxy()  # enable proxy before import torch

    import flashinfer

    # use ops in flashinfer ...
    ```

    ```bash
    PADDLE_COMPATIBLE_API=1 python example.py
    ```

## Why this is opt-in

- We added a simple check for the environment variable `PADDLE_COMPATIBLE_API`. When set, the compatibility hooks and small source adjustments are enabled. This keeps the default behavior unchanged for regular PyTorch or Paddle users.

## Small changes requested in flashinfer

- JIT-related logic: some JIT code in flashinfer assumes PyTorch's directory layout. We do not aim to mirror directory structure 1:1. We request a small refactor to decouple the logic from the exact torch package file layout (make paths configurable or resolve modules by import names).
- `setup.py` / AOT build: during AOT compilation `setup.py` currently does `import torch`. For compatibility builds we need the build to perform `paddle.compat.enable_torch_proxy()` early (before `import torch`), or otherwise provide a small hook so the build imports load `paddle` instead.
- To make these changes backward-compatible, our patch adds an environment-variable driven path (`PADDLE_COMPATIBLE_API`) inside flashinfer; if present, we enable the compatibility adjustments only in that mode.

Would these minimal, environment-gated changes be acceptable to the flashinfer maintainers?

## What we tested

- We tested the `flashinfer.fused_moe.cutlass_fused_moe` interface. With the compatibility mode enabled and some additional Python-side compatibility work in progress, we successfully ran fp16 unit tests for that interface.
- We plan to incrementally expand test coverage and run more of flashinfer's unit tests in CI as part of further collaboration.

## Next steps (proposed)

- If you’re open to collaboration, we can:
  - Open a PR with the smallest possible changes to flashinfer (clearly marked and gated).
  - Add flashinfer tests into PaddlePaddle's CE system (run daily under `PADDLE_COMPATIBLE_API=1`) and gradually increase coverage.

Thank you for reviewing this PR — we welcome your feedback on the minimal integration approach and are ready to iterate on the branch or make any changes you prefer.

[^1]: https://github.com/PaddlePaddle/Paddle/tree/develop/paddle/phi/api/include/compat
[^2]: https://github.com/PaddlePaddle/Paddle/blob/b38a9503d4f3f7c84af44a6399bb76ee043e7616/python/paddle/compat.py#L110


yzh119:
Hi @SigureMo we plan to go with tvm-ffi to replace the current PyTorch bindings (wip in #1641), it should satisfy what paddle need (can you double checking it?)

Cc @tqchen

SigureMo:
Hi @yzh119 — thanks for the work here! I only learned about the recent TVM FFI efforts in the last couple of days. TVM FFI is indeed an excellent FFI solution for the ML systems — thanks to you and @tqchen for driving this.

From what I see, TVM FFI can nicely decouple flashinfer from PyTorch by providing a framework-agnostic binding layer, which aligns well with the goals of our compatibility approach. This should remove a lot of pain points in our custom C++ operator ecosystem — at minimum we wouldn’t need to worry about C++ ABI/operator registration compatibility anymore. I did a quick look into the implementation and it seems we would likely only need a small adaptation for CUDA stream handling in TVM FFI (see: https://github.com/apache/tvm/blob/a819115375568e52f9d2d7376cdbb0a23346c3cb/ffi/python/tvm_ffi/cython/function.pxi#L110-L124). So I’m looking forward to your refactor.

Separately, TVM FFI as a more general, framework-agnostic custom-op solution opens up additional possibilities and could offer more options for our ecosystem compatibility strategy. Do you have any plans to promote or adopt TVM FFI in projects beyond flashinfer? If so, that could help more custom-op projects decouple from the PyTorch ecosystem and move toward a framework-agnostic custom-op ecosystem. @tqchen

tqchen:
thanks @SigureMo , yes, we do plan to bring tvm ffi as an independent project that benefit all, we are still at the bring up stage so didn't communicate broadly, but yes the goal is to make it a general project that can be used across all deep learning frameworks, compilers, and libraries

yzh119:
https://github.com/flashinfer-ai/flashinfer/pull/1641 is merged, @SigureMo would you mind checking whether it's helpful for paddle compatibility?

SigureMo:
> #1641 is merged, @SigureMo would you mind checking whether it's helpful for paddle compatibility?

@yzh119 Thanks for the work on #1641! I can confirm the C++ layer no longer depends on PyTorch after that change, which removes the adapter maintenance we were carrying on our side—really appreciate it.

I did notice the Python JIT workflow still references torch headers and some torch-specific compile flags.

https://github.com/flashinfer-ai/flashinfer/blob/08b8da35ffc26454415a9b833a69e114a645ede3/flashinfer/jit/cpp_ext.py#L98-L119

Do you plan to remove those as well?

On the E2E validation: we already landed the Paddle prerequisites (PaddlePaddle/Paddle#75193 and PaddlePaddle/Paddle#75205), so I’m optimistic flashinfer will run on Paddle as smoothly as it does on PyTorch. I’ll run the verification soon—likely right after the holiday.

yzh119:
> Do you plan to remove those as well?

Yes most of them are no longer required, updated in https://github.com/flashinfer-ai/flashinfer/pull/1795

PR #1644, opened 2025-09-05
Added mx_fp4 support using the cudnn backend

### LINKED ISSUE #1592: MXFP4 * MXFP4 GEMM
Request from @mgoin for vLLM.

Motivation from @LopezCastroRoberto: 

> we have some ongoing case studies (QAT and PTQ) where MXFP4 with online transformations looks particularly relevant, so having those kernels would be great.

nvmbreughe:
Fixes #1592

Updated mm_fp4 API:
* alpha -- now optional (only needed for nvfp4, not for mx_fp4)
* use_nvfp4 -- new parameter to distinguish between nvfp4 and mx_fp4

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/flashinfer/gemm.py
+++ b/flashinfer/gemm.py
@@ -1466,6 +1466,7 @@ def build_cudnn_gemm_block_scale_dequantize_graph(
     o_type,
     block_size,
     device,
+    use_nvfp4,
 ):
     _check_cudnn_availability()
     stream = torch.cuda.current_stream(device)
@@ -1490,12 +1491,7 @@ def build_cudnn_gemm_block_scale_dequantize_graph(
             data_type=scale_type,
             reordering_type=cudnn.tensor_reordering.F8_128x4,
         )
-        global_scale_cudnn_tensor = graph.tensor(
-            name="global_scale",
-            dim=(1, 1, 1),
-            stride=(1, 1, 1),
-            data_type=cudnn.data_type.FLOAT,
-        )
+
         dequant_a_tensor = graph.block_scale_dequantize(
             a_cudnn_tensor,
             block_descale_a_cudnn_tensor,
@@ -1518,19 +1514,30 @@ def build_cudnn_gemm_block_scale_dequantize_graph(
         )
         c_tensor.set_data_type(cudnn.data_type.FLOAT)
 
-        c_final_cudnn_tensor = graph.mul(
-            name="scale_mul",
-            a=c_tensor,
-            b=global_scale_cudnn_tensor,
-            compute_data_type=cudnn.data_type.FLOAT,
-        )
+        c_final_cudnn_tensor = c_tensor
+
+        # if use_nvfp4 is True, we need to multiply the output by the global scale
+        if use_nvfp4:
+            global_scale_cudnn_tensor = graph.tensor(
+                name="global_scale",
+                dim=(1, 1, 1),
+                stride=(1, 1, 1),
+                data_type=cudnn.data_type.FLOAT,
+            )
+            c_final_cudnn_tensor = graph.mul(
+                name="scale_mul",
+                a=c_tensor,
+                b=global_scale_cudnn_tensor,
+                compute_data_type=cudnn.data_type.FLOAT,
+            )
+            global_scale_cudnn_tensor.set_uid(UIDs.ALPHA_UID.value)
+
         c_final_cudnn_tensor.set_name("c_final").set_output(True).set_data_type(o_type)
 
         a_cudnn_tensor.set_uid(UIDs.A_UID.value)
         b_cudnn_tensor.set_uid(UIDs.B_UID.value)
         block_descale_a_cudnn_tensor.set_uid(UIDs.BLOCK_DESCALE_A_UID.value)
         block_descale_b_cudnn_tensor.set_uid(UIDs.BLOCK_DESCALE_B_UID.value)
-        global_scale_cudnn_tensor.set_uid(UIDs.ALPHA_UID.value)
         c_final_cudnn_tensor.set_uid(UIDs.O_UID.value)
 
         graph.validate()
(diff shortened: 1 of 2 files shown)


flashinfer/gemm.py
```diff
@@ -1883,22 +1903,29 @@ def mm_fp4(
...
+        raise ValueError("alpha must be provided for nvfp4 quantization.")
+
     if out_dtype not in (torch.bfloat16, torch.float16):
         raise ValueError(
             f"Unsupported output dtype: {out_dtype}. "
             f"Only torch.bfloat16 and torch.float16 are supported for FP4 GEMM operations."
         )
-    if block_size != 16:
-        raise ValueError("Only block_size = 16 is supported for FP4 GEMM operations.")
+    if block_size != 16 and not (block_size == 32 and not use_nvfp4):
```
ttyio: change the check here, block_size 16 is for nvfp4 only?

flashinfer/gemm.py
```diff
@@ -1883,22 +1903,30 @@ def mm_fp4(
...
         raise ValueError(
             f"Unsupported output dtype: {out_dtype}. "
             f"Only torch.bfloat16 and torch.float16 are supported for FP4 GEMM operations."
         )
-    if block_size != 16:
-        raise ValueError("Only block_size = 16 is supported for FP4 GEMM operations.")
+
+    if use_nvfp4 and block_size != 16:
+        raise ValueError("nvfp4 only supports block_size = 16.")
+    if not use_nvfp4 and block_size not in [16, 32]:
```
ttyio: should we only allow block size 32 for mxfp4?
I checked some reference: 
https://docs.vllm.ai/en/latest/api/vllm/model_executor/layers/quantization/mxfp4.html#__codelineno-0-279
https://developer.nvidia.com/blog/nvfp4-trains-with-precision-of-16-bit-and-speed-and-efficiency-of-4-bit/

flashinfer/gemm.py
nvmbreughe: mxfp4 can use both 16 or 32 (in theory it can be arbitrary). I updated the check to make this more clear, and also randomized block size in the tester.

flashinfer/gemm.py
```diff
@@ -1828,14 +1842,17 @@ def mm_fp4(
...
-        Block size for FP4 quantization, only 16 is supported.
+        Block size for FP4 quantization, only 16 and 32 are supported.
 
     use_8x4_sf_layout: bool
         Whether to use 8x4 scale factor layout or 128x4 scale factor layout, defaults to False.
 
     backend: Literal["cudnn", "trtllm", "cutlass"]
         Backend to use, defaults to "cudnn".
 
+    use_nvfp4: bool
```
yzh119: How about inferring from scale data types instead?
When descale tensor has data type `float8_e4m3fn`, uses mxfp4, when descale tensor has data type `float8_e8m0fnu` (currently represented by int8 instead because e8m0 is only available from torch 2.8 on).

flashinfer/gemm.py
```diff
@@ -1939,15 +1967,16 @@ def mm_fp4(
             expanded_b_descale_shape,
             expanded_b_descale_stride,
             cudnn.data_type.FP4_E2M1,
-            torch.float8_e4m3fn,
+            torch.float8_e4m3fn if use_nvfp4 else torch.float8_e8m0fnu,
```
yzh119: Also note that `torch.float8_e8m0fnu` is introduced in torch 2.8 and we should care about the case of torch 2.7 before it's end-of-life.

flashinfer/gemm.py
nvmbreughe: The problem is that torch.uint8 may also be used for nvfp4.
If we want to avoid an extra boolean, we can distinguish either by 1) checking if alpha is None, as mxfp4 does not need a global scale, or 2) the block size (16 for nvpf4, 32 for mxfp4). My preference would be 1) as technically non-standard mxpf4 can have different block sizes. Does that sound ok? Or do we prefer an explicit parameter as we have now?

tests/test_mm_fp4.py
```diff
@@ -53,11 +69,13 @@ def test_mm_fp4(m, n, k, res_dtype, backend, use_128x4_sf_layout, auto_tuning):
             mat2_fp4.T,
             input_inv_s,
             mat2_inv_s.T,
-            alpha,
+            alpha if use_nvfp4 else None,
             res_dtype,
             res,
+            block_size=16 if use_nvfp4 else 32,
```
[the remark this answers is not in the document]
nvmbreughe: Check was updated to 32 only.

flashinfer/gemm.py
nvmbreughe: Thank you for pointing this out.
Since mxfp4 needs e8m0, is it ok to skip the test for python < 2.8?

flashinfer/gemm.py
nvmbreughe: Addressed with option 1).

flashinfer/gemm.py
nvmbreughe: Addressed. Thanks again.

yzh119: > Since mxfp4 needs e8m0, is it ok to skip the test for python < 2.8?

Before torch2.7, we use uint8 as container of e8m0 (and we still keep this behavior for quantization kernels), we don't rely on the torch operators upon e8m0 so I think it's okay to keep the test for earlier torch versions.

nvmbreughe: Hi @yzh119 , I removed the dependency on the torch type completely so no version check is needed. The code looks cleaner this way. Thanks again for your feedback.

Just fyi: we do need to tell cudnn about the scale type when building the cudnn graph, but that is a different type: cudnn.data_type.FP8_E8M0. This is orthogonal to the pytorch version, and the _check_cudnn_fp4_availability is a sufficient condition for this data type to be present.

yzh119: Good to know, yes I think `_check_cudnn_fp4_availability` should work.

Issue #1634, opened 2025-09-04
Feature Request: Raise exception if sampling params don't match expected shape


netanel-haber:
The docs clearly state the expected value constraints, which is great, [e.g.](https://docs.flashinfer.ai/generated/flashinfer.sampling.top_k_sampling_from_probs.html):
> top_k (Union[torch.Tensor, int]) – Either a scalar or a tensor of shape (batch_size,), representing the threshold for top-k sampling. If a scalar, the same threshold is used for all requests. If a tensor, each request has its own threshold.

I erroneously debugged weird correctness issues that turned out to be passing a scalar tensor as top_k, for example. It would have saved me time if flashinfer just asserted that it got either an actual primitive scalar or a tensor of the correct size. Currently it just silently produces wrong results.

The code below clearly showcases the current surprising (IMO) behavior:
```python
import torch
from flashinfer.sampling import top_k_renorm_probs
torch.manual_seed(42)

K = 3
batch_size, vocab_size = 2, 4
pre_norm_prob = torch.rand(batch_size, vocab_size).to(0)
prob = pre_norm_prob / pre_norm_prob.sum(dim=-1, keepdim=True)

print(top_k_renorm_probs(prob, K))
"""
tensor([[0.3201, 0.3319, 0.0000, 0.3480],
        [0.2187, 0.3366, 0.0000, 0.4446]], device='cuda:0')"""

print(top_k_renorm_probs(prob, torch.tensor([[K] * vocab_size] * batch_size, dtype=torch.int).to(0)))
"""
tensor([[0.3201, 0.3319, 0.0000, 0.3480],
        [0.2187, 0.3366, 0.0000, 0.4446]], device='cuda:0')"""

print(top_k_renorm_probs(prob, torch.tensor(K, dtype=torch.int).to(0)))
"""
tensor([[0.3201, 0.3319, 0.0000, 0.3480],
        [0.1913, 0.2943, 0.1257, 0.3887]], device='cuda:0')"""

print(top_k_renorm_probs(prob, torch.tensor([K], dtype=torch.int).to(0)))
"""
tensor([[0.3201, 0.3319, 0.0000, 0.3480],
        [0.1913, 0.2943, 0.1257, 0.3887]], device='cuda:0')"""
```

raayandhar:
@yzh119 I'd like to contribute to FlashInfer. Could this be a good first issue?

yzh119:
[...]

PR #1652, opened 2025-09-08
fix: add _check_tensor_params to check correct sampling parameters and dtype validation in decode.py

raayandhar:
## 📌 Description

This adds the `_check_tensor_params` function in `sampling.py` that is used in the different sampling functions to check that the parameters make sense for this function (see related issues). This also adds some `dtype` validation checks in `decode.py` to make sure we don't silently produce wrong results when passing in the wrong `dtype`. 

I tried to make just one small function that would cover the cases. But please let me know if there is a case I missed. Looking forward to the feedback. 

## 🔍 Related Issues

This was originally from this issue https://github.com/flashinfer-ai/flashinfer/issues/1634 that showed that we can pass relatively bogus inputs to sampling function and get wrong output. Now, when running his code, only the first case passes, which seems correct to me. The second case is passing a 2D tensor (2, 4) when the docs should indicate it should be (2,) (which passes). If it should be the other way around, let me know. The output is the same as well. 

For `dtype` validation, this was originally from https://github.com/flashinfer-ai/flashinfer/issues/1654. I add a check that is very similar to the ones I saw in `mla.py`.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

Added unit tests `test_check_tensor_params_top_k`, `test_check_tensor_params_top_p`, `test_check_tensor_params_min_p` that cover different functions and three different cases. Tested the `dtype` validation changes locally.

### DIFF
--- a/flashinfer/decode.py
+++ b/flashinfer/decode.py
@@ -834,12 +834,12 @@ class BatchDecodeWithPagedKVCacheWrapper:
         Parameters
         ----------
         indptr : torch.Tensor
-            The indptr of the paged kv cache, shape: ``[batch_size + 1]``
+            The indptr of the paged kv cache, shape: ``[batch_size + 1]``, dtype: ``torch.int32``
         indices : torch.Tensor
-            The page indices of the paged kv cache, shape: ``[kv_indptr[-1]]``
+            The page indices of the paged kv cache, shape: ``[kv_indptr[-1]]``, dtype: ``torch.int32``
         last_page_len : torch.Tensor
             The number of entries in the last page of each request in the paged kv
-            cache, shape: ``[batch_size]``
+            cache, shape: ``[batch_size]``, dtype: ``torch.int32``
         num_qo_heads : int
             The number of query/output heads
         num_kv_heads : int
@@ -895,6 +895,16 @@ class BatchDecodeWithPagedKVCacheWrapper:
 
         The :meth:`plan` method cannot be used in Cuda Graph or in ``torch.compile``.
         """
+        for tensor, name in [
+            (indptr, "indptr"),
+            (indices, "indices"),
+            (last_page_len, "last_page_len"),
+        ]:
+            if tensor.dtype != torch.int32:
+                raise ValueError(
+                    f"{name} must have dtype torch.int32, got {tensor.dtype}"
+                )
+
         self._workspace_size = (
             self._float_workspace_buffer.numel()
             * self._float_workspace_buffer.element_size()
@@ -1609,12 +1619,12 @@ class BatchDecodeMlaWithPagedKVCacheWrapper:
         Parameters
         ----------
         indptr : torch.Tensor
-            The indptr of the paged kv cache, shape: ``[batch_size + 1]``
+            The indptr of the paged kv cache, shape: ``[batch_size + 1]``, dtype: ``torch.int32``
         indices : torch.Tensor
-            The page indices of the paged kv cache, shape: ``[qo_indptr[-1]]``
+            The page indices of the paged kv cache, shape: ``[qo_indptr[-1]]``, dtype: ``torch.int32``
         last_page_len : torch.Tensor
             The number of entries in the last page of each request in the paged kv
-            cache, shape: ``[batch_size]``
+            cache, shape: ``[batch_size]``, dtype: ``torch.int32``
         num_qo_heads : int
             The number of query/output heads
         head_dim_compressed_kv : int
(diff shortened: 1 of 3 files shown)


netanel-haber:
Perhaps add a unittest that uses originial issue's repro code with `self.assertRaises`?

flashinfer/sampling.py
```diff
@@ -463,6 +463,13 @@ def _to_tensor_scalar_tuple(x):
         return (None, x)
 
 
+def _check_tensor_param(param, tensor):
+    if isinstance(param, torch.Tensor) and (
+        param.dim() == 0 or param.dim() > 1 or param.shape[0] != tensor.shape[0]
+    ):
+        raise ValueError("Wrong input format")
```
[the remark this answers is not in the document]
netanel-haber: Second this in spirit - the exception message should be more descriptive.

raayandhar:
> Perhaps add a unittest that uses originial issue's repro code with `self.assertRaises`?

Yes, good idea will add.

Edit: have added this unit test.

flashinfer/sampling.py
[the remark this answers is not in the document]
raayandhar: Yes, will update.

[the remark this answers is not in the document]
raayandhar: Added some better error messages, let me know what you think?

tests/test_sampling.py
```diff
@@ -557,6 +557,50 @@ def test_chain_speculative_sampling(
         assert torch.all(emitted_num + 1 == (output_token_ids != -1).sum(dim=1))
 
 
+def test_check_tensor_param():
+    torch.manual_seed(42)
+
+    K = 3
+    batch_size, vocab_size = 2, 4
```
yzh119: better to parametrize these arguments with `pytest.parametrize` and try more combinations.

raayandhar: Yes, sounds good, will do

raayandhar: I parameterized it with some basic numbers, happy to change on your suggestion.
I am also wondering if we should testing the other sampling functions too, since they have this check. Looking for your feedback on how we can accomplish this, if you think it's a good idea? Maybe try to parameterize the function too. But that sounds messy since inputs are not always the same. Can try splitting into cases, but introduces more a lot more code. Looking forward to your feedback.

raayandhar: I have picked parameterizations that are consistent with the other tests in the file. I also added two other tests, one for `min_p_sampling`and `top_p_renorm_probs`. I think these are good because they also test the `float` case and the sampling case (returning `(batch_size,)` instead of `(batch_size, num_classes)`. 

I think if we want to cover every single case, I can add these tests into the other testing functions, instead of having them as separate cases, include them in i.e. `test_top_p_renorm_probs`, etc. I think this might be too invasive, though.

Let me know your thoughts @yzh119.

raayandhar:
I have added `dtype` validation in `decode.py` that is consistent with the same way it's done in `mla.py`. I also included this validation not only for `last_page_len` like the issue poster describes but for `indptr` and `indices` as well. Please let me know if that is not correct.

flashinfer/decode.py
```diff
@@ -895,6 +895,16 @@ def plan(
 
         The :meth:`plan` method cannot be used in Cuda Graph or in ``torch.compile``.
         """
+        for tensor, name in [
```
yzh119, line 898: Later on we can move them to C++ side to save python overhead.

flashinfer/sampling.py
```diff
@@ -463,6 +463,27 @@ def _to_tensor_scalar_tuple(x):
         return (None, x)
 
 
+def _check_tensor_param(param: Any, tensor: torch.Tensor) -> None:
```
yzh119, line 466: This could also be C++ side function.

raayandhar:
> Hi @raayandhar Thanks for your contribution, and I'm good with the changes here.
> 
> We can keep improving this by moving the checks to C++ side to reduce python overhead (in later PRs).

I see, that sounds better. I can try that as well and make a PR later. There are other checks that I was looking at as reference that can also be moved to C++ side beyond the ones here so I will try to include those as well.

yzh119:
Hi @raayandhar it's better to do that after #1641 where we changed a lot on C++ side structure (we will merge it today).

raayandhar:
> Hi @raayandhar it's better to do that after #1641 where we changed a lot on C++ side structure (we will merge it today).

I see, thanks for letting me know.

PR #1659, opened 2025-09-09
Tiny allow compiling with line info and release moe

fzyzcjy:
## 📌 Description

useful for ncu

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/flashinfer/jit/core.py
+++ b/flashinfer/jit/core.py
@@ -197,6 +197,10 @@ def gen_jit_spec(
         # non debug mode
         cuda_cflags += ["-DNDEBUG"]
 
+    # useful for ncu
+    if bool(os.environ.get("FLASHINFER_JIT_LINEINFO", "0")):
+        cuda_cflags += ["-lineinfo"]
+
     if extra_cflags is not None:
         cflags += extra_cflags
     if extra_cuda_cflags is not None:


flashinfer/jit/core.py
```diff
@@ -197,6 +197,10 @@ def gen_jit_spec(
         # non debug mode
         cuda_cflags += ["-DNDEBUG"]
 
+    # useful for ncu
+    if bool(os.environ.get("FLASHINFER_JIT_LINEINFO", "0")):
```
yzh119, line 201: it's covered in `FLASHINFER_JIT_VERBOSE`

fzyzcjy: good point, btw it seems ptxas's register-usage-level (enabled in the verbose path) is:

```
--register-usage-level <0..10>                      (-regUsageLevel)            
        Controls the aggressiveness of optimizations that affect register usage.
        ([0..10], default = 5) Higher values aggressively optimize the source program,
        trading off additional register usage for potential improvements in the generated
        code. Lower values inhibit optimizations that aggressively increase register
        usage. This option can work in conjunction with -maxrregcount and CUDA launch
        bounds. This is a BETA feature for advanced users and there is no guarantee
        that the implementation stays consistent between ptxas releases.
        Default value:  5.
```

thus I am not very sure whether it changes program behavior

yzh119, line 201: `bool("0")` evaluates to `True`, so `bool(os.environ.get("FLASHINFER_JIT_LINEINFO", "0"))` will be True if this environment variable is not set.

This bug is fixed in #1872 , note that `-lineinfo` will greatly increase binary size, see my comment in https://github.com/flashinfer-ai/flashinfer/pull/1872#issuecomment-3374843208

cc @fzyzcjy @zhyncs  for viz

fzyzcjy: Good point, I should have written `bool(int(...))`...

PR #1661, opened 2025-09-09
perf: skip kv-tile computation out of sliding window in FA2

happierpig:
## 📌 Description

This PR skips redundant kv-tiles loading&computation in FA2's sliding window implementation, which is used by GPT-OSS inference scenarios. Main modification:
1. Make `PrefillPlanSM80` aware of sliding window: skipping redundant attention states merge
2. Add `kv-start-idx` offsets at the start of each device kernel, to skip the window computation. As it maintains the original 0-base `kv-tile-idx`, it should be compatible w/ jit `sink-token` implementation.
image

## 🔍 Related Issues
This should backup https://github.com/vllm-project/vllm/pull/23010.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/include/flashinfer/attention/scheduler.cuh
+++ b/include/flashinfer/attention/scheduler.cuh
@@ -497,7 +497,7 @@ inline auto PrefillSplitQOKVIndptr(IdType* qo_indptr_h, IdType* kv_indptr_h,
                                    uint32_t total_num_rows, uint32_t batch_size,
                                    uint32_t num_qo_heads, uint32_t num_kv_heads, uint32_t head_dim,
                                    uint32_t page_size, uint32_t max_batch_size_if_split,
-                                   bool enable_cuda_graph) {
+                                   bool enable_cuda_graph, int32_t window_left) {
   std::vector<IdType> request_indices, qo_tile_indices, kv_tile_indices, merge_indptr, o_indptr;
   merge_indptr.push_back(0);
   o_indptr.push_back(0);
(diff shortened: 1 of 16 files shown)


include/flashinfer/attention/scheduler.cuh
```diff
@@ -554,17 +554,25 @@ inline auto PrefillSplitQOKVIndptr(IdType* qo_indptr_h, IdType* kv_indptr_h,
     }
   }
 
+  // Calculate the actual needed CTA when considering sliding window
+  std::vector<int64_t> cliped_kv_len_arr(batch_size);
```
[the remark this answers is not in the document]
yzh119: agreed with gemini here, please rename to "clipped", or "effective"

tests/test_sliding_window.py
```diff
@@ -207,14 +217,15 @@ def test_single_prefill_sliding_window(
...
-@pytest.mark.parametrize("qo_len", [37, 47])
-@pytest.mark.parametrize("window_left", [13, 33])
-@pytest.mark.parametrize("num_kv_heads", [1, 4])
-@pytest.mark.parametrize("num_qo_heads", [4, 8])
+@pytest.mark.parametrize("batch_size", [12, 17, 71])
+@pytest.mark.parametrize("kv_len", [54, 397, 1177, 2999])
+@pytest.mark.parametrize("qo_len", [1, 37, 47])
+@pytest.mark.parametrize("window_left", [13, 33, 111])
+@pytest.mark.parametrize("num_kv_heads", [32])
+@pytest.mark.parametrize("num_qo_heads", [32])
```
yzh119: Please don't remove the GQA case, changing sliding windows' behavior might influence GQA because of the use of GQA packing.

include/flashinfer/attention/cascade.cuh
```diff
@@ -386,6 +386,9 @@ __global__ void PersistentVariableLengthMergeStatesKernel(
 
 #pragma unroll 1
   for (uint32_t i = cta_id; i < seq_len * num_heads; i += num_ctas) {
+    // NOTE (Yilong): necessary to prevent hazard on smaller `num_index_sets`
+    __syncthreads();
```
Edenzzzz, line 390: do you mean `v_smem` and `s_smem` pollution here?

yzh119:
This PR brings 2% performance degradation for decoding (w/o sliding window), would you mind investigating that?

One easy solution is to separate the sliding window/non sliding window kernels at compile time, but better to understand where the degradation is coming from.

include/flashinfer/attention/cascade.cuh
happierpig: Yes, dependency between `cp.async` and `v_smem`. It may happen that the next CTA starts loading from gmem before the previous CTA consumes all the v_smem. I will get precision issue without this `__syncthreads()`.

happierpig:
> This PR brings 2% performance degradation for decoding (w/o sliding window), would you mind investigating that?
> 
> One easy solution is to separate the sliding window/non sliding window kernels at compile time, but better to understand where the degradation is coming from.

Sure! It will be good if there is a reproducible perf benchmark on hopper.

happierpig:
@yzh119 With `benchmarks/bench_batch_decode.py` and H200, i cannot reproduce the bandwidth drop. The small delta looks like measurement error.
image

If perf degrades, probably it's because this PR adds one more `__sync_threads()` in `MergeStates` (https://github.com/flashinfer-ai/flashinfer/blob/88e333e038c6fce317e959261f165a0618fa4f3c/include/flashinfer/attention/cascade.cuh#L389), which should be necesseary for precision IMO.

include/flashinfer/attention/scheduler.cuh
```diff
@@ -554,16 +554,25 @@ inline auto PrefillSplitQOKVIndptr(IdType* qo_indptr_h, IdType* kv_indptr_h,
     }
   }
 
+  // Calculate the actual needed CTA when considering sliding window
+  std::vector<int64_t> effective_kv_len_arr(batch_size);
+  for (uint32_t i = 0; i < batch_size; ++i) {
+    // pad CTA_TILE_Q to consider the causal kv-len
+    effective_kv_len_arr[i] =
+        std::min(window_left >= 0 ? ceil_div(window_left + cta_tile_q, page_size) : kv_len_arr[i],
+                 kv_len_arr[i]);
```
Edenzzzz, line 563: ~~Looks like this is not the causal kv len? Should we use `packed_causal_kv_end` when computing `num_chunks_kv` and computing `new_batch_size`?~~

https://github.com/flashinfer-ai/flashinfer/blob/6832b4672cdd74ff049a565aea216bd8c035e114/include/flashinfer/attention/scheduler.cuh#L933

EDIT: Looks like this will not help much, because if #q tiles is large it will usually not trigger split kv, if it's small then causality won't increase the kv length diff between q tiles to above chunk size

include/flashinfer/attention/cascade.cuh
Edenzzzz: Might actually be better to add sync after the for loop in all state sync util functions
https://github.com/flashinfer-ai/flashinfer/blob/35e099ee7f344d2ecf65429e2b93e2587f590e7b/include/flashinfer/attention/cascade.cuh#L153

PR #1662, opened 2025-09-09
benchmark: add cupti support to benchmark

nv-yunzheq:
## 📌 Description

Add CUPTI support for more accurate benchmarking. 
Require cupti-python>=13.0.0. Off by default

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/flashinfer/testing/utils.py
+++ b/flashinfer/testing/utils.py
@@ -21,6 +21,7 @@ from typing import Tuple, Any
 
 import os
 import sys
+import warnings
 
 import numpy as np
 import torch
@@ -642,11 +643,12 @@ def bench_gpu_time_with_cupti(
     """
     Benchmark GPU time using by using CUPTI to measure the kernel execution time.
     """
+    import warnings
     try:
         from cupti import cupti
         from functools import partial
     except ImportError:
-        print("CUPTI is not installed. Falling back to bench_gpu_time.")
+        warnings.warn("CUPTI is not installed. Falling back to bench_gpu_time.")
         return bench_gpu_time(
             fn, dry_run_iters, repeat_iters, dry_run_time_ms, repeat_time_ms,
             l2_flush, l2_flush_size_mb, l2_flush_device, sleep_after_run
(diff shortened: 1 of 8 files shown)


bkryu:
Changes seem mostly mechanical except for the addition of benchmarking with CUPTI. Upvoted, but can I get some insight on the implications of using CUPTI?

I'm thinking about:
- Impact on benchmarking time. Does each test case take much longer than using CUDA events?
- Do the measured times differ from using CUDA events or CUDA graphs?
- Do measurements done with CUPTI get affected by throttling in the same way as we would for CUDA events?

flashinfer/testing/utils.py
```diff
@@ -629,6 +630,226 @@ def bench_gpu_time(
...
+        List[float]: Measured times in milliseconds per iteration.
+    """
+    # check if CUPTI is installed and its version is >= 13.0.0
+    try:
+        from cupti import cupti
+        from importlib.metadata import version as importlib_metadata_version
+
+        cupti_version = importlib_metadata_version("cupti-python")
+        if int(cupti_version.split(".")[0]) < 13:
+            raise Exception("CUPTI needs to be >= 13.0.0.")
```
yzh119: can you also add some instructions on how to install the latest version of cupti (e.g. `pip install -U nvidia-cuda-cupti`)

flashinfer/testing/utils.py
```diff
@@ -629,6 +630,226 @@ def bench_gpu_time(
...
+        from functools import partial
+    except (ModuleNotFoundError, Exception) as e:
+        if isinstance(e, ModuleNotFoundError):
+            warnings.warn(
+                "CUPTI is not installed. Falling back to use cuda events.",
```
yzh119: ditto

nv-yunzheq:
> Changes seem mostly mechanical except for the addition of benchmarking with CUPTI. Upvoted, but can I get some insight on the implications of using CUPTI?
> 
> I'm thinking about:
> 
> * Impact on benchmarking time. Does each test case take much longer than using CUDA events?
> * Do the measured times differ from using CUDA events or CUDA graphs?
> * Do measurements done with CUPTI get affected by throttling in the same way as we would for CUDA events?

For CUPTI, it would capture all the start and end time of cuda runtime calls on CPU side, and all the corresponding start and end time of cuda kernels on GPU side.
We also inject a marker before and after each python function call. By comparing these markers and start time of cuda runtime calls, we could know which kernels being launched for one python function call. Then we use the span from 'min start time of all kernels’ to ‘max end time of all kernels’ to get the kernel time for one python function call.

> Impact on benchmarking time. Does each test case take much longer than using CUDA events?

It should not. But I haven't done any test.

> Do the measured times differ from using CUDA events or CUDA graphs?

Yes, it differs. But ideally this should be closer to the timing result from nsys profile.

> Do measurements done with CUPTI get affected by throttling in the same way as we would for CUDA events?

Yes.

PR #1664, opened 2025-09-10
feat: Support incremental prefill in flashinfer_benchmark.py and benchmark minor updates

bkryu:
## 📌 Description

`flashinfer_benchmark.py`'s attention prefill currently assumes `s_qo = s_kv`, which blocks measuring incremental performance where `s_qo < s_kv`. Current PR enables benchmarking for this case.

Additional minor updates and fixes to the benchmark code:
* **Removed `try/catch` blocks during benchmarking**: A try-catch block containing PyTorch tensors can lead to memory leaks, especially on CUDA devices, if exceptions interrupt the normal lifecycle of the tensors and result in tensors not being properly cleared or released.
* **Raise errors only when necessary** while support checks result in graceful exits for better benchmarking behavior when running batch tests.
* Head dim (192 or 128) constraint in `testBatchPrefillWithRaggedKVCacheWrapper` removed as it is no longer applicable.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/benchmarks/routines/attention.py
+++ b/benchmarks/routines/attention.py
@@ -38,7 +38,8 @@ def run_attention_test(args):
     elif args.routine == "BatchMLAPagedAttentionWrapper":
         return testBatchMLAPagedAttentionWrapper(args)
     else:
-        raise ValueError(f"Unsupported routine: {args.routine}")
+        print(f"[ERROR] Unsupported routine: {args.routine}")
+        return []
 
 
 def parse_attention_args(line, parser):
@@ -213,16 +214,19 @@ def testBatchDecodeWithPagedKVCacheWrapper(args):
     kv_init_dtype = torch.bfloat16
     rtol = 2e-1
     atol = 1e-2
+    res = []
 
     # Handle different query data types.
     q_dtype = dtype_str_to_torch_dtype(args.q_dtype)
     if q_dtype not in [torch.bfloat16, torch.float8_e4m3fn]:
-        raise ValueError(f"Unsupported q_dtype: {args.q_dtype}")
+        print(f"[ERROR] Unsupported q_dtype: {args.q_dtype}")
+        return res
 
     # Handle different KV cache data types.
     kv_dtype = dtype_str_to_torch_dtype(args.kv_dtype)
     if kv_dtype not in [torch.bfloat16, torch.float8_e4m3fn]:
-        raise ValueError(f"Unsupported kv_dtype: {args.kv_dtype}")
+        print(f"[ERROR] Unsupported kv_dtype: {args.kv_dtype}")
+        return res
 
     # Parse and validate backend configurations
     backends = args.backends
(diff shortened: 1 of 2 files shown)


benchmarks/routines/attention.py
```diff
@@ -1150,6 +1174,11 @@ def testBatchPrefillWithRaggedKVCacheWrapper(args):
         ]:
             print("[INFO] CUTLASS backend does not support FP8. Skipping.")
             remove_cutlass = True
+        if not (
+            (head_dim_qk == 128 and head_dim_qk == head_dim_vo) or head_dim_qk == 192
+        ):
+            print("[INFO] CUTLASS backend requires head dimension to be 128 or 192")
```
yzh119, line 1180: We can add other head dimensions in cutlass backends as well but thanks for adding this check here!

benchmarks/routines/attention.py
```diff
@@ -719,29 +723,49 @@ def testBatchPrefillWithPagedKVCacheWrapper(args):
...
+    # If s_qo == s_kv, then make sampled actual_seq_lens_kv the same as actual_seq_lens_q.
+    # IF s_qo < s_kv, then sample actual_seq_lens_kv separately. Then ensure actual_seq_lens_kv is at least as long as actual_seq_lens_q.
     actual_seq_lens_q = sample_actual_seq_lens(
         s_qo, batch_size, None, args.random_actual_seq_len
     )
-    actual_seq_lens_kv = actual_seq_lens_q.clone()
+    if s_qo == s_kv:
+        if args.verbose >= 2:
+            print(
+                "[VVERBOSE] s_qo == s_kv, making actual_seq_lens_kv the same as actual_seq_lens_q"
```
yzh119, line 750: What does the first "V" mean in "VVERBOSE", is that a typo?

bkryu: Short answer: vverbose = very verbose

Longer answer: I have set up the input flags to allow [multi-level verbosity settings](https://github.com/flashinfer-ai/flashinfer/blob/main/benchmarks/flashinfer_benchmark.py#L93-L95) in `flashinfer_benchmark.py` (also see [count action](https://docs.python.org/3/library/argparse.html#action) in argparse).

At runtime, the verbosity level is set by the number of `v`'s provided as an input flag. For example, `-v` sets `args.verbose = 1` and `-vv` sets `args.verbose = 2` and so on. In the benchmark code, I have generally been following the two-level structure where:
* If `-v` (verbose), print verbosity level 1 outputs as `[VERBOSE] message ...`
* If`-vv` (very verbose), print verbosity level 2 outputs as `[VVERBOSE] message ...` (in addition to level 1 outputs)

As of today, what should be at verbosity level 1 vs. 2 is not very well defined, but I consider debug messages to be at level 2. I chose the current approach in case the verbosity level could be used as logging levels in the future (could also extend indefinitely to `-vvv` or `-vvvv` 😄 )

yzh119: Thanks for your explanation!

PR #1665, opened 2025-09-10
test: update fused_moe test to random scale factor

yyihuang:
## 📌 Description

We could use rand scale factor rather than fixed constant when testing fp8_block_scaled fused_moe. 
We might use the same scale_factor random generation (by normal dist) rather than the dumped workload in bench.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/tests/test_trtllm_gen_fused_moe.py
+++ b/tests/test_trtllm_gen_fused_moe.py
@@ -1626,7 +1626,6 @@ def run_moe_reference_fp4(args, quant_mode: QuantMode):
 def run_moe_reference_dsfp8(args):
     """FP8 block-scale reference implementation."""
     # Generate block scales at runtime for FP8 block scaling
-    # Use deterministic scales for testing consistency
     hidden_states_scale = 2.0 * torch.rand(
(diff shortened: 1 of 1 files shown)


yzh119 (commented):
Can you explain the motivation of this PR in the description?

yyihuang:
> Can you explain the motivation of this PR in the description?

Added.

yzh119:
Agreed that rand is better than ones for testing.

But even rand might not precisely model data distribution, if you look at other unittests, one possible way is:
1. generate randn data (e.g. in f32)
2. quantize it to fp4/fp8 (and get the quantized data and scale tensor).

a reference can be found at https://github.com/flashinfer-ai/flashinfer/blob/7c587eb16c9236b7797525411fe8aedea0c9ec05/tests/test_groupwise_scaled_gemm_mxfp4.py#L275-L278

in this way both the fp4/fp8 data and scale should be closer to real distribution.

Can we do the same thing in this PR? (If it's too complicated, we can leave it for later PRs).

Also, please consider updating the title of the PR, it looks confusing.

yzh119 (approved):

PR #1667, opened 2025-09-10
Refactor Blackwell unit test scripts

dierksen:
## 📌 Description

Refactor the Blackwell unit test scripts so that they will run to completion regardless of failures and output JUnit xml for rendering pass/fail status.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,6 +76,7 @@ coverage.xml
 .hypothesis/
 .pytest_cache/
 cover/
+junit/
 
 # Translations
 *.mo
--- a/scripts/run_test_blackwell_attention_kernels.sh
+++ b/scripts/run_test_blackwell_attention_kernels.sh
@@ -3,14 +3,21 @@
 set -eo pipefail
 set -x
 
-pytest -s tests/test_blackwell_fmha.py
-pytest -s tests/test_deepseek_mla.py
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -76,6 +76,7 @@ coverage.xml
 .hypothesis/
 .pytest_cache/
 cover/
+junit/
```
yzh119, line 79: better to also add some junit installation scripts (e.g. in [dockerfile](https://github.com/flashinfer-ai/flashinfer/tree/main/docker)) because junit is not installed on all environments (e.g. the github CI isn't using it).

dierksen: We don't actually need junit itself; this is just pytest outputting files that we don't want added to git.

dierksen:
> > they will run to completion regardless of failures and output JUnit xml for rendering pass/fail status.
> 
> Could we make it (continue on failure/or fail fast) an option? In some cases we want to continue on failure to get full picture of failed tests, while in the other case we want to fail fast (especially for public pull requests, which will trigger CI per commit) to avoid wasting GPU hours.

Added an env variable check; if `RUN_TO_COMPLETION` is _not_ set, then it will exit immediately on failure.

PR #1668, opened 2025-09-10
TGV GEMM as a BF16 backend alternative to cuBLAS

yangs75:
## 📌 Description

We are adding a new BF16 GEMM backend named TGV as a more performant alternative to cuBLAS on small problem sizes. TGV GEMMs are optimized for minimum latency use cases and shown to achieve up to 60% speedup on B200 (gpcclk=1800MHz) compared to cuBLAS in our unit tests (see attached table). 

## 🔍 Related Issues

N/A

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

- [x] Tests have been added or updated as needed [added tests: tests/test_tgv_gemm.py, benchmarks/bench_tgv_gemm.py].
- [x] All tests are passing.

tgv_gemm

### DIFF
--- a/benchmarks/bench_tgv_gemm.py
+++ b/benchmarks/bench_tgv_gemm.py
@@ -9,7 +9,7 @@ import csv
 import torch
 import torch.nn.functional as F
 
-from flashinfer import tgv_gemm_bf16_sm100, autotune
+from flashinfer import tgv_gemm_sm100, autotune
 
 
 def test_tgv_gemm_bf16_sm100_perf():
@@ -104,7 +104,7 @@ def test_tgv_gemm_bf16_sm100_perf():
         # Warmup
         with autotune(tune_mode=True):
             for _ in range(3):
-                _ = tgv_gemm_bf16_sm100(A, B, bias)
+                _ = tgv_gemm_sm100(A, B, bias)
 
         torch.cuda.synchronize()
 
@@ -113,7 +113,7 @@ def test_tgv_gemm_bf16_sm100_perf():
         # Start graph capture
         with torch.cuda.graph(tgv_graph):
             for _ in range(100):
-                _ = tgv_gemm_bf16_sm100(A, B, bias)
+                _ = tgv_gemm_sm100(A, B, bias)
 
         # Warmup the graph
         tgv_graph.replay()
@@ -136,7 +136,7 @@ def test_tgv_gemm_bf16_sm100_perf():
         pdl_graph = torch.cuda.CUDAGraph()
         with torch.cuda.graph(pdl_graph):
             for _ in range(100):
-                _ = tgv_gemm_bf16_sm100(A, B, bias, pdl=True)
+                _ = tgv_gemm_sm100(A, B, bias, pdl=True)
 
         # Warmup the graph
         pdl_graph.replay()
@@ -195,7 +195,7 @@ def test_tgv_gemm_bf16_sm100_correctness():
     reference = torch.matmul(A, B) + bias.unsqueeze(0)
 
     # Test with TGV runner
-    out = tgv_gemm_bf16_sm100(A, B, bias)
+    out = tgv_gemm_sm100(A, B, bias)
 
     # Check correctness
     cos_sim = F.cosine_similarity(reference.reshape(-1), out.reshape(-1), dim=0)
--- a/benchmarks/bench_tgv_gemm.py
+++ b/benchmarks/bench_tgv_gemm.py
@@ -1,0 +1,229 @@
+#!/usr/bin/env python3
+"""
+Test bench for tgv_gemm_bf16_sm100() function.
+Tests the autotuner integration with TGV BF16 GEMM kernels.
+"""
+
+import time
+import csv
+import torch
+import torch.nn.functional as F
+from cuda import cudart
+from flashinfer.gemm import get_gemm_sm100_module_tgv_bf16
+
+from flashinfer import tgv_gemm_bf16_sm100, autotune
+
+
+def test_tgv_gemm_bf16_sm100_perf():
+    """Test tgv_gemm_bf16_sm100 with different problem sizes."""
+    print(f"\n=== Testing tgv_gemm_bf16_sm100 with different sizes ===")
+    
+    # Test different problem sizes
+    test_cases = [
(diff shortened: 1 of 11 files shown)


csrc/tgv_gemm.cu
```diff
@@ -0,0 +1,210 @@
...
+  TORCH_CHECK(tactic >= 0 && tactic < globalConfigs.size(), "tactic must be between 0 and ",
+              globalConfigs.size());
+  return globalConfigs[tactic];
+}
+
+
+at::Tensor bf16_gemm_impl(at::Tensor const& mat1, at::Tensor const& mat2, 
+                          std::optional<at::Tensor> bias, int64_t tactic, bool pdl) {
+  
+  // No heuristic for now, we use 128x8 with 6 DMA stages as the default tactic.
```
Yang-YiFan: Could you make `TGV_GEMM_CONFIG(64, 8, 8)` (i.e. cta_m=64, cta_n=8, DMA stage=8) to be the default tactic. And fix the comment (128x8, should be 64x8).

flashinfer/gemm.py
```diff
@@ -865,6 +865,129 @@ def get_gemm_sm120_module_cutlass_fp4():
...
+        (64, 8, 8),
+        (64, 8, 10),
+        (64, 8, 12),
+        (64, 16, 6),
+        (64, 16, 8),
+        (64, 16, 10),
+        (64, 32, 6),
+        (64, 32, 8),
+        (64, 64, 6),
+        (128, 64, 6),
```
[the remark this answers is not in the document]
Yang-YiFan: yeah looks like there is a config mismatch

csrc/tgv_gemm.cu
```diff
@@ -0,0 +1,210 @@
...
+    TORCH_CHECK(bias.value().size(0) == M, "Bias tensor must have M elements");
+    TORCH_CHECK(bias.value().scalar_type() == at::ScalarType::BFloat16, "Bias tensor must be bfloat16");
+    TORCH_CHECK(bias.value().stride(0) == 1, "Bias tensor must be M contiguous");
+  }
+
+  // Get CUDA stream
+  cudaStream_t stream = at::cuda::getCurrentCUDAStream();
+
+  // Kernel config constants
+  using TypeA = cutlass::bfloat16_t;
```
Yang-YiFan: seems like the dtype and input layout are both hard coded here and in [tgv_gemm_template.h](https://github.com/flashinfer-ai/flashinfer/pull/1668/files#diff-eefd2eda6222f5bad2f2853bfee7d215df4caaaed84b71d027965ec82ea22706R35). Is this the correct way?

happierpig:
Hey, wondering if this only works for SM100? Thanks!

yzh119:
> Hey, wondering if this only works for SM100? Thanks!

The kernel is written with blackwell features so currently I don't expect this could work on other hardwares. cc @yangs75 does the idea of tgv also work for hopper?

csrc/tgv_gemm.cu
yzh119: if we want to support fp16 as well, consider dispatching like this (`cutlass_dtype_t` will convert dtype to cutlass dtype: https://github.com/flashinfer-ai/flashinfer/blob/6dff34d1b8aadec626bd507715a5f45999458329/include/flashinfer/cutlass_utils.cuh#L52-L92):
https://github.com/flashinfer-ai/flashinfer/blob/6dff34d1b8aadec626bd507715a5f45999458329/csrc/group_gemm.cu#L30-L31

flashinfer/gemm.py
```diff
@@ -865,6 +865,129 @@ def get_gemm_sm120_module_cutlass_fp4():
...
+        tgv_bf16_gemm_runner=tgv_bf16_gemm_runner,
+    )
+
+
+def tgv_gemm_bf16_sm100(
+    a: torch.Tensor,
+    b: torch.Tensor,
+    bias: torch.Tensor,
+    pdl: bool = False,
+) -> torch.Tensor:
```
yzh119, line 987: can you add some docstring about input shape (e.g. b is column major) and dtype assumptions?

Yang-YiFan:
The code can be adapted to use the hopper mma instructions. Though the mma synchronization intrinsics and tmem related logic needs to be updated.

happierpig:
> The code can be adapted to use the hopper mma instructions. Though the mma synchronization intrinsics and tmem related logic needs to be updated.

@Yang-YiFan Thanks for replying! Will the idea also provide speedup theoretically on Hopper? The small batch (low-latency) GEMM on hopper is also super slow (including CUTLASS/CUBLAS/Triton etc.), with less than 50%ish bandwidth utilization (if i recall correctly).

Yang-YiFan:
> > The code can be adapted to use the hopper mma instructions. Though the mma synchronization intrinsics and tmem related logic needs to be updated.
> 
> @Yang-YiFan Thanks for replying! Will the idea also provide speedup theoretically on Hopper? The small batch (low-latency) GEMM on hopper is also super slow (including CUTLASS/CUBLAS/Triton etc.), with less than 50%ish bandwidth utilization (if i recall correctly).

The benefits should transfer to hopper too. But I can't estimate how much because hopper's DRAM BW is lower and latency is shorter, so you suffer less from latency problems on small batch inference on hopper. If you really want to achieve the theoretical 100% DRAM utilization (which you might not even be able to due to other critical path latency), mega kernel is the solution.

benchmarks/bench_tgv_gemm.py
```diff
@@ -0,0 +1,239 @@
...
+            cublas_graph.replay()
+
+        torch.cuda.synchronize()
+
+        # Benchmark using CUDA graph
+        start_time = time.time()
+        cublas_graph.replay()
+        torch.cuda.synchronize()
+        end_time = time.time()
+        cublas_avg_time = (end_time - start_time) / 100
```
yzh119, line 101: Consider using https://docs.flashinfer.ai/generated/flashinfer.testing.bench_gpu_time_with_cudagraph.html#flashinfer.testing.bench_gpu_time_with_cudagraph

yzh119 (approved):
Appended some commits to address the formats and more data types support comments.

Ready to merge on my side, waiting for @Yang-YiFan for double confirmation.

PR #1675, opened 2025-09-11
feat: Batch-size invariant FA2 Prefill & Decode

Edenzzzz:
## 📌 Description

As mentioned in https://thinkingmachines.ai/blog/defeating-nondeterminism-in-llm-inference/, split-kv can cause non-determinism. This PR introduces a fixed split size to produce exact precision when batch size changes. 
~~Compatibility with CUDA graph is left for future work, as varied seqlen can cause #CTAs to exceed the number when the graph was captured.~~
Also adds a `disable_split_kv` flag for CUDA Graph mode.

cc @Fridge003 I will likely be busy with some papers this semester, but this should be a useful starting point
image

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests


tests/test_invariant_batch_decode.py
```diff
@@ -0,0 +1,184 @@
...
+        lse_tensor_cores[:invariant_bs],
+        lse_tensor_cores_invariant,
+        rtol=1e-7,
+        atol=1e-7,
+    )
+
+    # test that without fixed split size, precision is different
+    # TODO: this works for the first 29 cases, but then fails with "illegal memory access"..?
+
+    # wrapper_tensor_cores.plan(
```
Edenzzzz: feel free to debug this if anyone has time
image

Edenzzzz:
cc @yzh119

yzh119:
The easiest way to turn off split-k in flashinfer is to change these two lines to `std::numeric_limits<int>::max()`:
1. https://github.com/flashinfer-ai/flashinfer/blob/88e333e038c6fce317e959261f165a0618fa4f3c/include/flashinfer/attention/scheduler.cuh#L199
2. https://github.com/flashinfer-ai/flashinfer/blob/88e333e038c6fce317e959261f165a0618fa4f3c/include/flashinfer/attention/scheduler.cuh#L527

Edenzzzz:
Yes, but simply turning it off would hurt performance

tests/test_invariant_batch_decode.py
```diff
@@ -0,0 +1,184 @@
...
+        page_size,
+        pos_encoding_mode=pos_encoding_mode,
+        data_type=torch.float16,
+        q_data_type=torch.float16,
+        fixed_split_size=fixed_split_size,
+    )
+    o_tensor_cores_invariant, lse_tensor_cores_invariant = wrapper_tensor_cores.run(
+        q[:invariant_bs], kv_data, return_lse=True
+    )
+    torch.testing.assert_close(
```
yzh119: can we add utility functions to compare bitwise identical check?

yzh119:
One minor note, the attention merge kernel (https://github.com/flashinfer-ai/flashinfer/blob/88e333e038c6fce317e959261f165a0618fa4f3c/include/flashinfer/attention/cascade.cuh#L342-L467) uses parallel reduction, I'm not sure if we need to change it to sequential reduction to guarantee full reproducibility.

Current reduction order:
1. thread 0-8: out[0] = merge(0, 16, 32, ...)
2. thread 9-16: out[1] = merge(1, 17, 33, ...)
3. ...
4. thread 120-127: out[15] = merge( 15, 31, 47, ...)

and at here https://github.com/flashinfer-ai/flashinfer/blob/88e333e038c6fce317e959261f165a0618fa4f3c/include/flashinfer/attention/cascade.cuh#L333
we reduce them into:
merge(out[0], out[1], ... out[15])

the order might change for different kv-length.

tests/test_invariant_batch_decode.py
Edenzzzz: switched to `torch.equal`, should do the job 
image

Edenzzzz:
Though I thought the threads in a block just collectively reduce along the head dim, one position & head at a time in `threadblock_sync_state`? No matter which head/seq_id it picks up.  
As long as changing one request's kv len doesn't affect other requests, it should be fine

yzh119:
@happierpig  has done some work on changing the reduction order before.

> As long as changing one request's kv len doesn't affect other requests, it should be fine

It matters if you want to guarantee the reproducibility across prefill and decode, my question is what kind of reproducibility do you want to achieve in this PR?

Edenzzzz:
> It matters if you want to guarantee the reproducibility across prefill and decode, my question is what kind of reproducibility do you want to achieve in this PR?

We want to ensure that changing the batch size(adding requests) does not change the output of individual requests, using one kernel. Not sure how reduction affects that? They both use the same reduction and batch prefill kernel

happierpig:
> the order might change for different kv-length.

I think the current `merge_states` should be fine for the batch-invariant reproducibility (which IMO means a single request's attention is exactly the same result no matter how many requests are batched together). As the parallel reduction in merge states is deterministic given an arbitrary kv-len, the order will be the same for an arbitrary kv-len, therefore the results.

Edenzzzz:
~~One more note, in the case of CUDA graph, we can binary-search a split size that just launches below `max_batch_size_if_split` CTAs, when the provided one would launch too many CTAs. This can be done in a follow up PR~~
It's better to disable split-kv for cuda graph, as if you split too much and hit that binary search, chunk size is again non-deterministic

Fridge003:
Do we need to add a flag for disabling split-kv? Since finally we want to use cuda graph for decoding.

Edenzzzz:
> Do we need to add a flag for disabling split-kv? Since finally we want to use cuda graph for decoding.

It should work now

yzh119:
To be more clear on my comments about "prefill and decode consistency":

I'm worried about the use cases where we manually merge attention outputs from different KV components (e.g. in chunked-prefill, speculative decoding), it's okay to ignore these cases at this moment but we should know their possible effect on reprodubility.

Edenzzzz:
> I'm worried about the use cases where we manually merge attention outputs from different KV components (e.g. in chunked-prefill, speculative decoding), it's okay to ignore these cases at this moment but we should know their possible effect on reprodubility.

Yes, I think if the chunk size changes, then setting a fixed split size can still lead to differences in the tail chunk for each request.
* Two issues:
1. sgl uses a [`new_token_ratio`](https://github.com/sgl-project/sglang/blob/9752861002d7b49e65f1c5ffcce2ea3dae90966b/python/sglang/srt/managers/schedule_policy.py#L310) which increases when decode hits OOM and decreases when it successfully runs. Different chunk sizes can lead to kv len being truncated in the tail chunk.
For example, assume chunk size is 4096 and kv split size is 2048, the first req is computed using split-kv. **However, if the chunk size then drops to 2048, it will not use split kv.** 
2. (Should be fine as FCFS is used by default) If you use Longest Prefix Match instead of FCFS, the order in which the scheduler fills up that chunk size is non-deterministic (truncate the last request). 

* As for speculative decoding, the varied # of tokens to verify might also lead to non-deterministic tail (though not sure if ppl ever use spec dec in RL)

In these cases we need to consistently use one BatchPagedPrefill kernel, and avoid merging different chunks using a separate `merge_state` kernel.
Radix cache shouldn't cause problems as long as you save kv cache first and only use the paged kernel.

cc @Fridge003 @merrymercy @zhyncs

Fridge003:
> > I'm worried about the use cases where we manually merge attention outputs from different KV components (e.g. in chunked-prefill, speculative decoding), it's okay to ignore these cases at this moment but we should know their possible effect on reprodubility.
> 
> 
> 
> Yes, I think if the chunk size changes, then setting a fixed split size can still lead to differences in the tail chunk for each request.
> 
> Two issues:
> 
> 1. sgl uses a [`new_token_ratio`](https://github.com/sgl-project/sglang/blob/9752861002d7b49e65f1c5ffcce2ea3dae90966b/python/sglang/srt/managers/schedule_policy.py#L310) which increases when decode hits OOM and decreases when it successfully runs. Different chunk sizes can lead to kv len being truncated in the tail chunk.
> 
> 2. (Should be fine as FCFS is used by default) If you use Longest Prefix Match instead of FCFS, the order in which the scheduler fills up that chunk size is non-deterministic (truncate the last request). 
> 
> 
> 
> As for speculative decoding, the varied # of tokens to verify can also lead to non-deterministic tail (though not sure if ppl ever use spec dec in RL)
> 
> In these cases we have to disable split kv.
> 
> So we will need to stick to FCFS and avoid changing `new_token_ratio`.
> 
> 
> 
> cc @Fridge003 @merrymercy @zhyncs
> 
> 

For deterministic inference we can add some assumptions, like disabling radix cache, disabling speculative decoding, or even disabling chunked prefill. Part of the speed can be sacrificed for more stable output

Edenzzzz:
Prefill tests also pass
image

csrc/batch_prefill.cu
```diff
@@ -47,7 +47,8 @@ at::Tensor BatchPrefillWithKVCachePlan(
     at::Tensor page_locked_int_workspace_buffer, at::Tensor qo_indptr, at::Tensor kv_indptr,
     at::Tensor kv_len_arr, int64_t total_num_rows, int64_t batch_size, int64_t num_qo_heads,
     int64_t num_kv_heads, int64_t page_size, bool enable_cuda_graph, int64_t head_dim_qk,
-    int64_t head_dim_vo, bool causal, int64_t window_left) {
+    int64_t head_dim_vo, bool causal, int64_t window_left, int64_t fixed_split_size = -1,
```
yzh119: I don't encourage setting up default value in C++ APIs (as these APIs are directly exposed as python bindings), any use case of them?

tvm_binding/batch_prefill.cu
```diff
@@ -46,7 +46,8 @@ IntTuple BatchPrefillWithKVCachePlan(
     DLTensor* page_locked_int_workspace_buffer, DLTensor* qo_indptr, DLTensor* kv_indptr,
     IntTuple kv_len_arr, int64_t total_num_rows, int64_t batch_size, int64_t num_qo_heads,
     int64_t num_kv_heads, int64_t page_size, bool enable_cuda_graph, int64_t head_dim_qk,
-    int64_t head_dim_vo, bool causal, int64_t window_left, TVMStreamHandle cuda_stream) {
+    int64_t head_dim_vo, bool causal, int64_t window_left, int64_t fixed_split_size,
```
yzh119, line 49: For TVM bindings, @MasterJH5574 could probably help with checking them.

yzh119 (commented):
LGTM in general, left some minor suggestions.

csrc/batch_prefill.cu
Edenzzzz: just example default values for python interface writers, removing now

tvm_binding/batch_prefill.cu
MasterJH5574: Sorry I just notice this change.  On TVM side we don't use `fixed_split_size` at this moment, so will need to remove it from the parameter list and use a default value.  We will submit a fix later.

MasterJH5574: Updated in https://github.com/flashinfer-ai/flashinfer/pull/1680

PR #1688, opened 2025-09-16
gemm: Enabled alpha with the mx_fp4 format

nvmbreughe:
## 📌 Description

While the mx_fp4 format uses per-block scales only, some cases do benefit from having a global scale to avoid truncating the range. This PR enables this and makes the following changes:
* construct the cuDNN graph with a global scale if one is provided
* Adjust the mm_fp4 API with an explicit use_nvfp4 bool: previously the API could use the presence of an "alpha" factor to distinguish between nvfp4 (required global scale) and mx_fp4 formats. Now we have an explicit parameter to indicate this: use_nvfp4.

## 🔍 Related Issues

This change was decided after a feedback discussion with the requester of [issue #1592](https://github.com/flashinfer-ai/flashinfer/issues/1592).

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/flashinfer/gemm.py
+++ b/flashinfer/gemm.py
@@ -1652,6 +1652,7 @@ def build_cudnn_gemm_block_scale_dequantize_graph(
     o_type,
     block_size,
     device,
+    alpha,
     use_nvfp4,
 ):
     _check_cudnn_availability()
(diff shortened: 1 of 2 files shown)


yzh119:
Hi @nvmbreughe can you provide some background of this PR? (e.g. in the PR description).

nvmbreughe:
> Hi @nvmbreughe can you provide some background of this PR? (e.g. in the PR description).

Done. Thanks for reminding me!

PR #1695, opened 2025-09-16
[cute_dsl] add gemm + all reduce (two_shot) 

Amir-19:
## 📌 Description

this pr adds gemm overlapped with two-shot allreduce (with multimem instructions) 

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/flashinfer/cute_dsl/gemm_allreduce_two_shot.py
+++ b/flashinfer/cute_dsl/gemm_allreduce_two_shot.py
@@ -1,31 +1,24 @@
-import argparse
-import ctypes
-import os
-from math import prod
-from functools import partial
 from typing import Optional, Tuple, Type, Union
 
 import torch
-import torch.distributed._symmetric_memory as symm_mem
 import torch.distributed as dist
 from cuda import cuda
 
 import cutlass
 import cutlass.cute as cute
-import cutlass.cute.testing as testing
-import cutlass.torch as cutlass_torch
 import cutlass.utils as utils
 import cutlass.pipeline as pipeline
 import cutlass.utils.blackwell_helpers as sm100_utils
 import cutlass.utils.distributed_helpers as distributed_helpers
 from cutlass.cute.nvgpu import cpasync, tcgen05
-from cutlass.cute.runtime import from_dlpack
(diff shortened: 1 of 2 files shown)


flashinfer/cute_dsl/gemm_allreduce_two_shot.py
```diff
@@ -0,0 +1,2030 @@
...
+* The contiguous dimension of A/B/C tensors must be at least 16 bytes aligned,
+  i.e, number of elements is a multiple of 4, 8, and 16 for TFloat32,
+  Float16/BFloat16, and Int8/Uint8/Float8, respectively.
+* OOB tiles are not allowed when TMA store is disabled
+* when all_reduce is not "none", M and N must be multiple of 128, world_size must be 8
+
+"""
+
+
+class PersistentDenseGemmKernel:
```
yzh119, line 104: How many of them overlaps with https://github.com/flashinfer-ai/flashinfer/blob/4d4e8c05729a07fc4a04ee764d3186a9a6c718ec/flashinfer/cute_dsl/blockscaled_gemm.py#L462? Any chance of reusing some common components?

Amir-19: I can check, but these are all examples from cute_dsl and keeping them as close to the original implementation makes it easier to maintain and update.

yzh119: > but these are all examples from cute_dsl and keeping them as close to the original implementation makes it easier to maintain and update.

Yes but they will be diverged from original examples soon as we keep adding features upon them. Making them modular should still be important.

Amir-19: there are similarities between these two but if we want to extract the overlapping parts into separate common functions, there would be branching. also porting other kernels, would require further changes to these common files and the logic become unnecessarily complicated and hard to read. This gemm+ar kernel shares all the compute part with dense gemm persistent and not the existing blockscaled gemm in flashinfer.

yzh119 (approved):
LGTM

PR #1696, opened 2025-09-16
DRAFT: Support Kimi-K2 for TRT: templatize number of experts

GordonGustafson:
## Goal: Support Kimi-K2 Family of Models

The DeepSeek-v3 family of models use 256 experts, while the Kimi-K2 family of models use 384 experts. The number of experts is currently a hardcoded constant in the DeepSeek routing kernels, and must remain a compile-time constant in order to be used as the `__launchbounds__` for some kernels.

  ### Proposed Solution: Template-configurable Number of Experts

  In [include/flashinfer/trtllm/fused_moe/RoutingKernel.h](https://github.com/flashinfer-ai/flashinfer/blob/main/include/flashinfer/trtllm/fused_moe/RoutingKernel.h), change
  ```
  template <typename InputT_, typename OutputT_, bool UsePdl_>
  struct KernelParamsBase {
      ...
  }
  ```
  to
  ```
  template <typename InputT_, typename OutputT_, bool UsePdl_, int NumExperts_>
  struct KernelParamsBase {
      static constexpr bool NumExperts = NumExperts_;
      ...
  }
  ```
Then change [csrc/trtllm_fused_moe_routing_deepseek.cu](https://github.com/flashinfer-ai/flashinfer/blob/main/csrc/trtllm_fused_moe_routing_deepseek.cu), [csrc/trtllm_fused_moe_routing_renormalize.cu](https://github.com/flashinfer-ai/flashinfer/blob/main/csrc/trtllm_fused_moe_routing_renormalize.cu), [csrc/trtllm_fused_moe_routing_llama4.cu](https://github.com/flashinfer-ai/flashinfer/blob/main/csrc/trtllm_fused_moe_routing_llama4.cu), to use `KernelParams::NumExperts` as the number of experts (might need some boilerplate to "pass" it from `KernelParamsBase` to `KernelParams` as well).

  ### Benefits
  - Easy to support different numbers of experts in the future.
  - Zero runtime overhead.

  ### Costs
  - (NEW) The numbers of experts for Llama4 and `renormalize` kernels need to be known at compile time, where previously only the *max* number of experts was known at compile time. This means that `flashinfer` users passing custom numbers of experts to these kernels will have their code broken if we don't include that number of experts in the `run` logic dispatching. If this is a concern we *might* be able to keep the number of experts defined at compile time in the deepseek kernel and at runtime in the Llama4 and renormalize kernels, but this would come at the cost of some complexity.

  ### Alternatives Considered

  1. New file called `csrc/trtllm_fused_moe_routing_kimi.cu` that is an exact copy of [csrc/trtllm_fused_moe_routing_deepseek.cu](https://github.com/flashinfer-ai/flashinfer/blob/main/csrc/trtllm_fused_moe_routing_deepseek.cu) except it hardcodes 384 experts. This makes maintenance of both files more difficult, and doesn't make it easy to support different numbers of experts in the future.

  2. Make [csrc/trtllm_fused_moe_routing_deepseek.cu](https://github.com/flashinfer-ai/flashinfer/blob/main/csrc/trtllm_fused_moe_routing_deepseek.cu) accept a compiler flag defining the number of experts (`-DNUM_EXPERTS 256` and `-DNUM_EXPERTS 384`). This moves complexity to the build stage, which doesn't seem like the right place to manage it.

### DIFF
--- a/csrc/trtllm_fused_moe_routing_renormalize.cu
+++ b/csrc/trtllm_fused_moe_routing_renormalize.cu
@@ -210,8 +210,8 @@ __global__ void __launch_bounds__(NumThreadsHist)
 
     routingTopKExperts<BaseType, InputT, VecSize, KernelParams::DoSoftmaxBeforeTopK>(
         warp, allScores, allExpertIdx, warpTopKScore, warpTopKExpertIdx, laneIdx,
-        KernelParams::NumExperts, params.mTopK, params.mPtrScores + scoreOffset, params.mNormTopkProb,
-        params.mApplySoftmaxAfterTopK);
+        KernelParams::NumExperts, params.mTopK, params.mPtrScores + scoreOffset,
+        params.mNormTopkProb, params.mApplySoftmaxAfterTopK);
 
     if (laneIdx < params.mTopK) {
       PackedScoreIdx<OutputT> packedScore{static_cast<OutputT>(warpTopKScore[laneIdx]),
@@ -226,9 +226,8 @@ __global__ void __launch_bounds__(NumThreadsHist)
 template <int NumExperts>
 void runImpl(Data const& data, void* stream) {
   // Validate that the runtime value matches the template parameter
-  TORCH_CHECK(data.mNumExperts == NumExperts,
-              "Renormalize routing kernel expects #experts ", data.mNumExperts,
-              " to match template parameter ", NumExperts);
+  TORCH_CHECK(data.mNumExperts == NumExperts, "Renormalize routing kernel expects #experts ",
+              data.mNumExperts, " to match template parameter ", NumExperts);
 
   TORCH_CHECK(data.mPtrExpertIdx != nullptr || data.mPtrScores != nullptr,
               "Routing kernel requires at least one input parameter");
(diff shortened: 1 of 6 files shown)


csrc/trtllm_fused_moe_routing_renormalize.cu
```diff
@@ -271,24 +276,58 @@ void run(Data const& data, void* stream) {
...
+  TORCH_CHECK(data.mPtrPermutedIdxSize != nullptr && data.mPtrCtaIdxXyToBatchIdx != nullptr &&
+                  data.mPtrCtaIdxXyToMnLimit != nullptr && data.mPtrNumNonExitingCtas != nullptr,
+              "Llama4 routing kernel expects permuted idx and grouped Gemm launch config buffers");
+  TORCH_CHECK(data.mTopK <= MaxNumTopExperts,
+              "Routing kernel expects topK experts <= ", MaxNumTopExperts, ", got ", data.mTopK);
+  TORCH_CHECK(data.mPaddingLog2 < 8, "Routing kernel expects padding log2 < 8, got ",
+              data.mPaddingLog2);
+
+  // Dispatch to the appropriate template instantiation based on the number of experts
+  switch (data.mNumExperts) {
```
GordonGustafson, line 312: I'm not familiar enough with the all the models to known which values to include here. Please let me know if you have suggestions! As I mentioned in the `Costs` section of the description, it might be possible to keep the number of experts defined at runtime for this kernel and at compile-time in the DeepSeek kernel, but this would come at the cost of complexity.

yzh119: Let's keep this as is until we reach a case not captured here.

yzh119 (approved):
LGTM

csrc/trtllm_fused_moe_routing_renormalize.cu
yzh119: Blackwell UTs failed with error:
```
RuntimeError: Unsupported number of experts: 72
```

Seems we should also add number of experts for 384 experts under different TP/EP settings.

GordonGustafson: Addressed, I added `384/{1,2,4,8,16}`, `256`, and `72`.

GordonGustafson:
~Please don't merge yet, I'm still doing some testing~

EDIT: testing is done, tried it on Kimi-K2 and DeepSeek for both fp4 and fp8 quantization.

zhyncs:
Hi @yzh119 After this PR, may we release a new version? Thanks!

zhyncs:
All passed @yzh119

PR #1710, opened 2025-09-17
test: skip the unsupported test cases for sm120/121

yongwww:
## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Reviewer Notes

cc: @yzh119

### DIFF
--- a/tests/test_batch_attention.py
+++ b/tests/test_batch_attention.py
@@ -23,6 +23,7 @@ from jit_utils import (
     gen_persistent_batch_attention_modules,
     gen_prefill_attention_modules,
 )
+from flashinfer.utils import get_compute_capability
 
 
 @pytest.fixture(autouse=True, scope="module")
@@ -211,6 +212,10 @@ def test_batch_attention_correctness(
     test_dtype,
     logits_soft_cap,
 ):
+    compute_capability = get_compute_capability(torch.device(device="cuda"))
+    # TODO(P1): Most of these tests pass on SM120 and SM121 GPUs. We need triage these at some point.
+    if compute_capability[0] == 12:
+        pytest.skip("These tests are only guaranteed to work on SM100 and SM103 GPUs.")
     num_qo_heads = num_kv_heads * gqa_group_size
     kv_lens = [p[0] for p in seq_len_pairs]
     qo_lens = [p[1] for p in seq_len_pairs]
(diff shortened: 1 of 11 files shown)


tests/test_batch_attention.py
```diff
@@ -211,6 +212,10 @@ def test_batch_attention_correctness(
     test_dtype,
     logits_soft_cap,
 ):
+    compute_capability = get_compute_capability(torch.device(device="cuda"))
+    # TODO(P1): Most of these tests pass on SM120 and SM121 GPUs. We need triage these at some point.
+    if compute_capability[0] == 12:
+        pytest.skip("These tests are only guaranteed to work on SM100 and SM103 GPUs.")
```
yzh119: Firstly, this kernel should work for all architectures (sm_80, sm_90, ...). The reason it fails on sm120 is because the tile size/number of stages is too large for sm_120 (and we should fix the behavior later), I encourage xfail instead (and add some explainations).

tests/test_batch_attention.py
```diff
@@ -238,6 +243,7 @@ def test_batch_attention_correctness(
         num_kv_heads=4,
         gqa_group_size=7,
         head_dim=128,
+        v_scale=2.0,
```
yzh119: why changing the v_scale here?

tests/test_xqa.py
```diff
@@ -167,6 +168,9 @@ def test_xqa(
     use_attention_sinks,
     use_sliding_window,
 ):
+    compute_capability = get_compute_capability(torch.device(device="cuda"))
+    if compute_capability[0] != 9:
+        pytest.skip("These tests are only guaranteed to work on Hopper.")
```
yzh119:
```suggestion
        pytest.skip("XQA only supports on Hopper at this moment")
```

tests/test_batch_attention.py
yongwww: will ran into `TypeError: test_batch_attention_correctness() missing 1 required positional argument: 'v_scale'` if tried to run `python tests/test_batch_attention.py`. Probably i should not touch this in __main__ since pytest won't trigger it

tests/test_batch_attention.py
yongwww: resolved in https://github.com/flashinfer-ai/flashinfer/commit/b4129bd4a136ba397ac714f47ce56f6c91954422

PR #1721, opened 2025-09-18
fix the dequantize_block in the trtllm_cutlass fuse moe test

rainj-me:
## 📌 Description

The dequantize_block logic in test_trtllm_cutlass_fused_moe.py file is not correct. The test case is not failed due to the HIDDEN_SIZES intended set to 128, which makes the tensor multiply succeed like

```
x_quant.shape = torch.Size([1, 128])
scales.shape = torch.Size([1, 1, 128])
```

However, if we change the HIDDEN_SIZES to 256 and keep block size to 128, the tensor multiply would failed, since now the 

```
x_quant.shape = torch.Size([1, 256])
scales.shape = torch.Size([1, 2, 128])
```

With the logic in this PR, the scales will reshape to (1, 256) which is the same shape as x_quant, now the tensor multiply would succeed

```
x_quant.shape = torch.Size([1, 256])
scales.shape = torch.Size([1, 256])
```

The logic also works with batch size > 1

## 🔍 Related Issues

N/A

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/tests/test_trtllm_cutlass_fused_moe.py
+++ b/tests/test_trtllm_cutlass_fused_moe.py
@@ -215,7 +215,7 @@ BATCH_SIZES = [
     1,
 ]
 HIDDEN_SIZES = [
-    128,
+    256,
 ]
 NUM_EXPERTS = [2]
 TOP_K_VALUES = [2]
@@ -884,6 +884,7 @@ def dequantize_block(
     scales: torch.Tensor,
     dtype: torch.dtype,
     original_shape: tuple,
+    block_size_n: int = 128,
 ) -> torch.Tensor:
     """
     Dequantize a block-quantized tensor.
@@ -904,8 +905,8 @@ def dequantize_block(
         if dim != -1:
             a = a.transpose(dim, -1)
         # Broadcast and reshape
-        a_broadcasted = a.unsqueeze(-1).expand(*a.shape, 128)
-        a_reshaped = a_broadcasted.reshape(*a.shape[:-1], a.shape[-1] * 128)
+        a_broadcasted = a.unsqueeze(-1).expand(*a.shape, block_size_n)
+        a_reshaped = a_broadcasted.reshape(*a.shape[:-1], a.shape[-1] * block_size_n)
         # Move back if needed
         if dim != -1:
             a_reshaped = a_reshaped.transpose(dim, -1)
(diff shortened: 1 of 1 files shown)


rainj-me:
Without this PR, the failure log for the test_trtllm_cutlass_fused_moe.py if we change the HIDDEN_SIZES to 256 is 

```
>       x_dequant = x_quant.to(dtype) * scales.to(dtype)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E       RuntimeError: The size of tensor a (256) must match the size of tensor b (128) at non-singleton dimension 2

tests/test_trtllm_cutlass_fused_moe.py:932: RuntimeError
```

tests/test_trtllm_cutlass_fused_moe.py
```diff
@@ -924,10 +930,10 @@ def transform_dim(a: torch.Tensor, dim: int = -1) -> torch.Tensor:
...
 
         # Handle padding
-        if in_dim % 128 != 0:
-            scales = scales[..., : in_dim % 128, :]
-        if out_dim % 128 != 0:
-            scales = scales[..., :, : out_dim % 128]
+        if in_dim % block_size_n != 0:
+            scales = scales[..., : in_dim % block_size_n, :]
+        if out_dim % block_size_n != 0:
+            scales = scales[..., :, : out_dim % block_size_n]
```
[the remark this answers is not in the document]
rainj-me: the in_dim and out_dim in the scales per block.

tests/test_trtllm_cutlass_fused_moe.py
```diff
@@ -215,7 +215,7 @@ def compute_with_experts(
     1,
 ]
 HIDDEN_SIZES = [
-    128,
+    256,
```
wenscarl, line 218: Thanks for the catch. Could you please add another case that hidden_size is not divisible of 128 to trigger the padding logic?

wenscarl (approved):

tests/test_trtllm_cutlass_fused_moe.py
rainj-me: Sure, will fix it.

rainj-me: @wenscarl , I believe the padding logic and the per block quantization logic also not correct.

PR #1723, opened 2025-09-19
Fix DeepSeek quality for TRTLLM fused MoE routing

GordonGustafson:
In our tests DeepSeek-V3 model quality was severely degraded when using `MaxNumTopGroups = 16`. This PR makes it so we use `MaxNumTopGroups = 16` for Kimi-K2 and `MaxNumTopGroups = 4` for DeepSeek.

### DIFF
--- a/csrc/trtllm_fused_moe_routing_deepseek.cu
+++ b/csrc/trtllm_fused_moe_routing_deepseek.cu
@@ -24,7 +24,14 @@ namespace routingDeepSeek {
 
 static constexpr int NumTopGroupScores = 2;
 static constexpr int MaxNumTopExperts = 8;
-static constexpr int MaxNumTopGroups = 16;
+
+__host__ __device__ int constexpr getMaxNumTopGroups(const bool useGroups, const int numExperts) {
+  if (useGroups || numExperts <= 256) {
+    return 4;
+  } else {
+    return 16;
+  }
+}
 
 template <typename KernelParams>
 __global__ void routingMainKernel(KernelParams params) {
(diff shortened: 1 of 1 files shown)


zhyncs (approved):

yzh119 (commented):
> In our tests DeepSeek-V3 model quality was severely degraded when using MaxNumTopGroups = 16

`MaxNumTopGroups` will determine register usage thus influencing performance.

Currently the runtime (max) limit is 4, https://github.com/flashinfer-ai/flashinfer/blob/ea322abeb17c8ad0247e6486347ed618bab94388/csrc/trtllm_fused_moe_runner.cu#L61, in what case would we reach 16?

GordonGustafson:
@yzh119 We have this check:
```
    TORCH_CHECK(data.mNumExperts <= WarpSize * MaxNumTopGroups, "Routing kernel expects #experts ",
                data.mNumExperts, " <= WarpSize * MaxNumTopGroups ", WarpSize * MaxNumTopGroups);
```
Kimi's number of experts is 384, so MaxNumTopGroups for Kimi needs to be at least 12 (32 * 12 = 384). If we have a preference for 12 instead of 16 I could test that.

yzh119:
My question is, which `routingMethodType` do you expect to use for Kimi-K2? If you reuse `RoutingMethodType::DeepSeekV3` then you will run into the check defined in https://github.com/flashinfer-ai/flashinfer/blob/ea322abeb17c8ad0247e6486347ed618bab94388/csrc/trtllm_fused_moe_runner.cu#L61. So I guess we will use other routing method?

Also, would you mind checking https://github.com/flashinfer-ai/flashinfer/pull/1724? Our internal CI failed after https://github.com/flashinfer-ai/flashinfer/pull/1696

PR #1755, opened 2025-09-22
Fix tests/test_trtllm_gen_attention.py::test_trtllm_batch_prefill, ::test_trtllm_batch_decode mismatch error

kahyunnam:
## 📌 Description

Fix failing tests in` tests/test_trtllm_gen_attention.py::test_trtllm_batch_prefill ` and `tests/test_trtllm_gen_attention.py::test_trtllm_batch_decode` due to `AssertionError: Tensor-likes are not close!`. An example error message was (all the failing tests had similar error messages): 

```
E       AssertionError: Tensor-likes are not close!
E
E       Mismatched elements: 1 / 169269760 (0.0%)
E       Greatest absolute difference: 2.0 at index (60221, 1, 77) (up to 1.0 allowed)
E       Greatest relative difference: 1.0 at index (60221, 1, 77) (up to 0.4 allowed)
```

which suggests that this is not a functional bug, but the testing could be relaxed to allow for a small number of mismatches (in this case, 1 out of 169 million elements), especially since the abs/rel difference is not wildly out of range (2 vs 1, 1 vs 0.4) and the failing tests are low precision types (fp8-fp8-nvfp4)

This change catches the assertion error and allows a small mismatch rate (1e-7).

## 🧪 Test results

(Tested on B300)

```
pytest tests/test_trtllm_gen_attention.py::test_trtllm_batch_decode
================================================== 11664 passed, 1296 skipped in 582.92s (0:09:42) ===================================================
```
```
 tests/test_trtllm_gen_attention.py::test_trtllm_batch_prefill 
================================================================ 972 passed in 46.09s ================================================================
```

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

- [x ] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [ x] I have installed the hooks with `pre-commit install`.
- [ x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

## 🧪 Tests

- [ x] Tests have been added or updated as needed.
- [ x] All tests are passing (`unittest`, etc.).

### DIFF
--- a/tests/test_trtllm_gen_attention.py
+++ b/tests/test_trtllm_gen_attention.py
@@ -400,8 +400,8 @@ def test_trtllm_batch_prefill(
 
     allowed_mismatch_rate = 1e-7
 
-    # convert to float32 for fp8 is not supported by assert_close
     try:
+        # convert to float32 for fp8 is not supported by assert_close
         torch.testing.assert_close(
             output.float() * o_scale, output_ref.float(), rtol=rtol, atol=atol
         )
--- a/tests/test_trtllm_gen_attention.py
+++ b/tests/test_trtllm_gen_attention.py
@@ -398,6 +398,7 @@ def test_trtllm_batch_prefill(
     else:
         rtol, atol = 1e-2, 1e-2
 
+    # Arbitary small mismatch rate
     allowed_mismatch_rate = 1e-7
 
     try:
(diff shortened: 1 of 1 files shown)


tests/test_trtllm_gen_attention.py
```diff
@@ -622,12 +644,34 @@ def test_trtllm_batch_decode(
...
+    allowed_mismatch_rate = 5e-5
+
+    try:
+        torch.testing.assert_close(
+            output.float() * o_scale,
+            output_ref.float(),
+            rtol=rtol,
+            atol=atol,
+        )
+    except AssertionError as err:
```
nvmbreughe: [nit] Should we create a utils function since this logic is being used at least twice?

tests/test_trtllm_gen_attention.py
```diff
@@ -398,10 +398,32 @@ def test_trtllm_batch_prefill(
...
+    # Arbitary small mismatch rate
+    allowed_mismatch_rate = 1e-7
+
+    try:
+        # convert to float32 for fp8 is not supported by assert_close
+        torch.testing.assert_close(
+            output.float() * o_scale, output_ref.float(), rtol=rtol, atol=atol
+        )
+    except AssertionError as err:
+        # Allow for a small mismatch rate for low precision outputs
```
yzh119: Can we summarize a function for this kind of check (it will be useful especially for low-precision kernels), such as `flashinfer.testing.assert_close` with an parameter of `max_failed_ratio`

tests/test_trtllm_gen_attention.py
nvmbreughe: Seems like I duplicated Zihao's comment :).

tests/test_trtllm_gen_attention.py
kahyunnam: I started to refactor a shared function, but then I saw `assert_close_with_mismatch_tolerance`  is already implemented/imported  from `conftest` and used in this file .... so I changed the two tests to use that instead!

tests/test_trtllm_gen_attention.py
kahyunnam: (Commented on Zihao's comment as well but) I updated to re-use `assert_close_with_mismatch_tolerance` from conftest

nvmbreughe: excellent

PR #1759, opened 2025-09-23
fix(cleanup): ensure repository URL has no trailing slash

tarukumar:
## 📌 Description
Remove trailing slash from FLASHINFER_CUBINS_REPOSITORY before concatenating the filename, to avoid redundant double slashes in the URL. Even though servers may handle double slashes, this makes URL construction consistent and cleaner.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/flashinfer/jit/cubin_loader.py
+++ b/flashinfer/jit/cubin_loader.py
@@ -155,7 +155,9 @@ def get_cubin(name, sha256, file_extension=".cubin", session=None):
     if cubin:
         return cubin
     # either the file does not exist or it is corrupted, we'll download a new one.
-    uri = Path(FLASHINFER_CUBINS_REPOSITORY) / cubin_fname
+
+    base = FLASHINFER_CUBINS_REPOSITORY.rstrip("/")
+    uri = urljoin(base + "/", cubin_fname)
     logger.info(f"Fetching cubin {name} from {uri}")
     download_file(uri, cubin_path, session=session)
     return load_cubin(cubin_path, sha256)


flashinfer/jit/cubin_loader.py
```diff
@@ -154,7 +154,7 @@ def get_cubin(name, sha256, file_extension=".cubin", session=None):
     if cubin:
         return cubin
     # either the file does not exist or it is corrupted, we'll download a new one.
-    uri = FLASHINFER_CUBINS_REPOSITORY + "/" + cubin_fname
+    uri = f"{FLASHINFER_CUBINS_REPOSITORY.rstrip('/')}/{cubin_fname}"
```
yzh119: Can we use pathlib instead? which is more robust than string concatenations

flashinfer/artifacts.py
```diff
@@ -107,9 +108,7 @@ def get_cubin_file_list():
     ]:
         cubin_files += [
             (kernel + name, extension)
-            for name, extension in get_available_cubin_files(
-                FLASHINFER_CUBINS_REPOSITORY + "/" + kernel
-            )
+            for name, extension in get_available_cubin_files(base + "/" + kernel)
```
yzh119: ditto

flashinfer/jit/cubin_loader.py
tarukumar: Pathlib is not a good solution for handling URIs. I tried it just for the sake of it, but I’ve switched to using urllib, which seems more appropriate.

yzh119 (approved):

PR #1769, opened 2025-09-25
feat: add xqa fp8 mha and fp8 kv cache

qsang-nv:
## 📌 Description

Add xqa fp8 mha and fp8 kv cache. Add fp8 mla for sm120. Use vllm kv layout.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * MLA-based attention path and dedicated MLA entrypoints (SM120/121)
  * FP8 KV-cache support with optional paged KV layout and separate K/V cache inputs
  * Asynchronous tensor-map/TMA and matrix-descriptor primitives for high-throughput GPU transfers
  * Dtype-driven config and expanded GPU SM gating for builds/runtimes

* **Bug Fixes**
  * Improved numerical stability for attention mask initialization

* **Tests**
  * Expanded coverage for MLA, FP8, FP16/BF16, and new cache layouts

* **Documentation**
  * Added XQA API docs and new public symbols

### DIFF
--- a/csrc/xqa/utils.cuh
+++ b/csrc/xqa/utils.cuh
@@ -46,7 +46,7 @@ __constant__ constexpr float kE4M3_MAX = 448.F;
 constexpr uint32_t kMAX_SMEM_SIZE = (99u << 10);
 #elif __CUDA_ARCH__ == 800 || __CUDA_ARCH__ == 870
 constexpr uint32_t kMAX_SMEM_SIZE = (163u << 10);
-#elif __CUDA_ARCH__ == 900 || __CUDA_ARCH__ == 1000
+#elif __CUDA_ARCH__ == 900 || __CUDA_ARCH__ == 1000 || __CUDA_ARCH__ == 1030
 constexpr uint32_t kMAX_SMEM_SIZE = (227u << 10);
 #endif
 #endif
--- a/flashinfer/aot.py
+++ b/flashinfer/aot.py
@@ -356,7 +356,7 @@ def gen_attention(
 
 
 def gen_xqa(
-    fp16_input_: List[bool],
+    input_type_: List[torch.dtype],
     fp8_kv_cache_: List[bool],
     token_per_page_: List[int],
     head_size_: List[int],
@@ -371,32 +371,20 @@ def gen_xqa(
     if not has_sm90 and not has_sm100 and not has_sm120 and not has_sm121:
         return  # XQA requires SM90+
 
-    sm_versions = []
-    if has_sm90:
-        sm_versions.append(90)
-    if has_sm100:
-        sm_versions.append(100)
-    if has_sm120:
-        sm_versions.append(120)
-    if has_sm121:
-        sm_versions.append(121)
-
     for (
-        fp16_input,
+        input_type,
         fp8_kv_cache,
         token_per_page,
         head_size,
         head_grp_size,
         use_sliding_window,
-        sm_version,
     ) in product(
-        fp16_input_,
+        input_type_,
         fp8_kv_cache_,
         token_per_page_,
         head_size_,
         head_grp_size_,
         use_sliding_window_,
-        sm_versions,
     ):
         # Skip invalid configurations
         if head_size % 16 != 0 or head_size > 256 or head_size < 16:
@@ -406,21 +394,19 @@ def gen_xqa(
 
         if fp8_kv_cache:
             kv_cache_dtype = torch.float8_e4m3fn
-        elif fp16_input:
-            kv_cache_dtype = torch.float16
         else:
-            kv_cache_dtype = torch.bfloat16
+            kv_cache_dtype = input_type
+
         yield gen_xqa_module(
-            input_dtype=torch.float16 if fp16_input else torch.bfloat16,
+            input_dtype=input_type,
             kv_cache_dtype=kv_cache_dtype,
             page_size=token_per_page,
             head_dim=head_size,
             head_group_ratio=head_grp_size,
             use_sliding_window=use_sliding_window,
-            sm_version=sm_version,
         )
 
-    if has_sm120:
+    if has_sm120 or has_sm121:
         for token_per_page in token_per_page_:
             yield gen_xqa_module_mla(
                 input_dtype=torch.float8_e4m3fn,
@@ -429,19 +415,6 @@ def gen_xqa(
                 head_dim=576,
                 head_group_ratio=128,
                 use_sliding_window=False,
-                sm_version=120,
-            )
-
-    if has_sm121:
-        for token_per_page in token_per_page_:
-            yield gen_xqa_module_mla(
-                input_dtype=torch.float8_e4m3fn,
-                kv_cache_dtype=torch.float8_e4m3fn,
-                page_size=token_per_page,
-                head_dim=576,
-                head_group_ratio=128,
-                use_sliding_window=False,
-                sm_version=121,
             )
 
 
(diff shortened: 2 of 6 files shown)


csrc/flashinfer_xqa_binding.cu
```diff
@@ -16,8 +16,8 @@
 
 #include "pytorch_extension_utils.h"
 
-void xqa_wrapper(int64_t multiProcessorCount, int64_t nbKHeads, int64_t slidingWinSize,
-                 double qScale, at::Tensor output,
+void xqa_wrapper(bool run_fp8_mha, int64_t multiProcessorCount, int64_t nbKHeads,
```
jdebache: Instead of making this a flag, could we pass a dtype?

Same for the other places where we pass:

- the type of the input (only bf16 and fp16 supported I think)
- the type of the kv-cache (fp8 or bf16)
- the type in which we perform arithmetic (the same type as the kv-cache I think?)

csrc/xqa/utils.cuh
```diff
@@ -31,7 +31,13 @@
 #include "barriers.cuh"
 
 inline constexpr float log2e = 1.4426950408889634;  // std::log2(M_E)
-inline constexpr float safeInitRowMax = -1e+30F;
+// we used an optimization where exp(x-rowMax) is computed as:
```
jdebache, line 34: That's interesting: what were the symptoms of the instability? Accuracy loss?

tests/attention/test_xqa.py
```diff
@@ -354,4 +364,21 @@ def cache_head_at(
         kernel_output = output[req][b][
             idx_k_head * head_grp_size : (idx_k_head + 1) * head_grp_size
         ].to(torch.float32)
-        assert torch.allclose(ref_output, kernel_output, atol=0.01, rtol=0.01)
+        if fp8_kv_cache or run_fp8_mha:
+            atol = 0.05
```
jdebache: How did you tune this tolerance? Can it be smaller?

qsang-nv: From 0.01 to 0.05, add 0.01 every step. And it can't be smaller from my test.

csrc/xqa/utils.cuh
qsang-nv: This is copied from https://github.com/NVIDIA/TensorRT-LLM/commit/c1aa7f31d966c6868fb4224b46e1e8652d9ee0e9, you may ask the author, I am not sure about this question.

flashinfer/jit/xqa.py
```diff
@@ -58,18 +71,28 @@ def gen_xqa_module(
     else:
         flag_sliding_window = ["-DSLIDING_WINDOW=0"]
 
+    if sm_version == 100:
```
sricketts: Is it possible to add SM103 support by targeting SM100f instead of SM100a?

And similarly, can we add SM121 support by targeting SM120f instead of SM120a?

qsang-nv: What's the difference in those archs? I mean SM103/SM100f/SM100a, and SM121/SM120f/SM120a.

sricketts: The "a" means "arch-specific" and the "f" means "family".

SM100 and SM103 are in the "SM100f family". 

SM100a (SM100 arch-specific) will only run on SM100 devices. 

But I believe SM103 devices have a superset of the SM100 features, and therefore if you target SM100f instead of SM100a during compilation, your cubin will be able to run on SM103 as well, without any loss of optimization on either device. So I think it's strictly better than targeting SM103a.

SM121 and SM120 have a similar story: it's better to target SM120f as a compilation target, yielding a cubin that will run on both SM120 and SM121 devices without any compromise to performance.

See this documentation for details: https://docs.nvidia.com/cuda/cuda-c-programming-guide/#family-specific-features

@aleozlx can you confirm my understanding?

aleozlx: yes sm_100f is known as family specific or family conditional. this is important to enhance device compatibility in a sort of out of the box fashion.

i'd like to point out a few things tho based on my experience:
1. with strictly jit compilation where the premise is that at runtime fewer devices are in the compatibility question, the arch conditionals may be the safest way to target the instruction supersets. (when the target is available to compile by the toolkit at the time of implementation)
2. family conditionals are important for compatibility story (indeed aligning with your understanding conceptually) but it is not without inherent engineering complexity. from an engineer's perspective i naturally experience a slightly more complicated story at the levels beneath. i'll spare the details but leave it as a reliability intuition (in the context of jit).

however, i want to bring this [CompilationContext.get_nvcc_flags_list()](https://github.com/flashinfer-ai/flashinfer/blob/bd98dacf67b78d81eed48fab2592b2bb48d51bcd/flashinfer/compilation_context.py#L50
) up for consideration to not hard code any guidance either way but have it abstracted so we can adjust if situation changes. briefly, how this works is for each op/backend if you whitelist your supported targets, this function shall serve as the mapping to provide the recommended flags. 

we can put in sm103 support (likely fine for attn) supposing our cicd will catch the issue if not

sricketts: Makes sense, let me clarify my guidance then:

1. Since SM120 and SM121 are identical architectures, naively I would think that supporting both is only marginally harder than supporting one.
2. The story seems even a little harder for SM100 and SM103, because they are not identical, but if you don't care about SM103-specific features, I would think the marginal effort of supporting SM103 shouldn't be massive.
3. Therefore the default design for any solution for SM100 and SM120 should at least try to include SM103 and SM121 support, or should at least be designed with SM103 and SM121 in mind, even if some of the details are left as a future TODO.
4. What compilation targets you use, and how you architect the code to query for those compilation targets, is an engineering implementation detail, and I probably shouldn't be opinionated about that. :) 

The problem here is that the PR doesn't address (3), maybe because @qsang-nv didn't know about (1) and (2). I'm only proposing that we re-think the design here with the above in mind.

@aleozlx @qsang-nv do you agree?

csrc/flashinfer_xqa_binding.cu
qsang-nv: Now it is passing dtype in flashinfer/flashinfer/xqa.py

flashinfer/jit/xqa.py
qsang-nv: Thanks for the detailed explanation! I've added support for sm100f and sm121a.

tests/attention/test_xqa.py
```diff
@@ -166,26 +182,23 @@ def test_xqa(
     nb_k_heads,
     seq_len,
     tokens_per_page,
-    use_fp16,
+    fp16_input,
```
nvmbreughe: Can we make this parameter more explicit? like `input_data_type`?

flashinfer/aot.py
```diff
@@ -356,44 +356,94 @@ def gen_attention(
 
 
 def gen_xqa(
-    use_fp16_: List[bool],
+    fp16_input_: List[bool],
+    fp8_kv_cache_: List[bool],
```
nvmbreughe, line 360: same here, maybe we can make the data types explicit instead of a combination of bools.
It seems like the options are:

-             kv_cache_dtype = torch.float8_e4m3fn
-             kv_cache_dtype = torch.float16
-             kv_cache_dtype = torch.bfloat16

yzh119:
Hi @qsang-nv xqa UT failed on https://gitlab-master.nvidia.com/dl/flashinfer/flashinfer-ci/-/jobs/225084624, would you mind taking a look?

flashinfer/jit/xqa.py
yzh119: We should rely on compilation context instead: https://github.com/flashinfer-ai/flashinfer/blob/d4a3ff4356aeaeaa2e67a8b176d72a749d96a089/flashinfer/compilation_context.py#L50-L52

tests/attention/test_xqa.py
qsang-nv: Done.

flashinfer/aot.py
qsang-nv: The kv cache option remains to be True/False as kv cache must be fp8 or consistent with input data type, combination such as (input fp16, kv cache bf16) is invalid. Input now use explicit data types.

PR #1771, opened 2025-09-25
Waive / disable test_mla_decode_kernel.py::test_mla_decode_kernel for not sm80 

kahyunnam:
## 📌 Description

test_mla_decode_kernel.py::test_mla_decode_kernel uses use_tensor_cores=false, which calls the internal sm80 MLA decode kernel, which doesn't support B40 chips (sm120).

This change adds checks to the test to skip when running on B40, and in the code before calling the internal sm80 MLA decode kernel to return a better error message. 

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

- [ x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [ x] I have installed the hooks with `pre-commit install`.
- [ x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

## 🧪 Tests

- [ x] Tests have been added or updated as needed.
- [x ] All tests are passing (`unittest`, etc.).

### DIFF
--- a/flashinfer/decode.py
+++ b/flashinfer/decode.py
@@ -61,6 +61,8 @@ from .utils import (
     register_fake_op,
     ceil_div,
     round_up,
+    get_compute_capability,
+    GPUArchitectureError,
 )
 
 
@@ -1755,6 +1757,15 @@ class BatchDecodeMlaWithPagedKVCacheWrapper:
             * attention output, shape: ``[batch_size, num_qo_heads, head_dim]``
             * logsumexp of attention scores, shape: ``[batch_size, num_qo_heads]``.
         """
+
+        # MLA decode kernel supports SM80 only
+        major, minor = get_compute_capability(q_nope.device)
+        device_arch = major * 10 + minor
+        if device_arch != 80:
+            raise GPUArchitectureError(
+                f"MLA decode kernel is not supported on this GPU (SM{device_arch}). "
+                "Supported architecture: SM80."
+            )
         window_left = self._window_left
         logits_soft_cap = self._logits_soft_cap
         sm_scale = self._sm_scale
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -146,6 +146,17 @@ def get_device_properties(device: torch.device):
     return torch.cuda.get_device_properties(device)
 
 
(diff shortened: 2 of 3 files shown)


flashinfer/decode.py
```diff
@@ -1755,6 +1756,15 @@ def run(
             * attention output, shape: ``[batch_size, num_qo_heads, head_dim]``
             * logsumexp of attention scores, shape: ``[batch_size, num_qo_heads]``.
         """
+
+        # MLA decode kernel supports up to SM80 only
+        major, minor = get_compute_capability(q_nope.device)
+        device_arch = major * 10 + minor
+        if device_arch != 80:
+            raise ValueError(
```
nvmbreughe: Instead of a value error, could we throw a flashinfer.utils.GPUArchitectureError?

nvmbreughe: Take a look at e.g., test_mm_fp4. We can even use it to pytest.xfail/pytest.skip.

flashinfer/decode.py
```diff
@@ -1755,6 +1756,15 @@ def run(
             * attention output, shape: ``[batch_size, num_qo_heads, head_dim]``
             * logsumexp of attention scores, shape: ``[batch_size, num_qo_heads]``.
         """
+
+        # MLA decode kernel supports up to SM80 only
```
sricketts: nit: do you mean "supports SM80 only"? 

"supports up to SM80" implies that it supports device_arch <= 80, which is not what the logic below is checking for.

kahyunnam: yes, updated to "`# MLA decode kernel supports SM80 only`"

flashinfer/decode.py
kahyunnam: Updated to raise GPUArchitectureError; 

> Take a look at e.g., test_mm_fp4. We can even use it to pytest.xfail/pytest.skip.

This kind of seems to do the same thing I had originally? (manual if compute_capability check, then pytest.skip)

But I do get the gist of what you're saying here. I added a functools test decorator that specifically catches the GPUArchitectureError to skip a test in the latest revision

yzh119 (approved):

PR #1779, opened 2025-09-26
bugfix: remove the filelock cleanup logic in cubin_loader.py

yzh119:
## 📌 Description

This PR removes cleanup of filelock files after downloading cubins, it may cause race conditions in concurrent environments where multiple processes might be downloading the same cubin simultaneously.

The filelock library already has robust internal mechanisms to handle lock cleanup. By removing the manual cleanup in  the finally block, we allow filelock to manage its own lock files correctly. This prevents race conditions where one process might delete a lock file while another process is still using it, especially in concurrent download scenarios.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Reviewer Notes

cc @nvpohanh @joker-eph @wenscarl

### DIFF
--- a/flashinfer/jit/cubin_loader.py
+++ b/flashinfer/jit/cubin_loader.py
@@ -106,12 +106,6 @@ def download_file(
         )
         return False
 
-    finally:
-        # Clean up the lock file
-        if os.path.exists(lock_path):
-            os.remove(lock_path)
-            logger.info(f"Lock file {lock_path} removed.")
-
 
 def load_cubin(cubin_path, sha256) -> bytes:
     """


joker-eph:
> The filelock library already has robust internal mechanisms to handle lock cleanup.

Do you have reference to this? I believe I did some research before adding this and this was fixing an actual bug.

Stonesjtu:
> > The filelock library already has robust internal mechanisms to handle lock cleanup.
> 
> Do you have reference to this? I believe I did some research before adding this and this was fixing an actual bug.

@joker-eph you can see this comment, AFAIK `file-lock` doesn't do clean-up for *flock* mechanism (which should be the default for most linux servers) since it causes race-condition.

https://github.com/tox-dev/filelock/blob/f7c3f8ea7defc21b3f0d736bb36711769cce840e/src/filelock/_unix.py#L57-L64

However when using `SoftFileLock`, the `.lock` file should be cleaned up as expected.

https://github.com/tox-dev/filelock/blob/f7c3f8ea7defc21b3f0d736bb36711769cce840e/src/filelock/_soft.py#L37-L42

PR #1783, opened 2025-09-26
Fix autotune profile min shape bigger than max shape.

weireweire:
## 📌 Description

The code assume the `opt_shapes` got from `spec.gen_tuning_buckets` is non-decreasing order, but actually the tuning config like [HERE](https://github.com/flashinfer-ai/flashinfer/blob/f21f94abab89a3c5193e38cfdc0edae3c9dd790c/flashinfer/fused_moe/core.py#L531) can use decreasing order, which make the min shape bigger than max shape.

Since current code base don't use the min max value. So it didn't bring issue yet.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/flashinfer/autotuner.py
+++ b/flashinfer/autotuner.py
@@ -637,6 +637,11 @@ class AutoTuner:
                 )
             else:
                 opt_shapes = spec.gen_tuning_buckets
+
+            # Normalize candidate buckets to be monotonically non-decreasing and non-empty
+            opt_shapes = tuple(sorted(set(opt_shapes)))
+            assert len(opt_shapes) > 0, "Empty tuning buckets are not allowed"
+
             opt_shapes_max = {
                 v1: v2
                 for v1, v2 in zip(opt_shapes, tuple(opt_shapes[1:]) + (float("inf"),))


flashinfer/autotuner.py
```diff
@@ -637,6 +637,11 @@ def _generate_optimization_profiles(
                 )
             else:
                 opt_shapes = spec.gen_tuning_buckets
+
+            # Normalize candidate buckets to be monotonically non-decreasing and non-empty
+            opt_shapes = tuple(sorted(set(opt_shapes)))
+            assert len(opt_shapes) > 0, "Empty tuning buckets are not allowed"
```
pavanimajety, line 643: @weireweire Is it requried to have assert for opt_shapes here? This leads to an error in vLLM during eager mode when we try to autotune. 

`vllm serve openai/gpt-oss-20b --tool-server demo --port 36759 --seed 0 --enforce-eager` leads to an error with this assert 

@nvpohanh

pavanimajety: It might be okay to not have this assert because I assume autotuner would always return a default value?

weireweire: if it's empty, will use runner_id=0 and tactic=-1.
Did your issue gone when remove this assert? If yes I think we can remove it and use default tactic.

pavanimajety: Yes, the issue went away when the assertion was commented out. Thanks for confirming! I’ll send out a small PR

PR #1790, opened 2025-09-27
test: Force upgrade cudnn backend for blackwell attention kernel UT

jimmyzho:
## 📌 Description

Following tests need higher version of nvidia-cudnn-cu12 package:
test_cudnn_prefill_deepseek.py
test_cudnn_prefill.py

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/scripts/task_test_blackwell_kernels.sh
+++ b/scripts/task_test_blackwell_kernels.sh
@@ -5,7 +5,7 @@ set -x
 : ${JUNIT_DIR:=$(realpath ./junit)}
 
 pip install -e . -v
-pip install --upgrade nvidia-cudnn-cu12
+pip install --upgrade "nvidia-cudnn-cu12>=9.11.0.98"
 
 EXIT_CODE=0
 scripts_to_run=(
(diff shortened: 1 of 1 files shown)


scripts/task_test_blackwell_kernels.sh
```diff
@@ -5,6 +5,7 @@ set -x
 : ${JUNIT_DIR:=$(realpath ./junit)}
 
 pip install -e . -v
+pip install --upgrade "nvidia-cudnn-cu12>=9.11.0.98"
```
nvmbreughe, line 8: I think we want to get away from installing packages in the test scripts and instead have them in the Docker container. 
Seems like the docker is set up to install that version: https://github.com/flashinfer-ai/flashinfer/blob/main/docker/install/install_python_packages.sh#L37 ?

What is the version that you were getting?

yzh119: Yes we should avoid adding more installation scripts here.

PR #1794, opened 2025-09-27
chore: improved URL handling for CUBIN/artifacts downloads

hypdeb:
Encountered some issues when adding some cubin's. This change set and tests should ensure future contributors do not run into the same issues.

- Warp `urljoin` into a helper which joins URLs independently of the presence of a trailing slash
- Add unit tests for this utility and the functions that use it

### DIFF
--- a/flashinfer/jit/cubin_loader.py
+++ b/flashinfer/jit/cubin_loader.py
@@ -33,14 +33,22 @@ FLASHINFER_CUBINS_REPOSITORY = os.environ.get(
     "https://edge.urm.nvidia.com/artifactory/sw-kernelinferencelibrary-public-generic-local/",
 )
 
+
 def safe_urljoin(base, path):
     """Join URLs ensuring base is treated as a directory."""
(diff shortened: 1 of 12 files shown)


flashinfer/jit/cubin_loader.py
```diff
@@ -133,26 +138,24 @@ def load_cubin(cubin_path, sha256) -> bytes:
     return b""
 
 
-def get_cubin(name, sha256, file_extension=".cubin", session=None):
+def get_cubin(file_name: str, sha256: str, session=None) -> bytes:
```
cyx-6, line 149: please also update the c++ side, where calling `getCubin` defined in `include/flashinfer/cubin_loader.h`. They leverage the default value of file extension to append `.cubin` automatically.

jdebache: Thanks, I missed those usages. I assumed all cubin file paths would come from the Python side somehow.

yzh119 (approved):
LGTM

PR #1798, opened 2025-09-28
bugfix: remove the append "a" logic if user specifies cuda arch explicitly

yzh119:
## 📌 Description

In our design of compilation context we will append "a" to cuda archs if the major arch is >= 9, however, we shouldn't change it if user specify them explicitly.

This PR changes the logic so that we only append "a" to cuda arch if `FLASHINFER_CUDA_ARCH_LIST` is not set explicitly, otherwise follow the user specified cuda arch.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests


aleozlx:
lgtm

PR #1809, opened 2025-09-29
Support checks PoC

nvmbreughe:
## 📌 Description

This PR adds is_*supported checks for backend and compute capability, through decorators. 

1. This allows us to check support before running
2. It also wraps the original function so it calls back the support check before running.
3. The wrapped function also adds an optional parameter "skip_check". A quick measurement show only minimal impact (14.51s without checks, 14.58s with checks for all of test_mm_fp4), so we should further benchmark the usefulness of this feature.

Example:
Screenshot 2025-10-12 at 9 39 06 PM

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/flashinfer/gemm.py
+++ b/flashinfer/gemm.py
@@ -83,6 +83,9 @@ from .utils import (
 
 DEFAULT_WORKSPACE_SIZE = 32 * 1024 * 1024
 
+# Error messages
+CUDNN_FP4_MXFP4_SM120_CUDNN_VERSION_ERROR = "cudnn FP4 GEMM with mxfp4 quantization is not supported on SM120 with cuDNN backend version < 9.14.0."
+
 
 def _match_sm_version(device: torch.device, sm_version: list[str]):
     major, minor = get_compute_capability(device)
@@ -1766,9 +1769,7 @@ def _cudnn_gemm_fp4_requirement(
         and _match_sm_version(a.device, ["120"])
         and cudnn.backend_version() < 91400
     ):
-        raise LibraryError(
-            "cudnn FP4 GEMM with mxfp4 quantization is not supported on SM120 with cuDNN backend version < 9.14.0."
-        )
+        raise LibraryError(CUDNN_FP4_MXFP4_SM120_CUDNN_VERSION_ERROR)
 
     _check_cudnn_fp4_availability()
 
--- a/tests/gemm/test_mm_fp4.py
+++ b/tests/gemm/test_mm_fp4.py
@@ -9,6 +9,7 @@ from flashinfer import (
     mxfp4_quantize,
 )
 from flashinfer.utils import get_compute_capability, LibraryError
+from flashinfer.gemm import CUDNN_FP4_MXFP4_SM120_CUDNN_VERSION_ERROR
 
 
 # TODO: Consdier splitting this function up for the various backends
@@ -92,14 +93,8 @@ def test_mm_fp4(
         assert cos_sim > 0.97
     except LibraryError as e:
         # TODO: Remove this check once cuDNN backend version is updated to 9.14.0
-        if (
-            backend == "cudnn"
-            and not use_nvfp4
-            and (compute_capability[0] == 12 and compute_capability[1] == 0)
-        ):
-            pytest.xfail(
-                "cudnn FP4 GEMM with mxfp4 quantization is not supported on SM120 with cuDNN backend version < 9.14.0."
-            )
+        if str(e) == CUDNN_FP4_MXFP4_SM120_CUDNN_VERSION_ERROR:
+            pytest.xfail(str(e))
         else:
             pytest.fail(str(e))
 


flashinfer/utils.py
```diff
@@ -740,3 +748,64 @@ def get_shuffle_matrix_sf_a_row_indices(
     row_indices = get_shuffle_matrix_a_row_indices(input_tensor, epilogue_tile_m)
 
     return row_indices
+
+
+def supports_backends(
```
sricketts: I think it might help with review if you write a docstring for `supports_backends` that defines the semantics of the arguments.

flashinfer/gemm.py
```diff
@@ -1999,6 +2000,11 @@ def _expand_block_scale_tensor_shape(block_scale_tensor, batch_size):
     return (tuple(block_scale_shape), tuple(block_scale_stride))
 
 
+@supports_backends(
+    ["cudnn", "trtllm", "cutlass"],
```
sricketts: Is this redundant with the declaration of the `backend` parameter?

`backend: Literal["cudnn", "trtllm", "cutlass"] = "cudnn",`

flashinfer/utils.py
```diff
@@ -740,3 +748,64 @@ def get_shuffle_matrix_sf_a_row_indices(
...
+    backends, capabilities=None, anti_capabilities=None, capability_tensor_arg=None
```
sricketts: nit: I wonder if "cc" or "compute_capabilities" would be more clear than "capabilities" -- or did you mean to signal something more generic than compute capabilities?

sricketts (changes requested):

flashinfer/utils.py
```diff
@@ -740,3 +748,134 @@ def get_shuffle_matrix_sf_a_row_indices(
...
+                )
+            if not is_problem_size_supported(*args, **kwargs):
+                raise ValueError(f"Problem size is not supported for {func.__name__}")
+            return func(*args, **kwargs)
+
+        wrapper.is_compute_capability_supported = is_compute_capability_supported
+        wrapper.is_backend_supported = is_backend_supported
+        wrapper.is_problem_size_supported = is_problem_size_supported
+        wrapper.__name__ = func.__name__
+        wrapper.__doc__ = func.__doc__
```
nvjullin: Can use [functools.wraps](https://docs.python.org/3/library/functools.html#functools.wraps) for more robust wrapping and standardized interface.

flashinfer/utils.py
```diff
@@ -740,3 +748,134 @@ def get_shuffle_matrix_sf_a_row_indices(
...
+                    f"Problem size check function is not provided for {func.__name__}"
+                )
+            else:
+                return problem_size_check(*args, **kwargs)
+
+        def wrapper(*args, **kwargs):
+            backend = kwargs.get("backend")
+            capability = None
+            if capability_tensor_arg:
+                tensor = kwargs.get(capability_tensor_arg)
```
nvjullin: Is there a reason why we need `capability_tensor_arg` instead of finding `torch.Tensor`s automatically and get the capability from them? We can also assert that they're all on the same device.

flashinfer/utils.py
```diff
@@ -740,3 +748,134 @@ def get_shuffle_matrix_sf_a_row_indices(
...
 
     return row_indices
+
+
+def supports_backends(
+    backends,
+    capabilities=None,
+    anti_capabilities=None,
+    capability_tensor_arg=None,
+    problem_size_check=None,
```
nvjullin: Add type hints, perhaps excluding `problem_size_check` if it's too tedious or just `typing.Callable`.

nvjullin:
The checks currently live very far away from the implementation and updating them to be consistent with each other can eventually become a maintenance problem. The conditional checks are also quite tricky to get correct. For example, it's not easy to tell if the mxfp4 checks are correct.
```python
    if not use_nvfp4 and block_size != 32:
        raise ValueError("mxfp4 supports block_size = 32.")

```
Shouldn't the checks be reordered to avoid confusing error messages?
1. User tries trtllm + block_size=16 and gets rejected by block_size=32
2. User then tries trtllm + block_size=32 and gets rejected by only cudnn is supported

---

Instead of having one top level `supports_backends`, perhaps consider a two level design:
1. Local requirement decorator `requirement` written for each backend entrypoint
2. Top level `backend_requirement` that composes requirements

For example:
```python
def cudnn_gemm_fp4_requirement(
    # ...
):
        if (
            not use_nvfp4
            and _match_sm_version(a.device, ["120"])
            and cudnn.backend_version() < 91400
        ):
            raise LibraryError(
                "cudnn FP4 GEMM with mxfp4 quantization is not supported on SM120 with cuDNN backend version < 9.14.0."
            )

        _check_cudnn_fp4_availability()
        # ...

@requirement(cudnn_gemm_fp4_requirement, capability=["100", "101", "102"])
def execute_cudnn_gemm_fp4_graph(
    # ...


@backend_requirement({
    "cudnn": execute_cudnn_gemm_fp4_graph.requirement,
    "trtllm": #...
})
def mm_fp4(
    # ...
```

This also means that all requirements are enforced to be local to the backend and won't affect each other.

flashinfer/utils.py
nvmbreughe: Residing on the same device is not a requirement for comm kernels.
But we could consider dynamically finding the tensors and ensuring they have the required capability. Agreed it would be cleaner.

nvmbreughe:
> Instead of having one top level `supports_backends`, perhaps consider a two level design:
> 
> 1. Local requirement decorator `requirement` written for each backend entrypoint
> 2. Top level `backend_requirement` that composes requirements

Thank you @nvjullin for the excellent suggestion.
I think these are two discussions:
1. Separate the support checks
2. Separate the execution routines

While both are valid points, we prioritize separating the checks for now. Not all APIs are as cleanly to separate (2) atm and there is a plan for a more OO Backend class @Anerudhan. That does overlap somewhat with the support checks, as eventually we would be able to do something like cudnn_backend->check_mmfp4_support(). 

So I think as an intermediary step, and to get tighter checks in, we could do something like this:


```python
@supported_compute_capability(["100", "101", "102"])
def cudnn_gemm_fp4_requirement(
    # ...
):
        if (
            not use_nvfp4
            and _match_sm_version(a.device, ["120"])
            and cudnn.backend_version() < 91400
        ):
            raise LibraryError(
                "cudnn FP4 GEMM with mxfp4 quantization is not supported on SM120 with cuDNN backend version < 9.14.0."
            )

        _check_cudnn_fp4_availability()
        # ...


@backend_requirement({
    "cudnn": execute_cudnn_gemm_fp4_graph.requirement,
    "trtllm": #...
    },
   common_check=common_fp4_checks # To be called by all backend checks
})
def mm_fp4(
    )
    if backend == "cudnn":
              # cudnn path
    elif backend == "trtllm":
              # trtllm path         
```

flashinfer/gemm.py
nvmbreughe: Unfortunately with the current design (using a dictionary) it is no longer possible to avoid this.

nvjullin:
> While both are valid points, we prioritize separating the checks for now. Not all APIs are as cleanly to separate (2) atm and there is a plan for a more OO Backend class @Anerudhan. That does overlap somewhat with the support checks, as eventually we would be able to do something like cudnn_backend->check_mmfp4_support().

I wasn't aware, thanks for the info. LGTM.

nvmbreughe:
> > While both are valid points, we prioritize separating the checks for now. Not all APIs are as cleanly to separate (2) atm and there is a plan for a more OO Backend class @Anerudhan. That does overlap somewhat with the support checks, as eventually we would be able to do something like cudnn_backend->check_mmfp4_support().
> 
> I wasn't aware, thanks for the info. LGTM.

Thank you for the excellent suggestions, @nvjullin

tests/gemm/test_mm_fp4.py
```diff
@@ -100,7 +101,7 @@ def test_mm_fp4(
                 "cudnn FP4 GEMM with mxfp4 quantization is not supported on SM120 with cuDNN backend version < 9.14.0."
```
sricketts: The logic in the `if` and the `xfail` msg look duplicated from `_cudnn_gemm_fp4_requirement` in `gemm.py`. Can we DRY it somehow?

nvmbreughe: Good point. As a side note: I had was thinking yesterday to extend this for other things: e.g., throw a pytest.skip if the SM is not listed in the decorator. But then we are conflating unit tests and code base logic. So I believe we should in that case be WET (write everything twice).

Again for here, we should be DRY as you pointed out. I addressed it by identifying the exact error message,

PR #1810, opened 2025-09-30
tests: Update support for tgv_gemm to SM100 only and add to ut

jimmyzho:
## 📌 Description

add tgv_gemm to tests and update support surface

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests


flashinfer/gemm.py
```diff
@@ -1014,8 +1014,8 @@ def tgv_gemm_sm100(
         - Tensor b is expected to be in column-major layout (transposed from typical PyTorch row-major)
     """
     # Verify SM100 architecture support
-    if not _match_sm_version(a.device, ["100", "103", "110"]):
-        raise ValueError("TGV GEMM requires SM100, SM103, or SM110 architecture")
+    if not _match_sm_version(a.device, ["100"]):
```
yzh119: @yangs75 @yang-yifan have you tested with b300? I guess the kernel should also be compatible with `sm_103`?

jimmyzho: I can confirm that the unit tests can pass with B300 - if they were meant to be compatible with SM103

yzh119: sm_100 kernels should be compatible with sm_103 if compiled with `sm_100f`, but I believe these kernels haven't used sm_103 specific features.
How about removing `110` and keep `["100", "103"]`?

Yang-YiFan: Yeah it should work on B200 and B300 since it hasn’t used any sm103 specific features.

flashinfer/gemm.py
```diff
@@ -924,10 +925,17 @@ def gen_gemm_sm100_module_tgv(dtype: torch.dtype = torch.bfloat16) -> JitSpec:
         )
         write_if_different(dest_path, source)
 
+    if _match_sm_version(torch.device("cuda"), ["100"]):
+        sm_nvcc_flags = sm100a_nvcc_flags
+    elif _match_sm_version(torch.device("cuda"), ["103"]):
+        sm_nvcc_flags = sm103a_nvcc_flags
+    else:
+        raise ValueError(f"Unsupported SM version {torch.device('cuda').sm_version}")
+
```
yzh119: Hi @jimmyzho we shouldn't use any gpu related functions inside `gen_*_module` files because they are designed to be host-only (CPU machines can compile them as well).

You can either separate these two functions:
- gen_gemm_sm100_tgv_module
- gen_gemm_sm103_tgv_module

Or use sm100f_nvcc_flags here. (`sm_100f` should be compatible with both b200 and b300).

yzh119: sm100f_nvcc_flags is not defined in https://github.com/flashinfer-ai/flashinfer/blob/d252ed77539170636c33ec424950cc6a2f6b5864/flashinfer/jit/core.py#L76, you can add another line there:
```
sm100f_nvcc_flags = ["-gencode=arch=compute_100f,code=sm_100f"] + common_nvcc_flags
```

yzh119 (approved):
LGTM

yzh119:
Another issue emerges, sm_100f is not available until cuda 12.9+

yzh119:
Hi @jimmyzho I added some version checks in https://github.com/flashinfer-ai/flashinfer/pull/1810/commits/51f4c995a24cb6a195ec2c392797139bc0548444 to fix the CI error, does that look good to you (some random files introduced in this commit was removed in https://github.com/flashinfer-ai/flashinfer/pull/1810/commits/215f06e90f808de52d0fc65defdc26dbf8387fff)?

PR #1824, opened 2025-09-30
[Perf] Cache device property functions to avoid recomputation

Jialin:
## 📌 Description

Reduce overhead among cuda launches by caching device property functions. In small batches, we observed GPU bubbles which means for certain cases, CPU workload (e.g. cuda launch preparations) delays GPU kernel launches.

In this PR, we simply cache device property functions to reduce the CPU workload overhead.

**Before**
Screenshot 2025-09-30 at 3 41 21 PM

**After**
Screenshot 2025-09-30 at 3 54 54 PM

## 🔍 Related Issues

N/A

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Reviewer Notes

N/A

### DIFF
--- a/flashinfer/utils.py
+++ b/flashinfer/utils.py
@@ -17,7 +17,6 @@ limitations under the License.
 import functools
 import math
 from enum import Enum
-from functools import cache
 from typing import Callable, Dict, Iterable, Optional, Sequence, Tuple, Union
 
 import torch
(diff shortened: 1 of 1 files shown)


flashinfer/utils.py
```diff
@@ -17,6 +17,7 @@
 import functools
 import math
 from enum import Enum
+from functools import cache
```
yzh119: There are some other functions in this file using `functools.cache`, we should unify to style to either all using `@functools.cache` or `@cache`

Jialin: Updated.

yzh119 (approved):
LGTM

Issue #1806, opened 2025-09-29
Allow using system libraries when compiling from source


BwL1289:
For barebones package managers, it would be helpful to have a `USE_SYSTEM_LIBS` flag to toggle in order to use system versions of `cutlass` and `spdlog` when building from source. 

Here's an [example](https://github.com/awslabs/aws-crt-python/blob/c4a41abfd8f7bb196c98604cc1d63302f2d09016/setup.py#L176-L179) from aws-crt-python.

yzh119:
Hi @BwL1289 agreed that's better option then current design, we should also do the same thing for nvshmem.

Would you mind creating a PR for this feature?

BwL1289:
> Would you mind creating a PR for this feature?

@yzh119 don't have the bandwidth, but if I did I would!

BwL1289:
Why was this closed?

PR #1827, opened 2025-10-01
chore: allow custom paths for external dependencies like CUTLASS

yzh119:
## 📌 Description

Implement features mentioned in #1806 , now user can specify custom dependencies path (e.g. if they are installed as system library) via environment variables `FLASHINFER_CUTLASS_INCLUDE_PATH` etc.

User can run `python -m flashinfer show-config` to display these environment variables.

cc @bwl1289

## 🔍 Related Issues

#1806

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/flashinfer/comm/nvshmem.py
+++ b/flashinfer/comm/nvshmem.py
@@ -13,10 +13,14 @@ from ..jit import gen_jit_spec
 
 def gen_nvshmem_module() -> JitSpec:
     lib_dirs = jit_env.get_nvshmem_lib_dirs()
+    # Check new environment variable first, then fall back to old one for backward compatibility
+    ldflags_env = os.environ.get("FLASHINFER_NVSHMEM_LDFLAGS") or os.environ.get(
+        "NVSHMEM_LDFLAGS", ""
+    )
     ldflags = (
         [f"-L{lib_dir}" for lib_dir in lib_dirs]
         + ["-lnvshmem_device"]
-        + shlex.split(os.environ.get("NVSHMEM_LDFLAGS", ""))
+        + shlex.split(ldflags_env)
     )
 
     return gen_jit_spec(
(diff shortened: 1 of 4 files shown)


flashinfer/comm/nvshmem.py
```diff
@@ -13,10 +13,14 @@
 
 def gen_nvshmem_module() -> JitSpec:
     lib_dirs = jit_env.get_nvshmem_lib_dirs()
+    # Check new environment variable first, then fall back to old one for backward compatibility
+    ldflags_env = os.environ.get("FLASHINFER_NVSHMEM_LDFLAGS") or os.environ.get(
```
aleozlx, line 17: shall we have documentation collecting these FLASHINFER* vars in some .md file, and briefly mention what they may affect and the order in which related vars are consulted. this could help both users and developers understand the expected behavior (which could otherwise run away after a few ppl modifying the same)

aleozlx (approved):
lgtm

flashinfer/comm/nvshmem.py
yzh119: currently user can run
```
python -m flashinfer show-config
```
do display these environment variables, but yes let's clearly document it.

zhyncs:
@yzh119 even we don't want to release rc version, we need to fix the release issue https://github.com/flashinfer-ai/flashinfer/actions/runs/18338855616/job/52229146277

flashinfer/jit/env.py
```diff
@@ -74,15 +74,39 @@ def _get_workspace_dir_name() -> pathlib.Path:
...
+def get_spdlog_include_dir():
+    path = os.environ.get("FLASHINFER_SPDLOG_INCLUDE_PATH")
+    if path is not None:
+        return pathlib.Path(path)
+
+    # Fall back to default path
+    return _package_root / "data" / "spdlog" / "include"
+
+
+# For backward compatibility, keep these as properties
```
nvmbreughe, line 100: [nit] Maybe mark as deprecated? In next major version we can then get rid of it.

PR #1841, opened 2025-10-01
bugfix: fix synchronize logic error in tests/comm/test_trtllm_alltoall.py

nv-yunzheq:
## 📌 Description

Fix the incorrect sychronization logic error in the unit test that causes it to hang.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/tests/comm/test_trtllm_alltoall.py
+++ b/tests/comm/test_trtllm_alltoall.py
@@ -702,6 +702,8 @@ def test_moe_alltoall_prepare(
     )
 
     stream = torch.cuda.Stream()
+    cur_stream = torch.cuda.current_stream()
+    stream.wait_stream(cur_stream)
     with torch.cuda.stream(stream):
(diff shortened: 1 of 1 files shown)


nvmbreughe:
LGTM.

Just curious how this resolve the hang? Is it because "moe_prepare" runs twice for rank 0, and before we allowed them to overlap, whereas now there is a sync in between?

sync diagrams

Incomming arrow means that the destination stream waits on the source stream

nv-yunzheq:
@nvmbreughe You diagram should be correct.
It seems that the first moe_prepare is to jit the kernel with ep_size = 1 and ep_rank = 0. While the second moe_prepare would reset everything and run on a multi-stream setting. We probably could put the first moe_prepare in the default stream, as I don't see a benefit of having it runing on a separate stream.

PR #1843, opened 2025-10-02
feat: add warp-level persistent qk norm

happierpig:
## 📌 Description
Recent models are using QK normalization right before RoPE and core self-attention (e.g., Qwen-3, Wan). Existing RMSNorm implementation in FlashInfer falls short on optimal for:
1. Extra shared memory reduction step.
2. Do not support non-contiguous layout on the middle dimension. E.g., q maybe [batch_size, :num_qo_heads, head_dim].

This PR implements a persistent version of RMSNorm, where each head is unrolled with each warp.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/csrc/flashinfer_norm_binding.cu
+++ b/csrc/flashinfer_norm_binding.cu
@@ -19,8 +19,6 @@ using tvm::ffi::Tensor;
 
 void rmsnorm(Tensor out, Tensor input, Tensor weight, double eps, bool enable_pdl);
 
-void qk_rmsnorm(Tensor out, Tensor input, Tensor weight, double eps, bool enable_pdl);
-
 void fused_add_rmsnorm(Tensor input, Tensor residual, Tensor weight, double eps, bool enable_pdl);
 
 void gemma_rmsnorm(Tensor out, Tensor input, Tensor weight, double eps, bool enable_pdl);
@@ -29,7 +27,6 @@ void gemma_fused_add_rmsnorm(Tensor input, Tensor residual, Tensor weight, doubl
                              bool enable_pdl);
 
 TVM_FFI_DLL_EXPORT_TYPED_FUNC(rmsnorm, rmsnorm);
-TVM_FFI_DLL_EXPORT_TYPED_FUNC(qk_rmsnorm, qk_rmsnorm);
 TVM_FFI_DLL_EXPORT_TYPED_FUNC(fused_add_rmsnorm, fused_add_rmsnorm);
 TVM_FFI_DLL_EXPORT_TYPED_FUNC(gemma_rmsnorm, gemma_rmsnorm);
 TVM_FFI_DLL_EXPORT_TYPED_FUNC(gemma_fused_add_rmsnorm, gemma_fused_add_rmsnorm);
--- a/csrc/norm.cu
+++ b/csrc/norm.cu
@@ -23,55 +23,59 @@ using tvm::ffi::Tensor;
 
 void rmsnorm(Tensor output, Tensor input, Tensor weight, double eps, bool enable_pdl) {
   CHECK_LAST_DIM_CONTIGUOUS_INPUT(input);
-  CHECK_LAST_DIM_CONTIGUOUS_INPUT(weight);
-  CHECK_DEVICE(input, weight);
-  CHECK_DIM(2, input);   // input: (batch_size, hidden_size)
-  CHECK_DIM(1, weight);  // weight: (hidden_size)
-  TVM_FFI_ICHECK_EQ(input->shape[1], weight->shape[0]);
-  unsigned int batch_size = input->shape[0];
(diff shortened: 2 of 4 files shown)


csrc/flashinfer_norm_binding.cu
```diff
@@ -19,6 +19,8 @@ using tvm::ffi::Tensor;
 
 void rmsnorm(Tensor out, Tensor input, Tensor weight, double eps, bool enable_pdl);
 
+void qk_rmsnorm(Tensor out, Tensor input, Tensor weight, double eps, bool enable_pdl);
```
yzh119: why do we need a new API for this?

yzh119 (commented):
> Do not support non-contiguous layout on the middle dimension

Can you explain more?

happierpig:
> > Do not support non-contiguous layout on the middle dimension
> 
> Can you explain more?

Here is a detailed motivation example:

When serivng Qwen3-8B, there is a RMSNorm performed on each `head_dim` (https://github.com/huggingface/transformers/blob/9f2d5666f8fda5b647e5f64dfc8ba1edd7a87a1e/src/transformers/models/qwen3/modeling_qwen3.py#L183).
```python
q, k, v = self.wqkv(x).split([q_size, kv_size, kv_size], dim=-1)
q = q.view(seqlen, self.num_qo_heads, self.head_dim)
k = k.view(seqlen, self.num_kv_heads, self.head_dim)
v = v.view(seqlen, self.num_kv_heads, self.head_dim)
# step 1. apply RMSNorm to q/k
# step 2. RoPE
# step 3. attention
```
Prior to this PR, we need one additional `.contiguous` to feed into the existing API.
```python
q = flashinfer.norm.rmsnorm(q.reshape(-1, head_dim), self.weight, self.eps)
q = q.view(seqlen, self.num_qo_heads, self.head_dim)
# same to k
```

csrc/flashinfer_norm_binding.cu
happierpig: I see. We can keep the `rmsnorm` API, and dispatch to the persistent version if `input.dim() == 3`

yzh119 (commented):
So functionality-wise it should be a extension to the original rmsnorm API right? We support non-contiguous layout because we passed the strides to the kernel.

I would recommend making your implementation the default implementation of rmsnorm, and reuse the original API.

happierpig:
> So functionality-wise it should be a extension to the original rmsnorm API right? We support non-contiguous layout because we passed the strides to the kernel.
> 
> I would recommend making your implementation the default implementation of rmsnorm, and reuse the original API.

Original APi doesn't support this kind of non-contiguous. As there is only one stride (`stride_input`) in the original API, [batch_size, :num_heads, head_dim] is not support as `dim=1` is contiguous while `dim=0` is not. Both `stride_input_n` and `stride_input_h` are needed to index such tensors. https://github.com/flashinfer-ai/flashinfer/blob/398dc267760dd261c1b3c9f5cbadb6f84dcc5f3c/include/flashinfer/norm.cuh#L32

yzh119:
> Original APi doesn't support this kind of non-contiguous. As there is only one stride (stride_input) in the original API, [batch_size, :num_heads, head_dim] is not support as dim=1 is contiguous while dim=0 is not. Both stride_input_n and stride_input_h are needed to index such tensors. 

Yes, but can add support for 3d input to rmsnorm api, right?

* When input is 3d tensor, use the persistent kernel implemented by you.
* When input is 2d tensor, view it as 3d (1, b, d), and we can still use your persistent kernel.

and we can safely remove the original implementation.

happierpig:
> > Original APi doesn't support this kind of non-contiguous. As there is only one stride (stride_input) in the original API, [batch_size, :num_heads, head_dim] is not support as dim=1 is contiguous while dim=0 is not. Both stride_input_n and stride_input_h are needed to index such tensors.
> 
> Yes, but can add support for 3d input to rmsnorm api, right?
> 
> * When input is 3d tensor, use the persistent kernel implemented by you.
> * When input is 2d tensor, view it as 3d (1, b, d), and we can still use your persistent kernel.
> 
> and we can safely remove the original implementation.

IMO this warp-level persistent version has less parallelism in `hidden_dim` dim, which may fail short on normal RMSNorm cases (i.e., [batch_sizem 4096]). Is it okay to keep both implementation and selectively dispatch?

yzh119 (commented):
> IMO this warp-level persistent version has less parallelism in hidden_dim dim, which may fail short on normal RMSNorm cases (i.e., [batch_size, m 4096]). Is it okay to keep both implementation and selectively dispatch?

Sounds good, I agree with keep both implementations the.

Can you add more docstring on clarifying the difference between `qk_rmsnorm` and `rmsnorm`, or unify the two interfaces.

Issue #1833, opened 2025-10-01
ci coverage: add unittest for flashinfer-cli


yzh119:
Currently flashinfer-cli functionalities do not have test coverage and was sometimes broken by new commits, we should also cover flashinfer-cli in our unittests.

sricketts:
I can add more CLI test coverage as a follow up to #1846.

sricketts:
#1848 (ready for review) is intended to address this issue.

sricketts:
#1848 resolves this. Closing.

PR #1726, opened 2025-09-19
ci/cd: bringup flashinfer-jit-cache package

yzh119:
## 📌 Description

Move aot modules to a standalone package `flashinfer-jit-cache`.

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [x] Tests have been added or updated as needed.
- [x] All tests are passing (`unittest`, etc.).

PR #1846, opened 2025-10-02
bugfix: show-config command

sricketts:
## 📌 Description

I believe the show-config command line option was broken by #1794, because it change the return type of `get_artifacts_status` from a list of size 3 tuples to a tuple of size 2 tuples. This code in `__main__.py` expects the former:

```
    status = get_artifacts_status()
    num_downloaded = sum(1 for _, _, exists in status if exists)
    total_cubins = len(status)
```

This PR has a fix and a test to protect against future breaks.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Reviewer Notes

Is this the right place to add CLI tests? Note that I've added a `tests/cli` subdirectory.

### DIFF
--- a/flashinfer/__main__.py
+++ b/flashinfer/__main__.py
@@ -114,7 +114,7 @@ def show_config_cmd():
     click.secho("=== Downloaded Cubins ===", fg="yellow")
 
     status = get_artifacts_status()
-    num_downloaded = sum(1 for _, _, exists in status if exists)
+    num_downloaded = sum(1 for _, exists in status if exists)
     total_cubins = len(status)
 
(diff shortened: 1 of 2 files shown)


sricketts:
I don't think we want to monkeypatch `get_artifacts_status` in the test -- will update.

yzh119:
Yes I also had a fix in #1726 , but might be better to merge this first.

We need to add unittest for cli functionality: https://github.com/flashinfer-ai/flashinfer/issues/1833

yzh119 (approved):

PR #1852, opened 2025-10-03
Fix: Prevent race condition in cubin loader when file is being consumed

yzh119:
## 📌 Description
There's a race condition between cubin download and compilation:

  Timeline of the race condition:
  1. Worker1: Acquires lock -> downloads cubin -> releases lock
  2. Worker1: Starts compile operation (reading the cubin file) without lock protection
  3. Worker2: Acquires lock -> detects file corruption (sha256 mismatch) -> overwrites the file Worker1 is reading!
  4. Worker1: Compilation fails or uses corrupted data

  Root cause: The critical region (lock scope) only protects the download operation, but doesn't protect the subsequent compile/load operations. We cannot simply extend the lock scope because:
  - load cubin + build op + load op are in different code paths
  - Cannot use the same lock across the entire pipeline (across C++/python)

###  Solution

  Use atomic file rename to leverage filesystem inode semantics:

  1. Download/copy to temporary file: {local_path}.tmp
  2. Atomically rename to final path: os.replace(tmp, local_path)

  Why this works:
  - os.replace() is an atomic operation
  - On Unix systems, when Worker2 replaces the file, Worker1's open file descriptor still points to the old inode
  - Worker1 can safely continue reading the old file
  - New readers (after the replace) see the new file
  - The old inode is only deleted when all file descriptors are closed

###  Changes

  - flashinfer/jit/cubin_loader.py:85: Atomic rename for local file copy
  - flashinfer/jit/cubin_loader.py:103: Atomic rename for URL downloads

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Reviewer Notes

cc @nvpohanh @nvjullin @joker-eph

### DIFF
--- a/flashinfer/jit/cubin_loader.py
+++ b/flashinfer/jit/cubin_loader.py
@@ -80,7 +80,9 @@ def download_file(
             # Handle local file copy
             if os.path.exists(source):
                 try:
-                    shutil.copy(source, local_path)
+                    temp_path = f"{local_path}.tmp"
+                    shutil.copy(source, temp_path)
+                    os.replace(temp_path, local_path)  # Atomic rename
                     logger.info(f"File copied successfully: {local_path}")
                     return True
                 except Exception as e:
@@ -93,8 +95,12 @@ def download_file(
                     response = session.get(source, timeout=timeout)
                     response.raise_for_status()
 
-                    with open(local_path, "wb") as file:
+                    temp_path = f"{local_path}.tmp"
+                    with open(temp_path, "wb") as file:
                         file.write(response.content)
+
+                    # Atomic rename to prevent readers from seeing partial writes
+                    os.replace(temp_path, local_path)
 
                     logger.info(
                         f"File downloaded successfully: {source} -> {local_path}"


joker-eph:
> Worker2: Acquires lock -> detects file corruption (sha256 mismatch) -> overwrites the file Worker1 is reading!

How is this possible that there is sha256 mismatch?
Why would worker1 be happy with the file?

yzh119:
Hi @joker-eph 

> How is this possible that there is sha256 mismatch?


The race condition occurs because `load_cubin` is called outside the lock:

- Worker 1 starts downloading (inside lock), Worker 2's `load_cubin` reads the file while it's being written -> fails
- Worker 1 releases lock and calls `load_cubin` to verify
- Worker 2 acquires lock and starts downloading/overwriting 
- Worker 1 reads while Worker 2 writes -> SHA256 mismatch

The filelock only protects downloads, not the read operations. The atomic rename ensures readers never see a file being written.

flashinfer/jit/cubin_loader.py
```diff
@@ -93,9 +95,13 @@ def download_file(
                     response = session.get(source, timeout=timeout)
                     response.raise_for_status()
 
-                    with open(local_path, "wb") as file:
+                    temp_path = f"{local_path}.tmp"
```
nvmbreughe: [nit] Duplication from L83. Maybe reuse?

yzh119: sounds good

yzh119: cleaned up in https://github.com/flashinfer-ai/flashinfer/pull/1852/commits/8d1e855f26dc9203c398eb6f388b705b27b9f9d8

yzh119:
@joker-eph I created a test case in https://github.com/flashinfer-ai/flashinfer/pull/1852/commits/de1a41f4c6044b197df4f84dc3daf0ff15f98e2e, in earlier codebase, it failed at iteration 83 in one run (the exact iteration varies due to race condition timing).

PR #1870, opened 2025-10-06
ci: Create Github Action to Automate CODEOWNER update

yzh119:
## 📌 Description

Duplicate of #1869 but created from flashinfer/workflow-update-codeowner to make sure we have permission.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests


nvmbreughe:
LGTM

.github/workflows/update-codeowners.yml
```diff
@@ -0,0 +1,113 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+
+      - name: Run CODEOWNERS analyzer
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          python scripts/codeowner_analyzer.py \
```
sricketts, line 38: Can we use `--allowed-users-file` here to explicitly control the group of trusted codeowners?

.github/workflows/update-codeowners.yml
```diff
@@ -0,0 +1,113 @@
+name: Update CODEOWNERS
+
+on:
+  schedule:
+    # Run weekly on Monday at 00:00 UTC
+    - cron: '0 0 * * 1'
+  workflow_dispatch: # Allow manual triggering
+  pull_request:
```
sricketts: Am I reading this correctly that it will run on every PR in addition to running weekly?

.github/workflows/update-codeowners.yml
yzh119: Yes, do you want this list to be public or private?

.github/workflows/update-codeowners.yml
yzh119: removed in https://github.com/flashinfer-ai/flashinfer/pull/1870/commits/75476931c7dd0c59222cdfdb6ab95cd08451abd7

.github/workflows/update-codeowners.yml
sricketts: I think public is ok. This will give people an opportunity to either ask to be removed (if they no longer want to be a codeowner) or ask to be included (if they want to be but aren't for some reason).

nvmbreughe:
Should we add tests/ to the codeowners anyway (noticed it got intentionally excluded)? If a test fails, it may be good to find the person who is most knowledgeable about it. Sometimes it's easy to find the associated module, sometimes it may not be. 

Just a nitpick, since we could also use git blame.

yzh119:
Updated the script to use github CLI to extract username.

Added `scripts/authorized_codeowner.txt` per @sricketts 's suggestion.

> Should we add tests/ to the codeowners anyway (noticed it got intentionally excluded)? If a test fails, it may be good to find the person who is most knowledgeable about it. Sometimes it's easy to find the associated module, sometimes it may not be.

Great suggestion! we can work on this in later PRs :) @nvmbreughe

.github/workflows/update-codeowners.yml
```diff
@@ -0,0 +1,111 @@
...
+        with:
+          python-version: '3.11'
+
+      - name: Run CODEOWNERS analyzer
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          python scripts/codeowner_analyzer.py \
+            --output .github/CODEOWNERS \
+            --depth 3 \
```
nvmbreughe: [nit] Should we make this variables? Could lead to copy-paste issues in the "Create Pull Request" step.

yzh119: updated in: https://github.com/flashinfer-ai/flashinfer/pull/1870/commits/55bc90ec210e6387981fcd33611abcae96893f3b, where both the script input arguments and pull request templates using the same set of values.

PR #1894, opened 2025-10-09
fix vllm graph register and add test

NVShreyas:
## 📌 Description

The all-reduce buffers are incorrectly registered during the cuda graph capture phase when using vLLM custom all reduce. The base_ptr is not initialized and leads to illegal memory access error.
vLLM's implementation with proper memory initialization - https://github.com/vllm-project/vllm/blob/0d21b9b51eccabfa1f8114eab2df61d75459bee7/csrc/custom_all_reduce.cuh#L452

This PR ensures that the memory pointer is initialized correctly and adds tests for vLLM custom all reduce during cuda graph capture

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/include/flashinfer/comm/vllm_custom_all_reduce.cuh
+++ b/include/flashinfer/comm/vllm_custom_all_reduce.cuh
@@ -44,7 +44,7 @@ struct cuda_error : public std::runtime_error {
 namespace vllm {
 
 constexpr int kMaxBlocks = 36;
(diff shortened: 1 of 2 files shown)


NVShreyas:
> Hi @NVShreyas would you mind providing more context in the PR description?

Hi @yzh119 , I've added some context

Issue #1898, opened 2025-10-09
[Bug] trtllm-gen attention kernel issues for batch size 1


bkryu:
Description of the issue and test case added in [PR1897](https://github.com/flashinfer-ai/flashinfer/pull/1897). Reproduced here:

Trtllm-gen's attention kernels have been discovered to fail tests when batch size is 1. When batch size 1, 

[STILL REQUIRES A FIX] `test_trtllm_batch_decode`: produces incorrect outputs with newly added parameters
```
## Running pytest ./tests/attention/test_trtllm_gen_attention.py::test_trtllm_batch_decode -v
>                   torch.testing.assert_close(
                        output.float(),
                        output_wrapper.float(),
                        rtol=1e-1,
                        atol=1e-1,
                    )
E                   AssertionError: Tensor-likes are not close!
E                   
E                   Mismatched elements: 1480 / 8192 (18.1%)
E                   Greatest absolute difference: 64.021484375 at index (0, 46, 106) (up to 0.1 allowed)
E                   Greatest relative difference: 1.625 at index (0, 56, 109) (up to 0.1 allowed)
```

[UPDATE: NOW FIXED in #1912] `test_trtllm_gen_prefill_deepseek`: can trigger an IMA with the newly added parameters
```
## Running pytest ./tests/attention/test_trtllm_gen_attention.py::test_trtllm_gen_prefill_deepseek -v
>           default_generator.manual_seed(seed)
E           torch.AcceleratorError: CUDA error: an illegal memory access was encountered
E           CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
E           For debugging consider passing CUDA_LAUNCH_BLOCKING=1
E           Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.

/opt/conda/envs/py312/lib/python3.12/site-packages/torch/cuda/random.py:129: AcceleratorError
```

bkryu:
`test_trtllm_gen_prefill_deepseek` IMA issue has been fixed in #1912 

`test_trtllm_batch_decode` outpupt mismatch still needs to be fixed.

nvpohanh:
@bkryu is this still an issue?

bkryu:
> [@bkryu](https://github.com/bkryu) is this still an issue?

No the issue has been resolved. Closing. Thanks @nvpohanh for checking in

PR #1912, opened 2025-10-10
fix: Fix trtllm-gen prefill IMA when batch_size==1

bkryu:
## 📌 Description

Current PR fixes the test and benchmark codes IMAs when running trtllm-gen paged & ragged prefill with batch size 1 -- the issue was described in https://github.com/flashinfer-ai/flashinfer/issues/1898

Root cause of the issue: `flashinfer.prefill.trtllm_ragged_attention_deepseek` and `flashinfer.prefill.trtllm_batch_context_with_kv_cache` both require `max_q_len` to match the length of the query when batch size is 1.

**Updated PR:**
Issue has been addressed from the kernel-side so that the "*`max_q_len` to match the length of the query when batch size is 1*" is no longer required. 

Current PR updates trtllm-gen FMHA cubins to latest and brings minor updates to kernel metadata.

Unit test results after PR: 
```
$ pytest tests/attention/test_trtllm_gen_attention.py 
...
platform linux -- Python 3.12.11, pytest-8.4.2, pluggy-1.6.0
rootdir: /flashinfer
configfile: pytest.ini
collected 2320 items   
...
2055 passed, 264 skipped, 1 xfailed in 224.43s (0:03:44)
```

**Description of previous solution:**
~~Updating `max_q_len` to `cum_seq_lens_q[-1].item()` within the `trtllm_ragged_attention_deepseek` or `trtllm_batch_context_with_kv_cache` functions are not a viable option because the CPU-side synchronization breaks the deterministic and fully device-side execution required during CUDA graph capture. The workaround was thus to update the test & benchmark codes that call the trtllm prefill functions, and clearly state in the docstring that when batch_size == 1, max_q_len must match the query size.~~

## 🔍 Related Issues

https://github.com/flashinfer-ai/flashinfer/issues/1898

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Enabled an additional native attention backend for batch_size=1 where other constraints allow, expanding supported scenarios.

* **New Features**
  * Added a kernel parameter to enable future block-sparse attention support (no change to current behavior).

* **Documentation**
  * Updated benchmarks docs to note support for an additional attention optimization/backend.

* **Tests**
  * Expanded test coverage with configurable sequence lengths and a dedicated batch-size-1 test; removed a previous expected-failure for that path.

### DIFF
--- a/include/flashinfer/trtllm/fmha/kernelParams.h
+++ b/include/flashinfer/trtllm/fmha/kernelParams.h
@@ -152,6 +152,8 @@ struct KernelParams {
   int32_t mStartTokenIdxSfO;
   // The sum of sequence lengths for Q and K/V.
   int32_t mSumOfSeqLensQ, mSumOfSeqLensKv;
+  // The flag to use block sparse attention.
+  bool mUseBlockSparseAttention;
 
   // Create the TMA shape/stride for Q.
   template <class FmhaOptions>
@@ -699,6 +701,8 @@ struct KernelParams {
     params.mStartTokenIdxSfO = options.mSfStartTokenIdx;
     params.mScaleSfKv = options.mScaleSfKv;
     params.ptrSoftmaxStats = options.softmaxStatsPtr;
+    // TODO: Integrate trtllm block-sparse attention kernels when needed.
+    params.mUseBlockSparseAttention = false;
     return params;
   }
 };


benchmarks/routines/attention.py
```diff
@@ -932,6 +928,12 @@ def testBatchPrefillWithPagedKVCacheWrapper(args):
         v_fp8 = (v_data / v_scale).to(kv_dtype)
         kv_cache = torch.cat([k_fp8, v_fp8], dim=1)
 
+    if batch_size == 1:
+        # trtllm kernel requires max_q_len to be the same as the seqlen of the query when batch_size=1
```
yzh119: Why `qo_indptr[-1]` could be different to `s_qo`, is it because we want to be compatible with cudagraphs and `s_qo` will always be the maximum length?

bkryu: Short answer is yes.

Longer answer: In a `batch_size > 1` situation, the CUDA graph containing `prefill.trtllm_batch_context_with_kv_cache()` can be reused with multiple sequence lengths but not when `batch_size==1`. For example, 
* If batch_size is 3 and we have two batches with query lengths `[100, 200, 300]` and `[16, 500, 1024]`, we can set `s_qo=1024`, when we construct the CUDA graph and use the same CUDA graph for the two batches.
* However for batch_size=1, where we have batches of query lengths `[100]` and `[1024]`, a CUDA graph must be constructed each time -- first with `s_qo=100` and second with `s_qo=1024`.

Not sure whether the above is a real concern at the framework level. Nevertheless, `s_qo` goes in as the `max_q_len` [input argument](https://docs.flashinfer.ai/generated/flashinfer.prefill.trtllm_batch_context_with_kv_cache.html) where it is the *max sequence length for query*. We may at least want to consider whether the wording in the documentation is clear 😄

yzh119:
Hi @bkryu does upgrading to latest trtllm-gen fixing the issue?

bkryu:
> Hi @bkryu does upgrading to latest trtllm-gen fixing the issue?

Hi @yzh119, I'm currently checking. Upgrading to the latest trtllm-gen does fix the batch size 1 unit test, but I am seeing some errors in otherplaces. Will verify what is happening before marking the PR as ready

yzh119 (approved):
LGTM

yzh119:
Failed UT are not relevant and are scheduled to be fixed in #1950 and #1953

bkryu:
@nvmbreughe , can I get a review on the PR? Zihao and Perkz already approved, but due to code owner review requirements, it seems like I need a review from you

nvmbreughe (approved):
LGTM!

PR #1922, opened 2025-10-13
add blockwise gemm cute dsl

Amir-19:
## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive Python tests for blockwise GEMM validating multiple dtypes, layouts, alignment, iterations, and conditional hardware skips to ensure correctness.
  * Added FP8 W8A8 block GEMM benchmarking and comparison tooling to exercise multiple backends, measure TFLOP/s, validate correctness across providers, and record benchmark results for tuning.


kaixih:
Thx for the PR. Can I have the motivation of the blockwise gemm? like what kind of blockwise gemm is this (deepseek style or mxfp8/nvfp4). And how does it compare to the deepgemm and cutlass gemm in terms of performance? @Amir-19

Amir-19:
> Thx for the PR. Can I have the motivation of the blockwise gemm? like what kind of blockwise gemm is this (deepseek style or mxfp8/nvfp4). And how does it compare to the deepgemm and cutlass gemm in terms of performance? @Amir-19

yes this is DeepSeek style blockwise gemm. the implementation should follow cutlass cpp implementation closely. the main motivation is to put the ground work for gemm + all reduce:https://github.com/flashinfer-ai/flashinfer/pull/2007

flashinfer/cute_dsl/blockwise_gemm.py
```diff
@@ -0,0 +1,2970 @@
...
+    b_major: str,
+    c_major: str,
+    ab_dtype: Type[cutlass.Numeric],
+    sf_dtype: Type[cutlass.Numeric],
+    c_dtype: Type[cutlass.Numeric],
+    acc_dtype: Type[cutlass.Numeric],
+    use_2cta_instrs: bool,
+    mma_tiler_mn: Tuple[int, int],
+    cluster_shape_mn: Tuple[int, int],
+    sm_count: int,
```
yzh119: this value doesn't have to be part of the compilation arguments

yzh119 (commented):
Overall LGTM

flashinfer/cute_dsl/blockwise_gemm.py
Amir-19: removed `sm_count`.

flashinfer/cute_dsl/blockwise_gemm.py
```diff
@@ -0,0 +1,2955 @@
...
+    sfa_torch: torch.Tensor,
+    b_torch: torch.Tensor,
+    sfb_torch: torch.Tensor,
+    c_torch: torch.Tensor,
+    *,
+    ab_dtype: str,
+    sf_dtype: str,
+    c_dtype: str,
+    acc_dtype: str,
+    **kwargs,
```
kaixih, line 2915: Is this intended to be a user-facing interface, or will it eventually be called internally by something like `fp8_mm`?
I’m asking because we may want to follow a consistent naming convention for the API and its parameters. (@yzh119 do we have one?)
I understand that the backend will eventually add more features such as communication, but it would still be good to have a well-defined interface at this stage.
Also, adding a docstring to clarify the behavior of this function would be helpful.

Amir-19: good point! I think we can remove `ab_dtype`, `sf_dtype`, and `c_dtype` and infer them inside this function. for `acc_dtype`, I'm not sure if we can use a predefined logic or it should be explicitly specified.

PR #1947, opened 2025-10-19
chore: agentic workflow for automatic version bump

yzh119:
## 📌 Description

Add a new GitHub Actions workflow that automatically determines semantic version bumps using Google Gemini AI to analyze commit messages.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Chores**
  * Added an automated version-bump workflow (manual or scheduled) that analyzes recent commits, proposes semantic version bumps, includes AI reasoning and key changes in PRs, supports manual override, and prevents duplicate bump PRs.
* **New Features**
  * Added automated tagging and release workflow that creates a tag and triggers a release when a merged PR has the version-bump label.
  * PRs now include a generated, user-facing body with current/new version, bump type, AI reasoning, key changes, and semantic version guidance.

### DIFF
--- a/scripts/ai_determine_version_bump.py
+++ b/scripts/ai_determine_version_bump.py
@@ -94,6 +94,53 @@ def get_commits_since_tag(tag: str) -> list[dict]:
         return []
 
 
+def build_analysis_prompt(commits_summary: str, current_version: str) -> str:
+    """Build the AI analysis prompt (shared by all AI providers)."""
+    return f"""You are analyzing git commits for a CUDA kernel library called FlashInfer to determine the appropriate semantic version bump.
+
+Current version: {current_version}
+
+Versioning rules for this project (from CONTRIBUTING.md):
+FlashInfer follows a "right-shifted" versioning scheme (major.minor.patch[.post1]):
+- MAJOR increment: architectural milestones and/or incompatible API changes (breaking changes to public APIs), similar to PyTorch 2.0
+- MINOR increment: significant backwards-compatible new features (major functionality additions)
+- PATCH increment: small backwards-compatible features (e.g. new kernels, new SM support, etc.) and backwards-compatible bug fixes
+- POST (e.g. .post1): optional suffix for quick follow-up release with just backwards-compatible bug fixes (not used in this analysis)
+
+Here are the commits since the last release:
+
+{commits_summary}
+
+Please analyze these commits and determine:
+1. Whether there are any breaking API changes or architectural milestones (MAJOR bump needed)
+2. Whether there are significant new features (MINOR bump needed)
+3. Whether there are small features or bug fixes (PATCH bump needed)
+4. If no significant changes, return "none"
+
+Respond in JSON format:
+{{
+    "bump_type": "major|minor|patch|none",
+    "reasoning": "Detailed explanation of your decision",
+    "key_changes": ["list of most important changes that influenced the decision"]
+}}
+
+Important considerations:
+- New kernel implementations, new SM support, performance improvements are PATCH-level changes
+- MINOR bumps are for significant/major feature additions only, not incremental improvements
+- Internal refactoring, test updates, documentation changes alone don't warrant a version bump
+- API signature changes or removed functionality are MAJOR bumps
+- Focus on changes that affect users of the library, not internal changes
+"""
+
+
+def extract_json_from_response(text: str) -> str:
+    """Extract JSON from response that might be wrapped in markdown code blocks."""
+    json_match = re.search(r"```(?:json)?\s*(\{.*?\})\s*```", text, re.DOTALL)
+    if json_match:
+        return json_match.group(1)
+    return text
+
+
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-tag-and-release.yml
```diff
@@ -0,0 +1,98 @@
+name: Auto Tag and Release
+
+on:
+  pull_request:
```
nvmbreughe, line 4: I would be careful with automatically tagging. We want to make sure the tagged commit has a clean internal CI.

nvmbreughe: We could read a summary of the CI report and map it to the commit.

scripts/ai_determine_version_bump.py
```diff
@@ -0,0 +1,564 @@
...
+                        "files_changed": files_changed,
+                    }
+                )
+
+        return commits
+    except subprocess.CalledProcessError:
+        return []
+
+
+def analyze_with_openai(commits_summary: str, current_version: str) -> dict:
```
nvmbreughe: I would provide some more descriptive names for these analyze_* functions. Like version_bump_from_commit_analysis_openai.

scripts/ai_determine_version_bump.py
```diff
@@ -0,0 +1,564 @@
...
+    if not api_key:
+        raise ValueError("OPENAI_API_KEY not set")
+
+    model = os.getenv("OPENAI_MODEL", "gpt-4o")
+
+    from openai import OpenAI
+
+    client = OpenAI(api_key=api_key)
+
+    prompt = f"""You are analyzing git commits for a CUDA kernel library called FlashInfer to determine the appropriate semantic version bump.
```
nvmbreughe: This prompt is duplicated across the different models. I would suggest to assign it to a variable, or put it into a separate file.

nvmbreughe: actually these functions all look very similar. Maybe we can just have one function and pass in the model as a paramerer?

.github/workflows/auto-tag-and-release.yml
bkryu: The automatic tagging seems to happen if and only if a PR with a `version-bump` label gets merged, which should only get merged after the CI has been reviewed. The PR seems to remove the additional step of manually tagging & releasing.

Usually a version bump PR like [1921](https://github.com/flashinfer-ai/flashinfer/pull/1921) is more of an empty PR that updates version.txt -- the internal CI should have been triggered on the PR and the report should have already been reviewed before checking the PR in.

scripts/ai_determine_version_bump.py
```diff
@@ -0,0 +1,564 @@
...
+
+    # Fallback to basic analysis
+    print(
+        "Warning: No AI providers available, falling back to basic analysis",
+        file=sys.stderr,
+    )
+    return fallback_analysis(commits)
+
+
+def fallback_analysis(commits: list[dict]) -> dict:
```
bkryu, line 320: The fallback feels quite forced with regex matches. I wonder whether we realistically believe that a version bump PR opened based on this fallback logic will be any useful. It will require manual intervention to look at the actual commits that have been made.

scripts/ai_determine_version_bump.py
```diff
@@ -0,0 +1,564 @@
...
+        result_text = json_match.group(1)
+
+    return json.loads(result_text)
+
+
+def analyze_with_ai(commits: list[dict], current_version: str) -> dict:
+    """
+    Use AI to analyze commits and determine version bump.
+
+    Tries providers in order: OpenAI -> Claude -> Gemini -> Fallback
```
bkryu, line 266: If the cost of the API calls are not expensive, would the quality of response be better with an ensemble method? 

i.e. we make analyze with all three AIs (or up to three if not available), and make them vote on the best version number to bump to

scripts/ai_determine_version_bump.py
```diff
@@ -0,0 +1,564 @@
...
+
+Respond in JSON format:
+{{
+    "bump_type": "major|minor|patch|none",
+    "reasoning": "Detailed explanation of your decision",
+    "key_changes": ["list of most important changes that influenced the decision"]
+}}
+
+Important considerations:
+- Internal refactoring, test updates, documentation changes alone don't warrant a version bump
```
bkryu: Maybe also add benchmark code updates in list of things that don't warrant a version bump?

.github/workflows/auto-tag-and-release.yml
nvmbreughe: Agreed, but I believe the thinking is that we can have PRs in between the version bump and the actual tagged version, for agility reasons.

scripts/ai_determine_version_bump.py
yzh119: For public CI, I only enabled gemini at this moment (which is free under concern request rate), agreed voting could be better but I guess it's not a very complicated task for LLM at the moment.

PR #1954, opened 2025-10-20
Feature: Support Relu2 activation in fused MoE

amirkl94:
## 📌 Description
Added support for Relu2 activation in cutlass fp8 FusedMoE path.
`Relu2(x) = Relu(x)^2`.

Validated this works correctly on H100 and B200.

## Summary by CodeRabbit

* **New Features**
  * Added Relu2 as a selectable activation across MOE operations and exposed activation_type configuration to public MOE APIs and runner interfaces (Swiglu remains the default).
* **Behavior**
  * Certain GEMM execution paths now explicitly reject Relu2 and raise a clear runtime error instead of silently failing.

### DIFF
--- a/csrc/nv_internal/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_gemm_template_dispatch.h
+++ b/csrc/nv_internal/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_gemm_template_dispatch.h
@@ -927,6 +927,9 @@ void MoeGemmRunner<T, WeightType, OutputType, ScaleBiasType>::moeGemmBiasAct(
     case ActivationType::Geglu:
       runGemm<cutlass_extensions::EpilogueOpDefaultFtGelu>(inputs, hopper_inputs);
       break;
+    case ActivationType::Relu2:
+      TLLM_THROW("Relu2 is not supported.");
+      break;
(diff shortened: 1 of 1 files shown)


flashinfer/fused_moe/core.py
```diff
@@ -73,6 +73,19 @@ class RoutingMethodType(IntEnum):
     Unspecified = 6
 
 
+# Copied from csrc/nv_internal/tensorrt_llm/kernels/cutlass_kernels/include/common.h
```
djns99, line 76: Any chance we could add this to the python bindings instead of duplicating this manually?
Not sure if there is a good way to do this for enums, but it would be nicer if we could keep these in sync

amirkl94: I agree with this but there's an issue I'm not sure how to solve:
For the bindings to be present, we need to load the compiled binaries, and this is done lazily  from `get_cutlass_fused_moe_module()`. This means we won't be able to expose it from the python package and then this enum will be redundant.
Do you have a suggestion on how to solve this? If so I'd try and apply to `RoutingMethodType` as well as I think it has the same issue.

yzh119: one option could be making these enum classes not just-in-time loaded. we can do that in a future PR.

PR #1969, opened 2025-10-23
feat: enable deepgemm jit for fp8 block-scale on SM90

djmmoss:
## 📌 Description

Enable JIT compile for the FP8 DeepGEMM kernels, NVRTC is currently disabled it uses NVCC by default.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Refactor**
  * JIT include directory discovery now uses the flashinfer-python package instead of the previous package.
  * Updated resolved include path to the flashinfer data location.
  * Runtime compilation now consistently uses NVCC; the prior environment-variable toggle was removed.
  * Updated warning text when the expected package installation cannot be found.

### DIFF
--- a/csrc/nv_internal/tensorrt_llm/deep_gemm/compiler.cuh
+++ b/csrc/nv_internal/tensorrt_llm/deep_gemm/compiler.cuh
@@ -178,7 +178,7 @@ std::vector<std::filesystem::path> getJitIncludeDirs() {
                               "nv_internal" / "tensorrt_llm");
       }
     } else {
-      TLLM_LOG_WARNING("Failed to find TensorRT LLM installation, DeepGEMM will be disabled.");
+      TLLM_LOG_WARNING("Failed to find FlashInfer installation, DeepGEMM will be disabled.");
     }
   }
   return includeDirs;


csrc/nv_internal/tensorrt_llm/deep_gemm/compiler.cuh
```diff
@@ -125,7 +125,7 @@ std::vector<std::filesystem::path> getJitIncludeDirs() {
   static std::vector<std::filesystem::path> includeDirs;
   if (includeDirs.empty()) {
     // Command to execute
-    char const* cmd = "pip show tensorrt_llm 2>/dev/null";
+    char const* cmd = "pip show flashinfer-python 2>/dev/null";
```
yzh119, line 128: What's the purpose of this command?

csrc/nv_internal/tensorrt_llm/deep_gemm/compiler.cuh
```diff
@@ -174,15 +174,11 @@ std::vector<std::filesystem::path> getJitIncludeDirs() {
...
-        if (!kJitUseNvcc) {
-          includeDirs.push_back(std::filesystem::path(location) / "tensorrt_llm" / "include" /
-                                "cuda" / "include");
-        }
+        includeDirs.push_back(std::filesystem::path(location) / "flashinfer" / "data" / "csrc" /
+                              "nv_internal" / "tensorrt_llm");
       }
     } else {
-      TLLM_LOG_WARNING("Failed to find TensorRT LLM installation, DeepGEMM will be disabled.");
+      TLLM_LOG_WARNING("Failed to find FlashInfer installation, DeepGEMM will be disabled.");
```
yzh119, line 181: I guess we can safely assume flashinfer is installed if this function is called?

yzh119 (approved):

csrc/nv_internal/tensorrt_llm/deep_gemm/compiler.cuh
djmmoss: For the DeepGEMM JIT, it needs the header files in deep_gemm/, this command finds the installation path which is then used further down to add the deep_gemm/ to the -I

yzh119: I tend to move the logic to python, `pip show flashinfer-python` doesn't necessarily show the correct package information (e.g. at AOT time when the package is not installed yet).

yzh119: Or we can obtain the include path from python and pass the value to C++.

djmmoss: I think this is where a refactor might be necessary, unfortunately these deep_gemm kernels aren't captured as part of AOT.

PR #1973, opened 2025-10-23
Feature: Add support for L40 FusedMoE in cutlass path

amirkl94:
## 📌 Description
Fixed a few compilation issues for L40, and removed 1 gemm tactic for `sm == 89` that crashes due to:
```
Assertion failed: GPU lacks the shared memory resources to run GroupedGEMM kernel
```

## 🧪 Tests

Ran `pytest tests/moe/test_trtllm_cutlass_fused_moe.py` manually on an L40 GPU and verified all tests passed.

## Summary by CodeRabbit

* **New Features**
  * Official support for SM89 target: build/JIT flags and a public generation path to target it.

* **Bug Fixes / Compatibility**
  * Clarified FP8/FP4 dispatch: FP8 paths enabled for SM89; FP4 usage remains gated and now requires explicit enablement.

* **Performance**
  * Adjusted kernel/tile selection order for certain FP8 paths to prefer SM89-optimized options.

* **Chores**
  * Reduced logging severity for failed tactic profiling to warn/debug.

### DIFF
--- a/csrc/nv_internal/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_tma_warp_specialized_traits.h
+++ b/csrc/nv_internal/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_tma_warp_specialized_traits.h
@@ -49,12 +49,15 @@ template <typename T, typename WeightType,
               TmaWarpSpecializedGroupedGemmInput::EpilogueFusion::NONE>
 constexpr bool isValidBlackwellMOESpecialisation() {
 #if defined(CUTLASS_ARCH_MMA_SM100_SUPPORTED)  // TODO Is there a better choice
-  return (cutlass::platform::is_same<T, WeightType>::value
+  return (cutlass::platform::is_same<T, WeightType>::value ||
 #if defined(ENABLE_FP4)
-          || (cutlass::platform::is_same<T, __nv_fp8_e4m3>::value &&
-              cutlass::platform::is_same<WeightType, __nv_fp4_e2m1>::value))
+          (cutlass::platform::is_same<T, __nv_fp8_e4m3>::value &&
+           cutlass::platform::is_same<WeightType, __nv_fp4_e2m1>::value)
+#else
+          false
 #endif
-         && cutlass::platform::is_same<EpilogueTag, cutlass_extensions::EpilogueOpDefault>::value &&
+              ) &&
+         cutlass::platform::is_same<EpilogueTag, cutlass_extensions::EpilogueOpDefault>::value &&
          Fusion == TmaWarpSpecializedGroupedGemmInput::EpilogueFusion::NONE;
 #else
   return false;  // CUTLASS_ARCH_MMA_SM100_SUPPORTED is set when Blackwell kernels are enabled


csrc/nv_internal/tensorrt_llm/kernels/cutlass_kernels/cutlass_heuristic.cpp
```diff
@@ -158,7 +158,14 @@ std::vector<CutlassTileConfig> get_candidate_tiles(
...
     case CutlassGemmType::Fp8:
       if (config_type_param & CutlassGemmConfig::GROUPED_GEMM) {
-        if (sm == 89 || sm >= 120) {
+        if (sm == 89) {
+          return {CutlassTileConfig::CtaShape32x128x64_WarpShape32x32x64,
+                  CutlassTileConfig::CtaShape64x128x64_WarpShape64x32x64,
+                  CutlassTileConfig::CtaShape64x64x128_WarpShape32x64x64,
+                  CutlassTileConfig::CtaShape128x64x64_WarpShape64x32x64,
+                  CutlassTileConfig::CtaShape128x256x64_WarpShape64x64x64,
+                  CutlassTileConfig::CtaShape256x128x64_WarpShape64x64x64};
```
djns99: Are there any SM89 GPUs that can support the `CtaShape16x256x128_WarpShape16x64x128`, or is this an SM120 addition?
Having a small M value like this helps with low latency cases, so I'd want to understand why its not supported before disabling it

djns99: At the very least can you leave a comment saying what the difference between the two lists are, so people dont have to manually compare the items

yzh119:
There are some compilation errors on CI, @amirkl94 would you mind taking a look?
https://ci.tlcpack.ai/blue/organizations/jenkins/flashinfer-ci/detail/PR-1973/4/pipeline

csrc/nv_internal/tensorrt_llm/kernels/cutlass_kernels/cutlass_heuristic.cpp
amirkl94: > Are there any SM89 GPUs that can support the `CtaShape16x256x128_WarpShape16x64x128`, or is this an SM120 addition? Having a small M value like this helps with low latency cases, so I'd want to understand why its not supported before disabling it

This is a removal from the sm89 path. When I tested it on an `L40` GPU I got `Assertion failed: GPU lacks the shared memory resources to run GroupedGEMM kernel` .
It might be that on other sm89 GPUs it will pass, the main issue is that this was the default tactic that was chosen when trying to use `FusedMoE`, I believe that moving it to be the last will also fix my issue.

amirkl94:
> There are some compilation errors on CI, @amirkl94 would you mind taking a look? https://ci.tlcpack.ai/blue/organizations/jenkins/flashinfer-ci/detail/PR-1973/4/pipeline

Pushed a commit to fix this

csrc/nv_internal/tensorrt_llm/kernels/cutlass_kernels/cutlass_heuristic.cpp
amirkl94: I tried moving this tile config to be the last one and now the default tactic won't fail on l40. The issue is that if `autotuner` is on then the tactics that use this tile config will report an error with a stacktrace which looks bad.
@yzh119 do you think it'll be ok to change the errors that happen in the autotuner to be debug logs? Otherwise it means users will get spammed with error messages when they run autotuning on L40 FusedMoE.

csrc/nv_internal/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_tma_warp_specialized_traits.h
```diff
@@ -47,7 +48,8 @@ template <typename T, typename WeightType,
           TmaWarpSpecializedGroupedGemmInput::EpilogueFusion Fusion =
               TmaWarpSpecializedGroupedGemmInput::EpilogueFusion::NONE>
 constexpr bool isValidBlackwellMOESpecialisation() {
-#if defined(CUTLASS_ARCH_MMA_SM100_SUPPORTED)  // TODO Is there a better choice
+#if defined(CUTLASS_ARCH_MMA_SM100_SUPPORTED) && \
+    defined(ENABLE_FP4)  // TODO Is there a better choice
```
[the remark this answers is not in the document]
amirkl94: The issue here is:
```
cutlass::platform::is_same<WeightType, __nv_fp4_e2m1>::value
```
Which I think isn't defined unless ENABLE_FP4 is defined.

csrc/nv_internal/tensorrt_llm/kernels/cutlass_kernels/cutlass_heuristic.cpp
djns99: I think the autotuner should still output warnings, but just make them say "Skipping tactic x due to error. This tactic may not be supported by the current GPU architecture".
That said I know there is a difference of opinion on whether we should proactively filter them as you have done here, the argument being that we should be able to do the due diligence to determine what tactics are supported so that we can raise an error when a tactic fails when it shouldn't. So I can see either side.

amirkl94: I agree. I think that since this PR is somewhat time critical I'd rather move the faulty tile config to be the last so as not to remove it, and change the log to be a warning as you suggested and maybe a debug log with the stacktrace.
I'll also open an issue to take a deeper look into regarding if this specific tile config is relevant for sm89 or not as I think it might take me some time.
@djns99 , @yzh119 Does that sound ok to you?

djns99: I'm happy with that resolution

yzh119 (approved):
LGTM, should be ready to merge once CI passed.

nvmbreughe:
LGTM

Issue #1722, opened 2025-09-18
[Feature Request] "auto" backend for mm_fp4


sricketts:
`mm_fp4` currently does not have an `auto` backend, so the user has to select it manually. The ask is to add an `auto` backend with:

- logic to select the best backend based on problem shape and target device
- some way to tell the user which backend was selected, so the user can address (for example) weight shuffling requirements (this was input from @nvpohanh)

bkryu:
PR 1979 with the `mm_fp4(backend='auto')` option has been checked in yesterday.

PR #1113, opened 2025-06-04
Add CUTLASS fused moe kernels from TensorRT-LLM.

wenscarl:
## 🔍 Related Issues
Supported data types are: fp32/bf16/fp16/float8_e4m3fn/float8_e2m1
The kernels also support expert/tensor parallellism.

This PR also exposes quantization methods for nvfp4.

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [ ] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [ ] I have installed the hooks with `pre-commit install`.
- [ ] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [ ] Tests have been added or updated as needed.
- [ ] All tests are passing (`unittest`, etc.).

PR #1979, opened 2025-10-25
feat: Add backend='auto' to mm_fp4 and enable autotune for backend='cudnn'

bkryu:
## 📌 Description
Current PR:
* Introduces an `auto` backend to `mm_fp4` that can be autotuned. **It replaces `cudnn` as the default.**
  * Implementation matches `bmm_fp8`'s auto backend support.
* Allows `cudnn` backend to be autotuned.
* Added unit test test cases for backend=auto
 
Behavior of `auto` backend:
* Examines CUDA version & cuDNN version and calls either `cutlass` or `cudnn` kernel backends. `trtllm` kernel is not considered due to a non-interchangeable interface with other backends.
  * `auto` backend therefore only supports inputs runnable by `cutlass` and/or `cudnn.
* Non-autotuned behavior: 
  * Constructs an ordered list of backends (cudnn, cutlass) or (cutlass, cudnn) where ordering is based on previous microbenchmark study results.
    * If CUDA 12 --> cutlass comes to front.
    * If CUDA 13 and cuDNN version < 9.15 --> cutlass comes front
    * If CUDA 13 and cuDNN version >= 9.15 --> cudnn comes front
  * If kernel is not available from a support check, it is removed from the list.
* Autotune behavior:
  * If backend is explicitly provided --> Autotunes within the backend. Same as previous behavior, but now autotuning is supported for cudnn.
  * If `backend='auto'` --> Autotunes within and across backends (cudnn & cutlass) and chooses the best config of best backend. `trtllm` kernel is not considered 
* A lot of helper functions to `mm_fp4` were refactored to enable cross-backend autotuning. Refactoring was done to match cross-backend autotune-enabled `bmm_fp8` as a reference. 

### Pytest outputs
`pytest tests/gemm/test_mm_fp4.py`
* SM100 (B200) CUDA 13 & cuDNN 9.15: `900 passed, 2532 skipped in 125.19s (0:02:05)`
* SM100 (B200) CUDA 12 & cuDNN 9.15: `900 passed, 2532 skipped in 125.67s (0:02:05)`
* SM120 (RTX 5090) CUDA 13 & cuDNN 9.15: `720 passed, 2712 skipped in 76.50s (0:01:16)`

### Example microbenchmark outputs:
On SM100 (B200) CUDA 13 & cuDNN 9.15
```
flashinfer/benchmarks# python3 flashinfer_benchmark.py --routine mm_fp4 --m 1024 --n 7168 --k 4608 --out_dtype bfloat16 --backends cudnn cutlass trtllm auto --use_128x4_sf_layout --use_nvfp4 --refcheck
[PERF] cudnn          :: median time 0.018 ms; std 0.000 ms; achieved tflops 3797.932 TFLOPs/sec; achieved tb_per_sec 1.884 TB/sec
[PERF] cutlass        :: median time 0.020 ms; std 0.000 ms; achieved tflops 3440.640 TFLOPs/sec; achieved tb_per_sec 1.707 TB/sec
[PERF] trtllm         :: median time 0.031 ms; std 0.000 ms; achieved tflops 2187.427 TFLOPs/sec; achieved tb_per_sec 1.085 TB/sec
[PERF] auto           :: median time 0.018 ms; std 0.000 ms; achieved tflops 3840.714 TFLOPs/sec; achieved tb_per_sec 1.905 TB/sec
/flashinfer/benchmarks# python3 flashinfer_benchmark.py --routine mm_fp4 --m 1024 --n 7168 --k 4608 --out_dtype bfloat16 --backends cudnn cutlass trtllm auto --use_128x4_sf_layout --refcheck
[INFO] cutlass backend does not support this configuration: ValueError: Only cudnn and auto FP4 GEMM supports mxfp4 quantization.
[INFO] trtllm backend does not support this configuration: ValueError: Only cudnn and auto FP4 GEMM supports mxfp4 quantization.
[PERF] cudnn          :: median time 0.021 ms; std 0.000 ms; achieved tflops 3238.249 TFLOPs/sec; achieved tb_per_sec 1.606 TB/sec
[PERF] auto           :: median time 0.021 ms; std 0.000 ms; achieved tflops 3237.753 TFLOPs/sec; achieved tb_per_sec 1.606 TB/sec median time 0.009 ms; std 0.000 ms; achieved tflops 938.356 TFLOPs/sec; achieved tb_per_sec 2.069 TB/sec

## Autotune
/flashinfer/benchmarks# python3 flashinfer_benchmark.py --routine mm_fp4 --m 1024 --n 7168 --k 4608 --out_dtype bfloat16 --backends cudnn cutlass trtllm auto --use_128x4_sf_layout --use_nvfp4 --refcheck --autotune
2025-11-11 23:43:23,715 - INFO - autotuner.py:256 - flashinfer.jit: [Autotuner]: Autotuning process starts ...
2025-11-11 23:43:25,789 - INFO - autotuner.py:262 - flashinfer.jit: [Autotuner]: Autotuning process ends
2025-11-11 23:43:25,790 - INFO - autotuner.py:256 - flashinfer.jit: [Autotuner]: Autotuning process starts ...
2025-11-11 23:43:26,251 - INFO - autotuner.py:262 - flashinfer.jit: [Autotuner]: Autotuning process ends
2025-11-11 23:43:26,251 - INFO - autotuner.py:256 - flashinfer.jit: [Autotuner]: Autotuning process starts ...
2025-11-11 23:43:26,327 - INFO - autotuner.py:262 - flashinfer.jit: [Autotuner]: Autotuning process ends
2025-11-11 23:43:26,327 - INFO - autotuner.py:256 - flashinfer.jit: [Autotuner]: Autotuning process starts ...
2025-11-11 23:43:26,335 - INFO - autotuner.py:262 - flashinfer.jit: [Autotuner]: Autotuning process ends
[PERF] cudnn_autotune :: median time 0.016 ms; std 0.000 ms; achieved tflops 4129.171 TFLOPs/sec; achieved tb_per_sec 2.048 TB/sec
[PERF] cutlass_autotun:: median time 0.019 ms; std 0.000 ms; achieved tflops 3513.845 TFLOPs/sec; achieved tb_per_sec 1.743 TB/sec
[PERF] trtllm_autotune:: median time 0.026 ms; std 0.000 ms; achieved tflops 2613.338 TFLOPs/sec; achieved tb_per_sec 1.296 TB/sec
[PERF] auto_autotune  :: median time 0.016 ms; std 0.000 ms; achieved tflops 4128.768 TFLOPs/sec; achieved tb_per_sec 2.048 TB/sec

/flashinfer/benchmarks# python3 flashinfer_benchmark.py --routine mm_fp4 --m 1024 --n 7168 --k 4608 --out_dtype bfloat16 --backends cudnn cutlass trtllm auto --use_128x4_sf_layout --refcheck --autotune
[INFO] cutlass backend does not support this configuration: ValueError: Only cudnn and auto FP4 GEMM supports mxfp4 quantization.
[INFO] trtllm backend does not support this configuration: ValueError: Only cudnn and auto FP4 GEMM supports mxfp4 quantization.
2025-11-11 23:43:37,942 - INFO - autotuner.py:256 - flashinfer.jit: [Autotuner]: Autotuning process starts ...
2025-11-11 23:43:43,116 - INFO - autotuner.py:262 - flashinfer.jit: [Autotuner]: Autotuning process ends
2025-11-11 23:43:43,116 - INFO - autotuner.py:256 - flashinfer.jit: [Autotuner]: Autotuning process starts ...
2025-11-11 23:43:43,124 - INFO - autotuner.py:262 - flashinfer.jit: [Autotuner]: Autotuning process ends
[PERF] cudnn_autotune :: median time 0.020 ms; std 0.000 ms; achieved tflops 3370.154 TFLOPs/sec; achieved tb_per_sec 1.672 TB/sec
[PERF] auto_autotune  :: median time 0.020 ms; std 0.000 ms; achieved tflops 3370.692 TFLOPs/sec; achieved tb_per_sec 1.672 TB/sec
```

On SM100 (B200) CUDA 12 & cuDNN 9.15
```
flashinfer/benchmarks# python3 flashinfer_benchmark.py --routine mm_fp4 --m 1024 --n 7168 --k 4608 --out_dtype bfloat16 --backends cudnn cutlass trtllm auto --use_128x4_sf_layout --use_nvfp4 --refcheck
[PERF] cudnn          :: median time 0.023 ms; std 0.001 ms; achieved tflops 2975.898 TFLOPs/sec; achieved tb_per_sec 1.476 TB/sec
[PERF] cutlass        :: median time 0.020 ms; std 0.000 ms; achieved tflops 3370.423 TFLOPs/sec; achieved tb_per_sec 1.672 TB/sec
[PERF] trtllm         :: median time 0.031 ms; std 0.000 ms; achieved tflops 2187.427 TFLOPs/sec; achieved tb_per_sec 1.085 TB/sec
[PERF] auto           :: median time 0.020 ms; std 0.000 ms; achieved tflops 3371.229 TFLOPs/sec; achieved tb_per_sec 1.672 TB/sec
(py312) root@84ef83abb1b5:/flashinfer/benchmarks# python3 flashinfer_benchmark.py --routine mm_fp4 --m 1024 --n 7168 --k 4608 --out_dtype bfloat16 --backends cudnn cutlass trtllm auto --use_128x4_sf_layout --refcheck
[INFO] cutlass backend does not support this configuration: ValueError: Only cudnn and auto FP4 GEMM supports mxfp4 quantization.
[INFO] trtllm backend does not support this configuration: ValueError: Only cudnn and auto FP4 GEMM supports mxfp4 quantization.
[PERF] cudnn          :: median time 0.021 ms; std 0.000 ms; achieved tflops 3238.249 TFLOPs/sec; achieved tb_per_sec 1.606 TB/sec
[PERF] auto           :: median time 0.021 ms; std 0.000 ms; achieved tflops 3238.249 TFLOPs/sec; achieved tb_per_sec 1.606 TB/sec

## Autotune
/flashinfer/benchmarks# python3 flashinfer_benchmark.py --routine mm_fp4 --m 1024 --n 7168 --k 4608 --out_dtype bfloat16 --backends cudnn cutlass trtllm auto --use_128x4_sf_layout --use_nvfp4 --refcheck --autotune
2025-11-11 23:42:43,378 - INFO - autotuner.py:256 - flashinfer.jit: [Autotuner]: Autotuning process starts ...
2025-11-11 23:42:45,451 - INFO - autotuner.py:262 - flashinfer.jit: [Autotuner]: Autotuning process ends
2025-11-11 23:42:45,451 - INFO - autotuner.py:256 - flashinfer.jit: [Autotuner]: Autotuning process starts ...
2025-11-11 23:42:45,910 - INFO - autotuner.py:262 - flashinfer.jit: [Autotuner]: Autotuning process ends
2025-11-11 23:42:45,910 - INFO - autotuner.py:256 - flashinfer.jit: [Autotuner]: Autotuning process starts ...
2025-11-11 23:42:45,986 - INFO - autotuner.py:262 - flashinfer.jit: [Autotuner]: Autotuning process ends
2025-11-11 23:42:45,986 - INFO - autotuner.py:256 - flashinfer.jit: [Autotuner]: Autotuning process starts ...
2025-11-11 23:42:45,993 - INFO - autotuner.py:262 - flashinfer.jit: [Autotuner]: Autotuning process ends
[PERF] cudnn_autotune :: median time 0.021 ms; std 0.000 ms; achieved tflops 3190.355 TFLOPs/sec; achieved tb_per_sec 1.583 TB/sec
[PERF] cutlass_autotun:: median time 0.019 ms; std 0.000 ms; achieved tflops 3551.330 TFLOPs/sec; achieved tb_per_sec 1.762 TB/sec
[PERF] trtllm_autotune:: median time 0.026 ms; std 0.000 ms; achieved tflops 2621.440 TFLOPs/sec; achieved tb_per_sec 1.300 TB/sec
[PERF] auto_autotune  :: median time 0.019 ms; std 0.000 ms; achieved tflops 3551.628 TFLOPs/sec; achieved tb_per_sec 1.762 TB/sec
flashinfer/benchmarks# python3 flashinfer_benchmark.py --routine mm_fp4 --m 1024 --n 7168 --k 4608 --out_dtype bfloat16 --backends cudnn cutlass trtllm auto --use_128x4_sf_layout --refcheck --autotune
[INFO] cutlass backend does not support this configuration: ValueError: Only cudnn and auto FP4 GEMM supports mxfp4 quantization.
[INFO] trtllm backend does not support this configuration: ValueError: Only cudnn and auto FP4 GEMM supports mxfp4 quantization.
2025-11-11 23:42:55,176 - INFO - autotuner.py:256 - flashinfer.jit: [Autotuner]: Autotuning process starts ...
2025-11-11 23:42:58,600 - INFO - autotuner.py:262 - flashinfer.jit: [Autotuner]: Autotuning process ends
2025-11-11 23:42:58,601 - INFO - autotuner.py:256 - flashinfer.jit: [Autotuner]: Autotuning process starts ...
2025-11-11 23:42:58,608 - INFO - autotuner.py:262 - flashinfer.jit: [Autotuner]: Autotuning process ends
[PERF] cudnn_autotune :: median time 0.021 ms; std 0.000 ms; achieved tflops 3238.249 TFLOPs/sec; achieved tb_per_sec 1.606 TB/sec
[PERF] auto_autotune  :: median time 0.021 ms; std 0.000 ms; achieved tflops 3238.249 TFLOPs/sec; achieved tb_per_sec 1.606 TB/sec
```

On SM120 (RTX 5090) CUDA 13 & cuDNN 9.15
```
/flashinfer/benchmarks$ python3 flashinfer_benchmark.py --routine mm_fp4 --m 1024 --n 7168 --k 4608 --out_dtype bfloat16 --backends cudnn cutlass trtllm auto --use_128x4_sf_layout --use_nvfp4 --refcheck
[INFO] trtllm backend does not support this configuration: BackendSupportedError: mm_fp4 does not support backend 'trtllm' with capability 120
[PERF] cudnn          :: median time 0.058 ms; std 0.000 ms; achieved tflops 1167.143 TFLOPs/sec; achieved tb_per_sec 0.579 TB/sec
[PERF] cutlass        :: median time 0.060 ms; std 0.000 ms; achieved tflops 1135.056 TFLOPs/sec; achieved tb_per_sec 0.563 TB/sec
[PERF] auto           :: median time 0.058 ms; std 0.000 ms; achieved tflops 1158.952 TFLOPs/sec; achieved tb_per_sec 0.575 TB/sec
/flashinfer/benchmarks$ python3 flashinfer_benchmark.py --routine mm_fp4 --m 1024 --n 7168 --k 4608 --out_dtype bfloat16 --backends cudnn cutlass trtllm auto --use_128x4_sf_layout --refcheck
[INFO] cutlass backend does not support this configuration: ValueError: Only cudnn and auto FP4 GEMM supports mxfp4 quantization.
[INFO] trtllm backend does not support this configuration: BackendSupportedError: mm_fp4 does not support backend 'trtllm' with capability 120
[PERF] cudnn          :: median time 0.054 ms; std 0.000 ms; achieved tflops 1241.735 TFLOPs/sec; achieved tb_per_sec 0.616 TB/sec
[PERF] auto           :: median time 0.054 ms; std 0.000 ms; achieved tflops 1241.735 TFLOPs/sec; achieved tb_per_sec 0.616 TB/sec
```

## 🔍 Related Issues

#1722 

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * "auto" backend selection for FP4 matrix ops; command-line backend option now accepts "auto"
  * cuDNN, CUTLASS and TRTLLM offered as selectable FP4 GEMM backends; backend choice guided by CUDA/cuDNN heuristics

* **Improvements**
  * Runtime capability checks replace static backend lists; unsupported backends are pruned dynamically
  * Heuristic-driven auto selection required for automatic mode; autotune/warmup expanded and validation tolerance relaxed

* **Tests**
  * Tests extended to cover auto-backend scenarios and relaxed constraints

### DIFF
--- a/tests/utils/test_decorators.py
+++ b/tests/utils/test_decorators.py
@@ -344,7 +344,27 @@ def test_suitable_auto_backends():
     def _cudnn_check(x, backend):
         return x.shape[0] > 5
 
-    @backend_requirement({"cutlass": _cutlass_check, "cudnn": _cudnn_check})
+    # When using an auto backend, some heuristic function must exist
+    def _heuristic_func(suitable_backends, x, backend):
+        candidate_backends = None
+        if x.shape[0] > 5:
+            candidate_backends = ["cudnn", "cutlass"]
+        else:
+            candidate_backends = ["cutlass", "cudnn"]
+
+        heuristic_backends = []
+        for backend in candidate_backends:
+            if backend in suitable_backends:
+                heuristic_backends.append(backend)
+        return heuristic_backends
+
+    @backend_requirement(
+        backend_checks={
+            "cutlass": _cutlass_check,
+            "cudnn": _cudnn_check,
+        },
+        heuristic_func=_heuristic_func,
+    )
     def my_kernel(x, backend="auto"):
         backends = my_kernel.suitable_auto_backends
         if x.shape[0] > 5:


benchmarks/routines/gemm.py
```diff
@@ -823,7 +823,7 @@ def testMmFp4(args):
             print(
                 "[INFO] cutlass backend does not support mxfp4 quantization (use_nvfp4=False)"
             )
-            backends.remove("cutlass")
+            remove_cutlass = True
```
nvmbreughe: Another way to avoid these remove_backend_x bools is to call the related backend check (which should be annoated with the decorator), or have the decorator return a filtered list as I proposed. https://github.com/flashinfer-ai/flashinfer/pull/2000#discussion_r2474225984

flashinfer/gemm/gemm_base.py
```diff
@@ -1942,70 +2093,128 @@ def mm_fp4(
...
-            cudnn.data_type.FP4_E2M1,
-            _torch_data_type_to_cudnn_data_type(out_dtype),
-            block_size,
-            a.device,
-            alpha is not None,
-            use_nvfp4,
-        )
+    # Auto-select the best backend
+    if backend == "auto":
+        cuda_major, _ = get_cuda_version(a.device)
```
nvmbreughe: These checks should be part of the _auto_gemm_fp4_requirement check.

I think a cleaner way would be to move the generation of the list of candidate_backends in the @backend_requirement decorator, where "auto" backend is treated specially. It lists the required checks for each backend already. An alternative is that we create a separate decorator that composes and uses the backend checks of the  backend_requirement

nvmbreughe: The danger here is that we may be repeating some checks, but not all of them.

flashinfer/gemm/gemm_base.py
```diff
@@ -1942,70 +2093,128 @@ def mm_fp4(
...
+        # If cuda version is 13 or greater:
+        # cudnn is more performant if cudnn version is 9.14 or greater.
+        if cuda_major >= 13 and cudnn.backend_version() >= 91400:
+            candidate_backends = ("cudnn", "cutlass")
+        # Otherwise, prioritize cutlass
+        else:
+            candidate_backends = ("cutlass", "cudnn")
+
+        # Filter to only supported backends for this compute capability
+        # Note: The requirement function already validated that at least one backend is supported
```
nvmbreughe: So this is the dangerous part: at this point, we know 1 backend replied that its check is ok. But we are considering all backends. Maybe cudnn supports it but not trtllm or cutlass.

flashinfer/gemm.py
```diff
@@ -1942,70 +2093,128 @@ def mm_fp4(
...
+        # Filter to only supported backends for this compute capability
+        # Note: The requirement function already validated that at least one backend is supported
+        backends = []
+        for candidate in candidate_backends:
+            # mypy requires explicit type casting for the backend literal
+            backend_literal = cast(
+                Literal["cudnn", "trtllm", "cutlass", "auto"], candidate
+            )
+            try:
+                _check_mm_fp4_problem_size(
```
nvmbreughe: we shouldn't have to do this here.

flashinfer/gemm.py
```diff
@@ -1942,70 +2093,128 @@ def mm_fp4(
...
+        else:
+            candidate_backends = ("cutlass", "cudnn")
+
+        # Filter to only supported backends for this compute capability
+        # Note: The requirement function already validated that at least one backend is supported
+        backends = []
+        for candidate in candidate_backends:
+            # mypy requires explicit type casting for the backend literal
+            backend_literal = cast(
+                Literal["cudnn", "trtllm", "cutlass", "auto"], candidate
```
nvmbreughe: why is auto added back?

flashinfer/gemm.py
```diff
@@ -1942,70 +2093,128 @@ def mm_fp4(
...
-            b_descale = b_descale.view(torch.uint8)
-
-        # Dispatch to the correct module based on device architecture
-        major, _ = get_compute_capability(a.device)
-        if major == 12:
-            gemm_module = get_gemm_sm120_module_cutlass_fp4()
+    # At this point, backends contains a supported backend if specified, or all supported backends if backend='auto'.
+    runners = []
+    for cur_backend in backends:
+        if cur_backend == "cudnn":
```
nvmbreughe: Is there a cleaner way than the if-then-else structure? Maybe some dictionary that maps between backend and runner retrieval?

flashinfer/gemm.py
```diff
@@ -1942,70 +2093,128 @@ def mm_fp4(
...
+            runners.append(_cudnn_gemm_fp4_runner())
+        elif cur_backend == "trtllm":
+            runners.append(
+                get_trtllm_fp4_gemm_module().trtllm_fp4_gemm_runner(use_8x4_sf_layout)
+            )
+        elif cur_backend == "cutlass":
+            if a.dtype == torch.uint8 and a_descale.dtype == torch.float8_e4m3fn:
+                a_descale = a_descale.view(torch.uint8)
```
nvmbreughe: This seems like an implementation detail, and maybe needs to be moved to the cutlass runner itself, just like we do with the cudnn_runner.

benchmarks/routines/gemm.py
nvmbreughe: Regardless whether you stuff it into the decorator, this will be a pattern that will happen for all APIs, so we should think about encapsulating the "if backend and checks_dont_pass: filter_it_out".

flashinfer/gemm/gemm_base.py
```diff
@@ -1942,70 +2093,128 @@ def mm_fp4(
...
         else:
-            gemm_module = get_gemm_sm100_module_cutlass_fp4()
+            # Should not reach this
+            raise ValueError(f"Unsupported backend: {cur_backend}")
 
-        gemm_module.cutlass_fp4_gemm(
-            a, b.T, a_descale, b_descale.T, alpha, out, workspace_buffer
-        )
+    # Now we have a list of runners for desired & supported backends.
+    tuner = AutoTuner.get()
```
nvmbreughe, line 2134: I think it is great that we unify the autotuning logic, and separating the different runner implementations.

flashinfer/gemm.py
bkryu: Agree and this allows removal of the if-then-else structure above. Updated in latest commit

flashinfer/gemm.py
bkryu: It is a good idea. Updated in latest commit to use a dictionary

flashinfer/gemm.py
bkryu: Auto is actually not being added here since the `cast()` is telling pre-commit tests that `backend_literal` will be one of `["cudnn", "trtllm", "cutlass", "auto"]` while `candidate_backends` will never contain `auto`. 

However, there is no need for `auto` to be there and I can see it being confusing so I have removed in the latest commit

flashinfer/gemm/gemm_base.py
bkryu: When writing the code path for this PR, I noted that the following questions had to be answered at different times by the auto backend logic:
1. Is there at least one runnable backend for the given input params -- for early error raising
2. What are the runnable backends for the given input params -- to consider which backends to choose from
3. In the current GPU/CUDA/cuDNN environment, what is the preferred ordering of backends -- for heuristics

The current implementation in the PR answers 1 in  `@backend_requirement` and 2 & 3 in the body of the `mm_fp4` while you're suggesting putting 2 inside  `@backend_requirement`. I agree that this helps us avoid repeating checks but this will involve--as you raised--a special treatment for the auto backend and a change to `backend_requirement`. We can discuss

flashinfer/gemm.py
bkryu: I'm finding that we actually do because the result of `_check_mm_fp4_problem_size` depends on the backend. 

With mxfp4 cases, for example, we need to check each backend so that we can keep cudnn and reject all other backends.

flashinfer/gemm/gemm_base.py
bkryu: You are correct here. In the latest commit, I now check whether the backend is supported generally + for the inputs

benchmarks/routines/gemm.py
bkryu: This is a good idea. I have removed these hard coded checks entirely and have started using the checkers in the latest

bkryu: Now this is addressed with the latest decorator update #2029

flashinfer/gemm/gemm_base.py
```diff
@@ -93,6 +93,10 @@ def _match_sm_version(device: torch.device, sm_version: list[str]):
     return device_arch in sm_version
 
 
+def get_cuda_version(device: torch.device):
+    return tuple(map(int, torch.version.cuda.split(".")))  # (major, minor)
+
```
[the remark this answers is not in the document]
bkryu: Fixed in latest commit

flashinfer/gemm/gemm_base.py
bkryu: Now this is addressed with the latest decorator update #2029

flashinfer/gemm/gemm_base.py
bkryu: Now this is addressed with the latest decorator update #2029

flashinfer/gemm/gemm_base.py
```diff
@@ -1845,19 +1961,60 @@ def _cutlass_gemm_fp4_requirement(
...
+        "cudnn": _cudnn_gemm_fp4_requirement,
         "cutlass": _cutlass_gemm_fp4_requirement,
-    },
-    common_check=_check_mm_fp4_problem_size,  # Shape checks common to all backends
+        # Does not consider trtllm due to different interface.
+    }
+
+    for candidate in candidate_backends:
+        checker = backend_checkers[candidate]
+        if hasattr(
```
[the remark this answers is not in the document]
bkryu: Deleted in refactor

flashinfer/utils.py
```diff
@@ -1028,9 +1028,20 @@ def suitable_auto_backends(cc, *args, **kwargs):
...
+                    if (
+                        req_checker(*args, **kwargs)
+                        and req_checker.is_compute_capability_supported(cc)
+                        and (
+                            (common_check is None)
+                            or common_check(
+                                *args, backend=backend, **kwargs_without_backend
+                            )
+                        )
+                    ):
```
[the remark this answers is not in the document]
bkryu: req_checker does not use the backend variable

flashinfer/utils.py
```diff
@@ -1028,9 +1028,20 @@ def suitable_auto_backends(cc, *args, **kwargs):
             for backend in backend_checks:
                 req_checker = backend_checks[backend]
                 try:
-                    if req_checker(
-                        *args, **kwargs
-                    ) and req_checker.is_compute_capability_supported(cc):
+                    # Remove 'backend' from kwargs to explicitly pass it to the common check.
```
nvmbreughe: This seems dangerous. Now it is possible that common_check will use it's default value for backend (e.g., cudnn) while you are checking another (e.g., trtllm).

flashinfer/utils.py
```diff
@@ -1028,9 +1028,20 @@ def suitable_auto_backends(cc, *args, **kwargs):
...
+                    kwargs_without_backend = {
+                        k: v for k, v in kwargs.items() if k != "backend"
+                    }
+                    if (
+                        req_checker(*args, **kwargs)
+                        and req_checker.is_compute_capability_supported(cc)
+                        and (
+                            (common_check is None)
+                            or common_check(
```
nvmbreughe: We already ran common_check in line 1024

flashinfer/gemm/gemm_base.py
```diff
@@ -1677,6 +1654,145 @@ def mm_fp8(
...
+                b_descale,
+                alpha,
+                out_dtype,
+                out,
+                block_size,
+                use_nvfp4,
+                workspace_buffer,
+            ) = inputs
+
+            real_a_shape, real_a_stride = _get_real_fp4_shape_from_packed_uint8(a)
```
nvmbreughe: Seems to be the same like _cudnn_gemm_fp4 does. Why we need it twice?

flashinfer/gemm/gemm_base.py
```diff
@@ -1930,70 +2089,80 @@ def mm_fp4(
...
-            gemm_module = get_gemm_sm120_module_cutlass_fp4()
-        else:
-            gemm_module = get_gemm_sm100_module_cutlass_fp4()
+    backend_to_runner_factory = {
+        "cudnn": lambda: _cudnn_gemm_fp4_runner(),
+        "trtllm": lambda: get_trtllm_fp4_gemm_module().trtllm_fp4_gemm_runner(
+            use_8x4_sf_layout
+        ),
+        "cutlass": lambda: get_gemm_sm120_module_cutlass_fp4().cutlass_fp4_gemm_runner()
+        if major == 12
```
nvmbreughe: [nit] This feels a bit weird. It's not a heuristic, but also not a compatibility check. Perhaps we should have a wrapper function around this: cutlass_fp4_gemm() that branches for the multiple SMs.

flashinfer/utils.py
bkryu: Yes, we did for backend='auto', but not for each backend, and we do need for each backend when constructing the `suitable_backends` array

The result of mm_fp4's common_check depends on the backend ([see these lines](https://github.com/bkryu/flashinfer/blob/5fd8e7641ca4e19faef95e83ba32af546f95e41b/flashinfer/gemm/gemm_base.py#L1844-L1847))
```
    if backend != "trtllm" and use_8x4_sf_layout:
        raise ValueError("Only TRTLLM FP4 GEMM supports 8x4 scale factor layout.")
    if backend not in ["cudnn", "auto"] and not use_nvfp4:
        raise ValueError("Only cudnn and auto FP4 GEMM supports mxfp4 quantization.")
```
For mxfp8 cases, without checking here, the common check will say cutlass is a viable option to pick where in reality it is not.

flashinfer/utils.py
bkryu: Hi Max, in this part of the code we are in `suitable_auto_backends()` which gets called only when backend='auto', so I believe the mentioned risk does not apply.

`kwargs_without_backend` is used only `common_check` for each backend considered for auto.

nvmbreughe: common_check actually takes a look at "backend" also in some cases: e.g., _check_mm_fp4_problem_size:

``` 
if backend != "trtllm" and use_8x4_sf_layout:
        raise ValueError("Only TRTLLM FP4 GEMM supports 8x4 scale factor layout.")
```

Maybe that is not a good practice and we should add these to the backend specific checks. If that is the case we should orthogonally enforce that common_check should not have the 'backend' parameter. This is, imho, better practice than manipulating the kwargs here.

flashinfer/gemm/gemm_base.py
bkryu: This is a good point, the code is mostly repeated and I should refactor here.

Context is: In this function, I am constructing graph to extract the number of execution plans only, where `_cudnn_gemm_fp4` additionally executes the graph. I will create e.g. a `get_cudnn_fp4_gemm_graph()` to reduce redundancy.

flashinfer/gemm/gemm_base.py
bkryu: Thanks I see why you are saying this is odd. Let me see if I can refactor this.

flashinfer/utils.py
bkryu: Yes, hence we need to run common_check for each backend that can be considered by the auto backend, and we need to do a substitution -- otherwise, we only check against backend='auto'

> If that is the case we should orthogonally enforce that common_check should not have the 'backend' parameter. This is, imho, better practice than manipulating the kwargs here.

I agree with you on this point. However, the current decorator apparatus does not have a way to check this  because `is_backend_supported` is agnostic to the inputs. Would you suggest adding another function to the decorator like `input_backend_check` that checks whether the provided inputs are supported for a specific backend?

nvmbreughe: Good point: we need
* is_backend_supported agnostic to inputs (to quickly gauge CC and backend support)
* is_shape_supported to see whether there are any mathematical issues with the inputs
* some combination where we check whether the backend's support surface covers the shape. This is now part of is_shape_support, which is a little bit of an abuse in a sense. 


Given that, I would leave the decorator as is.

nvmbreughe: Also your changes here look good to me. I would however add a comment to explain this.

flashinfer/utils.py
jimmyzho: If we move the backend-specific checks in `common_check` to their respective backend checkers, we would then not need to call `common_check` for every backend in this function, right?

flashinfer/utils.py
jimmyzho: > I agree with you on this point. However, the current decorator apparatus does not have a way to check this because is_backend_supported is agnostic to the inputs. Would you suggest adding another function to the decorator like input_backend_check that checks whether the provided inputs are supported for a specific backend?

Just to clarify, is this to be specific to APIs with 'auto' option, or across the board? I thought `is_backend_supported` is already agnostic to inputs, it only checks for existence of backend and cc support.

> is_shape_supported to see whether there are any mathematical issues with the inputs

This check could be covered (plus support surface) already by the individual backend-specific checker functions (_cudnn_xxx_requirement, ...) as it already looks into the inputs

I'm not sure if we're on the same page here, please let me know if this makes sense

flashinfer/gemm/gemm_base.py
bkryu: Refactored in latest commit

flashinfer/gemm/gemm_base.py
bkryu: Refactored in the latest commit 😃

flashinfer/utils.py
bkryu: Yes this is actually a good point. cc @nvmbreughe

tests/gemm/test_mm_fp4.py
```diff
@@ -105,5 +94,38 @@ def test_mm_fp4(
             pytest.fail(str(e))
 
 
+# TODO: Consdier splitting this function up for the various backends
```
aleozlx, line 97: what does that address?

bkryu: Short answer: My guess is maybe to reduce the total number of test cases including those that get skipped?

I can only guess because I copied over the comment when I was doing the refactor

flashinfer/gemm/gemm_base.py
aleozlx: on a related note, i wish we can have a uniform standard to test per op per backend about the followings: 1. top1, 
2. topN-autotuing (and show which tactic failed), 
3. specific tactic, 
4. sweep across all kernels (optional, if kernels are ported from elsewhere and already tested)

flashinfer/gemm/gemm_base.py
```diff
@@ -406,87 +410,50 @@ def fp8_gemm_sm100(
...
-            [fp4_runner],
-            tuning_config,
-            inputs,
-        )
+            def forward(
+                self,
+                inputs: List[torch.Tensor],
+                tactic: int = -1,
+                do_preparation: bool = False,
+                **kwargs,
```
aleozlx, line 427: i wonder what are the kwargs used for

flashinfer/gemm/gemm_base.py
```diff
@@ -1829,19 +1968,60 @@ def _cutlass_gemm_fp4_requirement(
...
 @backend_requirement(
     {
-        "cudnn": _cudnn_gemm_fp4_requirement,  # Each backend has its own requirement function
+        "cudnn": _cudnn_gemm_fp4_requirement,
         "trtllm": _trtllm_gemm_fp4_requirement,
         "cutlass": _cutlass_gemm_fp4_requirement,
     },
-    common_check=_check_mm_fp4_problem_size,  # Shape checks common to all backends
+    common_check=_check_mm_fp4_problem_size,
+    heuristic_func=_heuristic_func_mm_fp4,  # result stored in mm_fp4.suitable_auto_backends
```
aleozlx, line 2025: i find heuristic_func to be undocumented
https://github.com/flashinfer-ai/flashinfer/blob/main/flashinfer/utils.py#L859

how to make sense of 
1. heuristic_fun being a "backend_requirement" 
2. and what it is expected to do (requirements on the actual function passed to it) and 
3. the interaction it may have with backend arg 
4. and in what scenarios such interactions happen

we need to be more clear so it doesn't run away to be just another arbitrary code injection point

flashinfer/gemm/gemm_base.py
```diff
@@ -1829,19 +1968,60 @@ def _cutlass_gemm_fp4_requirement(
...
+    alpha: Optional[torch.Tensor] = None,
+    out_dtype: torch.dtype = torch.bfloat16,
+    out: Optional[torch.Tensor] = None,
+    block_size: int = 16,
+    use_8x4_sf_layout: bool = False,
+    backend: Literal["cudnn", "trtllm", "cutlass", "auto"] = "cudnn",
+    use_nvfp4: bool = True,
+):
+    r"""
+    Heuristic function for mm_fp4 backend selection. Routes to either cudnn or cutlass, but not trtllm.
```
aleozlx: why? what is defining such behavior is ok vs not ok?

flashinfer/gemm/gemm_base.py
bkryu: No use in this specific instance. However, included because the parent class `TunableRunner` defines the [forward function](https://github.com/flashinfer-ai/flashinfer/blob/main/flashinfer/autotuner.py#L220-L226) with `**kwargs`

flashinfer/gemm/gemm_base.py
bkryu: Generally it is highly undesired, but in this case we discussed that it is inevitable without over-engineering  a solution. 

As also mentioned in the description of the PR, `backend='trtllm'` uses a different API definition from the other two backends. This means if we change the backend from cudnn or cutlass to trtllm, mm_fp4's output changes. Moreover, when inside of mm_fp4, it is impossible to determine whether the input was prepared for `backend='trtllm'` in mind or not -- we have painted ourselves into a corner when first writing mm_fp4. Given this situation, we discussed and converged among us and with frameworks developers that an auto option that only chooses between cudnn or cutlass because (1) vLLM and SGLang currently use cudnn or cutlass unless an env variable that explicitly requires trtllm is set; (2) cudnn + cutlass generally outperforms trtllm in most test cases; and (3) we do not want a major overhaul of the mm_fp4 API

bkryu: However, I agree that I provide a bit more context in the comments here 😅. Will add in the next commit

flashinfer/gemm/gemm_base.py
bkryu: Agree with your comment. @nvmbreughe, @jimmyzho, and I have reached an amorphous consensus on how the function should behave, but we should document it. Will update in next commit

flashinfer/gemm/gemm_base.py
bkryu: @aleozlx agree that this could be nice to have. Would be nice if we can come up with a design and proceed it from there

flashinfer/gemm/gemm_base.py
bkryu: Added comments in `utils.py` that explain the expected behavior of `heuristic_func`

flashinfer/gemm/gemm_base.py
bkryu: Added a bit more context on why trtllm is not considered in comment.

flashinfer/utils.py
```diff
@@ -880,6 +880,12 @@ def backend_requirement(
         backends. Should accept the same arguments as the decorated function and return
         True if requirements are met, False otherwise.
         In the case where the kernel function does not have any specific backends, this can be decorated with @supported_compute_capability to specify the function's supported compute capabilities.
+    heuristic_func : callable, optional
+        An optional function that performs heuristic backend selection when backend is "auto". Does not do anything if backend is not "auto".
```
aleozlx: is it a bug on FI if this is missing but user tries "auto", is there defined behavior there?

flashinfer/gemm/gemm_base.py
aleozlx: great. added another comment there
and leaving this open for a little in case others wanna comment

flashinfer/gemm/gemm_base.py
aleozlx: right and the abstract class does not seem do document that
i wonder if we wanna revisit that and understand what it is for (even if not in this PR)
it didn't exist until this https://github.com/flashinfer-ai/flashinfer/pull/1113
oh it's from trtllm and also no documentation there... 
https://github.com/NVIDIA/TensorRT-LLM/blob/04fb481da3c0b8fda68375d0f2ce4cdb6551b02d/tensorrt_llm/_torch/autotuner.py#L187-L200

i wonder if this is fundamental enough infra that we wanna maintain or plan to keep syncing with them
i feel the reason cannot just be "the code next to/above it look like this too"

i'll gather this in a doc somewhere then resolve it
cc @nvmbreughe

tests/gemm/test_mm_fp4.py
aleozlx: ok will ask more offline

flashinfer/utils.py
bkryu: Yes, if an API decorated with `backend_requirement`  has a backend='auto' option, but it is missing the heuristic function, it would be a bug.
(Limiting to "decorated with `backend_requirement` because some APIs like attention still are not wrapped. There is ongoing work to do so)

bkryu: However, to explicitly address your question, if a heuristic function is missing the behavior word be falling back to the ordering of backends in [backend_checks.keys()](https://github.com/bkryu/flashinfer/blob/2fa33877529df1b294ee9c0adceeadcefe8e844b/flashinfer/utils.py#L1034-L1049)

flashinfer/gemm/gemm_base.py
aleozlx: asked ppl working with trtllm, informal answer it's to supply hints to the tuning process (eg. which part to tune or other conditions)

documented code ownership question, will move this discussion off of this review

flashinfer/gemm/gemm_base.py
aleozlx: moving to design doc

flashinfer/utils.py
aleozlx: should we put in an assertion "an API decorated with backend_requirement has a backend='auto' option, then should have a heuristic_func"

aleozlx:
looks good ✅  no other comments from me

flashinfer/utils.py
jimmyzho: Yes, if backend is set to "auto" but no heuristic function is passed in, suitable_auto_backends() will return the list of compatible backends in the order that @bkryu described. We should decide whether this behaviour is acceptable, or add some default heuristic function?

benchmarks/routines/gemm.py
```diff
@@ -904,6 +868,34 @@ def run_backend(backend):
                 backend=backend,
                 use_nvfp4=use_nvfp4,
             )
+        except Exception as e:
+            print(
+                f"[INFO] {backend} backend does not support this configuration: {type(e).__name__}: {e}"
+            )
+            backends_to_remove.append(backend)
```
jimmyzho, line 875: Is this try-except block trying to verify whether each backend supports the input configuration? I'm thinking whether it would be a good idea to expose the _is_problem_size_supported API in the decorator so that we can avoid doing something like this, and make a call directly to query this. E.g `flashinfer.gemm.mm_fp4.is_problem_size_supported(args...)`

bkryu: This is an interesting suggestion, but I also wonder whether it is a common use case. Essentially what we are doing here is reverse engineering what backends are supported for my inputs. 

For a framework running end-to-end inference, this would not be a use case so I'd say we probably don't need to worry about it. If anything, we should advocate using the "auto" backend and deliver good backend selection accuracy, which could be done through enabling autotuning.

For a benchmarking script like the current one, I'd see the use case of `is_problem_size_supported`, but I'd claim that the current flow is too bad because the exception is raised very early before the kernels are launched.

flashinfer/gemm/gemm_base.py
```diff
@@ -93,6 +93,10 @@ def _match_sm_version(device: torch.device, sm_version: list[str]):
     return device_arch in sm_version
 
 
+def get_cuda_version():
```
yzh119: You can try relying on this https://github.com/flashinfer-ai/flashinfer/blob/b9964cc30edd0ddac84e68ef2528eeefb59c96e2/flashinfer/jit/cpp_ext.py#L64

Another tricky issue is that cuda runtime version doesn't always match the nvcc version. Which one do you think we should rely on?

bkryu: Thanks this is a great point. I removed the reinvented `get_cuda_version` and imported the one you pointed to in the latest commit.

> Another tricky issue is that cuda runtime version doesn't always match the nvcc version. Which one do you think we should rely on?

Regarding this comment, it's hard to think or test in this case, because here we are only using the major version (13 vs. 12) to make any heuristic decisions. If the CUDA runtime version and nvcc mismatch in their major version that would be a more significant issue that we likely don't need to address here.

Regarding in the future though, when a case emerges in which a minor or patch version differs between CTK, runtime, and nvcc we will need to test and decide what makes sense

flashinfer/utils.py
aleozlx: i think an assert is lowerest cost solution if 99% we expect it to exist. contributors supply this function not the user.
and don't need to bother about what exactly happens in 1% just make it loud something bad is there
adding a default is ok but then you need to explain/document the default - more cost

flashinfer/gemm/gemm_base.py
aleozlx: i believe we try to check the respective version where the requirement is from instead of leaning into one or the other in cudnn for e.g.
secondarily perhaps we could give a warning when these two major version don't match (or a flag to explicitly enable such situation if user intended to do that to try something) and test is once in the beginning of the library, imo

benchmarks/routines/gemm.py
nvmbreughe: Actually, when I designed the decorator I had exactly in mind what Jimmy is nudging at: callling is_problem_size_supported before we even run the operation.

benchmarks/routines/gemm.py
```diff
@@ -904,6 +868,34 @@ def run_backend(backend):
...
+
+    # Remove unsupported backends
+    for backend in backends_to_remove:
+        backends.remove(backend)
+
+    if len(backends) == 0:
+        print("[ERROR] No backends passed validation. Exiting.")
+        return
+
+    def run_backend(backend):
```
nvmbreughe, line 885: [nit] Is the goal of this function to ensure DRY and avoid copy paste arguments with the parameters? If so another way is to use in your "for cur_backend in backends" on L911, `run_backend=flashinfer.gem.mm_fp4`

flashinfer/utils.py
nvmbreughe: Right seems like low cost to add the assert. Dev's will have examples to look at how a heuristic function should look like. We could add a unit test with a very simple heuristic function.

The danger is that we may inadvertently change the order by shuffling the order in the backend_checks dictionary. This would change auto behavior.

tests/gemm/test_mm_fp4.py
nvmbreughe: I believe this comment was added by me. The idea is that we now have a combinatorial space of parameters but only a subset is supported and the rest skipped. by splitting this up per tests we will 1) reduce the number of parameters and 2) the values those parameters support for the specific backend.

flashinfer/gemm/gemm_base.py
```diff
@@ -1687,7 +1824,7 @@ def _check_mm_fp4_problem_size(
     out: Optional[torch.Tensor] = None,
     block_size: int = 16,
     use_8x4_sf_layout: bool = False,
-    backend: Literal["cudnn", "trtllm", "cutlass"] = "cudnn",
+    backend: Literal["cudnn", "trtllm", "cutlass", "auto"] = "auto",
```
nvmbreughe: I guess now that we have factored out any checks on backend, we no longer need the parameter here?

flashinfer/gemm/gemm_base.py
```diff
@@ -1677,6 +1662,158 @@ def mm_fp8(
...
+    a_descale: torch.Tensor,
+    b_descale: torch.Tensor,
+    alpha: Optional[torch.Tensor] = None,
+    out_dtype: torch.dtype = torch.bfloat16,
+    out: Optional[torch.Tensor] = None,
+    block_size: int = 16,
+    use_nvfp4: bool = True,
+    tactic: int = -1,
+):
+    _check_cudnn_availability()
```
nvmbreughe: I could be wrong, but could we add this check to _cudnn_gemm_fp4 and avoid keep checking this?

flashinfer/gemm/gemm_base.py
```diff
@@ -1677,6 +1662,158 @@ def mm_fp8(
...
+                a_descale,
+                b_descale,
+                alpha,
+                out_dtype,
+                out,
+                block_size,
+                use_nvfp4,
+                workspace_buffer,
+            ) = inputs
+            _cudnn_gemm_fp4(
```
nvmbreughe, line 1800: Why not just call execute_cudnn_gemm_fp4_graph here?
We already have the graph from get_valid_tactics.

flashinfer/gemm/gemm_base.py
```diff
@@ -1677,6 +1662,158 @@ def mm_fp8(
...
+                b_descale,
+                alpha,
+                out_dtype,
+                out,
+                block_size,
+                use_nvfp4,
+                workspace_buffer,
+            ) = inputs
+
+            graph = _get_cudnn_fp4_gemm_graph(
```
nvmbreughe, line 1765: In general I am having a hard time to mentally put everything together with the runner design. I don't immediately see how we can avoid it, but it may be helpful to add some code comments for people who look at it for the first time:
* we need to build a cudnn graph, and cache it so any checks and execution can use the same one
* we need to build plans, and pick one of those
* we need to run the graph with some of the plans of before (execute_cudnn_gemm_fp4_graph) and use the cached graph.

benchmarks/routines/gemm.py
bkryu: I can see the initial appeal of this idea and how being able to call a support check seems nice at first glance, but I am still not able to see the use case. 

If we hypothetically support this, how do we expect frameworks to use it? Would they have a preferred ordering of backends in mind and try to check one by one, and pick the first one that succeeds? and then if no backend is supported maybe fallback to a non-FlashInfer choice?

My stance is:
* Frameworks should be encouraged to use the auto option for optimal backend selection, not checking them one by one
* If there is a support gap, we want the frameworks to come to us so we can fix it instead of falling back to some other kernel provider.

However, if I am missing something and there is a need for this, I would advocate for a followup PR, just to avoid the scope creep of this PR. The size of this PR has grown quite a bit since its inception and is becoming  a bit daunting 😓

benchmarks/routines/gemm.py
bkryu: Here it is more of a consistency with other APIs that can be benchmarked -- in the case of attention, the function name is actually different. For example, see [these lines](https://github.com/bkryu/flashinfer/blob/8e335ebfef45420db76ee94cfb0c3ffa54115979/benchmarks/routines/attention.py#L503-L535)

flashinfer/gemm/gemm_base.py
bkryu: The parameters will be unused but when we run these checks, we simply give all arguments via `backend_checks[backend](*args, **kwargs)` in [utils.py](https://github.com/bkryu/flashinfer/blob/8e335ebfef45420db76ee94cfb0c3ffa54115979/flashinfer/utils.py#L1056-L1060) so it's likely best to keep the function signature the same.

However, I can note that the backend parameter will be unused.

flashinfer/gemm/gemm_base.py
bkryu: It is already in `_cudnn_gemm_fp4_requirement` when it calls `_check_cudnn_fp4_availability`.

I am realizing that there are a few redundant checks, which is a good catch. Deleting them

flashinfer/gemm/gemm_base.py
bkryu: get_valid_tactics does not get called when not autotuning. In [choose_one function in autotuner.py](https://github.com/bkryu/flashinfer/blob/8e335ebfef45420db76ee94cfb0c3ffa54115979/flashinfer/autotuner.py#L433-L450), which is a bit tortuous to follow, we find that runner=0 tactic=-1  is picked automatically when not autotuning. More on this in the response to your comment below

flashinfer/gemm/gemm_base.py
bkryu: I agree with your sentiment here; it is a bit convoluted. I also don't see how we can easily clean up the runner design. The graph constructed only once during `_cudnn_gemm_fp4_requirement()` time with `@functools.cache`:

1. `_cudnn_gemm_fp4_requirement()` calls `create_cudnn_execution_plans_fp4_gemm()` to build and cach the graph.
2. When we call `get_valid_tactics()` and `_cudnn_gemm_fp4()` the cached graph should be retrieved.

Sprinkled a few comments to make this clear.

flashinfer/gemm/gemm_base.py
nvmbreughe: I think you're right. Though that breaks our previous thinking to make this function backend independent. One thing we could do is have the decorator explicitly remove the backend from kwargs when calling this function. This way we enforce it to not have a backend arg.... 
Not sure what is most complex: 

1. making the shape_check functions like these have the exact same function signature like the backend_check functions but with the implicit understanding that we should not be using the backend parameter in any of the conditions, or
2. making the shape_check function the exact same signature minus the backend param.

I may have slight preference for (1) how it is right now, but neither are perfect. CC+ @jimmyzho in case you have any thoughts on this

bkryu: In the latest, I at least annotated unused parameters with `#unused` in the function signature, but neither are indeed perfect.

benchmarks/routines/gemm.py
nvmbreughe: > However, if I am missing something and there is a need for this, I would advocate for a followup PR, just to avoid the scope creep of this PR. The size of this PR has grown quite a bit since its inception and is becoming a bit daunting 😓

Yes that is not a problem.

Just to provide context of the use case -- I think is_problem_size_supported itself is not as big of a use case. But the complete picture: backend + problem size is:

A DLFW may not want to use auto, and avoid any checks. They could have determined themselves which backend is most performant (perhaps using auto initially) and bake in which backends they want. They can then call the api with skip_check=True, to avoid overhead of any checks. CC+ @elfiegg

PR #2000, opened 2025-10-29
refactor: backend_requirement + supported_compute_capability decorator for gemm

jimmyzho:
## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Release Notes

* **New Features**
  * Added batch POD attention with paged KV cache support.
  * Added RoPE quantization with paged KV cache append functionality.
  * Extended MLA attention with LSE base configuration option.
  * Expanded Hopper (H100/H200) FMHA optimizations with GMMA support.

* **Improvements**
  * Enhanced FP8 quantization paths for grouped GEMM operations.
  * Improved input validation with explicit error handling for attention kernels.
  * Optimized device context management via RAII guards.

* **Documentation**
  * Updated GPU/CUDA support information and API logging guidance.
  * Expanded benchmark suite with FP8 and softmax performance measurements.

### DIFF
--- a/flashinfer/deep_gemm.py
+++ b/flashinfer/deep_gemm.py
@@ -45,7 +45,12 @@ from .artifacts import ArtifactPath
 from .cuda_utils import checkCudaErrors
 from .jit.cubin_loader import get_cubin
 from .jit.env import FLASHINFER_CUBIN_DIR
-from .utils import ceil_div, round_up
+from .utils import (
+    ceil_div,
+    round_up,
+    supported_compute_capability,
+    backend_requirement,
+)
 
 
 class GemmType(enum.Enum):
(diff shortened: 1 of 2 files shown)


flashinfer/deep_gemm.py
```diff
@@ -1358,24 +1363,27 @@ def m_grouped_fp8_gemm_nt_masked_sm10x(
     runtime(**all_kwargs)
 
 
-def m_grouped_fp8_gemm_nt_contiguous(
+@supported_compute_capability([100, 103])
```
nvmbreughe, line 1366: Ah, the compute capability will not be checked actually on the common_check function.
We would either need to:
1. Redesign the decorator so that we can just add @supported_compute_capability to  m_grouped_fp8_gemm_nt_contiguous directly; OR
2. Change the decorator's implementations of wrapper and is_compute_capability_supported so it also checks this on the common_check.
3. Add a backend parameter to m_grouped_fp8_gemm_nt_contiguous. Though this is deepgemm and I don't think we want to call this a backend for now.


My preference would go to option 2

flashinfer/deep_gemm.py
```diff
@@ -1423,6 +1464,72 @@ def m_grouped_fp8_gemm_nt_contiguous(
     impl(a, sfa, b, sfb, d, m_indices)
 
 
+@supported_compute_capability([100, 103])
```
nvmbreughe, line 1467: same here, the compute capabilities will be ignored.

flashinfer/gemm.py
```diff
@@ -2009,6 +2006,70 @@ def mm_fp4(
...
+    _validate_fp8_output_dtype(dtype)
+    return True
+
+
+@backend_requirement(
+    {
+        "cudnn": _cudnn_bmm_fp8_requirement,
+        "cublas": _cublas_bmm_fp8_requirement,
+        "cutlass": _cutlass_bmm_fp8_requirement,
+        "auto": _cublas_bmm_fp8_requirement,  # cublas default
```
nvmbreughe: auto doesn't default to cublas. It actually sets `backends = ["cutlass", "cublas", "cudnn"]` , this usually will be autotuned across backends.

That's not trivial to resolve. One way may be to just create an aggregate function "check_all_backends" that calls the 3 checker functions.

Note that the function that takes all these backends is fp8_gemm_sm100, which also does a bunch of checks. So perhaps we need to rely on those, since the checks here are very minimal anyway.

But this does remain a general issue: "How do we isolate backend support checks when the auto backend is chosen". I think the behavior that we wish is "filter out unsuitable backends", rather than just rejecting auto when one of the backends fails. Maybe this can be an additional "attribute" that the decorator adds for the special auto backend. To illustrate, we would have:
bmm_fp8.suitable_auto_backends() return ["cutlass", "cublas", "cudnn"] if all checks pass, or ["cutlass", "cublas"] if cudnn's check failed. bmm_fp8.suitable_auto_backends() would return None if the selected backend was not "auto".

flashinfer/deep_gemm.py
jimmyzho: I think 2 also makes most sense. In a lot of the APIs there are also no 'backend' arg to be passed in so we can't only check @ supported_compute_capability there. I can change this in a separate PR

nvmbreughe: A separate PR is fine, since we wouldn't cause a regression (we didn't have CC checks before the current PR anyway).

flashinfer/gemm.py
jimmyzho: > bmm_fp8.suitable_auto_backends() return ["cutlass", "cublas", "cudnn"] if all checks pass, or ["cutlass", "cublas"] if cudnn's check failed.

In this decorator, we could set an attribute with the final list of supported backends for the actual function to reference when passing into the autotuner, is this what you had in mind?

PR #2013, opened 2025-10-30
More realistic bench for POD Attn

Edenzzzz:
## 📌 Description

Use real head sizes, seq lens and add comparison with sequential prefill + decode.
Results on H100 (without overlap, which only adds ~150GB/s for persistent):
image
cc @yzh119 
## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Release Notes

* **New Features**
  * Added comprehensive performance benchmarking for batch attention operations with detailed timing measurements.
  * Introduced sequential dual-kernel benchmark path with extended memory bandwidth reporting.

* **Tests**
  * Updated benchmark test configurations to use deterministic, fixed values for improved reproducibility.
  * Adjusted benchmark parameters for consistency across test iterations.

### DIFF
--- a/benchmarks/bench_mixed_attention.py
+++ b/benchmarks/bench_mixed_attention.py
@@ -72,6 +72,24 @@ def run_bench(
     measurements = bench_gpu_time(lambda: wrapper_old.run(q, kv_data))
     ms_old = np.median(measurements)
 
+    wrapper_persistent = flashinfer.BatchAttention(kv_layout="NHD")
+    wrapper_persistent.plan(
+        q_indptr.to(device),
+        kv_indptr.to(device),
+        torch.arange(num_blocks, dtype=torch.int32, device=device),
+        seq_lens.to(device),
+        num_qo_heads,
+        num_kv_heads,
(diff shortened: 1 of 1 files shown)


Edenzzzz:
> I suppose the benefit of POD mainly coming from overlapping?
> 
> LGTM overall, we will revamp the OSS attention code in the coming release and let's check the performance later.

Yes, though when I tested overlap persistent kernel, the benefit is not as significant as in POD, which is confusing because it should achieve statically planned overlap instead of opportunistic, and the block launch/warmup overhead should be lower
Will make a bench/issue against FA3 later
d09a73481279ca3dc8d71e7aa970a5ab

912d8ba42e62c44405df133d7012cdad
image

Edenzzzz:
@yzh119 ncu profiling results show POD has less branching and higher memory throughput
image
image
image
[persistent_vs_pod.ncu-rep.zip](https://github.com/user-attachments/files/23266615/persistent_vs_pod.ncu-rep.zip)

PR #2015, opened 2025-10-30
Support cc common check decorator for empty backends

jimmyzho:
## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved backend/compute-capability validation with clearer errors and correct fallback when backend-specific checks are absent.

* **New Features**
  * Decorated functions expose runtime attributes to query backend availability and choices.
  * Default-backend behavior: kernels use a default when none is passed.

* **Compatibility**
  * Expanded supported compute-capability set and raised minimum cuDNN package requirements.

* **Tests**
  * Added tests for empty-backend common-checks and default-backend behavior.

* **Chores**
  * Version bumped to 0.5.1.

### DIFF
--- a/flashinfer/utils.py
+++ b/flashinfer/utils.py
@@ -975,13 +975,12 @@ def backend_requirement(
                 return False
 
         def is_compute_capability_supported(cc):
-            # Handle empty backend_checks case
+            # In case there is only 1 implicit backend, the compute capability support needs to be added to the common check
             if not has_backend_choices():
                 # No backend-specific checks, only check common_check
                 if not hasattr(common_check, "is_compute_capability_supported"):
                     raise ValueError(
-                        f"Invalid is_compute_capability_supported call: {common_check.__name__} \
-                        does not have is_compute_capability_supported decorator"
+                        f"Invalid is_compute_capability_supported call: {common_check.__name__} does not have is_compute_capability_supported decorator"
                     )
                 return common_check.is_compute_capability_supported(cc)
             else:
@@ -1018,7 +1017,7 @@ def backend_requirement(
         def has_backend(backend: str) -> bool:
             # Whether the given backend exists in the API
             return backend in backend_checks
-        
+
         # @brief: Wrapper function that calls the orignal, decorated function, after applying a number of checks.
         # @note that here we manually apply defaults to the arguments in the wrapper function when doing validation.
         @functools.wraps(func)
--- a/tests/utils/test_decorators.py
+++ b/tests/utils/test_decorators.py
@@ -138,18 +138,29 @@ def test_backend_requirement_empty_backends_with_common_check_cc():
     # Check compute capability support (only common_check)
     assert unsupported_kernel.is_compute_capability_supported(42) is True
     assert unsupported_kernel.is_compute_capability_supported(75) is False
-    # Undefined behaviour: is_backend_supported with no backend choices
-    with pytest.raises(
-        ValueError,
-        match="Invalid is_backend_supported call: no backend choices for unsupported_kernel",
-    ):
-        unsupported_kernel.is_backend_supported(None, 42)
-        unsupported_kernel.is_backend_supported(None, 75)
+
+    # The following tests are for when no backend choices are provided, where
+    # `is_backend_supported` is undefined behaviour and will raise error.
+    # We also enforce the `common_check` function when using `@backend_requirement` decorator.
+    # It must also be decorated with `@supported_compute_capability`.
+
+    # Raise error: is_backend_supported cannot be called with no backend choices.
+    for backend in [
+        ("random_backend", 42),
+        ("random_backend", 75),
+        (None, 42),
+        (None, 75),
+    ]:
+        with pytest.raises(
+            ValueError,
+            match="Invalid is_backend_supported call: no backend choices for unsupported_kernel",
+        ):
+            unsupported_kernel.is_backend_supported(backend[0], backend[1])
 
     # Test compute capability support during kernel runtime
     x = torch.randn(10, 10, device="cuda")
 
-    # Will always raise error because no real compute capability is supported
+    # Error: no real compute capability is supported
     with pytest.raises(
         BackendSupportedError, match="does not support compute capability"
     ):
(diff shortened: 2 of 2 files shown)


flashinfer/utils.py
```diff
@@ -973,6 +991,13 @@ def is_compute_capability_supported(cc):
...
         def is_problem_size_supported(*args, **kwargs):
             backend = kwargs.get("backend")
+
+            # Handle empty backend_checks case
+            if not backend_checks:
+                if common_check is not None:
+                    return common_check(*args, **kwargs)
+                return True  # No checks at all
+
             if backend not in backend_checks:
```
nvmbreughe, line 1003: This will raise since kwargs.get("backend") is None

tests/utils/test_decorators.py
```diff
@@ -115,6 +115,63 @@ def my_kernel(x, backend="cudnn"):
...
+        {},
+        common_check=_common_check,
+    )
+    def supported_kernel(x):
+        return x * 2
+
+    assert supported_kernel.is_compute_capability_supported(actual_capability) is True
+    assert supported_kernel.is_backend_supported(None, actual_capability) is True
+    assert (
+        supported_kernel.is_backend_supported("any_backend", actual_capability) is True
```
nvmbreughe: I think the desired behavior is to return False in this case, cause there isn't really a specific backend. We could add an additional ".has_backend()" to the decorator.

tests/utils/test_decorators.py
```diff
@@ -115,6 +115,63 @@ def my_kernel(x, backend="cudnn"):
...
+
+    @backend_requirement(
+        {},
+        common_check=_common_check,
+    )
+    def supported_kernel(x):
+        return x * 2
+
+    assert supported_kernel.is_compute_capability_supported(actual_capability) is True
+    assert supported_kernel.is_backend_supported(None, actual_capability) is True
```
nvmbreughe: This should be undefined behavior so a test is not relevant.

flashinfer/utils.py
```diff
@@ -951,6 +952,14 @@ def backend_requirement(
 
     def decorator(func):
         def is_backend_supported(backend, cc=None):
+            # No backend-specific checks
+            if not backend_checks:
```
nvmbreughe: I would return False in this case

flashinfer/utils.py
```diff
@@ -964,6 +973,15 @@ def is_backend_supported(backend, cc=None):
...
 
         def is_compute_capability_supported(cc):
+            # Handle empty backend_checks case
+            if not backend_checks:
+                # No backend-specific checks, only check common_check
+                if common_check is not None and hasattr(
+                    common_check, "is_compute_capability_supported"
+                ):
+                    return common_check.is_compute_capability_supported(cc)
+                return True  # No restrictions at all
```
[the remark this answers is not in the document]
nvmbreughe: You can disregard this comment if you agree with the earlier one I left on is_backend_supported

flashinfer/utils.py
```diff
@@ -973,6 +991,13 @@ def is_compute_capability_supported(cc):
 
         def is_problem_size_supported(*args, **kwargs):
             backend = kwargs.get("backend")
+
+            # Handle empty backend_checks case
```
nvmbreughe, line 999: I don't think this is needed here. I would suggest adding an extra statement to the bottom:
```
if common_check is not None and backend is not None:
....
elif common_check is not None and backend is None
....
elif backend is not None:
...
```

nvmbreughe (changes requested):

flashinfer/utils.py
jimmyzho: Is this line 1001? Here, I thought this would imply the there is a `backend_checks` and therefore kwargs.get("backend") will be non-null

tests/utils/test_decorators.py
```diff
@@ -115,6 +115,130 @@ def my_kernel(x, backend="cudnn"):
...
+    assert unsupported_kernel.is_compute_capability_supported(42) is True
+    assert unsupported_kernel.is_compute_capability_supported(75) is False
+    # Undefined behaviour: is_backend_supported with no backend choices
+    with pytest.raises(
+        ValueError,
+        match="Invalid is_backend_supported call: no backend choices for unsupported_kernel",
+    ):
+        unsupported_kernel.is_backend_supported(None, 42)
+        unsupported_kernel.is_backend_supported(None, 75)
+
```
nvmbreughe, line 159: [nit] maybe also check the output of this call for a string. I see you did this later also for "random_backend"

tests/utils/test_decorators.py
```diff
@@ -115,6 +115,130 @@ def my_kernel(x, backend="cudnn"):
...
+    assert supported_kernel.is_compute_capability_supported(actual_capability) is True
+    # Undefined behaviour: is_backend_supported with no backend choices
+    with pytest.raises(
+        ValueError,
+        match="Invalid is_backend_supported call: no backend choices for supported_kernel",
+    ):
+        supported_kernel.is_backend_supported(None, actual_capability)
+    assert supported_kernel.has_backend("random_backend") is False
+
+    result = supported_kernel(x)
```
nvmbreughe, line 194: [nit] I think this is the main desired behavior that this PR adds. I would suggest to point that out here with a comment.

flashinfer/utils.py
```diff
@@ -951,27 +952,47 @@ def backend_requirement(
...
+                return common_check.is_compute_capability_supported(cc)
+            else:
+                # True if any backend requirement supports this cc
+                return any(
+                    hasattr(checker, "is_compute_capability_supported")
+                    and checker.is_compute_capability_supported(cc)
+                    for checker in backend_checks.values()
+                )
 
         def is_problem_size_supported(*args, **kwargs):
```
nvmbreughe: Note this function was renamed yesterday.

nvmbreughe (approved):
LGTM

flashinfer/utils.py
```diff
@@ -951,27 +952,47 @@ def backend_requirement(
...
-            return False
 
         def is_compute_capability_supported(cc):
-            # True if any backend requirement supports this cc
-            return any(
-                hasattr(checker, "is_compute_capability_supported")
-                and checker.is_compute_capability_supported(cc)
-                for checker in backend_checks.values()
-            )
+            # Handle empty backend_checks case
```
nvmbreughe: I would rephrase the comment: "In case there is only 1 implicit backend, the compute capability support needs to be added to the common check".

PR #2029, opened 2025-11-03
feat: suitable_auto_backends to prune auto backends, bmm_fp8 refactor, heuristic_func intake

jimmyzho:
## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Improvements**
  * Expanded FP8 BMM backend support with explicit Cutlass SM10x/SM12x handling, safer fallbacks (no unconditional hard failures), and richer auto-selection that exposes viable backends and respects device capabilities.
  * Added heuristic-driven backend preference for auto and cutlass paths.

* **Refactor**
  * Backend gating reorganized into per-backend capability checks, a shared problem-size pre-check, and heuristic selection; decorator now exposes suitable_auto_backends and capability extraction.

* **Tests**
  * Added tests validating auto backend discovery and heuristic ordering.

### DIFF
--- a/flashinfer/utils.py
+++ b/flashinfer/utils.py
@@ -1020,14 +1020,17 @@ def backend_requirement(
             # Whether the given backend exists in the API
             return backend in backend_checks
 
-        def suitable_auto_backends(*args, **kwargs):
+        def suitable_auto_backends(cc, *args, **kwargs):
             if common_check is not None and not common_check(*args, **kwargs):
                 return False
             suitable_backends = []
             # Check for each backend support
             for backend in backend_checks:
+                req_checker = backend_checks[backend]
                 try:
-                    if backend_checks[backend](*args, **kwargs):
+                    if req_checker(
+                        *args, **kwargs
+                    ) and req_checker.is_compute_capability_supported(cc):
                         suitable_backends.append(backend)
                 except ValueError:
                     continue
@@ -1038,6 +1041,25 @@ def backend_requirement(
                 return False
             wrapper.suitable_auto_backends = suitable_backends
             return True
+
+        def _get_capability(*args, **kwargs):
+            capability = None
+            # Find the first tensor argument.
+            # Assume all tensors are on the same device/capability.
+            # We could consider check all tensors at a performance cost.
+            tensor_arg = None
+            all_args = args + tuple(kwargs.values())
+            for value in all_args:
+                if isinstance(value, torch.Tensor):
+                    tensor_arg = value
+                    break
+
+            if tensor_arg is not None:
+                # Get compute capability from the first tensor
+                # Assume all tensors are on the same device/capability
+                major, minor = get_compute_capability(tensor_arg.device)
+                capability = major * 10 + minor
+            return capability
 
         # @brief: Wrapper function that calls the orignal, decorated function, after applying a number of checks.
         # @note that here we manually apply defaults to the arguments in the wrapper function when doing validation.
@@ -1055,24 +1077,8 @@ def backend_requirement(
                 bound_args.apply_defaults()
                 # Convert to kwargs for validation functions
                 kwargs_with_defaults = dict(bound_args.arguments)
-
                 backend = kwargs_with_defaults.get("backend")
-
-                capability = None
-                # Find the first tensor argument.
-                # Assume all tensors are on the same device/capability.
-                # We could consider check all tensors at a performance cost.
-                tensor_arg = None
-                for value in kwargs_with_defaults.values():
-                    if isinstance(value, torch.Tensor):
-                        tensor_arg = value
-                        break
-
-                if tensor_arg is not None:
-                    # Get compute capability from the first tensor
-                    # Assume all tensors are on the same device/capability
-                    major, minor = get_compute_capability(tensor_arg.device)
-                    capability = major * 10 + minor
+                capability = _get_capability(*args, **kwargs)
                 if not has_backend_choices() and common_check is None:
                     raise ValueError(
                         f"Invalid @backend_requirement decorator usage: no backend choices and no common_check for {func.__name__}"
@@ -1080,7 +1086,9 @@ def backend_requirement(
 
                 if has_backend_choices():
                     if backend == "auto":
-                        if not suitable_auto_backends(**kwargs_with_defaults):
+                        if not suitable_auto_backends(
+                            capability, **kwargs_with_defaults
+                        ):
                             raise BackendSupportedError(
                                 f"No suitable auto backends found for {func.__name__}"
                             )
@@ -1109,7 +1117,8 @@ def backend_requirement(
             elif skip_check and heuristic_func is not None:
                 if kwargs.get("backend") == "auto":
                     # This needs to be called for heuristic function
-                    suitable_auto_backends(*args, **kwargs)
+                    capability = _get_capability(*args, **kwargs)
+                    suitable_auto_backends(capability, *args, **kwargs)
 
             return func(*args, **kwargs)
 


yzh119:
Hi @jimmyzho would you mind merging with (or rebase onto) main branch?

flashinfer/gemm.py
```diff
@@ -2013,6 +2009,69 @@ def mm_fp4(
     return out
 
 
+@supported_compute_capability([89, 90, 100, 103, 120])
```
bkryu: In the mm_fp4 case, we made a mistake of missing out a few SM archs like 110 and 121 in the first iteration.

Can you double check that we are not missing anything?

flashinfer/utils.py
```diff
@@ -989,6 +989,17 @@ def _is_problem_size_supported(*args, **kwargs):
...
 
+        def suitable_auto_backends(*args, **kwargs):
+            suitable_backends = [
+                backend
+                for backend in backend_checks
+                if backend_checks[backend](*args, **kwargs)
+            ]
+            if not suitable_backends:
+                return False
+            wrapper.suitable_auto_backends = suitable_backends
```
bkryu: Previously in bmm_fp8 when the backend was set to auto, the ordering of backends was `backends = ["cutlass", "cublas", "cudnn"]`. However, now it depends on the 

However, with the logic here in `gemm.py`'s bmm_fp8 decorator :
```
@backend_requirement(
    {
        "cudnn": _cudnn_bmm_fp8_requirement,
        "cublas": _cublas_bmm_fp8_requirement,
        "cutlass": _cutlass_bmm_fp8_requirement,
    },
    common_check=_check_bmm_fp8_problem_size,
)
```
the ordering becomes `wrapper.suitable_auto_backends=["cudnn", "cublas", "cutlass"]`.

My understanding is that this changes the kernel choice when we do not autotune because we select the first one in the list. Can you confirm whether this is true or not?

Additionally,  can you check whether `if backend_checks[backend](*args, **kwargs)` here is strong enough? If we imagine extending to mm_fp4, the common check will likely need to called for each backend so we may need `if backend_checks[backend](*args, **kwargs) and common_check(*args, **kwargs)` plus the ability to catch errors raised.

flashinfer/utils.py
```diff
@@ -983,6 +983,17 @@ def is_problem_size_supported(*args, **kwargs):
...
+            return True
```
bkryu: Same comment as I was asking above.

flashinfer/utils.py
bkryu: One more comment, we want `suitable_auto_backends` to also call a *heuristic function* that returns an ordered list of preferred backends. In `bmm_fp8`'s case, the function should return `["cutlass", "cublas", "cudnn"]` and the `suitable_auto_backends` should filter out those that are not supported from the heuristic list.

This will be needed because an auto backend for `mm_fp4` will need to dynamically decide whether to recommend cudnn or cutlass depending on the user's environment. The current implementation can only return a filtered list based on the static list.

bkryu (changes requested):
Thanks @jimmyzho for laying the foundation for backend=auto options. 

I left a few comments asking for clarifications and some minor changes. Can you take a look?

flashinfer/utils.py
jimmyzho: > the ordering becomes wrapper.suitable_auto_backends=["cudnn", "cublas", "cutlass"]

Yes, this is correct. One idea is that we can pass in the `{"cutlass": ..., "cublas":... , "cudnn": ...}` dict in order of what we want so that `suitable_auto_backends` will return the desired order. Another way is that the user can explicitly give the ordered list of auto backends as a new parameter into the decorator.

jimmyzho: > common check will likely need to called for each backend

One clarification, the common check function is backend-agnostic. However, if a problem size fails the common check, then all backends will be unsuitable.

bkryu: > One clarification, the common check function is backend-agnostic. However, if a problem size fails the common check, then all backends will be unsuitable.

Not sure whether it is intended but in the mm_fp4 case, this is not true. [_check_mm_fp4_problem_size](https://github.com/flashinfer-ai/flashinfer/blob/1e75bff99c175f5344ca079d40950106db157aee/flashinfer/gemm.py#L1696-L1754) has

```
    if backend != "trtllm" and use_8x4_sf_layout:
        raise ValueError("Only TRTLLM FP4 GEMM supports 8x4 scale factor layout.")
```
Maybe these checks are placed here because if we were to move these checks to the backend-specific check, the code would have to be duplicated

bkryu: > Yes, this is correct. One idea is that we can pass in the {"cutlass": ..., "cublas":... , "cudnn": ...} dict in order of what we want so that suitable_auto_backends will return the desired order. Another way is that the user can explicitly give the ordered list of auto backends as a new parameter into the decorator.

What we likely want is something dynamic that can take in a heuristic function, where we provide the function for each of bmm_fp8(), mm_fp4() and etc. Afterwards. the function for `bmm_fp8()` it should always return [cutlass, cublas, cudnn] while `mm_fp4()`'s heuristic function should check the CUDA version and cudnn version to return a preferred list.

bkryu (approved):
Thank you for the update @jimmyzho. LGTM

tests/utils/test_decorators.py
```diff
@@ -327,6 +327,77 @@ def my_kernel(x, backend="cudnn"):
...
+
+    @supported_compute_capability([80, 86, 89, 90, actual_capability])
+    def _cutlass_check(x, backend):
+        return x.shape[0] > 10
+
+    @supported_compute_capability([75, 80, 86, 89, 90, actual_capability])
+    def _cudnn_check(x, backend):
+        return x.shape[0] > 5
+
+    def _heuristic_func(suitable_backends, x, backend):
```
nvmbreughe, line 388: It's not immidiatly clear why these parameters are needed (suitable_backends, x, backend)

nvmbreughe: Perhaps add a simple decision tree that uses backends from suitable_backends. Not sure about backend param.

flashinfer/gemm.py
```diff
@@ -2013,6 +1997,101 @@ def mm_fp4(
...
+
+def _heuristic_func_bmm_fp8(
+    suitable_backends: List[str],
+    A: torch.Tensor,
+    B: torch.Tensor,
+    A_scale: torch.Tensor,
+    B_scale: torch.Tensor,
+    dtype: torch.dtype,
+    out: Optional[torch.Tensor] = None,
+    backend: Literal["cudnn", "cublas", "cutlass", "auto"] = "cublas",
```
nvmbreughe, line 2063: Is this param unused?

flashinfer/utils.py
```diff
@@ -1052,27 +1073,44 @@ def wrapper(*args, **kwargs):
...
                         )
-                if not _is_problem_size_supported(**kwargs_with_defaults):
-                    raise ValueError(
-                        f"Problem size is not supported for {func.__name__}"
-                    )
+                    if not _is_problem_size_supported(**kwargs_with_defaults):
+                        raise ValueError(
+                            f"Problem size is not supported for {func.__name__}"
+                        )
+            elif skip_check and heuristic_func is not None:
```
nvmbreughe, line 1117: We want to keep the skip_check path as low weight as possible. Especially in case of a non-auto backend (say cudnn), we don't necesarily want to do these argument binding here. 

So perhaps we want to make it : elif skip_check and backend=="auto" and heuristic_func is not None ?

nvmbreughe: Also as code rabit pointed out, these kwargs_with_defaults are actually not being used?

flashinfer/utils.py
```diff
@@ -1052,27 +1073,44 @@ def wrapper(*args, **kwargs):
...
+                                f" with capability {capability}" if capability else ""
+                            )
+                            raise BackendSupportedError(
+                                f"{func.__name__} does not support backend '{backend}'{extra}"
+                            )
+                        if not _is_problem_size_supported(**kwargs_with_defaults):
+                            raise ValueError(
+                                f"Problem size is not supported for {func.__name__}"
+                            )
                 else:
```
nvmbreughe, line 1107: Maybe add a comment here "If the function doesnt have backends (i.e., there is only 1, implicit backend), run the following checks.

flashinfer/gemm.py
jimmyzho: Yes - but since it may vary what the heuristic function may actually use, I'm assuming to just pass all params of the original func signature + suitable backends after support filtering

PR #2042, opened 2025-11-05
test: Fix test_sampling.py on Spark

bkryu:
## 📌 Description

Current PR fixes `test_sampling.py::test_softmax` on Spark by inserting a `torch.cuda.synchronize()` before calling the softmax function.

tl; dr why it works: PDL is enabled in these tests. Investigation shows that when PDL is enabled, `logits.view(-1).index_fill_(0, inf_idx, float("-inf"))` that prepares the inputs overlaps with the `probs = flashinfer.sampling.softmax(logits, temperature=temperature_arr)` function itself. Hence, we need to ensure that the input preparation is complete before running the softmax function to get the correct output.

#### Observations
`test_sampling.py::test_softmax` fails on select cases Spark. Example output
```
# pytest tests/utils/test_sampling.py::test_softmax
=================================================================================================================================================== test session starts ===================================================================================================================================================
platform linux -- Python 3.12.11, pytest-8.4.2, pluggy-1.6.0
rootdir: /flashinfer
configfile: pytest.ini
collected 324 items                                    
...
================================================================================================================================================= short test summary info =================================================================================================================================================
FAILED tests/utils/test_sampling.py::test_softmax[True-True-1.0-normal_distribution(std=1)-128256-989] - AssertionError: assert False
FAILED tests/utils/test_sampling.py::test_softmax[True-True-1.0-normal_distribution(std=5)-128256-989] - AssertionError: assert False
FAILED tests/utils/test_sampling.py::test_softmax[True-True-1.0-gumbel_distribution(beta=0.1)-128256-989] - AssertionError: assert False
======================================================================================================================================== 3 failed, 321 passed, 1 warning in 10.33s
```

Observations from debugging:
* When outputs are printed, rows containing all `nan`s are produced in the output of `probs = flashinfer.sampling.softmax(logits)`
* Surprisingly, the test passes with `CUDA_LAUNCH_BLOCKING=1 pytest tests/utils/test_sampling.py::test_softmax`
* `compute-sanitizer` does not detect any IMAs
* Running only a failed test results in a pass:
```
$ pytest tests/utils/test_sampling.py::test_softmax[True-True-1.0-normal_distribution\(std=1\)-128256-989]
...
1 passed, 1 warning in 0.80s
```

Towards a fix:
* I empirically find that the test passes:
   * when the reference `torch.softmax()` is called before `flashinfer.sampling.softmax()` (currently reference is called after)
   * when pdl is disabled in [line 67](https://github.com/flashinfer-ai/flashinfer/blob/main/tests/utils/test_sampling.py#L67) with `probs = flashinfer.sampling.softmax(logits, temperature=temperature_arr,enable_pdf=False)`
   * when `torch.cuda.synchronize()` is inserted in the line 64 as in this PR.
```
    if neg_inf_input:
        # assign random logits to -inf
        num_inf = torch.randint(0, logits.numel() - 1, (), device=logits.device).item()
        inf_idx = torch.randperm(logits.numel(), device=logits.device)[:num_inf]
        logits.view(-1).index_fill_(0, inf_idx, float("-inf"))
        torch.cuda.synchronize() ## This fixes the issue for some reason!

    if temperature_arr:
        temperature_arr = torch.full((batch_size,), temperature, device="cuda:0")
        probs = flashinfer.sampling.softmax(logits, temperature=temperature_arr)
        logits_scaled = logits / temperature_arr.unsqueeze(-1)
```
but **does not fix the issue if I place the synchronization any earlier**

An nsys profile shows that surprisingly the `logits.view(-1).index_fill_(0, inf_idx, float("-inf"))` and `flashinfer.sampling.softmax(logits, temperature=temperature_arr)` can overlap execution when pdl is enabled. 
Screenshot 2025-11-04 at 5 49 50 PM

This means that the softmax kernel is launching before inputs are done being prepared when `neg_inf_input=True`. Hence, placing a `torch.cuda.synchronize()` after the fill or disabling pdl can solve the issue. With the current PR, the nsys timeline changes to:
Screenshot 2025-11-04 at 5 51 32 PM
and the unit test passes.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Release Notes

* **Bug Fixes**
  * Improved synchronization of concurrent operations to ensure proper execution order and prevent potential timing-related issues.

### DIFF
--- a/tests/utils/test_sampling.py
+++ b/tests/utils/test_sampling.py
@@ -61,6 +61,7 @@ def test_softmax(
         num_inf = torch.randint(0, logits.numel() - 1, (), device=logits.device).item()
         inf_idx = torch.randperm(logits.numel(), device=logits.device)[:num_inf]
         logits.view(-1).index_fill_(0, inf_idx, float("-inf"))
+        torch.cuda.synchronize()  # wait for the index_fill_ to finish because it can overlap with the softmax kernel
 
     if temperature_arr:
         temperature_arr = torch.full((batch_size,), temperature, device="cuda:0")


tests/utils/test_sampling.py
```diff
@@ -61,6 +61,7 @@ def test_softmax(
         num_inf = torch.randint(0, logits.numel() - 1, (), device=logits.device).item()
         inf_idx = torch.randperm(logits.numel(), device=logits.device)[:num_inf]
         logits.view(-1).index_fill_(0, inf_idx, float("-inf"))
+        torch.cuda.synchronize()  # wait for the index_fill_ to finish because it can overlap with the softmax kernel
```
yzh119, line 64: That's interesting, why would this happen as all these operators should be executed within the same stream?

bkryu: Current evidence points to some issue with PDL where `index_filll_()` is overlapping with `flashinfer.sampling.softmax()`. As per why the GPU thinks it's okay to overlap, I am still investigating.

bkryu: Based on digging in the code and searching, our `OnlineSoftmaxFusedKernel` [allows overlapping with a preceding producer kernel via PDL](https://github.com/flashinfer-ai/flashinfer/blob/6d19a75a6963b0996b7b2cfe5f9e385c5cc41a66/include/flashinfer/sampling.cuh#L1365-L1379), but it is up to the producer kernel to tell when the consumer can start. It appears that torch's `index_fill_()` is allowing its consumer kernel to launch too early.

Upon profiling B200, we do see that the `index_fill_()` function also overlaps with softmax on B200, but the test does not fail on B200. 
Screenshot 2025-11-05 at 10 23 58 AM

Some notable differences might be:
* On B200, `index_fill_()` finishes in 1.7 usec with a 0.3 usec overlap (~18%)
* On Spark, `index_fill_()`finishes in 4.1 usec with a 2.7 usec overlap (~66%)

A plausible explanation of why we only see this failure on Spark is that on B200 or any other PDL enabled card, the `index_fill_` completes too quickly to expose the issue. 

Problematic test cases here all have vocab_size=128256, batch_size=989. logits has size 128256*989=126,845,184 elements (~121 MB for float32). Spark has 24MB L2 cache & 270 GB/s mem bandwidth while all other cards over SM90 (that allows PDL) have  substantially larger L2 cache size & mem bandwidths so it is expected that `index_fill_` runs much quicker on all other card and complete preparing the input before the softmax kernel starts accessing it.

yzh119: I see, thanks for collecting the profiling results.

I think the question might exist in torch's `index_fill_` implementation? It might launch dependents earlier than expected.

Another fix for this PR could be disable pdl for sm_121.

bkryu: > I think the question might exist in torch's index_fill_ implementation? It might launch dependents earlier than expected.

Yes I agree with this; it appears that it is running dependents too early. 

> Another fix for this PR could be disable pdl for sm_121.

This also works, but I went with the "synchronize before launching softmax" approach here because we are currently inadvertently overlapping `index_fill_` with `flashinfer.sampling.softmax` on all SKUs and this ensures they do not overlap. Let me know if you think there is a reason to disable pdl for sm121 instead. 

If not, I'd say the PR is ready otherwise 😄

yzh119: We are not sure if it's a sm_121 specific issue or a more general one (but not revealed because of kernel duration difference). Let's merge this one first, thanks for working on this!

yzh119 (approved):

PR #2069, opened 2025-11-10
minor: canonicalize TFLOPS calculation 

Edenzzzz:
## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Refactor**
  * TFLOPS computation standardized across attention benchmarks so reported performance metrics consistently account for actual sequence and batch lengths.

* **Bug Fixes**
  * Added checks to prevent invalid mixed-length causal inputs, avoiding misleading benchmark results.

* **Chores**
  * Renamed timing parameter in the benchmark utility for clearer intent.

### DIFF
--- a/flashinfer/testing/utils.py
+++ b/flashinfer/testing/utils.py
@@ -279,7 +279,9 @@ def attention_flops(
     """
     # Causal attention requires kv_len >= q_len
     if qo_seqlen > kv_seqlen:
-        causal = False
+        raise ValueError(
+            "qo_seqlen must be less than or equal to kv_seqlen for causal attention"
+        )
 
     if causal:
         bmm1_flops = (
@@ -330,7 +332,9 @@ def attention_flops_with_actual_seq_lens(
     # Causal attention requires kv_len >= q_len
     # Otherwise right align if kv_len > q_len
     if causal and (actual_seq_lens_q > actual_seq_lens_kv).any():
-        causal = False
+        raise ValueError(
+            "actual_seq_lens_q must be less than or equal to actual_seq_lens_kv for causal attention"
+        )
 
     if causal:
         bmm1_flops = (


benchmarks/bench_hopper_attention.py
```diff
@@ -176,14 +186,15 @@ def bench_batch_paged_prefill(
...
-            )
+        return attention_tflops_per_sec_with_actual_seq_lens(
+            batch_size * seq_len,
+            batch_size * seq_len,
+            head_dim,
+            head_dim,
+            num_qo_heads,
+            causal,
+            ms,
+        )
```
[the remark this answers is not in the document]
Edenzzzz: fixed

yzh119:
Hi @Edenzzzz thanks for the PR, did you observe different benchmark results?

Edenzzzz:
Actually I realized it's the same for regular seq lengths, but we could still use it to unify the calculation 😂

Edenzzzz:
Should it be like this? (causal requires kv_len >= qo_len, when kv_len > qo_len it's right aligned)
```
  if causal and (actual_seq_lens_q > actual_seq_lens_kv).any():
      causal = False
```

flashinfer/testing/utils.py
```diff
@@ -277,6 +277,10 @@ def attention_flops(
     Returns:
         total_flops (int): Total FLOPs for the layer.
     """
+    # Causal attention requires kv_len >= q_len
+    if qo_seqlen > kv_seqlen:
+        causal = False
```
yzh119: It looks confusing to me, why not directly raise runtime error?

If we need to support right align in the future, we should have more arguments in this function.

flashinfer/testing/utils.py
```diff
@@ -323,6 +327,11 @@ def attention_flops_with_actual_seq_lens(
     Returns:
         total_flops (int): Total FLOPs for the layer.
     """
+    # Causal attention requires kv_len >= q_len
+    # Otherwise right align if kv_len > q_len
+    if causal and (actual_seq_lens_q > actual_seq_lens_kv).any():
+        causal = False
```
yzh119: ditto

yzh119 (commented):
Made a few comments, otherwise LGTM

flashinfer/testing/utils.py
Edenzzzz: Currently causal is already right-aligned? Anyway, will raise an error
https://github.com/flashinfer-ai/flashinfer/blob/11177e8d0ebf32599786afcdc46211855d37074b/include/flashinfer/attention/prefill.cuh#L1598

PR #2072, opened 2025-11-10
[Test] Optimize test_trtllm_gen_fused_moe.py

jiahanc:
## 📌 Description
Currently `test_llm_gen_fused_moe` took long time, make some optimization to speed up
- Add autotuner choice to turn on/off autotuner in test
- Optimize token count for better coverage
- Optimize `check_accuracy` to speed up

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Tests**
  * Add configurable autotune flag (enabled/disabled) across MoE test paths and parametrized cases; propagate it through runtime/test flows
  * Ensure autotune-aware kernel invocations respect a max-token tuning limit (4096)
  * Clear tuner state between runs to improve cache isolation
* **Bug Fixes**
  * Tighten accuracy checks to enforce finiteness and use isclose-based matching with early exit when thresholds met

### DIFF
--- a/tests/moe/test_trtllm_gen_fused_moe.py
+++ b/tests/moe/test_trtllm_gen_fused_moe.py
@@ -49,6 +49,10 @@ from flashinfer.fused_moe.core import (
 from flashinfer.utils import get_compute_capability
 
 
+# Max num tokens to tune for trtllm-gen fused moe
+TUNE_MAX_NUM_TOKENS = 4096
+
+
 def check_cuda(err):
     """Unified CUDA error checking function used throughout the file."""
     if err != runtime.cudaError_t.cudaSuccess:
@@ -208,7 +212,7 @@ class CUDAGraphMoE:
             routing_method_type=self.config["routing_method_type"],
             gated_act_type=self.config["gated_act_type"],
             do_finalize=True,
-            tune_max_num_tokens=4096,
+            tune_max_num_tokens=TUNE_MAX_NUM_TOKENS,
         )
         return output  # Extract tensor from tuple
 
@@ -800,7 +804,7 @@ class FP8BlockScaleMoe(Moe):
                 use_shuffled_weight=static_data["use_shuffled_weight"],
                 weight_layout=static_data["weight_layout"],
                 enable_pdl=enable_pdl,
-                tune_max_num_tokens=4096,
+                tune_max_num_tokens=TUNE_MAX_NUM_TOKENS,
             )
         return output.to(torch.float)
 
@@ -977,7 +981,7 @@ class FP8PerTensorMoe(Moe):
                 == RoutingMethodType.Llama4,  # Use_routing_scales_on_input
                 None,
                 routing_method_type,
-                tune_max_num_tokens=4096,
+                tune_max_num_tokens=TUNE_MAX_NUM_TOKENS,
             )
 
         return output.to(torch.float)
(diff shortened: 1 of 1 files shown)


tests/moe/test_trtllm_gen_fused_moe.py
```diff
@@ -2267,7 +2282,7 @@ def run_moe_test(
 
 
 # Test: Renormalize routing
-@pytest.mark.parametrize("num_tokens", [1, 8, 1024, 3072])
+@pytest.mark.parametrize("num_tokens", [8, 768, 3072])
```
jiahanc, line 2286: token 1 and 8 uses same kernel, test 8 only. Replace 1024 to 768 to speed up while using same kernel

tests/moe/test_trtllm_gen_fused_moe.py
```diff
@@ -2586,7 +2605,7 @@ def test_topk_routing(
 
 
 # Test: Llama4 routing
-@pytest.mark.parametrize("num_tokens", [1, 8, 1024])
+@pytest.mark.parametrize("num_tokens", [8, 768, 3072])
```
yzh119: ditto

tests/moe/test_trtllm_gen_fused_moe.py
```diff
@@ -1408,20 +1418,18 @@ def routing_reference_topk(expert_logits, top_k, num_experts, padding):
...
     """Unified accuracy checking function with detailed error reporting."""
-    if torch.any(torch.isnan(a)):
-        raise Exception("NaN in reference output")
-    if torch.any(torch.isnan(b)):
-        raise Exception("NaN in actual output")
-    if torch.any(torch.isinf(a)):
-        raise Exception("Inf in reference output")
-    if torch.any(torch.isinf(b)):
-        raise Exception("Inf in actual output")
+    if not torch.isfinite(a).all():
```
yzh119, line 1425: It should be equivalent to `torch.any(torch.isinf(a))`, do you observe any performance difference?

tests/moe/test_trtllm_gen_fused_moe.py
```diff
@@ -207,6 +208,7 @@ def _run_moe_computation(self, runtime_args):
             routing_method_type=self.config["routing_method_type"],
             gated_act_type=self.config["gated_act_type"],
             do_finalize=True,
+            tune_max_num_tokens=4096,
```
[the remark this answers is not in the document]
yzh119: I think Gemini's suggestion is fair.

tests/moe/test_trtllm_gen_fused_moe.py
```diff
@@ -2295,8 +2307,9 @@ def run_moe_test(
                 "routing_method_type": RoutingMethodType.Renormalize,
                 "compatible_moe_impls": [FP8BlockScaleMoe, FP4Moe, BF16Moe],
                 "compatible_intermediate_size": [384, 768, 1024],
+                "enable_autotune": True,
```
yzh119, line 2320: Are there any cases where this argument is not set to True in this unittest?

tests/moe/test_trtllm_gen_fused_moe.py
yzh119: thanks for clarification

tests/moe/test_trtllm_gen_fused_moe.py
jiahanc: There are 2 separate call `torch.isnan` and `torch.isinf` to check if there is nan or inf in tensor. This could be done in 1 `torch.isfinite(a).all()`

tests/moe/test_trtllm_gen_fused_moe.py
jiahanc: Yes, there are 2 test config set to false with 2 reasons (DSLite and Renorm)
1. Want to test the accuracy without autotune
2. The config is not a real world model config, so should be safe to disable autotune

yzh119: thanks for explaination

yzh119 (approved):

PR #2074, opened 2025-11-10
MNNVL All Reduce for large number of tokens

nvmbreughe:
## 📌 Description

This PR does two things:
* Add a check for the number of tokens and raise an exception if the max token size was exceeded
* Adds an optional parameter to allow users to dial in an arbitrary workspace

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * Added an optional configurable workspace buffer size for all-reduce operations with a sensible default to preserve backwards compatibility.
  * Runtime input validation now enforces 2D inputs and token-count limits, with clearer error messages guiding corrective actions.

* **Tests**
  * Expanded test coverage for workspace behavior: default sizing, explicit sizing, and negative tests for insufficient workspace.
  * Tests now allow supplying an explicit workspace size to validate allocation and reuse scenarios.

### DIFF
--- a/flashinfer/comm/trtllm_mnnvl_ar.py
+++ b/flashinfer/comm/trtllm_mnnvl_ar.py
@@ -138,6 +138,7 @@ def get_allreduce_mnnvl_workspace(
     Args:
         mapping: Tensor parallel mapping configuration containing rank info
         dtype: Data type of the tensors being reduced
+        buffer_size_in_bytes: Optional buffer size. Practically, assign this to 3 * 2 * dtype.itemsize * hidden_dim * max_tokens
 
     Returns:
         Tuple containing:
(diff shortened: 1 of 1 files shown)


flashinfer/comm/trtllm_mnnvl_ar.py
yzh119, line 139: Please also add `buffer_size_in_bytes` to docstring.

cyx-6 (approved):
Lgtm

flashinfer/comm/trtllm_mnnvl_ar.py
```diff
@@ -223,6 +225,12 @@ def trtllm_mnnvl_all_reduce(
         [Optional] out: Output tensor to store the result (required if wait_for_results is True)
 
     """
+
+    if inp.shape[0] > buffer_M:
```
wenscarl, line 235: What If the inp is a 3D tensor? Could it be better:
```
if int(torch.prod(torch.tensor(inp.shape[:-1]))) > buffer_M:
...
```

wenscarl (approved):

flashinfer/comm/trtllm_mnnvl_ar.py
nvmbreughe: Good catch!

flashinfer/comm/trtllm_mnnvl_ar.py
nvmbreughe: Added a check to ensure its a 2D tensor. Thanks for pointing this out. The check was done on the testing side, but we do need to enforce it here in addition.

PR #2080, opened 2025-11-12
chore: update requires-python in pyproject.toml

raayandhar:
## 📌 Description

When installing I ran into this error:

```
  × No solution found when resolving dependencies for split (markers: python_full_version == '3.9.*'):
  ╰─▶ Because the requested Python version (>=3.9, <4.0) does not satisfy Python>=3.10 and nvidia-cutlass-dsl==4.2.1 depends on
      Python>=3.10, we can conclude that nvidia-cutlass-dsl==4.2.1 cannot be used.
      And because only nvidia-cutlass-dsl<=4.2.1 is available and your project depends on nvidia-cutlass-dsl>=4.2.1, we can conclude
      that your project's requirements are unsatisfiable.

      hint: Pre-releases are available for `nvidia-cutlass-dsl` in the requested range (e.g., 4.3.0.dev0), but pre-releases weren't
      enabled (try: `--prerelease=allow`)

      hint: The `requires-python` value (>=3.9, <4.0) includes Python versions that are not supported by your dependencies (e.g.,
      nvidia-cutlass-dsl==4.2.1 only supports >=3.10). Consider using a more restrictive `requires-python` value (like >=3.10).

      hint: While the active Python version is 3.12, the resolution failed for other Python versions supported by your project.
```
or in short, `nvidia-cutlass-dsl==4.2.1 only supports >=3.10`. So I just bumped the `pyproject.toml` to be stricter.

## 🔍 Related Issues

Hoping to avoid an issue post here I guess?

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Reviewer Notes

Please feel free to close if you don't think this is necessary. It's a super nitty thing anyways.

## Summary by CodeRabbit

* **Chores**
  * Raised minimum supported Python from 3.9 to 3.10.

* **Documentation**
  * Installation prerequisites updated: removed Python 3.9 and added Python 3.14.

* **Bug Fixes**
  * Iteration behavior tightened across components to enforce matching sequence lengths, replacing silent truncation with explicit validation to surface mismatches earlier.

### DIFF
--- a/benchmarks/bench_batch_attention.py
+++ b/benchmarks/bench_batch_attention.py
@@ -436,7 +436,7 @@ def main(args: argparse.Namespace) -> None:
     records_new = []
     records_separate = []
     for cfg_id, (decode_case, prefill_case) in enumerate(
-        zip(decode_lens, prefill_lens, strict=False), start=1
+        zip(decode_lens, prefill_lens, strict=True), start=1
     ):
         prefill_kv_lens = [p[0] for p in prefill_case]
         prefill_qo_lens = [p[1] for p in prefill_case]
--- a/benchmarks/bench_mixed_attention.py
+++ b/benchmarks/bench_mixed_attention.py
@@ -218,7 +218,7 @@ if __name__ == "__main__":
     head_dim = 128
 
     for idx, (p_q_lens, p_kv_lens, d_q_len, d_kv_len) in enumerate(
-        zip(p_q_configs, p_kv_configs, d_q_len_configs, d_kv_len_configs, strict=False)
+        zip(p_q_configs, p_kv_configs, d_q_len_configs, d_kv_len_configs, strict=True)
     ):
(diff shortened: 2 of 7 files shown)


raayandhar:
OK, so since in Python 3.10+, `zip()` added the keyword-only argument `strict` and now ruff requires that we specify this. There are two options, we can either ignore it in `pyproject.toml`, or we can add it. For now, I've added it with the most recent commit, but I can reverse this instead if that's better. I've done this by just applying the `--unsafe-fixes` flag. But to me it seems safe/reasonable to apply `strict=True` to most/all of them, not sure.

yzh119 (commented):
Yes `nvidia-cutlass-dsl` only supports python 3.10+, and as python 3.9 reached its end-of-life: https://devguide.python.org/versions/, I have no problem with this change.

yzh119:
I think `strict=True` looks better to me.

PR #2033, opened 2025-11-04
use scalar for kv_scale in xqa

qsang-nv:
## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [ ] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [ ] I have installed the hooks with `pre-commit install`.
- [ ] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [ ] Tests have been added or updated as needed.
- [ ] All tests are passing (`unittest`, etc.).

## Summary by CodeRabbit

* **Breaking Changes**
  * Public xqa/xqa_mla entry points now accept kv_scale as a plain float (default 1.0) instead of a 1-element tensor. Update call sites accordingly.

* **Documentation**
  * Docstrings updated to reflect kv_scale as float.

* **Tests**
  * Tests updated to pass scalar kv_scale, with added parameterization and conditional skip for FP8 kv-cache scenarios.

PR #2084, opened 2025-11-13
[API change] Allow using torch.Tensor for scales for trtllm-gen attention

IwakuraRein:
## 📌 Description

- change `bmm1_scale` and `bmm2_scale` to `Union[float, torch.Tensor]`. notice that when using tensor, it must be applied by log2e
- **remove the `bmm1_scale_log2_tensor` and `bmm2_scale_tensor` in the `xqa_batch_decode_with_kv_cache_mla`**
- update trtllm-gen FMHA kernels

TODO: do the same refactor for xqa kernels. The support for the device side scales was removed in #2033 

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * Attention scale parameters now accept either floats or 1-element tensors across prefill, decode and runtime; tensor scales are validated and applied on-device and pointer-backed scale paths are supported.

* **Chores**
  * Updated FMHA artifact path and checksum constants; added a public utility import and removed an obsolete inline comment.

* **Tests**
  * Updated tests to exercise device/tensor-or-scalar scale flows, removed legacy per-tensor call-site args, and added device-scale parametrization for several test variants.

### DIFF
--- a/tests/attention/test_trtllm_gen_attention.py
+++ b/tests/attention/test_trtllm_gen_attention.py
@@ -339,40 +339,7 @@ def unpack_compare_nvfp4(
     return output_unpacked, output_ref
 
 
-@pytest.mark.parametrize("kv_layout", ["HND", "NHD"])
-@pytest.mark.parametrize(
-    "batch_size,page_size,num_kv_heads,head_grp_size",
-    [
-        (4, 16, 2, 1),
-        (4, 32, 4, 5),
-        (4, 64, 4, 8),
-        (128, 16, 2, 5),
-        (128, 32, 4, 1),
-        (128, 64, 2, 8),
-        (256, 16, 4, 8),
-        (256, 32, 2, 8),
-        (256, 64, 4, 1),
-        (256, 64, 4, 5),
-    ],
-)
-@pytest.mark.parametrize("window_left", [-1])  # todo(Siyuan): add 127 window_left
-@pytest.mark.parametrize(
-    "q_dtype,kv_dtype,o_dtype",
-    [
-        ("bf16", "bf16", "bf16"),
-        ("fp16", "fp16", "fp16"),
-        ("fp8", "fp8", "bf16"),
-        ("fp8", "fp8", "fp16"),
-        ("fp8", "fp8", "fp8"),
-        ("fp8", "fp8", "nvfp4"),
-    ],
-)
-@pytest.mark.parametrize("enable_pdl", [True, False, None])
-@pytest.mark.parametrize("enable_sink", [True, False])
-@pytest.mark.parametrize("max_q_len", [511])
-@pytest.mark.parametrize("max_kv_len", [2047])
-@pytest.mark.parametrize("device_scale", [True, False])
-def test_trtllm_batch_prefill(
+def _test_trtllm_batch_prefill(
     kv_layout,
     batch_size,
     page_size,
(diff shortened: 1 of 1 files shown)


flashinfer/decode.py
```diff
@@ -2203,6 +2203,8 @@ def trtllm_batch_decode_with_kv_cache(
             o_scale=o_scale,
         )
     elif backend == "trtllm-gen":
+        print(f"{bmm1_scale=}")
```
huanghua1994: Remove debug print?

csrc/trtllm_fmha_kernel_launcher.cu
```diff
@@ -78,9 +80,10 @@ void trtllm_paged_attention_launcher(
     int64_t max_kv_len, int64_t num_pages_in_mem_pool, int64_t num_qo_heads, int64_t num_kv_heads,
     int64_t head_dim_qk, int64_t head_dim_vo, int64_t page_size, int64_t kv_stride_keys_values,
     int64_t kv_stride_heads, int64_t kv_stride_batch, int64_t max_num_blocks_per_seq,
-    double bmm1_scale, double bmm2_scale, double o_sf_scale, int64_t o_sf_vec_size,
-    int64_t o_sf_start_index, int64_t window_left, int64_t sum_seq_q, int64_t sm_count,
-    bool enable_pdl, int64_t workspace_size, cudaStream_t stream) {
+    double bmm1_scale, double bmm2_scale, const float* bmm1_scale_log2_ptr,
+    const float* bmm2_scale_ptr, double o_sf_scale, int64_t o_sf_vec_size, int64_t o_sf_start_index,
```
wenscarl, line 84: In the python API(decode.py), there is only bmm1_scale, but there are both bmm1_scale and bmm1_scale_ptr. Some comment here to indicate the difference in usage would be better.

csrc/trtllm_fmha_kernel_launcher.cu
```diff
@@ -197,11 +204,12 @@ inline bool is_4bit(Data_type data_type) { return data_type == Data_type::DATA_T
 void trtllm_paged_attention_decode(TensorView out, Optional<TensorView> out_scale_factor,
                                    TensorView query, TensorView key_cache, TensorView value_cache,
                                    TensorView workspace_buffer, TensorView block_tables,
-                                   TensorView seq_lens, int64_t max_kv_len, double bmm1_scale,
-                                   double bmm2_scale, double o_sf_scale, int64_t o_sf_vec_size,
-                                   int64_t o_sf_start_index, int64_t window_left, int64_t sm_count,
-                                   bool enable_pdl, int64_t workspace_size,
-                                   Optional<TensorView> attention_sinks) {
+                                   TensorView seq_lens, int64_t max_kv_len,
+                                   Variant<double, ffi::Tensor> bmm1_scale,
```
yzh119, line 208: cc @tqchen , I suppose Variant is a legit ABI across languages right?

flashinfer/decode.py
```diff
@@ -2172,6 +2174,11 @@ def trtllm_batch_decode_with_kv_cache(
         )
 
     if backend == "xqa":
+        # TODO(Siyuan): support device scale factors, which was removed in #2033
```
yzh119, line 2177: A reminder that we should fix the behavior in a future PR.

IwakuraRein:
> The unittest should cover both cases (using scalar bmm scale or device bmm scale),

@yzh119  At least [fp8](https://github.com/flashinfer-ai/flashinfer/blob/main/tests/attention/test_trtllm_gen_attention.py#L153-L154) is using tensor for bmm1_scale and bmm2_scale in the unit test now. For fp16 it's a float number.

Should we add the host scale for fp8 and device scale for fp16 in the unit test? That would double the unit test size though.

yzh119:
Lots of UT failed on gb300, but the error seems to be "no space on disk".

yzh119 (approved):
Failed UTs are not relevant (will be fixed in https://github.com/flashinfer-ai/flashinfer/pull/2097) and this PR itself LGTM, thanks for your contributions.

PR #2089, opened 2025-11-14
ci: Reduce test time by moving compilation off-line

kahyunnam:
## 📌 Description

Download `flashinfer-cubin` and `flashinfer-jit-cache `to avoid compilation. (Unless the JIT kernel is not in the `flashinfer-jit-cache`; then it will still JIT compile during test runtime. We could set `export FLASHINFER_DISABLE_JIT = 1 `to avoid this, but then it will "skip" a lot of tests that use JIT kernels that aren't found in `flashinfer-jit-cache`.)

## 🔍 Related Issues

Issue was discussed on slack. "Ideally, we would move that compilation off-line which would reduce test time & make kernel hang detection much easier. "

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Chores**
  * Improved runtime install flow to detect CUDA, compute an effective JIT architecture mapping, and install matching precompiled kernel artifacts plus local package sources; these steps run only outside dry-run mode and verify installation by showing configuration.
  * Simplified build parallelism calculation to a constant division by 8 (with existing safety guards retained).

* **Bug Fixes**
  * Missing precompiled kernel artifacts now cause an explicit error/abort instead of a warning.


scripts/task_test_blackwell_kernels.sh
```diff
@@ -25,7 +25,34 @@ if [[ "$1" == "--dry-run" ]] || [[ "${DRY_RUN}" == "true" ]]; then
...
+        esac
+    fi
+    echo "Using CUDA stream: ${CUDA_STREAM}"
+    echo ""
+
+    # Install precompiled kernels
+    echo "Installing flashinfer-cubin from PyPI/index..."
+    pip install -q flashinfer-cubin
+    echo "Installing flashinfer-jit-cache for ${CUDA_STREAM} from https://flashinfer.ai/whl/${CUDA_STREAM} ..."
+    pip install -q --extra-index-url "https://flashinfer.ai/whl/${CUDA_STREAM}" flashinfer-jit-cache
```
yzh119: Unfortunately, this approach may not work because `flashinfer-jit-cache` and `flashinfer-cubin` are only built when we tag new versions, so they may be out-of-sync with the latest codebase.

Workarounds include adding standalone workflow to build these wheels for latest codebase (and upload the artifacts), we have such workflows set up for nightly release: https://github.com/flashinfer-ai/flashinfer/actions/runs/19318123225, but I suppose it will take some time to propagate that to gitlab CI and so on.

Or we can try adding `warmup_jit` functions (https://github.com/flashinfer-ai/flashinfer/blob/54101e9533bd65af6393e7f86c30c0bff9794499/tests/attention/test_batch_attention.py#L29-L51) ahead of each UT to pre-compile all kernels used in the UT with multiple CPU cores.

kahyunnam: Ah right that's true. 

I think the `warmup_jit` idea is the best solution here -- though it does seem like most of the UT already have this?

kahyunnam: Is the "nightly release" different from latest codebase build?

bkryu: Also the script `task_test_blackwell_kernels.sh` is used to run unit tests on PRs, in which case there are no downloadable compiled jit-caches. Like Zihao is suggesting, we would need to compile for the PR.

kahyunnam: OK, updated the PR. This will be merged after some internal gitlab Flashinfer-ci updates to add the new pipeline stage for precompiling.

scripts/task_test_blackwell_kernels.sh
```diff
@@ -25,7 +25,38 @@ if [[ "$1" == "--dry-run" ]] || [[ "${DRY_RUN}" == "true" ]]; then
...
+    # Install precompiled kernels (require CI build artifacts)
+    : ${DIST_CUBIN_DIR:="../dist/${CUDA_VERSION}/cubin"}
+    : ${DIST_JIT_CACHE_DIR:="../dist/${CUDA_VERSION}/jit-cache"}
+
+    if [ -d "${DIST_CUBIN_DIR}" ] && ls "${DIST_CUBIN_DIR}"/*.whl >/dev/null 2>&1; then
+        echo "Installing flashinfer-cubin from ${DIST_CUBIN_DIR} ..."
+        pip install -q "${DIST_CUBIN_DIR}"/*.whl
+    else
+        echo "ERROR: flashinfer-cubin wheel not found in ${DIST_CUBIN_DIR}. Ensure the CI build stage produced the artifact." >&2
+        exit 1
```
bkryu: I suggest that we print a warning message and proceed if the cubin and jit-cache wheel is not detected instead of returning an error here.

The reason being if someone wants to run this script locally, they should be able to just run `bash scripts/tast_test_blackwell_kernels.sh` without needing to worry about wheels not being there.

I'm open to suggestions from anyone else though.

scripts/task_test_blackwell_kernels.sh
```diff
@@ -25,7 +25,38 @@ if [[ "$1" == "--dry-run" ]] || [[ "${DRY_RUN}" == "true" ]]; then
...
+    fi
+
+    if [ -d "${DIST_JIT_CACHE_DIR}" ] && ls "${DIST_JIT_CACHE_DIR}"/*.whl >/dev/null 2>&1; then
+        echo "Installing flashinfer-jit-cache from ${DIST_JIT_CACHE_DIR} ..."
+        pip install -q "${DIST_JIT_CACHE_DIR}"/*.whl
+    else
+        echo "ERROR: flashinfer-jit-cache wheel not found in ${DIST_JIT_CACHE_DIR} for ${CUDA_VERSION}. Ensure the CI build stage produced the artifact." >&2
+        exit 1
```
bkryu: Ditto here

scripts/task_test_blackwell_kernels.sh
kahyunnam: That makes sense to me; I removed the "exit 1" for both the cubin / jit-cache else logic, so it will continue with the pip installation with compile from scratch as the alternative.

yzh119 (approved):

Issue #2094, opened 2025-11-14
Add optimal gb200 moe comm kernels


nvrohanv:
Flashinfer seems to be missing the latest MOE comm kernels for multinode-nvlink/gb200.

TRTLLM's path is mnnvl_moe_alltoallv_combine -> torch.ops.trtllm.moe_comm -> moeCommOp -> tensorrt_llm::kernels::moeAllToAll, see [https://github.com/NVIDIA/TensorRT-LLM/blob/222bc911cd35405f3539c366da6c03c00e9a7fb7/cpp/tensorrt_llm/kernels/fusedMoeCommKernels.cu#L1406](https://nam11.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FNVIDIA%2FTensorRT-LLM%2Fblob%2F222bc911cd35405f3539c366da6c03c00e9a7fb7%2Fcpp%2Ftensorrt_llm%2Fkernels%2FfusedMoeCommKernels.cu%23L1406&data=05%7C02%7Crohanv%40nvidia.com%7C14b30ef960d549a1baab08de23053810%7C43083d15727340c1b7db39efd9ccc17a%7C0%7C0%7C638986702517301506%7CUnknown%7CTWFpbGZsb3d8eyJFbXB0eU1hcGkiOnRydWUsIlYiOiIwLjAuMDAwMCIsIlAiOiJXaW4zMiIsIkFOIjoiTWFpbCIsIldUIjoyfQ%3D%3D%7C0%7C%7C%7C&sdata=zFUC6%2FdVn8mVyY0OHq8zTmRxUxScao3sQZOTovM1bMw%3D&reserved=0)

[...]

PR #1245, opened 2025-07-14
Mnnvl memory with custom communicator

wenscarl:
## 📌 Description

Make Mnnvl memory initialization independent of mpi4py communicator.

## 🔍 Related Issues

Depends by https://github.com/flashinfer-ai/flashinfer/pull/1550

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [ ] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [ ] I have installed the hooks with `pre-commit install`.
- [ ] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [ ] Tests have been added or updated as needed.
- [ ] All tests are passing (`unittest`, etc.).

PR #2102, opened 2025-11-18
Port TRT-LLM communication kernels to flashinfer

djns99:
## 📌 Description

This ports the latest MNNVL A2A communication implementation from TRT-LLM

## 🔍 Related Issues

https://github.com/flashinfer-ai/flashinfer/issues/2094 

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * Added a throughput‑optimized Mixture‑of‑Experts all‑to‑all backend with dispatch, combine and sanitization phases and a high‑level Python class for workspace management and zero‑copy payloads.
  * New runtime tunables for MOE A2A behavior and block‑size tuning; KV‑cache time output setting renamed.

* **Documentation**
  * Added API reference for the new A2A throughput backend.

* **Tests**
  * Added comprehensive single‑ and multi‑rank end‑to‑end tests and updated test runners to include A2A scenarios.

### DIFF
--- a/flashinfer/comm/trtllm_moe_alltoall.py
+++ b/flashinfer/comm/trtllm_moe_alltoall.py
@@ -12,7 +12,7 @@ from typing import Optional
 import torch
 import functools
 
-from .api_logging import flashinfer_api
+from ..api_logging import flashinfer_api
 
 from .mnnvl import MnnvlMemory, MnnvlConfig
 from .mapping import Mapping


flashinfer/comm/trtllm_moe_alltoall.py
```diff
@@ -0,0 +1,556 @@
...
+            top_k: Number of experts per token
+            num_experts: Total number of experts
+            workspace_size_per_rank: Size of workspace per rank in bytes (default: 512MB)
+        """
+        # Initialize constants from C++
+        self._init_constants()
+
+        # Initialize MNNVL memory system
+        MnnvlMemory.initialize()
+
```
trevor-m, line 545: https://github.com/flashinfer-ai/flashinfer/pull/1245 adds an optional custom communicator for the MnnvlMemory initialization. This is necessary because sglang/vllm don't use MPI so the previous bootstrapping wouldn't work. We will need to allow users to pass through the custom communicator here as well.

1. Can you add an arg to this function
```
config: Optional[MnnvlConfig] = None
```

2. Then after MnnvlMemory.initialize(), add
```
if config:
            MnnvlMemory.set_comm_from_config(mapping, config)  # type: ignore[attr-defined]
```

I think that should take care of it. Thanks!

djns99: Done, please let me know if it looks correct

csrc/nv_internal/tensorrt_llm/kernels/communicationKernels/moeAlltoAllKernels.cu
```diff
@@ -0,0 +1,873 @@
...
+  } else {                                              \
+    using POLICY = WarpPolicy;                          \
+    __VA_ARGS__                                         \
+  }
+
+#if DISABLE_TIMEOUT
+#define check_timeout(s) false
+#else
+// 300 * 2000 MHz - should be high enough on any GPU but will prevent a hang
+#define check_timeout(s) ((clock64() - (s)) > (300ll * 2000ll * 1000ll * 1000ll))
```
djns99, line 115: I have manually added this, can I get someone to sanity check my logic here

scripts/task_test_multi_node_comm_kernels.sh
```diff
@@ -6,13 +6,14 @@ set -x
...
-echo "Cache cleaned."
-echo ""
+# echo "Cleaning Python bytecode cache..."
+# find . -type d -name __pycache__ -exec rm -rf {} + 2>/dev/null || true
+# find . -type f -name '*.pyc' -delete 2>/dev/null || true
+# echo "Cache cleaned."
+# echo ""
 
-pip install -e . -v
+# pip install -e . -v
```
djns99, line 15: Im pretty sure this file is intended to be run with mpirun, so doing this logic on every rank would be incorrect behaviour. I don't see where this task is used though, so I could be misunderstanding something

flashinfer/comm/trtllm_moe_alltoall.py
```diff
@@ -0,0 +1,619 @@
...
+    metainfo: torch.Tensor,
+    ep_rank: int,
+    invalid_expert_id: int,
+):
+    return get_mnnvl_a2a_module().moe_a2a_sanitize_expert_ids(
+        expert_ids, workspace, metainfo, ep_rank, invalid_expert_id
+    )
+
+
+def moe_a2a_get_workspace_size_per_rank(
```
djns99, line 359: This API is also an addition of mine, some extra eyes on this would be helpful

flashinfer/comm/trtllm_moe_alltoall.py
```diff
@@ -0,0 +1,619 @@
...
+
+    Example:
+        >>> moe_a2a = MoeAlltoAll(mapping, max_num_tokens=2048, top_k=2, num_experts=8)
+        >>> recv = moe_a2a.dispatch(experts, [hidden, ids, scales], batch_size)
+        >>> output = moe_a2a.combine(processed, batch_size)
+    """
+
+    # Single shared workspace across the process
+    # _WORKSPACE: Optional[dict] = None
+    _WORKSPACE_CACHE: dict[tuple[int, int, int, int], dict] = {}
```
djns99, line 408: I updated this to use a cache so that we support different shaped communicators existing at the same time. Previously this assumed all communicators had the same shape which made testing tricky.
I would appreciate if people could double check my working here

flashinfer/comm/trtllm_moe_alltoall.py
trevor-m: Thanks, looks good!

csrc/nv_internal/tensorrt_llm/kernels/communicationKernels/moeAlltoAllKernels.cu
trevor-m: Looks good to me

flashinfer/comm/trtllm_moe_alltoall.py
trevor-m: This looks good to me. I don't think sglang will need different shaped communicators right now but may be nice for other cases or in the future.

flashinfer/comm/trtllm_moe_alltoall.py
trevor-m: It looks like TRT-LLM just added (in the last 24 hours) something similar which looks slightly more user friendly: https://github.com/NVIDIA/TensorRT-LLM/commit/8b5ededc83bf10d16619a1d8364a886d25fe4a75

Maybe in a future PR that could be integrated to flashinfer

djns99: Yeah I am in touch with the author, I was going to port across as part of this, it should be a small change

flashinfer/comm/trtllm_moe_alltoall.py
djns99: Yeah mainly useful for the tests, in trt-llm each test is launched under MPI so has a new state, but we dont do that in flashinfer.

flashinfer/comm/trtllm_moe_alltoall.py
djns99: @trevor-m @bobboli  I ported those changes into this MR. Can you take a look and see if it looks sensible. I made some tweaks to make it slightly more flexible, since @bobboli's version was specific to TRT-LLM's usage

scripts/task_test_multi_node_comm_kernels.sh
djns99: @dierksen @nvmbreughe  will this cause any issues that you know?

dierksen: I don't know of any issues. I don't think we actually use this in CI at the moment.

flashinfer/comm/trtllm_moe_alltoall.py
bobboli: LGTM.

bobboli: BTW, I am still seeing names like "MNNVL". This term is not precise as the kernel is also supposed to work for the nvlink domain within a single node like DGX B200. Could you modify it to be sth like "trtllm" "nvlink_one_sided" to reduce discrepancy?

djns99: Currently this uses `MnnvlMemory.initialize()` in the MoeAllToAll class and uses that to allocate the workspace, does that not imply that this is still MNNVL only?
I agree we can rename, but I wonder if it would be more consistent with flashinfer to keep the mnnvl name?
That said, better to fix the name now while no-one is using it

djns99: Also `nvlink_one_sided` is really a technical detail that I don't think we should be exposing to end users. Instead should I call it `nvlink_moe_alltoall` or something similar. Though I guess the dispatch is not specific to MOE

djns99: Ok I removed mnnvl from the name of the C++ bindings and JIT module. The only reference is the MnnvlMemory object used in MoeAllToll

flashinfer/comm/trtllm_moe_alltoall.py
```diff
@@ -0,0 +1,719 @@
...
+        moe_a2a_initialize=moe_a2a_initialize,
+        moe_a2a_dispatch=moe_a2a_dispatch,
+        moe_a2a_combine=moe_a2a_combine,
+        moe_a2a_sanitize_expert_ids=moe_a2a_sanitize_expert_ids,
+        moe_a2a_get_metainfo_index_pairs=moe_a2a_get_metainfo_index_pairs,
+        moe_a2a_get_aux_data_size=moe_a2a_get_aux_data_size,
+    )
+
+
+def moe_a2a_initialize(
```
aleozlx, line 202: i believe we started marking these public apis with a decorator `@flashinfer_api` it adds logging and helps readability for users and code reviewers

good job having already added to the documentation too

csrc/nv_internal/cpp/common/envUtils.cpp
```diff
@@ -328,4 +323,37 @@ uint16_t getEnvNixlPort() {
...
+  if (block <= 0) block = 256;
+  if (block > 1024) block = 1024;
+  // Round to nearest multiple of 32 (warp size)
+  block = (block + 31) / 32 * 32;
+  if (block == 0) block = 256;
+  return block;
+}
+
+int getEnvMoeA2ADispatchBlockSize() {
+  static int const kBlock = sanitizeBlockSize(getIntEnv("TLLM_MOE_A2A_DISPATCH_BLOCK_SIZE"));
```
aleozlx, line 348: should we document env vars too?

flashinfer/comm/trtllm_moe_alltoall.py
djns99: Done I decorated this and the class methods, that will mean that it gets logged twice if users use the helper class. Let me know if you would prefer something different

csrc/nv_internal/cpp/common/envUtils.cpp
djns99: @bobboli can you advise? Should we remove these or do you know the cases these should be enabled

flashinfer/comm/trtllm_moe_alltoall.py
aleozlx: **i think it's ok.** the outer level logged as `MoeAlltoAll` and the inner level logged as `moe_a2a_*`. as a user looking at the logging, i would grep either the class or the underlying api functions that are also considered public api.
cudnn has also the precedence of allowing public api calling public api to be logged, even tracking the depth and tweak formatting. overall i think this is fine & no concerns for now
cc @bkryu for vis

bkryu: This should be okay; there are similar examples within the library as well.

One example can be found in `fp4_quantization.py` where `fp4_quantize()` and `mxfp4_quantize()` are both decorated functions, but `mxfp4_quantize()`'s [implementation](https://github.com/flashinfer-ai/flashinfer/blob/main/flashinfer/fp4_quantization.py#L877C5-L890) calls `fp4_quantize`so there is double-logging being done

aidando73:
Just tested this PR

```python
trt-llm mnnvl bf16
[COMMS] hidden=7168, intermediate=2048, experts=256, topk=8, topk_group=4, n_group=8, cuda_graph=True, dtype=bf16, loop=10, world_size=16
|   batch_size |   trt_mnnvl |
|--------------|-------------|
|           32 |          45 |
|          128 |          44 |
|          256 |          47 |
|          512 |          54 |

flashinfer
|   batch_size |   flashinfer_mnnvl_new |
|--------------|------------------------|
|           32 |                     44 |
|          128 |                     46 |
|          256 |                     47 |
|          512 |                     51 |
```

Can repro bf16. Thanks :)

But when I send nvfp4 quantized tensors - it's slower by about 6-8 us.

```python
[COMMS] hidden=7168, intermediate=2048, experts=256, topk=8, topk_group=4, n_group=8, cuda_graph=True, dtype=bf16, loop=10, world_size=16
|   batch_size |   trt_mnnvl_nvfp4_nvfp4 |
|--------------|-------------------------|
|           32 |                      38 |
|          128 |                      41 |
|          256 |                      42 |
|          512 |                      48 |

fp4 - x_quant, x_scales (nvfp4) - 16 ranks
|   batch_size |   flashinfer_mnnvl_new_nvfp4_nvfp4 |
|--------------|------------------------------------|
|           32 |                                 46 |
|          128 |                                 48 |
|          256 |                                 48 |
|          512 |                                 53 |
```

Any idea why?

yzh119:
Hi @aidando73 can you share your benchmark script?

scripts/task_test_multi_node_comm_kernels.sh
yzh119: I believe these code are designed to get around some python side module refactoring (while the CI node have python cache for older versions, which should not be the case as we are running UTs inside docker).

tests/comm/test_mnnvl_memory.py
```diff
@@ -122,7 +122,7 @@ def test_mnnvl_memory(self):
         reason="Mnnvl memory is not supported on this platform",
     )
     def test_moe_alltoall_multi_rank_single_gpu(self):
-        torch.cuda.set_device(self.rank)
+        torch.cuda.set_device(self.local_rank)
```
yzh119, line 125: Is it a bug with the original test?

yzh119 (approved):
The PR itself is in good shape and I believe it's ready to be merged. How about we study the performance gap and fix in future PRs? @djns99 @aidando73

aidando73:
Here it is - disclaimer it's fairly vibecoded

https://gist.github.com/aidando73/b1cccf6f5650e30d24c6faea129ad9a0. It's `flashinfer_mnnvl_new_nvfp4_nvfp4` vs `trt_mnnvl_nvfp4_nvfp4`.

I think it's related to sending multiple tensors. This is bf16 - with sending 2 tensors

```
Flashinfer mnnvl new correct numerics:
Timings (microseconds):
|   batch_size |   flashinfer_mnnvl_new |
|--------------|------------------------|
|           32 |                     51 |
|          128 |                     51 |
|          256 |                     52 |
|          512 |                     55 |

Timings (microseconds):
|   batch_size |   trt_mnnvl |
|--------------|-------------|
|           32 |          39 |
|          128 |          42 |
|          256 |          44 |
|          512 |          52 |
```

Happy to hop on a call as well

aidando73:
@yzh119 Yeah sgtm

aidando73:
Oh - are you guys not using the FIFO queue implementation here? Looks like this implementation is a PUT -> barrier - what was the rationale? Seems like the FIFO queue version was performing pretty well.

djns99:
> Oh - are you guys not using the FIFO queue implementation here? Looks like this implementation is a PUT -> barrier - what was the rationale? Seems like the FIFO queue version was performing pretty well.

@aidando73  I believe this approach is faster than FIFO in TRT-LLM for some of the DeepSeek workloads we have tested. In general it is less memory efficient, but because it is one shot it is better for throughput cases.
@bobboli can you share more details

tests/comm/test_mnnvl_memory.py
djns99: Yeah, these tests aren't run in CI currently. I tested them with 2 nodes + mpirun, they pass with this change

csrc/nv_internal/tensorrt_llm/kernels/communicationKernels/moeAlltoAllKernels.cu
```diff
@@ -0,0 +1,874 @@
...
+#include "tensorrt_llm/common/envUtils.h"
+#include "tensorrt_llm/kernels/communicationKernels/moeAlltoAllKernels.h"
+
+namespace tensorrt_llm::kernels::moe_alltoall {
+
+#define ENABLE_DEBUG_PRINT 0
+#define DISABLE_SYNC_FOR_PROFILING 0
+
+#ifndef DISABLE_TIMEOUT
+#define DISABLE_TIMEOUT 0
```
djns99, line 34: @aidando73 can you try benchmark with timeout disabled and see if that changes perf?
I added this logic because I think it is good to have some error handling rather than hangs, but if its a perf issue we can turn off by default. It should only be a couple extra instructions on the critical path, but its possible it breaks some optimisations

aidando73:
> better for throughput cases.

Oh so should I be testing with bigger batch sizes. Do you consider batch size 256, 512 a throughput case

csrc/nv_internal/tensorrt_llm/kernels/communicationKernels/moeAlltoAllKernels.cu
aidando73: Ok trying I'm assuming I set this at runtime: FLASHINFER_EXTRA_CUDAFLAGS="-DDISABLE_TIMEOUT=1"

aidando73:
> In general it is less memory efficient

For my understanding - when you say the FIFO implementation is more memory efficient - that's because we're copying from shared memory -> fabric -> shared memory? And the shared memory <-> global memory copies are overlapped with compute / nvlink transfers so we kind of get them for free?

csrc/nv_internal/tensorrt_llm/kernels/communicationKernels/moeAlltoAllKernels.cu
aidando73:
```
Timings (microseconds):
|   batch_size |   flashinfer_mnnvl_new |
|--------------|------------------------|
|           32 |                     48 |
|          128 |                     50 |
|          256 |                     51 |
|          512 |                     55 |
```

No much difference.

Confirmed that that flag is being set as well:

```
Running comms benchmark for batch size (global): 512 (impl=flashinfer_mnnvl_new)
DISABLE_TIMEOUT: 1
DISABLE_TIMEOUT: 1
```

aidando73:
> better for throughput cases.

Tested on 1024 and 2048 - seems like the cross-over point is 1024 for these comms

```bash
flashinfer:
|         1024 |                     65 |
|         2048 |                     82 |
trt_llm:
|   batch_size |   trt_mnnvl |
|--------------|-------------|
|         1024 |          66 |
|         2048 |          96 |
```

csrc/nv_internal/tensorrt_llm/kernels/communicationKernels/moeAlltoAllKernels.cu
djns99: Ok good to know that isn't impacting perf

djns99:
> For my understanding - when you say the FIFO implementation is more memory efficient - that's because we're copying from shared memory -> fabric -> shared memory? And the shared memory <-> global memory copies are overlapped with compute / nvlink transfers so we kind of get them for free?

Memory efficient here I just mean that we have to allocate the max number of tokens on all ranks up front. i.e. the workspace is bigger in this version

bobboli:
> > For my understanding - when you say the FIFO implementation is more memory efficient - that's because we're copying from shared memory -> fabric -> shared memory? And the shared memory <-> global memory copies are overlapped with compute / nvlink transfers so we kind of get them for free?
> 
> Memory efficient here I just mean that we have to allocate the max number of tokens on all ranks up front. i.e. the workspace is bigger in this version

Yes, `flashinfer_mnnvl_new` may require larger memory, but it's not a consideration for kernel speed.

Please see https://github.com/flashinfer-ai/flashinfer/issues/2222#issuecomment-3659570965 for more discussions.

PR #2116, opened 2025-11-19
update autotuner input tensor random range

jiahanc:
## 📌 Description

Update autotuner input tensor random range from [0,1) to [-5,5) for larger range and closer to real tensor

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tensor initialization used during autotuning: values are now drawn from a symmetric range around zero ([-5, 5]) with a more uniform-like distribution, yielding more consistent and stable parameter tuning results.

### DIFF
--- a/flashinfer/autotuner.py
+++ b/flashinfer/autotuner.py
@@ -62,8 +62,8 @@ class DynamicTensorSpec:
         if self.tensor_initializers is None:
             self.tensor_initializers = [
                 lambda shapes, dtype, device: (
-                    torch.randn(shapes, device=device).to(dtype) * 10 - 5
-                )
+                    torch.rand(shapes, device=device) * 10 - 5
+                ).to(dtype)
                 for _ in range(len(self.input_idx))
             ]
 


flashinfer/autotuner.py
```diff
@@ -61,8 +61,8 @@ def __post_init__(self):
         # Set default tensor_initializers if not provided
         if self.tensor_initializers is None:
             self.tensor_initializers = [
-                lambda shapes, dtype, device: torch.randn(shapes, device=device).to(
-                    dtype
+                lambda shapes, dtype, device: (
+                    torch.randn(shapes, device=device).to(dtype) * 10 - 5
                 )
```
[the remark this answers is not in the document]
jiahanc: Nice advice, thank you!

flashinfer/autotuner.py
```diff
@@ -61,9 +61,9 @@ def __post_init__(self):
         # Set default tensor_initializers if not provided
         if self.tensor_initializers is None:
             self.tensor_initializers = [
-                lambda shapes, dtype, device: torch.randn(shapes, device=device).to(
```
yzh119, line 64: randn is gaussian distribution, which is different from your desciption:

> input tensor random range from [0,1) to [-5,5) for larger range 

where [0, 1) is a uniform distribution.

I have no idea about the real data distribution tbh and changing it to [-5, 5) seems fine. Just a heads up to make sure it's not a typo.

flashinfer/autotuner.py
```diff
@@ -761,8 +761,8 @@ def _create_tensor_like(
     def _prepare_input_tensors(
         self, profile: OptimizationProfile, inputs: List[torch.Tensor]
     ) -> List[torch.Tensor]:
-        default_initializer = lambda shapes, dtype, device: torch.rand(
```
yzh119, line 764: this is indeed uniform distribution [0, 1)

flashinfer/autotuner.py
jiahanc: thanks for pointing out. Is there a reason why `randn` here but `rand` in the other place?
Reason to change to [-5,5) is @rosenrodt did some work on MXFP4 tuning experiment and found this range can get better autotuner config than [0,1)

rosenrodt: I speculated [-5, 5) is than [0, 1) because the latter could truncate to 0s, thus affecting the power profile during autotune and less representative of the power profile of the actual workload.

yzh119: Yes I think if [-5, 5) is better let's use it, data distribution affects kernel execution time.

yzh119 (approved):

PR #2056, opened 2025-11-06
Add custom communicator for trtllm_mnnvl_ar

wenscarl:
## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [ ] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [x] Tests have been added or updated as needed.
- [x] All tests are passing (`unittest`, etc.).

## Summary by CodeRabbit

* **New Features**
  * Added optional communication-backend parameter for multi-node memory and buffer allocation to allow using a provided communicator for handle transfer.

* **Bug Fixes / Reliability**
  * Multi-node synchronization now uses the provided communicator's barrier when available, preserving previous behavior otherwise.

* **Tests**
  * Added end-to-end tests covering custom communication backends and multi-node all-reduce synchronization.

PR #2118, opened 2025-11-20
Refactor trtllm_mnnvl_allreduce

timlee0212:
## 📌 Description

This PR porting all changes in [TensorRT-LLM#8018](https://github.com/NVIDIA/TensorRT-LLM/pull/8018) into Flashinfer.

Apart from the changes mentioned in the original PR, this PR also introduce new API interface as `trtllm_mnnvl_allreduce` and `trtllm_mnnvl_fused_allreduce_add_rmsnorm` to replace the original ones. The workspace allocation is wrapped as an entire class with a given buffer size and the user does not need to worry about the details inside. 

This PR adds support for IPC Socket based mcast device memory bootstrap so that it can run on DGX machine that does not support fabric handle.

@wenscarl This PR also incorporate the changes made in #2056 and should be able to replace that PR. A bcast interface is added to the comm backend as this is needed  during the handle transfer.

The old API is tagged as deprecated and redirected to the new APIs. The user of the old API should not need to make any changes.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * Fused all‑reduce with optional RMSNorm fusion and selectable one‑shot/two‑shot strategies; new Python APIs and workspace utilities; IPC-based handle exchange and bcast support.

* **Improvements**
  * Pluggable handle‑exchange backends (Fabric/POSIX), stricter I/O and shape validation, renamed/standardized fusion entry points and parameter surfaces, cached CUDA SM count for tuning, and safer lifecycle/cleanup.

* **Tests**
  * MPI‑aware tests for fused and legacy flows, workspace-based runs, synchronization, and expanded sequence/hidden‑size coverage.

### DIFF
--- a/flashinfer/comm/trtllm_mnnvl_ar.py
+++ b/flashinfer/comm/trtllm_mnnvl_ar.py
@@ -60,10 +60,19 @@ class MNNVLAllreduceFusionWorkspace:
         comm_backend: Optional[CommBackend] = None,
     ):
         """
-        Initialize the MNNVL Allreduce Fusion Workspace. comm_backend will be used for creating the workspace and synchronization. If not provided, MPIBackend will be used which will use COMM_WORLD for synchronization.
+        Initialize the MNNVL Allreduce Fusion Workspace. The workspace will be allocated and initialized based on the provided problem size. If max_num_tokens is larger than the one-shot threshold, the workspace will be created according to the max of required one-shot size at threshold, or the required two-shot size. Note that the workspace is not bind to the given problem size. It can be reused for different problem size without reinitialization given the allocated size is sufficient.
+
+        If the buffer_size_in_bytes is provided, the workspace will be created according to the provided size. The user is expected to use the utility function get_required_buffer_size_bytes to calculate the required size. The actual allocation size may be larger due to alignment requirements. This covers the advanced used case, for example, the user may want to enforce oneshot strategy and ignore the heuristics.
+
+        Either max_num_tokens or buffer_size_in_bytes must be provided.
+
+        comm_backend will be used for creating the workspace and synchronization. If not provided, MPIBackend will be used which will use COMM_WORLD for synchronization.
 
         Args:
             mapping: Mapping configuration containing rank info
+            max_num_tokens: The maximum number of tokens in the input tensor.
+            hidden_dim: The hidden dimension of the tensors to be reduced.
+            dtype: The data type of the tensors to be reduced.
             buffer_size_in_bytes: The requested size in bytes for each lamport buffer. The actual allocation size may be larger due to alignment requirements. The actual usable size will be NUM_LAMPORT_BUFFERS * actual_buffer_size_per_lamport_buffer.
         """
 


flashinfer/comm/trtllm_mnnvl_ar.py
```diff
@@ -159,38 +434,26 @@ def get_allreduce_mnnvl_workspace(
...
-    # This is a buffer to maintain the state of this allreduce Op
-    # [Buffer_ptr, Clear_ptr, Buffer_size, num_tokens_prev, atomic access counter]
-    buffer_flags = torch.tensor(
-        [0, 2, max_num_elements, 0, 0],
-        dtype=torch.uint32,
-        device=torch.device("cuda", mapping.local_rank),
+    # Redirect to the new workspace allocation logic. The new kernel needs the new flag buffer layout.
+    workspace = MNNVLAllreduceFusionWorkspace(
+        mapping, buffer_size_in_bytes, comm_backend_for_handle_transfer
     )
```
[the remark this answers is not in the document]
timlee0212: This can be waived as this is in legacy API path.

include/flashinfer/comm/trtllm_mnnvl_allreduce.cuh
```diff
@@ -95,581 +96,1126 @@ inline __device__ __nv_half fromFloat<__nv_half>(float val) {
...
+    *reinterpret_cast<PackedType*>(&prenormedPtr[threadOffset]) = packedAccum.packed;
+    // =============================== Rmsnorm ================================
+    PackedVec<PackedType, T> gamma;
+    gamma.packed = *reinterpret_cast<PackedType const*>(&gammaPtr[packedIdx * kELTS_PER_THREAD]);
+
+    float threadSum = 0.F;
+#pragma unroll
+    for (int i = 0; i < kELTS_PER_THREAD; i++) {
+      // FIXME: Use float square if accuracy issue
+      threadSum += toFloat<T>(packedAccum.elements[i] * packedAccum.elements[i]);
```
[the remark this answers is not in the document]
timlee0212: This can be waived as the accuracy satisfied the requirement in testing.

flashinfer/comm/mnnvl.py
```diff
@@ -979,17 +1144,23 @@ def __init__(
...
+            comm_backend_for_handle_transfer,
         )
         self.buf_size = buf_size
         self.local_device = device
 
     def lamport_initialize(self, rank: int, dtype: torch.dtype):
         self.mcast_device_memory.lamport_initialize(rank, dtype)
 
-    def get_mc_buffer(
+    def get_multicast_buffer(
```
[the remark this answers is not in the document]
timlee0212: This class is used internally, and left as a placeholder but not implemented. Thus, a breaking changes is fine. Tag @nvmbreughe for confirmation.

flashinfer/comm/trtllm_mnnvl_ar.py
```diff
@@ -25,104 +27,379 @@ def mpi_barrier():
...
-            buffer_flags,
-            launch_with_pdl,
+
+def trtllm_mnnvl_allreduce(
+    input: torch.Tensor,
+    workspace: MNNVLAllreduceFusionWorkspace,
+    launch_with_pdl: bool,
+    output: Optional[torch.Tensor] = None,
+    strategy: MNNVLAllreduceFusionStrategy = MNNVLAllreduceFusionStrategy.AUTO,
+) -> torch.Tensor:
```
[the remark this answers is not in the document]
timlee0212: Do we want this check to be on the execution path? Or should we assuming it is the user's liability to ensure it does not overflow.

flashinfer/comm/trtllm_mnnvl_ar.py
```diff
@@ -25,104 +27,379 @@ def mpi_barrier():
...
+            launch_with_pdl: Whether to launch with PDL
+            use_oneshot: Whether to use one-shot (true) or two-shot (false)
+            output: Output tensor
+            residual_out: Residual output tensor (if rmsnorm)
+            gamma: Gamma tensor (if rmsnorm)
+            epsilon: Epsilon value (if rmsnorm)
+        """
+        print(
+            f"[Rank {rank}] Inside Kernel: multicast_buffer_ptr: {multicast_buffer_ptr:x}, buffer_ptrs_dev: {buffer_ptrs_dev:x}, buffer_ptr_local: {buffer_ptr_local:x}, buffer_flags_mnnvl: {buffer_flags_mnnvl}"
+        )
```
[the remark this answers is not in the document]
timlee0212: Fixed

tests/comm/test_trtllm_mnnvl_allreduce.py
```diff
@@ -295,6 +406,9 @@ def run_mnnvl_ar_full(
         rank_failed = True
         failure_message = f"FAILED[rank={rank}]: seq_lens={seq_lens}, fusion={fusion}, dtype={dtype} failed: {e}"
         print(failure_message)
+        import traceback
+
+        print(traceback.format_exc())
```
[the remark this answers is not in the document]
timlee0212: Fixed

csrc/trtllm_mnnvl_allreduce.cu
```diff
@@ -26,77 +26,86 @@ using tvm::ffi::Optional;
...
+    cudaError_t status;
+    if (use_oneshot) {
+      status = oneshotAllreduceFusionDispatch<c_type>(params);
+    } else {
+      status = twoshotAllreduceFusionDispatch<c_type>(params);
+    }
     TVM_FFI_ICHECK(status == cudaSuccess)
-        << "twoshot_rmsnorm_dispatch_hidden_dim failed with error code "
+        << "twoshot_allreduce_dispatch_world_size failed with error code "
         << cudaGetErrorString(status);
```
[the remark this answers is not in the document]
timlee0212: Fixed

flashinfer/comm/trtllm_mnnvl_ar.py
```diff
@@ -25,104 +27,376 @@ def mpi_barrier():
...
+    NUM_LAMPORT_BUFFERS = 3
+
+    def __init__(
+        self,
+        mapping: Mapping,
+        buffer_size_in_bytes: Optional[int] = None,
+        comm_backend: Optional[CommBackend] = None,
+    ):
+        """
+        Initialize the MNNVL Allreduce Fusion Workspace. COMM_WORLD will be used for creating the workspace and synchronization. The process might hang if the intended communication group in mapping is not COMM_WORLD.
```
nvmbreughe: Is there a way we can check this?

flashinfer/comm/trtllm_mnnvl_ar.py
```diff
@@ -25,104 +27,376 @@ def mpi_barrier():
...
+MNNVL_ONE_SHOT_THRESHOLD = 64 * 1024 * 8 * 2
+
+
+class MNNVLAllreduceFusionWorkspace:
+    NUM_LAMPORT_BUFFERS = 3
+
+    def __init__(
+        self,
+        mapping: Mapping,
+        buffer_size_in_bytes: Optional[int] = None,
```
nvmbreughe, line 59: Could you provide guidance for buffer_size_in_bytes? E.g., in function of number of tokens and hidden size? Or just refer to get_required_buffer_size_bytes

flashinfer/comm/trtllm_mnnvl_ar.py
```diff
@@ -25,104 +27,376 @@ def mpi_barrier():
...
+        self.mcast_buffer_handle = McastGPUBuffer(
+            self.workspace_size_bytes,
+            mapping.tp_size,
+            mapping.tp_rank,
+            torch.device("cuda", mapping.local_rank),
+            mapping.is_multi_node(),
+            comm_backend,
+        )
+
+        # We use FP32 for sentinel value regardless of the real dtype
```
nvmbreughe, line 155: Why? Before we used the real dtype

flashinfer/comm/trtllm_mnnvl_ar.py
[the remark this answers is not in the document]
nvmbreughe: We do want this check. I recently added it because it did bite others.

flashinfer/comm/trtllm_mnnvl_ar.py
nvmbreughe, line 59: Another option would be to replace "buffer_size_in_bytes" by the parameters that get_required_buffer_size_bytes takes, and just call this from the init function. Seems more user friendly.

If you do want to just allocate a blob of memory, we could still have buffer_size_in_bytes as an **addtional** parameter that would override whatever is calculated.

csrc/trtllm_mnnvl_allreduce.cu
```diff
@@ -26,77 +26,92 @@ using tvm::ffi::Optional;
     }                                                                               \
   }()
 
-void trtllm_mnnvl_all_reduce(TensorView in, int64_t multicast_buffer_ptr, int64_t buffer_ptrs_dev,
-                             int64_t buffer_M, TensorView buffer_flags_mnnvl, int64_t nranks,
-                             int64_t rank, bool wait_for_results, bool launch_with_pdl,
-                             Optional<TensorView> out) {
-  cudaSetDevice(in.device().device_id);
-  auto stream = get_stream(in.device());
+// FIXME: is bool flag for oneshot a good idea? Trying to avoid defining a new type/enum at this
```
nvmbreughe: I don't think it is a problem

flashinfer/comm/trtllm_mnnvl_ar.py
```diff
@@ -25,104 +27,376 @@ def mpi_barrier():
     MPI.COMM_WORLD.Barrier()
 
 
+class MNNVLAllreduceFusionStrategy(Enum):
+    ONESHOT = 0
+    TWOSHOT = 1
+    AUTO = 99
+
+    @staticmethod
+    def is_one_shot(
```
nvmbreughe: Maybe rename this to "heuristic_for_one_shot" or something like that.

flashinfer/comm/trtllm_mnnvl_ar.py
timlee0212: We are using LDG.128 for all data read/write, and if the allocation is 4-byte (word) aligned, reading/writing each word can be considered atomic. Thus, it is sufficient to check in FP32 granularity regardless of the dtype. This will simplify the buffer management as decoupling the sentinel value from the dtype.

flashinfer/comm/trtllm_mnnvl_ar.py
timlee0212: I don't think that is a good design as it might give the user an impression that the allocated workspace will ONLY support the set of parameters `(max_num_tokens, hidden_dim, dtype, strategy)`

But actually, the workspace usage is quite flexible and as long as the required workspace size is smaller than the allocation, it will work. Thus, the intended usage is the user checks the required workspace size (or we can check it when calling the allreduce function, but at some cost) before using it.

timlee0212: just refer to `get_required_buffer_size_bytes`

flashinfer/comm/trtllm_mnnvl_ar.py
timlee0212: Forgot to update the doc. Fixed.

csrc/trtllm_mnnvl_allreduce.cu
timlee0212: Comment removed.

flashinfer/comm/trtllm_mnnvl_ar.py
timlee0212: Done.

flashinfer/comm/trtllm_mnnvl_ar.py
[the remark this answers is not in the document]
timlee0212: Added.

include/flashinfer/comm/trtllm_mnnvl_allreduce.cuh
```diff
@@ -95,581 +96,1126 @@ inline __device__ __nv_half fromFloat<__nv_half>(float val) {
...
+    __syncthreads();
+#endif
+    if (tid == 0) {
+#if (defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 1000))
+      asm volatile("red.async.release.global.gpu.add.u32 [%0], %1;" ::"l"(mFlagAccessPtr), "r"(1)
+                   : "memory");
+#elif (defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 700))
+      asm volatile("red.release.global.gpu.add.u32 [%0], %1;" ::"l"(mFlagAccessPtr), "r"(1)
+                   : "memory");
+#else
```
nvmbreughe, line 237: Does this mean the kernel is supported for archs < 700? What's the minimal requirement?
For the API we use the @backend_requirement decorator, which lists supported SMs. So as a minimum I think we can list: 70,80,90,100,103,110,120

Would you agree? Further back is probably not as relevant.

include/flashinfer/comm/trtllm_mnnvl_allreduce.cuh
```diff
@@ -95,581 +96,1126 @@ inline __device__ __nv_half fromFloat<__nv_half>(float val) {
...
+  // Store vectorized data
+  reinterpret_cast<PackedType*>(
+      &scatterBufDest[destTokenOffset * tokenDim * WorldSize + rank * tokenDim])[packedIdx] =
+      val.packed;
 
-  // Perform the initial input reduction
-  if (NUM_INPUTS > 0) {
-    T_IN accum[ELTS_PER_THREAD];
-    float4* accum4 = (float4*)&accum;
+  flag.clearDirtyLamportBuf(inputPtrs[rank], MNNVLTwoShotStage::SCATTER);
```
nvmbreughe, line 816: Instead of clearing the buffer here, could we assume we have properly initialized lamport buffers at the start? 
And then, at the end (e.g., right after we call PDL), we can clear the buffers, so that a next kernel using the same workspace can also assume properly initialized lamport buffers.

include/flashinfer/comm/trtllm_mnnvl_allreduce.cuh
```diff
@@ -95,581 +96,1126 @@ inline __device__ __nv_half fromFloat<__nv_half>(float val) {
...
-      for (int i = 1; i < NUM_INPUTS; i++) {
-        float4 data = *(float4*)&sh_input[i * NUM_THREADS * ITERS + k * ELTS_PER_THREAD];
-        T_IN* p_d = (T_IN*)&data;
-        for (int x = 0; x < ELTS_PER_THREAD; x++) {
-          accum[x] += p_d[x];
+        // Check validity across all elements
+#pragma unroll
+        for (int i = 0; i < kLAMPORT_ELTS_PER_PACKED; i++) {
+          valid &= !isNegZero(valuesLamport[r].elements[i]);
         }
```
nvmbreughe, line 838: As long as one of the values in WorldSize isNegZero, we will keep reading the others over and over again. Perhaps these are all cache hits, in which case: ignore my comment. Otherwise, would it make sense to break out of the inner loop (kLAMPORT_ELTS_PER_PACKED) early?

include/flashinfer/comm/trtllm_mnnvl_allreduce.cuh
yzh119: This macro is fairly common in barrier and semaphore utility functions, largely due to the memory consistency qualifiers introduced with the Volta architecture. For example, CUTLASS uses a similar pattern:

https://github.com/NVIDIA/cutlass/blob/e67e63c331d6e4b729047c95cf6b92c8454cba89/include/cutlass/barrier.h#L116-L129

That said, I believe our usage here simply follows established convention. Given that our minimum supported architecture is sm_75, we shouldn't actually need these qualifiers.

timlee0212: The kernel needs multicast to work, which at least requires SM90 and needs NVSwitch

include/flashinfer/comm/trtllm_mnnvl_allreduce.cuh
timlee0212: We use triple buffer so you can move the buffer clear anywhere. I found the current arrangement get the best performance.

If your assumption is using single buffer, clear the buffer at the end then assume the buffer is initialized for the next kernel, this won't work as it requires membar.sys which is very expensive. Moreover, we need to order the buffer write of the next kernel, other ranks after the buffer clear of this kernel, which has to use a flag and invalidate the usage of lamport protocol.

include/flashinfer/comm/trtllm_mnnvl_allreduce.cuh
timlee0212: It will be cache hit if it is still invalid. 

For the inner loop, I had a version that only check one element for validity. But there is no architecture guarantee that a 128B read/write is atomic, which mean beak the inner loop early could, in rare case, cause wrong result.

flashinfer/comm/mnnvl.py
```diff
@@ -821,60 +962,105 @@ def _alloc_mn_mcast_mem(self, buf_size: int):
...
-        all_fabric_handles = comm.allgather(my_fabric_handle.data)
+        if (
+            self._shareable_handle_type
+            == cuda.CUmemAllocationHandleType.CU_MEM_HANDLE_TYPE_FABRIC
+        ):
+            # All-gather fabric handles
+            all_shareable_uc_handles = self.comm_backend.allgather(
+                local_shareable_uc_handle.data
+            )
+        else:
```
nvmbreughe: [nit] Add a comment regarding handle type in this case: CU_MEM_HANDLE_TYPE_POSIX_FILE_DESCRIPTOR

flashinfer/comm/mnnvl.py
```diff
@@ -821,60 +962,105 @@ def _alloc_mn_mcast_mem(self, buf_size: int):
...
             )
         else:
-            mc_fabric_handle = None
-
-        # Broadcast multicast handle
-        mc_fabric_handle_data = comm.bcast(
-            mc_fabric_handle.data if mc_fabric_handle else None, root=0
-        )
+            shareable_mc_handle = None
+        if (
```
nvmbreughe: With all the branching on this handle type, would it make sense to build a class that encapsulate this logic?

Roughly thinking:

```
class ABCCommBufferAllocator:
def allocate_UnicastBuffer()
def allocate_MulitcastBuffer()

class CUFabricCommBufferAllocator(ABCCommBufferAllocator)
class CUPosixFDCommBufferAllocator(ABCCommBufferAllocator)
```

flashinfer/comm/mnnvl.py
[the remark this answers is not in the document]
nvmbreughe: agreed

flashinfer/comm/mnnvl.py
timlee0212: Branching logic no longer exists in the latest commit.

flashinfer/comm/mnnvl.py
timlee0212: Refactored in the latest commit.

flashinfer/comm/trtllm_mnnvl_ar.py
```diff
@@ -25,102 +27,418 @@ def mpi_barrier():
...
+        elem_size = torch.tensor([], dtype=dtype).element_size()
+        if strategy == MNNVLAllreduceFusionStrategy.AUTO:
+            strategy = MNNVLAllreduceFusionStrategy.select_strategy(
+                tp_size, num_tokens, hidden_dim, dtype
+            )
+
+        if strategy == MNNVLAllreduceFusionStrategy.ONESHOT:
+            # For one-shot, each rank needs to store num_tokens * tp_size tokens
+            buffer_size = num_tokens * hidden_dim * tp_size * elem_size
+        else:
```
nvmbreughe, line 217: As discussed we would like to make one-shot vs two-shot at runtime, so here I would add MAX_ONE_SHOT_SIZE= MNNVL_ONE_SHOT_THRESHOLD

and buffer_size = max(...., MAX_ONE_SHOT_SIZE). And add a comment explaining that this workspace can be used for both one-shot and two-shot.

nvmbreughe: Forgot to press submit. Ignore this.

nvmbreughe (approved):

flashinfer/comm/trtllm_mnnvl_ar.py
```diff
@@ -25,102 +27,451 @@ def mpi_barrier():
...
+        if buffer_size_in_bytes is None:
+            assert (
+                max_num_tokens is not None
+                and hidden_dim is not None
+                and dtype is not None
+            ), (
+                "max_num_tokens, hidden_dim, and dtype must be provided if buffer_size_in_bytes is not provided."
+            )
+
+            # If the user want to explictly use one-shot pass the threshold, which requires larger workspace size,
```
nvmbreughe: Maybe we can add a check for this:
if (one_shot_size_bytes > MNNVL_ONE_SHOT_THRESHOLD) and strategy == MNNVLAllreduceFusionStrategy.ONESHOT:
raise ValueError("Please set the workspace when using oneshot beyond treshold).

flashinfer/comm/trtllm_mnnvl_ar.py
```diff
@@ -25,102 +27,451 @@ def mpi_barrier():
...
+        dtype: Optional[torch.dtype] = None,
+        buffer_size_in_bytes: Optional[int] = None,
+        comm_backend: Optional[CommBackend] = None,
+    ):
+        """
+        Initialize the MNNVL Allreduce Fusion Workspace. comm_backend will be used for creating the workspace and synchronization. If not provided, MPIBackend will be used which will use COMM_WORLD for synchronization.
+
+        Args:
+            mapping: Mapping configuration containing rank info
+            buffer_size_in_bytes: The requested size in bytes for each lamport buffer. The actual allocation size may be larger due to alignment requirements. The actual usable size will be NUM_LAMPORT_BUFFERS * actual_buffer_size_per_lamport_buffer.
```
nvmbreughe, line 76: [nit] document the new parameters, and also bring up the point that buffer_size_in_bytes is needed when we force one-shot.

timlee0212: Done

flashinfer/comm/trtllm_mnnvl_ar.py
timlee0212: I assume this is considered resolved?

This function is used purely as an utility for calculating required workspace given a problem size and implementaation. It should not implement the logic we discuss, which is implemented in the constructor.

flashinfer/comm/trtllm_mnnvl_ar.py
```diff
@@ -25,102 +27,460 @@ def mpi_barrier():
...
-            residual,
-            buffer_flags,
-            launch_with_pdl,
+
+def trtllm_mnnvl_allreduce(
+    input: torch.Tensor,
+    workspace: MNNVLAllreduceFusionWorkspace,
+    launch_with_pdl: bool,
+    output: Optional[torch.Tensor] = None,
+    strategy: MNNVLAllreduceFusionStrategy = MNNVLAllreduceFusionStrategy.AUTO,
```
wenscarl, line 313: Before calling this API, users are expected to create
`workspace = MNNVLAllreduceFusionWorkspace(...)`,
which internally decides whether to use the one-shot or two-shot strategy.
But what happens if the framework (user) chooses a different strategy than the one determined by the workspace?

nvmbreughe: It's possible, and by design actually.

There is a check "if not workspace.is_buffer_size_sufficient(...)" that happens before the allreduce executes.

nvmbreughe: The longer answer is that: (1) the workspace creation will use a heuristic to find the optimal situation: create one-shot space (the largest one for the same problem space) up to a certain threshold, and two-shot after that. (*) 
Then at runtime (2) when allreduce gets called, we check the heuristic again, and use one-shot/two-shot based on that.

However, this means we could fall below the one-shot threshold at runtime. Therefore, internally we also made sure that enough memory was allocated for the largest one-shot problem sizes (refer to trtllm_mnnvl_ar.py:L88).

Now, if a user intentionally wants to use one-shot all the time, they are expected to calculate the memory themselves (see the comments).

wenscarl (approved):
LGTM.

flashinfer/comm/trtllm_mnnvl_ar.py
```diff
@@ -25,102 +27,460 @@ def mpi_barrier():
...
+        dtype: torch.dtype,
+        strategy: MNNVLAllreduceFusionStrategy = MNNVLAllreduceFusionStrategy.AUTO,
+    ) -> bool:
+        """
+        Calculate the required buffer size for a given problem size.
+        """
+        required_buffer_size = self.get_required_buffer_size_bytes(
+            tp_size, num_tokens, hidden_dim, dtype, strategy
+        )
+        if required_buffer_size > self.buffer_size_bytes:
```
nvpohanh, line 191: simplify this to: `return required_buffer_size <= self.buffer_size_bytes`

flashinfer/comm/trtllm_mnnvl_ar.py
```diff
@@ -25,102 +27,460 @@ def mpi_barrier():
...
-            gamma,
-            epsilon,
-            residual,
-            buffer_flags,
-            launch_with_pdl,
+
+def trtllm_mnnvl_allreduce(
+    input: torch.Tensor,
+    workspace: MNNVLAllreduceFusionWorkspace,
+    launch_with_pdl: bool,
```
nvpohanh, line 311: should we just make this optional and default to `True`?

flashinfer/comm/trtllm_mnnvl_ar.py
```diff
@@ -25,102 +27,460 @@ def mpi_barrier():
...
+        1. Scatter each GPU's input shard to other ranks. Each rank will received all shards of a slice of tokens.
+        2. Each rank perform reduction on the local tokens.
+        3. Each rank broadcast the result to all ranks.
+        Suitable for large data size and is optimized for balancing throughput and latency.
+
+    Args:
+        input: Local Input Shard [num_tokens, hidden_dim]
+        workspace: MNNVLAllreduceFusionWorkspace
+        launch_with_pdl: Whether to launch with PDL
+        output: Output tensor to store the result, empty tensor will be created if not provided.
```
nvpohanh, line 332: Not a blocker for this PR, but just a question: is there a plan to fuse FP8-Quant and NVFP4-Quant into this kernel later?

Currently, the frameworks are using the AR+Norm+Q fused kernels in https://github.com/flashinfer-ai/flashinfer/blob/main/flashinfer/comm/trtllm_ar.py . It becomes quite difficult for us to choose among these different options.

cc @nvmbreughe

PR #2130, opened 2025-11-21
A unified API for the MNNVL and single-node/multi-GPU AllReduce kernels.

nvmbreughe:
## 📌 Description

A unified API for the MNNVL and single-node AllReduce kernels.
* This introduces the API's `create_allreduce_fusion_workspace`, and `allreduce_fusion`
* The backend ("trtllm" or "mnnvl") is chosen during workspace creation. We can either pick it explicitly, or use the "auto" backend to have a heuristic pick the best backend.
* The API can be used for both single-node allreduce, as well as for multi-node allreduce.

Test with
```
mpirun -np 4 pytest tests/comm/test_allreduce_unified_api.py
mpirun -np 4 pytest tests/comm/test_allreduce_negative.py
```

note: mpirun is needed for the mnnvl backend, as illustrated in the test commands above

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * Unified AllReduce Fusion API with multi-backend support (auto/TRTLLM/MNNVL) and public workspace types
  * Common workspace base with lifecycle management and automatic cleanup warnings

* **Bug Fixes / Validation**
  * Stronger input/workspace validation with aggregated error messages
  * Idempotent destroy semantics for safer resource cleanup

* **Deprecations**
  * Legacy AllReduce APIs deprecated in favor of the unified API

* **Tests**
  * Expanded negative, integration, and cross-backend correctness tests plus MPI/CUDA test helpers

### DIFF
--- a/flashinfer/comm/__init__.py
+++ b/flashinfer/comm/__init__.py
@@ -48,6 +48,8 @@ from .allreduce import TRTLLMAllReduceFusionWorkspace as TRTLLMAllReduceFusionWo
 from .allreduce import allreduce_fusion as allreduce_fusion
 from .allreduce import (
     create_allreduce_fusion_workspace as create_allreduce_fusion_workspace,
+)
+
 # MNNVL A2A (Throughput Backend)
 from .trtllm_moe_alltoall import MoeAlltoAll as MoeAlltoAll
 from .trtllm_moe_alltoall import moe_a2a_combine as moe_a2a_combine


flashinfer/comm/allreduce.py
```diff
@@ -0,0 +1,856 @@
...
+
+def _workspace_creation_heuristic(
+    suitable_backends: list[str],
+    backend: str,
+    world_size: int,
+    rank: int,
+    max_token_num: int,
+    hidden_dim: int,
+    dtype: torch.dtype,
+    topology: str,
```
yzh119:
```suggestion
    topology: Literal["single_node", "multi_node"],
```

flashinfer/comm/allreduce.py
```diff
@@ -0,0 +1,746 @@
...
+    heuristic_func=_workspace_creation_heuristic,
+)
+def create_allreduce_fusion_workspace(
+    backend: Literal["trtllm", "mnnvl", "auto"] = "auto",
+    world_size: int = None,
+    rank: int = None,
+    max_token_num: int = None,
+    hidden_dim: int = None,
+    dtype: torch.dtype = None,
+    topology: str = "single_node",
```
nvcastet: I don't think it is needed longer term since we will use the same pytorch symmetric API to allocate symmetric memory for single and multi-node (under the cover pytorch/NCCL/NVSHMEM will detect platform and decides the right mem allocation handle)

flashinfer/comm/allreduce.py
```diff
@@ -0,0 +1,746 @@
...
+# ============================================================================
+# MAIN API - NO backend parameter, infers from workspace type
+# ============================================================================
+
+
+def allreduce_fusion(
+    input: torch.Tensor,
+    workspace: AllReduceFusionWorkspace,
+    pattern: int,
+    launch_with_pdl: bool = False,
```
nvcastet, line 456: Why the advantage to give pdl control to the user?

flashinfer/comm/allreduce.py
```diff
@@ -0,0 +1,746 @@
...
+    - AllReduce + Residual + RMSNorm + Quantization (FP8/FP4)
+
+    **Note on Workspace Reusability:**
+    You can reuse the same workspace with different (token_num, hidden_dim) combinations
+    as long as `workspace.is_sufficient_for(token_num, hidden_dim, tp_size, dtype)` returns True.
+
+    Args:
+        input: Input tensor [token_num, hidden_dim]
+        workspace: Workspace object (type determines backend)
+        pattern: Fusion pattern (AllReduceFusionPattern constant, 0-5)
```
nvcastet, line 490: All they all 2-kernel overlap or some are real fusion kernels?

flashinfer/comm/allreduce.py
```diff
@@ -0,0 +1,746 @@
...
+
+
+@backend_requirement(
+    backend_checks={
+        "trtllm": _trtllm_workspace_check,
+        "mnnvl": _mnnvl_workspace_check,
+    },
+    heuristic_func=_workspace_creation_heuristic,
+)
+def create_allreduce_fusion_workspace(
```
nvcastet, line 286: Could `create_allreduce_fusion_workspace` take an optional `workspace` argument? If `workspace` is big enough or too big this is a noop (maybe just updating backend selection). If it is too small, destroy current workspace and allocate a bigger one.

When we switch to mem pool, we should be able to call `create_allreduce_fusion_workspace` at each forward pass  and memory will just get reused from the mempool (instead of new allocations).
CC @Amir-19

flashinfer/comm/allreduce.py
```diff
@@ -0,0 +1,746 @@
...
+    Backend selection (checks + heuristics) handled by @backend_requirement decorator.
+
+    **Important: Workspace Reusability**
+    The workspace is allocated based on the total size (max_token_num * hidden_dim * dtype_size).
+    You can reuse the same workspace with different shapes as long as the total size fits:
+
+    - Workspace(max_token_num=2048, hidden_dim=4096) can handle:
+      - (token_num=2048, hidden_dim=4096) ✓
+      - (token_num=1024, hidden_dim=4096) ✓
+      - (token_num=4096, hidden_dim=2048) ✓ (same total size)
```
nvcastet, line 310: I only see FW adjusting the num of tokens but `hidden_dim` should be fixed per model.

flashinfer/comm/allreduce.py
```diff
@@ -0,0 +1,746 @@
...
+    Examples:
+        >>> # Auto-select best backend based on topology
+        >>> workspace = create_allreduce_fusion_workspace(
+        ...     backend="auto",
+        ...     world_size=8,
+        ...     rank=0,
+        ...     max_token_num=2048,
+        ...     hidden_dim=4096,
+        ...     dtype=torch.bfloat16,
+        ...     topology="single_node"
```
nvcastet, line 349: Could we had a check now to detect topology? before we switch to the mempool allocation?

flashinfer/comm/allreduce.py
nvmbreughe: with one-shot mnnvl it's real fusion. And I think similar for the trtllm_ar kernels. It's just two-shot mnnvl that is the 2-kernel overlap.

flashinfer/comm/allreduce.py
nvmbreughe: We have been doing this for all our APIs, but I am not sure why. Maybe because not all archs support it?

flashinfer/comm/allreduce.py
nvmbreughe: Hi @nvcastet , could you elaborate what you mean with a topology check?

flashinfer/comm/allreduce.py
nvmbreughe: Done. Thanks!

flashinfer/comm/allreduce.py
nvmbreughe: What's the advantage of having
`workspace= create_allreduce_fusion_workspace(old_workspace)
`
vs

`workspace = old_workspace if condition else create_allreduce_fusion_workspace()`

?

flashinfer/comm/allreduce.py
nvmbreughe: Discussed this live, and resulted in creating the one-shot vs two-shot decision to be delayed until runtime (aka not during workspace creation time).

wenscarl:
test_trtllm_mnnvl_allreduce_custom_comm.py was not updated in #2118. Could you update it in this PR?

flashinfer/comm/allreduce.py
```diff
@@ -0,0 +1,702 @@
...
+        ... )
+        >>>
+        >>> # Pre-allocate output tensors
+        >>> prenorm = torch.empty_like(hidden_states)
+        >>> normed = torch.empty_like(hidden_states)
+        >>>
+        >>> # Call fusion - backend inferred from workspace type
```
yzh119, line 539: Good design :)

tests/comm/test_allreduce_unified_api.py
```diff
@@ -0,0 +1,311 @@
...
+    if backend in ("trtllm", "auto"):
+        init_torch_distributed_from_mpi()
+        process_group = dist.group.WORLD
+
+    if local_rank == 0:
+        print(f"Running AllReduce test with {world_size} ranks, backend={backend}")
+        print(f"Rank {rank} using GPU {torch.cuda.current_device()}")
+
+    eps = 1e-5
+    torch.manual_seed(42 + rank)
```
yzh119, line 212: Is it designed to make sure each rank was initialized in a different way? (I'm curious if we should have a standard for data initialization for communication APIs).

tests/comm/test_allreduce_unified_api.py
```diff
@@ -0,0 +1,311 @@
...
+            print("output[1] (first 10 values):", output[1].flatten()[:10])
+            print(
+                "reference_output[1] (first 10 values):",
+                reference_output[1].flatten()[:10],
+            )
+
+    torch.testing.assert_close(
+        output[0],
+        reference_output[0],
+        rtol=0.05,
```
yzh119, line 113: We might consider different tolerance for different output data types in the future (if we add more precision output data types).

flashinfer/comm/allreduce.py
nvcastet: It was for the case when you would support growing the workspace:
If needed_size > old_workspace.size:
   old_workspace.destroy()
   return new_workspace

But we discussed that we may not be able to destroy current workspace due to previous CUDA graph captures.
So you can ignore my comment for now.

flashinfer/comm/allreduce.py
nvcastet: Something like that check: https://github.com/pytorch/pytorch/blob/eb66e1d2b590865be962b8677acc31728d3ad953/aten/src/ATen/cuda/PeerToPeerAccess.cpp#L146-L180

If fabric is enabled you would have a fabric workspace (for MNNVL) otherwise to default to single-node workspace.

tests/comm/test_allreduce_unified_api.py
nvmbreughe: Good catch! I think it must have been a copy paste error actually: the data is only initialized for rank 0, so this would not have an effect.

flashinfer/comm/allreduce.py
yzh119: the main purpose of having this control is for debugging as we observed some issues of pdl on sm121.

For deployment it should always be turned on IMO, so I suppose the default value should be None and we will call https://github.com/flashinfer-ai/flashinfer/blob/3a301a102c6a2ba7b52030910899a405bad156b0/flashinfer/utils.py#L615 to automatically determine it.

Issue #978, opened 2025-03-28
`top_k_top_p_sampling_from_logits` incompatible with torch.compile + CUDAGraph


sharvil:
Python version: 3.10
PyTorch version: 2.6.0
FlashInfer version: 0.2.3+cu124torch2.6

Minimal repro:

```py
import torch

from flashinfer.sampling import top_k_top_p_sampling_from_logits


def sample(x):
  return top_k_top_p_sampling_from_logits(
    logits=x,
    top_k=0,
    top_p=1.0,
  )

sample = torch.compile(sample, mode='reduce-overhead', fullgraph=True)
sample(torch.randn([1, 32]).cuda())
```

Produces the following error:

```
torch._dynamo.exc.Unsupported: call_function UserDefinedClassVariable(<class 'types.SimpleNamespace'>) [] {'sampling_from_probs': NestedUserFunctionVariable(), 'top_p_sampling_from_probs': NestedUserFunctionVariable(), 'top_k_sampling_from_probs': NestedUserFunctionVariable(), 'min_p_sampling_from_probs': NestedUserFunctionVariable(), 'top_k_top_p_sampling_from_probs': NestedUserFunctionVariable(), 'top_p_renorm_probs': NestedUserFunctionVariable(), 'top_k_renorm_probs': NestedUserFunctionVariable(), 'top_k_mask_logits': NestedUserFunctionVariable(), 'chain_speculative_sampling': NestedUserFunctionVariable()}

from user code:
   File ".../compile.py", line 7, in sample
    return top_k_top_p_sampling_from_logits(
  File "/home/sharvil/.conda/envs/lmnt/lib/python3.10/site-packages/flashinfer/sampling.py", line 810, in top_k_top_p_sampling_from_logits
    masked_logits = top_k_mask_logits(logits, top_k)
  File "/home/sharvil/.conda/envs/lmnt/lib/python3.10/site-packages/flashinfer/sampling.py", line 1129, in top_k_mask_logits
    return get_sampling_module().top_k_mask_logits(
  File "/home/sharvil/.conda/envs/lmnt/lib/python3.10/site-packages/flashinfer/sampling.py", line 393, in get_sampling_module
    _sampling_module = SimpleNamespace(
```

Setting `fullgraph=False` succeeds.

sharvil:
Fails on nightly as well:


```
[...]

PR #2132, opened 2025-11-23
feat: add seed offset args to sampler to allow cuda graph support

ksukrit:
## 📌 Description

This PR adds optional seed/offset args to all the sampler functions to prevent calling the `get_seed_and_offset` function. If that function is not called, we can potentially make the sampler forward call as part of CUDAGraph and use that to replay it.

We can directly compute the Seed/offset values, before launching the graph in a similar way to as it is being done in the current method and pass them when making the flashinfer call

## 🔍 Related Issues
#978 : top_k_top_p_sampling_from_logits incompatible with torch.compile + CUDAGraph

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * Optional seed and offset parameters added to sampling APIs to enable deterministic RNG control while remaining optional.

* **Tests**
  * New tests verify reproducible sampling when using the same seed/offset and variability when different values are used.

### DIFF
--- a/tests/utils/test_sampling.py
+++ b/tests/utils/test_sampling.py
@@ -726,6 +726,84 @@ def test_check_tensor_param_top_k(batch_size, vocab_size, k):
     assert samples.shape == normalized_prob.shape
 
 
+@pytest.mark.parametrize("batch_size", [1, 99, 989])
+@pytest.mark.parametrize("vocab_size", [111, 32000, 128256])
+def test_sampling_from_probs_seed_offset_reproducibility(batch_size, vocab_size):
+    """Test that explicit seed/offset produces reproducible results."""
+    torch.manual_seed(42)
+    pre_norm_prob = torch.rand(batch_size, vocab_size, device="cuda:0")
+    normalized_prob = pre_norm_prob / pre_norm_prob.sum(dim=-1, keepdim=True)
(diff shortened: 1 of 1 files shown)


yzh119 (commented):
Hi @ksukrit it's a good feature to have.

Would you mind adding unittest for this?

Also, it would be to support an array of seed/offset, where each of them is a int64 array, to describe per-batch seed/offset. (It requires some modification to the kernels).

ksukrit:
> Hi @ksukrit it's a good feature to have.
> 
> Would you mind adding unittest for this?
> 
> Also, it would be to support an array of seed/offset, where each of them is a int64 array, to describe per-batch seed/offset. (It requires some modification to the kernels).

Sure thing, will add the unittests for this. But I had a quick question, right now the kernels just take a single int offset/seed value right ? The main purpose of this was to avoid the `get_seed_and_offset` and offset call and allow cuda graph capture. 

Is it okay if I take up the seed/offset array changes for the batch in a separate PR @yzh119

yzh119:
> Is it okay if I take up the seed/offset array changes for the batch in a separate PR

Sure, we can do that in a separate PR.

PR #2134, opened 2025-11-23
fix(trtllm): reset negative strideBatch to 0 for ragged KV layout to …

YAMY1234:
## Description

Fix TMA descriptor failures for ragged KV layouts in the TRT-LLM FMHA path.

When using `trtllm_ragged_attention_deepseek` with a non-paged, non-contiguous KV layout (ragged KV), the KV batch dimension is effectively collapsed (shape `[head_dim, sum_seq_lens_kv, num_heads_kv, 1]`). However, `kStrideBatch` / `vStrideBatch` can be set to a large `numel()`-based value that overflows a 32-bit `int` and becomes a negative sentinel. This negative stride is then interpreted as a `uint64_t` in `buildNdTmaDescriptor`, producing an enormous `strideInBytes` and causing `cuTensorMapEncodeTiled` to fail with:

> Error: Failed to initialize the TMA descriptor due to invalid argument

This PR updates `makeStrideKv` so that for ragged KV layouts (`!isPagedKv && !isContiguousKv`), any negative `strideBatch` is treated as a sentinel and clamped to `0`. This matches the actual memory layout (no real batch stride for the collapsed batch dimension) and prevents overflow in the TMA descriptor.

## 🔍 Related Issues

- SGLang: DeepSeek-R1 + `trtllm_ragged_attention_deepseek` on SM100 with long KV (e.g., 128k total KV tokens) failing at `buildNdTmaDescriptor` with invalid TMA configuration.

[[Bug] [DeepSeek-R1] Error: Failed to initialize the TMA descriptor due to invalid argument on B200](https://github.com/sgl-project/sglang/issues/13775)

### Pre-commit Checks

- [x] I have installed `pre-commit` (e.g., `pip install pre-commit`).
- [x] I have run `pre-commit run --all-files` and fixed any reported issues.

## Tests

## Reviewer Notes

- The change is intentionally minimal and only affects ragged KV layouts.
- The goal is to keep the existing device kernels intact while fixing the host-side TMA descriptor construction for this layout.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented negative batch stride for ragged key/value cache layouts, avoiding invalid descriptor errors in large or irregular attention workloads.

* **Tests**
  * Added an integration test that reproduces large ragged KV scenarios to verify the clamped stride behavior on CUDA systems.
  * Updated a communication test to skip when required GPUs aren't available instead of failing, improving test robustness.

### DIFF
--- a/tests/attention/test_trtllm_ragged_kv_stride.py
+++ b/tests/attention/test_trtllm_ragged_kv_stride.py
@@ -23,9 +23,9 @@ def test_trtllm_ragged_kv_large_stride_overflow():
 
     device = torch.device("cuda")
     compute_capability = get_compute_capability(device)
-    if compute_capability[0] < 10:
+    if compute_capability[0] != 10:
         pytest.skip(
-            f"TRTLLM ragged attention requires SM100+ GPUs, got sm{compute_capability[0]}{compute_capability[1]}"
+            f"TRTLLM-gen ragged attention requires SM100 and SM103 GPUs, got sm{compute_capability[0]}{compute_capability[1]}"
         )
 
     torch.manual_seed(42)


YAMY1234:
> Thanks for the bugfix and it looks good to me overall, would you mind adding a unittest for the case
> 
> > However, kStrideBatch/vStrideBatch can be set to a large numel()-based value that overflows into a negative int32 sentinel. This negative stride is then cast to in buildNdTmaDescriptor, producing an enormous strideInBytes and causing cuTensorMapEncodeTiled to fail with:
> 
> which used to fail but works after your fix?

Thanks for the suggestion! Unittest is added @yzh119

PR #2175, opened 2025-12-04
fix: compile flags for trtllm fmha_v2 

jimmyzho:
## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Chores**
  * Removed noisy runtime console prints during build/generation.
  * Updated CUDA compiler requirements to target CUDA 12 and added a new compiler flag for compatibility.

* **Bug Fixes**
  * Added an early check that raises a clear error on unsupported GPU devices (SM120a), preventing misruns.

* **Tests**
  * Test now skips automatically when the required SM120a GPU support is not present.

### DIFF
--- a/flashinfer/jit/attention/modules.py
+++ b/flashinfer/jit/attention/modules.py
@@ -1901,7 +1901,7 @@ def gen_trtllm_fmha_v2_module() -> JitSpec:
     source_paths = kernel_paths + [binding_source_path]
 
     nvcc_flags = current_compilation_context.get_nvcc_flags_list(
-        supported_major_versions=[8, 9, 10, 11, 12]
+        supported_major_versions=[12]
     )
     nvcc_flags.append(f"-I{jit_env.FLASHINFER_CSRC_DIR / 'fmha_v2'}")
     nvcc_flags.append("-Wno-deprecated-gpu-targets")
--- a/flashinfer/prefill.py
+++ b/flashinfer/prefill.py
@@ -3603,6 +3603,8 @@ def fmha_v2_prefill_deepseek(
         If return_lse is True, the output will be a tuple of two tensors, the first is the output tensor, the second is the lse tensor.
         If return_lse is False, the output will be a single tensor.
     """
+    if not is_sm120a_supported(query.device):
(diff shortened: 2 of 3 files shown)


flashinfer/jit/attention/fmha_v2/generator_utils.py
```diff
@@ -3711,10 +3711,10 @@ def generate_files(specs_names):
     ]
     if "CUDA_PATH" in os.environ:
         cmd[0] = os.environ["CUDA_PATH"] + "/bin/" + cmd[0]
-    print('Running command "{}" to build "bin/print_traits.exe":'.format(" ".join(cmd)))
+    # print('Running command "{}" to build "bin/print_traits.exe":'.format(" ".join(cmd)))
```
yzh119, line 3714: Are these changes relevant to the PR?

flashinfer/jit/attention/modules.py
```diff
@@ -1901,9 +1901,10 @@ def gen_trtllm_fmha_v2_module() -> JitSpec:
     source_paths = kernel_paths + [binding_source_path]
 
     nvcc_flags = current_compilation_context.get_nvcc_flags_list(
-        supported_major_versions=[10, 11, 12]
+        supported_major_versions=[8, 9, 10, 11, 12]
```
bkryu: Do we expect trtllm fmha v2 to run on Ampere and Hopper cards? Asking because I've never worked with the trtllm fmha v2 module

flashinfer/jit/attention/fmha_v2/generator_utils.py
jimmyzho: not really, just commenting these out from the original trtllm script to clean up the stdout

flashinfer/jit/attention/modules.py
jimmyzho: That's a good point - this backend technically does support >= SM80, but since this backend is an exclusive use case for the  6000D cards, I believe we can keep our support surface to just SM120

yzh119 (approved):

PR #2111, opened 2025-11-19
refactor: update fa3 codebase and fix hopper unittest [part 1]

yzh119:
## 📌 Description

This PR refactors the out-dated fa3 codebase, more specifically, for page_size>1, the page offset calculation is performed inside the kernel, without the need of a standalone function call to block_sparse_indices_to_vector_sparse_offsets, and optimize the kv_offset calculation with prefetching and shuffling.

This PR also fixes the failed unittest on hopper.

However, the FA3 structure in our codebase is still terrible outdated without important features such as `IntraWGOverlap` and `RescaleOBeforeGemm`, will follow up soon in a later PR.

## 🔍 Related Issues

This PR should fixes https://github.com/flashinfer-ai/flashinfer/issues/1647

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [x] Tests have been added or updated as needed.
- [x] All tests are passing (`unittest`, etc.).

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced consistent K/V page-strides and stride_n for sparse paged attention with runtime validation.

* **New Features**
  * Exposed page-interval stride metadata for paged KV caches.
  * Expanded FP8 support: per-head/per-tensor scale propagation and configurable output dtype; added ragged-KV execution path and benchmarks.

* **Refactor**
  * Unified sparse attention to a page-based addressing model and simplified workspace/epilogue behavior; removed legacy block-to-vector conversion helpers and related APIs.

* **Tests**
  * Added FP8 ragged/paged tests, expanded page-size coverage, and removed obsolete vector-sparse unit test.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #2192, opened 2025-12-09
refactor: update fa3 codebase [part 2]

yzh119:
## 📌 Description

Follow up of #2111 , adding more optimizations in fa3 (mostly around fp8).

## 🔍 Related Issues

#2111 

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Reviewer Notes

cc @Edenzzzz 

## Release Notes

* **New Features**
  * Enhanced FP8 quantization support with optimized data transpose operations.
  * Added L2 cache-aware scheduling for improved performance.
  * Introduced Longest-Processing-Time scheduling optimization.
  * New FlashInfer vs FA3 benchmarking suite.

* **Performance Improvements**
  * Optimized kernel pipelines for Cutlass 3.6+ compatibility.
  * Improved paged KV cache handling and memory efficiency.
  * Enhanced attention computation across variable sequence lengths.

### DIFF
--- a/benchmarks/bench_fa3_comparison.py
+++ b/benchmarks/bench_fa3_comparison.py
@@ -6,10 +6,8 @@ random.seed(42)
 device = "cuda"
 dtype = torch.float16
 
-import sys
-
-sys.path.insert(0, "/home/zhye/flash-attention/hopper")
 from flash_attn_interface import flash_attn_varlen_func as fa3_varlen_func
+from flash_attn_interface import flash_attn_with_kvcache as fa3_kvcache_func
 import flashinfer
 from flashinfer.testing import (
     bench_gpu_time_with_cuda_event as bench_gpu_time_with_cupti,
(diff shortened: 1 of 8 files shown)


include/flashinfer/attention/scheduler.cuh
```diff
@@ -796,7 +796,7 @@ inline cudaError_t PrefillPlan(void* float_buffer, size_t float_workspace_size_i
   return cudaSuccess;
 }
 
-inline float cost_function(int qo_len, int kv_len) { return 2 * float(qo_len) + kv_len; }
+inline float cost_function(int qo_len, int kv_len) { return 0.05 * float(qo_len) + kv_len; }
```
Edenzzzz, line 799: Is this based on profiling? This will also change the cost for persistent FA2?
(I tried fitting a quadratic cost function [over a bunch of seq lens ](https://github.com/Edenzzzz/flashinfer/blob/6582d5f04cb9b6c962682f33473f5d95683d253e/profiler/scan_batch_attention.py#L138)for persistent FA2 + overlap, didn't work well)
I think FA4 just strides by num CTAs without any load balancing? I plan to try opportunistically fetching from the two work queues next in the persistent FA2, like in POD

yzh119: Yes it's based on profiling.

But as you mention, none of them work well (the cost model can not model the interference of L2, etc).

Edenzzzz: FA3's dynamic LPT scheduler (it might not be necessary to still stride by `num_ctas`)
https://github.com/Dao-AILab/flash-attention/blob/c783ab2f7e05ba1cd79ecfe0e6e109a4e3f6e542/hopper/tile_scheduler.hpp#L338

qqbbiu:
BatchPrefillWithRaggedKVCacheWrapper fp8 kernel sliding window hang, but BatchPrefillWithPaggedKVCacheWrapper works, is there any bug?

include/flashinfer/attention/scheduler.cuh
Edenzzzz: In FA3 LPT/load balancing only happens among the query tiles whose KV cache can all fit into the L2 cache to avoid eviction, which may take a significant amount of time, we can do the same for FA2?
https://github.com/Dao-AILab/flash-attention/blob/c783ab2f7e05ba1cd79ecfe0e6e109a4e3f6e542/hopper/tile_scheduler.hpp#L254

Issue #2170, opened 2025-12-03
[Feature Request] Expand API Logging to include Data Dump & Replay mode


bkryu:
#2108 and #2153 recently introduced API logging feature has been recently introduced in. 

The API logging feature:
* Allows logging level to be controlled via `FLASHINFER_LOGLEVEL` and destination via `FLASHINFER_LOGDEST`.
* Has the ability to log input and output tensor shapes, scalar values, and various metadata.
* More information on [logging in the docs](https://docs.flashinfer.ai/logging.html)

An example at level 3 or is:
```
================================================================================
[2025-11-20 17:51:58] FlashInfer API Call: BatchDecodeWithPagedKVCacheWrapper.__init__
--------------------------------------------------------------------------------
Positional input arguments:
  arg[0]:
    <flashinfer.decode.BatchDecodeWithPagedKVCacheWrapper object at 0x1234399e3410>
  arg[1]:
    Tensor(
      shape=(134217728,)
      stride=(1,)
      dtype=torch.int8
      device=cuda:0
      requires_grad=False
      is_contiguous=True
    )
  arg[2]:
    'HND'
Keyword input arguments:
  use_cuda_graph=
    True
  use_tensor_cores=
    False
...
```
which helps reproduce a lot of errors. 

However, a further desired feature is a "flight recorder" mode to support data recording of API calls for the purposes of reproducing testing and benchmark errors.

In addition to the existing logging features, the flight recorder & replay would ideally
* Enable replaying sequences of API calls for regression testing, benchmarking, or issue reproduction.
* Log input and output values for all data types.
* Include sufficient metadata to support deterministic replay (e.g., API version, library build identifier).
* A streamlined Replay Mode for reading recorded logs and replay the API call.
   * Even better if multiple API calls can be replayed in order.
* Validate replay correctness
   * Compare original vs replayed execution results (if applicable)
* Security and Privacy
  * Clearly document what data is recorded to avoid unintentional leaks, print a warning on the console when any data is stored.

bkryu:
Testing out implementation in #2206. Should be fully ready-to-merge in a week.

bkryu:
PR has been merged. Closing issue as complete

PR #2206, opened 2025-12-11
feat: Input/output Dump + Replay Mode for API Logging Level 10

bkryu:
## 📌 Description

**Flight Recorder: Level 10 API Logging with Tensor Dumps**
This PR provides a "Flight Recorder" mechanism within FlashInfer that captures detailed API execution traces (inputs, outputs, metadata) to enable exact reproduction of bugs, performance regression testing, and offline benchmarking.

The feature is implemented as Level 10 of the existing API logging module. The expected workflow is:
* Capture (Online):
  - User sets `FLASHINFER_LOGLEVEL=10`. Then makes a FlashInfer API call.
  - Before API execution: Inputs are serialized to `inputs.pt` (or `input.safetensors`) and metadata to `metadata.json`.
  - API execution: The actual kernel is invoked.
  - After API execution: Result is serialized to `outputs.pt`(or `input.safetensors`)  and completion status is appended to `metadata.jsonl`.
* Storage:
  - `FLASHINFER_DUMP_DIR` sets the dump location (default: `flashinfer_dumps`).
  - Each call gets a unique directory: `{timestamp}_pid{process_id}_{func_name}_call{sequence}`.
  - Tensor data is stored using `torch.save` (default) or optionally `safetensors`.
  - Metadata uses JSONL format (JSON Lines) for crash-safe, append-only logging.
  - A central `session.jsonl` provides a quick-scan overview of all recorded API calls.
 Non-contiguous tensors are needed to reproduce certain API calls like `mm_fp4()`.
* Replay (Offline):
  - User runs `flashinfer replay --dir <dump_path>`.
  - System loads inputs, dynamically resolves the function, executes it, and compares results.

**Environment Variables**

Variable | Default | Description
-- | -- | --
FLASHINFER_DUMP_DIR | flashinfer_dumps | Directory to save dump files
FLASHINFER_DUMP_MAX_SIZE_GB | 20 | Maximum total dump size in GB
FLASHINFER_DUMP_MAX_COUNT | 1000 | Maximum number of API calls to dump
FLASHINFER_DUMP_INCLUDE | (empty) | Comma-separated fnmatch patterns to include
FLASHINFER_DUMP_EXCLUDE | (empty) | Comma-separated fnmatch patterns to exclude
FLASHINFER_DUMP_SAFETENSORS | 0 | Set to 1 to use safetensors format (no pickle, but loses stride info)

**Dump Filtering**
For large-scale inference with many API calls, users can filter which APIs are dumped:
```
# Only dump decode-related APIs
export FLASHINFER_DUMP_INCLUDE="*decode*"

# Dump everything except __init__ and plan methods
export FLASHINFER_DUMP_EXCLUDE="*.__init__,*.plan"

# Only dump run() methods from wrapper classes
export FLASHINFER_DUMP_INCLUDE="*Wrapper.run"
```

Filter logic:
1. If `DUMP_INCLUDE` is set, only APIs matching at least one pattern are dumped
2. If `DUMP_EXCLUDE` is set, APIs matching any pattern are skipped
3. Both can be combined (include filter applied first, then exclude)

**Implementation Details**
- TTensor Serialization: Uses `torch.save` by default instead of raw bytes or simplified custom formats. FlashInfer kernels often rely on specific memory layouts (strides, non-contiguous views). `torch.save` preserves these properties faithfully, whereas reconstructing them from shape/stride metadata is error-prone and complex. An optional `safetensors` format is available for pickle-free serialization, though it does not preserve tensor strides.
- Multi-Process Support:
  - Dump directory names include the process ID (pid{PID}) for identification in multi-GPU environments.
  - The replay system uses a composite registry key (process_id, self_id) to prevent object collisions when replaying dumps from multiple processes.
- Stateful Replay: To support stateful APIs (e.g., BatchDecodeWithPagedKVCacheWrapper), the replay system tracks object identity across __init__, plan(), and run() calls. The object_registry maps (process_id, self_id) tuples to live object instances during replay.

## 🔍 Related Issues

#2170 

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * Level 10 "Flight Recorder": full input/output dumps with replayable session format and a CLI replay command reporting per-item outcomes.

* **Configuration**
  * Environment-configurable dump settings (dir, size/count limits), include/exclude filtering patterns, process-id substitution, and optional SafeTensors output.

* **Documentation**
  * Expanded docs covering dump layout, JSONL/session format, filtering examples, manual replay workflows, and replay examples.

* **Tests**
  * Comprehensive tests validating Level 10 logging, safetensors, filtering, sequencing, and CLI replay.

### DIFF
--- a/flashinfer/api_logging.py
+++ b/flashinfer/api_logging.py
@@ -1437,8 +1437,7 @@ def flashinfer_api(func: Callable = None) -> Callable:
         - 1: Log function name only (logged BEFORE execution - crash-safe)
         - 3: Log function name + inputs/outputs with metadata (inputs logged BEFORE execution - crash-safe)
         - 5: Log function name + inputs/outputs with metadata + tensor statistics (inputs logged BEFORE execution - crash-safe)
-        - 8: Level 5 logging + metadata-only dump (no tensor data saved)
-        - 10: Level 8 logging + dump tensors to disk for reproducibility (preserves stride/contiguity)
+        - 10: Level 5 logging + dump metadata and input/output tensors to disk for reproducibility (preserves stride/contiguity)
 
     FLASHINFER_LOGDEST : str (default: "stdout")
         - "stdout": Log to standard output
@@ -1455,6 +1454,16 @@ def flashinfer_api(func: Callable = None) -> Callable:
 
     FLASHINFER_DUMP_MAX_COUNT : int (default: 1000)
         - Maximum number of function call dumps
+
+    FLASHINFER_DUMP_SAFETENSORS : int (default: 0)
+        - 0: Use torch.save format (preserves stride/contiguity)
+        - 1: Use safetensors format (no pickle, but loses stride info)
+
+    FLASHINFER_DUMP_INCLUDE : str (default: "")
+        - Comma-separated list of patterns to include for dumping (fnmatch-style)
+
+    FLASHINFER_DUMP_EXCLUDE : str (default: "")
+        - Comma-separated list of patterns to exclude for dumping (fnmatch-style)
 
     Examples
     --------


flashinfer/api_logging.py
```diff
@@ -82,11 +104,830 @@ def _setup_logger():
...
+    kwargs: dict,
+    self_id: Optional[int] = None,
+) -> Optional[str]:
+    """
+    Dump function inputs to disk BEFORE execution (crash-safe).
+
+    This function:
+    1. Extracts tensors and metadata from inputs
+    2. Creates a timestamped directory
+    3. Saves inputs.pt and partial metadata.json
```
yzh119, line 320: can we consider safetensor and jsonl (and align with flashinfer-bench's trace collection format)?

cc @yyihuang (seems your #2305 have lots of overlap with this PR), can you coordinate and reach a consensus?

docs/logging.rst
```diff
@@ -116,3 +218,327 @@ Level 0 has zero overhead
...
+        else:
+            diff = (res - expected).abs().max().item()
+            print(f"\n❌ Output mismatch! Max diff: {diff}")
+
+Example 2: ``BatchDecodeWithPagedKVCacheWrapper`` - Stateful Wrapper Class
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+**Producer Script** (``batch_decode_producer.py``):
+
+This script demonstrates logging with a stateful wrapper class that requires ``__init__``, ``plan``, and ``run`` calls.
```
wenscarl, line 468: What if in the dump folder there are multiple __init__, plan, and runs. Is there a way to pick any combination of them in the replay mode? e.g. `--init init_0 --plan plan_3 --run run_2`.

flashinfer/api_logging.py
bkryu: This is great feedback. In the latest commit, I switched:
- from `metadata.json` to `metadata.jsonl`.
- from only `[intput.output.pt]` to giving an option to dump as safetensors via `FLASHINFER_DUMP_SAFETENSORS`. As we discussed, dumping via torch.pt retains all non-contiguity and custom strides while safetensors loses them, so it's likely best to keep both.

docs/logging.rst
bkryu: When multiple wrapper instances are created with interleaved `plan()`/`run()` calls, the system correctly tracks them:
1. Each dump's `metadata.jsonl` contains self_id (Python's id(self)) and `process_id`
2. During `replay_sequence()`, the `object_registry` uses composite keys `(process_id, self_id)` to maintain separate instances
    - This composite keys `(process_id, self_id)` was been added in the past few days, after we tried out this PR last month to debug the MLA IMA issue and struggled with confusion 😓 
4. Chronological replay automatically routes each plan()/run() to the correct instanceFor example, with two interleaved instances:

```
__init__ (self_id=12345) → registry[(pid, 12345)] = new instance A
__init__ (self_id=67890) → registry[(pid, 67890)] = new instance B
plan() (pid, self_id=12345)   → calls A.plan()
plan() (pid, self_id=67890)   → calls B.plan()
run() (pid, self_id=67890)    → calls B.run()
run() (pid, self_id=12345)    → calls A.run()
```

Regarding mixing the function calls from different wrapper instances `(--init init_0 --plan plan_3 --run run_2)`, the current implementation does not support it because it would not be a valid replay. One could potentially change the `(pid, self_id)` in the metadata files and force a replay, but combining calls from different instances would likely fail or produce incorrect results.

wenscarl: The composite key is good enough! LGTM.

flashinfer/api_logging.py
yyihuang: We can generate a separate jsonl file with another workflow in flashinfer-bench.

tests/utils/test_logging_replay.py
```diff
@@ -0,0 +1,1067 @@
...
+    )
+
+    # Verify tensors were loaded
+    assert len(result["args"]) == 2
+    assert result["args"][0].shape == (32, 128)
+    assert result["args"][1].shape == (128, 8, 128)
+
+
+if __name__ == "__main__":
+    pytest.main([__file__, "-v", "-s"])
```
yyihuang, line 1067: Will add another test with positional arguments only, for flashinfer-bench workflow.

bkryu:
@yzh119 , the PR seems ready. The original intent was to enable the dumps for debugging, but @yyihuang has expressed interest in the feature for flashinfer-bench, and will come up with a followup PR making further changes for flashinfer-bench.

Can I ask you to review once more?

Issue #2187, opened 2025-12-08
[Feature Request] Use rotating buffers for in benchmark functions


bkryu:
[flashinfer.testing.bench_gpu_time_with_cupti](https://github.com/flashinfer-ai/flashinfer/blob/main/flashinfer/testing/utils.py#L646C5-L646C30) today supports flushing the L2 cache before each run, but [bench_gpu_time_with_cuda_event](https://github.com/flashinfer-ai/flashinfer/blob/main/flashinfer/testing/utils.py#L551) does not.

This means that only in environments where `cupti-python` is installed we are able to microbenchmark performance with a cold L2, which is not widespread today (requires CUDA 13). 

It would be helpful to implement a rotating buffer using `input[round % N]` and `output[round % N]` where `N` could be set according to L2 cache size to use a cold L2.

sricketts:
@bkryu is this done as of #2213?

bkryu:
Yes this request was completed in  #2213 . I can close the issue.

PR #2213, opened 2025-12-12
feat: Cold L2 Cache Benchmarking with Rotating Buffers

bkryu:
### Problem

`bench_gpu_time_with_cudagraph` captures multiple kernel iterations within a graph to amortize timing overhead. However, the L2 cache becomes hot after the first iteration, leading to misleadingly better performance for memory-bound kernels.

### Solution

Introduce a unified `cold_l2_cache` parameter that automatically selects the appropriate strategy:

| Mode | Strategy | Mechanism |
|------|----------|-----------|
| CUPTI | L2 Flush | Flush cache between iterations |
| CUDA Events (no graph) | L2 Flush | Flush cache between iterations |
| CUDA Events + Graph | Rotating Buffers | Clone tensors, cycle through different memory regions |

### API Changes

# New unified interface
```
bench_gpu_time(
    fn=my_kernel,
    input_args=(x, y, out),    # NEW: explicit arguments for rotation
    cold_l2_cache=True,        # NEW: default True, auto-selects strategy
    use_cuda_graph=True,
)
```
**Deprecated:** `l2_flush`, `l2_flush_size_mb`, `l2_flush_device` (replaced by `cold_l2_cache`)

### Files Changed

- `flashinfer/testing/utils.py`: Core benchmarking utilities with rotating buffer support
- `benchmarks/routines/{attention,gemm,moe}.py`: Updated to use new API
- Deprecated: `l2_flush`, `l2_flush_size_mb`, `l2_flush_device` (replaced by cold_l2_cache)

## 🔍 Related Issues

#2187 

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * Added rotating-buffer support and a cold-L2 benchmarking mode to better simulate cold cache GPU timing.

* **Chores**
  * Expanded benchmark/test harness APIs to accept and propagate explicit per-run inputs through autotune and timing workflows.
  * Unified GPU timing utilities to accept input args/kwargs across CUPTI, CUDA-graph, and event paths, including deprecation warnings for legacy L2-flush options.

### DIFF
--- a/flashinfer/testing/utils.py
+++ b/flashinfer/testing/utils.py
@@ -765,9 +765,7 @@ def bench_gpu_time_with_cuda_event(
     """
     Benchmark kernel execution time using CUDA events (no CUDA graphs).
 
-    This is the simplest benchmarking method. It measures the total time from kernel
-    launch to completion, which includes both CPU-side kernel launch latency and
-    actual GPU kernel execution time. Best suited for kernels where launch overhead
+    This is the simplest benchmarking method. Best suited for kernels where launch overhead
     is negligible compared to execution time.
 
     The function performs:
@@ -1245,9 +1243,6 @@ def bench_gpu_time_with_cudagraph(
     through during graph capture, ensuring each kernel invocation operates on
     different memory regions. The number of buffer copies is automatically
     calculated based on the device's L2 cache size.
-
-    Note: For CUDA graphs, L2 flush alone is insufficient because the flush only
-    occurs between graph replays, not between kernel iterations within the graph.
 
     Args:
         fn (Callable): The kernel function to benchmark.


flashinfer/testing/utils.py
```diff
@@ -1008,17 +1409,86 @@ def bench_gpu_time(
...
+        l2_flush (bool): If True, flush L2 cache before each iteration (default: True).
+        l2_flush_size_mb (int): Size of buffer for L2 flush in MB (default: 256).
+        l2_flush_device (str): Device for the flush buffer (default: "cuda").
+        sleep_after_run (bool): If True, sleep briefly after iterations (default: False).
+        enable_cupti (bool): If True, use CUPTI for timing (default: False).
+        use_cuda_graph (bool): If True, use CUDA graph timing (default: False).
+        num_iters_within_graph (int): Kernel calls per graph (CUDA graph mode only,
+            default: 10).
+        rotate_buffers (bool): If True, use rotating buffers for cold-L2 benchmarking
+            (CUDA graph mode only, default: False).
```
yzh119: Why do we have this constraint? From my understanding rotate_buffer should also work without CUDAGraph?

bkryu: There is no technical constraint, but I did not implement it because when running outside of CUDA graph, we can simply flush L2 between runs with a `buffer.zero_()` ([here](https://github.com/bkryu/flashinfer/blob/6194eacbe4f94aa6f6751d0f6e8460c3b3582e1f/flashinfer/testing/utils.py#L861-L870))
```
    torch.cuda.synchronize()
    for iter_idx in range(repeat_iters):
        if l2_flush:
            buffer.zero_()
        start_events[iter_idx].record()
        call_fn()
        end_events[iter_idx].record()
```

It is the CUDA graph + CUDA events measurements that primarily requires rotating buffers because we launch the kernel [num_iters_within_graph times within the CUDA graph ](https://github.com/bkryu/flashinfer/blob/6194eacbe4f94aa6f6751d0f6e8460c3b3582e1f/flashinfer/testing/utils.py#L1334-L1345) and cannot flush the cache in between.
```
    g = torch.cuda.CUDAGraph()
    with torch.cuda.graph(g):
        if rotate_buffers and num_rotations > 1:
            # Capture with rotating buffers: use buffer[iter % num_rotations]
            for iter_idx in range(num_iters_within_graph):
                buf_idx = iter_idx % num_rotations
                call_fn_with_rotation(buf_idx)
        else:
            # Non-rotating capture (uses original args if provided)
            for _ in range(num_iters_within_graph):
                call_fn()
    torch.cuda.synchronize()
...
...
    torch.cuda.synchronize()
    for iter_idx in range(repeat_iters):
        if l2_flush:
            buffer.zero_()
        start_events[iter_idx].record()
        g.replay()
        end_events[iter_idx].record()
```

However, I can see the untidiness of users needing to think about L2 flush vs. rotating buffers.

Do you have any thoughts that come to mind whether or how we might want to clean this up?

bkryu: @yzh119, I have revised the confusing L2 flush vs. rotating buffer interface  in this PR by introducing a unifying `cold_l2_cache` variable that internally decides whether to flush L2 or use a rotating buffer so that the user does not need to worry.

Can you give a another look at the description and changes?

flashinfer/testing/utils.py
```diff
@@ -30,6 +30,206 @@
...
+    """
+    l2_size = get_l2_cache_size(device)
+    total_bytes = _calculate_tensor_bytes(tensors)
+
+    if total_bytes == 0:
+        return 1  # No tensors to rotate
+
+    # Use aggressive threshold: only skip rotation if tensors far exceed L2 (5x)
+    # This ensures cache effects are truly negligible even with prefetching
+    safe_cache_threshold = l2_size * 5
```
nvmbreughe, line 125: What's the origin of this number? Was it empirically decided?

flashinfer/testing/utils.py
```diff
@@ -770,12 +1095,22 @@ def func_buffer_completed(
...
+    def call_fn():
+        if has_args:
+            fn(*input_args, **input_kwargs)
+        else:
+            fn()
+
+    buffer = None
+    if _do_l2_flush:
+        l2_flush_size = int(_l2_flush_size_mb) * 1024 * 1024
+        buffer = torch.empty(l2_flush_size, device=_l2_flush_device, dtype=torch.int8)
```
nvmbreughe, line 1110: How do we guarantee that this writes to the L2 cache, and doesn't bypass it when writing to dram?

bkryu: The lines like [buffer.zero_()](https://github.com/flashinfer-ai/flashinfer/pull/2213/changes#diff-64724513908634bf5770f1d25a17409a55b9653bb1e3cdad9d87631f437886d8R884) is the part where the L2 cache is flushed.

1. We are doing a global store here on a tensor that matches the size of the devices l2 cache. Besides the fact that we know generally DRAM access is always through L2 (not familiar with exceptions) and that we have seen this approach to work, there is no guarantee.
2. Annectodally, Triton in [triton.testing](https://github.com/triton-lang/triton/blob/main/python/triton/testing.py#L182) uses the same approach. It calls a clear_cache function where for NVIDIA GPU runtimes, uses [this implementation](https://github.com/triton-lang/triton/blob/main/third_party/nvidia/backend/driver.py#L754-L764) that does `cache.zero_()`, and `triton.testing.do_bench()` is a fairly standard function used by many, and we at least match the methodology.

flashinfer/testing/utils.py
bkryu: Yes it was empirically set. I found that 1 or 2 is often not sufficient enough to clear out the cache, so I opted for a high enough number that is not too big. In a handful of cases, 5 or higher seems to have no measurable difference in performance while 1 or 2 was seeing "kernel runs too quickly to be true"

nv-yunzheq:
LGTM

aleozlx:
looks cool no comments on my end

PR #2217, opened 2025-12-14
feat: Support unpadded output hidden size for trtllm_fp4_block_scale_moe

elvischenv:
## 📌 Description

Support unpadded output hidden size for `trtllm_fp4_block_scale_moe` by passing the `output` tensor with unpadded hidden size, so that we don't need to do slicing after MoE.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected output dimension handling in MoE operations to use actual output tensor shapes rather than assumed values.
  * Enhanced validation for pre-allocated output tensors with stricter shape and dtype checks.
  * Improved consistency between different code execution paths for MoE operations.

### DIFF
--- a/csrc/nv_internal/cpp/common/envUtils.cpp
+++ b/csrc/nv_internal/cpp/common/envUtils.cpp
@@ -222,11 +222,6 @@ bool getEnvDisaggLayerwise() {
   return disaggLayerwise;
 }
 
-bool getEnvParallelCacheSend() {
-  static bool const parallelCacheSend = getBoolEnv("TRTLLM_PARALLEL_CACHE_SEND");
-  return parallelCacheSend;
-}
-
 bool getEnvRequestKVCacheConcurrent() {
   static bool const requestKVCacheConcurrent = getBoolEnv("TRTLLM_REQUEST_KV_CACHE_CONCURRENT");
   return requestKVCacheConcurrent;
@@ -277,7 +272,7 @@ size_t getEnvAllReduceWorkspaceSize() {
   return workspaceSize;
 }
 
-std::string getEnvKVCacheTransferOutputPath() {
+std::string const& getEnvKVCacheTimeOutputPath() {
   static std::string outputPath = getStrEnv("TRTLLM_KVCACHE_TIME_OUTPUT_PATH").value_or("");
   return outputPath;
 }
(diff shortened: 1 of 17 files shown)


flashinfer/fused_moe/core.py
```diff
@@ -1899,7 +1909,7 @@ def _fake_trtllm_fp4_block_scale_moe(
         tune_max_num_tokens: int,
     ):
         seq_len = hidden_states.shape[0]
-        hidden_size = hidden_states.shape[1]
+        hidden_size = hidden_states.shape[1] if output is None else output.shape[1]
```
yzh119, line 1912: Hi @elvischenv just want to make sure my understanding is right, does this mean hidden_states.shape[1] is the effective hidden dimension where `output.shape[1]` could be the padded hidden dimension?

elvischenv: I don't think so. `hidden_states.shape[1]` is already padded, the input should be padded before the MoE kernel. Before this PR, the API does not have info of the original unpadded hidden dim, the `output` will be in the same shape with `hidden_states`. We have to slice the output to original unpadded hidden size, which involves extra overhead.

This PR allows us to pass `output` with unpadded hidden dim, then the kernel will write the results in such dim by setting `args->hidden_size_output = output.size(1);`. Then we won't need the slice anymore.

flashinfer/fused_moe/core.py
```diff
@@ -1753,6 +1753,16 @@ def trtllm_fp4_block_scale_moe_op(
...
+        else:
+            check_shape_dtype_device(
+                output, None, torch.bfloat16, hidden_states.device, "output"
+            )
+            assert output.shape[0] == num_tokens, (
+                f"output.shape[0]={output.shape[0]} must be equal to {num_tokens}"
+            )
+            assert output.shape[1] <= hidden_size, (
+                f"output.shape[1]={output.shape[1]} must be less than or equal to {hidden_size}"
+            )
```
elvischenv, line 1765: We do this assertion since the `output` could be unpadded so that smaller than the (padded) hidden size.

flashinfer/fused_moe/core.py
yzh119: Thank you for the explanation!

PR #2233, opened 2025-12-17
feat: Fused RMSNorm + FP4 Quantization Kernels in CuTe-DSL

bkryu:
## Fused RMSNorm + FP4 Quantization Kernels (CuTe-DSL)

This PR adds two high-performance fused kernels implemented in CuTe-DSL for Blackwell (SM100+) GPUs:

- **`rmsnorm_fp4quant`**: Fuses RMSNorm and FP4 quantization into a single kernel
- **`add_rmsnorm_fp4quant`**: Fuses residual add, RMSNorm, and FP4 quantization into a single kernel

### Features
- Supports both **NVFP4** (block_size=16, E4M3 scales) and **MXFP4** (block_size=32, UE8M0 scales) formats
- FP16 and BF16 input support
- Hidden sizes from 64 to 32K+

### Files Added
- `flashinfer/cute_dsl/rmsnorm_fp4quant.py` - RMSNorm + FP4 quant kernel
- `flashinfer/cute_dsl/add_rmsnorm_fp4quant.py` - Add + RMSNorm + FP4 quant kernel
- `flashinfer/cute_dsl/__init__.py` - Module exports
- `tests/norm/test_rmsnorm_fp4_quant_cute_dsl.py` - Unit tests
- `tests/norm/test_add_rmsnorm_fp4_quant_cute_dsl.py` - Unit tests
- `benchmarks/bench_cute_dsl_rmsnorm_fp4quant.py` - Benchmark script
- `benchmarks/bench_cute_dsl_add_rmsnorm_fp4quant.py` - Benchmark script

### Benchmark outputs
<details>
<summary>B200 (SM100)</summary>

```
$ python3 benchmarks/bench_cute_dsl_rmsnorm_fp4quant.py 
================================================================================
Fused RMSNorm + FP4 Quantization Benchmark
================================================================================
GPU Compute Capability: SM100

Running sanity check...
  OK: (128, 256) - FP4 match
  OK: (512, 1024) - FP4 match
  OK: (1024, 2048) - FP4 match
✓ Confirmed: CuTe-DSL output is equivalent to RMSNorm + fp4_quantization


Batch    Hidden   CuTe-DSL (µs)  BW (GB/s)  RMSNorm (µs)  FP4Q (µs)   Separate (µs)  vs Separate 
-------------------------------------------------------------------------------------------------
1000     1536     4.4            898.6      3.8           3.8         7.6            1.74x       
1000     2048     5.2            1019.4     4.2           4.2         8.4            1.62x       
[...]

PR #2243, opened 2025-12-19
feat: RMSNorm/Fused RMSNorm + FP8 Quantization kernels

BLaZeKiLL:
## 📌 Description

FP8 model inference requires multiple intermediate quantization kernels, which can be avoided by fusing norm and quantization kernels. Consumers like sglang and vllm can lower to these norm + quant fusion kernels using custom torch compile passes

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### Reference
I have been working on adding custom fusion passes to sglang as part of the following [RFC](https://github.com/sgl-project/sglang/issues/10118) and would like to use flashinfer's norm kernels for the norm quant fusions instead of migrating vllm kernels to sglang as part of the following [MR](https://github.com/sgl-project/sglang/pull/10549)

### Implementation
I realise that existing kernels (at least for rmsnorm) can be modified to add the scale parameter as an optional parameter, thereby avoiding most code duplication. However, as an initial implementation, I have opted for a separate implementation route. This can be refactored if required.

For fused_add_rmsnorm_quant, I don't think an in-place update would be possible since dtypes for input and output differ

Currently, FP8_E3M4 numeric limits (448) have been hard-coded, as I am not aware of getting this value at compile time without including c10 headers from torch, and not sure if that is acceptable post tvm ffi migration

Following is a snippet from VLLM, and I have seen similar code for getting the FP8 numeric limits
```cpp
#include <c10/util/Float8_e4m3fn.h>

template <typename T,
          typename = std::enable_if_t<std::is_same_v<T, c10::Float8_e4m3fn> ||
                                      std::is_same_v<T, c10::Float8_e4m3fnuz> ||
                                      std::is_same_v<T, int8_t>>>
struct quant_type_max {
  static constexpr T val() { return std::numeric_limits<T>::max(); }
};
```

The best option in my mind is to introduce `include/flashinfer/fp8.h` containing something similar to the above snippet, and also support e5m2

### Tests
atol and rtol for the fp8 assertions had to be high due to the low precision nature of the data, but with tolerances of 1e-2, just a few tests fail with a single element mismatch

## Summary by CodeRabbit

* **New Features**
  * Added quantized RMSNorm and fused quantized RMSNorm (residual-add) with configurable scale, eps, and PDL toggle.
  * Supports FP16/FP8 paths and optional per-token or per-tensor scaling; outputs are clamped for quantized formats.

* **Tests**
  * Added tests validating quantized normalization and fused-residual flows across dtypes, batch sizes, scaling modes, and PDL configurations.

### DIFF
--- a/csrc/flashinfer_norm_binding.cu
+++ b/csrc/flashinfer_norm_binding.cu
@@ -17,8 +17,14 @@
 
 void rmsnorm(TensorView out, TensorView input, TensorView weight, double eps, bool enable_pdl);
 
+void rmsnorm_quant(TensorView out, TensorView input, TensorView weight, double scale, double eps,
+                   bool enable_pdl);
+
 void fused_add_rmsnorm(TensorView input, TensorView residual, TensorView weight, double eps,
                        bool enable_pdl);
+
+void fused_add_rmsnorm_quant(TensorView output, TensorView input, TensorView residual,
+                             TensorView weight, double scale, double eps, bool enable_pdl);
 
 void gemma_rmsnorm(TensorView out, TensorView input, TensorView weight, double eps,
                    bool enable_pdl);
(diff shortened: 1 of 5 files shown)


yzh119 (approved):
I'm good with the PR and implementation in general.

Please also note that @bkryu is working on another PR #2233 using cute-dsl and we might shift to cute-dsl for development of these kernels in the future.

bkryu:
> I'm good with the PR and implementation in general.
> 
> Please also note that @bkryu is working on another PR #2233 using cute-dsl and we might shift to cute-dsl for development of these kernels in the future.

At the least, I can see if I can benchmark the code and see if there is a need to/can write a cute-dsl based implementation

PR #2308, opened 2026-01-08
Fix: FilteredTopKUnifiedKernel read value out of length

HarryWu99:
## 📌 Description

I found an issue in the implementation of `top_k_ragged_transform` where elements beyond the specified `lengths` range can be read. This leads to out-of-bounds indices in the result.

Minimal repro:

```python
torch.set_default_device(torch.cuda.current_device())
torch.manual_seed(22)
topk = 4
block_logits = torch.randn(1, 8, dtype=torch.float32)
print(block_logits)

idxs = flashinfer.top_k_ragged_transform(
    block_logits,
    offsets=torch.zeros(1, dtype=torch.int32),
    lengths=torch.tensor([5], dtype=torch.int32),
    k=topk
)
print(idxs)
print(block_logits[:, :5].topk(k=4).indices)
```

Output:

```text
tensor([[ 0.4519,  1.0099, -0.3167, -2.1224,  1.0826,  1.9583,  0.2751,  0.0463]],
       device='cuda:0')
tensor([[0, 4, 1, 5]], device='cuda:0', dtype=torch.int32)
tensor([[4, 1, 0, 2]], device='cuda:0')
```

In this example, `lengths = [5]`, so indices in `idxs` should be strictly less than 5. The returned index `5` is clearly out of bounds.

The root cause is that the `vec_load` logic does not properly guard against reading past the valid range near the boundary. After fixing the boundary handling in `vec_load`, this case produces the expected result.

Affected APIs:
- `top_k_ragged_transform`
- `top_k_page_table_transform`

I also checked the implementation of `RadixTopKMultiCTA`, and it does not appear to suffer from this bug.

## 🔍 Related Issues

None

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

```
>> python tests/utils/test_topk.py
Testing page table transform...
Testing ragged transform...
Testing trivial cases...
Testing variable lengths...
Testing large scale...

Testing SGLang-style comparisons...

All tests passed!
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved boundary-safe processing for vectorized and tail elements to prevent out-of-range reads and ensure consistent histogram/top-k results on non-multiple lengths.
* **Tests**
  * Added a test verifying top-k behavior with ragged/out-of-length inputs, ensuring outputs remain within valid index ranges and match a reference implementation.

### DIFF
--- a/include/flashinfer/topk.cuh
+++ b/include/flashinfer/topk.cuh
@@ -1994,7 +1994,7 @@ __global__ void __launch_bounds__(FILTERED_TOPK_BLOCK_THREADS)
   for (int base = tx * VEC_SIZE; base < aligned_length; base += BLOCK_SIZE * VEC_SIZE) {
     score_vec.cast_load(&score[base]);
 #pragma unroll
-    for (int j = 0; j < VEC_SIZE && base + j < length; ++j) {
+    for (int j = 0; j < VEC_SIZE; ++j) {
       const auto bin = Traits::ToCoarseKey(score_vec[j]);
       atomicAdd(&s_histogram[bin], 1);
     }
@@ -2043,7 +2043,7 @@ __global__ void __launch_bounds__(FILTERED_TOPK_BLOCK_THREADS)
     for (int base = tx * VEC_SIZE; base < aligned_length; base += BLOCK_SIZE * VEC_SIZE) {
       score_vec.cast_load(&score[base]);
 #pragma unroll
-      for (int j = 0; j < VEC_SIZE && base + j < length; ++j) {
+      for (int j = 0; j < VEC_SIZE; ++j) {
         const auto bin = static_cast<int>(Traits::ToCoarseKey(score_vec[j]));
         if (bin > threshold_bin) {
           const auto pos = atomicAdd(&s_counter, 1);


yzh119:
@claude  Can you create regression tests that demonstrate the bug fix / feature in this PR? The tests should fail on the old codebase and pass on the new one.

HarryWu99:
I updated the logic for detecting the out of length condition following the guidance from Gemini, and added a corresponding unit test. The new test fails on the previous implementation and passes after this fix.

I also ran the performance benchmark benchmarks/bench_topk.py, and did not observe any noticeable performance differences compared to the previous version.

after fix
```
====================================================================================================
top_k_ragged_transform: Fused top-k + ragged index transform (dtype=FP32)
====================================================================================================
 batch    seq_len      k |   FlashInfer
----------------------------------------------------------------------
     1      32768   2048 |      23.95us
     4      32768   2048 |      24.45us
     8      32768   2048 |      24.37us
    16      32768   2048 |      25.54us
```
before fix
```
====================================================================================================
top_k_ragged_transform: Fused top-k + ragged index transform (dtype=FP32)
====================================================================================================
 batch    seq_len      k |   FlashInfer
----------------------------------------------------------------------
     1      32768   2048 |      24.26us
     4      32768   2048 |      23.60us
     8      32768   2048 |      24.58us
    16      32768   2048 |      24.85us
```

PR #2341, opened 2026-01-13
chore: MoE benchmark effective BW fix for trtllm_block_scale_moe

rosenrodt:
## 📌 Description

The MoE benchmark script overestimates the num bytes loaded by assuming all experts are active. I saw effective BW exceeds 3x the peak BW of some system as a result. The fix is to calculate the routed experts (topk_ids) on the host side and count the unique number of experts, the same logic `cutlass_fused_moe` does.

While investigating the above issue, I also found data init of routing_bias using `rand()` results in very skewed expert distribution (repro cmd below gives 18 active out of 128 experts). I'd like to change it to `ones()*0.1` for smoother expert distribution (noe giving 114 out of 128), while maintaining the same load/compute behavior in the kernels.

```
python3 flashinfer_benchmark.py --routine trtllm_fp4_block_scale_moe --num_tokens 32 --hidden_size 7168 --intermediate_size 2048 --num_experts 128 --routing_method deepseek_v3 --top_k 8 --n_group 8 --topk_group 4 --routed_scaling_factor 2.5 --use_routing_bias --use_shuffled_weight --generate_repro_command -vv
```

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * Added support for nvfp4 and mxfp4 quantization formats in bandwidth calculations.
  * Introduced routing support for DeepSeekV3 method.

* **Improvements**
  * Enhanced routing bias initialization for more consistent expert distribution.
  * Expanded routing computation utilities for greater flexibility.

* **Tests**
  * Updated benchmark test data to align with new routing and quantization logic.

### DIFF
--- a/benchmarks/routines/moe.py
+++ b/benchmarks/routines/moe.py
@@ -1233,6 +1233,7 @@ def testCutlassFusedMoe(args):
         weight_format=variant,
         routing_logits_dtype=router_logits.dtype,
         active_experts=int(selected_experts.unique().numel()),
+        verbose=args.verbose,
     )
 
     print_perf_metrics(backend, median_time, std_time, tflops, tb_per_sec)
@@ -1502,6 +1503,7 @@ def testTrtllmFp8BlockScaleMoe(args):
         weight_format="fp8",
         routing_logits_dtype=routing_logits.dtype,
         active_experts=int(selected_experts.unique().numel()),
+        verbose=args.verbose,
     )
 
     backend = "trtllm"
@@ -1733,6 +1735,7 @@ def testTrtllmFp8PerTensorScaleMoe(args):
         weight_format="fp8",
         routing_logits_dtype=routing_logits.dtype,
         active_experts=int(selected_experts.unique().numel()),
+        verbose=args.verbose,
     )
 
     backend = "trtllm"


benchmarks/routines/moe.py
```diff
@@ -430,21 +435,24 @@ def calculate_moe_bandwidth(
...
     # Get effective byte sizes
     def get_effective_bytes(dtype: torch.dtype, fmt: Optional[str]) -> float:
-        if fmt == "fp4":
-            return 0.5
-        if fmt == "fp8":
+        if fmt == "nvfp4":
+            return 0.5 + 1 / 16
+        elif fmt == "mxfp4":
+            return 0.5 + 1 / 32
+        elif fmt == "fp8":
```
nv-yunzheq, line 455: Is the weight for fp8 block-scaled? i.e. MXFP8?

nv-yunzheq:
LGTM.
For TRTLLM gen moe case, the bandwidth is still an estimate as the selected experts are not the exact same compute from the kernel itself. But it's better and closer to what we previously have.

benchmarks/routines/moe.py
rosenrodt: I have not considered block-scale fp8 (DeepSeek style) yet. In that case, it should be 1 fp32 scale every 128x128 block for weight, and 1 fp32 scale every 128x1 for activation.

rosenrodt:
> LGTM. For TRTLLM gen moe case, the bandwidth is still an estimate as the selected experts are not the exact same compute from the kernel itself. But it's better and closer to what we previously have.

Routing is computed internally in fairly high precision (bf16 or fp32). Therefore as long as the math is equivalent I think we're good. I will let @ChristinaZ comment if that is really the case for routing + topK in trtllm_block_scale_moe.

PR #2352, opened 2026-01-14
Added the cudnn backend Ragged KV Cache wrapper

Anerudhan:
## 📌 Description

Added the cudnn backend Ragged KV Cache wrapper
Fixed the test_prefill.py to not use torch.ones (accidentally did it before)

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

- [x ] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).

## 🧪 Tests

- [x ] All tests are passing (`unittest`, etc.).

## Summary by CodeRabbit

* **New Features**
  * Added explicit cuDNN-native backend support and enhanced batched prefill to handle variable-length sequences and indptr-based offsets for cuDNN paths.

* **Tests**
  * Expanded and hardened cuDNN tests: broader parameter sets, randomized initializations, larger workspace, and a wrapper-based test flow for more reliable validation.

* **Documentation**
  * Updated README and benchmark backend lists to include cuDNN-native and reflect backend support changes.

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -16,7 +16,7 @@ Currently supports testing attention, gemm, fused MOE, normalization, and quanti
     - `BatchPrefillWithPagedKVCacheWrapper` - Prefill attention with paged KV cache.
         - Also supports computationally similar `cudnn_batch_prefill_with_kv_cache` and  `trtllm_batch_context_with_kv_cache`.
     - `BatchPrefillWithRaggedKVCacheWrapper` - Prefill attention with ragged KV cache.
-        - Also supports computationally similar `cudnn_batch_prefill_with_kv_cache` and  `trtllm_ragged_attention_deepseek`.
+        - Also supports computationally similar `cudnn_batch_prefill_with_kv_cache` (cudnn-native) and  `trtllm_ragged_attention_deepseek`.
     - `BatchMLAPagedAttentionWrapper` - MLA attention proposed in DeepSeek series of models.
         - Also supports computationally similar `trtllm_batch_decode_with_kv_cache_mla`.
 - GEMM:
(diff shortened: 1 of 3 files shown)


tests/attention/test_cudnn_prefill_deepseek.py
```diff
@@ -103,45 +104,45 @@ def test_cudnn_prefill_deepseek(
...
+        num_kv_heads=num_kv_heads,
+        head_dim_qk=head_dim_qk,
+        head_dim_vo=head_dim_vo,
+        causal=causal,
+        sm_scale=scale,
+        q_data_type=torch.bfloat16,
+        kv_data_type=torch.bfloat16,
+        o_data_type=torch.bfloat16,
+        seq_lens=actual_seq_lens_kv,
+        seq_lens_q=actual_seq_lens_q,
```
bkryu, line 126: Do we envision deprecating `cudnn_batch_prefill_with_kv_cache` entirely?

If not, I suggest we test both `cudnn_batch_prefill_with_kv_cache` and the wrapper. Maybe should be a test parameter `@pytest.mark.parametrize("use_wrapper", [True, False])` and then handling an if statement in the function body, or even writing a separate test.

One value of keeping both is it serves as a reference for "how do I call this API" because we know that whoever authored an API in FlashInfer would have written the test script with the correct usage.

flashinfer/prefill.py
bkryu: Document 'cudnn' here?

flashinfer/prefill.py
```diff
@@ -2598,6 +2598,12 @@ def plan(
         max_item_len_ptr: Optional[torch.Tensor] = None,
         fixed_split_size: Optional[int] = None,
         disable_split_kv: bool = False,
+        seq_lens: Optional[torch.Tensor] = None,
```
bkryu, line 2601: Should we call this `seq_lens_kv` for consistency?

Anerudhan: Keeping it `seq_lens` for consistency with the PagedKVCache variant

tests/attention/test_cudnn_prefill_deepseek.py
Anerudhan: Not deprecating `cudnn_batch_prefill_with_kv_cache`, but not using it as a API as well since we want to use a top level unified API.

flashinfer/prefill.py
Anerudhan: Done. Thanks

PR #2442, opened 2026-01-30
Fix autotuner oom

### LINKED ISSUE #2357: Autotuning failsafe fallback to top1 tactic
Sometimes auto-tuning trigger OOM and cause problem at frameworks level.
Based on some early discussion, this is an edge case that likely need some graceful degradation, thus we are thinking of doing a failsafe fallback to top1 tactic upon potential OOM so as to avoid causing functionality blocker by this issue.

zack041:
## 📌 Description
Add graceful OOM handling during autotuning. When `torch.cuda.OutOfMemoryError` occurs, the autotuner now clears CUDA cache and falls back to the default tactic `(runners[0], -1)` instead of crashing. The try-except block wraps the entire profiling loop, covering methods like `_prepare_input_tensors()` that could also cause OOM. OOM from the inner profiling loop is raised to be caught by the outer exception handler.

## 🔍 Related Issues
Fixes #2357 

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Reviewer Notes
No tests added because OOM during autotuning is difficult to reliably reproduce in a test environment.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved profiling error handling so individual tactic failures are caught, logged, recorded, and do not abort tuning.
  * Added robust out-of-memory handling that clears GPU resources and falls back to safe/previous configurations instead of crashing.
  * Ensured tuning continues after non‑OOM errors, preserves cache/metrics consistency, and still selects the best measured configuration when available.

### DIFF
--- a/flashinfer/autotuner.py
+++ b/flashinfer/autotuner.py
@@ -479,7 +479,10 @@ class AutoTuner:
                         # TODO: use FakeTensor here.
                         valid_tactics = r.get_valid_tactics(tensors, p)
                         runner_arg_names = runner_arg_names_map[r]
-                        if "do_preparation" in runner_arg_names and len(valid_tactics) > 0:
+                        if (
+                            "do_preparation" in runner_arg_names
+                            and len(valid_tactics) > 0
+                        ):
                             r(tensors, tactic=-1, do_preparation=True, **kwargs)
                         for tac in valid_tactics:
                             try:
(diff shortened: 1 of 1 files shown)


yzh119:
@flashinfer-bot  run

yzh119:
@flashinfer-bot run

PR #2248, opened 2025-12-19
bugfix: skip CUTLASS kernel generation when AOT cache exists

yongwww:
Fixes #2093 

prior to the pr, when `flashinfer-jit-cache` is installed, the `gen_cutlass_fused_moe_module()`  function was unconditionally attempting to create directories and generate kernel source files (`.generated.cu`). This caused "Permission denied" errors in Docker containers where `/usr/local/lib/python3.12/dist-packages/` is not writable at runtime.

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [ ] Tests have been added or updated as needed.
- [ ] All tests are passing (`unittest`, etc.).

## Release Notes

* **Chores**
  * Optimized kernel generation process for AOT (ahead-of-time) compilation. The build now checks for existing compiled artifacts and skips regeneration when available, potentially reducing build times.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

Issue #2093, opened 2025-11-14
[BUG] Flashinfer with jit-cache and cubin still attempts to jit


bbartels:
I encountered an issue with flashinfer attempting to create a directory in order to cache JIT output for `qwen3-coder-480b-a35b`. I would have assumed in a situation with `flashinfer-jit-cache` and `flashinfer-cubin` it would not additionally require jitting.
This was found using the vllm Dockerfile which includes both dependencies.
For details please refer to: https://github.com/vllm-project/vllm/issues/28440

```
2025-11-11T02:42:39.924015483Z
[1;36m(EngineCore_DP0 pid=523)[0;0m   File "/usr/local/lib/python3.12/dist-packages/flashinfer/fused_moe/core.py", line 818, in cutlass_fused_moe
2025-11-11T02:42:39.924015483Z
[1;36m(EngineCore_DP0 pid=523)[0;0m     return get_cutlass_fused_moe_module(device_arch).cutlass_fused_moe(
2025-11-11T02:42:39.924015483Z
[1;36m(EngineCore_DP0 pid=523)[0;0m            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2025-11-11T02:42:39.924015483Z
[1;36m(EngineCore_DP0 pid=523)[0;0m   File "/usr/local/lib/python3.12/dist-packages/flashinfer/fused_moe/core.py", line 286, in get_cutlass_fused_moe_module
2025-11-11T02:42:39.924015483Z
[1;36m(EngineCore_DP0 pid=523)[0;0m     module = gen_cutlass_fused_moe_sm90_module(use_fast_build).build_and_load()
2025-11-11T02:42:39.924015483Z
[1;36m(EngineCore_DP0 pid=523)[0;0m              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2025-11-11T02:42:39.924015483Z
[1;36m(EngineCore_DP0 pid=523)[0;0m   File "/usr/local/lib/python3.12/dist-packages/flashinfer/jit/fused_moe.py", line 77, in gen_cutlass_fused_moe_sm90_module
2025-11-11T02:42:39.924015483Z
[1;36m(EngineCore_DP0 pid=523)[0;0m     return gen_cutlass_fused_moe_module(nvcc_flags, "90", use_fast_build)
2025-11-11T02:42:39.924015483Z
[...]

PR #2445, opened 2026-01-30
bugfix: fix stub generation directory in fused_moe module

yzh119:
## 📌 Description

bugfix to #2093, the fundamental issue is we should not write to `jit_env.FLASHINFER_CSRC_DIR` (which might be read-only) for fused-moe module, instead we should use `FLASHINFER_GEN_SRC_DIR` which is supposed to be writable.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Bug Fixes**
  * JIT kernel generation now correctly uses user-writable cache directories instead of package directories, resolving compatibility issues in post-installation and read-only environments.

* **Documentation**
  * Updated JIT directory rules clarifying which directories are writable versus read-only. Added details on GPU auto-detection and CUDA architecture environment variable controls.

### DIFF
--- a/.claude/skills/add-cuda-kernel/SKILL.md
+++ b/.claude/skills/add-cuda-kernel/SKILL.md
@@ -234,6 +234,7 @@ def gen_scale_module(dtype_in, dtype_out):
 - No Jinja template needed for simple operations
 - Just copy source files to generation directory
 - URI uniquely identifies the module configuration
+- **NEVER write to package directories** - see "JIT Directory Rules" in `CLAUDE.md`
 
 ### (Optional) Specifying Supported CUDA Architectures
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -189,6 +189,17 @@ FlashInfer's JIT system has three layers:
 - `sources`: List of .cu/.cpp files to compile
 - `extra_cuda_cflags`, `extra_cflags`, `extra_ldflags`: Compiler flags
 
+### JIT Directory Rules
+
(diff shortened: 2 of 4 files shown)


yongwww (approved):

yongwww:
x-post from #2248: it would be good to also have this https://github.com/flashinfer-ai/flashinfer/pull/2248#discussion_r2646340901. Because writing files here isn’t expected (no matter where we write them).

yzh119:
> x-post from https://github.com/flashinfer-ai/flashinfer/pull/2248: it would be good to also have this https://github.com/flashinfer-ai/flashinfer/pull/2248#discussion_r2646340901. Because writing files here isn’t expected (no matter where we write them).

I have another branch working on that (defer file generation), I'm not saying your #2248 is wrong, but there are actually much more modules we need to fix if you want to unify the behavior of jit file generation, but you only change thw behavior of `gen_cutlass_fused_moe_module`.

yzh119:
More specifically:

* Almost all functions inside`jit/attention/modules.py` write stub files
* `jit/gemm/core.py`
* `jit/activation.py`
* `jit/fused_moe.py`: where your #2248 is trying to fix.

If we don't want to generate files at all, we should fix all of these modules's behavior.

yongwww:
Yeah, I agree PR #2248 didn’t cover all modules, and this PR is a simpler approach. I’ve closed #2248 in favor of this one.

PR #2455, opened 2026-01-31
Fix nan sampling

zack041:
## 📌 Description

Fix illegal memory access in `TopKSamplingFromProbKernel` and `TopPSamplingFromProbKernel` when input probabilities contain NaN values. 

Illegal memory access occurs for top k sampling and top p sampling (but not the other sampling functions) when a whole row is the filled with NaN. In this case, no valid token is found during sampling, meaning the last_valid_id remains as a garbage value. Thus when computing `pivot_0 = probs[row_idx * d + sampled_id]`, illegal memory access arises.

## 🔍 Related Issues

#2402 

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [x] Tests have been added or updated as needed.
- [x] All tests are passing (`unittest`, etc.).

## Reviewer Notes

I'm currently using 0 as the placeholder token, let me know if a different value is preferred. I will add the optional NaN counting feature in a follow up PR.

## Release Notes

* **Bug Fixes**
  * Enhanced autotuner robustness with improved error handling during profiling—automatically recovers from failures, clears memory as needed, and falls back to safe configurations.
  * Fixed sampling to correctly handle invalid or problematic input values and return safe defaults.

* **Tests**
  * Expanded test coverage to validate proper handling of edge cases in sampling operations.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #2456, opened 2026-01-31
fix: fix illegal memory access for NaN input in sampling kernels

### LINKED ISSUE #2402: bug: improve sampling kernels robustness when inputs include NaN
The sampling kernels in flashinfer do not properly handles NaN input values and may lead to illegal memory access in this case. Per discussion w/ @youkaichao and @WoosukKwon , the behavior of sampling kernel for NaN input should be:

* No crash and illegal memory access - Return undefined/placeholder token ids for affected requests
* No blocking - NaN in one request should not affect other requests in the batch
* Optional NaN counting - Allow users to verify output reliability

zack041:
### Summary

Fix illegal memory access when input probabilities contain NaN values. Added `valid` output tensor so callers can distinguish failed sampling from legitimately sampling token 0. Also added missing `@register_fake_op` for `min_p_sampling_from_probs` to support `torch.compile`.

#### API
- New `return_valid: bool = False` parameter for all sampling functions
- When `True`, returns `(samples, valid)` tuple

#### Kernel
- Added `bool* valid` output to sampling kernels
- Initialize `last_valid_id = -1` before sampling
- `valid[bx] = false` when no valid token found (NaN input)
- `valid[bx] = true` when valid

#### PyTorch Integration
- Add missing `@register_fake_op` for `min_p_sampling_from_probs`

#### Affected Functions
- `sampling_from_probs`
- `top_k_sampling_from_probs`
- `top_p_sampling_from_probs`
- `min_p_sampling_from_probs`
- `top_k_top_p_sampling_from_probs`

### Usage
```python
# Default
samples = flashinfer.sampling.top_k_sampling_from_probs(probs, top_k=50)

# With validity check
samples, valid = flashinfer.sampling.top_k_sampling_from_probs(probs, top_k=50, return_valid=True)
# valid[i] == False means NaN or invalid input for row i
```

Fixes #2402

## 🔍 Related Issues

#2402

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Reviewer Notes

I'm currently using 0 as the placeholder token, let me know if a different value is preferred. I will add the optional NaN counting feature in a follow up PR.

## Summary by CodeRabbit

* **New Features**
  * Sampling APIs can optionally return a per-sample boolean "valid" mask alongside outputs (use a new return flag).

* **Bug Fixes**
  * Per-iteration state reset and improved handling of invalid/NaN or out-of-range probability rows: such cases now produce a neutral zero sample and mark valid=false.

* **Tests**
  * Added tests verifying behavior with NaN/invalid probability inputs and the new valid-mask return.

### DIFF
--- a/csrc/sampling.cu
+++ b/csrc/sampling.cu
@@ -107,6 +107,9 @@ void sampling_from_probs(TensorView probs, TensorView output, TensorView valid,
                          Optional<TensorView> maybe_offset_arr, uint64_t offset_val) {
   CHECK_INPUT(probs);
   CHECK_DIM(2, probs);  // probs: (batch_size, vocab_size)
+  CHECK_INPUT(valid);
+  CHECK_DIM(1, valid);
+  CHECK_DEVICE(valid, probs);
   CHECK_MAYBE_INPUT_TYPES(maybe_indices, dl_int32, dl_int64);
   CHECK_MAYBE_SAME_DTYPE(maybe_indices, output);
   validate_seed_offset_tensors(maybe_seed_arr, maybe_offset_arr, probs);
@@ -144,6 +147,9 @@ void top_p_sampling_from_probs(TensorView probs, TensorView output, TensorView v
                                uint64_t offset_val) {
   CHECK_INPUT(probs);
   CHECK_DIM(2, probs);  // probs: (batch_size, vocab_size)
+  CHECK_INPUT(valid);
+  CHECK_DIM(1, valid);
+  CHECK_DEVICE(valid, probs);
   CHECK_MAYBE_INPUT_TYPES(maybe_indices, dl_int32, dl_int64);
   CHECK_MAYBE_SAME_DTYPE(maybe_indices, output);
   validate_seed_offset_tensors(maybe_seed_arr, maybe_offset_arr, probs);
@@ -187,6 +193,9 @@ void top_k_sampling_from_probs(TensorView probs, TensorView output, TensorView v
   CHECK_DEVICE(output, probs);
   CHECK_DIM(2, probs);   // probs: (batch_size, vocab_size)
   CHECK_DIM(1, output);  // output: (batch_size)
+  CHECK_INPUT(valid);
+  CHECK_DIM(1, valid);
+  CHECK_DEVICE(valid, probs);
   CHECK_MAYBE_INPUT_TYPES(maybe_indices, dl_int32, dl_int64);
   CHECK_MAYBE_SAME_DTYPE(maybe_indices, output);
   validate_seed_offset_tensors(maybe_seed_arr, maybe_offset_arr, probs);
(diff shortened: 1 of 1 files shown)


yzh119:
Hi @zack041 can you add unittest for NaN inputs, and we might also fix the behavior of `SamplingFromProbKernel`.

zack041:
> Hi @zack041 can you add unittest for NaN inputs, and we might also fix the behavior of `SamplingFromProbKernel`.

Hi @yzh119 , just looking for some clarification. I've added the unittest in the PR and the function is test_sampling_nan_input(), are you looking for some additional checks? 

I'll add the same fix to SamplingFromProbKernel and include a test for it. For the other sampling kernels, should I also fix them in this PR or leave them in a follow-up?

zack041:
Hi @yzh119 , I fixed `SamplingFromProbKernel` and I'm back with some updates. Several sampling methods (including SamplingFromProbKernel) did not cause illegal memory access but instead assigns garbage value to `.last_valid_id`. I had problems capturing this case using the previous approach, therefore I added a restraint `max_valid_index < (int)d` to protect the assigning of `.last_valid_id` in `DeviceSamplingFromProb` to solve. I'm not sure if the scope of this problem is going beyond the scope of this issue, what is our next steps?

yzh119:
Thanks for the fix @zack041, the direction looks right. A few things to address before we merge:

1. **`TopKTopPSamplingFromProbKernel` is not fixed.** The combined top-k + top-p kernel has the same pattern but wasn't touched — it will have the same crash on all-NaN input. Please add the same fix there.

2. **Test coverage for non-NaN rows.** The test only sets row 0 to NaN. Please also assert that the other rows still produce valid token indices (i.e., NaN in one request doesn't affect other requests in the batch).

3. **Set `success` output to `false` for NaN rows.** Currently NaN rows silently output index 0, which is indistinguishable from legitimately sampling token 0. The `success` output tensor should be set to `false` for these rows so callers can detect the failure.

Once these are addressed, let's get this merged.

zack041:
> Thanks for the fix @zack041, the direction looks right. A few things to address before we merge:
> 
> 1. **`TopKTopPSamplingFromProbKernel` is not fixed.** The combined top-k + top-p kernel has the same pattern but wasn't touched — it will have the same crash on all-NaN input. Please add the same fix there.
> 2. **Test coverage for non-NaN rows.** The test only sets row 0 to NaN. Please also assert that the other rows still produce valid token indices (i.e., NaN in one request doesn't affect other requests in the batch).
> 3. **Set `success` output to `false` for NaN rows.** Currently NaN rows silently output index 0, which is indistinguishable from legitimately sampling token 0. The `success` output tensor should be set to `false` for these rows so callers can detect the failure.
> 
> Once these are addressed, let's get this merged.

Hi @yzh119 , the fixes are added, detailed description has been updated in the PR description.

yzh119:
@flashinfer-bot run

yzh119:
@flashinfer-bot run

yzh119:
Hi @kahyunnam would you taking a look at this PR as well? Relevant issue #2455 .

vadiklyutiy:
Could you update, are you continue to work on this PR?

Issue #2306, opened 2026-01-07
[Feature Request] BLASST: Dynamic BLocked Attention Sparsity via Softmax Thresholding


wenscarl:
https://www.arxiv.org/pdf/2512.12087

babaozhouy5:
Any updates?

DomBrown:
I'm working on this for Blackwell, starting with the prefill kernel. I don't have an ETA yet, but hopefully not too long

DomBrown:
PR open

bobboli:
Hopper prefill based on fmha_v2: https://github.com/flashinfer-ai/flashinfer/pull/2446

bobboli:
Blackwell prefill and decode based on trtllm-gen: https://github.com/flashinfer-ai/flashinfer/pull/2477

saltyminty:
Marking as closed since the above merged PRs address this issue.

PR #2477, opened 2026-02-03
feat: Add TRTLLM-Gen Skip-Softmax kernels for prefill and decode

DomBrown:
## 📌 Description

This PR adds initial Blackwell support for Skip-Softmax attention sparsity. See:

- NVIDIA Blog [post](https://developer.nvidia.com/blog/accelerating-long-context-inference-with-skip-softmax-in-nvidia-tensorrt-llm/)
- BLASST Paper on [arXiv](https://arxiv.org/abs/2512.12087)

The skip-softmax threshold is added as an optional parameter. If it is `None` then 'normal' attention is used, otherwise we will use skip-softmax with the specified threshold.

I have added unit tests by parameterising the existing tests with a skip threshold of zero. This allows us to compare to a 'normal' attention reference kernel.

## 🔍 Related Issues

#2306 is closed by this PR.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Reviewer Notes

I'm new to FlashInfer so may have missed an API update somewhere. I will address this if any issues are found :)

## Summary by CodeRabbit

* **New Features**
  * Added optional softmax-skipping controls (enable flag and threshold scale factor) across prefill and decode public APIs to allow selective skip-softmax behavior while preserving defaults.

* **Chores**
  * Updated artifact repository paths and checksums for FMHA artifacts.

* **Tests**
  * Extended and parameterized attention tests and helpers to cover and gate the new softmax-skipping paths.

### DIFF
--- a/flashinfer/decode.py
+++ b/flashinfer/decode.py
@@ -1249,7 +1249,11 @@ class BatchDecodeWithPagedKVCacheWrapper:
         q_len_per_req : int
             The number of query tokens per request, if not provided, will be set to ``1``.
         skip_softmax_threshold_scale_factor: Optional[float] = None
-            threshold scale factor for skipping softmax.
+            threshold scale factor for skipping softmax operations.
+            Providing a value for this parameter enables skip-softmax sparsity as described in: https://arxiv.org/abs/2512.12087
+            If no value is provided, then standard attention is used.
+            Setting the threshold to a higher value generally increases kernel performance at the cost of accuracy degradation.
+            The actual threshold value equals the provided threshold_scale_factor divided by the context length.
         Returns
         -------
         Union[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]]
@@ -2224,7 +2228,11 @@ def trtllm_batch_decode_with_kv_cache(
         When None, all requests use uniform query length specified by ``q_len_per_req``.
 
     skip_softmax_threshold_scale_factor: Optional[float] = None
-        threshold scale factor for skipping softmax.
+        threshold scale factor for skipping softmax operations.
+        Providing a value for this parameter enables skip-softmax sparsity as described in: https://arxiv.org/abs/2512.12087
+        If no value is provided, then standard attention is used.
+        Setting the threshold to a higher value generally increases kernel performance at the cost of accuracy degradation.
+        The actual threshold value equals the provided threshold_scale_factor divided by the context length.
 
     Returns
     -------
--- a/flashinfer/prefill.py
+++ b/flashinfer/prefill.py
@@ -3625,7 +3625,11 @@ def trtllm_batch_context_with_kv_cache(
     sinks : Optional[List[torch.Tensor]] = None
         additional value per head in the denominator of the softmax.
     skip_softmax_threshold_scale_factor: Optional[float] = None
-        threshold scale factor for skipping softmax.
+        threshold scale factor for skipping softmax operations.
+        Providing a value for this parameter enables skip-softmax sparsity as described in: https://arxiv.org/abs/2512.12087
+        If no value is provided, then standard attention is used.
+        Setting the threshold to a higher value generally increases kernel performance at the cost of accuracy degradation.
+        The actual threshold value equals the provided threshold_scale_factor divided by the context length.
     Returns
     -------
     out: Union[torch.Tensor, FP4Tensor]


tests/attention/test_trtllm_gen_attention.py
```diff
@@ -1297,6 +1336,7 @@ def test_trtllm_batch_decode_head_dim_256(
 @pytest.mark.parametrize("max_in_kv_len", [4096, 8192, 16384, 32768, 65536, 131072])
 @pytest.mark.parametrize("head_dim", [128])
 @pytest.mark.parametrize("device_scale", [True, False])
+@pytest.mark.parametrize("skips_softmax", [False])
```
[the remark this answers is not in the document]
DomBrown: This is deliberate and will be enabled at a later time.

DomBrown:
Looks like some failures occurred. I will look tomorrow -- looks like I missed a parameter somewhere

DomBrown:
Should be fixed now. If you could trigger CI that would be great (I don't think I'm allowed yet)

yzh119:
@flashinfer-bot run

flashinfer/decode.py
```diff
@@ -1241,6 +1248,8 @@ def run(
             Only supported for >= sm90, and currently only for FA2 and CUDA core decode.
         q_len_per_req : int
             The number of query tokens per request, if not provided, will be set to ``1``.
+        skip_softmax_threshold_scale_factor: Optional[float] = None
+            threshold scale factor for skipping softmax.
```
bkryu: Can you provide a bit more detail in the docstring here for the user? What should the user takeaway from this? Perhaps providing the URL to the paper (https://arxiv.org/abs/2512.12087) and referring to it might help.

If the default value is `None`, will it not be skipping at all or is there some internal default value? What are the tradeoffs for setting to a higher value? If you could provide something like "setting the threshold to a higher value generally increases kernel performance at the risk of accuracy degradation. Typical values range from ...", it would make the documentation much friendlier

Moreover, it should help frameworks integrating this new addition understand how to use it well.

flashinfer/decode.py
```diff
@@ -2207,6 +2223,9 @@ def trtllm_batch_decode_with_kv_cache(
         Only supported by trtllm-gen backend. Must be provided together with ``max_q_len``.
         When None, all requests use uniform query length specified by ``q_len_per_req``.
 
+    skip_softmax_threshold_scale_factor: Optional[float] = None
+        threshold scale factor for skipping softmax.
+
```
bkryu, line 2236: Ditto

flashinfer/prefill.py
```diff
@@ -3608,7 +3624,8 @@ def trtllm_batch_context_with_kv_cache(
         Layout of kv-cache, can be "HND" or "NHD", default is "HND".
     sinks : Optional[List[torch.Tensor]] = None
         additional value per head in the denominator of the softmax.
-
+    skip_softmax_threshold_scale_factor: Optional[float] = None
+        threshold scale factor for skipping softmax.
```
bkryu: Ditto

flashinfer/decode.py
DomBrown: Updated

tests/attention/test_trtllm_gen_attention.py
```diff
@@ -558,6 +568,7 @@ def _test_trtllm_batch_prefill(
         kv_layout=kv_layout,
         enable_pdl=enable_pdl,
         sinks=(sink if enable_sink else None),
+        skip_softmax_threshold_scale_factor=skip_softmax_threshold_scale_factor,
```
yzh119, line 571: This means we didn't test the case where skip_softmax_threshold_scale_factor is not 0 right?
Is there a way we can test the `skip_softmax_threshold_scale_factor != 0`?

DomBrown: We would have to write an entire reference kernel in torch. It's something I intend to come back to later. It's non trivial because the skipping is related to the tile size (the skip happens at tile granularity), which changes based on the selected cubin under the hood.

DomBrown: This is currently tested very robustly in trtllm-gen, I think it's safe for now

yzh119: sounds good, we can implement it later if we have good ideas about how to verify it.

yzh119:
@yongwww can you take a look at failed UT?

yongwww:
> @yongwww can you take a look at failed UT?

pr #2500 broke the workflow, we might need to revert that change for now: https://github.com/flashinfer-ai/flashinfer/pull/2524

PR #2495, opened 2026-02-04
fix: add support check for gemm config for cutlass moe

nv-yunzheq:
## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Added runtime checks to block unsupported kernel configurations and provide clearer, actionable error messages for incompatible settings.
  * Enforced output-size alignment and disallowed certain fusion options when a memory-limited execution mode is selected, preventing silent failures and aiding profiling/tuning.


csrc/nv_internal/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_gemm_template_dispatch.h
```diff
@@ -980,6 +980,16 @@ template <typename T, typename WeightType, typename OutputType, typename ScaleBi
 void MoeGemmRunner<T, WeightType, OutputType, ScaleBiasType>::moeGemmBiasAct(
     GroupedGemmInput<T, WeightType, ScaleBiasType, OutputType> inputs,
     TmaWarpSpecializedGroupedGemmInput hopper_inputs) {
+  // For NoSmem epilogue schedule, output N must be 256-bit aligned.
+  // For gated activation, this is automatic if the usual alignment requirement is met.
+  // This check is here so the autotuner can catch invalid tactics during profiling.
+  if (inputs.gemm_config.epilogue_schedule == cutlass_extensions::EpilogueScheduleType::NO_SMEM &&
```
djns99: If we put this in runGemm/dispatchToArch we don't need to have two copies of this check.

Maybe [here](https://github.com/flashinfer-ai/flashinfer/blob/9bf007d7403f0a394fda44abd4170b354cac3f05/csrc/nv_internal/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_gemm_template_dispatch.h#L809) since this is only relevant for SM90+

nv-yunzheq: `dispatchToArch` doesn't work as we could not know if the activation is gated or not in the function.
`runGemm` works, but to align with [the logic ](https://github.com/flashinfer-ai/flashinfer/blob/9bf007d7403f0a394fda44abd4170b354cac3f05/csrc/fused_moe/cutlass_backend/cutlass_fused_moe_kernels.cuh#L3602)in the moe runner code, I think it's better to split the gemm1 and gemm2 logic separately to make it more clear and align with the original logic

djns99: We dont need to check `isGatedActivation` here
```
  size_t const fc1_out_size =
      ((!use_ampere_activation_fusion) && is_gated_activation) ? inter_size * 2 : inter_size;
```
This line sets the value of N correctly. The original check is only working with `inter_size` so needs to explicitly check we are in the non-gated case

nv-yunzheq: Thanks. Updated to `dispatchToArch`

yzh119 (approved):

Issue #2493, opened 2026-02-04
Perf improvement for CuTe DSL GDN Decode Kernel


kahyunnam:

ameynaik-hub:
ameynaik-hub

ameynaik-hub:
CUTE= this version
FI = flashinfer pretranpose version gdn_decode.py 
spd = speedup.

B | T=1 CUTE | T=1 FI | T=1 spd | T=2 CUTE | T=2 FI | T=2 spd | T=3 CUTE | T=3 FI | T=3 spd | T=4 CUTE | T=4 FI | T=4 spd
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
1 | 10.2 | 9.2 | 0.90x | 10.2 | 13.3 | 1.30x | 12.3 | 15.4 | 1.25x | 12.3 | 15.4 | 1.25x
2 | 10.2 | 11.3 | 1.10x | 10.2 | 15.4 | 1.50x | 12.3 | 15.4 | 1.25x | 12.3 | 17.4 | 1.41x
4 | 10.2 | 11.2 | 1.10x | 10.2 | 17.3 | 1.69x | 12.3 | 19.4 | 1.58x | 12.3 | 19.4 | 1.58x
8 | 10.2 | 13.3 | 1.30x | 12.3 | 19.4 | 1.58x | 14.4 | 23.6 | 1.64x | 14.5 | 25.6 | 1.77x
16 | 12.4 | 17.4 | 1.41x | 16.4 | 27.6 | 1.69x | 18.5 | 33.8 | 1.83x | 20.5 | 37.9 | 1.85x
32 | 18.5 | 25.6 | 1.39x | 24.6 | 41.8 | 1.70x | 26.7 | 52.2 | 1.96x | 30.7 | 62.5 | 2.03x
64 | 28.7 | 50.2 | 1.75x | 36.9 | 72.7 | 1.97x | 45.1 | 92.3 | 2.05x | 51.2 | 113.6 | 2.22x
128 | 51.1 | 91.1 | 1.78x | 65.6 | 130.0 | 1.98x | 77.9 | 173.1 | 2.22x | 92.2 | 214.0 | 2.32x
256 | 93.1 | 173.1 | 1.86x | 120.9 | 240.6 | 1.99x | 145.4 | 324.6 | 2.23x | 172.1 | 406.5 | 2.36x
[...]

PR #2422, opened 2026-01-27
refactor: reduce hopper's gdn prefill compilation time and fix docstring.

yzh119:
## 📌 Description

This PR implements these features:
1. accelerate hopper's gdn prefill compilation time by split compilation
2. fix the docstring of gdn prefill kernel, instead of [N, H, K, V], it expects [N, H, V, K]

## 🔍 Related Issues

#2276 

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [x] Tests have been added or updated as needed.
- [x] All tests are passing (`unittest`, etc.).

## Reviewer Notes

cc @guangyunh-nv 

## Release Notes

* **New Features**
  * Enhanced JIT module generation for GDN prefill kernels with template-driven compilation and separate kernel instantiation.

* **Improvements**
  * JIT specification now intelligently handles C++ standard flags, applying defaults only when not already specified.

* **Documentation**
  * Clarified final state memory layout description for GDN prefill operations.

PR #2405, opened 2026-01-23
perf: improve gdn decode cute-dsl kernels

yzh119:
## 📌 Description

Follow up of #2370 , this PR improves the benchmark scripts and add comparison with baselines:
* benchmark using cupti with l2 flush
* compare with sglang's `fused_sigmoid_gating_delta_rule_update` function (with tile size optimization mentioned by @ vadiklyutiy).

this PR also implements some optimizations on the original gdn kernel:
* use fastmath as much as we can
* change "/" to multiply
* Use `cutlass.range_constexpr` and `cutlass.const_expr` whenever possible
* fuse scale and inv_norm_q
* For mtp, store state in registers directly, without load/write to shared memory, and remove cpasync
* Vectorized memory access.

## Performance on B200

Non MTP setting
```
> python benchmarks/bench_gdn_decode.py --batch-size 1 2 4 8 16 32 64 128 256 --compare --verify

=== Correctness Verification ===
Batch=8:
  Pretranspose: PASS
  Nontranspose: PASS
Batch=16:
  Pretranspose: PASS
  Nontranspose: PASS
Batch=32:
  Pretranspose: PASS
  Nontranspose: PASS
Batch=64:
  Pretranspose: PASS
  Nontranspose: PASS


========================================================================================================================
GDN Decode Benchmark: FlashInfer vs Triton, Pretranspose vs Nontranspose
Config: q_heads=16, k_heads=16, v_heads=32, head_size=128, dtype=bfloat16, qk_l2norm=ON
========================================================================================================================

 batch | FI-PreTr FI-NonTr | TR-PreTr TR-NonTr | FI/TR-Pre FI/TR-Non | Pre/Non-FI Pre/Non-TR
       |     (us)     (us) |     (us)     (us) |   speedup   speedup |    speedup    speedup
------------------------------------------------------------------------------------------------------------------------
     1 |     3.74     5.06 |     5.95     4.35 |    1.59x    0.86x |    1.35x    0.73x
     2 |     4.29     5.89 |     6.37     5.02 |    1.49x    0.85x |    1.37x    0.79x
     4 |     5.41     7.78 |     7.58     6.66 |    1.40x    0.86x |    1.44x    0.88x
     8 |     7.65    12.03 |     9.95    10.21 |    1.30x    0.85x |    1.57x    1.03x
    16 |    12.61    19.30 |    16.83    15.81 |    1.34x    0.82x |    1.53x    0.94x
[...]

PR #2370, opened 2026-01-18
feat: [Qwen3-Next] Add Cute DSL GDN decode kernel and  tests

HongliMi:
Co-author:  @zhou9402 @liz-badada @xutizhou

## 📌 Description

This PR integrates **three versions** of the Gated Delta Rule (GDN) **Decode kernels** into FlashInfer, implemented using **CuTe DSL** for SM90 (Hopper)  and SM100 (Blackwell) GPUs. These kernels enable efficient linear attention decoding for models like Qwen3-Next.

### 🎯 Features Added

1. **Pretranspose Decode Kernel** (`gated_delta_rule_decode_pretranspose`)
   - State layout: `[B, HV, V, K]` (K-last)

### 🏗️ Architecture

All three versions follow FlashInfer's integration pattern:
- **CuTe DSL kernels** with JIT compilation and caching
- **Python API layer** with `@flashinfer_api` decorator
- **Reference implementations** for correctness validation
- **Comprehensive unit tests** with various configurations (GQA, GVA, etc.)

### 📊 Performance Highlights

**B200**（HBM3e，192GB，8TB/s）
```
GDN Decode Comparison: FlashInfer (Pretranspose) vs Triton
Config: q_heads=16, k_heads=16, v_heads=32, head_size=128, dtype=bfloat16, qk_l2norm=ON
----------------------------------------------------------------------------------------------------
 batch FlashInfer(us)   Triton(us)  FI TFLOPS  TR TFLOPS    Speedup
----------------------------------------------------------------------------------------------------
     1           4.03         5.92       0.78       0.53       1.47x
     2           4.58         6.46       1.37       0.97       1.41x
     4           5.74         7.49       2.19       1.68       1.30x
     8           8.61         9.95       2.92       2.53       1.16x
    16          14.19        16.83       3.55       2.99       1.19x
    32          24.51        31.90       4.11       3.16       1.30x
    64          49.60        58.50       4.06       3.44       1.18x
   128          92.26       113.90       4.36       3.54       1.23x
   256         175.36       225.57       4.59       3.57       1.29x
   512         340.45       442.26       4.73       3.64       1.30x
----------------------------------------------------------------------------------------------------
```

### 📁 Files Changed

**Core Implementation:**
- `flashinfer/gdn_decode.py`
  - Three CuTe DSL kernel variants
  - JIT compilation with caching
  - API functions with comprehensive validation

**Testing:**
- `tests/gdn/test_decode_delta_rule.py`
  - Unit tests for all three versions
[...]

PR #2498, opened 2026-02-05
Ameyn/gdn decode cutedsl kernel

ameynaik-hub:
Implements high-performance Gated Delta Rule linear attention kernel supporting fixed sequence lengths T=1, T=2, T=3, T=4 using NVIDIA CuTe-DSL.

Key features:
- H state layout: K-last [B, HV, V, K] where K is the contiguous (fastest) dimension
- L2-normalized Q/K with configurable scale
- Gated exponential decay via softplus
- Delta rule updates: v_delta = beta * (v - pred)
- Async H memory loading with aggressive pipelining
- BF16 tensors with FP32 compute

Also includes:
- benchmark_gated_delta_rule.py: Simple benchmark script for measuring kernel perf
- Updated __init__.py exports

## 📌 Description
Implements high-performance Gated Delta Rule linear attention kernel supporting fixed sequence lengths T=1, T=2, T=3, T=4 using NVIDIA CuTe-DSL.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * Added CUDA-accelerated Gated Delta Rule kernel with optimized paths for sequence lengths 1–4 and new public API entries gated_delta_rule and GatedDeltaRuleKernel (exported when available).

* **Benchmarks**
  * New benchmark for the gated_delta_rule kernel and integration of an improved CuTe-DSL variant into the benchmark suite and CLI, with per‑T summaries and performance reporting.

* **Tests**
  * End-to-end tests exercising the improved kernel for T=1–4, availability guards, and updated reference tests to support explicit state storage dtype handling.

### DIFF
--- a/benchmarks/bench_gdn_decode.py
+++ b/benchmarks/bench_gdn_decode.py
@@ -64,7 +64,7 @@ from flashinfer.testing import bench_gpu_time
 
 # Import the gdn_decode_klast_bf16_state kernel for benchmarking (T=1..4, bf16 state, K-last)
 try:
-    from flashinfer.cute_dsl.gated_delta_rule import (
+    from flashinfer.gdn_kernels.gdn_decode_bf16_state import (
         gated_delta_rule as gdn_decode_klast_bf16_state,
     )
 
--- a/flashinfer/gdn_kernels/gdn_decode_bf16_state.py
+++ b/flashinfer/gdn_kernels/gdn_decode_bf16_state.py
@@ -13,8 +13,12 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 
-Gated Delta Rule Kernel (Unified Modular) using CuTe-DSL
-========================================================
+Gated Delta Rule Decode Kernel (BF16 Hidden State) - CuTe-DSL Implementation
+============================================================================
+
+RELOCATED: This file was previously located at flashinfer/cute_dsl/gated_delta_rule.py
+           and has been moved to flashinfer/gdn_decode/gdn_decode_bf16_state.py
+           to better reflect its domain-specific purpose (GDN decode with BF16 state).
 
 High-performance CUDA kernel implementing the Gated Delta Rule linear attention
 mechanism for decode-phase inference, supporting sequence lengths T=1, T=2, T=3, T=4.
--- a/flashinfer/gdn_decode.py
+++ b/flashinfer/gdn_decode.py
@@ -62,7 +62,7 @@ except ImportError:
 
 # GDN decode K-last bf16 state kernel (T=1..4, bf16 state, K-last layout) - optional backend
 try:
-    from .cute_dsl.gated_delta_rule import (
+    from .gdn_kernels.gdn_decode_bf16_state import (
         gated_delta_rule as _gated_delta_rule_gdn_decode_klast_bf16_state,
     )
 
--- a/flashinfer/gdn_kernels/__init__.py
+++ b/flashinfer/gdn_kernels/__init__.py
@@ -1,0 +1,33 @@
+"""
+GDN (Gated Delta Rule) Kernels - CuTe DSL Implementations
+=========================================================
+
+This module provides CuTe-DSL implementations of GDN kernels.
+
+The main gdn_decode.py and gdn_prefill.py files at the top level contain reference
+implementations and JIT-compiled kernels. This submodule provides high-performance
+CuTe DSL variants optimized for specific use cases.
+
+Exported Kernels:
+- gated_delta_rule: BF16 hidden state decode kernel (T=1,2,3,4)
+- GatedDeltaRuleKernel: Kernel class for advanced usage
+"""
+
+from typing import Optional, Type
+
+try:
+    from .gdn_decode_bf16_state import (
+        gated_delta_rule,
+        GatedDeltaRuleKernel,
+    )
+
+    _has_cute_dsl = True
+except ImportError:
+    _has_cute_dsl = False
+    gated_delta_rule = None  # type: ignore
+    GatedDeltaRuleKernel: Optional[Type] = None  # type: ignore
+
+__all__ = [
+    "gated_delta_rule",
+    "GatedDeltaRuleKernel",
+]
--- a/tests/gdn/test_decode_delta_rule.py
+++ b/tests/gdn/test_decode_delta_rule.py
@@ -43,7 +43,7 @@ from flashinfer.utils import get_compute_capability
 
 # Import the gdn_decode_klast_bf16_state kernel (T=1..4, bf16 state, K-last layout)
 try:
-    from flashinfer.cute_dsl.gated_delta_rule import (
+    from flashinfer.gdn_kernels.gdn_decode_bf16_state import (
         gated_delta_rule as gdn_decode_klast_bf16_state,
     )
 


yzh119:
@ameynaik-hub would you mind comparing with https://github.com/flashinfer-ai/flashinfer/blob/main/flashinfer/gdn_decode.py ? It's better to unify the file location and interface.

ameynaik-hub:
@yzh119 compared benchmark here https://github.com/flashinfer-ai/flashinfer/issues/2493#issuecomment-3850159474 

https://github.com/flashinfer-ai/flashinfer/blob/main/flashinfer/gdn_decode.py kernel supports
K-major and V-major `h` state layout. Also I believe it supports seqlen > 4. 

on other hand, this kernel only support V-major (K contiguous , k -last dim) `h` state layout which is used for qwen models and fixed seqlen <=4 only. 

Any suggestions on how to merge?

vadiklyutiy:
> @ameynaik-hub would you mind comparing with https://github.com/flashinfer-ai/flashinfer/blob/main/flashinfer/gdn_decode.py ? It's better to unify the file location and interface.

and test as well

vadiklyutiy:
> @yzh119 compared benchmark here [#2493 (comment)](https://github.com/flashinfer-ai/flashinfer/issues/2493#issuecomment-3850159474)
> 


I think it's worth to copy results here

vadiklyutiy:
New kernel is always better than old with pretranspose except batch=1 that we can excuse. Does it make sense to keep old pretranspose varian?

aditya-narayan5:
@ameynaik-hub  I might have overlooked details during reading but does anyone have a source for this:

> V-major (K contiguous , k -last dim) h state layout which is used for qwen models

ameynaik-hub:
> New kernel is always better than old with pretranspose except batch=1 that we can excuse. Does it make sense to keep old pretranspose varian?


@vadiklyutiy since it has support for T>4 (seqlen) and also support for v-last dim of `h` I thought it can be kept as is?

ameynaik-hub:
https://github.com/flashinfer-ai/flashinfer/pull/2498#issuecomment-3855179165

https://github.com/flashinfer-ai/flashinfer/pull/2422 @aditya-narayan5

vadiklyutiy:
> > New kernel is always better than old with pretranspose except batch=1 that we can excuse. Does it make sense to keep old pretranspose varian?
> 
> @vadiklyutiy since it has support for T>4 (seqlen) and also support for v-last dim of `h` I thought it can be kept as is?

As I remember there were several different interfaces/functions. I meant that processes case seq_len=1, k-last.
Of course I have not tried to propose unique interfaces, just want to avoid duplication

yzh119:
@ameynaik-hub can you reuse the gdn decode interface and set your implementation as default when we use k-layout (V-major) layout and T <=4?

kahyunnam (approved):
@ameynaik-hub this LGTM as the initial PR to just get the kernel into the Flashinfer codebase, approved. (still need a codeowner approval for this directory to merge ... maybe @yzh119 or @bkryu ). We can follow up with another PR for some work for the API integration + Flashinfer testing/benchmarking. Let me know if you have the bandwidth to work on this integration; if not, I can probably work on it next week.

yzh119:
Hi @kahyunnam , I encourage not to push more code to `flashinfer/cute_dsl` anymore as we plan to categorize modules by functionalities, not sources. Other PRs are starting to remove codes out of `flashinfer.cute_dsl`.

The required work to unify the existing gdn decode and this PR should be minimal, I can help if you need any assitance with it.

flashinfer/cute_dsl/benchmark_gated_delta_rule.py
```diff
@@ -0,0 +1,167 @@
...
+import torch
+
+
+def get_l2_cache_size():
+    """Get L2 cache size in bytes for the current GPU."""
+    return torch.cuda.get_device_properties(0).L2_cache_size
+
+
+def benchmark(
+    func, num_iterations=100, n_warmup=10, flush_l2=True, use_dummy_matmul=True
```
yzh119: We already have benchmarking APIs https://docs.flashinfer.ai/generated/flashinfer.testing.bench_gpu_time.html#flashinfer.testing.bench_gpu_time, please refer to https://github.com/flashinfer-ai/flashinfer/blob/main/benchmarks/bench_gdn_decode.py#L1088-L1095 on how to use these APIs.

flashinfer/gdn_kernels/gdn_decode_bf16_state.py
```diff
@@ -0,0 +1,1823 @@
...
+        if T == 1:
+            launch_fn = gated_delta_rule_launch_seqlen1
+        elif T == 2:
+            launch_fn = gated_delta_rule_launch_seqlen2
+        elif T == 3:
+            launch_fn = gated_delta_rule_launch_seqlen3
+        else:  # T == 4
+            launch_fn = gated_delta_rule_launch_seqlen4
+
+        _compiled_kernels[cache_key] = cute.compile(
```
yzh119, line 2025: please enable tvm-ffi with  `options="--enable-tvm-ffi",`, otherwise the host side overhead of converting torch.tensor to dlpack would be non-neglibile.

Reference can be found at https://github.com/flashinfer-ai/flashinfer/pull/2279

flashinfer/cute_dsl/benchmark_gated_delta_rule.py
vadiklyutiy: I just curious why that wasn't addressed and pointed in #2370 but was here?

yzh119: The benchmarking script in #2370 should use bench_gpu_time as well. I didn't notice that when reviewing that PR (it's my bad) and it was fixed in #2405.

Let's get it right in one shot in this PR.

flashinfer/gdn_kernels/gdn_decode_bf16_state.py
ameynaik-hub: Thanks will do.

ameynaik-hub:
> @ameynaik-hub can you reuse the gdn decode interface and set your implementation as default when we use k-layout (V-major) layout and T <=4?

@yzh119 yeah I can do that. 

I noticed that my kernel utilizes bf16 inputs for the state `h`, which appears to be compatible with sglang and other models. However, it seems that the default state for `h` is fp32. I am investigating the possibility of optimizing this configuration for this specific case. 

cc: @vadiklyutiy @kahyunnam

yzh119:
cc @guangyunh-nv (gdn prefill's author) do you have any insights on whether it's okay to use f16 state?

guangyunh-nv:
FP16 state might not be a good choice due to the dynamic range limitation. The range is not guaranteed and purely depends on the model activation dynamics. So this purely depends on the model behavior.

Datapoint: 
Previously with MiniMax M1 (lighting attention, similar to GDN but gating is not data dependent), the model only runs with BF16 due to its dynamical range grows exponentially somehow and then saturate.

BF16 maybe safer, but the accuracy would be a problem. If algorithm side decide that this is tolerable, then I think we should support it.

I think what we can do at the moment is to tolerate all kinds of state input, and learn from the production environment feedback :)

xutizhou:
Have you tested the end-to-end accuracy with FP16 SSM state? Does the performance remain unchanged on common benchmarks such as MMLU, GSM8K, etc.?

vadiklyutiy:
> I think what we can do at the moment is to tolerate all kinds of state input, and learn from the production environment feedback :)

I don't think this is right approach. Imagine car manufacturer introduce new self driving car and say let learn from road environment. Don't think a lot of people would be happy.

guangyunh-nv:
> > I think what we can do at the moment is to tolerate all kinds of state input, and learn from the production environment feedback :)
> 
> I don't think this is right approach. Imagine car manufacturer introduce new self driving car and say let learn from road environment. Don't think a lot of people would be happy.

This is called false analogy. A proper analogy will be flashinfer is part supplier. It is the model supplier to decide which part satisfy the minimium safety requirement.

ameynaik-hub:
Is someone looking into fp16/bf16 state benchmarks? 

meanwhile I am looking into fp32 state based improvements. looking at the current implementation https://github.com/flashinfer-ai/flashinfer/pull/2405 I see it is at 50-60% SOL at BS >=64, with bf16 I was able to achieve 60-75% SOL at BS>=64. so I am hoping there is still opportunity to improve.

vadiklyutiy:
> I noticed that my kernel utilizes bf16 inputs for the state `h`, which appears to be compatible with sglang and other models. However, it seems that the default state for `h` is fp32. I am investigating the possibility of optimizing this configuration for this specific case.

In vllm `h.dtype=bf16`. So, seems no need to worry about that

aditya-narayan5:
@ameynaik-hub Thanks for linking to the PR by @yzh119:

>fix the docstring of the GDN prefill kernel; instead of N,H,K,V, it expects N,H,V,K 

Could the maintainers clarify **why** the N,H,V,K layout is preferred? 

Most GDN reference implementations (e.g., the FLA project) expect the recurrent state in N,H,K,V, and the reference used in the tests in PR #2276 follows the same convention, we explicitly transpose the kernel output from N,H,V,K for comparison against the reference.

PR #2370 already provides both a pre‑transpose (N,H,V,K) and a **non‑transpose (N,H,K,V) path**, and the inline comments **recommend the non‑transpose version**. I’ve searched the Qwen3 documentation and the official Hugging Face Transformers implementation (which uses FLA under the hood) and haven’t found any explicit requirement for the N,H,V,K layout.

Is there a downstream consumer or hardware target that mandates this layout?

vadiklyutiy:
> @ameynaik-hub Thanks for linking to the PR by @yzh119:
> 
> > fix the docstring of the GDN prefill kernel; instead of N,H,K,V, it expects N,H,V,K
> 
> Could the maintainers clarify **why** the N,H,V,K layout is preferred?
> 
> Most GDN reference implementations (e.g., the FLA project) expect the recurrent state in N,H,K,V, and the reference used in the tests in PR #2276 follows the same convention, we explicitly transpose the kernel output from N,H,V,K for comparison against the reference.
> 
> PR #2370 already provides both a pre‑transpose (N,H,V,K) and a **non‑transpose (N,H,K,V) path**, and the inline comments **recommend the non‑transpose version**. I’ve searched the Qwen3 documentation and the official Hugging Face Transformers implementation (which uses FLA under the hood) and haven’t found any explicit requirement for the N,H,V,K layout.
> 
> Is there a downstream consumer or hardware target that mandates this layout?

@yzh119 convinced me in Slack that k-last is better. vllm already moved to k-last layout, and this is btw give perf improvement itself:)
Also FI's GDN decode use K-last layout. 
@yzh119 @guangyunh-nv could you pls comment on that.

ameynaik-hub:
generally speaking if it is k-last (k contiguous), you can vectorize the reads along k-dim, and the inner products with k and q can be performed parallelly across rows more easily.

guangyunh-nv:
@vadiklyutiy The reason for K-last is as follows,

In prefill, you need to repeatly update State in the kernel mainloop, so you better put state in register (hopper). This means you put S in the A operand. This further avoids the repeated transposition and updating the State in B in SMEM style kernel. It is very critical to kernel performance.

I believe Blackwell has the same constraint as due to the TMEM accesse pattern.

ameynaik-hub:
cute dsl bf16 `h` perf on B200


batch | T | time(us)
-- | -- | --
1 | 1 | 3.62
1 | 2 | 5.86
1 | 3 | 6.94
1 | 4 | 7.79
4 | 1 | 4.9
4 | 2 | 6.62
4 | 3 | 7.65
4 | 4 | 8.64
8 | 1 | 7.04
8 | 2 | 8.67
8 | 3 | 9.95
8 | 4 | 11.39
16 | 1 | 9.74
16 | 2 | 12.9
16 | 3 | 15.15
16 | 4 | 17.54
32 | 1 | 16.06
32 | 2 | 21.57
32 | 3 | 25.79
32 | 4 | 28.61
64 | 1 | 27.01
64 | 2 | 34.88
64 | 3 | 40.67
64 | 4 | 49.76
128 | 1 | 48.9
128 | 2 | 60.7
128 | 3 | 72.26
128 | 4 | 89.04
256 | 1 | 91.66
256 | 2 | 112.54
256 | 3 | 134.88
256 | 4 | 166.32
512 | 1 | 177.06
512 | 2 | 214.53
512 | 3 | 258.75
512 | 4 | 320.8

flashinfer/gdn_kernels/gdn_decode_bf16_state.py
```diff
@@ -0,0 +1,2046 @@
...
+Key Features:
+- Unified kernel architecture: T=2/3/4 share a single compile-time specialized kernel
+  using Constexpr dispatch, while T=1 uses a separate kernel with persistent K-in-registers
+- L2-normalized Q/K with configurable scale
+- Gated exponential decay of hidden state H via softplus
+- Delta rule updates: v_delta = beta * (v - pred)
+- Bank-conflict-free cross-warp reductions
+- Async H memory loading with aggressive pipelining
+- BF16 tensors with FP32 compute for numerical stability
+- GQA (grouped-query attention) support with configurable H (query) and HV (value) heads
```
guangyunh-nv, line 35: GQA: `HQ = <IntegerRatio> * HV`
GVA: `<IntegerRatio> * HQ = HV`, I confirmed this naming in a group chat with GDN paper author :)

aditya-narayan5:
Thanks to everyone for clarifying, this helped :) @vadiklyutiy  @guangyunh-nv  @ameynaik-hub

guangyunh-nv:
Out of curisity, why the perf degenerate when batch goes from 1 to 4.

batch | T | time(us)
-- | -- | --
1 | 1 | 3.62
4 | 1 | 4.9

I'd assume you have not reached the parallelism limit and bandwidth limit, then the time should remain a constant.

ameynaik-hub:
> Out of curisity, why the perf degenerate when batch goes from 1 to 4.

@guangyunh-nv for BS <=4 I use 4 ctas per head. so for BS=1 it has 32*4 = 128 CTAs and for BS=4 it has 32*4*4 = 512 CTAs
at this regime it is latency bound, MIO throttle appears to be the reason based on ncu reports.

tests/gdn/reference_delta_rule.py
```diff
@@ -137,6 +137,7 @@ def blockwise_linear_attention(
     | torch.Tensor = 1.0,  # float or tensor with num_elems == num_qo_heads
     decay_exponent_offset=0,
     kv_dtype: torch.dtype = torch.float32,
```
guangyunh-nv: I think you should remove `kv_dtype`, it is named as kv_dtype because in some earlier papers call the state directly as KV. It is not dtype for K and V

tests/gdn/reference_delta_rule.py
```diff
@@ -257,6 +263,7 @@ def delta_rule(
     beta: torch.Tensor | None = None,  # [total_seq_len, num_qo_heads]
     scale_factor=1.0,
     kv_dtype: torch.dtype = torch.float32,
```
guangyunh-nv: and here

tests/gdn/reference_delta_rule.py
```diff
@@ -357,6 +365,7 @@ def blockwise_delta_rule(
     block_size: int = 32,
     scale_factor=1.0,
     kv_dtype: torch.dtype = torch.float32,
```
guangyunh-nv: and here

flashinfer/gdn_kernels/gdn_decode_bf16_state.py
ameynaik-hub: @guangyunh-nv sorry do you mean it is called GVA and not GQA?

tests/gdn/reference_delta_rule.py
ameynaik-hub: done, can you please confirm if the change is okay? Thanks!

benchmarks/bench_gdn_decode.py
```diff
@@ -1800,6 +1832,49 @@ def verify_correctness_pretranspose(
 # ============================================================================
 
 
+def gdn_decode_klast_bf16_state_wrapper(
+    q: torch.Tensor,  # [B, T, H_Q, K] where T=1,2,3,4
+    k: torch.Tensor,  # [B, T, H_K, K]
+    v: torch.Tensor,  # [B, T, HV, V]
+    state: torch.Tensor,  # [B, HV, V, K] - K-fast layout (pretranspose)
```
yzh119:
```suggestion
    state: torch.Tensor,  # [B, HV, V, K] - K-last layout (pretranspose)
```

flashinfer/gdn_kernels/gdn_decode_bf16_state.py
yzh119, line 1: Can we moved it to flashinfer/gdn_decode.py (or making it a module)? We don't want to keep the directory`flashinfer/cute_dsl` in the future.

vadiklyutiy:
I would like to speed up this review a bit. The Qwen3.5 headline model has been released today.

flashinfer/gdn_kernels/gdn_decode_bf16_state.py
ameynaik-hub: pushed a change, does this structure look okay to you? 
added in flashinfer/gdn_kernels 
eventually we can not gdn_decode.py under it?

ameynaik-hub:
I have made some improvement for fp32 h state version of gdn_decode

BS,FI-PreTr (us),BS32opt-V2 (us),Speedup
1,3.74,3.52,1.06x
4,5.31,5.31,1.00x
8,7.68,7.58,1.01x
16,13.39,12.74,1.05x
32,22.99,22.72,1.01x
64,54.85,42.38,1.29x
128,92.21,81.39,1.13x
256,172.64,158.62,1.09x
512,336.56,313.69,1.07x

Zihao, I am thinking of creating a seperate PR for it. what do you think? @yzh119

Issue #2400, opened 2026-01-22
[Perf] BatchPrefillWithPagedKVCacheKernel with 'fa3' backend performance degradation in v0.6.1.


XuZhang99:
In v0.6.1, bf16 datatype, compared to `fa2` backend, BatchPrefillWithPagedKVCacheKernel with `fa3` backend has performance degradation, it takes about 2.5x as long as `fa2` backend.

bkryu:
Hi @XuZhang99 , thank you for bringing this to our attention. I can reproduce some regressions on H200. Will begin investigating the issue.

bkryu:
@XuZhang99, would it be possible to share the problem sizes or a reproducer for the cases in which you are seeing performance issues?

XuZhang99:
Yes, I can, but it's a little bit complex, I find this issue when I integrate flashinfer v0.6.1 into [xllm](https://github.com/jd-opensource/xllm), here is my pr: https://github.com/jd-opensource/xllm/pull/764.

If you want to reproduce the same issue, it maybe will take some time to setup env. I don't know if you are willing, but I will provide the details if you want. @bkryu

bkryu:
Perhaps having more details could be useful. Can you:

1. Clarify from which version to 0.6.1 you are going to? Are you bumping from 0.5.3? 0.6.0?
2. Which GPU and CUDA version?
3. For problem size collection, instead of myself trying to run xllm, which you also raise that might be a bit complex 😅, can you try proving a FlashInfer API log? The logging can be done by setting the following environment variables (see [logging docs](https://docs.flashinfer.ai/logging.html)):
```
export FLASHINFER_LOGLEVEL=5
export FLASHINFER_LOGDEST=flashinfer_api.log
```

This will produce a log at `flashinfer_api.log`. For example, a log from the following benchmark:
```
flashinfer/benchmarks$ python3 flashinfer_benchmark.py --routine BatchPrefillWithPagedKVCacheWrapper --backends fa2 fa3 --page_size 16 --batch_size 1 --s_qo 8192 --s_kv 8192 --num_qo_heads 64 --num_kv_heads 8 --head_dim_qk 128 --head_dim_vo 128 --refcheck --causal --q_dtype bfloat16 --kv_dtype bfloat16 --allow_output_mismatch --use_cupti
[WARNING] --use_cupti is deprecated and will be removed in a future release. CUPTI is now enabled by default.
[...]

PR #2530, opened 2026-02-09
pick fa2 for BatchDecodeWithPagedKVCacheWrapper auto backend

saltyminty:
## 📌 Description

BatchDecodeWithPagedKVCacheWrapper used to be hardcoded to select fa2, but now selects fa3 on Hopper due to the refactoring to use determine_attention_backend, which is slower than fa2. Revert to hardcode to fa2 for non-fp8 workloads.

Add auto backend support to flashinfer_benchmark.py for BatchDecodeWithPagedKVCacheWrapper and BatchPrefillWithPagedKVCacheWrapper.

## 🔍 Related Issues

https://github.com/flashinfer-ai/flashinfer/issues/2400

## 🚀 Pull Request Checklist

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * Added an "auto" backend option that records and displays the actual backend as auto(resolved_backend).
  * Performance outputs now include a resolved_backend column showing which backend was used.
  * Auto routing differs for FP8 vs non-FP8 data paths when selecting the actual backend; supported wrappers limited to decode/prefill paged-KV variants.

* **Documentation**
  * CLI/help text updated to document the --backends auto option and supported combinations.

* **Compatibility**
  * Deprecated backend names normalized for consistent reporting.

### DIFF
--- a/benchmarks/routines/attention.py
+++ b/benchmarks/routines/attention.py
@@ -666,8 +666,15 @@ def testBatchDecodeWithPagedKVCacheWrapper(args):
                 o_dtype=q_dtype,
             )
             resolved_backend = resolved_backends.get(backend, backend)
+            wrapper = backend_wrappers.get(backend)
+            if (
+                wrapper is not None
+                and resolved_backend == "fa2"
+                and wrapper.use_tensor_cores
+            ):
+                resolved_backend = "fa2_tc"
             display_backend = (
-                f"auto({resolved_backend})" if backend == "auto" else backend
+                f"auto({resolved_backend})" if backend == "auto" else resolved_backend
             )
             print_perf_metrics(
                 display_backend, median_time, std_time, tflops, tb_per_sec


saltyminty:
bot run

benchmarks/routines/attention.py
```diff
@@ -196,6 +197,13 @@ def parse_attention_args(line, parser):
 
     # Normalize backend names (handle deprecated names)
     args.backends = normalize_backends(args.backends)
+    if "auto" in args.backends and args.routine not in [
+        "BatchDecodeWithPagedKVCacheWrapper",
+        "BatchPrefillWithPagedKVCacheWrapper",
+    ]:
+        raise ValueError(
+            "backend=auto is only supported for BatchDecodeWithPagedKVCacheWrapper and BatchPrefillWithPagedKVCacheWrapper."
+        )
```
bkryu: Unnecessary because the modified backends support list in `flashinfer_benchmark_utils.py`should handle this. Let's reduce hard coded checks as much as possible.

benchmarks/routines/attention.py
```diff
@@ -498,6 +507,9 @@ def testBatchDecodeWithPagedKVCacheWrapper(args):
                 data_type=kv_dtype,
                 block_tables=block_tables,
             )
+            resolved_backends[backend] = backend_wrappers[backend]._backend
+        else:
+            resolved_backends[backend] = backend
```
bkryu, line 504: If a user ran `auto`, it should display `auto`, even if the underlying kernel got routed to a specific backend. The reason here is resolving backends does not help the user; often times, a user may want to know "should I use `auto` or backend X, but typically does not need to know what is running underneath.

Unless you had other specific reasons, we should not resolve backends.

saltyminty: I initially put this in for debugging purposes (so that we would know which backend actually got selected/run).

For example, this is the sample output from `benchmarks/flashinfer_benchmark.py`

```
[PERF] fa2            :: median time 0.488 ms; std 0.001 ms; achieved tflops 8.794 TFLOPs/sec; achieved tb_per_sec 2.203 TB/sec
[PERF] auto(fa2)      :: median time 0.326 ms; std 0.001 ms; achieved tflops 13.190 TFLOPs/sec; achieved tb_per_sec 3.304 TB/sec
```

Would you prefer if it just read auto?

bkryu: It seems like I misread; the current format should be fine.

Unrelated: it's odd that `fa2` and `auto(fa2)` produce different times

saltyminty: Discussed offline: the auto "fa2" selection has tensor cores enabled, so it's actually equivalent to "fa2_tc." Updated to display fa2_tc for clarity.

flashinfer/decode.py
```diff
@@ -1040,14 +1040,20 @@ def plan(
...
+                        self._backend = determine_attention_backend(
+                            self.device,
+                            PosEncodingMode[pos_encoding_mode].value,
+                            False,  # use_fp16_qk_reductions
+                            False,  # use_custom_mask
+                            q_data_type,
+                            kv_data_type,
+                        )
+                    else:
+                        self._backend = "fa2"
```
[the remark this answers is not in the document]
saltyminty: Not an issue.

PR #2550, opened 2026-02-12
Added comments on potential race conditions for a cluster size > 1

christindbose:
## 📌 Description

Adds comments on potential race conditions if this kernel is launched with a thread block cluster size > 1. Ideally we can also throw in an assert to ensure that the cluster size is equal to 1 (this needs including some libraries though). The race condition is because of the specific nature of arriving on an mbarrier and the CTA ID associated with it. 

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

- [✅] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [✅] I have installed the hooks with `pre-commit install`.
- [✅ ] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

## 🧪 Tests

- [✅] All tests are passing (`unittest`, etc.).

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a runtime check enforcing single-block cluster usage to prevent synchronization races and improve GPU computation reliability.

* **Documentation**
  * Clarified inline comments about expected GPU cluster configuration and synchronization assumptions.

### DIFF
--- a/include/flashinfer/attention/hopper/mainloop_mma.cuh
+++ b/include/flashinfer/attention/hopper/mainloop_mma.cuh
@@ -11,6 +11,7 @@
 #include <cutlass/cutlass.h>
 #include <cutlass/numeric_conversion.h>
 #include <cutlass/numeric_types.h>
+#include <cooperative_groups.h>
 
 #include "variants.cuh"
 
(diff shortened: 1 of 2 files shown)


christindbose:
fyi, the fix to avoid race conditions (for a cluster size > 1) is also straightforward. Lemme know if I need to update that. It involves updating the mbarrier locally instead of remote.

include/flashinfer/attention/hopper/quantization/mainloop_mma.cuh
```diff
@@ -79,7 +80,10 @@ CUTLASS_DEVICE void mma_fp8(const Params& mainloop_params, AttentionVariant& var
   if (work_idx != 0) {
     int lane_predicate = cute::elect_one_sync();
     if (cutlass::canonical_warp_idx_sync() == Ktraits::NUM_WARPS - 1 && lane_predicate) {
-#pragma unroll
+    // This assumes a thread block cluster size of 1 (1 CTA/cluster).
++   // A cluster size > 1 could cause race conditions with barrier_O.arrive() below.
++   assert(cooperative_groups::this_cluster().size() == 1);
```
yzh119, line 85: Is it a good idea to assert inside kernels? Why not move the assertion to host side?

include/flashinfer/attention/hopper/mainloop_mma.cuh
```diff
@@ -81,6 +82,9 @@ CUTLASS_DEVICE void mma_f16(
   if (work_idx != 0) {
     int lane_predicate = cute::elect_one_sync();
     if (cutlass::canonical_warp_idx_sync() == Ktraits::NUM_WARPS - 1 && lane_predicate) {
+    // This assumes a thread block cluster size of 1 (1 CTA/cluster).
++   // A cluster size > 1 could cause race conditions with barrier_O.arrive() below.
++   assert(cooperative_groups::this_cluster().size() == 1);
```
yzh119, line 87: ditto

Issue #2504, opened 2026-02-05
[Bug] Autotune picks suboptimal tactic for trtllm_fp4_block_scale_moe kernel at small batch sizes


XiaotongJiang:
## Summary

When using autotune for the `trtllm_fp4_block_scale_moe` kernel, we notice that for small batch sizes, autotune consistently picks suboptimal tactics compared to optimal config even for the same cache key.

## Example

For batch size 1:

**our autotune picks:**
```
bmm_MxE4m3_MxE2m1MxE4m3_Fp32_t128x32x256_s5_et128x32_m256x32x32_cga2x1x1_16dp256b_rM_TN_transOut_schedS_biasM_bN_tma_tmaOpt_clmp_swiGlu_dynBatch_sm100f
```
(tile 32, s5 scheduler)

**optimal autotune picks:**
```
bmm_MxE4m3_MxE2m1MxE4m3_Fp32_t128x16x256u2_s6_et128x16_m256x16x32_cga2x1x1_16dp256b_rM_TN_transOut_schedP2x1x2x3_biasM_bN_tma_tmaOpt_clmp_swiGlu_dynBatch_sm100f
```

## Questions

1. What could be causing autotune to select suboptimal tactics for small batch sizes?
2. Is there a script or guidance available for generating `tuning_configs` for kernels? We would like to bypass the autotune issue by using a pre-configured tuning file similar to:
   https://github.com/flashinfer-ai/flashinfer/blob/main/flashinfer/tuning_configs/v0_1_trtllm_fused_moe_NVIDIA_B200.py

## Environment

- Kernel: `trtllm_fp4_block_scale_moe`
- Batch size: 1 (and other small batch sizes)
- GPU: SM100 (B200)

bkryu:
Hi @XiaotongJiang. Thanks for reporting the issue.

> What could be causing autotune to select suboptimal tactics for small batch sizes?

This could be due to multiple things, but can be summarized as:
1. differences in the autotuner's performance measuring environment and e2e/production environment. Currently the autotuner measures time with a hot L2 cache (which I believe should be fixed). In an e2e setting, this might be different. 
2. For small batch sizes where execution times are very small, noise can contribute to the autotuner selecting suboptimal config.

> Is there a script or guidance available for generating tuning_configs for kernels? We would like to bypass the autotune issue by using a pre-configured tuning file similar to:
[...]

PR #2554, opened 2026-02-13
feat: Add autotuner config caching, thread safety, and documentation

bkryu:
## 📌 Description

**Summary**
This PR adds the ability to persist and reload autotuner configs via a JSON cache file, making it easy for users and inference frameworks to tune once and reuse optimal kernel configurations across runs. It also adds thread safety to the `AutoTuner` singleton and atomic file writes, along with a new documentation page and benchmark harness integration.

**Motivation**
Today, autotuned kernel configs are computed at runtime (expensive). Users and inference frameworks (e.g. vLLM, SGLang) have no easy way to:
* Tune once offline and reuse the results across deployments.
* Incrementally build up a config file as new input shapes are encountered.
* Skip re-profiling for shapes that already have known-good configs.

**Config caching via `autotune(cache=path)`**
The `autotune()` context manager now accepts an optional cache parameter:
```
# Tune and save results
# equivalent to flashinfer.autotune(cache="my_configs.json")
with flashinfer.autotune(True, cache="my_configs.json"): 
    <autotunable_flashinfer_api>

# Now that my_configs.json exists, will load cached configs (no profiling)
with flashinfer.autotune(cache="my_configs.json"):
    <autotunable_flashinfer_api>
```

* On entry, configs are loaded from the cache file if it exists.
* On exit (when `tune_mode=True`), configs are saved back, merging any previously loaded entries with newly profiled ones.
* Cache files are plain JSON, human-readable, and portable across FlashInfer versions (uses runner class names, not positional indices).
* Incremental tuning is supported: multiple sessions accumulate configs in the same file.
* Already-tuned shapes are skipped during profiling (cache hit), with a once-per-(op, runner) log message.

**Thread safety**
* AutoTuner.get() now uses double-checked locking (`threading.Lock`) for safe singleton creation.
* All mutable state operations (`search_cache`, `choose_one`, `save_configs`, `load_configs`, `clear_cache`, mode-flag save/restore) are protected by a threading.RLock on the instance.
* GPU profiling is serialized via the lock, preventing concurrent measurements from interfering.

**Atomic file writes**
* `save_configs()` writes to a temp file then calls `os.replace()` for atomic replacement.
* Concurrent readers never see a partially-written file.
* Process crashes mid-write leave the previous file intact.

**Benchmark harness integration**
* Added `--autotune_cache` to `flashinfer_benchmark.py` shared args.
* All 4 GEMM routines (bmm_fp8, bmm_mxfp8, mm_fp4, mm_mxfp8) and both MoE routines (trtllm_fp4_block_scale_moe, cutlass_fused_moe) pass cache= to autotune() and support cache-only inference via `AutoTuner.get().load_configs()`.

**Documentation**
* New `docs/autotuning.rst` page covering: what autotuning is, how to enable it, config caching (save/load/incremental/cache hits), API reference with behavior matrix, benchmark harness usage, cache file format, and multi-thread/multi-process considerations.

**Files changed**

File | Change
-- | --
flashinfer/autotuner.py | Core: autotune(cache=), save_configs, load_configs, _tactic_to_json/_json_to_tactic, search_cache 4-level fallback, threading locks, atomic writes
tests/utils/test_autotuner_configs.py | New: 35 tests covering save/load round-trips, tactic serialization, autotune(cache=) context manager, search_cache fallback chain, atomic writes, and thread safety
benchmarks/flashinfer_benchmark.py | Added --autotune_cache shared argument
benchmarks/routines/gemm.py | Pass cache= to autotune() in all 4 GEMM routines; load-only fallback
benchmarks/routines/moe.py | Pass cache= to autotune() in both MoE routines; load-only fallback
docs/autotuning.rst | New: full documentation page
docs/index.rst | Added autotuning to toctree

**Example from microbenchmarks**
```
## Save configs the first time:
flashinfer/benchmarks$ python3 flashinfer_benchmark.py --routine mm_fp4 --m 1024 --n 7168 --k 4608 --out_dtype bfloat16 --backends trtllm --use_128x4_sf_layout --use_nvfp4 --refcheck --autotune --autotune_cache autotune_cache.json
2026-02-13 16:32:16,790 - INFO - autotuner.py:313 - flashinfer.jit: [Autotuner]: Autotuning process starts ...
2026-02-13 16:32:16,914 - INFO - autotuner.py:320 - flashinfer.jit: [Autotuner]: Autotuning process ends
2026-02-13 16:32:16,920 - INFO - autotuner.py:1002 - flashinfer.jit: [Autotuner]: Saved 11 configs to autotune_cache.json (11 new, 0 from previous config)
[PERF] trtllm_autotune:: median time 0.027 ms; std 0.000 ms; achieved tflops 2473.969 TFLOPs/sec; achieved tb_per_sec 1.227 TB/sec


## Same command triggers cache hit and skips autotune.
flashinfer/benchmarks$ python3 flashinfer_benchmark.py --routine mm_fp4 --m 1024 --n 7168 --k 4608 --out_dtype bfloat16 --backends trtllm --use_128x4_sf_layout --use_nvfp4 --refcheck --autotune --autotune_cache autotune_cache.json
2026-02-13 16:32:22,802 - INFO - autotuner.py:1044 - flashinfer.jit: [Autotuner]: Loaded 11 configs from autotune_cache.json
2026-02-13 16:32:22,802 - INFO - autotuner.py:313 - flashinfer.jit: [Autotuner]: Autotuning process starts ...
2026-02-13 16:32:22,814 - INFO - autotuner.py:496 - flashinfer.jit: [Autotuner]: Config cache hit for fp4_gemm (runner=TrtllmFp4GemmRunner, source=config file)
2026-02-13 16:32:22,825 - INFO - autotuner.py:320 - flashinfer.jit: [Autotuner]: Autotuning process ends
2026-02-13 16:32:22,832 - INFO - autotuner.py:1002 - flashinfer.jit: [Autotuner]: Saved 11 configs to autotune_cache.json (0 new, 11 from previous config)
[PERF] trtllm_autotune:: median time 0.027 ms; std 0.000 ms; achieved tflops 2485.513 TFLOPs/sec; achieved tb_per_sec 1.233 TB/sec

## The config is a simple json with API, runner type, problem size, and selected tactic:
flashinfer/benchmarks$ cat autotune_config.json 
{
 "_metadata": {
    "flashinfer_version": "0.6.3",
    "cuda_version": "13.0",
    "cublas_version": "13.2.1",
    "cudnn_version": "91900",
    "gpu": "NVIDIA B200"
  },
  "_last_updated_by": {
    "flashinfer_version": "0.6.3",
    "cuda_version": "13.0",
    "cublas_version": "13.2.1",
    "cudnn_version": "91900",
    "gpu": "NVIDIA B200"
  },
  "('fp4_gemm', 'TrtllmFp4GemmRunner', ((1, 2304), (2304, 7168), (-1, 288), (288, 7168), (), (0,), (-1, 7168), (0,), (0,), (33554432,)))": [
    "TrtllmFp4GemmRunner",
    21
  ],
  "('fp4_gemm', 'TrtllmFp4GemmRunner', ((1024, 2304), (2304, 7168), (-1, 288), (288, 7168), (), (0,), (-1, 7168), (0,), (0,), (33554432,)))": [
    "TrtllmFp4GemmRunner",
    23
  ],
...
...
  "('fp4_gemm', 'TrtllmFp4GemmRunner', ((8, 2304), (2304, 7168), (-1, 288), (288, 7168), (), (0,), (-1, 7168), (0,), (0,), (33554432,)))": [
    "TrtllmFp4GemmRunner",
    21
  ]
}
```

## 🔍 Related Issues

https://github.com/flashinfer-ai/flashinfer/issues/2504
https://github.com/flashinfer-ai/flashinfer/issues/2620

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * Persistent save/load for autotuning results with optional cache path and automatic save on exit
  * CLI/benchmark integration to load/use a persistent autotune cache
  * Environment-aware cache validation, atomic file writes, and thread-safe cache handling

* **Documentation**
  * Comprehensive autotuning guide with examples, behavior matrix, and API reference

* **Tests**
  * Extensive tests for serialization, cache round-trips, concurrency, and edge cases

### DIFF
--- a/csrc/flashinfer_sampling_binding.cu
+++ b/csrc/flashinfer_sampling_binding.cu
@@ -20,8 +20,9 @@ using tvm::ffi::Optional;
 void softmax(TensorView workspace_buffer, TensorView logits, TensorView output,
              Optional<TensorView> maybe_temperature_arr, double temperature_val, bool enable_pdl);
 
-void sampling_from_probs(TensorView probs, TensorView output, Optional<TensorView> maybe_indices,
-                         bool deterministic, Optional<TensorView> maybe_seed_arr, uint64_t seed_val,
+void sampling_from_probs(TensorView probs, TensorView output, TensorView valid,
+                         Optional<TensorView> maybe_indices, bool deterministic,
+                         Optional<TensorView> maybe_seed_arr, uint64_t seed_val,
                          Optional<TensorView> maybe_offset_arr, uint64_t offset_val);
 
 void sampling_from_logits(TensorView logits, TensorView output, Optional<TensorView> maybe_indices,
@@ -29,28 +30,28 @@ void sampling_from_logits(TensorView logits, TensorView output, Optional<TensorV
                           uint64_t seed_val, Optional<TensorView> maybe_offset_arr,
                           uint64_t offset_val);
 
-void top_p_sampling_from_probs(TensorView probs, TensorView output,
+void top_p_sampling_from_probs(TensorView probs, TensorView output, TensorView valid,
                                Optional<TensorView> maybe_indices,
                                Optional<TensorView> maybe_top_p_arr, double top_p_val,
                                bool deterministic, Optional<TensorView> maybe_seed_arr,
                                uint64_t seed_val, Optional<TensorView> maybe_offset_arr,
                                uint64_t offset_val);
 
-void top_k_sampling_from_probs(TensorView probs, TensorView output,
+void top_k_sampling_from_probs(TensorView probs, TensorView output, TensorView valid,
                                Optional<TensorView> maybe_indices,
                                Optional<TensorView> maybe_top_k_arr, int64_t top_k_val,
                                bool deterministic, Optional<TensorView> maybe_seed_arr,
                                uint64_t seed_val, Optional<TensorView> maybe_offset_arr,
                                uint64_t offset_val);
 
-void min_p_sampling_from_probs(TensorView probs, TensorView output,
+void min_p_sampling_from_probs(TensorView probs, TensorView output, TensorView valid,
                                Optional<TensorView> maybe_indices,
                                Optional<TensorView> maybe_min_p_arr, double min_p_val,
                                bool deterministic, Optional<TensorView> maybe_seed_arr,
                                uint64_t seed_val, Optional<TensorView> maybe_offset_arr,
                                uint64_t offset_val);
 
-void top_k_top_p_sampling_from_probs(TensorView probs, TensorView output,
+void top_k_top_p_sampling_from_probs(TensorView probs, TensorView output, TensorView valid,
                                      Optional<TensorView> maybe_indices,
                                      Optional<TensorView> maybe_top_k_arr, double top_k_val,
                                      Optional<TensorView> maybe_top_p_arr, double top_p_val,
--- a/csrc/sampling.cu
+++ b/csrc/sampling.cu
@@ -101,11 +101,15 @@ void sampling_from_logits(TensorView logits, TensorView output, Optional<TensorV
   });
 }
 
-void sampling_from_probs(TensorView probs, TensorView output, Optional<TensorView> maybe_indices,
-                         bool deterministic, Optional<TensorView> maybe_seed_arr, uint64_t seed_val,
+void sampling_from_probs(TensorView probs, TensorView output, TensorView valid,
+                         Optional<TensorView> maybe_indices, bool deterministic,
+                         Optional<TensorView> maybe_seed_arr, uint64_t seed_val,
                          Optional<TensorView> maybe_offset_arr, uint64_t offset_val) {
   CHECK_INPUT(probs);
   CHECK_DIM(2, probs);  // probs: (batch_size, vocab_size)
+  CHECK_INPUT(valid);
+  CHECK_DIM(1, valid);
+  CHECK_DEVICE(valid, probs);
   CHECK_MAYBE_INPUT_TYPES(maybe_indices, dl_int32, dl_int64);
   CHECK_MAYBE_SAME_DTYPE(maybe_indices, output);
   validate_seed_offset_tensors(maybe_seed_arr, maybe_offset_arr, probs);
@@ -119,6 +123,7 @@ void sampling_from_probs(TensorView probs, TensorView output, Optional<TensorVie
   DISPATCH_DLPACK_IDTYPE_TO_CTYPE(output.dtype(), IdType, [&] {
     cudaError_t status = sampling::SamplingFromProb<float, IdType>(
         static_cast<float*>(probs.data_ptr()), static_cast<IdType*>(output.data_ptr()),
+        static_cast<bool*>(valid.data_ptr()),
         maybe_indices.has_value() ? static_cast<IdType*>(maybe_indices.value().data_ptr())
                                   : nullptr,
         batch_size, vocab_size, deterministic,
@@ -134,7 +139,7 @@ void sampling_from_probs(TensorView probs, TensorView output, Optional<TensorVie
   });
 }
 
-void top_p_sampling_from_probs(TensorView probs, TensorView output,
+void top_p_sampling_from_probs(TensorView probs, TensorView output, TensorView valid,
                                Optional<TensorView> maybe_indices,
                                Optional<TensorView> maybe_top_p_arr, double top_p_val,
                                bool deterministic, Optional<TensorView> maybe_seed_arr,
@@ -142,6 +147,9 @@ void top_p_sampling_from_probs(TensorView probs, TensorView output,
                                uint64_t offset_val) {
   CHECK_INPUT(probs);
   CHECK_DIM(2, probs);  // probs: (batch_size, vocab_size)
+  CHECK_INPUT(valid);
+  CHECK_DIM(1, valid);
+  CHECK_DEVICE(valid, probs);
   CHECK_MAYBE_INPUT_TYPES(maybe_indices, dl_int32, dl_int64);
   CHECK_MAYBE_SAME_DTYPE(maybe_indices, output);
   validate_seed_offset_tensors(maybe_seed_arr, maybe_offset_arr, probs);
@@ -157,6 +165,7 @@ void top_p_sampling_from_probs(TensorView probs, TensorView output,
   DISPATCH_DLPACK_IDTYPE_TO_CTYPE(output.dtype(), IdType, [&] {
     cudaError_t status = sampling::TopPSamplingFromProb<float, IdType>(
         static_cast<float*>(probs.data_ptr()), static_cast<IdType*>(output.data_ptr()),
+        static_cast<bool*>(valid.data_ptr()),
         maybe_indices.has_value() ? static_cast<IdType*>(maybe_indices.value().data_ptr())
                                   : nullptr,
         has_top_p_arr ? static_cast<float*>(maybe_top_p_arr.value().data_ptr()) : nullptr,
@@ -173,7 +182,7 @@ void top_p_sampling_from_probs(TensorView probs, TensorView output,
   });
 }
 
-void top_k_sampling_from_probs(TensorView probs, TensorView output,
+void top_k_sampling_from_probs(TensorView probs, TensorView output, TensorView valid,
                                Optional<TensorView> maybe_indices,
                                Optional<TensorView> maybe_top_k_arr, int64_t top_k_val,
                                bool deterministic, Optional<TensorView> maybe_seed_arr,
@@ -184,6 +193,9 @@ void top_k_sampling_from_probs(TensorView probs, TensorView output,
   CHECK_DEVICE(output, probs);
   CHECK_DIM(2, probs);   // probs: (batch_size, vocab_size)
   CHECK_DIM(1, output);  // output: (batch_size)
+  CHECK_INPUT(valid);
+  CHECK_DIM(1, valid);
+  CHECK_DEVICE(valid, probs);
   CHECK_MAYBE_INPUT_TYPES(maybe_indices, dl_int32, dl_int64);
   CHECK_MAYBE_SAME_DTYPE(maybe_indices, output);
   validate_seed_offset_tensors(maybe_seed_arr, maybe_offset_arr, probs);
@@ -199,6 +211,7 @@ void top_k_sampling_from_probs(TensorView probs, TensorView output,
   DISPATCH_DLPACK_IDTYPE_TO_CTYPE(output.dtype(), IdType, [&] {
     cudaError_t status = sampling::TopKSamplingFromProb<float, IdType>(
         static_cast<float*>(probs.data_ptr()), static_cast<IdType*>(output.data_ptr()),
+        static_cast<bool*>(valid.data_ptr()),
         maybe_indices.has_value() ? static_cast<IdType*>(maybe_indices.value().data_ptr())
                                   : nullptr,
         has_top_k_arr ? static_cast<float*>(maybe_top_k_arr.value().data_ptr()) : nullptr,
@@ -215,7 +228,7 @@ void top_k_sampling_from_probs(TensorView probs, TensorView output,
   });
 }
 
-void min_p_sampling_from_probs(TensorView probs, TensorView output,
+void min_p_sampling_from_probs(TensorView probs, TensorView output, TensorView valid,
                                Optional<TensorView> maybe_indices,
                                Optional<TensorView> maybe_min_p_arr, double min_p_val,
                                bool deterministic, Optional<TensorView> maybe_seed_arr,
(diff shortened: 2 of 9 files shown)


nvpohanh:
I think this feature itself will be very helpful for frameworks, but I do not like them being env vars (we already have tons of env vars). If possible, could we make it a formal Python interface? Something like:
```
flashinfer.set_autotuner_config_save_path("/path/to/my/dir")
flashinfer.set_autotuner_config_load_path("/path/to/my/dir")
```
or you can design it using the autotuner class.

also, please document it clearly about the multi-thread/multi-process support. For example, can two process write to the same path in parellel?

nvpohanh:
> Since the autotuner only works when
> ```
> with flashinfer.autotune():
>     ...
> ```
> why don't we just
> ```
> with flashinfer.autotune(cache="/path/to/cache"):
>    ...
> ```
> ?

great suggestion by @nvjullin

bkryu:
> > Since the autotuner only works when
> > ```
> > with flashinfer.autotune():
> >     ...
> > ```
> > why don't we just
> > ```
> > with flashinfer.autotune(cache="/path/to/cache"):
> >    ...
> > ```
> > ?
> great suggestion by @nvjullin

@nvpohanh Agree that this is a much cleaner approach. I have a working version now in this PR(description updated as well). Currently looking reconsidering the multi-thread and multi-process safe design.

docs/autotuning.rst
```diff
@@ -0,0 +1,392 @@
...
+is logged if mismatches are detected (e.g. different GPU, FlashInfer version,
+or cuBLAS version).
+
+Tactics are typically integers, but some runners use compound tactics (e.g.
+``(tile_size, gemm1_tactic, gemm2_tactic)``).  These are serialized as nested
+JSON arrays and restored to tuples on load.
+
+The file is human-readable and portable.  Because it stores runner class names
+(not positional indices), it is robust to changes in runner ordering across
+FlashInfer versions.
```
bkryu: Needs to be updated -- portability guarantee will be far too difficult to enforce and the current design actually did not implement this, but I forgot to update.

docs/autotuning.rst
```diff
@@ -0,0 +1,392 @@
...
+
+The file contains two special metadata keys:
+
+- ``_metadata``: Records the environment that **originally created** the cache
+  file.  This is preserved across subsequent incremental saves, so users can
+  always trace provenance back to the initial tuning session.
+- ``_last_updated_by``: Records the environment that **last wrote** to the
+  cache file.  Updated on every save.
+
+On load, ``_metadata`` is compared against the current environment. A warning
```
nvjullin: Caches being silently wrong is an incredible pain to debug, so I suggest that we err on the side of caution and either:
1. be 100% certain the cache is valid and use it
2. error out or ignore the cache, explaining why the cache may not be valid

Even if developers are diligent and read all warnings (I don't 😂), end users of frameworks definitely will not. It's better to be always correct than always fast.

bkryu: This is valuable feedback and I agree with erring on the side of caution. Will address

bkryu: Addressed this as:

> * `_metadata`: Records the environment that created the cache file (FlashInfer version, CUDA, cuBLAS, cuDNN, GPU). On load, it is compared against the current environment. **If any field differs, the entire cache is skipped** no configs are loaded, and the file will not be overwritten on exit; i.e. the autotuner would behave as if the cache file input was not provided (cache=None). This prevents silently using invalid tactics and avoids destroying configs tuned for a different environment.  A warning is logged with the mismatch details and a suggestion to use a different cache path for the current environment.

nvpohanh:
@nvjullin Could you review this again? Thanks!

docs/autotuning.rst
```diff
@@ -0,0 +1,400 @@
...
+        with open(path) as f:
+            merged.update(json.load(f))
+
+    with open("configs_merged.json", "w") as f:
+        json.dump(merged, f, indent=2, sort_keys=True)
+
+.. warning::
+
+   Atomic file writes rely on ``os.replace()``, which is atomic on local
+   filesystems (ext4, XFS, APFS, NTFS).  On **NFS** mounts, ``os.replace()``
```
nvjullin: I believe even NFS and lustre are atomic (but I'm no expert on linux filesystems). See https://stackoverflow.com/a/7054851 for example. Flashinfer cubin downloading also rely on atomicity of replace.

flashinfer/autotuner.py
```diff
@@ -844,6 +1091,184 @@ def _prepare_input_tensors(
...
+            original_metadata = disk_configs.pop(_METADATA_KEY, None)
+            disk_configs.update(configs)
+            configs = disk_configs
+        except (FileNotFoundError, json.JSONDecodeError):
+            pass  # file doesn't exist yet or is being replaced -- proceed with what we have
+
+        # Atomic write: write to a temp file then replace the target.
+        # This prevents readers from seeing a partially-written file and
+        # guards against data loss if the process is killed mid-write.
+        dir_name = os.path.dirname(abs_path)
```
nvjullin, line 1191: Perhaps comment here on a subtlety: for replace to be atomic, source and destination must be on same filesystem.

bkryu: Thanks @nvjullin. Like you said for `os.replace()` (POSIX rename()) to be atomic, source and destination must be on the same filesystem. We already ensure this — `tempfile.mkstemp(dir=dir_name)` creates the temp file in the same directory as the target, so the rename is always a same-filesystem operation. I've added a comment in the code making this explicit.

docs/autotuning.rst
bkryu: You're right that the warning is a bit too overly cautious and not entirely correct. 

I've softened the warning to a note note and ton'e down the message. Like you mentioned, NFS and Lustre both implement `rename()` atomically.

tests/autotuner/test_autotuner_configs.py
danisereb, line 1: Maybe the test can be moved to `tests/autotuner/` ?

That path (and AutoTuner unit tests) was recently created in this PR:
https://github.com/flashinfer-ai/flashinfer/pull/2617

bkryu: Thanks @danisereb, this is something I missed. Moved the test file in https://github.com/flashinfer-ai/flashinfer/pull/2554/commits/04566f52a278eb346cb47b115d7f69870889d900

tests/autotuner/test_autotuner_configs.py
```diff
@@ -0,0 +1,893 @@
...
+
+        threads = [threading.Thread(target=run_autotune) for _ in range(5)]
+        for t in threads:
+            t.start()
+        for t in threads:
+            t.join()
+
+        # After all threads complete, tuning mode should be restored to False
+        assert not tuner.is_tuning_mode, "is_tuning_mode was not properly restored"
+        assert len(errors) == 0, f"Errors in threads: {errors}"
```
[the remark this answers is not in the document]
bkryu: Good catch — the interleaving you described is a real race. The save/restore pattern works for nested single-threaded contexts but not for concurrent overlapping ones with different exit orders. I've removed the flaky assertion (the test still validates no crashes) and documented the "last-exiter wins" semantics in both the test and the code. The practical impact is low since GPU profiling is already serialized by _lock regardless of the mode flag. Reference counting could be added later if strict semantics are needed.

flashinfer/autotuner.py
```diff
@@ -254,19 +390,59 @@ def __hash__(self):
...
+        autotune_enabled = tune_mode and not old_mode
     if autotune_enabled:
         logger.info("[Autotuner]: Autotuning process starts ...")
     try:
         yield
     finally:
-        AutoTuner.get().is_tuning_mode = old_mode
+        with tuner._lock:
+            tuner.is_tuning_mode = old_mode
         if autotune_enabled:
```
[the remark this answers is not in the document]
bkryu: Great suggestion — implemented the reference counting approach you proposed. _active_tuning_contexts tracks the number of active autotune(True) contexts, and is_tuning_mode is derived from _active_tuning_contexts > 0. This means:
- An autotune(False) context can no longer clobber an active autotune(True) on another thread
- Concurrent autotune(True) contexts exit cleanly regardless of order — the last one to exit drops the counter to 0
- The thread safety test now asserts both is_tuning_mode == False and _active_tuning_contexts == 0 after all threads complete

flashinfer/autotuner.py
```diff
@@ -16,13 +19,146 @@
...
+    if isinstance(tactic, (tuple, list)):
+        return [_tactic_to_json(v) for v in tactic]
+    # Handle foreign iterable types (e.g. TVM FFI Array<int64_t>) that are
+    # not plain tuple/list but still support iteration.
+    if hasattr(tactic, "__iter__") and not isinstance(tactic, (str, bytes, dict)):
+        return [_tactic_to_json(v) for v in tactic]
+    # Coerce numpy / pybind int types to plain Python int for JSON safety.
+    if isinstance(tactic, int):
+        return int(tactic)
+    return tactic
```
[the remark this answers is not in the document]
bkryu: Good catch — bool being a subclass of int in Python is a classic gotcha. Fixed by checking isinstance(tactic, bool) before the int check so booleans are preserved through the round-trip.

flashinfer/autotuner.py
```diff
@@ -844,6 +1091,184 @@ def _prepare_input_tensors(
...
+                json.dump(ordered, f, indent=2)
+            os.replace(tmp_path, abs_path)
+        except BaseException:
+            with contextlib.suppress(OSError):
+                os.unlink(tmp_path)
+            raise
+
+        with self._lock:
+            self._dirty = False
+
```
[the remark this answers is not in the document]
bkryu: Valid concern — the unlocked I/O window between snapshot and _dirty = False could lose new results. However, holding the lock for the entire I/O (including _collect_metadata() ctypes probing, disk re-read, and atomic write) would block all threads during save — including inference threads calling search_cache() — causing latency spikes.

Instead, I've implemented a generation counter (_dirty_seq) that increments each time a new profiling result is added. save_configs snapshots the counter before releasing the lock, and only clears _dirty if the counter hasn't advanced by the time I/O completes. If new results arrived during the save, _dirty stays True and the next autotune() exit will save them. This avoids the race without holding the lock during I/O.

flashinfer/autotuner.py
```diff
@@ -254,19 +390,59 @@ def __hash__(self):
...
+    tuner = AutoTuner.get()
+
+    # Load configs from cache file on entry (if it exists).
+    # cache_valid is False when the file exists but has a metadata mismatch;
+    # in that case we skip saving on exit to avoid overwriting configs from
+    # a different environment.
+    cache_valid = True
+    if cache is not None and os.path.isfile(cache):
+        cache_valid = tuner.load_configs(cache)
+
```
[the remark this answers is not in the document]
bkryu: Good catch. Fixed by clearing _file_configs and _logged_file_hits whenever a cache path is specified, before attempting to load. This ensures only configs from the specified file are active, even if a different cache was loaded in a previous autotune() block.

nv-yunzheq (approved):

aleozlx:
bot run tests look good

flashinfer/autotuner.py
```diff
@@ -16,13 +19,148 @@
...
+    meta: Dict[str, str] = {}
+    meta["flashinfer_version"] = _flashinfer_version
+    meta["cuda_version"] = getattr(torch.version, "cuda", None) or "unknown"
+    meta["cublas_version"] = _get_cublas_version()
+    try:
+        meta["cudnn_version"] = str(torch.backends.cudnn.version())
+    except Exception:
+        meta["cudnn_version"] = "unknown"
+    try:
+        meta["gpu"] = torch.cuda.get_device_name(0)
```
wenscarl: How about `meta["gpu"] = torch.cuda.get_device_name(torch.cuda.current_device())`?

flashinfer/autotuner.py
```diff
@@ -844,6 +1111,189 @@ def _prepare_input_tensors(
...
+                runner_id, tactic, _opt_profile = cache_value
+
+                # Use hash-free key: (custom_op, runner_class_name, profile)
+                file_key = str((custom_op, runner_class_name, profile))
+
+                # Store runner class name (not positional index) for robustness
+                tactic_json = _tactic_to_json(tactic)
+                configs[file_key] = [runner_class_name, tactic_json]
+
+        num_new = len(configs) - num_previous
```
wenscarl: num_new in the save log message is misleading when disk has extra entries? May move it after the disk merge?

wenscarl (approved):

flashinfer/autotuner.py
bkryu: Good catch — fixed in latest commit. `torch.cuda.current_device()` correctly reflects the active device

flashinfer/autotuner.py
bkryu: Thanks for this and I agree how this can be misleading. Moved num_new to after the disk merge so the log message accurately reflects how many entries were added to the final file, including any extra entries from other sessions that got merged in.

PR #2560, opened 2026-02-13
fix: guard CUTLASS FMHA against SM12x and fix fmha_v2 SM121a check

blake-snc:
## Summary

- **Remove SM12x from CUTLASS FMHA support**: `get_fmha_module()` and `gen_fmha_cutlass_sm100a_module()` incorrectly included SM12x GPUs (RTX 5090, DGX Spark) in their support checks. SM12x lacks the `tcgen05` MMA instructions required by the CUTLASS FMHA SM100 kernel (`SM100_MMA_F16BF16_SS/TS`, `SM100_MMA_F8F6F4_SS/TS`), causing compile failures when using `backend="cutlass"` or `fmha_varlen()`. Changed `supported_major_versions` from `[10, 11, 12]` to `[10, 11]` and added a clear error message for SM12x users pointing them to `backend='fa2'`.

- **Fix `fmha_v2_prefill_deepseek` SM121a check**: The SM12x guard only checked `is_sm120a_supported()` (SM120 = RTX 5090, minor=0) but not `is_sm121a_supported()` (SM121 = DGX Spark, minor=1). DGX Spark users were incorrectly rejected from using the fmha_v2 DeepSeek prefill kernel.

## Validated on NVIDIA GB10 (DGX Spark, SM 12.1)

| Test | Result |
|------|--------|
| CUTLASS FMHA correctly rejects SM12x with clear error | PASS |
| FA2 prefill works (max_diff=0.0078 vs SDPA reference) | PASS |
| XQA decode works (no NaN) | PASS |
| `determine_attention_backend()` returns "fa2" for SM12x | PASS |
| `fmha_v2_prefill_deepseek` accepts SM121a | PASS |

## Test plan

- [ ] Verify CUTLASS FMHA still works on SM100a (B200/GB200)
- [ ] Verify `fmha_varlen()` raises clear error on SM12x instead of compile failure
- [ ] Verify `fmha_v2_prefill_deepseek()` works on both SM120 (RTX 5090) and SM121 (DGX Spark)
- [ ] Run existing CI tests

Contributed by [Second Nature Computing](https://joinsecondnature.com)

## Summary by CodeRabbit

* **Bug Fixes**
  * FMHA optimized kernel now targets only SM100a/SM110a devices; other devices will receive an updated compatibility message with a suggested alternative backend.
  * Removed CUDA 12+ compilation support for the optimized path.
  * Prefill behavior updated: the alternate prefill path will not proceed on non-SM12x hardware and now raises a clear, explicit message.

### DIFF
--- a/flashinfer/prefill.py
+++ b/flashinfer/prefill.py
@@ -53,7 +53,6 @@ from .utils import (
     canonicalize_torch_dtype,
     determine_attention_backend,
     device_support_pdl,
-    get_compute_capability,
     get_device_sm_count,
     is_float8,
     is_sm100a_supported,


blake-snc:
Regarding the suggestion to use `get_device_capability()[0] != 12` instead of `is_sm120a_supported() or is_sm121a_supported()`: the `is_sm12xa_supported()` utility functions are the standard pattern used throughout flashinfer (they also check CUDA toolkit version requirements). Using a raw capability check here would be inconsistent with the rest of the codebase.

That said, if a future SM122a variant appears, adding `is_sm122a_supported()` to the check is a one-line change. Happy to refactor if the maintainers prefer a different approach.

eugr:
> That said, if a future SM122a variant appears, adding `is_sm122a_supported()` to the check is a one-line change. Happy to refactor if the maintainers prefer a different approach.

@blake-snc - would it be better to introduce is_sm12x_family_supported() to cover all such cases? Because even though it's a one line change, there are still some places in the code where sm120 is included and sm121 is ignored, even though they are pretty much identical. Even in new-ish PRs. Point in case: https://github.com/flashinfer-ai/flashinfer/pull/2460

And that's with sm121 being out in the wild since October.

blake-snc:
@eugr Good call. I just opened #2574 which adds `is_sm12x_supported()` to `utils.py` using a `major == 12` check (matching the pattern of `is_sm100a_supported`). This way future SM12x variants are automatically covered without needing to add and wire up individual `is_sm122a_supported()` functions.

The PR updates the existing call sites in `gemm_base.py` and the DeepSeek MLA test. Once it lands, our other SM120 PRs (#2559, #2560, #2561) can be rebased to use it too.

yzh119:
@flashinfer-bot run

blake-snc:
The internal CI pipeline shows 9/20 passed — are the failures related to this PR or pre-existing? Our changes only narrow CUTLASS FMHA support (removing SM12x) and fix the SM121a check in `fmha_v2_prefill_deepseek`, so they shouldn't affect existing SM100/SM110 tests.

Happy to investigate if there's something specific we need to fix.

blake-snc:
Hey @yzh119 — this PR now has merge conflicts with main. Here's what changed upstream since your approval:

1. **CUTLASS FMHA**: Main now includes SM12x in `get_fmha_module()` with `is_sm12x_supported()` and `supported_major_versions=[10, 11, 12]`. Our PR removes SM12x from that path (since SM12x lacks tcgen05 MMA). These are conflicting approaches — should SM12x be in the CUTLASS FMHA path or not?

2. **fmha_v2 DeepSeek SM121a**: Main now uses `is_sm12x_supported()` which covers all SM12x variants, so our SM121a-specific fix is already addressed.

Happy to rebase if you can clarify the intended direction for (1). If CUTLASS FMHA genuinely works on SM12x with newer CUDA, we can drop that part and just rebase cleanly.

blake-snc:
Update: resolved the merge conflicts and rebased onto main. PR is mergeable now.

The resolution keeps our original intent:
- CUTLASS FMHA guard: SM12x removed from `get_fmha_module()` and `supported_major_versions` (SM12x lacks tcgen05)
- fmha_v2 DeepSeek: adopted upstream's `is_sm12x_supported()` which is more general than our original per-variant check

PR #2576, opened 2026-02-18
Remove compute capability restrictions from routerGemm and fused_topk_deepseek

yzh119:
## Summary
- Remove `@supported_compute_capability` and `@backend_requirement` decorators from `routerGemm` (`mm_M1_16_K7168_N128`, `mm_M1_16_K7168_N256`) and `fused_topk_deepseek` APIs
- Both kernels use standard CUDA operations with SM90+ PDL features guarded by `#if __CUDA_ARCH__ >= 900`, so they work on all GPU architectures
- `routerGemm` was previously restricted to SM100 only; `fused_topk_deepseek` was restricted to SM89/90/100/103/120/121
- Shape/config validation is now called directly in the function body instead of via decorator
- Router GEMM tests no longer skip on non-SM100 GPUs

## Test plan
- [ ] Run `pytest tests/model_optimizations/test_dsv3_router_gemm.py` on non-SM100 GPU (e.g. A100, H100)
- [ ] Run `pytest tests/model_optimizations/test_dsv3_fused_routing.py` on non-SM100 GPU
- [ ] Verify negative tests still raise ValueError for invalid inputs

## Summary by CodeRabbit

* **Refactor**
  * Restructured internal validation logic for DSv3 routing and GEMM kernels, moving validation checks to explicit runtime calls instead of decorator-based checks while maintaining the same functional behavior.

* **Tests**
  * Updated test execution to run unconditionally on all CUDA devices, removing hardware-specific conditional skipping.

### DIFF
--- a/flashinfer/fused_moe/fused_routing_dsv3.py
+++ b/flashinfer/fused_moe/fused_routing_dsv3.py
@@ -5,12 +5,9 @@ from types import SimpleNamespace
 import torch
 from flashinfer.utils import (
     register_custom_op,
-    supported_compute_capability,
-    backend_requirement,
 )
 
 
-@supported_compute_capability([89, 90, 100, 103, 120, 121])
 def _check_dsv3_fused_routing_supported(
     scores,
     bias,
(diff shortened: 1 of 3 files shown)


aleozlx (approved):
lgtm

PR #2574, opened 2026-02-17
feat: add is_sm12x_supported() helper for SM12x family detection

blake-snc:
## Summary

Adds `is_sm12x_supported()` to `flashinfer/utils.py` as a convenience helper that covers the entire SM12x GPU family (SM120a, SM121a, and future variants like SM122a) without requiring callers to enumerate each minor version.

Uses a `major == 12` check, matching the existing pattern of `is_sm100a_supported()` (`major == 10`). This means future SM12x variants are automatically covered without code changes.

**Motivation:** SM121a (DGX Spark) keeps getting missed when only SM120a is checked. This was noted by @eugr in #2560, and PR #2460 is another example where SM121a was not included alongside SM120a.

## Changes

| File | Change |
|------|--------|
| `flashinfer/utils.py` | Add `is_sm12x_supported()` with `major == 12` check |
| `flashinfer/gemm/gemm_base.py` | Replace 3 instances of `is_sm120a_supported(a.device) or is_sm121a_supported(a.device)` |
| `tests/attention/test_fmha_v2_prefill_deepseek.py` | Update skip guard to use `is_sm12x_supported()` |

The individual `is_sm120a_supported()` and `is_sm121a_supported()` functions are preserved for cases that need variant-specific behavior.

Validated on DGX Spark (SM121a, CUDA 13.0).

[Second Nature Computing](https://joinsecondnature.com)

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

## Summary by CodeRabbit

* **Refactor**
  * Consolidated separate SM120/SM121 capability checks into a unified SM12x check and updated the public import surface accordingly.
  * Introduced explicit CUDA-version gating for SM12x variants and clarified related compatibility/error messages.

* **Tests**
  * Updated GPU compatibility tests and skip logic/messages to target SM12x architecture support.

PR #2561, opened 2026-02-13
feat: SM120 standard attention kernels + CuTe DSL backend

blake-snc:
## Summary

Adds three attention paths for **SM120 GPUs** (RTX 5090, DGX Spark GB10) that lack `tcgen05` instructions:

- **fmha_v2 standard attention kernels** for head_dim={64, 128} with BF16 and FP16 (FP32 accumulator), using Ampere HMMA (`sm_mma=80`) with `SEPARATE_Q_K_V` layout
- **CuTe DSL flash attention backend** (`cute_dsl_prefill_sm120`) wrapping CUTLASS's `FlashAttentionForwardSm120` with JIT compilation and kernel caching
- **SM12x routing fixes**: CUTLASS FMHA SM100 no longer incorrectly routes SM12x (would crash on `tcgen05`), `fmha_v2_prefill_deepseek` now accepts SM121a

### Changes

| File | Change |
|------|--------|
| `flashinfer/jit/attention/fmha_v2/generate_kernels.py` | Add standard attention kernel enumeration for SM120 (bf16 + fp16_fp32, hd=64/128) |
| `csrc/trtllm_fmha_v2_binding.cu` | Add `TRTLLMFMHAv2StandardRun` binding for standard attention dispatch |
| `flashinfer/jit/attention/modules.py` | Add SM120 standard attention kernel files; fix SM100 module to exclude SM12x |
| `flashinfer/prefill.py` | Add `fmha_v2_prefill_sm120()` API; fix SM12x routing in CUTLASS FMHA and DeepSeek paths |
| `flashinfer/cute_dsl_attention.py` | New: CuTe DSL flash attention wrapper with alignment-aware tensor conversion |

### Validation on DGX Spark (SM 12.1)

| Test | Result |
|------|--------|
| fmha_v2 BF16 hd=128 causal (MHA) | PASS (max_diff=0.003906) |
| fmha_v2 BF16 hd=64 causal | PASS (max_diff=0.003906) |
| fmha_v2 FP16 hd=128 non-causal | PASS (max_diff=0.000488) |
| fmha_v2 GQA 32:8 BF16 hd=128 | PASS |
| CuTe DSL BF16 hd=128 causal | PASS (max_diff=0.001953) |
| CuTe DSL FP16 hd=64 non-causal | PASS (max_diff=0.000244) |
| fmha_v2 DeepSeek MLA BF16 | PASS |

### Context

[...]

PR #2446, opened 2026-01-30
feat: Add TRTLLM fmha_v2 library for SM90 attention with Skip-Softmax 

jimmyzho:
## 📌 Description

Integrates TRT-LLM's Fused Multi-Head Attention v2 (FMHAv2) kernels into FlashInfer as a prefill attention backend via a new trtllm_fmha_v2_prefill Python API, backed by a CUDA runtime dispatch layer (fmha_v2_run.cu) that bridges FlashInfer's tensor conventions with TRT-LLM's fused kernel infrastructure.

- Data types: FP16, BF16, FP8 (E4M3) inputs with configurable output dtype (e.g., FP8 → BF16/FP16). FP32 accumulation for BF16/E4M3; fused BMM1/BMM2 scaling for quantized inference.
- Input layouts: Packed QKV [T, 3, H, D], Contiguous Q+KV [T, H, D]+[T, 2, H_kv, D], Separate Q/K/V, and Q + Paged KV Cache (page sizes 32/128, HND/NHD layouts). Block tables auto-expanded from [B, M] → [B, 2, M] for separate K/V offset addressing.
- Masking: Causal, sliding window (window_left), chunked causal (power-of-2 sizes).
- Ragged batching: Variable-length sequences via cumulative sequence length tensors.
- GQA/MQA: Grouped-query and multi-query attention (h % h_kv == 0).
- Attention sinks: Per-head sink values in softmax denominator.
- Logits soft capping: Fused tanh-based softcapping with configurable scale.
- ALiBi: Optional ALiBi positional encoding.
- Softmax stats (LSE) export: Optional save_softmax_stats returns [max, sum_exp] per token/head.
- Non-blocking execution

Also adds skip-softmax optimization introduced in [TRT-LLM's Documentation](https://github.com/NVIDIA/TensorRT-LLM/blob/main/docs/source/blogs/tech_blog/blog16_Accelerating_Long_Context_Inference_with_Skip_Softmax_Attention.md#accelerating-long-context-inference-with-skip-softmax-attention)

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [ ] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [ ] I have installed the hooks with `pre-commit install`.
- [ ] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

[...]

PR #2598, opened 2026-02-20
feat: add CuTe DSL flash attention backend for SM120 GPUs

blake-snc:
## Summary

Adds an optional attention backend using NVIDIA's CuTe DSL (CUTLASS Python DSL) for **SM120 GPUs** (RTX 5090, DGX Spark GB10) that lack `tcgen05` instructions. The kernel uses SM80-compatible HMMA tensor core instructions (`mma.sync.aligned.m16n8k16`) which work on SM120.

This is extracted from #2561 as a standalone module, decoupled from the fmha_v2 standard attention work (which was integrated into #2446 per discussion with @jimmyzho).

### API

```python
from flashinfer.cute_dsl_attention import cute_dsl_prefill_sm120

out = cute_dsl_prefill_sm120(q, k, v, causal=True)
```

### Requirements

- `nvidia-cutlass-dsl` package (`pip install nvidia-cutlass-dsl`)
- `cuda-python` package (`pip install cuda-python`)
- CUTLASS repo with `FlashAttentionForwardSm120` example (set `CUTLASS_FA_SM120_PATH` or clone to `~/cutlass/`)

### Review feedback addressed (from #2561)

| Issue | Fix |
|-------|-----|
| Unbounded `_compile_cache` | Bounded to 64 entries with FIFO eviction |
| `num_threads` missing from cache key | Included in cache key |
| Hardcoded dev-specific paths | Removed; uses env var + standard locations only |
| Missing `@flashinfer_api` | Added |
| Missing `out` parameter | Added optional `out` parameter |
| `torch.Tensor.dim_order()` compat | Fallback for PyTorch < 2.7 via stride-derived order |
| `cuda-python` import guard | Protected with try-except; supports both cuda-python < 12.6.2 and >= 12.6.2 import paths |
| Missing input validation | Full q/k/v shape, dtype, device validation |
| `@functools.cache` unbounded | Changed to `@functools.lru_cache(maxsize=32)` |

### Validation on DGX Spark (SM121a, CUDA 13.0)

| Test | Result |
|------|--------|
| BF16 b=1 h=4 s=64 d=64 non-causal | PASS (max_diff=0.000000) |
| BF16 b=1 h=4 s=64 d=64 causal | PASS (max_diff=0.000000) |
| BF16 b=1 h=4 s=128 d=128 non-causal | PASS (max_diff=0.000977) |
| BF16 b=1 h=4 s=128 d=128 causal | PASS (max_diff=0.000001) |
| BF16 b=2 h=4 s=64 d=64 non-causal | PASS (max_diff=0.000977) |
| BF16 b=2 h=4 s=64 d=64 causal | PASS (max_diff=0.000000) |
| FP16 b=1 h=4 s=64 d=64 non-causal | PASS (max_diff=0.000122) |
| FP16 b=1 h=4 s=64 d=64 causal | PASS (max_diff=0.000244) |
| FP16 b=1 h=4 s=128 d=128 causal | PASS (max_diff=0.000244) |
| Pre-allocated `out` parameter | PASS |

### Dependencies

- This PR uses `is_sm120a_supported`/`is_sm121a_supported` individually; once #2574 merges, these can be replaced with the unified `is_sm12x_supported()` helper.

[Second Nature Computing](https://joinsecondnature.com)

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

## Summary by CodeRabbit

* **New Features**
  * CuTe DSL–based flash-attention prefill support for SM120/SM121 GPUs with JIT-compiled, cached kernels and concurrency-safe compilation.
  * Automatic backend selection to prefer CuTe DSL when compatible, with transparent fallback to alternate prefill paths for unsupported cases (grouped-query attention, CUDA graphs, differing per-sequence lengths or head mismatches).
  * Automatic tensor layout/format handling, configurable tuning knobs, robust input validation, and clear dependency checks.

### DIFF
--- a/flashinfer/prefill.py
+++ b/flashinfer/prefill.py
@@ -3312,7 +3312,6 @@ class BatchPrefillWithRaggedKVCacheWrapper:
                 q.device,
                 "out",
             )
-<<<<<<< HEAD
         if self._backend == "cute-dsl":
             # These checks live here (not in plan()) because return_lse and
             # scale parameters are run()-time arguments that can vary between
--- a/flashinfer/prefill.py
+++ b/flashinfer/prefill.py
@@ -3080,13 +3080,25 @@ class BatchPrefillWithRaggedKVCacheWrapper:
                 logits_soft_cap > 0,  # use_logits_soft_cap
                 use_fp16_qk_reduction,
             )
+            # cute_dsl limitations: no GQA, no CUDA graphs, and seqlen_q must equal
+            # seqlen_k per sequence (full prefill only, not append/chunked).
+            if self._backend == "cute_dsl" and (
+                num_qo_heads != num_kv_heads
+                or self.is_cuda_graph_enabled
+                or not torch.equal(
+                    qo_indptr_host[1:] - qo_indptr_host[:-1],
+                    kv_indptr_host[1:] - kv_indptr_host[:-1],
+                )
+            ):
+                self._backend = "fa2"
+
             if self._backend == "cutlass":
                 # insert qo_indptr.device to 9th position (0-indexed) of get_module_args
                 new_get_module_args = (
                     get_module_args[:9] + (qo_indptr.device,) + get_module_args[9:]
                 )
                 self._cached_module = get_fmha_module(*new_get_module_args)
-            elif self._backend != "cudnn":
+            elif self._backend not in ("cudnn", "cute_dsl"):
                 self._cached_module = get_batch_prefill_module(
                     self._backend, *get_module_args
                 )
@@ -3096,7 +3108,7 @@ class BatchPrefillWithRaggedKVCacheWrapper:
                 self._cached_module, qo_indptr, kv_indptr, num_qo_heads, causal
             )
             self._max_qo_len = torch.max(qo_indptr[1:] - qo_indptr[:-1]).item()
-        elif self._backend not in ("cudnn", "cute-dsl"):
+        elif self._backend not in ("cudnn", "cute_dsl"):
             assert self._cached_module is not None, "cached module is not initialized"
             args = [
                 self._float_workspace_buffer,
@@ -3300,6 +3312,7 @@ class BatchPrefillWithRaggedKVCacheWrapper:
                 q.device,
                 "out",
             )
+<<<<<<< HEAD
         if self._backend == "cute-dsl":
             # These checks live here (not in plan()) because return_lse and
             # scale parameters are run()-time arguments that can vary between
--- a/flashinfer/prefill.py
+++ b/flashinfer/prefill.py
@@ -54,6 +54,7 @@ from .utils import (
     _get_cache_buf,
     _unpack_paged_kv_cache,
     canonicalize_torch_dtype,
+    _should_use_cute_dsl,
     determine_attention_backend,
     device_support_pdl,
     get_device_sm_count,
@@ -3067,6 +3068,17 @@ class BatchPrefillWithRaggedKVCacheWrapper:
                     q_data_type,
                     kv_data_type,
                 )
+                # cute_dsl is only supported by BatchPrefill, so check here
+                # rather than in determine_attention_backend (which is shared
+                # by single_prefill, decode, sparse, etc.)
+                if self._backend == "fa2" and _should_use_cute_dsl(
+                    self.device,
+                    PosEncodingMode[pos_encoding_mode].value,
+                    self._custom_mask_buf is not None,
+                    q_data_type,
+                    kv_data_type,
+                ):
+                    self._backend = "cute_dsl"
 
             get_module_args = (
                 q_data_type,
--- a/flashinfer/prefill.py
+++ b/flashinfer/prefill.py
@@ -3092,6 +3092,11 @@ class BatchPrefillWithRaggedKVCacheWrapper:
             ):
                 self._backend = "fa2"
 
+            if self._backend == "cute_dsl":
+                # Cache CPU indptrs to avoid GPU→CPU sync on every run() call
+                self._qo_indptr_cpu = qo_indptr_host
+                self._kv_indptr_cpu = kv_indptr_host
+
             if self._backend == "cutlass":
                 # insert qo_indptr.device to 9th position (0-indexed) of get_module_args
                 new_get_module_args = (
--- a/flashinfer/prefill.py
+++ b/flashinfer/prefill.py
@@ -3078,7 +3078,7 @@ class BatchPrefillWithRaggedKVCacheWrapper:
                     q_data_type,
                     kv_data_type,
                 ):
-                    self._backend = "cute_dsl"
+                    self._backend = "cute-dsl"
 
             get_module_args = (
                 q_data_type,
@@ -3094,7 +3094,7 @@ class BatchPrefillWithRaggedKVCacheWrapper:
             )
             # cute_dsl limitations: no GQA, no CUDA graphs, and seqlen_q must equal
             # seqlen_k per sequence (full prefill only, not append/chunked).
-            if self._backend == "cute_dsl" and (
+            if self._backend == "cute-dsl" and (
                 num_qo_heads != num_kv_heads
                 or self.is_cuda_graph_enabled
                 or not torch.equal(
@@ -3104,7 +3104,7 @@ class BatchPrefillWithRaggedKVCacheWrapper:
             ):
                 self._backend = "fa2"
 
-            if self._backend == "cute_dsl":
+            if self._backend == "cute-dsl":
                 # Cache CPU indptrs to avoid GPU→CPU sync on every run() call
                 self._qo_indptr_cpu = qo_indptr_host
                 self._kv_indptr_cpu = kv_indptr_host
@@ -3115,7 +3115,7 @@ class BatchPrefillWithRaggedKVCacheWrapper:
                     get_module_args[:9] + (qo_indptr.device,) + get_module_args[9:]
                 )
                 self._cached_module = get_fmha_module(*new_get_module_args)
-            elif self._backend not in ("cudnn", "cute_dsl"):
+            elif self._backend not in ("cudnn", "cute-dsl"):
                 self._cached_module = get_batch_prefill_module(
                     self._backend, *get_module_args
                 )
@@ -3125,7 +3125,7 @@ class BatchPrefillWithRaggedKVCacheWrapper:
                 self._cached_module, qo_indptr, kv_indptr, num_qo_heads, causal
             )
             self._max_qo_len = torch.max(qo_indptr[1:] - qo_indptr[:-1]).item()
-        elif self._backend not in ("cudnn", "cute_dsl"):
+        elif self._backend not in ("cudnn", "cute-dsl"):
             assert self._cached_module is not None, "cached module is not initialized"
             args = [
                 self._float_workspace_buffer,
--- a/flashinfer/prefill.py
+++ b/flashinfer/prefill.py
@@ -3068,7 +3068,7 @@ class BatchPrefillWithRaggedKVCacheWrapper:
                     q_data_type,
                     kv_data_type,
                 )
-                # cute_dsl is only supported by BatchPrefill, so check here
+                # cute-dsl is only supported by BatchPrefill, so check here
                 # rather than in determine_attention_backend (which is shared
                 # by single_prefill, decode, sparse, etc.)
                 if self._backend == "fa2" and _should_use_cute_dsl(
@@ -3092,7 +3092,7 @@ class BatchPrefillWithRaggedKVCacheWrapper:
                 logits_soft_cap > 0,  # use_logits_soft_cap
                 use_fp16_qk_reduction,
             )
-            # cute_dsl limitations: no GQA, no CUDA graphs, and seqlen_q must equal
+            # cute-dsl limitations: no GQA, no CUDA graphs, and seqlen_q must equal
             # seqlen_k per sequence (full prefill only, not append/chunked).
             if self._backend == "cute-dsl" and (
                 num_qo_heads != num_kv_heads
--- a/flashinfer/utils.py
+++ b/flashinfer/utils.py
@@ -464,6 +464,17 @@ def is_cutlass_backend_supported(
     return True
 
 
+@functools.lru_cache(maxsize=1)
+def _is_cute_dsl_available() -> bool:
+    """Check if nvidia-cutlass-dsl is available for CuTe DSL backends."""
+    try:
+        import cutlass.cute  # noqa: F401
+
+        return True
+    except ImportError:
+        return False
+
+
 def determine_attention_backend(
     device: torch.device,
     pos_encoding_mode: int,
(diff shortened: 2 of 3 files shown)


blake-snc:
Added `determine_attention_backend()` routing in commit 648a155a.

SM120a/SM121a devices now automatically select the `cute_dsl` backend when `nvidia-cutlass-dsl` is installed. The check gates on:
- `is_sm120a_supported(device) or is_sm121a_supported(device)`
- `PosEncodingMode.NONE` (no RoPE/ALiBi)
- no custom mask
- FP16 or BF16 dtypes
- `cutlass.cute` importable (`_is_cute_dsl_available()`, LRU-cached)

Falls back to `fa2` if any condition fails (including if nvidia-cutlass-dsl isn't installed), so existing users are unaffected. Routing and backend land atomically in this PR.

blake-snc:
**Update on `determine_attention_backend()` routing (reverted, 04d0f767)**

I attempted to add SM12x routing in this PR (commit 648a155a) but reverted it immediately after discovering it breaks the existing pipeline.

**Root cause**: `determine_attention_backend()` is consumed by `BatchPrefillWithRaggedKVCacheWrapper.plan()` which unconditionally calls `get_batch_prefill_module(backend, ...)`. That function asserts `backend in ["fa2", "fa3"]` — it has no knowledge of `"cute_dsl"`.

The `cute_dsl` backend is architecturally different from fa2/fa3: it uses the CUTLASS Python DSL rather than the TVM-FFI JIT module path. Routing through `determine_attention_backend()` requires first adding special-case handling in `prefill.py`'s `plan()` and `run()` methods (following the existing `"cudnn"` and `"trtllm-gen"` patterns).

**Remaining integration work** (separate from what's currently in this PR):
1. In `BatchPrefillWithRaggedKVCacheWrapper.plan()`: skip `get_batch_prefill_module()` when backend is `"cute_dsl"`
2. In `BatchPrefillWithRaggedKVCacheWrapper.run()`: add a branch that calls `cute_dsl_prefill_sm120()` directly (analogous to the cudnn dispatch at end of `run()`)
3. Same treatment for `BatchPrefillWithPagedKVCacheWrapper`
4. Then re-add `determine_attention_backend()` routing

Leaving it out for now to keep the PR focused on the backend implementation itself.

blake-snc:
Follow-up to the revert above — the full integration is now in commit c3c70784.

**What changed:**

`utils.py` — `_is_cute_dsl_available()` + routing in `determine_attention_backend()`: SM120a/SM121a now selects `cute_dsl` when `nvidia-cutlass-dsl` is installed, falls back to `fa2` otherwise.

`prefill.py` — `BatchPrefillWithRaggedKVCacheWrapper`:
- `plan()`: skips JIT module loading and plan step for `cute_dsl`; downgrades to `fa2` when GQA is active, CUDA graphs are enabled, or `seqlen_q != seqlen_k` per sequence (the kernel requires full prefill, not append/chunked)
- `run()`: converts ragged flat tensors to per-sequence dense format, calls `cute_dsl_prefill_sm120()`, reassembles output

**Validated on SM121a (DGX Spark GB10):**
| Config | Backend | max_diff vs fa2 |
|---|---|---|
| non-GQA, causal=True | cute_dsl | 0.0156 (1 BF16 ULP) |
| non-GQA, causal=False | cute_dsl | 0.0039 |
| GQA (32H/8KV) | fa2 (fallback) | 0.0 |
| Sq≠Sk (append) | fa2 (fallback) | 0.0 |
| head_dim=64 | cute_dsl | 0.0156 |
| batch=1 | cute_dsl | 0.0078 |

blake-snc:
Friendly ping — CI is green and this has been open for a couple weeks. Would love to get a review when someone has bandwidth. Thanks!

flashinfer/utils.py
```diff
@@ -499,6 +510,15 @@ def determine_attention_backend(
...
         return "fa3"
+    elif (
+        (is_sm120a_supported(device) or is_sm121a_supported(device))
+        and not use_custom_mask
+        and pos_encoding_mode == PosEncodingMode.NONE.value
+        and dtype_q in {torch.float16, torch.bfloat16}
+        and dtype_kv in {torch.float16, torch.bfloat16}
+        and _is_cute_dsl_available()
+    ):
+        return "cute_dsl"
```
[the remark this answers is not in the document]
saltyminty: Seems this change was unreverted – what ended up being the reasoning for this, and is this an actual issue?

flashinfer/prefill.py
```diff
@@ -3071,7 +3083,26 @@ def run(
...
+            for i in range(batch_size):
+                q_i = q[qo_indptr_cpu[i] : qo_indptr_cpu[i + 1]].unsqueeze(0)
+                k_i = k[kv_indptr_cpu[i] : kv_indptr_cpu[i + 1]].unsqueeze(0)
+                v_i = v[kv_indptr_cpu[i] : kv_indptr_cpu[i + 1]].unsqueeze(0)
+                out_i = cute_dsl_prefill_sm120(
+                    q_i, k_i, v_i, causal=self._causal, sm_scale=sm_scale
+                ).squeeze(0)
+                out_slices.append(out_i)
+            out.copy_(torch.cat(out_slices, dim=0))
+            return (out, lse) if return_lse else out
```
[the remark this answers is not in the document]
saltyminty: +1

saltyminty (commented):
Largely looks good, just wanted a follow up on some of the coderabbit comments

flashinfer/utils.py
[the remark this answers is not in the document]
blake-snc: The revert was of the initial attempt that broke the pipeline (it tried to return `"cute_dsl"` from `determine_attention_backend()` but `get_batch_prefill_module()` only accepts `"fa2"`/`"fa3"`). The re-addition in c3c70784 includes the full integration — `plan()` skips JIT module loading for cute_dsl and `run()` dispatches directly to `cute_dsl_prefill_sm120()`. So the routing change in utils.py is intentional and the pipeline now handles it correctly end-to-end.

flashinfer/prefill.py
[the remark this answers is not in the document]
blake-snc: Good catch — I'll add a `NotImplementedError` guard for `return_lse=True` in the cute_dsl branch. The underlying CuTe DSL kernel doesn't compute LSE currently.

blake-snc:
@saltyminty thanks for the review — addressed everything in the last two commits:

- **`return_lse` guard** (25e78869): raises `NotImplementedError` when `return_lse=True` with cute_dsl since the kernel doesn't compute LSE
- **head_dim consistency check** (66346165): validates `key.shape[-1]` and `value.shape[-1]` match query
- **stride order in cache key** (66346165): prevents wrong compiled kernel reuse for different memory layouts
- **GPU→CPU sync fix** (66346165): indptrs are now cached on CPU during `plan()` instead of calling `.cpu()` every `run()`, and writes directly into output slices instead of cat + copy

The `utils.py` routing question is answered inline — the revert was of the first broken attempt, the current code includes the full `plan()`/`run()` integration.

blake-snc:
@saltyminty — all the review feedback has been addressed (NotImplementedError for return_lse in 25e78869, head_dim consistency check, and the other items in 66346165). CI is green. Would you be able to take another look?

flashinfer/prefill.py
```diff
@@ -3071,7 +3088,32 @@ def run(
                 q.device,
                 "out",
             )
-        if self._backend == "cutlass":
+        if self._backend == "cute_dsl":
+            from .cute_dsl_attention import cute_dsl_prefill_sm120
+
+            if return_lse:
```
saltyminty, line 3336: Should this be checked as part of the fallback to fa2 on 2871?

flashinfer/utils.py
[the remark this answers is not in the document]
saltyminty: Does the above CodeRabbit comment still need to be addressed? I don't believe the other APIs have proper support for cutedsl backend.

flashinfer/prefill.py
blake-snc: Good catch — the broader issue was that `determine_attention_backend()` was returning `"cute_dsl"` globally when only `BatchPrefill` handles it. Fixed in 111612e — moved the cute_dsl selection out of `determine_attention_backend()` into `BatchPrefill.plan()` via a separate `_should_use_cute_dsl()` helper. Now `single_prefill`, decode, sparse all correctly get `"fa2"` on SM12x. The fallback from cute_dsl to fa2 for unsupported configs (GQA, CUDA graphs, append/chunked) is still in place.

flashinfer/utils.py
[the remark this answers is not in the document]
blake-snc: You're right, this was a real issue — `determine_attention_backend()` returning `"cute_dsl"` would break any API path that passes the backend string to JIT module generators since they only handle `"fa2"` and `"fa3"`. Fixed in the same commit by scoping cute_dsl selection to `BatchPrefill.plan()` only. `determine_attention_backend()` now returns `"fa2"` for SM12x unconditionally.

blake-snc:
Related: opened #3016 to enable fmha_v2 HMMA kernels as an SM12x backend for when nvidia-cutlass-dsl isn't installed. The two are complementary — cute_dsl (this PR) takes priority when available, fmha_v2 (#3016) serves as the optimized fallback for MHA S>=256.

flashinfer/prefill.py
```diff
@@ -3096,7 +3125,7 @@ def plan(
                 self._cached_module, qo_indptr, kv_indptr, num_qo_heads, causal
             )
             self._max_qo_len = torch.max(qo_indptr[1:] - qo_indptr[:-1]).item()
-        elif self._backend not in ("cudnn", "cute-dsl"):
+        elif self._backend not in ("cudnn", "cute_dsl"):
```
saltyminty: There seems to be some mixture of using "cute-dsl" and "cute_dsl" – I believe "cute-dsl" was the standard prior to this change, should we align on one?

Otherwise looks good

saltyminty (changes requested):

flashinfer/prefill.py
blake-snc: Good catch — aligned on `"cute-dsl"` in commit `d3776c82`. The 5 spots in `flashinfer/prefill.py` that had introduced `"cute_dsl"` (underscore) are now back to `"cute-dsl"` (hyphen) to match the existing convention used in `tests/utils/test_fp8_quantize.py`, `tests/utils/test_fp4_quantize_padding.py`, etc. Python identifiers (module / function / attribute names like `_cute_dsl_wrapper`, `BatchPrefillCuteDSLWrapper`) stay as `cute_dsl` since they need PEP 8 underscores.

blake-snc: Follow-up — two corrections to my prior reply:

1. I misattributed the underscore-only constraint on identifiers to PEP 8; it's actually a Python *syntax* constraint (hyphens aren't legal identifier characters). PEP 8 is the style guide (snake_case for functions/modules, CamelCase for classes) and doesn't address hyphens at all. The end behavior is the same — module/function/attribute names like `flashinfer/cute_dsl/...`, `cute_dsl_prefill_sm120`, `_cute_dsl_wrapper`, `BatchPrefillCuteDSLWrapper` are all required by the language, not by style — but the framing was off.

2. On a more careful re-scan I had missed two comments in `flashinfer/prefill.py` that informally referenced the backend with the underscore form (`# cute_dsl is only supported by BatchPrefill`, `# cute_dsl limitations: ...`). Fixed in `6917dd9b` so a reader sees consistent backend naming throughout the file, not just in the string literals. No other stragglers across the rest of the PR's comments / docstrings.

flashinfer/prefill.py
saltyminty, line 2676: This constructor path ends up getting skilled when `backend == "auto"` – would this be an issue if auto selects cutedsl later down the line?

saltyminty (commented):
Sorry for late review – left one comment

flashinfer/prefill.py
blake-snc: Following up on your `prefill.py:2676` comment, @saltyminty — good catch, and yes, it's a real issue. Ran it down on DGX Spark (sm_121a) today.

It's actually worse than the constructor being skipped: `backend="auto"` could upgrade to cute-dsl *after* the cute-dsl plan branch had already been skipped, so `_cute_dsl_wrapper` stays `None` from `__init__` and `run()` crashes outright. I have a fix ready (resolve auto before the plan branch, construct the wrapper lazily, share one path with explicit `backend="cute-dsl"`) plus regression tests — happy to push it if useful.

While I was in there I hit something bigger worth raising, though: after the rebase over main's modular CuTe DSL FMHA package (#2805), the cute-dsl path this PR wires up is the tcgen05/TMEM mainloop — which only runs on datacenter Blackwell (sm_100/101/103). On sm12x both the explicit and auto cute-dsl paths now fail at `plan()` with `OpError: expects arch to be one of [sm_100a, ..., sm_103f]`, and the original sm120 mma.sync kernel module this PR added (`cute_dsl_attention.py`) is no longer wired to anything. So as it stands the routing sends consumer Blackwell to a kernel that can't run there.

Given that, #3016 (fmha_v2 HMMA for SM120 standard shapes, already approved) may be the cleaner path — it covers sm12x prefill natively and could supersede this PR. If you'd rather keep this open as the tracking issue for a native sm12x (mma.sync m16n8k16) mainloop in the modular cute_dsl package, I'm glad to take that follow-up on against the new structure. Whatever you think makes the most sense.

For the record on the other threads: `utils.py:521` — cute-dsl selection is kept out of the shared `determine_attention_backend` (lives in `_should_use_cute_dsl`, BatchPrefill-only); `prefill.py:3128` — user-facing strings standardized on "cute-dsl" (module path stays `cute_dsl`); `prefill.py:3104`/`:3336` — the run()-time `NotImplementedError` guards are there because `return_lse`/scales are run-time args that vary per call.

blake-snc:
@saltyminty was there something remaining here? Trying to catch up on stale open source work after a heavy workload the last month.

geraldstanje1:
hi @mmangkad - could you kindly reply to the question above?

blake-snc:
Thanks for the ping @geraldstanje1, and @saltyminty for the earlier reviews.

Circling back after a while: in the months this PR has been open, main has independently added a full CuTe DSL attention backend for BatchPrefill under `flashinfer/cute_dsl/attention/` (`BatchPrefillCuteDSLWrapper`, `ALiBiAttention`, `SoftCappingAttention`, and the plan/run wiring in `prefill.py`). That supersedes the bulk of this PR. The standalone `cute_dsl_attention.py` here is no longer needed, since `prefill.py` now uses the packaged wrapper from main.

The only piece not already in main is the automatic sm120/sm121 selection of a specialized backend. That is the same goal as #3016, which adds fmha_v2 HMMA auto-routing for sm12x and is the path being taken to keep these devices off generic FA2. Between main's cute_dsl backend and #3016's routing, what this PR set out to add is now covered.

So I'll close this as superseded. If a cute_dsl-specific auto-route is ever wanted alongside fmha_v2, I'm happy to open a small focused PR for just that, gated on nvidia-cutlass-dsl availability. Thanks all.

PR #2602, opened 2026-02-20
fix: get tensors by const ref to not rely on deleted move constructor for `TensorView`

jdebache:
## 📌 Description

Getting a `Tensor` our of an `Array` as a `TensorView` attempts to call the deleted move-constructor `TensorView(Tensor&&)`. We can instead get a const ref to the tensors out of the array.

Error messages before these changes:
```
/workspace/flashinfer/csrc/fused_moe/cutlass_backend/flashinfer_cutlass_fused_moe_binding.cu(980): error: function "tvm::ffi::TensorView::TensorView(tvm::ffi::Tensor &&)" (declared at line 717 of /workspace/venv/lib/python3.12/site-packages/tvm_ffi/include/tvm/ffi/container/tensor.h) cannot be referenced -- it is a deleted function
TensorView fc1_global = quant_scales.value()[1];
^
```

No errors after the fix.

## Summary by CodeRabbit

* **Refactor**
  * Internal optimization to quantization handling in the fused mixture of experts module for improved code efficiency.

### DIFF
--- a/csrc/fused_moe/cutlass_backend/flashinfer_cutlass_fused_moe_binding.cu
+++ b/csrc/fused_moe/cutlass_backend/flashinfer_cutlass_fused_moe_binding.cu
@@ -976,10 +976,10 @@ class FusedMoeRunner : public tvm::ffi::ModuleObj {
       TVM_FFI_ICHECK_EQ(quant_scales.value().size(), 4)
       "Expecting 4 quant scales for W4A8_MXFP4_MXFP8 quantization";
 
-      TensorView fc1_weight_block = quant_scales.value()[0];
-      TensorView fc1_global = quant_scales.value()[1];
-      TensorView fc2_weight_block = quant_scales.value()[2];
-      TensorView fc2_global = quant_scales.value()[3];
+      auto const& fc1_weight_block = quant_scales.value()[0];
+      auto const& fc1_global = quant_scales.value()[1];
+      auto const& fc2_weight_block = quant_scales.value()[2];
+      auto const& fc2_global = quant_scales.value()[3];
 
       // The input for scale fc1_weight_block / fc2_weight_block is packed into INT32
       constexpr int FP8_PER_INT32 = 4;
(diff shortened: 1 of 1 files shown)


jdebache:
> Do we need to update other modules?

I think usages like: https://github.com/flashinfer-ai/flashinfer/blob/main/csrc/gdn_prefill_launcher.cu#L137 are fine, since the underlying type is `TensorView`, so it shouldn't attempt to move?

Other than that, I just went through every reference to `TensorView` and couldn't find any other similar usage.

jdebache:
Looking at the test failures, e.g.
```
FAILED tests/gdn/test_decode_delta_rule.py::test_gdn_decode_klast_bf16_state_kernel[bfloat16-32-16-16-32-128-3-auto-True-True] - AssertionError: State mismatch for gdn_decode_klast_bf16_state kernel (B=32, T=3)
FAILED tests/gdn/test_decode_delta_rule.py::test_gdn_decode_klast_bf16_state_kernel[bfloat16-64-16-16-32-128-3-auto-True-True] - AssertionError: State mismatch for gdn_decode_klast_bf16_state kernel (B=64, T=3)
```

Seems unrelated to these changes? I'm not sure what `gdn` is.

Also this timeout:
```
2026-02-20T19:01:26.268403Z 01O tests/utils/test_logits_processor.py ................................... [  5%]
2026-02-20T19:01:45.112068Z 01O ........................................................................ [ 17%]
2026-02-20T19:07:08.200551Z 01O ............................s..s..s..................................... [ 29%]
2026-02-20T19:09:32.223400Z 00O section_end:1771614572:step_script
2026-02-20T19:09:32.223500Z 00O+[0K[0;33mWARNING: step_script could not run to completion because the timeout was exceeded. For more control over job and script timeouts see: https://docs.gitlab.com/ee/ci/runners/configure_runners.html#set-script-and-after_script-timeouts[0;m
2026-02-20T19:09:33.659520Z 00O [31;1mERROR: Job failed: execution took longer than 4h0m0s seconds
```

Issue #2687, opened 2026-03-04
[RFC] Unified GDN Decode/Prefill API


kaixih:
# GDN Unified API Design Plan

## 0. Status Snapshot (2026-05-13)

This issue is still open and is in the **FlashInfer Roadmap** project with status **In Progress**. It has also become more of an umbrella tracker for GDN work over time: most of the important kernel-side features have landed or have dedicated follow-up PRs, while the unified API work itself is still open.

**Completed / mostly landed:**
- Decode / MTP side: BF16 MTP + intermediate-state support (#2679), FP32 VK decode pool/index support (#2727), BF16 decode negative-index handling (#2810), and separate input/output pool indices for decode (#2905).
- Prefill infrastructure: SM90 checkpointing infrastructure (#2908) and SM100 / Blackwell prefill support (#3001).

**Still open:**
- Unified decode API: #2706 is still open but stale/conflicted (`mergeStateStatus=DIRTY`, last updated 2026-04-08). It likely needs a rebase plus another API-alignment pass before it can land.
- Prefill initial-state pool / log-gate support: #3167 is still open and conflicted (`mergeStateStatus=DIRTY`, last updated 2026-05-07). It mainly covers input-side initial-state indexing; paged checkpoint/final-state writeback is still missing.
- Remaining broader gaps: prefill BF16 state, prefill paged writeback / radix-cache reuse, unified API tests, and docs.

**Recent GDN-related activity:**
- #3312 is open (`mergeStateStatus=BLOCKED`, opened 2026-05-13): fixes Blackwell GDN chunked kernel compilation with cute-dsl 4.5.0.
- #3262 is open: `import flashinfer.gdn_decode` requires CUDA devices; #3293 is the corresponding open fix (`mergeStateStatus=BLOCKED`).
- #3298 is open: several cute-dsl kernels, including Blackwell GDN prefill, pass `storage.<scalar>` instead of `storage.<scalar>.ptr`; #3300 is the corresponding open fix (`mergeStateStatus=BLOCKED`).

## 1. Current API Landscape

FlashInfer currently exposes **four separate GDN decode entry points**, each covering a different subset of the feature matrix:

### 1.1 `gated_delta_rule_decode_pretranspose` (VK layout, primary)
- State layout: `[B, HV, V, K]` (K-last / VK)
- **BF16 state**: Yes — via BF16 fast path (requires T ∈ {1,2,3,4}, K=V=128)
[...]

PR #2706, opened 2026-03-06
feat(gdn): add unified decode API and deprecation shims (RFC 5.7, 5.8)

Dayuxiaoshui:
## 📌 Description

Implements RFC 5.7/5.8: add unified GDN decode API `gated_delta_rule_decode_unified` (state_layout, state_indices, intermediate_states_buffer, disable_state_update), rename backends to `_*_impl` with dispatch by layout/dtype/T, add deprecation shims for the three legacy decode APIs, export in __init__.py, and add tests/gdn/test_gdn_decode_unified.py (cross-check, intermediate_buffer, edge cases, error paths).
## 🔍 Related Issues

Closes https://github.com/flashinfer-ai/flashinfer/issues/2687

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * Unified decode API that selects the appropriate backend at runtime.

* **Deprecations**
  * Backend-specific decode variants deprecated; legacy shims emit warnings.

* **Improvements**
  * Stronger validation with clearer error messages for invalid state/layout/parameter combos.
  * Expanded test coverage and updated benchmarks to verify parity across implementations.


Dayuxiaoshui:
@kaixih Quick note on naming: the RFC suggests the unified entry point be called gated_delta_rule_decode, but that name is already used by the existing KV-layout API. To avoid breaking callers, this PR introduces the unified API as gated_delta_rule_decode_unified and keeps the old names as deprecation shims. If you prefer to switch to gated_delta_rule_decode for the unified API (and e.g. rename the current KV-layout one or deprecate it under a different name), I can follow up with a small rename patch. Please advise.

kaixih:
On naming: we'd recommend the following approach rather than keeping _unified in the name (which is an implementation detail callers shouldn't see):                                                
                                                                                                                                                                                                      
  1. Rename the existing gated_delta_rule_decode (KV layout) → gated_delta_rule_decode_kv: this is descriptive, clearly signals it's the legacy KV-layout-specific path, and the function is not currently exported from flashinfer/__init__.py so the blast radius is small.                                                                                                                        
  2. Use gated_delta_rule_decode for the unified API: the clean name with no suffix, as the RFC originally proposed.                                                                                 
  3. Keep the old gated_delta_rule_decode_pretranspose and gated_delta_rule_mtp as deprecation shims (as you've already done), and add gated_delta_rule_decode_kv as a shim for the renamed KV function.

@kahyunnam @yzh119 how do you think?

Dayuxiaoshui:
@kaixih I agree with this viewpoint, and I will make corrections and resubmit.

Dayuxiaoshui:
@kaixih Thanks, done. We renamed the KV path to gated_delta_rule_decode_kv, use gated_delta_rule_decode for the unified API, and kept the three shims (pretranspose, kv, mtp) as you suggested.

Dayuxiaoshui:
cc @kaixih

flashinfer/gdn_decode.py
```diff
@@ -625,3 +627,391 @@ def gated_delta_rule_mtp(
...
+def _check_state_indices_bounds(state_indices: torch.Tensor, pool_size: int) -> None:
+    """Validate that all state_indices are in [0, pool_size). Raises ValueError if not."""
+    if state_indices.numel() == 0:
+        return
+    bad = (state_indices < 0) | (state_indices >= pool_size)
+    if bad.any().item():
+        first_bad = state_indices[bad].flatten()[0].item()
+        raise ValueError(
+            f"state_indices must be in [0, pool_size={pool_size}); got out-of-range value {first_bad}"
+        )
```
kaixih, line 708: Do we need this bounds check? Two concerns: (1) the range validation adds overhead that may not be acceptable on the hot path — we generally rely on the framework/caller to ensure indices are valid. (2) More importantly, negative values are intentional: they represent padding (dummy) entries during CUDA graph capture and should be supported, not rejected.

flashinfer/gdn_decode.py
```diff
@@ -625,3 +627,391 @@ def gated_delta_rule_mtp(
...
+
+
+@flashinfer_api
+def gated_delta_rule_decode(
```
kaixih, line 712: maybe adding @backend_requirement for the sm90+ guard.

flashinfer/gdn_decode.py
```diff
@@ -625,3 +627,391 @@ def gated_delta_rule_mtp(
...
+            )
+
+    # state_layout == "VK", float32
+    if state.dtype != torch.float32:
+        raise ValueError(
+            f"VK layout supports bfloat16 or float32 state, got {state.dtype}"
+        )
+    if T == 1:
+        if use_pool:
+            raise NotImplementedError(
```
kaixih, line 884: can we double check if this is supported? from existing code, I can see:
```
 run_pretranspose_decode(
        h0_source,
        A_log,
        a,
        dt_bias,
        q,
        k,
        v,
        b,
        output,
        B,
        T,
        H,
        HV,
        K,
        V,
        scale,
        use_qk_l2norm,
        use_pool_indexing=use_pool_indexing,
        initial_state_indices=initial_state_indices,
    )
```

Dayuxiaoshui: Will route to run_pretranspose_decode with pool and remove the NotImplementedError.

flashinfer/gdn_decode.py
Dayuxiaoshui: Makes sense ，dropped the Python bounds check so negative indices stay as padding for the kernel, and noted that in the docstring.

Dayuxiaoshui:
cc @kaixih

PR #2710, opened 2026-03-06
fix(mm_fp4): use per-call workspace buffer to prevent race condition

### LINKED ISSUE #2708: SM120 CUTLASS FP4 GEMM: missing GDC compile flags cause PDL race condition — output tile corruption under concurrency
## Summary

All CUTLASS GEMM templates in FlashInfer use `enablePDL=true` (Programmatic Dependent Launch), but the JIT compilation is missing `-DCUTLASS_ENABLE_GDC_FOR_SM100=1` and `-DCUTLASS_ENABLE_GDC_FOR_SM90=1` compile flags. Without these flags, `wait_on_dependent_grids()` and `launch_dependent_grids()` in CUTLASS `grid_dependency_control.h` compile as **empty no-ops**, eliminating the synchronization barriers needed for safe PDL execution.

This causes intermittent output tile corruption (NaN/garbage in 128-aligned blocks) under high concurrency on SM120 (Blackwell) and potentially SM100/SM103.

## Root Cause

In `cutlass/include/cutlass/arch/grid_dependency_control.h`:

```cpp
CUTLASS_DEVICE void wait_on_dependent_grids() {
#if (defined(CUTLASS_GDC_ENABLED))
  asm volatile("griddepcontrol.wait;");
#endif
}
```

`CUTLASS_GDC_ENABLED` is only defined when `CUTLASS_ENABLE_GDC_FOR_SM100` is passed as a compile flag. In `flashinfer/jit/gemm/core.py`, only the TGV GEMM module passes this flag — all FP4, FP8, and MXFP8 GEMM modules are missing it.

Result: PDL enables kernel overlap at the host level, but the device-side barriers are compiled out. Dependent kernels read stale global memory before the previous kernel completes writeback.

## Affected modules

- `fp4_gemm_cutlass` (SM100)
- `fp4_gemm_cutlass_sm103` (SM103)
- `fp4_gemm_cutlass_sm120` (SM120)
- `fp8_gemm_cutlass` (SM100)
- `mxfp8_gemm_cutlass` (SM100)
- `gemm_sm120` (SM120 FP8 groupwise)

## Symptoms

- Output tiles not written — contiguous 128-aligned blocks matching CTA tile boundaries
- `CUDA_LAUNCH_BLOCKING=1` eliminates the bug (confirms race condition)
- Retry with identical inputs produces correct output
- Non-NaN elements are bit-exact identical to cudnn (`max_diff=0.0`)
- Only under high concurrency (64+ requests), never standalone `mm_fp4()`
- cudnn backend is always correct

## Fix

PR #2716 — adds both GDC compile flags to all affected GEMM JIT modules.

## Environment

- FlashInfer 0.6.4
- Hardware: 8x NVIDIA RTX PRO 6000 Blackwell (SM120, 96GB)
- SGLang with TP=8, EAGLE-v2, GLM-5-NVFP4-MTP model
- PyTorch 2.12.0.dev, CUDA 12.8+

## Related

- Fix PR: #2716
- SGLang issue: https://github.com/sgl-project/sglang/issues/20043
- SGLang workaround PR: https://github.com/sgl-project/sglang/pull/20047

voipmonitor:
## Summary

- Replace the shared cached workspace buffer (`_get_cache_buf`) in `mm_fp4()` with a fresh `torch.empty()` allocation per call to prevent race conditions when multiple `mm_fp4()` calls execute concurrently.

## Root Cause

`_get_cache_buf` returns a globally cached singleton buffer per device. When multiple `mm_fp4()` calls run concurrently during inference (e.g., in tensor-parallel or pipeline-parallel setups), they share the same workspace buffer, causing CUTLASS kernel corruption that manifests as NaN values in 128-aligned output tiles.

## Fix

Replace:
```python
workspace_buffer = _get_cache_buf(
    "mm_fp4_workspace", DEFAULT_WORKSPACE_SIZE, a.device
)
```

With:
```python
workspace_buffer = torch.empty(DEFAULT_WORKSPACE_SIZE, dtype=torch.uint8, device=a.device)
```

Fresh allocation has zero overhead thanks to PyTorch's CUDA allocator caching — previously freed buffers of the same size are returned instantly from the free list without any `cudaMalloc`.

Fixes #2708

## Test plan

- [ ] Verify NaN values no longer appear in concurrent `mm_fp4()` calls
- [ ] Verify no performance regression (PyTorch CUDA allocator caching makes fresh allocation effectively free)

## Summary by CodeRabbit

**Refactor**
- Updated workspace buffer allocation strategy for matrix operations to use direct allocation instead of cached buffers, improving memory predictability and lifecycle management during tensor computations.

### DIFF
--- a/flashinfer/gemm/gemm_base.py
+++ b/flashinfer/gemm/gemm_base.py
@@ -3854,9 +3854,7 @@ def mm_fp4(
             dtype=out_dtype,
         )
 
-    workspace_buffer = _get_cache_buf(
-        "mm_fp4_workspace", DEFAULT_WORKSPACE_SIZE, a.device
-    )
+    workspace_buffer = torch.empty(DEFAULT_WORKSPACE_SIZE, dtype=torch.uint8, device=a.device)
 
     # Auto-select the best backend
     if backend == "auto":


voipmonitor:
Closing — shared workspace is NOT the root cause. On clean image: workspace fix still crashes, only torch.zeros on output buffer prevents NaN. The CUTLASS SM120 kernel genuinely skips writing output tiles. Investigation continues in #2708.

voipmonitor:
This PR was based on an incorrect hypothesis (shared workspace race condition). The actual root cause was identified as missing GDC compile flags for PDL synchronization.

See #2708 and the fix in #2716.

PR #2716, opened 2026-03-07
fix(jit): GEMM kernels produce NaN under concurrency — missing GDC flags cause PDL synchronization barriers to compile as no-ops

voipmonitor:
## Summary

All CUTLASS GEMM templates use `enablePDL=true` (Programmatic Dependent Launch), but the JIT compilation is missing `-DCUTLASS_ENABLE_GDC_FOR_SM100=1` and `-DCUTLASS_ENABLE_GDC_FOR_SM90=1` compile flags. Without these flags, `wait_on_dependent_grids()` and `launch_dependent_grids()` in CUTLASS `grid_dependency_control.h` compile as **empty no-ops**, eliminating the synchronization barriers needed for safe PDL execution.

## Root Cause

In `cutlass/include/cutlass/arch/grid_dependency_control.h`:

```cpp
CUTLASS_DEVICE void wait_on_dependent_grids() {
#if (defined(CUTLASS_GDC_ENABLED))  // only defined when CUTLASS_ENABLE_GDC_FOR_SM100 is set
  asm volatile("griddepcontrol.wait;");
#endif
}
```

The `CUTLASS_GDC_ENABLED` macro is only defined when `CUTLASS_ENABLE_GDC_FOR_SM100` is passed as a compile flag. Without it, PDL launches kernels with overlap enabled at the host level (`cudaLaunchAttributeProgrammaticStreamSerialization`), but the device-side synchronization barriers are compiled out — creating a race condition.

## Symptoms

On SM120 (Blackwell RTX PRO 6000 / RTX 5090) with high concurrency (64+ simultaneous requests in SGLang with TP=8):
- CUTLASS FP4 GEMM intermittently fails to write output tiles
- Unwritten tiles contain uninitialized memory (NaN/garbage)
- NaN blocks are always contiguous and 128-aligned, matching CTA tile boundaries
- `CUDA_LAUNCH_BLOCKING=1` eliminates the bug (confirms race condition)
- cudnn backend is unaffected (does not use CUTLASS PDL)
- Retry with identical inputs produces correct output

## Fix

Add `-DCUTLASS_ENABLE_GDC_FOR_SM100=1` and `-DCUTLASS_ENABLE_GDC_FOR_SM90=1` to all affected GEMM JIT modules:
- `fp4_gemm_cutlass` (SM100)
- `fp4_gemm_cutlass_sm103` (SM103)
- `fp4_gemm_cutlass_sm120` (SM120)
- `fp8_gemm_cutlass` (SM100)
[...]

PR #2720, opened 2026-03-07
fix: replace griddepcontrol inline PTX with CUDA runtime API

### LINKED ISSUE #2558: griddepcontrol.wait should use "memory" clobber
Someone at Nvidia recently made me aware of this code: https://github.com/flashinfer-ai/flashinfer/blob/292f9be3f5f6d76248d4d3577c167fd178d7952d/include/flashinfer/norm.cuh#L53

There are two issues with it:
1. There is no need to use `asm` for this, as the wrappers `cudaTriggerProgrammaticLaunchCompletion` and `cudaGridDependencySynchronize` have existed essentially since the same toolkit which supports the `griddepcontrol` PTX statements.
2. If the wrappers shouldn't be used for some reason, it's important to make `griddepcontrol.wait` use the `"memory"` clobber, anything else is undefined behavior and might lead to incorrect compiler optimization.

A quick search reveals that this UB issue appears in a few other places, e.g. https://github.com/flashinfer-ai/flashinfer/blob/292f9be3f5f6d76248d4d3577c167fd178d7952d/include/flashinfer/activation.cuh#L37 as well.

It would be great to check the code in general for this kind of issue, not just with `griddepcontrol`.
Also, quick plug for https://github.com/NVIDIA/cccl/blob/v3.4.0.dev/libcudacxx/include/cuda/ptx : 
for many instructions, you can just 

```
#include <cuda/ptx>
namespace ptx = ::cuda::ptx;
...
ptx::cp_async_bulk(...)
```

It is part of the CUDA toolkit

he-yufeng:
Fixes #2558 - replaces asm volatile griddepcontrol with cudaGridDependencySynchronize/cudaTriggerProgrammaticLaunchCompletion across 19 files

## Summary by CodeRabbit

* **Refactor**
  * Updated kernel synchronization operations across fused MOE, attention cascade, GEMM, normalization, sampling, and related compute kernels to support CUDA 12+ and Blackwell GPUs (SM 100).

### DIFF
--- a/csrc/fused_moe/cutlass_backend/cutlass_fused_moe_kernels.cuh
+++ b/csrc/fused_moe/cutlass_backend/cutlass_fused_moe_kernels.cuh
@@ -214,7 +214,7 @@ __global__ void buildMinLatencyActiveExpertMapsKernel(
     int const start_expert, int const end_expert, int const num_experts_per_node,
     bool const smart_routing, int const cluster_rank, int const cluster_size,
     int const num_experts_smem) {
-#if (defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 900))
+#if (__CUDACC_VER_MAJOR__ >= 12 && defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 900))
   cudaGridDependencySynchronize();
 #endif
   // Use one block to process the min latency case
@@ -307,7 +307,7 @@ __global__ void buildMinLatencyActiveExpertMapsKernel(
       expert_first_token_offset[i_exp] = num_active * num_tokens;
     }
   }
-#if (defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 900))
+#if (__CUDACC_VER_MAJOR__ >= 12 && defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 900))
   cudaTriggerProgrammaticLaunchCompletion();
 #endif
 }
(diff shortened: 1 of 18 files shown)


he-yufeng:
### Details

This replaces all 79 occurrences of inline PTX `griddepcontrol` assembly with proper CUDA runtime API wrappers across 19 files:

- `asm volatile("griddepcontrol.wait;")` -> `cudaGridDependencySynchronize()`
- `asm volatile("griddepcontrol.launch_dependents;")` -> `cudaTriggerProgrammaticLaunchCompletion()`

**Why**: The inline PTX lacked a `"memory"` clobber, which is UB. The compiler can freely reorder global memory accesses across the barrier, potentially breaking PDL synchronization. The CUDA runtime wrappers include proper compiler barriers and have been available since CUDA 12.0.

**Scope**: Covers all C/C++ files on current main. The earlier `claude/issue-2558-20260216-0413` branch only touched 14 files and is now 71 commits behind, missing new files in `csrc/fused_moe/cutlass_backend/` and `csrc/xqa/`.

**Not covered**:
- CUTLASS wrapper callsites in `tgv_gemm.cuh` (uses `cutlass::arch::wait_on_dependent_grids()`, correctness depends on compile flags addressed in #2716)
- Python DSL files using MLIR `llvm.inline_asm` (different mechanism, may need separate fix)

he-yufeng:
Ping @yzh119 — any thoughts on this? It's been open a couple weeks. The inline PTX UB was reported by an NVIDIA engineer in #2558 so figured it's worth addressing.

he-yufeng:
Rebased this onto current `main` and addressed the review nits that were still relevant after the rebase:

- kept the CUDA runtime PDL calls behind an explicit CUDA 12 + SM90 guard, matching the pattern used elsewhere in the tree
- removed the duplicate synchronize/completion calls in `trtllm_moe_allreduce_fusion.cuh`
- left the remaining non-PDL SM90 guard in `xqa/utils.cuh` alone because it still protects an unrelated inline asm helper

Validation from my side:

- `git diff --check upstream/main..HEAD`
- duplicate-call inspection in the touched allreduce fusion kernel

I could not run CUDA compilation locally because this Windows machine does not have `nvcc` installed.

he-yufeng:
The current red `Test Results Summary` is an authorization gate rather than a CUDA build/test failure. The job reports:

```text
CI skipped (pending authorization)
A contributor in @flashinfer-ai/ci-users can comment `@flashinfer-bot run` to approve.
```

Pre-commit, docs, Claude review, label, and permission-check jobs are green. Could someone with CI permission trigger the GPU/AOT jobs when convenient?

PR #2722, opened 2026-03-08
feat(prefill): add Block Extend attention for diffusion LLMs

fdz-1999:
### What is Diffusion LLM (DLLM)?

Diffusion LLM (DLLM) is an emerging text generation paradigm. Unlike traditional Auto-Regressive LLMs that generate one token at a time, DLLM generates multiple tokens **in parallel at the block level**. In each iteration, all tokens within the current block are produced simultaneously through multi-step denoising. As a result, tokens within the same block require **bidirectional visibility**, while tokens in subsequent blocks must be completely invisible — this is the semantic origin of the Block Extend Mask.

### Similarity Between Block Diffusion and Chunked Prefill

The execution flow of Block Diffusion closely resembles SGLang's existing **Chunked Prefill** — both split the full sequence into chunks and process them step by step:

| Computation Phase | Chunked Prefill | Block Diffusion |
|-------------------|----------------|-----------------|
| Context query (Q_curr × KV_prev) | Full Attention | Full Attention (**identical**) |
| Intra-block query (Q_curr × KV_curr) | **Causal Attention Mask** | **Full Attention (bidirectional)** |

The only difference lies in the intra-block query: Chunked Prefill uses a causal mask (earlier tokens cannot attend to later ones), whereas Block Diffusion requires **bidirectional full attention** since tokens within the same block are generated in parallel.

### Why Native Kernel Support is Needed

Chunked Prefill is SGLang's default and well-established execution path. Given the strong similarity between Block Diffusion and Chunked Prefill, the initial approach was to reuse the Chunked Prefill path — splitting by DLLM block size and using `causal=False` for the current chunk to approximate the Block Extend mask.

However, this indirect Cascade Attention-based approach has fundamental limitations:

1. **chunk_size is locked to dllm_block_size**: `causal=False` is only correct when the chunk exactly equals one complete DLLM block; larger chunk sizes that would reduce the number of iteration steps cannot be used
2. **2–3 kernel launches per step** (current chunk + prefix + merge_state), with CPU overhead growing linearly with the number of steps
3. **Unable to exploit the structural sparsity of the Block Extend mask**: invisible KV tiles are still loaded and computed

An alternative is **Custom Mask** (`MaskMode::kCustom`). A dense boolean mask for a 32K sequence requires about 1 GiB per request. Even a packed bitmask requires about 128 MiB and still scales as O(qo_len × kv_len).

Therefore, we implemented the native `MaskMode::kBlockExtend` mask mode in FlashInfer, embedding the Block Extend mask semantics directly into the CUDA kernel:

- **No additional mask tensor**: the mask is computed via integer division in registers
- **Tile-level skip**: fully invisible KV tiles are skipped at CTA granularity, avoiding their memory loads and computation
- **Single prefill operation**: avoids Cascade's explicit current-chunk, prefix-attention, and state-merge sequence
- **Flexible chunk_size**: chunk_size can be much larger than dllm_block_size, significantly reducing the number of iteration steps

## 📌 Description

Add DLLM Block Extend Attention with native `MaskMode::kBlockExtend` tile-level skip optimization for Diffusion LLM inference.

### Block Extend Mask Rule
image

mask[q, k] = floor(q_global / B) >= floor(k_global / B)

Tokens within the same DLLM block are bidirectionally visible; they can attend to previous blocks but not subsequent blocks. Here, `B` is `block_size`, `q_global = q_offset + q_index`, and `k_global = kv_offset + k_index`.

### Core API

| Layer | Public API | Use Case |
|-------|------------|----------|
| Single Prefill | `single_prefill_with_kv_cache(..., block_extend=True, block_size=..., q_offset=..., kv_offset=...)` | Single request, incremental chunk prefill |
| Batch Ragged | `BatchPrefillWithRaggedKVCacheWrapper(..., block_extend=True, block_size=...)` | Multi-request parallel, ragged KV; offsets are supplied during `plan()` |
| Batch Paged | `BatchPrefillWithPagedKVCacheWrapper(..., block_extend=True, block_size=...)` | Multi-request parallel, paged KV; offsets are supplied during `plan()` |

### Why faster than Custom Mask

| Dimension | Custom Mask | Block Extend |
|-----------|------------|--------------|
| Memory | O(qo_len × kv_len) mask tensor | No additional mask tensor |
| Invisible tiles | Requires loading and evaluating the materialized mask | Derive visibility from offsets and skip fully invisible KV tiles |
| Fully visible tiles | Load mask + element-wise check | Direct MMA without loading a mask tensor |
| Mask source | Global memory read | Register computation (integer division) |

### Why faster than Cascade Attention (SGLang-style)
image

| Dimension | Cascade | Block Extend |
|-----------|---------|--------------|
| Execution per step | Separate current-chunk, prefix-attention, and state-merge operations | One Block Extend prefill operation |
| `chunk_size` constraint | Must equal `dllm_block_size` for this formulation | Not locked to `dllm_block_size`; larger and non-aligned chunks are supported |
| Steps (tokens=256, B=32) | 256/32 = 8 steps | chunk=256 → 1 step |
| SM utilization | Small chunk → low occupancy | Larger chunk → better SM saturation |

## 🔍 Related Issues

None

### Pre-commit checks

- [ ] I have installed `pre-commit` by running `pip install pre-commit`
- [ ] I have installed the hooks with `pre-commit install`
- [ ] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues

## 🧪 Tests

- [x] Correctness: block extend vs custom_mask reference, cascade vs blockwise extend validation
- [x] Performance: FlashInfer Block Extend vs Custom Mask and SGLang Cascade benchmarks on NVIDIA H20

### Kernel Benchmark — NVIDIA H20

Benchmarks were collected with CUDA Graphs on an NVIDIA H20 using the FA2 and
FA3 backends. Results are specific to this hardware and configuration.

Here, `chunk_size` is the number of query tokens processed per incremental prefill step. It is distinct from `dllm_block_size`, which is the logical block size used by the Block Extend mask. The tables report the best-performing chunk size from each tested sweep.

The SGLang-style Cascade baseline is constrained to `chunk_size = dllm_block_size = 32`. Block Extend can use larger chunks, so the reported CUDA Graph speedup includes both kernel efficiency and the reduction in iteration count.

### Single-request incremental prefill

Configuration: 8,192 physical KV tokens, one request, and `dllm_block_size=32`. We swept `chunk_size ∈ {32, 64, 128, 256}`; the table reports the best-performing configuration (`chunk_size=256`).

| Backend | Method | Chunk size (tokens/step) | CUDA Graph time | vs. SGLang Cascade | vs. Custom Mask |
|---|---|---:|---:|---:|---:|
| FA2 | `single_prefill_with_kv_cache` | 256 | 18.432 ms | 1.69x | 1.51x |
| FA2 | `BatchPrefillWithRaggedKVCacheWrapper` | 256 | 21.024 ms | 1.48x | 1.32x |
| FA3 | `single_prefill_with_kv_cache` | 256 | 10.988 ms | 8.70x | 2.81x |
| FA3 | `BatchPrefillWithRaggedKVCacheWrapper` | 256 | 11.085 ms | 8.62x | 2.78x |

### Multi-request batch prefill

Configuration: 256 requests × 512 physical KV tokens per request and `dllm_block_size=32`. We swept `chunk_size ∈ {32, 64, 128, 256}`; the table reports the best-performing configuration (`chunk_size=128`).

| Backend | Method | Chunk size (tokens/step) | CUDA Graph time | vs. SGLang Cascade | vs. Custom Mask |
|---|---|---:|---:|---:|---:|
| FA2 | `BatchPrefillWithRaggedKVCacheWrapper` | 128 | 19.605 ms | 1.29x | 1.44x |
| FA3 | `BatchPrefillWithRaggedKVCacheWrapper` | 128 | 13.359 ms | 6.28x | 2.10x |

Speedups are computed as baseline latency divided by Block Extend latency; higher is better. The `vs. Custom Mask` values compare against the Custom Mask baseline at the same backend and chunk size. The benchmark numbers are specific to the NVIDIA H20 and this configuration; they should not be interpreted as a direct comparison across GPU models.

---

### LLaDA2.0-flash-CAP End-to-End Benchmark Results

> **Environment**: NVIDIA H20; integrated with SGLang on an internal build and tested with the LLaDA2.0-flash-CAP model.
>
> - **Short prompt**: 4K input tokens, 1.5K output tokens
> - **Long prompt**: 24K input tokens, 1.5K output tokens
> - **Throughput** (tokens/s): higher is better
> - **TTFT** (Time To First Token, ms): lower is better
> - **TPOT** (Time Per Output Token, ms): lower is better
>
> Version labels refer to the corresponding internal SGLang/model builds. The `(no cache)` suffix follows the original benchmark configuration naming.

| Version | Metric | Short Prompt / Concurrency 1 | Short Prompt / Concurrency 4 | Long Prompt / Concurrency 1 | Long Prompt / Concurrency 4 |
| :--- | :--- | ---: | ---: | ---: | ---: |
| 260209 (no cache) | Throughput | 460.60 | 1241.64 | 2789.68 | 5939.58 |
|  | TTFT | 269.55 | 818.44 | 794.71 | 2704.73 |
|  | TPOT | 8.01 | 11.62 | 5.80 | 9.39 |
| 251225 (no cache) | Throughput | 448.03 | 911.07 | 1396.97 | 2197.70 |
|  | TTFT | 1162.47 | 5826.76 | 6341.80 | 22274.96 |
|  | TPOT | 7.66 | 12.48 | 8.48 | 16.70 |

#### TTFT Optimization Highlights

TTFT is strongly influenced by prefill performance, although these end-to-end measurements also include framework and scheduling effects:

| Scenario | 251225 (no cache) | 260209 (no cache) | Speedup |
| :--- | ---: | ---: | ---: |
| Short Prompt / Concurrency 1 | 1162.47 ms | 269.55 ms | **4.31x** |
| Short Prompt / Concurrency 4 | 5826.76 ms | 818.44 ms | **7.12x** |
| Long Prompt / Concurrency 1 | 6341.80 ms | 794.71 ms | **7.98x** |
| Long Prompt / Concurrency 4 | 22274.96 ms | 2704.73 ms | **8.23x** |

Across these internal builds, the observed TTFT is reduced by **4.3x–8.2x**. These are system-level results and should not be attributed solely to the Block Extend kernel.

refs : https://github.com/sgl-project/sglang/issues/12766


ClawSeven:
Hi, @yzh119, @bkryu, 

This PR is quite impactful for block diffusion LLM — we’ve seen several times improvement in TTFT in our production environment at AntGroup when using SGLang.

By the way, I’m the code owner of SGLang-dLLM. I’d expect to collaborate with you to make FlashInfer as the official recommended backend for SGLang-dLLM. I think it would be helpful for both projects and the broader community.
Looking forward to working together!

SGLang dLLM roadmap: https://github.com/sgl-project/sglang/issues/14199

include/flashinfer/attention/default_prefill_params.cuh
```diff
@@ -405,8 +456,18 @@ struct BatchPrefillPagedParams {
...
+  // Get batch item's KV global offset (for Cascade Current Chunk)
+  __host__ __device__ __forceinline__ uint32_t get_kv_block_expanding_offset(uint32_t batch_idx) const {
+    return (maybe_kv_block_expanding_offset != nullptr) ? maybe_kv_block_expanding_offset[batch_idx] : 0;
+  }
 };
 
 }  // namespace flashinfer
 
-#endif  // FLASHINFER_DECODE_PARAMS_CUH_
+#endif  // FLASHINFER_DECODE_PARAMS_CUH_
```
yzh119: Please fix the lint issue (new line in the end).

include/flashinfer/utils.cuh
```diff
@@ -518,6 +523,12 @@ __device__ __forceinline__ uint32_t dim4_offset(const uint32_t& dim_c, const uin
...
+// DLLM Block Expanding related type traits (defined once here to avoid ODR violations)
+DEFINE_HAS_MEMBER(dllm_block_size)
+DEFINE_HAS_MEMBER(q_block_expanding_offset)
+DEFINE_HAS_MEMBER(maybe_q_block_expanding_offset)  // per-batch q_offset pointer for FA3 BatchPrefill
+DEFINE_HAS_MEMBER(kv_block_expanding_offset)       // kv_offset for Cascade Current Chunk
+DEFINE_HAS_MEMBER(maybe_kv_block_expanding_offset) // per-batch kv_offset pointer for FA3 BatchPrefill
 }  // namespace flashinfer
 
-#endif  // FLASHINFER_UTILS_CUH_
+#endif  // FLASHINFER_UTILS_CUH_
```
yzh119: ditto.

yzh119 (commented):
Hi @fdz-1999 thanks for the great work and it make sense to me in general.

Adding a new mask type would significantly bloat binary size (in jit-cache, etc.), can we make it a standalone class instead of changing `csrc/batch_prefill_customize_config.jinja` etc?

fdz-1999:
> Hi @fdz-1999 thanks for the great work and it make sense to me in general.
> 
> Adding a new mask type would significantly bloat binary size (in jit-cache, etc.), can we make it a standalone class instead of changing `csrc/batch_prefill_customize_config.jinja` etc?

Thanks for the review and the concern about binary size — it's a really important consideration for a JIT-heavy project like FlashInfer, and I want to make sure the impact is clear.

I looked into this carefully, and I believe the current approach should not cause binary bloat for non-DLLM users. Here's the reasoning:

**1) Mode 4 is not compiled by default**

The JIT default in `modules.py` is `mask_modes=[0, 1, 2, 3]` across all 4 code paths (single prefill, batch ragged, batch paged, SM90 variants). Mode 4 (`kBlockExpanding`) is only compiled when DLLM wrappers explicitly request it via `mask_modes=[4]`. So for existing users, there are no extra `.cu` files generated and no extra JIT cache entries.

**2) Jinja template changes are conditionally compiled away**

The `get_q/kv_block_expanding_offset()` methods added to `RaggedParams` / `PagedParams` are guarded by Jinja `{% if %}` conditions:

```jinja
{% if 'maybe_q_block_expanding_offset' in additional_params_decl %}
    return maybe_q_block_expanding_offset[batch_idx];
{% else %}
    return 0;
{% endif %}
```

Non-DLLM callers never pass `maybe_q_block_expanding_offset`, so these methods render as `return 0` and get inlined away by the compiler.

**3) The kernel-level changes follow the existing MaskMode pattern**

The additions in `prefill.cuh` and the Hopper files are all `if constexpr (MASK_MODE == MaskMode::kBlockExpanding)` branches — the same mechanism used by `kCausal`, `kCustom`, and `kMultiItemScoring`. When mode 4 is not instantiated, these branches are eliminated at compile time with zero code generation.

The only "global" changes are:
- `mask.cuh`: one enum member (`kBlockExpanding = 4U`) — no code generation impact
- `utils.cuh`: one `case` in `DISPATCH_MASK_MODE` — only reached at runtime when mode 4 is explicitly requested, and non-DLLM users never pass mode 4

**4) Regarding the standalone class alternative**

I did consider this direction. The challenge is that block expanding only differs from the existing kernel at a few specific points (mask predicate, iteration count, mask_iteration boundary — ~50 lines of `if constexpr` branches in `prefill.cuh`, with similar patterns in the 4 Hopper files). The rest of the kernel — QKV matmul, online softmax, tiling, shared memory management, warp-level MMA — is identical.

A standalone implementation would need to duplicate the full prefill kernel stack (`prefill.cuh` + 4 Hopper files, ~4500 lines total), plus maintain a separate set of Jinja templates and JIT pipeline. This would create a significant ongoing maintenance burden, especially when upstream kernel optimizations or bugfixes need to be synced.

Given that the current approach has zero impact on non-DLLM users (no extra compilation, no binary size increase, no runtime overhead), I think extending the existing MaskMode infrastructure is the more sustainable path. But I'm happy to discuss further if there are specific scenarios you're concerned about!

include/flashinfer/utils.cuh
fdz-1999: done

include/flashinfer/attention/default_prefill_params.cuh
fdz-1999: done

fdz-1999:
@yzh119  I've replied to your binary-size concern in the comment above with a detailed breakdown, and the remaining code-level comments have been addressed as well. If any concern remains or needs further refinement, please let me know and I'll follow up accordingly. Could you please take another look at this PR when you have a chance? A re-review would be much appreciated — thank you!

YangXu1990uiuc:
makes sense to move this forward in FlashInfer rather than keep it downstream — the demand is concrete. The PR comes out of Ant Group's SGLang-dLLM production deployment (they report several-x TTFT improvement in the PR thread), and their code owner has explicitly offered to make FlashInfer the official recommended backend for SGLang-dLLM. The kernel approach itself is sound: it reuses the existing causal two-boundary tile structure (fully-visible tiles → direct MMA, boundary tiles → register-computed mask, invisible tiles → skipped at CTA granularity), with zero mask memory. Zihao has already said the direction makes sense in general.

That said, I'd ask for two design changes before it merges:

1. Keep the new mask mode out of the shared compilation space. As it stands, the PR adds kBlockExpanding to the shared batch_prefill_customize_config.jinja (and siblings), which enumerates one more mask-mode value for every existing prefill URI — inflating jit-cache/AOT binary size for all users, dLLM or not. This is Zihao's concern on the thread and I agree. The fix: instantiate a small dedicated set of dLLM URIs where mask_mode is fixed to kBlockExpanding and the other axes are restricted to what dLLM actually needs (fp16/bf16, head_dim 64/128, ragged + paged) — a separate small cartesian product, not a new value multiplied into the big one. The PR already defines dedicated block_expanding_* URIs, so this is mostly about reverting the shared jinja/config edits and moving the dispatch into a standalone config/variant class.

2. Converge the API surface. The PR introduces a parallel flashinfer/dllm/ package with four new APIs, its own module cache, and its own AOT path. We're actively trying to reduce attention-API fragmentation, and structured masks already have upstream precedent (multi-item scoring). Preferred shape: expose this as a mask option on the existing prefill APIs / wrappers (e.g. mask_mode="block_diffusion" plus block_size and q_offset/kv_offset) rather than a new API family users have to discover separately.

Separately, a few reviewer-flagged correctness items should be cleared before merge (zero-visible-KV TMA hang path in the Hopper mainloop, module URI missing idtype, fp8 silently coerced to fp16) — but those are normal review iteration, not direction questions.

fdz-1999:
> makes sense to move this forward in FlashInfer rather than keep it downstream — the demand is concrete. The PR comes out of Ant Group's SGLang-dLLM production deployment (they report several-x TTFT improvement in the PR thread), and their code owner has explicitly offered to make FlashInfer the official recommended backend for SGLang-dLLM. The kernel approach itself is sound: it reuses the existing causal two-boundary tile structure (fully-visible tiles → direct MMA, boundary tiles → register-computed mask, invisible tiles → skipped at CTA granularity), with zero mask memory. Zihao has already said the direction makes sense in general.
> 
> That said, I'd ask for two design changes before it merges:
> 
> 1. Keep the new mask mode out of the shared compilation space. As it stands, the PR adds kBlockExpanding to the shared batch_prefill_customize_config.jinja (and siblings), which enumerates one more mask-mode value for every existing prefill URI — inflating jit-cache/AOT binary size for all users, dLLM or not. This is Zihao's concern on the thread and I agree. The fix: instantiate a small dedicated set of dLLM URIs where mask_mode is fixed to kBlockExpanding and the other axes are restricted to what dLLM actually needs (fp16/bf16, head_dim 64/128, ragged + paged) — a separate small cartesian product, not a new value multiplied into the big one. The PR already defines dedicated block_expanding_* URIs, so this is mostly about reverting the shared jinja/config edits and moving the dispatch into a standalone config/variant class.
> 2. Converge the API surface. The PR introduces a parallel flashinfer/dllm/ package with four new APIs, its own module cache, and its own AOT path. We're actively trying to reduce attention-API fragmentation, and structured masks already have upstream precedent (multi-item scoring). Preferred shape: expose this as a mask option on the existing prefill APIs / wrappers (e.g. mask_mode="block_diffusion" plus block_size and q_offset/kv_offset) rather than a new API family users have to discover separately.
> 
> Separately, a few reviewer-flagged correctness items should be cleared before merge (zero-visible-KV TMA hang path in the Hopper mainloop, module URI missing idtype, fp8 silently coerced to fp16) — but those are normal review iteration, not direction questions.

Below is a cleaner, review-friendly version:

---

@YangXu1990uiuc Thanks — I agree with both design requirements. I’ve reworked the series accordingly and will keep this PR in Draft while I complete refreshed validation against the latest `main` baseline.

### Keep `kBlockExpanding` out of the shared compilation space

The shared customize generators continue to compile only the existing mask modes: `[0, 1, 2, 3]`.

`kBlockExpanding` is generated exclusively through the dedicated:

- `gen_customize_block_extend_single_prefill_module`
- `gen_customize_block_extend_batch_prefill_module`

paths, where mask mode is fixed to `4`. The dLLM product is explicitly limited to the required variants:

- `fp16` / `bf16`
- head dimensions `64` / `128`
- ragged and paged batch variants

The remaining Jinja-template changes do not add a fifth specialization to existing prefill URIs. They render the exact dispatcher required by each generated module.

The SM90 templates also expose `get_qo_len` and `get_kv_len`, which are required for Hopper tile-visibility handling, including the zero-visible-KV path.

As a result, existing shared prefill modules gain no additional JIT-cache or AOT footprint from this feature.

### Converge on the existing prefill API surface

The canonical user-facing path remains the existing prefill API:

```python
single_prefill_with_kv_cache(
    ...,
    block_diffusion=True,
    dllm_block_size=...,
    q_offset=...,
    kv_offset=...,
)
```

For batched prefill, the existing ragged/paged wrappers are used:

```python
BatchPrefillWith*KVCacheWrapper(
    ...,
    block_diffusion=True,
    dllm_block_size=...,
)
```

with `q_offsets` and `kv_offsets` provided during `plan()`.

The `flashinfer.dllm` entry points are transitional SGLang-compatibility shims, not a permanent second API surface. They currently delegate to the native prefill path and shared variant wiring, with no independent module cache or AOT product.

Once SGLang migrates to the canonical `block_diffusion=True` call sites, these shims will be removed in a follow-up commit within this PR series. The final merged tree will expose exactly one dLLM prefill API surface. Removing the shim is purely a deletion of a delegating namespace and introduces no additional API-visible changes beyond what is already under review.

### Correctness and validation

I also addressed the review feedback:

- Dedicated batch URI identity now includes `idtype`.
- The dLLM path validates its supported dtype/head-dimension combinations rather than silently coercing FP8.
- Targeted SM90 tests cover the Hopper zero-visible-KV regression.

The targeted zero-visible and named-option tests pass. I’m rerunning the focused full suite after syncing with the latest `main`, and will mark the PR ready for review once that validation completes.

fdz-1999:
@BangXu1990uiuc Thanks for the detailed review. I reworked the PR around compilation-space isolation and API convergence, and addressed the remaining correctness concerns.

The API cleanup is included in [d78d7c28](https://github.com/flashinfer-ai/flashinfer/commit/d78d7c28929838533fcb7deac778e7d0b95c32c0).

### Compilation-space isolation

The shared prefill generators continue to instantiate only the existing mask modes:

```text
[0, 1, 2, 3]
```

Block Extend is generated only through dedicated single- and batch-prefill generators, with the mask mode fixed internally to `kBlockExtend`.

The dedicated specialization space is limited to:

- FP16 / BF16
- head dimensions 64 / 128
- single, ragged, and paged prefill
- FA2 / FA3

Existing shared prefill URIs do not gain a Block Extend specialization. Block Extend variants are generated only through the dedicated generators.

### Unified prefill API

The parallel `flashinfer.dllm` API family has been removed from the final public path. Block Extend is exposed through the existing prefill APIs:

```python
single_prefill_with_kv_cache(
    ...,
    block_extend=True,
    block_size=...,
    q_offset=...,
    kv_offset=...,
)
```

and the existing batch wrappers:

```python
BatchPrefillWith*KVCacheWrapper(
    ...,
    block_extend=True,
    block_size=...,
)
```

Per-request `q_offsets` and `kv_offsets` are supplied through `plan()`.

The public `plan(mask_mode=...)` parameter has been removed from both batch wrappers. Block Extend is selected internally through `block_extend=True`, so callers cannot bypass the supported FA2/FA3 backend restrictions.

### Correctness and validation

The following review issues have been addressed:

- Dedicated batch module identities include `idtype`.
- Unsupported dtype/head-dimension combinations are rejected.
- FP8 is rejected instead of being silently coerced to FP16.
- The existing SM90 zero-visible-KV fast path is preserved for Block Extend: the producer skips invalid loads, while the consumer writes zero output and `-inf` LSE. Regression tests cover zero-visible → normal requests in the same batch for both ragged and paged layouts.

### FA3 custom-mask reference

FA3 single-prefill custom packed-mask support was added on SM90 so that Block Extend is validated against a backend-matched reference:

- FA2 Block Extend vs. FA2 custom mask
- FA3 Block Extend vs. FA3 custom mask

This avoids cross-backend numerical differences affecting the correctness comparison.

### Block Extend-aware planning

The batch planner now schedules against the logically visible KV length instead of always using the full physical KV-cache length. The original indptr and page table remain unchanged for kernel addressing; only the logical length used for scheduling is reduced.

The main planning implementation is included in [`20e2e91d`](https://github.com/flashinfer-ai/flashinfer/commit/20e2e91d0f56863e38bc61427dc9c3586a806312), with follow-up physical-offset fixes in `5d44f009` and `a0c8cea3`.

### Benchmark

Benchmark environment: NVIDIA H20, CUDA Graph enabled, release-mode JIT compilation.

| KV length | Batch size | Effective KV length | Baseline (us) | Optimized (us) | Speedup |
|---:|---:|---:|---:|---:|---:|
| 16384 | 1 | 8448 | 206.8 | 138.1 | 1.50x |
| 32768 | 1 | 16640 | 383.5 | 225.8 | 1.70x |
| 65536 | 1 | 33024 | 742.6 | 407.4 | 1.82x |

Speedup is computed as `baseline / optimized`; higher is better.

### Validation

The focused correctness suite was run on an NVIDIA H20:

```bash
FLASHINFER_JIT_DEBUG=0 \
python -m pytest -v -s \
  tests/attention/test_block_extend_attention.py
```

Result:

```text
14 passed, 2 warnings
```

Coverage includes FP16/BF16, FA2/FA3, single/ragged/paged prefill, backend-matched custom-mask references, query and KV offsets, non-power-of-two and non-aligned shapes, zero-visible → normal regressions, and public `plan()` signature validation.

Numerical thresholds:

```text
FP16: max absolute difference < 1e-3
BF16: max absolute difference < 1e-2
```

@BangXu1990uiuc @yzh119 @bkryu, could you please take another look when convenient? Thanks!

fdz-1999:
ready for review

fdz-1999:
ready for review

Issue #2733, opened 2026-03-09
[Bug] flashinfer JIT ops are incompatible with torch.compile fullgraph mode


Johnsonms:
### Summary
Any flashinfer function backed by a JIT module fails when used inside `torch.compile(..., fullgraph=True)`, causing the entire model compilation to abort.

### Error
torch._dynamo.exc.Unsupported: Attempted to call function marked as skipped Explanation: Dynamo does not know how to trace the builtin posix.stat  from user code:
```
get_norm_module() -> gen_norm_module().build_and_load()-> JitSpec.is_aot -> aot_path.exists() -> os.stat()
```

  ### Root Cause

`register_custom_op` / `register_fake_op` in `utils.py` are currently no-ops.
This means flashinfer ops (`_rmsnorm`, `apply_rope`, `sampling_from_probs`, etc.) are plain Python functions — Dynamo traces into their implementations and hits `Path.exists()` → `posix.stat`, which it cannot handle.

  ### Affected Modules

  All modules using `@register_custom_op`: `activation`, `cascade`, `decode`,
  `norm`, `page`, `prefill`, `quantization`, `rope`, `sampling`, `gemm`,
  `fused_moe`, and more.

  ### Expected Behavior

  ```python
  compiled = torch.compile(flashinfer.rmsnorm, fullgraph=True, backend="eager")
  compiled(x, w)  # should work without graph breaks 

 ###  Fix
  Re-enable torch.library.custom_op and torch.library.register_fake in
  utils.py. Dynamo will treat flashinfer ops as opaque custom ops, call fake
  implementations for shape inference, and never enter the Python body.

PR #2734, opened 2026-03-09
fix: torch.compile(fullgraph=True) compatibility for norm ops

Johnsonms:
## 📌 Description
  https://github.com/flashinfer-ai/flashinfer/issues/2733

`register_custom_op` / `register_fake_op` in `utils.py` were no-ops, causing JIT-backed FlashInfer ops to behave like plain Python functions. Under `torch.compile(fullgraph=True)`, Dynamo would trace into the Python implementation and hit `Path.exists() -> posix.stat`, which it cannot trace, breaking compilation.

  This PR re-enables `torch.library.custom_op` and
  `torch.library.register_fake` so Dynamo treats these ops as opaque custom ops
  instead of entering the Python body.

  It also removes the now-unnecessary `@torch.compiler.disable` from
  `get_norm_module` in `norm.py`, and adds
  `tests/norm/test_norm_torch_compile.py` with 40 regression cases covering:

  - `flashinfer.rmsnorm`
  - `flashinfer.fused_add_rmsnorm`
  - `flashinfer.gemma_rmsnorm`
  - `flashinfer.gemma_fused_add_rmsnorm`

  all under `torch.compile(fullgraph=True)`.

  Scope note: the regression coverage added in this PR is for norm ops.

  ## 🔍 Related Issues

  Fixes #<issue_number>

  ## 🚀 Pull Request Checklist

  ### ✅ Pre-commit Checks

  ## 🧪 Tests

    - `pytest tests/norm/test_norm_torch_compile.py -v`
    - `40 passed`

  ## Reviewer Notes

  This PR fixes the `torch.compile(fullgraph=True)` regression for the norm path.
  The custom-op registration change is shared infrastructure, but the added
  regression coverage in this PR is norm-only.

## Release Notes

* **Tests**
  * Added comprehensive torch.compile compatibility tests for multiple RMSNorm variants across various batch sizes, hidden dimensions, and data types (float16, bfloat16), validating compiled execution against reference results.

* **Improvements**
  * Replaced previous stub behavior with direct PyTorch library integration for custom/fake operation registration, improving runtime alignment with PyTorch.

### DIFF
--- a/flashinfer/utils.py
+++ b/flashinfer/utils.py
@@ -344,24 +344,19 @@ else:
         device_types: Optional[Union[str, Sequence[str]]] = None,
         schema: Optional[str] = None,
     ) -> Callable:
-        # NOTE(Zihao): torch.library.custom_op has significant overhead as mentioned in the following link
-        # https://github.com/vllm-project/vllm/blob/36e76700453924c8d421db99af70a88a1df835cd/vllm/utils.py#L1660-L1674
-
-        # return torch.library.custom_op(
-        #     name,
-        #     fn,
-        #     mutates_args=mutates_args,
-        #     device_types=device_types,
-        #     schema=schema,
-        # )
-        return lambda x: x
+        return torch.library.custom_op(
+            name,
+            fn,
+            mutates_args=mutates_args,
+            device_types=device_types,
+            schema=schema,
+        )
 
     def register_fake_op(
         name: str,
         fn: Optional[Callable] = None,
     ) -> Callable:
-        # return torch.library.register_fake(name, fn)
-        return lambda x: x
+        return torch.library.register_fake(name, fn)
 
 
 def determine_gemm_backend(device: torch.device) -> str:
--- a/tests/norm/test_norm_torch_compile.py
+++ b/tests/norm/test_norm_torch_compile.py
@@ -1,0 +1,146 @@
+# Copyright (c) 2025 by FlashInfer team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Tests for norm functions compatibility with torch.compile.
+
+Verifies that rmsnorm / fused_add_rmsnorm / gemma_rmsnorm /
+gemma_fused_add_rmsnorm can be traced by Dynamo with fullgraph=True
+(i.e. no graph breaks, no posix.stat issues).
+"""
+
+import pytest
+import torch
+from torch.torch_version import TorchVersion
+from torch.torch_version import __version__ as torch_version
+
+if TorchVersion(torch_version) < TorchVersion("2.4"):
+    pytest.skip("torch.compile support requires torch >= 2.4", allow_module_level=True)
+
+import flashinfer
+
+
+def _rmsnorm_ref(x: torch.Tensor, w: torch.Tensor, eps: float = 1e-6) -> torch.Tensor:
+    orig = x.dtype
+    x = x.float()
+    x = x * torch.rsqrt(x.pow(2).mean(-1, keepdim=True) + eps)
+    return (x * w.float()).to(orig)
+
+
+def _gemma_rmsnorm_ref(
+    x: torch.Tensor, w: torch.Tensor, eps: float = 1e-6
+) -> torch.Tensor:
+    orig = x.dtype
+    x = x.float()
+    x = x * torch.rsqrt(x.pow(2).mean(-1, keepdim=True) + eps)
+    return (x * (w.float() + 1.0)).to(orig)
+
+
+def _fused_add_rmsnorm_ref(
+    x: torch.Tensor, residual: torch.Tensor, w: torch.Tensor, eps: float = 1e-6
+):
+    residual = residual + x
+    orig = residual.dtype
+    r = residual.float()
+    r = r * torch.rsqrt(r.pow(2).mean(-1, keepdim=True) + eps)
+    return (r * w.float()).to(orig), residual
+
+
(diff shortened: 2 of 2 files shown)


tests/norm/test_norm_torch_compile.py
```diff
@@ -0,0 +1,114 @@
...
+def test_gemma_rmsnorm_torch_compile(batch_size, hidden_size, dtype):
+    x = torch.randn(batch_size, hidden_size, dtype=dtype, device="cuda")
+    w = torch.zeros(hidden_size, dtype=dtype, device="cuda")  # gemma: (w+1)*norm
+
+    # warmup
+    flashinfer.gemma_rmsnorm(x, w)
+
+    compiled = torch.compile(flashinfer.gemma_rmsnorm, fullgraph=True, backend="eager")
+    out = compiled(x, w)
+    ref = flashinfer.gemma_rmsnorm(x, w)
```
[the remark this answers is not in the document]
Johnsonms: fixed

tests/norm/test_norm_torch_compile.py
```diff
@@ -0,0 +1,114 @@
...
+
+Verifies that rmsnorm / fused_add_rmsnorm / gemma_rmsnorm /
+gemma_fused_add_rmsnorm can be traced by Dynamo with fullgraph=True
+(i.e. no graph breaks, no posix.stat issues).
+"""
+
+import pytest
+import torch
+
+import flashinfer
```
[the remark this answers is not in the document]
Johnsonms: Fixed

flashinfer/utils.py
```diff
@@ -344,24 +344,19 @@ def register_custom_op(
...
-
-        # return torch.library.custom_op(
-        #     name,
-        #     fn,
-        #     mutates_args=mutates_args,
-        #     device_types=device_types,
-        #     schema=schema,
-        # )
-        return lambda x: x
+        return torch.library.custom_op(
```
yzh119, line 347: If we turn this on again, it's more than just normalization kernel and we probably need to start working on cudagraph sanity check across the codebase (but let's see how CI works first).

As far as I remember sglang wraps these API again with torch library decorators so it should not be a problem?

nvpohanh: @Johnsonms is it possible for SGLang to wrap the FlashInfer op on SGLang-side?

Johnsonms: Thanks @nvpohanh @yzh119 !

Yes @yzh119, you are right — SGLang (sgl-kernel) already wraps these ops. For flexibility, we currently support JIT kernels, where the SGLang wrapper directly calls the FlashInfer JIT implementation.

The issue with torch.compile(..., fullgraph=True) is not just the wrapper itself. Dynamo still sees the underlying FlashInfer JIT-backed APIs as regular Python functions, so it traces into them.

During tracing it eventually hits the JIT loading logic (e.g., Path.exists() / os.stat()), which Dynamo cannot capture. As a result, CUDA graph–wise compilation breaks or fails in this path.

nvpohanh: @yzh119 Is it possible that in FlashInfer, we somehow "detect" if it is in capture mode and if so, we skip those JIT loading logic?

nvjullin: @Johnsonms I'm a bit confused.
If sglang wraps with custom op, why does dynamo trace into the op?
If not, why not?

Johnsonms: Good question @nvjullin , Let me take time to investigate and get the everything clear tomorrow

Johnsonms: There is an online urgent issue, having to work there, will do this in weekend. Sorry about that

DevashishLal-CB: @Johnsonms @yzh119 @nvpohanh I am using these kernels in sglang for fusion by registering them on sglang side

https://github.com/sgl-project/sglang/pull/10549/changes#diff-a3005a66a8755c5a5c2043c67c32bf0be63134bd45b4cd70455ada4c06cb3d0fR51

It does work but it would be nice to enable this support in flashinfer to avoid this duplication for all the kernels I want to use.

not sure if performance is still a concern but vllm and sglang work around this by using the direct registration of custom ops

https://github.com/vllm-project/vllm/blob/36e76700453924c8d421db99af70a88a1df835cd/vllm/utils.py#L1696

Johnsonms: Thanks @nvjullin — I looked into this and the root cause is now clearer.

The FlashInfer JIT kernel APIs (e.g. `flashinfer.norm.rmsnorm`) currently appear as plain Python functions to Dynamo. Although FlashInfer has `register_custom_op` / `register_fake_op` decorators, those registrations are effectively no-ops right now, so the functions are not exposed as real PyTorch custom ops.

On the SGLang side we call these FlashInfer Python APIs directly. Since Dynamo does not see an opaque custom op boundary, it traces into the Python implementation. During tracing it eventually reaches the JIT loading logic (`build_and_load()`, `Path.exists()`, `os.stat()`), which Dynamo cannot capture, and this breaks `torch.compile(..., fullgraph=True)`.

This is also why the internal `torch.ops.sgl_kernel.*` (or we called AOT) path works: those kernels are properly registered as PyTorch operators in C++, so Dynamo treats them as opaque ops and does not trace into them.

The current workaround (as @DevashishLal-CB  mentioned) is to register the kernels as custom ops on the SGLang side. A cleaner fix would be for FlashInfer to expose these JIT-backed APIs as proper torch custom ops so Dynamo treats them as opaque operators.

nvjullin: Is this a correct summary then?
1. `torch.compile` works after wrapping flashinfer functions in custom ops
2. But sglang would like the custom op wrapping be in flashinfer instead

nvpohanh: @xinli-sw could you find someone from vLLM team to review this? I think the original perf concern was raised by vLLM team

benchislett: I assume vLLM will want to wrap the custom op itself, but not sure about the performance implications here.

benchislett: I have seen the `posix.stat` crash before so we will want some kind of workaround for this issue. I don't have a strong intuition for whether FI should be the one to wrap the op or not.

nvjullin: > not sure about the performance implications here

I believe the history was vLLM asked flashinfer to turn **off** wrapping, because they wanted to hack into the torch custom op machinery https://github.com/vllm-project/vllm/pull/9896. The main motivation being perf.

Issue #2340, opened 2026-01-12
`chunk_gated_delta_rule` for Blackwell


vadiklyutiy:
Right now `chunk_gated_delta_rule` supports only Hopper 

https://github.com/flashinfer-ai/flashinfer/blob/2062decf4f5465974806a547450e3a150d68718c/flashinfer/gdn_prefill.py#L81

It'd be good to have it for Blackwell as well.

kahyunnam:
Checked in with DianZhang Chen; he has finished basic logic and is now working on optimization. The ETA is around mid-to-late February (though the Chinese New Year holiday may impact this)

dianzhangchen:
> Checked in with DianZhang Chen; he has finished basic logic and is now working on optimization. The ETA is around mid-to-late February (though the Chinese New Year holiday may impact this)

hi @kahyunnam , I just submitted a PR here: [2742](https://github.com/flashinfer-ai/flashinfer/pull/2742). thanks

kahyunnam:
Still in review, moving to April roadmap

guangyunh-nv:
I think #3001 fixed this.

PR #2729, opened 2026-03-09
[gdn] cuTile GDN prefill on Blackwell (SM100)

xutizhou:
## Summary      

  Adds accuracy tests and benchmarks for the cuTile-based GDN prefill kernel on                                                                                                                                                                                                                                                                                                           
  NVIDIA Blackwell (SM100 / B200), comparing against the FLA Triton baseline.
                                                                                                                                                                                                                                                                                                                                                                                          
  ## What's included                                                                                                                                                                                                                                                                                                                                                                      
  
  - `tests/gdn/test_prefill_cutile_blackwell.py` — 33 parametrized accuracy tests                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
  - `benchmarks/bench_gdn_prefill_cutile_vs_fla_blackwell.py` — kernel latency                                                                                                                                                                                                                                                                                                            
    comparison via `torch.profiler` CUDA events.                                                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                    
  ## Qwen3.5 GDN Parameters                                                                                                                        
                                                                                                                                                   
  | Parameter | Value | HF config field |                                                                                                          
  |-----------|-------|-----------------|
  | K (key head dim) | 128 | `linear_key_head_dim` |                                                                                               
  | V (value head dim) | 128 | `linear_value_head_dim` |                                                                                           
  | H (value heads) | 32 (4B) / 64 (397B-A17B) | `linear_num_value_heads` |                                                                        
  | H_k (key heads, GQA) | 16 | `linear_num_key_heads` |                                                                                           
  | BT (chunk size) | 64 | FlashInfer/FLA default |                                                                                                
  | q/k/v dtype | bfloat16 | model weight dtype |                                                                                                  
  | SSM state dtype | **float32** | sglang `mamba2_state_dtype` default |                                                                          
  | g, beta dtype | float32 | numerical stability | 

### Qwen3.5-397B TP8 (H=8, K=128, V=128)

| Config | FLA (ms) | cuTile (ms) | Speedup |
|--------|----------|-------------|---------|
| B=1, T=2048 | 0.092 | 0.056 | **1.65x** |
| B=2, T=2048 | 0.125 | 0.080 | **1.55x** |
| B=4, T=2048 | 0.184 | 0.120 | **1.54x** |
| B=8, T=1024 | 0.171 | 0.107 | **1.60x** |
[...]

PR #2742, opened 2026-03-10
Add blackwell GDN prefill impl

dianzhangchen:
## 📌 Description

Gated Delta Networks (GDN) chunked linear attention kernel for NVIDIA Blackwell (SM100).
Implements the Chunk-wise Gated Delta Rule linear attention using CuTe-DSL, for Blackwell Architecture.

Key Features:
- Supports Persistent & Non-persistent modes    
- Supports fixed-length and variable-length sequences (cu_seqlens)
- Supports grouped value attention (GVA) where h_v is a multiple of h_q
- Supports head dimension (128)
- Supports input data types (f16/bf16)
- Supports output data types (f16/bf16)
- Supports initial_state to provide the initial state
- Supports output_final_state flag to return the final state
- State input and output are in f32 (fp16/bf16 not supported yet)

## Performance on B200

+ dtype: bfloat16
+ use_initial_state
+ output_final_state=True
+ Fixlen case

| batch_size | seq_len | qk_heads | v_heads | head_dim | dtype | gdn_ms | fla_ms | speedup |
|---|---:|---:|---:|---:|---|---:|---:|---:|
| 1  | 512  | 96  | 96  | 128 | bfloat16 | 0.0501 | 0.2119 | 4.2275 |
| 1  | 1024 | 96  | 96  | 128 | bfloat16 | 0.0869 | 0.2233 | 2.5696 |
| 1  | 4096 | 96  | 96  | 128 | bfloat16 | 0.3082 | 0.7741 | 2.5118 |
| 1  | 8192 | 96  | 96  | 128 | bfloat16 | 0.6009 | 1.4919 | 2.4827 |
| 9  | 512  | 32  | 32  | 128 | bfloat16 | 0.0986 | 0.3030 | 3.0740 |
| 9  | 1024 | 32  | 32  | 128 | bfloat16 | 0.1735 | 0.5529 | 3.1858 |
| 9  | 4096 | 32  | 32  | 128 | bfloat16 | 0.6189 | 2.0558 | 3.3219 |
| 9  | 8192 | 32  | 32  | 128 | bfloat16 | 1.2123 | 4.0530 | 3.3434 |
| 33 | 512  | 32  | 32  | 128 | bfloat16 | 0.3461 | 1.0199 | 2.9469 |
| 33 | 1024 | 32  | 32  | 128 | bfloat16 | 0.6159 | 1.9385 | 3.1474 |
| 33 | 4096 | 32  | 32  | 128 | bfloat16 | 2.2590 | 7.4736 | 3.3084 |
| 33 | 8192 | 32  | 32  | 128 | bfloat16 | 4.4938 | 14.8813 | 3.3115 |
| 1  | 512  | 148 | 148 | 128 | bfloat16 | 0.0521 | 0.2208 | 4.2371 |
| 1  | 1024 | 148 | 148 | 128 | bfloat16 | 0.0955 | 0.3089 | 3.2342 |
| 1  | 4096 | 148 | 148 | 128 | bfloat16 | 0.3232 | 1.1166 | 3.4551 |
| 1  | 8192 | 148 | 148 | 128 | bfloat16 | 0.6310 | 2.1957 | 3.4799 |

## Issues

+ In the current version, we do not parallelize over the sequence-length dimension, so performance is limited when both batch size and head count are small. In such cases, only part of the 148 SMs are used, so the GPU is not fully utilized.
+ Current version supports GVA only; GQA is not supported.
+ State input and output are FP32 only; FP16 is not supported in the current version.

## 🔍 Related Issues

[chunk_gated_delta_rule for Blackwell #2340](https://github.com/flashinfer-ai/flashinfer/issues/2340)

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * Blackwell GPU support for GDN (Gated Delta Network) linear attention with SM-specific dispatch.
  * Built-in performance benchmark tool (fixed/variable lengths, sweep mode, formatted reports).
  * Exposes optional kernel entry for Blackwell prefill and public symbols for kernel access.

* **Tests**
  * Comprehensive GPU test suite: fixed/variable lengths, grouped-value attention variants, final-state outputs, and determinism across runs.

### DIFF
--- a/benchmarks/bench_blackwell_gdn_prefill.py
+++ b/benchmarks/bench_blackwell_gdn_prefill.py
@@ -31,9 +31,11 @@ from typing import List
 
 import torch
 import torch.nn.functional as F
+import numpy as np
+from flashinfer.testing import bench_gpu_time
 
 # Blackwell GDN prefill
-from flashinfer.gdn_kernels.blackwell_prefill.gdn import chunk_gated_delta_rule
+from flashinfer.gdn_prefill import chunk_gated_delta_rule
 
 # FLA baseline
 from fla.ops.gated_delta_rule.chunk import chunk_gated_delta_rule_fwd as fla_base
@@ -99,53 +101,35 @@ def benchmark_gdn_fixlen(
     output_final_state = True
 
     o = torch.zeros_like(v)
-    # Warmup
-    for _ in range(warmup_iters):
-        chunk_gated_delta_rule(
-            q,
-            k,
-            v,
-            g,
-            beta,
-            None,
-            h0,
-            output_final_state,
-            None,
-            False,
-            o,
-            None,
-        )
-    torch.cuda.synchronize()
+    state_output = torch.randn(
+        (batch_size, num_v_heads, head_dim, head_dim),
+        dtype=torch.float32,
+        device=device,
+    )
 
-    # Benchmark
-    torch.cuda.reset_peak_memory_stats()
-    start_event = torch.cuda.Event(enable_timing=True)
-    end_event = torch.cuda.Event(enable_timing=True)
+    fn_gdn = lambda: chunk_gated_delta_rule(
+        q,
+        k,
+        v,
+        g,
+        beta,
+        None,
+        h0,
+        output_final_state,
+        None,
+        False,
+        o,
+        state_output,
+    )
 
-    o.zero_()
+    kernel_times_ms = bench_gpu_time(
+        fn_gdn,
+        enable_cupti=True,
+        dry_run_iters=warmup_iters,
+        repeat_iters=benchmark_iters,
+    )
 
-    start_event.record()
-    for _ in range(benchmark_iters):
-        chunk_gated_delta_rule(
-            q,
-            k,
-            v,
-            g,
-            beta,
-            None,
-            h0,
-            output_final_state,
-            None,
-            False,
-            o,
-            None,
-        )
-
-    end_event.record()
-    torch.cuda.synchronize()
-
-    elapsed_ms = start_event.elapsed_time(end_event)
-    avg_latency_ms = elapsed_ms / benchmark_iters
+    avg_latency_ms = np.average(kernel_times_ms)
 
     # FLA baseline
     fla_latency_ms = None
@@ -161,19 +145,13 @@ def benchmark_gdn_fixlen(
             output_final_state=output_final_state,
         )
 
-        for _ in range(warmup_iters):
-            fn_fla()
-        torch.cuda.synchronize()
-
-        fla_start = torch.cuda.Event(enable_timing=True)
-        fla_end = torch.cuda.Event(enable_timing=True)
-        fla_start.record()
-        for _ in range(benchmark_iters):
-            fn_fla()
-        fla_end.record()
-        torch.cuda.synchronize()
-
-        fla_latency_ms = fla_start.elapsed_time(fla_end) / benchmark_iters
+        fla_times_ms = bench_gpu_time(
+            fn_fla,
+            enable_cupti=True,
+            dry_run_iters=warmup_iters,
+            repeat_iters=benchmark_iters,
+        )
+        fla_latency_ms = np.average(fla_times_ms)
 
     result = {
         "batch_size": batch_size,
@@ -239,53 +217,29 @@ def benchmark_gdn_varlen(
     o = torch.zeros_like(v)
     state_output = torch.zeros_like(h0, dtype=torch.float32)
 
-    # Warmup
-    for _ in range(warmup_iters):
-        chunk_gated_delta_rule(
-            q,
-            k,
-            v,
-            g,
-            beta,
-            None,
-            h0,
-            True,
-            cu_seqlens_tensor,
-            False,
-            o,
-            state_output,
-        )
-    torch.cuda.synchronize()
+    fn_gdn = lambda: chunk_gated_delta_rule(
+        q,
+        k,
+        v,
+        g,
+        beta,
+        None,
+        h0,
+        True,
+        cu_seqlens_tensor,
+        False,
+        o,
+        state_output,
+    )
 
-    # Benchmark
-    torch.cuda.reset_peak_memory_stats()
-    start_event = torch.cuda.Event(enable_timing=True)
-    end_event = torch.cuda.Event(enable_timing=True)
+    kernel_times_ms = bench_gpu_time(
+        fn_gdn,
+        enable_cupti=True,
+        dry_run_iters=warmup_iters,
+        repeat_iters=benchmark_iters,
+    )
 
-    o.zero_()
-    state_output.zero_()
-
-    start_event.record()
-    for _ in range(benchmark_iters):
-        chunk_gated_delta_rule(
-            q,
-            k,
-            v,
-            g,
-            beta,
-            None,
-            h0,
-            True,
-            cu_seqlens_tensor,
-            False,
-            o,
-            state_output,
-        )
-    end_event.record()
-    torch.cuda.synchronize()
-
-    elapsed_ms = start_event.elapsed_time(end_event)
-    avg_latency_ms = elapsed_ms / benchmark_iters
+    avg_latency_ms = np.average(kernel_times_ms)
 
     return {
         "num_seqs": num_seqs,
--- a/benchmarks/bench_blackwell_gdn_prefill.py
+++ b/benchmarks/bench_blackwell_gdn_prefill.py
@@ -1,0 +1,543 @@
+"""
+Copyright (c) 2026 by FlashInfer team.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+"""
+Performance benchmark for GDN (Gated Delta Network) linear attention.
+
+Usage:
+    python bench_blackwell_gdn_prefill.py
+    python bench_blackwell_gdn_prefill.py --batch-size 8 --seq-len 1024
+    python bench_blackwell_gdn_prefill.py --varlen --batch-size 4 --seq-len 2048
+    python bench_blackwell_gdn_prefill.py --varlen --cu-seqlens 0 512 1024 2048
+    python bench_blackwell_gdn_prefill.py --sweep
+"""
+
+import argparse
+import sys
+from typing import List
+
+import torch
+import torch.nn.functional as F
+
+# Blacwell GDN prefill
+from flashinfer.gdn_kernels.blackwell_prefill.gdn import chunk_gated_delta_rule
+
+# FLA baseline
+from fla.ops.gated_delta_rule.chunk import chunk_gated_delta_rule_fwd as fla_base
+
+
+def benchmark_gdn_fixlen(
+    batch_size: int,
+    seq_len: int,
+    num_qk_heads: int,
+    num_v_heads: int,
+    head_dim: int,
+    dtype: torch.dtype = torch.bfloat16,
+    warmup_iters: int = 10,
+    benchmark_iters: int = 100,
+    use_initial_state: bool = True,
+) -> dict:
+    """
+    Benchmark GDN with fixed-length sequences.
+
+    Returns:
+        Dictionary with benchmark results.
+    """
+    device = "cuda"
+    torch.cuda.reset_peak_memory_stats()
+
+    # Create input tensors
+    q = torch.randn(
+        (batch_size, seq_len, num_qk_heads, head_dim), dtype=dtype, device=device
+    )
+    k = F.normalize(
+        torch.randn(
+            batch_size,
+            seq_len,
+            num_qk_heads,
+            head_dim,
+            dtype=torch.float32,
+            device=device,
+        ),
+        p=2,
+        dim=-1,
+    ).to(dtype)
+    v = torch.randn(
+        (batch_size, seq_len, num_v_heads, head_dim), dtype=dtype, device=device
+    )
+    g = F.logsigmoid(
+        torch.rand(
+            1, seq_len * batch_size, num_v_heads, dtype=torch.float32, device=device
+        )
+    )
+    beta = torch.rand(
+        1, seq_len * batch_size, num_v_heads, dtype=torch.float32, device=device
+    ).sigmoid()
+
+    if use_initial_state:
+        h0 = torch.randn(
+            (batch_size, num_v_heads, head_dim, head_dim),
+            dtype=torch.float32,
+            device=device,
+        )
+    else:
+        h0 = None
+
+    output_final_state = True
+
+    o = torch.zeros_like(v)
+    # Warmup
+    for _ in range(warmup_iters):
+        chunk_gated_delta_rule(
+            q,
+            k,
+            v,
+            g,
+            beta,
+            None,
+            h0,
+            output_final_state,
+            None,
+            False,
+            o,
+            None,
+        )
+    torch.cuda.synchronize()
+
+    # Benchmark
+    torch.cuda.reset_peak_memory_stats()
+    start_event = torch.cuda.Event(enable_timing=True)
+    end_event = torch.cuda.Event(enable_timing=True)
+
+    o.zero_()
+
+    start_event.record()
+    for _ in range(benchmark_iters):
+        chunk_gated_delta_rule(
+            q,
+            k,
+            v,
+            g,
+            beta,
+            None,
+            h0,
+            output_final_state,
+            None,
+            False,
+            o,
+            None,
+        )
+
+    end_event.record()
+    torch.cuda.synchronize()
+
+    elapsed_ms = start_event.elapsed_time(end_event)
+    avg_latency_ms = elapsed_ms / benchmark_iters
+
+    # FLA baseline
+    fla_latency_ms = None
+    if num_qk_heads == num_v_heads:
+        fn_fla = lambda: fla_base(
+            q,
+            k,
+            v,
+            g,
+            beta,
+            None,
+            initial_state=h0,
+            output_final_state=output_final_state,
+        )
+
+        for _ in range(warmup_iters):
+            fn_fla()
+        torch.cuda.synchronize()
+
+        fla_start = torch.cuda.Event(enable_timing=True)
+        fla_end = torch.cuda.Event(enable_timing=True)
+        fla_start.record()
+        for _ in range(benchmark_iters):
+            fn_fla()
+        fla_end.record()
+        torch.cuda.synchronize()
+
+        fla_latency_ms = fla_start.elapsed_time(fla_end) / benchmark_iters
+
+    result = {
+        "batch_size": batch_size,
+        "seq_len": seq_len,
+        "num_qk_heads": num_qk_heads,
+        "num_v_heads": num_v_heads,
+        "head_dim": head_dim,
+        "dtype": str(dtype).split(".")[-1],
+        "gdn_ms": avg_latency_ms,
+    }
+
+    if fla_latency_ms is not None:
+        result["fla_ms"] = fla_latency_ms
+        result["speedup"] = (
+            fla_latency_ms / avg_latency_ms if avg_latency_ms > 0 else float("nan")
+        )
+
+    return result
+
+
+def benchmark_gdn_varlen(
+    cu_seqlens: List[int],
+    num_qk_heads: int,
+    num_v_heads: int,
+    head_dim: int,
+    dtype: torch.dtype = torch.bfloat16,
+    warmup_iters: int = 10,
+    benchmark_iters: int = 100,
+) -> dict:
+    """
+    Benchmark GDN with variable-length sequences.
+
+    Returns:
+        Dictionary with benchmark results.
+    """
+    device = "cuda"
+    torch.cuda.reset_peak_memory_stats()
+
+    total_len = cu_seqlens[-1]
+    num_seqs = len(cu_seqlens) - 1
+    cu_seqlens_tensor = torch.tensor(cu_seqlens, device=device)
+
+    # Create input tensors
+    q = torch.randn((1, total_len, num_qk_heads, head_dim), dtype=dtype, device=device)
+    k = F.normalize(
+        torch.randn(
+            1, total_len, num_qk_heads, head_dim, dtype=torch.float32, device=device
+        ),
+        p=2,
+        dim=-1,
+    ).to(dtype)
+    v = torch.randn((1, total_len, num_v_heads, head_dim), dtype=dtype, device=device)
+    g = F.logsigmoid(
+        torch.rand(1, total_len, num_v_heads, dtype=torch.float32, device=device)
+    )
+    beta = torch.rand(
+        1, total_len, num_v_heads, dtype=torch.float32, device=device
+    ).sigmoid()
+    h0 = torch.randn(
+        (num_seqs, num_v_heads, head_dim, head_dim), dtype=torch.float32, device=device
+    )
+
+    o = torch.zeros_like(v)
+    state_output = torch.zeros_like(h0, dtype=torch.float32)
+
+    # Warmup
+    for _ in range(warmup_iters):
+        chunk_gated_delta_rule(
+            q,
+            k,
+            v,
+            g,
+            beta,
+            None,
+            h0,
+            True,
+            cu_seqlens_tensor,
+            False,
+            o,
+            state_output,
+        )
+    torch.cuda.synchronize()
+
+    # Benchmark
+    torch.cuda.reset_peak_memory_stats()
+    start_event = torch.cuda.Event(enable_timing=True)
+    end_event = torch.cuda.Event(enable_timing=True)
+
+    o.zero_()
+    state_output.zero_()
+
+    start_event.record()
+    for _ in range(benchmark_iters):
+        chunk_gated_delta_rule(
+            q,
+            k,
+            v,
+            g,
+            beta,
+            None,
+            h0,
+            True,
+            cu_seqlens_tensor,
+            False,
+            o,
+            state_output,
+        )
+    end_event.record()
+    torch.cuda.synchronize()
+
+    elapsed_ms = start_event.elapsed_time(end_event)
+    avg_latency_ms = elapsed_ms / benchmark_iters
+
+    return {
+        "num_seqs": num_seqs,
+        "total_len": total_len,
+        "avg_seq_len": total_len // num_seqs,
+        "num_qk_heads": num_qk_heads,
+        "num_v_heads": num_v_heads,
+        "head_dim": head_dim,
+        "dtype": str(dtype).split(".")[-1],
+        "gdn_ms": avg_latency_ms,
+    }
+
+
+def benchmark_sweep(
+    warmup_iters: int = 10,
+    benchmark_iters: int = 100,
+    dtype: torch.dtype = torch.bfloat16,
+) -> None:
+    """Run a sweep of benchmark configurations for both fixlen and varlen."""
+    head_dim = 128
+
+    fixlen_configs = [
+        # (batch_size, seq_len, num_qk_heads, num_v_heads)
+        # Typical use cases:
+        (1, 512, 96, 96),
+        (1, 1024, 96, 96),
+        (1, 4096, 96, 96),
+        (1, 8192, 96, 96),
+        (9, 512, 32, 32),
+        (9, 1024, 32, 32),
+        (9, 4096, 32, 32),
+        (9, 8192, 32, 32),
+        (33, 512, 32, 32),
+        (33, 1024, 32, 32),
+        (33, 4096, 32, 32),
+        (33, 8192, 32, 32),
+        # Full SM utilization:
+        (1, 512, 148, 148),
+        (1, 1024, 148, 148),
+        (1, 4096, 148, 148),
+        (1, 8192, 148, 148),
+    ]
+
+    varlen_configs = [
+        # (cu_seqlens, num_qk_heads, num_v_heads) — derived from fixlen_configs
+        ([sl * i for i in range(bs + 1)], nqk, nv)
+        for bs, sl, nqk, nv in fixlen_configs
+    ]
+
+    print("\n" + "#" * 80)
+    print(" BENCHMARK SWEEP")
+    print("#" * 80)
+
+    fixlen_results = []
+    for i, (bs, sl, nqk, nv) in enumerate(fixlen_configs):
+        label = f"[{i + 1}/{len(fixlen_configs)}] bs={bs}, sl={sl}, nqk={nqk}, nv={nv}"
+        print(f"  Running fixlen  {label} ...", end="", flush=True)
+        try:
+            result = benchmark_gdn_fixlen(
+                batch_size=bs,
+                seq_len=sl,
+                num_qk_heads=nqk,
+                num_v_heads=nv,
+                head_dim=head_dim,
+                dtype=dtype,
+                warmup_iters=warmup_iters,
+                benchmark_iters=benchmark_iters,
+            )
+            fixlen_results.append(result)
+            msg = f"  GDN: {result['gdn_ms']:.3f} ms"
+            if "fla_ms" in result:
+                msg += f"  FLA: {result['fla_ms']:.3f} ms  ({result['speedup']:.2f}x)"
+            print(msg)
+        except Exception as e:
+            print(f"  FAILED: {e}")
+            torch.cuda.empty_cache()
+
+    varlen_results = []
+    for i, (cu_seqlens, nqk, nv) in enumerate(varlen_configs):
+        num_seqs = len(cu_seqlens) - 1
+        total_len = cu_seqlens[-1]
+        label = f"[{i + 1}/{len(varlen_configs)}] seqs={num_seqs}, total={total_len}, nqk={nqk}, nv={nv}"
+        print(f"  Running varlen  {label} ...", end="", flush=True)
+        try:
+            result = benchmark_gdn_varlen(
+                cu_seqlens=cu_seqlens,
+                num_qk_heads=nqk,
+                num_v_heads=nv,
+                head_dim=head_dim,
+                dtype=dtype,
+                warmup_iters=warmup_iters,
+                benchmark_iters=benchmark_iters,
+            )
+            varlen_results.append(result)
+            print(f"  GDN: {result['gdn_ms']:.3f} ms")
+        except Exception as e:
+            print(f"  FAILED: {e}")
+            torch.cuda.empty_cache()
+
+    print_results_table(fixlen_results, "Fixed-Length Sweep Results")
+    print_results_table(varlen_results, "Variable-Length Sweep Results")
+
+
+def print_results_table(results: List[dict], title: str = "Benchmark Results"):
+    """Print benchmark results in a formatted table."""
+    if not results:
+        return
+
+    print(f"\n{'=' * 80}")
+    print(f" {title}")
+    print(f"{'=' * 80}")
+
+    # Get all keys from first result
+    keys = list(results[0].keys())
+
+    # Calculate column widths
+    widths = {}
+    for key in keys:
+        max_len = len(key)
+        for r in results:
+            val = r[key]
+            if isinstance(val, float):
+                val_str = f"{val:.4f}" if val < 100 else f"{val:.2f}"
+            else:
+                val_str = str(val)
+            max_len = max(max_len, len(val_str))
+        widths[key] = max_len + 2
+
+    # Print header
+    header = " | ".join(f"{key:^{widths[key]}}" for key in keys)
+    print(header)
+    print("-" * len(header))
+
+    # Print rows
+    for r in results:
+        row_vals = []
+        for key in keys:
+            val = r[key]
+            if isinstance(val, float):
+                val_str = f"{val:.4f}" if val < 100 else f"{val:.2f}"
+            else:
+                val_str = str(val)
+            row_vals.append(f"{val_str:^{widths[key]}}")
+        print(" | ".join(row_vals))
+
+    print(f"{'=' * 80}\n")
+
+
+def main():
+    parser = argparse.ArgumentParser(description="GDN Performance Benchmark")
+    parser.add_argument("--batch-size", "-b", type=int, default=1, help="Batch size")
+    parser.add_argument(
+        "--seq-len", "-t", type=int, default=8192, help="Sequence length"
+    )
+    parser.add_argument(
+        "--num-qk-heads",
+        "-nqk",
+        type=int,
+        default=96,
+        help="Number of qk attention heads",
+    )
+    parser.add_argument(
+        "--num-v-heads", "-nv", type=int, default=96, help="Number of v attention heads"
+    )
+    parser.add_argument(
+        "--head-dim", "-d", type=int, default=128, help="Head dimension"
+    )
+    parser.add_argument("--warmup", type=int, default=10, help="Warmup iterations")
+    parser.add_argument("--iters", type=int, default=100, help="Benchmark iterations")
+    parser.add_argument(
+        "--varlen", action="store_true", help="Run variable-length benchmark"
+    )
+    parser.add_argument(
+        "--cu-seqlens",
+        type=int,
+        nargs="+",
+        default=None,
+        help="Cumulative sequence lengths for varlen mode (e.g. --cu-seqlens 0 512 1024 2048)",
+    )
+    parser.add_argument(
+        "--sweep",
+        action="store_true",
+        help="Run a sweep of many test cases (fixlen + varlen)",
+    )
+
+    args = parser.parse_args()
+
+    if args.head_dim != 128:
+        print(f"Error: head_dim must be 128, got {args.head_dim}")
+        sys.exit(1)
+
+    print("\n" + "=" * 60)
+    print(" GDN (Gated Delta Network) Linear Attention Benchmark")
+    print(" GPU:", torch.cuda.get_device_name(0))
+    print("=" * 60)
+
+    if args.sweep:
+        benchmark_sweep(
+            warmup_iters=args.warmup,
+            benchmark_iters=args.iters,
+        )
+        return
+
+    if args.varlen:
+        if args.cu_seqlens is not None:
+            cu_seqlens = args.cu_seqlens
+            if cu_seqlens[0] != 0:
+                print(f"Error: cu_seqlens must start with 0, got {cu_seqlens[0]}")
+                sys.exit(1)
+            if len(cu_seqlens) < 2:
+                print("Error: cu_seqlens must have at least 2 elements (e.g. 0 1024)")
+                sys.exit(1)
+        else:
+            cu_seqlens = [args.seq_len * i for i in range(args.batch_size + 1)]
+        print("\nRunning variable-length benchmark:")
+        print(f"  cu_seqlens: {cu_seqlens}")
+        print(f"  num_qk_heads: {args.num_qk_heads}")
+        print(f"  num_v_heads: {args.num_v_heads}")
+        print(f"  head_dim: {args.head_dim}")
+
+        result = benchmark_gdn_varlen(
+            cu_seqlens=cu_seqlens,
+            num_qk_heads=args.num_qk_heads,
+            num_v_heads=args.num_v_heads,
+            head_dim=args.head_dim,
+            warmup_iters=args.warmup,
+            benchmark_iters=args.iters,
+        )
+        print_results_table([result], "Variable-Length Benchmark")
+    else:
+        # Fixed-length benchmark
+        print("\nRunning fixed-length benchmark:")
+        print(f"  batch_size: {args.batch_size}")
+        print(f"  seq_len: {args.seq_len}")
+        print(f"  num_qk_heads: {args.num_qk_heads}")
+        print(f"  num_v_heads: {args.num_v_heads}")
+        print(f"  head_dim: {args.head_dim}")
+        print(f"  warmup: {args.warmup}")
+        print(f"  iters: {args.iters}")
+
+        result = benchmark_gdn_fixlen(
+            batch_size=args.batch_size,
+            seq_len=args.seq_len,
+            num_qk_heads=args.num_qk_heads,
+            num_v_heads=args.num_v_heads,
+            head_dim=args.head_dim,
+            warmup_iters=args.warmup,
+            benchmark_iters=args.iters,
+        )
+
+        print_results_table([result], "Fixed-Length Benchmark")
+
+
+if __name__ == "__main__":
+    main()
--- a/benchmarks/bench_blackwell_gdn_prefill.py
+++ b/benchmarks/bench_blackwell_gdn_prefill.py
@@ -436,19 +436,19 @@ def print_results_table(results: List[dict], title: str = "Benchmark Results"):
 
 def main():
     parser = argparse.ArgumentParser(description="GDN Performance Benchmark")
-    parser.add_argument("--batch-size", "-b", type=int, default=1, help="Batch size")
+    parser.add_argument("--batch-size", "-b", type=int, default=4, help="Batch size")
     parser.add_argument(
-        "--seq-len", "-t", type=int, default=8192, help="Sequence length"
+        "--seq-len", "-t", type=int, default=4096, help="Sequence length"
     )
     parser.add_argument(
         "--num-qk-heads",
         "-nqk",
         type=int,
-        default=96,
+        default=32,
         help="Number of qk attention heads",
     )
     parser.add_argument(
-        "--num-v-heads", "-nv", type=int, default=96, help="Number of v attention heads"
+        "--num-v-heads", "-nv", type=int, default=32, help="Number of v attention heads"
     )
     parser.add_argument(
         "--head-dim", "-d", type=int, default=128, help="Head dimension"
--- a/benchmarks/bench_blackwell_gdn_prefill.py
+++ b/benchmarks/bench_blackwell_gdn_prefill.py
@@ -32,7 +32,7 @@ from typing import List
 import torch
 import torch.nn.functional as F
 
-# Blacwell GDN prefill
+# Blackwell GDN prefill
 from flashinfer.gdn_kernels.blackwell_prefill.gdn import chunk_gated_delta_rule
 
 # FLA baseline
--- a/benchmarks/bench_blackwell_gdn_prefill.py
+++ b/benchmarks/bench_blackwell_gdn_prefill.py
@@ -33,6 +33,7 @@ import torch
 import torch.nn.functional as F
 import numpy as np
 from flashinfer.testing import bench_gpu_time
+from flashinfer.utils import is_sm100a_supported, is_sm110a_supported
 
 # Blackwell GDN prefill
 from flashinfer.gdn_prefill import chunk_gated_delta_rule
@@ -426,6 +427,11 @@ def main():
     )
 
     args = parser.parse_args()
+
+    device = torch.device("cuda")
+    if not (is_sm100a_supported(device) or is_sm110a_supported(device)):
+        print("Error: bench_blackwell_gdn_prefill.py requires an SM100A/SM110A GPU.")
+        sys.exit(1)
 
     if args.head_dim != 128:
         print(f"Error: head_dim must be 128, got {args.head_dim}")
--- a/benchmarks/bench_blackwell_gdn_prefill.py
+++ b/benchmarks/bench_blackwell_gdn_prefill.py
@@ -33,13 +33,18 @@ import torch
 import torch.nn.functional as F
 import numpy as np
 from flashinfer.testing import bench_gpu_time
-from flashinfer.utils import is_sm100a_supported, is_sm110a_supported
+from flashinfer.utils import get_compute_capability
 
 # Blackwell GDN prefill
 from flashinfer.gdn_prefill import chunk_gated_delta_rule
 
 # FLA baseline
 from fla.ops.gated_delta_rule.chunk import chunk_gated_delta_rule_fwd as fla_base
+
+
+def check_compute_capability(device: torch.device) -> bool:
+    major, minor = get_compute_capability(device)
+    return (major, minor) in ((10, 0), (10, 3), (11, 0))
 
 
 def benchmark_gdn_fixlen(
@@ -429,8 +434,11 @@ def main():
     args = parser.parse_args()
 
     device = torch.device("cuda")
-    if not (is_sm100a_supported(device) or is_sm110a_supported(device)):
-        print("Error: bench_blackwell_gdn_prefill.py requires an SM100A/SM110A GPU.")
+    is_device_support = check_compute_capability(device)
+    if not is_device_support:
+        print(
+            "Error: bench_blackwell_gdn_prefill.py requires an SM100/SM103/SM110 GPU."
+        )
         sys.exit(1)
 
     if args.head_dim != 128:
--- a/flashinfer/gdn_kernels/__init__.py
+++ b/flashinfer/gdn_kernels/__init__.py
@@ -48,12 +48,6 @@ except ImportError:
     get_vec_size_mtp = None  # type: ignore
 
 
-try:
-    from .blackwell_prefill.gdn import chunk_gated_delta_rule
-except ImportError:
-    chunk_gated_delta_rule = None  # type: ignore
-
-
 __all__ = [
     "gated_delta_rule",
     "GatedDeltaRuleKernel",
@@ -62,5 +56,4 @@ __all__ = [
     "run_mtp_decode",
     "get_tile_v_mtp",
     "get_vec_size_mtp",
-    "chunk_gated_delta_rule",
 ]
--- a/flashinfer/gdn_kernels/__init__.py
+++ b/flashinfer/gdn_kernels/__init__.py
@@ -47,6 +47,13 @@ except ImportError:
     get_tile_v_mtp = None  # type: ignore
     get_vec_size_mtp = None  # type: ignore
 
+
+try:
+    from .blackwell_prefill.gdn import chunk_gated_delta_rule
+except ImportError:
+    chunk_gated_delta_rule = None  # type: ignore
+
+
 __all__ = [
     "gated_delta_rule",
     "GatedDeltaRuleKernel",
@@ -55,4 +62,5 @@ __all__ = [
     "run_mtp_decode",
     "get_tile_v_mtp",
     "get_vec_size_mtp",
+    "chunk_gated_delta_rule",
 ]
--- a/flashinfer/gdn_kernels/blackwell_prefill/gdn.py
+++ b/flashinfer/gdn_kernels/blackwell_prefill/gdn.py
@@ -3782,12 +3782,18 @@ class GDN:
         initial_state_f32_iter: Optional[cute.Pointer],
         state_output: Optional[cute.Pointer],
         scale: Optional[float],
-        cum_seqlen_q: Optional[cute.Tensor] = None,
+        cum_seqlen: Optional[cute.Pointer] = None,
         stream: cuda.CUstream = None,
     ):
         """Host-side entry: build tensor layouts, TMA descriptors, smem storage, and launch the kernel."""
 
         b, s_q, s_sum, h_q, h_v, d = problem_size
+
+        cum_seqlen_q = (
+            None
+            if cum_seqlen is None
+            else cute.make_tensor(cum_seqlen, cute.make_layout((b)))
+        )
 
         h_r = h_v // h_q
         o_offset = 0 if cum_seqlen_q is None else (-s_q * d * h_r * h_q)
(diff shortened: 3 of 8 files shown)


flashinfer/gdn_kernels/__init__.py
```diff
@@ -47,6 +47,13 @@
     get_tile_v_mtp = None  # type: ignore
     get_vec_size_mtp = None  # type: ignore
 
+
+try:
+    from .blackwell_prefill.gdn import chunk_gated_delta_rule
+except ImportError:
+    chunk_gated_delta_rule = None  # type: ignore
```
hlu1: Can you call this function on SM100 in https://github.com/flashinfer-ai/flashinfer/blob/main/flashinfer/gdn_prefill.py#L86? The `chunk_gated_delta_rule` function in this file should be the only API for the user to call the prefill gdn kernel. It only supports SM90 on the main branch.

dianzhangchen: @hlu1 sure, thanks for pointing this out

tests/gdn/test_gdn_prefill_blackwell.py
```diff
@@ -0,0 +1,728 @@
...
+        h0 = torch.randn((B, H, D, D), dtype=torch.float)
+
+        q, k, v, beta, g, h0 = map(lambda x: x.to(device), (q, k, v, beta, g, h0))
+
+        results = []
+        for _run_idx in range(3):
+            state_output = torch.zeros_like(h0, dtype=torch.float)
+
+            o, state_output = chunk_gated_delta_rule(
+                q.clone(),
```
yzh119: Why do we need these clones here?

tests/gdn/test_gdn_prefill_blackwell.py
```diff
@@ -0,0 +1,728 @@
...
+            beta.clone(),
+            None,
+            h0.clone(),
+            True,
+            cu_seqlens_tensor.clone(),
+            False,
+            None,
+            state_output,
+        )
+        torch.cuda.synchronize()
```
yzh119: and the synchronization

tests/gdn/test_gdn_prefill_blackwell.py
```diff
@@ -0,0 +1,728 @@
...
+                initial_state=h0_v_major[i].clone(),
+                output_final_state=True,
+            )
+            if i == 0:
+                ref_ht = ref_ht_i
+            else:
+                ref_ht = torch.cat((ref_ht, ref_ht_i), dim=0).contiguous()
+
+        ref_ht = torch.transpose(ref_ht, -1, -2).contiguous()
+        torch.cuda.synchronize()
```
yzh119: ditto

tests/gdn/test_gdn_prefill_blackwell.py
```diff
@@ -0,0 +1,728 @@
...
+            beta.clone(),
+            None,
+            h0.clone(),
+            True,
+            None,
+            False,
+            None,
+            state_output,
+        )
+        torch.cuda.synchronize()
```
yzh119: ditto

tests/gdn/test_gdn_prefill_blackwell.py
```diff
@@ -0,0 +1,728 @@
...
+        g = F.logsigmoid(torch.rand(1, T * B, H_V, dtype=torch.float32))
+        beta = torch.rand(1, T * B, H_V, dtype=torch.float32).sigmoid()
+        h0 = torch.randn((B, H_V, D, D), dtype=torch.float)
+
+        q, k, v, beta, g, h0 = map(lambda x: x.to(device), (q, k, v, beta, g, h0))
+
+        state_output = torch.zeros_like(h0, dtype=torch.float)
+
+        o, state_output = chunk_gated_delta_rule(
+            q.clone(),
```
yzh119: ditto

tests/gdn/test_gdn_prefill_blackwell.py
```diff
@@ -0,0 +1,728 @@
...
+            beta.clone(),
+            None,
+            h0.clone(),
+            True,
+            None,
+            False,
+            None,
+            state_output,
+        )
+        torch.cuda.synchronize()
```
yzh119: dito

tests/gdn/test_gdn_prefill_blackwell.py
```diff
@@ -0,0 +1,728 @@
...
+    scale: float = None,
+    initial_state: torch.Tensor = None,
+    output_final_state: bool = False,
+):
+    """Reference implementation of gated delta rule (recurrent version)."""
+    q, k, v, beta, g = map(
+        lambda x: x.transpose(1, 2).contiguous().to(torch.float32), [q, k, v, beta, g]
+    )
+    B, H, T, K, V = *k.shape, v.shape[-1]
+    o = torch.zeros(B, H, T, V).to(v)
```
yzh119, line 59: it's much faster to directly create a GPU tensor than moving tensors from CPU to GPU, especially in unittests.


```suggestion
    o = torch.zeros(B, H, T, V, device=v.device)
```

tests/gdn/test_gdn_prefill_blackwell.py
```diff
@@ -0,0 +1,728 @@
...
+    v: torch.Tensor,
+    beta: torch.Tensor,
+    g: torch.Tensor,
+    scale: float = None,
+    initial_state: torch.Tensor = None,
+    output_final_state: bool = False,
+):
+    """Reference implementation of gated delta rule (recurrent version)."""
+    q, k, v, beta, g = map(
+        lambda x: x.transpose(1, 2).contiguous().to(torch.float32), [q, k, v, beta, g]
```
yzh119, line 56: it will be much faster if we can pass gpu tensors direclty.

tests/gdn/test_gdn_prefill_blackwell.py
```diff
@@ -0,0 +1,728 @@
...
+            (3, 128, 131, 177, testtype),
+        ],
+    )
+    def test_fixlen_output(self, set_seed, H, D, T, B, dtype):
+        """Test fixed-length GDN output correctness."""
+        _skip_if_gdn_blackwell_prefill_not_available()
+
+        device = "cuda"
+
+        q = torch.randn((B, T, H, D), dtype=dtype)
```
yzh119: create these tensors on GPU

benchmarks/bench_blackwell_gdn_prefill.py
```diff
@@ -0,0 +1,543 @@
...
+            None,
+            False,
+            o,
+            None,
+        )
+    torch.cuda.synchronize()
+
+    # Benchmark
+    torch.cuda.reset_peak_memory_stats()
+    start_event = torch.cuda.Event(enable_timing=True)
```
yzh119: Please consider using https://docs.flashinfer.ai/generated/flashinfer.testing.bench_gpu_time.html#flashinfer.testing.bench_gpu_time for benchmarking (and enable cupti), see https://github.com/flashinfer-ai/flashinfer/blob/main/.claude/skills/benchmark-kernel/SKILL.md for details.

flashinfer/gdn_kernels/__init__.py
dianzhangchen: @hlu1 done，change the api to gdn_prefill

tests/gdn/test_gdn_prefill_blackwell.py
dianzhangchen: @yzh119 thanks for pointing this out. For unit tests, we do not trust the code under test. Therefore, we cannot guarantee whether these tensors have been incorrectly modified by the function being tested, So i use clones.

And i have remove these clones.

tests/gdn/test_gdn_prefill_blackwell.py
dianzhangchen: thanks for pointing this out, done, remove synchronization

tests/gdn/test_gdn_prefill_blackwell.py
dianzhangchen: ditto  and done

tests/gdn/test_gdn_prefill_blackwell.py
dianzhangchen: ditto  and done

tests/gdn/test_gdn_prefill_blackwell.py
dianzhangchen: ditto  and done

tests/gdn/test_gdn_prefill_blackwell.py
dianzhangchen: ditto and done

tests/gdn/test_gdn_prefill_blackwell.py
dianzhangchen: thanks for pointing this out, done, q k v is  GPU tensor

tests/gdn/test_gdn_prefill_blackwell.py
dianzhangchen: thanks for pointing this out, done, q k v is GPU tensor

tests/gdn/test_gdn_prefill_blackwell.py
dianzhangchen: ditto and done

benchmarks/bench_blackwell_gdn_prefill.py
dianzhangchen: @yzh119  thanks for pointing this out, done, use bench_gpu_time func

kaixih (commented):
I know this PR is focused on SM100+ prefill support, but just want to clarify: since this prefill kernel outputs FP32 state while the fast decode path requires BF16 state, users will still need an explicit dtype conversion between prefill and decode until [gap 5.4](https://github.com/flashinfer-ai/flashinfer/issues/2687) (prefill BF16 state) is addressed. Is this a known limitation or something we plan to tackle soon?

dianzhangchen:
> I know this PR is focused on SM100+ prefill support, but just want to clarify: since this prefill kernel outputs FP32 state while the fast decode path requires BF16 state, users will still need an explicit dtype conversion between prefill and decode until [gap 5.4](https://github.com/flashinfer-ai/flashinfer/issues/2687) (prefill BF16 state) is addressed. Is this a known limitation or something we plan to tackle soon?

hi @kaixih , this is a known limitation and will be addressed in the next version.

vadiklyutiy:
What is preventing us from merging this PR?

flashinfer/gdn_prefill.py
```diff
@@ -211,3 +173,136 @@ def chunk_gated_delta_rule(
         return output, output_state
     else:
         return output
+
+
+@supported_compute_capability([90, 100, 110])
```
ZJY0516: Does this support B300(sm103)?

yzh119: Yes this should support sm103

yzh119: Does https://github.com/flashinfer-ai/flashinfer/pull/2742/commits/03233c07276caacd9b60abfe76152f01b9fcabe2 work for you?

ZJY0516: Thanks for replying. Could we please update `@supported_compute_capability` to include 103 as well?

flashinfer/gdn_prefill.py
```diff
@@ -175,7 +181,7 @@ def chunk_gated_delta_rule_hopper(
         return output
 
 
-@supported_compute_capability([90, 100, 103, 110])
```
yzh119: Why do we remove 110 here?

flashinfer/gdn_prefill.py
dianzhangchen: hi @ZJY0516, i updated the code. It should work now. @yzh119 Thanks for the revision. Since sm103 is supported in CUDA 12.8, I added `is_sm103a_supported` instead of `is_sm100f_supported`, which requires CUDA 12.9

flashinfer/gdn_prefill.py
dianzhangchen: I haven’t tested sm110 yet, so support is not confirmed.

ZJY0516:
I find that this PR does not perform well for small head configurations.

```
GPU: NVIDIA B300 SXM6 AC
Model reference: Qwen3.5-397B-A17B (h_k=16, h_v=64, d=128, GVA ratio=4)

Config                          h_qk  h_v  FlashInfer  FLA/Triton   Speedup
───────────────────────────────────────────────────────────────────────────
TP4  B=1  S=8192                   4   16      0.748ms      0.395ms     0.53x ✗
TP4  B=1  S=4096                   4   16      0.388ms      0.213ms     0.55x ✗
TP4  B=1  S=2048                   4   16      0.207ms      0.119ms     0.58x ✗
TP2  B=1  S=8192                   8   32      0.749ms      0.566ms     0.76x ✗
TP1  B=1  S=8192                  16   64      0.747ms      0.971ms     1.30x ✓
Symmetric B=4 S=4096              32   32      0.411ms      1.095ms     2.67x ✓
```

```python
import torch
import torch.nn.functional as F
import numpy as np
from flashinfer.testing import bench_gpu_time
from flashinfer.gdn_prefill import chunk_gated_delta_rule as fi_gdn
from fla.ops.gated_delta_rule.chunk import chunk_gated_delta_rule_fwd as fla_gdn

CONFIGS = [
    # (batch, seq_len, h_qk, h_v, d, label)
    # Qwen3.5-397B (h_k=16, h_v=64, d=128) under different TP
    (1, 8192, 4, 16, 128, "TP4  B=1  S=8192"),
    (1, 4096, 4, 16, 128, "TP4  B=1  S=4096"),
    (1, 2048, 4, 16, 128, "TP4  B=1  S=2048"),
    (1, 8192, 8, 32, 128, "TP2  B=1  S=8192"),
    (1, 8192, 16, 64, 128, "TP1  B=1  S=8192"),
    # Original benchmark config (symmetric heads)
    (4, 4096, 32, 32, 128, "Symmetric B=4 S=4096"),
]

WARMUP = 5
ITERS = 20


def bench_flashinfer(B, T, h_qk, h_v, d):
    device = "cuda"
    dtype = torch.float16
    q = torch.randn((B, T, h_qk, d), dtype=dtype, device=device)
    k = F.normalize(
        torch.randn(B, T, h_qk, d, dtype=torch.float32, device=device), p=2, dim=-1
    ).to(dtype)
    v = torch.randn((B, T, h_v, d), dtype=dtype, device=device)
    g = F.logsigmoid(torch.rand(1, T * B, h_v, dtype=torch.float32, device=device))
    beta = torch.rand(1, T * B, h_v, dtype=torch.float32, device=device).sigmoid()
    h0 = torch.randn((B, h_v, d, d), dtype=torch.float32, device=device)
    state_out = torch.zeros_like(h0)

    fn = lambda: fi_gdn(q, k, v, g, beta, None, h0, True, None, False, None, state_out)
    times = bench_gpu_time(fn, enable_cupti=True, dry_run_iters=WARMUP, repeat_iters=ITERS)
    torch.cuda.empty_cache()
    return np.average(times)


def bench_fla(B, T, h_qk, h_v, d):
    device = "cuda"
    dtype = torch.float16
    # FLA doesn't support GVA (h_qk != h_v), expand q/k to h_v
    h = h_v
    q = torch.randn((B, T, h, d), dtype=dtype, device=device)
    k = F.normalize(
        torch.randn(B, T, h, d, dtype=torch.float32, device=device), p=2, dim=-1
    ).to(dtype)
    v = torch.randn((B, T, h_v, d), dtype=dtype, device=device)
    g = F.logsigmoid(torch.rand(1, T * B, h_v, dtype=torch.float32, device=device))
    beta = torch.rand(1, T * B, h_v, dtype=torch.float32, device=device).sigmoid()
    h0 = torch.randn((B, h_v, d, d), dtype=torch.float32, device=device)

    fn = lambda: fla_gdn(q, k, v, g, beta, None, initial_state=h0, output_final_state=True)
    times = bench_gpu_time(fn, enable_cupti=True, dry_run_iters=WARMUP, repeat_iters=ITERS)
    torch.cuda.empty_cache()
    return np.average(times)


def main():
    print(f"\nGPU: {torch.cuda.get_device_name(0)}")
    print(f"Model reference: Qwen3.5-397B-A17B (h_k=16, h_v=64, d=128, GVA ratio=4)")
    print()
    header = f"{'Config':<30s}  {'h_qk':>4s} {'h_v':>4s}  {'FlashInfer':>10s}  {'FLA/Triton':>10s}  {'Speedup':>8s}"
    print(header)
    print("─" * len(header))

    for B, T, h_qk, h_v, d, label in CONFIGS:
        fi_ms = bench_flashinfer(B, T, h_qk, h_v, d)
        fla_ms = bench_fla(B, T, h_qk, h_v, d)
        speedup = fla_ms / fi_ms
        marker = "✓" if speedup > 1.0 else "✗"
        print(
            f"{label:<30s}  {h_qk:>4d} {h_v:>4d}  {fi_ms:>9.3f}ms  {fla_ms:>9.3f}ms  {speedup:>7.2f}x {marker}"
        )

if __name__ == "__main__":
    main()
```

dianzhangchen:
> I find that this PR does not perform well for small head configurations.
> 
> ```
> GPU: NVIDIA B300 SXM6 AC
> Model reference: Qwen3.5-397B-A17B (h_k=16, h_v=64, d=128, GVA ratio=4)
> 
> Config                          h_qk  h_v  FlashInfer  FLA/Triton   Speedup
> ───────────────────────────────────────────────────────────────────────────
> TP4  B=1  S=8192                   4   16      0.748ms      0.395ms     0.53x ✗
> TP4  B=1  S=4096                   4   16      0.388ms      0.213ms     0.55x ✗
> TP4  B=1  S=2048                   4   16      0.207ms      0.119ms     0.58x ✗
> TP2  B=1  S=8192                   8   32      0.749ms      0.566ms     0.76x ✗
> TP1  B=1  S=8192                  16   64      0.747ms      0.971ms     1.30x ✓
> Symmetric B=4 S=4096              32   32      0.411ms      1.095ms     2.67x ✓
> ```
> 
> ```python
> import torch
> import torch.nn.functional as F
> import numpy as np
> from flashinfer.testing import bench_gpu_time
> from flashinfer.gdn_prefill import chunk_gated_delta_rule as fi_gdn
> from fla.ops.gated_delta_rule.chunk import chunk_gated_delta_rule_fwd as fla_gdn
> 
> CONFIGS = [
>     # (batch, seq_len, h_qk, h_v, d, label)
>     # Qwen3.5-397B (h_k=16, h_v=64, d=128) under different TP
>     (1, 8192, 4, 16, 128, "TP4  B=1  S=8192"),
>     (1, 4096, 4, 16, 128, "TP4  B=1  S=4096"),
>     (1, 2048, 4, 16, 128, "TP4  B=1  S=2048"),
>     (1, 8192, 8, 32, 128, "TP2  B=1  S=8192"),
>     (1, 8192, 16, 64, 128, "TP1  B=1  S=8192"),
>     # Original benchmark config (symmetric heads)
>     (4, 4096, 32, 32, 128, "Symmetric B=4 S=4096"),
> ]
> 
> WARMUP = 5
> ITERS = 20
> 
> 
> def bench_flashinfer(B, T, h_qk, h_v, d):
>     device = "cuda"
>     dtype = torch.float16
>     q = torch.randn((B, T, h_qk, d), dtype=dtype, device=device)
>     k = F.normalize(
>         torch.randn(B, T, h_qk, d, dtype=torch.float32, device=device), p=2, dim=-1
>     ).to(dtype)
>     v = torch.randn((B, T, h_v, d), dtype=dtype, device=device)
>     g = F.logsigmoid(torch.rand(1, T * B, h_v, dtype=torch.float32, device=device))
>     beta = torch.rand(1, T * B, h_v, dtype=torch.float32, device=device).sigmoid()
>     h0 = torch.randn((B, h_v, d, d), dtype=torch.float32, device=device)
>     state_out = torch.zeros_like(h0)
> 
>     fn = lambda: fi_gdn(q, k, v, g, beta, None, h0, True, None, False, None, state_out)
>     times = bench_gpu_time(fn, enable_cupti=True, dry_run_iters=WARMUP, repeat_iters=ITERS)
>     torch.cuda.empty_cache()
>     return np.average(times)
> 
> 
> def bench_fla(B, T, h_qk, h_v, d):
>     device = "cuda"
>     dtype = torch.float16
>     # FLA doesn't support GVA (h_qk != h_v), expand q/k to h_v
>     h = h_v
>     q = torch.randn((B, T, h, d), dtype=dtype, device=device)
>     k = F.normalize(
>         torch.randn(B, T, h, d, dtype=torch.float32, device=device), p=2, dim=-1
>     ).to(dtype)
>     v = torch.randn((B, T, h_v, d), dtype=dtype, device=device)
>     g = F.logsigmoid(torch.rand(1, T * B, h_v, dtype=torch.float32, device=device))
>     beta = torch.rand(1, T * B, h_v, dtype=torch.float32, device=device).sigmoid()
>     h0 = torch.randn((B, h_v, d, d), dtype=torch.float32, device=device)
> 
>     fn = lambda: fla_gdn(q, k, v, g, beta, None, initial_state=h0, output_final_state=True)
>     times = bench_gpu_time(fn, enable_cupti=True, dry_run_iters=WARMUP, repeat_iters=ITERS)
>     torch.cuda.empty_cache()
>     return np.average(times)
> 
> 
> def main():
>     print(f"\nGPU: {torch.cuda.get_device_name(0)}")
>     print(f"Model reference: Qwen3.5-397B-A17B (h_k=16, h_v=64, d=128, GVA ratio=4)")
>     print()
>     header = f"{'Config':<30s}  {'h_qk':>4s} {'h_v':>4s}  {'FlashInfer':>10s}  {'FLA/Triton':>10s}  {'Speedup':>8s}"
>     print(header)
>     print("─" * len(header))
> 
>     for B, T, h_qk, h_v, d, label in CONFIGS:
>         fi_ms = bench_flashinfer(B, T, h_qk, h_v, d)
>         fla_ms = bench_fla(B, T, h_qk, h_v, d)
>         speedup = fla_ms / fi_ms
>         marker = "✓" if speedup > 1.0 else "✗"
>         print(
>             f"{label:<30s}  {h_qk:>4d} {h_v:>4d}  {fi_ms:>9.3f}ms  {fla_ms:>9.3f}ms  {speedup:>7.2f}x {marker}"
>         )
> 
> if __name__ == "__main__":
>     main()
> ```

hi @ZJY0516 , this is a known issue, as mentioned in the PR: "In the current version, we do not parallelize over the sequence-length dimension, so performance is limited when both batch size and head count are small. In such cases, only part of the 148 SMs are used, so the GPU is not fully utilized."

arpera:
What is the current status of this PR? Is it going to be merged soon, or work in progress still?

dianzhangchen:
> I find that this PR does not perform well for small head configurations.


Hi @ZJY0516 , i rerun this code on B200 and B300 — here are the results. One tip: use the `cu13` version of cutedsl. 

again as mentioned in the PR: "In the current version, we do not parallelize over the sequence-length dimension, so performance is limited when both batch size and head count are small. In such cases, only part of the 148 SMs are used, so the GPU is not fully utilized."

**GPU: NVIDIA B200**

```shell
GPU: NVIDIA B200
Model reference: Qwen3.5-397B-A17B (h_k=16, h_v=64, d=128, GVA ratio=4)

Config                          h_qk  h_v  FlashInfer  FLA/Triton   Speedup
───────────────────────────────────────────────────────────────────────────
TP4  B=1  S=8192                   4   16      0.559ms      0.437ms     0.78x ✗
TP4  B=1  S=4096                   4   16      0.286ms      0.306ms     1.07x ✓
TP4  B=1  S=2048                   4   16      0.147ms      0.241ms     1.64x ✓
TP2  B=1  S=8192                   8   32      0.562ms      0.593ms     1.06x ✓
TP1  B=1  S=8192                  16   64      0.569ms      1.015ms     1.78x ✓
Symmetric B=4 S=4096              32   32      0.299ms      1.139ms     3.81x ✓
```


**GPU: NVIDIA B300 SXM6 AC**

```shell
GPU: NVIDIA B300 SXM6 AC
Model reference: Qwen3.5-397B-A17B (h_k=16, h_v=64, d=128, GVA ratio=4)

Config                          h_qk  h_v  FlashInfer  FLA/Triton   Speedup
───────────────────────────────────────────────────────────────────────────
TP4  B=1  S=8192                   4   16      0.623ms      0.423ms     0.68x ✗
TP4  B=1  S=4096                   4   16      0.322ms      0.303ms     0.94x ✗
TP4  B=1  S=2048                   4   16      0.170ms      0.241ms     1.41x ✓
TP2  B=1  S=8192                   8   32      0.632ms      0.578ms     0.92x ✗
TP1  B=1  S=8192                  16   64      0.642ms      0.978ms     1.52x ✓
Symmetric B=4 S=4096              32   32      0.338ms      1.101ms     3.25x ✓
```

ZJY0516:
> In the current version, we do not parallelize over the sequence-length dimension, so performance is limited when both batch size and head count are small

@dianzhangchen Thanks for replying. Any plan to optimize it?

dianzhangchen:
> What is the current status of this PR? Is it going to be merged soon, or work in progress still?

hi @arpera , this PR should be ready to merge soon — just waiting for the unit tests to pass and the review to be approved. thanks.

dianzhangchen:
hi, @yzh119 I don't have permission to trigger the CI job via `/bot run`. Could you please help grant access? thanks!

dianzhangchen:
> Thanks for replying. Any plan to optimize it?

hi @ZJY0516 , There’s no plan to optimize it in the current version for now. If this is indeed a real requirement, we can evaluate whether to support it in a future update.

vadiklyutiy:
Actually `B=1` `S=8192` is one of the most frequently real life use case. Every prompt longer than 8192 will be chunked by 8192 pieces and processes one by one. 
Without good perf for it vllm can't use FI's GDN.    

@ZJY0516 Do you know current Hopper version has the same problem?

vadiklyutiy:
And one more thing
Data from @ZJY0516 https://github.com/flashinfer-ai/flashinfer/pull/2742#issuecomment-4128400827 and @dianzhangchen https://github.com/flashinfer-ai/flashinfer/pull/2742#issuecomment-4159774822 are not consistent. For B300 `TP4  B=1  S=4096` showed  0.213ms vs 0.303ms

ZJY0516:
> Actually `B=1` `S=8192` is one of the most frequently real life use case. Every prompt longer than 8192 will be chunked by 8192 pieces and processes one by one. Without good perf for it vllm can't use FI's GDN.
> 
> @ZJY0516 Do you know current Hopper version has the same problem?

I think hopper kernel of flashinfer is very efficient due to parallelism and warm specialization

vadiklyutiy:
> > Actually `B=1` `S=8192` is one of the most frequently real life use case. Every prompt longer than 8192 will be chunked by 8192 pieces and processes one by one. Without good perf for it vllm can't use FI's GDN.
> > @ZJY0516 Do you know current Hopper version has the same problem?
> 
> I think hopper kernel of flashinfer is very efficient due to parallelism and warm specialization

Did you test it with batch=1?

vadiklyutiy:
I figured out one more problematic thing. 
Seems that implementation try to tune kernel for every seq_len. If I run S=8192 I get the compilation for 8192 only. If after I run with 8191, I get the new compilation. 
@dianzhangchen Could you confirm, is it correct?

vadiklyutiy:
I expand a bit bench script above: add all Qwen3.5 shapes and add Blackwell support. 

Results for H200 looks very good - FI's always better. 

```
$ python gdn_prefill_bench.py

GPU: NVIDIA H200 [Hopper (SM90)]
Models: Qwen3.5 family (397B, 122B, 35B, 27B, 9B, 4B, 2B, 0.8B), d=128

Heads            Seqlens           h_qk  h_v        FI Hopper (SM90)  FLA/Triton   Speedup
------------------------------------------------------------------------------------------
397B/122B TP8    1x8192               2    8                  0.399ms      0.558ms     1.40x +
397B/122B TP8    1x4096               2    8                  0.209ms      0.458ms     2.19x +
397B/122B TP8    1x2048               2    8                  0.115ms      0.439ms     3.81x +
397B/122B TP8    6144+2048            2    8                  0.302ms      0.509ms     1.68x +
397B/122B TP8    4096+4096            2    8                  0.208ms      0.478ms     2.30x +
397B/122B TP8    2048+6144            2    8                  0.304ms      0.508ms     1.67x +
397B/122B TP8    1024+7168            2    8                  0.351ms      0.534ms     1.52x +
397B/122B TP8    2048x4               2    8                  0.115ms      0.464ms     4.02x +
397B/122B TP8    1024x8               2    8                  0.071ms      0.463ms     6.54x +

397B/122B TP4    1x8192               4   16                  0.397ms      0.594ms     1.50x +
397B/122B TP4    1x4096               4   16                  0.209ms      0.475ms     2.27x +
397B/122B TP4    1x2048               4   16                  0.115ms      0.453ms     3.93x +
397B/122B TP4    6144+2048            4   16                  0.304ms      0.549ms     1.80x +
397B/122B TP4    4096+4096            4   16                  0.211ms      0.515ms     2.44x +
397B/122B TP4    2048+6144            4   16                  0.303ms      0.549ms     1.81x +
397B/122B TP4    1024+7168            4   16                  0.350ms      0.572ms     1.64x +
397B/122B TP4    2048x4               4   16                  0.117ms      0.512ms     4.37x +
397B/122B TP4    1024x8               4   16                  0.075ms      0.519ms     6.96x +

397B/122B TP2    1x8192               8   32                  0.399ms      0.707ms     1.77x +
397B/122B TP2    1x4096               8   32                  0.210ms      0.512ms     2.44x +
397B/122B TP2    1x2048               8   32                  0.116ms      0.467ms     4.02x +
397B/122B TP2    6144+2048            8   32                  0.306ms      0.714ms     2.33x +
397B/122B TP2    4096+4096            8   32                  0.213ms      0.707ms     3.32x +
397B/122B TP2    2048+6144            8   32                  0.305ms      0.713ms     2.34x +
397B/122B TP2    1024+7168            8   32                  0.357ms      0.710ms     1.99x +
397B/122B TP2    2048x4               8   32                  0.130ms      0.718ms     5.51x +
397B/122B TP2    1024x8               8   32                  0.132ms      0.737ms     5.57x +

397B/122B TP1    1x8192              16   64                  0.401ms      1.156ms     2.88x +
397B/122B TP1    1x4096              16   64                  0.211ms      0.652ms     3.09x +
397B/122B TP1    1x2048              16   64                  0.117ms      0.500ms     4.26x +
397B/122B TP1    6144+2048           16   64                  0.311ms      1.164ms     3.75x +
397B/122B TP1    4096+4096           16   64                  0.217ms      1.161ms     5.35x +
397B/122B TP1    2048+6144           16   64                  0.310ms      1.166ms     3.75x +
397B/122B TP1    1024+7168           16   64                  0.360ms      1.164ms     3.23x +
397B/122B TP1    2048x4              16   64                  0.227ms      1.179ms     5.20x +
397B/122B TP1    1024x8              16   64                  0.242ms      1.206ms     4.99x +

35B/9B/4B TP1    1x8192              16   32                  0.400ms      0.705ms     1.76x +
35B/9B/4B TP1    1x4096              16   32                  0.210ms      0.510ms     2.43x +
35B/9B/4B TP1    1x2048              16   32                  0.116ms      0.467ms     4.01x +
35B/9B/4B TP1    6144+2048           16   32                  0.306ms      0.711ms     2.33x +
35B/9B/4B TP1    4096+4096           16   32                  0.213ms      0.712ms     3.34x +
35B/9B/4B TP1    2048+6144           16   32                  0.306ms      0.712ms     2.33x +
35B/9B/4B TP1    1024+7168           16   32                  0.356ms      0.714ms     2.01x +
35B/9B/4B TP1    2048x4              16   32                  0.125ms      0.717ms     5.71x +
35B/9B/4B TP1    1024x8              16   32                  0.133ms      0.738ms     5.55x +

27B TP1          1x8192              16   48                  0.401ms      0.944ms     2.35x +
27B TP1          1x4096              16   48                  0.211ms      0.575ms     2.72x +
27B TP1          1x2048              16   48                  0.117ms      0.485ms     4.13x +
27B TP1          6144+2048           16   48                  0.308ms      0.906ms     2.94x +
27B TP1          4096+4096           16   48                  0.214ms      0.952ms     4.45x +
27B TP1          2048+6144           16   48                  0.309ms      0.946ms     3.06x +
27B TP1          1024+7168           16   48                  0.359ms      0.949ms     2.64x +
27B TP1          2048x4              16   48                  0.223ms      0.907ms     4.06x +
27B TP1          1024x8              16   48                  0.187ms      0.931ms     4.97x +

2B/0.8B TP1      1x8192              16   16                  0.399ms      0.592ms     1.48x +
2B/0.8B TP1      1x4096              16   16                  0.208ms      0.479ms     2.30x +
2B/0.8B TP1      1x2048              16   16                  0.116ms      0.462ms     3.99x +
2B/0.8B TP1      6144+2048           16   16                  0.304ms      0.554ms     1.83x +
2B/0.8B TP1      4096+4096           16   16                  0.209ms      0.513ms     2.45x +
2B/0.8B TP1      2048+6144           16   16                  0.303ms      0.553ms     1.82x +
2B/0.8B TP1      1024+7168           16   16                  0.351ms      0.575ms     1.64x +
2B/0.8B TP1      2048x4              16   16                  0.118ms      0.518ms     4.39x +
2B/0.8B TP1      1024x8              16   16                  0.077ms      0.523ms     6.84x +

Sym h32          1x8192              32   32                  0.399ms      0.705ms     1.77x +
Sym h32          1x4096              32   32                  0.209ms      0.513ms     2.45x +
Sym h32          1x2048              32   32                  0.115ms      0.465ms     4.03x +
Sym h32          6144+2048           32   32                  0.306ms      0.712ms     2.33x +
Sym h32          4096+4096           32   32                  0.214ms      0.712ms     3.33x +
Sym h32          2048+6144           32   32                  0.307ms      0.714ms     2.33x +
Sym h32          1024+7168           32   32                  0.354ms      0.712ms     2.01x +
Sym h32          2048x4              32   32                  0.125ms      0.719ms     5.74x +
Sym h32          1024x8              32   32                  0.135ms      0.737ms     5.47x +
```

<details>
<summary> gdn_prefill_bench.py </summary>

```
import torch
import torch.nn.functional as F
import numpy as np
from flashinfer.testing import bench_gpu_time
from flashinfer.utils import get_compute_capability
from fla.ops.gated_delta_rule.chunk import chunk_gated_delta_rule_fwd as fla_gdn

HEAD_CONFIGS = [
    # (h_qk, h_v, d, label)
    # Qwen3.5-397B and 122B (h_k=16, h_v=64, d=128) under different TP
    ( 2,  8, 128, "397B/122B TP8"),
    ( 4, 16, 128, "397B/122B TP4"),
    ( 8, 32, 128, "397B/122B TP2"),
    (16, 64, 128, "397B/122B TP1"),
    # Qwen3.5-35B, 9B and 4B (h_k=16, h_v=32, d=128)
    (16, 32, 128, "35B/9B/4B TP1"),
    # Qwen3.5-27B (h_k=16, h_v=48, d=128)
    (16, 48, 128, "27B TP1"),
    # Qwen3.5-2B and 0.8B (h_k=16, h_v=16, d=128)
    (16, 16, 128, "2B/0.8B TP1"),
    # Symmetric heads
    (32, 32, 128, "Sym h32"),
]

SEQ_CONFIGS = [
    # (cu_seqlen_endpoints, label)
    # endpoints are cumulative positions (leading 0 added automatically)
    ((8192,),                                    "1x8192"),
    ((4096,),                                    "1x4096"),
    ((2048,),                                    "1x2048"),
    ((1024 * 6, 8192),                           "6144+2048"),
    ((1024 * 4, 8192),                           "4096+4096"),
    ((1024 * 2, 8192),                           "2048+6144"),
    ((1024 * 1, 8192),                           "1024+7168"),
    ((2048, 2048 * 2, 2048 * 3, 8192),           "2048x4"),
    (tuple(1024 * (i + 1) for i in range(8)),    "1024x8"),
]

WARMUP = 5
ITERS = 20


def detect_gpu_arch():
    major, minor = get_compute_capability(torch.device("cuda"))
    if major == 9:
        return "hopper"
    elif major == 10:
        return "blackwell"
    else:
        raise RuntimeError(f"Unsupported GPU: SM{major}{minor}")


def _get_fi_fn(arch):
    if arch == "hopper":
        from flashinfer.gdn_prefill import chunk_gated_delta_rule_hopper
        return chunk_gated_delta_rule_hopper
    else:
        from flashinfer.gdn_kernels.blackwell_prefill import (
            chunk_gated_delta_rule as chunk_gated_delta_rule_blackwell,
        )
        return chunk_gated_delta_rule_blackwell


def bench_flashinfer(endpoints, h_qk, h_v, d, arch, fi_fn):
    device = "cuda"
    dtype = torch.float16
    N = len(endpoints)
    T = endpoints[-1]
    cu_seqlens = torch.tensor([0] + list(endpoints), dtype=torch.int32, device=device)
    h0 = torch.randn((N, h_v, d, d), dtype=torch.float32, device=device)

    if arch == "hopper":
        # Hopper: packed 3D tensors (T, H, D), 2D gates (T, H_v)
        q = torch.randn((T, h_qk, d), dtype=dtype, device=device)
        k = F.normalize(
            torch.randn(T, h_qk, d, dtype=torch.float32, device=device), p=2, dim=-1
        ).to(dtype)
        v = torch.randn((T, h_v, d), dtype=dtype, device=device)
        g = F.logsigmoid(torch.rand(T, h_v, dtype=torch.float32, device=device))
        beta = torch.rand(T, h_v, dtype=torch.float32, device=device).sigmoid()
        state_out = torch.zeros_like(h0)
        fn = lambda: fi_fn(
            q, k, v, g, beta, None, h0, True, cu_seqlens, False, None, state_out,
        )
    else:
        # Blackwell: batched 4D tensors (1, T, H, D), 3D gates (1, T, H_v)
        q = torch.randn((1, T, h_qk, d), dtype=dtype, device=device)
        k = F.normalize(
            torch.randn(1, T, h_qk, d, dtype=torch.float32, device=device), p=2, dim=-1
        ).to(dtype)
        v = torch.randn((1, T, h_v, d), dtype=dtype, device=device)
        g = F.logsigmoid(torch.rand(1, T, h_v, dtype=torch.float32, device=device))
        beta = torch.rand(1, T, h_v, dtype=torch.float32, device=device).sigmoid()
        state_out = torch.zeros_like(h0)
        fn = lambda: fi_fn(
            q, k, v, g, beta, None, h0, True, cu_seqlens, False, None, state_out,
        )

    times = bench_gpu_time(fn, enable_cupti=True, dry_run_iters=WARMUP, repeat_iters=ITERS)
    torch.cuda.empty_cache()
    return np.average(times)


def bench_fla(endpoints, h_qk, h_v, d):
    device = "cuda"
    dtype = torch.float16
    h = h_v
    N = len(endpoints)
    T = endpoints[-1]
    cu_seqlens = torch.tensor([0] + list(endpoints), dtype=torch.int32, device=device)

    q = torch.randn((1, T, h, d), dtype=dtype, device=device)
    k = F.normalize(
        torch.randn(1, T, h, d, dtype=torch.float32, device=device), p=2, dim=-1
    ).to(dtype)
    v = torch.randn((1, T, h_v, d), dtype=dtype, device=device)
    g = F.logsigmoid(torch.rand(1, T, h_v, dtype=torch.float32, device=device))
    beta = torch.rand(1, T, h_v, dtype=torch.float32, device=device).sigmoid()
    h0 = torch.randn((N, h_v, d, d), dtype=torch.float32, device=device)

    fn = lambda: fla_gdn(q, k, v, g, beta, None, initial_state=h0, output_final_state=True, cu_seqlens=cu_seqlens)
    times = bench_gpu_time(fn, enable_cupti=True, dry_run_iters=WARMUP, repeat_iters=ITERS)
    torch.cuda.empty_cache()
    return np.average(times)


def main():
    arch = detect_gpu_arch()
    arch_label = {"hopper": "Hopper (SM90)", "blackwell": "Blackwell (SM100)"}[arch]
    fi_fn = _get_fi_fn(arch)

    print(f"\nGPU: {torch.cuda.get_device_name(0)} [{arch_label}]")
    print(f"Models: Qwen3.5 family (397B, 122B, 35B, 27B, 9B, 4B, 2B, 0.8B), d=128")
    print()
    fi_col = f"FI {arch_label}"
    header = f"{'Heads':<15s}  {'Seqlens':<16s}  {'h_qk':>4s} {'h_v':>4s}  {fi_col:>22s}  {'FLA/Triton':>10s}  {'Speedup':>8s}"
    print(header)
    print("-" * len(header))

    for h_qk, h_v, d, h_label in HEAD_CONFIGS:
        for endpoints, s_label in SEQ_CONFIGS:
            fi_ms = bench_flashinfer(endpoints, h_qk, h_v, d, arch, fi_fn)
            fla_ms = bench_fla(endpoints, h_qk, h_v, d)
            speedup = fla_ms / fi_ms
            marker = "+" if speedup > 1.0 else "-"
            print(
                f"{h_label:<15s}  {s_label:<16s}  {h_qk:>4d} {h_v:>4d}  {fi_ms:>21.3f}ms  {fla_ms:>9.3f}ms  {speedup:>7.2f}x {marker}"
            )
        print()

if __name__ == "__main__":
    main()

```

</details>

vadiklyutiy:
Perf data for B200

```
GPU: NVIDIA B200 [Blackwell (SM100)]
Models: Qwen3.5 family (397B, 122B, 35B, 27B, 9B, 4B, 2B, 0.8B), d=128

Heads            Seqlens           h_qk  h_v    FI Blackwell (SM100)  FLA/Triton   Speedup
------------------------------------------------------------------------------------------
397B/122B TP8    1x8192               2    8                  0.595ms      0.422ms     0.71x -
397B/122B TP8    1x4096               2    8                  0.317ms      0.313ms     0.99x -
397B/122B TP8    1x2048               2    8                  0.180ms      0.253ms     1.41x +
397B/122B TP8    6144+2048            2    8                  0.451ms      0.369ms     0.82x -
397B/122B TP8    4096+4096            2    8                  0.313ms      0.319ms     1.02x +
397B/122B TP8    2048+6144            2    8                  0.449ms      0.366ms     0.81x -
397B/122B TP8    1024+7168            2    8                  0.516ms      0.391ms     0.76x -
397B/122B TP8    2048x4               2    8                  0.179ms      0.272ms     1.51x +
397B/122B TP8    1024x8               2    8                  0.111ms      0.276ms     2.50x +

397B/122B TP4    1x8192               4   16                  0.590ms      0.456ms     0.77x -
397B/122B TP4    1x4096               4   16                  0.317ms      0.324ms     1.02x +
397B/122B TP4    1x2048               4   16                  0.178ms      0.259ms     1.45x +
397B/122B TP4    6144+2048            4   16                  0.453ms      0.401ms     0.88x -
397B/122B TP4    4096+4096            4   16                  0.317ms      0.354ms     1.12x +
397B/122B TP4    2048+6144            4   16                  0.449ms      0.401ms     0.89x -
397B/122B TP4    1024+7168            4   16                  0.519ms      0.431ms     0.83x -
397B/122B TP4    2048x4               4   16                  0.180ms      0.359ms     1.99x +
397B/122B TP4    1024x8               4   16                  0.114ms      0.369ms     3.23x +

397B/122B TP2    1x8192               8   32                  0.597ms      0.606ms     1.02x +
397B/122B TP2    1x4096               8   32                  0.317ms      0.358ms     1.13x +
397B/122B TP2    1x2048               8   32                  0.180ms      0.275ms     1.53x +
397B/122B TP2    6144+2048            8   32                  0.461ms      0.612ms     1.33x +
397B/122B TP2    4096+4096            8   32                  0.320ms      0.610ms     1.90x +
397B/122B TP2    2048+6144            8   32                  0.455ms      0.611ms     1.34x +
397B/122B TP2    1024+7168            8   32                  0.523ms      0.611ms     1.17x +
397B/122B TP2    2048x4               8   32                  0.188ms      0.618ms     3.28x +
397B/122B TP2    1024x8               8   32                  0.190ms      0.613ms     3.23x +

397B/122B TP1    1x8192              16   64                  0.601ms      1.027ms     1.71x +
397B/122B TP1    1x4096              16   64                  0.321ms      0.546ms     1.70x +
397B/122B TP1    1x2048              16   64                  0.182ms      0.331ms     1.81x +
397B/122B TP1    6144+2048           16   64                  0.469ms      1.023ms     2.18x +
397B/122B TP1    4096+4096           16   64                  0.331ms      1.028ms     3.11x +
397B/122B TP1    2048+6144           16   64                  0.470ms      1.029ms     2.19x +
397B/122B TP1    1024+7168           16   64                  0.544ms      1.031ms     1.90x +
397B/122B TP1    2048x4              16   64                  0.326ms      1.039ms     3.18x +
397B/122B TP1    1024x8              16   64                  0.336ms      1.032ms     3.07x +

35B/9B/4B TP1    1x8192              16   32                  0.597ms      0.606ms     1.02x +
35B/9B/4B TP1    1x4096              16   32                  0.318ms      0.360ms     1.13x +
35B/9B/4B TP1    1x2048              16   32                  0.180ms      0.281ms     1.56x +
35B/9B/4B TP1    6144+2048           16   32                  0.464ms      0.609ms     1.31x +
35B/9B/4B TP1    4096+4096           16   32                  0.323ms      0.609ms     1.89x +
35B/9B/4B TP1    2048+6144           16   32                  0.457ms      0.611ms     1.34x +
35B/9B/4B TP1    1024+7168           16   32                  0.525ms      0.614ms     1.17x +
35B/9B/4B TP1    2048x4              16   32                  0.189ms      0.618ms     3.28x +
35B/9B/4B TP1    1024x8              16   32                  0.192ms      0.614ms     3.20x +

27B TP1          1x8192              16   48                  0.599ms      0.867ms     1.45x +
27B TP1          1x4096              16   48                  0.320ms      0.481ms     1.50x +
27B TP1          1x2048              16   48                  0.181ms      0.314ms     1.73x +
27B TP1          6144+2048           16   48                  0.465ms      0.802ms     1.73x +
27B TP1          4096+4096           16   48                  0.325ms      0.859ms     2.64x +
27B TP1          2048+6144           16   48                  0.458ms      0.860ms     1.88x +
27B TP1          1024+7168           16   48                  0.527ms      0.863ms     1.64x +
27B TP1          2048x4              16   48                  0.321ms      0.810ms     2.53x +
27B TP1          1024x8              16   48                  0.264ms      0.819ms     3.11x +

2B/0.8B TP1      1x8192              16   16                  0.589ms      0.453ms     0.77x -
2B/0.8B TP1      1x4096              16   16                  0.316ms      0.324ms     1.03x +
2B/0.8B TP1      1x2048              16   16                  0.178ms      0.259ms     1.45x +
2B/0.8B TP1      6144+2048           16   16                  0.451ms      0.406ms     0.90x -
2B/0.8B TP1      4096+4096           16   16                  0.315ms      0.355ms     1.13x +
2B/0.8B TP1      2048+6144           16   16                  0.447ms      0.403ms     0.90x -
2B/0.8B TP1      1024+7168           16   16                  0.515ms      0.426ms     0.83x -
2B/0.8B TP1      2048x4              16   16                  0.181ms      0.364ms     2.01x +
2B/0.8B TP1      1024x8              16   16                  0.116ms      0.373ms     3.22x +

Sym h32          1x8192              32   32                  0.593ms      0.608ms     1.03x +
Sym h32          1x4096              32   32                  0.317ms      0.360ms     1.14x +
Sym h32          1x2048              32   32                  0.180ms      0.284ms     1.57x +
Sym h32          6144+2048           32   32                  0.460ms      0.609ms     1.32x +
Sym h32          4096+4096           32   32                  0.320ms      0.608ms     1.90x +
Sym h32          2048+6144           32   32                  0.455ms      0.611ms     1.34x +
Sym h32          1024+7168           32   32                  0.522ms      0.613ms     1.17x +
Sym h32          2048x4              32   32                  0.187ms      0.617ms     3.30x +
Sym h32          1024x8              32   32                  0.191ms      0.612ms     3.21x +
```

dianzhangchen:
Hi @vadiklyutiy, thanks for the comprehensive shapes — very helpful. The benchmarks confirm that performance improves with larger batch size/head count, so the lack of seqlen-dim parallelism is a real limitation. That said, Hopper still beats FLA at batch=1, suggesting there's room for overall optimization beyond just seqlen splitting. I'll keep working on it — one thing to try in the next version: fp16 for the invert op (Hopper uses fp16, Blackwell currently uses tf32).

dianzhangchen:
> I figured out one more problematic thing. Seems that implementation try to tune kernel for every seq_len. If I run S=8192 I get the compilation for 8192 only. If after I run with 8191, I get the new compilation. @dianzhangchen Could you confirm, is it correct?

hi @vadiklyutiy , Good catch, thanks! Fixed.

flashinfer/gdn_kernels/blackwell_prefill/gdn.py
```diff
@@ -0,0 +1,4657 @@
...
+        self,
+        val: cutlass.Float32,
+        sGateCumsum: cute.Tensor,
+        thr_mma: cute.ThrMma,
+        tOut: cute.Tensor,
+        bar_id: Int32,
+        mask: cutlass.Constexpr[cutlass.Boolean] = False,
+        tail_count: Int32 = 128,
+    ):
+        """Build the 128x128 causal gamma matrix in TMEM: gamma[i,j] = exp(cumsum[i]) * exp(-cumsum[j]) for j <= i."""
```
bestzsq: Hi, excellent work!, During my test, I have found that the calculated results for with input data turn out to be nan output. This is because changing `exp(cumsum[i] - cumsum[j])` to `exp(cumsum[i]) * exp(-cumsum[i])` may lead to unstable numerical computation results.

benchmarks/bench_blackwell_gdn_prefill.py
```diff
@@ -0,0 +1,503 @@
...
+    parser.add_argument(
+        "--sweep",
+        action="store_true",
+        help="Run a sweep of many test cases (fixlen + varlen)",
+    )
+
+    args = parser.parse_args()
+
+    device = torch.device("cuda")
+    if not (is_sm100a_supported(device) or is_sm110a_supported(device)):
```
nvpohanh: what if `is_sm103a_supported()`?

vadiklyutiy:
I just found https://github.com/flashinfer-ai/flashinfer/pull/2729

Does somebody know how this PR and #2729 connected/synced?

ZJY0516:
> I just found #2729
> 
> Does somebody know how this PR and #2729 connected/synced?

I tried that before. I remember it doesn't support varlen, and the author hasn't updated it in a while.

vadiklyutiy:
> > I just found #2729
> > Does somebody know how this PR and #2729 connected/synced?
> 
> I tried that before. I remember it doesn't support varlen, and the author hasn't updated it in a while.

thx for clarification

vadiklyutiy:
> Hi @vadiklyutiy, thanks for the comprehensive shapes — very helpful. The benchmarks confirm that performance improves with larger batch size/head count, so the lack of seqlen-dim parallelism is a real limitation. That said, Hopper still beats FLA at batch=1, suggesting there's room for overall optimization beyond just seqlen splitting. I'll keep working on it — one thing to try in the next version: fp16 for the invert op (Hopper uses fp16, Blackwell currently uses tf32).

The problem is that in real life there are no large batches. Usually batch size is 1,2,3. Prompt input length is large enough. vllm max tokens is 8192 by default and commonly 1,2 prompt is enough to fulfill.

ZJY0516:
> > Hi @vadiklyutiy, thanks for the comprehensive shapes — very helpful. The benchmarks confirm that performance improves with larger batch size/head count, so the lack of seqlen-dim parallelism is a real limitation. That said, Hopper still beats FLA at batch=1, suggesting there's room for overall optimization beyond just seqlen splitting. I'll keep working on it — one thing to try in the next version: fp16 for the invert op (Hopper uses fp16, Blackwell currently uses tf32).
> 
> The problem is that in real life there are no large batches. Usually batch size is 1,2,3. Prompt input length is large enough. vllm max tokens is 8192 by default and commonly 1,2 prompt is enough to fulfill.

I think vllm always uses batch=1 (varlen api)

vadiklyutiy:
> > > Hi @vadiklyutiy, thanks for the comprehensive shapes — very helpful. The benchmarks confirm that performance improves with larger batch size/head count, so the lack of seqlen-dim parallelism is a real limitation. That said, Hopper still beats FLA at batch=1, suggesting there's room for overall optimization beyond just seqlen splitting. I'll keep working on it — one thing to try in the next version: fp16 for the invert op (Hopper uses fp16, Blackwell currently uses tf32).
> > 
> > 
> > The problem is that in real life there are no large batches. Usually batch size is 1,2,3. Prompt input length is large enough. vllm max tokens is 8192 by default and commonly 1,2 prompt is enough to fulfill.
> 
> I think vllm always uses batch=1 (varlen api)

Yes, correct vllm always use batch=1. 
But I speculated that in that msg with "performance improves with larger batch" @dianzhangchen meant that batch is several sequences processing together as varlen, because my latest scripts and results don't contain clasical batch results at all - all testcases are batch=1

flashinfer/gdn_kernels/blackwell_prefill/gdn.py
dianzhangchen: hi @bestzsq , thanks for pointing this out—fixed. Reverted to `exp(cumsum[i] - cumsum[j])`

benchmarks/bench_blackwell_gdn_prefill.py
dianzhangchen: Hi @nvpohanh, on sm103 devices, the `is_sm100a_supported` check passes. Thanks for pointing out the confusion—I now use `check_compute_capability` to target sm100, sm103, and sm110.

dianzhangchen:
> > > > Hi @vadiklyutiy, thanks for the comprehensive shapes — very helpful. The benchmarks confirm that performance improves with larger batch size/head count, so the lack of seqlen-dim parallelism is a real limitation. That said, Hopper still beats FLA at batch=1, suggesting there's room for overall optimization beyond just seqlen splitting. I'll keep working on it — one thing to try in the next version: fp16 for the invert op (Hopper uses fp16, Blackwell currently uses tf32).
> > > 
> > > 
> > > The problem is that in real life there are no large batches. Usually batch size is 1,2,3. Prompt input length is large enough. vllm max tokens is 8192 by default and commonly 1,2 prompt is enough to fulfill.
> > 
> > 
> > I think vllm always uses batch=1 (varlen api)
> 
> Yes, correct vllm always use batch=1. But I speculated that in that msg with "performance improves with larger batch" @dianzhangchen meant that batch is several sequences processing together as varlen, because my latest scripts and results don't contain clasical batch results at all - all testcases are batch=1

@vadiklyutiy yes, correct—here “batch” is from the kernel’s perspective

kahyunnam:
It would be good to close this out soon. Will approve after CICD passes

vadiklyutiy:
> It would be good to close this out soon. Will approve after CICD passes

IMO perf isn't good enough. Lets don't do the same mess as we did with GDN decode

vadiklyutiy:
In practice with this perf results vllm can't use this implementation

dianzhangchen:
Since the current implementation uses TF32 for inversion, it underperforms FLA in some cases, so we’re closing this PR for now. A colleague is working on an fp16 version that outperforms FLA across all cases. Please stay tuned.

RRoboter:
Thanks for the work! I have a question regarding block matrix multiplication. Based on my derivation, four MMA operations are insufficient to compute a complete matrix inverse. The output of invert_sub_tiled_mma_ts_l1 only yields partial_inv_30 and partial_inv_31. It seems an additional 64x64 MMA is required to obtain the full inv_30 and inv_31.  Is the current approach of using four MMAs an intentional approximate calculation by design?

bestzsq:
> Since the current implementation uses TF32 for inversion, it underperforms FLA in some cases, so we’re closing this PR for now. A colleague is working on an fp16 version that outperforms FLA across all cases. Please stay tuned.

Hi @dianzhangchen, Is using fp16 precision sufficient for computing the inversion? Are there any additional caveats? In flash-linear-attention, the default precision for the inversion calculation is fp32. I examined the input range for the matrix operations involved in the inversion and found that the absolute maximum value (amax) of the inputs hovered around 1.0. Then attempted to switch the calculation precision to fp16, and it appeared to work without any issues.

dianzhangchen:
> Thanks for the work! I have a question regarding block matrix multiplication. Based on my derivation, four MMA operations are insufficient to compute a complete matrix inverse. The output of invert_sub_tiled_mma_ts_l1 only yields partial_inv_30 and partial_inv_31. It seems an additional 64x64 MMA is required to obtain the full inv_30 and inv_31. Is the current approach of using four MMAs an intentional approximate calculation by design?

Hi @RRoboter , there is no approximation here. In this version, four operations are actually sufficient. I also tried using more GEMMs, and while that improved performance, it led to a slight drop in accuracy.

dianzhangchen:
> > Since the current implementation uses TF32 for inversion, it underperforms FLA in some cases, so we’re closing this PR for now. A colleague is working on an fp16 version that outperforms FLA across all cases. Please stay tuned.
> 
> Hi @dianzhangchen, Is using fp16 precision sufficient for computing the inversion? Are there any additional caveats? In flash-linear-attention, the default precision for the inversion calculation is fp32. I examined the input range for the matrix operations involved in the inversion and found that the absolute maximum value (amax) of the inputs hovered around 1.0. Then attempted to switch the calculation precision to fp16, and it appeared to work without any issues.

Hi @bestzsq , I think FP16 is sufficient for most cases.

RRoboter:
> > Thanks for the work! I have a question regarding block matrix multiplication. Based on my derivation, four MMA operations are insufficient to compute a complete matrix inverse. The output of invert_sub_tiled_mma_ts_l1 only yields partial_inv_30 and partial_inv_31. It seems an additional 64x64 MMA is required to obtain the full inv_30 and inv_31. Is the current approach of using four MMAs an intentional approximate calculation by design?
> 
> Hi @RRoboter , there is no approximation here. In this version, four operations are actually sufficient. I also tried using more GEMMs, and while that improved performance, it led to a slight drop in accuracy.

Thanks for your reply. Based on my derivation, I found that Inv_30 is still missing the term Inv_33 @ D_32 @ Inv_20, and Inv_31 is missing Inv_33 @ D_32 @ Inv_21. Since Inv_33 @ D_32 has already been calculated in the SS_L0 mma, one more mma operation would be needed. Is it because of the characteristics of the D matrix that Inv_20 and Inv_21 are close to zero, and can therefore be ignored?

dianzhangchen:
> > > Thanks for the work! I have a question regarding block matrix multiplication. Based on my derivation, four MMA operations are insufficient to compute a complete matrix inverse. The output of invert_sub_tiled_mma_ts_l1 only yields partial_inv_30 and partial_inv_31. It seems an additional 64x64 MMA is required to obtain the full inv_30 and inv_31. Is the current approach of using four MMAs an intentional approximate calculation by design?
> > 
> > 
> > Hi @RRoboter , there is no approximation here. In this version, four operations are actually sufficient. I also tried using more GEMMs, and while that improved performance, it led to a slight drop in accuracy.
> 
> Thanks for your reply. Based on my derivation, I found that Inv_30 is still missing the term Inv_33 @ D_32 @ Inv_20, and Inv_31 is missing Inv_33 @ D_32 @ Inv_21. Since Inv_33 @ D_32 has already been calculated in the SS_L0 mma, one more mma operation would be needed. Is it because of the characteristics of the D matrix that Inv_20 and Inv_21 are close to zero, and can therefore be ignored?

Hi @RRoboter , Sorry, I'm not quite sure what the subscript numbers mean in your notation (e.g., Inv_33, Inv_20, etc.). However, no "close to zero, therefore ignored" approximation is used here. One key point is that the matrix is lower triangular.

RRoboter:
> > > > Thanks for the work! I have a question regarding block matrix multiplication. Based on my derivation, four MMA operations are insufficient to compute a complete matrix inverse. The output of invert_sub_tiled_mma_ts_l1 only yields partial_inv_30 and partial_inv_31. It seems an additional 64x64 MMA is required to obtain the full inv_30 and inv_31. Is the current approach of using four MMAs an intentional approximate calculation by design?
> > > 
> > > 
> > > Hi @RRoboter , there is no approximation here. In this version, four operations are actually sufficient. I also tried using more GEMMs, and while that improved performance, it led to a slight drop in accuracy.
> > 
> > 
> > Thanks for your reply. Based on my derivation, I found that Inv_30 is still missing the term Inv_33 @ D_32 @ Inv_20, and Inv_31 is missing Inv_33 @ D_32 @ Inv_21. Since Inv_33 @ D_32 has already been calculated in the SS_L0 mma, one more mma operation would be needed. Is it because of the characteristics of the D matrix that Inv_20 and Inv_21 are close to zero, and can therefore be ignored?
> 
> Hi @RRoboter , Sorry, I'm not quite sure what the subscript numbers mean in your notation (e.g., Inv_33, Inv_20, etc.). However, no "close to zero, therefore ignored" approximation is used here. One key point is that the matrix is lower triangular.

Thanks, perhaps I missed some details; I will check again.

RRoboter:
> > > > > Thanks for the work! I have a question regarding block matrix multiplication. Based on my derivation, four MMA operations are insufficient to compute a complete matrix inverse. The output of invert_sub_tiled_mma_ts_l1 only yields partial_inv_30 and partial_inv_31. It seems an additional 64x64 MMA is required to obtain the full inv_30 and inv_31. Is the current approach of using four MMAs an intentional approximate calculation by design?
> > > > 
> > > > 
> > > > Hi @RRoboter , there is no approximation here. In this version, four operations are actually sufficient. I also tried using more GEMMs, and while that improved performance, it led to a slight drop in accuracy.
> > > 
> > > 
> > > Thanks for your reply. Based on my derivation, I found that Inv_30 is still missing the term Inv_33 @ D_32 @ Inv_20, and Inv_31 is missing Inv_33 @ D_32 @ Inv_21. Since Inv_33 @ D_32 has already been calculated in the SS_L0 mma, one more mma operation would be needed. Is it because of the characteristics of the D matrix that Inv_20 and Inv_21 are close to zero, and can therefore be ignored?
> > 
> > 
> > Hi @RRoboter , Sorry, I'm not quite sure what the subscript numbers mean in your notation (e.g., Inv_33, Inv_20, etc.). However, no "close to zero, therefore ignored" approximation is used here. One key point is that the matrix is lower triangular.
> 
> Thanks, perhaps I missed some details; I will check again.

Thank you very much. Your approach is accurate and more efficient. I noticed that my prior derivation contained several redundant computational steps.

PR #2753, opened 2026-03-11
Fix MXFP8 fc1 weight shape check for non-gated activations

### LINKED ISSUE #2731: Potentially superfluous check that disables non gated activations in the cutlass fused moe API
https://github.com/flashinfer-ai/flashinfer/blob/2bb3e9e67aa5f7b05b7f867cac780bba74bf3920/csrc/fused_moe/cutlass_backend/flashinfer_cutlass_fused_moe_binding.cu#L893-L904

This check is triggered for both non-gated and gated-activations. This should only be triggered if we are using gated activations.

he-yufeng:
Fixes #2731.

The MXFP8 quantization path in `getQuantParams()` hardcodes `inter_size * 2` for the fc1 weight block N-dimension check, assuming gated activations (where gate + up weights are concatenated). This causes the check to fail for non-gated activations where the multiplier should be 1.

Replace the hardcoded `* 2` with `* (isGatedActivation(base_activation_type) ? 2 : 1)` in all three MXFP8 `getQuantParams` overloads. Also updated the error messages to show the actual expected multiplier.

This matches the existing pattern in the NVFP4 path (line ~1131) which already correctly guards the `* 2` with an `isGatedActivation` check.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected internal size validation to account for gated activations by using the proper multiplier when certain quantization formats are used.
  * Updated validation error messages to report the adjusted expected size when gated activations affect parameter layout.

### DIFF
--- a/csrc/fused_moe/cutlass_backend/flashinfer_cutlass_fused_moe_binding.cu
+++ b/csrc/fused_moe/cutlass_backend/flashinfer_cutlass_fused_moe_binding.cu
@@ -901,8 +901,8 @@ class FusedMoeRunner : public tvm::ffi::ModuleObj {
                   TmaWarpSpecializedGroupedGemmInput::MXFPXBlockScaleVectorSize ==
               TmaWarpSpecializedGroupedGemmInput::alignToSfDim(
                   hidden_size, TmaWarpSpecializedGroupedGemmInput::MinKDimAlignmentMXFPX))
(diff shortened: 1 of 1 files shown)


aleozlx:
looks like this PR is doing the same as this one
https://github.com/flashinfer-ai/flashinfer/pull/3082

aleozlx (approved):

aleozlx:
ok to close as dup tho? thanks for the contrib

PR #2764, opened 2026-03-12
fused_moe: pre-filter SM89 tactics with zero occupancy on SM120 Blackwell

dryu-nv:
## Problem

Pure FP8 MoE on SM120 (Blackwell CC 12.0) has no native SM120 TMA warp-specialized kernels (`isValidSM120MOESpecialisation<fp8,fp8>()` returns `false`), so the GEMM runner falls back to SM89 (Ada Lovelace) Ampere-style kernels. Some SM89 tile configs require more shared memory than SM120 provides, causing `GemmGrouped::maximum_active_blocks()` to return 0.

The static heuristic in `estimate_best_config_from_occupancies` is supposed to skip configs with 0 occupancy, but the occupancy query used `compute_occupancy_for_kernel<GemmKernel>()` (querying `cutlass::Kernel<GemmKernel>`) while execution used `GemmGrouped::maximum_active_blocks()` (querying a different entry point). These could disagree for SM89 configs on SM120, causing the heuristic to select a config that later asserts at execution:

```
GPU lacks the shared memory resources to run GroupedGEMM kernel
```

During autotuning all candidate tactics are tried, so each failing tactic produces a WARNING log and wastes profiling time.

## Fix

1. **`genericMoeGemmKernelLauncher::call`**: Use `GemmGrouped::maximum_active_blocks()` for the occupancy query, matching the execution path's check. Adds a comment explaining the SM89→SM120 fallback rationale.

2. **`MoeGemmRunner::queryOccupancyForConfig()`**: New method that dispatches with `inputs.occupancy` set so tensor pointers are never dereferenced (the `inputs.occupancy != nullptr` path returns early). Wraps in try-catch to return 0 for unsupported arch/config combinations (e.g., SM90 with FP8 Ampere configs hitting the "No fallback FP8 implementation available" assertion). TMA warp-specialized configs return 1 conservatively.

3. **`CutlassMoeFCRunnerInterface`**: Add `queryOccupancyForConfig()` pure virtual method and `CutlassMoeFCRunner` override.

4. **`flashinfer_cutlass_fused_moe_binding.cu`**: Expose `get_tactic_occupancy(tactic_id)` TVM-FFI binding.

5. **`fused_moe/core.py` `MoERunner.get_valid_tactics()`**: Pre-filter tactics with zero occupancy, eliminating failing tactics before autotuning begins. Falls back gracefully if the binding is unavailable or if all tactics report 0.

## Effect

On SM120 with pure FP8 MoE, the autotuner will only profile tactics that can actually run, eliminating the WARNING spam and wasted profiling time.

## Summary by CodeRabbit

* **New Features**
  * Add occupancy-query capability for MOE GEMM configurations: users can query the maximum active blocks per SM for a given tactic/config without launching kernels.
  * Introduce occupancy-based tactic pre-filtering: automatically excludes tactics with zero occupancy on the current device, and falls back to the full candidate set if all tactics are filtered.

### DIFF
--- a/csrc/fused_moe/cutlass_backend/flashinfer_cutlass_fused_moe_binding.cu
+++ b/csrc/fused_moe/cutlass_backend/flashinfer_cutlass_fused_moe_binding.cu
@@ -722,6 +722,17 @@ class FusedMoeRunner : public tvm::ffi::ModuleObj {
       return Function::FromTyped([this]() -> int64_t { return mGemm1TacticCount; });
     } else if (name == "get_gemm2_tactic_count") {
       return Function::FromTyped([this]() -> int64_t { return mGemm2TacticCount; });
+    } else if (name == "get_tactic_occupancy") {
+      // Returns the max active blocks per SM for the given tactic index.
+      // Returns 0 if the tactic is not supported on the current device (e.g., SM89 tile
+      // configs with insufficient shared memory when running on SM120 Blackwell).
+      return Function::FromTyped([this](int64_t tactic_id) -> int64_t {
+        if (tactic_id < 0 || tactic_id >= static_cast<int64_t>(mAllProfiles.size())) {
+          return 0;
+        }
+        return static_cast<int64_t>(
+            mKernelRunner->queryOccupancyForConfig(mAllProfiles[tactic_id]));
+      });
     } else if (name == "run_moe") {
       return Function::FromTyped(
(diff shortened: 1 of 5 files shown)


aleozlx:
pls help address merge conflicts

dryu-nv:
Closing this PR as better fix (PR 3032) is already merged

PR #2780, opened 2026-03-13
fix(jit): enable GDC for CUTLASS GEMM PDL — SM100 flag only

voipmonitor:
## Summary

Re-applies #2716 (reverted in #2737) with the fix for the AOT build failure.

**Only `-DCUTLASS_ENABLE_GDC_FOR_SM100=1`** is added. The `-DCUTLASS_ENABLE_GDC_FOR_SM90=1` flag that broke AOT builds is intentionally omitted.

## Why the original PR broke AOT

`sm90_gemm_tma_warpspecialized_cooperative.hpp:794` has a direct `#ifdef CUTLASS_ENABLE_GDC_FOR_SM90` guard (not `CUTLASS_GDC_ENABLED`) that calls `scheduler.is_last_tile()`. When compiling SM120 kernels with that flag, the SM120 scheduler (`PersistentTileSchedulerSm100StreamK`) doesn't have `is_last_tile()` → compilation error.

## Why SM100 flag alone is sufficient

CUTLASS 4.2.1 `grid_dependency_control.h` defines `CUTLASS_GDC_ENABLED` for the entire SM100 family (SM100/101/103/120/121) when `CUTLASS_ENABLE_GDC_FOR_SM100` is set. This enables `griddepcontrol.wait` and `griddepcontrol.launch_dependents` device-side barriers for all affected architectures.

## Why this is needed

All affected GEMM kernels hardcode `enablePDL=true`, which enables host-side kernel overlap. Without the GDC compile flag, the device-side synchronization barriers compile as no-ops → race condition → NaN/garbage output tiles under concurrency.

## Affected modules

- `fp4_gemm_cutlass` (SM100)
- `fp4_gemm_cutlass_sm103` (SM103)
- `fp4_gemm_cutlass_sm120` (SM120)
- `fp8_gemm_cutlass` (SM100)
- `mxfp8_gemm_cutlass` (SM100)
- `gemm_sm120` (SM120 FP8 groupwise)

(`tgv_gemm` already had the SM100 flag.)

## Test plan

- [ ] AOT build with `FLASHINFER_CUDA_ARCH_LIST="12.1a"` (the exact config that broke before)
- [ ] AOT build with full arch list `"7.5 8.0 8.9 9.0a 10.0a 12.0a"`
- [ ] FP4 GEMM correctness under concurrent streams on SM120

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Chores**
[...]

PR #2786, opened 2026-03-14
feat: K=64 block-scaled MoE GEMM for SM120 (RTX PRO 6000)

brandonmmusic-max:
## Summary

- Adds K=64 tile shapes to the CUTLASS block-scaled grouped GEMM for SM120 (Blackwell workstation GPUs with 99KB SMEM)
- Fixes two CUTLASS-level bugs that prevented K=64 block-scaled compilation (via submodule update)
- **2x single-user decode throughput** on RTX PRO 6000

## Problem

On SM120 (RTX PRO 6000, DGX Spark — 99KB SMEM):
- K=128 tiles compile but **overflow SMEM at runtime** → "Failed to initialize cutlass TMA WS grouped gemm"
- K=64 tiles **couldn't compile** due to two independent CUTLASS bugs (see below)

This leaves **no usable block-scaled tile shapes on SM120**.

### Bug 1: TMA `basis_get` with zero-stride basis (`copy_traits_sm90_tma.hpp`)

In `fill_tma_gmem_shape_stride`, when a TMA basis element is constant-zero (`_0`, representing a broadcast dimension where SFVectorSize elements share one scale factor), `basis_get(_0, gmem_shape)` falls back to the generic overload which returns the *entire* `gmem_shape` tuple instead of a scalar. This tuple can't convert to `uint64_t`.

For K≥128, the zero-stride mode gets absorbed into adjacent TMA dimensions during basis computation, so the bug is latent. For K=64, the different SMEM layout structure exposes the standalone `_0` basis element.

### Bug 2: SF SMEM layout nesting mismatch (`sm120_blockscaled_mma_builder.inl`)

The scale factor SMEM layout uses `kBasicBlockShape = Shape<SFVectorSize, MMA_NSF>` where `MMA_NSF = AtomK / SFVectorSize`.

For fp8×fp4 (MMA atom K=32, SFVectorSize=32): `MMA_NSF=1`, so `kBasicBlockShape = Shape<32, 1>`. For K=64, only 2 SFs along K (`64/32=2`), but `Blk_SF=4` creates an outer K shape with a zero dimension, triggering: `"TMA requires CTA_Tile and SLayout top-level size equivalence."`

For fp4×fp4 (MMA atom K=64): `MMA_NSF=2`, and the outer K shape is trivially `Shape<1, 1>` — this is why fp4×fp4 K=64 already compiles but fp8×fp4 doesn't.

### `3rdparty/cutlass` (submodule → [brandonmmusic-max/cutlass@flashinfer-sm120-k64-fixes](https://github.com/brandonmmusic-max/cutlass/tree/flashinfer-sm120-k64-fixes))

**`include/cute/atom/copy_traits_sm90_tma.hpp`**:
- In `fill_tma_gmem_shape_stride`, detect `is_constant<0>` basis elements and set shape=1, stride=0 directly instead of calling `basis_get` which returns a tuple

**`include/cutlass/gemm/collective/builders/sm120_blockscaled_mma_builder.inl`**:
- Added `EffBlk_SF = min(K/SFVectorSize, Blk_SF)` to clamp the SF block size when K is small
- Added `FoldSFIntoBasicBlock` — when `EffBlk_SF > MMA_NSF`, folds SF into `kBasicBlockShape` so outer K dimensions become trivial (all 1s), allowing TMA to construct valid descriptors

*These CUTLASS fixes are generic and will be upstreamed to NVIDIA/cutlass. Once merged upstream, the submodule can be pointed back to the official repo.*

### `flashinfer/jit/gemm/cutlass/generate_kernels.py`
- Added `[128,128,64]`, `[128,256,64]`, `[256,128,64]` to SM120 `cta_shapes_mnk`
- Added `_sm120_supported_shapes` set for 99KB SMEM constraint filtering
- Updated fp8×fp4 and fp4×fp4 filters to use shape set instead of single-shape check

### `csrc/.../moe_gemm_template_dispatch_tma_ws.h`
- Added K=64 entries to `are_tile_shapes_supported_sm120()`
- Added `SHAPE_CASE(120, ...)` for K=64 dispatch

## Benchmark Results

**Hardware:** 4x RTX PRO 6000 Blackwell (96GB, SM 12.0, 99KB SMEM)
**Model:** Qwen3.5-397B-A17B-NVFP4, TP=4, MTP=5
**Environment:** CUDA 13.2, CUTLASS 4.4.1, vLLM 0.17.1rc1, FlashInfer 0.6.6

### Decode Throughput

| Users | Before (tok/s) | After (tok/s) | Improvement |
|-------|---------------|--------------|-------------|
| 1 | 142 | **283** | +99% |
| 4 | 250 | **850** | +240% |
| 8 | 510 | **1,283** | +151% |
| 16 | — | **1,624** | — |

### Context Length Scaling (single user, 1K output)

| Context | tok/s |
|---------|-------|
| Short | 283 |
| 1K | 277 |
| 4K | 247 |
| 16K | 183 |
| 32K | 141 |

## Notes

- Some FP4×FP4 K=64 tiles with large M/N (128×256×64, 256×128×64) still overflow 99KB SMEM at runtime — the autotuner correctly skips these
- The 128×128×64 tiles fit for both FP4×FP4 and FP8×FP4
- The CUTLASS submodule points to a temporary fork branch; once the fixes are upstreamed to NVIDIA/cutlass, this can be rebased onto the official commit

## Test Plan

- [x] All SM120 kernel files compile (nvcc sm_120a)
- [x] Full ninja rebuild — zero errors
- [x] fused_moe_120.so links successfully
- [x] vLLM serves requests correctly with new kernels
- [x] Autotuner correctly skips tiles that overflow SMEM
- [x] Benchmark shows significant throughput improvement
- [ ] CI pre-commit / compilation checks pass

Related: vllm-project/vllm#30135

### DIFF
--- a/flashinfer/jit/gemm/cutlass/generate_kernels.py
+++ b/flashinfer/jit/gemm/cutlass/generate_kernels.py
@@ -1009,6 +1009,7 @@ def generate_sm80_fused_grouped_gemm_operations():
         (16, 256, 64),
         (32, 128, 64),
         (64, 128, 64),
+        (128, 128, 64),
     ]
 
     stages = [2, 3, 4]
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "3rdparty/cutlass"]
 	path = 3rdparty/cutlass
-	url = https://github.com/NVIDIA/cutlass.git
+	url = https://github.com/brandonmmusic-max/cutlass.git
 [submodule "3rdparty/spdlog"]
 	path = 3rdparty/spdlog
 	url = https://github.com/gabime/spdlog.git


brandonmmusic-max:
Pre-built Docker image with this fix applied is available:

```
docker pull verdictai/vllm-blackwell-k64:latest
```

Based on `voipmonitor/llm-pytorch-blackwell:nightly-cuda132` with CUDA 13.2, vLLM 0.17.1rc1, FlashInfer 0.6.6. Ready to use on RTX PRO 6000 / DGX Spark (SM120) hardware.

brandonmmusic-max:
### Updated Benchmark Results — Methodology Correction

After more thorough testing, the initial benchmark numbers were inflated by MTP speculative decoding achieving near-100% acceptance on trivial thinking tokens (`<think></think>`). I want to be transparent about what the numbers represent.

**Corrected results with thinking disabled and real prompts (substantive generation):**

| Users | System tok/s | Per-user tok/s |
|-------|-------------|---------------|
| 1 | 136 | 136 |
| 2 | 217 | 109 |
| 4 | 342 | 85 |
| 8 | 472 | 59 |

**Benchmark methodology breakdown:**

| Scenario | 1-user tok/s | Notes |
|----------|-------------|-------|
| Short prompt, thinking ON (synthetic peak) | 283 | MTP inflated by trivial think tokens |
| Real prompt, thinking ON | 161 | Think tokens still boost MTP acceptance |
| **Real prompt, thinking OFF (actual usage)** | **~130-136** | **Actual usable throughput** |
| Pre-patch baseline (community reports) | ~110 | Same hardware, no K=64 fix |

The K=64 kernel patch provides a real **~20-25% improvement** over the pre-patch baseline on identical hardware. The fix unblocks SM120 GPUs from falling back to slow GEMM paths by giving the autotuner CUTLASS tiles that fit within 99KB SMEM.

Note: Engine-level throughput (what most benchmarks report) will differ from these end-to-end API measurements which include HTTP overhead. The relative improvement from the K=64 patch is consistent across measurement methods.

eugr:
Is this guarded to sm120 only or sm120 AND sm121? Flashinfer arch guards are arch-specific, not family specific.

brandonmmusic-max:
Fixed the ruff format issue (one tuple per line). Pre-commit should pass now. Could a maintainer run `@flashinfer-bot run` to authorize CI? Thanks!

brandonmmusic-max:
**Re: SM120 vs SM121 guard**

Both are covered. The C++ dispatch at `moe_gemm_template_dispatch_tma_ws.h:336` guards with:

```cpp
Arch::kMinComputeCapability == 120 || Arch::kMinComputeCapability == 121
```

And `are_tile_shapes_supported_sm120()` is called for both SM120 and SM121. The Python codegen comment has been updated to say "SM120/SM121" for clarity.

SM121 (DGX Spark / GB10) has the same ~99KB SMEM constraint as SM120, so the K=64 tiles are needed on both.

**Re: coderabbit nitpick**

Fixed in 7145384 — hoisted `_sm120_supported_shapes` out of the loop.

brandonmmusic-max:
**Re: cta_shapes_mnk alignment with _sm120_supported_shapes**

The extra shapes in `cta_shapes_mnk` (`128×128×256`, `128×256×128`, `256×128×128`) are still used by non-block-scaled dtype paths (e.g. FP8×FP8). The `_sm120_supported_shapes` filter only applies to the `e4m3×e2m1` and `e2m1×e2m1` block-scaled combinations. Collapsing the two would break codegen for those other dtypes.

RobTand:
Using this patch in production on DGX Spark (SM121) — it gives ~2x decode throughput improvement with K=64 tiles vs K=128. Running Nemotron-3-Super-120B and Qwen3.5-122B NVFP4 models via vLLM.

Confirmed it works cleanly on top of the CUTLASS 4.4.2 upgrade merged in #2798. cc @aleozlx since you're actively testing SM121 on #2780.

brandonmmusic-max:
> Is this guarded to sm120 only or sm120 AND sm121? Flashinfer arch guards are arch-specific, not family specific.

There’s another gentleman below who says he’s running it on the DGX spark who gotten 2x decode with the patch, so it’ll work on sm 121 as well.

RobTand:
@brandonmmusic-max  I have a bunch of other changes that depend on your changes (specifically for spark). Let me know if you need any help getting these out. Nice work btw. I know they aren't leading to quite the performance gains you'd expected but they're still necessary for nvfp4 support and surely nvfp4 gates other enhancements.

RobTand:
I think i tagged you in the other one I was working on (https://github.com/NVIDIA/cutlass/pull/3121).

aleozlx:
there appears to be errors on sm120

aleozlx:
[sm120_errors.txt](https://github.com/user-attachments/files/26152621/sm120_errors.txt)

flashinfer/jit/gemm/cutlass/generate_kernels.py
```diff
@@ -992,7 +1009,6 @@ def generate_sm80_fused_grouped_gemm_operations():
         (16, 256, 64),
         (32, 128, 64),
         (64, 128, 64),
-        (128, 128, 64),
```
aleozlx, line 995: i want to seek some clarification why this tile is removed?

brandonmmusic-max: Good catch — this removal was unintentional. The `(128, 128, 64)` tile here is for SM80 f16/bf16 fused grouped GEMM and is completely unrelated to the SM120 block-scaled changes. I'll revert this line.

I'm also updating this PR to include the two CUTLASS-level fixes that are required for K=64 block-scaled tiles to actually compile. Without them, the CI hits a TMA static assertion (`"TMA requires CTA_Tile and SLayout top-level size equivalence."`) because the scale factor shared memory layout produces a zero-size dimension when `K/SFVectorSize < Blk_SF` (i.e. 64/32=2 < 4).

The two CUTLASS fixes (added via submodule update pointing to my CUTLASS fork):

1. **`copy_traits_sm90_tma.hpp`** — Handle zero-stride basis elements in `fill_tma_gmem_shape_stride`. When a basis element is constant-zero (broadcast dimension, e.g. SFVectorSize elements sharing one scale factor), `basis_get(_0, gmem_shape)` returns the entire tuple instead of a scalar. Fix: detect `is_constant<0>` and set shape=1, stride=0 directly.

2. **`sm120_blockscaled_mma_builder.inl`** — Clamp `Blk_SF` to `min(K/SFVectorSize, Blk_SF)` and fold the effective block into `kBasicBlockShape` when the tile K is too small for the default block size. This keeps outer K dimensions trivial (all 1s) so TMA can construct valid descriptors.

These are generic fixes that benefit any SM120 device with <227KB SMEM (e.g. RTX PRO 6000 with 99KB). I plan to upstream them to NVIDIA/cutlass as well.

brandonmmusic-max:
Updated — reverted the SM80 tile removal and added the two CUTLASS fixes via submodule update. Ready for another build test when you get a chance.

aleozlx:
thanks for persistent support

may i ask which commit is relevant that we are trying to run from

> the two CUTLASS fixes via submodule update

https://github.com/brandonmmusic-max/cutlass.git ?

i looked at the branch but it's not obvious to me which commit was related

https://github.com/brandonmmusic-max/cutlass/commits

therefore i cannot issue ci run for unvetted code (this is the reason CI's don't just automatically run). thx for understanding

also note that if you wanted the cutlass version bump, our main branch has bumped it to a new version

brandonmmusic-max:
Totally fair, I should have made that clearer — sorry about that.

It's one commit: [`ede914e4`](https://github.com/brandonmmusic-max/cutlass/commit/ede914e4) on the `flashinfer-sm120-k64-fixes` branch. Just two files:

1. `copy_traits_sm90_tma.hpp` — handles zero-stride basis elements in `fill_tma_gmem_shape_stride`
2. `sm120_blockscaled_mma_builder.inl` — clamps `Blk_SF` to `min(K/SFVectorSize, Blk_SF)` for K=64

I checked and the bug is still there in v4.4.2 (`da5e086`) — line 195 of the builder still divides by `Blk_SF` unconditionally, so K=64 with SFVectorSize=32 gives you a zero-size dimension and the TMA assertion blows up.

A few ways we could handle this — happy to go with whatever's easiest for you:

**A)** I rebase the fix onto v4.4.2 so you can see a clean diff against the version main already uses

**B)** Drop the submodule change entirely and just patch the files inline somewhere in `csrc/nv_internal/` — no external code to vet

**C)** I push the fix upstream to NVIDIA/cutlass first and we wait for it to land before merging this

The FlashInfer-side changes (generate_kernels.py + dispatch) are independent — they just need these two CUTLASS fixes to actually compile.

johnnynunez:
> Totally fair, I should have made that clearer — sorry about that.
> 
> It's one commit: [`ede914e4`](https://github.com/brandonmmusic-max/cutlass/commit/ede914e4) on the `flashinfer-sm120-k64-fixes` branch. Just two files:
> 
> 1. `copy_traits_sm90_tma.hpp` — handles zero-stride basis elements in `fill_tma_gmem_shape_stride`
> 2. `sm120_blockscaled_mma_builder.inl` — clamps `Blk_SF` to `min(K/SFVectorSize, Blk_SF)` for K=64
> 
> I checked and the bug is still there in v4.4.2 (`da5e086`) — line 195 of the builder still divides by `Blk_SF` unconditionally, so K=64 with SFVectorSize=32 gives you a zero-size dimension and the TMA assertion blows up.
> 
> A few ways we could handle this — happy to go with whatever's easiest for you:
> 
> **A)** I rebase the fix onto v4.4.2 so you can see a clean diff against the version main already uses
> 
> **B)** Drop the submodule change entirely and just patch the files inline somewhere in `csrc/nv_internal/` — no external code to vet
> 
> **C)** I push the fix upstream to NVIDIA/cutlass first and we wait for it to land before merging this
> 
> The FlashInfer-side changes (generate_kernels.py + dispatch) are independent — they just need these two CUTLASS fixes to actually compile.

@depaulmillz

PR #2792, opened 2026-03-16
feat: Support padding tokens with seqlen=0 for rope+quant+kv cache update fusion kernel

elvischenv:
## 📌 Description

vLLM is using seqlen=0 padding tokens for running a full cudagraph: https://github.com/vllm-project/vllm/blob/95c0f928cdeeaa21c4906e73cee6a156e1b3b995/vllm/v1/worker/gpu/model_runner.py#L652-L654

Update the following functions:
`get_batch_indices_positions_kernel`: initialize `batch_indices`/`positions` to `-1`/`0` for recognizing the padding tokens
`rope_quantize_fp8_append_paged_kv_cache`: skip those padding tokens

Testing:
`pytest -v -s tests/attention/test_rope.py::test_rope_quantize_fp8_append_paged_kv_cache_padding`
```
======= 4 passed in 1.66s =======
```

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Paged KV cache append now reliably skips padding entries, preventing unintended overwrites during attention quantization and positional encoding processing.
  * Deterministic initialization of internal index/position buffers to avoid uninitialized memory affecting padding behavior.

* **Tests**
  * Added parameterized regression test confirming padding entries are ignored and prefilled KV cache data remains exactly unchanged.


include/flashinfer/pos_enc.cuh
```diff
@@ -859,169 +859,175 @@ __global__ void RopeQuantizeAppendPagedKVCacheKernel(
...
-    // Load cos/sin for RoPE processing blocks only
-    if ((tx * vec_size < rope_dim) && (by < k_rope_end)) {
-      int sin_offset = rope_dim / 2;
-      int vec_idx;
-      if constexpr (interleave) {
-        vec_idx = (tx * vec_size) / 2;  // Force integer division
-      } else {
-        vec_idx = (tx * vec_size) % half_rope_dim;
+    // skip padding tokens with batch_indices < 0
+    if (batch_indices[idx] >= 0) {
```
elvischenv, line 863: The main change is just this line. The following is just indent formatting.

elvischenv:
Hi @yzh119, could you help review this? We need this fix for integrating this kernel to vLLM. Thanks!

elvischenv:
cc @kahyunnam for viz.

nvpohanh:
@elvischenv could you rebase again?

nvpohanh:
@bkryu Could you review this or assign someone to review this? Thanks!

nvpohanh:
@bkryu could you review this? Thanks

flashinfer/page.py
bkryu, line 1: Hi @elvischenv, the changes generally look correct and the CI does seem to pass.

However, Have you tried measuring the performance implications? Asking because `torch.full` or `torch.zeros` tend to call memset which have a higher overhead than `torch.empty`. I'm wondering whether there will be a noticeable performance difference from it.

nvpohanh: @elvischenv is this called per decoding step or per attention layer? If this is per decoding step, I am less worried about the additional memsets. But if this is per-layer, it may be noticeable perf overhead.

elvischenv: `get_batch_indices_positions` is a helper function, preparing the needed arguments for `rope_quantize_fp8_append_paged_kv_cache`, should only be called per decoding step. Then the whole iteration can reuse the same `batch_indices` and `positions`, which won't produce the noticeable overhead.

nvpohanh: @bkryu Once per decoding step should be okay? Do you agree?

nvpohanh:
@bkryu Are the failures known ones or caused by this PR?

flashinfer/page.py
bkryu: I agree that it should be fine.

bkryu (approved):
CI failures are unrelated. LGTM!

PR #2521, opened 2026-02-08
Feat/gdn decode pooled

xutizhou:
## 📌 Description

This PR adds pool-indexed (indirect) state access to the GDN decode kernel, enabling zero-copy integration with SGLang's state pool architecture.

### Background: SGLang's State Pool Architecture

In SGLang, when serving linear attention models (like Qwen3-Next using Gated Delta Rule), we maintain a **state pool** to store recurrent states for all active requests:

`ssm_states: [num_layers, pool_size, num_heads, head_dim, head_dim]`

where `pool_size` = `max_num_reqs` (maximum concurrent requests).

Each active request has a `req_pool_idx` that maps it to a slot in this pool. The mapping is **not contiguous** - requests come and go, so indices can be scattered (e.g., a batch of 4 requests might have pool indices `[3, 7, 12, 25]`).

### Motivation

The current GDN decode kernel expects state with shape `[B, H, K, V]` where B equals batch size and there's a 1:1 mapping (batch index i → state index i). To use it with SGLang's pool, we would need to:

1. **Gather** states from pool indices before kernel call
2. Run kernel on contiguous `[B, H, K, V]` state  
3. **Scatter** updated states back to pool indices

This adds 2 extra memory copy operations per decode step.

### Changes

This PR adds a `state_indices` parameter for **zero-copy pool access**:

```python
def gated_delta_rule_decode_pretranspose(
    q, k, v, beta,
    state,           # Can be [pool_size, H, K, V] instead of [B, H, K, V]
    state_indices,   # NEW: int32 tensor [B] mapping batch_idx -> pool_idx
    ...
)
```

When `state_indices` is provided:
- Kernel uses indirect addressing: `state[state_indices[batch_idx]]` instead of `state[batch_idx]`
- Negative indices (padding slots for CUDA graph) skip computation and write zeros to output
- Eliminates gather/scatter overhead + host-side `torch.where` for padding (~37μs/call)

## 🔍 Related Issues

- [sgl-project/sglang#18361](https://github.com/sgl-project/sglang/pull/18361) - FlashInfer K-last GDN integration into SGLang

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
[...]

PR #2793, opened 2026-03-16
Feat [GDN]: Added Negative Index Padding for BF16 decode & MTP

Itssshikhar:
## 📌 Description

  This PR implements negative-index padding for the remaining pooled GDN runtime paths: BF16 decode and MTP.
 
  There is already exist support for pooled state access for GDN decode through `initial_state` plus `initial_state_indices`, where each batch row maps to a slot in a persistent state pool instead of `assuming batch_idx == state_idx`.
 
  In this, a negative pooled index means:
 
  - this batch row is padding
  - there is no active sequence behind it
  - the kernel must skip all stateful work for that row
 
  The FP32 pooled decode path already had this behavior from earlier work. The BF16 fast path and MTP path did not.

  ## What this PR changes
 
 
  ### BF16 decode
 
  Adds negative-index padding handling to the BF16 VK-layout fast path:
 
 
  - T=1 main kernel
  - T=1 low-batch kernel
  - unified T=2/3/4 kernel
 
 
  For padded rows, these kernels now:
 
  - skip the normal state/gate/update path
  - explicitly zero the related output slice
  - avoid going to pooled state
 
  ### MTP
 
  Adds explicit padding output zeroing to the MTP pooled path so negative-index rows do not get stale output values.
 
 
  For padded rows, MTP now:
 
  - skips recurrent computation
  - skips state update
  - zeros output explicitly
  - preserves state-side buffers for those rows

## 🔍 Related Issues

  - #2687: Idea originally came from the proposed changes in GDN Decode/Prefill API.
  - #2521: pooled indirect state access and negative-index padding for the FP32 path.
  - #2727: non-contiguous pooled decode support.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed padding slot handling to properly zero outputs for invalid batch indices instead of undefined behavior.
  * Improved memory access safety with explicit guard checks to prevent illegal accesses.

* **Tests**
  * Added comprehensive test suite validating padding behavior across decode and multi-token processing scenarios, including all-padding edge cases.

### DIFF
--- a/flashinfer/gdn_kernels/gdn_decode_bf16_state.py
+++ b/flashinfer/gdn_kernels/gdn_decode_bf16_state.py
@@ -789,304 +789,309 @@ def gated_delta_rule_decode_kernel_seqlen1(
         cutlass.Float32, cute.make_layout((32, 4), stride=(1, 32))
     )
 
-    h_global = gH[(pool_batch_idx, value_head_idx, None, None)]
-
-    # Launch first 2 async loads
-    load_h_chunk_async(h_sh_chunk0, h_global, tidx, 0)
-    nvvm.cp_async_commit_group()
-    load_h_chunk_async(h_sh_chunk1, h_global, tidx, 32)
-    nvvm.cp_async_commit_group()
-
-    # L2 normalization
-    q_head = gQ[(batch_idx, 0, query_head_idx, None)]
-    k_head = gK[(batch_idx, 0, query_head_idx, None)]
-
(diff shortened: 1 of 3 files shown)


Itssshikhar:
@kaixih would really appreciate inputs on this.

nvpohanh:
@kaixih is this still needed?

Issue #1419, opened 2025-08-08
Ragged prefill SM90 fails with illegal memory access for total seqlen >= 16384


Edenzzzz:
Not an urgent issue, just came across this in a random test

<img width="1237" height="346" alt="Image" src="https://github.com/user-attachments/assets/8dc8c8be-b54d-4ecf-8444-33e485f89da6" />

```
import numpy as np
import torch
from triton.testing import do_bench

import flashinfer

# page_block_size = 16
page_block_size = 1
num_kv_heads = 4
num_qo_heads = 32
head_dim = 128


def bench_batch_decode(
    batch_size,
    seq_len,
    num_qo_heads,
    num_kv_heads,
    head_dim,
    page_block_size,
    q_dtype,
    kv_dtype,
):
    np.random.seed(42)
    seq_lens = torch.full((batch_size,), seq_len)
    seq_lens_blocks = torch.ceil(seq_lens / page_block_size).int()
    kv_indptr = torch.cat([torch.tensor([0]), torch.cumsum(seq_lens_blocks, 0)], dim=0)
    kv_indptr = kv_indptr.int()
    last_page_len = seq_lens - (seq_lens_blocks - 1) * page_block_size
    last_page_len = last_page_len.int()
    num_blocks = kv_indptr[-1].item()

    q = torch.rand(batch_size, num_qo_heads, head_dim, dtype=q_dtype, device="cuda:0")
    kv_data = torch.randn(
        num_blocks, 2, page_block_size, num_kv_heads, head_dim, device="cuda:0"
    ).to(kv_dtype)
    workspace_buffer = torch.empty(
        128 * 1024 * 1024, dtype=torch.uint8, device="cuda:0"
    )
    wrapper = flashinfer.BatchDecodeWithPagedKVCacheWrapper(
        workspace_buffer, kv_layout="NHD", use_tensor_cores=True
    )
    wrapper.plan(
        kv_indptr.to(0),
        torch.arange(num_blocks).int().to(0),
        last_page_len.to(0),
        num_qo_heads,
        num_kv_heads,
        head_dim,
        page_block_size,
        data_type=kv_dtype,
        q_data_type=q_dtype,
    )

    ms = do_bench(lambda: wrapper.run(q, kv_data))

    wrapper_ragged = flashinfer.BatchPrefillWithRaggedKVCacheWrapper(
        torch.empty(256 * 1024 * 1024, dtype=torch.uint8, device="cuda:0"),
[...]

PR #2801, opened 2026-03-17
[fix] bugfix 1419: Add batch size shape validation in decode and prefill run() APIs

qsang-nv:
## 📌 Description

This PR fixes https://github.com/flashinfer-ai/flashinfer/issues/1419, adding shape checks in several run() APIs to validate q.shape[0] before launching kernels, preventing silent OOB reads that cause CUDA crashes.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Added runtime validations to batched decode (including MLA path) to detect query/batch-size mismatches early.
  * Added runtime validations to prefill to ensure token/count consistency across paged and ragged flows, with backend-specific checks.
  * Error messages now report exact expected shapes/counts to make misconfigurations easier to diagnose.

### DIFF
--- a/flashinfer/prefill.py
+++ b/flashinfer/prefill.py
@@ -1847,12 +1847,13 @@ class BatchPrefillWithPagedKVCacheWrapper:
         self._max_item_len_ptr = max_item_len_ptr
 
         # NOTE(Zihao): only required if qo_indptr/paged_kv_indptr are device tensors
+        qo_indptr_host = qo_indptr.to("cpu")
+        self._qo_indptr_last = int(qo_indptr_host[-1])
+        total_num_rows = self._qo_indptr_last
         if max_token_per_sequence is not None:
             self._max_q_len = max_token_per_sequence
         else:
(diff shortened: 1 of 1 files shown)


flashinfer/prefill.py
```diff
@@ -3113,6 +3130,23 @@ def run(
         _check_cached_qkv_data_type(
             q, k, self._cached_q_data_type, self._cached_kv_data_type
         )
+        # Validate q shape matches qo_indptr
+        expected_qo_indptr_last = self._qo_indptr_buf[-1].item()
```
yzh119: `self._qo_indptr_buf[-1].item()` would be a blocking (requires synchronization) operation if `_qo_indptr_buf` is gpu tensor, and I would encourage moving it to `plan` if possible, we should get rid of synchronization in `run` functions.

qsang-nv: Done, now the size is obtained in `plan` and saved as a scalar to be used for checking.

PR #2808, opened 2026-03-17
Fix bmm_fp8 cublasLt handle usage in autotuned cublas runner  #26381

baonudesifeizhai:
## 🔍 Related Issues
 https://github.com/vllm-project/vllm/issues/26381
Root cause:
torch.cuda.current_blas_handle() returns cublasHandle_t, but bmm_fp8 reinterpreted it as cublasLtHandle_t before calling cublasLt APIs. This can fail in autotune/profiling paths on Blackwell.

on vllm side 
before: https://paste.ubuntu.com/p/npS2tkZY2c/
after : https://paste.ubuntu.com/p/c6Ys69PqvR/

and :  
```
 CUDA_VISIBLE_DEVICES=4,5,6,7 \
VLLM_DISABLE_COMPILE_CACHE=1 \
TORCHINDUCTOR_FORCE_DISABLE_CACHES=1 \
python examples/basic/offline_inference/generate.py \
  --model redhatai/meta-Llama-3.1-8B-FP8 \
  --tensor-parallel-size 4 \
  --kv-cache-dtype fp8 \
  --trust-remote-code \
  --max-model-len 4096 \
  --max-tokens 32 \
  --temperature 0
```
:https://paste.ubuntu.com/p/rPYcXQSwPC/

image

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Refactor

* Simplified FP8 batch matrix multiplication interfaces by removing the need to provide an external cuBLAS handle.
* Improved FP8 GEMM execution by automatically managing and reusing cuBLAS resources for each GPU device.
* Updated FP8 algorithm selection and execution workflows to use this internal resource management consistently.

### DIFF
--- a/csrc/bmm_fp8.cu
+++ b/csrc/bmm_fp8.cu
@@ -17,11 +17,43 @@
 #include <driver_types.h>
 
 #include <flashinfer/gemm/bmm_fp8.cuh>
+#include <unordered_map>
 
 #include "tvm_ffi_utils.h"
 
+namespace {
+
+struct ThreadLocalCublasLtHandles {
+  ~ThreadLocalCublasLtHandles() {
+    // Best-effort cleanup during thread teardown.
+    for (auto& [_, handle] : handles) {
+      if (handle != nullptr) {
+        (void)cublasLtDestroy(handle);
+      }
+    }
+  }
+
+  std::unordered_map<int, cublasLtHandle_t> handles;
+};
+
+cublasLtHandle_t get_cublaslt_handle(int device_id) {
+  static thread_local ThreadLocalCublasLtHandles cache;
+
+  if (auto it = cache.handles.find(device_id); it != cache.handles.end()) {
+    return it->second;
+  }
+
+  ffi::CUDADeviceGuard device_guard(device_id);
+  cublasLtHandle_t handle = nullptr;
+  FLASHINFER_CUBLAS_CHECK(cublasLtCreate(&handle));
+  cache.handles.emplace(device_id, handle);
+  return handle;
+}
+
+}  // namespace
+
 void bmm_fp8(TensorView A, TensorView B, TensorView D, TensorView A_scale, TensorView B_scale,
-             TensorView workspace_buffer, int64_t cublas_handle) {
+             TensorView workspace_buffer) {
   CHECK_CUDA(A);
   CHECK_CUDA(B);
   CHECK_CUDA(D);
@@ -44,9 +76,9 @@ void bmm_fp8(TensorView A, TensorView B, TensorView D, TensorView A_scale, Tenso
         auto k = A.size(2);
         auto n = B.size(2);
 
-        auto lt_handle = reinterpret_cast<cublasLtHandle_t>(cublas_handle);
         ffi::CUDADeviceGuard device_guard(A.device().device_id);
         auto stream = get_stream(A.device());
+        auto lt_handle = get_cublaslt_handle(A.device().device_id);
 
         auto status = flashinfer::bmm_fp8::bmm_fp8_internal_cublaslt(
             workspace_buffer.data_ptr(), workspace_buffer.numel(),
--- a/csrc/bmm_fp8.cu
+++ b/csrc/bmm_fp8.cu
@@ -17,11 +17,43 @@
 #include <driver_types.h>
 
 #include <flashinfer/gemm/bmm_fp8.cuh>
+#include <unordered_map>
 
 #include "tvm_ffi_utils.h"
 
+namespace {
+
+struct ThreadLocalCublasLtHandles {
+  ~ThreadLocalCublasLtHandles() {
+    // Best-effort cleanup during thread teardown.
+    for (auto& [_, handle] : handles) {
+      if (handle != nullptr) {
+        (void)cublasLtDestroy(handle);
+      }
+    }
+  }
+
+  std::unordered_map<int, cublasLtHandle_t> handles;
+};
+
+cublasLtHandle_t get_cublaslt_handle(int device_id) {
+  static thread_local ThreadLocalCublasLtHandles cache;
+
+  if (auto it = cache.handles.find(device_id); it != cache.handles.end()) {
+    return it->second;
+  }
+
+  ffi::CUDADeviceGuard device_guard(device_id);
+  cublasLtHandle_t handle = nullptr;
+  FLASHINFER_CUBLAS_CHECK(cublasLtCreate(&handle));
+  cache.handles.emplace(device_id, handle);
+  return handle;
+}
+
+}  // namespace
+
 void bmm_fp8(TensorView A, TensorView B, TensorView D, TensorView A_scale, TensorView B_scale,
-             TensorView workspace_buffer, int64_t cublas_handle) {
+             TensorView workspace_buffer) {
   CHECK_CUDA(A);
   CHECK_CUDA(B);
   CHECK_CUDA(D);
@@ -44,9 +76,9 @@ void bmm_fp8(TensorView A, TensorView B, TensorView D, TensorView A_scale, Tenso
         auto k = A.size(2);
         auto n = B.size(2);
 
-        auto lt_handle = reinterpret_cast<cublasLtHandle_t>(cublas_handle);
         ffi::CUDADeviceGuard device_guard(A.device().device_id);
         auto stream = get_stream(A.device());
+        auto lt_handle = get_cublaslt_handle(A.device().device_id);
 
         auto status = flashinfer::bmm_fp8::bmm_fp8_internal_cublaslt(
             workspace_buffer.data_ptr(),
@@ -64,8 +96,7 @@ void bmm_fp8(TensorView A, TensorView B, TensorView D, TensorView A_scale, Tenso
 }
 
 int64_t bmm_fp8_get_algos(TensorView A, TensorView B, TensorView D, TensorView A_scale,
-                          TensorView B_scale, TensorView workspace_buffer, int64_t cublas_handle,
-                          TensorView algo_buffer) {
+                          TensorView B_scale, TensorView workspace_buffer, TensorView algo_buffer) {
   CHECK_CUDA(A);
   CHECK_CUDA(B);
   CHECK_CUDA(D);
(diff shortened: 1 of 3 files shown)


yzh119 (commented):
Hi @baonudesifeizhai thanks for the fix, it's look good.

> torch.cuda.current_blas_handle() returns cublasHandle_t, but bmm_fp8 reinterpreted it as cublasLtHandle_t before calling cublasLt APIs

I noticed that cublasLtHandle was also used in pytorch, do you know if it's exposed in python (https://github.com/pytorch/pytorch/blob/98849e6ad451547ea09a958757e6bb422996e65f/aten/src/ATen/cuda/CublasHandlePool.cpp#L417)?

baonudesifeizhai:
only have this  torch.cuda.current_blas_handle()   . but it lose context  , can only get an integer handle for the current state at the Python layer.

maybe we can use at::cuda::getCurrentCUDABlasHandle()  ..but  i want decoupling from pytorch side...
> Hi @baonudesifeizhai thanks for the fix, it's look good.
> 
> > torch.cuda.current_blas_handle() returns cublasHandle_t, but bmm_fp8 reinterpreted it as cublasLtHandle_t before calling cublasLt APIs
> 
> I noticed that cublasLtHandle was also used in pytorch, do you know if it's exposed in python (https://github.com/pytorch/pytorch/blob/98849e6ad451547ea09a958757e6bb422996e65f/aten/src/ATen/cuda/CublasHandlePool.cpp#L417)?

baonudesifeizhai:
Is anything  failed on cicd..?

pjh4993:
Thanks for pinning this down. I hit the same crash (a hard SIGSEGV during FlashInfer autotune warmup on B200, no Python traceback) while serving an FP8/NVFP4 model and checking #3085

cuda-gdb on the isolated call shows `PC=0x0` from `call *(%rax)` with `[rax]==NULL`, every frame inside `libcublasLt.so`, i.e. cuBLASLt dispatching through a garbage function pointer read from the mis-typed handle, exactly as you describe.

**A/B verified on B200** (driver 580.126.09, cuBLAS 12.8), fresh JIT build, only the handle changed:

| handle | result |
| --- | --- |
| `reinterpret_cast<cublasLtHandle_t>(current_blas_handle())` | SIGSEGV |
| a real `cublasLtCreate()` handle (this PR) | OK, incl. under `autotune()` (output cos 0.999) |

So the driver-version correlation in #3085 (580.x fails, 590 passes) is a symptom of the UB (benign on some arch/driver layouts, fatal on Blackwell), not a driver bug.

**One update needed before this lands on current `main`:** since this PR (March), #3437 added two more cuBLAS entry points in `csrc/bmm_fp8.cu`, `bmm_fp8_get_algos` and `bmm_fp8_run_with_algo`, and both repeat the same `reinterpret_cast<cublasLtHandle_t>(cublas_handle)` (with matching `current_blas_handle()` call sites in `gemm/gemm_base.py`). They are on the autotune path, so as-is the fix would still crash there. The complete change routes all three functions through `get_cublaslt_handle()` and drops the `cublas_handle` parameter from all three (the C++ defs, the FFI binding, and the Python calls).

I verified the 3-site version end-to-end on B200 (the cuBLAS runner now tunes under `autotune()` and returns correct output) and put it up for reference here: https://github.com/pjh4993/flashinfer/pull/1 (all three sites, param dropped from the C++/binding/Python, plus a small regression test: `bmm_fp8` `backend="cublas"` under `autotune()` on SM100). Happy to push it to your branch, or you can cherry-pick from there, whichever you prefer.

This also resolves vllm-project/vllm#39814 (same crash on a plain FP8 model).

cc @dhiraj113 @bkryu

ZiyiTsang:
I face the problem in issue with 5090 and solve by using this PR. The further merge to this PR may be useful to flashinfer.

baonudesifeizhai:
hmm interested...

bkryu:
Thanks @pjh4993, @ZiyiTsang , and @baonudesifeizhai. I trying to put in effort to followup on issues and PRs stuck in limbo.

@baonudesifeizhai, would you be so kind to resolve the merge conflict?

baonudesifeizhai:
yeap  will do

baonudesifeizhai:
could you tell me how to reproduce?@ZiyiTsang
> I face the problem in issue with 5090 and solve by using this PR. The further merge to this PR may be useful to flashinfer.

baonudesifeizhai:
We monkeypatched torch.cuda.current_blas_handle() to return an invalid non-null handle (0x1) during forced cuBLAS FP8 autotuning.
- Main branch: crashed in cublasLtMatmulAlgoGetHeuristic via bmm_fp8_get_algos with exit code 139.
- Rebased fix: did not access the injected PyTorch handle and completed successfully.
- Finding: externally supplied invalid or stale handles can crash the autotuning path. Internally owned, per-thread/per-device cuBLASLt handles isolate FlashInfer from this failure.
- Caveat: this is deterministic fault injection, not proof that PyTorch naturally returns invalid handles.

ZiyiTsang:
> could you tell me how to reproduce?@ZiyiTsang
> 
> > I face the problem in issue with 5090 and solve by using this PR. The further merge to this PR may be useful to flashinfer.

Sure. We hit this while serving an NVFP4-quantized Qwen3.5 (hybrid GDN) with sglang on 2× RTX 5090. This is my env:

**Environment**

| Item | Version |
|---|---|
| GPU | 2× NVIDIA GeForce RTX 5090 (sm_120, 32 GB) |
| Driver | 580.95.05 |
| CUDA (user space) | 13.0 |
| torch | 2.13.0+cu130 |
| flashinfer | 0.6.18 (PyPI wheel, `[cu13]` extra; pinned by sglang 0.5.19) |
| apache-tvm-ffi | 0.1.11 |
| nvidia-cublas | 13.1.1.3 |
| sglang | 0.5.19 |

**reproduces**

sglang routes ModelOpt FP8 linear layers through `flashinfer.bmm_fp8(..., backend="cublas")` on sm90/sm100/sm120. With any NVFP4 checkpoint (FP8 layers included), the server dies during FlashInfer autotune warmup / cuda-graph capture, 4 out of 4 launches:

```
RuntimeError: Check failed: (status == CUBLAS_STATUS_SUCCESS) is false:
bmm_fp8_internal_cublaslt failed: the library was not initialized
```

**A/B experiment** (what convinced us it's the handle, done while debugging)

| Variant | Handle reaching `cublasLtMatmul` | Result |
|---|---|---|
| A — stock 0.6.18 | `torch.cuda.current_blas_handle()` reinterpreted as `cublasLtHandle_t` | `CUBLAS_STATUS_NOT_INITIALIZED`, deterministic |
| B | real `cublasLtCreate()` handle (ctypes, monkeypatched into `torch.cuda.current_blas_handle`) | works, cos-sim 0.999999 vs dequant reference |

I did not run this PR's branch itself — I just use claude to patched `gemm/gemm_base.py` locally following the same approach (internal per-device `cublasLtCreate` handle instead of the reinterpreted torch cuBLAS handle). With that patch all sglang launches pass and output numerics check out.

csrc/bmm_fp8.cu
```diff
@@ -17,11 +17,43 @@
...
 
+namespace {
+
+struct ThreadLocalCublasLtHandles {
+  ~ThreadLocalCublasLtHandles() {
+    // Best-effort cleanup during thread teardown.
+    for (auto& [_, handle] : handles) {
+      if (handle != nullptr) {
+        (void)cublasLtDestroy(handle);
+      }
```
bkryu: `thread_local` destructors for the main thread run at process exit, potentially after the CUDA driver has begun teardown — and calling `cublasLtDestroy` can cause static-destruction-order crash. PyTorch deliberately leaks its cuBLAS handles at shutdown rather than destroying them.

Two safe options:
1. Leak intentionally: delete the destructor and leave a comment stating that handles are deliberately not destroyed because  thread-local destruction can outlive the CUDA context (matching PyTorch's practice). Worker threads that exit mid-process leak one  small handle each — bounded by pool size, harmless.
2. Guard the destroy: bail out of the loop if the context is already gone, e.g.
```
~ThreadLocalCublasLtHandles() {
  int dev;             
  if (cudaGetDevice(&dev) != cudaSuccess) return;  // driver tearing down — leak deliberately
  for (auto& [_, handle] : handles)
    if (handle) (void)cublasLtDestroy(handle);
}
```
2. (Note the guard is best-effort — cudaGetDevice succeeding doesn't fully guarantee the Lt library state is intact, which is why  option 1 is the simpler and battle-tested choice.)

I'd go with option 1.

baonudesifeizhai:
@ZiyiTsang  i rebased ...i did not have 5090 only sm100+ series... did it fixed?

csrc/bmm_fp8.cu
bkryu:
```suggestion
  // Handles are deliberately never destroyed. A thread_local destructor on the
  // main thread runs during process exit, after the CUDA driver may already be
  // tearing down, and calling cublasLtDestroy into a dying context is the
  // classic static-destruction-order crash; following PyTorch in this convention.
  struct ThreadLocalCublasLtHandles {
    std::unordered_map<int, cublasLtHandle_t> handles; 
  }; 
  ```
    
  So the full surrounding code after applying it reads:
  ```C++
  namespace {
  
  // Handles are deliberately never destroyed. A thread_local destructor on the
  // main thread runs during process exit, after the CUDA driver may already be
  // tearing down, and calling cublasLtDestroy into a dying context is the
  // classic static-destruction-order crash; following PyTorch in this convention.
  struct ThreadLocalCublasLtHandles {
    std::unordered_map<int, cublasLtHandle_t> handles;
  };
  
  cublasLtHandle_t get_cublaslt_handle(int device_id) {
    static thread_local ThreadLocalCublasLtHandles cache;
    ...
  }
```

bkryu (commented):
Thanks @baonudesifeizhai for the rebase I left one comment that felt more critical. 

After this I can approve and merge the PR.

bkryu:
@baonudesifeizhai, the PR looks good but can you do a `pre-commit run --all-files` 😅

bkryu:
@baonudesifeizhai, sorry for the confusion. The main branch's pre-commit is currently broken. We have a fix in flight with https://github.com/flashinfer-ai/flashinfer/pull/5113. Currently all PRs are having this issue

Once 5113 is merged, would you be so kind to merge main or rebase?

baonudesifeizhai:
pre-commit run --all-files passes on the PR branch itself. The PR merge check is failing on flashinfer_blackwell_moe_ep_layer_sm100.cu, introduced by the latest upstream main. The current main pre-commit run fails on the same file due to duplicate top-level files: entries. I’ll rebase once the upstream fix lands.
> @baonudesifeizhai, the PR looks good but can you do a `pre-commit run --all-files` 😅

bkryu:
#5113 is merged so we should be good to go. Let me trigger the CI once the pre-commit clears 👍

bkryu:
@flashinfer-bot run

bkryu (approved):

PR #2810, opened 2026-03-17
feat(gdn): add padding index guard for bf16 decode kernel

kaixih:
Clamp negative slot indices to 0 before passing to the bf16 fast-path kernel to prevent out-of-bounds memory access when padding indices (-1) are present in initial_state_indices.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented negative batch indices during decoding to avoid invalid indexing, improving stability and correctness of initial-state handling in batch inference.

* **Tests**
  * Added new bf16 padding-indices test to validate handling of mixed padding and valid indices and related state updates.
  * Marked numerous tests as temporarily skipped due to CI failures to gate unstable coverage.

### DIFF
--- a/tests/gdn/test_decode_delta_rule.py
+++ b/tests/gdn/test_decode_delta_rule.py
@@ -23,7 +23,6 @@ import random
 import torch
 import pytest
 
-pytestmark = pytest.mark.skip(reason="Temporarily skipped due to CI failures.")
 
 try:
     from .reference_delta_rule import decode_delta_rule, verify_delta_rule
@@ -203,6 +202,7 @@ def _test_decode_kernel_pretranspose(
     "num_q_heads, num_k_heads, num_v_heads",
     [(16, 16, 32)],
 )
+@pytest.mark.skip(reason="Temporarily skipped due to CI failures.")
 @pytest.mark.parametrize("batch_size", [1, 2, 4, 8, 16, 32, 64, 128, 256, 512])
 @pytest.mark.parametrize("dtype", ["bfloat16"])
 def test_decode_kernel_basic_pretranspose(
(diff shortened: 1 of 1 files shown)


kaixih:
This resolves the issue: https://github.com/sgl-project/sglang/issues/20791

@zihaoye PTAL

flashinfer/gdn_kernels/gdn_decode_bf16_state.py
```diff
@@ -1994,6 +1994,7 @@ def gated_delta_rule(
     else:
         h_slot_indices = initial_state_indices
 
+    h_slot_indices = h_slot_indices.clamp(min=0)  # guard -1 padding
```
yzh119: Can we use inplace update: https://docs.pytorch.org/docs/stable/generated/torch.Tensor.clamp_.html

kaixih: I feel the out-of-place clamp (current state) is safer. After the call, if the user inspects their initial_state_indices (aliased as h_slot_indices here) tensor, the padding markers are gone, ie. silently corrupted. This would break any logic that relies on those -1s after the decode step (e.g. a loop that checks which slots are padding).

Ideally, we should have change the inside of the kernel to avoid loading when idx < 0. I don't have time to test that now.

kaixih: it is actually easier to change. I modified the kernel and let it to use 0 if the index is negative. PTAL.

kaixih:
Redirect negative pool_batch_idx to slot 0 (null buffer) inside all 3 bf16 decode kernel variants to prevent OOB memory access on padding indices. Also adds a unittest and enables it in CI by replacing the stale module-level skip with per-test skips.

tests/gdn/test_decode_delta_rule.py
```diff
@@ -203,6 +202,7 @@ def _test_decode_kernel_pretranspose(
     "num_q_heads, num_k_heads, num_v_heads",
     [(16, 16, 32)],
 )
+@pytest.mark.skip(reason="Temporarily skipped due to CI failures.")
```
yzh119, line 205: Do we still need this? cc @bkryu (seems it was first introduced in https://github.com/flashinfer-ai/flashinfer/pull/2600).

PR #2818, opened 2026-03-19
fix: fall back to CUDA events when CUPTI driver version < 13.0

sha7doww:
## Description

On systems where `cupti-python >= 13` is installed but the CUDA **driver** is older than 13.0, `cupti.activity_enable()` raises `NotSupportedError`. Previously this call happened **after** the `try` block, so the exception was unhandled and `bench_gpu_time_with_cupti` crashed instead of falling back to CUDA events.

**Root cause:** The driver-support check was missing from the guarded import block.

**Fix:** Added a probe (`activity_enable` + `activity_disable` on `RUNTIME`) inside the existing `try` block so the `NotSupportedError` triggers the existing CUDA-event fallback path.

## Related Issues

Fixes a crash when using `bench_gpu_time(enable_cupti=True)` on machines with CUDA driver < 13.0 but `cupti-python >= 13` installed.

## Pre-commit

- [x] `pre-commit run --all-files` passes (all 14 hooks green)

## Tests

- [x] Added `tests/utils/test_cupti_fallback.py` — mocks `cupti.activity_enable` to raise an exception and verifies:
  - `bench_gpu_time_with_cupti` does **not** crash
  - Returns valid timing results via the CUDA-event fallback
  - Emits the expected `UserWarning` about falling back

```
$ pytest tests/utils/test_cupti_fallback.py -v
PASSED
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GPU benchmarking robustness by verifying both CUPTI version and runtime driver support before use; falls back to alternative CUDA benchmarking when CUPTI is unavailable or activity enabling fails, emitting a warning.

* **Tests**
  * Added a test validating CUPTI fallback behavior, ensuring timing results remain positive and a fallback warning is issued when CUPTI activity enabling errors.

### DIFF
--- a/flashinfer/testing/utils.py
+++ b/flashinfer/testing/utils.py
@@ -1006,7 +1006,7 @@ def bench_gpu_time_with_cupti(
         # Use 2x L2 size to ensure complete flush
         _l2_flush_size_mb = (l2_size * 2) // (1024 * 1024)
 
-    # check if CUPTI is installed and its version is >= 13.0.0
+    # check if CUPTI is installed, version >= 13, and the driver supports it
     try:
         from cupti import cupti
         from importlib.metadata import version as importlib_metadata_version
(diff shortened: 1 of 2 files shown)


yzh119 (commented):
Hi @bkryu can you help review?

PR #2811, opened 2026-03-18
CuteDSL MoE fix redundant output buffer zeroing

leejnau:
## 📌 Description

The CuteDSL MoE pipeline redundantly zeroed the entire max_num_tokens output buffer before each GEMM2 scatter-add, costing ~3.7 ms/fwd across 61 layers in DeepSeek R1. This PR replaces it with a dense zero of only the active [:num_tokens] slice, overlapped with GEMM1 on an auxiliary stream — matching TRT-LLM's original zeroing strategy.

## 🔍 Related Issues

[feat: cuteDSL fp4 moe for better DSR1 performance.](https://github.com/flashinfer-ai/flashinfer/pull/2398)

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [x] Tests have been added or updated as needed.
- [x] All tests are passing (`unittest`, etc.).

## Summary by CodeRabbit

* **Refactor**
  * Improved memory management and per-batch output slicing for CUDA-graph execution; allocation/zeroing now targets only the active output slice.
* **Bug Fixes / Reliability**
  * Added validation of provided output buffers; callers must supply zero-initialized preallocated outputs to avoid corruption (runtime no longer implicitly zeroes caller buffers).
* **Documentation**
  * Clarified output ownership, zero-initialization contract, and atomic-accumulate behavior in user-facing docs.

PR #2775, opened 2026-03-12
chore: cute dsl nvfp4 moe clean up

nv-yunzheq:
## 📌 Description

This is to clean up cute dsl nvfp4 moe
1. Remove incorrect statement in code that sm110 is supported
2. Add cute dsl moe to benchmark script
3. Adjust autouning strategy and tactics. And use unit test to test all tactics
4. Remove unused blockscaled fp4 grouped gemm in moe
5. Make pdl as a function parameter instead of environment variable to align with the rest of library
6. Add CC support decorator to cute dsl moe function

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [ ] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [ ] I have installed the hooks with `pre-commit install`.
- [ ] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [ ] Tests have been added or updated as needed.
- [ ] All tests are passing (`unittest`, etc.).

## Summary by CodeRabbit

* **New Features**
  * Added CuteDSL FP4 block-scaled MoE benchmark and end-to-end test path; PDL (prefetch-dependent launch) is now runtime-configurable.

* **Bug Fixes**
  * Disabled the problematic tile-size=256 tactic.
  * Clarified supported Blackwell architectures to SM100 family (SM100/SM103).

* **Refactor**
  * Simplified MoE tactic selection to runtime validation with deterministic tactic shapes.
  * Replaced environment-based PDL flag with a runtime parameter.

* **Tests**
  * Added comprehensive MoE tactic validation tests and benchmark samples (some entries duplicated).

PR #2398, opened 2026-01-22
feat: cuteDSL fp4 moe for better DSR1 performance.

nv-yunzheq:
## 📌 Description

cuteDSL fp4 moe from TRTLLM with fusion. issue #2259
The collective fusion performance improvement collected from TRTLLM [PR 8880](https://github.com/NVIDIA/TensorRT-LLM/pull/8880),  [PR 9288](https://github.com/NVIDIA/TensorRT-LLM/pull/9288),  [PR 9618](https://github.com/NVIDIA/TensorRT-LLM/pull/9618),.

Introducing two new API for moe
`cute_dsl_fused_moe_nvfp4`, `CuteDslMoEWrapper`. 
The two API are functionally equivalent. The `cute_dsl_fused_moe_nvfp4` function could run the operation directly, while the wrapper API splits tensor setup and execution to better support cuda graph. Using wrapper API is recommended.

The PR also introduces autotune functionality for this function.

Performance data:
```
$ python benchmarks/bench_moe_deepseek.py --num-tokens 1,2,4,8,16,32,64,128,256,512,1024,2048,4096,8192,16384
====================================================================================================
DeepSeek-V3 MoE Benchmark: CuteDSL vs CUTLASS vs TRTLLM (EP=1)
====================================================================================================
Model: hidden=7168, intermediate=2048, experts=256, top_k=8
EP Config: 256 local experts (simulating 1-way parallelism)
CUDA Graph: enabled, CUPTI: enabled
----------------------------------------------------------------------------------------------------
Tokens |     CuteDSL     |     CUTLASS     |     TRTLLM      | Speedup (CuteDSL/X) |  Winner
       |      ms  TFLOPS |      ms  TFLOPS |      ms  TFLOPS |  CUTLASS   TRTLLM |
----------------------------------------------------------------------------------------------------
     1 |   0.064    10.9 |   0.098     7.2 |   0.053    13.2 |    1.53x    0.83x |  TRTLLM
     2 |   0.077    18.2 |   0.109    13.0 |   0.064    21.9 |    1.40x    0.83x |  TRTLLM
     4 |   0.097    28.9 |   0.131    21.5 |   0.086    32.7 |    1.35x    0.88x |  TRTLLM
     8 |   0.098    57.5 |   0.133    42.4 |   0.092    61.4 |    1.36x    0.94x |  TRTLLM
    16 |   0.102   110.3 |   0.138    82.0 |   0.104   108.5 |    1.35x    1.02x | CuteDSL
[...]

PR #2840, opened 2026-03-20
Fix CUDA stream race in CuteDSL MoE when CUDA graphs are disabled

leejnau:
## 📌 Description

Gate async memset on CUDA graph mode in the CuteDSL MoE wrapper. When CUDA graphs are enabled, the same pre-allocated moe_output buffer is reused on every graph replay, so zeroing it on a dedicated per-instance auxiliary stream overlaps with GEMM1 and hides the cost. When CUDA graphs are disabled, buffers are allocated dynamically each call and don't need async zeroing — a synchronous .zero_() on the main stream is sufficient and avoids a race condition where multiple MoE layers would share the module-level singleton stream from _get_cuda_graph_resources().

The fix moves stream and event creation from _allocate_buffers() into the if use_cuda_graph: block of __init__, so non-graph wrappers never create CUDA stream resources they don't need. The functional API (_cute_dsl_fused_moe_nvfp4_impl) is similarly updated to only use async memset when the caller provides an explicit auxiliary stream.

## 🔍 Related Issues

[feat: cuteDSL fp4 moe for better DSR1 performance.](https://github.com/flashinfer-ai/flashinfer/pull/2398)
[chore: cute dsl nvfp4 moe clean up](https://github.com/flashinfer-ai/flashinfer/pull/2775)
[CuteDSL MoE fix redundant output buffer zeroing](https://github.com/flashinfer-ai/flashinfer/pull/2811/changes)

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/flashinfer/fused_moe/cute_dsl/fused_moe.py
+++ b/flashinfer/fused_moe/cute_dsl/fused_moe.py
@@ -388,11 +388,18 @@ class CuteDslMoEWrapper:
         self.device = device
         self.enable_pdl = enable_pdl
 
-        # Pre-allocated buffers
+        # Pre-allocated tensor buffers (only when CUDA graphs are enabled)
         self._moe_sort_buffers: Optional[Dict[str, torch.Tensor]] = None
         self._gemm1_output: Optional[torch.Tensor] = None
         self._gemm1_output_scale: Optional[torch.Tensor] = None
         self._moe_output: Optional[torch.Tensor] = None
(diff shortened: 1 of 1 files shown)


nvpohanh:
@leejnau is this still needed?

leejnau:
> @leejnau is this still needed?

No.  I'll close it.

PR #2841, opened 2026-03-20
[Perf] Add FMHAv2 to flashinfer_benchmark.py and eliminate unnecessary H2D

jimmyzho:
## 📌 Description

This PR wires `trtllm_fmha_v2_prefill` into the `flashinfer_benchmark` harness and fixes several CUDA-graph-incompatible allocations on the FMHAv2 prefill hot path.

### Benchmark integration

- Add `trtllm-fmha-v2` as a backend option for `BatchPrefillWithPagedKVCacheWrapper` and `BatchPrefillWithRaggedKVCacheWrapper` in `benchmarks/routines/attention.py`.
- Register the backend in `flashinfer_benchmark_utils.py` compute-capability map for SM90 and SM120.
- Add it to the refcheck reference priority list (cudnn > cudnn-native > trtllm-gen > trtllm-native > trtllm-fmha-v2).
- For paged runs, thread a separate `kv_token_indptr` (cumulative token counts) since TRT-LLM expects token-based indptr while FlashInfer's paged wrapper uses page-based.
- For paged runs, `kv_cache.contiguous()` is only materialized when `trtllm-fmha-v2` is in the backend list (FlashInfer's `as_strided` HND presentation does not match the physical layout the kernel expects).

### CUDA-graph fixes

1. **`scale_bmm2` lives entirely in the kernel arg buffer.** The previous code path required a device-resident copy of the BMM2 scale (`params.scale_bmm2_d`) so the SM90 warp-specialized epilogue could dereference it. Populating that buffer required either a per-call host tensor allocation (graph-incompatible) or a `cudaMemcpyAsync` from a pageable host source (also graph-incompatible — the source pointer is captured by the memcpy node and goes out of scope after the wrapper returns). The fix patches the WS epilogue (`csrc/fmha_v2/fmha/warpspec/epilogue.h:1136`) to read the host-encoded `params.scale_bmm2` directly — that uint32 is captured by value into the kernel arg buffer, so replays naturally pick up new scale values. With the WS epilogue fixed, the binding stops allocating any device slot, stops issuing any memcpy, and Python passes a plain `float`. Non-WS epilogues (`hopper/`, `gmem_tile_o*.h`) already read `params.scale_bmm2` either directly or via a `d ? *d : scale_bmm2` fallback, so this change is local to the SM90 WS path; SM80/86/89/120/121 are unaffected. Marked `// DIVERGENCE` in the vendored TRT-LLM source for the next upstream rebase.
2. **Conditional `softmax_stats` allocation.** Removed the unconditional workspace allocation; `softmax_stats_ptr` is `nullptr` unless the caller provides an output tensor.
3. **Eliminated `expanded_block_tables`.** Previously, `trtllm_fmha_v2_prefill` called `torch.stack([bt*2, bt*2+1], dim=1).contiguous()` on every invocation to pre-expand `[B, M]` logical page indices into `[B, 2, M]` K/V pool offsets. This allocated 2–4 temporary GPU tensors per call and could trigger caching-allocator synchronization under memory pressure. The fix adds a `mUsesSharedPagedKvIdx` flag to `Kv_block_array` and teaches both the Ampere tiled path (`gmem_tile_qkv_packed.h`) and the Hopper warp-specialized DMA path (`warpspec/dma.h`) to compute the interleaved pool offset (`page_idx * 2 + kv_type`) on-the-fly during block-offset loads. The original `[B, M]` block_tables is now passed straight through to the kernel. The flag defaults to `false` (via constructor init + `memset`), so all legacy `[B, 2, M]` callers (e.g., `trtllm_fmha_v2_binding.cu`) are unchanged.
4. **`scale_softmax = 1.0` unconditionally.** The C++ auto-detect (when `scale_softmax == 0.0`) covers FP16/INT8/E4M3 but has no BF16 branch, which would zero out the softmax output. Passing 1.0 explicitly avoids the bug.

CUDA graph wrapping remains disabled for `trtllm-fmha-v2`, `trtllm-native`, and `cudnn-native` in the benchmark runners pending further validation.

### Tests

- `test_fmha_v2_prefill.py`: temporarily skip FP8 dtype combinations in the main parametrize block due to unmitigated hangs (tracked inline). Narrow one FP8 case from `float16` → `bfloat16` output dtype.

## 🔍 Related Issues

None.

## 🧪 Tests

- [x] Existing `tests/attention/test_fmha_v2_prefill.py` updated; FP16/BF16 paged + ragged layouts pass on SM90/SM120.
- [ ] FP8 tests re-enabled once hangs are resolved.

## Reviewer Notes

- The WS epilogue patch in `csrc/fmha_v2/fmha/warpspec/epilogue.h` is a one-line behavioral change (`*params.scale_bmm2_d` → `params.scale_bmm2`) on a vendored TRT-LLM kernel. Annotated with a `DIVERGENCE` comment so the next upstream rebase doesn't silently revert it. Other consumers of `params.scale_bmm2_d` (the Hopper non-WS and Ampere/Blackwell paths in `gmem_tile_o*.h`) already read it via `d ? *d : scale_bmm2`; with `scale_bmm2_d` now consistently `nullptr` in the FFI path they fall through to `params.scale_bmm2`, which `set_alpha` already populates. No kernel re-tuning needed.
- The `expanded_block_tables` elimination is the largest functional change — see `csrc/fmha_v2/fmha/paged_kv_cache.h`, `gmem_tile_qkv_packed.h`, `warpspec/dma.h`, and `fmha_v2_run.cu`. The kernel-side cost is one int multiply-add and a uniform branch per block-offset load (hidden by global memory / TMA latency); the host-side win is eliminating 2–4 tensor allocations per call.


benchmarks/routines/attention.py
```diff
@@ -1158,6 +1171,16 @@ def to_float8(x, dtype=torch.float8_e4m3fn):
         v_quantized, _ = to_float8(v_data, kv_dtype)
         kv_cache = torch.cat([k_quantized, v_quantized], dim=1)
 
+    _fmha_v2_bmm2_scale = v_scale if v_scale is not None else 1.0
+
+    # trtllm-fmha-v2 expects physically-HND paged KV, but FlashInfer's kv_cache uses
+    # as_strided to present HND shape with NHD physical layout. Make a contiguous copy
+    # so the kernel reads data in the correct order.
+    if "trtllm-fmha-v2" in backends:
+        _fmha_v2_kv_cache = kv_cache.contiguous()
```
saltyminty, line 1182: This check needs to also filter for non-fp4, since kv_cache would be a tuple and crash on attempting .contiguous.

tests/attention/test_fmha_v2_prefill.py
```diff
@@ -859,7 +859,7 @@ def test_trtllm_fmha_v2_prefill(
     [
         (torch.float16, torch.float16),
         (torch.bfloat16, torch.bfloat16),
-        (torch.float8_e4m3fn, torch.float16),
+        (torch.float8_e4m3fn, torch.bfloat16),
```
saltyminty, line 862: Just verifying that this is intended?

saltyminty (changes requested):

tests/attention/test_fmha_v2_prefill.py
jimmyzho: Yes, accidentally changed the fp8 test in this [PR to skip fp8 tests](https://github.com/flashinfer-ai/flashinfer/pull/3050) so was never caught. Reversing this now.

benchmarks/routines/attention.py
jimmyzho: Filtered for tensor kv_cache only. Fp4 will be removed from list of backends later for this

csrc/fmha_v2_run.cu
```diff
@@ -606,22 +595,27 @@ void fmha_v2_run(
...
-      tokens_per_block, qkv_packed_d, q_d, k_d, v_d, contiguous_kv_d, kv_cache_pool_ptr,
-      kv_cache_block_offsets_d, packed_mask_d, nullptr, attention_sinks_d,
-      static_cast<void*>(cum_seq_lens_kv.data_ptr()), static_cast<void*>(cum_seq_lens_q.data_ptr()),
-      o.data_ptr(), nullptr, nullptr, softmax_stats_ptr, scale_bmm2_d.data_ptr(), scale_bmm1,
-      scale_softmax, scale_bmm2, softcapping_scale_bmm1, false, false, false, has_alibi,
-      skip_softmax_threshold_scale_factor);
-
-  // For Q_PAGED_KV layout, override mMaxBlocksPerSeq to match the actual block_tables stride
-  // that we used when expanding the block offsets from [B, M] to [B, 2, M]
+  set_params(params_v2, launch_params, data_type, acc_type, output_dtype, input_layout, b, s_q, s,
```
saltyminty: is_paged_hnd was removed from this callsite for `set_params` but not removed from the associated inline function

jimmyzho: thanks for catch, this was an accident.

akhilg-nv:
This approach for fixing the scale_bmm2 H2D & tiny kernel LGTM. Do we have a trace with these changes to confirm all the small kernels are removed?

jimmyzho:
@akhilg-nv Running nsys on one of the benchmark routines:
Screenshot 2026-05-05 at 9 56 21 AM

saltyminty (approved):

saltyminty:
Per offline discussion: ragged tests/benchmarks need a minor refactor to account for sm120 fallback on `SEPARATE_Q_K_V`

PR #2848, opened 2026-03-21
[MoE] Fix weight_scale_vec_size check for models with non-aligned scale dimensions

Tib-Gridello:
## Summary

`nvfp4_block_scale_interleave` adds alignment padding to scale tensors. For models where `hidden_size / sf_block_size` is not a multiple of 4, this inflates `numel()` and makes `weight_scale_vec_size` compute to values like 31 instead of 32, failing the strict `== 16 || == 32` check.

**Example:** gpt-oss-120b has `hidden_size=2880`, `sf_block_size=32` → scale cols = 90 (not divisible by 4). After interleaving: 518,400 → 529,920 elements. `weight_scale_vec_size` = 31 instead of 32.

## Change

Relax the check from `== 16 || == 32` to `>= 14 && <= 32`. The `MxE2m1` dtype selection (`<= 16 ? E2m1 : MxE2m1`) remains correct. Also adds the actual value to the error message.

## Impact

- Models with `hidden_size % 128 == 0` (e.g. Llama, Qwen): `weight_scale_vec_size` = 32 exactly → no change
- Models with non-aligned dimensions (e.g. gpt-oss-120b: 2880): `weight_scale_vec_size` = 31 → now passes

## Testing

Tested on RTX PRO 6000 Blackwell Max-Q (SM120) with gpt-oss-120b MXFP4 via vLLM 0.18.0 + `VLLM_USE_FLASHINFER_MOE_MXFP4_MXFP8_CUTLASS=1`:
- Output correctness verified (math, code gen, JSON, 5-run determinism)
- 28% faster prompt processing vs Marlin fallback

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error messages with detailed diagnostic information for MOE kernel operations.
  * Expanded parameter validation range to support a wider set of configurations.

### DIFF
--- a/csrc/trtllm_fused_moe_kernel_launcher.cu
+++ b/csrc/trtllm_fused_moe_kernel_launcher.cu
@@ -1996,8 +1996,8 @@ Array<Tensor> trtllm_fp4_block_scale_moe(
       (local_num_experts * intermediate_size * intermediate_size_factor * hidden_size) /
       gemm1_weights_scale.numel();
 
-  TVM_FFI_ICHECK(weight_scale_vec_size == 16 || weight_scale_vec_size == 32)
-      << "unsupported weight_scale_vec_size.";
+  TVM_FFI_ICHECK(weight_scale_vec_size >= 14 && weight_scale_vec_size <= 32)
+      << "unsupported weight_scale_vec_size: " << weight_scale_vec_size;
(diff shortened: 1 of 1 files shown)


Tib-Gridello:
Closing — on further review, this check is only hit by the TRTLLM path, not the CUTLASS path that SM120 uses. The fix may be correct but we only verified output correctness on the CUTLASS path, not TRTLLM. Filing as an issue instead for maintainers who understand the TRTLLM kernel memory layout to evaluate.

PR #2866, opened 2026-03-23
[fmha-v2] Remove H2D transfer for bmm2 scale, enabling cuda graphs

akhilg-nv:
## 📌 Description

Removes a host to device transfer for setting the bmm2 scale, enabling cuda graphs. Potentially, we can also enable the API to accept a tensor for this scale, such that it can be computed elsewhere and passed in as a constant.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Chores**
  * Internal performance optimizations for tensor creation; no user-visible changes.

* **Refactor**
  * Simplified and standardized tensor construction logic to improve reliability and consistency across numeric types, reducing internal complexity without impacting functionality.

### DIFF
--- a/flashinfer/prefill.py
+++ b/flashinfer/prefill.py
@@ -121,7 +121,7 @@ def _create_scale_bmm2_d_tensor(
         return result
     else:
         # FP8, INT8, etc. use FP32 accumulation - create FP32 tensor and view as int32
-        return torch.tensor([scale_bmm2], dtype=torch.float32, device=device).view(
+        return torch.full((1,), scale_bmm2, dtype=torch.float32, device=device).view(
             torch.int32
         )
 
--- a/flashinfer/prefill.py
+++ b/flashinfer/prefill.py
@@ -111,14 +111,18 @@ def _create_scale_bmm2_d_tensor(
     """
     if data_dtype == torch.float16:
         # Create int32 buffer on device, write FP16 value to lower 16 bits via view
-        result = torch.zeros(1, dtype=torch.int32, device=device)
-        result.view(torch.float16)[0] = scale_bmm2
-        return result
+        return (
+            torch.full((1,), scale_bmm2, dtype=torch.float16, device=device)
+            .view(torch.uint16)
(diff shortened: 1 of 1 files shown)


flashinfer/prefill.py
```diff
@@ -111,17 +111,21 @@ def _create_scale_bmm2_d_tensor(
     """
     if data_dtype == torch.float16:
         # Create int32 buffer on device, write FP16 value to lower 16 bits via view
-        result = torch.zeros(1, dtype=torch.int32, device=device)
-        result.view(torch.float16)[0] = scale_bmm2
-        return result
+        return (
+            torch.full((1,), scale_bmm2, dtype=torch.float16, device=device)
```
nvpohanh, line 115: I think this still causes a tiny Fill kernel. 

If we want to eliminate this kernel as well, the solution would be to accept `bmm1_scale` and `bmm2_scale` as a torch.Tensor so that the framework (like SGLang) can provide the scales as device tensors directly (and framework can cache that across decoding steps). 

https://github.com/akhilg-nv/flashinfer/blob/bdf29115facde5097b050c5ffdf60f0eae9826f9/flashinfer/prefill.py#L4088-L4089

See this as an example: https://github.com/akhilg-nv/flashinfer/blob/bdf29115facde5097b050c5ffdf60f0eae9826f9/flashinfer/prefill.py#L3725-L3726

nvpohanh: cc @jimmyzho

akhilg-nv: Correct, Jimmy has this draft [PR](https://github.com/flashinfer-ai/flashinfer/pull/2841) up which allows the scale to be `Union[float, torch.Tensor]`. I think we will still want to keep the logic in this PR for the case where the input is a float, but perhaps it may be better to force it to be a tensor?

akhilg-nv:
Closing as its addressed by https://github.com/flashinfer-ai/flashinfer/pull/2841

Issue #2887, opened 2026-03-24
[DeepSeek][Perf] Post-MoE AllReduce trigger_completion_at_end=False


benchislett:
TRTLLM AR+Norm Fusion kernel uses `trigger_completion_at_end=launch_with_pdl`, and does not expose this option to the caller interface.

Using `trigger_completion_at_end=False` can enable **4%+ E2E speedup for DeepSeek** in vLLM. FlashInfer should either expose this to the caller, or set it to False internally. Setting it to `launch_with_pdl` effectively nullifies the option.

```
        trtllm_allreduce_fusion(
            allreduce_in=input_flat,
            world_size=workspace.world_size,
            world_rank=workspace.rank,
            token_num=token_num,
            hidden_dim=hidden_dim,
            workspace_ptrs=workspace.workspace_tensor,
            launch_with_pdl=launch_with_pdl,
            trigger_completion_at_end=launch_with_pdl,  # Same meaning
            fp32_acc=fp32_acc,
            pattern_code=pattern,  # type: ignore[arg-type]
            use_oneshot=use_oneshot,
            allreduce_out=output_flat,
            residual_in=residual_in_flat,
            residual_out=residual_out_flat,
            norm_out=norm_out_flat,
            quant_out=quant_out_flat,
            scale_out=scale_out,  # scale_out is not reshaped
            rms_gamma=rms_gamma,  # 1D tensor, no reshape needed
            rms_eps=rms_eps,
            scale_factor=scale_factor,
            layout_code=layout_code,  # type: ignore[arg-type]
            metadata=workspace.metadata,
        )
```

This leads to very little PDL overlap between the Fused-AR and the subsequent QKV Projection in DeepSeek:

<img width="1826" height="97" alt="Image" src="https://github.com/user-attachments/assets/ce2915f5-b463-4de3-8245-1fbfada1e8e9" />

Changing it to `False` will have the PDL trigger early:
```
  cudaGridDependencySynchronize();
  if constexpr (!TriggerCompletionAtEnd) {                                                                                                                                                                                                                                                                                              
      cudaTriggerProgrammaticLaunchCompletion();
  }
```

<img width="1491" height="97" alt="Image" src="https://github.com/user-attachments/assets/12c02c89-8166-4e6a-baa2-c8c2538b921d" />

[...]

PR #2894, opened 2026-03-25
fix: expose trigger_completion_at_end through unified API

nv-yunzheq:
## 📌 Description

Fix issue #2887, original logic of `trigger_completion_at_end` equals to `enable_pdl` is wrong. Expose as a separate parameter in the unified API.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * Added an optional control to choose when kernel completion is signaled, enabling early signaling to allow overlap with follow-up work or deferring until kernels finish for safety; execution paths now respect this option to improve collective communication flexibility and performance.

* **Documentation**
  * Clarified API semantics and updated wording around programmatic dependent launch and completion behavior.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,48 @@
+# Configuration for actions/labeler
+# Docs: https://github.com/actions/labeler
+
+"op: attention":
+  - changed-files:
+      - any-glob-to-any-file:
+          - "flashinfer/attention.py"
+          - "flashinfer/decode.py"
+          - "flashinfer/prefill.py"
+          - "flashinfer/mla/**"
+          - "flashinfer/xqa.py"
+          - "flashinfer/jit/attention/**"
+          - "include/flashinfer/attention/**"
+          - "csrc/fmha_v2/**"
+          - "csrc/xqa/**"
+          - "csrc/cutlass_mla*"
+          - "tests/attention/**"
+
+"op: moe":
+  - changed-files:
+      - any-glob-to-any-file:
+          - "flashinfer/fused_moe/**"
+          - "flashinfer/jit/fused_moe.py"
+          - "flashinfer/jit/moe_utils.py"
+          - "csrc/fused_moe/**"
+          - "csrc/trtllm_fused_moe*"
+          - "csrc/moe_utils*"
+          - "include/flashinfer/trtllm/fused_moe/**"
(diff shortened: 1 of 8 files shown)


flashinfer/comm/allreduce.py
```diff
@@ -611,7 +619,9 @@ def _flatten_checked(t, name):
             hidden_dim=hidden_dim,
             workspace_ptrs=workspace.workspace_tensor,
             launch_with_pdl=launch_with_pdl,
-            trigger_completion_at_end=launch_with_pdl,  # Same meaning
+            trigger_completion_at_end=trigger_completion_at_end
+            if trigger_completion_at_end is not None
+            else launch_with_pdl,
```
benchislett: I'm not entirely sure this default is needed. The PDL sync will be a no-op if PDL is not enabled anyways, so there's no real difference if it's at the start or the end. Might be easier to just have it be `trigger_completion_at_end: bool = True` with the default being the safe option, and can be overrided to False for the opt-in optimization

benchislett: Or just have it default to True when left as None

nv-yunzheq: Agree, addressed in the latest commit.

flashinfer/comm/allreduce.py
```diff
@@ -494,7 +495,14 @@ def allreduce_fusion(
...
                  Note: MNNVL only supports patterns 0 and 1
-        launch_with_pdl: Use Persistent Dependency Launch
+        launch_with_pdl: Use Programmatic Dependent Launch
+        trigger_completion_at_end: Controls when PDL completion is signaled.
+                     None (default): follows launch_with_pdl for backward compatibility.
+                     True: signal completion after the kernel finishes (safe, no overlap).
+                     False: signal completion early, allowing the next PDL-aware kernel
+                     to overlap with this one. Only safe when the subsequent kernel also
+                     uses cudaGridDependencySynchronize(). Can yield 4%+ E2E speedup
+                     for models like DeepSeek.
```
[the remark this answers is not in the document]
benchislett: Agree with the bot, MNNVL has it hard-coded in the middle-ish of the kernel. Should be made clear that this is not a universal option

PR #2897, opened 2026-03-26
Add Relu2 activation support in CUTLASS MoE backend and fix autotuner async CUDA error handling

askliar:
## Summary
- **Add Relu2 (squared ReLU) activation support** for the CUTLASS MoE GEMM path, enabling models that use Relu2 as their MoE gate activation (e.g. Nemotron-H MTP draft model) to run through the FlashInfer CUTLASS backend instead of throwing `InvalidType` at runtime.
- **SM121 (GB10) tile filtering for CUTLASS MoE GEMM** — GB10 has ~99 KB SMEM per block (vs ~228 KB on GB200/SM120). Three of the four FP4 grouped GEMM tiles and several Ampere-style bf16 tiles exceed this budget and fail at runtime. This PR filters them out at the runtime tactic selection level so the autotuner never tries known-bad configurations.
- **Fix autotuner async CUDA error propagation** — when a tactic probe launches a kernel that fails asynchronously (e.g. `cudaErrorIllegalInstruction`), the sticky error was not cleared and would surface later during CUDA graph capture or inference. Added `torch.cuda.synchronize()` drain and demoted the "Skipping tactic" log from WARNING to DEBUG since these failures are expected and recoverable.
- **Separate JIT cache key for SM121** — SM12.1 now normalizes to arch key `121f` instead of sharing `120f` with SM12.0, ensuring the JIT cache is isolated per architecture.
## Depends on
This PR should be merged **after** #2913 (enable GDC for CUTLASS fused MoE PDL on SM12x), which adds the `-DCUTLASS_ENABLE_GDC_FOR_SM100=1` compile flags that this PR's SM120 module generator relies on.
### Relu2 activation (`epilogue_helpers.h`, `moe_gemm_template_dispatch.h`)
Two pieces were missing:
1. `epilogue_helpers.h` had no `EpilogueOpDefaultRelu2` tag struct or `Epilogue` partial specialization, so there was no CUTLASS epilogue type for Relu2.
2. `moeGemmBiasAct()` had no `case ActivationType::Relu2`, causing it to fall through to `InvalidType` and throw.
The `Relu2` functor itself (`relu(x)²`) already existed in `fused_activations.h` — this PR just wires it into the epilogue dispatch.
### SM121 tile filtering (`cutlass_heuristic.cpp`, `compilation_context.py`)
| GPU | Arch | SMEM/block |
|-----|------|-----------|
| GB200 | SM120 | ~228 KB |
| GB10 | SM121 | ~99 KB |
**FP4 grouped GEMM (TMA Warp-Specialized path):**
- Added `get_candidate_configs_sm121()` that returns only `CtaShape128x128x64B` (~73 KB, fits in 99 KB). The other three SM120 tiles (128x128x128B, 256x128x64B, 128x256x64B) all exceed 99 KB.
- Changed dispatch from `sm >= 120` to `sm == 120` + separate `sm == 121` branch.
**Ampere-style bf16/FP8 grouped GEMM (SM80 path):**
- Added `filter_sm121` lambda in `get_candidate_tiles()` that removes tiles where both M >= 128 and N >= 128, since these also exceed SM121's SMEM budget.
**JIT cache separation:**
- `compilation_context.py`: SM12.1 now returns `(12, "1f")` instead of `(12, "0f")`, giving GB10 cache path `~/.cache/flashinfer/<version>/121f/` separate from GB200's `120f/`.
### Autotuner robustness (`autotuner.py`)
- Demoted "Skipping tactic" log from `WARNING` to `DEBUG` — tile-incompatibility failures are expected on SM121 and should not spam logs.
- Added `torch.cuda.synchronize()` drain after failed tactic probes to clear sticky async CUDA errors (e.g. `cudaErrorIllegalInstruction` from failed TMA WS GEMM probes) before they surface during CUDA graph capture.

## Test plan

- Existing MoE tests pass (`pytest tests/moe/`)
- Relu2-specific tests: `pytest tests/moe/test_trtllm_gen_fused_moe.py -k Relu2`
- Manual: run Nemotron-H MTP on GB10 (SM121) with FlashInfer CUTLASS backend:
  - No `WARNING [Autotuner]: Skipping tactic` log spam
  - No sticky CUDA errors during graph capture
  - Only `CtaShape128x128x64B` FP4 grouped GEMM tile selected
  - JIT cache at `~/.cache/flashinfer/0.6.7/121f/`

### DIFF
--- a/csrc/nv_internal/tensorrt_llm/kernels/cutlass_kernels/cutlass_heuristic.cpp
+++ b/csrc/nv_internal/tensorrt_llm/kernels/cutlass_kernels/cutlass_heuristic.cpp
@@ -128,6 +128,20 @@ std::vector<CutlassTileConfig> get_candidate_tiles(
     gemm_type = CutlassGemmType::Fp8;
   }
 
+  // SM121 (GB10) has ~99 KB SMEM — Ampere-style tiles where both M>=128 and N>=128
+  // exceed the SMEM budget. Filter them out so the autotuner doesn't waste time on
+  // known-bad configs.
+  auto filter_sm121 = [sm](std::vector<CutlassTileConfig> configs) {
+    if (sm != 121) return configs;
+    std::vector<CutlassTileConfig> filtered;
+    for (auto const& c : configs) {
+      TileShape ts = get_cta_shape_for_config(c);
+      if (ts.m >= 128 && ts.n >= 128) continue;
+      filtered.push_back(c);
+    }
+    return filtered;
+  };
+
   std::vector<CutlassTileConfig> base_configs{
       CutlassTileConfig::CtaShape32x128x64_WarpShape32x32x64,
       CutlassTileConfig::CtaShape64x128x64_WarpShape32x64x64};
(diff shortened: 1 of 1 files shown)


flashinfer/jit/fused_moe.py
```diff
@@ -73,6 +75,7 @@ def gen_cutlass_fused_moe_sm100_module(use_fast_build: bool = False) -> JitSpec:
         "-DENABLE_FP8",
         "-DENABLE_FP4",
         "-DUSING_OSS_CUTLASS_MOE_GEMM",
+        "-DCUTLASS_ENABLE_GDC_FOR_SM100=1",
```
amirkl94, line 78: @askliar What does this flag do?
Is there a way to add it in some common flags area or something?

amirkl94:
General question, why is the change in https://github.com/flashinfer-ai/flashinfer/pull/1954 not enough for your case? It probably goes for a different code-path but I'd like to make sure we're not missing a different bug that causes this.

flashinfer/autotuner.py
```diff
@@ -781,6 +765,13 @@ def choose_one(
                                         f"[Autotuner]: Failed when profiling {r} {tac}, shapes={shapes}. Error occurred: {e}"
                                     )
 
+                                    # Clear any pending async CUDA errors (e.g.
```
amirkl94, line 784: A bit of nitpicking but consider splitting this into a different PR

flashinfer/jit/fused_moe.py
johnnynunez: needed for activate PDL https://github.com/flashinfer-ai/flashinfer/issues/2708

eugr:
@askliar - according to [CUDA documentation](https://docs.nvidia.com/cuda/cuda-programming-guide/05-appendices/compute-capabilities.html), sm120 also has 99KB of shared memory, which makes sense, because sm120/sm121 are pretty much identical. Not sure where 228KB for sm120 comes from?

Also, I haven't looked at more detail, but I assume it will build correctly for Spark if FLASHINFER_CUDA_ARCH_LIST=12.1a (not 12.1f), right?

askliar:
@eugr just FYI, I have split this PR into two separate ones:
(Relu2, merged) https://github.com/flashinfer-ai/flashinfer/pull/2927
(Tactic errors, in review) https://github.com/flashinfer-ai/flashinfer/pull/2926

I will keep updating the latter with your comments!
Now to your questions:
1. Your comment on SM120 is correct, my understanding was wrong! Thank you and I will update FlashInfer to reflect that.
2. 12.1a/f are both okay, yes! 

Thank you for your contributions!

eugr:
Thanks!

PR #2903, opened 2026-03-27
fix: avoid re-downloading BMM export headers when flashinfer-cubin is installed

yzh119:
## Summary
- `download_trtllm_headers()` stored BMM export headers at a custom path (`CUBIN_DIR/flashinfer/trtllm/batched_gemm/trtllmGen_bmm_export/`), but `download_artifacts()` (used when building the flashinfer-cubin wheel) stores them at the canonical artifact hash path. The separate `get_file()` function only checked the destination path, so it never found the pre-packaged files and always re-downloaded ~17 header files from the network on every startup (×2 for each TP worker).
- Remove `download_trtllm_headers()`, `get_file()`, and rename `get_cubin()` → `get_artifact()`. Two unified primitives remain:
  - **`get_artifact(name, sha256)`** — fetches any artifact (cubins, headers, checksums, metainfo) into the canonical path, reusing flashinfer-cubin files when available
  - **`ensure_symlink(link, target)`** — maps C++ include paths to the artifact directory
- Callers (fused_moe, moe_utils) use these two primitives directly. A `get_cubin = get_artifact` alias is kept for backward compatibility.

## Related issue
- vllm-project/vllm#38110

## Test plan
- [x] 7 unit tests covering `get_artifact()` local cache (cubins, checksums.txt, metainfo headers, sha256 mismatch) and the BMM header symlink pattern (no download, stale dir replacement, idempotent)
- [ ] Verify on a gpt-oss-20b-sm100-fi-mxfp4-mxfp8-trtllm run with flashinfer-cubin installed that "Fetching file from" log messages no longer appear


aleozlx:
question: why did this PR touch the cutlass submodule?

aleozlx (commented):
lgtm except for the one question above

flashinfer/jit/cubin_loader.py
```diff
@@ -198,121 +197,51 @@ def load_cubin(cubin_path: str, sha256: str) -> bytes:
...
-            cached_hash = f.read().strip()
-        if cached_hash != artifact_path:
-            raise RuntimeError(
-                f"Detected inconsistent cached artifacts. "
-                f"(Cached trtllm headers were downloaded for artifact "
-                f"'{cached_hash}', but current code expects "
-                f"'{artifact_path}'). "
-                f"Please clear the cache to confirm and allow the new headers to be downloaded: "
-                f"rm -rf {header_dest_dir}."
-            )
```
jimmyzho, line 305: In the new implementation, is there a way we can ensure that the downloaded headers correspond to the correct hash? E.g if a user git checkouts to an older branch, then the artifactory + headers might be different corresponding to a different hash.

@aleozlx I remember discussing this before with you

aleozlx:
> s different hash i

which different hash did you mean? about cutlass or bmm artifacts?

jimmyzho:
> > s different hash i
> 
> which different hash did you mean? about cutlass or bmm artifacts?

Different bmm artifacts

aleozlx:
oh i think you mean the comment you left earlier.. i have addressed locally. let me push it

aleozlx:
i'm making a pass at other chatbot comments to double check...

flashinfer/jit/cubin_loader.py
aleozlx: resolved at the latest

aleozlx:
standby as i test the full workflow locally to double check everything

aleozlx:
cmds 

1  pip install --upgrade setuptools
    2  pip install --no-build-isolation -e .
    3  pushd flashinfer-cubin/
    4  pip install --no-build-isolation -e . -v
    5  export FLASHINFER_NO_DOWNLOAD=1
    6  export FLASHINFER_LOGGING_LEVEL=debug
    7  popd
    8  pytest tests/moe/test_trtllm_gen_fused_moe.py -v --log-cli-level=DEBUG

aleozlx:
local tests are looking good, no download observed by the aforementioned approach

aleozlx:
disclaimer on the last commit: it says debug but the debug var introduced is harmless and can help future debugging this area. hence leaving it in

aleozlx:
i'll send out a .post1 release that cherry picks it to resolve the filed issue, as planned, once this gets merged

aleozlx:
irrelevant err on gb200/300 (likely infra error)
```
FileExistsError: [Errno 17] File exists: '/workspace/junit'
❌ FAILED: tests/moe/test_trtllm_cutlass_fused_moe.py
==========================================
[93/118] Processing: tests/moe/test_trtllm_gen_fused_moe.py
```

aleozlx:
AOT builds has timed out

Issue #2023, opened 2025-11-01
[Bug] Autotuning + trtllm_fp4_block_scale_routed_moe Issue


varun-sundar-rabindranath:
Hey guys, I am facing an issue with `trtllm_fp4_block_scale_routed_moe` with autotuner. 

### Issue:
 Calling `trtllm_fp4_block_scale_routed_moe` with autotune results in an error, 
```
(EngineCore_DP0 pid=2361841) ERROR 11-01 12:49:12 [core.py:800]   File "/home/varun-sundar-rabindranath/code/vllm/vllm/model_executor/layers/fused_moe/trtllm_moe.py", line 140, in apply
(EngineCore_DP0 pid=2361841) ERROR 11-01 12:49:12 [core.py:800]     trtllm_fp4_block_scale_routed_moe(**kwargs)
(EngineCore_DP0 pid=2361841) ERROR 11-01 12:49:12 [core.py:800]   File "/home/varun-sundar-rabindranath/code/vllm/vllm-test/lib/python3.12/site-packages/flashinfer/fused_moe/core.py", line 2061, in trtllm_fp4_block_scale_routed_moe
(EngineCore_DP0 pid=2361841) ERROR 11-01 12:49:12 [core.py:800]     return get_trtllm_moe_sm100_module().trtllm_fp4_block_scale_moe(
(EngineCore_DP0 pid=2361841) ERROR 11-01 12:49:12 [core.py:800]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP0 pid=2361841) ERROR 11-01 12:49:12 [core.py:800]   File "/home/varun-sundar-rabindranath/code/vllm/vllm-test/lib/python3.12/site-packages/flashinfer/fused_moe/core.py", line 1540, in trtllm_fp4_block_scale_moe_op
(EngineCore_DP0 pid=2361841) ERROR 11-01 12:49:12 [core.py:800]     _, tactic = tuner.choose_one(
(EngineCore_DP0 pid=2361841) ERROR 11-01 12:49:12 [core.py:800]                 ^^^^^^^^^^^^^^^^^
(EngineCore_DP0 pid=2361841) ERROR 11-01 12:49:12 [core.py:800]   File "/home/varun-sundar-rabindranath/code/vllm/vllm-test/lib/python3.12/site-packages/flashinfer/autotuner.py", line 477, in choose_one
[...]

PR #2916, opened 2026-03-30
fix: Fix autotuner crash on meta-device tensor in trtllm_fp4_block_scale_routed_moe

bkryu:
#### Summary
- Fixes `RuntimeError: Cannot pack tensors on meta when trtllm_fp4_block_scale_routed_moe` is called with autotuning enabled
- Ensures the autotuner profiles the correct kernel code path (no-routing) when routing is pre-computed

#### Root Cause
Wh#en `trtllm_fp4_block_scale_routed_moe` is called, `routing_logits` is `None` because routing has already been done (pre-computed `topk_ids` are provided instead). To give the autotuner a tensor with the right shape/dtype for profile generation, a placeholder was created with `device="meta"`:

```
torch.empty(num_tokens, num_experts, dtype=routing_dtype, device="meta")
```
This worked without autotuning because `choose_one` returns early (only inspects .size() for the cache key, never passes the tensor to a kernel).

With autotuning enabled, `choose_one` enters the profiling loop, which calls `_create_tensor_like` on the placeholder. That method copies `origin_tensor.device`, so the derived profiling tensor is also on `"meta"`. When the profiling path calls `MoERunner.forward`, this meta tensor is passed to the C++ kernel via TVM FFI, which attempts DLPack conversion and fails: the meta device has no real memory.

#### Fix
Three changes in `flashinfer/fused_moe/core.py:`

- Replace `device="meta"` with `device=hidden_states.device` — the placeholder is now a real CUDA tensor so the autotuner can safely derive profiling tensors from it.
- Pass `skip_routing=(routing_logits is None)` through `kwargs` to `choose_one`, signaling that routing was pre-computed.
- In `MoERunner.forward`, set `routing_logits = None` when `skip_routing=True` — this ensures the C++ kernel takes the same no-routing code path during profiling as it does in production. Without this, the autotuner would profile with routing computation enabled (random `routing_logits` data), potentially selecting a suboptimal tactic for the actual inference path where routing is skipped.

#### Unit test changes
Added test_fp4_routed_moe_autotune_no_crash regression test in `tests/autotuner/test_trtllm_fused_moe_autotuner_integration.py` The test calls `trtllm_fp4_block_scale_routed_moe` inside `autotune(True)` with `num_tokens=1` and `num_tokens=16`, verifying no crash occurs.

Main branch fails the newly added tests before the changes in `flashinfer/fused_moe/core.py:`
```
$ pytest tests/autotuner/test_trtllm_fused_moe_autotuner_integration.py
...
tests/autotuner/test_trtllm_fused_moe_autotuner_integration.py ....FF...   
...
FAILED tests/autotuner/test_trtllm_fused_moe_autotuner_integration.py::test_fp4_routed_moe_autotune_no_crash[4-16-1] - RuntimeError: Cannot pack tensors on meta
FAILED tests/autotuner/test_trtllm_fused_moe_autotuner_integration.py::test_fp4_routed_moe_autotune_no_crash[4-16-16] - RuntimeError: Cannot pack tensors on meta
====================================================================================== 2 failed, 7 passed in 4.08s ======================================================================================
```

After fix:
```
$ pytest tests/autotuner/test_trtllm_fused_moe_autotuner_integration.py
...
tests/autotuner/test_trtllm_fused_moe_autotuner_integration.py .........                                                                                                                           [100%]

=========================================================================================== 9 passed in 6.47s ============================================================================================
```

## 🔍 Related Issues

#2023

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * Added skip routing support for Mixture of Experts operations, enabling optimization when routing computation can be bypassed.

* **Tests**
  * Added regression test for FP4 routed Mixture of Experts autotuning to ensure stability across token configurations.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,48 @@
+# Configuration for actions/labeler
+# Docs: https://github.com/actions/labeler
+
+"op: attention":
+  - changed-files:
+      - any-glob-to-any-file:
+          - "flashinfer/attention.py"
+          - "flashinfer/decode.py"
+          - "flashinfer/prefill.py"
+          - "flashinfer/mla/**"
+          - "flashinfer/xqa.py"
+          - "flashinfer/jit/attention/**"
+          - "include/flashinfer/attention/**"
+          - "csrc/fmha_v2/**"
+          - "csrc/xqa/**"
+          - "csrc/cutlass_mla*"
+          - "tests/attention/**"
+
+"op: moe":
+  - changed-files:
+      - any-glob-to-any-file:
+          - "flashinfer/fused_moe/**"
+          - "flashinfer/jit/fused_moe.py"
+          - "flashinfer/jit/moe_utils.py"
+          - "csrc/fused_moe/**"
+          - "csrc/trtllm_fused_moe*"
+          - "csrc/moe_utils*"
+          - "include/flashinfer/trtllm/fused_moe/**"
+          - "tests/moe/**"
+
+"op: comm":
+  - changed-files:
+      - any-glob-to-any-file:
+          - "flashinfer/comm/**"
+          - "flashinfer/jit/comm.py"
+          - "include/flashinfer/comm/**"
+          - "tests/comm/**"
+
+"op: gemm":
+  - changed-files:
+      - any-glob-to-any-file:
+          - "flashinfer/gemm/**"
+          - "flashinfer/jit/gemm/**"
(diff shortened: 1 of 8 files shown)


aleozlx (approved):

nvpohanh:
cc @trevor-m who is working on integrating routed moe into SGL

Issue #2920, opened 2026-03-31
Regression (0.6.6 → 0.6.7): enable_pdl=True on Ampere emits Hopper-only PTX (griddepcontrol), ptxas fails


yihwang-nv:
**Summary**

When using flashinfer-python 0.6.7, several norm/activation entry points that accept enable_pdl appear to JIT-compile through a CUTLASS DSL path that emits PTX using griddepcontrol / .wait / .launch_dependents, which require .target sm_90 or higher. If the NVVM/ptxas target is the actual GPU architecture (e.g. sm_86 / A10, sm_80 / A100), ptxas aborts and the failure surfaces as a CUTLASS DSL DSLRuntimeError (internal compiler ICE style message).

With 0.6.6, the same call pattern with enable_pdl=True did not hit this failure in our environment, so this looks like a regression in 0.6.7 (or an unintended breaking change in JIT/codegen behavior), not merely “user error” unless PDL is explicitly documented as Hopper-only and callers are expected to never pass enable_pdl=True below sm_90.

**Expected behavior**

Either:

1. enable_pdl=True on GPUs with CC < 9.0 should not generate sm_90-only instructions (automatically fall back to a non-PDL kernel), or
2. The API should fail fast with a clear error that PDL requires Hopper+, before emitting invalid PTX for the chosen arch.

**Actual behavior**

JIT compilation fails at ptxas with errors such as:

```
flashinfer: 0.6.7
GPU: NVIDIA A100 80GB PCIe sm_80
Traceback (most recent call last):
  File "/home/yihwang/.local/lib/python3.12/site-packages/nvidia_cutlass_dsl/python_packages/cutlass/base_dsl/dsl.py", line 1044, in compile_and_jit
    kernel = self.compiler_provider.compile_and_jit(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/yihwang/.local/lib/python3.12/site-packages/nvidia_cutlass_dsl/python_packages/cutlass/base_dsl/compiler.py", line 186, in compile_and_jit
    self.compile(
  File "/home/yihwang/.local/lib/python3.12/site-packages/nvidia_cutlass_dsl/python_packages/cutlass/base_dsl/compiler.py", line 161, in compile
    raise e
  File "/home/yihwang/.local/lib/python3.12/site-packages/nvidia_cutlass_dsl/python_packages/cutlass/base_dsl/compiler.py", line 148, in compile
[...]

PR #2928, opened 2026-03-31
fix: clamp enable_pdl=True to False on SM < 90 to prevent PDL PTX on Ampere

bkryu:
## 📌 Description

In 0.6.7, the norm kernels switched from the CUDA JIT path to the CuTe DSL path by
default. The CuTe DSL kernels emit `griddepcontrol.wait` / `griddepcontrol.launch_dependents` PTX when `enable_pdl=True`, which requires `.target sm_90` or higher.

Fix: Add ensures a clamp such that `enable_pdl=False` on unsupported hardware, while preserving the existing behavior for None (auto-detect) and False (opt-out).

Affected APIs
- `rmsnorm()`
- `rmsnorm_quant()`
- `fused_add_rmsnorm()`
- `fused_add_rmsnorm_quant()`
- `gemma_rmsnorm()`
- `gemma_fused_add_rmsnorm()`

PDL Behavior summary

| enable_pdl arg | HW supports PDL (SM>=90) | Before (0.6.7) | After (fix)
-- | -- | -- | --
None | Yes (Hopper+) | PDL enabled | PDL enabled
None | No (Ampere) | PDL disabled | PDL disabled
True | Yes (Hopper+) | PDL enabled | PDL enabled
**True** | **No (Ampere)** | **ptxas crash** | **PDL disabled**
False | Yes (Hopper+) | PDL disabled | PDL disabled
False | No (Ampere) | PDL disabled | PDL disabled

## 🔍 Related Issues

#2920

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Optimizations**
  * Improved device capability detection for normalization operations, enabling better automatic selection of optimal execution paths based on hardware compatibility.

### DIFF
--- a/flashinfer/norm/__init__.py
+++ b/flashinfer/norm/__init__.py
@@ -134,7 +134,7 @@ def _rmsnorm_impl(
     eps: float,
     enable_pdl: Optional[bool],
 ) -> None:
-    if enable_pdl is None:
+    if enable_pdl is None or enable_pdl:
         enable_pdl = device_support_pdl(input.device)
     if _USE_CUDA_NORM:
         get_norm_module().rmsnorm(out, input, weight, eps, enable_pdl)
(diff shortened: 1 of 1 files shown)


bkryu:
> I suppose this pattern was also used in other kernels? Do we need to populate to the entire codebase (other than only norm).

This is a good point. This PR directly addresses the regression reported in #2920, but I supposed that it is true. Some kernels like CuTe DSL MoE will be fine because it only supports SM100/103 anyways, others may not be. 

I can create a followup PR after this one patching all the holes.

PR #2931, opened 2026-03-31
feat: add get flashinfer-trace interface .fi_trace

yyihuang:
## 📌 Description

Adds a **trace layer** to FlashInfer so every public kernel can be described as a portable benchmark / replay definition without tying the description to any particular launcher.

- **`flashinfer/trace/template.py`** — new `TraceTemplate` schema with named `axes` (`Var`/`Const`), typed `inputs`/`outputs` (`Tensor`/`Scalar`), optional `reference` implementation, and tag/constraint metadata.
- **`flashinfer/trace/templates/*.py`** — one module per operator family (attention, cascade, GDN, GEMM, MoE, norm, activation, sampling). Each file declares the schema and, where feasible, an executable reference.
- **`@flashinfer_api(trace=...)`** (extends the existing decorator in `flashinfer/api_logging.py`) — attaches `.fi_trace()` to the decorated function/method and, when `FLASHINFER_TRACE_DUMP=1`, writes a per-shape JSON definition to `FLASHINFER_TRACE_DUMP_DIR` **before** the kernel runs (crash-safe).
- **`fi_trace()` helpers** — public entry points for programmatic trace generation from any `@flashinfer_api`-decorated API or a bound method.
- **`tests/trace/`** — template-consistency tests (signature ↔ axes/inputs), end-to-end reference checks, and an `example.py` that drives a realistic workload and dumps 45 `tests/trace/fi_trace_out/*.json` definitions across LLaMA-3.1, DeepSeek-V3, Gemma, Qwen3-Next, etc.

### Why
- Lets external tooling (e.g. `flashinfer-bench`) consume a single self-describing JSON per op instead of reverse-engineering Python call sites.
- Gives each kernel a single authoritative description of its shape constraints, keeping plan→run wrappers, benchmarks, and regression tests aligned.
- Zero overhead at `FLASHINFER_LOGLEVEL=0` / `FLASHINFER_TRACE_DUMP` unset (decorator is a no-op in that path).

### Covered APIs
Attention (paged/ragged prefill, paged decode, MLA), sampling (top-k/top-p/top-k-top-p), GEMM (bf16, fp8, mxfp8, fp4), fused MoE (fp8/fp4 block-scale × 6 routing methods), norm (rmsnorm, fused-add, quant variants, Gemma variants, layernorm), activation (silu/gelu/gelu-tanh + mul), cascade merge (state/state-in-place/states), and GDN (decode, MTP, chunk prefill).

## 🔍 Related Issues
- External consumer: https://github.com/flashinfer-ai/flashinfer-bench

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit`.

## 🧪 Tests

- [x] Tests have been added or updated as needed (`tests/trace/test_fi_trace.py`, `tests/trace/test_fi_trace_template_consistency.py`).
- [x] All tests are passing locally (`pytest tests/trace/ -v` → 139 passed).

## Reviewer Notes

- The decorator ordering is load-bearing: `@flashinfer_api(trace=...)` **must** be innermost so trace dump runs even when surrounding `@backend_requirement` raises for unsupported capability. For `mm_fp4` / `mm_mxfp8` on SM<100 the outer `@backend_requirement` raises before the dump, which is why their JSONs are only regenerated on Blackwell. See `tests/trace/example.py` for the realistic workload.
- Subsequent fix-up commit drops three redundant `@flashinfer_api` decorators that caused double-logging at `FLASHINFER_LOGLEVEL=3+` (subclass `__init__` overrides and `trtllm_low_latency_gemm` internal helper).
- Known remaining CodeRabbit items (follow-up work, not blocking this PR): some MoE reference helpers still hardcode `H`/`I`/`top_k`/`n_group`; `gdn_prefill_trace` lacks the head-ratio constraints that `gdn_decode` / `gdn_mtp` already have; the E2E test synthesizer uses `0` for `int32` inputs, which makes some synthesized definitions nonsensical.


flashinfer/gemm/gemm_base.py
```diff
@@ -2084,6 +2091,7 @@ def build_cudnn_gemm_fp4_graph_override_shape(
     return graph
 
 
+@flashinfer_api
```
bkryu: Not sure if this is a FlashInfer API that we expect users to run. Do we need the label here?

flashinfer/gemm/gemm_base.py
```diff
@@ -2319,6 +2327,7 @@ def build_cudnn_gemm_mxfp8_graph_override_shape(
...
```
bkryu: Ditto

flashinfer/gemm/gemm_base.py
```diff
@@ -2565,6 +2574,7 @@ def build_cudnn_gemm_with_per_tensor_q_graph_override_shape(
...
```
bkryu: Ditto

flashinfer/gemm/gemm_base.py
```diff
@@ -2893,6 +2903,7 @@ def build_cudnn_gemm_bf16_graph_override_shape(
...
```
bkryu: Ditto

bkryu (commented):
Overall looks good to me. Left a comment about some (I believe) non-APIs that we may not want to decorate.

Also, not sure if you intended the PR to be exhaustive, but I think you missed:
* cuDNN and TRTLLM attention variants
* CUTLASS Fused MoE
* Quantization APIs
* RoPe APIs

Ubospica (approved):
LGTM

nv-yunzheq (approved):

yzh119 (approved):
LGTM, I would encouraging update the AI review template to let them check this whenever there are new APIs.

yongwww (approved):
LGTM. Please fix the pre-commit

bkryu (approved):

aleozlx (approved):

PR #2942, opened 2026-04-01
[Perf] Refactor MoE autotuning to set valid topk ids in routed MoE tuning

wzhao18:
## 📌 Description
This PR fixes various issues in Trtllm MoE autotuning:
- Currently trtllm MoE uses [torch.empty-initialized topk ids](https://github.com/flashinfer-ai/flashinfer/blob/23b3279868432f1e93d3ce0c6217eb10131d18de/flashinfer/fused_moe/core.py#L837) for routed MoE autotuning, this results in misrepresented MoE expert distribution, and leads to suboptimal tuned kernel performance. This PR initializes topk ids to be within valid ranges for expert indices.
-  Now the dynamic dimension for `hidden_states_scale` is always set to `0`, but for certain ops like fp8 moe, the num_tokens dimension is at `1`. Thus the cache will not correctly and other batch sizes will not use the tuned kernel. This PR fixes this by setting the dynamic dimension for hidden states scale differently across ops.

This PR also adds progress bar to the tuning process - this helps diagnose whether the caching works properly. For instance, previously running autotune for different batch sizes will each trigger a new tuning process, which suggests the cache does not work properly. The progress bar is only shown if there is at least one cache miss, so there won't be flooded logging.

## Benchmark results:
Using minimax m2.5 dimensions for benchmarking MoE autotuning.
```sh
SCRIPT="benchmarks/bench_trtllm_gen_fused_moe_autotuner.py"
COMMON="--num-experts 256 --hidden-size 3072 --intermediate-size 1536 --top-k 8 --tp 2"
BATCH_SIZES=(1 8 64 128 256 512 1024 2048)

for quant_mode in "Fp8-Block" "MxFP4xBf16"; do
    for routed in "" "--routed"; do
        echo "========================================"
        echo "quant_mode=${quant_mode} routed=${routed:-false}"
        echo "========================================"
        python "$SCRIPT" \
            --quant-mode "$quant_mode" \
            --num-tokens "${BATCH_SIZES[@]}" \
            $routed \
            $COMMON
    done
done
```

**FP8 Routed:**
main:
```
quant_mode=Fp8-Block  routing=routed  experts=256  hidden=3072  intermediate=768  top_k=8
    num_tokens    no_autotune (ms)    autotuned (ms)    speedup
  ------------  ------------------  ----------------  ---------
             1               0.026             0.028       0.91x
             8               0.078             0.078       1.00x
            64               0.251             0.251       1.00x
           128               0.287             0.287       1.00x
           256               0.327             0.327       1.00x
           512               0.483             0.482       1.00x
          1024               0.555             0.555       1.00x
          2048               0.667             0.668       1.00x
```

branch:
```
quant_mode=Fp8-Block  routing=routed  experts=256  hidden=3072  intermediate=768  top_k=8
    num_tokens    no_autotune (ms)    autotuned (ms)    speedup
  ------------  ------------------  ----------------  ---------
             1               0.025             0.025       1.02x
             8               0.081             0.081       1.00x
            64               0.251             0.251       1.00x
           128               0.288             0.288       1.00x
           256               0.330             0.310       1.06x
           512               0.490             0.334       1.47x
          1024               0.555             0.433       1.28x
          2048               0.667             0.592       1.13x
```
There is no speedup in main because the caching does not work (the benchmark script only runs autotune for one batch size, the rest of batch sizes fails to look up the cache due to incorrect dynamic dimension set for `hidden_states_scale`).

**FP4 Routed:**
main:
```
quant_mode=MxFP4xBf16  routing=routed  experts=256  hidden=3072  intermediate=768  top_k=8
    num_tokens    no_autotune (ms)    autotuned (ms)    speedup
  ------------  ------------------  ----------------  ---------
             1               0.044             0.043       1.03x
             8               0.166             0.179       0.93x
            64               0.592             0.623       0.95x
           128               0.675             0.731       0.92x
           256               0.955             0.759       1.26x
           512               1.619             0.785       2.06x
          1024               1.617             0.792       2.04x
          2048               1.774             1.774       1.00x
```

branch:
```
quant_mode=MxFP4xBf16  routing=routed  experts=256  hidden=3072  intermediate=768  top_k=8
    num_tokens    no_autotune (ms)    autotuned (ms)    speedup
  ------------  ------------------  ----------------  ---------
             1               0.044             0.042       1.05x
             8               0.184             0.182       1.01x
            64               0.595             0.592       1.01x
           128               0.690             0.667       1.04x
           256               0.955             0.683       1.40x
           512               1.619             0.727       2.23x
          1024               1.631             0.770       2.12x
          2048               1.775             1.128       1.57x
```
The speedup is better than main because of the properly initialized `topk_ids` during autotuning.

## Ablation
To ensure the new initialization of `topk_ids` is meaningful, I compared the benchmark results with toggling the two initializers with everything else the same:
```python
if moe_inputs.topk_ids is not None:
       spec["topk_ids"] = _init_packed_topk_ids.                         # new
       spec["topk_ids"] = lambda shapes, dtype, device: torch.empty(     # before
             shapes, dtype=dtype, device=device
       )
```

Results:
**FP8 Routed:**
torch.empty:
```
quant_mode=Fp8-Block  routing=routed  experts=256  hidden=3072  intermediate=768  top_k=8
    num_tokens    no_autotune (ms)    autotuned (ms)    speedup
  ------------  ------------------  ----------------  ---------
             1               0.025             0.030       0.83x
             8               0.085             0.085       0.99x
            64               0.255             0.267       0.96x
           128               0.285             0.295       0.97x
           256               0.331             0.340       0.97x
           512               0.491             0.450       1.09x
          1024               0.556             0.779       0.71x
          2048               0.668             0.697       0.96x
```

_init_packed_topk_ids:
```
quant_mode=Fp8-Block  routing=routed  experts=256  hidden=3072  intermediate=768  top_k=8
    num_tokens    no_autotune (ms)    autotuned (ms)    speedup
  ------------  ------------------  ----------------  ---------
             1               0.025             0.023       1.06x
             8               0.078             0.078       1.00x
            64               0.256             0.256       1.00x
           128               0.288             0.288       1.00x
           256               0.328             0.309       1.06x
           512               0.491             0.335       1.47x
          1024               0.557             0.438       1.27x
          2048               0.667             0.594       1.12x
```
We can see using torch.empty to initialize the topk_ids in routed MoE kernel results in worse performance than the untuned kernel due to the noisy data. The new initialization brings noticeably speedup at larger batch sizes.

## E2E Evaluation:
Compared no tuning, tuning on main branch, and turning on PR branch on Minimax FP8 1K/1k on vllm.
isl1024_osl1024

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * Routed fused‑MoE runtime support for FP4 and FP8; supports routed top‑k inputs and reports consolidated no‑autotune vs autotuned medians and speedups across multiple token sizes.
  * Benchmark CLI: multi-value --num-tokens, --tp, and --routed with runtime validation; per-token-size benchmarking.

* **Tests**
  * Expanded autotuning smoke tests for FP4 and FP8 covering routed and non‑routed modes.

* **Chores**
  * Show progress during tuning (progress bar) and added build dependency for it.

### DIFF
--- a/flashinfer/gemm/gemm_base.py
+++ b/flashinfer/gemm/gemm_base.py
@@ -214,7 +214,7 @@ def _cutlass_mm_bf16_requirement(
     return True
 
 
-@supported_compute_capability([100, 103])
+@supported_compute_capability([80, 86, 89, 90, 100, 103])
 def _cudnn_mm_bf16_requirement(
     a: torch.Tensor,
     b: torch.Tensor,
@@ -224,15 +224,6 @@ def _cudnn_mm_bf16_requirement(
     pdl: bool = False,
     backend: Literal["cudnn", "cutlass", "tgv", "auto"] = "cudnn",
 ):
-    if bias is not None:
-        raise ValueError(
-            "You cannot use the cuDNN backend with a bias. Use the TGV backend instead."
-        )
-    if pdl:
-        raise ValueError(
-            "The cuDNN backend does not support PDL. Use the TGV backend instead."
-        )
-
     _validate_bf16_output_dtype(out_dtype)
     _check_cudnn_availability()
 
@@ -308,16 +299,18 @@ def _heuristic_func_mm_bf16(
 ):
     heuristic_backends = []
     if bias is not None or pdl:
-        # cuDNN and CUTLASS don't support bias/pdl, only TGV does
+        # CUTLASS doesn't support bias/pdl, only TGV and cuDNN do
         if "tgv" in suitable_backends:
             heuristic_backends.append("tgv")
-    else:
         if "cudnn" in suitable_backends:
             heuristic_backends.append("cudnn")
+    else:
         if "cutlass" in suitable_backends:
             heuristic_backends.append("cutlass")
         if "tgv" in suitable_backends:
             heuristic_backends.append("tgv")
+        if "cudnn" in suitable_backends:
+            heuristic_backends.append("cudnn")
     return heuristic_backends
 
 
@@ -1658,7 +1651,8 @@ class UIDs(Enum):
     BLOCK_DESCALE_B_UID = 4
     A_SCALE_UID = 5
     B_SCALE_UID = 6
-    O_UID = 7
+    BIAS_UID = 7
+    O_UID = 8
 
 
 def _check_cudnn_availability():
@@ -2113,7 +2107,7 @@ def execute_cudnn_gemm_fp4_graph_override_shape(
 
     c_shape, c_stride = _get_bf16_3d_shape_stride(c_final)
 
-    if real_a_stride[2] != 1 or real_b_stride[1] != 1:
+    if real_a_stride[-1] != 1 or real_b_stride[-2] != 1:
         raise ValueError(
             f"a and b must be k-major (contiguous along the K dimension), "
             f"got a stride={tuple(real_a_stride)}, b stride={tuple(real_b_stride)}"
@@ -2671,8 +2665,23 @@ def _cudnn_gemm_fp8_runner():
     return CudnnFp8GemmRunner()
 
 
+def _get_3d_shape_stride_from_vector(vector: torch.Tensor, dim: int = 0):
+    """Expand 1d vector to 3d tensor for cuDNN"""
+    if vector.dim() != 1:
+        raise ValueError(f"Expected 1D vector, got {vector.dim()}D tensor")
+    n = vector.shape[0]
+    p = vector.stride(0)
+    shape = [1, 1, 1]
+    stride = [1, 1, 1]
+    shape[dim] = n
+    stride[dim] = p
+    return (tuple(shape), tuple(stride))
+
+
 def _get_bf16_3d_shape_stride(tensor: torch.Tensor):
     """Expand 2d tensor to 3d tensor for cuDNN"""
+    if tensor.dim() != 2 and tensor.dim() != 3:
+        raise ValueError(f"Expected 2D or 3D tensor, got {tensor.dim()}D tensor")
     shape = list(tensor.shape)
     stride = list(tensor.stride())
 
@@ -2691,6 +2700,9 @@ def build_cudnn_gemm_bf16_graph(
     b_stride,
     o_type,
     device,
+    bias_is_not_none,
+    bias_shape,
+    bias_stride,
     policy=None,
 ):
     _check_cudnn_availability()
@@ -2711,11 +2723,29 @@ def build_cudnn_gemm_bf16_graph(
             B=b_cudnn_tensor,
             compute_data_type=cudnn.data_type.FLOAT,
         )
-        c_cudnn_tensor.set_name("c").set_output(True).set_data_type(o_type)
+        c_cudnn_tensor.set_data_type(cudnn.data_type.FLOAT)
+
+        if bias_is_not_none:
+            bias_cudnn_tensor = graph.tensor(
+                name="bias",
+                dim=bias_shape,
+                stride=bias_stride,
+                data_type=cudnn.data_type.BFLOAT16,
+            )
+            c_final_cudnn_tensor = graph.add(
+                name="bias_add",
+                a=c_cudnn_tensor,
+                b=bias_cudnn_tensor,
+            )
+            bias_cudnn_tensor.set_uid(UIDs.BIAS_UID.value)
+        else:
+            c_final_cudnn_tensor = c_cudnn_tensor
+
+        c_final_cudnn_tensor.set_name("c_final").set_output(True).set_data_type(o_type)
 
         a_cudnn_tensor.set_uid(UIDs.A_UID.value)
         b_cudnn_tensor.set_uid(UIDs.B_UID.value)
-        c_cudnn_tensor.set_uid(UIDs.O_UID.value)
+        c_final_cudnn_tensor.set_uid(UIDs.O_UID.value)
 
         graph.validate()
         graph.build_operation_graph()
@@ -2726,12 +2756,22 @@ def build_cudnn_gemm_bf16_graph(
         return graph
 
 
-def execute_cudnn_gemm_bf16_graph(graph, a, b, c_final, workspace, tactic: int = -1):
-    variant_pack = {
-        UIDs.A_UID.value: a,
-        UIDs.B_UID.value: b,
-        UIDs.O_UID.value: c_final,
-    }
+def execute_cudnn_gemm_bf16_graph(
+    graph, a, b, bias, c_final, workspace, tactic: int = -1
+):
+    if bias is not None:
+        variant_pack = {
+            UIDs.A_UID.value: a,
+            UIDs.B_UID.value: b,
+            UIDs.BIAS_UID.value: bias,
+            UIDs.O_UID.value: c_final,
+        }
+    else:
+        variant_pack = {
+            UIDs.A_UID.value: a,
+            UIDs.B_UID.value: b,
+            UIDs.O_UID.value: c_final,
+        }
 
     stream = torch.cuda.current_stream(a.device)
     cudnn_handle = _get_cudnn_handle(a.device, stream)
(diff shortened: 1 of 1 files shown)


flashinfer/fused_moe/core.py
```diff
@@ -893,28 +898,100 @@ def __init__(
...
+            input_idx = tuple(i for i, _, _ in sorted_inputs)
+
+            num_tokens = moe_inputs.hidden_states.shape[0]
+
+            def _dynamic_dim(name: str) -> int:
+                if name == "hidden_states_scale":
+                    # Layout varies by op: find which dim equals num_tokens.
+                    t = moe_inputs.hidden_states_scale
+                    return next(i for i, s in enumerate(t.shape) if s == num_tokens)
+                return MoEInputs._DYNAMIC_DIM[name]
```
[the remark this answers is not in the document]
wzhao18: You are right, let me fix this.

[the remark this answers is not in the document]
wzhao18: resolved now

benchmarks/bench_trtllm_gen_fused_moe_autotuner.py
```diff
@@ -80,135 +108,62 @@ def bench_trtllm_gen_fused_moe_autotuner_fp8(
...
+            [hidden_states_scale_scalar * w13_scale] * num_experts, device=device
+        )
+        output1_scales_gate_scalar = torch.ones(
+            num_experts, device=device, dtype=torch.float32
+        )
+        output2_scale_scalar = torch.tensor(
+            [hidden_states_scale_scalar * w2_scale] * num_experts, device=device
+        )
+    elif quant_mode == "Fp8-Block":
+        _, w13_scalar = fp8_quantize(w13)
```
qiching: i think here should be 
`w13, w13_scalar = fp8_quantize(w13)`
`w2, w2_scalar = fp8_quantize(w2)`

since `fp8_quantize` returns `(quantized_tensor, scale)`, but here only scale is retrieved, and the original bf16 weights recast. this will causes `gemm*_weights` and `gemm*_weights_scale` not the align, so will be inaccurate.

flashinfer/autotuner.py
```diff
@@ -751,6 +754,14 @@ def choose_one(
                         inputs=tensors,
                     )
                     if not is_cache_hit:
+                        if pbar is None:
+                            pbar = tqdm.tqdm(
+                                total=len(profiles),
+                                initial=_step,
+                                desc=f"[AutoTunner]: Tuning {custom_op}",
+                                unit="profile",
```
bkryu, line 762: I see a typo; should be `[AutoTuner]`

flashinfer/autotuner.py
```diff
@@ -7,6 +7,8 @@
 import os
 import tempfile
 import threading
+
+import tqdm
```
bkryu, line 11: It seems like we have a new top-level import for `tqdm`. 

While `tqdm` is likely to be installed in an environment running e2e workloads, I'm a bit ambivalent about whether adding it as a dependency.

flashinfer/autotuner.py
wzhao18: yes thanks!

benchmarks/bench_trtllm_gen_fused_moe_autotuner.py
qiching: @aleozlx can you review and correct me

aleozlx: idk the reason for this actually

cc @wzhao18

flashinfer/fused_moe/core.py
bkryu: There seems to be an assertion mismatch with new tuning config for DeepSeek FP8

What exists today (main): `MoERunner.forward()` has an assertion that assumes `hidden_states_scale.shape[0] == num_tokens`. On main this was never a problem because the old class-level `tuning_config_with_hidden_states_scales` hardcoded `dynamic_dim = 0` for every tensor, so the autotuner always synthesized `hidden_states_scale` with `shape[0] == num_tokens` — even for DeepSeek FP8 MoE, where the real layout is `[hidden_size//128, num_tokens]`. The cache key was silently wrong (which is what this PR fixes), but the assertion never fired.

What this PR changes: `_make_tuning_config()` now correctly sets `dynamic_dim = 1` for `hidden_states_scale` when `fp8_quantization_type == DeepSeekFp8`. This means the autotuner will synthesize tensors with shape `[hidden_size//128, num_tokens]` — matching the real DeepSeek FP8 layout. This is the right fix for the cache key.

What breaks: During autotuning, those synthesized tensors flow through `runner.forward()` for profiling. The assertion at line 1070 fires because `shape[0]` is now `hidden_size//128`, not `num_tokens`. The AssertionError is caught by the autotuner's except Exception handler (`autotuner.py` line 784), so it doesn't crash — but profiling attempts can be marked failed, and tuning silently falls back to the default. The net effect is that DeepSeek FP8 autotuning produces no benefit.

Note: `forward()` already handles this correctly after the assertion — at later in `forward()`, the DeepSeek path creates a fresh `current_hidden_states_scale = torch.full((hidden_size//128, num_tokens), ...)` and ignores the input. So the assertion is the only thing blocking correct execution.

Suggested fix:
```
# Replace lines 1070-1073:
if hidden_states_scale is not None:
    assert hidden_states_scale.dim() == 2, (
        "hidden_states_scale must be a 2D tensor"
    )
    if self.fp8_quantization_type == Fp8QuantizationType.DeepSeekFp8:
        assert hidden_states_scale.shape[1] == num_tokens, (
            f"DeepSeekFp8 hidden_states_scale shape {tuple(hidden_states_scale.shape)} "
            f"expects num_tokens={num_tokens} at dim 1"
        )
    else:
        assert hidden_states_scale.shape[0] == num_tokens, (
            f"hidden_states_scale shape {tuple(hidden_states_scale.shape)} "
            f"expects num_tokens={num_tokens} at dim 0"
        )
```
That covers the what/why/impact/fix in a way that's constructive and gives the PR author a clear path to resolution.

qiching: yesterday i reviewed and also found some concerned here, but i do not think this analysis is totally right. before reaching line 1070, forward() will have`trtllm_gen_dtype_has_scale(self.dtype_act)` which returns `false` for e4m3(ds fp8), setting `hidden_states_scale = None`. the assertion then short circuits on is `None` and never checks `shape[0]`. So there is no assertionError, and autotuning works as intended.

but the assertion is still inconsistent with the layout modeled by `_make_tuning_config`, so updating it for clarity is make sense, but it is not a functional bug.

qiching: ref https://github.com/flashinfer-ai/flashinfer/blob/34b6550e6b0b032d52d28197e875c6ed2c31d70f/flashinfer/tllm_enums.py#L76-L85

bkryu: Ah okay this sounds correct. Thanks for the clarification!

wzhao18: Thanks for looking at this! I think the proposed fix makes it consistent. I will apply the fix.

flashinfer/autotuner.py
wzhao18: tqdm seems already a dependency of flashinfer - https://github.com/flashinfer-ai/flashinfer/blob/fe0539318dcc31c76a33a7ed2ab0ee3c94fe6bad/requirements.txt#L14

flashinfer/autotuner.py
wzhao18: fixed.

benchmarks/bench_trtllm_gen_fused_moe_autotuner.py
wzhao18: Thanks for flagging this! This is now fixed.

wzhao18:
@aleozlx @bkryu The CI seems passing but the full pipeline fails. Is that expected? or we should try a re-run?

jimmyzho:
> The CI seems passing but the full pipeline fails. Is that expected? or we should try a re-run?

Seems to be just due to flaky CI, I will rerun and merge once it looks good

wzhao18:
@aleozlx both CI and internal pipeline seem passing. Merge?

PR #2944, opened 2026-04-01
feat: Add CuTe DSL grouped-gemm + combine fusion support

nvcastet:
## Summary

Enables https://github.com/sgl-project/sglang/pull/21877

- Add fused grouped-GEMM + combine operation to the CuTe DSL Blackwell masked GEMM kernel, enabling the GEMM epilogue to perform weighted scatter-reduce (combine) directly into multi-rank output buffers using `cp.reduce.async.bulk` PTX instructions (`bf16` and `f32` variants)
- Extend `MaskedScheduler` with `is_swap_ab` support so the scheduler correctly computes tile coordinates when A/B inputs are swapped (needed for combine fusion where output is M-major)
- Add `barrier_flag_local`/`barrier_flag_multicast` parameters for cross-rank synchronization via spin-lock barriers
- Add multi-GPU test covering the fused grouped-gemm + combine path

## Key changes
- **`flashinfer/gemm/kernels/grouped_gemm_masked_blackwell.py`**: New `cp_reduce_bf16_add`/`cp_reduce_f32_add` PTX wrappers, custom `make_fused_smem_layout_epi` for the combine path, `is_swap_ab` logic in scheduler, and new parameters (`topk_weights`, `idx_src_info`, `rank_src_info`, `out_ptrs`, barrier flags) threaded through the full stack
- **`tests/comm/test_multi_gpu_cute_dsl_blockscaled_gemm_fusion.py`**: New multi-GPU test that validates the fused gemm+combine against a reference implementation using `mpirun`

## Test plan
- [ ] `mpirun -np 4 pytest tests/comm/test_multi_gpu_cute_dsl_blockscaled_gemm_fusion.py` on SM100+ hardware
- [ ] Verify existing grouped-gemm tests still pass (non-fusion path unchanged)

## Summary by CodeRabbit

* **New Features**
  * Block-scaled GEMM combine-fusion: fused weight prefetch, per-tile async reduction stores, and optional AB swap.
  * Host API extended with optional top-k weights, routing info, output-pointer inputs, rank/count, and barrier-flag controls for distributed fusion runs.

* **Tests**
  * Added multi-GPU integration test validating the combine-fusion path, routing, and distributed synchronization.

* **Bug Fixes**
  * Persistent scheduler sizing improved by capping active clusters to available tiles; stage computation updated for fusion prefetch buffers.

### DIFF
--- a/flashinfer/gemm/kernels/grouped_gemm_masked_blackwell.py
+++ b/flashinfer/gemm/kernels/grouped_gemm_masked_blackwell.py
@@ -455,6 +455,14 @@ def cp_reduce_bf16_add(dst_addr, src_addr, size, *, loc=None, ip=None):
     """Async bulk copy-reduce (add) from shared memory to global memory.
 
     PTX instruction: cp.reduce.async.bulk.global.shared::cta.bulk_group.add.bf16.noftz [dstMem], [srcMem], size;
+
+    Performs an asynchronous bulk reduction operation that atomically adds bf16 values
+    from shared memory (CTA scope) to global memory. The operation is part of a bulk group.
+
+    Args:
+        dst_addr: Destination global memory address (pointer)
+        src_addr: Source shared memory address (pointer)
+        size: Size in bytes of the data to reduce (must be a multiple of 4)
     """
     llvm.inline_asm(
         None,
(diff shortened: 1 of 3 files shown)


gyhintel:
Hi, May I ask a question? Why is "is_swap_ab" needed? The comment says it's "needed for combine fusion where output is M-major."  The mma result in tensor memory is M-major. But I think it can be load to RF and then store it  to shared memory via "stmatrix.xxx.**trans**". Then the result in shared memory is N-major and it didn't need the "swap_ab". Thanks!

nvcastet:
> Hi, May I ask a question? Why is "is_swap_ab" needed? The comment says it's "needed for combine fusion where output is M-major." The mma result in tensor memory is M-major. But I think it can be load to RF and then store it to shared memory via "stmatrix.xxx.**trans**". Then the result in shared memory is N-major and it didn't need the "swap_ab". Thanks!

Good question, with Blackwell subtiling pipelining the mem copy out of TMEM to REG/smem/gmem  the subtile shape is <128,32> for mma tile <128,128>. Which means we can only write 32 elts per token per subtile which is not efficient for mem writes, with swap_ab, we can write 128 contiguous elts per token per subtile (256B for bf16).

aleozlx (approved):
lgtm

PR #2955, opened 2026-04-02
unifying all reduce memory allocation for single-node and multi-node nvlink

Amir-19:
## 📌 Description
The goal of this PR is to unify memory allocation for all reduce to use torch symmetric memory instead of custom allocators in in flashinfer.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

```
FLASHINFER_DISABLE_VERSION_CHECK=1 PYTHONPATH=.  mpirun --allow-run-as-root  -np <num_gpus> pytest tests/comm/test_allreduce_unified_api.py -vv -s
<num_gpus>:8 pass
<num_gpus>:4 pass
<num_gpus>:2 pass
```
```
FLASHINFER_DISABLE_VERSION_CHECK=1 PYTHONPATH=. pytest tests/comm/test_trtllm_allreduce_fusion.py -vv -s

pass
```
```
FLASHINFER_DISABLE_VERSION_CHECK=1 PYTHONPATH=.  mpirun --allow-run-as-root -np 4 pytest tests/comm/test_trtllm_mnnvl_allreduce.py -vv -s

<num_gpus>:4 pass

```
```
FLASHINFER_DISABLE_VERSION_CHECK=1 PYTHONPATH=. pytest tests/comm/test_trtllm_mnnvl_allreduce_custom_comm.py -vv -s

pass
```
## Summary by CodeRabbit

* **Improvements**
  * Enabled symmetric CUDA memory allocation for collective operations, improving stability and performance across PyTorch versions.
  * Unified workspace allocation and lifecycle for all-reduce/fusion paths, simplifying resource management and reducing teardown issues.

* **Bug Fixes**
  * Prevented a reset in distributed group tracking on older PyTorch releases, reducing failures during group teardown.

* **Tests**
  * Enhanced test logging and updated helpers to exercise the new workspace handling.

### DIFF
--- a/flashinfer/comm/torch_symmetric_memory.py
+++ b/flashinfer/comm/torch_symmetric_memory.py
@@ -1,0 +1,15 @@
+import torch
+import torch.distributed._symmetric_memory as symm_mem
+
+
+def _alloc_symm_buffer_bytes(
+    size_bytes: int, tp_size: int, dtype: torch.dtype, device: torch.device, group_name: str
+) -> tuple[list[int], torch.Tensor]:
+    numel = size_bytes // torch.tensor([], dtype=dtype).element_size()
+    tensor = symm_mem.empty(numel, dtype=dtype, device=device)
+    handle = symm_mem.rendezvous(tensor, group=group_name)
+    ptrs: list[int] = []
+    for peer in range(tp_size):
+        buf = handle.get_buffer(peer, (numel,), dtype, storage_offset=0)
+        ptrs.append(buf.data_ptr())
+    return ptrs, tensor, handle
--- a/flashinfer/comm/torch_symmetric_memory.py
+++ b/flashinfer/comm/torch_symmetric_memory.py
@@ -1,7 +1,95 @@
+import functools
 from typing import Any
 
 import torch
 import torch.distributed._symmetric_memory as symm_mem
+import torch.distributed.distributed_c10d as c10d
+
+_compat_patched = False
+
+
+def _patch_group_count_reset() -> None:
+    """Prevent group_count from resetting to 0 on WORLD destruction (2.10 only).
+
+    On PyTorch 2.10, ``destroy_process_group(WORLD)`` resets
+    ``_world.group_count`` to 0.  When the WORLD group is later recreated via
+    ``init_process_group``, it receives the same name (e.g. ``"0"``) as the
+    previous incarnation.  The C++ ``group_info_map`` still holds the old
+    entry whose store is now a zombie (the underlying TCPStore server was torn
+    down with the old process group).  Because ``set_group_info`` enforces
+    single-init per name, we cannot update the stale entry.
+
+    By preserving the counter, each recreated WORLD group gets a unique name
+    (``"0"``, ``"1"``, ``"2"``, …).  ``set_group_info`` always sees a fresh
+    key, the rendezvous uses the live store, and the old zombie entries are
+    simply never looked up again.
+    """
+    global _compat_patched
+    if _compat_patched:
+        return
+    _compat_patched = True
+
+    import torch.distributed as dist
+
+    _original_destroy = dist.destroy_process_group
+
+    @functools.wraps(_original_destroy)
+    def _patched_destroy(group=None):
+        saved_count = c10d._world.group_count
+        _original_destroy(group)
+        # WORLD destruction resets group_count to 0 – restore it so the next
+        # init_process_group picks a name that is fresh in the C++ map.
+        if group is None:
+            c10d._world.group_count = saved_count
+
+    dist.destroy_process_group = _patched_destroy
+
+
+def _enable_symm_mem_compat(group_name: str) -> None:
+    """Pre-initialize symmetric memory for a process group (PyTorch 2.10 compat).
+
+    PyTorch 2.10's ``enable_symm_mem_for_group`` wraps the group store in an
+    extra ``PrefixStore("symmetric_memory-{ranks}", ...)``.  PyTorch 2.11
+    removed this indirection and resolves the group store directly in C++ via
+    ``group->getStore()``.  On 2.10 the extra layer can cause the store-based
+    handle exchange inside ``make_peer_alloc_info`` to hang on certain
+    topologies (Blackwell MNNVL fabric handles).
+
+    This helper mimics the 2.11 behaviour: it calls ``set_group_info`` with the
+    group's native store (no extra prefix) and populates the Python-side guard
+    dict so that ``enable_symm_mem_for_group`` becomes a no-op for this group.
+
+    It also patches ``destroy_process_group`` to preserve the group counter so
+    that if the WORLD group is destroyed and recreated, the new group receives a
+    unique name that does not collide with stale entries in the C++ map.
+    """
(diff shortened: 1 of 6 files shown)


flashinfer/comm/torch_symmetric_memory.py
```diff
@@ -0,0 +1,127 @@
...
+
+import torch
+import torch.distributed._symmetric_memory as symm_mem
+import torch.distributed.distributed_c10d as c10d
+
+_compat_patched = False
+
+
+def _patch_group_count_reset() -> None:
+    """Prevent group_count from resetting to 0 on WORLD destruction (2.10 only).
```
nvcastet: @kwen2501 Is this hack necessary for pytorch 2.10?

nvcastet: @kwen2501 Could you have a look at this PR?

flashinfer/comm/trtllm_ar.py
```diff
@@ -451,27 +454,44 @@ def trtllm_create_ipc_workspace_for_all_reduce(
...
+        (buffer_size, torch.float32),
+        (buffer_size, torch.float32),
+        (flag_size, torch.int32),
+        (flag_size, torch.int32),
+        (lamport_buffer_size, torch.float16),
+        (lamport_buffer_size, torch.float16),
+        (lamport_buffer_size, torch.float16),
     ]:
         # all sizes should be aligned to 1LU << 21 bytes (2MB)
         aligned_size = round_up(size, 1 << 21)
```
nvcastet: Why ? torch symm memory will use a mempool under the cover so that you can have smaller requests. Probably good to be make sure it is 16B aligned but not 2MB.

flashinfer/comm/trtllm_mnnvl_ar.py
```diff
@@ -173,9 +201,24 @@ def __init__(
...
-        self.mc_ptr = self.mcast_buffer_handle.get_multicast_ptr()
+        self.uc_ptrs_dev = self.handle.buffer_ptrs_dev
+        self.uc_ptr_local = self.handle.buffer_ptrs[self.rank]
+        self.mc_ptr = self.handle.multicast_ptr
+
+    def lamport_initialize(self, rank: int, dtype: torch.dtype, allocated_size: int):
+        if dtype == torch.bfloat16 or dtype == torch.float16:
+            neg_zero = 0x8000
+            dsize = 2
+            memset_func = cuda.cuMemsetD16
```
nvcastet: why can't you use `tensor.fill_(-0.0)` ?

flashinfer/comm/torch_symmetric_memory.py
kwen2501: Hmm, do we need to support the case of in-process restart (hence calling `init_process_group` twice)?

flashinfer/comm/torch_symmetric_memory.py
```diff
@@ -0,0 +1,127 @@
...
+    PyTorch 2.10's ``enable_symm_mem_for_group`` wraps the group store in an
+    extra ``PrefixStore("symmetric_memory-{ranks}", ...)``.  PyTorch 2.11
+    removed this indirection and resolves the group store directly in C++ via
+    ``group->getStore()``.  On 2.10 the extra layer can cause the store-based
+    handle exchange inside ``make_peer_alloc_info`` to hang on certain
+    topologies (Blackwell MNNVL fabric handles).
+
+    This helper mimics the 2.11 behaviour: it calls ``set_group_info`` with the
+    group's native store (no extra prefix) and populates the Python-side guard
+    dict so that ``enable_symm_mem_for_group`` becomes a no-op for this group.
```
kwen2501: Sorry I am a bit confused. 
torch 2.11 purposely deprecates the `enable_symm_mem_for_group` API. 
Should user just check the torch version, and call `enable_symm_mem_for_group` when version is lower than 2.11? That is it.

flashinfer/comm/torch_symmetric_memory.py
```diff
@@ -0,0 +1,127 @@
...
+        Tuple of (per-peer data pointers, local tensor, symmetric memory handle).
+    """
+    # Ensure symmetric memory is set up with the correct store before
+    # rendezvous.  On PyTorch 2.10 this avoids an extra PrefixStore layer
+    # that can cause hangs on Blackwell/MNNVL topologies.
+    _enable_symm_mem_compat(group_name)
+
+    elem_size = torch.empty(0, dtype=dtype).element_size()
+    numel = size_bytes // elem_size
+    tensor = symm_mem.empty(numel, dtype=dtype, device=device)
```
kwen2501, line 70: Would it be more ergonomic if the API asks for shape or numel instead of size_bytes?

flashinfer/comm/torch_symmetric_memory.py
```diff
@@ -0,0 +1,127 @@
...
+    handle = symm_mem.rendezvous(tensor, group=group_name)
+    ptrs: list[int] = [
+        handle.get_buffer(peer, (numel,), dtype, storage_offset=0).data_ptr()
+        for peer in range(world_size)
+    ]
+    return ptrs, tensor, handle
```
kwen2501, line 76: nit: `ptrs` and `handle` are redundant to each other in this return. When user has the `handle`, they can get the ptrs themselves.

kwen2501:
One general thought I have is that:
do those trtllm ops still need to have the sticky "workspace"?
whenever they need such workspace, they can now just call:
```
t = symm_mem.empty(shape, dtype, device)
```
This call uses internal pooling of torch symm_mem, thus reusable.

kwen2501:
Correction:
Oh, for Lamport, I see why you need the workspace -- ping-pong buffering?

nvpohanh:
cc @wenscarl not sure how this will affect #3053

flashinfer/comm/trtllm_ar.py
Amir-19: fixed!

flashinfer/comm/trtllm_mnnvl_ar.py
Amir-19: fixed! thank you! I learned something new.

flashinfer/comm/torch_symmetric_memory.py
Amir-19: I haven't investigate deeply to what exactly is happening but if I do 
```
    torch_version = tuple(int(x) for x in torch.__version__.split(".")[:2])
    if torch_version >= (2, 11):
        return
    from torch.distributed._symmetric_memory import enable_symm_mem_for_group
    enable_symm_mem_for_group(group_name)
```
`mpirun -np 2 pytest tests/comm/test_allreduce_unified_api.py -vv -s` hangs after the first test case. adding `_patch_group_count_reset()` fixes the issue.

flashinfer/comm/torch_symmetric_memory.py
Amir-19: we can add another api or option to use this api with shape or numel but I wanted to be least intrusive in the kernels.

aleozlx (approved):

wenscarl:
> cc @wenscarl not sure how this will affect #3053

Verified this PR doesn't resolve the hang in #3053.

PR #2684, opened 2026-03-04
Dispatch GQA group size 5, 6, 7

ahadnagy:
## 📌 Description

Models like Qwen3.5-27B use 24 Q heads with 4 KV heads, producing a
GQA group_size of 6. The existing macro only handled {1, 2, 3, 4, 8},
causing "Unsupported group_size: 6" errors at batch decode plan time.

Add the missing group sizes 5, 6, and 7 so that non-power-of-2 GQA
ratios are dispatched correctly.
## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [ ] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [ ] I have installed the hooks with `pre-commit install`.
- [ ] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [ ] Tests have been added or updated as needed.
- [ ] All tests are passing (`unittest`, etc.).

## Summary by CodeRabbit

* **New Features**
  * Extended inference engine support for additional grouped query attention configurations, enabling broader compatibility with diverse model parameters and enabling more flexible inference operations across different model architectures.

PR #2986, opened 2026-04-05
Add GQA group_size 5, 6, 7 to DISPATCH_GQA_GROUP_SIZE

arbi-dev:
## Summary

`DISPATCH_GQA_GROUP_SIZE` only handles group sizes 1, 2, 3, 4, 8. Any other value hits a runtime error:

```
RuntimeError: Unsupported group_size: 6
```

This breaks several popular models with non-power-of-2 GQA ratios:

| Model | Q heads | KV heads | group_size |
|-------|---------|----------|------------|
| Qwen3.5-27B | 24 | 4 | **6** |
| InternLM2.5-20B | 48 | 8 | **6** |
| Qwen2.5-7B | 28 | 4 | **7** |
| Yi-1.5-34B | 56 | 8 | **7** |

This PR adds explicit `constexpr` cases for group sizes 5, 6, and 7, so all sizes 1-8 are supported. Each adds one template instantiation per call site, matching the existing dispatch pattern.

## Why this hasn't been reported widely

Most users access FlashInfer through vLLM or SGLang, which use `BatchDecodeWithPagedKVCacheWrapper`. That wrapper handles GQA at the Python level and doesn't go through `DISPATCH_GQA_GROUP_SIZE`. The error only manifests when calling the lower-level C++ kernel dispatch directly (e.g., custom attention backends or quantized KV cache implementations that bypass the Python wrapper).

## Test plan

- [x] Verified Qwen3.5-27B (group_size=6) runs correctly after fix
- [ ] CI: existing unit tests should pass (no behavior change for sizes 1-4, 8)

AI-assisted: Claude Opus 4.6 assisted with code generation and model survey.

## Summary by CodeRabbit

* **Bug Fixes**
  * Extended support for additional group sizes (5, 6, and 7) in grouped query attention operations that were previously unsupported.

### DIFF
--- a/include/flashinfer/utils.cuh
+++ b/include/flashinfer/utils.cuh
@@ -147,6 +147,15 @@
   } else if (group_size == 4) {                              \
     constexpr size_t GROUP_SIZE = 4;                         \
     __VA_ARGS__                                              \
+  } else if (group_size == 5) {                              \
+    constexpr size_t GROUP_SIZE = 5;                         \
+    __VA_ARGS__                                              \
+  } else if (group_size == 6) {                              \
+    constexpr size_t GROUP_SIZE = 6;                         \
+    __VA_ARGS__                                              \
+  } else if (group_size == 7) {                              \
+    constexpr size_t GROUP_SIZE = 7;                         \
+    __VA_ARGS__                                              \
   } else if (group_size == 8) {                              \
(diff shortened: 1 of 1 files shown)


yzh119:
Hi @arbi-dev  can you see my comments in https://github.com/flashinfer-ai/flashinfer/pull/2684#pullrequestreview-3890637272

PR #2927, opened 2026-03-31
feat: SM121 (GB10) tile filtering and autotuner robustness

askliar:
## Summary

- **Unified SMEM tile filter** — extracted `tile_fits_smem(sm, m, n, k_elem, config, stages=2)` helper that encodes the per-architecture SMEM limit (102,400 B for SM 8.6/8.9/12.x, per CUDA Programming Guide) and is shared by both the SM12x TMA WS path and the Ampere mainloop path.
- **SM12x FP4 tile filtering** — `get_candidate_configs_sm120` now uses `tile_fits_smem` to filter tiles that exceed the 100 KB SMEM budget. Tile table stores K in SMEM elements (FP4 unpacked: K_elem = K_packed_bytes × 2). SM121 reuses the same function via the `sm` parameter.
- **Ampere path SMEM filtering** — `get_candidate_configs` Ampere loop uses the same `tile_fits_smem` helper, replacing the previous inline `smem_limit`/`sizeof_a`/`sizeof_b` variables.
- **Debug logging** — filtered tiles emit `TLLM_LOG_DEBUG` with SM, M, N, K, and stages. Visible in debug builds or when the logger level is explicitly lowered; silent in release by default.
- **Fix autotuner async CUDA error propagation** — `torch.cuda.synchronize()` drain after failed tactic probes clears sticky async CUDA errors (e.g. `cudaErrorIllegalInstruction`) before they surface during CUDA graph capture.
- **Autotuner tactic skip counter** — instead of per-tactic `WARNING` log spam, a single `INFO`-level summary "Skipped N unsupported tactic(s)" is emitted per profile shape; per-tactic detail remains at `DEBUG`.
- **Separate JIT cache key for SM121** — SM12.1 normalizes to arch key `121f` instead of sharing `120f` with SM12.0.

## Architecture notes

| GPU | Arch | SMEM/block (optin) |
|-----|------|--------------------|
| GB200 | SM100 | ~228 KB |
| RTX PRO 6000 Blackwell | SM120 | 100 KB (101,376 B optin) |
| GB10 (Project DIGITS) | SM121 | 100 KB (same as SM120) |

SM120 and SM121 share the same 100 KB SMEM budget. `tile_fits_smem` uses `smem_limit = 102400` for both, consistent with the CUDA Programming Guide for SM 8.6/8.9/12.x.

**FP4 tiles and SMEM check (SM12x, 2 stages, sizeof=1):**

| Tile | M | N | K_elem | 2×(M+N)×K | Fits? |
[...]

PR #2997, opened 2026-04-06
feat: offline tactics blocklist for GPU-aware autotuner pruning

qiching:
## Summary

Add a GPU-aware pruning framework that records known-invalid CUTLASS tactics via offline profiling and filters them before autotuning, eliminating redundant failures, noisy error logs, and wasted GPU cycles across all autotuned MoE backends.

### Motivation

The FlashInfer autotuner currently attempts every registered tactic during profiling, including those that are guaranteed to fail on certain GPUs due to shared memory (SMEM) constraints. On SM121 (DGX Spark / GB10), this results in:

- **19 tactics failing at init** on every autotuning run (SMEM OOM before kernel launch)
- Noisy error logs that obscure real issues and confuse users
- Wasted profiling time on dead-code paths

This is a runtime complement to PR #2927 (compile-time SMEM heuristic filtering): that PR prunes tiles that *statically* cannot fit; this PR catches tactics that pass static checks but fail at runtime on specific hardware.

### Approach

Offline probe-and-record: a standalone script exercises each MoE backend on the target GPU, captures which tactics fail during profiling, and serializes them to a JSON blocklist. At runtime, the autotuner loads this file and skips blocked tactics *before* profiling begins.
### 1. `TacticsBlocklist` class (`flashinfer/tactics_blocklist.py`)
- **load()**: Reads a JSON blocklist with GPU metadata validation (auto-skips if GPU mismatch)
- **filter()**: Removes known-invalid tactics from the candidate list before profiling; unknown/new tactics always pass through (safe-by-default)
- **save()**: Atomic write (tempfile + `os.replace`) to prevent corruption from interrupted runs

### 2. Autotuner integration (`flashinfer/autotuner.py`)
- Loads blocklist from `FLASHINFER_TACTICS_BLOCKLIST` env var on init
- Applies `filter()` after `get_valid_tactics()` but before profiling loop
- Records failed tactics in `stats.failed_tactics` for the probe script to collect

### 3. Offline probe script (`scripts/generate_tactics_blocklist.py`)
5 probe paths covering all autotuned MoE backends:
| Probe | Backend | Quant |
|-------|---------|-------|
| `NvFP4xNvFP4` | TRT-LLM | FP4 block-scale |
| `Fp8-Block` | TRT-LLM | FP8 block-scale |
| `NvFP4-CUTLASS` | CUTLASS | FP4 (vLLM path) |
| `Fp8-PerTensor-CUTLASS` | CUTLASS | FP8 per-tensor |
| `BF16-CUTLASS` | CUTLASS | Unquantized BF16 (MTP/drafter) |

### 4. Test suite (`tests/autotuner/test_tactics_blocklist.py`)
37 unit tests covering load/filter/save semantics and autotuner integration — no GPU required.

## Usage

```bash
# Step 1: Generate blocklist on target GPU
python scripts/generate_tactics_blocklist.py \
    --num-experts 256 --hidden-size 7168 \
    --output /tmp/tactics_blocklist.json

# Step 2: Use at runtime
export FLASHINFER_TACTICS_BLOCKLIST=/tmp/tactics_blocklist.json
python your_app.py  # blocked tactics silently skipped before profiling
```

### GB200 (SM100) / GB300 (SM103)
All tactics pass, probe correctly reports "no blocklist file needed." Confirms zero false-positive risk on GPUs with sufficient SMEM.

### DGX Spark / GB10 (SM121), tested by @askliar

19 invalid tactics captured across 3 quant modes:

| Operation | Failed Tactics | Count |
|-----------|---------------|-------|
| `gemm1::MoERunner` | 4, 5, 7, 10, 11, 19, 20 | 7 |
| `gemm2::MoERunner` | 16, 17, 20, 21, 22, 23, 28, 29, 34, 35, 40, 41 | 12 |

<details>
<summary>Full SM121 probe output</summary>

```
[NvFP4-CUTLASS]
  Probing NvFP4xNvFP4 via CUTLASS path ...
    trtllm::fused_moe::gemm1::MoERunner: 4 failed tactic(s)  [10, 11, 4, 5]
    trtllm::fused_moe::gemm2::MoERunner: 10 failed tactic(s) [34, 35, 16, 17, 20, 21, 22, 23, 28, 29]

[Fp8-PerTensor-CUTLASS]
  Probing Fp8-PerTensor via CUTLASS path ...
    trtllm::fused_moe::gemm1::MoERunner: 2 failed tactic(s)  [19, 20]
    trtllm::fused_moe::gemm2::MoERunner: 2 failed tactic(s)  [40, 41]

[BF16-CUTLASS]
  Probing BF16 via CUTLASS path ...
    trtllm::fused_moe::gemm1::MoERunner: 2 failed tactic(s)  [5, 7]

Saved 19 invalid tactic(s) to /tmp/spark_whitelist.json
Completed in 74.8s
```

</details>

Generated JSON:
```json
{
  "_metadata": {
    "flashinfer_version": "0.6.9",
    "cuda_version": "13.0",
    "gpu": "NVIDIA GB10",
    "generator_version": "1.0"
  },
  "invalid_tactics": {
    "trtllm::fused_moe::gemm1::MoERunner": [19, 4, 5, 20, 7, 10, 11],
    "trtllm::fused_moe::gemm2::MoERunner": [34, 35, 40, 41, 16, 17, 20, 21, 22, 23, 28, 29]
  }
}
```

## Design Decisions

- **Blocklist*: Unknown and new tactics pass through by defaultm, safer for forward compatibility when new kernels are added
- **GPU metadata check**: Blocklist generated for GPU A is auto-skipped on GPU, preventing cross-device false positives
- **Atomic save**: Uses `tempfile` + `os.replace` to prevent corrupt JSON from interrupted probe runs
- **no regression risk**: When no blocklist is loaded, all code paths are unchanged

## Related PRs
- #2927 — Compile-time SMEM heuristic filtering for SM120 (complementary approach)
- #2926 — ReLU2 / unquantized MoE support for SM121

## PR Checklist
- [x] Pre-commit hooks pass
- [x] 37 unit tests added and passing
- [x] E2E validated on GB200 (SM100), GB300 (SM103), GB10 (SM121)
- [x] No breaking changes to existing API

## Acknowledgments
- @djns99  for the inspire this approach
- @askliar for contributed CUTLASS probe paths and validated on DGX Spark (SM121)
- @aleozlx  for guidance

## Summary by CodeRabbit

* **New Features**
  * Added blocklist support to skip known-invalid tactics during autotuning profiling
  * New CLI command to generate GPU-specific tactics blocklists from local hardware
  * Autotuner now tracks failed tactics for offline blocklist generation

[Review Change Stack](https://app.coderabbit.ai/change-stack/flashinfer-ai/flashinfer/pull/2997?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


flashinfer/tactics_whitelist.py
```diff
@@ -0,0 +1,184 @@
...
+            _METADATA_KEY: meta,
+            "invalid_tactics": serialized,
+        }
+
+        dir_name = os.path.dirname(os.path.abspath(path))
+        if dir_name:
+            os.makedirs(dir_name, exist_ok=True)
+
+        with open(path, "w") as f:
+            json.dump(output, f, indent=2)
```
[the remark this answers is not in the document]
qiching: updated.

flashinfer/autotuner.py
```diff
@@ -793,6 +826,15 @@ def choose_one(
...
                                         torch.cuda.cudart().cudaGetLastError()
 
+                                    # Record the failed tactic value for the
+                                    # whitelist generator.
+                                    ft_key = f"{custom_op}::{r.__class__.__name__}"
+                                    if ft_key not in self.stats.failed_tactics:
+                                        self.stats.failed_tactics[ft_key] = set()
+                                    self.stats.failed_tactics[ft_key].add(
+                                        _tactic_to_json_hashable(tac)
+                                    )
```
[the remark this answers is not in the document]
qiching: update

flashinfer/autotuner.py
```diff
@@ -573,6 +593,16 @@ def __init__(self, warmup=3, repeat=10, stream_delay_micro_secs=1000):
...
 
+        # Offline tactics whitelist (loaded via env var or explicit call).
+        # Lazy import to avoid circular dependency (tactics_whitelist
+        # imports _METADATA_KEY / _collect_metadata from this module).
+        from .tactics_whitelist import TacticsWhitelist
+
+        self._whitelist = TacticsWhitelist()
+        _wl_path = os.environ.get("FLASHINFER_TACTICS_WHITELIST")
+        if _wl_path and os.path.isfile(_wl_path):
+            self._whitelist.load(_wl_path)
```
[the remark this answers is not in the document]
qiching: updated.

flashinfer/autotuner.py
```diff
@@ -489,6 +506,9 @@ class AutoTunerStatistics:
     ] = field(default_factory=dict)
     tuned_op_total_configs: Dict[str, int] = field(default_factory=dict)
     tuned_op_successful_configs: Dict[str, int] = field(default_factory=dict)
+    # Maps "custom_op::RunnerClass" to sets of tactic values that failed.
+    # Used by the offline whitelist generator to extract per-tactic pass/fail.
+    failed_tactics: Dict[str, Set] = field(default_factory=dict)
```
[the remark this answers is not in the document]
qiching: updated.

flashinfer/tactics_blocklist.py
```diff
@@ -0,0 +1,197 @@
...
+)
+from .jit.core import logger
+
+
+class TacticsBlocklist:
+    """Blocklist-based tactic filter backed by an offline-generated JSON file.
+
+    Stores known-invalid tactics and removes them from the candidate list.
+    Unknown or new tactics are allowed through, which is safer than a strict
+    allowlist that would block anything not explicitly approved.
```
djns99, line 50: Do tactics actually have information that can be used for "unrecognised" tactics. From memory it just returns opaque integers so there is no way to determine if the meaning has changed under the hood

djns99 (commented):
Will look more comprehensively in a bit. Basic approach seems sensible, ideally this would be part of CI, do you have any thoughts on how to achieve that?

flashinfer/tactics_blocklist.py
qiching: For CI integration, we could add a workflow that runs scripts on each supported GPU type after a release or on a scheduled basis, then commits the resulting JSON files to the repo. Happy to explore this as a follow up.


For second:
ou're right that tactic IDs are opaque integers with no inherent versioning. I've addressed this in the latest commit: `load()` now validates all metadata (flashinfer_version, cuda_version, cublas_version, cudnn_version, gpu) against the current runtime. If any field mismatches, the blocklist is skipped with a warning.

How do you think? thanks!

djns99: I think this makes sense. Though perhaps we should fail and ask users to explicitly opt-out of using the blocklist with an API param? Not sure if this is important enough to fail things for though

I was thinking about how to CI this, maybe it makes sense to have official "golden" files for some reference model shapes (say deepseek/qwen) and then CI reruns the autotuner to check the same configs pass. Then we can fail CI if one of the tactics starts failing, or if new tactics are added

flashinfer/tactics_blocklist_gen.py
```diff
@@ -0,0 +1,547 @@
...
+
+
+PROBE_FUNCTIONS = {
+    "NvFP4xNvFP4": _probe_fp4,
+    "Fp8-Block": _probe_fp8_block,
+    # CUTLASS-path probes (for vLLM flashinfer_cutlass backend)
+    "NvFP4-CUTLASS": _probe_fp4_cutlass,
+    "Fp8-PerTensor-CUTLASS": _probe_fp8_per_tensor_cutlass,
+    "BF16-CUTLASS": _probe_bf16_cutlass,
+}
```
djns99, line 508: Do we need some MXFP8 and MXINT4 cases?

scripts/generate_tactics_blocklist.py
```diff
@@ -0,0 +1,547 @@
...
+    parser.add_argument("--hidden-size", type=int, default=7168)
+    parser.add_argument("--intermediate-size", type=int, default=2048)
+    parser.add_argument("--top-k", type=int, default=8)
+    parser.add_argument(
+        "--device",
+        type=str,
+        default="cuda:0",
+        help="CUDA device to probe (default: cuda:0)",
+    )
+    args = parser.parse_args()
```
djns99: We should have some pre-baked recipes for deepseek/qwen/kimi/gpt-oss

scripts/generate_tactics_blocklist.py
```diff
@@ -0,0 +1,547 @@
...
+                token_selected_experts=token_selected_experts,
+                token_final_scales=token_final_scales,
+                fc1_expert_weights=fc1,
+                fc2_expert_weights=fc2,
+                output_dtype=torch.bfloat16,
+                quant_scales=quant_scales,
+                input_sf=input_sf,
+                output=output,
+                enable_pdl=enable_pdl,
+                activation_type=ActivationType.Swiglu,
```
djns99: I think we should also have a Relu2 activation function to test non-gated activation

flashinfer/tactics_blocklist_gen.py
```diff
@@ -0,0 +1,547 @@
...
+    )
+    w2_scale = torch.full(
+        (
+            num_experts,
+            hidden_size // scale_vec_size,
+            intermediate_size // scale_vec_size,
+        ),
+        w2_sc.item(),
+        device=device,
+    )
```
djns99, line 206: Not in the scope of this PR, but @aleozlx its probably worth having a generic "dummy weight setup" utility function/helper since we have a lot of tests that all do the same sort of thing

flashinfer/tactics_blocklist.py
qiching: I lean towards the current warning and skip approach, since failing would break users who simply upgraded a library without regenerating blocklists, and the fallback behavior (no blocklist = full profiling) is safe. If needed, we could add a strict mode via env var (e.g. FLASHINFER_TACTICS_BLOCKLIST_STRICT=1) in the future, but I'd keep the default forgiving.

Love the golden file idea! We could maintain reference blocklists for standard shapes and have CI re-probe periodically to detect regressions happy to design this in a follow up PR.

scripts/generate_tactics_blocklist.py
qiching: @askliar how do you think?

flashinfer/tactics_blocklist_gen.py
qiching: The current 5 probes cover NvFP4/FP8-block (TRT-LLM) and FP4/FP8-per-tensor/BF16 (CUTLASS). I'll add MXFP8 and MXINT4 probe paths in a follow-up PR, they need a fresh on-hardware probe run to validate.

scripts/generate_tactics_blocklist.py
qiching: Agreed, presets would make this much easier to use. I'll add named recipes (deepseek/qwen/kimi/gpt-oss) as a follow-up so users don't have to pass shapes manually.

scripts/generate_tactics_blocklist.py
qiching: Agreed it'd broaden coverage beyond the gated path. I'll bring it in via a follow-up rather than re-trigger hardware validation on this PR.

scripts/generate_tactics_blocklist.py
```diff
@@ -0,0 +1,547 @@
+#!/usr/bin/env python3
```
yzh119: Do you expect this script to be part of the package? (by default everything under `scripts` wouldn't be part of the python wheel.

scripts/generate_tactics_blocklist.py
askliar: Yes, that would make sense to me. Especially given that Nemotron is using Relu2

qiching: so follow-up separate PR or we add in this PR？

askliar: adding it to this one would make most sense. It's a 1-2 lines change

qiching: yes. i will add now and could you test e2e if after i update it? thanks!

qiching: Updated! Could you please run the e2e on your end when you get a chance?

two changes in this push:

Non-gated Relu2 probe (BF16-Relu2-CUTLASS). It's a bit more than a one-liner because Relu2 is non-gated: GEMM1 produces an intermediate_size projection instead of Swiglu's 2*intermediate_size, so it has its own input setup. _run_cutlass_probe now takes an activation_type arg defaulting to Swiglu, so all existing probes are unchanged.

Packaged the generator:
Unit tests pass 42/42 locally. The new probe path is the only thing that hasn't been hardware-validated yet, that's the part I'd love your e2e to confirm. Thanks!

scripts/generate_tactics_blocklist.py
qiching: Done! Thanks! I moved the standalone scripts/generate_tactics_blocklist.py into the package as flashinfer/tactics_blocklist_gen.py and exposed it through the CLI:
```
flashinfer generate-tactics-blocklist --output tactics_<gpu>.json
```

The heavy imports are kept lazy and only pulled in when the command actually runs, so the package import path stays clean. pip-installed users can now run the probe without needing the source tree.

scripts/generate_tactics_blocklist.py
qiching: @askliar @djns99 Could you have time to take a look? Thanks!

djns99: Looks ok to me. Nemotron 3 Super is the main Relu2 model I know of - it uses FP8. We should add a case for that (can be a follow-up) when you have time to run the probe

aleozlx (approved):

Issue #2555, opened 2026-02-13
SM120 attention kernels exist but are blocked by wiring issues (fmha_v2, backend selector, MLA)


blake-snc:
## Summary

We've done a deep investigation of FlashInfer's SM120 support on an NVIDIA DGX Spark (GB10, SM121a, CUDA 13.0) and found that **SM120 attention kernels already exist in the codebase but are not wired into the runtime pipeline**. Three specific wiring issues prevent SM120 from using its optimized paths:

## Issue 1: FMHA_V2 SM120 kernels gated behind `ENABLE_SM120`

**Location:** `flashinfer/jit/attention/fmha_v2/generator_utils.py:6772-6786`

SM120-specific FMHA_V2 kernels are only generated if the `ENABLE_SM120` environment variable is set:

```python
if "ENABLE_SM120" in os.environ:
    enumerate_hmma_flash_kernels(specs, sm=120, dtype="fp16")
    enumerate_hmma_flash_kernels(specs, sm=120, dtype="bf16")
    # ...
```

Three SM120 kernel files are referenced in `jit/attention/modules.py:1918-1920` (BF16 and FP8 variants) but never get generated in standard builds. This means `gen_trtllm_fmha_v2_module()` can't find them at runtime.

**Expected:** SM120 kernels should be generated by default when building for SM120 targets.

## Issue 2: `determine_attention_backend()` routes SM120 to generic FA2

**Location:** `flashinfer/utils.py:455-497`

```python
def determine_attention_backend(device, ...):
    if is_sm90a_supported(device) and is_fa3_backend_supported(...):
        return "fa3"
    else:
        return "fa2"
```

This only checks for SM90. SM120 falls through to generic FA2 (Ampere-level), even though SM120-specific FMHA_V2 kernels exist. The function should have an SM120 path that routes to the fmha_v2 SM120 backend.

## Issue 3: `determine_mla_backend()` routes SM120 to FA2 instead of XQA

**Location:** `flashinfer/utils.py:561-562`

```python
def determine_mla_backend(device):
    return "fa3" if is_sm90a_supported(device) else "fa2"
```

SM120 gets `fa2` for MLA, but FlashInfer already has a functional XQA MLA backend specifically for SM120 (`data/csrc/xqa/mla_sm120.cu`, compiled with `supported_major_versions=[12]` in `jit/xqa.py:137`). This function should return `"xqa"` for SM120.

## What Currently Works on SM120

| Component | Status |
|---|---|
[...]

PR #3016, opened 2026-04-08
feat: enable fmha_v2 HMMA attention for SM120 standard shapes

blake-snc:
## Summary

Enables the existing TRT-LLM fmha_v2 HMMA flash attention kernels for SM120 (RTX 5090, DGX Spark) with standard head dimensions (64/128/256). These kernels use Ampere-compatible HMMA tensor core instructions (`sm_mma=80`) which SM120 supports natively.

The fmha_v2 SM120 kernel infrastructure already existed for DeepSeek MLA shapes (192x128) but standard attention kernels (bf16/fp16) were not enumerated in the JIT generation path. This PR adds them and wires fmha_v2 into `BatchPrefillWithRaggedKVCacheWrapper` as an automatic backend for SM12x MHA at sequence lengths >= 256.

## Changes

- `flashinfer/jit/attention/fmha_v2/generate_kernels.py`: Add `enumerate_hmma_flash_kernels(specs, sm=120, dtype="bf16")` and `fp16` for standard shapes
- `flashinfer/utils.py`: Add `_should_use_fmha_v2_sm120()` helper
- `flashinfer/prefill.py`: Wire fmha_v2 into BatchPrefill plan/run with conditional selection

## Backend selection logic

```
SM12x + MHA + seqlen >= 256 + bf16/fp16 + no custom mask → fmha_v2
SM12x + GQA or seqlen < 256 → fa2 (fallback)
```

## SM121a (DGX Spark) validation

**Correctness** (14/14 configs, all vs f32 reference):

| Config | Error |
|--------|-------|
| B=1 S=64 H=4 D=64 bf16 causal | 0.007 |
| B=2 S=128 H=4 D=64 bf16 causal | 0.007 |
| B=1 S=64 H=8 D=128 bf16 causal | 0.008 |
| B=1 S=256 H=4 D=128 bf16 noncausal | 0.001 |
| B=1 S=128 H=4 D=128 fp16 causal | 0.003 |

**Performance** (CUDA event timing, bf16 causal):

| Config | fmha_v2 | FA2 | ratio |
|--------|---------|-----|-------|
| B=2 S=512 H=4 D=64 | 0.050ms | 0.073ms | **0.69x** |
| B=2 S=512 H=4 D=128 | 0.037ms | 0.041ms | **0.91x** |
| B=1 S=1024 H=4 D=128 | 0.047ms | 0.053ms | **0.88x** |
| B=1 S=128 H=4 D=64 | 0.020ms | 0.010ms | 2.07x (FA2 wins, below threshold) |

fmha_v2 is 10-31% faster than FA2 at S>=512, but slower at small sequences due to CTA launch overhead. The S>=256 threshold avoids the regression region.

## Related

- Addresses Issue #2555 (Issue 1: SM120 kernel generation, Issue 2: backend routing)
- Complements PR #2598 (CuTe DSL backend — different kernel system, independent)

Contributed by Second Nature Computing (https://joinsecondnature.com)

## Summary by CodeRabbit

* **New Features**
  * Expanded SM120 GPU support with optimized flash-attention kernels for bfloat16 and float16.
  * Automatic selection of the optimized prefill backend for compatible SM120 workloads (>=256 sequence length and matching Q/KV layout).
  * New prefill execution path that uses packed Q/KV inputs and cached sequence/index buffers for faster runs.
  * Note: the optimized prefill path does not support returning LSEs (NotImplemented).

### DIFF
--- a/flashinfer/prefill.py
+++ b/flashinfer/prefill.py
@@ -3359,12 +3359,14 @@ class BatchPrefillWithRaggedKVCacheWrapper:
                     "than the expected [total_q_tokens, num_heads] LSE tensor."
                 )
             fmha_v2_mask_mode = "causal" if self._causal else "padding"
-            # Pack separate q, k, v into [num_tokens, 3, num_heads, head_dim]
-            # for PACKED_QKV layout (the only layout with SM120 standard kernels)
-            qkv_packed = torch.stack([q, k, v], dim=1)
+            # Use CONTIGUOUS_Q_KV layout: pass Q separately and stack only
+            # K+V into [num_tokens, 2, num_kv_heads, head_dim]. This avoids
+            # the 3-way torch.stack copy that PACKED_QKV requires.
+            # (fmha_v2 auto-selection already requires MHA, so Q/K head counts match)
+            kv_packed = torch.stack([k, v], dim=1)
             out = trtllm_fmha_v2_prefill(
-                qkv_packed,
-                input_layout="PACKED_QKV",
+                (q, kv_packed),
+                input_layout="CONTIGUOUS_Q_KV",
                 workspace_buffer=self._float_workspace_buffer,
                 seq_lens=self._fmha_v2_seq_lens,
                 max_q_len=self._fmha_v2_max_q_len,
--- a/flashinfer/prefill.py
+++ b/flashinfer/prefill.py
@@ -3364,6 +3364,14 @@ class BatchPrefillWithRaggedKVCacheWrapper:
             # the 3-way torch.stack copy that PACKED_QKV requires.
             # (fmha_v2 auto-selection already requires MHA, so Q/K head counts match)
             kv_packed = torch.stack([k, v], dim=1)
+            # Pass window_left and sinks for sliding-window + attention sink
+            # support. sinks is stored on the wrapper by vLLM's metadata builder.
+            fmha_v2_sinks = getattr(self, "_sinks", None)
+            fmha_v2_kwargs = {}
+            if self._window_left >= 0:
+                fmha_v2_kwargs["window_left"] = self._window_left
+            if fmha_v2_sinks is not None:
+                fmha_v2_kwargs["sinks"] = fmha_v2_sinks
             out = trtllm_fmha_v2_prefill(
                 (q, kv_packed),
                 input_layout="CONTIGUOUS_Q_KV",
@@ -3379,6 +3387,7 @@ class BatchPrefillWithRaggedKVCacheWrapper:
                 out=out,
                 mask_mode=fmha_v2_mask_mode,
                 save_softmax_stats=False,
+                **fmha_v2_kwargs,
             )
             return out
         elif self._backend == "cute-dsl":
--- a/flashinfer/prefill.py
+++ b/flashinfer/prefill.py
@@ -3006,6 +3006,10 @@ class BatchPrefillWithRaggedKVCacheWrapper:
         self._max_token_per_sequence = max_token_per_sequence
         self._max_sequence_kv = max_sequence_kv
 
+        # Precompute sequence length maxima (used by fmha_v2 and other backends)
+        max_seq_in_batch = int(kv_len_arr.max().item())
+        max_qo_len = int((qo_indptr_host[1:] - qo_indptr_host[:-1]).max().item())
+
         if self._backend == "cute-dsl":
             if custom_mask is not None or packed_custom_mask is not None:
                 raise NotImplementedError(
@@ -3074,10 +3078,6 @@ class BatchPrefillWithRaggedKVCacheWrapper:
                 # launch overhead for small sequences or high head counts.
                 # Only supports: MHA, standard symmetric head dims, no sliding
                 # window, no softcap, equal q/kv segments, NHD layout.
-                max_seq_in_batch = int(kv_len_arr.max().item())
-                max_qo_len = int(
-                    (qo_indptr_host[1:] - qo_indptr_host[:-1]).max().item()
-                )
                 same_qk_segments = torch.equal(qo_indptr_host, kv_indptr_host)
                 is_standard_shape = head_dim_qk == head_dim_vo and head_dim_qk in (
                     64,
--- a/flashinfer/prefill.py
+++ b/flashinfer/prefill.py
@@ -53,6 +53,7 @@ from .utils import (
     _get_cache_alibi_slopes_buf,
     _get_cache_buf,
     _unpack_paged_kv_cache,
+    _should_use_fmha_v2_sm120,
     canonicalize_torch_dtype,
     determine_attention_backend,
     device_support_pdl,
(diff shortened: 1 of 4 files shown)


blake-snc:
Addressed the review feedback — here's the status of each item:

**Critical — `max_qo_len`/`max_seq_in_batch` scope** (@coderabbitai): Already fixed. Both are computed unconditionally at lines 2973-2974 (before the `if self._backend == "auto":` block), so explicit `backend="fmha_v2"` works correctly.

**Major — `mask_mode` variable reuse** (@coderabbitai): Already fixed. The fmha_v2 path uses a dedicated `fmha_v2_mask_mode` variable (line 3272) to avoid type confusion with the int `mask_mode` used elsewhere.

**Major — TRT-LLM raw stats as LSE** (@coderabbitai): Already handled. `return_lse=True` raises `NotImplementedError` with an explanation of the format mismatch (lines 3268-3271). Proper LSE conversion would require understanding TRT-LLM's `(total_tokens, num_heads, 2)` stats format.

**Major — workspace buffer snapshot** (@coderabbitai): Not an issue. `run()` reads `self._float_workspace_buffer` at call time (line 3281), so `reset_workspace_buffer()` correctly updates the reference. No stale snapshot.

**Medium — GPU-CPU sync in run()** (@gemini-code-assist): Already fixed. `max_q_len`, `max_kv_len`, and int32 indptrs are cached during `plan()` (lines 3037-3043) and reused in `run()` without device sync.

**Medium — torch.stack overhead** (@gemini-code-assist): Acknowledged. PACKED_QKV is the only layout with SM120 standard kernels currently. Separate Q/K/V support would require generating additional kernel variants — tracked as a future optimization.

blake-snc:
Pushed a fix for the `torch.stack` overhead: switched from `PACKED_QKV` (3-way stack of Q,K,V) to `CONTIGUOUS_Q_KV` (Q passed separately, only K+V stacked). This reduces the per-call copy from `3*N*H*D` to `2*N*H*D` elements — ~33% less memory copy overhead.

`CONTIGUOUS_Q_KV` is already supported by the SM120 standard fmha_v2 kernel generator (only `SEPARATE_Q_K_V` is excluded). The fmha_v2 guard already requires MHA (`num_qo_heads == num_kv_heads`), so the layout is safe.

Fully eliminating the copy would require `SEPARATE_Q_K_V` support for standard head dims, but that's gated on TRT-LLM kernel generation changes (currently only available for DeepSeek MLA 192/128).

blake-snc:
Merged in latest main (62 commits) to clear the conflict so this is back in shape for re-review.

**Single conflict** was in `csrc/fmha_v2_run.cu` lines 270-285, where main rewrote the surrounding block from a plain `#if 0` into `#if 0 / #else / #endif` with the active `#else` setting `launch_params.flash_attention = true` unconditionally. I kept the original commentary and the `is_paged_kv` clause in the dead `#if 0` branch for documentation, and accepted main's `#else` as the active path.

Net effect: my original Q_PAGED_KV fix is functionally obviated — main's `flash_attention = true` is a strict superset of the Q_PAGED_KV-only condition I added, so the silent-fallback scenario it guarded against can no longer happen. The dead-code fix is just historical context now.

The other commits on this branch remain valuable independently:
- Attention sinks + `window_left` passthrough for fmha_v2 SM120
- `CONTIGUOUS_Q_KV` layout perf optimization (per-call copy reduction vs `PACKED_QKV`)
- Dispatch guards: `max_item_len_ptr`, `kv_layout` NHD, `prefix_len_ptr`, `token_pos_in_items_ptr`

Submodules: `3rdparty/cutlass` bumped `da5e086d → b46b16d0` to match main; `3rdparty/cccl` added (new in main).

**Validation on SM121a (DGX Spark GB10) before push:**
- `pytest tests/attention/test_fmha_v2_prefill.py::test_fmha_v2_prefill_deepseek` → 9/9 configs pass in 36s (BF16/FP8 mixes × seq_len ∈ {1024, 4096, 8192}, head_dim_qk=192, head_dim_v=128).
- The other 2336 tests in that file skip on SM121a because they're gated to `is_sm90a_supported()` — pre-existing upstream, not introduced by the merge. Hopper CI via `/bot run` should exercise those.

Ready for re-review.

saltyminty (approved):

blake-snc:
Thanks for the approval and the bot run @saltyminty. I can't see the internal pipeline from outside — could you (or the bot) share the failing job names/archs for the 9/20? This PR only widens the fmha_v2 dispatch predicate for SM120 standard shapes, gated on cc 12.x, so non-sm12x failures would be surprising and I'd like to know whether these are the known-flaky suites or something this predicate actually reaches. If any failing job is sm12x, I have DGX Spark (sm_121) hardware and can repro and fix immediately. A re-run to rule out flakes also works if that's easier.

saltyminty:
Related CI failures:
On RTX Pro 6000 and 5090:
tests/attention/test_non_contiguous_prefill.py::test_batch_ragged_prefill_packed_input: fails for `seq_len=257`, MHA shapes `num_qo_heads == num_kv_heads`, and head dims 64/128/256

`python3 -m pytest -q tests/attention/test_non_contiguous_prefill.py::test_batch_ragged_prefill_packed_input \
  -k '257 and (64 or 128 or 256)'`

blake-snc:
Rebased onto current `main` and fixed the failing CI check. Ready for another look.

**Rebase:** the only conflict was in `prefill.py`, where `main` has since evolved the `cute-dsl` run() branch to support LSE. I kept the `fmha_v2` branch and chained it onto main's current `cute-dsl` handling.

**CI fix** (`test_batch_ragged_prefill_packed_input`, seq_len >= 256): the `CONTIGUOUS_Q_KV` layout stacks K+V into a contiguous buffer but forwarded Q directly. The kernel requires Q contiguous, so a non-contiguous (packed/strided) Q was read with the wrong strides and produced garbage output for those shapes. The fix is one line, `q = q.contiguous()` before the call. It is a no-op when Q is already contiguous, so the copy-avoidance optimization is preserved, and a copy only for non-contiguous Q.

**Validation on DGX Spark (GB10, sm_121a):** `fmha_v2` packed vs contiguous is now bit-identical and matches a torch-SDPA reference. The full `tests/attention/test_non_contiguous_prefill.py` passes (1320 passed, 0 failed). Thanks @saltyminty for the earlier review.

flashinfer/prefill.py
```diff
@@ -3534,6 +3539,39 @@ def plan(
...
+                    and prefix_len_ptr is None
+                    and token_pos_in_items_ptr is None
+                    and max_item_len_ptr is None
+                    and _should_use_fmha_v2_sm120(
+                        self.device,
+                        PosEncodingMode[pos_encoding_mode].value,
+                        self._custom_mask_buf is not None,
+                        q_data_type,
+                        kv_data_type,
+                    )
```
saltyminty, line 3577: I believe we also need a gate against:
use_fp16_qk_reduction=False (the kernel enables this by default on bf16)
dtype_q!=dtype_kv (the kernel assumes they are the same)

saltyminty (commented):
Left one minor comment but otherwise LGTM

blake-snc:
Thanks @saltyminty. Pushed both guards to the fmha_v2 auto-selection in `plan()`:

- `q_data_type == kv_data_type`. The wrapper already rejects mixed q/kv dtype upstream, so this is belt-and-suspenders, but it makes the constraint explicit at the selection site.
- `use_fp16_qk_reduction`. Since the kernel reduces QK in fp16 on bf16, only auto-select fmha_v2 when the caller allows fp16 reduction, mirroring `is_fa3_backend_supported` (which returns False when `use_fp16_qk_reductions` is set).

One behavioral point worth confirming, since it changes the default: `use_fp16_qk_reduction` defaults to False, and on SM120 FA2 ignores the flag anyway (identical output either way, so there is no fp16-reduction path there). So these guards make fmha_v2 effectively opt-in, selected only when the caller passes `use_fp16_qk_reduction=True`, and a default prefill call stays on FA2 rather than picking up the fmha_v2 speedup. Is that the scoping you intend (fmha_v2 as the opt-in fp16-reduction fast path), or would you rather it stay auto-selected by default with the fp16 reduction documented? Happy to adjust either way.

Since the existing test uses the default and now routes to FA2 for those cases, I can add a focused test exercising the fmha_v2 path (MHA, seq >= 256, `use_fp16_qk_reduction=True`) to keep it covered. Let me know if you would like that in this PR.

Validated on DGX Spark (sm_121a): with `use_fp16_qk_reduction=True` and matching dtype, fmha_v2 is selected and packed vs contiguous is bit-identical (matches a torch-SDPA reference); default and mixed-dtype cases route to FA2.

PR #3019, opened 2026-04-08
docs(gdn): document -1 padding index semantics for pool+indices path

kaixih:
Document the `-1` padding index semantics for `gated_delta_rule_decode_pretranspose`'s pool+indices path which is a subtle but important behavioral difference between the bf16 fast path (redirects to a sacrificial slot 0 null buffer, output undefined) and the float32 legacy path (skips entirely, output zeroed) that framework integrators must be aware of when handling inactive sequences. 

cc. @kahyunnam @hlu1

## Summary by CodeRabbit

* **Documentation**
  * Clarified documentation regarding edge case handling and behavior across different computational backends, including how padding sequences are processed and what output to expect in specific scenarios.

### DIFF
--- a/flashinfer/gdn_decode.py
+++ b/flashinfer/gdn_decode.py
@@ -166,6 +166,21 @@ def gated_delta_rule_decode_pretranspose(
             entry to its slot in ``initial_state``.  Required when ``initial_state``
             is provided.
 
+            **Padding / inactive sequences**: set the index to ``-1`` for any batch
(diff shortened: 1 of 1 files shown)


hlu1 (commented):
LGTM

kahyunnam (approved):

vadiklyutiy:
vLLM uses `0` as padded index

PR #3032, opened 2026-04-10
fused_moe: pre-filter SM89 tactics with zero occupancy on SM120 Blackwell (fix review feedback on #2764)

### LINKED ISSUE #3031: fused_moe: SM89 tactics fail on SM120 Blackwell for pure FP8 MoE (shared memory assertion)
## Environment

| Component | Version |
|-----------|---------|
| GPU | 8× NVIDIA RTX PRO 6000 Blackwell Server Edition |
| Compute Capability | SM 12.0 |
| CUDA Driver / Runtime | 13.0 / 12.9 |
| vLLM | 0.17.0 |
| flashinfer | 0.6.x (bundled with vLLM 0.17.0) |
| PyTorch | 2.10.0+cu130 |
| Python | 3.12 |
| OS | Ubuntu 24.04, x86_64 |

**Model under test:** `nvidia/nemotron-3-nano` — pure FP8, TP=8 (MoE architecture)

Note: Dense FP8 models and NVFP4 MoE models hit a separate code path and are **not** affected by this issue. This is specific to **pure FP8 MoE** on SM120/SM121.

---

## Problem

During NIM server startup, the flashinfer MoE autotuner profiles CUTLASS FP8 GroupedGEMM tactics loaded from the `fused_moe_120` JIT cache. All tactics using the `cutlass::arch::Sm89` (Ada Lovelace) dispatch path fail with a shared memory assertion error on Blackwell (CC 12.0). This affects all TP workers simultaneously and repeats across multiple tactic indices per MoE layer.

The autotuner catches the failures at profiling time and skips the failing tactics, so the server does eventually start. However, it falls back to non-Blackwell-native kernels, resulting in **suboptimal MoE throughput**.

---

## Reproduction

```bash
docker run --gpus all \
  -e VLLM_USE_FLASHINFER_MOE_FP4=1 \
  nvcr.io/nvstaging/nim-internal/nemotron-nano-3:2.0.0-rc.20260311082200 \
  --model nvidia/nemotron-3-nano --quantization fp8 --tensor-parallel-size 8
```

---

## Error Output

**Startup warning (all 8 TP workers, tactic indices 19–20+):**

```
2026-03-11 16:06:48,177 - WARNING - autotuner.py:496 - flashinfer.jit:
[Autotuner]: Skipping tactic <MoERunner object> 19, due to failure while profiling:
[TensorRT-LLM][ERROR] Assertion failed: GPU lacks the shared memory resources
to run GroupedGEMM kernel
(moe_gemm_template_dispatch.h:175)
```

**Stack trace (key frames):**

```
dispatchGemmConfig<
  __nv_fp8_e4m3,           // input
  __nv_fp8_e4m3,           // weight
  __nv_bfloat16,           // output
  cutlass::arch::Sm89,     // ← Ada Lovelace target, running on Blackwell
  EpilogueOpDefault,
  GemmShape<16, 256, 128>, // tile shape
  GemmShape<16, 64, 128>   // warp shape
>
CutlassMoeFCRunner<__nv_fp8_e4m3, ...>::gemm1(...)
GemmProfilerBackend::runProfiler(...)
```

---

## Root Cause

The bug is in `moe_gemm_template_dispatch.h:793-801`:

```cpp
} else if (sm_ >= 90) {
    // For SM120+ pure FP8 MoE (not FP8 x FP4), redirect to SM89 (Ada) FP8 kernel implementations.
    if constexpr (use_fp8 && !use_wfp4afp8) {
        if (sm_ >= 120) {
            cutlass_kernels_oss::dispatchMoeGemmToCutlass<T, WeightType, ScaleBiasType,
                                                          cutlass::arch::Sm89,  // ← Ada on Blackwell
                                                          EpilogueTag>(inputs, multi_processor_count_);
            return;
        }
    }
```

This is **intentional** — no SM120-native TMA kernels exist for pure FP8 MoE (only the FP4/NVFP4 path has SM120 TMA coverage). When the SM89 kernels are dispatched on Blackwell, `GemmGrouped::maximum_active_blocks()` returns **0** for several SM89 tile shapes (e.g. `GemmShape<16, 256, 128>`) because Blackwell's shared memory architecture does not satisfy the SM89 occupancy requirements for those tile configs. The occupancy-zero assertion fires (line 175), which the autotuner catches and skips — but the surviving tactics are sub-optimal for Blackwell.

A secondary bug existed in the occupancy-query path itself: `genericMoeGemmKernelLauncher::call` was using `compute_occupancy_for_kernel<GemmKernel>()` for occupancy queries while the actual execution path used `GemmGrouped::maximum_active_blocks()`. This mismatch allowed configs with zero execution-path occupancy to pass the pre-check and only fail at profiling time.

---

## Workaround

Use NVFP4 instead of pure FP8 for MoE models on Blackwell, which takes the SM120-native TMA path and avoids the SM89 fallback entirely:

```bash
VLLM_USE_FLASHINFER_MOE_FP4=1  # use NVFP4 MoE path (Blackwell-native)
```

Or disable flashinfer MoE entirely to fall back to a non-CUTLASS kernel:

```bash
VLLM_USE_FLASHINFER_MOE_FP4=0
```

---

## Fix

PR #3032 (superseding #2764 by @dryu-nv) adds an occupancy pre-filter in `MoERunner.get_valid_tactics()`:

1. **C++ (`flashinfer_cutlass_fused_moe_binding.cu`):** Exposes `get_tactic_occupancy(tactic_id)` FFI binding using `GemmGrouped::maximum_active_blocks()` — the same occupancy path used at execution time — to safely probe each tactic without running it. Protected by `mMutex` to prevent races with concurrent `setTactic()`/`runMoe()` calls.

2. **C++ (`moe_gemm_template_dispatch.h`):** `queryOccupancyForConfig()` implementation routes through the full dispatch chain and catches unsupported configs via try/catch, returning 0 for any config that would fail.

3. **Python (`core.py`):** `get_valid_tactics()` pre-filters the tactic list by calling `get_tactic_occupancy()` on each candidate before passing them to the autotuner. Zero-occupancy tactics are dropped silently; if the FFI call fails unexpectedly, a `logger.warning` is emitted and the tactic is included as a safe fallback.

---

## Related

- #2764 — original fix by @dryu-nv (superseded by #3032)
- #2776 — NVFP4 MoE crash on GB10/SM121 (separate but related; fixed by #2798 / CUTLASS 4.4.2 upgrade)
- #2577 — NVFP4 mm_fp4 GEMM failures on SM120 (broader SM120 NVFP4 compatibility)

aniskumar-nv:
## Summary

Supersedes #2764 (by @dryu-nv). This PR carries the original fix forward and addresses all review feedback raised by CodeRabbit and Gemini Code Assist.

Closes #3031.

### Original fix (#2764)
On SM120 (Blackwell), pure FP8 MoE falls back to `cutlass::arch::Sm89` kernels. Some SM89 tile configs exceed Blackwell's shared memory, causing the autotuner to hit a shared memory assertion and skip those tactics, degrading MoE throughput. The fix adds an occupancy pre-filter in `MoERunner.get_valid_tactics()` so zero-occupancy tactics are dropped before autotuning begins.

### Changes in this PR (on top of #2764)

**`csrc/fused_moe/cutlass_backend/flashinfer_cutlass_fused_moe_binding.cu`**
- Add `std::lock_guard<std::mutex> lock(mMutex)` inside the `get_tactic_occupancy` lambda to protect access to `mAllProfiles` and `mKernelRunner` from concurrent `setTactic()`/`runMoe()` calls (thread safety issue flagged by CodeRabbit)

**`flashinfer/fused_moe/core.py`**
- Capture exception `e` and emit `logger.warning(...)` in the `get_occ(t)` failure path so FFI/C++ errors are no longer silently swallowed (logging issue flagged by Gemini)
- Return `[-1]` sentinel when `all_tactics` is empty (zero-tactic stage), preventing `get_valid_tactics()` from returning an empty list and violating its contract (edge case flagged by CodeRabbit)

## Test plan

- [ ] Run FP8 MoE autotuner on SM120 Blackwell (RTX PRO 6000 / GB10) — confirm no `GPU lacks shared memory` warnings
- [ ] Verify no regression on SM90 (H100) FP8 MoE autotuner
- [ ] Confirm dense NVFP4 models unaffected

## Summary by CodeRabbit

* **New Features**
  * Exposed an occupancy query so callers can determine whether a kernel tactic is supported on the device before execution.

* **Refactor**
  * Tactic selection now pre-validates candidates using occupancy results, preserves safe fallbacks when queries are missing or fail, logs warnings for query errors, and avoids returning an empty candidate list.

### DIFF
--- a/csrc/fused_moe/cutlass_backend/flashinfer_cutlass_fused_moe_binding.cu
+++ b/csrc/fused_moe/cutlass_backend/flashinfer_cutlass_fused_moe_binding.cu
@@ -725,6 +725,7 @@ class FusedMoeRunner : public tvm::ffi::ModuleObj {
       // Returns 0 if the tactic is not supported on the current device (e.g., SM89 tile
       // configs with insufficient shared memory when running on SM120 Blackwell).
       return Function::FromTyped([this](int64_t tactic_id) -> int64_t {
+        std::lock_guard<std::mutex> lock(mMutex);
         if (tactic_id < 0 || tactic_id >= static_cast<int64_t>(mAllProfiles.size())) {
           return 0;
         }
--- a/csrc/fused_moe/cutlass_backend/flashinfer_cutlass_fused_moe_binding.cu
+++ b/csrc/fused_moe/cutlass_backend/flashinfer_cutlass_fused_moe_binding.cu
@@ -720,6 +720,17 @@ class FusedMoeRunner : public tvm::ffi::ModuleObj {
       return Function::FromTyped([this]() -> int64_t { return mGemm1TacticCount; });
     } else if (name == "get_gemm2_tactic_count") {
       return Function::FromTyped([this]() -> int64_t { return mGemm2TacticCount; });
+    } else if (name == "get_tactic_occupancy") {
+      // Returns the max active blocks per SM for the given tactic index.
+      // Returns 0 if the tactic is not supported on the current device (e.g., SM89 tile
+      // configs with insufficient shared memory when running on SM120 Blackwell).
+      return Function::FromTyped([this](int64_t tactic_id) -> int64_t {
+        if (tactic_id < 0 || tactic_id >= static_cast<int64_t>(mAllProfiles.size())) {
+          return 0;
+        }
+        return static_cast<int64_t>(
+            mKernelRunner->queryOccupancyForConfig(mAllProfiles[tactic_id]));
+      });
     } else if (name == "run_moe") {
       return Function::FromTyped(
           [this](TensorView output, TensorView input, TensorView token_selected_experts,
--- a/csrc/nv_internal/tensorrt_llm/kernels/cutlass_kernels/include/moe_gemm_kernels.h
+++ b/csrc/nv_internal/tensorrt_llm/kernels/cutlass_kernels/include/moe_gemm_kernels.h
@@ -315,6 +315,12 @@ class MoeGemmRunner {
 
   [[nodiscard]] int getSM() const;
 
+  // Query the occupancy (max active blocks per SM) for a given GEMM configuration.
+  // Returns 0 if the configuration is not supported on the current device.
+  // This is useful for pre-filtering tactics that would fail at execution time due to
+  // insufficient shared memory resources (e.g., SM89 tile configs run on SM120 Blackwell).
+  int queryOccupancyForConfig(cutlass_extensions::CutlassGemmConfig const& config);
+
  private:
   template <typename EpilogueTag>
   void dispatchToArch(GroupedGemmInput<T, WeightType, ScaleBiasType, OutputType> inputs,
--- a/csrc/nv_internal/tensorrt_llm/kernels/cutlass_kernels/include/moe_kernels.h
+++ b/csrc/nv_internal/tensorrt_llm/kernels/cutlass_kernels/include/moe_kernels.h
@@ -455,6 +455,9 @@ class CutlassMoeFCRunnerInterface {
   virtual void setTactic(std::optional<cutlass_extensions::CutlassGemmConfig> gemm1_config,
                          std::optional<cutlass_extensions::CutlassGemmConfig> gemm2_config) = 0;
   virtual std::vector<cutlass_extensions::CutlassGemmConfig> getTactics(MoeGemmId gemm_id) = 0;
+  // Query occupancy for a GEMM config without executing the kernel.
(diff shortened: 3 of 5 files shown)


aleozlx (approved):

aniskumar-nv:
@aleozlx anything required from my side for the next steps? Want to get this merged.

PR #3062, opened 2026-04-14
fix: let FLASHINFER_CUBIN_DIR env var override flashinfer-cubin package path

### LINKED ISSUE #2834: Permission denied when running FP8 model: /usr/local/lib/python3.10/dist-packages/flashinfer_cubin/cubins/flashinfer
Hi,

I am trying to run inference for the model `Qwen3.5-397B-A17B-FP8` with vLLM, but I get the following error:

`Permission denied: '/usr/local/lib/python3.10/dist-packages/flashinfer_cubin/cubins/flashinfer'`

At the same time, inference for `Qwen3.5-397B-A17B` works successfully in the same environment.

How can I fix this problem?

### Environment
- OS: Ubuntu Jammy 22.04

### Installation steps
```bash
apt-get update && apt-get install -y \
  libxcb1 \
  build-essential \
  python3-dev

pip3 install --extra-index-url https://download.pytorch.org/whl/cu130 --extra-index-url https://pypi.org/simple \
  https://github.com/vllm-project/vllm/releases/download/v0.17.1/vllm-0.17.1+cu130-cp38-abi3-manylinux_2_35_x86_64.whl \

pip3 install --upgrade --index-url https://pypi.org/simple \
  flashinfer-python flashinfer-cubin

pip3 install --upgrade --index-url https://flashinfer.ai/whl/cu130 \
  flashinfer-jit-cache
```

### Error log
```text
(EngineCore_DP0 pid=488) {"message": "Worker proc VllmWorker-6 died unexpectedly, shutting down executor."}
(EngineCore_DP0 pid=488) Process EngineCore_DP0:
(EngineCore_DP0 pid=488) Traceback (most recent call last):
(EngineCore_DP0 pid=488)   File "/usr/lib/python3.10/multiprocessing/process.py", line 314, in _bootstrap
(EngineCore_DP0 pid=488)     self.run()
(EngineCore_DP0 pid=488)   File "/usr/lib/python3.10/multiprocessing/process.py", line 108, in run
(EngineCore_DP0 pid=488)     self._target(*self._args, **self._kwargs)
(EngineCore_DP0 pid=488)   File "/usr/local/lib/python3.10/dist-packages/vllm/v1/engine/core.py", line 1104, in run_engine_core
(EngineCore_DP0 pid=488)     raise e
(EngineCore_DP0 pid=488)   File "/usr/local/lib/python3.10/dist-packages/vllm/v1/engine/core.py", line 1090, in run_engine_core
(EngineCore_DP0 pid=488)     engine_core = EngineCoreProc(*args, engine_index=dp_rank, **kwargs)
(EngineCore_DP0 pid=488)   File "/usr/local/lib/python3.10/dist-packages/vllm/tracing/otel.py", line 178, in sync_wrapper
(EngineCore_DP0 pid=488)     return func(*args, **kwargs)
(EngineCore_DP0 pid=488)   File "/usr/local/lib/python3.10/dist-packages/vllm/v1/engine/core.py", line 834, in __init__
(EngineCore_DP0 pid=488)     super().__init__(
(EngineCore_DP0 pid=488)   File "/usr/local/lib/python3.10/dist-packages/vllm/v1/engine/core.py", line 120, in __init__
(EngineCore_DP0 pid=488)     num_gpu_blocks, num_cpu_blocks, kv_cache_config = self._initialize_kv_caches(
(EngineCore_DP0 pid=488)   File "/usr/local/lib/python3.10/dist-packages/vllm/tracing/otel.py", line 178, in sync_wrapper
(EngineCore_DP0 pid=488)     return func(*args, **kwargs)
(EngineCore_DP0 pid=488)   File "/usr/local/lib/python3.10/dist-packages/vllm/v1/engine/core.py", line 252, in _initialize_kv_caches
(EngineCore_DP0 pid=488)     available_gpu_memory = self.model_executor.determine_available_memory()
(EngineCore_DP0 pid=488)   File "/usr/local/lib/python3.10/dist-packages/vllm/v1/executor/abstract.py", line 136, in determine_available_memory
(EngineCore_DP0 pid=488)     return self.collective_rpc("determine_available_memory")
(EngineCore_DP0 pid=488)   File "/usr/local/lib/python3.10/dist-packages/vllm/v1/executor/multiproc_executor.py", line 389, in collective_rpc
(EngineCore_DP0 pid=488)     return aggregate(get_response())
(EngineCore_DP0 pid=488)   File "/usr/local/lib/python3.10/dist-packages/vllm/v1/executor/multiproc_executor.py", line 372, in get_response
(EngineCore_DP0 pid=488)     raise RuntimeError(
(EngineCore_DP0 pid=488) RuntimeError: Worker failed with error '[Errno 13] Permission denied: '/usr/local/lib/python3.10/dist-packages/flashinfer_cubin/cubins/flashinfer'', please check the stack trace above for the root cause
```

### Diagnostics
I checked the path mentioned in the error, and it does not exist:

```text
$ ls /usr/local/lib/python3.10/dist-packages/flashinfer_cubin/cubins/flashinfer

ls: cannot access '/usr/local/lib/python3.10/dist-packages/flashinfer_cubin/cubins/flashinfer': No such file or directory

$ ls /usr/local/lib/python3.10/dist-packages/flashinfer_cubin/cubins/

1fddc48b7b48af33914d040051b3e2ee9ba4701e  a72d85b019dc125b9f711300cb989430f762f5a6  b55211623be7f5697c5262ffd8361fc06c147bc9  f1ed60e5666a7620683a8c34a41c850a25029b35
```

Any guidance would be appreciated.

### LINKED ISSUE #2976: FLASHINFER_CUBIN_DIR env var ignored when flashinfer-cubin package is installed — breaks non-root containers
## Problem

When the `flashinfer-cubin` package is installed, `FLASHINFER_CUBIN_DIR` environment variable is silently ignored. This makes it impossible to redirect cubin storage to a writable directory in containerized environments where the process runs as a non-root user with an arbitrary UID/GID.

This causes `PermissionError: [Errno 13]` crashes on startup in Kubernetes/OpenShift deployments where security contexts enforce non-root execution with arbitrary UIDs that have no write access to the site-packages directory.

Related: #2834 (same symptom, no root cause identified)

## Root Cause

In [`flashinfer/jit/env.py`](https://github.com/flashinfer-ai/flashinfer/blob/main/flashinfer/jit/env.py), `_get_cubin_dir()` has this priority order:

```python
def _get_cubin_dir():
    # 1. flashinfer-cubin package if installed  ← ALWAYS WINS
    if has_flashinfer_cubin():
        import flashinfer_cubin
        return pathlib.Path(flashinfer_cubin.get_cubin_dir())

    # 2. Environment variable  ← only checked if package NOT installed
    env_dir = os.getenv("FLASHINFER_CUBIN_DIR")
    if env_dir:
        return pathlib.Path(env_dir)

    # 3. Default cache directory
    return FLASHINFER_CACHE_DIR / "cubins"
```

When `flashinfer-cubin` is installed (which it is in all pre-built container images), the env var check is never reached. The returned path points inside `site-packages/flashinfer_cubin/cubins/`, which is root-owned.

Then in [`flashinfer/jit/cubin_loader.py`](https://github.com/flashinfer-ai/flashinfer/blob/main/flashinfer/jit/cubin_loader.py), `get_artifact()` calls:

```python
os.makedirs(os.path.dirname(local_path), exist_ok=True)
```

This attempts to create subdirectories (e.g., `cubins/flashinfer/`) inside the package's own installation directory at runtime. When the process runs as a non-root user (e.g., UID 1003020000 on OpenShift), this fails with `PermissionError`.

## Reproduction

```bash
# Install flashinfer with pre-built cubins
pip install flashinfer-python flashinfer-cubin

# Try to redirect cubins to a writable location — this is silently ignored
export FLASHINFER_CUBIN_DIR=/tmp/flashinfer_cubins

# Run as non-root user (simulating OpenShift/k8s)
python -c "
from flashinfer.jit.env import FLASHINFER_CUBIN_DIR
print(f'FLASHINFER_CUBIN_DIR = {FLASHINFER_CUBIN_DIR}')
# Prints: site-packages/flashinfer_cubin/cubins  (NOT /tmp/flashinfer_cubins)
"
```

## Impact

- **vLLM** ships with `flashinfer-cubin` pre-installed in container images
- **Kubernetes/OpenShift** enforces non-root execution with arbitrary UID/GID via `runAsUser`/`runAsGroup` security contexts
- There is **no user-configurable workaround** — the env var that should fix this is ignored
- Container builders must resort to `chmod -R 777` on the package directory, which is a security anti-pattern

## Proposed Fix

Change `_get_cubin_dir()` to check the environment variable **first**, before falling back to the package:

```python
def _get_cubin_dir():
    # 1. Environment variable takes priority (allows override for non-root containers)
    env_dir = os.getenv("FLASHINFER_CUBIN_DIR")
    if env_dir:
        return pathlib.Path(env_dir)

    # 2. flashinfer-cubin package if installed
    if has_flashinfer_cubin():
        import flashinfer_cubin
        # version check ...
        return pathlib.Path(flashinfer_cubin.get_cubin_dir())

    # 3. Default cache directory
    return FLASHINFER_CACHE_DIR / "cubins"
```

This preserves the existing default behavior (package wins when no env var is set) while giving users an escape hatch for restricted environments.

An alternative would be to also seed the env-var directory from the package contents on first use, so pre-built cubins are still available without re-downloading.

## Environment

- FlashInfer 0.6.6
- vLLM 0.18.0
- Container runtime: OpenShift 4.x with arbitrary UID/GID security context
- Host OS: RHEL 9 / Ubuntu 22.04

ianliuy:
## What's broken?

When `flashinfer-cubin` is installed (all pre-built container images), setting `FLASHINFER_CUBIN_DIR` to redirect cubin storage to a writable directory is silently ignored. Non-root containers crash with `PermissionError` on startup.

## Who is affected?

All non-root K8s/OpenShift deployments using pre-built images with `flashinfer-cubin` (e.g., vLLM containers). No workaround exists.

## Why does it happen?

`_get_cubin_dir()` checks the `flashinfer-cubin` package **before** the env var. When the package is installed, the env var is never reached. Introduced in PR #1718 (intentional priority, but didn't consider non-root containers).

Every other env var in `env.py` follows "env var overrides automatic discovery." This was the sole exception.

## How did we fix it?

Swap priority: env var -> package -> default cache. Pure code-block reorder (~8 lines moved + docstring/comments updated). Zero new logic.

**Before:**
```python
def _get_cubin_dir():
    if has_flashinfer_cubin():          # always wins
        return package_path
    env_dir = os.getenv("FLASHINFER_CUBIN_DIR")  # unreachable
    ...
```

**After:**
```python
def _get_cubin_dir():
    env_dir = os.getenv("FLASHINFER_CUBIN_DIR")  # user override first
    if env_dir:
        return pathlib.Path(env_dir)
    if has_flashinfer_cubin():          # package fallback
        return package_path
    ...
```

Default behavior (no env var set) is **unchanged**.

> **Note:** If you previously had a stale `FLASHINFER_CUBIN_DIR` that was silently ignored, it will now take effect. Unset it if unneeded.

## How do we know it works?

Added `tests/test_env.py` with 4 regression tests covering the full priority matrix:
- env var + package -> env var wins (the bug)
- package only -> package wins
- env var only -> env var wins
- neither -> default cache dir

All 4 pass. Pre-commit hooks (ruff, mypy, format) all pass.

Fixes #2976
Related: #2834

@aleozlx @yzh119

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CUBIN directory resolution to honor `FLASHINFER_CUBIN_DIR` first, with fallback to the installed `flashinfer_cubin` location when unset.
  * When `FLASHINFER_CUBIN_DIR` is set alongside the installed package, a warning is emitted to clarify which source is used.

* **Tests**
  * Added regression tests covering CUBIN directory selection priority across environment and package availability scenarios.

### DIFF
--- a/flashinfer/jit/env.py
+++ b/flashinfer/jit/env.py
@@ -59,11 +59,16 @@ _package_root: pathlib.Path = pathlib.Path(__file__).resolve().parents[1]
 def _get_cubin_dir():
     """
     Get the cubin directory path with the following priority:
-    1. flashinfer-cubin package if installed
-    2. Environment variable FLASHINFER_CUBIN_DIR
+    1. Environment variable FLASHINFER_CUBIN_DIR
+    2. flashinfer-cubin package if installed
     3. Default cache directory
     """
-    # First check if flashinfer-cubin package is installed
+    # First check environment variable
+    env_dir = os.getenv("FLASHINFER_CUBIN_DIR")
+    if env_dir:
+        return pathlib.Path(env_dir)
+
+    # Then check if flashinfer-cubin package is installed
     if has_flashinfer_cubin():
         import flashinfer_cubin
 
@@ -81,11 +86,6 @@ def _get_cubin_dir():
             )
 
         return pathlib.Path(flashinfer_cubin.get_cubin_dir())
-
-    # Then check environment variable
-    env_dir = os.getenv("FLASHINFER_CUBIN_DIR")
-    if env_dir:
-        return pathlib.Path(env_dir)
 
     # Fall back to default cache directory
     return FLASHINFER_CACHE_DIR / "cubins"
--- a/flashinfer/jit/env.py
+++ b/flashinfer/jit/env.py
@@ -18,10 +18,17 @@ limitations under the License.
 # Do "from .jit import env as jit_env" and use "jit_env.xxx" instead.
 # This helps AOT script to override envs.
 
+import logging
 import os
 import pathlib
 from ..compilation_context import CompilationContext
 from ..version import __version__ as flashinfer_version
+
+# NOTE: use stdlib logging (namespaced under flashinfer.jit) instead of the
+# FlashInferJITLogger from core.py -- core.py imports this module, so importing
+# it here would create a circular import. A plain warning is fine because
+# _get_cubin_dir() only runs once, at module import.
+logger = logging.getLogger("flashinfer.jit")
 
 
 def has_flashinfer_jit_cache() -> bool:
@@ -66,6 +73,12 @@ def _get_cubin_dir():
     # First check environment variable
     env_dir = os.getenv("FLASHINFER_CUBIN_DIR")
     if env_dir:
+        if has_flashinfer_cubin():
+            logger.warning(
+                "FLASHINFER_CUBIN_DIR=%s overrides the installed flashinfer-cubin "
+                "package; cubins will be read from that path instead of the package.",
+                env_dir,
+            )
         return pathlib.Path(env_dir)
 
     # Then check if flashinfer-cubin package is installed
--- a/tests/test_env.py
+++ b/tests/test_env.py
@@ -1,0 +1,103 @@
+"""Regression tests for _get_cubin_dir() priority — issue #2976.
+
+env.py imports CompilationContext (CUDA deps), so we load it in isolation
+with lightweight stubs to keep tests runnable without a GPU.
+
+    python -m pytest tests/test_env.py -v --noconftest
+"""
+
+import importlib.util
+import pathlib
+import sys
+import types
+
+_REPO_ROOT = pathlib.Path(__file__).resolve().parents[1]
+
+
+def _load_env_module():
+    """Load flashinfer.jit.env with minimal stubs (no CUDA required)."""
+    stubs = {
+        "flashinfer": types.ModuleType("flashinfer"),
+        "flashinfer.jit": types.ModuleType("flashinfer.jit"),
+        "flashinfer.version": types.ModuleType("flashinfer.version"),
+        "flashinfer.compilation_context": types.ModuleType(
+            "flashinfer.compilation_context"
+        ),
+    }
+    stubs["flashinfer"].__path__ = [str(_REPO_ROOT / "flashinfer")]
+    stubs["flashinfer.jit"].__path__ = [str(_REPO_ROOT / "flashinfer" / "jit")]
+    stubs["flashinfer.version"].__version__ = "0.0.0+test"
+    stubs["flashinfer.version"].__git_version__ = "test"
+
+    class _Stub:
+        def __init__(self):
+            self.TARGET_CUDA_ARCHS = set()
+
+    stubs["flashinfer.compilation_context"].CompilationContext = _Stub
+
+    saved = {k: sys.modules.get(k) for k in stubs}
+    sys.modules.update(stubs)
+    try:
+        spec = importlib.util.spec_from_file_location(
+            "flashinfer.jit.env", str(_REPO_ROOT / "flashinfer" / "jit" / "env.py")
+        )
+        mod = importlib.util.module_from_spec(spec)
+        sys.modules["flashinfer.jit.env"] = mod
+        spec.loader.exec_module(mod)
+        return mod
+    finally:
+        for k, v in saved.items():
+            if v is None:
+                sys.modules.pop(k, None)
+            else:
+                sys.modules[k] = v
+
+
+_env = _load_env_module()
+
+
+def _fake_cubin_pkg(path):
+    """Return a stub ``flashinfer_cubin`` module pointing at *path*."""
+    m = types.ModuleType("flashinfer_cubin")
+    m.__version__ = "0.0.0+test"
+    m.get_cubin_dir = lambda: path
+    return m
+
+
+# -- priority tests (regression for #2976) ----------------------------------
+
+
+def test_env_var_overrides_package(monkeypatch, tmp_path):
+    """FLASHINFER_CUBIN_DIR must take priority over the installed package."""
+    env_dir = str(tmp_path / "env_cubins")
+    pkg_dir = str(tmp_path / "pkg_cubins")
+    monkeypatch.setenv("FLASHINFER_CUBIN_DIR", env_dir)
+    monkeypatch.setenv("FLASHINFER_DISABLE_VERSION_CHECK", "1")
+    monkeypatch.setattr(_env, "has_flashinfer_cubin", lambda: True)
+    monkeypatch.setitem(sys.modules, "flashinfer_cubin", _fake_cubin_pkg(pkg_dir))
+    assert _env._get_cubin_dir() == pathlib.Path(env_dir)
+
+
+def test_package_used_when_no_env_var(monkeypatch, tmp_path):
+    """Without the env var, the package path should be returned."""
+    pkg_dir = str(tmp_path / "pkg_cubins")
+    monkeypatch.delenv("FLASHINFER_CUBIN_DIR", raising=False)
+    monkeypatch.setenv("FLASHINFER_DISABLE_VERSION_CHECK", "1")
+    monkeypatch.setattr(_env, "has_flashinfer_cubin", lambda: True)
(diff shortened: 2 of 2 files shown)


yankeexe:
Would be really great to have this merged because this is the default expected behavior.

almersawi:
Can we merge this MR?

flashinfer/jit/env.py
```diff
@@ -59,11 +59,16 @@ def has_flashinfer_cubin() -> bool:
...
     Get the cubin directory path with the following priority:
-    1. flashinfer-cubin package if installed
-    2. Environment variable FLASHINFER_CUBIN_DIR
+    1. Environment variable FLASHINFER_CUBIN_DIR
+    2. flashinfer-cubin package if installed
     3. Default cache directory
     """
-    # First check if flashinfer-cubin package is installed
+    # First check environment variable
+    env_dir = os.getenv("FLASHINFER_CUBIN_DIR")
```
bkryu, line 74: Can we add a one-time logger.warning when both the env var is set and has_flashinfer_cubin() is true, so the bypass is visible?

This way there will be less surprises.

tests/test_env.py
```diff
@@ -0,0 +1,103 @@
...
+    stubs["flashinfer.compilation_context"].CompilationContext = _Stub
+
+    saved = {k: sys.modules.get(k) for k in stubs}
+    sys.modules.update(stubs)
+    try:
+        spec = importlib.util.spec_from_file_location(
+            "flashinfer.jit.env", str(_REPO_ROOT / "flashinfer" / "jit" / "env.py")
+        )
+        mod = importlib.util.module_from_spec(spec)
+        sys.modules["flashinfer.jit.env"] = mod
```
bkryu, line 45: I believe this entry is never restored: the `finally` block puts back the four stubs keys, but `flashinfer.jit.env` keeps pointing at this stub-loaded copy for the rest of the pytest session. 

Fix is two lines — include "flashinfer.jit.env" in the saved dict and restore it in the same finally:
saved = {k: sys.modules.get(k) for k in (*stubs, "flashinfer.jit.env")}

tests/test_env.py
```diff
@@ -0,0 +1,103 @@
...
+        return mod
+    finally:
+        for k, v in saved.items():
+            if v is None:
+                sys.modules.pop(k, None)
+            else:
+                sys.modules[k] = v
+
+
+_env = _load_env_module()
```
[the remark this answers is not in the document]
bkryu: Can you address the coderabbit comment?

bkryu (commented):
Thanks @ianliuy for opening the PR.

I agree that explicit user configuration beating auto-discovery is the right convention and the diff is simple and simply correct.

Left a few minor comments, especially for the test file

flashinfer/jit/env.py
jimmyzho: See additions

jimmyzho:
@bkryu Just applied your suggestions - it uses a plain `logger.warning`, not a `warning_once`, to avoid pulling the logger out of `core.py` to avoid a circular import. We can defer this to a separate PR - for now, in production, _get_cubin_dir() should only called once.

bkryu (approved):
Approved Pending CI. Thanks @jimmyzho and @ianliuy

yankeexe:
Finally! 😮‍💨

PR #3063, opened 2026-04-14
fix(autotuner): hybrid bucket spacing and cache-key fix for fused MoE

StudyingShao:
## 📌 Description

  Power-of-2 tuning buckets create exponentially growing gaps at large `num_tokens` values (e.g. 1024 gap between bucket
  1024 and 2048). For MoE workloads, `avg_tokens_per_expert` can cross multiple tile selection boundaries within a single
  gap, causing non-power-of-2 inputs (like 1536) to be mapped to a much smaller bucket and receive a kernel optimized for
  the wrong workload size — sometimes running slower than larger batch sizes.

  This PR replaces pure power-of-2 buckets with a four-phase hybrid spacing scheme: power-of-2 up to 256, linear step 256
  up to 2048, linear step 512 up to 4096, then power-of-2 again. This keeps bucket count manageable (21 vs 14 for
  max=8192) while ensuring adjacent buckets never span a tile selection boundary.

  This PR also fixes a bug in `AutoTuner._find_nearest_profile` where only the first tensor in a multi-tensor
  `DynamicTensorSpec` had its dimension mapped to the nearest bucket. For ops like `trtllm_fp8_per_tensor_scale_moe` that
  declare 5 input tensors sharing the same `num_tokens` dimension, the remaining 4 tensors retained their original shape —
   producing a different cache key at inference time than what was stored during profiling. This caused a silent cache
  miss and fallback to the default tactic for any non-bucket-boundary token count (e.g. bs=384 mapped to bucket 512 but
  always fell back to TileN=8).

  ## 🔍 Related Issues

  N/A

  ## 🚀 Pull Request Checklist

  Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are
  complete.

  ### ✅ Pre-commit Checks

  ## 🧪 Tests

  ## Reviewer Notes

  - The bucket spacing issue was observed with `trtllm-gen` fused MoE FP8-Per-Tensor (num_experts=24, top_k=8):
  `num_tokens=1536` was mapped to bucket 1024, selecting a small-tile kernel (`t128x32x256`) instead of the large-tile
  kernel (`t128x128x128`) that 2048 gets — making 1536 slower than 2048.
  - The cache-miss bug compounded this: even for the hybrid-bucket case, bs=384 mapped correctly to bucket 512 but always
  cache-missed because `_find_nearest_profile` only bucketed the output buffer tensor, leaving the other 4 input tensors
  at their original 384-token shapes. After fixing both, bs=384 goes from 1.00x to 2.40x speedup (Fp8-Per-Tensor,
  num_experts=192, top_k=8, hidden=4096, intermediate=1536).
  - The hybrid bucket fix applies to CUTLASS MoE, CuTE DSL MoE, GEMM, and low-latency GEMM paths for consistency, though
  only the trtllm-gen path has been benchmarked.
  - The cache-miss fix applies to any op whose `DynamicTensorSpec` lists more than one input tensor sharing a dynamic
  dimension.
  - Phase thresholds (256/2048/4096) and step sizes (256/512) are tuned for current max tile sizes (128/256). If future
  kernels introduce larger tiles, these may need adjustment.
  - Old functions `get_power_of_2_num_tokens_buckets` and `get_last_power_of_2_num_tokens_buckets` are removed;
  `last_positive_power_of_2` is retained as it's still used by autotuner unit tests.


StudyingShao:
Closed. Move to #3115 .

PR #3076, opened 2026-04-15
fix: fix logical bug reported in #2768

### LINKED ISSUE #2768: [bug] cumsumlog calculation in gated delta net.
https://github.com/flashinfer-ai/flashinfer/blob/f07faf964a4d4aecba92254161a4d4ae3079cf4c/include/flashinfer/flat/hopper/collective/flat_collective_tma_warpspecialized_delta_rule.hpp#L325        

The cumsumlog number accumulation across iter might be wrong? The `sum` should be accumulated.

guangyunh-nv:
Fix #2768

This is not trigger as size(frag) is 2 in the kernel.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected computation logic in tensor acceleration backend to improve computational accuracy.


guangyunh-nv:
Abandoned, code ported to cutedsl.

PR #3091, opened 2026-04-16
Vendor CCCL v3.3.2 from GitHub instead of relying on CTK-bundled copy

kahyunnam:
## 📌 Description

Pin CCCL (CUB, libcudacxx, Thrust) to a specific release tag as a git submodule under `3rdparty/cccl`, replacing the implicit dependency on whatever version ships with the user's CUDA Toolkit. This enables the CCCL team to land improvements (e.g., TopK, DeviceTransform, fast_mod_div) independently of CTK releases, and lets FlashInfer adopt new CCCL features immediately.

### Vendoring infrastructure (`c2cf0708`)
- Add `3rdparty/cccl` submodule at CCCL v3.3.2 (maps to CTK 13.2)
- Wire vendored CCCL into JIT include paths using `-I` (not `-isystem`) so it takes precedence over CTK headers, per [CCCL guidelines](https://github.com/NVIDIA/cccl/issues/527)
- Remove `$cuda_home/include/cccl` from system includes
- Package CCCL headers (cub, libcudacxx, thrust) into the wheel via `pyproject.toml` and `build_backend.py`
- Update `modal_runner.py` fallback clone for CCCL
- Remove dead `#if CUDA_VERSION` guards for `cub::Max`/`cub::Min` which no longer exist in CCCL 3.x; unconditionally use `cuda::maximum`/`cuda::minimum`

### Adopt `cub::DeviceTransform` for LSE computation (`9fda09f2`)
- Replace `ComputeLSEFromMDKernel` (hand-rolled element-wise kernel with manual PDL asm, launch config, bounds checking) with a single `cub::DeviceTransform::Transform` call
- `DeviceTransform` automatically provides PDL on SM90+, vectorized loads, software prefetch, auto-tuned occupancy, and bulk copy (TMA) on Hopper+
- Uses a named functor (`MDToLSE`) instead of a lambda to work around an nvcc name-mangling bug
- `log2f` replaces the PTX-only `math::ptx_log2` since the functor must be `__host__ __device__`; with `-use_fast_math`, nvcc emits the same `lg2.approx.ftz.f32` instruction on device

### Adopt `cuda::fast_mod_div` for fast integer division (`ac916940`, `ddd77d1c`)
- Replace the `FastModDivInt32` polyfill in `kernelParams.h` with a type alias to `cuda::fast_mod_div<int32_t>` — the struct it was explicitly polyfilling (binary layout is identical)
- Reimplement `flashinfer::uint_fastdiv` in `fastdiv.cuh` as a thin API-compatible wrapper around `cuda::fast_mod_div<uint32_t>`, preserving the default constructor, implicit conversions, and `divmod()` method used by ~30 call sites in the attention, page, MLA, and RoPE kernels

### Not changed
- `trtllm::dev::IntFastDiv` (MoE routing) — vendored TRT-LLM code with different binary layout and negative divisor support that `cuda::fast_mod_div` doesn't provide
- NV-internal / cuDNN / CUTLASS fast-divmod implementations — external codebases, best left alone

## 📊 Performance
Benchmarked `top_k` (BF16, non-deterministic, random input) across 162 configurations (batch ∈ {1, 16, 64, 256, 2048, 4096}, seq_len ∈ {256..524288}, k ∈ {256..4096}) on an B200 with CUDA 13.0.
**CCCL 3.3.2 (vendored) vs CCCL 3.0 (CTK-bundled):**
- Mean speedup: 1.00x | Median: 1.00x
- No regressions or improvements beyond measurement noise (±2% at small problem sizes)
- Scatter plot shows all 162 points on the y=x diagonal
This is expected — CCCL 3.0 → 3.3.2 is only 3 minor versions apart, and FlashInfer's TopK uses block-level CUB primitives which are stable across minor releases. Users on **older CTKs** (e.g., 12.6 shipping CCCL 2.5) would see larger gains from the version jump. The primary value of this PR is infrastructure: FlashInfer can now adopt new CCCL features (TopK improvements, DeviceTransform, fast_mod_div, segmented scan) independently of CTK releases, and the CCCL team can land targeted optimizations by bumping the submodule tag.

image

## 🔍 Related Issues

https://github.com/flashinfer-ai/flashinfer/issues/3096

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

- [x] `test_trtllm_gen_prefill` passes on SM100 (exercises `DeviceTransform` LSE, `FastModDivInt32`, and `uint_fastdiv` code paths)
- [x] `test_batch_prefill_with_paged_kv_cache` passes on SM80+ (exercises `uint_fastdiv` in core attention)
- [x] All pre-commit hooks pass (clang-format, ruff, mypy)

## Reviewer Notes

- The CCCL submodule adds to wheel size. Only the header directories needed at JIT time (`cub/cub/`, `libcudacxx/include/`, `thrust/thrust/`) are packaged — not the full CCCL repo.
- `cuda::fast_mod_div` has a deleted default constructor, so `uint_fastdiv` is kept as a thin wrapper to preserve the existing API contract (default-constructible, implicit conversions, `.divmod()` method) without touching ~30 call sites.
- The `DeviceTransform` change uses a named functor instead of a lambda due to an nvcc name-mangling bug with `__host__ __device__` lambdas in inline functions used as kernel template arguments.
- Bumping the CCCL submodule to a newer tag in the future is a one-line `git submodule update` — no code changes needed unless new APIs are deprecated.

## Summary by CodeRabbit

* **Chores**
  * Vendored NVIDIA CCCL added and packaged with distributed headers.
  * Build/packaging updated to include CCCL header trees.
  * Runtime tooling now ensures CCCL is present (clones vendored sources if missing).
  * CUDA-related kernel and math handling unified for improved compatibility and correctness.

### DIFF
--- a/include/flashinfer/trtllm/fmha/lse.cuh
+++ b/include/flashinfer/trtllm/fmha/lse.cuh
@@ -18,43 +18,23 @@ limitations under the License.
 
 #include <cuda.h>
 
+#include <cmath>
+#include <cub/device/device_transform.cuh>
+
 #include "../../math.cuh"
 #include "../../utils.cuh"
 
 namespace flashinfer {
 
-__global__ void ComputeLSEFromMDKernel(float2* __restrict__ md, float* __restrict__ lse, int n) {
-  int elem_idx = blockIdx.x * blockDim.x + threadIdx.x;
-  if (elem_idx >= n) return;
-#if (__CUDACC_VER_MAJOR__ >= 12 && defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 900))
-  asm volatile("griddepcontrol.wait;");
-#endif
-  float2 md_elem = md[elem_idx];
-  float m = md_elem.x;
-  float d = md_elem.y;
-  lse[elem_idx] = math::log2e * m + math::ptx_log2(d);
-#if (__CUDACC_VER_MAJOR__ >= 12 && defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 900))
-  asm volatile("griddepcontrol.launch_dependents;");
-#endif
-}
+struct MDToLSE {
+  __host__ __device__ float operator()(float2 md_elem) const {
+    return math::log2e * md_elem.x + log2f(md_elem.y);
+  }
+};
 
-inline cudaError_t ComputeLSEFromMD(float2* md, float* lse, int n, bool launch_with_pdl,
+inline cudaError_t ComputeLSEFromMD(float2* md, float* lse, int n, bool /*launch_with_pdl*/,
                                     cudaStream_t stream) {
-  int num_threads = std::min(1024, UpPowerOfTwo(n));
-  int num_blocks = ceil_div(n, num_threads);
-  cudaLaunchConfig_t config;
-  config.gridDim = num_blocks;
-  config.blockDim = num_threads;
-  config.dynamicSmemBytes = 0;
-  config.stream = stream;
-  cudaLaunchAttribute attrs[1];
-  attrs[0].id = cudaLaunchAttributeProgrammaticStreamSerialization;
-  attrs[0].val.programmaticStreamSerializationAllowed = launch_with_pdl;
-  config.numAttrs = 1;
-  config.attrs = attrs;
-
-  FLASHINFER_CUDA_CALL(cudaLaunchKernelEx(&config, ComputeLSEFromMDKernel, md, lse, n));
-  return cudaSuccess;
+  return cub::DeviceTransform::Transform(md, lse, n, MDToLSE{}, stream);
 }
 
 };  // namespace flashinfer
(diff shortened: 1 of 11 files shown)


include/flashinfer/trtllm/fmha/lse.cuh
```diff
@@ -18,43 +18,23 @@ limitations under the License.
...
-#endif
-}
+struct MDToLSE {
+  __host__ __device__ float operator()(float2 md_elem) const {
+    return math::log2e * md_elem.x + log2f(md_elem.y);
+  }
+};
 
-inline cudaError_t ComputeLSEFromMD(float2* md, float* lse, int n, bool launch_with_pdl,
+inline cudaError_t ComputeLSEFromMD(float2* md, float* lse, int n, bool /*launch_with_pdl*/,
```
kahyunnam, line 35: Note for reviewers: launch_with_pdl is unused — DeviceTransform enables PDL unconditionally on SM90+ via its internal launcher. On pre-Hopper GPUs, the PDL instructions compile to no-ops. This means callers that pass false will still get PDL when the GPU supports it, which is probably harmless (PDL is a performance hint, not semantic).

nv-yunzheq (approved):
Makes sense to me. Code looks good to me. Please wait for unit test clean to get it merged

PR #3097, opened 2026-04-17
Support NVFP4 KV for prefill and batch attention kernels

Tom-Zheng:
## 📌 Description

This MR supports NVFP4 KV input for batch prefill and batch attention kernels. It widely supports all arch (SM80+).

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * NVFP4 (packed FP4) KV-cache support across prefill, attention, and decode with optional per-block scale-factor inputs and expanded accepted formats for scale tensors.
  * New CPU dequantization fallback for older GPUs and broader FP4 vectorized handling.

* **Bug Fixes**
  * Runtime guard disables an incompatible backend for packed-KV (uint8) cases.

* **Tests**
  * Added extensive NVFP4 unit tests and test helpers for prefill, attention, and decode.

### DIFF
--- a/flashinfer/decode.py
+++ b/flashinfer/decode.py
@@ -444,9 +444,9 @@ def single_decode_with_kv_cache(
     q_scale : Optional[float]
         The calibration scale of query for fp8 input, if not provided, will be set to ``1.0``.
     k_scale : Optional[float]
-        The calibration scale of key for fp8 or nvfp4 input, if not provided, will be set to ``1.0``.
+        The calibration scale of key for fp8 input, if not provided, will be set to ``1.0``.
     v_scale : Optional[float]
-        The calibration scale of value for fp8 or nvfp4 input, if not provided, will be set to ``1.0``.
+        The calibration scale of value for fp8 input, if not provided, will be set to ``1.0``.
     window_left : int
         The left (inclusive) window size for the attention window, when set to ``-1``, the window
         size will be set to the full length of the sequence. Defaults to ``-1``.
@@ -1297,6 +1297,8 @@ class BatchDecodeWithPagedKVCacheWrapper:
               where dim 1 holds k (index 0) and v (index 1) scales.
 
             Both tensors have dtype ``torch.float8_e4m3fn``.
+
+            Currently, NVFP4 KV supports `fa2` and `trtllm-gen` backend.
         Returns
         -------
         Union[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]]
@@ -2293,9 +2295,7 @@ def trtllm_batch_decode_with_kv_cache(
     max_q_len: Optional[int] = None,
     cum_seq_lens_q: Optional[torch.Tensor] = None,
     skip_softmax_threshold_scale_factor: Optional[float] = None,
-    kv_cache_sf: Optional[
-        Union[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]]
-    ] = None,
+    kv_cache_sf: Optional[Tuple[torch.Tensor, torch.Tensor]] = None,
     uses_shared_paged_kv_idx: bool = True,
 ) -> Union[torch.Tensor, FP4Tensor]:
     """
@@ -2305,15 +2305,11 @@ def trtllm_batch_decode_with_kv_cache(
         query tensor with shape [num_tokens, num_heads, head_dim], num_tokens = total query tokens in the batch.
 
     kv_cache : Union[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]]
-        The paged KV-Cache stored as a tuple of tensors or a single tensor:
-
-        * a tuple ``(k_cache, v_cache)`` of 4-D tensors, each with shape
-          ``[num_pages, num_kv_heads, page_size, head_dim]`` if :attr:`kv_layout` is ``HND``,
-          or ``[num_pages, page_size, num_kv_heads, head_dim]`` if :attr:`kv_layout` is ``NHD``.
-        * a single 5-D tensor with shape
-          ``[num_pages, 2, num_kv_heads, page_size, head_dim]`` if :attr:`kv_layout` is ``HND``,
-          or ``[num_pages, 2, page_size, num_kv_heads, head_dim]`` if :attr:`kv_layout` is ``NHD``,
-          where dim 1 holds k (index 0) and v (index 1).
+        If kv_cache is a single tensor, it should be a tensor with shape [num_pages, 1 or 2, num_kv_heads, page_size, head_dim] if :attr:`kv_layout` is ``HND``,
+        or [num_pages, 1 or 2, page_size, num_kv_heads, head_dim] if :attr:`kv_layout` is ``NHD``.
+        If kv_cache is a tuple of two tensors, it should be a tuple of two tensors with shape [num_pages, num_kv_heads, page_size, head_dim] if :attr:`kv_layout` is ``HND``,
+        or [num_pages, page_size, num_kv_heads, head_dim] if :attr:`kv_layout` is ``NHD``.
+        The first tensor is the key cache, and the second tensor is the value cache.
 
         **Contiguity requirements (trtllm-gen backend):**
 
@@ -2403,17 +2399,10 @@ def trtllm_batch_decode_with_kv_cache(
         Setting the threshold to a higher value generally increases kernel performance at the cost of accuracy degradation.
         The actual threshold value equals the provided threshold_scale_factor divided by the context length.
 
-    kv_cache_sf : Optional[Union[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]]] = None
-        Per-block scale factors for NVFP4 KV cache. Accepts the same formats as
-        ``kv_cache``:
-
-        * a tuple ``(k_scales, v_scales)`` of 4-D tensors, each with shape
-          ``[num_pages, num_kv_heads, page_size, head_dim // 16]`` in HND layout.
-        * a single 5-D tensor with shape
-          ``[num_pages, 2, num_kv_heads, page_size, head_dim // 16]`` in HND layout,
-          where dim 1 holds k (index 0) and v (index 1) scales.
-
-        Both tensors have dtype ``torch.float8_e4m3fn``.
+    kv_cache_sf : Optional[Tuple[torch.Tensor, torch.Tensor]] = None
+        Per-block scale factors for NVFP4 KV cache, as a tuple of ``(k_scales, v_scales)``.
+        Each scale tensor has shape ``[num_pages, num_kv_heads, page_size, head_dim // 16]``
+        in HND layout, with dtype ``torch.float8_e4m3fn``.
 
         **Contiguity requirements (trtllm-gen backend):**
 
@@ -2435,7 +2424,18 @@ def trtllm_batch_decode_with_kv_cache(
     """
     enable_pdl = device_support_pdl(query.device) if enable_pdl is None else enable_pdl
 
-    k_cache, v_cache = _unpack_paged_kv_cache(kv_cache, kv_layout)
+    if isinstance(kv_cache, tuple):
+        k_cache, v_cache = kv_cache
+    else:
+        if kv_cache.shape[1] == 1:
+            k_cache, v_cache = kv_cache, kv_cache
+        else:
+            assert kv_cache.shape[1] == 2, (
+                "When kv_cache is a single tensor, the second dimension must be 1 or 2"
+            )
+            # NOTE(Zihao): unbind transforms [num_pages, 2, ...] to ([num_pages, ...], [num_pages, ...])
+            # it doesn't change underlying storage
+            k_cache, v_cache = kv_cache.unbind(dim=1)
 
     if (
         k_cache.dtype == torch.uint8 or v_cache.dtype == torch.uint8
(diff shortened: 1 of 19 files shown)


flashinfer/decode.py
```diff
@@ -1448,13 +1449,24 @@ def run(
...
+                    # has a different optional-param layout than prefill.py's paged_run
+                    run_args += [paged_kv_cache]
+
+                run_args += [
                     self._num_qo_heads,
                     self._num_kv_heads,
                     self._block_tables,
                     self._kv_lens_buffer,
                     page_size,
+                    None,  # max_q_len (not applicable for decode)
```
qsang-nv: Is it possible that `self._cached_module.paged_run` resolves to the one in `get_trtllm_gen_decode_module` (L2117-L2160)? That signature has `... page_size, max_kv_len, sinks, ...` — no `max_q_len` / `batch_size` / `cum_seq_lens_q` / `cum_seq_lens_kv` slots. The 4 `None` placeholders would shift everything by one and trigger `assert max_kv_len is not None` at L2169. Could the trtllm-gen path need a separate `run_args` branch?

flashinfer/prefill.py
```diff
@@ -3271,6 +3320,10 @@ def run(
             logits_soft_cap = 0.0
         if sm_scale is None:
             sm_scale = 1.0 / math.sqrt(q.size(-1))
+        if q_scale is not None:
```
qsang-nv: Small concern: the new `sm_scale *= q_scale / k_scale` at L3323-3326 is unconditional, but the cuDNN ragged branch also passes `q_scale/k_scale` to cuDNN which re-applies them via `descale_q/k` on the FP8 graph (cudnn/prefill.py:389-391). The NVFP4 test (BF16 Q) won't hit this, but a future FP8 Q + cuDNN ragged user would get `q_scale² × k_scale²`. Consider gating the fold on `self._backend != "cudnn"` to keep the cuDNN path untouched.

flashinfer/prefill.py
```diff
@@ -465,6 +471,8 @@ def get_batch_prefill_module(backend, *args):
             "int_workspace_buffer",
             "o",
             "maybe_lse",
+            "maybe_k_cache_sf",
```
qsang-nv: `maybe_k_cache_sf` / `maybe_v_cache_sf` (and paged `key_block_scales` / `value_block_scales`) are read-only in the kernel — only used as load sources in `produce_kv_sf` / `page_produce_kv_sf`. Should they be dropped from `mutates_args`?

flashinfer/quantization/fp4_quantization.py
```diff
@@ -875,6 +937,16 @@ def e2m1_and_ufp8sf_scale_to_float(
     major, minor = get_compute_capability(
         torch.device("cuda:0")
     )  # select any cuda device to get a compute capability
+    if major * 10 + minor < 90:
```
qsang-nv, line 940: The inline comment at L940 says "No SM100 kernel available", but the gate is `major * 10 + minor < 90`. Shouldn't it be `< 100`?

flashinfer/prefill.py
```diff
@@ -3284,18 +3337,28 @@ def run(
                 check_shape_dtype_device(
                     lse, (q.size(0), q.size(1)), torch.float32, q.device, "lse"
                 )
+        # Unpack kv_cache_sf for NVFP4 ragged KV
+        k_sf, v_sf = None, None
+        if kv_cache_sf is not None:
+            if isinstance(kv_cache_sf, tuple):
```
qsang-nv, line 3364: This is the only spot that accepts `tuple | Tensor` but not `list` — would trigger `AttributeError` (not `TypeError`) on `list` or invalid input. Comparing with the other four `kv_cache_sf` unpackers gets a bit tangled:

- `decode.py` L1298 + L1448, `prefill.py` L2280 (paged), L4055 (trtllm): annotated as `Tuple[Tensor, Tensor]` but impl accepts `tuple | list | Tensor`, error message says `"tuple/list ... or stacked tensor"` — annotation lags behind impl
- `attention.py` L182 uses `_unpack_paged_kv_cache` which depends on `kv_layout`

Could all five be routed through a single `_unpack_kv_cache_sf()` helper with one canonical type annotation and error message? That also fixes this ragged path.

include/flashinfer/attention/prefill.cuh
```diff
@@ -377,6 +428,163 @@ __device__ __forceinline__ void page_produce_kv(typename KTraits::SharedStorage*
...
+  constexpr uint32_t NUM_SF_ITERS = (SF_TOTAL_BYTES / 4 + THREADS_PER_CTA - 1) / THREADS_PER_CTA;
+
+  uint8_t* sf_smem = produce_v ? smem_storage->v_sf_smem : smem_storage->k_sf_smem;
+  const uint32_t thread_id = warp_idx * 32 + lane_idx;
+
+#pragma unroll
+  for (uint32_t k = 0; k < NUM_SF_ITERS; ++k) {
+    const uint32_t flat_uint32_idx = thread_id + k * THREADS_PER_CTA;
+    const uint32_t flat_byte = flat_uint32_idx * 4;
+    // sf_smem_col is 4-byte aligned since SF_COLS is a multiple of 4 for all valid head_dims.
```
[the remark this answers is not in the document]
Tom-Zheng: Fixed comment

flashinfer/decode.py
```diff
@@ -1448,13 +1449,24 @@ def run(
                     rope_theta,
                     0,  # token_pos_in_items_len
                     self._workspace_size,
-                    paged_kv_cache,
+                ]
+
+                if self._backend == "trtllm-gen":
+                    # decode.py's trtllm-gen paged_run (get_trtllm_gen_decode_module)
+                    # has a different optional-param layout than prefill.py's paged_run
+                    run_args += [paged_kv_cache]
```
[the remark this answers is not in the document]
Tom-Zheng: Not an issue.

flashinfer/decode.py
Tom-Zheng: It makes sense, modified.

flashinfer/quantization/fp4_quantization.py
Tom-Zheng: Fixed the comment. The kernel actually works for SM90 and SM100.

flashinfer/prefill.py
Tom-Zheng: I have unified them. Thanks.

flashinfer/prefill.py
Tom-Zheng: Done. Thanks

flashinfer/prefill.py
Tom-Zheng: Done, thanks.

flashinfer/prefill.py
qsang-nv: Same fix still needed at L643-644 for `paged_run`'s `key_block_scales` / `value_block_scales`.

Tom-Zheng: Done. Thanks!

qsang-nv (approved):
LGTM

flashinfer/decode.py
```diff
@@ -1258,9 +1264,9 @@ def run(
         q_scale : Optional[float]
             The calibration scale of query for fp8 input, if not provided, will be set to ``1.0``.
         k_scale : Optional[float]
-            The calibration scale of key for fp8 input, if not provided, will be set to ``1.0``.
+            The calibration scale of key for fp8 or nvfp4 input, if not provided, will be set to ``1.0``.
         v_scale : Optional[float]
-            The calibration scale of value for fp8 input, if not provided, will be set to ``1.0``.
+            The calibration scale of value for fp8 or nvfp4 input, if not provided, will be set to ``1.0``.
```
bkryu, line 1269: Can we clarify in the docstring, maybe in the `__init__` function in the wrapper which backends support the newly added NVFP4 KV?

I think I know the answer (`fa2`) but would be help to be clear for users.

flashinfer/decode.py
```diff
@@ -2267,11 +2305,15 @@ def trtllm_batch_decode_with_kv_cache(
...
-        The first tensor is the key cache, and the second tensor is the value cache.
+        The paged KV-Cache stored as a tuple of tensors or a single tensor:
+
+        * a tuple ``(k_cache, v_cache)`` of 4-D tensors, each with shape
+          ``[num_pages, num_kv_heads, page_size, head_dim]`` if :attr:`kv_layout` is ``HND``,
+          or ``[num_pages, page_size, num_kv_heads, head_dim]`` if :attr:`kv_layout` is ``NHD``.
+        * a single 5-D tensor with shape
+          ``[num_pages, 2, num_kv_heads, page_size, head_dim]`` if :attr:`kv_layout` is ``HND``,
+          or ``[num_pages, 2, page_size, num_kv_heads, head_dim]`` if :attr:`kv_layout` is ``NHD``,
+          where dim 1 holds k (index 0) and v (index 1).
```
bkryu: Are we changing the capability to provide `[num_pages, 1,  page_size, num_kv_heads, head_dim]`?

flashinfer/decode.py
```diff
@@ -2647,10 +2679,15 @@ def xqa_batch_decode_with_kv_cache(
...
-        or [num_pages, num_kv_heads, page_size, head_dim] if :attr:`kv_layout` is ``HND``.
+        The paged KV-Cache stored as a tuple of tensors or a single tensor:
+
+        * a tuple ``(k_cache, v_cache)`` of 4-D tensors, each with shape
+          ``[num_pages, page_size, num_kv_heads, head_dim]`` if :attr:`kv_layout` is ``NHD``,
+          or ``[num_pages, num_kv_heads, page_size, head_dim]`` if :attr:`kv_layout` is ``HND``.
+        * a single 5-D tensor with shape
+          ``[num_pages, 2, page_size, num_kv_heads, head_dim]`` if :attr:`kv_layout` is ``NHD``,
+          or ``[num_pages, 2, num_kv_heads, page_size, head_dim]`` if :attr:`kv_layout` is ``HND``,
+          where dim 1 holds k (index 0) and v (index 1).
```
bkryu: Ditto about single tensor losing the capability to use `kv_cache[1]==1`

flashinfer/prefill.py
```diff
@@ -2175,9 +2232,9 @@ def run(
         q_scale : Optional[Union[float, torch.Tensor]]
             The calibration scale of query for fp8 input, if not provided, will be set to ``1.0``.
         k_scale : Optional[Union[float, torch.Tensor]]
-            The calibration scale of key for fp8 input, if not provided, will be set to ``1.0``.
+            The calibration scale of key for fp8 or nvfp4 input, if not provided, will be set to ``1.0``.
```
bkryu, line 2237: Again would be good to clarify in the `__init__` function of the wrapper or anywhere else which backend supports nvfp4

flashinfer/decode.py
Tom-Zheng: Added in flashinfer/decode.py:1301 and flashinfer/prefill.py:2270. Thanks.

flashinfer/decode.py
Tom-Zheng: My intent was to unify the paged_kv APIs. But is not necessary in this PR. I have revert the changes.

flashinfer/decode.py
Tom-Zheng: Reverted. Same reason as above.

flashinfer/prefill.py
Tom-Zheng: Added at flashinfer/prefill.py:2270

bkryu (approved):

leonardHONG:
The default `BatchDecodeWithPagedKVCacheWrapper` (`use_tensor_cores=False`) doesn't appear to support NVFP4 KV — the legacy decode kernel has no FP4 path, and `kv_cache_sf` never gets threaded through to `run_args` in the non-tensor-cores branch.

  Seems intentional (tests use `use_tensor_cores=True`, docstring only mentions fa2/trtllm-gen). Would a guard be welcome — auto-route or raise `NotImplementedError`? Happy to send a PR either way.

Tom-Zheng:
> The default `BatchDecodeWithPagedKVCacheWrapper` (`use_tensor_cores=False`) doesn't appear to support NVFP4 KV — the legacy decode kernel has no FP4 path, and `kv_cache_sf` never gets threaded through to `run_args` in the non-tensor-cores branch.
> 
> Seems intentional (tests use `use_tensor_cores=True`, docstring only mentions fa2/trtllm-gen). Would a guard be welcome — auto-route or raise `NotImplementedError`? Happy to send a PR either way.

Thanks for you catch. I think a `NotImplementedError` should be enough.

leonardHONG:
> > The default `BatchDecodeWithPagedKVCacheWrapper` (`use_tensor_cores=False`) doesn't appear to support NVFP4 KV — the legacy decode kernel has no FP4 path, and `kv_cache_sf` never gets threaded through to `run_args` in the non-tensor-cores branch.
> > Seems intentional (tests use `use_tensor_cores=True`, docstring only mentions fa2/trtllm-gen). Would a guard be welcome — auto-route or raise `NotImplementedError`? Happy to send a PR either way.
> 
> Thanks for you catch. I think a `NotImplementedError` should be enough.

Thanks @Tom-Zheng — will send a PR shortly.

PR #3100, opened 2026-04-17
Add int8 paged KV support to main paths

lesj0610:
## 📌 Description

The main paged-KV path had no int8 support. This PR extends the following to accept int8 KV cache:

- append
- single decode, single prefill
- batch decode, batch prefill

On Hopper, auto backend selection routes to FA2 when FA3 int8 KV is unavailable, so no combination falls through to an unsupported path.

Tested on Ampere (A100) and Hopper (H100):

```bash
python -m pytest tests/attention/test_int8_paged_kv.py -v
```

7 tests passed on both architectures. The int4 part is in a separate follow-up PR.

### ✅ Pre-commit Checks
- [x] I have installed pre-commit by running `pip install pre-commit` (or used your preferred method).

### 🧪 Tests

Retargeted against `main` after the `v0.6.8` release branch cut.

## Summary by CodeRabbit

* **New Features**
  * Added Int8 quantized key/value cache support for paged and single-step attention; improved Int8 vector handling for efficient kernels

* **Bug Fixes**
  * Fixed prefill scaling to avoid double-applying V scales; added validated scalar/tensor scale handling
  * FA3 backend explicitly disabled for Int8 KV paths

* **Tests**
  * Added extensive tests validating Int8 paged KV behavior, FA3 gating, and FP8 scale correctness

### DIFF
--- a/csrc/page.cu
+++ b/csrc/page.cu
@@ -90,7 +90,7 @@ void append_paged_kv_cache(TensorView append_key, TensorView append_value, Tenso
 
   ffi::CUDADeviceGuard device_guard(append_key.device().device_id);
   const cudaStream_t stream = get_stream(append_key.device());
-  bool success = DISPATCH_DLPACK_DTYPE_TO_CTYPE(paged_k_cache.dtype(), c_type, [&] {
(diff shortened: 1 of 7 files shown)


lesj0610:
addressed in latest commits.

PR #3101, opened 2026-04-17
Add int4 paged KV support to main paths

lesj0610:
## 📌 Description

Builds on the int8 paged-KV work in #3100 to add int4 support.

Incremental review against the int8 branch:
https://github.com/lesj0610/flashinfer/compare/codex/int8-paged-kv-main-v068...codex/int4-paged-kv-main-v068

`torch.uint8` is already used in some paths as an FP4 container, so a plain uint8 input creates a semantic conflict. An explicit `INT4Tensor` wrapper is used to keep the contract unambiguous. Storage is packed uint8 with grouped fp16 scales (`group_size=32`).

The implementation goes through staged dequantization to fp16 before calling existing kernels. On Hopper, auto backend selection falls back to FA2 the same way as in #3100. The following are not included in this PR:

- CUDA graph: explicitly blocked, as the staging step requires temporary allocation
- Native FA3, XQA, and TRTLLM-gen int4 paths

Because the upstream PR base must stay on `main`, GitHub still shows the shared int8 commits in the main PR diff until #3100 lands. The compare link above shows the int4-only incremental delta.

Tested on Ampere (A100) and Hopper (H100):

```bash
python -m pytest tests/attention/test_int4_paged_kv.py -v
```

51 tests passed on both architectures.

### ✅ Pre-commit Checks
- [x] I have installed pre-commit by running `pip install pre-commit` (or used your preferred method).

### 🧪 Tests

Depends on #3100.

Retargeted against `main` after the `v0.6.8` release branch cut.

### DIFF
--- a/csrc/page.cu
+++ b/csrc/page.cu
@@ -90,7 +90,7 @@ void append_paged_kv_cache(TensorView append_key, TensorView append_value, Tenso
 
   ffi::CUDADeviceGuard device_guard(append_key.device().device_id);
   const cudaStream_t stream = get_stream(append_key.device());
-  bool success = DISPATCH_DLPACK_DTYPE_TO_CTYPE(paged_k_cache.dtype(), c_type, [&] {
+  bool success = DISPATCH_DLPACK_DTYPE_TO_CTYPE_QKV(paged_k_cache.dtype(), c_type, [&] {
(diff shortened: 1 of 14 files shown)


lesj0610:
addressed in latest commits.

PR #2886, opened 2026-03-24
fix moe benchmark

nv-yunzheq:
## 📌 Description

Fix inconsistent between autotune and real benchmark for deepseek moe benchmark script

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [ ] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [ ] I have installed the hooks with `pre-commit install`.
- [ ] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [ ] Tests have been added or updated as needed.
- [ ] All tests are passing (`unittest`, etc.).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed autotune log interleaving with benchmark results output
  * Improved benchmark accuracy via one-time pre-warm and GPU synchronization
  * Safer memory handling to avoid buffer overflow when larger token batches are selected

* **Refactor**
  * Integrated autotuning into the main benchmark workflow so autotune runs during benchmarking rather than beforehand

PR #3126, opened 2026-04-20
autotuner: check cache before synthesizing profile input tensors

leejnau:
## 📌 Description

  AutoTuner.choose_one's tuning-mode loop calls _prepare_input_tensors(p, inputs) before checking the cache. On a cache hit the synthesized tensors are thrown away, but their torch.rand / torch.randint kernel launches already happened on the
  device. For any caller that runs choose_one repeatedly inside autotune(True) (e.g. a benchmark sharing its tuning and measurement call-sites), those kernels recur on every warm-cache call and get attributed to the measured region by CUPTI /
  nsys.

Move search_cache above _prepare_input_tensors in the tuning-mode loop. Skip synthesis on cache hit; synthesize as before on cache miss. The lookup passes the caller's inputs (rather than synthesized tensors), which aligns it with the
  non-tuning branch and the post-loop search_cache. Safe under the existing get_cache_key_extras contract (dtype-like properties preserved by synthesis).
  
## 🔍 Related Issues

https://github.com/flashinfer-ai/flashinfer/pull/2398

https://github.com/flashinfer-ai/flashinfer/pull/2886/

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Refactor**
  * Reduced runtime overhead during tuning by checking the cache before preparing/synthesized inputs, avoiding unnecessary expensive tensor synthesis on cache hits.
* **Documentation**
  * Clarified cache behavior and requirements: cache-key extras must be synthesis-invariant, and input shapes vs. actual inputs influence separate matching steps.

### DIFF
--- a/flashinfer/autotuner.py
+++ b/flashinfer/autotuner.py
@@ -365,6 +365,12 @@ class TunableRunner(ABC):
         Override this method to differentiate cache entries that share the same
         input shapes but differ in other properties (e.g. output dtype).
         The returned tuple must be hashable.
+
+        Returned values must be synthesis-invariant: the same tuple must
+        be produced for the caller's real inputs and for tensors the
+        autotuner would synthesize for the same profile (i.e., depend only
+        on dtype, is-None flags, or scalar-argument values -- not on
+        per-tensor content).
         """
         return ()
 
@@ -619,6 +625,16 @@ class AutoTuner:
         Returns:
             A tuple containing:
             [is_cache_hit, runner_id, tactic, stored_profile]
+
+        Note:
+            input_shapes and inputs feed orthogonal paths inside
+            this method: input_shapes flows only into
+            _find_nearest_profile (bucket matching) and inputs
+            flows only into r.get_cache_key_extras(inputs).  Callers
+            may therefore pass them describing different tensor sets
+            (e.g. a profile's opt_shapes alongside the caller's real
+            inputs) so long as get_cache_key_extras is
+            synthesis-invariant; see: TunableRunner.get_cache_key_extras.
         """
         with self._lock:
             for r in runners:
--- a/flashinfer/autotuner.py
+++ b/flashinfer/autotuner.py
@@ -745,15 +745,34 @@ class AutoTuner:
             pbar = None
             for _step, p in enumerate(profiles):
                 try:
-                    tensors = self._prepare_input_tensors(p, inputs)
+                    # Check the cache before synthesizing profile inputs.
+                    # `_prepare_input_tensors` launches a GPU kernel per
+                    # `DynamicTensorSpec`; on a cache hit the synthesized
+                    # tensors are immediately discarded.  Skipping them
+                    # matters for callers that invoke `choose_one`
+                    # repeatedly inside `autotune(True)` -- otherwise every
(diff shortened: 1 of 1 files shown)


flashinfer/autotuner.py
```diff
@@ -745,15 +745,34 @@ def choose_one(
...
+                    # dtype-like properties preserved by the synthesis
+                    # initializers.  Matches the non-tuning branch above
+                    # and the post-loop `search_cache` call below.
                     is_cache_hit, runner_id, tactic, _ = self.search_cache(
                         custom_op,
                         runners,
                         p.get_opt_shapes(),
                         tuning_config,
-                        inputs=tensors,
+                        inputs=inputs,
```
qiching, line 787: first i think your pr is very make sense to me. 
i have a qq:
after this change `input_shapes=p.get_opt_shapes()` and `inputs=inputs` no longer describe the same tensors. It is safe because inside `search_cache` they feed orthogonal paths, `input_shapes` to `_find_nearest_profile`, `inputs` to  `get_cache_key_extras`, and all current `get_cache_key_extras` overrides are synthesis invariant (dtype or is-None or scalars). 

Could you confirm if my understanding correct? and as a small nit, would you mind adding a line to search_cache's docstring making this explicit?

leejnau: Yes, your understanding is correct!
 * `input_shapes` flows only into `_find_nearest_profile(input_shapes, tuning_config)` → bucket matching.
 * `inputs` flows only into `r.get_cache_key_extras(inputs)` → per-runner extras.

The two never interact inside `search_cache`, so it's safe for the caller to pass them describing different tensors provided `get_cache_key_extras()` is synthesis-invariant (same return for the caller's real tensors vs. autotuner-synthesized tensors of the same profile).

nvpohanh:
Thanks for fixing this!

flashinfer/autotuner.py
qiching: thanks for confirming! LGTM on the core change. 

Could you also add the short docstring note to `search_cache` capturing exactly what you wrote above, that `input_shapes` flows `into _find_nearest_profile` and `inputs` flows into `get_cache_key_extras` only, and the caller is responsible for `get_cache_key_extras` being synthesis-invariant? 

Then future contributors do not have to re derive this from the PR.

Thanks

leejnau: > thanks for confirming! LGTM on the core change.
> 
> Could you also add the short docstring note to `search_cache` capturing exactly what you wrote above, that `input_shapes` flows `into _find_nearest_profile` and `inputs` flows into `get_cache_key_extras` only, and the caller is responsible for `get_cache_key_extras` being synthesis-invariant?
> 
> Then future contributors do not have to re derive this from the PR.
> 
> Thanks

I believe the docstring should be there on the second commit:  https://github.com/flashinfer-ai/flashinfer/commit/c7100c7a6e8ed65f16cd1a2dc6780ce23b00e6e5
Does that look ok or I missed something?

qiching: sry. I missed it. LGTM

nv-yunzheq (approved):

Issue #3130, opened 2026-04-21
[Bug] MLA decode return all zeros when index of page blocks is large


Tracin:
I am using flashinfer-python 0.6.8.post1
The attached script can reproduce the issue.

[debug_flashinfer.py](https://github.com/user-attachments/files/26921564/debug_flashinfer.py)

Tracin:
A fix PR. https://github.com/flashinfer-ai/flashinfer/pull/3136

saltyminty:
Marking as closed now that 3136 has been merged. Feel free to reopen if you this again!

PR #3136, opened 2026-04-21
fix(mla): widen page index to int64_t to avoid 32-bit overflow

Tracin:
## 📌 Description
In the MLA decode/prefill KV load path, `indices[q] * ckv_stride_page` was computed in 32-bit because `IdType` is `int32_t` and `*_stride_page` is `uint32_t`; the product wraps modulo 2^32 before any widening to `int64_t` (Hopper) or pointer arithmetic (FA2). For large page pools (e.g. page_idx ~1M with page_size=32, kv_lora_rank=512, stride=16384) the true product exceeds 2^32 and the kernel reads the wrong page, producing all-zero outputs. Cast the selected page index to `int64_t` at all three sites (mla.cuh NUM_MMA_KV==1 and !=1 branches, and mla_hopper.cuh prefetch_offset) so the multiply executes in 64-bit.

## 🔍 Related Issues

https://github.com/flashinfer-ai/flashinfer/issues/3130

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed integer overflow in page-index arithmetic that could select the wrong memory page for very large attention KV caches, improving reliability and correctness for large-context workloads.

* **Tests**
  * Added a regression test that reproduces the page-index overflow case and verifies correct outputs for large KV cache scenarios, with gating to ensure it runs only when sufficient GPU support and memory are available.

### DIFF
--- a/include/flashinfer/attention/mla.cuh
+++ b/include/flashinfer/attention/mla.cuh
@@ -201,12 +201,17 @@ __device__ __forceinline__ void load_kv(
       uint32_t packed_block_iter = packed_block_iter_base + lane_idx / 8 + warp_idx_in_wg * 4;
       block_size.divmod(packed_block_iter, q, r);
 
-      DTypeKV* ckv_ptr = ckv +
-                         (packed_block_iter < packed_kv_bound ? indices[q] : 0) * ckv_stride_page +
-                         r * ckv_stride_n + (lane_idx % 8) * upcast_size<DTypeKV>();
-      DTypeKV* kpe_ptr = kpe +
-                         (packed_block_iter < packed_kv_bound ? indices[q] : 0) * kpe_stride_page +
-                         r * kpe_stride_n + (lane_idx % 8) * upcast_size<DTypeKV>();
+      // Cast page index to int64_t before multiplying to avoid overflow.
+      DTypeKV* ckv_ptr =
+          ckv +
+          static_cast<int64_t>(packed_block_iter < packed_kv_bound ? indices[q] : 0) *
+              ckv_stride_page +
+          r * ckv_stride_n + (lane_idx % 8) * upcast_size<DTypeKV>();
+      DTypeKV* kpe_ptr =
+          kpe +
+          static_cast<int64_t>(packed_block_iter < packed_kv_bound ? indices[q] : 0) *
+              kpe_stride_page +
+          r * kpe_stride_n + (lane_idx % 8) * upcast_size<DTypeKV>();
 
 #pragma unroll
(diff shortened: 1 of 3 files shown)


qsang-nv:
Nice fix — cast is in the right place and all three call sites are covered.

One suggestion: add a minimal regression test that forces `page_idx * stride_page > 2^32`. This is a silent-output-corruption bug (wrong output, no crash), so without a guard test, a future refactor of `IdType` or `*_stride_page` types could easily reintroduce it.

You don't need a huge KV cache — a sparse `kv_indices` with a few large index values pointing at a small real allocation should hit the overflow path in ~30 lines in `tests/attention/test_mla_decode_kernel.py`.

Tracin:
@qsang-nv Thanks for the review! However I do not get how `large index values pointing at a small real allocation`. I suppose we need a real address for large page_idx * stride_page.

qsang-nv:
@Tracin You are right, we do need a real address, however, it can be smaller than the script you provided in the issue.

The overflow triggers as soon as `page_idx × ckv_stride_page` exceeds `2^32`, so we just need `page_idx` to be **just past that threshold** — not at 1M.

**Overflow threshold:**

- `ckv_stride_page = ckv_cache.stride(0)` is in **elements**, and for a contiguous `[num_pages, page_size, kv_lora_rank]` tensor equals `page_size × kv_lora_rank = 32 × 512 = 16384`.
- Overflow happens when `page_idx × 16384 ≥ 2^32`.
- `2^32 / 16384 = 262144`, so the minimum `page_idx` that triggers it is **`262144`**.

**Allocation needed:**

Tensors are contiguous, so to have `page_idx = 262144` as a valid index, `max_pages` must be at least `262145`. Using `max_pages = 262170` to match the 26-page decode scenario in your debug script:

- `ckv_cache` elements = `262170 × 32 × 512` = `4,294,983,680` ≈ `4.295 × 10^9` (bf16 → ×2 bytes) → **~8.59 GB**
- `kpe_cache` elements = `262170 × 32 × 64` = `536,884,480` (bf16 → ×2 bytes) → **~1.07 GB**
- **Total ≈ 9.66 GB**

Fits on most modern GPUs. Gate with `@pytest.mark.skipif(torch.cuda.mem_get_info()[0] < 12 * 1024**3, ...)` to skip on small-memory runners. I think we can add this test to CI as-is first. If the memory footprint turns out to cause trouble on some runner, we can remove it as I see there are comments waring about overflow.

Tracin:
@qsang-nv I see. Test is added!

qsang-nv (approved):
CI passed, LGTM

saltyminty (approved):

PR #3142, opened 2026-04-21
[cli] add CLI helper for flashinfer-jit-cache and flashinfer-cubin wheel installs

### LINKED ISSUE #3033: CLI utility that can aot download the jit cache
From discussion with @mgoin:

vLLM upstream would like to improve the default vLLM experience. Today, it does not include `flashinfer-jit-cache`, for a couple of reasons:

1. `flashinfer-jit-cache` is not on pypi (too big)
2. pip does not allow dispatch based on CUDA version, so there's no clean way to add the right jit cache as a dependency

The proposed solution is to add a flashinfer CLI utility that can aot download the jit cache based on CUDA version in the environment. So the flow would be something like:

```
pip install vllm
flashinfer download-jit-cache
```
optional flag for CUDA version, otherwise, pick based on the environment. Ideally could also split by SM architecture.

@mgoin will post the current UX for comparison.

dierksen:
## Summary
Add CLI commands to install the matching flashinfer-jit-cache and/or flashinfer-cubin wheels from the FlashInfer wheel index.

Also update documentation to reflect using `--index-url` for `flashinfer-cubin` instead of installing my pypi.

## What changed
- detect the FlashInfer and CUDA versions and build the matching pip install command
- document the new command in the CLI docs
- add CLI tests for version/index selection and dry-run behavior
- update docs to use `--index-url` for cubin wheel

## Why
Packages is too large for PyPI, so users need a helper that can resolve the correct custom index-url and matching package version automatically.

## Validation
- Click dry-run smoke test in  verified the generated command resolves to  with 
-  was not available in the current environment, so the added CLI tests were not executed here

Closes #3033

## Summary by CodeRabbit

- **New Features**
  - Added new CLI commands to install pre-built cubin and JIT-cache wheels.
  - Added `download-kernels` to install both kernel artifact sets together.
  - Improved CUDA-version handling with auto-detection, compatibility mapping, nightly installs, and `--dry-run`.
  - Added a hidden `download-jit-cache` alias for convenience.

- **Documentation**
  - Updated README quickstart and artifact management CLI docs to use the new `flashinfer` subcommands.
  - Clarified what cubin downloads include and documented new install/download command behavior.

- **Tests**
  - Added/expanded mocked CLI tests covering exact install invocations, dry-run behavior, CUDA fallback/validation, aliases, and failure handling.


dierksen:
@flashinfer-bot run

flashinfer/__main__.py
```diff
@@ -34,9 +38,13 @@
 # Import __version__ from centralized version module
 from .version import __version__
 
+_SUPPORTED_JIT_CACHE_CUDA_VERSIONS = tuple(
+    Version(version) for version in ("12.8", "12.9", "13.0")
```
bkryu, line 45: Hard coding here does not seem very ideal. Any way to avoid this?

flashinfer/__main__.py
```diff
@@ -235,6 +453,122 @@ def clear_cubin_cmd():
...
+    cubin_index_url,
+    jit_cache_index_url,
+    nightly,
+    dry_run,
+):
+    """Install matching flashinfer-cubin and flashinfer-jit-cache wheels."""
+    _install_cubin_wheel(flashinfer_version, cubin_index_url, nightly, dry_run)
+    _install_jit_cache_wheel(
+        cuda_version, flashinfer_version, jit_cache_index_url, nightly, dry_run
+    )
```
bkryu, line 585: Here we install cubins first and then jit-cache. If there is a failure in the cubin installation, it will not proceed with the jit-cache wheel. Do we want this to be the default behavior?

bkryu (commented):
Tested on local CUDA 13.1 toolkit + torch-cu130 virtual environment and works as expected.

Left a few non-blocking questions before approving.

flashinfer/__main__.py
dierksen: I added a note, but it's basically synced to the paths in our release pipeline. I suppose we could load it from there and/or have a common source of truth for both, and then make sure that gets packaged in on releases?

flashinfer/__main__.py
dierksen: Fixed (along with deduplicating a function that coderabbit mentioned). Now it will attempt to do both and then report failures afterward.

yongwww (approved):

flashinfer/__main__.py
mgoin: Nit: It would be useful to expand this to a list of aliases. For instance if a user does `--cuda-version 13.2` since they have 13.2 locally, but the 13.0 build is compatible, we resolve to 13.0 internally.

bkryu (approved):

mgoin (approved):
LGTM, let's go for it!

PR #3145, opened 2026-04-22
Fix OOB crash in intermediate_states indexing for GDN decode MTP kernel

wenscarl:
co-authored by @YAMY1234 
## Problem

  `gdn_decode_bf16state_mtp_kernel` crashes with an out-of-bounds GPU memory write when
  `intermediate_states_buffer` is provided and `pool_size > B` with `initial_state_indices`
  pointing to upper pool slots (the normal serving scenario).

  **Affected path:** `flashinfer/gdn_kernels/gdn_decode_bf16_state.py:1873`

  \```python
   Before (wrong)
  flat_idx = cache_idx * T * HV + i_t * HV + i_hv

   After (correct)
  flat_idx = i_n * T * HV + i_t * HV + i_hv
  \```

  ## Root Cause

  When `intermediate_states` support was added to the BF16 state kernel, the author reused
  the `cache_idx`-based addressing pattern from the persistent state pool access:

  \```python
  flat_state_idx = cache_idx * HV + i_hv   # correct: h0_source is pool-scoped
  \```

  ...and extended it by analogy to add a `T` dimension for `intermediate_states`. This is
  wrong because the two buffers have different ownership semantics:

  - **`h0_source`** (the state pool) is pool-scoped — persists across decode steps, one slot
    per concurrent request in the system → correctly indexed by `cache_idx` (pool slot)
  - **`intermediate_states`** is batch-scoped — a per-forward-pass output capturing states at
    each of the T steps → should be indexed by `i_n` (batch position)

  The float32 counterpart `gdn_decode_mtp.py` has always used `i_n` correctly. The BF16
  kernel diverged when it was written.

  The bug was invisible in existing tests because they always set `pool_size = batch_size`
  with `initial_state_indices = arange(B)`, making `cache_idx == i_n` identically. The
  buggy and correct indexing produce the same result in that configuration. The docstring
  describing the buffer shape as `[pool_size, T, HV, V, K]` further reinforced the incorrect
  mental model.

  The crash only manifests in the realistic serving scenario where `pool_size >> B` and
  `initial_state_indices` contains values ≥ `B`, causing `cache_idx`-based writes to go
  beyond the end of a batch-sized buffer.

  ## Fix

  Change `cache_idx` to `i_n` at the `intermediate_states` indexing site, and update the
  docstring to reflect the correct buffer shape `[B, T, HV, V, K]` instead of
  `[pool_size, T, HV, V, K]`.

  ## Test Changes

  The existing `test_gdn_decode_bf16_state_mtp_kernel` always used `pool_size = batch_size`,
  which masked the bug entirely. Two changes are made:

  1. **`pool_size_multiplier` parameter** added to the helper
     `_test_gdn_decode_bf16_state_mtp_kernel`. When `> 1`, it sets
     `pool_size = batch_size * pool_size_multiplier` and assigns each batch entry to an upper
     pool slot (`initial_state_indices = arange(B) + pool_size - B`), so `cache_idx >= B`
     for every entry. The `intermediate_states_buffer` is allocated with `batch_size` as its
     first dimension — the semantically correct size — which is smaller than `pool_size`.
     With the buggy `cache_idx` indexing the kernel writes out of bounds and crashes; after
     the fix it produces results matching the reference.

  2. **`@pytest.mark.parametrize("pool_size_multiplier", [1, 4])`** added to the public
     test, doubling the existing matrix with a realistic pool-vs-batch configuration. The
     `pool_size_multiplier=4, cache_intermediate_states=True` cases are the ones that
     directly catch this bug.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed intermediate-state indexing so cached states are written and read per batch slice consistently.

* **Tests**
  * Expanded tests to cover pool sizes larger than the batch (parameterized), including remapped indices and updated assertions for those scenarios.

* **Documentation**
  * Updated API/docs and launcher expectations to reflect the new intermediate-states buffer shape.

### DIFF
--- a/flashinfer/gdn_kernels/gdn_decode_bf16_state.py
+++ b/flashinfer/gdn_kernels/gdn_decode_bf16_state.py
@@ -1095,7 +1095,7 @@ MTP_ILP_ROWS = 8  # Process 8 V-rows simultaneously per group iteration
 @cute.kernel
 def gdn_decode_bf16state_mtp_kernel(
     h0_source: cute.Tensor,  # [pool_size * HV, V, K] as BF16
-    intermediate_states: cute.Tensor,  # [pool_size * T * HV, V, K] as BF16 (or dummy)
+    intermediate_states: cute.Tensor,  # [B * T * HV, V, K] as BF16 (or dummy)
     vec_size: cutlass.Constexpr[int],
     num_v_tiles: cutlass.Constexpr[int],
     tile_v: cutlass.Constexpr[int],
@@ -2024,7 +2024,7 @@ def gdn_decode_bf16state_mtp_kernel(
 @cute.jit
 def run_gdn_decode_bf16state_mtp(
     h0_source: cute.Tensor,  # [pool_size * HV, V, K] BF16
-    intermediate_states: cute.Tensor,  # [pool_size * T * HV, V, K] BF16 (or dummy)
+    intermediate_states: cute.Tensor,  # [B * T * HV, V, K] BF16 (or dummy)
     A_log: cute.Tensor,
     a: cute.Tensor,
     dt_bias: cute.Tensor,
@@ -2524,7 +2524,7 @@ def gated_delta_rule_mtp(
         initial_state_indices: [B] int32 - indices into state pool (read)
         output_state_indices: Optional [B] int32 - indices for writing updated state.
             Defaults to initial_state_indices when None.
-        intermediate_states_buffer: Optional [pool_size, T, HV, V, K] bf16
+        intermediate_states_buffer: Optional [B, T, HV, V, K] bf16
         disable_state_update: bool - if True, don't update initial state
         scale: Optional, default 1/sqrt(K)
         output: Optional pre-allocated output tensor [B, T, HV, V] bf16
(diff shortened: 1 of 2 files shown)


kahyunnam (commented):
Please fix the pre-commit failure here: https://github.com/flashinfer-ai/flashinfer/actions/runs/24785336107/job/72527611010?pr=3145

flashinfer/gdn_kernels/gdn_decode_bf16_state.py
ameynaik-hub: - from `[pool_size * T * HV, V, K]`
    - to `[B * T * HV, V, K]`

ameynaik-hub: - from `[pool_size * T * HV, V, K]`
    - to `[B * T * HV, V, K]`

ameynaik-hub: - from `Optional [pool_size, T, HV, V, K] bf16`
    - to `Optional [B, T, HV, V, K] bf16`

ameynaik-hub (approved):
Thanks for the fix.
The code fix looks correct to me. One small cleanup before merge: a few comments/docstrings still describe `intermediate_states` as pool-scoped, but this PR makes it batch-scoped.

ameynaik-hub (approved):

nvpohanh:
@wenscarl Could you rebase this?

wenscarl:
adopted by another merged PR.

PR #3151, opened 2026-04-23
perf: Add no-bias path for tinygemm_bf16

bkryu:
## 📌 Description

This PR updates tinygemm_bf16 to use a dedicated no-bias kernel path when `bias=None`, **avoiding the previous zero-bias materialization in the Python wrapper**. It also adds shared benchmark-harness support and benchmark smoke coverage for `tinygemm_bf16`, while keeping the branch scoped to BF16-only changes.

If you want, I can also turn that into a full PR body with Summary and Test plan sections.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * Added a BF16 GEMM routine with benchmarking and optional correctness validation.
  * Added a no-bias GEMM variant to run without requiring a bias tensor.

* **Performance**
  * Reduced overhead when bias is absent by avoiding unnecessary bias allocation.

* **Compatibility**
  * BF16 routine is available only on supported GPU backends/architectures.

### DIFF
--- a/csrc/tinygemm2.cu
+++ b/csrc/tinygemm2.cu
@@ -141,7 +141,7 @@ __device__ uint32_t elect_one_sync() {
 // ============================================================================
 
 template <int WARP_TILE_M, int TILE_M, int TILE_N, int TILE_K, int STAGES, int STAGE_UNROLL,
-          bool USE_PDL = false>
+          bool USE_PDL = false, bool HAS_BIAS = true>
 __global__ __launch_bounds__(384, 1) void tinygemm_kernel(
     __nv_bfloat16* output, __nv_bfloat16* weights, __nv_bfloat16* activations, __nv_bfloat16* bias,
     int M, int N, int K, const __grid_constant__ CUtensorMap weight_map,
@@ -280,7 +280,9 @@ __global__ __launch_bounds__(384, 1) void tinygemm_kernel(
   // Compute threads
   else {
     // Sneak the bias load into the compute warps since they're just waiting
-    if (threadIdx.x < TILE_M) sh_bias[threadIdx.x] = bias[mib + threadIdx.x];
+    if constexpr (HAS_BIAS) {
+      if (threadIdx.x < TILE_M) sh_bias[threadIdx.x] = bias[mib + threadIdx.x];
+    }
 
     int stage = warp_id;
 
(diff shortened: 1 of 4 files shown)


csrc/tinygemm2.cu
jimmyzho: rather than only guarding the initialization of `bias_lo` and `bias_hi` can we guard this entire block under HAS_BIAS so there will be no chance of the fadd showing up?

jimmyzho: We can also guard this allocation with `HAS_BIAS`

flashinfer/gemm/routergemm.py
```diff
@@ -305,7 +305,22 @@ def tinygemm2_op_impl(
...
+        "flashinfer::tinygemm2_nobias_op",
+        mutates_args=["out"],
+    )
+    def tinygemm2_nobias_op_impl(
+        input: torch.Tensor,
+        weight: torch.Tensor,
+        out: torch.Tensor,
+        use_pdl: bool = False,
+    ) -> None:
+        module.tinygemm2_nobias_op(input, weight, out, use_pdl)
```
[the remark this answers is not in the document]
jimmyzho: should this be considered?

jimmyzho (commented):
overall looks good, have some minor suggestions

csrc/tinygemm2.cu
bkryu: Thanks @jimmyzho, Good point. I'd expect the compiler to remove it but it's better to be clear

bkryu: Yes, I updated this to keep the diff minimal.

Instead of reworking the shared-memory layout, I changed the declaration to a compile-time guarded shared array of just 1 float 
```
__shared__ nv_bfloat16 sh_bias[HAS_BIAS ? TILE_M : 1];
```
That keeps the original kernel structure intact while ensuring the no-bias specialization does not reserve a full TILE_M bias buffer.

flashinfer/gemm/routergemm.py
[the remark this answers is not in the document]
bkryu: @jimmyzho , the signature for `tinygemm2_op_impl`right above also uses 
```
    def tinygemm2_op_impl(
        input: torch.Tensor,
        weight: torch.Tensor,
        bias: torch.Tensor,
        out: torch.Tensor,
        use_pdl: bool = False,
    ) -> None:
        module.tinygemm2_op(input, weight, bias, out, use_pdl)
```
so for the sake of consistency and minimal diff, I lean towards keeping as is. LMK if you think it is best to address with an `# noqa: A002` as suggested above in both places.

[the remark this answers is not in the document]
jimmyzho: both seems fine, we can just keep it as is for consistency!

jimmyzho (approved):
lgtm!

aleozlx (approved):

PR #3001, opened 2026-04-07
[feat] Add blackwell GDN prefill kernel

jiahanc:
## 📌 Description

- Add Blackwell GDN prefill kernel written in cutedsl

```
GPU: NVIDIA B200 [Blackwell (SM100)]
Models: Qwen3.5 family (397B, 122B, 35B, 27B, 9B, 4B, 2B, 0.8B), d=128
FLA 0.4.2, Triton 3.5.1, PyTorch 2.11.0+cu130

Heads            Seqlens           h_qk  h_v    FI Blackwell (SM100)   TFLOPS  FLA/Triton   Speedup
---------------------------------------------------------------------------------------------------
397B/122B TP8    1x8192               2    8                  0.330ms    12.8      0.338ms     1.02x
397B/122B TP8    1x4096               2    8                  0.180ms    12.0      0.215ms     1.19x
397B/122B TP8    1x2048               2    8                  0.101ms    10.7      0.213ms     2.11x
397B/122B TP8    6144+2048            2    8                  0.258ms    16.6      0.285ms     1.10x
397B/122B TP8    4096+4096            2    8                  0.179ms    23.9      0.239ms     1.34x
397B/122B TP8    2048+6144            2    8                  0.256ms    16.8      0.285ms     1.11x
397B/122B TP8    1024+7168            2    8                  0.296ms    14.5      0.308ms     1.04x
397B/122B TP8    2048x4               2    8                  0.101ms    42.3      0.208ms     2.06x
397B/122B TP8    1024x8               2    8                  0.063ms    68.2      0.214ms     3.40x

397B/122B TP4    1x8192               4   16                  0.368ms    23.3      0.424ms     1.15x
397B/122B TP4    1x4096               4   16                  0.178ms    24.2      0.235ms     1.32x
397B/122B TP4    1x2048               4   16                  0.100ms    21.4      0.220ms     2.20x
397B/122B TP4    6144+2048            4   16                  0.257ms    33.5      0.377ms     1.47x
397B/122B TP4    4096+4096            4   16                  0.179ms    47.9      0.329ms     1.84x
397B/122B TP4    2048+6144            4   16                  0.259ms    33.2      0.376ms     1.45x
397B/122B TP4    1024+7168            4   16                  0.298ms    28.9      0.401ms     1.35x
397B/122B TP4    2048x4               4   16                  0.104ms    83.0      0.334ms     3.21x
397B/122B TP4    1024x8               4   16                  0.068ms   126.3      0.342ms     5.03x

[...]

PR #3155, opened 2026-04-23
fix(gdn): use physical SM count for SM100 persistent prefill kernel

arpera:
## 📌 Description

Fixes the `num_sm` issue CodeRabbit flagged on #3001 but which was not applied before merge: https://github.com/flashinfer-ai/flashinfer/pull/3001#discussion_r3048905545

The raw `HardwareInfo().get_max_active_clusters(1)` call returns 0 / stale values in spawned subprocesses (e.g. vLLM's EngineCore workers) where the CUDA driver API context has not been made current yet. The persistent tile scheduler then leaves some CTAs without any work and the kernel deadlocks at first call. Switch to `get_num_sm(q.device)`, matching the SM120 MoE dispatch.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Refactor**
  * Kernel compilation now derives device-specific SM and cluster counts at runtime, improving GPU resource allocation and leading to more consistent performance across different CUDA devices.


jiahanc (approved):
Thanks for the fix!

vadiklyutiy:
@jiahanc any idea why we didn't catch it with unit tests?

jiahanc:
> @jiahanc any idea why we didn't catch it with unit tests?

could be because the unit test doesnt have as high pressure as framework side, so bug is not exposed

kahyunnam (approved):
lgtm, thanks!

PR #3166, opened 2026-04-24
Refactor Part 3- Add block-per-token feature in the customized routing method

ChristinaZ:
## 📌 Description
Adds `routingIndicesBlockScoresKernel` — a block-per-token top-K kernel modeled on TRT-LLM's `deepseek_v3_topk_kernel` (no-groups path) — to `routingCustom`, and wires it into the dispatcher alongside the fused
cluster / warp-per-token paths.  Also removes a stale `num_experts % 4 == 0` precondition that no kernel actually depends on.
### Motivation
The fused single-cluster and warp-per-token top-K kernels have every cluster warp carry K-sized per-thread arrays.  For high-K / high-E configs (e.g. Nemotron Super V3 E=512/K=22 on B200), this produces large spill slots and caps occupancy at 1 block/SM.  Small-BS configs also suffer from idle-warp barrier stalls on the cluster path.
### Design
- **One block per token, 256 threads.**
- **Phase 1:** block-parallel preprocess via `PreprocessPolicy::applyToSmem` — all threads cooperatively write
  per-expert topK-key (and optional aux) values into smem.
- **Phase 2:** warp 0 alone runs a sort-based `reduceTopK` over `NumChunks = ceil(MaxNumExperts / 32)` elements per lane.
- **Phase 3:** warp 0 applies `PostprocessPolicy::applyWithAux`.
Only warp 0 holds K-sized register arrays, so per-block register pressure stays low (~64 regs/thread) and occupancy rises to 2 blocks/SM.
### Policy opt-in
Each preprocess / postprocess policy declares `static constexpr bool kSupportsBlockPerToken = true/false;`.
`PolicyPairSupportsBlockPerToken<Pre, Post>` is simply the AND of the two flags — no per-pair trait specialisation needed.  Policies that also need a separate `smemAux` array declare `static constexpr bool kNeedsAux = true;` (today only `ScaledSumNormalizePostprocess`).  

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

> pytest tests/moe/test_trtllm_gen_fused_moe.py::test_deepseek_ngroup1_block_per_token_routing -v

## Summary by CodeRabbit

* **Performance**
  * Added a block-per-token TopK scoring path and a runtime heuristic (env override) to choose the fastest kernel per workload.

* **New Features**
  * Block-per-token routing pathway with opt-in preprocess/postprocess policies and optional auxiliary support.
  * Split-pipeline mode to precompute block scores and defer permutation handling.

* **Improvements**
  * Removed expert-count divisibility restriction for more flexible configurations.

* **Tests**
  * Added routing tests targeting the new block-per-token path.

* **Chores**
  * Adjusted build script parallelism for non-aarch64 systems.

### DIFF
--- a/tests/moe/test_trtllm_gen_fused_moe.py
+++ b/tests/moe/test_trtllm_gen_fused_moe.py
@@ -3987,6 +3987,12 @@ def test_tier_1024_experts_routing(
             FP8BlockScaleMoe(fp8_quantization_type=QuantMode.FP8_BLOCK_SCALE_DEEPSEEK),
             id="FP8_Block_DeepSeek",
         ),
+        pytest.param(BF16Moe(), id="BF16xBF16"),
+        pytest.param(FP8PerTensorMoe(), id="FP8_PerTensor"),
+        pytest.param(
+            FP4Moe(quant_mode=QuantMode.FP4_NVFP4_NVFP4),
+            id="NvFP4xNvFP4",
+        ),
     ],
 )
 @pytest.mark.parametrize(
@@ -4005,7 +4011,12 @@ def test_tier_1024_experts_routing(
                 "routed_scaling": 2.5,
                 "has_routing_bias": True,
                 "routing_method_type": RoutingMethodType.DeepSeekV3,
-                "compatible_moe_impls": [FP8BlockScaleMoe],
+                "compatible_moe_impls": [
+                    FP8BlockScaleMoe,
+                    BF16Moe,
+                    FP8PerTensorMoe,
+                    FP4Moe,
+                ],
                 "compatible_intermediate_size": [512],
                 "enable_autotune": False,
             },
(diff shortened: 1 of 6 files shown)


tests/moe/test_trtllm_gen_fused_moe.py
```diff
@@ -3969,6 +3969,114 @@ def test_tier_1024_experts_routing(
...
+                "use_shuffled_weight": False,
+                "layout": WeightLayout.MajorK,
+                "compatible_moe_impls": [FP8BlockScaleMoe],
+            },
+            id="NoShuffle_MajorK",
+        ),
+    ],
+)
+@pytest.mark.parametrize("activation_type", [ActivationType.Swiglu])
+def test_deepseek_ngroup1_block_per_token_routing(
```
amitz-nv, line 4089: Are other quantizations besides FP8 affected by this change? If so, I think it would be safer to include them in the test parameterization, and also include relu^2 non-gated activation which Nemotron 3 Super uses.

jiahanc (approved):
lgtm, thanks for contribution!

tests/moe/test_trtllm_gen_fused_moe.py
amitz-nv: @jiahanc @zhou-yuxin Please address this comment before merge

zhou-yuxin: Fine, I am working on this~

samuellees (approved):
LGTM~

PR #3169, opened 2026-04-24
fix: add jitter to cubin download backoff

### LINKED ISSUE #3164: [Bug][Infra] Intermittent HTTP 403 errors downloading cubin files from Artifactory
### Summary
 Since nightly release #224 (Apr 14, 2026 US evening time), cubin downloads from the NVIDIA Artifactory endpoint are intermittently returning HTTP 403 errors, even though the files are accessible when the URL is opened in a browser. The failures appear consistent across multiple days and affect multiple kernels.
First Occurrence
https://github.com/flashinfer-ai/flashinfer/actions/runs/24434926267/job/71387057610

### Error log:
```
WARNING - cubin_loader.py:119 - flashinfer.jit: Downloading
  https://edge.urm.nvidia.com/artifactory/sw-kernelinferencelibrary-public-generic-local//fmha/trtllm-gen/.cubin: attempt N
  failed: 403 Client Error: Forbidden
  INFO    - cubin_loader.py:125 - flashinfer.jit: Retrying in N seconds...
  ERROR   - cubin_loader.py:128 - flashinfer.jit: Max retries reached. Download failed.
```

### Behavior

  - Files ARE accessible via browser — suggests a transient auth/rate-limit or CDN issue at the Artifactory endpoint, not a
  missing-file problem
  - Failures are intermittent: some downloads within the same run succeed after retries, others exhaust all retries and fail
  - Consistently reproducible across nightly CI runs from Apr 14 to present

  ### Impact

  - Nightly CI failures due to cubin download failures
  - Tests that require downloaded cubins (e.g., TRT-LLM attention tests) fail with `Max retries reached`

pluh-nv:
The retry loop in download_file used a fully deterministic exponential backoff (5s, 10s, 20s). Combined with 4 parallel download threads sharing one session and many concurrent CI runners, this produces lockstep retries that hammer the same CDN edge — a plausible contributor to the intermittent 403s reported in #3164.

Switch to full jitter (uniform[0, base*2^(attempt-1)]) to spread retries across the window and break the thundering-herd pattern.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved download retry logic to use jittered exponential backoff for more resilient recovery from transient failures.
  * Clarified retry termination so reported attempt counts match actual behavior.
  * Randomized retry delays are now logged to aid troubleshooting and visibility.


flashinfer/jit/cubin_loader.py
```diff
@@ -121,8 +122,12 @@ def download_file(
                     )
 
                     if attempt < retries:
-                        backoff_delay = delay * (2 ** (attempt - 1))
-                        logger.info(f"Retrying in {backoff_delay} seconds...")
+                        # Full jitter: uniform[0, base*2^(attempt-1)]. Prevents
+                        # lockstep retries from the 4 parallel download threads
```
yongwww: wondering if it is possible to figure out a reproducible script to verify the issue and confirm the patch works

flashinfer/jit/cubin_loader.py
```diff
@@ -121,8 +122,12 @@ def download_file(
...
+                        # (and from many CI runners) hammering the same CDN edge.
+                        backoff_cap = delay * (2 ** (attempt - 1))
+                        backoff_delay = random.uniform(0, backoff_cap)
```
aleozlx: i think we may try `backoff_delay = backoff_cap + random.uniform(0, backoff_cap)` this lower-bounds the backoff_delay to at least backoff_cap. The exponential part can be important to adapt to dynamic network congestion (hypothesis being we are getting 403 from potentially other actors competing for server bandwidth, causing the problem to depend on time of day, so not just a rate limit thing).

if we do this formula as i mentioned, to compensate for the whole value being potentially 2x larger, we could make the range loop zero-indexed `for attempt in range(retries)` if we wanna be conservative

This way we achieve three things

- jitter added
- exponential lower bound remains
- the magnitude of each delay value is unchanged (being conservative)

aleozlx (approved):
lgtm

aleozlx:
this keeps hitting AOT time outs. i think it could be CI machine dependent

Issue #3170, opened 2026-04-24
DGX Spark (SM121) Current Support Audit


kahyunnam:
# FlashInfer — DGX Spark (SM121) Current Support Audit

**Date:** 2026-04-24
**Scope:** Audit of SM121 (compute capability 12.1, DGX Spark) support across FlashInfer ops, APIs, backends, and testing.
**Goal:** Improve Spark usability on FlashInfer by identifying existing gaps.

**Note: skip to "15. Consolidated Action Items" for TLDR.**

---

## Architecture Context

| Gencode | Target | CUDA Requirement |
|---------|--------|------------------|
| `compute_120a` / `sm_120a` | RTX 5090 (SM120) | CUDA 12.8 |
| `compute_120f` / `sm_120f` | SM12x family (120 + 121) | CUDA 12.9 |
| `compute_121a` / `sm_121a` | DGX Spark (SM121) | CUDA 12.9 |

### SM120 vs SM121: Implications for FlashInfer

Per the [CUDA 13.2 Programming Guide — Compute Capabilities](https://docs.nvidia.com/cuda/archive/13.2.0/cuda-programming-guide/05-appendices/compute-capabilities.html) (Tables 29–33), SM120 and SM121 share a single "12.x" spec column: identical shared memory (99 KB/block, 100 KB/SM), warps (48/SM), registers (64K/SM), tensor core data types (TF32, BF16, FP16, FP8, FP6, FP4, INT8), TMA, and cluster support. [Table 28](https://docs.nvidia.com/cuda/archive/13.2.0/cuda-programming-guide/05-appendices/compute-capabilities.html#compute-capabilities-family-specific-compatibility) confirms `compute_120f` is compatible with **both** CC 12.0 and 12.1. The chip-level differences (SM121 = aarch64 host, unified LPDDR5x, 84 SMs vs SM120's x86_64, discrete GDDR7, up to 170 SMs) do not affect kernel correctness — only SM-count-based heuristics and host toolchain.

### `f` vs `a` compilation targets: performance implications

[...]

PR #3182, opened 2026-04-26
test: align test_fmha_v2_prefill SM gating with is_sm12x_supported

leonardHONG:
## 📌 Description

  The function-level skip in `test_fmha_v2_prefill.py` checks `is_sm120a_supported`, which only matches sm_120, but the actual FMHAv2 dispatch (`prefill.py:4498`, `:4540`) is gated on `is_sm12x_supported` —so SM121 ends up skipped from these tests even though the kernel path supports it.

  This swaps the test gates to `is_sm12x_supported`. The top-level skip message already said `"SM12x"` (only the gate was lagging). The three sub-skips (FP8 / SEPARATE_Q_K_V / SLIDING_WINDOW) previously said `"SM120+"`; they're updated to `"SM12x"` to match what the gate  actually means.

  `sm_120` behavior is unchanged: both helpers evaluate True for sm_120 on CUDA ≥ 12.8 (`is_sm120a_supported`: `major == 12 and minor==0`;`is_sm12x_supported`: `major == 12`, with a CUDA-version branch that collapses to ≥ 12.8 on minor 0).

  Verified on RTX Pro 6000 (sm_120, CUDA 12.9): 233 passed, 2112 skipped,0 failed.

  ## 🔍 Related Issues

  Refs #3170 (Action Item 1 / T1).

  ## 🚀 Pull Request Checklist

  ### ✅ Pre-commit Checks

  - [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your  preferred method).

  ## 🧪 Tests

## Summary by CodeRabbit

* **Tests**
  * Updated GPU support detection logic in attention tests to use updated SM version targeting for test gating and conditional skips.


kahyunnam (approved):
lgtm, thanks!

bkryu (approved):

qsang-nv (approved):

PR #3187, opened 2026-04-27
Add `set_autotune_process_group` to synchronize tactic choice across ranks

### LINKED ISSUE #3186: `AutoTuner.choose_one` picks divergent tactics per rank under TP, deadlocking NCCL symmetric-memory allocations
## Summary

`flashinfer.autotuner.AutoTuner.choose_one` runs entirely independently on every rank, with no cross-rank coordination. GPU timing noise makes different ranks pick different "winning" tactics for the same `(op, runner, profile)` key. When the chosen kernel is then invoked under a `use_symmetric_memory(...)` context (common in TP serving with NCCL symmetric memory), the per-rank tactic mismatch makes the kernel allocate different-sized scratch tensors per rank, and the resulting `ncclCommWindowRegister(NCCL_WIN_COLL_SYMMETRIC)` calls deadlock because the collective requires every rank to call it with the same size in the same order.

The bug is reproducible deterministically on Qwen3-235B + TP=8 + flashinfer + `--enable-symm-mem` in SGLang: CUDA-graph capture hangs at exactly `bs=144, step 30/52`. The straggler rank is non-deterministic (we observed TP2 and TP5 across separate runs), consistent with timing-noise-driven decisions.

This issue has been independently observed and partially worked around downstream (sglang [#18908][sglang-18908], [#23317][sglang-23317]), but the root cause has never been addressed at the autotuner layer.

## Environment

```
GPU         8x NVIDIA B200 (SM100)
CUDA        12.x (lmsysorg/sglang:latest container)
FlashInfer  0.6.7.post3 (also reproduces on 0.6.9 main)
Framework   SGLang (autotune ON, --enable-symm-mem)
Model       Qwen/Qwen3-235B-A22B-Instruct-2507 (94-layer MoE, trtllm_bf16_moe path)
Backend     flashinfer attention + flashinfer_trtllm MoE runner
```

## Reproduction

```bash
docker run -d --gpus all --shm-size 32g --network host \
    --ulimit memlock=-1 --init --name sglang-bench \
    -v <hf-cache>:/root/.cache/huggingface \
    --entrypoint sleep lmsysorg/sglang:latest infinity

docker exec sglang-bench python3 -m sglang.launch_server \
    --model-path Qwen/Qwen3-235B-A22B-Instruct-2507 \
    --tp 8 --attention-backend flashinfer --enable-symm-mem \
    --port 30000
```

Observed (truncated):

```
Capturing batches (bs=152 avail_mem=15.10 GB):  58%|██▌| 30/52 [00:07<00:05]
Capturing batches (bs=144 avail_mem=15.07 GB):  58%|██▌| 30/52 [00:07<00:05]
                                                       ^^^^^^^^^^^^^^^^^^^^^
                                                        progress freezes here
                                                        for hours
```

No exception, no NCCL error, no watchdog. All 8 scheduler procs sit at 99 % CPU (busy NCCL spin) forever.

## Trigger

The hang occurs during CUDA-graph capture when the model forward calls into the trtllm-gen BF16 MoE op (`flashinfer/fused_moe/core.py::trtllm_bf16_moe`) inside a SGLang `use_symmetric_memory(tp_group)` context. SGLang already auto-enables `enable_flashinfer_allreduce_fusion=True` for this config — the MoE compute and the cross-rank allreduce are fused into a single kernel pipeline, so the *kernel-side* communication path is not the problem. The deadlock fires *before the kernel runs*, in the host-side scratch-buffer allocation step:

```
1. AutoTuner.choose_one returns a tactic    (host-side, before kernel launch)
2. Kernel allocates scratch buffers via torch.empty(...) inside
   use_symmetric_memory(...) -> ncclMemAlloc + ncclCommWindowRegister
   -> COLLECTIVE -- must match across ranks   <-- HANG HERE
3. Fused MoE+allreduce kernel runs           (never reached on stragglers)
```

Smaller models (≤ 8B class) finish capture before the divergence window opens, which is why per-commit CI doesn't surface this. The bug requires a model with enough autotune-eligible ops × distinct profiles to fan the per-rank cache out enough that at least one allocation size disagrees.

## Direct measurement of per-rank tactic divergence

We instrumented `AutoTuner.choose_one` with a probe that logs every `profiling_cache[cache_key] = (runner_id, tactic, p)` write per rank:

```python
# inside choose_one, right after the cache-set
self.profiling_cache[cache_key] = (runner_id, tactic, p)
# --- probe ---
import torch.distributed as dist
rank = dist.get_rank() if dist.is_initialized() else "?"
with open(f"/tmp/tactic_log_rank_{rank}.txt", "a") as fh:
    fh.write(f"{custom_op}\t{p.get_opt_shapes()}\trunner={runner_id}\ttactic={tactic}\n")
```

Ran the failing config until the hang, then diffed the 8 per-rank logs.
Result for `flashinfer::trtllm_bf16_moe` across the 13 distinct profiles that got tuned before the hang:

| `num_tokens` (profile leading dim) | distinct tactics across 8 ranks |
|---:|---:|
| 1    | **8** (every rank disagrees) |
| 2    | 6 |
| 4    | 6 |
| 8    | 6 |
| 16   | 7 |
| 32   | 4 |
| 64   | 7 |
| 128  | 7 |
| 256  | 6 |
| 512  | 3 |
| 1024 | 6 |
| 2048 | 2 |
| 4096 | 3 |

**0 of 13 profiles have unanimous agreement across ranks.** For `num_tokens=1`, every one of the 8 ranks picked a different tactic — the maximum possible divergence on 8 ranks. Example:

| Rank | Tactic | Measured time |
|---|---|---|
| 0 | `(32, 59)` | 0.119 ms |
| 5 | `(32, 14)` | 0.119 ms |
| 7 | `(32, 11)` | 0.119 ms |

All three measurements are numerically identical to 3 decimal places — they're all in the "near-tied top cluster" of tactics. Which one wins the local `argmin` race depends entirely on per-rank timing noise.

## py-spy evidence at the hang point

```
TP0..TP4, TP6, TP7  →  torch.distributed.barrier(...)  in pynccl_allocator.py
                       (waiting on the next ncclCommWindowRegister
                        collective)

TP5 (or TP2 in another run, randomised)  →  trtllm_bf16_moe_op
                       (flashinfer/fused_moe/core.py:1301)
                       blocked inside the C++ kernel-launch shim,
                       presumably on its own ncclCommWindowRegister
                       call which has a different size argument from
                       what the other 7 ranks are about to issue
```

The non-determinism of the straggler rank rules out "the bug lives on a specific rank" and instead points directly at "the rank whose noisy timing happened to disagree on this iteration".

## Root cause

`AutoTuner.choose_one` in tune mode (lines ~656–825 of
`flashinfer/autotuner.py` on main):

```python
for r_id, r in enumerate(runners):
    for tac in r.get_valid_tactics(tensors, p):
        time_measured = self._profile_single_kernel(r, tensors, tac, ...)
        if time_measured < min_time:
            min_time = time_measured
            runner_id, tactic = r_id, tac
self.profiling_cache[cache_key] = (runner_id, tactic, p)
```

`grep -n "torch.distributed\|dist\." flashinfer/autotuner.py` returns nothing — the autotuner is fundamentally unaware of the distributed context it's running inside. Every rank is running its own local tournament with its own noisy stopwatch, so different ranks pick different winners whenever multiple tactics are tied within the noise band.

The link from "different tactic" to "different `ncclCommWindowRegister` size" is the kernel's internal scratch shape: each tactic uses a different tile/pipeline configuration, so `[N_experts, tile, hidden]` buffers come out at different sizes. Under `use_symmetric_memory(...)`, those allocations route through SGLang's NCCL-aware allocator plugin, which calls `ncclCommWindowRegister(NCCL_WIN_COLL_SYMMETRIC)` per allocation. The collective deadlocks on size mismatch.

## Expected vs. actual behaviour

**Expected**: under TP, all ranks are computing the same model on the same input, so they should converge on the same kernel tactic for the same shape. The cache should be coherent across ranks.

**Actual**: `profiling_cache` is per-process and per-rank-noisy. Two ranks computing the same forward end up running different kernel binaries.

This is invisible without symmetric memory (regular allreduce tolerates small per-rank speed differences). With symm-mem on the fast path, the divergence becomes a hard collective deadlock.

## Fix direction 

The minimal fix is to add a small opt-in API (`set_autotune_process_group(group)`) that makes `_profile_single_kernel` all reduce the measured time across ranks before the `argmin` comparison. With identical (averaged) inputs, every rank's `argmin` produces the same answer.

## Related

- sglang [#18908][sglang-18908] — same root cause identified months ago (Alison Shao, NVIDIA): *"different ranks may profile different kernel configs, causing NCCL collective divergence"*. The PR took the workaround route (disable symm-mem during autotune + capture) and was closed without merging after reviewer feedback that disabling symm-mem during capture defeats the purpose.
- sglang [#23317][sglang-23317] — independent re-attempt of the same workaround; superseded by the autotuner-side fix.
- sglang [#20153][sglang-20153] — adjacent issue: per-comm-group symmetric memory pool isolation.
- vLLM [#33996][vllm-33996] — RFC for NCCL symmetric memory option; affected the same way once enabled.

[sglang-18908]: https://github.com/sgl-project/sglang/pull/18908
[sglang-23317]: https://github.com/sgl-project/sglang/pull/23317
[sglang-20153]: https://github.com/sgl-project/sglang/pull/20153
[vllm-33996]: https://github.com/vllm-project/vllm/issues/33996

thanhhao98:
Fixes #3186 (full bug context, reproduction, py-spy and per-rank tactic-divergence measurement live in the issue).

## Summary

Adds an opt-in API to `flashinfer.autotuner`:

- `set_autotune_process_group(group: Optional[ProcessGroup]) -> None`
- `get_autotune_process_group() -> Optional[ProcessGroup]`

When a process group is set, `_profile_single_kernel` all-reduces the measured kernel timing across that group before the autotuner's `argmin`. Every rank therefore sees the same (mean) timing for a given `(runner, tactic, profile)` combination and picks the same tactic. With the group unset (default), behaviour is unchanged.

## Why isn't this already covered by AllReduce fusion?

A reasonable first reaction: SGLang already auto-enables FlashInfer's allreduce fusion (`enable_flashinfer_allreduce_fusion=True`) on SM100 for Qwen3MoE, so the MoE compute and the cross-rank allreduce are *already* fused into a single kernel pipeline. Why does fusion not prevent the hang?

Because the deadlock fires **before the fused kernel runs**, in the host-side scratch-buffer allocation step:

```
1. Autotune picks tactic                            ← host-side, before any kernel launch
2. Allocate scratch buffers via torch.empty(...)    ← inside use_symmetric_memory(...)
                                                       → ncclMemAlloc + ncclCommWindowRegister
                                                       → COLLECTIVE, must match across ranks
                                                       ← ranks deadlock here
3. Run fused MoE+allreduce kernel                   ← never reached on the straggler rank
```

Allreduce fusion fuses step 3 (kernel pipeline), not step 2 (host allocation). Different tactics chosen in step 1 produce different scratch shapes in step 2, and `ncclCommWindowRegister` requires identical sizes across ranks or it deadlocks. The fusion stays on; we just need step 1 to produce the same answer on every rank.

A more aggressive design — making `trtllm_bf16_moe` manage its own scratch internally without ever calling `torch.empty`, and pre-registering symmetric input/output buffers up front — would also close the gap, but it's a much larger refactor (Python + C++ + CUDA launcher API). This PR takes the cheaper path: keep the existing allocation contract and synchronise the tactic choice at its source.

## Design

```python
# flashinfer/autotuner.py

_tune_process_group: Optional["torch.distributed.ProcessGroup"] = None

def set_autotune_process_group(group): ...
def get_autotune_process_group(): ...

# Inside AutoTuner._profile_single_kernel, right after pure_profile(...):
if _tune_process_group is not None:
    import torch.distributed as dist
    time_tensor = torch.tensor([avg_time], dtype=torch.float64)
    dist.all_reduce(time_tensor, op=dist.ReduceOp.SUM, group=_tune_process_group)
    avg_time = time_tensor.item() / dist.get_world_size(_tune_process_group)
```

Rationale:

- **Mean reduction**, not `MAX` or `MIN`. Mean smooths out GPU noise symmetrically. `MAX` overweights the slowest rank; `MIN` overweights the fastest. `SUM`+divide is chosen over `AVG` for backend portability — `AVG` is NCCL-only, `SUM` works on both gloo and NCCL.
- **CPU tensor, gloo subgroup recommended.** The sync must not interfere with CUDA streams. NCCL groups also work but serialise against compute on the device. The comparison value is a scalar, so overhead is negligible either way.
- **Opt-in via explicit setter.** Zero change when the hook is not set (the default). Single-GPU users and multi-GPU users not on symmetric-memory paths pay nothing.
- **No rank-0 asymmetry.** All ranks contribute their timing, all ranks agree on the mean, every rank's `argmin` returns the same answer.

## Correctness

- **Caller contract — identical tuning loop on every rank.** The reduction adds one collective per tactic measurement, so every rank must enter `_profile_single_kernel` the same number of times in the same order. Concretely: `runner.get_valid_tactics(...)` must return the same list on every rank (no per-rank arch-conditional filtering or rank-dependent early exits), and callers must drive autotune with the same input shapes on every rank. Violating this deadlocks the reduction itself — a failure mode that looks identical to the symm-mem hang this API exists to prevent — so the docstring calls it out explicitly.
- **Collective cardinality preserved on the exception path.** If `pure_profile` raises on one rank (OOM, ptxas mismatch, illegal instruction in a bad tactic, etc.), that rank must still participate in the per-tactic reduce, or peers are left blocked waiting for a matching call that will never come. The implementation catches exceptions inside `_profile_single_kernel`, sets `avg_time = float("inf")`, runs the reduce unconditionally, then re-raises so `choose_one`'s outer handler (logging, stats, OOM fallback) still runs. `inf + finite = inf`, so the failing tactic is correctly disqualified on every rank.
- **Caches stay coherent.** With identical reduced measurements, every rank reaches the same `(runner_id, tactic)`, populates the same `profiling_cache` keys with the same values, and later cache lookups return the same answer on all ranks.

## Overhead

For a Qwen3-235B autotune run (~10 candidate tactics × ~13 shape buckets × handful of ops): roughly a few hundred 8-byte CPU all-reduces on a gloo subgroup. Sub-millisecond per reduction, invisible against profile-measurement time. Verified end-to-end: total autotune wall-clock unchanged within noise.

## API Compatibility

Purely additive. No existing function signature or behaviour changes. Callers that never call `set_autotune_process_group` observe no difference.

## Tests

`tests/autotuner/test_autotuner_distributed.py` (CPU-only, no GPU/NCCL required). The production `_profile_single_kernel` needs CUDA, so coverage is split into source-level AST guards on the real functions plus a runtime gloo check of the reduce *pattern*:

- **`test_profile_single_kernel_preserves_collective_cardinality_on_exception`** — AST guard on the real `_profile_single_kernel`: `pure_profile(...)` is inside a `try/except` that sets `avg_time = float("inf")`, `dist.all_reduce` appears AFTER the `try/except` (runs on both success and caught-exception paths), and a `raise` follows so `choose_one`'s outer handler still sees the failure.
- **`test_choose_one_oom_preserves_cardinality_under_group`** — AST guard on the real `choose_one`: the per-tactic `OutOfMemoryError` handler gates on `_tune_process_group` and disqualifies the tactic with `inf` instead of unconditionally re-raising (which would early-return `runners[0], -1` and desync the per-tactic all-reduce).
- **`test_exception_path_does_not_deadlock_the_reduce`** — gloo 2-rank runtime check of the `try/except`→`inf`→all-reduce *pattern* (a standalone reimplementation, **not** a call into `_profile_single_kernel`, which needs CUDA; the AST guards above cover the production path). Rank 0 fails, rank 1 succeeds; both must still reach the reduce and observe `(inf + 2) / 2 == inf`, else rank 1 times out via `pytest.mark.timeout`.

Existing autotuner tests pass unchanged.

## Validation

End-to-end on the configuration that #3186 reproduces (Qwen3-235B + TP=8 + flashinfer + `--enable-symm-mem`, on `colossus_b200_1`):

| Quantity | Before fix | After fix |
|---|---|---|
| Cross-rank tactic agreement (instrumented) | 0 / 13 profiles | **13 / 13 profiles, byte-identical caches across ranks** |
| CUDA-graph capture | hangs at `bs=144, 30/52` | 13 s, completes |
| gsm8k accuracy (200 questions) | N/A | 0.975 |
| `SGLANG_DEBUG_SYMM_MEM=1` warnings under load | N/A | 0 (symm-mem genuinely active at replay) |

## Downstream wire-up (reference)

A downstream framework using flashinfer under a TP process group needs to call the setter around its autotune block. Example for SGLang:

```python
from flashinfer.autotuner import autotune, set_autotune_process_group

set_autotune_process_group(tp_group.cpu_group)  # cpu_group uses gloo
try:
    with autotune(True):
        run_warmup_forwards()
finally:
    set_autotune_process_group(None)
```

Guarded behind `try: import` on the caller side keeps that code compatible with both patched and unpatched flashinfer versions.

## Summary by CodeRabbit

* **New Features**
  * Added public APIs to configure a distributed process group for autotuning, enabling synchronized shared profiling measurements across ranks.

* **Bug Fixes**
  * Improved distributed autotuning reliability so collective timing reductions still complete when per-kernel profiling fails, while the original profiling error is still surfaced correctly.

* **Tests**
  * Added multi-rank distributed tests that verify synchronization, correct handling of profiling exceptions, and deadlock prevention.


thanhhao98:
### Re-verified end-to-end on latest sglang after the rebase (8×B200)

Controlled before/after on `Qwen3-235B-A22B-Instruct-2507`, `--tp 8 --attention-backend flashinfer --enable-symm-mem`. Stack: this branch (flashinfer 0.6.14 + fix, built from source) + sglang latest `main` with the downstream wire-up. **Only variable = whether `set_autotune_process_group(tp_group.cpu_group)` is called** (the fix is opt-in, so flashinfer is byte-identical in both runs).

| | BEFORE (wire-up off) | AFTER (wire-up on) |
|---|---|---|
| CUDA-graph capture | **hangs** after autotune — all 8 GPUs 0% util, 7 ranks spin-wait in `hrtimer_nanosleep`, one rank `R(running)` | **completes 52/52 in 55 s**, server serves |
| Per-rank scratch alloc | **divergent** — straggler rank holds 167423 MiB vs 163769 on the other 7 (mismatched size → symm-mem collective deadlock) | uniform |
| Cross-rank tactic agreement | divergent (hung before autotune caches were saved) | **8/8 per-rank caches byte-identical, 66/66 profiles agree** |
| GSM8K (200 q) | n/a (never serves) | **0.965** |

The deadlock reproduces on current `main` of both projects, and this PR's tactic-sync resolves it: every TP rank converges on the same tactic, so the `ncclCommWindowRegister` scratch allocation matches and CUDA-graph capture completes with symmetric memory active.

aleozlx:
pls help review @qiching

flashinfer/autotuner/autotuner.py
```diff
@@ -1485,12 +1521,59 @@ def _run_kernels():
...
+        # fallback).
+        profile_exc: Optional[BaseException] = None
+        try:
+            with _profile_measurement_scope():
+                # warm up, no timing
+                for _ in range(self.warmup):
+                    runner(input_tensor_batches[-1], tactic=tactic, **kwargs)
+
+                avg_time = pure_profile(stream, self.repeat)
+        except BaseException as e:  # noqa: BLE001
```
qiching, line 1753: i think here just handle the generic exception path correct, but not OOM. `choose_one` reraises OutOfMemoryError past the tactic loop and the outer handler does return runners[0], -1, so an OOM rank leaves `choose_one` while peers advance to the next tactic and hang on `all_reduce`. OOM is per-rank, like independent TP processes so the "identical loop" contract cannot prevent it. 
so i will suggest: `when _tune_process_group` is not None, treat OOM like other exceptions (empty_cache(), record inf, continue) instead of early-returning.

flashinfer/autotuner/autotuner.py
```diff
@@ -726,6 +726,42 @@ def is_in_profile_measurement() -> bool:
...
+) -> None:
+    """All-reduce (mean) per-tactic profile timings across ``group`` so every
+    rank's ``argmin`` picks the same tactic.
+
+    Without it, GPU timing noise makes ranks diverge on tactic choice, which
+    deadlocks NCCL symmetric-memory allocation (``NCCL_WIN_COLL_SYMMETRIC``).
+    Prefer a CPU (``gloo``) subgroup; a NCCL group also works. ``None``
+    disables (default); not thread-safe.
+
+    Caller contract: every rank must enter ``_profile_single_kernel`` the same
```
qiching, line 849: Two more items belong here, both diverge per-rank reduce counts: 1) skip_ops must match across ranks, `choose_one` returns before the loop for skipped ops; 2) `profiling_cache` must be identical at entry, a hit on one rank vs. miss on another diverges the count, so the group must be set from the first `choose_one` with identical starting caches.

qiching (changes requested):

tests/autotuner/test_autotuner_distributed.py
```diff
@@ -0,0 +1,209 @@
...
+    has_reraise = any(
+        isinstance(n, ast.Raise) for stmt in post_try for n in ast.walk(stmt)
+    )
+    assert has_reraise, (
+        "the caught exception must be re-raised after the reduce so "
+        "choose_one's outer handler still records the failure"
+    )
+
+
+def _gloo_worker_exception_path(
```
qiching: i just to set expectations: this worker reimplements the try or except to inf to `all_reduce` pattern rather than calling the `real _profile_single_kernel` . So it validates that gloo behaves as assumed for a copy of the logic, not the production path, so the AST test is what actually guards the production function. That's a reasonable split given the CUDA dependency; I'd just phrase the PR text accordingly so it doesn't read as an e2e test of `_profile_single_kernel`.

qiching (changes requested):

thanhhao98:
Thanks @qiching — all three are spot-on. Addressed in the latest commit:

**[OOM cardinality]** resolved; the OOM *outside* `_profile_single_kernel` (input synthesis / `do_preparation`) can still hit the outer early-return — documented in the caller-contract rather than fixed, since there's no clean in-loop continuation there.

**[Caller contract]** resolved 

**[Test framing]** resolved

aleozlx (commented):
code review: looks good

waiting for CI run

aleozlx:
sorry the above comment of mine is posted on the wrong PR..

thanhhao98:
Sure @qiching  :D I updated as your comment, pls help to review again when you have time. Thanks!

aleozlx (approved):

aleozlx:
need extra approval cc @qiching

qiching (approved):
LGTM

aleozlx:
@thanhhao98 could help address merge conflict ? thanks

aleozlx:
🤖 Merged latest `main` into this branch to resolve conflicts and pushed the merge commit (GPG-signed). **1 conflicted file, 1 hunk** — resolved and verified (pre-commit clean: ruff / ruff-format / mypy / merge-conflict check; `py_compile` OK).

Root cause: `main` moved `flashinfer/autotuner.py` into a package (`flashinfer/autotuner/autotuner.py`). Git handled the rename; the feature logic (the `dist.all_reduce` in `_profile_single_kernel`) auto-merged cleanly. Only the definition block conflicted.

**Resolution**

- **`flashinfer/autotuner/autotuner.py`** — two independent additions collided at the same spot: this PR's `set_autotune_process_group` / `get_autotune_process_group`, and main's new `ProfilingCacheKey` dataclass. → Kept both.
- **`flashinfer/autotuner/__init__.py`** — re-exported `set_autotune_process_group` / `get_autotune_process_group` from the new package `__init__` (import block + `__all__`) to match main's packaging convention. This isn't a git conflict (the file didn't exist on the PR side), but without it the new public API wouldn't resolve as `flashinfer.autotuner.set_autotune_process_group`.
- **Typing import** — re-added `typing.Optional`. main modernized to `X | None`, but this PR's forward-ref annotations (`Optional["torch.distributed.ProcessGroup"]`) are in runtime-evaluated positions (module-level var, function param, return annotation) with no `from __future__ import annotations`, so `"torch..." | None` would raise `TypeError` at import. `Optional[...]` is the correct form here.

nvpohanh:
cc @leejnau since this may influence how SGLang autotunes FlashInfer comm ops

PR #3192, opened 2026-04-27
Yanqinz/fix cudnn sm120 nan

yanqinz2:
## 📌 Description

Aligns the cuDNN GEMM runners with the AutoTuner's bucket bookkeeping, fixes a rare NaN/Inf in FP4 GEMM caused by mismatched tactic indices, and adds a few production-quality plumbing improvements.

1. Root-cause fix for FP4 GEMM NaN/Inf
Symptom: flashinfer.mm_fp4(..., backend="cudnn") produced sparse NaN/Inf (typically a few values out of millions) at specific shapes (e.g. M=2001, K=2688, N=10304) under sglang prefill workloads.

Root cause: CudnnFp4GemmRunner._get_override_graph (and the BF16 analog) computed the cuDNN override-shape cache_m using last_positive_power_of_2(actual_m) (a floor-to-power-of-2 function), while a comment claimed the call was a ceiling. The function flashinfer was actually calling is largest 2^k <= x, so for a non-power-of-2 actual_m (e.g. 2001) it produced cache_m=1024, smaller than the runtime M.

The cuDNN graph cache is keyed on cache_m. The AutoTuner's plan-index lookup is keyed on the bucket the runtime size maps to (map_to_tuning_buckets). When those two bucket functions disagreed — which they did, in subtle ways — a tactic that was profiled on graph A could be silently applied to graph B at runtime, where the same plan index points to a different cuDNN engine. This produced rare NaN/Inf at the shapes where the two functions diverged.

Fix: Make the cuDNN runner share the AutoTuner's bucket function.

flashinfer/autotuner.py: new public method AutoTuner.get_effective_map_to_tuning_buckets(tuning_config, spec_idx=0) that returns the currently effective map_to_tuning_buckets for a given TuningConfig, applying any active with autotune(tuning_buckets=..., round_up=...) overrides.
flashinfer/gemm/gemm_base.py:
_cudnn_gemm_fp4_runner and _cudnn_gemm_bf16_runner now take a m_bucket_mapper argument; the inner _get_override_graph is converted from @staticmethod to an instance method and computes cache_m = self._m_bucket_mapper(actual_m).
mm_fp4 and bf16_gemm_sm100 query the active mapper via tuner.get_effective_map_to_tuning_buckets(...) and inject it into the runner factory.
After the fix, the cuDNN graph used at runtime is always the same graph the AutoTuner profiled tactics against, regardless of whether the user is in default mode, autotune(round_up=True), autotune(tuning_buckets=...), or any nested combination.

2. Stride-flag profile/runtime alignment for BF16 GEMM
build_cudnn_gemm_bf16_graph_override_shape bakes is_a_k_major / is_b_k_major into the graph cache key (the build-time stride layout differs between row- and column-major). The AutoTuner profile path uses torch.rand-synthesized tensors which are always contiguous (k-major), but the runtime tensor a may be column-major in real workloads. That made profile time and runtime build different cuDNN graphs from the cache's perspective and apply tactic indices across them.

Fix: Capture is_a_k_major / is_b_k_major from the real input tensors at runner construction time (in bf16_gemm_sm100) and store them on the runner instance. _get_override_graph now reads self._is_*_k_major instead of the local tensor's stride, so profile and runtime always build the same graph.

3. cuDNN graph cache is now bounded + manually clearable
The 8 build_cudnn_gemm_*_graph_* helpers were a mix of @functools.cache (unbounded, 6 uses) and @functools.lru_cache(maxsize=1024) (2 uses). On a long-running serving process with diverse shapes (different batch sizes, model layers, sequence lengths), the unbounded cache grows monotonically.

Fix:

All 8 helpers now use @functools.lru_cache(maxsize=1024). Cold shapes are evicted automatically; hot shapes stay resident. GPU memory usage is bounded.
New public function flashinfer.gemm.clear_cudnn_graph_cache() for the cases LRU cannot handle: development hot-patching, deliberate cuDNN library version switching, controlled GPU memory release, and unit-test cleanup. Calls .cache_clear() on every cuDNN graph builder.
The other @functools.cache decorators (JIT module loaders) are unchanged — those have a small bounded number of unique entries (~tens, by hardware × dtype combinations) and don't need eviction.

4. Out-of-range autotune cache miss warning
When a user ran with autotune(False, cache="cfg.json") for inference but the runtime input shape was outside the tuning bucket range used during profiling, the lookup silently fell back to tactic=-1 and the user had no way to notice they were missing the tuned tactic.

Fix (flashinfer/autotuner.py): in the non-tuning branch of choose_one, when search_cache returns a fallback and the user has loaded autotune data (file configs or in-memory profiling cache), emit a logger.warning once per (custom_op, profile-shape signature) pair pointing the user at expanding tuning_buckets / max_num_tokens. New _logged_cache_miss_oor: Set tracks dedup; clear_cache() clears it so a re-tune resets the deduplication.

5. Cache file metadata: include cudnn-frontend version
AutoTuner.save_configs writes a _metadata block to JSON; load_configs rejects (and refuses to overwrite) caches whose metadata doesn't match the current environment. Previously the check covered flashinfer_version, cuda_version, cublas_version, cudnn_version (backend), and gpu — but not cudnn-frontend (the Python wrapper) version.

policy=ALL plan_index ordering can shuffle when only the frontend is upgraded but the backend is the same. A cache saved under one frontend can have stale indices under another with no warning.

Fix (_collect_metadata): also record cudnn_frontend_version (from cudnn.__version__, falls back to "unknown" if cudnn is not importable). Caches saved by older flashinfer versions that predate this metadata field will be rejected on load and need to be regenerated — which is the correct behavior: those caches were never validated against frontend version and may already be stale.

load_configs docstring updated to enumerate the metadata fields and call out the migration cost.

## 🔍 Related Issues

Recently, sglang reported a silent nan encountered during running NVIDIA-Nemotron-3-Nano-30B-A3B-NVFP4 on sm120/sm121

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * Expose effective tuning-bucket mapping from the autotuner.
  * Ability to clear cuDNN graph-builder and tuning caches (also resets out-of-range warning state).
  * Broader BF16 backend eligibility on newer GPU architectures.

* **Bug Fixes**
  * Improved fallback warnings for operations outside tuned ranges (deduplicated).
  * Override-shape execution now respects runtime layout and active bucket mapping to avoid stale plan mismatches.

* **Tests**
  * GEMM tests now exercise additional cuDNN configurations.

* **Chores**
  * Bounded internal caches to limit growth and reduce resource use.

### DIFF
--- a/flashinfer/autotuner.py
+++ b/flashinfer/autotuner.py
@@ -1051,18 +1051,35 @@ class AutoTuner:
 
                     # If the user has loaded an autotune cache (via
                     # ``autotune(cache=...)``) or warmed up profiling
-                    # results in the current process but this particular
-                    # input shape still falls back, that almost always
-                    # means the runtime input is *outside the tuned
-                    # range* (e.g. ``max_num_tokens`` was 2048 during
-                    # tuning, runtime sees 4000).  This is a silent
-                    # perf regression: the fallback tactic is correct
-                    # but not optimised for this shape.  Warn once per
+                    # results *for this specific custom_op* in the
+                    # current process but this particular input shape
+                    # still falls back, that almost always means the
+                    # runtime input is *outside the tuned range* (e.g.
+                    # ``max_num_tokens`` was 2048 during tuning,
+                    # runtime sees 4000).  This is a silent perf
+                    # regression: the fallback tactic is correct but
+                    # not optimised for this shape.  Warn once per
                     # unique (op, profile-signature) pair so the user
                     # can extend ``tuning_buckets`` / re-tune.
-                    has_tune_data = bool(self._file_configs) or bool(
-                        self.profiling_cache
+                    #
+                    # The has-tune-data check is intentionally scoped
+                    # to ``custom_op`` -- unrelated ops with their own
+                    # cache entries should not trigger a "tuned range
+                    # exceeded" warning for an op that was never tuned
+                    # in the first place.  ``profiling_cache`` keys are
+                    # tuples ``(custom_op, ..., extras)`` and
+                    # ``_file_configs`` keys are
+                    # ``str((custom_op, runner_class_name, profile))``,
+                    # so we filter by ``custom_op`` on each.
+                    op_has_profiling = any(
+                        isinstance(k, tuple) and len(k) > 0 and k[0] == custom_op
+                        for k in self.profiling_cache
                     )
+                    file_key_op_prefix = f"({repr(custom_op)}, "
+                    op_has_file = any(
+                        k.startswith(file_key_op_prefix) for k in self._file_configs
+                    )
+                    has_tune_data = op_has_profiling or op_has_file
                     if has_tune_data:
                         try:
                             signature = self._find_nearest_profile(
--- a/flashinfer/autotuner.py
+++ b/flashinfer/autotuner.py
@@ -148,7 +148,28 @@ def _get_cublas_version() -> str:
 
 
 def _collect_metadata() -> Dict[str, str]:
-    """Collect environment metadata that can affect tactic-to-kernel mappings."""
+    """Collect environment metadata that can affect tactic-to-kernel mappings.
+
+    Tactics in flashinfer's autotune cache are stored as plan indices into
+    cuDNN's ``policy=ALL`` plan list (or backend-internal kernel ids for
+    other backends).  Anything that can shuffle that ordering must be
+    captured here so ``load_configs`` rejects a cache that no longer
+    matches the runtime environment.
+
+    Specifically tracked:
+
+        * ``flashinfer_version``  -- our own bucketing / ordering changes
+        * ``cuda_version``        -- CUDA driver/runtime ABI
+        * ``cublas_version``      -- cuBLAS plan availability inside cuDNN
+        * ``cudnn_version``       -- cuDNN **backend** version
+        * ``cudnn_frontend_version`` -- cuDNN-frontend Python wrapper
+                                        version (independent of backend);
+                                        plan_index ordering can change
+                                        when only the frontend is updated
+                                        but the backend is not
+        * ``gpu``                 -- device name (different SM may have
+                                     different available engines)
+    """
     meta: Dict[str, str] = {}
     meta["flashinfer_version"] = _flashinfer_version
     meta["cuda_version"] = getattr(torch.version, "cuda", None) or "unknown"
@@ -157,6 +178,17 @@ def _collect_metadata() -> Dict[str, str]:
         meta["cudnn_version"] = str(torch.backends.cudnn.version())
     except Exception:
         meta["cudnn_version"] = "unknown"
+    try:
+        # cudnn-frontend is an optional dependency; failing to import is
+        # not fatal -- we just record "unknown" and let the metadata
+        # check fall back to the backend-version field.
+        import cudnn as _cudnn_frontend
+
+        meta["cudnn_frontend_version"] = getattr(
+            _cudnn_frontend, "__version__", "unknown"
+        )
+    except Exception:
+        meta["cudnn_frontend_version"] = "unknown"
     try:
         meta["gpu"] = torch.cuda.get_device_name(torch.cuda.current_device())
     except Exception:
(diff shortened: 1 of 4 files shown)


flashinfer/gemm/__init__.py
```diff
@@ -91,6 +92,7 @@
     "mm_M1_16_K7168_N256",
     "tinygemm_bf16",
     "is_cudnn_override_shape_available",
+    "clear_cudnn_graph_cache",
```
dhiraj113: Are all these apis of gemm now?

yanqinz2: It should be debug/develop use only. I will remove it from public api, and add corresponding docs.

flashinfer/autotuner.py
```diff
@@ -964,6 +1036,40 @@ def choose_one(
...
+                    # range* (e.g. ``max_num_tokens`` was 2048 during
+                    # tuning, runtime sees 4000).  This is a silent
+                    # perf regression: the fallback tactic is correct
+                    # but not optimised for this shape.  Warn once per
+                    # unique (op, profile-signature) pair so the user
+                    # can extend ``tuning_buckets`` / re-tune.
+                    has_tune_data = bool(self._file_configs) or bool(
+                        self.profiling_cache
+                    )
+                    if has_tune_data:
```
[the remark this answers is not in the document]
yanqinz2: fixed

flashinfer/gemm/gemm_base.py
```diff
@@ -1766,6 +1786,50 @@ def is_cudnn_override_shape_available() -> bool:
...
+        build_cudnn_gemm_mxfp8_graph_override_shape,
+        build_cudnn_gemm_with_per_tensor_q_graph,
+        build_cudnn_gemm_with_per_tensor_q_graph_override_shape,
+        build_cudnn_gemm_bf16_graph,
+        build_cudnn_gemm_bf16_graph_override_shape,
+        create_cudnn_execution_plans_mxfp8_gemm,
+    )
+    for fn in cached_builders:
+        if hasattr(fn, "cache_clear"):
+            fn.cache_clear()
```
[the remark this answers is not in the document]
yanqinz2: fixed

flashinfer/gemm/gemm_base.py
```diff
@@ -3029,24 +3093,76 @@ def _cudnn_gemm_bf16(
...
+                else map_to_hybrid_bucket_uncapped
+            )
+            # Default to k-major (the convention torch.rand-synthesized
+            # profile tensors use) when caller didn't specify.
+            self._is_a_k_major = True if is_a_k_major is None else is_a_k_major
+            self._is_b_k_major = True if is_b_k_major is None else is_b_k_major
+
+        def _get_override_graph(self, a, b, bias, out):
+            a_shape, _a_stride = _get_bf16_3d_shape_stride(a)
+            b_shape, _b_stride = _get_bf16_3d_shape_stride(b)
```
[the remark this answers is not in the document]
yanqinz2: fixed

Kh4L (approved):
LGTM and tested on my end

flashinfer/gemm/gemm_base.py
```diff
@@ -1994,6 +2014,78 @@ def is_cudnn_override_shape_available() -> bool:
...
+    """
+    cached_builders = (
+        build_cudnn_gemm_fp4_graph,
+        build_cudnn_gemm_fp4_graph_override_shape,
+        build_cudnn_gemm_mxfp8_graph_override_shape,
+        build_cudnn_gemm_with_per_tensor_q_graph,
+        build_cudnn_gemm_with_per_tensor_q_graph_override_shape,
+        build_cudnn_gemm_bf16_graph,
+        build_cudnn_gemm_bf16_graph_override_shape,
+        create_cudnn_execution_plans_mxfp8_gemm,
```
dhiraj113, line 2075: Can all these be removed from `flashinfer/gemm/__init__.py` now?

Kh4L (approved):

dhiraj113 (approved):
Approving the change as it fixes a customer bug, which has been verified by the customer.

flashinfer/gemm/gemm_base.py
yanqinz2: Yes

nv-yunzheq (approved):
Approve given multiple approvals already

PR #3205, opened 2026-04-29
build: add sccache-backed jit-cache builds and AOT diagnostics

dierksen:
## Description

This PR adds sccache-backed compiler caching for the FlashInfer jit-cache/AOT build paths and folds in the AOT memory diagnostics work from #3204.

The main pieces are:

- Add compiler launcher support in the JIT build generation so `nvcc` and host compiler invocations can be wrapped by `sccache` or another launcher.
- Install and configure `sccache` in the release/nightly jit-cache wheel builds and the PR AOT build/import jobs.
- Use the shared S3 cache in read-write mode when AWS credentials are available, and fall back to anonymous read-only mode for PR jobs without credentials.
- Print `sccache --show-stats` at the end of the relevant jobs so cache hit rates and compile behavior are visible in CI logs.
- Keep `FLASHINFER_NVCC_THREADS` supported, but default it back to `1` after CI showed higher nvcc internal threading was slower for this workload.
- Increase build-level parallelism through `MAX_JOBS`, bounded by CPU count and an AOT memory budget so the larger CUDA builds do not overrun the runner.
- Add AOT memory monitoring/report files around install, wheel build, import/config, and module verification steps to diagnose OOMs and runner shutdowns.
- Clean stale jit-cache build artifacts before building and pick the newest wheel from `dist/`.
- Add tests for the nvcc flag generation, launcher-compatible depfile/debug flags, and build regeneration behavior.

## Context

This PR supersedes #3204. The memory monitor and AOT `MAX_JOBS` safety work from #3204 are included here, but the final script shape is different because it is integrated with the sccache setup and the later threading/concurrency experiments. Merging #3204 separately would still create a conflict in `scripts/task_test_jit_cache_package_build_import.sh`.

The current direction is to use sccache for the big rebuild/retry win while keeping nvcc internal threading conservative. The PR still allows `FLASHINFER_NVCC_THREADS` to be overridden for experiments, but the default path lets ninja drive parallelism through `MAX_JOBS`.

## CI Notes

The latest PR Test run completed successfully after the infrastructure-triggered AOT rerun path. The Release workflow's jit-cache jobs completed and showed useful sccache hit rates; the remaining Release failure was in `build-flashinfer-cubin` while downloading cubins, which appears separate from the sccache/AOT build changes.

## Related

Supersedes #3204.

## Performance comparison

This looks at the mean and median time to complete each AOT build type over the last 48 hours compared with the most recent 2 builds on this branch.

  | job | 48h mean | 48h median | [#4214](https://github.com/flashinfer-ai/flashinfer/actions/runs/25402485487)  | #4214 vs median | [#4284](https://github.com/flashinfer-ai/flashinfer/actions/runs/25457504611) | #4284 vs median |
  |---|---:|---:|---:|---:|---:|---:|
  | x64 cu126 | 1:09:34 | 51:55 | 42:28 | 18.2% faster | 31:36 | 39.1% faster |
  | x64 cu128 | 1:45:24 | 1:24:04 | 1:09:46 | 17.0% faster | 45:04 | 46.4% faster |
  | x64 cu129 | 2:36:17 | 2:25:29 | 2:12:39 | 8.8% faster | 59:50 | 58.9% faster |
  | x64 cu130 | 2:27:16 | 2:04:49 | 1:48:57 | 12.7% faster | 59:34 | 52.3% faster |
  | arm64 cu126 | 1:23:52 | 1:30:43 | 1:13:56 | 18.5% faster | 50:43 | 44.1% faster |
  | arm64 cu128 | 2:12:08 | 2:25:56 | 2:04:47 | 14.5% faster | 1:12:37 | 50.2% faster |
  | arm64 cu129 | 3:45:38 | 4:05:35 | 3:38:02 | 11.2% faster | 1:31:15 | 62.8% faster |
  | arm64 cu130 | 3:34:21 | 3:53:18 | 3:21:49 | 13.5% faster | 1:34:48 | 59.4% faster |


scripts/build_flashinfer_jit_cache_whl.sh
```diff
@@ -8,18 +8,32 @@ echo "=========================================="
...
-  MAX_JOBS=1
-elif (( NPROC < MAX_JOBS )); then
-  MAX_JOBS=$NPROC
+
+# Set NVCC_THREADS to match number of gencode targets (capped at 8), unless already set
+NUM_ARCHS=$(echo "${FLASHINFER_CUDA_ARCH_LIST}" | wc -w)
+NVCC_THREADS=${FLASHINFER_NVCC_THREADS:-${NUM_ARCHS}}
+if (( NVCC_THREADS > 8 )); then NVCC_THREADS=8; fi
+if (( NVCC_THREADS > NPROC )); then NVCC_THREADS=${NPROC}; fi
+if (( NVCC_THREADS < 1 )); then NVCC_THREADS=1; fi
```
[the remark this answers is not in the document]
dierksen: --no-isolation fails here. I previously enabled it but had to revert.

flashinfer/jit/cpp_ext.py
```diff
@@ -91,6 +93,113 @@ def is_cuda_version_at_least(version_str: str) -> bool:
...
+    )
+
+
+def get_nvcc_parallelism_flags(cuda_version: Optional[Version] = None) -> List[str]:
+    """Build nvcc flags controlled by FlashInfer parallelism environment variables."""
+    threads = get_nvcc_threads()
+    flags = [f"--threads={threads}"]
+    if threads > 1 and should_use_nvcc_jobserver(cuda_version):
+        flags.append("--jobserver")
+    return flags
```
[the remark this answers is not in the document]
dierksen: This is wrong. The flag was introduced in 13.0.

flashinfer/jit/core.py
```diff
@@ -431,7 +429,7 @@ def gen_jit_spec(
         cflags.insert(0, "-std=c++17")
 
     cuda_cflags = [
-        f"--threads={os.environ.get('FLASHINFER_NVCC_THREADS', '1')}",
+        *get_nvcc_parallelism_flags(),
```
yongwww, line 432: for curiosity, why do we need this change?

flashinfer/jit/cpp_ext.py
```diff
@@ -91,6 +91,37 @@ def is_cuda_version_at_least(version_str: str) -> bool:
...
+
+    return int_value
+
+
+def get_nvcc_threads() -> int:
+    """Return the requested number of nvcc worker threads."""
+    return _get_positive_int_env("FLASHINFER_NVCC_THREADS", 1)
+
+
+def get_nvcc_parallelism_flags(cuda_version: Optional[Version] = None) -> List[str]:
```
yongwww: if get_nvcc_parallelism_flags is needed. then we can consider merging _get_positive_int_env and get_nvcc_threads into get_nvcc_parallelism_flags. The two helpers are only used here

scripts/build_flashinfer_jit_cache_whl.sh
```diff
@@ -8,18 +8,41 @@ echo "=========================================="
...
-  MAX_JOBS=1
-elif (( NPROC < MAX_JOBS )); then
-  MAX_JOBS=$NPROC
+
+NVCC_THREADS=${FLASHINFER_NVCC_THREADS:-1}
+if ! [[ "$NVCC_THREADS" =~ ^[1-9][0-9]*$ ]]; then
+  echo "Invalid FLASHINFER_NVCC_THREADS=${NVCC_THREADS}; using 1"
+  NVCC_THREADS=1
+fi
+if (( NVCC_THREADS > 8 )); then NVCC_THREADS=8; fi
```
yongwww: Why is NVCC_THREADS capped at 8 here? Is there a specific reason for this upper bound?

scripts/build_flashinfer_jit_cache_whl.sh
```diff
@@ -51,6 +76,70 @@ echo "::group::Install build system"
 pip install --upgrade build
 echo "::endgroup::"
 
+install_sccache() {
```
yongwww: would it be possible to have the scccahe install in the docker

flashinfer/jit/core.py
dierksen: This was related to an earlier experiment I'd done with modifying the nvcc thread count and also trying to integrate the GNU make jobserver. I wound up pulling that out because performance was actually _worse_ when using a higher thread count (or at least the 8 that I was trying). I plan to leave it like this for now to make it easier to try again with different values.

scripts/build_flashinfer_jit_cache_whl.sh
dierksen: Yes, but I'd need to do that in a separate commit because we don't use updated docker images on pre-submit tests. I can follow up to replace that later though, then remove the install here.

scripts/build_flashinfer_jit_cache_whl.sh
dierksen: it's just a guardrail; right now it's a pretty conservative upper bound based on what we do currently.

scripts/task_test_jit_cache_package_build_import.sh
```diff
@@ -72,25 +164,166 @@ python -c "import torch; print(torch.__version__)"
 CUDA_VERSION=$(python3 -c 'import torch; print(torch.version.cuda)' | cut -d'.' -f1,2 | tr -d '.')
 echo "Detected CUDA version: cu${CUDA_VERSION}"
 
+# Parallelism: default back to one nvcc worker per compile and let ninja provide
```
yongwww: Looks like there's some duplication with scripts/build_flashinfer_jit_cache_whl.sh. It would be good to unify them

yongwww (approved):
overall lgtm

PR #3207, opened 2026-04-30
fix: guard NVFP4 KV cache on legacy decode kernel

leonardHONG:
## 📌 Description

When `BatchDecodeWithPagedKVCacheWrapper` is used with the default `use_tensor_cores=False`, NVFP4 KV cache (`torch.uint8` packed FP4) goes down the legacy decode path. That path has no FP4 support, and the `kv_cache_sf` information does not get threaded through the non-tensor-cores branch, so this can silently produce incorrect output instead of failing explicitly.

This PR adds guards in the legacy batch decode planning paths that raise `NotImplementedError` and point users to the supported alternatives:`use_tensor_cores=True` or `backend="trtllm-gen"`. This follows the direction suggested in #3097.

It also updates the `__init__` docstrings on `BatchDecodeWithPagedKVCacheWrapper` and `CUDAGraphBatchDecodeWithPagedKVCacheWrapper` so that the `use_tensor_cores` parameter explicitly documents the NVFP4 KV requirement. Previously, that note only appeared on `run()`.

## 🔍 Related Issues

Follow-up to #3097.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used another preferred method).

## 🧪 Tests

- [x] All relevant tests are passing.

Ran on RTX PRO 6000:

- `test_legacy_decode_rejects_nvfp4_paged` — confirms that the guard raises on the default wrapper. PASSED (~74s, mostly first-time `flashinfer` import).
- `test_legacy_decode_accepts_nvfp4_with_tensor_cores` — confirms that `use_tensor_cores=True` is not rejected by the new guard. PASSED (~15s).
- `test_fast_decode_plan_rejects_nvfp4` — confirms that the same guard also covers `fast_decode_plan`. PASSED.
- `test_batch_decode_with_paged_kv_cache_nvfp4[q_dtype1-128-32-4-16-512-12]` — existing #3097 NVFP4 KV decode coverage; no regression observed. PASSED (~34s).
- `tests/attention/test_tensor_cores_decode.py -k "fast_decode_plan or fast_plan"` — all 864 selected existing tests passed, confirming that the new guard does not reject legitimate non-`uint8` KV usage. PASSED (~82s).

## Reviewer Notes

The scope here is intentionally narrow: only `BatchDecodeWithPagedKVCacheWrapper.plan()` and `fast_decode_plan` are guarded.

`single_decode_with_kv_cache` does not accept NVFP4 KV in its signature. `trtllm_batch_decode_with_kv_cache` also does not go through the legacy decode kernel — its execution paths are `trtllm-gen` and `xqa`. As a result, neither requires an additional guard in this PR.

### DIFF
--- a/flashinfer/decode.py
+++ b/flashinfer/decode.py
@@ -72,6 +72,7 @@ from .utils import (
     canonicalize_torch_dtype,
     determine_attention_backend,
     device_support_pdl,
+    is_legacy_decode_supported,
     get_device_sm_count,
     is_float8,
     register_custom_op,
@@ -82,6 +83,12 @@ from .utils import (
     GPUArchitectureError,
     SINGLE_KERNEL_TMP_SIZE,
     prepare_jit_additional_args,
+)
+
+
+_NVFP4_LEGACY_DECODE_MSG = (
+    "NVFP4 KV cache (uint8 packed FP4) is not supported by the legacy decode kernel. "
+    "Use use_tensor_cores=True or backend='trtllm-gen'."
 )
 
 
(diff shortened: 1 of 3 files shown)


leonardHONG:
Hi,@bkryu @qsang-nv, could you add `run-ci` when convenient?

Thanks!

leonardHONG:
Pushed an empty commit to retrigger workflows.

PR #3153, opened 2026-04-23
feat: Add support for LoRa delta in MOE mxint4 x bf16, MXFP8 & BF16 to trtllm backend

djns99:
## 📌 Description

This PR exposes a LoRa delta parameter in the `trtllm_*_moe_routed` APIs for MXINT4xBF16 and BF16. It also returns the result of FC1 (after activation) for the consumption when calculating LoRa deltas for FC2.

The intended use-case is for frameworks to implement calculating per-token LoRa deltas and have the trtllm MOE backed consume them.

Key points:

1. LoRa deltas are expected in expanded format: `[num_tokens, top_k, intermediate_size * 2]`
2. LoRa deltas are applied before activation, and only support swiglu activation
3. Using LoRa requires frameworks to calculate the routing. This prevents issues where different top-k results are calculated
4. The framework should parse the expanded FC1 output according to the returned permuted indices map

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [ ] Tests have been added or updated as needed.
- [ ] All tests are passing (`unittest`, etc.).

## Summary by CodeRabbit

* **New Features**
  * Added LoRA GEMM1 delta support for routed Mixture of Experts operations
  * Introduced MXINT4 block-scale MoE kernel implementation
  * Enhanced MoE kernels with optional activation output return capability

* **Improvements**
  * Enhanced bias-aware kernel configuration filtering for GEMM operations
  * Refined GEMM1 bias handling with row index permutation support
  * Updated MoE kernel launchers with extended parameter plumbing for bias modes

[Review Change Stack](https://app.coderabbit.ai/change-stack/flashinfer-ai/flashinfer/pull/3153?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #3208, opened 2026-04-30
fix: Update top-k generation not to produce duplicate expert ids

djns99:
## 📌 Description

Top-K ids in MOE should not have duplicate expert IDs. While this doesn't cause issues in most cases, there are a number of places that use randint() to achieve this which can lead to unexpected behaviour in downstream code that assumes there are no duplicates

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Refactor**
  * Standardized per‑token top‑k expert selection across routing, tuning, benchmarks, docs, and tests by introducing a shared helper that generates consistent, no‑replacement top‑k samples. Updated initializers, examples, test fixtures, and benchmarks to use the unified generator; added module logging for diagnostic visibility.

### DIFF
--- a/flashinfer/fused_moe/core.py
+++ b/flashinfer/fused_moe/core.py
@@ -15,6 +15,7 @@ limitations under the License.
 """
 
 import functools
+import math
 from dataclasses import dataclass
 from types import SimpleNamespace
 from typing import Any, Dict, List, Optional, Tuple, Union
@@ -64,6 +65,7 @@ from ..utils import (
 from .utils import (
     get_hybrid_num_tokens_buckets,
     map_to_hybrid_bucket,
+    make_random_topk_ids,
 )
 from ..tllm_enums import (
     ActivationType,
@@ -1086,8 +1088,11 @@ def get_trtllm_moe_sm100_module():
             num_local_experts = self.num_local_experts
 
             def _init_packed_topk_ids(shapes, dtype, device):
-                expert_ids = torch.randint(
-                    0, num_local_experts, shapes, dtype=torch.int32, device=device
+                expert_ids = make_random_topk_ids(
+                    num_experts=num_local_experts,
+                    num_tokens=math.prod(shapes[:-1]),
+                    top_k=shapes[-1],
+                    device=device,
                 )
                 expert_weights = torch.ones(
                     shapes, dtype=torch.bfloat16, device=device
--- a/flashinfer/fused_moe/core.py
+++ b/flashinfer/fused_moe/core.py
@@ -1093,7 +1093,7 @@ def get_trtllm_moe_sm100_module():
                     num_tokens=math.prod(shapes[:-1]),
                     top_k=shapes[-1],
                     device=device,
-                )
+                ).view(shapes)
                 expert_weights = torch.ones(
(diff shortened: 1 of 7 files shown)


flashinfer/fused_moe/utils.py
```diff
@@ -423,3 +423,21 @@ def get_piecewise_cuda_graph_flag() -> bool:
...
+) -> torch.Tensor:
+    """
+    Pick ``top_k`` distinct experts (no replacement) for each of ``num_tokens`` tokens.
+
+    Returns a ``[num_tokens, top_k]`` int32 tensor whose rows contain unique
+    values in ``[0, num_experts)``.
+    """
+    if num_tokens == 0 or num_experts == 0 or top_k == 0:
+        return torch.empty(num_tokens, top_k, dtype=torch.int32, device=device)
+
```
qiching, line 448: maybe we should consider adding an explicit check for `top_k > num_experts`. now `torch.multinomial` will raise an opaque error. A clear ValueError would help debugging.

flashinfer/fused_moe/core.py
```diff
@@ -1086,9 +1088,12 @@ def _make_tuning_config(
             num_local_experts = self.num_local_experts
 
             def _init_packed_topk_ids(shapes, dtype, device):
-                expert_ids = torch.randint(
-                    0, num_local_experts, shapes, dtype=torch.int32, device=device
-                )
+                expert_ids = make_random_topk_ids(
```
qiching, line 1100: i suggest:
```
expert_ids = make_random_topk_ids(
    num_experts=max(num_local_experts, shapes[-1]),  # pool must be >= top_k
```

flashinfer/fused_moe/cute_dsl/tuner.py
```diff
@@ -285,13 +287,12 @@ def __init__(
                             1, 128, shapes, dtype=torch.uint8, device=device
                         ),
                         # 2: token_selected_experts — expert indices [0, num_experts)
-                        lambda shapes, dtype, device: torch.randint(
-                            0,
-                            max(num_experts, 1),
-                            shapes,
-                            dtype=torch.int32,
+                        lambda shapes, dtype, device: make_random_topk_ids(
+                            num_experts=max(num_experts, 1),
```
qiching, line 300: same EP concern as `core.py`: `max(num_experts, 1)` does not guard against `num_experts < top_k`. Should be `max(num_experts, shapes[-1], 1)` to ensure the sampling pool is always >= the number of samples requested.

flashinfer/fused_moe/core.py
djns99: Is there a legitimate use case where we pass an incorrect num experts? Or should this always be an error case?

qiching: thanks for the update! i see you are now using `self.num_experts`, global count instead of `num_local_experts` is cleaner, avoids the EP edge case. and the defensive clamp in`make_random_topk_ids` is a good to be safe. very good.

qiching (approved):
LGTM

djns99:
Closed by https://github.com/flashinfer-ai/flashinfer/pull/3153

PR #3210, opened 2026-04-30
fix(dcp_alltoall): require MNNVL workspace, drop broken plain-memory path

davidjpyu:
## Summary

Follow-up to #2951. The merged DCP A2A code shipped with two latent foot-guns that this PR cleans up:

1. The `mapping=None` branch in `decode_cp_a2a_allocate_workspace` returns a per-rank `torch.zeros` tensor — this deadlocks at runtime on any real multi-GPU setup.
2. The MNNVL workspace tensor's lifetime is pinned via a private attribute, which is silently lost across ordinary tensor operations and would let the fabric memory be unmapped while the kernel still holds raw pointers into it.

Both are addressed below.

## Bug 1: workspace VA mismatch (silent deadlock)

`getFifoBasePtr` in `csrc/nv_internal/tensorrt_llm/kernels/helixAllToAll.cu:177` addresses peer FIFOs via:

```cpp
auto* mappedMemory = params.workspace + mappedMemoryRank * params.workspaceStrideInU64;
```

This pointer arithmetic only resolves correctly when the workspace is a single unified VA spanning all CP ranks — i.e. MNNVL fabric memory. With per-rank `torch.zeros`, rank 0 writing to "rank 1's FIFO" lands in rank 0's own memory; peer 1 never sees it and the FIFO consumer signaling hangs forever.

Reproduced on H200 NVL with 4 GH200s — first `decode_cp_a2a_alltoall` call hangs, no CUDA error, no assertion. TRT-LLM upstream (`tensorrt_llm/_torch/distributed/ops.py:386`) only supports MNNVL — there is no plain-memory branch. The fallback added during the FlashInfer port was wrong from the start.

### Why CI didn't catch Bug 1

1. `tests/comm/test_dcp_alltoall.py` simulates `cp_size` ranks on **one** GPU with **one shared** workspace tensor — pointer arithmetic on the same allocation works, so the bug is invisible.
2. `tests/comm/test_mnnvl_dcp_alltoall.py::TestMnnvlDcpAlltoall` exercises real multi-GPU but only on the MNNVL path.
3. `TestMnnvlDcpDeviceMemoryFallback` asserts shape only — never actually calls `alltoall`, so the deadlock never fires.

## Bug 2: workspace keep-alive via tensor private attribute

The previous code did:

```python
workspace = mnnvl_mem.as_torch_strided_tensor(torch.int64)
workspace._mnnvl_mem = mnnvl_mem  # prevent GC of MNNVL handle
return workspace
```

`MnnvlMemory.__del__` calls `close_mnnvl_memory` which unmaps the underlying fabric VA. The `workspace._mnnvl_mem` private attribute is the only thing keeping the wrapper alive. **But torch tensor private attributes are NOT preserved across ordinary operations** — `workspace[r]`, `.view()`, `.contiguous()`, `.clone()`, `.to(...)`, slicing, indexing all return a fresh tensor without the attribute. Any caller that derives a view or slice and drops the original would silently free the workspace while the kernel still holds raw pointers into it. Current direct callers happen to be safe, but this is a buried mine.

### Bug 1 — drop the broken plain-memory path
- **Rename** `decode_cp_a2a_allocate_workspace` → `decode_cp_a2a_allocate_mnnvl_workspace`, matching `trtllm_mnnvl_ar` naming style. The MNNVL requirement is now obvious at the call site.
- **Make `mapping` the only required argument.** Drop the `mapping=None` branch entirely. The redundant `cp_size` and `cp_rank` parameters were also removed — `mapping` already carries that info, and a separate path was a double-source-of-truth footgun.
- **Refactor the single-GPU sim test** to use a local `_alloc_sim_workspace(cp_size)` helper that does `torch.zeros(...)` directly — that's what the test actually needs and what its docstring claims (it does *not* need the public allocator).
- **Drop `TestMnnvlDcpDeviceMemoryFallback`** — the path it covered no longer exists.
- **Use the public allocator in the multi-rank test** (`test_mnnvl_dcp_alltoall.py::_allocate_mnnvl_workspace_once`) instead of manually instantiating `MnnvlMemory`, so the public API is exercised end-to-end.
- Refresh module docstring to make the MNNVL requirement explicit.

### Bug 2 — robust workspace keep-alive
- Replace `workspace._mnnvl_mem = mnnvl_mem` with a module-level `_workspace_keepalive: Dict[int, MnnvlMemory]` keyed by `workspace.data_ptr()`. The dict pins each `MnnvlMemory` for the process lifetime, so the workspace stays mapped regardless of what the caller does with the returned tensor. This matches how TRT-LLM holds workspace ownership in its `HelixAllToAllNative._cache` class-level dict.

### Final allocator signature
```python
decode_cp_a2a_allocate_mnnvl_workspace(mapping: Mapping, *, mnnvl_config: Optional[MnnvlConfig] = None) -> torch.Tensor
```

## Test plan

All verified on dlcluster GB200-NVL72 (compute capability 10.0a, CUDA 13):

- [x] `tests/comm/test_dcp_alltoall.py` (single-GPU sim, container `flashinfer/flashinfer-ci-cu130`) — **29/29 PASSED in 5.16s**
- [x] `mpirun -launcher fork -np 4 pytest tests/comm/test_mnnvl_dcp_alltoall.py` — single-node 4-rank MNNVL — **8/8 PASSED on all 4 ranks in 1.34s**
- [x] `srun -N 2 --ntasks-per-node=4 --mpi=pmix pytest tests/comm/test_mnnvl_dcp_alltoall.py` — **multi-node 8-rank MNNVL** (2 nodes × 4 GPUs, cp_size=8, container `nvcr.io/nvidia/pytorch:26.02-py3` with HPC-X) — **8/8 PASSED on all 8 ranks in ~17s**

The multi-node run exercises real cross-node fabric memory allocation via `cuMemCreate` with FABRIC handles — same code path Helix production uses on NVL72.

🤖 AI-assisted (Claude Code)


davidjpyu:
Thanks @gemini-code-assist and @coderabbitai for the review.

Addressed all 5 Gemini suggestions in commit b87aa37f:

- **HIGH** `tests/comm/test_mnnvl_dcp_alltoall.py:38` — `_allocate_mnnvl_workspace_once` now calls the public `decode_cp_a2a_allocate_mnnvl_workspace` instead of manually instantiating `MnnvlMemory`. The multi-rank test now exercises the public API end-to-end.
- **MEDIUM** `flashinfer/comm/dcp_alltoall.py:161` (and 18, 181) — Dropped redundant `cp_size` and `cp_rank` parameters; `mapping` is now the only required arg. Final signature: `decode_cp_a2a_allocate_mnnvl_workspace(mapping, *, mnnvl_config=None)`.
- **MEDIUM** `tests/comm/test_dcp_alltoall.py:185` — Removed the now-tautological `test_allocate_returns_correct_shape_and_dtype` (it asserted properties of a local `torch.zeros` helper). Real shape/dtype assertion lives in `test_mnnvl_dcp_alltoall.py::test_workspace_shape`.

CodeRabbit's suggestion to validate `cp_size == mapping.cp_size` and `cp_rank == mapping.cp_rank` is moot now that those parameters no longer exist — the inconsistent-input footgun is removed at the API level rather than caught at runtime.

Re-tested on dlcluster GB200-NVL72:
- single-GPU sim: 29/29 passed (was 30 before removing the redundant test)
- single-node 4-rank MNNVL via mpirun: 8/8 passed on all ranks
- multi-node 8-rank MNNVL via srun --mpi=pmix (cp_size=8, 2 nodes × 4 GPUs): 8/8 passed on all ranks (already verified at the previous commit and unchanged by this API-only refactor)

aleozlx (approved):

PR #3198, opened 2026-04-28
fix(cute_dsl/moe): correct off-by-one in get_max_num_tiles to match TRT-LLM

leejnau:
## 📌 Description

Replace `get_max_num_tiles` in `flashinfer/fused_moe/cute_dsl/moe_utils.py` with TRT-LLM's compact closed-form. The previous formula over-allocated by 1 tile whenever `(E - L) % T != 0`, where `E = num_tokens * top_k`, `L = num_local_experts`, `T = tile_size`.

## Fix

```python
return (num_expanded_tokens + (tile_size - 1) * num_local_experts) // tile_size
```

Provably equivalent to `L + floor((E - L) / T)`, the tight worst-case bound on `sum_e ceil(K_e / T)`. Matches TRT-LLM's `tensorrt_llm/_torch/custom_ops/cute_dsl_custom_ops.py`.

Concrete examples at `top_k=8`, `num_tokens=16384`:
  - `num_local_experts=32`, `tile_size=128`: 1056 → **1055**
  - `num_local_experts=32`, `tile_size=256`: 544 → **543**
 
## 🔍 Related Issues

https://github.com/flashinfer-ai/flashinfer/pull/3171

https://github.com/flashinfer-ai/flashinfer/pull/2398

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [x] Tests have been added or updated as needed.
- [x] All tests are passing (`unittest`, etc.).

## Summary by CodeRabbit

* **Refactor**
  * Improved MOE tile computation to use a single tight closed-form bound, simplifying logic and ensuring consistent worst-case tiling.

* **Tests**
  * Added comprehensive tests validating the new tile-bound across edge cases (including zero tokens and threshold cases), monotonic behavior, and exact tightness between tiled and permuted token counts.

PR #3171, opened 2026-04-25
fix(cute_dsl/moe): correct tile_size=256 gemm2 tactic enumeration

leejnau:
## 📌 Description

`get_gemm2_valid_tactics(tile_size)` in `flashinfer/fused_moe/cute_dsl/tuner.py` ignored its `tile_size` argument and unconditionally returned 1-CTA tactics (`mma_m=128, cluster=(1, X)`). At `tile_size=256` the upstream gemm1 kernel emits a 2-CTA `mma_m=256` layout that the 1-CTA gemm2 cannot consume — every `(gemm1, gemm2)` tactic pair at `tile_size=256` was structurally broken. This was the underlying cause of #3067 and the workaround that disabled `tile_size=256` in the  autotuner.

This PR parameterizes `get_gemm2_valid_tactics` on `tile_size` (mirroring the existing `get_gemm1_valid_tactics` pattern) and lifts the `[128]` gate to `[128, 256]`. At `tile_size=128` the returned tactics are unchanged (backward-compatible); at `tile_size=256` it returns the corresponding 4 2-CTA entries.

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [x] Tests have been added or updated as needed.
- [x] All tests are passing (`unittest`, etc.).

## Summary by CodeRabbit

* **Refactor**
  * MOE tuner now derives GEMM tactic shapes from tile configuration, expanding candidate tactics (8 → 16) and enabling multi-CTA behavior for larger tile sizes.
* **Tests**
  * New unit tests ensure generated tactic lists follow requested tile-size parameters and that paired upstream/downstream tactics remain layout-compatible.

PR #3115, opened 2026-04-18
perf(autotuner): replace power-of-2 token buckets with hybrid spacing & fix missing routing_replay_out arg

StudyingShao:
## 📌 Description

This PR includes two improvements:

1. **perf(autotuner): Replace power-of-2 token buckets with hybrid spacing** — Pure power-of-2 spacing creates huge gaps at large values (e.g. a jump from 1024 to 2048), forcing the autotuner to pick a kernel optimised for a very different workload size. The new hybrid scheme uses four phases with progressively coarser spacing:
   - Phase 1: `[min .. 256]` — power-of-2 (step ×2)
   - Phase 2: `(256 .. 2048]` — linear step 256
   - Phase 3: `(2048 .. 4096]` — linear step 512
   - Phase 4: `(4096 .. max]` — power-of-2 (step ×2)

   All callsites in MoE, GEMM, and low-latency GEMM autotuners are updated to use the new `get_hybrid_num_tokens_buckets` / `map_to_hybrid_bucket` API.

2. **fix: Pass missing `routing_replay_out` arg to `trtllm_fp8_per_tensor_scale_moe`** — Two call sites in `fused_moe/core.py` were missing the `routing_replay_out` argument, causing it to be silently dropped.

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [x] Tests have been added or updated as needed.
- [x] All tests are passing (`unittest`, etc.).

## Reviewer Notes

**Changed files:**
- `flashinfer/fused_moe/utils.py` — Core implementation: new `get_hybrid_num_tokens_buckets`, `map_to_hybrid_bucket`, `map_to_hybrid_bucket_uncapped`; removed old `get_last_power_of_2_num_tokens_buckets`
- `flashinfer/fused_moe/core.py` — Updated all MoE autotuner callsites + added missing `routing_replay_out` arg
- `flashinfer/fused_moe/cute_dsl/tuner.py` — Updated CuTe DSL FP4 MoE tuner callsite
[...]

PR #3216, opened 2026-05-01
fix(cute_dsl/moe): make autotuner bucket configuration adapt to runtime input

leejnau:
## 📌 Description

The autotuner's `DynamicTensorSpec` in `flashinfer/fused_moe/cute_dsl/tuner.py` declared `gen_tuning_buckets` as the pre-computed tuple `get_hybrid_num_tokens_buckets(8192)` and `map_to_tuning_buckets` as `lambda x: map_to_hybrid_bucket(x,8192)`. The hardcoded 8192 cap silently clamped any runtime workload larger than that to the 8192-bucket's cached tactic — at DeepSeek-V3 prefill (N=16384) fi profiled at half the per-expert workload and used a tactic optimized for the wrong shape. 

This PR replaces the pre-computed tuple with the bare callable form (`get_hybrid_num_tokens_buckets`) and switches the mapper to the uncapped variant `map_to_hybrid_bucket_uncapped` (added alongside the hybrid-bucket scheme for exactly this case). The autotuner now invokes them with the actual input dim at autotune time, matching TRT-LLM's pattern at `cute_dsl_custom_ops.py:2390-2391` and flashinfer's own pattern at `gemm/gemm_base.py:_FP8_GEMM_SM100_TUNING_CONFIG`.

## 🔍 Related Issues

https://github.com/flashinfer-ai/flashinfer/pull/3171
https://github.com/flashinfer-ai/flashinfer/pull/3198
https://github.com/flashinfer-ai/flashinfer/pull/3115

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Bug Fixes**
  * MoE autotuner now uses uncapped dynamic hybrid bucket mapping instead of a fixed-bounded set, improving adaptation to varying input token sizes.

* **Tests**
  * Added offline tests validating autotuner bucket configuration: dynamic bucket generation, responsiveness to input size, monotonic mapping behavior, large-input scaling, and alignment with expected power-of-2 bucket values.

### DIFF
--- a/flashinfer/fused_moe/cute_dsl/tuner.py
+++ b/flashinfer/fused_moe/cute_dsl/tuner.py
@@ -43,7 +43,7 @@ from ...autotuner import (
 )
 from ..utils import (
     get_hybrid_num_tokens_buckets,
-    map_to_hybrid_bucket,
+    map_to_hybrid_bucket_uncapped,
 )
 
 logger = logging.getLogger(__name__)
@@ -278,8 +278,15 @@ class CuteDslFusedMoENvfp4Runner(TunableRunner):
                 DynamicTensorSpec(
                     input_idx=(0, 1, 2, 3, 11),
                     dim_idx=(0, 0, 0, 0, 0),
-                    gen_tuning_buckets=get_hybrid_num_tokens_buckets(8192),
-                    map_to_tuning_buckets=lambda x: map_to_hybrid_bucket(x, 8192),
+                    # Pass bucket generators as bare callables (matching
+                    # TRT-LLM's pattern at cute_dsl_custom_ops.py:2390-2391
+                    # and 2700-2703, and flashinfer's own pattern at
+                    # `gemm/gemm_base.py:_FP8_GEMM_SM100_TUNING_CONFIG`).
+                    # The autotuner invokes them with the actual input dim
+                    # at autotune time, so the bucket set adapts to the
+                    # workload — no hardcoded cap needed.
+                    gen_tuning_buckets=get_hybrid_num_tokens_buckets,
+                    map_to_tuning_buckets=map_to_hybrid_bucket_uncapped,
                     tensor_initializers=[
                         # 0: x — FP4 quantized input (uint8 packed)
                         lambda shapes, dtype, device: torch.randint(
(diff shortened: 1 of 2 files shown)


flashinfer/fused_moe/cute_dsl/tuner.py
```diff
@@ -278,8 +278,15 @@ def __init__(
                 DynamicTensorSpec(
                     input_idx=(0, 1, 2, 3, 11),
                     dim_idx=(0, 0, 0, 0, 0),
-                    gen_tuning_buckets=get_hybrid_num_tokens_buckets(8192),
-                    map_to_tuning_buckets=lambda x: map_to_hybrid_bucket(x, 8192),
+                    # Pass bucket generators as bare callables (matching
```
qiching: the TRT-LLM line numbers (2390-2391, 2700-2703) will go stale. the code is self-explanatory given the function names. I'd trim to 1-2 lines max, e.g.:
```
# bare callables, autotuner adapts bucket set to actual input dim
# (matches gemm_base.py _FP8_GEMM_SM100_TUNING_CONFIG pattern).
```

tests/moe/test_cute_dsl_fused_moe.py
```diff
@@ -673,6 +673,190 @@ def test_consistent_with_get_max_num_tiles(
...
+            num_local_experts=256,
+        )
+
+    def test_gen_tuning_buckets_is_callable_not_static_tuple(self):
+        """``gen_tuning_buckets`` must be a callable that adapts to the
+        actual input dim at autotune time — not a pre-computed
+        tuple/sequence that bakes in a hardcoded cap.
+        """
+        runner = self._make_runner()
+        spec = runner.tuning_config.dynamic_tensor_specs[0]
```
qiching: If `gen_tuning_buckets` is a tuple, callable(tuple_instance) is already false, so the first assertion fails before the second is ever reached. the second assertion is dead. could you check? maybe either remove it or swap the order.

qiching (commented):
every test method calls `self._make_runner()` independently. Since the runner is stateless for these checks, i will recommend `@pytest.fixture` would reduce boilerplate and test runtime.

qiching (approved):
good! module-scoped fixture that replaces `_make_runner` and the repeated construction of the runner for each test with a `bucket_spec` pytest fixture. This is better! the module scope means the runner is constructed only once for the entire test module, making it more efficient!

aleozlx (commented):
reviewed. will wait for bot run

aleozlx (approved):

PR #3252, opened 2026-05-06
fix(cute_dsl/moe): unbias autotuner profiling for tile_size enumeration

leejnau:
## 📌 Description

`CuteDslMoEWrapper.__init__` pre-allocates `_gemm1_output`, `_gemm1_output_scale`, and `_moe_sort_buffers` sized for `self.tile_size` only. The `use_prealloc` gate in `_forward_with_tactic` (`fused_moe.py`) is `tile_size == self.tile_size and self.use_cuda_graph and num_tokens <= self.max_num_tokens`, so during autotune profiling the mismatched-`tile_size` tactics fall through to dynamic `torch.empty()` allocation while matching ones run on the prealloc. The autotuner sees **asymmetric allocation overhead** between tactic groups and consistently picks the matching `tile_size` even when intrinsic kernel performance favors the other — at EP=8/16 N=16384, fi locks to `tile_size=128` in 14/14 cache entries while TRT-LLM picks `tile_size=256` more often.

The fix includes three coordinated changes: (1) `tuner.py` lifts the hardcoded `[128, 256]` to a module-level `VALID_TILE_SIZES` tuple — single source of truth for tactic enumeration AND prealloc sizing; (2) `fused_moe.py:_allocate_buffers` sizes buffers to fit any `tile_size in VALID_TILE_SIZES` (`max_num_permuted_tokens` increases with `tile_size` → use `max(VALID_TILE_SIZES)`; `max_num_tiles` decreases → use `min(VALID_TILE_SIZES)`); (3) the prealloc gate becomes `tile_size in VALID_TILE_SIZES`. Both tactic groups now reuse the prealloc; profiling is unbiased.

## 🔍 Related Issues

https://github.com/flashinfer-ai/flashinfer/pull/3216
https://github.com/flashinfer-ai/flashinfer/pull/3171

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CUDA preallocation gating so preallocated buffers are reused safely when CUDA graphs are enabled, avoiding allocation mismatches across supported tile sizes and improving memory efficiency.

* **New Features**
  * Autotuner now scopes per-tactic timing so preallocation is skipped during measurement windows but used outside them; buffer sizing now supports all valid tile sizes.

* **Tests**
  * Added CPU/GPU tests validating preallocation capacity and correct gating behavior during tuning and inference.

### DIFF
--- a/flashinfer/autotuner.py
+++ b/flashinfer/autotuner.py
@@ -640,6 +640,52 @@ def autotune(
             tuner.save_configs(cache)
 
 
+# Thread-local "currently inside the per-tactic measurement window" flag.
+# Set/cleared by ``_profile_measurement_scope`` and queried by
+# ``is_in_profile_measurement``.  Distinct from
+# ``AutoTuner.is_tuning_mode``, which is True for the entire ``autotune(True)``
+# context (cache hits, prep calls, post-``choose_one`` runs, and concurrent
+# threads inclusive); ``is_in_profile_measurement`` is True only on the
+# specific thread that is actively timing a tactic, and only during the
+# warmup + measurement window inside ``_profile_single_kernel``.
+_profile_measurement_thread_local = threading.local()
+
+
+@contextlib.contextmanager
+def _profile_measurement_scope():
+    """Mark the calling thread as inside the autotuner's per-tactic
+    measurement window.  Nested entries are honored via prev/restore so
+    correctness doesn't depend on a single entry path.
+    """
+    prev = getattr(_profile_measurement_thread_local, "active", False)
+    _profile_measurement_thread_local.active = True
+    try:
+        yield
+    finally:
+        _profile_measurement_thread_local.active = prev
+
+
+def is_in_profile_measurement() -> bool:
+    """Return True iff the calling thread is currently inside the
+    autotuner's per-tactic measurement window (warmup + timed run inside
+    ``AutoTuner._profile_single_kernel``).
+
+    This is *narrower* than ``AutoTuner.get().is_tuning_mode``:
+
+    - ``is_tuning_mode`` is True for the entire ``autotune(True)`` context,
+      including cache lookups, ``do_preparation`` calls, the final invocation
+      after ``choose_one`` returns, and any concurrent threads' work.
+    - ``is_in_profile_measurement`` is True only on the calling thread, and
+      only during the actual measurement window of a single tactic.
+
+    Wrappers that need to bypass per-call optimizations (e.g. CUDA-graph
+    preallocated buffers sized for a specific construction-time tile) so
+    that the autotuner's tactic comparison is unbiased should consult
+    ``is_in_profile_measurement()`` rather than ``is_tuning_mode``.
+    """
+    return getattr(_profile_measurement_thread_local, "active", False)
+
+
 @dataclass
 class AutoTunerStatistics:
     """Statistics collected by the AutoTuner.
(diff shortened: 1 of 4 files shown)


nv-yunzheq (approved):

flashinfer/fused_moe/cute_dsl/fused_moe.py
```diff
@@ -488,13 +488,24 @@ def _forward_with_tactic(
...
+        #   it, biasing the autotuner toward picking the matching
+        #   ``tile_size`` regardless of intrinsic kernel performance.
+        # - the tactic's tile_size doesn't match ``self.tile_size``:
+        #   the prealloc layout would be wrong.
+        # - the batch exceeds what the buffers were sized for (e.g.
+        #   autotuner probing larger buckets).
         num_tokens = x.shape[0]
         use_prealloc = (
             self.use_cuda_graph
+            and not AutoTuner.get().is_tuning_mode
```
qiching: two qq:
`AutoTuner.get()` is called on every forward invocation in the inference path. should double checked locking singleton so the overhead is negligible, one is None check + attribute return, but worth confirming this meets latency requirements?

another is `is_tuning_mode` is read without holding `_lock`. Under CPython this is fine,d boolean reads are atomic due to the GIL, but it's worth noting as a potential concern if the runtime ever changes?

Just reference. could you consider and answer?

qiching (commented):
using `threading.local()` is correct! 
but i have another concerns that if Autotuner profiling might execute the runner in a subthread. If the `runner()` call during warmup or just measure is executed in another thread, the subthread will not see `active=True`, and the gate will fail.

could we consider this make sense?

flashinfer/autotuner.py
```diff
@@ -640,6 +640,52 @@ def autotune(
...
 
+# Thread-local "currently inside the per-tactic measurement window" flag.
+# Set/cleared by ``_profile_measurement_scope`` and queried by
+# ``is_in_profile_measurement``.  Distinct from
+# ``AutoTuner.is_tuning_mode``, which is True for the entire ``autotune(True)``
+# context (cache hits, prep calls, post-``choose_one`` runs, and concurrent
+# threads inclusive); ``is_in_profile_measurement`` is True only on the
+# specific thread that is actively timing a tactic, and only during the
+# warmup + measurement window inside ``_profile_single_kernel``.
+_profile_measurement_thread_local = threading.local()
```
qiching, line 651: using `threading.local()` is correct! 
but i have another concerns that if Autotuner profiling might execute the runner in a subthread. If the `runner()` call during warmup or just measure is executed in another thread, the subthread will not see `active=True`, and the gate will fail.

could we consider this make sense?

flashinfer/fused_moe/cute_dsl/fused_moe.py
leejnau: addressed in https://github.com/flashinfer-ai/flashinfer/pull/3252/changes/f460328166117057cd324321e5954b8b0a254ce3

flashinfer/autotuner.py
leejnau: In the current autotuner implementation, _profile_single_kernel() invokes runner(...) synchronously on the same Python thread for both warmup and timed measurement, so the threading.local() flag is visible to the wrapper when _forward_with_tactic() runs.
 
The CUDA kernels themselves launch asynchronously, but the Python-side allocation/prealloc gate is evaluated before launch, on that same caller thread. So the thread-local scope covers the decision we need here.

If we later refactor autotuner profiling to dispatch runner(...) from a Python worker thread, then we should either propagate the profiling context into that worker or replace this with an explicit per-call kwarg/context token. For today’s code path, threading.local() is probably the correct narrower signal: it avoids leaking profiling behavior into concurrent inference threads while still covering the actual measured runner calls.

qiching: now approach is correct for today's synchronous profiling path. Left a note for future: if `_profile_single_kernel` ever dispatches `runner()` calls from a worker thread, the `threading.local()` flag won't propagate.

qiching (approved):

nv-yunzheq (approved):

Issue #3085, opened 2026-04-16
[Bug]: cuBLAS bmm_fp8 segfaults on B200 during autotuning


ZhanqiuHu:
## Environment

- **flashinfer-python**: 0.6.7
- **apache-tvm-ffi**: 0.1.9
- **PyTorch**: 2.11.0+cu130
- **CUDA runtime**: 13.0.88
- **nvidia-cublas**: 13.2.1.1
- **GPU**: NVIDIA B200 (SM100)
- **Driver**: 580.126.09
- **Python**: 3.12.3
- **OS**: Ubuntu 24.04.4 LTS

## Description

`flashinfer.bmm_fp8` segfaults inside the `autotune()` context on B200. The crash is at `gemm_base.py:131` — the cuBLAS runner's `forward` calling `bmm_fp8_internal_cublaslt`. Without `autotune()`, the same call works fine (CUTLASS SM10x is used as default).

Related vLLM issue: https://github.com/vllm-project/vllm/issues/39814

## Reproducer

```python
import torch
import flashinfer
from flashinfer.autotuner import autotune

torch.manual_seed(42)
_ = torch.randn(16, 16, device="cuda") @ torch.randn(16, 16, device="cuda")
torch.cuda.synchronize()

def to_float8(x, dtype=torch.float8_e4m3fn):
    finfo = torch.finfo(dtype)
    amax = x.abs().amax().clamp(min=1e-12)
    scale = finfo.max / amax
    x_scl_sat = (x * scale).clamp(min=finfo.min, max=finfo.max)
    return x_scl_sat.to(dtype), scale.float().reciprocal()

M, K, N = 8192, 2688, 5376

A_bf16 = torch.randn(M, K, device="cuda", dtype=torch.bfloat16)
A_fp8, A_scale = to_float8(A_bf16)

B_raw = torch.randn(N, K, device="cuda", dtype=torch.bfloat16)
B_fp8, B_scale = to_float8(B_raw)
B_fp8 = B_fp8.t()  # [K, N] column-major

# Works
result = flashinfer.bmm_fp8(
    A_fp8.unsqueeze(0), B_fp8.unsqueeze(0),
    A_scale, B_scale, torch.bfloat16,
    out=None, backend="auto",
)
print(f"Without autotune: OK {result.shape}")

# Segfaults
with torch.inference_mode(), autotune():
    result = flashinfer.bmm_fp8(
[...]

PR #3256, opened 2026-05-07
fix(gemm): exclude cublas from bmm_fp8 heuristic on sm_120 (RTX 5090)

### LINKED ISSUE #3255: bmm_fp8: CUBLAS_STATUS_NOT_SUPPORTED on sm_120 (RTX 5090) — autotuner selects cublas on small shapes, fails on real model tensor shapes
## Environment

- GPU: NVIDIA GeForce RTX 5090 (sm_120, 32 GB × 2)
- Driver: 580.126.09 / CUDA 13.0
- flashinfer: 0.6.8.post1 (cu130/torch2.11 wheel)
- vLLM: 0.20.1
- Model: speakleash/Bielik-11B-v2.3-Instruct-FP8 (2× RTX 5090, tensor-parallel)

## Symptom

vLLM crashes during FP8 model warmup:

```
RuntimeError: bmm_fp8_internal_cublaslt failed: CUBLAS_STATUS_NOT_SUPPORTED
  File ".../flashinfer/data/csrc/bmm_fp8.cu", line 57
```

The error occurs consistently during the autotuner benchmark phase — not on the first tiny synthetic shape, but on a later shape that matches real model tensor dimensions.

## Root cause

`_heuristic_func_bmm_fp8` in `flashinfer/gemm/gemm_base.py` unconditionally appends `cublas` whenever `"cublas"` is in `suitable_backends`, even on sm_120 where `cutlass_sm12x` was just selected:

```python
# gemm_base.py (both 0.6.8.post1 and current main)
is_sm120_supported = _match_sm_version(A.device, ["120", "121"])  # ← computed

if "cutlass" in suitable_backends and not is_e5m2:
    if is_sm_supported:
        heuristic_backends.append("cutlass_sm10x")
    elif is_sm120_supported:
        heuristic_backends.append("cutlass_sm12x")   # correct for sm_120
if "cublas" in suitable_backends:                    # ← no is_sm120_supported guard!
    heuristic_backends.append("cublas")
```

The autotuner benchmarks `cublas` on small synthetic shapes (e.g. the initial 1-element tensor the heuristic call uses to probe). cuBLASLt's heuristic query succeeds on those shapes (returns N>0 results), so `cublas` wins. On real model tensor shapes during the actual inference warmup, `bmm_fp8_internal_cublaslt` returns `CUBLAS_STATUS_NOT_SUPPORTED` and crashes.

`is_sm120_supported` is already computed in the function — the guard was simply never added.

## Fix

One-line guard:

```diff
-    if "cublas" in suitable_backends:
+    if "cublas" in suitable_backends and not is_sm120_supported:
         heuristic_backends.append("cublas")
```

`cutlass_sm12x` is the correct and sufficient backend for sm_120 — it handles all K values via padding (as the existing comment notes). Excluding `cublas` from the heuristic on sm_120 lets the autotuner use `cutlass_sm12x` consistently.

## Workaround (for 0.6.8.post1 users)

Set `VLLM_PREFER_FLASHINFER_OPS=0` to bypass flashinfer FP8 GEMM and fall back to `torch._scaled_mm`. This is effective but sacrifices flashinfer performance.

## Verification

After applying the patch on RTX 5090 hardware (same driver/CUDA as above):
- autotuner completes cleanly and selects `cutlass_sm12x`
- full 924-record FP8 inference run completes without errors
- `VLLM_PREFER_FLASHINFER_OPS=0` workaround no longer needed

A PR with the fix is attached.

Related: #3085 (sm_120 JIT compilation — separate but co-occurring issue on RTX 5090 with CUDA < 12.9)

YASoftwareDev:
Fixes #3255.

## Problem

`_heuristic_func_bmm_fp8` in `flashinfer/gemm/gemm_base.py` adds `cublas` to the heuristic
candidate list unconditionally — even on sm_120 (RTX 5090 / GeForce Blackwell) where
`cutlass_sm12x` was already selected. The variable `is_sm120_supported` is computed two
lines above the guard but never used to exclude `cublas`.

The autotuner benchmarks on small synthetic shapes where cuBLASLt's heuristic lookup
returns N>0 results, so `cublas` wins the profile. At inference time, on real model tensor
shapes, `bmm_fp8_internal_cublaslt` returns `CUBLAS_STATUS_NOT_SUPPORTED` and crashes:

```
RuntimeError: bmm_fp8_internal_cublaslt failed: CUBLAS_STATUS_NOT_SUPPORTED
  File ".../flashinfer/data/csrc/bmm_fp8.cu", line 57
```

## Fix

```diff
-    if "cublas" in suitable_backends:
+    if "cublas" in suitable_backends and not is_sm120_supported:
         heuristic_backends.append("cublas")
```

`cutlass_sm12x` handles all K values on sm_120 via padding (the existing comment already
documents this). The cublas exclusion lets the autotuner use `cutlass_sm12x` consistently
on sm_120.

## Testing

Tested on RTX 5090 hardware (driver 580.126.09 / CUDA 13.0, flashinfer 0.6.8.post1,
vLLM 0.20.1, model speakleash/Bielik-11B-v2.3-Instruct-FP8, 2× RTX 5090 tensor-parallel):

- **Before**: autotuner selects `cublas`, crashes with `CUBLAS_STATUS_NOT_SUPPORTED`
- **After**: autotuner selects `cutlass_sm12x`, 924-record inference run completes cleanly

Note: this fix targets the heuristic logic only. The pre-built sm_120 AOT kernels are
absent from the cu130/torch2.11 wheel (separate issue #3085); users on CUDA < 12.9 still
need nvcc ≥ 12.9 for JIT compilation. This fix is independent of that.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected backend selection for FP8 batched matrix operations on recent SM12x GPUs so the most compatible implementation is chosen, preventing runtime failures and improving stability during matrix workloads.

### DIFF
--- a/flashinfer/gemm/gemm_base.py
+++ b/flashinfer/gemm/gemm_base.py
@@ -6036,7 +6036,13 @@ def _heuristic_func_bmm_fp8(
         elif is_sm120_supported:
             # supports all K values through padding
             heuristic_backends.append("cutlass_sm12x")
-    if "cublas" in suitable_backends:
+    if "cublas" in suitable_backends and not is_sm120_supported:
+        # cuBLASLt bmm_fp8 returns CUBLAS_STATUS_NOT_SUPPORTED on sm_120 (RTX 5090 /
+        # GeForce Blackwell) for real model tensor shapes.  The autotuner profiles on
+        # small synthetic shapes where cuBLASLt heuristics succeed and incorrectly
+        # selects cublas; at inference time the same kernel fails.
+        # cutlass_sm12x is the correct backend for sm_120 — it handles all K values
+        # via padding and is already added above.
         heuristic_backends.append("cublas")
     if CUDNN_AVAILABLE and "cudnn" in suitable_backends:
         heuristic_backends.append("cudnn")
--- a/flashinfer/gemm/gemm_base.py
+++ b/flashinfer/gemm/gemm_base.py
@@ -5972,7 +5972,7 @@ def _cudnn_bmm_fp8_requirement(
     return True
 
 
-@supported_compute_capability([89, 90, 100, 103, 110, 120, 121])
+@supported_compute_capability([89, 90, 100, 103, 110])
 def _cublas_bmm_fp8_requirement(
     A: torch.Tensor,
     B: torch.Tensor,
@@ -5982,6 +5982,10 @@ def _cublas_bmm_fp8_requirement(
     out: Optional[torch.Tensor] = None,
     backend: Literal["cudnn", "cublas", "cutlass", "auto"] = "cublas",
 ):
+    # sm_120/121 (RTX 5090 / GeForce Blackwell) excluded: cuBLASLt bmm_fp8 returns
+    # CUBLAS_STATUS_NOT_SUPPORTED for real model tensor shapes on sm_120, even though
+    # the autotuner succeeds on the small synthetic shapes used during profiling.
+    # cutlass_sm12x is the correct backend for sm_120 (handles all K values via padding).
     return True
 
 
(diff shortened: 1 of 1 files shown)


YASoftwareDev:
Thanks for the reviews, @gemini-code-assist and @coderabbitai.

Both points from Gemini are addressed in the follow-up commit (6e19046):

1. **Comment placement** — moved the explanatory comment above the `if "cublas" in suitable_backends and not is_sm120_supported:` guard. The previous placement inside that branch was logically confusing: the block only executes when `is_sm120_supported` is `False`, so a comment about sm_120 behaviour would never be reached on an actual sm_120 device.

2. **`_cublas_bmm_fp8_requirement`** — removed `120, 121` from `@supported_compute_capability([89, 90, 100, 103, 110, 120, 121])`. This is the correct authoritative gate: it ensures that explicit `bmm_fp8(..., backend='cublas')` calls on sm_120 raise `BackendSupportedError` rather than crashing inside cuBLASLt. The heuristic guard in `_heuristic_func_bmm_fp8` is now defence-in-depth — cublas won't appear in `suitable_backends` on sm_120 at all.

bkryu:
As explained in  #3255, cublas is a properly supported backend in `bmm_fp8` on SM120/121. Hence we should not be excluding the backend from the heursitic.

I will close the PR but feel free to rekindle the discussion in #3255

PR #3260, opened 2026-05-07
non-override tactic control

yanqinz2:
## 📌 Description

It also fixes a tactic mismatch hazard in the non-override path. Autotuning buckets dynamic M values, but static cuDNN graphs are rebuilt with the actual runtime M. A tactic index profiled on a bucket-M graph may not refer to the same execution plan in the actual-M graph. To avoid applying an invalid or mismatched tactic, the non-override cuDNN paths now expose only the fallback tactic -1, forcing runtime to use the cuDNN heuristic path for the actual static graph.

This update also brings the cuDNN FP8 and MXFP8 GEMM paths in line with the existing BF16/FP4 override-shape design.

For FP8, the per-tensor quantized cuDNN graph builders were renamed to follow the same naming convention as the other GEMM paths: build_cudnn_gemm_fp8_graph and build_cudnn_gemm_fp8_graph_override_shape. The cuDNN FP8 runner now supports the override-shape execution path, using the autotuner’s effective M-bucket mapper to build a reusable bucketed graph and passing the runtime shapes through override_shapes.

For MXFP8, the cuDNN graph construction was refactored to match the FP4/BF16 structure. The previous create_cudnn_execution_plans_mxfp8_gemm + _get_cudnn_mxfp8_gemm_graph split was replaced with a single build_cudnn_gemm_mxfp8_graph builder that owns graph creation, support checking, and plan building. The MXFP8 runner now also supports the override-shape path with bucketed graph reuse.

For both FP8 and MXFP8, the non-override static cuDNN path no longer participates in autotune plan-index profiling, since static graphs are rebuilt for the actual runtime M and cannot safely reuse tactic indices profiled on bucketed M shapes.

## 🔍 Related Issues

test_bmm_bf16 failure on B300

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **API Changes**
  * Removed legacy cuDNN GEMM override helpers from the public API; core GEMM wrappers remain.
  * Public exports now include available CuTe‑DSL kernels at import time when present.

* **Improvements**
  * Safer override‑shape handling with deterministic fallback when dynamic-shape support is unavailable.
  * Improved cache invalidation to avoid stale execution plans.

* **Tests**
  * Added GPU tests validating override‑shape execution across dynamic sizes and quantized modes.


flashinfer/gemm/gemm_base.py
```diff
@@ -4778,41 +4807,15 @@ def get_valid_tactics(
 
             # currently cudnn backend does not support alpha for dynamic-shape
             # remove this restriction once cudnn suppport it
-            if is_cudnn_override_shape_available():
+            if self._use_override_shape:
```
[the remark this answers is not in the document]
yanqinz2: fixed

flashinfer/gemm/gemm_base.py
```diff
@@ -488,9 +444,9 @@ def mm_bf16(
     pdl: bool = False,
     out: Optional[torch.Tensor] = None,
     out_dtype: torch.dtype = torch.bfloat16,
-    backend: Literal[
-        "cudnn", "cutlass", "tgv", "cublaslt", "tinygemm", "auto"
-    ] = "cudnn",
+    backend: Literal["cudnn", "cutlass", "tgv", "cublaslt", "auto"] = "cudnn",
+    *,
+    enable_override_shape: bool = True,
```
dhiraj113: Is this API level change strictly necessary?

flashinfer/gemm/gemm_base.py
```diff
@@ -3114,29 +3156,63 @@ def _cudnn_gemm_fp8(
...
     class CudnnFp8GemmRunner(TunableRunner):
+        def __init__(self, m_bucket_mapper):
+            super().__init__()
+            self._m_bucket_mapper = (
+                m_bucket_mapper
+                if m_bucket_mapper is not None
+                else map_to_hybrid_bucket_uncapped
+            )
+            self._use_override_shape = is_cudnn_override_shape_available()
+
```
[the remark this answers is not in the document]
yanqinz2: We decided not to add override shape switch

flashinfer/gemm/gemm_base.py
```diff
@@ -1281,6 +1281,11 @@ def fp8_gemm_sm100(
     workspace_buffer: torch.Tensor,
     runner_names: List[str],
 ) -> None:
+    tuner = AutoTuner.get()
+    effective_m_bucket_mapper = tuner.get_effective_map_to_tuning_buckets(
+        _FP8_GEMM_SM100_TUNING_CONFIG, spec_idx=0
+    )
```
dhiraj113, line 1287: Is this is only used by `cudnn`, can this be moved to under cudnn runner.

dhiraj113 (approved):

flashinfer/gemm/gemm_base.py
yanqinz2: Yes, it would be cleaner to move its lookup into the cuDNN runner construction path. We can update this in a follow-up change.

PR #3265, opened 2026-05-07
build(jit-cache): split flashinfer-jit-cache wheels by SM family

### LINKED ISSUE #3033: CLI utility that can aot download the jit cache

### LINKED ISSUE #3257: [Bug]GitHub Release upload fails for `flashinfer-jit-cache` cu130 x86_64 wheel due to 2 GiB asset limit
### Summary

Detect create_release failure in : [Nightly Release 245](https://github.com/flashinfer-ai/flashinfer/actions/runs/25415540774) 
Example log: https://github.com/flashinfer-ai/flashinfer/actions/runs/25415540774/job/74564447018 

  The nightly release workflow fails during the `create-release` job when uploading the `flashinfer-jit-cache` wheel for CUDA 13.0 x86_64.

  The build appears to complete and earlier JIT cache wheels upload successfully, but the `cu130 x86_64` wheel is rejected by GitHub Releases because the asset size exceeds the 2 GiB limit.

  ### Failure log

  ```
  Processing jit-cache-cu128-x86_64...
  ✅ Uploaded jit-cache-cu128-x86_64
  Processing jit-cache-cu128-aarch64...
  ✅ Uploaded jit-cache-cu128-aarch64
  Processing jit-cache-cu129-x86_64...
  ✅ Uploaded jit-cache-cu129-x86_64
  Processing jit-cache-cu129-aarch64...
  ✅ Uploaded jit-cache-cu129-aarch64
  Processing jit-cache-cu130-x86_64...
  HTTP 422: Validation Failed (https://uploads.github.com/repos/flashinfer-ai/flashinfer/releases/318788053/assets?label=&name=flashinfer_jit_cache-0.6.11.dev20260507%2Bcu130-cp39-abi3-manylinux_2_28_x86_64.whl)
  size must be less than 2147483648
  Process completed with exit code 1.
```

  ### Impact

  The release creation fails after several assets have already been uploaded. The failing asset is:

  flashinfer_jit_cache-0.6.11.dev20260507+cu130-cp39-abi3-manylinux_2_28_x86_64.whl

  This also appears related to a later test job failure where pip install rejects the downloaded cu130 JIT cache wheel as invalid.

  ### Suspected cause

  The cu130 JIT cache wheel likely grew beyond GitHub Release’s per-asset size limit of 2147483648 bytes.

  Recent changes that may have contributed to the size increase include the SM120/SM121 CUTLASS GEMM expansion, especially:

  a11ce269 perf: Integrate CUTLASS Small Tile N Blockscaled GEMMs/Grouped GEMMs for SM120 and SM121 (#3152)

  That PR adds additional SM120/SM121 CUTLASS GEMM and grouped GEMM kernel variants, including small-N tile shapes such as 128x32 and 128x64, plus extra dispatch/config variants. Since the cu130 JIT cache build includes many GPU arch targets, this can
  significantly increase the packaged .so payload.

  ### Possible fixes (By Codex)

  - Split flashinfer-jit-cache wheels by GPU architecture or kernel family.
  - Reduce the default AOT contents for cu130.
  - Temporarily gate or exclude some newly added SM120/SM121 CUTLASS GEMM variants from the release JIT cache package.
  - Publish oversized JIT cache artifacts somewhere other than GitHub Releases, or split them before upload.

### PR DESCRIPTION
**Pull request by dierksen**
The cu130 flashinfer-jit-cache wheel grew to 2.0 GB and started failing to upload as a GitHub Release asset (per-asset 2 GiB limit; see #3257). Each new SM target appended ~150-200 MB compressed to every wheel, and cu130 carries 8 (sm75/80/89/90a/100a/103a/110a/120f).

Split each (CUDA, CPU-arch) wheel into three by GPU SM family:
- sm9x   - Ampere/Ada/Hopper       (<= sm90a)
- sm10x  - Datacenter Blackwell    (sm100a/103a/110a)
- sm12x  - Consumer Blackwell      (sm120f, future sm121a)

Same package name everywhere; the family is encoded in the PEP 440 local-version, so wheels resolve as e.g. 'flashinfer-jit-cache== 0.6.11+cu130.sm10x'. Existing 'pip install flashinfer-jit-cache' still works once the right pin is given.

Driven by a new 'flashinfer install-jit-cache-wheel' subcommand that detects FlashInfer version, CUDA version, and GPU compute capability (via torch.cuda.get_device_capability) and runs the matching pip install. Honors --cuda-version, --sm-family, --nightly, --dry-run. Modeled on the CLI scaffolding from #3142 with the family dimension added.

Build side: 'FLASHINFER_JIT_CACHE_SM_FAMILY' env var, when set, filters 'FLASHINFER_CUDA_ARCH_LIST' to the family's archs and appends '.<family>' to the local-version suffix. Release / nightly workflows gain an 'sm_family' matrix dimension; the upload-to-release loop iterates over all three families. The wheel-index regex accepts the new local-version shape and remains compatible with the legacy '+cuXY' format.

Closes #3257
Closes #3033

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Testing Done

Local DGX Spark validation:

- Confirmed host is `aarch64` with NVIDIA GB10, compute capability `12.1`; PyTorch sees CUDA `13.0` and reports the device as `(12, 1)`.
- Ran `flashinfer install-jit-cache-wheel --dry-run` from the source checkout. It now resolves from `version.txt` when package metadata is `0.0.0+unknown`, detects CUDA `13.0`, selects `sm12x`, and prints `flashinfer-jit-cache==0.6.11+cu130.sm12x`.
- Verified uv-style venv behavior: this environment has no `python -m pip`, so the CLI now falls back to `uv pip install --python ...`.
- Ran the live install path against `https://flashinfer.ai/whl/cu130`; it reached the resolver cleanly and failed only because `0.6.11+cu130.sm12x` is not yet published.
- Downloaded the PR-built `jit-cache-cu129-aarch64-sm12x` artifact from Release run `25528598737`, served it through a local simple index, installed it with explicit `--cuda-version cu129 --sm-family sm12x --index-url ...`, imported `flashinfer_jit_cache`, verified `FLASHINFER_AOT_DIR` points at the installed package cache, then uninstalled it because the local machine is CUDA 13.0.
- Checked the Release run artifacts: the matching `jit-cache-cu130-aarch64-sm12x` artifact is absent because that job failed while downloading build dependencies (`IncompleteRead` for `nvidia_cublas`), not because of the CLI install path.

Merge/conflict validation:

- Merged current `upstream/main` and resolved conflicts in `pr-test.yml`, `release.yml`, and `nightly-release.yml`, preserving both `FLASHINFER_JIT_CACHE_SM_FAMILY` forwarding and upstream sccache/NVCC env forwarding.
- Parsed the touched workflow YAML files with `yaml.safe_load`.
- Ran `git diff --cached --check`.
- Ran `python -m py_compile flashinfer/__main__.py flashinfer/jit/env.py tests/cli/test_cli_cmds.py`.
- Ran `python -m pytest tests/cli/test_cli_cmds.py -q` (`18 passed`).

Review-comment follow-up:

- Reviewed the unresolved Gemini and CodeRabbit comments. Addressed Gemini's arch-list parsing case by accepting integer-style entries like `90`, `100`, and `120` in the SM-family filter, and moved the duplicated SM-family helpers into `build_utils.py` for reuse by both the CLI and jit-cache build backend.
- Addressed CodeRabbit's `--nightly` concern by rejecting nightly installs when the resolved FlashInfer version is not a dev release; dev versions still exact-pin the matching SM-family wheel, e.g. `flashinfer-jit-cache==0.6.11.dev20260508+cu130.sm12x`.
- Ran `python -m py_compile build_utils.py flashinfer/__main__.py flashinfer-jit-cache/build_backend.py tests/cli/test_cli_cmds.py`.
- Ran `python -m pytest tests/cli/test_cli_cmds.py -q` (`21 passed`, with the expected PyTorch GB10 capability warning from this host's torch build).
- Ran `git diff --check`.
- Ran release CLI dry-run: `flashinfer install-jit-cache-wheel --cuda-version cu130 --sm-family sm12x --dry-run`, which resolves `flashinfer-jit-cache==0.6.11+cu130.sm12x` and the `uv pip install --python ...` command.
- Ran nightly CLI dry-runs for both paths: stable `0.6.11` now fails early with the new explanatory error, while explicit `0.6.11.dev20260508` resolves `flashinfer-jit-cache==0.6.11.dev20260508+cu130.sm12x` against `https://flashinfer.ai/whl/nightly/cu130` with `--pre`.

Human feedback follow-up:

- Updated `flashinfer install-jit-cache-wheel` autodetection to inspect every visible CUDA device instead of only device 0. It selects a wheel only when the visible GPUs are covered by one jit-cache SM-family wheel, and otherwise fails with guidance to pass `--sm-family` or build from source with an explicit `FLASHINFER_CUDA_ARCH_LIST`.
- Treated Blackwell-family wheels as `sm80` base arch plus native Blackwell archs. The build-side family filter now keeps/adds `8.0` for `sm10x` and `sm12x` only when a native arch for that family is present.
- Kept `sm12x` default arch lists on `12.0f`; the family-specific `sm120f` target covers DGX Spark / GB10 (`sm121`) without adding an exact `12.1a` target by default.
- Added installed-wheel compatibility validation for `flashinfer-jit-cache` local-version SM suffixes. On CUDA hosts, a wrong-family installed wheel now fails fast; on this DGX Spark, `0.6.11+cu130.sm12x` validates and `0.6.11+cu130.sm9x` fails with an expected-family error.
- Reused the shared Python family filter in the AOT build/import test script instead of maintaining a second shell implementation.
- Ran `python -m py_compile build_utils.py flashinfer/__main__.py flashinfer/jit/env.py flashinfer-jit-cache/build_backend.py tests/cli/test_cli_cmds.py`.
- Ran `python -m pytest tests/cli/test_cli_cmds.py -q` (`27 passed`, with the expected PyTorch GB10 capability warning from this host's torch build).
- Ran `uvx ruff check ...` and `uvx ruff format --check ...` over the touched Python files.
- Ran `git diff --check`, `bash -n scripts/task_test_jit_cache_package_build_import.sh`, and parsed the touched workflow YAML files with `yaml.safe_load`.
- Verified the family filter outputs `sm10x: 8.0 10.0a 10.3a 11.0a` and `sm12x: 8.0 12.0f` for the CUDA 13.0 release arch list.
- Re-ran the DGX Spark CLI dry-run; it still detects `sm12x` and resolves `flashinfer-jit-cache==0.6.11+cu130.sm12x` using the `uv pip install --python ...` fallback.

SM121 target cleanup:

- Removed the temporary `12.1a` additions from release/nightly/default jit-cache arch lists and docs; `sm12x` now defaults to `8.0 12.0f`.
- Kept explicit `12.1a` support in the parser/filter if a user supplies it manually, but release artifacts no longer build it by default.
- Re-ran `python -m pytest tests/cli/test_cli_cmds.py -q` (`27 passed`), `uvx ruff check tests/cli/test_cli_cmds.py`, `uvx ruff format --check tests/cli/test_cli_cmds.py`, `bash -n scripts/task_test_jit_cache_package_build_import.sh`, `git diff --check`, workflow YAML parsing, and the DGX Spark CLI dry-run.

SM110 architecture split:

- Limited `11.0a` / `sm110` jit-cache build coverage to CUDA 13.0 `aarch64` release, nightly, and PR AOT build/import arch lists. CUDA 13.0 `x86_64` lists now omit `11.0a`.
- Updated source-build docs to omit `11.0a` from the generic x86-oriented examples and call out adding it for Jetson AGX Thor / T5000 aarch64 targets.
- Re-ran workflow YAML parsing, `bash -n scripts/task_test_jit_cache_package_build_import.sh`, `git diff --check`, `uvx ruff check tests/cli/test_cli_cmds.py`, and `python -m pytest tests/cli/test_cli_cmds.py -q` (`27 passed`).
- Verified the generated CUDA 13 arch lists: `x86_64` -> `7.5 8.0 8.9 9.0a 10.0a 10.3a 12.0f`; `aarch64` -> `7.5 8.0 8.9 9.0a 10.0a 10.3a 11.0a 12.0f`.

## Summary by CodeRabbit

* **New Features**
  * Added a CLI command to detect CUDA/SM family and install per‑SM‑family JIT‑cache wheels (nightly, dry‑run, and explicit overrides supported).

* **Documentation**
  * Updated README and installation docs with the new CLI, per‑SM‑family wheel scheme, offline init, source‑build flags, and nightly install guidance.

* **Tests / CI**
  * CI matrices, workflows, and tests updated to build, name, upload, and consume JIT‑cache artifacts per (CUDA, arch, SM‑family).

* **Chores**
  * Build/index tooling and version handling improved to produce and recognize SM‑family segmented wheels.

### CODE REVIEW

File: flashinfer/__main__.py [hunk 1]
```diff
@@ -60,6 +72,29 @@ def _ensure_modules_registered():
...
+            "Install torch or pass --sm-family explicitly."
+        ) from e
+
+    if not torch.cuda.is_available() or torch.cuda.device_count() == 0:
+        raise click.ClickException(
+            "No CUDA device visible — cannot autodetect SM family. "
+            "Pass --sm-family {sm9x|sm10x|sm12x} or run on a host with the target GPU."
+        )
+
+    major, minor = torch.cuda.get_device_capability(0)
```
kahyunnam: how would we handle multiple GPU systems? this only detects index 0, right?

kahyunnam (commented):
re: the 3 GPU SM families in the PR description, I think @aleozlx  mentioned earlier in the thread that each device typically requires 8.0 plus their native arch -- should we add sm80a compialtion to sm10x and sm12x subwheels as well?

yongwww:
`sm10x - Datacenter Blackwell (sm100a/103a/110a)`, this is a bit confusing to me: sm10x does not seem to include sm_110a, and sm_110a maps to Thor/Tegra rather than datacenter Blackwell

docs/installation.rst
```diff
@@ -43,11 +43,28 @@ FlashInfer provides three packages:
 .. code-block:: bash
 
     pip install flashinfer-python flashinfer-cubin
-    # JIT cache package (replace cu129 with your CUDA version: cu128, cu129, or cu130)
-    pip install flashinfer-jit-cache --index-url https://flashinfer.ai/whl/cu129
+    # JIT cache package: autodetects CUDA + GPU SM family and runs the right pip install.
+    flashinfer install-jit-cache-wheel
```
yongwww, line 47: using the command `flashinfer install-jit-cache-wheel` to install pip wheel is weird to me, and downstream projects will have to change accordingly. Would be good to get comments from other folks. cc @yzh119  @aleozlx  @nv-yunzheq

dierksen: #3033 actually requested this explicitly, which is why I folded it into here.

The existing install methods should still work (just with additional parameters). We also already have a CLI function to install cubins, so this seems to be a logical extension.

dierksen:
> `sm10x - Datacenter Blackwell (sm100a/103a/110a)`, this is a bit confusing to me: sm10x does not seem to include sm_110a, and sm_110a maps to Thor/Tegra rather than datacenter Blackwell

Yeah, that one in particular does rub me the wrong way, but if I understand correctly, SM110 (previously SM101) maps closer to the datacenter Blackwell chips than to the SM120f chips.

That said, maybe we should split things up more to align better with different systems? Using the table from https://developer.nvidia.com/cuda/gpus and our support matrix from https://github.com/flashinfer-ai/flashinfer/#gpu-support, we could probably split things up like:

SM75-89: x86_64 and aarch64
SM90: x86_64 and aarch64
SM100: x86_64 and aarch64
SM110: aarch64 only
SM120f: x86_64 and aarch64

Thor stands out as relatively unique, and this would help keep the binary size lower for all of them.

dierksen:
Looking at the breakdown in jit-cache wheel sizes:

For cu130:

  | split | x86_64 | aarch64 |
  |---|---:|---:|
  | SM75-89 | ~474 MB | ~469 MB |
  | SM90 + SM80 | ~526 MB | ~517 MB |
  | SM100 + SM80, without SM103 | ~503 MB | ~484 MB |
  | SM100+SM103 + SM80 | ~957 MB | ~920 MB |
  | SM103 + SM80, if separate | ~431 MB | ~418 MB |
  | SM110 + SM80 | n/a | ~354 MB |
  | SM120f + SM80 | 437 MB | 428 MB |

Currently in this PR:

  | wheel | x86_64 | aarch64 |
  |---|---:|---:|
  | sm9x = 75/80/89/90 | 820 MB | 808 MB |
  | sm10x = 80/100/103 plus 110 on aarch64 | 957 MB | 1,215 MB |
  | sm12x = 80/120f | 437 MB | 428 MB |

And the original:

  | wheel | size |
  |---|---:|
  | x86_64 | 1,883 MB |
  | aarch64 | 2,139 MB |


Another thing that changes is that if we align with PyTorch and shift to 12.6/13.0/13.2, we might be able to drop SM100+ on 12.x and only provide those for 13.x.

dierksen:
Questions from meeting today:

- should we just break the wheels down to individual SMs?
- who else needs to review this for downstream dependencies?
- can we make explicit dependencies for the higher-SM wheels on SM 80 and split that out?
- could we potentially get these small enough to host on pypi?

README.md
```diff
@@ -102,10 +102,13 @@ pip install flashinfer-python
 
 ```bash
 pip install flashinfer-python flashinfer-cubin
-# JIT cache (replace cu129 with your CUDA version)
-pip install flashinfer-jit-cache --index-url https://flashinfer.ai/whl/cu129
+# JIT cache: autodetect CUDA + GPU SM family and run the matching pip install.
+# Use --dry-run to preview, --sm-family / --cuda-version to override.
+flashinfer install-jit-cache-wheel
```
mgoin, line 107: Can you make sure to test this within regular python venvs and also with `uv` venvs? I commonly use `uv pip install` so I forget if there sometimes can be issues mixing `pip install` and `uv pip install` within the same environment

build_utils.py
```diff
@@ -21,6 +21,14 @@
 from typing import Optional
 
 
+SM_FAMILY_ORDER = ("sm9x", "sm10x", "sm110", "sm12x")
+SM_FAMILY_BASE_ARCHS = {
+    "sm10x": ("8.0",),
+    "sm110": ("8.0",),
+    "sm12x": ("8.0",),
+}
```
mgoin, line 29: Why are these all 8.0 for base arches?

dierksen: 8.0 is needed in all of the wheels as a baseline (IIUC, @aleozlx can clarify), and this is just how it's currently encoding that for the 10.0+ wheels. We could also change to just have everything explicitly list 8.0 in each of them.

mgoin:
Making a note from offline discussion that we will still need something like `flashinfer download-jit-cache --all` , because we need to download kernels for all SMs when building multi-arch docker images, like in vLLM

dierksen:
> Making a note from offline discussion that we will still need something like `flashinfer download-jit-cache --all` , because we need to download kernels for all SMs when building multi-arch docker images, like in vLLM

This might require some additional planning to implement. With the current setup, I'm not sure if the .so's would resolve properly if we download them separately and then try to access them. 

What we might need to do instead is parcel out the existing wheel structure into shards that are under the 2GB release artifact limit enforced by GitHub, then reconstruct that after downloading the shards, only when using the `--all` flag proposed. That way, users who just want to use the standard `pip install flashinfer-jit-cache --index-url=...` can still do so with the smaller packages provided, or you can use the new `flashinfer install-jit-cache-wheels --all` to install the full 2GB+ package.

Any thoughts on this implementation or other suggestions? Right now we can't release at all due to the GitHub artifact cap, so we're kind of stuck until we change _something_.

PR #3277, opened 2026-05-09
comm: avoid torch symmetric memory by default for TRTLLM allreduce workspaces

mmangkad:
## 📌 Description

PR #2955 changed TRTLLM allreduce workspace allocation to use `torch.distributed._symmetric_memory` by default. After that change, every TRTLLM allreduce workspace creation depends on PyTorch's symmetric-memory allocation and rendezvous behavior. In practice this path appears less stable for these communication kernels than the previous FlashInfer/TensorRT-style `SymmDeviceMemory` allocator: I encountered fabric-level TRTLLM allreduce/fusion failures on one SM103 system, while another SM103 system worked, which suggests the default path may be sensitive to platform, fabric, driver, or environment differences. There may also be other failure modes from using torch symmetric memory as the default here that have not surfaced yet.

This PR restores the FlashInfer/TensorRT-style `SymmDeviceMemory` allocator as the default for TRTLLM allreduce workspaces. That keeps the default allreduce workspace path closer to the original TRTLLM implementation and avoids making PyTorch symmetric memory a hard dependency for these kernels. The torch symmetric-memory path is still available as an explicit opt-in for deployments that want to test or use it by setting `FLASHINFER_TRTLLM_AR_USE_TORCH_SYMM_MEM=1`.

## 🔍 Related Issues

Related to #2955.

### ✅ Pre-commit Checks

## 🧪 Tests

## Reviewer Notes

This PR does not remove the torch symmetric-memory path. It changes it from the default allocator to an opt-in allocator for TRTLLM allreduce workspaces.

## Summary by CodeRabbit

* **Refactor**
  * Workspace allocation for TRTLLM and MNNVL all‑reduce fusion now supports an optional Torch symmetric‑memory mode and unifies allocation/teardown across modes.
* **New Features**
  * Public APIs now expose a flag to opt into Torch symmetric memory for workspace allocation; process‑group validation is performed when enabled.
* **Documentation**
  * Docstrings clarified to describe the “group” parameter as the workspace‑allocation process group.

[Review Change Stack](https://app.coderabbit.ai/change-stack/flashinfer-ai/flashinfer/pull/3277?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/flashinfer/comm/allreduce.py
+++ b/flashinfer/comm/allreduce.py
@@ -69,7 +69,7 @@ from .trtllm_ar import trtllm_moe_finalize_allreduce_fusion
 
 from .mapping import Mapping
 
-from .mnnvl import CommBackend, SymmDeviceMemory
+from .mnnvl import CommBackend
 
 # Note: AllReduceFusionPattern and QuantizationSFLayout are pseudo-types (classes with int constants)
 # Import them for runtime use but type hint as int for mypy compatibility
@@ -107,6 +107,7 @@ class TRTLLMAllReduceFusionWorkspace(AllReduceFusionWorkspace):
         dtype: torch.dtype = torch.float16,
         comm_backend: Optional[CommBackend] = None,
         group: Optional[ProcessGroup] = None,
+        use_torch_symm_mem: bool = False,
     ):
         """
         Create TensorRT-LLM AllReduce fusion workspace.
@@ -118,7 +119,9 @@ class TRTLLMAllReduceFusionWorkspace(AllReduceFusionWorkspace):
             hidden_dim: Hidden dimension size
             dtype: Data type
             comm_backend: Communication backend
-            group: Process group for symmetric memory rendezvous. Defaults to torch.distributed.group.WORLD.
+            group: Process group for workspace allocation. Defaults to torch.distributed.group.WORLD.
+            use_torch_symm_mem: If True, use torch symmetric memory for workspace
+                allocation. Defaults to False (uses FlashInfer/TensorRT-style SymmDeviceMemory).
         """
         super().__init__(tp_size, tp_rank)
 
@@ -133,12 +136,13 @@ class TRTLLMAllReduceFusionWorkspace(AllReduceFusionWorkspace):
             create_metadata=True,
             use_fp32_lamport=dtype == torch.float32,
             use_symm_dev_mem=True,
+            use_torch_symm_mem=use_torch_symm_mem,
         )
 
         # Store essential attributes for easy access
         # Cast to 3-tuple to make linter happy, since we always call with create_metadata=True
         workspace_tuple = cast(
-            Tuple[List[List[int]], torch.Tensor, List[SymmDeviceMemory], dict],
+            Tuple[List[List[int]], torch.Tensor, List[Any], dict],
             self._internal_workspace,
         )
         self.ipc_handles = workspace_tuple[0]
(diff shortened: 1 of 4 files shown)


flashinfer/comm/trtllm_mnnvl_ar.py
```diff
@@ -169,8 +180,11 @@ def __init__(
             f"[MNNVL Allreduce] Actual allocated size: {allocated_size} bytes, Actual buffer size per lamport buffer: {self.buffer_size_bytes} bytes, total workspace: {self.workspace_size_bytes} bytes."
         )
 
-        # lamport initialize tensor to negative zero.
-        self.tensor.fill_(-0.0)
+        if self._uses_torch_symm_mem:
+            # lamport initialize tensor to negative zero.
+            self.tensor.fill_(-0.0)
+        else:
+            self.mcast_buffer_handle.lamport_initialize(mapping.tp_rank, torch.float32)
```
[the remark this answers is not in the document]
mmangkad: Thanks for the review. I checked this against the pre-#2955 MNNVL implementation in `v0.6.9rc1`, and this path already called `lamport_initialize(..., torch.float32)` regardless of the workspace `dtype`. The MNNVL workspace uses FP32 Lamport sentinels here, so switching this to the tensor dtype would be a behavior change rather than restoring the previous allocator behavior. I kept `torch.float32` and added a code comment to make that explicit.

flashinfer/comm/trtllm_mnnvl_ar.py
```diff
@@ -138,26 +139,39 @@ def __init__(
...
                 else torch.distributed.group.WORLD
             )
             group_name = group.group_name
+            self.ptrs, self.tensor, self.handle = _alloc_symm_buffer_bytes(
+                requested_workspace_size,
+                mapping.tp_size,
+                torch.float32,
+                device,
+                group_name,
+            )
```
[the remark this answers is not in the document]
mmangkad: Thanks, this is a valid case for the torch symmetric-memory opt-in path. I added a runtime guard before `_alloc_symm_buffer_bytes` that checks the `TorchDistBackend` group size/rank against `Mapping.tp_size/tp_rank` and raises a clear `ValueError` on mismatch. That keeps the peer count used for rendezvous and the later `buffer_ptrs[self.rank]` indexing consistent.

flashinfer/comm/trtllm_ar.py
```diff
@@ -31,7 +32,52 @@
 
 logger = logging.getLogger(__name__)
 from .cuda_ipc import cudart
-from .torch_symmetric_memory import _alloc_symm_buffer_bytes
+
+
+def _use_torch_symm_mem_for_trtllm_ar(device: torch.device) -> bool:
+    """Return whether TRTLLM allreduce workspaces should use torch symm memory."""
+    if os.getenv("FLASHINFER_TRTLLM_AR_USE_TORCH_SYMM_MEM", "0") == "1":
```
nvpohanh: Let's avoid using env var for this. I think we should make this an explicit FlashInfer interface (e.g. a function argument)

mmangkad: @nvpohanh updated

flashinfer/comm/allreduce.py
```diff
@@ -107,6 +107,7 @@ def __init__(
         dtype: torch.dtype = torch.float16,
         comm_backend: Optional[CommBackend] = None,
         group: Optional[ProcessGroup] = None,
+        use_torch_symm_mem: bool = False,
```
nvpohanh, line 110: @aleozlx is this considered breaking backward compatibility? or is this fine?

PR #3286, opened 2026-05-11
feat(cute_dsl/moe): deterministic balanced autotune profile inputs

leejnau:
## Summary

Reduces fi-vs-trt parity distance for CuteDSL NVFP4 MoE autotune selections by giving the autotuner a deterministic, balanced approx-max-load distribution for `token_selected_experts` during profiling, plus seeded random tensor initializers for the other dynamic inputs. Brings fi's autotune-profile machinery into structural alignment with trt-llm's `inputs_pre_hook` mechanism (all 6 of trt-llm's CuteDSL `TuningConfig`s set this hook).

Three commits on the branch:
- `8faa2359` — Path A+C: seed all 4 `tensor_initializers` + port `TuningConfig.inputs_pre_hook` + `CuteDslMoEInputsHelper` (the load-bearing change)
- `a009ca18` — small parity-cleanup: align `use_cold_l2_cache=True` with the rest of the codebase / trt-llm (provably no-op for DeepSeek-V3, value is forward-looking; flagged separately so reviewers don't conflate it with the A+C perf claim)
- `819bd648` — pin the `CuteDslMoEInputsHelper` input-layout contract with a small unit test

## Empirical validation

10-run 30-cell sweep on NGC rc14, B200, DeepSeek-V3 (hidden=7168, intermediate=2048, num_experts=256, top_k=8):

| | baseline | with this patch | delta |
|---|---:|---:|---:|
| Overall mean \|Δ%\| from trt parity | 8.48% | **6.45%** | −2.03pp |
| Small (N=1..256) | 10.50% | 7.44% | −3.07pp |
| Mid (N=512..2048) | 6.54% | 5.80% | −0.74pp |
| Large (N=4096..16384) | 4.35% | 4.13% | −0.21pp |

Cells closer to parity: 19 / further: 7 / neutral: 19. Cells now within 1% of trt parity: 19 of 45 (was ~12 of 45 baseline).

## Trade-offs (deliberately retained for parity gain)

- N=4096 EP=1 regresses from +1.24% to +8.89% (single cell; deterministic seed lands on a sub-optimal pick at this bucket).
- N=2048 EP=16 still bimodal under the seeded distribution; mean +19.87%.

## Test plan

- [x] `pytest tests/moe/test_cute_dsl_fused_moe.py::TestInputsHelperContract` — new unit test pins `CuteDslMoEInputsHelper.inputs_pre_hook`'s input-layout contract (replaces `inputs[2]`, passes through rest, deterministic across instances).
- [x] `pytest tests/moe/test_cute_dsl_fused_moe.py` — existing CuteDSL MoE tests pass (no behavior change on the fast path).
- [x] 10-run 30-cell parity sweep on B200 rc14 — all 450 cells `parity_ok=True`; numerical equivalence preserved.

## Summary by CodeRabbit

* **New Features**
  * Optional input-preprocessing callback for autotuning profiles, preserved across derived configs
  * Deterministic MoE profiling inputs via a new helper and seeded initialization for reproducible tuning
  * Autotune runner enables a cold L2 cache option to improve MoE tuning realism

* **Tests**
  * Added CPU tests validating the input-preprocessing hook replaces the expected tensor, preserves others, and is deterministic

[Review Change Stack](https://app.coderabbit.ai/change-stack/flashinfer-ai/flashinfer/pull/3286?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/flashinfer/autotuner.py
+++ b/flashinfer/autotuner.py
@@ -319,6 +319,13 @@ class TuningConfig:
     constraint_specs: Tuple[ConstraintSpec, ...] = ()
     use_cold_l2_cache: bool = False
     use_cuda_graph: bool = False
+    # Optional callback invoked once per profile bucket, after dynamic
+    # tensors are synthesized but before the per-tactic profile loop.
+    # Receives the full list of tensors and returns a (possibly modified)
+    # list. Use this to inject a deterministic, realistic distribution
+    # for inputs whose default tensor_initializer would be random
+    # (e.g. token_selected_experts in MoE workloads).
+    inputs_pre_hook: Optional[Callable] = None
 
 
 @dataclass(unsafe_hash=True)
@@ -1195,6 +1202,11 @@ class AutoTuner:
                     if not is_cache_hit:
                         # Synthesize inputs only on the profiling path.
                         tensors = self._prepare_input_tensors(p, inputs)
+                        # Apply the optional inputs_pre_hook to inject a
+                        # deterministic / realistic distribution before
+                        # the per-tactic profile loop.
+                        if tuning_config.inputs_pre_hook is not None:
+                            tensors = list(tuning_config.inputs_pre_hook(tensors))
                         if pbar is None:
                             pbar = tqdm.tqdm(
                                 total=len(profiles),
--- a/flashinfer/autotuner.py
+++ b/flashinfer/autotuner.py
@@ -1035,6 +1035,7 @@ class AutoTuner:
             constraint_specs=tuning_config.constraint_specs,
             use_cold_l2_cache=tuning_config.use_cold_l2_cache,
             use_cuda_graph=tuning_config.use_cuda_graph,
+            inputs_pre_hook=tuning_config.inputs_pre_hook,
         )
         self._override_config_cache.setdefault(tuning_config, {})[cache_key] = (
             new_config
--- a/flashinfer/fused_moe/cute_dsl/_inputs_helper.py
+++ b/flashinfer/fused_moe/cute_dsl/_inputs_helper.py
@@ -1,0 +1,186 @@
+"""
+Copyright (c) 2025 by FlashInfer team.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+"""Inputs helper for CuteDSL MoE autotune profiling.
+
+Provides a deterministic, balanced approx-max-load distribution for
+``token_selected_experts`` during autotune profiling, using rejection
+sampling so each token's top_k slots are independently filled.
+
+Without this hook, fi's autotune profile uses random expert assignments
+(via ``torch.randint`` in the tensor_initializers), which:
+
+  1. Makes profile-time inputs non-deterministic across process invocations,
+     causing run-to-run tactic-pick variance at marginal cells.
+
+  2. Produces a uniform-random expert distribution that does not match
+     real workload distributions; per-tactic profile times don't reflect
+     production performance differences -- e.g. tile=256 2-CTA may
+     profile competitively under uniform-random load but underperform
+     at runtime.
+
+Reference: ports trt-llm's GroupedGemmInputsHelper.generate_num_tokens_per_expert
+and generate_token_selected_experts from
+tensorrt_llm/_torch/custom_ops/cute_dsl_custom_ops.py (lines 31-167) and the
+``inputs_pre_hook`` mechanism from CuteDslFusedMoENvfp4InputsHelper in
+tensorrt_llm/_torch/modules/fused_moe/fused_moe_cute_dsl.py.
+"""
+
+import math
+from typing import List
+
+import torch
+
+
+class CuteDslMoEInputsHelper:
+    """Generates a balanced approx-max-load distribution for autotune profiling.
+
+    Used as ``TuningConfig(inputs_pre_hook=helper.inputs_pre_hook, ...)`` in
+    the CuteDSL MoE autotune setup. The hook intercepts profile-time inputs
+    after the autotuner synthesizes them and replaces ``token_selected_experts``
+    with a deterministic balanced assignment.
+
+    Args:
+        num_experts: Total number of experts in the MoE layer.
+        top_k: Number of experts each token routes to.
+        num_local_experts: Number of experts processed by this rank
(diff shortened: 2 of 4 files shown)


flashinfer/fused_moe/cute_dsl/_inputs_helper.py
```diff
@@ -0,0 +1,186 @@
...
+            limit = self.top_k - (self.num_experts - j)
+            prioritized = (
+                torch.nonzero(num_selected_experts <= limit).squeeze(-1).tolist()
+            )
+            if len(prioritized) > 0:
+                p_set = set(prioritized)
+                selection_order_j = prioritized + [
+                    i for i in selection_order_j if i not in p_set
+                ]
+            for i in selection_order_j:
```
qiching, line 162: assume when `num_tokens_on_curr_rank < num_local_experts`, the base of `divmod(num_tokens_on_curr_rank, num_local_experts)` is 0, and the remaining experts should be assigned 0 tokens. For DSv3 with EP=16 and num_tokens=1: average=0.5, extra≈0.83, num_tokens_on_curr_rank=2, divmod(2,16)=(0,2) so experts 0 and 1 should each receive 1 token, while experts 2–15 should each receive 0 tokens; however, the current logic causes experts 2–15 to incorrectly receive 1 additional token each.

it deviates.

so how do you think?

flashinfer/fused_moe/cute_dsl/_inputs_helper.py
```diff
@@ -0,0 +1,186 @@
...
+                torch.randperm(num_tokens) for _ in range(self.num_local_experts)
+            ]
+
+        for j, num_tokens_j in enumerate(num_tokens_per_expert):
+            selection_order_j = selection_orders[j].tolist()
+            # Prioritize tokens that risk not hitting top_k before all
+            # experts are exhausted. (For typical MoE configs with
+            # num_local_experts < num_experts the prioritized list is
+            # empty; preserved for parity with trt's algorithm.)
+            limit = self.top_k - (self.num_experts - j)
```
qiching, line 148: i  agree with Gemini and i think here we should use global index:
```
global_j = j + self.local_expert_offset
limit = self.top_k - (self.num_experts - global_j)
```

flashinfer/autotuner.py
```diff
@@ -319,6 +319,13 @@ class TuningConfig:
     constraint_specs: Tuple[ConstraintSpec, ...] = ()
     use_cold_l2_cache: bool = False
     use_cuda_graph: bool = False
+    # Optional callback invoked once per profile bucket, after dynamic
+    # tensors are synthesized but before the per-tactic profile loop.
+    # Receives the full list of tensors and returns a (possibly modified)
+    # list. Use this to inject a deterministic, realistic distribution
+    # for inputs whose default tensor_initializer would be random
+    # (e.g. token_selected_experts in MoE workloads).
+    inputs_pre_hook: Optional[Callable] = None
```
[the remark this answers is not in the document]
leejnau: addressed in https://github.com/flashinfer-ai/flashinfer/pull/3286/changes/36699b54046cdd3ff3c0e5eef1422a035970278c

flashinfer/fused_moe/cute_dsl/_inputs_helper.py
```diff
@@ -0,0 +1,186 @@
...
+
+        # Seeded random permutations on CPU for cross-process reproducibility.
+        with torch.random.fork_rng(devices=[]):
+            torch.manual_seed(self.seed)
+            selection_orders = [
+                torch.randperm(num_tokens) for _ in range(self.num_local_experts)
+            ]
+
+        for j, num_tokens_j in enumerate(num_tokens_per_expert):
+            selection_order_j = selection_orders[j].tolist()
```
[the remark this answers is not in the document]
leejnau: This is actually done deliberately to match the TRT-LLM behavior:

https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/custom_ops/cute_dsl_custom_ops.py#L152-L160

flashinfer/fused_moe/cute_dsl/_inputs_helper.py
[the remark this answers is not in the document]
leejnau: This is actually done deliberately to match the TRT-LLM behavior:

https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/custom_ops/cute_dsl_custom_ops.py#L146-L147

leejnau: This was deliberately done for parity with TRT-LLM.  TRT-LLM uses local `j` identically at https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/custom_ops/cute_dsl_custom_ops.py#L146-L147, and the goal is port alignment.  Switching to global `j` would deviate from that.  We could instead fix it here and file an issue to TRT-LLM.  (This would also apply to Gemini's other comment.)

flashinfer/fused_moe/cute_dsl/_inputs_helper.py
leejnau: This is the same issue Gemini flagged earlier in this file, and your example reinforces this.  However, TRT-LLM has identical behavior at https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/custom_ops/cute_dsl_custom_ops.py#L151-L159 

Since the goal is a faithful port, ensuring FlashInfer's autotune sees the same profile inputs TRT-LLM does, fixing here would deviate.  We could instead fix it here and file an issue to TRT-LLM.

qiching (approved):
Agree to merge. how about add a comment with the issue link to these two bug reports in the code to avoid repeated flagging?

leejnau:
> Agree to merge. how about add a comment with the issue link to these two bug reports in the code to avoid repeated flagging?

* Filed both as a single upstream tracking issue: https://github.com/NVIDIA/TensorRT-LLM/issues/14146
* Added inline comments at both sites pointing to the TRT-LLM issue (commit 04b28b37) so future readers don't have to re-flag.

qiching (approved):
LGTM now! thanks your work

flashinfer/fused_moe/cute_dsl/tuner.py
```diff
@@ -294,25 +304,48 @@ def __init__(
...
+                        # for cross-process determinism of autotune picks
+                        # (matches trt-llm's seed=515 convention).
                         lambda shapes, dtype, device: torch.randint(
-                            0, 256, shapes, dtype=torch.uint8, device=device
+                            0,
+                            256,
+                            shapes,
+                            dtype=torch.uint8,
+                            device=device,
+                            generator=torch.Generator(device=device).manual_seed(515),
```
nv-yunzheq, line 316: Do you want to use different seed value for tensor initialization?

leejnau: The shared seed is ok because the four lambdas draw from different shapes/dtypes/distributions, so the outputs are uncorrelated in practice despite sharing a seed.

nv-yunzheq (approved):
Approved as the unit test passed

PR #3235, opened 2026-05-05
Support Kimi K2.5 H64 CuTe DSL MLA decode

saltyminty:
https://github.com/flashinfer-ai/flashinfer/issues/3161

## Summary

- Enable CuTe DSL MLA decode for Kimi K2.5-style 64 query heads.
- Allow H=64 split-KV by padding split-KV workspace storage to the physical 128-head MMA lane width.
- Fold 64-head coverage into the existing CuTe DSL MLA decode test sweep.

## Notes

An experimental 64-wide MMA-M path was investigated separately, but it was reverted and is not part of this PR. The branch intentionally keeps the GPU-validated padded implementation.

## Validation

- `pre-commit run --files flashinfer/cute_dsl/attention/collective_builder.py flashinfer/cute_dsl/attention/mla_config.py flashinfer/cute_dsl/attention/mla_decode.py flashinfer/cute_dsl/attention/mla_decode_fp8.py flashinfer/cute_dsl/attention/scheduler/mla_persistent.py flashinfer/cute_dsl/attention/wrappers/batch_mla.py tests/attention/test_cute_dsl_mla_decode.py`
- SM100 smoke: H=64 CuTe DSL MLA public API with computed `split_kv=32`, output shape `(1, 1, 64, 512)`, dtype `torch.float16`, no NaNs.
- SM100 padded benchmark: B=1/S=128 `0.027255 ms`, B=1/S=512 `0.027000 ms`, B=4/S=128 `0.026925 ms`, B=4/S=512 `0.026979 ms`.

## Summary by CodeRabbit

* **New Features**
  * Relaxed head-count constraints so MLA decode accepts smaller head dimensions and more configurations.

* **Bug Fixes**
  * Workspace sizing now pads head dimension to 128 for accumulator/layout computations and when computing workspace for split-KV, preventing layout/address issues.
  * Removed runtime rejection for small head counts.

* **Tests**
  * Expanded MLA decode tests to cover multiple head counts and added NaN/Inf (finiteness) assertions.

PR #2805, opened 2026-03-17
[CuTe DSL] Add modular FMHA prefill and MLA decode attention kernels

pgera:
## Summary
Modular rewrite of the CuTe DSL attention kernels, refactored into composable building blocks. Includes FMHA prefill (from #1549) and MLA decode (replacing the monolithic `flashinfer/mla/cute_dsl/` kernels), with bug fixes, new attention variants, and comprehensive tests.

### Architecture
- Composable "roles" (Loader, MMA, Softmax, Correction, Epilogue) connected by declarative pipeline topologies
- Shared infrastructure across FMHA prefill and MLA decode (pipeline topology, collective builder, mainloop spec)
- MLA-specific roles for paged KV, page-table loading, and split-KV reduction

### FMHA prefill features
- bf16/fp16, head_dim 64/128, GQA/MHA, causal/sliding-window masks
- Custom logits transforms (sigmoid, sigmoid-tanh via MUFU.TANH), output transforms
- Attention variants via `score_mod` hook: ALiBi (1-D params), RPE (2-D params), SoftCapping (compile-time only)
- Attention sink with custom `update_statistics`
- Variable-length sequences (ragged layout)
- Wired into `BatchPrefillWithRaggedKVCacheWrapper` via `backend="cute-dsl"`
- TVM-FFI compile path: `make_fake_compact_tensor` for compile-time tracing, PyTorch tensors at runtime

### MLA decode features
- FP16/BF16 and FP8 (separate kernel class + roles)
- Paged KV cache with configurable page sizes
- Persistent tile scheduler with split-KV reduction
- Replaces ~9,700 lines of monolithic code with modular roles

### Bug fixes vs #1549
- **Causal mask boundary + PV accumulate**: fixed off-by-one in causal mask and accumulation bug
- **Attention sink dtype mismatch**: wrapper hardcoded fp16 for sink tensor regardless of input dtype
- **Attention sink M_D_update domain**: corrected domain conversion and exp2 scaling in online softmax
- **Sliding window mask** (4 issues): missing left-bound check, incorrect trip count/start block for symmetric windows, KV coordinate offset mismatch, and Q/K offset for unequal sequence lengths
- **SigmoidAttention bias**: bias not converted to log-base-2 domain

## Key files
| Path | Description |
|------|-------------|
| `flashinfer/cute_dsl/attention/prefill.py` | FMHA prefill kernel |
| `flashinfer/cute_dsl/attention/mla_decode.py` | MLA decode kernel (FP16/BF16) |
[...]

PR #3296, opened 2026-05-12
Restore monolithic CuTe-DSL MLA decode alongside modular, gated by cute_dsl_impl=

pgera:
## Summary

PR #2805 refactored the monolithic CuTe-DSL MLA decode kernel into a modular structure and removed the original implementation. The original authors want it kept available because the modular path is still maturing. This PR restores it under the existing \`backend=\"cute-dsl\"\` user surface (no new backend name) and exposes implementation selection via a new \`cute_dsl_impl=\` keyword argument on \`trtllm_batch_decode_with_kv_cache_mla\`.

- **\"auto\"** (default): monolithic by default, automatically promoted to modular when the call uses a modular-only feature (currently \`sinks\`).
- **\"modular\"**: strict, always run the modular kernels.
- **\"monolithic\"**: strict, always run the monolithic kernels; raises \`ValueError\` if the call uses any modular-only feature.

The dispatcher strips modular-only kwargs (\`sinks=None\`) before forwarding to monolithic, so callers can pass \`sinks=\` unconditionally without breaking the monolithic path.

### Sinks support on cute-dsl backend

\`trtllm_batch_decode_with_kv_cache_mla(sinks=...)\` on \`backend=\"cute-dsl\"\` now constructs an \`AttentionWithSink\` variant inside the modular standalone, instead of being rejected at the API boundary. \`AttentionWithSink\` gained value-based \`__hash__\`/\`__eq__\` (keyed on \`(type, shape, dtype)\`) so \`@functools.cache\` on \`_compile_mla_kernel\` correctly reuses compiled kernels across invocations with the same shape — without this, a fresh variant per call hashed by object identity, JIT-recompiled the kernel on every iteration, and made cuda-graph + sinks bench measurements appear to hang.

### Layout

\`\`\`
flashinfer/cute_dsl/attention/
  monolithic/         - restored kernels (verbatim from before #2805,
                        relocated to live next to the modular code).
                        Includes the H<128 / Kimi K2.5 fix from #3235
                        backported.
  wrappers/           - existing modular standalone + wrapper.
  mla_dispatch.py     - new dispatcher in front of both impls.
\`\`\`

### Bench

\`benchmarks/bench_trtllm_gen_mla.py\` grows a focused 6-cell \`with_sinks=True\` sub-sweep (B in {1,16,128} × S in {1024,8192} at q_len=1, page=64, bf16) on top of the existing main sweep, instead of doubling the full grid. Argument list deduplicated into a \`common_kwargs\` dict so warmup and benchmark calls cannot drift.

Sinks overhead is ~free on both backends (worst case +1.9% at the smallest cell). Cross-backend ranking does not change with sinks enabled.

## Test plan

Existing standalone and public-API tests in \`tests/attention/test_cute_dsl_mla_decode.py\` now parametrize over modular/monolithic via a \`cute_dsl_impl\` fixture, doubling coverage on the same shapes. New minimal sinks tests pin the auto/modular dispatch branches and the monolithic+sinks \`ValueError\` contract. Wrapper sinks numerics remain covered by the pre-existing \`test_cute_dsl_mla_decode_attention_sink\`.

- [x] All pre-commit hooks pass on changed files (mypy, ruff check, ruff format, EOF, whitespace, etc.).
- [x] \`pytest tests/attention/test_cute_dsl_mla_decode.py -v\` — full sweep (544 cases incl. parametrized modular/monolithic) passes on B200.
- [x] \`pytest tests/attention/test_cute_dsl_mla_decode.py -k sinks\` — 3 new sinks integration tests pass.
- [x] \`pytest tests/attention/test_trtllm_gen_mla.py -v\` — unaffected, passes.
- [x] H=64 / Kimi K2.5 backport on monolithic exercised via the existing \`num_heads ∈ [128, 64]\` parametrization × \`cute_dsl_impl=monolithic\`.
- [x] Bench \`benchmarks/bench_trtllm_gen_mla.py --backend cute-dsl\` and \`--backend trtllm-gen\` both run cleanly through the focused sinks sub-sweep.

## Summary by CodeRabbit

* **New Features**
  * Optional "sinks" support for an alternate softmax path and a cute_dsl_impl option to choose/auto-select modular vs monolithic MLA decode implementations.
  * New monolithic CuTe-based MLA kernel targeting Blackwell hardware.

* **Performance / Reliability**
  * Improved kernel caching/variant keying to enable reuse across variant instances.
  * Benchmark updated to exercise sinks-enabled and sinks-disabled paths.

* **Tests**
  * Added tests for implementation selection, sinks behavior, and related error/shape validations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/flashinfer-ai/flashinfer/pull/3296?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


flashinfer/cute_dsl/attention/monolithic/mla_decode_fp16.py
```diff
@@ -0,0 +1,4259 @@
...
+        if H > 128:
+            return False
+        if S < 1 or S > 4:
+            return False
+        if K <= 0:
+            return False
+        return True
+
+
+def run(
```
Observer007: Should we remove the `run` related functions?

pgera: Yes, removed.

saltyminty (approved):
Approved – unsure if the comments about assert statements is a real issue or intended behavior, please check.

saltyminty:
~~/bot run~~
nvm, needs to resolve merge conflicts first

PR #3299, opened 2026-05-12
fix(fmha_v2): two small perf/correctness wins on Hopper FP8 prefill

jdebache:
Two independent fixes for the Hopper warp-spec FP8 path of `trtllm_fmha_v2_prefill`, found while
  profiling against FA3 on long-context FP8 paged prefill. Each commit's message has the full
  investigation; this is the summary.

  ## Commits

  - **`fix: gate skip-softmax vote-write and named-barrier behind ENABLE_SKIP_SOFTMAX`**
    In `csrc/fmha_v2/fmha/warpspec/compute.h`, the per-tile `*skip_softmax_vote = 1` store and the
    matching `named_barrier_wait(SKIP_SOFTMAX_BARRIER_ID + threadIdx.x/128, 128)` were emitted
    unconditionally, even for kernel variants compiled with `ENABLE_SKIP_SOFTMAX = false`. The
    companion read path in `warpspec/epilogue.h` is already gated, so the unconditional write +
    barrier were paying a per-KV-iteration cost for a feature the kernel was not using. Wrap both
    in `if constexpr (Kernel_traits::ENABLE_SKIP_SOFTMAX)`.

  - **`fix: nullify softmax_stats_ptr when caller does not request LSE`**
    In `csrc/fmha_v2_run.cu`, the binding was unconditionally allocating a workspace-backed
    scratch buffer for softmax stats when `save_softmax_stats=False`, leaving
    `params.softmax_stats_ptr` non-null. The dispatcher's runtime predicate
    `params.softmax_stats_ptr != nullptr` then unconditionally selected the heavier
    `return_softmax_stats=True` kernel variant, which performs per-row cross-warp shuffles and
    writes ~50 MB of stats to global memory per call at long-context prefill shapes — even when no
    caller wanted them. Set `softmax_stats_ptr = nullptr` when no LSE tensor is provided.

  ## Measurements (GH200 144G, SM90, CUDA 13.0, FP8 e4m3, paged prefill, causal, s=65536)

  | Shape | Pre-fix v2 | Post-fix v2 | FA3 | Δ from fixes |
  |---|---|---|---|---|
  | hd=256, 4Q/4KV, page_size=16 | 16.08 ms | 15.85 ms | 14.20 ms | -1.4% |
  | hd=128, 96Q/8KV, page_size=64 | 181.17 ms | 179.03 ms | 161.75 ms | -1.2% |

  Both fixes are correctness/cleanliness wins; perf impact is modest but consistent. The
  softmax_stats fix in particular eliminates ~50 MB of unnecessary global writes per call at
  typical long-context prefill shapes.

  ## Test plan

  - [ ] `pytest tests/attention/test_*` (especially anything that exercises
    `trtllm_fmha_v2_prefill` or warp-spec FP8 paged prefill).
  - [ ] Confirm kernel-name change via ncu: post-fix v2 kernel should not have `_softmax_` substring
    when `save_softmax_stats=False`.
  - [ ] No-op check: callers that pass `save_softmax_stats=True` should still receive correct LSE
    output.

## Release Notes

* **Bug Fixes**
  * Enabled FP8 (e4m3) execution on SM90 by removing the prior “known to hang” restriction/warning so the FP8 path runs normally.
  * Reduced skip-softmax overhead by only performing the related setup and synchronization when skip-softmax is enabled.

* **Tests**
  * Updated attention prefill tests to run FP8 (e4m3) cases on SM90, including skip-softmax scenarios, and verify results.


nvpohanh:
@jdebache Could you rebase this PR? thanks!

jimmyzho (approved):

nv-yunzheq (approved):
Approve as @jimmyzho have reviewed. And there is no issue with unit test

PR #3330, opened 2026-05-15
feat(trace): add check callbacks to trace templates

### LINKED ISSUE #3266: Feature request: add check() method to fi_trace TraceTemplate for built-in correctness validation
## Summary

Add a `check()` callable field to `TraceTemplate` (in `flashinfer/trace/template.py`) and **emit its source into the generated trace JSON**, alongside the existing `init()` and `reference()` fields. The `check()` callable takes two sets of outputs — the kernel-under-test outputs and the reference outputs — performs standard correctness checks (closeness, mismatch percentage, cosine similarity), and returns `True` when the actual outputs are validated against the reference.

## Motivation

`TraceTemplate` already declares an `outputs` schema, an `init()` callable (added in a separate PR), and an optional `reference` callable, all of which are serialized into the trace JSON (see `flashinfer/trace/template.py` and examples in `flashinfer/trace/templates/norm.py`). However, every consumer of fi_trace currently has to roll their own correctness comparison: manually invoking the reference, writing `torch.allclose` / `torch.testing.assert_close` calls, computing mismatch ratios, etc. This leads to:

- Inconsistent tolerance choices across kernels (different teams using different `atol`/`rtol`).
- Boilerplate in every test / benchmark / autotuner harness that wants to validate a candidate implementation.
- No standard way to surface "soft" correctness signals (mismatch %, cosine similarity) which are often more useful than a binary `allclose` for low-precision kernels (FP8/FP4).
- The trace JSON describes *how to run* a kernel and *what reference output to compare against*, but not *how to decide if the candidate output is acceptable* — leaving that decision implicit.

A first-class `check()` on the trace, emitted into the JSON the same way `init()` and `reference()` are, would centralize this logic and make fi_trace usable as a *correctness oracle* in addition to a benchmark schema. Downstream tools that consume trace JSON (autotuners, regression harnesses) could then load and run `check()` directly without re-implementing it.

## Proposed API

Add an optional `check` callable to `TraceTemplate.__init__`, store it as `self.check`, and serialize its source via `inspect.getsource(template.check)` into the JSON under a new `"check"` key — mirroring the existing handling of `reference` in `build_fi_trace_fn`.

```python
class TraceTemplate:
    def __init__(
        self,
        ...
        reference: Optional[Callable] = None,
        check: Optional[Callable] = None,    # NEW
        ...
    ) -> None:
        ...
        self.check = check
```

Recommended `check()` signature for template authors:

```python
def check(
    actual: Dict[str, torch.Tensor],
    reference: Dict[str, torch.Tensor],
    *,
    rtol: Optional[float] = None,    # per-dtype default if None
    atol: Optional[float] = None,    # per-dtype default if None
    max_mismatch_pct: float = 0.0,   # allowed % of elements outside tolerance
    min_cos_sim: float = 1.0 - 1e-3, # minimum acceptable cosine similarity
) -> bool:
    """Return True iff `actual` matches `reference` within thresholds.

    Both dicts are keyed by the names declared in the template's `outputs`.
    The default implementation (when authors don't override) should compute
    closeness, mismatch %, and cosine similarity per output and return True
    when every output passes all three thresholds.
    """
```

## JSON output (after change)

```json
{
  "name": "fused_add_rmsnorm_h5120",
  "op_type": "rmsnorm",
  "axes": { ... },
  "inputs": { ... },
  "outputs": { ... },
  "init": "def _fused_add_rmsnorm_init(...): ...",
  "reference": "def _fused_add_rmsnorm_reference(...): ...",
  "check": "def _fused_add_rmsnorm_check(actual, reference, ...): ..."
}
```

## Example usage

```python
from flashinfer.trace.templates.norm import rmsnorm_trace

actual    = {"output": flashinfer.rmsnorm(x, w)}
reference = {"output": rmsnorm_trace.reference(x, w)}

assert rmsnorm_trace.check(actual, reference)
```

## Implementation sketch

1. Add `check: Optional[Callable] = None` to `TraceTemplate.__init__` (`flashinfer/trace/template.py`) and store as `self.check`.
2. In `build_fi_trace_fn`, after the existing `reference` source-emission block, add an analogous block:
   ```python
   if template.check is not None:
       try:
           result["check"] = inspect.getsource(template.check)
       except (OSError, TypeError):
           pass
   ```
3. Update the `TraceTemplate` docstring to describe `check`.
4. Add a default per-dtype tolerance helper that template authors can call from their `check()` implementations (covers FP32/BF16/FP16/FP8/FP4).
5. (Follow-up) Add `check()` implementations to a handful of representative templates (rmsnorm, gemm, attention) to demonstrate the pattern; remaining templates can adopt incrementally.

## Open questions

1. Should `check()` also accept the *inputs* and call `self.reference(**inputs)` internally, so callers can pass `(inputs, actual)` instead of `(actual, reference)`? More ergonomic but couples `check()` to `reference` being present.

yyihuang:
## Summary
- add optional TraceTemplate.check and emit its source into fi_trace JSON
- add public default_check/default_tolerances helpers for reference-vs-actual output validation
- add representative checks for norm, GEMM, and single-request attention templates

Closes #3266.

## Tests
- python -m ruff check flashinfer/trace/__init__.py flashinfer/trace/template.py flashinfer/trace/templates/__init__.py flashinfer/trace/templates/attention.py flashinfer/trace/templates/gemm.py flashinfer/trace/templates/norm.py tests/trace/test_fi_trace.py
- python -m ruff format --check flashinfer/trace/__init__.py flashinfer/trace/template.py flashinfer/trace/templates/__init__.py flashinfer/trace/templates/attention.py flashinfer/trace/templates/gemm.py flashinfer/trace/templates/norm.py tests/trace/test_fi_trace.py
- python -m pytest tests/trace/test_fi_trace.py -q
- python -m pytest tests/trace/test_fi_trace_template_consistency.py -q

## Summary by CodeRabbit

* **New Features**
  * Added trace correctness validation with dtype-aware default tolerances and a standard check API; many trace templates (attention, GEMM/BMM, norm) now include validation checks and generated trace definitions may embed check logic.

* **Documentation**
  * Template docs updated to describe optional reference/check callables and the expected check contract.

* **Tests**
  * Expanded and consolidated tests to cover default tolerances, per-template checks, and representative acceptance/rejection behaviors; test utilities unified to use the new checker.

[Review Change Stack](https://app.coderabbit.ai/change-stack/flashinfer-ai/flashinfer/pull/3330?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


nv-yunzheq:
Questions. I think there are function contain more than one output, are those functions considered in the design here? Or there is no such functions.

aleozlx (approved):

nv-yunzheq (approved):

yongwww (approved):

PR #3336, opened 2026-05-15
Replace SM120 W4A16 MoE kernels

lukealonso:
## 📌 Description

Replace the previous SM120 W4A16 MoE kernel split with a packed-route design.
The new W4A16 path groups routed token rows by expert before GEMM work, then drives the
fused FC1/FC2 kernel from packed route indices, block expert IDs, and a route count.

This removes the old static, dynamic, and micro W4A16 kernel files and adds W4A16-specific
prepared-weight, route-packing, FP4 helper, host utility, and fused-kernel modules under the
moe_w4a16_ prefix. The FlashInfer b12x MoE dispatch, workspace allocation, tracing, and tests
are updated to use the new W4A16 backend while preserving the existing caller-facing interfaces.

  ## 🔍 Related Issues

  None.

  ## 🚀 Pull Request Checklist

  ### ✅ Pre-commit Checks

  - [X] I have installed pre-commit by running pip install pre-commit (or used your preferred method).
  - [X I have installed the hooks with pre-commit install.
  - [X] I have run the hooks manually with pre-commit run --all-files and fixed any reported issues.

  ## 🧪 Tests

 ## Reviewer Notes
 
Please focus review on the W4A16 kernel replacement itself:

  - the packed-route layout and route-block metadata consumed by the fused kernel
  - the FC1/activation/FC2 dataflow over that packed routed layout
  - workspace sizing and reuse for graph-captured execution
  - preservation of the existing FlashInfer b12x MoE API surface for W4A16 callers

  The intended behavior for the full-expert path is that topk_ids are valid dense expert IDs. Skipped/nonlocal routing goes through the mapped route path.

## Summary by CodeRabbit

* **New Features**
  * Added explicit quantization mode (quant_mode) and source_format inputs, W4A16 route-packing execution path, and W4A16 weight-preparation/packing utilities.

* **Improvements**
  * Unified quant_mode-aware workspace allocation and dispatch for SM120/SM121 MoE kernels; wrapper and trace templates surface new inputs and validation.

* **Bug Fixes / Stability**
  * CUDA-graph capture now requires preallocated outputs and rejects incompatible quant_mode/source_format combinations.

* **Tests**
  * Expanded tests to cover quant_mode, W4A16 routing, workspace contracts, and CUDA-graph guards.

[Review Change Stack](https://app.coderabbit.ai/change-stack/flashinfer-ai/flashinfer/pull/3336)


aleozlx (commented):
code review - looks good

aleozlx:
pls address failed pre-commit check. thx

bkryu (approved):

aleozlx (approved):

lantudou:
@lukealonso We found a memory-lifetime issue in the W4A16 prepared-weight path introduced by this PR while integrating it with vLLM on a 32 GiB SM120 GPU. The full report and reproduction are in vllm-project/vllm#49031.

The current flow keeps both layouts resident:

1. vLLM owns the source `w1/w2` expert parameters on GPU.
2. `_get_w4a16_packed_weights()` creates full `packed_w13/packed_w2` copies.
3. `_W4A16_WEIGHT_CACHE` retains the prepared object for the process lifetime.
4. `B12xMoEWrapper.run()` still requires the source tensors on every call, and the cache key includes their `data_ptr()` values.

This means the caller cannot release or CPU-offload the source weights after preparation without invalidating the cache lookup. On an RTX 5090, a Qwen3.6-35B-A3B compressed-tensors W4A16 checkpoint fits using Marlin, but the B12x path OOMs during first-forward weight preparation because the source and packed expert layouts overlap in GPU memory.

There is also a separate preparation peak in `_nvfp4_compute_scale_factor()`: converting the complete scale tensor to FP32 and creating `nonzero_mask` caused a 768 MiB allocation in this reproduction. A scalar max reduction avoids that particular allocation, but it does not solve the persistent source-plus-packed residency.

Would you be open to exposing an explicit prepared-weight lifecycle, e.g., `prepare_weights(...) -> prepared` plus `run_prepared(..., prepared_weights)`, where execution no longer requires the source tensors? 

The exact API is not important; the useful contract for framework integrations would be:
* Prepared weights are caller-owned.
* Execution can bypass the global pointer-keyed cache.
* Preparation happens outside CUDA graph capture.
* Source tensors may be released after successful preparation.

With that contract, vLLM can prepare one MoE layer at a time in `process_weights_after_loading()`, register the prepared tensors as layer-owned buffers, and deregister the original expert parameters. A vLLM-only workaround would otherwise need to depend on FlashInfer's private CuTeDSL launch functions or inspect `_W4A16_WEIGHT_CACHE`, which would be brittle.

lazypool (commented):
nice work!

PR #3378, opened 2026-05-21
Fixes flashinfer memleak issue when using the latency backend for fused_moe

sighingnow:
Hi @wzhao18 @aleozlx  we found a host memory leak caused by the missing cache introduced in this PR: https://github.com/flashinfer-ai/flashinfer/pull/2942. After running for about 12 hours run, it will cause about 100GB host memory leak.

We have drafted a patch but not sure if it is applicable, could you folks help to check and consider if it worth merge?

Thanks!

## Summary by CodeRabbit

* **Performance Enhancements**
  * Improved Mixture of Experts (MoE) tuning configuration caching with enhanced validation logic and support for special hardware configurations.
  * Optimized hybrid token bucket computation by introducing intelligent caching mechanisms to prevent redundant calculations.
  * Refined tensor dimension handling for greater stability and improved performance in configuration management.

[Review Change Stack](https://app.coderabbit.ai/change-stack/flashinfer-ai/flashinfer/pull/3378?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/flashinfer/fused_moe/core.py
+++ b/flashinfer/fused_moe/core.py
@@ -1113,8 +1113,53 @@ def get_trtllm_moe_sm100_module():
                 tune_max_num_tokens: Upper bound for the num_tokens tuning buckets.
                 **kwargs: Extra TuningConfig kwargs (e.g. use_cold_l2_cache).
             """
-            num_experts = self.num_experts
+            # Validate hidden_states_scale layout at call time (asserts must run
+            # every call), but resolve the dynamic dim to a hashable int so the
+            # actual TuningConfig can be cached. Without caching, every MoE
+            # forward allocates a fresh TuningConfig + DynamicTensorSpec +
+            # ~5 lambdas + bucket tuple, which leaks ~25 MB Own Memory across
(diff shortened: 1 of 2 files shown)


wzhao18:
Hi @sighingnow, thank you very much for finding out the bug and putting up the fix.

Could you help me understand why the tuning config and the num_token bucket are not freed after every call? It would be helpful to understand the source of the stale memory so that we can either free it properly, or make sure not to introduce memory leaking in other allocation.

Could you answer the Gemini review? I think what it is flagging is a valid concern.

wzhao18:
One additional question: Is the workload that exhibits memleak running with CUDA graph? If so, I wonder why there would be host-side accumulation.

wzhao18:
cc @qiching for viz

sighingnow:
I'm not sure about the detail. Actually we found the regression between v0.6.7 and v0.6.9 and claude-code help to figure out the root cause.

I'll paste more after digging into the detail.

samuellees:
I'm also curious about the root cause of host mem leak

Issue #3402, opened 2026-05-25
Is it possible to enable swap_ab with `mm_fp4` `cute-dsl` backend in more small M scenarios?


b8zhong:
Hi, @nv-yunzheq, currently, it requires being a multiple of 8. In TRTLLM backend, it seems to be unconditionally enabled with 8x4 SF layout there.

I read some of the implementation in https://github.com/flashinfer-ai/flashinfer/pull/2917, but it's still a little unclear to me. But I think this could be beneficial when the tile N is small (I tried 8/16/32 with autotune, the tile sizes used by trtllm-gen), which could bring a decent speedup for M = 16, 32, so I'm wondering if it could work as well.

- https://github.com/flashinfer-ai/flashinfer/pull/3403 enables smaller tile N, so this work should be completed.

### Potential action items to do for FI team:
- Is possible to enable swap_ab for non multiples of 8 in NVFP4 GEMM, and thus take advantage of smaller tile size in these cases.

nvpohanh:
> Is possible to enable swap_ab for non multiples of 8 in NVFP4 GEMM, and thus take advantage of smaller tile size in these cases.

@nv-yunzheq Could you help to answer this? Thanks!

bkryu:
Asked @Vinnie6167 to take a look. Assigning myself to track status as @Vinnie6167 is from the core FI team but has not been added to the repo yet 😅 

The process here would involve:
1. Checking the status of `swap_ab` configs -- whether they exist but are disabled or require separate implementation.
2. If exist but disabled, see if we can enable them and expand the heuristic.
3. If they do not exist, scope out the effort to implement and decide how to proceed.

Vinnie6167:
Drafted an MR [here](https://github.com/flashinfer-ai/flashinfer/pull/3464)

Vinnie6167:
#3464 is a WIP. Need to test correctness mor.

Issue #3295, opened 2026-05-12
The autotune speed of `mm_fp4` with `backend=cute-dsl` is slow


b8zhong:
Hi, could we possibly have some offline heuristics (after collecting autotuner results), and maybe reduce the number of candidates being autotuned?

nvpohanh:
@b8zhong Could you provide some numbers about how slow it is? thanks!

b8zhong:
For 

```
python3 -m sglang.launch_server \
  --model-path nvidia/DeepSeek-V3.2-NVFP4 \
  --tp 4 \
  --ep 1 \
  --quantization modelopt_fp4 \
  --tool-call-parser deepseekv32 \
  --reasoning-parser deepseek-v3
```

(The max num tokens for autotune will be 512)

It will take **over 5 minutes** in just the GEMM autotune warmup. Maybe it's a combination of compilation and autotune increasing the number also

nvpohanh:
@b8zhong What is the accepted range? Something like <1 minute?

b8zhong:
@nvpohanh 2 minutes is fine

bkryu:
@b8zhong and @nvpohanh this might sound like a non-answer, but what if we just don't autotune with the `cute-dsl` backend?

In #2940, we implemented a heuristic for the `mm_fp4(backend='cute-dsl')` that reduced the autotuned vs. non-autotuned geomean performance gap from 16% to 2%. Currently, most of the autotuning slowness comes from compiling the kernel. Foreseeable options include (a) parallelizing compilation; (b) including the cute-dsl kernels in the jit-cache wheel; or (c) further improving the heuristic to close the 2% gap. However, both of these avenues will require some effort and given that the headroom is generally small at ~2%, I wonder whether it is worth trying to push this.

How the heuristic works, if you are interested: The heuristic works by spending <100us for the first time an (N, K) combination is seen by mm_fp4 and predicting which config will be most performant for bins of M values. Afterwards, subsequent calls on the seen (N,K) combination will simply look up the cached heuristic result (< 200 nsec). cc @Vinnie6167

b8zhong:
@bkryu , sure. However, currently we just run a dummy forward pass, so we can't really exclude it in that case (other than maybe doing some monkeypatching, but that seems potentially unstable). And another problem is the common case is NVFP4 + MoE, and we still want autotune for `trtllm-gen` MoE, so we also don't want to disable it entirely.

[...]

PR #3403, opened 2026-05-25
Enable smaller tile N for SM100 Cute-DSL NVFP4 GEMM

b8zhong:
## 📌 Description

Use tile N = 8, 16, 32

image

Note: The speedup of M = 1 is actually not due to narrow N (as it could not use swapAB), but actually because the autotuner is consistently picking incorrect tactic. Just hardcode it in these cases

## 🔍 Related Issues
https://github.com/flashinfer-ai/flashinfer/issues/3402

### ✅ Pre-commit Checks

## 🧪 Tests

on SM103:
```
pytest tests/gemm/test_mm_fp4.py
========================================================================= test session starts ==========================================================================
platform linux -- Python 3.12.3, pytest-9.0.3, pluggy-1.6.0
rootdir: /sgl-workspace/flashinfer
configfile: pytest.ini
plugins: anyio-4.13.0, typeguard-4.5.1
collected 12072 items  
...
tests/gemm/test_mm_fp4.py: 310 warnings
  /sgl-workspace/flashinfer/flashinfer/gemm/kernels/dense_blockscaled_gemm_sm100.py:1365: DeprecationWarning: The 'alignment' parameter of CooperativeGroup's constructor is deprecated and will be removed in a subsequent release, please remove it from your code.
    c_producer_group = pipeline.CooperativeGroup(

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
==================================================== 3168 passed, 8904 skipped, 2230 warnings in 576.26s (0:09:36) =====================================================
```

## Summary by CodeRabbit

* **Improvements**
  * Broader SM100 block‑scaled GEMM tiling options and refined selection to better handle small‑N and medium‑N workloads while skipping invalid small‑N choices.
  * Added stricter runtime validation to reject unsupported or inefficient kernel configurations and enforce additional constraints for certain tiling cases.

* **Tests**
  * Expanded FP4 GEMM test coverage with additional m dimension values to exercise the new tiling and validation paths.

[Review Change Stack](https://app.coderabbit.ai/change-stack/flashinfer-ai/flashinfer/pull/3403?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


flashinfer/gemm/gemm_base.py
```diff
@@ -4261,9 +4264,28 @@ def _get_sm100_block_scaled_tactics(
...
+            (128, 32),
+            (128, 16),
+            (128, 8),
+            (256, 32),
+            (256, 16),
+            (256, 8),
+        ]
+    else:
+        _tile_candidates = _SM100_MMA_TILER_MN_CANDIDATES
+    _swap_ab_candidates = (True,) if m == 64 else (False, True)
```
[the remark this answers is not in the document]
b8zhong: Good point.

bkryu:
@Vinnie6167 can you help review the heuristic here?

bkryu:
> Note: The speedup of M = 1 is actually not due to narrow N (as it could not use swapAB), but actually because the autotuner is consistently picking incorrect tactic. Just hardcode it in these cases


@b8zhong, regarding this, there is a proposed fix that is being examined at https://github.com/flashinfer-ai/flashinfer/pull/3373

flashinfer/gemm/gemm_base.py
```diff
@@ -4261,9 +4264,28 @@ def _get_sm100_block_scaled_tactics(
...
+            _tile_candidates = [(128, 64)]
+    elif m in (8, 16):
+        _tile_candidates = [
+            (128, 32),
+            (128, 16),
+            (128, 8),
+            (256, 32),
+            (256, 16),
+            (256, 8),
+        ]
```
Vinnie6167: For (m=1) and (m in {8, 16}), are these candidates set based on performance data or based on the support surface of the CuTe backend? i.e. does the CuTe backend only support the tile candidate (256, 64) when m == 1 and n <= 1024? If not, I would rather update the scoring function in utils.py to naturally select the correct tile, instead of hard-coding the correct tile size for these shapes. This will make our heuristic more generalizable.

b8zhong: It's to reduce tile candidates that are being compiled (since the cute-dsl backend autotune time is already long, https://github.com/flashinfer-ai/flashinfer/issues/3295), since otherwise it would take a long time (also I think for common LLM shape with very small num tokens, these tiles are typically chosen most of the time)

b8zhong:
Thanks @bkryu ! I think it's applicable in our case, since this gemm may only take 8 us, so that PR could fix this issue

flashinfer/gemm/gemm_base.py
Vinnie6167: In that case, can we move the config filtering logic to the Runner impelementation (`CuteDSLFp4GemmRunner`). The `_get_sm100_block_scaled_tactics` function should still return all supported configs to autotune on, and the `CuteDSLFp4GemmRunner` should control what configs we autotune over.

b8zhong: Just moved it to the runner.

flashinfer/gemm/kernels/utils.py
```diff
@@ -91,6 +94,8 @@ def _compute_tactic_for_m(rep_m, n, real_k, sm_count, m_aligned):
     best_score = -1.0
 
     for tile_m, tile_n in _SM100_MMA_TILER_MN_CANDIDATES:
+        if tile_n < 64 and prob_n > tile_n:
+            continue
```
Vinnie6167, line 98: I believe this should also be moved to the Runner?

b8zhong: I think _compute_tactic_for_m could not call can_implement, thus it would not guard against this incorrect scenario (this is needed for correctness). so I believe we can't in this case

Vinnie6167: I see, then this is fine. I may need to update this heuristic in a future MR to make sure correctness is there for all tactics returned.

bkryu (approved):

qiching (commented):
two comments:

**1. m=1 hardcode may be unnecessary after #3373**

The m=1 case is a ~7–8µs kernel where the autotuner mispicks due to inflated event timing (delay kernel too short). If #3373 merges first and the autotuner can now accurately time these fast kernels, the m=1 hardcode becomes a redundant workaround. **But it worth verifying.** the m=1 benchmark with #3373 applied and autotuner enabled that is no hardcode, to see if it picks correctly on its own.

**2. m∈{8,16} narrow tile logic is the real value here**

`tile_N < 64` didn't exist before, the speedups are genuine capability wins, not timing fixes. On the scoring vs hardcode question: I understand the compile-time concern (#3295), so restricting candidates for small-M makes sense as a practical tradeoff. But at minimum we should document _why_ these specific tiles are chosen (perf data, not just arbitrary selection), so future contributors know whether to expand or prune.

@b8zhong @bkryu

b8zhong:
(Am trying merging in main and seeing if autotuner is now correct on it's own)

b8zhong:
@qiching After testing the latest main, no. It's still not picking the right tile. So I think we'd better leave it for now
Screenshot 2026-05-27 at 12 21 30 PM

qiching:
> @qiching After testing the latest main, no. It's still not picking the right tile. So I think we'd better leave it for now Screenshot 2026-05-27 at 12 21 30 PM

got it, thanks for testing! sounds like for m=1 the kernel is so fast that even with accurate timing the autotuner cannot distinguish between candidates. makes sense to keep the hardcode. LGTM on that point!

qiching (approved):
LGTM

PR #3407, opened 2026-05-26
Unify B12x Decode with CUTLASS Prefill under a single API

askliar:
## Release Notes

* **New Features**
  * Added configurable hybrid MoE dispatch in `B12xMoEWrapper`: prefill can route to a CUTLASS-based path and fall back to b12x decoding using a `cutlass_prefill_threshold` (or an environment-variable override) for NVFP4.
  * Added support for registering CUTLASS-format prefill weights (including activation-specific parameters when needed).

* **Benchmarks**
  * Added a token-sweep benchmark to compare B12x vs CUTLASS fused-MoE performance, with optional CSV output.

* **Tests**
  * Added CUDA/SM120/SM121-gated tests covering threshold behavior, weight registration requirements, environment-variable handling, activation mapping, and cross-backend numerical parity.

### DIFF
--- a/benchmarks/b12x_vs_cutlass_moe_sweep.py
+++ b/benchmarks/b12x_vs_cutlass_moe_sweep.py
@@ -1,0 +1,347 @@
+"""
+Sweep benchmark: B12xMoEWrapper vs cutlass_fused_moe (NVFP4) on SM120/121.
+
+Charts the prefill/decode crossover so we can pin a default
+``cutlass_prefill_threshold`` inside ``B12xMoEWrapper``.
+
+For each ``num_tokens`` in the sweep we run both backends with their native
+weight packings (built from independent BF16 baselines — perf is the only
+metric, so we don't need bit-equal weights) and record median GPU time via
+CUPTI.
+
+Usage:
+    python benchmarks/b12x_vs_cutlass_moe_sweep.py \\
+        --hidden 6144 --intermediate 3072 --num_experts 128 --top_k 8 \\
+        --activation relu2 \\
+        --num_tokens 1 4 8 16 32 64 128 256 512 1024 2048 4096 8192 \\
+        --out sweep.csv
+"""
+
+import argparse
+import csv
+import sys
+from pathlib import Path
+
+import numpy as np
+import torch
+
+sys.path.insert(0, str(Path(__file__).resolve().parent))
+from routines.moe import _create_nvfp4_moe_test_data, _activation_kwarg
+from routines.moe_utils import (
+    FLOAT4_E2M1_MAX,
+    FLOAT8_E4M3_MAX,
+    compute_routing,
+)
+
+import flashinfer  # noqa: F401
+from flashinfer import (
+    ActivationType,
+    B12xMoEWrapper,
+    cutlass_fused_moe,
+    fp4_quantize,
+)
+from flashinfer.testing.utils import bench_gpu_time
+
+
+def _round_up(x: int, y: int) -> int:
+    return (x + y - 1) // y * y
+
+
+def build_cutlass_nvfp4_weights(
+    *, num_experts: int, hidden: int, intermediate: int, is_gated: bool, device
+):
+    """Build CUTLASS-NVFP4-format weights from a fresh bf16 baseline.
+
+    Mirrors the inline path in benchmarks/routines/moe.py::testCutlassFusedMoe
+    (variant="nvfp4"). Returns the tensors that ``cutlass_fused_moe`` expects.
+    """
+    e = num_experts
+    n = intermediate
+    k = hidden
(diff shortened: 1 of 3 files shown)


tests/moe/test_b12x_cutlass_hybrid.py
```diff
@@ -0,0 +1,399 @@
+"""
```
qiching: i think the tests lack any numerical correctness checks:

all assertions only check shape, dtype, `isfinite`, and dispatch call counts and by my understanding, the core premise of the entire PR is that “b12x and Cutlass weights share FP4 numeric values,” but not a single test verifies whether the outputs of the two paths are consistent near the threshold. Assume there is any mismatch in the Cutlass prefill path regarding quantization or activation mapping, the tests will still fail even if they all show passed.

And , `_make_b12x_weights(seed=0)` and `_make_cutlass_weights(seed=1)` use completely independent random weights, so even if we wanted to perform cross check, we couldn’t. So my recommendation is we can use the same BF16 baseline to generate both layouts simultaneously, and add a `test_hybrid_matches_pure_b12x` test, and compare the hybrid output with the pure B12X output at `m == threshold`.

qiching (changes requested):
_run_cutlass_prefill takes unquantized bf16 as input, with input_sf=None, but the benchmark uses explicit pre quantization; the input conventions for `cutlass_fused_moe` are inconsistent between the two. 

should we need to clarify: When `cutlass_fused_moe` is in NVFP4 mode and given a bf16 input with `input_sf=None`, does it internally quantize using `quant_scales[0]`
 If it requires pre-quantized FP4 input, then the prefill path is incorrect. We must be confirmed, and it would be best to cover this scenario in the tests.

benchmarks/b12x_vs_cutlass_moe_sweep.py
```diff
@@ -0,0 +1,347 @@
...
+        cutlass_us = float("nan")
+        if not args.skip_cutlass:
+            out = torch.empty(m, args.hidden, dtype=torch.bfloat16, device=device)
+            x_q, x_sf = fp4_quantize(x_bf16, cutlass_w["a1_gs"])
+
+            def run_cutlass(x, te, rw, w1q, w2q, o):
+                return cutlass_fused_moe(
+                    x,
+                    te.to(torch.int),
+                    rw,
```
qiching, line 284: te.to(torch.int) and ·w1q.contiguous().view(torch.long)· are executed every time a `repeat` loop runs, which unnecessarily increases the execution time of CUTLASS. We need to move these conversions to the `cutlass_args` preparation phase.

qiching (changes requested):

benchmarks/b12x_vs_cutlass_moe_sweep.py
```diff
@@ -0,0 +1,347 @@
...
+
+
+def parse_args():
+    p = argparse.ArgumentParser()
+    p.add_argument("--hidden", type=int, required=True)
+    p.add_argument("--intermediate", type=int, required=True)
+    p.add_argument("--num_experts", type=int, required=True)
+    p.add_argument("--top_k", type=int, required=True)
+    p.add_argument(
+        "--activation",
```
qiching, line 126: I will recommend to explicitly map the dictionary; if an unsupported activation is encountered, raise an exception instead of falling through.

qiching (changes requested):

flashinfer/fused_moe/cute_dsl/b12x_moe.py
```diff
@@ -279,6 +297,14 @@ def __init__(
         )
         self.source_format = source_format
 
+        env_override = os.environ.get("FLASHINFER_B12X_CUTLASS_PREFILL_THRESHOLD")
```
qiching: we must either only consult the env var when the kwarg is left at its default, or at minimum call this precedence out loudly in the docstring.

qiching (changes requested):

qiching (changes requested):
some nit on doc: call out the two real tradeoffs of enabling hybrid dispatch.

- Weight memory ~doubles. With cutlass_prefill_threshold > 0 the caller must keep both the b12x-format and the CUTLASS-format FP4 weights resident at the same time. For large MoE models this is a significant memory cost, please state it explicitly in the main class docstring (not just in register_cutlass_prefill_weights) so users can weigh it.
- CUDA graph captures a single path per shape. The b12x/CUTLASS branch is a host-side Python decision, so any captured graph freezes whichever path the token count selected at capture time. This is fine for per-shape-bucket graphs, but worth a one-line note in the docstring so users serving with CUDA graphs aren't surprised.

qiching (approved):
LGTM now

PR #3424, opened 2026-05-27
Fused FP8 blockwise MoE megakernel For BS <= 8

yugong333:
## 📌 Description

Design a FP8 quantized MoE mega kernel to accelerate the MoE model inference. The proposed kernel fuses the token routing, quantization and SiLU together into a big kernel, and iterates the expert in parallel.

This kernel is designed specific for 
- TopK = 8
- FP8 blockwise quantization with block size [128, 128]
- Hidden size = 2048
- MoE intermediate size = 512

Some initial results compared with Triton kernel in vLLM:
image

Run `python flashinfer/benchmarks/bench_megamoe.py` to Compared with cutlass in Flashinfer (without routing):

           shape |               mono_moe |   cutlass (no routing) |  speedup |  
         m=1,k=8 |     0.0185 ± 0.0001 ms |     0.0812 ± 0.0002 ms |    4.38x |  
         m=2,k=8 |     0.0215 ± 0.0001 ms |     0.0890 ± 0.0001 ms |    4.14x |  
         m=4,k=8 |     0.0383 ± 0.0002 ms |     0.0970 ± 0.0002 ms |    2.54x |  
         m=8,k=8 |     0.0627 ± 0.0001 ms |     0.1073 ± 0.0002 ms |    1.71x |  

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added fused Mixture-of-Experts (MoE) kernel with block-FP8 quantization for NVIDIA Hopper (SM90a) architecture
  * Added `mono_moe` API supporting configurable top-K expert routing with softmax/sigmoid scoring
  * Includes in-kernel routing and renormalization for optimized inference

* **Tests**
  * Added correctness validation tests comparing new implementation to reference
  * Added performance benchmarking suite

* **Documentation**
  * Added trace templates and examples for MoE operations


csrc/fused_moe/megamoe/README.md
```diff
@@ -0,0 +1,188 @@
...
+```
+
+Capture was impossible under the previous cooperative launch path.
+
+### Further reading
+
+The full design — seed correctness argument, fence discipline,
+per-site producer/consumer-set analysis, and the Phase-2a layout
+alignment that enables `expert_barrier` at site #2 — lives in
+`.kiro/specs/moe-monokernel-software-grid-sync/design.md`.
```
kailashbuki: Remove the reference to kiro-cli artifact.

csrc/fused_moe/megamoe/tests/README.md
```diff
@@ -0,0 +1,45 @@
+# moe_monokernel standalone device tests
+
+Compilable, self-contained sanity checks for the low-level PTX wrappers that
+back the TMA + WGMMA up-projection path (spec:
+`.kiro/specs/tma-wgmma-weight-load/`).
```
kailashbuki: Remove the reference to kiro-cli artifact.

csrc/fused_moe/megamoe/README.md
```diff
@@ -0,0 +1,188 @@
...
+### Why software barriers
+
+Phase boundaries inside `moe_kernel_topk` used to synchronize through
+`cudaLaunchCooperativeKernel` + `cooperative_groups::this_grid().sync()`.
+Cooperative launch blocks `cudaStreamBeginCapture`, so the kernel could
+not be captured into a CUDA Graph and paid a per-decode-step host launch
+overhead. The primitives in `src/moe_grid_barrier.h` replace that with a
+standard `<<<grid, block, smem, stream>>>` launch plus a software barrier
+built on global-memory atomic counters. See
+`.kiro/specs/moe-monokernel-software-grid-sync/` for the full spec.
```
kailashbuki: Remove the reference to kiro-cli artefact.

csrc/fused_moe/monomoe/src/moe_tma.cu
```diff
@@ -0,0 +1,269 @@
...
+      global_strides, box_dim, element_strides, CU_TENSOR_MAP_INTERLEAVE_NONE,
+      CU_TENSOR_MAP_SWIZZLE_128B, CU_TENSOR_MAP_L2_PROMOTION_L2_128B,
+      CU_TENSOR_MAP_FLOAT_OOB_FILL_NONE);
+
+  TORCH_CHECK(
+      res == CUDA_SUCCESS,
+      "cuTensorMapEncodeTiled failed for down-projection weights: CUresult=", static_cast<int>(res),
+      " (num_experts=", num_experts, ", K=", K, ", N=", N, ", row_box=", row_box, ")");
+
+  return desc;
```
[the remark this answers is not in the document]
yugong333: The claim is that "downstream loader/barrier and SHM layout are hard-coded to a 128×128 (16KB) tile" while the factory accepts 256. That is false against 
  current code — row_box=256 is already fully plumbed end-to-end:

  1. Loader propagates the tile size, not hardcoded 128:
    - moe_down_projection.cu:272 — DOWN_COL_TILE = CoreDims::DOWN_COL_TILE (128 or 256)
    - :284 — DOWN_COL_HALVES = DOWN_COL_TILE / 128 (1 or 2)
  2. mbarrier tx_bytes scales with the tile (the exact thing the finding says to fix):
    - :298-300 — DOWN_W_TX_BYTES_TOTAL = 16384 * DOWN_COL_HALVES * K_SUBSTEPS_DOWN → 32 KB per TMA when tile=256, not a fixed 16 KB.
    - :511 arms bar_w[0] with that computed total.
  3. SHM slot layout is sized by the tile: w_down_wgmma[slot][kk * DOWN_COL_TILE][...] (:515), and out_accum[BS][DOWN_COL_TILE] (:378,401).
  4. WGMMA A descriptor still references a single 128-row sub-atom per call (:621-629), exactly as the factory comment (moe_tma.cu:160-162) documents — so
  there's no inconsistency to reconcile.
  5. The config is real and active: moe_internal.h:256 — DOWN_COL_TILE = (use_tma && BS<=8) ? 256 : 128, and the shipped Dims_BS8_..._WGMMA_TMA variant is
  exactly that case. The wrapper passes row_box = DOWN_COL_TILE = 256.

  So the factory's row_box==128 || row_box==256 guard is correct as-is, and both branches are consistent through the loader, tx_bytes, SHM, and WGMMA
  descriptor. The reviewer was likely looking at the DOWN_W_TX_BYTES_PER_HALF = 16384 constant (:298) in isolation and missing that it's multiplied by
  DOWN_COL_HALVES.

csrc/fused_moe/monomoe/src/moe_tma.cu
```diff
@@ -0,0 +1,269 @@
...
+      // reads the pointer value, it never writes through it.
+      const_cast<void*>(activations_ptr), global_dim, global_strides, box_dim, element_strides,
+      CU_TENSOR_MAP_INTERLEAVE_NONE, CU_TENSOR_MAP_SWIZZLE_128B, CU_TENSOR_MAP_L2_PROMOTION_L2_128B,
+      CU_TENSOR_MAP_FLOAT_OOB_FILL_NONE);
+
+  TORCH_CHECK(res == CUDA_SUCCESS,
+              "cuTensorMapEncodeTiled failed for down-projection "
+              "activations: CUresult=",
+              static_cast<int>(res), " (temp_rows=", temp_rows, ", N=", N, ")");
+
```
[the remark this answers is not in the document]
yugong333: 1. The premise about TMA semantics is incorrect. The finding asserts that CU_TENSOR_MAP_FLOAT_OOB_FILL_NONE "allows out-of-bounds reads" past global_dim[1].
  It does not. Tiled-mode TMA always bounds-checks box coordinates against globalDim, regardless of the OOB-fill mode. When a box extends past globalDim[1],
  the hardware does not issue raw global-memory loads for the out-of-bounds rows — it simply doesn't fetch them. The FLOAT_OOB_FILL_* mode only controls what
  lands in the SMEM destination for those OOB coordinates (NONE = left untouched, ZERO = zero-filled); it has nothing to do with whether DRAM past the tensor
  is touched. So there is no cudaErrorIllegalAddress / unbounded read to fix here.

  2. Even setting that aside, the buffer is physically over-allocated to cover the tail. The descriptor's outer extent is TEMP_ROWS_TMA = BS*SPEC_MAX_TOPK = 
  64 (moe_internal.h:193), but the actual temp_fp8 allocation is TEMP_ROWS = BS*SPEC_MAX_TOPK + 8 = 72 rows (:184, :229). A worst-case nonempty expert starts
  at row ≤63 and the box_dim[1]=8 load reads rows [63, 71) — all within the 72-row allocation. The +8 padding is exactly a full box-height of slack. So even a
  physical overhang would stay inside allocated memory.

  3. The proposed fix would likely break the build. This descriptor's data type is CU_TENSOR_MAP_DATA_TYPE_UINT8 (:254). The CU_TENSOR_MAP_FLOAT_OOB_FILL_*
  zero-fill modes are only valid for floating-point tensor types; requesting FILL_ZERO on a UINT8 descriptor would make cuTensorMapEncodeTiled return an error
  (caught by the TVM_FFI_ICHECK at :261). So the suggested change isn't just unnecessary — it would regress a currently-working path.

csrc/fused_moe/megamoe/moe_wrapper.cu
```diff
@@ -0,0 +1,266 @@
...
+  }
+
+// TMA + WGMMA + SWIZZLE_128B variant of the BS8 path — the only BS8
+// implementation.  Selects the TMA-based weight + activation load path
+// in Phase 3 via `KernelConfig::USE_TMA = true`.  Up-projection weights
+// must be repacked via `interleave_for_tma_wgmma_up` (gate/up row
+// interleave for single-issue TMA); down-projection weights are passed
+// raw row-major (the TMA hardware applies the core-matrix XOR swizzle
+// at write time).
+MOEMONOKERNEL_TOPK_WRAPPER_IMPLEMENTATION(
```
djns99: Does this really need to be a macro? 
Macro is much worse, not least because it has bad ergonomics - disables syntax highlighting, editing it requires adding new `\` for every line, debug/crash logs do not give you the actual line/function

Can we instead use a 
```
// Declaration
template<class Dims> void moe_monokernel_topk_launcher(...) { ... } 

// Explicit instantiation
template void moe_monokernel_topk_launcher<moe_monokernel::Dims_BS8_E256_Qwen3_5_35B_BlockFP8_WGMMA_TMA>(...);
```

flashinfer/fused_moe/__init__.py
```diff
@@ -78,6 +78,13 @@
     fill_w_ptr as fill_w_ptr,
     has_bgmv_moe as has_bgmv_moe,
 )
+from .megamoe import (  # noqa: F401
+    mono_moe as mono_moe,
+    has_megamoe as has_megamoe,
+    alloc_scratchpad as alloc_scratchpad,
+    get_scratchpad_size_bytes as get_scratchpad_size_bytes,
+    interleave_for_tma_wgmma_up as interleave_for_tma_wgmma_up,
+)
```
djns99, line 87: I notice we have many different names for this `mega kernel`, `mega moe` `mono kernel` and `mono moe`. Can we pick one and stick with it?

I like "MOE Mega Kernel" the most but "MOE Mono Kernel" or "Mono MOE" is fine too.

MegaMOE is already used for DeepSeek MegaMOE, lets NOT use that. We have a WIP design for this coming. We will need to figure out how to retrofit this implementation into that interface.

csrc/fused_moe/megamoe/moe_wrapper.cu
```diff
@@ -0,0 +1,266 @@
...
+    /* Standard (non-cooperative) launch.  The kernel reaches grid-wide                            \
+       happens-before via the software Grid_Barrier / Partial_Barrier                              \
+       primitives in `src/moe_grid_barrier.h` (spec R1.1, R5.1, Design                             \
+       Component C "Launch form") rather than                                                      \
+       `cooperative_groups::this_grid().sync()`.  Using standard                                   \
+       `cudaLaunchKernel` is what lets the migrated kernel be captured                             \
+       into a CUDA Graph. */                                                                       \
+    CUDA_CHECK(cudaLaunchKernel(                                                                   \
+        (const void*)moe_kernel_topk<dims>, dim3(dims::KernelConfig::GRID_SIZE, 1, 1),             \
+        dim3(dims::KernelConfig::BLOCK_SIZE, 1, 1), kernel_args, shmem_size, stream));             \
```
djns99: Can we just use standard `moe_kernel_topk<dims><<<grid, block, shmem_size, stream>>>(...)` API? 
This will be easier to read and wont need the ugly `kernel_args` struct. This also indicates to the reader we aren't doing anything special

csrc/fused_moe/megamoe/moe_wrapper.cu
```diff
@@ -0,0 +1,266 @@
...
+        cudaFuncGetAttributes(&fa, moe_kernel_topk<dims>);                                         \
+        int sm_count = 0;                                                                          \
+        cudaDeviceGetAttribute(&sm_count, cudaDevAttrMultiProcessorCount, 0);                      \
+        int smem_opt_in = 0;                                                                       \
+        cudaDeviceGetAttribute(&smem_opt_in, cudaDevAttrMaxSharedMemoryPerBlockOptin, 0);          \
+        (void)smem_opt_in;                                                                          \
+        (void)max_blocks_per_sm;                                                                    \
+        (void)fa;                                                                                   \
+        /* Hard co-residency assertions for the software grid barrier                              \
+           (spec R4.1, R4.2, R4.3 / Design Component C "Co-residency                               \
```
djns99: What are all these spec references? Can we tidy up these comments

csrc/fused_moe/megamoe/moe_wrapper.cu
```diff
@@ -0,0 +1,266 @@
...
+                           (void*)&down_weights_desc,                                              \
+                           (void*)&down_activations_desc};                                         \
+    const cudaStream_t stream = get_stream(activations_in.device());                               \
+    CUDA_CHECK(cudaFuncSetAttribute(moe_kernel_topk<dims>,                                         \
+                                    cudaFuncAttributeMaxDynamicSharedMemorySize, shmem_size));     \
+    /* One-time diagnostic: compute and print occupancy + shmem so that a                          \
+       cooperative-launch failure is easy to diagnose. */                                          \
+    {                                                                                              \
+      static bool _diag_printed = false;                                                           \
+      if (!_diag_printed) {                                                                        \
```
djns99: This looks like debugging code, should it still be here?

csrc/fused_moe/megamoe/moe_wrapper.cu
```diff
@@ -0,0 +1,266 @@
...
+            cudaOccupancyDefault);                                                                 \
+        cudaFuncAttributes fa;                                                                     \
+        cudaFuncGetAttributes(&fa, moe_kernel_topk<dims>);                                         \
+        int sm_count = 0;                                                                          \
+        cudaDeviceGetAttribute(&sm_count, cudaDevAttrMultiProcessorCount, 0);                      \
+        int smem_opt_in = 0;                                                                       \
+        cudaDeviceGetAttribute(&smem_opt_in, cudaDevAttrMaxSharedMemoryPerBlockOptin, 0);          \
+        (void)smem_opt_in;                                                                          \
+        (void)max_blocks_per_sm;                                                                    \
+        (void)fa;                                                                                   \
```
djns99: Why retrieve these only to ignore them and never use them?

csrc/fused_moe/megamoe/moe_wrapper.cu
```diff
@@ -0,0 +1,266 @@
...
+      /* Caller-provided scratchpad: key the guard on buffer identity                              \
+         (ptr, size, device) and re-zero whenever it changes, so each                              \
+         distinct allocation is initialized exactly once.  A process-wide                          \
+         one-shot flag would be WRONG here — it zeroes only the first                             \
+         buffer ever seen and launches every subsequent DISTINCT scratchpad                        \
+         (other stream/device or a freshly-malloc'd buffer) with                                   \
+         uninitialized barrier counters, deadlocking the seed/spin protocol                        \
+         or silently corrupting results.  A reused buffer still pays the                           \
+         zero-init only on its first launch (self-maintaining ping-pong                            \
+         reset thereafter). */                                                                     \
```
djns99: Lots of wall of text comments like this, maybe these would be better expressed with a single design document and a shorter comment so the documentation is not scattered piecemeal across the many files

csrc/fused_moe/megamoe/megamoe_binding.cu
```diff
@@ -0,0 +1,39 @@
...
+// field is added — see the TEMP_FP8_OFFSET regression anchor in
+// moe_wrapper.cu).  The kernel's barrier counters live at the tail of this
+// struct, so the buffer MUST be at least this large.
+int64_t moe_monokernel_scratchpad_size() {
+  return static_cast<int64_t>(sizeof(moe_monokernel::MoEGemmSpec<
+                                     moe_monokernel::Dims_BS8_E256_Qwen3_5_35B_BlockFP8_WGMMA_TMA>));
+}
+
+TVM_FFI_DLL_EXPORT_TYPED_FUNC(
+    moe_monokernel_topk, moe_monokernel_topk_BS8_E256_Qwen3_5_35B_BlockFP8_WGMMA_TMA_impl);
```
djns99: Rather than aliasing the hardcoded function, would we be better actually having a single function called `moe_monokernel_topk` or similar (see comment on naming)

```
void moe_monokernel_topk(...)
{
   if(BS <= 8 && Experts == 256 && experts_weights_up.shape() ==  (...) && ...) {
      moe_monokernel_topk_BS8_E256_Qwen3_5_35B_BlockFP8_WGMMA_TMA_impl(...)
   } else {
      // Throw not implemented
   }
}

**Review by djns99 — COMMENTED** [review]
Just reviewed the API currently, will try find time to dive into the implementation later today.

File: csrc/fused_moe/monomoe/src/moe_down_projection.cuh [hunk 14]
```diff
@@ -0,0 +1,1005 @@
+
+#pragma once
```
djns99, line 2: If this is supposed to be included can we use `*.cuh` or `*.inl` extensions. Same with all the other files

csrc/fused_moe/megamoe/src/ptx_utils.h
```diff
@@ -0,0 +1,467 @@
+#ifndef PTX_UTILS_H
+#define PTX_UTILS_H
+
+#pragma once
+
+#include <cstdint>
+
+#include <cuda.h>
+#include <cuda_bf16.h>
+#include <cuda_fp8.h>
```
djns99: I have not checked for sure, but I think many of these operations already exist in `cute` which is already depended on by FlashInfer. Would it be worth using those so we have a centralised location?

Im ok if there are practical reasons to do it this way (not least of all compile times 😅), but might be a nice ergonomic consideration

csrc/fused_moe/monomoe/src/moe_internal.h
```diff
@@ -0,0 +1,1304 @@
...
+  std::uint32_t id;
+};
+
+// ── WGMMA / TMA opt-in detection (forward declarations) ────────────────────
+// These SFINAE helpers let `MoEGemmSpec<Dims>` pick the variant-dependent
+// `DOWN_COL_TILE` below (128 normally, 256 for the BS8 TMA+WGMMA variant
+// after Phase 2a layout alignment).  Full definitions (with detailed
+// comments) live further down the file; the forward declarations here only
+// need to expose `::value` so compile-time expressions can use them.
+template <typename Dims>
```
djns99, line 75: Not sure if this is ported from a larger implementation and there is more to follow. But can we simplify this and get rid of all these SFINAE things and just use `Dims::KernelConfig::USE_WGMMA` directly.

This might _slightly_ help with compile times, and just overall make it simpler to read

csrc/fused_moe/megamoe/src/moe_internal.h
```diff
@@ -0,0 +1,1304 @@
...
+  } grid_barrier;
+  struct {
+    uint32_t expert_slot[Dims::NUM_EXPERTS][2];
+    uint32_t colstripe_slot[DOWN_GRID][2];
+  } partial_barrier;
+};
+
+// Maximum supported dimensions for shared memory and scratchpad allocation
+// sizes
+#if USE_SMALL_SETUP
```
djns99: Is this something that should be exposed somewhere for testing? Or can we remove it?

csrc/fused_moe/megamoe/src/moe_internal.h
```diff
@@ -0,0 +1,1304 @@
...
+  static constexpr unsigned K_DIM_PADDED_W = Dims::HIDDEN_STATES;
+  static constexpr unsigned K_DIM_HALF_PADDED_A = Dims::HIDDEN_STATES / 2;
+};
+
+// 1 tile per warp
+// 20 warps x 2 params x 1k = 20k pre-fetch
+template <typename Dims>
+struct MoE_SHM {
+  using CoreDims = MoECoreDims<Dims>;
+  union U {
```
djns99: There's only one item in this union? What is the point of it?

csrc/fused_moe/megamoe/src/moe_prepare.cu
```diff
@@ -0,0 +1,18 @@
...
+#error Do not include this file directly.
+#endif
+
+#include <cstdint>
+
+#include "moe_internal.h"
+
+// This file previously contained the BS64 prepare_moe_topk_BSx_Ey function.
+// The BS64 path has been removed; only the BS8 path (prepare_moe_topk_BS8
+// in moe_routing.cu) remains.
```
djns99: Can we just remove this file then?

flashinfer/fused_moe/__init__.py
yugong333: Thank you! I use MonoMoe for now but very open to new names.

csrc/fused_moe/megamoe/moe_wrapper.cu
yugong333: I create a `DESIGN.md` doc to illustrate the overall principle, but keep the specific local comments. Please let me know if it's better to move comments to the design doc. Thanks!

docs/design_docs/monomoe_kernel.md
```diff
@@ -0,0 +1,233 @@
+# MonoMoe kernel — design
```
djns99, line 1: Does the `docs/design-docs` make more sense to place this

flashinfer/fused_moe/__init__.py
djns99: MoneMoe is fine! Thank you

csrc/fused_moe/megamoe/moe_wrapper.cu
djns99: Perfect, just make sure you point the comments to the design doc so people can find it 🙂

csrc/fused_moe/monomoe/src/moe_internal.h
```diff
@@ -0,0 +1,1092 @@
...
+struct MoE_SHM {
+  using CoreDims = MoECoreDims<Dims>;
+  // NOTE: `TinyDataWGMMA_TMA` (and its trailing static_asserts) are
+  // over-indented by one level — they were once wrapped in a
+  // `union U { ... } u;` that held multiple SHM variants.  Only this one
+  // variant remains, so the single-member (byte-identical) union wrapper was
+  // removed and the struct is now a direct member `tiny_wgmma_tma`; the body
+  // indentation is left as-is to keep the diff small.  The real SHM
+  // byte-sharing is done by the inner anonymous unions below.
```
djns99: This comment doesn't make sense in the context of a new contribution. We can clean up the old references and fix the indenting

djns99 (commented):
LGTM, happy with the resolution to most of the issues, just a couple of cleanup items I noticed on a quick second pass

docs/design_docs/monomoe_kernel.md
yugong333: Thanks a lot! Moved the design doc to `docs/-design_docs` now.

flashinfer/fused_moe/monomoe.py
```diff
@@ -0,0 +1,289 @@
...
+        out: optional bf16 output buffer ``[M, K]``; allocated if omitted.
+        scratchpad: optional reusable uint8 scratchpad from
+            :func:`alloc_scratchpad`; allocated per-call if omitted.
+        interleave_up: apply the gate/up TMA repack to ``expert_weights_up``
+            (default True).
+
+    Returns:
+        bf16 MoE output ``[M, K]``.
+    """
+    if not (1 <= top_k <= 8):
```
nv-yunzheq, line 294: Should there be a sm version check since it only supports sm90? There is a `@supported_compute_capability([90])` function decorator you could use.

nv-yunzheq (commented):
Did not review the kernel code. But the interface looks good.

flashinfer/fused_moe/monomoe.py
yugong333: Thank you for the comment! 
I have added the function here https://github.com/flashinfer-ai/flashinfer/pull/3424/changes#diff-6663f4813d35c176a97d0ddd7ba5a4e184a572c80abaecaec6252bbbb05284e7R212

jhaotingc:
Hi @nv-yunzheq the failed Blackwell test seems not relevant to this PR. The mono MoE test on H100 passed. 
May we merge this PR?

nv-yunzheq (approved):

PR #3426, opened 2026-05-28
feat(cutile): introduce cuTile backend (mm_bf16 + bmm_bf16 + gemm_fp8_nt_groupwise)

yifeis-nv:
Adds **cuTile** (CUDA Tile Python) backend support for three GEMM operations in FlashInfer, porting kernels from NVIDIA TileGym to use the public `cuda.tile` API without TileGym runtime dependencies:

- `mm_bf16(backend="cutile")`
- `bmm_bf16(backend="cutile")`
- `gemm_fp8_nt_groupwise(backend="cutile")` — K-major scale only

Performance:

All measurements: CUDA event timing, warmup=20, repeats=50, median latency. Cosine similarity = 1.000 on all shapes.

### `mm_bf16` — H100 NVL (sm_90)

| Shape (M×N×K) | cuDNN (µs) | tinygemm (µs) | **cuTile (µs)** | vs cuDNN | vs tinygemm |
|---|---|---|---|---|---|
| 64×2048×2048 | 99.71 | 66.30 | **58.88** | +1.69× | +1.13× |
| 128×4096×4096 | 98.94 | 120.35 | **58.82** | +1.68× | +2.04× |

> Geomean: **1.69× vs cuDNN**, **1.52× vs tinygemm**  
> tgv not supported on sm_90 (skipped). Larger shapes (K=11008, M≥512) hit memory limits on this node.

### `gemm_fp8_nt_groupwise` K-major — B300 SXM6 AC (sm_103)

| Shape (M×N×K) | cutlass (µs / TFLOPS) | **cuTile (µs / TFLOPS)** | cuTile / cutlass |
|---|---|---|---|
| 256×1024×1024 | 60.64 / 8.85 | **64.29 / 8.35** | 1.06× (≈ tied) |
| 1024×4096×4096 | 68.10 / 504.6 | **71.33 / 481.7** | 1.05× (≈ tied) |
| 2048×4096×4096 | 69.06 / 995.1 | **88.54 / 776.1** | 1.28× slower |
| 4096×4096×4096 | 78.50 / 1750.9 | **173.06 / 794.2** | 2.20× slower |

> cuTile is within ~5% of cutlass for M ≤ 1024. Performance gap widens at larger M — autotune config investigation ongoing.

## Build system note: PEP 517 isolation

`build_backend.py` installs the `cuda-tile` compile chain (nvcc, tileiras, nvvm…)
with `--no-deps` to avoid pulling in `nvidia-cuda-runtime-cu13`, which conflicts
with torch's `nvidia-cuda-runtime-cu12` (both expose `libcudart.so` at different
major versions, breaking cudnn at import time).

When `pip install` runs **without `--no-build-isolation`** (the default), pip
creates a clean PEP 517 build environment that contains only the declared build
dependencies such as `setuptools` and `packaging` — it does **not** include `pip`
itself or `uv`. As a result, we cannot invoke `pip install` from within that
environment to resolve this conflict. The function therefore falls back to a
warn-and-continue approach: the compile chain is pre-installed on CI images, and
a clean PyPI install will surface a clear `ImportError` at first cuTile use rather
than aborting the build.


flashinfer/gemm/kernels/cutile/bmm_bf16_cutile.py
bkryu, line 1: Can we rename this file `bmm_bf16.py` or `bmm_bf16_cutile.py`?

flashinfer/gemm/kernels/gemm_fp8_nt_groupwise_cutile.py
bkryu, line 1: Can we rename this file `gemm_fp8_nt_groupwise.py` or `gemm_fp8_nt_groupwise_cutile.py`?

flashinfer/gemm/kernels/cutile/mm_bf16_cutile.py
bkryu, line 1: Can we rename this file `mm_bf16.py` or `mm_bf16_cutile.py`?

bkryu (dismissed):
@yifeis-nv thanks for contributing these files. The PR generally looks good to me except for some file names and locations.

I'm guessing you saw [flashinfer/cute_dsl/](https://github.com/flashinfer-ai/flashinfer/tree/main/flashinfer/cute_dsl) that has some cute-dsl kernels and thus created `flashinfer/cutile/`. However, we are trying to move away from adding kernels there.

We should place these cuTile GEMM kernels in [flashinfer/gemm/kernels](https://github.com/flashinfer-ai/flashinfer/tree/main/flashinfer/gemm/kernels). You can either move the `cutile` directory into there or just move the kernel files into `flashinfer/gemm/kernels`.

For the individual file names I added separate comments.

flashinfer/gemm/kernels/cutile/mm_bf16_cutile.py
yifeis-nv: Done

flashinfer/gemm/kernels/gemm_fp8_nt_groupwise_cutile.py
yifeis-nv: Done

flashinfer/gemm/kernels/cutile/bmm_bf16_cutile.py
yifeis-nv: Done

dhiraj113:
@yifeis-nv Is the PR description AI generated or human written?

flashinfer/gemm/gemm_base.py
```diff
@@ -359,6 +359,34 @@ def _tgv_gemm_requirement(
...
+    out: Optional[torch.Tensor] = None,
+    out_dtype: torch.dtype = torch.bfloat16,
+    bias: Optional[torch.Tensor] = None,
+    pdl: bool = False,
+    backend: Literal[
+        "cudnn", "cutlass", "tgv", "cublaslt", "tinygemm", "cutile", "auto"
+    ] = "cudnn",
+):
+    if out_dtype not in (torch.bfloat16, torch.float16, torch.float32):
+        raise ValueError(
```
dhiraj113, line 378: Why do you want to raise an error here?

yifeis-nv:
> @yifeis-nv Is the PR description AI generated or human written?

The description is generated by AI, I will rewrite it.

flashinfer/gemm/gemm_base.py
yifeis-nv: Those functions&dtype will be supported in the following PRs. I think if merge them all in one pr will make it hard to review. This PR wants to import a stable cutile backend, so that it will be easier for more and more cutile kernels for flashinfer.

flashinfer/gemm/kernels/cutile/bmm_bf16_cutile.py
dhiraj113: Instead, can you move these files into a new directory under `kernels` called `cutile`. The original file name would be fine.

tests/gemm/test_bmm_bf16.py
```diff
@@ -4,6 +4,7 @@
 
 from flashinfer import autotune, bmm_bf16
 from flashinfer.gemm.gemm_base import CUDNN_AVAILABLE
+from flashinfer.gemm.kernels.cutile_common import is_cuda_tile_available
```
dhiraj113: Is `is_cuda_tile_available` an API now?

dhiraj113: If so, its location should be different than `flashinfer.gemm.kernels.`

tests/gemm/test_bmm_bf16.py
```diff
@@ -50,5 +57,32 @@ def test_bmm_bf16(b, m, n, k, res_dtype, backend):
...
+    if not bmm_bf16.is_backend_supported("cutile", cc_num):
+        pytest.skip("cuTile backend not supported on current compute capability.")
+    if not is_cuda_tile_available():
+        pytest.skip("cuda-tile / tileiras compiler not available in this environment.")
+
+    torch.random.manual_seed(0)
+    A = torch.randn(4, 128, 256, device="cuda", dtype=torch.bfloat16)
+    B = torch.randn(4, 256, 256, device="cuda", dtype=torch.bfloat16).transpose(-2, -1)
+
+    from flashinfer.gemm.kernels.bmm_bf16_cutile import _BMM_BF16_TUNE_CACHE
```
dhiraj113: How is the user suppose to know that this is an API?

build_backend.py
```diff
@@ -607,6 +607,88 @@ def _install_nvep_runtime_wheels(built_nixl: bool, built_nccl: bool) -> None:
         ) from e
 
 
+def _install_cuda_tile_compile_deps() -> None:
```
dhiraj113, line 610: Can you explain how the contents of this file are used?

yifeis-nv: There is a libcudart version conflict. The straightforward way to get the tileiras compiler is to install `cuda-tile[tileiras]>=1.4.0`. However, that extra also pulls in `nvidia-cuda-runtime==13.3.* `(`libcudart.so.13`), which clashes with the `nvidia-cuda-runtime-cu12` (`libcudart.so.12`) bundled with **PyTorch**. When importing, cudnn sees both versions on the library path and raises a `“Multiple libcudart libraries found”` error, which breaks all cudnn tests.

To avoid this, we remove the `[tileiras]` extra from requirements.txt and instead install only the compile-time wheels (nvcc, tileiras, nvvm, nvjitlink, crt) using `--no-deps`. These components don’t require libcudart at compile time—they only generate PTX or cubin—and the resulting binaries run against PyTorch’s cu12 at test time. So skipping the transitive cu13 runtime dependency is safe.

yifeis-nv: BTW, there is the PEP 517 isolated build environment. When pip install runs without `--no-build-isolation`, pip creates a clean build environment that includes only the declared build dependencies (such as setuptools and packaging), and does not include `pip` itself or `uv`. As a result, we cannot invoke pip install from within that environment to solve the conflict of `nvidia-cuda-runtime`.

tests/gemm/test_bmm_bf16.py
yifeis-nv: Good catch — importing a private variable is not a stable API. Replaced with make_bmm_bf16_tune_cache(), a public factory function defined at the top of bmm_bf16_cutile.py that returns a fresh, isolated cache dict.

tests/gemm/test_bmm_bf16.py
yifeis-nv: You are right, thx. Moved the export to flashinfer.gemm directly (same as is_cute_dsl_available in flashinfer.cute_dsl).

build_backend.py
dhiraj113: Thanks for the explanation. Could you please add this informatikon to your MR too.

yifeis-nv: I've updated both the PR description and the docstring in build_backend.py to include this explanation.

dhiraj113 (approved):

PR #3428, opened 2026-05-28
fix: scope trtllm-gen block_size guard to its own backend

### LINKED ISSUE #3425: [Bug]Regression: `backend="cute-dsl"` MLA decode via API rejects `page_size=128`
## Summary

  `tests/attention/test_cute_dsl_mla_decode.py::test_cute_dsl_mla_decode_via_api  now fails for `page_size=128` with:

  ```
  ValueError: Supported block_size are 32 and 64, got 128
```

  This appears to be a regression introduced between:

  - good/base: 56d537a106024eb25f4d4a186eadc226990a9185
  - bad/head: c802a05d9806a3d8ec72014f9cec4d67da0d5e8e

  The likely culprit is:

  - 0b755de7060980b6ced02359aac6921aef5418d2
  - PR #3355: MLA Decode Autotuning Across TRTLLM-Gen and CuTe Backends

  ## Failing Tests

  Examples:

  FAILED tests/attention/test_cute_dsl_mla_decode.py::test_cute_dsl_mla_decode_via_api[monolithic-128-512-4]
  FAILED tests/attention/test_cute_dsl_mla_decode.py::test_cute_dsl_mla_decode_via_api[monolithic-128-512-128]
  FAILED tests/attention/test_cute_dsl_mla_decode.py::test_cute_dsl_mla_decode_via_api[monolithic-64-128-1]
  FAILED tests/attention/test_cute_dsl_mla_decode.py::test_cute_dsl_mla_decode_via_api[monolithic-64-512-128]

  ## Root Cause Hypothesis (By  Codex)

  PR #3355 refactored trtllm_batch_decode_with_kv_cache_mla to share dispatch and autotuning logic between trtllm-gen and cute-dsl.

  As part of that refactor, the following validation is now executed before backend-specific runner selection:
```
  block_size = kv_cache.size(-2)
  if block_size != 32 and block_size != 64:
      raise ValueError(f"Supported block_size are 32 and 64, got {block_size}")
```
  This check appears to be trtllm-gen specific. However, it now also runs for explicit backend="cute-dsl" calls.

  The failing test uses:

  backend="cute-dsl"
  page_size=128

  Before #3355, the explicit cute-dsl path directly called cute_dsl_mla_decode, so page_size=128 was not rejected by the trtllm-gen block-size guard.

  ## Expected Behavior

  For backend="cute-dsl", page_size=128 should be validated by the cute-dsl compatibility checks, not rejected by the trtllm-gen-only block_size in {32, 64} restriction.

  ## Actual Behavior

  The public MLA API rejects the call before reaching the cute-dsl backend:

  ValueError: Supported block_size are 32 and 64, got 128

  ## Suggested Fix  (By  Codex)

  Make the block-size restriction backend-specific:

  - Keep block_size in (32, 64) as a hard error for backend="trtllm-gen".
  - Do not apply that restriction to explicit backend="cute-dsl".
  - For backend="auto", exclude trtllm-gen from candidate runners when block_size is unsupported, but still allow cute-dsl if its own compatibility check passes.

  Pseudo-structure:

  trtllm_gen_reason = None
  if block_size not in (32, 64):
      trtllm_gen_reason = (
          f"trtllm-gen supports block_size 32 and 64, got {block_size}"
      )

  if backend == "trtllm-gen" and trtllm_gen_reason is not None:
      raise ValueError(trtllm_gen_reason)

  runner_names = []
  if backend in ("auto", "trtllm-gen") and trtllm_gen_reason is None:
      runner_names.append("trtllm-gen")
  if backend in ("auto", "cute-dsl") and cute_dsl_reason is None:
      runner_names.append("cute-dsl")

  ## Additional Test Coverage

  The existing backend="cute-dsl" test already catches the explicit-backend regression.

  It may also be useful to add coverage for:

  backend="auto"
  page_size=128

  Expected behavior: auto should skip trtllm-gen and dispatch to cute-dsl if cute-dsl supports the configuration.

cc: @aleozlx @Vinnie6167

### PR DESCRIPTION
**Pull request by elwhyjay**
## 📌 Description

After #3355 unified the trtllm-gen and cute-dsl dispatch in `trtllm_batch_decode_with_kv_cache_mla`, the trtllm-gen `block_size in (32, 64)` check runs before backend selection and rejects valid cute-dsl configurations such as `page_size=128`.

This PR scopes the guard to trtllm-gen by converting it into a `trtllm_gen_reason` variable, following the existing `cute_dsl_reason` pattern. The dispatch then consults both reasons:

- `backend="cute-dsl"`, `page_size=128`: now runs (was rejected since #3355).
- `backend="trtllm-gen"`, `page_size=128`: raises `trtllm-gen requires block_size in (32, 64), got 128`.
- `backend="auto"`, `page_size=128`: dispatches to cute-dsl when its compatibility check passes; raises if neither backend is compatible.
- `backend="auto"`, `page_size in (32, 64)`: unchanged.

`_check_trtllm_gen_mla_shape` is still called unconditionally for shape normalization; its `block_num % (128 / block_size) != 0` constraint is trivially satisfied for `block_size=128`.

## 🔍 Related Issues

Fixes #3425.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

Added two regression tests in `tests/attention/test_cute_dsl_mla_decode.py`:

- `test_mla_decode_auto_dispatches_to_cute_dsl_for_block_size_128` - `backend="auto"` with `page_size=128` runs end-to-end via cute-dsl.
- `test_mla_decode_trtllm_gen_rejects_block_size_128` - explicit `backend="trtllm-gen"` raises with the backend-specific message.

The pre-existing `test_cute_dsl_mla_decode_via_api[*-128-*]` covers the explicit cute-dsl regression case once dispatch is corrected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic backend selection for MLA decoding to only choose compatible implementations, record incompatibility reasons, and raise clearer errors when configurations aren't supported (e.g., unsupported block sizes)
* **Tests**
  * Added regression tests validating MLA decoding with larger block sizes and confirming correct backend routing and error behavior

### CODE REVIEW

File: tests/attention/test_cute_dsl_mla_decode.py [hunk 1]
```diff
@@ -1465,3 +1465,58 @@ def soft_capping_score_mod(score, batch_idx, qo_idx, kv_idx, head_idx):
...
+    assert torch.isfinite(out).all()
+
+
+def test_mla_decode_trtllm_gen_rejects_block_size_128():
+    skip_if_unsupported()
+    from flashinfer.mla import trtllm_batch_decode_with_kv_cache_mla
+
+    args = _mla_decode_block_size_128_inputs()
+    with pytest.raises(ValueError, match=r"trtllm-gen requires block_size"):
+        trtllm_batch_decode_with_kv_cache_mla(**args, backend="trtllm-gen")
```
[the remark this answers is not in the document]
elwhyjay: Both tests call `skip_if_unsupported()` on the first line, which already gates on `is_sm100a_supported(device) or is_sm110a_supported(device)` (and on `is_cute_dsl_available()`). The existing guard matches the SM range the suggestion targets, so I think no extra check is needed here. Happy to swap it for an explicit `is_sm100a_supported(...)` call if a tighter scope is preferred.

flashinfer/mla/_core.py
```diff
@@ -1810,8 +1810,11 @@ def trtllm_batch_decode_with_kv_cache_mla(
     sm_count = get_device_sm_count(query.device)
 
     block_size = kv_cache.size(-2)
+    trtllm_gen_reason: Optional[str] = None
```
aleozlx: [nit] shall we make it more clear smth like "trtllmgen_not_supported_reason"

concern about future misuse of this var

aleozlx:
pls address merge conflicts & thx for the contrib

i expect a few others to review too

flashinfer/mla/_core.py
elwhyjay: Done. Renamed to `trtllm_gen_not_supported_reason`

elwhyjay:
HI @aleozlx! Thank you for quick review.Rebased onto main and resolved the conflict. The conflict came from main, which removed the `block_size in (32, 64)` guard outright rather than adding block_size=128 support to trtllm-gen. As a result, current main still appends trtllm-gen as an auto candidate for unsupported block sizes. I re-applied the guard as a backend-specific `trtllm_gen_not_supported_reason`, so `backend="trtllm-gen"` raises a clear error and `backend="auto"` excludes trtllm-gen while still dispatching to cute-dsl.

Vinnie6167:
The changes here look good to me 👍

qsang-nv (approved):

elwhyjay:
@Vinnie6167 @qsang-nv @aleozlx Thanks for the review, Could you run the CI?

aleozlx:
auto merge enabled

Issue #3436, opened 2026-05-28
Autotuning is often silently a no-op for cuBLAS bmm_fp8 and cuDNN bmm_fp8/mm_fp4 in common configurations


bkryu:
## Summary

When autotuning is enabled for the `cuBLAS` and `cuDNN` backends of `bmm_fp8` (and `cuDNN` for `mm_fp4`), the autotuner appears to run but, in several common configurations, silently selects the default/heuristic tactic instead of a tuned one. There is no error — the call succeeds and produces correct results — but the expected tuning benefit never materializes.

This shows up in three related ways:

1. **Benchmark gate:** `flashinfer_benchmark.py --routine bmm_fp8` won't autotune
   `cudnn`/`cublas` at all — they are not in the routine's autotune allowlist.
2. **cuBLAS `bmm_fp8`:** the tuned tactic is not applied at runtime for any **M that doesn't exactly equal a profiled bucket** (i.e. essentially all serving traffic once you tune over a range).
3. **cuDNN `bmm_fp8` / `mm_fp4`:** on cuDNN versions without the override-shape feature, the runner exposes only a single heuristic plan and ignores the selected tactic, so autotuning has nothing to choose and nothing to apply.

## Affected APIs / backends

- `flashinfer.bmm_fp8(..., backend="cublas")`
- `flashinfer.bmm_fp8(..., backend="cudnn")`
- `flashinfer.mm_fp4(..., backend="cudnn")`
- `benchmarks/routines/gemm.py` (`testBmmFp8` autotune allowlist)

### (1) Benchmark refuses to autotune cudnn/cublas

```bash
python3 benchmarks/flashinfer_benchmark.py --routine bmm_fp8 \
  --batch_size 1 --m 4 --n 4096 --k 4096 \
  --input_dtype fp8_e4m3 --mat2_dtype fp8_e4m3 --out_dtype bfloat16 \
  --backends cudnn cublas --autotune
```
Output:
```
[INFO] cudnn backend does not support autotune
[INFO] cublas backend does not support autotune
[ERROR] No backends passed validation. Exiting.
```

### (2) cuBLAS: tuned tactic not applied for non-bucket-aligned M

Tune over a range (e.g. at M=2048, which profiles buckets up to 2048), then run
inference at an M that rounds to a *different* bucket (e.g. M=100):

```python
import torch
from flashinfer import autotune, bmm_fp8

def mk(m):
[...]

PR #3437, opened 2026-05-28
Fix silent no-op autotuning for cuBLAS `bmm_fp8` and cuDNN `bmm_fp8`/`mm_fp4`

bkryu:
### Summary 
Autotuning the `cuBLAS`/`cuDNN` backends of `bmm_fp8` (and `cuDNN` for `mm_fp4`) silently selected the default heuristic tactic instead of a tuned one in several common configurations as reported in #3436. This PR makes the tuned tactic actually get enumerated and applied, and enables these backends in the benchmark/tests.

### Changes
- **cuBLAS `bmm_fp8` cache key** (`CublasFp8GemmRunner`): `get_cache_key_extras` included `a.shape` (dynamic `M`), violating the autotuner's synthesis-invariant contract — the runtime key missed the bucketed profile key and fell back to `tactic=-1`. Now returns dtypes only; added `_algo_cache_key` (full shapes) for the internal cuBLASLt algo cache.
- **cuDNN `bmm_fp8` / `mm_fp4` non-override path** (`_cudnn_gemm_fp8_runner`, `CudnnFp4GemmRunner`): built with `HEURISTICS_CHOICE` (one plan) and hardcoded `tactic=-1`. Now uses `build_plan_policy.ALL` and passes the selected tactic through. 
  - Override path requires cuDNN 9.21 and autotuning works as expected. No changes for override path.
  - Non-override path now mirrors the override path with this PR.
- **Benchmark/tests**: add `cudnn`/`cublas` to `testBmmFp8`'s autotune allowlist and relax the `test_bmm_fp8` autotune guard.

## 🔍 Related Issues

- #3436

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed FP8 GEMM caching to avoid collisions and ensure correct tuned parameters are applied during execution.
  * Improved runtime fallback when tuned tactics are out-of-range to use a safe heuristic.

* **Performance**
  * Autotuning now supports the cuBLAS backend in addition to existing backends.
  * Autotuner includes workspace sizing in tuning keys and evaluates a fuller set of plans for more reliable results.

* **Tests**
  * Added tests to validate autotuner cache reuse across shape bucket boundaries.

### DIFF
--- a/flashinfer/gemm/gemm_base.py
+++ b/flashinfer/gemm/gemm_base.py
@@ -138,12 +138,22 @@ def get_gemm_module():
                 self._algo_cache: dict = {}
 
             def get_cache_key_extras(self, inputs: List[torch.Tensor]) -> tuple:
+                # Must be synthesis-invariant: including a.shape (dynamic M)
+                # makes the runtime key miss the bucketed profile key, so the
+                # tuned tactic is dropped for tactic=-1. Shapes are already in
+                # the autotuner's input_shapes; add only dtypes.
+                a, b, _, _, out, _ = inputs
+                return (a.dtype, b.dtype, out.dtype)
+
+            def _algo_cache_key(self, inputs: List[torch.Tensor]) -> tuple:
+                # Internal cuBLASLt algo-enumeration cache: this one is
+                # shape-specific, so key on full shapes (incl. M).
                 a, b, _, _, out, _ = inputs
                 return (a.shape, b.shape, a.dtype, b.dtype, out.dtype)
 
             def _get_algos(self, inputs):
                 a, b, scale_a, scale_b, out, workspace_buffer = inputs
-                key = self.get_cache_key_extras(inputs)
+                key = self._algo_cache_key(inputs)
                 cached = self._algo_cache.get(key)
                 if cached is not None:
                     return cached
@@ -3193,6 +3203,9 @@ def _cudnn_gemm_fp8_runner():
             if self._use_override_shape:
                 graph = self._get_override_graph(a, b, out)
             else:
+                # ALL exposes every heuristic plan to the autotuner;
+                # HEURISTICS_CHOICE would collapse to a single plan. Graph
+                # is @lru_cache'd, so all plans are built once per shape.
                 graph = build_cudnn_gemm_fp8_graph(
                     a_shape=a.shape,
                     a_stride=a.stride(),
@@ -3202,7 +3215,7 @@ def _cudnn_gemm_fp8_runner():
                     b_type=_torch_data_type_to_cudnn_data_type(b.dtype),
                     o_type=_torch_data_type_to_cudnn_data_type(out.dtype),
                     device=a.device,
-                    policy=cudnn.build_plan_policy.HEURISTICS_CHOICE,
+                    policy=cudnn.build_plan_policy.ALL,
                 )
 
             return list(range(graph.get_execution_plan_count()))
@@ -3228,6 +3241,8 @@ def _cudnn_gemm_fp8_runner():
                     tactic=max(tactic, 0),
                 )
             else:
+                # Apply the tuned tactic. tactic>=0 -> specific plan
+                # (policy=ALL); tactic==-1 -> cheap HEURISTICS_CHOICE default.
                 _cudnn_gemm_fp8(
                     workspace_buffer,
                     a,
@@ -3236,7 +3251,7 @@ def _cudnn_gemm_fp8_runner():
                     scale_b,
                     out,
                     out.dtype,
-                    tactic=-1,
+                    tactic=tactic,
                 )
             return out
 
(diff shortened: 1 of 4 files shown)


yanqinz2:
lgtm

nv-yunzheq (approved):

flashinfer/gemm/gemm_base.py
```diff
@@ -3236,7 +3256,7 @@ def forward(
                     scale_b,
                     out,
                     out.dtype,
-                    tactic=-1,
+                    tactic=tactic,
```
qiching, line 3268: Since M is a static dim of the non override graph, the `policy=ALL` plan list (count + order) isn't guaranteed identical for bucket-M vs real-M. So `tactic` may select a suboptimal plan (silent perf regression will defeats the PR's purpose) or be out of range → `execute_plan_at_index` raises more likely with `round_up=True`. numerics are safe; same hazard in cuDNN FP4 and cuBLAS (where it becomes a `ValueError`). Cutlass is immune since its tactics are shape independent.

Could we:

- Add a test that tunes at one bucket then runs at a **non-bucket** M (real-M ≠ bucket-M), asserting no range error and that the tuned plan is applied.
- Clamp `tactic` to the runtime plan/algo count and fall back to `-1` on out-of-range, instead of raising.

bkryu: I see what you mean. To be precise about scope: this is the non-override path only. The override path builds at the bucketed `cache_m` and feeds the real shape via `override_shapes`, so its plan list is identical at tune and runtime; the non-override path rebuilds at the real `M`, where the cuDNN plan list / cuBLASLt algo list can differ in length and order vs the bucket.

I addressed both requests:
1. Clamp + fallback (no more raising): `execute_cudnn_gemm_fp8_graph` and `execute_cudnn_gemm_fp4_graph` now clamp tactic to `-1` when it's `>=  graph.get_execution_plan_count()` (placed before the workspace-size query, which would also have indexed out of range), falling back to `graph.execute()`.  `CublasFp8GemmRunner.forward` now falls back to the default heuristic instead of ValueError on an out-of-range/empty algo list. I fall back to -1 (the engine's  heuristic best) rather than clamping to count-1, since an arbitrary in-range plan has no tuning basis and -1 matches the autotuner's own fallback. CUTLASS is  untouched (shape-independent tactics).
2. Test: as suggested added `test_autotuner_gemm_cross_bucket_m` (parametrized over cudnn/cublas) — tunes at M=256, runs at M=100 (real-M ≠ bucket-M=128), and asserts no range  error + finite output + the tuned bucket entry is reused. The clamp logic itself confirmed by injecting an  out-of-range index (15 plans, tactic=65 --> falls back, no raise); and a forced-non-override cross-M sweep runs cleanly for both FP8 and FP4.

Will retrigger the CI

qiching (approved):
LGTM!

dhiraj113 (approved):

Saddss:
Independent validation of #3436 / #3437 on **SM120** (RTX 5090, driver 580.159).

### Integration (vLLM 0.23.0 serving)
- Framework: vLLM **0.23.0**, `-O3`, `quantization=modelopt`, `kv-cache-dtype=fp8`, MoE cutlass backend
- Code path: `FlashInferFP8ScaledMMLinearKernel` → `bmm_fp8(..., backend="auto")`
- Workload: sustained online inference, QPS 5.7, 12 × 30s rounds (3 runs per config)

| FlashInfer | `No tuned config` + `tactic=-1` per run | Notes |
|---|---|---|
| **0.6.12** | **57 / 57 / 57** | reproducible across 3 runs |
| **0.6.8.post1** (control) | **0 / 0 / 0** | same vLLM + serve flags |
| **0.6.13rc1** | **0** | micro + serving startup confirm fix |

This matches the root cause described here (`CublasFp8GemmRunner.get_cache_key_extras` including dynamic `M`).

### Minimal repro (pure FlashInfer)
Tune at `M=2048`, then infer at non-bucket `M=100` (`backend=cublas`) and `M=96` (`backend=auto`) with `autotune(False)`:

```python
import torch
from flashinfer import autotune, bmm_fp8

def mk(m, n=4096, k=4096):
    a = (torch.randn(1, m, k, device="cuda", dtype=torch.bfloat16) * 0.1).to(torch.float8_e4m3fn)
    b = (torch.randn(1, n, k, device="cuda", dtype=torch.bfloat16) * 0.1).transpose(-2, -1).contiguous().to(torch.float8_e4m3fn)
    return a, b, torch.tensor(1.0, device="cuda"), torch.tensor(1.0, device="cuda")

a, b, sa, sb = mk(2048)
with autotune(True):
    bmm_fp8(a, b, sa, sb, torch.bfloat16, backend="cublas")

a, b, sa, sb = mk(100)
with autotune(False):
    bmm_fp8(a, b, sa, sb, torch.bfloat16, backend="cublas")  # 0.6.12: WARNING tactic=-1

a, b, sa, sb = mk(96)
with autotune(True):
    bmm_fp8(a, b, sa, sb, torch.bfloat16, backend="auto")
with autotune(False):
    bmm_fp8(a, b, sa, sb, torch.bfloat16, backend="auto")    # 0.6.12: WARNING tactic=-1
```

On **0.6.12** stderr shows `No tuned config covers fp8_gemm ... tactic=-1` for both shapes. On **0.6.13rc1** there are no such warnings.

We are pinning **0.6.8.post1** until **0.6.13 stable** — could you share the expected release timeline?

PR #3455, opened 2026-05-29
fix(gemm): cuDNN mxfp8 override-shape block-scale descale shape (broken on cuDNN >=9.21)

YangXu1990uiuc:
Fixes `bmm_mxfp8(backend="cudnn")`, which is **broken for every shape on cuDNN ≥ 9.21** (`CUDNN_STATUS_NOT_SUPPORTED_INVALID_DYNAMIC_SHAPE`), and hardens the scale-layout contract. Two commits:

### 1. Correct the override-shape block-scale descale shape
`execute_cudnn_gemm_mxfp8_graph_override_shape` passed the runtime scale buffer's **flat 1D** shape (`list(a_descale.shape)`) as the override for a tensor the graph declares **3D** (`[batch, dim_m, dim_k]`, `F8_128x4` reordered) → cuDNN rejects the mismatch. Masked because existing tests only use 128-aligned M, and on cuDNN < 9.21 the non-override path runs instead. Present since override-shape was introduced (#2790): FP4 used `_expand_block_scale_tensor_shape`, mxfp8 was handed the flat `.shape`.

**Fix:** describe the descale override with the same 3D `[batch, dim_m, dim_k]` shape/stride the graph was built with, recomputed for the actual M via `_calculate_block_scale_dims` (mirrors FP4). Pure shape/stride metadata — no data movement, no quantizer change, cuDNN-path only.

Root of the asymmetry: **`nvfp4_quantize` returns a 2D scale, `mxfp8_quantize` returns a 1D-flat scale** — so `_expand` (which assumes 2D/3D) handled FP4 but not mxfp8.

### 2. Reject non-swizzled scales on the cuDNN mxfp8 path
The cuDNN graph hardwires `F8_128x4` (128-padded) scales — there is no linear/non-swizzled path; a linear scale only matched at 128-aligned M by coincidence and silently produced wrong results otherwise. Since `mxfp8_quantize` returns 1D for *both* layouts (rank can't disambiguate), validate the scale **length** against `_mxfp8_swizzled_scale_len` and raise a clear error on mismatch. Also skips the `(cudnn, non-swizzled)` combo in `test_bmm_mxfp8` (it tested an unsupported layout). Downstream check: vllm/sglang use linear scales but never via `bmm_mxfp8`, and the cutlass non-swizzled path is untouched — so this is safe.

### Tests / validation (Blackwell SM100 + cuDNN 9.30)
- New `test_bmm_mxfp8_cudnn_dynamic_m`: non-128-aligned M (130/200/257/384/1000), swizzled — 5/5.
- Full aligned-M `test_bmm_mxfp8[cudnn]` matrix goes green (was 288/288 failing on cuDNN 9.30).
- Non-aligned linear → clear `ValueError`; swizzled → correct.

### Known separate follow-ups (not this PR)
- **Batched b>1 at non-128-aligned M** still NaNs: `mxfp8_quantize` flattens `[b,m,k]→[b*m,k]` and pads the combined dim, so the swizzled scale isn't per-batch-padded the way the cuDNN bmm graph expects. Needs per-batch SF padding (quantizer-side).
- flashinfer CI should add a **cuDNN ≥9.21** leg — this whole path was untested there.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed block-scale tensor dimension handling for MXFP8 cuDNN graph execution to respect actual runtime shapes.
  * Added input validation to reject non-swizzled MXFP8 scale tensors for the cuDNN backend with guidance to use the alternate backend.

* **Tests**
  * Tightened cuDNN test coverage and added a regression test for MXFP8 matrix multiply with variable M sizes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/flashinfer-ai/flashinfer/pull/3455?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/flashinfer/gemm/gemm_base.py
+++ b/flashinfer/gemm/gemm_base.py
@@ -8427,6 +8427,28 @@ def bmm_mxfp8(
     if resolved_backend == "cudnn":
         if not CUDNN_AVAILABLE:
             raise ValueError("cudnn is not available")
+        # The cuDNN mxfp8 graph requires the F8_128x4-swizzled scale layout.
+        # Non-swizzled / linear scales are not supported -- the graph's
+        # reordering + 128-padding only matched a linear scale at 128-aligned M
+        # by coincidence, and silently produced wrong results otherwise.
+        # mxfp8_quantize returns a 1D buffer for *both* layouts, so we can't
+        # distinguish by rank; the swizzled (128x4) buffer has a specific length
+        # (M/N padded to 128, K grouped by 4), which differs from the linear
+        # length at non-128-aligned M.  Reject on length mismatch.
+        exp_a = _mxfp8_swizzled_scale_len(
+            A.shape[0] * A.shape[1], A.shape[2], SfLayout.layout_128x4
+        )
+        exp_b = _mxfp8_swizzled_scale_len(
+            B.shape[0] * B.shape[2], B.shape[1], SfLayout.layout_128x4
+        )
+        if A_scale.numel() != exp_a or B_scale.numel() != exp_b:
+            raise ValueError(
+                "bmm_mxfp8(backend='cudnn') requires swizzled (F8_128x4) scale "
+                "factors (mxfp8_quantize(..., is_sf_swizzled_layout=True)); got "
+                f"A_scale.numel()={A_scale.numel()} (expected {exp_a}), "
+                f"B_scale.numel()={B_scale.numel()} (expected {exp_b}). "
+                "Use the cutlass backend for non-swizzled scales."
+            )
         mxfp8_gemm_sm100(
             A,
             B,
--- a/flashinfer/gemm/gemm_base.py
+++ b/flashinfer/gemm/gemm_base.py
@@ -2800,18 +2800,35 @@ def execute_cudnn_gemm_mxfp8_graph_override_shape(
         UIDs.BLOCK_DESCALE_B_UID.value,
         UIDs.O_UID.value,
     ]
+
+    # The block-scale tensors are declared in the graph as 3D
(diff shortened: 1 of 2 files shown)


YangXu1990uiuc:
Closing as superseded. Recap for the record:

- **Commit 1** (override-shape block-scale descale fix) already landed on `main` via **#3489 "add_cudnn_mxfp8"** (@yanqinz2) — `execute_cudnn_gemm_mxfp8_graph_override_shape` recomputes the 3D descale shape via `_calculate_block_scale_dims`, with a comment citing this PR. GPU-verified on SM100/B200 (cuDNN 9.24): the dynamic-M cases (M ∈ {130,200,257,384,1000}) all pass on main.

- **Commit 2** (reject non-swizzled scales via a numel check) is **not worth landing**. Measured on main: a linear scale on the cuDNN bmm path is never rejected by cuDNN — it's silently degraded at 128-aligned M (cos ≈ 0.92) and an out-of-bounds scale read (NaN) at non-aligned M. A numel-based guard is length-consistent with the quantizer (no false-reject) but structurally incomplete: linear and swizzled 1D scales have identical numel whenever `(b·m) % 128 == 0`, so the check passes the bad scale through in exactly the silently-degraded case and only fires where the output is already NaN. Linear vs swizzled scales are indistinguishable at the API boundary. #3882 handles this the right way (document swizzled-required + drop the pretend-linear test paths + the 0.99 cos threshold). Full analysis in #3882.

- The one surviving increment — the **dynamic-M regression test** `test_bmm_mxfp8_cudnn_dynamic_m` — is being contributed separately (folded into #3882 or a small standalone PR), since it guards the #3489 fix and main has no non-128-aligned-M coverage.

PR #3357, opened 2026-05-18
feat: Add 8x4 swizzle layout support to MXFP4 and MXFP8 CuTe-DSL kernels

bkryu:
## 📌 Description
Brings the CuTe-DSL backend to parity with CUDA: `mxfp8_quantize` and the MXFP4 path of `fp4_quantize` previously raised ValueError when asked for 8x4. NVFP4 `cute-dsl` already supported 8x4.

### Changes ###
  
Kernels (`flashinfer/quantization/kernels/mxfp{4,8}_quantize.py`) -- Same pattern previously used for `NVFP4QuantizeSwizzledKernel`.
- Added `SF_LAYOUT_8x4` constant and threaded sf_layout through `MXFP{4,8}QuantizeSwizzledKernel`.
- Added a `_compute_sf_offset` JIT helper that dispatches between `compute_sf_index_swizzled_{128x4,8x4}_gpu`.
- Added `sf_layout` to the `@functools.cache` key on the compile helpers. 
  
Notes
- No public-API breakage (the new mxfp8 kwarg is keyword-only with False default).
- High-level mxfp4_quantize still hardcodes 128x4 on both backends — parity preserved; 8x4 reachable via fp4_quantize.

In all cases tested, the CUDA and CuTe-DSL backends have 100% bitwise matches in quantized output and scale factor tensors. Perf numbers from B200:
<details>
  <summary>Backend performance comparisons from bench_nvfp4_quantize_backend_comparison.py</summary>

nvfp4_quantize_backend_comparison_linear_bfloat16
nvfp4_quantize_backend_comparison_swizzled_128x4_bfloat16
nvfp4_quantize_backend_comparison_swizzled_8x4_bfloat16

</details>
<details>
  <summary>Backend performance comparisons from bench_mxfp4_quantize_backend_comparison.py</summary>

mxfp4_quantize_backend_comparison_linear_bfloat16
mxfp4_quantize_backend_comparison_swizzled_128x4_bfloat16
mxfp4_quantize_backend_comparison_swizzled_8x4_bfloat16

</details>
<details>
  <summary>Backend performance comparisons from bench_mxfp8_quantize_backend_comparison.py</summary>
mxfp8_backend_comparison_linear_bfloat16
mxfp8_backend_comparison_swizzled_128x4_bfloat16
mxfp8_backend_comparison_swizzled_8x4_bfloat16

</details>

## 🔍 Related Issues

#3356

## 🚀 Pull Request Checklist

[...]

PR #3457, opened 2026-05-29
feat(quantize): opt-in rank_preserving scale-factor convention (mxfp8 + nvfp4 batched)

YangXu1990uiuc:
## The mess today

The block-scale quantizers return scale factors (SF) with **inconsistent rank conventions**, for historical reasons (they were added in separate feature PRs):

| quantizer | SF shape today |
|---|---|
| `nvfp4_quantize` (2D matmul) | **2D** `[M_pad, sf_cols]`; a 3D input is folded into M and collapsed to 2D |
| `nvfp4_batched_quantize` (bmm) | per-batch correct, but SF **flattened** to `[B, M_pad*sf_cols]` |
| `mxfp8_quantize` | **1D flat**, regardless of input rank; a 3D input folds the batch into `M` and pads the combined `B*M` |

For a batched `[B, M, K]` input, folding the batch into `M` and padding the *combined* dim is fine at 128-aligned `M`, but at non-aligned `M` the **per-batch padding rows don't exist** in the flat buffer — so batched consumers (notably the cuDNN `bmm` path) can't be made correct downstream. The 1D-vs-2D-vs-flattened divergence is also a recurring footgun in the GEMM↔cuDNN integration.

We can't just change the default shapes: vLLM and SGLang consume the flat `mxfp8_quantize` buffer and reshape it themselves, so flipping the default would break them.

## What this PR adds

An **opt-in** `rank_preserving: bool = False` on the two quantizers a batched GEMM would call, with **one shared convention**: rank mirrors the input, padding is per batch.

**`mxfp8_quantize`:**
- `False` (default, unchanged): flat 1D buffer, exactly as today.
- `True`: 2D `[M,K]` → **2D `[M_pad, K_blocks_pad]`** (a reshape); 3D `[B,M,K]` → **3D `[B, M_pad, K_blocks_pad]`**, padded per batch (swizzled 3D is quantized per batch since the rows can't be recovered from the `B*M`-flattened buffer; linear 3D is reshaped).

**`nvfp4_batched_quantize`:**
- `False` (default, unchanged): per-batch SF flattened as `[B, M_pad*sf_cols]`.
- `True`: per-batch SF as **3D `[B, M_pad, sf_cols]`** — a pure reshape of the already-per-batch buffer (the native batched kernel already pads each batch independently).

So `mxfp8_quantize(rank_preserving=True)` and `nvfp4_batched_quantize(rank_preserving=True)` produce the **same `[B, M_pad, *]` layout**, which is exactly what the cuDNN `bmm` descale tensor expects — unblocking batched mxfp8 `bmm` at non-128-aligned `M`, and giving a future `bmm_fp4` the same convention out of the box.

Notes:
- `rank_preserving` was deliberately **not** added to `nvfp4_quantize` (the 2D matmul quantizer): fp4 already has the dedicated `nvfp4_batched_quantize` for batched use, so adding a per-batch loop there would be a redundant third path. A docstring note now points batched fp4 callers at `nvfp4_batched_quantize`.
- `mxfp8_quantize` + `backend="cute-dsl"` + `rank_preserving=True` raises `NotImplementedError` (CUDA backend only).

## How to transition

1. Today: nothing changes — default is `False` everywhere.
2. Batched consumers (incl. the upcoming bmm paths) pass `rank_preserving=True` and get the unified `[B, M_pad, *]` layout; non-batched callers can pass it on 2D input to get the nvfp4-style 2D scale.
3. Once vLLM/SGLang and other consumers migrate, flip the defaults to `True` and deprecate the flat forms — converging mxfp8 and nvfp4 onto one rank-preserving scale convention.

## Testing (Blackwell SM100)

- **mxfp8:** 2D output is a byte-exact reshape of the 1D buffer; 3D swizzled equals per-batch independent quantization and carries the larger per-batch element count (the actual fix); 3D linear is a plain reshape; cute-dsl opt-in is rejected. Default path unchanged — 418 existing quantize tests pass.
- **nvfp4 batched:** rank_preserving output is a byte-exact reshape of the flat per-batch SF and each batch equals an independent `fp4_quantize`.

Separate from #3455 (cuDNN mxfp8 override-shape fix); this is the quantizer-side groundwork.

## Summary by CodeRabbit

* **New Features**
  * Added opt-in rank-preserving mode for FP8 and FP4 quantization to preserve per-batch/2D scale-factor shapes for batched inputs.

* **Documentation**
  * Clarified docstrings on how batched inputs fold to 2D and pointed to batched APIs for per-batch behavior.

* **Tests**
  * Added tests validating rank-preserving shapes and equivalence to non-ranked outputs across 2D/3D, swizzled/linear layouts, and backend behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/flashinfer-ai/flashinfer/pull/3457?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/flashinfer/quantization/fp4_quantization.py
+++ b/flashinfer/quantization/fp4_quantization.py
@@ -1176,6 +1176,12 @@ def nvfp4_quantize(
           When `per_token_activation=True`, returns a third tensor containing
           per-token FP32 scales.
 
+    Note:
+        This is the 2D-matrix quantizer; a batched ``[B, M, K]`` input has its batch
+        folded into M (the scale factor is collapsed to 2D ``[-1, sf_cols]``). For batched
+        (bmm) use, prefer :func:`nvfp4_batched_quantize`, which pads each batch's M
+        independently and supports ``rank_preserving=True`` for a per-batch 3D scale.
+
     Warning:
         The "cute-dsl" backend is **experimental** and not part of the stable API.
         It may change or be removed in future versions without notice.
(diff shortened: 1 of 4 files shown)


flashinfer/quantization/fp8_quantization.py
```diff
@@ -207,6 +241,40 @@ def mxfp8_quantize(
             SfLayout.layout_128x4 if is_sf_swizzled_layout else SfLayout.layout_linear
         )
 
+    if rank_preserving and backend != "cuda":
+        raise NotImplementedError(
+            "rank_preserving=True is currently only supported with backend='cuda'."
+        )
```
bkryu, line 247: As you can see in the PR description of #3357 (expand to see plots), `mxfp8_quantize` is meaningfully faster with `backend='cute-dsl'` so it would be great to add support.

Do we need kernel-level changes to support `rank_preserving=True` with `backend='cuda'`?

flashinfer/quantization/fp4_quantization.py
bkryu, line 1: Question: What about `mxfp4_quantize()` Can we give similar treatments as nvfp4 and mxfp8 quantize?

bkryu (commented):
Thanks @YangXu1990uiuc, left a few minor comments.

I wrote the `backend='cute-dsl'` kernels and they are substantially more performant than the cuda kernels so I'm trying to understand what has to be done to bring support parity.

PR #2996, opened 2026-04-06
fix: tinygemm2 hang issue due to barrier sync

jimmyzho:
## 📌 Description

The `tinygemm2` kernel hung intermittently on H100 CI runners. Running `compute-sanitizer --tool synccheck` confirmed a barrier synchronization failure.

## Root Cause

The kernel launches 384 threads (12 warps of 32):

| Warps | Threads   | Role                                      |
|-------|-----------|-------------------------------------------|
| 0-3   | 0-127     | Compute (HMMA + reduction)                |
| 4-7   | 128-255   | DMA -- weight TMA (1 elected per warp)    |
| 8-11  | 256-383   | DMA -- activation TMA (1 elected per warp)|

The original control flow in `csrc/tinygemm2.cu` was:

```cpp
// BEFORE (buggy)
if (warp_id >= 4 && elect_one_sync()) {
    // DMA work -- only 8 elected threads enter
}
else if (warp_id < 4) {
    // Compute work -- 128 threads enter
    reduction_buffer[threadIdx.x] = accum4;
    __syncthreads();          // <-- BUG: only warps 0-3 reach this
    if (warp_id == 0) { /* reduction + output */ }
}
```

**248 out of 384 threads** (all of warps 4-11, elected or not) never reached
`__syncthreads()`. Per the CUDA spec, `__syncthreads()` requires **all threads
in the CTA** to participate. Violating this is undefined behavior -- on some
H100 configurations it silently works, on others the block hangs forever.

### File: `csrc/tinygemm2.cu` (lines 198-391)

Two structural changes, no logic changes:

**1. Separate the warp-id check from the election** so non-elected DMA threads
   fall through to `__syncthreads()`:

```cpp
// AFTER (fixed)
if (warp_id >= 4) {
    // Outer block: ALL threads in warps 4-11 enter here.
    // Non-elected threads skip the inner block and fall through.
    if (elect_one_sync()) {
        // DMA work -- only 8 elected threads
    }
}
else {                          // warp_id < 4
    // Compute work -- 128 threads
    reduction_buffer[threadIdx.x] = accum4;
}
```

**2. Move `__syncthreads()` and the warp-0 reduction outside the if/else**
   so all 384 threads participate:

```cpp
// AFTER (fixed) -- after the if/else block closes
__syncthreads();                // ALL 384 threads reach this

if (warp_id == 0) {
    // reduction + output (unchanged)
}
```

## 🚀 Pull Request Checklist

[...]

PR #2914, opened 2026-03-30
feat: Add cuBLASLt backend for `mm_bf16` and enable multi-tactic autotuning for FP8/MXFP8 runners

vadiklyutiy:
## Summary

- **Add cuBLASLt backend for `mm_bf16`**: new `backend="cublaslt"` option (gated to SM100/SM103). Autotuning across all available cuBLASLt algorithms via `get_valid_tactics()`. 
- **Enable multi-tactic autotuning for single-tactic GEMM runners**: `CublasFp8GemmRunner`, `CudnnFp8GemmRunner`, and `CudnnMxfp8GemmRunner` previously hardcoded a single tactic (`return [0]` / `return [-1]`), preventing the autotuner from exploring better algorithms (the same are done for FP4 and FP16). Now all three enumerate available algorithms/plans via `get_valid_tactics()` and pass the selected tactic through to execution.
- **Improve test coverage**: added `cublaslt` and `auto` backends + `auto_tuning` parameter to `test_mm_bf16.py`, `auto` backend to `test_bmm_bf16.py`, and a dedicated edge-case test for zero-algorithm handling.
## Test Results

All GEMM/BMM test suites pass with 0 failures:

| Test File | Passed | Failed |
|-----------|--------|--------|
| `test_mm_bf16.py` | 1441 | 0 |
| `test_mm_fp8.py` | 30 | 0 |
| `test_mm_fp4.py` | 1440 | 0 |
| `test_mm_mxfp8.py` | 1843 | 0 |
| `test_bmm_bf16.py` | 144 | 0 |
| `test_bmm_fp8.py` | 1188 | 0 |
| `test_bmm_mxfp8.py` | 288 | 0 |
| **Total** | **6374** | **0** |

Test additions:
- `test_mm_bf16.py`: added `cublaslt` and `auto` to backend parametrize, added `auto_tuning` parameter, added `test_cublaslt_bf16_runner_zero_algos` edge-case test.
- `test_bmm_bf16.py`: added `auto` backend.

## Summary by CodeRabbit

* **New Features**
  * Added a cuBLASLt BF16 GEMM backend with selectable algorithm tactics; module generated for supported hardware.
  * Exposed APIs to enumerate and run FP8/BF16 GEMM algorithms for explicit tactic selection.
  * Expanded backend choices with "cublaslt" and broader "auto" routing.

* **Bug Fixes**
  * Robustified autotuner hashing to avoid failures with unhashable attributes.

* **Tests**
[...]

PR #3459, opened 2026-05-29
perf(gemm): multi-tactic STAGES autotuning for TinyGEMM BF16 backend

askliar:
## Summary

The TinyGEMM BF16 backend was wired into the `mm_bf16` autotuner in #3203, but its runner exposed only a **single tactic** (`get_valid_tactics -> [0]`), so the autotuner could never explore the kernel's pipeline-depth (`STAGES`) knob. This PR turns `STAGES` into a tunable tactic, mirroring the cuBLASLt change in #2914 that upgraded single-tactic GEMM runners (`CublasFp8GemmRunner`, cuDNN FP8/MXFP8) to enumerate real algorithms via `get_valid_tactics()` and pass the selected tactic through to execution.

Previously the kernel auto-selected the largest pipeline depth that fits in shared memory (16 on SM90/SM100, 4 on SM120). Now the autotuner profiles all fitting multiples of 4:
- **SM90 / SM100**: `{4, 8, 12, 16}`
- **SM120**: `{4, 8}`

## Changes

- **`csrc/tinygemm2.cu`**: hoist the fixed tile config to namespace scope (single source of truth), add `compute_valid_stages()`, and rework `launch_tinygemm2<HAS_BIAS>` to take an explicit `stages` argument that dispatches over the four `launch_tinygemm2_impl<STAGES,...>` instantiations. `stages <= 0` preserves the historical auto-selection. New FFI exports: `tinygemm2_get_valid_stages` (fills a CPU int32 buffer, returns count), `tinygemm2_op_with_stages`, `tinygemm2_nobias_op_with_stages`. The original `tinygemm2_op` / `tinygemm2_nobias_op` are unchanged (auto path).
- **`flashinfer/gemm/routergemm.py`**: expose the two new ops as registered custom ops, plus a cached `get_valid_stages(device_index)` helper, through the module `SimpleNamespace`.
- **`flashinfer/gemm/gemm_base.py`**: `TinyGemmBf16GemmRunner.get_valid_tactics` returns the device's valid `STAGES`; `forward` dispatches to the `*_with_stages` ops for `tactic >= 0` and the auto path for the `-1` fallback.
- **`tests/gemm/test_mm_bf16.py`**: add `test_tinygemm_bf16_runner_multi_tactic` (verifies tactic enumeration and correctness of every tactic + the fallback). The existing parametrized test already covers `tinygemm`/`auto` with `auto_tuning=True` end-to-end.

## Notes / things for reviewers to check

- `STAGES = 8` and `STAGES = 12` are **newly instantiated** depths — the kernel previously only ever ran 4 or 16. They satisfy the documented "multiple of 4" constraint (one slot per compute warp) and the stage-cycling logic is generic, but they have not been exercised in production before (cf. the prior barrier-sync hang fix #2996). The new runner-level test is intended to surface any correctness/hang issue at these depths.
- The auto (non-tuned, `tactic == -1`) path is unchanged on all archs.

## Testing

> ⚠️ Implemented in a CPU-only environment — **GPU tests have not been run yet.** Python compiles and pre-commit (ruff + clang-format) passes. Please run on SM90/SM100/SM120:

```bash
pytest tests/gemm/test_mm_bf16.py -k "tinygemm or auto" -v
pytest tests/gemm/test_mm_bf16.py::test_tinygemm_bf16_runner_multi_tactic -v
```


.github/workflows/issue-claim.yml
```diff
@@ -8,8 +8,6 @@ name: Issue Claim
 on:
   issue_comment:
     types: [created]
-  schedule:
```
dhiraj113, line 11: Was this intended?

dhiraj113:
@askliar Is this PR no longer needed?

askliar:
@dhiraj113 sorry for slow response but yes, let's put it on hold for now.

Issue #2409, opened 2026-01-23
perf: `trtllm_ragged_attention` requires zeroed output and workspace for non-causal attention


mgoin:
## Summary

When using `trtllm_ragged_attention` (via `trtllm_ragged_attention_deepseek` in vLLM) for non-causal attention (`is_causal=False`) with variable sequence lengths, the output tensor and workspace buffer must be pre-zeroed by the caller for correct results. This adds significant overhead on the critical path during chunked prefill.

## Expected Behavior

The kernel should handle uninitialized output and workspace memory internally, producing correct results without requiring the caller to zero-fill tensors beforehand.

## Context

In vLLM's MLA attention implementation, we use `trtllm_ragged_attention_deepseek` for chunked context prefill (non-causal attention over cached KV). We discovered that accuracy dropped significantly without explicit zeroing:

```python
# vllm/model_executor/layers/attention/mla_attention.py

def _run_prefill_context_chunk_trtllm_ragged(
    self, prefill: MLACommonPrefillMetadata, chunk_idx: int, q, k, v
):
    """TRT-LLM ragged attention for context chunks (non-causal)."""
    from flashinfer.prefill import trtllm_ragged_attention_deepseek

    # ISSUE: These zero-fills are required for correctness but add overhead on the hot path
    out = torch.zeros(
        q.shape[0],
        q.shape[1],
        v.shape[2],
        device=q.device,
        dtype=q.dtype,
    )
    prefill.workspace_buffer.fill_(0)

    attn_out, lse = trtllm_ragged_attention_deepseek(
        query=q,
        key=k,
        value=v,
        workspace_buffer=prefill.workspace_buffer,
        seq_lens=prefill.chunked_context.seq_lens[chunk_idx],
        # ... other params ...
        is_causal=False,  # Non-causal for context chunks
        return_lse=True,
        out=out,
    )
```

NOTE: The causal case (`_run_prefill_new_tokens_trtllm_ragged`) does NOT require this zeroing and works correctly without it.

pavanimajety:
@PerkzZheng for viz

PerkzZheng:
@mgoin this kernel doesn't require zero initialisation of the output tensor. Can you prepare a standalone test so that we can reproduce this issue easily ?

saltyminty:
@mgoin checking again if you're still running into this issue, and if so, do you have a standalone repro?

PR #3469, opened 2026-05-31
chore: explicitly set mMultiCtasKvMode in ragged attention launcher

namgyu-youn:
## 📌 Description

Explicitly set `mMultiCtasKvMode = false` in `trtllm_ragged_attention_launcher` for clarity, matching the paged context path. Confirmed this is a no-op: `TllmGenSelectKernelParams` already forces `Disabled` for `Context`-type kernels regardless of this field, and the struct is zero-initialized by default anyway.

## 🚀 Pull Request Checklist

- [x] `pre-commit install`
- [x] `pre-commit run --all-files`

## 🧪 Tests

- [x] `pytest tests/attention/` with `is_causal=False` ragged attention cases

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ragged attention processing during context handling.
  * Enhanced inference reliability by ensuring attention execution uses the appropriate key-value processing configuration.
  * No interface changes are required; existing integrations continue to work as before.

### DIFF
--- a/csrc/trtllm_fmha_kernel_launcher.cu
+++ b/csrc/trtllm_fmha_kernel_launcher.cu
@@ -724,6 +724,7 @@ void trtllm_ragged_attention_launcher(
 
   runner_params.mKernelType = FmhaKernelType::Context;
   runner_params.mTileScheduler = TileScheduler::Persistent;
+  runner_params.mMultiCtasKvMode = false;
   runner_params.mMaskType =
       is_causal ? TrtllmGenAttentionMaskType::Causal : TrtllmGenAttentionMaskType::Dense;
 


namgyu-youn:
cc @yyihuang. Could you please look onto this? It fixes an uninitialized `mMultiCtasKvMode` in the ragged attention launcher that can silently produce wrong results with `is_causal=False`.

csrc/trtllm_fmha_kernel_launcher.cu
```diff
@@ -725,6 +725,7 @@ void trtllm_ragged_attention_launcher(
 
   runner_params.mKernelType = FmhaKernelType::Context;
   runner_params.mTileScheduler = TileScheduler::Persistent;
+  runner_params.mMultiCtasKvMode = false;
```
saltyminty, line 728: Not sure if I'm missing something – isn't this param already default false? We manually memset to 0 on `fmhaRunnerParams.h::381`.

This is still a good change/better explicit structure, but unsure why this closes https://github.com/flashinfer-ai/flashinfer/issues/2409

saltyminty (approved):
Approved but please see comment

csrc/trtllm_fmha_kernel_launcher.cu
namgyu-youn: Oh thanks for catching that, it's already default false.

Also, after looking again, I don't think this PR is actually related to #2409 — see PerkzZheng's comment (https://github.com/flashinfer-ai/flashinfer/issues/2409#issuecomment-3797672120): the kernel shouldn't need output zero-init at all, and no repro was ever confirmed. So this isn't a fix for that issue.

Updated the PR description to just call this a clarity change. Could you take another look?

saltyminty (approved):

PR #3471, opened 2026-06-01
Add optional MagiAttention FFA wrapper

DwenGu:
## Summary

  This PR adds an optional FlashInfer-style wrapper for MagiAttention single-GPU forward Flex Flash Attention.

  The new API is exposed under:

  from flashinfer.ffa_kernels import FlexFlashAttentionWrapper

  It follows FlashInfer’s existing plan() / run() style while keeping MagiAttention as an optional runtime dependency. The wrapper accepts
  MagiAttention’s range-based mask representation directly through q_ranges, k_ranges, and attn_type_map.
 
## Request For Maintainer Feedback

  The main goal of this PR is to validate whether this integration style is acceptable for FlashInfer:

  - Is flashinfer.ffa_kernels.FlexFlashAttentionWrapper an appropriate namespace?
  - Is keeping MagiAttention as a lazy optional dependency acceptable?
  - Is the plan() / run() wrapper shape consistent with FlashInfer’s API conventions?
  - Should this wrapper stay outside the top-level flashinfer.__init__ export for now?
  - Does it support the open‑source license of MagiAttention?

  If this API direction is acceptable, I can follow up with a separate PR for benchmark integration into FlashInfer’s unified benchmark

 

  ## What This PR Includes

  This PR adds:

  - flashinfer/ffa_kernels/__init__.py
  - flashinfer/ffa_kernels/flex_flash_attn.py
  - tests/ffa/test_flex_flash_attn_wrapper.py

  The wrapper supports:

  - plan(q_ranges, k_ranges, attn_type_map, ...)
  - run(q, k, v, out=None, lse=None, return_lse=False)
  - run_return_lse(q, k, v, ...)

  Optional MagiAttention arguments supported through plan() include:

  - sink
  - sink_layout
  - softmax_scale
  - softcap
  - deterministic
  - sm_margin
  - auto_range_merge
  - sparse_load
  - ref_block_size
  - max_seqlen_q

  The wrapper is intended for single-GPU forward/prefill-style attention with MagiAttention FFA range masks.

 ## What This PR Does Not Include

  This PR intentionally does not change existing FlashInfer attention kernels.

  It does not include:

  - top-level flashinfer.__init__ export
  - FlashInfer unified benchmark integration
  - paged KV cache decode support
  - backward support
  - distributed MagiAttention support
  - changes to existing prefill/decode/sparse kernels

  MagiAttention is imported lazily only when run() is called, so importing FlashInfer does not require MagiAttention to be installed.

  ## Testing

  Targeted tests were run locally:

  python -m pytest tests/ffa/test_flex_flash_attn_wrapper.py -q
  19 passed, 1 skipped

  Related regression checks:

  python -m pytest tests/ffa -q
  19 passed, 1 skipped

  python -m pytest tests/test_artifacts.py tests/utils/test_decorators.py tests/ffa -q
  35 passed, 1 skipped

  python -m compileall -q flashinfer/ffa_kernels tests/ffa
  PASS

  pre-commit run --files flashinfer/ffa_kernels/__init__.py flashinfer/ffa_kernels/flex_flash_attn.py tests/ffa/
  test_flex_flash_attn_wrapper.py
  PASS

  The tests cover:

  - lazy optional dependency behavior
  - plan() metadata validation
  - run() behavior
  - run_return_lse() behavior
  - user-provided out / lse buffers
  - missing MagiAttention dependency error path
  - correctness against direct MagiAttention when MagiAttention + CUDA are available
  - full, causal, inverse causal, bidirectional causal masks
  - GQA
  - multi-range and complex multi-segment masks

 ## Local Performance Data

  These numbers were collected locally on an NVIDIA H20-3e GPU with bf16. They are not part of this PR and are only provided as context for
  the integration.

 
image

  The wrapper overhead is effectively negligible compared with direct MagiAttention. The performance profile is therefore dominated by
  MagiAttention itself, not the FlashInfer wrapper layer.

## Summary by CodeRabbit

* **New Features**
  * Added an experimental MagiAttention-based Flex Flash Attention adapter with support for `NHD`/`HND` layouts and optional `lse` output.
  * Added API documentation and reference docs for the new adapter.

* **Bug Fixes**
  * Improved input validation and error messages for invalid layouts and range settings.
  * Ensured traced attention dispatch behaves correctly across supported layouts.

* **Tests**
  * Added unit tests (including traced-dispatch behavior) and optional real-path GPU correctness tests.
  * Added a dedicated CI workflow to run MagiAttention Flex Flash Attention tests in the CI container.


qsang-nv:
**Integration shape diverges from how FlashInfer absorbs other attention backends, and the wrapper is a no-op without a separate third-party install**

Two related concerns about how this is wired in.

**1. It re-exports MagiAttention's native API instead of adapting it into FlashInfer's.**

`FlexFlashAttentionWrapper` surfaces MagiAttention's own kwargs verbatim under a new `flashinfer.ffa_kernels` namespace — `plan()` just caches them (`q_ranges`, `k_ranges`, `attn_type_map`, `sink_layout`, `auto_range_merge`, `ref_block_size`, `sparse_load`, `max_seqlen_q`, ...) and `run()` splats them straight into `flex_flash_attn_func(**kwargs)`. There's no normalization onto FlashInfer's conventions, and it isn't plugged into any existing abstraction.

That's a different direction from how FlashInfer integrates other (incl. external) attention implementations today:

- **`backend=` dispatch** — the main wrappers (`BatchPrefillWithPagedKVCacheWrapper`, `BatchDecodeWithPagedKVCacheWrapper`) take `backend="auto"` and branch internally across `fa2 / fa3 / trtllm-gen / cute-dsl / xqa`, all behind one normalized `plan()/run()` API, with `determine_attention_backend(...)` auto-selecting.
- **Adapter registry** — `flashinfer/parallel_attention/attention_ops.py` integrates the external `flash_attn_interface` (FA3) via `@AttentionOpManager.register_attn("flash-attn3")`. The `FlashAttn3` adapter takes FlashInfer-normalized args, normalizes layout/dtype in (`convert_qkv_layout`) and output out, and **hides the external library's native API** behind FlashInfer's interface.

The common discipline in both: the external impl is *adapted into* FlashInfer's interface; callers program against FlashInfer, not the external lib. This PR instead exposes MagiAttention's interface as-is, so a user of `FlexFlashAttentionWrapper` is effectively coding against MagiAttention with a `flashinfer.` import prefix. Practical fallout: it can't be auto-selected, can't reuse the workspace/CUDA-graph plumbing, and FlashInfer ends up owning a public API surface (`q_ranges`/`attn_type_map`/`sink_layout`/...) it doesn't control. (This is also why `plan()` here is effectively a no-op — it isn't wired into that plumbing.)

**2. As a standalone passthrough to an uncontrolled external lib, it's a no-op out of the box and its core path isn't covered by CI.**

`magi_attention` is an external package (SandAI's [MagiAttention](https://github.com/SandAI-org/MagiAttention), Apache-2.0). It isn't vendored, isn't in `requirements.txt`/`pyproject.toml`, and isn't a transitive dep — so FlashInfer install `run()` always raises the install-hint `ImportError`. That's acceptable for an *optional* feature, but the real kernel path then isn't exercised by CI: the correctness tests gate on `pytest.importorskip("magi_attention")` (`_skip_without_magi_attention_cuda`), so on CI (no MagiAttention installed) every wrapper-vs-MagiAttention and vs-torch case is **skipped** — only the monkeypatched pure-Python unit tests run. The "19 passed" was on a machine with MagiAttention installed; on CI the meaningful coverage drops to near zero, so regressions in the actual integration would go unnoticed.

**Suggested direction**

Follow the `parallel_attention` adapter precedent: keep MagiAttention's native call internal and expose a FlashInfer-normalized surface, registered as a selectable backend rather than a parallel namespace, e.g.:

```python
@AttentionOpManager.register_attn("magi-ffa")
class MagiFFA:
    def __call__(self, query, key, value, *, q_ranges, k_ranges,
                 attn_type_map=None, return_lse=False, tensor_layout="NHD", **kw):
        ffa = _load_flex_flash_attn_func()  # lazy import stays
        q, k, v = _normalize_layout(query, key, value, tensor_layout)
        out, meta = ffa(q=q, k=k, v=v, q_ranges=q_ranges, k_ranges=k_ranges,
                        attn_type_map=attn_type_map, **kw)
        return _normalize_out(out, meta, tensor_layout, return_lse)
```

Plus: (a) decide how CI can cover the real path — e.g. an optional/nightly job that `pip install`s MagiAttention and runs `tests/ffa`, otherwise document that this path is locally-verified-only; (b) make the optional/experimental status explicit in the docstring so users know an extra install is required; (c) reconsider the `ffa_kernels` name — `gdn_kernels`/`kda_kernels` hold actual in-tree CuTe-DSL kernels, whereas this module ships no kernel, only a passthrough, so the name is misleading.

A note on arch gating: please don't hardcode an SM90-only check (in the wrapper or the test skip). MagiAttention FFA is multi-arch — the native FFA kernel targets Hopper, and the `FFA_FA4` backend extends to Blackwell (SM100) and Ampere (SM80) as of v1.1.1 with multi-arch CUDA builds — so a hardcoded `is_sm90a_supported` gate would wrongly exclude supported GPUs and will drift from upstream. Let MagiAttention own arch support.

If FFA's range-mask model genuinely doesn't fit FlashInfer's existing `indptr` + causal/custom-mask abstractions and a dedicated entry point is warranted, that's defensible — but the adapter discipline (normalize layout/output, hide the native API) should still apply.

qsang-nv:
Thanks for the substantial revision — the earlier architectural feedback is addressed: the `plan()/run()` shell is gone, the namespace is now `flashinfer.magi_ffa`, layout/output/LSE are normalized, the native `(out, meta)` return is hidden, the dedicated-entry-point rationale is documented, and there's correctly no hardcoded SM90 gate. The guarded optional-dependency pattern is precedented by `parallel_attention`; this implementation is even lazier in that it defers the import until first use. Remaining items are all should-fixes:

**1. Pin and validate the supported optional-dependency environment.**
The guarded import is legitimate and precedented; the remaining issue is reproducibility. FlashInfer and MagiAttention declare conflicting Cutlass DSL constraints (MagiAttention pins `nvidia-cutlass-dsl==4.3.5`/`==4.4.2`; FlashInfer requires `>=4.5.0`), so this can't be a resolver-clean extra and must be documented as a *validated override* rather than a supported resolution. Please:
- Pin the tested MagiAttention version (author verified `1.1.0.post10` on H20) and document the exact install order:
  ```bash
  pip install magi_attention==<validated-version>
  pip install "nvidia-cutlass-dsl>=4.5.0"   # restore after MagiAttention downgrades it
  ```
- Reproduce that override in the opt-in CI job so it doesn't silently drift as either side bumps versions.
- Describe **Hopper** as the FlashInfer-validated configuration. Blackwell/Ampere (FFA_FA4) may be noted as *upstream-supported* but are **not validated** under this DSL override, so FlashInfer should not present them as validated configurations. (Accurate arch picture: FlashInfer-validated = Hopper/H20; MagiAttention-upstream-claimed = Blackwell/Ampere; no SM90 runtime gate needed.)

**2. Opt-in CI doesn't use the repo's CI image.** The workflow's `python -m venv --system-site-packages` reuses the runner *host*, not the Docker image the comment references, so PyTorch/CUDA build deps aren't guaranteed. Existing GPU jobs enter `flashinfer-ci` via `bash ci/bash.sh "${DOCKER_IMAGE}" ...` (see `pr-test.yml`) — run inside that image, which is also the natural home for reproducing the pinned install above (a `pinned + latest` matrix would additionally catch upstream drift).

**3. Bitwise-equality asserts on a nondeterministic kernel.** `_assert_matches_direct_magi_attention` compares two separate kernel launches (native vs adapter) with `rtol=0, atol=0`. With atomic reductions — especially the `auto_range_merge` overlapping-`q_ranges` case (`[[0,32],[0,32]]`) — run-to-run results aren't guaranteed bit-identical. Enable deterministic mode for the exact-equality cases, or use a small tolerance for the overlap/auto-merge case. (This revives the earlier CodeRabbit point — still valid in the new code.)

**4. Public `@flashinfer_api` API missing `trace=`.** `flex_flash_attn` is decorated with `@flashinfer_api` but carries no `TraceTemplate`, which the CLAUDE.md "Trace Template Checklist" asks for on public APIs decorated with it. Please add a `trace=` template per the checklist.

---

Non-blocking: the PR description still describes the removed `plan()/run()` API end-to-end (`ffa_kernels.FlexFlashAttentionWrapper`, old `tests/ffa/test_flex_flash_attn_wrapper.py` path) and should be rewritten for the current `flashinfer.magi_ffa.flex_flash_attn`; and add `persist-credentials: false` to the checkout step.

qsang-nv:
Thanks — the previous round's four items are all addressed in code: the MagiAttention pin (`1.1.0.post10`) with a CI install-time version assertion and a first-use warning for non-validated versions, the opt-in job moved into the repo CI image via `ci/bash.sh`, tolerance-based comparisons plus an explicit `deterministic=True` bit-exact case, and a full NHD `trace=` template (reference/init/check) plus an HND schema-only template, example calls, and committed JSON. The docs page with the install order and the Hopper-vs-upstream support matrix is a good addition. Two actionable issues came in with the trace/CI work, plus one lower-priority pre-existing Trace Apply limitation.

**1. The opt-in workflow cannot validate the PR that introduces it.**
`magi-ffa-optin-test.yml` is new in this PR and its only trigger is `workflow_dispatch`. GitHub only delivers `workflow_dispatch` for workflows that already exist on the default branch — the file is not on `main` (404), and no such workflow is registered in Actions, so there is no run. That means the pinned, resolver-inconsistent Cutlass-DSL override this integration depends on has **not** been validated by CI. Two places already state that this coverage exists, while the API docstring also implies an available opt-in CI path:
- `docs/api/magi_ffa.rst`: "Hopper (SM90) — FlashInfer-validated (**opt-in CI** + verified on H20)"
- `tests/ffa/test_flex_flash_attn.py`: "In CI, the real path **is covered by** the opt-in workflow"
- `_flex_flash_attn.py` docstring: "...or in an opt-in CI job"

**Suggested fix — fold the job into `pr-test.yml`, gated on the existing `gate` authorization plus a path check.** Unlike `workflow_dispatch`, `pull_request`-triggered workflows use the workflow version from the PR merge ref, so this takes effect within this PR and needs no separate landing PR. Reusing the existing gate matters: team members are authorized by membership, while external PRs are authorized by the `run-ci` label — and `@flashinfer-bot run` only adds that label (`ci-bot-commands.yml`), while `run-magi-ffa` doesn't exist in the repo's labels. So a job keyed on a new label would never fire via the documented bot command, and for an external author the existing gate would reject it anyway. For an external PR carrying `run-ci`, the gate short-circuits on the label before the secret/team-membership path, so it works even though fork PR secrets are unavailable. The path check keeps the Hopper job off unrelated PRs; `setup` already computes the changed-file list, so it can emit one more output:

```yaml
  # setup.outputs:  magi_ffa_changed: ${{ steps.check.outputs.magi_ffa }}
  # inside the existing "Check Skip Conditions" step, reusing $CHANGED:
  #   if echo "$CHANGED" | grep -qE '^(flashinfer/magi_ffa/|tests/ffa/|tests/trace/test_flex_flash_attn_reference_correctness\.py|scripts/task_test_magi_ffa\.sh)'; then
  #     echo "magi_ffa=true" >> $GITHUB_OUTPUT
  #   else
  #     echo "magi_ffa=false" >> $GITHUB_OUTPUT
  #   fi

  magi-ffa-real-path:
    name: MagiAttention FFA real-path tests (Hopper)
    needs: [gate, setup]
    if: |
      needs.gate.outputs.authorized == 'true' &&
      needs.setup.outputs.magi_ffa_changed == 'true'
    runs-on: [self-hosted, linux, x64, gpu, h100]
    # ... checkout, then:
    #   bash ci/bash.sh "${DOCKER_IMAGE}" ./scripts/task_test_magi_ffa.sh
```

(`scripts/task_test_magi_ffa.sh` can stay as-is. Whether to also match `flashinfer/trace/templates/attention.py` is a judgement call — including it catches Magi-only trace edits but makes unrelated attention-trace changes pull in an expensive H100 job.)

A standalone workflow is fine too, as long as it keeps an equivalent permission gate and a real trigger path — please don't run fork code on the self-hosted GPU runner ungated, and don't use `pull_request_target` to check out and execute the PR head, because that would run untrusted fork code in the privileged base-repository context. Either way, please trigger it once on this PR and link the green run; until such a run exists, the three statements above should say "verified locally on H20" rather than CI-covered.

**2. The trace definition silently drops behavior-bearing MagiAttention kwargs.**
`flex_flash_attn` publicly accepts and forwards `**kwargs` (`softmax_scale`, `softcap`, `sink`, `deterministic`, `auto_range_merge`, ...), but none are represented in the template's `inputs`. Auto-dump does `bound = _sig.bind(...); fi_trace_fn(**dict(bound.arguments))`, so with a trailing `**kwargs` in the signature those extras land under a nested `"kwargs"` entry that the template never reads. Consequences beyond imprecise benchmark metadata:
- `softmax_scale` / `softcap` / `sink` change numerics, but the emitted definition contains no corresponding inputs, so replay has no way to reproduce the original invocation; it can also be deduplicated against a default-path definition that describes different behavior;
- `deterministic` / `auto_range_merge` change which kernel path is benchmarked;
- replay and a matched Trace Apply solution are invoked **without** these arguments, which can silently produce different attention results.

Two workable directions:
- **Model them explicitly** — promote the supported numerics- and kernel-selecting options to named parameters and corresponding template inputs: for example, `softmax_scale` / `softcap` / `deterministic` / `auto_range_merge` as scalars, and `sink` as an optional tensor with the appropriate axes. `sink_layout` must also be represented explicitly or non-default values rejected. Any tuning option left in `**kwargs` (`sm_margin`, `sparse_load`, `ref_block_size`, `max_seqlen_q`, ...) must cause trace/apply fallback when supplied; otherwise replay and benchmark metadata still describe a different kernel invocation.
- **Or make tracing default-path-only** — keep `**kwargs`, but have the trace dispatch decline to emit a definition whenever unsupported kwargs are present, and add a runtime Trace Apply guard that calls the original function for such invocations. Rejecting only in the trace dispatch is not sufficient, because Trace Apply consumes the registered templates and does not invoke that dispatcher when routing a live call.

Please add a regression case with a non-default `softmax_scale` or `sink` to lock the chosen behavior down.

**3. (Lower priority) The published HND trace definition isn't selectable by Trace Apply.**
`enable_apply()` gates on `template0 = templates[0]` via `_template_matches_any`, and `_make_wrapper` uses that same `template0` for `definition_name()` and `adapt_and_call()`. Since NHD is registered first, an HND-only solution never wraps the API. If an NHD solution is present, HND calls are still processed through the NHD template: `extract_axes` fills each axis from the *first* map that resolves it, so the NHD extractor interprets `[H, N, D]` as `[N, H, D]`, and the same NHD template is then used for naming and adaptation.

To be fair, this is a **pre-existing Trace Apply limitation**, not something this PR introduces — `trtllm_batch_decode_trace_dispatch` (2 templates) and `trtllm_batch_decode_mla_trace_dispatch` (3 templates) have the same constraint. I would not ask this PR to implement dispatch-aware Trace Apply. What I'd ask for here: please explicitly document that the HND definition is for trace schema/dump only and is **not supported safely by Trace Apply** — the existing "stays schema-only" comment describes the missing reference/init/check, not the Apply limitation. This documentation would only scope the advertised support; it would not make Trace Apply safe. Please track the actual runtime dispatch/fallback fix in a separate Trace Apply issue.

---

Non-blocking: the PR description still describes the removed `plan()/run()` API end-to-end (`ffa_kernels.FlexFlashAttentionWrapper`, old `tests/ffa/test_flex_flash_attn_wrapper.py` path) and should be rewritten for the current `flashinfer.magi_ffa.flex_flash_attn`.

PR #3474, opened 2026-06-01
fix: zero TRTLLM-GEN counter workspace region before each decode launch

### LINKED ISSUE #2408: [Feature Request] Remove the requirement for preinitializing workspace and output tensors to zeros for trtllm attention kernels
Hi,

We currently notice that the accuracy of MHA Non Causal attention drops when the output buffer and workspace is not preinitialized to zeros - like here

https://github.com/vllm-project/vllm/blob/fec9da0af48da5fe9523914aeddf2c34f53166a9/vllm/model_executor/layers/attention/mla_attention.py#L1586

Is it possible to remove this requirement in the means of the kernel handling the initialization? 

Relevant TRTLLM tests in flashinfer - https://github.com/flashinfer-ai/flashinfer/blob/446b2c55a7b786c75bad0b4791a6074249d2c0a8/tests/attention/test_trtllm_gen_attention.py#L262-L264

namgyu-youn:
## 📌 Description

TRTLLM-GEN decode kernels require the first 8 MB of the workspace (multi-CTA semaphore counters) to be zeroed before each call. Previously callers worked around this by zeroing the entire 128–256 MB buffer. This PR automates zero-initialization of the TRTLLM-GEN workspace counter region during decode, exposing a C++ helper.

**Changes:**
- Adds `trtllm_fmha_counter_workspace_bytes()` in C++ (TVM-FFI exported) and mirrors it as `_TRTLLM_GEN_MLA_COUNTER_REGION_BYTES = 8192 * 256 * 4` in Python.
- Calls `workspace_buffer[:_TRTLLM_GEN_MLA_COUNTER_REGION_BYTES].zero_()` before each launch in all four TRTLLM decode paths. Per-call (not once at init) because cute-dsl shares the workspace and may corrupt the counter region during autotune (~5 µs on B200).
- Updates docstrings; context/prefill paths are unchanged.

## 🔍 Related Issues

close #2408

## 🚀 Pull Request Checklist

- [x] I have installed the hooks with `pre-commit install`

## 🧪 Tests

- [x] `tests/attention/test_trtllm_gen_attention.py` is updated to apply corresponding changes.
- [x] All tests (`pytest -sv tests/attention/test_trtllm_gen_attention.py::test_create_workspace_buffers`) passing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented stale TRT-LLM-gen counter/semaphore state by zero-initializing the counter region before every decode launch, including paged decode and autotuning runners.

* **New Features**
  * Added and exported a helper that reports the required TRT-LLM-gen counter-region workspace size (bytes), with a runtime consistency check against the Python constant.

* **Documentation**
  * Clarified workspace initialization requirements: decode may require counter-region reset for TRT-LLM-gen, while prefill/context modes do not use the counter region and don’t require zero-initialization.


namgyu-youn:
@gemini-code-assist review again

namgyu-youn:
cc @qsang-nv. Could you please look onto this? It automates zero-initialization of the TRTLLM-GEN decode workspace counter region, replacing the current workaround of zeroing the entire buffer.

qsang-nv:
Thanks — replacing callers' full-buffer zeroing with an 8 MB counter-region memset is the right direction for #2408. A few issues before this can be approved:

**1. A fifth decode call site is missed.** `trtllm_batch_decode_with_kv_cache_mla` has a direct FFI call to `trtllm_paged_attention_decode` in the `cum_seq_lens_q` (var-q / spec-decode) branch (`flashinfer/mla/_core.py` ~L2952) that bypasses the autotune runner and does not zero the counter region. Under the new contract this PR advertises (callers no longer pre-zero), this path launches with garbage semaphores. Please add the same `zero_()` there and update that function's docstring, which still says "Must be zero-initialized on first use".

**2. The new C++ comment is incorrect.** It says "the kernel does NOT self-reset after use", but the kernel does reset the counters at the end of every launch: the existing comment above `_TRTLLM_GEN_MLA_COUNTER_REGION_BYTES` in `flashinfer/mla/_core.py` documents this (and this PR leaves it in place, so the two comments would contradict each other), and the decode test asserts the first 8 MB is all-zero *after* the kernel runs — which passes today with a buffer zeroed only once. The actual reason per-call zeroing helps is that *other* workspace sharers (xqa, other kernels, framework-side reuse) can dirty the region between launches — please state that instead. The cute-dsl motivation in the PR description is also stale: when cute-dsl shares the buffer with trtllm-gen (the autotune scenario), its scratch is already offset past the counter region via `_cute_dsl_workspace_view`.

**3. `trtllm_fmha_counter_workspace_bytes()` is dead code.** It is exported but never called — Python still uses the hardcoded `_TRTLLM_GEN_MLA_COUNTER_REGION_BYTES`, so the PR adds a second source of truth for the 8192×256×4 magic numbers rather than removing one. Either use it (e.g. validate the Python constant once at module load) or drop the export.

**4. Testing.** The PR description cites `tests/attention/test_trtllm_gen_attention.py::test_create_workspace_buffers`, but that file was split into `_decode`/`_prefill` and no test with that name exists anywhere in the repo. Since this touches the trtllm-gen decode hot path, please post B200 results for `test_trtllm_gen_attention_decode.py`, `test_trtllm_gen_attention_prefill.py` (it imports `create_workspace_buffers` from the decode file), and `test_trtllm_gen_mla.py` (covers the var-q branch).

Minor: the 8-line comment in the MLA runner `forward()` should be trimmed to the actual invariant — the placement/perf narrative ("~5us on B200", "Done inside forward() rather than …") doesn't belong in code.

namgyu-youn:
close this PR because #3582 superseded this issue.

PR #3492, opened 2026-06-02
feat(moe): add SWIGLUSTEP activation to CUTLASS fused MoE

### LINKED ISSUE #3486: Step-3.7-Flash-NVFP4 - enable support for MoEActivation.SWIGLUSTEP activation.
When launching stepfun-ai/Step-3.7-Flash-NVFP4 on DGX Spark, vLLM chooses VLLM_CUTLASS instead of FLASHINFER_CUTLASS. When forcing FLASHINFER_CUTLASS MoE backend, it fails with the following error:

```
ValueError: NvFp4 MoE backend 'FLASHINFER_CUTLASS' does not support the deployment configuration since kernel does not support MoEActivation.SWIGLUSTEP activation.
```

Launch command:

```bash
vllm serve stepfun-ai/Step-3.7-Flash-NVFP4 \
 --max-model-len auto \
 -tp 2 \
 --gpu-memory-utilization 0.8 \
 --trust-remote-code \
 --reasoning-parser step3p5  \
 --enable-auto-tool-choice \
 --tool-call-parser step3p5 \
 --load-format fastsafetensors \
 --kv-cache-dtype fp8 \
 --distributed-executor-backend ray \
 --moe-backend FLASHINFER_CUTLASS
```

bkryu:
## 📌 Description

Adds support for the `SwigluStep` activation (clipped `SwiGLU`, used by Step-3.5/3.7-Flash) to the CUTLASS fused-MoE path (`cutlass_fused_moe`).

What `SwigluStep` computes (matches vLLM's `SwigluStepAndMul` reference exactly. See https://github.com/vllm-project/vllm/pull/34478):

```
out = min(silu(gate), limit) * clamp(up, -limit, limit)      # limit = 7.0 for Step-3.5 and 3.7
```
  
Note this is distinct from the existing `SwigluBias` adaptor, which clamps the gate input before the sigmoid; `SwigluStep` clamps the output of SiLU (upper-bound only) and clamps up to [-limit, limit].

Because the activation is applied in the shared `doActivation` / `doGatedActivation` kernels, this works across all `cutlass_fused_moe` dtypes — `bf16`, `fp8`, and `nvfp4`.

####Changes
  
- Enum: added `SwigluStep = 8`, kept in sync across `flashinfer/tllm_enums.py`, `cutlass_kernels/include/common.h`, and `trtllm/fused_moe/runner.h` (+ `serializeActivationType`).
- Kernel (`csrc/fused_moe/cutlass_backend/cutlass_fused_moe_kernels.cuh`): new `SwigluStepAdaptor`, wired into both `doGatedActivation` and the `doActivation` fn_list. The adaptor's limit defaults to 7.0 and is only overridden when the caller passes a per-expert `swiglu_limit` tensor — so selecting `SwigluStep` "just works" with the Step-3.x default while still allowing a custom limit.
- Binding (`flashinfer_cutlass_fused_moe_binding.cu`): the existing swiglu_* → SwigluBias promotion is now guarded so an explicit SwigluStep is preserved while `swiglu_limit` still flows through.
- No public API change: reuses the existing optional `swiglu_limit` parameter on `cutlass_fused_moe`.
- Benchmark: `flashinfer_benchmark.py --routine cutlass_fused_moe` now supports --activation-type SwigluStep across all variants (gated sizing + per-expert limit plumbing); README updated.

Example microbenchmark outputs -- collected on B200 but runnable on SM100/103/120/121
```
# Swiglu
$ python3 flashinfer_benchmark.py --routine cutlass_fused_moe --num_tokens 1 --hidden_size 2048 --intermediate_size 768 --num_experts 128 --top_k 8 --cutlass_variant nvfp4 --quantized_input --input_dtype bfloat16 --activation-type Swiglu --tp_size 1 --tp_rank 0 --ep_size 1 --ep_rank 0 --use_cuda_events --num_iters 50
[PERF] cutlass        :: median time 0.048 ms; std 0.000 ms; achieved tflops 1.559 TFLOPs/sec; achieved tb_per_sec 0.439 TB/sec

# SwigluStep -- Slightly slower due to the clamping in activation
$ python3 flashinfer_benchmark.py --routine cutlass_fused_moe --num_tokens 1 --hidden_size 2048 --intermediate_size 768 --num_experts 128 --top_k 8 --cutlass_variant nvfp4 --quantized_input --input_dtype bfloat16 --activation-type SwigluStep --tp_size 1 --tp_rank 0 --ep_size 1 --ep_rank 0 --use_cuda_events --num_iters 50
[PERF] cutlass        :: median time 0.049 ms; std 0.000 ms; achieved tflops 1.549 TFLOPs/sec; achieved tb_per_sec 0.436 TB/sec
```

## 🔍 Related Issues

#3486

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * Added SwiGLU Step activation for MoE workloads; treated as a gated activation and respects its compile-time default step limit unless an explicit per-expert limit is provided.

* **Documentation**
  * CLI docs updated to list expanded activation options under the --activation-type flag.

* **Tests**
  * Added and expanded tests covering SwiGLU Step across CPU, NVFP4, and fused kernels; test helpers accept activation type and optional step-limit.

### DIFF
--- a/benchmarks/routines/moe.py
+++ b/benchmarks/routines/moe.py
@@ -76,6 +76,40 @@ def _activation_kwarg(fn, activation_type: ActivationType) -> dict:
     return {}
 
 
+# Activations that split FC1 output into gate/up halves (FC1 weight has 2*intermediate rows).
+GATED_ACTIVATIONS = (
+    ActivationType.Swiglu,
+    ActivationType.Geglu,
+    ActivationType.SwigluStep,
+)
+
+
+def _swiglu_limit_kwarg(
+    fn,
+    activation_type: ActivationType,
+    num_local_experts: int,
+    device,
+    limit: float = 7.0,
+) -> dict:
+    """Return the per-expert ``swiglu_limit`` kwarg for activations that clamp the gate.
+
+    Step-3's SwigluStep clamps ``silu(gate)`` and ``up`` to ``[-limit, limit]`` (limit=7.0 in
+    the model). The limit is passed as a per-expert float32 tensor sized to the local expert count.
+    """
+    if activation_type != ActivationType.SwigluStep:
+        return {}
+    if "swiglu_limit" not in inspect.signature(fn).parameters:
+        raise ValueError(
+            "The installed flashinfer version does not support 'swiglu_limit' "
+            "(required for ActivationType.SwigluStep)."
+        )
+    return {
+        "swiglu_limit": torch.full(
+            (num_local_experts,), limit, device=device, dtype=torch.float32
+        )
+    }
+
+
 def run_moe_test(args):
     """
     Run a MOE test.
@@ -872,7 +906,7 @@ def testCutlassFusedMoe(args):
 
     # Create base tensors
     activation_type = args.activation_type
-    is_gated = activation_type in (ActivationType.Swiglu, ActivationType.Geglu)
+    is_gated = activation_type in GATED_ACTIVATIONS
     w1_rows = (2 if is_gated else 1) * intermediate_size
 
     torch.manual_seed(args.random_seed)
@@ -935,6 +969,11 @@ def testCutlassFusedMoe(args):
 
     w31_local, w2_local = build_tp_shards(w31_ep, w2_ep)
 
+    # Per-expert clamp limit for SwigluStep (empty dict for other activations).
+    swiglu_limit_kwarg = _swiglu_limit_kwarg(
+        cutlass_fused_moe, activation_type, w31_local.shape[0], device
+    )
+
     # Prepare variant-specific inputs (outside of the timed/captured region)
     variant = getattr(args, "cutlass_variant", "base")
     out = torch.empty_like(x)
@@ -957,6 +996,7 @@ def testCutlassFusedMoe(args):
                 output=out,
                 enable_pdl=args.enable_pdl,
                 **_activation_kwarg(cutlass_fused_moe, activation_type),
+                **swiglu_limit_kwarg,
             )
 
         input_args_for_bench = (
@@ -1026,6 +1066,7 @@ def testCutlassFusedMoe(args):
                 output=out,
                 enable_pdl=args.enable_pdl,
                 **_activation_kwarg(cutlass_fused_moe, activation_type),
+                **swiglu_limit_kwarg,
             )
 
         input_args_for_bench = (
@@ -1113,6 +1154,7 @@ def testCutlassFusedMoe(args):
                 output=out,
                 enable_pdl=args.enable_pdl,
                 **_activation_kwarg(cutlass_fused_moe, activation_type),
+                **swiglu_limit_kwarg,
             )
 
         input_args_for_bench = (
@@ -1165,7 +1207,7 @@ def testCutlassFusedMoe(args):
         num_experts,
         top_k,
         median_time,
-        is_gated=args.activation_type in (ActivationType.Swiglu, ActivationType.Geglu),
+        is_gated=is_gated,
     )
     tb_per_sec = calculate_moe_kernel_bandwidth(
         num_tokens,
@@ -1181,7 +1223,7 @@ def testCutlassFusedMoe(args):
         routing_logits_dtype=router_logits.dtype,
         active_experts=int(selected_experts.unique().numel()),
         verbose=args.verbose,
-        is_gated=args.activation_type in (ActivationType.Swiglu, ActivationType.Geglu),
+        is_gated=is_gated,
     )
 
     print_perf_metrics(backend, median_time, std_time, tflops, tb_per_sec)
--- a/benchmarks/routines/moe.py
+++ b/benchmarks/routines/moe.py
@@ -84,6 +84,7 @@ def is_gated_activation(activation_type: ActivationType) -> bool:
     return activation_type in (
         ActivationType.Swiglu,
         ActivationType.Geglu,
+        ActivationType.SwigluBias,
         ActivationType.SwigluStep,
     )
 
--- a/benchmarks/routines/moe.py
+++ b/benchmarks/routines/moe.py
@@ -76,12 +76,16 @@ def _activation_kwarg(fn, activation_type: ActivationType) -> dict:
     return {}
 
 
-# Activations that split FC1 output into gate/up halves (FC1 weight has 2*intermediate rows).
-GATED_ACTIVATIONS = (
-    ActivationType.Swiglu,
-    ActivationType.Geglu,
-    ActivationType.SwigluStep,
-)
+def is_gated_activation(activation_type: ActivationType) -> bool:
+    """Whether the activation splits FC1 output into gate/up halves (FC1 weight has 2*intermediate
+    rows). SwigluStep's clamp limit defaults to 7.0 (the Step-3 model value) in the kernel, so no
+    swiglu_limit tensor needs to be passed.
+    """
+    return activation_type in (
+        ActivationType.Swiglu,
+        ActivationType.Geglu,
+        ActivationType.SwigluStep,
+    )
 
 
 def run_moe_test(args):
@@ -880,7 +884,7 @@ def testCutlassFusedMoe(args):
 
     # Create base tensors
     activation_type = args.activation_type
-    is_gated = activation_type in GATED_ACTIVATIONS
+    is_gated = is_gated_activation(activation_type)
     w1_rows = (2 if is_gated else 1) * intermediate_size
 
     torch.manual_seed(args.random_seed)
--- a/benchmarks/routines/moe.py
+++ b/benchmarks/routines/moe.py
@@ -84,32 +84,6 @@ GATED_ACTIVATIONS = (
 )
 
 
-def _swiglu_limit_kwarg(
-    fn,
-    activation_type: ActivationType,
-    num_local_experts: int,
-    device,
-    limit: float = 7.0,
-) -> dict:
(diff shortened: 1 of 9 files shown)


tests/moe/test_trtllm_cutlass_fused_moe.py
```diff
@@ -314,7 +331,14 @@ def compute_with_experts(
         w3_expert, w1_expert = torch.chunk(w31_expert, 2, dim=0)
 
         expert_inputs = x[batch_idx]
-        if alpha is not None and limit is not None and beta is not None:
+        if activation_type == ActivationType.SwigluStep:
+            # Step-3 clipped SwiGLU: min(silu(gate), limit) * clamp(up, -limit, limit).
+            # Defaults to the kernel's default limit (7.0) when not otherwise specified.
+            step_limit = 7.0 if limit is None else limit
+            x1 = F.silu(expert_inputs @ w1_expert.t()).clamp(max=step_limit)
+            x2 = (expert_inputs @ w3_expert.t()).clamp(min=-step_limit, max=step_limit)
```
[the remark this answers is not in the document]
bkryu: Good catch — agreed. Fixed in [commit/latest push]: the SwigluStep branch now indexes limit per-expert so a per-expert swiglu_limit tensor clamps correctly instead of broadcasting/erroring in the next commit

csrc/fused_moe/cutlass_backend/cutlass_fused_moe_kernels.cuh
```diff
@@ -2027,6 +2027,24 @@ struct SwigluBiasAdaptor {
...
+  constexpr static bool IS_GLU = true;
+  float alpha = 1.0f;
+  float beta = 0.0f;
+  float limit = 7.0f;
+
+  template <class T>
+  __device__ T operator()(T const& gate, T const& linear) const {
+    cutlass::epilogue::thread::SiLu<T> fn{};
+    T gate_clamped = cutlass::minimum<T>{}(fn(gate), limit);
+    T linear_clamped = cutlass::maximum<T>{}(cutlass::minimum<T>{}(linear, limit), -limit);
```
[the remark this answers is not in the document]
bkryu: Thanks — generally I agree explicit casts are good hygiene for half/bf16, but I don't think it applies here, and `static_cast<T>(limit)` would actually be incorrect in this context.
  
The activation functor is never instantiated with a half/bf16 `T`. The kernels deliberately compute the activation in fp32: at both call sites the inputs are converted to `ComputeElem = cutlass::Array<float, N>` before invoking the adaptor (doActivationKernel and doGatedActivationKernel), with the existing comment "BF16 isn't   supported, use FP32 for activation function." So T deduces to `cutlass::Array<float, N>` — element type float.

Given that, `cutlass::minimum<T>{}(fn(gate), limit)` resolves to the `cutlass::minimum<Array<float, N>>` scalar overload `operator()(Array<float, N> const&, float const& scalar) (cutlass/array.h:911)`. limit is already float, so there's no narrowing conversion at all.

`static_cast<T>(limit)` would mean `static_cast<cutlass::Array<float, N>>(limit)` — casting a scalar to an Array, which isn't what we want (and changes overload resolution / may not compile). If we wanted an explicit cast it would have to be to the element type (float), which is a no-op here since limit is already float.

For consistency, this is also the exact pattern the neighboring `SwigluBiasAdaptor` and `GLUAdaptor` use (passing the raw float limit/alpha), and it compiles warning-free and passes. So I'd prefer to leave it as-is to stay consistent with the sibling adaptors.

csrc/fused_moe/cutlass_backend/flashinfer_cutlass_fused_moe_binding.cu
```diff
@@ -511,19 +520,28 @@ class FusedMoeRunner : public tvm::ffi::ModuleObj {
...
     if (swiglu_beta.has_value()) {
       CHECK_INPUT_AND_TYPE(swiglu_beta.value(), dl_float32);
       TVM_FFI_ICHECK_EQ(swiglu_beta.value().size(0), num_experts_on_rank)
       "swiglu_beta must have num_experts_on_rank elements.";
-      base_activation_type = ActivationType::SwigluBias;
+      // SwigluStep also uses swiglu_limit but must not be promoted to SwigluBias.
+      if (base_activation_type != ActivationType::SwigluStep) {
+        base_activation_type = ActivationType::SwigluBias;
+      }
     }
```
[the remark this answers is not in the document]
bkryu: Fixed — added the missing <<. While I was there I grepped the file for the same pattern and found one more identical instance at line ~1075 (the W4A8_MXFP4_MXFP8 quant-scale check, behind #ifdef USING_OSS_CUTLASS_MOE_GEMM), so I fixed that one as well. Both are pre-existing.

benchmarks/routines/moe.py
```diff
@@ -76,6 +76,14 @@ def _activation_kwarg(fn, activation_type: ActivationType) -> dict:
     return {}
 
 
+# Activations that split FC1 output into gate/up halves (FC1 weight has 2*intermediate rows).
+GATED_ACTIVATIONS = (
```
aleozlx: [nit] typically is_gated() folds the "in" operator in as well

include/flashinfer/trtllm/fused_moe/runner.h
```diff
@@ -166,7 +166,8 @@ enum class ActivationType : int64_t {
   SwigluBias = 5,
   Relu2 = 6,
   Identity = 7,
-  InvalidType = 8,  // Must be last
+  SwigluStep = 8,
+  InvalidType = 9,  // Must be last
```
aleozlx, line 170: hmm i did't notice this before... changing value is a bit odd tho it is only internal.

csrc/fused_moe/cutlass_backend/flashinfer_cutlass_fused_moe_binding.cu
```diff
@@ -345,19 +345,28 @@ class FusedMoeRunner : public tvm::ffi::ModuleObj {
       CHECK_INPUT_AND_TYPE(swiglu_alpha.value(), dl_float32);
       TVM_FFI_ICHECK_EQ(swiglu_alpha.value().size(0), num_experts_on_rank)
           << "swiglu_alpha must have num_experts_on_rank elements.";
-      base_activation_type = ActivationType::SwigluBias;
+      // SwigluStep also uses swiglu_limit but must not be promoted to SwigluBias.
+      if (base_activation_type != ActivationType::SwigluStep) {
```
aleozlx: suggest flip it to positive logic

assign bias

if step
  assign step

prevents future injection into the unbounded negative logic

benchmarks/routines/moe.py
bkryu: Good point let me change this to `is_gated_activation()`

include/flashinfer/trtllm/fused_moe/runner.h
bkryu: Shifted the enum value because I explicitly saw `// Must be last` 👀. 

LMK if you think another fix might be better

csrc/fused_moe/cutlass_backend/flashinfer_cutlass_fused_moe_binding.cu
bkryu: Sure, flipped to positive logic. The logic here pertains to promoting the base activation from `Swiglu` to `SwigluBias`. Swiglu + swiglu_alpha/beta/limit maps to the SwigluBias kernel, and everything else is left as the coaller set it.

benchmarks/routines/moe.py
```diff
@@ -872,7 +884,7 @@ def testCutlassFusedMoe(args):
 
     # Create base tensors
     activation_type = args.activation_type
-    is_gated = activation_type in (ActivationType.Swiglu, ActivationType.Geglu)
+    is_gated = is_gated_activation(activation_type)
     w1_rows = (2 if is_gated else 1) * intermediate_size
```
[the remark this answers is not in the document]
bkryu: TRTLLM does not support the newly added activation so we can leave it as now to minimize diffs.

include/flashinfer/trtllm/fused_moe/runner.h
aleozlx: yeah i think is fix would be out of scope for this PR
 no action requried here

aleozlx (approved):

flashinfer/tllm_enums.py
```diff
@@ -38,7 +38,8 @@ class ActivationType(IntEnum):
     SwigluBias = 5
     Relu2 = 6
     Identity = 7
-    InvalidType = 8
+    SwigluStep = 8
```
djns99: Can we have identity as the second-last item?

I consider Identity and InvalidType as special cases.

Maybe we should change their value to a larger value (100/101) as well so we dont have to change the enum values to add new items in future?

djns99 (approved):

flashinfer/tllm_enums.py
bkryu: Sure I can make identity the second last item.

Regarding making them 100/101, I'll defer to a future followup PR for now

djns99: Yeah I guess 100/101 might not work with the current direct cast -> array lookup logic.

Maybe we can make "invalid" 100, and move identity first.

Can be a follow-up PR

PR #3504, opened 2026-06-04
Add MXFP8 MoE SwiGLU OA parameters

ruoqianguo:
## 📌 Description

This PR adds optional per-expert FC1 gated-activation parameters to the TensorRT-LLM FP8 block-scale / MXFP8 MoE path.

Changes include:

  - Expose `gemm1_alpha`, `gemm1_beta`, and `gemm1_clamp_limit` on `trtllm_fp8_block_scale_moe` and `trtllm_fp8_block_scale_routed_moe`.                                                               
  - Validate the optional tensors as float32 `[local_num_experts]` in the C++ launcher.
  - Forward the parameters to the trtllm-gen fused FC1 SwiGLU epilogue.
  - Update trace/reference plumbing for the same activation semantics.
  - Add an MXFP8 MoE regression test covering default behavior, no-op parameters, alpha+clamp, and non-default beta behavior.                                                             

  The new parameters default to `None`, so existing public Python API call sites preserve the previous behavior.          

## 🔍 Related Issues

N/A

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

Validation run:

  - `git diff --check`
  - `python3.11 -m py_compile tests/moe/test_trtllm_gen_fused_moe.py flashinfer/fused_moe/core.py flashinfer/trace/templates/moe.py`
  - B300: `tests/moe/test_trtllm_gen_fused_moe.py::test_mxfp8_block_scale_moe_swiglu_oa_activation_params -q -s`
 

## Reviewer Notes

For MXFP8 block-scale MoE, these values are forwarded as raw per-expert scalars. FlashInfer does not apply host-side scalar dequant-scale conversion.

  gemm1_beta is included together with gemm1_alpha and gemm1_clamp_limit because trtllm-gen supports the full fused SwiGLU epilogue parameter set, and the beta path needs explicit coverage when non-default values are provided.

## Summary by CodeRabbit

* **New Features**
  * FP8 block-scale MoE adds three optional per-expert SwiGLU parameters (gemm1_alpha, gemm1_beta, gemm1_clamp_limit) across public APIs, trace templates, runtime paths, and kernel calls to enable per-expert activation tuning and optional clamping.
* **Validation**
  * Runtime validation ensures these params are only accepted with MxFp8 quantization + SwiGLU and checks device, dtype, shape, contiguity, and per-expert length.
* **Tests**
  * End-to-end and trace tests added for validation, default/no-op, and alpha/beta/clamp behavior.


csrc/trtllm_fused_moe_kernel_launcher.cu
```diff
@@ -360,6 +360,18 @@ class FusedMoeLauncher {
...
+                                              std::string const& tensor_name) const {
+    if (!tensor.has_value()) {
+      return;
+    }
+    auto const& value = tensor.value();
+    TVM_FFI_ICHECK_EQ(value.dtype(), dl_float32) << tensor_name << " must be float32.";
+    TVM_FFI_ICHECK_EQ(value.ndim(), 1) << tensor_name << " must be 1D.";
+    TVM_FFI_ICHECK_EQ(value.size(0), args->local_num_experts)
+        << tensor_name << " must have shape [local_num_experts].";
+  }
```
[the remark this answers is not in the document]
ruoqianguo: Agreed. These optional tensors are passed as raw device pointers to the CUDA kernel, so they should be validated as CUDA tensors on the same device as hidden_states.

flashinfer/trace/templates/moe.py
```diff
@@ -91,8 +94,14 @@ def _fp8_moe_run_experts(
...
         G1 = A_e.matmul(W13[le].t())
         X1, X2 = G1[:, :I], G1[:, I:]
-        silu_X2 = X2 / (1.0 + torch.exp(-X2))
-        O = (silu_X2 * X1).matmul(W2[le].t())
+        if gemm1_clamp_limit is not None:
+            limit = gemm1_clamp_limit[le].to(torch.float32)
+            X1 = torch.clamp(X1, min=-limit, max=limit)
+            X2 = torch.clamp(X2, max=limit)
+        alpha = 1.0 if gemm1_alpha is None else gemm1_alpha[le].to(torch.float32)
+        beta = 0.0 if gemm1_beta is None else gemm1_beta[le].to(torch.float32)
```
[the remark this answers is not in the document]
ruoqianguo: Agreed. The trace/reference path should explicitly move the scalar values to the activation tensor device when converting to float32.

flashinfer/trace/templates/moe.py
```diff
@@ -2334,6 +2433,24 @@ def _unpack_int4(packed):
...
+            dtype="float32",
+            optional=True,
+            description="Per-expert SwiGLU beta.",
+        ),
+        "gemm1_clamp_limit": Tensor(
+            ["num_local_experts"],
+            dtype="float32",
+            optional=True,
+            description="Per-expert SwiGLU clamp limit.",
+        ),
```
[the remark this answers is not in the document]
ruoqianguo: > medium
> 
> The `gemm1_alpha`, `gemm1_beta`, and `gemm1_clamp_limit` inputs are added to the `trtllm_fp4_block_scale_routed_moe_trace` template, but the corresponding reference implementation `_fp4_moe_run_experts` (and `_trtllm_fp4_block_scale_routed_moe_reference`) does not currently support or forward these parameters. This could lead to reference mismatch errors if FP4 block-scale routed MoE is ever tested with non-default values for these parameters.

Thanks for pointing this out. I checked the current diff and this PR does not add these optional parameters to trtllm_fp4_block_scale_routed_moe_trace. The FP4 standard trace already had related optional inputs before this PR, while the FP4 reference path does not currently consume them. Since this PR is scoped to FP8 block-scale / MXFP8 MoE support, I will avoid expanding FP4 activation semantics in this change. FP4 reference support for these parameters can be handled as a separate follow-up if needed.

ruoqianguo:
@samuellees  Please review this, thanks

tests/moe/test_trtllm_gen_fused_moe.py
```diff
@@ -2374,7 +2395,35 @@ def run_moe_dequant(args, quant_mode: QuantMode):
...
             my_x2 = my_a[:, args.intermediate_size :]
-            activation_output[i : i + my_num_tokens] = activation_func(my_x2) * my_x1
+            if args.gemm1_clamp_limit is not None:
+                limit = args.gemm1_clamp_limit[expert_idx].to(
+                    device=my_x1.device, dtype=torch.float
+                )
+                my_x1 = torch.clamp(my_x1, min=-limit, max=limit)
+                my_x2 = torch.clamp(my_x2, max=limit)
+            if args.gemm1_alpha is not None or args.gemm1_beta is not None:
+                assert int(args.activation_type) == int(ActivationType.Swiglu)
```
samuellees, line 2458: Please add this check into our kernel interface at C++ or python level

tests/moe/test_trtllm_gen_fused_moe.py
```diff
@@ -4657,6 +4732,170 @@ def test_fp8_block_scale_routed_activation_type_relu2_smoke():
     assert mismatch_pct < 10, f"Mismatch percentage is {mismatch_pct:.2f}%"
 
 
+def test_mxfp8_block_scale_moe_swiglu_oa_activation_params(cache_permute_indices):
+    """MxFP8 MoE applies raw trtllm-gen fused FC1 SwiGLU alpha/clamp params."""
```
samuellees: Do we support MXFP8? If yes, related test should be added.
If no, please modify api doc

flashinfer/fused_moe/core.py
```diff
@@ -1904,6 +1907,9 @@ def trtllm_fp8_block_scale_moe_op(
         gemm1_weights: torch.Tensor,
         gemm1_weights_scale: torch.Tensor,
         gemm1_lora_delta: Optional[torch.Tensor],
+        gemm1_alpha: Optional[torch.Tensor],
```
samuellees, line 1983: alpha, beta, clamp;  which one is necessary? What if user just pass one? What's the valid combination please?

flashinfer/fused_moe/core.py
```diff
@@ -3026,6 +3044,10 @@ def trtllm_fp8_block_scale_moe(
             Column order matches topk_indices. When None (default), zero overhead - the
             kernel skips the write entirely. Buffer may be larger than num_tokens for CUDA
             graph pre-allocation; only rows [0, num_tokens) are written.
+        gemm1_alpha / gemm1_beta / gemm1_clamp_limit: optional ``[local_num_experts]``
```
samuellees: Can you describe the math formula, default value, please?

tests/moe/test_trtllm_gen_fused_moe.py
```diff
@@ -4657,6 +4732,170 @@ def test_fp8_block_scale_routed_activation_type_relu2_smoke():
...
+    output_oa_ref, output_oa = run_case(
+        gemm1_alpha=alpha, gemm1_clamp_limit=clamp_limit
+    )
+    output_minimax_oa_ref, output_minimax_oa = run_case(
+        gemm1_alpha=alpha, gemm1_beta=beta, gemm1_clamp_limit=clamp_limit
+    )
+
+    torch.testing.assert_close(output_oa, output_oa_ref, atol=1e-1, rtol=1e-1)
+    torch.testing.assert_close(
+        output_minimax_oa, output_minimax_oa_ref, atol=1e-1, rtol=1e-1
```
samuellees: Please make sure the threshold makes sense for mxfp8^ ^

samuellees (commented):
Left some comments

flashinfer/fused_moe/core.py
djns99: Could we also consider a wrapper class

```
class ExtraSwiGLUParams:
    ...
```

I know this is not consistent with the rest of the API, so happy for a no answer, but lets at least have a discussion (+ @aleozlx for feedback based on unified MOE design)

tests/moe/test_trtllm_gen_fused_moe.py
ruoqianguo: Agreed. The reference currently has this assertion only in the test path, but the public/kernel interface should reject unsupported activation combinations earlier. I will add an interface-level validation so these optional OA parameters are only accepted with ActivationType.Swiglu, matching the current reference and test coverage.

flashinfer/fused_moe/core.py
ruoqianguo: None of them is required. Each parameter is optional and can be provided independently:
- `gemm1_alpha=None` means alpha = 1.0
- `gemm1_beta=None` means beta = 0.0
- `gemm1_clamp_limit=None` means no clamp / +inf

Any subset is valid for SwiGLU. If only one is passed, the other parameters use their defaults. I will document this explicitly and add validation that these OA parameters are only used with `ActivationType.Swiglu` in this PR.

flashinfer/fused_moe/core.py
ruoqianguo: Sure. I will add the formula and defaults to the API docstring. The intended SwiGLU OA math is:

Let the GEMM1 output be split into `X1` (linear/up half) and `X2` (gate half). If `gemm1_clamp_limit` is provided:
  `X1 = clamp(X1, -limit, limit)` and `X2 = clamp(X2, max=limit)`.

Then the fused activation output is:
  `Y = X2 * sigmoid(alpha * X2) * (X1 + beta)`.

Defaults are alpha = 1.0, beta = 0.0, and no clamp when the corresponding tensor is None. For MXFP8 block-scale MoE these are raw values; no host-side dequant-scale conversion is applied.

tests/moe/test_trtllm_gen_fused_moe.py
ruoqianguo: Thanks for clarifying. The intended scope of this PR is MXFP8 block-scale MoE only for the new OA parameters.

The `trtllm_fp8_block_scale_moe` entry point is shared by the existing FP8 block-scale modes, so I agree we should make the constraint explicit rather than leaving it implicit. I will add validation/documentation so `gemm1_alpha`, `gemm1_beta`, and `gemm1_clamp_limit` are only accepted for `Fp8QuantizationType.MxFp8` with `ActivationType.Swiglu` in this PR.

If we want to support the other blockwise FP8 mode through the same API later, we should add a matching numerical test for that mode and document the parameter semantics separately.

tests/moe/test_trtllm_gen_fused_moe.py
ruoqianguo: I checked the existing MXFP8 tolerances in the test suite. The `atol=1e-1, rtol=1e-1` used here follows the existing MXFP8xMXFP8 CUTLASS MoE test, which compares the MXFP8 kernel against a dequantized reference with the same tolerance: [test_moe_mxfp8_mxfp8](https://github.com/flashinfer-ai/flashinfer/blob/main/tests/moe/test_trtllm_cutlass_fused_moe.py#L1478-L1573)

It is also stricter than the generic FP8 block-scale MoE randomized tolerance in this file, which allows partial element mismatch for broader cases. This OA test uses a deterministic identity-like setup, so requiring all elements to match within `1e-1/1e-1` is reasonable. I can add a short comment near the assertion to make this rationale explicit.

flashinfer/fused_moe/core.py
ruoqianguo: > Could we also consider a wrapper class
> 
> ```
> class ExtraSwiGLUParams:
>     ...
> ```
> 
> I know this is not consistent with the rest of the API, so happy for a no answer, but lets at least have a discussion (+ @aleozlx for feedback based on unified MOE design)

For the broader unified MoE API redesign (cc @aleozlx ), happy to fold these into a ExtraSwiGLUParams dataclass in a follow-up if that's the agreed pattern; in this PR keeping kwargs for backward compat

samuellees: Thanks @djns99 @ruoqianguo !

samuellees (approved):
LGTM, waiting CI～

ruoqianguo:
Hello @samuellees , The CI has finished running, and the failures reported in CI are unrelated to my changes. The error in tests/moe/test_trtllm_gen_fused_moe.py::test_bf16_moe_swiglu_oa_activation_params was already fixed in this PR (which has been merged): https://github.com/flashinfer-ai/flashinfer/pull/3580

samuellees:
https://github.com/flashinfer-ai/flashinfer/actions/runs/27333207695/job/81015763162?pr=3504#step:5:11175 
Hi @ruoqianguo , seems this CI error is related with this PR, could you address this please?

ruoqianguo:
Hello @samuellees ,The CI failure was caused by omitted optional MoE trace inputs (gemm1_alpha, gemm1_beta, gemm1_clamp_limit) being emitted with dtype "unknown", which the trace consistency test rejects. The fix makes fi_trace fall back to the descriptor-declared dtype for omitted optional tensor inputs, while preserving the actual runtime tensor dtype when the input is provided. I have addressed the ci error. Thanks

PR #888, opened 2025-02-23
feat - support mla kvcache store

baowendin:
### Summary
related to [#877 ](https://github.com/flashinfer-ai/flashinfer/issues/877)
This PR implement MLA cache store，and passed correctness test in the case of ckv_dim=512 and kpe_dim=64，but no further performance test yet. Sincerely hope somebody who familiar with CUDA can help improve performance.

PR #3524, opened 2026-06-06
perf: vectorize AppendPagedKVMlaCache to 128-bit access

aws-jiadingg:
## 📌 Description

`AppendPagedKVMlaCache` (the MLA paged-KV append used by DeepSeek-V3/R1 MLA serving) hard-coded `vec_size = 2`, i.e. 4-byte (2× bf16) global loads/stores. With `head_dim_ckv=512` this leaves the memory subsystem 4× under-utilized per transaction.

This sets `vec_size` to use 128-bit (16-byte) vectorized access — `std::max(16/sizeof(DType), HEAD_CKV_DIM/128)` — mirroring the idiom already used by the sibling `AppendPagedKVCache` kernel (page.cuh:303). For bf16 this is 8 elems/thread (16 B); for fp8 it is 16 elems/thread (16 B). The `/128` (vs the sibling's `/32`) keeps access at exactly 128-bit for the 512-wide ckv dim. The k-pe path (`head_dim_kpe=64`) remains correctly covered.

**Measured on H200 (CUDA 12.8), bf16, head_dim_ckv=512/kpe=64, page_size=64:**

| append tokens (nnz) | before | after | speedup |
|---|---|---|---|
| 16384 | 15.7 µs | 8.8 µs | **1.80×** |
| 32768 | 49.1 µs | 27.2 µs | **1.80×** |
| 65536 | 93.0 µs | 49.8 µs | **1.87×** |

Small batches (nnz ≤ 4096) are launch-latency-bound and unchanged.

**Correctness:** bf16 output is bit-identical to the original (KV-cache checksum matches across the sweep). The fp8 (e4m3) path was verified clean under `compute-sanitizer memcheck` (0 errors).

## 🔍 Related Issues

N/A

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

- [] All tests are passing (`unittest`, etc.).

## Reviewer Notes

Pure vectorization-width change in the launcher; kernel body unchanged. Behavior-preserving — only the load/store granularity differs. Matches the existing `AppendPagedKVCache` vec_size idiom (using `/128` instead of `/32` so the 512-wide ckv dim stays at exactly 128-bit access).

## Summary by CodeRabbit

* **Refactor**
  * Optimized kernel memory operation performance for improved efficiency in paged KV cache append operations.

### DIFF
--- a/include/flashinfer/page.cuh
+++ b/include/flashinfer/page.cuh
@@ -851,7 +851,8 @@ cudaError_t AppendPagedKVMlaCache(paged_kv_mla_t<DType, IdType> paged_kv, DType*
   constexpr uint32_t HEAD_KPE_DIM = 64;
   FLASHINFER_CHECK(head_dim_ckv == HEAD_CKV_DIM, "head_dim_ckv must be equal to 512");
   FLASHINFER_CHECK(head_dim_kpe == HEAD_KPE_DIM, "head_dim_kpe must be equal to 64");
-  constexpr uint32_t vec_size = 2;
+  constexpr uint32_t vec_size =
+      std::max(uint32_t(16 / sizeof(DType)), uint32_t(HEAD_CKV_DIM / 128));
 
   uint32_t bdx = HEAD_CKV_DIM / vec_size;
   uint32_t num_threads = bdx;


jhaotingc:
Originated from https://github.com/flashinfer-ai/flashinfer/pull/888/changes#diff-051a27a481433e5a5e8a0fa0999b0ae9611084497df5a584d8d36f6654c490c3R621.
Hi @yzh119, seems that other PageKVCache kernel has the same [std::max(16 / sizeof(DType), HEAD_DIM / 32);](https://github.com/aws-jiadingg/flashinfer/blame/09872a1d79fa616d9b2d1def05ad5119d25b5ee0/include/flashinfer/page.cuh#L303) but the MLA one. 
Would you mind reviewing this PR, thank you!

include/flashinfer/page.cuh
```diff
@@ -569,7 +569,8 @@ cudaError_t AppendPagedKVMlaCache(paged_kv_mla_t<DType, IdType> paged_kv, DType*
   constexpr uint32_t HEAD_KPE_DIM = 64;
   FLASHINFER_CHECK(head_dim_ckv == HEAD_CKV_DIM, "head_dim_ckv must be equal to 512");
   FLASHINFER_CHECK(head_dim_kpe == HEAD_KPE_DIM, "head_dim_kpe must be equal to 64");
-  constexpr uint32_t vec_size = 2;
+  constexpr uint32_t vec_size =
```
saltyminty, line 854: With this change, `uint32_t(HEAD_CKV_DIM / 128)`=4, so we're effectively having vec_size to a minimum of 4. Would this change tensor alignment assumptions anywhere?

jhaotingc:
```
FAILED tests/moe/test_cute_dsl_fused_moe.py::TestCuteDslMoEWrapper::test_cuda_graph_wrapper_lifetime_after_autotune - AssertionError: autotune(True) did not populate a CuteDslMoEWrapper::run cache entry
assert False
```

There seem to be failed test here @aws-jiadingg

include/flashinfer/page.cuh
```diff
@@ -569,7 +569,8 @@ cudaError_t AppendPagedKVMlaCache(paged_kv_mla_t<DType, IdType> paged_kv, DType*
...
+      std::max(uint32_t(16 / sizeof(DType)), uint32_t(HEAD_CKV_DIM / 128));
```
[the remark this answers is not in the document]
aws-jiadingg: Agreed that, for current constants it's equivalent.The HEAD_CKV_DIM / 128 floor only matters if someone adds a dtype with sizeof > 4 in the future. btw, I was copying this from the non-MLA pattern (std::max(16/sizeof(DType), HEAD_DIM/32)) for consistency:

```
  - https://github.com/flashinfer-ai/flashinfer/blob/main/include/flashinfer/page.cuh#L379
  - https://github.com/flashinfer-ai/flashinfer/blob/main/include/flashinfer/page.cuh#L419
```

include/flashinfer/page.cuh
aws-jiadingg: Thanks for flagging this. No alignment issue because vec_size * sizeof(DType) always equals 16 bytes (one 128-bit load) regardless of dtype. Each thread reads at byte offset `tx * vec_size * sizeof(DType) = tx * 16B` within the row, which is 16B-aligned by construction. The inner dims (512 / 64) are asserted at lines 850–853 and are divisible by the largest possible vec_size (16 for fp8), so row strides also stay aligned.

Btw, the non-MLA AppendPagedKVCache uses the same pattern (page.cuh:379 (https://github.com/flashinfer-ai/flashinfer/blob/main/include/flashinfer/page.cuh#L379)).

aws-jiadingg:
> ```
> FAILED tests/moe/test_cute_dsl_fused_moe.py::TestCuteDslMoEWrapper::test_cuda_graph_wrapper_lifetime_after_autotune - AssertionError: autotune(True) did not populate a CuteDslMoEWrapper::run cache entry
> assert False
> ```
> 
> There seem to be failed test here @aws-jiadingg

This looks like a pre-existing main failure. Other unrelated PRs are hitting the same ~10/20 CI pass rate around the same timeframe (e.g., [#3846](https://github.com/flashinfer-ai/flashinfer/pull/3846), [#3871](https://github.com/flashinfer-ai/flashinfer/pull/3871)). This PR is a one-line vec_size constant in the MLA KV-append launcher (page.cuh) — no code path to CuteDslMoEWrapper. Looks like a flaky/broken test on current main?

saltyminty (approved):

PR #3551, opened 2026-06-09
fix: handle negative draft_token_ids in ChainSpeculativeSampling

### LINKED ISSUE #879: Possible Bug in chain_speculative_sampling: output_emitted_token_num Exceeds Expected Limit
I was testing chain_speculative_sampling and noticed that output_emitted_token_num sometimes exceeds the expected limit. 
Specifically, in my test case, it should never be greater than 3, but in some instances, it returns 4, even though probability values for such a case don’t exist.

I used the following test script:

```python
import torch
from flashinfer.sampling import chain_speculative_sampling

def run_chain_speculative_sampling(draft_probs, draft_token_ids, verify_probs):
    batch_size, num_draft_tokens = draft_token_ids.shape
    uniform_samples = torch.rand(batch_size, num_draft_tokens + 1, device=draft_token_ids.device)
    accepted_token_ids, output_accepted_token_num, output_emitted_token_num =\
        chain_speculative_sampling(draft_probs, draft_token_ids, uniform_samples, verify_probs)
    
    return accepted_token_ids, output_accepted_token_num, output_emitted_token_num

token_ids = torch.tensor([[ 1,  0,  1, -1, -1]], device='cuda:0')
draft_probs = torch.tensor([[[0.5883, 0.4117, 0.0000, 0.0000],
                             [1.0000, 0.0000, 0.0000, 0.0000],
                             [0.3803, 0.5435, 0.0762, 0.0000],
                             [0.0000, 0.0000, 0.0000, 0.0000],
                             [0.0000, 0.0000, 0.0000, 0.0000]]], device='cuda:0')

verify_probs = torch.tensor([[[0.4555, 0.5445, 0.0000, 0.0000],
                             [1.0000, 0.0000, 0.0000, 0.0000],
                             [0.5783, 0.2831, 0.0000, 0.1386],
                             [0.2346, 0.6850, 0.0804, 0.0000],
                             [0.0000, 0.0000, 0.0000, 0.0000],
                             [0.0000, 0.0000, 0.0000, 0.0000]]], device='cuda:0')

for i in range(10):
    accepted_token_ids, output_accepted_token_num, output_emitted_token_num = run_chain_speculative_sampling(draft_probs, token_ids, verify_probs)

    print(f"------------------------------------------------------")
    print(f"accepted_token_ids : {accepted_token_ids.squeeze(0).tolist()}")
    print(f"output_accepted_token_num : {output_accepted_token_num.item()}")
    print(f"output_emitted_token_num : {output_emitted_token_num.item()}")

```

Reason for having unnecessary extra tokens and probabilities is because I have detected abnormal output when I was testing batching. I have isolated the batch that was causling the problem, and could still generate the same behavior.

Here are some sample outputs from multiple runs:

```python
------------------------------------------------------
accepted_token_ids : [1, 0, 3, -1, -1, -1]
output_accepted_token_num : 3
output_emitted_token_num : 2
------------------------------------------------------
accepted_token_ids : [1, 0, 1, -1, 3, -1] # Unexpected 3 at 4th index!
output_accepted_token_num : 4 # Unexpected!
output_emitted_token_num : 4  # Unexpected!
------------------------------------------------------
accepted_token_ids : [1, 0, 0, -1, -1, -1]
output_accepted_token_num : 3
output_emitted_token_num : 2
------------------------------------------------------
accepted_token_ids : [1, 0, 1, -1, 3, -1] # Unexpected 3 at 4th index!
output_accepted_token_num : 4 # Unexpected!
output_emitted_token_num : 4  # Unexpected!
```

My torch version is 2.4.0 with cuda version 12.1, and I've tested the case in both flashinfer 0.2.0 and 0.2.1

Oxygen56:
## Summary

Fixes #879.

`ChainSpeculativeSampling` kernel used `draft_token_ids[i]` directly as an array index into `draft_probs` and `target_probs` without validating the value. When the draft model produces fewer tokens than `num_speculate_tokens`, the remaining positions are filled with `-1` sentinel values. Using `-1` as an index causes **out-of-bounds memory access**, reading garbage values that happen to satisfy the acceptance condition (`u * 0 < garbage_q`), leading to:

- Invalid `-1` tokens being "accepted" as draft output
- `output_emitted_token_num` exceeding the number of valid draft tokens
- Nondeterministic incorrect results

## Changes

- **`include/flashinfer/sampling.cuh`**: Add `draft_id < 0` guard in both the contiguous-acceptance loop (line ~1884) and the post-rejection counting loop (line ~1900). Negative IDs are treated as immediate rejections.
- **`tests/utils/test_sampling.py`**: Add regression test using the exact reproduction case from #879, verifying that `emitted_token_num` never exceeds the count of valid (non-negative) draft tokens.

## Test plan

- [ ] Existing `test_chain_speculative_sampling` tests pass (no behavioral change for valid inputs)
- [ ] New `test_chain_speculative_sampling_negative_draft_ids` verifies the fix over 50 trials with the exact inputs from the bug report

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed speculative sampling so early-terminated draft tokens with sentinel values no longer cause incorrect outputs or indexing errors.
* **Tests**
  * Added a regression test to ensure sampling correctly handles trailing sentinel draft tokens and does not emit or accept tokens beyond valid drafts.

### DIFF
--- a/include/flashinfer/sampling.cuh
+++ b/include/flashinfer/sampling.cuh
@@ -1882,7 +1882,7 @@ __global__ void ChainSpeculativeSampling(DType* draft_probs, IdType* draft_token
   uint32_t pos = num_speculative_tokens;
   for (uint32_t i = 0; i < num_speculative_tokens; ++i) {
     IdType draft_id = draft_token_ids[row_idx * num_speculative_tokens + i];
-    if (draft_id < 0) {
+    if (static_cast<int32_t>(draft_id) < 0) {
       pos = i;
       break;
     }
@@ -1900,8 +1900,8 @@ __global__ void ChainSpeculativeSampling(DType* draft_probs, IdType* draft_token
   uint32_t emitted_token_num = pos;
   uint32_t accepted_token_num = pos;
   for (uint32_t i = pos; i < num_speculative_tokens; ++i) {
-    int draft_id = draft_token_ids[row_idx * num_speculative_tokens + i];
-    if (draft_id < 0) {
+    IdType draft_id = draft_token_ids[row_idx * num_speculative_tokens + i];
+    if (static_cast<int32_t>(draft_id) < 0) {
       break;
     }
     float q = target_probs[(row_idx * (num_speculative_tokens + 1) + i) * d + draft_id],
--- a/include/flashinfer/sampling.cuh
+++ b/include/flashinfer/sampling.cuh
@@ -1882,11 +1882,14 @@ __global__ void ChainSpeculativeSampling(DType* draft_probs, IdType* draft_token
   uint32_t pos = num_speculative_tokens;
   for (uint32_t i = 0; i < num_speculative_tokens; ++i) {
     IdType draft_id = draft_token_ids[row_idx * num_speculative_tokens + i];
+    if (draft_id < 0) {
+      pos = i;
+      break;
+    }
     float q = target_probs[(row_idx * (num_speculative_tokens + 1) + i) * d + draft_id],
           p = draft_probs[(row_idx * num_speculative_tokens + i) * d + draft_id];
     float u = curand_uniform(&curand_state);
     if (u * p < q) {
-      // accept the draft models output
       output_token_ids[row_idx * (num_speculative_tokens + 1) + i] = draft_id;
     } else {
       pos = i;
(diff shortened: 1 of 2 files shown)


kahyunnam:
bot?

Oxygen56:
Hi, yes, I am a real person and I opened this PR myself. Happy to follow up on any code or test concerns here.

PR #3564, opened 2026-06-10
perf(moe): Enhance CuteDSL NVF4 MOE Perf

liyuhannnnn:
## 📌 Description

Cherry pick perf optimization changes from TRTLLM PR https://github.com/NVIDIA/TensorRT-LLM/pull/15092.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved grouped GEMM/MoE finalize behavior by refining epilogue synchronization and accumulator consumption semantics, improving output stability.
  * Enhanced handling of padded/invalid rows and per-row scaling metadata to reduce incorrect results in edge cases.

* **Performance**
  * Improved finalize execution efficiency with dedicated metadata staging and streamlined per-row scaling in the output pipeline.
  * Updated kernel tactic validation to correctly respect the runtime dtype of final scaling values.


aleozlx:
waiting on CI result to approve

flashinfer/fused_moe/cute_dsl/blockscaled_contiguous_grouped_gemm_finalize_fusion.py
leejnau, line 480: The TRT-LLM path this was ported from requires `token_final_scales.dtype == torch.float32`. Since this PR now sets `alpha * token_scale` through shared memory as `final_scale_dtype`, allowing bf16/fp16 here can introduce extra rounding versus the old path.

Maybe we should match TRT-LLM and reject non-fp32 routing scales for this kernel unless/until `meta_scale` is fp32?

liyuhannnnn: Added cantImplement rule to only accept `final_scale_dtype==fp32`.

flashinfer/fused_moe/cute_dsl/blackwell/blockscaled_contiguous_gather_grouped_gemm_act_fusion.py
```diff
@@ -1263,8 +1263,10 @@ def kernel(
 
         # Pipeline Init: Initialize acc_pipeline (barrier) and states
         acc_pipeline_producer_group = pipeline.CooperativeGroup(pipeline.Agent.Thread)
-        num_acc_consumer_threads = len(self.epilog_warp_id) * (
-            2 if use_2cta_instrs else 1
+        num_acc_consumer_threads = (
+            len(self.epilog_warp_id)
+            * self.threads_per_warp
+            * (2 if use_2cta_instrs else 1)
```
vitamin-chaos, line 1269: I'd like to ask whether removing 'elect_one' this way is more of a habitual writing style, and whether there would be any performance difference

leejnau (approved):

aleozlx (approved):

flashinfer/fused_moe/cute_dsl/blackwell/blockscaled_contiguous_gather_grouped_gemm_act_fusion.py
liyuhannnnn: >> I'd like to ask whether removing 'elect_one' this way is more of a habitual writing style

I don't think `elect_one` is a habitual writing style for all kinds of situation, for tma/mma style, elect one should be used. For epilogue acc pipeline(with 128 threads in epilogue), we generally should not use elect one and let all threads participant the pipeline release.

>> and whether there would be any performance difference

Furtherly, about the reason of performance difference, normally in our cutedsl codes, I expect the acc pipeline early release happens just after we load acc from tmem, however, it doesn't happen as expect if we add elect_one, once I remove elect one, it happens. I know if have elect one and the acc pipeline release pattern doesn't have direct relationship, something may happen on compiler side.

vitamin-chaos: Thanks for the detailed explanation.

PR #3573, opened 2026-06-10
Feat/moe autotuner ep aware

qiching:
## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

- **New Features**
  - Added distributed autotuning synchronization to keep selected kernel tactics consistent across devices/process groups.
  - Added balanced “local expert” top‑k routing for MoE to match sharded expert layouts.
- **Bug Fixes**
  - Improved autotuning correctness for sharded expert routing and scenarios where routing logits are provided.
  - Updated legacy/opt-in dummy routing behavior via an environment-variable switch.
  - Refined autotuning selection to compare against a heuristic baseline with a configurable switch margin.
- **Tests**
  - Added multi-rank and single-rank benchmark/validation scripts for determinism, latency, and process-group behavior.

### DIFF
--- a/flashinfer/autotuner/autotuner.py
+++ b/flashinfer/autotuner/autotuner.py
@@ -6,6 +6,7 @@ import inspect
 import itertools
 import json
 import os
+import sys
 import tempfile
 import threading
 import weakref
@@ -647,6 +648,7 @@ def autotune(
     tuning_buckets: tuple[int, ...] | None = None,
     round_up: bool | None = None,
     skip_ops: str | set[str] | None = None,
+    distributed_process_group: Any | None = None,
 ):
     """Context manager for autotuning with optional file-based caching.
 
@@ -714,6 +716,16 @@ def autotune(
             ``autotune(skip_ops={"A"})`` skips both ``"A"`` and ``"B"``.
             Common op names: ``"fp4_gemm"``, ``"bf16_gemm"``,
             ``"fp8_gemm"``, ``"mxfp8_gemm"``.
+        distributed_process_group: Optional ``torch.distributed`` process group
+            used by the ``broadcast`` distributed tuning strategy (enabled via
+            ``FLASHINFER_AUTOTUNE_DISTRIBUTED=broadcast``) to scope the
+            cross-rank tactic broadcast.  Pass the group whose ranks shard the
+            work and must agree on the tactic -- e.g. the MoE expert-parallel
+            group.  ``None`` (default) uses the WORLD group, which is correct
+            when WORLD is a single EP group, but in multi-dimensional layouts
+            (e.g. EP8 + DP2 over 16 ranks) you should pass the EP group so the
+            broadcast does not cross independent EP-group boundaries.  No effect
+            unless the broadcast strategy is enabled.
 
     Raises:
         ValueError: If ``tuning_buckets`` is provided but empty.
@@ -850,6 +862,24 @@ def autotune(
 
         if autotune_enabled:
             logger.info("[Autotuner]: Autotuning process ends")
+            # Align tactics across ranks once the outermost tuning
+            # context closes, so every rank leaves with the same per-key
+            # decision.  No-op unless FLASHINFER_AUTOTUNE_DISTRIBUTED is set.
+            # ``distributed_process_group`` scopes the broadcast to the ranks
+            # that actually shard the work (e.g. the MoE expert-parallel group);
+            # ``None`` falls back to the default WORLD group.
+            #
+            # An exception may be propagating out of the tuning body on *some*
+            # ranks but not others. The broadcast is a collective, so the
+            # decision to run it must be identical on every rank -- otherwise a
+            # rank that raised would skip it while a rank that finished normally
+            # would block forever inside ``broadcast_object_list``. We therefore
+            # hand the local exception state to the sync, which reaches a
+            # collective agreement (all ranks broadcast, or all skip).
+            tuner._maybe_sync_distributed_cache(
+                process_group=distributed_process_group,
+                local_had_exception=sys.exc_info()[0] is not None,
+            )
 
         # Save configs on exit when tuning with a cache path, but only if
         # new profiling results were added this session.  save_configs
@@ -1088,9 +1118,24 @@ class AutoTuner:
     def __init__(
         self, warmup: int = 3, repeat: int = 10, stream_delay_micro_secs: int = 5000
     ):
-        self.repeat = repeat
-        self.warmup = warmup
+        # Allow env overrides so profiling fidelity (back-to-back iters used to
+        # rank tactics) can be tuned without code changes.  A too-small
+        # ``repeat`` can under-measure low-occupancy tactics that only show
+        # their true cost in steady state, making the autotuner pick a tactic
+        # that is slower than the heuristic default in real (CUDA-graph) decode.
+        self.repeat = int(os.environ.get("FLASHINFER_AUTOTUNE_REPEAT", repeat))
+        self.warmup = int(os.environ.get("FLASHINFER_AUTOTUNE_WARMUP", warmup))
         self.stream_delay_micro_secs = stream_delay_micro_secs
+        # Relative speed-up an explicit tactic must beat the op's built-in
+        # heuristic-default (tactic=-1) by before the autotuner switches away
+        # from that default.  This guarantees the tuned choice is *never slower
+        # than default* (the autotuner keeps the default unless a tactic is
+        # genuinely faster) and removes run-to-run / cross-rank flapping among
+        # near-tied tactics.  Override via FLASHINFER_AUTOTUNE_SWITCH_MARGIN
+        # (e.g. "0.0" to restore pure fastest-wins behavior).
+        self.switch_margin = max(
+            0.0, float(os.environ.get("FLASHINFER_AUTOTUNE_SWITCH_MARGIN", "0.03"))
+        )
         self.profiling_cache: dict[
             ProfilingCacheKey, tuple[Any, OptimizationProfile | None]
         ] = {}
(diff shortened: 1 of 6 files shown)


flashinfer/fused_moe/core.py
```diff
@@ -1190,26 +1184,54 @@ def __init__(
...
+                        local_expert_offset,
+                        local_expert_offset + max(num_local_experts, 1),
+                        shapes,
+                        dtype=torch.int32,
+                        device=device,
+                    )
+                else:
+                    expert_ids = make_balanced_local_topk_ids(
+                        num_tokens=math.prod(shapes[:-1]),
+                        top_k=shapes[-1],
```
jiahanc, line 1280: When routing_logits is provided, trtllm moe  creates topk_ids as an empty but non-None tensor. Consequently, the moe_inputs.topk_ids is not None check registers _init_packed_topk_ids for this placeholder.

During autotuning, a token profile such as M=256 expands the placeholder from (0,) to (256,). The new initializer then interprets this one-dimensional shape as num_tokens=1, top_k=256, causing make_balanced_local_topk_ids to raise when there are 128 local experts—even though the actual runtime top_k is 8.

flashinfer/autotuner.py
```diff
@@ -672,6 +684,23 @@ def autotune(
...
+            # that actually shard the work (e.g. the MoE expert-parallel group);
+            # ``None`` falls back to the default WORLD group.
+            #
+            # Skip the collective while an exception is propagating: if one rank
+            # raised inside the tuning body it would still reach this ``finally``
+            # and block in ``dist.broadcast_object_list`` while the other ranks
+            # are elsewhere, deadlocking the whole group. A hung collective is
+            # not an exception, so the try/except inside the sync cannot recover
+            # from it -- we must simply not enter it during unwind.
+            if sys.exc_info()[0] is None:
```
jiahanc: Correct me if I am wrong, this seems doesnt fix the deadlock. `sys.exc_info()` only reflects the current rank. If rank 0 raises, it skips the broadcast, while a rank that exits the tuning body normally still enters broadcast_object_list and waits for rank 0.

flashinfer/fused_moe/core.py
qiching: thanks. works!

flashinfer/autotuner.py
qiching: thanks. updated.

jiahanc (approved):
LGTM, thanks for the contribution!

flashinfer/autotuner.py
```diff
@@ -672,6 +683,15 @@ def autotune(
...
             logger.info("[Autotuner]: Autotuning process ends")
+            # Align tactics across ranks once the outermost tuning
+            # context closes, so every rank leaves with the same per-key
+            # decision.  No-op unless FLASHINFER_AUTOTUNE_DISTRIBUTED is set.
+            # ``distributed_process_group`` scopes the broadcast to the ranks
+            # that actually shard the work (e.g. the MoE expert-parallel group);
+            # ``None`` falls back to the default WORLD group.
+            tuner._maybe_sync_distributed_cache(
+                process_group=distributed_process_group
+            )
```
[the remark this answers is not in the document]
jdebache: Isn't the guarantee of `finally` that the program always goes there, no matter whether an exception is thrown or not?

flashinfer/autotuner/autotuner.py
```diff
@@ -813,9 +843,24 @@ class AutoTuner:
...
 
     def __init__(self, warmup=3, repeat=10, stream_delay_micro_secs=5000):
-        self.repeat = repeat
-        self.warmup = warmup
+        # Allow env overrides so profiling fidelity (back-to-back iters used to
+        # rank tactics) can be tuned without code changes.  A too-small
+        # ``repeat`` can under-measure low-occupancy tactics that only show
+        # their true cost in steady state, making the autotuner pick a tactic
+        # that is slower than the heuristic default in real (CUDA-graph) decode.
+        self.repeat = int(os.environ.get("FLASHINFER_AUTOTUNE_REPEAT", repeat))
```
jdebache, line 1130: I don't get the presence of repeat as state to the autotuner singleton. As the comment says, number of iterations should depend on what operation / kernel we are measuring and should be set by the caller?

flashinfer/autotuner/autotuner.py
```diff
@@ -813,9 +843,24 @@ class AutoTuner:
...
+        self.warmup = int(os.environ.get("FLASHINFER_AUTOTUNE_WARMUP", warmup))
```
jdebache, line 1131: Same comment as for repeat.

flashinfer/autotuner/autotuner.py
```diff
@@ -813,9 +843,24 @@ class AutoTuner:
...
         self.stream_delay_micro_secs = stream_delay_micro_secs
+        # Relative speed-up an explicit tactic must beat the op's built-in
+        # heuristic-default (tactic=-1) by before the autotuner switches away
+        # from that default.  This guarantees the tuned choice is *never slower
```
jdebache: "guarantees" is too strong. It makes it less likely that heuristic beats tuned, but doesn't provide any guarantee I think. This is all empirical.

flashinfer/autotuner/autotuner.py
```diff
@@ -813,9 +843,24 @@ class AutoTuner:
...
+        # than default* (the autotuner keeps the default unless a tactic is
+        # genuinely faster) and removes run-to-run / cross-rank flapping among
+        # near-tied tactics.  Override via FLASHINFER_AUTOTUNE_SWITCH_MARGIN
+        # (e.g. "0.0" to restore pure fastest-wins behavior).
+        self.switch_margin = max(
+            0.0, float(os.environ.get("FLASHINFER_AUTOTUNE_SWITCH_MARGIN", "0.03"))
```
jdebache, line 1143: Does it have to be an environment variable?

Also, same reasoning as for "warmup" and "repeat": I don't think it should be set globally.

flashinfer/autotuner/autotuner.py
```diff
@@ -1280,10 +1325,42 @@ def choose_one(
...
                         skipped_count = 0
+
+                        # Measure the op's built-in heuristic default (tactic=-1)
+                        # on the fallback runner (runners[0]) up-front and use it
+                        # as the baseline.  An explicit tactic is only adopted when
+                        # it beats this baseline by ``switch_margin`` (decision made
+                        # after the loop) -- this guarantees the tuned choice is
+                        # *never slower than default* and removes run-to-run /
+                        # cross-rank flapping among near-tied tactics.
+                        default_time = float("inf")
```
jdebache, line 1668: If by "inf" we mean "we don't have a measurement yet", then use "None" instead.

flashinfer/autotuner/autotuner.py
```diff
@@ -1280,10 +1325,42 @@ def choose_one(
...
+                                torch.cuda.synchronize()
+                            with contextlib.suppress(Exception):
+                                torch.cuda.cudart().cudaGetLastError()
+                            logger.debug(
+                                f"[Autotuner]: default tactic (-1) baseline "
+                                f"profiling failed for {custom_op}: {e}"
+                            )
+
+                        # Best explicit (runner, tactic) candidate found so far.
+                        best_time = float("inf")
```
jdebache, line 1691: Same here: no information is better represented by "None" than by "inf".

flashinfer/autotuner/autotuner.py
```diff
@@ -1280,10 +1325,42 @@ def choose_one(
...
+                        best_runner_id, best_tactic = None, None
```
jdebache, line 1692: You could introduce a small type to wrap these three pieces of information, as they will always be set / not set together. Either we have a best runner / tactic / time, or we don't.

flashinfer/autotuner/autotuner.py
```diff
@@ -1404,6 +1523,110 @@ def _get_input_sizes(self, inputs: List[torch.Tensor]) -> List[torch.Size]:
 
         return sizes
 
+    def _maybe_sync_distributed_cache(
+        self, process_group: Any = None, local_had_exception: bool = False
```
jdebache: Can we type `process_group`?

flashinfer/autotuner/autotuner.py
```diff
@@ -1404,6 +1523,110 @@ def _get_input_sizes(self, inputs: List[torch.Tensor]) -> List[torch.Size]:
...
+                peer that finished normally blocks in it forever.  All ranks
+                first exchange this flag via a small ``all_gather_object`` (a
+                collective every rank always reaches here), and the broadcast is
+                run only when *no* rank in the group failed.
+
+        Only the (runner_id, tactic) decision is communicated -- never the
+        ``OptimizationProfile`` object -- so there is nothing fragile to
+        pickle, and each rank keeps its own (shape-identical) profile.
+        """
+        mode = os.environ.get("FLASHINFER_AUTOTUNE_DISTRIBUTED", "").lower()
```
jdebache, line 1919: You could try creating a `DistributedAutotuningStrategy` union type that is either `None`, or `Broadcast`, which is a type containing simply a process group. Then switch on it. Then leave it to the caller to pass `None` or some strategy type. No need to rely on environment variables.

flashinfer/autotuner/autotuner.py
```diff
@@ -1404,6 +1523,110 @@ def _get_input_sizes(self, inputs: List[torch.Tensor]) -> List[torch.Size]:
...
+
+        # Collectively agree on whether to broadcast. Every rank reaches this
+        # ``all_gather_object`` unconditionally (so it cannot deadlock on its
+        # own), then we broadcast only if no rank is unwinding an exception.
+        # This keeps the participate/skip decision identical across ranks even
+        # when a failure is asymmetric (some ranks raised, others did not).
+        try:
+            world_size = dist.get_world_size(group=process_group)
+            exc_flags: List[Any] = [None] * world_size
+            dist.all_gather_object(
```
jdebache, line 1939: So all ranks perform the tuning, but we only keep rank 0's result? Wouldn't it be better to either:
- Only have rank 0 do the tuning ?
- Have the ranks "vote" on which tactic is best? What if rank 0 is the odd-one-out, that for some reason found a tactic to be better, while all it's peers agree on another one?

flashinfer/autotuner/autotuner.py
qiching: updated

flashinfer/autotuner/autotuner.py
qiching: updated

flashinfer/autotuner/autotuner.py
aleozlx, line 1491: (can be separate PR: document exception handling expectation - autotuning crashing handling)

aleozlx:
prefer to have (at least) a one-liner/ bullet points PR description

aleozlx:
please link releated issue

aleozlx:
reviewed in a meeting. sounds good overall
but there seems to be a few more reviewer comments to address

PR #3591, opened 2026-06-11
test(moe): align EP-offset tests and docs with global expert ids (gh #3547)

### LINKED ISSUE #3547: [Bug] Unified MoE API (#3093): trtllm_fp4_routed returns all-zeros for expert-parallel shards (local_expert_offset > 0)
**Affected:** unified MoE API, PR #3093 (`moe_api` branch), `flashinfer/fused_moe/runners.py`. SM100, NVFP4. `cute_dsl_nvfp4` is **not** affected.

### Summary
Driving `MoELayer` with an expert-parallel shard — `num_experts` (global) > `local_num_experts`, `local_expert_offset > 0`, weights prepared for the local shard, `selected_experts` passed as **global** ids (the contract `test_pack_inputs_applies_local_expert_offset` already encodes) — `trtllm_fp4_routed` produces **bit-exactly zero** output (`(out == 0).all()` is True). `cute_dsl_nvfp4` on the identical config is correct. EP is the standard large-MoE deployment mode, so this is a silent total wrong-answer on a common path.

### Repro / isolating diagnostic (H=I=512, T=128, top_k=4, SM100)
Cosine vs an independent fp32 reference, sweeping `(global, local, offset)`:
```
global=32  local=32 offset=0  : cute_dsl 0.9995   trtllm 0.9995   ok  (no EP)
global=64  local=32 offset=0  : cute_dsl 0.9995   trtllm 0.9995   ok  (global>local, offset 0)
global=64  local=32 offset=32 : cute_dsl 0.9995   trtllm 0.0000   BAD (all-zero)
global=128 local=32 offset=64 : cute_dsl 0.9995   trtllm 0.0000   BAD (all-zero)
```
The trigger is specifically **`local_expert_offset > 0`** — `global > local` at offset 0 is fine.

### Likely root cause — offset applied twice
`TrtllmFp4RoutedRunner` subtracts the offset in `pack_inputs` (`runners.py:257`: `ids = act.selected_experts - self._local_expert_offset`, producing local ids in `[0, local)`) **and also** forwards `local_expert_offset=self._local_expert_offset` to the kernel (`runners.py:305`). If the kernel subtracts it again, the ids fall outside `[0, local_num_experts)`, no expert matches, and the finalize writes nothing → all-zero. `cute_dsl_nvfp4` does **not** pre-subtract (it passes `selected_experts` as-is) and applies the offset once, so it is correct.

### Why CI missed it
`test_unified_moe.py::test_pack_inputs_applies_local_expert_offset` checks only the *packing* in isolation (offset ∈ {0, 32, 96}); there is no full EP-forward accuracy test, so the double-application in the kernel-call path is never exercised.

### Suggested fix
Apply the offset exactly once — either keep the `pack_inputs` subtraction and pass `local_expert_offset=0` to the kernel call, or stop pre-subtracting in `pack_inputs` and let the kernel own it (matching `cute_dsl`). Add a full EP-forward accuracy test.

(Found by a unified-API MoE fuzzer; file:line refs at PR #3093 head `89466b1a`.)

waynehacking8:
## Description

**Rebased 2026-07-02 — scope reduced to tests + docs.** The EP-offset double-subtraction this PR originally fixed was fixed on main by #3686 (`pack_inputs` now packs GLOBAL expert ids; the kernel owns the global→local mapping via `mLocalExpertsStartIdx` in `include/flashinfer/trtllm/fused_moe/RoutingKernel.h`). What #3686 did **not** update:

- `tests/moe/test_unified_moe.py::TestTrtllmEPOffset::test_pack_inputs_applies_local_expert_offset` still asserts the **old** pre-subtracted packing (`decoded == selected_experts - local_expert_offset`), which now contradicts `pack_inputs` — for `local_expert_offset > 0` it fails wherever it runs (it is `sm100_required`).
- `tests/moe/test_unified_moe_fuzz.py` still carries the gh #3547 `_KNOWN_FAILURES` xfail entry for a now-fixed bug (per the ledger's own design, a fixed bug's entry must be removed) plus stale "pending gh #3547" CI-gate comments.
- The `TrtllmFp4RoutedRunner` class/`pack_inputs` docstrings still describe the pre-subtracted packing.

This PR aligns all of the above with the merged contract and adds an EP-shard forward regression test (`test_ep_shard_forward_matches_offset_zero`): `offset > 0` output must match the bit-identical `offset = 0` run and must not be all-zero (the gh #3547 symptom), so the double-subtraction can't silently come back.

Fixes #3547 (the code fix landed in #3686; this closes the loop with regression coverage).

## Related Issues

gh #3547 · supersedes-note: code fix landed via #3686

## Verification

- `ruff check` + `ruff format --check` clean on the three touched files
- `pytest tests/moe/test_unified_moe.py` on RTX PRO 6000 (SM120, CUDA 13): **99 passed, 11 skipped** (the `sm100_required` tests, including the ones this PR edits, skip on SM120 — they need SM100 CI to execute)
- `pytest --collect-only` on both test files: 196 collected, no import errors

AI-assisted (Claude Code).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed TRTLLM FP4 routed MoE all-zero outputs for nonzero expert offsets by keeping global expert IDs in packed top-k IDs and applying expert offset remapping during kernel execution.

* **Tests**
  * Updated input-packing assertions to verify global expert ID preservation and correct shard-range mapping.
  * Added a GPU correctness regression test comparing nonzero-offset results against the zero-offset baseline.
  * Removed the prior expected-failure entry for this configuration from the fuzz/ledger logic.

* **Documentation**
  * Clarified routed MoE packed top-k ID semantics and kernel-side offset behavior.

### DIFF
--- a/tests/moe/test_unified_moe.py
+++ b/tests/moe/test_unified_moe.py
@@ -886,6 +886,9 @@ class TestTrtllmEPOffset:
             + local_expert_offset
         )
         final_scales = torch.rand(num_tokens, top_k, device=device)
+        # One negative weight makes the low-16-bit check below sensitive to a
+        # dropped & 0xFFFF mask (bf16 sign bit would smear into the id field).
+        final_scales[0, 0] = -final_scales[0, 0]
         act_pack = MoEActivationPack(
             hidden_states_q=torch.zeros(
                 num_tokens, hidden_size // 2, dtype=torch.uint8, device=device
@@ -925,7 +928,13 @@ class TestTrtllmEPOffset:
         # Global ids land inside this rank's shard; the kernel maps them to
         # [0, local_num_experts) by subtracting local_expert_offset.
         assert int(decoded_ids.min()) >= local_expert_offset
-        assert int(decoded_ids.max()) < local_expert_offset + local_num_experts
+        # Low 16 bits hold the bf16 gate-weight bits.
+        expected_bits = (
+            final_scales.to(torch.bfloat16).view(torch.int16).to(torch.int32) & 0xFFFF
+        )
+        assert torch.equal(topk_ids & 0xFFFF, expected_bits)
+        # The offset travels to the kernel as a separate argument.
+        assert runner._static_kwargs["local_expert_offset"] == local_expert_offset
 
     @pytest.mark.parametrize("local_expert_offset", [32, 96])
     def test_ep_shard_forward_matches_offset_zero(self, local_expert_offset):
--- a/tests/moe/test_unified_moe.py
+++ b/tests/moe/test_unified_moe.py
@@ -847,17 +847,18 @@ class TestUnifiedMoEDispatch:
 
 @sm100_required
 class TestTrtllmEPOffset:
-    """TRTLLM routed packing must shift global expert ids down by the local
-    shard offset.
+    """TRTLLM routed packing must keep expert ids GLOBAL.
 
-    The packed int32 top-k id is ``((expert_id - offset) << 16) | bf16(weight)``;
-    the kernel indexes local experts as ``[0, local_num_experts)``.  Before the
-    CR3 fix, ``pack_inputs`` defaulted the offset to 0, so a nonzero local shard
-    produced out-of-range local expert ids.
+    The packed int32 top-k id is ``(global_expert_id << 16) | bf16(weight)``;
+    the kernel performs the global→local mapping itself by subtracting
+    ``local_expert_offset`` (``mLocalExpertsStartIdx`` in RoutingKernel.h).
+    Pre-subtracting the offset in ``pack_inputs`` applied it twice, pushing
+    every id below the local shard and silently zeroing the output for any
+    ``local_expert_offset > 0`` (gh #3547).
     """
 
     @pytest.mark.parametrize("local_expert_offset", [0, 32, 96])
-    def test_pack_inputs_applies_local_expert_offset(self, local_expert_offset):
+    def test_pack_inputs_preserves_global_expert_ids(self, local_expert_offset):
         device = torch.device("cuda", torch.cuda.current_device())
         num_experts = 128
         local_num_experts = 32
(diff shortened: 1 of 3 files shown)


tests/moe/test_unified_moe_fuzz.py
```diff
@@ -66,7 +66,7 @@
 Coverage today: NVFP4 (CuteDSL + TRTLLM-FP4-routed) on SM100 -- the only wired MVP runners.
 
 OPT-IN: this suite is gated behind FLASHINFER_UMOE_FUZZ (see the pytestmark below) and is
-SKIPPED unless that env var is set -- waived in CI pending gh #3547 and root-cause of a
+SKIPPED unless that env var is set -- waived in CI pending root-cause of a
 whole-process device-side-assert abort that would block B200 CI. Run it explicitly:
```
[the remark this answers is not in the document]
waynehacking8: Done in 941df6e — aligned all remaining `#3547` strings in the CI-gate block (lines ~174-188), the skip reason, and the roadmap credit line. The opt-in gate itself stays (the whole-process abort root-cause is still outstanding).

waynehacking8:
Rebased onto current main (`8fc7f079`) — now MERGEABLE. Since #3686 landed the pack-global-ids fix itself, this PR's scope is reduced to what #3686 left stale: `TestTrtllmEPOffset` (still asserts the old pre-subtracted packing, so it now fails for `offset > 0` on SM100), the fuzzer's gh #3547 `_KNOWN_FAILURES` entry + CI-gate comments, and the runner docstrings — plus a new EP-shard forward regression test so the double-subtraction can't silently return. Title/description updated to match. Verified locally: ruff clean; `test_unified_moe.py` 99 passed / 11 skipped on SM120 (the `sm100_required` tests need SM100 CI to execute).

waynehacking8:
Note: this also fixes #3823 (the b300 test_pack_inputs_applies_local_expert_offset assertion) - same root cause, and #3826 now proposes the same test change independently.

jdebache:
I can see a couple of failures in CI, but please don't worry about them / spend time trying to fix them. They are very likely to be unrelated to your changes. I will seek advice internally on how to best resolve them.

waynehacking8:
Thanks for the update - no rush. If any of them do turn out to touch the moe test paths this PR rewrites, happy to take a look.

mhoqueanik:
Codes look correct. Failing on CI. I see three common issues-

1. Shared memory limit exceeded
```
python/tvm_ffi/cython/function.pxi:929: RuntimeError: Error in function 'BatchPrefillWithRaggedKVCacheDispatched' at /tmp/flashinfer/include/flashinfer/attention/prefill.cuh:4070: Required shared memory (132176 bytes) for head_dim_qk=512, head_dim_vo=512, cta_tile_q=32 exceeds this GPU's per-block limit (101376 bytes); this configuration is not supported on this architecture.
E   RuntimeError: Error in function 'BatchPrefillWithRaggedKVCacheDispatched' at /tmp/flashinfer/include/flashinfer/attention/prefill.cuh:4070: Required shared memory (132176 bytes) for head_dim_qk=512, head_dim_vo=512, cta_tile_q=32 exceeds this GPU's per-block limit (101376 bytes); this configuration is not supported on this architecture.
``` 
2. Issue in cutedsl (a new version was released a few days ago, it might be incompatible with current code)
```
/opt/conda/envs/py312/lib/python3.12/site-packages/nvidia_cutlass_dsl/dsl_packages/cutlass/base_dsl/runtime/cuda.py:360: cutlass.base_dsl.common.DSLCudaRuntimeError:
E   cutlass.base_dsl.common.DSLCudaRuntimeError: 
    error: cudaErrorUnsupportedPtxVersion (error code: 222) 
``` 
3. Script taking too long
```
============================= test session starts ==============================
platform linux -- Python 3.12.13, pytest-9.0.3, pluggy-1.6.0
rootdir: /tmp/flashinfer
configfile: pytest.ini
collected 160 items
[2026-07-05T15:33:26.007] error: *** STEP 1468220.3 ON 2u1g-b650-1121 CANCELLED AT 2026-07-05T15:33:26 DUE TO TIME LIMIT ***
```

kahyunnam (approved):
LGTM

tests/moe/test_unified_moe.py
```diff
@@ -915,13 +916,84 @@ def test_pack_inputs_applies_local_expert_offset(self, local_expert_offset):
...
+        # Upper 16 bits hold the GLOBAL expert id, exactly as selected.
+        decoded_ids = topk_ids >> 16
+        assert torch.equal(decoded_ids, selected_experts), (
+            f"offset={local_expert_offset}: packed ids "
+            f"{decoded_ids} != global ids {selected_experts}"
+        )
+        # Global ids land inside this rank's shard; the kernel maps them to
+        # [0, local_num_experts) by subtracting local_expert_offset.
+        assert int(decoded_ids.min()) >= local_expert_offset
+        assert int(decoded_ids.max()) < local_expert_offset + local_num_experts
```
feih-nv: Does this make sense?
> the test currently asserts only the upper 16 bits, a regression in the low bits (e.g. a dropped & 0xFFFF mask letting a negative bf16 sign bit smear into the expert-id field) would pass while silently corrupting per-expert scales; and packing global ids but dropping the separate local_expert_offset kwarg would also pass while reproducing the #3547 symptom.

```suggestion
        assert int(decoded_ids.max()) < local_expert_offset + local_num_experts
        # Low 16 bits hold the bf16 gate-weight bits.
        expected_bits = (
            final_scales.to(torch.bfloat16).view(torch.int16).to(torch.int32) & 0xFFFF
        )
        assert torch.equal(topk_ids & 0xFFFF, expected_bits)
        # The offset travels to the kernel as a separate argument.
        assert runner._static_kwargs["local_expert_offset"] == local_expert_offset
```

waynehacking8: Makes sense - applied in e3fbfd10, plus one flipped sign in final_scales: with all-positive rand() weights a dropped mask is invisible to both asserts (bit-replay shows the smear lands in the id field, so the existing global-id assert catches it once a sign bit exists), while your low-16 assert pins the dtype/packing-order half. The static-kwarg assert checks out against runners.py (populated in pack_inputs before the test reads it). I can't execute this class locally (sm100-gated, I'm on SM120), so this is verified by source + a standalone bit replay - CI will exercise it.

aleozlx (approved):

PR #3608, opened 2026-06-12
fix: pass XQA NVFP4 scale-factor strides

Njuapp:
## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * Extended MHA and FlashInfer launches to pass through additional 4-bit KV-cache stride parameters for SF (“scaling factor”) layouts, improving correct SF-based K/V loading.
* **Bug Fixes**
  * Added runtime validation to reject unsupported NVFP4 KV cache page sizes (16–32), matching documented support.
* **Documentation**
  * Clarified that NVFP4 KV cache page sizes 64–128 are supported; 16–32 are not.
* **Tests**
  * Expanded NVFP4 batch decode coverage with additional SF layouts/stride scenarios.

### DIFF
--- a/csrc/xqa/mha.cu
+++ b/csrc/xqa/mha.cu
@@ -2911,8 +2911,7 @@ void launchMHA(
   uint32_t const stride_token_in_heads = static_cast<uint32_t>(kv_stride_token / validElemsPerHead);
   uint32_t const stride_head_in_heads = static_cast<uint32_t>(kv_stride_head / validElemsPerHead);
 #if ENABLE_4BIT_KV_CACHE
-  uint32_t const sf_elems_per_head =
-      validElemsPerHead / CacheElemConverter::QuantVectorSize;
+  uint32_t const sf_elems_per_head = validElemsPerHead / CacheElemConverter::QuantVectorSize;
   uint32_t const sf_stride_page_in_heads =
       static_cast<uint32_t>(sf_stride_page / sf_elems_per_head);
   uint32_t const sf_stride_token_in_heads =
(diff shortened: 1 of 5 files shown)


samuellees:
cc @Tom-Zheng @qsang-nv

qsang-nv:
Thanks for the fix — decoupling the SF strides from the data-cache strides is the right call. A few things to address before this lands:

**1. CI is red — please run pre-commit.**
Both `pre-commit` (clang-format on `csrc/xqa/xqa_wrapper.cu` and `csrc/xqa/mha.h`) and the test job are failing. `pre-commit run --all-files` should clear the formatting hunks.

**2. The page-size guard is on the wrong entry point.**
The new `page_size in (16, 32)` check lives in `trtllm_batch_decode_with_kv_cache` (xqa branch), but the canonical XQA entry `xqa_batch_decode_with_kv_cache` has no such guard. That's the function the existing NVFP4 test calls, and the one users hit directly — calling it with NVFP4 + page_size 16/32 still reaches the kernel and silently produces wrong results. Please move/duplicate the check into `xqa_batch_decode_with_kv_cache` (or a shared helper) so both paths are protected.

**3. The fix isn't actually exercised by any test.**
The only NVFP4 XQA test (`tests/attention/test_xqa_batch_decode.py::test_xqa_batch_decode_nvfp4_kv`) is SM120/121-gated (CI runners skip it), and — more importantly — it builds both the data cache and the SF cache with the same `torch.stack([k, v], dim=1)` layout. After conversion to head units the SF strides end up equal to the data-cache strides, so the test passes identically with and without this patch; it cannot catch the bug being fixed. Please add a case where the SF and data layouts diverge (e.g. separately-allocated/contiguous SF, or page_size=128), and post SM120/SM121 local results since CI cannot run this path.

qsang-nv (approved):

PR #3616, opened 2026-06-12
feat: improve gated delta rule benchmark script

guangyunh-nv:
## 📌 Description

1. Add more benchmark cases, long single seq and large batch are now covered.
2. Use cuda graph because cutedsl and fla both have large launch overhead.
3. Add cooling time due internal machine overheat.
4. Use rotation buffer to avoid L2 caching effect.

### ✅ Pre-commit Checks

## 🧪 Tests

## Benchmark Improvements

* **Chores**
  * Enhanced the benchmark harness with improved timing and memory management capabilities.
  * Added new command-line options including cooling time intervals, CUPTI profiling, and CUDA graph support.
  * Updated benchmark configurations targeting larger sequence lengths and new performance patterns.

### DIFF
--- a/benchmarks/bench_gdn_prefill.py
+++ b/benchmarks/bench_gdn_prefill.py
@@ -27,6 +27,7 @@ Usage:
 
 import argparse
 import sys
+import time
 
 import numpy as np
 import torch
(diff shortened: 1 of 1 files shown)


jiahanc (approved):
LGTM

PR #3644, opened 2026-06-15
fix(moe): return bf16 expert_weights from trtllm_fp4_block_scale_moe (#3595)

### LINKED ISSUE #3595: trtllm_fp4_block_scale_moe allocates FP32 topk_weights for DeepSeekV3 despite BF16 CUDA routing output
## Summary

In `trtllm_fp4_block_scale_moe`, the Python wrapper allocates `topk_weights` using `routing_logits.dtype` when routing logits are provided. For FP32 routing logits, this creates an FP32  `topk_weights` buffer.

However, the C++ DeepSeekV3 routing path hard-codes the routing weight output dtype to BF16:

- Python allocation: `flashinfer/fused_moe/core.py`
- DeepSeekV3 routing output: `csrc/trtllm_fused_moe_runner.cu`

This creates a dtype/memory interpretation mismatch for:

```python
trtllm_fp4_block_scale_moe(
    routing_logits=fp32_logits,
    routing_method_type=RoutingMethodType.DeepSeekV3,
    ...
)
```

The issue is especially visible with `do_finalize=False`, where the returned expert_weights tensor can have FP32 metadata even though the routing kernel writes BF16 values.

## Potential Solution

For `routing_method_type == RoutingMethodType.DeepSeekV3`:

- Allocate internal topk_weights as `torch.bfloat16`, regardless of `routing_logits.dtype`.
- Validate user-provided unpacked `topk_weights` is BF16.
- Document that DeepSeekV3 expert_weights are BF16, including for `do_finalize=False`.

This matches the current C++ kernel behavior and the documented unpacked precomputed routing contract.

This is minimally invasive but changes the visible `dtype` of returned `expert_weights` for `do_finalize=False` from FP32 to BF16 in the FP32-logits + DeepSeekV3 case. That said, the current FP32 tensor appears to be incorrectly typed because the CUDA writer emits BF16 data.

waynehacking8:
## Description

Fixes #3595.

`trtllm_fp4_block_scale_moe` allocates the `expert_weights` output buffer with
`routing_logits.dtype` and **returns that buffer verbatim** when
`do_finalize=False`:

```python
# do_finalize=False return path
return [
    torch.from_dlpack(intermediate_output[0]),
    topk_weights,                                # <-- python-allocated buffer
    torch.from_dlpack(intermediate_output[2]),
]
```

But the trtllm-gen routing kernel **always writes the expert weights in
bfloat16**, for *every* routing method — `routingData.mDtypeOutput` is
hard-set to `Bfloat16` in each branch of `Runner::run` in
`csrc/trtllm_fused_moe_runner.cu` (e.g. for `DeepSeekV3`:

```cpp
routingData.mDtypeOutput =
    btg::Dtype::Bfloat16;               // for DeepSeek, the expW is currently always bfloat16
routingData.mDtypeInput = dtypeLogits;  // routing logits can be bfloat16 or fp32
```

and `routingData.mPtrTopKWeights = expertWeights;`).

So when the routing logits are fp32 — the conventional **DeepSeekV3** case —
the returned `expert_weights` is an **fp32 tensor over bf16 data**: the caller
reinterprets bf16 bytes as fp32 and reads garbage.

## Fix

Allocate the output buffer as `bfloat16` to match the kernel's actual output
dtype, independent of `routing_logits.dtype`.

This is a **no-op** for:
- the common **bf16 routing-logits** path (`routing_logits.dtype` was already bf16), and
- **`do_finalize=True`** (the buffer is internal workspace; the C++ side tracks
  its true bf16 dtype and an over-sized fp32 buffer just wastes bytes without
  corrupting anything),

so existing tests are unaffected. The bug only surfaced on the
`do_finalize=False` + fp32-logits path, which had no coverage.

### Scope

I verified the other trtllm-gen fused-MoE ops (`trtllm_fp8_per_tensor_scale_moe`,
`trtllm_fp8_block_scale_moe`) and they are **not** affected: their
`do_finalize=False` path returns `torch.from_dlpack(intermediate_output[...])`
views, which carry the kernel's true bf16 dtype rather than the python-allocated
buffer. So the fix is correctly scoped to the FP4 op, exactly as the issue
describes.

## Test

Added `test_fp4_block_scale_deepseekv3_unfinalized_weight_dtype`: runs the FP4
op with fp32 DeepSeekV3 routing logits and `do_finalize=False`, asserting the
returned `expert_weights` is `bfloat16`. (trtllm-gen kernels are SM100-class, so
this runs on the H100 CI; it is skipped on other architectures by the existing
gating.) The public-API `Returns` docstring now documents the bf16 contract.

---

 AI-assisted (Claude). Reviewed and verified by the author.

## Bug Fixes
- Ensured MoE fused kernel routing always emits **expert weights as bfloat16**, including when `do_finalize=False` and routing logits are not bfloat16.
- Added input validation for precomputed routing to require **top‑k weights are bfloat16**, preventing fp32 from being misinterpreted.

## Tests
- Added a regression test for **DeepSeekV3 (fp32 routing logits)** to confirm unfinalized expert weights are **bfloat16**.
- Extended the MoE test harness with an optional check to verify this behavior.

### DIFF
--- a/flashinfer/fused_moe/core.py
+++ b/flashinfer/fused_moe/core.py
@@ -2289,9 +2289,14 @@ def get_trtllm_moe_sm100_module():
                 "either topk_ids or routing_logits must be provided."
             )
             assert topk_ids.dtype == torch.int32, "topk_ids must be an int32 tensor."
-            routing_dtype = torch.bfloat16
-        else:
-            routing_dtype = routing_logits.dtype
+        # The trtllm-gen routing kernel always emits expert weights as bfloat16
+        # (routingData.mDtypeOutput is hard-set to Bfloat16 for every routing
+        # method in csrc/trtllm_fused_moe_runner.cu), independent of the
+        # routing_logits dtype. This buffer is returned verbatim to the caller
+        # when do_finalize=False, so it must be bfloat16 regardless of
+        # routing_logits.dtype (e.g. fp32 DeepSeekV3 logits); otherwise the
+        # returned expert_weights mislabels bf16 data as fp32. See #3595.
+        routing_dtype = torch.bfloat16
         hidden_size = hidden_states.shape[-1]
         if hidden_states.dtype == torch.uint8:
             hidden_size = hidden_size * 2
@@ -3901,6 +3906,10 @@ def trtllm_fp4_block_scale_moe(
     List[torch.Tensor]
         ``[output]`` when ``do_finalize`` is ``True``, otherwise
         ``[gemm2_output, expert_weights, expanded_idx_to_permuted_idx]``.
+        The ``expert_weights`` tensor is always ``bfloat16`` (the routing
+        kernel emits bf16 weights for every routing method), regardless of
+        the ``routing_logits`` dtype — including the ``do_finalize=False``
+        path and fp32 ``DeepSeekV3`` logits.
     """
     _validate_routing_replay_out(routing_replay_out, top_k)
     return get_trtllm_moe_sm100_module().trtllm_fp4_block_scale_moe(
--- a/flashinfer/fused_moe/core.py
+++ b/flashinfer/fused_moe/core.py
@@ -2311,6 +2311,14 @@ def get_trtllm_moe_sm100_module():
             assert topk_weights is not None, (
                 "topk_weights must be provided for UnpackedPrecomputed mode"
             )
+            # The finalize kernel reads the expert weights as args.mDtypeExpW,
+            # which is bfloat16 for this op (see runner.h: expert_weights is
+            # "[num_tokens, top_k] in bfloat16 = mDtypeExpW"). A user-provided
+            # fp32 buffer would be reinterpreted as bf16, so reject it up front.
+            assert topk_weights.dtype == torch.bfloat16, (
+                "topk_weights must be bfloat16 for UnpackedPrecomputed mode, got "
+                f"{topk_weights.dtype}"
+            )
         else:
             # For Mode 1 (FromLogits) and Mode 2 (PackedPrecomputed), allocate OUTPUT buffers
             if topk_ids is None:
--- a/tests/moe/test_trtllm_gen_fused_moe.py
+++ b/tests/moe/test_trtllm_gen_fused_moe.py
@@ -2027,3 +2027,43 @@ def test_moe_lora_delta(
 
     torch.testing.assert_close(delta_args_dequant.gemm1_lora_delta, delta)
     assert (delta_reference - zero_reference).abs().max().item() > 0.05
+
+
+def test_fp4_block_scale_deepseekv3_unfinalized_weight_dtype(cache_permute_indices):
+    """Regression for #3595.
+
+    With fp32 DeepSeekV3 routing logits and ``do_finalize=False``, the returned
+    ``expert_weights`` must be bfloat16: the trtllm-gen routing kernel always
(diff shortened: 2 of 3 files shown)


waynehacking8:
Good catch — addressed in 138d92c.

I traced the consumption side to confirm the contract before adding the guard: the finalize kernel reads the expert weights as `args.mDtypeExpW` (`trtllm_fused_moe_runner.cu` → `finalizeData.mDtypeExpW = args.mDtypeExpW`; `dev_kernel.cu` reads `expertWeightsPtr[...]` as `TypeExpW`). `args.mDtypeExpW` is hard-defaulted to `Bfloat16` in `runner.h` and is never reassigned anywhere in `csrc/`, and `runner.h` documents `expert_weights` as *"[num_tokens, top_k] in bfloat16 = mDtypeExpW"*. (The `DevKernel.h` dispatch does compile fp32-expW paths, but this launcher never selects them.)

So in `UnpackedPrecomputed` mode a user-supplied fp32 `topk_weights` would be silently reinterpreted as bf16. The routed `(ids, weights)` tuple API already documents the weights as `bfloat16`, so this just enforces the existing contract up front. Added as an assertion in the same validation block.

waynehacking8:
Rebased onto latest `main` to clear the conflict. The only collision was in `run_moe_test`'s signature: `main` added `check_reference=True` (#3592-era) and this PR added `verify_unfinalized_weight_dtype=False` at the same trailing position — they're independent kwargs (used at the `if check_reference:` and `if verify_unfinalized_weight_dtype:` guards respectively), so I kept both. No logic changed; net diff is unchanged from the prior head.

waynehacking8:
@coderabbitai Those four `match=` patterns (L5205/5213/5226/5234) aren't introduced by this PR — `git blame` attributes them to 897cddd6 (Ruoqian Guo, merged to `main` earlier today); they only appear in the diff because this branch was just rebased onto latest `main`. This PR's own diff adds no `pytest.raises(match=...)` patterns (its only `MxFp8`/`Swiglu` references are enum values, not regex). I'm leaving them untouched to keep the PR scoped to the bf16 `expert_weights` fix; the regex-escaping nit is best handled in the PR/commit that owns those lines.

waynehacking8:
@coderabbitai No need for a tracking issue, thanks. Those four `match=` patterns are pre-existing (897cddd6, not this PR), and the unescaped `.` still matches the literal `.` in the expected strings, so the `pytest.raises` assertions pass correctly — it's a cosmetic robustness nit, not a latent failure. Not worth a standalone issue; best folded into any future cleanup that already touches those lines.

waynehacking8:
Rebased onto current `main` (was conflicting). The `core.py` fix applied cleanly. The regression test needed re-targeting: `run_moe_test` moved to `tests/moe/trtllm_gen_fused_moe_utils.py` and `TUNE_MAX_NUM_TOKENS` was removed upstream since this PR was opened, so the `verify_unfinalized_weight_dtype` path + the DeepSeekV3 fp32 regression test are re-added against the current structure (`tune_max_num_tokens` now defaults to 8192).

Verified locally on current `main`: `py_compile` + pytest collection + `ruff` clean, and the new test skips cleanly on SM120 (trtllm-gen FP4 MoE is unsupported there); the FP4 dtype assertion exercises on CI / datacenter Blackwell. AI-assisted rebase.

aleozlx (approved):

PR #3650, opened 2026-06-16
fix: eliminate NaN output rows in FP8 block-scale grouped MoE GEMM

anish-shanbhag:
## Summary

The TensorRT-LLM-derived FP8 block-scale **grouped** MoE GEMM (`deep_gemm` / `fp8_blockscale_gemm`) can emit **NaN output rows**, and a single NaN can silently corrupt *valid* rows that merely share an output tile with a padded/garbage row. This PR makes the epilogue NaN-safe with no change to the tensor-core compute path: valid rows remain bit-identical.

📄 Full mechanism write-up (exact carrier lines, probes, and every fix attempt): https://gist.github.com/anish-shanbhag/5ea4a939068bd15779ca3de1d82bcd57

## How it shows up

Found while serving a large sparse FP8 MoE model end-to-end (Qwen3.5-397B-A17B, FP8 `[128×128]` block-scale, 8×H100, TP8+EP8, SGLang) through the cutlass FP8 block-scale grouped-MoE path. The model produced **token salad** — GSM8K dropped to **1.14%** (vs **95.98%** on the reference Triton MoE) — despite the kernel being "callable" and fast. Per-layer numerical comparison against a bf16-dequant oracle showed the kernel was *correct on valid rows* (cosine ≈ 0.9997) yet the served output was garbage, which localized the problem to **specific output rows going NaN**, not a systematic math error.

## Root cause

Grouped GEMM pads each expert's token count to a 32-row multiple, computes the padded rows, then discards them. Two things combine:

1. **The FC2 (second feed-forward) GEMM reads input rows in the tile-padding region** — past the expert's real token count — that `doActivation` never wrote. The per-token requant turns those un-written rows into **NaN fp8**, so the GEMM reads a **NaN operand** and the accumulator goes non-finite. (Measured on the **swap-AB** kernel, used at small M: `accum` is `NaN` with a *finite* scale — a contracted NaN operand, not `0 * Inf`. The **static** kernel used at higher concurrency was separately observed to overflow toward `Inf` and hit `0 * Inf`, so the exact form may be path-dependent; the sanitize covers both.)
2. The NaN is then carried to a valid token **in the epilogue**: the swap-AB transposed store stages the padding lane's accumulator into the shared-memory output line of the **valid token adjacent to the tile boundary**, and the boundary-clipped store writes that line to the valid token's output row.

This is **structural to grouped-GEMM tiling that straddles expert boundaries**, and it is worst under **sparse routing**: at low concurrency this model runs ~1.85 tokens/expert, so nearly every expert's tile is 1–2 valid rows surrounded by garbage padded rows — a maximal contamination surface. Denser MoEs (large batches, fewer experts per tile) hit the same boundary effect but dilute it to a tiny, usually-invisible fraction of rows, which is likely why it has not been reported before. The kernel is vendored from TensorRT-LLM, whose own MoE orchestrator pads/zeros differently, so the trigger does not fire there.

## The fix

Neutralize the NaN in the grouped-GEMM epilogue with a single, uniform guard applied to **both** implementations of the kernel — **finite-sanitize the accumulator before the bf16 store** (`if (!isfinite(final_accum[q])) final_accum[q] = 0.f;`):

- **`csrc/nv_internal/tensorrt_llm/deep_gemm/fp8_gemm_impl.cuh`** — the deep_gemm path, used whenever the JIT compiler is available (the default). Sanitize at **both** epilogue sites: the static `fp8_gemm_kernel` and the swap-AB `fp8_gemm_kernel_swapAB`.
- **`.../cutlass_kernels/fp8_blockscale_gemm/fp8_blockscale_gemm_kernel.cuh`** — the `TRTLLM_DG_ENABLED=0` cutlass-blockscale fallback. The identical sanitize before its bf16 store.

Zeroing a non-finite lane is safe because **valid output rows are always finite** — only padded/cross-boundary rows (which are discarded, but can bleed a NaN into a valid row through the reused shared-memory staging buffer) ever go non-finite. The wgmma tensor-core compute path is unchanged; valid rows are bit-identical (matched ratio 1.0).

## Validation

**Kernel-level, patched vs. unpatched** (same kernel both ways; the only changed variable is `fp8_gemm_impl.cuh`, `isfinite` count 0 → 2). A real-weights reproducer — Qwen3.5 layer-3 expert weights, serving-shaped global-sparse top-k routing, small M (1/2/5/8 across ranks 5/6/7, 24 seeds + a 399-forward broad sweep):

| | unpatched (stock) | patched (sanitize) |
|---|---|---|
| grid forwards with NaN output | **123 / 288** | **0 / 288** |
| broad-sweep forwards with NaN | **148 / 399** | **0 / 399** |

The NaN fires on ~43% of small-M forwards on the stock kernel and is fully eliminated by the patch; valid rows are unchanged (cosine vs a bf16-dequant oracle 0.9930–0.9997 on finite forwards). The failing forwards show finite FC1 magnitudes and `fc1_inf = 0`, confirming the NaN is the FC2 cross-boundary path the sanitize targets — not an FC1 overflow.

End-to-end corroboration:
- Eliminating the NaN restored correctness: **GSM8K 1.14% → 0.953** on the served config (coherent English, correct multi-step arithmetic).
- An expert-distribution trace confirmed the mechanism: the NaN perturbs the router and **collapses routing from 159 → 72 active experts/layer** — also why the broken path looked deceptively faster (it did ~half the padded MoE work). With the fix, routing and accuracy are restored.

**On tests:** no unit test ships with this PR. The NaN is value-correlated — it needs real expert weights with global-sparse routing (and the resulting stale padding rows) and does **not** reproduce from synthetic random inputs (a sweep over M, routing concentration, activation magnitude, and even injected extreme weight-scale dynamic range does not trigger it). flashinfer's kernel tests are synthetic-input only, so a real-weights reproducer is out of scope for the suite; and a synthetic test would pass on **both** the patched and unpatched kernel — i.e. would not guard the fix — so shipping one would be misleading. The reproducer above is the validation of record.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed FP8 matrix multiplication (GEMM) operations to prevent NaN/Inf values from contaminating results.
  * Improved output reliability in FP8 blockscale GEMM operations through enhanced accumulator validation.


csrc/nv_internal/tensorrt_llm/kernels/cutlass_kernels/fp8_blockscale_gemm/fp8_blockscale_gemm_kernel.cuh
```diff
@@ -478,32 +478,37 @@ __global__ void __launch_bounds__(TILE_M == 64 ? 256 : 384, 1)
...
+              final_accum[i * 4 + 0] += cross_0 ? 0.f : scale_0 * accum[i * 4];
+              final_accum[i * 4 + 1] += cross_0 ? 0.f : scale_0 * accum[i * 4 + 1];
             }
             for (int i = 0; i < WGMMA_OP::NUM_ACCUM / 4; i++) {
-              final_accum[i * 4 + 2] += scale_1 * accum[i * 4 + 2];
-              final_accum[i * 4 + 3] += scale_1 * accum[i * 4 + 3];
+              final_accum[i * 4 + 2] += cross_1 ? 0.f : scale_1 * accum[i * 4 + 2];
+              final_accum[i * 4 + 3] += cross_1 ? 0.f : scale_1 * accum[i * 4 + 3];
             }
           }
```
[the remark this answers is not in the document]
anish-shanbhag: Thanks — this is no longer applicable. This fallback kernel has been simplified to match the deep_gemm path: the `cross_0`/`cross_1` boundary gating is removed entirely, and the NaN is now handled by a single finite-sanitize of the accumulator before the bf16 store (`if (!isfinite(final_accum[q])) final_accum[q] = 0.f;`) — identical to `fp8_gemm_impl.cuh`. So there are no per-iteration ternaries left in the epilogue to hoist. (Updated in the latest commit on this branch.)

csrc/nv_internal/tensorrt_llm/deep_gemm/fp8_gemm_impl.cuh
```diff
@@ -363,6 +363,15 @@ __global__ void __launch_bounds__(
         }
       });
 
+      // Finite-sanitize the accumulator before the bf16 STSM write-back. Cross-boundary
+      // padding rows accumulate over in-bounds-but-stale fp8 input and can reach Inf/NaN;
+      // although that is clipped at the store, the bf16 NaN can bleed into a valid row
```
djns99, line 368: Im not super familiar with this kernel, but how does "bf16 NaN can bleed into a valid row" happen? I would assume valid rows should never read from uninitialized shared memory locations?

djns99: I think the real bug is https://github.com/anish-shanbhag/flashinfer/blob/bb557734577b3bf00ea1675f0ac441888d8ace6e/csrc/nv_internal/tensorrt_llm/deep_gemm/fp8_gemm_impl.cuh#L288

Im not sure final_accum is actually initialized to 0, only the first element (need a C++ expert here).
Accum is entirely uninitialised

djns99: Can you try change

```
float accum[WGMMA::kNumAccum], final_accum[WGMMA::kNumAccum] = {0};
```

To:

```
float accum[WGMMA::kNumAccum] = {0}, final_accum[WGMMA::kNumAccum] = {0};
```

And remove this change and see if that also fixes it?

anish-shanbhag: > how does "bf16 NaN can bleed into a valid row" happen? I would assume valid rows should never read from uninitialized shared memory locations?

From my understanding, the padded rows are actually not uninitialized memory that valid rows read; but rather the kernel actively computes them, and their NaN is written into a staging buffer that valid rows share.

From Claude, this is the sequence of events:

## 1. The NaN is *computed* by the kernel from stale *input* — not read from uninitialized output

- **Stale data lives on the *input* side.** The wgmma runs over the full `BLOCK_M` tile. When an expert has, say, 2 real tokens, the other ~126 rows of its tile are padding, and those rows of the *expanded-input* buffer hold leftover bytes (from a previous expert/iteration, or never written). That is the only "uninitialized/stale memory" in the story.
- **The NaN itself is a floating-point result the kernel produces.** The MMA accumulates over that stale fp8 input for the padded rows, and the chain (confirmed by a device dump) is:

  ```
  stale fp8 input  ->  K-accumulation goes non-finite (overflows toward Inf,
                       and/or garbage fp8 bytes decode to non-finite)
                   ->  scale-promotion `scale * accum`, where a padded row's
                       per-token scale is 0  ->  0 * Inf = NaN
  ```

So the padded rows are "garbage" in the sense of *don't-care results that the kernel actively computes*, not unwritten output memory.

## 2. Valid and padded rows go into the *same* shared staging buffer, via a transposed store-matrix

The epilogue does not store registers straight to global memory — it stages the whole tile through `smem_d` using `stmatrix.trans` (`SM90_U32x4_STSM_T`):

```cpp
SM90_U32x4_STSM_T<nv_bfloat162>::copy(..., smem_d + warp_idx*16 + i*16*BLOCK_M + tid);
//  tid = lane_idx*BLOCK_M (+8 variants)  -> the M (row/token) dimension is the interleaved stride
```

The `_T` transpose rearranges each thread's fragment so that in `smem_d` the **M-row dimension is interleaved at sub-vector granularity** within the warp's region, and that buffer is shared across the tile's valid + padded rows (and reused across tiles in the persistent kernel). The subsequent coalesced `smem_d -> gmem` store therefore reads words that pack lanes from *adjacent* M-rows together.

## 3. That interleave is the bleed

A valid row's store transaction can pull bf16 lanes physically written by an adjacent **padded** row — i.e. the padded row's NaN lands in the valid row's stored output. It is not "valid row reads uninitialized smem"; it is "the transposed, shared staging layout does not keep valid and padded rows in disjoint store words, so a computed-NaN padded lane co-locates with a valid lane."

djns99: > The NaN is computed by the kernel from stale input

I think this is the same, if stale input makes it into valid data is that not also a bug? I guess `0  ->  0 * Inf = NaN` this makes sense, I just can't see where that is happening form smem_d? We only write to smem_d on line 384, and dont do any math on it afterwards that I can see?
Good chance I am missing something here though

anish-shanbhag: @djns99 I tried the accum change and it didn't fix the issue; it seems like accum being uninitialized is actually harmless because `accum` is overwritten every stage by the wgmma ([here](https://github.com/anish-shanbhag/flashinfer/blob/bb557734577b3bf00ea1675f0ac441888d8ace6e/csrc/nv_internal/tensorrt_llm/deep_gemm/fp8_gemm_impl.cuh#L333) and [here](https://github.com/anish-shanbhag/flashinfer/blob/bb557734577b3bf00ea1675f0ac441888d8ace6e/csrc/nv_internal/tensorrt_llm/deep_gemm/fp8_gemm_impl.cuh#L722)) - when `k=0` the last `scale_d` parameter is 0, so the wgmma overwrites the uninitialized accum. But let me know if you think we should make that change anyway to avoid future bugs

djns99: Ok thanks for trying. I am not sure I fully understand how NaN in `smem_d` manages to propagate to a valid block (is it in a subsequent kernel the invalid value is read?), but the evidence speaks for itself.

anish-shanbhag: Looking into your last comment now

anish-shanbhag: I dug a bit deeper and wrote up an analysis of how the NaN propagates to a valid block, including some attempted (but unsuccessful) input-side fixes and a self-contained repro script here: https://gist.github.com/anish-shanbhag/5ea4a939068bd15779ca3de1d82bcd57

Seems like the best viable fix is still the one in this PR (on the output side) - but let me know what you think

djns99: First thing, thanks for your work. If this fixes it I am happy to get the fix merged.

I'm still getting mixed messages. What is the first location a valid value (not padding, not OOB) becomes NaN?

I think I understand if it is this case:

> So the tile rows in [num_moe_inputs, max_shape_m_32_align_padded) are an over-read of padding rows doActivation never wrote, whose requantized fp8 is NaN (Where the NaN is born, below, traces why). The bad input is on the input side.

If `doActivation` does not correctly initialise padded values and then a bug happens when we _quantize_ in a future kernel, because NaN padding causes the FP8 scale factor to become NaN as well. I would think the best fix here would be to just skip reading the padding in the quantize kernel right (if this is not possible then this PR is a fine solution)?

I'm like 90% sure this is the problem - but it says:

> Short answer: It is not a subsequent kernel. The NaN reaches the valid token inside this same FP8 grouped-GEMM kernel

Afaict NO valid data is corrupted in _this_ kernel. Padding is corrupted, but that is _only_ a problem if a future kernel reads those values in an unguarded way (i.e. what I said above about quantization kernel).

Also confusing - in one place you say:
> A valid row's own final_accum is always finite, so this touches only padding-derived lanes → valid rows bit-identical to the unpatched kernel.
No NaN ever enters smem_d or the global output, so the boundary store (Step 4) and anything downstream see only finite values → 123/288 → 0/288 NaN forwards; GSM8K 0.953; ~329 µs (perf-neutral).

Why is it a problem if "A valid row's own final_accum is always finite, so this touches only padding-derived lanes"?  To me this means no valid data is corrupted.
Just writing a NaN to memory is safe if it isn't then read anywhere.

In another it says:
> *g_data_addr = *s_data_addr;                     // the store — writes the (NaN) smem_d line to a valid token's gmem row

if this is true and we are overwriting VALID data then `isfinite()` check doesn't help. Writing zero is just as incorrect as writing NaN

PR #3654, opened 2026-06-16
JIT Cache / AOT build requirements doc

dierksen:


docs/design_docs/jit_cache_aot_kpis.md
```diff
@@ -0,0 +1,676 @@
...
+
+- requires `FLASHINFER_CUDA_ARCH_LIST`;
+- redirects `jit_env` paths to the source tree and a build directory;
+- calls the same `gen_*_module()` functions used by runtime JIT;
+- enumerates a default coverage matrix in `flashinfer/aot.py`;
+- builds the generated `JitSpec` list via a multi-module Ninja file;
+- copies `cached_ops/<name>/<name>.so` into `flashinfer_jit_cache/jit_cache/<name>/<name>.so`;
+- packages those `.so` files in a platform-specific wheel.
+
+AOT coverage is therefore an explicit registry problem: if a runtime call generates a `JitSpec.name` absent from the wheel, `flashinfer-jit-cache` cannot satisfy that call and runtime JIT is required unless disabled.
```
sricketts: > and runtime JIT is required unless disabled

This seems to be implying that somehow by disabling runtime JIT, it no longer becomes required ... but that doesn't seem right. I think what happens is that if there's a miss in the JIT cache, we return some kind of error, right?

docs/design_docs/jit_cache_aot_kpis.md
```diff
@@ -0,0 +1,676 @@
...
+The architecture set is controlled by `CompilationContext` in [`flashinfer/compilation_context.py`](../../flashinfer/compilation_context.py):
+
+- `FLASHINFER_CUDA_ARCH_LIST` wins when set.
+- Otherwise local CUDA devices are detected through PyTorch.
+- The arch list contributes to the runtime JIT workspace path.
+- Some generators further restrict supported major versions and fail early when no supported target architecture remains.
+
+## 3. Use Case Classes
+
+### 3.1 Kernel development
```
sricketts: minor: I might move this use case towards the bottom, since it's more of an advanced / internal use case.

docs/design_docs/jit_cache_aot_kpis.md
```diff
@@ -0,0 +1,676 @@
...
+- Run package-level tests in an isolated install environment.
+
+## 4. Requirements and KPIs
+
+| Use case | Requirement | KPI |
+| --- | --- | --- |
+| Kernel development | Source changes are picked up without reinstall | Time from source edit to passing targeted test; number of manual reinstall steps, target 0 |
+| Kernel development | Debuggable compiler/runtime failures | Presence of generated source, `build.ninja`, JIT log, and optional `-lineinfo`/debug build artifacts |
+| General users | First API call succeeds on supported systems | First-call success rate by API, CUDA version, and GPU arch |
+| General users | First-use overhead is understandable | Cold first-call latency and warm repeated-call latency per representative API |
```
sricketts: What does it mean for the first-use overhead to be "understandable"?

I think at a very coarse granularity, we want it to be clear whether the first use will: hit in the jit-cache wheel, hit in the cubin wheel, hit in the user cache, or miss. 

But then on miss, the latency can vary widely -- there could be a cubin fetch (network latency) or local compile (which itself can vary from C++ template case to faster compilation flows like CuTeDSL). I'm not sure whether it's easy for the user to predict what will happen here or understand why it happened.

Are we saying we want there to be some kind of clear log that explains what happened?

In any case, I'm not sure how the KPIs in this row relate to the requirement.

I think maybe what we want is to split two things:
* understandable behavior
* sufficiently low overhead for each use case (and we can work with the community to define some actual numbers here)

docs/design_docs/jit_cache_aot_kpis.md
```diff
@@ -0,0 +1,676 @@
...
+- Wheel size visibility.
+- Automated missing-coverage reports before publishing.
+- Reproducible package index layout.
+
+Primary package mode:
+
+- Build all package artifacts in CI.
+- Run package-level tests in an isolated install environment.
+
+## 4. Requirements and KPIs
```
sricketts, line 221: This is a long list that may be hard for people to fully absorb and give useful feedback on. But I like that we've thought so thoroughly about the requirements.

Maybe we can split it like this:
1. High level requirements - here I'd think about a "KPI" as something important and measurable that we might talk about on a slide or blog for the community ... something like "__ seconds to fully compile for model foo for SM100"
2. Details - e.g. "Wrong package combinations fail early" is important, but not that remarkable ... more of an engineering detail that people would expect "just works"

sricketts: Thinking more about (1), I think it's something like:
* how big the packages are
* how good the model coverage is if you install all of the packages
* how long it takes to JIT when you don't install all of the packages

... of course others should chime in if this misses something. But if we have a crisp set of high level things to optimize for, then we can optimize over time and build a great product.

All of the other functional details have to work too, but to me they feel like check-boxes that are more straight-forward, and I'd move them to (2).

docs/design_docs/jit_cache_aot_kpis.md
```diff
@@ -0,0 +1,676 @@
...
+
+Measure at least these package/cache scenarios:
+
+| Scenario | Packages | Cache state | Environment |
+| --- | --- | --- | --- |
+| Core only, true cold | `flashinfer-python` | empty JIT cache and empty cubin cache | downloads/JIT allowed |
+| Core plus cubin | `flashinfer-python`, `flashinfer-cubin` | empty JIT cache | downloads avoided for packaged cubins |
+| Core plus JIT cache | `flashinfer-python`, `flashinfer-jit-cache` | empty cubin cache | JIT avoided for covered modules |
+| Fully prebuilt/offline | all three packages | empty user cache | `FLASHINFER_DISABLE_JIT=1`, `FLASHINFER_NO_DOWNLOAD=1` |
+| Warm cache | any chosen package set | populated runtime cache | normal production env |
```
sricketts: I'm not sure I understand the "warm cache" scenario ... what are we trying to learn?

The "any chosen package set" seems like a broad bucket of possibilities that will make the results from this test scenario kind of fuzzy to interpret.

docs/design_docs/jit_cache_aot_kpis.md
```diff
@@ -0,0 +1,676 @@
...
+
+- Which FlashInfer APIs are reached in their supported serving paths?
+- Which models, dtypes, quantization modes, and attention backends must be covered?
+- Which head dimensions, page sizes, sliding-window modes, logits-soft-cap modes, index dtypes, and MLA/XQA/MoE variants are used?
+- Which GPU architectures and CUDA versions are in their release/support matrix?
+- Is runtime `nvcc` available in production images?
+- Are production images expected to work without runtime `nvcc`, or is `nvcc` intentionally part of the serving image contract?
+- Is network egress available during startup?
+- Is there an explicit cold-start, engine-start, or first-token latency SLA?
+- Is a startup warmup phase allowed, and how long may it run?
```
sricketts: I think we know that startup warmup is allowed in the "Production serving with warmup allowed" use case. I would constrain this question to that use case -- e.g. "In the `Production serving with warmup allowed` use case, how much overhead is acceptable for kernel compilation?"

docs/design_docs/jit_cache_aot_kpis.md
```diff
@@ -0,0 +1,676 @@
...
+
+- The request is not a production blocker today, but the customer is motivated and framed it as internal housekeeping.
+- P1 seems reasonable if the goal is a documented, supported custom AOT build path with manifest-based pruning.
+- A three-week estimate is plausible for a minimal productized path if scoped to JIT-cache `.so` packaging and validation. It is likely larger if it also includes automatic framework trace generation, arbitrary manifest-to-generator reverse mapping, or pruned cubin packaging.
+
+## 8. Questions for vLLM, SGLang, and Customers
+
+Ask each integration team for a concrete workload matrix, not just a package preference.
+
+Minimum questions:
```
sricketts: I think let's distill this list of questions to target the list of "high level requirements/KPIs" from above. For each of those, what information do we need from the FWs to nail down the details?

docs/design_docs/jit_cache_aot_kpis.md
```diff
@@ -0,0 +1,676 @@
...
+
+- Which FlashInfer APIs are reached in their supported serving paths?
```
sricketts: I don't think we need to ask this, for a couple of reasons:
* we already know the rough answer
* any details or gaps are not fundamental -- the goal is for FlashInfer to cover all of the important primitives, so FlashInfer should know which those are and make sure they work well

docs/design_docs/jit_cache_aot_kpis.md
```diff
@@ -0,0 +1,676 @@
+# FlashInfer JIT Cache and AOT Wheel Use Cases
+
+*Working note - v0.1 - June 2026*
+
+## 1. Purpose
+
+FlashInfer has three related but distinct runtime artifact paths:
```
sricketts:
```suggestion
The choice of JIT vs AOT deployment of GPU kernels for inference requires reasoning about tradeoffs across binary size, warmup overhead, and runtime performance. FlashInfer leaves these tradeoffs to the user, providing three related but distinct runtime artifact paths:
```

docs/design_docs/jit_cache_aot_kpis.md
```diff
@@ -0,0 +1,676 @@
...
+
+- `flashinfer-python`: the core package. It generates CUDA/C++ sources, compiles JIT shared libraries on demand, and can download cubin artifacts on demand.
+- `flashinfer-jit-cache`: an optional package containing pre-built JIT shared libraries (`.so` files) for a specific FlashInfer version, CUDA build, and architecture set.
+- `flashinfer-cubin`: an optional package containing pre-downloaded cubin/header artifacts used by some TensorRT-LLM, DeepGEMM, cuDNN, and CuTe DSL backed paths.
+
+This document captures the current design, likely use cases, requirements, KPIs, and concrete measurement methods. It is based on the repository implementation and should be treated as a starting point before validating requirements with integration teams such as vLLM, SGLang, and customers that want self-built AOT packages.
```
sricketts:
```suggestion
However, today there are several problems:

* It's not obvious what the JIT overhead implications are for some given subset of packages
* The binary size of the `flashinfer-jit-cache` package is growing rapidly, and it's not obvious what to do about that
* Users generally complain about a bucket of UX issues related to the above, and without an overall strategy, we're solving each one locally rather than moving in some coherent longer term direction

This document is a proposal to address these problems. It summarizes the current design, use cases, requirements, KPIs, and concrete measurement methods. The intent is to iterate on this with the community to set a long term vision for improving the JIT/AOT experience for FlashInfer.
```

sricketts (commented):
Great starting point, thanks @dierksen.

docs/design_docs/jit_cache_aot_kpis.md
dierksen: This is just ensuring that the JIT cache is caching JIT compilation, rather than  rebuilding every time or depending on AOT compilations being loaded through the JIT cache. It's more for completeness than anything.

PR #3664, opened 2026-06-17
fix(mla): make cute-dsl decode eligibility gate impl-aware

zcnrex:
## Problem

`trtllm_batch_decode_with_kv_cache_mla(backend="cute-dsl")` (or `backend="auto"`) raises for valid MLA-decode configs that the kernel can actually run, e.g.:

```
ValueError: cute-dsl backend (MLA decode kernel) cannot implement this configuration:
cute_dsl_mla_decode: unsupported configuration (q_len=8, num_heads=16, page_size=64,
in_dtype=torch.bfloat16, out_dtype=torch.bfloat16)
```

`q_len=8, num_heads=16` is well within the monolithic kernel's `fold_sq` limit (`num_heads * F <= 128`, here `F=8` → `16*8=128`).

q_len=8 is a case for DFlash https://www.lmsys.org/blog/2026-06-15-next-generation-speculative-decoding-dflash-v2/

## Root cause

The eligibility gate `_cute_dsl_incompatibility_reason` always validated against the **modular** kernel's constraints (`MLAConfig.can_implement`: hard-caps `seq_len_q <= 4`, no folding), **even when the call resolves to the monolithic impl** (the default, `_DEFAULT_IMPL="monolithic"`). The monolithic kernel folds `seq_len_q` into the 128-wide MMA-M tile, so `q_len > 4` is supported there — but the gate rejected it before dispatch using the wrong impl's rules.

## Fix

Resolve the impl the same way the dispatcher does (`mla_dispatch._resolve_impl`), then validate against the matching `_check_can_implement` (monolithic vs modular). Also:
- Surfaces the `monolithic` + `sinks` conflict as a clean reason string instead of a late forward-time crash.
- Keeps `ImportError` handling around the cute-dsl import so `backend="auto"` still degrades gracefully (drops cute-dsl, falls back to trtllm-gen) when cute-dsl deps are absent.

## Validation

On B200, `q_len=8, num_heads=16, bf16, page_size=64` (bs=10) vs an fp32 torch MTP reference on byte-identical inputs:

| seq | backend | SNR (dB) | cos | time (µs) |
|---|---|---|---|---|
| 2,048 | cute-dsl | 54.84 | 0.999998 | 18.2 |
| 2,048 | trtllm-gen | 54.64 | 0.999998 | 37.0 |
| 50,000 | cute-dsl | 55.44 | 0.999999 | 127.3 |
| 50,000 | trtllm-gen | 55.41 | 0.999999 | 803.6 |

The monolithic cute-dsl kernel computes the previously-rejected config correctly (matches trtllm-gen and the torch reference).

## Summary by CodeRabbit

* **Refactor**
  * Improved CuteDSL compatibility checking by determining the effective CuteDSL implementation and using the right eligibility logic for compatibility decisions.
  * When using automatic backend selection, the `cute_dsl_impl` preference is now honored, and incompatibility reasons are presented more clearly if requirements can’t be met.

### DIFF
--- a/flashinfer/mla/_core.py
+++ b/flashinfer/mla/_core.py
@@ -2139,7 +2139,7 @@ def _cute_dsl_incompatibility_reason(
     try:
         from ..cute_dsl.attention.mla_dispatch import _resolve_impl
         resolved_impl = _resolve_impl(requested=cute_dsl_impl, kwargs={"sinks": sinks})
-    except ValueError as e:
+    except (ValueError, ImportError) as e:
         return f"cute-dsl backend (MLA decode kernel): {e}"
 
     try:
--- a/flashinfer/mla/_core.py
+++ b/flashinfer/mla/_core.py
@@ -2089,17 +2089,10 @@ def _cute_dsl_incompatibility_reason(
     Used by both the explicit ``backend="cute-dsl"`` path (raises the reason
     as a ``ValueError``) and the ``backend="auto"`` filter (silently drops
     cute-dsl from the runners list).
-
-    ``cute_dsl_impl`` mirrors the dispatcher's impl selection so the
-    eligibility check validates against the constraints of the impl that will
-    actually run (modular vs monolithic).
     """
     cc = get_compute_capability(query.device)
-    if cc[0] not in (10, 11):
-        return (
-            "cute-dsl backend (MLA decode kernel) uses tcgen05 2-SM instructions, "
-            f"supported only on SM100-SM110 (datacenter Blackwell); got SM{cc[0]}{cc[1]}"
-        )
+    if cc[0] < 10:
+        return f"cute-dsl backend (MLA decode kernel) requires SM100+, got SM{cc[0]}{cc[1]}"
     if isinstance(bmm1_scale, torch.Tensor):
         return (
             "cute-dsl backend (MLA decode kernel) does not support tensor bmm1_scale, "
--- a/flashinfer/mla/_core.py
+++ b/flashinfer/mla/_core.py
@@ -2082,12 +2082,17 @@ def _cute_dsl_incompatibility_reason(
     is_var_seq: bool,
     return_lse: bool,
     lse: Optional[torch.Tensor],
+    cute_dsl_impl: str = "auto",
 ) -> Optional[str]:
     """Return None if cute-dsl can handle this call, else a human-readable reason.
 
     Used by both the explicit ``backend="cute-dsl"`` path (raises the reason
     as a ``ValueError``) and the ``backend="auto"`` filter (silently drops
     cute-dsl from the runners list).
+
+    ``cute_dsl_impl`` mirrors the dispatcher's impl selection so the
+    eligibility check validates against the constraints of the impl that will
+    actually run (modular vs monolithic).
     """
     cc = get_compute_capability(query.device)
     if cc[0] < 10:
@@ -2132,7 +2137,16 @@ def _cute_dsl_incompatibility_reason(
 
     _, q_len, num_heads, _ = query.shape
     try:
-        from ..cute_dsl.attention.wrappers.batch_mla import _check_can_implement
+        from ..cute_dsl.attention.mla_dispatch import _resolve_impl
+        resolved_impl = _resolve_impl(requested=cute_dsl_impl, kwargs={"sinks": sinks})
+    except ValueError as e:
+        return f"cute-dsl backend (MLA decode kernel): {e}"
+
+    try:
+        if resolved_impl == "monolithic":
+            from ..cute_dsl.attention.monolithic.mla_decode import _check_can_implement
+        else:
+            from ..cute_dsl.attention.wrappers.batch_mla import _check_can_implement
 
         _check_can_implement(
             torch_dtype=query.dtype,
@@ -3043,6 +3057,7 @@ def trtllm_batch_decode_with_kv_cache_mla(
         is_var_seq,
         return_lse,
         lse,
+        cute_dsl_impl,
     )
     if backend == "cute-dsl":
         if cute_dsl_reason is not None:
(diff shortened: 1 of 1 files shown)


waynehacking8:
Thanks for making the eligibility gate impl-aware! Adjacent SM12x gap in the same function: the `if cc[0] < 10:` arch guard lets SM120/SM121 (consumer/workstation Blackwell) through, but the cute-dsl MLA-decode kernels are tcgen05/2-SM (SM100/SM103/SM110 only). After this PR resolves the impl, the monolithic/modular `_check_can_implement` has no arch gate, so an explicit `backend="cute-dsl"` on SM120 still proceeds into a tcgen05 op and dies with a cryptic error instead of the clean reason string.

On an RTX PRO 6000 (SM120) the analogous tcgen05 path (mamba SSD, same `MmaF16BF16Op`) fails with:

```
OpError: expects arch to be one of [sm_100a, sm_100f, sm_101a, sm_101f, sm_103a, sm_103f, sm_110a, sm_110f], but got sm_120a
```

Would it make sense to also reject SM12x here, e.g. `if cc[0] not in (10, 11):` (the tcgen05 op's own supported set)? Happy to send a small follow-up on top of this PR if useful.

zcnrex:
> Thanks for making the eligibility gate impl-aware! Adjacent SM12x gap in the same function: the `if cc[0] < 10:` arch guard lets SM120/SM121 (consumer/workstation Blackwell) through, but the cute-dsl MLA-decode kernels are tcgen05/2-SM (SM100/SM103/SM110 only). After this PR resolves the impl, the monolithic/modular `_check_can_implement` has no arch gate, so an explicit `backend="cute-dsl"` on SM120 still proceeds into a tcgen05 op and dies with a cryptic error instead of the clean reason string.
> 
> On an RTX PRO 6000 (SM120) the analogous tcgen05 path (mamba SSD, same `MmaF16BF16Op`) fails with:
> 
> ```
> OpError: expects arch to be one of [sm_100a, sm_100f, sm_101a, sm_101f, sm_103a, sm_103f, sm_110a, sm_110f], but got sm_120a
> ```
> 
> Would it make sense to also reject SM12x here, e.g. `if cc[0] not in (10, 11):` (the tcgen05 op's own supported set)? Happy to send a small follow-up on top of this PR if useful.

Thanks for the comment. I have updated the code. Please take a look again.

waynehacking8:
Confirmed on an RTX PRO 6000 (SM120, CUDA 13.0) with f839824: `_cute_dsl_incompatibility_reason(...)` now returns

> cute-dsl backend (MLA decode kernel) uses tcgen05 2-SM instructions, supported only on SM100-SM110 (datacenter Blackwell); got SM120

so `backend="cute-dsl"` raises this cleanly and `backend="auto"` drops cute-dsl as intended — no more cryptic tcgen05 `OpError` on SM12x. LGTM, thanks for the quick turnaround! 🙏

flashinfer/mla/_core.py
```diff
@@ -1944,16 +1944,24 @@ def _cute_dsl_incompatibility_reason(
...
+    ``cute_dsl_impl`` mirrors the dispatcher's impl selection so the
+    eligibility check validates against the constraints of the impl that will
+    actually run (modular vs monolithic).
     """
     cc = get_compute_capability(query.device)
-    if cc[0] < 10:
-        return f"cute-dsl backend (MLA decode kernel) requires SM100+, got SM{cc[0]}{cc[1]}"
+    if cc[0] not in (10, 11):
+        return (
+            "cute-dsl backend (MLA decode kernel) uses tcgen05 2-SM instructions, "
```
saltyminty: This seems like an unnecessary change, and the new error message isn't more helpful than the original message.

flashinfer/mla/_core.py
```diff
@@ -1944,16 +1944,24 @@ def _cute_dsl_incompatibility_reason(
...
     lse: Optional[torch.Tensor],
+    cute_dsl_impl: str = "auto",
 ) -> Optional[str]:
     """Return None if cute-dsl can handle this call, else a human-readable reason.
 
     Used by both the explicit ``backend="cute-dsl"`` path (raises the reason
     as a ``ValueError``) and the ``backend="auto"`` filter (silently drops
     cute-dsl from the runners list).
+
+    ``cute_dsl_impl`` mirrors the dispatcher's impl selection so the
```
saltyminty: docstring reads a little weird here

saltyminty (commented):
Left two comments about comments/docstrings but otherwise looks good.

flashinfer/mla/_core.py
zcnrex: reverted

flashinfer/mla/_core.py
zcnrex: Removed this docstring as there was no other docstring.

saltyminty (approved):

leejnau:
@zcnrex please make sure you've run the pre-commit steps (https://github.com/flashinfer-ai/flashinfer/blob/main/.pre-commit-config.yaml):

```
ruff format..............................................................Failed
- hook id: ruff-format
- files were modified by this hook

1 file reformatted, 766 files left unchanged

pre-commit hook(s) made changes.
If you are seeing this message in CI, reproduce locally with: `pre-commit run --all-files`.
To run `pre-commit` as part of git workflow, use `pre-commit install`.
All changes made by hooks:
diff --git a/flashinfer/mla/_core.py b/flashinfer/mla/_core.py
index 2a12d9c..28cac52 100644
--- a/flashinfer/mla/_core.py
+++ b/flashinfer/mla/_core.py
@@ -2134,6 +2134,7 @@ def _cute_dsl_incompatibility_reason(
     _, q_len, num_heads, _ = query.shape
     try:
         from ..cute_dsl.attention.mla_dispatch import _resolve_impl
+
         resolved_impl = _resolve_impl(requested=cute_dsl_impl, kwargs={"sinks": sinks})
     except (ValueError, ImportError) as e:
         return f"cute-dsl backend (MLA decode kernel): {e}"
Error: Process completed with exit code 1.
```

PR #3464, opened 2026-05-29
Enable SM100 FP4 swap_ab for non-8 M

Vinnie6167:
## 📌 Description

This PR removes the guard against using swap AB for the CuTe DSL mm_fp4 backend when M is not divisible by 8. This guard was unnecessary and limiting available tactics. See [#3402](https://github.com/flashinfer-ai/flashinfer/issues/3402).

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [x] Tests have been added or updated as needed.
- [x] All tests are passing (`unittest`, etc.).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of FP4 matrix operations to detect invalid configurations earlier with clearer error messages.

* **Performance**
  * Streamlined tactic selection logic for matrix computations for better efficiency.

PR #3667, opened 2026-06-17
Enable SwapAB in `mm_fp4` `cute-dsl` backend when M is not a multiple of 8.

### LINKED ISSUE #3648: mm_fp4 with cuteDSL backend will pick wrong autotune results for tile, cluster size on SM10X
Here are the following shapes, which the problem is encountered in https://github.com/sgl-project/sglang/issues/28224

| Shape | default tile | default cluster | autotuned tile | autotuned cluster |
| :--- | :--- | :--- | :--- | :--- |
| (1,1024)×(1024,8192) | 128×64 | (1,4) | 128×128 | (2,2) |
| (1,3584)×(3584,8192) | 128×64 | (1,4) | 128×128 | (1,4) |

Here are the performance detail on SM103.

| Shape | default tile/cluster | autotuned tile/cluster | default | autotuned | faster |
| :--- | :--- | :--- | :--- | :--- | :--- |
| (1,1024)×(1024,8192) | 128×64, (1,4) | 128×128, (2,2) | 4.90 µs | 5.56 µs | default by 12% |
| (1,3584)×(3584,8192) | 128×64, (1,4) | 128×128, (1,4) | 7.96 µs | 8.90 µs | default by 11% |

```python3
import torch
import flashinfer.gemm.kernels.utils as ku
from flashinfer import mm_fp4, nvfp4_quantize, SfLayout

torch.manual_seed(0)
dev = "cuda"

# (name, M, K, N, default_tactic, autotuned_tactic)
# tactic = (mma_tiler_mn, cluster_shape_mn, swap_ab, use_prefetch, kernel_type, use_tma_store)
# name, M, K, N, dict of {label: tactic}
CASES = [
    ("(1,1024)x(1024,8192)", 1, 1024, 8192, {
        "heuristic":     ((128, 64),  (1, 4), False, False, "sm100", None),
        "old_cache":     ((128, 128), (2, 2), False, False, "sm100", None),
        "fresh_autotune":((128, 128), (1, 1), False, False, "sm100", None),
    }),
    ("(1,3584)x(3584,8192)", 1, 3584, 8192, {
        "heuristic":     ((128, 64),  (1, 4), False, False, "sm100", None),
        "fresh_autotune":((128, 128), (1, 4), False, False, "sm100", None),
    }),
]


def prep(m, k, n):
    a = torch.randn(m, k, device=dev, dtype=torch.bfloat16)
    b = torch.randn(n, k, device=dev, dtype=torch.bfloat16)  # weight (n,k), pass b.T as (k,n)
    gsa = (448 * 6) / a.float().abs().max()
    gsb = (448 * 6) / b.float().abs().max()
    a4, a_s = nvfp4_quantize(a, gsa, sfLayout=SfLayout.layout_128x4, do_shuffle=False)
    b4, b_s = nvfp4_quantize(b, gsb, sfLayout=SfLayout.layout_128x4, do_shuffle=False)
    alpha = 1.0 / (gsa * gsb)
    out = torch.empty(m, n, device=dev, dtype=torch.bfloat16)
    return a4, b4, a_s, b_s, alpha, out


def run(args):
    a4, b4, a_s, b_s, alpha, out = args
    mm_fp4(a4, b4.T, a_s, b_s.T, alpha, torch.bfloat16, out,
           block_size=16, backend="cute-dsl", use_nvfp4=True)


from flashinfer.testing import bench_gpu_time_with_cupti


def time_it(args):
    for _ in range(10):
        run(args)
    torch.cuda.synchronize()
    times = bench_gpu_time_with_cupti(lambda: run(args))
    return sum(times) / len(times) * 1000  # us (times are ms)


for name, m, k, n, tactics in CASES:
    args = prep(m, k, n)
    results = {}
    for label, tactic in tactics.items():
        ku._SM100_MM_FP4_TACTIC_CACHE.clear()
        orig = ku._select_sm100_mm_fp4_cute_dsl_tactic
        ku._select_sm100_mm_fp4_cute_dsl_tactic = lambda *a, _t=tactic, **kw: _t
        gb._select_sm100_mm_fp4_cute_dsl_tactic = lambda *a, _t=tactic, **kw: _t
        try:
            results[label] = (time_it(args), tactic)
        finally:
            ku._select_sm100_mm_fp4_cute_dsl_tactic = orig
            gb._select_sm100_mm_fp4_cute_dsl_tactic = orig
    print(f"\n=== {name} ===")
    best = min(results, key=lambda l: results[l][0])
    for label, (t, tactic) in sorted(results.items(), key=lambda x: x[1][0]):
        mark = "  <-- fastest" if label == best else ""
        print(f"  {label:15s} tile={tactic[0]} cluster={tactic[1]}: {t:7.2f} us{mark}")
```

b8zhong:
## 📌 Description

Enable SwapAB in `mm_fp4` `cute-dsl` backend when M is not a multiple of 8, and through offline tuning, use M <= 32 as a heuristic for SwapAB (typically, this is the point where there is no benefit).

Note: it's very similar to https://github.com/flashinfer-ai/flashinfer/pull/3464, except add the cuda-graph and cold-L2 mode, since otherwise the autotuner will fail to profile the correct timings. (Note: this is currently already used in `trtllm_fp*_block_scale_moe` autotuning. This also fixes https://github.com/flashinfer-ai/flashinfer/issues/3648, where the autotuner fails to pick the fastest tile on Llama 3.3 70B FP4 shapes

As well, fix some warnings emitted from the DSL, with 4.5.0

## 🔍 Related Issues

For example, here are the speedups in the M = 1 case in DSV3 shapes:

| K    | N    | Before — Tile Size | Before — Cluster Shape | Before — Swap AB | Before (µs) | After — Tile Size | After — Cluster Shape | After — Swap AB | After (µs) | Speedup |
|-----:|-----:|:------------------:|:----------------------:|:----------------:|------------:|:-----------------:|:---------------------:|:---------------:|-----------:|:-------:|
| 256  | 7168 | 128 × 64           | 1 × 2                  | No               | 3.81        | 128 × 32          | 1 × 1                 | Yes             | 3.26       | 1.17×   |
| 2048 | 7168 | 128 × 64           | 1 × 4                  | No               | 6.02        | 128 × 16          | 1 × 1                 | Yes             | 5.47       | 1.10×   |
| 2304 | 7168 | 128 × 64           | 1 × 4                  | No               | 6.16        | 128 × 16          | 1 × 1                 | Yes             | 5.57       | 1.11×   |
| 3584 | 1024 | 256 × 64           | 4 × 4                  | No               | 5.82        | 128 × 8           | 1 × 1                 | Yes             | 4.90       | 1.19×   |
| 3584 | 9216 | 128 × 128          | 1 × 4                  | No               | 9.06        | 128 × 16          | 1 × 1                 | Yes             | 7.20       | 1.26×   |

In the E2E framework speedup, it could be around 135 -> 140 TPS in BS = 1 decoding.

Here are the cases for Llama 3.3 70B FP4 + TP4:

| Shape (M × K) × (K × N)  | Before — Tile Size | Before — Cluster Shape | Before — Swap AB | Before (µs) | After — Tile Size | After — Cluster Shape | After — Swap AB | After (µs) | Speedup |
|:------------------------:|:------------------:|:----------------------:|:----------------:|------------:|:-----------------:|:---------------------:|:---------------:|-----------:|:-------:|
| (1, 1024) × (1024, 8192) | 128 × 64           | 1 × 4                  | No               | 4.99        | 128 × 8           | 1 × 1                 | Yes             | 4.32       | 1.16×   |
| (1, 3584) × (3584, 8192) | 128 × 64           | 1 × 4                  | No               | 7.90        | 128 × 16          | 1 × 1                 | Yes             | 6.94       | 1.14×   |

### ✅ Pre-commit Checks

## 🧪 Tests

```
pytest tests/gemm/test_mm_fp4.py
=========================================================================================================== test session starts ===========================================================================================================
platform linux -- Python 3.12.3, pytest-9.0.3, pluggy-1.6.0
rootdir: /sgl-workspace/flashinfer
configfile: pytest.ini
plugins: anyio-4.13.0, typeguard-4.5.2
collected 23957 items  
...
============================================================================================================ warnings summary =============================================================================================================
tests/conftest.py:16
tests/conftest.py:16
  /sgl-workspace/flashinfer/tests/conftest.py:16: DeprecationWarning: tcgen05.OperandMajorMode is deprecated, use cute.nvgpu.OperandMajorMode instead
    from cutlass.cute.nvgpu.tcgen05 import OperandMajorMode

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
======================================================================================= 6336 passed, 17621 skipped, 2 warnings in 797.85s (0:13:17) =======================================================================================
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed SM100 block-scaled GEMM setup to consistently respect the B operand dtype.
  * Improved FP4 (SM100, cute-dsl) tactic selection and swap handling, with updated validity rules based on N alignment and broader small-M support.
  * Added/strengthened FP4 validation so the cute-dsl backend is only used when N is 8-aligned (otherwise it errors/falls back).

* **Performance**
  * Improved FP4 autotuning defaults by enabling CUDA Graphs and cold L2 caching.

* **Tests**
  * Expanded FP4 coverage and added a negative test for misaligned N behavior.

### DIFF
--- a/flashinfer/gemm/gemm_base.py
+++ b/flashinfer/gemm/gemm_base.py
@@ -4433,7 +4433,6 @@ def _get_sm100_block_scaled_tactics(
     )
 
     batch_size = 1
-    m_aligned = m % 8 == 0
     n_aligned = n % 8 == 0
 
     valid_tactics = []
@@ -4442,7 +4441,7 @@ def _get_sm100_block_scaled_tactics(
             for swap_ab in (False, True):
                 if not swap_ab and not n_aligned:
                     continue
-                if swap_ab and not m_aligned:
+                if swap_ab and not n_aligned:
                     continue
 
                 if swap_ab:
@@ -5645,15 +5644,7 @@ def _cute_dsl_gemm_fp4_runner(
                 a.device,
             )
 
-            if m == 1:
-                if n <= 1024:
-                    allowed_tiles = {(256, 64)}
-                elif n >= 8192:
-                    allowed_tiles = {(128, 128)}
-                else:
-                    allowed_tiles = {(128, 64)}
-                sm100_base = [t for t in sm100_base if t[0] in allowed_tiles]
-            elif m in (8, 16):
+            if m <= 32:
                 allowed_tiles = {
                     (128, 8),
                     (128, 16),
@@ -5743,6 +5734,9 @@ def _cute_dsl_gemm_fp4_runner(
             sf_vec_size = 16 if use_nvfp4 else 32
             sf_dtype = cutlass.Float8E4M3FN if use_nvfp4 else cutlass.Float8E8M0FNU
             batch_size = 1
+
+            if n % 8 != 0:
+                raise ValueError(f"CuTe-DSL FP4 GEMM requires N % 8 == 0, got n={n}")
 
             if tactic is None or tactic == -1:
                 # Use analytical heuristic to pick the best tactic based on
@@ -6133,6 +6127,8 @@ def _mxfp8_swizzled_scale_len(m: int, k: int, swizzle_layout: SfLayout) -> int:
 
 
 _MM_FP4_TUNING_CONFIG_8x4 = TuningConfig(
+    use_cuda_graph=True,
+    use_cold_l2_cache=True,
     dynamic_tensor_specs=(
         DynamicTensorSpec(
             (0,),  # a_tensor_index
@@ -6162,6 +6158,8 @@ _MM_FP4_TUNING_CONFIG_8x4 = TuningConfig(
 
 
 _MM_FP4_TUNING_CONFIG_128x4 = TuningConfig(
+    use_cuda_graph=True,
+    use_cold_l2_cache=True,
     dynamic_tensor_specs=(
         DynamicTensorSpec(
             (0,),  # a_tensor_index
--- a/flashinfer/gemm/gemm_base.py
+++ b/flashinfer/gemm/gemm_base.py
@@ -5488,7 +5488,7 @@ def _cutlass_gemm_fp4_requirement(
 @supported_compute_capability([100, 103])
 def _cute_dsl_gemm_fp4_requirement(
     a: torch.Tensor,  # unused
-    b: torch.Tensor,  # unused
+    b: torch.Tensor,
     a_descale: torch.Tensor,  # unused
     b_descale: torch.Tensor,  # unused
     alpha: Optional[torch.Tensor] = None,  # unused
@@ -5496,9 +5496,7 @@ def _cute_dsl_gemm_fp4_requirement(
     out: Optional[torch.Tensor] = None,  # unused
     block_size: int = 16,  # unused
     use_8x4_sf_layout: bool = False,
-    backend: Literal[
-        "cudnn", "trtllm", "cutlass", "cute-dsl", "b12x", "auto"
-    ] = "auto",  # unused
+    backend: Literal["cudnn", "trtllm", "cutlass", "cute-dsl", "b12x", "auto"] = "auto",
     use_nvfp4: bool = True,
     enable_pdl: bool = True,  # unused
 ):
@@ -5508,6 +5506,11 @@ def _cute_dsl_gemm_fp4_requirement(
     # preparation for 128x4 layout.
     if use_8x4_sf_layout:
         raise ValueError("cute_dsl FP4 GEMM only supports 128x4 scale factor layout.")
+    # N must be 8-aligned; raise only for explicit cute-dsl (never auto-selected).
+    if b.shape[1] % 8 != 0:
+        if backend != "cute-dsl":
+            return False
+        raise ValueError(f"CuTe-DSL FP4 GEMM requires N % 8 == 0, got n={b.shape[1]}")
     _check_cute_dsl_availability()
     return True
 
@@ -5732,9 +5735,6 @@ def _cute_dsl_gemm_fp4_runner(
             sf_vec_size = 16 if use_nvfp4 else 32
             sf_dtype = cutlass.Float8E4M3FN if use_nvfp4 else cutlass.Float8E8M0FNU
             batch_size = 1
-
-            if n % 8 != 0:
-                raise ValueError(f"CuTe-DSL FP4 GEMM requires N % 8 == 0, got n={n}")
 
             if tactic is None or tactic == -1:
                 # Use analytical heuristic to pick the best tactic based on
--- a/flashinfer/gemm/gemm_base.py
+++ b/flashinfer/gemm/gemm_base.py
@@ -4439,9 +4439,7 @@ def _get_sm100_block_scaled_tactics(
     for mma_tiler_mn in _SM100_MMA_TILER_MN_CANDIDATES:
         for cluster_shape_mn in _SM100_CLUSTER_SHAPE_MN_CANDIDATES:
             for swap_ab in (False, True):
-                if not swap_ab and not n_aligned:
-                    continue
-                if swap_ab and not n_aligned:
+                if not n_aligned:
                     continue
 
                 if swap_ab:
--- a/flashinfer/gemm/kernels/utils.py
+++ b/flashinfer/gemm/kernels/utils.py
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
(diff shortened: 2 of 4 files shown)


flashinfer/gemm/gemm_base.py
```diff
@@ -4442,7 +4441,7 @@ def _get_sm100_block_scaled_tactics(
             for swap_ab in (False, True):
                 if not swap_ab and not n_aligned:
                     continue
-                if swap_ab and not m_aligned:
+                if swap_ab and not n_aligned:
                     continue
```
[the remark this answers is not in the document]
b8zhong: Just cleaned it up.

flashinfer/gemm/gemm_base.py
```diff
@@ -5744,6 +5735,9 @@ def forward(
             sf_dtype = cutlass.Float8E4M3FN if use_nvfp4 else cutlass.Float8E8M0FNU
             batch_size = 1
 
+            if n % 8 != 0:
+                raise ValueError(f"CuTe-DSL FP4 GEMM requires N % 8 == 0, got n={n}")
```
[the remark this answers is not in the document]
b8zhong: Nice suggestion

tests/gemm/test_mm_fp4.py
leejnau, line 151: Maybe add a focused regression test for the new `N % 8` guard at `flashinfer/gemm/gemm_base.py:5509`, using a shape like `m=16, n=130, k=128` with `backend="cute-dsl"`, `use_128x4_sf_layout=True`, and `fp4_type="nvfp4"`.
This can stay separate from the global `n` parametrization to avoid expanding the full backend matrix.

flashinfer/gemm/kernels/utils.py
leejnau: Minor cleanup: this now should be `next_positive_power_of_2` not `last_positive_power_of_2`.

leejnau: Minor cleanup: after the `swap_ab` heuristic change, `_compute_tactic_for_m(..., m_aligned)` no longer appears to use `m_aligned`, so the aligned/non-aligned cache entries at `utils.py:163-171` should now be identical. Maybe simplify the cache to key only by `bucket`, or add a short comment if keeping the duplicate entries is intentional.

flashinfer/gemm/gemm_base.py
```diff
@@ -5511,6 +5506,11 @@ def _cute_dsl_gemm_fp4_requirement(
     # preparation for 128x4 layout.
     if use_8x4_sf_layout:
         raise ValueError("cute_dsl FP4 GEMM only supports 128x4 scale factor layout.")
+    # N must be 8-aligned; raise only for explicit cute-dsl (never auto-selected).
+    if b.shape[1] % 8 != 0:
+        if backend != "cute-dsl":
+            return False
+        raise ValueError(f"CuTe-DSL FP4 GEMM requires N % 8 == 0, got n={b.shape[1]}")
```
Vinnie6167, line 5513: I believe this is too restrictive. We can  use the cute-dsl backend where b.shape[1] % 8 != 0 as long as swap_ab==true and b.shape[0] % 8 == 0.

tests/gemm/test_mm_fp4.py
b8zhong: Sure. Added that.

flashinfer/gemm/kernels/utils.py
b8zhong: Good point.

b8zhong: Just removed it.

flashinfer/gemm/gemm_base.py
b8zhong: @Vinnie6167 I actually tested this at first. And just tested it again. Did it work for you?

When testing this locally, for (before swap) m,n = 16,130 will be wrong, but 16,128 will be correct. So I think it's actually still required (otherwise, it would require some change to TMA store)

leejnau (approved):

flashinfer/gemm/gemm_base.py
Vinnie6167: I see. I just tested myself as well. It is correct that we need N % 8 == 0, with or without swap AB. Thank you for checking.

bkryu (approved):

Issue #2490, opened 2026-02-04
[Bug] GDN prefill kernel produces NaN


ZJY0516:
[gdn_debug_inputs.pt.zip](https://github.com/user-attachments/files/25076044/gdn_debug_inputs.pt.zip)

I'm trying to use flashinfer gdn prefill kernel in vLLM, see https://github.com/vllm-project/vllm/pull/32846.

But I find it will always produce NaN

```python
import torch
from flashinfer.gdn_prefill import chunk_gated_delta_rule


def main():
    data = torch.load("gdn_debug_inputs.pt", map_location="cuda")

    q = data["q"]
    k = data["k"]
    v = data["v"]
    g = data["g"]
    beta = data["beta"]
    state = data["state"]
    cu_seqlens = data["cu_seqlens"]

    output, final_state = chunk_gated_delta_rule(
        q=q,
        k=k,
        v=v,
        g=g,
        beta=beta,
        initial_state=state,
        output_final_state=True,
        cu_seqlens=cu_seqlens,
    )
    print(output, final_state)


if __name__ == "__main__":
    main()

```

<details>
<summary>Output</summary>

```text
tensor([[[nan, nan, nan,  ..., nan, nan, nan],
         [nan, nan, nan,  ..., nan, nan, nan],
         [nan, nan, nan,  ..., nan, nan, nan],
         ...,
         [nan, nan, nan,  ..., nan, nan, nan],
         [nan, nan, nan,  ..., nan, nan, nan],
         [nan, nan, nan,  ..., nan, nan, nan]],

        [[nan, nan, nan,  ..., nan, nan, nan],
         [nan, nan, nan,  ..., nan, nan, nan],
         [nan, nan, nan,  ..., nan, nan, nan],
         ...,
         [nan, nan, nan,  ..., nan, nan, nan],
         [nan, nan, nan,  ..., nan, nan, nan],
         [nan, nan, nan,  ..., nan, nan, nan]],

        [[nan, nan, nan,  ..., nan, nan, nan],
         [nan, nan, nan,  ..., nan, nan, nan],
         [nan, nan, nan,  ..., nan, nan, nan],
         ...,
         [nan, nan, nan,  ..., nan, nan, nan],
         [nan, nan, nan,  ..., nan, nan, nan],
         [nan, nan, nan,  ..., nan, nan, nan]],

        ...,

        [[nan, nan, nan,  ..., nan, nan, nan],
[...]

Issue #3491, opened 2026-06-02
Megathread for overhaul GDN support with cutedsl


guangyunh-nv:
This issue tracks cutedsl overhaul for gated delta rule or GDN for flashinfer as well as latter improvement.

### Support matrix
|       | Fully Fused Prefill | CP Prefill 
|-------|---------------------|------------
| SM90  | <ul><li>[x] #3477</li></ul> | <ul><li>[x] #3481</li></ul>
| SM100 | <ul><li>[x] #3001</li></ul> | <ul><li>[x] #4078 </li></ul>
| SM120 | <ul><li>[x] #3479</li></ul> | <ul><li>[x] #3659</li></ul>

### Progress

- [x] #3477 
- [x] #3536 
- [x] #3478
- [x] #3479 
- [x] #3480 
- [x] #3481 
- [x] #3613
- [x] #3616 
- [x] #3659
- [x] #3788
- [x] #4133 
- [x] #4311 
- [ ] #4374

kahyunnam:
@guangyunh-nv can we close this? Feel free to re-open if there's something that's still a todo.

PR #3680, opened 2026-06-19
fix(gdn): add g_space parameter to chunk_gated_delta_rule

shbhm-ptl:
## 📌 Description

`chunk_gated_delta_rule` expects the forget gate `g` in **linear space**, but callers like vLLM pass `g = logsigmoid(raw_gate)` in ln-space for numerical stability, which produces all-NaN output with no actionable error.

This was reported and discussed in #2490 (closed). The kernel author identified the root cause as a gating-space mismatch and the reporter resolved their immediate issue with a `g=torch.exp(g)` workaround, but two commenters — including the kernel author — proposed the proper fix: a `g_space` flag so callers don't need to know the kernel's internal space convention. That fix was never implemented, so this PR follows through on it.

## 🔍 Related Issues

Related to #2490 (the suggested fix discussed there, implemented here)

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Reviewer Notes

Adds `g_space: str = "linear"` (default, fully backward-compatible). Valid values are `"linear"`, `"ln"`, `"log2"` — the API converts `g` to linear space once, before dispatch, so all three kernel backends (SM90/SM100/SM120) benefit automatically.

Test coverage in `tests/gdn/test_gdn_prefill_g_space.py`:
- Invalid `g_space` raises `ValueError` (hardware-independent)
- `g_space="ln"` output matches manual `torch.exp(g)` + `g_space="linear"`
- `g_space="log2"` output matches manual `torch.exp2(g)` + `g_space="linear"`
- Default `g_space="linear"` is unchanged from current behavior
- Regression test simulating the exact vLLM `logsigmoid` pattern from the original report

## Summary by CodeRabbit

* **Tests**
  * Added broader coverage for gate-space handling in attention-related functionality.
  * Verified invalid input values are rejected consistently.
  * Confirmed outputs remain stable and finite across supported gate-space options.
  * Added checks for backward compatibility between default and explicit settings.
  * Included regression coverage for log-space gate inputs on supported GPU hardware.


guangyunh-nv:
#3491 This should be addressed and ultimately unified to log2 space. This should be trivial for sm90 and sm120 kernels by modifying alpha processing logic.

PR #3682, opened 2026-06-20
fix(attention): add LEFT_SLIDING_WINDOW drain loop to FP8 Hopper prefill (#3578)

### LINKED ISSUE #3578: SM90 SWA + FP8 Q prefill hangs
With FP8 KV and sliding-window prefill (window_left=511), FlashInfer FA3/native prefill works when Q is BF16. If Q is FP8 (float8_e4m3fn), the kernel hangs for seq_len > 256 on H200/SM90. The Python call returns, but the following CUDA sync hangs

here's  a repro:

```
#!/usr/bin/env python3
"""Minimal FlashInfer repro for FP8-Q paged prefill hang.

Usage:

    python flashinfer_swa_fp8_prefill_min_repro.py 256  # expected to finish
    python flashinfer_swa_fp8_prefill_min_repro.py 257  # hangs on post-run sync

Observed on H200 / SM90. The FlashInfer wrapper resolves backend="auto" to
FA3 for this case. The key signal is that the pre-run CUDA sync completes,
wrapper.run(...) returns, and the post-run CUDA sync hangs for num_tokens >= 257.
"""

import sys
import time

import torch
import flashinfer


# Everything below is intentionally fixed; only num_tokens varies.
BACKEND = "auto"
KV_LAYOUT = "NHD"
WINDOW_LEFT = 511
PAGE_SIZE = 32
NUM_QO_HEADS = 64
NUM_KV_HEADS = 8
HEAD_DIM = 128
Q_DTYPE = torch.float8_e4m3fn
KV_DTYPE = torch.float8_e4m3fn
OUT_DTYPE = torch.bfloat16
WORKSPACE_BYTES = 128 * 1024 * 1024
INPUT_SCALE = 0.05
SEED = 0


def main() -> int:
    if len(sys.argv) != 2:
        print(f"usage: {sys.argv[0]} <num_tokens>", file=sys.stderr)
        return 2

    num_tokens = int(sys.argv[1])
    device = torch.device("cuda:0")
    torch.manual_seed(SEED)

    num_pages = (num_tokens + PAGE_SIZE - 1) // PAGE_SIZE
    last_page_len = num_tokens % PAGE_SIZE or PAGE_SIZE

    q_bf16 = (
        torch.randn(
            num_tokens,
            NUM_QO_HEADS,
            HEAD_DIM,
            dtype=torch.bfloat16,
            device=device,
        )
        * INPUT_SCALE
    )
    q = q_bf16.to(Q_DTYPE)

    kv_bf16 = (
        torch.randn(
            num_pages,
            2,
            PAGE_SIZE,
            NUM_KV_HEADS,
            HEAD_DIM,
            dtype=torch.bfloat16,
            device=device,
        )
        * INPUT_SCALE
    )
    k_bf16, v_bf16 = torch.chunk(kv_bf16, 2, dim=1)
    k_scale = (k_bf16.abs().amax().clamp(min=1e-12) / 256).item()
    v_scale = (v_bf16.abs().amax().clamp(min=1e-12) / 256).item()
    kv_cache = torch.cat(
        [(k_bf16 / k_scale).to(KV_DTYPE), (v_bf16 / v_scale).to(KV_DTYPE)],
        dim=1,
    )

    qo_indptr = torch.tensor([0, num_tokens], dtype=torch.int32, device=device)
    paged_kv_indptr = torch.tensor([0, num_pages], dtype=torch.int32, device=device)
    paged_kv_indices = torch.arange(num_pages, dtype=torch.int32, device=device)
    paged_kv_last_page_len = torch.tensor(
        [last_page_len], dtype=torch.int32, device=device
    )

    workspace = torch.empty(WORKSPACE_BYTES, dtype=torch.uint8, device=device)
    wrapper = flashinfer.BatchPrefillWithPagedKVCacheWrapper(
        workspace, kv_layout=KV_LAYOUT, backend=BACKEND
    )
    wrapper.plan(
        qo_indptr,
        paged_kv_indptr,
        paged_kv_indices,
        paged_kv_last_page_len,
        NUM_QO_HEADS,
        NUM_KV_HEADS,
        HEAD_DIM,
        PAGE_SIZE,
        causal=True,
        window_left=WINDOW_LEFT,
        q_data_type=Q_DTYPE,
        kv_data_type=KV_DTYPE,
        o_data_type=OUT_DTYPE,
    )

    print(
        f"num_tokens={num_tokens} resolved_backend={wrapper._backend} "
        f"q={tuple(q.shape)}/{q.dtype} kv={tuple(kv_cache.shape)}/{kv_cache.dtype} "
        f"window_left={WINDOW_LEFT} k_scale={k_scale:.8g} v_scale={v_scale:.8g}",
        flush=True,
    )

    torch.cuda.synchronize()
    print("after_pre_sync_before_run", flush=True)

    start = time.time()
    out = wrapper.run(q, kv_cache, q_scale=1.0, k_scale=k_scale, v_scale=v_scale)
    print(
        f"after_run_before_sync elapsed_s={time.time() - start:.6f} "
        f"out={tuple(out.shape)}/{out.dtype}",
        flush=True,
    )

    torch.cuda.synchronize()
    print(f"after_post_sync elapsed_s={time.time() - start:.6f}", flush=True)
    return 0


if __name__ == "__main__":
    raise SystemExit(main())
```

whycoming:
## Summary
Fixes #3578: FP8-Q sliding-window prefill deadlocks on Hopper (SM90).

## Root cause (confirmed by repro + source read)
The FA3 FP8 prefill consumer `mma_fp8` (`attention/hopper/quantization/mainloop_mma.cuh`) is missing the `LEFT_SLIDING_WINDOW` drain loop that the BF16 `mma_f16` has (`attention/hopper/mainloop_mma.cuh`, added in #673). The FP8 producer loads K/V down to `swa_begin`, but the consumer stops releasing pipeline stages at `swa_end + 1`, leaving the `pipeline_k`/`pipeline_vt` stages in `(swa_begin, swa_end]` unconsumed. The warp-specialized producer then blocks on `producer_acquire`: the launch returns and the following `cudaStreamSynchronize` never returns. It reproduces once the sequence crosses one KV tile past the window (empirically at seq_len 257 with the 256-wide tiling); the producer/consumer mismatch is window-independent in the source.

Reachable through the **ragged and paged** batch prefill wrappers, which pass `window_left` into the shared `mma_fp8` consumer. The single-prefill API guards FP8 with `assert window_left == -1`. fp16/bf16 are unaffected because `mma_f16` already has the drain loop (which is why it is used here as the correctness reference).

## Fix
Mirror the BF16 drain loop with the FP8-specific substitutions (`pipeline_vt`, `variant.PQuantize`, `convert_layout_acc_Aregs_fp8`, `permute_regs_A_to_C`), gated by `if constexpr (LEFT_SLIDING_WINDOW)` so non-SWA codegen is unchanged. The one consumer fix covers both batch paths.

## Verification (H20 / SM90, CUDA 12.8, torch 2.9.1)
Issue repro (paged, FP8 q/kv, `window_left=511`):

| seq_len | before | after |
|---|---|---|
| 256 | pass | pass |
| 257, 384, 512, 640, 1024 | hang | pass |

- FP8 output vs the bf16 (`mma_f16`) reference: MSE ~1e-6, cos 0.94-0.98 -- the **same agreement as the always-correct seq_len=256 baseline**, so the residual is FP8 P-requantization noise, not a correctness gap from the fix. Holds for real sliding-window cases (`swa_begin > 0`; window 128/256 at long seq).
- New `tests/attention/test_hopper_fp8_sliding_window.py`: 18 cases (ragged + paged x seq_len {257,512,1024} x window_left {128,256,511}) pass, under `pytest-timeout` so a regression surfaces as a failure instead of a CI hang.
- Existing non-SWA FP8 paged test (`test_batch_prefill_paged_gqa`) still passes -- the `if constexpr` gate leaves non-SWA codegen unchanged.
- pre-commit formatters clean: clang-format 19.1.1, ruff 0.12.8.

## Notes
- New test is a separate file for focus; happy to fold it into `test_hopper_fp8_attention.py`.

## Release Notes

* **Bug Fixes**
  * Fixed FP8-quantized attention computation with sliding window support on Hopper GPUs
  
* **Tests**
  * Added regression tests for FP8 sliding window prefill operations (ragged and paged batches)


whycoming:
Hi @bkryu, thanks for running the internal `tests/attention` pipeline on this PR 🙏

As an external contributor I can't access `nv/flashinfer-ci`, so I only see `#57024192: 8/20 passed` with no logs on my end. One data point that might help triage: on the **public** `JIT Unittest (H100)` job, the new `tests/attention/test_hopper_fp8_sliding_window.py` is collected and **passes 18/18** — and that's SM90/Hopper, the exact architecture this fix targets. So the drain-loop fix and its regression test look green where they're meant to run.

Given that, could you help confirm whether any of the failing internal jobs are actually in **this PR's test or the `mma_fp8` change** — versus pre-existing / unrelated failures (a different arch, or flaky jobs)? If they're unrelated I think this is ready; if any are on me, just point me at the failing test names + error type (timeout vs. assertion) and I'll fix right away. Thanks again!

bkryu (approved):
Hi @whycoming thanks for the contribution and trying to check the. 

The internal CI looks good. Failures are from some recently starting failures we are trying to fix  😄.

PR #3684, opened 2026-06-22
feat(attention): asymmetric VO-split NVFP4 paged prefill (qk=512/vo=256) for Gemma-4 on SM120/121

jethac:
## Summary

Add an **asymmetric VO-split NVFP4 paged-prefill path** to the FA2 attention kernels so consumer/SoC Blackwell (SM120 / SM121) can serve models whose global-attention layers are **512-wide** — specifically **Gemma 4** — with a 4-bit (NVFP4) KV cache.

Upstream NVFP4 paged prefill is symmetric `head_dim ≤ 256` only. Gemma 4's full-attention layers are `head_dim_qk = 512, head_dim_vo = 256` (value reuses key), which hits three walls: the equal-K/V-stride host check, the VO-split `static_assert(!is_fp4)`, and the output-width hardcode. This PR adds the FP4 V-load + explicit scale-factor strides + asymmetric K/V offsets so the kernel serves the asymmetric `(qk=512, vo=256)` shape directly, and keys `CTA_TILE_Q` on `head_dim_qk` so the dispatch is feasible at 512.

## What's in it

- **NVFP4 KV-cache support for FA2 paged attention (SM120)** — the FP4 V-load into the VO-split path, the `[all-data | all-SF]` contiguous scale-factor layout (the trtllm-gen per-page swizzle doesn't commute with head-dim slicing), and the asymmetric K/V byte/scale offsets.
- **SM121 (DGX Spark / GB10) dispatch** + a heuristic regression test.
- **`CTA_TILE_Q` keyed on `head_dim_qk`** for the asymmetric `(QK=512, VO=256)` case (the dispatcher otherwise picks an infeasible tile / `max_mma_kv: 0`).
- **`plan()`: move `mask_indptr` to the custom mask's device before `segment_packbits`** — a standalone correctness fix (the paged + ragged wrappers passed a CPU `mask_indptr`).

## Validation

- **Numerics (RTX 5090, SM120):** `qk=512 / vo=256` paged NVFP4 prefill `max_abs_err ≈ 0.0047` (the FP4 e2m1 rounding floor itself is ~0.0156); output correctly sized by `vo=256`; symmetric `qk=vo=256/128` cases unregressed; multi-tile seq 512/1024 + causal all sane.
- **End-to-end (via vLLM):** the full Gemma 3 + Gemma 4 family serves a calibrated NVFP4 KV cache through this kernel on **both** SM120 (RTX 5090 / RTX PRO 6000) and SM121 (GB10) — facts match `bf16`. (vLLM side: companion PR linked below.)

## Notes
- The CC 12.0 / 12.1 findings reproduce identically on an RTX PRO 6000 and a GB10 — one fix serves the 5090, the PRO 6000, the Spark, and the upcoming RTX Spark.
- Companion vLLM PR (serving orchestration + the two-pass VO split + built-in calibration) is forthcoming; I will cross-link it here once opened.
- Happy to split the standalone `mask_indptr` device fix into its own PR if preferred.

## NVFP4 split-KV correctness (added)

While validating this kernel under SGLang's radix cache (and as a heads-up for vLLM's
automatic prefix caching), I found a real correctness bug in the **NVFP4 split-KV**
path -- distinct from the VO-split work, and shared by every consumer of this kernel.

**Symptom.** With a 4-bit KV cache and prefix reuse on, retrieval cliffs: a Gemma-4
E2B/E4B needle is retrieved perfectly up to ~550 tokens of *reused-prefix* context,
then collapses (degenerate repetition / `<pad>`) from ~680 tokens on. `bf16` KV has no
such cliff. The tell: `--disable-radix-cache` (any cold, contiguous prefill) makes it
vanish.

**Root cause.** Not page-table fragmentation, not calibration (both ruled out by an
instrumented A/B). It's the **extend geometry** -- a short query window attending a
long *cached* KV (`qo_len << kv_len`) -- which makes the FA2 scheduler pick
**split-KV** (flash-decoding). NVFP4 KV is packed `uint8` with a per-16-element FP8
block scale; `kv_chunk_size` is not aligned to those blocks, so a split boundary can
land mid-block, and the small per-split chunk also trips the 1-byte-KV `NUM_MMA_KV`
tile floor. A position-aligned dense-cache trace shows the final query token's layer-0
attention output drift ~0.1-1% under split-KV vs the cold full-prefill for an
*identical* query -- small, but it compounds over ~30 layers until small models lose
the needle. A full-prefill (`qo_len == kv_len`) never splits, which is why dense-prefill
tests stay green while prefix caching / long-context decode break.

**Fix (this push).** Gate split-KV off when `kv_data_type` is NVFP4 (`uint8` /
`float4_e2m1fn_x2`) in both the paged and ragged prefill `plan()`; FP8 / 16-bit KV are
untouched. ~5 lines, provably correct, at a small decode/short-query parallelism cost
**for 4-bit KV only**. Adds `test_nvfp4_split_kv_gate_dtype_logic` (hardware-free
dtype-classification guard).

**Verified.** Gemma-4 E2B/E4B radix-on retrieval holds to 1448 tokens (== `bf16`) with
the gate vs cliffing at ~600 without it -- on both **sm120** (RTX PRO 6000) and
**sm121** (GB10 / DGX Spark). The full Gemma-4 multimodal matrix (text/image/audio,
nvfp4 vs bf16) is green across E2B/E4B/12B/26B-A4B/31B with radix caching **on**.

**Follow-up.** A perf-preserving version keeps flash-decoding by making FP4 split points
16-token-aligned (respecting the scale blocks) and clearing the `NUM_MMA_KV` floor per
chunk -- left as a follow-up so the correctness fix can land now.

## Release Notes
- **New Features**
  - Improved paged/prefill NVFP4 KV support with separate K/V stride handling and NVFP4 scale-factor stride wiring, including generated JIT parameter setup.
  - Enhanced FP4 GEMM auto-selection for SM12x (e.g., SM120/SM121) to prefer `b12x` first.
- **Bug Fixes**
  - Fixed NVFP4 decode/prefill output sizing and KV stride/softmax-factor loading consistency.
  - Prevented uninitialized `lse` values in NVFP4 SM120 attention and corrected KV attention tiling logic to use Q/K dimensions.
- **Documentation**
  - Updated source-install guidance and CUDA-arch examples (including `12.1a`).
- **Tests**
  - Added coverage for SM121 NVFP4 backend selection and JIT attention stride/SF generation.

## Build-size note (per review)

`DISPATCH_GQA_GROUP_SIZE` gains a `group_size == 6` case (needed for Gemma-4-class GQA ratios), which adds one instantiation per macro user - roughly +20% compile time/binary size for the translation units that expand this macro.


jethac:
A heads-up for reviewers before you spend time here:

I'm extending this work to add **multimodal** support for Gemma 3/4, which I erroneously omitted from the initial version. Gemma 3/4 are natively multimodal, and serving image (and, for Gemma 4, audio) inputs through the NVFP4 VO-split prefill requires the model's span-level bidirectional image masking to flow through the kernel's custom-mask path. Without it the contribution only covers text prompts, which isn't a complete Gemma 3/4 claim.

I'm validating the multimodal path now and will push the update shortly. I'd appreciate holding final review until it lands — apologies for the churn.

jethac:
Update — the multimodal support has landed, on the vLLM side: vllm-project/vllm#46329 (commit *"Gemma 3/4 multimodal: NVFP4 KV mm-prefix span masking on FA2"*). It serves image-token spans bidirectionally via FlashInfer's existing FA2 packed `custom_mask` path, and the only FlashInfer-side requirement — moving `mask_indptr` onto the custom mask's device when the mask is on GPU while the indptr arrays are on CPU — is already included in **this** PR. So #3684 needs no further changes for multimodal.

Validated reading images back under NVFP4 KV, matching bf16: `sm120` Gemma-3-4B + Gemma-4-31B (VO split), `sm121`/GB10 Gemma-3-4B + Gemma-4-E4B (VO split). Thanks for the patience on the churn.

jethac:
Thanks for the reviews. Pushed one fix (7f549fe0) and notes on the rest:

- **`flashinfer/decode.py` (out_head_dim)** — fixed: the output-width doubling is now gated on `v_cache.dtype == torch.uint8`, not just `kv_cache_sf is not None`, so a stray scale-factor tensor on a non-uint8 cache fails the shape check instead of silently mis-allocating.
- **`flashinfer/cute_dsl/utils.py` (cute import)** — `cutlass.cute` is already imported at the top of the file (`import cutlass.cute as cute`, line 25); this PR only adds the `OperandMajorMode` shim beneath it, so there's no missing import. No change needed.
- **`flashinfer/jit/attention/modules.py` (FP4 gating in `gen_customize_batch_attention_module`)** — this PR routes NVFP4 KV through `gen_customize_batch_prefill_module`, which carries the `require_fp4_kv_cache` guard; it does not add FP4 wiring to the batch-attention generator, so that path isn't FP4-reachable from this change. Happy to add the same guard there as a separate hardening if you'd prefer.
- **`include/flashinfer/attention/prefill.cuh` (CTA pruning, QK vs VO)** — intentional for the asymmetric VO-split: `FA2DetermineCtaTileQ` selects `CTA_TILE_Q` from `head_dim_qk`, so the first prune clause keys on QK while the MMA/smem clauses key on VO. Verified on sm120 at qk=512/vo=256.

jethac:
Pushed a follow-up commit (`8984eb3`): **disable split-KV for NVFP4 KV** in the prefill `plan()`.

Found this validating the kernel under prefix caching. NVFP4 split-KV (flash-decoding) corrupts reads when a short query attends a long *cached* KV (`qo_len << kv_len`): a split boundary lands mid 16-element scale block, and the small per-split chunk also trips the 1-byte-KV `NUM_MMA_KV` floor. It's invisible to dense full-prefill tests (which never split) but cliffs prefix-cached retrieval / long-context decode -- Gemma-4 E2B/E4B drop the needle past ~600 tokens of reused context while `bf16` holds. Gating split-KV off for FP4 `kv_data_type` (`uint8` / `float4_e2m1fn_x2`) restores parity out to 1448 tokens, validated on **sm120 (RTX PRO 6000)** and **sm121 (GB10 / DGX Spark)**. FP8 / 16-bit KV are untouched.

Worth flagging for the companion vLLM PR and anyone else consuming this kernel: since the fix lives in `plan()`, every consumer inherits it -- but **without it, any engine running NVFP4 KV with prefix caching shares this bug**. Full root-cause + the position-aligned dense-cache trace numbers are in the PR description under "NVFP4 split-KV correctness". The perf-preserving version (scale-block-aligned FP4 splits) is left as a follow-up so the correctness fix can land now.

jethac:
Thanks @coderabbitai — verified all five against the current tree and pushed `a1975a5`. Verdict on each:

**1. `init_rope_freq` bound `NUM_MMA_D_VO/2` → `NUM_MMA_D_QK/2` (Critical).** Real. `rope_freq` is declared `[NUM_MMA_D_QK/2][4]` and the rotary appliers index it up to `NUM_MMA_D_QK/2`, so the asymmetric VO-split dispatch (qk=512, vo=256) that this PR makes reachable was leaving the upper half uninitialized. Fixed. Note our own runtime path applies RoPE before attention (`pos_encoding_mode=NONE`), so `init_rope_freq` isn't on it — this only bit in-kernel-RoPE callers of the asymmetric kernel, but it's a correctness bug regardless.

**2. Guard the deswizzle smem store (Major).** Fixed, guarded on `flat_byte < SF_TOTAL_BYTES` rather than `in_bounds` so the in-buffer padding-row zero-fill is preserved. For context: this branch is behind `FLASHINFER_PAGED_V_SF_DESWIZZLE`, which defaults to `0`, so it's compiled out of shipped builds; the live `cp_async` store below was already predicated. Latent, but the guard is correct and free.

**3. Include FP4 SF smem in the `NUM_MMA_KV` budget (Major).** Real and on the FP4 path — `kKVSmemPerMmaKV` counted KV data + repack (+ VO-split `p_smem` at the paged site) but not `k_sf_smem`/`v_sf_smem`, which scale with `CTA_TILE_KV`. Added the term at all three dispatchers. This is safe-by-construction: a larger per-tile cost only *lowers* `max_num_mma_kv_smem`, and any tile that actually fits in smem (every validated config, since it runs) stays selectable — so it can't regress a working config, only stop the dispatcher from over-selecting a tile whose `SharedStorage` would exceed the limit. Recompiling the JIT + an sm120 smoke before this merges, since it touches live dispatch.

**4. Gate `out_head_dim` doubling on the V-cache dtype, not `kv_cache_sf` (Major).** Fixed at all three prefill sites — gated on `kv_cache_sf is not None and v.dtype == torch.uint8`, matching the decode-side guard. (Kept the `kv_cache_sf` half too: doubling should require the scale factors to be present *and* the packed uint8 storage.)

**5. Mirror the `mask_indptr` device fix in the ragged custom-mask path (Major).** Fixed — `mask_indptr.to(custom_mask.device)` before `segment_packbits`, same as the paged path. Drive-by since it's the identical one-liner; happy to split it out if you'd rather keep the diff scoped to the NVFP4 changes.

jethac:
Smoke for `a1975a5` on real sm120 (RTX PRO 6000 Blackwell, CC 12.0, CUDA 13.0): branch built from source, JIT-recompiled, full nvfp4 suite green.

```
tests/attention/test_nvfp4_attention_sm120.py ......  (8 passed in 135.28s)
  accuracy[s128-d64 / s256-d128 / s4096-d64 / s4096-d128 / s8192-d64, causal+noncausal]
  causal_mask_column_order
  split_kv_gate_dtype_logic
```

The 135s includes the JIT recompile, so the budget/rope/smem-guard edits compile and the nvfp4 prefill numerics are unchanged. (The unit suite covers symmetric head_dim 64/128; I'm running a Gemma-4-12B end-to-end pass next to exercise the asymmetric 512/256 VO-split path through the new budget term.)

jethac:
Correction + fuller validation on sm120 (RTX PRO 6000 Blackwell, CUDA 13.0). My earlier "smoke" note was imprecise: `test_nvfp4_attention_sm120.py` exercises the standalone `nvfp4_attention_sm120` kernel, **not** the `prefill.cuh` dispatchers this PR's `a1975a5` actually edits, so it didn't compile-test those changes. Re-ran against the right targets:

**1. nvfp4 paged prefill (the path `a1975a5` touches) — green.**
`test_batch_prefill_with_paged_kv_cache_nvfp4` (32 params: batch×kv_len×qo_len×page_size×{fp16,bf16}) all pass. This JIT-compiles the edited `BatchPrefill*Dispatched` translation unit (budget term + rope + smem guard all build) and exercises the `out_head_dim` doubling on the real paged path.

**2. Found and fixed a separate uninitialized-read in `nvfp4_attention_sm120` (`d1fd883`).** The `fwd` entry point only zeroes `lse` in the `seq_len==0` branch; the normal path passes the caller's `torch.empty` lse straight to the kernel, which writes `out` for every row but not every `lse` entry → flaky NaNs depending on allocator state (passed alone, failed when run after the paged test dirtied the pages). Fixed by zeroing `lse` before launch, mirroring the empty-seq branch. Re-ran the dirty ordering: **40/40 deterministic**, including the case that flaked.

**3. Gemma-4-12B end-to-end on the 512/256 VO-split path — full retrieval.** Served `google/gemma-4-12b-it` nvfp4 (radix-on, FlashInfer backend) with this branch's `prefill.py` + `prefill.cuh` compiled in (confirmed the budget term is in the served kernel), needle sweep 168→1448: every context hits (`ans=49173`). The one miss is the cold first-request JIT compile timing out, not a quality miss — the next request at a larger context retrieves cleanly. So the asymmetric VO-split dispatch is bf16-faithful with the new budget/`out_head_dim` code live.

qsang-nv:
Thanks for this work. A related set of PRs (#3739, #3740, #3741, and #3748) is now near the end of review and is expected to land first. In particular, #3739 and #3748 overlap with this PR in the FA2 large-head/NVFP4 prefill and scale-stride paths, so merging them will create conflicts in several core attention files.
To avoid reviewing code that will shortly need substantial conflict resolution, could you please rebase this PR onto main after those PRs land and reconcile the overlapping implementations? We will review the rebased version afterward, with particular attention to the remaining asymmetric (head_dim_qk=512, head_dim_vo=256) and SM120/SM121-specific changes.

jethac:
Thanks @qsang-nv — that makes complete sense, and I'd much rather rebase onto the finished versions than have you review code that's about to need conflict resolution.

Understood on the sequencing: with #3739 and #3741 merged and #3748 / #3740 / #3822 landing now, I'll rebase #3684 onto main once they're in and reconcile the overlapping prefill paths onto @lesj0610's implementations rather than my own. That should reduce the diff to the asymmetric `head_dim_qk=512` / `head_dim_vo=256` VO-split and the SM120/SM121-specific pieces you flagged.

@lesj0610's large-head NVFP4 work is excellent. Splitting it into focused pieces (the paged NVFP4 KV dequant in #3748, slot-mapping append in #3740, workspace sizing in #3741) is a cleaner decomposition of this path than my single PR, and the vectorized V-scale dequant in #3822 (four scales as one `uint32` + `fast_dequant_f8f16x4`) is a genuinely nice optimization — I'll be glad to build the VO-split on top of it. Reconciling onto their versions will make this PR smaller and better.

For context, this VO-split is also the base for a native `mma.sync.kind::mxf4nvf4.block_scale` QKᵀ I've been validating on SM120/121 — happy to bring it as a stacked PR once this lands, if useful.

I'll ping the thread when the rebase is up. Thanks again.

jethac:
For context on where this PR is headed — I've written up the campaign it's the foundation for: **A4Q**, a native `mma.sync.kind::mxf4nvf4.block_scale` QKᵀ that consumes the NVFP4 KV cache directly (zero dequant) on SM120/121. The asymmetric `qk=512 / vo=256` VO-split in this PR is exactly what lets Gemma-4's wide heads reach that path on the tensor core instead of dropping to a software fallback — so the kernel stacks straight on top of it.

The write-up has the measured numbers, the bit-exact cert methodology, and an honest accounting of where the end-to-end win lands (and where it washes to nothing):

https://x.com/jetha/status/2073322454198649215

Next step here is unchanged — rebase onto the overlapping NVFP4-paged-KV PRs once they land, then the A4Q kernel goes up as a stacked PR. Glad to take any questions in the thread.

qsang-nv:
All related PRs are merged, you can start rebasing now @jethac.

TechPrototyper:
Cross-reference: we filed a detailed corruption report for the combined stack (this branch @ `d1fd883` + vllm-project/vllm#46329 @ `bd1e159`) over in the vLLM PR: https://github.com/vllm-project/vllm/pull/46329#issuecomment-4918945200

TL;DR: on SM120 (RTX 5090) with a hybrid-attention Qwen3.6-27B, nvfp4 KV deterministically corrupts every second request (fp8 on the identical build is clean); bisection rules out prefix caching, CUDA graphs, VOSPLIT, and block size — and the 06-25 split-KV chunk-boundary fix from this branch was already included in our build. We also observed write/read-inconsistent behavior for KV scales ≠ 1.0. Full environment, bisection matrix, and repro details in the linked comment.

TechPrototyper:
Following up on my report from 2026-07-08: the every-second-request corruption on hybrid models (sm120) turned out to be **not a FlashInfer bug** — the FA2 paged nvfp4 reader computes everything correctly, including the SF addressing via the explicit per-tensor strides this branch introduced.

Root cause was on the vLLM side: the global `[all-data | all-scales]` KV page layout (built around the assumption that the reader derives the SF stride as `data_stride / 8`, which is no longer true on this branch) made logical KV blocks non-byte-contiguous, breaking vLLM's hybrid-model KV tensor sharing — GDN state writes overlapped the attention FP4 data region. Full analysis and fix (switching sm12x to the per-page `[data | scale]` layout, consumed by this reader via its SF-tensor strides): https://github.com/vllm-project/vllm/pull/46329#issuecomment-4927977267 and https://github.com/jethac/vllm/pull/1.

Verified on RTX 5090 with a hybrid Qwen3.6-27B: clean across all probes, 441k KV tokens at util 0.97 (+58% vs fp8 KV).

jethac:
Thanks @qsang-nv — rebased onto current main (3fd5c55b) as requested; the branch is updated (head `785d0e01`, 10 commits).

**Reconciliation with the merged NVFP4 series (#3739 / #3741 / #3740 / #3748 / #3822).** The overlap resolved cleanly in main's favor and the diff shrank from 827+/156− across 27 files to 648+/115− across 26. Where main already had an equivalent, ours was deleted: the explicit SF-stride generator machinery went in favor of #3748's `GetFP4ScaleStrides` + static `k/v_sf_stride_*` params; the shared-V-smem offset stash went in favor of #3739's `page_produce_kv_on_the_fly`; our dispatcher smem-reject macro and coarser SF budget term went in favor of #3739's budgeting (which is a more precise superset — it carries the `NUM_WARPS_KV` factor ours lacked); and our single/ragged explicit SF strides were reverted to match #3748's deliberate paged-only scoping. What remains is the asymmetric-head substance this PR is about: separate K/V stride families in `paged_kv_t` (`protective_get_{k,v}_offset`), CTA_TILE_Q dispatch keyed on `head_dim_qk` (utils.cuh heuristic, `KernelTraits::IsInvalid`, jinja instantiation lists, scheduler threading rebased onto #3741's `PrefillPlanImpl`), the `init_rope_freq` bound fix (`NUM_MMA_D_QK/2`), the NVFP4 split-KV disable at plan(), uint8-gated NVFP4 output sizing, SM121 (GB10) enablement, the fp4x2 vector casts, and the lse-zeroing fix in the sm120 binding.

**Validation (DGX Spark, sm121, throwaway venv, fresh JIT cache).** Upstream NVFP4 test batteries on the rebased branch: **486 passed, 0 failed** — `test_batch_prefill_kernels -k nvfp4` (56, incl. large_head/rope/strided-scale-views), `test_fp4_kv_quantization` (54, incl. asymmetric dk64/dv128), `test_page -k nvfp4` (70 + 1 skip), `test_workspace_size` (6), `test_single_prefill -k nvfp4` (10), `test_batch_decode_kernels -k nvfp4` (289), plus the sm120-gated suite (1 + 7 arch skips, identical to pre-rebase). End-to-end with the qk512/vo256 VO-split active (Gemma-4-12B, calibrated NVFP4 KV): fidelity 8/8, 8k-context retrieval margin 16/16 argmax recall with mean margin within 0.007% of the pre-rebase baseline. One test in our own earlier commit asserted a gate string the generator never emitted (`test_attention_utils`); the head commit fixes the assertion to check the gate that is actually emitted (test-only change).

**Follow-up worth flagging (from #3740, not blocking this PR).** The NVFP4 slot-mapping append kernels carry fully separate K/V data and SF strides, but a single `HEAD_DIM` template parameter drives both the K and V quantize loops, and the host wrapper derives it from the single `paged_kv_t::head_dim` member (`head_dim = paged_kv.head_dim * 2`). The append helper therefore can't write an asymmetric (K=512, V=256) VO-split cache. No impact here — this PR is the read path and vLLM performs its own cache writes — but any future user combining #3740's append with the VO-split layout will hit it silently. Happy to send a small follow-up PR parameterizing the K/V loops separately if that's welcome. Relatedly, a one-line polish I'd suggest (can fold in or follow up): template `get_paged_kv_offset_for_logical_row` on `produce_v` so the shared-smem path routes through `protective_get_{k,v}_offset` instead of addressing V with K strides (today unreachable for asymmetric/FP4 shapes, but no longer guarded by a host-side stride-equality check).

Also — thanks @TechPrototyper for the sm120 hybrid (Qwen3.6-27B) stress report and especially the follow-up confirming the FA2 paged NVFP4 reader computes correctly, including the SF addressing via the per-tensor strides this branch introduces. The KV-layout question your fix addresses is vLLM-side; happy to continue that thread on vllm#46329.

For context: this VO-split path is the base for a follow-on I've been validating — a native `mma.sync.kind::mxf4nvf4.block_scale` QKᵀ that reads the NVFP4 KV directly (zero dequant) on SM120/121, compute-sanitizer-clean and bit-identical across sm120/sm121. Happy to open it as a stacked PR once this lands, if useful.

lesj0610:
@jethac Thanks for the detailed update. On the #3740 follow-up: that PR was scoped to the existing symmetric paged-KV contract. The slot-mapping variant added independent K/V data and SF strides for addressing, but not independent K/V head dimensions. You're right that separate K/V dimensions are needed before the append path can support a QK=512 / VO=256 layout.
On the safety point, an asymmetric combination passed through #3740's public API isn't mishandled silently. It's rejected up front by the host-side dimension and shape checks. The unsafe case would arise only if those checks were bypassed or later relaxed. A focused follow-up adding separate K/V dimensions and asymmetric tests sounds right to me. For get_paged_kv_offset_for_logical_row, I'd address the produce_v routing in #3684, since that's where the separate K/V stride families are introduced.

TechPrototyper:
Thanks @jethac — glad the reports were useful, and great to see the rebase onto main. Offer: we can validate the rebased head (`785d0e01`) on our SM120 setup (RTX 5090, Qwen3.6-27B hybrid — 16 full-attention + 48 GDN layers, NVFP4 KV in continuous production use). We'd rebuild against the new SHA and run our acceptance matrix (repeated-request consistency, long-generation, 260k-context fills, throughput) and report results here within a day or two. If there are specific paths you or @qsang-nv want exercised on consumer Blackwell (e.g. the asymmetric K/V append follow-up from #3740), happy to include those.

TechPrototyper:
As promised — validation results for the rebased head (`785d0e01`, v0.6.14) on consumer Blackwell:

**Setup:** RTX 5090 (SM120), driver 595.58.03, CUDA 13.0.2, torch 2.11.0+cu130. FlashInfer built from this branch @ `785d0e01`, JIT-only (packaging note: `flashinfer-cubin` 0.6.14 isn't on PyPI yet, so we dropped cubin for the build — the import-time version check fails otherwise). vLLM from jethac/vllm @ `9c8a79ac0` (nvfp4-fa2-consumer-blackwell incl. the hybrid-sharing fix). Model: Qwen3.6-27B hybrid (16 full-attn + 48 GDN layers), NVFP4 KV, block 32, util 0.97, max-model-len 262144. We also exercise the FA2 verify path via MTP speculative decoding (k=2).

**Results — all green, no regressions vs. our previous build (`d1fd883`):**
- 20/20 byte-identical greedy completions (repeated-request consistency — the original corruption reproducer)
- **255,823-token context fill** at 1,839 tk/s prefill, correct summarization referencing full-depth content
- 52k prefill at 5,109 tk/s; decode 110 tk/s with MTP (72.9 without)
- vision + tool-calling checks clean; KV pool 416,933 tokens
- sm12x JIT compile from a cold cache: ~4 min on first start, stable afterwards

From the SM120-hybrid corner this head looks release-ready. Happy to re-run on future revisions — the matrix is automated on our side.

qsang-nv:
Thanks for the careful rebase — reconciling onto the merged NVFP4 series in main's favor and shrinking the diff to the asymmetric substance is exactly what we asked for, and the validation on this head is strong (your SM121 batteries + Gemma-4 e2e, plus @TechPrototyper's independent SM120 run). On our side I ran `tests/attention/test_batch_attention.py -k nvfp4` on an H100 at `785d0e01`: **32 passed** — the touched persistent NVFP4 path is covered for the tested equal-stride cases, so no need to re-run that battery yourself.

Remaining items before this is mergeable:

**[Blocker] `decode.cuh` still addresses V with K strides, and the host guard is gone.** `batch_decode.cu` now accepts independent K/V strides (the per-dim equality ICHECK was removed), but the decode kernel computes a single `kv_offset[j]` via `protective_get_kv_offset` (K strides) and uses it for both `paged_kv.k_data +` and `paged_kv.v_data +` (decode.cuh ~486/533). A caller passing genuinely separately-strided K/V pools gets silently corrupted V reads where it used to get a loud error. Either make the decode kernels V-stride-aware, or keep the stride-equality check in `batch_decode.cu` until they are.

**[Blocker] `get_paged_kv_offset_for_logical_row` produce_v routing.** As you noted (and @lesj0610 seconded doing it in this PR, since the separate stride families originate here): the shared-smem path still routes V through `protective_get_kv_offset`. Please fold in your own suggested fix, e.g.

```cpp
if constexpr (produce_v) {
  return paged_kv.protective_get_v_offset(...);
} else {
  return paged_kv.protective_get_k_offset(...);
}
```

Same rationale as above: this PR legalizes unequal K/V strides, so every consumer reachable from the updated entry points must either support them or reject them explicitly.

**[Medium] SM121 `mm_fp4` default flip to `b12x`.** This changes backend selection for *all* SM121 users of `mm_fp4(backend="auto")`, and the previous code deliberately excluded SM121 ("cutlass/cudnn are faster in most cases"). Please provide GB10 benchmark data justifying the flip, narrow the policy change, or move it into a separately justified PR. Also please reword the project-specific comments (`the campaign's VO-split / NVFP4 KV path needs...`, `NOTE(spark-hijinks)`) into neutral technical rationale.

**[Medium] The smem probe in `FA2DetermineCtaTileQ`.** Three issues: (1) the `(512,256)` example in the probe's comment is unreachable — the preceding `qk >= 512` branch returns early. Please replace it with an actually reachable configuration that requires this fallback, and ideally add coverage for that configuration; for the commonly tested dimensions up to 256, the probe appears not to change the selection. (2) It checks `cudaDevAttrMaxSharedMemoryPerMultiprocessor`, while the dispatcher-side guard uses `MaxSharedMemoryPerBlockOptin` — please use the per-block opt-in limit so the two budgets can't disagree near the boundary. (3) "sizeof assumed 2 … only configurations that could not run at all are moved" is inaccurate for 1-byte KV: the 2× overestimate can move a *working* fp8/fp4 config from CTA16 to CTA64 (a perf change, not an availability one).

**[Minor]**
- The split-KV gate itself is a reasonable conservative fix, but the root-cause text doesn't hold together: scale blocks run along `head_dim` (16 elements each), while split-KV partitions along the *token* axis, so a chunk boundary can't land "mid-block". The empirical evidence is convincing — please reframe the description as an empirical workaround and pin the actual mechanism (the 1-byte-KV `NUM_MMA_KV` tile floor looks like the more plausible culprit). Also worth checking whether decode's own partition-KV path shares the issue for NVFP4.
- `DISPATCH_GQA_GROUP_SIZE` gaining group size 6 adds one more instantiation to every kernel using the macro (~+20% for those TUs) — fine, but please mention it in the PR description.
- The new scalar-appending logic in `run()`: a declared-but-unprovided scalar outside the fixed `jit_scalar_values` dict raises a bare `KeyError`; the `scalar_start` slicing is also hard to follow. Consider simplifying and raising a named error.
- Nit: the jinja lists instantiate `CTA_TILE_Q=32` for `qk>=512 && vo<512` even though the heuristic never selects it and `IsInvalid()` prunes the body — dropping 32 there would remove the dead dispatch wrapper.

TechPrototyper:
@jethac — two offers from the SM120 side, whichever is useful:

1. **Same-day validation** of the stride-related fixes once you push them — our acceptance matrix (hybrid Qwen, paged NVFP4 KV, ~1h turnaround incl. build) exercises exactly the paged read/write paths with separate K/V stride families.
2. If it helps, we can run the **decode partition-KV check** @qsang-nv mentioned on our RTX 5090 — split-size sweep in isolation, results reported here either way. Happy to share any repro data that comes out of it in case it's useful for the split-KV gate's description, too.

(GB10/SM121 benchmarking we unfortunately can't help with — no SM121 hardware here.)

jethac:
Thanks @qsang-nv for the thorough review, and for running the H100 battery — appreciated. All items are addressed; the branch is updated (head `cab25c90`, 8 commits on top of the previously reviewed `785d0e01`, no history rewritten). Point by point:

**decode.cuh addressing V with K strides (blocker).** Took the host-side check you offered as option B: `batch_decode.cu` now ICHECKs per-dimension K/V stride equality before constructing `paged_kv_t`, with an error message naming the limitation (the decode path does not yet support asymmetric K/V strides). The constructor call keeps independent stride arguments, which is equivalent once equality is enforced. The check is now also exercised: a new negative test constructs genuinely unequal K/V stride families and asserts the ICHECK fires with the stated message, alongside an equal-stride positive control that runs and matches the decode reference — so the rejection contract is tested, not just present. (`d77e0238`, test in `cab25c90`)

**produce_v routing in prefill.cuh (blocker).** `get_paged_kv_offset_for_logical_row` now takes a `bool produce_v` template parameter and routes via `if constexpr` to `protective_get_v_offset` / `protective_get_k_offset`; both call sites (the only two — grep-verified) pass `<produce_v, KTraits>`. This closes the shared-smem path that addressed V with K strides. The commit also credits @lesj0610, whose shared-smem produce path this touches. (`675746dd`)

**SM121 `mm_fp4` backend-ordering default (medium).** Withdrawn from this PR. `flashinfer/gemm/gemm_base.py` and `tests/gemm/test_mm_fp4.py` are reverted to match main exactly (`git diff main` on both files is empty), and the heuristic test asserting the flipped ordering is removed; `b12x` remains selectable as an explicit backend. I'll propose the default flip as a separate PR backed by GB10 benchmark data, so it can be evaluated on its own merits.  (`9c3d67da`)

**Project-specific comments (medium).** Removed — the tagged notes are neutral now, and the diff against main contains no remaining project-specific lines. (`a40d2c63`, `9c3d67da`)

**Shared-memory fallback comment + probe (medium).** Three parts:
1. The unreachable (512,256) example is replaced with the true reachable worst case: (256,256) at 2-byte KV = 16·256·2 + (256+256)·16·4·2 = 72KB. Since that fits the opt-in limit of every current CC≥8 part (≥99KB), the comment now states explicitly that the fallback is a forward-looking guard mirroring the dispatcher's check, rather than something any current config triggers — I didn't want to invent a failing config that doesn't exist.
2. The probe now queries `cudaDevAttrMaxSharedMemoryPerBlockOptin`, which is the binding term of the dispatcher's `min(per-SM/num_ctas, opt-in)` in the tight `num_ctas=1` case where its guard fires.
3. The 1-byte-KV overestimate is fixed at the source: `sizeof(DTypeKV)` is visible at TU scope in `batch_prefill.cu`, so `kv_dtype_bytes` is threaded as a defaulted trailing parameter through `FA2DetermineCtaTileQ` / `PrefillSplitQOKVIndptr` / `PrefillPlanImpl` / `PrefillPlan` / `PrefillPlanWorkspaceSize`, passed at both the plan and workspace-size entry points (kept mutually consistent) and at single-prefill dispatch. The comment records the remaining approximations (2-byte Q assumption; FP4 SF/repack bytes uncounted) — for callers using the default, the residual is a conservative overestimate affecting tile choice, not correctness. (`4d2fbc9e`)

**Split-KV gate description (minor).** You're right that the stated mechanism was wrong — SF blocks run along head_dim while split-KV partitions along tokens, so split-KV never splits a scale block. The gate docstring, both call-site comments, and the test docstring are rewritten to describe what was actually observed (numerical divergence with split-KV enabled on the NVFP4 paged path), with the NUM_MMA_KV tile floor named as an unconfirmed hypothesis. Zero behavior change. (`a40d2c63`)

**Bare KeyError + defensive slicing (minor).** The lookup now raises `ValueError` naming the offending scalar and listing the derivable set; the `max(0, …)` clamp is dropped, with a comment recording the invariant it was masking (`prepare_jit_additional_args` returns at least one entry per tensor name). (`2b52729c`)

**Dead CTA_TILE_Q=32 instantiation (minor).** The jinja instantiation lists are now three-way — vo≥512 → [16,32]; qk≥512 && vo<512 → [16]; else [16,64,128] — and the `IsInvalid` comment documents the same mapping. Before removing the instantiation I checked a shipped .so with `nm`: upstream already leaves `Dispatched<32,…>` as an undefined lazily-bound symbol for vo<512 modules, so this introduces no new linkage hazard. (`f7a2b3d9`)

**GQA group size 6 (minor).** No code change, per your suggestion; I've added a note to the PR description instead: the `group_size == 6` case (needed for Gemma-4-class GQA ratios) adds one instantiation per macro user — roughly +20% compile time/binary size for the translation units that expand `DISPATCH_GQA_GROUP_SIZE`.

**New in-tree tests (`cab25c90`).** Your framing — this PR legalizes unequal K/V strides, so every reachable consumer must support them or reject them explicitly — is now proven in-tree on both halves: `test_batch_prefill_with_paged_kv_cache_nvfp4_asymmetric` runs the asymmetric read at (512,256) and (256,128) across page sizes and kv-head counts against a dequantize-then-attend reference built on #3748's `nvfp4_kv_dequantize_paged` (kernel and reference consume identical quantized bytes; existing nvfp4 tolerances, not weakened) — 8/8 pass; and `test_batch_decode_rejects_unequal_kv_strides_nvfp4_contract` proves the decode-side explicit rejection (with a positive control showing the failure is for stride inequality, not the allocation pattern). With these, `-k nvfp4` selections go 56→64 (prefill) and 289→290 (decode), all green.

**Revalidation at the new head, both consumer arches.** SM120 (RTX 5090, x86_64, nvcc 13.0.88, fresh clone + fresh JIT workspace at this exact head): **504 passed / 1 skipped / 0 failed** across the same battery — including the sm120-gated suite actually executing (8/8; it can only arch-skip on sm121) and both new tests. SM121 (DGX Spark, throwaway venv, fresh JIT workspace so the changed instantiation lists compiled from scratch): the same NVFP4 batteries as at `785d0e01` — 486 passed, 0 failed (batch_prefill 56, fp4_kv_quantization 54, page 70+1 skip, workspace_size 6, single_prefill 10, batch_decode 289, sm120-gated suite 1+7 arch skips) plus the JIT gate (2). End-to-end with the qk512/vo256 VO-split active (Gemma-4-12B, calibrated NVFP4 KV): fidelity 8/8; 8k-context retrieval margin 16/16 argmax recall with mean margin within 0.007% of the previous head's reference.

@TechPrototyper — thank you for both offers, and yes to both. (1) A same-day SM120 validation of the stride fixes at this head would be very welcome once it's pushed — the two stride-path changes (`d77e0238`, `675746dd`) are exactly the ones that benefit most from an independent sm120 run. (2) Please do run the decode partition-KV check on your 5090: the new negative test exercises the stride-rejection contract, but your partition-KV sweep covers actual split-size execution behavior that no in-tree test reaches. I'll ping this thread as soon as the head is up.

---

TechPrototyper:
**Independent SM120 validation (RTX 5090) at `cab25c90`.** NVFP4 attention battery: **466 passed, 1 skipped, 0 failed, 0 errors** (32 pre-existing SM120 xfails in `test_batch_attention.py`, unrelated). Both stride tests pass — `test_batch_prefill_with_paged_kv_cache_nvfp4_asymmetric` (8/8) and `test_batch_decode_rejects_unequal_kv_strides_nvfp4_contract`; the sm120-gated `test_nvfp4_attention_sm120.py` suite executes (not arch-skipped) 8/8, incl. accuracy s128–s8192; NVFP4 prefill/decode selection counts match your reference (64 / 290). (Scope: the coherent NVFP4 attention files — `-k nvfp4 tests/` verbatim also pulls the ~53k-test trtllm-gen collection, a separate path — so not a bit-exact rerun of your 504 count; no NVFP4 test fails.)

On the decode partition-KV check you asked for: split-KV is force-disabled for NVFP4 KV at `plan()` (`_nvfp4_kv_requires_disabled_split_kv`), so nothing in-tree reaches it enabled. Happy to exercise the split-size path however you'd like it reached — or is the gate the intended end state here?

Per your rebased-head plan (https://github.com/vllm-project/vllm/pull/46329#issuecomment-4987235429) — standing by for the ~1h sm120 pass on the HND head whenever it's up (hybrid Qwen3.6-27B repro, Gemma 3/4 incl. the head-512 VO-split, accuracy vs bf16).

qsang-nv:
Thanks — the eight follow-up commits address the previous round cleanly, and I've verified them against the diff: the decode-side stride guard (with the negative/positive test pair), the `produce_v` routing with both call sites updated, the full withdrawal of the SM121 `b12x` default, the `kv_dtype_bytes` threading kept consistent between `PrefillPlan` and `PrefillPlanWorkspaceSize`, and the honest rewrite of the split-KV gate docs. Two new items surfaced while verifying, plus a rebase note.

**[Medium] The smem-probe comment's "no current config triggers this" claim doesn't hold.** The VO side is indeed constrained — within the probe's `vo < 512` range, there is no compilable VO dimension between 256 and 512 (`static_assert(NUM_MMA_D_VO % NUM_MMA_D_VO_TILE == 0)`), and VO=512 takes the earlier branch — but the QK side has no such constraint under `pos_encoding_mode=NONE`, and `plan()` doesn't validate head dims: the JIT path compiles what it's asked for. Reachable counterexample: `(qk=448, vo=256)`, 2-byte KV, short-q: `16*448*2 + (448+256)*16*4*2 = 104448 B > 101376 B` — the probe fires today on 99KB-opt-in parts. And at those dims with FP8 KV, this revision changes the selection: the previous head (assumed 2-byte) returned CTA64, the actual 1-byte cost (59392 B) now returns CTA16. So the fallback is not purely forward-looking and the change does alter tile selection for reachable configs. Please correct the comment and commit message, and either add a `(448,256)`-style regression test, or enforce the claimed supported head-dim set in `plan()`/JIT validation and document that restriction.

**[Medium] The new asymmetric NVFP4 test doesn't execute the second blocker's fix.** `page_produce_kv_on_the_fly<produce_v>` only runs under `USE_KV_SHARED_SMEM`, which excludes FP4 and requires `HEAD_DIM_QK == HEAD_DIM_VO` — so the asymmetric NVFP4 test exercises the (already-split) `thr_local_kv_offset_{k,v}` path, not `get_paged_kv_offset_for_logical_row`. To actually pin the fix, please add: 16-bit KV, `qk = vo = 512`, K and V pools built as views of differently-padded parent tensors (identical shapes, unequal stride families), NHD at minimum (ideally HND too), checked against a reference. That configuration goes through the shared-KV-smem path, and since it's SM80+ it can run on the existing H100 runner — unlike the SM100+-gated asymmetric test, whose coverage only exists on internal runners.

Also note the branch currently has merge conflicts with main — please rebase and resolve them, and re-run the new stride tests on the rebased head.

jethac:
Thanks @qsang-nv — both catches are correct, and both are now fixed and pinned in-tree. The branch is rebased onto current main (`28b51d80`) per your note, so this update rewrites history for the first time in this PR: new head `741b6372`, 20 commits over main. Point by point:

**Smem-probe comment overclaim (medium).** You're right, and the counterexample is exact: nothing validates head dims at plan()/JIT time, so `(qk=448, vo=256)` reaches the probe's short-q branch, and at 2-byte KV `16*448*2 + (448+256)*16*4*2 = 104448 > 101376` — the probe fires today on 99KB-opt-in parts, and my "no reachable configuration triggers this" claim (comment and commit message) was wrong. Also correct that the `kv_dtype_bytes` accuracy changes tile selection at those dims (59392 B at 1 byte → CTA16 where the 2-byte assumption returned CTA64). Since the rebase rewrites history anyway, I corrected this at the source rather than stacking a fixup: the rebased probe commit (`88b6725b`) now carries the honest comment (the reachable (448,256) example, the 1-byte selection change, and the remaining approximations) and a rewritten message that names the earlier claim as wrong. On top of that I took your option 1 — `test_batch_prefill_paged_cta_tile_q_smem_probe_qk448_vo256` (`773c747a`): it computes the expected tile from the device's actual `shared_memory_per_block_optin` (so the assertion is exact on H100 and 99KB parts alike), asserts the planned `cta_tile_q` through `PrefillPlanInfo` (the technique your `test_fp8_prefill.py` uses), and for 2-byte KV runs the kernel against an exact float32 reference — on my 5090 the probe fires, CTA64 is selected, and the kernel runs and matches, which is precisely the "fallback keeps it dispatchable" behavior the comment now claims and no more. For 1-byte KV the pin is plan-level (planned tile = 16, the flip), with the reason stated in the test: the FA2 1-byte producers require head_dim to be a multiple of 128 elements (the 128b-per-lane load loop steps `NUM_MMA_D` by 8, and the k128B swizzle needs an 8-aligned upcast stride), so no currently-runnable 1-byte configuration reaches the flipped region — the largest runnable in-branch fp8 config, (384, 256), costs 94208 B and never fires even at the 2-byte estimate. I did not take option 2 (enforcing a supported head-dim set in plan()/JIT validation): rejecting dimensions that plan() accepted yesterday is a user-visible behavior change I'd rather not couple to this PR — happy to do it as a follow-up if you'd prefer belt-and-braces.

**Asymmetric test doesn't execute the produce_v fix (medium).** Correct again — `USE_KV_SHARED_SMEM` excludes FP4 and requires `HEAD_DIM_QK == HEAD_DIM_VO`, so the NVFP4 asymmetric test exercises the prefetched `thr_local_kv_offset_{k,v}` path and never reaches `get_paged_kv_offset_for_logical_row<produce_v>`. Added exactly the test you specified (`741b6372`): 16-bit KV, `qk = vo = 512`, K and V pools as views of differently padded parent tensors (identical logical shapes, unequal stride families — the decode negative test's construction), parametrized over NHD **and** HND and over short-q/long-q, checked against an exact float32 reference. That the config takes the shared-KV-smem path is established two ways: by static reasoning, stated in the docstring — at (512, 512) fp16 the planner picks CTA_TILE_Q=16 (NUM_WARPS_KV=4; NUM_MMA_D_VO=32 % 4 == 0) or CTA_TILE_Q=32 (`kLargeHeadWarpSplit`, NUM_WARPS_KV=2), and `USE_KV_SHARED_SMEM` holds in both, so the qo_len parametrization covers both tiles — and by execution: with `f0c27652` (the produce_v routing fix) reverted in a scratch build at this head (fresh JIT workspace), the new test fails on all four parametrizations — 99.2% of output elements mismatch, greatest absolute difference 1.33 against the 2e-3 tolerance — and passes 4/4 with the fix restored. So the test demonstrably catches the exact bug it pins. It is SM80+ and runs on the standard runners.

**Rebase.** Done, onto `28b51d80`. Two commits needed manual resolution, both against #3871's `uniform_q_len` threading through the same plan-path signatures our `head_dim_qk`/`kv_dtype_bytes` threading extends (`PrefillSplitQOKVIndptr` / `PrefillPlanImpl` / `PrefillPlan` / `PrefillPlanWorkspaceSize` plus the two FFI entry points in `csrc/batch_prefill.cu`): the resolution keeps upstream's parameter and its new uniform-q-len branch, threads `head_dim_qk`/`kv_dtype_bytes` into that branch's `FA2DetermineCtaTileQ` calls as well, and drops the now-stale `(void)head_dim_qk;` that upstream added while the parameter was unused. Your #3863 fp8-h512 clamp auto-merged cleanly around our single-prefill dispatch change (the call now reads `FA2DetermineCtaTileQ(packed_qo_len, HEAD_DIM_VO, HEAD_DIM_QK, sizeof(DTypeKV))` under your new `kLargeHeadWarpSplit` comment). Everything else is byte-identical under `git range-diff`: 15/18 prior commits unchanged, 2 conflict-resolved as above, 1 (the probe commit) additionally carrying the folded comment/message correction from item 1.

**Revalidation at the rebased head** (SM120, RTX 5090, x86_64, driver 580.126.09, nvcc 13.0.88, torch 2.11.0+cu130, fresh clone + fresh JIT workspace): batch_prefill `-k nvfp4` 64, fp4_kv_quantization 54, page `-k nvfp4` 70+1 skip, workspace_size 6, single_prefill `-k nvfp4` 10, batch_decode `-k nvfp4` 290, the sm120-gated suite 14/14 executing (8 as before plus the 6 tests main itself has added to that file since our merge-base, which the rebase picks up), JIT gate 2 — **510 passed / 1 skipped / 0 failed**, plus the two new tests above (2 + 4) for **516 / 1 / 0** total, and the reverted-fix red run as the negative control.

@TechPrototyper — on your partition-KV question: yes, the plan()-level split-KV disable is the intended end state *for this PR* — it documents an empirically observed divergence, and removing it is a perf investigation I'd rather not couple to these correctness fixes. The rebased head is up; the same-day SM120 pass you offered would be very welcome again, particularly on the two new stride tests and the rebase seam (the #3871/#3863 interactions).

TechPrototyper:
**Independent SM120 validation (RTX 5090) at the rebased head `741b6372`** — same-day, as offered. Fresh clone at the exact SHA via the PR ref, merge-base verified as `28b51d80`, editable install, fresh JIT workspace, `TORCH_CUDA_ARCH_LIST=12.0`, nvcc 13.0.88, torch 2.11.0+cu130. One deliberate environment difference worth having on record: driver **595.58.03** (vs. your 580.126.09) and a containerized runner — so this is also an independent driver-branch datapoint.

| Suite | Result |
|---|---|
| **New:** `test_batch_prefill_paged_shared_kv_smem_unequal_kv_strides` (NHD+HND × qo 17/65) | **4/4 passed** |
| **New:** `test_batch_prefill_paged_cta_tile_q_smem_probe_qk448_vo256` (fp16 + fp8) | **2/2 passed** |
| batch_prefill `-k nvfp4` | **64 passed** |
| fp4_kv_quantization | **54 passed** |
| single_prefill `-k nvfp4` | **10 passed** |
| batch_decode `-k nvfp4` | **278 passed / 0 failed** (12 env-OOM skips, see below) |
| sm120-gated suite | **14/14** (incl. the 6 main-added tests) |
| JIT gate | **2/2** |
| workspace_size | **6/6** |
| Rebase seam: `test_fp8_prefill.py` complete | **1605 passed / 0 failed** |
| page `-k nvfp4` | 40 passed / 29 env-OOM skips / 1 env-flaky fail (not PR-attributable, see below) |

Every PR-touched path is green, all counts match your reference where our environment permits, and the seam commits (#3871 `uniform_q_len` threading, #3863 clamp) hold under the full fp8 prefill battery.

**One for-the-record observation, explicitly not attributed to this PR.** In our containerized runs, `tests/conftest.py`'s OOM guard converted 29 page + 12 decode tests into skips (genuine `torch.cuda.OutOfMemoryError`s that your bare-metal runs don't hit), and main's `test_..._cuda_graph_bad_scale` — untouched by this PR — flaked in suite context: 2 fail / 2 pass at this head, 3/3 pass on bare main `28b51d80`, 6/6 pass isolated on both. The failure mode is the expected device-side error on the bad-scale graph replay not surfacing, and it only occurred when the page suite ran after heavier suites in the same container — i.e., VRAM-pressure-mediated, plausibly driver-branch-sensitive (595 vs 580). Since the test's subprocess-trap design is main's, not yours, we'll take this to a separate flashinfer issue with the full 2×2 matrix if maintainers want it — flagging it here only so nobody bisects it into this PR later.

On split-KV: understood and agreed — documenting the observed divergence and decoupling the perf investigation is the right scope; the split-size-sweep offer stands for that follow-up.

*(Run and analysis prepared with Claude (Anthropic) per the project's AI policy; job specs, build and test logs retained, all judgment calls human-reviewed.)*

qsang-nv:
One more item, found by running the two new tests on an H100 (SM90) at `741b6372` — which neither of the SM120 validation environments could surface:

**[Test blocker] The fp8 parametrization of `test_batch_prefill_paged_cta_tile_q_smem_probe_qk448_vo256` errors on pre-SM100 machines, including the H100 runner.** `_fa2_head_dim_nvcc_flags` restricts non-NVFP4 1-byte large-head modules to `supported_major_versions=[10, 11, 12]`, so on an SM90-only system the JIT spec generation inside `plan()` raises before the tile-selection assertion is reached:

```
RuntimeError: No supported CUDA architectures found for major versions [10, 11, 12].
```

Reproduced verbatim on our H100 (raised via `_fa2_prefill_head_dim_nvcc_flags` → `get_nvcc_flags_list`). `skip_if_head_dim_unsupported()` only gates the 16-bit large-head path (SM80+), so it doesn't catch this. Please add a dtype-aware skip mirroring the module gate, e.g.

```python
if kv_dtype.itemsize == 1 and get_compute_capability(torch.device("cuda:0"))[0] < 10:
    pytest.skip("FP8 KV with head_dim > 256 requires SM100 or newer")
```

and narrow the docstring's "exact on every architecture" to "every supported architecture". The fp16 parametrization is unaffected and keeps the 2-byte fallback's kernel coverage on SM80/90.

For the record, the rest of the H100 run at this head is green: probe test fp16 param passes (large-smem branch: planned `cta_tile_q == 16`, kernel matches the reference), and `test_batch_prefill_paged_shared_kv_smem_unequal_kv_strides` passes 4/4 (NHD/HND × short-q/long-q) — an SM90 datapoint for the `produce_v` fix to go with the two SM120 environments.

jethac:
@qsang-nv Good catch, thank you — you're right that the FP8 parametrization of
`test_batch_prefill_paged_cta_tile_q_smem_probe_qk448_vo256` errors before the tile
assertion on pre-SM100. As you traced: `_fa2_head_dim_nvcc_flags` restricts the
non-NVFP4 1-byte large-head modules to major versions [10, 11, 12], so on SM90 the
JIT spec-gen inside `plan()` raises `No supported CUDA architectures found for major
versions [10, 11, 12]` before the probe assertion is ever reached.
`skip_if_head_dim_unsupported()` only gates the 16-bit path, so it doesn't cover this.

Fixed by adding a dtype-aware skip that mirrors the module gate, placed where
`kv_dtype` is known and before `plan()`:

```python
if kv_dtype.itemsize == 1 and get_compute_capability(torch.device("cuda:0"))[0] < 10:
    pytest.skip("FP8 KV with head_dim > 256 requires SM100 or newer")
```

I also narrowed the docstring wording "exact on every architecture" ->
"exact on every supported architecture". The fp16 parametrization is unaffected
(2-byte fallback, SM80+).

I reproduced your environment on an H100 (SM90, CUDA 13.0 / torch cu130) to confirm:

- `test_batch_prefill_paged_cta_tile_q_smem_probe_qk448_vo256`: the fp8 param now
  **skips cleanly** ("FP8 KV with head_dim > 256 requires SM100 or newer") instead of
  raising, and the fp16 param **passes** (large-smem CTA16 branch, matches reference).
  `1 passed, 1 skipped`.
- `test_batch_prefill_paged_shared_kv_smem_unequal_kv_strides`: **4/4 passed**
  (NHD/HND x short-q/long-q) — the produce_v routing on our own SM90.
- The rest of the nvfp4 prefill/decode battery that runs on SM90 is green with the
  expected arch-skips and no errors.

This branch is also rebased onto current `main` (the PR had gone stale). Thanks again
for the review, and for running the H100 pass on your side.

jethac:
Bonus datapoint: beyond the H100 (SM90) reproduction above, I rented an **SM100 datacenter Blackwell B200 on RunPod** to exercise the one architecture in this module's `supported_major_versions=[10, 11, 12]` set that had otherwise never actually run here — the asymmetric qk512/vo256 VO-split path works there too.

Built from this branch's HEAD, editable install, fresh JIT, compiled natively for `compute_100a`:

- `test_batch_prefill_paged_cta_tile_q_smem_probe_qk448_vo256`: the fp8 (448, 256) parametrization **runs (does not skip) and passes** on SM100 — the dtype-aware skip I added above correctly fires only on pre-SM100, so datacenter Blackwell keeps exercising the real tile-selection assertion.
- `test_batch_prefill_paged_shared_kv_smem_unequal_kv_strides`: 4/4 (the produce_v routing on SM100).
- The full NVFP4 prefill + decode battery: **503 passed / 14 skipped / 1 failed**, no kernel errors.

The single failure was arch-independent, not an SM100 finding: the host-side `test_customize_batch_prefill_nvfp4_large_head_uses_prefill_flags` still called `gen_customize_batch_prefill_module` with empty additional-tensor lists after the NVFP4 scale-factor input contract landed, so it raised `ValueError` before reaching its flag assertions — and it fails identically on every architecture (it never touches the GPU). Fixed in the latest commit on this branch by passing the two `maybe_k_cache_sf` / `maybe_v_cache_sf` tensors, mirroring the production caller.

Net: the qk512/vo256 kernel is now demonstrated green on SM90 (skip path), SM100 (B200), and SM120/121 — no gate narrowing needed.

qsang-nv:
LGTM, please resolve conflicts and I will trigger CI.

jethac:
@qsang-nv Rebased onto current `main` (`e683e3078`) — head is now `2ed09bd3a`. The rebase applied cleanly (the "conflicting" state was GitHub's merge-base check rather than real content conflicts; `main` hadn't touched the files this PR carries). Nothing else changed vs the head you LGTM'd — the 22 commits, the SM100 skip-fix, and the tests are all preserved. Ready for you to trigger CI whenever convenient. Thanks!

qsang-nv:
Please run pre-commit locally.https://github.com/flashinfer-ai/flashinfer/actions/runs/30316271761/job/90155731131?pr=3684

TechPrototyper:
**Independent sm_121 validation at this head (`2ed09bd3a`) — first non-author GB10 datapoint: all PR-relevant suites green.**

New capability on our side since the last datapoint: an ASUS Ascent GX10 (NVIDIA GB10, **sm_121**, 121 GB unified) now sits next to the RTX 5090 — so this PR's `supported_major_versions`/sm_12x claims can be exercised on real consumer *and* SoC Blackwell independently of the author's hardware.

**Setup:** this branch at `2ed09bd3a`, editable install, fresh JIT compiled natively on-device (`.cache/flashinfer/0.6.15/121a/`), PyTorch 2.11.0+cu130, CUDA 13.0, driver 580.142, inside the `vllm/vllm-openai:latest` aarch64 container. `FLASHINFER_DISABLE_VERSION_CHECK=1` (no 0.6.15 `flashinfer-cubin` on PyPI yet).

| Suite | Result |
|---|---|
| `tests/jit/test_attention_utils.py` | **2 passed** (incl. the fixed NVFP4 SF-tensor flag probe) |
| `tests/jit/test_jit_cpp_ext.py` | **20 passed** |
| `tests/attention/test_nvfp4_attention_sm120.py` | **14 passed** — the binding's `minor == 0 \|\| minor == 1` gate verified on real 12.1 silicon |
| `tests/attention/test_batch_prefill_kernels.py` | **8737 passed / 1320 skipped / 5760 xfailed / 0 real failures** (54 initial errors were my container missing cuBLAS dev headers for the *test-side* `fp4_quantization` JIT helper — `cublasLt.h` not found; after `apt-get install libcublas-dev-13-0`, the full nvfp4 subset re-ran **64/64 green**) |
| ↳ the two new tests, explicitly | `cta_tile_q_smem_probe_qk448_vo256` 2/2, `shared_kv_smem_unequal_kv_strides` 4/4 — **6/6**, matching my sm_120 run at `741b6372` |
| `tests/attention/test_batch_decode_kernels.py` | **3260 passed / 0 failed** |

One environment trap worth recording for anyone else validating on GB10 with the vLLM image: the image ships `flashinfer-jit-cache 0.6.14+cu130` AOT modules, and those load *ahead of* a source checkout — their pre-PR `nvfp4_attention_sm120` binding still hard-requires CC 12.0 and rejects sm_121 (13 failures that look like PR failures but aren't). `pip uninstall flashinfer-jit-cache` + clean JIT cache, and the PR's own binding compiles and passes 14/14.

Net: at the head qsang-nv LGTM'd, sm_121 now has independent data — prefill, decode, the VO-split stride tests and the sm12x NVFP4 kernel suite are all green on GB10.

*Disclosure per repo AI policy: validation run AI-assisted under my direction; hardware and results are ours.*

TechPrototyper:
@jethac in case it saves you a cycle on the pre-commit ask above: the failures are exactly three mechanical items — clang-format on 5 C++/CUDA files, `ruff format` on 2 test files, and one `SIM105` in `flashinfer/cute_dsl/utils.py` (`try/except/pass` → `contextlib.suppress(AttributeError)` + the `import contextlib`). We ran the full hook set locally on `2ed09bd3a` with those applied: 13/13 hooks pass, no semantic changes (formatting only, plus the suppress rewrite). Happy to send the diff (8 files, +21/−24) if useful — it applies cleanly to the current head.

*Disclosure per repo AI policy: AI-assisted verification.*

jethac:
Thanks so much for the review and the LGTM, @qsang-nv — much appreciated.

The only outstanding item was the failing pre-commit check. I ran the pinned hooks locally (`pre-commit run --all-files`) on a clean Linux checkout and pushed the result in 47aee407: clang-format on the touched C++/CUDA files, ruff-format on two tests, and one `SIM105` in `flashinfer/cute_dsl/utils.py` (`try/except/pass` → `contextlib.suppress(AttributeError)`). It's formatting/style only — no kernel or semantic changes — and all hooks now pass cleanly.

Special thanks to @TechPrototyper for pinning down the exact set of failures and for the independent sm_121 validation of this path — both were a big help.

Should be green for CI now; ready when you are.

jethac:
Amended the previous commit to correct @TechPrototyper's `Co-authored-by` email (`tech@smartlogics.net`) so the attribution links properly at merge — **the tree is byte-identical** to the prior head (message-only change), so this doesn't affect the review. Ready for CI.

TechPrototyper:
**CI-failure triage: both red signals look infrastructural, not PR-attributable — data below.**

**1. The GitLab pipeline result (`0/18 executed test jobs passed`) is the run started 2026-07-28 on the *previous* head** (pipeline #59878090, triggered before the pre-commit fix / amend). Its failure signature is uniform across every architecture and CUDA version — including **H100 and GB200/GB300, which this PR's sm12x-gated paths don't touch** — and in every case it is `script failed before producing a JUnit report` or explicit `CI infrastructure failure` (both 5090 jobs). Zero JUnit artifacts were produced anywhere. A change set that actually broke tests would produce *failing tests*, not a pipeline that dies before reporting on all 18 jobs uniformly.

**2. The `AOT Build Import (arm64, cu128)` failure is a spot-instance shutdown**, from the job log at 09:39:43Z, ~53 min in:
```
##[error]The runner has received a shutdown signal. This can happen when the
runner service is stopped, or a manually started runner is canceled.
##[error]The operation was canceled.
```
(with `task_monitor_spot.sh` active and the analyze step emitting `is_spot_termination` / `rerun_matrix`). The build itself was progressing normally until the instance was reclaimed.

**3. Positive controls on the actual code:** the GitHub-Actions JIT unittests pass on the current head (`JIT Unittest (T4)` 26m, `JIT Unittest 1 (A10G)` 1h19m, both green), and the full battery on real consumer Blackwell was green independently — sm_121/GB10 at the tree-equivalent predecessor head ([battery](https://github.com/flashinfer-ai/flashinfer/pull/3684#issuecomment-5124692235): nvfp4 suite 14/14, prefill 8737 passed, decode 3260/3260, new tests 6/6).

Net: nothing in either red signal points at the change set; a fresh pipeline run on `e79f21b6f` should be all that's needed.

*Disclosure per repo AI policy: AI-assisted triage; log forensics verified by us.*

jethac:
> **⚠️ Superseded (2026-07-31):** `main` moved again, so the branch has since been rebased onto current `main` and is now conflict-free at **`fa894014e`** (`MERGEABLE`, 0 behind) — that's the head ready for CI. The note below referred to the earlier head `e79f21b6` and a stale-pipeline signal that has since self-resolved (the `AOT Build Import` reruns went green); please disregard it for triage.

<details><summary>Original note (2026-07-30, head <code>e79f21b6</code>) — kept for record</summary>

@qsang-nv — following your LGTM, the branch is at `e79f21b6` (rebased onto current `main` + the pre-commit fixes applied); nothing else changed vs the head you approved. Whenever you have a cycle to trigger CI — the two red signals on this head are both infrastructural, not the change set (@TechPrototyper ran the log forensics):

- The GitLab pipeline showing `0/18` is the **07-28 run on the previous, pre-amend head** (pipeline #59878090) — failures are uniform across *every* arch including **H100 and GB200/GB300**, which this PR's sm12x-gated paths don't touch, and every job died with `script failed before producing a JUnit report` / `CI infrastructure failure` (zero JUnit artifacts anywhere). Broken code produces failing tests, not a pipeline that dies before reporting on all 18 jobs uniformly.
- The `AOT Build Import` jobs were **spot-instance cancellations** (`runner has received a shutdown signal`, ~53 min in), not build errors.

Positive controls on the code are green: `pre-commit` passes on this head, the GitHub-Actions JIT unittests pass (T4 + A10G), and there's now independent consumer *and* SoC Blackwell data — @TechPrototyper's RTX 5090 (sm120) and ASUS GX10 / GB10 (sm121) batteries are green at this tree (nvfp4 suite 14/14, prefill 8737 passed, decode 3260/3260, the two new VO-split stride tests 6/6).

A fresh pipeline run on `e79f21b6` should be all that's needed. Thanks again for the review.

</details>

jethac:
@qsang-nv rebased onto current `main` — now conflict-free at `fa894014e` (MERGEABLE, 0 behind), nothing else changed vs the head you LGTM'd; ready for you to trigger CI whenever.

jethac:
@qsang-nv the CI run ([30594351621](https://github.com/flashinfer-ai/flashinfer/actions/runs/30594351621)) has been stuck ~1.5 days on `JIT Unittest (H100)` — still `in_progress` with no movement since 07-31 13:46 (looks like a hung/queued runner), so it never completes. The only real failure is `JIT Unittest 4 (A10G)` → `tests/test_artifacts.py::test_get_subdir_file_list`, an unrelated cubin-download mock test (this PR touches attention kernels, not `artifacts`/`cubin`/`deep-gemm`; it fails on a stale-mock 404 for a `deep-gemm` checksum, almost certainly red on `main` too). Could you cancel + re-trigger the run? `T4` and A10G shards 1–3/5 are green. Thanks!

jethac:
Update — run [30594351621](https://github.com/flashinfer-ai/flashinfer/actions/runs/30594351621) completed: H100 ran the full suite green (attention/nvfp4/gemm/moe); the only red is `tests/test_artifacts.py`, which is byte-identical to `main` here and fails on a missing `DEEPGEMM_RUBIN` cubin mock (added by the SM107 v0.6.16 merge-back, #4280) — pre-existing on main, unrelated to this PR.

qsang-nv (approved):

TechPrototyper:
Since the internal CI's RTX 5090 rows were reported as infrastructure failures, here is an independent datapoint from consumer hardware, at the current PR head.

**Environment.** RTX 5090 (sm_120), driver 595.71.05, CUDA 13.0, torch 2.11.0+cu130, containerized Linux. flashinfer checked out at `dd25a783` ("Merge branch 'main' into nvfp4-vosplit-rederive"), editable install, JIT kernel path (`TORCH_CUDA_ARCH_LIST=12.0`, `MAX_JOBS=8`), pytest with 1800 s per-test timeout.

**Results — zero failures, zero errors across all three suites:**

| Suite | passed | failed | errors | skipped | xfailed | xpassed | wall time |
| `tests/attention` | **117,829** | 0 | 0 | 265,520 | 6,717 | 11,242 | 1 h 24 min |
| `tests/gemm -k fp4` | **8,453** | 0 | 0 | 20,170 | — | — | 18 min |
| `tests/moe -k fp4` | **125** | 0 | 0 | 9,117 | — | — | 9 min |

Skips are the expected non-sm120 paths (SM100/hopper-only kernels, unsupported dtypes/head-dims). No test hit the 1800 s timeout.

Two environment notes for reproducibility:
- `scipy` had to be added to the image (`tests/attention/test_block_sparse.py` and `test_hopper_fp8_attention.py` import it at collection time; without it the whole attention suite aborts on collection unless `--continue-on-collection-errors` is set).
- The 11,242 xpassed in the attention suite are tests marked xfail that pass on this platform; not investigated further, mentioned for completeness.

junit XMLs for all three suites available on request.

qsang-nv:
@flashinfer-bot run

jethac:
@TechPrototyper thank you for this. I'd been meaning to rent a 5090 on RunPod and run exactly this myself and just haven't had the time — you saved me a real chunk of work. The `scipy` collection-time import is a good catch too; pre-existing gap in `tests/attention` rather than anything this PR introduces.

Your run was at `dd25a783`, so I've re-run at the current head `00054844` on a rented 5090 — the upstream merge in between touches one file this PR also touches (`flashinfer/mla/_core.py`).

**Environment.** RTX 5090 (sm_120), driver 580.126.09, CUDA 13.0 (nvcc 13.0.88), torch 2.9.1+cu130, editable/JIT install at `00054844`, `TORCH_CUDA_ARCH_LIST=12.0`. Different torch minor from the run above, so between us this has sm120 coverage on two torch versions.

**19,202 tests executed, zero failures, zero errors:**

| suite | executed | failed | skipped | time |
|---|---|---|---|---|
| `tests/attention -k fp4` | 508 | 0 | 31,229 | 31 min |
| `tests/gemm -k fp4` | 6,551 | 0 | 22,072 | 12 min |
| `tests/jit` | 129 | 0 | 0 | <1 s |
| `test_batch_{decode,prefill}_kernels.py`, unfiltered | 12,014 | 0 | 7,117 | 21 min |

The last row is the two attention test files this PR modifies, run without the fp4 filter, since the PR also edits shared headers (`prefill.cuh`, `scheduler.cuh`, `page.cuh`, `vec_dtypes.cuh`) that non-fp4 paths compile against. `tests/jit` is included because the PR touches `test_attention_utils.py` and `test_jit_cpp_ext.py`.

`test_nvfp4_attention_sm120.py` ran 14/14 with nothing skipped: accuracy across s128–s8192 at d64/d128 causal and non-causal, structured-Q correction, LSE, causal mask column order, split-KV gate dtype logic.

Attention skips are arch gating rather than the feature being passed over — 15,584 `trtllm-gen requires SM100/SM103`, 9,792 xqa backend limits, 2,560 SM100/SM103-only.

@qsang-nv — the `@flashinfer-bot run` from Aug 11 hasn't reported back. Could you check whether that pipeline stalled, or re-trigger it? From the Aug 3 run (#60789494), none of the red was attributable here: 5090 lost both CUDA versions to infra failure, GB200/13.0 was an infra interrupt, and the B300/12.9 `test_trtllm_gen_attention_decode` mismatch plus `test_trtllm_gen_fused_moe` timeout both pass on 13.0, with attribution flagged uncertain by the bot. GB300, H100 and RTX Pro 6000 were green on both.

junit XMLs available if useful.

PR #3687, opened 2026-06-22
Fix AutoTuner memory leak for MoE kernels

jdebache:
# Fix AutoTuner memory leak for MoE kernels

## Root cause

`AutoTuner._find_nearest_profile` was decorated with `@lru_cache(maxsize=None)`.
Its cache key includes the `TuningConfig` argument, which is hashed via
`DynamicTensorSpec.__hash__` — and that hash uses `id()` of callable fields
(specifically `map_to_tuning_buckets`).

Any call site that constructs a fresh `TuningConfig` on every inference call
produces a new callable object with a new `id()`, which means a new cache entry
on every call — regardless of whether the input shape changed.  At 1000 calls/s
this grows at ~1.2 GB/day.

### `flashinfer/autotuner/` (new submodule)

Converted `autotuner.py` into a proper package:

- `autotuner.py` — original implementation, unchanged except two inline
  `lambda shapes, dtype, device: (torch.rand(...) * 10 - 5).to(dtype)` defaults
  replaced with the named `autotuner_initializer_rand_scaled`
- `initializers.py` — **new**: six named, reusable tensor initializers following
  the `(shapes, dtype, device) -> Tensor` contract:
  `empty`, `zeros`, `ones`, `randn`, `rand`, `rand_scaled`. Although we don't hash the initializers as part of the key currently, having the most well-known options with stable object identity is setting up for the future improvement.
- `__init__.py` — re-exports all public symbols so existing
  `from flashinfer.autotuner import ...` imports continue to work unchanged

While working on understanding the auto-tuner logic, I have also modernized and extended the typing annotations.

### `flashinfer/fused_moe/utils.py`

Added a cached factory next to `map_to_hybrid_bucket`:

```python
@functools.cache
def make_hybrid_bucket_mapper(max_num_tokens: int) -> Callable[[int], int]:
    return functools.partial(map_to_hybrid_bucket, max_num_tokens=max_num_tokens)
```

The `@functools.cache` ensures the same `partial` object is returned for the same
`max_num_tokens`, keeping its `id()` — and therefore the `DynamicTensorSpec`
hash — stable across calls.

### `flashinfer/fused_moe/core.py`

- All three `lambda x: map_to_hybrid_bucket(x, tune_max_num_tokens)` occurrences
  replaced with `make_hybrid_bucket_mapper(tune_max_num_tokens)`
- Six inline initializer lambdas in `_make_tuning_config` replaced with named
  functions from `initializers.py`

### `flashinfer/trtllm_low_latency_gemm.py`

`TuningConfig` hoisted from per-call construction to a module-level singleton
`_LLGEMM_TUNING_CONFIG`.

### Minor fixes
- Removed dependency from auto-tuner on MoE module by extracting a couple of generic utilities.

## Test results

Measured with 5 000 calls, single shape, directly against `_find_nearest_profile`:

| | Leaky | Fixed |
|---|---|---|
| Cache growth | +5 000 (1 per call) | +511 (1 per distinct shape) |
| Python allocation | 6 166 KB | 638 KB gross, bounded |
| Per-call cost | 1 263 B | 131 B |
| **Extrapolated @ 1M calls** | **~1.2 GB** | ~128 MB (then plateaus) |

## Release Notes

* **New Features**
  * Added reusable AutoTuner tensor initializer utilities and exported bucket-mapping helpers.
  * Added a cached hybrid token bucketing mapper for MoE autotuning.

* **Improvements / Bug Fixes**
  * Improved autotuner profiling cache identity and persistence for more consistent results.
  * Fixed an AutoTuner memory leak by stabilizing tuning-cache inputs.
  * Shared common bucketing and tuning configuration logic across MoE/MLA-related autotuning paths.

* **Tests**
  * Added a regression test to detect cache-growth/memory regressions.

* **Documentation**
  * Updated autotuner leak-fix notes.


flashinfer/autotuner/autotuner.py
```diff
@@ -1602,8 +1736,8 @@ def _generate_optimization_profiles(
     @classmethod
     @lru_cache(maxsize=None)
     def _find_nearest_profile(
-        cls, shapes: Tuple[torch.Size], tuning_config: TuningConfig
-    ) -> Tuple:
+        cls, shapes: tuple[tuple[int, ...], ...], tuning_config: TuningConfig
+    ) -> tuple[tuple[int, ...], ...]:
```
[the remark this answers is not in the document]
jdebache: I think `_find_nearest_profile` is called prior to checking the profiling cache, which means it is on the hot-path no matter what. I assume the reason caching was added to this function was to mitigate a specific performance issue. Waiting on maintainers to weigh in on this.

CHANGES_AUTOTUNER_LEAK.md
```diff
@@ -0,0 +1,78 @@
+# AutoTuner memory leak — branch `fix/autotuner-memory-leak`
```
qiching: `CHANGES_AUTOTUNER_LEAK.md` please move to PR description.

flashinfer/autotuner/autotuner.py
```diff
@@ -1600,10 +1736,10 @@ def _generate_optimization_profiles(
         return generated_profiles
 
     @classmethod
-    @lru_cache(maxsize=None)
+    @functools.cache
```
qiching: i think it still unbounded, should add maxsize. please use lru_cache(maxsize=N) so a future regression degrades to cache-thrash instead of host OOM.

tests/autotuner/test_autotuner_core.py
```diff
@@ -1015,3 +1015,65 @@ def test_skip_ops_restored_after_context():
 
     # After context, skip_ops should be empty — op goes through normal path
     assert tuner._effective_skip_ops == frozenset()
+
+
+def test_find_nearest_profile_lru_cache_memory_leak():
```
qiching: by my understanding, thhe regression test does not test the leak. this function just calls with a stable named function and asserts the cache grows by N, that's asserting normal behavior? shoudl we construct an equivalent TuningConfig repeatedlyand assert the cache does not grow. As written it would still pass even if the leak came back.

qiching (changes requested):
Design-level concern:

This fix works by giving the map_to_tuning_buckets callables a stable identity so the same args return the same object and the cache key stays stable. That fixes the known MoE call sites, but it's a fix-by-convention rather than fix-by-design that the underlying root cause is that `DynamicTensorSpec.__hash__` hashes callables by `id()`. Any future call site that passes a fresh lambda or closure as map_to_tuning_buckets will silently reintroduce the exact same unbounded cache leak, with nothing to catch it.


the current approach is fine as an immediate fix, but I don't want to merge it as the final state without either the hash fix or the bounded cache.

jdebache:
Hey @qiching,

I agree with you. Fundamentally, the autotuner's interface does not allow safe caching, as it accepts objects that cannot be compared by value.

I would like to follow up on these changes with a more definitive (and invasive) set of changes. This is also the reason why I've invested a bit into adding typing annotations.

I've added a max-size to the caches as you suggest to guard against leaks for now.

aleozlx:
cc @qiching

aleozlx (approved):

jdebache:
I see there are conflicts. Will resolve a bit later.

qiching (approved):
LGTM

PR #3694, opened 2026-06-23
feat(attention): FP8 KV cache support for Hopper SM90 MLA

### LINKED ISSUE #2144: Q bf16 kv fp8 for MLA
Hi!

In sglang we cast the entire kv cache for each layer to bf16 from fp8 which results in performance degradation on H100.

https://github.com/sgl-project/sglang/blob/dbab5d50a3d6d1fc0169c04a33f1b6dcefdeec04/python/sglang/srt/layers/attention/flashinfer_mla_backend.py#L639

since flashinfer unsupoorts mixed types for q and kv for MLA

Can we support it?

qsang-nv:
## Summary

Add native FP8 e4m3 KV cache support to the Hopper SM90 MLA kernel (closes #2144). Eliminates SGLang's per-layer `.to(q.dtype)` cast workaround on FP8 KV deployments, saving one HBM round-trip per layer on DeepSeek-V3/V4.

## Approach

Hopper WGMMA has no BF16×FP8 mixed instruction, so the kernel keeps KV in FP8 through HBM/L2/shmem and dequants to BF16 in a shared-memory staging buffer just before WGMMA. WGMMA stays BF16×BF16 — bit-identical to the existing BF16 fast path. The dequant uses `vec_cast<bf16, fp8_e4m3>` (`fast_dequant_f8f16x4`) and applies the per-tensor scale in-kernel via packed `__hmul2`.

Shared-memory budget is tight (~225KB / 228KB) — FP8 KV halves the per-stage KV footprint, but the cross-stage BF16 staging buffers add it back; tracked by a defensive `static_assert`.

## Scope (locked)

- `q_data_type = torch.bfloat16`
- `kv_data_type = torch.float8_e4m3fn`
- `backend = "fa3"` (SM90 only; FA2 / CUTLASS rejected at plan time)
- `head_dim_ckv = 512`, `head_dim_kpe = 64` (DeepSeek-V3 shape; enforced by C++ `static_assert` + Python plan-time check)

Full-FP8 (Q=fp8) and per-head scales are explicit follow-ups.

## API change

`BatchMLAPagedAttentionWrapper.run()` gains two keyword-only optional floats (semantics: `real = quantized * scale`):

```python
def run(self, q_nope, q_pe, ckv_cache, kpe_cache, *,
        ckv_scale: Optional[float] = None,
        kpe_scale: Optional[float] = None,
        ...)
```

Required when `kv_data_type == float8_e4m3fn`; rejected otherwise. Matches the "unscaled cache + read-time compensation" convention already used by `trtllm_mla` / `tokenspeed_mla` / `dsa_backend` (cache stores `quantize(K)`, kernel applies scale at dequant time).

`plan()` adds four validation layers for FP8 KV: kv_data_type allow-list, `backend == "fa3"`, `q_data_type == bfloat16`, head-dim lock. `run()` validates all four tensor dtypes against planned dtypes to catch silent reinterpret bugs.

## Tests

New `tests/attention/test_deepseek_mla_fp8_kv.py` — **170 cases**:

- 144 cases: main matrix (batch × kv_len × qo_len × page_size × num_heads × causal) vs BF16 reference, `rtol=atol=1e-2`
- 9 cases: scale sensitivity sweep (`ckv_scale`/`kpe_scale` ∈ {0.01, 0.1, 1.0})
- 1 case: zero-KV → zero-output
- 1 case: KPE-dominant deterministic regression (catches FP8 KPE shmem-row aliasing — fixed via narrower `SWIZZLE_MODE_KPE_RAW`)
- 15 cases: plan-time and run-time rejection paths (unsupported kv_dtype, unsupported head_dim, FP16 Q rejection, keyword-only enforcement, dtype mismatch on `run()`, required-scale enforcement)

BF16/FP16 path: zero behavioral change — all gating is `if constexpr (KTraits::USE_KV_REPACK)` where `USE_KV_REPACK = std::is_same_v<DTypeKV, __nv_fp8_e4m3>`. Existing `tests/attention/test_deepseek_mla.py` BF16 sample case verified on H100.

Tested on H100 (SM90), CUDA 13.2, torch 2.11.

## Out of scope / follow-ups

- **AOT**: FP8 KV MLA is JIT-only in this PR (`flashinfer/aot.py:365-378` comment retained). Prebuilt wheels still don't ship FP8 KV MLA — can be added in a follow-up.
- **Full-FP8** (Q=fp8) — separate kernel variant; not requested by current op.

## Companion SGLang change

SGLang side wired up in a parallel PR (sglang fork branch: skips the per-layer `.to(q.dtype)` cast on FP8 KV path and passes `ckv_scale = kpe_scale = layer.k_scale_float` to `wrapper.run()`. Merge order: this PR + release first, then the SGLang one (pip wheel dependency).

## 🔍 Related Issues
https://github.com/flashinfer-ai/flashinfer/issues/2144

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * Added support for separate scaling values when running FP8 KV-cache attention.
  * Expanded MLA planning and execution to better validate supported data types and backend-specific settings.
  * Updated trace templates and test coverage to include the new FP8 KV-cache flow.

* **Bug Fixes**
  * Improved runtime checks to catch mismatched tensor types and missing scale values earlier.
  * Added Hopper-specific handling to avoid shared-memory corruption in narrow KV layouts.


flashinfer/mla/_core.py
```diff
@@ -1578,6 +1578,51 @@ def plan(
...
+            # Fail before JIT-compiling a module the device can't actually run.
+            major, minor = get_compute_capability(self.device)
+            if major != 9:
+                raise ValueError(
+                    "FP8 kv_data_type for MLA requires an SM90 (Hopper) device, "
+                    f"got SM{major}{minor}."
+                )
+            # The FP8 KV path was developed and tested for BF16 Q; while
+            # `vec_cast<half, __nv_fp8_e4m3>` exists in vec_dtypes.cuh, the
+            # FP16 Q + FP8 KV combination has no test coverage in this PR.
```
saltyminty: Seems like unnecessary comment content (likely a remnant of iterating with agents). The comments in this section can probably be cleaned up/simplified in general.

flashinfer/mla/_core.py
```diff
@@ -1668,6 +1724,9 @@ def run(
         page_table: Optional[torch.Tensor] = None,
         return_lse_base_on_e: bool = False,
         o_scale: Optional[float] = None,
+        *,
```
saltyminty, line 1720: scales would also need to be added to the trace templates (flashinfer/trace/templates/attention.py::mla_paged_decode_trace)

tests/attention/test_deepseek_mla_fp8_kv.py
```diff
@@ -0,0 +1,791 @@
+"""
```
saltyminty: Not blocking but is there a reason for creating a new `test_deepseek_mla_fp8_kv.py` vs adding to the existing `test_deepseek_mla.py` file?

Also, I think many of these test cases are unnecessary long term, i.e. test_fp8_kv_plan_rejects_unsupported_kv_dtype, test_fp8_kv_plan_rejects_unsupported_head_dim, test_fp8_kv_plan_rejects_fp16_q, etc.

qsang-nv: Thanks. I trimmed the plan-time reject tests (kv_dtype allowlist + head_dim) — those are guarded by downstream C++ static_assert / JIT crash, so the Python tests were UX-only. Kept the 4 reject tests that protect silent-wrong-output paths (Q dtype, run() tensor dtype mismatch, missing scales, keyword-only marker) since none of those have a downstream backstop. 15 → 7 cases. I also moved all tests to `test_deepseek_mla.py`.

flashinfer/mla/_core.py
qsang-nv: Done

flashinfer/mla/_core.py
qsang-nv: Done

saltyminty (approved):

saltyminty (approved):
Approved conditional on CI

PR #3701, opened 2026-06-23
feat: allow using NVL a2a kernels without MNNVL as long as all peers are within the same node / intra-node NVL domain

jdebache:
## 📌 Description

**What:** Let the NVLink all-to-all / MNNVL allreduce-fusion kernels run on a **single
node without multi-node NVLink fabric (IMEX)**, as long as every peer shares the same
intra-node NVLink domain — previously these paths required full MNNVL fabric support and
were unusable otherwise.

**How:** `MnnvlMemory` now probes CUDA **FABRIC** handle support per rank and AND-reduces
it across the group (all ranks must pick the same handle path or the rendezvous
deadlocks). When fabric is unavailable, it exports the shareable memory as
`CU_MEM_HANDLE_TYPE_POSIX_FILE_DESCRIPTOR` and exchanges the fds over **AF_UNIX
`SCM_RIGHTS`** instead of fabric handles. `MNNVLAllReduceFusionWorkspace` also fails fast
with an actionable error when the multicast pointer is null (hardware without NVLink
multicast) instead of proceeding silently.

This PR also carries the **comm-backend refactor** that makes the above clean:

- New `flashinfer/comm/abstractions.py` — `CommBackend` as a structural `typing.Protocol`
  (leaf module, no inheritance).
- New `flashinfer/comm/comm_backend.py` — `MPIBackend` / `TorchDistBackend` adapters;
  fixes an inverted `MPIBackend.Split` (it mutated `self` into the sub-group and returned
  a fresh `COMM_WORLD` adapter, so `set_comm_from_config` got the wrong communicator).
- New `flashinfer/comm/fd_exchange.py` — the single home for fd passing (AF_UNIX
  `SCM_RIGHTS`, no `CAP_SYS_PTRACE`), deduplicating three prior copies of the logic.
- `flashinfer/comm/mnnvl.py` — handle exchange rewritten from an ABC hierarchy to a frozen
  dataclass + free functions; MNNVL capability probes (`is_multicast_supported`,
  `all_ranks_agree`, `all_ranks_support_mnnvl`) consolidated out of `allreduce.py`;
  `CommBackend.allgather` now returns a tuple.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

New CPU-only unit tests (no GPU required):

- `tests/comm/test_fd_exchange.py` — `exchange_fds` / `broadcast_fd` + the low-level
  `SCM_RIGHTS` primitives (15 tests, forkserver multiprocess).
- `tests/comm/test_comm_backend.py` — `TorchDistBackend` over real gloo across processes,
  fake-comm `MPIBackend` adapter tests, and a real mpi4py test (`mpirun -np N`). Gated to
  **skip** cleanly when gloo / mpi4py are absent.

## Reviewer Notes

- The AND-reduce of FABRIC support in `MnnvlMemory._resolve_fabric_support`: correctness
  hinges on **every** rank calling it unconditionally and agreeing on FABRIC vs POSIX-fd,
  or the handle-exchange rendezvous deadlocks.
- The `MPIBackend.Split` semantics fix (returns the sub-group, leaves `self` unchanged).
- The multiprocess gloo test is intentionally kept at `world_size=2` with the `spawn`
  start method; larger fork/forkserver meshes are unsafe here because the eager
  `import flashinfer` starts native threads that poison a fork.

### DIFF
--- a/flashinfer/comm/allreduce.py
+++ b/flashinfer/comm/allreduce.py
@@ -70,7 +70,12 @@ from .trtllm_ar import trtllm_moe_finalize_allreduce_fusion
 
 from .mapping import Mapping
 
-from .mnnvl import CommBackend, SymmDeviceMemory
+from .mnnvl import (
+    CommBackend,
+    SymmDeviceMemory,
+    all_ranks_support_mnnvl,
+    is_multicast_supported,
+)
 
 # Note: AllReduceFusionPattern and QuantizationSFLayout are pseudo-types (classes with int constants)
 # Import them for runtime use but type hint as int for mypy compatibility
@@ -274,22 +279,6 @@ def _trtllm_workspace_check(
     return world_size <= 16
 
 
-def _mnnvl_workspace_check(
-    backend: str,
-    world_size: int,
-    rank: int,
-    max_token_num: int,
-    hidden_dim: int,
-    dtype: torch.dtype,
-) -> bool:
-    """
-    Check if mnnvl backend CAN be used for workspace creation.
-
-    """
-
-    return True
-
-
 # ============================================================================
 # HEURISTIC - Performance-based backend selection
 # ============================================================================
@@ -461,13 +450,9 @@ def create_allreduce_fusion_workspace(
             dtype=dtype,
         ):
             suitable_backends.append("trtllm")
-        if _mnnvl_workspace_check(
-            backend=backend,
-            world_size=world_size,
-            rank=rank,
-            max_token_num=max_token_num,
-            hidden_dim=hidden_dim,
-            dtype=dtype,
+        local_mnnvl_supported = is_multicast_supported(torch.cuda.current_device())
+        if all_ranks_support_mnnvl(
+            local_mnnvl_supported, world_size, comm_backend, group
         ):
             suitable_backends.append("mnnvl")
 
--- a/flashinfer/comm/comm_backend.py
+++ b/flashinfer/comm/comm_backend.py
@@ -20,6 +20,7 @@ so any object exposing the same methods (rank/size/allgather/bcast/barrier/
 Split) is a valid backend. Nothing here depends on CUDA.
 """
 
+from collections.abc import Sequence
 from typing import Any
 
 
@@ -83,6 +84,25 @@ class MPIBackend:
         split = MPIBackend()
         split._mpicomm = self._mpicomm.Split(color, key)
         return split
+
+
+def _split_partition(all_info: Sequence[tuple[int, int, int]]) -> list[list[int]]:
+    """Full color partition for an MPI-style ``Split``.
+
+    ``all_info`` is the allgathered ``(color, key, global_rank)`` from every rank.
+    Returns one rank-list per color (colors sorted ascending); within each color,
+    ranks are ordered by ``(key, global_rank)`` -- lower key first, ties broken by
+    rank, matching ``MPI_Comm_split``. The result is independent of the calling
+    rank: every rank must build the *same* partition, because
+    ``torch.distributed.new_group`` names each sub-group by a global counter (not
+    by its member ranks), so ranks that create only their own sub-group get
+    colliding rendezvous keys and deadlock.
+    """
+    colors = sorted({c for c, _, _ in all_info})
+    return [
+        [r for _, r in sorted((k, r) for c, k, r in all_info if c == col)]
+        for col in colors
+    ]
 
 
 class TorchDistBackend:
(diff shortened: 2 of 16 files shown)


flashinfer/comm/mnnvl.py
```diff
@@ -484,6 +557,122 @@ def new_mnnvl_memory_address(mapping: Mapping, size: int):
...
+        comm_size = comm.Get_size()
+
+        # Place the rendezvous socket inside a private, per-exchange directory
+        # created with 0700 permissions (mkdtemp also gives it an unpredictable
+        # name).  This prevents another local process from pre-binding the path
+        # or connecting to our listener to inject a bogus handle fd: only the
+        # owning user can traverse the directory, and the peer ranks of the same
+        # job share that uid.  A predictable name under the world-writable temp
+        # dir would otherwise be open to squatting/impersonation.
+        xchg_dir = tempfile.mkdtemp(prefix=f"cuda_fd_xchg_{os.getpid()}_{comm_rank}_")
```
djns99: No objection to the precaution, but is this security surface a real threat?
If I understand the cases for this:

1. Random name: prevent squatting/accidental collisions
2. Permission 0700: Prevent other users injecting malicious data

Whats not covered - anything malicious under the same user

flashinfer/comm/mnnvl.py
```diff
@@ -484,6 +557,122 @@ def new_mnnvl_memory_address(mapping: Mapping, size: int):
...
+        sock_path = os.path.join(xchg_dir, "fd.sock")
+
+        # try/finally guarantees the private dir + socket are removed on every
+        # exit path (including the connect-timeout RuntimeError below); a unique
+        # mkdtemp name is not self-healing across runs the way the old fixed
+        # name was, so a leak here would accumulate stale 0700 dirs under /tmp.
```
djns99: Would `with tempfile.TemporaryDirectory() as tmpdirname:` not be a more robust solution?

bobboli (commented):
Left one focused comment on the remaining backend-selection consistency issue.

flashinfer/comm/allreduce.py
```diff
@@ -223,9 +223,28 @@ def _mnnvl_workspace_check(
...
+        from flashinfer.cuda_utils import checkCudaErrors
+
+        device_idx = torch.cuda.current_device()
+        multicast_supported = checkCudaErrors(
+            cuda.cuDeviceGetAttribute(
+                cuda.CUdevice_attribute.CU_DEVICE_ATTRIBUTE_MULTICAST_SUPPORTED,
+                device_idx,
+            )
+        )
+        return multicast_supported != 0
```
bobboli: This MNNVL backend check is local, but `backend="auto"` selection needs to be consistent across all ranks. If one rank returns false here while another returns true, ranks can construct different workspace backend types and diverge. Can we allgather this capability over the same process group / comm backend and only enable MNNVL when all ranks report multicast support? This would mirror the communicator-wide `_resolve_fabric_support()` decision in `MnnvlMemory`.

flashinfer/comm/mnnvl.py
jdebache: I think it's unlikely that this could be used in a real-world attack, but the cost of handling it is minimal, and, I think, good practice, since the only component that needs to be able to identify this file is this one (e.g. it doesn't need to be shared externally for communication).

flashinfer/comm/mnnvl.py
jdebache: Done.

bobboli (approved):

flashinfer/comm/allreduce.py
jdebache: Done.

flashinfer/comm/allreduce.py
```diff
@@ -223,9 +223,65 @@ def _mnnvl_workspace_check(
...
+        # different but co-spanning comm (e.g. the MNNVL constructor's MPI
+        # barrier).
+        import torch.distributed as dist
+
+        if dist.is_initialized():
+            comm = TorchDistBackend(group=group)
+        else:
+            comm = MPIBackend()
+
+    return all(comm.allgather(local_supported))
```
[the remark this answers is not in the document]
jdebache: Addressed.

flashinfer/comm/comm_backend.py
```diff
@@ -0,0 +1,172 @@
...
+        # then sort), so a lazy generator won't do, and its parameter is list[int].
+        my_group_ranks = [
+            r
+            for _, r in sorted(
+                ((k, r) for c, k, r in all_info if c == color),
+                key=lambda kr: kr[0],
+            )
+        ]
+
+        new_group = self._dist.new_group(ranks=my_group_ranks)  # pyright: ignore
```
bobboli: `torch.distributed.new_group()` does not have `MPI_Comm_split` semantics: all world ranks must create every subgroup in the same order and pass the same `ranks` list. Here ranks with different colors call it once with different lists, which can hang. I think the correct way is to create all color groups in deterministic order on every rank, then return the group matching this rank's color. A 4-rank test with two ranks per color would cover this case.

dierksen:
@flashinfer-bot run

aleozlx (approved):

PR #3705, opened 2026-06-23
Fix cudart library resolution when stubs are loaded

### LINKED ISSUE #3676: import flashinfer.comm fails at import time when a co-loaded library matches "libcudart" (e.g. tilelang libcudart_stub.so) — find_loaded_library picks the non-runtime lib
## Summary

`import flashinfer.comm` can fail at **import time** with `AttributeError: undefined symbol: cudaDeviceReset` when another package legitimately loaded in the same process provides a shared object whose filename contains `libcudart` but which is **not** the full CUDA runtime.

At import, `flashinfer/comm/cuda_ipc.py` constructs a module-level `CudaRTLibrary()` (line 194). With no explicit path, `CudaRTLibrary.__init__` calls `find_loaded_library("libcudart")`, which returns the **first** `/proc/self/maps` line containing the substring `libcudart`. Because `/proc/self/maps` is address-sorted, that's whichever `libcudart`-named mapping has the lowest base address — which is not necessarily the real CUDA runtime. If the selected library doesn't export `cudaDeviceReset`, `__init__` raises.

A concrete trigger is [`tilelang`](https://github.com/tile-ai/tilelang), which ships and loads `tilelang/lib/libcudart_stub.so` (a cudart shim that does not export `cudaDeviceReset`). `tilelang` and `flashinfer-python` are independent packages; both end up in the same process when a downstream framework (e.g. vLLM) imports both. This is legitimate usage — the user has no way to know one library's presence will break another's import.

## Minimal reproduction (arch-independent; verified on H100)

No special model/kernels needed — just three imports in a fresh process:

```bash
# flashinfer-python==0.6.12, tilelang==0.1.9, torch==2.11.0+cu130 (CUDA 13.0), Python 3.12
python3 -c "import torch; import tilelang; import flashinfer.comm"
```

Result:

```
  File ".../flashinfer/comm/__init__.py", line 1, in <module>
    from .cuda_ipc import CudaRTLibrary, create_shared_buffer, free_shared_buffer
  File ".../flashinfer/comm/cuda_ipc.py", line 194, in <module>
    cudart = CudaRTLibrary()
  File ".../flashinfer/comm/cuda_ipc.py", line 134, in __init__
    f = getattr(self.lib, func.name)
  File "/usr/lib/python3.12/ctypes/__init__.py", line 392, in __getattr__
    func = self.__getitem__(name)
  File "/usr/lib/python3.12/ctypes/__init__.py", line 397, in __getitem__
    func = self._FuncPtr((name_or_ordinal, self))
```

The failure is load-order dependent: it occurs when the non-runtime `libcudart`-named library maps at a lower address than the real `libcudart.so.13` (the case in a single process that imports `tilelang` before `flashinfer.comm`).

## Workaround (no code change)

Preloading the real CUDA runtime so it is the first `libcudart` match resolves it:

```bash
LD_PRELOAD=/usr/local/cuda-13.0/lib64/libcudart.so.13 \
  python3 -c "import torch; import tilelang; import flashinfer.comm"   # imports cleanly
```

Verified on the same H100 environment.

## Suggested robustness improvement

Make `libcudart` resolution robust to other libraries that happen to share the name, so co-loading a `libcudart`-named shim doesn't break the import. Options:

- In `find_loaded_library` (or before accepting its result), select the `libcudart` mapping that actually provides the runtime API rather than the first substring match — e.g. skip candidates that don't resolve the symbols `CudaRTLibrary` needs (`cudaDeviceReset`, etc.), continuing the scan to the real runtime.
- Optionally prefer/allow an explicit `so_file` for the module-level `CudaRTLibrary()` (the constructor already accepts `so_file`), as vLLM's analogous `CudaRTLibrary` in `cuda_wrapper.py` does.

## Environment

| Component | Version |
|---|---|
| flashinfer-python | 0.6.12 |
| tilelang | 0.1.9 |
| torch | 2.11.0+cu130 |
| CUDA | 13.0 |
| Python | 3.12 |
| GPU | NVIDIA H100 (reproduces on any CUDA GPU; this is a host-side library-resolution issue, not arch-specific) |
| OS | Ubuntu 24.04 |

yinli-systems:
## Summary

Fixes #3676.

`CudaRTLibrary()` previously resolved `libcudart` by taking the first `/proc/self/maps` entry whose filename matched `libcudart*`. That can select a co-loaded stub such as `tilelang/lib/libcudart_stub.so`, which has the right filename shape but does not export the CUDA runtime symbols FlashInfer needs. Importing or constructing `flashinfer.comm.CudaRTLibrary` then fails before it can reach the real CUDA runtime.

This change keeps explicit `so_file` handling strict, but makes automatic discovery try every loaded `libcudart*` candidate and accept the first one that exports the required CUDA runtime functions.

## Tests

- `PYTHONPYCACHEPREFIX=/private/tmp/flashinfer-pycache /usr/bin/python3 -m py_compile flashinfer/comm/cuda_ipc.py tests/comm/test_cuda_ipc_cudart_resolution.py`
- `PYTHONPYCACHEPREFIX=/private/tmp/flashinfer-pycache /usr/bin/python3 -m pytest --noconftest -q tests/comm/test_cuda_ipc_cudart_resolution.py`
- `UV_CACHE_DIR=/private/tmp/uv-cache-flashinfer UV_TOOL_DIR=/private/tmp/uv-tools-flashinfer uvx ruff check flashinfer/comm/cuda_ipc.py tests/comm/test_cuda_ipc_cudart_resolution.py`
- `UV_CACHE_DIR=/private/tmp/uv-cache-flashinfer UV_TOOL_DIR=/private/tmp/uv-tools-flashinfer uvx ruff format --check flashinfer/comm/cuda_ipc.py tests/comm/test_cuda_ipc_cudart_resolution.py`
- `git diff --check`

Note: the focused pytest uses `--noconftest` locally because this machine does not have `tvm_ffi`, and the repository-wide `tests/conftest.py` imports the full FlashInfer JIT stack. The added test directly loads `flashinfer/comm/cuda_ipc.py` and only exercises the host-side cudart resolution logic.

## Summary by CodeRabbit

* **Refactor**
  * Improved CUDA runtime library discovery and dynamic loading to reliably pick a compatible `cudart` when multiple candidates exist, ensuring required exports are present before use.

* **Tests**
  * Added automated coverage for CUDA runtime resolution, verifying that auto-selection prefers a valid library over stubs missing required symbols, and that explicit paths fail when symbols are absent (without caching incompatible results).

### DIFF
--- a/flashinfer/comm/cuda_ipc.py
+++ b/flashinfer/comm/cuda_ipc.py
@@ -119,24 +119,53 @@ class CudaRTLibrary:
     #  to the corresponding dictionary
     path_to_dict_mapping: Dict[str, Dict[str, Any]] = {}
 
+    @staticmethod
+    def _load_exported_functions(so_file: str) -> Dict[str, Any]:
+        """Load cudart and bind every runtime symbol required by this wrapper."""
+        lib = CudaRTLibrary.path_to_library_cache.get(so_file)
+        if lib is None:
+            lib = ctypes.CDLL(so_file)
+        _funcs = {}
+        for func in CudaRTLibrary.exported_functions:
+            f = getattr(lib, func.name)
+            f.restype = func.restype
+            f.argtypes = func.argtypes
+            _funcs[func.name] = f
+        CudaRTLibrary.path_to_library_cache[so_file] = lib
+        CudaRTLibrary.path_to_dict_mapping[so_file] = _funcs
+        return _funcs
+
+    @staticmethod
+    def _find_loadable_cudart() -> str:
+        """Return the first loaded libcudart candidate that exports required symbols."""
+        candidates = []
+        with open("/proc/self/maps") as f:
+            for line in f:
+                if "libcudart" not in line or "/" not in line:
+                    continue
+                path = line[line.index("/") :].strip()
+                filename = path.split("/")[-1]
+                if not filename.rpartition(".so")[0].startswith("libcudart"):
+                    continue
+                if path not in candidates:
+                    candidates.append(path)
+        for path in candidates:
+            try:
+                CudaRTLibrary._load_exported_functions(path)
+            except (AttributeError, OSError):
+                continue
+            return path
+        raise AssertionError("libcudart is not loaded in the current process")
+
     def __init__(self, so_file: Optional[str] = None):
         if so_file is None:
-            so_file = find_loaded_library("libcudart")
-            assert so_file is not None, "libcudart is not loaded in the current process"
-        if so_file not in CudaRTLibrary.path_to_library_cache:
-            lib = ctypes.CDLL(so_file)
-            CudaRTLibrary.path_to_library_cache[so_file] = lib
-        self.lib = CudaRTLibrary.path_to_library_cache[so_file]
+            so_file = self._find_loadable_cudart()
 
         if so_file not in CudaRTLibrary.path_to_dict_mapping:
-            _funcs = {}
-            for func in CudaRTLibrary.exported_functions:
-                f = getattr(self.lib, func.name)
-                f.restype = func.restype
-                f.argtypes = func.argtypes
-                _funcs[func.name] = f
-            CudaRTLibrary.path_to_dict_mapping[so_file] = _funcs
-        self.funcs = CudaRTLibrary.path_to_dict_mapping[so_file]
+            self.funcs = self._load_exported_functions(so_file)
+        else:
+            self.funcs = CudaRTLibrary.path_to_dict_mapping[so_file]
+        self.lib = CudaRTLibrary.path_to_library_cache[so_file]
 
     def CUDART_CHECK(self, result: cudaError_t) -> None:
         if result != 0:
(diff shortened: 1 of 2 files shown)


yinli-systems:
Thanks, this is addressed in the current head. `_load_exported_functions()` now only updates `path_to_library_cache` and `path_to_dict_mapping` after every required cudart symbol is resolved successfully. The explicit-path regression test also verifies that a stub missing `cudaDeviceReset` raises and leaves both caches untouched.

yinli-systems:
Closing this to consolidate the fix on #3930, which is now the actively claimed implementation and has current stub-resident GPU integration evidence. After re-auditing both approaches, the filename matcher there is the smaller change at the actual selection boundary, while this PR adds a broader symbol-probing/cache refactor. I do not see a unique reason to keep two competing resolver implementations open. The symbol-completeness and cache-pollution tests from this PR remain useful context if reviewers want additional defense-in-depth coverage on #3930.

PR #3707, opened 2026-06-23
Yanqinz/autotuner tactic

yanqinz2:
This PR contains a series of changes to make cudnn GEMM backend stable during autotuning.
1. Bump flashinfer dependency requirement of cudnn frontend version from v1.13.0 to v1.25.0
2. Replace fixed plan index cached in autotuner with explicit cudnn engine/knob combination
3. Add a fallback plan for cudnn backend when cached execution plan is not available to avoid abrupt crash

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Refactor**
  * Unified cuDNN GEMM execution around a **tactic → plan_index** workflow, including discovery of engine/knob tactics, consistent plan selection, and plan-specific workspace sizing (covers FP4/BF16 and related paths).
* **Bug Fixes**
  * Added resilience for tactic-based execution: if a tactic fails, the system warns and automatically falls back to the default tactic.
* **Tests**
  * Updated autotuner tests to validate cuDNN engine/knob tactics and to verify expected warning behavior; expanded MXFP8 bmm warning assertions.
* **Dependencies**
  * Increased minimum `nvidia-cudnn-frontend` to **≥ 1.25.0**.
* **Chores**
  * Broadened autotuner tactic handling to support non-integer representations; skipped legacy cuDNN integer tactic cache entries when autotuning.

### DIFF
--- a/flashinfer/autotuner/autotuner.py
+++ b/flashinfer/autotuner/autotuner.py
@@ -487,8 +487,8 @@ class FakeTensor:
 class TunableRunner(ABC):
     @abstractmethod
     def get_valid_tactics(
-        self, inputs: list[torch.Tensor], profile: OptimizationProfile
-    ) -> list[int]:
+        self, inputs: List[torch.Tensor], profile: OptimizationProfile
+    ) -> List[Any]:
         """One tactic corresponding to one cuda kernel normally, but how to interpret the meaning
         of tactic is pure internal details of the runner.
 
@@ -527,8 +527,8 @@ class TunableRunner(ABC):
     @abstractmethod
     def forward(
         self,
-        inputs: list[torch.Tensor],
-        tactic: int = -1,
+        inputs: List[torch.Tensor],
+        tactic: Any = -1,
         do_preparation: bool = False,
         **kwargs,  # all others are keyword args only
     ) -> Any:
@@ -1314,7 +1314,7 @@ class AutoTuner:
         tuning_config: TuningConfig,
         inputs: list[torch.Tensor],
         **kwargs,
-    ) -> tuple[TunableRunner, int]:
+    ) -> Tuple[TunableRunner, Any]:
         """Choose the best runner and tactic combination through performance profiling.
 
         Args:
@@ -1325,9 +1325,9 @@ class AutoTuner:
             **kwargs: Arbitrary keyword arguments, will be passed to get_valid_tactics and forward method of each runner
 
         Returns:
-            Tuple[TunableRunner, int]: A tuple containing:
+            Tuple[TunableRunner, Any]: A tuple containing:
                 - The selected runner implementation
-                - The best tactic ID for that runner (-1 if using fallback)
+                - The best tactic for that runner (-1 if using fallback)
 
         Note:
             The method profiles different implementations and tactics to find the
--- a/flashinfer/autotuner/autotuner.py
+++ b/flashinfer/autotuner/autotuner.py
@@ -2168,13 +2168,31 @@ class AutoTuner:
                 )
                 return False
 
+        skipped_legacy_cudnn_tactics = 0
         with self._lock:
             for key, value in configs.items():
                 runner_name = value[0]
                 tactic = _json_to_tactic(value[1])
+                if (
+                    runner_name.startswith("Cudnn")
+                    and isinstance(tactic, int)
+                    and tactic >= 0
+                ):
+                    skipped_legacy_cudnn_tactics += 1
+                    continue
                 self._file_configs[key] = (runner_name, tactic)
 
-        logger.info(f"[Autotuner]: Loaded {len(configs)} configs from {path}")
+        if skipped_legacy_cudnn_tactics:
+            logger.warning(
+                f"[Autotuner]: Skipped {skipped_legacy_cudnn_tactics} legacy "
+                "cuDNN config(s) using integer plan-index tactics. They will "
+                "be re-tuned when autotuning is enabled."
+            )
+
+        logger.info(
+            f"[Autotuner]: Loaded {len(configs) - skipped_legacy_cudnn_tactics} "
+            f"configs from {path}"
+        )
         return True
 
     def _prepare_input_tensors_with_batches(
--- a/flashinfer/autotuner/autotuner.py
+++ b/flashinfer/autotuner/autotuner.py
@@ -487,8 +487,8 @@ class FakeTensor:
 class TunableRunner(ABC):
     @abstractmethod
     def get_valid_tactics(
-        self, inputs: List[torch.Tensor], profile: OptimizationProfile
-    ) -> List[Any]:
+        self, inputs: list[torch.Tensor], profile: OptimizationProfile
+    ) -> list[Any]:
         """One tactic corresponding to one cuda kernel normally, but how to interpret the meaning
         of tactic is pure internal details of the runner.
 
@@ -527,7 +527,7 @@ class TunableRunner(ABC):
     @abstractmethod
     def forward(
         self,
-        inputs: List[torch.Tensor],
+        inputs: list[torch.Tensor],
         tactic: Any = -1,
         do_preparation: bool = False,
         **kwargs,  # all others are keyword args only
@@ -1314,7 +1314,7 @@ class AutoTuner:
         tuning_config: TuningConfig,
         inputs: list[torch.Tensor],
         **kwargs,
-    ) -> Tuple[TunableRunner, Any]:
+    ) -> tuple[TunableRunner, Any]:
         """Choose the best runner and tactic combination through performance profiling.
 
         Args:
--- a/flashinfer/gemm/gemm_base.py
+++ b/flashinfer/gemm/gemm_base.py
@@ -2300,8 +2300,6 @@ def _cudnn_bf16_gemm_usable_or_skip(
 
 def _is_cublas_fp4_available_in_cudnn():
     """Check if cuBLAS backend for FP4 GEMM is available in cuDNN."""
-
-    # Check cuDNN backend version for FP4 support (requires cudnn_version == 9.11.1 or cudnn_version >= 9.13)
     backend_version = cudnn.backend_version()
     CUDNN_VERSION_9_11_1 = 91101
     CUDNN_VERSION_9_13_0 = 91300
@@ -2315,71 +2313,49 @@ def _check_cudnn_override_shape_availability():
     """Raise if the installed cuDNN backend does not support is_override_shape_enabled."""
     _check_cudnn_availability()
     backend_version = cudnn.backend_version()
-    if backend_version < 92100:
+    if backend_version < 92301:
         raise RuntimeError(
-            f"cuDNN override-shape GEMM requires backend version >= 92100 (9.21.0), "
+            f"cuDNN override-shape GEMM requires backend version >= 92301 (9.23.1), "
             f"found {backend_version}. "
             f"Please upgrade cuDNN: pip install --upgrade nvidia-cudnn-cu12 nvidia-cudnn-frontend"
         )
-    try:
-        version_str = cudnn.__version__
-        major, minor = map(int, version_str.split(".")[:2])
-        required_frontend_version = (1, 24) if backend_version >= 92300 else (1, 20)
-        if (major, minor) < required_frontend_version:
-            raise RuntimeError(
-                f"cuDNN override-shape GEMM requires cudnn-frontend version >= "
-                f"{required_frontend_version[0]}.{required_frontend_version[1]}, found {version_str}. "
-                f"Please upgrade: pip install --upgrade nvidia-cudnn-frontend"
-            )
-    except (AttributeError, ValueError, IndexError) as e:
-        raise RuntimeError(
-            "Unable to determine cudnn-frontend version. "
-            "Override-shape GEMM requires cudnn-frontend >= 1.20, or >= 1.24 with cuDNN backend >= 9.23.0"
-        ) from e
 
 
 def _is_cudnn_override_shape_available() -> bool:
     """Return True if the installed cuDNN backend supports is_override_shape_enabled."""
     if not CUDNN_AVAILABLE:
         return False
-    try:
-        backend_version = cudnn.backend_version()
-        if backend_version < 92100:
-            return False
-        version_str = cudnn.__version__
-        major, minor = map(int, version_str.split(".")[:2])
-        required_frontend_version = (1, 24) if backend_version >= 92300 else (1, 20)
-        return (major, minor) >= required_frontend_version
-    except Exception:
+    if cudnn.backend_version() < 92301:
         return False
+    return True
 
 
-def _get_cudnn_workspace_size(graph, tactic: int) -> int:
-    if tactic < 0:
+def _get_cudnn_workspace_size(graph, plan_index: int) -> int:
+    if plan_index < 0:
         return graph.get_workspace_size()
-    return graph.get_workspace_size_plan_at_index(tactic)
+    return graph.get_workspace_size_plan_at_index(plan_index)
 
 
 def _get_cudnn_override_shape_workspace_size(
     graph,
-    tactic: int,
+    plan_index: int,
     cudnn_handle,
     override_uids,
     override_shapes,
     override_strides,
 ) -> int:
     if cudnn.backend_version() >= 92300:
-        if tactic < 0:
+        if plan_index < 0:
             return graph.get_workspace_size(
                 cudnn_handle, override_uids, override_shapes, override_strides
             )
         return graph.get_workspace_size_plan_at_index(
-            tactic, cudnn_handle, override_uids, override_shapes, override_strides
+            plan_index, cudnn_handle, override_uids, override_shapes, override_strides
         )
     else:
-        if tactic < 0:
+        if plan_index < 0:
             return graph.get_workspace_size()
-        return graph.get_workspace_size_plan_at_index(tactic)
+        return graph.get_workspace_size_plan_at_index(plan_index)
 
 
 def _cudnn_graph_engine_knob_tactics(graph) -> List[tuple]:
--- a/flashinfer/gemm/gemm_base.py
+++ b/flashinfer/gemm/gemm_base.py
@@ -3728,11 +3728,9 @@ def build_cudnn_gemm_bf16_graph(
     bias_is_not_none,
     bias_shape,
     bias_stride,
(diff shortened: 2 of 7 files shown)


flashinfer/gemm/gemm_base.py
```diff
@@ -2166,71 +2165,148 @@ def _check_cudnn_override_shape_availability():
...
+                "cuDNN GEMM engine/knob tactic did not match any built execution "
+                "plan; falling back to default tactic=-1.",
+                stacklevel=3,
+            )
+    else:
+        plan_index = tactic
+
+    if plan_index >= graph.get_execution_plan_count():
+        plan_index = -1
+    return plan_index
```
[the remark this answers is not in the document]
yanqinz2: Typically there will be less than 10 execution plan, and it should hit the matching plan in the very beginning. Should not be an obvious overhead.

flashinfer/gemm/gemm_base.py
```diff
@@ -2631,15 +2693,9 @@ def build_cudnn_gemm_fp4_graph_override_shape(
...
-    graph.create_execution_plans([cudnn.heur_mode.A, cudnn.heur_mode.B])
-
-    if alpha_is_not_none and not _is_cublas_fp4_available_in_cudnn():
-        graph.deselect_engines(["eng0"])
-
-    graph.check_support()
-    graph.build_plans(policy)
+    _finalize_cudnn_graph_for_tactic(
+        graph, tactic, [cudnn.heur_mode.A, cudnn.heur_mode.FALLBACK]
+    )
```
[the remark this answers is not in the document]
yanqinz2: No need

flashinfer/gemm/gemm_base.py
```diff
@@ -2166,71 +2165,148 @@ def _check_cudnn_override_shape_availability():
...
-        if backend_version < 92100:
-            return False
-        version_str = cudnn.__version__
-        major, minor = map(int, version_str.split(".")[:2])
-        required_frontend_version = (1, 24) if backend_version >= 92300 else (1, 20)
-        return (major, minor) >= required_frontend_version
-    except Exception:
+    if cudnn.backend_version() < 92301:
         return False
+    return True
```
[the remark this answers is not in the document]
yanqinz2: Added

YangXu1990uiuc:
Overall, the engine/knob tactic representation and the cuDNN version gating look reasonable. 

I have the following requested changes/hardening suggestions:

### 1. Fix the MXFP8 test so it actually exercises the override-shape path

`test_bmm_mxfp8.py` currently constructs B as:

```python
mat2 = torch.randn([b, n, k]).transpose(-2, -1).contiguous()
mat2_q, mat2_scale = mxfp8_quantize(mat2, ...)
```

This produces a row-major logical `[B, K, N]` tensor and quantizes along N. However, the API contract requires B to be column-major, with quantization performed along the contraction K dimension.

It should instead be constructed as:

```python
weight = torch.randn([b, n, k], device="cuda", dtype=input_dtype)
weight_q, weight_scale = mxfp8_quantize(weight, is_sf_swizzled_layout)

B = weight_q.transpose(-2, -1)
reference = torch.bmm(input_mat, weight.transpose(-2, -1))
```

Please also assert the expected stride, and ideally ensure that no fallback warning is emitted. The current test passes numerically only because the runner silently falls back to the fixed-shape path, so it does not cover the new override-shape execution path.

This test bug predates this PR, but the new fallback behavior makes it especially easy for the test to pass without exercising the intended code.

### 2. Do not treat a failed tactic's fallback execution as a successful tactic during profiling

Fallback is desirable for inference robustness, but during autotuning, a failed tactic should be reported back to the autotuner as failed rather than being timed as `tactic=-1`.

Otherwise, the autotuner may persist a tactic that never executed and actually measured the fallback implementation. A preferable policy is:

- During profiling: propagate the tactic failure so the autotuner marks it unsupported.
- After all candidate tactics fail: use `tactic=-1` as the final fallback.
- Outside profiling/in normal inference: retain the current warning-and-fallback behavior.

This preserves serving reliability while ensuring that tuning results represent the requested tactics.

### 3. Include BF16 matrix layout in the persistent cache key

The in-memory runner identity distinguishes the A/B layout flags, but the persisted file-cache key currently does not distinguish:

```python
is_a_k_major
is_b_k_major
```

As a result, different NT/TN/layout configurations can reuse a tactic tuned for another layout after serialization. Please include these flags in `get_cache_key_extras()` or otherwise encode them in the persistent key.

This may be uncommon, but cache-key bugs are extremely difficult to diagnose because they produce stable, configuration-dependent performance regressions.

### 4. Version the persisted cuDNN tactic representation

This PR changes cuDNN tactics from integer plan indices to engine/knob descriptions. Please add a tactic-schema/cache-format version and invalidate or explicitly reject legacy non-negative integer cuDNN tactics.

FlashInfer version metadata helps for normal releases, but it does not fully protect development builds, same-version branch changes, manually copied configs, or wildcard metadata. A persisted plan index must never be interpreted as a tactic under the new representation.

yanqinz2:
Regarding above comment from @YangXu1990uiuc 
1. Fixed.
2. For now, I’d prefer not to change this behavior. Propagating tactic failures differently during profiling would require a relatively broad autotuner change, and I don’t want to take that risk in this PR. In practice, this fallback path should be rare, so I think the current behavior is acceptable for now.
3. Fixed.
4. For the cache format/versioning concern, I think the cleaner path after this refactor is to ask users to regenerate their autotune cache rather than continuing to support legacy cuDNN cache entries. The old integer plan-index representation is no longer compatible with the new engine/knob tactic format, so re-autotuning is the safest and simplest migration path.

tests/autotuner/test_autotuner_bmm_fp8.py
```diff
@@ -6,6 +6,7 @@
 from flashinfer.gemm.gemm_base import (
     _FP8_GEMM_SM100_TUNING_CONFIG,
     _cudnn_gemm_fp8_runner,
+    _is_cudnn_engine_knob_tactic,
```
dhiraj113: Is this API? Can the import of this be avoided in the test harness?

yanqinz2: Sure, removed import this internal api, and add a local helper function instead

dhiraj113:
While I am reviewing this could you please take a look at the issues that Claude has flagged - 

PR #3707 Review — cuDNN Tactic Representation Migration
This PR replaces integer plan-index tactics with stable (engine_id, knob_items) tuples for cuDNN GEMM runners (FP8, MXFP8, FP4, BF16-FP4), making autotuned plans resilient to plan-list reordering across cuDNN versions. It also unifies the graph build/execute patterns through _finalize_cudnn_graph_for_tactic and _get_cudnn_plan_index_for_tactic, adds graceful fallback on tactic failure, and bumps the cuDNN frontend requirement to ≥1.25.0. The scope is large (~600 LOC changed) and architecturally sound, but there are several issues worth addressing before merge.

Findings
1. [flashinfer/gemm/gemm_base.py:2808](vscode-webview://1u1dd2eak15tlbkvegmiiln9vditk52ur305l352mcham2jmleil/flashinfer/gemm/gemm_base.py#L2808) — build_cudnn_gemm_fp4_graph_override_shape omits deselect_eng0, silently allowing engine 0 to be autotuned on the override-shape path when it should be excluded

build_cudnn_gemm_fp4_graph correctly passes deselect_eng0=alpha_is_not_none and not _is_cublas_fp4_available_in_cudnn() to _finalize_cudnn_graph_for_tactic. The override-shape variant accepts the same alpha_is_not_none parameter but passes nothing for deselect_eng0 (defaults to False). When alpha is present and cuBLAS FP4 is unavailable in the installed cuDNN version, the override-shape graph builder exposes engine 0 to the autotuner — the exact scenario the WAR was designed to block. An autotuned (engine_id=0, ...) tactic will be stored and reapplied at inference, bypassing the guard on the non-override-shape path and potentially producing wrong results on affected cuDNN versions.
Fix: add deselect_eng0=alpha_is_not_none and not _is_cublas_fp4_available_in_cudnn() to the _finalize_cudnn_graph_for_tactic call in build_cudnn_gemm_fp4_graph_override_shape.

2. [flashinfer/gemm/gemm_base.py:2275](vscode-webview://1u1dd2eak15tlbkvegmiiln9vditk52ur305l352mcham2jmleil/flashinfer/gemm/gemm_base.py#L2275) — Frontend version check removed; mismatched frontend/backend gives a cryptic AttributeError instead of a diagnostic

_check_cudnn_override_shape_availability previously verified that the installed nvidia-cudnn-frontend version was ≥1.20 (or ≥1.24 for backend ≥9.23.0) and raised a clear RuntimeError with upgrade instructions if not. The PR removes the check entirely, relying solely on requirements.txt >=1.25.0. A user who satisfies the package floor but later upgrades only the cuDNN backend (e.g., via a separate CUDA install) will hit AttributeError: 'graph' object has no attribute 'get_engine_and_knobs_at_index' deep inside _cudnn_graph_engine_knob_tactics, with no hint that the frontend is out of date. The symptom would look like a graph API crash rather than a version mismatch.

3. [flashinfer/gemm/gemm_base.py:182](vscode-webview://1u1dd2eak15tlbkvegmiiln9vditk52ur305l352mcham2jmleil/flashinfer/gemm/gemm_base.py#L182) — _cudnn_graph_engine_knob_tactics silently drops plans that raise RuntimeError from get_engine_and_knobs_at_index, with no diagnostic

The function skips any plan index where get_engine_and_knobs_at_index raises. The AttributeError branch is effectively dead given the >=1.25.0 requirement, but the RuntimeError branch is live if the cuDNN backend raises for certain engine types. Plans dropped this way are never offered to the autotuner — the fastest kernel could be silently excluded. Previously, list(range(N)) enumerated all plan indices unconditionally. The behavior change is correct when the skipped plans are genuinely invalid, but a logger.debug() message ("Skipping plan index N in tactic enumeration: <exc>") would make silent exclusions auditable.

4. [flashinfer/gemm/gemm_base.py:206](vscode-webview://1u1dd2eak15tlbkvegmiiln9vditk52ur305l352mcham2jmleil/flashinfer/gemm/gemm_base.py#L206) — _get_cudnn_plan_index_for_tactic does an O(N) plan scan on every forward call, plus a redundant sort of already-sorted knob items

When a tactic is an engine/knob tuple, this function iterates all execution plans in the graph on every call. Graphs are LRU-cached by shape, but _get_cudnn_plan_index_for_tactic is invoked at execution time (inside execute_cudnn_gemm_*_graph), so for a hot inference loop the O(N) scan runs on every token. Additionally, target_knob_items is re-sorted at line ~2216 even though tactics stored by _cudnn_graph_engine_knob_tactics already contain pre-sorted knob tuples. The plan-index lookup could be precomputed once when the graph is finalized (e.g., stored in a dict alongside the cached graph), or the redundant sort could at least be removed.

5. [flashinfer/gemm/gemm_base.py:3187](vscode-webview://1u1dd2eak15tlbkvegmiiln9vditk52ur305l352mcham2jmleil/flashinfer/gemm/gemm_base.py#L3187) — build_cudnn_gemm_fp8_graph LRU cache halved to 1024, while tactic is now a cache key dimension

The old code keyed the cache on (shape, stride, dtype, device, policy) with policy taking one of two values (HEURISTICS_CHOICE or ALL). The new code keys on tactic, also taking one of two values (-1 or 0). The effective number of entries per shape pair doubles from 1 to 2, meaning the 1024-entry cache is equivalent to ~512 shape-pair slots. The old 2048-entry cache held ~1024. For decoding workloads that sweep through hundreds of M buckets with two tactic values each, this 2× reduction in effective shape coverage doubles eviction frequency and cuDNN graph rebuild cost.

6. [flashinfer/gemm/gemm_base.py:3519](vscode-webview://1u1dd2eak15tlbkvegmiiln9vditk52ur305l352mcham2jmleil/flashinfer/gemm/gemm_base.py#L3519) — Tactic normalization expression copy-pasted 5+ times across gemm_base.py and gemm_bf16_fp4_cudnn.py

0 if _is_cudnn_engine_knob_tactic(tactic) or tactic >= 0 else -1 (with the accompanying comment # tactic value only be 0 or -1 to hit the graph cache) appears verbatim in 5–6 _get_override_graph methods across both files. A one-line helper _tactic_for_graph_cache(tactic) -> int would keep the rule in one place and eliminate the risk of a future change being applied to only some of the copies.

7. [flashinfer/gemm/gemm_base.py:196](vscode-webview://1u1dd2eak15tlbkvegmiiln9vditk52ur305l352mcham2jmleil/flashinfer/gemm/gemm_base.py#L196) — _is_cudnn_engine_knob_tactic accepts list as well as tuple, but no production path ever produces a list tactic

_json_to_tactic recursively converts all JSON arrays to tuples before storing tactics, and _cudnn_graph_engine_knob_tactics always constructs tuples. The list branch of the isinstance(tactic, (tuple, list)) check is unreachable dead code. Keeping it creates the impression that lists are a valid tactic representation (they would be unhashable in an LRU cache if ever passed directly), and makes the type contract ambiguous. Restricting to isinstance(tactic, tuple) tightens the contract with no behavior change.

8. [flashinfer/gemm/gemm_base.py:247](vscode-webview://1u1dd2eak15tlbkvegmiiln9vditk52ur305l352mcham2jmleil/flashinfer/gemm/gemm_base.py#L247) — Integer out-of-range plan index silently clamped to -1; inconsistent with explicit warning for knob-mismatch

When _get_cudnn_plan_index_for_tactic receives an engine/knob tactic that doesn't match any plan, it emits warnings.warn. When it receives an integer plan index that is out of range (the plan_index >= get_execution_plan_count() guard), it silently falls back to -1 with no log message. A stale integer tactic (e.g., from a manually edited cache file) thus fails invisibly, making it impossible to detect cache poisoning from debug logs alone. A logger.warning (or warnings.warn) consistent with the knob-mismatch path would improve diagnosability.

flashinfer/autotuner/autotuner.py
```diff
@@ -2168,13 +2168,31 @@ def load_configs(self, path: str) -> bool:
                 )
                 return False
 
+        skipped_legacy_cudnn_tactics = 0
```
dhiraj113, line 2171: Shouldn't the autotuner be agnostic to any particular backend?

flashinfer/autotuner/autotuner.py
```diff
@@ -2168,13 +2168,31 @@ def load_configs(self, path: str) -> bool:
...
         with self._lock:
             for key, value in configs.items():
                 runner_name = value[0]
                 tactic = _json_to_tactic(value[1])
+                if (
+                    runner_name.startswith("Cudnn")
```
dhiraj113, line 2177: Are you sure you want to have code like this?

flashinfer/autotuner/autotuner.py
```diff
@@ -2168,13 +2168,31 @@ def load_configs(self, path: str) -> bool:
...
+                    and isinstance(tactic, int)
+                    and tactic >= 0
+                ):
+                    skipped_legacy_cudnn_tactics += 1
```
dhiraj113, line 2181: Is this bool assignment done through increment operation?

flashinfer/autotuner/autotuner.py
YangXu1990uiuc: needed till autotuner able to reject cache with legacy cudnn version

flashinfer/autotuner/autotuner.py
YangXu1990uiuc: needed till autotuner able to reject cache with legacy cudnn version

flashinfer/gemm/gemm_base.py
```diff
@@ -2277,71 +2276,149 @@ def _check_cudnn_override_shape_availability():
...
+        )
+        tactics.append((int(engine_id), knob_items))
+    return tactics
+
+
+def _is_cudnn_engine_knob_tactic(tactic) -> bool:
+    return isinstance(tactic, (tuple, list)) and len(tactic) == 2
+
+
+def _cudnn_knob_items_to_dict(knob_items) -> dict:
```
dhiraj113, line 2356: Is this a function that needs to be in a file called `gemm_base.py`?

flashinfer/autotuner/autotuner.py
YangXu1990uiuc: this is integer counter

YangXu1990uiuc:
the CI failure on H100 tests/trace/test_cudnn_batch_prefill_reference_correctness.py needs PR #3784, unrelated to this change

flashinfer/gemm/gemm_base.py
```diff
@@ -2277,71 +2276,149 @@ def _check_cudnn_override_shape_availability():
...
+    return {
+        cudnn.knob_type(int(knob_type)): int(value) for knob_type, value in knob_items
+    }
+
+
+def _get_cudnn_plan_index_for_tactic(graph, tactic) -> int:
```
dhiraj113, line 2362: Is this function specific to gemm? If not, move it to a different file liek `cudnn_utils.py` or something.

dhiraj113 (approved):

flashinfer/gemm/gemm_base.py
yanqinz2: Currently only gemm is using it. I would like to move it to a standalone file while another place is starting to use it.

flashinfer/gemm/gemm_base.py
yanqinz2: Same above

yanqinz2:
Regarding @dhiraj113 comments above:
1. override shape only supports cudnn >= 9.23.1, which does not need to deselect eng0
2. will bump the fe version to 1.25.0, then no fe version check is needed
3. addressed by adding a logging information
4. N is typically ~10, so I decide no more complex caching in this round
5. addressed by restoring 1024 to 2048, and also update all the graph cache to 2048 for consistency
6. addressed by abstract to a function: _tactic_for_graph_cache
7. addressed by constrains the condition
8. addressed by printing a warning info

YangXu1990uiuc:
Following up on review point #2 from this PR ("do not treat a failed tactic's fallback execution as a successful tactic during profiling"), which was deferred here as potentially needing a broad autotuner change: it turned out to be a local change — `is_in_profile_measurement()` already exists and `choose_one` already disqualifies a raised tactic — so each cuDNN fallback site just re-raises during profiling and keeps warn+fallback for serving. Done in #4132 (GPU-tested on SM100). Thanks for flagging it originally.

PR #3714, opened 2026-06-24
Update CUTLASS DSL CooperativeGroup usage

xiufanl:
## Summary

- Remove the deprecated third `alignment` argument from CUTLASS DSL `pipeline.CooperativeGroup` calls in the persistent GEMM TMA-store producer setup.
- Keep the cooperative group size unchanged and use the current `CooperativeGroup(agent, size)` API.

## Fix

Current CUTLASS DSL wheels no longer accept a separate positional alignment argument for `CooperativeGroup`. These FlashInfer kernels were passing the same thread count as both the group size and the legacy alignment value, which causes a `TypeError` before the kernel can be compiled with newer CUTLASS DSL packages.

This change updates the FlashInfer kernel code to avoid specifying alignment in the first place, matching the current CUTLASS DSL API while preserving the intended producer group size.

## Verification

- Ran an AST audit over all FlashInfer Python sources: 153 `CooperativeGroup` calls across 22 files, with 0 calls using a third positional argument, `alignment=`, or hidden `*args`.
- Checked every `PipelineTmaStore.create(... producer_group=...)` setup and confirmed the producer group resolves to a two-argument `CooperativeGroup` call.
- Confirmed there are no non-Python `CooperativeGroup` references in the repository.
- Ran `python -m py_compile` on the two touched files.
- Ran `git diff --check`.


brandon-yujie-sun:
Thanks @xiufanl the change LGTM

PR #3717, opened 2026-06-24
Fix CuteDSL paged decode non-causal boundary mask

### LINKED ISSUE #3716: [Bug] CuteDSL paged decode with `is_causal=False` uses padding KV lanes
## Environment

- GPU: NVIDIA B200, `sm_100`
- FlashInfer: `0.6.13rc2` and current `main` (`015f5973`)
- PyTorch: `2.9.1+cu128`
- CUDA driver: 580.126.09

## Description

`BatchDecodePagedCuteDSLWrapper` gives wrong output when `is_causal=False` and `seq_len` is not aligned to the 256-token sequence tile.
The non-causal path skips the boundary mask pass. KV lanes after `seq_len` remain in the score tile and enter the softmax denominator. With `q = 0`, attention should be uniform over valid KV tokens only, but the output is scaled as if padding lanes were valid.
The causal path is correct.

## Minimal Repro

```python
import torch
from flashinfer.cute_dsl.attention import BatchDecodePagedCuteDSLWrapper

D, page, H = 128, 16, 8
dev = torch.device("cuda")
dt = torch.bfloat16


def check(seq_len, is_causal):
    npages = (seq_len + page - 1) // page
    indptr = torch.tensor([0, npages], dtype=torch.int32, device=dev)
    indices = torch.arange(npages, dtype=torch.int32, device=dev)
    seq_lens = torch.tensor([seq_len], dtype=torch.int32, device=dev)

    torch.manual_seed(1)
    k_cache = torch.randn(npages, page, H, D, device=dev, dtype=dt) * 0.1
    v_cache = torch.randn(npages, page, H, D, device=dev, dtype=dt) * 0.1
    q = torch.zeros(1, H, D, device=dev, dtype=dt)
    ref = v_cache.reshape(-1, H, D)[:seq_len].float().mean(0)

    wrapper = BatchDecodePagedCuteDSLWrapper(
        torch.empty(512 * 1024 * 1024, dtype=torch.uint8, device=dev)
    )
    wrapper.plan(
        indptr,
        indices,
        seq_lens,
        num_qo_heads=H,
        num_kv_heads=H,
        head_dim=D,
        page_size=page,
        q_data_type=dt,
        kv_data_type=dt,
        sm_scale=1.0 / (D**0.5),
        reduction="none",
        kv_splits=1,
        q_len_per_req=1,
        is_causal=is_causal,
        max_kv_len=seq_len,
    )
    out = wrapper.run(q, k_cache, v_cache).reshape(H, D).float()
    torch.cuda.synchronize()
    ratio = (ref.abs().mean() / (out.abs().mean() + 1e-12)).item()
    max_abs = (out - ref).abs().max().item()
    print(is_causal, seq_len, ratio, max_abs)


for s in (16, 64, 128, 240, 256, 257, 512):
    check(s, is_causal=False)
for s in (16, 64, 128, 240, 256, 257, 512):
    check(s, is_causal=True)
```

## Observed

Before the fix:

```text
is_causal=False seq_len=  16  ref/out=16.000  max_abs=8.61e-02
is_causal=False seq_len=  64  ref/out= 4.000  max_abs=3.89e-02
is_causal=False seq_len= 128  ref/out= 2.000  max_abs=1.63e-02
is_causal=False seq_len= 240  ref/out= 1.067  max_abs=1.33e-03
is_causal=False seq_len= 256  ref/out= 1.000  max_abs=6.02e-05
is_causal=False seq_len= 257  ref/out= 1.940  max_abs=1.09e-02
is_causal=True  all tested seq_len values: ref/out=1.000
```

## Expected

`is_causal=False` should mask KV lanes where `key_pos >= seq_len`. All tested sequence lengths should have `ref/out` close to 1.0.

elwhyjay:
## 📌 Description

Fix CuteDSL paged decode for `is_causal=False` when `seq_len` is not aligned to the 256-token sequence tile.

The old path skipped the boundary mask loop for non-causal decode. Scores for KV lanes after `seq_len` stayed finite, so padding lanes entered the softmax denominator. This made short or unaligned sequences return scaled-down outputs.

This PR:
- masks non-causal boundary tiles with `key_pos >= seq_len`;
- keeps the causal mask condition unchanged;
- adds a paged non-causal decode regression test with `q_len_per_req > 1`.

Validation summary on B200:

- `q = 0` repro: before the fix, `ref/out` followed padding scale (`16.0` at `seq_len=16`, `4.0` at `seq_len=64`); after the fix, all checked lengths are `1.000`.
- Random GQA + multi-query checks match fp32 reference and prefill within `2.7e-4` max absolute error.

## 🔍 Related Issues

Fixes #3716

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * Added non-causal paged attention decode with correct boundary handling when sequence lengths don’t align.
  * Expanded non-causal decoding to support variable query lengths per request.

* **Bug Fixes**
  * Corrected non-causal masking so out-of-range KV lanes are excluded consistently (using proper negative-infinity masking) for split/boundary cases.

* **Tests**
  * Added regression coverage for non-causal paged decode, including variable-length queries and split/boundary scenarios, validated against a reference implementation.


elwhyjay:
Hi @qsang-nv, gentle ping on this PR when you have a chance. It fixes a CuteDSL paged decode correctness issue and adds an attention regression test. Thanks!

qsang-nv:
Two non-blocking follow-ups before merge:

**1. The `tma_mask` docstring is now stale.** It still states that out-of-bounds scores are "masked to 0 instead of -inf. Mathematically equivalent softmax...". That's exactly the behavior this PR fixes — after this change OOB lanes are masked to `-inf` via the boundary-mask loop, and the "mathematically equivalent" claim was the root cause (padding entered the softmax denominator). Please update the docstring so it no longer describes the buggy path.

**2. The new test never exercises a masked boundary tile on a non-zero `kv_split_idx`.** That's the part of the fix most worth pinning down (the `is_last_split` / `is_last_phase` selection), and the current matrix misses it:
- `kv_len=129` is the only length that triggers masking, but `kv_splits` is clamped by `ceil_div(max_kv_len, 256)` (literal 256, not the 128 `sequence_tile` used when `q_len_per_req>1`), so `ceil(129/256)=1` → `kv_splits` is forced to 1 and the boundary tile always lands on split 0.
- `kv_len=1024` does get `kv_splits>1`, but it's tile-aligned and never enters the mask branch.

Since auto `kv_splits` also depends on `sm_count` (device-specific), I'd add a case that pins it explicitly and uses variable per-request lengths, e.g.:

```python
q_len_per_req = 4          # -> 128 sequence_tile, 2 softmax warpgroups
seq_lens = [16, 256, 511]
kv_splits = 2              # passed explicitly to plan(), reduction="kernel"
is_causal = False
```

With a 128 tile: `seq=511` → 4 tiles, last tile on `kv_split_idx=1`, and that split runs 2 iterations so the masked tile also falls on `softmax_phase=1`; `seq=16` is a short partial tile on split 0; `seq=256` is the aligned control. (Heads-up: `seq=16` leaves split 1 with zero iterations, so this path also touches empty-split kernel reduction — if it trips, confirm it's not pre-existing before attributing it to this PR. `_make_paged_kv` will also need a variable-length variant.)

elwhyjay:
Hi @qsang-nv. Thanks for the careful review!

I updated the stale `tma_mask` docstring and added a targeted regression test with variable sequence lengths, explicit `kv_splits=2`, `reduction="kernel"`, and `is_causal=False`. The new case covers the masked boundary tile on a non-zero `kv_split_idx` / later softmax phase as suggested.

Validation:
`python -m pytest tests/attention/test_cute_dsl_decode.py::test_cute_dsl_decode_paged_non_causal_split_boundary_tile -q`

Result: passed.

qsang-nv:
Two more non-blocking follow-ups on the updated version:

**1. Strengthen `test_cute_dsl_decode_paged_non_causal_split_boundary_tile` — swap `511` for `385`.** `[16, 256, 511]` does land the masked tile on `kv_split_idx=1` / `softmax_phase=1` as intended, but at `seqlen=511` (128-token tile) that boundary tile holds **127 valid + 1 OOB** token. If a phase-1 mask regression slipped through, only 1 padding lane would leak into a 511-token softmax — a ~0.1–0.2% perturbation, at/below the `rtol=atol=5e-3` used here, so the test could still pass. `seqlen=16` detects a dropped mask strongly, but only on split 0 / phase 0.

`385` keeps every property while flipping the boundary tile to **1 valid + 127 OOB**:
- `ceil(385/128)=4` tiles, last tile on `kv_split_idx=1`;
- split 1 runs 2 iterations → last tile on `softmax_phase=1`;
- a dropped mask now leaks 127 lanes into a 385-token softmax → error well above tolerance.

So `seq_lens_host = [16, 256, 385]` puts strong detection exactly on the split/phase cell this test is meant to pin.

**2. Two comments are stale now that this block handles the non-causal path too:**
- `gqa_decode_paged.py:1413` — `# Causal masking for spec decode verification`, but the block below now also sets up the non-causal (`else`) boundary mask.
- `gqa_decode_paged.py:1481` — `# Apply causal mask`, but the loop body now applies both causal and non-causal (`key_pos >= seqlen`) masks.

Reword both to cover both paths (e.g. "boundary mask setup" / "apply boundary mask (causal or non-causal)").

elwhyjay:
Hi @qsang-nv, thanks again for catching this. I updated the test case to use `385` and refreshed the stale boundary-mask comments.

richardmcai:
@elwhyjay the tma_mask path has been removed in https://github.com/flashinfer-ai/flashinfer/pull/3794, can you confirm if this covers your use case?

elwhyjay:
@richardmcai Thanks for checking. From my read, #3794’s new DenseMask path should cover the same correctness issue. I’d prefer to keep it moving if it is ready to merge first. Then your PR can rebase on top and preserve the same behavior after removing the tma_mask path.

qsang-nv (approved):

PR #3719, opened 2026-06-24
Change the default of Flashinfer autotune from without CUDA graph, to with CUDA graph.

b8zhong:
## 📌 Description

For example, without this, in small problem sizes, the autotuner will not pick the fastest tactic.

Example A: in mm_fp4 https://github.com/flashinfer-ai/flashinfer/pull/3667, when M is < 32, some incorrect tactics will be chosen.
Example B: in bmm_fp8, for Nemotron 3 NVFP4 shapes, there will be a 10% slowdown when M <= 256, by picking CUTLASS backend instead of cuBLAS.

AFAIK, all ops in the repo should be safe with this change.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * CUDA graph execution is now enabled by default during autotuning, which may improve profiling consistency and performance for supported workflows.

* **Bug Fixes**
  * Adjusted autotuning settings for GEMM so certain precision modes use safer defaults, helping avoid issues during profiling and kernel selection.


b8zhong:
@leejnau @qiching Let me know if there's anything I can do for this PR, i.e I might missed some ops that might not be compatible causing failures. We could just skip them for now.

b8zhong:
For the bmm_fp8 and mm_fp4 path, I can collect some data later.

But here is some comprehensive data for mm_bf16 (which is very important for integration) with using cold L2 and CUDA graph, vs not using any of these, for M from 1 to 64.

image

qiching (commented):
I am supportive of the direction that CUDA graph profiling is strictly more accurate for tactic selection.

before approving the global default flip, I'd like to understand the CI failures.

b8zhong:
The external CI failures look unrelated. (Note: I can't check internal CI)

qiching:
from my view:
`test_mm_mxfp8` fails: This PR flips use_cuda_graph default to True, so mm_mxfp8 is now profiled under CUDA graph. The cute-dsl backend is not graph-capture-safe so may lead to illegal memory access. That error is sticky, so every later test in the worker cascades with the same error.

`test_groupwise_scaled_gemm_fp8` is a separate effect of the same switch: CUDA-graph profiling on 8192³ shapes blows up memory so lead to OOM/timeout.

potential fix: don't revert the global default; set use_cuda_graph=False explicitly on the non-graph-safe configs (_MM_MXFP8_TUNING_CONFIG + the fp8-groupwise config), and open a follow-up to make cute-dsl capture-safe.


@leejnau @b8zhong

b8zhong:
Okay, @qiching can you please point me where test_groupwise_scaled_gemm_fp8 is using autotune? Because I cannot find it

leejnau:
@b8zhong you probably want to refer to these:
https://github.com/flashinfer-ai/flashinfer/issues/3768
https://github.com/flashinfer-ai/flashinfer/pull/3861

vadiklyutiy:
Could you pls clarify what is the status of this PR? What is the blocker?

b8zhong:
Hi, can you join https://sgl-fru7574.slack.com/archives/C09NG5Q0LEP, and we can ask @leejnau about it

PR #3720, opened 2026-06-24
feat(gdn): add output-only BF16-state WY MTP decode kernel

ameynaik-hub:
# feat(gdn): add output-only BF16-state WY MTP decode kernel

## 📌 Description

This PR adds a chunked-prefill-style **WY-representation decode kernel** for GDN (Gated Delta Rule) multi-token processing (MTP) with fixed T=16 tiling. Instead of per-token recurrence, all draft tokens are processed in a single kernel launch using T×T GEMMs and a log-depth Neumann inverse on tensor cores.

**Key characteristics:**
- State precision: BF16
- Frozen state (output-only, no state updates) — targets the MTP speculative-decode **verify** path
- API matches `gdn_decode_bf16_state.gated_delta_rule_mtp` (drop-in)
- Supports the output-only frozen-state subset with K = V = 128
- Registered as `flashinfer.gdn_kernels.gated_delta_rule_mtp_wy_output_only`

### Modified files
- **Added:** `flashinfer/gdn_kernels/gdn_decode_bf16_wy_output_only.py` (kernel + host wrapper)
- **Modified:** `flashinfer/gdn_kernels/__init__.py` (registration)
- **Modified:** `tests/gdn/test_decode_delta_rule.py` (`test_gdn_decode_bf16_wy_output_only_mtp_kernel`)
- **Modified:** `benchmarks/bench_gdn_decode.py` (`--version bf16_wy_output_only` head-to-head vs the bf16-state MTP kernel)

---

Three follow-up changes to the kernel.

### 1. NaN-safe decay matrix (correctness fix, always on)

The decay-matrix element for (row r, col c) was computed as `exp(cumsum_r) · exp(−cumsum_c)`. For strong real decay (large `A_log`), `exp(−cumsum_c)` overflows to `+inf` while `exp(cumsum_r)` underflows to `0`, giving `0 · inf = NaN` and silently breaking MTP verify (gsm8k → 0). Synthetic benches with weak `A_log ~ randn` masked this.

**Fix:** store the log-domain cumsum and form the decay as `exp(cumsum_r − cumsum_c)`, which is bounded ≤ 1 in the causal r ≥ c region. Mathematically identical, NaN-free. Validated: gsm8k 0.000 → 0.585 on real verify inputs.

### 2. Strided q/k/v read — drop `.contiguous()` (flag: `SGLANG_GDN_WY_STRIDED_QKV`, default off)

In the verify path, q/k/v are column slices of the fused conv output (`mixed_qkv[:, slice]`), so their token stride is `conv_dim`, not compact. The host previously `.contiguous()`-materialized all three (~11 µs; v alone ~8 MB). The kernel's existing per-token stride vars now accept the real conv row stride (head/element strides unchanged, so smem/ldmatrix/MMA are untouched); the wrapper passes the row stride and skips the three copies, with a `.contiguous()` fallback when strides don't match. Native path (T ∈ {4, 8}) only.

### 3. Native a/b read — drop a/b staging (flag: `SGLANG_GDN_WY_NATIVE_AB`, default off)

Even on the native path, the wrapper zero-pad-staged a/b into T_KERNEL(=16) buffers (~4 µs, 2 copies). An `ab_native` ctor kwarg makes the a/b batch stride use `n_valid·HV`, and the warp-3 a/b load + gamma compute gate by `n_valid` rows so tail lanes are never loaded. Gamma is a causal prefix-sum and the native path returns a compact `[B, T]` output, so the unloaded tail rows cannot affect valid rows. Native path (T ∈ {4, 8}) only.

### Env flags (default off; set on the serving process)

| Flag | Change | Effect |
|---|---|---|
| `SGLANG_GDN_WY_NATIVE_T` | (pre-existing) | q/k/v native load + in-kernel smem-tail zero (no q/k staging) |
| `SGLANG_GDN_WY_STRIDED_QKV` | 2 | read q/k/v from strided conv-output slices (no `.contiguous()`) |
| `SGLANG_GDN_WY_NATIVE_AB` | 3 | read a/b native `[B, n_valid, HV]` (no staging) |

The NaN fix (change 1) has no flag — it is always on.

---

## 📊 Performance

Kernel-only median time (µs) via CUDA-graph replay, HV = 64, BF16, output-only (`disable_state_update=True`), measured with `benchmarks/bench_gdn_decode.py`:

| Batch | T=4 | T=8 | T=16 |
|-------|-------|-------|-------|
| 1 | 4.29 | 4.09 | 4.50 |
| 2 | 4.29 | 4.09 | 4.30 |
| 4 | 4.91 | 4.70 | 5.32 |
| 8 | 6.14 | 6.34 | 7.16 |
| 16 | 9.62 | 9.82 | 10.85 |
| 32 | 15.35 | 15.97 | 19.14 |
| 64 | 30.30 | 31.12 | 36.04 |
| 128 | 55.29 | 56.31 | 66.15 |
| 256 | 105.77 | 107.48 | 127.07 |

## 🔁 Reproducing the benchmark

Head-to-head vs the bf16-state MTP kernel (includes a built-in WY-vs-state `max|Δ|` cross-check plus timing), sweeping T ∈ {4, 8, 16} × batch ∈ {1…256}:

```bash
python benchmarks/bench_gdn_decode.py \
    --version bf16_wy_output_only \
    --batch-size 1 2 4 8 16 32 64 128 256 \
    --seq-len 4 8 16 \
    --num-q-heads 16 --num-k-heads 16 --num-v-heads 64 --head-size 128 \
    --dtype bfloat16 \
    --warmup 10 --iters 100
```

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * Added a new BF16 output-only decode benchmark mode with per-shape correctness checks and reported speedups, using CUDA-graph timing when possible with automatic fallback.
  * Introduced an optional BF16 output-only decode kernel, exposed when available, including optimized short-sequence handling.

* **Tests**
  * Added a BF16 output-only decode test that runs on supported GPUs and reuses the existing MTP validation harness to confirm correctness.


kahyunnam:
@ameynaik-hub the pre-commit check is failing; please check out the commands for "✅ Pre-commit Checks" in the PR description

flashinfer/gdn_kernels/gdn_decode_bf16_wy_output_only.py
```diff
@@ -0,0 +1,2187 @@
...
+            if _fresh or _RESTAGE:  # always fill a fresh buffer; else honor _RESTAGE
+                qb[:, :T].copy_(q)
+                kb[:, :T].copy_(k)
+                vb[:, :T].copy_(v)
+                ab[:, :T].copy_(a)
+                bb[:, :T].copy_(b)
+            q, k, v, a, b = qb, kb, vb, ab, bb
+
+    pool_size = h0.shape[0]
+    _num_sms = 148  # B200
```
kahyunnam: See Gemini comment for above -- hardcoding this may pick wrong occupancy on H100.

flashinfer/gdn_kernels/gdn_decode_bf16_wy_output_only.py
```diff
@@ -0,0 +1,2187 @@
...
+    # One CTA per (b, hv) — full V tile per CTA. Per-CTA SMEM ~51.5 KB -> <=4 CTAs/SM.
+    _total_ctas = HV * B
+    _needed = math.ceil(_total_ctas / _num_sms)
+    mbp = max(1, min(_needed + 1, 4))
+    # T-aware Phase-2 squaring depth.
+    t_disc = 4 if T <= 4 else (8 if T <= 8 else 16)
+    # n_valid in the key: native (n_valid<T) vs staged (n_valid=T_KERNEL) compile to
+    # different kernels (different q/k batch stride + the smem-tail-zero path).
+    cache_key = (str(device), B, pool_size, mbp, t_disc, n_valid, _qkv_rs, _ab_native_flag)
```
kahyunnam: This cache key has both B and pool_size, wouldn't this need to re-compile per batch/pool? This seeme likely to not be not viable for production

flashinfer/gdn_kernels/gdn_decode_bf16_wy_output_only.py
```diff
@@ -0,0 +1,2187 @@
...
+        # staged buffer. tidx in [0,THREADS=128) covers K_DIM=128 cols exactly;
+        # pad cols [K_DIM:K_PADDED) are never read. The following sync_threads()
+        # publishes both the loaded rows and these zeros before the L2-norm reads.
+        if const_expr(self._n_valid < T):
+            for _zr in cutlass.range_constexpr(self._n_valid, T):
+                sK.iterator[_zr * K_PADDED + tidx] = io(0.0)
+                sQ.iterator[_zr * K_PADDED + tidx] = io(0.0)
+        sync_threads()
+
+        # L2 norm for K (warps 0,1) and Q (warps 2,3) — t-aware warp skip.
```
kahyunnam, line 1076: does this need to be guarded by `if const_expr(use_qk_l2norm)`?

kahyunnam (commented):
Some comments below; also, please remove "GDN_WY_VERIFY_CHANGES.md" file from PR, it seems like an internal note

flashinfer/gdn_kernels/gdn_decode_bf16_wy_output_only.py
ameynaik-hub: resolved

flashinfer/gdn_kernels/gdn_decode_bf16_wy_output_only.py
ameynaik-hub: fixed.

kahyunnam (approved):

kahyunnam:
will merge and cherry-pick pending above CI bot.

PR #3740, opened 2026-06-26
Add NVFP4 slot-mapping paged KV append

lesj0610:
## 📌 Description

Add FlashInfer-native NVFP4 paged KV append APIs for runtimes that write paged KV cache entries directly.

This PR:
- Adds the base NVFP4 paged KV append API.
- Adds a slot-mapping variant that interprets slot mappings as flat paged-cache locations and skips negative padding slots.
- Exposes FFI and Python bindings that accept device scalar scale tensors.
- Adds coverage for layouts, append tensor contiguity, slot index dtypes, boundary page sizes, and padding slots.

## 🔍 Related Issues

N/A

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

Validation performed:
- `python -m pytest tests/attention/test_page.py::test_nvfp4_quantize_append_paged_kv_cache_with_slot_mapping -q -s -x --tb=short`
- `python -m pytest tests/attention/test_page.py::test_nvfp4_quantize_append_paged_kv_cache -q -s -x --tb=short`
- Direct packed-cache view smoke check on SM86 for NHD and HND layouts
- `git diff --check`
- `python -m py_compile flashinfer/page.py tests/attention/test_page.py flashinfer/__init__.py`

Not yet verified:
- Full CUDA/JIT regression matrix across all dtype, page-size, layout, and slot-mapping combinations.

## Reviewer Notes

## Summary by CodeRabbit

* **New Features**
  * Added support for FP4-quantized paged KV-cache append operations, including a variant that works with slot mappings.
  * Exposed the new functionality in the public API and low-level interface.

* **Bug Fixes**
  * Added stronger input validation for tensor types, shapes, layouts, and scale values.
  * Improved handling of empty inputs, invalid slots, and no-op cases so caches remain unchanged when expected.

* **Tests**
  * Expanded GPU test coverage across layouts, dtypes, head sizes, slot-mapping cases, and scale validation.


qsang-nv:
Went through the new append paths. +1 on the two scale/device issues already raised, with some additions, plus one more item.

**1. Non-positive device scalars silently zero the KV cache (blocker)**

`_as_float32_scalar_tensor` only validates the value for CPU tensors (`value.device.type == "cpu" and value.item() <= 0.0`); a CUDA scalar with `0` / negative / `NaN` skips the check, and the C++ side only validates dtype/numel. In the kernel, `global_scale <= 0` (and `NaN`, since `NaN > 0.0f` is false) makes both `sf_value` and `output_scale` zero, so the packed KV **and** block scales are written as all-zeros — silent data corruption, not an error. This affects **both** the base and slot-mapping APIs (the device-scalar path is the one this PR is built around), and the tests only use a positive float / positive device scalar, so the bad device-scalar path is uncovered. Please make the kernel reject non-positive/NaN scales explicitly (device-side error) or validate on the host, and add a device-scalar regression test.

**2. Base API is missing `CHECK_DEVICE` on `batch_indices` / `positions`**

The slot-mapping API correctly does `CHECK_DEVICE(slot_mapping, append_key)`, but the base API checks `append_value`/caches/`kv_*` against `append_key` and omits `batch_indices` and `positions` — they only get `CHECK_INPUT` (validates "is CUDA + contiguous", not the device id). On multi-GPU this lets a wrong-device index tensor reach the kernel as an illegal access. Two lines:

```cpp
CHECK_DEVICE(batch_indices, append_key);
CHECK_DEVICE(positions, append_key);
```

**3. New public APIs are missing `trace=` wiring**

Both new entrypoints (`nvfp4_quantize_append_paged_kv_cache` at `page.py:529` and `..._with_slot_mapping` at `:622`) are bare `@flashinfer_api`. The sibling `append_paged_kv_cache` in the same module already wires `trace=append_paged_kv_cache_trace`, so there's no category exemption — per CLAUDE.md's trace-template checklist every public `@flashinfer_api` needs one (without it `fi_trace()` raises "No fi_trace spec is registered" and auto-dump emits no JSON). The existing template can't be reused as-is: it doesn't model the packed NVFP4 cache, the FP8 scale cache, `k_scale`/`v_scale`, the slot mapping, or the in-place writes. Proper fix: add two TraceTemplates in `flashinfer/trace/templates/page.py`, wire `trace=` on both decorators, add example calls in `tests/trace/example.py`, and commit the regenerated `fi_trace_out/*.json`.

lesj0610:
@qsang-nv Pushed an update for the three points.

For scale validation: the Python side now rejects non-finite or non-positive scales for both APIs, including CUDA scalar tensors (the device type guard is removed so `.item()` always runs). Added `std::isfinite` on the C++ tensor-scale path too as defense-in-depth.

`CHECK_DEVICE(batch_indices, append_key)` and `CHECK_DEVICE(positions, append_key)` added to the base append API, matching what the slot-mapping API already does.

Added trace templates for both NVFP4 append APIs, wired `trace=` on the decorators, and added example calls. The templates model the packed uint8 K/V cache and FP8 scale cache as in-place outputs. No numeric reference — the missing piece was the trace spec itself.

Local checks:

- pre-commit: passed
- py_compile: passed
- bad-scale tests (CUDA scalar 0 / negative / NaN, Python float, CPU tensor): 7 passed
- trace/template tests: 728 passed, 148 skipped

The focused NVFP4 CUDA kernel test still hits a CUDA 13 fastdiv compile error before the test body runs, so that part is not verified locally.

lesj0610:
I’m currently fixing the environment issue, and I’ll add the verification results once it’s completed.

lesj0610:
Follow-up on the local CUDA test.

The previous fastdiv compile failure was from my local editable setup. The worktree was missing the vendored CCCL submodule / editable data links, so JIT fell back to the CUDA toolkit headers and failed before reaching the test body. After fixing that setup, the CUDA tests run.

Additional local checks on RTX 3090 (SM86):

- bad-scale tests: 7 passed
- NVFP4 append page tests: 66 passed
- trace/template tests: 728 passed, 148 skipped

So the focused NVFP4 CUDA append path is verified locally now.

lesj0610:
I found one more issue in the scale validation. The `.item()` host read is not capture-safe. vLLM calls the slot-mapping append during CUDA graph capture, so the host sync breaks the capture.

I fixed this so eager calls still reject bad scales, but the captured path does not do the host read or stream synchronization.

Verified with the focused NVFP4 slot-mapping tests, including CUDA graph capture/replay: 5 passed. I also checked the vLLM path with Gemma4 GGUF TP=1 + nvfp4 KV cache; CUDA graph capture completed, the API server reached startup, and a short chat completion returned 200.

qsang-nv:
Re-checked the latest revision — three items still open, one confirmed by actually running the committed artifact.

**1. CUDA-graph path still accepts invalid scales (blocker)**

Both Python (`_as_float32_scalar_tensor`) and C++ (`CheckPositiveFiniteFloatScalar`) skip value validation while capturing. Since the graph records the scale tensor's *address*, passing an invalid value during capture — or mutating the tensor to 0/negative/NaN after capture — makes replay silently write all-zero KV (the original silent-corruption mode). The graph test only uses `1.0`, so this isn't covered. Either add graph-compatible validation, or explicitly document that in capture mode the scale must be pre-validated and immutable, and add a graph test with a bad scale to lock the behavior.

**2. Eager path does up to 4 host syncs per append (major)**

For a CUDA scalar both layers validate: Python `k_scale.item()` + `v_scale.item()`, then C++ `CheckPositiveFiniteFloatScalar(k_scale)` + `(v_scale)`, each a D2H copy + `cudaStreamSynchronize`. That's 4 host syncs on the decode hot path, and the Python/C++ checks are redundant. Validate in a single layer, ideally batching both scales into one D2H + sync.

**3. Generated trace JSON is not self-contained (blocker)**

`_render_init_source` inlines only `_init_helpers.py`, the functions in `_trace_init_dependencies`, and the init itself. The new init calls `_append_paged_kv_cache_init` and `_nvfp4_paged_cache_shapes`, both defined in `templates/page.py` (not `_init_helpers.py`) and not declared as init dependencies, so the dumped JSON has the calls but not the definitions. Executing the committed JSON's `init` fails:

```text
def _append_paged_kv_cache_init present: False
def _nvfp4_paged_cache_shapes present: False
RESULT: NameError -> name '_append_paged_kv_cache_init' is not defined
```

Fix: declare `_trace_init_dependencies = (_append_paged_kv_cache_init, _nvfp4_paged_cache_shapes)` on the init (or move the helpers into `_init_helpers.py`), regenerate the JSON, and add a test that executes the dumped `init` so this is caught.

lesj0610:
@qsang-nv Updated.

For the graph path, I only changed the slot-mapping kernel. That one takes k_scale/v_scale as device pointers, so graph replay reads whatever value sits in the tensor at replay time. I added a device-side positive/finite guard — chose trap over documenting because silent corruption is harder to catch in production than a crash. The base append API passes scales by value so capture copies them, host-side validation is enough.

For the eager path, scale normalization and validation are now in one helper that handles all input types. Python floats and CPU tensors are checked immediately, CUDA tensors are batched into one D2H transfer. The duplicate C++ host syncs are removed.

For trace, I declared _trace_init_dependencies on the two NVFP4 inits and regenerated both JSONs. Added a test that exec()s the dumped init and calls the function, so a missing dependency will break it.

Local:
- pre-commit: passed
- py_compile: passed
- tests/attention/test_page.py: 70 passed
- tests/trace/test_fi_trace.py: 27 passed

lesj0610:
@qsang-nv Would you mind checking #3748 too?

qsang-nv:
**[Blocker] Capture-mode scale validation still materializes temporary tensors for non-CUDA-tensor scales**

In `_as_float32_scalar_tensors`, capture detection is gated on whether an input is *already* a CUDA tensor:

```python
has_cuda_tensor = any(isinstance(v, torch.Tensor) and v.is_cuda for _, v in values)
stream_capturing = has_cuda_tensor and torch.cuda.is_current_stream_capturing()
```

So when the scales are Python floats or CPU tensors, `stream_capturing` is `False` even while capturing, and the function falls through to `torch.tensor(..., device=...)` / `.to(device=...)`, creating a temporary CUDA tensor inside the capture. The graph records that tensor's device pointer, but its lifetime ends when the API returns, so replay can read reused/freed memory. The device-side `trap` guard doesn't save this — the reused memory may hold a valid-looking positive value, so the kernel proceeds with the wrong scale (silent wrong output). CPU scales additionally hit `.item()` + an H2D copy during capture.

Reproduced by forcing `is_current_stream_capturing()` to return True and calling the extracted helper:

```text
[two python floats]          capturing=True -> NO raise; tensors on ['cuda','cuda']  (temp created in capture)
[two CPU tensors]            capturing=True -> NO raise; tensors on ['cuda','cuda']  (temp created in capture)
[two CUDA tensors on cuda:0] capturing=True -> NO raise; returns the original tensors (correct)
```

Only the already-on-device CUDA-tensor case is handled correctly. This affects the slot-mapping API (`..._with_slot_mapping`, which routes scales through this helper); the base API takes a host `double` and is unaffected. It looks like a regression from the "Simplify NVFP4 scale tensor validation" change — the earlier per-value `value.is_cuda and capturing` check didn't have this gap.

Fix: gate capture detection on `is_current_stream_capturing()` alone (not `has_cuda_tensor`), and while capturing, reject any scale that would need an on-the-fly `torch.tensor(...)` / `.to(device)` (float, CPU, or wrong-device) rather than materializing a temporary — require a contiguous CUDA tensor already on the target device.

lesj0610:
Ah, this is a gap I missed while optimizing. I’ll fix it right away.

lesj0610:
Fixed. The capture check no longer depends on whether the input is already a CUDA tensor. Float and CPU tensor scales are rejected before any temporary CUDA allocation during capture. Added a regression test for both cases.

Local: tests/attention/test_page.py slot-mapping focused — 50 passed.

qsang-nv:
**[Minor] Capture detection isn't scoped to `append_key.device` (multi-GPU gap)**

Nice fix on the float/CPU temporary-scale rejection. One remaining edge: the capture check uses `torch.cuda.is_current_stream_capturing()`, which takes no device argument and queries the *current device's* current stream (docstring: "capture is underway on the current CUDA stream ... if a CUDA context does not exist on the current device, returns False"). If a capture runs on `append_key.device` (say `cuda:1`) while the process's current device is still `cuda:0`, it returns False and the function silently falls back to the eager path: float/CPU scales get materialized via `torch.tensor(device=...)` / `.to(device)` (a temporary whose pointer the graph records → use-after-free on replay), and a CUDA-tensor scale hits the `torch.stack(...).cpu()` validation (a host sync inside capture).

Scope the capture check to `append_key.device` — e.g. wrap it in `with torch.cuda.device(device): torch.cuda.is_current_stream_capturing()`, or query that device's stream directly (`torch.cuda.current_stream(device)` accepts a device). Low priority — it only triggers on the uncommon multi-GPU setup where the capture device isn't the current device — but the capture check is correctness-critical, so worth closing.

lesj0610:
@qsang-nv Fixed — the capture check now runs under `with torch.cuda.device(append_key.device)`.

Added a two-GPU regression test for this case. Local slot-mapping tests passed, and full `test_page.py` passed on a free GPU (72 passed, 1 skipped). pre-commit passed.

qsang-nv:
Root-caused the CI failures on `test_page.py` — it's a device-placement issue in the test's dequant reference helper, not in the append kernel. Deterministic per GPU arch, which is why your SM86 run passed while CI (SM90+) fails all 50 dequant-checking cases.

`_nvfp4_dequant_linear` assumes `e2m1_and_ufp8sf_scale_to_float` preserves the input device, but that helper dispatches by compute capability:
- **SM < 90** (your RTX 3090): pure-PyTorch fallback — device-preserving, CUDA in → CUDA out;
- **SM ≥ 90** (CI): the `..._sm100` op is host-side — it copies inputs to CPU and **always returns a CPU tensor** (`out = torch.zeros(..., device="cpu")`; see the `NOTE(Zihao): this is another cpu op` comment).

So on SM90+ `k_dequant` lands on CPU, and both `k_dequant[valid_slots]` (CUDA indices) and `_assert_nvfp4_quantized_close(cpu, cuda)` raise the exact CI error.

One-line fix in the helper (no-op on pre-SM90):

```python
def _nvfp4_dequant_linear(packed, scales, global_scale):
    ...
    return dequant.to(packed.device).reshape(*packed.shape[:-1], packed.shape[-1] * 2)
```

Verified on an H100 (SM90) at `aa0efe8a`: without the fix, the failing cases reproduce CI's `indices should be either on cpu or on the same device as the indexed tensor (cpu)` exactly; with it, the full nvfp4 subset of `test_page.py` passes — **70 passed, 1 skipped** (the two-GPU capture case on a single-GPU node).

(Unifying the device behavior of `e2m1_and_ufp8sf_scale_to_float` itself across arches would be the deeper fix, but that's out of scope here — the existing `NOTE(Zihao)` comment already flags it.)

lesj0610:
@qsang-nv Thanks for running this down, and for verifying on the H100. Applied in b3bf0048. I put the `.to(packed.device)` after the reshape instead of before, just to match how the other fp4 test helpers do it (utils_fp4.py, test_xqa_batch_decode.py), otherwise it's exactly your one-liner.

I can't exercise the sm100 path on my 3090 so the SM90+ side is on CI, but the nvfp4 subset of test_page.py stays green locally (70 passed, 1 skipped).

qsang-nv (approved):

PR #3741, opened 2026-06-26
Add caller-owned workspace sizing helper

lesj0610:
## 📌 Description

Add a caller-owned workspace sizing helper so integrations can allocate the FlashInfer float workspace to the actual planned requirement instead of always reserving the default workspace size.

This PR:
- Adds count-only support to `AlignedAllocator` so plan allocation offsets can be reused for sizing without touching workspace memory.
- Adds batch decode and batch prefill `workspace_size` FFI functions that return required float and int workspace bytes.
- Exposes `workspace_size` on generated and JIT batch decode/prefill Python modules, with `None` for TRTLLM-gen where the helper is not available.
- Adds CUDA tests that verify decode and FA2 prefill can plan with buffers sized exactly from `workspace_size`.
- Leaves `plan_info` layout and runtime run paths unchanged.

## 🔍 Related Issues

N/A

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

Validation performed:
- `FLASHINFER_DISABLE_VERSION_CHECK=1 MAX_JOBS=4 FLASHINFER_NVCC_THREADS=2 python -m pytest tests/attention/test_workspace_size.py -q -s -x --tb=short`

Not yet verified:
- Full CUDA/JIT regression matrix across all dtype, mask, sliding-window, positional-encoding, and backend combinations.

## Reviewer Notes

## Summary by CodeRabbit

* **New Features**
  * Added workspace-size reporting for paged-KV batched decode and KV-cache prefill.
  * Planning modules/wrappers now expose `workspace_size` to let callers size buffers before running.
* **Bug Fixes**
  * Improved workspace sizing/allocation behavior for more accurate buffer requirements.
  * Clarified that required float workspace buffers must be 16-byte aligned.
* **Tests**
  * Added CUDA-only tests validating decode and prefill workspace sizing and planning with exact-sized buffers.

### DIFF
--- a/csrc/batch_decode.cu
+++ b/csrc/batch_decode.cu
@@ -80,6 +80,52 @@ Array<int64_t> BatchDecodeWithPagedKVCachePlan(
   return Array(plan_info.ToVector());
 }
 
+Array<int64_t> BatchDecodeWithPagedKVCacheWorkspaceSize(
+    TensorView device_buffer, TensorView indptr, int64_t batch_size, int64_t num_qo_heads,
+    int64_t num_kv_heads, int64_t page_size, bool enable_cuda_graph, int64_t window_left,
+    double logits_soft_cap, int64_t head_dim_qk, int64_t head_dim_vo, TensorView empty_q_data,
+    TensorView empty_kv_data) {
+  (void)window_left;
+  (void)logits_soft_cap;
+  (void)empty_q_data;
+  (void)empty_kv_data;
+  CHECK_INPUT_TYPE(indptr, dl_int32);
+
+  TVM_FFI_ICHECK_EQ(head_dim_qk, head_dim_vo)
+      << "CUDA cores template only supports equal head dim for QK and VO, please use tensor "
+         "cores template for different head dim";
+
+  ffi::CUDADeviceGuard device_guard(device_buffer.device().device_id);
+  const cudaStream_t stream = get_stream(device_buffer.device());
+  size_t float_workspace_size_in_bytes = 0;
+  size_t int_workspace_size_in_bytes = 0;
+
+  DISPATCH_context(
+      DTypeQ, DTypeKV, DTypeO, IdType, HEAD_DIM_QK, HEAD_DIM_VO, POS_ENCODING_MODE,
+      USE_SLIDING_WINDOW, USE_LOGITS_SOFT_CAP, AttentionVariant, Params, [&] {
+        DISPATCH_GQA_GROUP_SIZE(num_qo_heads / num_kv_heads, GROUP_SIZE, {
+          auto work_estimation_func = BatchDecodeWithPagedKVCacheWorkEstimationDispatched<
+              GROUP_SIZE, HEAD_DIM_QK, POS_ENCODING_MODE, AttentionVariant, Params>;
+          cudaError_t status =
+              DecodePlanWorkspaceSize<HEAD_DIM_QK, POS_ENCODING_MODE, AttentionVariant, Params>(
+                  float_workspace_size_in_bytes, int_workspace_size_in_bytes,
+                  static_cast<IdType*>(indptr.data_ptr()), batch_size, num_qo_heads, page_size,
+                  enable_cuda_graph, /*stream=*/stream, work_estimation_func);
+
+          TVM_FFI_ICHECK(status == cudaSuccess)
+              << "BatchDecodeWithPagedKVCache workspace size failed with error "
+              << cudaGetErrorString(status);
+          return true;
+        });
+      });
+
+  std::vector<int64_t> workspace_sizes = {
+      static_cast<int64_t>(float_workspace_size_in_bytes),
+      static_cast<int64_t>(int_workspace_size_in_bytes),
+  };
+  return Array(workspace_sizes);
+}
+
 void BatchDecodeWithPagedKVCacheRun(TensorView float_workspace_buffer,
                                     TensorView int_workspace_buffer, Array<int64_t> plan_info_vec,
                                     TensorView q, TensorView paged_k_cache,
--- a/csrc/batch_decode_jit_binding.cu
+++ b/csrc/batch_decode_jit_binding.cu
@@ -27,6 +27,12 @@ Array<int64_t> BatchDecodeWithPagedKVCachePlan(
     int64_t window_left, double logits_soft_cap, int64_t head_dim_qk, int64_t head_dim_vo,
     TensorView empty_q_data, TensorView empty_kv_data);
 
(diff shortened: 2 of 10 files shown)


qsang-nv:
Thanks for putting this together — a caller-owned workspace sizing helper is a genuinely useful addition for memory-tight deployments, and the design is clean and low-risk (no changes to `run` paths or `plan_info` layout). I verified the size accounting line-by-line against `DecodePlan`/`PrefillPlan` and it matches today (allocation order, padding, and the `padded_batch_size`/grid estimation are all consistent). A few things before merge:

**1. Public entry point + docs for the new API**

Since this is explicitly a *caller-owned* helper, its only consumers are external integrations — but there is currently no ergonomic, documented way to reach it. The only exercised path is the private `wrapper._cached_module.workspace_size(...)` (as the test does), and there is no docstring describing the calling convention: e.g. that the `indptr`/`qo_indptr` arguments must be **host** tensors (mirroring `plan`), that the first argument is a device tensor used only to select the device/stream, and that the return value is `(float_bytes, int_bytes)`. Could you add a stable public entry point (a wrapper method or a public function) plus a short docstring/example for the "size → allocate → plan" flow? `_cached_module` shouldn't really be the public contract.

**2. Test coverage gap on the split-KV prefill path**

`test_batch_prefill_workspace_size_plans_with_exact_buffers` uses `disable_split_kv=True`, so `split_kv` is False and `float_workspace_size` is 0 (the test only asserts `>= 0`). This means the most error-prone part of the prefill accounting — `tmp_v`/`tmp_s` float workspace, `merge_indptr`, `block_valid_mask` — is never exercised. The decode test does hit split-KV (kv_len=4096), but prefill doesn't. Could you add a prefill case with split-KV enabled (drop `disable_split_kv`, use a longer kv_len) that plans with the exact returned sizes? The `enable_cuda_graph=True` path is also untested in both tests.

**3. Eliminate the duplicated plan logic**

`DecodePlanWorkspaceSize`/`PrefillPlanWorkspaceSize` are hand-copied mirrors of `DecodePlan`/`PrefillPlan`, kept in step only by the four `IMPORTANT: keep ... in sync` comments. If someone later changes an allocation in `*Plan` and misses the size variant, the failure is silent: under-counting → caller allocates too little → `plan` overflows; over-counting → wasted memory (defeating the PR's purpose). Templating the plan routine on the allocator type (real allocator for `*Plan`, counting allocator for sizing) would let both share a single code path and remove this class of bug entirely.

lesj0610:
@qsang-nv Thank you for the thorough review. All three points make sense and I agree with the direction.

For the public API, I added workspace_size(...) as a proper method on both decode and prefill wrappers, so callers do not need to touch _cached_module anymore. The docstring explains the host indptr convention, device/stream selection from the wrapper workspace buffer, the (float_bytes, int_bytes) return value, and the intended "size then allocate then plan" workflow.

For the split-KV prefill coverage, I found that combining use_cuda_graph=True with fixed_split_size can hit the padded_batch_size limit depending on GPU SM count, so I split the test into two separate cases. One exercises split-KV with use_cuda_graph=False, and the other exercises CUDA graph sizing without fixed split. This keeps both paths covered without relying on that scheduler edge case.

For the duplicated plan logic, I refactored the allocation sequence so real plan and workspace sizing share the same code path. The "keep in sync" comments are gone now because the two paths no longer duplicate the allocation sequence.

Pushed the update, please check when you get a chance.

qsang-nv:
The templated `*PlanImpl<MATERIALIZE>` refactor and the new public `workspace_size` method both look good. Two items still need addressing before the exact-size path is safe:

**1. The CUDA-graph and fixed-split test cases aren't actually separated**

In `_run_batch_prefill_workspace_size_plan`, the `fixed_split_size` argument is overridden unconditionally:

```python
def _run_batch_prefill_workspace_size_plan(use_cuda_graph=False, fixed_split_size=None):
    ...
    fixed_split_size = 16          # overrides the argument
    ...
    if fixed_split_size is not None:   # therefore always true
        plan_kwargs = {"fixed_split_size": fixed_split_size, "disable_split_kv": False}
```

So both tests run with `fixed_split_size=16`, and `test_..._cuda_graph` is still `use_cuda_graph=True` **with** `fixed_split_size=16` — exactly the GPU-dependent combination the split was meant to avoid (the CUDA-graph padding limit depends on `num_sm`). Dropping that override line lets the argument take effect. Note that once the CUDA-graph case no longer forces a fixed split, `split_kv` is decided by `PrefillBinarySearchKVChunkSize` and isn't guaranteed, so `assert float_workspace_size > 0` may need to relax to `>= 0` (or the case should explicitly force a split).

**2. The exact-size contract still depends on an unchecked base alignment**

The counting allocator assumes offset 0 already satisfies every alignment (the first 16-byte alloc gets `padding = 0`), whereas the real allocator calls `std::align` from the actual base pointer. `torch.empty` buffers are fine, but `reset_workspace_buffer` accepts arbitrary tensors — including a non-16B-aligned slice. In that case `workspace_size()` under-counts the leading padding and the subsequent `plan` overflows (surfaced as a generic buffer-overflow rather than an alignment error). The docstring now documents the 16-byte requirement, but it isn't enforced. A cheap base-pointer alignment check in `workspace_size()` / `plan()` (failing with a clear message) would make the contract robust.

lesj0610:
@qang-nv Pushed a follow-up for this.

The fixed-split and CUDA graph cases are now actually separate. The helper no longer forces fixed_split_size=16, and the CUDA graph case allows zero float workspace when the scheduler does not need split-kv workspace.

For the alignment issue, the paged decode/prefill wrappers now reject non-16-byte-aligned workspace buffers at construction/reset time and before workspace_size()/plan(). Empty buffers are skipped. I also added CUDA regression coverage using a one-byte-sliced workspace buffer for both decode and prefill.

Local checks passed for py_compile, ruff, pre-commit, and the new alignment tests. The full workspace_size test file still cannot complete locally because the exact-size plan cases hit a CUDA 13 fastdiv JIT compile error before the test body runs.

qsang-nv:
Ran `test_workspace_size.py` on **H100 (SM90) with CUDA 13.2 (V13.2.51)** at PR head `93e355ef` — **6 passed** (61s):

- exact-size plan: decode (cuda_graph=False/True), prefill (fixed-split, cuda_graph)
- unaligned-buffer rejection: decode + prefill

No fastdiv JIT compile error on this CUDA 13.2 setup, so the compile failure you hit appears environment-specific (a particular CUDA 13 minor / build config) rather than something in this PR or CUDA 13 in general. The exact-size plan path and the alignment guard are functionally verified here.

qsang-nv (approved):

lesj0610:
@qsang-nv Thanks, I appreciate it.

PR #3739, opened 2026-06-26
Enable Ampere FA2 large-head attention

lesj0610:
## 📌 Description

Enable FA2 large-head attention on SM80+ instead of treating the whole large-head path as SM100+ only.

This PR:
- Enables FA2 large-head 16-bit KV modules on SM80+.
- Enables NVFP4 large-head FA2 prefill on SM80+.
- Keeps NVFP4 large-head decode on SM100+ only.
- Adds VO-split handling and shared-memory budgeting updates for large-head prefill paths.
- Adds focused paged/ragged/single prefill coverage and updates large-head skip logic in related tests.
- Collapses the duplicate single/batch prefill NVCC flag helper into one shared helper.

The NVFP4 part is motivated by long-context serving on pre-SM100 GPUs. On Ampere cards like RTX 3090, large-head models hit the KV-cache capacity wall pretty fast, and NVFP4 is a practical way to push that limit back while staying on the FlashInfer FA2 prefill path.

That said, the support is intentionally partial on pre-SM100. This PR only enables NVFP4 large-head FA2 prefill on SM80+. Decode large-head NVFP4 kernels stay SM100+-only; I have not validated them on Ampere and do not want to claim otherwise. The scope is just unblocking the prefill/read path for longer-context workloads, not full one-byte large-head decode on Ampere.

Scope:
- NVFP4 large-head support on pre-SM100 is prefill-only.
- NVFP4 large-head decode remains SM100+ only.
- Does not change FA3 routing or holistic batch-attention generation.
- Does not include model-specific routing changes.
- Default AOT jit-cache exclusion for head_dim=512 is already handled by #3769; this PR leaves that list unchanged.

## 🔍 Related Issues

N/A

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [x] Tests have been added or updated as needed.
- [ ] All tests are passing (`unittest`, etc.).

Validation performed:
- `pre-commit run --all-files`
- `git diff --check`
[...]

PR #3748, opened 2026-06-28
Add paged NVFP4 KV dequantization helper

lesj0610:
## 📌 Description

Add a paged NVFP4 KV dequantization helper for cases where the caller already has paged NVFP4 K/V cache and needs contiguous K/V tensors again.

This PR:
- Adds `nvfp4_kv_dequantize_paged` for paged NVFP4 K/V cache input.
- Gathers physical pages by caller-provided block tables and sequence lengths.
- Writes dequantized K and V into caller-owned contiguous output tensors.
- Supports tuple or stacked paged cache input, NHD and HND layouts, int32/int64 block tables, fp16/bf16 output, non-contiguous outer strides, and different K/V head dimensions.
- Rejects block tables that cannot cover the requested output length before launching the CUDA kernel.
- Keeps the existing flat `nvfp4_kv_dequantize` API and existing paged attention paths unchanged.

The main point is simple: when K/V is stored in paged NVFP4 cache, caller should not need to manually gather pages and unpack E2M1 values on the Python side. This helper keeps that logic in FlashInfer.

Scope note: the explicit FP4 scale-stride handling in this PR is applied to the paged prefill path. Ragged params carry the same fields for parameter/JIT layout consistency, but ragged prefill still uses the existing compact scale layout because there is no current caller passing strided NVFP4 scale views through the ragged path. Ragged strided-scale support can be handled separately if a caller needs it.

## 🔍 Related Issues

N/A

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

Validation performed:
- `pre-commit run --files csrc/fp4_kv_dequantization.cu tests/utils/test_fp4_kv_quantization.py`
- `git diff --check`
- `python -m py_compile flashinfer/quantization/fp4_quantization.py tests/utils/test_fp4_kv_quantization.py`
- `FLASHINFER_DISABLE_VERSION_CHECK=1 FLASHINFER_WORKSPACE_BASE=/tmp/flashinfer-jit-test-paged-dequant-resume PYTHONPATH=. python -m pytest -q tests/utils/test_fp4_kv_quantization.py::test_nvfp4_kv_dequantize_paged tests/utils/test_fp4_kv_quantization.py::test_nvfp4_kv_dequantize_paged_rejects_short_block_tables`

Not yet verified:
- Full CUDA/JIT regression matrix across all page sizes, layouts, output dtypes, and larger batch shapes.
- Performance comparison with a consumer-side gather/dequant implementation.

## Reviewer Notes

## Summary by CodeRabbit

* **New Features**
  * Added paged NVFP4 KV-cache dequantization that writes dequantized K and V outputs.
  * Supports both NHD and HND cache layouts with per-page FP8 E4M3 scaling, including functional trace support.
  * Extended paged KV prefill to handle strided FP4 scale-factor tensors.
* **Bug Fixes**
  * Improved validation for kv layout, block-table coverage, scalar scale tensors, and required head/dimension alignment.
* **Tests**
  * Expanded coverage for layout/dtype combinations, non-contiguous and stacked inputs, long contexts, and rejection cases; added coverage for strided scale views.
* **Chores**
  * Exposed the new paged dequantization API at the package top level and related modules.


qsang-nv:
This PR bundles two separable changes — the new `nvfp4_kv_dequantize_paged` helper and a change to the existing paged-prefill NVFP4 scale path (`page_produce_kv_sf` now takes real scale strides instead of deriving them from KV strides). Two blockers plus smaller items:

**1. [Blocker] persistent BatchAttention reads scales at 8× the correct offset**

`page_produce_kv_sf` dropped its internal `/ SF_CONTAINERS` (÷8) conversion and now expects real scale strides from the caller. The paged-prefill call sites were updated to pass `k_sf_stride_*`, but `persistent.cuh` (329/337/378/396) still passes `k_stride_*` / `v_stride_*` (KV strides) — 8× the correct scale stride for a compact layout, so persistent BatchAttention reads wrong/out-of-bounds scales. Those 4 are the only callers left unconverted. `tests/attention/test_batch_attention.py::test_batch_attention_nvfp4` covers this path and wasn't run.

**2. [Blocker] trace template isn't self-contained for HND, and the trace checklist is incomplete**

The NHD/HND templates share `_nvfp4_kv_dequantize_paged_init` (default `kv_layout="NHD"`), but `axes` has no `kv_layout` entry, so nothing feeds it — the HND JSON's `init` builds NHD-shaped inputs (the existing test hides this by passing `fi_init(..., kv_layout="HND")`). Also, no `tests/trace/example.py` call and no `fi_trace_out/*.json` are committed, so the trace checklist isn't finished. Since `_render_init_source` uses `inspect.getsource(init)` (a `functools.partial` isn't serializable, and `kv_layout` isn't an integer axis), the clean fix is a dedicated HND init wrapper that delegates to the base init, with the base init pulled in via `_trace_init_dependencies` (same pattern as the NVFP4-append trace). Then add NHD/HND examples, regenerate the JSON, and add a test that executes each dumped `init`.

**3. [Minor] Empty output launches a zero-sized grid**

`dim3 grid(batch_size * max_seq_len * num_heads)` has no non-zero guard; an empty batch / `max_seq_len == 0` / zero heads → `grid.x == 0` → `invalid configuration argument`. Return early for empty output and add an empty-input test.

**Scope (non-blocking):** the dequant helper and the prefill scale-stride change are independent (the helper has its own kernel and doesn't depend on the prefill change). Splitting them would let the prefill change — which alters a shared function used by attention — be reviewed on its own. Not required for correctness.

**Validation — please run locally (no special hardware needed):** these paths are all SM80+ (`nvfp4_kv_dequantize_paged` is `@supported_compute_capability([80, 86, 89, 90, ...])`, and the NVFP4 prefill path runs on your RTX 3090 as in #3739), so please run the affected tests and paste results rather than leaving them to CI:
- `tests/attention/test_batch_attention.py::test_batch_attention_nvfp4` (the persistent-scale regression, item 1)
- the NVFP4 paged-prefill cases in `tests/attention/test_batch_prefill_kernels.py`
- `tests/utils/test_fp4_kv_quantization.py`
- `tests/trace/` (trace consistency + executing the dumped `init` from item 2)

lesj0610:
@qsang-nv Fixed the four persistent page_produce_kv_sf callsites to use the real scale strides.

The generated PersistentParams did not have scale stride fields, so I added them and wired the same GetFP4ScaleStrides helper that paged prefill uses. Moved it to tvm_ffi_utils.h for both paths.

NHD and HND trace init are now separate wrappers with the layout hardcoded, so the dumped JSON is self-contained. Added example calls and committed the generated JSONs. The test execs the dumped init without passing kv_layout.

nvfp4_kv_dequantize_paged returns early when batch_size, max_seq_len, or num_heads is zero.

SM86 local results:
- pre-commit: passed
- test_batch_prefill_kernels -k nvfp4: 50 passed
- test_fp4_kv_quantization: 38 passed, 16 skipped
- test_fi_trace: 27 passed
- full tests/trace: 821 passed, 196 skipped, 10 failed

The 10 trace failures are CUTLASS DSL norm compile errors (`llvm.mlir.global_dtors op requires attribute data`), not from the new NVFP4 dequant trace. The same 10 fail on the base commit f4841360.

test_batch_attention_nvfp4 also fails on SM86 with the cooperative launch limit, same on f4841360. Both failures are pre-existing.

qsang-nv:
Thanks — both blockers are addressed, and I verified the persistent-scale fix at runtime.

**persistent scale (blocker) — fixed & verified.** The four `persistent.cuh` call sites now pass `params.k_sf_stride_*`, `PersistentParams` carries the fields, and `GetFP4ScaleStrides` is shared via `tvm_ffi_utils.h`. I ran `test_batch_attention_nvfp4` on an H100 (SM90) — the cooperative-launch limit you hit on SM86 doesn't trigger there, and every case passes:

```text
tests/attention/test_batch_attention.py::test_batch_attention_nvfp4
32 passed, 2 warnings in 3314.32s
```

**trace HND (blocker) — fixed & verified.** The dedicated `_nvfp4_kv_dequantize_paged_hnd_init` (layout hardcoded) plus `_trace_init_dependencies` makes the dumped JSON self-contained. Executing the committed HND JSON's `init` without passing `kv_layout` now builds correct HND-shaped inputs (`k_cache` dim1 == num_heads), no NameError.

**[Nit]** The two generated JSONs are committed and the example calls added, but the "Results:" file list at the top of `tests/trace/example.py` isn't updated with `nvfp4_kv_dequantize_paged_h2_dk64_dv128_ps4.json` and `nvfp4_kv_dequantize_paged_hnd_h2_dk64_dv128_ps4.json` (step 5 of the trace checklist). Please add them so the list stays complete.

qsang-nv:
There are conflicts, pls resolve.@lesj0610

lesj0610:
@qsang-nv I resolved the conflicts.

qsang-nv (approved):

PR #3771, opened 2026-06-30
[cute-dsl] MLA decode: add causal toggle for non-causal multi-query (EAGLE tree verify)

thanhhao98:
## What
Add a `causal: bool = True` parameter to the monolithic cuteDSL MLA-decode kernel, threaded `cute_dsl_mla_decode` → `_get_compiled_mla_kernel` → `BlackwellMultiHeadLatentAttentionForwardFP8.__init__`. When `causal=False`, the per-row MTP `k_bound` is set to `K` (every query attends the full sequence) instead of the spec-decoding causal bound `K - (S_q - 1) + q_tok`.

## Why
The fold kernel currently hardcodes the MTP causal-among-Q mask, which is correct for a linear MTP/chain. **EAGLE tree-drafting verify** needs a **non-causal shared-prefix pass** — all `q_len` draft tokens must attend the *entire* prefix (the tree structure among draft tokens is applied separately). There is no way to request that today on the decode path (prefill gained an `is_causal` toggle, decode did not). This unblocks a downstream sglang PR that enables EAGLE tree (topk>1) verify on the trtllm_mla / cutedsl_mla MLA backends.

## Compatibility
`causal=True` is the default → **no behavior change** for existing callers (the MTP-masked path covered by `test_cute_dsl_mla_decode_fold_sq` is unchanged).

## Test
Adds `test_cute_dsl_mla_decode_noncausal` (fp16 + fp8, H∈{16,64}, q_len∈{2,4}, bs∈{1,4}) asserting `causal=False` output matches the non-causal PyTorch reference (`apply_mtp_mask=False`); the existing fold_sq test covers `causal=True`.

Draft — opening for early review; the consuming sglang tree-on-MLA PR follows.

### DIFF
--- a/flashinfer/cute_dsl/attention/monolithic/mla_decode.py
+++ b/flashinfer/cute_dsl/attention/monolithic/mla_decode.py
@@ -131,6 +131,7 @@ def _get_compiled_mla_kernel(
     skip_correction_threshold: float = 0.0,
     is_workspace_size_zero: bool = False,
     enable_pdl: bool = False,
+    causal: bool = True,
 ) -> Callable:
     """Compile and cache an MLA decode kernel.
 
@@ -181,6 +182,7 @@ def _get_compiled_mla_kernel(
         num_heads=num_heads,
         seq_len_q=seq_len_q,
         fold_sq=fold_sq,
+        causal=causal,
     )
 
     # All dimensions as sym_int — this matches the original kernel's use of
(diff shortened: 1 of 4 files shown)


thanhhao98:
Consumer: sgl-project/sglang#29706 (EAGLE tree drafting on trtllm_mla/cutedsl_mla) calls `cute_dsl_mla_decode(..., causal=False)` from this PR for its non-causal shared-prefix verify pass — it's blocked on this toggle releasing.

PR #3772, opened 2026-06-30
feat(bench): adds variable-length top-k transform benchmark for decode/prefill

### LINKED ISSUE #9628: Add benchmarks for variable-size segments to FlashInfer
Skimming FlashInfer's `bench_topk.py`, the transform benchmarks (page_table / ragged) look like they use a fixed segment length for every row. Our assumption is that production sparse-attention top-k is variable-length per row: in decode each request has a different context length, and in prefill the valid length grows causally across query positions.

To have a benchmark more reflective of production workloads, we want to contribute a variable-length benchmark to FlashInfer.

The issue can be closed when a PR is merged or if the overall idea is rejected by the FlashInfer community.

elstehle:
## 📌 Description

Adds a `varlen` benchmark mode to `benchmarks/bench_topk.py` that exercises the fused top-k transform APIs (`top_k_page_table_transform`, `top_k_ragged_transform`) with realistic per-row variable segment lengths, instead of the fixed `lengths == seq_len` that the existing transform benchmarks use for every row.

**Why:** production sparse-attention top-k (e.g. DeepSeek-V3.2 DSA, ragged/paged KV) selects top-k over a per-row valid window described by `lengths` (and a `row_to_batch` mapping). The existing benchmarks always pass a uniform length, which hides the behaviors that matter most in production: load imbalance across rows/CTAs, the trivial `length <= k` short-circuit, and multi-CTA scheduling. This mode gives a realistic comparison to production workloads.

**What it adds:**

- Two regimes:
  - `decode`: one row per request; independent context lengths drawn from a distribution (`uniform`, `lognormal` short-skewed serving mix, `bimodal` short/long mix that stresses imbalance and the trivial path).
  - `prefill` (`causal`): `q_len` rows per request with causal-monotonic lengths that grow across query positions, wired via a `row_to_batch` mapping.
- Baselines per case (non-deterministic mode): FlashInfer default, FlashInfer `clusters` (SM100 only), and `torch.topk` over a tensor that is masked to the valid window once, outside the timed region, so it isolates selection cost rather than masking overhead.
- Deterministic and tie-break parity with the existing top-k ops, via the shared `--deterministic` and `--tie-break` flags: `--deterministic` also times the deterministic path and reports its slowdown versus non-deterministic, and `--tie-break` (which implies deterministic) adds tie-small and tie-large columns. The `clusters` column is omitted in these modes because that path is non-deterministic only.
- Ragged `offsets` are built as the exclusive prefix-sum of `lengths` (a packed KV layout); the value does not affect timing, so this only affects realism.
- Reports per-case length stats (`min`/`mean`/`max`) and `triv%` (fraction of rows with `length <= k`) so runs with different length distributions stay comparable.
- New flags: `--op varlen` (also included in `--op all`), `--length-dist`, `--varlen-k`, `--varlen-q-len`. Length draws use a fixed-seed generator for reproducibility, and per-case `RuntimeError`/OOM is caught and reported.

No library or kernel code is changed; this is a benchmark-script-only change.

## 🔍 Related Issues

N/A. Motivated by variable-length sparse-attention top-k (DeepSeek-V3.2 DSA).

### ✅ Pre-commit Checks

## 🧪 Tests

> Benchmark-only change (no pytest coverage). Variable-length correctness for these APIs is already covered in `tests/utils/test_topk.py` (e.g. `test_page_table_transform_variable_lengths`, `test_ragged_transform_variable_lengths`, `test_top_k_transform_with_row_starts`), which pass on B200 (SM100). Smoke-tested on B200 via `python benchmarks/bench_topk.py --op varlen`, including the `--deterministic` and `--tie-break` modes.

## Reviewer Notes

- Scoped as a separate `varlen` op rather than extending the existing fixed-length transform sweeps, so the scaling numbers are preserved and the SGLang-comparison paths (which assume uniform lengths) are untouched.
- The `torch(mask)` baseline masks invalid positions outside the timed region on purpose; including the mask would unfairly penalize torch, since the kernel reads `lengths` directly without materializing a masked tensor.
- `Clusters` column is omitted off SM100 (the clusters path requires Blackwell) and under `--deterministic`/`--tie-break` (the clusters path runs only in the non-deterministic mode).
- Memory note: the prefill regime expands rows (`num_requests x q_len`), so the heaviest default case (`r16 x q128 x l131072`, fp32) allocates about 2 GB for scores plus a matching masked copy. Fine on datacenter GPUs and OOM-guarded; happy to trim the default grid or make the batch and `max_len` sets CLI-configurable if preferred.

## Summary by CodeRabbit

* **New Features**
  * Added a new variable-length benchmark mode for top-k operations to better reflect realistic per-request workloads.
  * Extended CLI options to select the new mode and configure length distribution, query length, and top-k.

* **Benchmarking Improvements**
  * Added detailed generation of variable-length inputs and summary statistics for each benchmark case.
  * Benchmarks now compare page-table and ragged-style transforms, with optional cluster-path timing when supported.

### DIFF
--- a/benchmarks/bench_topk.py
+++ b/benchmarks/bench_topk.py
@@ -840,6 +840,8 @@ def bench_varlen_transform(
     dtype: torch.dtype,
     generator: torch.Generator,
     has_clusters: bool,
+    deterministic: bool = False,
+    compare_tie_break: bool = False,
 ) -> dict:
     """Benchmark a transform API on realistic variable-length segments."""
     device = torch.device("cuda")
@@ -852,20 +854,28 @@ def bench_varlen_transform(
     num_rows = scores.size(0)
     k = case.k
 
-    if transform == "page_table":
-
-        def flashinfer_fn():
+    def run(deterministic_mode, tie_break=TopKTieBreak.NONE):
+        if transform == "page_table":
             return flashinfer.top_k_page_table_transform(
-                scores, src_page_table, lengths, k, row_to_batch=row_to_batch
+                scores,
+                src_page_table,
+                lengths,
+                k,
+                row_to_batch=row_to_batch,
+                deterministic=deterministic_mode,
+                tie_break=tie_break,
             )
-
-    else:
-
-        def flashinfer_fn():
-            return flashinfer.top_k_ragged_transform(scores, offsets, lengths, k)
+        return flashinfer.top_k_ragged_transform(
+            scores,
+            offsets,
+            lengths,
+            k,
+            deterministic=deterministic_mode,
+            tie_break=tie_break,
+        )
 
     set_topk_algo("default")
-    fi_ms = bench_median_ms(flashinfer_fn)
+    fi_ms, fi_nondeterministic_ms = bench_flashinfer_modes(run, deterministic)
 
     len_min, len_mean, len_max, trivial_frac = summarize_lengths(lengths, k)
     result = {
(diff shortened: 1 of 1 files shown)


kahyunnam (approved):
LGTM!

PR #3776, opened 2026-06-30
perf(moe): support caller-provided combine output

samnordmann:
## Summary

Allow `MoeAlltoAll.combine()` to write into an optional caller-provided output tensor. The existing allocating API remains unchanged, and the output-buffer path uses the same communication and combine kernels.

Related vLLM integration: https://github.com/vllm-project/vllm/pull/47156

## Rationale

vLLM already preallocates the final MoE output. The current path creates a temporary tensor and then copies it:

```python
# Existing
temporary = combine(payload)
output.copy_(temporary)

# Proposed
combine(payload, output=output)
```

Writing directly to the caller-owned tensor removes one allocation and one serialized output-sized D2D copy per MoE layer. The communication algorithm and combine kernel are unchanged.

## Performance

One GB200 node, four GPUs, EP4, hidden size 8192, top-k 22, 512 experts, and BF16 output:

| Timed path, 8192 tokens/rank | Baseline | Direct output | Delta, 95% CI |
| --- | ---: | ---: | ---: |
| Eager combine + output copy | 817.648 us | 777.528 us | **-4.91%** [-4.93%, -4.89%] |
| CUDA-graph dispatch + combine | 1423.616 us | 1386.640 us | **-2.59%** [-2.61%, -2.57%] |

## Summary of updates

* **New Features**
  * Added a “combine into” mode that lets you supply a pre-allocated output tensor for MoE all-to-all combine.
  * Extended the Python `moe_a2a_combine` and `MoeAlltoAll.combine` APIs with an optional `output` parameter.

* **Bug Fixes**
  * Added early validation for `output` (must be CUDA, contiguous, and match the expected shape/dtype) before combine runs.

* **Tests**
  * Added tests for early rejection of CPU/non-contiguous outputs.
  * Expanded single-GPU coverage for in-workspace payloads and optional external output-buffer usage.

### DIFF
--- a/csrc/trtllm_moe_alltoall.cu
+++ b/csrc/trtllm_moe_alltoall.cu
@@ -331,8 +331,10 @@ void moeA2ACombineIntoOp(TensorView payload, int64_t localNumTokens, TensorView 
   TVM_FFI_ICHECK_EQ(output.ndim(), 2) << "output must be a 2D tensor";
   TVM_FFI_ICHECK_EQ(output.size(0), output_shape[0]);
   TVM_FFI_ICHECK_EQ(output.size(1), output_shape[1]);
-  TVM_FFI_ICHECK(output.dtype() == outputDtype_.value_or(payload.dtype()))
-      << "output dtype does not match output_dtype";
+  auto const expectedOutputDtype = outputDtype_.value_or(payload.dtype());
+  TVM_FFI_ICHECK(output.dtype() == expectedOutputDtype)
+      << "output dtype must match " << (outputDtype_.has_value() ? "output_dtype" : "payload dtype")
+      << " (expected " << expectedOutputDtype << ", got " << output.dtype() << ")";
   params.ep_size = static_cast<int>(epSize);
   params.ep_rank = static_cast<int>(epRank);
   params.local_num_tokens = static_cast<int>(localNumTokens);
--- a/csrc/trtllm_moe_alltoall.cu
+++ b/csrc/trtllm_moe_alltoall.cu
@@ -273,12 +273,12 @@ nvinfer1::DataType toNvDataType(DLDataType dtype) {
   return nvinfer1::DataType::kFLOAT;
 }
 
-Tensor moeA2ACombineOp(TensorView payload, int64_t localNumTokens, TensorView workspace,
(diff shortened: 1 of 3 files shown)


csrc/trtllm_moe_alltoall.cu
```diff
@@ -396,6 +401,23 @@ Tensor moeA2ACombineOp(TensorView payload, int64_t localNumTokens, TensorView wo
...
       << "moe_a2a_combine launch failed: " << cudaGetErrorString(err);
+}
+
+Tensor moeA2ACombineOp(TensorView payload, int64_t localNumTokens, TensorView workspace,
+                       TensorView metainfo, int64_t runtimeMaxTokensPerRank, int64_t epRank,
+                       int64_t epSize, int64_t topK, int64_t combinePayloadOffset,
+                       bool payloadInWorkspace, Optional<DLDataType> outputDtype_,
+                       Optional<TensorView> outputScales, double outputScalarScale,
+                       int64_t sfLayout) {
+  int64_t const elementsPerToken = payload.size(2);
```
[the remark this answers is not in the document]
samnordmann: Addressed in `3ba76b44`: the allocating wrapper now validates that `payload` is a valid CUDA tensor with rank 3 before reading `payload.size(2)`. Clang-format and mypy (Python 3.12) pass.

aleozlx (approved):

csrc/trtllm_moe_alltoall.cu
```diff
@@ -326,8 +326,15 @@ Tensor moeA2ACombineOp(TensorView payload, int64_t localNumTokens, TensorView wo
   auto const output_shape = outputDtype_.has_value() && outputDtype_.value() == dl_uint8
                                 ? tvm::ffi::Shape{localNumTokens, elementsPerToken / 2}
                                 : tvm::ffi::Shape{localNumTokens, elementsPerToken};
-  Tensor output =
-      alloc_tensor(output_shape, outputDtype_.value_or(payload.dtype()), payload.device());
+  CHECK_INPUT(output);
```
amirkl94, line 329: Can we move this check to the python path? I think this will show a clearer exception for the user

samnordmann: done

PR #3778, opened 2026-06-30
fix(mla): validate paged kv-cache shape in BatchMLAPagedAttentionWrapper.run

### LINKED ISSUE #3219: [BUG] BatchMLAPagedAttentionWrapper(backend="fa2") returns zeros for non-contiguous kv_indices on Ampere bf16
## Summary

`flashinfer.mla.BatchMLAPagedAttentionWrapper(backend="fa2")` produces incorrect attention output when `kv_indices` contains non-contiguous (scattered) physical page indices, on Ampere (SM 8.6) with bfloat16 inputs.

The same logical KV data placed at scattered physical pages produces a different result (specifically, **all zeros**) than the same data placed at contiguous physical pages. Both should produce identical attention output for identical Q.

This was observed when integrating the wrapper into vLLM's MLA decode path on Ampere, where the BlockManager fragments page assignments across the global cache pool. Single-prompt isolation (one page assigned, trivially "contiguous") produces correct output; multi-prompt scenarios (scattered pages) produce gibberish from the very first decode token.

## Environment

- **GPU**: NVIDIA RTX A4000 (Ampere SM 8.6)
- **PyTorch**: 2.11.0+cu130
- **FlashInfer**: 0.6.8.post1
- **dtype**: torch.bfloat16
- **backend**: "fa2"

## Minimal repro

Self-contained, ~150 lines, no external dependencies beyond `torch` and `flashinfer`. Runs in under a minute.

```python
"""
Minimal reproduction of FlashInfer fa2 MLA paged-indexing bug on Ampere.

The BatchMLAPagedAttentionWrapper(backend="fa2") produces different output
when given the SAME logical KV data via two different paged layouts:
  (A) data placed at physical pages [0, 1, 2, ...] with kv_indices = [0, 1, 2, ...]
  (B) data placed at scattered physical pages [P0, P1, P2, ...] with kv_indices = [P0, P1, P2, ...]

Both should produce identical attention output for the same q.
"""
import sys
import torch
import flashinfer


def build_metadata(num_pages, page_size, kv_lens, page_assignment, device):
    batch_size = len(kv_lens)
    qo_indptr = torch.arange(0, batch_size + 1, dtype=torch.int32, device=device)
    pages_per_batch = [len(p) for p in page_assignment]
    kv_indptr = torch.zeros(batch_size + 1, dtype=torch.int32, device=device)
    cum = 0
    for i, n in enumerate(pages_per_batch):
        cum += n
        kv_indptr[i + 1] = cum
    flat = []
    for p in page_assignment:
        flat.extend(p)
    kv_indices = torch.tensor(flat, dtype=torch.int32, device=device)
    kv_len_arr = torch.tensor(kv_lens, dtype=torch.int32, device=device)
    return qo_indptr, kv_indptr, kv_indices, kv_len_arr


def main() -> int:
    cap = torch.cuda.get_device_capability(0)
    print(f"Device: {torch.cuda.get_device_name(0)}, compute capability {cap}")
    print(f"FlashInfer: {flashinfer.__version__}\n")

    num_heads = 4
    head_dim_ckv = 512
    head_dim_kpe = 64
    page_size = 64
    pool_size = 256
    batch_size = 4
    kv_len = 50

    device = torch.device("cuda:0")
    dtype = torch.bfloat16

    ckv_pool = torch.zeros((pool_size * page_size, 1, head_dim_ckv), dtype=dtype, device=device)
    kpe_pool = torch.zeros((pool_size * page_size, 1, head_dim_kpe), dtype=dtype, device=device)

    torch.manual_seed(42)
    test_ckv = [torch.randn(page_size, 1, head_dim_ckv, dtype=dtype, device=device) for _ in range(batch_size)]
    test_kpe = [torch.randn(page_size, 1, head_dim_kpe, dtype=dtype, device=device) for _ in range(batch_size)]

    workspace = torch.empty(64 * 1024 * 1024, dtype=torch.int8, device=device)
    sm_scale = 1.0 / ((128 + 64) ** 0.5)

    def run_with_assignment(label, page_assignment):
        ckv_pool.zero_()
        kpe_pool.zero_()
        for batch_idx, pages in enumerate(page_assignment):
            phys_page = pages[0]
            start = phys_page * page_size
            end = start + page_size
            ckv_pool[start:end] = test_ckv[batch_idx]
            kpe_pool[start:end] = test_kpe[batch_idx]

        qo_indptr, kv_indptr, kv_indices, kv_len_arr = build_metadata(
            pool_size, page_size, [kv_len] * batch_size, page_assignment, device
        )
        wrapper = flashinfer.mla.BatchMLAPagedAttentionWrapper(workspace, backend="fa2")
        wrapper.plan(qo_indptr, kv_indptr, kv_indices, kv_len_arr,
                     num_heads, head_dim_ckv, head_dim_kpe, page_size,
                     False, sm_scale, dtype, dtype)
        torch.manual_seed(99)
        q_nope = torch.randn(batch_size, num_heads, head_dim_ckv, dtype=dtype, device=device)
        q_pe = torch.randn(batch_size, num_heads, head_dim_kpe, dtype=dtype, device=device)
        out = wrapper.run(q_nope, q_pe, ckv_pool, kpe_pool, return_lse=False)
        torch.cuda.synchronize()
        print(f"--- {label} ---")
        print(f"  page_assignment={page_assignment}")
        print(f"  output norm = {out.norm().item():.6f}")
        return out

    out_a = run_with_assignment("Case A: contiguous pages", [[0], [1], [2], [3]])
    out_b = run_with_assignment("Case B: scattered pages",  [[16], [47], [89], [203]])

    diff = (out_a.float() - out_b.float()).abs()
    rel_err = diff.max().item() / max(out_a.float().abs().max().item(), 1e-9)
    print(f"\n=== max abs diff = {diff.max().item():.6e}, relative error = {rel_err:.6e} ===")
    if rel_err < 1e-2:
        print("PASS")
        return 0
    print("FAIL — fa2 MLA mishandles non-contiguous kv_indices on Ampere")
    return 1


if __name__ == "__main__":
    sys.exit(main())
```

## Observed output on RTX A4000 (SM 8.6) with bf16

```
Device: NVIDIA RTX A4000, compute capability (8, 6)
FlashInfer: 0.6.8.post1

--- Case A: contiguous pages ---
  page_assignment=[[0], [1], [2], [3]]
  output norm = 32.750000

--- Case B: scattered pages ---
  page_assignment=[[16], [47], [89], [203]]
  output norm = 0.000000

=== max abs diff = 1.710938e+00, relative error = 1.000000e+00 ===
FAIL — fa2 MLA mishandles non-contiguous kv_indices on Ampere
```

With contiguous `kv_indices`, the kernel produces sensible nonzero attention output (norm 32.75). With scattered `kv_indices` over the same logical data, the kernel produces a tensor of all zeros. This points to the kernel mishandling page-table indirection when indices are non-contiguous.

## Impact

This blocks `BatchMLAPagedAttentionWrapper(backend="fa2")` for any production inference engine on Ampere hardware with paged KV caches and multi-sequence batching, because real-world page assignments are inherently scattered. We hit it integrating the wrapper into vLLM's MLA backend — once we worked around it via a "compaction" step (gathering active pages into a contiguous buffer and remapping `kv_indices` to `[0..N-1]` before calling the wrapper), the entire test prompt set produced coherent output.

## Workaround for downstream users

```python
ckv_paged = ckv_pool.view(num_pages_total, page_size, 1, head_dim_ckv)
kpe_paged = kpe_pool.view(num_pages_total, page_size, 1, head_dim_kpe)
sel = kv_indices.long()
ckv_compact = ckv_paged.index_select(0, sel).reshape(-1, 1, head_dim_ckv).contiguous()
kpe_compact = kpe_paged.index_select(0, sel).reshape(-1, 1, head_dim_kpe).contiguous()
new_kv_indices = torch.arange(kv_indices.shape[0], dtype=torch.int32, device=device)
wrapper.plan(qo_indptr, kv_indptr, new_kv_indices, kv_len_arr, ...)
out = wrapper.run(q_nope, q_pe, ckv_compact, kpe_compact)
```

The `index_select` overhead is within measurement noise on the integration path we tested (vLLM MLA decode at TP=8 on 8× A4000), so this workaround is functionally viable for production while the root cause is being investigated.

## Hypothesis on root cause

We did not patch the kernel directly. Empirically, the failure mode is consistent with an offset calculation that assumes contiguous physical pages — for example, computing a base pointer as `cache_base + kv_indices[0] * page_stride` and then advancing by `page_stride` for subsequent pages, instead of resolving each `kv_indices[i]` independently. Sequential `kv_indices = [0, 1, 2, ...]` always works; scattered ones produce zeros.

## Related

- Issue #3047 — earlier reported correctness issues on the FlashInfer-backed MLA path with batch-composition dependency. This bug exhibits the same "fails on certain batch compositions" pattern.

## What would resolve

A fix to the fa2 MLA kernel's paged-indexing on non-Blackwell architectures so that arbitrary (non-contiguous, non-monotonic) `kv_indices` produce correct output, matching the contiguous-indices case.

Happy to provide additional diagnostics, capture Nsight traces, or test fixes against the integration. The standalone repro above and a vLLM-side end-to-end repro (linked from https://github.com/MidasMining/universal-translator) both reproduce the failure.

zhangchushu:
## 📌 Description
  
  The fa2/fa3 paged MLA kernels derive each kv-cache slot offset from the
  `page_size` captured in `plan()`, assuming the layout
  `[num_pages, page_size, head_dim]`. Passing a collapsed layout such as
  `[num_pages * page_size, 1, head_dim]` silently addresses the wrong rows and
  produces all-zero / incorrect output instead of raising.

  This PR adds `_check_fa_mla_cache_shape()`, called eagerly in
  `BatchMLAPagedAttentionWrapper.run()`, validating `ckv_cache`/`kpe_cache`
  rank, `page_size` (dim-1), head dims, and matching `num_pages` — turning a
  silent correctness bug into a clear `ValueError`.

  ## 🔍 Related Issues

  Related to #3219

  ## 🚀 Pull Request Checklist
  
  ### ✅ Pre-commit Checks
  - [x] Installed `pre-commit`
  - [x] Installed the hooks
  - [x] Ran the hooks and fixed any reported issues (all passed)

  ## 🧪 Tests
  - [x] Added regression test `test_batch_mla_collapsed_cache_shape_raises` (fa2/fa3)
  - [x] All tests passing

## Summary by CodeRabbit

* **Bug Fixes**
  * Added stronger validation for paged MLA KV-cache tensor shapes to fail fast when inputs don’t match the planned page layout.
  * Improved runtime checks to ensure the KV-cache matches the planned page size, head dimensions, and page count before attention kernels run.
  * Added regression tests to verify an invalid “collapsed” KV-cache layout raises a clear `ValueError` for supported backends.

Note: defensive guard against an invalid (collapsed) cache layout only — does not implement arbitrary-page FA2 indexing, so it does not fully  close #3219.

### DIFF
--- a/flashinfer/mla/_core.py
+++ b/flashinfer/mla/_core.py
@@ -75,6 +75,73 @@ def _check_cutlass_shape(q_nope_pe, ckv_kpe_cache, kv_len, page_table):
         )
 
 
+def _check_fa_mla_cache_shape(
+    ckv_cache: torch.Tensor,
+    kpe_cache: torch.Tensor,
+    page_size: int,
+    head_dim_ckv: int,
+    head_dim_kpe: int,
+) -> None:
+    """Validate the paged kv-cache shape for the fa2/fa3 MLA kernels.
+
+    The fa2/fa3 paged MLA kernels index the kv-cache with the layout
+    ``[num_pages, page_size, head_dim]``, deriving each slot offset from the
+    ``page_size`` captured in ``plan()`` -- not from the tensor's own dim-1. A
+    collapsed layout such as ``[num_pages * page_size, 1, head_dim]`` therefore
+    addresses the wrong rows and silently produces all-zero / incorrect output
+    instead of raising. This validates eagerly so the failure is obvious.
+
+    See https://github.com/flashinfer-ai/flashinfer/issues/3219.
+
+    Args:
+        ckv_cache: Compressed-KV cache, expected ``[num_pages, page_size,
+            head_dim_ckv]``.
+        kpe_cache: Key positional-embedding cache, expected ``[num_pages,
+            page_size, head_dim_kpe]``.
+        page_size: Page size passed to ``plan()``; must match dim-1 of both caches.
+        head_dim_ckv: Expected last dim of ``ckv_cache``.
+        head_dim_kpe: Expected last dim of ``kpe_cache``.
+
+    Raises:
+        ValueError: If either cache is not 3D, dim-1 does not equal
+            ``page_size``, the last dim does not match the corresponding head
+            dim, or the two caches disagree on ``num_pages``.
+    """
+    if ckv_cache.ndim != 3 or kpe_cache.ndim != 3:
+        raise ValueError(
+            "ckv_cache and kpe_cache must be 3D "
+            "[num_pages, page_size, head_dim], got "
+            f"ckv_cache.shape={tuple(ckv_cache.shape)}, "
+            f"kpe_cache.shape={tuple(kpe_cache.shape)}"
+        )
+    if ckv_cache.shape[1] != page_size or kpe_cache.shape[1] != page_size:
+        raise ValueError(
+            "ckv_cache/kpe_cache dim-1 must equal the page_size passed to "
+            f"plan() (page_size={page_size}), got "
+            f"ckv_cache.shape={tuple(ckv_cache.shape)}, "
+            f"kpe_cache.shape={tuple(kpe_cache.shape)}. The MLA paged kv-cache "
+            "layout is [num_pages, page_size, head_dim]; a collapsed layout "
+            "such as [num_pages * page_size, 1, head_dim] is a common cause of "
+            "all-zero or incorrect output "
+            "(https://github.com/flashinfer-ai/flashinfer/issues/3219)."
+        )
+    if ckv_cache.shape[2] != head_dim_ckv:
+        raise ValueError(
+            f"ckv_cache last dim must equal head_dim_ckv={head_dim_ckv}, got "
+            f"{ckv_cache.shape[2]}"
+        )
+    if kpe_cache.shape[2] != head_dim_kpe:
+        raise ValueError(
+            f"kpe_cache last dim must equal head_dim_kpe={head_dim_kpe}, got "
+            f"{kpe_cache.shape[2]}"
+        )
+    if ckv_cache.shape[0] != kpe_cache.shape[0]:
+        raise ValueError(
+            "ckv_cache and kpe_cache must have the same num_pages (dim-0), got "
+            f"{ckv_cache.shape[0]} and {kpe_cache.shape[0]}"
+        )
+
+
 @dataclass(frozen=True)
 class MLAHeadDimensions:
     """
@@ -1604,6 +1671,8 @@ class BatchMLAPagedAttentionWrapper:
             self._kv_len_arr_buf = kv_len_arr.to(self.device, non_blocking=True)
         self._causal = causal
         self._page_size = page_size
+        self._head_dim_ckv = head_dim_ckv
+        self._head_dim_kpe = head_dim_kpe
         self._sm_scale = sm_scale
         self._use_profiler = use_profiler
 
@@ -1767,6 +1836,13 @@ class BatchMLAPagedAttentionWrapper:
                 )
         num_heads = q_nope.shape[1]
         page_size = self._page_size
+        _check_fa_mla_cache_shape(
+            ckv_cache,
+            kpe_cache,
+            page_size,
+            self._head_dim_ckv,
+            self._head_dim_kpe,
+        )
         sm_scale = self._sm_scale
         causal = self._causal
         mask_mode = MaskMode.CAUSAL.value if causal else MaskMode.NON_CAUSAL.value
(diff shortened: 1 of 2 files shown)


zhangchushu:
@coderabbitai review

bkryu:
Hi @zhangchushu , please check [this comment](https://github.com/flashinfer-ai/flashinfer/issues/3219#issuecomment-5457495061) in Issue #3219.

The documented `ckv_cache` shape is `[num_pages, page_size, head_dim_ckv]` so

> The fa2/fa3 paged MLA kernels derive each kv-cache slot offset from the
page_size captured in plan(), assuming the layout
[num_pages, page_size, head_dim].

is not an assumption, it is an expected contract from the user. I will close the PR but please feel to reopen if you feel like the explanation is incorrect.

PR #3793, opened 2026-07-01
fix(grouped_mm): require cuDNN >= 9.21 for the FP8 MOE grouped matmul

### LINKED ISSUE #3792: [Bug] cuDNN FP8 MOE grouped matmul silently computes only the first expert group with backend 9.18–9.20 (SM120)
## Summary

`flashinfer.grouped_mm.grouped_mm_fp8` (cuDNN backend) returns **silently wrong results** with cuDNN backend **9.18.0 – 9.20.0**: only the **first expert group** is computed; every group `e >= 1` of the output is left **zero**. `graph.check_support()` passes, no error or warning is raised. Fixed in cuDNN **9.21.0**.

The current version floor admits the broken versions: `_CUDNN_MOE_MIN_VERSION = 91800` (`flashinfer/grouped_mm/cudnn/core.py:48`) is used for `grouped_mm_fp8` (`flashinfer/grouped_mm/core.py:269`), and the test gate `requires_cudnn_moe` (`tests/grouped_mm/conftest.py:28`) is `>= 91800` as well.

**Blast radius:** `torch 2.11.0+cu130` pins `nvidia-cudnn-cu13==9.19.0.56` — i.e. a stock cu13 PyTorch environment gets exactly a broken version, and MoE FP8 grouped GEMM silently produces garbage for every expert but the first.

## Reproduction (RTX PRO 6000 Blackwell, SM120, CUDA 13, torch 2.11.0+cu130)

```python
import torch
from flashinfer.grouped_mm import grouped_mm_fp8

torch.manual_seed(42)
E, tpe, k, n = 4, 64, 256, 256
a = torch.randn(E * tpe, k, device="cuda").clamp(-1, 1).to(torch.float8_e4m3fn)
b = torch.randn(E, n, k, device="cuda").clamp(-1, 1).to(torch.float8_e4m3fn)
m_indptr = (torch.arange(E + 1, device="cuda") * tpe).to(torch.int32)
alpha = torch.tensor([1.0], dtype=torch.float32, device="cuda")

out = grouped_mm_fp8(a, b, m_indptr, alpha=alpha, out_dtype=torch.bfloat16)

for e in range(E):
    s, t = e * tpe, (e + 1) * tpe
    ref = a[s:t].float() @ b[e].float().T
    ok = torch.isclose(out[s:t].float(), ref, atol=0.125, rtol=0.125).float().mean()
    zero = (out[s:t] == 0).float().mean()
    print(f"expert {e}: match={ok:.3f} zero-frac={zero:.3f}")
```

With cuDNN 9.19.0.56:

```
expert 0: match=1.000 zero-frac=0.004
expert 1: match=0.014 zero-frac=1.000
expert 2: match=0.010 zero-frac=1.000
expert 3: match=0.008 zero-frac=1.000
```

Same pattern for `float8_e5m2`, all shapes tried (`tpe ∈ {32..256}`, `k ∈ {256,512}`, `n ∈ {256,1024}`), with and without `alpha`. `pytest tests/grouped_mm/test_grouped_mm_fp8.py` → **58 failed** (everything with `num_experts > 1`).

## What rules out the wrapper

`grouped_mm_bf16` goes through the **same** `_run_cudnn_moe_grouped_gemm` / `_build_cudnn_moe_grouped_gemm_graph` code path (same `first_token_offset` construction, `flashinfer/grouped_mm/cudnn/core.py:248`) and is correct on every cuDNN version tested — `tests/grouped_mm/test_grouped_mm_bf16.py` = 46 passed on 9.19.0.56. Only the FP8 data types trip it, so the `first_token_offset`/graph semantics on the FlashInfer side are fine; the regression is inside the cuDNN backend's FP8 MOE kernels.

## cuDNN backend bisect (pip `nvidia-cudnn-cu13`, `tests/grouped_mm/test_grouped_mm_fp8.py`)

| cuDNN backend | result |
| --- | --- |
| 9.18.0.77 | 58 failed / 34 passed |
| 9.18.1.3 | 58 failed / 34 passed |
| 9.19.0.56 (torch 2.11 cu13 pin) | 58 failed / 34 passed |
| 9.19.1.2 | 58 failed / 34 passed |
| 9.20.0.48 | 58 failed / 34 passed |
| **9.21.0.82** | **92 passed** |
| 9.22.0.52 | 92 passed |
| 9.23.2.1 | 92 passed |

(The passing subset on broken versions is the `num_experts = 1` / empty-expert cases.)

Bonus datapoint: with 9.23.2.1 the whole `tests/grouped_mm/` directory is **305 passed** on SM120, including the `>= 9.21` block-scale MXFP8/FP4 suites.

## Scope caveat

Verified on **SM120 (RTX PRO 6000 Blackwell, CUDA 13)** only — I don't have SM90/SM100 hardware to check whether 9.18–9.20 mis-execute there too. bf16 vs fp8 discrimination above is from the same SM120 box.

## Proposed fix

Mirror the existing block-scale floor: add `_CUDNN_MOE_FP8_MIN_VERSION = 92100` and use it in `grouped_mm_fp8`'s runtime check (`flashinfer/grouped_mm/core.py:269`), plus a matching `requires_cudnn_moe_fp8` test gate, so the broken combination fails fast with an actionable error instead of silently corrupting MoE outputs. PR incoming.

Found while validating current `main` (`8fc7f079`) on RTX PRO 6000 for the v0.6.14 SM12x release push (#3783). AI-assisted (Claude Code) under human direction; every number above is from live runs on the real card.

waynehacking8:
## Description

cuDNN backends **9.18.0 – 9.20.0** pass `check_support()` for the FP8 MOE grouped matmul but **silently compute only the first expert group** — every group `e >= 1` of the output is left zero, with no error or warning (full analysis, repro, and per-version bisect in #3792). bf16 through the same graph-builder path is correct on every version, so the wrapper semantics are fine; the regression is inside the cuDNN backend's FP8 MOE kernels, fixed in 9.21.0. Notably, `torch 2.11.0+cu130` pins `nvidia-cudnn-cu13==9.19.0.56` — a broken version — so a stock cu13 PyTorch environment hits this.

This PR mirrors the existing block-scale floor (`_CUDNN_MOE_BLOCK_SCALE_MIN_VERSION`):

- add `_CUDNN_MOE_FP8_MIN_VERSION = 92100` and use it in `grouped_mm_fp8`'s runtime check, so the broken combination fails fast with an actionable `RuntimeError` instead of silently corrupting MoE outputs
- add a matching `requires_cudnn_moe_fp8` test gate and switch the FP8 kernel tests to it; `TestGroupedMmFp8Validation` keeps the broad `>= 9.18` gate (argument validation runs before the version check, so those stay covered on 9.18–9.20 installs)
- add `test_version_floor_enforced`: monkeypatches `backend_version()` to 9.20.0 and asserts the `RuntimeError`, so the floor itself is exercised on any installed cuDNN

Scope note: the bisect evidence is from SM120 (I don't have SM90/SM100 hardware to check whether 9.18–9.20 also mis-execute there). Since the failure mode is silent wrong results and 9.21+ is freely available, a global floor seemed the safer call — happy to make it arch-conditional if you know 9.18–9.20 to be good on other archs.

## Related Issues

Fixes #3792

## Verification (RTX PRO 6000, SM120, CUDA 13, torch 2.11.0+cu130)

| cuDNN backend | `test_grouped_mm_fp8.py` | `test_grouped_mm_bf16.py` |
| --- | --- | --- |
| 9.23.2.1 | 93 passed | 46 passed |
| 9.19.0.56 (broken, torch pin) | 7 passed (validation + floor test) / 86 skipped | 46 passed |

Direct `grouped_mm_fp8` call on 9.19: `RuntimeError: cuDNN grouped_mm_fp8 requires backend version >= 92100, found 91900.`

`ruff check` / `ruff format --check` clean on all touched files.

## Pull Request Checklist

- [x] Pre-commit checks (ruff clean on touched files)
- [x] Tests added (`test_version_floor_enforced`) and updated (gate switch); verified on real SM120 hardware under both a broken and a fixed cuDNN

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved FP8 grouped matrix-multiplication version checks to require the correct cuDNN minimum version.
  * Added clearer handling for cuDNN versions that support earlier grouped operations but not FP8 behavior fully.
  * Updated test coverage to skip FP8 cases unless the runtime meets the FP8-specific cuDNN requirements.

### DIFF
--- a/flashinfer/grouped_mm/core.py
+++ b/flashinfer/grouped_mm/core.py
@@ -22,6 +22,7 @@ from ..api_logging import flashinfer_api
 from ..utils import backend_requirement, supported_compute_capability
 from .cudnn import (
     _CUDNN_MOE_BLOCK_SCALE_MIN_VERSION,
+    _CUDNN_MOE_FP8_MIN_VERSION,
     _CUDNN_MOE_MIN_VERSION,
     _check_cudnn_version,
     _run_cudnn_moe_block_scale_grouped_gemm_fp4,
@@ -266,7 +267,7 @@ def grouped_mm_fp8(
         out_dtype = out.dtype
 
     if backend == "cudnn":
-        _check_cudnn_version(_CUDNN_MOE_MIN_VERSION, "grouped_mm_fp8")
+        _check_cudnn_version(_CUDNN_MOE_FP8_MIN_VERSION, "grouped_mm_fp8")
         return _run_cudnn_moe_grouped_gemm(
             a,
             b,
--- a/flashinfer/grouped_mm/cudnn/__init__.py
+++ b/flashinfer/grouped_mm/cudnn/__init__.py
@@ -9,6 +9,7 @@ helpers below.
 
 from .core import (
     _CUDNN_MOE_BLOCK_SCALE_MIN_VERSION,
+    _CUDNN_MOE_FP8_MIN_VERSION,
     _CUDNN_MOE_MIN_VERSION,
     _check_cudnn_version,
     _run_cudnn_moe_block_scale_grouped_gemm_fp4,
@@ -18,6 +19,7 @@ from .core import (
 
 __all__ = [
     "_CUDNN_MOE_BLOCK_SCALE_MIN_VERSION",
+    "_CUDNN_MOE_FP8_MIN_VERSION",
     "_CUDNN_MOE_MIN_VERSION",
     "_check_cudnn_version",
     "_run_cudnn_moe_block_scale_grouped_gemm_fp4",
--- a/flashinfer/grouped_mm/cudnn/core.py
+++ b/flashinfer/grouped_mm/cudnn/core.py
@@ -46,6 +46,11 @@ except OSError as e:
 # ---------------------------------------------------------------------------
 
 _CUDNN_MOE_MIN_VERSION = 91800  # 9.18.0
+# FP8 needs a higher floor than bf16: backends 9.18.0-9.20.0 pass
+# check_support() for the FP8 MOE grouped matmul but silently compute only the
+# first expert group, leaving every other group's output zero (gh #3792,
+# verified on SM120 / RTX PRO 6000; fixed in 9.21.0).
+_CUDNN_MOE_FP8_MIN_VERSION = 92100  # 9.21.0
 _CUDNN_MOE_BLOCK_SCALE_MIN_VERSION = 92100  # 9.21.0
 
 
--- a/tests/grouped_mm/conftest.py
+++ b/tests/grouped_mm/conftest.py
@@ -35,6 +35,16 @@ requires_cudnn_moe_block_scale = pytest.mark.skipif(
     reason="cuDNN MOE block-scale requires backend >= 9.21.0 and a frontend exposing moe_grouped_matmul_mode",
 )
 
+# FP8 needs a higher floor than bf16: backends 9.18.0-9.20.0 pass
+# check_support() but silently compute only the first expert group (gh #3792,
+# fixed in 9.21.0).
+requires_cudnn_moe_fp8 = pytest.mark.skipif(
+    not CUDNN_AVAILABLE or CUDNN_BACKEND_VERSION < 92100 or not CUDNN_HAS_MOE_API,
+    reason="cuDNN MOE FP8 requires backend >= 9.21.0 (9.18-9.20 silently compute "
+    "only the first expert group, gh #3792) and a frontend exposing "
+    "moe_grouped_matmul_mode",
+)
+
 
 def _requires_supported_cc(check_fn):
     # Skip mark whose allowlist is the `_supported_ccs` set attached to
--- a/tests/grouped_mm/test_grouped_mm_fp8.py
+++ b/tests/grouped_mm/test_grouped_mm_fp8.py
@@ -8,6 +8,7 @@ from flashinfer.grouped_mm import grouped_mm_fp8
 from .conftest import (
     ref_grouped_mm,
     requires_cudnn_moe,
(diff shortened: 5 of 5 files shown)


waynehacking8:
Re CodeRabbit's nitpick (add `trace=` to the public `grouped_mm` APIs): agreed the gap is real, but it's pre-existing for all four `grouped_mm_*` entry points and predates this PR — this diff only changes a version constant, a check-site, and test gates, so wiring `TraceTemplate`s here (templates + `tests/trace/example.py` + regenerated `fi_trace_out` JSONs, per the CLAUDE.md checklist) would bury the correctness fix in an unrelated change. Happy to do it as a follow-up PR if maintainers want it.

waynehacking8:
Additional validation for the 9.21 floor: shape-fuzzed the grouped_mm family on cuDNN 9.23.2.1 (SM120) with 240 randomized valid-contract configs - non-uniform and empty expert splits, num_experts 1..64, cum_m 1..4096, unaligned k/n from 8 to 1024 (including 24/120/136), e4m3/e5m2 mixes, alpha on/off - against the loop-matmul reference at the tests' tolerance. Zero silent mismatches, zero errors. So post-9.21 the family looks robust across shapes; the 9.18-9.20 first-group-only corruption remains the one failure mode found.

waynehacking8:
#3797 landed the same >=9.21 floor (now a single global _CUDNN_MOE_MIN_VERSION = 92100 in grouped_mm/cudnn/core.py), so main already rejects 9.18-9.20 instead of silently computing one group. Closing this as superseded. The 240-config shape-fuzz test from this branch is still available if anyone wants the coverage.

Issue #3570, opened 2026-06-10
Non-Causal TRTLLM-Gen support for Decode


benchislett:
See https://github.com/flashinfer-ai/flashinfer/issues/2826 for context.

For speculative decoding (Qlen <= 16) we typically use `trtllm_batch_decode_with_kv_cache`. Currently we only have support for non-causal in the context pathway, which does not help for speculative decoding.

A key requirement here is compatibility with FP8 KV-Cache. Ideally we would also have MLA support.

For DFlash speculative decoding, this is on the pareto for nearly all models served on B200s, particularly showing strong results at low-latency serving. Long-context use-cases are underperforming because we do not have efficient attention kernels for this use-case and we cannot currently apply quantized kv-cache on Blackwell as there is no available kernel.

For diffusion LLMs (LLada, [DiffusionGemma](https://blog.google/innovation-and-ai/technology/developers-tools/diffusion-gemma-faster-text-generation/)) depending on the block size we may also want to use the decode-optimized kernel.

elwhyjay:
Opened PR #3629 for this. Tests are written but blocked on Dense paged GQA generation cubins missing from the current flashinfer-cubin package - a follow-up TRT-LLM PR is needed to unblock them.

saltyminty:
Duplicate of https://github.com/flashinfer-ai/flashinfer/issues/3335, but leaving this open for tracking.

saltyminty:
We currently don't plan on extending trtllm-gen to support non-causal decode; this path is instead supported by the PrimTS path (see discussion here: https://github.com/flashinfer-ai/flashinfer/pull/3629#issuecomment-5419054188).

Current ongoing work:
- #4739 exposes `backend="prims-ts"` and `is_causal=False` through `BatchDecodeWithPagedKVCacheWrapper`.
- #4414 adds the mixed-precision path needed for BF16 queries with FP8/NVFP4 KV cache. 

Can you try these out and let us know if they work for you?

Issue #3335, opened 2026-05-15
TRTLLM-gen attention with non-causal support for decode


xinli-sw:
TRTLLM-gen has support for non-causal (bidirectional) attention. We would like support for this to accelerate inference for [DFlash](https://huggingface.co/Qwen/Qwen3.5-9B). Could flashinfer expose the non-causal support for these kernels? Currently causal=True must be used with [TRTLLM-gen attention](https://github.com/flashinfer-ai/flashinfer/blob/9276e441898789e3ff21c397ca942103bf76cb12/flashinfer/decode.py#L2489) as far as I can tell

Prefill path was added in https://github.com/flashinfer-ai/flashinfer/pull/3020

benchislett:
Note that DFlash-style non-causal-decode is basically just standard decode attention without masking within the (fairly short) nextn query block. This is a subcase of tree verification attention (where any dynamic attention pattern within the query block is permitted), for which we will (eventually) also want kernels.

At this point it seems much simpler to me to support non-causal in the decode spec-verify path, but we would also be very happy with a tree verification kernel as well.

In either case: we will want this for GQA _and_ MLA decode. FP8 KV-Cache support is also desirable.

nvpohanh:
cc @elvischenv

saltyminty:
We currently don't plan on extending trtllm-gen to support non-causal decode; this path is instead supported by the PrimTS path (see discussion here: https://github.com/flashinfer-ai/flashinfer/pull/3629#issuecomment-5419054188). The core kernels (including non-causal decode) landed in #4357

Current ongoing work:

- https://github.com/flashinfer-ai/flashinfer/pull/4739 exposes backend="prims-ts" and is_causal=False through BatchDecodeWithPagedKVCacheWrapper.
- https://github.com/flashinfer-ai/flashinfer/pull/4414 adds the mixed-precision path needed for BF16 queries with FP8/NVFP4 KV cache. 

Can you try these out and let us know if they work for you, and what coverage gaps still exist?

PR #3794, opened 2026-07-01
feat(cute-dsl): sliding window and attention sinks for GQA decode (non-paged + paged)

richardmcai:
## 📌 Description

Adds sliding window attention (left and right bounds) and attention sink support to the CuTe-DSL GQA decode backend, for both the non-paged and paged kernel paths. Also introduces a new `AttentionMask` abstraction that makes masking configurations JIT-composable.

- **Attention sinks**: removes the `NotImplementedError` guard in `decode.py` and `batch_decode.py`; the `sinks` tensor is now passed through to the cute-dsl kernel for both paged and non-paged decode.
- **Sliding window (left + right)**: removes the sliding-window `NotImplementedError` guard; `window_left` is wired through to the kernel. A new `window_right` parameter is added to `BatchDecodeWithPagedKVCacheWrapper.plan` (keyword-only, defaults to `0`; `cute-dsl` backend only) to support right-boundary masking.
  - set `window_left=-1, window_right=-1` for dense mask
  - set `window_left=-1, window_right=0` for causal mask
  - set `window_left>=0, window_right>=0` for sliding window
- **`AttentionMask` ABC **: new JIT composable masking API in `fusion/mask.py` that allows users to define their own custom masking configurations (supports gqa decode only for now)
- **`plan` keyword-only args**: `pos_encoding_mode`, `window_left`, and all optional kwargs after `page_size` are now keyword-only (`*` separator) to prevent future positional-arg confusion, with a deprecation shim added to not break existing positional call usages. No existing call sites are affected (verified across tests and benchmarks). Existing users passing positional args will receive a deprecation warning, no usages will be broken until the deprecation shim is removed in a future release.

## 🔍 Related Issues

https://github.com/flashinfer-ai/flashinfer/issues/3335

https://github.com/flashinfer-ai/flashinfer/issues/3570

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

- [x] `pytest tests/attention/test_cute_dsl_decode.py -v` — new sliding-window and sink test cases
- [x] `pytest tests/attention/test_cute_dsl_decode.py -k sliding_window`
- [x] `pytest tests/attention/test_cute_dsl_decode.py -k sink`

## Summary of changes

- **New Features**
  - Added runtime-configurable attention masking for CuTe DSL decode, including sliding-window support via `window_left` and `window_right`.
  - Added optional “attention sink” logits support to decode, with updated wrapper and backend plumbing.
  - Updated CuTe DSL batch decode planning/execution APIs to accept the new window bounds and sink options.

- **Bug Fixes**
  - Improved masked softmax stability by using a safer reduction initialization when all values may be masked.

- **Tests**
  - Added/updated integration tests for sliding-window equivalence and sink correctness, and adjusted expectations for newly supported features.


flashinfer/cute_dsl/attention/wrappers/batch_decode.py
```diff
@@ -126,6 +127,18 @@ def _resolve_reduction(reduction: str, kv_splits: int, o_dtype: "cutlass.dtype")
...
+def _get_mask_config(
+    window_left: Optional[int],
+    window_right: Optional[int],
+) -> AttentionMask:
+    if window_left is None and window_right is None:
+        return DenseMask()
+    elif window_left is None and window_right == 0:
+        return CausalMask()
+    else:
+        return SlidingWindowMask(window_left, window_right)
```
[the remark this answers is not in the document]
richardmcai: The implementation can accept negative window bounds, only the wrapper treats -1 as disabling the bound.

flashinfer/cute_dsl/attention/wrappers/batch_decode.py
```diff
@@ -807,6 +838,8 @@ def run(
             o_partial,
             l_partial,
             m_partial,
+            None,  # sink_h
```
saltyminty: Seems we don't expose sinks in the non-paged wrapper path, is this intended?

flashinfer/cute_dsl/attention/wrappers/batch_decode.py
```diff
@@ -520,14 +546,16 @@ def plan(
         num_qo_heads: int,
         num_kv_heads: int,
         head_dim: int,
+        *,
```
saltyminty: We're introducing an API breaking change in this PR, which should only be done in major versions. We should probably introduce window_right at the end for now and make this change for 0.7.0

flashinfer/cute_dsl/attention/wrappers/batch_decode.py
richardmcai: Since sinks is passed at run() and not plan(), we only compile the no-sink variant after plan(), then recompile with sink at run() if the user passes sink. Similar behavior with LSE and skip softmax features.

flashinfer/cute_dsl/attention/wrappers/batch_decode.py
richardmcai: do you mean for `BatchDecodeWithPagedKVCacheWrapper`? This cute dsl wrapper shouldnt be used by other users afaik.

Also how should we introduce the change? Should I just add a TODO note and open a followup PR?

flashinfer/cute_dsl/attention/wrappers/batch_decode.py
```diff
@@ -861,14 +894,16 @@ def plan(
         num_kv_heads: int,
         head_dim: int,
         page_size: int,
+        *,
```
aleozlx: Hi please provide justification for this change

flashinfer/cute_dsl/attention/wrappers/batch_decode.py
aleozlx: please refer to semantic versioning in the contributing guide

https://github.com/flashinfer-ai/flashinfer/blob/main/CONTRIBUTING.md#release-versioning

concretely, it's possible to track this as a comment or an issue. the 0.6 api could not be changed throughout 0.7 during a posted deprecation period. only at 0.8 is the breaking change allowed to be applied

flashinfer/cute_dsl/attention/wrappers/batch_decode.py
richardmcai: this makes all args after `page_size` keyword-arg only, so that they cannot be passed in positionally. Claude and Codex found that this already matches all existing usages of `plan` in the repo, so this would not break any existing usages internally. This also already matches the signature of `run` which makes all args besides QKV keyword only.

Doing this allows us to expose features in the attention kernels without potentially breaking users who depend on positional arguments. e.g. if a user depended on positional arguments, the new `window_right` arg I'm adding would have to be appended to the end of the function signature instead of its most natural position after `window_left`.

If there is a user who is depending on this positional call signature (unlikely) then we should break them earlier in the interest of enabling new features in the long term.

richardmcai: > To reduce disruption during deprecation and removal, we prefer "keyword only" (after an *, see [PEP-3102](https://peps.python.org/pep-3102/)) for parameters that are likely to come and go (e.g. perf parameters).

the contributing notes already suggests keyword only so this would just make the keyword mandatory.

richardmcai: also this is the wrong file to look at, this is just a shim to map `BatchDecodeWithPagedKVCacheWrapper` in decode.py to the cute dsl kernel. The real potential (but unlikely) breakage is in decode.py

richardmcai:
@saltyminty @aleozlx I restored the positional argument ordering in `BatchDecodeWithPagedKVCacheWrapper.plan()` and added a note for followup. the wrappers in batch_decode.py I left as keyword-only since these are internal shims that dont need to have stable API, and also they are only a month old and not tagged  to any major release so I highly doubt anyone has picked it up as a dependency.

flashinfer/cute_dsl/attention/wrappers/batch_decode.py
aleozlx: Yes I added that line to the contribution guide essentially 

But I was asking about what warrants a breaking change. Sorry my question wasn't clear 

The breaking change is what needs justification against, not why this line is good

aleozlx: It will reliably be detected by our QA anyway and I'd always comeback to this question for clarity. Without it I'd strip it from release branch on the spot

richardmcai: My justification for breaking is that this will keep the function signature readable and easy to maintain in the long term as we continue to add new kernel features, otherwise we can only update the function signature in chronological order which makes it confusing to understand. A developer might not realize `window_right` is supported if they dont see it immediately after `window_left`.

aleozlx: This is insufficient in my opinion for the following reasons:

1. this is covered by `@flashinfer_api` so it is a semantically versioned point
2. this is used by the user. they have even filed issues on it https://github.com/flashinfer-ai/flashinfer/issues/3716
3. i fully support your enthusiasm to make the api much better "early", but not "retroactively" and immediately breaking user's code
4. the suggested approach is to follow deprecation procedure patiently

richardmcai: Interesting, i did not realize that users were already using this internal wrapper...

In that case how can we indicate whether an API should or shouldn't be stable for these internal classes? Should we even make a guarantee for this?

richardmcai: @coderabbitai recommend a soft api deprecation path based on this discussion

aleozlx: >how can we indicate whether an API should or shouldn't be stable

`@flashinfer_api` decorator is the guarantee indicator

sure i'm ok with compatible soft changes. i'd leave that discretion up to attention code owners @saltyminty @qsang-nv

richardmcai: I guess my questions remaining are
1. should we deprecate positional args in `plan`, and if so should we add this soft deprecation path in this PR or a followup?
2. Should we do the same for these internal wrappers for cute dsl backend? Or is it fine to make a hard break since we havent annotated with `@flashinfer_api`
3. If it is ok to break the internal wrappers, should we eventually promote it to a stable API in the future since users seem interested in it?

aleozlx: >should we deprecate positional args in plan, and if so should we add this soft deprecation path in this PR or a followup?

if we do it in any "soft"/compatible way i actually support you doing it and put in the deprecation notice. separate or follow up PR both fine as long as we are on the same page. 

>since we havent annotated with @flashinfer_api

they have no stability guarantee on the contrary. anything goes

>should we eventually promote it to a stable API 

it's possible but until we decorate it, it's not a problem as we have to draw a line somewhere and QA rely on it to detect what is ok or not and what is the focus

richardmcai:
@aleozlx @saltyminty @qsang-nv added a shim to deprecate positional args in plan/run methods without breaking users. It seems I marked the cute dsl wrappers with @flashinfer_api so to preserve semver I applied the same changes to the cute dsl decode as well. Heres a summary from Codex

### Summary:
- Kept legacy optional positional calls working for:
  - `BatchDecodeWithPagedKVCacheWrapper.plan`
  - `BatchDecodeCuteDSLWrapper.plan`
  - `BatchDecodeCuteDSLWrapper.run`
  - `BatchDecodePagedCuteDSLWrapper.plan`
  - `BatchDecodePagedCuteDSLWrapper.run`
- Added `DeprecationWarning` for optional positional arguments and documented that they should be passed by keyword.
- Kept newly added parameters keyword-only where there is no released positional API:
  - `window_right`
  - `sinks`
- Preserved deprecated `is_causal` compatibility in the CuTe DSL plan wrappers by translating:
  - `is_causal=True` -> `window_left=None, window_right=0`
  - `is_causal=False` -> `window_left=None, window_right=None`
- Reject ambiguous `is_causal` usage when combined with explicit `window_left` or `window_right`.
- Reject duplicate legacy positional-plus-keyword values in the shim, matching normal Python call semantics.
- Leave unexpected kwargs to be rejected by the keyword-only `_plan_impl` / `_run_impl` signatures.
- Added focused monkeypatched tests for the shim behavior without triggering kernel compilation.

richardmcai:
the CI failures appear unrelated, all the B200/B300 tests passed. Will let others confirm.

flashinfer/cute_dsl/attention/wrappers/batch_decode.py
saltyminty: Yeah that makes sense – but BatchDecodeCuteDSLWrapper::run and _run_impl (batch_decode.py::952) don't take sinks as a parameter either

saltyminty (approved):

flashinfer/cute_dsl/attention/wrappers/batch_decode.py
richardmcai: sure, I can add that

PR #3489, opened 2026-06-01
add_cudnn_mxfp8

yanqinz2:
## 📌 Description

Added _cudnn_mm_mxfp8_requirement and registered the "cudnn" backend in @backend_requirement for mm_mxfp8.

Updated _heuristic_func_mm_mxfp8 so that backend="auto" falls back to cuDNN whenever CUTLASS is not selectable and cuDNN is available.

Added _cudnn_mm_mxfp8_runner enabling the cuDNN backend to participate in the existing tuner.choose_one autotuning flow.

Correct the layout passing logic to cudnn override shape path, to avoid cudnn runtime invalid dynamic shape error.

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [ ] Tests have been added or updated as needed.
- [ ] All tests are passing (`unittest`, etc.).

## Summary by CodeRabbit

* **New Features**
  * cuDNN backend support added for MXFP8 mm operations on NVIDIA GPUs (compute capabilities 10.0, 10.3, 11.0, 12.0, 12.1); cuDNN can be selected where available.

* **Bug Fixes**
  * Improved MXFP8 layout, stride and block-scale handling to preserve expected memory/layout behavior across backends.
  * Fixed MXFP8 input quantization and construction to maintain correctness.

* **Behavior**
  * cuDNN is excluded from automatic backend candidates unless explicitly requested.

PR #3798, opened 2026-07-02
fix(gemm): handle linear MXFP8 scales in bmm_mxfp8 cuDNN

### LINKED ISSUE #3663: [Bug] test_bmm_mxfp8 Cosine similarity 0.8984 is too low (expected > 0.9) on B300
Found this issue in [54916177](https://nv/flashinfer-ci/-/pipelines/54916177)
Failed Jobs: [unit_test_b300: [cu130]](https://nv/flashinfer-ci/-/jobs/342190325) 
History:

- 2026-06-10: PASSED
- 2026-06-12: PASSED
- 2026-06-13: PASSED
- 2026-06-14: PASSED 
- 2026-06-16: PASSED

tests/gemm/test_bmm_mxfp8.py
```shell
E   AssertionError: Cosine similarity 0.8984 is too low (expected > 0.9)
    assert tensor(0.8984, device='cuda:0', dtype=torch.bfloat16) > 0.9
/workspace/flashinfer/tests/gemm/test_bmm_mxfp8.py:74: AssertionError: Cosine similarity 0.8984 is too low (expected > 0.9)
-- generated xml file: /tmp/junit/tests_gemm_test_bmm_mxfp8.py.687030424.xml ---
=========================== short test summary info ============================
FAILED tests/gemm/test_bmm_mxfp8.py::test_bmm_mxfp8[False-cudnn-res_dtype0-False-input_dtype0-128-512-128-1] - AssertionError: Cosine similarity 0.8984 is too low (expected > 0.9)
assert tensor(0.8984, device='cuda:0', dtype=torch.bfloat16) > 0.9
```

coder-2011:
## Description

Fixes `bmm_mxfp8` cuDNN scale-layout handling for MXFP8 linear scale tensors.

Previously, `bmm_mxfp8` always built the cuDNN MXFP8 graph with scale tensors declared as `F8_128x4`. That was correct for swizzled scales, but incorrect when the inputs were quantized with `is_sf_swizzled_layout=False`, which produces linear scales.

This PR adds `sf_layout` argument to `bmm_mxfp8` and maps it to the cuDNN scale tensor reordering, where

- `SfLayout.layout_linear` -> `cudnn.tensor_reordering.NONE`
- `SfLayout.layout_128x4` -> `cudnn.tensor_reordering.F8_128x4`

The existing default remains `SfLayout.layout_128x4` to preserve current behavior. CUTLASS BMM continues to require 128x4 swizzled scales.

The BMM MXFP8 test now passes the scale layout used by the test into `bmm_mxfp8`, while leaving quantization behavior unchanged.

## Related Issues

Fixes #3663

### Pre-commit Checks

## Tests

- [x] Relevant tests are passing.

Tested with:

```bash
.venv/bin/pre-commit run --all-files
```

```bash
.venv/bin/python -m pytest -q \
  'tests/gemm/test_bmm_mxfp8.py::test_bmm_mxfp8[False-cudnn-res_dtype0-False-input_dtype0-128-512-128-1]' \
  'tests/gemm/test_bmm_mxfp8.py::test_bmm_mxfp8[False-cudnn-res_dtype0-True-input_dtype0-128-512-128-1]' \
  'tests/gemm/test_bmm_mxfp8.py::test_bmm_mxfp8[True-cudnn-res_dtype0-False-input_dtype0-128-512-128-1]' \
  'tests/gemm/test_bmm_mxfp8.py::test_bmm_mxfp8[True-cudnn-res_dtype0-True-input_dtype0-128-512-128-1]'
```

Random-seed stress for the original failing cuDNN linear-scale shape:

```text
backend=cudnn
input_dtype=torch.bfloat16
res_dtype=torch.bfloat16
is_sf_swizzled_layout=False
sf_layout=SfLayout.layout_linear
sampled seeds=10,000 unique seeds from range(0, 2**31)
minimum cosine=0.94921875
```

## Summary

* **New Features**
  * Added runtime-selectable FP8 scale reordering for MXFP8 GEMM/BMM, supporting linear and 128x4 layouts.
  * Extended `bmm_mxfp8` with a new keyword-only `sf_layout` argument to choose the scale format.
* **Bug Fixes**
  * Improved cuDNN MXFP8 autotuning/execution so selected scale-layout tactics don’t mix across different reordering layouts.
  * Added stricter validation for supported `sf_layout` options across backends.
* **Tests**
  * Updated MXFP8 BMM tests to cover both supported `sf_layout` values.


flashinfer/gemm/gemm_base.py
```diff
@@ -8707,6 +8720,7 @@ def _cudnn_bmm_mxfp8_requirement(
     dtype: torch.dtype,
     out: Optional[torch.Tensor] = None,
     backend: Literal["cudnn", "auto"] = "cudnn",
+    sf_layout: SfLayout = SfLayout.layout_128x4,
```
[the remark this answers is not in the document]
coder-2011: makes sense, resolved

flashinfer/gemm/gemm_base.py
```diff
@@ -8707,7 +8720,11 @@ def _cudnn_bmm_mxfp8_requirement(
...
 ):
+    sf_layout = SfLayout(sf_layout)
+    if sf_layout not in (SfLayout.layout_linear, SfLayout.layout_128x4):
+        return False
     return _cudnn_available_or_raise_for_backend(backend)
```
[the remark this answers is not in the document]
coder-2011: cuDNN MXFP8 matmul is supported on SM12x.

NVIDIA’s cuDNN release notes explicitly [mention:](https://docs.nvidia.com/deeplearning/cudnn/backend/v9.23.2/release-notes.html#:~:text=mxfp8%20block-scaled%20matrix%20multiplication%20on%20certain%20Blackwell-architecture%20GPUs%20with%20compute%20capability%2012.0%20and%2012.1%20%28sm_120%20and%20sm_121%29) 

“mxfp8 block-scaled matrix multiplication on certain Blackwell-architecture GPUs with compute capability 12.0 and 12.1 (sm_120 and sm_121)”

coder-2011:
ping @aleozlx  @yzh119

waynehacking8:
Ran this on an RTX PRO 6000 (SM120, CUDA 13, cudnn-frontend 1.25.0 / cudnn 9.19), since the cudnn path is admitted there at runtime (_cudnn_bmm_mxfp8_requirement allows [100, 103, 110, 120, 121]) even though the test file skips cudnn for anything but SM10x. The fix checks out: with linear scales on the cudnn backend, main produces scrambled output against a reference computed from the dequantized fp8 operands (median rel err ~0.99, cos 0.37) while this branch is exact modulo output rounding (max rel err 0.4%, cos 0.999999). Swizzled is unchanged between main and this branch (cos 0.9993 vs the pre-quant reference) and agrees with the cutlass backend, so no regression there.

One thing I had to work around to measure this, worth noting while you're in the test file: plain randn inputs can't see this bug -- every 32-block quantizes to nearly the same e8m0 scale, so main's broken linear path still scores cos 0.93 on my box and clears the test's 0.9 bar (with a median per-element rel error of 31%). I only got discrimination by scaling each 32-element block by a 2^-4..2^4 ladder and checking per-element error against the dequantized-operand reference.

That ladder input also exposed an issue in the existing test construction that your PR inherits: the test quantizes mat2 as contiguous [b, k, n], which blocks scales along n, while the graph declares b_descale as [b, k/32, n] with K-major strides (and the bmm_mxfp8 docstring says to pass B column-major). With block-varying magnitudes even the intended swizzled config drops to cos 0.756 using the test's construction, vs 0.9993 when B is quantized as [b, n, k] (blocks along k) and passed transposed -- on both cudnn and cutlass, main and PR alike. So as written the test can't catch scale-layout bugs, which is presumably how the linear-layout one shipped silently. Happy to share the probe script if useful.

coder-2011:
Thx, I think the B-axis construction bug can be solved in a future PR, probably by me.

coder-2011:
@dhiraj113 @yzh119 @aleozlx @bkryu

I woud appreciate if someone can take a look a this!

aleozlx:
let me help you escalate.

adding @YangXu1990uiuc

flashinfer/gemm/gemm_base.py
```diff
@@ -2897,6 +2898,8 @@ def build_cudnn_gemm_mxfp8_graph_override_shape(
     _check_cudnn_override_shape_availability()
     if policy is None:
         policy = cudnn.build_plan_policy.HEURISTICS_CHOICE
+    if scale_reordering is None:
```
dhiraj113, line 2904: Can you do this through the function default argument instead.

flashinfer/gemm/gemm_base.py
```diff
@@ -8517,10 +8520,13 @@ def build_cudnn_gemm_mxfp8_graph(
     block_size,
     o_type,  # cudnn.data_type, BF16 or FP16
     device,
+    scale_reordering=None,
```
dhiraj113, line 8621: Same here. Make the default value explicit in the function args.

flashinfer/gemm/gemm_base.py
```diff
@@ -8650,6 +8656,7 @@ def _cudnn_gemm_mxfp8(
     out: Optional[torch.Tensor] = None,
     workspace_buffer: torch.Tensor = None,
     tactic: int = -1,
+    scale_reordering=None,
```
dhiraj113, line 8757: here too.

flashinfer/gemm/gemm_base.py
```diff
@@ -8684,7 +8692,7 @@ def _cudnn_gemm_mxfp8(
     )
 
 
-def _cudnn_gemm_mxfp8_runner():
+def _cudnn_gemm_mxfp8_runner(scale_reordering):
```
dhiraj113: This looks incorrect. Why do you need pass `scale_reordering` this way when other things are not being passed like this.

flashinfer/gemm/gemm_base.py
```diff
@@ -8902,6 +8924,8 @@ def bmm_mxfp8(
     dtype: torch.dtype,
     out: Optional[torch.Tensor] = None,
     backend: Literal["cudnn", "cutlass", "auto"] = "auto",
+    *,
```
dhiraj113: API shouldn't be extended like this.

flashinfer/gemm/gemm_base.py
```diff
@@ -8902,6 +8924,8 @@ def bmm_mxfp8(
...
+    sf_layout: SfLayout = SfLayout.layout_128x4,
```
dhiraj113, line 9026: Since you are modifying the API to run the test, could you explain how this is a bug fix?

dhiraj113:
Also, the test in https://github.com/flashinfer-ai/flashinfer/issues/3663 doesn't pass with your change.

dhiraj113:
Here's the code review from Claude. See if it helps - 

Code Review: PR #3798 — fix(gemm): handle linear MXFP8 scales in bmm_mxfp8 cuDNN
What it does: Threads a scale_reordering parameter through the cuDNN MXFP8 GEMM graph builders/runners and adds a keyword-only sf_layout argument to bmm_mxfp8, mapping SfLayout.layout_linear → cudnn.tensor_reordering.NONE and layout_128x4 → F8_128x4. Previously the reordering was hardcoded to F8_128x4, so linearly-arranged scales (from mxfp8_quantize(..., is_sf_swizzled_layout=False)) were misinterpreted. The core mapping logic, the str(scale_reordering) cache-key addition, and the sf_layout forwarding through @backend_requirement are all correct — with a correctly-constructed B I measured cos_sim=1.0000 for both layouts.

I set up git (added coder-2011 remote, checked out fix-bmm-mxfp8-scale-reordering) and verified findings by running on the branch. Findings ranked most-severe first:

1. tests/gemm/test_bmm_mxfp8.py:40-44 — The PR does not make its target test pass; the cuDNN/SM100 case still errors out.
The PR claims to fix #3663 (B300 = SM100, cuDNN backend) and edits this exact test, but running the target case on the branch still fails:


ValueError: cuDNN mxfp8 GEMM requires B to be column-major [batch, k, n]
(stride [k*n, 1, k] …); got b.shape=(16, 128, 512), b.stride()=(65536, 512, 1).
The test builds mat2 = randn([b,n,k]).transpose(-2,-1).contiguous(), and .contiguous() makes B row-major — rejected by the cuDNN column-major guard (added in #3489). This hits both linear and swizzled cuDNN cases. The PR adds sf_layout= plumbing but never fixes the mat2 construction, so on the very hardware the issue is about, the test errors before the kernel runs. Fix: quantize the [b,n,k] weight and transpose the quantized result without .contiguous() (exactly what the error message prescribes). I confirmed that with correct B construction the fix yields cos_sim=1.0.

2. flashinfer/gemm/gemm_base.py:8559-8571 — Linear-scale path reuses the 128x4-padded descriptor → silently wrong results for non-aligned shapes.
_calculate_block_scale_dims always pads m→⌈m/128⌉·128, n→⌈n/128⌉·128, k/32→⌈·/4⌉·4. The graph builder uses these padded dims for the scale-tensor descriptor regardless of layout. But linear scales are unpadded (numel = b·m·⌈k/32⌉), so when m, n, or k/32 isn't 128x4-aligned, cuDNN reads with a wrong-shaped/strided descriptor. Empirically, k=96 (k/32=3): linear → cos_sim=0.7031 (wrong), swizzled → 1.0000. The test grid only uses aligned shapes (m,n ∈ {128,256,512}, k ∈ {128,256,512,1024}), so it never catches this. For the linear path the descriptor should use the unpadded dims [b, m, ⌈k/32⌉] / [b, ⌈k/32⌉, n], not the padded ones. At minimum, the linear layout should be validated/rejected for non-aligned shapes and the docstring should state the alignment restriction.

3. flashinfer/gemm/gemm_base.py:8794-8808 — mxfp8_gemm_sm100 adds scale_reordering as a required positional parameter (no default), inconsistent with every other touched function.
Every other function in this PR uses scale_reordering=None. Here it's required-positional. Only one caller exists so it doesn't break today, but it's a gratuitous API-shape inconsistency and a latent break for any future caller. Low severity — give it a default (= None) for consistency.

Bottom line: The fix's logic is sound and works for the aligned shapes the test uses, but (1) the shipped test still fails on the issue's own hardware because the unrelated mat2 row-major bug isn't fixed, and (2) the linear path is only correct for 128x4-aligned shapes — a real latent correctness bug the test can't see. I'd ask the author to fix the test's B construction (so the test actually exercises and validates the fix) and either generalize the descriptor for linear layout or explicitly guard/document the alignment restriction. The branch is checked out locally if you want to iterate on it.

flashinfer/gemm/gemm_base.py
coder-2011: I might be mistaken but, `cudnn.*` objects are never function defaults in the codebase, bc flashinfer would crash if we dont have `cudnn` imported.

ex.

```python
def build_cudnn_gemm_fp4_graph(
    ...,
    policy=None,
):
    if policy is None:
        policy = cudnn.build_plan_policy.HEURISTICS_CHOICE
 ```

agent said that there arent any `cudnn.*` function defaults in flashinfer

flashinfer/gemm/gemm_base.py
coder-2011: fixed!

flashinfer/gemm/gemm_base.py
coder-2011: fixed!

flashinfer/gemm/gemm_base.py
coder-2011: There was an issue, and I had to add support for a diff SfLayout to fix an oversight, which seems like a bug fix to me. The test didn't test the right things.

coder-2011:
Tests do pass on my end, on cuDNN 9.10.2. There was a failure bc of other unrelated issues. 

Claude explicitly said:
>This hits both linear and swizzled cuDNN cases. The PR adds sf_layout= plumbing but never fixes the mat2 construction, so on the very hardware the issue is about, the test errors before the kernel runs.

My tests say the same thing. I can fix this in a follow up, or now if u think thats better.

I appreciate the thorough review, I learned alot! lmk if I made any stupid mistakes, and I will fix them. I would love to contribute to flashinfer alot more in the future

@dhiraj113

flashinfer/gemm/gemm_base.py
dhiraj113: Okay. I guess the test was wrong not the API.

flashinfer/gemm/gemm_base.py
dhiraj113: follow the style of `def mm_fp4`.

coder-2011: fixed that too

PR #3799, opened 2026-07-02
fix(gdn): rewrite sm_120a cluster TMA loads only; keep CuTeDSL's CUBIN when the compiler emits none

### LINKED ISSUE #3795: [Bug] GDN sm_120a prefill fails with cudaErrorUnsupportedPtxVersion on CUDA-13.0 drivers: the TMA patch path reloads PTX through the driver JIT
## Summary

The GDN sm_120a prefill path (`chunk_gated_delta_rule`) fails on any **CUDA 13.0 driver** with

```
cutlass.base_dsl.common.DSLCudaRuntimeError: cudaErrorUnsupportedPtxVersion (error code: 222)
```

at `flashinfer/gdn_kernels/delta_rule_dsl/custom_compile_cache.py` `_load_cuda_library`. On current main (`8fc7f079`), `pytest tests/gdn/test_prefill_delta_rule.py` = **2629 failed / 6 passed / 912 skipped** on an RTX PRO 6000 (SM120) with driver 580.159.03.

## Root cause

The sm_120a TMA patch path is the **only** CuTe-DSL consumer that round-trips through the **driver's PTX JIT**:

1. `_patch_sm120a_tma` extracts the compiled kernel's PTX text and rewrites the cluster-TMA instructions to their `shared::cta` forms (sm_120a has no cluster TMA).
2. `_install_patched_ptx_loader` then loads the **raw patched PTX text** via `cuLibraryLoadData(..., CU_JIT_TARGET=COMPUTE_120A)` — i.e. the *driver* must JIT the PTX.
3. Every `nvidia-cutlass-dsl` 4.5.x build (4.5.0/4.5.1/4.5.2 — the full range `requirements.txt` admits with `>=4.5.0`) bundles **CUDA 13.1** and emits **`.version 9.1`** PTX. A CUDA-13.0 driver (580.xx series) cannot JIT PTX 9.1, hence error 222.

This is why only GDN fails while every other CuTe-DSL surface passes on the same box/driver (b12x fused MoE 142 passed, SM120 MXFP8 grouped GEMM 69 passed, unified MoE 99 passed today): the normal DSL pipeline materializes SASS through its own bundled toolchain and never asks the driver to JIT PTX.

No driver floor is documented — `requirements.txt` only pins `nvidia-cutlass-dsl>=4.5.0` — so a stock torch-2.11-cu130 + 580-driver machine (a common production setup, and driver updates are often not in the user's control on shared clusters) hits this on the very feature the v0.6.14 highlights lead with.

## Evidence (RTX PRO 6000, SM120, CUDA 13.0, driver 580.159.03, torch 2.11.0+cu130)

- Extracted the actual patched PTX from the failing path: header is `.version 9.1` / `.target sm_120a`.
- `nvidia-cutlass-dsl` 4.5.0, 4.5.1, 4.5.2 all report `DSLCudaVersion(major=13, minor=1)` — no 13.0-emitting release satisfies the `>=4.5.0` floor.
- Assembling that same patched PTX with a ptxas >= 13.1 (e.g. the 13.2 ptxas from the pip `nvidia/cu13` toolchain wheel) and loading the **cubin** through the identical `cuLibraryLoadData` call **succeeds on the 13.0 driver**, and the kernels run correctly:

| `tests/gdn/test_prefill_delta_rule.py` | result |
| --- | --- |
| main (`8fc7f079`), driver PTX JIT | 2629 failed / 6 passed / 912 skipped |
| with ptxas-assembly fix, same driver | **2635 passed / 0 failed / 912 skipped** |

## Proposed fix

In `_install_patched_ptx_loader`, assemble the patched PTX to a cubin with the first working `ptxas` (candidates: `FLASHINFER_PTXAS` env override, `PATH`, pip toolchain wheels `nvidia/cu1x/bin`, triton's bundled ptxas, `CUDA_HOME`; the assembly attempt itself is the version check) and hand `cuLibraryLoadData` SASS instead of PTX. Fall back to the current driver-JIT path with a diagnostic warning when no usable ptxas exists. PR incoming.

Found while validating current main on RTX PRO 6000 for the v0.6.14 SM12x release (#3783 follow-up in the proper venue this time). AI-assisted (Claude Code) under human direction; all numbers are from live runs on the real card.

waynehacking8:
## Description

The sm_120a TMA patch path exists because the CUDA-12.9-compiler DSL wheel (`nvidia-cutlass-dsl-libs-base`) emits `shared::cluster` TMA loads that sm_120a does not support - triggering a syscall-emulation path where `setmaxnreg` is ignored (ptxas C7506), register reallocation fails, and the kernel silently runs ~32% slower (guangyunh-nv's NCU measurement in this thread). The cu13 wheel (`nvidia-cutlass-dsl-libs-cu13`) emits no cluster loads; which wheel is active is determined by an install-order file collision (NVIDIA/cutlass#3170, fixed upstream in cutlass-dsl 4.6).

Final shape after review (thread history: driver-JIT retry -> wheel-collision diagnosis -> store-rewrite verification):

- `_patch_sm120a_tma_ptx` performs only the `shared::cluster -> shared::cta` load rewrite - exactly the transformation that removes C7506 and the register spills.
- The former TMA-store cache-hint rewrite is dropped: it does not affect C7506 (guangyunh-nv's isolation), and the PTX reload it forced measured ~9% slower than CuTeDSL's own CUBIN on the cu13 path (826.0 us vs 748.7 us median, 50 CUDA-event reps, B1/H16/S8192/D128).
- When the active compiler emitted no cluster loads (cu13 wheel), the PTX is unchanged and the early-return keeps CuTeDSL's original in-process CUBIN: no PTX reload, no driver PTX-JIT, no version coupling.
- When cluster loads are present (base wheel), the rewritten PTX loads through the driver as before, and a version-downgrade retry (RuntimeWarning on engage) guards any future wheel that both emits cluster loads and stamps a PTX ISA newer than the driver - the original cudaErrorUnsupportedPtxVersion failure mode of #3795.

Once flashinfer pins and verifies cutlass-dsl 4.6, this whole file can be retired.

## Related Issues

Fixes #3795

## Verification (RTX PRO 6000, SM120, CUDA 13.0 driver 580.159.03)

| path | evidence |
| --- | --- |
| cu13 wheel (active `_cutlass_ir` CL-36699951 confirmed) | prefill + CP suites: 2816 passed via the unpatched DSL CUBIN path; ~9% faster than the previous store-rewrite behavior |
| base wheel (payload override per the reproduction in this thread) | pre-patch PTX `.version 8.8` with 6 cluster loads captured; cluster-only rewrite applied; checkpoint tests 17 passed through the patched loader |
| retry guard | engages only on cudaErrorUnsupportedPtxVersion (structured error-code check), warns once per compile; verified earlier in the thread (2635 passed through the retry path when the re-entry was still forced) |

`ruff check` / `ruff format --check` clean.


waynehacking8:
Addressed all three review comments in the follow-up commit: FLASHINFER_PTXAS pointing at a missing file is now warned about and not yielded (the warning previously said "ignoring it" while the generator still yielded it - good catch); site-dir discovery suppresses TypeError and filters falsy entries; the cubin read moved inside the try so an unreadable file falls back to the next candidate instead of raising. Re-verified after the changes: ruff clean, checkpoint tests 17 passed on the affected SM120 / CUDA-13.0-driver box.

waynehacking8:
Second review round swept:

- CodeRabbit's cuda_nvcc finding: valid - added nvidia/cuda_nvcc/bin (the cu12-era pip NVCC wheel layout) to the ptxas candidates alongside the cu13/cu12/triton paths.
- CodeRabbit's S603 noqa nitpick: skipped - this repo's ruff config selects only E/F/B/SIM (pyproject.toml [tool.ruff.lint]), bandit rules are not enabled, so the annotation would be dead weight; ruff check passes as-is.
- gemini-code-assist's three comments reviewed the pre-fix range (up to 41386fc) and are identical to CodeRabbit's first round - all three were already addressed in 2e670c8e (FLASHINFER_PTXAS no longer yielded when missing, TypeError suppressed + falsy site dirs filtered, cubin read moved inside the try).

Re-verified after the change: ruff clean, checkpoint test passes on the affected SM120 / CUDA-13.0-driver machine.

waynehacking8:
Extended verification: the sm_120a context-parallel prefill path (delta_rule_cp_sm120.py) shares the patched loader and is equally affected/fixed. On the same CUDA-13.0-driver box: main fails tests/gdn/test_prefill_cp_delta_rule.py with the identical cudaErrorUnsupportedPtxVersion at the first patch-path test (1 failed / 95 passed with -x; the passing configs are the non-patched ones), while this branch runs the full suite 181 passed / 0 failed. GDN decode (gdn_decode.py) uses plain CUDA JIT kernels, not the DSL patch path, so it is unaffected either way - tests/gdn/test_decode_pretranspose_* pass (14 passed) and the long decode suite is running clean.

flashinfer/gdn_kernels/delta_rule_dsl/custom_compile_cache.py
```diff
@@ -142,6 +142,93 @@ def _patch_sm120a_tma_ptx(ptx_text: str) -> str:
     return patched
 
 
+def _ptxas_candidates():
```
guangyunh-nv: cutedsl itself does not actually call to ptxas directly, what if the system doesn't have a ptxas binary?

guangyunh-nv:
What happen if you just force override the ptx version to 9.0?

flashinfer/gdn_kernels/delta_rule_dsl/custom_compile_cache.py
waynehacking8: Good question - three parts:

1. No ptxas found: _assemble_sm120a_cubin returns None and the loader falls back to exactly the pre-PR behavior (driver PTX JIT), plus a RuntimeWarning naming the driver requirement and the FLASHINFER_PTXAS override. So a system without any ptxas is no worse off than today: it works when the driver is at least as new as the DSL's toolkit, and fails with the same error 222 (now with a diagnostic) when older.

2. In practice a usable ptxas is almost always present: any environment that can JIT flashinfer's own CUDA kernels has a full toolkit (CUDA_HOME/bin/ptxas), and torch, triton, and the pip CUDA toolchain wheels (nvidia/cu13/bin, nvidia/cuda_nvcc/bin) each bundle one. The realistic no-ptxas case is a pure flashinfer-jit-cache/AOT deployment, where the fallback (or a new-enough driver) applies. The candidate does have to be at least as new as the DSL's bundled toolkit - the assembly attempt itself is that check, older ones are skipped naturally (on my box the CUDA_HOME 13.0 ptxas is skipped and the pip cu13 wheel's 13.2 one wins).

3. I also verified the in-process alternative on this box: nvJitLink (cuda.bindings.nvjitlink, already imported ecosystem-wise by this module's dependencies) accepts PTX input and emits a cubin - but the torch-2.11-pinned nvidia-nvjitlink-cu13==13.0.88 rejects the DSL's PTX with exactly the same version wall: 'ptxas application ptx input, line 9; fatal: Unsupported .version 9.1; current version is 9.0'. So it is another candidate with the same version caveat, not a way around it. Happy to put nvjitlink first in the chain (in-process, no subprocess) with the binary ptxas candidates as fallback if you prefer that shape.

Underlying asymmetry, for context: the DSL itself never needs an external ptxas because its runtime materializes SASS for the original PTX in-process at compile time; this patch path re-enters PTX-land after that step, which is what re-couples the load to some assembler or driver version. The deeper fix - applying the TMA rewrite before the DSL's in-process assembly - would live inside cutlass-dsl and is beyond what I can reach from here.

waynehacking8:
@guangyunh-nv tested your suggestion on the affected box (SM120, CUDA 13.0 driver 580.159.03, cutlass-dsl 4.5.2) - it works end to end:

- Rewrote .version 9.1 -> 9.0 inside _patch_sm120a_tma_ptx (one regex on the text we already patch), forced the ptxas path off so the load went through the plain driver PTX JIT with CU_JIT_TARGET=COMPUTE_120A.
- Driver accepts it: library loads, the kernel enumerates, and the full tests/gdn/test_prefill_delta_rule.py suite passes 2635 / 0 failed - so this kernel's PTX uses nothing 9.1-specific and the numerics are identical.

That is strictly simpler than the ptxas machinery in this PR (no external binary, no candidate discovery). If a future DSL emitted genuinely 9.1-only instructions the downgraded load would fail loudly at parse, same visibility as today's error 222.

Two shapes I can rework the PR into - your call:

1. Unconditional: always rewrite the minor version to .0 of the same major in _patch_sm120a_tma_ptx (validated above; smallest diff, ~3 lines).
2. On-demand: attempt the load with the original header first; only on cudaErrorUnsupportedPtxVersion rewrite 9.x -> 9.0 and retry. Zero change for driver-new-enough setups, rescue path only exercised when needed - slightly more code but never touches PTX that already works.

I would drop the ptxas/subprocess code entirely in either case. Happy to push whichever you prefer (my lean is 2).

guangyunh-nv:
@waynehacking8 Exercise option 2. Produce a warning when PTX version overriding happens.

waynehacking8:
@guangyunh-nv done - reworked to option 2 in 5db311a5: the loader tries the original header first and only on cudaErrorUnsupportedPtxVersion rewrites the .version minor to the major's baseline and retries, with a RuntimeWarning when the override happens. The ptxas discovery / subprocess machinery is removed entirely. Re-verified on the affected box (CUDA 13.0 driver, DSL 4.5.2 emitting .version 9.1): prefill suite 2635 passed / 0 failed, CP suite 181 passed / 0 failed, and the warning fires exactly once per compile when the retry engages. Title and description updated to match.

waynehacking8:
Addressed CodeRabbit's nitpick on the retry condition: detection now compares the exception's structured error_code attribute (DSLCudaRuntimeError exposes it; CUresult CUDA_ERROR_UNSUPPORTED_PTX_VERSION and cudaError_t cudaErrorUnsupportedPtxVersion share the numeric value, both referenced symbolically) instead of substring-matching the stringified message; the substring test remains only as a fallback for exception types without a code. Re-verified on the affected box: the retry engages through the structured path, the warning fires once per compile, checkpoint tests 17 passed.

waynehacking8:
One more finding that may change how you want to resolve this, @guangyunh-nv: on the currently supported cutlass-dsl range the sm_120a TMA patch itself appears to be obsolete, and retiring it would remove the PTX re-entry (and with it this whole driver-version coupling) at the root.

Evidence from my SM120 box:

- Emission: I captured the pre-patch PTX from a real sm_120a compile on nvidia-cutlass-dsl 4.5.2 - it contains zero shared::cluster instructions (the cluster-to-cta load rewrite finds nothing to replace anymore). The only lines the patch still touches are the 16 tile-store .L2::cache_hint forms.
- Validity: ptxas 13.2 assembles that unpatched PTX for sm_120a cleanly (rc 0), so the cache-hint store form is accepted for this target by the current toolchain.
- Behavior: with _needs_sm120a_tma_patch forced to False, tests/gdn/test_prefill_delta_rule.py = 2635 passed / 0 failed on DSL 4.5.2, and on DSL 4.5.0 (the requirements.txt floor) the checkpoint test plus the full CP suite = 182 passed / 0 failed. Patch-free, the DSL loads its own compiled artifact in-process, so there is no PTX text reload and no driver PTX-JIT involved at all - the cudaErrorUnsupportedPtxVersion failure class cannot occur.

I assume the patch dates from an older DSL that did emit cluster TMA for sm_120a; with the floor now at >=4.5.0 it looks like dead code, though you would know better whether any configuration outside the gdn test suites still needs it.

Happy to take this PR either way: (a) rework it to retire the patch path entirely (bigger deletion, removes the root cause), or (b) keep the current shape - the version-downgrade retry only ever engages if the patch path is exercised, so it stays as a harmless guard if you prefer to keep the patch for safety. Your call.

guangyunh-nv:
@waynehacking8 The background is with some older versions of anything (find it out, should be `nvidia-cutlass-dsl`) during the kernel development:
```
`ptxas info    : (C7506) Potential Performance Loss: 'setmaxnreg' ignored to maintain compatibility into 'extern' call.`
```
is encountered by extracting PTX and compiling with ptxas. It is caused the TMA instruction with `::cluster` modifier in the PTX. shared::cluster is not supported and results in a syscall emulation, if further causes register reallocation failure which leads to unexpected register spill.

This problem causes performance issue with no error or warning observable at all if compile directly with cutedsl. Only huge perf degeneration is observed. Sweep the dsl package versions to see if problem is gone, since when.

waynehacking8:
Ran the sweep you asked for, @guangyunh-nv - per DSL version, I captured the pre-patch PTX for the sm_120a prefill kernel on this branch's code, counted ::cluster TMA instructions, and assembled the unpatched PTX with ptxas 13.2 -v to catch the C7506 signature and spill stats:

| nvidia-cutlass-dsl | ::cluster lines in PTX | ptxas C7506 warnings | max spill | max regs |
| --- | --- | --- | --- | --- |
| 4.5.0 | 0 | 0 | 8 bytes | 168 |
| 4.5.1 | 0 | 0 | 8 bytes | 168 |
| 4.5.2 | 0 | 0 | 8 bytes | 168 |
| 4.4.2 / 4.4.1 / 4.4.0 | kernel does not compile (cute.nvgpu.OperandMajorMode absent) | - | - | - |

So: gone since at least 4.5.0. Every DSL version the current kernel can compile on emits no ::cluster TMA for sm_120a, ptxas raises no C7506, and the register picture is healthy (8-byte spill, 168 regs - nothing like the reallocation-failure spills you described). The 4.4.x line cannot build today's kernel at all because the cute.nvgpu API it uses did not exist yet, which fits the ::cluster/C7506 era being pre-4.5 development builds. requirements.txt already floors at >=4.5.0, so within the supported range the hazard cannot reproduce.

Together with the earlier behavioral runs (full prefill suite 2635 passed and CP suite 182 passed with the patch disabled, on 4.5.2 and 4.5.0 respectively), both the correctness and the silent-perf failure modes look retired on the supported range. Happy to also A/B the patched-vs-unpatched kernel latency on this card if you want a perf number before deciding retire-vs-keep.

guangyunh-nv:
Thanks for running the sweep. I reran this from PR #3799's exact merge base and got the opposite result: the current SM120 kernel still emits `shared::cluster` TMA loads with `nvidia-cutlass-dsl==4.5.2`, and the original C7506/spill failure is reproducible.

### Revision and method

- PR merge base: `651d877827a825d73bd304e7f54bff0649a69904`
- PR head: `96c3bb86a12f5bf40c4357a55b2b60c2c00bfc07`
- Local `delta_rule_sm120.py` blob: `58acfaed0e6a9daa577f8b4084a873ec673b40b9`
- PR-head `delta_rule_sm120.py` blob: `58acfaed0e6a9daa577f8b4084a873ec673b40b9`

Thus the tested kernel source is byte-identical to the PR head. On a detached worktree at the merge base, I added a diagnostic-only switch to `custom_compile_cache.py` which:

1. dumps the PTX immediately before `_patch_sm120a_tma_ptx`,
2. dumps the PTX immediately after it, and
3. allows the patched loader to be bypassed with `FLASHINFER_GDN_SM120A_TMA_PATCH=0`.

The two generated files are different and reproducible across the enabled/disabled runs:

| PTX | SHA256 | `shared::cluster` TMA loads |
| --- | --- | ---: |
| unpatched | `5f82a7916dc5936e1a659c08b5b106df9b43c68e63f1d42f598bb0ba624885cc` | 6 |
| patched | `238b91a0e130e6e2cbc093ceb577b7c06815e63112f71d98d823da021bef1100` | 0 |

The six unpatched instructions are the exact form handled by the workaround:

```ptx
cp.async.bulk.tensor.3d.shared::cluster.global.tile.mbarrier::complete_tx::bytes.L2::cache_hint ...
```

### `ptxas` result

I assembled both files with:

```bash
ptxas -arch=sm_120a -v <file>.ptx -o <file>.cubin
```

| PTX | C7506 | stack frame | spill stores | spill loads | registers |
| --- | ---: | ---: | ---: | ---: | ---: |
| unpatched | yes | 240 B | 364 B | 656 B | 168 |
| patched | no | 8 B | 8 B | 32 B | 168 |

The unpatched warning is the original failure signature:

```text
ptxas info : (C7506) Potential Performance Loss: 'setmaxnreg' ignored to maintain compatibility into 'extern' call.
```

I also isolated the two PTX transformations. Replacing only the six cluster-load instructions removes C7506 and produces the healthy 8/32-byte spill result. Applying only the TMA-store rewrite leaves C7506 and the 364/656-byte spills unchanged. Therefore the cluster-to-CTA load rewrite is what fixes this failure.

### Performance

I measured the same non-CP SM120 prefill kernel with NCU, using separate processes for patched and unpatched compilation:

- shape: `B=1, H=16, S=8192, D=128`
- fp16 Q/K/V, fp32 alpha/beta/state
- `use_cp=False`
- metric: `gpu__time_duration.sum`
- 20 captured launches per mode
- NCU clock control: `base`

| PTX | mean | median | min | max | standard deviation | NCU launch stack |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| patched | 910.600 us | 910.720 us | 908.896 us | 911.456 us | 0.663 us | 1024 B |
| unpatched | 1202.198 us | 1202.144 us | 1198.400 us | 1206.592 us | 1.997 us | 14848 B |

The patched kernel is **1.320x faster**. Equivalently, the unpatched kernel is **32.0% slower**, while still launching and producing finite outputs. This confirms that the failure remains silent at runtime.

### Environment

- GPU: NVIDIA RTX PRO 6000 Blackwell Server Edition, SM 12.0
- VBIOS: `98.02.AF.00.01`
- Driver: `595.58.03`
- CUDA Toolkit: `13.1`, build `V13.1.115`
- `ptxas`: CUDA 13.1, build `V13.1.115`
- Nsight Compute CLI: `2025.4.1.0`
- `nvidia-cutlass-dsl`: `4.5.2`
- PyTorch: `2.9.1+cu130`
- `cuda-bindings`: `13.3.1`
- Python: `3.13.14`
- OS: Ubuntu 24.04.3 LTS, Linux 6.14.0-29-generic

The previously reported combination of "0 cluster lines" and "8-byte spill" exactly matches the **patched** PTX in this reproduction, not the pre-patch PTX. Could you share the PTX-capture script or exact commands used for the sweep? I suspect the captured artifact had already passed through `_patch_sm120a_tma_ptx`.

waynehacking8:
Thank you for the rigorous reproduction - your result stands and mine was measured on a different axis than I claimed. Retracting the retire-the-patch suggestion: with ::cluster emission reproducible in your environment and your isolation showing the cluster-to-CTA load rewrite is exactly what removes C7506 and the 32% silent slowdown, the patch clearly has to stay.

What I can now say precisely about the divergence:

- Same source: my capture tree is detached at 651d877 - your exact merge base - so every file is identical, not just the delta_rule_sm120.py blob.
- Capture method (script: https://gist.github.com/waynehacking8/ec70399f3b61b5928f96bae97708e306): I intercept the text argument of _patch_sm120a_tma_ptx itself, i.e. the same point as your dump (1). The captured pre-patch PTX on my box is .version 9.1 / .target sm_120a, contains the 16 tile-store .L2::cache_hint lines and zero shared::cluster loads, sha256 b464341ae4d1fd288643d1ae659970e0e6c42002d5d78b78da92d5c4eed6610a - so it is genuinely pre-patch (your store-rewrite markers are present), just without the cluster loads yours has.
- Within my environment the result is stable across everything I can vary: shapes (T=8192/H=16 per your NCU config, down to T=128/H=4 - byte-identical PTX, the kernel is shape-generic), nvidia-cutlass-dsl 4.5.0/4.5.1/4.5.2, and cuda-bindings 13.0.3 vs 13.3.1 (upgraded to match yours - no change).

So the ::cluster emission is conditional on something in the environment rather than the DSL version alone, and my sweep answered the wrong axis. Remaining deltas between our setups: driver 595.58.03 vs 580.159.03, host toolkit 13.1 vs 13.0, torch 2.9.1 vs 2.11.0, Python 3.13 vs 3.10, and Server vs Max-Q Workstation SKU. My money is on the driver: if the DSL's TMA lowering keys on a driver-reported device capability, a 595-era driver could legitimately unlock the cluster form that 580 does not. I cannot upgrade the driver on this box (shared, 14 users) to confirm, but if you can dump the DSL's device/driver attribute view in your environment (or point me at where the lowering branches), I am glad to chase it from the other side.

Implication for this PR either way: since the patch stays, the PTX re-entry stays, and the retry-with-downgraded-header remains the fix for the driver-version coupling - which my environment does hit and where the full prefill and CP suites pass through the retry path (2635 and 181 passed). The earlier obsolescence comment should be read as environment-specific; apologies for the overreach.

guangyunh-nv:
I found the remaining environment axis. Your capture gist is correct; the divergence is not caused by the driver. It is caused by which CUTLASS DSL compiler wheel owns the overlapping installed files.

This is the same packaging problem reported in [NVIDIA/cutlass#3170](https://github.com/NVIDIA/cutlass/issues/3170): `nvidia-cutlass-dsl-libs-base` and `nvidia-cutlass-dsl-libs-cu13` install divergent binaries and Python files to the same paths. The distribution metadata can show both packages as installed, while installation order determines which `_cutlass_ir` compiler is actually active.

I reproduced both results on the same node, without changing the source, GPU, driver, host toolkit, Torch, or Python. I only selected a different 4.5.2 wheel payload through `PYTHONPATH`/`LD_LIBRARY_PATH`:

| Active 4.5.2 wheel | Embedded compiler | PTX header | `shared::cluster` loads | PTX SHA256 |
| --- | --- | --- | ---: | --- |
| `nvidia-cutlass-dsl-libs-base` | CUDA 12.9 / NVVM 20, CL-36006120 | `.version 8.8` | 6 | `5f82a7916dc5936e1a659c08b5b106df9b43c68e63f1d42f598bb0ba624885cc` |
| `nvidia-cutlass-dsl-libs-cu13` | CUDA 13.1 / NVVM 21, CL-36699951 | `.version 9.1` | 0 | `b464341ae4d1fd288643d1ae659970e0e6c42002d5d78b78da92d5c4eed6610a` |

The cu13 result is byte-for-byte your reported pre-patch PTX hash. It was generated here with driver `595.58.03`, so the `580` versus `595` driver difference is ruled out. The PTX compiler version comes from the wheel, not the host toolkit.

### Reproduction

Starting from the PR merge base and using your `capture_prepatch_ptx.py` gist:

```bash
rm -rf /tmp/dsl-wheels /tmp/dsl-base /tmp/dsl-cu13
mkdir -p /tmp/dsl-wheels /tmp/dsl-base /tmp/dsl-cu13

python -m pip download --no-deps -d /tmp/dsl-wheels \
  nvidia-cutlass-dsl-libs-base==4.5.2 \
  nvidia-cutlass-dsl-libs-cu13==4.5.2

unzip -q /tmp/dsl-wheels/nvidia_cutlass_dsl_libs_base-4.5.2-*.whl \
  -d /tmp/dsl-base
unzip -q /tmp/dsl-wheels/nvidia_cutlass_dsl_libs_cu13-4.5.2-*.whl \
  -d /tmp/dsl-cu13

for variant in base cu13; do
  root=/tmp/dsl-${variant}/nvidia_cutlass_dsl
  PYTHONPATH=${root}/python_packages:. \
  LD_LIBRARY_PATH=${root}/lib:${LD_LIBRARY_PATH} \
  CAP_OUT=/tmp/${variant}.ptx \
    python capture_prepatch_ptx.py

  sed -n '1,10p' /tmp/${variant}.ptx
  grep -c 'shared::cluster' /tmp/${variant}.ptx || true
  sha256sum /tmp/${variant}.ptx
done
```

The active compiler can also be identified directly:

```bash
P=$(python - <<'PY'
import cutlass._mlir._mlir_libs._cutlass_ir as m
print(m.__file__)
PY
)
sha256sum "$P"
strings "$P" | grep -E 'CL-[0-9]+|Cuda compilation tools, release|NVVM [0-9]+\.' | sort -u
```

Therefore the earlier DSL-version sweep held the cu13 wheel variant constant. It established that `libs-cu13` 4.5.0-4.5.2 does not emit cluster loads, but it did not test the corresponding `libs-base` compiler. A fresh plain `pip install nvidia-cutlass-dsl` selects the base dependency and can reproduce the cluster/C7506 path; `pip install 'nvidia-cutlass-dsl[cu13]'` selects the cu13 overlay, subject to the install-order collision described in #3170.

The CUTLASS maintainer reports that the overlapping-wheel issue is fixed in [the 4.6 release](https://github.com/NVIDIA/cutlass/issues/3170#issuecomment-4862661573). Until FlashInfer requires and verifies 4.6, the cluster-load workaround must remain for the base-wheel path.

This also sharpens the question for this PR: with the cu13 wheel, there are no cluster loads to rewrite. Only the 16 TMA-store forms currently change, which forces PTX re-entry and exposes the PTX 9.1/driver compatibility issue. Since the store-only rewrite does not affect C7506, we should separately verify whether it is still needed. If it can be removed, the cu13 path would keep CuTeDSL's original CUBIN and avoid the PTX-version retry entirely, while the base path would still apply the necessary cluster-to-CTA rewrite.

guangyunh-nv:
@aleozlx Can we bump flashinfer pinned cutlass package version to avoid this type of issue, so that the patch can be fully avoided and removed.

waynehacking8:
Excellent detective work on the wheel collision - that closes the divergence completely. Confirmed on my side with your identification command: my active _cutlass_ir is CL-36699951 (CUDA 13.1 compiler, the cu13 payload), which is why every capture here produced the 9.1/no-cluster PTX with exactly the hash you matched. And using your payload-override reproduction on this same box, the base wheel produces .version 8.8 with 6 cluster loads - both results now reproduce on one machine on both our setups.

I also ran the store-rewrite verification you asked for, on the cu13 path (B=1, H=16, S=8192, D=128, fp16, 50 CUDA-event reps, medians): store-rewrite-and-reload 826.0 us vs untouched DSL CUBIN 748.7 us - the store rewrite plus forced PTX reload costs about 9%, with byte-identical output signatures. So it is not just unnecessary, it is a small regression.

Pushed the rework in 1594d621 accordingly, matching the shape you outlined:

- _patch_sm120a_tma_ptx now performs only the shared::cluster -> shared::cta load rewrite. The store cache-hint rewrite is gone.
- When the compiler emitted no cluster loads (cu13 wheel), the text is unchanged and the existing early-return keeps CuTeDSL's original in-process CUBIN: no PTX reload, no driver JIT, retry never engages.
- When cluster loads are present (base wheel), the rewrite applies and the patched PTX (.version 8.8 there) loads through the driver as before; the version-downgrade retry stays as a small guard for any future wheel that both emits cluster loads and stamps a PTX ISA newer than the driver.

Verified both paths on this card: cu13 wheel = prefill + CP suites 2816 passed through the unpatched CUBIN path; base wheel (payload override) = 6 cluster loads captured pre-patch, checkpoint tests 17 passed through the cluster-rewritten loader. And agreed on the endgame: once flashinfer pins and verifies cutlass-dsl 4.6 per your ask to aleozlx, the whole file can go.

guangyunh-nv:
Superseded with #3844

waynehacking8:
Makes sense - the pin-side fix in #3844 is the cleaner resolution, thanks for closing the loop.

PR #3801, opened 2026-07-02
fix(cudnn): use a full-sequence Q batch stride in batch prefill (silent corruption for batch > 0 without ragged offsets)

### LINKED ISSUE #3800: [Bug] cudnn_batch_prefill_with_kv_cache: one-token Q batch stride silently corrupts every batch after the first when batch_offsets_q is not passed
### Summary

`cudnn_batch_prefill_with_kv_cache` returns **silently wrong attention output for every batch after the first** whenever the caller does not pass `batch_offsets_q` - which is exactly how flashinfer's own trace suite calls it (`tests/trace/test_cudnn_batch_prefill_reference_correctness.py`, red on current main for its `B2` case).

### Root cause

The Q tensor descriptor in `flashinfer/cudnn/prefill.py` uses a **one-token batch stride**:

```python
cudnn_q = g.tensor(
    name="q",
    dim=(graph_b, h_qo, graph_s_qo, d_qk),
    stride=(h_qo * d_qk, h_stride, s_stride, d_stride),  # batch stride = ONE token
    ...
)
```

so batch `b`'s queries are read starting at token `b` instead of token `b * s_qo`. Batch 0 is coincidentally correct; every later batch computes attention over the wrong query rows and writes finite, plausible-looking garbage. The O descriptor right below does it correctly (`graph_s_qo * d_vo * h_qo`), which is what exposes this as an oversight rather than a convention.

The public wrapper path (`BatchPrefillWithPagedKVCacheWrapper(backend="cudnn")`) always supplies ragged offsets, and `set_ragged_offset` overrides the batch stride - which is why `tests/attention/test_cudnn_prefill.py` passes while the raw-function path is broken.

### Evidence (RTX PRO 6000, SM120, CUDA 13; reproduces identically with cuDNN backend 9.19.0.56 and 9.23.2.1, cudart 12 and 13)

- Failing config `B=2, Hq=8, Hk=2, D=128, PS=16, q_len=32, kv_len=64` (the trace test): per-batch max abs diff vs reference = `[0.0078, 1.7183]` - batch 0 perfect, batch 1 100% of tokens wrong from token 0.
- Independent arbitration: a hand-written torch reference (per-page gather + causal softmax) agrees with the trace template's reference on both batches (max 0.0039) and disagrees with the kernel on batch 1 - the references are right, the kernel path is wrong.
- Scales with batch: `B=3` gives `[0.0078, 1.7183, 1.5195]`. Not GQA-specific: `Hq=Hk=8` gives `[0.0039, 1.6133]`.
- Passing `batch_offsets_q`/`batch_offsets_o` heals it: `[0.0078, 0.0039]` - confirming the ragged-offset override is what protects the wrapper path.

### Fix

Use a full-sequence batch stride, mirroring the O descriptor: `stride=(graph_s_qo * s_stride, h_stride, s_stride, d_stride)` (using the tensor's own per-token stride also keeps non-contiguous `q` correct). With that one-line change on this machine: the trace file goes 1 failed / 1 passed -> 2 passed, `tests/attention/test_cudnn_prefill.py` + `test_cudnn_prefill_deepseek.py` stay green (384 passed), and the B3/GQA/MHA probes above all drop to <= 0.0078. PR incoming.

Also noting while reading the same function: the `q.dim() == 4` branch unpacks three values from a 4-D `q.stride()`, which raises ValueError if ever taken - the 4-D path looks untested.

### Environment

RTX PRO 6000 Blackwell (SM120), CUDA 13.0, torch 2.11 cu130, flashinfer main 8fc7f079. Found during a systematic SM120 sweep of the trace suites; arbitration and fix verified on the same card.

waynehacking8:
## Description

`cudnn_batch_prefill_with_kv_cache` builds its Q tensor descriptor with a one-token batch stride (`h_qo * d_qk`), so whenever the caller does not pass `batch_offsets_q`, batch `b`'s queries are read starting at token `b` instead of token `b * s_qo`. Batch 0 is coincidentally correct; every later batch silently computes attention over the wrong query rows and returns finite, plausible-looking garbage. Full analysis, per-batch error maps, and independent-reference arbitration in #3800.

The public wrapper path (`BatchPrefillWithPagedKVCacheWrapper(backend="cudnn")`) always supplies ragged offsets, and `set_ragged_offset` overrides the batch stride - which is why the wrapper test suite passes while flashinfer's own trace suite (`tests/trace/test_cudnn_batch_prefill_reference_correctness.py`), which calls the function directly, is red on main for its `B2` case.

Fix: use `graph_s_qo * s_stride` (a full sequence, expressed via the tensor's own per-token stride so non-contiguous `q` stays correct), mirroring the O descriptor a few lines below which already does this correctly.

## Related Issues

Fixes #3800

## Verification (RTX PRO 6000, SM120, CUDA 13.0; identical results with cuDNN backend 9.19.0.56 and 9.23.2.1)

- `tests/trace/test_cudnn_batch_prefill_reference_correctness.py`: 1 failed / 1 passed -> **2 passed**.
- `tests/attention/test_cudnn_prefill.py` + `test_cudnn_prefill_deepseek.py`: **384 passed** before and after - the ragged-offset wrapper path is unaffected by construction.
- Direct-call probes without offsets: B=2 and B=3, GQA (Hq8/Hk2) and MHA (Hq8/Hk8) - every batch now matches an independently hand-written torch reference (per-page gather + causal softmax), max abs diff <= 0.0078, where batch 1 was 1.72 before the fix.

`ruff check` / `ruff format --check` clean.

## Pull Request Checklist

- [x] Pre-commit checks (ruff clean on the touched file)
- [x] Tests: the previously-red trace test now passes and covers this path; wrapper suites re-run green

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected query tensor batching in cuDNN SDPA graph construction, improving indexing behavior when ragged query offsets are not provided.
  * Fixed the batch stride used for query tensors so each batch now maps to the full query sequence layout as expected.

### DIFF
--- a/flashinfer/cudnn/prefill.py
+++ b/flashinfer/cudnn/prefill.py
@@ -196,7 +196,13 @@ if CUDNN_AVAILABLE:
             cudnn_q = g.tensor(
                 name="q",
                 dim=(graph_b, h_qo, graph_s_qo, d_qk),
-                stride=(h_qo * d_qk, h_stride, s_stride, d_stride),
+                # The batch stride must cover a full sequence of tokens
+                # (mirroring the O descriptor below). With the previous
+                # one-token stride (h_qo * d_qk), every batch after the first
+                # read its queries from the wrong rows whenever no
+                # batch_offsets_q ragged offset was supplied - silently wrong
+                # outputs for batch_size > 1.
+                stride=(graph_s_qo * s_stride, h_stride, s_stride, d_stride),
                 data_type=cudnn_q_data_type,
             )
 


YangXu1990uiuc:
Thanks @waynehacking8 — and thank you especially for #3800: the per-batch error maps and the independent-reference arbitration made the root cause unambiguous, and your diagnosis is exactly right — with no ragged offset attached, the one-token batch stride reads batch b's queries starting at token b. That analysis directly informed how this got fixed.

We ended up merging a different fix (#3784, plus #3921 as a companion), and you deserve an explanation of why rather than just a closed PR.

The core difficulty is that q is documented as a packed tensor (total_qo_tokens, h, d). For a packed ragged buffer, batch b starts at qo_indptr[b] — which no constant batch stride can express; that's precisely what cuDNN's ragged offsets exist for. The full-sequence stride in this PR is the right descriptor for a padded buffer with b·s_max rows (which is what the uniform-length repro allocates), but for a genuinely ragged packed input it still reads the wrong rows, and once total_tokens < b·s_max it reads past the end of the packed allocation. Since a padded buffer and a packed uniform-length buffer are indistinguishable from shapes alone, any stride choice just trades one silent corruption for another. So we opted to make the unaddressable case loud instead: #3784 raises a ValueError for batch_size > 1 without offsets, and #3921 drops the cost of supplying them to near zero — batch_offsets_units="tokens" accepts the plain qo_indptr/kv_indptr every serving framework already has, and actual_seq_lens_* can then be omitted entirely.

If you have a real use case where a padded, no-offsets Q is the natural calling convention, that would make a good feature discussion — as an explicit layout flag rather than stride inference — and we'd be glad to review it.

I want to highlight the quality of this report: live-hardware verification across two cuDNN backend versions, an independently hand-written reference, and disclosed methodology — this is exactly the kind of contribution this repo needs. Please keep them coming; the cuDNN path has historically been under-tested and careful eyes like yours are how it gets better!

waynehacking8:
@YangXu1990uiuc Thanks for the detailed explanation and the graceful handling of this PR!

Your architectural reasoning makes total sense. Raising a ValueError (#3784) for the unaddressable packed case is definitely the right engineering call to prevent trading one silent bug for another. I also really like the ergonomics improvement in #3921—supporting qo_indptr directly with near-zero overhead is a much cleaner way to handle this across serving frameworks.

I'm glad the per-batch error maps and reference probes in #3800 were helpful in pinpointing the root cause. I don't have a strong blocker for a dedicated "Padded Q" layout flag right now, as the new token-unit offsets cover my use case well.

Thanks again for the kind words and the great work on maintaining FlashInfer. I’ll go ahead and close this PR, and I look forward to contributing more to the cuDNN paths in the future!

PR #3805, opened 2026-07-02
Return selected backend from GEMM dispatchers

askliar:
## What changed

- Make `bf16_gemm_sm100`, `fp8_gemm_sm100`, and `mxfp8_gemm_sm100` return the concrete backend name corresponding to the runner selected by `AutoTuner.choose_one()`.
- Keep a backend-name list aligned with each dispatcher runner list and return the name at the selected runner index.
- Leave all public high-level GEMM signatures and tensor return values unchanged.

## Why

The selected runner already appears in logs, but callers of the lower GEMM dispatchers cannot read it programmatically. Returning the stable backend name consistently across BF16, FP8, and MXFP8 provides that information without broad public API changes.

## Validation

- `pre-commit run --files flashinfer/gemm/gemm_base.py`
- `pytest -q tests/autotuner/test_autotuner_core.py` (`86 passed`)

## Summary by CodeRabbit

* **New Features**
  * GEMM operations now return the selected backend name at runtime, providing clearer visibility into which execution path was chosen by the auto-tuner.
  * Updated the SM100 GEMM dispatch helpers for BF16, FP8, and MXFP8 to report the chosen backend after selection.

### DIFF
--- a/flashinfer/gemm/gemm_base.py
+++ b/flashinfer/gemm/gemm_base.py
@@ -1302,7 +1302,7 @@ def bf16_gemm_sm100(
     out: torch.Tensor,
     workspace_buffer: torch.Tensor,
     runner_names: List[str],
-) -> None:
+) -> str:
     use_sm_100f = is_sm100f_supported(a.device)
 
     tuner = AutoTuner.get()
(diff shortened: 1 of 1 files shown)


flashinfer/gemm/gemm_base.py
```diff
@@ -1313,23 +1313,29 @@ def bf16_gemm_sm100(
     is_b_k_major = b.stride(-2) == 1
 
     runners = []
+    backend_names = []
```
qiching, line 1424: `backend_names` and `runners` are two parallel lists kept in sync by hand, they are in the same order. 

 Consider appending them together so they can't drift:
```
runners, backend_names = [], []
def _add(name, runner):
    runners.append(runner)
    backend_names.append(name)
if "cudnn" in runner_names:
    _add("cudnn", _cudnn_gemm_bf16_runner(is_a_k_major=is_a_k_major, is_b_k_major=is_b_k_major))
if "cublaslt" in runner_names:
    _add("cublaslt", get_mm_bf16_cublaslt_module().cublaslt_bf16_gemm_runner())
```

Same applies to `fp8_gemm_sm100` and `mxfp8_gemm_sm100`. Otherwise looks correct.

qiching (changes requested):

qiching (approved):
LGTM now

PR #2140, opened 2025-11-24
Use global TuningConfig, to fix memory leak caused by AutoTuner LRU cache and dynamic lambda TuningConfig

juju812:
## 📌 Description

This PR is to fix a memory leak bug caused by AutoTuner LRU cache and dynamic lambda TuningConfig

## 🔍 Related Issues

https://github.com/flashinfer-ai/flashinfer/issues/2139

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [x] Tests have been added or updated as needed.
- [x] All tests are passing (`unittest`, etc.).

## Summary by CodeRabbit

* **Performance**
  * Reduced autotuner overhead by caching runner parameter names to avoid repeated signature inspection during profiling, speeding up tuning runs.
* **New Features**
  * Centralized reusable tuning presets for mixed-precision GEMM (FP8/FP4) with additional tuning presets to improve autotuning and execution efficiency.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

Issue #2139, opened 2025-11-24
NVFP4 CPU RAM leak caused by dynamic creation of TuningConfig objects


juju812:
## 1. Problem Description

When frameworks like vLLM invoke FlashInfer's `mm_fp4` (backend="cutlass") or `fp8_gemm_sm100` interfaces, a continuous increase in memory usage is observed, potentially leading to OOM (Out Of Memory) errors.

Screenshot below is based on version `0.3.1`

![image](https://s21.ax1x.com/2025/11/24/pZApNkt.png)

## 2. Root Cause Analysis

The root cause of the memory leak lies in the incompatibility between the `AutoTuner` caching mechanism and the dynamic creation of `TuningConfig` objects.

### 2.1 Call Chain

1.  **`AutoTuner` Caching Mechanism**:
    The `flashinfer.autotuner.AutoTuner` class has a method `_find_nearest_profile` decorated with `@lru_cache(maxsize=None)`:
    ```python
    @classmethod
    @lru_cache(maxsize=None)
    def _find_nearest_profile(
        cls, shapes: Tuple[torch.Size], tuning_config: TuningConfig
    ) -> Tuple:
    ```
    This method uses `(shapes, tuning_config)` as the cache key.

2. **Dynamic Creation of `TuningConfig`**:
    In the current code, `TuningConfig` objects are created anew inside the `mm_fp4` and `fp8_gemm_sm100` functions every time they are called:
    ```python
    # Example code
    def mm_fp4(...):
        # ...
        tuning_config = TuningConfig(
            dynamic_tensor_specs=(
                DynamicTensorSpec(...),
            ),
            constraint_specs=(
                ConstraintSpec(..., lambda shapes: ...), # <--- New lambda created each time
            ),
        )
        runner, tactic = tuner.choose_one(
            "fp4_gemm",
            runners,
            tuning_config,
            inputs,
        )
    ```

3. **Inconsistent Hash Values**:
    *   `TuningConfig` contains `ConstraintSpec`.
    *   `ConstraintSpec` contains a `lambda` function.
    *   In Python, executing a `lambda` expression creates a new function object each time, even if the code logic is identical. These function objects have different memory addresses (and thus different hash values).
    *   Consequently, the `TuningConfig` object generated in each call to `mm_fp4` is considered "unequal".

4.  **Unbounded Cache Growth**:
[...]

PR #3810, opened 2026-07-02
fix(fused_moe): FP4 MoE autotuner LRU cache leak

rjzhb:
## 📌 Description

The trtllm-gen fused-MoE ops (`trtllm_bf16_moe`, `trtllm_fp8_per_tensor_scale_moe`,
`trtllm_fp8_block_scale_moe`, `trtllm_fp4_block_scale_moe`,
`trtllm_mxint4_block_scale_moe`) call `MoERunner._make_tuning_config()` on **every**
forward. That method built the profiling-input initializers and the
`num_tokens -> bucket` mapper as **fresh closures each call**, so every forward
produced a brand-new `TuningConfig`.

`AutoTuner.choose_one()` -> `search_cache()` -> `_get_cache_key()` keys the
`@lru_cache(maxsize=None)` on `_find_nearest_profile` by that `TuningConfig`, and
`DynamicTensorSpec` hashes/compares `map_to_tuning_buckets` and
`tensor_initializers` **by object identity**. A fresh closure per forward therefore
mints a new, never-evicted cache entry on every forward -> unbounded host-memory
growth. 

This is effectively a regression of the class of bug fixed for the GEMM path in
#2140: the later MoE autotuning refactor (#2942 / #3115) reintroduced per-call
dynamic-lambda `TuningConfig`s on the MoE path.

## 🔧 Fix

Define the MoE profiling-input initializers once at module scope (`_MOE_INITS`),
and memoize the two that depend on run-time integers:
- `_moe_topk_ids_init(num_experts)`
- `_moe_map_bucket(tune_max_num_tokens)`

`_make_tuning_config()` now reuses these stable objects. Because the buckets are
already a value-hashed tuple and the mapper/initializers now have stable identity,
repeated forwards yield `TuningConfig`s that are hash- and eq-equal, so
`_find_nearest_profile`'s cache holds a single entry per distinct op/config instead
of one per forward. No behavioral change to the synthesized profiling tensors.

## 🔍 Related Issues

https://github.com/flashinfer-ai/flashinfer/issues/2139

## 🧪 Tests

Verified with the real `DynamicTensorSpec`/`TuningConfig` and the new helpers:
- before: two `_make_tuning_config`-style configs are hash≠ / eq≠ (distinct cache keys);
- after: hash= / eq= (same key);
- 100 simulated forwards execute `_find_nearest_profile`'s body once (was 100×).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the stability of MoE autotuning by reusing consistent input setup across runs.
  * Reduced unnecessary re-creation of internal tuning inputs, which can help make forward passes and tuning behavior more reliable.

### DIFF
--- a/flashinfer/fused_moe/core.py
+++ b/flashinfer/fused_moe/core.py
@@ -79,6 +79,76 @@ from .utils import (
 )
 
 
+# Module-level initializers for the autotuner's synthetic MoE profiling inputs.
+#
+# The AutoTuner keys its ``_find_nearest_profile`` LRU cache on the
+# ``TuningConfig`` passed to ``choose_one`` on every MoE forward, and a
+# ``DynamicTensorSpec`` hashes/compares its ``map_to_tuning_buckets`` and
+# ``tensor_initializers`` by object identity.  Building those closures fresh
+# inside ``_make_tuning_config`` on each call therefore mints a new cache key
+# every forward, so the (unbounded) cache grows without limit -- a host-side
+# memory leak first observed on MiniMax-M3.  Defining the initializers once at
+# module scope (and memoizing the two that depend on run-time integers) keeps
+# their identity stable across forwards so repeated calls reuse a single cache
+# entry.  See https://github.com/flashinfer-ai/flashinfer/issues/2139.
+_MOE_INITS = {
+    "output": lambda shapes, dtype, device: torch.empty(
+        shapes, dtype=dtype, device=device
+    ),
+    "hidden_states": lambda shapes, dtype, device: torch.randn(
+        shapes, device=device
+    ).to(dtype),
+    "routing_logits": lambda shapes, dtype, device: torch.rand(
+        shapes, dtype=dtype, device=device
+    ),
+    "expert_weights": lambda shapes, dtype, device: torch.ones(
+        shapes, dtype=dtype, device=device
+    ),
+    "hidden_states_scale": lambda shapes, dtype, device: torch.ones(
+        shapes, device=device
+    ).to(dtype),
+    "gemm1_lora_delta": lambda shapes, dtype, device: torch.zeros(
+        shapes, dtype=dtype, device=device
+    ),
+    "per_token_scale": lambda shapes, dtype, device: torch.ones(
+        shapes, device=device
+    ).to(dtype),
+}
+
+
+@functools.lru_cache(maxsize=32)
+def _moe_topk_ids_init(num_experts: int):
+    """Return a stable packed-topk-ids initializer for a given expert count.
+
+    Memoized on ``num_experts`` so the returned closure keeps a stable identity
+    across forwards; see the ``_MOE_INITS`` note above.
+    """
+
+    def _init(shapes, dtype, device):
+        expert_ids = make_random_topk_ids(
+            num_experts=num_experts,
+            num_tokens=math.prod(shapes[:-1]),
+            top_k=shapes[-1],
+            device=device,
+        ).view(shapes)
+        expert_weights = torch.ones(
+            shapes, dtype=torch.bfloat16, device=device
+        ).view(torch.int16)
+        return (expert_ids << 16) | expert_weights
+
+    return _init
+
+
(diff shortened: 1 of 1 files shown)


rjzhb:
The bug proposed by this PR affects all FP4 MoE kernels, including those used by the latest open-source models such as MiniMax.

junliu-mde:
+1 — we hit this exact leak in production (SGLang + trtllm-gen MoE, flashinfer 0.6.12, eager prefill) and independently traced it to the same chain: per-forward closures in `_make_tuning_config` → `DynamicTensorSpec.__hash__` on callable `id()` → `_find_nearest_profile`'s unbounded `lru_cache` never hits and retains every config. Memory dumps showed the leaked objects are exactly the lambdas/`_init_packed_topk_ids` this PR stabilizes.

Can also confirm bounding `_find_nearest_profile`'s cache eliminates the growth with no observable behavior change. Hope it lands soon.

image

aleozlx:
the purpose sounds similar to #3687 

could you help review @jdebache

cc @qiching

vitamin-chaos:
I think #3810 is still valuable because it fixes the known MoE path at the source by stabilizing all initializer and mapper identities, including _init_packed_topk_ids. #3687 bounds the worst-case cache growth, while #3810 prevents repeated cache misses and allocation churn. The two fixes are complementary.

jdebache:
#3687 should also be stabilizing initializer and mapper identities. Just with a different approach. I don't have a strong preference for one way or the other.

vitamin-chaos:
> I think #3810 is still valuable because it fixes the known MoE path at the source by stabilizing all initializer and mapper identities, including _init_packed_topk_ids. #3687 bounds the worst-case cache growth, while #3810 prevents repeated cache misses and allocation churn. The two fixes are complementary.

Sorry for not being clear earlier. I think we should keep _moe_topk_ids_init(num_experts) from #3810 and use it instead of the local _init_packed_topk_ids. #3687 does not fully fix this path because _init_packed_topk_ids is still created as a fresh local closure on every _make_tuning_config() call. DynamicTensorSpec.__hash__ does not include tensor_initializers, but the dataclass-generated __eq__ still compares them. Therefore, repeated configs have the same hash but are not equal, and _find_nearest_profile continues to miss. But I haven't verified this specifically. Please point it out if it's incorrect. If the above conclusion is correct, I think #3687 could incorporate this change from #3810, and then we could close #3810.

rjzhb:
> > I think #3810 is still valuable because it fixes the known MoE path at the source by stabilizing all initializer and mapper identities, including _init_packed_topk_ids. #3687 bounds the worst-case cache growth, while #3810 prevents repeated cache misses and allocation churn. The two fixes are complementary.
> 
> Sorry for not being clear earlier. I think we should keep _moe_topk_ids_init(num_experts) from #3810 and use it instead of the local _init_packed_topk_ids. #3687 does not fully fix this path because _init_packed_topk_ids is still created as a fresh local closure on every _make_tuning_config() call. DynamicTensorSpec.**hash** does not include tensor_initializers, but the dataclass-generated **eq** still compares them. Therefore, repeated configs have the same hash but are not equal, and _find_nearest_profile continues to miss. But I haven't verified this specifically. Please point it out if it's incorrect. If the above conclusion is correct, I think #3687 could incorporate this change from #3810, and then we could close #3810.

You're right — and I'd actually verified this before opening the PR. I measured the _find_nearest_profile miss rate before/after: with the per-call local closure it misses on essentially every call, whereas after switching spec["topk_ids"] to the memoized _moe_topk_ids_init(num_experts) (stable identity) the hit rate is high — one cache entry per distinct shape instead of one per forward.

jdebache:
Ohhhh, I see now that I missed one of the initializers. In my mind they were all handled 🤦. I also figured out why my test was not catching it. I will submit a follow-up PR. Thanks for pointing this out and apologies for the confusion.

aleozlx:
pls do not overwrite default PR template.

https://github.com/flashinfer-ai/flashinfer/blob/main/.github/pull_request_template.md

PR #3817, opened 2026-07-03
fix(sparse-mla-sm120): add TOPK=256 decode-dsv4 instantiation for DeepSeek-V4 DSpark

vedcsolution:
## Problem

Serving `deepseek-ai/DeepSeek-V4-Flash-DSpark` (DeepSeek-V4-Flash with DSpark speculative decoding) on SM120 (GB10 / DGX Spark, `sm_121a`) crashes at engine warmup with a hard C++ abort:

```
tvm.error.InternalError: Check failed: num_tokens > 64 (5 vs. 64)
  : Decode (num_tokens <= 64) must go through sparse_mla_sm120_decode_dsv3_2
    or sparse_mla_sm120_decode_dsv4; got num_tokens=5
  at .../csrc/sparse_mla_sm120.cu
```

## Root cause

The DSpark draft model decodes with `(num_heads=32, topk=256, d_qk=512, page_block_size=64)` (`num_tokens = num_speculative_tokens = 5`). `topk=256` is absent from both `_DECODE_DSV4_DISPATCH` (Python) and the `DSV4_DISPATCH` instantiation switch (CUDA), which only cover `TOPK ∈ {128, 512, 1024}`.

In `_paged_attention`, a `num_tokens <= _DECODE_MAX_TOKENS` shape that matches no decode table falls through to `sparse_mla_sm120_paged_attention` (the prefill orchestrator), which hard-asserts `num_tokens > 64` and aborts the process. The main-model decode (`topk=512`) is covered, so this only manifests once speculative decoding is enabled.

## Fix

- Add the `TOPK=256` column to the DSV4 decode dispatch — Python `_DECODE_DSV4_DISPATCH` and the CUDA `DSV4_DISPATCH` macros — for `NUM_HEADS ∈ {8, 16, 32, 64, 128}`. Shared memory and launch bounds in the decode-dsv4 kernel are driven by `DSV4_BI` (=64) and `HPB`, not `TOPK`, so `topk=256` runs within the same resource envelope as the existing `128/512/1024` instantiations (`256 = 4·64`, `num_splits = 4`).
- Guard the decode path: when `num_tokens <= _DECODE_MAX_TOKENS` and no decode table matches, raise a clear `ValueError` naming the shape instead of falling through to the paged kernel's C++ abort.
- Extend `test_sparse_mla_sm120_decode_dsv4` with `(8, 256)`, `(16, 256)`, and `(32, 256)` (compared against the reference implementation on-device).

## Scope / compatibility

- Confined to `sparse_mla_sm120` (Blackwell `sm_120`/`sm_121`). Other architectures neither compile nor load this module.
- Purely additive to the dispatch; existing shapes are unchanged.
- The guard only replaces an existing C++ abort with an actionable Python error; it never triggers for a supported shape.

## Validation

Built and served `DeepSeek-V4-Flash-DSpark` on GB10 / DGX Spark (`sm_121a`), TP=2, `num_speculative_tokens=5`:

- Engine starts; `/health` → 200.
- Spec decode healthy: mean acceptance length ~4.6/5, per-position acceptance `0.91, 0.79, 0.71, 0.65, 0.58`, ~73% average draft acceptance.
- No decode fallthrough / no abort.

### Independent hardware confirmation (@OsakaTX)

OsakaTX independently validated on 2× DGX Spark (GB10, `sm_121a`) serving `DeepSeek-V4-Flash-DSpark` at TP=2, with the AOT cache masked for isolation A/B testing:

- **Unpatched + topk=256 config entries only (no kernel instantiations):** all 12 topk=256 combos fail (`num_tokens ∈ {1,16,64}` × `with_sink ∈ {False,True}`), all 42 pre-existing combos pass. Error matches exactly: `Check failed: num_tokens > 64 (64 vs. 64)`.
- **With the full topk=256 instantiations:** all 54 combos pass against the bf16 reference.

His diff matched byte-equivalent, confirming the same root cause analysis.

### Notes for local patching

If applying this patch locally (e.g., on a prebuilt container image), the `flashinfer_jit_cache` AOT module for `sparse_mla_sm120` shadows patched source code. The JIT loader finds the stale AOT `.so` and never recompiles, reproducing the original error with the fix applied. **Mask or remove `jit_cache/sparse_mla_sm120/` to force a rebuild.**

## Related

- #3828 / #3896 cover the same topk=256 gap in the prefill path.


OsakaTX:
Independent confirmation on hardware — I hit this exact wall on 2× DGX Spark (GB10, `sm_121a`) serving `DeepSeek-V4-Flash-DSpark` at TP=2, and have been carrying a byte-equivalent local patch since July 3. Your diff matches what I derived from the same root cause: `(32, 256)` misses `_DECODE_DSV4_DISPATCH`, falls through to the prefill orchestrator, and dies on its `num_tokens > 64` check — an error that points everywhere except the missing instantiation. The Python-side guard you added is a nice touch; it would have saved me most of a debugging day.

Validation results from my rig, JIT-built from source with the AOT cache masked (isolation A/B, only the dispatch lines varied):

- **Unpatched main + `(8, 256)`/`(32, 256)` added to `_DSV4_DECODE_CONFIGS`:** exactly the 12 topk=256 combos fail (`num_tokens ∈ {1,16,64}` × `with_sink ∈ {False,True}`), all 42 pre-existing combos pass. Error verbatim: `Check failed: num_tokens > 64 (64 vs. 64) : Decode (num_tokens <= 64) must go through sparse_mla_sm120_decode_dsv3_2 or sparse_mla_sm120_decode_dsv4; got num_tokens=64`
- **With the topk=256 instantiations:** all 54 combos pass against the bf16 reference.

Two small suggestions:

1. Consider adding `(8, 256)` to `_DSV4_DECODE_CONFIGS` alongside `(32, 256)` — NH=8 is the padded small-TP corner case (zero-Q rows, write-gating), so it exercises a different code path than 32 and it passes on my hardware.
2. Maybe worth a line in the PR body for people patching locally: the prebuilt `flashinfer_jit_cache` module for `sparse_mla_sm120` shadows patched source — the loader finds the stale AOT `.so` and never recompiles, reproducing the original error with the fix applied. You have to mask/remove `jit_cache/sparse_mla_sm120/` to force the rebuild.

Related: #3828 notes the prefill path has the same missing-256 gap — that half is now filed as #3896.

vedcsolution:
Reposting — the previous one got mangled by backtick interpretation in the shell. Keeping the actual response below.

vedcsolution:
Thanks for the thorough independent validation, @OsakaTX — really appreciate you taking the time to A/B test it on your rig and share the results.

I've incorporated both suggestions:

1. **`(8, 256)` added to `_DSV4_DECODE_CONFIGS`** alongside `(16, 256)` in the test suite (commit 76fd3daf). Your reasoning about NH=8 exercising the write-gating / zero-Q row path is spot on.
2. **AOT cache note added to the PR body** — the `flashinfer_jit_cache` shadowing caught me too the first time; good to have it documented for anyone patching locally.

I've also linked #3896 for the prefill side. Looking forward to cross-referencing once that lands.

Thanks again — this kind of hardware-backed confirmation is exactly what makes open-source kernel work solid.

ormandj:
Tested on 2x RTX PRO 6000 (SM120, CUDA 13.0), applied to the 0.6.14 wheel along with #3834. Required for vLLM dspark speculative decoding on SM120 — the draft's noncausal index width is always 256 (cdiv(window+n,128)*128), and without this instantiation the decode path has no TOPK=256 kernel.

Testing, serving DeepSeek-V4-Flash-DSpark (TP=2, n=2 and n=5): 4 quality runs of a 35-conversation code-review agent workload (~450-475 turns and ~260-290k completion tokens per run, 5 concurrent), scanning every streamed reasoning+content turn for degenerate output (CJK/marker census) — 0 degenerate turns, 0 tool-call parser failures across all runs. Decode benchmarked at concurrency 1-64 (ctx 0 and 8k) with sane numbers throughout (194 tok/s cc1 → 1772 cc64). Several days of continuous agent traffic on top of that. #3834 is needed too or prefill hits the same missing instantiation.

ormandj:
Reverified the carried TOPK=256 decode instantiation against current vLLM main `d973cce3ca` with the 0.6.14 wheel plus #3817/#3834 source snapshots.

DeepSeek-V4-Flash-DSpark, 2x RTX PRO 6000 Max-Q (SM120), TP2, fixed probabilistic K=5, temperature 0, context 0, 30 s/cell: C1 187.29, C2 283.00, C4 490.12, C8 621.64 tok/s. Draft acceptance was 0.3429 / 0.3741 / 0.3832 / 0.3655. No request errors or inference-time JIT warnings.

Benchmark SHA256: `fa227030012a8b55545af6b6a50fa4adcbdff8d003bb16469dc8e2de024ed0c0`.

ducviet00:
Tried this with https://github.com/eugr/spark-vllm-docker/pull/304 and gott promising results - peak performance on 2x DGX Spark is 60 token/s. Thank you!

ormandj:
Head `76fd3daf` adds the missing TOPK=256 decode instantiations without changing existing dispatch shapes. OsakaTX independently reported on this PR that the new shapes failed before the patch while 42 existing cases passed, and that the patched SM121 matrix passed 54/54 against the bf16 reference. As previously reported here, this carry was also required for sustained SM120 TP2 DSpark serving; #3834 is the separate prefill half. This is dispatch and correctness evidence, not an isolated throughput claim. There are no unresolved review threads; requesting authorized CI and maintainer review for merge.

rafaelkallis:
have been running dsv4 flash dspark for 4 days now (cumulative 50M input tokens, 500K output tokens) using latest vllm and flashinfer with this PR plus the sibling prefill PR #3834 , did not observe any issues.

2x Asus GX10 GB10 SM121

vedcsolution:
Thanks @rafaelkallis — a 4-day, 50M-input/500K-output run with latest vLLM + flashinfer and zero issues is exactly the kind of sustained-production signal that helps kernel patches like this land. Really appreciate you taking the time to report it.

For context, this PR (decode topk=256) and the sibling #3834 (prefill topk=256) are the two halves of the same missing-instantiation gap for DSpark. Independent confirmations have now stacked up across both target archs — SM121 (DGX Spark GB10, @OsakaTX, 54/54 vs bf16 ref) and SM120 (RTX PRO 6000, @ormandj, 0 degenerate turns across 4 multi-day agent runs) — so your production datapoint on the same stack is a great capstone. Thanks again.

saltyminty:
This PR is superseded by and will be closed after https://github.com/flashinfer-ai/flashinfer/pull/4380

bkryu:
Closing as superseded by #4380

PR #3822, opened 2026-07-03
Optimize FP4 V-scale dequant for h512 FA2 prefill

lesj0610:
## 📌 Description

Optimize FP4 V-scale dequantization in the h512 FA2 NVFP4 paged prefill path.

On SM86, the h512 FP4 prefill kernel spends most of its time on V-scale dequant inside `vosplit_compute_pv` — a bypass probe showed about 57% of the kernel time. This PR loads 4 scale rows as one `uint32_t` on SM8x and converts them with the existing `fast_dequant_f8f16x4` helper, instead of doing 4 scalar byte loads and conversions per MMA tile. The scalar path is kept for the non-aligned case, for non-FP4 dtypes, and for SM90+.

The SM90+ gating is based on the SM120 result reported by waynehacking8: the ungated vector path was correct on RTX PRO 6000, but showed a small 1-2.6% regression on some 4K cases. So this revision keeps the measured SM86 speedup while leaving newer architectures on the existing scalar path until there is positive data for them. The KV cache format and smem layout are not touched.

Also added bf16 h512 test cases (with and without RoPE), because bf16 takes a different branch in `fast_dequant_f8f16x4`.

No new kernel instantiations — only the inner loop of the existing FA2 kernel changes. h512 is excluded from the default AOT config anyway, so there is no wheel size impact.

This is a follow-up to flashinfer-ai/flashinfer#3739, which this optimization was split from to keep that PR focused on large-head FA2 enablement.

## 🔍 Related Issues

Follow-up to flashinfer-ai/flashinfer#3739.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

Validation performed:
- h512 NVFP4 paged microbench on RTX 3090 (SM86), kv_len=8192: 4.70 ms -> 1.97 ms
- vLLM serving check (Gemma4 31B, TP=1, nvfp4 KV cache): 4.50 -> 5.58 tok/s, 16/16 requests OK
- h512 NVFP4 prefill correctness tests pass: fp16/bf16 x NONE/ROPE_LLAMA paged + ragged large-head, 6 tests total
- latest local check after SM8x gating: `CUDA_VISIBLE_DEVICES=1 FLASHINFER_DISABLE_VERSION_CHECK=1 python -m pytest tests/attention/test_batch_prefill_kernels.py -q -rs -k "nvfp4_large_head"` — 3 passed, 15852 deselected
- `pre-commit run --files include/flashinfer/attention/prefill.cuh`
- `git diff --check`

External SM120 result from waynehacking8:
- RTX PRO 6000, SM120, CUDA 13.0.88
- all six h512 NVFP4 tests passed and matched main on the tested shapes
- the ungated vector path showed a small regression on some 4K cases, so this revision disables it for SM90+

Not yet verified:
- SM90/H100 performance
- SM100/SM120 performance with this gated revision
- full dtype/page-size test matrix

## Summary by CodeRabbit

* **New Features**
  * Improved attention prefill performance for low-precision FP4/NVFP4 workloads on supported SM8x GPUs by enabling a grouped/vectorized V-scale path during dequant/scaling under specific packing conditions.
* **Tests**
  * Added NVFP4 paged KV-cache “large head” coverage with BF16 queries, including both default positional encoding and RoPE (LLaMA-style) variants, gated by head-dimension support.


lesj0610:
@qsang-nv Opened this as a follow-up to #3739. It only changes the FP4 V-scale inner loop in the h512 FA2 prefill path, about 2.4x on the microbench (4.70 -> 1.97 ms on RTX 3090). Would appreciate your review when you have time.

waynehacking8:
Datapoint from an RTX PRO 6000 (SM120, CUDA 13.0.88): all six h512 nvfp4 prefill tests pass on this branch (including the new bf16 ones), outputs agree with main on every shape I benched (mean |o| identical to 1e-6), and the alignment premise checks out statically - all three vosplit_compute_pv call sites pass warp_vo_base as a multiple of NUM_MMA_D_VO_PER_WARP and v_sf_smem is alignas(16), so the uint32 scale loads are safe. Perf is a different story on this arch though: interleaved median-of-100 CUDA-event timings over three rounds give main 1.562/1.775/1.761 ms vs this PR 1.603/1.792/1.774 ms for f16-4k / bf16-4k / bf16-2k-b4 paged h512 - a small but consistent 1-2.6% regression - while 8k/16k sit inside run-to-run noise, nothing like the 2.4x you measured on SM86. Looks like Blackwell doesn't bottleneck on the scalar scale loads the way SM86 does, so it might be worth checking whether the win holds on SM90/SM100 too, or gating the vectorized path by arch if it doesn't.

lesj0610:
Thanks for the SM120 run.

Gated the vectorized V-scale path to SM8x only — that's where I measured the scalar loads as the bottleneck. SM90+ stays on the existing scalar path, so the regression you saw should be gone. Updated the PR description with your SM120 numbers.

Don't have H100/SM100 locally, so keeping this SM8x-only for now.

waynehacking8:
Re-ran on the same RTX PRO 6000 at ac2c94c1: all six h512 tests pass and the regression is gone -- interleaved median-of-100 x3 rounds now puts main vs PR within 0.3% on the three shapes that regressed before (1.565/1.782/1.763 vs 1.565/1.781/1.763 ms), and mean |output| agrees with main to 1e-6 on every shape. The SM8x-only gate looks right from here.

qsang-nv (approved):

Issue #3168, opened 2026-04-24
[Bug] trtllm_fp4_block_scale_moe autotune corrupts device state on num_tokens=16384 bucket (SM100/SM103, NVFP4 MoE, vLLM-padded hidden 2688→2816)


stecasta:
## Summary
On SM100 (GB200) and SM103 (GB300), `flashinfer.fused_moe.trtllm_fp4_block_scale_moe` produces incorrect behavior when the MoE hidden dim is not a multiple of 512 and the caller has zero-padded weights to the next 256-multiple. Two manifestations, same root cause:

1. **Autotune corruption.** At `max_num_batched_tokens=16384`, autotune profile 6 corrupts the CUDA device context; the next probe surfaces `torch.AcceleratorError: CUDA error: an illegal memory access was encountered`.
2. **Silent numerical corruption.** Engine initialises, autotune completes, generation produces garbage — factual prompts return incorrect tokens.

Both disappear when `FLASHINFER_TRTLLM` is replaced by `FLASHINFER_CUTLASS` on the same padded weights.

## Hardware scope

Verified across four Blackwell devices on `nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-NVFP4` (hidden=2688):

| Device | SM | Selected backend | Behaviour |
|---|---|---|---|
| GB200 | 100 | `FLASHINFER_TRTLLM` + padding 2688→2816 | **Fails** |
| GB300 | 103 | `FLASHINFER_TRTLLM` + padding 2688→2816 | **Fails** |
| Thor AGX | 110 | `FLASHINFER_CUTLASS` | Passes |
| DGX Spark (GB10) | 121 | `VLLM_CUTLASS` (all FI rejected) | Passes |

SM100 / SM103 specific for the TRTLLM-GEN NVFP4 padded path.

## Reproduction
```bash
VLLM_USE_FLASHINFER_MOE_FP4=1 \
vllm serve nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-NVFP4 \
  --tensor-parallel-size 1 --trust-remote-code
```
The model's `hidden_size=2688`. vLLM's `align_trtllm_fp4_moe_hidden_dim_for_fi` zero-pads weights to 2816 at load time, then calls `trtllm_fp4_block_scale_moe`. First inference request hits either manifestation above.

## `max_num_batched_tokens` threshold

[...]

Issue #2776, opened 2026-03-12
[Bug] NVFP4 MoE models crash on GB10 (SM121) during CUDA graph capture


aniskumar-nv:
## Hardware

| | |
|---|---|
| GPU | NVIDIA GB10 (DGX Spark) |
| Compute Capability | SM 12.1 |
| Driver | 580.82.09 |
| CUDA | 13.0 |

```
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 580.82.09              Driver Version: 580.82.09      CUDA Version: 13.0     |
+-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
|   0  NVIDIA GB10                    Off |   0000000F:01:00.0 Off |                  N/A |
+-----------------------------------------------------------------------------------------+
```

## Environment

| | |
|---|---|
| vLLM | 0.17.1 |
| flashinfer | 0.6.4 |
| PyTorch | 2.10.0+cu130 |
| Python | 3.12 |
| `VLLM_USE_FLASHINFER_MOE_FP4` | `1` |

## Model

`nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-NVFP4` (NVFP4 MoE, TP=1)

Dense NVFP4 models (e.g. Llama 3.1 8B) are **not** affected.

## Steps to Reproduce

```bash
VLLM_USE_FLASHINFER_MOE_FP4=1 python -m vllm.entrypoints.openai.api_server \
  --model nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-NVFP4 \
  --tensor-parallel-size 1
```

## Error

Crash occurs at step ~32/51 during PIECEWISE CUDA graph capture:

```
Capturing CUDA graphs (mixed prefill-decode, PIECEWISE):  63%|██████▎| 32/51

torch.AcceleratorError: CUDA error: an illegal memory access was encountered
Search for `cudaErrorIllegalAddress` in https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html
```

## Traceback

```
File "vllm/v1/worker/gpu_model_runner.py", line 5363, in capture_model
    self._capture_cudagraphs(
File "vllm/v1/worker/gpu_model_runner.py", line 5464, in _capture_cudagraphs
    dummy_run(
[...]

PR #3840, opened 2026-07-05
fix(moe): fail loudly instead of IMA when no routing tier covers (numExperts, topK)

jinzh-nvidia:
## 📌 Description

The trtllm custom-routing launchers dispatch kernels through a compile-time
`PolicyTraits` tier list. When no tier covers the runtime `(numExperts, topK)`
for the active policy, the launchers previously only emitted `FLASHINFER_WARN`
and returned **without launching any kernel**. `run()` / `runPostTopKPipeline()`
then continued with **uninitialized routing output**, and the downstream grouped
GEMM read out-of-range indices and crashed with an **illegal memory access**.

This is reproducible with 512-expert TopK-only routing (`RoutingMethodType::TopK`,
i.e. `NoOpPreprocess + NoOpPostprocess`), whose `PolicyTraits` currently only
covers ≤256 experts: during autotune the routing kernel silently fails to launch
and the next launch surfaces `CUDA error: an illegal memory access was encountered`.

This PR makes the "no compiled tier" case **fail fast** instead of corrupting
device state:

- Every tier-dispatched routing launcher now returns whether a kernel was
  actually launched: `launchBlockKernel`, `launchDynBlockKernel`,
  `launchClusterKernel` (+ its `ForPolicy` / `ForBlockDim` helpers),
  `launchHistogramScoresKernel`, `launchBlockScoresKernel` → `bool`.
- `run()` and `runPostTopKPipeline()` now `FLASHINFER_CHECK` each launch and
  raise a clear host-side error (with the offending `numExperts` / `topK` and a
  hint to add a matching `Tier<E, K>`) **before** any downstream work.
- Adds `queryPolicyHasCompiledTier()` in `RoutingCustomPolicy.cuh`, mirroring the
  existing `query*` helpers so it stays exactly in sync with the dispatch predicate.
- `launchCoopKernel` is intentionally left unchanged — it already throws for
  `numExperts > 1024` and covers everything below, so it cannot silently no-launch.

**Scope:** behavior changes only on the previously-broken path. Supported configs
are unaffected; an uncovered `(policy, numExperts, topK)` now raises a
`RuntimeError` instead of triggering an IMA. This does **not** by itself add
support for a new config (that still requires adding a `Tier<E, K>`); it turns
silent memory corruption into an actionable error. This is the fail-fast behavior
previously suggested in reviews of #2803 (`launchCoopKernel`) and #3166
(`queryPolicySupportsBlockPerToken`).

## 🔍 Related Issues

No dedicated tracking issue. #3168 and #2776 report the same *symptom* class
(autotune / graph-capture IMA in `trtllm_fp4_block_scale_moe`) but a **different
root cause** (FP4 hidden-dim padding), so they are **not** fixed by this PR.

### ✅ Pre-commit Checks

## 🧪 Tests

## Reviewer Notes

- No new test was added: the change only converts a silently-broken path into a
  loud error, and existing `tests/moe/test_trtllm_gen_fused_moe.py` already
  covers no-regression. Ran the trtllm-gen subset on GB300 (SM103):
  **38 passed, 0 failed** (Sigmoid / DeepSeekV3 / TopK / Llama4 / Renormalize),
  and the 512-expert TopK repro now raises a clean `RuntimeError` instead of an
  IMA. Happy to add a `pytest.raises` negative test for the no-tier path if you
  prefer.
- Follow-up (separate PR): add `Tier<512, 8>` to
  `PolicyTraits<NoOpPreprocess, NoOpPostprocess>` to actually *support*
  512-expert TopK-only routing (this PR only makes it fail cleanly).
- The `FLASHINFER_CHECK` messages use the stream-style argument form (consistent
  with `utils.cuh` / `exception.h`), so `numExperts` / `topK` are printed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Routing now validates that a matching compiled kernel path exists before running, reducing silent failures for unsupported expert/top-k combinations.
  * When no matching configuration is available, the app now stops with clearer error messages instead of continuing incorrectly.
  * Improved dispatch handling across routing paths so supported configurations are launched reliably.

### DIFF
--- a/csrc/fused_moe/trtllm_backend/trtllm_fused_moe_routing_common.cu
+++ b/csrc/fused_moe/trtllm_backend/trtllm_fused_moe_routing_common.cu
@@ -18,10 +18,14 @@
 
 namespace moe::dev::routing {
 namespace routingCustom {
-// Forward declarations of launch functions
-void launchBlockKernel(Data const& data, uint32_t numThreadsHist, void* stream);
-void launchDynBlockKernel(Data const& data, uint32_t numThreadsHist, void* stream);
-void launchClusterKernel(Data const& data, void* stream);
+// Forward declarations of launch functions.
+// Block/DynBlock/Cluster return whether a compiled tier covered (numExperts, topK)
+// and the kernel was actually launched; the definitions live in
+// trtllm_fused_moe_routing_custom.cu. Keep these signatures in sync (the return
+// type is not part of the mangled name, so a mismatch would silently be UB).
+bool launchBlockKernel(Data const& data, uint32_t numThreadsHist, void* stream);
+bool launchDynBlockKernel(Data const& data, uint32_t numThreadsHist, void* stream);
+bool launchClusterKernel(Data const& data, void* stream);
 void launchCoopKernel(Data const& data, int numBlocksCoop, uint32_t numThreadsHist, void* stream);
 void launchInitExpertCounts(Data const& data, uint32_t numThreadsHist, void* stream);
 void launchHistogramKernel(Data const& data, int numBlocksHistogram, uint32_t numThreadsHist,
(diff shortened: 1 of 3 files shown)


aleozlx (commented):
code review: looks good

waiting for CI run

aleozlx (approved):

PR #3846, opened 2026-07-06
perf(moe): optimize one-sided EP4 NVFP4 dispatch

samnordmann:
## Summary

Optimize one-sided MoE dispatch by:

- constructing routing metadata cooperatively across a warp instead of using a
  serial thread-0 loop;
- retaining one compact destination per rank when `ep_size == 4` and
  `top_k > 4`;
- using a 128-thread CTA for that compact path when every payload is at most
  1,024 bytes per token.

The compact path reduces destination state from `top_k` entries to at most
four and no longer depends on an exact model payload fingerprint. Other EP
sizes and `top_k <= 4` retain the generic path. There is no API change.

The 128-thread choice was measured against 64 and 256 threads on two GB200
nodes. Compared with 64 threads, 128 was effectively neutral at 4-64
tokens/rank, 0.9-2.9% faster at 512 tokens/rank, and 2.5-10.0% faster at 8,192
tokens/rank.

## Performance

GB200, EP4, top-k22, H2048 NVFP4, CUDA graphs. Values average the
medians from two independent opposite-order runs.

| Tokens/rank | Dispatch baseline → candidate | Reduction | Dispatch + combine baseline → candidate | Reduction |
|---:|---:|---:|---:|---:|
| 4 | 19.68 → 13.79 us | 29.9% | 31.78 → 25.83 us | 18.7% |
| 64 | 22.15 → 15.52 us | 29.9% | 35.59 → 28.95 us | 18.7% |
| 512 | 46.90 → 18.21 us | 61.2% | 67.77 → 38.71 us | 42.9% |
| 8192 | 456.43 → 53.04 us | 88.4% | 599.90 → 192.65 us | 67.9% |

End-to-end Nemotron Ultra NVFP4 was measured with vLLM on one GB200 node,
`TP=1`, `DP=4`, `EP=4`, concurrency 16, and `ISL/OSL=50000/2048`.
Four balanced ABBA/BAAB allocations covered 3,072 requests.

| Metric | Baseline | Candidate | Geometric-mean change |
|---|---:|---:|---:|
| Aggregate output throughput | 532.75 tok/s | 562.09 tok/s | +5.5% |
| Mean TPOT | 23.12 ms | 23.06 ms | -0.3% |
| Mean TTFT | 13.16 s | 10.96 s | -16.7% |

## Validation

- all supported top-k values: 200 generic-path changing-payload graph replays each;
- exact packed-NVFP4 target: 1,000 replays in vLLM payload order and 200 in
  the alternate FlashInfer order;
- nonmatching top-k22 payload: 200 replays through the generic fallback;
- four-rank GB200 CUDA-graph validation for compact EP4 `top_k` 6, 8, 10,
  16, and 22 with packed H2048 payloads, plus a wide-BF16 case that retains
  the configured CTA size (Lyris job `2301707`, commit `8a5a6f93`);
- local and GitHub `pre-commit`.

## Summary by CodeRabbit

* **New Features**
  * Added a compact EP4 dispatch mode for MOE AlltoAll routing, improving how repeated destinations are handled.
  * Updated dispatch launch behavior to better match the payload size and expert routing layout.

* **Tests**
  * Added a new test covering compact EP4 dispatch to validate token routing and dispatched outputs across ranks.

### DIFF
--- a/csrc/nv_internal/tensorrt_llm/kernels/communicationKernels/moeAlltoAllKernels.cu
+++ b/csrc/nv_internal/tensorrt_llm/kernels/communicationKernels/moeAlltoAllKernels.cu
@@ -32,6 +32,8 @@ namespace tensorrt_llm::kernels::moe_alltoall {
 #define DISABLE_SYNC_FOR_PROFILING 0
 
 constexpr int kEp4Size = 4;
+constexpr int kCompactDispatchMaxPayloadBytes = 1024;
+constexpr int kCompactDispatchBlockSize = 128;
 
 #ifndef DISABLE_TIMEOUT
 #define DISABLE_TIMEOUT 0
@@ -333,6 +335,7 @@ __global__ void moeA2ADispatchKernel(
     int num_payloads,                       // Number of payloads
     int max_tokens_per_rank,                // Maximum tokens per rank
     int local_num_tokens, int rank_id, int ep_size, int num_experts_per_rank) {
+  static_assert(!COMPACT_EP4 || TOP_K > kEp4Size);
   int thread_idx = threadIdx.x;
   int local_token_idx = blockIdx.x;
 
@@ -547,23 +550,19 @@ void moe_a2a_dispatch_launch(MoeA2ADispatchParams const& params) {
 
   int block_size = tensorrt_llm::common::getEnvMoeA2ADispatchBlockSize();
 
-  bool const is_ep4_topk22 = params.ep_size == kEp4Size && params.top_k == 22;
-  // H2048 NVFP4 dispatch carries packed activations (1024 B), scales (128 B),
-  // expert IDs (88 B), and routing weights (88 B). Callers may order the
-  // auxiliary payloads differently.
-  bool const has_h2048_nvfp4_payloads = params.num_payloads == 4 &&
-                                        kernel_ptrs.payload_bytes_per_token[0] == 1024 &&
-                                        ((kernel_ptrs.payload_bytes_per_token[1] == 128 &&
-                                          kernel_ptrs.payload_bytes_per_token[2] == 88 &&
-                                          kernel_ptrs.payload_bytes_per_token[3] == 88) ||
-                                         (kernel_ptrs.payload_bytes_per_token[1] == 88 &&
-                                          kernel_ptrs.payload_bytes_per_token[2] == 88 &&
-                                          kernel_ptrs.payload_bytes_per_token[3] == 128));
-  bool const use_compact_ep4 = is_ep4_topk22 && has_h2048_nvfp4_payloads;
-  // H2048 NVFP4 has only 64 aligned 16-byte activation vectors. More than four warps
-  // leaves most threads idle during the dominant payload copy.
-  if (use_compact_ep4 && block_size > 128) {
-    block_size = 128;
+  bool const use_compact_ep4 = params.ep_size == kEp4Size && params.top_k > kEp4Size;
+
+  int max_payload_bytes_per_token = 0;
+  for (int i = 0; i < params.num_payloads; ++i) {
+    if (kernel_ptrs.payload_bytes_per_token[i] > max_payload_bytes_per_token) {
+      max_payload_bytes_per_token = kernel_ptrs.payload_bytes_per_token[i];
+    }
+  }
+  // Small payloads do not have enough 16-byte vectors to use larger CTAs. A GB200
+  // sweep found 128 threads faster than both 64 and 256 at prefill token counts.
+  if (use_compact_ep4 && max_payload_bytes_per_token <= kCompactDispatchMaxPayloadBytes &&
+      block_size > kCompactDispatchBlockSize) {
+    block_size = kCompactDispatchBlockSize;
   }
 
   // Configure kernel launch: one block per token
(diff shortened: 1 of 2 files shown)


aleozlx (approved):

samnordmann:
CI failure diagnosis: all five failures were in the new compact-EP4 test before kernel launch. The test added an `int32` route tensor to a default `torch.arange` (`int64`), promoting `token_selected_experts` to `int64`; the C++ API correctly rejected it.

Fixed in `de4e5256` by constructing the token offsets as `int32`. Local pre-commit passes and the new official CI matrix is running. The production kernel was not implicated; its four-rank GB200 CUDA-graph validation remains green.

samnordmann:
@aleozlx can you please trigger CI again?

amirkl94:
@samnordmann Can this be beneficial for every `ep_size < top_k` ? Specifically, I'm wondering if this can be utilized for multi-node as well.

samnordmann:
> @samnordmann Can this be beneficial for every `ep_size < top_k` ? Specifically, I'm wondering if this can be utilized for multi-node as well.

Good question. Conceptually yes: when `ep_size < top_k`, there are at most `ep_size` unique destinations, so the compact representation can generalize.
The benefit is not guaranteed, since the generic path already removes duplicate copies; this only reduces local metadata and fanout overhead, so gains should decrease as `ep_size` approaches `top_k`.
It should also work across nodes over MNNVL, but this PR only validates EP4. I suggest keeping its scope narrow and evaluating EP8/EP16 and multi-node separately.

samnordmann:
The internal pipeline failure [57559844](https://gitlab-master.nvidia.com/dl/flashinfer/flashinfer-ci/-/pipelines/57559844) is unrelated to the PR.

@aleozlx Could we rerun the pipeline ?

xxi-nv:
@bobboli Hi bo, could you help to review this PR and do you want to import to TRTLLM?

bobboli:
Hi @samnordmann, the results look astonishing!

I wonder for this top22 + EP4 is a somehow special case, how much "constructing routing metadata cooperatively" and "compact path" contributes separately? I think the former is a generic optimization. For the later one, I would like to see if it deserves generic support instead of hard-coding for ep size.

samnordmann:
> Hi @samnordmann, the results look astonishing!
> 
> I wonder for this top22 + EP4 is a somehow special case, how much "constructing routing metadata cooperatively" and "compact path" contributes separately? I think the former is a generic optimization. For the later one, I would like to see if it deserves generic support instead of hard-coding for ep size.

Thanks! Yes, EP4 + top-22 is a particularly favorable case. I measured each stage separately, on GB200 nodes. Here is the incremental dispatch latency reduction:

Tokens/rank | Cooperative routing | Compact destinations | 128-thread CTA | Compounded
-- | -- | -- | -- | --
4 | 16.8% | 16.6% | 0.2% | 31.9%
64 | 17.3% | 14.2% | 0.4% | 29.2%
512 | 28.2% | 43.4% | 4.3% | 61.3%
8192 | 34.9% | 80.0% | 4.8% | 88.0%

I decided to put the differents optimizations in the same PR bc the patch remains relatively small.

The PR already enables the "compact destinations" EP4 with every `top_k > 4` (only the EP size is specialized). I agree the principle extends to `ep_size < top_k`, but I would preserve constexpr counts rather than runtime, e.g. specialize EP2/EP4/EP8 kernels. I could add other counts as follow up after benchmark, wdyt?

bobboli:
Thanks for the breakdown. I ported this to TRT-LLM and compared using `ep_size` as a template argument, which enables compile-time unrolling, against using it as a normal runtime argument. The runtime version retained essentially all the benefit in my B300 tests, suggesting the gain comes from iterating over `ep_size` instead of `top_k`, rather than from specialization or unrolling.

Based on this, it may be worth considering a generic runtime compact path when `ep_size < top_k`, rather than adding separate EP2/EP4/EP8 template specializations.

samuellees:
cc @b8zhong

samnordmann:
> Thanks for the breakdown. I ported this to TRT-LLM and compared using `ep_size` as a template argument, which enables compile-time unrolling, against using it as a normal runtime argument. The runtime version retained essentially all the benefit in my B300 tests, suggesting the gain comes from iterating over `ep_size` instead of `top_k`, rather than from specialization or unrolling.
> 
> Based on this, it may be worth considering a generic runtime compact path when `ep_size < top_k`, rather than adding separate EP2/EP4/EP8 template specializations.

Thank you for the insight. Can you share a pointer of the trtllm patch?

bobboli:
> > Thanks for the breakdown. I ported this to TRT-LLM and compared using `ep_size` as a template argument, which enables compile-time unrolling, against using it as a normal runtime argument. The runtime version retained essentially all the benefit in my B300 tests, suggesting the gain comes from iterating over `ep_size` instead of `top_k`, rather than from specialization or unrolling.
> > Based on this, it may be worth considering a generic runtime compact path when `ep_size < top_k`, rather than adding separate EP2/EP4/EP8 template specializations.
> 
> Thank you for the insight. Can you share a pointer of the trtllm patch?

https://github.com/NVIDIA/TensorRT-LLM/pull/16628/changes.
Perf numbers are tested by agent. Please take it with a grain of salt.

PR #3863, opened 2026-07-07
fix(attention): clamp CTA_TILE_Q for fp8 h512 ragged prefill on 99KB-smem GPUs

### LINKED ISSUE #3843: [bug] test_batch_prefill_with_ragged_kv_cache_fp8 error: this configuration is not supported on this architecture.
### Summary
Found this issue in latest nightly [unit-test: 56860555 ](https://nv/flashinfer-ci/-/pipelines/56860555) 
### Observed jobs:
[unit_test_5090: [cu129]](https://nv/flashinfer-ci/-/jobs/355030343)
[unit_test_5090: [cu130]](https://nv/flashinfer-ci/-/jobs/355030362)

```python
tests/attention/test_fp8_prefill.py
``` 
```shell
 E   RuntimeError: Error in function 'BatchPrefillWithRaggedKVCacheDispatched' at /tmp/flashinfer/include/flashinfer/attention/prefill.cuh:4070: Required shared memory (132176 bytes) for head_dim_qk=512, head_dim_vo=512, cta_tile_q=32 exceeds this GPU's per-block limit (101376 bytes); this configuration is not supported on this architecture.
 python/tvm_ffi/cython/function.pxi:929: RuntimeError: Error in function 'BatchPrefillWithRaggedKVCacheDispatched' at /tmp/flashinfer/include/flashinfer/attention/prefill.cuh:4070: Required shared memory (132176 bytes) for head_dim_qk=512, head_dim_vo=512, cta_tile_q=32 exceeds this GPU's per-block limit (101376 bytes); this configuration is not supported on this architecture.
 E   RuntimeError: Error in function 'BatchPrefillWithRaggedKVCacheDispatched' at /tmp/flashinfer/include/flashinfer/attention/prefill.cuh:4070: Required shared memory (132176 bytes) for head_dim_qk=512, head_dim_vo=512, cta_tile_q=32 exceeds this GPU's per-block limit (101376 bytes); this configuration is not supported on this architecture.
``` 
### Suggested Fix (by Codex)
Skip when requested memory exceeds the opt-in limit in tests/attention/test_fp8_prefill.py. 
```python
_HEAD_DIM_512_SMEM_BYTES = 132176

def _max_dynamic_smem_per_block() -> int:
    props = torch.cuda.get_device_properties(torch.device("cuda:0"))
    return getattr(
        props, "shared_memory_per_block_optin", props.shared_memory_per_block
    )

def skip_if_head_dim_unsupported(head_dim: int):
    if head_dim > 256 and not head_dim_512_supported():
        pytest.skip("head_dim > 256 is only supported on SM100 or newer")
    if head_dim == 512 and _max_dynamic_smem_per_block() < _HEAD_DIM_512_SMEM_BYTES:
        pytest.skip(
            "head_dim=512 prefill requires ~132 KB shared memory per block, "
            "which exceeds this GPU's per-block limit"
        )
```

qsang-nv:
## 📌 Description

Fixes the SM120 (RTX 5090) nightly failures in `tests/attention/test_fp8_prefill.py`:
FA2 ragged batch prefill with FP8 KV at `head_dim=512` failed to launch with
`Required shared memory (132176 bytes) ... exceeds this GPU's per-block limit (101376 bytes)`.

**Root cause.** FP8 KV is excluded from `USE_SOFTMAX_VO_SPLIT` (16-bit/FP4 only), so the
ragged and single-prefill kernels use the non-VO-split `SharedStorage`, whose CTA_TILE_Q=32
layout carries a cross-warp output-merge buffer of `NUM_WARPS_KV(4) × 32 × min(head_dim,256)`
floats — 132KB regardless of `NUM_MMA_KV`, which can never fit the 101376B per-block opt-in
limit of SM86/89/120/121-class GPUs. The paged kernel (`SharedStoragePaged`, VO-split) fits
at CTA32 and is unaffected.

**Fix.** Clamp `CTA_TILE_Q` to 16 (~82KB, launches; same config short-q already uses) at plan
time when the device's `MaxSharedMemoryPerBlockOptin` cannot fit the CTA32 layout. Since the
paged and ragged wrappers share one `plan` export and must not be clamped alike, this adds a
new **additive** `ragged_plan` export with the same signature — the existing `plan` /
`workspace_size` arity and paged semantics are unchanged (no ABI break), and older modules
without the symbol fall back to `plan` via `getattr`. The FP8 predicate comes from the
compile-time `DTypeKV` already visible in the plan TU; large-smem GPUs (H100/A100/B200) keep
CTA_TILE_Q=32.

**Tests.** Adds three regression tests: ragged fp8 h512 long-q asserts the planned
`cta_tile_q` (16 on 99KB devices, 32 otherwise) and runs; paged fp8 h512 asserts CTA32 is
kept everywhere (guards against silent perf regression); single-prefill fp16-Q/fp8-KV h512
long-q numerics vs the 16-bit reference.

Verified on RTX PRO 6000 Blackwell (SM120 — same CC 12.0 and 101376B opt-in limit as the
failing 5090 runners): the failing cases reproduce the exact CI error before the change;
after it, `test_fp8_prefill.py` passes 1603/1603, and bf16 h512 (ragged+paged) plus the full
`test_single_prefill.py` are unregressed. SM121 shares the same limit and the gate is
attribute-based, so it is expected to behave identically (not tested locally, no hardware).

## 🔍 Related Issues

Fixes #3843

### ✅ Pre-commit Checks

## 🧪 Tests

## Reviewer Notes

This is the availability fix only: CTA16 is a slower tile for long-q, but the alternative
today is a hard launch failure. Extending the VO-split softmax path to FP8 KV (keeping CTA32
on 99KB parts) is a separate perf follow-up.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved FP8 attention behavior for large head dimensions, aligning kernel layout selection with 16-bit paths.
  * Fixed planning for FP8 prefill cases so long-sequence and ragged inputs use the expected tile and warp configuration.
  * Added coverage to verify FP8 outputs match FP16-dequantized references for single, paged, and ragged prefill scenarios.

### DIFF
--- a/csrc/batch_prefill.cu
+++ b/csrc/batch_prefill.cu
@@ -44,13 +44,22 @@ using namespace flashinfer;
 using tvm::ffi::Array;
 using tvm::ffi::Optional;
 
-Array<int64_t> BatchPrefillWithKVCachePlan(
+// FP8 KV runs the ragged (and single-prefill) kernels on the non-VO-split
+// SharedStorage whose cross-warp output-merge buffer cannot fit CTA_TILE_Q=32 at
+// head_dim>=512 on ~99KB-smem GPUs; the paged kernel's SharedStoragePaged is
+// VO-split and unaffected. FA2DetermineCtaTileQ needs this distinction, so the
+// ragged wrapper plans through a separate `ragged_plan` export instead of the
+// shared `plan` (whose signature and paged semantics stay unchanged).
+constexpr bool kFp8KvNoVOSplit =
+    std::is_same_v<DTypeKV, __nv_fp8_e4m3> || std::is_same_v<DTypeKV, __nv_fp8_e5m2>;
+
+static Array<int64_t> BatchPrefillWithKVCachePlanImpl(
     TensorView float_workspace_buffer, TensorView int_workspace_buffer,
     TensorView page_locked_int_workspace_buffer, TensorView qo_indptr, TensorView kv_indptr,
-    TensorView kv_len_arr, int64_t total_num_rows, int64_t batch_size, int64_t num_qo_heads,
-    int64_t num_kv_heads, int64_t page_size, bool enable_cuda_graph, int64_t head_dim_qk,
-    int64_t head_dim_vo, bool causal, int64_t window_left, int64_t fixed_split_size,
-    bool disable_split_kv, int64_t num_colocated_ctas = 0) {
+    int64_t total_num_rows, int64_t batch_size, int64_t num_qo_heads, int64_t num_kv_heads,
+    int64_t page_size, bool enable_cuda_graph, int64_t head_dim_qk, int64_t head_dim_vo,
+    int64_t window_left, int64_t fixed_split_size, bool disable_split_kv,
+    int64_t num_colocated_ctas, bool fp8_kv_no_vo_split) {
   size_t float_workspace_size_in_bytes =
       float_workspace_buffer.size(0) * get_element_size(float_workspace_buffer);
   size_t int_workspace_size_in_bytes =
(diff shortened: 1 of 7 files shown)


bkryu (commented):
Hi @qsang-nv, the previous PR https://github.com/flashinfer-ai/flashinfer/pull/3576 had a 2x2 Q and KV warp layout that fit in the 99KB SMEM of SM86/89/120. The current PR fixes the SMEM issue with clamping.

The 2x2 warp layout seems to have been removed in #3739. I'm inclined to believe that FP8 KV performance is better with the 2x2 layout. do you mind taking a look?

qsang-nv:
You're right — thanks for catching this. #3576's layout gate was `sizeof(DTypeKV) <= 2`
(FP8 included, deliberately — the 2×2 layout is what made the merge storage fit 99KB parts),
and #3739 narrowed it to `(sizeof == 2) || is_fp4` while adding the VO-split paths, dropping
FP8 back to the 1×4 layout whose 132KB merge buffer can't launch. I found no benchmark or
correctness result in #3739 motivating that change, so this looks like a regression from
coupling the layout gate with the new VO-split dtype gate, not a day-one limitation.

I restored FP8 in the 2×2 gate locally (ragged + single dispatch sites, keeping FP8 out of
`kVOSplitDispatch`/`USE_SOFTMAX_VO_SPLIT`). On an RTX PRO 6000 (SM120): all 2592
h512-related tests pass, including the previously-crashing FP8 ragged cases and partial
Q-tile lengths. Perf vs this PR's CTA16 clamp (FP8 KV, h512, causal, ragged; CUDA-event
timing, cold L2, median):

| bs × seqlen | heads | CTA16 clamp | 2×2 CTA32 | time reduction |
|---|---:|---:|---:|---:|
| 1 × 8192  | 32/8 | 84.8 ms   | 81.3 ms   | 4.1% |
| 1 × 16384 | 32/8 | 333.8 ms  | 321.1 ms  | 3.8% |
| 1 × 32768 | 32/8 | 1324.6 ms | 1275.3 ms | 3.7% |
| 4 × 4096  | 32/8 | 86.6 ms   | 82.0 ms   | 5.3% |
| 1 × 8192  | 4/4  | 11.19 ms  | 11.38 ms  | −1.7% |

So for the target GQA configs the 2×2 is consistently ~4-5% faster (4/4 MHA is neutral),
and it fixes the layout at its single point of truth instead of adding planner-side
formulas and a new `ragged_plan` export. I'll rework the PR accordingly, keeping a
regression test that pins FP8-h512 ragged to CTA32 on a 99KB part, plus explicit
partial-Q-tile coverage given the historical 2×2 numerical issue in #523.

bkryu (approved):

PR #3870, opened 2026-07-07
fix(autotuner): time tactics with %globaltimer under Confidential Computing

elvischenv:
## Problem

CUDA event timing (cudaEventElapsedTime) is unreliable when confidential compute (CC) is enabled. This adds an alternative timing backend that uses a small CUDA kernel reading %globaltimer before and after the profiled work, then computes elapsed time on the host.

## Fix

Time candidate tactics with the GPU's `%globaltimer` register (a tiny JIT stamp kernel) instead of CUDA events — monotonic and CC-safe. The return value/signature of the timed path is unchanged, so `choose_one` and the tuning-cache format are untouched.

- `flashinfer/utils.py`: `is_confidential_compute()` (NVML CC detection) and `get_globaltimer_kernel()` (lazy `load_inline` stamp kernel), both `@lru_cache`'d and reusable.
- `flashinfer/autotuner.py`: `AutoTuner.__init__` selects the timing backend once; `pure_profile` brackets the run through a unified `record_start` / `record_end` / `elapsed_time` path (globaltimer stamps vs CUDA events).

**Off-CC behavior is unchanged** — by default the `%globaltimer` path is used only when CC is detected; otherwise the legacy `cudaEvent` timer is kept. If the stamp kernel fails to build, it falls back to `cudaEvent`.

Mirrors TensorRT-LLM PR [#11657](https://github.com/NVIDIA/TensorRT-LLM/pull/11657).

## Control

`FLASHINFER_AUTOTUNE_TIMER`:
- `globaltimer` — force `%globaltimer`
- `cuda_event` — force the legacy cudaEvent timer
- unset / anything else (default) — `%globaltimer` iff CC is detected

CC detection is via NVML; override with `FLASHINFER_CONFIDENTIAL_COMPUTE=1/0`.

## Test

`tests/autotuner/test_global_timer.py`:
- `test_global_timer_vs_cuda_event[shape-mode]` — times identical GEMM work with both timers, interleaved over 6 trials, in **eager** and **CUDA-graph** modes; asserts both means > 0 and that they agree within `max(0.01 ms, 5% · event_mean, 3 · combined_SEM)`.
- `test_globaltimer_monotonic` — sanity: two stamps around real work are strictly increasing, and more work reads a larger delta.
- Skips cleanly without CUDA or if the stamp kernel can't JIT-build.

Passing on H100:
```
5 passed in 41.31s
```

## Summary by CodeRabbit

* **New Features**
  * Added an alternate GPU timing backend for autotuning, supporting an environment-controlled “global timer” mode with automatic detection in confidential-compute environments.
  * If the global-timer kernel isn’t available, the autotuner automatically falls back to standard CUDA event timing.
* **Bug Fixes**
  * Improved autotuning timing reliability in confidential-compute environments by using a CC-safe timing path when possible.
* **Tests**
  * Added CUDA tests covering global-timer monotonicity and agreement with CUDA events, including CUDA Graph and GEMM scenarios.


nvpohanh (approved):

nvpohanh:
cc @leejnau for SGLang request tracking. thanks

flashinfer/utils.py
```diff
@@ -1367,3 +1372,67 @@ def prepare_jit_additional_args(
...
+
+        # per-process name so the 4 TP ranks don't race on the JIT build dir
+        tag = os.environ.get("LOCAL_RANK") or str(os.getpid())
+        mod = load_inline(
+            name=f"flashinfer_globaltimer_{tag}",
+            cpp_sources="void get_globaltimer_timestamp(torch::Tensor);",
+            cuda_sources=_GLOBALTIMER_KERNEL_CU,
+            functions=["get_globaltimer_timestamp"],
+            verbose=False,
+        )
```
qiching, line 1433: Per-rank name is counterproductive and still races across nodes.

Suggest dropping the tag and using a fixed name (`flashinfer_globaltimer`) so torch's built-in lock does the right thing. If isolation is truly needed, key it on a globally-unique id (hostname + global RANK), not LOCAL_RANK.

flashinfer/autotuner/autotuner.py
```diff
@@ -1030,6 +1034,29 @@ def __init__(
             TuningConfig, dict[tuple[tuple[int, ...] | None, bool], TuningConfig]
         ] = weakref.WeakKeyDictionary()
 
+        # Timing backend: globaltimer kernel vs cuda events.
+        # FLASHINFER_AUTOTUNE_TIMER env var overrides auto-detection:
+        #   "globaltimer" -> force globaltimer
+        #   "cuda_event"  -> force cuda events
+        #   unset/default -> auto-detect via is_confidential_compute()
+        timer_env = os.getenv("FLASHINFER_AUTOTUNE_TIMER", "").lower()
```
qiching, line 1085: the import hunk in this file only touches the `flashinfer.utils` import, no import os is added. If flashinfer/autotuner/autotuner.py doesn't already import os at module level, the first `AutoTuner.get()` will raise NameError. Please double-check and add import os if missing.

qiching (changes requested):

flashinfer/autotuner/autotuner.py
elvischenv: os is already imported in `flashinfer/autotuner/autotuner.py`
https://github.com/flashinfer-ai/flashinfer/blob/ce40d25a4cd05b2f28c935b99c9ea5091b4e64d2/flashinfer/autotuner/autotuner.py#L8

flashinfer/utils.py
elvischenv: Updated to use the fixed name `flashinfer_globaltimer`

qiching (approved):
LGTM

PR #3871, opened 2026-07-07
Fix: graph-safe uniform multi-token decode on FA2 tensor-core path

lucifer1004:
### 📌 Description

The tensor-core decode path (`BatchDecodeWithPagedKVCacheWrapper` with `use_tensor_cores=True`, fa2/fa3 backends) always plans `qo_indptr` as one row per request with a non-causal mask, regardless of `q_len_per_req`. Three consequences:

1. **Correctness**: `q_len_per_req > 1` (the spec-decode verify shape: every request carries `1 + num_draft` tokens) silently produces wrong results — only the first `batch_size` rows of `q` are mapped, and there is no causal masking inside each request's token block. `q_len_per_req` is mid-migration from `run()` to `plan()` (soft-deprecation warning added earlier); this PR completes the migration for the tensor-core path.
2. **Capability**: uniform multi-token batches could not be captured in CUDA graphs on the generic path, even though the decode path's replay-safety contract (host-side planning depends only on frozen shapes; per-request kv lengths are read by the kernel from device-side indptr at run time) extends naturally to a fixed `q_len_per_req`. Serving stacks therefore cap FlashInfer GQA spec-decode at single-token capture and fall back to piecewise graphs on architectures without trtllm-gen kernels (e.g. SM120 / RTX PRO 6000).
3. **Workspace**: under `enable_cuda_graph` the prefill scheduler sizes `cta_tile_q` for the ragged worst case (`total_num_rows - batch_size + 1` rows in one request). For a uniform multi-token batch this inflates the tile (e.g. 16 → 128) and the split-kv scratch: at batch 8 / `q_len_per_req=3` / 64:8 heads the plan demands 197 MB, overflowing the common 128 MB workspace.

Changes:

- `plan()` / `fast_decode_plan()` / `workspace_size()`: scale `qo_indptr` and `total_num_rows` by `q_len_per_req`, plan causal when `q_len_per_req > 1`, refresh the persistent `qo_indptr` buffer under cuda-graph mode, and stop discarding `q_len_per_req` in `workspace_size()`.
- `run()`: dispatch `MaskMode.CAUSAL` from the planned q_len; raise `ValueError` when the `q` tensor does not match the planned `q_len_per_req` on the fa2/fa3 path (trtllm-gen and cute-dsl keep their run-time `q_len` flexibility).
- scheduler: new `uniform_q_len` plan hint (0 keeps the existing ragged semantics unchanged). When set under `enable_cuda_graph`, `cta_tile_q` and `total_num_tiles_q` are sized for the exact uniform length, and the per-request qo lengths are validated against the promise at plan time — misuse raises instead of silently corrupting replays. The 197 MB case above drops below the default 128 MB workspace.
- All 10 Python call sites of the fa2 plan binding updated for the new arity; `pod`/`sparse`/`prefill` pass 0 (no behavior change), and decode passes 0 when `q_len_per_req == 1`, so existing single-token decode plans are bit-identical.

### 🔍 Related Issues

Spec-decode verify batches on GQA models via the generic (non-trtllm-gen) path; motivated by enabling FULL-cudagraph MTP on SM120 in vLLM (integration follow-up on the vLLM side will detect this capability from the plan signature).

### ✅ Pre-commit Checks

- [x] pre-commit run on all changed files (clang-format, ruff check/format, mypy: clean)

### 🧪 Tests

- New `test_cuda_graph_uniform_multi_token_decode_with_paged_kv_cache`: 16 parametrized cases (batch 4/8 × q_len 2/3 × kv-heads 2/8 × GQA group 4/8), capture at one set of kv lengths, then two re-plan + replay rounds with different kv lengths, each compared against a fresh `BatchPrefillWithPagedKVCacheWrapper` reference (causal, scaled qo_indptr). With the uniform hint the outputs are bitwise-identical to the reference on SM120.
- New `test_tensor_core_decode_rejects_mismatched_q_len` covering the run-time guard.
- Existing cuda-graph decode suite unaffected: 450 passed (433 pre-existing + 17 new), 0 failed.

### Reviewer Notes

- The replay-safety argument: `plan_info` quantities that pin the captured kernel (grid, `cta_tile_q`, `padded_batch_size`) depend only on `(batch_size, total_num_rows, uniform_q_len)`, all frozen per capture; per-request tile assignments are rewritten into device buffers by each re-plan outside the graph, and kv lengths are read from device-side indptr inside the kernel — same contract that already makes single-token decode graph-safe.
- The `uniform_q_len` validation loop is O(batch) over an array the planner already builds.
- fa3/sm90 has a separate plan entry point (`BatchPrefillWithKVCacheSM90Plan`) and is untouched.

## Summary by CodeRabbit

* **New Features**
  * Added multi-token-per-request paged-KV decode support via `q_len_per_req` (default `1`), with uniform query-length handling to keep planning and execution consistent (including `fast_decode_plan`).
* **Bug Fixes**
  * Improved correctness and error handling for CUDA-graph and tensor-core decode when `q_len_per_req > 1`, including stricter validation and “frozen” shape enforcement.
* **Tests**
  * Added coverage for CUDA-graph + tensor-core uniform multi-token decode, `q_len_per_req` frozen behavior, and plan/run mismatch failures.


flashinfer/decode.py
```diff
@@ -1279,6 +1282,8 @@ def plan(
             logits_soft_cap = 0.0
 
         qo_indptr_host = _get_range_buf(batch_size + 1, "cpu")
+        if q_len_per_req > 1:
+            qo_indptr_host = qo_indptr_host * q_len_per_req
```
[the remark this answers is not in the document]
lucifer1004: Done in 99ddd50e — added the `q_len_per_req >= 1` validation and the tensor-core requirement check in `plan()`. One note on scope: the wrapper already forces `use_tensor_cores=True` for the trtllm-gen and cute-dsl backends at construction, so `not self.use_tensor_cores` exactly identifies the standard-kernel path and this check does not restrict those backends. Covered by a new case in `test_tensor_core_decode_rejects_mismatched_q_len`.

flashinfer/decode.py
```diff
@@ -1762,6 +1770,20 @@ def run(
             # Infer runtime q_len from q.size(0). Doesn't need to match planned q_len
             q_len_per_req = q.size(0) // actual_batch_size
 
+        planned_q_len = getattr(self, "_q_len_per_req", 1) or 1
+        if (
+            self.use_tensor_cores
+            and self._backend not in ("trtllm-gen", "cute-dsl")
+            and q_len_per_req != planned_q_len
+        ):
```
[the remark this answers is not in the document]
lucifer1004: Done in 99ddd50e — `run()` now raises for `q_len_per_req > 1` on the non-tensor-core path, ahead of the planned-vs-inferred consistency check.

flashinfer/decode.py
```diff
@@ -3429,6 +3456,7 @@ def fast_decode_plan(
     - Remove unnecessary host-to-device copy for the metadata buffers.
     """
     batch_size = len(last_page_len)
+    self._q_len_per_req = q_len_per_req
```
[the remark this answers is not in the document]
lucifer1004: Done in 99ddd50e — the same `q_len_per_req >= 1` and tensor-core-requirement validation is applied in `fast_decode_plan()` before the planned value is stored.

flashinfer/decode.py
```diff
@@ -1098,26 +1099,29 @@ def workspace_size(
...
+                batch_size * q_len_per_req,  # total_num_rows
                 batch_size,
                 num_qo_heads,
                 num_kv_heads,
                 page_size,
                 self.is_cuda_graph_enabled,
                 head_dim,
                 head_dim,
-                False,  # causal
+                q_len_per_req > 1,  # causal
```
bkryu: Can we separate out the variable as `is_causal` and make `is_causal=q_len_per_req > 1` for now? The current implementation assumes equivalence between causal and q_len_per_req>1. However, DFlash will require causal=False, so we'd set the stage a bit cleanly for it.

flashinfer/decode.py
bkryu, line 1: Throughout the PR here, I am suspecting that changing the `q_len_per_req` after the wrapper has been created and planned once will cause issues, even when the `plan()` is called with a new `q_len_per_req`.

I know that changing `q_len_per_req` is an unusual use case, but if this is the case, can you add a comment stating that Once a wrapper is planned with a `q_len_per_req` value, it cannot be re-planned with a different value?

flashinfer/decode.py
```diff
@@ -1278,7 +1282,16 @@ def plan(
         if logits_soft_cap is None:
             logits_soft_cap = 0.0
 
+        if q_len_per_req < 1:
+            raise ValueError(f"q_len_per_req must be >= 1, got {q_len_per_req}")
         qo_indptr_host = _get_range_buf(batch_size + 1, "cpu")
+        if q_len_per_req > 1:
```
bkryu, line 1430: If we are not enabling `q_len_per_req>1` for the `fa3` backend, can we raise an error or note somewhere about it?

flashinfer/decode.py
bkryu: Can you check whether this docstring is still correct?

bkryu (commented):
Thank you @lucifer1004 left a few comments

flashinfer/decode.py
lucifer1004: Thanks, the mask decision is now an explicit is_causal = q_len_per_req > 1.

flashinfer/decode.py
lucifer1004: Non-graph wrappers can re-plan with a different q_len_per_req safely (each plan rebuilds qo_indptr and the run-time guard validates against the planned value), and now error will be raised if user tries to do the same in graph mode.

flashinfer/decode.py
lucifer1004: Yes, `NotImplementedError` will be raised now for fa3 backend.

flashinfer/decode.py
lucifer1004: It has been updated.

flashinfer/pod.py
saltyminty, line 1113: Would we also need to wire uniform_q_len into this call?

lucifer1004: Nice catch! Fixed in the latest commit.

flashinfer/decode.py
```diff
@@ -1414,6 +1460,8 @@ def _plan_impl(
             self._paged_kv_indices_buf[: len(indices)].copy_(
                 indices, non_blocking=(indices.device == self.device) and non_blocking
             )
+            if self.use_tensor_cores and q_len_per_req > 1:
```
saltyminty: self.use_tensor_cores is gated on backend being trtllm-gen or cute-dsl, but we want this to run for fa2. I don't believe this is safe check to make (?)

Otherwise LGTM

lucifer1004: Added a gate

saltyminty:
pipeline failures look unrelated

saltyminty (approved):

PR #3803, opened 2026-07-02
Fix 0/0 NaN in GLM52 routing renorm on sigmoid underflow

xianbaoqian:
The DeepSeekV3 no routing path (SigmoidBias preprocess, ScaledSumNormalizePostprocess) renormalizes the top-K weights as weight = sigmoidScore * routeScale / (sum + mSumEpsilon), where sum is the sum of the selected experts' un-biased sigmoid scores. mSumEpsilon defaults to 0 on this path.

Selection uses the bias-added score, so a token can select K experts whose pre-bias logits are all strongly negative; their sigmoids underflow to exactly 0.0 in bf16, so sum is bit-exactly 0 and the divide is 0.0/0.0 = NaN, poisoning the token's whole output row.

Set mSumEpsilon = 1e-20f, matching the DeepSeek-V3 reference gate (modeling_deepseek.py normalizes with sum + 1e-20) and the sibling MiniMax2 branch in this file (already 1e-20f). Normal tokens (sum >> 1e-20) are bit-identical; degenerate tokens no longer NaN.

Link: https://github.com/deepseek-ai/DeepSeek-VL2/blob/main/deepseek_vl2/models/modeling_deepseek.py

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [x] Tests have been added or updated as needed.
- [x] All tests are passing (`unittest`, etc.).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved routing stability for DeepSeekV3 in the no-groups path.
  * Prevented rare `NaN` values during normalization when values underflow to zero.
  * Aligned normalization behavior more closely with reference DeepSeek-V3 results.

PR #3875, opened 2026-07-07
fix: preserve DeepSeek no-group sigmoid routing weights

alexeldeib:
## Summary

DeepSeek no-group routing selects experts with `sigmoid(logit) + bias`, but its
final `ScaledSumNormalize` weights must be normalized from the unbiased
`sigmoid(logit)`.

The warp-per-token TopK paths previously recovered that value as
`selected_score - bias`. For GLM-5.2, selected sigmoid values can be around
`1e-8` to `1e-7` while routing biases are around `12`. In FP32, adding the tiny
sigmoid to the bias can round back to the bias, so subtracting the bias yields
zero for every selected expert.

This PR:

- recomputes `sigmoid(raw)` from the selected experts' original logits in both
  the array and lane-owned warp-per-token paths;
- declares that raw-score dependency through a postprocess capability trait;
- keeps the block-per-token path unchanged, where the unbiased sigmoid is
  already retained in `smemAux`;
- enforces `topK <= numExperts` before both raw and precomputed routing branches;
- adds SM100 regression coverage for the array TopK and high-expert lane-owned
  HistogramScores paths, with PDL disabled and enabled; and
- corrects the adjacent epsilon and routing-bias dtype documentation.

## Relationship to #3803 and #3946

#3803 and #3946 add a `1e-20` denominator guard to the no-group and grouped
DeepSeek paths. Those changes prevent division by zero, but they cannot recover
nonzero numerator values already lost through `(sigmoid + bias) - bias`
cancellation.

This PR restores the unbiased-score dataflow used by the legacy DeepSeek kernel
and the current block-per-token path. The existing denominator guard complements
this fix.

## Validation

- Repository pre-commit hooks pass for all modified files.
- Python byte-compilation passes for the modified Python files.
- The original array-path B200 regression passed with the fix and failed on the
  baseline because `output[:, 0]` was `0` instead of `routeScale=2.5`.
- The rebased regression now also includes an `E=896`, `K=16` lane-owned
  HistogramScores case.
- e2e validated with glm5.2 on gb200

## Summary by CodeRabbit

* **Bug Fixes**
  * Added stricter runtime validation for fused MoE routing to reject invalid expert configurations when `top_k` exceeds `num_experts`, preventing unstable or erroneous results.
  * Improved DeepSeekV3-style fused MoE normalization for the `n_group=1` case by using an unbiased sigmoid computation for more numerically correct BF16 routing outputs.
* **Documentation**
  * Updated `routing_bias` parameter docs to allow `float32` or `bfloat16`.
  * Clarified the epsilon behavior used during routing normalization.
* **Tests**
  * Added a CUDA-only regression test (SM100a-gated) covering the DeepSeekV3 `n_group=1` sigmoid-bias scenario with expected output shape and finiteness checks.

### DIFF
--- a/csrc/fused_moe/trtllm_backend/trtllm_fused_moe_routing_custom.cu
+++ b/csrc/fused_moe/trtllm_backend/trtllm_fused_moe_routing_custom.cu
@@ -1479,6 +1479,9 @@ void run(Data const& data, void* stream) {
   TVM_FFI_ICHECK(data.mPtrTopKPacked != nullptr || data.mPtrScores != nullptr ||
                  data.mPtrTopKIds != nullptr)
       << "Routing kernel requires at least one input parameter";
+  TVM_FFI_ICHECK_LE(data.mTopK, data.mNumExperts)
+      << "Routing kernel expects topK (" << data.mTopK << ") to be <= numExperts ("
+      << data.mNumExperts << ").";
 
   // When topK is already computed (mPtrTopKIds or mPtrTopKPacked without scores),
   // delegate to the shared post-topK pipeline which handles all path selection
--- a/include/flashinfer/trtllm/fused_moe/RoutingCustomPolicy.cuh
+++ b/include/flashinfer/trtllm/fused_moe/RoutingCustomPolicy.cuh
@@ -386,14 +386,19 @@ struct SumNormalizePostprocess {
   }
 };
 
-/// ScaledSumNormalize: recovers un-biased sigmoid scores by subtracting per-expert bias from the
-/// selection scores (sigmoid + bias), then normalizes by sum and applies routeScale.
+/// ScaledSumNormalize: normalizes un-biased sigmoid scores by sum and applies
+/// routeScale. SigmoidBias selection uses sigmoid(raw) + bias as the top-K key,
+/// but final weights must use sigmoid(raw). Warp-per-token routing recomputes
+/// selected sigmoid(raw) from raw logits to avoid cancellation when a tiny
+/// sigmoid is rounded away by a large bias; block-per-token routing reads the
+/// same un-biased sigmoid from smemAux.
 /// Used by DeepSeek-style routing: final_weight = sigmoid(raw) * routeScale / (sum + epsilon).
-/// DeepSeek uses epsilon=0 (no guard); MiniMax2 uses epsilon=1e-20 to prevent division by zero.
+/// The DeepSeek and MiniMax2 runners use epsilon=1e-20 to prevent division by zero.
 struct ScaledSumNormalizePostprocess {
   /// Opts into the block-per-token kernel (provides applyWithAux below).
   static constexpr bool kSupportsBlockPerToken = true;
   static constexpr bool kSupportsLaneOwnedTopK = true;
+  static constexpr bool kNeedsRawScores = true;
 
   /// Needs per-expert aux data (un-biased sigmoid) in the block-per-token
   /// kernel — paired with SigmoidBiasPreprocess, which writes the un-biased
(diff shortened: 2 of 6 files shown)


include/flashinfer/trtllm/fused_moe/RoutingCustomPolicy.cuh
```diff
@@ -435,6 +439,28 @@ struct ScaledSumNormalizePostprocess {
...
+      int32_t const expertIdx = warpTopKExpertIdx[laneIdx];
+      sigmoidScore = sigmoid_accurate(static_cast<float>(ptrScores[expertIdx]));
+    }
+
+    float sum = cg::reduce(warp, sigmoidScore, cg::plus<float>());
+    if (laneIdx < topK) {
+      warpTopKScore[laneIdx] =
+          static_cast<DataType>(sigmoidScore * params.routeScale / (sum + params.sumEpsilon));
+    }
+  }
```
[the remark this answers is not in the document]
alexeldeib: Addressed by enforcing `topK <= numExperts` in `routingCustom::run()` before the raw-score top-k kernels launch. That is the invariant the downstream routing writes also require; guarding only this raw-logit reload would avoid one read but would still leave invalid top-k expert IDs available to later writes. With `topK <= numExperts`, `reduceTopK` has enough real expert indices and its tie-break prefers smaller indices, so padded `>= numExperts` indices are not selected.

tests/moe/test_trtllm_gen_deepseek_sigmoid_bias.py
```diff
@@ -0,0 +1,62 @@
+"""
```
jiahanc: This test seems don't test anything in flashinfer. Can you add the tests that tests the DS no-group sigmoid routing?

alexeldeib: yeah this is slop from local testing, let me fix, apologies

alexeldeib: realized I only pushed the logic not the tests then pushed bad tests -_-

alexeldeib: should be a bit better?

csrc/fused_moe/trtllm_backend/trtllm_fused_moe_routing_custom.cu
```diff
@@ -1501,6 +1501,9 @@ void run(Data const& data, void* stream) {
   TVM_FFI_ICHECK(data.mPtrPermutedIdxSize != nullptr && data.mPtrCtaIdxXyToBatchIdx != nullptr &&
                  data.mPtrCtaIdxXyToMnLimit != nullptr && data.mPtrNumNonExitingCtas != nullptr)
       << "Custom routing kernel expects permuted idx and grouped Gemm launch config buffers";
+  TVM_FFI_ICHECK_LE(data.mTopK, data.mNumExperts)
+      << "Routing kernel expects topK (" << data.mTopK << ") to be <= numExperts ("
+      << data.mNumExperts << ").";
```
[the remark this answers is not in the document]
alexeldeib: Moved the invariant before the precomputed-routing early return in 61f2a854; also adopted the raw-score trait, SM100a gate, and routing-bias dtype doc fixes.

tests/moe/test_trtllm_gen_fused_moe.py
```diff
@@ -1306,6 +1306,145 @@ def test_deepseek_ngroup1_block_per_token_routing(
     )
 
 
+@pytest.mark.parametrize("enable_pdl", [False, True])
+@pytest.mark.parametrize(
+    "num_tokens, num_experts, top_k, hidden_size, intermediate_size",
+    [
+        pytest.param(8, 384, 6, 512, 512, id="array-topk"),
```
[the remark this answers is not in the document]
alexeldeib: Added BF16-aware output tolerances and removed the unreachable bias-subtraction overloads in 98b9b9df.

simon-mo:
Thanks @alexeldeib for the patch. @jiahanc can we please get this in? we are seeing this for GLM as well (same arch) and it is causing nvfp4 glm to generate ! (token id 0) and cause de-generate output.

simon-mo:
Validated this on a standalone DGX B300 (SM103, CUDA 13.0, PyTorch 2.11+cu130).

Source provenance was explicit: I installed each checkout editable, removed the image's packaged `flashinfer-jit-cache==0.6.13+cu130`, used fresh JIT workspaces, and confirmed compilation for `compute_103a`. The image packages `nvrtc.h` under the Python CUDA wheel, so source builds additionally needed `CPATH=/usr/local/lib/python3.12/dist-packages/nvidia/cu13/include`.

Results:

- latest `main` `e9bb80f9fd4194a3a74bc9e1ea9f5d1c8bb77b11`, with only this PR's regression test applied: **4/4 failed** (array TopK and lane-owned HistogramScores, PDL off/on). Selected IDs were valid and outputs finite, but every checked first-column value was `0` instead of `routeScale=2.5`.
- this PR `98b9b9dfe10354a3d91367a12bf1dc919266dddb`: **4/4 passed** on the same B300.
- GLM-shaped standalone repro (`num_tokens=32768, E=256, K=8`, `sigmoid(raw)=2e-7`, bias `12`, PDL on): latest-main forced histogram produced finite all-zero routed output; this PR's histogram path produced the expected `2.5` exactly. Latest-main forced block-per-token also produced `2.5`, confirming the existing workaround.
- In that routing-focused synthetic case, PR histogram median was 0.255 ms vs 0.326 ms for forced block-per-token (30 iterations). This is not a whole-model throughput estimate, but it confirms the fixed default path does not need the workaround for correctness.

This directly confirms the dataflow described here: in FP32, `12 + 2e-7 == 12`, so subtracting the bias loses every selected numerator; the epsilon guard keeps the result finite but cannot restore the weights. Recomputing `sigmoid(raw)` for selected experts restores both affected warp paths.

One CI note: GitHub currently reports the workflows for this head as `action_required` with zero jobs (including PR Test and pre-commit), so they appear to need maintainer approval before running.

jiahanc (approved):
LGTM, thanks for the fix!

PR #3879, opened 2026-07-08
fix: error in blockReduceSumV2 when blockDim.x not a multiple of 32

jdebache:
### Symptom

tests/comm/test_allreduce_unified_api.py::test_allreduce_unified on 4× B300, trtllm backend, hidden_size=2880, fusion=True, for seq_lens [15] (bf16) and [27,11,24,256] (fp16 + bf16). The fused RMS-norm output (output[0]) comes out wrong while the allreduce+residual output (output[1]) is fine. In the log, every printed element of the failing norm output is a uniform ~1.061× of the reference — that ratio is √(9/8) = 1.0607, i.e. the RMS denominator is computed over exactly 8/9 of the row.

### Root cause

Commit dc5d2bc3 changed the block reduction mask in `trtllm_allreduce_fusion.cuh`. When blockDim.x is not a multiple of 32, the last (partial) warp still writes its sum, but the floored mask means nobody reads it. For hidden=2880, fp16/bf16 (VEC_SIZE=8): threads_per_token = 360. The launcher picks cluster_size=8 → 45 threads/block, then doubles to reach ≥128: blockDim=180 (5 full warps + one 20-thread partial warp), cluster=2. The sum-of-squares then covers 2×160×8 = 2560 of 2880 elements → norm output scaled by √(2880/2560) ≈ 1.0607. That matches the observed error to three decimals.

### Proposed fix

Ceiling integer arithmetic, keeping the intent of the optimization: threadIdx.x < ((blockDim.x + 31) >> 5) — in both blockReduceSumV2 and blockReduceMaxV2

## 🔍 Related Issues

No logged issue, fixes CI pipeline failure on B300.

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [X] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [X] I have installed the hooks with `pre-commit install`.
- [X] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

## 🧪 Tests

- [NA] Tests have been added or updated as needed.
- [NA] All tests are passing (`unittest`, etc.).

### Perf A/B: `blockReduceSumV2`/`MaxV2` partial-warp fix (4× B300, TP=4, bf16)

| tokens | hidden | reduction changed | fused old (µs) | fused fixed (µs) | Δ fused | unfused old (µs) | unfused fixed (µs) |
[...]

PR #3880, opened 2026-07-08
fix(comm): make TRT-LLM reductions partial-warp safe

samnordmann:
## 📌 Description

Fix the TRT-LLM block sum and max reductions for block sizes that are not a
multiple of 32.

The original code had two problems for a trailing partial warp:

1. `blockDim.x >> 5` omitted that warp from the second-stage reduction.
2. The warp reduction used `__shfl_xor_sync(0xffffffff, ...)`, although not all
   lanes named in the mask exist in a partial warp. Reads from inactive source
   lanes are undefined by the [CUDA shuffle contract](https://docs.nvidia.com/cuda/cuda-c-programming-guide/#warp-shuffle-functions).

On GB200, the ceiling warp-count fix alone makes a 180-thread sum return 192
instead of 180. This change incorporates that fix and makes both reduction
stages partial-warp safe:

- preserve the existing full-warp fast path;
- use the active-lane mask for the trailing warp;
- ignore XOR partners outside the active-lane range; and
- apply the same handling to sum and max reductions.

The focused CUDA test covers a one-lane tail (129 threads), a 31-lane tail
(159), the observed failure (180), and an aligned control (192).

## 🔍 Related Issues

Builds on and supersedes #3879.

### ✅ Pre-commit Checks

- [x] I have run `pre-commit run --all-files` and fixed reported issues.

## 🧪 Tests

Local:

4x GB200

- public AR+RMSNorm and AR+RMSNorm+dynamic-FP8 validation passed for one-shot
  and two-shot at the partial-warp shape `16x2880`;
- 52 public benchmark validation instances passed across all ranks, including
  CUDA graph replay; and
- same-node exact-base/candidate A/B showed no material performance regression
  (median-of-three deltas from `-2.5%` to `+3.4%`).

Focused reduction test: 4 passed for 129, 159, 180, and 192 threads.

## Reviewer Notes

Official A10G, T4, and H100 jobs are still skipped because this contributor is
not in `@flashinfer-ai/ci-users`. An authorized maintainer must comment
`@flashinfer-bot run` before merge.

### DIFF
--- a/include/flashinfer/comm/trtllm_allreduce_fusion.cuh
+++ b/include/flashinfer/comm/trtllm_allreduce_fusion.cuh
@@ -390,7 +390,8 @@ inline __device__ float reciprocal_approximate_ftz(float a) {
 
 namespace utils {
 
-#define FINAL_MASK 0xffffffff
+static constexpr int kWarpSize = 32;
+static constexpr unsigned int kFullWarpMask = 0xffffffffU;
 
 template <typename T, int NUM>
 __inline__ __device__ T warpReduceSumV2(T* val) {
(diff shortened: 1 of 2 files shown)


samnordmann:
@flashinfer-bot run

aleozlx (approved):

PR #3882, opened 2026-07-08
fix: make mxfp8 gemm test pass by having it quantize along the correct dimension

jdebache:
## 📌 Description

The CI failure (test_bmm_mxfp8[...cudnn...-128-128-128-1], flaky cos-sim ~0.898 vs 0.9) was caused by the test misusing bmm_mxfp8: B quantized along n instead of the reduction dim k, and linear-layout scales fed to a cudnn graph that only reads F8_128x4-swizzled scales. That degraded accuracy from ~0.999 to mean ~0.92, leaving the unseeded test to cross the 0.9 threshold ~2% of the time at the smallest shape.

### Changes

In `flashinfer/gemm/gemm_base.py`, we have added warnings when using unsupported scale layouts. Even though it was previously claimed that 2D non-swizzled / linear was supported, the kernels only ever read 1D 128x4-swizzled scales. Note that we have chosen to NOT throw on these paths, to avoid breaking some unforeseen use-case out there. Nevertheless, as the tests show, 2D non-swizzled / linear result in wrong output.

| Backend | How scale is read | Swizzle | Linear path? | If a wrong (linear/2D) scale is passed |
|---|---|---|---|---|
| **cutlass SM100** | flat ptr + `Sm1xxBlkScaledConfig::tile_atom_to_shape_SF*(problem_shape)`, vec 32; scale strides never read | 128x4 fixed | **no** | warns only |
| **cutlass SM120/121** | flat ptr + compile-time `Sm1xxBlockScaledConfig<32>` atom | 128x4 fixed | **no** | **hard reject** — Python req + C++ `ICHECK_EQ(ndim,1)` |
| **cute-dsl (SM100)** | `make_ptr` + fixed 6D `BlockScaledBasicChunk` layout `((32,4),(32,4))` from `sf_m/sf_n/sf_k` | 128x4, vec 32 | **no** | **hard reject** — `ValueError` if `ndim!=1` |
| **trtllm (SM100/103)** | raw ptr + `SfLayout` **enum** as cubin-selection key; no `Linear` cubin compiled | **A: 128x4 or 8x4; B: 128x4 (+shuffle)** | **no** | warns only |
| **cuDNN** | 3D descriptor + hardcoded `tensor_reordering.F8_128x4` over a 1D swizzled buffer | 128x4 fixed | **no** | mm: reject (`ndim==1`, `use_8x4=False`); bmm: no ndim gate |

In `examples/pytorch/flashinfer_modules.py`, we have corrected the example to use `bmm_mxfp8` by quantizing along the correct axis and dropping the superfluous and possibly mis-interpreted batch dimension from the scales.

In `benchmarks/routines/gemm.py`, we remove the dead code for linear scales as they are never supported and passing correctly swizzled scales on the `auto` backend path.

In `tests/gemm/test_bmm_mxfp8.py` and `tests/gemm/test_mm_mxfp8.py`, we remove the linear scale paths, fix the scale layout on the other paths and tighten the tolerances now that usage is correct.

## 🔍 Related Issues

Example logs of failure in CI:
```
2026-07-05T14:09:41.327317Z 01O tests/gemm/test_bmm_mxfp8.py ........................................... [  7%]
2026-07-05T14:09:41.327318Z 01O ...............................F........................................ [ 19%]
2026-07-05T14:09:41.327319Z 01O .............................sssssssssssssssssssssssssssssssssssssssssss [ 32%]
2026-07-05T14:09:41.327319Z 01O ssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss [ 44%]
2026-07-05T14:09:41.327320Z 01O sssssssssssssssssssssssssssss........................................... [ 57%]
2026-07-05T14:09:41.327333Z 01O ........................................................................ [ 69%]
2026-07-05T14:09:41.327333Z 01O .............................sssssssssssssssssssssssssssssssssssssssssss [ 82%]
2026-07-05T14:09:41.327334Z 01O ssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss [ 94%]
2026-07-05T14:09:41.327334Z 01O sssssssssssssssssssssssssssss                                            [100%]
2026-07-05T14:09:41.327335Z 01O 
2026-07-05T14:09:41.327335Z 01O =================================== FAILURES ===================================
2026-07-05T14:09:41.327336Z 01O E   AssertionError: Cosine similarity 0.8984 is too low (expected > 0.9)
2026-07-05T14:09:41.327337Z 01O     assert tensor(0.8984, device='cuda:0', dtype=torch.bfloat16) > 0.9
2026-07-05T14:09:41.327337Z 01O ----------------------------- Captured stderr call -----------------------------
2026-07-05T14:09:41.327338Z 01O 2026-07-05 05:42:58,019 - INFO - autotuner.py:651 - flashinfer.jit: [Autotuner]: Autotuning process starts ...
2026-07-05T14:09:41.327338Z 01O 2026-07-05 05:42:58,021 - INFO - autotuner.py:674 - flashinfer.jit: [Autotuner]: Autotuning process ends
2026-07-05T14:09:41.327339Z 01O /workspace/flashinfer/tests/gemm/test_bmm_mxfp8.py:74: AssertionError: Cosine similarity 0.8984 is too low (expected > 0.9)
2026-07-05T14:09:41.327340Z 01O =============================== warnings summary ===============================
2026-07-05T14:09:41.327341Z 01O tests/conftest.py:16
2026-07-05T14:09:41.327341Z 01O tests/conftest.py:16
2026-07-05T14:09:41.327341Z 01O   /workspace/flashinfer/tests/conftest.py:16: DeprecationWarning: tcgen05.OperandMajorMode is deprecated, use cute.nvgpu.OperandMajorMode instead
2026-07-05T14:09:41.327342Z 01O     from cutlass.cute.nvgpu.tcgen05 import OperandMajorMode
2026-07-05T14:09:41.327343Z 01O 
2026-07-05T14:09:41.327343Z 01O -- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
2026-07-05T14:09:41.327344Z 01O -- generated xml file: /tmp/junit/tests_gemm_test_bmm_mxfp8.py.687030424.xml ---
2026-07-05T14:09:41.327344Z 01O =========================== short test summary info ============================
2026-07-05T14:09:41.327345Z 01O FAILED tests/gemm/test_bmm_mxfp8.py::test_bmm_mxfp8[True-cudnn-res_dtype0-False-input_dtype0-128-128-256-1] - AssertionError: Cosine similarity 0.8984 is too low (expected > 0.9)
2026-07-05T14:09:41.327346Z 01O assert tensor(0.8984, device='cuda:0', dtype=torch.bfloat16) > 0.9
2026-07-05T14:09:41.327346Z 01O ====== 1 failed, 287 passed, 288 skipped, 2 warnings in 105.05s (0:01:45) ======
2026-07-05T14:09:41.327347Z 01O ❌ FAILED: tests/gemm/test_bmm_mxfp8.py (pytest exit code: 1)
```

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

- [X] All tests are passing (`unittest`, etc.).

## Summary

* **Bug Fixes**
  * Improved MXFP8 GEMM/BMM diagnostics with clearer warnings when operand or scale tensor layouts don’t match expected contiguity/swizzle/column-major requirements.
  * Updated MXFP8 GEMM/BMM trace checking and reference math for 1D 128x4-swizzled E8M0 scales (including decode/unswizzle behavior).
  * Relaxed certain MXFP8 backend eligibility checks to warn instead of failing on non-critical scale-layout mismatches.

* **Documentation**
  * Refreshed MXFP8 MM/BMM documentation and the PyTorch example to match supported swizzled scale layouts.

* **Tests**
  * Increased cosine-similarity thresholds and aligned trace tolerance expectations.


tests/gemm/test_mm_mxfp8.py
danisereb, line 1: Are changes in this file required?
This is a test for `mm_mxfp8` and not `bmm_mxfp8` as far as I remember.

@aleozlx consider removing `bmm_mxfp8`, I don't think vLLM or SGLang use it (but need to confirm).

`mm_mxfp8` is used in **vLLM** (I merged a PR that uses it with ModelOpt MXFP8 models).

jdebache: Yes, effectively the same issue exists there. With the fix, we can tighten the tolerance: see https://github.com/flashinfer-ai/flashinfer/pull/3882/changes#diff-b7c4e1b4ed48ae9eb886c900b42ede8a8fd209a271c6dc02ed8c6fc6eb4f58f3R74

flashinfer/gemm/gemm_base.py
```diff
@@ -5119,11 +5122,11 @@ def mm_mxfp8(
     )
 
     assert a_descale.ndim in (1, 2), (
-        f"mm_mxfp8: a_descale must be 1D (swizzled) or 2D (non-swizzled), "
+        f"mm_mxfp8: a_descale must be 1D (swizzled) or 2D (trtllm shuffled), "
         f"got {a_descale.ndim}D with shape {a_descale.shape}, dtype={a_descale.dtype}"
     )
     assert b_descale.ndim in (1, 2), (
-        f"mm_mxfp8: b_descale must be 1D (swizzled) or 2D (non-swizzled), "
+        f"mm_mxfp8: b_descale must be 1D (swizzled) or 2D (trtllm shuffled), "
```
danisereb: Can you explain what "trtllm shuffled" means?
I assume it's related to use of this MXFP8 GEMM in TRTLLM?

Maybe there's a better name for this type of shuffling ("trtllm shuffled" sounds confusing to me).

tests/gemm/test_mm_mxfp8.py
danisereb: OK, PR LGTM

flashinfer/gemm/gemm_base.py
jdebache: It refers to the 8x4 layout format that the trtllm backend supports.

YangXu1990uiuc:
@jdebache — I looked into whether the cuDNN `bmm_mxfp8` path should **hard-reject** a non-swizzled (linear) scale (the approach in #3455's 2nd commit), since it bears on #3882. Short version: a hard reject isn't the right tool — your warn + doc + drop-the-linear-tests is — but there's a cheap mm/bmm **parity** gap worth a look. Everything below measured on `main` (SM100/B200, cuDNN 9.24), `bmm_mxfp8(backend="cudnn")`.

**What main does today with a linear scale** (b=1, n=k=256):

| scale | M | `A_scale.numel()` vs swizzled-expected | result |
|---|---|---|---|
| swizzled | 128 / 200 | matches | OK, cos 0.999 |
| **linear** | 128 | **1024 == 1024** | finite but degraded, **cos 0.923** |
| **linear** | 200 | 1600 != 2048 | out-of-bounds scale read → NaN |

cuDNN never raises — it's silently degraded at 128-aligned M and an OOB scale read (NaN observed) at non-aligned M. The linear M=128 cos is **0.923**, which passes the old `min_cos_sim=0.9` but fails your new `0.99` — so the threshold bump is what actually flags this at test time.

**Why a numel-based reject can't fully fix it.** `mxfp8_quantize` returns a 1D buffer for both layouts, and (for k a multiple of 128) `swizzled_len = pad(m,128)·(k/32)` vs `linear_len = m·(k/32)` — equal **iff `(b·m) % 128 == 0`** (the quantizer pads the flattened b·m dim, so that's the exact blind spot). A length check therefore passes the bad scale through in exactly the silently-degraded case and only fires where the output is already NaN. Linear vs swizzled 1D scales are indistinguishable by dtype/rank/length at aligned sizes, so "reject non-swizzled" can't be made watertight — I don't think the hard reject is worth adding (false confidence, and it conflicts with your warn-not-reject choice).

**One thing #3882 leaves open, if you want it.** `mm_mxfp8` already validates the 1D scale length (`_check_mm_mxfp8_problem_size`, gemm_base.py ~L4447); the **bmm** cuDNN path doesn't. And #3882's `_warn_mxfp8_gemm_strides` checks operand *strides* (row/col-major), which is orthogonal to the *scale-layout* mistake — a correctly-strided call with a linear scale trips nothing. So a small **warn** mirroring mm's length check would turn the non-aligned-M NaN into an actionable message and restore mm/bmm parity (it still can't catch the aligned-M silent case — that's what the doc + 0.99 threshold are for). Happy to add it here or fold into #3882; not a blocker, your call.

**On #3455:** I'm closing it as superseded — the override-shape fix already landed via #3489, and commit-2's reject is the incomplete approach above. The one piece worth keeping is its **dynamic-M regression test** (`test_bmm_mxfp8_cudnn_dynamic_m`, M ∈ {130,200,257,384,1000}) — green on main, and it guards the #3489 fix, which your `test_bmm_mxfp8` doesn't exercise (you keep M ∈ {128,256,512}). I'll fold it into #3882 or send a tiny standalone PR, whichever avoids a conflict.

jdebache:
Should also fix:
- https://github.com/flashinfer-ai/flashinfer/issues/4090

flashinfer/gemm/gemm_base.py
```diff
@@ -2935,6 +2936,45 @@ def execute_cudnn_gemm_fp4_graph_override_shape(
...
+            "is most likely wrong, please check accuracy."
+        )
+    if b.stride(-2) != 1:
+        jit_logger.warning_once(
+            f"{backend} mxfp8 GEMM expects B to be column-major [batch, k, n] "
+            f"(K contiguous); got b.stride()={tuple(b.stride())}. Quantize the "
+            "contiguous [b, n, k] weight and pass the transpose of the "
+            "quantized tensor, e.g. B = mxfp8_quantize(weight)[0].transpose(-2, -1). "
+            "The result is most likely wrong, please check accuracy."
+        )
```
dhiraj113, line 2950: Wouldn't it be better to add this as backedn supportedness check and return not supported instead of a warning like this?

flashinfer/gemm/gemm_base.py
```diff
@@ -5125,19 +5182,22 @@ def mm_mxfp8(
...
-          where M_padded = round_up(m, 8 if 8x4 layout else 128), K_padded = round_up(k // 32, 4)
-        dtype: uint8.
+        Block scale tensor for A, uint8 (fp8 e8m0), 1D swizzled layout:
+        shape (M_padded * K_padded,) where M_padded = round_up(m, 8 if 8x4
+        layout else 128) and K_padded = round_up(k // 32, 4). Produced by
+        ``mxfp8_quantize(..., is_sf_swizzled_layout=True)``. The 8x4 layout
+        (``use_8x4_sf_layout=True``) is only consumed by the trtllm backend.
+        2D linear scales are misinterpreted by every backend: SM12x CUTLASS
+        rejects them; the other backends emit a warning and produce results
+        that are most likely wrong.
```
dhiraj113: would a future TODO be to reject these cases?

tests/gemm/test_bmm_mxfp8.py
```diff
@@ -26,8 +27,13 @@ def test_bmm_mxfp8(
         pytest.skip("bmm_mxfp8 cudnn backend requires SM10x.")
     if backend == "cutlass" and compute_capability[0] != 12:
         pytest.skip("bmm_mxfp8 cutlass backend requires SM12x.")
-    if backend == "cutlass" and not is_sf_swizzled_layout:
-        pytest.skip("bmm_mxfp8 cutlass backend on SM12x only supports swizzled layout.")
+    if not is_sf_swizzled_layout:
+        pytest.skip(
+            "bmm_mxfp8 backends read scale factors in the F8_128x4 swizzled "
+            "layout; linear-layout scales would be misinterpreted."
```
dhiraj113: Is skipping the test the right way to handle this? Shouldn't the api return unsupported for such cases?

dhiraj113 (approved):

flashinfer/gemm/gemm_base.py
jdebache: That was my first instinct, but I was worried to change the behaviour in case someone, somehow, uses the API this way and is fine with it. I can't imagine why at this time, but...

flashinfer/gemm/gemm_base.py
jdebache: Same as above.

flashinfer/gemm/gemm_base.py
dhiraj113: I think it should be okay to raise, especially given that the warning says the results are likely wrong. It's also okay if you want to differ raising the error for later.

aleozlx:
seems CI caught an error on h100

https://github.com/flashinfer-ai/flashinfer/actions/runs/30298598108/job/90085814969?pr=3882

```
=================================== FAILURES ===================================
____________________ test_trtllm_warns_for_linear_b_descale ____________________

monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x7877b72649b0>

    def test_trtllm_warns_for_linear_b_descale(
        monkeypatch: pytest.MonkeyPatch,
    ):
        messages: list[str] = []
>       monkeypatch.setattr(gemm_base.jit_logger, "warning_once", messages.append)
                            ^^^^^^^^^^^^^^^^^^^^
E       AttributeError: module 'flashinfer.gemm.gemm_base' has no attribute 'jit_logger'

tests/gemm/test_mm_mxfp8.py:526: AttributeError
```

jdebache:
My bad, made a small change after starting the local tests... Fixing it and doing a full local run again.

bkryu:
Both internal and external CI results look clean. Merging PR

PR #1951, opened 2025-10-20
bugfix: fix failed unittest `test_green_ctx` and `test_jit_example` on spark (sm_121)

yzh119:
## 📌 Description

There are three failed unittests on spark (sm_121):
* tests/utils/test_green_ctx.py
* tests/utils/test_jit_example.py
* tests/utils/test_sampling.py

First one is because spark has small number of SMs (48) and we don't have a guard on green context splitting.
Second one is an unknown issue (logits don't match with reference) and probably related to barriers on sm_121, xfail now and will fix later.

The last one will be fixed by another PR from @bkryu , this PR fixes the first two issues.

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [x] Tests have been added or updated as needed.
- [ ] All tests are passing (`unittest`, etc.).

## Summary by CodeRabbit

* **Tests**
  * Tests now pre-check GPU resources and auto-skip with informative messages including available and requested SM counts to avoid spurious failures.
  * Added a conditional xfail for GPUs with compute capability 12.1 to avoid false negatives on that hardware.
  * Tightened a sampling test by adding a relative tolerance for more robust numerical validation.

* **Bug Fixes**
  * Improved runtime error handling to surface clearer guidance when GPU SM resources are insufficient.

PR #3890, opened 2026-07-09
fix: skip LogitsTransform on lanes beyond the split-KV chunk boundary in FA2 kernels

yichengj0:
## 📌 Description

Resolves item 11 of the SM121 support audit (#3170): re-verify the SM121 `xfail` on
`tests/utils/test_jit_example.py::test_dump_logits`, added in #1951 when the cause was
unknown.

Re-verified on DGX Spark hardware: the test still fails, but not because of SM121
numerics. The cause is a split-KV bug in the FA2 kernels that Spark's SM count happens
to expose:

- With split-KV, a CTA's kv chunk can end in the middle of a KV tile. For the lanes
  past the chunk end, the kernel invoked the attention variant's `LogitsTransform`
  hook before the mask discarded them. Attention output stays correct, but a variant
  with side effects (such as the `DumpLogits` JIT example, which writes logits to
  global memory) publishes stale values for kv positions its CTA doesn't own.
- Most GPUs get tile-aligned chunks, so those lanes never exist and CI never saw this.
  Spark's 48 SMs produce a misaligned chunk size for the test shape.

Fix: `logits_transform` now receives the chunk boundary and skips lanes beyond it.
Kernel output is unchanged, since those lanes were always masked out right after; only
what the hook observes changes. The test now passes on Spark, so the xfail is removed.

## 🔍 Related Issues

#3170 (item 11), #1951 (introduced the xfail).

## 🧪 Tests

On DGX Spark (GB10, SM121, 48 SMs, CUDA 13, aarch64):

- `test_dump_logits`: fails before the fix (2.1% of dumped logits wrong,
  deterministic), passes after. A stricter NaN-prefilled variant of the check confirms
  every logits slot is written correctly.
- `tests/utils/test_jit_example.py`: 9 passed, 1 skipped (sm90a-only).
- Regression slices over the touched kernels, all passing: single prefill with long
  unaligned kv lengths (48 tests), sliding-window prefill (1868 tests).
- The persistent kernel (`test_batch_attention.py`) is a pre-existing xfail on SM12x,
  so it cannot be exercised on this machine. The `persistent.cuh` change was
  compile-verified for `sm_100a` with `logits_soft_cap` enabled, which instantiates
  the changed call sites.
- `pre-commit run` on changed files: clean.

## Reviewer Notes

- No public API change. The JIT customization interface (`REGISTER_LOGITS_TRANSFORM`,
  the variant's `LogitsTransform` signature) is untouched. `logits_transform` is an
  internal helper referenced only by `prefill.cuh` and `persistent.cuh`; #776 changed
  `logits_mask` the same way when it added `chunk_end` there.
- An alternative fix, rounding the chunk size up to a tile multiple, was rejected
  because it changes work distribution across chunks on all architectures. The chosen
  change leaves numerics untouched.
- Cost is one integer compare per fragment lane, mirroring index math `logits_mask`
  already does.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved attention logits processing to respect end-of-range boundaries, preventing stale/out-of-range intermediate values from being read or written in prefill and masked attention paths.
  * Updated masked and drain logits transformations to pass correct end-position information for more consistent behavior.
* **Tests**
  * Removed a conditional expected-failure for the `test_dump_logits` scenario on a specific hardware configuration so it runs normally.


saltyminty (approved):
Approved pending CI

saltyminty:
/bot run tests/utils/test_jit_example.py tests/attention/test_single_prefill.py tests/attention/test_batch_prefill_kernels.py

kahyunnam (approved):

PR #3834, opened 2026-07-04
fix(sparse-mla-sm120): add TOPK=256 DSV4 prefill instantiation

waynehacking8:
## Description

The SM120 sparse-MLA DSV4 prefill dispatch (`dispatch_dsv4_single` in `csrc/sparse_mla_sm120_prefill.cu`) only instantiates `topk in {128, 512, 1024, 2048}`, so `topk=256` (a valid `4 x BI(64)` shape, hit by DeepSeek-V4 + DSpark configs) fails with `Unsupported sparse-MLA prefill configuration: model=DSV4 num_heads=32 topk=256 ...`. This adds the missing `topk == 256` branch plus test coverage.

This is the prefill half of #3828; PR #3817 covers the decode half. No overlap: this PR touches only the prefill `.cu`, and the test configs are distinct from #3817's decode entries in the same test file (trivial rebase for whichever lands second).

Compute-mode choice: BF16, not FP8. Measured on an RTX PRO 6000 (SM120, CUDA 13), BF16 beats FP8 at `topk=256` for every shape tested, consistent with the existing small-K-loop comment (the FP8 Q-quantize prologue doesn't amortise at K=256):

| num_heads | num_tokens | FP8 median | BF16 median |
|---|---|---|---|
| 32 | 128 | 53.7 us | 51.0 us |
| 32 | 1024 | 278.5 us | 254.0 us |
| 32 | 4096 | 991.2 us | 910.6 us |
| 128 | 128 | 141.3 us | 127.0 us |
| 128 | 1024 | 958.5 us | 862.2 us |
| 128 | 4096 | 4442.1 us | 4330.8 us |

Verified on the same card: the issue's prefill repro (num_heads=32, topk=256, num_tokens=128) fails on current main and runs + matches the reference with this change.

## Related Issues

Prefill portion of #3828 (decode portion: #3817).

### Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

## Tests

- [x] Tests have been added or updated as needed.
- [x] All tests are passing (`unittest`, etc.): new configs `(16, 256)` and `(32, 256)` exercise the NHG=1 and NHG=2 routes; full `tests/attention/test_sparse_mla_sm120.py` = 145 passed on SM120.

## Reviewer Notes

[...]

PR #3896, opened 2026-07-09
fix(sparse-mla-sm120): add topk=256 prefill-dsv4 dispatch (fixes #3828)

### LINKED ISSUE #3828: [SM120] Sparse-MLA DSV4 decode+prefill kernels not instantiated for topk=256
## Summary

The SM120 sparse-MLA **DSV4** kernels are only instantiated for `topk ∈ {128, 512, 1024, 2048}`. An attention call with **`topk=256`** has no kernel in either the decode or the prefill path and fails:

**decode** (`num_tokens ≤ 64`) — `_paged_attention` in `flashinfer/mla/_sparse_mla_sm120.py` finds `(num_heads, topk)` absent from `_DECODE_DSV4_DISPATCH`, so it falls through to the prefill orchestrator (`csrc/sparse_mla_sm120.cu`), which is decode-rejecting:
```
Check failed: num_tokens > 64 (7 vs. 64) : Decode (num_tokens <= 64) must go through
sparse_mla_sm120_decode_dsv3_2 or sparse_mla_sm120_decode_dsv4; got num_tokens=7
```

**prefill** (`num_tokens > 64`) — `sparse_mla_prefill_dispatch` → `dispatch_dsv4_single` in `csrc/sparse_mla_sm120_prefill.cu` has no `topk == 256` branch and returns false:
```
Unsupported sparse-MLA prefill configuration: model=DSV4 num_heads=32 topk=256
page_block_size=64 topk_extra=0 extra_page_block_size=0
```

`topk=256` is a valid shape (`256 = 4 × BI(64)`, between the instantiated 128 and 512); it's simply missing from the instantiation grids.

## Repro / context

Serving DeepSeek-V4 + DSpark speculative decoding on vLLM, SM120 (RTX PRO 6000, TP=2). The DSpark draft's sliding-window (`swa_only`) layers issue sparse-MLA attention with `num_heads=32` (64 heads / TP2), `topk=256`, `page_block_size=64`, `d_qk=512`, `extra_topk=0`. Warmup hits the decode path (`num_tokens ≤ 64`) → the fallthrough assert; the flashinfer autotune dummy run hits the prefill path (`num_tokens > 64`) → the unsupported-config error.

## Fix

Adding the `topk=256` instantiations resolves it end-to-end (verified serving):

- `csrc/sparse_mla_sm120_decode_dsv4.cu` — `DSV4_DISPATCH(H, 256)` for `H ∈ {8,16,32,64,128}`
- `flashinfer/mla/_sparse_mla_sm120.py` — the `(H, 256)` pairs in `_DECODE_DSV4_DISPATCH`
- `csrc/sparse_mla_sm120_prefill.cu` (`dispatch_dsv4_single`) — `else if (topk == 256) DISPATCH_BY_NH_CM(FP8, 256);` (FP8 per the existing perf-mode comment; smem-safe since FP8-512 already builds)

Tested against main @ `b5ac097e`.

OsakaTX:
## 📌 Description

Implements the prefill half of #3828, exactly as proposed there by @eous: the
SM120 sparse-MLA DSV4 **prefill** path (`dispatch_dsv4_single` in
`csrc/sparse_mla_sm120_prefill.cu`) has no `topk == 256` branch — the dispatch
chain covers {128, 512, 1024, 2048} — so a prefill-shaped call
(`num_tokens > 64`) with `topk=256` returns false and fails:

```
Check failed: (ok) is false: Unsupported sparse-MLA prefill configuration:
model=DSV4 num_heads=32 topk=256 page_block_size=64 topk_extra=0 extra_page_block_size=0
```

As #3828 notes, `topk=256` is a valid shape (4 × BI(64), between the
instantiated 128 and 512); it's simply missing from the grid. The shape occurs
in practice via DeepSeek-V4-Flash-DSpark under tensor-parallel=2 — the decode
half of the same gap is addressed by #3817.

### Change

- `csrc/sparse_mla_sm120_prefill.cu`: `else if (topk == 256) DISPATCH_BY_NH_CM(FP8, 256);`
  in `dispatch_dsv4_single` — FP8 compute mode per the existing comment (the
  BF16 QK fast path is reserved for the small K-loop at 128; FP8-512 already
  builds, so the smaller 256 tile is smem-safe).
- `tests/attention/test_sparse_mla_sm120.py`: add `(16, 256)` and `(32, 256)`
  to `_DSV4_PREFILL_CONFIGS` (16 exercises the MG_N_HG_T=1 routing, 32 is the
  motivating TP=2 shape).

## 🧪 Testing

Isolation A/B on GB10 (sm_121a), CUDA 13 — identical container and environment,
module JIT-built from source in both legs, only the two dispatch lines varied.
This complements the RTX PRO 6000 (SM120) serving validation reported in #3828 —
two architectures, same result:

- **A (unmodified main + the new test configs):** exactly the 8 `topk=256`
  prefill combos fail (`num_tokens ∈ {128, 256}` × `with_sink ∈ {False, True}` ×
  NH ∈ {16, 32}) with the error above; all 36 pre-existing prefill combos pass.
- **B (this branch):** all prefill combos pass, new shapes verified against the
  bf16 reference implementation.

## 🔗 Related

- Fixes #3828 (prefill half; analysis and fix design are @eous's — this PR
  implements it with test coverage and additional hardware validation)
- #3817 — the decode half of the same missing-topk=256 gap (independently
  confirmed on my hardware in that thread)

## Summary by CodeRabbit

* **New Features**
  * Expanded support for additional prefill configurations, including `topk=256`, for certain sparse attention workloads.
* **Tests**
  * Added test coverage for two more head-count and `topk=256` combinations to improve validation of the new supported settings.

### DIFF
--- a/csrc/sparse_mla_sm120_prefill.cu
+++ b/csrc/sparse_mla_sm120_prefill.cu
@@ -293,6 +293,8 @@ inline bool dispatch_dsv4_single(int num_heads, int topk, const bf16* Q, const u
   // amortises FP8's higher Tensor-Core throughput.
   if (topk == 128)
     DISPATCH_BY_NH_CM(BF16, 128);
+  else if (topk == 256)
+    DISPATCH_BY_NH_CM(FP8, 256);
   else if (topk == 512)
     DISPATCH_BY_NH_CM(FP8, 512);
   else if (topk == 1024)
--- a/tests/attention/test_sparse_mla_sm120.py
+++ b/tests/attention/test_sparse_mla_sm120.py
@@ -890,6 +890,8 @@ def test_sparse_mla_sm120_prefill_glm_nsa_arbitrary_fp32(num_heads: int) -> None
 
 _DSV4_PREFILL_CONFIGS = [
     (16, 128),
+    (16, 256),
+    (32, 256),
     (32, 512),
     (64, 1024),
     (128, 1024),


waynehacking8:
Heads up: #3834 (opened 07-04) implements this same prefill-half of #3828 -- same dispatch branch and the same (16,256)/(32,256) test configs -- but with BF16 CM instead of FP8, picked from measurement on an RTX PRO 6000 (BF16 beat FP8 by 2.5-9% at all six shapes I benched; the small-K-loop prologue argument in the PR body). ormandj's multi-day SM120 serving soak upthread on #3817 ran with the BF16 version of this branch. Not territorial about which PR lands, but the compute-mode choice is a real fork in the road and probably deserves a deliberate call by the maintainers.

OsakaTX:
> Heads up: #3834 (opened 07-04) implements this same prefill-half of #3828 -- same dispatch branch and the same (16,256)/(32,256) test configs -- but with BF16 CM instead of FP8, picked from measurement on an RTX PRO 6000 (BF16 beat FP8 by 2.5-9% at all six shapes I benched; the small-K-loop prologue argument in the PR body). ormandj's multi-day SM120 serving soak upthread on #3817 ran with the BF16 version of this branch. Not territorial about which PR lands, but the compute-mode choice is a real fork in the road and probably deserves a deliberate call by the maintainers.

Good catch, and I appreciate the friendliness. Haven't submitted many PRs before, mostly do private dev work. Agree the CM choice is the real decision and it belongs to the maintainers. Your RTX PRO 6000 data is compelling for sm_120. One datapoint is missing: sm_121a (GB10/DGX Spark), which is where the (32,256) shape actually bites in practice (DSv4 DSpark under vLLM TP=2). I'll try and bench both CM variants at the six shapes on GB10 hardware and post numbers here within a day or two. If BF16 holds on sm_121a as well, #3834 is the better patch and I'm happy to close this one in its favor.  The fix landing matters more than whose branch carries it.

waynehacking8:
sm_121a numbers would genuinely complete the picture -- it's a different chip and a different cubin variant (121a vs 120f), so I wouldn't assume my sm_120 ordering transfers. And the goodwill cuts both ways: if FP8 wins on GB10 I'm happy to flip #3834's CM or split it per-arch, whichever the maintainers prefer. Looking forward to your numbers.

OsakaTX:
Closing this in favor of #3834. I benchmarked both compute modes for the `topk == 256` DSv4 prefill dispatch on GB10 (sm_121a) and BF16 is the better default.

**Method.** #3834 and this PR differ by exactly one token in the dispatch (`DISPATCH_BY_NH_CM(BF16, 256)` vs `FP8`). To isolate that single variable with no base-commit confounds, I built both on the same checkout, forcing a clean JIT recompile from source each time (uninstalled `flashinfer-jit-cache` and `flashinfer-cubin`). KV input is identically packed via `quantize_kv_dsv4` in both, so this measures the compute-mode path only, not KV memory. Timing is CUDA-event based, 60 warmup + 5x300 iters, median of reps (per-config std < 1 us). Both builds pass the `(16, 256)` and `(32, 256)` prefill correctness tests on sm_121a.

**Result (BF16 time saved vs FP8, positive = BF16 faster):**

| heads | prefill tokens | full (256 valid idx) | half (128 valid, test setup) |
|---|---|---|---|
| 16 | 128 | +12.8% | +12.6% |
| 16 | 256 | +10.1% | +9.3% |
| 16 | 512 | +13.1% | +1.8% |
| 16 | 1024 | +5.3% | ~0% |
| 16 | 2048 | +4.3% | +1.3% |
| 32 | 128 | +19.1% | +19.2% |
| 32 | 256 | +16.4% | +15.9% |
| 32 | 512 | +1.8% | +1.1% |
| 32 | 1024 | +0.2% | +1.3% |
| 32 | 2048 | +1.9% | -0.2% |

BF16 is faster than or equal to FP8 in every configuration. The margin is largest at short prefills (about 10 to 19% at 128 to 256 tokens), where the kernel is latency-bound and the FP8 dequant path costs the most, and it narrows toward parity at long prefills (about 0 to 5% at 1024 to 2048 tokens) as the kernel becomes bandwidth-bound. FP8 shows no throughput advantage, so #3834 is the right choice for this dispatch on GB10. Thanks @waynehacking8.

PR #3899, opened 2026-07-09
jit: memory-aware default for _get_num_workers() (fixes #3634)

### LINKED ISSUE #3634: JIT compile of SM120/SM121a CUTLASS FP4 kernels OOMs on first forward pass — ninja default parallelism unsafe on unified-memory systems
## Environment

| | |
|---|---|
| Hardware | NVIDIA DGX Spark, GB10 (Grace Blackwell), ARM64 |
| Compute Capability | SM121a (12.1a) — uses SM120 CUTLASS kernels |
| Memory | 121 GiB unified memory (CPU + GPU share the same pool) |
| CUDA | 13.0 |
| FlashInfer | 0.6.12 |
| vLLM | 0.23.0 (V1 engine, EngineCore subprocess) |
| Model | NvFP4-quantized MoE — `saricles/Qwen3-Coder-Next-NVFP4-GB10` |

## What happens

On the first forward pass (triggered by vLLM's KV-cache profiling dummy run), FlashInfer JIT-compiles two large CUTLASS FP4 GEMM kernel sets for SM121a:

- `fused_moe_120` — **85 CUDA source files**
- `fp4_gemm_cutlass_sm120` — linear layers

`ninja` is invoked without a `-j` cap, so it uses its platform default (= number of CPU cores = 20 on the DGX Spark). Each `nvcc` process compiles a dense CUTLASS kernel and peaks at **~3–8 GiB RAM**. With 20 parallel jobs that is up to **160 GiB** — far exceeding available memory even on a 121 GiB unified-memory system (more so when the model is already partially loaded).

Result: the Linux kernel OOM-kills nvcc workers:

```
ninja: build stopped: subcommand failed.   # exit code 137 (SIGKILL) / 143 (SIGTERM) / 4 (nvcc-internal)
NVRM: nvCheckOkFailedNoLog: Check failed: Out of memory [NV_ERR_NO_MEMORY]
```

The build fails silently. With vLLM the EngineCore subprocess crashes mid-init with no actionable error — only `Engine core initialization failed. Failed core proc(s): {}` surfaces at the vLLM level.

## Root cause

`cpp_ext.py::_get_num_workers()` returns `None` when `MAX_JOBS` is not set, which causes `run_ninja()` to omit the `-j` flag entirely. ninja then defaults to using all available CPU cores. There is no fallback cap or memory-aware limit.

```python
def _get_num_workers() -> Optional[int]:
    max_jobs = os.environ.get("MAX_JOBS")
    if max_jobs is not None and max_jobs.isdigit():
        return int(max_jobs)
    return None   # ← ninja falls back to os.cpu_count() = OOM on unified-memory systems
```

## Workaround (confirmed working)

Pre-compile the `.so` files **before** loading the model (so ~74 GiB is free) with `MAX_JOBS=4`:

```bash
MAX_JOBS=4 python -c "
import flashinfer
# trigger JIT compile of SM120 CUTLASS ops
"
```

With `MAX_JOBS=4`: 4 × 8 GiB = 32 GiB ≪ 74 GiB available → completes cleanly in ~40 min. The cached `.so` files are reused on subsequent runs without recompilation.

`FLASHINFER_NVCC_THREADS` (controls threads *within* each nvcc invocation) is orthogonal and does not help here — the problem is the *number of parallel nvcc processes*, not the per-process thread count.

## Suggested fix

Add a safe default in `_get_num_workers()` — either a fixed conservative fallback, or a memory-aware estimate:

```python
def _get_num_workers() -> Optional[int]:
    max_jobs = os.environ.get("MAX_JOBS")
    if max_jobs is not None and max_jobs.isdigit():
        return int(max_jobs)
    # Safe fallback: avoid OOM on unified-memory / memory-constrained systems.
    # SM120 CUTLASS kernels can peak at ~8 GiB per nvcc process.
    try:
        import psutil
        available_gib = psutil.virtual_memory().available / (1024 ** 3)
        return max(1, int(available_gib // 10))  # conservative: 10 GiB headroom per job
    except ImportError:
        return None  # keep existing behaviour if psutil not available
```

Alternatively: document `MAX_JOBS` prominently in the installation / first-run documentation for SM120/SM121a users.

## Notes

- The `.so` files produced by a successful pre-compile (with `MAX_JOBS=4`) work correctly at runtime — the OOM only occurs during the *build* phase, not during inference.
- This affects any system where unified memory or an already-resident model leaves less than ~160 GiB free, which will include most DGX Spark deployments.
- Tested on: FlashInfer 0.6.12 inside LocalAI v4.4.3 `l4t-cuda13-arm64` container (vLLM 0.23.0 backend).

pos-ei-don:
## Summary

`_get_num_workers()` currently returns `None` when `MAX_JOBS` is unset, which lets ninja fall back to `cpu_count() + 2` parallel jobs. Large CUTLASS translation units (e.g. the SM120 NVFP4 GEMM kernels this repo builds by default) peak at ~8 GiB RSS per nvcc, so a 22-way parallel JIT compile tries to allocate ~176 GiB of anonymous memory. On memory-constrained hosts and especially on unified-memory systems (DGX Spark GB10: 121 GiB shared between CPU and GPU, 20 cores, plus a possibly-resident model), this either OOM-kills nvcc mid-build or — with `kernel.hardlockup_panic=0`, which is the default on many ARM64 distros — sends the kernel into a hard lockup that only a hardware watchdog can recover from.

This PR ports the exact same guard that already lives in `scripts/jit_cache_build_common.sh::compute_jit_cache_parallelism()` (added in #3205) from the AOT path to the runtime JIT path. Same formula, same 8 GB/job baseline, same idea — just applied where end-users hit it.

Closes #3634.

## Formula

```
budget = max(0, (MemAvailable_GiB - headroom_GiB) // mem_per_job_GiB)
num_workers = max(1, min(cpu_count, budget))
```

Defaults (overridable via env):

- `FLASHINFER_JIT_JOB_MEMORY_GB=8` — matches the AOT baseline in `jit_cache_build_common.sh`
- `FLASHINFER_JIT_MEMORY_HEADROOM_GB=16` — for OS + a possibly-resident model when the JIT runs mid-serving

Explicit `MAX_JOBS=N` still wins, preserving the existing behaviour for callers that already tuned it. When memory cannot be determined (`/proc/meminfo` unreadable and no `SC_AVPHYS_PAGES`), `_get_num_workers()` returns `None` as before, so ninja's `cpu_count + 2` default kicks in on those platforms.

## Impact examples

| Host | `MemAvailable` | Jobs before | Jobs after |
|------|---------------:|------------:|-----------:|
| DGX Spark GB10 (20 cores, 121 GiB UMA) | 105 GiB | 22 (ninja default) | **11** |
| DGX Spark GB10, model resident | 40 GiB  | 22                 | **3**  |
| Workstation (16 cores, 64 GiB) | 60 GiB  | 18                 | **5**  |
| Small VM (8 cores, 16 GiB)     | 12 GiB  | 10                 | **1**  |

Values are computed against `_get_available_memory_gib()` at ninja-start time. `NVCC_THREADS` (`--threads=N` per nvcc) stays untouched; it is orthogonal.

## Test plan

- [x] `python -c "import ast; ast.parse(open('flashinfer/jit/cpp_ext.py').read())"` — syntax OK
- [x] Manual unit-check with monkey-patched `/proc/meminfo` for 105 / 40 / 12 GiB → 11 / 3 / 1 jobs
- [ ] Live rebuild on GB10 after the fix — I will follow up in the thread once I have a controlled repro slot

## Reproduction of the pre-patch failure

Today I purged one CUTLASS FP4 kernel directory (`~/.cache/flashinfer/0.6.12/121a/cached_ops/fp4_gemm_cutlass_sm120/`) on my DGX Spark to force a targeted recompile — nothing dramatic, just the ~85 kernels in that one module. With the current default MAX_JOBS behaviour, ninja fanned out enough concurrent nvcc processes that the box went into a hard lockup within ~10 minutes and was recovered by the ARM SBSA hardware watchdog (`ACPI GTDT: found 1 SBSA generic Watchdog(s)` in the boot log). No panic message, no OOM in `journalctl -b -1` — the classic `hardlockup_panic=0` failure mode. `MAX_JOBS=4` avoids it manually, but that is exactly the knowledge users should not need to have for a first-run JIT.

## Notes for review

- I kept the environment variable naming under the `FLASHINFER_JIT_*` prefix that already exists for JIT-related knobs (`FLASHINFER_JIT_VERBOSE`, `FLASHINFER_JIT_DEBUG`, etc.).
- No `psutil` dependency added — `/proc/meminfo` + `os.sysconf("SC_AVPHYS_PAGES")` cover Linux and most other POSIX systems without pulling in a new requirement.
- Deliberately did not add a hard cap. The AOT script uses `nproc` as its ceiling and I mirrored that — a hard cap here would also slow down the maintainers' AOT builds if they later share this helper.
- Non-Linux fallback: `os.sysconf("SC_AVPHYS_PAGES")` returns free-not-available, so it under-schedules slightly on macOS / BSD. That is on the safe side for a JIT default.

## Summary by CodeRabbit

- **New Features**
  - Enhanced JIT build parallelism so it scales more accurately with available system memory and CPU capacity.
  - Added smarter handling for concurrency limits, including configurable memory-based budgeting and headroom to reduce the risk of unstable builds.
- **Bug Fixes**
  - Improved fallback behavior when environment settings are missing or invalid, ensuring the build can still proceed safely (including limiting to a single job when necessary).

### DIFF
--- a/flashinfer/jit/cpp_ext.py
+++ b/flashinfer/jit/cpp_ext.py
@@ -366,9 +366,13 @@ def _get_available_memory_gib() -> Optional[float]:
     except (OSError, ValueError, IndexError):
         pass
     try:
-        return (os.sysconf("SC_AVPHYS_PAGES") * os.sysconf("SC_PAGE_SIZE")) / (1024**3)
+        avphys_pages = os.sysconf("SC_AVPHYS_PAGES")
+        page_size = os.sysconf("SC_PAGE_SIZE")
+        if avphys_pages > 0 and page_size > 0:
+            return (avphys_pages * page_size) / (1024**3)
     except (AttributeError, OSError, ValueError):
-        return None
+        pass
+    return None
 
 
 def _get_num_workers() -> Optional[int]:
--- a/flashinfer/jit/cpp_ext.py
+++ b/flashinfer/jit/cpp_ext.py
@@ -341,11 +341,62 @@ def generate_ninja_build_for_op(
     return "\n".join(lines)
 
 
+def _env_int(name: str, default: int, minimum: int) -> int:
+    value = os.environ.get(name)
+    if value is None:
+        return default
+    try:
+        parsed = int(value)
+    except ValueError:
+        logger.warning("Ignoring invalid %s=%r; using %d.", name, value, default)
+        return default
+    if parsed < minimum:
+        logger.warning("Ignoring %s=%r (< %d); using %d.", name, value, minimum, default)
+        return default
+    return parsed
+
+
+def _get_available_memory_gib() -> Optional[float]:
+    """Best-effort available system memory in GiB; None if undetermined."""
+    try:
+        with open("/proc/meminfo") as f:
+            for line in f:
+                if line.startswith("MemAvailable:"):
+                    return int(line.split()[1]) / (1024**2)  # kB -> GiB
+    except (OSError, ValueError, IndexError):
+        pass
+    try:
+        return (os.sysconf("SC_AVPHYS_PAGES") * os.sysconf("SC_PAGE_SIZE")) / (1024**3)
+    except (AttributeError, OSError, ValueError):
+        return None
+
+
 def _get_num_workers() -> Optional[int]:
     max_jobs = os.environ.get("MAX_JOBS")
     if max_jobs is not None and max_jobs.isdigit():
         return int(max_jobs)
-    return None
+    # Without -j, ninja defaults to cpu_count()+2 parallel jobs. Large CUTLASS
+    # translation units peak at multiple GiB of RSS per nvcc process (the AOT
+    # build in scripts/jit_cache_build_common.sh budgets 8 GB/job for the same
+    # reason), which OOMs or hard-locks memory-constrained and unified-memory
+    # systems (e.g. DGX Spark GB10: 20 cores sharing 121 GiB with the GPU,
+    # where the model may already be resident when JIT compilation starts).
+    # Mirror the AOT logic: budget per-job memory against MemAvailable, keep
+    # headroom for concurrent allocations, and never exceed the CPU count.
+    mem_per_job_gib = _env_int("FLASHINFER_JIT_JOB_MEMORY_GB", 8, 1)
+    headroom_gib = _env_int("FLASHINFER_JIT_MEMORY_HEADROOM_GB", 16, 0)
+    available_gib = _get_available_memory_gib()
+    if available_gib is None:
+        return None  # cannot determine memory: keep previous behavior
+    budget_jobs = int(max(0.0, available_gib - headroom_gib) // mem_per_job_gib)
+    num_workers = max(1, min(os.cpu_count() or 1, budget_jobs))
+    if budget_jobs < 1:
+        logger.warning(
+            "Low available memory (%.1f GiB); limiting JIT compilation to a "
+            "single job. Set MAX_JOBS to override.",
+            available_gib,
+        )
+    return num_workers
 
 
 def run_ninja(workdir: Path, ninja_file: Path, verbose: bool) -> None:


pos-ei-don:
Addressed the review feedback in [5b47baef](https://github.com/pos-ei-don/flashinfer/commit/5b47baef): the `os.sysconf()` fallback now checks that `SC_AVPHYS_PAGES` and `SC_PAGE_SIZE` return positive values before multiplying, and falls through to `return None` if either is `-1` or otherwise invalid. This preserves the "undetermined memory → keep previous default" path in `_get_num_workers()` and avoids the spurious low-memory warning that would otherwise fire.

The `/proc/meminfo` primary path is unchanged, so this only hardens the fallback that gets used on non-Linux or when `/proc` isn't readable. Ready for another pass whenever a maintainer has cycles.

pos-ei-don:
Adding a quick "still relevant" data point in case it helps: I upgraded the surrounding runtime this week (LocalAI **v4.7.1** on the same DGX Spark GB10, arm64-l4t-cuda-13 image) and the failure mode this PR addresses is unchanged — the runtime JIT `_get_num_workers()` still uses the current formula, and I still keep the local `MAX_JOBS` cap in the sglang backend's `run.sh` to prevent the compile-storm hard-lockup reproduced back on 2026-07-09.

Nothing new to add on the review comments (the `os.sysconf()` fallback fix from [5b47baef](https://github.com/pos-ei-don/flashinfer/commit/5b47baef) is still the current tip of the branch). Just flagging that this is still the shipping default six weeks after #3634 was filed. Happy to help if there's anything I can do to move it along.

PR #3901, opened 2026-07-09
feat(topk): Add top_k_varlen with GVR and radix backends for sparse-attention KV selection

dhiraj113:
## 📌 Description

Sparse attention has to pick the top-K KV positions for each request on every decode step, and requests in a batch have different sequence lengths. This adds flashinfer.top_k_varlen to do that, with three backends:

radix – CuTe DSL single-pass multi-CTA radix. Handles variable lengths natively (no logit masking) and is Blackwell-only. This is the default when no pre_idx hint is passed.
gvr – Guess-Verify-Refine. Uses the previous step's top-K indices (pre_idx) as a warm start, since a layer's attention pattern barely moves from one step to the next. Blackwell-only, with a load-balance mode for ragged batches.
radix_cutlass – masked fallback on the existing CUTLASS radix kernel. Runs on any GPU.
backend="auto" uses GVR when a pre_idx hint is available, otherwise the CuTe DSL radix on Blackwell, otherwise the CUTLASS fallback.

The GVR and radix kernels are ported from TRT-LLM and tuned to match it. Across the shapes we measured (batch × seq-len × K, bf16 and fp32 on B200) latency is on par with the equivalent TRT-LLM path. Two changes did most of that: SM-aware multi-CTA chunking on the radix path, which also fixes a shared-memory overflow at large N, and feeding the per-CTA scan width into the GVR launch heuristic.

Also in here:

tests/topk/test_topk_varlen.py – covers the three backends, the multi-CTA path, CUDA-graph capture/replay, ragged and seq_len <= top_k cases, and that the backends agree with torch.topk.
benchmarks/routines/topk_varlen.py – compares the backends on a given shape and reference-checks the results.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

Unit tests — covers all three backends, the multi-CTA path, CUDA-graph capture/replay, ragged and seq_len <= top_k cases, and cross-backend agreement with torch.topk:

pytest tests/topk/test_topk_varlen.py
Trace-template reference correctness:

pytest tests/trace/test_top_k_varlen_reference_correctness.py
Benchmark with a reference check across backends (radix / gvr / radix_cutlass):

python benchmarks/flashinfer_benchmark.py --routine top_k_varlen \
    --batch_size 16 --max_seq_len 8192 --top_k 512 --refcheck
The GVR and CuTe-DSL radix paths need Blackwell (sm_100+); radix_cutlass runs anywhere.

## Summary by CodeRabbit

* **New Features**
  * Added a unified `top_k_decode` API with `radix`, `gvr`, or `auto` backend selection, plus `load_balance` for Blackwell-optimized hybrid GVR decoding.
  * Added/expanded options: `return_values`, `next_n`, `compress_ratio`, and support for preallocated output buffers.
  * Expanded the public API surface with `top_k_decode` and new GVR tuning configurations (including load-balanced settings).

* **Tests**
  * Added GPU and radix correctness tests covering values, grouping (`next_n`), compression, preallocation, determinism, fallback behavior, and configuration validation.

* **Benchmarks**
  * Added `top_k_decode` to sampling benchmarks with backend-aware runs and result verification.


flashinfer/topk_varlen/kernels/gvr_topk_decode.py
```diff
@@ -0,0 +1,2256 @@
...
+            smem_input = smem.allocate_tensor(
+                element_type=self.dtype,
+                layout=cute.make_ordered_layout((self.smem_cache_elems,), order=(0,)),
+                byte_alignment=128,
+            )
+        else:
+            smem_input = None
+
+        # ---- Degenerate path: N <= top_k → copy input as-is ----
+        if N <= cutlass.Int32(top_k):
```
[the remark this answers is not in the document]
dhiraj113: is this still an issue

flashinfer/__init__.py
```diff
@@ -227,6 +227,7 @@
 from .topk import top_k_page_table_transform as top_k_page_table_transform
 from .topk import top_k_ragged_transform as top_k_ragged_transform
 from .topk import TopKTieBreak as TopKTieBreak
+from .topk_varlen import top_k_varlen as top_k_varlen
```
leejnau: Consider doing the CuTe DSL imports lazily or guarding them with ImportError. Because top_k_varlen is eagerly exported from flashinfer.__init__, this optional cutlass import currently makes unrelated imports such as from flashinfer import aot fail when nvidia-cutlass-dsl is absent. Should the radix_cutlass fallback remain usable without the Python CUTLASS DSL package?

flashinfer/topk_varlen.py
```diff
@@ -0,0 +1,995 @@
...
+    return_output_values: bool,
+    out_indices: Optional[torch.Tensor],
+    out_values: Optional[torch.Tensor],
+) -> Tuple[torch.Tensor, Optional[torch.Tensor]]:
+    """Run GVR without load-balancing: one CTA per row, rows sorted longest-first."""
+    cute_dtype = torch_to_cutlass_dtype(logits.dtype)
+
+    # Sort requests longest-first (LJF) so early waves process the heaviest rows;
+    # short rows fill the tail, eliminating the load-imbalance tail penalty.
+    order_row = torch.argsort(seq_lens[::next_n], descending=True).to(torch.int32)
```
leejnau: Consider removing the [::next_n] slice here since seq_lens is already request-level with shape num_rows // next_n, so this slice makes order_row too short when next_n > 1. The kernel then indexes beyond it for later requests. It would also be useful to test load_balance=False with next_n > 1.

flashinfer/topk_varlen.py
```diff
@@ -0,0 +1,995 @@
...
+    """CuTe DSL multi-CTA radix top-k: native varlen, no logit masking needed."""
+    cute_dtype = torch_to_cutlass_dtype(logits.dtype)
+    num_rows, N = logits.shape
+    # Note: the row->request mapping (num_rows // next_n) is handled inside the
+    # compiled kernel (row_idx // next_n indexes seq_lens; see radix_topk.py),
+    # so next_n only needs to be forwarded to _compile_radix below.
+
+    # Adjust seq_lens for compress_ratio (match _run_radix_cutlass semantics).
+    if compress_ratio > 1:
+        seq_lens = (seq_lens // compress_ratio).clamp(max=N).to(torch.int32)
```
leejnau: Consider applying compression after the per-row next_n adjustment. This divides the request-level length first, while the radix kernel subsequently subtracts next_n in compressed-index units. The intended length is (seq_len - next_n + row_offset + 1) // compress_ratio. For example, seq_len=16385, next_n=2, and compress_ratio=4 makes row 0 scan 4095 entries instead of 4096. Could we also add a test combining next_n > 1 with compression?

flashinfer/topk_varlen.py
```diff
@@ -0,0 +1,995 @@
...
+# cap at N≈115K bf16 / 57K fp32) and (b) left the kernel's multi-CTA path dead,
+# serializing very long rows on one CTA. These helpers mirror TRT-LLM's
+# ``_compute_max_chunk`` / ``_get_chunk_config`` (cute_dsl_custom_ops.py) so
+# FlashInfer picks the same SMEM-bounded chunk_size and CTA count.
+
+_RADIX_NUM_COPY_BITS = 256  # 256-bit vectorized SMEM loads (matches the kernel)
+# sm_100a opt-in dynamic SMEM capacity (bytes). Same value cutlass reports via
+# get_smem_capacity_in_bytes("sm_100"); inlined to avoid a device query on the
+# hot path.
+_SM100_SMEM_CAPACITY = 232448
```
leejnau: Consider deriving the shared-memory capacity from the target architecture. The radix backend also accepts SM120/121, but this calculation always assumes SM100’s 232,448-byte capacity.  Could the architecture-specific capacity be passed into the chunk configuration?

flashinfer/topk_varlen.py
```diff
@@ -0,0 +1,995 @@
...
+    compress_ratio=1,
+    next_n=1,
+    return_values=False,
+    out_indices=None,
+    out_values=None,
+    backend="auto",
+    load_balance=True,
+):  # extra kwargs mirror the public signature; unused by the check
+    """GVR LB: requires Blackwell hardware, CuTe DSL, and a pre_idx hint."""
+    return is_cute_dsl_available() and pre_idx is not None
```
leejnau: Consider including GVR’s problem constraints in this suitability check.  Since backend="auto" uses this result to select GVR, but currently any Blackwell call with pre_idx passes even with an unsupported top_k or compression ratio, misaligned row width, or a load-balanced batch above 1024. Those calls then fail later instead of falling back to radix.  Should this return False whenever the GVR configuration is not runnable?

flashinfer/topk_varlen.py
```diff
@@ -0,0 +1,995 @@
...
+        .view(torch.int32)
+        .view(max_num_groups, _RADIX_STATE_SIZE)
+    )
+
+    compiled(
+        logits,
+        row_states,
+        seq_lens,
+        out_indices,
+        out_values,
```
leejnau: Consider guarding this argument the way the GVR paths do: out_values if return_output_values else None. Since _compile_radix keys its cache on return_output_values and passes None as the fake output-values tensor when it is False, a caller who supplies out_values= without return_values=True hands a real tensor to a kernel compiled to expect None.  Since _run_gvr and _run_gvr_lb already apply that guard, the radix path is the only inconsistent one.  Should we also add a test that passes out_values with return_values=False across all three backends?

flashinfer/topk_varlen.py
```diff
@@ -0,0 +1,995 @@
...
+    out_values=None,
+    backend="auto",
+    load_balance=True,
+):  # extra kwargs mirror the public signature; unused by the check
+    """GVR LB: requires Blackwell hardware, CuTe DSL, and a pre_idx hint."""
+    return is_cute_dsl_available() and pre_idx is not None
+
+
+def _top_k_varlen_heuristic(suitable_backends, **kwargs):
+    """GVR (needs pre_idx) > radix (CuTe DSL, Blackwell) > radix_cutlass (all GPUs)."""
```
leejnau: Consider spelling out the full public signature here instead of **kwargs. On the skip_check=True path, backend_requirement forwards positional args into heuristic_func in utils.py, so top_k_varlen(logits, seq_lens, top_k, backend="auto", skip_check=True) raises TypeError. The existing heuristics (_heuristic_func_mm_fp4) declare each parameter explicitly for this reason.  Should we also add a skip_check=True case to the backend-selection tests?

flashinfer/cute_dsl/top_k/pdl_utils.py
```diff
@@ -0,0 +1,60 @@
...
+These inline PTX wrappers are used by the GVR Top-K kernel to express
+grid-level dependency scheduling on Blackwell (sm_100+).
+"""
+
+from cutlass._mlir.dialects import llvm
+from cutlass.cutlass_dsl import dsl_user_op
+
+
+@dsl_user_op
+def griddepcontrol_wait(*, loc=None, ip=None) -> None:
```
leejnau: Consider importing griddepcontrol_wait / griddepcontrol_launch_dependents from cutlass.cute.arch instead of adding a new module. They already exist upstream (used directly in flashinfer/gemm/kernels/dense_blockscaled_gemm_sm100.py:44) and are duplicated once more in fused_moe/cute_dsl/blackwell/utils.py:369 and this would be a third copy.

flashinfer/cute_dsl/top_k/config.py
```diff
@@ -0,0 +1,195 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Kernel-level configuration dataclasses for GVR Top-K kernels.
+
+These are internal tuning knobs that most callers should not need to touch.
+The public API (``top_k_varlen``) accepts an optional
```
leejnau: In this docstring  top_k_varlen has no config argument, so there's nothing to omit or pass. Same wording appears in the GvrTopKConfig class docstring just below.

flashinfer/cute_dsl/top_k/config.py
```diff
@@ -0,0 +1,195 @@
...
+            use_constant_hint=False,
+            min_blocks_per_mp=min_blocks_per_mp,
+            use_256bit_load=use_256bit_load,
+            enable_warp_parallel_reduce=enable_warp_parallel_reduce,
+        )
+
+
+@dataclass
+class GvrTopKLBConfig:
+    """Kernel-level configuration for :func:`flashinfer.gvr_topk_lb_prepare`
```
leejnau: Consider pointing this at top_k_varlen(backend="gvr", load_balance=True) instead since flashinfer.gvr_topk_lb_prepare and gvr_topk_lb_decode aren't exported.

flashinfer/topk_varlen.py
```diff
@@ -0,0 +1,995 @@
...
+    if next_n > 1:
+        row_seq_lens = seq_lens.repeat_interleave(next_n)
+        row_offsets = torch.arange(
+            next_n, device=logits.device, dtype=torch.int32
+        ).repeat(seq_lens.shape[0])
+        row_seq_lens = (row_seq_lens - next_n + row_offsets + 1) // compress_ratio
+    else:
+        row_seq_lens = (seq_lens // compress_ratio).clamp(max=N)
+
+    col_idx = torch.arange(N, device=logits.device).unsqueeze(0)
```
leejnau: Consider pushing the seq_lens bound into the kernel, or at minimum caching the col_idx row. This allocates an int64 arange(N) plus a full [num_rows, N] masked copy on every call, which is ~32 MB written per invocation at 128×131072 bf16, and the radix_topk op then allocates its own output_indices that gets copy_d into out_indices at line 658.

flashinfer/topk_varlen.py
```diff
@@ -0,0 +1,995 @@
...
+    out_indices: Optional[torch.Tensor],
+    out_values: Optional[torch.Tensor],
+) -> Tuple[torch.Tensor, Optional[torch.Tensor]]:
+    """LB prepare (sort rows by length) + GVR decode."""
+    cute_dtype = torch_to_cutlass_dtype(logits.dtype)
+    batch_size = seq_lens.shape[0]
+    max_batch_size = _lb_max_batch_size(batch_size)
+    lb_cfg = GvrTopKLBConfig(max_batch_size=max_batch_size)
+
+    order_row = torch.empty(max_batch_size, dtype=torch.int32, device=logits.device)
```
leejnau: Consider allowing caller-provided reusable order_row and counters workspace. These buffers are allocated on every call even though seq_lens is layer-invariant. Also, counters can use torch.empty because the prepare kernel overwrites both entries. Using the global _get_cache_buf directly may introduce races between concurrent streams.

tests/topk/test_topk_varlen.py
```diff
@@ -0,0 +1,1169 @@
...
+    pre_idx = torch.zeros(batch_size, top_k, dtype=torch.int32, device="cuda")
+    for r in range(batch_size):
+        pre_idx[r, 0] = int(logits_f32[r, : seq_len_list[r]].argmax().item())
+    pre_idx[:, 1:] = torch.arange(1, top_k, dtype=torch.int32, device="cuda")
+    return logits, seq_lens, pre_idx
+
+
+@requires_blackwell
+@pytest.mark.parametrize("load_balance", [True, False])
+def test_load_balance_modes(load_balance):
```
leejnau: Consider adding a next_n > 1 case to this parametrization.  Since load_balance=False routes to _run_gvr, whose order_row construction is the one place next_n is mishandled.   Every existing GVR test uses next_n=1 or the LB path, so nothing currently exercises that combination.

tests/topk/test_topk_varlen.py
```diff
@@ -0,0 +1,1169 @@
...
+    indices, _ = flashinfer.top_k_varlen(
+        logits, seq_lens, top_k, pre_idx=None, next_n=next_n, backend="radix"
+    )
+    torch.cuda.synchronize()
+    _check_correct(indices, logits, seq_lens, top_k, next_n=next_n)
+
+
+@requires_blackwell
+@pytest.mark.parametrize("top_k", [512, 1024])
+def test_radix_compress_ratio(top_k):
```
leejnau: Consider adding a case that combines next_n > 1 with compress_ratio > 1, since test_radix_next_n and test_radix_compress_ratio each vary one axis, so the interaction is untested which is where _run_radix (topk_varlen.py:704) diverges from _run_radix_cutlass by dividing before subtracting next_n. Since _check_correct already computes N_eff using the radix_cutlass convention, the test should fail.

tests/topk/test_topk_varlen.py
```diff
@@ -0,0 +1,1169 @@
...
+    lf = logits.float()
+    for row in range(batch_size):
+        expected = lf[row][indices[row].long()]
+        assert torch.allclose(expected, values[row].float(), rtol=1e-3, atol=1e-3), (
+            f"row={row}: values do not match logits[row, indices]"
+        )
+
+
+@requires_blackwell
+def test_radix_preallocated_outputs():
```
leejnau: Consider covering out_values passed with return_values=False.  Both _run_gvr and _run_gvr_lb guard this argument but _run_radix (topk_varlen.py:760) forwards it unconditionally into a kernel compiled to expect None. The existing preallocation tests all set return_values=True, so the mismatch is never hit.

flashinfer/topk_varlen/topk_varlen.py
```diff
@@ -0,0 +1,995 @@
...
+
+# ---------------------------------------------------------------------------
+# Supported compute-capability sets (major * 10 + minor)
+# ---------------------------------------------------------------------------
+
+# All SM tiers FlashInfer ships kernels for.
+_ALL_CCS = [75, 80, 86, 89, 90, 100, 103, 110, 120, 121]
+
+# GVR requires Blackwell sm_100a or newer.
+_BLACKWELL_PLUS_CCS = [100, 103, 110, 120, 121]
```
leejnau, line 72: Consider narrowing this list, or gating gvr separately from radix. The GVR LB path launches cluster_size=4 CTA clusters, but this admits SM103/110/120/121 while the PR reports B200-only measurements and the repo already treats SM120/121 as a 1×1×1-cluster target (flashinfer/jit/gemm/core.py:416). Since backend="auto" picks gvr whenever pre_idx is present, an SM120 caller gets the clustered path with no fallback.  Should we restrict these to the validated tiers and let radix_cutlass serve the rest?

flashinfer/topk_varlen.py
```diff
@@ -0,0 +1,995 @@
...
+                       calls the FlashInfer CUTLASS radix top-K.  Runs on any GPU.
+``"auto"``           — GVR (if pre_idx provided) > radix (Blackwell) >
+                       radix_cutlass (default).
+"""
+
+import functools
+import math
+from typing import Literal, Optional, Tuple
+
+import cutlass
```
kahyunnam: Eager cutlass import breaks `from flashinfer import aot` / AOT wheel builds flashinfer/topk_varlen.py:41-42, flashinfer/__init__.py:230

`topk_varlen` does unconditional `import cutlass` / `import cutlass.cute`, and `__init__.py` eagerly re-exports it. Sibling CuTe APIs use `contextlib.suppress(ImportError)` / try-except.

**CI proof** (AOT Build Import x64 cu129):
```
from flashinfer import aot
  File ".../__init__.py", line 230
    from .topk_varlen import top_k_varlen as top_k_varlen
  File ".../topk_varlen.py", line 41, in <module>
    import cutlass
ModuleNotFoundError: No module named 'cutlass'
```
(isolated `flashinfer-jit-cache` build env lacks cutlass). `radix_cutlass` should remain usable without the Python CUTLASS DSL package — same ask as `@leejnau` on `__init__.py:230`.

flashinfer/topk_varlen/topk_varlen.py
```diff
@@ -0,0 +1,995 @@
...
+        ctas_per_group = max(1, math.ceil(N / max_chunk))
+        chunk_size = math.ceil(N / ctas_per_group)
+        chunk_size = ((chunk_size + vec_size - 1) // vec_size) * vec_size
+        chunk_size = min(chunk_size, max_chunk)
+    else:
+        # Small batch, spare SMs: split each row across ideal_ctas_per_group CTAs.
+        chunk_size = math.ceil(N / ideal_ctas_per_group)
+        chunk_size = max(chunk_size, 8192)  # min chunk: amortize per-CTA reduce
+        ctas_per_group = math.ceil(N / chunk_size)
+        if ctas_per_group == 2 and chunk_size < 32768:
```
kahyunnam, line 536: Same idea as here for chunk_size as https://github.com/flashinfer-ai/flashinfer/pull/3901/changes#r3675791303

kahyunnam (commented):
Agree with most of @leejnau 's comments

flashinfer/topk_varlen.py
```diff
@@ -0,0 +1,1086 @@
...
+        row_states_buffer,
+        top_k,
+        False,  # deterministic
+        0,  # tie_break = TopKTieBreak.NONE
+        False,  # dsa_graph_safe
+    )
+
+    if return_output_values:
+        # Gather values at selected indices — O(num_rows*top_k) vs O(num_rows*N).
+        out_values.copy_(torch.gather(logits, 1, out_indices.long()))
```
leejnau: Consider clamping the index before gathering, since radix_topk_ragged_transform writes -1 into surplus slots when length <= top_k (include/flashinfer/topk.cuh:2395), so this gather gets a negative index and trips a device-side bounds assert, which is reachable via radix_cutlass + return_values=True + any row with seq_len < top_k. The old masked_fill path never gathered, so this is new. Could we clamp and zero the sentinel slots, and add that combination to test_seq_len_less_than_top_k?

benchmarks/routines/flashinfer_benchmark_utils.py
```diff
@@ -886,6 +887,19 @@ def dtype_str_to_torch_dtype(dtype_str):
...
+        # radix = CuTe DSL multi-CTA radix (Blackwell sm_100+); gvr = GVR LB
+        # (Blackwell); radix_cutlass = masked CUTLASS radix fallback (any GPU).
+        "7.5": ["radix_cutlass"],
+        "8.0": ["radix_cutlass"],
+        "8.6": ["radix_cutlass"],
+        "8.9": ["radix_cutlass"],
+        "9.0": ["radix_cutlass"],
+        "10.0": ["radix", "gvr", "radix_cutlass"],
+        "10.3": ["radix", "gvr", "radix_cutlass"],
+        "12.0": ["radix", "gvr", "radix_cutlass"],
```
leejnau: _GVR_CCS is now [100, 103], so gvr here (and on 12.1 below) will fail BackendSupportedError when the default backend set is used on SM120/121. Consider dropping it from both 12.x entries.

tests/topk_varlen/test_topk_varlen.py
```diff
@@ -0,0 +1,1376 @@
...
+    subtracting next_n in compressed-index units gives the wrong column bound
+    (off by up to compress_ratio-1 columns per row).
+    """
+    dtype, top_k, next_n, compress_ratio = torch.bfloat16, 512, 2, 4
+    # N=4096 logit columns = 16384 uncompressed tokens per column.
+    # Use seq_lens near a compress_ratio boundary (e.g. 16385) so the
+    # integer-division order matters: (16385-2+0+1)//4=4096 vs (16385//4)-2+0+1=4095.
+    N, batch_size = 4096, 8
+    num_rows = batch_size * next_n
+    logits, _, seq_lens = _make_inputs(
```
leejnau, line 909: Consider explicitly setting seq_lens to N * compress_ratio + 1 here. Despite the comment mentioning 16385, _make_inputs sets it to N * compress_ratio, which is 16384. At that divisible boundary, the old divide-before-adjust implementation produces the same result, so this test would not catch the regression it describes.

tests/topk/test_topk_varlen.py
```diff
@@ -0,0 +1,1376 @@
...
+    logits, pre_idx, seq_lens = _make_inputs(batch_size, N, top_k, dtype, seed=3)
+    pre_idx_arg = pre_idx if _IS_BLACKWELL else None
+
+    idx1, _ = flashinfer.top_k_varlen(logits, seq_lens, top_k, pre_idx=pre_idx_arg)
+    torch.cuda.synchronize()
+    idx2, _ = flashinfer.top_k_varlen(logits, seq_lens, top_k, pre_idx=pre_idx_arg)
+    torch.cuda.synchronize()
+
+    for row in range(batch_size):
+        assert set(idx1[row].cpu().tolist()) == set(idx2[row].cpu().tolist()), (
```
leejnau: Consider checking both results with _check_correct instead of requiring identical index sets. On non-Blackwell GPUs this uses radix_cutlass with deterministic=False, and BF16 values can tie at the K-th boundary, allowing two correct calls to select different tied indices. This exact assertion is currently failing in the H100 CI job.

flashinfer/trace/templates/topk.py
```diff
@@ -0,0 +1,153 @@
...
+            ["batch_size", "max_seq_len"],
+            description="Decode-step attention logits (bfloat16 / float16 / float32).",
+        ),
+        "seq_lens": Tensor(
+            ["batch_size"],
+            dtype="int32",
+            description="Effective KV-cache length per request.",
+        ),
+        "top_k": Scalar("int32", description="K — number of top elements to select."),
+        "pre_idx": Tensor(
```
leejnau, line 135: Consider ensuring the top_k axis is extracted from the scalar top_k input before this optional tensor. _build_axis_extractors currently finds top_k in pre_idx first and stops, but generated trace samples omit optional inputs. The extractor therefore returns None without falling back to the scalar, causing the trace-consistency test to report Const axes missing values: ['top_k'].

flashinfer/topk_varlen/topk_varlen.py
```diff
@@ -0,0 +1,1200 @@
...
+
+@backend_requirement(
+    {
+        "radix": _radix_top_k_varlen_check,
+        "gvr": _gvr_top_k_varlen_check,
+        "radix_cutlass": _radix_cutlass_top_k_varlen_check,
+    },
+    heuristic_func=_top_k_varlen_heuristic,
+)
+@flashinfer_api(trace=top_k_varlen_trace)
```
leejnau, line 972: The rebase picked up tests/trace/template_registry.py, whose _TRACE_REGISTRATION_MODULES must list every module with @flashinfer_api(trace=...) test_template_registry.py::test_registration_module_inventory_is_complete asserts exact set equality against an AST scan. Consider adding "flashinfer.topk_varlen" after "flashinfer.topk" (line 81 of that file), otherwise that test fails on this branch.

leejnau (approved):

flashinfer/topk_varlen/kernels/__init__.py
bkryu, line 1: We are trying to avoid dumping everything in `flashinfer/cute_dsl`.

Can you create a `top_k` or `sampling` directory in `flashinfer`, create a `kernels` directory inside there, and move the kernel source code in there? (For example, see [flashinfer/quantization](https://github.com/flashinfer-ai/flashinfer/tree/main/flashinfer/quantization))

flashinfer/topk_varlen/topk_varlen.py
bkryu, line 1: Ditto for this file. As we create `flashinfer/sampling` or `flashinfer/top_k` let's plcae this file in there as well for better organization

tests/topk_varlen/test_topk_varlen.py
bkryu, line 1: As we create this `tests/topk` directory, can we move `tests/utils/test_topk.py` to `tests/topk`?

Afterwards, we should be able to trigger `/bot run tests/topk` to test both files and iterate on this PR quickly instead of `/bot run`

benchmarks/routines/flashinfer_benchmark_utils.py
```diff
@@ -886,6 +897,20 @@ def dtype_str_to_torch_dtype(dtype_str):
...
+        "7.5": ["radix_cutlass"],
+        "8.0": ["radix_cutlass"],
+        "8.6": ["radix_cutlass"],
+        "8.9": ["radix_cutlass"],
+        "9.0": ["radix_cutlass"],
+        "10.0": ["radix", "gvr", "radix_cutlass"],
+        "10.3": ["radix", "gvr", "radix_cutlass"],
+        "12.0": ["radix", "radix_cutlass"],
+        "12.1": ["radix", "radix_cutlass"],
+    },
```
bkryu: Since `top_k_varlen`   has the `backend_requirement` decorator, we will not need to have these hard coded version checks. `mm_fp4` and its test code may be a good reference on how it is handled. Can you double check and modify?

bkryu (commented):
Thanks @dhiraj113, this seems like a great PR.

Added some minor comments, focusing on organization of files.

tests/topk_varlen/test_topk_varlen.py
dhiraj113: I don't want to touch `test_topk.py` as moving that would need moving the flashinfer `topk.py` to maintain symmetry. Also, `sampling` can't be used as the directory/package name as a file named `sampling.py` already exists. These are good suggestions, but I will leave it for a future cleanup MR. Here, I want to restrict the scope only to the newly added `top_k_varlen` api and its associated files.

dhiraj113:
@bkryu addressed your comments. Please take a look.

bkryu (approved):

saltyminty (approved):
Late but LGTM

dhiraj113:
Blocked due to lack of a functioning CI/CD.

PR #3903, opened 2026-07-09
feat(comm): extend trtllm_allreduce to SM12x and fix lamport buffer pointer packing

yichengj0:
## 📌 Description

Resolves item 14 of the SM121 support audit (#3170). The fused TRT-LLM allreduce module (`trtllm_comm`) refuses to build on SM12x because its JIT gate only lists SM90 and SM100, so those GPUs get no fused allreduce at all.

The gate came from the original TRT-LLM port, not from a hardware limit. SM12x has everything these kernels use (verified on RTX 5080 and DGX Spark: thread-block clusters, distributed shared memory, programmatic dependent launch). Two real bugs sat behind the gate:

- **Empty kernel on SM12x.** Separate from the JIT gate, the kernel header excludes SM12x with a compile-time check. Widening the JIT gate alone would build the lamport kernel (the low-latency path for small batches) with an empty body: small-batch decode calls would launch it, do nothing, and return garbage with no error.
- **Wrong pointer layout in the host wrapper (all architectures).** The wrapper writes the lamport buffer pointers into a fixed-size array with one spacing between buffer sets, and the kernels read them back with another. The two layouts only agree at world size 16. At any other world size the kernel reads slots that were never written, then crashes or corrupts. No test ever caught this: the lamport kernel only runs at 16 tokens or fewer, and the existing tests use 64 and 128.

Fixes:

- The wrapper now writes the pointer array with the same spacing the kernels read.
- The compile-time exclusion is removed, so the lamport kernel gets a real body on SM12x.
- The JIT gate now includes major version 12, and AOT builds include the module for every arch it supports.
- Launch cluster sizes are clamped to what the device reports through `cudaOccupancyMaxPotentialClusterSize`, instead of assuming a size per arch. SM90, SM100, SM120, and SM121 all report 8, so behavior is unchanged where the kernels already ran.
- The test suites skip cleanly on unsupported machines instead of erroring.

SM90 and SM100 compile to the same machine code as before; the compile-time check change does not affect them.

## 🔍 Related Issues

#3170 (item 14, cross-cutting issue 5).

## 🧪 Tests

On DGX Spark (GB10, SM121, CUDA 13, aarch64) and RTX 5080 (SM120, CUDA 13.3, x86_64):

- JIT build through `gen_trtllm_comm_module` (`sm_121a` and `sm_120f`).
- A functional test with two ranks sharing a single GPU over CUDA IPC. The lamport, oneshot, twoshot, and fusion paths all match fp32 references on both GPUs, including the small-batch lamport shapes the existing suites never reach. Before the pointer fix, those shapes crash or corrupt.
- Both GPUs report max cluster size 8 for the real kernels and pass cluster-size-8 launches.
- Cross-compilation for `sm_90a` and `sm_100a` is clean.
- `pre-commit run` is clean.

## Reviewer Notes

- The lamport dispatch thresholds (16 tokens or fewer, hidden size 256 or more) carry over from TRT-LLM's datacenter tuning. Whether lamport beats plain oneshot on SM12x needs a real multi-GPU measurement.
- No API change. Test changes are only skip guards.

## Summary by CodeRabbit

* **New Features**
  * Broadened TRT-LLM communication-kernel build support to additional GPU generations (SM90/100/100f/103/120/120f/121).
  * Added CUDA major version 12 handling for NVCC flag selection.
* **Bug Fixes**
  * Added topology-limit validation for all-reduce IPC packing.
  * Corrected triple-buffer lamport IPC pointer packing stride.
  * Expanded architecture-guarded synchronization/launch-completion behavior.
  * Updated fusion and MoE cluster sizing to clamp using each device’s maximum supported cluster size.
* **Tests**
  * Tests now run only when CUDA is available and compute capability is supported (9/10/12), with lamport buffers re-initialized per shape and expanded token coverage (including 8).


csrc/trtllm_allreduce.cu
```diff
@@ -143,12 +143,15 @@ void trtllm_custom_all_reduce(TensorView in, TensorView out, int64_t tp_size, in
       TVM_FFI_ICHECK(lamport_peer_comm_buffer_ptrs_2.has_value())
           << "lamport_peer_comm_buffer_ptrs_2 is required if lamport_peer_comm_buffer_ptrs_0 "
              "is provided";
+      // Kernels read the triple-buffered lamport pointers at stride
+      // MAX_RANKS_PER_NODE, not tp_size.
       for (int i = 0; i < tp_size; ++i) {
```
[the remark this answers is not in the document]
yichengj0: Added in d3a00179.

include/flashinfer/utils.cuh
```diff
@@ -359,6 +359,22 @@ inline std::pair<int, int> GetCudaComputeCapability() {
...
+  config.blockDim = block_dim;
+  int cluster_size = 0;
+  cudaError_t status = cudaOccupancyMaxPotentialClusterSize(&cluster_size, kernel, &config);
+  if (status != cudaSuccess || cluster_size < 1) {
+    (void)cudaGetLastError();
+    return 1;
+  }
+  return cluster_size;
+}
+
```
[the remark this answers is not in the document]
yichengj0: This is a capability check (can the device launch multi-block clusters), not an occupancy optimization. The result is min-ed with the same heuristic sizes the launchers have always used, so on arches where these kernels already ran the final (cluster size, block size) combinations are identical to the previous hardcoded 8; a blockDim-dependent launch failure would predate this PR. Verified on SM120 and SM121 with block sizes 128-1024: all launches at the probed cluster size succeed. A per-launch re-query would add an occupancy call on the decode hot path without covering a reachable failure mode.

ormandj:
End-to-end SM120 TP2 integration evidence is available for head `d3a00179`. Against an otherwise fixed PyNCCL control, the complete required stack—#3903, the vLLM SM12x selector, and an initialization-only CUDA-runtime resolver—changes sustained MTP:0 decode at C1/C2/C4/C8/C16/C32 by `+5.18/+5.14/+4.10/+3.22/+3.22/+2.15%`; exact 8K/64K/128K cold prefill changes by `+1.39/+0.25/+0.40%`. The candidate selected the `trtllm` workspace, started with zero restarts, completed every decode cell without errors or underfill, and passed sequential plus C8 semantic correctness. These percentages are joint-stack evidence, not isolated attribution to this PR. Both prior technical review threads are resolved. Requesting authorized CI and maintainer review for merge; this addresses #3170 item 14.

Caveat added 2026-07-24 for accuracy: the percentages above come from a single run per concurrency cell. I have since measured up to 6.9% run-to-run drift on a settled server with this workload, so individual cell percentages should be treated as provisional; the consistent sign and monotonic trend across all six cells still indicate a real effect, and the kernel-level evidence in my later comment is unaffected.

kahyunnam (approved):
lgtm! One small comment: can we make sure CI tests hit lamport path?

yichengj0:
Good catch. The fusion suite already hits its lamport kernel (token_num=1 with use_oneshot=True), but test_trtllm_allreduce.py did not: its token counts (64/128) are all above the lamport threshold of 16, which is also why the pointer bug fixed here was never caught. Added token_num=8 in cfda5d12 so ONESHOT + RESIDUAL_RMS_NORM dispatches to the lamport kernel in the single-node comm CI job.

yichengj0:
Pushed a commit that fixes a bug in the new test that causes it to fail in multi-gpu settings

ormandj:
End-to-end confirmation on SM120: 2x RTX PRO 6000 Blackwell (TP2), DeepSeek-V4-Flash MTP2 with CUDA graphs, SGLang main `3d91a569ce` + sgl-project/sglang#30700 + sgl-project/sglang#32330, FlashInfer 0.6.14 + this PR at `1eb77ab2` (+ #3930).

- Workspace initialization succeeds on both ranks (`backend=trtllm`, world_size=2), including under CUDA graph capture and replay.
- Torch profiler traces show the one-shot Lamport kernel replacing NCCL ring all-reduce 1:1 (87 kernels per decode graph, 1,113 across the capture), summed all-reduce time 0.772 → 0.394 ms per graph.
- C1 sustained decode, five consecutive 30 s runs per cell on a settled server: NCCL control medians 126.25 tok/s (mean 125.75), with this PR's kernel active medians 132.85 (mean 134.31) — +5.2% on medians, +6.8% on means. On a separate tuned stack, per-graph all-reduce time was 0.367 ms versus vLLM's 0.364 ms on the same hardware; the end-to-end figures from that stack were single runs and are omitted.
- GSM8K 200-question accuracy at parity across NCCL/Lamport cells (97.5% / 97.0%, 0 invalid). Zero errors or restarts across sustained decode and profiling runs.

Downstream status: sgl-project/sglang#32330 (SM120 enablement on the SGLang side) is open and maintainer-acknowledged, waiting specifically on this PR landing in a stable FlashInfer release.

include/flashinfer/comm/trtllm_allreduce.cuh
```diff
@@ -414,7 +414,7 @@ __global__ void rms_norm_kernel(AllReduceParams<T> params) {
   local_final_output_buffer += block_offset;
   intermediate_buffer += block_offset;
 
-#if (defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 900) && (__CUDA_ARCH__ < 1200))
+#if (defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 900))
   cudaGridDependencySynchronize();
```
Anerudhan, line 418: Do you want to put a different upper limit `1300`, so it is not auto chosen for future architectures

yichengj0: I considered a cap, but I think open-ended is safer here. Arch gating already happens in the JIT gate (`supported_major_versions`), which fails loudly. With a `< 1300` cap, once SM13x gets added to that gate the module would build and launch fine, but the guard would compile the sync out of the kernel and produce wrong results with no error, which is how the old `< 1200` bound broke SM12x.

Anerudhan: SG.

Anerudhan (approved):

PR #3918, opened 2026-07-10
[fix] fix autotuner non tensor guard

jiahanc:
## 📌 Description

Fix AutoTuner cold-L2 profiling for operators whose input list contains non-Tensor values, such as `torch.dtype` or optional `None` arguments.

  This regression was introduced while refactoring AutoTuner in [#3687](https://github.com/flashinfer-ai/flashinfer/pull/3687). 

 During the AutoTuner refactor in [#3687](https://github.com/flashinfer-ai/flashinfer/pull/3687), guards for non-Tensor inputs were accidentally removed. 
  ```python
  input.numel() * input.element_size()
  if isinstance(input, torch.Tensor)
  else 0
```
As a result, MXFP8 and some MoE tuning paths attempted to call Tensor methods on `torch.dtype` or `None`. Profiling failures were caught internally, causing all tactics to be skipped and AutoTuner to fall back to the default tactic.

  ## Fix

Only calculate buffer sizes and clone inputs that are `torch.Tensor` instances. Preserve other inputs unchanged.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved autotuning batching to correctly handle mixed inputs that include non-tensor values.
  * Updated profiling preparation to size cold-cache buffers using tensor memory only, while preserving non-tensor entries unchanged.
  * Clones are now applied only to tensor inputs when generating additional batches; non-tensors are kept as-is.
* **Tests**
  * Added regression coverage to verify non-tensor preservation and correct tensor cloning behavior during batching.

### DIFF
--- a/flashinfer/autotuner/autotuner.py
+++ b/flashinfer/autotuner/autotuner.py
@@ -2181,14 +2181,19 @@ class AutoTuner:
 
     def _prepare_input_tensors_with_batches(
         self,
-        inputs: list[torch.Tensor],
+        inputs: list[Any],
         tuning_config: TuningConfig,
-    ) -> list[list[torch.Tensor]]:
+    ) -> list[list[Any]]:
         """Create multiple input copies to flush the L2 cache between profiling iterations."""
         if not tuning_config.use_cold_l2_cache:
             return [inputs]
 
-        one_buffer_bytes = sum(input.numel() * input.element_size() for input in inputs)
+        one_buffer_bytes = sum(
+            input.numel() * input.element_size()
+            if isinstance(input, torch.Tensor)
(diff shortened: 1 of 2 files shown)


kahyunnam (approved):

qiching (commented):
LGTM. 
and also #3687 Is this the only place in the refactoring where the `guard` was omitted? Have you used `grep` to check for other places where `.numel()`, `.size()`, `.element_size()`, or `.clone()` are called directly on input parameters to confirm that no similar omissions were missed?

PR #3923, opened 2026-07-10
Fix SM120 DSV4 topk-256 decode dispatch

voipmonitor:
## Summary

Add the missing SM120 DSV4 sparse-MLA decode instantiations for `topk=256` and fail explicitly when a decode-sized DSV4 shape is unsupported instead of silently falling through to the generic paged-attention path.

## Why

DeepSeek-V4-Flash-DSpark can build non-causal sparse-attention metadata with a 128-token context plus a padded draft block. That produces a `topk=256` decode bucket. The existing DSV4 dispatch table supports 128, 512, and 1024, so valid DSpark decode requests miss the specialized SM120 kernel.

The added instantiations cover the DSV4 head counts used by TP1/TP2/TP4 (`8`, `16`, `32`, and `64`). The Python dispatch table is kept in sync with the C++ switch.

For decode-sized inputs, an unsupported shape now raises a diagnostic containing model type, token count, head count, top-k, QK width, page-block size, and extra top-k. Prefill-sized inputs retain the existing generic path.

## Validation

- `git diff --check`
- `python3 -m py_compile flashinfer/mla/_sparse_mla_sm120.py tests/attention/test_sparse_mla_sm120.py`
- Added `(num_heads=32, topk=256)` to the existing SM120 DSV4 decode matrix.
- Confirmed the dispatch is present in the image used for TP2 DSpark end-to-end validation.

## Summary by CodeRabbit

* **New Features**
  * Added SM120 sparse MLA decode support for TOPK=256 across 8, 16, 32, and 64 attention heads.

* **Bug Fixes**
  * Unsupported decode configurations now return a clear error instead of silently using a fallback path.

* **Tests**
  * Added coverage for the 32-head, TOPK=256 configuration.


voipmonitor:
Closing in favor of the earlier and broader #3817, which covers the same DSV4 topk=256 decode dispatch, includes NUM_HEADS=128, tests, and the unsupported-shape guard. The release image will use #3817 together with prefill fix #3896.

PR #3930, opened 2026-07-12
fix(comm): make find_loaded_library skip look-alike libs like libcudart_stub.so

### LINKED ISSUE #3676: import flashinfer.comm fails at import time when a co-loaded library matches "libcudart" (e.g. tilelang libcudart_stub.so) — find_loaded_library picks the non-runtime lib

aryanputta:
Fixes #3676.

## Problem
`import flashinfer.comm` can fail at import time with `AttributeError: undefined symbol: cudaDeviceReset` when another package in the same process ships a shared object whose filename merely *contains* `libcudart` but is not the real CUDA runtime.

`find_loaded_library("libcudart")` scanned `/proc/self/maps` and returned the **first** line containing the substring `libcudart`. Because `/proc/self/maps` is address-sorted, that is whichever `libcudart`-named mapping has the lowest base address. When [`tilelang`](https://github.com/tile-ai/tilelang) is co-loaded (e.g. both it and flashinfer are pulled in by vLLM), its `libcudart_stub.so` can map below the real `libcudart.so.13` and get selected. The stub does not export `cudaDeviceReset`, so `CudaRTLibrary.__init__` raises at import.

The existing `assert filename.rpartition(".so")[0].startswith(lib_name)` did not catch this: the stub stem `libcudart_stub` still `startswith("libcudart")`.

## Fix
Scan every mapping instead of breaking on the first hit, and accept only filenames that are really `lib_name`: `lib_name.so[.ver]` or `lib_name-<hash>.so[.ver]` (the hash form the original comment already documented). This rejects `libcudart_stub.so` while still matching the real runtime regardless of load order.

## Verification
Arch-independent, no GPU needed (the bug is in path selection). Ran the edited function against a mocked `/proc/self/maps`:
- stub mapped *below* the real runtime -> returns `libcudart.so.13` (previously returned the stub)
- hashed name `libcudart-<hash>.so.11` -> matched
- only the stub present -> `None` (so `__init__`'s existing assert reports "not loaded" instead of a confusing undefined-symbol crash)
- library not loaded -> `None`

AI-assisted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of loaded CUDA runtime libraries.
  * Prevented incorrect matches for stub libraries and dotted look-alike filenames.
  * Preserved support for exact library names, version or hash suffixes, and hyphenated variants.
  * Returns no result when a valid matching library cannot be found.

### DIFF
--- a/flashinfer/comm/cuda_ipc.py
+++ b/flashinfer/comm/cuda_ipc.py
@@ -47,24 +47,25 @@ def find_loaded_library(lib_name) -> Optional[str]:
     shared libraries loaded by the process. We can use this file to find the path of the
     a loaded library.
     """  # noqa
-    found = False
+    # A plain `lib_name in line` substring match on the first hit is not enough:
+    # another loaded library can have a filename that merely contains lib_name but
+    # is not the library we want (e.g. tilelang's libcudart_stub.so vs the real
+    # libcudart.so.13). Because /proc/self/maps is address-sorted, the first hit is
+    # whichever mapping has the lowest base address, so the stub can win and later
+    # raise "undefined symbol" on load. Scan every mapping and keep only names that
+    # are really lib_name, i.e. `lib_name.so[.ver]` or `lib_name-<hash>.so[.ver]`.
+    # See https://github.com/flashinfer-ai/flashinfer/issues/3676.
     with open("/proc/self/maps") as f:
         for line in f:
-            if lib_name in line:
-                found = True
-                break
-    if not found:
-        # the library is not loaded in the current process
-        return None
-    # if lib_name is libcudart, we need to match a line with:
-    # address /path/to/libcudart-hash.so.11.0
-    start = line.index("/")
-    path = line[start:].strip()
-    filename = path.split("/")[-1]
-    assert filename.rpartition(".so")[0].startswith(lib_name), (
-        f"Unexpected filename: {filename} for library {lib_name}"
-    )
-    return path
+            if lib_name not in line or "/" not in line:
+                continue
+            path = line[line.index("/"):].strip()
+            filename = path.split("/")[-1]
+            stem = filename.rpartition(".so")[0]
+            if stem == lib_name or stem.startswith(lib_name + ".") or stem.startswith(lib_name + "-"):
+                return path
+    # the library is not loaded in the current process
+    return None
 
 
 class CudaRTLibrary:


ormandj:
Reproduced #3676 on SM120 TP2: substring lookup selected TileLang's `libcudart_stub.so`, and FlashInfer workspace initialization failed while binding `cudaDeviceReset`. After replacing substring lookup with filename-based matching, the same `trtllm` configuration initialized its workspace and reached readiness. A later integration run using the resolver carry started with zero restarts and passed semantic correctness. This is initialization evidence, not a performance claim.

I prepared a signed-off commit [`bd6765de`](https://github.com/ormandj/flashinfer/commit/bd6765dea271b23a579938132f8ca1b9cbf6a2a5) directly against this head for both current review findings. Its anchored matcher accepts only `libcudart.so[.N...]` or a nonempty lowercase-hex `libcudart-<hash>.so[.N...]`; it rejects underscore, arbitrary-hyphen, dotted, empty-hash, and path lookalikes. It adds CPU cases for exact, versioned, hash-mangled, collision, load-order, and no-match behavior. This stricter matcher builds on the filename-based resolver approach in #3677 and vllm-project/vllm#47586; credit for those earlier fixes remains with their authors. The focused pytest file passes 12/12 in an isolated uv Python 3.12 environment, and `pre-commit run --all-files` passes every hook. Requesting review of the matcher/test update rather than opening another resolver PR.

ormandj:
This fix was adopted in our SM120 stack (SGLang TP2, TileLang installed as an SGLang dependency) after earlier testing hit IPC initialization failures from `find_loaded_library` selecting TileLang's `libcudart_stub.so`. With this PR at `e855cc25` (alongside #3903), TRT-LLM all-reduce workspace initialization succeeds on both ranks and survives CUDA graph capture and replay. Tested on SGLang main `3d91a569ce`, FlashInfer 0.6.14; end-to-end numbers in flashinfer-ai/flashinfer#3903 and sgl-project/sglang#32330.

ormandj:
Correction to the above after discussion with SGLang maintainers: current SGLang main does not need this fix — sgl-project/sglang#30580 (merged Jul 12) lazy-loads TileLang so the stub library is not resident during IPC initialization. Our failure came from a configuration that prewarms TileLang at load. The fix may still be worth having as a robustness improvement in `find_loaded_library` itself, independent of consumer-side workarounds.

vedularaghu:
Worth landing this — the underlying selection bug is still live on `main` and in v0.6.18.
`find_loaded_library` is unchanged, and `CudaRTLibrary.__init__` still calls it with no
`so_file`.

I hit it independently on a serving stack that has TileLang resident before IPC
initialization, so the consumer-side workaround noted above (lazy-loading TileLang) does
not generalize — this really does want fixing in `find_loaded_library` itself, as you
suggested.

Your matcher looks right to me for both spellings a real runtime uses, the plain
`libcudart.so.<ver>` and the hashed `libcudart-<hash>.so.<ver>`, while still rejecting
`libcudart_stub.so`. Worth noting that a simpler-looking tightening to
`filename.startswith(lib_name + ".so")` would silently break the hashed form, so the stem
comparison you went with matters.

One detail that makes this easy to under-prioritize: since 0.6.17 the handle is resolved
lazily, so the original import-time reproduction no longer fails even though the bug is
still there — the failure just moves to the first `cudart.*` access. Details in
https://github.com/flashinfer-ai/flashinfer/issues/3676#issuecomment-5498963361 rather
than repeating them here. Landing this fixes the selection regardless of when the handle
is resolved.

Is this blocked on anything I can help with? Happy to test a build against a
stub-resident environment if that would move it along.

PR #3946, opened 2026-07-13
Fix the routing inconsistency for num_groups > 1

b8zhong:
(For example, this epsilon is in official DSV3.2 inference demo https://github.com/deepseek-ai/DeepSeek-V3.2-Exp, and it's currently applied when num_groups = 1 (GLM 5.x), but not this path. Add it for potential fix for instability when using f32 expert correction bias.


aleozlx:
thanks for the contribution

could you elaborate how this works?

aleozlx (approved):
code review: looks good. asked a question about the context

b8zhong:
Hi @aleozlx , it's a similar fix to https://github.com/flashinfer-ai/flashinfer/pull/3803. 

For sigmoid, 0.5*tanh(0.5*x) + 0.5, and if the logit value is very negative, and tanh = -1, then 0.5 x -1 + 0.5 = 0. In rare cases when they are all negative, it would cause 0 division and thus NaN.

(This PR fixes for when num_groups > 1, ie. Deepseek V3.2/3.1/V3 original models. The other PR only fixes for num_groups = 1, e.g GLM-5.2, Kimi K2.6.)

PR #3949, opened 2026-07-13
perf(attention): reuse TRT-LLM-gen counter buffers

saltyminty:
## 📌 Description

Fix the B300 TRT-LLM-native decode regression caused by allocating and zeroing the multi-CTA KV counter sidecar inside every timed call.

- Preallocate separate reusable multi-CTA counter buffers for the direct GQA and MLA benchmark paths before timing and CUDA graph capture.
- Add optional caller-owned counter-buffer reuse to the TRT-LLM-gen MLA API. Supplied buffers are validated for device, contiguity, and capacity; callers that omit the buffer retain the internal allocation fallback.
- Keep caller-owned MLA buffers out of synthetic autotune profiles. The TRT-LLM MLA runner lazily allocates and reuses runner-owned profiling storage, then receives the caller buffer only for the selected runner's final real request. This allows autotune profile batches to exceed the request batch without resizing or replacing caller storage.
- Add exact GPT-OSS and MLA regression coverage for repeated reuse, pointer stability, kernel self-reset, invalid buffers, and the autotune profiling/final-call ownership contract.

Fresh validation at PR head `78b622da` reran the exact B300/SM103 GPT-OSS case against the pre-regression parent using one warm process and five measured processes per ref. Median TRT-LLM-native throughput was `15.330 TFLOP/s` for this PR versus `15.313 TFLOP/s` for the parent. Reconstructed from the fixed operation count, latency was `0.008755233 ms` versus `0.008764953 ms`, making this PR `0.110894%` faster. All 12 benchmark processes passed refcheck; both refs used the same CUDA-event fallback because CUPTI was unavailable.

## 🔍 Related Issues

- [NVBug 6438060](https://nvbugspro.nvidia.com/bug/6438060)

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

Validation includes the focused B300 correctness suite (`28 passed`), the broader decode/MLA/prefill regression matrix (`1392 passed`), focused B200 caller-buffer validation (`6 passed`), targeted B200 autotune/final-call validation (`7 passed`), and the fresh exact B300 parent-versus-current benchmark above. Repository and changed-file pre-commit coverage passed, including mypy and Ruff.

## Reviewer Notes

Please focus on the caller-owned buffer lifetime and autotune ownership contract. Synthetic autotune profiles use reusable runner-owned storage so profile batch buckets can exceed the real request batch safely. Only the selected TRT-LLM runner's final request receives caller-owned storage. Caller buffers must remain alive for all launches or graph replays and must not be shared by overlapping streams or graphs.

## Summary by CodeRabbit

* **New Features**
  * Added optional multi-CTA KV counter buffer support for TRT-LLM attention and MLA decoding.
  * Caller-provided buffers can be reused across decode runs, including autotuned execution.
  * Added backend-aware handling to ensure the counter buffer is only applied where supported.
* **Bug Fixes**
  * Improved correctness when using preallocated output tensors for MLA decoding (including in-place behavior).
* **Tests**
  * Expanded tests to verify counter-buffer reuse, all-zero preservation, validation errors, and autotune behavior.

### DIFF
--- a/benchmarks/routines/attention.py
+++ b/benchmarks/routines/attention.py
@@ -24,7 +24,11 @@ except OSError as e:
 from flashinfer import autotune
 from flashinfer.fp4_quantization import nvfp4_quantize_paged_kv_cache
 from flashinfer.prefill import trtllm_fmha_v2_prefill
-from flashinfer.utils import is_sm12x_supported
+from flashinfer.utils import (
+    get_device_sm_count,
+    get_trtllm_gen_multi_ctas_kv_counter_bytes,
+    is_sm12x_supported,
+)
 from flashinfer.testing.utils import (
     attention_tb_per_sec_with_actual_seq_lens,
     attention_tflops_per_sec_with_actual_seq_lens,
@@ -571,6 +575,14 @@ def testBatchDecodeWithPagedKVCacheWrapper(args):
 
     scale = float(1.0 / (head_dim_qk**0.5))
     workspace_buffer = torch.empty(512 * 1024 * 1024, dtype=torch.int8, device=device)
+    gqa_multi_ctas_kv_counter_buffer = None
+    if "trtllm-native" in backends:
+        counter_bytes = get_trtllm_gen_multi_ctas_kv_counter_bytes(
+            batch_size, num_qo_heads, get_device_sm_count(device)
+        )
+        gqa_multi_ctas_kv_counter_buffer = torch.zeros(
+            counter_bytes, dtype=torch.uint8, device=device
+        )
 
     if args.verbose >= 2:
         print(f"[VVERBOSE] {kv_cache.shape = }")
@@ -699,6 +711,7 @@ def testBatchDecodeWithPagedKVCacheWrapper(args):
                 mask=speculative_mask,
                 kv_cache_sf=kv_cache_sf,
                 enable_pdl=args.enable_pdl,
+                multi_ctas_kv_counter_buffer=gqa_multi_ctas_kv_counter_buffer,
             )
         else:
             print(f"[ERROR] Backend {backend} not supported")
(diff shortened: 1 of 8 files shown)


qsang-nv:
Verified the autotune ownership contract at head `f919c89f`: `choose_one` is invoked without extra kwargs, so profiling never sees the caller buffer (runner-owned storage grows instead), and only the final real call receives it. No blockers from my side. A few non-blocking suggestions:

1. **Alignment validation gap**: the launcher requires the counter pointer to be 16-byte aligned (`csrc/trtllm_fmha_kernel_launcher.cu:242`), but `_resolve_trtllm_gen_multi_ctas_kv_counter_buffer` only checks device/contiguity/size. A contiguous-but-offset view (e.g. `torch.zeros(n + 1, dtype=torch.uint8, device="cuda")[1:]`) passes Python validation and only fails at the kernel boundary — under autotune that means the final real request, after tuning already succeeded. Suggest adding a `data_ptr() % 16 == 0` check that raises `ValueError`, plus a "contiguous but offset" sibling of the existing strided-view test.
2. **Share the resolver**: `decode.py` and `prefill.py` carry duplicated inline validation that lacks the contiguity (and alignment) checks the MLA resolver has. Moving the resolver into `flashinfer/utils.py` next to `get_trtllm_gen_multi_ctas_kv_counter_bytes` and reusing it in all three paths would fix the asymmetry once.
3. **Orphaned comment**: in `TrtllmGenMlaDecodeRunner.__init__`, the "Allocated lazily for autotune profiling…" comment now dangles after `self.lse = lse`, while the attribute it documents (`_multi_ctas_kv_counter_buffer`) moved up next to `self.sm_count`. Move the comment next to the attribute.
4. Nit: in `forward()`, the `None` branch calls `_resolve_trtllm_gen_multi_ctas_kv_counter_buffer(None, ...)`, which is just `_get_trtllm_gen_multi_ctas_kv_counter_buffer` with extra indirection — call it directly.

saltyminty:
/bot run tests/attention tests/autotuner/test_autotuner_mla_decode.py

qsang-nv (approved):

PR #3952, opened 2026-07-14
CuTe-DSL modular prefill: sliding-window support + band-mask performance

pgera:
# CuTe-DSL modular prefill: sliding-window support + band-mask performance

## Summary

The modular CuTe-DSL prefill (SM100) mapped `causal=True + window_left >= 0`
to a plain causal mask — wrong results, and full O(s²/2) work. This PR fixes
the mask semantics with composable per-row bands, removes the banded regime's
dead work, makes window sizes runtime arguments (one compiled kernel per mask
kind), and routes windowed plans to this path. It also adds native LSE
output and mixed K/V dtype support (bf16/fp16 Q/K + fp8 V) to the modular
kernel, so windowed plans never leave it at run time.

## Changes

- **Composable band masks (correctness fix)**: the `MaskType` enum becomes a
  band `[lo(q), hi(q))` — causal composes with `window_left`; `window_left`
  alone is a left-bounded non-causal window (FA2 semantics); new
  `window_right` plan() parameter for symmetric windows. Loader / MMA /
  softmax / correction all derive trip counts from the same band helpers.
- **Dead work removal**: per-half band iteration (head/tail peel) — each
  128-row half of the CTA runs exactly its own band's trips instead of the
  union with ~1 dead step per stage; warp-ballot rescale skip in correction
  (skips the O-tile TMEM roundtrip when the factor is exactly 1).
- **Mask codegen**: hoisted row-invariant bounds, trace-time-constant K
  coordinates, `cutlass.select_` for masked writes; one shared path for all
  four mask kinds (avoids an MLIR lowering pathology that caused register
  spills on the release toolchain).
- **Runtime window values**: `MaskSpec` is two presence bits; window sizes
  are runtime `Int32` kernel args, causal is a right bound with value 0, and
  the `k < seqlen_k` tail check is always on. Mask compile matrix: 4 kernels.
- **Scheduling by mask kind**: banded masks default to non-persistent launch,
  unmasked stays persistent (`FLASHINFER_CUTE_PREFILL_PERSISTENT` overrides).
- **`sm_scale` default fix**: cute-dsl wrapper plan() now defaults to
  `1/sqrt(head_dim_qk)` (was 1.0, silently unscaled).
- **Vendored FMHA windowed compile fix**: the #3857 path crashed on any
  `window_left >= 0` (no window axis in the trace signature, compile cache,
  or artifact names); it now JIT-compiles with window-presence bits.
- **Native LSE**: `run(return_lse=True)` is computed in the correction warp
  from the final (row_sum, row_max) — log2-domain, matching the vendored
  kernel's convention. Zero cost when off (the non-LSE kernel's SASS is
  byte-identical) and ≲1% when on. Supported for score-mod and sink
  variants; logits-transform variants (no softmax) raise.
- **Mixed K/V dtypes**: bf16/fp16 Q/K with fp8 V. P converts to V's dtype
  for the PV MMA, V slots on the shared K/V smem ring re-arm their barrier
  with V's byte count, and the wrapper JIT-compiles per-V-dtype kernel
  variants at run time (V's dtype is a run-time property; plan()'s
  `kv_data_type` declares K and V together).
- **Routing**: `variant is None and head_dim_qk == 128 and window_left < 0`
  routes to the #3857 vendored kernel; everything else — windows, variants,
  other head dims, any V dtype, `return_lse` — runs the modular path. The
  gate is decided entirely at plan(); there is no run-time fallback.

## Correctness

New causal+window, left-only, and symmetric-window tests in
`tests/attention/test_modular_fmha_prefill.py`, including windowed masks
over mixed-length ragged batches (each sequence gets its own band
geometry; short sequences exercise the item-skip path). Validated on B200 against an
exact f32 reference over {8/8, 32/8, 64/8, 32/32} heads × 5 shapes × 4 masks
in both scheduler modes (152/152), with bitwise-identical outputs across
scheduler modes.

Also new: uniform-fp8 (e4m3) tests calibrated against the vendored kernel's
fp8 error on identical inputs (both ~0.066 max vs f32); mixed bf16-QK/fp8-V
tests against a dequantized-V f32 reference (~0.05 max — S stays bf16-exact);
LSE tests (log2-domain, f32 logsumexp references) covering windowed, ragged
varlen, ALiBi, and sink variants, with the output tensor bit-identical
between LSE and non-LSE runs.

## Performance vs trtllm-gen

B200 (SM100a), torch 2.11+cu130, `nvidia-cutlass-dsl` 4.6.0. CUPTI
kernel-only, median of 30 iters, cold L2; each cell benches both backends
back-to-back in one process, median of 3 passes, single node and session.
GQA 32/8 heads, head_dim 128, bf16. cute-dsl =
`BatchPrefillWithRaggedKVCacheWrapper(backend="cute-dsl")`, i.e. what the
public wrapper runs per mask kind: vendored route for unmasked/causal,
modular kernel for windows. trtllm-gen = paged wrapper, HND, page 16,
preallocated output. Ratio = cute-dsl / trtllm (< 1: cute-dsl faster).
GPU clocks are not lockable on these nodes, so absolute times are only
comparable within a table (each table is built from back-to-back
same-process measurements; the ratios are clock-robust).

Modular vs vendored, why windows route to the modular kernel (B=1 S=16K,
median of 3 passes):

| config | modular (µs) | vendored (µs) | modular/vendored |
|--|--:|--:|:--:|
| unmasked | 3540.4 | 2893.8 | 1.22 |
| causal | 1808.3 | 1532.7 | 1.18 |
| causal + w=127 | 265.0 | 359.8 | 0.74 |
| causal + w=511 | 314.4 | 620.9 | 0.51 |
| causal + w=1023 | 412.0 | 748.1 | 0.55 |

Shape grid (increasing S; per shape: unmasked, causal, windows in increasing
size). The kernel column shows which kernel the wrapper routes to: windowed
rows run this PR's modular path; unmasked/causal rows run the #3857 vendored
kernel and are included for context only.

| B | S | mask | kernel | cute-dsl (µs) | trtllm (µs) | ratio |
|--:|--:|--|--|--:|--:|:--:|
| 64 | 512 | unmasked | vendored | 330.9 | 280.8 | 1.18 |
| 64 | 512 | causal | vendored | 511.7 | 269.4 | 1.90 |
| 64 | 512 | w=256 | modular | 570.9 | 300.6 | 1.90 |
| 16 | 1024 | unmasked | vendored | 249.9 | 236.4 | 1.06 |
| 16 | 1024 | causal | vendored | 318.9 | 185.7 | 1.72 |
| 16 | 1024 | w=256 | modular | 302.7 | 169.2 | 1.79 |
| 1 | 4096 | unmasked | vendored | 201.8 | 227.4 | 0.89 |
| 1 | 4096 | causal | vendored | 159.3 | 150.0 | 1.06 |
| 1 | 4096 | w=256 | modular | 76.4 | 52.5 | 1.46 |
| 1 | 4096 | w=1024 | modular | 115.6 | 87.9 | 1.32 |
| 1 | 4096 | w=2048 | modular | 151.1 | 121.0 | 1.25 |
| 4 | 4096 | unmasked | vendored | 824.4 | 825.0 | 0.99 |
| 4 | 4096 | causal | vendored | 557.9 | 483.7 | 1.15 |
| 4 | 4096 | w=256 | modular | 288.3 | 172.4 | 1.67 |
| 4 | 4096 | w=1024 | modular | 428.5 | 284.5 | 1.51 |
| 4 | 4096 | w=2048 | modular | 519.2 | 399.7 | 1.30 |
| 1 | 16384 | unmasked | vendored | 2845.0 | 3376.5 | 0.86 |
| 1 | 16384 | causal | vendored | 1568.4 | 1758.7 | 0.86 |
| 1 | 16384 | w=256 | modular | 292.8 | 178.1 | 1.62 |
| 1 | 16384 | w=1024 | modular | 415.0 | 304.8 | 1.36 |
| 1 | 16384 | w=2048 | modular | 576.9 | 480.3 | 1.20 |
| 1 | 16384 | w=4096 | modular | 881.7 | 808.0 | 1.09 |
| 1 | 16384 | w=8192 | modular | 1402.0 | 1314.6 | 1.06 |
| 1 | 32768 | unmasked | vendored | 12151.0 | 13847.2 | 0.88 |
| 1 | 32768 | causal | vendored | 6353.5 | 7176.0 | 0.89 |
| 1 | 32768 | w=256 | modular | 580.1 | 334.8 | 1.77 |
| 1 | 32768 | w=1024 | modular | 798.7 | 603.6 | 1.33 |
| 1 | 32768 | w=2048 | modular | 1145.3 | 983.6 | 1.16 |
| 1 | 32768 | w=4096 | modular | 1955.4 | 1740.8 | 1.11 |
| 1 | 32768 | w=8192 | modular | 3443.3 | 3392.4 | 1.01 |

Window-size sweep (B=1, S=16384):

| window | cute-dsl (µs) | trtllm (µs) | ratio |
|--:|--:|--:|:--:|
| 127 | 236.2 | 145.6 | 1.62 |
| 255 | 269.7 | 167.9 | 1.61 |
| 511 | 312.9 | 216.8 | 1.44 |
| 1023 | 411.3 | 304.6 | 1.35 |
| 2047 | 578.0 | 480.0 | 1.20 |
| 4095 | 883.8 | 804.9 | 1.10 |
| 8191 | 1387.7 | 1320.7 | 1.04 |

Mixed V dtype (bf16 Q/K + fp8 V, B=1 S=8192, same protocol): mixed inputs
force the vendored path onto JIT compilation (its prebuilt artifact matrix
has no mixed-dtype axis), where the modular kernel also wins causal:

| config | modular (µs) | vendored-JIT (µs) | modular/vendored |
|--|--:|--:|:--:|
| unmasked | 739.4 | 706.3 | 1.05 |
| causal | 481.8 | 548.1 | 0.88 |
| causal + w=127 | 117.7 | 183.9 | 0.64 |
| causal + w=1023 | 208.4 | 395.2 | 0.53 |

## Summary by CodeRabbit

* **New Features**
  * Added flexible attention band masking, including causal, symmetric, left-only, and right-bounded sliding windows.
  * Added optional log-sum-exp (LSE) output, including variable-length, ALiBi, and attention-sink scenarios.
  * Added support for mixed Q/K/V data types in prefill attention.
  * Added automatic scale selection when no scale is provided.
* **Bug Fixes**
  * Improved kernel selection and fallback behavior for windowed attention and unsupported data-type combinations.
* **Tests**
  * Expanded coverage for masking, FP8, mixed dtypes, LSE results, and ragged inputs.

### DIFF
--- a/flashinfer/attention/cute_dsl/fmha.py
+++ b/flashinfer/attention/cute_dsl/fmha.py
@@ -400,8 +400,14 @@ def cute_dsl_fmha_ragged_prefill(
 
     if kernel_fn is None:
         gpu_arch = _get_gpu_arch(q.device)
+        # Q and K feed the same GEMM operand dtype: neither the prebuilt
+        # artifacts nor the JIT compile (single qk_dtype) support q != k.
+        if q.dtype != k.dtype:
+            raise ValueError(
+                f"cute-dsl FMHA requires q.dtype == k.dtype, got {q.dtype} and {k.dtype}"
+            )
         try:
-            if q.dtype != k.dtype or k.dtype != v.dtype:
+            if k.dtype != v.dtype:
                 raise RuntimeError("Separate dtype_qk / dtype_vo requires JIT")
             if window_left != -1 or window_right != -1:
                 # The exported artifact matrix has no window axis: every
--- a/flashinfer/attention/cute_dsl/fmha.py
+++ b/flashinfer/attention/cute_dsl/fmha.py
@@ -403,6 +403,13 @@ def cute_dsl_fmha_ragged_prefill(
         try:
             if q.dtype != k.dtype or k.dtype != v.dtype:
                 raise RuntimeError("Separate dtype_qk / dtype_vo requires JIT")
+            if window_left != -1 or window_right != -1:
+                # The exported artifact matrix has no window axis: every
+                # prebuilt variant is traced with window_size_left=None (and
+                # window_size_right present only for causal), so a windowed
+                # call into one fails the FFI signature check.  The causal
+                # artifact lookup would falsely succeed here — force JIT.
+                raise RuntimeError("windowed DSL FMHA variants are not exported")
             kernel_fn = get_cute_dsl_fmha_kernel(
                 gpu_arch,
                 q.dtype,
@@ -439,6 +446,8 @@ def cute_dsl_fmha_ragged_prefill(
                 use_skip_softmax,
                 enable_pdl,
                 q.device,
+                has_window_left=window_left != -1,
+                has_window_right=is_causal or window_right != -1,
             )
             # JIT kernels are compiled with --enable-tvm-ffi; the CuTe-native branch
             # would not accept them.
--- a/flashinfer/cute_dsl/attention/fusion/mask.py
+++ b/flashinfer/cute_dsl/attention/fusion/mask.py
@@ -1,20 +1,233 @@
 # Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: BSD-3-Clause
 
-"""Mask types and masking helper functions for attention kernels.
+"""Composable band-mask helpers for attention kernels.
 
-All helpers are standalone @cute.jit functions that take mask_type and
-window_left as compile-time parameters, so they can be reused across
-different kernel variants (prefill, decode).
+The set of KV positions a query row attends to is modeled as a contiguous
+band.  With ``offset = seqlen_k - seqlen_q`` (queries right-aligned to the
+end of KV, the append/prefill-with-cache convention), row ``q`` sees::
+
+    k in [ lo(q), hi(q) )
+    lo(q) = max(0, q + offset - window_left)                if window_left >= 0 else 0
+    hi(q) = min(seqlen_k, q + offset + 1)                   if causal
+          = min(seqlen_k, q + offset + window_right + 1)    elif window_right >= 0
+          = seqlen_k                                        otherwise
+
+``MaskSpec`` describes which bounds exist.  The spec is a compile-time
+parameter: absent bounds compile away via ``cutlass.const_expr``, so e.g. a
+no-mask kernel contains no masking code.  ``causal`` and ``window_left`` are
+independent, matching the FlashInfer convention elsewhere
+(``include/flashinfer/attention/variants.cuh``): ``window_left`` bounds
+lookback and composes with ``causal``; ``window_right`` bounds lookahead for
+non-causal masks (e.g. symmetric windows) and is mutually exclusive with
+``causal``.
+
+All helpers are standalone ``@cute.jit`` functions so they can be reused
+across kernel variants (prefill, decode).
 """
 
-import enum
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, is_dataclass
 
 import cutlass
 import cutlass.cute as cute
 from cutlass.cute.typing import Int32, Float32, Optional
+
+
+@dataclass(frozen=True)
+class MaskSpec:
+    """Compile-time description of the attention mask band.
+
+    Attributes:
+        causal: right bound follows the diagonal (``k <= q + offset``).
+        window_left: max lookback distance; ``-1`` = unbounded.
+        window_right: max lookahead distance for non-causal masks;
+            ``-1`` = unbounded.  Mutually exclusive with ``causal``.
+        check_kv_bounds: some ``seqlen_k`` is not a multiple of the KV tile,
+            so the trailing partial tile needs ``k < seqlen_k`` masking even
+            when no other bound is active.
+    """
+
+    causal: bool = False
+    window_left: int = -1
+    window_right: int = -1
+    check_kv_bounds: bool = False
+
+    def __post_init__(self):
+        if self.causal and self.window_right >= 0:
+            raise ValueError(
+                "window_right is mutually exclusive with causal "
+                "(causal already bounds lookahead at 0)"
+            )
+
+    @property
+    def has_left_bound(self) -> bool:
+        return self.window_left >= 0
+
+    @property
+    def has_right_bound(self) -> bool:
+        return self.causal or self.window_right >= 0
+
+    @property
+    def needs_masking(self) -> bool:
+        return self.has_left_bound or self.has_right_bound or self.check_kv_bounds
+
+
+@cute.jit
+def get_kv_block_range(
+    spec: MaskSpec,
+    blk_coord: cute.Coord,
+    tile_shape: cute.Shape,
+    seqlen_k: Int32,
+    seqlen_q: Int32,
+) -> tuple[Int32, Int32]:
+    """[start, end) KV block range visited for this Q tile (union over rows).
+
+    ``start`` is nonzero only with window_left; ``end`` shrinks with causal
+    or window_right.
+    """
+    qk_offset = seqlen_k - seqlen_q
+    start_block = 0
+    if cutlass.const_expr(spec.window_left >= 0):
+        first_q = blk_coord[0] * tile_shape[0] + qk_offset
+        min_kv = cutlass.max(0, first_q - spec.window_left)
+        start_block = min_kv // tile_shape[1]
+    last_q = (blk_coord[0] + 1) * tile_shape[0] - 1 + qk_offset
+    end_elem = seqlen_k
+    if cutlass.const_expr(spec.causal):
+        end_elem = cutlass.min(seqlen_k, last_q + 1)
+    elif cutlass.const_expr(spec.window_right >= 0):
+        end_elem = cutlass.min(seqlen_k, last_q + spec.window_right + 1)
+    end_block = cute.ceil_div(end_elem, tile_shape[1])
+    return start_block, end_block
+
+
+@cute.jit
+def get_trip_count(
+    spec: MaskSpec,
+    blk_coord: cute.Coord,
+    tile_shape: cute.Shape,
+    seqlen_k: Int32,
+    seqlen_q: Int32,
+) -> Int32:
+    """Number of KV tile blocks to process for this Q tile."""
+    start_block, end_block = get_kv_block_range(
+        spec, blk_coord, tile_shape, seqlen_k, seqlen_q
+    )
+    return end_block - start_block
+
+
+@cute.jit
+def get_trip_segments(
+    spec: MaskSpec,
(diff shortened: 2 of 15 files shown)


flashinfer/attention/cute_dsl/fmha.py
```diff
@@ -403,6 +403,13 @@ def cute_dsl_fmha_ragged_prefill(
         try:
             if q.dtype != k.dtype or k.dtype != v.dtype:
                 raise RuntimeError("Separate dtype_qk / dtype_vo requires JIT")
```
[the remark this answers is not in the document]
pgera: Fixed in f13d017b: `q.dtype != k.dtype` now raises a hard `ValueError` before artifact/JIT resolution (the JIT compile builds both Q and K fake tensors from a single `qk_dtype`, so it could never serve that case); only a separate V dtype falls back to JIT.

pgera:
@flashinfer-bot run

flashinfer/cute_dsl/attention/fusion/mask.py
```diff
@@ -17,6 +34,324 @@
...
+
+@cute.jit
+def apply_mask(
+    spec: MaskSpec,
+    acc_qk: cute.Tensor,
+    index_qk: cute.Tensor,
+    seqlen_k: Int32,
+    causal_offset: Int32,
+    index_qk_static: cute.Tensor,
+    window_left: Int32,
```
PetersonGuo, line 290: window_left and window_right would be layer constant. is there a reason why it was left as a runtime parameter instead? Could we compile kernels with the layer shapes instead?

flashinfer/cute_dsl/attention/wrappers/batch_prefill.py
```diff
@@ -206,10 +207,11 @@ def plan(
         head_dim_qk,
         head_dim_vo=None,
         causal=True,
-        sm_scale=1.0,
+        sm_scale=None,
         q_data_type=torch.float16,
         kv_data_type=torch.float16,
         window_left: int = -1,
+        window_right: int = -1,
```
PetersonGuo: most models are still causal and window_right=0 represents this. Can we default this to 0 instead and remove the causal bool?

tests/attention/test_modular_fmha_prefill.py
```diff
@@ -829,7 +835,13 @@ def attention_sliding_window_ref(
     qk_offset = kv_len - qo_len
     q_idx = torch.arange(qo_len, device=q.device).unsqueeze(1)
     k_idx = torch.arange(kv_len, device=q.device).unsqueeze(0)
-    mask = torch.abs(k_idx - (q_idx + qk_offset)) <= window_left
+    mask = torch.ones(qo_len, kv_len, dtype=torch.bool, device=q.device)
+    if causal:
+        mask &= k_idx <= q_idx + qk_offset
+    elif window_right >= 0:
```
PetersonGuo: The causal branch looks redundant here

flashinfer/cute_dsl/attention/fusion/mask.py
PetersonGuo: would also remove the need for MaskSpec

PetersonGuo (changes requested):

flashinfer/prefill.py
```diff
@@ -3502,7 +3502,14 @@ def plan(
...
             )
             # Route compatible calls to faster trtllm CuTe DSL FMHA kernels.
-            self._cute_dsl_use_fmha = variant is None and head_dim_qk == 128
+            # Sliding-window plans stay on the modular cute-dsl path: it
+            # measures faster than the FMHA route on windows (its banded
+            # masks skip dead work), and the FMHA route's prebuilt artifact
+            # matrix has no windowed variants (windowed calls there need a
+            # JIT compile).
+            self._cute_dsl_use_fmha = (
+                variant is None and head_dim_qk == 128 and window_left < 0
```
saltyminty, line 3511: The modular cutedsl path isn't fully featured, i.e. it doesn't support return_lse and requires QKV to have the same dtype, so the new `window_left < 0` could lead to a unsupported regression compared to the previous path. I believe this warrants additional guards before choosing `self._cute_dsl_use_fmha`

flashinfer/cute_dsl/attention/wrappers/batch_prefill.py
```diff
@@ -230,14 +232,21 @@ def plan(
...
         q_data_type : torch.dtype
             Data type for queries (float16, bfloat16, or float8_e4m3fn).
         kv_data_type : torch.dtype
             Data type for keys/values.
         window_left : int
-            Sliding window size. -1 disables sliding window.
+            Max lookback distance: query ``q`` attends to keys
+            ``k >= q + (kv_len - qo_len) - window_left``. -1 = unbounded.
+            Composes with ``causal`` (causal sliding window attention).
+        window_right : int
```
saltyminty, line 266: For backward compatibility, we should be appending all arguments to the end until the next major release.

saltyminty:
/bot run tests/attention/test_modular_fmha_prefill.py tests/attention/test_cute_dsl_fmha_backend.py

flashinfer/prefill.py
pgera: I added LSE support and mixed KV types ( V= fp8, K = bf16) to the modular path. The features should be at parity now.

flashinfer/cute_dsl/attention/wrappers/batch_prefill.py
pgera: Ack

flashinfer/cute_dsl/attention/wrappers/batch_prefill.py
pgera: The causal bool is a part of the plan API shared across diff backends. So didn't want to break that.

flashinfer/cute_dsl/attention/fusion/mask.py
pgera: The window bounds can be compiled in, but it doesn't show any perf benefits. The core logic depends on the existence of windows. Keeping them runtime vals means that we don't need to JIT compile a new kernel for a different bound. There are 2 flags, so the universe is just 4 types of kernels.

pgera:
@flashinfer-bot run

tests/attention/test_modular_fmha_prefill.py
pgera: Ack

xrq-phys (approved):
`fmha.py` runner changes LGTM. Thanks so much for spotting this!

saltyminty (approved):

PR #3955, opened 2026-07-14
feat: expose trtllm-gen block-sparse attention for decode

PerkzZheng:
## What

Expose trtllm-gen block-sparse attention through `trtllm_batch_decode_with_kv_cache` via a new `enable_block_sparse_attention` flag (decode/generation phase only). Each KV head attends to its own subset of KV cache pages:

- `block_tables`: `[num_kv_heads, batch_size, max_num_pages_per_seq]` — selected page ids packed densely at the front of each row, ascending original order.
- `seq_lens`: `[num_kv_heads, batch_size]` — surviving KV token count per (kv head, sequence).

The KV cache layout is unchanged. `mUseBlockSparseAttention` is a pure runtime parameter of the existing paged-KV generation cubins (added in trtllm-gen), so **no new cubins / kernel-selection changes are required** — this PR only wires the flag (previously hardcoded `false` in `kernelParams.h`) through the runner params, FFI launcher, and Python API.

Also fixes `csrc/fmhaReduction.cu` to index the per-KV-head sequence lengths in the multi-CTA-KV reduction when the flag is set.

Rejected combinations (validated with clear errors): context/prefill phase, sliding window, sparse MLA, skip-softmax, and the non-shared paged-KV index layout (kernel supports it; deferred).

## Tests

New `tests/attention/test_trtllm_gen_block_sparse_decode.py` (78 tests, B200):
- correctness matrix vs a pure-torch gather+SDPA reference with **different random page sets per KV head** (fp16/bf16/fp8 KV, page size 16/64, GQA, spec decode, 8k seqlens exercising the multi-CTA-KV reduction, non-page-aligned per-head lengths);
- full-density per-head tables match the dense kernel bitwise;
- per-head-difference guard (disjoint page sets per head must change only that head's output);
- validation-error coverage.

Regressions on SM100: `test_trtllm_gen_attention_decode.py` 67096 passed, `test_trtllm_gen_attention_prefill.py` 11568 passed, `test_trtllm_gen_mla.py` 7695 passed (cute-dsl cases excluded: pre-existing `nvidia-cutlass-dsl` env mismatch, fails identically on main), `tests/trace` 879 passed.

AI-assisted.

## Release Notes

* **New Features**
  * Added opt-in block-sparse attention for decode/generation via `enable_block_sparse_attention`, including new per–KV-head `seq_lens`/`block_tables` shape requirements.

* **Tests**
  * Added long-running GPU coverage for correctness, dense equivalence, per–KV-head behavior, and input/parameter validation.
  * Added tests ensuring trace-rendered init/reference code executes standalone.

* **Documentation / Trace Support**
  * Added block-sparse TRT-LLM paged decode trace/template support and a new corresponding trace JSON, plus improvements to reference code rendering.


qsang-nv:
Reviewed the wiring and the new tests — a few issues to address:

**1. [P1] Spec decode implicitly requires the speculative tokens' pages to be selected, but the API doesn't say so.**
The causal boundary is derived from the compacted per-head length, both inside the generation cubins and in the reduction fix ([fmhaReduction.cu#L86-L104](https://github.com/flashinfer-ai/flashinfer/blob/2be18d1d7e931c3338633ba0813babc5a0657c80/csrc/fmhaReduction.cu#L86-L104)): query `i` attends to compacted positions `j <= L - q_len_per_req + i`. That is only correct when the last `q_len_per_req` compacted tokens are exactly the speculative tokens. If a per-head table drops any page holding speculative-token KV, the surviving tokens are historical and should be fully visible, yet the kernel still masks the trailing ones — silently wrong output, no error.

The tests [partially encode this assumption](https://github.com/flashinfer-ai/flashinfer/blob/2be18d1d7e931c3338633ba0813babc5a0657c80/tests/attention/test_trtllm_gen_block_sparse_decode.py#L194-L206) by always retaining the tail page and selecting at least two pages. However, the second page is random, so when the speculative-token suffix crosses a page boundary, the preceding page is not guaranteed to be selected. The torch reference also applies the same compacted-length causal formula, so it cannot detect this contract violation — kernel and reference stay consistent while both diverge from the intended semantics.

Since the mask semantics is burned into the cubins and the tables live on device (no cheap host-side validation), please document the exact contract under `enable_block_sparse_attention`:
- when `q_len_per_req > 1`, all pages intersecting the original KV suffix that contains the speculative tokens must be selected, so that the last `q_len_per_req` compacted tokens are the speculative ones;
- each per-head surviving length must be `>= q_len_per_req`, otherwise the first query's boundary becomes non-positive.

It would also be good to make the test's page selection explicitly include all speculative-suffix pages (rather than relying on the tail page alone), so the tested domain matches the documented contract.

**2. [P2] Block-sparse calls emit a corrupt fi_trace schema.**
The API is bound to the static [`trtllm_batch_decode_trace`](https://github.com/flashinfer-ai/flashinfer/blob/2be18d1d7e931c3338633ba0813babc5a0657c80/flashinfer/trace/templates/attention.py#L1773-L1815) template, which declares `block_tables: [batch_size, max_pages_per_seq]` and `seq_lens: [batch_size]`. The axis extractors read shape dims positionally without a rank check, so a sparse call binds `batch_size = num_kv_heads` and `max_pages_per_seq = batch_size`, and the dumped trace cannot reproduce the call. `flashinfer_api(trace=...)` already accepts a dispatch callable that selects a template from the invocation kwargs — please add a block-sparse template (with its own reference/init and example JSON) behind a dispatch on `enable_block_sparse_attention`, or at minimum skip trace emission for sparse calls so no corrupt trace is produced.

**3. [P3] Explicit `skip_softmax_threshold_scale_factor=0.0` is wrongly rejected.**
The [Python check](https://github.com/flashinfer-ai/flashinfer/blob/2be18d1d7e931c3338633ba0813babc5a0657c80/flashinfer/decode.py#L3276-L3279) rejects any non-`None` value, but the C++ layer (`value_or(0.0f)`, `skips_softmax = value != 0.0f`) and the dense path both treat `0.0` as disabled. Only nonzero values need rejecting, so explicitly passing the default (common with config-driven callers) doesn't fail.

PerkzZheng:
@qsang-nv thanks for the comments,
1. it is not needed as speculative decoding is a causal mask even for block sparse, which means only current token_q's previous kv tokens will be attended, and only the last visible tokenKv's block will be force selected. 
2. this has been addressed in the latest commit
3. I think it is not necessary for now since we don't expect the two features used together anyway.

qsang-nv:
**Follow-up on the fi_trace fix:** the new dispatch/template resolves the schema mismatch, but the checked-in JSON is not runnable standalone. `exec()`-ing the dumped `init` and `reference` strings and calling them raises:

- `init`: `NameError: name '_trtllm_batch_decode_init' is not defined`
- `reference`: `NameError: name '_trtllm_kv_from_cache' is not defined`

Both [the new reference](https://github.com/flashinfer-ai/flashinfer/blob/f64e0bfdfdde34f86bbe83383e78033249fd6e71/flashinfer/trace/templates/attention.py#L1823-L1880) and [the new init](https://github.com/flashinfer-ai/flashinfer/blob/f64e0bfdfdde34f86bbe83383e78033249fd6e71/flashinfer/trace/templates/attention.py#L1883-L1928) call module-level helpers, but the renderers only export the target function's own source: `_render_init_source` inlines `_init_helpers` plus functions listed in `_trace_init_dependencies`, and `_render_reference_source` has no dependency capture at all.

Suggested fix:
- `init`: attach `_trace_init_dependencies = (_trtllm_batch_decode_init,)` — the existing convention already used by the moe/page/quantize templates; `_trtllm_batch_decode_init` itself only needs `torch`/`math`, so no transitive deps.
- `reference`: either inline the `_trtllm_kv_from_cache` logic, or add an equivalent dependency-capture to `_render_reference_source`. The renderer route is preferable: the dense `_trtllm_batch_decode_reference` calls the same helper, so any future dense dump would hit the same NameError.
- Add a test that `exec()`s the rendered `init`/`reference` strings from a dumped JSON and invokes them once — dump-time verification runs the live module functions (where the helpers resolve), so it cannot catch broken rendered artifacts, and the consistency tests never execute these strings.

After the fix, please regenerate `trtllm_batch_decode_block_sparse_h16_kv2_d128_ps16.json`.

PerkzZheng:
Good catch — the rendered strings were indeed not standalone (dump-time verification calls the live module functions, so the missing inlining was invisible). Fixed in e370a0bc:

qsang-nv (approved):

saltyminty (approved):

PR #3962, opened 2026-07-14
feat: fused FP8 quantization output for LayerNorm

elwhyjay:
## 📌 Description

FlashInfer provides fused norm + FP8 quantization for the RMSNorm family (`rmsnorm_quant`, `fused_add_rmsnorm_quant`), but not for LayerNorm. Models that use LayerNorm (OPT, Falcon, BERT family, ViT/DiT variants, Whisper/CLIP style encoders) currently have to run layernorm and quantization as two kernels, which costs about 7 bytes of traffic per element instead of 3.

The kernel side is already there: `generalLayerNorm` in `include/flashinfer/norm.cuh` implements per-tensor and per-token quantization paths, but the `LayerNorm` host launcher passes nullptr for all quant arguments, with a note:

    // TODO(kaixih): add support for fp8 quantization if needed

This PR connects the existing per-tensor path rather than writing a new kernel, so the diff stays small.

- Adds `flashinfer.layernorm_quant(out, input, gemma, beta, scale, eps)`. Output dtype (float8_e4m3fn or float8_e5m2) is taken from the preallocated `out` tensor, same as `rmsnorm_quant`.
- Adds fp8 `QuantTypeStaticVals` specializations (constants match TRT-LLM `quantTypeUtils.cuh`) and the missing e5m2 `cuda_cast` specializations. The FP8 dispatch macro instantiates both e4m3 and e5m2, so the e5m2 casts are needed for compilation.
- Adds a `LayerNormQuant` host launcher, guarded by `ENABLE_FP8`. The csrc launcher keeps a non-FP8 build working (`test_norm_compilation_without_fp8` passes).
- Changes the per-tensor scale in `generalLayerNorm` to be applied as division (`out = normed / scale`). The kernel originally multiplied, but this path was unreachable from every existing call site (the pointer is always nullptr), so no current behavior changes. Keeping the multiply semantics would make `layernorm_quant` and `rmsnorm_quant` disagree on what `scale` means, which seemed worse for fusion passes that target both. Happy to adjust if there is a concern I am missing here.
- Trace template, docs entry, tests, and a `layernorm_quant` benchmark routine.
## Performance

RTX 5090, bf16 input, e4m3 output, eps 1e-6. Fused kernel vs. unfused baseline (`flashinfer.layernorm` followed by eager torch div/clamp/cast), median over 100 iterations under CUDA graph:

| batch | hidden | unfused (us) | fused (us) | speedup |
|------:|-------:|-------------:|-----------:|--------:|
|   256 |   4096 |         13.2 |        6.9 |   1.91x |
|  1024 |   4096 |         36.9 |       21.4 |   1.73x |
|  4096 |   4096 |        178.9 |       71.8 |   2.49x |
|   256 |   8192 |         21.1 |        8.4 |   2.52x |
|  1024 |   8192 |         69.2 |       24.9 |   2.77x |
|  4096 |   8192 |        499.6 |       80.6 |   6.20x |
|  1024 |  16384 |        167.2 |       43.8 |   3.82x |
|  4096 |  16384 |       1066.5 |      148.7 |   7.17x |

The eager baseline materializes intermediate tensors for div/clamp/cast, so large shapes gain more than the pure traffic ratio (7B vs 3B per element, about 2.3x). At batch 4096 / hidden 16384 the fused kernel reaches about 1.35 TB/s on this card.

Reproduce with:

    python benchmarks/flashinfer_benchmark.py --routine layernorm_quant \
        --batch_size 1024 --hidden_size 4096 --refcheck

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

- `tests/utils/test_norm.py::test_layernorm_quant`: e4m3 and e5m2, batch 1 to 989, hidden 111 to 16384 (odd sizes cover the non-vectorized path), scale 0.01/1.0/10.0, tolerance rtol=1/atol=1 following the existing FP8 norm tests. 168 cases pass.
- `test_layernorm_quant_invalid_inputs`: rejects fp16 input, non-contiguous input, and non-scalar scale.
- `test_norm_compilation_without_fp8` passes unchanged (norm module still builds without `ENABLE_FP8`).
- Full `tests/utils/test_norm.py` passes (2887 tests).
- `pytest tests/trace/` template consistency and end-to-end tests pass for the new trace.
- CUDA graph capture/replay smoke test passes; scale is read on device at kernel execution time, so updating the scale tensor between replays works.

## Summary by CodeRabbit

* **New Features**
  * Added `flashinfer.layernorm_quant` (LayerNorm + FP8 quantization) with CUDA support across supported compute capabilities.
  * Exposed the API at both `flashinfer.norm` and the top-level `flashinfer` namespace.
  * Added benchmark and trace coverage for the quantized LayerNorm operation.
* **Documentation**
  * Updated normalization API docs to include `layernorm_quant`.
* **Tests**
  * Added correctness/invalid-input tests, plus trace example coverage and expected outputs for `layernorm_quant`.
* **Bug Fixes**
  * Improved trace output dtype resolution when traced tensors are not available.


benchmarks/routines/norm.py
```diff
@@ -1015,6 +1017,196 @@ def run_backend(backend, out_tensor, input_tensor, weight):
...
+            median_time = np.median(backend_times[backend])
+            std_time = np.std(backend_times[backend])
+
+            # Memory bandwidth calculation for LayerNorm + Quant
+            # Read: input tensor + gamma/beta tensors
+            # Write: output tensor (quantized, smaller dtype)
+            num_elements = np.prod(input_shape)
+            problem_bytes = (
+                num_elements * input_dtype.itemsize  # input read
+                + 2 * hidden_size * torch.float32.itemsize  # gamma and beta read
```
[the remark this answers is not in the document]
elwhyjay: `torch.dtype` does not seem to have element_size(). I also ran this routine end to end with --refcheck on torch 2.9 and it completed without errors.

elwhyjay:
Gentle ping @kahyunnam @yzh119 @bkryu @yongwww .Would you be able to review this when you get a chance? Thank you.

bkryu:
Thank you @elwhyjay for the PR. Adding the motivation in the description helps us understand.

@kahyunnam , the PR looks generally good to me, but can you take another pass?

tests/utils/test_norm.py
```diff
@@ -347,6 +357,46 @@ def test_layernorm(batch_size, hidden_size, dtype):
...
+    x = torch.randn(batch_size, hidden_size, dtype=torch.bfloat16, device="cuda")
+    gamma = torch.randn(hidden_size, dtype=torch.float32, device="cuda")
+    beta = torch.randn(hidden_size, dtype=torch.float32, device="cuda")
+    scale = torch.tensor([quant_scale], dtype=torch.float32, device="cuda")
+
+    y_ref = layer_norm_quant(x, gamma, beta, scale, quant_dtype)
+    y = torch.empty_like(x, dtype=quant_dtype)
+    flashinfer.norm.layernorm_quant(y, x, gamma, beta, scale)
+
+    torch.testing.assert_close(y_ref.float(), y.float(), rtol=1, atol=1)
```
kahyunnam: rtol/atol=1 seems a bit loose?

kahyunnam (approved):
small nit but overall looks good! approving

tests/utils/test_norm.py
elwhyjay: Thank you for the review @kahyunnam . I agree with it. rtol/atol=1 only bounds a per-element FP8 step, so a systematic off-by-one would still pass. Since ref and kernel share the same quantize path, how about checking the fraction of mismatched elements instead?
```python
tol = torch.finfo(quant_dtype).eps * y_ref.float().abs()  # one FP8 step at |ref|
mismatched = ((y_ref.float() - y.float()).abs() > tol).float().mean().item()
assert mismatched < 0.01
```
This keeps the FP8 rounding slack but catches a systematic error. Let me know what you think.

kahyunnam: `finfo(fp8).eps` can change a bit depending on e4m3 vs e5m2; let's reuse an existing helper: `flashinfer.trace.default_check` + default_tolerances (FP8 defaults: rtol=0.1, atol=0.1) supports max_mismatch_pct. Trace templates for FP8 norm use the same family of checks.

also, the ref and kernel slightly differ: right now the pytest ref is fp32-only, let's round normalized output to bf16 before / scale and FP8 cast (same as the benchmark refcheck)

elwhyjay: Thanks @kahyunnam. Applied it. PTAL

kahyunnam:
Thank you for your efforts here @elwhyjay ! Setting this to automerge when pre-merge passes.

PR #3912, opened 2026-07-10
fix: autotuner memory leak follow up

jdebache:
## 📌 Description

In #3687, we partially fixed the identity of objects used to identify cache entries in the cache of `_find_nearest_profile`. However, as correctly pointed out in #3810, we did not handle the initializer for the MoE TopK IDs.

This PR fixes this gap and adds a test to preserve the correct behaviour.

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [X] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [X] I have installed the hooks with `pre-commit install`.
- [X] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

## 🧪 Tests

- [X] Tests have been added or updated as needed.
- [X] All tests are passing (`unittest`, etc.).

## Summary by CodeRabbit

* **Performance**
  * Improved Mixture-of-Experts autotuning efficiency by reusing the expert-routing initialization used to build tuning configurations.

* **Reliability / Bug Fixes**
  * Improved autotuner profile-cache stability by ensuring equivalent tuning configurations deduplicate correctly.
  * Updated integration-test cache injection to target the correct operator entry.

* **Tests**
  * Added regression coverage for profile-cache deduplication and verifying cache growth when new initializers are recreated.
  * Made VSA BlockSparseAttention tests conditional on the presence of `quack`.

* **Chores**
  * Updated benchmark cache-update keying to match loader expectations.

PR #3970, opened 2026-07-14
Autotuner mem leak follow up 2

jdebache:
## 📌 Description

This PR fixes two autotuner cache issues that can degrade long-running MLA decode workloads or leave on-disk tuning caches permanently unusable.

- Memoizes MLA decode `TuningConfig` objects using the stable `(buckets, num_pages, profile_seq_len)` inputs. This reuses initializer closures across equivalent dispatcher calls and prevents unbounded growth in `_find_nearest_profile` caused by fresh, unequal closures on every call.
- Distinguishes hard metadata mismatches from indeterminate metadata in persisted autotuner caches:
  - Definite environment conflicts remain protected from reads and overwrites.
  - Missing, `unknown`, or legacy `None` metadata is ignored on load and replaced with freshly tuned configurations on the next save, allowing the cache to heal.
- Re-checks cache metadata at save time so concurrent cache replacements are evaluated against the latest on-disk state.
- Improves cuDNN version detection by falling back to the cuDNN frontend when PyTorch has not initialized cuDNN.
- Deduplicates repeated metadata-mismatch warnings by severity while retaining details at debug level.
- Updates the autotuning documentation and adds regression coverage for MLA config reuse, cache growth, metadata classification, cache healing, hard-mismatch protection, concurrent save behavior, and cuDNN version detection.

## 🔍 Related Issues

- Follow-up to #3912

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

Added and updated tests cover:

- Stable MLA decode tuning-config identity and `_find_nearest_profile` cache size.
- Hard, soft, missing, wildcard, and corrupt metadata handling.
- Healing indeterminate cache files while preserving definite cross-environment conflicts.
- Save-time metadata revalidation and warning deduplication.
- PyTorch and cuDNN frontend version-detection paths.

## Reviewer Notes

Please focus on the hard-versus-indeterminate metadata classification and the save-time behavior: a definite mismatch must never overwrite another environment's cache, while an indeterminate cache must be replaceable so it does not force repeated retuning. For the MLA change, please verify that `(buckets, num_pages, profile_seq_len)` fully captures the values used by the memoized tuning configuration.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved autotuning cache compatibility by distinguishing hard vs soft environment-metadata mismatches.
  * Hard conflicts now skip cache usage; soft/indeterminate mismatches trigger “cache healing” on the next tuned save.
  * Added clearer, one-time-per-process warnings for mismatch details, and ensures poisoned caches can be overwritten.
  * Autotuning now records the cuDNN **backend** version for more reliable compatibility checks.
* **Performance**
  * Reuses memoized MLA decode tuning configurations for equivalent inputs.
* **Documentation**
  * Updated autotuning docs to describe hard/soft mismatch and cache healing behavior.
* **Tests**
  * Expanded coverage for mismatch classification, healing behavior, warning deduplication, and cuDNN backend detection.

### DIFF
--- a/docs/autotuning.rst
+++ b/docs/autotuning.rst
@@ -230,12 +230,25 @@ The ``_metadata`` key records the environment that created the cache file
 
 On load, ``_metadata`` is compared against the current environment.  If any
 field differs (e.g. different GPU, FlashInfer version, or cuBLAS version),
-the **entire cache is skipped** — no configs are loaded, and the file will
-not be overwritten on exit; i.e. the autotuner would behave as if the cache
-file input was not provided (cache=None). This prevents silently using invalid
-tactics and avoids destroying configs tuned for a different environment.  A
-warning is logged with the mismatch details and a suggestion to use a
-different cache path for the current environment.
+the **entire cache is skipped** — no configs are loaded.  This prevents
+silently using invalid tactics.  A warning is logged once per process with
+the mismatch details; repeats for further cache files are logged at DEBUG.
+
+What happens to the file on the next save depends on the kind of mismatch:
+
+* The saved value is **definite** (e.g. ``"cudnn_version": "91900"`` vs
+  current ``92101``): the file belongs to a different environment sharing
+  this cache path.  It is never overwritten — the autotuner behaves as if
+  ``cache=None`` — so configs tuned for that environment survive.  Use a
+  different cache path for the current environment, or delete the file to
+  re-prime it.
+* The saved value is **indeterminate** (``"unknown"``, recorded when the
+  writer could not detect e.g. its cuDNN version, or a field missing from
+  files written by older FlashInfer versions): the file's configs cannot
+  be attributed to any environment, so the next tuned save **replaces**
+  the file with freshly tuned configs stamped with the current
+  environment's metadata.  The cache heals itself instead of forcing a
+  full re-tune on every startup forever.
 
 Advanced users can bypass individual checks by manually editing the JSON file
 and setting a metadata field to ``"*"``.  For example, setting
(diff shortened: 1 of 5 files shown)


qiching (approved):
Make sense to me. LGTM.

i have a qq: In `_classify_metadata_mismatches`, when both saved and current metadata are indeterminate, when both saved and current cudnn_version are "unknown", they compare equal and the cache is loaded as a match, is that intended, or should equal but indeterminate still be ignored on load?

jdebache:
Yeah I was also wondering about that. This seems to be the current behaviour, so I didn't want to change it. But I also think that ideally one would never accept to load something that is "unknown" from disk. Which then raises the question: why even save "unknown" tagged results to disk at all?

qiching:
> Yeah I was also wondering about that. This seems to be the current behaviour, so I didn't want to change it. But I also think that ideally one would never accept to load something that is "unknown" from disk. Which then raises the question: why even save "unknown" tagged results to disk at all?

makes sense to keep it out of this PR. Might be worth a follow-up to just never persist or load unknown-tagged entries.

wongsingfo:
Thanks for addressing the MLA path by memoizing `_mla_decode_tuning_config`.

However, this does not cover the MoE side of the original performance issue. `MoERunner._make_tuning_config()` still constructs a new `TuningConfig` on every invocation:

https://github.com/jdebache/flashinfer/blob/b58a53c5a5a97866c63d6d6e5da956d2b6480de7/flashinfer/fused_moe/core.py#L1231-L1316

Although these configurations are logically equivalent, repeatedly constructing them and using the full `TuningConfig` as part of the `_find_nearest_profile` cache key can still incur significant hashing and equality-comparison overhead on the forward path.

I opened #3984 as an attempt to address this more generally by normalizing the nearest-profile cache key to include only the fields actually used for profile selection. This should cover both the MLA and MoE paths.

PR #3976, opened 2026-07-14
feat(moe): add deterministic CuTe DSL NVFP4 finalize mode

zianglih:
## 📌 Description

@humansand

Add an optional bitwise-deterministic finalize mode to CuTe DSL NVFP4 MoE for eager execution and CUDA graph replay.

The existing GEMM2 fused finalize atomically scatters routed expert outputs into each token row. Its accumulation order can vary when a token has multiple local routes, causing bitwise output drift across identical launches.

This PR mirrors the `cutlass_fused_moe` contract:

- `use_fused_finalize=True` remains the default for the existing atomic fused path and best performance.
- `use_fused_finalize=False` makes GEMM2 write each active route to a unique expanded `(token, top-k slot)` row, then uses `moe_unpermute` to apply router weights and reduce top-k routes in fixed order.
- `bench_moe_deepseek.py --functional-api --no-fused-finalize` selects the deterministic path for matched performance comparisons.
- `moe_unpermute` receives PyTorch's current CUDA stream explicitly so the reduction is captured and replayed by CUDA graphs.
- The unified MoE fuzz contract continues to classify the default CuTe DSL atomic path as nondeterministic; the existing CUDA graph replay test now exercises both finalize modes.

The existing CuTe DSL CUDA graph replay matrix now runs both finalize modes: the fused path retains its FP4 tolerance, while the deterministic path requires bitwise-identical replay output.

## 🔍 Related Issues

- Follow-up to #3645.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

> If you are unsure about how to set up `pre-commit`, see the [pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

Validated on NVIDIA B200 with CUDA 13:

```bash
FLASHINFER_DISABLE_VERSION_CHECK=1 python3 -m pytest -s \
  tests/moe/test_cute_dsl_fused_moe.py::TestCuteDslMoEWrapper::test_wrapper_accuracy \
  tests/moe/test_cute_dsl_fused_moe.py::TestCuteDslMoEWrapper::test_wrapper_cuda_graph \
  tests/moe/test_cute_dsl_fused_moe.py::TestExpertParallelism::test_functional_with_ep
# 76 complete matrix cases passed as part of an 81-case B200 run

CUDA_VISIBLE_DEVICES=0 FLASHINFER_DISABLE_VERSION_CHECK=1 python3 -m pytest -q \
  tests/moe/test_cute_dsl_fused_moe.py \
  -k test_deterministic_finalize_numerical_accuracy
# 2 passed in 20.25s

pre-commit run --all-files
# Passed
```

The 76-case regression matrix comprises the complete 48-case wrapper accuracy matrix, complete
24-case CUDA graph matrix, and all four functional EP combinations. The original
144-case functional numerical matrix remains unchanged for the default fused
finalize path. Deterministic-finalize numerical coverage is isolated to two
representative cases spanning both activations, both activation-scaling modes,
and small and large top-k shapes; both passed on B200.

## 📊 Benchmark

Measured on NVIDIA B200 with CUDA 13, CUDA graphs enabled, per-token activation scaling, and 4over6 disabled. Timing used CUPTI with `cupti-python==13.3.0`, `nvidia-cuda-cupti==13.3.75`, and `cuda-python==13.2.0`. A direct `bench_gpu_time(..., enable_cupti=True)` smoke test returned five samples with a 0.001024 ms median, and neither benchmark log reported a missing-CUPTI or timing-fallback warning.

```bash
# Atomic fused finalize (default)
python3 benchmarks/bench_moe_deepseek.py \
  --functional-api \
  --use-per-token-activation \
  --num-tokens 1,2,4,8,16,32,64,128,256,512,1024,2048,4096

# Deterministic two-stage finalize
python3 benchmarks/bench_moe_deepseek.py \
  --functional-api \
  --use-per-token-activation \
  --no-fused-finalize \
  --num-tokens 1,2,4,8,16,32,64,128,256,512,1024,2048,4096
```

| Tokens | Atomic fused (ms) | Deterministic (ms) | Deterministic overhead |
|---:|---:|---:|---:|
| 1 | 0.064 | 0.073 | +14.1% |
| 2 | 0.100 | 0.112 | +12.0% |
| 4 | 0.156 | 0.174 | +11.5% |
| 8 | 0.263 | 0.282 | +7.2% |
| 16 | 0.444 | 0.465 | +4.7% |
| 32 | 0.693 | 0.715 | +3.2% |
| 64 | 0.933 | 0.953 | +2.1% |
| 128 | 1.023 | 1.045 | +2.2% |
| 256 | 1.062 | 1.078 | +1.5% |
| 512 | 1.073 | 1.088 | +1.4% |
| 1024 | 1.091 | 1.129 | +3.5% |
| 2048 | 1.138 | 1.200 | +5.4% |
| 4096 | 1.520 | 1.663 | +9.4% |

Geometric-mean deterministic overhead is **5.94%** across the full sweep, **7.03%** for 1-128 tokens, and **4.21%** for 256-4096 tokens. Values are calculated from the benchmark's printed millisecond results.

## Reviewer Notes

Please focus on the deterministic GEMM2 output layout and the `input_is_expanded` mode added to `moe_unpermute`. The deterministic `use_fused_finalize=False` path adds an expanded-route workspace and a separate fixed-order reduction kernel; the atomic fused finalize remains the default, matching `cutlass_fused_moe`.

### Why normalized `activation_type` uses a local `activation` binding

`normalize_cute_dsl_moe_activation_type` returns an `ActivationType`, while the existing API parameter is annotated as `int`. Reassigning the normalized value back to `activation_type` leaves mypy treating it as an `int`, so subsequent `.value` and `.name` accesses fail type checking. The local `activation` binding preserves the normalized enum type without adding casts, type ignores, or unrelated public annotation changes. This is typing-only and does not change runtime behavior.

### Why `moe_unpermute` receives an explicit CUDA stream

The deterministic finalize calls `moe_unpermute` from PyTorch and may run inside a non-default stream or CUDA graph capture. The binding function `get_current_stream()` resolves through `TVMFFIEnvGetStream`, which does not track `torch.cuda.current_stream()`. Python therefore passes the current PyTorch stream pointer explicitly so the reduction is ordered after GEMM2 and captured in the same CUDA graph. All four dtype-specialized `moe_unpermute` bindings carry the argument because Python selects among them from the input and routing-scale dtypes. A zero pointer retains the existing TVM FFI stream fallback. This follows the established `moe_sort` and `moe_output_memset_inplace` pattern.

### Why this CuTe DSL change touches CUDA C++

The CuTe DSL MoE pipeline is hybrid: GEMM1 and GEMM2 are CuTe DSL kernels, while routing and finalize utilities such as `moe_sort`, `moe_permute`, and `moe_unpermute` are CUDA C++ kernels loaded through the `moe_utils` JIT module. The deterministic path calls the existing CUDA C++ `moeUnpermute` kernel after the CuTe DSL GEMM2 completes; the atomic fused path does not call it.

Previously, `moeUnpermute` assumed its input rows were in expert-permuted order and resolved every `(token, top-k slot)` through `expanded_idx_to_permuted_idx`. The deterministic GEMM2 path instead writes each route exactly once into a compact `[num_tokens * top_k, hidden_size]` buffer in expanded `(token, top-k slot)` order. `input_is_expanded=True` changes only the source-row selection to use that expanded index directly. The mapping is still checked so masked or non-local EP routes, represented by a negative permuted index, are skipped.

Because this utility crosses the Python/JIT boundary, the new layout mode must be threaded through `flashinfer/fused_moe/cute_dsl/moe_utils.py`, `csrc/moe_utils_binding.cu`, the C++ declaration, and the CUDA kernel launcher. Keeping the old interface would require either a larger expert-padded output workspace or a duplicate reduction kernel.

## Summary by CodeRabbit

* **New Features**
  * Added `use_fused_finalize` option to switch MoE finalization between fused (default) and deterministic modes, including the resulting output layout.
  * Added `input_is_expanded` support for MoE unpermutation to correctly handle expanded vs unexpanded token layouts.
  * Improved MoE unpermutation execution by honoring an explicit CUDA stream.
  * Added a benchmark CLI flag to compare fused vs deterministic finalize modes.
* **Bug Fixes**
  * Normalized activation handling consistently across MoE execution paths.
* **Tests**
  * Expanded accuracy and CUDA-graph coverage for both finalize modes, with replay checks adjusted per mode.

### DIFF
--- a/csrc/nv_internal/tensorrt_llm/kernels/cuteDslKernels/moeUtils.cu
+++ b/csrc/nv_internal/tensorrt_llm/kernels/cuteDslKernels/moeUtils.cu
@@ -190,7 +190,7 @@ __global__ void moeUnpermuteKernel(InputType const* permuted_input, InputType* o
       auto const* src_ptr =
           reinterpret_cast<ElemCopyType const*>(permuted_input) + input_idx * kCopyPerToken;
       *reinterpret_cast<ElemCopyType*>(rmem) = src_ptr[i];
-      TopKScaleType const scale = topk_scales[token_idx * top_k + k];
+      TopKScaleType const scale = topk_scales[expanded_idx];
 
 #pragma unroll
       for (int32_t j = 0; j < kElemPerCopy; j++) {
--- a/csrc/nv_internal/tensorrt_llm/kernels/cuteDslKernels/moeUtils.cu
+++ b/csrc/nv_internal/tensorrt_llm/kernels/cuteDslKernels/moeUtils.cu
@@ -160,7 +160,7 @@ template <typename InputType, typename TopKScaleType, int32_t kThreadsPerBlock>
 __global__ void moeUnpermuteKernel(InputType const* permuted_input, InputType* output,
                                    int32_t const* expanded_idx_to_permuted_idx,
                                    TopKScaleType const* topk_scales, int32_t const hidden_size,
-                                   int32_t const top_k) {
+                                   int32_t const top_k, bool const input_is_expanded) {
   using AccumType = float;
   int32_t constexpr kElemPerCopy = elemPerCopy<InputType>();
   // Need int64_t to prevent overflow when computing pointer offsets.
@@ -181,12 +181,14 @@ __global__ void moeUnpermuteKernel(InputType const* permuted_input, InputType* o
       rmemAccum[j] = 0;
     }
     for (int32_t k = 0; k < top_k; k++) {
-      int32_t const permuted_idx = expanded_idx_to_permuted_idx[token_idx * top_k + k];
+      int32_t const expanded_idx = token_idx * top_k + k;
+      int32_t const permuted_idx = expanded_idx_to_permuted_idx[expanded_idx];
       if (permuted_idx < 0) {
         continue;
       }
+      int32_t const input_idx = input_is_expanded ? expanded_idx : permuted_idx;
       auto const* src_ptr =
-          reinterpret_cast<ElemCopyType const*>(permuted_input) + permuted_idx * kCopyPerToken;
+          reinterpret_cast<ElemCopyType const*>(permuted_input) + input_idx * kCopyPerToken;
       *reinterpret_cast<ElemCopyType*>(rmem) = src_ptr[i];
       TopKScaleType const scale = topk_scales[token_idx * top_k + k];
 
@@ -211,7 +213,7 @@ template <typename InputType, typename TopKScaleType>
 void moeUnpermute(InputType const* permuted_input, InputType* output,
                   int32_t const* expanded_idx_to_permuted_idx, TopKScaleType const* topk_scales,
                   int32_t const num_tokens, int32_t const hidden_size, int32_t const top_k,
-                  bool enable_pdl, cudaStream_t stream) {
+                  bool input_is_expanded, bool enable_pdl, cudaStream_t stream) {
   int32_t constexpr kThreadsPerBlock = 256;
   int32_t constexpr kElemPerCopy = elemPerCopy<InputType>();
   TLLM_CHECK_WITH_INFO(hidden_size % kElemPerCopy == 0, "hidden_size must be divisible by %d.",
@@ -233,15 +235,15 @@ void moeUnpermute(InputType const* permuted_input, InputType* output,
   config.numAttrs = 1;
   config.attrs = attrs;
   cudaLaunchKernelEx(&config, kernel, permuted_input, output, expanded_idx_to_permuted_idx,
-                     topk_scales, hidden_size, top_k);
+                     topk_scales, hidden_size, top_k, input_is_expanded);
 }
 
-#define INSTANTIATE_MOE_UNPERMUTE(InputType, TopKScaleType)                                  \
-  template void moeUnpermute<InputType>(InputType const* permuted_input, InputType* output,  \
-                                        int32_t const* expanded_idx_to_permuted_idx,         \
-                                        TopKScaleType const* topk_scales,                    \
-                                        int32_t const num_tokens, int32_t const hidden_size, \
-                                        int32_t const top_k, bool enable_pdl, cudaStream_t stream)
+#define INSTANTIATE_MOE_UNPERMUTE(InputType, TopKScaleType)                          \
+  template void moeUnpermute<InputType>(                                             \
+      InputType const* permuted_input, InputType* output,                            \
+      int32_t const* expanded_idx_to_permuted_idx, TopKScaleType const* topk_scales, \
+      int32_t const num_tokens, int32_t const hidden_size, int32_t const top_k,      \
+      bool input_is_expanded, bool enable_pdl, cudaStream_t stream)
 
 INSTANTIATE_MOE_UNPERMUTE(half, float);
 INSTANTIATE_MOE_UNPERMUTE(half, half);
--- a/flashinfer/fused_moe/cute_dsl/blackwell/utils.py
+++ b/flashinfer/fused_moe/cute_dsl/blackwell/utils.py
@@ -315,6 +315,25 @@ def atomic_add_func(rOut_epi_packed, scatter_out_offset, loc=None, ip=None):
 
 
 @dsl_user_op
+def blk_copy(dst_gemm, src_smem, size, loc=None, ip=None):
+    llvm.inline_asm(
+        None,
+        [
+            dst_gemm.iterator.toint(loc=loc, ip=ip).ir_value(loc=loc, ip=ip),
+            src_smem.iterator.toint(loc=loc, ip=ip).ir_value(loc=loc, ip=ip),
+            size.ir_value(loc=loc, ip=ip),
+        ],
+        "cp.async.bulk.global.shared::cta.bulk_group [$0], [$1], $2;",
+        "l,r,r",
+        has_side_effects=True,
+        is_align_stack=False,
+        asm_dialect=llvm.AsmDialect.AD_ATT,
+        loc=loc,
+        ip=ip,
+    )
+
+
+@dsl_user_op
 def blk_reduce_bf16(dst_gemm, src_smem, size, loc=None, ip=None):
     llvm.inline_asm(
         None,
--- a/flashinfer/fused_moe/cute_dsl/fused_moe.py
+++ b/flashinfer/fused_moe/cute_dsl/fused_moe.py
@@ -77,6 +77,7 @@ from ...utils import supported_compute_capability
 from .moe_utils import (
     moe_output_memset_inplace,
     moe_sort,
+    moe_unpermute,
     normalize_cute_dsl_moe_activation_type,
 )
 from .blockscaled_contiguous_gather_grouped_gemm_act_fusion import (
@@ -166,6 +167,7 @@ def _moe_core_impl(
     # Options
     output_dtype: torch.dtype = torch.bfloat16,
     use_async_memset: bool = True,
+    use_fused_finalize: bool = False,
     enable_pdl: bool = True,
     activation_type: int = ActivationType.Swiglu.value,
     swiglu_alpha: float = DEFAULT_SWIGLU_ALPHA,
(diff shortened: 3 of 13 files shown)


csrc/nv_internal/tensorrt_llm/kernels/cuteDslKernels/moeUtils.cu
```diff
@@ -181,12 +181,14 @@ __global__ void moeUnpermuteKernel(InputType const* permuted_input, InputType* o
...
+      int32_t const permuted_idx = expanded_idx_to_permuted_idx[expanded_idx];
       if (permuted_idx < 0) {
         continue;
       }
+      int32_t const input_idx = input_is_expanded ? expanded_idx : permuted_idx;
       auto const* src_ptr =
-          reinterpret_cast<ElemCopyType const*>(permuted_input) + permuted_idx * kCopyPerToken;
+          reinterpret_cast<ElemCopyType const*>(permuted_input) + input_idx * kCopyPerToken;
       *reinterpret_cast<ElemCopyType*>(rmem) = src_ptr[i];
       TopKScaleType const scale = topk_scales[token_idx * top_k + k];
```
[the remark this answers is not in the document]
zianglih: Good catch. `expanded_idx` already represents `token_idx * top_k + k`, so I changed the scale lookup to `topk_scales[expanded_idx]` in 29103f1b. The CuTe accuracy and CUDA-graph/determinism suites, TRTLLM determinism test, and both 4over6 determinism variants pass.

zianglih:
@coderabbitai review

flashinfer/fused_moe/cute_dsl/fused_moe.py
```diff
@@ -537,10 +565,10 @@ def _forward_with_tactic_weak(*args, **kwargs):
             top_k=top_k,
             num_local_experts=self.num_local_experts,
             local_expert_offset=local_expert_offset,
-            use_fused_finalize=True,
+            use_fused_finalize=False,
```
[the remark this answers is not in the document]
zianglih: Good catch. Addressed in 77fa36eb: `CuteDslMoEWrapper` now accepts and stores `use_fused_finalize`, and forwards it to both the standard and per-token runners. The benchmark wrapper path now accepts the same option, and a focused wrapper test verifies that both runner configurations receive `True`. Validation: the targeted B200 pytest passed (`1 passed`) and `pre-commit run --all-files` passed.

zianglih:
@coderabbitai review

flashinfer/fused_moe/cute_dsl/blackwell/utils.py
```diff
@@ -314,6 +314,25 @@ def atomic_add_func(rOut_epi_packed, scatter_out_offset, loc=None, ip=None):
...
+def blk_copy(dst_gemm, src_smem, size, loc=None, ip=None):
+    llvm.inline_asm(
+        None,
+        [
+            dst_gemm.iterator.toint(loc=loc, ip=ip).ir_value(loc=loc, ip=ip),
+            src_smem.iterator.toint(loc=loc, ip=ip).ir_value(loc=loc, ip=ip),
+            size.ir_value(loc=loc, ip=ip),
+        ],
+        "cp.async.bulk.global.shared::cta.bulk_group [$0], [$1], $2;",
+        "l,r,r",
```
aleozlx, line 327: is $1 not a 64b pointer?

tests/moe/test_cute_dsl_fused_moe.py
```diff
@@ -1249,6 +1249,7 @@ class TestCuteDslFusedMoeFunctional:
     @pytest.mark.parametrize(
         "hidden_size,intermediate_size", [(256, 512), (1024, 2048)]
     )
+    @pytest.mark.parametrize("use_fused_finalize", [False, True])
```
aleozlx: Should we separate it out of the Cartesian product and test a few combinations? (trying to regularize testing surface expansion)

zianglih: Agreed. Commit `e5944d2a` removes `use_fused_finalize` from the broad Cartesian product, preserving the original 144-case matrix for the default fused path. The deterministic finalize path now has two representative numerical cases that share the same assertion implementation: SwiGLU/per-tensor/top-k 1/small shape and ReLU2/per-token/top-k 8/large shape. Both targeted cases passed on B200 (`2 passed in 20.25s`), and `pre-commit run --all-files` passes.

flashinfer/fused_moe/cute_dsl/blackwell/utils.py
zianglih: A generic LLVM pointer is 64-bit, but `$1` is passed through `src_smem.iterator.toint()`, not as `llvm_ptr`. CuTe DSL's `Pointer.toint()` is address-space aware: it returns `Int64` for global/generic memory and `Int32` for shared memory. That makes `$0` a 64-bit global address and `$1` a 32-bit shared address, matching NVIDIA CCCL/CUTLASS's `cp.async.bulk.global.shared` constraints (`"l"` for global, `"r"` for shared). So `"l,r,r"` is intentional here; no code change is needed.

IwakuraRein (approved):

aleozlx (approved):

zianglih:
## B200 per-token deterministic-finalize benchmark

Reran the per-token NVFP4 MoE sweep with deterministic two-stage finalize (`--no-fused-finalize`) on an NVIDIA B200 using CUDA 13.2, PyTorch 2.13.0+cu132, CUDA Graph, and CUPTI timing. Tested source revision: `6dbe15e7`.

```bash
FLASHINFER_DISABLE_VERSION_CHECK=1 \
FLASHINFER_CUDA_ARCH_LIST=10.0a \
CUDA_VISIBLE_DEVICES=0 \
FLASHINFER_NVFP4_4OVER6=1 \
FLASHINFER_NVFP4_4OVER6_E4M3_USE_256=1 \
FLASHINFER_NVFP4_4OVER6_ERR_MODE=MSE \
FLASHINFER_NVFP4_4OVER6_ERR_USE_FAST_MATH=1 \
FLASHINFER_DISABLE_FP4_QUANT_FAST_MATH=1 \
python3 benchmarks/bench_moe_deepseek.py \
  --use-per-token-activation \
  --no-fused-finalize \
  --num-tokens 1,2,4,8,16,32,64,128,256,512,1024,2048,4096
```

Configuration: EP=1, TP=1, 256 experts, top-k=8, hidden size 7168, intermediate size 2048. Initial activation quantization is excluded. CUTLASS is omitted because it does not consume the per-token activation scale.

| Tokens | CuTe DSL (ms) | TRTLLM (ms) | TRTLLM / CuTe DSL latency | Faster backend |
|---:|---:|---:|---:|:---|
| 1 | 0.073 | 0.060 | 0.82x | TRTLLM |
| 2 | 0.113 | 0.095 | 0.83x | TRTLLM |
| 4 | 0.177 | 0.147 | 0.83x | TRTLLM |
| 8 | 0.290 | 0.239 | 0.82x | TRTLLM |
| 16 | 0.475 | 0.391 | 0.82x | TRTLLM |
| 32 | 0.735 | 0.603 | 0.82x | TRTLLM |
| 64 | 0.985 | 0.819 | 0.83x | TRTLLM |
| 128 | 1.079 | 0.900 | 0.83x | TRTLLM |
| 256 | 1.136 | 0.940 | 0.83x | TRTLLM |
| 512 | 1.174 | 0.963 | 0.82x | TRTLLM |
| 1024 | 1.184 | 1.071 | 0.90x | TRTLLM |
| 2048 | 1.330 | 1.159 | 0.87x | TRTLLM |
| 4096 | 1.672 | 1.709 | 1.02x | CuTe DSL |

The geomean TRTLLM / CuTe DSL latency ratio across all 13 shapes is **0.850x**. Equivalently, CuTe DSL is **1.176x slower** on geomean for this sweep. CuTe DSL becomes narrowly faster at 4096 tokens (`1.02x`).

One qualification: for 1-32 tokens, the TRTLLM autotuner reported that these shapes were outside its tuning bucket range and used the fallback tactic. The table reports the measured fallback performance for those rows.

PR #3984, opened 2026-07-15
perf: normalize autotuner nearest-profile cache keys

wongsingfo:
## 📌 Description

Motivation: We identified CPU contention that interfered with model forward execution, increasing TTFT by approximately ~4%. `py-spy` collected 142,500 samples from the forward path. Approximately 95% of the sampled cumulative GIL time was below:

```text
AutoTuner._get_cache_key
  -> AutoTuner._find_nearest_profile
```

Root cause: `AutoTuner._find_nearest_profile` currently keys its bounded LRU cache with the complete `(shapes, tuning_config)` pair. MLA creates per-call initializer closures. `DynamicTensorSpec.__hash__` excludes initializers while dataclass equality includes them, so equivalent MLA configs have the same hash but compare unequal:

```text
hash(config_a) == hash(config_b)
config_a != config_b
```

Those MLA lookups repeatedly probe the same hash bucket, walk its equality collision chain, miss, and insert another entry into the 16,384-entry LRU. Once full, the cache stays bounded but continues allocating, evicting, and performing the same comparisons. This is bounded cache churn and Python/GIL overhead.

This PR:

- keeps the existing `_find_nearest_profile(shapes, tuning_config)` interface, but backs it with an LRU keyed by a normalized `_NearestProfileKey`;
- includes only fields read by nearest-profile selection: dynamic input/dim indices, mapper identity, and constrained input/dim indices;

A/B cache-path measurements:

| Scenario | Base | This PR |
| --- | ---: | ---: |
| Rebuilt config with a fresh initializer (1,000 calls) | 449.940 µs/call | 5.082 µs/call |
| Actual MLA config builder + lookup (1,000 calls) | 477.659 µs/call | 8.159 µs/call |

## 🔍 Related Issues

- Follow-up to [[#3687](https://github.com/flashinfer-ai/flashinfer/pull/3687)](https://github.com/flashinfer-ai/flashinfer/pull/3687)
- Follow-up to [[#3912](https://github.com/flashinfer-ai/flashinfer/pull/3912)](https://github.com/flashinfer-ai/flashinfer/pull/3912)

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * Added `inputs_pre_hook` to the tuning configuration for custom preprocessing during autotuning.

* **Performance Improvements**
  * Improved nearest-profile caching to reuse results more reliably.
  * Cache-key stability now ignores initializer-closure details.

* **Bug Fixes**
  * Fixed how `tensor_initializers` are applied—now driven by `input_idx` to ensure correct per-input initialization.

* **API Changes**
  * Moved `tensor_initializers` ownership to `TuningConfig`.
  * Removed `FakeTensor` from the public API.

* **Tests**
  * Expanded coverage for cache-key behavior and initializer mapping, including deduplication and sparse/MLA scenarios.


jdebache:
@wongsingfo : thanks for reporting this.

Your measurements are compelling. I will try to reproduce.

jdebache:
I can repro. I think your fix is going the right direction, but we can push it a bit more. I will do a couple of experiments and come back to you.

jdebache:
@wongsingfo: here is my suggestion: https://github.com/flashinfer-ai/flashinfer/pull/4096.

If you pull the initializers out, then everything becomes simpler.

wongsingfo:
> @wongsingfo: here is my suggestion: #4096.
> 
> If you pull the initializers out, then everything becomes simpler.

@jdebache Thanks for the suggestion! Pulling the initializers out is indeed a much cleaner approach. I'll update the PR accordingly.

flashinfer/autotuner/autotuner.py
```diff
@@ -394,7 +394,7 @@ def get_config_path(is_module: bool):
         )
 
 
-@dataclass(slots=True)
+@dataclass(slots=True, unsafe_hash=True)
```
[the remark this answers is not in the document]
jdebache: `gen_tuning_buckets` is a tuple, not a list. Moreover, constraints are relevant (see the implementation of the nearest profile method).

The point here is to remove the asymmetry between equality and hash so that a hit by hash almost always implies a hit by equality, resulting in most hash buckets having 1 element and collapsing the lookup cost in the cache to O(1) in practice.

jdebache:
@wongsingfo: please re-run your benchmarks to verify that this version also addresses your original concerns, and that we haven't missed anything.

wongsingfo:
@jdebache  I reran the benchmark, and the results confirm that this PR addresses the original problem.

- `main@41155ec2` is the baseline from when I originally discovered the issue.
- `main@e4b8594e` includes the fix from #3970, which mitigates the issue.
- This PR fixes the underlying cache-key problem by decoupling nearest-profile lookup from per-call initializer identity. It also cleans up initializer ownership and applies the fix consistently across both MLA and MoE configurations.

| Scenario | `main@41155ec2` | `main@e4b8594e` (after fix in #3970) | This PR |
| --- | ---: | ---: | ---: |
| Rebuilt config with a fresh initializer | 1010.091 µs/call | 1017.070 µs/call | 2.971 µs/call |
| Actual MLA config builder + lookup | 1264.313 µs/call | 10.335 µs/call | 9.534 µs/call |
| Actual MoE config builder + lookup | 1247.152 µs/call | 15.318 µs/call | 14.240 µs/call |

Note that these measurements were collected on a different machine (due to some resource constraints...), so the absolute numbers are not directly comparable with the results in the original PR description.

jdebache:
Great! Let's now wait on CI to complete.

wongsingfo:
The latest CI run appears to have failed because the self-hosted runners were shut down, rather than because of a test failure. Could someone with Actions permissions please rerun the failed jobs? Thanks!

jdebache:
CI looks good, I'm looking to get approval from code owner. It might take a couple more days due to limited bandwidth.

aleozlx (approved):

jdebache:
Hey @wongsingfo, we are trying to get this merged, but noticed that there is now a pre-commit failure. Would you mind addressing it? Apologies again for the delay on this.

flashinfer/mla/_core.py
```diff
@@ -2411,6 +2411,7 @@ def init_causal_seqlens_kv_global(shapes, dtype, device):
                 tensor_initializers=tensor_initializers,
             ),
         ),
+        tensor_initializers=((1, init_block_tables), (2, init_seq_lens)),
```
qiching: Two lines above, DynamicTensorSpec still passes `tensor_initializers=tensor_initializers`, but that field no longer exists so i think this raises TypeError on every MLA decode autotune call, and public CI misses it because no shard runs tests/autotuner/. 

Also, with enable_dcp=True index 4 is in input_idx but isn't registered here, please add (4, init_causal_seqlens_kv_global) and cover it in the new index assertion.

qiching (changes requested):

flashinfer/mla/_core.py
wongsingfo: Thanks for catching the issues. I’ve fixed both in the latest commit and added coverage test.

wongsingfo:
> Hey @wongsingfo, we are trying to get this merged, but noticed that there is now a pre-commit failure. Would you mind addressing it? Apologies again for the delay on this.

Hi @jdebache, I've already fixed it in the latest commit.

qiching (approved):
LGTM.

PR #3395, opened 2026-05-23
feat(attention): add SM120 sparse MLA kernels

lucifer1004:
## Summary

Sparse-MLA paged attention for SM120a / SM121a, covering both the DSv4 family
(`d_qk=512`) and the DSv3.2 / GLM-5.1 family (`d_qk=576`). Adds decode
(warp-spec, split-K) and prefill (TMA + warp-spec) kernels in the SM120 sparse
MLA JIT/AOT module, with dispatch through the existing MLA Python APIs.

## Surfaces

The user-facing surface is limited to extending existing MLA entry points with
SM120/SM121 sparse backend coverage. SM120-specific dispatchers, runners, and
direct kernel wrappers remain internal implementation details; users should
continue to call the public `flashinfer.mla` APIs below. The public APIs keep
their existing decode-oriented names, while the SM120 sparse backend internally
selects decode or prefill kernels from the flattened token count.

### Public Python APIs

- `flashinfer.mla.trtllm_batch_decode_sparse_mla_dsv4(...)`
  - Extends the DSv4 sparse-MLA decode entry point to SM120/SM121.
  - SM100/SM103 continue to route through the TRTLLM-GEN path.
  - SM120/SM121 route through the new packed sparse backend. The SM120 path
    accepts BF16 query tensors, a packed uint8 SWA KV cache with 584 bytes per
    token, optional attention sinks, and an optional second packed KV cache via
    `extra_sparse_indices` / `extra_sparse_topk_lens`.
  - `swa_topk_lens` describes the active SWA segment. When the optional
    compressed segment is present, `extra_sparse_topk_lens` describes its active
    length. Both sparse segments share one online-softmax denominator.

- `flashinfer.mla.trtllm_batch_decode_with_kv_cache_mla(...)`
  - Extends the existing MLA decode API with an SM120/SM121 sparse backend for
    DSv3.2 / GLM-style packed sparse MLA (`d_qk=576`, 656 bytes per token).
  - With `backend="auto"` on SM120/SM121 and `sparse_mla_top_k > 0`, the call
    selects the sparse backend. `backend="sparse"` selects it explicitly.
  - `kv_scale_format` controls the packed-cache scale convention:
    `"auto"` / `"pow2_fp32"` for DSv3.2-style power-of-two FP32 scales, and
    `"arbitrary_fp32"` for GLM-style arbitrary FP32 scales.

### Internal / JIT Implementation Details

[...]

PR #3989, opened 2026-07-15
fix(DeepSeek-v4-flash-dspark sm120 error)

### LINKED ISSUE #3988: Sm120 decode dsv4 topk256 bad routing and shape not accepted
Following vllm-project/vllm#46995 with TP-4 and DeepSeekV4-Flash instead of Pro, to test specdec on my local machine, with this command

```bash
CUDA_VISIBLE_DEVICES=4,5,6,7 vllm serve deepseek-ai/DeepSeek-V4-Flash-DSpark   --tensor-parallel-size 4 --trust-remote-code --kv-cache-dtype fp8   --speculative_config '{"method": "dspark", "num_speculative_tokens": 7, "draft_sample_method": "probabilistic"}' --gpu-memory-utilization 0.7   --port 8006
```

Resulted in this error, which should be possible to replicate with vllm 0.25.0 and flashinfer-python 0.6.14:
```
(Worker_TP1 pid=2362331) ERROR 07-15 15:25:16 [multiproc_executor.py:1004]   File "/home/atatjer/src/personalized-continual-speculator/.venv/lib/python3.12/site-packages/flashinfer/data/csrc/sparse_mla_sm120.cu", line 261, in void flashinfer::sparse_mla_sm120::SparseMlaSm120PagedAttention(tvm::ffi::TensorView, tvm::ffi::TensorView, tvm::ffi::TensorView, tvm::ffi::TensorView, tvm::ffi::TensorView, double, int64_t, tvm::ffi::Optional<tvm::ffi::TensorView>, tvm::ffi::Optional<tvm::ffi::TensorView>, tvm::ffi::Optional<tvm::ffi::TensorView>, tvm::ffi::Optional<tvm::ffi::TensorView>, tvm::ffi::Optional<tvm::ffi::TensorView>)
(Worker_TP1 pid=2362331) ERROR 07-15 15:25:16 [multiproc_executor.py:1004]     TVM_FFI_ICHECK_GT(num_tokens, 64)
(Worker_TP1 pid=2362331) ERROR 07-15 15:25:16 [multiproc_executor.py:1004]     
(Worker_TP1 pid=2362331) ERROR 07-15 15:25:16 [multiproc_executor.py:1004] tvm.error.InternalError: Check failed: num_tokens > 64 (7 vs. 64) : Decode (num_tokens <= 64) must go through sparse_mla_sm120_decode_dsv3_2 or sparse_mla_sm120_decode_dsv4; got num_tokens=7
```

NVIDIA RTX PRO 6000 Blackwell Server Edition, 595.71.05
2.11.0+cu130 13.0
Python 3.12.13
6.8.0-124-generic
1c005b4d
vllm==0.25.0

aldakata:
## 📌 Description

**Problem:** DeepSeek-V4-Flash-DSpark's draft module issues sparse MLA decode calls
with (num_heads=16 at TP4, topk=256). This shape is absent from both the decode
dispatch table ({128,512,1024} × nh) and the prefill dsv4_single ladder, so draft
calls fall through to the prefill orchestrator, whose num_tokens>64 guard rejects
them — making DSpark spec-decode unusable on SM120. Repro (vLLM 0.25.0):

[your command block]
[your error block]

**Fix:** adds topk=256 instantiations:
- decode: (nh, 256) for nh ∈ {8,16,32,64,128} in `_DECODE_DSV4_DISPATCH` + the
  `DSV4_DISPATCH` ladder (kernel is TOPK-generic: BI=64-chunked smem,
  dynamic-split merge; 256 = 4 splits, between validated 2 and 8)
- prefill: topk==256 in `dispatch_dsv4_single`, **ComputeMode::BF16** — measured
  6–10% faster than FP8 at K=256 (64/128 heads × 128–1024 tokens, RTX PRO 6000
  Blackwell), placing the BF16/FP8 crossover above 256
- also fixes `bench_sparse_mla_sm120.py`: `quantize_kv_model1` returned an NHD
  view rejected by `_check_sm120_dsv4_kv_cache_layout`; broken at the first stock
  config since its introduction in #3395 (metadata-only fix)

**Verification:**
- pytest `test_sparse_mla_sm120.py` incl. new (nh,256) decode+prefill configs,
  all green vs the SDPA reference
- greedy outputs token-identical spec-on vs spec-off (V4-Flash-DSpark, TP4)
- bench (256 rows interpolate 128/512 neighbors cleanly):

### DSv4 decode path (num_tokens ≤ 64) — new topk=256 rows (bold) with 128/512 neighbors

| num_heads | topk | num_tokens | lat (µs) | kv BW (GB/s) | attn TFLOPs |
|----------:|-----:|-----------:|---------:|-------------:|------------:|
| 16 | 128 | 16 | 23.7 | 50.4 | 2.83 |
| **16** | **256** | **16** | **25.2** | **94.7** | **5.32** |
| 16 | 512 | 16 | 27.8 | 172.2 | 9.66 |
| 32 | 128 | 16 | 25.2 | 47.4 | 5.32 |
| **32** | **256** | **16** | **27.3** | **87.6** | **9.83** |
| 32 | 512 | 16 | 37.5 | 127.5 | 14.30 |
| 128 | 128 | 16 | 35.5 | 33.7 | 15.13 |
| **128** | **256** | **16** | **56.3** | **42.5** | **19.07** |
| 128 | 512 | 16 | 76.8 | 62.3 | 27.96 |
| 128 | 128 | 32 | 60.1 | 39.8 | 17.88 |
| **128** | **256** | **32** | **81.0** | **59.1** | **26.53** |
| 128 | 512 | 32 | 116.7 | 82.0 | 36.79 |

### DSv4 prefill path (num_tokens > 64) — topk=256 uses ComputeMode::BF16

| num_heads | topk | num_tokens | lat (µs) | kv BW (GB/s) | attn TFLOPs |
|----------:|-----:|-----------:|---------:|-------------:|------------:|
| 64 | 128 | 128 | 73.7 | 129.8 | 29.13 |
| **64** | **256** | **128** | **96.3** | **198.8** | **44.62** |
| 64 | 512 | 128 | 145.4 | 263.2 | 59.07 |
| 128 | 128 | 128 | 102.4 | 93.4 | 41.94 |
| **128** | **256** | **128** | **133.1** | **143.8** | **64.54** |
| 128 | 512 | 128 | 204.8 | 186.9 | 83.89 |
| 64 | 128 | 512 | 190.5 | 200.9 | 45.10 |
| **64** | **256** | **512** | **252.4** | **303.3** | **68.08** |
| 64 | 512 | 512 | 400.9 | 381.9 | 85.71 |
| 128 | 128 | 1024 | 617.5 | 124.0 | 55.65 |
| **128** | **256** | **1024** | **840.2** | **182.2** | **81.79** |
| 128 | 512 | 1024 | 1384.5 | 221.1 | 99.27 |

### ComputeMode comparison at topk=256 (prefill, idle GPU, otherwise identical builds)

| (num_heads, num_tokens) | BF16 (µs) | FP8 (µs) | BF16 advantage |
|:------------------------|----------:|---------:|---------------:|
| (64, 128)   | 96.3  | 102.4 | 6.0%  |
| (128, 128)  | 133.1 | 143.4 | 7.2%  |
| (64, 512)   | 252.4 | 274.4 | 8.0%  |
| (128, 1024) | 840.2 | 938.0 | 10.4% |

**Environment:** 4× RTX PRO 6000 Blackwell (sm_120a), driver 595.71.05, CUDA 13.2,
torch 2.11.0, vLLM 0.25.0, kernels JIT-compiled from this branch.

**Side observations (not addressed here):** both dispatch-failure asserts report
misleading preconditions (size-based wording when the failure is table
membership — suggest printing the failing (nh, topk) tuple); prefill's NH ladder
lacks case 8 (TP8 would fail even at supported topks).

## 🔍 Related Issues
Fixes #3988
Related: vllm-project/vllm#46995 (DSpark support — this provides the SM120
instantiations for the Flash draft's topk=256; was #46995 validated on SM100/Pro?)

## Summary by CodeRabbit

- **New Features**
  - Expanded sparse MLA DSV4 support to handle additional head-count and top-k configurations, including top-k 256.
  - Added broader coverage across decode and prefill workloads.

- **Tests**
  - Extended automated coverage to validate the newly supported DSV4 configurations.

- **Benchmarks**
  - Expanded benchmark scenarios to measure performance across the additional head-count, token, and top-k combinations.


aldakata:
@bkryu is there anything I can do to help speed this up? Happy to help here or with other PRs.
Thanks

croberts012:
**Repro data point — patch fixes dispatch but kernel launch fails on SM121 (GB10 / DGX Spark)**

Thanks for this fix — adding a hardware data point from testing it on **GB10 (Grace-Blackwell, SM121 / compute capability 12.1)**, since the original report was from RTX PRO 6000 (SM120).

**Setup:**
- 4× DGX Spark (GB10, 128 GB unified), driver 580.159.03, CUDA 13
- vLLM v0.25.0 (source build @ `702f4814`, `TORCH_CUDA_ARCH_LIST=12.1a`)
- FlashInfer `v0.6.15` (@ `8eccd0c`) + this PR's diff (head `1459c5d`), built from source, JIT kernel path (no `flashinfer-cubin` package)
- Model: `deepseek-ai/DeepSeek-V4-Flash-0731`, DSpark `{"method":"dspark","num_speculative_tokens":5}`

**Result:** the dispatch fix works — the original `num_tokens > 64 (5 vs. 64)` prefill-fallthrough rejection is gone. But the selected kernel then fails to launch:

```
RuntimeError: Check failed: (ok) is false: decode-dsv4 launch failed (unsupported shape or kernel error)
  at flashinfer::sparse_mla_sm120::SparseMlaSm120DecodeDsv4
```

**Isolated repro with this PR's own test** (clean container, fresh JIT cache, single GPU — no vLLM involved):

```python
test_sparse_mla_sm120_decode_dsv4(num_heads=16, topk=256, num_tokens=1, with_sink=False)
# -> RuntimeError: Check failed: (ok) is false: decode-dsv4 launch failed
```

Fails at the smallest parametrization, so it looks like the topk=256 instantiation's launch configuration (shared memory / occupancy?) exceeds what an SM121 SM provides, rather than anything shape-routing related. Happy to run diagnostics or test revisions on this hardware if useful — DGX Spark clusters are a common target for this model family.

croberts012:
**Correction + stronger result: PR works on SM121 — our earlier failure was a stale AOT artifact**

Retracting my launch-failure report above after deeper investigation: no kernel was ever refused. Our container carried the `flashinfer-jit-cache` wheel's prebuilt `sparse_mla_sm120.so` (built pre-PR), and `JitSpec.build_and_load()` prefers the AOT artifact over compiling patched source — so the dispatch chain fell through (`TVM_FFI_ICHECK(ok)`) having found no topk=256 instantiation, and the error message made it look like a launch rejection.

With that artifact removed so the patched source JIT-compiles: **GB10/SM121 passes the full sweep — 79/79** (all `_DSV4_DECODE_CONFIGS` × num_tokens {1,16,64} × sink {±}, incl. every topk=256 case, plus prefill and truncation tests; no regression on 128/512/1024). Measured resource data: dynamic smem for topk=256 is byte-identical to topk=128 (90,800 B vs the 101,376 B optin limit; TOPK only affects the chunk-loop trip count) — there's no resource concern on either SM120 or SM121.

We're now running this PR in production: DeepSeek-V4-Flash-0731 with DSpark (k=5, greedy) on a 4× DGX Spark TP4 cluster — ~+30% single-stream decode, valid structured output under speculation. Draft acceptance is ~30%/token at TP4 (vs ~60% reported at TP2 with the loader fix), consistent with the known acceptance degradation at higher node counts.

Two small suggestions from the debugging trail: (1) splitting the `ICHECK` into "no dispatch case for (num_heads, topk)" vs "CUDA launch rejected" would have made this a 5-minute diagnosis; (2) a stale AOT cache silently desynchronizes the Python and C++ dispatch tables — a version/hash check between them would catch it. Happy to share the SM121 validation logs. This PR looks correct and complete as-is — hope it can move forward.

jasl:
Validated this on SM121a (2× NVIDIA GB10 / DGX Spark), which I think is the
hardware gap in the current review — the kernels declare `[120, 121]` but the
reports on vllm-project/vllm#50720 span both, so it is worth having a data point
from the 121 side. **It fixes the reported failure, and the reference tests pass.**

### Before / after, at the exact shape from vllm#50720

Called `_sparse_mla_sm120_paged_attention` directly at DSv4 shapes
(`d_qk=512`, 584 B/token, `page_block_size=64`), `num_tokens=7` — that 7 is
`num_speculative_tokens` from the DSpark draft in the linked issue:

| num_heads | topk | stock 0.6.16 | with #3989 |
|---|---|---|---|
| 16 | 128 | OK | OK |
| 16 | **256** | `Check failed: num_tokens > 64 (7 vs. 64)` | OK |
| 32 | 128 | OK | OK |
| 32 | **256** | `Check failed: num_tokens > 64 (7 vs. 64)` | OK |
| 128 | 128 | OK | OK |
| 128 | **256** | `Check failed: num_tokens > 64 (7 vs. 64)` | OK |

`_DECODE_DSV4_DISPATCH` goes 15 → 20 entries, `topk ∈ {128, 256, 512, 1024}`.
The stock error is the one three people reported on that issue: `_paged_attention`
consults `_decode_dsv4_dispatchable`, but a **miss is unhandled** — it falls
straight through to `module.sparse_mla_sm120_paged_attention`, the prefill
orchestrator, which then rejects the call for being decode-sized.

Which suggests a second, independent fix worth making: two lines before that
final call — if `num_tokens <= _DECODE_MAX_TOKENS`, raise something naming the
actual `(num_heads, topk)` and the instantiated `topk` values. As it stands, any
future uninstantiated shape produces an assertion about `num_tokens`, which is
not the thing that is wrong; that is why the downstream issue took three reports
and an instrumented run to localise. Happy to open that as a separate PR.

### Reference tests

`tests/attention/test_sparse_mla_sm120.py` on the patched build:

```
-k decode_dsv4    75 passed
-k prefill_dsv4   44 passed
```

### One thing I would add to the PR

The decode `.cu` instantiates five new shapes:

```
DSV4_DISPATCH(8, 256)  (16, 256)  (32, 256)  (64, 256)  (128, 256)
```

but `_DSV4_DECODE_CONFIGS` only gains `(8, 256)` and `(16, 256)`, so
**`(32, 256)`, `(64, 256)` and `(128, 256)` would ship untested.** I added those
three to the list and ran the full parametrization — `num_tokens ∈ {1, 16, 64}`
× `with_sink ∈ {off, on}`, so 18 additional cases — and all pass. This is a
coverage gap rather than a defect, but worth folding into the PR.

(The `75 passed` above is that widened grid: 12 configs × 3 × 2, plus the three
unparametrized decode tests.)

### BF16 vs FP8 at topk=256 — the choice is right, by ~5%

`dispatch_dsv4_single` sends 256 through BF16 while 512/1024 use FP8, on the
reasoning that FP8 only amortises its Q-quantize prologue at larger K. Since
`DISPATCH_BY_NH_CM` is generic over the compute mode, that is one token to
change, so I built it both ways on an otherwise idle SM121a. **`topk=128` and
`512` are identical in the two builds and act as a control.**

| num_heads | topk | num_tokens | BF16 build | FP8-at-256 build | |
|---|---|---|---|---|---|
| 64 | 128 | 128 | 228.4 us | 227.4 us | control, 0.4% |
| 64 | **256** | 128 | **293.0 us** | 307.2 us | **BF16 4.8% faster** |
| 64 | 512 | 128 | 440.4 us | 442.4 us | control, 0.5% |
| 128 | 128 | 128 | 369.7 us | 368.7 us | control, 0.3% |
| 128 | **256** | 128 | **470.9 us** | 498.8 us | **BF16 5.9% faster** |
| 128 | 512 | 128 | 713.7 us | 716.7 us | control, 0.4% |
| 128 | 128 | 1024 | 2679.8 us | 2639.0 us | control, 1.5% |
| 128 | **256** | 1024 | **3400.6 us** | 3588.1 us | **BF16 5.5% faster** |
| 128 | 512 | 1024 | 5297.7 us | 5305.4 us | control, 0.1% |

The controls agree within 1.5%, so the 4.8–5.9% gap is outside run-to-run
spread. FP8 at 256 is *correct* — all 44 prefill tests pass either way — just
slower. The existing comment's reasoning holds at 256.

Two measurement notes, in case anyone reproduces this:

* These are from a **second** pass over the grid; the first is discarded.
  `benchmarks/bench_sparse_mla_sm120.py` as written reports its early configs
  badly warm-up-contaminated — it gave me `(16, 128, 16)` at 387 us, slower than
  `(16, 512, 16)` at 75.7 us, and every prefill number came out ~2x its
  steady-state value. A warmup pass inside the bench would be worth having.
* The bench does cover `(64, 256)` and `(128, 256)` in prefill even though the
  test grid does not, since prefill dispatches on `topk` alone.

### Drive-by: the benchmark cannot run on SM121

`benchmarks/bench_sparse_mla_sm120.py` guards on `is_sm120a_supported`, so it
exits with `Sparse-MLA SM120 requires sm120a.` on SM121a — even though
`_SparseMLAPagedAttentionRunner` is decorated `@supported_compute_capability([120, 121])`
and the test file uses the wider `is_sm12x_supported`. I had to widen the guard
locally to produce the numbers above. Since this PR already touches that file,
it might be a good place to align it.

### Applying it without an AOT rebuild

For anyone hitting this before the fix lands in a release: these kernels are
JIT-built from sources shipped inside the wheel
(`gen_sparse_mla_sm120_module().build_and_load()` reads
`flashinfer/data/csrc/`), so patching the installed package is enough — no
FlashInfer build:

```bash
PKG=$(python -c "import flashinfer,os;print(os.path.dirname(flashinfer.__file__))")
git clone https://github.com/flashinfer-ai/flashinfer.git && cd flashinfer
git fetch origin pull/3989/head:pr3989 && git checkout pr3989
cp csrc/sparse_mla_sm120_decode_dsv4.cu csrc/sparse_mla_sm120_prefill.cu "$PKG/data/csrc/"
cp flashinfer/mla/_sparse_mla_sm120.py "$PKG/mla/"
rm -rf ~/.cache/flashinfer/*sparse_mla_sm120*
```

The JIT rebuilds on the next call. That is what the table above was measured on.

One caveat if you reproduce this: don't run the tests from inside the clone.
`tests/__init__.py` and `tests/attention/__init__.py` both exist, so pytest puts
the repo root on `sys.path` and `import flashinfer` resolves to the source tree,
which has no `data/csrc/` — every JIT build then fails on a missing `.cu` and it
looks like a kernel problem. Copy the test file out first.

saltyminty:
This PR is superseded by and will be closed after https://github.com/flashinfer-ai/flashinfer/pull/4380

bkryu:
Closing as superseded by #4380

Issue #3999, opened 2026-07-16
[Bug]`test_bmm_fp8.py` regression: cuDNN tuple tactic can be replayed on cuBLAS runner after #3707


cindyzxq:
## Summary

Found unit-test error in https://nv/flashinfer-ci/-/jobs/365155065/raw `tests/gemm/test_bmm_fp8.py` started failing with:
```
    TypeError: '>=' not supported between instances of 'tuple' and 'int'
    /workspace/flashinfer/flashinfer/gemm/gemm_base.py:214
```
This appears to be a regression introduced by #3707 (b35396c1 Yanqinz/autotuner tactic).#3707 changed cuDNN GEMM tactics from integer plan indices to structured (engine_id, knob_items) tuples, but the autotuner cache still stores runner_id as an index relative to the runner list used during the original tuning call. When the same cached cuDNN entry is reused from a later call whose runner list has a different ordering, the tuple tactic can be dispatched to a non-cuDNN runner.

## Failure
Observed in:  tests/gemm/test_bmm_fp8.py
Failure snippet:
```
    E   TypeError: '>=' not supported between instances of 'tuple' and 'int'
    E   /workspace/flashinfer/flashinfer/gemm/gemm_base.py:214
```
    The failing line is in the cuBLAS FP8 BMM runner: 
    if tactic >= 0:
    This runner expects an integer tactic, but receives a tuple.

## Suspected Cause (By Codex)
After #3707, CudnnFp8GemmRunner.get_valid_tactics() returns tuple tactics via:
```
    return _cudnn_graph_engine_knob_tactics(graph)
```
The autotuner stores cache entries as:
```
    self.profiling_cache[cache_key] = (runner_id, tactic, p)
```
where runner_id is only meaningful for the runner list used during that specific choose_one() call.

A likely sequence is:

    1. backend="cudnn" tunes bmm_fp8; runner list is [CudnnFp8GemmRunner], so cached value stores runner_id=0 and
       tactic=(engine_id, knob_items).
    2. Later backend="auto" runs for the same shape/profile. For FP8 e5m2 cases, CUTLASS is excluded, so runner list can be
       [CublasFp8GemmRunner, CudnnFp8GemmRunner].
    3. search_cache() finds the cuDNN cache key, but returns stale runner_id=0.
    4. Current runners[0] is cuBLAS, so cuBLAS receives the cuDNN tuple tactic and fails at tactic >= 0.

[...]

PR #4004, opened 2026-07-16
Yanqinz/fix autotuner using autotuning plan from other runners

yanqinz2:
## 📌 Description

This PR removes the explicit runner position index used in autotuner to prevent it using autotuning plan from other runners, which can cause either silent perf loss, or even crash

## 🔍 Related Issues

https://github.com/flashinfer-ai/flashinfer/issues/3999

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved autotuning cache lookup reliability across runner configurations.
  * Preserved distinct tuning results for configurations with different extra parameters.
  * Added support for non-integer tactic values in tuning configurations.
  * Ensured cached tactics are matched to the correct runner rather than relying on runner order.
  * Improved compatibility when loading and saving existing tuning configuration files.

### DIFF
--- a/flashinfer/autotuner/autotuner.py
+++ b/flashinfer/autotuner/autotuner.py
@@ -1229,34 +1229,33 @@ class AutoTuner:
             synthesis-invariant; see: TunableRunner.get_cache_key_extras.
         """
         with self._lock:
-            for r_id, r in enumerate(runners):
-                extras = r.get_cache_key_extras(inputs) if inputs is not None else ()
-                cache_key = AutoTuner._get_cache_key(
-                    custom_op, r, input_shapes, tuning_config, extras
+            runner_keys = [
+                (
+                    r_id,
+                    AutoTuner._get_cache_key(
+                        custom_op,
+                        r,
+                        input_shapes,
+                        tuning_config,
+                        r.get_cache_key_extras(inputs) if inputs is not None else (),
+                    ),
                 )
-                # 1. In-memory cache (from live tuning)
+                for r_id, r in enumerate(runners)
+            ]
+
+            # 1. In-memory cache (from live tuning)
+            for r_id, cache_key in runner_keys:
                 if cache_key in self.profiling_cache:
                     tactic, stored_profile = self.profiling_cache[cache_key]
                     return True, r_id, tactic, stored_profile
 
-                # Build the hash-free file key used by both user configs and bundled configs.
-                # Include extras (index 4) so that runner specific parameters
-                # are not lost on disk.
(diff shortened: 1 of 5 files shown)


tests/autotuner/test_autotuner_core.py
```diff
@@ -231,11 +231,39 @@ def test_search_cache_hit_and_miss():
     assert miss == (False, 0, -1, None)
 
     key = AutoTuner._get_cache_key("dummy", runner, shapes, config)
-    tuner.profiling_cache[key] = (0, 1, None)
+    tuner.profiling_cache[key] = (1, None)
     hit = tuner.search_cache("dummy", [runner], shapes, config)
     assert hit == (True, 0, 1, None)
 
 
+def test_search_cache_hit_resolves_runner_id_against_current_list():
```
qiching, line 239: This regression test only covers in-memory runner-id resolution. The second bug this PR claims to fix that a `_file_configs` hit on an earlier listed runner shadowing a fresher in-memory result on a later-listed runner, has no direct test. Could we add one test that runners [A, B], seed `_file_configs` for A and `profiling_cache` for B, then assert search_cache returns B?

qiching (changes requested):

flashinfer/autotuner/autotuner.py
```diff
@@ -1193,50 +1196,51 @@ def search_cache(
             synthesis-invariant; see: TunableRunner.get_cache_key_extras.
         """
         with self._lock:
-            for r in runners:
-                extras = r.get_cache_key_extras(inputs) if inputs is not None else ()
-                cache_key = AutoTuner._get_cache_key(
-                    custom_op, r, input_shapes, tuning_config, extras
+            runner_keys = [
```
qiching: We now eager build cache keys for all runners before any lookup, but the old code short circuited on an in-memory hit for runner 0. Since `search_cache` runs on every inference forward, can you confirm `get_cache_key_extras` is cheap for the real runners like cuDNN or cuBLAS? For short runner lists this is fine, just want to rule out a per forward regression.

yanqinz2: Now only build cache keys in the in-memory iteration.

tests/autotuner/test_autotuner_core.py
yanqinz2: Added test: test_search_cache_in_memory_beats_file_config_across_runners

qiching (approved):
LGTM!

PR #4007, opened 2026-07-16
feat: add collect-env environment report tool + issue templates (bug, feature request)

YangXu1990uiuc:
## 📌 Description

Issue reporters often can't state their environment precisely, and the most common unreproducible-issue root cause is version confusion — multiple copies of cuDNN/cuBLAS/CUDA runtime installed where the **loaded** one is not the one the user assumes.

This PR adds an environment-forensics tool and wires it into a new (first) GitHub issue template:

```bash
python -m flashinfer.collect_env        # or: flashinfer collect-env [--json]
```

**What the report covers**

- flashinfer / flashinfer-cubin / flashinfer-jit-cache versions (with explicit mismatch flag), JIT cache + local cubin store stats, resolved target CUDA archs
- GPU/driver properties in **CUDA enumeration order** (nvidia-smi order as no-torch fallback), CUDA toolkit **as resolved by flashinfer's JIT** vs `CUDA_HOME`
- **Loaded vs installed GPU libraries**: force-loads the libs torch actually uses, parses `/proc/self/maps` for what is truly mapped, scans `LD_LIBRARY_PATH` / site-packages / `CUDA_HOME` / ldconfig for other on-disk installs; aggregated per directory with pip provenance, ⚠ on real conflicts
- torch build info, cuDNN frontend + loaded backend version, relevant packages (incl. vLLM/SGLang and their declared flashinfer pins, stated factually), `FLASHINFER_*`/`CUDA*`/… env vars, GPU topology

**Design constraints** (documented in the module docstring)

- stdlib-only at module level and every probe individually guarded → still produces a report when `import flashinfer` or torch is broken; the file can be `curl`-ed and run standalone with bare Python
- strictly offline/read-only (no downloads, unlike `show-config`'s artifact status), ~7 s wall clock

Example of the headline section on a box with a pip-wheel CUDA stack plus a local toolkit:

```text
==== GPU Libraries: loaded vs on disk ====
libcublas  ⚠ other installs on disk (check which one you expect):
    LOADED   .../.venv/lib/python3.12/site-packages/nvidia/cu13/lib/  [pip nvidia-cublas==13.0.0.19]
    on disk  /usr/local/cuda-13.2/targets/x86_64-linux/lib/  [v13.3.0.5]
```

## 🔍 Related Issues

Improves triage for env-dependent reports in general (e.g. the recurring "works on my machine" class).

### ✅ Pre-commit Checks

## 🧪 Tests

`tests/utils/test_collect_env.py` asserts the core contract (collection never raises, sections present, JSON-serializable) and runs GPU-less. Manually verified on a multi-arch box (A100/L40S/H100/B200-class, driver 595.71.05): full run in a venv, standalone run with bare `/usr/bin/python3` (no torch/flashinfer — degrades gracefully), `--json` parses.

## Reviewer Notes

- `/proc/self/maps` is Linux-only; on other platforms that section degrades to the on-disk scan.
- AI-assisted (Claude Code).

## Summary by CodeRabbit

* **New Features**
  * Added a `collect-env` CLI command to generate detailed environment reports (runtime, CUDA/GPU details, loaded libraries, and version info).
  * Supports formatted text or pretty-printed JSON output.
  * Added GitHub issue forms for bug reports and feature requests with prompts for reproducible steps and environment output.

* **Documentation**
  * Updated quick-reference materials with environment-report commands, including an optional `--json` mode.

* **Tests**
  * Added tests covering report section presence/structure, FlashInfer↔flashinfer version consistency, text formatting, and JSON-serializability.

---

**Update (2026-07-17):** also adds `🚀 Kernel / Feature Request` (`feature_request.yml`), a slimmed revision of the template draft circulated on Slack, incorporating the thread feedback (thanks @Kaustubh, Brian, Jingfan):

- Two clean templates with an explicit routing rule in each description: *broken / wrong / slower than before* → Bug Report; *new capability / speedup* → Feature Request
- Required fields: 4 (+ a "searched existing issues" checkbox on both templates); the remaining fields are grouped under a visible "everything below is optional" divider
- Hardware dropdown gains Ampere/Ada; inference engine is multi-select; Evidence dropdown folded into the Problem field; Impact label/multi-select contradiction fixed
- Both templates reference the same environment tool: `python -m flashinfer.collect_env`; bug-template reproducer guidance now asks for a short standalone script rather than a full `vllm serve` run

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,7 +1,17 @@
 name: 🐛 Bug Report
-description: Report a correctness, crash, or performance bug in FlashInfer
+description: >
+  Report something broken, wrong, or slower than it used to be — correctness
+  issues, crashes, build/install failures, or performance regressions. For new
+  capabilities or speedups, use 🚀 Kernel / Feature Request instead.
 labels: ["bug"]
 body:
+  - type: checkboxes
+    id: presubmit
+    attributes:
+      label: Before submitting
+      options:
+        - label: I have searched existing issues and did not find a solution or an existing report.
+          required: true
   - type: textarea
     id: environment
     attributes:
@@ -38,10 +48,12 @@ body:
     attributes:
       label: Minimal reproduction
       description: >
-        Smallest self-contained script or command that reproduces the issue.
-        Include exact shapes/dtypes and the backend selected (fa2/fa3/cudnn/
-        trtllm/cutlass/...). If the failure involves JIT compilation, re-run
-        with `FLASHINFER_JIT_VERBOSE=1` and attach the log.
+        Smallest self-contained script that reproduces the issue — a short
+        standalone Python script is far more actionable than a full
+        `vllm serve ...` invocation. Include exact shapes/dtypes and the
+        backend selected (fa2/fa3/cudnn/trtllm/cutlass/...). If the failure
+        involves JIT compilation, re-run with `FLASHINFER_JIT_VERBOSE=1` and
+        attach the log.
       render: python
     validations:
       required: true
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -25,15 +25,8 @@ body:
         The report includes which cuDNN/cuBLAS/CUDA libraries are *actually
         loaded* vs merely installed — this resolves most "works on my machine"
         divergence, so please don't trim it.
-      value: |
-        <details>
-        <summary>Output of python -m flashinfer.collect_env</summary>
-
-        ```text
-        PASTE OUTPUT HERE
-        ```
-
-        </details>
+      render: text
+      placeholder: Paste the full output of `python -m flashinfer.collect_env` here.
     validations:
       required: true
   - type: textarea
@@ -51,14 +44,23 @@ body:
         Smallest self-contained script that reproduces the issue — a short
         standalone Python script is far more actionable than a full
         `vllm serve ...` invocation. Include exact shapes/dtypes and the
-        backend selected (fa2/fa3/cudnn/trtllm/cutlass/...). If the failure
-        involves JIT compilation, re-run with `FLASHINFER_JIT_VERBOSE=1` and
-        attach the log.
+        backend selected (fa2/fa3/cudnn/trtllm/cutlass/...).
       render: python
     validations:
       required: true
   - type: textarea
+    id: logs
+    attributes:
+      label: Error messages / logs
+      description: >
+        Stack trace, error output, or logs. If the failure involves JIT
+        compilation, re-run with `FLASHINFER_JIT_VERBOSE=1` and include that
+        log.
+      render: text
+    validations:
+      required: false
+  - type: textarea
     id: extra
     attributes:
       label: Additional context
-      description: Stack traces, logs, related issues, workarounds you tried.
+      description: Related issues, workarounds you tried, anything else useful.
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,52 @@
+name: 🐛 Bug Report
+description: Report a correctness, crash, or performance bug in FlashInfer
+labels: ["bug"]
+body:
+  - type: textarea
+    id: environment
+    attributes:
+      label: Environment
+      description: >
+        Run `python -m flashinfer.collect_env` (or `flashinfer collect-env`)
+        in the exact environment (same venv/container) where the bug
+        reproduces and paste the full output below. If `import flashinfer` itself is broken, download the
+        standalone script and run it with plain Python:
+        `curl -OL https://raw.githubusercontent.com/flashinfer-ai/flashinfer/main/flashinfer/collect_env.py && python collect_env.py`.
+        The report includes which cuDNN/cuBLAS/CUDA libraries are *actually
+        loaded* vs merely installed — this resolves most "works on my machine"
+        divergence, so please don't trim it.
+      value: |
+        <details>
+        <summary>Output of python -m flashinfer.collect_env</summary>
+
+        ```text
+        PASTE OUTPUT HERE
+        ```
+
+        </details>
+    validations:
+      required: true
+  - type: textarea
+    id: description
+    attributes:
+      label: Bug description
+      description: What happened, and what did you expect to happen instead?
+    validations:
+      required: true
+  - type: textarea
+    id: reproduce
+    attributes:
+      label: Minimal reproduction
+      description: >
+        Smallest self-contained script or command that reproduces the issue.
+        Include exact shapes/dtypes and the backend selected (fa2/fa3/cudnn/
+        trtllm/cutlass/...). If the failure involves JIT compilation, re-run
+        with `FLASHINFER_JIT_VERBOSE=1` and attach the log.
+      render: python
+    validations:
+      required: true
+  - type: textarea
+    id: extra
+    attributes:
+      label: Additional context
+      description: Stack traces, logs, related issues, workarounds you tried.
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -56,11 +56,12 @@ body:
       description: Select every confirmed target.
       multiple: true
       options:
-        - SM80 / SM86 (A100, A30, RTX 30-series)
+        - SM80 / SM86 / SM87 (A100, A30, RTX 30-series, Jetson Orin)
         - SM89 (L4, L40S, RTX 40-series)
         - SM90 (H100, H200, H20)
         - SM100 (B200, GB200)
         - SM103 (B300, GB300)
+        - SM110 (Jetson AGX Thor, DRIVE AGX Thor)
         - SM120 (RTX PRO 6000, RTX 50-series)
         - SM121 (GB10 — DGX Spark, RTX Spark)
         - Future architecture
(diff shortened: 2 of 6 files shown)


.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,186 @@
...
+    id: hardware
+    attributes:
+      label: Target hardware
+      description: Select every confirmed target.
+      multiple: true
+      options:
+        - SM80 / SM86 - Ampere (A100, A30, RTX 30)
+        - SM89 - Ada (L4, L40S, RTX 40)
+        - SM90 - Hopper (H100, H200, or H20)
+        - SM100 - Blackwell datacenter (B200 or GB200)
```
aleozlx: this does not align with product nomenclature

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,64 @@
...
+  issues, crashes, build/install failures, or performance regressions. For new
+  capabilities or speedups, use 🚀 Kernel / Feature Request instead.
+labels: ["bug"]
+body:
+  - type: checkboxes
+    id: presubmit
+    attributes:
+      label: Before submitting
+      options:
+        - label: I have searched existing issues and did not find a solution or an existing report.
```
aleozlx, line 13: for future, i believe it's common practice that a bot flags dup issues and "threatens" to close if not justifying against the duplication. e.g. claude code repo

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,64 @@
...
+        ```
+
+        </details>
+    validations:
+      required: true
+  - type: textarea
+    id: description
+    attributes:
+      label: Bug description
+      description: What happened, and what did you expect to happen instead?
```
aleozlx, line 36: we can separate this into two fields

https://github.com/anthropics/claude-code/blob/main/.github/ISSUE_TEMPLATE/bug_report.yml?plain=1#L30-L49

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,186 @@
...
+      required: true
+
+  - type: dropdown
+    id: inference-engine
+    attributes:
+      label: Inference engine
+      description: Select every engine or integration path this applies to.
+      multiple: true
+      options:
+        - vLLM
```
aleozlx, line 80: for future, this info can be used to auto-label stuff with a git action

aleozlx (commented):
reviewed. one blocking comment on product names

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,64 @@
+name: 🐛 Bug Report
```
aleozlx, line 1: are the emojis in the "name" necessary? i don't get it

.github/ISSUE_TEMPLATE/feature_request.yml
YangXu1990uiuc: This comment is anchored on the previous revision — 4e77e038 (pushed after Brian raised the same point on Slack) dropped the architecture marketing names and segment qualifiers entirely. The dropdown now uses the SM number as the key with product examples only: `SM100 (B200, GB200)`, `SM103 (B300, GB300)`, `SM120 (RTX PRO 6000, RTX 50-series)`, `SM121 (GB10 — DGX Spark, RTX Spark)`. Please take a look at the current revision — if a specific product name is still off, happy to fix.

.github/ISSUE_TEMPLATE/bug_report.yml
YangXu1990uiuc: Agreed — a dedup bot that flags likely duplicates and auto-closes unless the reporter justifies (like anthropics/claude-code) is a good follow-up. Out of scope here; this checkbox is the manual placeholder until then.

.github/ISSUE_TEMPLATE/bug_report.yml
YangXu1990uiuc: Done in 33ebdfdc — split a dedicated optional "Error messages / logs" textarea (render: text) out of the reproduction field, following the claude-code structure. The repro field now asks only for the script.

.github/ISSUE_TEMPLATE/feature_request.yml
YangXu1990uiuc: Agreed — that's part of why these are dropdowns rather than free text: the selections render machine-parseably in the issue body, so a labeler action can map hardware/engine choices to labels. Good follow-up once the templates land.

.github/ISSUE_TEMPLATE/bug_report.yml
YangXu1990uiuc: Ecosystem convention — vLLM ([bug](https://github.com/vllm-project/vllm/blob/main/.github/ISSUE_TEMPLATE/400-bug-report.yml), [feature](https://github.com/vllm-project/vllm/blob/main/.github/ISSUE_TEMPLATE/500-feature-request.yml)) and PyTorch ([bug](https://github.com/pytorch/pytorch/blob/main/.github/ISSUE_TEMPLATE/bug-report.yml), [feature](https://github.com/pytorch/pytorch/blob/main/.github/ISSUE_TEMPLATE/feature-request.yml)) use this exact 🐛/🚀 pair in their template names (SGLang too, with 🐞). They help visual scanning in the template chooser, so keeping them.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,52 @@
...
+        divergence, so please don't trim it.
+      value: |
+        <details>
+        <summary>Output of python -m flashinfer.collect_env</summary>
+
+        ```text
+        PASTE OUTPUT HERE
+        ```
+
+        </details>
```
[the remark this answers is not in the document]
YangXu1990uiuc: Good catch — fixed in 33ebdfdc: switched to `placeholder` + `render: text`, so required validation actually requires user input and the submitted output gets auto code-fenced (which also replaces the details wrapper).

flashinfer/collect_env.py
```diff
@@ -0,0 +1,619 @@
...
+def _installed_distributions():
+    """{normalized_name: version} for every installed distribution."""
+    import importlib.metadata
+
+    dists = {}
+    for dist in importlib.metadata.distributions():
+        name = (dist.metadata.get("Name") or "").strip()
+        if name:
+            dists[name.lower()] = dist.version
+    return dists
```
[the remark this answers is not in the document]
YangXu1990uiuc: Fixed in 33ebdfdc — `_installed_distributions` now applies the same `[-_.]+ → -` canonicalization as `_check_pin`.

flashinfer/collect_env.py
```diff
@@ -0,0 +1,619 @@
...
+        if name in pkgs:
+            pkgs[name] += _framework_pin_note(name, dists)
+    return pkgs
+
+
+def _get_env_vars():
+    info = {}
+    for k in sorted(os.environ):
+        if k.startswith(_ENV_PREFIXES) or k in _ENV_EXACT:
+            info[k] = os.environ[k]
```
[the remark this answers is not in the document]
YangXu1990uiuc: Applied the name-based redaction in 33ebdfdc (KEY/TOKEN/SECRET/PASSWORD/CREDENTIAL → `<redacted>`). Deliberately not rewriting home-directory prefixes: absolute library/executable paths are the forensic payload of this report (loaded-vs-installed disambiguation), and masking them costs more signal than the mild PII it hides.

flashinfer/collect_env.py
```diff
@@ -0,0 +1,619 @@
...
+}
+
+
+def _describe_lib_dir(family, dirpath, paths, dists):
+    """One line of provenance for a directory holding this family's .so files:
+    filename-embedded version and/or the owning pip package."""
+    tags = []
+    versions = set()
+    for p in paths:
+        m = re.search(r"\.so\.(\d+(?:\.\d+)+)$", p)
```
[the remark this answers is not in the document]
YangXu1990uiuc: Done in 33ebdfdc — the regex now accepts single-number suffixes (`(\d+(?:\.\d+)*)`).

aleozlx (approved):
lgtm!

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,186 @@
...
+      required: true
+
+  - type: dropdown
+    id: hardware
+    attributes:
+      label: Target hardware
+      description: Select every confirmed target.
+      multiple: true
+      options:
+        - SM80 / SM86 (A100, A30, RTX 30-series)
```
aleozlx: SM87 Orin? (avoid getting filed into unknown)

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,186 @@
...
+        - SM89 (L4, L40S, RTX 40-series)
+        - SM90 (H100, H200, H20)
+        - SM100 (B200, GB200)
+        - SM103 (B300, GB300)
+        - SM120 (RTX PRO 6000, RTX 50-series)
```
aleozlx, line 65: SM110 Thor? (avoid getting filed into unknown)

.github/ISSUE_TEMPLATE/feature_request.yml
YangXu1990uiuc: Added in edd2607f — folded into the Ampere row: `SM80 / SM86 / SM87 (A100, A30, RTX 30-series, Jetson Orin)`.

.github/ISSUE_TEMPLATE/feature_request.yml
YangXu1990uiuc: Added in edd2607f as `SM110 (Jetson AGX Thor, DRIVE AGX Thor)`. Double-checked the number since Thor was renumbered: CC 10.1 (`sm_101`) in the CUDA 12.x era → CC 11.0 (`sm_110`) since CUDA 13.0 — consistent with our `is_sm110a_supported` (`major == 11` and CUDA ≥ 13.0) and nvcc 13.x's arch list (`compute_110` present, `compute_101` gone).

PR #4017, opened 2026-07-17
skip invalid experts in dispatch

gnovack:
## 📌 Description

When running data parallel deployments, the input batch size is often padded to the largest batch size across all ranks. Inference engines can save communication b/w and compute by marking these pad tokens with an invalid expert id (for example see: https://github.com/vllm-project/vllm/blob/67fe73b2b47e2927af1e5ec6fb91e16e3b075940/vllm/model_executor/layers/fused_moe/modular_kernel.py#L1155-L1160). 

I found that DeepEP has an explicit check for `-1` destination expert id (https://github.com/deepseek-ai/DeepEP/blob/main/deep_ep/include/deep_ep/impls/dispatch.cuh#L104C24-L104C36), but the NVLink One-Sided dispatch in flashinfer does not have such a check. Instead, this kernel currently dispatches all of these pad tokens to rank 0, which can actually degrade performance in cases where the input is heavily padded

This PR adds logic to skip dispatching pad tokens, reusing the existing optional `invalid_token_expert_id` parameter of `MoeAlltoAll.dispatch()`.

### Benchmarks

Benchmarks results were collected using `moe_comm.py` on 4xGB200. Padding was added to simulate one rank processing the full batch size, while others process only 50% of the full batch size worth of tokens.

| Num Tokens             | Before (ms) | After this change (ms) |
| ---------------------- | ----------- | ---------------------- |
| 1024 (3 ranks w/ 512)  | 0.136       | 0.098                  |
| 2048 (3 ranks w/ 1024) | 0.252       | 0.145                  |
| 4096 (3 ranks w/ 2048) | 0.474       | 0.276                  |
| 8192 (3 ranks w/ 4096) | 0.974       | 0.478                  |

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## New Features

* Added optional support for marking invalid expert assignments during MoE all-to-all dispatch.
* Invalid assignments are treated as padding and excluded from routing and data movement.
* Exposed the invalid-expert option through the public Python dispatch API.

## Bug Fixes

* Prevented invalid expert entries from producing destination routing or copy indices.
* Improved validation for configurations containing invalid expert assignments.


aleozlx:
pls address conflict

csrc/nv_internal/tensorrt_llm/kernels/communicationKernels/moeAlltoAllKernels.cu
```diff
@@ -354,10 +355,11 @@ __global__ void moeA2ADispatchKernel(
     uint64_t already_copied = 0;
     for (int k = 0; k < TOP_K; k++) {
       int expert_id = token_selected_experts[local_token_idx * TOP_K + k];
+      bool is_invalid = has_invalid_expert_id && (expert_id == invalid_expert_id);
       // Use contiguous partitioning to determine target rank
       int target_rank = compute_target_rank_id(expert_id, num_experts_per_rank);
```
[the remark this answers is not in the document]
aleozlx: @gnovack could you double check if this makes sense?

aleozlx:
code review: looks good

waiting for testing

csrc/nv_internal/tensorrt_llm/kernels/communicationKernels/moeAlltoAllKernels.cu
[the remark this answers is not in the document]
gnovack: Yeah this makes sense and i initially considered this approach. I opted to explicitly check `expert_id == invalid_expert_id` since i thought it could be a problem to silently skip any expert_id that is `> ep_size * num_experts_per_rank`, since this could indicate some underlying configuration issue. 

My thinking was this: in a happy path, the only expert_id outside of the valid range should be the `invalid_expert_id` placeholder - any other out of range value should result in a failure.

csrc/trtllm_moe_alltoall.cu
```diff
@@ -205,6 +205,10 @@ Tuple<Array<int64_t>, Array<int64_t>, int64_t> moeA2ADispatchOp(
   params.max_tokens_per_rank = static_cast<int>(runtimeMaxTokensPerRank);
   params.top_k = static_cast<int>(topK);
   params.token_selected_experts = static_cast<int32_t const*>(tokenSelectedExperts.data_ptr());
+  if (invalidExpertId.has_value()) {
+    params.has_invalid_expert_id = true;
+    params.invalid_expert_id = static_cast<int32_t>(invalidExpertId.value());
+  }
```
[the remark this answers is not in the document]
bobboli: Too defensive, not necessary.

bobboli, line 282: Add an `else` branch here, do not initialize `has_invalid_expert_id` in struct definition of `MoeA2ADispatchParams.`

csrc/nv_internal/tensorrt_llm/kernels/communicationKernels/moeAlltoAllKernels.cu
```diff
@@ -333,7 +333,8 @@ __global__ void moeA2ADispatchKernel(
     const DispatchKernelPointers ptrs,      // Struct containing all kernel pointers
     int num_payloads,                       // Number of payloads
     int max_tokens_per_rank,                // Maximum tokens per rank
-    int local_num_tokens, int rank_id, int ep_size, int num_experts_per_rank) {
+    int local_num_tokens, int rank_id, int ep_size, int num_experts_per_rank,
+    bool has_invalid_expert_id, int32_t invalid_expert_id) {
```
bobboli: Maybe no need for `has_invalid_expert_id`, `invalid_expert_id` default to -1 is enough.

csrc/nv_internal/tensorrt_llm/kernels/communicationKernels/moeAlltoAllKernels.cu
```diff
@@ -354,10 +355,11 @@ __global__ void moeA2ADispatchKernel(
     uint64_t already_copied = 0;
     for (int k = 0; k < TOP_K; k++) {
       int expert_id = token_selected_experts[local_token_idx * TOP_K + k];
+      bool is_invalid = has_invalid_expert_id && (expert_id == invalid_expert_id);
```
bobboli: We can pass `invalid_expert_id` into `compute_target_rank_id`, and check if expert is invalid at the beginning. If invalid, `compute_target_rank_id` returns -1. We don't handle if `expert_id` contains values other than {`invalid_expert_id`} or [0, ep_size) - that would be too defensive. We can doc in the API that if `expert_id` is not within that set, the behavior is undefined.

csrc/nv_internal/tensorrt_llm/kernels/communicationKernels/moeAlltoAllKernels.cu
```diff
@@ -354,10 +355,11 @@ __global__ void moeA2ADispatchKernel(
...
       // Use contiguous partitioning to determine target rank
       int target_rank = compute_target_rank_id(expert_id, num_experts_per_rank);
 
-      if (already_copied & (1ULL << target_rank)) {
+      if (is_invalid || already_copied & (1ULL << target_rank)) {
```
bobboli: Based on the prev comment about `compute_target_rank_id`, `is_invalid` -> `target_rank > 0` might be more elegant.

csrc/trtllm_moe_alltoall.cu
gnovack: Added an `else` branch to set `invalid_expert_id` to default -1 if not provided. Also removed `has_invalid_expert_id` from the struct definition based on your comment below

csrc/nv_internal/tensorrt_llm/kernels/communicationKernels/moeAlltoAllKernels.cu
gnovack: Removed `has_invalid_expert_id`

csrc/nv_internal/tensorrt_llm/kernels/communicationKernels/moeAlltoAllKernels.cu
gnovack: Updated `compute_target_rank_id` to return `-1` if `expert_id == invalid_expert_id`

csrc/nv_internal/tensorrt_llm/kernels/communicationKernels/moeAlltoAllKernels.cu
gnovack: Code is a bit different after rebase, but i added a `target_rank >= 0` to the condition after `compute_target_rank_id`

bobboli (approved):

gnovack:
Addressed the most recent merge conflicts. @aleozlx @bobboli  please take another look when you get a chance. thanks!

aleozlx (approved):

WoosukKwon:
@aleozlx Can you please help merge this?

PR #4018, opened 2026-07-17
feat(attention): cuTile paged/ragged prefill + paged/MLA decode

yifeis-nv:
## 📌 Description

Adds cuTile-authored attention kernels (public `cuda.tile` Python API) as an **opt-in `backend="cutile"`** on existing FlashInfer wrappers — defaults unchanged.

**Why FlashInfer wants this:** written in the high-level `cuda.tile` Python API instead of CUDA/CUTLASS templates, giving contributors a maintainable path to author and tune Blackwell (SM100/SM103) attention. Registered on existing wrappers so they validate against FlashInfer's own references, at parity on standard shapes.

**Added**
- `fmha_decode_bsr_cutile` — **paged GQA decode** (split-KV reduction, persistent scheduling, CUDA-graph-capturable `run()`) → `BatchDecodeWithPagedKVCacheWrapper(backend="cutile")`; and **DeepSeek MLA paged decode** (head_dim_ckv=512 + head_dim_kpe=64, compressed latent KV) → `BatchMLAPagedAttentionWrapper(backend="cutile")`.
- `fmha_prefill_bsr_cutile` — paged / ragged prefill, two-stage causal masking → `BlockSparseAttentionWrapper(backend="cutile")` and `BatchPrefillWithRaggedKVCacheWrapper(backend="cutile")`.

### Decode perf — B300 / sm103, bf16, GQA group 8 (heads 64/8, head_dim 128)

| batch · s_kv | cuTile ÷ trtllm-gen (SOTA) | cuTile vs fa2_tc |
|---|---|---|
| 1 · 1024 | 0.63× | — |
| 16 · 2048 | 0.98× | +32% |
| 200 · 8192 | 0.97× | +5% |

Decode is GQA-group-sensitive; small groups (e.g. GQA-4) currently run ~0.3× — a kernel-formulation follow-up, not a default regression.

### MLA decode perf — B200 / sm100, bf16, DeepSeek MLA (ckv=512 + kpe=64, page=64), num_heads=32

This PR adds an **arch-gated `TRANS_QK`** operand-swap that makes the QK-GEMM M-dim = BLOCK_N (128) to fill the Blackwell WGMMA tile (small query-groups otherwise leave it 3/4 empty):

| batch · s_kv | pre-TRANS_QK | **this PR** | speedup | trtllm-gen (SOTA) |
|---|---|---|---|---|
| 16 · 2048 | 0.040 ms | **0.023 ms** | 1.7× | 0.018 ms |
| 200 · 8192 | 1.386 ms | **0.497 ms** | **2.8×** | 0.543 ms |

`TRANS_QK` is gated to datacenter Blackwell (sm100/sm103) via `_supports_trans_qk`; sm120 / A100 keep the byte-identical original path (the transposed [512, BLOCK_H] accumulator would spill on their smaller SMs). Correctness unchanged (MLA 18/18). cuTile MLA decode now essentially matches trtllm-gen at large batch.

### Prefill perf — B300 / sm103, dense paged, vs fa2

| shape | cuTile ÷ fa2 |
|---|---|
| bs4 · s2048 · g4 · causal | 0.93× |
| bs1 · s4096 | 1.11× |

### Backend-comparison benchmarks

`benchmarks/bench_{ragged_prefill,paged_prefill,mla_decode,gqa_decode}_backend_comparison.py` (canonical `--providers`/`--csv`, inline correctness verify, `bench_gpu_time` kernel self-time, speedup-vs-trtllm table + heatmap). Two-flashinfer: cuTile from this checkout vs the trtllm-gen SOTA from an installed flashinfer, measured in separate processes and merged.

Full ocean-workload sweeps on **B200 (sm100)**, median kernel time, cuTile ÷ trtllm-gen (>1 = cuTile faster):

| op | sweep | geomean | range |
|---|---|---|---|
| **ragged prefill** (hd_qk=192, DeepSeek) | batch{1,16} × s_kv{1k…8k} | **0.60×** | 0.53–0.67× |
| **paged prefill** | page{16,64} × batch{1,16} × s_kv{1k…8k} | 0.60× | **page64 0.79–1.13×** · page16 0.33–0.48× |
| **MLA decode** (num_heads=32) | page{32,64} × batch{1…256} × s_kv{1k…8k} | 0.85× | **page64 batch≥128: 1.17–1.28×** · page32 0.57–0.85× |
| **GQA decode** | page{32,64} × batch{1…256} × s_kv{1k…8k} | 0.85× | **batch≥64: 0.88–0.97×** · batch≤16 0.56–0.89× |

Reading: cuTile **beats** trtllm-gen on MLA decode in the throughput regime (page64, batch ≥128) and on small-batch paged prefill with page64; it **converges to ~0.96–0.97×** on GQA decode at serving scale. It lags on **small pages** (page16 prefill, page32 MLA) and on **ragged prefill hd192**, both of which are cuTile-compiler scheduling effects under active work upstream — neither is on a default path here, since the backend is opt-in.

## 🔍 Related Issues

Part of a 3-way split of the cuTile migration (attention / quantization / GEMM), replacing #3959. Sibling PRs: #4019 (quantization), #4020 (GEMM).

### ✅ Pre-commit Checks

## 🧪 Tests

Measured on B300 / sm103:
- decode native test: 180 passed / 900 skipped
- block-sparse prefill (narrow slice): 12 passed
- standalone prefill: 7 passed
- decode `run()` CUDA-graph capture: replay bit-exact vs eager
- MLA paged decode (`BatchMLAPagedAttentionWrapper(backend="cutile")`) vs torch reference: 18 passed (batch×seq×page_size×heads sweep + preallocated-out)
- ragged prefill wrapper (`BatchPrefillWithRaggedKVCacheWrapper(backend="cutile")`) wired + tested

## Reviewer Notes

The backend is opt-in on every wrapper — no default path changes. The perf table above is the standing on the public toolchain over the full ocean workload sweep, including the regimes where cuTile currently lags (small pages, ragged hd192); those are tracked as kernel/compiler follow-ups rather than blockers for landing an opt-in backend.

## Summary by CodeRabbit

* **New Features**
  * Added cuTile-backed attention for paged and ragged KV prefill.
  * Added cuTile paged decode support for standard and MLA KV cache layouts, including split-KV reduction.
  * Introduced a new cuTile backend option for batch decode and block-sparse attention with backend-specific runtime limitations.

* **Bug Fixes**
  * Improved block-sparse test coverage to work when SciPy is unavailable.

* **Tests**
  * Added CUDA-only cuTile prefill tests and GPU-only cuTile MLA decode tests.
  * Expanded decode and block-sparse tests to run across the cuTile backend.

### DIFF
--- a/benchmarks/bench_gqa_decode_backend_comparison.py
+++ b/benchmarks/bench_gqa_decode_backend_comparison.py
@@ -1,0 +1,470 @@
+"""
+Copyright (c) 2025 by FlashInfer team.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+Benchmark: GQA Decode Backend Comparison (cuTile vs trtllm-gen SOTA)
+
+Compares FlashInfer's cuTile GQA-decode backend against the trtllm-gen SOTA
+kernel over ocean-eval's DecodePaged workload (LLaMA-style GQA: 128 query heads
+/ 8 KV heads, head_dim=128, bf16), sweeping batch x s_kv x page_size:
+
+  - ``cutile`` : this checkout's ``BatchDecodeWithPagedKVCacheWrapper`` planned/run
+                 with ``backend="cutile"`` (pure cuda.tile kernel, kv_layout="NHD").
+  - ``trtllm`` : ``flashinfer.decode.trtllm_batch_decode_with_kv_cache`` (trtllm-gen,
+                 kv_layout="HND"), the ocean-eval dashboard SOTA baseline.
+
+TWO-FLASHINFER NOTE: the two providers are meant to be measured in SEPARATE
+processes (this checkout for ``cutile``; the image's installed flashinfer for
+``trtllm`` -- this checkout's trtllm path may crash on some arches). Each
+provider independently try/except-es and degrades to N/A (NaN), so the two
+result CSVs can be produced separately and merged. Baseline default = trtllm.
+
+Usage (typical two-run merge):
+    python3 bench_gqa_decode_backend_comparison.py --providers cutile --csv cutile.csv
+    python3 bench_gqa_decode_backend_comparison.py --providers trtllm --csv trtllm.csv
+    # then concatenate cutile.csv + trtllm.csv (drop the duplicate header)
+
+Requirements:
+    - cuda.tile toolchain (cutile provider)
+    - flashinfer with trtllm-gen cubins (trtllm provider)
+    - matplotlib for the optional heatmap
+"""
+
+import argparse
+import csv as _csv
+import math
+import numpy as np
+import torch
+from typing import Callable, Dict, List, Tuple
+
+from flashinfer.testing.utils import bench_gpu_time
+
+# LLaMA-style GQA problem geometry (fixed).
+NUM_QO_HEADS = 128
+NUM_KV_HEADS = 8
+HEAD_DIM = 128
+DTYPE = torch.bfloat16
+
+
+# --------------------------------------------------------------------------- #
+# Provider availability
+# --------------------------------------------------------------------------- #
+def _has_cutile() -> bool:
+    try:
+        import cuda.tile  # noqa: F401
+        from flashinfer.decode import (  # noqa: F401
+            BatchDecodeWithPagedKVCacheWrapper,
+        )
+
+        return True
+    except Exception:
+        return False
+
+
+def _has_trtllm() -> bool:
+    try:
+        from flashinfer.decode import trtllm_batch_decode_with_kv_cache  # noqa: F401
+
+        return True
+    except Exception:
+        return False
+
+
+ALL_PROVIDERS = ["cutile", "trtllm"]
+_AVAIL = {
+    "cutile": _has_cutile,
+    "trtllm": _has_trtllm,
+}
+
+
+# --------------------------------------------------------------------------- #
+# Shared paged-GQA input generation
+# --------------------------------------------------------------------------- #
+class _GQAInputs:
+    """Raw tensors shared by both providers (identical values -> comparable outputs).
+
+    KV cache is generated in NHD layout ``[num_pages, page, num_kv_heads, head_dim]``;
+    the trtllm (HND) path takes a transposed view of the same values.
+    """
+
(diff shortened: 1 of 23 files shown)


yifeis-nv:
Thanks for the reviews. Pushed `adf4ac011` addressing the three genuine issues, validated on B300 (sm103): decode native `180 passed / 900 skipped`, standalone prefill `7 passed`.

**Fixed**
- **Decode split-KV reduce (non-pow2 `NUM_KV_SPLITS`)** — `attn_splitk_out` is now loaded with `padding_mode=ZERO`, so a non-power-of-two split count can't read past the buffer. The padded split rows carry weight 0 (their lse is padded to `-inf`), so the zeros don't perturb the weighted reduction.
- **Ragged prefill padded-KV softmax denominator** — the final `BLOCK_N` tile is `PAD_ZERO`-loaded, so padded columns had `qk == 0` and were adding `exp2(-m_ij)` mass to the denominator (their V rows are zero, so only the denominator was corrupted). Now masked with the `offs_n < off_band_hi` boundary, matching the paged path. No-op when `off_band_hi` is a multiple of `BLOCK_N`.
- **GQA ratio guard** — both prefill wrappers now reject `num_qo_heads % num_kv_heads != 0`.

- **`decode.py` `plan()` "stale block table"** — false positive: `self._block_tables` is reassigned from the `plan()` argument at the top of every `plan()` call (`decode.py:1511`), so the `if self._block_tables is None:` branch rebuilds from the current CSR state each plan; nothing stale survives across plans.
- **`run()` `block_tables.to(...)` during capture** — the tensor is already staged on `self.device` as `int32` at plan time, so `.to(device, int32)` returns the same tensor (no alloc, no H2D) and is capture-safe.
- **Reject unsupported block-sparse options** — already enforced in the cuTile `plan()` (`sparse.py:441-507`): `mask`/`packed_mask`, `logits_soft_cap`, `pos_encoding_mode`, `causal` under the gathered-block mapping, non-divisible GQA, and out-of-range indices all raise today.
- **Unequal Q/KV causal (bottom-right alignment)** — not reachable through the public wrappers: block-sparse forces `causal=False`, and the causal prefill path is equal-length self-attention. The reachable unequal-length non-causal path is exactly what the padded-KV mask above fixes.
- **Output-dtype override on the early-return / block-sparse paths** — the kernels allocate output from `q` (dtype = `q.dtype`), which is the cuTile backend's contract; a distinct planned `o_data_type` isn't part of it.
- **Test SM/availability gating + independent-reference cases** — the cuTile tests already skip when `cuda.tile` is unavailable and run on the SM100/SM103 CI. Test-strengthening (SM100 gate helper, non-`BLOCK_N`-divisible KV reference case) is a fair follow-up; leaving it out of this change to keep the diff to the correctness fixes.

yifeis-nv:
@flashinfer-bot run

saltyminty (approved):

saltyminty:
Checking

saltyminty:
Verified that the cutile tests pass locally, but CI doesn't run them so adding the `is_cuda_tile_available()` as is the precedent elsewhere in flashinfer. Will open a follow-up PR to enable CI.

PR #4019, opened 2026-07-17
feat(quantization): cuTile per-token-group 8bit quant + fused RoPE-FP8

yifeis-nv:
## 📌 Description

Adds two cuTile quantization kernels (public `cuda.tile` API), opt-in `backend="cutile"`.

**Why FlashInfer wants this:** `rope_quantize_fp8` fuses RoPE + FP8 quantization into one pass for the MLA / DeepSeek paged-KV append path (replaces separate rope + quant launches); `per_token_group_quant_8bit` provides the block-scaled FP8/INT8 quantizer feeding the groupwise FP8 GEMM path.

**Added**
- `per_token_group_quant_8bit_cutile` — FP8 / INT8 per-token-group quant, row/col-major scale.
- `rope_quantize_fp8_cutile` — fused RoPE + FP8 quant (MLA 2D latent key) → `rope_quantize_fp8(..., backend="cutile")`.

### Benchmarks

Canonical `benchmarks/bench_*_backend_comparison.py` (argparse `--providers`/`--csv`, inline correctness verify, `bench_gpu_time` kernel self-time, speedup-vs-SOTA table + heatmap):
- `bench_per_token_group_quant_8bit.py` — cuTile vs **sgl_kernel** (SOTA) / SGLang triton / torch, over a (num_tokens × hidden_dim) sweep.
- `bench_rope_quantize_fp8_backend_comparison.py` — fused MLA RoPE+fp8 cuTile vs the **native fused CUDA** kernel (SOTA) / torch, over a num_tokens sweep.

Measured on **B200 (sm100)**, median kernel time, speedup = SOTA ÷ cuTile (>1 = cuTile faster).

**`per_token_group_quant_8bit`** vs sgl_kernel — sweep num_tokens{1…4096} × hidden{2048,4096,7168}, **geomean 0.72×**. cuTile is at parity while the op is launch-bound and falls behind once it becomes bandwidth-bound:

| num_tokens | hidden 2048 | 4096 | 7168 |
|---|---|---|---|
| 1–64 | 1.00× | 1.00× | 1.00× |
| 128 | 0.99× | 1.00× | 0.79× |
| 256 | 1.00× | 0.80× | 0.71× |
| 512 | 0.80× | 0.71× | 0.57× |
| 1024 | 0.71× | 0.58× | 0.50× |
| 4096 | 0.45× | 0.42× | 0.39× |

SGLang's triton kernel tracks cuTile closely (0.9–1.0× across the sweep); torch is 3–6× slower than both. So cuTile is competitive with sgl_kernel at small token counts and ~2–2.5× behind at serving-scale batches — a per-token-group memory-access follow-up, not a default-path regression (the backend is opt-in).

**`rope_quantize_fp8` (MLA, 128 heads, head_size 576)** vs the native fused CUDA kernel — sweep num_tokens{1…4096}, **geomean 0.87×**: cuTile **wins at ntok ≤ 16** (1.00–1.10×), ~0.90–0.94× at 32–128, settling at 0.75–0.83× for ntok ≥ 256. Both are ~3× faster than the torch reference.

## 🔍 Related Issues

Part of a 3-way split of the cuTile migration (attention / quantization / GEMM), replacing #3959. Sibling PRs: #4018 (attention), #4020 (GEMM).

### ✅ Pre-commit Checks

## 🧪 Tests

Measured on B300 / sm103:
- rope MLA + per-token-quant: 66 passed
- GQA/MHA rope correctly gated (`NotImplementedError`, not a crash)
- new coverage: non-power-of-two `group_size`, column-major scale strides, TMA scale padding, UE8M0 power-of-two scales

## Reviewer Notes

Carries the shared `cached_replace_hints` helper in `flashinfer/cutile/cutile_common.py` (also in the GEMM PR, #4020) — whichever merges second drops the duplicate hunk on rebase.

## Summary by CodeRabbit

* **New Features**
  * Added public per-token-group 8-bit quantization API with FP8/INT8 outputs, row/column-major scales, TMA-aligned scales, and UE8M0 encoding.
  * Added a selectable `cutile` backend for fused FP8 RoPE quantization.
  * Re-exported an additional FP8 quantization utility in the public quantization API.
* **Performance**
  * Reduced repeated cuTile compilation overhead via cached hinted-kernel launches.
* **Tests**
  * Extended RoPE and per-token-group 8-bit quantization tests to validate both CUDA and cuTile paths and scale layout/edge cases.

### DIFF
--- a/flashinfer/rope.py
+++ b/flashinfer/rope.py
@@ -1377,6 +1377,7 @@ def rope_quantize_fp8(
     q_nope_out: Optional[torch.Tensor] = None,
     k_nope_out: Optional[torch.Tensor] = None,
     enable_pdl: bool = False,
+    backend: str = "cuda",
 ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
     r"""Apply RoPE (Rotary Positional Embeddings) and quantize to FP8 format.
 
@@ -1423,12 +1424,56 @@ def rope_quantize_fp8(
         Pre-allocated output tensor for quantized key (non-rotary). If ``None``, allocated automatically.
     enable_pdl : bool
         Whether to enable PDL (Programmatic Dependent Launch). Default: ``False``.
+    backend : str
+        Implementation backend. ``"cuda"`` (default) uses the fused CUDA kernel;
+        ``"cutile"`` uses the cuda.tile Python kernel.
 
     Returns
     -------
     Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]
         Quantized tensors: (q_rope_out, k_rope_out, q_nope_out, k_nope_out).
     """
+    if backend == "cutile":
+        if is_neox:
+            # The cuTile kernel implements only the interleaved (GPT-J) rotary
+            # layout. is_neox defaults to True, so surface a clear error instead
+            # of the kernel's bare AssertionError (which -O would also strip).
+            raise NotImplementedError(
+                "backend='cutile' supports is_neox=False (interleaved/GPT-J "
+                "rotary) only; got is_neox=True."
+            )
+        if k_rope.ndim != 2:
+            # The kernel addresses the key as 2D [tokens, rope_dim] (single
+            # shared latent K head, the MLA use case). A 3D GQA/MHA key would
+            # fail deep in autotune with an opaque TileTypeError.
+            raise NotImplementedError(
+                "backend='cutile' rope_quantize_fp8 supports MLA-style 2D key "
+                f"tensors (single shared K head) only; got {k_rope.ndim}D key "
+                "(GQA/MHA multi-head K is not supported)."
+            )
+        if cos_sin_cache.dtype != torch.float32:
+            raise ValueError("cos_sin_cache should be float32")
+        from .quantization.kernels.cutile.rope_quantize_fp8_cutile import (
+            rope_quantize_fp8_cutile,
+        )
+
+        return rope_quantize_fp8_cutile(
+            q_rope,
+            k_rope,
+            q_nope,
+            k_nope,
+            cos_sin_cache,
+            pos_ids,
+            is_neox=is_neox,
+            quantize_dtype=quantize_dtype,
+            quant_scale_q=quant_scale_q,
+            quant_scale_kv=quant_scale_kv,
+            q_rope_out=q_rope_out,
+            k_rope_out=k_rope_out,
+            q_nope_out=q_nope_out,
+            k_nope_out=k_nope_out,
+        )
+
     if cos_sin_cache.dtype != torch.float32:
         raise ValueError("cos_sin_cache should be float32")
 
--- a/flashinfer/rope.py
+++ b/flashinfer/rope.py
@@ -1433,6 +1433,11 @@ def rope_quantize_fp8(
     Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]
         Quantized tensors: (q_rope_out, k_rope_out, q_nope_out, k_nope_out).
     """
+    if backend not in ("cuda", "cutile"):
+        raise ValueError(
+            f"Unsupported backend for rope_quantize_fp8: {backend!r}; "
+            "expected 'cuda' or 'cutile'."
+        )
     if backend == "cutile":
         if is_neox:
             # The cuTile kernel implements only the interleaved (GPT-J) rotary
--- a/benchmarks/bench_per_token_group_quant_8bit.py
+++ b/benchmarks/bench_per_token_group_quant_8bit.py
@@ -1,34 +1,70 @@
 """
-Benchmark: per-token-group 8-bit quantization — cuTile backend vs SOTA.
+Copyright (c) 2025 by FlashInfer team.
 
-Compares FlashInfer's migrated ``backend="cutile"`` kernel (PR #4019) against
-the SGLang ``sgl_kernel`` reference (the SOTA baseline used by ocean-eval's
-dashboard for this op) and a PyTorch-native reference.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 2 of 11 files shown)


yifeis-nv:
Thanks for the reviews. Pushed `3d0989d40`. Validated on B300 (sm103): `per_token_group_quant` `26 passed`, `rope` cutile `48 passed / 576 skipped`.

**Fixed**
- **Padded-lane absmax corruption (genuine bug)** — `BLOCK = next_pow2(group_size)`, so for a non-power-of-two `group_size` the trailing lanes `[group_size, BLOCK)` gather **in-bounds data from the next group** (gather's `padding_value` only fires on true OOB), corrupting the per-group `absmax` and thus the scales. Now masked with `ct.where(mask, y, 0.0)` before the reduction, in both the row- and column-major kernels. Added a `test_non_pow2_group_size` regression test (group_size 96/48) that fails without the mask.
- **`dst_dtype` validation** — the `else: torch.finfo(dst_dtype)` branch silently accepted fp16/fp32; now rejects anything outside `{float8_e4m3fn, float8_e5m2, int8}`.
- **`column_major_scales` rank guard** — the permute / TMA / `g_id` addressing assume 2D; now rejects non-2D inputs instead of mis-allocating.
- **`rope_quantize_fp8_cutile` dtype guard** — rejects an inferred `quantize_dtype` that isn't an FP8 type before it reaches `ct.astype`.
- **`rope_quantize_fp8` backend validation** — unknown backend values now raise instead of silently falling through to CUDA (matches the sibling `per_token_group_quant_8bit`).
- **Tests** — switched to `flashinfer.utils.get_compute_capability`; added a column-major stride assertion and a UE8M0 power-of-two-scale assertion.

- **`cached_replace_hints` thread lock** — under CPython the `.get()`→`.move_to_end()` sequence has no GIL-release point, so the `popitem`/`move_to_end` `KeyError` race can't occur; the only effect of true concurrency is a rare, benign redundant `replace_hints` compile (the second insert just overwrites). Not worth a lock on the cache hot path. (Same helper appears in the GEMM PR; keeping both consistent.)
- **`check_bounds` on scatter/store** — the quantized-value scatters already pass `check_bounds=True`; the `y_scale` scatter indices are provably in range (`g_id ∈ [0, M)` with array size `M`; column-major offset `< num_groups * col_stride`), and the OOB-sentinel `ct.where` branch is never taken.
- **Unbounded tune cache** — the autotune cache is keyed on shape (matches the Triton convention); an LRU would thrash and re-run the (expensive) `exhaustive_search`. Distinct-shape count is bounded in practice.
- **`TraceTemplate` for `per_token_group_quant_8bit`** — bare `@flashinfer_api` is an established pattern in the same file; adding trace replay is a feature, tracked separately.
- **Separate Q/K non-RoPE widths** — by this API's contract `q_in`/`k_in` share `total_dim = rope_dim + no_rope_dim`, so `q_nope`/`k_nope` have equal width (e.g. MLA 512/512); the shared-width assumption is a design invariant, not a bug.

yifeis-nv:
Thanks for the latest pass. Addressed the critical finding in `per_token_group_quant_8bit_cutile.py` (pushed in `dd34416`):

**`.contiguous()` on the TMA-aligned scales (was line ~302) — valid, fixed.** The `scale_tma_aligned` branch returned `x_s_raw[:, :num_tokens].t().contiguous()`. The `.t()` already produces the intended `(num_tokens, num_groups)` column-major view (token dim unit-stride, group stride = `aligned_size`, preserving the 16B TMA leading-dim padding); the trailing `.contiguous()` repacked it into a tight row-major tensor, destroying both the column-major property and the padding this branch exists to produce — contradicting the API docstring ("column-major (Fortran) layout" + "padded ... leading dimension for TMA alignment"). Now returns the view directly.

Added `test_tma_aligned_scales` (num_tokens ∈ {128, 130}) asserting `stride(0)==1` (column-major) **and** `stride(1)==ceil_align(num_tokens,4)` (padding preserved), plus dequant correctness through the padded view — the previous code would have failed both stride assertions. This is also the coverage the earlier "add aligned-segment coverage" note asked for. Verified on B300/sm103: the new test passes (2/2) and the full quant file is green (28 passed).

The `.contiguous()`-was-here shape-tuple-unpacking nitpick is subsumed by the same region rewrite.

bkryu:
@flashinfer-bot run

flashinfer/rope.py
```diff
@@ -1377,6 +1377,7 @@ def rope_quantize_fp8(
     q_nope_out: Optional[torch.Tensor] = None,
     k_nope_out: Optional[torch.Tensor] = None,
     enable_pdl: bool = False,
+    backend: str = "cuda",
```
bkryu, line 1380: Hi @yifeis-nv, for `rope_quantize_fp8(backend='cutile')` I am questioning the value add of the kernel.

We get a lot of questions users about what backend they should pick for APIs with multiple backends. Often times, it is historical: the nvfp4_quantize has cuda and cute-dsl because the cuda backend was written first, then I wrote the cute-dsl kernel that outperforms the cuda kernel on most problem sizes. Here, it seems like the cutile kernel does not yet beat the CUDA implementation and has a narrower support surface.

I'd like to suggest that the cuTile kernel gets iterated a bit more until it can justify having multiple backends

bkryu (commented):
Hi @yifeis-nv thanks for opening the PR. It seems like the PR currently:
* Adds a new `flashinfer.quantization.per_token_group_quant_8bit` API with a cuTile kernel underneath
* Adds a new backend `flashinfer.rope.rope_quantize_fp8(backend='cutile')` to an existing API.

The `per_token_group_quant_8bit` as a new API that parallels SGLang's kernel is valuable, but I left a question on `rope_quantize_fp8(backend='cutile')`. Can you either comment on it or wait until we have a better implementation?

Sidenote: Due to us taking long to review the kernel, it seems like the branch is quite a bit behind main. Can you rebase or merge main?

bkryu:
@flashinfer-bot run

bkryu (approved):

PR #4020, opened 2026-07-17
feat(gemm): cuTile alpha-beta / masked-bmm / ragged-bmm / block-scaled FP8 GEMM

yifeis-nv:
## 📌 Description

Adds five cuTile GEMM kernels (public `cuda.tile` API), registered as `backend="cutile"` on existing `flashinfer.gemm` ops — defaults unchanged.

**Why FlashInfer wants this:** `gemm_fp8_nt_groupwise` gains a CUDA-graph-safe block-scaled FP8 path that beats cutlass on aligned MoE segments; `masked_bmm` / `ragged_bmm` / `gemm_alpha_beta` fill batched/ragged GEMM gaps in the public API.

**Added:** `gemm_alpha_beta`, `masked_bmm`, `ragged_bmm`, `ragged_block_scaled_bmm`, `gemm_fp8_nt_groupwise` (single + grouped + aligned-segment TMA fast path via `segment_alignment`).

### FP8 / bmm perf — B300 / sm103, FP8 vs cutlass; bmm vs cutlass bf16

| op · shape | cuTile | native | ratio |
|---|---|---|---|
| gemm_fp8_nt_groupwise single · 4096³ | 1535 TF | 1863 TF (cutlass) | 0.82× |
| gemm_fp8_nt_groupwise single · 2048³ / 2560³ | — | cutlass | 0.85× / 0.84× |
| group_gemm_fp8 **aligned-TMA** · m1024·g8 | — | cutlass | **1.03×** |
| group_gemm_fp8 **aligned-TMA** · m512·g4 / m128·g4 | — | cutlass | 0.66× / 0.60× |
| group_gemm_fp8 arbitrary (gather) · m1024·g8 | 524 TF | cutlass | 0.29× |
| masked_bmm (bmm_bf16) · 8×1024³ | 737 TF | 625 TF (cutlass) | 1.22× |

`gemm_fp8` single's ~0.82× is a cuTile-vs-cutlass FP8-MMA gap (shared by the persistent bf16 GEMM); aligned callers get the TMA fast path (`segment_alignment=128`). Arbitrary (non-aligned) segments use the gather path.

### Dense bf16 GEMM benchmark

`benchmarks/bench_mm_bf16_backend_comparison.py` (canonical `--providers`/`--csv`, inline correctness verify, `bench_gpu_time` kernel self-time, speedup-vs-CUTLASS table + heatmap): cuTile `mm_bf16` vs **CUTLASS** (SOTA) / cuDNN / torch, over square {1024…8192}³ plus weight-fixed m-sweeps at (n,k) = (4096,4096) and (8192,8192).

On **B200 (sm100)**, cuTile ÷ each backend (>1 = cuTile faster) — **geomean 1.34× vs CUTLASS, 1.15× vs cuDNN, 0.86× vs torch/cuBLAS**:

| m · n · k | cuTile TFLOP/s | ÷ cutlass | ÷ cudnn | ÷ torch |
|---|---|---|---|---|
| 1024³ | 161 | 1.17× | 3.90× | 0.77× |
| 2048³ | 671 | 1.24× | 1.96× | 0.80× |
| 8192³ | 1415 | **3.07×** | 1.20× | 1.20× |
| 16 · 4096 · 4096 | 25 | 1.00× | 1.05× | 1.00× |
| 512 · 4096 · 4096 | 620 | 1.66× | 1.07× | 1.07× |
| 4096 · 4096 · 4096 | 1219 | 1.56× | 0.81× | 0.81× |
| 64 · 8192 · 8192 | 204 | 0.90× | 1.12× | 0.85× |
| 2048 · 8192 · 8192 | 1285 | 1.63× | 0.80× | 0.80× |

cuTile clearly beats FlashInfer's CUTLASS and cuDNN backends, with the lead growing with problem size (up to ~3× at 8192³); it trails only on tiny-M (16/64-row) shapes. Against vendor cuBLAS/torch it trades the lead by shape — winning the large square GEMMs (8192³: 1.20×) and the 512-row weight-fixed shapes, behind on the rest.

`cuBLASLt` / `TGV` / `TinyGEMM` report **N/A with a reason** in the sweep — these are benchmark-image build limits, not B200 unsupport: TGV needs `nvidia-cutlass-dsl` with CuTe-Experimental (the bundled 4.4.2 is rejected), cuBLASLt's JIT source `mm_bf16_cublaslt.cu` is absent from the image, and TinyGEMM's JIT build fails on a missing symbol.

## 🔍 Related Issues

Part of a 3-way split of the cuTile migration (attention / quantization / GEMM), replacing #3959. Sibling PRs: #4018 (attention), #4019 (quantization).

### ✅ Pre-commit Checks

## 🧪 Tests

Measured on B300 / sm103:
- group_gemm_fp8 `-k cutile`: 144 passed / 144 skipped
- ragged_block_scaled_bmm: 2 passed — bit-identical to native `group_gemm_fp8_nt_groupwise` (trtllm)
- ragged_bmm: 9 passed / 6 skipped · gemm_alpha_beta: 16 passed (incl. the `beta==0` NaN-C regression) · masked_bmm: 8 passed
- CUDA-graph capture + replay: bit-exact vs eager

## Reviewer Notes

Carries the shared `cached_replace_hints` helper in `flashinfer/cutile/cutile_common.py` (also in the quantization PR, #4019) — whichever merges second drops the duplicate hunk on rebase.

## Summary by CodeRabbit

* **New Features**
  * Added new cuTile GEMM entry points: `masked_bmm`, `ragged_bmm`, `ragged_block_scaled_bmm`, and `gemm_alpha_beta`.
  * Added `segment_alignment` support to grouped FP8 GEMM on the cuTile backend.
* **Performance**
  * Improved grouped FP8 execution by moving from per-group launches to a single fused persistent on-device kernel.
  * Added LRU memoization for hinted cuTile kernels to reduce repeated recompiles.
* **Bug Fixes**
  * Ensures `beta=0` ignores existing output values to prevent NaN/invalid propagation.
* **Tests**
  * Added new cuTile correctness coverage for the new GEMM entry points and alignment behavior.


yifeis-nv:
Thanks for the reviews. Pushed `7bd8ffc88`. Validated on B300 (sm103): `ragged_block_scaled 2 passed`, `ragged_bmm 9 passed / 6 skipped`, `gemm_alpha_beta 16 passed`, `masked_bmm 8 passed`.

**Fixed**
- **Block-scaled `BLOCK_N=256` broadcast (silent wrong output, sm100/sm103)** — the kernel loads a single `b_scale` per `(pid_q, offs_bsn, k)` and broadcasts it across the whole `BLOCK_N` tile, so a `BLOCK_N=256` swapped config applied the first 128-column group's scale to the second group. The sm90 branch already caps this at 128 with an in-code note explaining the exact hazard; the sm100/sm103 (`else`) branch was still emitting `BLOCK_N=256`. Capped it to 128 there too.
- **`ragged_bmm` grid sizing** — the persistent grid now sizes on `total_m` instead of the host `max_m` hint (3 sites). A stale/zero hint gave 0 tiles → empty grid → unwritten output. The kernel already reads its per-tile bound from `max_m_device`; this only fixes the launched program count. Matches the `ragged_block_scaled_bmm` sibling.
- **Aligned-TMA dispatch guard** — `gemm_fp8_nt_groupwise` now routes to the aligned `ragged_block_scaled_bmm` fast path only when `block_n == block_k and K % block_k == 0` (that kernel indexes `b_scale` as `n_offset // BLOCK_K` with a fixed `BLOCK_K=128`). Other granularities fall back to the arbitrary-granularity gather kernel.
- **Alias check** — rejects only storage overlap between the writable output `C` and an input; `a`/`b` sharing storage (e.g. a Gram matrix `A @ A.T`) is read-only and allowed.
- **Device binding** — the launch stream and SM-count query now bind to `a.device` across `masked_bmm` / `ragged_bmm` / `ragged_block_scaled_bmm` / `gemm_alpha_beta` (multi-GPU correctness).
- **Tests** — `ragged_bmm` now asserts the FP8-rejection contract (the backend is fp16/bf16-only) instead of expecting success (this case had been failing on sm100/sm103); tightened the block-scaled tolerance from `rtol=1.0` (which accepted 100%-off output) to `2e-2`.

- **`cached_replace_hints` thread lock** — under CPython the `.get()`→`.move_to_end()` sequence has no GIL-release point, so the `popitem`/`move_to_end` race can't occur; the only effect of true concurrency is a rare, benign redundant `replace_hints` compile. Kept lock-free (same helper/decision as the quantization PR).
- **`TraceTemplate` for the new GEMM APIs** — bare `@flashinfer_api` is a supported pattern here; trace replay is a feature, tracked separately.
- **Per-candidate `c.clone()` in autotune** — runs once per config in a one-time cached tuning pass; the M×N clone is negligible against per-candidate compile + benchmark.
- **Test compute-capability `major*10+minor`** — `is_backend_supported(backend, cc)` keys on the coarse `major*10` form, so refining to `103` would wrongly skip cutile on sm103 nodes. Left as `major*10`.

dhiraj113 (commented):
Review of the cuTile GEMM additions, scoped to **crashes and correctness bugs only** for now (design/naming/benchmark-hygiene notes to follow separately if useful).

Every finding below was verified by **executing a repro against this PR's head (`10b5ad9c`) on a B200 (sm100)** — each inline comment carries the exact command and its observed output. This was an AI-assisted review (per the repo's agent-review guidance in `docs/code_review_guidance.md`); all findings survived an adversarial verification pass before posting, and none are speculative.

**Crashes / silent corruption:**
1. `masked_scaled_bmm`: a `Q>1024` call permanently corrupts the cached per-device scratch tensor — every subsequent call on that device fails (`masked_scaled_bmm_cutile.py:66`).
2. `masked_bmm`: zero-sized `Q`/`M`/`N` launches a zero-sized grid → opaque error, where `torch.bmm` returns an empty tensor (`masked_bmm_cutile.py:329`).
3. `ragged_scaled_bmm`: an understated host `max_m` silently returns uninitialized `torch.empty` rows — no `max_m_device` defense exists in this op, unlike its siblings (`ragged_scaled_bmm_cutile.py:134`).
4. `ragged_bmm`: the documented `max_m_device` defense-in-depth is circular on the default path — when `max_m_device=None` (the only path the public API uses), it is synthesized from the same untrusted host `max_m` (`ragged_bmm_cutile.py:766`).

**Validation gaps that turn bad inputs into silent wrong numerics or opaque internal errors:**
5. `_cutile_ragged_block_scaled_bmm_requirement` is a no-op (`return True` under a docstring claiming validation) — wrong-dtype inputs are silently accepted or die deep in `ct.mma` (`gemm_base.py:1208`).
6. `ragged_block_scaled_bmm`: `a_scale` K-granularity is never validated — the missing counterpart of the `b_scale` guard added earlier in this PR; a mismatched `a_scale` is silently zero-padded or mis-applied (`ragged_block_scaled_bmm_cutile.py:308`).
7. `masked_scaled_bmm`: `max_m_device` is never checked against `a.device` (the sibling op got this check after earlier review), and the CPU-tensor branch hides a host sync that contradicts the docstring and breaks CUDA-graph capture (`masked_scaled_bmm_cutile.py:725`).

**Test coverage bugs (shipped-but-never-executed paths):**
8. The 106-entry masked_bmm shape sweep is truncated by `[:2]` to a single case tested twice (`test_masked_bmm_cutile.py:120`).
9. `MIXED_PREC` (fp8×fp4) and `mxfp4` are accepted by the gate but never tested (`test_masked_scaled_bmm_cutile.py:157`).
10. `ragged_block_scaled_bmm` is only ever tested with perfectly uniform segments — the prefix-sum schedule's motivating cases (imbalanced/zero-row) are unverified (`test_ragged_block_scaled_bmm_cutile.py:309`).
11. The new `segment_alignment>=128` TMA fast path is exercised by nothing in `tests/` or `benchmarks/` (`test_groupwise_scaled_gemm_fp8.py`).
12. All group-GEMM tests use equal-size groups; the persistent-kernel rewrite's uneven/zero-length group handling is untested (`test_groupwise_scaled_gemm_fp8.py`).

**Benchmark:**
13. One failing cuTile config aborts the whole `bench_masked_scaled_bmm` sweep (unguarded `verify_cutile`), instead of degrading to `--` like other backends (`bench_masked_scaled_bmm_backend_comparison.py:404`).

**Good news from the same probing:** every untested branch we actually executed — MIXED_PREC/mxfp4, non-uniform + zero-row segments, uneven/zero-length groups, and the TMA fast path — produced correct numerics vs independent torch references on sm100. So items 9–12 look like test additions, not kernel rework.

flashinfer/gemm/kernels/cutile/masked_scaled_bmm_cutile.py
```diff
@@ -0,0 +1,786 @@
...
+    BLOCK = (1 << max(0, (Q - 1)).bit_length()) if Q > 1 else 1
+    BLOCK = max(BLOCK, 32)  # Minimum block size for efficiency
+    BLOCK = min(BLOCK, 1024)
+
+    if Q > BLOCK:
+        # For large Q, fall back to torch.max on device (still async, no CPU sync).
+        # torch.max requires out[0].dtype to match input.dtype, so cast masked_m
+        # to int32 first if needed.
+        src = masked_m if masked_m.dtype == torch.int32 else masked_m.to(torch.int32)
+        torch.max(
```
dhiraj113: **[CRASH]** The Q > 1024 fallback in `_masked_m_max_device_cutile` permanently poisons the cached `max_m` scratch: `torch.max(..., out=)` silently resizes it from `(1,)` to 0-dim, and every subsequent call on that device crashes.

`masked_scaled_bmm(..., max_m_device=None)` with `masked_m.numel() > 1024` reaches this fallback (via masked_scaled_bmm_cutile.py:736) because `BLOCK` is capped at 1024. `torch.max`'s dim-reduction over a 1D tensor requires a 0-dim `out`, so PyTorch resizes `max_m_buf` in place from `(1,)` to `()`, emitting only a deprecation `UserWarning`. Since that buffer is the very object stored in the module-level `_max_m_cache[device]` (masked_scaled_bmm_cutile.py:51-54), the damage is twofold: the current call feeds a rank-0 `max_m` to `_masked_scaled_bmm_kernel`'s rank-1 `ct.load(max_m, index=(0,), shape=(1,))`, and every later call on that device — including small-Q calls that previously worked — reuses the resized buffer and dies in `_masked_m_max_kernel`'s `ct.store` with `TileTypeError: Index size 1 does not match the array rank 0`.

A shape-preserving reduction fixes both at once, e.g. `torch.amax(src, dim=0, keepdim=True, out=max_m_buf)` (still async, keeps the `(1,)` shape), or copy the reduction result into the cached buffer rather than passing it as `out=`.

<details>
<summary>Repro — verified on B200 (sm100), PR head 10b5ad9c</summary>

Exercises the helper directly (the exact path `masked_scaled_bmm` takes at line 736) to avoid JIT-compiling the full bmm kernel for 1025 experts.

```bash
python - <<"PY"
import torch, warnings
from flashinfer.gemm.kernels.cutile import masked_scaled_bmm_cutile as m

small = torch.tensor([3, 7, 5, 2], device="cuda", dtype=torch.int32)
r = m._masked_m_max_device_cutile(small)
print("small Q=4    -> shape", tuple(r.shape), "value", int(r.item()))  # (1,), 7: OK

big = torch.randint(1, 100, (1025,), device="cuda", dtype=torch.int32)
with warnings.catch_warnings(record=True) as w:
    warnings.simplefilter("always")
    r2 = m._masked_m_max_device_cutile(big)  # Q > 1024 -> torch.max(out=) fallback
print("big Q=1025   -> shape", tuple(r2.shape),
      "| cached scratch now", tuple(m._max_m_cache[small.device].shape))
if w:
    print("UserWarning:", str(w[0].message).splitlines()[0])

try:
    r3 = m._masked_m_max_device_cutile(small)  # identical to the first call
    print("small Q=4 again -> shape", tuple(r3.shape))
except Exception as e:
    print("small Q=4 again CRASHED:", type(e).__name__, "-", str(e).splitlines()[0][:110])
PY
```

Observed:
```
small Q=4    -> shape (1,) value 7
big Q=1025   -> shape () | cached scratch now ()
UserWarning: An output with one or more elements was resized since it had shape [1], which does not match the required output shape []. This behavior is deprecated, ...
small Q=4 again CRASHED: TileTypeError - Index size 1 does not match the array rank 0
```
</details>

flashinfer/gemm/kernels/cutile/masked_bmm_cutile.py
```diff
@@ -0,0 +1,483 @@
...
+        raise ValueError("A matrix must be contiguous")
+    if not b.is_contiguous():
+        raise ValueError("B matrix must be contiguous")
+    if not masked_m.is_contiguous():
+        raise ValueError("Masked matrix must be contiguous")
+    if masked_m.shape.numel() != Q:
+        raise ValueError(
+            f"Masked matrix must have the same shape as the number of batches ({Q})"
+        )
+    c = torch.empty((Q, M, N), device=a.device, dtype=a.dtype)
```
dhiraj113, line 414: **[CRASH]** A zero-sized dim (Q, M, or N == 0) produces a zero-sized launch grid — `masked_bmm` raises an opaque error instead of returning an empty tensor.

When any of Q/M/N is 0, `total_tiles = tiles_per_batch * Q` is 0, so `num_programs = min(NUM_SMS // num_ctas, total_tiles) * occupancy` evaluates to 0 in both the autotune path (`masked_bmm_cutile.py:329`) and the default path (`masked_bmm_cutile.py:447`), and the resulting `(0, 1, 1)` grid is rejected by the cuda.tile launcher. On the default path this surfaces as `ValueError: No valid config found in search space` (every autotune candidate fails to launch); with `FLASHINFER_CUTILE_AUTOTUNE_DISABLED=1` it is `RuntimeError: Failed to launch cuTile kernel: invalid argument`. `torch.bmm` returns empty tensors for the same shapes.

Suggested direction: early-return the already-allocated empty `c` here when `Q == 0 or M == 0 or N == 0` (mirroring the `Q == 0` guard in `masked_scaled_bmm_cutile.py`'s `_masked_m_max_device_cutile`); alternatively clamp `num_programs >= 1` — the persistent loop `range(pid, total_tiles, num_programs)` already no-ops on an empty range.

<details>
<summary>Repro — verified on B200 (sm100), PR head 10b5ad9c</summary>

```bash
for AT in 0 1; do
FLASHINFER_CUTILE_AUTOTUNE_DISABLED=$AT python - <<'PY'
import os, torch
from flashinfer.gemm import masked_bmm
mode = "autotune" if os.environ["FLASHINFER_CUTILE_AUTOTUNE_DISABLED"] == "0" else "autotune-disabled"
for name, sa, sb, q in [("Q=0", (0, 16, 64), (0, 64, 32), 0), ("M=0", (2, 0, 64), (2, 64, 32), 2)]:
    a = torch.rand(sa, device="cuda", dtype=torch.float16)
    b = torch.rand(sb, device="cuda", dtype=torch.float16)
    print(f"{name}: torch.bmm -> shape {tuple(torch.bmm(a, b).shape)}")
    try:
        c = masked_bmm(a, b, torch.zeros(q, device="cuda", dtype=torch.int32))
        torch.cuda.synchronize()
        print(f"{name}: masked_bmm [{mode}] -> OK {tuple(c.shape)}")
    except Exception as e:
        print(f"{name}: masked_bmm [{mode}] -> {type(e).__name__}: {e}")
PY
done
```

Observed:
```
Q=0: torch.bmm -> shape (0, 16, 32)
Q=0: masked_bmm [autotune] -> ValueError: No valid config found in search space.
Config: namespace(BLOCK_M=128, BLOCK_N=128, BLOCK_K=64, GROUP_SIZE_M=8, num_ctas=1, occupancy=1)
RuntimeError: cuGraphAddKernelNode: invalid argument
M=0: torch.bmm -> shape (2, 0, 32)
M=0: masked_bmm [autotune] -> ValueError: No valid config found in search space.
Q=0: masked_bmm [autotune-disabled] -> RuntimeError: Failed to launch cuTile kernel: invalid argument
M=0: masked_bmm [autotune-disabled] -> RuntimeError: Failed to launch cuTile kernel: invalid argument
```
</details>

flashinfer/gemm/kernels/cutile/ragged_scaled_bmm_cutile.py
```diff
@@ -0,0 +1,680 @@
...
+    SCALE_REP_M: ct.Constant[int],
+    SCALE_REP_K: ct.Constant[int],
+    MIXED_PREC: ct.Constant[int],
+):
+    pid = ct.bid(0)
+    if ELEM_PER_BYTE_A == 2:
+        num_k_tiles = ct.num_tiles(a, axis=1, shape=(BLOCK_M, BLOCK_K // 2))
+    else:
+        num_k_tiles = ct.num_tiles(a, axis=1, shape=(BLOCK_M, BLOCK_K))
+    num_pid_m = ct.cdiv(max_m, BLOCK_M)
```
dhiraj113: **[SILENT CORRUPTION]** An understated host `max_m` makes `ragged_scaled_bmm` silently return uninitialized `torch.empty` memory for every row beyond the first `cdiv(max_m, BLOCK_M)` M-tiles of each segment.

Unlike its siblings in this PR, this kernel has no device-side defense: the per-segment M-tile bound is `num_pid_m = ct.cdiv(max_m, BLOCK_M)` (`ragged_scaled_bmm_cutile.py:134`) computed straight from the host int, and both launch paths size the persistent grid from that same host value (`grid_fn` at `ragged_scaled_bmm_cutile.py:346`, `_ragged_scaled_bmm_default_launch` at `:427`). Contrast `ragged_bmm_cutile.py:63-64`, which reads `max_m_runtime` from a `max_m_device` tensor and sizes its grid from `total_m` (commit 41d0e831 fixed exactly this bug class there), and `ragged_block_scaled_bmm`, which derives per-expert bounds from `m_indptr`. So a stale/underestimated `max_m` hint (easy to hit when reused across MoE routing steps) schedules too few tiles, and the tail rows of longer segments ship as whatever the allocator had in the `torch.empty` output (`:608`) — no error, no warning, and neither this docstring nor `gemm_base.py`'s mentions the hazard.

Suggested direction: mirror one of the siblings — read the per-tile bound from a `max_m_device` tensor (caller-supplied, or synthesized internally e.g. via `torch.diff(m_indptr).amax()`) and size the grid from `total_m`, or derive `valid_m` per segment from `m_indptr` the way `ragged_block_scaled_bmm` now does.

<details>
<summary>Repro — verified on B200 (sm100), PR head 10b5ad9c</summary>

```bash
python - <<"PY"
import os
os.environ["FLASHINFER_CUTILE_AUTOTUNE_DISABLED"] = "1"  # deterministic single-config launch
import torch
from flashinfer.gemm import ragged_scaled_bmm

torch.manual_seed(0)
dev, Q, N, K, VEC = "cuda", 2, 128, 128, 32
total_m = 640
m_indptr = torch.tensor([0, 512, 640], dtype=torch.int32, device=dev)  # segments: 512, 128 rows
a = (torch.randn(total_m, K, device=dev) * 0.25).to(torch.float8_e4m3fn)
b = (torch.randn(Q, N, K, device=dev) * 0.25).to(torch.float8_e4m3fn)
# MX-swizzled scale tensors, all scales = 1.0 (e8m0 biased exponent 127 == 2^0)
a_scale = torch.full((total_m // 128, K // VEC // 4, 2, 256), 127, dtype=torch.uint8, device=dev)
b_scale = torch.full((Q, N // 128, K // VEC // 4, 2, 256), 127, dtype=torch.uint8, device=dev)
call = lambda mm: ragged_scaled_bmm(a, b, a_scale, b_scale, m_indptr, mm, "mxfp8", backend="cutile")

c_ok = call(512)   # control: correct host max_m (true max segment = 512)
sent = torch.full((total_m, N), 1234.5, device=dev)
del sent           # prefill the allocator block the kernel's internal torch.empty output will reuse
c_bad = call(128)  # understated host max_m: no error, no warning
torch.cuda.synchronize()

match = (c_bad == c_ok).all(dim=1)
unwritten = (c_bad == 1234.5).all(dim=1)
bad = (~match).nonzero().flatten()
print(f"rows bit-exact vs max_m=512 control: {int(match.sum())}/{total_m}")
print(f"rows never written (still hold allocator sentinel 1234.5): {int(unwritten.sum())}/{total_m}")
if bad.numel():
    print(f"corrupt rows: {bad.numel()}, range [{bad.min().item()}..{bad.max().item()}]")
PY
```

Observed:
```
rows bit-exact vs max_m=512 control: 256/640
rows never written (still hold allocator sentinel 1234.5): 384/640
corrupt rows: 384, range [128..511]
```

Rows 128..511 (the tail of the 512-row segment) are exactly the rows past the single `cdiv(128, BLOCK_M=128)` M-tile the understated hint allows per segment.
</details>

flashinfer/gemm/kernels/cutile/ragged_bmm_cutile.py
```diff
@@ -0,0 +1,868 @@
...
+        out_dtype = a.dtype
+    c = torch.empty((total_m, N), device=a.device, dtype=out_dtype)
+
+    # Materialize fallback max_m_device if the caller didn't pass one. The
+    # kernel always reads its grid bound from a device tensor (defense-in-depth),
+    # so we keep the call sites uniform.
+    # Note: this mirrors the NVT triton kernel, which always reads its grid
+    # bound from a device tensor.
+    if max_m_device is None:
+        max_m_device = torch.tensor([max_m], dtype=torch.int32, device=a.device)
```
dhiraj113: **[BUG]** With `max_m_device=None` — the default, and the path every test and the public `gemm_base.py` `ragged_bmm` caller actually hits — the documented max_m defense-in-depth is circular, so an understated host `max_m` silently leaves real output rows as uninitialized `torch.empty` garbage.

The kernel deliberately reads its per-tile loop bound from `max_m_device` (`ragged_bmm_cutile.py:63-64`) to "prevent silent output corruption when the caller passes too small a host-side max_m hint" (docstring, `ragged_bmm_cutile.py:50-54`). But when the caller omits `max_m_device`, `ragged_bmm_cutile.py:766-767` synthesizes it as `torch.tensor([max_m])` — from the exact same possibly-wrong host value — so `num_pid_m` is just as understated and rows beyond `cdiv(max_m, BLOCK_M)*BLOCK_M` in each segment are never written into the `torch.empty` output (`ragged_bmm_cutile.py:759`), with no error. This is a different gap from the earlier (addressed) review thread about `max_m==0` with an explicit `max_m_device`: here the protection never engages at all on the default path, which is the only one `tests/gemm/test_ragged_bmm_cutile.py` exercises.

Suggested direction: when `max_m_device is None`, derive it from ground truth instead of echoing the hint — e.g. `torch.diff(m_indptr).amax().to(torch.int32).reshape(1)` (stays on-device, no sync) — or, if the extra kernel launch is unwanted, amend the docstrings (`ragged_bmm_cutile.py:50-54, 717-721`) so an accurate `max_m` is documented as a hard correctness requirement rather than a hint with a safety net.

<details>
<summary>Repro — verified on B200 (sm100), PR head 10b5ad9c</summary>

```bash
python - <<"PY"
import torch
from flashinfer.gemm import ragged_bmm

torch.manual_seed(0)
Q, SEG, K, N = 2, 512, 64, 128  # m_indptr=[0,512,1024]; true max segment = 512
a = torch.randn(Q * SEG, K, device="cuda", dtype=torch.bfloat16)
b = torch.randn(Q, N, K, device="cuda", dtype=torch.bfloat16)
m_indptr = torch.tensor([0, SEG, 2 * SEG], device="cuda", dtype=torch.int32)
ref = torch.cat([a[q*SEG:(q+1)*SEG].float() @ b[q].float().T for q in range(Q)])

c_ok = ragged_bmm(a, b, m_indptr, max_m=SEG)  # correct hint, max_m_device=None
print("correct max_m=512   : max abs err =", (c_ok.float() - ref).abs().max().item())

# Prefill the allocator block that the output's torch.empty will recycle.
sent = torch.full((Q * SEG, N), float("nan"), device="cuda", dtype=torch.bfloat16)
del sent
torch.cuda.synchronize()

c_bad = ragged_bmm(a, b, m_indptr, max_m=64)  # understated hint, max_m_device=None
torch.cuda.synchronize()
nan_rows = int(torch.isnan(c_bad).any(dim=1).sum())
print(f"understated max_m=64: {nan_rows}/{Q*SEG} output rows still hold the NaN "
      f"sentinel (never written), no error raised")
PY
```

Observed (autotune sweep tables elided):
```
correct max_m=512   : max abs err = 0.12384033203125
understated max_m=64: 512/1024 output rows still hold the NaN sentinel (never written), no error raised
```
</details>

flashinfer/gemm/gemm_base.py
```diff
@@ -962,6 +962,554 @@ def bmm_bf16(
...
+    max_m: int,
+    max_m_device: Optional[torch.Tensor] = None,
+    transpose_a: bool = False,
+    transpose_b: bool = True,
+    out_dtype: Optional[torch.dtype] = None,
+    segment_alignment: int = 128,
+    backend: Literal["cutile"] = "cutile",
+):
+    """Validate shapes, dtypes and backend support for the cuTile ragged_block_scaled_bmm path."""
+    return True
```
dhiraj113, line 1334: **[BUG]** `_cutile_ragged_block_scaled_bmm_requirement` validates nothing — its docstring says "Validate shapes, dtypes and backend support" but the body is just `return True`.

Every other `_cutile_*_requirement` gate added in this PR (`masked_bmm`, `gemm_alpha_beta`, `ragged_bmm`, `masked_scaled_bmm`, `ragged_scaled_bmm`) rejects at least the wrong-dtype case with a clean `ValueError` at the API boundary. Because this one checks nothing, `ragged_block_scaled_bmm` with plain bf16 `a`/`b` runs end-to-end through a dtype combination the API neither documents nor tests, while an fp8/bf16 mismatch escapes the gate and surfaces as an opaque `cuda.tile` `TileTypeError` from `ct.mma` deep inside the JIT'd kernel (`ragged_block_scaled_bmm_cutile.py:145`) instead of a `ValueError`. Suggest validating `a`/`b` dtypes and the `a_scale`/`b_scale` shapes/dtypes here, mirroring the sibling gates (e.g. `_cutile_masked_bmm_requirement`).

<details>
<summary>Repro — verified on B200 (sm100), PR head 10b5ad9c</summary>

```bash
python - <<"PY"
import torch
from flashinfer.gemm import masked_bmm, ragged_block_scaled_bmm

dev = "cuda"
Q, M, N, K = 1, 128, 128, 128
m_indptr = torch.tensor([0, M], dtype=torch.int32, device=dev)
a_scale = torch.ones(M, 1, dtype=torch.float32, device=dev)
b_scale = torch.ones(Q, 1, 1, dtype=torch.float32, device=dev)
b_bf16 = torch.randn(Q, N, K, device=dev, dtype=torch.bfloat16)

# Contrast: sibling gate (_cutile_masked_bmm_requirement) rejects a bad dtype cleanly
try:
    masked_bmm(torch.randn(1, 8, 8, device=dev), torch.randn(1, 8, 8, device=dev),
               torch.full((1,), 8, dtype=torch.int32, device=dev))
except ValueError as e:
    print(f"masked_bmm  fp32/fp32 : clean ValueError: {e}")

for a, label in [
    (torch.randn(M, K, device=dev, dtype=torch.bfloat16), "bf16/bf16"),
    (torch.randn(M, K, device=dev).to(torch.float8_e4m3fn), "fp8 /bf16"),
]:
    try:
        out = ragged_block_scaled_bmm(a, b_bf16, a_scale, b_scale, m_indptr, M)
        torch.cuda.synchronize()
        err = (out.float() - a.float() @ b_bf16[0].float().T).abs().max()
        print(f"ragged_block_scaled_bmm {label}: SILENTLY ACCEPTED; "
              f"out.dtype={out.dtype}, max|out - a@b.T|={err:.3g}")
    except ValueError as e:
        print(f"ragged_block_scaled_bmm {label}: clean ValueError: {e}")
    except Exception as e:
        print(f"ragged_block_scaled_bmm {label}: {type(e).__module__}.{type(e).__name__}:")
        for ln in str(e).splitlines()[:3]:
            print(f"    {ln}")
PY
```

Observed:
```
masked_bmm  fp32/fp32 : clean ValueError: The masked_bmm cuTile backend supports float16 / bfloat16 inputs only; got torch.float32.
ragged_block_scaled_bmm bf16/bf16: SILENTLY ACCEPTED; out.dtype=torch.bfloat16, max|out - a@b.T|=0.125
ragged_block_scaled_bmm fp8 /bf16: cuda.tile._exception.TileTypeError:
    x and y must have the same dtype unless they are int8/uint8, got float8_e4m3fn bfloat16
      ".../flashinfer/gemm/kernels/cutile/ragged_block_scaled_bmm_cutile.py", line 145, col 25-63, in _ragged_block_scaled_bmm_kernel:
                        c_mma = ct.mma(a_block, b_block, acc=mma_zeros)
```
</details>

flashinfer/gemm/kernels/cutile/ragged_block_scaled_bmm_cutile.py
```diff
@@ -0,0 +1,434 @@
...
+    # [Q, rnb, rkb], hence the N granularity is N // rnb. A mismatch (e.g. a
+    # 256-wide N granularity with BLOCK_K=128) would apply one column group's
+    # scale to another and silently produce wrong numbers, so reject it.
+    n_scale_granularity = N // rnb
+    if n_scale_granularity != BLOCK_K:
+        raise ValueError(
+            f"b_scale N granularity ({n_scale_granularity} = N {N} // {rnb}) must equal "
+            f"BLOCK_K ({BLOCK_K}); this kernel indexes b_scale on N by BLOCK_K. Use a "
+            "b_scale whose N and K granularities match, or the gather kernel."
+        )
```
dhiraj113, line 395: **[BUG]** `a_scale`'s K-granularity is never validated — a mismatched `a_scale` is silently accepted and produces wrong numerics, while the equivalent `b_scale` mismatch raises a clean ValueError.

The validation at `ragged_block_scaled_bmm_cutile.py:309` destructures `total_ma, rka = a_scale.shape` but only asserts `total_ma == total_m`; `rka` is never compared against `cdiv(K, BLOCK_K)`. The kernel reads `a_scale` at `index=(m_tile_start + pid_m, k)` for every K tile with `padding_mode=ct.PaddingMode.ZERO` (lines 150–155), so an `a_scale` with fewer K tiles than expected (e.g. quantized with 256-wide K blocks while `BLOCK_K=128`) has its out-of-range tiles scaled by 0, and those K tiles' contributions silently vanish; a too-wide `a_scale` (e.g. 64-wide K blocks) is likewise accepted and applies wrong-granularity scales. This is exactly the failure class the b_scale N-granularity guard at lines 368–379 was hardened against earlier in this PR — `a_scale`'s K axis never got the counterpart check, and the API-level `_cutile_ragged_block_scaled_bmm_requirement` in `gemm_base.py` doesn't catch it either.

Suggested direction: next to the b_scale guard, validate that `a_scale`'s K granularity equals `BLOCK_K` (i.e. `rka == cdiv(K, BLOCK_K)`) and raise the same style of ValueError.

<details>
<summary>Repro — verified on B200 (sm100), PR head 10b5ad9c</summary>

```bash
python - <<"PY"
import torch
from flashinfer.gemm import ragged_block_scaled_bmm

torch.manual_seed(0)
dev, fp8 = "cuda", torch.float8_e4m3fn
Q, seg_m, N, K, BK = 2, 128, 256, 512, 128  # BLOCK_K = VEC_SIZE = K // b_scale.shape[2] = 128
total_m, k_tiles = Q * seg_m, K // BK
m_indptr = torch.tensor([0, seg_m, total_m], dtype=torch.int32, device=dev)
mx = torch.finfo(fp8).max
a = (torch.randn(total_m, K, device=dev) * 0.3 * mx).clamp(-mx, mx).to(fp8)
b = (torch.randn(Q, N, K, device=dev) * 0.3 * mx).clamp(-mx, mx).to(fp8)
a_scale = torch.rand(total_m, k_tiles, device=dev) * 1e-2  # [256, 4]: 128-wide K blocks
b_scale = torch.rand(Q, N // BK, k_tiles, device=dev) * 1e-2

def run(a_s):
    return ragged_block_scaled_bmm(a, b, a_s, b_scale, m_indptr, seg_m,
                                   out_dtype=torch.bfloat16, backend="cutile").float()

def ref(a_s_elem):  # torch dequant reference from a per-element [total_m, K] a-scale
    c = torch.zeros(total_m, N, device=dev)
    for q in range(Q):
        s, e = m_indptr[q].item(), m_indptr[q + 1].item()
        bs = b_scale[q].repeat_interleave(BK, 0).repeat_interleave(BK, 1)
        c[s:e] = (a.float()[s:e] * a_s_elem[s:e]) @ (b.float()[q] * bs).t()
    return c

relerr = lambda o, r: ((o - r).abs().max() / r.abs().max()).item()
print(f"correct a_scale [256,4]:    rel err {relerr(run(a_scale), ref(a_scale.repeat_interleave(BK, 1))):.4f}")
half = a_scale[:, :2].contiguous()  # caller quantized A with 256-wide K blocks -> [256, 2]
out = run(half)  # accepted with no error
print(f"too-narrow a_scale [256,2]: rel err {relerr(out, ref(half.repeat_interleave(2 * BK, 1))):.4f} vs intended 256-wide dequant")
print(f"                            rel err {relerr(out, ref(torch.cat([half, torch.zeros_like(half)], 1).repeat_interleave(BK, 1))):.4f} vs 'K tiles 2..3 scaled by ZERO'")
try:  # contrast: the equivalent b_scale N-granularity mismatch is cleanly rejected
    ragged_block_scaled_bmm(a, b, a_scale, b_scale[:, :1].contiguous(), m_indptr, seg_m)
except ValueError as e:
    print(f"b_scale N-granularity mismatch -> ValueError: {e}")
PY
```

Observed:
```
correct a_scale [256,4]:    rel err 0.0025
too-narrow a_scale [256,2]: rel err 0.7778 vs intended 256-wide dequant
                            rel err 0.0030 vs 'K tiles 2..3 scaled by ZERO'
b_scale N-granularity mismatch -> ValueError: b_scale N granularity (256 = N 256 // 1) must equal BLOCK_K (128); this kernel indexes b_scale on N by BLOCK_K. Use a b_scale whose N and K granularities match, or the gather kernel.
```

The 0.0030 match against the "missing K tiles scaled by zero" reference confirms the mechanism: the out-of-range `a_scale` reads are zero-padded, not rejected.
</details>

flashinfer/gemm/kernels/cutile/masked_scaled_bmm_cutile.py
```diff
@@ -0,0 +1,786 @@
...
+
+    # NO host-side unswizzle — scales are passed raw 5D to the kernel.
+    # The kernel does in-register reshape/permute/reshape (same as Triton).
+    # This avoids extra memory copies that were a major source of overhead.
+
+    # Compute max(masked_m) on GPU — avoids GPU->CPU sync.
+    # The kernel uses max_m to dynamically reduce tile scheduling.
+    if max_m_device is not None:
+        if isinstance(max_m_device, torch.Tensor) and max_m_device.is_cuda:
+            max_m = max_m_device.to(torch.int32).reshape(1)
```
dhiraj113: **[BUG]** `max_m_device` is never validated against `a.device`, unlike the sibling `ragged_block_scaled_bmm`, which rejects exactly this input.

When `max_m_device` is a CUDA tensor, `masked_scaled_bmm_cutile.py:725-726` forwards it as `max_m_device.to(torch.int32).reshape(1)` — `.to(dtype)` preserves the device, so a tensor resident on a different GPU than `a` (e.g. precomputed once and reused across per-device calls in a multi-GPU MoE setup) flows into the kernel launched on `torch.cuda.current_stream(a.device)` and is dereferenced cross-GPU via `ct.load(max_m, ...)` — typically `cudaErrorIllegalAddress`, or silently wrong reads if P2P happens to be enabled (this case is from code reading; the repro box has a single GPU). `ragged_block_scaled_bmm_cutile.py:341-344` already guards its own `max_m_device` with a clean `ValueError`, but the guard was not carried over to this file. Relatedly, the else-branch at `:727-734` silently accepts a CPU tensor via `int(max_m_device.item())` plus a fresh H2D `torch.tensor(...)` per call — a host round-trip that contradicts the `:722` comment ("avoids GPU->CPU sync") and is unsafe under CUDA-graph capture.

Suggested direction: mirror the sibling's `max_m_device.device != a.device` → `ValueError` check, and either reject the silent CPU-tensor fallback or document its host-sync cost.

<details>
<summary>Repro — verified on B200 (sm100), PR head 10b5ad9c</summary>

```bash
FLASHINFER_CUTILE_AUTOTUNE_DISABLED=1 python - <<"PY"
import torch
from flashinfer.gemm import masked_scaled_bmm
from flashinfer.gemm.kernels.cutile.ragged_block_scaled_bmm_cutile import (
    ragged_block_scaled_bmm,
)

dev = torch.device("cuda")
Q, M, N, K = 1, 128, 128, 128
a = torch.randn(Q, M, K, device=dev).to(torch.float8_e4m3fn)
b = torch.randn(Q, N, K, device=dev).to(torch.float8_e4m3fn)
a_scale = torch.full((Q, 1, 1, 2, 256), 127, dtype=torch.uint8, device=dev)
b_scale = torch.full((Q, 1, 1, 2, 256), 127, dtype=torch.uint8, device=dev)
masked_m = torch.tensor([M], dtype=torch.int32, device=dev)
off_dev = torch.tensor([M], dtype=torch.int32)  # NOT on a.device

out = masked_scaled_bmm(a, b, a_scale, b_scale, masked_m, "mxfp8",
                        max_m_device=off_dev)
torch.cuda.synchronize()
print(f"masked_scaled_bmm: off-device max_m_device ({off_dev.device}) accepted "
      f"silently, out.shape={tuple(out.shape)}")

# Sibling op rejects the exact same tensor before any launch:
ra = torch.randn(M, K, device=dev).to(torch.float8_e4m3fn)
rb = torch.randn(1, N, K, device=dev).to(torch.float8_e4m3fn)
ra_s = torch.full((M, 4), 127, dtype=torch.uint8, device=dev).view(torch.float8_e8m0fnu)
rb_s = torch.full((1, N, 4), 127, dtype=torch.uint8, device=dev).view(torch.float8_e8m0fnu)
m_indptr = torch.tensor([0, M], dtype=torch.int32, device=dev)
try:
    ragged_block_scaled_bmm(ra, rb, ra_s, rb_s, m_indptr, M, max_m_device=off_dev)
except ValueError as e:
    print(f"ragged_block_scaled_bmm: rejected -> ValueError: {e}")
PY
```

Observed:
```
masked_scaled_bmm: off-device max_m_device (cpu) accepted silently, out.shape=(1, 128, 128)
ragged_block_scaled_bmm: rejected -> ValueError: max_m_device must be on cuda:0; got cpu
```
</details>

tests/gemm/test_masked_bmm_cutile.py
```diff
@@ -0,0 +1,162 @@
...
+        [
+            (
+                case["num_groups"],
+                case["max_m"],
+                case["expected_m_per_group"],
+                case["n"],
+                case["k"],
+            )
+            for case in list(enumerate_m_grouped_masked())[
+                :2
```
dhiraj113: **[BUG]** The `[:2]` slice collapses this 106-entry shape sweep to a single `(num_groups=6, m_per_group=512)` case tested twice — the DeepGEMM defaults and the entire GB200 sweep never run.

`enumerate_m_grouped_masked()` iterates the two `(n, k)` shapes in its inner loop, so the flattened list interleaves them per case: `[case0/nk0, case0/nk1, case1/nk0, ...]`. The slice at `test_masked_bmm_cutile.py:120-122` therefore keeps case 0 twice (differing only in `n`/`k`): of 106 generated entries covering 51 distinct `(num_groups, expected_m_per_group)` cases, only `(6, 512)` is ever exercised, while the explicitly labeled "DeepGEMM default cases" `(1, 1024)`, `(2, 512)`, `(4, 256)` and all 98 GB200 `num_ranks x num_tokens` entries are dead code. Since the comment says "Use smaller set for correctness", the intent was presumably a representative sample — consider truncating `cases` before the `(n, k)` product (or picking a handful of entries spanning different `num_groups`) rather than slicing the interleaved flat list.

<details>
<summary>Repro — verified at PR head 10b5ad9c (static enumeration, no GPU needed)</summary>

Run from the repo root:

```bash
python - <<"PY"
import sys

sys.path.insert(0, "tests/gemm")
from test_masked_bmm_cutile import enumerate_m_grouped_masked

full = list(enumerate_m_grouped_masked())
distinct = {(c["num_groups"], c["expected_m_per_group"]) for c in full}
kept = full[:2]  # what the parametrize actually keeps
print("generated entries              :", len(full))
print("distinct (groups, m/grp) cases :", len(distinct))
print("kept by [:2]                   :",
      [(c["num_groups"], c["expected_m_per_group"], c["n"], c["k"]) for c in kept])
print("distinct cases actually tested :",
      sorted({(c["num_groups"], c["expected_m_per_group"]) for c in kept}))
PY
```

Observed:
```
generated entries              : 106
distinct (groups, m/grp) cases : 51
kept by [:2]                   : [(6, 512, 4096, 7168), (6, 512, 7168, 2048)]
distinct cases actually tested : [(6, 512)]
```
</details>

tests/gemm/test_masked_scaled_bmm_cutile.py
```diff
@@ -0,0 +1,223 @@
...
+        return a, b, a_scale, b_scale, m_mask, reference
+
+    @pytest.mark.parametrize(
+        "num_groups, max_m, expected_m_per_group, n, k",
+        [
+            (2, 512, 64, 256, 256),
+            (4, 512, 128, 256, 256),
+        ],
+    )
+    @pytest.mark.parametrize("block_scale_type", ["mxfp8", "nvfp4"])
```
dhiraj113: **[TEST GAP]** `block_scale_type="mixed"` and `"mxfp4"` are accepted by the API but never executed by any test, so the MIXED_PREC fp8×fp4 kernel branch ships with zero numeric coverage.

`_cutile_masked_scaled_bmm_requirement` in gemm_base.py accepts all four values `{"nvfp4", "mxfp4", "mxfp8", "mixed"}`, and `"mixed"` selects a distinct kernel branch (`MIXED_PREC == 1`, masked_scaled_bmm_cutile.py:165-195, selector at masked_scaled_bmm_cutile.py:739) containing the fp8→fp4 downcast workaround the code itself flags as "may lose precision vs Triton"; `"mxfp4"` independently exercises the otherwise-untested fp4×fp4 branch. Since this parametrize only covers `["mxfp8", "nvfp4"]`, a regression in either branch (wrong cast, wrong `a_block`/`b_block` shape) would not be caught by CI. Good news: I ran both uncovered values on B200 against this file's own dequantized `torch.bmm` reference and they currently pass exactly (max_abs_err=0), so this is purely a test addition, not a kernel fix.

Suggested direction: extend the parametrize to all four values — `initialize_block_scaled` already supports them — and widen the sm103 `ct.unpack_from_bytes` skip at line 188 to also cover `"mixed"`/`"mxfp4"`, since those paths unpack fp4 bytes too.

<details>
<summary>Repro — verified on B200 (sm100), PR head 10b5ad9c</summary>

```bash
# 1) No collected test ever exercises 'mixed' or 'mxfp4':
pytest tests/gemm/test_masked_scaled_bmm_cutile.py --collect-only -q | grep -cE 'mixed|mxfp4'   # -> 0

# 2) Probe the two uncovered values against the test's own dequant reference:
python - <<"PY"
import os, random, sys, torch
os.environ["FLASHINFER_CUTILE_AUTOTUNE_DISABLED"] = "1"  # single-config launch; same kernel body
sys.path.insert(0, "tests/gemm")
from test_masked_scaled_bmm_cutile import Test_FlashInfer_MaskedScaledBMM
from flashinfer.gemm import masked_scaled_bmm

for bst in ["mxfp8", "mixed", "mxfp4"]:  # mxfp8 = CI-covered control
    torch.manual_seed(0); random.seed(0)
    a, b, a_scale, b_scale, m_mask, ref = (
        Test_FlashInfer_MaskedScaledBMM.initialize_block_scaled(
            2, 256, 64, 256, 256, block_scale_type=bst,
            compute_reference=True, out_dtype=torch.bfloat16))
    c = masked_scaled_bmm(a, b, a_scale, b_scale, m_mask, bst,
                          out_dtype=torch.bfloat16, backend="cutile")
    for i in range(2):
        c[i, m_mask[i]:, :] = 0
    err = (c.float() - ref.float()).abs().max().item()
    print(f"{bst:6s} a={a.dtype} b={b.dtype}: max_abs_err={err:.4g}")
PY
```

Observed:
```
0
mxfp8  a=torch.float8_e4m3fn b=torch.float8_e4m3fn: max_abs_err=0
mixed  a=torch.float8_e4m3fn b=torch.uint8: max_abs_err=0
mxfp4  a=torch.uint8 b=torch.uint8: max_abs_err=0
```
</details>

tests/gemm/test_ragged_block_scaled_bmm_cutile.py
```diff
@@ -0,0 +1,329 @@
...
+            max_m,
+        ) = self.prepare_data(
+            num_groups,
+            m,
+            n,
+            k,
+            trans_a,
+            trans_b,
+            out_dtype,
+            use_aligned_segments=True,
```
dhiraj113: **[TEST GAP]** The only `ragged_block_scaled_bmm` test builds perfectly uniform segments, so the ragged/imbalanced/zero-row cases the prefix-sum schedule was built for are never exercised.

`test_op` always calls `prepare_data(..., use_aligned_segments=True)` (test_ragged_block_scaled_bmm_cutile.py:309), which routes to `create_aligned_m_segments` — all `num_groups=4` segments get the identical size `aligned_m`. The ragged helper `create_ragged_m_segments` (line 19) is never called anywhere in the file, and `reference()` would even `assert segment_size > 0` (line 134) if a zero-row segment ever reached it. That leaves the per-expert prefix-sum scheduling (commit fe24edd7, whose own motivation is imbalanced shapes like `[1024, 7x128]` and varying MoE group sizes) — and the "matches native trtllm" claim for those shapes — with zero CI coverage.

Good news: I probed non-uniform 128-aligned segments including a zero-row segment against a torch dequant reference on B200 and the kernel matches (~2e-3 rel err), so this is a test addition, not a kernel fix. Suggest parametrizing `test_op` over a couple of hand-picked non-uniform `m_indptr` layouts (including an empty segment) and making `reference()` skip zero-row segments instead of asserting.

<details>
<summary>Repro — verified on B200 (sm100), PR head 10b5ad9c</summary>

```bash
python - <<"PY"
import torch
from flashinfer.gemm import ragged_block_scaled_bmm

torch.manual_seed(0)
DEV, N, K, BLK = torch.device("cuda"), 256, 256, 128
# Non-uniform 128-aligned segments incl. a zero-row one: sizes [256, 0, 128, 512].
# CI's only test builds num_groups identical segments (use_aligned_segments=True).
m_indptr = torch.tensor([0, 256, 256, 384, 896], dtype=torch.int32, device=DEV)
total_m, Q = 896, 4
fmax = torch.finfo(torch.float8_e4m3fn).max
a = ((torch.rand(total_m, K, device=DEV) - 0.5) * 2 * fmax).to(torch.float8_e4m3fn)
b = ((torch.rand(Q, N, K, device=DEV) - 0.5) * 2 * fmax).to(torch.float8_e4m3fn)
a_s = torch.rand(total_m, K // BLK, device=DEV) * 1e-2
b_s = torch.rand(Q, N // BLK, K // BLK, device=DEV) * 1e-2
c = ragged_block_scaled_bmm(a, b, a_s, b_s, m_indptr, max_m=512,
                            out_dtype=torch.bfloat16, backend="cutile")
ref = torch.zeros(total_m, N, device=DEV)
a_f = a.float() * a_s.repeat_interleave(BLK, dim=1)
for q in range(Q):
    s, e = int(m_indptr[q]), int(m_indptr[q + 1])
    if e == s:  # zero-row segment: the test file's reference() would assert here
        print(f"segment {q} rows [{s},{e}): zero-row")
        continue
    b_f = b[q].float() * b_s[q].repeat_interleave(BLK, 0).repeat_interleave(BLK, 1)
    ref[s:e] = a_f[s:e] @ b_f.t()
    err = (c[s:e].float() - ref[s:e]).abs().max().item()
    print(f"segment {q} rows [{s},{e}): max abs err {err:.4f} "
          f"(rel {err / ref[s:e].abs().max().item():.1e})")
torch.testing.assert_close(ref.to(torch.bfloat16), c, atol=1.0, rtol=2e-2)
print("OK: non-uniform + zero-row segments match the torch reference")
PY
```

Observed:
```
segment 0 rows [0,256): max abs err 0.4994 (rel 2.4e-03)
segment 1 rows [256,256): zero-row
segment 2 rows [256,384): max abs err 0.4829 (rel 2.7e-03)
segment 3 rows [384,896): max abs err 0.4999 (rel 1.7e-03)
OK: non-uniform + zero-row segments match the torch reference
```

The kernel is correct on these shapes today — the point is that no test in the file can catch a regression here, since only uniform segments are ever constructed.
</details>

tests/gemm/test_groupwise_scaled_gemm_fp8.py
```diff
@@ -201,6 +201,7 @@ def test_fp8_groupwise_gemm_small_batch_size(m, n, k, scale_major_mode):
     torch.testing.assert_close(c, ref_c, atol=1e-2, rtol=1e-2)
 
 
+@pytest.mark.parametrize("backend", ["trtllm", "cutile"])
```
dhiraj113, line 204: **[TEST GAP]** The new `segment_alignment >= 128` TMA fast path of the cutile group GEMM is exercised by zero tests — every test run takes the default gather path.

This test is the only caller of `group_gemm_fp8_nt_groupwise(..., backend="cutile")` in the suite and never passes `segment_alignment`, so the dispatch gate at `gemm_fp8_nt_groupwise_cutile.py:870-893` (which forwards `max_m=total_m`, `max_m_device`, `out`, transpose flags into `ragged_block_scaled_bmm`) is always False under pytest; `grep -rn segment_alignment tests/ benchmarks/` returns nothing. That standalone kernel has its own test file, but this integration wiring does not — and the parameter's own docstring (gemm_base.py:8382) calls it a caller contract where "a wrong value silently corrupts output", exactly the kind of path that wants a regression test. Good news: driving the branch through the public API with 128-aligned ragged segments works and matches a torch reference, so this is coverage-only today.

Suggested direction: parametrize this test (or add a small dedicated case) with `segment_alignment=128` and a 128-aligned ragged `m_indptr`, asserting the result against the existing reference.

<details>
<summary>Repro — verified on B200 (sm100), PR head 10b5ad9c</summary>

```bash
# 1) No test or benchmark ever passes segment_alignment (the TMA-path opt-in):
grep -rn segment_alignment tests/ benchmarks/ | wc -l    # -> 0

# 2) The path is live and reachable from the public API on sm100+:
python - <<"PY"
import torch
import flashinfer.gemm.kernels.cutile.ragged_block_scaled_bmm_cutile as rbsb
from flashinfer.gemm import group_gemm_fp8_nt_groupwise

calls, orig = [], rbsb.ragged_block_scaled_bmm
def spy(*a, **kw):
    calls.append(kw.get("segment_alignment"))
    return orig(*a, **kw)
rbsb.ragged_block_scaled_bmm = spy  # the dispatch imports this at call time

torch.manual_seed(0)
t, k, n, G, cum_m = 128, 256, 256, 2, 384
m_indptr = torch.tensor([0, 128, 384], dtype=torch.int32, device="cuda")  # ragged, 128-aligned
a = (torch.randn(cum_m, k, device="cuda") * 0.5).to(torch.float8_e4m3fn)
b = (torch.randn(G, n, k, device="cuda") * 0.5).to(torch.float8_e4m3fn)
a_scale = (torch.rand(cum_m, k // t, device="cuda") * 1.5 + 0.5).float()
b_scale = (torch.rand(G, n // t, k // t, device="cuda") * 1.5 + 0.5).float()

out = group_gemm_fp8_nt_groupwise(a, b, a_scale, b_scale, m_indptr,
    scale_major_mode="K", out_dtype=torch.bfloat16, backend="cutile",
    segment_alignment=128)  # <-- the argument no test ever passes

print("TMA branch taken:", calls == [128])
a_deq = a.float() * a_scale.repeat_interleave(t, 1)
b_deq = b.float() * b_scale.repeat_interleave(t, 1).repeat_interleave(t, 2)
ref = torch.cat([a_deq[s:e] @ b_deq[g].T for g, (s, e) in
    enumerate(zip(m_indptr[:-1].tolist(), m_indptr[1:].tolist()))]).to(torch.bfloat16)
torch.testing.assert_close(out, ref, atol=1e-2, rtol=1e-2)
print("matches torch reference, max_abs_err =", (out.float() - ref.float()).abs().max().item())
PY
```

Observed:
```
0
TMA branch taken: True
matches torch reference, max_abs_err = 0.03125
```
(`0` is the grep hit count; the assert_close passed — the 0.03 max abs error is within bf16 rtol on O(10) magnitudes.)
</details>

tests/gemm/test_groupwise_scaled_gemm_fp8.py
```diff
@@ -259,6 +277,7 @@ def test_fp8_groupwise_group_gemm(
         m_indptr,
         scale_major_mode=scale_major_mode,
         out_dtype=out_dtype,
+        backend=backend,
```
dhiraj113, line 280: **[TEST-GAP]** Every grouped-GEMM case in this suite builds perfectly uniform groups, so the persistent kernel's uneven/zero-length group handling is never exercised.

`m_indptr = torch.arange(0, group_size + 1, ...) * m` (test_groupwise_scaled_gemm_fp8.py:270) always yields `group_size` identical segments of length `m`, and the only other indptr construction in the file (`test_fp8_groupwise_group_deepgemm`, lines 310-328) is equally uniform. Yet the rewritten `_w8a8_group_fp8_matmul_kernel` (gemm_fp8_nt_groupwise_cutile.py) exists precisely for non-uniform MoE token counts: per-group `m_start`/`m_end` device loads (lines 661-663), the OOB-sentinel drop for tail rows spilling past a short group (line 731), and the `avg_m`-based config split (line 745) — none of which uniform groups can stress, so a regression in tile-to-group assignment for skewed or empty groups would pass this suite unnoticed.

Suggested direction: add a case (or parametrize `m_indptr`) with skewed sizes including a zero-length group, e.g. `[200, 0, 56, 192]` (avg_m < 256, swap_ab config) and `[512, 0, 4, 508]` (avg_m >= 256). Good news: both already match a torch reference exactly, so the strengthened test would be green on arrival.

<details>
<summary>Repro — verified on B200 (sm100), PR head 10b5ad9c</summary>

```bash
python - <<"PY"
import math, torch
from flashinfer.gemm import group_gemm_fp8_nt_groupwise
from flashinfer.testing.utils import dequantize_fp8, quantize_fp8

torch.random.manual_seed(0)
T = 128

def run(name, sizes, n=128, k=256):
    Q, total_m = len(sizes), sum(sizes)
    m_indptr = torch.cumsum(torch.tensor([0] + sizes), 0).to(torch.int32).cuda()
    a = torch.randn((total_m, k), dtype=torch.float, device="cuda")
    b = torch.randn((Q, n, k), dtype=torch.float, device="cuda") / math.sqrt(k)
    a_fp8, a_s = quantize_fp8(a, (total_m, k // T), (1, T), "K")
    b_fp8, b_s = quantize_fp8(b, (Q, n // T, k // T), (1, T, T), "K")
    a_dq, b_dq = dequantize_fp8(a_fp8, a_s, "K"), dequantize_fp8(b_fp8, b_s, "K")
    out = group_gemm_fp8_nt_groupwise(a_fp8, b_fp8, a_s, b_s, m_indptr,
        scale_major_mode="K", out_dtype=torch.bfloat16, backend="cutile")
    ref = torch.cat([a_dq[int(m_indptr[q]):int(m_indptr[q+1])] @ b_dq[q].T
                     for q in range(Q)]).to(torch.bfloat16)
    err = (out.float() - ref.float()).abs().max().item()
    print(f"{name}: sizes={sizes} max_abs_err={err:.4f}")

run("uneven+empty, avg_m<256 (swap_ab cfg) ", [200, 0, 56, 192])
run("uneven+empty, avg_m>=256 (BLOCK_M=128)", [512, 0, 4, 508])
PY
```

Observed (the untested shapes pass — confirming this is purely a coverage gap today):
```
uneven+empty, avg_m<256 (swap_ab cfg) : sizes=[200, 0, 56, 192] max_abs_err=0.0000
uneven+empty, avg_m>=256 (BLOCK_M=128): sizes=[512, 0, 4, 508] max_abs_err=0.0000
```
</details>

benchmarks/bench_masked_scaled_bmm_backend_comparison.py
```diff
@@ -0,0 +1,479 @@
...
+    ]
+    dtypes = [d.strip() for d in args.dtype.split(",") if d.strip()]
+    workloads = default_workloads()
+    out_dtype = torch.bfloat16
+
+    all_results = {}
+    for bst in dtypes:
+        print(f"\n{'#' * 78}\n# block_scale_type = {bst}\n{'#' * 78}")
+        print("Correctness (cuTile vs dequant torch.bmm, 2x512x256x256):")
+        ok, cos = verify_cutile(bst, out_dtype)
```
dhiraj113: **[BUG]** One failing cuTile config aborts the entire benchmark sweep, including valid configs queued after it.

`verify_cutile()` is called bare here, and inside it the `masked_scaled_bmm(...)` call at `bench_masked_scaled_bmm_backend_comparison.py:336` is unguarded — the `try/except` at lines 351-358 only wraps `torch.testing.assert_close`. So any exception from the cuTile backend (bad `block_scale_type`, unsupported arch, transient JIT/compile error) kills the whole run with a traceback before a single timing row is printed, whereas `bench_one()` (lines 247-270) correctly catches per-provider failures and degrades to `--`. The same pattern exists in `bench_ragged_scaled_bmm_backend_comparison.py` (bare `verify_cutile` at line 346, unguarded `ragged_scaled_bmm` at line 285).

Suggested direction: guard the kernel invocation in `verify_cutile` (or the call site) the same way `bench_one` does, reporting FAIL/N-A for that dtype and letting the remaining dtypes proceed.

<details>
<summary>Repro — verified on B200 (sm100), PR head 10b5ad9c</summary>

An invalid dtype is the cheapest deterministic trigger; any runtime cuTile failure behaves the same. The valid `mxfp8` sweep queued second never runs (exit code 1):

```bash
python benchmarks/bench_masked_scaled_bmm_backend_comparison.py --providers cutile --dtype bogus,mxfp8
```

Observed:
```
GPU Compute Capability: SM100
##############################################################################
# block_scale_type = bogus
##############################################################################
Correctness (cuTile vs dequant torch.bmm, 2x512x256x256):
  File ".../bench_masked_scaled_bmm_backend_comparison.py", line 404, in main
    ok, cos = verify_cutile(bst, out_dtype)
  File ".../bench_masked_scaled_bmm_backend_comparison.py", line 336, in verify_cutile
    c = masked_scaled_bmm(
ValueError: masked_scaled_bmm supports block_scale_type in ('nvfp4', 'mxfp4', 'mxfp8', 'mixed'); got 'bogus'.
```
</details>

yifeis-nv:
Hi @dhiraj113, thanks for your reviewing. Pushed fixes for all 13 findings (up to `74a20c314`), validated on B200 (sm100).

1. masked_scaled_bmm max_m fallback no longer writes the cached scratch — `24c901ac7`
2. zero-sized masked_bmm returns an empty output instead of building an invalid grid — `4e59429c8`
3. ragged_scaled_bmm loop bound now comes from device max_m, not the host hint — `8180970bc`
4. ragged_bmm max_m_device fallback now amax's over m_indptr diffs instead of the host hint — `85a66ee17`
5. ragged_block_scaled_bmm requirement validates NT layout / FP8 dtype / fp32 scales instead of `return True` — `edae42400`
6. a_scale K granularity checked (`K_A // rka == BLOCK_K`) — `46da2d17e`
7. masked_scaled_bmm requires a CUDA max_m_device on `a.device`; dropped the CPU/`.item()` path so it stays CUDA-graph safe — `e2a11c697`
8. masked_bmm sweep covers all group counts, not the first two shapes — `d68ba2d07`
9. added mxfp4 + mixed block-scale coverage to masked_scaled_bmm — `00a27a8aa`
10. ragged_block_scaled_bmm covers non-uniform segments — `14f1351dd` (+ alignment axis `4971c7573`)
11. coverage for the group-GEMM `segment_alignment>=128` TMA fast path — `ef437f3b2`
12. uneven + zero-length group coverage — `aa35c8571`
13. guarded verify_cutile in both bench sweeps so one bad config can't abort the run — `74a20c314`

Full cuTile test run is green: masked_bmm 128, masked_scaled_bmm 8, ragged_block_scaled 8, group GEMM 118 passed / 0 failed.

yifeis-nv:
Rebased onto latest `main` (base `2c78870`) and worked through the remaining bot findings.

Two were still valid and are now fixed:
- `ragged_block_scaled_bmm`: input-validation `assert`s converted to `raise ValueError` so they survive `python -O` (Gemini G2/G3).
- Documented `FLASHINFER_CUTILE_AUTOTUNE_DISABLED` in `docs/autotuning.rst`.

All other Gemini/CodeRabbit items were verified already addressed in the rebased tree (assert→ValueError in `masked_bmm`, `out=`/device-binding paths, TMA fast-path gate, test tolerances and capability checks). The `cutile_common.cached_replace_hints` duplicate hunk dropped out on rebase as noted, since #4019 landed first.

Validated on B200: the six `ragged_block_scaled_bmm` negative paths now raise `ValueError`, and valid inputs produce the same output as before.

flashinfer/gemm/kernels/cutile/masked_scaled_bmm_cutile.py
dhiraj113: Verified at 01c5c888 (`f2882623`): re-ran the original repro on a B200 — the Q=1025 fallback now returns shape `(1,)`, the cached scratch keeps its shape (same object, no resize warning), and the follow-up small-Q call succeeds. Resolved, thanks.

flashinfer/gemm/kernels/cutile/masked_bmm_cutile.py
dhiraj113: Verified at 01c5c888 (`616ecdf4`): all three degenerate cases (Q=0, M=0, N=0) now return empty tensors with shapes exactly matching `torch.bmm`, on both the autotune and `FLASHINFER_CUTILE_AUTOTUNE_DISABLED=1` paths (the early return sits after validation, before the branch, so one guard covers both), and the `out=` interplay is correct for empty shapes. Resolved, thanks.

flashinfer/gemm/kernels/cutile/ragged_scaled_bmm_cutile.py
dhiraj113: Verified at 01c5c888 (`bca72b9c`): the kernel now reads its loop bound from a device-side `max_m_device` computed via `torch.amax(seg_lens)` (async, CUDA-graph safe), and both launch paths size the grid from `total_m` so the device bound always has programs to run on. Re-ran the original repro: understated `max_m=128` on a 512-row segment now produces fully-written, bit-exact output (0 sentinel rows), and the correct-`max_m` baseline is unchanged. The Q==0 early return also keeps the `amax` well-defined. Resolved, thanks.

flashinfer/gemm/kernels/cutile/ragged_bmm_cutile.py
dhiraj113: Verified at 01c5c888 (`e06d521d`): the `max_m_device=None` fallback is now derived on-device from `m_indptr` diffs (`torch.amax(keepdim=True)`) instead of the circular `torch.tensor([max_m])`, grids are sized from `total_m`, and the understated-`max_m=64` repro now produces fully correct output (0 sentinel rows, deviation equal to bf16 rounding, same as the control). CUDA-graph safe (no host sync). Resolved, thanks.

flashinfer/gemm/gemm_base.py
dhiraj113: Verified at 01c5c888 (`19f8a3eb`): the gate now raises clean, descriptive `ValueError`s for non-FP8 a/b dtypes, non-fp32 block scales, and non-NT layout — both original failure modes (bf16/bf16 silently accepted; fp8/bf16 opaque `TileTypeError`) re-tested and both now fail cleanly at the gate, while valid FP8+fp32-scale inputs still pass. This meets (and slightly exceeds) sibling-gate parity. Resolved, thanks.

flashinfer/gemm/kernels/cutile/ragged_block_scaled_bmm_cutile.py
dhiraj113: Verified at 01c5c888 (`2494f9a3`): the `a_scale` K-granularity guard now sits right after the `b_scale` guard on the single launch path and raises a descriptive `ValueError` (consistent with `c885ea7a`'s ValueError-not-assert convention). Re-ran the repro: both the too-narrow (previously silent zero-padding) and too-wide (previously mis-applied scales) cases now raise cleanly; the correct-shape control still matches the torch reference (rel err 0.0025). Resolved, thanks.

flashinfer/gemm/kernels/cutile/masked_scaled_bmm_cutile.py
dhiraj113: Verified at 01c5c888 (`0fc10c50`): `max_m_device` is now required to be a CUDA tensor on `a.device` — CPU tensors and python ints are rejected with clean `ValueError`s, the hidden host-sync `.item()` fallback is fully removed, and a valid on-device tensor produces output bit-identical to the `max_m_device=None` path. The cross-GPU case is closed by construction (guard runs before any launch). Resolved, thanks.

tests/gemm/test_masked_bmm_cutile.py
dhiraj113: Verified at 01c5c888 (`f3cc7f51`): the `[:2]` truncation is replaced by a principled `select_correctness_cases()` sampler — 32 parametrize entries covering 16 distinct `(num_groups, m_per_group)` cases across all 9 group counts (including the previously-dead GB200 counts 8/9/18/36/72), each at its per-group m extremes; 128 tests collect cleanly. Nice fix — deterministic and CI-sized. Resolved, thanks.

tests/gemm/test_masked_scaled_bmm_cutile.py
dhiraj113: Verified at 01c5c888 (`67fb6922`): `block_scale_type` now parametrizes all 4 values, and I ran the 4 new mxfp4/mixed cases on a B200 — all pass against the dequantized torch reference, so the MIXED_PREC and fp4xfp4 branches are now genuinely exercised (the sm103 fp4-packing skip generalization looks right too). Resolved, thanks.

tests/gemm/test_ragged_block_scaled_bmm_cutile.py
dhiraj113: Mostly resolved at 01c5c888 (`380ff219` + `6d480353`): the new `segment_mode={uniform,ragged}` axis with `create_ragged_aligned_m_segments` produces genuinely non-uniform segments (the old dead helper is removed), the `segment_alignment={128,256}` axis is threaded through, and all 8 cases pass on a B200 — the imbalanced prefix-sum schedule is now exercised in CI. One small remainder from the original comment: **zero-row segments** are still never tested (the new generator is deliberately strictly positive). Our earlier probe showed the kernel handles a zero-row segment correctly, so this is purely a test add — a single case with one empty segment (e.g. sizes `[256, 0, 128, 512]`) would close it. Non-blocking.

tests/gemm/test_groupwise_scaled_gemm_fp8.py
dhiraj113: Verified at 01c5c888 (`ae04a6d4`): `test_fp8_groupwise_group_gemm_cutile_tma_fast_path` drives the public API with `segment_alignment` in {128, 256} and satisfies every conjunct of the TMA dispatch predicate — confirmed with a spy that the test actually dispatches into the TMA branch, and all 4 parametrizations pass on a B200. Resolved, thanks.

tests/gemm/test_groupwise_scaled_gemm_fp8.py
dhiraj113: Verified at 01c5c888 (`98775723`): the new `test_fp8_groupwise_group_gemm_uneven_segments` covers genuinely uneven groups including a 4-row group (forcing the OOB-sentinel tail-drop) and a zero-length group at front/middle/back positions, with an explicit prefix-sum `m_indptr` and a per-group reference that skips empty groups. Ran at head on a B200: cutile 3/3 and trtllm 3/3 pass. Resolved, thanks.

benchmarks/bench_masked_scaled_bmm_backend_comparison.py
dhiraj113: Verified at 01c5c888 (`0e5c8c76`): `verify_cutile()` is now guarded in **both** this script and the sibling `bench_ragged_scaled_bmm_backend_comparison.py`. Re-ran the original `--dtype bogus,mxfp8` repro on both: the bogus config degrades to `FAIL (...)` with `--` rows and the valid mxfp8 sweep runs to completion, exit 0. Resolved, thanks.

dhiraj113 (approved):
Re-reviewed at head `01c5c888` (rebased branch) — all 13 findings from the previous review were re-verified by re-running their original executed repros on a B200 (sm100): 12 fully resolved, 1 mostly resolved with a tiny test-only remainder (zero-row segments in `test_ragged_block_scaled_bmm_cutile.py` — the kernel handles them correctly per our earlier probe, so it's a single test case to add; detailed in the thread).

The fix series is exemplary — one commit per finding, each mechanism-level rather than symptom-level (device-derived `max_m` defenses in both ragged ops, gate validation parity, principled test sampling instead of the `[:2]` slice, and the previously-untested TMA / uneven-group / mixed-precision paths now genuinely exercised and passing).

Approving. (AI-assisted review per `docs/code_review_guidance.md`; every verdict above is backed by an executed repro at this head.)

dhiraj113:
@flashinfer-bot run

yifeis-nv:
@flashinfer-bot run

yifeis-nv:
@flashinfer-bot run

yifeis-nv:
Hi @dhiraj113 — thanks again for the approval! Since your last `/bot run tests/gemm` (pipeline #65703344, 15/16 passed) the branch has only picked up main-merges and one `ruff-format` pass — no substantive code changes — with the head now at `198f18c90` (merged latest `main` today).

That last run's only non-pass was RTX Pro 6000 Blackwell / **cu130**, classified *Unknown/infra* ("script failed before producing a JUnit report"); the same GPU on cu12.9 passed, and B300/GB200/GB300/H100 plus all 6 multi-GPU/multi-node jobs were green — so it looks like an infra flake rather than a code failure.

Could you re-run `/bot run tests/gemm` on the current head when you get a chance? That should clear the authorization check and re-run GPU CI. Appreciate it!

bkryu:
@flashinfer-bot run

PR #4027, opened 2026-07-17
MoE monokernel Bug fix, barrrier remove and kernel rewrite.

yugong333:
Find two bugs in previous design:
1. The cross-block handoffs counters are not reset correctly in the scratchpad. If the scratchpad is reused the down projection will fetch the wrong data based on the wrong status of the counter
2. inv_scale overflow on near-subnormal block maxima.

Fixes:

1.  Kernel rewrite. Replace the software grid barriers with sentinel/flag
cross-block handoffs:
- Delete moe_grid_barrier.h and the grid/partial-barrier counter protocol.
- Phase 3->4 handoff: each temp_act_scale cell is release-published after its fp8 payload segment and doubles as the readiness flag; the down-projection polls exactly the cells it consumes (per-expert granularity), so no barrier-counter rendezvous is needed.
- Phase 4->5 handoff: each block bumps its column stripe's parity-selected arrival counter; only that stripe's Phase-5 writer polls it.
- The launch-parity double-buffer makes the handoff state self-maintaining across CUDA-graph replays; the kernel stays capturable with a plain launch.

2. Replace the too-narrow subnormal check with an epsilon clamp on the block max, before computing either scale:

```
// Eps-clamp tiny maxima: blk_max slightly above FLT_MIN still overflows
// blk_inv_scale (448/blk_max > FLT_MAX for blk_max < ~1.32e-36), NaN-ing
// the whole block after the fp8 cast. 1e-10 matches vLLM's group-quant eps.
blk_max = fmaxf(blk_max, 1e-10f);

const float blk_act_scale = blk_max * FP8_MAX_INV;  // stored dequant scale
const float blk_inv_scale = FP8_MAX / blk_max;      // quant multiplier, now finite
```

3. Tests: tests/moe/test_monomoe.py gains scratchpad-reuse (no cross-launch contamination) and small-scale correctness coverage on top of the accuracy sweep (M in {1,2,8} x top_k {1,8}); all pass on H200.

| M | Old Kernel | Fixed Kernel |
|---|---|---|
| 1 | FAILED (reuse vs fresh: cos=0.000) | PASS (all > 0.9999) |
| 2 | FAILED (stale barrier race) | PASS (all > 0.9999) |
| 4 | FAILED (stale barrier race) | PASS (all > 0.9999) |
| 8 | PASS | PASS |

## Summary by CodeRabbit

* **New Features**
  * Added/standardized support for the fixed block-FP8 configuration (E=256, N=512, K=2048).

* **Bug Fixes**
  * Strengthened runtime shape and routing-input validation, including token-cap checks and clearer failure messages.
  * Improved numerical stability for small-scales and ensured scratchpad reuse does not contaminate results.

* **Documentation**
  * Refreshed MonoMoe design/docs to reflect the supported hardware, fixed-shape constraints, and execution flow.

* **Tests**
  * Updated and expanded MonoMoe FP8 reference comparisons and new stability/scratchpad reuse coverage.


csrc/fused_moe/monomoe/monomoe_wrapper.cuh
```diff
@@ -94,14 +100,23 @@ void monomoe_topk_launcher(TensorView activations_in, TensorView router_logits,
...
+    // so for M < BS the routing-window load reads exactly M rows and
+    // hardware-zeros the phantom rows.  Under CUDA graphs the descriptor is
+    // captured per graph (one graph per batch size), so the baked-in M
+    // always matches replays.
     activations_desc = create_activations_tma_desc(
-        reinterpret_cast<const void*>(activations_in_ptr), Dims::BS, Dims::HIDDEN_STATES);
-    down_weights_desc = create_down_weight_tma_desc(
-        reinterpret_cast<const void*>(expert_weights_down_ptr), Dims::NUM_EXPERTS,
-        Dims::HIDDEN_STATES, Dims::N, /*row_box=*/MoECoreDims<Dims>::DOWN_COL_TILE);
+        reinterpret_cast<const void*>(activations_in_ptr), num_tokens, Dims::HIDDEN_STATES);
```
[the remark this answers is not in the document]
yugong333: Three BS checks runs before this line 
1. FFI entry point — monomoe_binding.cu:81-86: monomoe_topk rejects m < 1 || m > Dims::BS with TVM_FFI_ICHECK before. This is exactly "the module's existing validation mechanism" the comment asks to use. 
2. Python API — flashinfer/fused_moe/monomoe.py:213-214: raise ValueError for m > 8 before the FFI call is ever made.
3. Device-side backstop — moe.cuh:299-300: assert(token_count <= Dims::BS) in the kernel.

So by the time control reaches the flagged create_activations_tma_desc(..., num_tokens, ...) at monomoe_wrapper.cuh:108, 1 <= num_tokens <= Dims::BS is already guaranteed. Adding a fourth check inside the launcher would be redundant defensive code guarding an unreachable state.

csrc/fused_moe/monomoe/src/moe.cuh
```diff
@@ -375,28 +305,14 @@ __global__ __launch_bounds__(Dims::KernelConfig::BLOCK_SIZE, 1) void moe_kernel_
...
-  moe_kernel_topk_BS8<Dims>(activations_in, token_count, router_logits, expert_weights_up,
-                            expert_scales_up, expert_weights_down, expert_scales_down,
-                            activations_out, top_k, scoring_func, renormalize, spec, shmem,
-                            up_weights_desc, activations_desc, down_weights_desc,
-                            down_activations_desc, grid_counters, grid_phase, expert_counters,
-                            expert_phase, colstripe_counters, colstripe_phase);
+  moe_kernel_topk_impl<Dims>(activations_in, token_count, router_logits, expert_weights_up,
+                             expert_scales_up, expert_weights_down, expert_scales_down,
+                             activations_out, top_k, scoring_func, renormalize, expert_bias,
+                             routed_scaling_factor, spec, shmem, up_weights_desc, activations_desc,
```
jhaotingc: routed_scaling_factor will this added arg be added into test?

csrc/fused_moe/monomoe/src/ptx_utils.h
jhaotingc, line 109: > with CuTe-delegated WGMMA control ops replaced by explicit inline PTX.

Is this removed? seems that it's still CuTe

jhaotingc (commented):
- Description says it adds cuda graph replay, adding a test would be nice
- test for added bias 
- consider splitting this PR into (1) bug fix (sync rewrite) (2) new routing API
- Some local benchmark/acc test that shows this PR resolve acc issue would be more direct, e.g. has this accuracy test been taken care of by test_monomoe_accuracy? why is test_monomoe_accuracy only testing 1 case? maybe H200 latency before/after for M={1,2,4,8}, top_k={1,8}

Thank you!

csrc/fused_moe/monomoe/src/ptx_utils.h
yugong333: Clarify the comments

yugong333:
> * Description says it adds cuda graph replay, adding a test would be nice
> 
>     * test for added bias
> 
>     * consider splitting this PR into (1) bug fix (sync rewrite) (2) new routing API
> 
>     * Some local benchmark/acc test that shows this PR resolve acc issue would be more direct, e.g. has this accuracy test been taken care of by test_monomoe_accuracy? why is test_monomoe_accuracy only testing 1 case? maybe H200 latency before/after for M={1,2,4,8}, top_k={1,8}
> 
> 
> Thank you!


Thank you @jhaotingc ! I have removed the additional routing method support and focus on the bug fix only now. 
The bugs are related to \1 barrier race and \2 small weight scale. The updated test script covers these two cases and sweeps all cases of BS = {1, 2, 4, 8} and topK = {1, 8}.

Thank you for you comments!

aleozlx (approved):
lgtm

PR #4028, opened 2026-07-17
gemm: avoid unsafe TinyGEMM shapes on SM103

### LINKED ISSUE #3848: BF16 TinyGEMM can hang on Blackwell SM103 for large projection shapes
## Summary

On Blackwell B300A / SM103, BF16 `mm_bf16` can hang indefinitely for some large projection shapes when TinyGEMM is used. This is visible both when calling `backend="tinygemm"` directly and when `backend="auto"` / vLLM's FlashInfer BF16 path autotunes across candidates that include `tinygemm2`.

The hang looks like a CUDA kernel that never completes: GPU utilization stays high and the Python process blocks in `torch.cuda.synchronize()` / the FlashInfer autotuner profiling synchronize. The CUDA context cannot be recovered in-process; the benchmark watchdog had to kill and restart the worker.

## Environment

- GPU: B300A / Blackwell, `torch.cuda.get_device_capability() == (10, 3)`
- Device name from PyTorch: `NVIDIA Graphics Device`
- FlashInfer: `0.6.12`
- PyTorch: `2.11.0+cu130`
- CUDA: `13.0`
- Python: `3.12.3`

## Minimal repro

This repro runs each case in a subprocess because a hung TinyGEMM launch leaves the CUDA context wedged.

```python
import multiprocessing as mp
import os
import sys


def run_case(m: int, n: int, k: int, bias_enabled: bool) -> None:
    import torch
    import flashinfer

    torch.manual_seed(0)
    a = torch.randn((m, k), device="cuda", dtype=torch.bfloat16)

    # mm_bf16 expects B as shape [k, n] with B.T contiguous.
    b_storage = torch.randn((n, k), device="cuda", dtype=torch.bfloat16)
    b = b_storage.t()

    bias = None
    if bias_enabled:
        bias = torch.randn((n,), device="cuda", dtype=torch.bfloat16)

    torch.cuda.synchronize()
    print(
        "running",
        {"m": m, "n": n, "k": k, "bias": bias_enabled},
        flush=True,
    )
    out = flashinfer.mm_bf16(
        a,
        b,
        bias=bias,
        pdl=False,
        backend="tinygemm",
    )
    torch.cuda.synchronize()
    print("done", out.shape, out.dtype, flush=True)


def main() -> int:
    cases = [
        # Observed watchdog hangs in a single-layer benchmark.
        (1, 7168, 5120, False),
        (1, 55296, 5120, False),
        (1, 5120, 27648, True),
        (8192, 5120, 27648, False),
    ]
    for case in cases:
        p = mp.Process(target=run_case, args=case)
        p.start()
        p.join(timeout=30)
        if p.is_alive():
            print("HANG", case, flush=True)
            p.kill()
            p.join()
            return 1
        if p.exitcode:
            print("FAILED", case, "exitcode", p.exitcode, flush=True)
            return p.exitcode
    return 0


if __name__ == "__main__":
    mp.set_start_method("spawn", force=True)
    raise SystemExit(main())
```

Expected: each call should either complete or return a regular Python/CUDA error.

Actual: at least some cases hang until the subprocess is killed.

## Additional evidence from a larger single-layer sweep

The local benchmark key format is:

```text
backend,pdl,bias,m,n,k
```

Examples of explicit TinyGEMM cells that the watchdog killed:

```text
STUCK on 'tinygemm,False,False,8192,5120,27648'
STUCK on 'tinygemm,False,True,1,5120,27648'
STUCK on 'tinygemm,False,False,1,55296,5120'
STUCK on 'tinygemm,False,False,1,7168,5120'
STUCK on 'tinygemm,False,True,1024,7168,5120'
```

We also saw `backend="auto"` / vLLM `vllm-auto` hangs on the same family of large-N / large-K shapes. In those cases the process was stuck under the autotuner profiling path:

```text
mm_bf16
  bf16_gemm_sm100
    AutoTuner.choose_one
      _profile_single_kernel
        pure_profile
          torch.cuda.Stream.synchronize  # never returns
```

The candidate ops loaded before the autotune wedge included:

```text
mm_bf16_cublaslt, bf16_gemm_cutlass, tgv_gemm_bf16, tinygemm2, trtllm_utils
```

We cannot prove from the auto-path stack alone which candidate is always responsible, but explicit `backend="tinygemm"` hangs on overlapping shapes make TinyGEMM a strong suspect. As a local mitigation, excluding `tinygemm` from BF16 `backend="auto"` candidates avoids exposing vLLM's BF16 FlashInfer autotune path to this hang.

## Notes

TinyGEMM is described as a small-M BF16 GEMM backend, but the current hard requirement checks allow some large projection shapes through. It may be safer for TinyGEMM to reject unsupported / out-of-design-range shapes before launch, or for BF16 auto selection to avoid TinyGEMM on these shapes.

sj0618:
## Summary

- guard BF16 TinyGEMM against the large SM103 projection shapes reported in #3848
- reject unsafe explicit `backend="tinygemm"` calls before launching a wedging kernel
- exclude the same shapes from `backend="auto"` candidate selection
- add regression coverage for reported shapes and the threshold boundary

## Testing

- `python3 -m py_compile flashinfer/gemm/gemm_base.py tests/gemm/test_mm_bf16.py`
- `ruff check flashinfer/gemm/gemm_base.py tests/gemm/test_mm_bf16.py`
- `git diff --check`
- pure guard behavior checks passed
- CUDA pytest could not run in this checkout because the sparse worktree omits the JIT package and no CUDA device is available

Closes #3848

## Summary by CodeRabbit

* **Bug Fixes**
  * Added an SM103-specific safety guard to prevent selecting the TinyGEMM backend for unsafe BF16 projection shapes.
  * Now raises a clear error to avoid a known hang when the BF16 projection is too large.

* **Tests**
  * Added SM103 TinyGEMM guard tests, including expected-boolean cases and a boundary check around the projection-size threshold.

### DIFF
--- a/flashinfer/gemm/gemm_base.py
+++ b/flashinfer/gemm/gemm_base.py
@@ -118,6 +118,30 @@ from ..utils import (
 )
 
 DEFAULT_WORKSPACE_SIZE = 32 * 1024 * 1024
+_SM103_TINYGEMM_MAX_PROJECTION_ELEMENTS = 32 * 1024 * 1024
+
+
+def _is_sm103_tinygemm_unsafe_shape(
+    compute_capability: Tuple[int, int], n: int, k: int
+) -> bool:
+    """Return whether TinyGEMM is unsafe for an SM103 projection shape.
+
+    TinyGEMM can hang indefinitely on B300A/SM103 for large BF16 projection
+    shapes. Keep the workaround scoped to the affected architecture and
+    disable it only once the N*K projection size reaches the smallest
+    reported failure threshold. See flashinfer-ai/flashinfer#3848.
+    """
+    return (
+        compute_capability == (10, 3)
+        and n * k >= _SM103_TINYGEMM_MAX_PROJECTION_ELEMENTS
+    )
+
+
+def _is_sm103_tinygemm_unsafe(a: torch.Tensor, b: torch.Tensor) -> bool:
+    return _is_sm103_tinygemm_unsafe_shape(
+        get_compute_capability(a.device), b.shape[1], a.shape[1]
+    )
+
 
 # sizeof(cublasLtMatmulAlgo_t) = uint64_t[8] = 64 bytes.
 # Shared by cuBLAS FP8, cuBLASLt BF16, and any other cuBLASLt-based runners.
@@ -445,6 +469,11 @@ def _tinygemm_mm_bf16_requirement(
         raise ValueError("The TinyGEMM backend requires a contiguous output tensor.")
     if bias is not None and not bias.is_contiguous():
         raise ValueError("The TinyGEMM backend requires a contiguous bias tensor.")
+    if _is_sm103_tinygemm_unsafe(a, b):
+        raise ValueError(
+            "The TinyGEMM backend is disabled for this large BF16 projection on SM103 "
+            "to avoid a known kernel hang; see flashinfer-ai/flashinfer#3848."
+        )
     return True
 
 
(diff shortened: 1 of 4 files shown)


bkryu:
Thanks @sj0618, the issue shold have been fixed in #3630 by fixing the root cause.

I will close the current PR. We always appreciate future contributions from the community :)

PR #4029, opened 2026-07-17
perf(gemm): Improve mm_fp4 cute-dsl autotune time via disk-cache and parallel compilation

bkryu:
## 📌 Description

Follow-up to #3948 (top-N tactic ranking) and built on #3874 (CuTe-DSL disk cache). Autotuned `mm_fp4(backend='cute-dsl')` still:
* Recompiles every tactic in-process
* Compiles serially, taking 2-3 seconds per kernel

This PR:

1. Persists mm_fp4 kernels to the on-disk CuTe-DSL cache.  Loading `JITLink` cached artifacts from disk is measured to be done in ~10 ms per kernel.
2. Parallelizes first-time compilation. The autotuner checks for not-yet-cached tactics with a subprocess pool (default 4, RAM-capped) that persists directly into the shared disk cache.
  * default pool size of 4 was set to be conservative. Empirically measured RAM (RSS) size per compilation was ~1GB per compilation.

### Autotune time improvements
Evaluated with
`python3 benchmarks/flashinfer_benchmark.py --routine mm_fp4 --m 256 --n 1024 --k 7168 --out_dtype bfloat16 --backends cute-dsl --use_128x4_sf_layout --use_nvfp4 --refcheck --autotune`

End-to-end process wall time on B200, 3 runs each:

| Scenario | Runs | Median | Speedup |
|---|---|---:|---:|                             
| Current main branch main (in-process serial compilation) | 94.9 / 93.5 / 93.2 s | **93.5 s** | 1.0× |
| this PR, first run; cold disk cache (parallel compile + persist) | 45.8 / 44.6 / 46.0 s | **45.8 s** | **~2×** |
| this PR, every subsequent process; hot disk cache (load from disk) | 4.38 / 4.35 / 4.35 s | **4.35 s** | **~20×** |

Autotuned results are unchanged: the same tactics are profiled with bitwise identical kernel and output; just changes in the compilation infra.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Performance**
  * Improved FP4 GEMM startup by precompiling eligible CuTe-DSL tactics.
  * Added parallel CuTe-DSL kernel compilation with persistent on-disk caching to reduce repeated build times.
  * Strengthened device-scoped caching to speed up subsequent launches.
* **Reliability**
  * Disk persistence failures now fall back gracefully, allowing kernels to compile on demand.
  * Updated FP4/MXF dtype handling and improved deterministic, collision-resistant kernel naming.
* **Tests**
  * Added coverage ensuring FP4 CuTe-DSL kernel names change with all codegen arguments and are filename-safe.

### DIFF
--- a/flashinfer/gemm/gemm_base.py
+++ b/flashinfer/gemm/gemm_base.py
@@ -53,6 +53,14 @@ from ..fused_moe.utils import (
     get_hybrid_num_tokens_buckets,
     map_to_hybrid_bucket_uncapped,
 )
+from .gemm_mm_fp4_cute_dsl import (
+    _TORCH_TO_CUTLASS_DTYPE_ATTR,
+    _blockscaled_gemm_cache_key_files,
+    _blockscaled_kernel_disk_name,
+    _compile_block_scaled_gemm,
+    _get_mm_fp4_cute_dsl_compile_workers,
+    run_mm_fp4_precompile_pool,
+)
 from .kernels.utils import (
     _SM100_CLUSTER_SHAPE_MN_CANDIDATES,
     _SM100_MMA_TILER_MN_CANDIDATES,
@@ -126,11 +134,6 @@ _CUBLASLT_MAX_ALGOS = 100
 
 # Error messages
 CUDNN_FP4_MXFP4_SM120_CUDNN_VERSION_ERROR = "cudnn FP4 GEMM with mxfp4 quantization is not supported on SM120/SM121 with cuDNN backend version < 9.14.0."
-
-_TORCH_TO_CUTLASS_DTYPE_ATTR = {
-    torch.bfloat16: "BFloat16",
-    torch.float16: "Float16",
-}
 
 
 def _match_sm_version(device: torch.device, sm_version: list[str]):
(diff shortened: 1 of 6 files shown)


flashinfer/jit/cute_dsl_core.py
```diff
@@ -184,6 +184,36 @@ def build(self) -> None:
...
+                )
+                shutil.rmtree(self.module_dir, ignore_errors=True)
+            try:
+                self._export()
+            except Exception as e:
+                logger.warning(
+                    f"Failed to persist CuTe-DSL kernel {self.name} to "
+                    f"{self.object_path}: {e}. The kernel will be recompiled "
+                    f"next run."
+                )
```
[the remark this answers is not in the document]
bkryu: Done — added # noqa: BLE001 -- persistence is best-effort and # noqa: BLE001 -- serial fallback is intentional at the two sites in next commit

flashinfer/gemm/gemm_mm_fp4_cute_dsl.py
```diff
@@ -0,0 +1,432 @@
...
+    if available is not None:
+        mem_cap = int(available // _MM_FP4_PRECOMPILE_WORKER_RAM_BYTES)
+        if mem_cap < workers:
+            logger.warning(
+                f"[mm_fp4 cute-dsl] capping tactic precompile workers "
+                f"{workers} -> {mem_cap} (host MemAvailable = "
+                f"{available / (1 << 30):.1f} GiB)."
+            )
+            workers = mem_cap
+    return workers
```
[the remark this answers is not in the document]
bkryu: Good catch. Added _cgroup_available_memory_bytes()  and now take the min of that and MemAvailable in next commit.

flashinfer/gemm/gemm_mm_fp4_cute_dsl.py
```diff
@@ -0,0 +1,432 @@
...
+            "mm_fp4",
+            kernel_name,
+            compile_fn,
+            _hash_source_files(tuple(payload["key_files"])),
+        )
+        if not spec.is_compiled:
+            spec.compile_and_persist()
+        return (kernel_name, None)
+    except Exception as e:  # noqa: BLE001 -- reported to the parent
+        return (kernel_name, f"{type(e).__name__}: {e}")
```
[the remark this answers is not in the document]
bkryu: Done in next commit — precompile_mm_fp4_tactics now takes the caller's device.

flashinfer/gemm/gemm_mm_fp4_cute_dsl.py
```diff
@@ -0,0 +1,432 @@
...
+
+        compiled_gemm = build_and_load_cute_dsl_kernel(
+            cache_module_name,
+            _blockscaled_kernel_disk_name(cache_key, batch_size, max_active_clusters),
+            compile_kernel,
+            extra_key_files=_blockscaled_gemm_cache_key_files(),
+        )
+
+    result = (compiled_gemm, max_active_clusters)
+    cache[cache_key] = result
```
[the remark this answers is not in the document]
bkryu: Agreed the in-memory key was missing device identity (a pre-existing gap on main)

Implemented with one deviation from the suggested mechanics in next commit. The in-memory key is now (device_index, cache_key).

flashinfer/gemm/gemm_mm_fp4_cute_dsl.py
```diff
@@ -0,0 +1,482 @@
...
+def _blockscaled_gemm_cache_key_files() -> tuple:
+    """Source files whose content invalidates the on-disk mm_fp4 kernels."""
+    from .kernels import (
+        dense_blockscaled_gemm_sm100,
+        dense_blockscaled_gemm_sm100_common,
+    )
+
+    return (
+        __file__,
+        dense_blockscaled_gemm_sm100.__file__,
```
qiching, line 38: forward passes `cache_module_name="mm_fp4"` unconditionally, so sm103 tactics get persisted to disk, but this key-files tuple omits `dense_blockscaled_gemm_sm103.py`, so if edit the sm103 kernel won't invalidate its cached `.o` and a stale kernel will be served. Please add the sm103 source here.

qiching (changes requested):

flashinfer/gemm/gemm_mm_fp4_cute_dsl.py
bkryu: Thanks @qiching, good point added in latest commit 8aae20acdebb358c7336068492e9412703eb30a0

qiching (approved):
LGTM

flashinfer/gemm/gemm_base.py
```diff
@@ -53,6 +53,13 @@
     get_hybrid_num_tokens_buckets,
     map_to_hybrid_bucket_uncapped,
 )
+from .gemm_mm_fp4_cute_dsl import (
+    _TORCH_TO_CUTLASS_DTYPE_ATTR,
+    _compile_block_scaled_gemm,
+    _device_index,
+    _mm_fp4_cache_key,
+    precompile_mm_fp4_tactics,
+)
```
dhiraj113, line 60: Are all these applicable only to `gemm_mm_fp4_cute_dsl`? If not, this structure doesn't look great. Please move them to separate utils files.

bkryu: Thanks @dhiraj113, good call. Removed `_TORCH_TO_CUTLASS_DTYPE_ATTR` and `_device_index` from here.
* I deleted  `_TORCH_TO_CUTLASS_DTYPE_ATTR` because there was a `torch_to_cutlass_dtype()` in `flashinfer/cute_dsl/utils.py`
  * Had to change all consumers of `_TORCH_TO_CUTLASS_DTYPE_ATTR` to use `torch_to_cutlass_dtype`
* `_device_index` moved to `flashinfer/utils.py` as get_device_index. It's a generic torch helper with nothing cute-dsl or nothing GEMM so The higher level `utils.py` is a good place

Changes made in the latest commit c437973

flashinfer/gemm/gemm_base.py
```diff
@@ -6112,17 +6026,7 @@ def forward(
...
-                mma_tiler_mn,
-                cluster_shape_mn,
-                swap_ab,
-                use_prefetch,
-                kernel_type,
-                use_tma_store,
-                enable_pdl,
-                out_dtype,
-            )
+            cache_key = _mm_fp4_cache_key(sf_vec_size, tactic, enable_pdl, out_dtype)
```
dhiraj113, line 6029: NIce clean up.

tests/jit/test_cute_dsl_cache.py
```diff
@@ -156,3 +156,76 @@ def test_nvfp4_kernel_name_is_symbol_safe():
...
+
+# ---------------------------------------------------------------------------
+# mm_fp4 (flashinfer/gemm/gemm_mm_fp4_cute_dsl.py) cache adopter
+# ---------------------------------------------------------------------------
+
+import torch  # noqa: E402
+
+from flashinfer.gemm.gemm_mm_fp4_cute_dsl import (  # noqa: E402
+    _blockscaled_kernel_disk_name,
+    _mm_fp4_cache_key,
```
dhiraj113, line 169: Is testing this not possible at flashinfer API level? Are all these APIs too?

bkryu: Thanks @dhiraj113, the imports here for the `test_mm_fp4_kernel_name_varies_with_every_argument` below intentionally import the internal helper functions directly, instead of going through `mm_fp4()`. 

As you asked, the reason for importing the helper is: what is being tested can't be reached from the public API. The test checks that changing any single kernel parameter changes the cache file name (i.e. so two different kernels can never end up colliding in one cached file name). These kernel parameters are things like tile shape, cluster shape, swap_ab a user cannot choose or even query through the public mm_fp4 API. Instead they are chosen by the heuristic or the autotuner. 

Our public APIs are decorated with `@flashinfer_api`. These undecorated functions also have the underscore prefix so I'd believe it is reasonably clear that these are not APIs, despite being imported in the test file.

dhiraj113: Okay, so this is more like an unit test then right? Not for now, but maybe such tests should be separated.

dhiraj113 (approved):

PR #4039, opened 2026-07-17
feat(msa): accept K/V views split from a packed paged KV cache

yichengj0:
## 📌 Description

Lets `msa_sparse_attention` and `msa_sparse_decode_attention` accept K/V views split from a paged KV cache that packs K and V in one `2 * head_dim` content dim per token. This is vLLM's KV-cache layout for MiniMax-M3 (vllm-project/vllm#44455), so it unblocks the vLLM integration of the MSA attend on SM120/SM121 (vllm-project/vllm#48994).

Design:

- The wrappers detect the split-view pattern and hand the kernel the whole packed cache; a `kv_packed` kernel variant reads K from the first half and V from the second half of each token's row. The contiguous path is unchanged.
- Packed views are accepted for paged bf16/fp16/fp8 caches only; NVFP4 and non-paged inputs still require contiguous K/V, since vLLM's packed cache is paged and never NVFP4, so nothing produces those combinations.
- The detection result is memoized so repeated calls with the same cache views skip it.
- Callers probe the capability via the new `msa_ops.SUPPORTS_PACKED_KV` flag.
- Also documents that `v_global_scale` applies to any KV dtype, which packed-cache callers rely on for the fp8 V descale.

### Performance

Each cell is the median time of the same op over 30 iterations with random block selections on a packed cache, divided by the same measurement on a contiguous cache (1.00 = accepting the packed layout costs nothing). Rows cover decode and prefill shapes at the two KV-cache dtypes the packed path accepts, bf16 and fp8. topk is 16, the MiniMax-M3 default; the head shape (8 query / 2 KV heads) is kept small so the kernels stay short and any per-call overhead would show. B=1 cells vary by a few percent between runs.

| shape (KV dtype, batch B, KV length S) | RTX 5080 | RTX PRO 6000 | GB10 |
|---|---|---|---|
| decode bf16 B=1 S=8192 | 1.01 | 0.98 | 0.80* |
| decode fp8 B=1 S=8192 | 1.04 | 1.02 | n/a** |
| decode bf16 B=8 S=8192 | 0.98 | 1.02 | 1.03 |
| decode fp8 B=8 S=8192 | 1.02 | 0.99 | n/a** |
| decode bf16 B=64 S=8192 | 0.98 | 1.01 | 0.99 |
| decode fp8 B=64 S=8192 | 1.00 | 1.01 | n/a** |
| decode bf16 B=256 S=8192 | 1.01 | 0.99 | 1.01 |
| decode fp8 B=256 S=8192 | 1.00 | 1.02 | n/a** |
| decode bf16 B=8 S=32768 | 1.01 | 1.03 | 1.09* |
| decode fp8 B=8 S=32768 | 1.02 | 1.05 | n/a** |
| prefill bf16 B=4 Q=512 S=8192 | 1.00 | 1.00 | 0.99 |
| prefill fp8 B=4 Q=512 S=8192 | 1.00 | 1.00 | n/a** |

\* Short GB10 runs vary a lot between repeats, likely because the board shifts clocks under its shared CPU/GPU power budget; read the starred cells as parity within that noise.

\** fp8 KV kernels currently fail to compile on SM121 with cutlass-dsl 4.5.2; this reproduces without this PR.

## 🔍 Related Issues

Follow-up to #3655 (MSA for SM120/SM121); enables the vLLM MiniMax-M3 integration in vllm-project/vllm#48994.

### ✅ Pre-commit Checks

## 🧪 Tests

- New `tests/msa_ops/test_packed_kv.py`: decode and prefill on packed split views vs contiguous copies, across HND/NHD cache layouts, MQA and GQA, and bf16/fp8; plus the capability flag, the memoized detection, and rejection of strided views that are not packed splits.
- RTX 5080 (two units) and RTX PRO 6000 (SM120): every packed case matches its contiguous reference bit-exactly, and the full `tests/msa_ops` suite passes.
- GB10 (SM121): the bf16 cases pass bit-exactly; fp8 does not compile there (see the performance footnote).

## Reviewer Notes

- Most of the diff is plumbing for the second compiled kernel variant, where mistakes fail loudly at compile or launch time. The exception is `_packed_kv_view` in `_common.py`: it reinterprets the caller's memory under new strides, so if it accepts a layout it should reject, the result is silently wrong attention output rather than an error. That function and its memoization are the parts to review closely.

## Summary by CodeRabbit

* **New Features**
  * Extended sparse attention benchmarks with a `--kv_layout` option supporting `flat`, `paged`, and `packed`.
  * Added packed-KV cache view support for sparse prefill and decode.
  * Exported a capability flag so applications can detect packed-KV support.
* **Bug Fixes**
  * Added validation for supported `kv_layout`/dtype combinations and for valid packed-KV view layouts.
  * Improved decode execution to correctly propagate packed-KV metadata.
* **Tests**
  * Added SM120-121 tests covering packed-KV behavior, layout/dtype coverage, and invalid view rejection.

### DIFF
--- a/benchmarks/routines/sparse_attention.py
+++ b/benchmarks/routines/sparse_attention.py
@@ -47,6 +47,11 @@ _ROUTINE_BACKENDS = {
 
 def run_sparse_attention_test(args):
     """Route an MSA routine to its test function."""
+    if args.kv_layout != "flat" and args.routine not in (
+        "MSASparseAttention",
+        "MSASparseDecode",
+    ):
+        raise ValueError(f"--kv_layout {args.kv_layout} unsupported for {args.routine}")
     if args.routine == "MSAProxyScore":
         return testMSAProxyScore(args)
     elif args.routine == "MSASparseAttention":
@@ -105,6 +110,17 @@ def parse_sparse_attention_args(line, parser):
         default="bfloat16",
         help="KV dtype: bfloat16/float16/fp8_e4m3/nvfp4 (prefill/decode/pipeline).",
     )
+    parser.add_argument(
+        "--kv_layout",
+        type=str,
+        default="flat",
+        choices=["flat", "paged", "packed"],
+        help=(
+            "KV storage for prefill/decode: flat varlen, a paged cache, or "
+            "K/V views split from a paged cache that packs K|V per token "
+            "(vLLM's layout). paged/packed exclude nvfp4."
+        ),
+    )
     args = parser.parse_args(line)
     if args.verbose >= 1:
         print(f"[INFO] {args = }")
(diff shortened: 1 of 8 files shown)


saltyminty (approved):

saltyminty (changes requested):
(previous approval was accidental, will review in a bit)

saltyminty (approved):
Approved pending CI

PR #4056, opened 2026-07-19
fix(norm): clamp CUDA FP8 quantization to the output dtype range

### LINKED ISSUE #4053: [Bug]: Legacy CUDA RMSNorm quantization clips E5M2 output to the E4M3 range
## Observed behavior

With the legacy CUDA norm backend, `rmsnorm_quant` and `fused_add_rmsnorm_quant` clamp E5M2 output to ±448, the E4M3 finite range.

E5M2 supports finite values through ±57344, so representable results between 448 and 57344 are silently reduced to 448.

This affects `FLASHINFER_USE_CUDA_NORM=1` and fallback use when the CuTe DSL implementation is unavailable.

## Reproduction

```python
import torch
import flashinfer

x = torch.ones(1, 128, dtype=torch.float16, device="cuda")
weight = torch.full((128,), 1024.0, dtype=torch.float16, device="cuda")
out = torch.empty_like(x, dtype=torch.float8_e5m2)

flashinfer.rmsnorm_quant(out, x, weight, 1.0, enable_pdl=False)
print(out[0, 0].float())
```

For these inputs, RMS normalization produces approximately 1024:

| Output type | Finite maximum | Input to clamp | Previous output | Expected output |
|---|---:|---:|---:|---:|
| E4M3 | 448 | 1024 | 448 | 448 |
| E5M2 | 57344 | 1024 | 448 | 1024 |

```text
previous = clamp(1024, -448, 448) = 448
expected = clamp(1024, -57344, 57344) = 1024
```

## Root cause

Both legacy CUDA kernels use a hard-coded ±448 clamp even though the output type is available as a template parameter.

## Scope

Affected:

- Legacy CUDA `rmsnorm_quant`
- Legacy CUDA `fused_add_rmsnorm_quant`
- E5M2 results above 448

Not affected:

- E4M3 output
- INT8
- The default CuTe implementation

## Success criteria

- Both kernels select the limit from the output FP8 type.
- E4M3 continues to saturate at 448.
- E5M2 preserves representable values above 448.
- Both APIs and formats receive regression coverage.

morluto:
## 📌 Description

The legacy CUDA implementations of `rmsnorm_quant` and `fused_add_rmsnorm_quant` clamped every FP8 output to ±448. That bound is correct for E4M3 but not E5M2, whose finite range extends to ±57344.

| Output type | Finite maximum | Input to clamp | Previous output | Correct output |
|---|---:|---:|---:|---:|
| E4M3 | 448 | 1024 | 448 | 448 |
| E5M2 | 57344 | 1024 | 448 | 1024 |

```text
previous = clamp(1024, -448, 448) = 448
correct  = clamp(1024, -57344, 57344) = 1024
```

This change adds a compile-time FP8 range helper and applies the selected output type in both kernels:

- E4M3 remains clamped to ±448.
- E5M2 is clamped to ±57344.
- Unsupported output template types fail at compile time.

Public regression tests use 1024, a value between the formats' maximum ranges, and cover both APIs with E4M3 and E5M2.

The CuTe DSL implementation, backend routing, and INT8 behavior are unchanged.

## 🔍 Related Issues

Fixes #4053.

## 🧪 Tests

- Added E4M3 and E5M2 regressions for `rmsnorm_quant` and `fused_add_rmsnorm_quant`
- Focused pre-commit hooks passed
- `python3 -m compileall -q tests/utils/test_norm.py`
- `git diff --check` passed

## Reviewer Notes

The core invariant is in `fp8_max<T>()`: only CUDA E4M3 and E5M2 output types are accepted, and both kernels use the selected value symmetrically.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated FP8 quantized RMSNorm and fused add+RMSNorm to clamp outputs using the selected FP8 format’s maximum finite magnitude (instead of a fixed range), preventing out-of-range values.
* **Tests**
  * Added CUDA-gated regression tests validating correct FP8 clamping for both FP8 e4m3 and e5m2 output types.

### DIFF
--- a/include/flashinfer/norm.cuh
+++ b/include/flashinfer/norm.cuh
@@ -18,6 +18,7 @@
 
 #include <cstdint>
 #include <numeric>
+#include <type_traits>
 
 #include "flashinfer/trtllm/common/cudaTypeUtils.cuh"
 #include "flashinfer/trtllm/common/cudaUtils.h"
@@ -32,6 +33,12 @@ namespace flashinfer {
 namespace norm {
 
 using namespace tensorrt_llm::common;
+
+template <typename T>
+constexpr float fp8_max() {
+  static_assert(std::is_same<T, __nv_fp8_e4m3>::value || std::is_same<T, __nv_fp8_e5m2>::value);
+  return std::is_same<T, __nv_fp8_e5m2>::value ? 57344.0f : 448.0f;
+}
 
 template <uint32_t VEC_SIZE, typename T>
 __global__ void RMSNormKernel(T* __restrict__ input, T* __restrict__ weight, T* __restrict__ output,
@@ -159,6 +166,7 @@ __global__ void RMSNormQuantKernel(T* __restrict__ input, T* __restrict__ weight
   const uint32_t num_threads = num_warps * warp_size;
   const uint32_t rounds = ceil_div(d, VEC_SIZE * num_threads);
   const float scale_inv = 1.0f / scale[0];
+  constexpr float output_max = fp8_max<O>();
   extern __shared__ float smem[];
 
   float sum_sq = 0.f;
@@ -214,7 +222,7 @@ __global__ void RMSNormQuantKernel(T* __restrict__ input, T* __restrict__ weight
     for (uint32_t j = 0; j < VEC_SIZE; j++) {
       output_vec[j] =
           float(input_vec[j]) * rms_rcp * (weight_bias + float(weight_vec[j])) * scale_inv;
-      output_vec[j] = fmaxf(-448.0f, fminf(output_vec[j], 448.0f));
+      output_vec[j] = fmaxf(-output_max, fminf(output_vec[j], output_max));
     }
     if ((i * num_threads + thread_id) * VEC_SIZE < d) {
       output_vec.cast_store(output + bx * stride_output + i * num_threads * VEC_SIZE +
@@ -528,6 +536,7 @@ __global__ void FusedAddRMSNormQuantKernel(T* __restrict__ input, T* __restrict_
   const uint32_t num_threads = num_warps * warp_size;
   const uint32_t rounds = ceil_div(d, VEC_SIZE * num_threads);
   const float scale_inv = 1.0f / scale[0];
+  constexpr float output_max = fp8_max<O>();
   extern __shared__ float smem[];
   float* smem_x = smem + ceil_div(num_warps, 4) * 4;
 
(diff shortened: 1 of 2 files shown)


lazypool (commented):
Reviewed the diff against `main`. The core fix is correct: E5M2's finite max is 57344 (= 1.75 × 2^15) and E4M3's is 448, and since the launcher dispatches `O` only via `DISPATCH_DLPACK_DTYPE_TO_CTYPE_FP8` (`csrc/tvm_ffi_utils.h`, e4m3/e5m2 only), the `fp8_max<O>()` static_assert cannot break other output types (e.g. int8) — no dispatch path instantiates the kernel with them today.

Two concerns before this lands:

1. **Tests never run in the default CI configuration.** `_USE_CUDA_NORM` is opt-in (`FLASHINFER_USE_CUDA_NORM=1`, `flashinfer/norm/__init__.py:60`), so both new tests `pytest.skip` on the default CuTe-DSL backend. That means the clamp fix has zero regression protection in default CI — the exact situation that let this bug rot in the first place. Consider dropping the skip and asserting both backends, or at least gating on "CUDA backend compiled" rather than "CUDA backend active".

2. **The issue's user-facing repro is still broken on the default backend.** On main, `rmsnorm_quant(out=e5m2)` on the default CuTe-DSL path raises `KeyError` (no `float8_e5m2` entry in the norm dtype map), so the API still fails for E5M2 even though the CUDA clamp is now right. Worth either explicitly scoping #4053 to the CUDA backend, or noting that the default backend still needs a follow-up (PR #4202 covers this).

Minor: `fp8_max<T>()` duplicates the concept of `QuantTypeStaticVals<T>::MAX_VAL` already in this file (lines 724-745, used by `layernorm_quant`). Reusing that trait would be a smaller, more consistent diff.

Also note this PR and #4202 edit the exact same two clamp lines in the same kernels — please coordinate so the second one to land rebases cleanly.

lazypool (commented):
Went through the diff and the fix itself looks right — 57344 for E5M2 checks out (1.75 * 2^15), and the static_assert shouldn't trip on anything else since the launcher only ever dispatches fp8 types for these kernels.

Two things I'd like to see addressed though:

1. The new tests skip whenever the CuTe-DSL backend is active, which is the default (FLASHINFER_USE_CUDA_NORM is opt-in). So in normal CI they'll just be skipped forever and the fix won't have any regression protection — that's basically how this bug stayed hidden in the first place. Could the tests run on both backends instead?

2. Even with this fix, the repro from the issue still fails on the default backend: rmsnorm_quant with an e5m2 output raises KeyError there (no entry in the norm dtype map). So users on the default path are still broken. Fine if the issue is scoped to the CUDA backend only, but then it should say so explicitly. FWIW #4202 handles the DSL side.

Minor thing: instead of introducing a new fp8_max helper, the file already has QuantTypeStaticVals with a MAX_VAL per type (layernorm_quant uses it) — reusing that would keep the change smaller.

Heads up: #4202 touches the exact same clamp lines, so whichever of the two lands second will need a rebase.

bkryu:
Thanks @morluto, I expect #4202 to supersede this PR by implementing fixes to both the CUDA and CuTe-DSL kernel. 

#4202 is approved and set to auto-merge so I will close this PR. Please feel free to reopen if you see a need.

PR #4067, opened 2026-07-20
perf(moe): phase NVFP4 dispatch payload copies at EP4/EP8/EP16

samnordmann:
## Summary

Follow-up to #3846. The generic dispatch kernel makes every CTA thread build remote destination pointers for every payload before determining whether that thread owns any bytes. NVFP4 has four predictably asymmetric payloads, so this change gives each payload only the workers it needs:

- two warps for packed activations;
- the required lanes of one warp for scales;
- two lane groups in the final warp for expert IDs and routing weights.

EP4 retains its compact per-rank destination list. EP8 and EP16 retain the existing top-k-sized representation; only their copy schedule changes. The phased path is gated to EP `{4, 8, 16}`, a 128-thread CTA, and validated four-payload NVFP4 layouts. Unsupported EP sizes, top-k values, widths, and payload layouts use the existing generic kernel.

The H2048/top-k-22 target retains compile-time payload widths. Validated smaller layouts use the same schedule with runtime activation and scale widths.

### Single op

GB200, H2048, top-k 22, CUDA graphs. EP8 and EP16 are true global expert-parallel groups spanning two and four nodes.

| Global EP | Nodes | Tokens/rank | Baseline | Candidate | Delta |
|---:|---:|---:|---:|---:|---:|
| 4 | 1 | 1 | 13.480 us | 12.862 us | -4.58% |
| 4 | 1 | 4 | 14.408 us | 13.490 us | -6.37% |
| 4 | 1 | 64 | 16.511 us | 15.284 us | -7.43% |
| 4 | 1 | 512 | 18.155 us | 17.804 us | -1.93% |
| 4 | 1 | 8192 | 55.072 us | 54.689 us | -0.70% |
| 8 | 2 | 1 | 16.270 us | 13.877 us | -14.71% |
| 8 | 2 | 4 | 18.394 us | 14.679 us | -20.20% |
| 8 | 2 | 64 | 21.253 us | 17.358 us | -18.33% |
| 8 | 2 | 512 | 40.520 us | 27.275 us | -32.69% |
| 8 | 2 | 1024 | 64.713 us | 37.797 us | -41.59% |
| 8 | 2 | 8192 | 308.292 us | 96.494 us | -68.70% |
| 16 | 4 | 1 | 18.107 us | 15.733 us | -13.11% |
| 16 | 4 | 4 | 21.639 us | 17.328 us | -19.92% |
| 16 | 4 | 64 | 25.264 us | 20.652 us | -18.26% |
| 16 | 4 | 512 | 55.275 us | 41.484 us | -24.95% |
| 16 | 4 | 1024 | 100.321 us | 68.156 us | -32.06% |
| 16 | 4 | 8192 | 373.115 us | 176.750 us | -52.63% |

### E2e workload

On the **Nemotron Ultra NVFP4 end-to-end** workload (2 GB200 nodes, TP1/DP8/EP8, FlashInfer one-sided, CUDA graphs, ISL/OSL 8192/8, concurrency 8), mean TTFT improves from 860.8 ms to 839.9 ms (**-2.43%**, 95% CI [-3.44%, -1.41%]). Throughput (+0.62%) and TPOT are statistically on par.

## Summary by CodeRabbit

* **New Features**
  * Added support for additional phased NVFP4 layouts in Mixture-of-Experts all-to-all dispatch.
  * Added a specialized H2048 phased dispatch path, including top‑k=22 configurations.
* **Performance**
  * Improved dispatch efficiency by optimizing activation, scale, and routing-metadata movement.
* **Bug Fixes**
  * Strengthened validation to ensure payload layouts match runtime settings.
* **Tests**
  * Updated NVFP4 payload generation and expanded phased H2048 coverage across supported configurations.

### DIFF
--- a/csrc/nv_internal/tensorrt_llm/kernels/communicationKernels/moeAlltoAllKernels.cu
+++ b/csrc/nv_internal/tensorrt_llm/kernels/communicationKernels/moeAlltoAllKernels.cu
@@ -425,9 +425,9 @@ __global__ void moeA2ADispatchKernel(
     int local_num_tokens, int rank_id, int ep_size, int num_experts, int eplb_stats_num_experts,
     bool enable_pdl) {
   static_assert(!COMPACT_EP4 || TOP_K > kEp4Size);
-  static_assert(!PHASE_H2048_NVFP4 || (COMPACT_EP4 && TOP_K == 22));
+  static_assert(!PHASE_H2048_NVFP4 || TOP_K == 22);
   static_assert(!PHASE_NVFP4_PAYLOADS ||
-                (COMPACT_EP4 && TOP_K % 2 == 0 &&
+                (TOP_K > kEp4Size && TOP_K % 2 == 0 &&
                  TOP_K * static_cast<int>(sizeof(int32_t)) / kDispatchMetadataVectorBytes <=
                      kDispatchHalfWarpSize));
   static_assert(!PHASE_H2048_NVFP4 || !PHASE_NVFP4_PAYLOADS);
@@ -791,15 +791,20 @@ void moe_a2a_dispatch_launch(MoeA2ADispatchParams const& params) {
       max_payload_bytes_per_token = kernel_ptrs.payload_bytes_per_token[i];
     }
   }
+  bool const supports_phased_nvfp4 =
+      params.ep_size == kEp4Size || params.ep_size == 8 || params.ep_size == 16;
+  bool const phased_nvfp4_layout =
+      supports_phased_nvfp4 &&
+      isPhasedNvfp4PayloadLayout(kernel_ptrs, params.num_payloads, params.top_k);
+
   // Small payloads do not have enough 16-byte vectors to use larger CTAs. A GB200
   // sweep found 128 threads faster than both 64 and 256 at prefill token counts.
-  if (use_compact_ep4 && max_payload_bytes_per_token <= kCompactDispatchMaxPayloadBytes &&
+  if ((use_compact_ep4 || phased_nvfp4_layout) &&
+      max_payload_bytes_per_token <= kCompactDispatchMaxPayloadBytes &&
       block_size > kCompactDispatchBlockSize) {
     block_size = kCompactDispatchBlockSize;
   }
-  bool const phase_nvfp4_payloads =
-      use_compact_ep4 && block_size == kCompactDispatchBlockSize &&
-      isPhasedNvfp4PayloadLayout(kernel_ptrs, params.num_payloads, params.top_k);
+  bool const phase_nvfp4_payloads = phased_nvfp4_layout && block_size == kCompactDispatchBlockSize;
 
   // Configure kernel launch: one block per token
   int grid_size = params.local_num_tokens;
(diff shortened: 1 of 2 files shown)


samnordmann:
cc @aleozlx for review

csrc/nv_internal/tensorrt_llm/kernels/communicationKernels/moeAlltoAllKernels.cu
```diff
@@ -48,6 +65,30 @@ __host__ __device__ inline T ceilDiv(T m, T n) {
...
+static bool isPhasedNvfp4PayloadLayout(DispatchKernelPointers const& ptrs, int num_payloads,
+                                       int top_k) {
+  if (num_payloads != 4 || top_k <= kEp4Size) {
+    return false;
+  }
+
+  int const activation_bytes = ptrs.payload_bytes_per_token[0];
+  int const scale_bytes = ptrs.payload_bytes_per_token[1];
+  int const metadata_bytes = top_k * static_cast<int>(sizeof(int32_t));
+  bool const validated_top_k_layout = top_k < 22 || activation_bytes == kNvfp4H2048ActivationBytes;
```
shaharmor98, line 82: For ultra, that makes perfect sense. However, I think we can safely expand it to support Nemotron-Super-3.5VL (which has a 1024 MoE hidden size and not 2048 as Ultra).
If that's fine, I'll open a subsequent PR once this one is merged to support it.

csrc/nv_internal/tensorrt_llm/kernels/communicationKernels/moeAlltoAllKernels.cu
```diff
@@ -314,8 +355,8 @@ __device__ void vectorized_dispatch_impl(uint8_t const* src_ptr, int bytes_per_t
   }
 
   // TODO: process all payloads. index could be reused.
```
shaharmor98: NIT - I might be wrong, but at first I was confused by this comment, since now we're technically not letting all threads process the payloads. Perhaps I'd adjust that comment. Feel free to ignore.

csrc/nv_internal/tensorrt_llm/kernels/communicationKernels/moeAlltoAllKernels.cu
```diff
@@ -684,12 +799,20 @@ void moe_a2a_dispatch_launch(MoeA2ADispatchParams const& params) {
       max_payload_bytes_per_token = kernel_ptrs.payload_bytes_per_token[i];
     }
   }
+  bool const supports_phased_nvfp4 =
+      params.ep_size == kEp4Size || params.ep_size == 8 || params.ep_size == 16;
```
shaharmor98, line 806: Curious to know if you also tried EP2?
Perhaps not for Ultra, but definitely for other models (e.g., Super-3.5VL), EP2 is also used.
If that's fine, I'll open a subsequent PR once this one is merged to support it.

tests/comm/test_mnnvl_moe_alltoall.py
```diff
@@ -709,6 +714,11 @@ def test_moe_a2a_dispatch(distribution, top_k, use_lora):
     safe_run(moe_a2a_dispatch_test_impl, distribution, top_k, use_lora)
 
 
+def test_moe_a2a_dispatch_phased_h2048_nvfp4():
```
shaharmor98, line 718: IIUC, this test validates the dispatch results but does not prove that the specialized kernel executes, which is the goal of this test.
AFAIK, safe_run only propagates MPI failures, and ep_size comes from the external MPI world; under the file's documented -n 2 command, this runs the generic fallback.
I do think that a test ensuring the correct phased path is selected is important.
WDYT?

tests/comm/test_mnnvl_moe_alltoall.py
```diff
@@ -207,8 +207,11 @@ def make_nvfp4_payloads(
     token_selected_experts: torch.Tensor,
     lora_ids: torch.Tensor | None = None,
 ) -> tuple[list, int]:
-    """Create the four NV FP4 payloads exactly as in single-GPU test, with an
-    optional 5th LoRA adapter ID payload."""
```
shaharmor98, line 211: I think this was accidentally deleted, as the helper still appends it and this still seems relevant.

shaharmor98 (approved):
LGTM. 
Left a few small comments

csrc/nv_internal/tensorrt_llm/kernels/communicationKernels/moeAlltoAllKernels.cu
samnordmann: Yes. H1024 is already eligible through the general runtime-width phased path for even top-k values below 22 at EP4/EP8/EP16; only the top-k-22 specialization is restricted to H2048. I clarified that distinction in the selector comment. Explicit Super H1024/EP2 coverage and measurements still make sense as a follow-up.

csrc/nv_internal/tensorrt_llm/kernels/communicationKernels/moeAlltoAllKernels.cu
samnordmann: I did not measure EP2, so I kept it out of this Ultra-focused PR. At EP2, simply enabling the current phased path may also leave avoidable overhead because it retains the top-k-sized destination representation despite at most two unique ranks.

tests/comm/test_mnnvl_moe_alltoall.py
samnordmann: Yes, this was accidentally dropped. fixed

tests/comm/test_mnnvl_moe_alltoall.py
samnordmann: Agreed.the previous test could pass through the EP2 generic fallback. It now skips unless the MPI world size is EP4/EP8/EP16, and the documented invocation is corrected to `mpirun -n 4`, so the H2048/top-k-22 case exercises the specialized branch. I also validated the updated test at EP4 with a 128-thread CTA

csrc/nv_internal/tensorrt_llm/kernels/communicationKernels/moeAlltoAllKernels.cu
samnordmann: Super H1024/EP2 follow-up would make much sense!
Just to clarify: only the top-k-22 specialization is restricted to 2048. I clarified this in comment.

csrc/nv_internal/tensorrt_llm/kernels/communicationKernels/moeAlltoAllKernels.cu
samnordmann: agreed. fixed the comment

aleozlx (approved):

PR #4068, opened 2026-07-20
perf(moe): avoid copying batched GEMM configs

DongyunZou:
# PR title

`perf(moe): avoid copies in TRT-LLM batched GEMM config scan`

# PR body

## 📌 Description

The TRT-LLM batched GEMM config scan copied both the config and its options on every iteration. This PR changes exactly those two bindings to read-only references:

```diff
-    auto const config = configs[i];
-    auto const options = config.mOptions;
+    auto const& config = configs[i];
+    auto const& options = config.mOptions;
```

The scan now reads the existing objects instead of copying them. Config filtering, the selected config, launch order, and GPU kernels are unchanged.

### Performance

Measured before and after the patch on an NVIDIA B200 (SM100) with an Intel Xeon 6960P host, driver 580.126.09, CUDA toolkit 13.2, and PyTorch 2.12.1+cu130.

Both revisions used the same in-tree one-token benchmark command, synthetic inputs, warm JIT caches, cold L2 cache, eager execution, and CUPTI timing with no CUDA-event fallback. Five independent processes were run per revision in alternating order; the table reports the median of their per-process medians. This CUPTI path measures the span from the first GPU activity to the last, so host gaps between launches are included.

| Revision | Median latency | Process range | Speedup |
|---|---:|---:|---:|
| Before patch (`f212ec82`) | 1.162 ms | 1.042–1.206 ms | 1.00x |
| After patch (`1b53b1cb`) | 0.398 ms | 0.359–0.410 ms | **2.92x** |

Each process used 3 dry runs and 30 measured iterations. All ten runs completed with `refcheck=True` and `use_cupti=True`.

Reproduction command:

```bash
python3 benchmarks/flashinfer_benchmark.py \
  --routine trtllm_fp8_block_scale_moe \
  --num_tokens 1 \
  --hidden_size 7168 \
  --intermediate_size 2048 \
  --num_experts 32 \
  --local_num_experts 32 \
  --top_k 8 \
  --n_group 8 \
  --topk_group 4 \
  --routing_method deepseek_v3 \
  --use_routing_bias \
  --no_cuda_graph \
  --refcheck \
  --dry_run_iters 3 \
  --num_iters 30
```

## 🔍 Related Issues

The copies were introduced in #3239.

### ✅ Pre-commit Checks

## 🧪 Tests

No unit test was added because this change only replaces two local copies with `const` references. The benchmark refcheck passed at both revisions, and `pre-commit run --all-files` passed at the PR head.

## Reviewer Notes

`configs` remains alive for the entire scan, so both references are valid. This PR does not cache or reorder configs.

## Summary by CodeRabbit

* **Performance**
  * Reduced unnecessary copying during batched matrix multiplication configuration selection, with no expected change to user-visible behavior.


Ubospica:
@IwakuraRein @yongwww could you take a look? Looks like a cpu overhead issue and will make MOE much slower.

Issue #3780, opened 2026-06-30
[sm90] Add support for MegaMOE in FlashInfer


mhoqueanik:
MegaMOE has potential to speedup a number of customer workloads. We should support the following data types:

- NVFP4
- MXFP8
- FP8 BlockScale
- BF16

Ideally we should make sure the MOE backend is adapted so that we can naturally use MegaMOE in downstream frameworks without needing a secondary integration path specific to MegaMOE

**Important** Please read the kernel API guidelines described here: https://github.com/flashinfer-ai/flashinfer/issues/3704

mhoqueanik:
[Pull-based dispatch] Update from Vincent Zhang: Non swapA/B version is functional ready, still working on the performance optimization
[Push-based dispatch] leonardHONG started working on it. Tentative timeline for alpha version: mid-Aug.

inocsin:
Per-tensor + blockwise with swapAB and non-swapAB is functional ready, still working on the performance

leonardHONG:
Hi! I have two API-shape questions from the SM90 push-based MegaMoE work before I turn the prototype into integration PRs.

1. **Kernel shape.** On Hopper, the best path I have is split behind the API: push dispatch into `symm_buffer`, in-tree grouped FP8 GEMM for FC1/FC2, then push combine. I tried a single fused megakernel first, but it was much slower on H800 because the existing grouped GEMM is far stronger on SM90. Would it still fit the intent of `mega_kernel(output, l1_weight, l2_weight, symm_buffer)` if that entry point orchestrates a split sequence internally, or is a single device kernel required?

2. **Activation pack.** The MVP `MoEActivationPack` looks NVFP4-shaped, with packed `uint8` hidden states. For the SM90 FP8-blockscale path, I’d prefer to pass bf16 hidden states and quantize to 1×128 FP8 inside the dispatch/prep step while staging into `symm_buffer`. Would you prefer adding an FP8-blockscale activation-pack variant aligned with `QuantVariant.DeepSeekFp8`, or should SM90 define its own activation view?

Thanks — the proposed API split has been useful for making the SM90 path fit the existing MoE integration.

mhoqueanik:
Hi @leonardHONG,

Glad to see the SM90 MegaMoE work is making good progress, and thanks for driving it!

[...]

Issue #3704, opened 2026-06-23
RFC: Adopt a Standard MegaMoE Kernel Integration API


mhoqueanik:
# MegaMoE Kernel Integration API

## Motivation

As we add MegaMoE support to FlashInfer, we need a standard interface between FlashInfer and MegaMoE kernels. Without one, each backend integrated into `moe_ep` would need to define its own preprocessing, workspace management, and kernel invocation interfaces.

We propose introducing a standardized **MegaMoE Kernel Integration API**.

## FlashInfer MegaMoE API

```
FlashInfer MoE EP Layer [Python]
           |
FlashInfer MegaMoE API [Python]
           |
MegaMoE Kernel Integration API [Python]
           |
       Mega Kernel [CUDA/CuteDSL/...]
```

## Design

A MegaMoE implementation consists of three components:

### 1. Data Preparation Utilities

- During initialization, there are functions that transform weights into formats required by the kernel. 
- During forward(), input tokens are quantized (typically from bf16) and placed into symm_buffer. It is recommended to fuse them into a single kernel (see [example](https://github.com/vllm-project/vllm/blob/7a341fa109e569c8d265545adfca756f42451244/vllm/models/deepseek_v4/xpu/model.py#L529)).

### 2. Scratch Space (`symm_buffer`)

A symm_buff workspace containing prepared inputs and storage for intermediate kernel results.

### 3. Kernel Interface

The kernel operates on preprocessed inputs:

```python
mega_kernel(
    output_tensor,
    l1_weight,
    l2_weight,
    symm_buffer,
)
```

The integration API owns preprocessing and workspace management; the kernel focuses solely on computation.

## Example: DeepGEMM SM100 MegaMoE

DeepGEMM's SM100 MegaMoE implementation already follows this structure:

| Function | Purpose |
|----------|---------|
| `per_token_cast_to_fp8` | Activation quantization |
| `per_token_cast_to_fp4` | Weight quantization |
| `transform_sf_into_required_layout` | Scale tensor layout transformation |
| `get_symm_buffer_for_mega_moe` | Scratch-space allocation |
| `transform_weights_for_mega_moe` | Weight layout transformation |
| `fp8_fp4_mega_moe` | MegaMoE kernel |
| `_stage_mega_moe_inputs`* | quant+place input tokens into symm_buff |

[...]

Issue #3692, opened 2026-06-22
Add support for MegaMOE in FlashInfer


djns99:
MegaMOE has potential to speedup a number of customer workloads. We should support the following data types:

- [ ] NVFP4
- [ ] MXFP8
- [ ] FP8 BlockScale
- [ ] BF16

Ideally we should make sure the MOE backend is adapted so that we can naturally use MegaMOE in downstream frameworks without needing a secondary integration path specific to MegaMOE

See also https://github.com/flashinfer-ai/flashinfer/issues/3196 https://github.com/flashinfer-ai/flashinfer/issues/3346

leonardHONG:
Hi,I'd be happy to take a first slice of this.

I have Hopper + NVLink access, so I can validate this end to end.

I was thinking of starting with DeepGEMM's existing FP8×FP4 Hopper MegaMOE path, exposing it through JitSpec / TVM-FFI / Python, and wiring it into the existing MoE backend instead of creating a separate path.

Does that sound like the right direction?

YangXu1990uiuc:
Hi @leonardHONG it's great you are interested in working on the hopper path! supporting NVFP4 or MXFP8 might be a bit challenging on Hopper since there is no native tensor core support for those.
Separately, @mhoqueanik will be working on the blackwell path for additional dtype coverage

mhoqueanik:
Hi @leonardHONG ! Thanks again for helping with the MegaMoE kernel development.

We just opened an RFC describing the proposed MegaMoE Kernel Integration API and how we envision integrating new kernels into FlashInfer. It may be useful to take a quick look while you're working on the Hopper kernel: https://github.com/flashinfer-ai/flashinfer/issues/3704

The API is still being discussed, so feel free to share feedback if anything seems awkward or if you see a better integration pattern based on your implementation experience.

leonardHONG:
Thanks both. I'll align the Hopper side with the #3704 API — the prep-utils / symm_buffer / kernel-interface split looks like a clean shape to follow for SM90.

One correction after checking DeepGEMM more carefully: I missed that its MegaMoE kernels are SM100-only (`sm100_fp8_fp4_mega_moe` / `sm100_bf16_mega_moe`), built around tcgen05 / TMEM / native FP8×FP4 MMA. The SM90 side seems to be GEMM-only, so there isn't an existing Hopper MegaMoE path to wrap.

[...]

PR #4069, opened 2026-07-20
feat(moe_ep): add SM90 push FP8 mega-MoE backend for Hopper

leonardHONG:
## 📌 Description

This PR registers `sm90_push_fp8`, a whole-layer expert-parallel MoE backend for SM90 GPUs connected through single-node NVLink, as a fourth mega backend behind the existing `register_mega_kernel` plugin surface.

The public entry point is `MoEEpLayer` with `Sm90PushFp8MegaMoeConfig`. Existing backend behavior is unchanged.

The steady-state forward path covers dispatch, FP8 block-scale expert FFN, and combine without host-side synchronization, and supports CUDA Graph capture after initialization. The diagram below shows the path with all three optional optimizations enabled:

```text
bump_tag
  -> wait_acks
  -> deduplicated dispatch
  -> wait_prefix
  -> compact
  -> FC1 with fused SwiGLU + 1x128 quantization
  -> FC2
  -> grouped combine with owner-side FP32 route reduction
  -> BF16 reduction of received owner rows
  -> ack
```

The backend provides three independently configurable optimizations. Validation and performance results below use all three unless stated otherwise.

* **Deduplicated dispatch** stores one payload row per token and destination rank instead of one per route. In the tested configurations, with all other options fixed, its output matched the per-route path under `torch.equal`. For `K=6`, `EP=4`, and random routing, it reduces dispatch payload bytes by approximately 43%.
* **Grouped combine** performs FP32 route reduction on the owner rank before pushing one quantized row per token. Its correctness gate compares the final output against the same reference used by the per-route path because the quantization point moves.
* **Fused FC1 epilogue** performs SwiGLU and 1x128 activation quantization inside the FC1 DeepGEMM epilogue, eliminating an approximately 1 GiB intermediate BF16 activation buffer per rank at the DSV3 EP8 shape. In the tested configurations, with all other options fixed, its output matched the unfused path under `torch.equal`. The unfused path remains the default and maintained reference; the fused path is opt-in.

## Motivation

The existing Hopper EP stack composes NCCL all-to-all with a local fused-MoE operator, paying for additional launches, intermediate buffers, and quantization boundaries.

`sm90_push_fp8` writes quantized payloads directly into peer-mapped symmetric memory and feeds received rows directly into the expert GEMM pipeline.

## Contracts introduced by this PR

* `capacity_factor` bounds the GEMM, TMA, and scale buffers, rather than only the protocol window. The workspace query takes explicit `expected_m` and `max_rows`, so block-M tactics and cubin keys remain independent of the memory bound.
* Building the grouped GEMM requires CUDA Toolkit 12.8 or newer. The generator explicitly rejects older toolkits, and AOT skips this module below CUDA 12.8, allowing JIT-cache wheels to continue building on CUDA 12.6. Loading and running the generated module also requires a CUDA runtime version 12.8 or newer.
* Runner initialization elects one NVCC-capable rank per cache and tactic group to compile cold DeepGEMM cubins. Other ranks load the resulting cubins from disk.

  * A warm cache works without NVCC.
  * A cold cache without NVCC fails during initialization with an error explaining how to select the fallback.
  * `TRTLLM_DG_ENABLED=0` selects the fixed-tactic CUTLASS fallback on the unfused path.
* Wait kernels use a `%globaltimer` deadline and publish a shared abort marker before trapping. Peers polling the same protocol window observe the marker and fail. Because a trap leaves a sticky CUDA error, the process launcher must terminate the full rank group from the CPU.
* This backend is a stateful variant of the mega-kernel contract:

  * construction binds transformed static weights;
  * `stage_inputs` pre-binds the caller's output tensor.

  The architecture and runbook documentation are updated accordingly, including narrowing the two-entry wording to buffer-oriented kernels. The `MegaKernelBackend` lifecycle itself is unchanged.

## Correctness

Full kernel acceptance ran on an 8× H800 NVLink node:

| Suite                                                     |           Scale | Result                       |
| --------------------------------------------------------- | --------------: | ---------------------------- |
| Single-GPU suite, 105 configurations                       |           1 GPU | all passed                   |
| Distributed suites                                         | EP2 / EP4 / EP8 | all passed                   |
| Uneven token distributions, zero-token ranks, launch skew  |         EP2–EP8 | all passed                   |
| CUDA Graph capture and replay                              | tested configurations | all passed            |
| 200-round oracle-checked soaks                             | every tested configuration | all passed        |
| Eight-GPU acceptance matrix at DSV3 shapes                 |             EP8 | all passed, no skipped cases |

The final PR tree was then revalidated across three Hopper variants:

| Suite                                                                          | Hardware       |               Result |
| ------------------------------------------------------------------------------ | -------------- | -------------------: |
| Single-GPU suite, all configurations                                           | 1× H20         | 130 passed, 0 failed |
| `torchrun` EP2                                                                 | 2× H100 NVLink |  15 passed, 0 failed |
| `torchrun` EP4                                                                 | 4× H20 NVLink  |  15 passed, 0 failed |
| `compute-sanitizer` memcheck / racecheck / initcheck, excluding trap and soaks | 1× H20         |                Clean |

Additional hardware-verified properties on the final PR tree:

* In the tested configurations, the fused FC1 output matched the unfused reference under `torch.equal`.
* Leader deduplication was verified by counting NVCC invocations against a shared cold cache: exactly one compilation per cubin across four ranks.
* Warm-cache execution without NVCC was exercised end to end.
* Cold-cache fail-fast behavior was exercised end to end.
* The CUTLASS fallback was exercised end to end.

## Performance

Performance numbers were collected during kernel acceptance on the 8× H800 node. The final PR tree was functionally revalidated as listed above.

### Setup

* H800 NVLink
* CUDA 12.9
* PyTorch 2.8
* barrier-aligned timing
* maximum latency across ranks
* autotuned baseline tactics
* identical correctness gates on both sides
* `T` denotes live tokens per rank

The tables report `baseline latency / sm90_push_fp8 latency`; values greater than `1×` favor this backend. `H`, `E`, and `K` denote hidden size, expert count, and top-k routes. `random` distributes routes across experts, while `hot1` concentrates routes on one expert.

### Versus NCCL all-to-all with the same FP8 block-scale compute

| Configuration           | Decode T64 | Throughput T2048 |
| ----------------------- | ---------: | ---------------: |
| EP4 SMALL, H4096 E32 K6 |   1.1–1.5× |         2.0–2.3× |
| EP4 DSV3, H7168 E256 K8 |      1.01× |         2.0–2.3× |

Against a wire-only NCCL transport reference that excludes compaction and combine reduction, the push transport measured **1.9–3.2×** faster.

This number should be interpreted only as a comparison against that specific measured reference. It is not a claim against third-party EP libraries.

### Versus NCCL + `cutlass_fused_moe` FP8

| Configuration | Decode T64 | T2048 random |      T2048 hot1 |
| ------------- | ---------: | -----------: | --------------: |
| EP4 DSV3      |      1.06× |         2.3× |            2.0× |
| EP8 DSV3      |      1.10× |     2.2–2.4× |            1.6× |
| EP4 SMALL     |       3.1× |         2.1× | See limitations |

At EP1 and equal FP8 precision, `sm90_push_fp8` measured **2.3–3.4×** faster than `cutlass_fused_moe` over `T64–T2048`.

### EP8 scaling

Configuration:

```text
DSV3
capacity = 2048
all three optimizations enabled
```

Results:

```text
T64:    0.758 ms
T2048:  3.238 ms
```

With capacity fixed at 2048, a 32× increase in live token count increased latency by 4.27×.

Additional observations:

* no scheduling cliff near `T128`;
* all-remote routing adds approximately 5%;
* running 64 live tokens with buffers sized for 2048 adds approximately 0.7%.

### Comparison with Pull Style

[From comment]

Ran the six suggested shapes on 8× H100 80GB SXM. These are back-to-back steady-state measurements with 20 warmups and 100 samples per path. The table reports p50 latency in milliseconds; the last column is Pull / Push, so values above 1 mean Push is faster.

| Geometry | Tokens/rank | Pull | Push | Speedup |
|---|---:|---:|---:|---:|
| DeepSeek V3 | 8 | 0.572 | 0.614 | 0.93× |
|  | 64 | 0.657 | 0.704 | 0.93× |
|  | 512 | 1.699 | 1.065 | 1.60× |
|  | 2048 | 3.497 | 2.728 | 1.28× |
|  | 8192 | 12.935 | 9.886 | 1.31× |
| Kimi K2.6 | 8 | 0.716 | 0.749 | 0.96× |
|  | 64 | 0.913 | 0.949 | 0.96× |
|  | 512 | 2.038 | 1.163 | 1.75× |
|  | 2048 | 3.867 | 2.776 | 1.39× |
|  | 8192 | 13.023 | 10.046 | 1.30× |
| DeepSeek V4 Flash | 8 | 0.386 | 0.410 | 0.94× |
|  | 64 | 0.436 | 0.468 | 0.93× |
|  | 512 | 0.871 | 0.581 | 1.50× |
|  | 2048 | 1.960 | 1.410 | 1.39× |
|  | 8192 | 5.752 | 4.609 | 1.25× |
| DeepSeek V4 Pro | 8 | 0.951 | 0.966 | 0.98× |
|  | 64 | 1.301 | 1.294 | 1.01× |
|  | 512 | 2.660 | 1.479 | 1.80× |
|  | 2048 | 4.315 | 3.168 | 1.36× |
|  | 8192 | 14.172 | 10.757 | 1.32× |

Push is about 4–5% slower at the smallest token counts, but becomes consistently faster from T512 onward, with gains of roughly 1.25–1.80×.

Correctness also passed across all measured cases (minimum cosine similarity 0.99816, maximum NRMSE 0.06063).

GPT-OSS 120B is not supported yet because its hidden dimensions are not 128-aligned, and Qwen3.5 397B uses top-k 10 while the current limit is 8.

## Limitations

* Supports:

  * SM90;
  * single-node, peer-accessible NVLink;
  * protocol limit `ep_size <= 32`, with hardware validation in this PR covering up to EP8;
  * `top_k` in `{1, 2, 4, 6, 8}`;
  * DeepSeek-style FP8 block scaling;
  * SwiGLU.

  Unsupported configurations raise an explicit error rather than silently falling back.

* The unfused FC1 path supports `intermediate_size <= 16384`; larger configurations require `fuse_fc1_epilogue=True`.

* At DSV3 decode shapes, the advantage over NCCL + `cutlass_fused_moe` is **1.06–1.10×**, because computation dominates.

* In the artificial SMALL `T2048` case where all `K` routes target one expert, the backend is 25% slower than the autotuned CUTLASS stack. Realistic `hot1` routing remains faster at EP4 and EP8.

* The fused FC1 epilogue does not yet have a small-M `swapAB` variant and can be disabled independently below its break-even point.

* Round tags use `uint32`. Reuse after `2^32` forwards on one pipe is a documented limit.

* Custom raw-stream execution through `bootstrap.stream` is rejected explicitly rather than silently ignored.

### Single-GPU and host-only suites

```bash
pytest \
  tests/moe_ep/test_sm90_push_fp8_kernel.py \
  tests/moe_ep/test_sm90_push_fp8_backend.py \
  tests/gemm/test_sm90_moe_gemm.py \
  tests/gemm/test_sm90_moe_gemm_contract.py
```

### Multi-rank tests

Requires at least two SM90 GPUs and uses `torch.distributed` only.

```bash
bash tests/moe_ep/run_tests.sh sm90_push
```

### Soak tests

```bash
SM90_PUSH_SOAK_ROUNDS=200 \
  bash tests/moe_ep/run_tests.sh sm90_push
```

### Benchmark

```bash
torchrun \
  --standalone \
  --nproc-per-node=8 \
  benchmarks/bench_sm90_push_megamoe.py \
  --config DSV3 \
  --dedup \
  --grouped-combine \
  --fuse-fc1 \
  --assert-cos-min 0.997
```

## 🔍 Related issues

This PR contributes the Hopper FP8 block-scale milestone under #3692 and the SM90 sub-issue #3780, and follows the whole-layer integration direction described in #3704.

### ✅ Pre-commit Checks

## 🧪 Tests

* [x] Tests added and updated.
* [x] Single-GPU suite passes.
* [x] EP2 distributed suite passes.
* [x] EP4 distributed suite passes.
* [x] CUDA Graph tests pass.
* [x] Soak tests pass on the listed hardware.

## Reviewer notes

Native NVFP4 and MXFP8 execution are out of scope because SM90 does not provide native block-scaled tensor-core instructions for those formats. A follow-up PR stacked on this one adds NVFP4 checkpoint support for SM90 on top of this backend, through online W4A8 decode kernels and a one-time requantization path that rides this FP8 backend unchanged.
The highest-value review areas are:

Entry points for the highest-risk areas:

* **Symmetric-window ordering and the acknowledgement protocol:**
  `kernel_src/sm90_push_megamoe/src/a2a/sm90_push_a2a_ops.cu`

* **Grouped-combine numerical behavior, where the quantization point moves:**
  The combine kernels in `kernel_src/sm90_push_megamoe/src/a2a/sm90_push_a2a_ops.cu`, gated by the reference comparison in `tests/moe_ep/test_sm90_push_fp8_kernel.py`

* **DeepGEMM fused-epilogue and workspace contracts:**
  `src/fp8_gemm/fp8_moe_fc1_fused.cuh``shim/gemm.py`

* **CUDA 12.8 gating and the AOT skip:**
  `shim/gemm.py``flashinfer/aot.py`

* **Collective initialization failure handling:**
  `shim/protocol.py`

## Summary by CodeRabbit

* **New Features**
  * Added SM90 FP8 MegaMoE support with pull- and push-style execution paths.
  * Added configurable FP8 formats, scaling modes, routing, reduction, CUDA graph, and distributed execution options.
  * Added weight preprocessing and public configuration interfaces.
  * Added benchmark tools for correctness, performance, token sweeps, and peer-to-peer bandwidth.

* **Documentation**
  * Expanded architecture guidance, setup instructions, tuning information, and reproducibility runbooks.

* **Tests**
  * Added broad correctness, validation, packaging, lifecycle, distributed, and CUDA graph coverage.


flashinfer/moe_ep/kernel_src/sm90/push_style_megamoe/src/fp8_gemm/fp8_moe_jit.cuh
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    try {
+      std::filesystem::rename(tmp_cubin_path, cubin_path);
+    } catch (std::filesystem::filesystem_error const&) {
+      if (!std::filesystem::exists(cubin_path)) {
+        std::filesystem::remove_all(tmp_path);
+        throw;
+      }
+    }
```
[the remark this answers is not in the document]
leonardHONG: Both directories are under the same cache root, so this does not cross filesystems. The atomic rename is intentional to avoid exposing a partially written cubin to another rank.

flashinfer/moe_ep/kernel_src/sm90/push_style_megamoe/src/a2a/sm90_push_a2a_ops.cu
```diff
@@ -0,0 +1,1854 @@
...
+}
+
+// unsigned increment: int32 overflow at 2^31 rounds would be UB; uint32
+// wrap is well-defined and every tag check is equality-based
+__global__ void bump_tag_kernel(int32_t* round_ctr) {
+  if (threadIdx.x == 0 && blockIdx.x == 0) {
+    auto* u = reinterpret_cast<uint32_t*>(round_ctr);
+    *u = *u + 1u;
+  }
+}
```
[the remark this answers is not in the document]
leonardHONG: Access through the matching unsigned type is valid C++. We use it intentionally so the tag wraps modulo 2^32 without signed-overflow UB.

tests/moe_ep/test_sm90_push_fp8_backend.py
```diff
@@ -0,0 +1,471 @@
...
+    with pytest.raises(MoEEpConfigError, match="int32"):
+        _forward(layer, x, ids.long(), weights)
+    with pytest.raises(MoEEpConfigError, match="float32"):
+        _forward(layer, x, ids, weights.half())
+
+    large_x, large_ids, large_weights = _make_inputs(
+        TOKEN_CAPACITY + 1, LOCAL_EXPERTS, 32, device
+    )
+    with pytest.raises(MoEEpConfigError, match="max_tokens_per_rank"):
+        _forward(layer, large_x, large_ids, large_weights)
```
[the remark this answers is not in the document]
leonardHONG: The test uses the public `MoEEpLayer` path, which rejects the oversized batch first with a `max_tokens_per_rank` error. The lower-level `token_capacity` check is never reached here, so the matcher is intentional.

tests/moe_ep/_sm90_push_fp8_baseline.py
```diff
@@ -0,0 +1,236 @@
...
+    for expert in range(num_experts):
+        quantized, scales = per_block_cast_to_fp8(w13[expert])
+        w13_fp8[expert].copy_(quantized)
+        w13_scales[expert].copy_(scales)
+        quantized, scales = per_block_cast_to_fp8(w2[expert])
+        w2_fp8[expert].copy_(quantized)
+        w2_scales[expert].copy_(scales)
+    result = (w13_fp8, w13_scales, w2_fp8, w2_scales)
+    _weight_cache[key] = result
+    return result
```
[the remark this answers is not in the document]
leonardHONG: Fixed it.

mhoqueanik:
@leonardHONG Thank you for the PR! let me run it on my end and get back to you!

mhoqueanik:
@leonardHONG My apologies for the delay! I am still restructuring moe_ep to adapt new dtypes and arch (+creating vLLM PR of fi moe_ep). Tentatively will process your PR and this https://github.com/flashinfer-ai/flashinfer/pull/4113 by mid-next week

mhoqueanik:
@leonardHONG I was able to reproduce your perf on my end. The numbers look great! Can you resolve the conflicts, and rebase it against this PR: https://github.com/flashinfer-ai/flashinfer/pull/4113? That way we can cleanly merge it

leonardHONG:
Rebased onto #4113 and resolved our conflicts. The branch builds and passes the single-GPU tests. Github currently shows #4113 conflicting with main as well, so I can rebase again after it’s updated.

mhoqueanik:
right @leonardHONG ! I have rebased the https://github.com/flashinfer-ai/flashinfer/pull/4113 fixing the conflicts. You can rebase this PR with it again (sorry for the duplicate work!). 

I tested it in 1x4 H100, will do a regression test of both PRs in 1x4 B100 - then we can start merging the PR into TOT.

leonardHONG:
No worries, thanks for fixing it! I’ve rebased onto the updated #4113.

mhoqueanik:
Awesome @leonardHONG ! Let me land the other PR then this one can go. Tests are passing on my end. Meanwhile, can you add a few perf microbench tables from [here](https://github.com/mhoqueanik/moe_ep_benchmark/tree/vllm_repro_8_gpu_v2)? you can adjust this for your cluster. It has 6 geometries.

You can also glance over the coderabbit issues that are raised meanwhile.

leonardHONG:
Thanks! I ran the six suggested shapes on 8× H100 80GB SXM. These are back-to-back steady-state measurements with 20 warmups and 100 samples per path. The table reports p50 latency in milliseconds; the last column is Pull / Push, so values above 1 mean Push is faster.

| Geometry | Tokens/rank | Pull | Push | Speedup |
|---|---:|---:|---:|---:|
| DeepSeek V3 | 8 | 0.572 | 0.614 | 0.93× |
|  | 64 | 0.657 | 0.704 | 0.93× |
|  | 512 | 1.699 | 1.065 | 1.60× |
|  | 2048 | 3.497 | 2.728 | 1.28× |
|  | 8192 | 12.935 | 9.886 | 1.31× |
| Kimi K2.6 | 8 | 0.716 | 0.749 | 0.96× |
|  | 64 | 0.913 | 0.949 | 0.96× |
|  | 512 | 2.038 | 1.163 | 1.75× |
|  | 2048 | 3.867 | 2.776 | 1.39× |
|  | 8192 | 13.023 | 10.046 | 1.30× |
| DeepSeek V4 Flash | 8 | 0.386 | 0.410 | 0.94× |
|  | 64 | 0.436 | 0.468 | 0.93× |
|  | 512 | 0.871 | 0.581 | 1.50× |
|  | 2048 | 1.960 | 1.410 | 1.39× |
|  | 8192 | 5.752 | 4.609 | 1.25× |
| DeepSeek V4 Pro | 8 | 0.951 | 0.966 | 0.98× |
|  | 64 | 1.301 | 1.294 | 1.01× |
|  | 512 | 2.660 | 1.479 | 1.80× |
|  | 2048 | 4.315 | 3.168 | 1.36× |
|  | 8192 | 14.172 | 10.757 | 1.32× |

Push is about 4–5% slower at the smallest token counts, but becomes consistently faster from T512 onward, with gains of roughly 1.25–1.80×.

Correctness also passed across all measured cases (minimum cosine similarity 0.99816, maximum NRMSE 0.06063).

GPT-OSS 120B is not supported yet because its hidden dimensions are not 128-aligned, and Qwen3.5 397B uses top-k 10 while the current limit is 8.

mhoqueanik:
Hi @leonardHONG , the numbers look good! A quick remark on the codes. I see that this PR touches other parts of FI such as gemm and deep_geem, etc. Do we need those for the push style to work? If not, can you revert those files and keep changes limited to only your backend folder only? preferably moe_ep/<push backend> and tests/moe_ep/ only? If other files needs separate fix, we can create another PR for those. What do you think?

leonardHONG:
Thanks for pointing that out! I’ve reverted the unrelated shared changes and kept the push-specific changes localized. The updated branch is pushed and tests are passing.

leonardHONG:
Fixed the actionable CodeRabbit issues and revalidated the changes.

flashinfer/moe_ep/backends/mega/kernel/mxfp8_cutedsl/backend.py
```diff
@@ -114,7 +114,7 @@ def validate_transformed_weights(
         )
 
     def _allocate_workspace(self, fleet_params: FleetParams) -> Any:
-        from .....kernel_src.cutedsl_megamoe import (
+        from .....kernel_src.sm100.cutedsl_megamoe import (
```
mhoqueanik, line 117: @leonardHONG I think these changes came from rebasing the TOT. these should be reverted back so this PR only focuses on the push style kernel codes

mhoqueanik:
Thank you @leonardHONG ! The coderabbit addressing looks good to me. But I still see the PRs touching the other folders. I think this happened when you rebased with TOT. can you revert those files to TOT and only keep the changes into /moe_ep/<your backend> and /tests/moe_ep/<your backend related files>? that way will can ensure this PR tracks only your push style kernel.

leonardHONG:
Thanks! I’ve dropped the benchmark, docs, and AOT changes. The only remaining exception is the small pyproject.toml entry needed to package the CUDA sources. Since this is still stacked on #4113, should I wait for it to land before rebasing onto main?

mhoqueanik:
I see! yeah https://github.com/flashinfer-ai/flashinfer/pull/4113 is still stuck on CI. I am trying to clear that, and this PR goes next

mhoqueanik:
Hi @leonardHONG , https://github.com/flashinfer-ai/flashinfer/pull/4113 is finally merged! You can rebase one last time and and address any coderabbit issue, then we can start the CI

leonardHONG:
Done, rebased onto the latest main and cleaned up the remaining coderabbit comments. Thanks!

flashinfer/moe_ep/kernel_src/sm90/push_style_megamoe/ACKNOWLEDGEMENT.md
```diff
@@ -0,0 +1,14 @@
...
+
+The private FP8 grouped-GEMM implementation builds on the DeepGEMM execution
+model and TensorRT-LLM CUDA support code shipped in FlashInfer's
+`csrc/nv_internal/tensorrt_llm` tree. The fused FC1 source retains the
+DeepSeek MIT, NVIDIA Apache-2.0, and FlashInfer Apache-2.0 notices that apply to
+its respective contributions. The remaining source files retain their own
+copyright and SPDX notices.
+
+DeepGEMM is maintained at <https://github.com/deepseek-ai/DeepGEMM> and
+TensorRT-LLM at <https://github.com/NVIDIA/TensorRT-LLM>.
```
mhoqueanik, line 14: @leonardHONG is this still correct? Did this PR add anything in `csrc/...`? You can put your name as contributor if you want.

@Anerudhan  @aleozlx for viz on LICENSE

flashinfer/moe_ep/modes/mega_layer.py
```diff
@@ -198,18 +198,17 @@ def forward(self, t: "MoEEpTensors") -> torch.Tensor:
...
-            quantize_input=quantize_input,
-        )
-
         y = torch.empty(
             t.num_tokens,
             self._fleet_params.token_hidden_size,
             dtype=torch.bfloat16,
             device=t.hidden_states.device,
         )
+        self._kernel.stage_inputs(
```
mhoqueanik, line 207: @leonardHONG could be unrelated moving?

mhoqueanik (approved):
Overall looks good to me. We can proceed with the merge once the LICENSE/ACKNOWLEDGE.md issue is clear.

flashinfer/moe_ep/modes/mega_layer.py
leonardHONG: This ordering is intentional for the push backend. `stage_inputs()` begins a stateful communication round, so allocating the output afterward could leave the runner staged if allocation fails. Allocating it first keeps allocation failures outside the protocol round.

flashinfer/moe_ep/kernel_src/sm90/push_style_megamoe/ACKNOWLEDGEMENT.md
leonardHONG: This PR doesn’t add or modify any files under `csrc/`. The private GEMM module uses the existing vendored DeepGEMM and TensorRT-LLM sources under `csrc/nv_internal/tensorrt_llm`, which is what the acknowledgement refers to.

flashinfer/moe_ep/modes/mega_layer.py
mhoqueanik: valid point @leonardHONG ! We didn't have this before because all other kernels were pull style

flashinfer/moe_ep/kernel_src/sm90/push_style_megamoe/ACKNOWLEDGEMENT.md
mhoqueanik: Makes sense!

aleozlx (approved):

huangzhilin-hzl:
Hi @leonardHONG， Thanks for the great work! Is there a benchmark comparing the pull- and push-style backends under the same setup? Which workloads currently favor each style?

leonardHONG:
> Hi @leonardHONG， Thanks for the great work! Is there a benchmark comparing the pull- and push-style backends under the same setup? Which workloads currently favor each style?

yep, i posted the 8× h100 comparison [[above](https://github.com/flashinfer-ai/flashinfer/pull/4069#issuecomment-5139124193)]. pull is a little better for small token counts, while push wins on larger workloads. push currently needs 128-aligned dimensions and top-k ≤ 8, but i may support more shapes later.

PR #4072, opened 2026-07-20
feat: add optional normed-output (y_out) to add_rmsnorm_fp4quant

nv-guomingz:
## 📌 Description

add_rmsnorm_fp4quant fuses add + RMSNorm + FP4 quantization into one CuTe-DSL kernel. It computes y = RMSNorm(residual + input) * weight and quantizes y to FP4, but the normed y never leaves registers — it's consumed only by the quantizer.

This is a problem when a second consumer needs y in its original precision. In a Qwen3-Next MoE layer, the post-attention norm feeds two readers at once:

the FP4 experts (want the FP4-quantized y), and
the router gate (wants y in bf16).
Today that forces two kernels — a separate fused_add_rmsnorm for the router plus a standalone fp4_quantize for the experts — reading the same normed tensor twice.

This PR adds an optional y_out tensor. When provided, the kernel also writes the normed bf16/fp16 y (one extra [M, H] store in the epilogue), so a single launch feeds both consumers. On Qwen3.6-35B-A3B-NVFP4 decode (B200) this collapses fused_add_rmsnorm + two FP4 quantizes into one kernel — roughly 8.4µs → 3.1µs per MoE layer per decode step.

✅ Key properties
Zero cost when unused. y_out=None is a compile-time specialization threaded through the JIT cache key, so the default path generates the identical kernel as before.
Non-invasive. The return-tuple arity is unchanged; y_out is written in place.
Numerically transparent. The FP4 payload, both scale-factor layouts, and the in-place residual update are byte-identical with and without y_out. y_out itself matches the fp32 reference within 1–2 ULPs, and is the un-scaled RMSNorm result (global_scale only affects the FP4 block scales, never y).
Small overhead when used. +0.19µs at M=4 (2.87µs → 3.06µs, CUDA-graph replay, PDL on) — just the extra store.

🧪 Tests
New TestYNormOutput class: correctness vs the fp32 reference, byte-identity of the FP4/scale/residual outputs with and without y_out, composition with output_both_sf_layouts, 3D inputs, MXFP4 (block_size=32), the global_scale ≠ 1 semantics, shape/dtype/device guards, and trace-dispatch refusal. 145 passed on B200.

📎 Notes
The trace template models only the core FP4-quantize signature, so a dispatch guard makes fi_trace refuse to emit a definition when y_out (or the pre-existing output_both_sf_layouts) is active, rather than silently dumping an incomplete one.
A full-file test run shows 96 unrelated failures in TestFusedVsSeparateFP4Quantize::test_mxfp4_fused_matches_separate; these are pre-existing on main (reproduced by reverting this kernel file) and untouched by this change — the edit is entirely guarded by y_out is not None.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

- **New Features**
  - Added optional `y_out` to `add_rmsnorm_fp4quant` to write `RMSNorm(x + residual) * weight` (unscaled) into a caller-provided FP16/BF16 tensor for both 2D and 3D inputs, including MXFP4 mode (block size 32).
  - `y_out` can be used alongside `output_both_sf_layouts` while preserving the existing FP4 payload, scale-factor outputs, and residual updates.

- **Documentation**
  - Updated tracing notes: trace/replay is not supported for `y_out` and `output_both_sf_layouts`.

- **Tests**
  - Added a dedicated test suite validating correctness, invariance vs baseline outputs, and argument validation.

### DIFF
--- a/tests/norm/test_add_rmsnorm_fp4_quant_cute_dsl.py
+++ b/tests/norm/test_add_rmsnorm_fp4_quant_cute_dsl.py
@@ -2211,5 +2211,431 @@ class TestOutputBothSFLayouts:
         )
 
 
+@cute_dsl_available
+@blackwell_required
+class TestYNormOutput:
+    """Optional normed-output (``y_out``).
+
+    When ``y_out`` is provided the kernel additionally stores the bf16/fp16
+    RMSNorm result (before quantization) so a single launch can feed both an
+    FP4 GEMM and a norm-precision consumer (e.g. an MoE router). The FP4
+    payload and scale-factor outputs must be unaffected by the extra store.
+    """
+
+    @pytest.mark.parametrize("batch_size", [1, 64])
+    @pytest.mark.parametrize("hidden_size", [256, 4096])
+    @pytest.mark.parametrize("dtype", [torch.float16, torch.bfloat16])
+    def test_y_out_matches_reference(self, batch_size, hidden_size, dtype):
+        """``y_out`` equals the reference RMSNorm(x+r)*weight in input dtype."""
+        from flashinfer.cute_dsl import add_rmsnorm_fp4quant
+
+        torch.manual_seed(42)
+        block_size = 16
+        eps = 1e-6
+
+        x = torch.randn(batch_size, hidden_size, device="cuda", dtype=dtype)
+        r = torch.randn(batch_size, hidden_size, device="cuda", dtype=dtype)
+        weight = torch.randn(hidden_size, device="cuda", dtype=dtype)
+
+        # Reference computed before the in-place residual update.
+        ref_norm = llama_rms_norm(x + r, weight, eps=eps)
+
+        y_out = torch.empty(batch_size, hidden_size, device="cuda", dtype=dtype)
+        add_rmsnorm_fp4quant(
+            x,
+            r,
+            weight,
+            eps=eps,
+            block_size=block_size,
+            y_out=y_out,
+        )
+
+        # Same rounding path as the reference (fp32 compute -> input dtype),
+        # so the stored norm should match within a couple of ULPs.
+        torch.testing.assert_close(
+            y_out.float(), ref_norm.float(), rtol=1e-2, atol=1e-2
+        )
+
+    @pytest.mark.parametrize("batch_size", [1, 64])
+    @pytest.mark.parametrize("hidden_size", [256, 2048])
+    @pytest.mark.parametrize("dtype", [torch.float16, torch.bfloat16])
+    def test_quant_outputs_unaffected_by_y_out(self, batch_size, hidden_size, dtype):
+        """FP4 payload and block scales are byte-identical with/without ``y_out``.
+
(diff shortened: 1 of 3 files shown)


tests/norm/test_add_rmsnorm_fp4_quant_cute_dsl.py
```diff
@@ -2211,5 +2211,431 @@ def test_is_sf_swizzled_layout_ignored_when_output_both(self):
...
+
+    When ``y_out`` is provided the kernel additionally stores the bf16/fp16
+    RMSNorm result (before quantization) so a single launch can feed both an
+    FP4 GEMM and a norm-precision consumer (e.g. an MoE router). The FP4
+    payload and scale-factor outputs must be unaffected by the extra store.
+    """
+
+    @pytest.mark.parametrize("batch_size", [1, 4, 7, 64, 256])
+    @pytest.mark.parametrize("hidden_size", [64, 256, 512, 2048, 4096])
+    @pytest.mark.parametrize("dtype", [torch.float16, torch.bfloat16])
```
bkryu: Can we reduce parameter combinations a bit? We are trying to prevent unit testing time blowing up.

tests/norm/test_add_rmsnorm_fp4_quant_cute_dsl.py
```diff
@@ -2211,5 +2211,431 @@ def test_is_sf_swizzled_layout_ignored_when_output_both(self):
...
+        # Same rounding path as the reference (fp32 compute -> input dtype),
+        # so the stored norm should match within a couple of ULPs.
+        torch.testing.assert_close(
+            y_out.float(), ref_norm.float(), rtol=1e-2, atol=1e-2
+        )
+
+    @pytest.mark.parametrize("batch_size", [1, 4, 7, 64, 256])
+    @pytest.mark.parametrize("hidden_size", [64, 256, 512, 2048, 4096])
+    @pytest.mark.parametrize("dtype", [torch.float16, torch.bfloat16])
+    def test_quant_outputs_unaffected_by_y_out(self, batch_size, hidden_size, dtype):
```
bkryu: Ditto about reducing parameter combinations a bit

bkryu (commented):
PR LGTM in general. Left a few comments about reducing parameter combinations.

Asking to update PR description as they become commit messages.
Also, the CI is failing at `pre-commit` checks. Please try `pre-commit run --all-files`

tests/norm/test_add_rmsnorm_fp4_quant_cute_dsl.py
nv-guomingz: Thanks @bkryu , I've updated the combination logic. The test case counts reduced from 145 to 37 now.

tests/norm/test_add_rmsnorm_fp4_quant_cute_dsl.py
nv-guomingz: Updated.

bkryu (approved):

PR #4081, opened 2026-07-22
feat(gdn): u/d cache spec-decode kernels for replayssm

ameynaik-hub:
## 📌 Description
Two CuTe-DSL kernels that make speculative decoding cheap for GDN
(Gated DeltaNet) models, plus tests and a benchmark.

The problem: during speculative decoding, the engine must be able to
roll back rejected draft tokens. For GDN models the naive way is to
save a full copy of the recurrent state for every draft position — but
GDN states are megabytes per layer per request, so those copies eat
both memory capacity and bandwidth.

The idea: do not save states at all. Save only the small per-token
ingredients (update vector u, normalized key k, decay g) in a small
ring per request. One kernel launch per layer per decode step then does
everything: computes the verify output for the draft tokens, appends
the new ingredients to the ring, and — only when a request's live
window is full, roughly once every 3-5 steps — folds the window into
the single checkpoint state. Rejected tokens cost nothing (just a
cursor that does not advance).

Ring layout and cursors: the ring has 32 physical slots per request
(`RING_SLOTS`), of which at most 16 are ever live (`W_RING`, the max
history window). Two per-request int32 cursors describe the window:
`cache_base` (where it starts) and `hist_len` (how long it is); row j
lives at physical slot `(cache_base + j) & 31`. The kernel treats both
cursors as READ-ONLY: history is read through the rotation, and new
tokens are always appended past the live window at
`(cache_base + hist_len + s) & 31` — for folding and non-folding
requests alike. Since window (≤16) plus appends (≤8) always fit in 32
slots, an append can never touch a slot any of the request's sibling
CTAs is still reading — the fold/append overlap is impossible by
construction, with no inter-CTA synchronization needed.

Cursor commits happen OUTSIDE the launch, after acceptance is known:
fold → `cache_base = (cache_base + hist_len) & 31; hist_len =
n_accepted`; no fold → `hist_len += n_accepted`. These are exactly the
cursor semantics of vLLM's Triton ReplaySSM spec backend, so a serving
integration can share one cursor set between that backend and this
kernel. For standalone use the wrapper can apply the fold commit for
you (`restart_hist_on_flush=True`); this path requires a caller-owned
`cache_base` tensor and validates `hist_len ∈ [0, 16]` /
`cache_base ∈ [0, 32)` (validation is skipped while a CUDA graph is
capturing; the eager warmup calls cover it). The serving path
(`restart_hist_on_flush=False`) stays sync-free.

Included:
- gdn_decode_bf16_wy_ucache_flush.py — the main fused verify+flush
  kernel (32-slot ring as above)
- gdn_decode_bf16_wy_ucache.py — a verify-only variant. NOTE: this
  variant still uses the legacy 16-deep flat layout (history at rows
  [0, hist_len), no cache_base) and is NOT ring-format compatible with
  the flush kernel — do not share one ring allocation between them.
- tests/gdn/test_decode_ucache.py — correctness vs a simple fp32
  reference, including wrapped-window cases (base+len crossing the ring
  boundary), a fold-never-overwrites-the-live-window property test, the
  strided (chunk-view) q/k/v + a/b path, and negative tests for the
  cursor-misuse guards
- benchmarks/bench_gdn_ucache_flush.py — latency benchmark
  (`--no-commit` times the pure kernel; `--base` measures wrapped
  windows)

Supported precision modes (`g_cache` is fp32 in every mode):

| # | mode / arm name | env knobs | q/k/v + output | SSM state | u/k cache | g (gate) |
|---|---|---|---|---|---|---|
| 1 | bf16 (default) | (none) | bf16 | bf16 | bf16 | fp32 |
| 2 | fp16_io | `IO_DTYPE=fp16` | fp16 | fp16 | fp16 | fp32 |
| 3 | fp16_state | `STATE_DTYPE=fp16` | bf16 | fp16 | bf16 | fp32 |
| 4 | ring_fp16 (cache) | `RING_DTYPE=fp16` | bf16 | bf16 | fp16 | fp32 |
| 5 | fp16_state_cache | `STATE_DTYPE=fp16 + RING_DTYPE=fp16` | bf16 | fp16 | fp16 | fp32 |

The fp16-state ("mixed") mode is the analogue of
mamba_ssm_cache_dtype=float16 in serving frameworks: only the
checkpoint gains fp16's extra mantissa bits; everything else stays bf16.

How to run:
  pytest tests/gdn/test_decode_ucache.py -v
  python benchmarks/bench_gdn_ucache_flush.py --arm bf16 --iters 1000 --no-commit
  python benchmarks/bench_gdn_ucache_flush.py --arm fp16_state --iters 1000 --no-commit

Benchmark — us per layer per decode step, 1x B200, Qwen3.5-122B GDN
geometry at TP1 (H=16 key heads, HV=64 value heads, K=V=128, T=4 draft
window, 32-slot ring / 16-row window); columns = fraction of requests
folding their window that step, scattered at exact counts (CUDA-graph
replay, CUPTI cold-L2, median of 1000, `--no-commit` = pure kernel
time; the cursor commit is caller-owned in serving and excluded here).
Wrapped windows (base near the ring boundary) measure the same as
base=0.

| batch | 0% fold | 20% fold | 40% fold | 80% fold |
|------:|--------:|---------:|---------:|---------:|
|     8 |     9.4 |     13.3 |     14.2 |     14.5 |
|    32 |    22.6 |     27.7 |     30.9 |     36.9 |
|    64 |    40.4 |     48.0 |     55.3 |     67.6 |
|   128 |    70.9 |     86.4 |    100.3 |    125.6 |
|   256 |   131.5 |    162.0 |    188.5 |    243.8 |

Scope/limits: validated on B200 (SM100), head dims K=V=128, draft
windows T in {4, 8}. Ring depth is fixed at 32 slots with a 16-row max
window (sized for T ≤ 8). The verify-only variant has not been
converted to the ring layout. Dtype selection via env vars and a couple
of API stubs are known follow-ups.

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary

- **New Features**
  - Added SM90+ Gated Delta-Net ucache support with conditional “verify+flush” (folding history into state) and configurable IO/state/ring dtype specialization.
  - Introduced new public decode entry point for verify+flush with options to control cache base and flush commit behavior.
- **Benchmarking**
  - Added a CUDA verify+flush benchmark sweeping batch sizes and flush rates, including configurable ring modes and cursor commit controls.
- **Tests**
  - Added fp32-oracle correctness and regression coverage for outputs, folded committed state, ring overwrite safety, and strided vs compact execution.


ameynaik-hub:
[@Johnny-Liou ](https://github.com/Johnny-Liou)

flashinfer/gdn_kernels/gdn_decode_bf16_wy_ucache_flush.py
```diff
@@ -0,0 +1,3749 @@
...
+  q, k                   [B, T, H,  K]     IO      in       draft query / key
+  v                      [B, T, HV, V]     IO      in       draft values
+  a, b                   [B, T, HV]        IO      in       per-token gate / beta
+  A_log                  [HV]              bf16    in       per-head log-decay (cast+cached)
+  dt_bias                [HV]              bf16    in       per-head time-step bias (cast+cached)
+  initial_state_source   [pool, HV, V, K]  STATE   in/out   checkpoint S0 (written on flush)
+  initial_state_indices  [B]               int32   in       per-request pool slot
+  k_cache                [pool, H,  16, K]  IO     in/out   ring: L2-normalized keys
+  u_cache                [pool, HV, 16, V]  IO     in/out   ring: correction vectors
+  g_cache                [pool, HV, 16]    f32     in/out   ring: cumulative log-decay
```
kahyunnam, line 70: 32 not 16 for ring slots?

tests/gdn/test_decode_ucache.py
```diff
@@ -0,0 +1,499 @@
+"""
```
kahyunnam, line 1: all cases load the flush module; verify-only has no oracle coverage, can we have a thin test for verify-only?

flashinfer/gdn_kernels/gdn_decode_bf16_wy_ucache.py
```diff
@@ -0,0 +1,2807 @@
...
+
+device = torch.device("cuda:0")
+
+# Problem dimensions. One CTA processes a full V tile per (request, head).
+T = 16
+K_DIM = 128
+V_DIM_C = 128  # full V tile per CTA
+BK_H = 16  # K-tile for the H GEMM (multiple of 16 for mma.k=16)
+EPS = 1e-6
+io = cutlass.BFloat16
```
kahyunnam, line 97: let's assert q/k/v/a/b are bf16 (or convert), and/or add dtype to the verify cache key once body can specialize. 

Context: trying to avoid GDN-C1 and GDN-C3 items from https://github.com/flashinfer-ai/flashinfer/issues/4214

kahyunnam (commented):
The pre-commit is red (https://github.com/flashinfer-ai/flashinfer/actions/runs/30186413244/job/89751881862?pr=4081) +  we probably need to export this from https://github.com/flashinfer-ai/flashinfer/blob/main/flashinfer/gdn_kernels/__init__.py#L94

flashinfer/gdn_kernels/gdn_decode_bf16_wy_ucache_flush.py
ameynaik-hub: during one kernel launch, two things happen to the same request's cache at the same time:
(1) if the row is flushing, the kernel is reading the history rows (up to 16 of them) to fold them into the state checkpoint, and 
(2) the kernel is WRITING the current step's new tokens (T=4) into the cache.
  
If the buffer were exactly 16 slots, there'd be nowhere to put the new tokens except by wrapping around to the front. Example: history is 13 rows (the flush threshold), so the new 4 tokens go to positions 13, 14, 15, 16  but 16 wraps to slot 0 on a 16-slot ring, which is a history row the flush is still reading. That's a read/write race, and there's no synchronization between the thread blocks to prevent it.
So the physical ring is bigger than the logical window: new tokens are always written PAST the live window into fresh slots, never on top of it. Worst case is a full 16-row window plus up to 8 new tokens = 24 slots needed at once → round up to the next power of two = 32.   

it also makes rollback trivial. The new tokens sit outside the committed window, and a cursor advances by the number of accepted tokens after the launch if some tokens are rejected, nothing in the live history was ever touched.

flashinfer/gdn_kernels/gdn_decode_bf16_wy_ucache.py
ameynaik-hub: fixed.

kahyunnam (approved):

Issue #3957, opened 2026-07-14
[Bug] nvfp4 unified-MoE on SM100: cumulative cross-call state corruption after ~50 varied-shape calls in one process (device-side gather assert; victim config moves)


YangXu1990uiuc:
## Summary

While enabling `tests/moe/test_unified_moe_fuzz.py` for CI, the fuzzer surfaced a **silent out-of-bounds device write** in the nvfp4 unified-MoE stack on SM100: an earlier fuzz config corrupts device memory, and a later config (`nvfp4_uniform_e16_k8_t3_h1536_i256_s46`) then trips a `vectorized_gather_kernel index out of bounds` device-side assert when the corrupted memory is consumed as gather indices. The failure is **deterministic** (2/2 accumulated runs stop at the same test id) but **state-dependent**: the "failing" config passes cleanly in isolation, so it is the victim, not the culprit.

This is the C3/C5 class from the release-quality plan (#3605) — the device-state probe is doing exactly what it was designed to do.

## Environment

- B200-class SM100 board, driver 595.71.05
- torch 2.9.1+cu130, CUDA 13.2 (JIT), cutlass-dsl 4.6.1
- flashinfer main @ `ce40d25a` (MegaMoE integration), JIT install
- Wired nvfp4 backends exercised by the run: `cute_dsl_nvfp4` + `trtllm_fp4_routed`

## Repro (deterministic, ~18 min on warm JIT cache)

```bash
FLASHINFER_UMOE_FUZZ=1 CUDA_VISIBLE_DEVICES=<sm100-idx> \
  pytest tests/moe/test_unified_moe_fuzz.py -q -x
```

Result (identical across 2 runs):

```
1 failed, 50 passed in ~18 min
FAILED tests/moe/test_unified_moe_fuzz.py::test_unified_moe_fuzz[nvfp4_uniform_e16_k8_t3_h1536_i256_s46]
/pytorch/.../IndexKernelUtils.cu:16: vectorized_gather_kernel: Assertion `ind >= 0 && ind < ind_dim_size` failed
```

- The assert is reported asynchronously at the fuzzer's device-state `torch.cuda.synchronize()`.
[...]

PR #4086, opened 2026-07-22
fix(moe): reject CuteDSL MoE tactics whose N-tiling overruns the output (#3957)

YangXu1990uiuc:
## 📌 Description

Stop-the-bleeding fix for #3957 (cumulative cross-call device-memory corruption on SM100). Root cause and full evidence chain are on the issue; short version: the CuteDSL MoE epilogues store **full CTA-tile rows with no column predicate** — the finalize kernel's raw-pointer `cp.reduce.async.bulk ... add` scatter and gemm1's SFC `autovec_copy` (the epilogue TODO, from #2398) — so any tactic whose N-tiling leaves a **partial CTA tile** or a **cluster-padding CTA** (the persistent scheduler pads the grid with an M-only validity guard) read-modify-writes past the real output columns into neighboring pool memory. Silent for most landings (add-of-zero canonicalizes NaN bit patterns → corrupts int data reinterpreted as bf16, e.g. trtllm's long-lived permute-index cache), an IMA when it falls off the last allocation.

**Changes:**
- finalize `can_implement`: require `n % (mma_tiler_n * cluster_n) == 0` (single idiomatic check rejects both failure modes; mechanisms documented in the comment).
- gemm1 `can_implement`: require `n % mma_tiler_n == 0` (`cluster_n == 1` already enforced there).
- tuner: the `DEFAULT_MOE_TACTIC` fallback is **gated on the same `can_implement`** — never fall back to an unvalidated tactic; refuse the shape (empty tactic list) if even the default cannot run safely.
- `tests/moe/test_cute_dsl_moe_can_implement.py`: 8 directed host-side tests pinning the accept/reject matrix (incl. the `n=384, mma_n=256, cluster_n=2` case a `ceil_div`-style check would miss).

**Validation (B200-class SM100):**
- Unpatched main: the accumulated fuzzer sweep deterministically cascades + hard-aborts at item ~54 (the #3957 signature).
- With the guards: **all 98 configs run to completion, no cascade, no abort**.
- Residual mxfp8/`trtllm_fp8_block` failures reproduce on unpatched main (different backend, untouched by this diff) and will be tracked separately.

**Out of scope (kernel-owner follow-ups, documented on #3957):** padding CTAs keeping cluster sync but skipping the scatter; predicated tail handling; a real coordinate predicate on gemm1's SFC store. The same kernels ship in TensorRT-LLM with weaker filtering — this guard should be forwarded upstream.

## 🔍 Related Issues

Fixes the tactic-launch surface of #3957. Unblocks #3958.

## 🧪 Tests

- [x] 8 new directed unit tests (host-side, no GPU) pass.
- [x] B200 accumulated-sweep A/B as described above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added stricter eligibility checks to reject MoE GEMM configurations that could cause out-of-bounds output writes or epilogue accumulation.
  * Improved tactic selection by enforcing required divisibility of the output N dimension across configured tiling and cluster shapes.

* **Tests**
  * Added host-side validation tests covering CuteDSL MoE eligibility checks for finalize and GEMM1 paths, including valid and invalid N-tiling scenarios.

### DIFF
--- a/flashinfer/fused_moe/cute_dsl/tuner.py
+++ b/flashinfer/fused_moe/cute_dsl/tuner.py
@@ -532,34 +532,23 @@ class CuteDslFusedMoENvfp4Runner(TunableRunner):
         valid_tactics = [t for t in ALL_MOE_TACTICS if _tactic_ok(t)]
 
         if not valid_tactics:
-            # gh #3957: never fall back to an UNVALIDATED tactic. The
-            # can_implement guards exist because the kernels' unpredicated
-            # epilogue stores overrun partial/padded N tiles; a fallback that
-            # bypasses them re-opens the same silent device-memory corruption.
-            if _tactic_ok(DEFAULT_MOE_TACTIC):
-                logger.warning(
-                    "No valid tactics found for problem dims "
-                    "(tokens=%d, hidden=%d, intermediate=%d, experts=%d, top_k=%d). "
-                    "Falling back to default tactic.",
-                    num_tokens,
-                    hidden_size,
-                    intermediate_size,
-                    num_local_experts,
-                    self.top_k,
-                )
-                valid_tactics = [DEFAULT_MOE_TACTIC]
-            else:
-                logger.warning(
-                    "cute_dsl MoE refuses problem dims "
-                    "(tokens=%d, hidden=%d, intermediate=%d, experts=%d, top_k=%d): "
-                    "no tactic (incl. default) passes can_implement (gh #3957).",
-                    num_tokens,
-                    hidden_size,
-                    intermediate_size,
-                    num_local_experts,
-                    self.top_k,
-                )
-                return []
+            # DEFAULT_MOE_TACTIC is a member of ALL_MOE_TACTICS, so an empty
+            # list means even the default fails can_implement -- do not fall
+            # back to it unvalidated (gh #3957). This early refusal is
+            # diagnostics/defense-in-depth: the kernel wrappers re-validate
+            # can_implement at launch and raise, so an unvalidated tactic
+            # cannot reach the device -- but refusing here avoids pointless
+            # profiling of a tactic that can only throw, and says why.
+            logger.warning(
+                "cute_dsl MoE refuses problem dims "
+                "(tokens=%d, hidden=%d, intermediate=%d, experts=%d, top_k=%d): "
+                "no tactic (incl. default) passes can_implement (gh #3957).",
+                num_tokens,
+                hidden_size,
+                intermediate_size,
+                num_local_experts,
+                self.top_k,
+            )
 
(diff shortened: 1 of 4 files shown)


YangXu1990uiuc:
Disposition of the second review round:

1. **Staging blocker — stale snapshot.** The review predated the commit: PR head `ef255f2b` (now `7d80fc65`) contains the final single-modulo finalize guard, the corrected gemm1 guard, the tuner change, and the 8 directed tests. Please re-review at the current head.
2. **Tuner fallback — both sub-points confirmed and taken.** `DEFAULT_MOE_TACTIC ∈ ALL_MOE_TACTICS` (verified at runtime), so the re-check branch was dead — removed in `7d80fc65`. The remaining early refusal (warn + empty list) is now documented honestly as diagnostics/defense-in-depth: the kernel wrappers re-validate `can_implement` at launch and raise `Unsupported configuration`, so an unvalidated tactic cannot reach the device either way. Making `MoELayer` *skip* a runner with no valid tactics (instead of surfacing the wrapper's error) is a real multi-backend dispatch improvement but out of scope for this fix — noted on #3957.
3. **Ledger backend scope — fixed in #3958 (`54590e35`).** Backend-scoped Findings no longer match config-level (`backend=None`) queries, so a future scoped quarantine cannot xfail the whole config and silence healthy backends; the new `skip_backend()` applies scoped crash-quarantines per-runner in the loop instead.

flashinfer/fused_moe/cute_dsl/tuner.py
```diff
@@ -528,21 +527,28 @@ def get_valid_tactics(  # type: ignore[override]
...
+            # back to it unvalidated (gh #3957). This early refusal is
+            # diagnostics/defense-in-depth: the kernel wrappers re-validate
+            # can_implement at launch and raise, so an unvalidated tactic
+            # cannot reach the device -- but refusing here avoids pointless
+            # profiling of a tactic that can only throw, and says why.
             logger.warning(
-                "No valid tactics found for problem dims "
-                "(tokens=%d, hidden=%d, intermediate=%d, experts=%d, top_k=%d). "
-                "Falling back to default tactic.",
+                "cute_dsl MoE refuses problem dims "
```
aleozlx: this logging text change seem unnecessary

aleozlx (approved):

flashinfer/fused_moe/cute_dsl/tuner.py
aleozlx: already addressed in my commit below

aleozlx:
this is been merged into #4186 as well

https://github.com/flashinfer-ai/flashinfer/pull/4186/commits

closing

Issue #3926, opened 2026-07-11
test(moe): cover do_finalize=False expert_weights dtype in the unified MoE fuzzer


aleozlx:
## Summary

The trtllm-gen routing kernel always emits expert weights in **bfloat16** (`routingData.mDtypeOutput` is hard-set to `Bfloat16` for every routing method in `csrc/trtllm_fused_moe_runner.cu`), independent of the `routing_logits` dtype. On the `do_finalize=False` path the `expert_weights` buffer is surfaced to the caller verbatim, so its dtype must be bf16 even when the routing logits are fp32 (the conventional DeepSeekV3 case). A mismatched fp32 label there reinterprets bf16 bytes as fp32 → garbage.

## Coverage gap

The routing-aware unified MoE fuzzer added in #3892 (`tests/moe/test_unified_moe_fuzz.py`) already parametrizes `routing_method × routing_input_mode × logits_dtype`, but it does not assert the dtype of the returned `expert_weights` on the `do_finalize=False` path. That specific combination (fp32 logits + `do_finalize=False`) is currently uncovered.

## Ask

Extend the unified MoE fuzzer (see #3892 for the relevant axes / oracle / dispatch code) so that, for each trtllm-gen fused-MoE operator:

- `trtllm_bf16_moe`
- `trtllm_fp8_per_tensor_scale_moe`
- `trtllm_fp8_block_scale_moe`
- `trtllm_mxint4_block_scale_moe`

run with fp32 (e.g. DeepSeekV3) routing logits and `do_finalize=False`, and assert the returned `expert_weights.dtype == torch.bfloat16` regardless of the `logits_dtype` axis.

This closes the fp32-logits + unfinalized path in the fuzzer and guards against future regressions of the routing-output dtype contract.

feih-nv:
TODO: Also cover fake-op contract parity (discovered in PR #4088 ). Both FP8 per-tensor fake implementations currently always return one tensor, while real operations return three tensors when `do_finalize=False`. Update both fake implementations when custom/fake-op registration is re-enabled.

PR #4088, opened 2026-07-22
feat: close feature gap by wiring up per-tensor routed FP8 fused-moe

jdebache:
## 📌 Description

FP8 per-tensor routed fused-MoE is supported at the kernel level. This PR exposes it and tests it.

### ✅ Pre-commit Checks

## 🧪 Tests

- [X] All tests are passing (`unittest`, etc.).

## Summary by CodeRabbit

- **New Features**
  - Added FP8 per-tensor MoE pre-routed execution that accepts packed expert top-k indices (routing logits can be omitted).
  - Introduced a new routed FP8 per-tensor MoE API and exposed it at the Python package top level.
  - Supports non-zero local expert offsets for correct shifted expert dispatch.
- **Tests**
  - Added parity tests comparing logits-based vs pre-routed outputs, including renormalize variants.
  - Added correctness coverage for non-zero expert offset scenarios.
- **Trace**
  - Added routed MoE trace template, example output, and trace-schema/reference tests.


csrc/trtllm_fused_moe_kernel_launcher.cu
```diff
@@ -2308,8 +2394,101 @@ Array<Tensor> trtllm_fp8_per_tensor_scale_moe(
...
+    int64_t local_expert_offset, int64_t local_num_experts, Optional<double> routed_scaling_factor,
+    bool use_routing_scales_on_input, int64_t routing_method_type, bool do_finalize,
+    bool enable_pdl, Array<int64_t> config_index, int64_t activation_type, bool norm_topk_prob,
+    Optional<TensorView> routing_replay_out) {
+  // Basic type validation
+  auto const dtype = hidden_states.dtype();
+  auto const activation = validateAndCastActivationType(activation_type);
+
+  TVM_FFI_ICHECK_EQ(expert_indices.dtype(), dl_int32)
+      << "FP8 MoE: expert_indices must be int32 (packed expert_id/weight).";
```
feih-nv, line 2435: This comment has not been fully addressed. 

Could we validate `expert_indices` fully at the start of this routed API?

This routed entry point has no routing logits, so it must always receive packed expert indices with shape `[num_tokens, top_k]`.

Currently, inputs such as these are not rejected clearly:

- a 1-D tensor such as `[num_tokens * top_k]`
- a 3-D tensor
- an empty tensor such as `[0, top_k]` when `hidden_states` contains tokens
- a tensor with the wrong top-k width, such as `[num_tokens, top_k + 1]`
- a noncontiguous tensor
- a CPU tensor or a tensor on a different GPU

Some malformed shapes cause `has_precomputed_routing()` to return false. The launcher then behaves as if no packed routing was supplied, but this API also has no logits, so the routing kernel may read an uninitialized buffer.

Please require:

```text
- CUDA tensor on the same device as `hidden_states`
- contiguous `int32`
- shape `[num_tokens, top_k]`
```

flashinfer/fused_moe/core.py
```diff
@@ -3571,6 +3778,141 @@ def trtllm_fp8_per_tensor_scale_moe(
         return result
 
 
+@flashinfer_api
+def trtllm_fp8_per_tensor_scale_routed_moe(
```
feih-nv, line 3784: Could this new public API be wired to a trace template?

Repository guidance requires public `@flashinfer_api` entry points to provide `trace=...`, add an example in `tests/trace/example.py`, regenerate the trace JSON, and run `tests/trace/`.

The routed per-tensor API currently has no trace template/example, so `fi_trace` cannot capture this path.

tests/moe/test_trtllm_gen_routed_fused_moe.py
```diff
@@ -458,6 +460,134 @@ def test_trtllm_gen_fp8_routed_fused_moe(
...
+        output1_scales_scalar=output1_scales_scalar,
+        output1_scales_gate_scalar=output1_scales_gate_scalar,
+        gemm2_weights=gemm2_weights,
+        output2_scales_scalar=output2_scales_scalar,
+        num_experts=num_experts,
+        top_k=top_k,
+        n_group=None,
+        topk_group=None,
+        intermediate_size=intermediate_size,
+        local_expert_offset=0,
```
feih-nv, line 576: Could we add a sharded expert-offset case perhaps?

flashinfer/fused_moe/core.py
```diff
@@ -2219,6 +2248,183 @@ def _fake_trtllm_fp8_per_tensor_scale_moe(
...
+            enable_pdl,
+            [-1, -1] if tactic == -1 else tactic,
+            activation_type,
+            True,  # norm_topk_prob: unused for pre-computed routing
+            routing_replay_out,
+        )
+        return _unpack_trtllm_moe_output(intermediate_output, output, do_finalize, None)
+
+    @register_fake_op("flashinfer::trtllm_fp8_per_tensor_scale_routed_moe")
+    def _fake_trtllm_fp8_per_tensor_scale_routed_moe(
```
feih-nv, line 2393: How should the fake op handle `do_finalize=False`?

The real routed op returns three tensors when finalization is disabled:
```text
gemm2_output
expert_weights
expanded_idx_to_permuted_idx
```
but the fake op always returns one `[num_tokens, hidden_size]` tensor.

This is also tracked by #3926 , here we may do 2 at least?

1. make its fake output match the real three-tensor result, or
2. explicitly reject do_finalize=False?

feih-nv (approved):

feih-nv:
Retriggered the failed 5090 job in the pipeline above, passed here https://gitlab-master.nvidia.com/dl/flashinfer/flashinfer-ci/-/jobs/374648320.

PR #4089, opened 2026-07-22
fix(mnnvl): use SCM_RIGHTS fd exchange for POSIX-FD path instead of pidfd_getfd

thanhhao98:
## Problem

`MnnvlMemory.open_mnnvl_memory` has two handle-exchange paths, selected in `get_allocation_prop` by CPU arch:
- **aarch64** (GB200/GB300) → `CU_MEM_HANDLE_TYPE_FABRIC`
- **x86_64** (B200/B300, Hopper) → `CU_MEM_HANDLE_TYPE_POSIX_FILE_DESCRIPTOR`

The POSIX-FD path imports peer ranks' CUDA VMM file descriptors via `pidfd_open` / **`pidfd_getfd`**, which require `PTRACE_MODE_ATTACH`. Inside a **default-capability container** this fails:

```
RuntimeError: pidfd_getfd(pidfd=..., fd=...) failed with errno 1: Operation not permitted.
  ... try adding --cap-add=SYS_PTRACE to your docker run command.
```

Two independent kernel gates block `pidfd_getfd` without `CAP_SYS_PTRACE`:
1. Docker's **default seccomp** profile returns `EPERM` for `pidfd_getfd`.
2. **yama `ptrace_scope=1`** forbids sibling→sibling ptrace (worker ranks are siblings).

So every consumer of `MnnvlMemory` on x86 (decode-CP all-to-all, MoE all-to-all) needs `--cap-add=SYS_PTRACE` — which many managed/hardened container platforms disallow.

## Fix

Reuse the **`PosixFDHandleExchanger`** (`IpcSocket` + `SCM_RIGHTS` over a Unix domain socket) that already exists in this file and is already used by `SymmDeviceMemory`. Passing fds via `SCM_RIGHTS` needs **no ptrace** — it works under the default seccomp profile and any `ptrace_scope`, with no added capability. The FABRIC path (aarch64) is unchanged.

```python
# before: pidfd_open + pidfd_getfd loop (needs CAP_SYS_PTRACE)
# after:
exchanger = PosixFDHandleExchanger(comm, comm_rank, comm_size)
try:
    all_handles_data = exchanger.allgather(exported_fabric_handle)
finally:
    exchanger.close()
```

This is why `SymmDeviceMemory` (e.g. the trtllm MNNVL all-reduce fusion) already runs on x86 Blackwell in a default container, while `MnnvlMemory` consumers do not — this change brings `MnnvlMemory` to the same, cap-free mechanism.

## Verification (8×B200, x86, single node, Docker, default caps, yama ptrace_scope=1)

Real serving run (DeepSeek-V2-Lite, tp4 = dcp4, SGLang `--dcp-comm-backend fi_a2a`, which allocates the MNNVL decode-CP workspace via `MnnvlMemory`):

| case | flags | result |
|---|---|---|
| stock flashinfer | no `--cap-add` | ❌ `pidfd_getfd ... errno 1: Operation not permitted` at workspace init |
| stock flashinfer | `--cap-add=SYS_PTRACE` | ✅ boots, GSM8K 0.66 |
| **this patch** | **no `--cap-add`** | ✅ **boots, GSM8K 0.66** |

(NCCL `a2a` baseline boots without any flag, GSM8K 0.63 — confirms DCP itself is fine; only the MNNVL fd exchange was gated.)

(A standalone multi-process `MnnvlMemory` alloc repro reproduces the same matrix and is available on request.)

## Notes
- Also fixes the FlashInfer MNNVL MoE all-to-all (`trtllm_moe_alltoall`, same `MnnvlMemory`) in default containers.
- No behavior change on aarch64 NVL72 (FABRIC path untouched).
- The x86 POSIX-FD path now additionally calls `comm.bcast` and `comm.barrier` (the old pidfd loop needed only `comm.allgather`). `MPIBackend` and `TorchDistBackend` implement both; a custom duck-typed `CommBackend` must too.
- POSIX-FD fds are closed after import (no fd leak, no allocation pinned past `close_mnnvl_memory`); the exchange raises an actionable error / times out instead of hanging on a misconfig. CPU-only coverage added in `tests/comm/test_posix_fd_exchanger.py`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved non-FABRIC memory handle exchange to use a POSIX FD passing approach for better compatibility and reliability.
  * Added clearer error messaging for single-node/network-namespace requirements.
  * Ensured file-descriptor cleanup to avoid leaks and potential hangs during FD receiving.
* **Tests**
  * Added a CPU-only distributed test covering POSIX FD exchanger rank-to-descriptor correctness across multiple world sizes.

### DIFF
--- a/flashinfer/comm/mnnvl.py
+++ b/flashinfer/comm/mnnvl.py
@@ -530,41 +530,17 @@ class MnnvlMemory:  # type: ignore[no-redef]
         ):
             all_handles_data = comm.allgather(exported_fabric_handle.data)
         else:
-            all_handles_data = comm.allgather(exported_fabric_handle)
-            all_pids = comm.allgather(os.getpid())
-            libc = ctypes.CDLL(None, use_errno=True)
-            syscall = libc.syscall
-            SYS_pidfd_open = 434
-            SYS_pidfd_getfd = 438
-            pidfds = []
-            for pid in all_pids:
-                pidfd = syscall(SYS_pidfd_open, pid, 0)
-                if pidfd < 0:
-                    err = ctypes.get_errno()
-                    raise RuntimeError(
-                        f"pidfd_open({pid}) failed with errno {err}: {os.strerror(err)}"
-                    )
-                pidfds.append(pidfd)
-
-            remote_fds = []
-            for pidfd, fd in zip(pidfds, all_handles_data, strict=True):
-                remote_fd = syscall(SYS_pidfd_getfd, pidfd, fd, 0)
-                if remote_fd < 0:
-                    err = ctypes.get_errno()
-                    error_msg = f"pidfd_getfd(pidfd={pidfd}, fd={fd}) failed with errno {err}: {os.strerror(err)}."
-                    if err == 1:  # EPERM
-                        error_msg += (
-                            " Permission denied. If running in a container, try adding --cap-add=SYS_PTRACE "
-                            "to your docker run command."
-                        )
-                    else:
-                        error_msg += (
-                            " This may be due to kernel version (requires Linux 5.6+)."
-                        )
-                    raise RuntimeError(error_msg)
-                remote_fds.append(remote_fd)
-
-            all_handles_data = remote_fds
+            # POSIX-FD single-node path (x86 Blackwell/Hopper, no NVLink fabric
+            # cluster): exchange the CUDA VMM file descriptors via SCM_RIGHTS over
+            # a Unix socket, reusing PosixFDHandleExchanger -- the same mechanism
+            # SymmDeviceMemory already uses. This avoids pidfd_getfd, which needs
+            # PTRACE_MODE_ATTACH (CAP_SYS_PTRACE + a permissive yama ptrace_scope)
+            # and fails with EPERM inside a default-capability container.
+            exchanger = PosixFDHandleExchanger(comm, comm_rank, comm_size)
+            try:
+                all_handles_data = exchanger.allgather(exported_fabric_handle)
+            finally:
+                exchanger.close()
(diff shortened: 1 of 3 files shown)


thanhhao98:
Thanks for the extremely thorough review — both required items are in, plus the two optional ones you flagged.

**Required**
- **F1 — close the exchanged fds.** The POSIX-FD branch now closes every fd it owns, gated on `is_posix_fd`: each `all_handles_data[i]` right after `cuMemImportFromShareableHandle`/map (peer fds, and the ring's iteration-0 self-dup for `i == comm_rank`), plus `exported_fabric_handle` after the loop. Mirrors `SymmDeviceMemory`'s `exchanger.cleanup()` / NCCL's close-after-import. Net transient footprint is now peak ~2, and the allocation is no longer pinned past `close_mnnvl_memory`. → "leaks nothing".
- **F2 — stale test gate.** Dropped `_check_pidfd_permissions()` (and its now-dead `ctypes`/`os` imports + false docstring) from `tests/comm/conftest.py`; `mnnvl_available()` is now just `MnnvlMemory.supports_mnnvl()`, so the MNNVL tests run in default-capability containers.

**Optional (did both)**
- **F4 — actionable failure + timeout.** `allgather` is wrapped to re-raise `OSError` as a `RuntimeError` explaining the one-node/shared-netns requirement, and `IpcSocket.recv_fd` now `settimeout(120)` so an asymmetric peer failure surfaces as `socket.timeout` instead of a hang (also benefits the existing `SymmDeviceMemory` user).
- **§6 — CPU-only coverage.** Added `tests/comm/test_posix_fd_exchanger.py`: spawns N∈{1,2,4} processes (gloo), each shares a rank-tagged `memfd`, allgathers, and asserts each received fd's content maps to the right source rank — no GPU. Catches any ring-ordering/indexing regression in the mechanism `open_mnnvl_memory` now depends on. Passes locally (3/3, 11s).
- **F3** — added a note to the PR description that the x86 path now also uses `comm.bcast`/`comm.barrier` (custom `CommBackend`s must implement them).

**Verification:** CPU test 3/3 green; re-ran the 8×B200 default-caps model boot (`--dcp-comm-backend fi_a2a`, no `--cap-add`) with the fd-close code — still boots and GSM8K matches the `--cap-add=SYS_PTRACE` baseline. FABRIC path still untouched.

Left F5 (same-node guard) as a follow-up per your note. Ready for re-review — thanks!

Issue #4087, opened 2026-07-22
[Bug] mxfp8/deepseekfp8 unified-MoE (trtllm_fp8_block): marginal numeric failures + a solo-config core dump on SM100 (pre-existing on main, surfaced by the fuzzer)


YangXu1990uiuc:
Found while validating #4086 on a B200-class SM100 (main @ `6104afe3`). Two distinct symptoms on the recently-added mxfp8/deepseekfp8 unified-MoE paths, both reproducing on **unpatched main** (the #4086 guards touch only CuteDSL nvfp4 kernels — different backend):

**1. Marginal tolerance failures (16 configs, backend `trtllm_fp8_block`)** in the accumulated fuzzer sweep (`FLASHINFER_UMOE_FUZZ=1 FLASHINFER_UMOE_FUZZ_SEED=100`): e.g.
```
trtllm_fp8_block mxfp8_FL_Renormalize_uniform_e32_k4_t127_h3072_i1024_s100: 74/390144 elems exceed tol (rtol=0.85 atol=0.00141; max|diff|=0.002881)
trtllm_fp8_block mxfp8_RenormalizeNaive_imbalanced_e16_k4_t256_h2048_i1024_s900013: 110/524288 elems exceed tol (atol=0.00138; max|diff|=0.002357)
```
Hair's-breadth exceedances (tens of elems out of ~500K, ~1.7-2x atol) that appear/disappear between runs — consistent with autotune-winner nondeterminism flipping marginal cases, i.e. either a tolerance-calibration gap in the new mxfp8 reference/tolerances or a genuinely marginal tactic. Full list of 16 ids available in the #4086 validation logs.

**2. A solo-config core dump**: `FLASHINFER_UMOE_FUZZ=1 FLASHINFER_UMOE_FUZZ_ONLY_SEED=900013 pytest tests/moe/test_unified_moe_fuzz.py` on unpatched main ends in `Aborted (core dumped)` — a single curated mxfp8 config, no accumulation needed. Needs its own triage (assert-class in the fp8-block path? — the same exit-abort decoding as #3957 applies: any device assert poisons the context and terminates at interpreter shutdown).

Both are independent of #3957/#4086 and belong to the owners of the mxfp8/deepseekfp8 unified-MoE integration (#4026-era). The fuzzer had been opt-in until #3958, so these paths had never been swept before.

YangXu1990uiuc:
## Dig results: two corrections and a mechanism hypothesis

[...]

PR #4093, opened 2026-07-22
fix(moe): MxFp8 weight scales must be 128x4 interleaved after the row shuffle (#4087)

YangXu1990uiuc:
## 📌 Description

Fixes the MxFp8 half of #4087. The unified-MoE MxFp8 weight preparation applied **only the row permutation** to W1/W2 scale factors, but the trtllm-gen kernels consume the legacy `shuffle_matrix_sf_a` layout: row permutation **then** 128x4 `block_scale_interleave`. The runner forces `use_shuffled_weight=True` for MxFp8 and the C++ launcher hard-requires it, so ~99.8% of scale bytes were read from the wrong interleaved address. Errors stayed sparse and sub-tolerance in existing tests because random weights repeat E8M0 exponents (a misplaced read often lands on an equal exponent and cancels), the original layout unit test hard-coded the permutation-only layout as expected, and the functional test's `atol=0.05` dwarfs the ~0.0025 output scale. Root-cause credit: external review pass on the #4087 coordinate-dump forensics.

**Changes**
- `prepare.py` MxFp8 branch: interleave the permuted scales for both W1 and W2, then reshape back to the logical `(rows, cols)` view the runner validates (the kernel consumes the raw interleaved bytes) — the legacy pipeline's exact idiom.
- Explicit precondition: the shuffled-weight layout requires `hidden_size` and `2*intermediate_size` divisible by 128 (the same constraint the legacy path carried implicitly); clear `ValueError` instead of a confusing downstream reshape failure.
- `tests/moe/test_mxfp8_weight_prepare_layout.py`: the w2 (non-gated) recipe is compared **byte-exact against legacy `shuffle_matrix_sf_a`** using sentinel data where every scale byte is distinct (with random data, misplacements cancel — the sentinel is what makes the test able to fail); a gated-path guard asserts the shipped layout is not permutation-only.

**Validation (B200-class SM100, main @ 6104afe3)**
- Layout tests: 5/5.
- Targeted fuzzer sweep over all 16 previously-failing configs: **17 passed / 3 failed / 0 skipped** — all 13 mxfp8 configs now genuinely run and pass (pass-vs-skip explicitly verified; an earlier draft of this fix silently broke the runner shape contract and produced skip-mirages, hence the tri-state check). The 3 remaining failures are the non-layout classes tracked on #4087: one nvfp4-FromLogits config and two deepseekfp8 configs (reference-pipeline gaps — DeepSeek reference missing the gate/up fp8 requant before activation; GEMM2 bf16 output store unmodeled).

## 🔍 Related Issues

Fixes the MxFp8 layout half of #4087. The deepseekfp8 reference calibration and the nvfp4-FL s120 failure remain tracked there.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MXFP8 weight-scale layout handling for shuffled weight formats.
  * Added validation for supported tensor dimensions to prevent invalid configurations.
  * Ensured scale data is correctly interleaved and reshaped for affected weight paths.

* **Tests**
  * Added GPU-specific regression coverage to verify scale ordering, layout correctness, and data preservation.

### DIFF
--- a/flashinfer/fused_moe/prepare.py
+++ b/flashinfer/fused_moe/prepare.py
@@ -649,6 +649,17 @@ def prepare_trtllm_fp8_block_weights(
             raise ValueError(
                 "MXFP8 requires hidden_size and intermediate_size divisible by 32."
             )
+        if hidden_size % 128 != 0 or (2 * intermediate_size) % 128 != 0:
+            # The shuffled-weight layout interleaves scale rows in 128-row
+            # tiles; the exact (non-padded) reshape below needs both scale
+            # matrices' row counts (2*intermediate and hidden) to be 128-row
+            # aligned -- same constraint the legacy pipeline carried
+            # implicitly (gh #4087).
+            raise ValueError(
+                "MXFP8 shuffled-weight layout requires hidden_size and "
+                "2*intermediate_size divisible by 128."
+            )
+        from ..fp4_quantization import block_scale_interleave
         from ..quantization.fp8_quantization import mxfp8_quantize
         from .core import (
             _maybe_get_cached_w3_w1_permute_indices,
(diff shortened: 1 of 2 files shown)


flashinfer/fused_moe/prepare.py
```diff
@@ -649,6 +649,17 @@ def prepare_trtllm_fp8_block_weights(
             raise ValueError(
                 "MXFP8 requires hidden_size and intermediate_size divisible by 32."
             )
+        if hidden_size % 128 != 0 or (2 * intermediate_size) % 128 != 0:
```
feih-nv, line 652: is `(2 * intermediate_size) % 128 != 0` correct? I heard `intermediate_size=64` cannot be supported

jdebache:
Related to https://github.com/flashinfer-ai/flashinfer/pull/3882?

PR #4098, opened 2026-07-22
Fix MXFP8 Test: CUTLASS requires a swizzled layout

depaulmillz:
## 📌 Description

This fixes a MXFP8 GEMM test. CUTLASS requires a swizzled layout, but non-swizzled layouts were being tested occasionally leading to IMA.

## 🔍 Related Issues

N/A

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Tests**
  * Updated matrix multiplication test logic to skip unsupported configurations when the required scaling-factor layout is not enabled, including CUTLASS-specific cases and automatic backend selection.  
  * Reduced coverage by running MXFP8 matrix-multiplication tests only with compatible swizzled layouts.  
  * Removed the scale contiguity requirement test.

### DIFF
--- a/tests/gemm/test_mm_mxfp8.py
+++ b/tests/gemm/test_mm_mxfp8.py
@@ -85,13 +85,8 @@ def _run_mm_mxfp8(
 ):
     _skip_if_unsupported(backend)
 
-    compute_capability = get_compute_capability(torch.device("cuda"))
-    is_sm12x = compute_capability[0] == 12
-
-    if is_sm12x and not is_sf_swizzled_layout:
-        pytest.skip(
-            "SM12x only supports swizzled 1D scales; no backend handles non-swizzled layout."
-        )
+    if not is_sf_swizzled_layout:
+        pytest.skip("All backends require swizzled layouts.")
 
     if backend == "trtllm":
         if not is_sf_swizzled_layout:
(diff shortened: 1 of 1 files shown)


tests/gemm/test_mm_mxfp8.py
```diff
@@ -100,9 +100,12 @@ def _run_mm_mxfp8(
             pytest.skip("trtllm does not support non-multiple of 256")
         if out_dtype != torch.bfloat16:
             pytest.skip("trtllm does not support non-bfloat16 output")
-    if backend == "cutlass":
+    # auto also maps to cutlass so validate whether to skip here
+    if backend == "cutlass" or backend == "auto":
```
dhiraj113: not sure this is a fix. It is not necessary that `auto` will always map to `cutlass`.

Can the same fix be done at the flashinfer code level, instead of doing it at the test code level by skipping a test.

depaulmillz: Updated to just remove the testing for swizzled layouts. All backends require swizzled layouts (trtllm, cutlass, and the current auto choice), and it is not possible to detect whether the layout is swizzled from within the mm_mxfp8 API since a `torch.Tensor` alone cannot convey that information.

jdebache:
I think this is a duplicate of: https://github.com/flashinfer-ai/flashinfer/pull/3882. Could you check? I think we also need to adjust the examples, traces, etc... for this to be complete.

depaulmillz:
Closing in favor of https://github.com/flashinfer-ai/flashinfer/pull/3882

PR #4106, opened 2026-07-23
Add distribution-aware autotuning for TRT-LLM MoE

samuel21119:
## 📌 Description

MoE performance depends on routing distribution as well as tensor shape. The existing TRTLLM-Gen autotuner benchmarks tactics with randomly generated routing inputs and reuses one shape-only winner for all requests of that shape. Different expert-skew patterns can change load balance, memory access, and communication behavior, making different tactics optimal.

This MR adds opt-in distribution-aware (DA) autotuning for TRTLLM-Gen MoE.

## 🔍 Related Issues

No.

### ✅ Pre-commit Checks

## 🧪 Tests

### Distribution-aware autotuning

Instead of benchmarking tactics with one random routing distribution, the autotuner generates routing logits from Dirichlet distributions with different concentration parameters, producing varying degrees of expert skew. It benchmarks every fused-MoE configuration (a **tactic**) for every routing distribution and token-count bucket, then records the fastest tactic.

Each routing distribution is represented by an **exemplar**: an *N*-dimensional normalized occupancy vector, where *N* is the number of local experts and each element is that expert's fraction of assigned tokens. The vector is sorted in ascending order so equivalent expert permutations have the same representation. This produces a `(token count, exemplar) -> tactic` mapping. Exemplars that choose the same tactic are merged by averaging their normalized occupancy vectors, reducing both storage and runtime lookup cost.

### Runtime kernel selection

At runtime, the expert assignments are already GPU-resident. The graph counts tokens per local expert, normalizes and sorts the counts to form the runtime occupancy vector, and compares it with the stored exemplars in the matching token bucket using cosine similarity. The nearest exemplar selects the tactic, and a CUDA conditional `SWITCH` graph node dispatches its fixed fused-MoE kernel. All selection work stays on device: no routing-value transfer, host-side lookup, synchronization, or CUDA-graph recapture is required.

## Implementation Details

The generic autotuner is extended with `DynamicValueSpec`, which supplies bucketed routing-value profiles and input generators for value-aware benchmarking. The autotuner records tactic timings for each representative routing distribution while also publishing the winner for the default, caller-provided profile to the existing shape-only cache. This preserves the existing eager-execution behavior and API.

The DA policy and graph runtime are implemented under `flashinfer/fused_moe/dist_aware/`. This code generates Dirichlet-based routing samples, builds the per-token-bucket exemplar plan, and owns plan serialization, compatibility checks, and the baseline-guard decision.

`flashinfer/fused_moe/core.py` integrates DA preparation into the existing public MoE wrappers rather than adding a parallel API. Each wrapper retains its normal validation and runner construction, while additionally providing a fixed-tactic, metadata-based launch path for graph replay. Unsupported configurations, unavailable plans, and eager calls fall back to the original monolithic wrapper path.

`DynamicValueSpec` is used only during autotuning, not for inference-time dispatch. During graph capture, the runtime computes live occupancy from GPU routing data, matches it on device, and selects a pre-captured fixed-tactic body. Outside CUDA graphs, eager execution continues to use the shape-only tactic selected from the default routing profile.

Before publishing a DA plan, a baseline guard compares recorded per-distribution latency against the corresponding shape-only NoDA tactic, including estimated selector/control overhead. If DA does not provide a sufficient win, the plan is rejected and graph capture uses the original monolithic NoDA path instead. DA is experimental and opt-in (`FLASHINFER_DIST_AWARE_AUTOTUNE=1`).

### Enable distribution-aware autotuning

DA-MoE is disabled by default. Enable it before starting the serving process or benchmark:

```bash
export FLASHINFER_DIST_AWARE_AUTOTUNE=1
```

No API change is required: supported TRTLLM blocked-MoE wrappers automatically tune during the normal AutoTuner warmup and realize the resulting plan during CUDA-graph capture. Set the variable to `0` or unset it to disable DA.

### Run the ordinary DeepSeek MoE benchmark

The existing DeepSeek benchmark exercises DA through the ordinary public MoE wrappers. Graph mode is the default; routed mode supplies caller-provided expert IDs and routing weights to the TRTLLM path:

```bash
# Run from the FlashInfer repository root.
export FLASHINFER_DIST_AWARE_AUTOTUNE=1

python -u benchmarks/bench_moe_deepseek.py \
  --num-tokens 64,128,256,512,1024 \
  --num-experts 128 \
  --ep 2 \
  --backends cutedsl,trtllm \
  --distributions ddist:0.1,ddist:1,ddist:2,ddist:3,ddist:4 \
  --routing-input-mode routed
```

`--backends` accepts any subset of `cutedsl,cutlass,trtllm`. Use `--ep` to derive the local expert count from `--num-experts`. `--no-cuda-graph` is an eager diagnostic and does not exercise device-side DA selection. The benchmark sets the DA distribution list from `--distributions` and performs its autotuning warmup before timed graph replay. Routed graph mode explicitly enables DA inside this benchmark; use the dedicated benchmark below for a matched graph-mode NoDA-versus-DA comparison.

### Tune and save the tuning cache

Pass a new JSON path to the normal `autotune(..., cache=path)` context. The cache is loaded on context entry and saved automatically on tuning-context exit. The dedicated benchmark exposes the same path as `--cache` (`--tuning-cache` and the historical `--bundle-output` spelling remain aliases):

```bash
# Run from the FlashInfer repository root.
python benchmarks/bench_trtllm_moe_da.py \
  --precision nvfp4 \
  --distributions uniform,ddist:1.1,ddist:1.5,ddist:2,ddist:3,ddist:4 \
  --num-tokens 8,32,64,128,256,512,1024,2048,4096,8192 \
  --num-experts 256 \
  --local-num-experts 32 \
  --top-k 8 \
  --hidden-size 7168 \
  --intermediate-size 2048 \
  --n-group 8 \
  --topk-group 4 \
  --tune-max-num-tokens 8192 \
  --warmup 3 \
  --iters 10 \
  --execution-mode graph \
  --cache "$artifact_root/tuning-cache.json" \
  --out "$artifact_root/results.csv" \
  --json-out "$artifact_root/results.json"
```

The cache contains ordinary AutoTuner tactics and namespaced DA records in one JSON file. DA records include the exact operation/configuration identity, compiled plan mode, selector exemplars, deduplicated bodies, and graph-free eager selection. The cache is environment-specific. A current-schema record mismatch is ignored and requires retuning; a definite cross-environment metadata conflict is ignored and is not overwritten. Use one writer per cache path while publishing DA records.

The default tuner catalog contains seven distributions (`ddist:1.1`, `ddist:1.3`, `ddist:1.5`, `ddist:1.7`, `ddist:2`, `ddist:2.5`, and `ddist:4`) with one realization each. The dedicated benchmark's replay default is the six-distribution list shown above. At most eight total selector exemplars may be requested, so `number of distributions × FLASHINFER_DA_DISTRIBUTION_SAMPLES` must not exceed eight.

### Reload a saved tuning cache

Use the same cache path and operation/configuration in a fresh process. `autotune(False, cache=path)` loads without profiling. For the dedicated benchmark, repeat the tuning command with `--skip-autotune` and new result paths:

```bash
python benchmarks/bench_trtllm_moe_da.py \
  --precision nvfp4 \
  --distributions uniform,ddist:1.1,ddist:1.5,ddist:2,ddist:3,ddist:4 \
  --num-tokens 8,32,64,128,256,512,1024,2048,4096,8192 \
  --num-experts 256 \
  --local-num-experts 32 \
  --top-k 8 \
  --hidden-size 7168 \
  --intermediate-size 2048 \
  --n-group 8 \
  --topk-group 4 \
  --tune-max-num-tokens 8192 \
  --warmup 1 \
  --iters 10 \
  --execution-mode graph \
  --cache "$artifact_root/tuning-cache.json" \
  --skip-autotune \
  --out "$artifact_root/cache-only-results.csv" \
  --json-out "$artifact_root/cache-only-results.json"
```

The public wrapper restores compatible DA state automatically before capture. The cache-only benchmark fails instead of silently profiling when the requested operation/configuration has no compatible saved record.

### Expected tuning time

Exact-current Full AC tuning and fresh-process cache times are below. The run reused a warm JIT workspace after the focused CUDA suites; these values measure the requested tuning/cache transactions rather than first-use compiler and download latency. They are diagnostic because the authorized node was oversubscribed.

| Precision | NoDA fresh (s) | DA fresh (s) | NoDA cache (s) | DA cache (s) |
|---|---:|---:|---:|---:|
| NVFP4 | 19.719 | 24.148 | 0.506 | 0.162 |
| MXFP4 | 27.123 | 28.500 | 0.510 | 0.176 |
| W4A16 | 10.874 | 69.608 | 0.512 | 1.158 |
| BF16 | 8.746 | 25.762 | 0.376 | 0.147 |
| FP8 per-tensor | 5.397 | 20.570 | 1.641 | 1.183 |
| FP8 block | 4.179 | 25.965 | 0.391 | 0.168 |
| MXFP8 | 32.849 | 38.034 | 0.511 | 0.173 |
| MXINT4 | 9.261 | 49.403 | 0.515 | 0.171 |

### Ordinary public-wrapper benchmark

The `benchmarks/bench_moe_deepseek.py` workflow above was measured on 1000 W B200 nodes with a 128-expert, EP=2 DeepSeek-shaped MoE, routed input, tokens 64–1024, and distributions `ddist:0.1`, `ddist:1`, `ddist:2`, `ddist:3`, and `ddist:4`. It produced 25 paired graph-mode TRTLLM rows. Each cell is the minimum–maximum `DA-disabled TRTLLM latency / DA-enabled TRTLLM latency` across the five distributions; values above 1.0 favor DA.

| 64 | 128 | 256 | 512 | 1024 |
|---:|---:|---:|---:|---:|
| 1.004–1.007x | 1.004–1.008x | 1.506–2.552x | 0.976–1.000x | 1.488–1.655x |

### Dedicated NoDA-versus-DA benchmark

DeepSeek-V3 EP8 shape: 256 global experts, 32 local experts, top-k 8, hidden size 7168, intermediate size 2048, `n-group=8`, and `topk-group=4`. It evaluates eight precision families, six routing distributions, and ten token counts with three warmups and the median of ten CUDA-graph replays per row.

| Precision | 8 | 32 | 64 | 128 | 256 | 512 | 1024 | 2048 | 4096 | 8192 | Overall |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| NVFP4 | 0.9974x | 1.0031x | 1.0163x | 1.1177x | 1.1160x | 1.0008x | 0.9983x | 0.9995x | 0.9932x | 0.9952x | 1.0227x |
| MXFP4 | 1.0063x | 1.0033x | 1.0661x | 1.3152x | 1.3387x | 0.9879x | 0.9895x | 1.0238x | 0.9948x | 0.9956x | 1.0651x |
| W4A16 | 1.0004x | 1.0220x | 1.2578x | 1.7240x | 1.4206x | 1.5738x | 0.9887x | 0.9983x | 0.9887x | 0.9740x | 1.1677x |
| BF16 | 1.0031x | 1.0063x | 1.0928x | 1.3677x | 1.2830x | 1.1377x | 1.1969x | 1.0281x | 0.9917x | 0.9647x | 1.0999x |
| FP8 per-tensor | 1.0046x | 1.0384x | 1.1651x | 1.1941x | 1.1557x | 1.1008x | 1.2138x | 0.9905x | 1.0100x | 1.0027x | 1.0844x |
| FP8 block | 1.0034x | 1.0807x | 1.3598x | 1.3423x | 1.1516x | 0.9923x | 1.1187x | 1.1861x | 0.9615x | 0.9850x | 1.1101x |
| MXFP8 | 1.0007x | 1.0654x | 1.2058x | 1.2346x | 1.2281x | 0.9996x | 1.0015x | 0.9721x | 1.0038x | 0.9885x | 1.0653x |
| MXINT4 | 1.0024x | 1.4091x | 2.0148x | 2.0042x | 1.5466x | 1.1914x | 1.3027x | 0.9778x | 0.9800x | 0.9822x | 1.2913x |

Fresh compiled plan outcomes were 330 `da_single_body`, 114 `da_fallback`, and 36 `da_switch` rows. Actual capture outcomes were 330 single-body, 114 fallback, 30 switch, and 6 pristine `noda_capture_fallback` rows. The carried per-row performance waiver remains explicit; the aggregate table is release evidence, not a claim that every row speeds up.

## Summary by CodeRabbit

* **New Features**
  * Added distribution-aware MoE routing/autotuning with CUDA Graph capture improvements and optional baseline-guard behavior.
  * Introduced value-aware autotuning (value buckets/profiling) with reusable profiling bundles.
  * Added routing-metadata replay execution across BF16/FP8/FP4/MXINT4, plus multi-tile DeepSeek routing on supported GPUs.
  * Expanded MoE benchmarks and CLI options (routing modes, distributions, precisions) with CSV reporting.

* **Bug Fixes**
  * Improved FP4 tensor dtype deduction for more scale layouts.
  * Corrected cuBLASLt handle management in FP8 paths.

* **Documentation**
  * Documented new distribution-aware and value-aware autotuning/dispatch environment options.


aleozlx:
@claude review

samuellees:
@samuel21119 Could you please address claude's comment and fix the conflict? thanks!

samuellees:
@claude review with insight: 1. Route-map padding
   main commit 8b6af1aa (#4237) pads `permuted_idx_to_token_idx` by one
   `int32` because the shipped routed-BMM cubins read one element past the
   logical end. The three new metadata allocations in
   `csrc/trtllm_fused_moe_kernel_launcher.cu:3759,3877,4202` still allocate
   exactly `max_num_padded_tokens`. Please check whether all three require
   `+1`, together with the validator at :315-316 and the fake shape in
   `flashinfer/fused_moe/core.py:4058`.

2. Multi-tile expert-count output
   `trtllm_moe_allocate_routing_metadata_multi_tile` passes
   `expert_count_histogram` as `mPtrExpertCounts` at launcher :4231.
   `routingIndicesMultiTileClusterKernel` only calls `routingPermutation`;
   that function appears to keep its histogram in shared memory without
   writing `mPtrExpertCounts`. Please confirm whether
   `FLASHINFER_DA_KNN_SELECT_FROM_ROUTING_COUNTS=1` therefore reads an
   unwritten metadata tensor in `da_capture.py:1136-1142`.

3. Non-power-of-two expert counts
   `da_knn_compact_counts_for_sort*` initializes `counts_int` only through
   `num_local_experts`, while `da_knn_finish_selection_from_counts` rounds
   `sort_items` up to a power of two. For 17 nonzero local experts,
   `sort_len` becomes 32 and `counts_int[17:32]` appears uninitialized before
   bitonic sort (`da_heuristic.cuh:244-247,358-361,446-453`). Please verify
   this case and the current test coverage.

4. FP32 unpacked routing weights
   The multi-tile API accepts BF16 or FP32 weights at launcher :4153-4158,
   but `can_use_multi_cluster` at :4164-4169 does not gate the weight dtype.
   The selected launcher later requires BF16 output at
   `trtllm_fused_moe_routing_deepseek.cu:510-514`. Please check whether a
   valid FP32 input enters this path and is then rejected instead of falling
   back.

5. SM120 eligibility
   Base TRT-LLM MoE accepts SM10.x and SM12.x at launcher :932-944, and the
   JIT spec builds major versions 10 and 12. DA eligibility in
   `da_state.py:134-145` has no architecture check, while the three metadata
   entry points at launcher :3742,3861,4142 require major 10. Please verify
   whether enabling DA on SM120 reaches these entry points instead of
   falling back before metadata preparation.

6. Distributed graph-profile timing
   `set_autotune_process_group()` promises a per-tactic reduction so ranks
   choose the same tactic. The reduction is in `_profile_single_kernel`
   (`autotuner.py:3420-3433`), but value-aware CUDA-graph profiling calls
   `session.measure()` directly at :2891-2924. Please check whether the
   process-group contract is skipped in the default DA graph-profile path.

7. Tactic blocklist in factorized tuning
   Initial factorized candidates are filtered at `autotuner.py:2868-2870`.
   The composed tactic at :3100-3107 and the exhaustive fallback candidates
   at :3127-3135 do not appear to pass through `_blocklist.filter`. Please
   verify whether a blocklisted tactic can be profiled in either path.

8. Bundle loading and round trip
   Shared `tune_and_publish_tactic()` calls `maybe_prepare_bundle()` only
   after value-aware tuning at `da_core.py:739-740`. FP4 has a separate
   preload at `core.py:3614-3616`, while the BF16/FP8/MXINT4 callers only use
   the shared helper. Please verify whether a fresh non-tuning process with
   an existing `FLASHINFER_DA_BUNDLE` loads it for those precisions.
   The writer always stores `"tactic_table": {}` at
   `da_profile.py:1647-1651`. The loader sets `_bundle_has_tactics` only when
   this table is nonempty at :534-540, and `da_core.py:390-397` uses that flag
   to suppress value profiling. Please check whether a nonempty bundle
   generated by this PR is loaded but then fully profiled again.

9. Current-main integration
    The PR currently conflicts with main in the FP4 `MoERunner` construction:
    main adds `num_fused_shared_experts`, while the PR adds
    `factorized_da_enabled`. After retaining both fields, DA capture still
    validates routing resources as exactly `[num_tokens, top_k]` at
    `da_capture.py:180-190`, whereas main can widen them to
    `top_k + num_fused_shared_experts`. Please check whether DA should be
    explicitly disabled for fused shared experts until that dimension is
    represented throughout the DA context and metadata ABI.

For confirmed items, please suggest minimal fixes and focused regression
tests. It would also be useful to distinguish code defects from unsupported
but intentional configurations.

feih-nv:
`FP8 per-tensor is excluded from this routed-input benchmark because its public API accepts logits rather than routed IDs and weights.`
I think this is already supported via #4088 by @jdebache .

rosenrodt:
Fixed in 5d0740e2c

> 1. Route-map padding

All routing-metadata allocation paths now allocate permuted_idx_to_token_idx with max_num_padded_tokens + 1. This will be removed after fixed cubins land.

> 2. Multi-tile expert-count output

Added fix for FLASHINFER_DA_KNN_SELECT_FROM_ROUTING_COUNTS=1 mode. That mode is not a default path and will be considered for removal. 

> 3. Non-power-of-two expert counts

Fair. We see more non-pow2 expert count especially for latent MoE models (Kimi-K3). Fixed.

> 4. FP32 unpacked routing weights

Fixed.

> 5. SM120 eligibility

No support for SM120. Mainly because the first supported MoE backend is SM10x only.

DA kernel dispatch has optimizations tied to TRTLLM MoE-like routing metadata. One of those is to speculatively compute routing metadata for all candidate tileN runners, while the DA dispatcher finalizes which tileN runner to choose from. 

IIUC CuteDSL SM10x MoE follows similar routing metadata structure. DA-MoE can be extended to that in the future as well.

> 6. Distributed graph-profile timing

Both ordinary and value-aware CUDA-graph measurements now use the same once-per-tactic process-group reduction helper.

> 7. Tactic blocklist in factorized tuning

Fixed and tested in [test_factorized_value_profile_never_profiles_blocklisted_composed_tactic](https://github.com/samuel21119/flashinfer/blob/5d0740e2cc25583c44d155989232b48495f33572/tests/autotuner/test_autotuner_core.py#L1267-L1335)

> 8. Bundle loading and round trip

Fixed in 5d0740e2c

> 9. Current-main integration

Fused shared experts is currently out of scope and disabled for DA-MoE. Will revisit.

rosenrodt:
@feih-nv 

> FP8 per-tensor is excluded from this routed-input benchmark because its public API accepts logits rather than routed IDs and weights.

Thanks. I see it was merged just days ago. The huge PR cannot keep up :) 

Will address that in follow-up.

feih-nv:
Additional review findings:

1. **DA replay drops optional GEMM1 activation parameters**

   The normal BF16 and FP8 block paths pass `gemm1_alpha`, `gemm1_beta`, and `gemm1_clamp_limit`, but their `_from_routing_metadata` replay APIs omit them (`flashinfer/fused_moe/core.py:2197-2248`, `:3189-3246`).

   CUDA-graph replay can therefore silently differ from eager execution. Please plumb these tensors through both replay APIs or disable replay when any are non-`None`, with graph-replay tests.

2. **Small-token tactic enumeration changes behavior with DA disabled**

   `MoERunner.get_valid_tactics()` returns no tactics on SM10x when the expanded token count is less than twice the expert count (`flashinfer/fused_moe/core.py:1584-1590`).

   This guard is not conditioned on DA. For a 256-expert, top-8 model, profiles below 64 tokens skip ordinary autotuning and use the built-in default tactic. Please justify this with measurements, scope it to DA, or move it into a separate performance change.

3. **New DA entry points lack `CUDADeviceGuard`**

   Existing launcher paths establish the tensor’s CUDA device, but the new metadata allocation, population, replay, and heuristic entry points do not. Representative locations are `csrc/trtllm_fused_moe_kernel_launcher.cu:3766`, `:3890`, `:3995`, and `:4077`.

   When the ambient device differs from the tensor device, these paths can use an incompatible stream/context. Please add device guards following the existing launcher convention and test alternating devices in one process.

4. **Capture resources have no eviction policy**

   `CAPTURE_KEEPALIVE` retains routing tensors indefinitely (`flashinfer/fused_moe/dist_aware/da_state.py:173-190`). Existing capture keys are reused, but shape or context churn permanently retains superseded metadata.

   Please define resource ownership and eviction when `CAPTURE_RESOURCES` entries are replaced, with a test covering repeated shape/context changes.

5. **Baseline fallback misses are cached permanently**

   `resolve_static_fallback()` stores `None` in `STATIC_FALLBACK_TACTICS` when NoDA timing is not available yet (`flashinfer/fused_moe/dist_aware/da_profile.py:962-1027`).

   An eager lookup before autotuning can therefore permanently suppress later baseline discovery, causing the guard to publish an unprotected DA plan with `missing_noda_baseline`. Please avoid caching misses or invalidate them when profiling results are published.

6. **Bundle loading can leave partially published state**

   `load_knn_v2_bundle()` publishes per-bucket Python and C++ selector state incrementally, but can later return failure for an invalid bucket (`flashinfer/fused_moe/dist_aware/da_profile.py:536-676`).

   The caller treats this as a failed load without rolling back earlier mutations. Please validate the full bundle before publication or explicitly roll back all state on failure.

7. **DA environment parsing is inconsistent**

   `flashinfer/autotuner/autotuner.py:57-58` enables value-aware tuning only for the exact value `"1"`, while `dist_aware/da_config.py:12-16` accepts broader truthy values.

   For example, `FLASHINFER_DIST_AWARE_AUTOTUNE=true` enables runtime/capture orchestration while skipping value-profile generation. Please use one shared boolean parser.

feih-nv:
Lower-priority follow-ups (by gpt-5.6) in case they make sense:

1. **Subprocess tests report PASS when the child skips**

   Several tests in `tests/moe/test_trtllm_da_autotune.py` run pytest in a subprocess and check only `returncode == 0`, while `_require_sm100()` runs inside the child (`:690`, `:837`, `:971`, `:1085`, and `:4091`).

   Pytest exits successfully for skipped tests, so non-SM100 CI can report these tests as passed without exercising CUDA graphs. Please check the architecture before spawning or verify an explicit child-execution marker.

2. **Routing IDs lack symmetric device and layout validation**

   The metadata APIs validate `topk_weights` for CUDA placement and contiguity but do not apply equivalent checks to `topk_ids` or `packed_topk_ids` (`csrc/trtllm_fused_moe_kernel_launcher.cu:3771-3798`).

   Please validate dtype, CUDA device, matching device, and contiguity for every routing input before passing its pointer to a kernel.

3. **DA global state is not thread-safe**

   Selector handles and the tactic, baseline-decision, bundle-context, and capture-resource maps are updated without synchronization (`flashinfer/fused_moe/dist_aware/da_state.py:155-178`).

   If concurrent initialization is supported, please protect state publication with a lock. Otherwise, document and enforce single-threaded initialization.

4. **`DynamicValueSpec.__eq__` compares hashes**

   `flashinfer/autotuner/autotuner.py:581-582` treats two specifications as equal when their hashes match. A collision can therefore make structurally different tuning configurations compare equal.

   Please construct one canonical field key and use it for both hashing and equality.

5. **DA-disabled calls still rebuild configuration**

   The main MoE wrappers construct a fresh `DAConfig` on every invocation (`flashinfer/fused_moe/core.py:2035`, `:2431`, `:3003`, `:3531`, and `:4348`), repeatedly parsing environment variables even when DA is disabled.

   Consider caching the environment-derived default configuration or adding a fast feature-disabled check before constructing DA execution objects.

6. **Auto-profile callbacks retain large tensors through positional coupling**

   `register_auto_profile_callback()` captures weights, biases, and scales until the autotune context exits (`flashinfer/fused_moe/dist_aware/da_profile.py:1700-1799`). `da_core.py:660-661` also retrieves `per_token_scale` through the positional expression `tuning_inputs[7]`.

   Please minimize the callback capture set and replace the positional index with a named field to avoid accidental lifetime extension and silent breakage if input ordering changes.

7. **Some public diagnostics and controls are inert**

   `FLASHINFER_DA_ROUTING_CONCURRENT` is parsed by an unused function in `csrc/trtllm_fused_moe_da_heuristic.cu:375-384`. Similarly, `get_da_fast_path_stats()` exposes `fire_multi`, `fire_single`, and `skip`, but those values are hardcoded in `da_capture.py:1213-1221`.

   Please wire these surfaces to real behavior or remove them until implemented.

8. **Distribution-generation helpers have duplicate implementations**

   `flashinfer/fused_moe/utils.py` duplicates distribution-assignment logic from `dist_aware/da_utils.py`, and the implementations already differ in some branches.

   Please keep one canonical implementation and re-export it where needed so benchmarks and the autotuner cannot drift.

9. **Unsupported expert counts can emit repeated warnings**

   `maxTokensMultiTileCluster()` is evaluated before the caller’s expert-count eligibility check. Models with unsupported expert counts can therefore emit a warning on every attempted eligibility check even though the multi-tile path is subsequently rejected.

   Please check the supported expert range before calling the warning-producing helper, or make the helper silent when used only as a predicate.

flashinfer/autotuner/autotuner.py
```diff
@@ -1737,11 +3247,69 @@ def choose_one(
...
+        return count
+
+    @staticmethod
+    def _map_dynamic_value_bucket(
+        spec: DynamicValueSpec,
+        inputs: List[torch.Tensor],
+        kwargs: Dict[str, Any],
+    ) -> int:
+        tensor = inputs[spec.input_idx]
+        num_params = len(inspect.signature(spec.map_to_value_bucket).parameters)
```
qiching: `_map_dynamic_value_bucket` runs on the non capturing `choose_one` path like once per MoE call at inference when DA is enabled, and it calls inspect.signature.parameters every time so pure host overhead on the dispatch critical path. 

Please precompute the callable arity once like in DynamicValueSpec.__post_init__ and store an int, so the runtime path is just a comparison. Same pattern shows up for the generator arity probes in `_apply_value_specs` and the value-sample loop, though those are tuning-time and lower priority.

qiching (changes requested):

flashinfer/autotuner/autotuner.py
```diff
@@ -1605,10 +2806,82 @@ def choose_one(
...
+                                    tuning_config,
+                                    r.get_cache_key_extras(tensors),
+                                )
+                                default_entry = self.profiling_cache.get(
+                                    shape_cache_key
+                                )
+                                if default_entry is not None:
+                                    default_tactic, _ = default_entry
+                                    try:
+                                        matches_profile_tile = int(
```
qiching: Suggest moving this into a TunableRunner hook so the DA runner owns the semantics and the core tuner stays layout agnostic.

rosenrodt: Addressed in a5df03f29. It now uses more DA-agnostic semantics:

```
measurement_tactics = r.get_measurement_only_tactics(default_tactic, tensors, p)
```

get_measurement_only_tactics() can be overridden by actual runners (e.g. MoERunners in flashinfer/fused_moe/core.py).

flashinfer/autotuner/autotuner.py
rosenrodt: > Please precompute the callable arity 

Addressed in a5df03f29. Thanks!

rosenrodt:
> * `tests.moe.test_trtllm_gen_fused_moe` — 20107 failures on RTX Pro 6000 Blackwell / CUDA 12.9, RTX Pro 6000 Blackwell / CUDA 13.0
>   
>   * RuntimeError: No CUDA GPUs are available
> * `tests.moe.test_trtllm_da_autotune` — 28 failures on RTX Pro 6000 Blackwell / CUDA 12.9, RTX Pro 6000 Blackwell / CUDA 13.0
>   
>   * RuntimeError: No CUDA GPUs are available

Error is node-specific. I don't repro the error locally.

rosenrodt:
@qiching @aleozlx is there any more changes we need to make?

@samuel21119 could you help update the PR description from `## Usage` and below? I refactored the PR so the tuning time vastly improves while maintaining same DA speedup. 

---

## Usage

### Enable distribution-aware autotuning

DA-MoE is disabled by default. Enable it before starting the serving process or benchmark:

```bash
export FLASHINFER_DIST_AWARE_AUTOTUNE=1
```

No API change is required: supported TRTLLM blocked-MoE wrappers automatically tune during the normal AutoTuner warmup and realize the resulting plan during CUDA-graph capture. Set the variable to `0` or unset it to disable DA.

### Run the ordinary DeepSeek MoE benchmark

The existing DeepSeek benchmark exercises DA through the ordinary public MoE wrappers. Graph mode is the default; routed mode supplies caller-provided expert IDs and routing weights to the TRTLLM path:

```bash
# Run from the FlashInfer repository root.
export FLASHINFER_DIST_AWARE_AUTOTUNE=1

python -u benchmarks/bench_moe_deepseek.py \
  --num-tokens 64,128,256,512,1024 \
  --num-experts 128 \
  --ep 2 \
  --backends cutedsl,trtllm \
  --distributions ddist:0.1,ddist:1,ddist:2,ddist:3,ddist:4 \
  --routing-input-mode routed
```

`--backends` accepts any subset of `cutedsl,cutlass,trtllm`. Use `--ep` to derive the local expert count from `--num-experts`. `--no-cuda-graph` is an eager diagnostic and does not exercise device-side DA selection. The benchmark sets the DA distribution list from `--distributions` and performs its autotuning warmup before timed graph replay. Routed graph mode explicitly enables DA inside this benchmark; use the dedicated benchmark below for a matched graph-mode NoDA-versus-DA comparison.

### Tune and save the tuning cache

Pass a new JSON path to the normal `autotune(..., cache=path)` context. The cache is loaded on context entry and saved automatically on tuning-context exit. The dedicated benchmark exposes the same path as `--cache` (`--tuning-cache` and the historical `--bundle-output` spelling remain aliases):

```bash
# Run from the FlashInfer repository root.
python benchmarks/bench_trtllm_moe_da.py \
  --precision nvfp4 \
  --distributions uniform,ddist:1.1,ddist:1.5,ddist:2,ddist:3,ddist:4 \
  --num-tokens 8,32,64,128,256,512,1024,2048,4096,8192 \
  --num-experts 256 \
  --local-num-experts 32 \
  --top-k 8 \
  --hidden-size 7168 \
  --intermediate-size 2048 \
  --n-group 8 \
  --topk-group 4 \
  --tune-max-num-tokens 8192 \
  --warmup 3 \
  --iters 10 \
  --execution-mode graph \
  --cache "$artifact_root/tuning-cache.json" \
  --out "$artifact_root/results.csv" \
  --json-out "$artifact_root/results.json"
```

The cache contains ordinary AutoTuner tactics and namespaced DA records in one JSON file. DA records include the exact operation/configuration identity, compiled plan mode, selector exemplars, deduplicated bodies, and graph-free eager selection. The cache is environment-specific. A current-schema record mismatch is ignored and requires retuning; a definite cross-environment metadata conflict is ignored and is not overwritten. Use one writer per cache path while publishing DA records.

The default tuner catalog contains seven distributions (`ddist:1.1`, `ddist:1.3`, `ddist:1.5`, `ddist:1.7`, `ddist:2`, `ddist:2.5`, and `ddist:4`) with one realization each. The dedicated benchmark's replay default is the six-distribution list shown above. At most eight total selector exemplars may be requested, so `number of distributions × FLASHINFER_DA_DISTRIBUTION_SAMPLES` must not exceed eight.

### Reload a saved tuning cache

Use the same cache path and operation/configuration in a fresh process. `autotune(False, cache=path)` loads without profiling. For the dedicated benchmark, repeat the tuning command with `--skip-autotune` and new result paths:

```bash
python benchmarks/bench_trtllm_moe_da.py \
  --precision nvfp4 \
  --distributions uniform,ddist:1.1,ddist:1.5,ddist:2,ddist:3,ddist:4 \
  --num-tokens 8,32,64,128,256,512,1024,2048,4096,8192 \
  --num-experts 256 \
  --local-num-experts 32 \
  --top-k 8 \
  --hidden-size 7168 \
  --intermediate-size 2048 \
  --n-group 8 \
  --topk-group 4 \
  --tune-max-num-tokens 8192 \
  --warmup 1 \
  --iters 10 \
  --execution-mode graph \
  --cache "$artifact_root/tuning-cache.json" \
  --skip-autotune \
  --out "$artifact_root/cache-only-results.csv" \
  --json-out "$artifact_root/cache-only-results.json"
```

The public wrapper restores compatible DA state automatically before capture. The cache-only benchmark fails instead of silently profiling when the requested operation/configuration has no compatible saved record.

### Expected tuning time

Exact-current Full AC tuning and fresh-process cache times are below. The run reused a warm JIT workspace after the focused CUDA suites; these values measure the requested tuning/cache transactions rather than first-use compiler and download latency. They are diagnostic because the authorized node was oversubscribed.

| Precision | NoDA fresh (s) | DA fresh (s) | NoDA cache (s) | DA cache (s) |
|---|---:|---:|---:|---:|
| NVFP4 | 19.719 | 24.148 | 0.506 | 0.162 |
| MXFP4 | 27.123 | 28.500 | 0.510 | 0.176 |
| W4A16 | 10.874 | 69.608 | 0.512 | 1.158 |
| BF16 | 8.746 | 25.762 | 0.376 | 0.147 |
| FP8 per-tensor | 5.397 | 20.570 | 1.641 | 1.183 |
| FP8 block | 4.179 | 25.965 | 0.391 | 0.168 |
| MXFP8 | 32.849 | 38.034 | 0.511 | 0.173 |
| MXINT4 | 9.261 | 49.403 | 0.515 | 0.171 |

## Benchmark Results

### Ordinary public-wrapper benchmark

The `benchmarks/bench_moe_deepseek.py` workflow above was measured on 1000 W B200 nodes with a 128-expert, EP=2 DeepSeek-shaped MoE, routed input, tokens 64–1024, and distributions `ddist:0.1`, `ddist:1`, `ddist:2`, `ddist:3`, and `ddist:4`. It produced 25 paired graph-mode TRTLLM rows. Each cell is the minimum–maximum `DA-disabled TRTLLM latency / DA-enabled TRTLLM latency` across the five distributions; values above 1.0 favor DA.

| 64 | 128 | 256 | 512 | 1024 |
|---:|---:|---:|---:|---:|
| 1.004–1.007x | 1.004–1.008x | 1.506–2.552x | 0.976–1.000x | 1.488–1.655x |

### Dedicated NoDA-versus-DA benchmark

DeepSeek-V3 EP8 shape: 256 global experts, 32 local experts, top-k 8, hidden size 7168, intermediate size 2048, `n-group=8`, and `topk-group=4`. It evaluates eight precision families, six routing distributions, and ten token counts with three warmups and the median of ten CUDA-graph replays per row.


| Precision | 8 | 32 | 64 | 128 | 256 | 512 | 1024 | 2048 | 4096 | 8192 | Overall |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| NVFP4 | 0.9974x | 1.0031x | 1.0163x | 1.1177x | 1.1160x | 1.0008x | 0.9983x | 0.9995x | 0.9932x | 0.9952x | 1.0227x |
| MXFP4 | 1.0063x | 1.0033x | 1.0661x | 1.3152x | 1.3387x | 0.9879x | 0.9895x | 1.0238x | 0.9948x | 0.9956x | 1.0651x |
| W4A16 | 1.0004x | 1.0220x | 1.2578x | 1.7240x | 1.4206x | 1.5738x | 0.9887x | 0.9983x | 0.9887x | 0.9740x | 1.1677x |
| BF16 | 1.0031x | 1.0063x | 1.0928x | 1.3677x | 1.2830x | 1.1377x | 1.1969x | 1.0281x | 0.9917x | 0.9647x | 1.0999x |
| FP8 per-tensor | 1.0046x | 1.0384x | 1.1651x | 1.1941x | 1.1557x | 1.1008x | 1.2138x | 0.9905x | 1.0100x | 1.0027x | 1.0844x |
| FP8 block | 1.0034x | 1.0807x | 1.3598x | 1.3423x | 1.1516x | 0.9923x | 1.1187x | 1.1861x | 0.9615x | 0.9850x | 1.1101x |
| MXFP8 | 1.0007x | 1.0654x | 1.2058x | 1.2346x | 1.2281x | 0.9996x | 1.0015x | 0.9721x | 1.0038x | 0.9885x | 1.0653x |
| MXINT4 | 1.0024x | 1.4091x | 2.0148x | 2.0042x | 1.5466x | 1.1914x | 1.3027x | 0.9778x | 0.9800x | 0.9822x | 1.2913x |

Fresh compiled plan outcomes were 330 `da_single_body`, 114 `da_fallback`, and 36 `da_switch` rows. Actual capture outcomes were 330 single-body, 114 fallback, 30 switch, and 6 pristine `noda_capture_fallback` rows. The carried per-row performance waiver remains explicit; the aggregate table is release evidence, not a claim that every row speeds up.

samuel21119:
> @qiching @aleozlx is there any more changes we need to make?
> 
> @samuel21119 could you help update the PR description from `## Usage` and below? I refactored the PR so the tuning time vastly improves while maintaining same DA speedup.
> ## Usage
> ### Enable distribution-aware autotuning
> 
> DA-MoE is disabled by default. Enable it before starting the serving process or benchmark:
> 
> ```shell
> export FLASHINFER_DIST_AWARE_AUTOTUNE=1
> ```
> 
> No API change is required: supported TRTLLM blocked-MoE wrappers automatically tune during the normal AutoTuner warmup and realize the resulting plan during CUDA-graph capture. Set the variable to `0` or unset it to disable DA.
> ### Run the ordinary DeepSeek MoE benchmark
> 
> The existing DeepSeek benchmark exercises DA through the ordinary public MoE wrappers. Graph mode is the default; routed mode supplies caller-provided expert IDs and routing weights to the TRTLLM path:
> 
> ```shell
> # Run from the FlashInfer repository root.
> export FLASHINFER_DIST_AWARE_AUTOTUNE=1
> 
> python -u benchmarks/bench_moe_deepseek.py \
>   --num-tokens 64,128,256,512,1024 \
>   --num-experts 128 \
>   --ep 2 \
>   --backends cutedsl,trtllm \
>   --distributions ddist:0.1,ddist:1,ddist:2,ddist:3,ddist:4 \
>   --routing-input-mode routed
> ```
> 
> `--backends` accepts any subset of `cutedsl,cutlass,trtllm`. Use `--ep` to derive the local expert count from `--num-experts`. `--no-cuda-graph` is an eager diagnostic and does not exercise device-side DA selection. The benchmark sets the DA distribution list from `--distributions` and performs its autotuning warmup before timed graph replay. Routed graph mode explicitly enables DA inside this benchmark; use the dedicated benchmark below for a matched graph-mode NoDA-versus-DA comparison.
> ### Tune and save the tuning cache
> 
> Pass a new JSON path to the normal `autotune(..., cache=path)` context. The cache is loaded on context entry and saved automatically on tuning-context exit. The dedicated benchmark exposes the same path as `--cache` (`--tuning-cache` and the historical `--bundle-output` spelling remain aliases):
> 
> ```shell
> # Run from the FlashInfer repository root.
> python benchmarks/bench_trtllm_moe_da.py \
>   --precision nvfp4 \
>   --distributions uniform,ddist:1.1,ddist:1.5,ddist:2,ddist:3,ddist:4 \
>   --num-tokens 8,32,64,128,256,512,1024,2048,4096,8192 \
>   --num-experts 256 \
>   --local-num-experts 32 \
>   --top-k 8 \
>   --hidden-size 7168 \
>   --intermediate-size 2048 \
>   --n-group 8 \
>   --topk-group 4 \
>   --tune-max-num-tokens 8192 \
>   --warmup 3 \
>   --iters 10 \
>   --execution-mode graph \
>   --cache "$artifact_root/tuning-cache.json" \
>   --out "$artifact_root/results.csv" \
>   --json-out "$artifact_root/results.json"
> ```
> 
> The cache contains ordinary AutoTuner tactics and namespaced DA records in one JSON file. DA records include the exact operation/configuration identity, compiled plan mode, selector exemplars, deduplicated bodies, and graph-free eager selection. The cache is environment-specific. A current-schema record mismatch is ignored and requires retuning; a definite cross-environment metadata conflict is ignored and is not overwritten. Use one writer per cache path while publishing DA records.
> 
> The default tuner catalog contains seven distributions (`ddist:1.1`, `ddist:1.3`, `ddist:1.5`, `ddist:1.7`, `ddist:2`, `ddist:2.5`, and `ddist:4`) with one realization each. The dedicated benchmark's replay default is the six-distribution list shown above. At most eight total selector exemplars may be requested, so `number of distributions × FLASHINFER_DA_DISTRIBUTION_SAMPLES` must not exceed eight.
> ### Reload a saved tuning cache
> 
> Use the same cache path and operation/configuration in a fresh process. `autotune(False, cache=path)` loads without profiling. For the dedicated benchmark, repeat the tuning command with `--skip-autotune` and new result paths:
> 
> ```shell
> python benchmarks/bench_trtllm_moe_da.py \
>   --precision nvfp4 \
>   --distributions uniform,ddist:1.1,ddist:1.5,ddist:2,ddist:3,ddist:4 \
>   --num-tokens 8,32,64,128,256,512,1024,2048,4096,8192 \
>   --num-experts 256 \
>   --local-num-experts 32 \
>   --top-k 8 \
>   --hidden-size 7168 \
>   --intermediate-size 2048 \
>   --n-group 8 \
>   --topk-group 4 \
>   --tune-max-num-tokens 8192 \
>   --warmup 1 \
>   --iters 10 \
>   --execution-mode graph \
>   --cache "$artifact_root/tuning-cache.json" \
>   --skip-autotune \
>   --out "$artifact_root/cache-only-results.csv" \
>   --json-out "$artifact_root/cache-only-results.json"
> ```
> 
> The public wrapper restores compatible DA state automatically before capture. The cache-only benchmark fails instead of silently profiling when the requested operation/configuration has no compatible saved record.
> ### Expected tuning time
> 
> Exact-current Full AC tuning and fresh-process cache times are below. The run reused a warm JIT workspace after the focused CUDA suites; these values measure the requested tuning/cache transactions rather than first-use compiler and download latency. They are diagnostic because the authorized node was oversubscribed.
> Precision 	NoDA fresh (s) 	DA fresh (s) 	NoDA cache (s) 	DA cache (s)
> NVFP4 	19.719 	24.148 	0.506 	0.162
> MXFP4 	27.123 	28.500 	0.510 	0.176
> W4A16 	10.874 	69.608 	0.512 	1.158
> BF16 	8.746 	25.762 	0.376 	0.147
> FP8 per-tensor 	5.397 	20.570 	1.641 	1.183
> FP8 block 	4.179 	25.965 	0.391 	0.168
> MXFP8 	32.849 	38.034 	0.511 	0.173
> MXINT4 	9.261 	49.403 	0.515 	0.171
> ## Benchmark Results
> ### Ordinary public-wrapper benchmark
> 
> The `benchmarks/bench_moe_deepseek.py` workflow above was measured on 1000 W B200 nodes with a 128-expert, EP=2 DeepSeek-shaped MoE, routed input, tokens 64–1024, and distributions `ddist:0.1`, `ddist:1`, `ddist:2`, `ddist:3`, and `ddist:4`. It produced 25 paired graph-mode TRTLLM rows. Each cell is the minimum–maximum `DA-disabled TRTLLM latency / DA-enabled TRTLLM latency` across the five distributions; values above 1.0 favor DA.
> 64 	128 	256 	512 	1024
> 1.004–1.007x 	1.004–1.008x 	1.506–2.552x 	0.976–1.000x 	1.488–1.655x
> ### Dedicated NoDA-versus-DA benchmark
> 
> DeepSeek-V3 EP8 shape: 256 global experts, 32 local experts, top-k 8, hidden size 7168, intermediate size 2048, `n-group=8`, and `topk-group=4`. It evaluates eight precision families, six routing distributions, and ten token counts with three warmups and the median of ten CUDA-graph replays per row.
> Precision 	8 	32 	64 	128 	256 	512 	1024 	2048 	4096 	8192 	Overall
> NVFP4 	0.9974x 	1.0031x 	1.0163x 	1.1177x 	1.1160x 	1.0008x 	0.9983x 	0.9995x 	0.9932x 	0.9952x 	1.0227x
> MXFP4 	1.0063x 	1.0033x 	1.0661x 	1.3152x 	1.3387x 	0.9879x 	0.9895x 	1.0238x 	0.9948x 	0.9956x 	1.0651x
> W4A16 	1.0004x 	1.0220x 	1.2578x 	1.7240x 	1.4206x 	1.5738x 	0.9887x 	0.9983x 	0.9887x 	0.9740x 	1.1677x
> BF16 	1.0031x 	1.0063x 	1.0928x 	1.3677x 	1.2830x 	1.1377x 	1.1969x 	1.0281x 	0.9917x 	0.9647x 	1.0999x
> FP8 per-tensor 	1.0046x 	1.0384x 	1.1651x 	1.1941x 	1.1557x 	1.1008x 	1.2138x 	0.9905x 	1.0100x 	1.0027x 	1.0844x
> FP8 block 	1.0034x 	1.0807x 	1.3598x 	1.3423x 	1.1516x 	0.9923x 	1.1187x 	1.1861x 	0.9615x 	0.9850x 	1.1101x
> MXFP8 	1.0007x 	1.0654x 	1.2058x 	1.2346x 	1.2281x 	0.9996x 	1.0015x 	0.9721x 	1.0038x 	0.9885x 	1.0653x
> MXINT4 	1.0024x 	1.4091x 	2.0148x 	2.0042x 	1.5466x 	1.1914x 	1.3027x 	0.9778x 	0.9800x 	0.9822x 	1.2913x
> 
> Fresh compiled plan outcomes were 330 `da_single_body`, 114 `da_fallback`, and 36 `da_switch` rows. Actual capture outcomes were 330 single-body, 114 fallback, 30 switch, and 6 pristine `noda_capture_fallback` rows. The carried per-row performance waiver remains explicit; the aggregate table is release evidence, not a claim that every row speeds up.

I updated the PR description. 👍

qiching (approved):
LGTM, approved

feih-nv:
Some review findings, ordered by priority:

1. **Profile arenas also affect DA-disabled autotuning**

   `core.py` assigns all dynamic MoE inputs to `profile_arena_input_indices` unconditionally. With two arenas and the default ten lanes, large token buckets can add several GB of autotuning memory even when DA is disabled.

   Scope the arena path to DA/value-aware profiling, or restore the previous L2-aware lane-count limit. Add a DA-off regression test confirming ordinary MoE autotuning uses the existing bounded clone path.

2. **Multiple distribution samples are not independent or reproducible**

   `RoutingRealizationKey.sample_index` is not used to seed generation. Expert IDs depend on global RNG state, while routing weights can repeat because `torch.rand()` runs in an unseeded `fork_rng()` context.

   Use a stable per-sample `torch.Generator` derived from the realization key and pass it to both `multinomial` and `rand`. Alternatively, restrict the current release to one sample and update the documentation.

3. **Baseline guard needs deterministic coverage and surfaced reasons**

   The guard is enabled by default and is the safety mechanism for rejecting noncompetitive DA plans, but tests generally disable it. `DAPlanCompiler` produces admission/rejection reasons that are not propagated into operation diagnostics or cache records.

   Add tests for switch and singleton admission/rejection, margin boundaries, control overhead, and NoDA fallback. Preserve `guard_reason` in runtime state and serialized plans.

4. **SM12 can publish a SWITCH plan that native code cannot execute**

   Native multi-body SWITCH routing is restricted to SM10, while Python eligibility can still tune and publish a switch plan on SM120/121. It then fails during resource preparation and becomes a sticky NoDA fallback.

   Gate multi-body DA by architecture during eligibility or plan compilation. SM12 can explicitly use singleton/NoDA until native SWITCH support is implemented.

5. **Eager execution differs from the documented contract**

   The PR states that eager execution continues using the shape-only winner. Current code can instead run `_eager_body`, selected from `ddist:1.1` or uniform profiling.

   Either restore shape-only eager behavior or document distribution-selected eager dispatch as an intentional DA-enabled behavior change, including its cache identity and small-token performance implications.

6. **Capture fallback diagnostics and lease failure handling are incomplete**

   A compiled `da_switch` can capture NoDA without a useful `capture_fallback_reason`. Also, `acquire_graph_lease()` clears pending dispatcher state before all graph/generation/lane validation succeeds, preventing retry after failure.

   Return structured fallback reasons and clear pending lease state only after successful lease registration.

7. **Transient preparation failures remain sticky**

   Preparation exceptions are cached by plan generation and binding signature. A temporary allocation or resource failure can therefore disable DA until retuning or process restart.

   Distinguish permanent compatibility failures from retryable resource failures and add a bounded retry path.

Additional follow-ups: harden namespaced cache writes with locking/CAS semantics, add native FFI validation for selector tensors/body indices, and update the PR text to replace stale `DynamicValueSpec` terminology with the implemented value-aware profiling APIs.

feih-nv (approved):
Approving with some comments that can be addressed in the following PR(s).

aleozlx:
for the recored, AOT tests all passed

once the new conflict is resolved, i ok to just merge it

PR #4108, opened 2026-07-23
feat: support native qk_rope_head_dim=0 sparse MLA decode in trtllm-gen

JustinTong0323:
## 📌 Description

Adds a native TRTLLM-GEN sparse MLA decode path for the shape with **no rotary tail**: `kv_lora_rank=512`, `qk_rope_head_dim=0`.

For this shape the query carries no RoPE component, and both KV TMA descriptors address a **single 512-wide cache pool**, so the kernel needs the per-request active lengths to bound the sparse gather. Today `trtllm_batch_decode_with_kv_cache_mla` only accepts `deepseek_mla_dimensions` and `smaller_mla_dimensions`, so a `qk_rope_head_dim=0` request is rejected as an unsupported MLA dimension.

This PR registers the new dimension set and threads an optional `sparse_mla_top_k_lens` tensor down to the launcher so the shape can be served natively.

**Changes**

- **`csrc/trtllm_fmha_kernel_launcher.cu`** — add an optional `sparse_mla_top_k_lens` argument to `trtllm_paged_attention_decode`. When the single-pool dynamic sparse MLA shape is detected (`sparse_mla_top_k_lens` present and MLA decode), pass the key cache as the sliding-window KV pool so the kernel reads the active per-token lengths. The launcher already rejects combining block-sparse attention with sparse MLA (`sparse_mla_top_k <= 0` check), so the two stay mutually exclusive.
- **`flashinfer/mla/_core.py`** — register `nope_mla_dimensions` (`kv_lora_rank=512`, `qk_rope_head_dim=0`); require `sparse_mla_top_k > 0` and a `sparse_mla_top_k_lens` tensor for this shape; thread the autotune profiling length through the decode tuning config so different `top_k` values key distinct autotune configs; expose `sparse_mla_top_k_lens` on the public `trtllm_batch_decode_with_kv_cache_mla`.
- **`flashinfer/decode.py`** — forward the new optional argument at the two existing kernel call sites.
- **`flashinfer/trace/templates/attention.py`** — declare the optional `sparse_mla_top_k_lens` input on the sparse MLA decode trace template so the trace schema matches the kernel signature.

The new argument is **optional and defaults to `None`**, so the `deepseek_mla_dimensions` / `smaller_mla_dimensions` decode paths are unchanged. `sparse_mla_top_k_lens` (one `int32` active length per query token) is supplied by the caller.

## 🔍 Related Issues

None.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

Static checks pass (`clang-format` / `ruff` / `mypy` via `pre-commit`). The new path has been exercised end-to-end in a downstream serving stack that computes `sparse_mla_top_k_lens` from the page table and drives this decode path. Happy to add a focused in-tree unit test for the `qk_rope_head_dim=0` dimension registration + argument threading — see Reviewer Notes.

## Reviewer Notes

- **Backward compatibility**: `sparse_mla_top_k_lens` is optional and defaults to `None`; all existing callers and the two established MLA dimension sets keep their current behavior.
- **Mutual exclusion**: block-sparse attention and sparse MLA are already mutually exclusive in the launcher (`sparse_mla_top_k <= 0` check), so the new single-pool path cannot be entered together with block-sparse.
- **Autotune keying**: the profiling length is threaded through the decode tuning config and into the cache key, so a dense request (`len(inputs)==4`) and a sparse request (`len(inputs)==5`) resolve to distinct autotune configs rather than mis-keying.
- I can add a unit test covering the dimension registration and the optional-argument threading if you'd like it in-tree — let me know the preferred test shape.

## Summary by CodeRabbit

* **New Features**
  * Added support for native no-RoPE MLA decoding.
  * Added optional per-query sparse attention lengths for supported MLA decode workloads.
  * Added validation for sparse attention length tensor type, shape, device, and contiguity.
  * Integrated sparse MLA inputs with direct decoding and autotuning paths.

* **Bug Fixes**
  * Improved handling of supported MLA head configurations during decode dispatch.


flashinfer/mla/_core.py
saltyminty, line 3220: Is it intended that we don't have a auto branch for nope?

saltyminty:
CI looks good

saltyminty (approved):
Left one comment but approved

flashinfer/mla/_core.py
JustinTong0323: Yes, on SM100/SM103 the call intentionally stays on the generic auto path. Since the no-RoPE shape requires sparse MLA, CuTeDSL is filtered out and trtllm-gen is the only remaining runner, so a separate branch would duplicate the existing routing.

PR #4119, opened 2026-07-23
perf(test): bulk-precompile XQA decode kernels to cut test wall time ~4x

bkryu:
## 📌 Description

`tests/attention/test_trtllm_gen_attention_decode_xqa.py` takes **40+ minutes** in CI where profiling shows the wall time is ~95% JIT compilation, not kernel execution. 

The file's parametrize matrix expands to cases that map to 408 unique XQA kernel compilations, all of which occur serially, one at a time. Each compilation takes around 5 seconds. 

Current PR adds a `tests/attention/conftest.py` that, right before the xqa test file runs, enumerates the exact kernels the collected cases will need and builds them **all in a single parallel `ninja` invocation** via `flashinfer.jit.build_jit_specs`, then stages the artifacts so the tests load them directly instead of recompiling.

**Measured on GB10 / SM121, cold cache (full file):**

| | Wall time |
|---|---|
| Before (serial first-touch compile) | ~40 min `4896 passed, 66912 skipped, 2 warnings in 2305.72s (0:38:25)` |
| After (parallel prebuild + load-only) | **628 s (10:28)** — `4896 passed, 66912 skipped ... in (0:10:28)` |

The parallel build compiles all 408 kernels in ~4.6 min (0.67 s/kernel amortized vs 5.9 s serial); the rest is unchanged test execution. No behavior change — same pass/skip counts.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **Tests**
  * Improved attention test startup time by bulk-precompiling supported FlashInfer XQA decode kernels during pytest collection.
  * Reduced repeated compilation overhead by compiling a deduplicated set of module specs in parallel.
  * Added optional progress/status reporting when terminal output is available.
  * Improved robustness by staging built artifacts ahead of runtime using safe atomic moves/hardlinks, reusing existing files when possible, and falling back to normal JIT behavior if bulk setup encounters any issues.

### DIFF
--- a/tests/attention/conftest.py
+++ b/tests/attention/conftest.py
@@ -55,38 +55,46 @@ def pytest_collection_modifyitems(session, config, items):
     if not xqa_items:
         return
 
+    # --collect-only enumerates tests without running them, so don't pre-build.
+    if config.getoption("--collect-only"):
+        return
+
     from flashinfer.jit.core import build_jit_specs
     from flashinfer.jit.xqa import gen_xqa_module
 
-    specs = {}
-    for it in xqa_items:
-        cs = getattr(it, "callspec", None)
-        if cs is None:
-            continue
-        kwargs = _executed_xqa_uri(cs.params)
-        if kwargs is None:
-            continue
-        try:
-            spec = gen_xqa_module(**kwargs)
-        except ValueError as e:
-            logger.debug("xqa-prebuild: skipping unsupported config %s: %s", kwargs, e)
-            continue
-        specs[spec.name] = spec
-
-    specs = list(specs.values())
-    if not specs:
-        return
-
     reporter = config.pluginmanager.getplugin("terminalreporter")
-    if reporter:
-        reporter.write_line(
-            f"[xqa-prebuild] compiling {len(specs)} XQA kernels in parallel..."
-        )
 
-    # Any failure here (e.g. compilation failure) must not abort collection
-    # for the whole suite. Fall back to the normal per-test first-touch JIT path,
-    # where a genuinely broken kernel surfaces as its own test failure instead.
+    # The whole prebuild is a best-effort optimization. Any failure here must not
+    # abort collection for the entire suite.
     try:
+        specs = {}
+        for it in xqa_items:
+            cs = getattr(it, "callspec", None)
+            if cs is None:
+                continue
+            kwargs = _executed_xqa_uri(cs.params)
+            if kwargs is None:
+                continue
+            try:
+                spec = gen_xqa_module(**kwargs)
+            except ValueError as e:
+                # Per-config unsupported (bad dtype/page_size/head_dim): skip just
+                # this config and keep prebuilding the rest.
+                logger.debug(
+                    "xqa-prebuild: skipping unsupported config %s: %s", kwargs, e
+                )
+                continue
+            specs[spec.name] = spec
+
+        specs = list(specs.values())
+        if not specs:
+            return
+
+        if reporter:
+            reporter.write_line(
+                f"[xqa-prebuild] compiling {len(specs)} XQA kernels in parallel..."
+            )
+
         # One ninja graph, built in parallel; skip_prebuilt reuses anything already AOT'd.
         build_jit_specs(specs, verbose=False)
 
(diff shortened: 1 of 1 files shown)


tests/attention/conftest.py
```diff
@@ -0,0 +1,104 @@
...
+        return
+
+    reporter = config.pluginmanager.getplugin("terminalreporter")
+    if reporter:
+        reporter.write_line(
+            f"[xqa-prebuild] compiling {len(specs)} XQA kernels in parallel..."
+        )
+
+    # One ninja graph, built in parallel; skip_prebuilt reuses anything already AOT'd.
+    build_jit_specs(specs, verbose=False)
```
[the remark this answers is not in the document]
bkryu: Good catch — agreed. A prebuild failure shouldn't be worse than the status quo. Made it a try-except block in next commit

tests/attention/conftest.py
```diff
@@ -0,0 +1,104 @@
...
+        if dst.exists() or not src.exists():
+            continue
+        dst.parent.mkdir(parents=True, exist_ok=True)
+        try:
+            os.link(src, dst)  # hardlink: instant, same filesystem
+        except OSError:
+            import shutil
+
+            shutil.copy2(src, dst)
+        staged += 1
```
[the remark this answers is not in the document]
bkryu: Good point and addresseed in next commit.  The os.link fast path is already atomic. Fallback wasn't.

tests/attention/conftest.py
```diff
@@ -0,0 +1,104 @@
...
+        cs = getattr(it, "callspec", None)
+        if cs is None:
+            continue
+        kwargs = _executed_xqa_uri(cs.params)
+        if kwargs is None:
+            continue
+        try:
+            spec = gen_xqa_module(**kwargs)
+        except Exception:
+            continue  # unsupported config the test would also skip
```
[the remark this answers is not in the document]
bkryu: Fixed in next commit. gen_xqa_module signals "unsupported config" specifically with ValueError (three of them, no other rejection type), so I narrowed the handler to except ValueError and log the skipped config at DEBUG.

tests/attention/conftest.py
```diff
@@ -0,0 +1,134 @@
...
+    for it in xqa_items:
+        cs = getattr(it, "callspec", None)
+        if cs is None:
+            continue
+        kwargs = _executed_xqa_uri(cs.params)
+        if kwargs is None:
+            continue
+        try:
+            spec = gen_xqa_module(**kwargs)
+        except ValueError as e:
```
saltyminty: Not too familiar with exact structure but per agent review – is it safe to only catch ValueError here? i.e unsupported gpu raises runtime error

tests/attention/conftest.py
```diff
@@ -0,0 +1,134 @@
...
+        head_dim=p["head_dim"],
+        head_group_ratio=p["head_grp_size"],
+        use_sliding_window=(p["window_left"] != -1),
+        output_dtype=_DT[o],
+        q_seq_len=p["q_len_per_req"],
+    )
+
+
+@pytest.hookimpl(trylast=True)
+def pytest_collection_modifyitems(session, config, items):
```
saltyminty, line 51: This runs unconditionally for any pytest collection – should we have a guard to exit when running under --collect-only?

bkryu: Agreed, added the guard. This matters concretely: the CI scheduler runs pytest --collect-only -q "$test_file" per file (scripts/test_utils.sh:284), so without the guard the full 408-kernel (~5 min) compile fired during that enumeration pass — semantically wrong (--collect-only shouldn't do run-preparation) and a real risk of tripping the collection step's timeout.

Fixed with an early guard for `--collect-only.`

tests/attention/conftest.py
bkryu: You're right. The value error by itself is not safe. 

Fixed by treating the two failure modes distinctly:
- Per-config unsupported → skip just that one config and prebuild the rest. This is expected control flow, logged at DEBUG.
- Whole-prebuild can't proceed → I moved the spec-generation loop inside the existing outer try/except that logs [xqa-prebuild] prebuild failed (…) and falls back to per-test first-touch JIT. Genuine bugs aren't hidden — they resurface per-test (the test's own build_and_load re-calls gen_xqa_module), just without taking down collection.

saltyminty (approved):

Issue #4070, opened 2026-07-20
[Feature] SM120/121 does not support Sink Attention


askliar:
As per vllm docs here: https://docs.vllm.ai/en/latest/design/attention_backends/#standard-attention-mha-mqa-gqa-backends, FlashInfer backend only supports sink attention when using trtllm-gen. 
On SM12x, trtllm-gen is not supported. However, attention sinks are required for efficiency gains, esp. in some of the newer models.

bkryu:
@yichengj0 is currently looking

bkryu:
Followup fix from PR #4199. Expected to be closed after it

bkryu:
#4199 is merged and should be available starting 0.6.16. Closing issue; @askliar , please feel free to reopen if you see any issues!

PR #4137, opened 2026-07-24
feat(xqa): ragged Q and per-row sliding-window masking for speculative decode

yichengj0:
## 📌 Description

XQA is the FlashInfer decode kernel used on SM120/121 for models with attention sinks (#4070). Serving those models with speculative decoding runs the draft-verification step through XQA as well, and two gaps blocked that:

- Every request in a batch had to verify the same number of draft tokens. The kernel indexes queries and masks through cumulative lengths internally, but the argument was never exposed.
- Sliding-window masking computed one window start from the last draft token's position and applied it to the whole draft block. Each draft token sits at its own position, so this masked out KV that the earlier draft tokens should still attend to.

Fixes:

- Plumb `q_cu_seq_lens` through the wrapper and Python API so each request can verify a different number of draft tokens, with host-side input validation. The SM90 fp8 path rejects ragged Q rather than run an unvalidated path.
- Compute the window per draft row: whole KV tiles are skipped conservatively, and the exact per-row edge is masked in the kernel.
- Extend tests and the benchmark to cover both draft-block mask modes (causal and full, selected with `--spec_dec_mask`), long contexts (split-KV path), zero-length drafts, and GQA group ratio 16.

Kernel changes and investigation by @bkryu.

## 📈 Performance

The table shows XQA's speedup over each baseline, measured as kernel time with CUPTI at batch size 1 with head_dim 128, 32 query heads over 2 KV heads, and a bf16 KV cache. Each cell gives the speedup at context lengths 1k, 4k, and 32k. The column m is the number of query tokens per request: m=1 is plain decode, and m=4 or 8 is draft-block verification in speculative decode. Each baseline is measured under full attention and under sliding-window attention with a 1024-token window (SWA 1024).

Baselines:

- vLLM's Triton unified-attention kernel, sinks enabled on both sides.
- FlashInfer fa2 wrappers, sinks disabled on both sides (fa2 has no sink support).

| GPU | m | vs Triton, full attn | vs Triton, SWA 1024 | vs fa2, full attn | vs fa2, SWA 1024 |
|---|---|---|---|---|---|
| GB10 | 1 | 1.7 / 0.9 / 1.1 | 1.5 / 1.2 / 2.3 | 0.8 / 0.9 / 1.0 | 0.9 / 0.7 / 0.8 |
| GB10 | 4 | 1.2 / 2.2 / 5.0 | 1.2 / 1.2 / 1.2 | 2.7 / 1.4 / 1.2 | 2.3 / 1.9 / 1.9 |
| GB10 | 8 | 1.1 / 1.8 / 4.0 | 1.2 / 1.1 / 0.9 | 1.8 / 1.4 / 1.0 | 2.5 / 2.1 / 1.5 |
| RTX PRO 6000 | 1 | 3.0 / 2.4 / 3.2 | 3.3 / 3.0 / 3.6 | 1.0 / 0.8 / 0.8 | 1.0 / 1.0 / 0.6 |
| RTX PRO 6000 | 4 | 2.1 / 6.0 / 18.0 | 1.9 / 1.9 / 1.2 | 3.7 / 2.3 / 1.5 | 2.7 / 3.0 / 2.1 |
| RTX PRO 6000 | 8 | 2.1 / 5.5 / 14.2 | 1.8 / 2.0 / 1.0 | 3.3 / 2.4 / 1.1 | 2.8 / 2.8 / 1.6 |

## 🔍 Related Issues

#4070 (attention sinks on SM120/121).

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

`tests/attention/test_xqa_batch_decode.py` passes on SM120 (RTX 5080, RTX PRO 6000) and SM121 (GB10). Outputs cross-checked against an independent reference kernel up to 64k context.

## Reviewer Notes

- Speculative-decode builds now assume the draft tokens form a linear chain rather than a tree (the `IS_SPEC_DEC_TREE` compile flag flips from 1 to 0). This is deliberate: the per-row window needs each draft token's sequence position, and only a linear chain defines one. The assumption only matters in sliding-window builds, and tree-shaped drafts never worked correctly with sliding windows, so no working caller changes behavior.
- Not validated on SM90/SM100 hardware. The changed paths are spec-dec only, and the SM90 fp8 wrapper rejects ragged explicitly; CI runs the tests on those arches.

## Summary by CodeRabbit

- **New Features**
  - Added ragged-Q speculative decoding support via optional `q_cu_seq_lens`.
  - Introduced configurable speculative draft attention masking with `--spec_dec_mask` (`causal`/`full`), including trace support.
  - Extended JIT/custom-op pathways to support ragged-Q specialization.

- **Bug Fixes**
  - Improved speculative decoding masking for sliding-window cases and fixed an empty-query edge case for ragged requests.

- **Tests**
  - Expanded coverage for ragged-Q, `causal`/`full` mask modes, sliding-window behavior, and KV cache variants.

### DIFF
--- a/benchmarks/routines/attention.py
+++ b/benchmarks/routines/attention.py
@@ -202,7 +202,14 @@ def parse_attention_args(line, parser):
         "--causal",
         action="store_true",
         default=False,
-        help="Causal masking. Note: not padding masking. Only used for prefill tests.",
+        help=(
+            "Causal masking. Note: not padding masking. Used for prefill tests, and "
+            "for the speculative-decode draft block (--s_qo > 1) in decode tests: "
+            "with --causal, draft token i attends to draft tokens j <= i (standard "
+            "speculative decoding); without it, every draft token attends to all "
+            "draft tokens (no causal masking within the draft block, e.g. "
+            "DFlash-style drafters). The KV prefix is always fully visible."
+        ),
     )
     parser.add_argument(
         "--random_actual_seq_len",
@@ -284,21 +291,31 @@ def sample_actual_seq_lens(max_seqlen, batch_size, device, random_actual_seq_len
     return actual_seq_lens
 
 
-def generate_speculative_causal_mask(batch_size, q_seq_len, device):
+def generate_speculative_mask(batch_size, q_seq_len, device, mask_mode="causal"):
     """
-    Generate a packed causal mask for speculative decode chunks (q_len > 1).
+    Generate a packed draft-block mask for speculative decode chunks (q_len > 1).
 
     Returned shape is [batch_size, q_seq_len, num_packed_masks_per_token * 2]
     with dtype uint16, where num_packed_masks_per_token = ceil(q_seq_len / 32).
-    Each query row i encodes allowed attention to draft-token columns j <= i
-    (strictly lower-triangular with diagonal) and masks out j > i.
+    With mask_mode="causal", each query row i encodes allowed attention to
+    draft-token columns j <= i (strictly lower-triangular with diagonal) and
+    masks out j > i. With mask_mode="full", every draft token attends to all
+    draft tokens (no causal masking within the draft block, e.g. DFlash-style
+    drafters); the KV prefix is always fully visible in both modes.
     The innermost dimension stores packed bits (uint32 words reinterpreted as
     uint16), matching the mask layout expected by decode APIs.
     """
     num_packed_masks_per_token = (q_seq_len + 31) // 32
     q_indices = torch.arange(q_seq_len, device=device, dtype=torch.int32).unsqueeze(1)
     kv_indices = torch.arange(q_seq_len, device=device, dtype=torch.int32).unsqueeze(0)
-    causal_bool_mask = kv_indices <= q_indices
+    if mask_mode == "causal":
+        causal_bool_mask = kv_indices <= q_indices
+    elif mask_mode == "full":
+        causal_bool_mask = torch.ones(
+            q_seq_len, q_seq_len, device=device, dtype=torch.bool
+        )
+    else:
+        raise ValueError(f"Unsupported spec-decode mask mode: {mask_mode}")
 
     padded_seq_len = num_packed_masks_per_token * 32
     if padded_seq_len > q_seq_len:
@@ -380,6 +397,8 @@ def testBatchDecodeWithPagedKVCacheWrapper(args):
     batch_size = args.batch_size
     s_qo = args.s_qo
     speculative_decode = s_qo > 1
+    # For speculative decode, --causal selects the draft-block mask.
+    causal = args.causal
     s_kv = args.s_kv
     num_qo_heads = args.num_qo_heads
     num_kv_heads = args.num_kv_heads
@@ -437,6 +456,13 @@ def testBatchDecodeWithPagedKVCacheWrapper(args):
     if "auto" in backends and speculative_decode:
         print("[INFO] auto backend is disabled for speculative decode. Skipping.")
         backends.remove("auto")
+
+    if speculative_decode and not causal and "trtllm-gen" in backends:
+        print(
+            "[WARNING] trtllm-gen wrapper backend applies implicit causal masking to "
+            "the draft block and ignores the non-causal (DFlash) mask; refcheck "
+            "against trtllm-native may mismatch."
+        )
 
     # Storage for timing results and outputs
     backend_times = {backend: [] for backend in backends}
@@ -568,7 +594,9 @@ def testBatchDecodeWithPagedKVCacheWrapper(args):
         * (s_qo * num_qo_heads * head_dim_qk)
     ).long()  # For cuDNN
     speculative_mask = (
-        generate_speculative_causal_mask(batch_size, s_qo, device)
+        generate_speculative_mask(
+            batch_size, s_qo, device, "causal" if causal else "full"
+        )
         if speculative_decode
         else None
     )
@@ -852,7 +880,7 @@ def testBatchDecodeWithPagedKVCacheWrapper(args):
                 cur_res["num_kv_heads"] = num_kv_heads
                 cur_res["head_dim_qk"] = head_dim_qk
                 cur_res["head_dim_vo"] = head_dim_vo
-                cur_res["causal"] = False
+                cur_res["causal"] = causal if speculative_decode else False
                 cur_res["q_dtype"] = q_dtype
                 cur_res["kv_dtype"] = kv_dtype
                 cur_res["avg_actual_seq_len"] = avg_seq_len_kv
--- a/benchmarks/routines/attention.py
+++ b/benchmarks/routines/attention.py
@@ -202,13 +202,19 @@ def parse_attention_args(line, parser):
         "--causal",
         action="store_true",
         default=False,
+        help="Causal masking. Note: not padding masking. Only used for prefill tests.",
+    )
+    parser.add_argument(
+        "--spec_dec_mask",
+        type=str,
+        choices=["causal", "full"],
+        default="causal",
         help=(
-            "Causal masking. Note: not padding masking. Used for prefill tests, and "
-            "for the speculative-decode draft block (--s_qo > 1) in decode tests: "
-            "with --causal, draft token i attends to draft tokens j <= i (standard "
-            "speculative decoding); without it, every draft token attends to all "
-            "draft tokens (no causal masking within the draft block, e.g. "
-            "DFlash-style drafters). The KV prefix is always fully visible."
+            "Draft-block mask for speculative decode (--s_qo > 1) in decode tests. "
+            "'causal': draft token i attends to draft tokens j <= i (standard "
+            "speculative decoding, the default). 'full': every draft token attends "
+            "to all draft tokens (e.g. DFlash-style drafters). The KV prefix is "
+            "always fully visible."
         ),
     )
     parser.add_argument(
@@ -292,18 +298,12 @@ def sample_actual_seq_lens(max_seqlen, batch_size, device, random_actual_seq_len
 
 
 def generate_speculative_mask(batch_size, q_seq_len, device, mask_mode="causal"):
-    """
-    Generate a packed draft-block mask for speculative decode chunks (q_len > 1).
+    """Packed draft-block mask for speculative decode (q_len > 1).
 
-    Returned shape is [batch_size, q_seq_len, num_packed_masks_per_token * 2]
-    with dtype uint16, where num_packed_masks_per_token = ceil(q_seq_len / 32).
-    With mask_mode="causal", each query row i encodes allowed attention to
-    draft-token columns j <= i (strictly lower-triangular with diagonal) and
-    masks out j > i. With mask_mode="full", every draft token attends to all
-    draft tokens (no causal masking within the draft block, e.g. DFlash-style
-    drafters); the KV prefix is always fully visible in both modes.
-    The innermost dimension stores packed bits (uint32 words reinterpreted as
-    uint16), matching the mask layout expected by decode APIs.
+    mask_mode "causal": draft token i attends to draft tokens j <= i;
+    "full": every draft token attends to all draft tokens. The KV prefix is
+    always fully visible. Returns [batch_size, q_seq_len,
+    ceil(q_seq_len / 32) * 2] uint16 (bit-packed, as decode APIs expect).
     """
     num_packed_masks_per_token = (q_seq_len + 31) // 32
     q_indices = torch.arange(q_seq_len, device=device, dtype=torch.int32).unsqueeze(1)
@@ -397,8 +397,7 @@ def testBatchDecodeWithPagedKVCacheWrapper(args):
     batch_size = args.batch_size
     s_qo = args.s_qo
     speculative_decode = s_qo > 1
-    # For speculative decode, --causal selects the draft-block mask.
-    causal = args.causal
+    spec_dec_mask_mode = args.spec_dec_mask
     s_kv = args.s_kv
     num_qo_heads = args.num_qo_heads
     num_kv_heads = args.num_kv_heads
@@ -457,7 +456,7 @@ def testBatchDecodeWithPagedKVCacheWrapper(args):
         print("[INFO] auto backend is disabled for speculative decode. Skipping.")
         backends.remove("auto")
 
-    if speculative_decode and not causal and "trtllm-gen" in backends:
+    if speculative_decode and spec_dec_mask_mode == "full" and "trtllm-gen" in backends:
         print(
             "[WARNING] trtllm-gen wrapper backend applies implicit causal masking to "
             "the draft block and ignores the non-causal (DFlash) mask; refcheck "
@@ -594,9 +593,7 @@ def testBatchDecodeWithPagedKVCacheWrapper(args):
         * (s_qo * num_qo_heads * head_dim_qk)
     ).long()  # For cuDNN
     speculative_mask = (
-        generate_speculative_mask(
-            batch_size, s_qo, device, "causal" if causal else "full"
-        )
+        generate_speculative_mask(batch_size, s_qo, device, spec_dec_mask_mode)
         if speculative_decode
         else None
     )
@@ -880,7 +877,9 @@ def testBatchDecodeWithPagedKVCacheWrapper(args):
                 cur_res["num_kv_heads"] = num_kv_heads
                 cur_res["head_dim_qk"] = head_dim_qk
                 cur_res["head_dim_vo"] = head_dim_vo
-                cur_res["causal"] = causal if speculative_decode else False
+                cur_res["causal"] = (
+                    spec_dec_mask_mode == "causal" if speculative_decode else False
+                )
                 cur_res["q_dtype"] = q_dtype
                 cur_res["kv_dtype"] = kv_dtype
                 cur_res["avg_actual_seq_len"] = avg_seq_len_kv
--- a/flashinfer/xqa.py
+++ b/flashinfer/xqa.py
@@ -42,6 +42,7 @@ def get_xqa_module(
     use_sliding_window: bool,
     output_dtype: torch.dtype,
     q_seq_len: int,
+    use_ragged_q: bool = False,
 ):
     module = gen_xqa_module(
         input_dtype,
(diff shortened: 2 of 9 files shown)


benchmarks/routines/attention.py
```diff
@@ -204,6 +204,19 @@ def parse_attention_args(line, parser):
...
+        choices=["causal", "full"],
+        default="causal",
+        help=(
+            "Draft-block mask for speculative decode (--s_qo > 1) in decode tests. "
+            "'causal': draft token i attends to draft tokens j <= i (standard "
+            "speculative decoding, the default). 'full': every draft token attends "
+            "to all draft tokens (e.g. DFlash-style drafters). The KV prefix is "
+            "always fully visible."
+        ),
+    )
```
bkryu, line 219: Can we just make the input `--causal`?
* If specdec:
  * If `--causal` provided --> causal
  * If not provided --> Full
* else:
  * `--causal` does nothing

This should be more consistent with the rest of the benchmarking code base

bkryu (commented):
Thanks @yichengj0, left one comment about benchmark flags

flashinfer/xqa.py
```diff
@@ -243,16 +249,28 @@ def xqa(
...
+        where each request's rows cover its own draft length.
         Data type should be torch.uint16 (bit-packed format, aligned to 32 bits).
+    q_cu_seq_lens : Optional[torch.Tensor], default=None
+        Cumulative draft lengths ``[batch_size + 1]`` (torch.int32 or
+        torch.uint32, on device) enabling ragged Q: requests may have
+        different draft lengths, including 0 for a request with no draft
+        tokens this step. When set, ``q`` and ``output`` are packed as
+        ``[total_q_tokens, num_q_heads, head_dim]``, ``q_seq_len`` must be
+        the maximum draft length, and ``mask`` rows are packed by the same
+        cumulative offsets.
```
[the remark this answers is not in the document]
yichengj0: Good catch, will note the SM90 fp8 limitation in the q_cu_seq_lens docstring.

flashinfer/xqa.py
```diff
@@ -266,8 +284,31 @@ def xqa(
...
+            "pointer on the GPU"
+        )
+        assert q_cu_seq_lens.numel() == seq_lens.shape[0] + 1, (
+            f"q_cu_seq_lens must have batch_size + 1 = {seq_lens.shape[0] + 1} "
+            f"entries, got {q_cu_seq_lens.numel()}"
+        )
+        assert output.shape == q.shape, "Output must match packed ragged q shape"
+        # Not checked (needs a device sync): entries non-decreasing, each
+        # length <= q_seq_len, last entry == q.shape[0]. Violations return
+        # garbage rows rather than raising.
```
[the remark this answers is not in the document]
yichengj0: Partially agree. The cheap metadata checks (same device as q, 1-D, contiguous) are worth adding and we will. The value checks need a .cpu() copy, which forces the device sync this path exists to avoid and breaks CUDA graph capture, so we're leaving those out. The docstring and an inline comment already state that violated offset invariants produce garbage output rather than an error.

benchmarks/routines/attention.py
yichengj0: The catch is `--causal` defaults to false, so bare spec-dec runs would silently switch to the full mask (today's bench always uses the causal mask for spec-dec, and trtllm-gen refcheck would mismatch by default). We want to keep causal as the default, so we chose a separate flag defaulting to causal. Happy to fold it into `--causal` if you're fine with full becoming the default, wdyt?

bkryu: Ah I see. In that case let's keep the change as is for backwards compatibility. Thanks for explaining 👍

qsang-nv:
https://github.com/flashinfer-ai/flashinfer/actions/runs/30125334423/job/89587528368?pr=4137
tests/attention/test_xqa_batch_decode.py failed on H100, please take a look.

saltyminty:
+1 to @qsang-nv's comment on the failure above – I believe it's just a unsupported case that needs to be skipped (fp8 kv + enable_sink=false)

saltyminty (approved):
Approved pending CI and fixing of the previously mentioned test failure.

yichengj0:
@qsang-nv @saltyminty Thanks for flagging. We reproduced this on an H100 and root-caused it. All 36 failures are on the SM90 fp8 path, which dispatches to the Hopper-specific `mha_sm90.cu`:

- The 32 `ragged_q` failures are this PR's own guard raising: `mha_sm90.cu` has no ragged-Q support.
- The 4 `spec_dec_sliding_window` failures are real. The `IS_SPEC_DEC_TREE=0` build enabled a dormant per-row sliding-window path in `mha_sm90.cu`, and it mis-masks full (non-causal) draft masks. The `enable_sink=True` variants pass because the existing sinks workaround already routes them to the generic `mha.cu`.

Instead of skipping, the pushed fix extends that fallback: SM90 fp8 spec-dec with a sliding window or ragged Q also runs on the generic kernel, verified on an H100.

The `mha_sm90.cu` gaps could be tracked in a follow-up issue so the Hopper fast path can be restored later.

qsang-nv:
Thanks for the quick fix on the H100 failures — routing SM90 fp8 back to the generic kernel rather than skipping the cases is the right call. Three things I'd like resolved before this lands, plus two non-blocking notes.

---

## Blocking

### 1. `qCuSeqLens` is dereferenced before `griddepcontrol.wait` (PDL race)

Worth flagging up front that this one is on the **SM12x path, not SM90**: `mha.cu` is the only kernel compiled for SM120/121 (`mha_sm90.cu` is added to the sources and `USE_SM90_MHA=1` only when `has_sm90`), `ENABLE_PDL` is 1 there (`__CUDA_ARCH__ >= 900`, and SM120 is 1200), and `device_support_pdl()` returns True for major >= 9. So this affects the PR's primary target architecture unconditionally.

[`mha.cu#L1599-L1610`](https://github.com/flashinfer-ai/flashinfer/blob/a51867d85d19fcbd3b3388df04d984275d1f45f9/csrc/xqa/mha.cu#L1599-L1610) reads `qCuSeqLens[idxReq]` and `qCuSeqLens[idxReq + 1]` before the acquire; `acqBulk()` (`griddepcontrol.wait`) is at [`L1685-L1688`](https://github.com/flashinfer-ai/flashinfer/blob/a51867d85d19fcbd3b3388df04d984275d1f45f9/csrc/xqa/mha.cu#L1685-L1688). PDL is on by default — `makeLaunchConfig` sets `cudaLaunchAttributeProgrammaticStreamSerialization` and the Python `enable_pdl` defaults to `device_support_pdl()`. Per the PTX contract, a dependent grid may begin executing before the prerequisite grid's writes are visible; `griddepcontrol.wait` is what establishes that visibility. Reading producer-written memory before it is a contract violation.

These are the first *request-indexing* loads placed before the acquire, and they are the dangerous ones: they gate an early return and they produce `reqSeqOffset`, which drives the mask row offset (L1626), the Q row offset (L1708) and **the output row offset** (L2780, `&output[reqSeqOffset * nbQHeads]`). A stale or garbage value doesn't merely perturb numerics — it can write into another request's output rows, or out of bounds if the buffer still holds allocator garbage, non-deterministically. Because the loaded value feeds a control-flow decision (the zero-length early return), this also isn't something the compiler could be sinking below the acquire.

`q_cu_seq_lens` is exactly the kind of buffer a preceding kernel produces on device — it's the prefix sum of accepted draft lengths. The API accepts it as a device tensor and performs no host-side synchronization, so it can naturally be produced by a preceding device kernel. If that producer is PDL-aware (issues `launch_dependents` before its writes retire), XQA can observe the pre-completion state.

No existing test intentionally supplies `q_cu_seq_lens` from a prerequisite kernel that triggers `launch_dependents` before completion. `test_xqa_batch_decode_ragged_q` constructs it through ordinary `torch.cumsum`/`torch.cat`, so the test does not establish the early-launch overlap needed to expose this race. A serving integration with a PDL-aware producer can.

Suggested fix — move the loads **and everything derived from them** below the acquire; the CTA-local shared-memory/barrier initialization can stay where it is:

```cpp
  __syncthreads();

#if ENABLE_PDL
  preExit();
  acqBulk();
#endif

#if SPEC_DEC
  bool const variableQSeqLen = qCuSeqLens != nullptr;
  uint32_t const actualQSeqLen =
      variableQSeqLen ? uint32_t(qCuSeqLens[idxReq + 1] - qCuSeqLens[idxReq]) : qSeqLen;
  uint32_t const reqSeqOffset = variableQSeqLen ? uint32_t(qCuSeqLens[idxReq]) : (qSeqLen * idxReq);
  if (variableQSeqLen && actualQSeqLen == 0) {
    return;
  }
  // ... nbQHeadTokens / totalNbHeadTokensInGrp / nbValidHeadTokens / `mask +=` follow here
#endif
```

Returning after `__syncthreads()` is safe: the condition is CTA-uniform because all threads in a CTA use the same `blockIdx.z` and indptr entries, the barriers are CTA-local, and the request's CTAs have not entered any cross-CTA semaphore protocol yet (that only happens in the multi-block reduction at the end).

Side note, not a request for this PR: `qScalePtr[0]` / `kvScalePtr[0]` at L1582-L1583 are also global loads ahead of the acquire. They predate this PR and are far lower risk (a stale scalar scale gives wrong numbers, not misaddressed writes), but they're worth a look when someone next audits this kernel for PDL.

### 2. The fallback is still under-scoped: non-causal masks on SM90 fp8 without a sliding window

The new guard at [`xqa.py#L386-L393`](https://github.com/flashinfer-ai/flashinfer/blob/a51867d85d19fcbd3b3388df04d984275d1f45f9/flashinfer/xqa.py#L386-L393) keys on `use_ragged_q or use_sliding_window`. But the breakage is tied to SWAP_AB, not to the window.

The SWAP_AB mask implementation at [`mha_sm90.cu#L1755-L1801`](https://github.com/flashinfer-ai/flashinfer/blob/a51867d85d19fcbd3b3388df04d984275d1f45f9/csrc/xqa/mha_sm90.cu#L1755-L1801) does not use `specDec` at all. It synthesizes the draft mask from the column index:

```cpp
uint32_t const maskCol = col / headGrpSize;
MaskType const bit_mask = (1ULL << (maskCol + 1)) - 1;   // hard-coded causal
```

The only consumers of the runtime mask (`specDec.needMask` / `specDec.loadTileMaskRow`) are at [`L2122`/`L2140-L2141`](https://github.com/flashinfer-ai/flashinfer/blob/a51867d85d19fcbd3b3388df04d984275d1f45f9/csrc/xqa/mha_sm90.cu#L2140-L2141), i.e. exclusively in the non-SWAP_AB variant. `tok0WinBeg` is likewise accepted but never used in the SWAP_AB version. And SWAP_AB is selected purely by `SPEC_Q_SEQ_LEN`, which [`jit/xqa.py#L91`](https://github.com/flashinfer-ai/flashinfer/blob/a51867d85d19fcbd3b3388df04d984275d1f45f9/flashinfer/jit/xqa.py#L91) emits whenever `q_seq_len * head_group_ratio <= 32`. That matches the failure pattern exactly: the 4 failing cases were `(4,2,32,2,4)` and `(4,4,64,4,2)` (product 8), while `(4,5,16,2,8)` (40) and `(4,4,32,2,16)` (64) passed.

So a `full` mask with `sliding_win_size == 0` and `q_seq_len * head_group_ratio <= 32` still reaches this path and returns causal results.

**The main test already covers that combination and passes, and that is itself the problem.** Every link is checkable: `cc == 9` + fp8 KV sets `run_sm90_fp8_mha = True`; `sinks is None`; not ragged; `window_left = -1` means no fallback (and in the previous revision there was no fallback at all); `USE_SM90_MHA == 1` on that runner is proven by the ragged-Q `ICHECK` firing there in the previous revision's run; `q_seq_len * head_group_ratio <= 32` gives SWAP_AB. So `test_xqa_batch_decode[full][bf16-fp8-bf16][window_left=-1][enable_sink=False]` ran a causal-masked kernel against a non-causal reference and still passed. That is a proof that the currently exercised inputs and fp8 tolerance did not distinguish causal from full, not evidence that the path is correct.

I want to be clear about what I think belongs here versus in a follow-up, because the SWAP_AB behavior itself predates this PR. What changed is the contract: before this PR the mask was documented as causal-only, so hard-coding causal was consistent with it. This PR promotes non-causal draft masks to an advertised feature —

```diff
-        Causal attention mask for speculative decoding mode (when ``q_seq_len > 1``).
+        Draft-block attention mask for speculative decoding mode (when
```
```diff
-        causal attention mask for xqa speculative decoding.
+        draft-block attention mask for xqa speculative decoding.
```

plus the new `--spec_dec_mask {causal,full}` benchmark flag and the `spec_dec_mask_mode` test parametrization. So the gap isn't inherited — the PR ships a documented mode that silently returns something else on one architecture.

**In this PR:**

1. Extend the dispatch guard so the advertised behavior is correct everywhere. "Fall back only when the mask isn't causal" isn't implementable — `xqa()` can't inspect a device-side mask without a sync. Either fall back whenever SM90 + fp8 + spec-dec + `q_seq_len * head_group_ratio <= 32`, or keep the Hopper kernel and suppress `-DSPEC_Q_SEQ_LEN` for masked spec-dec so `mha_sm90.cu` takes the mask-honoring non-SWAP_AB path — the same suppression mechanism this PR already added for ragged Q. Either way it's a couple of lines in the guard you just added, on the fallback path you already validated on H100, so it needs no SM90 hardware to land.
2. Add a deterministic mask-mode case. This is not an SM90 concern: the current parametrization ran a causal-masked kernel against a non-causal reference and passed, so it has no power to catch a mask-ignoring regression on *any* architecture — including SM120/121, where `full` is therefore exercised but not validated. E.g. use zero/equal Q and K so the logits are uniform, then give the future draft tokens clearly distinct V values.

**Follow-up:** making the SWAP_AB `warpGrpApplyMask` read the runtime mask (and actually use `tok0WinBeg`) is real kernel work that needs SM90 hardware to validate. Happy to see that tracked in the `mha_sm90.cu` issue you mentioned rather than done here.

### 3. The trace template's `q_cu_seq_lens` input is inert

[`attention.py#L2768`](https://github.com/flashinfer-ai/flashinfer/blob/a51867d85d19fcbd3b3388df04d984275d1f45f9/flashinfer/trace/templates/attention.py#L2768) declares `q_cu_seq_lens`, but `_xqa_batch_decode_reference` forwards `**kwargs` into `_trtllm_paged_attention_reference`, which only reads `cum_seq_lens_q` ([L1565](https://github.com/flashinfer-ai/flashinfer/blob/a51867d85d19fcbd3b3388df04d984275d1f45f9/flashinfer/trace/templates/attention.py#L1565)). The offsets are silently dropped and the reference falls back to `q_start = b * (num_tokens // batch_size)`.

Renaming the kwarg wouldn't be enough. That shared reference also hardcodes `causal=False` and has no notion of the packed bit-mask, and the template declares neither `mask` nor `q_len_per_req` — so a trace that sets `q_cu_seq_lens` trips `assert q_len_per_req > 1` before it gets anywhere. Either implement ragged query partitioning **and** packed-mask semantics in the trace reference (including `q_len_per_req` and `mask`), or drop `q_cu_seq_lens` from the trace schema in this PR and add it with the rest.

---

## Non-blocking

### 4. JIT module keying for ragged Q

Two small things in the same area.

**Normalization lives in the caller.** [`xqa.py#L373`](https://github.com/flashinfer-ai/flashinfer/blob/a51867d85d19fcbd3b3388df04d984275d1f45f9/flashinfer/xqa.py#L373) computes `ragged_build = use_ragged_q and q_seq_len * head_group_ratio <= 32` at the call site, while [`get_xqa_module`](https://github.com/flashinfer-ai/flashinfer/blob/a51867d85d19fcbd3b3388df04d984275d1f45f9/flashinfer/xqa.py#L35-L36) is `@functools.cache`d on the raw argument. If a process ever requests both raw `False` and raw `True` for a config with `q_seq_len * head_group_ratio > 32`, the two cache entries produce the same `op_name` and the second registration fails. Today's only in-tree caller normalizes, so this is latent. Note that normalizing inside the cached function body doesn't help — the cache key is computed from the raw arguments before the body runs. It needs an uncached wrapper:

```python
def get_xqa_module(..., q_seq_len, use_ragged_q=False):
    normalized = use_ragged_q and q_seq_len * head_group_ratio <= 32
    return _get_xqa_module_cached(..., q_seq_len, normalized)

@functools.cache
def _get_xqa_module_cached(..., q_seq_len, use_ragged_q):
    ...
```

**On SM12x the ragged variant is a redundant build.** `SPEC_Q_SEQ_LEN` appears zero times in `mha.cu`; outside the `static_assert` in `defines.h` it is used only by `mha_sm90.cu` to select SWAP_AB. So on an SM120/121-only build, suppressing it for ragged Q changes no generated code — it just adds the `_ragged_q` URI suffix and produces a second, binary-equivalent module: another full XQA JIT compile, another registered torch op, another resident module for the same shape. `gen_xqa_module` already computes `has_sm90` for the sources list; gating both the flag suppression and the URI suffix on it would keep non-SM90 builds down to one module.

### 5. `decode.py` docs don't reflect the SM90 fp8 fallback

The `q_cu_seq_lens` docstring at [`decode.py#L3606`](https://github.com/flashinfer-ai/flashinfer/blob/a51867d85d19fcbd3b3388df04d984275d1f45f9/flashinfer/decode.py#L3606) does not disclose that ragged Q uses the generic kernel on SM90 fp8. This matters more than in the previous revision: it used to raise there, and now it silently drops to the generic kernel, so users lose the Hopper fp8 path with no signal. The same applies to sliding-window dispatch: the fallback keys on `sliding_win_size > 0` rather than on whether the window actually truncates, so a windowed model on SM90 fp8 gives up the fast path even when `kv_len <= window`. Worth stating in both places.

Also +1 on tracking the `mha_sm90.cu` gaps in a follow-up issue so the Hopper path can be restored later.

yichengj0:
@qsang-nv Thanks again for the thorough review. All five items are addressed in #4199, and the SM90 kernel work (SWAP_AB mask support, `tok0WinBeg`, ragged Q) is tracked in #4198.

For the SWAP_AB gap we took the fallback option. The generic path is already validated on H100. Suppressing the specialization instead would move the Hopper kernel to its unswapped layout, which would need its own correctness and perf validation at these small shapes, and that effort is better spent restoring SWAP_AB properly in #4198.

Your scale-tensor side note turned out to be the same one-line move as the `qCuSeqLens` fix, so #4199 fixes it too.

PR #3502, opened 2026-06-04
feat(gdn): BF16 state recovery/decode kernel with per-request K and f…

ameynaik-hub:
Files changed:
  flashinfer/gdn_kernels/gdn_decode_bf16_state.py
  benchmarks/bench_gdn_decode.py
  tests/gdn/test_decode_delta_rule.py

# Three kernel modes

All three modes share the wide_vec kernel family (gdn_wide_vec_kernel for T>=2; gdn_wide_vec_kernel_t1 for T=1; small-batch ILP=4 path for B*HV < 128). The mode is selected at the Python wrapper layer via the disable_output, disable_state_update, and recovery_steps / accepted_steps flags. JIT cache keys differ across modes so each is compiled as its own specialization.

1. Recovery kernel  (disable_output=True, disable_state_update=False) Advance state through T tokens without per-token output. Used in the spec-decode reject branch where the verified prefix needs its state folded in but the speculated decode outputs are discarded. Supports per-request K via accepted_steps[i] (early-exit per CTA at accepted_steps[i]+1 iters).

2. Output-only / stateless decode  (disable_output=False, disable_state_update=True) Emit T per-token outputs, discard final state. Used in the spec-decode speculate branch where the kernel produces draft outputs but the candidate state is thrown away.

3. Fused recovery+decode  (disable_output=False, disable_state_update=False) Single launch collapsing the two-call chain state-only(K) -> update(T-K). First K iterations run state-only (Phase A), at the boundary h_K is written async to the state pool, remaining T-K iterations run full update with output (Phase B). h_T is discarded (spec-decode reject semantics: the "good" state is h_K). Phase split is either scalar recovery_steps: int OR per-request accepted_steps: Tensor[B] int32. When accepted_steps is set in fused-mode flags, my_K = accepted_steps[i] + 1 controls the per-CTA Phase A length and the scalar recovery_steps is ignored.

# Python API

  accepted_steps: Optional[Tensor[B] int32] = None
      Per-request K. In recovery mode: each CTA exits at accepted_steps[i]+1.
      In fused mode: each CTA does accepted_steps[i]+1 state-only iters,
      writes h_K to the state pool, then T-(accepted_steps[i]+1)
      output-emitting iters. None = uniform K=T.

  recovery_steps: int = 0
      Legacy scalar Phase A length for fused mode. Same value for every
      request. Ignored when accepted_steps is provided in fused-mode flags.

# Tests added (in tests/gdn/test_decode_delta_rule.py)

[...]

PR #3293, opened 2026-05-11
fix(gdn): allow importing gdn_decode without a CUDA device

kahyunnam:
## 📌 Description

- Defer `torch.cuda.get_device_properties()` and `torch.cuda.get_device_capability()` in `gdn_decode_bf16_state.py `from module scope to lazy `@functools.cache` helpers that derive the device from the input tensor (q.device), matching the convention used by the CuTe DSL MLA and MoE backends.
- Catch RuntimeError in addition to ImportError in all gdn_decode.py backend import blocks, matching the existing pattern in `gdn_kernels/__init__.py.`

Before this PR: 
```
CUDA_VISIBLE_DEVICES= python3 -c "import flashinfer.gdn_decode; print('Success: import worked without CUDA devices')"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/workspace/flashinfer/flashinfer/gdn_decode.py", line 58, in <module>
    from .gdn_kernels.gdn_decode_bf16_state import (
  File "/workspace/flashinfer/flashinfer/gdn_kernels/gdn_decode_bf16_state.py", line 1389, in <module>
    NUM_SMS = torch.cuda.get_device_properties(0).multi_processor_count
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/envs/py312/lib/python3.12/site-packages/torch/cuda/__init__.py", line 614, in get_device_properties
    _lazy_init()  # will define _get_device_properties
    ^^^^^^^^^^^^
  File "/opt/conda/envs/py312/lib/python3.12/site-packages/torch/cuda/__init__.py", line 410, in _lazy_init
    torch._C._cuda_init()
RuntimeError: No CUDA GPUs are available
```

After this PR: 
```
CUDA_VISIBLE_DEVICES= python3 -c "import flashinfer.gdn_decode; print('Success: import worked without CUDA devices')"
Success: import worked without CUDA devices
```

## 🔍 Related Issues

https://github.com/flashinfer-ai/flashinfer/issues/3262

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

[...]

Issue #3262, opened 2026-05-07
`import flashinfer.gdn_decode` requires CUDA devices available


feldsherov:
[This commit](https://github.com/flashinfer-ai/flashinfer/blame/c5c089bea20e50d5010e3185f3f14f485aaefc37/flashinfer/gdn_kernels/gdn_decode_bf16_state.py#L1389) makes impossible to `import flashinfer.gdn_decode` in CPU only environment.

This is quite inconvenient, when you have CPU only tests in code base with FI.

Reproduction:
```
# python3 example.py
Hi :)
# CUDA_VISIBLE_DEVICES= python3 example.py
Traceback (most recent call last):
  File "/traindata/svyatoslav/example.py", line 1, in <module>
    import flashinfer.gdn_decode
  File "/usr/local/lib/python3.12/dist-packages/flashinfer/gdn_decode.py", line 58, in <module>
    from .gdn_kernels.gdn_decode_bf16_state import (
  File "/usr/local/lib/python3.12/dist-packages/flashinfer/gdn_kernels/gdn_decode_bf16_state.py", line 1389, in <module>
    NUM_SMS = torch.cuda.get_device_properties(0).multi_processor_count
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch/cuda/__init__.py", line 686, in get_device_properties
    _lazy_init()  # will define _get_device_properties
    ^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch/cuda/__init__.py", line 478, in _lazy_init
    torch._C._cuda_init()
RuntimeError: No CUDA GPUs are available
# cat example.py
import flashinfer.gdn_decode

print("Hi :)")

# flashinfer show-config
=== Version Info ===
FlashInfer version: 0.6.11
flashinfer-cubin version: 0.6.11
flashinfer-jit-cache: Not installed
=== Torch Version Info ===
Torch version: 2.11.0+20260320.70d99e9.cu130
CUDA runtime available: Yes

=== Environment Variables ===
FLASHINFER_CACHE_DIR: /traindata/svyatoslav/.cache/flashinfer
FLASHINFER_CUBIN_DIR: /usr/local/lib/python3.12/dist-packages/flashinfer_cubin/cubins
FLASHINFER_CUDA_ARCH_LIST: {(10, '0a')}
[...]

PR #4144, opened 2026-07-25
fix(gdn): defer CUDA device queries in gdn_decode_bf16_state to lazy helpers

### LINKED ISSUE #4143: import flashinfer unconditionally crashes on machines without a CUDA device (regression of #3262/#3293)
## Environment

Reproduced in a clean `python:3.12-slim` container (CPU-only torch, no CUDA device), running `flashinfer/collect_env.py` standalone since `import flashinfer` itself is what fails:

```
### FlashInfer environment report
==== FlashInfer ====
flashinfer : <import failed: ModuleNotFoundError: No module named 'flashinfer'>

==== Python / Platform ====
Python            : 3.12.13 (main, Jul 14 2026, 02:09:00) [GCC 14.2.0]
Platform          : Linux-6.8.0-134-generic-x86_64-with-glibc2.41
Container         : yes

==== GPU / Driver ====
Driver version       : <nvidia-smi unavailable>
CUDA_VISIBLE_DEVICES :
GPUs (torch)         : torch.cuda.is_available() == False

==== PyTorch ====
torch                          : 2.13.0+cpu
torch.version.cuda             : None
```

## Bug description

`import flashinfer` (and `import flashinfer.gdn_decode`) crashes on any machine without a visible CUDA device. This is a regression: #3262 reported the same crash, #3293 fixed it by moving the module-scope `torch.cuda.get_device_properties()`/`get_device_capability()` calls in `gdn_decode_bf16_state.py` into lazy, device-parameterized helpers. #3502 added new code to the same file and reintroduced the exact eager module-scope pattern (new `NUM_SMS`/`_GPU_MAJOR`/`_USE_PACKED_FMA` module globals), without reusing the established fix.

Expected: importing `flashinfer` should never require a CUDA device to be present; kernels should only need CUDA when actually invoked.

## Minimal reproduction

```
CUDA_VISIBLE_DEVICES= python3 -c "import flashinfer"
```

## Error messages / logs

```
  File ".../flashinfer/__init__.py", line 107, in <module>
    from .gdn_prefill import chunk_gated_delta_rule as chunk_gated_delta_rule
  File ".../flashinfer/gdn_prefill.py", line 25, in <module>
    from .gdn_kernels import (
  File ".../flashinfer/gdn_kernels/__init__.py", line 22, in <module>
    from .gdn_decode_bf16_state import (
  File ".../flashinfer/gdn_kernels/gdn_decode_bf16_state.py", line 2701, in <module>
    NUM_SMS = torch.cuda.get_device_properties(0).multi_processor_count
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../torch/cuda/__init__.py", line 737, in get_device_properties
    _lazy_init()
    ^^^^^^^^^^^^
  File ".../torch/cuda/__init__.py", line 522, in _lazy_init
    raise AssertionError("Torch not compiled with CUDA enabled")
AssertionError: Torch not compiled with CUDA enabled
```

(On a CUDA-enabled torch build with `CUDA_VISIBLE_DEVICES=`, this raises `RuntimeError: No CUDA GPUs are available` at the same line instead, same underlying cause as #3262.)

`flashinfer/gdn_kernels/__init__.py` and `flashinfer/gdn_decode.py` both wrap this import in `except (ImportError, RuntimeError)`, so the CUDA-enabled-but-no-GPU case (`RuntimeError`) can look partially guarded elsewhere. The CPU-only-torch case raises `AssertionError` instead, which neither guard catches, and either way the crash happens at package-import time rather than being contained to the one optional backend.

## Additional context

`git log -G` on the two eager lines shows the full lifecycle: introduced (#2679) -> fixed (#3293, "allow importing gdn_decode without a CUDA device") -> reintroduced (#3502). I have a fix that reapplies #3293's pattern (lazy `get_num_sm`/`get_sm_version` helpers from `flashinfer.cute_dsl.utils`/`flashinfer.cute_dsl.fp4_common`, already used by every other file in `gdn_kernels/`) to the code #3502 added, plus a regression test that would have caught this reintroduction. Opening the PR now with `Fixes #<this issue>`.

AmirF194:
## 📌 Description

`gdn_decode_bf16_state.py` computed `NUM_SMS`/`_GPU_MAJOR`/`_USE_PACKED_FMA` via eager
`torch.cuda.get_device_properties(0)`/`get_device_capability(0)` calls at module scope, so
`import flashinfer` crashed on any machine without a visible CUDA device.

#3293 fixed this exact pattern for #3262 by moving the calls into lazy, device-parameterized
helpers (`get_num_sm`/`get_sm_version`, already the convention in every other `gdn_kernels/*.py`
file). #3502 added new code to the same file and reintroduced the eager module-scope pattern
without reusing that fix. `git log -G` on the two lines shows the full lifecycle: introduced
(#2679) -> fixed (#3293) -> reintroduced (#3502).

This PR reapplies #3293's fix to the code #3502 added: `_select_tile_v_for_mtp` and
`_get_bf16_mtp_config` now take a `device` keyword, threaded from `q.device` at each of the three
call sites (all three functions `assert q is not None` earlier in the same function body). The
module-scope globals are removed.

## 🔍 Related Issues

Fixes #4143

### ✅ Pre-commit Checks

## 🧪 Tests

## Reviewer Notes

Verified in a clean `python:3.12-slim` container (CPU-only torch, `CUDA_VISIBLE_DEVICES=`, no
GPU hardware): `import flashinfer` and `import flashinfer.gdn_decode` both crash on unpatched
`main` (`AssertionError: Torch not compiled with CUDA enabled` at the `NUM_SMS` line) and both
succeed after this change. `tests/test_gdn_decode_cpu_import.py` pins this as a subprocess test
(fails on main, passes on the branch, confirmed both ways) so a future addition to this file
can't reintroduce the eager pattern a third time.

I have not run this against an actual GPU: the three call sites already had `assert q is not
None` in scope before reaching the changed lines, and `get_num_sm`/`get_sm_version` are the
same helpers already exercised by every sibling kernel file, but the packed-FMA / SM-count
dispatch itself is unverified on real hardware in this session.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GPU-aware kernel selection across systems with multiple GPU types.
  * Fixed wide-vector and multi-token prediction execution paths to select appropriate hardware optimizations at runtime.
  * Improved reliability when importing the package without visible CUDA devices.
  * Ensured device-specific decisions are based on the active query tensor’s GPU.

* **Tests**
  * Added regression coverage for importing the package when GPUs are hidden.
  * Added validation for CUDA-dependent behavior in environments without visible devices.


AmirF194:
Checking in after a week with no review yet. Happy to split the fix further or add more test coverage if that would help; no rush.

tests/test_gdn_decode_cpu_import.py
bkryu, line 1: There is a [tests/gdn](https://github.com/flashinfer-ai/flashinfer/tree/main/tests/gdn) directory and we'd like to organize our test files.

However, what might be better is to simply put the tests in this file inside a suitable existing file because it covers only two cases.

bkryu (commented):
Thank you for the fix @AmirF194, I left one comment about the organization of added tests. Can you move the newly added tests into an existing file instead of adding a new file?

AmirF194:
Moved the two import-crash tests into `tests/gdn/test_decode_pretranspose_bf16_padded_pool.py` and deleted the standalone file. Reran them under CPU-only Docker (CUDA_VISIBLE_DEVICES unset): both still pass in the new location.

AmirF194:
Rebased past main to resolve a conflict: main added `dtype_compat.as_bf16` and a `_dtype_key` helper in this same file (unrelated to this fix) since the internal CI ran. Kept both changes; they don't overlap functionally with the lazy device-query fix here.

Re-verified in Docker (CPU-only, `CUDA_VISIBLE_DEVICES=` unset): `import flashinfer`, `import flashinfer.gdn_decode`, and `import flashinfer.gdn_kernels.gdn_decode_bf16_state` all succeed post-rebase, and both regression tests in `tests/gdn/test_decode_pretranspose_bf16_padded_pool.py` (`test_import_gdn_decode_without_cuda_device`, `test_import_flashinfer_without_cuda_device`) pass. `ruff check`/`ruff format --check` clean. Both commits SSH-signed and verified.

I don't have GPU hardware here to re-run the actual kernel numerics, so the internal CI pipeline is the source of truth for that (it previously reported 16/18, all unit tests green across every GPU/CUDA combination; the two `B300 (multi-GPU)` jobs came back with no individual test or infrastructure failure extractable from the report, so I can't tell from here whether that's related to this change or not). Happy to have `/bot run` re-triggered on the rebased head whenever convenient.

AmirF194:
Rebased past main again to clear a second conflict in the same file: main's new `build_and_load_cute_dsl_kernel`/`make_kernel_name` imports landed next to this fix's `get_sm_version`/`get_num_sm` imports, same shape as the 08-12 rebase. Kept both, no functional overlap (checked call sites). CPU-only import scope unchanged.

AmirF194:
Closing this one: #4507 just merged and removes these exact two module-scope calls (NUM_SMS/_GPU_MAJOR) as part of a broader per-device target refactor, so the import crash from #4143 is fixed there instead. Verified the new gdn_decode_bf16_state.py has zero torch.cuda calls left at module scope.

PR #4146, opened 2026-07-25
fix(cudnn): key prefill graphs by tensor descriptors

davidkny22:
## 📌 Description

The cuDNN prefill graph cache did not distinguish tensors with the same shape but different strides. On an RTX 4070 Laptop GPU, a contiguous K call followed by a same-value, token-strided K call reused the first graph and returned incorrect output (1.8754 maximum error against an FP32 PyTorch reference).

cuDNN bakes tensor descriptor metadata into the compiled graph. This patch:

- adds the dtype, device, shape, and stride metadata used to build the Q/K/V and optional tensor descriptors to the cache key;
- normalizes the output dtype the same way as the graph builder;
- derives `h_kv` and `d_vo` from V, matching the graph builder.

Pointers, tensor values, caller-provided output/LSE buffers, and the dynamic packed-token count remain outside the key, so calls with the same graph descriptors still reuse the cached graph.

## 🔍 Related Issues

None.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

RTX 4070 Laptop GPU, SM89, cuDNN 9.20.0:

- same-shape K stride regression: maximum error 1.8754 before, 0.00424 after;
- `pytest -q tests/attention/test_cudnn_graph_cache_key.py`: 9 passed;
- `pytest -q tests/attention/test_cudnn_prefill_token_indptr.py`: 65 passed, 65 skipped;
- paged prefill with causal/noncausal, LSE on/off, and repeated calls: 8 passed;

I did not complete the full `tests/attention -k cudnn` run locally because its largest cases do not fit on the laptop's 8 GB GPU. I tested mixed BF16/FP16 output at the key level only: a cold mixed-output call is already incorrect with cuDNN 9.20 on this GPU, so it cannot serve as a runtime regression for this cache fix.

## Reviewer Notes

The added key fields are limited to metadata used when building cuDNN tensor descriptors. Calls with identical descriptor metadata still share a cache entry.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cuDNN SDPA attention graph-cache correctness by making cache keys account for output data type, tensor layouts/strides/shapes/devices, and attention scale so different configurations no longer reuse stale graphs.
  * Fixed value-dimension derivation for prefill so KV sizing is consistent with the value cache.
  * Ensured cache reuse only occurs when tensor descriptors and relevant metadata are truly equivalent (including optional offset signatures and block tables).

* **Tests**
  * Added regression tests asserting cache-key uniqueness/reuse for output dtype, descriptor changes, scale, offsets stability, and block-table content.
  * Added CUDA-dependent stride coverage comparing cuDNN prefill results against a reference implementation.

### DIFF
--- a/flashinfer/cudnn/prefill.py
+++ b/flashinfer/cudnn/prefill.py
@@ -67,6 +67,18 @@ def _create_cudnn_handle(stream: torch.cuda.Stream):
     return _cudnn_handle
 
 
+def _tensor_descriptor_signature(
+    tensor: Optional[torch.Tensor], *, ignore_leading_dim: bool = False
+):
+    """Return hashable metadata used to build a cuDNN tensor descriptor."""
+    if tensor is None:
+        return None
+    shape = tuple(tensor.shape)
+    if ignore_leading_dim and tensor.dim() == 3:
+        shape = shape[1:]
+    return tensor.dim(), tensor.dtype, tensor.device, shape, tuple(tensor.stride())
+
+
 # Tensor ids
 class UIDs(Enum):
     RESERVED_INVALID_UID = 0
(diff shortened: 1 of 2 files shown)


davidkny22:
Ready for review. This fixes a silent wrong-output case: the cuDNN prefill graph cache does not distinguish tensors with the same shape but different strides, so a contiguous K call followed by a token-strided K call reuses the first graph and returns wrong values. Reproduced on an RTX 4070 Laptop at 1.8754 maximum error against an FP32 reference, with a regression test included.

cc @Anerudhan, since the graph key and handle construction are yours.

Issue #3628, opened 2026-06-13
[RFC] MXFP8 block-scaled prefill attention for SM120a (consumer Blackwell)


ultism:
## Summary

Add an open-source, source-level **MXFP8** prefill / ragged attention kernel for
**SM120a / SM121a** (consumer Blackwell — GeForce RTX 50 series, GB20x).
MXFP8 here means the OCP microscaling FP8 format: `e4m3` data with per-32-element
`ue8m0` block scales, accumulating in FP32.

FlashInfer currently has no block-scaled FP8 attention that runs on SM120. This RFC
proposes one, motivated by a concrete hardware property of consumer Blackwell measured
below.

### 1. The block-scaled tensor path escapes the consumer FP32-accumulate throttle (measured)

On consumer Blackwell the FP32-accumulate tensor throughput is halved **for the legacy
warp-MMA instructions** — but the **block-scaled tensor instruction is not throttled**.

Register-resident MMA microbenchmark, RTX 5060 Ti (SM120, GB206, 36 SMs, CUDA 13.3),
pure tensor-core issue rate, no memory traffic:

| inputs / accumulate | SASS instruction | TFLOP/s |
|---|---|---:|
| FP16 / FP16 acc | `HMMA.16816.F16` | ~103 |
| FP16 / FP32 acc | `HMMA.16816.F32` | ~51 |
| FP8 e4m3 / FP32 acc (plain) | `QMMA.16832.F32.E4M3.E4M3` | ~102 |
| FP8 e4m3 / FP16 acc (plain) | `QMMA.16832.F16.E4M3.E4M3` | ~206 |
| **MXFP8 e4m3 + ue8m0 block-scaled / FP32 acc** | **`QMMA.SF.16832.F32.E4M3.E4M3.E8`** | **~202** |

Key ratios:

- **MXFP8 block-scaled / BF16(FP32-acc) = 3.95×** (202 vs 51 TFLOP/s).
- Legacy FP32-acc throttle = 2.0× (`HMMA.16816.F16` vs `HMMA.16816.F32`), and it equally
  halves *plain* FP8 (`QMMA...F32` 102 vs `QMMA...F16` 206) — but **not** the block-scaled
  path (`QMMA.SF` stays at ~202 with FP32 accumulate).

In other words: on consumer Blackwell the block-scaled MXFP8 MMA is the only way to get
full-rate tensor throughput *together with* FP32 accumulation. A per-tensor / plain FP8
[...]

PR #4149, opened 2026-07-26
feat(attention): MXFP8 ragged prefill attention for SM120/121

ultism:
## Summary

Implements RFC #3628.

Add an **MXFP8 ragged prefill attention kernel for SM120a/SM121a** (consumer Blackwell — RTX 50 series, GB20x), exposed as `flashinfer.mxfp8_attention_sm120_fwd` and the plan/run wrapper `MXFP8AttentionSM120Wrapper`. FlashInfer currently has no block-scaled FP8 attention that runs on SM120; this PR adds one, following the standalone-op structure of `nvfp4_attention_sm120`.

## Motivation (from the RFC)

On consumer Blackwell, FP32-accumulate tensor throughput is halved for the legacy warp-MMA instructions — but the **block-scaled tensor instruction is not throttled** (register-resident MMA microbenchmark, pure issue rate):

| inputs / accumulate | SASS | TFLOP/s (GB206) | TFLOP/s (GB202) |
|---|---|---:|---:|
| FP16 / FP32 acc | `HMMA.16816.F32` | ~51 | ~253 |
| FP8 e4m3 / FP32 acc (plain) | `QMMA.16832.F32.E4M3.E4M3` | ~102 | ~506 |
| **MXFP8 e4m3 + ue8m0 block-scaled / FP32 acc** | **`QMMA.SF.16832.F32.E4M3.E4M3.E8`** | **~202** | **~993** |

The block-scaled MXFP8 MMA is the only way to get full-rate tensor throughput *with* FP32 accumulation on these parts, so the mainloop keeps both GEMMs on the SM120 block-scaled MMA atom (`SM120_16x8x32_TN_VS`). (GB202 numbers measured on RTX 5090 with the same `mma_peak` probe and cross-checked with cuBLAS — bf16→FP32 GEMM tops out at 237 TFLOP/s, 94% of the throttled roof.)

## What

- **Kernel** (`include/flashinfer/attention/sm120/mxfp8_attention_sm120/`): warp-specialized producer/consumer prefill kernel, TMA-fed, persistent with a host-built LPT (longest-processing-time) tile work-list (`BatchPrefillPersistentTileScheduler`). Framework-agnostic headers, raw pointers.
- **Contract**: ragged varlen FP8 (`float8_e4m3fn`) Q/K/V `[total_tokens, heads, 128]`, per-tensor scales (`q_scale·k_scale` fold into `sm_scale` on the host, `v_scale` into the PV output in-kernel — the kernel's `kUniformFp8` mode), GQA (`num_qo_heads % num_kv_heads == 0`), and FlashInfer slice-3 causal semantics (query `m` attends keys `[0, m + kv_len - qo_len]`), which covers prefix-cache hits and chunked-prefill continuation natively. Returns `(out [total_q, Hq, 128] fp16/bf16, lse [total_q, Hq] fp32)`.
- **Plan/run split** (`MXFP8AttentionSM120Wrapper`): `plan()` performs the batch-composition host work once per step (128-row padding layout, LPT work lists) and `run()` is tensors-only per layer — matching the calling convention of the other FlashInfer prefill wrappers. Scales/`sm_scale` are run-time (per-layer quantization); `causal` is plan-time (scheduler cost model). `mxfp8_attention_sm120_fwd` remains as a one-shot plan+run convenience.
- **Plumbing** mirroring `nvfp4_attention_sm120`: TVM-FFI binding (`csrc/mxfp8_attention_sm120/`), JIT module with `sm120a_nvcc_flags`, AOT registration, trace templates + `fi_trace_out` examples + `tests/trace/example.py` entry, API docs, and a standalone microbenchmark (`benchmarks/bench_mxfp8_attention_sm120.py`).

## Testing

- `tests/attention/test_mxfp8_attention_sm120.py` (SM120/SM121-gated): ragged GQA causal (bf16/fp16 out), ragged MHA non-causal, single long request (2048), prefix-append (`kv_len > qo_len`), wrapper plan-reuse, and the one-shot functional path. The reference replays the kernel's exact numerics — unnormalized P requantized to e4m3 at a fixed scale of 256, normalized by the unquantized row sum — so the gates are tight against masking/stride/GQA indexing bugs while tolerating only accumulation-order noise. **7/7 passing on RTX 5060 Ti (SM120).**
- Trace-template consistency tests pass (auto-discovered).
- ruff + clang-format (19.1.1) clean.
- Before upstreaming, the same kernel was validated end-to-end in vLLM 0.21 (Qwen3-8B-AWQ, fp8 KV cache, `enable_prefix_caching=True`, chunked prefill): long-context outputs matched the stock bf16-prefill backend token-for-token.

## Performance

RTX 5090 (sm_120), same ragged batch on both sides, event timing (min of 5 alternating A/B rounds x 100 iters) cross-checked against single-launch `gpu__time_duration` from `ncu --set full` (agreement <3%). Baseline: flashinfer `BatchPrefillWithRaggedKVCacheWrapper(backend="fa2")`, bf16.

| shape (identical batch both sides) | fa2 | ours | **speedup** | tensor-pipe active (fa2 / ours) |
|---|---:|---:|---:|---:|
| 16 mixed varlen requests, Hq8/Hkv2, causal | 0.388 ms | 0.141 ms | **2.74x** | 32.5% / 46.9% |
| same batch, Hq32/Hkv8, causal | 1.165 ms | 0.566 ms | **2.06x** | 42.7% / 49.3% |
| 8x2048 (Qwen3-8B-like), Hq32/Hkv8, causal | 1.317 ms | 0.614 ms | **2.14x** | 44.0% / 51.5% |
| single 16384-token request, Hq16/Hkv4, causal | 5.434 ms | 2.234 ms | **2.43x** | 41.5% / 54.8% |

Effective compute: fa2 saturates at ~200-209 TFLOP/s on the larger shapes — i.e. it already sits at 84-88% of the consumer-Blackwell `HMMA.F32` (bf16, FP32-accumulate) hardware ceiling (cuBLAS-confirmed). This kernel reaches 413-492 TFLOP/s with the same FP32-accumulate semantics on the block-scaled path, while its tensor pipe is only 47-55% busy — roughly 2x of headroom remains toward the block-scaled ceiling. Full methodology + raw reps: [`ultism/fa-blackwell` docs/5090_evidence.md](https://github.com/ultism/fa-blackwell/blob/master/docs/5090_evidence.md).

Note: these are kernel-level numbers for the attention itself; the op-level Python path additionally does the per-layer padding/scatter work described above.

## Follow-ups (not in this PR)

- `BatchPrefillWithRaggedKVCacheWrapper` dispatch integration (this PR is the standalone op only).

AI-assisted development (kernel authored and validated in an external worktree with ncu-guided optimization before upstreaming).

## Summary by CodeRabbit

* **New Features**
  * Added ragged FP8 prefill attention for compatible SM120/SM121 GPUs.
  * Added one-shot and reusable wrapper APIs with causal and non-causal support.
  * Added package exports, build integration, and trace definitions.
* **Documentation**
  * Added API documentation for the new attention functionality.
* **Tests**
  * Added coverage for GQA, MHA, prefix-append, plan reuse, long requests, causal masking, and zero-length requests.
* **Benchmarks**
  * Added a performance benchmark reporting latency and TFLOP/s.


ultism:
Thanks for the thorough review. Disposition of all 14 findings in 1f0c01d (8/8 attention tests re-pass on RTX 5060 Ti; ruff + clang-format clean):

**Accepted & fixed (11)**

- 🔴 Work lists on default device (`_to_i32` → `.cuda()`): real multi-GPU bug — `_build_lpt_work_lists` now takes the plan device and allocates via `torch.tensor(..., device=device)`.
- 🟠 `cudaFuncSetAttribute`/launch unchecked: both branches now propagate via `TVM_FFI_ICHECK`, plus a `cudaGetLastError` after launch.
- 🟠 Trace template output dtypes: `out`/`lse` were serializing as fp8 (inherited from inputs) — now `bfloat16`/`float32` in both templates; both `fi_trace_out` JSONs regenerated (diff is exactly the two dtype fields per file).
- 🟠 Hopper include guards + top-level `flashinfer` symbol collision: guards renamed to `FLASHINFER_ATTENTION_SM120_MXFP8_*`, schedulers moved into `flashinfer::mxfp8_attention_sm120`. `named_barrier.cuh` turned out to be entirely unused by this kernel (it synchronizes on raw `cutlass::arch::NamedBarrier` ids), so the file is deleted rather than carried as a second copy — this also resolves the `WarpScheduler` nit.
- 🟠 `WorkTileInfo` 10-of-11-member initializers (x4): confirmed — the missing `batch_idx` shifted `counter`/`ptr_begin`/`ptr_end` (latent, since `is_valid()` still returned false). Fixed in both schedulers' `get_initial_work`/`get_next_work`.
- 🟡 `out_Mnb` dereferenced under the `out_Ppre` guard: now independently null-checked.
- 🟡 RUF059 unused `total_q`: fixed.
- 🔵 Same-device validation: all tensor inputs/outputs are now checked against `q`'s device before the device guard.
- 🔵 Unused host includes (`<cstdio>`/`<cstring>`/`<vector>`): dropped.
- 🔵 `else` + `CUTLASS_PRAGMA_UNROLL` without braces (x3): braced.
- 🔵 Zero-length request coverage: added `test_mxfp8_attention_sm120_zero_length_request` with interleaved `qo_len == 0` requests (padding/scatter/work-list paths all skip them cleanly).

**Declined (1)**

- `tests/trace/example.py` fixture list: false positive — the expected-output list already contains exactly `mxfp8_attention_sm120_fwd_head_dim128.json` and `mxfp8_attention_sm120_run_head_dim128.json` (lines 63–64), and the regenerated JSONs match those names.

**Deferred (1)**

- Caching padded buffers on the plan / kernel-native `[Sk, Hkv, D]` V layout: agreed this is the right direction, but it's a design change, not a fix — the op-level per-layer padding/scatter cost is already documented in the PR description, and the wrapper-dispatch follow-up is where the V layout should be addressed (the kernel's TMA constraint that motivated the transpose is documented in the binding). Keeping this PR focused on the standalone op.

Issue #3920, opened 2026-07-10
[RFC] Autotuner v2: FlashInfer-managed persistent cache and deployment-matched measurement


YangXu1990uiuc:
**Status (2026-09-08)**: #3861 **merged** (`3a75b4e3`, ships in v0.7.0; opt-in, v1 untouched). Framework migration has started: vLLM draft vllm-project/vllm#55671, validated on 4×B200 at TP=1/2/4 (see the status comment below). Design rationale and the graduation plan live in-tree at `docs/design_docs/autotuner_v2.md`; `docs/autotuning.rst` documents `autotune_v2` as experimental. This RFC remains the discussion thread for the roadmap and open questions.

## Summary

FlashInfer should own the persistence and compatibility rules for its autotuning results. Frameworks such as vLLM and SGLang should decide *when* to warm up and how to coordinate distributed workers — but they should not construct FlashInfer cache keys, interpret its JSON schema, or manage cache files, as they do today.

The proposal has four parts:

1. **A managed persistent cache** (`autotune_v2()`): environment-hashed directories, one atomic entry per tuned operation, invalid entries are misses (never errors), old generations left on disk.
2. **A measurement policy** (`MeasurementPolicy`): tune the way you deploy — eager vs CUDA-graph execution as the primary axis, cold-L2 as a secondary one.
3. **A runner contract**: rules that make "serving executes the tactic the tuner chose" structurally guaranteed rather than convention-enforced.
4. **An accuracy harness**: tuner quality ("picks the fastest tactic") becomes a measured claim (regret vs. an oracle), not an assumption.

Autotune data is a disposable performance optimization. Cross-version cache migration is a non-goal: after an upgrade, paying the tuning cost once for the new environment is acceptable. This follows the compiler-cache patterns of Triton and `torch.compile`/Inductor: key artifacts by program + compiler + environment, publish atomically, treat unusable entries as work to rebuild, own the local cache, and expose only an opaque export/import boundary.

## Evidence 1: how frameworks actually use v1

A survey of vLLM and SGLang (HEAD, 2026-07) shows both use exactly one pattern — **one-shot startup warmup around a whole-model dummy forward, before CUDA-graph capture; serving always runs outside any context**:

[...]

Issue #3673, opened 2026-06-19
[Bug] fp8_gemm autotune tactic=-1 on 0.6.12 (vLLM ModelOpt FP8); CUDA crash on 0.6.13rc1 (SM120)


Saddss:
**This issue was closed prematurely and is retained only as a placeholder.**

Internal model identifiers were included by mistake. A corrected report will be filed after full investigation, using only desensitized workload description (quantization method, kernel path, hardware).

Please ignore the original content below this line.

---

[Content redacted]

Saddss:
Closing — opened prematurely. Will re-file after complete investigation with desensitized workload description (quantization/backend only, no internal model identifiers).

Issue #3566, opened 2026-06-10
bmm_fp8 cuDNN backend raises "Plan index N is invalid" at execute time on GB10 (sm_121a) for small-M FP8 GEMM, killing the engine


jeremy-newhouse:
## Environment

- Hardware: NVIDIA DGX Spark (GB10, sm_121a, integrated Blackwell, unified memory), single GPU
- Image: `vllm/vllm-openai:v0.22.0` (vLLM 0.22.0)
- flashinfer 0.6.11.post2, CUDA 13.0.x
- Model: `nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-NVFP4` (modelopt FP4), served single-stream (concurrency 1), vLLM V1 engine

## Summary

On GB10 (sm_121a), the cuDNN runner selected by `bmm_fp8` with `backend="auto"` builds a plan at autotune/build time that is then **rejected at execute time** for a small-M dense FP8 projection of shape ~`(1, 4096, 1024)`:

```
RuntimeError: Plan index 13 is invalid
```

The call path is `vllm.flashinfer_scaled_fp8_mm -> flashinfer.bmm_fp8 -> fp8_gemm_sm100 -> _cudnn_gemm_fp8` (the runner is appended in `fp8_gemm_sm100`, `flashinfer/gemm/gemm_base.py:1293` on current main). The error propagates as a vLLM `EngineDeadError` and kills the server mid-run.

The crash is shape/tactic-dependent: deterministic for a given shape, but whether a run reaches that shape depends on what it decodes. Short-output workloads hit it intermittently; sustained long-output decode hits it reliably.

## Reproduction

Serve the NVFP4 Nemotron-3-Super on a GB10 (V1 engine, concurrency 1). Two ways to reach it:

- **Short / intermittent:** `vllm bench serve --dataset-name sharegpt --max-concurrency 1` (~100 prompts). Two separate runs each crashed at request ~25-26 — one with `--kv-cache-dtype fp8`, one with the default KV dtype (which is how we ruled the KV cache path out; the trigger is the dense FP8 projection GEMM, independent of KV dtype). A separate 200-prompt run happened to complete — consistent with intermittent plan selection on short generations.
- **Sustained / reliable:** drive one long-output reasoning generation (an AIME-style problem with thinking enabled, tens of thousands of output tokens; we use the model's built-in MTP speculative decoding). The cuDNN `bmm_fp8` crashes ~97 s into sustained decode, every time.

[...]

PR #4165, opened 2026-07-27
fix(gemm): gate cuDNN out of SM12x bmm_fp8 auto when override_shape unavailable

Saddss:
## 📌 Description

Gate the cuDNN FP8 GEMM backend out of the `bmm_fp8(backend="auto")` candidate
list on SM12x when cuDNN `override_shape` is unavailable (cuDNN backend <
9.23.1). On this narrow configuration the cuDNN path is both unsafe and
unnecessarily slow; cublas/cutlass already cover the workload.

## Root cause

Two independent failure modes on SM12x (RTX 5090), both removed by the gate:

1. **Unbounded host-side compilation (perf).** Without `override_shape`, cuDNN
   builds a fresh `policy=ALL` execution graph per distinct serving `(M,K,N)`
   shape (~250 ms host-side; measured 236 ms/shape). Under load this is
   unbounded — RFC #3920, runner-contract rule 6.
2. **Async CUDA fault (crash).** The cuDNN FP8 serving path raises
   `cudaErrorMisalignedAddress` asynchronously at the next sync, which #3707's
   `forward()` try/except cannot catch (it only handles synchronous exceptions).

## 📊 Evidence chain (RTX 5090, SM120, vLLM nightly, online-replay load test)

Same container; only the FlashInfer wheel (and cuDNN package) differs.

**Reproducibility (input/output lengths only)**
- Model checkpoint (open-sourced): <https://huggingface.co/wangqia0309/gemma-4-26B-A4B-it_nvfp4_experts_fp8_dense_fp8_attn-kv_fp8> (NVFP4 mixed-precision: FP8 dense + attention, FP8 KV cache)
- Input: `max-model-len 8192`, `max-num-batched-tokens 8192`, `max-num-seqs 64`
- Output: `max-tokens 200`
- MTP: `num_speculative_tokens=3`
- vLLM dtype/quant config: `--quantization modelopt --kv-cache-dtype fp8 --moe-backend cutlass -O3 --gpu-memory-utilization 0.95`

| FI build | cuDNN in SM12x auto | override_shape | result |
|---|---|---|---|
| 0.6.15.post1 (BAD baseline) | yes | unavailable (backend 92000) | 6 rounds then crash; p50 96/78/71/78/74/40 s |
| main + #3707 | yes | unavailable (backend 92000) | ~5 rounds crash (`misaligned`); p50 26/17.6/17/17.5 s |
| **main + #3707 + this gate** | no (gated) | unavailable (backend 92000) | **12/12 PASS, steady p50 10.31 s, 0 crashes** |
| main + #3707 (NO gate) | yes | **available** (backend 92400, `nvidia-cudnn-cu13` 9.24) | **12/12 PASS, steady p50 10.69 s, 0 crashes** |

The last row is the **clean** override-available test: with the cu13 cuDNN 9.24
stack (override_shape available), cuDNN-in-auto does *not* crash and matches the
gated run's p50 — validating the conditional gate's `else` branch (keep cuDNN
where override_shape amortizes the build).

**Why the gate proves the faulting path is cuDNN (elimination).** The gate
removes *only* cuDNN from the candidate list; the `cutlass_sm12x` and `cublas`
kernels and the autotuner's within-backend winner selection are unchanged. At
serving time only the winning runner executes, and on 92000-without-override
the cuDNN runner is selected (per-tactic warm median ≈ 0.05–0.07 ms, winning
20/20 in the §7.3 micro-bench). Removing cuDNN eliminates the crash; the
remaining cutlass/cublas kernels do not fault (12/12). Therefore the faulting
path is cuDNN — no `compute-sanitizer` pinning needed.

**Independent reproduction:** handoff §6 — on post2, removing cuDNN from the
SM120 candidate list → 12/12, steady p50 ≈ 12 s.

## Scope and limitation

- **Scoped fix.** The gate is `is_sm120_supported and not
  _is_cudnn_override_shape_available()` — SM100/103/107/110 and SM12x-with-
  override are unchanged (cuDNN stays a candidate where M-bucketing amortizes
  the build).
- **"Upgrade cuDNN" is tested clean here.** With the matching CUDA-13 cuDNN
  (`nvidia-cudnn-cu13==9.24.0.43`, backend 92400, override_shape available) +
  cuDNN still in `auto`, the run is **12/12 PASS, p50 10.69 s, 0 crashes** — cuDNN
  is safe when `override_shape` is available, so the conditional gate keeps it.
  (An earlier `nvidia-cudnn-cu12==9.24` attempt was a CUDA 12/13 library
  mismatch that broke cuBLASLt symbol loading — not a cuDNN-FP8 signal.)
- **No upstream improvement yet:** vLLM v0.26.0 stable pins
  `flashinfer-python==0.6.14` (BAD per the §4 bisect); vLLM nightly pins
  0.6.15.post1 (BAD); FlashInfer main + #3707 still crashes on SM12x without
  this gate.

## 🔍 Related Issues

- RFC #3920 (Autotuner v2, runner-contract rule 6) — unbounded serving-time compilation
- #3707 (engine/knob tactics + sync fallback) — fixes stale plan-index crash, not the async fault
- #3437 (multi-tactic / `policy=ALL` for cuDNN) — introduced the per-shape ALL build cost
- #2914 (cuBLASLt + FP8 multi-tactic autotuning) — first BAD boundary
- #3566, #3673, #3255 — SM12x cuDNN/plan-index field reports

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

Added `test_bmm_fp8_heuristic_gates_cudnn_on_sm12x_without_override_shape` —
monkeypatches SM version + cuDNN availability so it runs on any host without
SM12x hardware or a cuDNN install; asserts cuDNN is excluded when
`override_shape` is unavailable and retained when available.

## Reviewer Notes

- Scope is deliberately minimal: one heuristic branch + one test. No change to
  cuDNN code, the override_shape path, or non-SM12x architectures.
- The one-shot warning uses a module-level `set` (mutation, no `global`) so it
  does not spam per-call (the heuristic runs on every `bmm_fp8(backend="auto")`).
- DCO: `Signed-off-by: Saddss <2872669061@qq.com>`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic FP8 matrix multiplication backend selection on supported GPUs.
  * Prevented incompatible cuDNN paths from being selected when required shape support is unavailable.
  * Preserved other compatible backend options for continued operation.


Saddss:
Hi @bkryu, @nv-yunzheq, and @qiching — sorry for the direct message.
Would any of you have time to help review this pull request? If the approach seems reasonable, could you then initiate the full CI process?
This is a narrowly targeted fix for the SM12x architecture, specifically related to the `bmm_fp8(backend="auto")` function. It excludes the use of cuDNN only when the `override_shape` parameter is not available (i.e., when using a cuDNN backend version earlier than 9.23.1). All other cuDNN versions and architectures that aren’t part of the SM12x family remain unaffected by this fix.
This change resolves an issue where the system crashed due to the asynchronous `cudaErrorMisalignedAddress` error, as well as problems related to repeated graph compilations with `policy=ALL` for different shapes of data, observed on the RTX 5090 / SM120. In tests, the modified version of the code completed all 12 online-replay rounds without any crashes, with a median execution time of 10.31 seconds. As a control, cuDNN 9.24, with the `override_shape` option enabled, also completed all 12 rounds without any crashes, but with a median execution time of 10.69 seconds.
The patch is intentionally small: it involves just one heuristic branch and one focused unit test. It’s closely related to issues #3437 and #3707.
Please let me know if any changes or additional reproduction data would help move this forward. Thanks!

nv-yunzheq:
@yanqinz2 Could you take a look?

yanqinz2:
On my radar. We acknowledge the host compilation overhead issue, but do not expect a hard failure. Let me investigate and review this PR

yanqinz2:
Hi @Saddss, for the host overhead issue, we can reproduce, but we cannot reproduce the  cudaErrorMisalignedAddress issue. Do you have any repro that can share with us?

Saddss:
@yanqinz2 Thanks for looking into this — and for confirming the host-compilation overhead. Here is a **self-contained repro** for the hard failure (verified on RTX 5090 / SM120 with the script below; **no private dataset**).

### Why an isolated GEMM microbench will not hit the hard failure

`cudaErrorMisalignedAddress` / the cuDNN execute fault is a **serving-path** issue:

1. vLLM calls `bmm_fp8(backend="auto")`.
2. On SM12x with cuDNN backend **< 9.23.1** (`override_shape` unavailable), the heuristic still includes `cudnn`.
3. Autotuner picks `CudnnFp8GemmRunner`. New MTP / continuous-batching `(M,K,N)` shapes then hit `build_plans(policy=ALL)` (hundreds of ms host) and later `execute_plan_at_index`.
4. The illegal access is in that cuDNN execute; CUDA may report it asynchronously at a later sync (e.g. `_get_valid_sampled_token_count`), so the stack top is often **not** cuDNN.

A single-op GEMM call (even with matching dtypes) stays on a stable shape, never exercises the dynamic serving path, and typically looks fine (kernel ~0.05ms, high cosine).

### What we verified

Pinned image (please **do not** use floating `:nightly` — cuDNN can move and the crash disappears once `override_shape` is available):

```
vllm/vllm-openai@sha256:b356b39e510376c56ae16f530cc6f2d0e10934631a009e30ba9c13cc02ae7ffe
# image ID b54e9e2f48ea, cuDNN backend 92000, FI 0.6.15.post1
```

- Target: open mirror with the same structure as our internal NVFP4-mixed checkpoint — https://huggingface.co/wangqia0309/gemma-4-26B-A4B-it_nvfp4_experts_fp8_dense_fp8_attn-kv_fp8 (place under `$MODELS_DIR/Kaon-V3-nvpf4-mixed/`)
- MTP draft: `google/gemma-4-26B-A4B-it-assistant` → `$MODELS_DIR/gemma-4-26B-A4B-it-assistant/`
- Serve: `modelopt` + `kv-cache-dtype fp8` + `moe-backend cutlass` + MTP `num_speculative_tokens=3`. **Do not swap the FlashInfer wheel** — built-in `0.6.15.post1` is the crashing config.
- Load: **synthetic multi-turn chat** embedded in the script (stdlib only; no private replay logs). On our RTX 5090 this reproduced within a few minutes of sustained ~QPS 5.7 load after readiness; p50 climbed to ~12s before the crash (host plan-build path).

Crash signatures (any one):

```
cudaErrorMisalignedAddress / misaligned address
CUDNN_STATUS_EXECUTION_FAILED_CUDA_DRIVER
illegal memory access was encountered (700)
```

### How to run

```bash
# MODELS_DIR/
#   Kaon-V3-nvpf4-mixed/
#   gemma-4-26B-A4B-it-assistant/
export MODELS_DIR=/path/to/models
chmod +x repro_sm120_cudnn_crash.sh
./repro_sm120_cudnn_crash.sh --rounds 12 --port 8001
# exit 0 = crash signature reproduced
```

Needs ≥~30 GiB free GPU memory. Script aborts early if cuDNN backend ≥ 92301 (wrong image).

### Full script

<details>
<summary>repro_sm120_cudnn_crash.sh (click to expand)</summary>

````bash
#!/usr/bin/env bash
# Reproduce SM120 bmm_fp8(backend="auto") cuDNN hard failure (FlashInfer PR #4165)
# ----------------------------------------------------------------------------
# Requires: RTX 5090 (SM120), >=~30 GiB free GPU, Docker --gpus all
# Models under MODELS_DIR:
#     https://huggingface.co/wangqia0309/gemma-4-26B-A4B-it_nvfp4_experts_fp8_dense_fp8_attn-kv_fp8
#   gemma-4-26B-A4B-it-assistant/     # MTP draft (e.g. google/gemma-4-26B-A4B-it-assistant)
#
# Image is DIGEST-PINNED (do NOT use floating :nightly):
#   vllm/vllm-openai@sha256:b356b39e510376c56ae16f530cc6f2d0e10934631a009e30ba9c13cc02ae7ffe
#   cuDNN backend 92000 (<92301 => override_shape unavailable), FI 0.6.15.post1
#
# Why isolated GEMM will not reproduce: the fault is in serving-time cuDNN
# execute_plan_at_index under MTP + dynamic shapes; CUDA may surface it async.
# Verified with THIS script's embedded synthetic multi-turn load (no private dataset):
#   cudaErrorMisalignedAddress on RTX 5090 within ~few minutes of load.
#
# Usage:
#   export MODELS_DIR=/path/to/models
#   chmod +x repro_sm120_cudnn_crash.sh && ./repro_sm120_cudnn_crash.sh [--rounds 12] [--port 8001]
# Exit 0 = crash signature reproduced.
# ----------------------------------------------------------------------------
set -euo pipefail

IMAGE="vllm/vllm-openai@sha256:b356b39e510376c56ae16f530cc6f2d0e10934631a009e30ba9c13cc02ae7ffe"
ROOT="${ROOT:-$(pwd)}"
MODELS_DIR="${MODELS_DIR:-$ROOT/models}"
TAG="repro-sm120-crash"
PORT=8001
ROUNDS=12
ROUND_SECONDS=30
TARGET_QPS=5.7
MAX_INFLIGHT=48
CRASH_RE='cudaErrorMisalignedAddress|misaligned address|CUDNN_STATUS_EXECUTION_FAILED|illegal memory access'

while [[ $# -gt 0 ]]; do
  case "$1" in
    --rounds)  ROUNDS="$2"; shift 2;;
    --port)    PORT="$2"; shift 2;;
    --seconds) ROUND_SECONDS="$2"; shift 2;;
    --qps)     TARGET_QPS="$2"; shift 2;;
    --models)  MODELS_DIR="$2"; shift 2;;
    --root)    ROOT="$2"; shift 2;;
    *) echo "unknown arg: $1" >&2; exit 2;;
  esac
done

TOTAL_SECONDS=$((ROUNDS * ROUND_SECONDS))
NAME="fi-$TAG"
LOG="$ROOT/logs/$TAG.serve.log"
LOAD_LOG="$ROOT/logs/$TAG.load.log"
CRASH_MARKER="$ROOT/logs/$TAG.crash.txt"
LOADGEN="$ROOT/logs/$TAG.loadgen.py"

mkdir -p "$ROOT/logs" "$ROOT/results"
rm -f "$LOG" "$LOAD_LOG" "$CRASH_MARKER" "$LOADGEN"
docker rm -f "$NAME" >/dev/null 2>&1 || true

if [[ ! -f "$MODELS_DIR/Kaon-V3-nvpf4-mixed/config.json" ]]; then
  echo "[repro] missing $MODELS_DIR/Kaon-V3-nvpf4-mixed" >&2; exit 1
fi
if [[ ! -f "$MODELS_DIR/gemma-4-26B-A4B-it-assistant/config.json" ]]; then
  echo "[repro] missing $MODELS_DIR/gemma-4-26B-A4B-it-assistant" >&2; exit 1
fi

free_mib=$(nvidia-smi --query-gpu=memory.free --format=csv,noheader,nounits | head -1 | tr -d ' ')
if (( free_mib < 30000 )); then
  echo "[repro] GPU free ${free_mib} MiB < 30000" >&2
  nvidia-smi --query-compute-apps=pid,process_name,used_memory --format=csv >&2 || true
  exit 1
fi

echo "[repro] image=$IMAGE"
echo "[repro] models=$MODELS_DIR port=$PORT qps=$TARGET_QPS duration=${TOTAL_SECONDS}s"

if ! docker image inspect "$IMAGE" >/dev/null 2>&1; then
  echo "[repro] pulling pinned image ..."
  docker pull "$IMAGE"
fi

GUARD_OUT="$(docker run --rm --gpus all --entrypoint python3 "$IMAGE" -c \
  'import cudnn,flashinfer; print(cudnn.backend_version(), flashinfer.__version__)' 2>&1)" \
  || { echo "[repro] version guard failed: $GUARD_OUT" >&2; exit 1; }
CUDNN_BE="$(echo "$GUARD_OUT" | awk '{print $1}')"
echo "[repro] cuDNN backend=$CUDNN_BE FI=$(echo "$GUARD_OUT" | awk '{print $2}')"
if [[ -z "$CUDNN_BE" || "$CUDNN_BE" -ge 92301 ]]; then
  echo "[repro] ABORT: backend=$CUDNN_BE >= 92301 (override_shape available) — will NOT crash. Use pinned digest." >&2
  exit 1
fi

KAON_REAL="$(readlink -f "$MODELS_DIR/Kaon-V3-nvpf4-mixed")"
ASST_REAL="$(readlink -f "$MODELS_DIR/gemma-4-26B-A4B-it-assistant")"
needs_hub_mount=0
for p in "$KAON_REAL" "$ASST_REAL"; do
  if find "$p" -maxdepth 1 -type l -lname '../../blobs/*' | grep -q .; then
    needs_hub_mount=1; break
  fi
done

MOUNT_ARGS=()
if (( needs_hub_mount )); then
  HF_HOME_HOST="$(readlink -f "${HF_HOME:-$HOME/.cache/huggingface}")"
  case "$KAON_REAL" in "$HF_HOME_HOST"/*) ;; *)
    echo "[repro] hub symlinks outside HF_HOME=$HF_HOME_HOST" >&2; exit 1;; esac
  case "$ASST_REAL" in "$HF_HOME_HOST"/*) ;; *)
    echo "[repro] draft hub symlinks outside HF_HOME=$HF_HOME_HOST" >&2; exit 1;; esac
  MOUNT_ARGS+=(-v "$HF_HOME_HOST:$HF_HOME_HOST:ro")
  MODEL_IN_CTR="$KAON_REAL"; DRAFT_IN_CTR="$ASST_REAL"
else
  MOUNT_ARGS+=(-v "$KAON_REAL:/models/Kaon-V3-nvpf4-mixed:ro")
  MOUNT_ARGS+=(-v "$ASST_REAL:/models/gemma-4-26B-A4B-it-assistant:ro")
  MODEL_IN_CTR="/models/Kaon-V3-nvpf4-mixed"
  DRAFT_IN_CTR="/models/gemma-4-26B-A4B-it-assistant"
fi

echo "[repro] starting $NAME ..."
docker run -d --name "$NAME" --gpus all --network host --entrypoint bash \
  -e FLASHINFER_DISABLE_VERSION_CHECK=1 -e FLASHINFER_LOGLEVEL=1 \
  "${MOUNT_ARGS[@]}" \
  "$IMAGE" -c '
    set -e
    rm -rf /root/.cache/vllm/flashinfer_autotune_cache || true
    python3 -c "import cudnn,flashinfer; print(\"CUDNN_BACKEND=\",cudnn.backend_version(),\"FI=\",flashinfer.__version__)"
    exec python3 -m vllm.entrypoints.openai.api_server \
      --model '"$MODEL_IN_CTR"' --host 0.0.0.0 --port '"$PORT"' \
      --served-model-name kaon-v3-nvpf4-mixed --trust-remote-code \
      --max-model-len 8192 --max-num-seqs 64 --max-num-batched-tokens 8192 \
      --gpu-memory-utilization 0.95 --language-model-only \
      --quantization modelopt --kv-cache-dtype fp8 --moe-backend cutlass \
      --speculative-config "{\"method\":\"mtp\",\"model\":\"'"$DRAFT_IN_CTR"'\",\"num_speculative_tokens\":3}" \
      --stream-interval 5 -O3
  ' >/dev/null

docker logs -f "$NAME" >"$LOG" 2>&1 &
LOG_PID=$!
WATCH_PID=""; LOAD_PID=""
cleanup() {
  kill "$LOG_PID" >/dev/null 2>&1 || true
  [[ -n "${WATCH_PID}" ]] && kill "$WATCH_PID" >/dev/null 2>&1 || true
  [[ -n "${LOAD_PID}" ]] && kill "$LOAD_PID" >/dev/null 2>&1 || true
}
trap cleanup EXIT

echo "[repro] waiting for readiness (up to 40 min) ..."
deadline=$((SECONDS + 2400))
while (( SECONDS < deadline )); do
  if curl --fail --silent "http://localhost:$PORT/v1/models" >/dev/null 2>&1; then
    echo "[repro] ready on :$PORT"; break
  fi
  if [[ "$(docker inspect --format '{{.State.Running}}' "$NAME")" != "true" ]]; then
    echo "[repro] container exited early:"; tail -n 40 "$LOG" >&2; exit 1
  fi
  sleep 5
done
if ! curl --fail --silent "http://localhost:$PORT/v1/models" >/dev/null 2>&1; then
  echo "[repro] readiness timed out" >&2; exit 1
fi

( while true; do
    if grep -qiE "$CRASH_RE" "$LOG" 2>/dev/null; then
      echo "[repro] CRASH signature detected"
      grep -niE "$CRASH_RE" "$LOG" | head -n 5 > "$CRASH_MARKER"
      break
    fi
    sleep 5
  done ) &
WATCH_PID=$!

cat >"$LOADGEN" <<'LOADGEN_EOF'
"""Synthetic chat load for SM120 cuDNN misaligned-address repro. Stdlib only.

No private replay dataset. Builds multi-turn conversations with growing
context + concurrent QPS so MTP + continuous batching produce new (M,K,N)
shapes and force cuDNN policy=ALL plan builds (crash path on SM120 without
override_shape).
"""

import argparse
import json
import random
import string
import threading
import time
import urllib.request
from concurrent.futures import FIRST_COMPLETED, ThreadPoolExecutor, wait

# Mix short and long turns so batched token counts wander (incl. MTP shapes).
TURN_CHARS = (128, 256, 512, 1024, 1536, 2048, 3072, 4096)
PAD = ("alpha bravo charlie delta echo foxtrot golf hotel india juliet " * 20)


def _pad(n: int) -> str:
    body = (PAD * ((n // len(PAD)) + 2))[:n]
    uniq = "".join(random.choices(string.ascii_lowercase, k=12))
    return f"{body} [id={uniq}]"


class ConversationPool:
    """In-memory multi-turn chats — closer to production replay than one-shot prompts."""

    def __init__(self, n: int = 64, max_turns: int = 8) -> None:
        self.max_turns = max_turns
        self.lock = threading.Lock()
        self.convs: list[list[dict]] = [[] for _ in range(n)]

    def next_messages(self) -> list[dict]:
        with self.lock:
            i = random.randrange(len(self.convs))
            hist = self.convs[i]
            if len(hist) >= self.max_turns * 2:
                hist.clear()
            n = random.choice(TURN_CHARS)
            if random.random() < 0.2:
                n = random.randint(48, 160)
            hist.append({"role": "user", "content": _pad(n)})
            # Copy so the worker can send without holding the lock.
            return list(hist)

    def record_assistant(self, messages: list[dict], content: str) -> None:
        # Best-effort: append assistant turn onto matching prefix length bucket.
        with self.lock:
            for hist in self.convs:
                if len(hist) == len(messages) and hist and hist[-1] is messages[-1]:
                    hist.append({"role": "assistant", "content": (content or "")[:500]})
                    return
            # Fallback: attach to a random open convo
            i = random.randrange(len(self.convs))
            if self.convs[i] and self.convs[i][-1]["role"] == "user":
                self.convs[i].append({"role": "assistant", "content": (content or "")[:500]})


POOL = ConversationPool()


def one_request(url: str, model: str, timeout: float) -> tuple[str, float, str]:
    messages = POOL.next_messages()
    payload = {
        "model": model,
        "messages": messages,
        "max_tokens": random.choice([64, 128, 200, 256]),
        "temperature": 0.7,
        "top_p": 0.85,
        "stream": False,
    }
    data = json.dumps(payload).encode()
    req = urllib.request.Request(
        url,
        data=data,
        headers={
            "Content-Type": "application/json",
            "Authorization": "Bearer " + ("a" * 32),
        },
        method="POST",
    )
    t0 = time.perf_counter()
    try:
        with urllib.request.urlopen(req, timeout=timeout) as resp:
            body = json.loads(resp.read())
            text = ""
            try:
                text = body["choices"][0]["message"]["content"]
            except Exception:
                pass
            POOL.record_assistant(messages, text)
            return "ok", time.perf_counter() - t0, ""
    except Exception as e:  # noqa: BLE001 - keep hammering through transient errors
        return "err", time.perf_counter() - t0, f"{type(e).__name__}: {e}"


def main() -> int:
    ap = argparse.ArgumentParser()
    ap.add_argument("--api-base", required=True)
    ap.add_argument("--model", default="kaon-v3-nvpf4-mixed")
    ap.add_argument("--qps", type=float, default=5.7)
    ap.add_argument("--seconds", type=float, required=True)
    ap.add_argument("--max-inflight", type=int, default=48)
    ap.add_argument("--timeout", type=float, default=600.0)
    args = ap.parse_args()

    url = args.api_base.rstrip("/") + "/chat/completions"
    interval = 1.0 / args.qps
    deadline = time.perf_counter() + args.seconds
    ok = err = 0
    latencies: list[float] = []
    lock = threading.Lock()

    def record(status: str, dt: float, msg: str) -> None:
        nonlocal ok, err
        with lock:
            if status == "ok":
                ok += 1
                latencies.append(dt)
            else:
                err += 1
                if err <= 8 or err % 25 == 0:
                    print(f"[load] err#{err} {msg}", flush=True)

    print(
        f"[load] start url={url} qps={args.qps} seconds={args.seconds} "
        f"max_inflight={args.max_inflight} (multi-turn synthetic, no dataset)",
        flush=True,
    )
    next_launch = time.perf_counter()
    with ThreadPoolExecutor(max_workers=args.max_inflight) as pool:
        pending: set = set()
        while time.perf_counter() < deadline:
            now = time.perf_counter()
            if now < next_launch:
                time.sleep(min(0.01, next_launch - now))
                continue
            next_launch = max(next_launch + interval, time.perf_counter())
            while len(pending) >= args.max_inflight:
                done, pending = wait(pending, return_when=FIRST_COMPLETED, timeout=1.0)
                for fut in done:
                    record(*fut.result())
            pending.add(pool.submit(one_request, url, args.model, args.timeout))
            if (ok + err) > 0 and (ok + err) % 20 == 0:
                p50 = sorted(latencies)[len(latencies) // 2] if latencies else 0.0
                print(
                    f"[load] sent={ok + err} ok={ok} err={err} "
                    f"inflight={len(pending)} p50_lat={p50:.2f}s",
                    flush=True,
                )
        while pending:
            done, pending = wait(pending, return_when=FIRST_COMPLETED, timeout=5.0)
            for fut in done:
                record(*fut.result())

    p50 = sorted(latencies)[len(latencies) // 2] if latencies else 0.0
    print(f"[load] done ok={ok} err={err} p50_lat={p50:.2f}s", flush=True)
    return 0 if ok > 0 else 1


if __name__ == "__main__":
    raise SystemExit(main())

LOADGEN_EOF

echo "[repro] synthetic multi-turn load ${TOTAL_SECONDS}s @ QPS ${TARGET_QPS} (no private dataset) ..."
set +e
python3 "$LOADGEN" \
  --api-base "http://localhost:$PORT/v1" \
  --model kaon-v3-nvpf4-mixed \
  --qps "$TARGET_QPS" --seconds "$TOTAL_SECONDS" \
  --max-inflight "$MAX_INFLIGHT" --timeout 600 \
  2>&1 | tee "$LOAD_LOG" &
LOAD_PID=$!

while kill -0 "$LOAD_PID" 2>/dev/null; do
  if [[ -s "$CRASH_MARKER" ]]; then
    echo "[repro] crash seen — stopping loadgen"
    kill "$LOAD_PID" >/dev/null 2>&1 || true
    break
  fi
  if [[ "$(docker inspect --format '{{.State.Running}}' "$NAME" 2>/dev/null || echo false)" != "true" ]]; then
    echo "[repro] container stopped"
    kill "$LOAD_PID" >/dev/null 2>&1 || true
    break
  fi
  sleep 5
done
wait "$LOAD_PID" 2>/dev/null || true
kill "$WATCH_PID" >/dev/null 2>&1 || true
sleep 2
set -e

echo "==== crash signature ===="
if [[ -s "$CRASH_MARKER" ]]; then
  cat "$CRASH_MARKER"; echo
  echo "[repro] RESULT: crash signature REPRODUCED"; exit 0
fi
if grep -qiE "$CRASH_RE" "$LOG"; then
  grep -niE "$CRASH_RE" "$LOG" | head -n 5
  echo "[repro] RESULT: crash signature REPRODUCED (post-scan)"; exit 0
fi
echo "[repro] RESULT: no crash within ${TOTAL_SECONDS}s"; exit 1

````

</details>

Happy to add a `CUDA_LAUNCH_BLOCKING=1` variant if you want the fault forced to the launch site, or tweak the loadgen if anything is unclear on your side.

yanqinz2:
Discussed with @YangXu1990uiuc , seems we don't have a short-term solution for this host overhead issue, unless the cudnn version bumped to v9.23.1. We agree to merge this PR at this moment. Please feel free to approve if it lgtu. @nv-yunzheq

Saddss:
@nv-yunzheq hello，could you please take a look?

bkryu (approved):
Thank you for your patience @Saddss . PR looks good. Merging

sybrengg:
Confirming this from a different direction, and adding one consequence that I don't think is captured here yet: on SM12x without `override_shape`, the per-shape `policy=ALL` rebuild isn't only a host-side stall — **it also leaks device memory**.

Each graph build loads a CUDA module that is never released. Measured on an SM120 (RTX PRO 6000) serving deployment, cuDNN backend `91701`, FlashInfer 0.6.14, continuous batching so M changes almost every step:

- CUPTI `MODULE_LOADED`: **~150/min, no plateau**
- CUPTI `MODULE_UNLOAD_STARTING`: **0**, for the entire process lifetime
- `nvidia-smi` on the engine pid: **+~10 MiB/min**, unbounded

I found it via a CUPTI injection library (`CUDA_INJECTION64_PATH`) rather than an allocator hook, because it is invisible to both the torch caching allocator and `cuMemAlloc`/`cudaMalloc` interposition — module loading goes through neither. The tell in the activity records was a flood of unfreed `CUPTI_ACTIVITY_MEMORY_KIND_DEVICE_STATIC` allocations, i.e. static `__device__` storage, which only arrives with a module load.

Routing that layer off cuDNN made both signals flat: module loads plateaued and device memory stopped growing entirely over a 12 h run. The mechanism you describe here — one graph build per distinct serving shape, unbounded without `cache_m` bucketing — accounts for the observed rate almost exactly.

So this fix likely resolves a class of "GPU memory slowly grows until OOM" reports on SM12x, not just the latency spikes. Worth mentioning in the symptom description if anyone goes looking.

I've opened #4426 applying the same condition to `fp8_gemm_sm100`, which callers can reach with a pre-built runner list without passing through `_heuristic_func_bmm_fp8`.

Issue #4160, opened 2026-07-27
[Bug]`layernorm_quant` FP8 e4m3 correctness failure on A10/SM86 with small scale


cindyzxq:
## Description

  `flashinfer.norm.layernorm_quant` added by #3962 appears to fail correctness on A10/SM86 in the public nightly pipeline. Observed in https://github.com/flashinfer-ai/flashinfer/actions/runs/30186569236 
  ```bash
  pytest -s tests/utils/test_norm.py

  Failing cases:

  tests/utils/test_norm.py::test_layernorm_quant[quant_dtype0-0.01-500-1]
  tests/utils/test_norm.py::test_layernorm_quant[quant_dtype0-0.01-1024-1]

  Parameters:

  batch_size = 1
  hidden_size = 500 / 1024
  quant_scale = 0.01
  quant_dtype = torch.float8_e4m3fn

  Failure:

  assert default_check([y_ref], [y], max_mismatch_pct=1.0, min_cos_sim=None)
```
  This means more than 1% of elements are not close under the default FP8 tolerance (rtol=0.1, atol=0.1).

  ## Why this looks SM86/A10 related

  layernorm_quant does not currently gate out SM86, so A10 appears to be an intended supported target.

  The suspicious path is the packed e4m3 conversion used by even hidden sizes. In include/flashinfer/trtllm/common/cudaTypeUtils.cuh, the float2 -> __nv_fp8x2_e4m3 conversion currently rounds through bf16 before converting to FP8:
```
  template <>
  __device__ inline __nv_fp8x2_e4m3 cuda_cast<__nv_fp8x2_e4m3, float2>(float2 val) {
    return __nv_fp8x2_e4m3(bf1622float2(float22bf162(val)));
  }
```
  The Python reference in tests/utils/test_norm.py only models the earlier bf16 rounding before applying scale:
```
  y = y.to(x.dtype).float() / scale
```
  With scale=0.01, small rounding differences are amplified by 100x and can cross FP8 e4m3 bucket boundaries. This likely explains why the failure shows up for float8_e4m3fn, small scale, and even hidden sizes (500, 1024) on A10/SM86.

  ## Suggested fix (By Codex)

  Change the packed e4m3 cast to convert directly from float2 to FP8, matching the implementation style already used in include/flashinfer/vec_dtypes.cuh:
```
  template <>
[...]

PR #4167, opened 2026-07-27
fix(norm): convert float2 to e4m3 directly in packed cast

elwhyjay:
## 📌 Description
`layernorm_quant` can miss the FP8 e4m3 correctness bound when the quantization scale is small. Reported for A10/SM86 in the nightly pipeline.

The packed `float2` to `__nv_fp8x2_e4m3` cast in `cudaTypeUtils.cuh` rounded through bf16 before quantizing:

    return __nv_fp8x2_e4m3(bf1622float2(float22bf162(val)));

`generalLayerNorm` applies the per-tensor scale before this cast, so the mantissa bits dropped by the bf16 step are scaled up by `1/scale`. At `scale=0.01` that is a 100x amplification, enough to push values into a neighbouring e4m3 bucket. The test reference models the bf16 rounding that the kernel does before scaling, but not this second rounding after it, so the two diverge.

This PR converts directly with `__nv_cvt_float2_to_fp8x2`, matching the style already used in `vec_dtypes.cuh`. The e5m2 packed cast already converts without a bf16 round trip, so it is unchanged, which matches the failures being e4m3 only. The cast is only reachable through `generalLayerNorm`, whose sole caller is `LayerNormQuant`, so `layernorm_quant` is the only affected API.

This PR also seeds `test_layernorm_quant`. The bf16 round trip is architecture independent, so the error was always present on the packed path; the check bounds the fraction of drifting elements, and unseeded inputs left that fraction varying per run, which is why it surfaced as an intermittent SM86 failure.

## 🔍 Related Issues
#4160
## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

Verified on an RTX A6000 (SM86, same architecture as the A10 in the nightly pipeline).

Before and after, over 30 seeds per shape at `scale=0.01`, e4m3, batch 1:

| hidden | seeds over 1% bound (before) | after | worst mismatch (before) | after |
|-------:|-----------------------------:|------:|------------------------:|------:|
|    500 |                         4/30 |  0/30 |                   1.20% | 0.00% |
|   1024 |                         3/30 |  0/30 |                   1.27% | 0.00% |

The two cases named in the report, `hidden=500` and `hidden=1024` at `scale=0.01`, pass after the fix.

- Full `tests/utils/test_norm.py`: 1543 passed, 1344 skipped. Skips are PDL cases that need Hopper or newer.
- `pytest tests/trace/ -k "norm"`: 123 passed, 21 skipped.
- `pre-commit run` on the changed files: all hooks pass.

Unrelated pre-existing failures: `pytest tests/trace/ -k "rope"` reports 24 failures on this machine both with and without this change, so they are not caused by it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved FP8 conversion accuracy and handling by converting values directly with saturation.

* **Tests**
  * Stabilized quantized layer normalization test results by making input generation deterministic.

### DIFF
--- a/include/flashinfer/trtllm/common/cudaTypeUtils.cuh
+++ b/include/flashinfer/trtllm/common/cudaTypeUtils.cuh
@@ -592,7 +592,11 @@ __device__ inline half2 cuda_cast<half2, __nv_fp8x2_e4m3>(__nv_fp8x2_e4m3 val) {
 
 template <>
 __device__ inline __nv_fp8x2_e4m3 cuda_cast<__nv_fp8x2_e4m3, float2>(float2 val) {
-  return __nv_fp8x2_e4m3(bf1622float2(float22bf162(val)));
+  // Direct conversion: an intermediate bf16 step drops mantissa bits that a
+  // small quantization scale then amplifies past e4m3 bucket boundaries.
+  __nv_fp8x2_e4m3 out;
+  out.__x = __nv_cvt_float2_to_fp8x2(val, __NV_SATFINITE, __NV_E4M3);
+  return out;
 }
 
 template <>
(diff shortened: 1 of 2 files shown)


kahyunnam (approved):
Thanks for the follow up fix @elwhyjay ! Approving, I can help merge after internal CI run clears.

bkryu (approved):
Thank you @elwhyjay, the internal CI looks good to me

Issue #4172, opened 2026-07-27
Bug: trtllm_fp4_block_scale_moe kernel hangs under interprocess CUDA events → TP deadlock (B200)


ZKeeer:
## Summary

When serving GLM-5.2-NVFP4 (`quantization=modelopt_fp4`) with vLLM + LMCache (`LMCacheMPConnector`, TP=8/EP=8 on 8×B200), the flashinfer `trtllm_fp4_block_scale_moe` GPU kernel hangs sporadically, causing a tensor-parallel collective deadlock. All 8 workers spin-wait indefinitely until vLLM's 300s `execute_model` timeout fires.

## Key findings

1. **FP8 works fine**: Same model architecture with `quantization=fp8` (same `FLASHINFER_TRTLLM` MoE backend, which dispatches to `trtllm_fp8_block_scale_moe`) completes stress tests with zero errors.

2. **Without LMCache works**: The baseline (vLLM's built-in OffloadingConnector, no LMCache) runs the same NVFP4 model stably for 3 days.

3. **Root cause**: LMCache creates `Event(interprocess=True)` + `record()` on the forward stream. This triggers PDL (Programmatic Dependent Launch) sensitivity specific to the FP4 cubin binary.

4. **PDL is active on B200**: The `trtllm::gen::launchKernel()` uses `cuLaunchKernelEx` with `CU_LAUNCH_ATTRIBUTE_PROGRAMMATIC_STREAM_SERIALIZATION`. `device_support_pdl()` returns True for compute capability >= 9.

## Hypothesis

The FP4 cubin has a grid-dependency synchronization path that is sensitive to CUDA driver state perturbation from interprocess CUDA events. The FP8 cubin (same launch path, different binary) does not exhibit this.

## Suggested fix

Disable PDL by default for the FP4 MoE kernel (`trtllm_fp4_block_scale_moe_op`), or provide an env-var override (`FLASHINFER_DISABLE_PDL_MOE=1`) for users to test/workaround.

## Environment

| Component | Version |
|-----------|---------|
| vLLM | 0.25.1 |
| flashinfer | 0.6.13 |
| flashinfer-cubin | 0.6.13 |
| Model | GLM-5.2-NVFP4 (`modelopt_fp4`) |
| Hardware | 8× NVIDIA B200 (CUDA 13.0) |
| Parallelism | TP=8, EP=8 |

Full diagnostic artifacts available upon request.

PR #4173, opened 2026-07-27
fix: disable PDL for FP4 MoE kernel to prevent TP deadlock (#4172)

ZKeeer:
## Summary

Fix for #4172: The `trtllm_fp4_block_scale_moe` kernel hangs under interprocess CUDA event interference when PDL is enabled on SM100+ GPUs.

## Changes

1. **Disable PDL by default for `trtllm_fp4_block_scale_moe_op`** — PDL (`CU_LAUNCH_ATTRIBUTE_PROGRAMMATIC_STREAM_SERIALIZATION`) can cause kernel hangs when interprocess CUDA events are recorded on the same stream. The FP8 counterpart (`trtllm_fp8_block_scale_moe`) is not affected.

2. **Add `FLASHINFER_DISABLE_PDL_MOE=1` env var override** — users can test whether PDL is the root cause, or re-enable PDL if their workload is not affected.

## Testing

- Verified on 8×B200 with GLM-5.2-NVFP4 (TP=8, EP=8), LMCache connector
- FP8 baseline unaffected (0 errors under ABA2 stress test)
- Without the fix: 3/3 NVFP4 runs deadlock within 5–33 minutes

## Rationale

PDL is a performance optimization (kernel overlap), not a correctness requirement. Disabling it for FP4 MoE eliminates the sensitivity at minimal perf cost.

## Summary by CodeRabbit

- **Configuration**
  - Added an environment variable to disable Programmatic Dependent Launch (PDL) for FP4 block-scaled MoE execution.
  - FP4 block-scaled MoE execution now disables this optimization by default unless explicitly enabled.
  - Set `FLASHINFER_DISABLE_PDL_MOE=1` to force PDL off at runtime.

### DIFF
--- a/flashinfer/fused_moe/core.py
+++ b/flashinfer/fused_moe/core.py
@@ -16,6 +16,7 @@ limitations under the License.
 
 import functools
 import math
+import os
 from dataclasses import dataclass
 from types import SimpleNamespace
 from typing import Any, Dict, List, Optional, Tuple, Union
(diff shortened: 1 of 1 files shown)


ZKeeer:
Closing — this was my own addition not based on a patch file. Keeping only issue #4172 for the bug report.

PR #4182, opened 2026-07-27
perf(kda): optimize indexed recurrent decode dispatch and paging

djmmoss:
## Summary

- Page indexed recurrent state directly inside both CuTe DSL kernel architectures,
  removing the full-state gather and scatter around standard decode. Dense calls
  reuse the same path through a cached identity mapping.
- Add a table-free SM100/SM103 dispatch heuristic that selects the one-warp or grouped
  architecture and its row/reduction or key/value split schedule from the workload.
- Cover arbitrary state slots, padded state strides, CUDA Graph replay, and dispatch
  boundaries.

## B200 performance

BF16 cold-L2 CUPTI timing with CUDA Graphs disabled; each row is the median of five
independent rounds. This compares upstream pre-PR `recurrent_kda` at `76c583655`
against this PR at `6039adef`. Both columns use the CuTe DSL backend.

| Matrix | Dimensions | Cases | Pre-PR geomean (us) | This PR geomean (us) | Speedup |
|---|---|---:|---:|---:|---:|
| Cake PR #4279 | D128; T=1–6; H=16; HV=16/32; N=1–128 | 30 | 31.281 | 29.875 | 1.047x |
| Kimi K3 PR #4182 | D128; T=1; H=HV=12/24/48/96; N=1–256 | 17 | 182.230 | 13.526 | 13.472x |
| Legacy PR #2572 | D64/128; T=1/4; H=HV=8/16/32; N=1–128 | 48 | 24.981 | 22.192 | 1.126x |
| **Overall** | **All matrices** | **95** | **38.272** | **22.310** | **1.715x** |

The current implementation passed output and full-state correctness across all 95
shapes.


flashinfer/kda_kernels/recurrent_kda.py
```diff
@@ -1754,30 +1829,47 @@ def run_recurrent_kda(
...
+                    f"ssm_state_indices must have shape [{B}] for standard decode, "
+                    f"got {list(ssm_state_indices.shape)}"
+                )
+            ssi = ssm_state_indices.to(torch.int32).contiguous()
         if initial_state is None:
-            state = torch.zeros(B, HV, V, K, device=device, dtype=torch.bfloat16)
-        elif ssm_state_indices is not None:
-            state = initial_state[ssm_state_indices].contiguous()
-            copy_back_indices = ssm_state_indices
+            max_idx = int(ssi.max().item()) + 1
```
jimmyzho, line 1864: This will leave a D2H sync everytime .item() is called + can break cuda graph capture. You can add
 ```
if ssm_state_indices is None: 
    state = torch.zeros(B, HV, V, K, device=device, dtype=torch.bfloat16)
else:
    max_idx = ...
```

File: flashinfer/kda_kernels/recurrent_kda.py [hunk 2]
```diff
@@ -63,9 +63,32 @@
...
+    num_tokens: int,
+    sequence_heads: int,
+    query_heads: int,
+    value_heads: int,
+    use_gate: bool,
+    compute_capability: tuple[int, int],
+) -> bool:
+    """Select the measured kernel architecture for the active GPU."""
+    if compute_capability not in TUNED_DISPATCH_COMPUTE_CAPABILITIES:
+        return num_tokens == 1 and sequence_heads < GROUPED_MIN_SEQUENCE_HEADS[head_dim]
```
jimmyzho, line 82: I believe with this change it is now:
sequence_heads (D=128, T=1) | pre-PR | PR fallback
-- | -- | --
< 128 | grouped | one-warp
128 … 1919 | one-warp | one-warp
`>=` 1920 | one-warp | grouped

Is this expected?

PR #4183, opened 2026-07-27
feat(moe_ep): fault-tolerance rank mask (NCCL-EP + NIXL-EP)

Anerudhan:
## Why

NCCL-EP gained fault tolerance ([`nccl_ep.h` `enable_mask` + `ncclEpMask*`](https://github.com/NVIDIA/nccl/blob/master/contrib/nccl_ep/include/nccl_ep.h#L227-L232)): a peer that times out during dispatch/combine is masked and skipped instead of tripping a GPU `trap()`. NIXL-EP has the equivalent (`update/query/clean_mask_buffer`). **`flashinfer.moe_ep` exposed neither** — it never set `enable_mask`, never called a mask API, and gave callers no way to learn a rank had died. One slow or dead EP rank killed the whole job.

This adds the FT surface over **both** transports. `moe_ep` is pure Python, so there is no CUDA/C++ change here.

---

## How the FT API is used

FT is opt-in per Fleet via a **knob**, not new `FleetParams` fields — `FleetParams` is the frozen *sizing* dataclass, while optional transport features already live in the knob namespace (cf. `FleetAlgoKnobTopologyCapacity`). The runtime API lands on the `Fleet` ABC as **concrete raising defaults**, not `@abstractmethod`, so no existing or out-of-tree Fleet breaks.

### 1. Probe, then enable

```python
from flashinfer.moe_ep import (
    FleetAlgoKnobFaultTolerance, FleetAlgoKnobTopologyCapacity,
    MoEEpLayer, supports_fault_tolerance,
)

# Needs more than the backend being built: nccl_ep also needs an nccl4py with
# GroupConfig.enable_mask AND a libnccl_ep exporting ncclEpMask*. Never raises.
assert supports_fault_tolerance("nccl_ep")

knobs = [FleetAlgoKnobFaultTolerance(timeout_ms=5000)]   # 0 = transport default
# nixl only: size the capacity for the largest world you will ever reach
knobs.append(FleetAlgoKnobTopologyCapacity(n=32))

layer = MoEEpLayer(bootstrap, fleet_params, weights, fleet_knobs=knobs,
                   backend=SplitConfig(comm=NCCLEPConfig(), kernel=IdentityConfig()))
```

`FleetAlgoKnobFaultTolerance(enabled=True, timeout_ms=0, reconcile_timeout_s=30.0, coordinator_takeover_s=10.0)`. **LOW_LATENCY only** on both transports — `validate_fleet_params` rejects FT + HIGH_THROUGHPUT at construction, because nccl leaves the mask buffer NULL under HT and the mask APIs then *abort the process*.

### 2. Serve, poll, recover

```python
fleet = layer._ensure_fleet()

for step in ...:
    layer.forward(t)

    # Between iterations only — both transports read the mask LIVE from the
    # dispatch/combine kernels, so mutating it mid-collective is a race.
    if fleet.query_fault():                       # free on nccl (pinned host flag)
        agreed = fleet.reconcile_active_mask()    # store-collective, death-tolerant
        fleet.clear_faults(readmit=False)         # re-arm detection; keep serving DEGRADED

        # ... later, if the peer comes back (collective over survivors, blocking):
        # fleet.clear_faults(readmit=True)
```

| Method | Collective? | Blocks host? | Stream-ordered? |
|---|---|---|---|
| `supports_fault_tolerance` | — | no | no |
| `query_fault()` | local | nccl no / nixl yes (small D2H) | nccl no / nixl yes |
| `query_active_mask(out=None)` | local | no | **yes** |
| `set_active_mask(mask)` | local (must be applied identically everywhere) | no | **yes** |
| `reconcile_active_mask()` | **store-collective**, tolerates dead ranks | yes (≤ timeout) | yes |
| `clear_faults(readmit=False)` | local | no | no |
| `clear_faults(readmit=True)` | **collective over survivors** | **yes** | yes |
| `active_mask_epoch` | — | no | no |

**Canonical mask: `int32[world_size]`, `1 = active`, CUDA tensor** — matching `ncclEpMaskQuery` and vLLM's `query_active_mask()` naming.

### 3. Rules that will bite you

1. **The steady state is read-only.** The transport discovers the fault and masks the peer; the application's job is to *notice* (`query_fault` / `query_active_mask`). `set_active_mask` is the exceptional reconciliation path, not a per-step call — most callers only reach it via `reconcile_active_mask()`. If you do write the mask, write it only between iterations (kernels read it live).
2. **All survivors must reconcile in the same iteration slot** — they must pass the same `active_mask_epoch`.
3. **`clear_faults(readmit=True)` and `update_topology()` are alternatives, not a sequence.** The former re-admits a merely-delayed rank on the *same* communicator; the latter destroys the group and builds a **new** `ncclComm_t`, which is the only way to add or replace a process. Re-admitting after a rebuild is meaningless; before one it is wasted work. (If you do both, `MaskClean` must come first — it needs a live handle on the current group.)
4. **No FT call during CUDA-graph capture** — but note **dispatch and combine themselves ARE safe to capture**: neither transport compacts the surviving ranks' layout, so they re-read the mask on every replay and a rank that fails later is still skipped. What must not be captured is a *decision about* fault state. `query_fault()` is a host read, not stream work, so it can't be captured at all — it returns the capture-time answer and freezes the branch taken on it into the graph forever. All four FT entry points raise on capture with a per-operation reason. *(Thanks @lrbison — an earlier draft of this said "stale offsets", which was wrong.)*
5. **A rank can be told it is dead.** `reconcile_active_mask()` raises `MoEEpRankEvictedError` when the survivors agreed *this* rank is gone. It can't apply that (a rank may not mask itself) and mustn't keep serving (peers stopped sending it tokens).
6. **Dropped tokens are not renormalized, and experts are not re-homed.** `y_degraded[t] == y_healthy[t] * Σ_{alive k} topk_weights[t][k]`. Both omissions are deliberate: implicit renormalization would add a kernel to every forward, hide a serving-quality event, and divide by zero when a token's whole top-k died; re-homing the dead rank's experts is an EPLB-style job that belongs to the framework. Keep serving on a partial mask with `reconcile_active_mask()` → `clear_faults(readmit=False)` — no `update_topology`, no new communicator. Runbook has the opt-in renormalization snippet.

---

### Group/Buffer creation (where masking is switched on)

```
nccl_ep                                        nixl_ep
-------                                        -------
NcclEpFleet.__init__                           NixlEpFleet.__init__
 ├ _index_knobs -> self._ft                     ├ _index_knobs -> self._ft
 ├ validate_fleet_params(fault_tolerance=)      ├ validate_fleet_params(fault_tolerance=)
 ├ _check_ft_supported()                        ├ nixl_ep.Buffer(..., timeout_ms=)   [TypeError -> actionable]
 │   ├ dataclasses.fields(GroupConfig)          ├ update_memory_buffers(cap, ...)
 │   │   -> needs "enable_mask"                 │   -> allocates mask_buffer[capacity], 0xFF-memset
 │   └ mask_ffi().available                     └ connect_ranks([0, world))
 │       -> needs the 5 ncclEpMask* symbols         -> unmasks [0, world); tail stays masked
 └ _build_group_config()
     ├ kwargs["enable_mask"] = True
     ├ kwargs["timeout_ns"] = timeout_ms * 1e6      (omitted when 0)
     └ nccl.ep.Group.create -> ncclEpCreateGroup
         -> allocates mask_buffer[nRanks] + pinned async-error flag
```

### `query_fault()`

```
nccl_ep                                        nixl_ep
NcclEpFleet.query_fault                        NixlEpFleet.query_fault
 ├ reject_graph_capture("query_fault")          ├ self.query_active_mask()      (kernel + D2H)
 │   ^ NOT because it touches a stream --        │   ^ which itself rejects capture
 │     because it does NOT: a host read          └ compare against self._ft_applied
 │     cannot be captured, so it would
 │     return the capture-time answer and       (no host error flag on this transport,
 │     freeze the branch into the graph          so the applied mask IS the state)
 └ mask_ffi().get_async_error(group)
    ├ group.get_async_error()  [native first]
    └ ctypes ncclEpGetAsyncError(group.ptr, &o)
        -> libnccl_ep.so -> reads PINNED HOST flag: no stream, no sync, free
```

### `query_active_mask()`

```
nccl_ep                                        nixl_ep
 ├ _reject_graph_capture()                      ├ _reject_graph_capture()
 ├ _ft_bufs() -> device int32[world]            ├ _ft_raw() -> device int32[CAPACITY]
 └ mask_ffi().mask_query(group, dev_ptr, str)   ├ buffer.query_mask_buffer(raw)
     -> ncclEpMaskQuery                         │   -> nixl_ep_cpp -> kernel copy
     -> D2D copy of mask_buffer                 │   (asserts numel == max_num_ranks)
     -> ALREADY 1 = active, identity            └ (raw[:world] == 0).to(int32)
                                                    ^ NOT 1-raw: buffer is 0xFF-memset so
                                                      untouched entries read back as -1
```

### `set_active_mask(mask)`

```
nccl_ep                                        nixl_ep
 ├ _normalize_mask -> list[int], rejects        ├ _normalize_mask (same guard)
 │   masking the local rank                     └ for each CHANGED rank r != self:
 ├ host.copy_(...)   PINNED staging buffer          buffer.update_mask_buffer(r, mask=(a==MASKED))
 │   ^ pinned because Update is stream-ordered:      -> atomicExch kernel, one launch EACH
 │     a pageable buffer mutated next call is        -> so we push only the DIFF; a blind
 │     a use-after-write race                          range(world) loop would inject `world`
 └ mask_ffi().mask_update(group, host_ptr, str)        launches into the steady state
     -> ncclEpMaskUpdate  (HOST ptr, unlike Query)
```

### `reconcile_active_mask()` — shared, transport-free

```
FaultToleranceMixin.reconcile_active_mask          (both backends)
 ├ local = self.query_active_mask().cpu().tolist()
 ├ store = self._ft_store()          -> resolve_rendezvous_store(subsystem="ft")
 └ reconcile_masks_via_store(store, rank, world, local, epoch=active_mask_epoch, ...)
     ├ 1. store.set("ft/gen{E}/local/{rank}", bytes(view))
     ├ 2. poll ONLY ranks we still believe alive, until timeout_s
     ├ 3. elementwise-AND what arrived; mask believed-alive ranks that never reported
     ├ 4. coord = min(active); coord publishes via ATOMIC compare_set("ft/gen{E}/decision")
     │      everyone else adopts whatever that key holds  <- kills split brain
     │      (coordinator itself dead -> mask it, re-elect, bounded by world_size)
     └ 5. _adopt(): raises MoEEpRankEvictedError if the decision masks US
 └ self.set_active_mask(agreed)   -> backend-specific path above
```

Deliberately **not** a `torch.distributed` allreduce: that would hang on exactly the rank being masked out. With a store, a missing key *is* the death signal.

### `clear_faults()`

```
nccl_ep                                        nixl_ep
readmit=False:                                 readmit=False:
 └ mask_ffi().error_clear -> ncclEpErrorClear    └ return    (no sticky flag to re-arm)

readmit=True:                                  readmit=True:
 ├ guard: a handle must exist                   ├ buffer.clean_mask_buffer()
 │   (MaskClean asserts on the LL staging       │   -> zeroes ALL `capacity` entries,
 │    buffer -> would SIGABRT from C)           │      marking the never-connected tail ACTIVE
 ├ warn under EXPERT_MAJOR                      ├ _ft_applied = [ACTIVE] * world
 │   (MaskClean computes reset offsets          └ re-mask [world, capacity)   <- fixes that bug
 │    assuming RANK_MAJOR)
 ├ mask_clean -> ncclEpMaskClean
 │   COLLECTIVE over survivors; internally cudaStreamSynchronize's
 └ error_clear -> ncclEpErrorClear
     ^ always paired: MaskClean does NOT clear the flag, which is exactly why
       `readmit` is a flag on one method rather than two a caller can mis-sequence
```

---

## Three transport facts that shaped the code

1. **NIXL's polarity is "nonzero = masked", not "1 = masked".** The buffer is `0xFF`-memset at allocation (an untouched entry reads back as `-1`) and the kernels test `!= 0`. The normalization must be `(raw == 0)`; the obvious `1 - raw` yields `2` for never-connected capacity-tail ranks and silently poisons every downstream `sum()`/`bool()`.
2. **`clean_mask_buffer` zeroes all *capacity* entries**, marking the never-connected tail **active** — a live bug on any fleet sized above its world.
3. **`disconnect_ranks` is suffix-only**, so NIXL cannot evict a *middle* rank; masked-and-degraded is the terminal state there.

Each has a regression test.

## nccl4py binding gap

`nccl4py` binds `GroupConfig.enable_mask`/`timeout_ns` but its `Group` stops at `create`/`create_handle`/`destroy`/`.ptr` — the five mask functions are unreachable from Python. This adds a ctypes shim on `Group.ptr` that **tries a native `Group` method first**, so it retires itself with no call-site churn once those bindings land. Note the symbols live in **`libnccl_ep.so`, not `libnccl.so.2`**; we bind the process-global namespace, which is the only resolution guaranteed to be the same library the caller's group came from.

## Reconciliation, and a hole the tests found

Each transport masks *locally* — NCCL-EP's header calls mask consistency "a framework-level concern" — so survivors can disagree, and disagreeing masks deadlock the next dispatch. The decision is published via a single atomic `compare_set`, which is what prevents a split brain when a straggler's key lands between two survivors' polls; there is a dedicated regression test for that timing.

Writing those tests surfaced a case the design missed: a rank that is alive but that some peer already timed out on gets ANDed out of the group. It can't apply that decision and can't ignore it, so it now raises `MoEEpRankEvictedError`.

## Also fixed (incidental)

- `validate_fleet_params` now rejects a nixl topology capacity **below** world size — it previously sailed through and went out of bounds inside the transport.
- `update_topology` now rejects growing past the capacity, for the same reason. `test_update_topology_diffs_ranks` was exercising that invalid config (4→6 on capacity 4); it now passes `capacity=8`, keeping its intent.

## Testing

`254 passed` in `tests/moe_ep/`; **166 FT-related tests pass**. The 6 failures in the full run are pre-existing and environmental (nvfp4/CuTeDSL ninja JIT) — confirmed failing identically on base `6258e522`.

The protocol, shim and both backends' wiring are covered host-only (HashStore + threads; a fake ctypes library; fake transports), so they run in CI with no GPU.

**Not yet run:** the 4-GPU tiers — `test_moe_ep_fault_tolerance_multirank.py` (stalls a middle rank and walks detect → reconcile → degrade → re-admit, asserting the degraded output *exactly* equals the surviving-weight scaling) and `smoke_ft_ep.py` (hard SIGTERM kill). This host has 1 GPU and no transport built. `bash tests/moe_ep/run_tests.sh ft` runs both.

## vLLM

**No vLLM code here**, by design. vLLM already owns the contract (`support_fault_tolerance` / `query_active_mask` / `query_fault` + the per-step hook in `gpu_model_runner.py`); the FlashInfer-EP managers just hardcode `False`, and since the manager is transport-parameterized, one base-class change covers both LL backends. `docs/design_docs/vllm_moe_ep_integration.md` §8 records the target shape and the four things that commit must get right — including that vLLM's existing nixl/deepep `query_active_mask()` return *raw* buffers with the opposite polarity, and that `self._fleets` holds several fleets sharing one EP group so FT state must be hoisted to a primary fleet.

## Review follow-up (commit 9)

@lrbison's review caught a real defect: `query_fault()` on nccl_ep had **no** capture guard, because I'd reasoned "host read, therefore capture-safe" — exactly backwards. Being a host read is *why* it can't be captured, so it silently returned the capture-time answer and froze the branch on it into the graph. NIXL's already raised (it goes via `query_active_mask`), so the same call behaved differently per backend. Both now raise, with a regression test each. The "stale offsets" rationale was also wrong and is replaced with per-operation reasons.

## Commits

1. `FleetAlgoKnobFaultTolerance` + Fleet FT API surface (no backend touched)
2. promote nixl_ep's store resolver to `core.bootstrap_utils` (pure move)
3. TCPStore-based active-mask reconciliation
4. ctypes shim for the `ncclEpMask*` API
5. wire nccl_ep `enable_mask`/`timeout_ns` + FT methods
6. wire nixl_ep `timeout_ms` + FT methods (carries the polarity/capacity/tail fixes)
7. multirank fault injection + FT smoke + `run_tests.sh ft`
8. docs + vLLM design note
9. `fix`: guard `query_fault` against graph capture + correct the capture rationale (review follow-up)

## Summary by CodeRabbit

* **New Features**
  * Added opt-in fault tolerance for MoE expert-parallel fleets, including timeouts and coordinator takeover.
  * Introduced public APIs to query faults, manage active-rank masks, reconcile fleet state, clear faults, and track mask epochs.
  * Added backend capability detection to ensure fault tolerance only activates when supported.
* **Documentation**
  * Added design docs and an operational runbook describing recovery behavior, ordering rules, and transport-specific constraints.
  * Added a vLLM integration design note for future wiring into the fault-tolerance flow.
* **Tests**
  * Added host-only unit tests, backend-specific mocks, NCCL ctypes-shim tests, multi-GPU end-to-end coverage, and FT smoke tests.

### DIFF
--- a/docs/design_docs/MoE_EP_impl.md
+++ b/docs/design_docs/MoE_EP_impl.md
@@ -18,6 +18,9 @@ lives in dispatch/combine).
 | `BootstrapConfig(world_size, rank, stream, nccl_comm)` | transport bootstrap |
 | `FleetParams(num_experts, max_tokens_per_rank, token_hidden_size, dtype_bytes, algorithm, layout)` | EP geometry |
 | `EpAlgorithm.{LOW_LATENCY, HIGH_THROUGHPUT}`, `EpLayout.{EXPERT_MAJOR, RANK_MAJOR, FLAT}` | algorithm / receive layout |
+| `FleetAlgoKnobFaultTolerance(enabled, timeout_ms, reconcile_timeout_s, coordinator_takeover_s)` | opt-in rank masking (LL only) |
+| `supports_fault_tolerance(backend)` | can this host serve the FT API? |
+| `Fleet.{query_fault, query_active_mask, set_active_mask, reconcile_active_mask, clear_faults, active_mask_epoch}` | FT runtime API — mask is `int32[world]`, `1 = active` |
 | `weights: MoEWeightPack` (layer arg) + `SplitConfig(kernel=FusedMoeKernelConfig(moe_config: MoEConfig))` | the expert GEMM (config from `flashinfer.fused_moe.api`) |
 
 `create_fleet(...)` raises `MoEEpNotBuiltError` (with rebuild hint) if the backend extension
--- a/docs/design_docs/moe_ep_architecture.md
+++ b/docs/design_docs/moe_ep_architecture.md
@@ -46,6 +46,7 @@ Kernels register via `@register_split_kernel` / `@register_mega_kernel` when `ba
 | `MoEWeightPack` | Canonical `w13` / `w2` (+ optional `w13_scale` / `w2_scale`); required `weights` arg at layer construction; `dummy_moe_weights()` for comm-only split |
 | `SplitConfig` | `comm` + `kernel` slots (default `NcclEpConfig` + `IdentityConfig`) |
 | `MegaConfig` | `megakernel`, `quantize_input`, `preprocess_weights`, optional `transformed_weights` |
+| `FleetAlgoKnobFaultTolerance` | Opt-in rank masking (`enabled`, `timeout_ms`, reconcile budgets) — see **Fault tolerance** |
 
 **Split:** pass `SplitConfig(comm=..., kernel=...)` or a comm string/config (kernel defaults to `IdentityConfig`). `fleet_knobs` tune transport. Fleet is lazy-created on first `forward()`; a new Handle per forward. `MoEEpSplitLayer.enable_timing` optionally records per-stage GPU ms in `last_timings_ms`.
 
@@ -270,3 +271,48 @@ sequenceDiagram
         Layer->>Runtime: finalize
     end
 ```
+
+## Fault tolerance
+
+Opt-in per Fleet with `FleetAlgoKnobFaultTolerance()`. Without it, a peer that
+stops responding during dispatch/combine trips a GPU `trap()` and takes the job
+down; with it, both transports mask the offending rank, skip it, and let the
+collective complete.
+
+**LOW_LATENCY only, on both transports** — `nccl_ep` leaves its mask buffer
+NULL under HIGH_THROUGHPUT (the mask APIs then abort the process) and `nixl_ep`
+has no HT mask at all. `validate_fleet_params` rejects the combination.
+
+### Mask convention
+
+Canonical across the package: **`int32[world_size]`, `1 = active`, `0 = masked`**,
+a CUDA tensor. This matches `ncclEpMaskQuery` and vLLM's `query_active_mask()`
+naming, so `nccl_ep` is a pass-through.
+
+`nixl_ep` differs on both axes and normalizes inside its own Fleet, so nothing
+else in the tree ever sees its convention:
+
+| | nccl_ep | nixl_ep |
+|---|---|---|
+| polarity | `1 = active` | **nonzero = masked** (buffer is `0xFF`-memset, so an untouched entry reads back as `-1`; kernels test `!= 0`) |
+| length | `world_size` | topology **capacity** (`query_mask_buffer` asserts on it) |
+| normalization | identity | `active = (raw[:world_size] == 0)` — **not** `1 - raw`, which yields `2` for never-connected tail ranks |
+
+### Fleet API
+
+| Method | Collective? | Blocks host? | Stream-ordered? |
+|---|---|---|---|
+| `supports_fault_tolerance` | — | no | no |
+| `query_fault()` | local | nccl no / nixl yes (small D2H) | nccl no / nixl yes |
+| `query_active_mask(out=None)` | local | no | **yes** |
+| `set_active_mask(mask)` | local (but must be applied identically everywhere) | no | **yes** |
+| `reconcile_active_mask()` | **store-collective**, tolerates dead ranks | yes (≤ timeout) | yes |
+| `clear_faults(readmit=False)` | local | no | no |
+| `clear_faults(readmit=True)` | **collective over survivors** | **yes** | yes |
+| `active_mask_epoch` | — | no | no |
+
+Reconciliation goes through the bootstrap rendezvous store
+(`resolve_rendezvous_store(..., subsystem="ft")`), never a `torch.distributed`
+allreduce: an allreduce over the EP group would hang on exactly the rank being
+masked out. See `core/comm/fault_tolerance.py` and the runbook for the protocol
+and the recovery state machine.
--- a/docs/design_docs/moe_ep_runbook.md
+++ b/docs/design_docs/moe_ep_runbook.md
@@ -331,21 +331,60 @@ DEGRADED ──peer gone for good──> update_topology(...)      [COLLECTIVE, 
 
 ### Ordering rules (all load-bearing)
 
-1. **Mutate the mask only between iterations.** Both transports read it *live*
-   from the dispatch/combine kernels, so changing it while a collective is in
-   flight is a race that can produce a half-masked dispatch. Retire the
-   iteration's `combine` (and `Handle.complete()` under
+0. **The steady state is read-only.** The transport discovers the fault: its
+   kernel times out on the peer and masks it. The application's job is to
+   *notice* — `query_fault()`, then `query_active_mask()`. `set_active_mask()`
+   is **not** a normal-path call; it exists so that reconciliation can impose an
+   agreed vector once survivors have compared views, and most callers should
+   only reach it indirectly through `reconcile_active_mask()`.
+
+1. **If you do write the mask, write it only between iterations.** Both
+   transports read it *live* from the dispatch/combine kernels, so changing it
+   while a collective is in flight is a race that can produce a half-masked
+   dispatch. Retire the iteration's `combine` (and `Handle.complete()` under
    `HandleAlgoKnobSplitOperation`) first.
+
 2. **All survivors must reconcile in the same iteration slot** — they must pass
    the same `active_mask_epoch`. Make the host's fault poll a synchronous
    decision point.
-3. `clear_faults(readmit=True)` **before** `update_topology`, never after:
-   `ncclEpMaskClean` needs a live handle on the *current* group, which
-   `update_topology` destroys and recreates.
-4. **No FT call may be issued during CUDA-graph capture.** A captured query
-   replays stale offsets; a captured set freezes one mask vector into every
-   replay. All stream-ordered entry points raise on capture. Call them from the
-   host between replays.
+
+3. **`clear_faults(readmit=True)` and `update_topology()` are alternatives, not
+   a sequence.**
+
+   * `clear_faults(readmit=True)` runs `ncclEpMaskClean` on the *same*
+     communicator: it re-admits a rank that was merely delayed.
+   * `update_topology()` destroys the group and builds a **new communicator**
+     (a fresh `ncclComm_t`, unless you adopted one via
+     `BootstrapConfig.nccl_comm`). That is the only way to add or replace a
+     process — matching `nccl_ep.h`, which notes that rank replacement needs
+     `ncclCommGrow` and a new EP group.
+
+   Re-admitting after a rebuild is meaningless (the group it targeted is gone)
+   and before one is wasted work. If you genuinely do both, `MaskClean` must
+   come first, since it needs a live handle on the *current* group.
+
+4. **No FT call may be issued during CUDA-graph capture** — but not for the
+   reason you might expect. Neither transport compacts the surviving ranks'
+   data layout, so **dispatch and combine are safe to capture**: they re-read
+   the mask on every replay, and a rank that fails later is still skipped.
+   Data from a masked rank is simply left in an unknown state with the error
+   flag raised.
+
+   What must not be captured is a *decision about* the fault state:
+
+   * `query_fault()` is a host-side read, not stream work, so it cannot be
+     captured at all. Inside a capture region it returns the capture-time
+     answer, and the branch taken on it is frozen into the graph's structure —
+     a graph that ignores faults forever because none had happened when it was
+     recorded. This is the quietest failure of the three, which is why it is
+     guarded even though it touches no stream.
+   * `set_active_mask()` bakes the rank and value into the captured operation,
+     so every replay re-applies that one mask.
+   * `query_active_mask()` can only be consumed on the host, which needs a sync
+     that capture forbids.
+
+   All four raise on capture with a per-operation explanation. Call them from
+   the host between replays.
 
 ### Backend asymmetries worth knowing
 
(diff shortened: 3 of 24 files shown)


mhoqueanik:
overall looks good to me. I will rebase this with my current working tree and share more feedback

lrbison:
from the NCCL perspective it looks about right to me. I scanned the documentation and the code.  I have a few comments about these "gotchyas":

> 1. Mutate the mask only between iterations (kernels read it live).

True, however the expectation is the kernel will discover the problem first, so torch application should be typically checking the mask/error rather than trying to update it.

> 3. clear_faults(readmit=True) before update_topology, never after: ncclEpMaskClean needs a live handle on the current group, which update_topology destroys.

As I understand it, update_topology() creates a new NCCL communicator? (it is the only option for NCCL to admit a new process).  In that case the graph has changed as well.

> 4. No FT call during CUDA-graph capture — a captured query replays stale offsets, a captured set freezes one mask into every replay. All stream-ordered entry points raise on capture.

I think I agree with this point just not the wording.  I agree that querying if mask has changed or if there has been an error during a graph capture is going to capture a graph that ignores errors by assuming whatever the capture saw will be true forever.

But regarding stale offsets, NCCL-EP doesn't "squeeze" the remaining ranks in data layout.  Instead the data arriving from those ranks is left in an unknown state, and an error is flagged.  The application may end up invoking the dispatch and combine again and again before the CUDA-graph is complete, and each time we respect the mask and don't wait on masked ranks.  In fact, the application may choose to re-route experts away from the failed rank with something like EPLB, in which case the rank remains failed, but the application can clear the error state and continue with a partial mask (without update_topology).  I think you call this clear_faults(readmit=False) right?

Anerudhan:
> True, however the expectation is the kernel will discover the problem first, so torch application should be typically checking the mask/error rather than trying to update it.

In steady state, `query_fault()`  and `query_active_mask()` are read only. 

`set_active_mask()` is not to be called, but if a kernel is timed out only the local rank will set the mask. 

If there are two rank which have different mask we need to combine them with bitwise and, to have agreed survivors ?

Done via `reconcile_active_mask()`. 

Hope that makes sense. Let me know if my understanding is incorrect.


> As I understand it, update_topology() creates a new NCCL communicator? (it is the only option for NCCL to admit a new process). In that case the graph has changed as well.

Yes. `update_topology` creates new communicator, and then creates a new group over it.

-  `clear_faults(readmit=True)` -->  `ncclEpMaskClean` on the  communicator: re-admits a merely-delayed rank.
- `update_topology()` → new communicator + new group: the only way to add or  replace a process.

> No FT call during CUDA-graph capture — a captured query replays stale offsets, a captured set freezes one mask into every replay. All stream-ordered entry points raise on capture.

Correct, will update the doc.

Question to you @lrbison 
`ncclEpMaskUpdate` has no callerd, so I have kept the reconciliation "push" step separable.  Is the framework pushing a
reconciled mask back via `ncclEpMaskUpdate` the use it was for ?

lrbison:
@Anerudhan 

Yes -- I think we are in agreement about query_active_mask / set_active_mask.

> If there are two rank which have different mask we need to combine them with bitwise and, to have agreed survivors

The approach we took was to assume that communication would naturally drive eventual consistency.  However you are right to explicitly synchronize masks at the FlashInfer level to be sure.


> `ncclEpMaskUpdate`: Is the framework pushing a reconciled mask back via ncclEpMaskUpdate the use it was for ?

Yes, the use case is framework-detected updates.  This could be recovering from a false-positive timeout: Some rank decided to move on and mask out a peer, but after the forward pass the framework did its own synchronization and decided that actually all peers are healthy and should rejoin without any process changes.  Or it could be some separate scale-down signal that application is handling by masking and exiting some of the ranks.  For scale-up or replacement the API is not useful unless NCCL eventually provides a way to admit new processes into an existing group (which isn't possible today).

saltyminty (approved):

lrbison (approved):

bkryu (dismissed):
Hi @Anerudhan, CI is failing on `test_deep_gemm_mega_kernel_vs_reference.py` in `cu130` configs. Plz take a look

mhoqueanik:
Hi @bkryu! Filed a fix PR here: https://github.com/flashinfer-ai/flashinfer/pull/4221

Anerudhan:
> #59999869

@mhoqueanik is going to be fixing this in a separate PR. Not introduced by this PR

nv-yunzheq (approved):

PR #4186, opened 2026-07-27
fix(moe): handle CuTe DSL finalize output tails

S1ro1:
## Summary

- clamp each fused-finalize bulk copy/reduction to the output columns remaining
  in the current N tile
- skip the output transfer when a cluster-padding CTA has no remaining columns
- remove the finalize-only exact-N-tiling restriction introduced by #4086
- retain the exact-N-tiling restriction for GEMM1, whose scale-factor output
  uses a separate unpredicated store path

## Why

The fused-finalize epilogue previously transferred one full compile-time CTA
tile for every valid row. This is unsafe when:

1. the final N tile is only partially populated, or
2. the persistent scheduler adds a padding CTA to complete an N cluster.

#4086 prevents those cases by rejecting any finalize configuration where
`N % (mma_n * cluster_n) != 0`. That is safe, but it removes otherwise useful
kernel configurations and can leave some widths with no eligible configuration.

This PR fixes the transfer itself. It computes the remaining columns for the
current tile, transfers only that many bytes, and performs no transfer when the
tile starts beyond the output width. All CTAs still execute the existing
commit/wait/barrier sequence.

### Correctness

Validated on NVIDIA GB200 with the newly enabled
`tile_m=256, mma_n=256, cluster_n=2` finalize configuration forced:

- `N=256`: exercises an empty cluster-padding CTA
- `N=384`: exercises a partially populated final N tile
- both numerical tests pass against the FP4 reference
- an additional 100 consecutive launches per case produced finite output and
  left a sentinel row immediately after the output unchanged

The focused host-side `can_implement` checks, Ruff checks, and formatting checks
also pass.

### Configuration coverage

The table counts complete GEMM1 + GEMM2 configurations exposed by the MoE
runner:

| Output width N | #4086 base | This PR |
| ---: | ---: | ---: |
| 256 | 12 / 16 | 16 / 16 |
| 384 | 4 / 16 | 16 / 16 |
| 512 | 16 / 16 | 16 / 16 |
| 2880 | 0 / 16 | 16 / 16 |
| 4096 | 16 / 16 | 16 / 16 |

For `N=2880`, all eight newly eligible finalize configurations were also
launched directly and produced finite output.

### Existing-case performance

The finalize kernel was benchmarked in isolation so the epilogue change was not
hidden by routing or GEMM1. The #4086 base and this PR were loaded in the same
process and measured in alternating pairs on one GB200:

- SM clock locked to 2062 MHz
- exact-tile widths `N=512, K=512` and `N=4096, K=1024`
- all eight previously valid finalize configurations at each width
- 31 hot-cache pairs with 250 launches per measurement
- compilation excluded from timing

Lower is better. Across the 16 exact-tile comparisons:

| Measurement | Median head/base | Range of per-configuration medians |
| --- | ---: | ---: |
| Hot cache | 0.9943 | 0.9882–0.9971 |
| Cold cache | 0.9934 | 0.9629–1.0249 |

No previously supported configuration showed a hot-cache slowdown. Cold-cache
single-launch measurements were noisier, but showed no systematic regression.

## Scope and dependencies

This PR changes only the finalize output transfer. GEMM1 keeps the conservative
guard from #4086.

The PR is stacked on #4086 and targets its upstream
`fix-3957-cluster-padding` branch so this diff contains only the durable
tail-handling change. It should be retargeted to `main` after #4086 merges.

The vLLM consumer is vllm-project/vllm#50030.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fused MoE handling for partial output tiles and cluster padding.
  * Prevented unsupported configurations from being selected during autotuning.
  * Limited output operations to valid columns, avoiding incorrect or out-of-bounds results.

* **Tests**
  * Added coverage for partial and exact tile sizes across multiple hidden dimensions.
  * Added validation for configuration acceptance and rejection without requiring GPU execution.


bkryu:
Hi @S1ro1, the current PR tries to merge into `flashinfer-ai:fix-3957-cluster-padding` instead of the `main` branch. Can you modify or re-open the PR so it merges into main?

S1ro1:
> Hi @S1ro1, the current PR tries to merge into `flashinfer-ai:fix-3957-cluster-padding` instead of the `main` branch. Can you modify or re-open the PR so it merges into main?

Hi, fixed the base to main, however this depends on the PR, if these 2 don't get merged in order I think it'd leave main in a "broken" state. I see the previous base is approved but just noting it down here

aleozlx (approved):

aleozlx:
added 0.6.18 label due to cherry picking to 0.6.17rc2 before merging to main

PR #4191, opened 2026-07-28
Split trtllm-gen cubin pins per arch

Vinnie6167:
## 📌 Description

The BMM & GEMM trtllm-gen cubins in ToT fail on SM 100. This MR introduces a WAR, loading an older set of published cubins when targeting SM 100.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

### DIFF
--- a/flashinfer/jit/fused_moe.py
+++ b/flashinfer/jit/fused_moe.py
@@ -281,12 +281,12 @@ def gen_trtllm_gen_fused_moe_sm100_module(enable_rubin: bool = False) -> JitSpec
     for header in BMM_EXPORT_HEADERS:
         h = get_artifact(f"{bmm_export_path}/{header}", get_meta_hash(checksum, header))
         assert h, f"{header} not found"
+    # Per-module export-header root: the Blackwell and Rubin variants must not
+    # share this symlink, or an AOT build (all modules generated, then compiled)
+    # lets the last gen_* call's target win and skews the other module's ABI.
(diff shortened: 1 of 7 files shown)


flashinfer/jit/fused_moe.py
kahyunnam, line 290: Does this symlink need to be different depending on enable_rubin as well, similar to gemm_path / gemm_export_path?

flashinfer/jit/gemm/core.py
kahyunnam: Same here for symlink needing to differ depending on enable_rubin

kahyunnam: same here

kahyunnam (commented):
In AOT, gen_all_modules() (aot.py) emits both when has_sm100 and has_sm107 (default arch list has 10.0a and 10.7a): each gen_* rewrites the same synlink, then all modules compile afterward. Last writer wins → Blackwell modules (enable_rubin=False, no -DTLLM_RUBIN_FEATURES) can compile against Rubin export headers → same GemmOptions ABI skew this PR is fixing.

Same issue in gen_trtllm_gen_gemm_module, gen_trtllm_low_latency_gemm_module, and gen_trtllm_gen_fused_moe_sm100_module

Image

flashinfer/jit/fused_moe.py
Vinnie6167: Right, fixed in cb2787dcd

Issue #4193, opened 2026-07-28
Public CI H100 failed


aleozlx:
### Before submitting

- [ ] I have searched existing issues and did not find a solution or an existing report.

### Environment

```text
N/A might be failed on CI job with multi-GPU

I have NOT searched existing issues and did not find a solution or an existing report

echo '===== JENKINS INFO ====='
+ echo NODE_NAME=flashinfer-h100-2gpu-a-i-0ba7dda1dd42ca685
+ echo EXECUTOR_NUMBER=0
+ echo WORKSPACE=/opt/dlami/nvme/actions-runner/runner-1/_work/flashinfer/flashinfer
+ echo BUILD_NUMBER=8864
+ echo WORKSPACE=/opt/dlami/nvme/actions-runner/runner-1/_work/flashinfer/flashinfer
+ echo '===== EC2 INFO ====='
+ ec2_metadata ami-id
+ curl -w '\n' -fsSL http://169.254.169.254/latest/meta-data/ami-id
===== JENKINS INFO =====
NODE_NAME=flashinfer-h100-2gpu-a-i-0ba7dda1dd42ca685
EXECUTOR_NUMBER=0
WORKSPACE=/opt/dlami/nvme/actions-runner/runner-1/_work/flashinfer/flashinfer
BUILD_NUMBER=8864
WORKSPACE=/opt/dlami/nvme/actions-runner/runner-1/_work/flashinfer/flashinfer
===== EC2 INFO =====
curl: (22) The requested URL returned error: 401

+ echo failed
failed
+ ec2_metadata instance-id
+ curl -w '\n' -fsSL http://169.254.169.254/latest/meta-data/instance-id
curl: (22) The requested URL returned error: 401

+ echo failed
failed
+ ec2_metadata instance-type
+ curl -w '\n' -fsSL http://169.254.169.254/latest/meta-data/instance-type
curl: (22) The requested URL returned error: 401

failed
+ echo failed
+ ec2_metadata hostname
+ curl -w '\n' -fsSL http://169.254.169.254/latest/meta-data/hostname
curl: (22) The requested URL returned error: 401

failed
+ echo failed
+ ec2_metadata public-hostname
Vulnerability Retbleed:                  Not affected
Vulnerability Spec rstack overflow:      Mitigation; Safe RET
Vulnerability Spec store bypass:         Mitigation; Speculative Store Bypass disabled via prctl
[...]

PR #4195, opened 2026-07-28
fix(comm): make TorchDistBackend.Split() synchronizing, and report every rank's fate on comm-test failure

aleozlx:
## 📌 Description

`tests/comm/test_comm_backend.py::test_torch_dist_backend_collectives[4]` failed once on the public H100 CI, and while investigating #4193 I found a genuine synchronization gap in `TorchDistBackend.Split()`. This PR closes that gap. It is **hardening prompted by** #4193, not a demonstrated fix for it — see Reviewer Notes before reading it as a resolution.

The observed CI failure was:

```
[rank2] Gloo connectFullMesh failed with [.../transport/tcp/pair.cc:547] Connection closed by peer [127.0.0.1]:18085
  ... TorchDistBackend.Split -> new_subgroups_by_enumeration -> ProcessGroupGloo(...)
```

**The gap.** Two facts combine into a real race window in `TorchDistBackend.Split`:

1. **`new_group()` does not synchronize ranks.** Its post-init store barrier only runs when `TORCH_DIST_INIT_BARRIER=1`, and that env var defaults to `0` (`_is_barrier_after_init` in torch's `distributed_c10d.py`). A rank therefore returns from `new_subgroups_by_enumeration()` as soon as *its own* side of the rendezvous is done.
2. **The gloo pair handshake is asymmetric.** Per `gloo/transport/tcp/pair.cc`: "one side takes a passive role and the other side takes an active role". The active side connects, writes a sequence number and returns; the passive side must still accept and *read* that sequence number.

Applied to the observed failure, that *would* look like this — presented as a hypothesis, since I could not reproduce it. With `world_size=4` and `color = rank % 2` the partition is `[[0, 2], [1, 3]]`. Rank 0 finished its half of sub-group `[0, 2]`, returned from `Split()`, published its result, called `destroy_process_group()` and exited — while rank 2 was still completing the mesh. Rank 2 then read EOF on the socket, which gloo reports as `Connection closed by peer`. Rank 0 exited `0` and only rank 2 died, which matches the CI log exactly (only `SpawnProcess-5` / `[rank2]` failed).

This is not confined to the test: `MnnvlMemory.set_comm_from_config()` in `flashinfer/comm/mnnvl.py` calls `Split()` the same way, so any caller doing rank-dependent work or tearing down right after a split is exposed.

**Fix.** Barrier on the **parent** group at the end of `TorchDistBackend.Split()`, making `Split()` collective *and* synchronizing — matching `MPI_Comm_split` (and `MPIBackend.Split`) semantics. No rank can leave `Split()` until every peer has finished its sub-group rendezvous. `Split()` runs once at setup, so the cost is irrelevant.

### Second change: make the next failure diagnosable

Independent of the barrier, and arguably the more useful half. When this test fails, the evidence that matters is the **first** death — and the harness was discarding it:

- `for p in procs: assert p.exitcode == 0` stopped at the first bad rank, so later ranks were never inspected. In the #4193 log we learned rank 2 raised and never found out what rank 3 did.
- A worker killed by a signal prints **no traceback at all**, so an OOM-kill or a runner shutdown was indistinguishable from a clean failure — it just surfaced as a peer that could not be reached.
- All workers shared the parent's stderr, so gloo/c10d output from N ranks interleaved into something unattributable.

Now each worker gets its own stdout/stderr file, redirected at the **fd** level so C++-side gloo and c10d logging is captured too, and every rank is reported on failure: status, whether it produced a result, and its output. Signal deaths are named, which separates an infra kill from a test failure at a glance:

```
AssertionError: workers [0, 1, 2, 3] did not exit cleanly:
  rank 0 (pid 68390): STILL RUNNING (timed out), NO RESULT
  rank 1 (pid 68391): KILLED by SIGKILL, NO RESULT
  rank 2 (pid 68392): STILL RUNNING (timed out), NO RESULT
  rank 3 (pid 68393): STILL RUNNING (timed out), NO RESULT
```

The failure is also bounded now: a worker that loses a peer blocks in a collective until gloo's own 30-minute timeout, turning a failure into a CI hang. Past the deadline survivors are killed and reported as timed out, with exit codes snapshotted first so the report says how each worker actually ended rather than how we ended it.

Verified by fault injection — one rank raising, and one rank `SIGKILL`ing itself — the report names the culprit with its traceback and marks the rest as victims, in 16s instead of hanging.

**Happy to split this into its own PR** if you'd rather it land independently of the barrier, which is the contested part. It touches the same function, so I kept them together to avoid a merge conflict rather than out of any conviction that they belong in one review.

## 🔍 Related Issues

Related to #4193 — but this PR **does not claim to fix it**, and deliberately does not use a closing keyword. See Reviewer Notes: the failure was never reproduced, and #4193 should stay open (and keep its `infra` label) independently of whether this merges.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

`tests/comm/test_comm_backend.py`: 9 passed, 1 skipped (mpi4py not installed).

I then tried hard to reproduce the failure on Linux/x86_64 with CPU-only gloo — the **tcp** transport, same as CI — on torch 2.11. Every pre-fix configuration below is green:

| build | scenario | runs | failures |
|---|---|---|---|
| pre-fix | pytest, idle host | 60 | 0 |
| pre-fix | pytest, 6-way concurrent, pinned to 2 CPUs | 72 | 0 |
| pre-fix | hard-exit stress, sub-groups of 2 / 4 / 8 ranks | 90 | 0 |
| pre-fix | hard-exit stress, 16 ranks timesharing 1 CPU | 20 | 0 |
| **pre-fix total** | | **242** | **0** |
| post-fix | pytest, idle host | 60 | 0 |

The "hard-exit stress" harness was written specifically to trigger the hypothesized mechanism: every rank calls `os._exit(0)` the instant `Split()` returns — no `destroy_process_group()`, sockets closed abruptly — and the sub-group is scaled up so the rendezvous has many more pairs and a wider spread between the first and last rank to finish. It did not reproduce.

## Reviewer Notes

**I could not reproduce the failure in 242 pre-fix runs, and I want that stated up front rather than buried.** Worse for my own hypothesis: the harness built specifically to trigger it (immediate `os._exit` after `Split()`, enlarged mesh, single-CPU starvation) is also green. That is evidence *against* the specific causal story, not for it. The chain is *inferred from the pytorch and gloo sources*, not demonstrated. Treat this PR as closing a synchronization gap that is provably there, rather than as a confirmed fix for #4193.

By the rule of three, 0 failures in 242 runs puts the 95% upper bound on the per-run failure rate at ~1.2%, and 1% is already in mild tension with the data — so the true rate is likely well under 1%. That has a direct consequence for review: **a green CI run on this PR is not evidence the fix works.** At p ≈ 1% a single run passes with ~99% probability even with the bug fully present; catching it with 95% confidence would take roughly 300 runs. Please don't read a green pipeline as validation here.

What **is** certain, independent of the repro: `new_group()` performs no cross-rank synchronization by default, so `Split()` today can return on one rank while a peer is still inside the sub-group rendezvous. That gap is real whether or not it is what bit CI on build 8864.

Note also that this is **not a regression** — `git log --follow` shows the current `Split()` implementation and its multi-process gloo test both landed together in 6104afe3 / #3701 on 2026-07-22, five days before the failure. There was no prior CI coverage that created gloo sub-groups, so nothing "started failing"; a new intermittent test started running. That is also why the gated per-PR check did not block it: a required check blocks deterministic failures, and #3701's own run was green.

Worth a reviewer's eye: the added barrier is on the parent group (`self._group`), which every rank reaches because `Split()` is already collective on that group (it starts with an `allgather`) — so it cannot deadlock, including when splitting an existing sub-group. It also adds no new device requirement beyond what the existing `all_gather_object` in `Split()` already imposes.

AI-assisted (Claude Opus 5).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved distributed subgroup splitting so all processes synchronize before the operation completes, preventing race conditions during setup and teardown.

* **Tests**
  * Expanded coverage for subgroup synchronization and collective operations.
  * Improved distributed test diagnostics, including per-process logs and clearer failure reporting.

### DIFF
--- a/flashinfer/comm/comm_backend.py
+++ b/flashinfer/comm/comm_backend.py
@@ -179,4 +179,14 @@ class TorchDistBackend:
         cur_group, _ = self._dist.new_subgroups_by_enumeration(
             _split_partition(all_info)
         )  # pyright: ignore
+        # new_group() does not synchronize the ranks by default (torch only runs
+        # its post-init store barrier when TORCH_DIST_INIT_BARRIER=1), so a rank
+        # can return from a sub-group rendezvous while a peer is still inside it.
+        # With gloo the rendezvous is eager: the peer that is still connecting
+        # then sees "connectFullMesh failed ... Connection closed by peer" as soon
+        # as the rank that raced ahead tears its process group down or exits.
+        # Barrier on the parent group so Split() is collective *and* synchronizing,
+        # matching MPI_Comm_split (and MPIBackend.Split) semantics. Split happens
+        # once at setup, so the cost is irrelevant.
+        self.barrier()
         return TorchDistBackend(group=cur_group)  # pyright: ignore
--- a/tests/comm/test_comm_backend.py
+++ b/tests/comm/test_comm_backend.py
@@ -22,8 +22,10 @@ logic is unit-tested against an injected fake mpi4py communicator.
 import importlib.util
 import multiprocessing as mp
 import os
+import queue
 import shutil
 import tempfile
+import time
 from typing import Any
 
 import pytest
@@ -63,6 +65,9 @@ def _backend_checks(rank: int) -> dict[str, Any]:
     res["sub_type"] = type(sub).__name__
     res["sub_rank"] = sub.Get_rank()
     res["sub_size"] = sub.Get_size()
+    # Collectives must work on the sub-group, not just report the right shape.
+    sub.barrier()
+    res["sub_allgather"] = sub.allgather(rank)
     return res
 
 
@@ -96,14 +101,28 @@ def _run_dist(world_size: int) -> dict[int, dict[str, Any]]:
     try:
         for p in procs:
             p.start()
+        # Drain the queue *before* join: a worker blocks on exit until its queue
+        # feeder thread has flushed, so join-then-read can deadlock. q.empty() is
+        # also unreliable, hence counting results instead of polling it.
+        results: dict[int, dict[str, Any]] = {}
+        deadline = time.monotonic() + _TIMEOUT_S
+        while len(results) < world_size:
+            remaining = deadline - time.monotonic()
+            if remaining <= 0:
+                break
+            try:
+                rank, res = q.get(timeout=min(remaining, 1.0))
+            except queue.Empty:
+                # Nothing more is coming once every worker is gone; fall through
+                # to the exitcode assert below, which gives the better message.
+                if all(p.exitcode is not None for p in procs):
+                    break
+                continue
+            results[rank] = res
         for p in procs:
-            p.join(timeout=_TIMEOUT_S)
+            p.join(timeout=max(0.0, deadline - time.monotonic()))
         for p in procs:
             assert p.exitcode == 0, f"worker pid={p.pid} exited with {p.exitcode}"
-        results: dict[int, dict[str, Any]] = {}
-        while not q.empty():
-            rank, res = q.get()
-            results[rank] = res
         assert len(results) == world_size, (
             f"expected {world_size} results, got {sorted(results)}"
         )
@@ -137,6 +156,8 @@ def test_torch_dist_backend_collectives(world_size: int) -> None:
         assert res["sub_type"] == "TorchDistBackend"
         assert res["sub_size"] == len(peers)
         assert res["sub_rank"] == peers.index(rank)
+        # Collectives on the sub-group see only the same-parity ranks.
+        assert res["sub_allgather"] == tuple(peers)
 
 
 class _FakeGroup:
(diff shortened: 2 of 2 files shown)


aleozlx:
Split into two PRs for clarity, as discussed:

- **#4205** — test instrumentation only (no change under `flashinfer/`). Per-rank output capture and full per-rank failure reporting, so the *next* occurrence of #4193 names the first process to die instead of only the process that noticed. Independently useful; mergeable on its own merits.
- **#4206** — the `Split()` parent-group barrier. Closes a synchronization gap that is provable from the torch and gloo sources, but **not** a demonstrated fix for #4193 (it never reproduced in 242 runs against the unfixed code).

They were together here only because both touched `_run_dist`; separating them means the diagnostics don't wait on the contested change. #4193 stays open.

Issue #4198, opened 2026-07-28
[Bug] mha_sm90.cu: spec-dec draft mask ignored in the SWAP_AB layout, wrong sliding-window masking for full masks, no ragged Q, scale loads before the PDL acquire


yichengj0:
### Before submitting

- [x] I have searched existing issues and did not find a solution or an existing report.

### Environment

SM90 only (`csrc/xqa/mha_sm90.cu`, the fp8 KV cache decode kernel). Reproduced on an H100 PCIe at 60783fb9 with torch 2.12.0+cu130 (CUDA 13.0), flashinfer built from source. The gaps are code-path level, not toolkit or driver specific, so the full collect_env output is omitted.

### Bug description

`mha_sm90.cu` has four gaps. The Python API routes calls hitting the first three to the generic `mha.cu` kernel (#4137, #4199), which is correct but gives up the Hopper fp8 fast path. This issue tracks fixing the kernel so the fallback can be removed.

- The SWAP_AB layout (used when `q_seq_len * head_group_ratio <= 32`) builds a causal mask from the column index and never reads the runtime draft mask, so a full (non-causal) draft mask returns causal results. Its window argument `tok0WinBeg` is accepted but unused.
- The per-row sliding-window path gets full draft masks wrong: on an H100, a draft row that should attend to a later draft token is off by 0.57 against a test tolerance of 0.1. Causal masks stay within tolerance.
- Ragged Q (`q_cu_seq_lens`) is not supported.
- A PDL ordering hazard, not limited to spec-dec: `qScalePtr` and `kvScalePtr` are loaded at the top of the kernel, before every live `acqBulk()`. Only direct `xqa()` callers passing device-tensor scales are exposed (the batch-decode API passes floats). Acquire-site analysis and fix sketch: [#4199 review](https://github.com/flashinfer-ai/flashinfer/pull/4199#issuecomment-5100990025).

### Minimal reproduction

On an H100, make `run_sm90_fp8_mha` unconditionally True in `flashinfer/xqa.py`, then:

```
pytest tests/attention/test_xqa_batch_decode.py -k "(mask_mode_deterministic or spec_dec_sliding_window or ragged_q) and fp8"
```

The deterministic mask test fails on the SWAP_AB gap, the sliding-window tests fail on full-mask configs, and the ragged tests raise.

The scale-load hazard has no functional repro (it needs a PDL producer grid still in flight at launch) and was found by inspection.

### Error messages / logs

[...]

PR #4199, opened 2026-07-28
fix(xqa): fix PDL load ordering and SM90 fp8 draft-mask dispatch

yichengj0:
## 📌 Description

Follow-up to #4137, addressing [the review feedback](https://github.com/flashinfer-ai/flashinfer/pull/4137#issuecomment-5099541418) that arrived after auto-merge. Thanks @qsang-nv for the detailed analysis.

Issues:

- Under programmatic dependent launch (PDL), the generic XQA kernel (`csrc/xqa/mha.cu`) read `q_cu_seq_lens` and the scale tensors before the acquire that makes a producer kernel's writes visible. `q_cu_seq_lens` drives the output row offset, so a stale read could write into the wrong request's rows.
- On SM90 with fp8 KV cache, the small-batch layout of the Hopper XQA kernel (`csrc/xqa/mha_sm90.cu`, used when `q_seq_len * head_group_ratio <= 32`) hardcodes a causal draft mask, so a full draft mask silently returned causal results. Existing refchecks use random data and cannot tell the two modes apart within fp8 tolerance.
- The XQA trace template declared a `q_cu_seq_lens` input that the trace reference silently dropped.

Fixes:

- Move the `q_cu_seq_lens` and scale-tensor loads below the PDL acquire in `mha.cu`.
- Extend the SM90 fp8 fallback so small-batch speculative decode also runs on the generic kernel, matching the sliding-window and ragged-Q cases. Restoring the Hopper fast path is tracked in #4198.
- Add a deterministic mask test: zero Q and K make each output row an exact mean of the visible V values, so any deviation from the requested mask fails loudly on every architecture. It runs two shapes, one that falls back to the generic kernel and one that stays on the Hopper kernel on SM90 fp8.
- Remove `q_cu_seq_lens` from the trace template until the trace reference supports ragged Q.
- Normalize the ragged-Q module key inside the module getter, and build a separate ragged variant only when an SM90 target is compiled; on other targets it is identical to the uniform module.
- Document the SM90 fp8 fallback in the `xqa()`, `xqa_batch_decode_with_kv_cache`, and `trtllm_batch_decode_with_kv_cache` docstrings.

## 🔍 Related Issues

#4198 (restore the Hopper fp8 fast path for speculative decode). Review thread: #4137.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

- Added `test_xqa_batch_decode_mask_mode_deterministic` (causal/full mask, bf16/fp8 KV, two head-group shapes) with exact expected outputs.
- On SM120 (RTX 5080): the new test, the ragged-Q and sliding-window suites (256 cases), and the trace suite (970 cases) pass.
- SM90/SM100 are covered by CI. On SM90 fp8 the new test's `head_grp_size=16` shape exercises the Hopper kernel and the `head_grp_size=4` shape exercises the widened fallback.

## Reviewer Notes

- Behavior change on SM90 with fp8 KV cache: small-batch speculative decode now runs on the generic kernel, including causal masks. The mask lives on the device, so dispatch cannot check its content without a sync. The Hopper kernel previously returned causal results regardless of the requested mask; #4198 restores that fast path.
- The `SPEC_Q_SEQ_LEN` build specialization is no longer reachable at run time but is kept: the fast-path restoration in #4198 re-enables it.
- Ragged workloads cannot be represented in trace dumps until the trace reference supports ragged Q; removing the inert input beats shipping a wrong reference.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved speculative decoding handling for variable-length queries, including safer execution ordering for mask/sequence offsets.
  * Refined SM90 FP8 KV-cache kernel selection, extending the conditions that fall back to the generic kernel (ragged queries, attention sinks, sliding-window, and small head-group bound).
  * Improved compilation/caching behavior to better match supported speculative-decoding configurations.

* **Documentation**
  * Updated the XQA documentation to clarify when the generic kernel is used for SM90 FP8 speculative decoding.

* **Tests**
  * Added a deterministic test covering causal/full speculative-decoding masks for both BF16 and FP8 KV caches.

### DIFF
--- a/csrc/xqa/mha.cu
+++ b/csrc/xqa/mha.cu
@@ -1579,8 +1579,6 @@ CUBIN_EXPORT __global__
 #endif
         uint32_t* __restrict__ semaphores = nullptr, void* __restrict__ scratch = nullptr) {
 
-  float const qScaleValue = qScalePtr != nullptr ? qScalePtr[0] : qScale;
-  float const kvCacheScaleValue = kvScalePtr != nullptr ? kvScalePtr[0] : kvCacheScale;
   assert(allowMultiBlockMode || gridDim.x == 1);
   bool const isMultiBlock = allowMultiBlockMode && (gridDim.x != 1);
   uint32_t const nbSubSeqPerSeq = allowMultiBlockMode ? gridDim.x : 1;
@@ -1594,37 +1592,7 @@ CUBIN_EXPORT __global__
   SharedMem& smem = *reinterpret_cast<SharedMem*>(&smemByteBuf[0]);
 
   uint32_t const idxReq = blockIdx.z;
-#if SPEC_DEC
-  // Variable query sequence length support.
-  bool const variableQSeqLen = qCuSeqLens != nullptr;
-  uint32_t const actualQSeqLen =
-      variableQSeqLen ? uint32_t(qCuSeqLens[idxReq + 1] - qCuSeqLens[idxReq]) : qSeqLen;
-  // Same as idxReq * qSeqLen if all sequences all the same.
-  // Take different beams as different requests/sequences currently.
-  uint32_t const reqSeqOffset = variableQSeqLen ? uint32_t(qCuSeqLens[idxReq]) : (qSeqLen * idxReq);
-  // A ragged request may have 0 draft tokens (all rejected). It owns no rows
-  // and min(row, actualQSeqLen - 1) would underflow, so exit before any
-  // barrier/semaphore interaction (uniform across the request's CTAs).
-  if (variableQSeqLen && actualQSeqLen == 0) {
-    return;
-  }
-
-  uint32_t const nbVHeads = nbKHeads;
-  uint32_t const nbQHeads = nbKHeads * headGrpSize;
-  uint32_t const nbQHeadTokens = nbQHeads * actualQSeqLen;
-  uint32_t const nbQKVHeads = nbQHeads + nbKHeads + nbVHeads;
-
-  uint32_t const nbTokenBlocksPerGrp = gridDim.y / nbKHeads;
-  uint32_t const idxHeadGrp = blockIdx.y / nbTokenBlocksPerGrp;  // inside one request
-  uint32_t const idxHeadTokenInGrp = (blockIdx.y % nbTokenBlocksPerGrp) * warpTile.y;
-  uint32_t const totalNbHeadTokensInGrp = actualQSeqLen * headGrpSize;
-  uint32_t const nbValidHeadTokens =
-      idxHeadTokenInGrp > totalNbHeadTokensInGrp
-          ? 0u
-          : mha::min(totalNbHeadTokensInGrp - idxHeadTokenInGrp, rowsPerBlock);
-  // Shift the mask ptr by batch_idx.
-  mask += reqSeqOffset * divUp(qSeqLen, 32u);
-#else
+#if !SPEC_DEC
   uint32_t const nbQHeads = nbKHeads * headGrpSize;
 
   uint32_t const idxHeadGrp = blockIdx.y;  // inside one request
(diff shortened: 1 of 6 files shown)


qsang-nv:
Verified the three items from the #4137 thread on `dda315f5`:

- **PDL ordering** — the `qCuSeqLens` and scale-tensor loads, the zero-length early return, and every derived index (`nbValidHeadTokens`, `mask +=`, …) all moved below `acqBulk()` in [`mha.cu#L1653-L1691`](https://github.com/flashinfer-ai/flashinfer/blob/dda315f5446acde347625a919add6cf31550d2b2/csrc/xqa/mha.cu#L1653-L1691). I checked the window that was vacated: L1582-L1651 now contains only asserts, `blockIdx`/`threadIdx` arithmetic and shared-memory barrier init — no global load is left ahead of the acquire — and the first uses of `qScaleValue`/`kvCacheScaleValue` are at L1806 and L2572.
- **SM90 fp8 dispatch** — the `swap_ab_eligible(...)` term closes the non-causal-mask hole, and the deterministic test's construction checks out (zero Q/K gives a uniform softmax; V values 1/2/4/8/16 are exact in e4m3; causal-vs-full expectations differ by 1.0 and 3.5 against `atol=5e-2`).
- **Module keying** — the uncached wrapper plus cached normalized helper is the right shape, and gating the ragged variant on an SM90 target is a real win: `SPEC_Q_SEQ_LEN` appears only in `defines.h`'s `static_assert` and `mha_sm90.cu`, and `mha.cu` doesn't include `specDec.h`, so re-emitting it on SM120/121 changes nothing and the ragged and uniform builds correctly collapse to one module there.

Three things below: one small ask for this PR, one for the #4198 follow-up, one doc nit.

---

### The deterministic test never reaches `mha_sm90.cu` on any architecture

The new test is pinned to `q_len=4, head_grp_size=4` ([L1085](https://github.com/flashinfer-ai/flashinfer/blob/dda315f5446acde347625a919add6cf31550d2b2/tests/attention/test_xqa_batch_decode.py#L1085)), so `q_seq_len * head_group_ratio == 16` and the new `swap_ab_eligible` term routes it to the generic kernel on SM90. SM100 and SM120/121 use `mha.cu` unconditionally. So across the whole matrix this test only ever validates `mha.cu`.

After the widened fallback, exactly one speculative-decode configuration still selects the Hopper kernel: fp8 KV, `q_seq_len * head_group_ratio > 32`, no sliding window, no ragged Q, no sinks — the non-SWAP_AB `warpGrpApplyMask` that does read the runtime mask. That path currently has only random-data refchecks, which is the exact class of coverage that failed to notice an entirely ignored mask mode last time.

One extra parametrize entry would cover it, e.g. a second shape with `num_kv_heads=2, head_grp_size=16` (`q_len=4` → 64 > 32). That gives the Hopper masking path a deterministic assertion today and a regression guard for when #4198 restores the fast path.

### Follow-up: the Hopper kernel still reads both scale tensors ahead of every acquire

This PR moves `qScalePtr` / `kvScalePtr` behind `acqBulk()` in `mha.cu`, but the same two loads sit at the very top of the Hopper kernel, [`mha_sm90.cu#L643-L644`](https://github.com/flashinfer-ai/flashinfer/blob/dda315f5446acde347625a919add6cf31550d2b2/csrc/xqa/mha_sm90.cu#L643-L644) — before even the `__CUDA_ARCH__ == 900` guard, and before every acquire site in the file.

Worth being precise about where those sites are, because the `acqBulk()` at [L1241-L1243](https://github.com/flashinfer-ai/flashinfer/blob/dda315f5446acde347625a919add6cf31550d2b2/csrc/xqa/mha_sm90.cu#L1241-L1243) is under `#if ENABLE_PDL == 1` and is dead here: `defines.h` selects `ENABLE_PDL 2` for `__CUDA_ARCH__ == 900`, which is the only configuration in which this kernel body exists. The live acquires are the four aggressive-mode sites at [L1269](https://github.com/flashinfer-ai/flashinfer/blob/dda315f5446acde347625a919add6cf31550d2b2/csrc/xqa/mha_sm90.cu#L1269), [L1290](https://github.com/flashinfer-ai/flashinfer/blob/dda315f5446acde347625a919add6cf31550d2b2/csrc/xqa/mha_sm90.cu#L1290), [L1316](https://github.com/flashinfer-ai/flashinfer/blob/dda315f5446acde347625a919add6cf31550d2b2/csrc/xqa/mha_sm90.cu#L1316) and [L1375](https://github.com/flashinfer-ai/flashinfer/blob/dda315f5446acde347625a919add6cf31550d2b2/csrc/xqa/mha_sm90.cu#L1375), all in the IO-warp branch and all after L643. The compute warps that consume `qScaleValue` at [L781-L783](https://github.com/flashinfer-ai/flashinfer/blob/dda315f5446acde347625a919add6cf31550d2b2/csrc/xqa/mha_sm90.cu#L781-L783) never execute an acquire at all.

Scope, so it can be prioritized properly: both ternaries short-circuit on a null pointer, so no load happens unless the caller passes tensor-backed scales. `xqa_batch_decode_with_kv_cache` passes Python floats, so the pointers are null there; the exposure is direct `xqa()` callers with device-tensor scales — a dynamic-quant serving stack, say. The route stays reachable after this PR for plain decode as well as for the `q_seq_len * head_group_ratio > 32` spec-dec case above.

The cheap fix keeps aggressive PDL intact: sink the two loads to their first use rather than the top of the kernel, placed after the compute warps' first `produced.wait()` — note that L776-L779 are `arrive()` calls, not waits, so L781 itself isn't yet ordered. Routing tensor-backed scales to the generic kernel would also work but is a heavy hammer for SM90 fp8 decode. Predates both PRs and lower risk than the `qCuSeqLens` case, so this reads as #4198 material rather than something to redo here.

### Nit: document the fallback under `window_left` too

The consolidated note in `xqa()` covers all the conditions, but the high-level [`window_left` docs](https://github.com/flashinfer-ai/flashinfer/blob/dda315f5446acde347625a919add6cf31550d2b2/flashinfer/decode.py#L3575-L3577) don't mention that SM90 fp8 speculative decode drops to the generic kernel for any configured window, including one that never truncates. The new note only sits under `q_cu_seq_lens`, so ragged-Q users see it and sliding-window users don't. While there: the `xqa()` note lists ragged Q, a positive `sliding_win_size` and the `<= 32` shape condition, but not the pre-existing `sinks is not None` fallback — worth listing all four in one place.

yichengj0:
@qsang-nv Thanks for the careful verification. All three items are handled:

- Added a `head_grp_size=16` shape (`q_len * ratio = 64`) to the deterministic mask test, so the non-SWAP_AB Hopper masking path gets a deterministic assertion. It passes on SM120, and H100 CI covers the Hopper leg.
- Agreed the `mha_sm90.cu` scale loads are #4198 material. Added them to that issue with your acquire-site analysis linked.
- Doc nits fixed: `window_left` now mentions the fallback, and the `xqa()` note lists all four conditions including sinks.

saltyminty (approved):
approved pending CI

PR #4202, opened 2026-07-28
fix: support fp8 e5m2 output in rmsnorm_quant and fused_add_rmsnorm_quant

elwhyjay:
## 📌 Description

E5M2 output is rarely used in inference, but the current code fails silently instead of rejecting it. This PR is a small fix to close that gap.

`rmsnorm_quant` and `fused_add_rmsnorm_quant` quantize to the dtype of the `out` tensor and the CUDA dispatch accepts e5m2, but:

- CuTe-DSL backend (default): the norm dtype map has no `float8_e5m2`  entry, so an e5m2 out tensor raises `KeyError`; the store epilogue and the clamp were e4m3 only anyway.
- CUDA fallback (`FLASHINFER_USE_CUDA_NORM=1`): the kernels clamp to `[-448, 448]` for any output dtype, silently collapsing the e5m2 range (max 57344).

Changes:

- `norm/utils.py`: `FLOAT8_E5M2_MAX`, e5m2 conversion intrinsics (hw via`cvt.rn.satfinite.e5m2x2.f32`, sw via bit manipulation), thin fp8 wrappers that select the variant from a constexpr dtype, and a
  `float8_e5m2` dtype map entry.
- `norm/kernels/{rmsnorm,fused_add_rmsnorm}.py`: quant epilogues dispatch on the output element type; the clamp bound follows the output dtype. Kernel signatures and compile cache keys are unchanged.
- `include/flashinfer/norm.cuh`: both quant kernels clamp with `QuantTypeStaticVals<O>::MAX_VAL`, as `layernorm_quant` already does. The trait moves above the kernels and drops its `ENABLE_FP8` guard:
  the fp8 types are always visible (`vec_dtypes.cuh` includes `cuda_fp8.h` unconditionally), and these kernels are instantiated for fp8 through `DISPATCH_DLPACK_DTYPE_TO_CTYPE_FP8` even in builds without `ENABLE_FP8`, so keeping the guard would break the no-fp8 build once they reference the trait (exercised by `test_norm_compilation_without_fp8`).
- Tests: `test_norm_quant` and `test_fused_add_rmsnorm_quant` parametrized over `quant_dtype` (e4m3fn, e5m2), same style as `test_layernorm_quant`.

Verified on RTX 5090: the full quant grids pass on both backends (3072 CuTe-DSL, 3085 CUDA cases). Both hw and sw conversion paths were checked bit-exact against torch fp8 casts over all finite fp16 values plus random and boundary floats, for both formats; since sm_120 always dispatches the hw path, the sw path was force-compiled with `use_hw_fp8=False` for this.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * Added FP8 E5M2 support for RMSNorm quantization and fused add + RMSNorm quantization.
  * Quantized outputs now support both FP8 E4M3 and E5M2 formats.
  * Clamping automatically uses the selected output format’s valid numeric range.

* **Documentation**
  * Clarified supported FP8 output formats and quantization behavior.

* **Tests**
  * Expanded coverage across both FP8 formats, including stride and large-tensor cases.

### DIFF
--- a/flashinfer/norm/__init__.py
+++ b/flashinfer/norm/__init__.py
@@ -199,7 +199,8 @@ def rmsnorm_quant(
     Parameters
     ----------
     out: torch.Tensor
-        The output tensor, will quantize the output to the dtype of this tensor.
+        The output tensor, will quantize the output to the dtype of this tensor,
+        which must be float8_e4m3fn or float8_e5m2.
     input: torch.Tensor
         Input tensor, 2D shape (batch_size, hidden_size).
     weight: torch.Tensor
@@ -312,7 +313,8 @@ def fused_add_rmsnorm_quant(
     Parameters
     ----------
     out: torch.Tensor
-        The output tensor, will quantize the output to the dtype of this tensor.
+        The output tensor, will quantize the output to the dtype of this tensor,
+        which must be float8_e4m3fn or float8_e5m2.
     input: torch.Tensor
         Input tensor, shape (batch_size, hidden_size).
     residual: torch.Tensor
--- a/flashinfer/norm/kernels/fused_add_rmsnorm.py
+++ b/flashinfer/norm/kernels/fused_add_rmsnorm.py
@@ -30,14 +30,14 @@ import torch
 from cutlass import Float32, Int32, Int64
 
 from ..utils import (
-    FLOAT8_E4M3_MAX,
     COPY_BITS,
+    get_fp8_max,
     rcp_approx_ftz,
-    cvt_and_store_f32_to_e4m3_hw,
-    cvt_and_store_f32_to_e4m3_sw,
-    cvt_and_store_8xf32_to_e4m3_hw,
-    cvt_and_store_4xf32_to_e4m3_hw,
-    cvt_and_store_2xf32_to_e4m3_hw,
+    cvt_and_store_f32_to_fp8_hw,
+    cvt_and_store_f32_to_fp8_sw,
+    cvt_and_store_8xf32_to_fp8_hw,
+    cvt_and_store_4xf32_to_fp8_hw,
+    cvt_and_store_2xf32_to_fp8_hw,
     has_hw_fp8_cvt,
     get_ptr_as_int64,
     get_sm_version,
@@ -426,7 +426,8 @@ class FusedAddRMSNormQuantKernel:
 
     Computes:
     1. residual = input + residual (in-place update)
-    2. output = clamp(residual / sqrt(mean(residual^2) + eps) * weight / scale, -448, 448)
+    2. output = residual / sqrt(mean(residual^2) + eps) * weight / scale,
+       clamped to the finite range of the FP8 output dtype (E4M3 or E5M2)
 
     """
 
@@ -737,6 +738,7 @@ class FusedAddRMSNormQuantKernel:
         # against M or used in the address arithmetic below.
         actual_row = Int64(bidx) * rows_per_block + row_in_block
         col_offset = lane_in_row * vec_size
+        fp8_max = get_fp8_max(mY.element_type)
 
         if cutlass.const_expr(self.use_hw_fp8 and vec_size == 8):
             for v in cutlass.range_constexpr(num_vec_blocks):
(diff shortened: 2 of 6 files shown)


lazypool (commented):
Reviewed the diff against `main`. This is the more complete of the two fixes for #4053 — it addresses the default CuTe-DSL backend (where e5m2 currently raises `KeyError`), the CUDA fallback clamp, and the API docs. A few notes:

1. **The `ENABLE_FP8` guard removal is necessary and correct.** The dispatch cases in `csrc/tvm_ffi_utils.h` are gated on `FLASHINFER_ENABLE_FP8_E4M3`/`FLASHINFER_ENABLE_FP8_E5M2`, and the JIT always defines those (`flashinfer/jit/core.py:548-549`) even in builds where `ENABLE_FP8` is omitted (as `test_norm_compilation_without_fp8` does). So `RMSNormQuantKernel` genuinely *is* instantiated with fp8 `O` in the no-`ENABLE_FP8` build, and the trait must be visible unconditionally. The move above the kernels is fine (it's a header-only trait). Worth a sentence in the commit message since it looks like a scope change at first glance.

2. **The SW E5M2 conversion checks out numerically.** I re-derived the boundary logic against the existing E4M3 SW path:
   - Underflow flush at `abs_bits <= 0x37000000` (= 2^-17) is exactly the RNE midpoint between 0 and the E5M2 min denormal 2^-16 — consistent with the E4M3 path flushing at 2^-10 (midpoint between 0 and 2^-9).
   - Saturation `min(norm_raw, 123)` = exp 30 / mant 11 = 57344, matching `cvt.rn.satfinite` semantics; the rounding-carry path (`radj` add before the min) cannot overflow into the inf exponent.
   - Denormal RNE (`dr_bit`/`dsticky`/`odd`) mirrors the E4M3 pattern.
   The bit-exactness claim vs torch over all finite fp16 + boundary floats is credible.

3. **Nit — NaN handling differs between hw and sw:** the SW path saturates NaN to 57344, while `cvt.rn.satfinite.e5m2.f32` produces the NaN pattern (0x7F). This matches the pre-existing E4M3 SW behavior, so it's not a regression, but worth documenting since the clamp (`fminf`/`fmaxf`) does not sanitize NaN inputs.

4. **Nit — inconsistent dispatch:** `get_fp8_max` raises `ValueError` on unknown dtypes, but the `cvt_and_store_*_to_fp8_*` wrappers silently fall back to E4M3 for anything that isn't `Float8E5M2`. Consider raising in the wrappers too, or reusing `get_fp8_max`'s check, so an unexpected output type fails loudly instead of producing E4M3 data.

5. **Overlap with #4056:** both PRs edit the same two clamp lines in the same kernels in `include/flashinfer/norm.cuh`. Since this PR is the broader of the two (both backends + tests running on the default config), it would make sense for #4056 to be closed in favor of this one (its stricter `static_assert` with a message could be folded in).

lazypool (commented):
Nice, this is the more complete fix of the two — the DSL path was just crashing with a KeyError on e5m2 outputs, which is honestly worse than the CUDA clamp issue, so good to see both covered.

A few things after going through it:

1. The ENABLE_FP8 guard removal: I was suspicious at first but checked and it's actually necessary. The dispatch in csrc/tvm_ffi_utils.h is gated on FLASHINFER_ENABLE_FP8_E4M3/E5M2, and the JIT always defines those even when ENABLE_FP8 isn't set (flashinfer/jit/core.py:548). So the kernels do get instantiated with fp8 types in the no-fp8 build and the trait has to be visible. Worth one line in the commit message though, since it looks like a scope change at a glance.

2. Spent a while on the SW e5m2 asm since it's the bulk of the change. The underflow flush at 2^-17 is the RNE midpoint between 0 and the min denormal 2^-16, which matches the e4m3 path flushing at 2^-10 (midpoint of 0 and 2^-9), and the saturation to 57344 (exp 30 / mant 11) mirrors what the hw satfinite does. Looks right to me, and the bit-exact claim against torch is reassuring.

3. One thing I noticed: NaN inputs. The SW path saturates them to 57344 while the hw path produces the NaN pattern. Same behavior as the existing e4m3 SW code, so not a regression, but worth a code comment since the clamp won't catch NaNs.

4. Minor: get_fp8_max raises on unknown dtypes but the cvt_and_store_* wrappers silently fall back to e4m3. Would be nice if they behaved the same way.

Also heads-up: #4056 edits the same clamp lines in norm.cuh, so there's a conflict coming — probably makes sense for that one to be superseded by this.

bkryu (approved):

PR #4212, opened 2026-07-28
feat: paged KV cache support for the modular CuTe-DSL Blackwell prefill

pgera:
## Summary

Adds paged KV cache support to the modular CuTe-DSL Blackwell (SM100) prefill kernel and routes `BatchPrefillWithPagedKVCacheWrapper(backend="cute-dsl")` to it. Paged execution runs at parity with the modular ragged kernel across page sizes 16–128, and within a few percent at page size 8 (dedicated page-table producer warp — see below), with bitwise-identical outputs on identical logical problems at every page size.

## Architecture

**Kernel structure** (context for where paging plugs in — unchanged by this PR): the modular prefill is a warp-specialized kernel, one CTA per (256-row Q tile, KV head group, batch item), 16 warps in fixed roles with per-role `setmaxnreg` budgets:

| warps | role | regs/thread |
|---|---|---|
| 0–3 / 4–7 | softmax warpgroups (one per 128-row Q half-tile) | 192 |
| 8–11 | correction (online-softmax rescale of O partials) | 96 (88 paged) |
| 12 | MMA (tcgen05 QK^T and PV) | 32 (40 paged) |
| 13 | load (TMA Q/K/V) | 32 (40 paged) |
| 14 | epilogue (TMA store O / LSE) | 32 (40 paged) |
| 15 | empty (page-table producer at page_size 8) | 24 |

Roles communicate through mbarrier pipelines: the load warp fills a double-buffered Q ring and a shared K/V smem ring (3 stages bf16, 4 stages fp8); the MMA warp consumes them into TMEM S tiles; softmax reads S and writes P; MMA runs PV; correction rescales output partials; the epilogue stores O. Wall time in the small-band regime is set by the per-KV-tile cadence of this pipeline, which is what paging must not disturb.

**Paged design** (per-page TMA mechanism from the in-repo MLA decode loader):

- **Layout**: the kernel reads K and V from page pools viewed as `(num_pages, page_size, H_kv, head_dim)`. Only `head_dim` must be contiguous; the other three strides are runtime arguments of a single compiled kernel (one per page size), so the same binary handles an NHD-contiguous pool, an HND pool passed as a transposed view, or the K/V slices of a combined `(num_pages, 2, ...)` cache. The TMA descriptor is built once on the host over the entire pool, and a copy selects which page to load by passing that page's physical index as one of its TMA coordinates — the page-table indirection is a coordinate, not per-page pointer arithmetic.
- **Logical/physical split**: the load warp is the only role that sees pages. It takes `kv_page_table` + `kv_page_indptr`; mask, softmax, correction and epilogue operate on logical token cumsums, untouched.
- **Zero-fill contract**: out-of-range page ids become −1 → the TMA coordinate goes out of bounds, the smem box is zero-filled, and complete-tx still credits the full byte count. This covers pages past the sequence tail *and*, on sliding-window kernels, table slots below the window start: serving frameworks repoint reclaimed slots at a shared null block that may hold garbage/NaN, and score masking cannot save the V path (P = +0.0, but 0×NaN = NaN through the PV MMA), so the load itself must be skipped. The window clamp runs only on an item's first KV tile (the window lower bound falls inside it by construction); the table read is min-clamped to the item's slice (on the page-table-ring path below, the same clamps are applied at produce time instead).
- **Page-table ring (page_size 8)**: at 16 pages per KV tile the loader's in-loop id loads — one global-latency load per TMA copy, twice per tile across the K and V loops — dominate its issue cadence. The otherwise-idle empty warp therefore stages each tile's ids into a 4-stage smem ring (`cp.async`, its own tile-scheduler instance, tail/window → −1 clamps applied at produce time), and the load warp reads them back at LDS latency — the MLA-decode pt-loader design. Compile-time gated to `pages_per_kv_tile >= 16`: at larger page sizes the ring's fixed per-tile handshake measured worse than the id loads it removes.
- **Register rebalance** (paged-only): one 8-register `setmaxnreg` quantum moves from correction to the mma/load/epilogue group. Spill-free either way; measured ~1–2% at ps16/32.

Non-paged kernels are unaffected: all paged code is `const_expr`-gated and ragged cubins compile byte-identical.

## Performance

B200, CUPTI cold-L2 interleaved, 3-pass medians (9 passes for the two >4 ms cells, which drift at 3), hq/hk = 32/8, scrambled page tables, ratio = paged / modular-ragged on identical logical problems (bitwise-identical outputs; sliding-N = sliding window of N + causal):

| dtype | d | B | S | mask | ragged (µs) | ps8 | ps16 | ps32 | ps64 | ps128 |
|---|---|---|---|---|---|---|---|---|---|---|
| bf16 | 128 | 1 | 16K | sliding-1024 | 374 | 1.071 | 1.027 | 1.017 | 1.024 | 1.013 |
| bf16 | 128 | 4 | 2K | sliding-512 | 158 | 1.069 | 1.025 | 1.019 | 1.026 | 1.009 |
| bf16 | 128 | 4 | 2K | causal | 172 | 1.072 | 1.041 | 1.007 | 1.014 | 1.000 |
| bf16 | 128 | 16 | 512 | causal | 97 | 1.058 | 1.015 | 1.011 | 1.019 | 1.000 |
| bf16 | 128 | 8 | 8K | causal | 4498 | 1.015 | 1.014 | 1.014 | 1.006 | 0.977 |
| bf16 | 128 | 2 | 16K | causal | 4221 | 1.031 | 1.018 | 1.009 | 1.019 | 1.021 |
| bf16 | 128 | 4 | 2K | full | 243 | 1.075 | 1.072 | 1.039 | 1.010 | 1.024 |
| bf16 | 128 | 16 | 512 | full | 94 | 1.091 | 1.080 | 1.028 | 1.018 | 1.013 |
| bf16 | 64 | 4 | 2K | sliding-512 | 132 | 1.066 | 1.013 | 0.995 | 0.998 | 0.992 |
| bf16 | 64 | 4 | 2K | causal | 150 | 1.080 | 1.058 | 1.037 | 1.035 | 1.009 |
| bf16 | 64 | 16 | 512 | causal | 79 | 1.083 | 1.052 | 1.034 | 1.028 | 1.013 |
| bf16 | 64 | 8 | 8K | causal | 3043 | 1.043 | 1.021 | 1.007 | 1.004 | 0.999 |
| fp8 | 128 | 4 | 2K | sliding-512 | 144 | 1.050 | 1.005 | 1.005 | 1.009 | 0.993 |
| fp8 | 128 | 4 | 2K | causal | 160 | 1.050 | 1.025 | 1.025 | 1.028 | 1.007 |
| fp8 | 128 | 8 | 8K | causal | 3160 | 1.023 | 1.015 | 1.008 | 1.010 | 0.999 |

## Tests

`tests/attention/test_modular_fmha_prefill_paged.py`: bitwise-vs-ragged anchor (identity + scrambled tables × masks × page sizes 8–128 × varlen incl. 1/7-token seqs × partial last page), pool-garbage immunity, NaN canaries (tail + null-block window slots) at head dims 64 and 128, window clamp boundary cases, LSE/fp8/mixed-dtype feature matrix, wrapper integration + rejection paths (dtype universe, combined-cache-with-mixed-dtypes, unsupported page size).

## Summary by CodeRabbit

* **New Features**
  * Added paged KV-cache support for CuTe DSL prefill attention.
  * Added page-table execution for variable-length pages and NHD/HND cache layouts.
  * Added support for causal and windowed attention, FP8 and mixed-dtype caches, LSE output, ALiBi, soft-capping, and selected ragged-cache workloads.
  * Added validation for cache shapes, layouts, dtypes, page sizes, and unsupported options.

* **Bug Fixes**
  * Improved handling of invalid, trailing, and window-excluded pages.

* **Tests**
  * Added comprehensive paged-versus-ragged output coverage across supported configurations.

### DIFF
--- a/flashinfer/prefill.py
+++ b/flashinfer/prefill.py
@@ -2412,7 +2412,11 @@ class BatchPrefillWithPagedKVCacheWrapper:
 
                 if cute_variant is not None:
                     raise NotImplementedError(
-                        "cute-dsl backend does not support combining ALiBi with logits_soft_cap"
+                        "the public cute-dsl route cannot combine ALiBi with "
+                        "logits_soft_cap (one plan-time variant per kernel); "
+                        "a fused custom variant can express both — use "
+                        "flashinfer.cute_dsl.attention."
+                        "BatchPrefillCuteDSLWrapper with plan(..., variant=...)"
                     )
                 cute_variant = SoftCappingAttention(cap=logits_soft_cap)
 
@@ -2761,7 +2765,11 @@ class BatchPrefillWithPagedKVCacheWrapper:
                 )
             if sinks is not None:
                 raise NotImplementedError(
-                    "cute-dsl paged backend does not support attention sinks"
+                    "the public cute-dsl paged route does not accept run-time "
+                    "attention sinks yet; the backend supports sinks as a "
+                    "plan-time variant — use flashinfer.cute_dsl.attention."
+                    "BatchPrefillCuteDSLWrapper with "
+                    "plan(..., variant=AttentionWithSink(sinks))"
                 )
             if skip_softmax_threshold_scale_factor is not None:
                 raise NotImplementedError(
@@ -3590,7 +3598,11 @@ class BatchPrefillWithRaggedKVCacheWrapper:
 
                 if variant is not None:
                     raise NotImplementedError(
-                        "cute-dsl backend does not support combining ALiBi with logits_soft_cap"
+                        "the public cute-dsl route cannot combine ALiBi with "
+                        "logits_soft_cap (one plan-time variant per kernel); "
+                        "a fused custom variant can express both — use "
+                        "flashinfer.cute_dsl.attention."
+                        "BatchPrefillCuteDSLWrapper with plan(..., variant=...)"
                     )
                 variant = SoftCappingAttention(cap=logits_soft_cap)
 
@@ -4033,6 +4045,19 @@ class BatchPrefillWithRaggedKVCacheWrapper:
                 raise NotImplementedError(
                     "cute-dsl backend does not support NVFP4 packed KV cache "
                     "(kv_cache_sf)"
+                )
+            # vLLM's metadata builder stashes attention sinks on the wrapper
+            # for the fmha_v2 route; the cute-dsl route would silently
+            # ignore them (wrong results), so fail loudly with the
+            # supported alternative.
+            if getattr(self, "_sinks", None) is not None:
+                raise NotImplementedError(
+                    "attention sinks were set on the wrapper (_sinks) but "
+                    "the public cute-dsl ragged route does not consume "
+                    "them; the backend supports sinks as a plan-time "
+                    "variant — use flashinfer.cute_dsl.attention."
+                    "BatchPrefillCuteDSLWrapper with "
+                    "plan(..., variant=AttentionWithSink(sinks))"
                 )
             # enable_pdl is a launch-latency hint: forwarded on the FMHA
             # route below; the modular kernel has no PDL support.
--- a/flashinfer/prefill.py
+++ b/flashinfer/prefill.py
@@ -1660,10 +1660,12 @@ class BatchPrefillWithPagedKVCacheWrapper:
             mask will be used in attention computation.
 
         backend : str
-            The implementation backend, could be ``auto``/``fa2``/``fa3``/``cudnn`` or ``trtllm-gen``.
+            The implementation backend, could be ``auto``/``fa2``/``fa3``/``cudnn``/``trtllm-gen``
+            or ``cute-dsl``.
             Defaults to ``auto``.
             If set to ``auto``, the wrapper will automatically choose the backend based on the
             device architecture and kernel availability.
+            The ``cute-dsl`` backend uses the CuTe DSL attention kernel for Blackwell (SM100+).
 
         jit_args : Optional[List[Any]]
             If provided, the wrapper will use the provided arguments to create the JIT module,
--- a/flashinfer/prefill.py
+++ b/flashinfer/prefill.py
@@ -1677,10 +1677,12 @@ class BatchPrefillWithPagedKVCacheWrapper:
         )
         _check_kv_layout(kv_layout)
 
+        self._cute_dsl_wrapper = None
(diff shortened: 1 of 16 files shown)


pgera:
@flashinfer-bot run

pgera:
@flashinfer-bot run

pgera:
@flashinfer-bot run

flashinfer/prefill.py
```diff
@@ -2375,7 +2379,73 @@ def plan(
...
+                head_dim_vo=head_dim_qk,
+                causal=causal,
+                sm_scale=_sm_scale,
+                q_data_type=q_data_type,
+                kv_data_type=kv_data_type if kv_data_type is not None else q_data_type,
+                window_left=window_left,
+                variant=cute_variant,
+                page_size=page_size,
+                paged_kv_indptr=paged_kv_indptr,
+                paged_kv_indices=paged_kv_indices,
```
saltyminty, line 2444: Should we be consuming the wrapper owned `self._qo_indptr_buf` and `self._paged_kv_*_buf`here instead of `qo_indptr` and `paged_kv_*`?

tests/attention/test_modular_fmha_prefill_paged.py
```diff
@@ -0,0 +1,710 @@
...
+
+
+SHAPES = [
+    ([512, 512], [512, 512]),  # uniform, page-aligned
+    ([7, 300, 1291], [23, 300, 1547]),  # varlen, partial last pages, s_k > s_q
+]
+
+
+@pytest.mark.parametrize("causal", [False, True])
+@pytest.mark.parametrize("page_size", [16, 64, 128])
```
saltyminty, line 235: We also support page_size=8, should we add that to the test sweep?

saltyminty (approved):

pgera:
@flashinfer-bot run

tests/attention/test_modular_fmha_prefill_paged.py
pgera: Yes, added. Also had to add a separate perf optimization for page_size 8 as the latency of id loads becomes quite high.

flashinfer/prefill.py
pgera: Yes, good catch.

PR #4168, opened 2026-07-27
fix(moe): reject incompatible output-scale cubins

rosenrodt:
## 📌 Description

Reject TRTLLM Gen batched-GEMM cubins whose output scaling-factor dtype does
not match the format allocated and consumed by the MoE pipeline.

For block-format FC1 output, MoE allocates the default block-SF representation:
E4M3 for NVFP4 and UE8M0 for MX formats. The affected
`bmm_E2m1xFp32_*` cubins instead write linear FP32 factors. On the small
routed-NVFP4 `32 x 1024`, 16-expert fixture, that writes roughly 44 KiB into a
roughly 16 KiB E4M3 buffer and then feeds the wrong representation to FC2.
The result is non-finite output and can surface later as
`cudaErrorInvalidAddressSpace`.

Qualification retains the existing one-way input per-token requirement and
checks the causal output-SF contract:

- block-format output must use that format's default SF dtype;
- non-block output must not produce an output SF tensor.

This removes exactly 132 `E2m1 -> FP32/Linear` FC1 cubins from the current
5,194-entry bundle. It removes no candidates from MXFP4+MXFP8, MXFP4+BF16,
DeepSeek FP8, MXFP8, or NVFP4 per-token execution.

The FP8 per-tensor and FP4 tactic-query constructors are also aligned with the
runtime constructor so enumeration and execution use the same per-token flags.
There is no replay-time predicate or kernel overhead.

## 🔍 Related Issues

No linked issue.

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

The changed-file hooks pass:

```text
pre-commit run --files \
  csrc/trtllm_batched_gemm_runner.cu \
  csrc/trtllm_fused_moe_kernel_launcher.cu \
  tests/moe/test_trtllm_gen_moe_autotune_tactics.py
```

The all-files check was not run.

## 🧪 Tests

- [x] Tests have been added or updated as needed.
- [x] All tests are passing (`unittest`, etc.).

The regression is index-free and sweeps every tactic advertised for the
[...]

PR #4122, opened 2026-07-24
Adds SM107 support

Vinnie6167:
## 📌 Description

Adds SM107 (compute capability 10.7) support to FlashInfer, so its existing kernels and backends compile and run on SM107 GPUs.

Key changes:
- SM107a enablement across existing backends — capability lists, JIT architecture flags, and dispatch paths extended to recognize and target sm_107a.
- CuTe DSL device-support check with sm_100f family fallback — before selecting a CuTe DSL backend, probe whether the installed DSL can target the device; when it lacks a native SM107 target, fall back to the sm_100f family target so kernels still compile and run correctly.
- Auto-detected CUTLASS SM107 support — for CUTLASS-based modules, the sm_107a→sm_100f mapping is now driven by probing the bundled CUTLASS headers, so it upgrades to native sm_107a automatically once CUTLASS gains SM107, with no code change.
- Toolchain-aware architecture probes (NVRTC, Triton) — availability checks so unsupported compiler/DSL toolchains skip cleanly instead of failing deep inside compilation.
- Arch-aware test handling — tests gate on, or adjust tolerances for, SM107 where hardware-dependent behavior differs.

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [ ] Tests have been added or updated as needed.
- [ ] All tests are passing (`unittest`, etc.).

## Summary by CodeRabbit

* **New Features**
  * Expanded SM107 (compute capability 10.7) support across FP4 quantization, GEMM, MoE, and attention workflows.
  * Added optional `use_fp16_softmax` and `uses_spcompress` controls for paged and ragged attention variant selection.
  * Added runtime CuTe DSL architecture validation before CuTe-based execution.
* **Bug Fixes**
  * Improved TensorRT-LLM attention kernel selection and shared-memory configuration, including new variant traits.
* **Documentation**
  * Updated source-install instructions to include SM107 in the CUDA architecture list.
* **Tests**
  * Enhanced architecture-aware test skipping and SM107 coverage; relaxed a few numerical tolerances.

PR #4215, opened 2026-07-28
fix(jit): pass map_sm107_to_100f in gen_moe_utils_module

aleozlx:
## 📌 Description

`gen_moe_utils_module()` is the only MoE JitSpec that does not pass `map_sm107_to_100f=True`. On Rubin (SM107) this breaks the build outright:

```
nvcc fatal   : Unsupported gpu architecture 'compute_107a'
```

No public CUDA toolkit (13.0 or 13.2) accepts `compute_107a`, so the SM107 → `sm100f` family mapping is required.

### Why the omission is not merely redundant

These flags are passed as `extra_cuda_cflags`. In `build_cuda_cflags` (`flashinfer/jit/cpp_ext.py`), the `module_has_gencode` branch lets module-level flags **override** the globally computed, correctly-mapped flags. So leaving the argument off does not fall back to the global mapping — it *removes* it.

### Blast radius

Strictly a no-op off Rubin. `CompilationContext.get_nvcc_flags_list` only rewrites the gencode when `(major, minor) == (10, "7a")`:

```python
if apply_sm107_mapping and major == 10 and minor == "7a":
    flags.append("-gencode=arch=compute_100f,code=sm_100f")
```

`apply_sm107_mapping` is additionally gated on `not cutlass_supports_sm107()`, so once the bundled CUTLASS gains native `compute_107a` support this stops applying automatically, with no further code change.

Affects the CuteDSL MoE path — `gen_moe_utils_module()` is reached from `aot.py` and `flashinfer/fused_moe/cute_dsl/moe_utils.py`.

## 🔍 Related Issues

Found while investigating the Rubin NVFP4 MoE failure. This is **independent of** and complementary to #4168 / #4213 (incompatible output-scale cubins) — that one fixes a runtime correctness bug, this one fixes a build break. Both are needed on Rubin.

### ✅ Pre-commit Checks

## 🧪 Tests

Verified on Hecate GR100 (cc 10.7, aarch64): without this change the `moe_utils` JIT module fails to compile on SM107; with it the module builds and the MoE suites run.

No test added — this is a build-configuration fix on a JitSpec, exercised by any SM107 build of the CuteDSL MoE path.

## Reviewer Notes

Single-line change plus comment. Every other MoE JitSpec already opts into this mapping; this one appears to have simply been missed when SM107 support was added in #4122.

### DIFF
--- a/flashinfer/jit/moe_utils.py
+++ b/flashinfer/jit/moe_utils.py
@@ -74,8 +74,14 @@ def gen_moe_utils_module() -> JitSpec:
         "-DENABLE_FP4",
     ]
 
+    # SM107 (Rubin) must be built for the sm100f family target: no public CUDA
+    # toolkit (13.0 or 13.2) accepts ``compute_107a``, so without this mapping
+    # nvcc aborts with "Unsupported gpu architecture 'compute_107a'". Every other
+    # MoE JitSpec already opts in; this one was missed. Because these flags are
+    # passed as ``extra_cuda_cflags`` they override the correctly-mapped global
+    # flags, so omitting it here is not merely redundant -- it breaks the build.
     nvcc_flags += current_compilation_context.get_nvcc_flags_list(
-        supported_major_versions=[10]
+        supported_major_versions=[10], map_sm107_to_100f=True
     )
 
     return gen_jit_spec(


kahyunnam (approved):
makes sense to me!

PR #4222, opened 2026-07-28
feat(attention): cuDNN paged prefill via direct mixed-form cu_seq_lens

egilliam-nv:
## 📌 Description

Extends the cuDNN prefill **direct (token-unit) sequence-length path** (#3921) to **paged KV**, using the cuDNN **mixed sequence-length form** (cumulative on Q, per-batch on KV) that landed in cuDNN backend 9.25 + cudnn-frontend (per-side relaxation).

**How it works.** The paged case binds:
- the token-unit `qo_indptr` as `cu_seq_len_q` — which *also* serves as the Q/O ragged offset (per-tensor multipliers applied in the graph builder), and
- the per-request KV lengths as `seq_len_kv`.

KV is addressed through the page tables, so — unlike the non-paged both-cumulative path — **no KV-side cumulative prefix sum is materialized, and no element-unit offset conversion is performed**. This is the FlashInfer paged shape the mixed-form work was built to enable.

Gated separately from the non-paged path via `_cudnn_supports_direct_seqlens(dtype, mixed=True)` (backend `>= 92500`, frontend `>= 1.27`). Below the gate, the existing legacy element-conversion paged graph is used unchanged — no behavior change for current cuDNN versions.

Aside: the direct path also sidesteps the int32 overflow latent in the legacy path's element-unit offsets (`token_offset * num_heads * head_dim`) for large models, since token-unit indptrs stay at token scale.

## Scope / status

Core low-level path only, kept localized to `flashinfer/cudnn/prefill.py`. **The following are deferred to a separate follow-up PR:**
- Wiring `BatchPrefillWithPagedKVCacheWrapper` (`backend="cudnn"`) to pass token-unit `qo_indptr` + `actual_seq_lens_kv` so the wrapper takes the direct path.
- LSE base-2 fold / de-pad for the unified-attention output contract.
- A NOT_SUPPORTED feature-probe fallback (the FE exposes no compiled/effective-version query, so the gate is a version compare today).

## 🔍 Related

- #3921 (token-unit indptr / cu_seq_len direct path — non-paged)
- cuDNN mixed-form sequence-length support (backend + cudnn-frontend, merged separately)
- Aligns with P2 of the paged-prefill unification proposal (cuDNN adapter: paged + cu_seq_len direct plumbing)

## 🧪 Tests

New `tests/attention/test_cudnn_prefill_paged_cu_seqlens.py`: compares the paged **direct** path against the paged **legacy** path (itself validated against the fa2 reference by `test_cudnn_prefill`) across batch × seqlen × page_size × GQA × causal — **32/32 passed on H100 (sm90)**. The test probes real mixed-paged capability and skips cleanly on older cuDNN.

Regression: `test_cudnn_prefill` (paged) 288 passed, `test_cudnn_prefill_token_indptr` (non-paged direct) 130 passed.

## Summary by CodeRabbit

- **New Features**
  - Added mixed sequence-length handling for paged attention.
  - Added support for independently supplied query and key/value sequence lengths.
  - Improved compatibility with direct sequence-length processing for standard and paged inputs.

- **Bug Fixes**
  - Updated processing to use direct handling when supported, with safe fallback behavior otherwise.

- **Tests**
  - Added coverage across batch sizes, sequence lengths, page sizes, head counts, and causal attention modes.

### DIFF
--- a/flashinfer/cudnn/prefill.py
+++ b/flashinfer/cudnn/prefill.py
@@ -18,7 +18,7 @@ except Exception:
 
 
 @functools.cache
-def _cudnn_supports_direct_seqlens(dtype: torch.dtype) -> bool:
+def _cudnn_supports_direct_seqlens(dtype: torch.dtype, *, mixed: bool = False) -> bool:
     """True if cuDNN can consume token-unit indptr buffers directly for `dtype`.
 
     Requires cu_seq_len_q/kv SDPA inputs and per-tensor ragged-offset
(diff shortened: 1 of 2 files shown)


saltyminty:
Agreed with the above CodeRabbit comments that the test coverage should be improved. Also, is there a reason for creating a new file as opposed to utilizing the existing `test_cudnn_prefill_paged_cu_seqlens.py`?

saltyminty (approved):

PR #4229, opened 2026-07-29
fix(xqa): read the runtime draft mask on the SWAP_AB spec-dec path

Smallfu666:
## 📌 Description

Partially addresses #4198: fixes the SWAP_AB runtime-mask and sliding-window gaps on the SM90 fp8 XQA spec-decode path. Ragged Q and the PDL scale-load hazard remain open (follow-ups).

On the SWAP_AB layout (`q_seq_len * head_group_ratio <= 32`), `warpGrpApplyMask` synthesized a causal mask from the column index and never read `specDec.params.mask`; the `tok0WinBeg` argument was accepted but unused, so sliding-window begin-masking was also lost on this path.

Changes:

- `csrc/xqa/mha_sm90.cu` — rewrite the `SWAP_AB && SPEC_DEC` variant of `warpGrpApplyMask`:
  - read the runtime draft mask through the same `SpecDec::needMask` / `loadTileMaskRow` helpers as the non-SWAP_AB variant, with the lookup transposed for this layout (mask row selected by the q-token column, bit index = the thread's KV row); `TileMaskRow` loads are deduplicated across adjacent columns sharing a q token;
  - apply the per-column sliding-window begin mask (`globalRow < tok0WinBeg + maskCol`), with signed comparisons (`tok0WinBeg` can be deeply negative) and explicit shift clamping instead of relying on PTX `shl.b64` clamp behavior;
  - fix the early-exit condition so window-edge tiles far from the draft region are still masked (previously the function returned early based on the draft region alone, which is one structural reason `tok0WinBeg` was dead);
  - keep the hard-coded causal behavior as a fallback when `params.mask == nullptr`.
- `flashinfer/xqa.py` — remove the `use_sliding_window` and `swap_ab_eligible` conditions from the SM90 fp8 MHA fallback guard (the kernel now handles both); keep the `use_ragged_q` and `sinks` fallbacks; update the note.
- `flashinfer/decode.py` — drop the now-stale docstring sentence saying sliding-window falls back to the generic kernel.
- `tests/attention/test_xqa_batch_decode.py` — comment-only: label the deterministic-test parametrizations (`(2,4)` → SM90 fp8 SWAP_AB path, `(2,16)` → non-SWAP_AB path).

The kernel change also clamps `maskStartRow` in the null-mask fallback (`cacheSeqLen >= SPEC_Q_SEQ_LEN ? cacheSeqLen - SPEC_Q_SEQ_LEN : 0`) so a `cacheSeqLen < SPEC_Q_SEQ_LEN` call from direct C++ users cannot underflow and skip masking in release builds.

## 🔍 Related Issues

Partially addresses #4198 (gaps (a) and (b); ragged Q and the `qScalePtr`/`kvScalePtr` PDL load-ordering item are intentionally left for follow-ups). Builds on the deterministic tests introduced in #4199.

### ✅ Pre-commit Checks

## 🧪 Tests

Validation on a matching reported CUDA/PyTorch stack — H100 (SXM) / SM90, CUDA 13.0 (nvcc 13.0.88), torch 2.12.0+cu130, source build (JIT), on current main + this patch. Notes: the original report is from an H100 PCIe; the node's kernel driver here is 550.127.08 with the CUDA 13.0 forward-compatibility package, not a native 580-series driver.

| test (H100) | before (unpatched, SM90 fp8 MHA forced on) | after (patched, default dispatch) |
|---|---|---|
| `spec_dec_sliding_window[300-full-False-…-63-…-2-32-2-4]` | FAILED | PASSED |
| `spec_dec_sliding_window[300-full-False-…-63-…-4-64-4-2]` | FAILED | PASSED |
| `spec_dec_sliding_window[300-full-False-…-127-…-2-32-2-4]` | FAILED | PASSED |
| `spec_dec_sliding_window[300-full-False-…-127-…-4-64-4-2]` | FAILED | PASSED |
| `mask_mode_deterministic[2-4-full-fp8]` | FAILED | PASSED |
| causal controls (deterministic causal + all causal window cases) | PASSED | PASSED |
| full `pytest tests/attention/test_xqa_batch_decode.py -k fp8` | — | 1380 passed / 480 skipped / 0 failed |

The 16 remaining before-failures in the forced-on targeted subset are the known sinks + spec-dec accuracy TODO (unchanged by this PR; sinks still fall back in normal dispatch).

Primary development validation — H200 / SM90, CUDA 13.0, torch 2.13.0, source build (JIT): same 5 cases fail before / pass after; post-rebase targeted regression 68 passed / 0 failed; full `-k fp8` (pre-rebase tree) 1378 passed / 0 failed.

- [x] Tests have been added or updated as needed (this PR relies on the deterministic + sliding-window tests introduced in #4199; the only test-file change here is comment labeling of the parametrizations).

## Reviewer Notes

- I could not reproduce out-of-tolerance full-mask + sliding-window results for **non-SWAP_AB** shapes (q_len×grp products 40/64 pass at tol 0.1 on both H100 and H200), so this PR only changes the SWAP_AB variant; the non-SWAP_AB `warpGrpApplyMask` is untouched.
- The known sinks + spec-dec accuracy issue on the forced SM90 path is unchanged (pre-existing TODO, still gated by the sinks fallback in normal dispatch).
- The `qScalePtr`/`kvScalePtr` PDL load-ordering hazard added to #4198 is not addressed here; happy to take it as a follow-up.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved speculative decoding mask handling for custom masks, causal masking, sequence boundaries, and sliding-window limits.
  * Prevented incorrect accumulator values when tokens fall outside valid masking or sequence ranges.
  * Improved SM90 FP8 speculative-decoding kernel selection, including clearer handling for ragged queries and sink values.

* **Documentation**
  * Updated XQA documentation to reflect current FP8 fallback behavior.
  * Clarified test coverage comments for SM90 head-group configurations.

### DIFF
--- a/csrc/xqa/mha_sm90.cu
+++ b/csrc/xqa/mha_sm90.cu
@@ -1758,41 +1758,119 @@ __device__ inline void warpGrpApplyMask(Gemm0Acc& acc, SpecDec const& specDec,
 #endif
                                         uint32_t cacheSeqLen, uint32_t idxTile, uint32_t warpRank) {
   constexpr uint32_t tileSize = gemm0CtaTileNbTokens;
-  static_assert(SPEC_Q_SEQ_LEN <= sizeof(MaskType) * 8, "not implemented");
-
-  assert(cacheSeqLen >= SPEC_Q_SEQ_LEN);
-  uint32_t const maskStartRow = cacheSeqLen - SPEC_Q_SEQ_LEN;
-  uint32_t const tileStartRow = tileSize * idxTile;
-  if (tileStartRow + tileSize < maskStartRow) {
-    return;
-  }
+  constexpr uint64_t fullMask = ~uint64_t{0};
+  static_assert(tileSize == sizeof(fullMask) * 8, "not implemented");
 
   uint32_t const idxInQuad = laneId() % 4;
   uint32_t const idxQuad = laneId() / 4;
+  uint32_t const tileStartRow = tileSize * idxTile;
 
+  if (specDec.params.mask == nullptr) {
+    // Fallback: no device-side mask available; synthesize a causal draft mask
+    // (legacy behavior).
+    static_assert(SPEC_Q_SEQ_LEN <= sizeof(MaskType) * 8, "not implemented");
+    // cacheSeqLen < SPEC_Q_SEQ_LEN is reachable via direct C++ callers; clamp
+    // instead of letting the uint32 subtraction underflow, which would make
+    // the early-exit below skip masking entirely in release builds.
+    uint32_t const maskStartRow = cacheSeqLen >= SPEC_Q_SEQ_LEN ? cacheSeqLen - SPEC_Q_SEQ_LEN : 0;
+    if (tileStartRow + tileSize < maskStartRow) {
+      return;
+    }
+#pragma unroll
+    for (uint32_t n = 0; n < acc.cols; n++) {
+#pragma unroll
+      for (uint32_t j = 0; j < GmmaAccCoreMat::cols; j++) {
+        uint32_t const col = GmmaAccCoreMat::cols * (4 * n + idxInQuad) + j;
+        uint32_t const maskCol = col / headGrpSize;
+        MaskType const bit_mask = (1ULL << (maskCol + 1)) - 1;
+
+#pragma unroll
+        for (uint32_t m = 0; m < acc.rows; m++) {
+#pragma unroll
+          for (uint32_t i = 0; i < GmmaAccCoreMat::rows; i++) {
+            uint32_t const row = gmma::instM * m + gmma::instM / 4 * warpRank + 8 * i + idxQuad;
+            uint32_t const globalRow = tileStartRow + row;
+            if (globalRow >= cacheSeqLen) {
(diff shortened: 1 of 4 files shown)


Smallfu666:
Closing this PR. This is not a claim that the underlying problem is fixed.

We are not continuing this work, so the PR should not stay open as though we were. #4198
stays open and is not affected by this: it tracks four gaps, this PR addressed two of them
(the SWAP_AB runtime mask and the sliding-window begin mask), and ragged Q and the PDL
scale-load ordering were always left as follow-ups. Please feel free to unassign me there, so it is
not waiting on me.

Closing changes nothing in-tree. The dispatch guard from #4199 is still live in
`flashinfer/xqa.py`, so the affected shapes keep falling back to the generic kernel exactly
as they do today.

The work stays on the PR for whoever picks it up — the rewritten SWAP_AB `warpGrpApplyMask`,
and the before/after validation in the description on H100 and H200. One caveat: my check
that the defect still survives on current `main` is a source-level reading of
`csrc/xqa/mha_sm90.cu`, not a re-run of those tests.

Issue #4012, opened 2026-07-17
GB300 (SM103): enable_flashinfer_autotune triggers cudaErrorNotPermitted during CUDA-graph capture for NVFP4 MoE


guanxingithub:
## Summary

On **GB300** (Blackwell-Ultra, SM103, `compute_cap 10.3`), enabling
`enable_flashinfer_autotune` for **NVFP4 block-scale MoE** models makes vLLM engine
startup crash at **CUDA-graph capture** with:

```
torch.AcceleratorError: CUDA error: operation not permitted
```
(`cudaErrorNotPermitted`), raised from `torch.cuda.graph(...).capture_end()`.

Importantly, the **FlashInfer autotune sweep itself completes cleanly**
(`[Autotuner]: Autotuning process ends`) — the crash happens *afterwards*, during
CUDA-graph capture. This is therefore distinct from #3197 (illegal-memory access
*during* the `trtllm_fp4_block_scale_moe` sweep), which appears resolved in 0.6.14.

## Environment

- FlashInfer **0.6.14**
- vLLM **0.24.0**
- GPU: **NVIDIA GB300** (SM103 / `compute_cap 10.3`), single GPU, TP=1
- `quantization=modelopt_fp4`, `enable_flashinfer_autotune=true`
- container run as root

## Reproduce (dummy weights are sufficient)

```bash
vllm bench throughput \
  --model=nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-NVFP4 \
  --load-format=dummy --quantization=modelopt_fp4 --tensor-parallel-size=1 \
  --num-prompts=128 --input-len=256 --output-len=256 --max-model-len=2048 \
  --gpu-memory-utilization=0.9 --trust-remote-code \
  --kernel-config '{"enable_flashinfer_autotune": true}'
```

## Observed

The FP4 autotune sweep finishes, then during
`Capturing CUDA graphs (mixed prefill-decode, PIECEWISE)` it dies deterministically at
~graph 26/51:

```
[Autotuner]: Autotuning process ends
Capturing CUDA graphs (mixed prefill-decode, PIECEWISE):  51%|#####     | 26/51
...
  File ".../torch/cuda/graphs.py", line 169, in capture_end
    super().capture_end()
torch.AcceleratorError: CUDA error: operation not permitted
RuntimeError: Engine core initialization failed.
```

Reproduced 2/2 with the crash at the same graph index.

## What isolates it (A/B)

| condition | result |
[...]

Issue #3530, opened 2026-06-08
[BUG] FlashInfer MoE autotune crashes on multi-node TP8 Qwen3.5-397B NVFP4 (GB200 / GB300)


vadiklyutiy:
## Run characteristics

For performance measurement, **612 vLLM server runs** were executed on **GB200** (306) and **GB300** (306), serving **Qwen3.5-397B-A17B-NVFP4** with multi-node TP8, vLLM 0.22.1rc1.dev231+gec0a31d4a, and FlashInfer fused MoE kernels. **8 runs failed** during FlashInfer MoE autotuning; the other 604 initialized successfully.

**This bug is flaky.** The same launch configuration can pass or crash across runs — failures are intermittent, not deterministic, and occur across unrelated workload settings (see tables below). A failed configuration often succeeds on retry.

**All 8 failures are TP8 without expert parallel (not TP8+TEP).**

| GPU | Cluster | Runs | FlashInfer crashes |
|-----|---------|------|--------------------|
| GB300 | lyris | 306 | 2 |
| GB200 | ptyche | 306 | 6 |
| **Total** | | **612** | **8** |

## Failure summary

Both failure modes below are **flaky** — observed on only a small fraction of runs with no consistent repro rate.

| Failure type | GB300 | GB200 | Total |
|--------------|-------|-------|-------|
| 1. Segmentation fault during FlashInfer MoE autotune | 0 | 4 | **4** |
| 2. CUDA illegal memory access during FlashInfer MoE autotune/init | 2 | 2 | **4** |

### Failure 1 — Segmentation fault during FlashInfer MoE autotune (4 runs)

FlashInfer autotuner completes (or nearly completes) `flashinfer::trtllm_fp4_block_scale_moe`, then crashes while tuning `flashinfer::trtllm_bf16_moe`:

```
!!!!!!! Segfault encountered !!!!!!!
... flashinfer::trtllm_bf16_moe ...
... FusedMoeLauncher::run -> cuFuncSetAttribute ... TrtllmGenBatchedGemmRunner ...
```

| GPU | Parallelism | Workload |
|-----|-------------|----------|
| GB200 | TP8 | conc=1, isl=8192, osl=1024 |
| GB200 | TP8 | conc=16, isl=1, osl=1000 |
| GB200 | TP8 | conc=16, isl=8192, osl=1024 |
| GB200 | TP8 | conc=2048, isl=1, osl=1000 |

[...]

PR #4237, opened 2026-07-29
fix(moe): pad trtllm-gen route map by one element to avoid OOB read

syuoni:
## 📌 Description

The routed dynamic-batch batched-GEMM kernels (`Bmm_*_dynB` with `-routeAct`)
read one `int32` past the end of `ptrRouteMap`, from the last batch-dim CTA, on
every launch. The read is speculative — its value is never consumed, so outputs
are always bitwise correct — but it faults with `CUDA_ERROR_ILLEGAL_ADDRESS`
whenever the allocation happens to end at a mapped-region boundary. That
allocator-placement dependence is what makes it surface as *flaky* MoE autotune
and inference crashes.

Root cause is kernel-side: an off-by-one clamp in the hoisted load-task
initializer (`WarpGrpThreadIdx` is clamped to the load-group size instead of
size − 1), so off-group threads compute row `tileN` of a `tileN`-row tile, i.e.
index `(ctaIdxY + 1) * tileN`. For the last CTA that is exactly one element past
the shape documented in `KernelParamsDecl.h`
(`[sum(divUpMul(N[bi], tileN) for bi in B)]`) — which is what we allocate. It has
been reported to the kernel owners and is being fixed there.

This PR is the integration-side workaround: allocate the route map with one
extra element, so the already-shipped prebuilt cubins stay in bounds. The
overrun is provably always exactly one `int32` from one CTA, so `+1` is
sufficient by construction, not a heuristic. The pad slot's contents are
irrelevant since the value is never used.

Two allocation sites, both `permuted_idx_to_token_idx` (→ `routeMap`). Marked
`WAR` + `TODO` so the `+1` can be dropped once regenerated cubins land.

## 🔍 Related Issues

Likely explains #3530 and #3168 (and possibly #4012, #2776) — all report
intermittent NVFP4 MoE autotune crashes, and #3168 additionally reports silent
garbage output, which is the expected signature of an OOB read that usually
lands on mapped memory.

### ✅ Pre-commit Checks

## 🧪 Tests

No test added — the failure is allocator-placement dependent and does not
reproduce deterministically through the public API. Validation was done on
`v0.6.15.post1` (B200, SM100, Kimi-K2.5 NVFP4 TP4):

| Check | stock | route map +1 |
|---|---|---|
| `compute-sanitizer` memcheck, M=2 bucket × 4 ranks | 40 × `Invalid __global__ read of size 4` | **0 violations** (816 profiles/rank) |
| Guard-page probe (route map placed at the tail of a VMM mapping) | faults on every launch | **clean** |
| Full-model TP4 autotune crash loop | crashed by attempt 2 in 3/3 loops | **6/6 clean** |

The guard-page probe is the load-bearing one: it removes all dependence on
allocator placement, so it is deterministic in both directions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability for fused mixture-of-experts routing operations by preventing out-of-bounds memory access in supported execution paths.
  * Applied the safeguard to both standard and FP4 block-scale routing workflows.

### DIFF
--- a/csrc/trtllm_fused_moe_kernel_launcher.cu
+++ b/csrc/trtllm_fused_moe_kernel_launcher.cu
@@ -450,8 +450,10 @@ class FusedMoeLauncher {
     expanded_idx_to_permuted_idx =
         alloc_tensor({args->num_tokens * totalExpertsPerToken}, dl_int32, hidden_states.device());
 
+    // WAR: the routed batched-GEMM kernels read one int32 past the end of the route map.
+    // TODO: drop the +1 once the fixed kernel cubins land.
     permuted_idx_to_token_idx =
-        alloc_tensor({max_num_padded_tokens}, dl_int32, hidden_states.device());
+        alloc_tensor({max_num_padded_tokens + 1}, dl_int32, hidden_states.device());
 
     if (gemm1_bias_type == batchedGemm::gemm::BiasType::Mn) {
       permuted_idx_to_expanded_idx =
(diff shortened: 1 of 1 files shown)


zhyncs (approved):

zhyncs:
Per our discussion, let's take this hot fix for now. The proper fix is expected to take longer to land, and we can replace this workaround once it's ready.

samuellees:
cc @rosenrodt  for viz

rosenrodt:
Thanks for the fix. Will notify here when cubin fix lands

samuellees (approved):
LGTM. For CI fails on GB300, it's about AllReduce and not relative with this PR. And this is not relative with 5090.

PR #4238, opened 2026-07-29
fix(trtllm-gen moe): persistent, CUDA-graph-capture-safe routing scratch

### LINKED ISSUE #3168: [Bug] trtllm_fp4_block_scale_moe autotune corrupts device state on num_tokens=16384 bucket (SM100/SM103, NVFP4 MoE, vLLM-padded hidden 2688→2816)

### LINKED ISSUE #3427: FlashInfer TRTLLM BF16 MoE IMA on B200 with SGLang RL rollout, correlated with t128x8x128 + schPd2x1x2x3 BMM kernel at effbs=1024
## Summary

We are enabling a Miles RL rollout backend to use SGLang with the FlashInfer TRTLLM BF16 MoE backend for Qwen3-30B-A3B on B200. The initial SGLang hot-update layout issue was addressed in [sgl-project/sglang#25692](https://github.com/sgl-project/sglang/pull/25692). After that fix, the run can start and produce sane rollout metrics, but we still intermittently hit a CUDA illegal memory access in the FlashInfer TRTLLM BF16 MoE path.

After several repro and profiling runs, the strongest signal we have is that the failure correlates with a specific BF16 BMM kernel family selected at larger effective rollout batch size.

Important caveat: this is only the pattern we observed, not proof that this kernel family is inherently broken. We are filing this to ask for help interpreting the coredump and narrowing down whether this is a FlashInfer kernel issue, an SGLang integration issue, or an invalid input/workspace contract issue.

## Environment

- GPU: B200 / SM100
- Model: `Qwen/Qwen3-30B-A3B`
- Workload: Miles RL rollout with SGLang
- SGLang MoE backend: FlashInfer TRTLLM BF16 MoE
- SGLang patch: [sgl-project/sglang#25692](https://github.com/sgl-project/sglang/pull/25692)
- `flashinfer-python`: `0.6.11.post1`
- `flashinfer-cubin`: `0.6.11.post1`
- `flashinfer-jit-cache`: `0.6.11.post1+cu129`
- `torch`: `2.11.0+cu129`
- `sglang`: `0.5.13.dev20+gd9a0158` plus the PR patch above
- `sglang-kernel`: `0.4.2.post2+cu129`

## Workload Shape

We tested with response length capped at 1024 to get a faster repro. This is intentionally a stability repro, not a reward-quality run.

The key sweep was:

- `BATCH_SIZE=32`, `N_SAMPLES_PER_PROMPT=8`, effective batch = 256, `SGLANG_CUDA_GRAPH_MAX_BS=256`
- `BATCH_SIZE=64`, `N_SAMPLES_PER_PROMPT=8`, effective batch = 512, `SGLANG_CUDA_GRAPH_MAX_BS=512`
- `BATCH_SIZE=128`, `N_SAMPLES_PER_PROMPT=8`, effective batch = 1024, `SGLANG_CUDA_GRAPH_MAX_BS=1024`

Common settings:

```bash
FAULT_TOLERANCE=disabled
STEPS=3000
SKIP_EVAL=1
ROLLOUT_MAX_RESPONSE_LEN=1024
WANDB_MODE=online
```

## Observed Pattern

All runs used FlashInfer TRTLLM BF16 MoE + the SGLang PR #25692 patch.

### effbs=256, maxbs=256, len1024 (GOOD)

- Clean in the observed window.
- Profiler saw `schPd2x1x2x3` variants, but not the suspicious `t128x8x128 + schPd2x1x2x3` combo.

### effbs=512, maxbs=512, len1024 (GOOD)

- Profile collected; no conclusive IMA in that profile window.
- Profiler saw `t128x8x128` variants, but they were `schedS` variants, not `schPd2x1x2x3`.

### effbs=1024, maxbs=1024, len1024 (BAD!!)

- Reproduced CUDA illegal memory access.
- Lightweight CUDA coredump points to this BF16 BMM kernel family:

```text
bmm_Bfloat16_Bfloat16Bfloat16_Fp32_t128x8x128_s4_et128x8_m128x8x16_c1x1x1_16dp256b_rM_BN_transOut_schPd2x1x2x3_bN_rgTma_clmp_dynB_sm100f
```

Another rank/device saw:

```text
bmm_Bfloat16_Bfloat16Bfloat16_Fp32_t128x8x128_s5_et128x8_m128x8x16_c1x1x1_16dp256b_rM_BN_transOut_schPd2x1x2x3_bN_rgTma_clmp_dynB_sm100f
```

Both were launched as:

```text
<<<(16,32,1),(256,1,1)>>>
```

So the pattern we see is:

- `effbs=256`: no `t128x8x128 + schPd2x1x2x3` (GOOD)
- `effbs=512`: `t128x8x128` appears, but as `schedS` (GOOD)
- `effbs=1024`: IMA, and coredump lands on `t128x8x128 + schPd2x1x2x3` (BAD!!)

Again, this is a correlation from our runs, not proof that this kernel name is the root cause.

## Coredump Details

The coredump was generated with lightweight CUDA coredump enabled. The run reached rollout 2, then the SGLang workers became unavailable and the router returned `503 no_available_workers`.

The coredump decoded with `cuda-gdb target cudacore` shows:

```text
warning: Encountered unhandled device exception (30)
CUDA Exception: Device Unknown Exception
```

For one dump:

```text
[Current focus set to CUDA kernel 0, grid 2320505, cluster (13,7,0), block (13,7,0), thread (192,0,0), device 4, sm 0, warp 6, lane 0]

#0  0x0000787e6ff09cf0 in bmm_Bfloat16_Bfloat16Bfloat16_Fp32_t128x8x128_s4_et128x8_m128x8x16_c1x1x1_16dp256b_rM_BN_transOut_schPd2x1x2x3_bN_rgTma_clmp_dynB_sm100f<<<(16,32,1),(256,1,1)>>> ()

errorpc 0x787e6ff09b20 <...+14880>
```

For another dump:

```text
[Current focus set to CUDA kernel 0, grid 2320505, cluster (5,7,0), block (5,7,0), thread (192,0,0), device 5, sm 18, warp 7, lane 0]

#0  0x00007c1dcff04970 in bmm_Bfloat16_Bfloat16Bfloat16_Fp32_t128x8x128_s5_et128x8_m128x8x16_c1x1x1_16dp256b_rM_BN_transOut_schPd2x1x2x3_bN_rgTma_clmp_dynB_sm100f<<<(16,32,1),(256,1,1)>>> ()

errorpc 0x7c1dcff047a0 <...+15008>
```

There was also an active finalize kernel:

```text
moe::dev::finalize::finalizeKernel<moe::dev::finalize::KernelParams<cutlass::bfloat16_t, cutlass::bfloat16_t, 4, true> >
```

## Things We Tried / Ruled Down

- The original SGLang hot-update packed/canonical layout issue is fixed by [sgl-project/sglang#25692](https://github.com/sgl-project/sglang/pull/25692).
- Triton MoE backend works for the same broad workload, so the issue appears specific to FlashInfer TRTLLM BF16 MoE or its integration path.
- We tried several sync/profiling/coredump variants. The visible Python-side failure location can move because CUDA errors surface asynchronously, but coredump consistently gives us the BF16 BMM kernel family above.
- Pure shape mismatch was the first bug we fixed; this remaining issue happens after that fix.

## What We Need Help With

Could you help us interpret whether this points to:

1. A FlashInfer TRTLLM BF16 MoE kernel bug for the `t128x8x128 + schPd2x1x2x3 + BN_transOut` family on SM100.
2. An input/workspace contract issue in the SGLang FlashInfer integration, e.g. invalid routing/topk/workspace/packed-weight assumptions that only surface in this kernel variant.
3. A CUDA graph or workspace lifetime issue. Some runs use cuda graph, but the coredump points into the FlashInfer BMM kernel rather than Python-side graph replay machinery.

If there is a recommended way to collect more useful debug info, e.g. a FlashInfer-specific trace, kernel argument dump, or a way to decode which input pointer/workspace address caused the MMU/device exception, we can rerun with that instrumentation.

## Artifacts

We have lightweight CUDA coredumps and decoded `cuda-gdb target cudacore` text available and can share them privately if useful. The artifact set contains:

```text
umbriel-b200-077.19283.1779946774.nvcudmp
umbriel-b200-077.19283.1779946774.target-cudacore.txt
umbriel-b200-077.19284.1779946773.nvcudmp
umbriel-b200-077.19284.1779946773.target-cudacore.txt
driver.log
output.log
```

janbernloehr:
## 📌 Description

Under CUDA-graph **piecewise** capture (one captured graph piece per decoder layer, e.g. SGLang's Piecewise CUDA Graph), the trtllm-gen fused-MoE launcher allocates its routing scratch (`expanded_idx_to_permuted_idx` and siblings) **per call** via `alloc_tensor`. Those allocations come from the graph's private memory pool, whose offsets are recycled across the separately-captured layer pieces. The routing cooperative kernel bakes the address of its in-bounds store `mPtrExpandedIdxToPermutedIdx[expandedIdx] = permutedIdx`; at replay that address aliases a **different** piece's allocation → memory corruption → a downstream trtllm-gen GEMM raises an illegal memory access. This is the `UTMALDG` fault in #3427 and the capture-time fault in #3168.

**Fix:** allocate the routing scratch from a **persistent, per-device arena** (`RoutingScratchArena`), created once and reused across all layers and graph replays, so the baked address stays valid. The scratch is pure per-call working memory (written by routing, consumed by the GEMMs in the same call), so sharing it is safe — verified bitwise-identical to per-call allocation (below). The arena is pre-sized to a generous max so it is allocated before capture and never re-allocated during capture; it costs a few MB per device.

`expanded_idx_to_permuted_idx` is returned to the caller when it is not finalized (`!do_finalize || return_activation_output`); it is kept per-call in exactly those cases (via a `mExpandedIdxReturned` flag set in `run()` before `prepare_routing()`) and shared only when it is not returned. Covers the bf16 / fp8-per-tensor / fp8-block-scale / mxint4 launchers (shared `prepare_routing_common`) and the FP4 launcher (which inlines the same allocation pattern).

**Root cause was localized on-device** by a marker-validated kernel-skip bisect on the reproducing workload (Qwen3-235B-A22B BF16, TP4, GB200/SM100, SGLang piecewise capture, `num_tokens=16384`):

| Experiment | Result |
|---|---|
| skip GEMM1 / GEMM2 / finalize individually | still IMA |
| skip routing + GEMM1 + GEMM2 + finalize | healthy |
| routing runs, all consumers skipped | **IMA** → routing itself issues the fault |
| skip `routingIndicesCoopKernel` only | healthy |
| skip only the store `mPtrExpandedIdxToPermutedIdx[expandedIdx] = permutedIdx` | healthy |
| over-allocate the buffer 4× per call | healthy |
| over-allocate 2× per call | **still IMA** (→ not a size/overrun bug) |
| allocate the same-size buffer **once, persistently** | **healthy** (→ the per-call graph-pool allocation is the problem) |

The empirical WAR boundary (`--piecewise-cuda-graph-max-tokens 8192`) equals `MaxNumTokensSingleCluster = 8192` — ≤8192 uses the single-cluster kernel (different write phase, healthy), >8192 selects the cooperative kernel (faults). Retaining the tensor's refcount does **not** help (address recycling, not object lifetime).

## 🔍 Related Issues

- Fixes #3427 (bf16, B200)
- Fixes #3168 (fp4 block-scale, clean at 8192 / corrupt at 16384)

### ✅ Pre-commit Checks

- [ ] `pre-commit run --all-files` clean

### 🧪 Tests

- Added `tests/moe/test_trtllm_gen_moe_cudagraph_capture_safety.py`: captures N *separate* single-MoE graphs sharing one mempool (the sglang piecewise-CUDA-graph structure) at a token count that selects the **cooperative** routing kernel (>8192 — the existing suite only covers ≤64 tokens, i.e. the single-cluster kernel), replays them, and asserts each replayed output matches its eager reference. This is new coverage for the coop-routing path the fix changes.
- **Scope of this test (honest):** the *full* illegal-memory-access reproduction of #3427/#3168 requires sglang's `torch.compile` piecewise-cudagraph backend; a raw `torch.cuda.graph` capture recycles the routing scratch but does not itself trigger the fault (verified on GB200 — a standalone 40-piece shared-pool capture stays clean with and without the fix). The end-to-end crash→heal was therefore validated at the **integration level** on the sglang server workload (see Reviewer Notes), and this unit test guards capture+replay correctness of the affected path rather than reproducing the sglang-specific crash. Happy to add a fuller integration repro if the maintainers have a preferred harness.

## Reviewer Notes

- **On-device validation:** on the *actual* failing workload (full 94-layer Qwen3-235B-A22B, TP4, `num_tokens=16384` piecewise capture), per-call reproduces the fault (`trtllm_fused_moe_dev_kernel.cu:991 ... illegal memory access`, all 4 TP) and the persistent arena is **healthy**. A per-call vs. persistent-arena numeric check on 16 sequential calls at 16384 tokens is **bitwise-identical** (`torch.equal`, `max_abs_diff=0`).
- **Sizing — two options, reviewer's choice.** This PR uses an internal arena pre-sized to `kMaxRoutingScratchTokens = 65536` (a few MB/device); smallest diff, no API change, matches the validated build. The fully-general alternative is a caller-provided workspace/max — the same pattern already used for `routing_replay_out`; happy to switch if preferred.
- **Concurrency.** The arena is keyed by `device_id` and mutex-guarded; it assumes routing scratch has no cross-call/cross-stream aliveness requirement (true today: written by routing, consumed by the GEMMs in the same call, on the same stream). If two-stream micro-batch overlap runs two MoE calls concurrently on one device, key by `(device_id, stream)` — a one-line change; flagged for input.
- **Scope of on-device validation:** bf16 is validated end-to-end; fp8-per-tensor / fp8-block-scale / mxint4 share `prepare_routing_common` verbatim, and FP4 receives the same edit **by code symmetry** — those precisions were not independently reproduced here.

Co-Authored with Claude Code (Fable 5).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Mixture-of-Experts routing during CUDA Graph capture and replay.
  * Prevented routing workspace reuse from causing incorrect or corrupted outputs across graph replays.
  * Preserved correct output handling when activation results are returned.

* **Tests**
  * Added regression coverage validating consistent MoE results across multiple captured and replayed CUDA graphs.

### DIFF
--- a/csrc/trtllm_fused_moe_kernel_launcher.cu
+++ b/csrc/trtllm_fused_moe_kernel_launcher.cu
@@ -22,6 +22,7 @@
 #include <iostream>
 #include <map>
 #include <memory>
+#include <mutex>
 #include <set>
 #include <tuple>
 #include <type_traits>
@@ -232,6 +233,49 @@ inline void check_gemm1_bias_mn(Optional<TensorView> const& gemm1_bias,
       << "gemm1_bias must have shape [num_tokens, top_k, 2 * intermediate_size].";
 }
 
+// Persistent, CUDA-graph-capture-safe routing scratch (flashinfer#3427, #3168).
+//
+// The trtllm-gen MoE routing scratch (expanded_idx_to_permuted_idx and its
+// siblings) used to be allocated per call via alloc_tensor(), i.e. from the
+// framework caching allocator. Under CUDA-graph *piecewise* capture (e.g. SGLang
+// PCG, one graph piece per decoder layer) those allocations come from the graph's
+// private memory pool, whose offsets are recycled across the separately captured
+// layer pieces. The routing cooperative kernel bakes the address of its in-bounds
+// store `mPtrExpandedIdxToPermutedIdx[expandedIdx] = permutedIdx`; at replay that
+// address aliases a *different* piece's allocation -> corruption -> a downstream
+// trtllm-gen GEMM raises an illegal memory access.
+//
+// The scratch is pure per-call working memory (written by routing, consumed by
+// the GEMMs within the same call), so a single persistent buffer per device,
+// allocated once and reused across all layers and graph replays, is correct and
+// keeps the baked address stable. It is pre-sized to a generous maximum so the
+// arena is allocated before capture and never re-allocated during capture (a
+// re-allocation mid-capture would re-introduce the pool-relative address). The
+// arena is a few MB per device.
+namespace {
+constexpr int64_t kMaxRoutingScratchTokens = 65536;  // upper bound on tokens processed per capture
+struct RoutingScratchSlot {
+  Tensor tensor;
+  int64_t capacity = 0;
+};
+struct RoutingScratchArena {
+  RoutingScratchSlot num_tokens_per_expert, total_num_padded_tokens, expanded_idx_to_permuted_idx,
+      permuted_idx_to_token_idx, permuted_idx_to_expanded_idx, expert_indexes,
+      expert_count_histogram, cta_idx_xy_to_batch_idx, cta_idx_xy_to_mn_limit, num_non_exiting_ctas;
+};
+std::mutex g_routing_scratch_mu;
+std::unordered_map<int, RoutingScratchArena> g_routing_scratch;  // key = device_id
+// Grow-only int32 slot; once seeded at kMaxRoutingScratchTokens it is not
+// re-allocated for any smaller/equal request, so its address is capture-stable.
+inline Tensor& ensure_routing_slot(RoutingScratchSlot& slot, int64_t numel, DLDevice dev) {
+  if (slot.capacity < numel) {
+    slot.tensor = alloc_tensor({numel}, dl_int32, dev);
+    slot.capacity = numel;
+  }
+  return slot.tensor;
+}
+}  // namespace
+
 class FusedMoeLauncher {
  protected:
   Optional<TensorView> routing_logits;
@@ -430,6 +474,11 @@ class FusedMoeLauncher {
   Tensor cta_idx_xy_to_mn_limit;
   Tensor num_non_exiting_ctas;
 
+  // Set by run() before prepare_routing(): true when expanded_idx_to_permuted_idx
+  // is returned to the caller (must stay per-call then; otherwise it may share the
+  // persistent arena). See run()'s return-array layout.
+  bool mExpandedIdxReturned = false;
+
   void* permuted_idx_to_expanded_idx_ptr() const {
     return permuted_idx_to_expanded_idx.defined() ? permuted_idx_to_expanded_idx.data_ptr()
                                                   : nullptr;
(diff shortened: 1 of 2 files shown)


csrc/trtllm_fused_moe_kernel_launcher.cu
```diff
@@ -232,6 +233,49 @@ inline void check_gemm1_bias_mn(Optional<TensorView> const& gemm1_bias,
...
+  Tensor tensor;
+  int64_t capacity = 0;
+};
+struct RoutingScratchArena {
+  RoutingScratchSlot num_tokens_per_expert, total_num_padded_tokens, expanded_idx_to_permuted_idx,
+      permuted_idx_to_token_idx, permuted_idx_to_expanded_idx, expert_indexes,
+      expert_count_histogram, cta_idx_xy_to_batch_idx, cta_idx_xy_to_mn_limit, num_non_exiting_ctas;
+};
+std::mutex g_routing_scratch_mu;
+std::unordered_map<int, RoutingScratchArena> g_routing_scratch;  // key = device_id
```
nvpohanh, line 267: I am wondering that instead of allocating this here in the .cu file, maybe FlashInfer should just expose the trtllm-gen moe scratch space API so that frameworks can allocate the scratch space and control the scratch space's lifetime? 

@leejnau could you take the action item to discuss this with FlashInfer team for the desired design? Thanks!

janbernloehr:
Closing — superseded by [#4319](https://github.com/flashinfer-ai/flashinfer/pull/4319) (the real fix: the Blackwell TMA 128 KiB-backing contract, not routing-scratch aliasing); thanks @aleozlx / @nvpohanh for the detailed root-cause work on #3427

Issue #4223, opened 2026-07-28
[Tracking][SM12x] Port latest b12x MoE and GEMM kernels for NVFP4 model targets


yichengj0:
## Background

FlashInfer's ports of the SM120/SM121 kernels from the b12x library (published as [sparkinfer](https://github.com/local-inference-lab/sparkinfer)) are 4 to 10 weeks behind upstream, which has since landed performance improvements and a correctness fix.

This issue tracks updating the kernels needed by the NVFP4 model checkpoints targeting SM120 and SM121: [Qwen3.6-35B-A3B-NVFP4](https://huggingface.co/nvidia/Qwen3.6-35B-A3B-NVFP4), [Qwen3.6-27B-NVFP4](https://huggingface.co/nvidia/Qwen3.6-27B-NVFP4), [Gemma-4-26B-A4B-NVFP4](https://huggingface.co/nvidia/Gemma-4-26B-A4B-NVFP4), [Gemma-4-31B-IT-NVFP4](https://huggingface.co/nvidia/Gemma-4-31B-IT-NVFP4), and Nemotron-family models that use relu2. b12x supports much more than these models use, so the work is scoped to their requirements rather than a full sync.

## MoE

- Precision
  - Update needed
    - NVFP4 W4A4 (Qwen3.6-35B, Gemma-4-26B, Nemotron variant)
    - NVFP4 W4A16 (Nemotron variants)
  - Not needed (no target checkpoint)
    - W8A8, W4A8, W6A8
- Activations
  - Re-apply after the update (exists only in FlashInfer)
    - gelu_tanh (Gemma-4)
  - Regression check only (supported on both sides)
    - relu2 (Nemotron)
    - silu (Qwen3.6-35B)
  - Not needed (no target checkpoint)
    - SiTU
- Expert parallelism (upstream supports it for the W4A16 family only)
  - Port the small Python orchestration module
  - Wire `expert_map` through the FlashInfer API
  - Lift the current hard reject in `b12x_fused_moe`

## GEMM

- Update or port needed
  - NVFP4 W4A4 dense GEMM (`mm_fp4(backend="b12x")`)
  - BF16 small-N GEMV (Qwen3.6 linear-attention projections): benchmark 1
    - New API, vLLM integration required
- Benchmark first (already covered by the existing CUTLASS kernel)
  - MXFP8 W8A8 (Nemotron variant): benchmark 2
- No action (no b12x kernel exists; covered by existing FlashInfer kernels)
[...]

PR #4255, opened 2026-07-30
feat(moe): sync SM12x W4A16 fused MoE family to b12x HEAD

yichengj0:
## 📌 Description

Updates the SM120/SM121 W4A16 (NVFP4 weights, bf16 activations) fused-MoE family to current [b12x](https://github.com/local-inference-lab/sparkinfer) upstream (`cc9b476`).

Changes:

- Cooperative persistent launches for the fused FC1/FC2 kernel, with occupancy-aware tile selection and launch bounds.
- A tensor-core decode path for small batches on the packed serving weights, folding the top-k sum into the FC2 store epilogue and removing a separate launch.
- Shape-stable route packing: the triton packing kernels specialize on a power-of-two capacity instead of the exact token count, so decode batch-size changes no longer recompile. The W4A16 workspaces are sized to the same capacity.
- A new `fp4_e8m0_k32` (MXFP4 K/32) weight source format, including scale-tail handling for TP shards that are not 128-aligned.

Public API and behavior changes:

- `b12x_fused_moe`'s `source_format` parameter accepts the new `fp4_e8m0_k32` value.
- W4A16 workspaces are now sized to the power-of-two route capacity, slightly larger than before. This is what lets decode batch-size changes reuse one compiled route-packing kernel instead of recompiling.
- Kernel launches now dispatch through torch custom ops, registered at module import. This brings the family in line with the library-wide convention it predated.

Upstream features with no FlashInfer target checkpoint are not included:

- NF3 (3-bit normal-float) weights. Their kernel branches stay: they compile out and keep future sync diffs clean.
- The NVFP4+NF3 hybrid entry points.
- The native-ModelOpt small-batch micro kernel.
- The SiTU and swiglu-oai activations.

## 📊 Performance

Speedup over the previous kernels at the MoE shape of a Nemotron variant, with relu2 and silu activations, run under CUDA graphs, median of three runs. Each cell reads RTX 5080 / RTX Pro 6000 Server Edition / GB10:

| Activation | m=1 | m=4 | m=16 | m=64 | m=2048 |
|---|---|---|---|---|---|
| relu2 | 1.07x / 1.15x / 1.11x | 1.03x / 1.07x / 1.04x | 1.00x / 1.00x / 1.02x | 0.99x / 1.00x / 1.03x | 12.91x / 18.52x / 1.00x |
| silu | 0.98x / 1.23x / 1.10x | 0.99x / 1.05x / 1.02x | 0.95x / 0.98x / 1.01x | 0.95x / 0.98x / 1.00x | 11.70x / 16.88x / 1.02x |

The m=2048 cells fix a stall rather than measure a general kernel win: the fused kernel uses grid-wide barriers but was not launched cooperatively, so at large m CTAs could spin waiting for peers that were never co-scheduled. GB10 kept the whole grid resident and never hit the stall, so its m=2048 cells have no win to inherit.

## 🔍 Related Issues

#4223 (item 3).

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

- `tests/moe/test_b12x_fused_moe.py` (142 cases) passes on SM121, both on the default dispatch and with the W4A16 path forced.
- Added `tests/moe/test_b12x_w4a16_route_pack.py` (64 cases): route packing against a host reference, covering `expert_map`, invalid expert ids, preallocated buffers, and the workspace capacity contract.
- Weight preparation is bit-identical to the previous implementation across the supported source formats and activations.

## Reviewer Notes

- Most of the diff is upstream code taken as is. FlashInfer-local changes are limited to the import remaps, the two support modules, the API-boundary checks, the workspace sizing in `moe_dispatch`, and the decode-path cap below.
- One tuning deviation from upstream: the tensor-core decode path is selected up to m = 4 instead of upstream's m = 8. The crossover to the route-packed GEMM varies by card, so the cap conservatively stops where the decode path wins or ties on every card measured.
- The fused kernel compiles at DSL optimizer level 3 instead of upstream's 2. Level 3 generates a faster mainloop on the DSL version pinned here, at the cost of a longer compile for this kernel.
- The kernels import activation metadata and a kernel compile cache from upstream's library. The two new modules, `moe_w4a16_activations.py` and `moe_w4a16_compiler.py`, are small local implementations of those two pieces.

## Summary by CodeRabbit

* **New Features**
  * Expanded Blackwell W4A16 fused MoE support for NF3, NVFP4, ModelOpt, CompressedTensors, and FP4 E8M0 formats.
  * Added configurable activations, SwiGLU parameters, routing, calibration, and launch options.
  * Improved route packing for large workloads, dynamic capacities, expert remapping, and invalid route IDs.
  * Added optimized quantization, dequantization, conversion, and matrix-operation support.
  * Added compile caching to reduce repeated kernel compilation.

* **Bug Fixes**
  * Improved workspace sizing and safe handling of zero-valued scales.

* **Tests**
  * Added coverage for route grouping, padding, capacity fallbacks, large shapes, and expert mapping.


flashinfer/fused_moe/cute_dsl/blackwell_sm12x/moe_w4a16_kernel.py
```diff
@@ -3563,46 +5852,626 @@ def compile_w4a16_topk_sum(
...
+    grid_x = fc1_mn_tiles // waves
+    if grid_x < 1 or grid_x > cap:
+        return max(cap, 1)
+    return grid_x
+
+
+@torch.library.custom_op(
+    "flashinfer::w4a16_fused_moe_launch",
+    mutates_args="unknown",
+)
```
bkryu, line 6101: These three ops register as import-time side effects with fixed global names. Since flashinfer doesn't appear to have a double-registration  guard in fused_moe/, worth confirming this module is imported exactly once — a re-import under a different path (editable-install quirk) would   hit torch's "operator already exists" error. Also mutates_args="unknown" is the safe-but-pessimistic choice; fine, just flagging it disables    some torch.compile fusion around these ops.

flashinfer/fused_moe/cute_dsl/blackwell_sm12x/moe_w4a16_kernel.py
```diff
@@ -3563,46 +5852,626 @@ def compile_w4a16_topk_sum(
...
+    tc_decode_fused_sum: bool,
+    stream_int: int,
+) -> None:
+    return None
+
+
+@torch.library.custom_op(
+    "flashinfer::w4a16_fused_moe_calibrated_launch",
+    mutates_args="unknown",
+)
```
bkryu, line 6266: Same here

flashinfer/fused_moe/cute_dsl/blackwell_sm12x/moe_w4a16_kernel.py
```diff
@@ -3563,46 +5852,626 @@ def compile_w4a16_topk_sum(
...
+        ),
+        m,
+        cuda.CUstream(stream_int),
+    )
+
+
+@torch.library.custom_op(
+    "flashinfer::w4a16_topk_sum_launch",
+    mutates_args=("output",),
+)
```
bkryu, line 6456: Same here

bkryu (commented):
Thanks @yichengj0 left a few comments!

flashinfer/fused_moe/cute_dsl/blackwell_sm12x/moe_w4a16_kernel.py
yichengj0: Confirmed the import side: this module is only reachable through one path, so registration runs once per process. `flashinfer::fp4_quantize` is the only other live registration in the library and works the same way. On `mutates_args`, agreed: adb976a5 replaces "unknown" with the actual lists.

flashinfer/fused_moe/cute_dsl/blackwell_sm12x/moe_w4a16_compiler.py
```diff
@@ -0,0 +1,110 @@
...
+        compiled = _COMPILE_CACHE.get(cache_key)
+    if compiled is not None:
+        return compiled
+
+    compile_callable = cute.compile
+    if dsl_compile_options is not None:
+        compile_callable = compile_callable[dsl_compile_options]
+    compiled = compile_callable(func, *args, **kwargs)
+    with _COMPILE_CACHE_LOCK:
+        _COMPILE_CACHE[cache_key] = compiled
```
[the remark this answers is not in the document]
yichengj0: Compiling the same kernel twice gives the same result, so a concurrent miss only costs a wasted compile. `clear_compile_cache()` is a test helper. Keeping the simple lock.

flashinfer/fused_moe/cute_dsl/blackwell_sm12x/moe_w4a16_fp4_helpers.py
```diff
@@ -197,6 +216,125 @@ def relu2_quantize_grouped_nvfp4_torch(
...
+    )
+    for group_idx in range(num_groups):
+        valid_rows = int(row_counts[group_idx].item())
+        if valid_rows == 0:
+            continue
+        x = input_tensor[group_idx, :valid_rows].float()
+        blocked = x.view(valid_rows, cols // MX_SF_VEC_SIZE, MX_SF_VEC_SIZE)
+        block_max = blocked.abs().amax(dim=-1, keepdim=True)
+        rounded, byte = pow2_ceil_ue8m0_torch(block_max * _INV_FLOAT8_E4M3_MAX)
+        inv = _ue8m0_output_scale_torch(byte)
```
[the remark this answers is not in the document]
yichengj0: Renamed in adb976a5.

tests/moe/test_b12x_w4a16_route_pack.py
```diff
@@ -0,0 +1,331 @@
...
+    monkeypatch.setattr(moe_dispatch, "get_num_sm", lambda device: 120)
+    slots, blocks, _, _, _ = moe_dispatch._w4a16_workspace_geometry(
+        routed_rows=routed_rows,
+        route_num_experts=num_experts,
+        num_topk=topk,
+        k=4096,
+        n=4096,
+        is_gated=True,
+        device=torch.device("cuda"),
+    )
```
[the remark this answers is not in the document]
yichengj0: Added in adb976a5.

aleozlx (approved):

Issue #4254, opened 2026-07-29
[CAKE] Long-term CAKE-generated Kernel Progress Tracker


yyihuang:
To track multi-platform kernels generated from CAKE 🍰.

We are more than happy to take any new kernel challenges!

**Feel free to reach out/leave your comments/create sub-issues here to submit your customized CAKE kernel request. And we will bake it for you!**

**We know many of you have been asking for training kernels—and we heard you.** You have had your CAKE going forward; now it is time to take it **back**! Follow the dedicated forward & backward tracker at #4642, bring us your toughest training bottleneck, and let us bake the full pass together. 🍰🔥

### Communication Kernel

- #4722 
- #4730 — opt-in SM100/SM103 fused MoE all-reduce (TP2/TP4/TP8); B200 TP8 `1.096503x` GM, 46/48 wins; 32 synccheck + 32 racecheck, zero errors
- #4784
- #4822 
- #5116 — parallelize SM103 TP8 packed QKV peer traversal (draft)
- #4875

#### Prefill

- #4262
- #4313
- #4351
- #4675
- #4728 
- #4845  2.82x vs flashkda

#### Decode

- #4279
- #4314
- #4562
- #4967 — fused KDA decode backend for B200 and B300, with FP32/BF16 recurrent state

#### Prefill and decode

- #4445
- #4535
- #4571

#### Context-parallel prefill

- #4539, from #4078

#### Genreal update

- #4572 — honor intermediate cache step strides
- #4745 

#### Source-only Blackwell backend

- #4581 — source-only generated backend for SM100 and SM103

#### Blackwell all-gather matmul

- #4722 — opt-in generated backend for SM100 and SM103, world sizes 2 and 4
- #5076 

#### BF16 BMM

- #4370 — opt-in generated backend for SM100a and SM103a

#### Deepgemm - FP8

- #4564 

#### TinyGEMM2

- #4274

#### Blackwell Router GEMM

- #4594

### AlphaMoE

- #4287
- #4339
- #4340

### MegaMoE

- #4810
- #5126
- #4819 
- #4970 

### BGMV MoE

- #4821 — prepared Blackwell backend for SM100

### DeepSeek Fused Routing

- #4587 — current Cake/FlashInfer: B200 1.0194–1.2101x and GB300 1.2665–1.5738x (15/15 each)

### MXFP8 Quantization

- #4820 — opt-in generated backend for grouped MXFP8 quantization on SM100 and SM103

[...]

PR #4262, opened 2026-07-30
feat(cake_kda): add optimized B200 recurrent prefill backend

yyihuang:
## 📌 Description

This PR adds an CAKE-generated optimized SM100a BF16 recurrent-KDA prefill backend and
dispatches to it through the existing `recurrent_kda` API.

The new backend coexists with the current CuTe-DSL implementation. Decode
(`T=1`), speculative decode, GQA, state-pool/checkpoint features, unsupported
layouts and dtypes, and non-B200 devices continue to use the existing backend
unchanged.

### Dispatch contract

The optimized path is selected only when all of the following hold:

- the device is exactly compute capability 10.0 (B200 / SM100a);
- the call is ordinary multi-token prefill: fixed `T > 1`, or packed input
  with more tokens than sequences;
- Q, K, V, and G are contiguous BF16 `[B, T, H, 128]` tensors with a shared
  head count, and beta is contiguous BF16 `[B, T, H]`;
- `A_log` is contiguous FP32 `[H]`, while `dt_bias` is contiguous FP32
  `[H, 128]` or `[H * 128]`;
- `use_qk_l2norm_in_kernel=True`, `use_gate_in_kernel=True`,
  `beta_is_logit=True`, and `lower_bound` is finite and negative;
- speculative decode, GQA, state indices, committed-state sources, and
  accepted-token/checkpoint features are disabled.

Packed input uses `B=1` and contiguous CUDA int32/int64 `cu_seqlens`. The
binding consumes int64 offsets; CUDA graph capture therefore requires callers
to supply int64 offsets directly. An optional contiguous CUDA int32
`seq_order` can order packed sequences for better tail utilization.

Calls that do not exactly match this contract fall back to the existing
CuTe-DSL path.

### Physical schedules and true in-place state update

- M64 is used only for fixed-layout `B=1, H=64`.
- M128 handles every other eligible fixed-layout call and all eligible packed
  calls.
- State layout remains BF16 `[N, H, V, K]` with `V=K=128`.
- A supplied `initial_state` is passed as both the initial- and final-state
  kernel pointer. The kernel updates that allocation directly; there is no
  state scratch allocation and no post-kernel `copy_`/memcpy.
- M128 assigns one CTA to each `(sequence, head)`. M64 assigns two CTAs whose
  64-row value partitions are disjoint. Each CTA loads all initial-state rows
  it owns before storing those same final-state rows, making exact pointer
  aliasing safe.
- The binding accepts either disjoint state buffers or an exact full-range
  alias and rejects partial overlap.
- When `initial_state=None`, final-state storage is created only if
  `output_final_state=True`.

The generated source emits `__restrict__` for every pointer and has no
per-argument alias escape hatch, so the removal of `__restrict__` from only the
two state parameters is an explicitly delimited FlashInfer ABI integration
patch. The frozen-source test restores those two qualifiers, removes the
existing tensor-map acquire integration prologue, and verifies the remaining
generated source against its normalized SHA256.

### CUDA graph and tensor-map safety

`RecurrentKDAPrefillWorkspace` owns optional final-state scratch for calls
without an initial state, beta padding, and separate 768-byte M64/M128
tensor-map descriptor blocks. It binds to one stream, is warmed eagerly
against the exact tensor signature, and is used by one captured
`recurrent_kda` invocation. A workspace that has participated in capture is
rejected by later Python calls; graph replay remains valid because it does not
re-enter Python.

Tensor maps are prepared outside capture and published to stable global
storage on the caller stream. Every consumer CTA executes
`fence.proxy.tensormap::generic.acquire.gpu` for all six maps, followed by a
CTA barrier, before any TMA instruction can consume them.

### Other integration

- Adds JIT and AOT registration for separate M64 and M128 translation units.
- Updates tracing and API documentation for `seq_order` and
  `prefill_workspace`.
- Adds correctness, routing/fallback, ABI, stream, overlap, workspace, and
  CUDA graph tests.
- Adds a CUPTI cold-L2 benchmark for the six fixed/packed contract shapes,
  including a commit- and binary-verified MoonshotAI/FlashKDA peer.

## 🧪 Validation

Final validation was run on pushed head
`f6c3b0c486787e64f76b2be91ddb239ea3ba66ce`, on an NVIDIA B200
(compute capability 10.0), PyTorch `2.13.0.dev20260503+cu132`, CUDA 13.2,
and Python 3.13.13.

| Gate | Result |
| --- | --- |
| Frozen-source/JIT specification tests | 8 passed |
| Full targeted recurrent-KDA prefill suite, including fixed/packed correctness, M64/M128, non-default streams, and CUDA graph capture/replay | 28 passed |
| Focused M64/M128 `compute-sanitizer --tool memcheck` | 5 passed; `ERROR SUMMARY: 0 errors` |
| Focused M64/M128 `compute-sanitizer --tool synccheck` | 5 passed; `ERROR SUMMARY: 0 errors` |
| Pre-commit hooks on the change | Passed |

The full repository test suite was not run.

### Fresh B200 CUPTI benchmark

The candidate is invoked only through the public
`flashinfer.kda_decode.recurrent_kda` entry point and performs its state update
in place inside the kernel. Its timed region contains no state memcpy.

The comparison uses six fixed/packed shapes with deterministic seeds, matching
tensor distributions, BF16 state, and `scale=1/sqrt(128)`. Preinitialized
rotating state buffers ensure each timed candidate invocation receives the
same initial state. Allocation, metadata creation, sequence ordering,
state-pool reset, and JIT/cache warmup are outside the measured region.

The peer is `MoonshotAI/FlashKDA._fwd_raw`, from source commit
[`d2ff19a6`](https://github.com/MoonshotAI/FlashKDA/commit/d2ff19a6a0c82f39f796f637ebd1c36090b1268f)
and CUTLASS
[`5c149f52`](https://github.com/NVIDIA/cutlass/commit/5c149f52a436782210263fb2f19b354443a61c6a).
The loaded extension SHA256 is
`997c3a1d1338f8bf9dba3c1a01386b1b74448214c294d64409454cc11141c04c`.
The benchmark records the source revision and independently computed extension
digest.

Timing uses CUPTI activity tracing, cold-L2 flushing, no CUDA graph, 20 ms
warmup, and 100 ms measurement per block. Each value is the median of two
independent block medians; the reported scopes occupy symmetric
`PR/raw/.../raw/PR` positions. Speedup is `FlashKDA / this PR`.

| Case | Layout / sequence lengths | Variant | This PR public in-place API (ms) | FlashKDA raw peer (ms) | Speedup |
| --- | --- | --- | ---: | ---: | ---: |
| H=96 fixed | `[8192]` | M128 | 0.506095 | 1.085277 | **2.1444x** |
| H=96 mixed | `[1300, 547, 2048, 963, 271, 3063]` | M128 | 0.391135 | 0.894702 | **2.2875x** |
| H=96 uniform | `[1024] × 8` | M128 | 0.434992 | 0.735535 | **1.6909x** |
| H=64 fixed | `[8192]` | M64 | 0.469711 | 0.992638 | **2.1133x** |
| H=64 mixed | `[1300, 547, 2048, 963, 271, 3063]` | M128 | 0.269903 | 0.676574 | **2.5067x** |
| H=64 uniform | `[1024] × 8` | M128 | 0.292567 | 0.495871 | **1.6949x** |
| **Geometric mean** | | | | | **2.0512x** |

All six output/state comparisons against FlashKDA passed at
`atol=rtol=1e-2`; the maximum observed absolute errors were `0.0009765625`
for output and `0.015625` for state. The benchmark JSON SHA256 is
`c34b27f9e8fb4e4dfbc500e6976e63668fcd0bff2bb00086cd516de59491f259`.

### Upstream recurrent-KDA comparison

Pinned upstream main does not implement ordinary multi-token recurrent-KDA
prefill through its public API: fixed `T != 1` is rejected, and packed standard
decode processes one token per sequence. Speculative decode has different
checkpoint/state semantics and is not an equivalent workload. The upstream
prefill comparison is therefore **unsupported / N/A**, rather than timing a
different operation.

## 🚀 Pull Request Checklist

- [x] Tests were added or updated.
- [x] Targeted tests and pre-commit hooks pass.
- [ ] Full repository test suite was not run.

## Reviewer Notes

The main review surfaces are:

1. strict eligibility and unchanged fallback behavior;
2. exact-alias state ABI and direct in-kernel update;
3. workspace lifetime, stream binding, and CUDA graph rules;
4. tensor-map publication/acquire ordering;
5. the two explicitly delimited frozen-CUDA integration patches and their
   integrity normalization.

Related to #4254

## Summary by CodeRabbit

* **New Features**
  * Added optimized recurrent Kimi Delta Attention prefill support for eligible B200/SM100a configurations.
  * Added M64 and M128 execution variants with eager execution and CUDA graph compatibility.
  * Added a phase-neutral `recurrent_kda` API supporting decode, speculative decode, and prefill workflows.
  * Added reusable prefill workspace support and public package exports.

* **Documentation**
  * Added API documentation covering prefill behavior, dispatch, workspaces, and supported configurations.

* **Tests**
  * Added comprehensive coverage for validation, routing, numerical correctness, workspace reuse, and CUDA graphs.


yzh119 (approved):

yiakwy-xpu-ml-framework-team:
Hi @yyihuang after checking the code, I found Cake backend is more like a TVM-FFI jit module warpper to compile the CUDA sources. 

We are actually did this in flash float : https://github.com/yiakwy-xpu-ml-framework-team/flash-float-jit-kernels

> flashFloat : Pure cuda with zero dependency of Cutlass and cutile; This is follow up of original design of flashinfer. 

Is that true ?

Note my motivation is that after trying to support many subprojects and I found CUDA is the best cross platform languages so that I can support for different devices.

PR #4263, opened 2026-07-30
fix(quantization): use 64-bit row addressing in per-token NVFP4 quantizer

S1ro1:
## 📌 Description

The per-token NVFP4 quantizer formed row views with `mInput[row_idx, None]` and `mOutput[row_idx, None]`. For sufficiently large 2D tensors, CuTe folds the row element offset into an Int32 value. With `K=2688`, rows at and above `ceil(2**31 / 2688) = 798916` cross that range.

This is reached by the padded GEMM1 intermediate for NVIDIA Nemotron 3 Super:

| Row tile | Intermediate shape | Elements | Result before this change |
|---:|---:|---:|---|
| 128 | `785920 x 2688` | 2,112,552,960 | Passes |
| 256 | `851456 x 2688` | 2,288,713,728 | CUDA address error |

The change constructs input and output row views from explicit 64-bit byte addresses. Quantization math, layouts, alignment, output format, and the public API are unchanged.

It also adds a focused GPU regression test on both sides of the boundary and with both default and explicitly disabled PDL.

## 🧪 Validation

The equivalent wheel-compatible patch was validated on GB200 with:

- all eight relevant fused-MoE configurations, with PDL enabled and disabled;
- quantizer-only probes at `M=785920`, `798848`, `799040`, and `851456`, with exact per-token scale checks;
- one-token, partial-routing, and full-routing cases;
- a complete autotuning pass over token buckets from 1 through 32768;
- randomized-data numerical comparisons.

The 128-row configurations remained within benchmark noise. At 32768 tokens, enabling the valid 256-row configurations improved the best measured time from 10.61 ms to 9.93 ms.

Targeted pre-commit hooks pass for both changed files.

## Reviewer Notes

The linear and swizzled NVFP4 quantizers contain similar row-slice expressions, but their currently supported shapes do not approach this address range. They are intentionally left out of this focused change and can be handled separately if desired.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved NVFP4 per-token quantization for very large tensors, preventing incorrect memory access when row offsets exceed 32-bit limits.
  * Ensured quantization results remain accurate for large workloads.

* **Tests**
  * Added regression coverage for row addressing beyond the 32-bit boundary, including validation of scales and packed FP4 output.

### DIFF
--- a/flashinfer/quantization/kernels/nvfp4_quantize.py
+++ b/flashinfer/quantization/kernels/nvfp4_quantize.py
@@ -39,7 +39,7 @@ import cutlass.cute.nvgpu.cpasync as cpasync
 import cutlass.pipeline as pipeline
 from cutlass.pipeline import pipeline_init_arrive, pipeline_init_wait
 import torch
-from cutlass import Float32, Int32, Uint8
+from cutlass import Float32, Int32, Int64, Uint8
 
 from ...api_logging import flashinfer_api
 from ...jit.cute_dsl_core import build_and_load_cute_dsl_kernel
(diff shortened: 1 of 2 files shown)


aleozlx:
>The 128-row configurations remained within benchmark noise. At 32768 tokens, enabling the valid 256-row configurations improved the best measured time from 10.61 ms to 9.93 ms.

could you elaborate on this? e.g. what benchmark have you already run? what shapes

also here is some relevant benchmark utility in FI: https://github.com/flashinfer-ai/flashinfer/tree/main/benchmarks#quantization-flags

cc @bkryu

aleozlx (commented):
code review: lgtm

S1ro1:
> > The 128-row configurations remained within benchmark noise. At 32768 tokens, enabling the valid 256-row configurations improved the best measured time from 10.61 ms to 9.93 ms.
> 
> could you elaborate on this? e.g. what benchmark have you already run? what shapes
> 
> also here is some relevant benchmark utility in FI: https://github.com/flashinfer-ai/flashinfer/tree/main/benchmarks#quantization-flags
> 
> cc @bkryu

Yeah the benchmark was based on those files, for shapes I don't remember exact ones (can gather it), but it was mostly shapes that would previously fail and otherwise Nemotron 3 Super shapes with various token lengths. Can rerun a larger suite if needed

aleozlx (approved):

PR #4270, opened 2026-07-30
Fix silent wrong results in the cuDNN SDPA prefill/decode paths

YangXu1990uiuc:
# Fix silent wrong results in the cuDNN SDPA prefill/decode paths

## Summary

Seven defects in the cuDNN SDPA integration cause silently wrong attention outputs — no
exception, no IMA, just wrong numbers. The two most severe are graph-cache key omissions:
the prefill key ignored every tensor stride and the decode key ignored dtype entirely, so a
second call that differs only in layout or precision replays the first call's compiled graph
against incompatible pointers. Decode additionally hardcoded `BFLOAT16` and a contiguous
stride tuple, so fp16 decode never worked at all, and it silently discarded caller-supplied
ragged offsets. The cuDNN paths now key on everything the built graph bakes in, derive dtype
and strides from the actual tensors, and reject the configurations they cannot honour.

### 1. Prefill graph-cache key omitted every tensor stride

`flashinfer/cudnn/prefill.py:175` (`_sdpa_prefill_key_fn`).

`_build_prefill_graph` bakes each tensor's strides into the graph via `set_stride()` on the
q/k/v tensor descriptors — cuDNN compiles a kernel for that exact layout. The cache key
carried shapes but no strides, so two calls with identical shapes and different layouts
(e.g. contiguous K/V vs. two views into a single interleaved KV buffer) shared one compiled
graph. The second call runs the first layout's kernel against the second layout's pointers.

The key now includes `q.stride()`, `k_cache.stride()`, `v_cache.stride()`, the K/V dtypes
(K/V may carry a dtype of their own, e.g. an FP8 KV cache, and the graph's data types come
from those tensors rather than from `q` alone), `o_data_type`, and the presence flags for the
`batch_offsets_*` / `actual_seq_lens_*` tensors — each of these changes the built graph's
structure via `set_ragged_offset()` or the padding-mask wiring.

### 2. Decode graph-cache key had no dtype at all

`flashinfer/cudnn/decode.py:53` (`_sdpa_decode_key_fn`).

The key was `(max_sequence_kv, q.shape, k_cache.shape)` plus `scale`. A bf16 call and an fp16
call at the same shapes shared one graph. This is worse than a stale replay: it bypasses
cuDNN's own support check, so a configuration cuDNN explicitly refuses on a cold cache
executes anyway against the other dtype's graph. Added dtypes, strides, `v_cache.shape`,
`block_size` (the page size is a graph constant) and the `batch_offsets_*` presence flags.

### 3. Decode hardcoded BFLOAT16 and contiguous strides

`flashinfer/cudnn/decode.py:145-150` and `:209-212`.

The Q tensor was built with `data_type=cudnn.data_type.BFLOAT16` and a hand-written
contiguous stride tuple `(h_qo * d_qk, d_qk, d_qk * h_qo, 1)`, ignoring both `q.dtype` and
`q.stride()`; the O tensor was likewise pinned to `BFLOAT16`. With an fp16 paged K/V cache
the resulting graph is mixed-dtype, so cuDNN rightly refuses to build it — meaning fp16
decode never worked on a cold cache, and only "succeeded" warm by colliding with a cached
bf16 graph (finding 2). Now uses `cudnn.datatypes._torch_to_cudnn_data_type(q.dtype)` and the
real `q.stride()`, exactly as `prefill.py` already does.

### 4. Decode silently discarded caller-supplied ragged offsets

`flashinfer/cudnn/decode.py:120-122` (before the fix) set `batch_offsets_q = None;
batch_offsets_o = None` with the comment *"WAR: override batch offsets for now, as it leads
to a poor performance"*. The graph's Q/O descriptors assume the packed
`(batch, heads, head_dim)` layout, so any caller passing offsets that describe a different
layout got wrong rows read and written with no diagnostic. The non-cuDNN fallback path *does*
honour the offsets, so the same call returned different results depending on whether
`cudnn` was importable.

**Judgement call — conservative option chosen.** Rather than re-enabling ragged offsets (a
performance change I cannot measure here) or keeping the silent drop, the cuDNN path now
honours `batch_offsets_q`/`batch_offsets_o` (the graph binds them as ragged offsets), keys the
graph on their shape/dtype/stride rather than mere presence, and rejects `batch_offsets_k`/
`batch_offsets_v`, which decode cannot use because the KV cache is addressed via `block_tables`
(`flashinfer/cudnn/decode.py:243-250`). This is a deliberate, visible API break: the two in-repo
callers (`tests/attention/test_cudnn_decode.py`, `benchmarks/routines/attention.py`) passed
the *trivial* packed offsets, for which dropping them was a no-op, so they simply stop passing
them. Also fixed the copy/paste in `_batch_decode_with_kv_cache` that forwarded
`batch_offsets_q` as `batch_offsets_o`.

### 5. Token-unit ragged-offset multiplier assumed a packed layout

`flashinfer/cudnn/prefill.py:361`, `:390`, `:408`.

On the direct `cu_seq_lens` path the `batch_offsets_*` are token-unit indptrs and cuDNN scales
them to element offsets with `set_ragged_offset_multiplier()`. The multiplier was hardcoded to
the *packed* element count (`h_qo * d_qk`, `h_kv * d_qk`, `h_kv * d_vo`) instead of the
tensor's own token stride, which the same code already unpacks as `s_stride` and passes to
`set_stride()` two lines above. With K/V as views into one interleaved KV buffer the token
stride is `2 * h_kv * d`, so every request after the first addressed the wrong tokens. This is
an independent bug from finding 1: it reproduces on a cold cache.

### 6. Decode block-table stride used floor division

`csrc/cudnn_sdpa_kernel_launcher.cu:1103`.

`num_pages_per_seq` was computed as `max_s_kv / page_size` while callers allocate the block
With `b=8, page_size=16, max_sequence_kv=1000` the launcher passes 62 while the table has 63
columns, so request `i` reads its page list at `i*62` — every batch element except 0 attends
over the wrong KV pages. Silent, no IMA, because the reads stay in bounds of the allocation.
This path is reached only when the `cudnn` Python package is unavailable (the cubin fallback).

### 7. Prefill LSE tensor allocated with `torch.empty`

`flashinfer/cudnn/prefill.py:902`. cuDNN writes only the rows covered by `actual_seq_len`, so
the padded rows of a dense LSE buffer are undefined. The allocation deliberately KEEPS
`torch.empty` (zero is not a neutral logsumexp and it would not cover caller-supplied
buffers); the constraint is documented instead, and the tests compare only valid rows.

## Evidence

Prefill stride collision, H100, bf16, B=2 s_q=64 s_kv=128 h=8 d=128. Config A = contiguous K/V;
config B = K and V as views into one interleaved KV buffer (token stride 2*h_kv*d). Each run in a
fresh process; error is max_abs against an fp32 reference (|ref|max = 1.52):

| run | order | err | checksum |
| --- | --- | --- | --- |
| cold | B alone | 0.0040 | 9565056 |
| control | B then B | 0.0040 | 9565056 |
| warm | A then B | 1.8333 | -13264255 |
| reverse | B then A | 1.9641 | 20915960 |

The control (same config twice) rules out "any second call is broken"; the collision is symmetric.
After the fix all four runs return 0.0040 with identical checksums.

Decode dtype, H100, paged KV, B=2 s_kv=128 h=8 d=128 page=16:

| cold | fp16 alone | cuDNN NOT_SUPPORTED (correctly refused) |
| --- | --- | --- |
| control | fp16 then fp16 | NOT_SUPPORTED |
| warm | bf16 then fp16 | SUCCEEDS, returns garbage (err 3.35 vs \|ref\|max 0.47) |

A configuration cuDNN explicitly rejects executes anyway against the cached bf16 graph. After the
key fix all three correctly raise. Note the deeper cause is Fix 3: with Q hardcoded to BFLOAT16
while the paged K/V cache is fp16, the graph is mixed-dtype, so cuDNN rightly refuses on a cold
cache — meaning fp16 decode never worked at all, warm or cold.

With findings 2 and 3 applied, fp16 decode now works, measured on the same box against an fp32
page-gather reference (B=2, h=8, d=128, page=16, s_kv=128):

| run | err | \|ref\|max |
| --- | --- | --- |
| bf16 (fresh process) | 0.00131 | 0.836 |
| fp16 after bf16, same process | 0.00017 | 0.469 |
| fp16 alone (fresh process) | 0.00022 | 0.469 |

Finding 5, same box, bf16, B=2 s_q=64 s_kv=128 h=8 d=128, `batch_offsets_units="tokens"`,
max_abs vs. an fp32 reference (|ref|max = 1.10):

| config | before | after |
| --- | --- | --- |
| A, contiguous K/V | 0.0031 | 0.0031 |
| B, interleaved KV views, cold process | 1.1871 | 0.0031 |
| A then B | 1.1871 | 0.0031 |

## Testing

Run on one H100 NVL (sm90), cuDNN 9.24.0, `CUDA_HOME` set so the JIT/cubin path builds:

```
pytest tests/attention/test_cudnn_graph_cache_key.py \
       tests/attention/test_cudnn_prefill_token_indptr.py \
       tests/attention/test_cudnn_decode.py
157 passed in 56.62s
```

Baseline for comparison, pristine tree, same box and same `CUDA_HOME`:
`tests/attention/test_cudnn_decode.py` -> `12 passed`.

Measured evidence for the two cache-key fixes is in the Evidence section above; both were
reproduced with cold/warm/control/reverse runs in separate processes.

Not covered by this PR: a `num_pages_per_seq` floor-vs-ceil discrepancy in
`csrc/cudnn_sdpa_kernel_launcher.cu:1103` relative to `prefill()` at `:576`. It was dropped
here because the decode tests pass with the current floor division, i.e. nothing in the suite
exercises it, so the change would have been unproven. It will be raised separately with a test
that forces the cubin fallback.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cuDNN SDPA graph caching to avoid stale reuse across differing tensor dtypes, strides, and KV/cache layouts (including scale/optional ragged inputs).
  * Corrected cuDNN decode batch-offset wiring and output dtype selection to match the query tensor.
  * Improved cuDNN prefill correctness for strided KV and tightened behavior for unsupported batch-offset inputs.

* **Tests**
  * Expanded cuDNN decode/prefill coverage for larger sequence lengths, strided KV views, packed batch offsets, and key/stride/dtype cache key differences.
  * Added checks that unsupported KV batch-offset inputs raise `NotImplementedError`.

### DIFF
--- a/flashinfer/cudnn/decode.py
+++ b/flashinfer/cudnn/decode.py
@@ -50,6 +50,13 @@ class UIDs(Enum):
     STATS_UID = 1001  # Stats tensor
 
 
+def _tensor_desc(t: Optional[torch.Tensor]):
+    """Everything about a tensor that tensor_like() bakes into the built graph."""
+    if t is None:
+        return None
+    return (tuple(t.shape), t.dtype, tuple(t.stride()))
+
+
 def _sdpa_decode_key_fn(
     q: torch.Tensor,
     k_cache: torch.Tensor,
@@ -69,15 +76,33 @@ def _sdpa_decode_key_fn(
         max_sequence_kv,
         tuple(q.shape),
         tuple(k_cache.shape),
+        tuple(v_cache.shape),
         # attn_scale is baked into the built graph as a compile-time constant;
         # omitting it silently replays a stale-scale graph on same-shape calls.
         scale,
+        # The graph's cuDNN data types come from these tensors. Without them a
+        # bf16 call and an fp16 call at the same shapes share one graph, which
+        # also bypasses cuDNN's own support check: a dtype cuDNN rejects on a
+        # cold cache executes anyway against the other dtype's graph.
+        q.dtype,
+        k_cache.dtype,
+        v_cache.dtype,
+        # strides are baked into the graph via set_stride()
+        q.stride(),
+        k_cache.stride(),
+        v_cache.stride(),
+        # page size is a graph constant
+        block_size,
         # These presence flags change the built graph's structure (padding
         # mask, paged tables) the same way: same-shape calls that differ only
         # in them must not share a graph.
         actual_seq_lens_q is not None,
         actual_seq_lens_kv is not None,
         block_tables is not None,
+        # The ragged-offset tensors are described to cuDNN with tensor_like(), so
+        # their shape/dtype/stride are baked into the graph, not just their presence.
+        _tensor_desc(batch_offsets_q),
+        _tensor_desc(batch_offsets_o),
     )
 
 
@@ -101,14 +126,13 @@ if CUDNN_AVAILABLE:
     ):
         handle = _create_cudnn_handle(torch.cuda.current_stream())
 
-        # WAR: override batch offsets for now, as it leads to a poor performance
-        batch_offsets_q = None
-        batch_offsets_o = None
-
         with cudnn.graph(handle) as (g, _):
             if q.dim() == 3:
                 s_qo = 1
                 b, h_qo, d_qk = q.shape[0], q.shape[1], q.shape[2]
+                b_stride, h_stride, d_stride = q.stride()
+                # s_qo == 1, so the s stride is never dereferenced
+                s_stride = b_stride
             elif q.dim() == 4:
                 b, h_qo, s_qo, d_qk = (
                     q.shape[0],
@@ -116,6 +140,7 @@ if CUDNN_AVAILABLE:
                     q.shape[2],
                     q.shape[3],
                 )
+                b_stride, h_stride, s_stride, d_stride = q.stride()
             else:
                 raise ValueError(f"q must have 3 or 4 dimensions, got {q.dim()}")
 
@@ -124,11 +149,13 @@ if CUDNN_AVAILABLE:
 
             d_vo = v_cache.shape[3]
 
+            cudnn_q_data_type = cudnn.datatypes._torch_to_cudnn_data_type(q.dtype)
+
             cudnn_q = g.tensor(
                 name="q",
                 dim=(b, h_qo, s_qo, d_qk),
-                stride=(h_qo * d_qk, d_qk, d_qk * h_qo, 1),
-                data_type=cudnn.data_type.BFLOAT16,
+                stride=(b_stride, h_stride, s_stride, d_stride),
+                data_type=cudnn_q_data_type,
             )
             if batch_offsets_q is not None:
                 ragged_q = g.tensor_like(batch_offsets_q)
@@ -191,7 +218,7 @@ if CUDNN_AVAILABLE:
             O.set_uid(UIDs.O_UID.value).set_output(True).set_dim(
                 [b, h_qo, s_qo, d_vo]
             ).set_stride([d_vo * h_qo, d_vo, d_vo * h_qo, 1]).set_data_type(
-                cudnn.data_type.BFLOAT16
+                cudnn_q_data_type
             )
 
         tensors_to_return = [cudnn_q, cudnn_k_cache, cudnn_v_cache, O]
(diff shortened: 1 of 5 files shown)


davidkny22:
Heads up, this overlaps #4146, which I opened on 07-25 for the same cache-key defect. Mine is narrower, just the prefill stride case plus a regression test in `tests/attention/test_cudnn_graph_cache_key.py`.

They conflict in `_sdpa_prefill_key_fn`, so they can't both land as they are. Happy to close mine in favor of this one, or to trim mine down to just the test if that's useful to you.

@yzh119 @saltyminty either way works for me, whichever is less work to review.

PR #4285, opened 2026-07-31
perf(moe): sync SM12x NVFP4 fused-MoE kernels to b12x HEAD

yichengj0:
## 📌 Description

Syncs the SM12x NVFP4 (W4A4) fused-MoE kernels behind `b12x_fused_moe` and `B12xMoEWrapper` with upstream [b12x](https://github.com/local-inference-lab/sparkinfer) at `f9be272`, reaching backend parity.

The module now has four kernels. `MoEDirectMicroKernel` takes the smallest decode batches, `MoEMicroKernel` and `MoEStaticKernel` cover the rest of decode with tile assignments fixed at launch, and `MoEDynamicKernel` covers prefill by distributing tiles through an in-kernel work queue.

Changes:

- Adopt upstream's current `MoEDirectMicroKernel`, replacing a stale copy in `moe_direct_micro_kernel.py` that was never routed to. It reads the top-k ids directly, so the smallest batches skip the routing pre-pass that dominates their runtime, and it now supports gelu_tanh.
- Let `MoEDynamicKernel` use tiles smaller than 128 rows, so sparse routing stops padding every expert up to 128.
- Launch all kernels cooperatively. They synchronize the whole grid between phases, which deadlocks if concurrent work keeps part of the grid off the GPU.
- Simplify `MoEDynamicKernel`'s work queue and size its pipeline stages from the real shared-memory footprint.
- Vectorize `MoEStaticKernel`'s scatter epilogue and refresh the decode tile heuristics.
- Restrict the fast FP4 quantizer to gated activations, since relu2's squared outputs need the exact one.

Public API and behavior changes: none.

## 📊 Performance

Versus main, measured over 38 target-model MoE shapes on three GPUs. Geomean speedup per shape group:

| shape group | DGX Spark | RTX 5080 | RTX Pro 6000 SE |
|---|---|---|---|
| decode, batch 1-2 | 1.06x | 1.30x | 1.40x |
| decode, batch 4-40 | 1.00x | 1.00x | 1.00x |
| prefill, batch 512-2048 | 1.10x | 1.10x | 1.09x |
| prefill, batch 8192 | 1.00x | 1.00x | 1.00x |

- The batch 1-2 gains come from the new `MoEDirectMicroKernel`; the shapes it serves win up to 2.3x.
- The prefill gains come from `MoEDynamicKernel` picking tiles smaller than 128 rows; those shapes win up to 1.26x.

## 🔍 Related Issues

#4223 (item 2).

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

- On SM121 (DGX Spark GB10): `tests/moe/test_b12x_fused_moe.py` (157 cases) and `tests/moe/test_unified_moe_b12x.py` (58 cases) pass, including tests that force each backend and a test that a smaller tile is picked again after a large-batch call.

## Reviewer Notes

Deviations from the b12x source, beyond import plumbing:

- gelu_tanh added to `MoEDirectMicroKernel`, matching the activation coverage of the module's other kernels.
- The module's API accepts reciprocal-form scales but upstream's kernel only takes multipliers, so dispatch inverts them before launch.
- `MoEDirectMicroKernel`'s cutover is re-measured: upstream's only alternative is its counterpart of `MoEDynamicKernel`, while here it competes with `MoEMicroKernel` and `MoEStaticKernel`.
- Upstream sizes scratch and builds the kernel together inside one plan object, so their tile sizes always match. FlashInfer has no plan object, so the workspace cache is keyed by tile size to give the same guarantee.


CLAUDE.md
```diff
@@ -575,6 +575,7 @@ Used by `flashinfer.trace` / `fi_trace`.
 | `FLASHINFER_ROUTING_FORCE_BLOCK_PER_TOKEN` | unset | `csrc/fused_moe/trtllm_backend/trtllm_fused_moe_routing_custom.cu` | Forces the TRT-LLM MoE custom-routing kernel into "one-block-per-token" mode regardless of the active routing policy. Mainly used to reproduce specific perf points. |
 | `FLASHINFER_B12X_MICRO_SHARE_INPUT` | `1` | `flashinfer/fused_moe/cute_dsl/blackwell_sm12x/moe_dispatch.py` | `0` disables the B12x MoE micro-batch input-sharing optimization. Internal/experimental — leave at the default unless investigating an SM12x MoE regression. |
 | `FLASHINFER_B12X_FORCE_MOE_W4A16` | unset | `flashinfer/fused_moe/cute_dsl/blackwell_sm12x/moe_dispatch.py` | When set (any non-empty value), forces the SM12x MoE dispatcher onto the W4A16 kernel path regardless of weight dtype. Internal/experimental — used to reproduce W4A16-specific issues. |
+| `FLASHINFER_B12X_MOE_TILE_MN` | unset | `flashinfer/fused_moe/cute_dsl/blackwell_sm12x/moe_dispatch.py` | Overrides the SM12x MoE decode tile (`64x128` or `128x128`), bypassing the routed-rows heuristic. Applies to the micro kernel and the multi-top-k static kernel only. Internal/experimental — benchmarking aid. |
```
bkryu: Can we refrain from adding a new environment variable `FLASHINFER_B12X_MOE_TILE_MN`?

I don't foresee any user using it so we can likely delete

flashinfer/fused_moe/cute_dsl/blackwell_sm12x/moe_dispatch.py
```diff
@@ -219,21 +241,41 @@ def _get_static_compact_cutover_pairs(activation_precision: str = "fp4") -> int:
...
     """
+    override = os.environ.get("FLASHINFER_B12X_MOE_TILE_MN")
+    if override:
+        if override not in ("64x128", "128x128"):
+            raise ValueError(
+                f"FLASHINFER_B12X_MOE_TILE_MN={override!r} is not supported. "
+                "The micro and static kernels only take 64x128 or 128x128."
+            )
+        tile_m, tile_n = override.split("x")
+        return (int(tile_m), int(tile_n))
```
bkryu: Let's delete the environment variable here. I don't think any customer will be using this one

yichengj0: Agreed, it was carried over from upstream as a benchmarking aid but has no real user. Removed in bf702509, along with the CLAUDE.md entry.

benchmarks/routines/moe.py
```diff
@@ -1711,16 +1711,20 @@ def testB12xFusedMoe(args):
...
+    }
     if activation_type not in _ACT_STR:
         raise ValueError(
-            f"b12x_fused_moe only supports Swiglu and Relu2 activations, "
+            f"b12x_fused_moe only supports Swiglu, Geglu, and Relu2 activations, "
             f"got {activation_type.name}"
         )
     activation_str = _ACT_STR[activation_type]
-    is_gated = activation_type == ActivationType.Swiglu
+    is_gated = activation_type in (ActivationType.Swiglu, ActivationType.Geglu)
```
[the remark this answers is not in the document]
yichengj0: Good catch. Applied in c0cb7d65: GegluTanh now maps to gelu_tanh alongside Geglu, and the gated check uses `ActivationType.is_gated` instead of the tuple.

flashinfer/fused_moe/cute_dsl/blackwell_sm12x/moe_direct_micro_kernel.py
```diff
@@ -1,19 +1,18 @@
...
+"""MoEDirectMicroKernel: direct-routed NVFP4 MoE decode kernel for SM12x.
+
+Unlike MoEMicroKernel (Triton route pre-pass, expert-major packed A), this
+kernel consumes raw per-token topk_ids/topk_weights with no routing
+pre-pass: 16-warp CTAs quantize each token row into shared memory, run FC1
+as software fp4 dot products over f16x2 HFMA, activate, quantize the
+intermediate, cross a resident-grid barrier (or per-token epoch barrier at
+m==1), and run FC2 with a bf16x2 atomic scatter into the token-major
+output.
 """
```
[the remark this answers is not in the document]
yichengj0: The docstring's pipeline description, including that wording, was removed in c6e89c79, which landed just before this review ran. The current text makes no claim about the FC2 store path.

aleozlx (approved):

PR #4286, opened 2026-07-31
feat: log git commit hash on import (FLASHINFER_LOGLEVEL >= 1)

aleozlx:
## 📌 Description

When a model serving run crashes, determining which FlashInfer commit was used requires reverse-engineering Docker image build timestamps — a process that currently takes hours per regression investigation.

This PR exposes the git commit hash captured at build time and logs it once on import, so crash logs always contain the exact commit without any manual archaeology:

```bash
$ FLASHINFER_LOGLEVEL=1 python -c "import flashinfer"
FlashInfer 0.6.16 (commit b9bb5074)
```

Changes:
- `flashinfer/__git_commit__` — new public attribute aliasing the existing `__git_version__` (full SHA-1 or `"unknown"` when building from a tarball without git). Both names are preserved for backward compatibility.
- **Import-time log**: when `FLASHINFER_LOGLEVEL >= 1`, prints `FlashInfer X.Y.Z (commit <short-sha>)` once on first import. Respects `FLASHINFER_LOGDEST` (stdout/stderr/file).
- **`collect-env`**: `python -m flashinfer.collect_env` now includes the git commit alongside the version.

### ✅ Pre-commit Checks

## 🧪 Tests

## Reviewer Notes

The git hash (`__git_version__`) was already captured during build via `build_backend.py` → `build_utils.get_git_version()` → `_build_meta.py`. This PR only exposes it and logs it — no change to the build system.

## Summary by CodeRabbit

- **New Features**
  - Added Git commit information to package metadata and environment reports.
  - Added optional import-time logging of the package version and shortened commit identifier.
  - Logging can be directed to standard output, standard error, or a file through environment settings.
- **Bug Fixes**
  - Logging failures, invalid settings, and unavailable metadata no longer interrupt package imports.
- **Compatibility**
  - Existing version metadata remains available through a backward-compatible alias.
  - Commit information falls back to an unknown value when unavailable.


flashinfer/collect_env.py
```diff
@@ -259,6 +259,7 @@ def _get_flashinfer_info():
         info["flashinfer"] = f"<import failed: {type(e).__name__}: {e}>"
         return info
     info["flashinfer"] = _guard(lambda: flashinfer.__version__, "?")
+    info["flashinfer git commit"] = _guard(lambda: flashinfer.__git_commit__, "?")
```
aleozlx: simplify to "flashinfer commit"

flashinfer/version.py
```diff
@@ -17,7 +17,11 @@
 # Centralized version information to avoid circular imports
 try:
     from ._build_meta import __version__ as __version__
-    from ._build_meta import __git_version__ as __git_version__
+    from ._build_meta import __git_commit__ as __git_commit__
```
bkryu: Not impacting released versions, but in a development environment, if a developer does a `git pull` in a local clone that is installed with editable install, I thinks this changes breaks `import flashinfer` because the `_build_meta.py` will be stale (need to re-install flashinfer from local).

The except block below won't catch it because it raises an `ImportError` (module found but cannot import name)

Something like
```
try:
    from . import _build_meta
    __version__ = _build_meta.__version__
    __git_commit__ = getattr(_build_meta, "__git_commit__", None) or getattr(
        _build_meta, "__git_version__", "unknown"
    )
except ModuleNotFoundError:
    __version__ = "0.0.0+unknown"
    __git_commit__ = "unknown"
__git_version__ = __git_commit__
```
should prevent development environments from breaking

flashinfer/__init__.py
```diff
@@ -262,3 +263,45 @@
...
+    #                so a logging misconfiguration can never block the import.
+    try:
+        if int(_os.environ.get("FLASHINFER_LOGLEVEL", "0")) < 1:
+            return
+        try:
+            _short = __git_commit__[:8] if __git_commit__ != "unknown" else "unknown"
+        except Exception:
+            _short = "unknown"
+        _line = f"FlashInfer {__version__} (commit {_short})\n"
+        _dest = _os.environ.get("FLASHINFER_LOGDEST", "stdout")
```
bkryu: Current `api_logging` supports `%i` → PID substitution in the destination path, which this reimplementation doesn't.

with this change, I'd think `FLASHINFER_LOGDEST=log_%i.txt` would open a literal `log_%i.txt`. Can we check this?

flashinfer/collect_env.py
aleozlx: Already addressed in the current code — the key is `"flashinfer commit"` (pushed in the first commit). Thanks!

flashinfer/version.py
aleozlx: Fixed in the latest commit — now loads `_build_meta` as a module and uses `getattr` with fallback to `__git_version__` for stale generated files, matching your suggested pattern. Also added `# type: ignore[attr-defined]` for mypy.

flashinfer/__init__.py
aleozlx: Confirmed and fixed — `FLASHINFER_LOGDEST=log_%i.txt` was opening a literal `log_%i.txt`. The latest commit applies `.replace("%i", str(os.getpid()))` on the dest string before branching on it, matching `api_logging.py`'s behaviour.

flashinfer/version.py
```diff
@@ -16,8 +16,19 @@
 
 # Centralized version information to avoid circular imports
 try:
-    from ._build_meta import __version__ as __version__
-    from ._build_meta import __git_version__ as __git_version__
+    from . import _build_meta  # type: ignore[attr-defined]
```
bkryu, line 19: Thanks now this fixes the case when there is a stale `_build_meta.py`, but now this might break when FlashInfer was never installed and `_build_meta.py` does not even exist.

It does not sounds like a common use case, but Claude is telling me it is a valid use case:

>   1. CI / lint / docs jobs that import the package but don't build it — a mypy, ruff, or Sphinx job that puts the repo on the path to introspect the Python API without running a full pip install.
> 2.  git clean -fdx (or any clean that removes untracked files) wipes the generated _build_meta.py; the next import before re-installing hits the missing-file path.

Root cause: The switch from from `._build_meta import ...` to ` from . import _build_meta` is the issue here. The two import forms raise different exceptions when `_build_meta.py` doesn't exist at all.
  - from `._build_meta import __version__` → `ModuleNotFoundError` → caught by exception below
  - from `. import _build_meta` → plain `ImportError` ("cannot import name") → not caught, so import flashinfer crashes

Fix: [`ModuleNotFoundError` is a subclass of `ImportError`](https://docs.python.org/3/library/exceptions.html#ModuleNotFoundError), so catching only the subclass misses the plain-`ImportError`. Thus widen the clause to the parent class — `except ImportError:` — which covers both the missing-file and stale-name cases in one line.

bkryu (commented):
Thanks @aleozlx, left one more comment after being suggested by my agent 😅

flashinfer/version.py
aleozlx: Great catch — fixed in the latest commit. Widened `except ModuleNotFoundError` to `except ImportError` so both cases are handled by one clause: `ModuleNotFoundError` (subclass) for the stale-attribute case, and plain `ImportError` for the missing-file case (CI lint jobs, post-`git clean`, etc.).

bkryu (approved):

aleozlx:
added 0.6.18 label due to cherry picking to 0.6.17rc2 before merging to main

PR #4292, opened 2026-07-31
fix(test): mock the DEEPGEMM_RUBIN pin in test_get_subdir_file_list

aleozlx:
## 📌 Description

`tests/test_artifacts.py::test_get_subdir_file_list` has been failing on `main` since #4280:

```
FileNotFoundError: [Errno 2] No such file or directory:
  '/tmp/pytest-of-root/pytest-0/test_get_subdir_file_list0/cubins/7ec7ac40b9fd48172651b77ff2ebe20d79decc39/deep-gemm/checksums.txt'
```

**Root cause.** #4280 added `ArtifactPath.DEEPGEMM_RUBIN` and appended it to the `cubin_dirs`
list walked by `get_subdir_file_list()`. The same PR registered `responses` mocks for the two
*other* new Rubin pins (`TRTLLM_GEN_BMM_RUBIN`, `TRTLLM_GEN_GEMM_RUBIN`) but not for the
deep-gemm one. Under `@responses.activate` an unregistered URL raises `ConnectionError`, so
`download_file()` exhausts its 4 retries and returns `False`; `get_checksums()` ignores that
return value and `open()`s a `checksums.txt` that was never written.

**The artifact itself is fine** — I fetched both deep-gemm manifests from artifactory (HTTP 200)
and the live SHA256 of `7ec7ac40…/deep-gemm/checksums.txt` matches the pin in
`CheckSumHash.DEEPGEMM_RUBIN` (`09e961d4…`). Only the unit test's mock registry was stale, so no
production download path is affected.

**Changes:**

- `tests/test_artifacts.py`: register the `DEEPGEMM_RUBIN` directory index and `checksums.txt`,
  mirroring the BMM/GEMM Rubin pattern (same index body, distinct per-pin hashes).
- `tests/test_artifacts.py`: extend the existing per-pin-hash-collision assertion to a shared
  deep-gemm kernel name, so deep-gemm gets the same regression coverage the trtllm-gen pins have.
- `flashinfer/artifacts.py`: `get_checksums()` now raises a `RuntimeError` naming the pin and the
  URL when the manifest download fails, instead of falling through to a bare `FileNotFoundError`
  on a local cache path. That failure mode is what made this read as a cache bug rather than
  "a pin was added without a mock".

## 🔍 Related Issues

Regression from #4280. Observed on unrelated PR CI, e.g.
https://github.com/flashinfer-ai/flashinfer/actions/runs/30607427968/job/91082694368

### ✅ Pre-commit Checks

## 🧪 Tests

- [ ] All tests are passing (`unittest`, etc.) — verification in progress on a dev box; draft until `tests/test_artifacts.py` is confirmed green.

## Reviewer Notes

CPU-only change; no kernel or GPU behavior is touched. The `get_checksums()` error-path change is
the only non-test edit — it converts a silent-download-failure into a diagnosable error and does
not alter the success path.

AI-assisted (Claude Code): diagnosis and patch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved checksum validation when artifact manifests are unavailable.
  * Errors now clearly identify the affected artifact and repository instead of producing a less informative file-not-found error.
  * Added fallback to cached checksum manifests when refreshing from a repository fails.
  * Prevented checksum collisions for shared kernel filenames across different DeepGEMM paths, including the Rubin variant.

### DIFF
--- a/flashinfer/artifacts.py
+++ b/flashinfer/artifacts.py
@@ -226,7 +226,15 @@ def get_checksums(subdirs):
             FLASHINFER_CUBINS_REPOSITORY, safe_urljoin(subdir, "checksums.txt")
         )
         checksum_path = FLASHINFER_CUBIN_DIR / safe_urljoin(subdir, "checksums.txt")
-        download_file(uri, checksum_path)
+        if not download_file(uri, checksum_path) and not checksum_path.is_file():
+            # Without this the next open() fails with a bare FileNotFoundError on
+            # the local cache path, which hides the real cause: the artifact pin
+            # is unreachable (typo'd/unpublished pin, or network/mirror failure).
+            raise RuntimeError(
+                f"Failed to fetch the checksum manifest for artifact pin '{subdir}' "
+                f"from {uri}. Check that the pin exists in "
+                f"{FLASHINFER_CUBINS_REPOSITORY} and is reachable."
+            )
         with open(checksum_path, "r") as f:
             for line in f:
                 sha256, filename = line.strip().split()
--- a/tests/test_artifacts.py
+++ b/tests/test_artifacts.py
@@ -4,6 +4,7 @@ from flashinfer.artifacts import (
     get_subdir_file_list,
 )
 
+import pytest
 import responses
 
 from flashinfer.jit.cubin_loader import safe_urljoin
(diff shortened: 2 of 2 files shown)


Aneureka (approved):
LGTM.

aleozlx:
closing in favor of #4301

Issue #4267, opened 2026-07-30
[Bug] FA2 attention silently outputs zeros when every logit in a row is below -5e4 (math::inf sentinel is a finite value)


tangcy98:
### Before submitting

- [x] I have searched existing issues and did not find a solution or an existing report.

### Environment

```text
### FlashInfer environment report
<!-- generated by `python -m flashinfer.collect_env`; paste into your issue -->

==== FlashInfer ====
flashinfer                   : 0.6.12
flashinfer file              : /usr/local/lib/python3.12/dist-packages/flashinfer/__init__.py
flashinfer-cubin             : 0.6.12
flashinfer-jit-cache         : 0.6.12+cu130  ⚠ MISMATCH vs flashinfer-python==0.6.12
JIT cache                    : /root/.cache/flashinfer (0 compiled .so)
Cubin store (local)          : /usr/local/lib/python3.12/dist-packages/flashinfer_cubin/cubins (15558 cubins, 1754 MiB)
Target CUDA archs (resolved) : {(12, '0f')}

==== Python / Platform ====
Python            : 3.12.3 (main, Jun 19 2026, 12:46:00) [GCC 13.3.0]
Python executable : /usr/bin/python
Platform          : Linux-5.15.0-181-generic-x86_64-with-glibc2.39
libc              : glibc 2.39
OS                : Ubuntu 24.04.3 LTS
Container         : no / not detected

==== GPU / Driver ====
Driver version       : 580.126.20
CUDA_VISIBLE_DEVICES : <unset>
GPU 0 (CUDA order)   : NVIDIA RTX PRO 5000 72GB Blackwell | SM120 | 110 SMs | 71.1 GiB
GPU 1 (CUDA order)   : NVIDIA RTX PRO 5000 72GB Blackwell | SM120 | 110 SMs | 71.1 GiB

==== CUDA Toolkit ====
nvcc on PATH                            : not found
CUDA_HOME (env)                         : <unset>
CUDA toolkit resolved by flashinfer JIT : /usr/local/cuda (version 13.0)

==== PyTorch ====
torch                          : 2.11.0+cu130
torch.version.cuda             : 13.0
torch CXX11 ABI                : True
torch compiled archs           : sm_75 sm_80 sm_86 sm_90 sm_100 sm_120
torch.backends.cudnn.version() : 91900
torch file                     : /usr/local/lib/python3.12/dist-packages/torch/__init__.py

==== cuDNN Frontend ====
[...]

PR #962, opened 2025-03-19
Fix compilation with FP16_QK_REDUCTION enabled.

diptorupd:
As described in #806 and #936, setting the cmake build flag `FLASHINFER_GEN_USE_FP16_QK_REDUCTIONS` to "true" causes a build failure due to `cuda_fp16.h` not supporting `constexpr` cast from `__half` type to `float`. Note that the issue is not just a CMake/C++ configuration issue the issue will be triggered even in the flashinfer JIT code compilation path as reported in #915.

The PR fixes #806 and #936 by adding a modified version of the FP16 header from the [FP16 library](https://github.com/Maratyszcza/FP16) that supports `constexpr` versions of the conversion functions. To make the conversion functions `constexpr`, I am using `std::bit_cast` that is the reason for bumping the required standard to 20.

With these changes I am able to build the C++ API with both `FLASHINFER_GEN_USE_FP16_QK_REDUCTIONS` ON and OFF.

Fixes #936 
Fixes #806

PR #4298, opened 2026-07-31
fix: make the FP16_QK_REDUCTION path compile and decode QK logits correctly

### LINKED ISSUE #4297: [BUG] FP16_QK_REDUCTION path does not compile, and decodes QK logits with a value cast instead of a bit cast
### Description

In `include/flashinfer/attention/prefill.cuh`, the `FP16_QK_REDUCTION` path
(`DTypeQKAccum = __half`) converts the QK accumulator fragment back to `float`
with:

```cpp
// include/flashinfer/attention/prefill.cuh:1400
logits = std::bit_cast<float>(fp16_ieee_to_fp32_value(s_frag[mma_q][mma_kv][reg_id]));
```

`fp16_ieee_to_fp32_value` is declared in `include/flashinfer/fp16.h:48` as:

```cpp
static constexpr float fp16_ieee_to_fp32_value(uint16_t h)
```

It expects the IEEE half **bit pattern** in a `uint16_t`. But the argument
passed is a `__half`, and nothing bit-casts it. What happens instead is
`__half`'s implicit conversion operator to `unsigned short`
(`cuda_fp16.hpp`):

```cpp
__CUDA_HOSTDEVICE__ operator unsigned short() const { return __half2ushort_rz(*this); }
```

`__half2ushort_rz` is a saturating round-toward-zero **numeric value**
conversion (`cvt.rzi.u16.f16` on device), not a reinterpret. So the bit pattern
handed to the decoder is the *integer value* of the half, not its bits.

Note also that the outer `std::bit_cast<float>` is a no-op —
`fp16_ieee_to_fp32_value` already returns `float`. It gives the line the
appearance of doing a bit-level conversion while the actual bit-level step is
missing.

### Impact

This corrupts essentially every logit, not just edge cases:

- Any **negative** half saturates to `0x0000` and decodes to `+0.0f`. This
  includes the `-inf` softmax mask sentinel (`KTraits::MaskFillValue`,
  `KTraits::getNegInf()`).
- Any **positive** half is truncated to its integer part, which is then
  interpreted as a half bit pattern, producing a tiny subnormal
  (`+1.0` -> `5.96e-08`, `+10.0` -> `5.96e-07`).
- Sweeping all 65536 half bit patterns: **Among 63,490 non-NaN half patterns,
  63,489 decode incorrectly; only `+0.0` matches bit-for-bit.**

The failure is **silent**. The symmetric write-back a few lines later is
correct:

```cpp
// :1415
s_frag[mma_q][mma_kv][reg_id] = std::bit_cast<half>(fp16_ieee_from_fp32_value(logitsTransformed));
```

(`fp16_ieee_from_fp32_value` returns `uint16_t`, so `std::bit_cast<half>` is
right.) The fragment therefore stays well-formed and no NaN or assert appears.
Because masked positions become `0.0` instead of `-inf`, they receive uniform
softmax weight, so the row degenerates to `lse == log2(kv_len)` and
`output == mean(V)`.

### The path does not even compile on current main

Before the numeric bug can be reached at all, the path fails to build. On
`main` with `-DFP16_QK_REDUCTION_SUPPORTED` and a `boost/math/ccmath` include
path, `nvcc` rejects `include/flashinfer/attention/variant_helper.cuh:86`:

```cpp
// AttentionVariantBase::OutputTransform
float d_rcp = (m != -math::inf) ? math::ptx_rcp(d) : 0.f;
```

`m` has type `T_M&`, which is `DTypeQKAccum` — `__half` on this path — while
`-math::inf` is a `float`. The comparison is ambiguous, since `__half` converts
to `float` and `float` converts to `__half`, and CUDA supplies `operator!=`
overloads for `__half`, `__half2`, `__nv_bfloat16` and `__nv_bfloat162`:

```
variant_helper.cuh(86): error: no operator "!=" matches these operands
    ... note #3326-D: function "operator==(const __half2 &, const __half2 &)"
        does not match because argument #1 does not match parameter
    ... note #3326-D: function "operator==(const __nv_bfloat16 &, const __nv_bfloat16 &)"
        does not match because argument #1 does not match parameter
    ... note #3328-D: built-in operator==(<nullptr>, <nullptr>)
        does not match because argument #1 does not match parameter
1 error detected in the compilation of
  ".../single_prefill_..._f16qk_True/single_prefill_kernel_mask_1.cu"
```

So `FP16_QK_REDUCTION` currently has **two independent latent defects**: this
compile error, and the value-cast bug above that it hides. Neither can be
observed without first supplying the define by hand.

One subtlety worth flagging for whoever fixes the compile error: `math::inf` is
the **finite** sentinel `5e4`, not an infinity. Widening `m` to `float` compiles
but is wrong — the `__half` mask fill value is
`fp16_ieee_from_fp32_value(-math::inf)` = `0xFA1A` = `-49984.0f`, which compares
unequal to `-50000.0f`, so a fully-masked row would be treated as live and take
`math::ptx_rcp(d)` with `d == 0`. Comparing in the accumulator type instead
(`static_cast<T_M>(-math::inf)`) is both unambiguous and correct, and is the
identity for `T_M = float`.

### Why this has gone unnoticed

The path appears to be unreachable from a stock build:

- The line was introduced verbatim by #962 ("Fix compilation with
  FP16_QK_REDUCTION enabled.", commit `3a695603`), which also added `fp16.h`.
  `git log -L1398,1402:include/flashinfer/attention/prefill.cuh` shows it as the
  most recent change to that region. That PR's goal was making the path
  **compile**, and with the implicit `operator unsigned short()` available it
  compiled cleanly — which is exactly why the value cast went unnoticed. The
  no-op outer `std::bit_cast<float>` reinforces the impression that the bit-level
  conversion was handled.
- `FP16_QK_REDUCTION_SUPPORTED` is referenced **only inside `prefill.cuh`
  itself** (lines 30, 329, 343, 1398, 1406, 1412). Nothing in the workflows,
  build files, or JIT flag lists defines it.
- Without the define, the `static_assert` at `:1405` (and `:342`) hard-fails the
  build when `DTypeQKAccum` is `__half`, so requesting fp16 QK reduction cannot
  silently fall through.
- Enabling it requires both `-DFP16_QK_REDUCTION_SUPPORTED` and a
  `boost/math/ccmath` include path (per the `static_assert` message).

So as far as I can tell no CI job compiles this path, which means neither defect
has ever been exercised. The two layers corroborate that directly: a path that
does not compile cannot have been built by anything, and a numeric bug sitting
behind a compile error cannot have been observed by anyone.

The history fits too. `prefill.cuh:1400` and `fp16.h` both arrived in #962
("Fix compilation with FP16_QK_REDUCTION enabled.", `3a695603`) — a PR whose
goal was to make this path *compile*. It did compile at the time, which is
exactly why the value cast was never examined numerically; and the path has
since rotted back into not compiling, without anything noticing.

Impact on current users is likely nil, but `use_fp16_qk_reduction` is part of the
public API surface, and this would ship broken the moment the define is wired up.

### Reproduction

The mechanism is deterministic and provable on the host, with no GPU. This
program contrasts the correct bit-cast decode against the value-cast decode that
`:1400` actually performs (emulating `__half2ushort_rz` per its definition in
`cuda_fp16.hpp`):

<details>
<summary><code>g++ -std=c++20 proof.cpp -o proof &amp;&amp; ./proof</code></summary>

```
half value                                           bits       correct(bit)   buggy(value) verdict
-inf  (softmax mask sentinel, KTraits::MaskFillValue) 0xFC00           -inf              0 DIVERGES
-10.0 (typical masked/low logit)                     0xC900            -10              0 DIVERGES
-1.0                                                 0xBC00             -1              0 DIVERGES
-0.125                                               0xB000         -0.125              0 DIVERGES
smallest negative subnormal                          0x8001   -5.96046e-08              0 DIVERGES
-0.0                                                 0x8000             -0              0 DIVERGES
+0.0                                                 0x0000              0              0 ok
+1.0                                                 0x3C00              1    5.96046e-08 DIVERGES
+10.0  (positive: value cast truncates to 10 -> decodes as 5.96e-07) 0x4900             10    5.96046e-07 DIVERGES
+inf                                                 0x7C00            inf           -nan DIVERGES

exhaustive sweep over all non-NaN half bit patterns:
  diverging patterns          : 63489
  negative patterns           : 31745
  negative patterns diverging : 31745 (all of them)
```

(full source attached / see linked PR)
</details>

I have also observed this on hardware, on an H200 (sm90), with
`-DFP16_QK_REDUCTION_SUPPORTED -std=c++20` plus a `boost/math/ccmath` shim.
Because of the compile error above, getting the kernel to build at all required
fixing `variant_helper.cuh:86` first; with only that compile fix in place and
`prefill.cuh:1400` left as-is, prefill cases returned `lse == log2(64)` exactly
with `output == mean(V)` — the signature of an all-zero `s_frag` entering the
softmax, matching the prediction above. The same harness on the same build with
`use_fp16_qk_reduction=False` was correct, so the only variable was the fp16
accumulator.

With both fixes applied, the same cases track an fp64 reference to within fp16
accumulation error (e.g. logits `-512/-256` -> `lse = -27.656` vs reference
`-27.644`; sign-mixed `+512/-256` -> `70.312` vs `70.289`; all-positive `+51200`
-> `6537.249` vs `6534.892`), and the fp16 and fp32 paths produce bit-identical
output on identical inputs.

### Suggested fix

Two one-line changes, one per defect.

`prefill.cuh:1400` — reinterpret the bits, matching the direction already used by
the write-back at `:1415` and by `KTraits::getNegInf()` at `:333`:

```cpp
logits = fp16_ieee_to_fp32_value(std::bit_cast<uint16_t>(s_frag[mma_q][mma_kv][reg_id]));
```

`variant_helper.cuh:86` — compare in the accumulator type, which is unambiguous
and, per the note above, the only variant that stays correct for the finite
sentinel (and is the identity when `T_M` is `float`):

```cpp
float d_rcp = (m != static_cast<T_M>(-math::inf)) ? math::ptx_rcp(d) : 0.f;
```

`<bit>` is already available via `fp16.h`. Happy to open a PR — I have both
changes built and passing on an H200 (sm90).

Separately, it may be worth deciding whether this path should be built by CI or
removed. Two independent defects accumulating unnoticed suggests that an
unbuilt, untested code path behind a manual define is not carrying its
maintenance cost.

Smallfu666:
Fixes #4297

### Problem

The `FP16_QK_REDUCTION` path (`DTypeQKAccum = __half`) has two independent
latent defects, the first hiding the second.

**1. It does not compile.** `variant_helper.cuh:86` in
`AttentionVariantBase::OutputTransform` compares `m` (type `T_M&`, i.e. `__half`
here) against `-math::inf` (a `float`). That is ambiguous — `__half` converts to
`float`, `float` converts to `__half`, and CUDA supplies `operator!=` for
`__half`/`__half2`/`__nv_bfloat16`/`__nv_bfloat162` — so nvcc rejects it with
"no operator `!=` matches these operands".

Note `math::inf` is the **finite** sentinel `5e4`, not an infinity, so widening
`m` to `float` would compile but silently break masked rows: the `__half` mask
fill value is `fp16_ieee_from_fp32_value(-math::inf)` = `0xFA1A` = `-49984.0f`,
which compares unequal to `-50000.0f`, so a fully-masked row would be treated as
live and take `math::ptx_rcp(d)` with `d == 0`.

**2. It decodes QK logits with a value cast.**
`include/flashinfer/attention/prefill.cuh:1400` passed a `__half` straight to
`fp16_ieee_to_fp32_value(uint16_t)` (`include/flashinfer/fp16.h:48`), which
expects the IEEE half **bit pattern**. No bit-cast was applied to the argument,
so `__half`'s implicit `operator unsigned short()` ran instead — a saturating
round-toward-zero **value** conversion (`__half2ushort_rz`).

Result: negative logits (including the `-inf` mask sentinel) saturate to
`0x0000` and decode to `+0.0f`; positive logits truncate to an integer that is
then read as a half bit pattern, yielding a subnormal. Among 63,490 non-NaN half
patterns, 63,489 decode incorrectly; only `+0.0` matches bit-for-bit.

It was silent because the symmetric write-back at `:1415` is correct, so the
fragment stayed well-formed. Masked positions gained uniform softmax weight,
collapsing rows to `lse == log2(kv_len)` and `output == mean(V)`.

The line was introduced verbatim by #962 (`3a695603`), whose goal was to make
this path *compile*; with `__half`'s implicit `operator unsigned short()`
available it compiled cleanly, so the value cast went unnoticed.

### Change

Two commits, one per defect. 2 files changed, +2/-2.

**1. `variant_helper.cuh` — compare in the accumulator type:**

```diff
-    float d_rcp = (m != -math::inf) ? math::ptx_rcp(d) : 0.f;
+    float d_rcp = (m != static_cast<T_M>(-math::inf)) ? math::ptx_rcp(d) : 0.f;
```

This resolves the ambiguity via the unambiguous
`operator!=(const __half&, const __half&)`. It is also the only variant that
stays correct given the finite sentinel: `static_cast<__half>` rounds to nearest
even exactly as `fp16_ieee_from_fp32_value` does, so it reproduces the `0xFA1A`
fill value and the comparison selects the masked branch. For `T_M = float` the
cast is the identity, so **fp32 behaviour is bit-for-bit unchanged**.

**2. `prefill.cuh` — bit-cast the fragment instead of value-converting it:**

```diff
-          logits = std::bit_cast<float>(fp16_ieee_to_fp32_value(s_frag[mma_q][mma_kv][reg_id]));
+          logits = fp16_ieee_to_fp32_value(std::bit_cast<uint16_t>(s_frag[mma_q][mma_kv][reg_id]));
```

`std::bit_cast` matches the existing idiom in this same file (`:1415`, `:333`);
`<bit>` is already pulled in by `fp16.h` under the same `#ifdef`. The outer
`std::bit_cast<float>` is dropped because it was a no-op —
`fp16_ieee_to_fp32_value` already returns `float`.

This also makes the line portable to builds that define
`__CUDA_NO_HALF_CONVERSIONS__` — as some PyTorch extension configurations do —
where `__half`'s `operator unsigned short()` is gated out and the old expression
is a **hard compile error** rather than a silent miscomputation. Both directions
were checked with nvcc (sm_80, C++20): the old line fails to compile with the
macro defined and compiles-but-miscomputes without it; the new line compiles and
is correct in both configurations.

### Testing

Verified statically, on the host, and on hardware.

**Host-side proof of the decode bug (deterministic, exhaustive).** Comparing the
correct bit-cast decode against the value-cast decode the old line performed:
**Among 63,490 non-NaN half patterns, 63,489 decode incorrectly; only `+0.0`
matches bit-for-bit.** All 31,745 negative patterns collapse to `+0.0`. Program
and full output in the linked issue.

**GPU confirmation — H200 (sm90).** Built upstream `main` with only these two
commits applied, using
`-DFP16_QK_REDUCTION_SUPPORTED -I<boost ccmath shim> -std=c++20`. The path
compiles and runs. All cases below are one build / one process; fp32 controls use
the same translation unit with `use_fp16_qk_reduction=False`, so they exercise the
`static_cast<T_M>` identity case and the `else if constexpr` branch. Reference is
fp64.

| case (raw QK dot) | fp16 lse | fp64 ref lse | out max-err |
|---|---|---|---|
| `-512 / -256` | -27.656 | -27.644 | 1.202e-04 |
| `+512 / -256` (sign mix) | 70.312 | 70.289 | 1.202e-04 |
| `+51200` (all positive) | 6537.249 | 6534.892 | 1.150e-04 |
| fully masked | finite sentinel | — | out **exactly 0** |

- The `-512/-256` case is the one that exposed the decode bug: before the fix it
  returned `lse = log2(64) = 6.00` exactly with `output = mean(V)`, the signature
  of `s_frag` arriving all-zero. It now tracks the fp64 reference to within fp16
  accumulation error.
- The positive and sign-mixed cases confirm the bug was never negatives-only.
- **No regression:** the four fp32 controls reproduce the fp64 reference with
  `|dlse| = 0.000`.
- **fp16 and fp32 now agree bit-for-bit on output** (`|do|max = 0.000e+00`) for
  all three live cases on identical inputs; lse differs only by fp16
  accumulation error.

Note on the fully-masked case: `math::inf` is the finite sentinel `5e4`, so on
`main` a fully-masked row reports a large negative lse rather than `-inf`. I
assert the in-scope property — output exactly `0` — and left that sentinel
behaviour alone.

**Out of scope, for transparency.** A case whose legitimate raw QK dot is
`-51200` still disagrees with the reference, because `-51200` is more negative
than the `-50000` mask sentinel and the kernel cannot distinguish the two. That
is a separate pre-existing issue. This is tracked separately in #4267 and is not
addressed by this PR. It fails
*identically* with `use_fp16_qk_reduction` on and off (`lse=-49978.047` vs
`-49993.891`, ref `-6522.892`), which is what shows it is unrelated to these
changes.

**No CI coverage exists for this path.** `FP16_QK_REDUCTION_SUPPORTED` is
referenced only inside `prefill.cuh`; nothing in the build or workflows defines
it, and the `static_assert` at `:1405` blocks the fp16-accum instantiation
without it. So existing CI never instantiates `FP16_QK_REDUCTION`. Note that the
generic `OutputTransform` in `variant_helper.cuh` is still compiled by existing
jobs, for `T_M = float`, where `static_cast<T_M>` is the identity — so that half
of the change is exercised today and cannot regress silently. What existing jobs
cannot verify is the FP16-specific behaviour: the `__half` comparison and the
`s_frag` bit-cast. I'm glad to add a test if you can point me at where a build with
the define should live.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected FP16 attention score conversion for more accurate calculations.
  * Improved handling of negative-infinity values in output transformations.

### DIFF
--- a/include/flashinfer/attention/prefill.cuh
+++ b/include/flashinfer/attention/prefill.cuh
@@ -1397,7 +1397,7 @@ __device__ __forceinline__ void logits_transform(
 
 #ifdef FP16_QK_REDUCTION_SUPPORTED
         if constexpr (std::is_same<DTypeQKAccum, __half>::value) {
-          logits = std::bit_cast<float>(fp16_ieee_to_fp32_value(s_frag[mma_q][mma_kv][reg_id]));
+          logits = fp16_ieee_to_fp32_value(std::bit_cast<uint16_t>(s_frag[mma_q][mma_kv][reg_id]));
         } else if constexpr (!std::is_same<DTypeQKAccum, __half>::value) {
           logits = s_frag[mma_q][mma_kv][reg_id];
         }
--- a/include/flashinfer/attention/variant_helper.cuh
+++ b/include/flashinfer/attention/variant_helper.cuh
@@ -83,7 +83,7 @@ struct AttentionVariantBase {
   REGISTER_M_D_UPDATE(params, kv_tile_idx, qo_head_idx, m, d, scale, { return; })
 
   REGISTER_OUTPUT_TRANSFORM(params, output, batch_idx, qo_idx, qo_head_idx, m, d, scale, {
-    float d_rcp = (m != -math::inf) ? math::ptx_rcp(d) : 0.f;
+    float d_rcp = (m != static_cast<T_M>(-math::inf)) ? math::ptx_rcp(d) : 0.f;
     float v_scale_val = get_v_scale(params);
     return output * d_rcp * v_scale_val;
   })


lazypool (commented):
I reviewed this. The fix is correct and complete.

PR #4133, opened 2026-07-24
[GDN] improve sm100 GDN performance

Observer007:
## What

Optimizes the Blackwell SM100 chunk-stream Gated Delta Net (GDN) prefill kernel.
Coauthored by @guangyunh-nv @jhjpark 

## Test environment

- **GPU:** NVIDIA B300 SXM6 AC — Blackwell (SM100)
- **Compile:** `--enable-tvm-ffi --opt-level 3`
- **Dtype:** bf16 in/out, fp32 state + accumulation
- **Benchmark:** `benchmarks/bench_gdn_prefill.py`, 2 warmup + 8 timed iters, CUDA graph
- **Baseline ("pre-PR"):** the initial functional port of this kernel (`--opt-level 2`)

## Performance

120 configurations (8 head configs x 15 sequence configs). **Median 1.25x** speedup
(range 1.07x-1.32x). No regressions (every config is faster than the baseline).

<details>
<summary>Full 120-configuration results (h_qk/h_v, seqlen, pre-PR ms, this-PR ms, speedup)</summary>

| h_qk/h_v | seqlen | pre-PR (ms) | this PR (ms) | speedup |
|---|---|---|---|---|
| 2/8 | 1x65536 | 2.352 | 1.827 | 1.29x |
| 2/8 | 1x32768 | 1.184 | 0.923 | 1.28x |
| 2/8 | 1x16384 | 0.601 | 0.470 | 1.28x |
| 2/8 | 1x8192 | 0.311 | 0.242 | 1.28x |
| 2/8 | 1x4096 | 0.165 | 0.129 | 1.28x |
| 2/8 | 1x2048 | 0.091 | 0.069 | 1.31x |
| 2/8 | 6144+2048 | 0.237 | 0.187 | 1.27x |
| 2/8 | 4096+4096 | 0.165 | 0.131 | 1.26x |
| 2/8 | 2048+6144 | 0.237 | 0.187 | 1.27x |
| 2/8 | 1024+7168 | 0.274 | 0.215 | 1.27x |
| 2/8 | 2048x4 | 0.093 | 0.074 | 1.25x |
| 2/8 | 1024x8 | 0.057 | 0.047 | 1.22x |
| 2/8 | 8192x8 | 0.315 | 0.246 | 1.28x |
| 2/8 | 8192x16 | 0.317 | 0.248 | 1.28x |
| 2/8 | 8192x32 | 0.627 | 0.490 | 1.28x |
[...]

PR #4311, opened 2026-08-01
fix(gdn): use block-end decay for SM100 state updates

guangyunh-nv:
## Summary

- restore the block-end cumulative decay when updating the recurrent SM100 GDN state
- remove the thread-local coordinate materialization used to select that scalar
- add a shared regression test covering full, partial, and recurrent blocks on all supported GDN architectures

## Root cause

PR #4133 replaced the fixed block-end lookup with an index derived from the last coordinate in each thread's partitioned accumulator fragment:

```python
max_coord = tTR_tCcShared[cute.size(tTR_tCcShared) - 1]
cumprod_total = sCumprod[max_coord[1], 0, gate_handle.index]
```

That max_coord coordinate is thread-dependent. Its second coordinate is:

```
tid % 4 == 0: coord[1] = 57
tid % 4 == 1: coord[1] = 59
tid % 4 == 2: coord[1] = 61
tid % 4 == 3: coord[1] = 63
```

Representative output:

```
tid=0   coord=(24, 57)
tid=1   coord=(24, 59)
tid=2   coord=(24, 61)
tid=3   coord=(24, 63)
...
tid=124 coord=(127, 57)
tid=125 coord=(127, 59)
tid=126 coord=(127, 61)
tid=127 coord=(127, 63)
```

The fix directly loads the block-end scalar:

```python
cumprod_total = sCumprod[self.b_t - 1, 0, gate_handle.index]
```

This remains correct for a partial block. The gate loader predicates out-of-bounds elements after initializing them to the multiplicative identity `1.0`. Their log2 contribution is therefore zero, so the inclusive prefix product remains constant after the final valid token and slot `BT - 1` contains the product over exactly the valid block.

## Reproduction

The regression test uses FP16, sequence lengths `[64, 111, 192]`, one head, normalized K, and alpha/beta sampled from `[0.99, 1.0)`. These lengths cover one full block, a non-multiple tail with recurrent carry, and three aligned blocks.

| Revision | Output max error | Output mismatches | State max error | State mismatches |
|---|---:|---:|---:|---:|
| Before #4133 (`6258e522`) | `5.80e-4` | `0 / 46976` | `2.28e-4` | `0 / 49152` |
| After #4133 (`f057e15b`) | `1.59e-2` | `1660 / 46976` | `1.01e-2` | `6361 / 49152` |
| This fix | `5.80e-4` | `0 / 46976` | `2.28e-4` | `0 / 49152` |

## Validation

```bash
PYTHONPATH=. python -m pytest -q \
  tests/gdn/test_prefill_delta_rule.py::test_prefill_block_end_decay
```

Result: `1 passed` on SM100 with CUDA 13. The test now collects on SM90, SM100, and SM12x so architecture CI covers the shared contract.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved gated delta network prefill behavior for partial chunks and block-boundary sequence lengths.
  * Ensured cumulative decay uses neutral padding values when processing incomplete chunks.

* **Tests**
  * Added GPU coverage comparing prefill results and final states against the blockwise reference implementation, including decay factors near one.

### DIFF
--- a/tests/gdn/test_prefill_delta_rule.py
+++ b/tests/gdn/test_prefill_delta_rule.py
@@ -171,6 +171,67 @@ def _test_prefill_kernel(
     torch.testing.assert_close(our_state, ref_state, atol=atol_kv, rtol=rtol_kv)
 
 
+@torch.inference_mode()
+def test_prefill_block_end_decay(qkv_factory, seed=0):
+    _skip_if_unsupported()
+    random.seed(seed)
+    torch.random.manual_seed(seed)
+    torch.cuda.manual_seed(seed)
+
+    seq_lens = [64, 111, 192]
+    total_seqlen = sum(seq_lens)
+    num_heads = 1
+    head_size = 128
+    dtype = torch.float16
+    device = torch.device("cuda")
+
+    with device:
+        q, k, v = qkv_factory(
+            seq_lens, num_heads, num_heads, num_heads, head_size, dtype
+        )
+        k = torch.nn.functional.normalize(k, p=2.0, dim=-1)
+        alpha = 0.99 + 0.01 * torch.rand(total_seqlen, num_heads)
+        beta = 0.99 + 0.01 * torch.rand(total_seqlen, num_heads)
+        cu_seqlens = torch.tensor(exclusive_cumsum(seq_lens), dtype=torch.int64)
+
+    our_o = torch.empty_like(q)
+    our_state = torch.empty(
+        (len(seq_lens), num_heads, head_size, head_size),
+        dtype=torch.float32,
+        device=device,
+    )
+    chunk_gated_delta_rule(
+        q,
+        k,
+        v,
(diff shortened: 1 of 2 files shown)


tests/gdn/test_prefill_delta_rule.py
```diff
@@ -171,6 +171,67 @@ def _test_prefill_kernel(
     torch.testing.assert_close(our_state, ref_state, atol=atol_kv, rtol=rtol_kv)
 
 
+@torch.inference_mode()
+def test_prefill_sm100_block_end_decay(qkv_factory, seed=0):
+    _skip_if_not_sm100()
```
cloudhan: Other archs should also be checked. Relax it.

Observer007 (approved):

jiahanc (approved):
Thanks for the fix!

PR #4312, opened 2026-08-01
fix(prefill): validate head_dim is a multiple of 64 in FA2 prefill APIs

### LINKED ISSUE #4299: [Bug] FA2 prefill never validates head_dim: values not a multiple of 64 give an illegal memory access, values not a multiple of 16 silently return wrong results
## Summary

FlashInfer's FA2 prefill path accepts any `head_dim` at the Python level,
JIT-compiles a module for it, and then misbehaves in one of two ways depending on
the value. Nothing rejects it, at any layer.

* `head_dim` a multiple of 16 but not of 64 (80, 96, 112, 144, 160, 176 …) →
  **`CUDA error: an illegal memory access was encountered`**.
* `head_dim` not a multiple of 16 (72, 136 …) → the call **returns normally with
  wrong results**, no error anywhere.

Neither case needs a non-default argument: three tensors and `causal=True`.

Why it is worth your time: **`head_dim=96` is Phi-3-mini's head dimension**
(3072/32), and `head_dim=80` is in the Phi-2 generation, so the faulting class is
reachable from a real model. #455 was a user hitting `head_dim=96` with Phi-3-mini
through SGLang; on that (2024-era decode) path it produced a clean dispatch error.
On today's FA2 prefill path the same head dim is a memory fault instead. The
silently-wrong class needs nothing exotic at all — just a head dim that is not a
multiple of 16.

## Reproducer

```python
import torch
import flashinfer

dev = torch.device("cuda:0")
head_dim = 96          # 96 -> illegal memory access.  72 -> returns wrong results.  64/128 -> fine.

q = torch.randn(128, 8, head_dim, dtype=torch.float16, device=dev)
k = torch.randn(256, 8, head_dim, dtype=torch.float16, device=dev)
v = torch.randn(256, 8, head_dim, dtype=torch.float16, device=dev)

o = flashinfer.single_prefill_with_kv_cache(q, k, v, causal=True)
torch.cuda.synchronize()

# fp32 reference
s = torch.einsum("ihd,jhd->hij", q.float(), k.float()) / head_dim**0.5
i = torch.arange(q.shape[0], device=dev)[:, None]
j = torch.arange(k.shape[0], device=dev)[None, :]
s = s.masked_fill(~(j <= i + (k.shape[0] - q.shape[0])), float("-inf"))
ref = torch.einsum("hij,jhd->ihd", s.softmax(-1), v.float())
print(head_dim, (o.float() - ref).abs().max().item() / ref.abs().max().item())
```

Expected outcomes, qualitatively:

* `head_dim = 96` (also 80, 112): raises before it prints. With
  `CUDA_LAUNCH_BLOCKING=1` the error is attributed to the prefill kernel launch —
  `RuntimeError: Check failed: (status == cudaSuccess) is false:
  SinglePrefillWithKVCache kernel launch failed, error: an illegal memory access was
  encountered`. Without it, it surfaces later at the next synchronisation as
  `torch.AcceleratorError`, which is what makes it hard to trace back to `head_dim`
  in an application.
* `head_dim = 72` (also 136): prints a relative error of order 100 rather than the
  ~1e-3 you get at 64 or 128. I am deliberately not promising an exact figure for
  *this* snippet: the numbers I measured came from my own harness (a naive fp32
  reference with explicit GQA broadcast and bottom-right causal alignment), where it
  was `max_abs = 203.7` / `max_rel_to_peak = 242.1` at head_dim 72 and
  `max_rel_to_peak = 284.7` at 136, against ~3e-4 at 64/128. The snippet is the
  minimal form of the same comparison; treat its magnitude as "enormous", not as a
  number to match.
* `head_dim = 64` or `128`: prints ~3e-4.

## Measured surface

`head_dim` swept in steps of 8, one fresh process per case (a CUDA illegal access
poisons the context, so an in-process loop manufactures fake failures after the
first real one), `CUDA_LAUNCH_BLOCKING=1`, fp16 Q/K/V, `backend="fa2"`.

**This sweep is partial.** 28 head dims (8 … 224, step 8) have been measured on the
single-request path so far; the rest of the range and the full batch-path rows are
still running. The table is evidence for the three regimes; it is **not** a complete
residue-class map and I would not want it read as one.

`single_prefill_with_kv_cache(q, k, v, causal=True)`, `qo_len=128`, `kv_len=256`:

| head_dim | outcome |
|---|---|
| 8 | the JIT compile itself fails (`ninja: build failed`) |
| 16, 24, 32, 40, 48, 56 | `E1` |
| **64** | **ok** — matches fp32 reference, ~3e-4 |
| **72** | **WRONG — returns successfully, rel. error 242** |
| **80, 88, 96, 104, 112, 120** | **IMA** |
| **128** | **ok** — ~3e-4 |
| **136** | **WRONG — returns successfully, rel. error 285** |
| **144, 152, 160, 168, 176, 184** | **IMA** |
| 192, 200, 208, 216, 224 | `E1` at this shape |

`IMA` = illegal memory access. `E1` = `FlashInfer Internal Error: Invalid
configuration : NUM_MMA_Q=… please create an issue`. The `E1` at 192+ is
shape-dependent tile configuration, not this bug — head_dim 192 does run correctly at
other shapes (e.g. `qo_len=1, kv_len=8192` with `disable_split_kv=True`).

Independently, on `BatchPrefillWithRaggedKVCacheWrapper` with
`plan(..., disable_split_kv=True)`, over 4 shape variants each
(`(qo_len, kv_len, num_qo_heads)` = `(128,256,8)`, `(1,8192,8)`, `(2048,2048,8)`,
`(128,256,32)`): head_dim **80, 96, 112 → illegal memory access in 12 of 12 cases**;
head_dim **64, 128 → ok in 8 of 8**, matching the reference. Together with the
single-request path that is **24 of 24** faulting cases for those three head dims,
and the outcome does not depend on shape.

What the measurements and the source agree on: the outcome is decided by
`D = floor(head_dim / 16)`. `D % 4 != 0` → out-of-bounds access. `D % 4 == 0` → the
kernel runs, but is correct only if `head_dim` is *exactly* `16*D`. `D < 4` → `E1`.

### Paged batch prefill: predicted, not yet measured

`page_produce_kv` and `page_produce_kv_on_the_fly` carry the identical offset
arithmetic (line numbers below), so I expect
`BatchPrefillWithPagedKVCacheWrapper` to behave exactly like the ragged wrapper:
shielded by the split-KV merge kernel by default, faulting with
`disable_split_kv=True`. **I have not measured the paged path** — that is a source
reading, not a result.

## Root cause

`HEAD_DIM_QK` arrives as a template parameter and is immediately turned into a tile
count by **integer division** (`include/flashinfer/attention/prefill.cuh:2479`, and
`:4096` / `:4296` for the batch entry points):

```c++
constexpr uint32_t NUM_MMA_D_QK = HEAD_DIM_QK / 16;
constexpr uint32_t NUM_MMA_D_VO = HEAD_DIM_VO / 16;
```

and `KernelTraits` then *redefines* the head dim from that count (`:252-253`):

```c++
static constexpr uint32_t HEAD_DIM_QK = NUM_MMA_D_QK * 16;
```

### Regime 1 — `head_dim % 16 != 0` → silent truncation

`head_dim = 72` becomes `NUM_MMA_D_QK = 4`, i.e. the kernel is built for head_dim
**64**, while the tensors it is handed still have 72 elements per row and the output
tensor is still 72 wide. It attends over the first 64 channels, writes 64 of the 72
output columns, and leaves the remaining 8 columns of every row at whatever the
freshly-allocated output buffer contained. No error is raised at any layer.

This is the class I would prioritise: the caller gets numbers back, and they are
wrong.

For what it is worth, the `% 16` half of the invariant *is* enforced elsewhere in the
tree — `csrc/page.cu:199` has `TVM_FFI_ICHECK_EQ(head_dim % 16, 0);`. Prefill has no
equivalent in C++ or Python; `csrc/single_prefill.cu:41` reads
`unsigned int head_dim_qk = q.size(2);` and checks nothing.

### Regime 2 — `head_dim % 64 != 0` → out-of-bounds shared memory access

The staging loops walk a swizzled shared-memory offset column-by-column and then
subtract exactly one row width to return to the row start. The column loop is bounded
by another integer division, so it only reaches the row end when the tile count is a
multiple of 4 — i.e. when `head_dim` is a multiple of **64**. `load_q_global_smem`:

```c++
for (uint32_t mma_do = 0; mma_do < KTraits::NUM_MMA_D_QK / 4; ++mma_do) {   // :931
  q_smem->template load_128b_async<...>(q_smem_offset_w, q_ptr, ...);
  q_smem_offset_w = q_smem->template advance_offset_by_column<8>(q_smem_offset_w, mma_do);
  q_ptr += 8 * upcast_size<DTypeQ>();
}
q_smem_offset_w =
    q_smem->template advance_offset_by_row<4, UPCAST_STRIDE_Q>(q_smem_offset_w)
    - 2 * KTraits::NUM_MMA_D_QK;                                            // :938-940
```

Offsets are counted in 16-byte `b128_t` units (`permuted_smem.cuh:74-125`) and
`advance_offset_by_column<8>` is `offset + 8`. With `D = NUM_MMA_D_QK` and fp16,
`UPCAST_STRIDE_Q = head_dim/8 = 2*D`, so the loop advances the column offset by
`8 * floor(D/4)` while the row step subtracts `2*D`. Those cancel **iff `D % 4 == 0`,
i.e. `head_dim % 64 == 0`**. At `head_dim = 96` (`D = 6`) the loop advances 8 and the
code subtracts 12, so the offset drifts −4 units per row group instead of returning to
the row start, and accesses computed from it leave the buffer.

The same `loop-then-subtract-a-full-row` pattern, with the same assumption, appears in
five places:

| function | loop bound | backtrack |
|---|---|---|
| `load_q_global_smem` | `prefill.cuh:931` | `:938-940` |
| `produce_kv` (ragged) | `:483` | `:495-496` |
| `page_produce_kv` (paged) | `:555` | `:567-569` |
| `page_produce_kv_on_the_fly` | `:639` | `:650-651` |
| `write_o_reg_gmem` | `:2006` | `:2014-2015` |

All three FA2 prefill kernels (`SinglePrefillWithKVCacheKernel:2448`,
`BatchPrefillWithRaggedKVCacheKernel:2682`,
`BatchPrefillWithPagedKVCacheKernel:4057`) go through these.

Two further observations:

1. **The invariant is already asserted in this very file**, only on paths the default
   configuration never instantiates:
   `static_assert(KTraits::NUM_MMA_D_QK % 4 == 0, "NUM_MMA_D_QK must be a multiple of
   4")` at `prefill.cuh:955`, `:991` and `:1278` — inside
   `q_smem_inplace_apply_rotary`, `q_smem_inplace_apply_rotary_with_pos` and
   `compute_qk_fp4_rope`. With `pos_encoding_mode="NONE"` none of them is compiled, so
   nothing stops the loader that violates the same invariant from building.
2. **`backend="auto"` actively routes these head dims into the affected kernel.**
   `determine_attention_backend` (`flashinfer/utils.py:571-576`) consults
   `is_fa3_prefill_head_dim_supported` (`:452` — `{64,128,256}` ∪ `{(192,128)}`) and,
   when it fails, returns `"fa2"`. There is no FA2-side counterpart to that predicate,
   so `head_dim=96` ends up in FA2 *because* FA3 declined it.

### Why the batch default path looks fine

`BatchPrefillWith{Ragged,Paged}KVCacheWrapper` reach the same kernel, but with
split-KV enabled the merge step (`VariableLengthMergeStates`,
`attention/cascade.cuh:687`, wrapped in `DISPATCH_HEAD_DIM` — `utils.cuh:194`, which
instantiates only `{64,128,256,512}`) rejects the head dim first, so you get a clean
`Unsupported head_dim: 96`. Pass the documented `plan(..., disable_split_kv=True)` and
the fault returns. The guard that exists lives in a kernel only some paths call.

## `compute-sanitizer`

memcheck localises regime 2, and the controls come back clean — which is what makes
the record quotable (`--report-api-errors no --check-api-memory-access no`; without
those, library-init API noise buries everything):

```
========= Invalid __shared__ read of size 16 bytes
=========     at void flashinfer::SinglePrefillWithKVCacheKernel<
                 flashinfer::KernelTraits<(flashinfer::MaskMode)1, 128u, 2u, 4u,
                 6u, 6u, 4u, 1u, (flashinfer::PosEncodingMode)0,
                 __half, __half, __half, float, int,
                 flashinfer::DefaultAttention<0,0,0,0>>, Params>(T2)+0x55e0
=========     by thread (31,3,0) in block (0,0,1)
=========     Access to 0xc420 is out of bounds
```

```
head_dim=96,  single_prefill      -> 48 errors, all "Invalid __shared__ read of size 16 bytes"
head_dim=64,  single_prefill      ->  0 errors      (control)
head_dim=96,  ragged, no split-KV -> 16 errors, same signature
head_dim=128, ragged, no split-KV ->  0 errors      (control)
```

`NUM_MMA_D_QK = NUM_MMA_D_VO = 6` in that mangled name is `head_dim = 96`, and 16
bytes is exactly one `b128_t` — the unit the swizzled offsets above are counted in.

Caveat: I did not map the SASS offset `+0x55e0` back to a source line (the module is
built without `-lineinfo`), so the sanitizer establishes *an* out-of-bounds shared
access in this kernel; the arithmetic above is what identifies the offset-drift family
as the mechanism, and **which of the five sites emits this particular access is not
proven.**

## Suggested resolution (your call)

* **Reject early.** Give FA2 prefill the counterpart of
  `is_fa3_prefill_head_dim_supported` and check `head_dim_qk` / `head_dim_vo` at the
  top of `single_prefill_with_kv_cache` and in `plan()`, before JIT compilation. For
  fp16 the predicate the kernel actually implements is `head_dim % 64 == 0` (fp8 KV
  would need `% 128` by the same arithmetic, since the loop divisor becomes
  `8 / sizeof(DTypeKV) == 8` — that is inference; I did not measure fp8). That is
  strictly better than today in every regime: the illegal access becomes an actionable
  error, the silent wrong answer becomes an actionable error, the `Invalid
  configuration … please create an issue` becomes an actionable error, and the user
  stops paying a 1–2 minute JIT compile before each failure.
* **Or support them**, by making the five loops iterate to the row end with a
  partial-tile tail. Note that this also has to fix the truncation in
  `NUM_MMA_D = HEAD_DIM / 16`, not just the offset drift.

If "reject" is the answer, I am happy to send that patch — it is small and the test
matrix above already exists. The only thing I would need from you is which predicate
should be authoritative.

Related but not a duplicate: #3176 / #3467 are the same class of defect in the same
function (an unvalidated Python-level input reaching the kernel) for `custom_mask`
rather than `head_dim`.

## What I am not claiming

* Not that upstream intends to support head_dim 72/80/96/112 — the evidence says the
  opposite. The report is that the failure mode is a fault or a wrong number instead
  of a rejection.
* **Not that the illegal access corrupts other allocations.** I looked for that
  specifically: sentinel guard regions around Q/K/V and (on the batch path, via
  `run(out=...)`) around the output, plus allocator-adjacency canaries around the
  output on the `single_prefill` path — 4 shape variants × 3 bad head dims × 2 code
  paths, `CUDA_LAUNCH_BLOCKING=1`, one process each. Every regime-2 case faulted;
  **no case was observed in which such a call returned successfully and memory outside
  the tensors had been modified**, and every 64/128 control ran with zero guard
  violations. What the sanitizer reports is an out-of-bounds *read* of shared memory,
  which is consistent with that. Once the fault fires the CUDA context is dead, so the
  guards cannot be read back for the failing cases — the claim is "no out-of-tensor
  corruption observed", not "impossible".
* The regime-1 wrongness is *inside* the output tensor (columns the kernel never
  writes), not corruption of neighbouring memory — guard regions were clean there too.
* The sweep is partial: 28 head dims, single-request path, plus 5 head dims × 4 shapes
  on the ragged batch path. The regime boundaries are consistent across everything
  measured and follow from the source arithmetic, but I have not enumerated the whole
  range yet.
* The paged path is a source reading, not a measurement.
* fp8 / NVFP4 KV, FA3, and `head_dim_qk != head_dim_vo` combinations were not probed.

## Environment

FlashInfer `main` @ `668a1ba1ca86432c79f6adad37ecfce8d06ec083` (0.6.17), built from
source, JIT only, on an H200 (sm90), CUDA 13.0, driver 580.65.06,
torch 2.13.0+cu130, Python 3.13. One fresh process per measurement.

lazypool:
## 📌 Description

Fixes #4299 — FA2 prefill never validated `head_dim`. The Python layer JIT-compiles a kernel per `head_dim` with no check, and the FA2 prefill kernels only work for head dims divisible by 64:

- `NUM_MMA_D_QK = HEAD_DIM_QK / 16` (`include/flashinfer/attention/prefill.cuh`) truncates via integer division, and the smem swizzle offsets (`include/flashinfer/permuted_smem.cuh`, `get_permuted_offset` / `advance_offset_by_column`) are consistent only for 64-multiples.
- `DISPATCH_HEAD_DIM` (`include/flashinfer/utils.cuh`) guards only the precompiled-cubin path (64/128/256/512), never the JIT path.

**The fix:** new `_check_head_dim(head_dim_qk, head_dim_vo)` in `flashinfer/utils.py` raising `ValueError` for non-multiples of 64, called before any JIT module build at every FA2 prefill entry point:
- `single_prefill_with_kv_cache` (prefill.py:1378)
- `BatchPrefillWithPagedKVCacheWrapper.workspace_size` (prefill.py:1970) and `plan` (prefill.py:2249)
- `BatchPrefillWithRaggedKVCacheWrapper.plan` (prefill.py:3365)

All supported dims (64/128/256/512, MLA 512/576/640) are multiples of 64, so no valid use is rejected.

## 🔍 Related Issues

- #4299

### Before the fix (reproduced on RTX 4090, sm_89)

Reproducer: random fp16 `q[128,8,d]`, `k/v[256,8,d]`, causal FA2 prefill via `single_prefill_with_kv_cache`, compared against an fp32 einsum reference.

| head_dim | kernel returns | sync | rel_err | verdict |
|---|---|---|---|---|
| 64 | yes | OK | 3.788e-04 | correct (fp16 accumulation noise) |
| 72 | yes | OK | **1.115e+00** | **silent wrong results** (garbage output, no NaN/Inf) |
| 96 | no | — | — | **illegal memory access** |

Raw outputs:

```
head_dim=96: launching FA2 prefill
RuntimeError: Check failed: (status == cudaSuccess) is false:
SinglePrefillWithKVCache kernel launch failed, error: an illegal memory access was encountered

head_dim=72: launching FA2 prefill
head_dim=72: kernel returned
head_dim=72: sync OK
head_dim=72 rel_err=1.115e+00 nan_or_inf_in_out=False
```

### After the fix (same host)

| head_dim | before | after |
|---|---|---|
| 64 | rel_err 3.788e-04 | unchanged, rel_err 3.788e-04 |
| 96 | illegal memory access | `ValueError: head_dim_qk and head_dim_vo must be multiples of 64, got head_dim_qk=96, head_dim_vo=96` (raised at prefill.py:1378, before any kernel/GPU work) |
| 72 | rel_err 1.115e+00 (silent garbage) | same `ValueError` with `head_dim_qk=72, head_dim_vo=72` |

Also unit-checked the helper directly: 64/128/256/512/576 accepted; 72/96/32/80 rejected. `python -m py_compile` passes on both files.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

- [ ✅] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [ ✅] I have installed the hooks with `pre-commit install`.
- [ ✅] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.
- [ ✅] Tests have been added or updated as needed.
- [ ✅] All tests are passing (`unittest`, etc.).

## Reviewer Notes

- No public API change; new internal helper follows the existing `_check_*` convention in `flashinfer/utils.py`.
- For NVFP4 KV caches (`kv_cache_sf`) the packed V's last dim is `head_dim//2`, so the single-prefill check uses `q.shape[-1]` and `out_head_dim` (which is `q`'s dim on that path).
- `flashinfer/decode.py` has the same latent issue (decode.py:612, 2006); left as a follow-up to keep this PR scoped to prefill.

## Bug Fixes

* Added validation for attention head dimensions during single, paged, and ragged prefill operations.
* Invalid or unsupported dimensions now produce clear errors before processing begins.
* Improved backend selection and workspace planning for configurations with separate query/key and value dimensions.
* Correctly validates packed value dimensions in NVFP4 prefill operations.
* Expanded validation coverage across supported prefill configurations.


lazypool (commented):
All codes have been reviewed manually. Please give this a check if you have any chance. Thanks!

Smallfu666:
Thanks for turning this round so fast — and for confirming it independently on different hardware. Your sm_89 numbers line up with what I measured on sm90 (head_dim 72 giving a relative error of order 1, 96 faulting), and your predicate matches my own sweep exactly: I swept head_dim 8…288 in steps of 8 across single, ragged and paged prefill (108 cases), and `floor(head_dim/16) % 4 == 0` — i.e. your `% 64` — held with no counterexample. I also checked every other backend's head-dim predicate (FA3 `{64,128,256}` ∪ `(192,128)`, cuDNN qk ∈ `{128,192}` / vo 128, trtllm/sage `{64,128,256}`, MLA 512/576/640) and nothing legitimately supported anywhere is a non-multiple of 64, so I don't see a regression risk on the reject side. Placement before backend resolution and before any JIT build looks right too.

Two things I'd raise:

**1. Is `% 64` sufficient once KV is fp8?** By the same arithmetic the predicate looks dtype-dependent. `produce_kv` (`include/flashinfer/attention/prefill.cuh:483`, backtrack at `:495-496`) advances the column offset by `8 * floor(NUM_MMA_D / (8/sizeof(DTypeKV)))` while the row step subtracts `sizeof(DTypeKV) * NUM_MMA_D`. For 2-byte KV those cancel iff `D % 4 == 0` (head_dim % 64), which is what you have. For **1-byte KV** it becomes `8 * floor(D/8)` versus `D`, cancelling only when `D % 8 == 0` — i.e. `head_dim % 128 == 0`. That would mean `% 64` still admits fp8 head_dim 64 / 192 / 320 / 448. Worth noting head_dim 64 with fp8 KV looks degenerate rather than merely misaligned: `floor(4/8) == 0`, so the inner loop body never executes, and `SharedStorageQKVO::USE_KV_REPACK` (`prefill.cuh:175`) is explicitly `false` for `HEAD_DIM_VO == 64`, so that configuration can't escape via the repack path either.

I want to be clear that I have **not** measured this — my sweep was fp16 KV throughout, and I don't think you tested fp8 either. So this is a question about whether the check should key off `sizeof(DTypeKV)` rather than a claim that something is broken. It seemed better to raise it than to let a `% 64` check imply fp8 is covered.

**2. The PR body ticks "Tests have been added", but no file under `tests/` changed.** Your manual verification looks right and matches the failure modes, it just isn't in the repo. Upstream `tests/attention/*.py` parametrise head_dim over `{64,128,256}` only, so there's no existing negative coverage to lean on and a negative-parametrisation case would be cheap to add.

Minor: `% 64` is necessary but not sufficient in the other direction — 192/320/448 are 64-multiples that no FA2 config actually supports and that still land on the `Invalid configuration : NUM_MMA_Q=…` pre-launch throw. Not a defect (that throw is clean and pre-launch), but the error text's "only support head dims divisible by 64" reads as slightly stronger than what holds.

bvolpato (commented):
Tested this PR head on RTX 5070 Ti (SM120) with PyTorch 2.13 and CUDA 13. FA2 head dim 64 matched my fp32 reference with max absolute error 3.33e-4. Head dims 72 and 96 both raised the intended `ValueError` before JIT. FP8 KV at head dim 64 also matched the dequantized fp16 reference exactly in my run, so I could not reproduce the concern that FP8 needs a `% 128` check.

Two things before approval:

1. `_check_head_dim` accepts zero and negative multiples of 64. With zero-sized Q heads, single prefill calculates the default scale first and raises `ZeroDivisionError` instead of the intended validation error. Please reject non-positive dimensions and run this check before calculating `sm_scale`.
2. This behavior needs a checked-in regression test. No test file changed, and existing attention tests only exercise supported dimensions. Focused coverage for 64, 72, 96, and zero through single prefill would protect the new guard. Plan-path coverage would be useful too if it stays small.

I also confirmed the open `workspace_size()` concern: SM90 auto selection can choose FA3 when dimensions are omitted, while `plan()` chooses FA2 for 512/512 after receiving them. That mismatch predates this PR, so I would track or fix it separately rather than attribute it to this patch.

lazypool (commented):
@Smallfu666 @bvolpato @coderabbitai Thanks for the detailed review. I addressed the remaining points in follow-up commit 514d7f4c:

- `_check_head_dim()` now rejects non-positive dimensions as well as non-multiples of 64, and single prefill validates before default `sm_scale` calculation. A zero-sized Q head now raises `ValueError` instead of `ZeroDivisionError`.
- Added checked-in regression coverage for single prefill head dimensions 64, 72, 96, and 0; invalid paged `workspace_size()` and `plan()` calls; and invalid ragged `plan()` calls.
- Paged `workspace_size()` now forwards `head_dim_qk` and `head_dim_vo` to backend selection, matching `plan()`.

I kept the `% 64` guard for FP8 KV. @bvolpato's SM120 result confirms that FP8 KV with `head_dim=64` is valid, so requiring `% 128` would reject a working configuration. The error text now frames `% 64` as the FA2 safety requirement, rather than implying every 64-multiple is a generally supported FA2 configuration.

Validation: `pytest -q tests/attention/test_prefill_head_dim_validation.py` passes 8/8 on RTX 4090 (SM89), PyTorch 2.6.0+cu124.

Smallfu666:
Thanks — I checked the follow-up changes in 514d7f4c. They address the validation-order and regression-coverage points I raised. No further comments from me.

bkryu:
@flashinfer-bot run

bkryu (commented):
Thanks @lazypool, the diagnosis and the fix looks correct to me. Running the CI so we can get it merged.

bkryu:
@lazypool, the CI results are indicating that `tests/attention/test_sm120_prims_prefill_backend.py` is failing with

```
  - tests/attention/test_sm120_prims_prefill_backend.py::test_cuda_graph_reads_updated_caller_block_table
    E   ValueError: head_dim_qk and head_dim_vo must be positive multiples of 64, got head_dim_qk=32, head_dim_vo=32. The FA2 prefill kernels only safely handle head dims divisible by 64; other dims lead to out-of-bounds accesses or silently wrong results.
    log: /tmp/junit/shards/shard-0000/units/unit-b09233b1f16ef331/batches/batch-963af618bf57b129.log
  - tests/attention/test_sm120_prims_prefill_backend.py::test_cuda_graph_rejects_uncompiled_specialization
    E   ValueError: head_dim_qk and head_dim_vo must be positive multiples of 64, got head_dim_qk=32, head_dim_vo=32. The FA2 prefill kernels only safely handle head dims divisible by 64; other dims lead to out-of-bounds accesses or silently wrong results.
```
Can you look into it?

flashinfer/prefill.py
```diff
@@ -1492,6 +1492,7 @@ def single_prefill_with_kv_cache(
     """
     _check_pos_encoding_mode(pos_encoding_mode)
     _check_kv_layout(kv_layout)
+
```
lazypool:
```suggestion
```

lazypool (commented):
@bkryu  It seems the previous commit ignored backends other than FA2.

For example, cute-dsl-prims supports `head_dim=32`, but are still incorrectly rejected.

Therefore, the new fix adds a conditional check, checking for a multiple of 64 only when the backend is FA2. The `head_dim` constraints for the various backends are as follows.

| Backend | head_dim constraint | Guard |
|---------|---------------------|-------|
| FA2 | Multiples of 64 | `_check_head_dim` (our fix) |
| FA3 | 64, 128, 256, (192,128) | Whitelist |
| cuDNN | (128,128), (192,128), (256,256) | Whitelist |
| CUTLASS | (128,128), (192,128) | Whitelist |
| fmha_v2 | 64, 128, 256 | Whitelist |
| cute-dsl-prims | Any | None needed |
| trtllm-gen | Own constraints | Whitelist |


Note that the backends marked as whitelist will not encounter the `head_dim` issue under FA2 at all, because they reject unsupported dims themselves. But cute-dsl-prims is special, and its why only it encounters the problem.

So now `_check_head_dim` will additionally accept a parameter, `backend`. Therefore, it is appropriate to place it after the backend is resolved. That is why I also moved the computation of `out_head_dim` back to its original position _(it was incorrectly moved up in a merge commit)_, which better follows _the principle of minimizing changes_.

lazypool:
and for @bkryu , there are still 2 questions that should be attentioned:

1. **ZeroDivisionError vs ValueError** I may place _check_head_dim() before everywhere `1.0 / math.sqrt(q.size(-1))`, which will change ZeroDivisionError to a more detailed ValueError (like what @bvolpato wants), but it will make the code more chaos... I'm not sure if it's worthwhile because I think the ZeroDivisionError is already clear enough...

2. **Decode path** `flashinfer/decode.py` has the same FA2 head_dim issue (lines612, 2006). The PR description scopes this to prefill only. Should we also fix decode in this PR, or as a follow-up?

Looking forward to your valuable feedback. :-)

PR #4313, opened 2026-08-01
feat(cake_kda): share recurrent prefill kernels across SM100 family

yyihuang:
> [!IMPORTANT]
> Follow-up to merged #4262. This PR contains only SM100-family packaging,
> routing, validation, and GB300 enablement; it does not add another frozen
> prefill kernel body.

## Description

Compile the existing frozen recurrent-KDA prefill M64 and M128 kernels once as
`sm_100f` on CUDA 12.9 and newer, and use those modules on both CC 10.0
(B200/GB200) and CC 10.3 (B300/GB300). CUDA 12.8 B200 keeps the exact
`sm_100a` compatibility modules because that toolkit predates `sm_100f`.

The M64 and M128 generated CUDA bodies remain byte-identical to #4262. JIT and
AOT use target-bearing module identities, and the binding accepts the family
module only on CC 10.0 or CC 10.3.

## Validation

Publication commit: `8d3bbb27e26795609638f8002d345917c6d802c5` (tree
`c859b6d5241b5994cc9ccf4ceb5faacce15f4057`), based directly on upstream
`main@4433996e` after #4279 merged. The incremental prefill implementation is
identical to the candidate validated on both GPUs: the ten non-AOT files are
blob-identical, and their stable patch-id is unchanged. The only replay
resolution was mechanical in `flashinfer/aot.py`: it preserves the merged
#4279 decode registration while applying the same prefill family registration.

| Gate | B200 | GB300 |
|---|---:|---:|
| Public prefill GPU tests | 39/39 passed | 39/39 passed |
| Targeted CPU tests at publication commit | 26 passed, 30 skipped | same commit |
| Pre-commit at publication commit | passed | same commit |

Strict CUPTI, cold-L2, balanced-process A/B below compares the family target
with the same frozen bodies compiled for the exact device target. Speedup is
`exact / sm100f`; values near 1 mean target consolidation is performance
neutral.

| GPU / run | Historical six | H6 four | All ten |
|---|---:|---:|---:|
| B200 | 0.9994x | 0.9992x | 0.9993x |
| GB300 primary ABBA | 0.9996x | 0.9905x | 0.9959x |
| GB300 reverse-BAAB confirmation | 0.9999x | 0.9957x | 0.9982x |

The only stable tiny-shape difference is `h6_packed_n8_t128`: `sm100f`
25.664 us versus exact `sm103a` 25.360 us (0.9882x, 0.304 us absolute).
Keeping an extra exact M128 target for this one coordinate would reintroduce a
third target and shape-specific routing, so this PR retains the simpler shared
family module.

The existing six-shape comparison with pinned official FlashKDA remains the
performance baseline documented by this PR; the new A/B isolates only the
physical target change.

Related to #4254.

## Summary by CodeRabbit

* **New Features**
  * Added recurrent KDA prefill support for SM100a and SM103a GPUs.
  * Added automatic selection of compatible kernels based on GPU architecture and CUDA version.
  * Added support for smaller head counts through automatic beta-data padding.
  * Improved hardware information included in benchmark results.

* **Bug Fixes**
  * Improved device and input validation, shared-memory checks, and launch error reporting.

* **Documentation**
  * Updated KDA prefill documentation with supported architectures and CUDA requirements.

### DIFF
--- a/benchmarks/bench_recurrent_kda_prefill.py
+++ b/benchmarks/bench_recurrent_kda_prefill.py
@@ -50,6 +50,7 @@ from flashinfer.utils import get_compute_capability
 FLASH_KDA_PEER_COMMIT = "d2ff19a6a0c82f39f796f637ebd1c36090b1268f"
 FLASH_KDA_CUTLASS_COMMIT = "5c149f52a436782210263fb2f19b354443a61c6a"
 DEFAULT_STATE_ROTATIONS = 512
+SUPPORTED_FLASH_KDA_ARCHS = {(10, 0): "sm100a", (10, 3): "sm103a"}
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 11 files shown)


yzh119:
What's the difference between this sm103 kernel compared to the sm100 kernel, any B300 specific features being used here?

Also, where does the host-side overhead coming from, we should remove that overhead.

yzh119 (approved):

PR #4314, opened 2026-08-02
feat(cake_kda): share recurrent decode kernels across SM100 family

yyihuang:
> [!IMPORTANT]
> Follow-up to merged #4279. This PR is a single incremental SM100-family
> target-consolidation commit based directly on upstream `main@4433996e`.

## Description

Compile the frozen recurrent-KDA decode portfolio as `sm_100f` on CUDA 12.9+
and share it between CC 10.0 (B200/GB200) and CC 10.3 (B300/GB300). The public
`backend="cake"` contract and all device split selectors remain unchanged.

Two physical-target exceptions are deliberate:

- CUDA 12.8 B200 retains the exact `sm_100a` portfolio because `sm_100f` is
  unavailable in that toolkit.
- GB300 retains exact `sm_103a` modules for the two direct-T1 variants. A
  repeated target-only A/B found a systematic approximately 2.1% regression
  when those variants were compiled as `sm_100f`.

All 23 frozen generated CUDA bodies are byte-identical to #4279. CUDA 12.9+
B200 registers 23 family modules. CUDA 12.9+ GB300 registers the family
portfolio plus two exact direct-T1 modules; T2-T6 use the family modules.

## Validation

Publication commit: `e2552d7b0671b6694564053d22f67a7f82ef023f` (tree
`d3312b2fbd8b52544f0a1ba6c9f1b8b318ca9aad`), based directly on upstream
`main@4433996e`. Its tree is byte-identical to ready commit `0c8212ad` and to
the candidate validated on both GPUs.

| Gate | B200 | GB300 |
|---|---:|---:|
| Public decode GPU tests | 151/151 passed | 151/151 passed |
| Targeted CPU tests at publication commit | 194 passed, 20 skipped | same commit |
| Pre-commit at publication commit | passed | same commit |

Strict CUPTI, cold-L2, balanced-process A/B below compares the final family /
hybrid build with the exact-target implementation. Speedup is
`exact / family-or-hybrid`.

| GPU | T1 | T2 | T3 | T4 | T5 | T6 | 30-shape geomean |
|---|---:|---:|---:|---:|---:|---:|---:|
| B200 | 1.0043x | 1.0031x | 1.0029x | 0.9954x | 1.0003x | 1.0006x | **1.0011x** |
| GB300 | 1.0015x | 0.9988x | 1.0138x | 1.0035x | 0.9998x | 0.9985x | **1.0027x** |

GB300 T1 in this final table is exact `sm103a` on both sides; the row verifies
that retaining the performance exception does not perturb public dispatch.
The existing #4279/upstream and CAKE-internal baseline measurements remain the
algorithm-level performance comparison; this A/B isolates physical-target
consolidation.

Follow-up to merged #4279. Related to #4254.

## Summary by CodeRabbit

- **New Features**
  - Added Flash-KDA decode support for SM103a and SM100-family targets, including direct single-token variants.
  - Added architecture-aware JIT and AOT module generation, loading, caching, and dispatch.
  - Added support for architecture-specific scheduling and value-split policies.

- **Bug Fixes**
  - Improved validation for GPU architecture and CUDA version compatibility.
  - Unsupported target and variant combinations are now rejected earlier.

- **Documentation**
  - Updated recurrent KDA guidance with supported architectures, compilation requirements, scheduling policies, and performance details.

- **Tests**
  - Expanded coverage for target selection, routing, compatibility checks, and numerical correctness.

### DIFF
--- a/benchmarks/bench_recurrent_kda_decode_export.py
+++ b/benchmarks/bench_recurrent_kda_decode_export.py
@@ -33,6 +33,7 @@ import statistics
 import subprocess
 from importlib.metadata import PackageNotFoundError, version
 from pathlib import Path
+from typing import Optional
 
 import torch
 import torch.nn.functional as F
@@ -47,11 +48,11 @@ EVOLUTION_PEER_SHA = "cea7f46ffc190cabf82c95a39cd0d2aa6c888c17"
 DATA_SEED = 42
 PRECOMPUTED_SEQUENCE_COUNTS = (8, 16, 32, 64, 128)
 T3_LOWER_BOUND_SEQUENCE_COUNTS = (1, 2, 4, 8, 16)
+SUPPORTED_FLASH_KDA_DECODE_ARCHS = {(10, 0): "sm100a", (10, 3): "sm103a"}
 
-# Keep the measured public-export routes in one place. T1 uses a direct
-# register-state body with split16 at N8 and split8 at N16+; the exported split
-# sweep selects split4 for T2 and split2 for T4. T3 has one exact lower-bound
-# split4 specialization.
+# Keep the measured public-export routes in one place. T3 has one exact
+# lower-bound split4 specialization. The final 30-shape matrix has
+# architecture-local T1 routes and one measured SM103a T4/N8 route override.
 EXPECTED_VARIANTS_BY_T = {
     2: "d128_t2_precomputed_split4",
     3: "d128_t3_lower_bound_split4",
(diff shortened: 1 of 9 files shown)


benchmarks/bench_recurrent_kda_decode_export.py
```diff
@@ -47,24 +48,33 @@
...
-    128: "d128_t1_precomputed_direct_split8",
+EXPECTED_T1_VARIANTS_BY_ARCH_AND_N = {
+    "sm100a": {
+        8: "d128_t1_precomputed_direct_split16",
+        16: "d128_t1_precomputed_direct_split8",
+        32: "d128_t1_precomputed_direct_split8",
+        64: "d128_t1_precomputed_direct_split8",
+        128: "d128_t1_precomputed_direct_split8",
+    },
+    "sm103a": {
```
yzh119, line 71: why the config here on sm103 is different to sm100?

yyihuang: This table mirrors the device-local runtime split policy; it is not a correctness or ISA constraint.

For T=1, split16 launches 16 CTAs per sequence/value-head with 8 value rows per CTA, while split8 launches 8 CTAs with 16 rows per CTA. Both variants use the same math and passed full output/state correctness.

We retained the B200 policy measured for #4279 and independently swept the same frozen variants on GB300. On B200, split16 won at N=8 (6.160 vs 7.024 us), while split8 won at N=16 (8.944 vs 9.521 us), so its policy crosses over after N=8. On GB300, split16 won through N=128. In the 12-round cyclic-order cold-L2 CUPTI confirmation, its median-paired speedups over split8 at N=16/32/64/128 were 1.1186x, 1.0663x, 1.0602x, and 1.0282x, with 12/12 wins at every point.

Therefore SM103 uses a separately measured dispatcher policy. Both architectures export the same frozen source variants; only runtime route selection differs.

yzh119 (approved):

PR #4319, opened 2026-08-03
fix(moe): pad BF16 TRTLLM-Gen intermediates to 128 KiB

nvpohanh:
[by Codex]

## Summary

- pad the BF16 TRTLLM-Gen MoE GEMM1 and GEMM2 intermediate allocations to at least 128 KiB
- reuse `Routing::maybeGetMinTokenCount`, matching the existing FP8/MXFP launchers
- calculate the GEMM1 and GEMM2 row counts independently because their hidden dimensions can differ

## Why

On Blackwell, the TRTLLM-Gen BMM2 kernel can use TMA `BASE_128KB` address generation based on the logical tensor-map shape. With a small eager decode batch, the BF16 launcher currently allocates only the logical 16/32-KiB intermediate. That leaves less than 128 KiB mapped after the activation base and can produce an illegal memory access in `UTMALDG.4D`.

The quantized launchers already call `maybeGetMinTokenCount` for both intermediates. Applying the same allocation policy to the BF16 launcher gives the TMA transaction the required backing without changing the logical token count passed to routing or GEMM.

## Validation

- B200x8 patched-JIT compile/load preflight
- six consecutive Miles/SGLang rollout runs with the original reproducer setting:
  - pinned image and model revision
  - FlashInfer TRTLLM backend, TP8
  - four train/rollout steps per run
  - 256 one-token post-update health probes per step boundary
  - no descriptor-bit override, TMA-OOB disablement, extra allocation-padding knob, or synchronization workaround

Result: six consecutive valid passes on B200x8. Every run exited normally, completed all four steps, passed the diagnostic evidence gate, and had zero IMA/sanitizer/segfault matches.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved BF16 Mixture-of-Experts processing by ensuring intermediate buffers have sufficient capacity for required token mappings.
  * Increased reliability for GEMM-based operations across varying tensor dimensions while preserving expected output shapes and device placement.
  * Helps prevent capacity-related processing failures in workloads with larger or uneven token distributions.


nvpohanh:
This will fix #3427

nvpohanh:
@leejnau could you add this PR into the tracker? thanks!

aleozlx (approved):

b8zhong:
Hi, is there anything blocking the merge of this PR? It occurs during GLM 5.2 draft MoE, so ideally I'm wondering if it can be merged soon

henrylhtsang:
@nvpohanh dumb question why not consider the cutlass fix for tma descriptor? since it seems like it is the same issue? or am I wrong?

```
  if (driver_version <= 13010) {
    if (cute::bits_to_bytes(
          cute::cosize(tensor_cwhdn.layout()) *
          cute::sizeof_bits<typename EngineA::value_type>::value) < 131072) {
      reinterpret_cast<uint64_t*>(&tma_desc)[1] &= ~(1llu << 21);
    }
  }
```

https://github.com/NVIDIA/cutlass/blob/6c68991985ca8b09594ac6fd43abbfd5830c4140/include/cute/atom/copy_traits_sm90_im2col.hpp#L474-L483

nvpohanh:
TBH I think that fix is a little ugly 😹 And padding workspace to 128KiB should be almost harmless

nlima-prog:
### Hatch Session

This PR was worked on in a Hatch coding session.

| | |
|---|---|
| **Session** | `6493ed5f-d798-40ac-88c2-392844ffd43c` |
| **User** | nlima-prog |
| **Environment** | Hatch (Local) |
| **Agent** | Abacus (Local) |
| **Sandbox** | local |
| **Time** | 2026-08-25T16:58:07.185Z |

PR #4324, opened 2026-08-03
MSA decode path improvements

rmhaskarnvidia:
## 📌 Description
Two independent changes to the SM120/SM121 MSA ops, both aimed at per-call overhead on the decode path -- 

1. Paged paths rebuilt a prefix sum they did not need 
`msa_proxy_score`, `msa_proxy_score_fp4` and `msa_sparse_decode_attention` each did this on every paged call:
```
cu_k = torch.zeros(batch_size + 1, dtype=torch.int32, device=dev)
cu_k[1:] = seqused_k.to(dev).cumsum(0)
```
so the kernel could do `seqlen_k = mCuK[b + 1] - mCuK[b]` and recover a number the caller already had in `seqused_k`.

On the paged path `page_table` supplies every KV address, so `k_start` is dead — it is read only in the flat branches (`sparse_decode_sm12x.py` lines 470/474/ 633/637/648, and the equivalents in the proxy kernels). Only `seqlen_k` is used, for masking.

The fix passes `seqused_k` straight through and takes the length directly under a `const_expr` paged branch, so there is no runtime cost to the distinction:

```python
if cutlass.const_expr(self._paged):
    k_start = cutlass.Int32(0)
    seqlen_k = mCuK[batch_idx]
else:
    k_start = mCuK[batch_idx]
    seqlen_k = mCuK[batch_idx + 1] - k_start
```
2. `msa_topk_select` takes only batch-wide scalars
.`num_valid_pages` and `force_end_blocks` were ints, so a caller whose query tokens have differing causal KV extents — any decode batch, any chunked prefill — could not express per-token semantics. The only recourse was to post-process on the host. vLLM's MiniMax-M3 integration does exactly that: bias each token's local window with a `scatter_` of `FLT_MAX`, then repair the output with a comparison, a `masked_fill_` and a `sort`.

`num_valid_pages` may now be an int32 tensor of shape `(total_qo_len,)`. The kernel then clamps each token's candidate range to its own extent and forces that token's own trailing `force_end_blocks`, which is what the scatter was emulating. Consequences:

* no selected index can exceed its token's extent, so no masking pass
* the ascending, `-1`-tail-padded contract holds by construction, so no sort
* the forced region is clamped in-kernel against `nvp`, so tokens shorter than the local window are handled instead of underflowing to negative blocks

Dispatch between the count-rank and radix kernels bounds by `max_k_tiles` on this path rather than the valid-page count: reading the tensor on the host would sync the stream, which is illegal under CUDA graph capture.

### Measurements
RTX PRO 6000 Blackwell Server Edition (SM120), MiniMax-M3 per-GPU TP4 indexer geometry (`num_idx_heads` 1, `head_dim` 128, page 128, topk 16, `init_blocks` 0, `local_blocks` 1), context 8192, heterogeneous per-request extents.

Indexer selection (proxy score + top-k), kernel launches and eager wall time per call: 

| batch | launches before | after | before | after |
|-------|-----------------|-------|--------|-------|
| 1 | 27 | 7 | 341.1 us | 107.1 us |
| 4 | 27 | 7 | 337.8 us | 107.1 us |
| 16 | 27 | 7 | 334.8 us | 106.1 us |
| 64 | 27 | 7 | 337.9 us | 108.9 us |

Paged sparse decode: **7 -\> 2** launches per call
Selections are identical to the host-side sequence they replace, verified per-token across the batch.

### End-to-end
vLLM serving MiniMax-M3 NVFP4 on 4x RTX PRO 6000, TP4, ISL 8192 / OSL 1024. . Requires the matching vLLM-side change to use FI backend for MSA

| conc | TPOT Triton | TPOT FlashInfer | ΔTPOT | TTFT Triton | TTFT FlashInfer | ΔTTFT |
|------|-------------|-----------------|-------|-------------|-----------------|-------|
| 1 | 9.28 ms | 9.38 ms | +1.1% | 1030.1 ms | 977.4 ms | −5.1% |
| 4 | 16.17 ms | 16.16 ms | −0.1% | 1229.7 ms | 1171.6 ms | −4.7% |
| 16 | 38.99 ms | 38.24 ms | −1.9% | 2023.9 ms | 1938.4 ms | −4.2% |
| 64 | 112.68 ms | **109.63 ms** | **−2.7%** | 4699.7 ms | **4460.5 ms** | **−5.1%** |

Before these changes the FlashInfer backend was a **+16.7% TPOT regression at concurrency 1 and aborted outright at 4 and 16** with an illegal memory access.

### Tests
New `tests/msa_ops/test_msa_topk_per_token.py`:

* per-token tensor equals invoking the scalar path once per token, across `force_begin`/`force_end` combinations and at both `max_k_tiles` values that straddle the count-rank/radix dispatch boundary (so it cross-checks the two kernels against each other on distinct-score inputs)
* output contract for every token, including extents shorter than the forced region — which the scalar API rejects, so only the in-kernel clamp handles it
* `force_end_blocks` forces each token's _own_ trailing blocks
* input guards
* CUDA-graph capture safety

The existing `tests/msa_ops/` suite passes (63 passed after fixing one test that called the private `_get_compiled_topk` positionally).
## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

- **New Features**
  - Added per-token valid-page counts for top-k page selection, including input validation and bounded results.
  - Improved paged KV-cache handling with direct per-request sequence lengths.
  - Added variable-length paged FP4 decoding for general and packed execution modes.

- **Bug Fixes**
  - Corrected sequence lengths, offsets, and page boundaries across paged and flat cache paths.
  - Improved forced-region clamping for variable-length requests.

- **Tests**
  - Added coverage for heterogeneous sequences, ragged pages, validation, bounded output, and CUDA Graph compatibility.

### DIFF
--- a/flashinfer/msa_ops/proxy_score.py
+++ b/flashinfer/msa_ops/proxy_score.py
@@ -32,9 +32,14 @@ from ._common import _BLK_KV
 _SF_VEC_SIZE = 16
 
 
-def _resolve_proxy_dims(cu_seqlens_q, cu_k, max_seqlen_q, max_k_tiles, output):
+def _resolve_proxy_dims(
+    cu_seqlens_q, cu_k, max_seqlen_q, max_k_tiles, output, k_is_lengths=False
+):
     """Resolve grid dims ``(max_seqlen_q, max_k_tiles)`` as Python ints; deriving
-    them reads ``cu_seqlens`` on host, so under CUDA-graph capture we raise."""
+    them reads ``cu_seqlens`` on host, so under CUDA-graph capture we raise.
+
+    ``k_is_lengths`` says ``cu_k`` holds per-request lengths rather than a
+    prefix sum (the paged path)."""
     if max_k_tiles is None and output is not None:
         max_k_tiles = int(output.shape[1])
     if max_seqlen_q is not None and max_k_tiles is not None:
@@ -50,7 +55,7 @@ def _resolve_proxy_dims(cu_seqlens_q, cu_k, max_seqlen_q, max_k_tiles, output):
         max_seqlen_q = int((cu_q_cpu[1:] - cu_q_cpu[:-1]).max().item())
     if max_k_tiles is None:
         cu_k_cpu = cu_k.cpu()
-        seqlens_k = cu_k_cpu[1:] - cu_k_cpu[:-1]
+        seqlens_k = cu_k_cpu if k_is_lengths else (cu_k_cpu[1:] - cu_k_cpu[:-1])
         max_k_tiles = int((seqlens_k.max().item() + _BLK_KV - 1) // _BLK_KV)
     return max_seqlen_q, max_k_tiles
 
@@ -382,8 +387,12 @@ def msa_proxy_score(
                 f"paged k must be (num_pages, num_kv_heads, {_BLK_KV}, {head_dim})"
             )
         batch_size = seqused_k.numel()
-        cu_k = torch.zeros(batch_size + 1, dtype=torch.int32, device=dev)
-        cu_k[1:] = seqused_k.to(dev).cumsum(0)
+        # Paged addressing comes entirely from page_table, so the kernel needs
+        # only each request's length. Pass seqused_k through instead of building
+        # a prefix sum (zeros + cumsum + slice-copy on every call) for the
+        # kernel to difference back apart; `paged` tells it which form this is.
+        cu_k = seqused_k if seqused_k.device == dev else seqused_k.to(dev)
+        cu_k = cu_k.contiguous()
         pt_dev = page_table.contiguous()
     else:
         if cu_seqlens_k is None:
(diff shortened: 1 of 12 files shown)


flashinfer/msa_ops/proxy_score.py
```diff
@@ -715,8 +727,12 @@ def msa_proxy_score_fp4(
                 f"paged k_fp4 must be (num_pages, num_kv_heads, {_BLK_KV}, {hd2})"
             )
         batch_size = seqused_k.numel()
-        cu_k = torch.zeros(batch_size + 1, dtype=torch.int32, device=dev)
-        cu_k[1:] = seqused_k.to(dev).cumsum(0)
+        # Paged addressing comes entirely from page_table, so the kernel needs
+        # only each request's length. Pass seqused_k through instead of building
+        # a prefix sum (zeros + cumsum + slice-copy on every call) for the
+        # kernel to difference back apart; `paged` tells it which form this is.
+        cu_k = seqused_k if seqused_k.device == dev else seqused_k.to(dev)
```
saltyminty: Wanted to check my understanding: we're passing non-cumulative `seqused_k` instead of cumulative `cu_k`, and `paged` tells the kernels whether to evaluate the tensor as non-cumulative or cumulative? I feel the `cu_k` naming is a bit misleading here

rmhaskarnvidia: That's the correct understanding. cu_k updated to k_len_or_cu to avoid the confusion.

saltyminty (approved):

PR #4328, opened 2026-08-04
feat(moe): add unified CUTLASS BF16 and W4A16 runners

feih-nv:
Expose the existing CUTLASS fused-MoE implementation through the unified API for BF16 and W4A16, with independent GEMM tuning, safe workspace reuse, and CUDA graph support.

Introduce a `build()` method to `MoERunner` to skip the JIT overhead for unsupported cases.

## 📌 Description

Add unified CUTLASS runners for:

- `CutlassBf16Config` → `CutlassBf16Runner`
- `CutlassW4A16Config` → `CutlassW4A16Runner`

`CutlassConfig` remains as a compatibility placeholder, but it is not registered with `MoELayer` and is not directly runnable. We can remove it later after we have all the CUTLASS variants supported (e.g. nvfp4).

CUTLASS BF16 follows the flat API architecture coverage and is currently GPU-validated here on SM90. CUTLASS W4A16 remains SM90-only because it uses Hopper-specific mixed-input layouts.

Key changes for the runners:

- Add BF16 and W4A16 weight preparation and `MoELayer` registration.
- Support packed precomputed routing, SwiGLU, and `do_finalize=True`.
- Tune GEMM1 and GEMM2 independently and combine their winners into one compound tactic.
- Use consistent token buckets for tuning and workspace allocation.
- Include fixed launch configuration such as `top_k` in persistent tuning keys.
- Reject inputs above `tune_max_num_tokens`.
- Retain bucket workspaces referenced by captured CUDA graphs.
- Add BF16 and W4A16 numerical, autotuning, workspace, and CUDA graph tests.

For the API change:
- Defer CUTLASS module loading and inner-runner construction from `__init__()` to the idempotent `build()` method.
- Have `MoELayer` enforce `check_support()` before `build()`, preventing unsupported configurations from triggering JIT compilation.
- Require direct-runner callers to complete the explicit lifecycle before packing, tuning, or execution.
- Ensure execution methods cannot implicitly load modules or query workspace sizes.

### ✅ Pre-commit Checks

- [x] I have installed and configured `pre-commit`.
- [x] I have run `pre-commit run --all-files` and fixed reported issues.

## 🧪 Tests

- [x] Focused tests pass on NVIDIA H100 (SM90).

```bash
python3 -m pytest -q -p no:cacheprovider \
  tests/moe/test_unified_moe_cutlass.py \
  tests/moe/test_unified_moe.py::TestBackendOptions
31 passed, 2 warnings in 5.60s
```


tests/moe/test_unified_moe_cutlass.py
```diff
@@ -0,0 +1,719 @@
...
+    w2 = (
+        torch.randn(
+            num_experts,
+            hidden_size,
+            intermediate_size,
+            device=device,
+            dtype=torch.bfloat16,
+        )
+        / 20
+    )
```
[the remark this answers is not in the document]
feih-nv: Agreed. Fixed in `8f8f4d86` by rescaling the activation and weight fixtures so reference outputs have meaningful magnitude, tightening W4A16 tolerances, and routing all numerical checks through a helper that explicitly verifies an all-zero output fails before comparing the real output.

W4A16 is compared against a reference using dequantized MXFP4 weights, so the updated tolerance measures kernel error against the quantized-weight contract rather than against the original BF16 weights.

tests/moe/test_unified_moe_cutlass.py
```diff
@@ -0,0 +1,719 @@
+"""Unified CUTLASS BF16 MoE adapter tests."""
```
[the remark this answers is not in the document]
feih-nv: Fixed in `8f8f4d86`. The module docstring now covers BF16 and W4A16/MXFP4 tests.

flashinfer/fused_moe/api.py
```diff
@@ -457,16 +465,101 @@ def __repr__(self) -> str:
...
         return "CutlassConfig()"
 
 
+@dataclass(frozen=True)
+class CutlassBf16Config:
+    """CUTLASS BF16 backend for the unified MoE API.
+
+    Architecture coverage follows the dense-BF16 legacy flat API. The unified
+    GPU tests currently exercise SM90.
+    """
```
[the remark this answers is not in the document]
feih-nv: Good point. Updated both public config docstrings to document packed precomputed routing, SwiGLU, `do_finalize=True`, and the lack of EP/shared-expert support. The W4A16 docstring now also documents the 128-divisibility requirements.

flashinfer/fused_moe/api.py
```diff
@@ -480,16 +487,111 @@ def __repr__(self) -> str:
 
 @dataclass(frozen=True)
 class CutlassConfig:
```
Aneureka, line 494: Is this actually used somewhere? I think we can at least deprecate this class.

flashinfer/fused_moe/runners.py
```diff
@@ -190,6 +196,439 @@ def check_support(self) -> None:
...
+            num_tokens, dtype=torch.int32, device=inputs[2].device
+        ).unsqueeze(1)
+        slots = torch.arange(
+            top_k, dtype=torch.int32, device=inputs[2].device
+        ).unsqueeze(0)
+        inputs[2].copy_((token_offsets * top_k + slots) % num_experts)
+        inputs[3].fill_(1.0 / top_k)
+        return inputs
+
+    def get_valid_tactics(self, inputs: List[torch.Tensor], _profile: Any) -> List[Any]:
```
Aneureka, line 341: It seems this API is to list the valid tactics but we currently choose the best one as implementation?

flashinfer/fused_moe/api.py
feih-nv: You’re right that `CutlassConfig` is no longer runnable or registered. It is currently retained only as a compatibility placeholder for existing imports and serialized configurations. We should mark it deprecated now and remove it once the concrete CUTLASS variants cover the remaining contracts. 
It is now used in `test_unified_moe.py`, this PR changed the bf16 tests to use `CutlassBf16Config` but I think there are still tests for nvfp4 and fp8 where we don't have the runners yet.

flashinfer/fused_moe/runners.py
feih-nv: Yes, this is intentional but semantically awkward. CUTLASS exposes independent GEMM1 and GEMM2 tactic spaces; returning every compound pair would turn the search into an O(n1*n2) Cartesian product. The current adapter preserves the legacy O(n1+n2) behavior by tuning each stage independently and returning one compound winner.

The lifecycle update defers module loading until after `check_support()`, and cache hits bypass this path, so the nested search occurs only on cold tuning. Longer term, this should use an explicit factorized/compound-tactic autotuner interface rather than selecting winners inside `get_valid_tactics()`.

flashinfer/fused_moe/api.py
feih-nv: marked as deprecated

flashinfer/fused_moe/runners.py
Aneureka: Got it. It seems the current autotune runner cannot express multi-stage autotuning semantics, so there is a gap we need to address. cc @aleozlx

feih-nv: Agreed. What it does not cover is interaction between stage choices, where the individually fastest tactics may not form the fastest end-to-end combination.

A practical extension would be optional top-k refinement:

1. Retain the top k tactics from each stage.
2. Profile only the k² resulting compound tactics end to end.
3. Cache the best compound tactic.

This gives a search cost of `O(n1 + n2 + k²)`, avoiding a full `O(n1 × n2)` Cartesian search. With k=1, it preserves the current behavior; offline or deeper tuning could use k=2 or k=3.

The cache could be layered accordingly:
```
stage/GEMM1/config → ranked GEMM1 candidates
stage/GEMM2/config → ranked GEMM2 candidates
compound/config    → selected (GEMM1, GEMM2) pair
```
On a compound-cache miss, each stage cache can be checked independently, so if GEMM1 is cached but GEMM2 is not, only GEMM2 needs to be retuned. To support refinement efficiently, the stage cache should retain ranked top-k candidates and timings rather than only the single winner.

The current CUTLASS implementation already approximates this layered structure through separate GEMM1/GEMM2 `choose_one()` calls plus the outer compound cache. The main architectural improvement would be to make staged tuning a first-class autotuner abstraction instead of initiating nested tuning from `get_valid_tactics()`.

I would keep the current independent two-stage implementation for this PR and treat first-class staged tuning, layered ranked caches, and optional top-k end-to-end refinement as a follow-up autotuner enhancement. What does it sound?

Aneureka: Yeah we can leave a `FIXME` comment here to refine the autotuner and unblock this PR.

Aneureka (approved):

Aneureka:
Approved. @aleozlx may take a look as well.

flashinfer/fused_moe/runners.py
feih-nv: Added via commit 1e08dfd06d2ac4a69a1d5039027b146187ca6dba

Issue #4317, opened 2026-08-02
[Bug] sm12x MoE CuTe-DSL kernels are never disk-cached — cute.compile hardcodes no_cache=True, so every fresh engine re-pays 15–40 s MLIR compiles per kernel shape


jhsmith409:
## Summary

The Blackwell sm12x fused-MoE path compiles all of its CuTe-DSL kernels through explicit
`cute.compile(...)`, which unconditionally disables the DSL's file cache. The only caching
layer is three in-process Python dicts, so **every engine start re-pays the full MLIR JIT
cost for every kernel specialization it touches** — measured at **~40 s** for the dynamic
kernel's first hit and **~15–18 s per small-`m` micro-kernel specialization**. In a serving
context (vLLM `--moe-backend=flashinfer_b12x`) these compiles land mid-request and freeze
the whole engine (vllm#47458: ~17.5 s stop-the-world stalls, all in-flight requests frozen
including mid-decode).

`CUTE_DSL_CACHE_DIR` / `CUTE_DSL_DISABLE_FILE_CACHING` have no effect on this path — the
gate that would consult them is short-circuited before they are read.

## Environment

- flashinfer 0.6.12, 0.6.13, and **0.6.15.post1** (all verified; `moe_dispatch.py` cache
  structure unchanged across them)
- nvidia-cutlass-dsl 4.5.0 and **4.6.0** (both verified)
- SM120: RTX 5070 Ti (×2, TP/PP), RTX 5090 (unsharded), RTX PRO 6000 Blackwell — reproduced
  on all three in the vllm#47458 thread
- Model: `nvidia/Qwen3.6-35B-A3B-NVFP4` via vLLM `--moe-backend=flashinfer_b12x`

## Mechanism (exact code paths)

1. **The DSL hardcodes `no_cache=True` for explicit compiles.**
   `CompileCallable._compile` sets `kwargs["no_cache"] = True` unconditionally:
   - nvidia-cutlass-dsl 4.5.0: `base_dsl/compiler.py:604-605`
   - nvidia-cutlass-dsl 4.6.0: `base_dsl/compiler.py:1319`

2. **The file cache is gated behind `if not no_cache`** — both the load and the dump —
   so the env vars are never even consulted for explicit compiles:
   - 4.5.0: `base_dsl/dsl.py:1645` (load), `:1736` (dump)
[...]

PR #4331, opened 2026-08-04
perf(moe): persist b12x MoE CuTe-DSL kernels to the disk cache

Smallfu666:
## 📌 Description

Rolls the on-disk CuTe-DSL kernel cache out to the three sm12x NVFP4 fused-MoE kernels (micro,
static, dynamic), addressing #4317. The infrastructure landed in #3874 / #4029; @bkryu noted in
#4317 that it still had to reach individual kernels and that the core team lacked bandwidth, so this
applies it to b12x MoE.

Follows the rollout note in `docs/design_docs/cute_dsl_kernel_cache.md`: each `cute.compile` becomes
a closure passed to `build_and_load_cute_dsl_kernel`. The three cache-key tuples become named
functions so both cache levels derive from one source of truth, and the artifact name appends a
digest of that tuple — the keys hold floats and `None` (`swiglu_*`), and `1.5` / `-1.5` both sanitize
to `1_5`, so a formatted name would not be injective.

As in the existing adopters, the kernels now compile against
`make_fake_stream(use_tvm_ffi_env_stream=True)`, so TVM-FFI supplies the caller's current stream and
the two launch sites no longer pass one (compiled signatures: 24 / 24 / 32 parameters).

**Not covered:** the direct-micro kernel (`compile_direct_micro_kernel`), which this module started
dispatching to recently. It compiles without `--enable-tvm-ffi` and launches through the DSL rather
than a TVM-FFI callable, so caching it is a separate change — #4317 is only partly closed by this PR.

### Measured — GB10 (sm_121), DSL 4.6.0, five kernel shapes per process

| process | total compile + load |
|---|---:|
| before, two runs | 17.92 s / 17.95 s |
| after, cold first run | 18.91 s |
| after, two later runs | **0.130 s / 0.134 s** (~135×) |

Warm, per kernel: dynamic 7.7 s → 1 ms; static 3.2 s → 1 ms; micro 2.3–3.2 s → 1 ms (the first warm
kernel pays 0.13 s of one-off module setup). The cold run costs a few percent for the export.
Artifacts are 161–262 KB each.

## 🔍 Related Issues

#4317 (partly — see the scope note above). Builds on the cache infrastructure from #3874 and #4029.

### ✅ Pre-commit Checks

- [x] `pre-commit` installed.
- [x] Hooks installed.
- [x] `pre-commit run` on the two changed files: all hooks pass, no files modified. (I ran it
      file-scoped rather than `--all-files`, since the remaining hooks are repo-wide.)

## 🧪 Tests

- [x] Tests added.
- [x] The new tests pass; see the caveat below for what does not run on my hardware.

- New `tests/moe/test_b12x_moe_kernel_cache.py`: 61 naming-contract tests — signature coverage,
  per-argument perturbation, symbol safety, cross-family collision — replicating
  `tests/jit/test_cute_dsl_cache.py` as the design doc asks of new adopters (happy to fold them into
  that file instead if you would rather keep all adopters together).
- Also checked on the same host: a corrupt artifact and a read-only cache directory each fall back to
  compiling with a warning, and editing a key source invalidates the module and recompiles it once.

**Not verified — please check on sm120 hardware.** `tests/moe/test_b12x_fused_moe.py`'s numerical
tests do not run on my GB10: the nvcc reference ops fail to build (`CUDA compiler and CUDA toolkit
headers are incompatible`). The suite gives an identical `142 failed, 15 passed` — the same 142 test
ids — on unmodified `main` and on this branch, so I have no accuracy signal either way. All timings
above are compile/load time; steady-state kernel performance should be unchanged, since it is the
same binary.

## Summary by CodeRabbit

* **New Features**
  * Added shared on-disk caching for static, micro, and dynamic MoE kernels.
  * Improved cache invalidation when source dependencies change.
  * Kernel launches now automatically use the caller’s active CUDA stream.

* **Tests**
  * Added comprehensive coverage for cache-key completeness, naming stability, symbol safety, and uniqueness across kernel types.


yashb98:
You asked for someone with sm12x hardware to run the numerical suite. Results below, plus two things worth knowing.

GB10 (DGX Spark, sm_121, aarch64), nvcc 13.0.88, torch 2.11.0+cu130, cutlass-dsl 4.5.2. Base is the `nightly-v0.6.17-20260804` wheel, whose `moe_dispatch.py` is byte-identical to your base `d7e390c1`; the PR arm is that same wheel with your head `96a8e813` overlaid. Both arms run as `PYTHONPATH=<unpacked wheel dir> python -m pytest` from a directory containing only `tests/`, so `flashinfer` resolves to the wheel (0.6.17.dev20260804) and not to a source checkout.

**`tests/moe/test_b12x_fused_moe.py`: 157 passed on base, 157 passed on this PR.** Same 157 tests you collect, no accuracy change.

```
base d7e390c1  157 passed in 1251.07s
PR   96a8e813  157 passed in 1114.84s
```

That wall-clock difference is not a cache win, base ran first with a cold nvcc cache and the PR second with it warm. Controlled timing below.

**Controlled timing**, five `test_numerical_accuracy` shapes per process, varying only the cute-dsl cache state with the nvcc cache warm throughout:

| condition | pytest time |
|---|---|
| base, no cache code | 61.10 s / 60.52 s |
| PR, `FLASHINFER_CUTE_DSL_DISABLE_CACHE=1` | 59.70 s / 59.29 s |
| PR, cache cold | 60.70 s |
| PR, cache warm | **1.79 s / 1.85 s** |

About 33x on this workload rather than your 135x, since mine includes imports and test execution while yours is compile+load only. Disabling the cache costs nothing measurable, if anything the disabled arm is ~2% under base, which is drift rather than a win.

I cannot give you a verdict on the cold-run penalty either way. My two no-cache arms differ from each other by 1.3 s, larger than the ~1 s export cost you measured, so on a 60 s workload it sits under my noise floor, and my arms ran in blocks so arm and time-order are confounded. For what it is worth the cold run sits +1.2 s above the cache-disabled arm, the same absolute size as your 17.92 s -> 18.91 s, so nothing here contradicts it. If you want it pinned down on sm12x the right measurement is to time the `build_and_load_cute_dsl_kernel` calls directly rather than pytest wall time. Happy to do that.

Note our DSL versions differ (4.5.2 here vs 4.6.0 on yours) and `meta.json` pins `cute_dsl_version`, so artifacts do not cross DSL versions. The before/after delta is internally consistent on one box; the absolute numbers are not comparable to yours.

**Both the export and the reload path are covered.** The PR run compiled and exported 68 distinct specializations into `b12x_moe_sm121a_cute_dsl/` (29 `static_*`, 25 `dynamic_*`, 14 `micro_*`, per `flashinfer_jit.log`), so all three converted sites are reached, with `meta.json` = `{"arch": "sm121a", "cute_dsl_version": "...==4.5.2", "source_sha256": ...}`. Those were all first-time misses, and on a miss `build_and_load()` returns the kernel it just compiled rather than reloading the `.o`, so that alone would only exercise the export path. So I ran the full suite twice back to back:

```
run A (populating)  157 passed in 1068.54s   63 new compiles -> 68 artifacts
run B (warm)        157 passed in  281.65s    0 new compiles
```

157 passed against a fully warm cache with zero recompilation, 3.8x faster end to end. The reload path is covered at full suite breadth, not just the five shapes above.

**On the build failure.** `CUDA compiler and CUDA toolkit headers are incompatible, please check your include paths` is not a torch check and is not sm12x-specific. It is a `#error` in CCCL's `cuda/std/__cccl/cuda_toolkit.h`, which FlashInfer vendors and puts ahead of the CTK copy with `-I` (`flashinfer/jit/cpp_ext.py`). It fires when the major.minor of the nvcc actually invoked differs from `CUDART_VERSION` in the `cuda_runtime.h` its include path resolves to, so the comparison that matters is nvcc against the headers it is picking up, not against `torch.version.cuda`. `FLASHINFER_JIT_VERBOSE=1` prints the failing command line with its include dirs. I cannot say what it is on your box, but I never reached nvcc at all, which may be the whole difference: with `flashinfer-jit-cache` installed and `torch.version.cuda` 13.0, the sm_121 reference `fp4_quantize` path resolves to the `120f` family module, which that wheel ships prebuilt. No nvcc module directory appears under `cached_ops/` in either of my arms.

**On the direct-micro kernel you scoped out**, so you can size the follow-up. Three consecutive processes of `TestMicroKernel::test_direct_micro_forced_accuracy` (10 shapes) against a warm cache give 136.97 s / 134.10 s / 135.69 s, and the artifact count never moves, so there is no cross-process reuse at all. That is ~13.5 s per shape against the ~11.8 s per shape the cached path saves here, so per kernel it is about the same size as what this PR already fixes, just paid in full in every process. It matches your scope note: the call is not wrapped in `build_and_load_cute_dsl_kernel` at all (`compile_direct_micro_kernel`, `moe_direct_micro_kernel.py:5037`, defaults `options=None` and only forwards it when truthy; its only caller, `moe_dispatch.py:1367` on your head, passes none), so it needs the wrapper as well as `--enable-tvm-ffi`. Happy to move this to its own issue rather than clutter the PR.

Minor: CodeRabbit's docstring point is right, the module docstring lists three contracts while the file enforces five, with `test_disk_name_is_stable_for_equal_keys` and `test_kernel_types_do_not_collide` undocumented.

Happy to re-run any of this, or the full suite on a rebase.

flashinfer/fused_moe/cute_dsl/blackwell_sm12x/moe_dispatch.py
```diff
@@ -593,7 +594,201 @@ def _get_weight_views(
...
+    The name is the *sole* per-kernel cache key — the module ``meta.json``
+    guards only module-wide facts (arch, DSL stack, source hashes) — so it has
+    to be injective in every codegen parameter. It is therefore derived from
+    the very tuple that keys the in-process cache: a readable shape prefix for
+    humans browsing ``cached_ops/``, plus a digest of the exact tuple.
+
+    The digest, rather than a formatted field list, is what makes the mapping
+    injective: the keys contain floats and ``None`` (``swiglu_alpha`` /
+    ``swiglu_beta`` / ``swiglu_limit``) whose textual forms would collide once
+    sanitized into a filename (``1.5`` and ``-1.5`` both sanitize to ``1_5``).
```
bkryu, line 652: Appreciate the explanation that it is not just for convenience, but also for correctness.

bkryu (commented):
Thank you @Smallfu666 for the contribution. Looks good to me and I can reproduce the cache behavior on DGX Spark; I see ~10 sec to compile a kernel every time to 0.12s per kernel after first compile. 

Can approve after the CI goes through

Smallfu666:
Thanks @bkryu. All unit-test jobs passed; the only failures are the two unclassified B300 multi-GPU jobs. Could you rerun the pipeline?

bkryu (approved):

bkryu:
> Thanks @bkryu. All unit-test jobs passed; the only failures are the two unclassified B300 multi-GPU jobs. Could you rerun the pipeline?

Thanks @Smallfu666, the failures are not related so there won't be a need to retrigger. We are currently having an issue with a failunre in the CI for the `PR Test / AOT Build Import (x64, cu126) (pull_request)` job. We are working on a fix and I expect the PR to be merged afterwards.

Smallfu666:
Sorry, closed by accident on mobile — reopened.

PR #4346, opened 2026-08-04
fix: optimize and correct SM120 NVFP4 paged prefill

### LINKED ISSUE #4269: [Performance] SM120 NVFP4 paged causal prefill is ~1.7-1.8x slower than FP8
## Summary

On SM120, FlashInfer's paged causal prefill with BF16 queries and an NVFP4 KV cache is about **1.7-1.8x slower than the matched FP8 KV-cache path** in our vLLM workload.

NVFP4 decode is faster than FP8 as expected for a bandwidth-bound workload, so the regression appears localized to the compute-heavy causal prefill path.

## Environment

- GPU: NVIDIA RTX PRO 5000 Blackwell (SM120)
- CUDA: 12.9
- PyTorch: 2.10.0+cu129
- FlashInfer: 0.6.15
- KV layout: HND
- Query dtype: BF16 at the model boundary (quantized as required by the selected path)
- KV formats compared:
  - NVFP4 packed data with per-16-element FP8 scale factors
  - FP8 KV cache
- Integration: vLLM FlashInfer backend

## Matched profile results

The two traces contain some disjoint startup/warmup regions, so the table below compares the same 48 causal paged-prefill calls.

| KV cache | Calls | Total | Mean per call |
|---|---:|---:|---:|
| NVFP4 | 48 | 145.81 ms | 3.04 ms |
| FP8 | 48 | 84.03 ms | 1.75 ms |

**NVFP4 / FP8 prefill ratio: 1.74x**

For context, in the same workload:

| Operation | NVFP4 | FP8 | Observation |
|---|---:|---:|---|
| Decode | 9.63 us | 12.35 us | NVFP4 is about 22% faster |
| KV append | 8.39 us | 3.01 us | NVFP4 is slower, but only about 7 ms total |
| Matched CUTLASS segment | 93.61 ms | 96.69 ms | Similar |
| All matched kernels | 297.73 ms | 238.45 ms | Prefill explains most of the gap |

The end-to-end capacity test also shows the tradeoff: NVFP4 avoids preemption and admits more concurrent requests, but wall time is worse because prefill is slower.

## Current hypothesis

The current SM120 FA2 NVFP4 prefill path appears to consume packed FP4 KV through an unpack/convert/scale sequence before the BF16-oriented QK MMA. In causal prefill, QK is compute-heavy and the reduced KV memory traffic does not offset that extra conversion work.

This is a hypothesis based on the kernel path and profile behavior, not yet a complete instruction-level Nsight Compute proof.

An A4Q-style path could avoid this cost by quantizing Q and using native block-scaled FP4 QK MMA (`use_nvf4_qk`-style dispatch), while retaining the packed NVFP4 KV representation.

## Questions

1. Is this performance behavior expected for the current SM120 NVFP4 paged-prefill implementation?
2. Is an A4Q/native block-scaled FP4 QK path planned for upstream FlashInfer?
3. Would maintainers accept a contribution wiring Q quantization and block-scaled FP4 MMA into the paged causal prefill path?
4. Are there existing benchmarks or recommended kernel parameters for comparing NVFP4 and FP8 prefill fairly on SM120?

## Related work

- vLLM tracking issue with the serving-level measurements: https://github.com/vllm-project/vllm/issues/50416
- NVFP4 KV support for prefill/batch attention: https://github.com/flashinfer-ai/flashinfer/pull/3097
- SM120/121 NVFP4 paged-prefill extensions: https://github.com/flashinfer-ai/flashinfer/pull/3684

I can provide the Nsight Systems JSON traces and a reduced benchmark case if that would help with kernel-level investigation.

alphapibeta:
## 📌 Description

This PR fixes the SM120 NVFP4 paged causal prefill regression tracked in #4269.

The paged-prefill path dequantizes packed NVFP4 KV values inline for BF16-oriented MMA. This change adds an amortized NVFP4-to-BF16 repack path following the existing FP8 repack design, so conversion and scale application happen once per tile instead of inside every MMA iteration.

The initial repack implementation reinterpreted BF16 bit patterns as FP16 values when DTypeQ was half. The PTX conversion produces BF16 bits, so the implementation now explicitly converts BF16 to DTypeQ before applying the scale.

## 🔍 Related Issues

Fixes #4269

### ✅ Pre-commit Checks

- [x] Pre-commit hooks have been installed in the CUDA 13 development environment.
- [x] Hooks have been installed.
- [x] `pre-commit run --all-files` has been run successfully.

## 🧪 Tests

- [x] Final previously failing FP16 NVFP4 case passes on CT102.
- [x] Full paged NVFP4 selection passes: 39/39.
- [x] Representative paged-prefill regression passes.
- [x] `git diff --check` passes.

### Changes

- Add NVFP4 repack dispatch for paged and ragged prefill K/V paths.
- Convert only the eight valid packed FP4 bytes loaded by `load_64b_async`.
- Apply the per-block scale factor during repack.
- Convert BF16 results correctly to FP16 or retain BF16 according to DTypeQ.
- Preserve existing FP8 and non-FP4 paths.
- Add a deterministic NVFP4 repack boundary test.
- Add the CUDA 13 `uint_fastdiv` compatibility fallback.

### Validation environment

- CT102
- CUDA 13.0.88
- 2x NVIDIA RTX PRO 4000 Blackwell
- SM120

### Performance

Matched CUDA-event benchmark after warmup, 60 iterations per case, on CT102:

- Workload: BF16 Q, HND, causal paged prefill; batch=4, KV length=2048, QO length=128, page size=16, Q heads=32, KV heads=8, head dimension=128.
- Parent baseline (commit `51920591`, with only the independent CUDA 13 `fastdiv.cuh` compatibility fix applied): median **0.2999 ms**.
- This PR: median **0.2888 ms**.
- Result: **3.7% lower kernel-run latency** for this workload.

A shorter 256-token KV case was within noise, so the result above is reported as the representative long-KV case rather than extrapolating a larger gain. No large serving benchmark was run while SGLang owned the GPUs; SGLang was paused only for this controlled benchmark and restarted afterward. The PR is open for review pending maintainer review and broader CI validation.

## Review follow-up

- Account for NVFP4 repack staging and scale-factor shared memory in all launch-sizing paths.
- Restore CUDA 13 uint_fastdiv zero-divisor/default-constructor parity with the CUDA 12 implementation.
- Add a CUDA test covering default, zero, and nonzero divisors.
- Revalidated on SM120: uint_fastdiv parity passed; NVFP4 boundary and previously failing FP16 case passed.
- pre-commit run --all-files passed.

## Summary by CodeRabbit

* **New Features**
  * Added support for NVFP4 key/value cache repacking in BF16 and FP16 prefill workflows, including single, ragged, and paged attention.
  * Added compatibility for CUDA 13 with built-in fast division and modulo operations.

* **Bug Fixes**
  * Preserved existing FP8 processing while improving FP4 data handling across supported prefill paths.

* **Tests**
  * Added coverage for NVFP4 paged-prefill boundary scenarios and CUDA fast division behavior.

### DIFF
--- a/include/flashinfer/attention/prefill.cuh
+++ b/include/flashinfer/attention/prefill.cuh
@@ -2626,7 +2626,7 @@ cudaError_t SinglePrefillWithKVCacheDispatched(Params params, typename Params::D
     // path is active, so NUM_MMA_KV is chosen to keep base+staging within the
     // occupancy budget. NOTE: single-prefill doesn't use the repack, but the
     // staging buffer still lives in SharedStorageQKVO, so it must be accounted.
-    constexpr bool kUseRepack = (sizeof(DTypeKV) == 1) && !is_fp4_type_v<DTypeKV> &&
+    constexpr bool kUseRepack = ((sizeof(DTypeKV) == 1) || is_fp4_type_v<DTypeKV>) &&
                                 (HEAD_DIM_VO != 64) && (HEAD_DIM_VO <= 256) && (CTA_TILE_Q > 16);
     constexpr bool kKVShared = !is_fp4_type_v<DTypeKV> && (HEAD_DIM_VO / 16 > 16) &&
                                ((HEAD_DIM_VO / 16) % NUM_WARPS_KV == 0) &&
@@ -4258,7 +4258,7 @@ cudaError_t BatchPrefillWithRaggedKVCacheDispatched(Params params, typename Para
   // staging buffer (sized max(HEAD_DIM_QK, HEAD_DIM_VO)) when the FP8 repack path
   // is active, so NUM_MMA_KV is chosen to keep base+staging within the occupancy
   // budget (otherwise the staging silently drops blocks/SM at large head dims).
-  constexpr bool kUseRepack = (sizeof(DTypeKV) == 1) && !is_fp4_type_v<DTypeKV> &&
+  constexpr bool kUseRepack = ((sizeof(DTypeKV) == 1) || is_fp4_type_v<DTypeKV>) &&
                               (HEAD_DIM_VO != 64) && (HEAD_DIM_VO <= 256) && (CTA_TILE_Q > 16);
   // Matches KernelTraits::USE_KV_SHARED_SMEM: at large head dims K and V
   // time-share one smem buffer, so the occupancy budget counts the K/V
@@ -4276,6 +4276,9 @@ cudaError_t BatchPrefillWithRaggedKVCacheDispatched(Params params, typename Para
       (kUseRepack ? ((HEAD_DIM_QK > HEAD_DIM_VO ? HEAD_DIM_QK : HEAD_DIM_VO) * 16 * NUM_WARPS_KV *
                      sizeof(DTypeQ))
                   : 0u) +
+      (is_fp4_type_v<DTypeKV>
+           ? ((HEAD_DIM_QK + HEAD_DIM_VO) * 16 * NUM_WARPS_KV / NVFP4_SF_VEC_SIZE)
+           : 0u) +
       (kVOSplitDispatch ? (CTA_TILE_Q * NUM_WARPS_KV * 16 * sizeof(DTypeQ)) : 0u);
   constexpr uint32_t kVOSplitFixedSmem =
       kVOSplitDispatch ? (NUM_WARPS_KV * CTA_TILE_Q * 8u + 2048u) : 0u;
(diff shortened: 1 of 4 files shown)


seanyourhighness:
Independent SM120/NVFP4 validation for this PR against FlashInfer
0.6.16.post3 (`9dc1b24`):

- backport applied cleanly and all 57 NVFP4 prefill correctness tests passed,
  including paged/ragged, head_dim 128/256, RoPE, strided scales, and the new
  repack-boundary case;
- at the DFlash q_len=8 / head_dim=256 / 180K-context geometry, the patched
  kernel was about 10-12% faster at c3-c8 (2.63 vs 2.93 ms/call at c8), with
  negligible c1/c2 change;
- matched server A/B at 90K showed 92.6/116.0/186.3 vs
  115.3/127.2/188.5 tok/s at c1/c3/c8, within observed boot/run variance;
- the integrated c3 slowdown persisted in both arms and was later localized to
  XQA execution context, not this prefill repack path.

Conclusion: safe and useful as a modest NVFP4 prefill optimization, but not the
server XQA speed fix.

Disclosure: AI-assisted analysis and write-up; I reviewed the patch and
verified the reported runs.

PR #4349, opened 2026-08-04
fix(rope): guard rotary_dim <= head_dim to prevent silent cross-head corruption

SuperMarioYL:
## 📌 Description

Adds a `head_dim < rotary_dim` guard to the 5 RoPE launchers that were missing it, mirroring the guard already present in `BatchQKApplyRotaryPosIdsCosSinCache` (`include/flashinfer/pos_enc.cuh:1301`).

**Why this matters:** the RoPE kernel `vec_apply_llama_rope` (`pos_enc.cuh:107-126`) computes `x + rotary_dim/2` with no bounds check. When `rotary_dim > head_dim` (a misconfiguration that is easy to hit when porting partial-RoPE configs between HF and vLLM — e.g. `rotary_dim=192` on a `head_dim=128` tensor), the offset reads past the current head's data into the adjacent head/token. The result is **silent cross-head corruption**: no crash, no NaN — just wrong attention output downstream, which is far harder to diagnose than a clear error.

Only 1 of the 6 RoPE launchers (`BatchQKApplyRotaryPosIdsCosSinCache`) currently guards this. This PR propagates that exact guard to the 5 siblings:

- `BatchQKApplyRotaryPosIds`
- `BatchQKApplyRotary`
- `BatchQKApplyRotaryInPlace`
- `BatchQKApplyLlama31Rotary`
- `BatchQKApplyLlama31RotaryPosIds`

The C++ guard is mirrored byte-for-byte from the existing one (`FLASHINFER_ERROR` + `std::ostringstream`, same message wording), placed at the top of each launcher body before any kernel dispatch. A matching host-side `ValueError` is also added to the 4 public Python wrappers in `flashinfer/rope.py` so Python callers get a clear error before any JIT/kernel launch.

## 🔍 Related Issues

Refs #4312 (the prefill `head_dim` validation PR) — same class of "validate head_dim/rotary_dim at the launch boundary" fix; this covers the RoPE side that #4312 does not touch. No filed issue exists for the RoPE `rotary_dim` gap specifically — the corruption is silent, so it tends to be attributed to model weights/config rather than the kernel.

### ✅ Pre-commit Checks

- [x] `ruff check` passes on the changed Python files (`flashinfer/rope.py`, `tests/test_rope_rotary_dim_bound.py`); `pos_enc.cuh` is C++ (out of ruff scope).
- [ ] Full `pre-commit run --all-files` (mypy / codespell / clang-format) deferred to CI — no local toolchain for the CUDA/mypy parts.

> Note: I am not on the `ci-users` team, so public CI will not auto-trigger — requesting `run-ci` (or `@flashinfer-bot run`) for the GPU test matrix.

## 🧪 Tests

- [x] Tests added: `tests/test_rope_rotary_dim_bound.py` — parametrized over all 4 public Python wrappers; the bad-path (`rotary_dim` 128/192 with `head_dim=64`) asserts `ValueError` (runs without a GPU — the Python guard fires pre-kernel); the happy-path (`rotary_dim == head_dim == 64`) is `skipif(not torch.cuda.is_available())` since it drives the kernel.
- [x] `python3 -m py_compile` passes on the changed files.

## Reviewer Notes

- The guard is mechanical and byte-identical to the existing `BatchQKApplyRotaryPosIdsCosSinCache` guard (`pos_enc.cuh:1301`), so review should be short — it is "propagate the guard the team already wrote to the 5 launchers that missed it."
- `BatchQKApplyRotaryInPlace` delegates to `BatchQKApplyRotary`, so the guard on the latter already covers the in-place path; I added it to the in-place entry too as defense-in-depth at the public boundary. Happy to drop that one if you would prefer a single source of truth.
- The C++ guards are a backstop for direct TVM-FFI / JAX callers; the Python guards cover the common Python-API path. Both layers mirror each other.
- Pre-empting the "this is just user error" concern: agreed that `rotary_dim > head_dim` is a caller misconfiguration, but the cost of a host-side bounds check (one comparison per launch) is negligible, and the alternative — silently wrong attention output attributed to model weights — is materially worse for users debugging partial-RoPE setups. The existing `BatchQKApplyRotaryPosIdsCosSinCache` guard (line 1301) already encodes this judgment; this PR just makes it consistent across the RoPE surface.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to rotary position embedding operations to reject configurations where `rotary_dim` exceeds the query or key head dimension.
  * Errors are now reported before processing begins, preventing invalid execution and improving clarity.
  * Valid configurations where `rotary_dim` equals the head dimension continue to work as expected.

* **Tests**
  * Added regression coverage for invalid, boundary, and mismatched query/key head-dimension configurations.

### DIFF
--- a/flashinfer/rope.py
+++ b/flashinfer/rope.py
@@ -512,6 +512,10 @@ def apply_rope_inplace(
     """
     if rotary_dim is None:
         rotary_dim = q.size(-1)
+    if rotary_dim > q.size(-1):
+        raise ValueError(
+            f"head_dim ({q.size(-1)}) must be >= rotary_dim ({rotary_dim})"
+        )
     _apply_rope(
         q, k, q, k, indptr, offsets, rotary_dim, interleave, rope_scale, rope_theta
     )
@@ -571,6 +575,10 @@ def apply_rope_pos_ids_inplace(
     """
     if rotary_dim is None:
         rotary_dim = q.size(-1)
+    if rotary_dim > q.size(-1):
+        raise ValueError(
+            f"head_dim ({q.size(-1)}) must be >= rotary_dim ({rotary_dim})"
+        )
     _apply_rope_pos_ids(
         q, k, q, k, pos_ids, rotary_dim, interleave, rope_scale, rope_theta
     )
@@ -668,6 +676,10 @@ def apply_llama31_rope_inplace(
     """
     if rotary_dim is None:
         rotary_dim = q.size(-1)
+    if rotary_dim > q.size(-1):
+        raise ValueError(
+            f"head_dim ({q.size(-1)}) must be >= rotary_dim ({rotary_dim})"
+        )
     _apply_llama31_rope(
         q,
         k,
@@ -748,6 +760,10 @@ def apply_llama31_rope_pos_ids_inplace(
     """
     if rotary_dim is None:
         rotary_dim = q.size(-1)
+    if rotary_dim > q.size(-1):
+        raise ValueError(
+            f"head_dim ({q.size(-1)}) must be >= rotary_dim ({rotary_dim})"
+        )
     _apply_llama31_rope_pos_ids(
         q,
         k,
--- a/flashinfer/rope.py
+++ b/flashinfer/rope.py
@@ -512,9 +512,9 @@ def apply_rope_inplace(
     """
     if rotary_dim is None:
         rotary_dim = q.size(-1)
-    if rotary_dim > q.size(-1):
+    if rotary_dim > min(q.size(-1), k.size(-1)):
         raise ValueError(
-            f"head_dim ({q.size(-1)}) must be >= rotary_dim ({rotary_dim})"
+            f"head_dim (q={q.size(-1)}, k={k.size(-1)}) must be >= rotary_dim ({rotary_dim})"
         )
     _apply_rope(
         q, k, q, k, indptr, offsets, rotary_dim, interleave, rope_scale, rope_theta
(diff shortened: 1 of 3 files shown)


lazypool (commented):
@SuperMarioYL Thanks, the `rotary_dim <= head_dim` guard is the right direction and matches the existing cos/sin-cache launcher.

One safety gap remains: the native wrappers derive `head_dim` from `q.size(2)`, but the same `rotary_dim` is also applied to `k` and there is no validation that `k.size(2)` matches Q or is at least `rotary_dim`.

For example, `q.shape[-1] == 128`, `k.shape[-1] == 64`, and `rotary_dim == 128` passes the new guard, but the K path can still read beyond one K head. Could we validate that Q and K have compatible head dimensions, or at minimum require `rotary_dim <= min(q.size(-1), k.size(-1))`, in both the Python and C++ entry points?

Please also add a regression case for mismatched Q/K head dimensions. This is especially important here because the stated purpose is preventing silent cross-head corruption.

SuperMarioYL:
Thanks @lazypool — good catch on the K path. Pushed `8d25bcce`.

**Python wrappers:** all four public wrappers now bound `rotary_dim <= min(q.size(-1), k.size(-1))` (was q-only), and the error names both head dims (`head_dim (q=..., k=...) must be >= rotary_dim (...)`). Your exact scenario (q=128, k=64, rotary_dim=128) now raises at the Python layer before any JIT/kernel work. Added a CPU (pre-kernel) regression test parametrized across the four wrappers for mismatched Q/K.

**C++ entry points:** the TVM bindings already enforce `TVM_FFI_ICHECK_EQ(q.size(2), k.size(2))` at the top of every `apply_rope*` in `csrc/rope.cu`, which is stronger than the min check — with equal head dims the launcher's existing `head_dim < rotary_dim` guard in `pos_enc.cuh` already covers K. So mismatched Q/K was already rejected at the C++ binding before the kernel; this change makes it fail earlier and with a clearer rotary-specific message rather than the q!=k equality error.

If you'd still like an explicit `min(q_head_dim, k_head_dim)` guard inside the launchers (`pos_enc.cuh`) for defense-in-depth, that's a small signature change (thread `k_head_dim` through the launchers + the `csrc/rope.cu` call sites) — I held off because it's redundant given the binding's q==k check and I can't build CUDA locally to verify it; happy to add it if you want the belt-and-suspenders.

PR #4356, opened 2026-08-05
fix(prefill): size split-KV tmp_v by output dtype

### LINKED ISSUE #4333: Question: Should split-KV batch_prefill_tmp_v workspace use sizeof_dtype_o instead of sizeof(float)?
Hi, I noticed a possible inconsistency in the FA2 batch prefill planner around the split-KV temporary output workspace.

In `PrefillPlanImpl`, the function accepts a `sizeof_dtype_o` argument:

```cpp
bool enable_cuda_graph, uint32_t sizeof_dtype_o, ...
```

but the argument is currently unused:

```cpp
(void)sizeof_dtype_o;
```

When split-KV is enabled, the temporary output buffer `batch_prefill_tmp_v` is allocated with `sizeof(float)`:

```cpp
plan_info.v_offset = float_allocator.aligned_alloc_offset(
    num_qo_heads * padded_batch_size * cta_tile_q * head_dim_vo * sizeof(float),
    16,
    "batch_prefill_tmp_v");
```

However, at run time this buffer is interpreted as `DTypeO*`:

```cpp
tmp_v = GetPtrFromBaseOffset<DTypeO>(float_buffer_ptr, plan_info.v_offset);
```

and the prefill kernel writes partial outputs through `params.o = tmp_v`, where `params.o` is also `DTypeO*`.

So I wanted to ask whether the `sizeof(float)` allocation is intentional here.

My understanding is that `tmp_s` / LSE should be float, but `tmp_v` represents partial output values and should probably be sized according to `DTypeO`, i.e. using `sizeof_dtype_o`. With the current code, fp16/bf16 output appears to over-allocate the `tmp_v` workspace by 2x. This is probably safe from a correctness perspective for current fp16/bf16 paths, but it makes the workspace size larger than necessary and seems inconsistent with the existing `sizeof_dtype_o` parameter.

Would it make sense to change this allocation to:

```cpp
num_qo_heads * padded_batch_size * cta_tile_q * head_dim_vo * sizeof_dtype_o
```

and pass `sizeof(DTypeO)` from `csrc/batch_prefill.cu` instead of the current hardcoded `/*sizeof_dtype_o=*/2`?

Thanks!

lazypool:
## 📌 Description

Fixes an FA2 split-KV workspace sizing inconsistency reported in #4333.

The prefill kernel stores normalized per-split partial outputs in `tmp_v` as `DTypeO`, but the planner previously reserved FP32-sized storage for that buffer. This PR passes `sizeof(DTypeO)` to the FA2 planner and uses it for `tmp_v`; `tmp_s` remains FP32 because it stores LSE values.

This does not alter attention math or output precision. It corrects the staging buffer allocation to match its existing `DTypeO` read/write contract. This PR affects only workspace sizing; the split-KV kernel and merge algorithm are unchanged.

## 🔍 Related Issues

Fixes #4333

## 🧪 Tests

- [x] Tests have been updated as needed and passes on RTX 4090.
- [x] `tests/attention/test_workspace_size.py::test_batch_prefill_workspace_size_plans_fixed_split_with_exact_buffers`

### Workspace Measurement

The before/after values were collected from the FA2 planner rather than from an allocated PyTorch tensor.

A paged-KV input was constructed with:

```text
batch_size=3, qo_len=64, kv_len=16384, page_size=16
num_qo_heads=16, num_kv_heads=4, head_dim=128
```

This produces a split-KV plan with 384 partitions. For each version, `workspace_size()` was called, exact-size byte workspaces were supplied to `plan()`, and the generated `PrefillPlanInfo` was inspected:

```text
tmp_v bytes = s_offset - v_offset
tmp_s bytes = float_workspace_size - s_offset
```

Both plans selected `padded_batch_size=384` and `cta_tile_q=128`.

```text
tmp_v = 16 heads * 384 partitions * 128 Q tile * 128 head_dim * element_size
```

| Allocation | Before | After |
|---|---:|---:|
| `tmp_v` | 384 MiB | 192 MiB |
| `tmp_s` | 3 MiB | 3 MiB |
| Total float workspace | 387 MiB | 195 MiB |

The old plan uses `sizeof(float)` for `tmp_v`. The fixed plan uses `sizeof(DTypeO)`, which is 2 bytes for this FP16 configuration. Planning succeeds with the reduced exact-size workspace.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` using the local `torch-base3.10` pyenv environment.
- [x] I have run the relevant hooks manually and fixed reported issues.

## Reviewer Notes

`tmp_v` is written as `DTypeO` by the split-KV prefill kernel and reloaded as the merge kernel input dtype. `tmp_s` remains intentionally allocated as `float` for FP32 LSE storage.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workspace allocation for prefill and decode operations across output data types of different sizes.
  * Fixed temporary buffer sizing to accurately reflect the selected output type.

* **Tests**
  * Added validation for split-workspace planning in batch decode and fixed-split prefill scenarios.
  * Expanded coverage for workspace-size calculations based on model dimensions and output data size.


lazypool (commented):
Self-review: the planner now receives `sizeof(DTypeO)` from the generated FA2 module and uses it only for the `tmp_v` allocation.

Appenhaimer:
Same issue in DecodePlanImpl as well.

`plan_info.v_offset = float_allocator.aligned_alloc_offset(
        num_qo_heads * padded_batch_size * HEAD_DIM * sizeof(float), 16, "batch_decode_tmp_v");`

Appenhaimer:
> Same issue in DecodePlanImpl as well.
> 
> `plan_info.v_offset = float_allocator.aligned_alloc_offset( num_qo_heads * padded_batch_size * HEAD_DIM * sizeof(float), 16, "batch_decode_tmp_v");`

Let's fix it n this PR as well.

lazypool:
@Appenhaimer Good catch — I verified BatchDecodeWithPagedKVCacheDispatched writes tmp_v via cast_store as DTypeO (decode.cuh:599), so the decode planner has the same mismatch. I'll extend the fix to DecodePlanImpl in this PR (incl. the MLA plan paths that call DecodePlan) and add a decode workspace-size regression test.

XFDG:
Cross-linking #5107 and #5166 because these PRs choose mutually exclusive contracts for the same split-KV decode buffer. On B200/SM100 with BF16, the current output-typed partial path produced 2063/4096 incorrectly rounded values for split B=1, versus 10/4096 on the non-split B=256 path. Keeping FP32 partials in #5166 reduced split B=1 to 0/4096 and all tested split batches B=1..128 to 0-8/4096. Commit 6438f97a in #5166 now changes the exact scheduler allocation line to DTypePartialO = float, so the branches cannot be combined silently. If #4356 is preferred for memory savings, the extra BF16 rounding becomes an intentional contract; if #5166 is preferred for precision, decode tmp_v must remain FP32. Please choose one decode policy before merging either PR unchanged.

PR #4360, opened 2026-08-05
fix(packaging): pin the tvm-ffi build ABI and declare it in wheel metadata

YangXu1990uiuc:
> **This does not block v0.6.16.post2.** That tag builds correctly today: it carries the old open range `apache-tvm-ffi>=0.1,<0.2`, and the highest non-yanked release on PyPI is now `0.1.13.post2`, so dispatching the release workflow resolves the right version. Ship post2 first; this PR is the longer-run fix so that outcome stops depending on when the button is pressed.

## Problem

`apache-tvm-ffi` is a **header-only ABI dependency**. `make_object`, `Optional<T>` and the `ModuleObj` vtable are all inlined into every `.so` we compile, so **the version resolved at build time is the ABI those artifacts require at run time**.

Today nothing pins it and nothing records it:

| site | constraint before |
|---|---|
| `requirements.txt` (runtime) | `>=0.1.6,!=0.1.8,!=0.1.8.post0,<0.2` |
| `pyproject.toml` `[build-system]` | `>=0.1.6,!=0.1.8,!=0.1.8.post0,<0.2` |
| `flashinfer-jit-cache/pyproject.toml` `[build-system]` | `>=0.1,<0.2` |
| **`flashinfer-jit-cache/pyproject.toml` `[project]`** | **`dependencies = []`** |
| `flashinfer-cubin/pyproject.toml` `[build-system]` | `>=0.1,<0.2` |
| 3 × docker / build scripts | 2 more spellings |

`scripts/build_flashinfer_jit_cache_whl.sh` runs `python -m build --wheel`, which creates a throwaway PEP 517 venv and resolves that open range from PyPI on every run. The nvcc lines in CI show it directly:

```
rc5 (run 30517013748):  -isystem /tmp/build-env-kqj5xz09/lib/python3.12/site-packages/tvm_ffi/include
GA  (run 30608215046):  -isystem /tmp/build-env-sxxbjhji/lib/python3.12/site-packages/tvm_ffi/include
```

So `v0.6.16rc5` (wheel uploaded 07-30 07:13Z) and `v0.6.16` (07-31 08:02Z) came out of the same source tree with **different ABIs**, because tvm-ffi 0.1.13 landed on PyPI at 07-30 15:47Z in between. Verified on the shipped wheels:

```
flashinfer_jit_cache-0.6.16rc5      fp4_kv_quant.so → U TVMFFIGetCustomAllocator absent
flashinfer_jit_cache-0.6.16         fp4_kv_quant.so → U TVMFFIGetCustomAllocator PRESENT  ← undefined symbol on <=0.1.12
flashinfer_jit_cache-0.6.16.post1   fp4_kv_quant.so → absent again
```

The build log records none of this — 246k lines, and `python -m build` swallows the isolated-env resolution behind a single `* Creating isolated environment: venv+pip...`.

## Changes

**1. Build sites pin one exact version: `0.1.13.post2`.** The newest release validated against the runtime floor, so upstream ABI fixes land rather than being frozen out, and every artifact from a given source tree has the same ABI regardless of when it is built.

**2. `flashinfer-jit-cache` declares a runtime constraint.** This is the package that ships the 959 prebuilt `.so` files carrying the ABI, and it had `dependencies = []`. pip was free to pair it with any `apache-tvm-ffi`, so the mismatch surfaced as a segfault at the first kernel call rather than an error at install time. This is the most load-bearing line in the diff.

Canonical strings, now in one place and CI-enforced:

```
build    apache-tvm-ffi==0.1.13.post2
runtime  apache-tvm-ffi>=0.1.11,!=0.1.13,!=0.1.13.post0,!=0.1.13.post1,<0.2
```

The floor (`0.1.11`) is deliberately **below** the build pin: it is what real deployments pin (upstream vLLM uses `apache-tvm-ffi==0.1.11`, which tilelang's `<0.1.13` cap also forces). That wider claim is empirical, not structural — a post2-built wheel runs on 0.1.11 only because the two are ABI-identical, which was measured rather than assumed. The test file records that a `BUILD_PIN` bump must be re-validated against the floor, with `FLASHINFER_DISABLE_JIT=1` so a failed `.so` load cannot be masked by a local rebuild.

## Why those three versions are excluded

tvm-ffi 0.1.13 shipped two independent ABI breaks. Measured with a compile probe mirroring our `FusedMoeRunner : public ModuleObj`, linked against each version's `libtvm_ffi.so`:

| tvm-ffi | `sizeof(Optional<Function>)` | `ModuleObj::GetFunction` return size |
|---|---|---|
| 0.1.10 / 0.1.11 / 0.1.12 | 8 | 8 |
| 0.1.13 | 16 | 16 |
| 0.1.13.post0 | 16 | 16 |
| 0.1.13.post1 | 16 | **8** (apache/tvm-ffi#700 patched only this method) |
| 0.1.13.post2 | **8** | **8** (apache/tvm-ffi#701 restored the general rule) |

`ModuleObj::GetFunction` is virtual and we override it in [`csrc/fused_moe/cutlass_backend/flashinfer_cutlass_fused_moe_binding.cu`](https://github.com/flashinfer-ai/flashinfer/blob/main/csrc/fused_moe/cutlass_backend/flashinfer_cutlass_fused_moe_binding.cu#L768) and [`csrc/fp8_blockscale_gemm_sm90_binding.cu`](https://github.com/flashinfer-ai/flashinfer/blob/main/csrc/fp8_blockscale_gemm_sm90_binding.cu#L57). C++ symbols and vtable slots don't encode the return type, so an 8-vs-16 disagreement corrupts the return slot — observed downstream as `Cannot find type info for type_index=1413787952` (`0x5444B130`, a pointer low-word read where a type index was expected; valid range is 64–127 static / ≥128 dynamic).

Everything else on the boundary is byte-identical across 0.1.10 → 0.1.13.post2 (`Any`, `AnyView`, `Object`, `ObjectRef`, `Function`, `Module`, `String`, `TVMFFIAny`, `TVMFFIObject`, `TVMFFITypeInfo`); exported symbols are additive-only with nothing removed; `kTVMFFIStaticObjectBegin`/`kTVMFFIDynObjectBegin` are unchanged and the new `kTVMFFIVisitInterrupt = 77` is appended without renumbering.

**The `!=` exclusions stay useful even after those releases are yanked upstream** — a PyPI yank is advisory (PEP 592), and an exact `==0.1.13.post0` in a downstream lockfile still selects it (that is exactly what the internal vLLM image pinned). With these exclusions that combination becomes a pip resolution error instead of a segfault.

Note the PEP 440 subtlety the range depends on: `!=0.1.13` excludes only `0.1.13` itself, **not** its post-releases, so `0.1.13.post2` still resolves. There is a regression test for that.

## Floor bump

`0.1.6` → `0.1.11`. `0.1.6`–`0.1.10` were advertised but never tested; 0.1.11 is the oldest version whose ABI we have actually measured, and the one upstream pins. The `!=0.1.8` exclusions are dropped since they now sit below the floor.

## Test

`tests/utils/test_dependency_constraints.py` holds the two canonical strings and asserts:
- every build site carries the exact pin,
- every runtime site carries the full range,
- the build pin is itself inside the runtime range (we can't compile against a version we then refuse to run on),
- the three ABI-divergent releases stay excluded and `0.1.13.post2` stays allowed,
- no tracked file names `apache-tvm-ffi` without being classified, so a new site can't be added silently.

CPU-only, no GPU, ~instant.

## Notes for review

- `flashinfer-cubin` ships only cubins (device code, no host ABI), so it gets no runtime constraint. Its build-time entry is kept and pinned in lockstep — it appears vestigial (nothing in that package references tvm-ffi) and could be dropped in a follow-up, but I left it in rather than risk the cubin release path.
- Deliberately **not** in this PR, but worth follow-ups: deriving the jit-cache runtime floor from the actually-resolved build version instead of a static literal; and an import-time guard for users who force-install an excluded version past pip.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility and stability by standardizing the `apache-tvm-ffi` version range across builds and runtime installations.
  * Excluded known incompatible releases to help prevent installation and ABI-related issues.
  * Added a controlled release-build version to improve consistency for packaged components.

* **Tests**
  * Added validation to ensure dependency constraints remain consistent across supported installation paths and release builds.
  * Added verification to detect ABI incompatibilities in packaged JIT-cache wheels before release.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,10 @@ nvep = []
 flashinfer = "flashinfer.__main__:cli"
 
 [build-system]
-requires = ["setuptools>=77", "packaging>=24", "apache-tvm-ffi>=0.1.6,!=0.1.8,!=0.1.8.post0,<0.2"]
+# apache-tvm-ffi must match requirements.txt: it is a header-only ABI dependency, so
+# the version resolved here is the ABI of anything compiled. Release builds narrow it
+# further via PIP_CONSTRAINT (scripts/build_constraints.txt).
+requires = ["setuptools>=77", "packaging>=24", "apache-tvm-ffi>=0.1.10,<0.2"]
 build-backend = "build_backend"
 backend-path = ["."]
 
--- a/docker/Dockerfile.flashinfer-nvep
+++ b/docker/Dockerfile.flashinfer-nvep
@@ -121,7 +121,7 @@ ENV VIRTUAL_ENV=${VENV}
 RUN uv pip install --python ${VENV}/bin/python \
         torch \
         setuptools packaging \
-        'apache-tvm-ffi>=0.1.6,<0.2,!=0.1.8,!=0.1.8.post0' \
+        'apache-tvm-ffi>=0.1.10,<0.2' \
         cython pybind11
 
 # moe_ep runtime base libraries — supplied by these pip wheels (not staged
--- a/docker/install/build_flashinfer_ep_pytorch.sh
+++ b/docker/install/build_flashinfer_ep_pytorch.sh
@@ -86,7 +86,7 @@ cd "${FI_SRC}"
 # without this upgrade.
 PIP_CONSTRAINT="" pip install --no-cache-dir -U \
     "setuptools>=77" "packaging>=24" \
-    "apache-tvm-ffi>=0.1.6,!=0.1.8,!=0.1.8.post0,<0.2"
+    "apache-tvm-ffi>=0.1.10,<0.2"
 PIP_CONSTRAINT="" BUILD_NIXL_EP=0 \
     pip install --no-cache-dir --no-build-isolation -e .
 
--- a/flashinfer-cubin/pyproject.toml
+++ b/flashinfer-cubin/pyproject.toml
@@ -1,5 +1,8 @@
 [build-system]
-requires = ["setuptools>=61.0", "wheel", "requests", "filelock", "torch", "tqdm", "numpy", "apache-tvm-ffi>=0.1,<0.2", "nvidia-ml-py"]
+# Cubins are device code with no host ABI surface, so this wheel has no runtime
+# apache-tvm-ffi requirement; the build-time entry only exists because the cubin
+# download path imports flashinfer.jit, which imports tvm_ffi.
+requires = ["setuptools>=61.0", "wheel", "requests", "filelock", "torch", "tqdm", "numpy", "apache-tvm-ffi>=0.1.10,<0.2", "nvidia-ml-py"]
 build-backend = "build_backend"
 backend-path = ["."]
 
--- a/flashinfer-jit-cache/pyproject.toml
+++ b/flashinfer-jit-cache/pyproject.toml
@@ -1,5 +1,8 @@
 [build-system]
-requires = ["setuptools>=77", "packaging>=24", "wheel", "tqdm", "torch", "ninja", "requests", "numpy", "nvidia-ml-py", "apache-tvm-ffi>=0.1,<0.2"]
+# Every .so in this wheel is compiled against apache-tvm-ffi's headers, so the version
+# resolved here is this package's ABI. Release builds pin it exactly via PIP_CONSTRAINT
+# (scripts/build_constraints.txt); the declared range is what consumers must satisfy.
+requires = ["setuptools>=77", "packaging>=24", "wheel", "tqdm", "torch", "ninja", "requests", "numpy", "nvidia-ml-py", "apache-tvm-ffi>=0.1.10,<0.2"]
 build-backend = "build_backend"
 backend-path = ["."]
 
@@ -28,7 +31,10 @@ classifiers = [
     "Topic :: Software Development :: Libraries :: Python Modules",
     "Topic :: Scientific/Engineering :: Artificial Intelligence",
 ]
-dependencies = []
+# This wheel ships prebuilt .so files that link libtvm_ffi, so it carries a real runtime
+# ABI requirement; an empty list let pip pair it with an incompatible apache-tvm-ffi and
+# fail as a segfault at the first kernel call rather than at install time.
+dependencies = ["apache-tvm-ffi>=0.1.10,<0.2"]
 
 [project.urls]
 Homepage = "https://github.com/flashinfer-ai/flashinfer"
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,10 @@
-apache-tvm-ffi>=0.1.6,!=0.1.8,!=0.1.8.post0,<0.2
+# Header-only ABI dependency: the version resolved at build time is the ABI our
+# compiled kernels need. 0.1.13/.post0/.post1 are ABI-divergent (object Optional<T>
+# became 16 bytes, changing the return storage of the virtual ModuleObj::GetFunction we
+# override) and are yanked upstream, so a resolver will not pick them; we rely on that
+# rather than carrying permanent exclusions. scripts/verify_jit_cache_abi.py checks a
+# built wheel actually loads on the floor declared here.
+apache-tvm-ffi>=0.1.10,<0.2
 click
 # cuda-python + nccl4py: runtime deps for the moe_ep EP transport backends
 # (default since the EP install became opt-out; CUDA-13 wheels only — see
--- a/scripts/build_constraints.txt
+++ b/scripts/build_constraints.txt
@@ -1,0 +1,12 @@
+# pip constraints for release builds of flashinfer-jit-cache.
+#
+# apache-tvm-ffi is a header-only ABI dependency, so the version resolved while
+# building freezes the ABI of every .so in the wheel. The package metadata keeps a
+# range (consumers must not be forced onto one release); this file pins the single
+# version our published wheels are built against, so the same source tree cannot
+# produce different ABIs on different days. Consumed via PIP_CONSTRAINT in
+# scripts/build_flashinfer_jit_cache_whl.sh.
+#
+# Must stay inside the range declared in requirements.txt — enforced by
+# tests/utils/test_dependency_constraints.py.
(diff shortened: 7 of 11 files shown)


pyproject.toml
```diff
@@ -40,7 +40,12 @@ nvep = []
 flashinfer = "flashinfer.__main__:cli"
 
 [build-system]
-requires = ["setuptools>=77", "packaging>=24", "apache-tvm-ffi>=0.1.6,!=0.1.8,!=0.1.8.post0,<0.2"]
+# apache-tvm-ffi is pinned exactly: it is a header-only ABI dependency, so the
+# resolved version freezes the ABI of everything we compile, and letting it float
+# meant the same source tree produced different ABIs on different days. The pin
+# tracks the newest release we have validated, which is not the floor of the
+# runtime range in requirements.txt — see tests/utils/test_dependency_constraints.py.
+requires = ["setuptools>=77", "packaging>=24", "apache-tvm-ffi==0.1.13.post2"]
```
aleozlx: no

aleozlx (changes requested):
== is not ok at FI level

YangXu1990uiuc:
@aleozlx you're right — `==` in `[build-system].requires` forces every source build (sdists, offline mirrors, distro packaging) onto one release, and it also makes `python -m build --no-isolation` hard-fail with `ERROR Missing dependencies` unless that exact version is preinstalled. I verified that failure mode; the docs tell people to run exactly that command for `flashinfer-jit-cache`.

Reworked: **determinism belongs in the release pipeline, not in consumer-facing metadata.**

- All `[build-system].requires` now carry the same range as the runtime constraint — no `==` anywhere in package metadata.
- Release builds of `flashinfer-jit-cache` (the only wheel that ships ABI-bound `.so` files) pin one version via `PIP_CONSTRAINT=scripts/build_constraints.txt` in `build_flashinfer_jit_cache_whl.sh`. Verified that `PIP_CONSTRAINT` does reach `python -m build`'s isolated env: without it the build resolved `0.1.13.post2`, with it `0.1.11`.

That keeps the original goal — the same source tree can't produce different ABIs on different days, which is how v0.6.16rc5 and v0.6.16 ended up differing — while leaving source builders on a range. Diff also trimmed (306 → 149 lines): dropped a build-time warning and README/docs notes that only existed to work around the hard pin, and cut the comment blocks down (rationale lives in the PR description).

Also folded in an external review pass: the constraint test now checks parsed TOML/requirements instead of raw text (it could previously pass on a matching comment while the real declaration drifted), fails instead of skipping on a real `git grep` error, and widened its scan to `apache[-/]tvm-ffi` — which immediately surfaced the `TVM_FFI_REF` override in `scripts/setup_test_env.sh`, now explicitly classified as an intentional escape hatch.

YangXu1990uiuc:
Two more changes.

**1. Floor lowered 0.1.11 → 0.1.10.** 0.1.10 exports an identical `TVMFFI*` symbol set to 0.1.11 and has identical boundary layouts (`sizeof(Optional<Function>)` and the `ModuleObj::GetFunction` return slot are both 8 bytes on 0.1.10/0.1.11/0.1.12/0.1.13.post2). Dropping a measured-compatible version was a policy break rather than an ABI necessity, and there are still consumers declaring `==0.1.10`. A lower floor can only ever keep us out of someone else's resolution conflict.

**2. Added `scripts/verify_jit_cache_abi.py`, run by the jit-cache build.** We were advertising a compatibility range we never tested. This resolves the floor from the declared constraint, downloads that `libtvm_ffi.so`, and fails if any kernel in the built wheel references a symbol it does not export. CPU-only, no GPU, no torch; `FLASHINFER_SKIP_ABI_VERIFY=1` opts out for offline builds.

It reproduces the incident against the artifacts that actually shipped:

```
# flashinfer_jit_cache-0.6.16      (built against tvm-ffi 0.1.13)
kernels reference 5 TVMFFI symbols: [... 'TVMFFIGetCustomAllocator' ...]
FAIL apache-tvm-ffi 0.1.10 does not export ['TVMFFIGetCustomAllocator']
       needed by fp4_kv_quant.so
OK   apache-tvm-ffi 0.1.13.post2 exports everything the kernels need
  exit=1

# flashinfer_jit_cache-0.6.16.post1 (built against 0.1.13.post0)
OK   apache-tvm-ffi 0.1.10 exports everything the kernels need
OK   apache-tvm-ffi 0.1.13.post2 exports everything the kernels need
  exit=0
```

Scope is stated in the docstring and worth repeating: this catches the undefined-symbol class only. The second break in 0.1.13 — object `Optional<T>` growing 8→16 bytes and with it the return slot of the virtual `ModuleObj::GetFunction` we override — keeps every symbol name intact and is invisible to `nm`. Closing that needs a GPU leg that installs the wheel at the floor with `FLASHINFER_DISABLE_JIT=1` (so a failed `.so` load can't be masked by a silent local rebuild) and exercises the two `ModuleObj` subclasses. Left as a follow-up rather than bolted on here.

YangXu1990uiuc:
Dropped the `!=` exclusions — constraint is now just `apache-tvm-ffi>=0.1.10,<0.2`.

Rationale: all three ABI-divergent releases (0.1.13, .post0, .post1) are yanked upstream, so a resolver will not select them. Reaching one requires an explicit `==` pin, and we're in direct contact with both major consumers (vLLM, SGLang) — not a scenario worth carrying a permanent exclusion list for. Accumulating `!=` clauses per incident is the pattern this repo has already pushed back on.

Kept the *reason* in a comment next to the requirement, since it explains why the floor and the release pin sit where they do.

Verifier still behaves correctly on the simplified range: it resolves the floor to 0.1.10, passes `0.6.16.post1`, and still fails `0.6.16` on the missing `TVMFFIGetCustomAllocator`.

yzh119:
cc @cyx-6 @tqchen have you ever looked into this issue before?

tqchen:
yes, i think we were in a thread the abi issue has been solved, the current range based approach looks fine to me

PR #4082, opened 2026-07-22
feat(moe): standalone trtllm-gen routing op + decomposed tests/moe routing matrix

aleozlx:
## 📌 Description

`tests/moe` multiplies routing-method axes against the full quant × layout × shape matrix, which is the main driver of the suite's size (e.g. `test_deepseekv3_routing` ~7.9k raw cases; the routed fused dense grid carried a 6× routing multiplier; the renormalize shard trio measured 123 min / 1,398 passing cases pre-split). Routing correctness is orthogonal to quantization — they share no kernel code path — so this PR decomposes `(routing methods) × (other params)` into `(routing methods, dense but cheap) + (few routing methods) × (other params)`:

**New standalone routing op + test**
- Split the `Routing::Runner` dispatcher out of `trtllm_fused_moe_runner.cu` into its own TU and expose it via a new TVM-FFI export `trtllm_gen_routing` with a lightweight JIT module (routing kernels only — no batched-GEMM stack).
- Python API `flashinfer.fused_moe.trtllm_gen_routing(...)`. Note: in from-logits mode the kernels emit no expert ids (`mPtrTopKIds` is input-only, `mPtrTopKPacked` is pipeline scratch); `topk_ids` is reconstructed from the permutation via `cta_idx_xy_to_batch_idx[slot // tile_tokens_dim]`.
- `tests/moe/test_trtllm_gen_routing.py`: all 9 routing methods × {top_k, num_experts (model shapes up to 512 experts / top_k 22), n_group/topk_group, bias & logits dtypes, tile_tokens_dim, load skew, edge token counts} against the existing `routing_reference_*` host oracles, plus expert-parallel shard cases and grouped-argument rejection tests.

**Fused-matrix reduction (coverage relocated, not deleted)**
- `test_trtllm_gen_routed_fused_moe.py`: dense grid pinned to Renormalize/packed + an 18-case method × format × quant parity smoke (routing here is host-precomputed, so the method axis only varied reference math now tested directly).
- `test_trtllm_gen_fused_moe.py::test_deepseekv3_routing`: dropped `kimi_k2`/`DSLite`/`GLM4_MoE` configs and orphaned intermediate sizes; kept `DSv3`, `nemotron_3_super`, and both fused-shared-expert variants (that fusion isn't covered standalone).
[...]

PR #4361, opened 2026-08-05
Add CUTLASS Primitives and Task Scheduling Blackwell MoE kernels

nekorobov:
## 📌 Description

This PR adds an experimental PrimsTS MoE backend for Blackwell GPUs under `flashinfer.fused_moe` and `flashinfer.prims_ts`. The kernels are ported from TRT-LLM Gen and built from CUTLASS primitives plus the task-scheduled (TS) scheduler. PrimsTS executes the FC1 and FC2 expert GEMMs while reusing the TRT-LLM Gen routing and finalization flow.

The new `prims_ts_*` entry points are modeled after the corresponding `trtllm_*` APIs, with backend-specific arguments and support restrictions documented and validated before launch. Existing TRT-LLM Gen APIs remain available and continue to use the native implementation.

### Supported configurations

- BF16 MoE: logits-based and routed APIs with BF16 activations and weights.
- FP8 per-tensor MoE: E4M3 activations and weights with per-tensor scalar scales.
- FP8 block-scale MoE: logits-based and routed APIs with DeepSeek FP8 and MXFP8 scale layouts.
- FP4 block-scale MoE: logits-based and routed APIs for NVFP4 × NVFP4, MXFP4 × MXFP8, and MXFP4 × BF16.
- Batched/grouped FC1 and FC2 GEMMs with FP32 accumulation, supported fused gated activations, bias, clamp, and fused quantization epilogues.
- Static hardware-grid and dynamic persistent CLC scheduling, PDL where supported, autotuning, custom/fake ops, trace definitions, and validation helpers.

### Current limitations

MXINT4 × BF16 is not supported. Other remaining gaps include selected LoRA, fused shared-expert, routing replay/policy, activation, and DSMEM split-K combinations. Accuracy qualification currently targets SM100a/B200; SM103a/B300 is architecture-gated but has not completed signoff qualification.

### Benchmark

B200, E/local-E=128, top-k=4, H=I=3072, SwiGLU

| Quant mode | Routing | Tokens | TRT default (ms) | TRT tuned (ms) | Prims default (ms) | Prims tuned (ms) | Tuned Prims / TRT |
|---|---|---:|---:|---:|---:|---:|---:|
| BF16 | logits | 1 | 0.056480 | 0.056000 | 0.059120 | 0.058080 | 1.037x |
| BF16 | logits | 32 | 0.617152 | 0.615536 | 0.629471 | 0.616959 | 1.002x |
| BF16 | logits | 128 | 0.987232 | 0.984191 | 0.998287 | 0.985871 | 1.002x |
| BF16 | logits | 512 | 1.367231 | 1.030591 | 1.501358 | 1.034463 | 1.004x |
| Fp8-Per-Tensor | logits | 1 | 0.036400 | 0.036304 | 0.039552 | 0.037536 | 1.034x |
| Fp8-Per-Tensor | logits | 32 | 0.353199 | 0.353264 | 0.370976 | 0.354448 | 1.003x |
| Fp8-Per-Tensor | logits | 128 | 0.517103 | 0.516880 | 0.539456 | 0.518624 | 1.003x |
| Fp8-Per-Tensor | logits | 512 | 0.638703 | 0.546576 | 0.915023 | 0.547472 | 1.002x |
| Fp8-Block | logits | 1 | 0.046415 | 0.045952 | 0.047456 | 0.045808 | 0.997x |
| Fp8-Block | logits | 32 | 0.074432 | 0.057904 | 0.133855 | 0.054576 | 0.943x |
| Fp8-Block | logits | 128 | 0.199392 | 0.088592 | 0.328543 | 0.079072 | 0.893x |
| Fp8-Block | logits | 512 | 0.694366 | 0.189536 | 1.093375 | 0.179423 | 0.947x |
| NvFP4xNvFP4 | precomputed | 1 | 0.024976 | 0.025008 | 0.028448 | 0.028448 | 1.138x |
| NvFP4xNvFP4 | precomputed | 32 | 0.191377 | 0.191584 | 0.219792 | 0.195217 | 1.019x |
| NvFP4xNvFP4 | precomputed | 128 | 0.309536 | 0.309392 | 0.348336 | 0.312592 | 1.010x |
| NvFP4xNvFP4 | precomputed | 512 | 0.368256 | 0.319840 | 0.499232 | 0.324416 | 1.014x |
| MxFP4xBf16 | precomputed | 1 | 0.032417 | 0.032624 | 0.205680 | 0.186992 | 5.732x |
| MxFP4xBf16 | precomputed | 32 | 0.232608 | 0.229984 | 2.195889 | 2.184864 | 9.500x |
| MxFP4xBf16 | precomputed | 128 | 0.392384 | 0.379632 | 3.775120 | 3.652624 | 9.621x |
| MxFP4xBf16 | precomputed | 512 | 0.814144 | 0.416896 | 8.844993 | 3.688272 | 8.847x |
| MxFP4xMxFP8 | precomputed | 1 | 0.026880 | 0.027264 | 0.033504 | 0.030304 | 1.112x |
| MxFP4xMxFP8 | precomputed | 32 | 0.185503 | 0.185344 | 0.281199 | 0.187632 | 1.012x |
| MxFP4xMxFP8 | precomputed | 128 | 0.300767 | 0.300864 | 0.456000 | 0.302623 | 1.006x |
| MxFP4xMxFP8 | precomputed | 512 | 0.479248 | 0.318720 | 0.964047 | 0.322560 | 1.012x |

MxFP4xBf16 performance difference will be addressed in the future PRs.

### ✅ Pre-commit Checks

- [x] `pre-commit` is available in the development environment.
- [x] The local Git hook has been installed with `pre-commit install`.
- [x] `pre-commit run --all-files` passes on the current branch.

## 🧪 Tests

- [x] Added standalone PrimsTS batched-GEMM correctness, schedule-validation, metadata, and no-hang coverage.
- [x] Added MoE backend integration and routed-input parity coverage.
- [x] Added trace examples and regenerated `tests/trace/fi_trace_out/` definitions.
- [x] Added focused regressions for reusable routed FFI tensors and autotuner cache-key compatibility.
- [x] Reduced the shared MoE integration matrix so PrimsTS coverage no longer multiplies the full TRT-LLM test grid.
- [ ] Complete SM100/SM103 GPU CI signoff.

## Reviewer Notes

The PrimsTS backend shares TRT-LLM Gen routing/finalization semantics but substitutes task-scheduled batched GEMMs for FC1/FC2. Unsupported shape, dtype, layout, scale, routing, and activation combinations are rejected before launch.

### Remaining follow-ups

- [x] **Performance qualification before merge:** benchmark representative supported configurations against TRT-LLM Gen on B200/SM100a, publish the environment/shapes and median latency, and verify or revise the intended “on par” claim. SM103a/B300 results belong with its separate signoff.
- [ ] **Persistent kernel cache after merge:** move the current process-local PrimsTS compiled-GEMM cache to FlashInfer's persistent CuTe-DSL cache lifecycle with architecture/version/source/config invalidation and cold/warm-process tests. Track this in a linked issue with an owner; it is not a known correctness blocker.

Review focus: kernel correctness on CUDA 12.9/13.x, representative end-to-end backend coverage, and sustainable CI collection/runtime.


flashinfer/__init__.py
```diff
@@ -89,12 +89,18 @@
...
+    prims_ts_fp4_block_scale_routed_moe,
+    prims_ts_fp8_block_scale_moe,
+    prims_ts_fp8_block_scale_routed_moe,
+    prims_ts_fp8_per_tensor_scale_moe,
     trtllm_fp4_block_scale_moe,
     trtllm_fp4_block_scale_routed_moe,
     trtllm_fp8_block_scale_moe,
     trtllm_fp8_block_scale_routed_moe,
     trtllm_fp8_per_tensor_scale_moe,
-    trtllm_fp8_per_tensor_scale_routed_moe,
```
rosenrodt, line 97: Removed by accident?

tests/moe/test_trtllm_gen_fused_moe.py
```diff
@@ -441,19 +461,18 @@ def test_deepseekv3_routing(
...
 @pytest.mark.parametrize(
-    ("activation_type", "alpha_value", "beta_value", "clamp_value"),
+    "activation_type",
     [
-        pytest.param(ActivationType.Swiglu, None, None, None, id="Swiglu"),
-        pytest.param(ActivationType.Geglu, None, None, None, id="Geglu"),
-        pytest.param(ActivationType.Situ, None, None, None, id="Situ_Defaults"),
-        pytest.param(ActivationType.Situ, 4.0, 25.0, None, id="Situ_kimi-k3"),
+        pytest.param(ActivationType.Swiglu, id="Swiglu"),
+        pytest.param(ActivationType.Geglu, id="Geglu"),
```
rosenrodt, line 475: Do we intend to remove alpha/beta/clamp and Situ tests?

If not, we may want specify which combo works for MoeGemmBackend.PRIMS_TS and for MoeGemmBackend.TRTLLM

tests/moe/test_trtllm_gen_fused_moe.py
```diff
@@ -1991,243 +2156,11 @@ def test_fp8_block_scale_moe_fused_shared_experts_reject_ep():
         )
 
 
-def test_fp4_block_scale_moe_fused_shared_experts_reject_ep():
```
rosenrodt, line 1994: Let's keep this test and see if MoeGemmBackend.PRIMS_TS satisfies the constraint.

tests/moe/test_trtllm_gen_fused_moe.py
```diff
@@ -1304,149 +1611,7 @@ def test_deepseek_ngroup1_block_per_token_routing(
...
-
-@pytest.mark.parametrize("enable_pdl", [False, True])
-@pytest.mark.parametrize(
-    "num_tokens, num_experts, top_k, hidden_size, intermediate_size",
-    [
-        pytest.param(8, 384, 6, 512, 512, id="array-topk"),
-        pytest.param(1025, 896, 16, 128, 128, id="lane-owned-topk"),
-    ],
-)
-def test_deepseek_ngroup1_sigmoid_bias_output_uses_unbiased_sigmoid(
```
rosenrodt, line 1318: Is removing existing tests intended? If not, let us keep the deleted tests and see if MoeGemmBackend.PRIMS_TS can satisfy the contract. Same applies to many existing test cases

flashinfer/prims_ts/moe/runner.py
```diff
@@ -0,0 +1,2408 @@
...
+        tactics: List[Any],
+        profile: OptimizationProfile,
+        **kwargs,
+    ) -> bool:
+        del profile
+        moe_inputs = MoeRunnerInputs.from_list(inputs)
+        hidden_states = moe_inputs.hidden_states
+        for tactic in tactics:
+            try:
+                self.forward(inputs, tactic=tactic, **kwargs)
```
rosenrodt, line 441: Minor/non-blocking: I think we can leverage this entrypoint to perform cpu-only compilation without actually running kernels. Additionally, can probably spawn threads to compile in paralalle.

flashinfer/prims_ts/batched_gemm/README.md
```diff
@@ -0,0 +1,294 @@
...
+  --Task-->    task-owned action
+  *            optional or feature-dependent path
+
+  A/B tensor args                         scale-factor args*           page/route args*
+        |                                       |                            |
+        | LoadA/LoadB: TMA                       | LoadSfA/LoadSfB: TMA       | Gather*: LDGSTS
+        v                                       v                            | or TMA gather4
+  +-----+----+                          +-------+--------+                   v
+  | SmemA/B  |                          | SmemSfA/SfB    |          +--------+---------+
+  | A/B buffer |                          | SF buffer        |          | SmemGather*      |
```
rosenrodt, line 147: Minor: text box boundary misaligned

flashinfer/__init__.py
nekorobov: Reverted

tests/moe/test_trtllm_gen_fused_moe.py
nekorobov: Reverted

tests/moe/test_trtllm_gen_fused_moe.py
nekorobov: Done

flashinfer/prims_ts/moe/runner.py
nekorobov: Done

flashinfer/prims_ts/batched_gemm/README.md
nekorobov: Done

flashinfer/quantization/fp4_quantization.py
```diff
@@ -802,6 +802,7 @@ def _fake_e2m1_and_ufp8sf_scale_to_float_sm100(
 
     # Register the module
     return SimpleNamespace(
+        raw_module=module,
```
aleozlx: quick question: what's the purpose of this

flashinfer/fused_moe/backends/trtllm/bf16_op.py
```diff
@@ -0,0 +1,255 @@
+# Copyright (c) 2026 by FlashInfer team.
```
aleozlx: why is this file here in this folder?

flashinfer/fused_moe/backends/trtllm/sm100_runner.py
```diff
@@ -0,0 +1,398 @@
...
+from flashinfer.tllm_enums import (
+    ActivationType,
+    DtypeTrtllmGen,
+    Fp8QuantizationType,
+    WeightLayout,
+    trtllm_gen_dtype_has_scale,
+)
+
+
+def create_trtllm_moe_runner_class(
```
aleozlx: notes: why is this needed?

flashinfer/fused_moe/shared/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+from .inputs import (
+    MoEInputs,
+    MoeRunnerInputs,
+    RoutingInputMode,
+    alloc_trtllm_moe_output,
+    unpack_trtllm_moe_output,
+)
+from .tuning import make_moe_tuning_config
+
+__all__ = [
```
aleozlx, line 27: these seems replicated vs existing ones

aleozlx (commented):
code reivew : to be continued

flashinfer/fused_moe/shared/__init__.py
nekorobov: Great catch, fixed

flashinfer/quantization/fp4_quantization.py
nekorobov: It was a workaround.

Fixed: added nvfp4_quant_and_per_token_scale_out_sm100 as a registered custom op that writes into caller-provided outputs. The allocating wrapper now calls that, Prims-TS uses the out-variant, and raw_module is removed.

flashinfer/fused_moe/backends/trtllm/bf16_op.py
nekorobov: It was only there to break a circular import: Prims-TS needed validate_bf16_gemm1_activation_params, which lived next to the BF16 op in core.py.

Fixed: moved the validator to fused_moe/shared/validation.py, put the BF16 op back inline in core.py with the other TRT-LLM cubin ops, and deleted backends/trtllm/bf16_op.py / validation.py.

flashinfer/fused_moe/backends/trtllm/sm100_runner.py
nekorobov: It preserved the old nested-class pattern that closed over the JIT moe_op. Not needed once the runner takes moe_op explicitly.

Fixed: MoERunner is now a normal module-level class with moe_op as its first __init__ arg (same pattern as the Prims-TS runners). The factory is gone; Tactic caches stay separate via id(self.moe_op) in the cache key.

tests/conftest.py
```diff
@@ -4,6 +4,19 @@
 from pathlib import Path
 from typing import Any, Dict, Set
 
+
+def _configure_cute_dsl_cache_dir():
```
aleozlx, line 8: i've seen this somewhere before, slightly surprised it is new here

adding @bkryu to double check

csrc/trtllm_fused_moe_kernel_launcher.cu
```diff
@@ -1025,8 +1025,6 @@ inline void check_gemm1_bias_mn(Optional<TensorView> const& gemm1_bias,
                                 batchedGemm::gemm::BiasType bias_type, int32_t num_tokens,
                                 int32_t top_k, int32_t intermediate_size) {
   if (bias_type == batchedGemm::gemm::BiasType::None) {
```
aleozlx, line 1128: The global reject that used to live here (`TVM_FFI_ICHECK(!gemm1_bias.has_value())` when `bias_type` is `None`) was removed — apparently so the FP8 block-scale path can accept a provided `gemm1_bias` with `BiasType::None` (it validates shape/dtype itself further down, around the `check_moe` for block scale). Side effect: every other launcher going through `check_gemm1_bias_mn` now silently *ignores* a passed `gemm1_bias` instead of erroring. Consider scoping the exception (e.g. an `allow_bias_with_none_type` parameter defaulted to `false`) so the other paths keep failing loudly on a bias that would never be applied.

csrc/trtllm_fused_moe_kernel_launcher.cu
```diff
@@ -1829,6 +1830,1382 @@ class Bf16MoeLauncher : public FusedMoeLauncher {
...
+    bool has_precomputed_indices = expert_indices.ndim() == 2 && expert_indices.size(0) > 0;
+    if (has_precomputed_indices) {
+      workspace.routing_expert_indexes =
+          static_cast<int*>(const_cast<void*>(expert_indices.data_ptr()));
+    }
+    bool has_precomputed_weights = expert_weights.ndim() == 2 && expert_weights.size(0) > 0;
+    if (has_precomputed_weights) {
+      FusedMoeLauncher::expert_weights = alloc_tensor(
+          {args->num_tokens, args->top_k}, expert_weights.dtype(), hidden_states.device());
+      workspace.expert_weights = const_cast<void*>(expert_weights.data_ptr());
```
aleozlx, line 2038: When `has_precomputed_weights` is true, this allocates a fresh `FusedMoeLauncher::expert_weights` that nothing ever writes — the workspace points at the caller's tensor — and `collect_routing_outputs()` then returns that unwritten buffer in slot 0. The Python side compensates via `_select_expert_weights` in `flashinfer/prims_ts/moe/runner.py`, so this is not a correctness bug, but it is a wasted `[num_tokens, top_k]` allocation per call and a confusing output contract. Consider skipping the allocation in this branch (and documenting that slot 0 is undefined in precomputed mode), or pointing slot 0 at the caller's tensor instead.

flashinfer/prims_ts/batched_gemm/README.md
```diff
@@ -0,0 +1,294 @@
...
+`--benchmark` runs the correctness check first; add `--skip-ref-check` for pure
+timing:
+
+```bash
+python -m batched_gemm.batched_gemm_run --benchmark --skip-ref-check
+```
+
+> Run as a **module** from the `dense_gemm_ts/` directory — the runner uses
+> package-relative imports. Shape and feature knobs (`--tile-n`, `--dtype-a`,
+> `--tile-scheduler`, …) are listed in `--help`.
```
aleozlx: Leftover from the source repo layout — there is no `dense_gemm_ts/` directory in this tree, and `python -m batched_gemm.batched_gemm_run` only resolves with `flashinfer/prims_ts/` as the working directory. Suggest pointing at the installed package path, which works from anywhere:

```suggestion
> Run as a **module** — the runner uses package-relative imports, e.g.
> `python -m flashinfer.prims_ts.batched_gemm.batched_gemm_run` (the command
> blocks above use the short form, which requires `flashinfer/prims_ts/` as the
> working directory). Shape and feature knobs (`--tile-n`, `--dtype-a`,
> `--tile-scheduler`, …) are listed in `--help`.
```

aleozlx (commented):
## Claude further review summary

**Layout compatibility:** Confirmed the PrimsTS MoE backend consumes the identical special weight layout as the TRT-LLM Gen backend: gate/up row interleaving, the epilogue-tile row shuffle (the `shuffle_matrix` tables in `batched_gemm_run.py` are byte-identical to `srcToDstBlk16/32RowMap` in `flashinfer/utils.py`), TRT-LLM Gen's BlockMajorK packing (consumed natively by the CuTe-DSL kernel), and the same scale-factor layouts. Weights preprocessed for the `trtllm_*` APIs work unchanged with the `prims_ts_*` APIs; `support.py` rejects unshuffled weights on every dtype path. Remaining layout gaps (per the README): unshuffled dispatch and MXINT4.

**Review findings:**
1. The FP8 per-tensor runner constructions dropped `use_per_token_scaling=use_routing_scales_on_input`, desyncing autotuner tactic-validity queries from the actual launch configuration under Llama4 routing (`usePerTokenScalingGemm1` is derived from `mUseRoutingScalesOnInput` at launch). **Fixed by the merge commit below** — the logits path explicitly, the routed path via main's merged UnpackedPrecomputed changes.
2. `check_gemm1_bias_mn` no longer rejects a provided `gemm1_bias` when the bias type is `None` — intentional for the new FP8 block-scale bias support, but the relaxation is global (inline comment).
3. `StagedMoeLauncher::prepare_routing` allocates an `expert_weights` tensor that is never written in precomputed-weights mode (benign — the Python side substitutes the caller's tensor — but a confusing output contract; inline comment).
4. Stale `dense_gemm_ts/` run instructions in the batched_gemm README (inline suggestion).

**Merge:** Merged latest main (`27a5a294`) into this branch as `e1bf6b8d`. CLAUDE.md / pyproject.toml conflicts were additive; in `fused_moe/core.py` the nested `MoERunner` re-added by main was dropped in favor of this PR's extracted `TrtllmMoERunner`, with main's new routed FP8 per-tensor ABI (`routing_input_mode` + `topk_weights` leading args, dispatch on `routing_logits is None` so UnpackedPrecomputed routing doesn't fall through to the logits entry point) ported into `backends/trtllm/sm100_runner.py::forward()`, and `use_per_token_scaling` restored on the logits path. The PR is mergeable again; C++ compile verification is deferred to CI.

*Review + merge assisted by Claude (Claude Code).*

aleozlx:
code review: done with minor comments

resolved merge conflict

our separate CI issue was resolved so restarting CI now

aleozlx:
@flashinfer-bot run

aleozlx:
@flashinfer-bot run

aleozlx:
@flashinfer-bot run

aleozlx:
@flashinfer-bot run

tests/conftest.py
bkryu: I checked what is happening here. Here we are deliberately setting `CUTE_DSL_CACHE_DIR` to override CuTe-DSL's internal compilation cache, which can hold the IR and other things that get the DSL likes to store, not the exported compiled kernel that gets disk-cached for `mm_fp4`, `b12x` MoE, and a few others. 

In short, it is separate from what you are remembering.

While redirecting the DSL internal cache to somewhere else in the test context is fine, I think a good suggestion would be for prims_ts kernels to adopt the kernel cache like the b12x MoE kernels did in #4331 or #4029

aleozlx:
@flashinfer-bot run

bkryu:
@flashinfer-bot run

bkryu:
@flashinfer-bot run

flashinfer/prims_ts/batched_gemm/README.md
nekorobov: Done

csrc/trtllm_fused_moe_kernel_launcher.cu
nekorobov: Done

csrc/trtllm_fused_moe_kernel_launcher.cu
nekorobov: Done

aleozlx:
@flashinfer-bot run

aleozlx:
looking into what's remaining


**A. PR-caused — independently confirmed by me on B200 (exact CI image, cu129, DSL 4.7.0) — ⚠️ STILL OPEN**

Root cause: MxFP4×BF16 prims_ts kernels fail NVVM compilation for sm_100a (new code introduced by this PR; deterministic; still unfixed on the current head):

`pytest tests/prims_ts/test_batched_gemm_mx.py -k MxFp4Bf16`

TestMxFp4Bf16Fc1::test_tile8_swiglu_persistent_correctness
TestMxFp4Bf16Fc1::test_tile8_swiglu_fp16_output_correctness
TestMxFp4Bf16Fc1::test_tile16_cluster2_swiglu_persistent_correctness
TestMxFp4Bf16Fc2::test_tile8_persistent_correctness
TestMxFp4Bf16Fc2::test_tile16_cluster2_persistent_correctness
TestMxFp4Bf16Fc2::test_tile128_static_correctness

Same root cause reaching the integration suite (confirmed instance; every PrimsTS+MxFP4xBf16 variant in the file fails identically):

`pytest "tests/moe/test_trtllm_gen_fused_moe.py::test_sigmoid_routing[PrimsTS-Swiglu-Shuffled_MajorK_PrimsTS-Sigmoid_128e_top8-MxFP4xBf16-1024-1024-8]"`

Control: the other 377 standalone tests/prims_ts tests all pass on B200 — including every hang-regression test — so this is the only broken kernel family on SM100.

**B. PR-caused — root-caused in code; ✅ FIXED by bba65a93**

~~Root cause: autotuner cache-key format change (is not None → _has_payload, shared/tuning.py) breaks forced-tactic lookup for the fp8 routed path only, because only that path passes an empty expert_weights placeholder. The fp4 key hit, the fp8 key didn't. The twelve = these six × 2 CUDA versions:~~

~~`pytest "tests/moe/test_trtllm_gen_moe_autotune_tactics.py::test_trtllm_fp8_routed_moe_all_tactics_correctness[4-128-3072-4096-{16,23,128}-{DeepSeekFp8,MxFp8}]"`~~

Fixed by the author in bba65a93 (test-side; in the current head). ✅ Confirmed by pipeline #63784427 on the merged head — the autotune failures are gone on all platforms. Residual (non-test): old persisted autotune configs are invalidated by the new key format — worth one line in the PR description.

**C. Reported as failures, actually cascade/capacity — independently characterized — ⚠️ STILL OPEN**
GB200/cu12.9, test_trtllm_gen_fused_moe, "5,829 failures, cudaErrorLaunchFailure" — one poisoned CUDA context cascades to every subsequent test in the job; the seed failure is most plausibly the family from (A) via the aarch64 DSL build (compiles into a bad cubin instead of being rejected). This is the one link I could not verify without a GB200.
B300/GB300 "98,766 failures — not executed due to timeout" — not failures at all: the file collects 49,383 tests (~2.7× main; independently counted) × 2 CUDA jobs = exactly 98,766. Pure walltime exhaustion from the moe_gemm_backend duplication.

How the expansion is formed (mechanics):

- **The ×2 is injected by a conftest hook, invisibly.** `tests/moe/conftest.py` parametrizes any test that merely *names* `moe_gemm_backend` as an argument — no decorator at the test site:

  ```python
  # tests/moe/conftest.py
  def pytest_generate_tests(metafunc):
      if "moe_gemm_backend" in metafunc.fixturenames:
          metafunc.parametrize(
              "moe_gemm_backend",
              [
                  pytest.param(MoeGemmBackend.TRTLLM, id="TRTLLM"),
                  pytest.param(MoeGemmBackend.PRIMS_TS, id="PrimsTS"),
              ],
          )
  ```

  The big integration tests (`test_renormalize_routing`, `test_sigmoid_routing`, `test_deepseek_routing`, the routed variants, …) all take this argument, so the ×2 lands on the outermost axis of their already-stacked `num_tokens × hidden × intermediate × moe_impl × routing × weight_processing × activation` products.

- **An inner axis widened too, so it's 1.66×, not a clean 2×.** New PrimsTS-only `weight_processing` entries were added next to the existing ones in `tests/moe/test_trtllm_gen_fused_moe.py`, and the PrimsTS variant is compatible with *more* `moe_impls` than its TRTLLM sibling:

  ```python
  # tests/moe/test_trtllm_gen_fused_moe.py (weight_processing params)
  pytest.param({..., "compatible_moe_impls": [FP4Moe, FP8PerTensorMoe, FP8BlockScaleMoe],
                "compatible_gemm_backends": [MoeGemmBackend.TRTLLM]},   id="Shuffled_MajorK"),
  pytest.param({..., "compatible_moe_impls": [..., BF16Moe],            # broader list
                "compatible_gemm_backends": [MoeGemmBackend.PRIMS_TS]}, id="Shuffled_MajorK_PrimsTS"),
  pytest.param({..., "compatible_gemm_backends": [MoeGemmBackend.TRTLLM,
                MoeGemmBackend.PRIMS_TS]},                              id="Shuffled_BlockMajorK"),
  ```

  Result: `test_trtllm_gen_fused_moe.py` collects **49,383** nodes — 30,808 PrimsTS + 18,575 TRTLLM (`pytest --collect-only -q` / `-k PrimsTS`).

- **Incompatible combos are filtered at runtime, not at collection.** The guard is a `pytest.skip` inside the test body (`tests/moe/utils.py:194`):

  ```python
  compatible_gemm_backends = weight_processing.get(
      "compatible_gemm_backends", [MoeGemmBackend.TRTLLM]
  )
  if moe_gemm_backend not in compatible_gemm_backends:
      pytest.skip(f"Incompatible: {moe_gemm_backend.value} backend with ...")
  ```

  Every incompatible combination is still collected, put in the shard manifest, scheduled, and set up before it skips — inflating both the schedule and the "not executed due to timeout" tally.

- **Wall-clock grows faster than node count.** Each genuinely-running PrimsTS combination JIT-compiles a CuTe-DSL kernel on first use (`flashinfer/prims_ts/batched_gemm/batched_gemm_run.py` → `cute.compile`, minutes per new dtype/tile/layout config), whereas the TRTLLM variants reuse prebuilt cubins fetched from the artifact repo.

- **Most of the duplicated coverage is redundant.** The same PrimsTS kernels are tested exhaustively and directly in `tests/prims_ts/` (383 tests, ~18 min total on B200). Suggested shape: keep exhaustive coverage standalone, run PrimsTS through the integration file only for a representative slice.

- **A compounding mitigation already exists: #4082.** It decomposes the *routing* axis of the same file on the same argument (routing correctness ⊥ quantization — no shared kernel code): a dense-but-cheap standalone routing suite + a small routing sample in the integration grid, removing e.g. the 6× routing multiplier from the routed dense grid. Since the `moe_gemm_backend` ×2 here multiplies whatever base matrix remains, #4082's reduction cuts both the TRTLLM and PrimsTS halves — the two fixes compound. Same decomposition pattern is the natural fix for the backend axis itself.

**D. NOT from the PR — pre-existing; ✅ CLEARED by merging current main (262795b8)**

~~RTX Pro 6000 / CUDA 13.0, all AttributeError: 'MoE*Kernel' object has no attribute '_collapse_to_vmk' (bot marked "Old"):~~

~~tests/moe/test_b12x_fused_moe.py — TestB12xFunctional (62), TestB12xWrapper (22), TestRelu2Activation (20), TestMicroKernel (2), TestB12xApiConsistency (1)~~
~~tests/moe/test_unified_moe_b12x.py::TestUnifiedB12xConformance (13)~~

These matched main's [#4602](https://github.com/flashinfer-ai/flashinfer/pull/4602) ("restore SM12x MoE kernels broken by self-resolved helper in borrowed dense methods"), which landed after this PR's previous merge-base. The branch now contains current main tip (f479fe28), so these should vanish on the next pipeline run.

**E. PR-caused, surfaced by pipeline #63784427 — independently reproduced on B200 — ⚠️ NEW, OPEN**

`RuntimeError: from_dlpack received an invalid capsule. Note that DLTensor capsules can be consumed only once` — deterministic on every SM100/103 platform, both CUDA versions, in:

`pytest tests/moe/test_prims_ts_fp8_block_scale_moe.py tests/moe/test_trtllm_gen_routed_fused_moe.py`

- failing frame chain: `fused_moe/backends/prims_ts/fp8_op.py:745 → :965 → flashinfer/prims_ts/moe/runner.py:2235`, where `forward()` converts the routing outputs:

  ```python
  # flashinfer/prims_ts/moe/runner.py:2235
  (expert_weights, expanded_idx_to_permuted_idx, ..., gemm2_output) = [
      torch.from_dlpack(t) for t in routing_out
  ]
  ```

- torch's traceback lands in the raw-capsule fallback branch of `torch/utils/dlpack.py` (`torch._C._from_dlpack(dlpack)` with `dlpack = ext_tensor`), meaning the elements of `routing_out` reach Python as **raw one-shot DLPack `PyCapsule`s** (no `__dlpack__` method), each consumable exactly once.
- the failing tests (`test_prims_ts_mxfp8_block_scale_routed_modes_match_logits` and its routed sibling) invoke the op repeatedly across routing input modes — a reused/cached capsule on the second invocation triggers the error.
- suggested fix: have the FFI boundary return NDArray-like objects (fresh `__dlpack__` per consumer) or mint a fresh capsule per call, rather than caching one-shot capsules across invocations.

Bottom line: ~~after the author's autotune fix (B) and a main merge (kills D),~~ B and D are done — remaining code work is the MxFP4×BF16 kernel family (**A**) and the DLPack capsule reuse in the prims_ts routing path (**E**), plus the suite-size problem (**C**), which is CI capacity, not correctness.


---

**Error-coverage map for the latest pipeline [#63784427](https://nv/flashinfer-ci/-/pipelines/63784427)** (run on the merged head 262795b8; how each reported item maps to the buckets above):

| Pipeline report item | Bucket | Reading |
|---|---|---|
| `tests.moe.test_trtllm_gen_fused_moe.py` — 136,176 failures, "not executed due to timeout" (B300/GB200 × cu12.9/13.0) | **C** | Pure walltime arithmetic: ~34k never-executed nodes per job × 4 jobs. Zero of these are observed test failures. **A is masked, not shown**: the `PrimsTS`+`MxFP4xBf16` integration variants sit inside this not-executed mass, and A's direct evidence (`tests/prims_ts/test_batched_gemm_mx.py`) is outside this pipeline's `tests/moe` scope entirely — so this run neither confirms nor refutes A (which stands on the independent B200 repro). The earlier GB200 `cudaErrorLaunchFailure` cascade did not recur; walltime hit first. |
| `tests.moe.test_prims_ts_fp8_block_scale_moe` — 4 failures, `from_dlpack ... invalid capsule` | **E (new)** | 1 test × 4 jobs. Reproduced independently on B200 (exact CI image): the prims_ts FFI op returns `routing_out` tensors as raw one-shot DLPack capsules, converted at `flashinfer/prims_ts/moe/runner.py:2235`; `test_prims_ts_mxfp8_block_scale_routed_modes_match_logits` invokes the op across multiple routing modes, and the second conversion of a reused capsule throws. Deterministic on every SM100/103 platform and both CUDA versions. |
| `tests.moe.test_trtllm_gen_routed_fused_moe` — 4 failures, same error | **E (new)** | Same root cause via the routed integration path. |
| *(absent)* `test_trtllm_gen_moe_autotune_tactics` | **B** | Fixed by bba65a93 — confirmed gone on all platforms. |
| *(absent)* RTX Pro 6000 `_collapse_to_vmk` failures | **D** | Cleared by syncing main (#4602 now in the branch) — RTX rows fully pass. |

Remaining work items: **A** (fix or gate MxFP4×BF16 prims_ts kernels on SM100), **C** (suite size — #4082 compounds), **E** (return NDArray objects or fresh per-call capsules from the prims_ts routing op instead of reusing one-shot DLPack capsules).

aleozlx:
@flashinfer-bot run

aleozlx:
## PR Review Screening

**CI verdict:** ✅ auto-run ok
**Review category:** `live` (rule fired: C1.2 net-new public interface family `prims_ts_*`; also C3.2 durable base-class/shared-abstraction changes and C3.1 experimental claimed without a tracked issue)
**Blocking checks:** S2 (template overwritten); C2.1 (perf parity claim without numbers in description)
**Release blocker:** no
**Early stop:** no

### Security
| Q | Answer | Evidence |
|---|--------|----------|
| S2 template overwritten | yes | body uses custom `# Summary / ## Feature support` format; 0 template headers, 0 checkboxes |
| S3 template obligations | unmet formally, met in substance | thorough description + 15k test LOC, but no checklist/Tests sections |

### Packaging
| Q | Answer | Evidence |
|---|--------|----------|
| C1.1 dependency bump | yes | `requirements.txt` + `pyproject.toml` add `nvidia-cutlass-dsl==4.7.0` (cu12/cu13 extras) |
| C1.2 public API changes | net-new family | `prims_ts_bf16_moe`, `prims_ts_bf16_routed_moe`, `prims_ts_fp8_per_tensor_scale_moe`, `prims_ts_fp8_block_scale_moe(+routed)`, `prims_ts_fp4_block_scale_moe(+routed)` exported via `flashinfer.fused_moe`; new `flashinfer.prims_ts` package. Signatures mirror the `trtllm_*` equivalents (prefix-interchangeable) — net-new, but shape-mirroring |
| C1.3 AOT/trace registration | partial gap | no nvcc `gen_*_module` (CuTe-DSL runtime compile → AOT n-a); trace templates registered + inventory updated (fcdcc5fb), but no `tests/trace/example.py` entries or regenerated `tests/trace/fi_trace_out/` JSONs per CLAUDE.md checklist |

### Presentation
| Q | Answer | Evidence |
|---|--------|----------|
| C2.1 perf claim backed by data | flag | "on par with TRT-LLM Gen … some configurations still have a performance gap" — no before/after numbers, GPU, or shapes in description |

### Implementation
| Q | Answer | Evidence |
|---|--------|----------|
| C3.1 experimental-track declared | claimed, no tracked issue | body says "experimental PrimsTS MoE kernels"; no experimental-terms issue linked |
| C3.2 durable areas touched | yes (durable) | `TunableRunner.precompile_tactics` hook in `flashinfer/autotuner/autotuner.py`; `fused_moe/core.py` restructured into `backends/`+`shared/`; autotuner cache-key semantics changed in `shared/tuning.py`; `csrc/trtllm_fused_moe_kernel_launcher.cu` +1.4k |
| C3.3 tests match behavior | yes, with caveats | 15k+ test LOC incl. per-kernel-family suites and hang-regression tests; known live failures (MxFP4×BF16 on SM100, dlpack capsule reuse) tracked in [the diagnosis comment](https://github.com/flashinfer-ai/flashinfer/pull/4361#issuecomment-5363720598); suite growth (~49k nodes in one file) currently exceeds CI walltime |
| C3.4 release blocker | no | feature PR |

Notes for the maintainer:
- The `live` routing matches reality: this introduces a parallel MoE backend family + scheduler package (~48k LOC) — an architecture-level addition.
- Experimental status should be formalized (tracked issue) so review applies the experimental bar deliberately rather than implicitly.

flashinfer/prims_ts/moe/runner.py
```diff
@@ -0,0 +1,2337 @@
...
+            tile_idx,
+            mn_limit,
+            num_non_exiting_ctas,
+            total_num_padded_tokens,
+            gemm1_output,
+            gemm1_output_scale,
+            activation_output,
+            activation_output_scale,
+            gemm2_output,
+        ) = [torch.from_dlpack(t) for t in routing_out]
```
aleozlx: **Confirmed crash (reproduced on B200 + all SM100/103 CI platforms):** the elements of `routing_out` arrive as raw one-shot DLPack `PyCapsule`s (torch's traceback lands in the no-`__dlpack__` fallback of `torch/utils/dlpack.py`), so this `torch.from_dlpack(t)` fails with `RuntimeError: from_dlpack received an invalid capsule` when the op is invoked repeatedly across routing input modes — e.g. `tests/moe/test_prims_ts_fp8_block_scale_moe.py::test_prims_ts_mxfp8_block_scale_routed_modes_match_logits` and the routed sibling in `test_trtllm_gen_routed_fused_moe.py`.

Fix direction: return NDArray-like objects (fresh `__dlpack__` per consumer) or mint a fresh capsule per call at the FFI boundary rather than reusing/caching one-shot capsules. Note the same unpack pattern exists at lines 638/931/1228/1381/1631/1941 — a shared helper would make the fix single-point.

flashinfer/fused_moe/shared/tuning.py
```diff
@@ -0,0 +1,143 @@
...
+
+    spec = {
+        "output": autotuner_initializer_empty,
+        "hidden_states": autotuner_initializer_randn,
+    }
+    if moe_inputs.routing_logits is not None:
+        spec["routing_logits"] = autotuner_initializer_rand
+    if _has_payload(moe_inputs.topk_ids):
+        spec["topk_ids"] = init_packed_topk_ids
+    if _has_payload(moe_inputs.expert_weights):
```
aleozlx: **Compat note (high confidence):** gating the dynamic-spec entries on `_has_payload()` (numel > 0) instead of main's `is not None` changes the autotuner profile-key format for the routed paths (empty placeholder tensors keep raw `(0,)` shape instead of bucketing). The in-repo test was reconciled in bba65a93, but **externally persisted tactics** (`FLASHINFER_AUTOTUNER_LOAD_FROM_FILE` / `autotune(cache=...)` files recorded under the old key format) silently stop matching and users re-tune after upgrade. Worth one line in the PR description/release notes; alternatively keep empty tensors in the spec to preserve key stability.

aleozlx (commented):
Agent code review per `docs/code_review_guidance.md` (posted on behalf of @aleozlx; an autonomous agent is also working this PR and will address findings). Findings labeled by confidence. Hardware-verified items reference the [diagnosis comment](https://github.com/flashinfer-ai/flashinfer/pull/4361#issuecomment-5363720598) for full evidence.

**Confirmed (reproduced on B200, exact CI image, DSL 4.7.0/cu12.9):**
1. **MxFP4×BF16 kernels fail NVVM compilation on `sm_100a`** — all 6 `TestMxFp4Bf16Fc1/Fc2` configs, deterministic `NVVM_ERROR_COMPILATION` in the cute-to-nvvm pipeline; likely in the BF16-activation path (`use_bf16_kbox_tma_a` / LDGSTS machinery). Repro: `pytest tests/prims_ts/test_batched_gemm_mx.py -k MxFp4Bf16`. On GB200/aarch64 this family plausibly compiles into a bad cubin instead → the `cudaErrorLaunchFailure` cascade. Either fix the kernel or gate the config out of `support.py` + skip the tests on SM100-family until fixed.
2. **DLPack capsule reuse crash** (inline comment on `runner.py:2235`).

**High confidence (code-read):**
3. Autotuner cache-key compat (inline comment on `shared/tuning.py:96`).
4. CI-budget: integration-suite growth (inline comment on `tests/moe/conftest.py:36`).
5. Trace checklist gap: the new `@flashinfer_api(trace=...)` ops lack `tests/trace/example.py` entries and regenerated `tests/trace/fi_trace_out/` JSONs (CLAUDE.md "Trace Template Checklist" steps 3–7).

**Flag-and-ask (minor):**
6. `flashinfer/prims_ts/moe/runner.py` repeats the identical 12-element `routing_out` unpack in 7 places (638/931/1228/1381/1631/1941/2235) — consider one helper; it is also the single choke point for fixing finding 2.
7. PR description overwrites the default template and the perf parity claim ("on par with TRT-LLM Gen") carries no numbers/GPU/shapes — both become the squash-merge commit message; please restore the template and attach benchmark data (guidance §6).

Not re-raised (already fixed on the branch): autotune forced-tactic key mismatch (bba65a93), `_collapse_to_vmk` (main merge 262795b8).

<sub>AI-assisted review — findings verified where stated; a maintainer's judgment supersedes.</sub>

tests/moe/conftest.py
```diff
@@ -4,22 +4,41 @@
...
 if _REPO_ROOT not in sys.path:
     sys.path.insert(0, _REPO_ROOT)
+
+
+@pytest.fixture(autouse=True)
+def _disable_prims_ts_exhaustive_checker_in_moe_integration_tests(monkeypatch):
+    monkeypatch.setenv("FLASHINFER_PRIMS_TS_DEBUG_CHECKS", "0")
+
+
+def pytest_generate_tests(metafunc):
```
aleozlx: **CI budget (high confidence, measured):** this hook re-parametrizes every test naming `moe_gemm_backend` across TRTLLM+PrimsTS. Combined with the PrimsTS-only `weight_processing` entries, `tests/moe/test_trtllm_gen_fused_moe.py` now collects **49,383 nodes** (30,808 PrimsTS + 18,575 TRTLLM vs ~18.5k before), and the internal SM100/103 jobs exceed walltime ("not executed due to timeout" for the whole file). Incompatible combos are filtered by *runtime* `pytest.skip` (`tests/moe/utils.py:194`), so they still inflate collection/scheduling.

Suggestion: keep exhaustive PrimsTS coverage in `tests/prims_ts/` (383 tests, ~18 min on B200) and run the PrimsTS backend through the integration file only for a representative slice — same decomposition #4082 applies to the routing axis; the two reductions compound.

flashinfer/prims_ts/batched_gemm/tmem_sf_resources.py
```diff
@@ -0,0 +1,874 @@
...
+
+    @cute.jit
+    def _cvt_e2m1x2_to_bf16x2_scaled(
+        self,
+        packed_word: Int32,
+        packed_scale_word: Int32,
+        byte_lane: Constexpr[int],
+        scale_lane: Constexpr[int],
+    ) -> Int32:
+        extract = (
```
dbari:
```suggestion
         # CUDA 12.x libNVVM rejects E2M1-to-BF16 PTX for sm_100a. Decode the
         # E2M1 values with integer operations there, while retaining the safe
         # UE8M0 conversion and packed BF16 multiply. CUDA 13+ keeps the fused
         # conversion below.
         if cutlass.target_version(max_version="12.9"):
             packed_byte = (packed_word >> Int32(byte_lane * 8)) & Int32(0xFF)
             scale_byte = (packed_scale_word >> Int32(scale_lane * 8)) & Int32(
                 0xFF
             )
             value = self._decode_cast_pair(packed_byte)
             scale_pair = scale_byte | (scale_byte << Int32(8))
             scale = self._cvt_ue8m0x2_to_bf16x2(scale_pair)
             return self._mul_bf16x2(value, scale)
 
         extract = (
```
@aleozlx this is the fix for point A from https://github.com/flashinfer-ai/flashinfer/pull/4361#issuecomment-5363720598 (MxFP4×BF16 prims_ts kernels fail NVVM compilation for sm_100a)

feih-nv:
@flashinfer-bot run

feih-nv:
Additional findings after checking the existing inline comments, review summaries, and discussion:

1. **Packaging blocker:** `flashinfer/prims_ts/moe/prims_ts_moe_configs.json` is not included in built wheels. `config_mapper.py` requires this file at runtime, while `include-package-data = false` and no package-data entry includes it. Non-editable installs will raise `FileNotFoundError`. Please explicitly package the JSON and add a wheel-install smoke test.

2. **Unpacked routing autotuning uses packed expert IDs:** all Prims-TS runners initialize `topk_ids` with `moe_topk_ids_init(..., packed=True)`. For `UnpackedPrecomputed`, this tensor must contain plain expert IDs; packed BF16 weight bits can produce out-of-range IDs during tuning. The initializer should depend on the routing mode.

3. **Expert-parallel GEMMs use the global expert count:** launch IO passes `kwargs["num_experts"]` into the tensor adapters, although the weight tensors contain `num_local_experts`. This controls the expert dimension of validation/tensor maps and can reject valid sharded weights or address beyond local storage. GEMM IO should use the local count; routing/finalization can continue using the global count.

4. **MajorK weights bypass shape validation:** `_validate_weight_storage()` immediately returns for `WeightLayout.MajorK`. Incorrect rank, expert count, K, or N dimensions can therefore reach the kernel and cause out-of-bounds accesses. MajorK should receive validation equivalent to the BlockMajorK path.

5. **MXFP8 scale validation checks only dtype:** `_check_fp8_scale_storage()` accepts any byte-sized FP8/uint8 tensor without validating its required shape or storage size. Undersized or incorrectly laid-out weight/activation scales can be passed to the kernel. Please validate the exact scale layout for each FC and quantization mode.

6. **The compiled GEMM cache is device-architecture blind:** `get_compiled_gemm()` keys only on `(cfg_hash, fc1_or_fc2, compile_options)`. A process using multiple GPU architectures can reuse code compiled for another device. The compile target/device architecture should be included in the key.

7. **The advertised drop-in FP8 API signatures differ:** `prims_ts_fp8_per_tensor_scale_moe` inserts `weight_layout` before `do_finalize` and additional scale arguments before `routing_replay_out`, while its docstring says it has the same arguments as the TRT-LLM API. Existing positional calls cannot safely swap backends. Please align the signatures or make backend-specific options keyword-only and document the difference.

flashinfer/fused_moe/shared/tuning.py
feih-nv: Addressed in 018aef79. Empty routing placeholders are again included in the dynamic tensor spec via `is not None`, preserving the legacy persisted profile/file-key shape. A regression test asserts the exact legacy `nearest_profile` and `file_key`; the focused tests pass.

flashinfer/prims_ts/moe/runner.py
feih-nv: Addressed in a2ab8fcf. Routed outputs now go through one shared helper that preserves Torch tensors recursively converted by TVM-FFI 0.1.11+, while retaining a reusable zero-copy DLPack fallback. Added regression coverage for tensor identity and repeated conversion; the focused tests pass.

tests/moe/conftest.py
feih-nv: Addressed in 088e841d. The large shared grid now defaults to TRT-LLM instead of globally multiplying every `moe_gemm_backend` consumer; the five PrimsTS-only weight-layout entries were removed from that grid, while a focused MXFP8 bias test still explicitly compares TRTLLM and PrimsTS. Collection dropped from 16,124 tests for the shared file to 6,126 tests for the shared file plus the focused PrimsTS file. Exhaustive PrimsTS coverage remains under `tests/prims_ts/`. Relevant pre-commit hooks and collection checks pass.

flashinfer/prims_ts/moe/runner.py
feih-nv: Additional follow-up in ef8420f8: the shared conversion helper now also preserves the optional `None` expert-weight slot returned for unpacked precomputed routing. This was found by rerunning the full SM100 file after the earlier TVM-FFI fix. The exact routed-mode regression now passes on CUDA 12.9 / SM100 (`1 passed`), and the helper has CPU coverage for Torch tensors, reusable FFI tensors, and optional outputs.

tests/conftest.py
feih-nv: Agreed on treating persistent Prims-TS kernel artifacts as a separate follow-up. This update only makes the existing in-process compiled GEMM cache safe across CUDA devices / CuTe-DSL compile architectures; adopting the shared persistent CuTe-DSL cache needs broader lifecycle and invalidation work, so I am leaving this thread open as a TODO.

feih-nv:
Up to `3acc1952`, all review findings listed above are addressed.

Two follow-ups remain, with different merge implications:

1. **Performance qualification — before merge.** Run an apples-to-apples PrimsTS versus TRT-LLM Gen benchmark on at least B200/SM100a across representative supported BF16, FP8 per-tensor, FP8 block-scale, and FP4 workloads. Each result must record the GPU and software versions, token/expert/top-k and hidden/intermediate shapes, quantization/routing mode, timing method, warmup/iteration count, and median latency. JIT compilation should be excluded from steady-state timing. Publish the table, define the threshold meant by “on par,” and revise or remove that claim if the results do not support it. SM103a/B300 performance belongs with its separate signoff.

2. **Persistent PrimsTS kernel artifact cache — tracked post-merge startup/performance work, not a known correctness blocker.** The current compiled-GEMM cache is only in-process. `ef8420f8` prevents unsafe reuse across devices/architectures, but new processes still compile tactics again. A follow-up should use FlashInfer's shared persistent CuTe-DSL cache lifecycle (`JitSpecCuteDsl` or equivalent), with keys/invalidation covering architecture, CUTLASS-DSL version, source digest, compile options, and tactic/config. It should test cold versus warm processes, invalidation, and heterogeneous devices. This should have a linked issue and owner rather than further expanding this PR.

The PR description now keeps these TODOs together under **Reviewer Notes → Remaining follow-ups** and explicitly marks the first as pre-merge and the second as post-merge tracking work.

feih-nv:
@flashinfer-bot run

feih-nv:
@flashinfer-bot run

feih-nv:
@flashinfer-bot run

feih-nv:
@flashinfer-bot run

docker/Dockerfile.flashinfer-nvep
```diff
@@ -121,7 +121,7 @@ ENV VIRTUAL_ENV=${VENV}
 RUN uv pip install --python ${VENV}/bin/python \
         torch \
         setuptools packaging \
-        'apache-tvm-ffi>=0.1.6,<0.2,!=0.1.8,!=0.1.8.post0' \
+        'apache-tvm-ffi>=0.1.11,<0.2' \
```
aleozlx, line 125: what's the reason for this change?

cc @Anerudhan can you double check if this line is stale and if the new change makes sense

flashinfer/prims_ts/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+
+from .cutlass_dsl import ensure_cutlass_dsl_experimental
+from .utils import (
+    configure_cute_dsl_cache_dir,
+    get_prims_ts_compile_options,
+    is_prims_ts_available,
+)
+
+configure_cute_dsl_cache_dir()
+ensure_cutlass_dsl_experimental()
```
kangbintNV: This module-level `ensure_cutlass_dsl_experimental()` call causes every import flashinfer to import cutlass.experimental.task_scheduling. With CUTLASS DSL 4.7.0, that import globally monkey-patches cutlass.utils.WorkTileInfo to convert each top-level tile coordinate to `Int32`. Modular FMHA passes `(m, 0, (batch, head))` to the same class, so the nested third coordinate becomes `Int32((Int32, Int32))` and fails with `(Int32(?), Int32(?))` to integer conversion is not supported.

This matches [failed Job 415276469](https://nv/flashinfer-ci/-/jobs/415276469), where all 188 runnable cases in `test_modular_fmha_prefill.py` failed, while the identical batch passed with 188 PASS and 20 SKIP in [base Job 415316375](https://nv/flashinfer-ci/-/jobs/415316375). Could we remove this module-level initialization, use a side-effect-free availability check such as `find_spec`, and lazily import the Prims-TS backend? The underlying CUTLASS monkey-patch should also be isolated or updated to support nested `cute.Coord` for processes that use both Prims-TS and FMHA.

docker/Dockerfile.flashinfer-nvep
dbari: This helps solve `RuntimeError: from_dlpack received an invalid capsule. Note that DLTensor capsules can be consumed only once`. Apache TVM-FFI added recursive framework conversion for tensors nested in Array/List/Map/Dict in v0.1.11. Older versions can expose nested outputs as cached one-shot PyCapsules, while v0.1.11+ gives each consumer a reusable tensor object.

feih-nv:
@flashinfer-bot run

feih-nv:
@coderabbitai review

feih-nv:
@flashinfer-bot run

aleozlx:
/bot run tests/prims_ts/test_batched_gemm_mx.py tests/gemm/test_mm_fp4_sm103.py tests/attention/test_modular_fmha_prefill.py tests/attention/test_modular_fmha_prefill_paged.py

aleozlx:
/bot run tests/prims_ts/test_batched_gemm_mx.py tests/gemm/test_mm_fp4_sm103.py tests/attention/test_modular_fmha_prefill.py tests/attention/test_modular_fmha_prefill_paged.py

feih-nv:
/bot run tests/prims_ts/test_batched_gemm_mx.py tests/gemm/test_mm_fp4_sm103.py tests/attention/test_modular_fmha_prefill.py tests/attention/test_modular_fmha_prefill_paged.py

feih-nv:
@flashinfer-bot run

feih-nv:
@flashinfer-bot run

dbari:
@flashinfer-bot run

rosenrodt:
@flashinfer-bot run

feih-nv:
@flashinfer-bot run

feih-nv:
/bot run tests/moe/test_trtllm_moe_valid_dims.py tests/moe/test_trtllm_gen_fused_moe.py tests/moe/test_trtllm_gen_routed_fused_moe.py tests/mamba/test_cake_ssd_combined.py

feih-nv:
/bot run tests/moe/test_trtllm_moe_valid_dims.py tests/moe/test_trtllm_gen_fused_moe.py tests/moe/test_trtllm_gen_routed_fused_moe.py tests/mamba/test_cake_ssd_combined.py

feih-nv:
@flashinfer-bot run

feih-nv (approved):
LGTM, let's address the TODOs in the description which should not change the code.

feih-nv:
@flashinfer-bot run

bkryu (approved):

feih-nv:
@flashinfer-bot run

PR #4373, opened 2026-08-06
feat: support FP8 KV NoPE MLA on SM90

JustinTong0323:
## Description

Allow `BatchMLAPagedAttentionWrapper` to run with `head_dim_kpe=0`, including FP8 KV caches on SM90.

When the positional-key-embedding dimension is zero, the KPE MMA loop is empty. The FA2 and FA3 paths initialize the QK accumulator from compressed KV instead. For FP8 KV, both native backends repack to BF16 and can consume an optional contiguous FP32 scale tensor with one scale per physical KV token and 128 CKV channels.

This change:

- supports zero-width KPE in FA2 and FA3;
- supports FP8 E4M3 KV with BF16 queries on SM90 for `head_dim_ckv=512` and `head_dim_kpe` in `{0, 64}`;
- adds optional `ckv_scale_arr` with shape `ckv_cache.shape[:-1] + (4,)` for per-token, per-group CKV dequantization;
- preserves existing per-tensor `ckv_scale` / `kpe_scale` behavior;
- validates backend, device, dtype, dimensions, scale shape, dtype, device, and contiguity;
- adds paged-MLA correctness coverage with a non-identity page table for FA2 and FA3.

## Related Issues

None.

## Pull Request Checklist

- [x] Pre-commit hooks pass on the changed files.
- [x] GPU tests pass.

## Tests

NVIDIA H200, CUDA 12.8, PyTorch 2.11.0+cu128:

- `test_batch_mla_without_kpe`: FA2 and FA3 passed.
- `test_batch_mla_fp8_nope_group_scales_matches_bf16_reference`: FA2 and FA3 passed.
- `test_fp8_kv_scales_are_keyword_only` and `test_fp8_kv_requires_scales`: passed.
- Representative existing KPE=64 per-tensor FP8 and row-aliasing regressions: passed.

## Reviewer Notes

- `ckv_scale_arr` follows the physical paged-cache layout used by the downstream SGLang integration: `[num_pages, page_size, head_dim_ckv / 128]`.
- Existing BF16/FP16 and FP8 per-tensor scale behavior is unchanged.
- The latest external attention CI failures before this update were not stable across reruns: HCA failures were classified as pre-existing, while the new failures alternated between GB200 multi-node NCCL setup and RTX Pro 6000 `cudaErrorUnknown` failures.

## Summary by CodeRabbit

* **New Features**
  * Added FP8 batched paged attention with optional per-token/per-channel CKV scaling.
  * Added support for FA2 and FA3 FP8 execution.
  * Expanded supported configurations to include zero- or 64-dimensional key positional encoding.

* **Bug Fixes**
  * Fixed attention computation without key positional encoding dimensions.
  * Added validation for incompatible scale inputs, shapes, layouts, and configurations.

* **Tests**
  * Added coverage for grouped CKV scaling and trace inputs.
  * Added BF16 reference comparisons across supported implementations.


JustinTong0323:
I added the requested SM90 FP8 KV NoPE support, including optional per-token/per-128-channel CKV scales for both FA2 and FA3, and rebased the branch onto the latest main. On an H200 with CUDA 12.8 / PyTorch 2.11.0+cu128, the new non-identity-page-table group-scale test passes on both FA2 and FA3; the existing KPE=64 per-tensor FP8 and row-aliasing regressions also pass. The current required check is red only because CI was skipped pending authorization for the new commit (`b72a3c14`); could you please run `@flashinfer-bot run` (and `/bot run tests/attention` if needed)? The prior external failures were not stable across reruns: HCA was classified as old, while the new failures alternated between GB200 NCCL setup and RTX Pro 6000 `cudaErrorUnknown`, so they do not appear attributable to this MLA change.

flashinfer/mla/_core.py
```diff
@@ -2321,6 +2322,7 @@ def run(
         *,
         ckv_scale: Optional[float] = None,
         kpe_scale: Optional[float] = None,
+        ckv_scale_arr: Optional[torch.Tensor] = None,
```
saltyminty, line 2324: Non-blocking but just want to hear thoughts: the intention is that we will accept only one of either `ckv_scale` or `ckv_scale_arr`, correct?

- Since these arguments are already keyword only, we are free to reorder them (i.e. putting `ckv_scale_arr` directly after `ckv_scale`)
- would it be cleaner to only keep one argument as `ckv_scale: Union[float, torch.Tensor]`?

saltyminty (approved):
Left one non-blocking comment.

JustinTong0323:
Regarding the public API advisory above: this change is backward-compatible. `ckv_scale_arr` is an optional keyword-only parameter defaulting to `None`. Existing per-tensor callers continue to use `ckv_scale=...` and `kpe_scale=...` unchanged, while grouped-scale callers explicitly opt in with `ckv_scale_arr=...`. No existing parameter is removed, renamed, or narrowed, so no deprecation or migration path is needed.

flashinfer/mla/_core.py
JustinTong0323: Yes, they are intended to be mutually exclusive. I reordered `ckv_scale_arr` directly after `ckv_scale` and now require exactly one of them for FP8 KV.

I kept them as separate arguments because the trace schema represents scalar and tensor inputs with different descriptors. A single `Union[float, torch.Tensor]` parameter cannot faithfully describe both forms in the generated benchmark definition. Internally, the grouped-scale path still passes 1.0 through the unused scalar launcher slot, so the C++ ABI remains unchanged.

leejnau:
@flashinfer-bot run

saltyminty (approved):
Supposed API breaking change is in a keyword only section and is not an issue.

PR #4387, opened 2026-08-07
feat(moe_ep): SM120 MXFP8 swap-AB CuTeDSL MegaMoE kernel

mhoqueanik:
# feat(moe_ep): SM120 MXFP8 swap-AB CuTeDSL MegaMoE kernel

## Summary

Adds an **SM120 (sm_120/sm_121) MXFP8 MegaMoE backend** to `flashinfer.moe_ep` — the swap-AB CuTeDSL megakernel (fused dispatch + grouped GEMM + combine) for the RTX PRO 6000 / GB10 (DGX Spark) class of Blackwell parts. Registered as `sm120_mxfp8_mxfp8_bf16_cutedsl` behind the standard `MegaKernelBackend` contract, on the same taxonomy layout as the SM100/SM90 backends.

> **Status: functional correctness is validated; performance tuning and benchmarking are ongoing.** The backend currently takes a knob dict only (no autotune yet), and no performance numbers are quoted in this PR — perf tables will follow once the tuning sweep lands.

## What's included

- **Vendored kernel drop**: `kernel_src/sm120/` — verbatim snapshot of the four kernel packages (`common/`, `src/`, `moe_sm120_mxfp8_swapab/`, `moe_mxfp8_glu/`) from the `sm120_swapab_wt` fork at `d19d30a` (branch `run/sm120-mxfp8-perf`), including two uncommitted worktree edits recorded in `VENDOR.md`. `src/` stays byte-for-byte upstream; all adaptation lives in the shim.
- **Shim layer**: mirrors the sm90 fork-tree conventions — `_paths` sibling-tree guard, per-tree `comm.py` (plus `zero_local_counter_regions`; this drop's kernel does not tail-clean its local counters), all-lazy `kernel_helpers`, and the SM120 MXFP8 frontend with `combine_output` + topk-reduce second stage, per-expert epilogue args, K-major weight views, and a mirrored-ABI-constant guard.
- **Backend**: `backends/mega/kernel/sm120/mxfp8_mxfp8_bf16_cutedsl/` with `Sm120_Mxfp8_Mxfp8_Bf16_Cutedsl_MegaMoeConfig` (native `token_back_mode` enum), K-major + interleave-8 weight preprocessing, torch-composed staging (no fused DataPreprocess in this tree), `validate_mega_arch_sm120` (exact sm_120/sm_121 family), and the sm120 runtime-requirements alias. Exported from `flashinfer.moe_ep`.
- **Rank-sharing bootstrap for single-GPU sm_12x boxes**: sm_12x cluster nodes have one GPU, and the drop's own harness runs N ranks per GPU. The FI runtime now folds `LOCAL_RANK` onto the physical GPUs (identity when there are enough), and under `MEGA_SINGLE_GPU_GLOO=1` inits the process group as gloo with a CPU-side NVSHMEM UID broadcast (NCCL cannot host two ranks on one device).
- **Tests**: sm120 mxfp8 mega multirank suite mirroring the sm100 one (layer-vs-shim parity across staged/prestaged/token-back profiles, all-gather torch-oracle anchor, registry/preprocess checks), a shared arch-free term-magnitude oracle-compare module, new `arch_sm120` pytest marker (with `arch_blackwell` tightened to the sm_10x family so sm100 suites stop collecting on sm_11x/12x hosts), and a `mega_sm120` entry in `run_tests.sh`.

## Upstream gaps found and guarded

Each was pinned by A/B against the drop's own runner and is recorded in `VENDOR.md` as a pending-upstream item; the FI backend rejects the broken configuration rather than silently running it:

| gap | disposition |
|---|---|
| `in_kernel_fc2_reduce` crashes with `cudaErrorIllegalAddress` (verified RTX PRO 6000, 4 ranks / 1 GPU, DSL 4.6.1); absent from the kernel team's own test scripts | rejected by the backend; ikr tests skipped with documented reason; shim keeps the plumbing for a fixed drop |
| `cluster_m > 1` fails `cute.compile` ("expects num_multicast to be 1 for non multicast G2S copies"); upstream scripts always use (1,1,1) | config pinned to cluster (1,1,1) |
| `gate_up_clamp` is dead plumbing — `kernel_fc12` stores it but never reads it; output bit-identical with/without (invisible to the drop's ±0.5-sparse test data) | backend rejects a set clamp; tests run clampless |
| `world_size=1` (`MEGA_NO_DIST`) numerics silently wrong for `mma_tiler` N=128 (5-20% of cells; reproduced upstream at their own standard geometry); same tile bit-exact at world_size=4 | documented in VENDOR.md; multirank is the validated path |

Two harness-level fixes that the correctness claims depend on: the oracle's `all_gather` is CPU-staged under the rank-sharing gloo group (gloo has no CUDA all_gather — the reference was previously built from corrupted operands), and the shim's `_main` smoke device folds onto physical GPUs.

## Usage

```python
from flashinfer.moe_ep import (
    BootstrapConfig, FleetParams, MegaConfig, MoEEpLayer, MoEEpTensors, MoEWeightPack,
    Sm120_Mxfp8_Mxfp8_Bf16_Cutedsl_MegaMoeConfig,
)

layer = MoEEpLayer(
    BootstrapConfig(world_size=world_size, rank=rank),
    FleetParams(num_experts=256, max_tokens_per_rank=tokens, token_hidden_size=7168),
    weights=MoEWeightPack(w13=w13_bf16, w2=w2_bf16),
    backend=MegaConfig(
        megakernel=Sm120_Mxfp8_Mxfp8_Bf16_Cutedsl_MegaMoeConfig(intermediate_size=2048, top_k=8),
    ),
)
out = layer.forward(MoEEpTensors(hidden_states=x_bf16, topk_ids=ids, topk_weights=w))
```

On a single-GPU sm_12x box, run multirank via the drop's env: `MEGA_SINGLE_GPU_GLOO=1 torchrun --nproc_per_node=4 ...`.

## Constraints

- Exact sm_120 / sm_121 family only (`validate_mega_arch_sm120`).
- `in_kernel_fc2_reduce`, `cluster != (1,1,1)`, and `gate_up_clamp` are rejected until the upstream drop fixes them (see table above).
- Validated against nvidia-cutlass-dsl 4.6.1.
- No autotuner yet — kernel knobs are passed as an explicit dict.

## Testing

- sm120 mxfp8 mega multirank suite green on RTX PRO 6000 (4 ranks / 1 GPU via the rank-sharing gloo bootstrap), DSL 4.6.1: layer-vs-direct-shim parity across staged / prestaged / large-token dispatch token-back profiles, plus the all-gather torch-oracle anchor (the sm120 reference pins `gate_up_interleave=8` + `apply_topk_in_fc1=True` itself).
- Registry, config-validation, and weight-preprocess checks.
- `run_tests.sh mega_sm120` (own torchrun process; kernel trees are process-exclusive).

## Performance

**Ongoing.** Kernel-level tuning (tiler/knob sweeps) and the standalone microbenchmark/e2e runs are in progress; numbers will be posted to this PR (or a follow-up) once the tuning sweep completes. Nothing in this PR should be read as a perf claim yet.

## Summary by CodeRabbit

* **New Features**
  * Added SM120 MXFP8 MegaMoE support, including fused execution, quantized inputs and weights, token communication, top-k reduction, and single- or multi-rank operation.
  * Added offline and collective autotuning for NVFP4 and MXFP8 workloads.
  * Added architecture-specific backend configurations and runtime validation for SM90, SM100, and SM120 hardware.

* **Improvements**
  * Introduced clearer backend and kernel names while preserving deprecated aliases for compatibility.
  * Improved workspace cleanup, distributed initialization, CUDA graph safeguards, and input validation.

* **Documentation**
  * Expanded architecture, tuning, compatibility, testing, and vendored-source guidance.


mhoqueanik:
**`valid_ab_tuple` NameError:**

> Fixed — the error path referenced an undefined `valid_ab_tuple` (masking the intended `ValueError` with a `NameError`); re-aligned with the reference copy in `kernel_src/cutedsl_megamoe`, which uses `valid_ab`.

**`_amax_lane` divergent `warp_redux_sync`:**

> Verified the finding is technically correct: both calls default to the full-warp membermask, so the `not first_half` call is a divergent full-mask `redux.sync` — UB. It is unreachable in this PR: the divergent path requires `sf_vec_size != 32`, the only wired sm120 backend is MXFP8 (`sf_vec_size = 32`, and `kernel_fc12.py` hard-rejects anything else), and the nvfp4 kernel that imports this epilogue is dormant (no shim entry point, no backend, no test). The sm90/sm100 copies of this file already carry the legal branchless half-mask form — the sm120 copy is a stale pre-fix snapshot, and it will pick up the same fix at the next vendored re-sync rather than being hand-edited (verbatim-drop policy).

**`topk_reduce` 16-byte alignment guard:**

> The kernel's inputs are internally allocated torch buffers (256-byte-aligned by the caching allocator); no shipping path passes a non-owning slice, so the stride checks cover the real invariants. Noting the alignment assumption is documented by `assumed_align=16` at the copy site; a host-side guard would belong upstream with the rest of the vendored validation.

**pyproject excludes for sm120 vendored src:**

> Already in place at head: `kernel_src/sm120/swapab_cutedsl_megakernel/src/` is in the mypy `exclude` (pyproject.toml:126) and ruff `extend-exclude` (:151).

**config compat aliases:**

> Already addressed at the package level: `Sm90PushFp8MegaMoeConfig` / `Sm90PullFp8MegaMoeConfig` are defined and exported from `flashinfer/moe_ep/__init__.py`; the deep per-backend module paths were never public API.

**VENDOR.md immutable SHA:**

> The file already documents this honestly: the pre-VENDOR.md drop's upstream SHA was not recorded and cannot be retroactively produced; the two partially-resynced files are pinned (`50117315d`), and VENDOR.md mandates a full SHA pin at the next complete re-sync. That is the remediation plan.

Anerudhan (approved):

mhoqueanik:
@flashinfer-bot run

PR #4393, opened 2026-08-07
[feat]custom all reduce kernel

qsang-nv:
# Add a custom all-reduce for intra-node PCIe machines without NVLink

## 📌 Description

Adds `pcie_ipc_comm`: a CUDA-IPC all-reduce for machines where every peer
transfer crosses the CPU root complex — no NVLink, no multicast.

**Why not one of the three all-reduce backends already here.** At 4 and 8 ranks
`vllm_custom_all_reduce` launches nothing at all without NVLink: every arm of
its dispatch except `world_size == 2` sits inside `if (full_nvlink_)`
(`include/flashinfer/comm/vllm_custom_all_reduce.cuh:477-486`), so on this
fabric the call falls through and returns. `trtllm_mnnvl_ar` needs multi-node
NVLink; `trtllm_ar` assumes an NVLink or multicast fabric. What is left is NCCL,
and NCCL leaves a large amount on the table at these sizes — hence the baseline
below.

**Why the kernels are shaped this way.** On such a fabric all-to-all peer writes
collapse to a fraction of what the same kernel achieves writing to a single
destination, and the collapse worsens the more blocks write at once. So the
kernels **stage** their pushes: at any instant each rank has one outbound and
one inbound stream. Total bytes are unchanged, only the order. The 8-rank path
keeps a 4+4 island decomposition so the scarce cross-socket links carry the
minimum. Copy-engine transfers are only ~24% faster than kernel writes here, so
the data plane stays in the kernel rather than moving to DMA.

## 🚀 Usage

Every method below is **collective**: every rank calls it, with identical
arguments, in the same order — including `destroy()`.

```python
import flashinfer.comm as comm

ws = comm.PcieIpcAllReduceWorkspace(
    group=tp_group,                          # see below
    max_numel=max_batch_tokens * hidden,     # capacity, fixed at construction
    dtype=torch.bfloat16,
)
try:
    out = ws.all_reduce(x) if ws.supports(x) else fallback(x)
finally:
    ws.destroy()
```

**`group`** is the tensor-parallel group — the ranks that participate in this
all-reduce, one process per GPU. With TP alone that is `dist.group.WORLD`; under
TP×PP or TP×DP it is the TP subgroup from whatever builds your parallel state.
Its size must be 2, 4 or 8, and it must not span hosts: CUDA IPC cannot cross a
node, so a multi-host group raises during construction rather than at the first
collective.

**`max_numel`** is the largest element count that will ever be reduced through
this workspace — `max_batch_tokens * hidden` for a decode-shaped workload. The
slab is IPC-shared once and its peer offsets are baked into the handle, so it
cannot grow; a larger tensor makes `supports()` answer `False` and the caller
falls back, rather than failing. Size it to the real workload rather than
rounding up: the epoch double buffer places its two halves `world_size *
max_numel` elements apart, so rounding up 4× at 8 ranks moves them 32× the
payload apart and costs measurable time at small batch.

**`supports()`** is a capability question — world size, dtype, capacity, and
enough payload for every rank to own a share — and what routes the rest
elsewhere. It is a pure function of shape and dtype, so every rank answers the
same without negotiating — which matters because a rank that opted out while
its peers opted in would hang them. It is not a claim that the shape is fast
here.

Launch configurations start from a seed: one crossover, keyed on the payload in
bytes rather than on a batch count, so it scales with the shape instead of being
fitted to one. Its constants are still defaults, not results — a workspace that
finds nothing tuned for it says so once. To measure the real ones, tune once —
the result is written to `FLASHINFER_AUTOTUNE_DIR` (or the workspace dir) and
later processes pick it up when they build a workspace with the same
`max_numel`, `max_blocks`, world size and fabric:

```python
ws.tune([hidden])                # the hidden sizes this job will run
```

If `group` is a strict subgroup of the default process group — the TP×PP and
TP×DP case above — pass the reduction group explicitly:
`ws.tune([hidden], tune_group=gloo_tp_group)`. `tune()` builds one itself only
when the workspace spans the whole default group, because `new_group()` is
collective over that group and building one here would hang a job whose
workspace is narrower; it raises rather than guessing.

The library's usual idiom does the same thing, and needs the timing reduction
installed so the ranks agree on a winner:

```python
set_autotune_process_group(gloo_group)
with flashinfer.autotune(True, cache=path):
    for batch in batches:
        ws.all_reduce(sample(batch))
```

**CUDA graphs.** Resolving a launch configuration makes the ranks agree on it
with a small reduction whose verdict is read back on the host, so the *first*
call at a given shape cannot happen inside a capture. Resolve them up front and
the capture touches no collective at all:

```python
ws.tune([hidden])                                   # or load a cache written earlier
ws.prepare(                                         # every shape you will capture
    [(b, hidden) for b in capture_buckets],
    dtype=torch.bfloat16,                           # once per dtype: the cache is keyed by it
)
# ... the framework captures its graphs as usual
```

`prepare()` is collective, must come **after** `tune()` — which clears the
in-process cache — and takes the *padded* batch sizes, since those are what a
serving framework captures. The resolution cache is keyed by
`(numel, hidden, dtype)`, so call it once per dtype the job will run; the
default is bfloat16, and preparing that alone leaves an fp16 capture failing
exactly as if nothing had been prepared. A shape or dtype left out is still
resolved lazily and still cannot be captured; capturing one raises an error
naming `prepare()`, rather than the CUDA-level "Cannot copy between CPU and CUDA
tensors during CUDA graph capture", which names neither this workspace nor the
remedy. Passing `config=` explicitly at the call site is the other way to keep a
capture free of resolution.

## 🔍 Results

8× L40S, bf16, clocks pinned. Median of three tuning sessions; within a
session, the median over timed replays of the per-iteration group maximum, 20
iterations captured per graph. Baseline is NCCL because that is what the caller
falls back to for a shape this reports unsupported. Every world size is measured
at every hidden size, which an earlier revision could not do — its tables
admitted TP2 ≤ 2048, TP4 == 4096 and TP8 ≥ 6144, and six of these nine
combinations were refused rather than slow. All 72 points beat NCCL, from 1.22×
to 5.92×.

These rows use configurations measured on this machine. An untuned machine runs
the seed instead: measured on the diagonal (TP2/2048, TP4/4096, TP8/6144), it is
15% slower at the TP4/TP8 median and 2.0× at its worst point, and within 1% at
TP2 — the price of not shipping one machine's constants as if they were
general.

| batch | TP2 h2048 | TP4 h2048 | TP8 h2048 | TP2 h4096 | TP4 h4096 | TP8 h4096 | TP2 h6144 | TP4 h6144 | TP8 h6144 |
|---|---|---|---|---|---|---|---|---|---|
| 1 | **3.86×** | **2.23×** | **2.69×** | **3.67×** | **1.70×** | **1.40×** | **3.56×** | **2.06×** | **1.87×** |
| 2 | **3.63×** | **1.62×** | **1.47×** | **3.29×** | **1.95×** | **1.33×** | **3.63×** | **1.56×** | **1.46×** |
| 4 | **3.75×** | **1.91×** | **1.37×** | **3.49×** | **1.46×** | **1.53×** | **3.43×** | **1.22×** | **1.23×** |
| 8 | **3.51×** | **1.47×** | **1.51×** | **3.26×** | **1.36×** | **1.41×** | **3.02×** | **1.62×** | **1.63×** |
| 16 | **3.23×** | **1.33×** | **1.44×** | **2.93×** | **1.90×** | **2.06×** | **2.73×** | **2.25×** | **2.59×** |
| 32 | **2.79×** | **1.88×** | **2.01×** | **2.60×** | **2.63×** | **3.04×** | **2.60×** | **3.08×** | **3.41×** |
| 64 | **2.52×** | **2.59×** | **2.92×** | **2.24×** | **3.24×** | **4.30×** | **1.88×** | **2.75×** | **3.24×** |
| 128 | **2.25×** | **3.27×** | **4.33×** | **1.71×** | **2.96×** | **3.66×** | **1.47×** | **3.47×** | **5.92×** |

```bash
sudo nvidia-smi -lgc 2520,2520          # the benchmark does not pin clocks
for n in 2 4 8; do                      # one world size per run, no subgroups
    for h in 2048 4096 6144; do
        torchrun --standalone --nproc_per_node=$n \
            benchmarks/comm/bench_pcie_ipc_all_reduce.py --hidden $h --tune \
            --tune-cache tune_tp${n}_h$h.json --json bench_tp${n}_h$h.json
    done
done
```

`--protocol-ab` rebuilds the kernels with one synchronisation mechanism compiled
out and times both in the same process (A-B-A, reporting the A/A spread as a
noise floor). The mechanisms with such a control — the call-level epoch, the
cross-island double buffer, the entry barriers — are within noise. Fixes without
a control group are not individually claimed to be free.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## 🧩 Design notes

**A seed, not a fitted table.** Admission is a capability check — world size in
{2,4,8}, whole 16-byte packs, at least one pack per rank — with no performance
judgement in it, so hidden sizes are unrestricted; the earlier tables admitted
TP2 ≤ 2048, TP4 == 4096 and TP8 ≥ 6144, which were the shapes they had been
fitted at rather than the shapes the kernels can run. What remains in the
library is one crossover keyed on payload bytes — push straight to every peer
while the payload is small, reduce-scatter/all-gather once it is not — because
that is a ratio of bytes to barrier latency, which ports between fabrics, unlike
a token count. The interconnect is still probed: `resolve_pcie_ipc_profile()`
works per GPU *pair*, keys by GPU UUID so `CUDA_VISIBLE_DEVICES` cannot skew it,
and decides with a pure function before any allocation — it now keeps two
topologies out of each other's autotune cache instead of selecting a table.

**Autotuning is where the numbers come from.**
`workspace.tune(hiddens)` measures every launchable configuration and persists
the winner; the library's `with autotune(True, cache=path)` idiom works too.
Three things make it safe for a collective that spins without a timeout:
candidates are **screened for correctness before they are timed**, with the
verdict reduced across the group (the autotuner ranks purely on wall time, and
this protocol's characteristic failure — a sentinel poll returning stale data —
is wrong *and* fast); a search **declines unless the timing reduction is
installed on every rank**, since tuning mode is process-global and a caller
tuning its GEMMs would otherwise sweep this operator with the ranks free to
disagree on a kernel; and **which shapes are supported does not change**,
because admission is a capability question and tuning only ranks configurations
for a shape already admitted. The seed stays the reference candidate every
measurement is ranked against, and the fallback for a stale or disagreeing
cache.

`TuningConfig` sets `use_cuda_graph=True`, and that is not a preference: without
capture the profiler issues each iteration separately, the host cannot keep up
with a few-microsecond collective, and the span it times is dominated by launch
gaps. Measured at 2 ranks batch 8, the ungraphed search picked a configuration
29% slower than the reference candidate; capture picked one within 3%.

**What tuning is worth: 15% at the TP4/TP8 median against the seed**, 2.0× at
the worst point. The tail is where the seed's coarseness shows: at TP8 batch 2 a
search finds `FLAT_STAGED(1,64)` and the seed asks for the one-shot push, which
is the narrow window a per-batch table used to encode by hand. TP2 is within 1%
of tuned at every batch, which is the one place the seed's single branch is
provably enough. Against configurations already measured on the same machine the
headroom
is far smaller — TP4 11% at the median, TP2 at the link ceiling, TP8 inside the
search's own repeatability. With ~190 candidates measured once each and the
leaders within a few percent, `argmin` selects the most favourable measurement
error rather than the best configuration.

The three sessions above separate that from harness noise, because they record
which configuration each search picked. Where all three converged on the same
one the spread is 0.3–3%; where they picked different ones it tracks the
configurations, worst at TP8 batch 1 — `UNSTAGED(128,1024)`, `(96,512)` and
`(32,1024)` on successive runs, 21.8/28.9/18.7 µs, a 47% spread. So the
measurement is repeatable and the *selection* is not, concentrated where fixed
costs dominate and many candidates sit within noise of each other. Fixing it
needs a second pass that re-measures the leaders, which the library's autotuner
has no stage for; until then a tuned result is worth re-checking at small batch.

**A scratch region holds one protocol family, not one kernel.** *Sentinel*
kernels poll for +0.0 meaning "not yet written" and store it back once a poll
succeeds; *barrier* kernels are content-blind and leave raw payload behind.
Nothing else sweeps the workspace — the host zeroes it once at init. So a region
may hold barrier kernels, or sentinel kernels that restore what the others poll,
but not both: a sentinel kernel landing on a barrier kernel's leftovers has its
all-gather poll return them immediately — wrong output on a subset of ranks, not
a hang. The epoch double buffer does not substitute for this; it guarantees the
other half is quiescent, not that it is clean.

**The scratch epoch is per call, not per block.** Per-block parity counts how
many times *that block* has run, so one change in grid size — which an ordinary
batch change makes — desynchronises the block ranges permanently. The epoch is
one value per launch, published by whichever block arrives last at kernel entry,
one counter per scratch region.

**Pack ownership below one pack per rank.** The reduce-scatter split hands each
rank `num_packs / world_size` packs, which is zero once the payload has fewer
16-byte packs than ranks; there `rsag_owner_for_pack()` named rank 0 while the
chunk ranges the same kernels walk gave the whole payload to the last rank, so a
kernel wrote to one owner and polled another. Sentinel loops have no timeout, so
that is a hang, not a wrong answer. It was unreachable while the
hidden-size guards held every admitted shape far above that floor, and it stays
unreachable through admission — the one-pack-per-rank rule is exactly that
floor, made explicit. The formulas now agree on the last rank regardless, so the
explicit-config path is correct too; covered by
`test_pcie_ipc_tiny_payload_every_variant`, which fails by hanging without it.

**Barrier generations compare in modular ring order.** `flag = load(slot) + 1`
overflows after ~2³¹ calls per block and a plain `<` then lets every barrier
through at the wrap. Counters increment as `uint32_t`; the spin compares
`(observed - expected) >= 0` reinterpreted as signed.

**Collective discipline.** `supports()` is a pure function of shape and dtype and
never reads rank-local state, so ranks cannot disagree about opting in; a device
mismatch *raises* rather than answering `False`, which would mean "use another
backend". Construction is a staged transaction — a rank that finds a problem
records an outcome and the group fails together at the next gather. The
workspace is caller-owned via the existing `create_shared_buffer()` /
`free_shared_buffer()`, because teardown needs a collective barrier between
"every rank unmaps its peers" and "every rank frees its slab", which a
destructor cannot express.

## ✅ Testing

Multi-GPU cases compare against NCCL at **zero tolerance** using small-integer
inputs, so the differing reduction order cannot mask an error.

- `tests/comm/test_pcie_ipc_all_reduce.py` — **30 passed** (8 GPUs). World sizes
  2/4/8 × bf16/fp16, all nine dispatchable (world size, variant) pairs,
  CUDA-graph capture and replay — including that a shape must be resolved
  before it can be captured: an unprepared one must raise an error naming
  `prepare()`, and a prepared one must capture, replay and still match NCCL.
  The older capture test warmed up three times before capturing, so it had been
  passing *because* of that warm-up; the dependency is now explicit rather than
  incidental,
  unsupported-shape and device/stream rejection, interleaved shapes with no
  intervening synchronisation, a grid-size change between calls, payloads with
  fewer 16-byte packs than ranks on every variant, and every TP8 variant
  interleaved on one workspace. That last one fails with ~3000 wrong
  elements if the island-agnostic kernel is moved into the barrier kernels'
  region — and its call order is load-bearing, because alternating the two
  families locks each onto one epoch half and the test then passes wherever the
  kernel is put. Six cover tuning: identical launch counts across ranks during a
  search, a corrupted candidate excluded from the timed set, a tuned
  configuration persisted and reloaded by a second process with every rank
  resolving the same kernel, the standard `autotune(True)` idiom, a tuning
  session that forgot the reduction group leaving the seed alone rather than
  hanging, and the untuned warning — which has to fire once rather than per
  call, stay silent while `autotune(True)` is open since that path skips the hot
  cache, and distinguish a machine nobody tuned from a cache written for a
  differently sized workspace.
- `tests/comm/test_pcie_ipc_cross_island_race.py` — **3 passed**, opt-in behind
  `FLASHINFER_TEST_PCIE_IPC_RACE=1`. Negative controls: each rebuilds a broken
  protocol through a compile-time switch and asserts the same sequence fails, so
  a passing test cannot be confused with one that never opened the window.
- `tests/comm/test_pcie_ipc_tuning.py` — **23 passed, 1 xfail, no GPU**. Tactics
  survive the JSON round-trip the cache needs; the candidate grid can name every
  configuration a search actually selected (block counts of 12 and 96 among
  them, so a powers-of-two grid would quietly cap the operator); every way a
  cached tactic can be stale ends at the seed rather than an exception; the
  correctness verdict reduces with the operator that makes it a group decision.
  Three synthesise a cache file and pin the coverage check: a workspace whose
  `max_numel`, world size, `max_blocks` or fabric differs from the tuned one
  reads as uncovered, an empty cache covers nothing, and a cache holding only
  the other 2-byte dtype still counts as covering. The strict xfail records the
  one remaining gap: the grid has no 3, and the seed asks for three ring blocks
  between 768 KiB and 1 MiB.
- `tests/comm/test_pcie_ipc_policy.py` — **40 passed, no GPU**. Admission is
  pinned exactly in both directions: a false yes reaches a hard check
  mid-collective, a false no routes a supported shape away. Of the seed only the
  shape is asserted — its constants belong to whatever machine measured them —
  and everything it returns passes the kernel's hard checks. One runs the other
  direction: every variant the dispatch can launch must be reachable from the
  tuner's candidates, because a kernel nothing can select is either dead code or
  an unexplored corner of the launch space, and from outside those look the same.
- `generation_reached()` is `constexpr` with `static_assert`s pinning the
  `INT32_MAX → INT32_MIN` and `UINT32_MAX → 0` boundaries; replacing it with a
  signed or an unsigned `<` fails the build.
- `pre-commit run --all-files` — clean.

**Only the policy tests and the static_asserts run in CI.** The multi-GPU cases
are run by hand on an 8-GPU box. The pytest entry added to
`scripts/task_test_single_node_comm_kernels.sh` puts them where the other comm
tests live, but no workflow invokes that script and it defaults to
`CUDA_VISIBLE_DEVICES=0`; the script now prints a loud banner when too few GPUs
are visible, because a run where every case silently skips still exits 0.
Wiring up an 8-GPU runner is a maintainer decision.

## ⚠️ Known limitations

- **An untuned machine runs the seed**, which is a workable configuration and
  not a fast one — 15% off tuned at the TP4/TP8 median on the machine measured
  here, 2.0× at the worst point, within 1% at TP2. `tune()` is one step per
  machine and world size, and the workspace warns once until it has run.
- **The tune cache is keyed on the workspace, not just the shape.** `max_numel`,
  `max_blocks`, world size and the fabric label are all in the key — `max_numel`
  because the epoch double buffer places its halves `world_size * max_numel`
  apart, so the best block count genuinely depends on it. A workspace built with
  different values therefore misses *every* entry rather than a few and runs the
  seed throughout. That is indistinguishable from a hit at any single shape, so
  it is detected against the loaded keys at construction and warned about on
  first use; the benchmark also marks rows that found no entry. Re-tune when the
  workspace parameters change.
- **float32 and `enable_pdl` are rejected.** The launcher hard-checks a 2-byte
  element size and instantiates only `half` and `nv_bfloat16`; the kernel
  templates carry a generic path, but nothing wider has been built or measured.
  PDL is refused because the TP8 block kernel triggers launch completion before
  two protocol stores; re-enabling needs a per-kernel audit and an SM90
  regression, and PDL does not exist on the hardware this targets.
- **A shape must be resolved before it can be captured.** The agreement that
  settles a launch configuration is a collective read back on the host, so it
  cannot run inside a CUDA graph capture. `prepare()` moves it to a point the
  caller chooses; it does not remove it. Nothing else about capture changed —
  replays carry no baked-in state, since the protocol state lives in device
  memory and is read at kernel entry.
- **No timeout.** Every wait is an unbounded spin, so ranks that disagree on
  shape, dtype or call order hang rather than raise.

## 📊 Notes for reviewers

- `include/flashinfer/comm/pcie_ipc_all_reduce.cuh` is the bulk of the diff. The
  signal-region layout comment above `phase_offset()` is the load-bearing
  invariant: epoch slots, barrier phases and barrier flags are three disjoint
  ranges, so correctness does not depend on each kernel remembering to skip
  phase 0.
- No architecture gate on the JIT spec. The kernels use only plain PTX
  loads/stores and CUDA IPC; the target is an interconnect, not an SM version.
- Size `max_numel` to the real workload. The epoch double buffer places its
  halves `world_size * max_numel` apart, so rounding up 4× at 8 ranks moves them
  32× the payload apart and costs measurable time at small batch.
- Following the precedent of the other `comm` trace templates, no entry was added
  to `tests/trace/example.py`: the workspace is an opaque multi-rank IPC handle
  that cannot be built in the single-process trace harness.

## Summary by CodeRabbit

* **New Features**
  * Added PCIe IPC all-reduce for intra-node PCIe systems with BF16 and FP16 support.
  * Added topology-aware launch selection, CUDA graph compatibility, workspace controls, and persistent autotuning.
  * Added tracing, public APIs, ahead-of-time kernel generation, and automatic fallback for unsupported shapes.
  * Added benchmarks for NCCL comparisons and protocol tuning.
* **Documentation**
  * Added usage, configuration, topology, tuning, and fallback guidance.
* **Tests**
  * Added comprehensive correctness, compatibility, tuning, and regression coverage.


ormandj:
TP2 full-model profiling at FlashInfer head `dca29052ac92789df4df95455170209a93b1ee73`:

- Hardware: 2× RTX PRO 6000 Blackwell Max-Q (SM120), TP2 over PCIe Gen4 x16, 300 W per GPU.
- Model: DeepSeek-V4-Flash-0731 with DSPARK block size 5 and BF16 all-reduce payloads.
- The matched NCCL and PCIe-IPC captures used the same image, source trees, model arguments, graph sizes, token IDs, runtime cache policy, and profiler settings. Only PCIe-IPC enablement and workspace capacity differed.
- Each decode arm retained 127 complete paired target-plus-draft scheduler steps.
- C1 mean step duration: 15.681 ms with NCCL and 14.965 ms with PCIe-IPC (-4.57%).
- C8 mean step duration: 35.399 ms with NCCL and 33.421 ms with PCIe-IPC (-5.59%).
- At C8, all 11,938 all-reduces used PCIe-IPC. The two rank timelines saved 1.734 ms and 1.781 ms per step, equal to 4.90% and 5.03% of their NCCL step durations.
- Prefill-sized reductions remained on NCCL; no prefill benefit was measured.

The SGLang consumer was tested at `3894b9014b30549fe7749c8dfeb0eaedecfa3e8f` with an explicit 786,432-element workspace. These are Nsight-delimited full-model step durations, not serving-throughput measurements.

A later integration using FlashInfer head `dca29052ac92789df4df95455170209a93b1ee73` completed five repetitions at every decode concurrency from C1 through C32, five cache-cold prefill repetitions at 8K, 32K, 64K, and 128K, GSM8K with 1,261/1,319 correct and zero request errors, and 8/8 long-output requests. These later results are integration coverage, not an NCCL-versus-PCIe-IPC attribution.

aleozlx (approved):

PR #4397, opened 2026-08-07
perf(blk64): build the Q/K/V tile layout in one pass

forrestl111:
# [Perf] bsa_attn_blk64: build the Q/K/V tile layout in one pass

## 📌 Description

`bsa_attn_blk64_fwd` rebuilds Q, K and V into the kernel's internal tiled layout
on **every call**. Per tensor that is currently `permute().contiguous()` →
`torch::zeros()` → `copy_()`, plus another `permute().reshape().contiguous()` for
K and V — three full passes over each activation where one suffices, and a zero
fill of the entire padded buffer when only the tail rows need zeroing.

Every destination layout is a **pure permutation** of the input, so it can be
allocated at its final shape and filled with a single strided copy:

```cpp
// before
auto q_bhsd = q.permute({0, 2, 1, 3}).contiguous();
auto q_padded = torch::zeros({batch, heads, rows_padded, kQkK}, q.options());
q_padded.narrow(2, 0, seq_q).copy_(q_bhsd);
auto q_tiled = q_padded.view({...}).contiguous();

// after
auto q_padded = torch::empty({batch, heads, rows_padded, kQkK}, q.options());
if (rows_padded > seq_q) {
  q_padded.narrow(2, seq_q, rows_padded - seq_q).zero_();   // pad rows only
}
q_padded.narrow(2, 0, seq_q).copy_(q.permute({0, 2, 1, 3})); // one strided copy
auto q_tiled = q_padded.view({...});                         // already contiguous
```

K and V are the same idea one level deeper: their
`(B, H, blocks, 2, 64, half)` destination is reachable from `(B, S, H, D)` by
view + permute, so it is filled directly. Whole blocks and the ragged tail block
are handled separately, and only the tail is zeroed.

Because the launch template now consumes strided BSHD directly, the Python
wrapper's three `q/k/v = x.contiguous()` lines become a wasted full-tensor copy
and are removed.

This is a **pure performance change** — no API, no semantics, no numerics.

Measured on MiniMax-H3 DiT attention (B200, S=37,736, 14 heads, head_dim 128,
bf16, sparsity 0.75):

| | per call |
| --- | ---: |
| re-tiling alone | 1.512 ms → **0.647 ms** (2.34x) |
| whole attention call | 4.78 ms → **3.47 ms** (1.38x) |
| copy kernels | 17 copies / 2252.6 µs → **6 copies / 777 µs** (2.9x) |

End to end in the caller this was written for (SGLang MiniMax-H3, 1344x768 / 5 s
/ 50 steps, 8x B200, Ulysses-8), DiT denoise goes 18.55 s → 17.32 s, i.e. the
patch alone is worth **1.071x** on a full video generation.

## 🔍 Related Issues

None — found while profiling `bsa_attn_blk64_fwd` as the attention kernel behind
a block-sparse backend for the MiniMax-H3 DiT in SGLang.

### ✅ Pre-commit Checks

## 🧪 Tests

No new tests: the change is a layout-construction rewrite with no observable
behaviour change, and the existing blk64 tests cover the kernel's contract.

Equivalence was verified **element-wise against the unpatched build**, not by
tolerance: same inputs, same seed, outputs compared with `torch.equal` →
**0 of 67,622,912 elements differ** (`out` is `[1, 37736, 14, 128]`). The three
re-tiled tensors `q_tiled`, `k_blocks`, `v_blocks` were also compared
individually against the old path and each passes `torch.equal`.

The ragged tail is the part worth a second pair of eyes, so it is exercised
explicitly: `seq_k` not a multiple of 64 (the `whole_blocks < total_sparse_blocks`
branch), and `seq_q` not a multiple of `kRows` (the Q pad-row branch).

## Reviewer Notes

Three things to focus on:

1. **The tail branches.** `whole_blocks = seq_k / kSparseBlockSize` splits the
   copy into whole blocks and a ragged remainder. The remainder path zeroes the
   tail block first and then fills `k_rest` rows — the zero has to come first,
   since `torch::empty` no longer pre-zeroes the buffer. Getting this wrong would
   leave uninitialised values in the phantom rows of the last block, which is
   exactly the case tolerance-based tests are worst at catching.

2. **`torch::zeros` → `torch::empty` + targeted `zero_()`.** This is where the
   saving mostly comes from, and it is also the only way this patch could
   introduce nondeterminism. Every element of every destination is now written by
   either a `copy_` or an explicit `zero_()`; if a reviewer finds a destination
   region covered by neither, that is a real bug and I would like to know.

3. **Dropping `.contiguous()` in the Python wrapper.** Safe only because the
   launch template permutes its inputs anyway. If any other call path into these
   symbols assumes contiguous q/k/v, that assumption is now load-bearing on this
   patch and I have missed it.

The K/V destinations differ only in the position of the block dimension
(`{0,3,1,4,2,5}` for K, `{0,3,1,4,5,2}` for V) — V's last two axes are swapped
because the kernel wants V transposed. That asymmetry is intentional, not a typo.

### DIFF
--- a/flashinfer/cute_dsl/sparse/blk64/flash_fwd_launch_template.h
+++ b/flashinfer/cute_dsl/sparse/blk64/flash_fwd_launch_template.h
@@ -44,6 +44,12 @@ std::vector<torch::Tensor> bsa_fused_fwd_blk64_launch(
   const int seq_q = static_cast<int>(q.size(1));
   const int heads = static_cast<int>(q.size(2));
   const int seq_k = static_cast<int>(k.size(1));
+  // MQA (a single KV head shared by every query head) is expressed by leaving
+  // the head dimension at 1 and letting the fill below broadcast it.
+  const int heads_kv = static_cast<int>(k.size(2));
+  TORCH_CHECK(heads_kv == heads || heads_kv == 1, "k/v must have either ", heads,
+              " heads or 1 (MQA), got ", heads_kv);
+  TORCH_CHECK(v.size(2) == heads_kv, "k and v must have the same number of heads");
 
   // Phantom block padding: round up to multiple of 8 (kSparseBlocksPerKV*2) for even kv_iters.
   constexpr int kAlign = kSparseBlocksPerKV * 2;  // 8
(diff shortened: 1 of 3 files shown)


hsr1234563:
I plan to replace the C++ implementation here with blk64 of sm100 implemented by cutedsl, so please wait for the update of the next version. Thanks，forrestl111

PR #4048, opened 2026-07-18
feat(moe): support BF16 activations in CuTe DSL NVFP4 MoE

zianglih:
## 📌 Description
@humansand

Adds an explicit `quant_mode="w4a16"` path to the SM100 CuTe DSL NVFP4 MoE APIs.

### Design walkthrough

- **Interface:** Explicit `quant_mode="w4a16"` mirrors the existing SM12x CuTe DSL MoE quant-mode dispatch instead of inferring the path from tensor dtype. W4A4/NVFP4 remains the default; W4A16 requires BF16 `x`, `x_sf=None`, `fc2_input_scale=None`, and no per-token scale. The unified runner uses the existing `MoEConfig.quant.variant`, `RoutingInputMode.PackedPrecomputed`, `TunableRunner`, `use_fused_finalize`, and `enable_pdl` contracts.
- **Weight decode:** NVFP4 E2M1 values and E4M3 scales are decoded online to BF16. CUDA 13.2+ uses direct explicit-RN BF16 PTX; earlier compilers use explicit-RN FP16 conversion/multiply followed by FP32-to-BF16 widening.
- **Weight layout:** TMA loads the unchanged public packed layout into SMEM. The transform assigns contiguous complete NVFP4 blocks to two decode warpgroups and writes BF16 in MMA/TMEM layout, so each E4M3 scale is decoded once and packed reads stay contiguous. For gated GEMM1 it also places 16 up rows beside their 16 gate rows, giving each epilogue warp both operands in its 32 TMEM lanes without a global repack.
- **Shared-memory reuse:** The E4M3 scale buffer is shared by its TMA and transform layouts, and gated decode uses composed views of the existing packed-weight and scale stages instead of allocating repacked copies. Decoded BF16 weights go directly to TMEM. Fused finalize reinterprets its single C stage as the contiguous reduction tile, while deterministic finalize assigns only unused SMEM to additional C stages.
- **Cache policy:** Activation TMA loads use evict-last because routed activation tiles are reused across output-weight M tiles. Streaming packed-weight TMA loads use evict-first so they do not displace those reusable activations from L2.
- **Warp specialization:** The 16-warp CTA uses warps 0-3 for the epilogue at 176 registers/thread; warps 4-7 for MMA, scheduling, weight TMA, and activation TMA at 80; and warps 8-15 as two four-warp weight-dequantization groups at the 128-register launch default. This exactly uses the 65,536-register CTA budget.
[...]

PR #4406, opened 2026-08-07
fix(monomoe): restore CUDA 12.0+ compatibility in tma_load_2d

aleozlx:
## 📌 Description

The rewrite in #4027 changed `tma_load_2d` in `csrc/fused_moe/monomoe/src/ptx_utils.h` from `.shared::cluster` to `.shared::cta`. The `.shared::cta` form requires PTX ISA 8.6 / CUDA 12.8+, breaking the `aot-build-import (cu126)` CI job with `ptxas: State space incorrect for instruction 'cp.async.bulk.tensor'`.

Fix: revert to `.shared::cluster` and remove the erroneous `.tile` qualifier (`.tile` is only valid with `.shared::cta` and was never correct on the cluster form). This matches the SM90 fallback in CUTLASS `cute/arch/copy_sm90_tma.hpp`, which uses `shared::cluster` without `.tile` for all pre-SM120 targets. Restores CUDA 12.0+ compatibility.

No performance impact: monomoe launches with cluster size 1, so `.shared::cluster` and `.shared::cta` are semantically identical for this kernel. The `.tile` qualifier is a PTX syntax disambiguator only; hardware behavior is unchanged.

The first commit added a CUDA >= 12.8 version gate as an initial workaround; the second commit replaces it with this root-cause fix and reverts the gate.

## 🔍 Related Issues

Reported in #4048 (comment): `ptxas monomoe_binding.ptx: State space incorrect for instruction 'cp.async.bulk.tensor'` on cu126 AOT build.

## 🚀 Pull Request Checklist

## 🧪 Tests

- [ ] `aot-build-import (x64, cu126)` and `aot-build-import (arm64, cu126)` pass in CI
- [ ] `aot-build-import (x64/arm64, cu128/cu129/cu130)` continue to pass

## Reviewer Notes

The PTX change is a one-liner: `shared::cta.global.tile` → `shared::cluster.global` (dropping `.tile`). CUTLASS's `cute/arch/copy_sm90_tma.hpp` uses exactly this form for the SM90 path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility for tensor loading across supported PTX and CUDA versions.
  * Updated tensor addressing behavior for single-CTA kernel launches.
  * Improved support across compatible hardware and toolchains.

### DIFF
--- a/csrc/fused_moe/monomoe/src/ptx_utils.h
+++ b/csrc/fused_moe/monomoe/src/ptx_utils.h
@@ -330,8 +330,12 @@ __device__ static __forceinline__ void tma_load_2d(CUtensorMap const& desc, std:
   std::uint32_t dst_addr = cvta_to_shared_u32(dst_smem);
   std::uint32_t bar_addr = cvta_to_shared_u32(bar_smem);
   std::uint64_t desc_addr = reinterpret_cast<std::uint64_t>(&desc);
+  // `.shared::cluster` without `.tile` works from PTX ISA 8.0 / CUDA 12.0+.
+  // `.shared::cta` and the `.tile` qualifier both require PTX ISA 8.6 /
+  // CUDA 12.8+.  This matches the fallback form in CUTLASS
+  // cute/arch/copy_sm90_tma.hpp (which also omits `.tile` for SM90).
   asm volatile(
-      "cp.async.bulk.tensor.2d.shared::cta.global.tile"
+      "cp.async.bulk.tensor.2d.shared::cluster.global"
       ".mbarrier::complete_tx::bytes"
       " [%0], [%1, {%2, %3}], [%4];\n"
       :
--- a/csrc/fused_moe/monomoe/src/ptx_utils.h
+++ b/csrc/fused_moe/monomoe/src/ptx_utils.h
@@ -330,10 +330,20 @@ __device__ static __forceinline__ void tma_load_2d(CUtensorMap const& desc, std:
   std::uint32_t dst_addr = cvta_to_shared_u32(dst_smem);
   std::uint32_t bar_addr = cvta_to_shared_u32(bar_smem);
   std::uint64_t desc_addr = reinterpret_cast<std::uint64_t>(&desc);
-  // `.shared::cluster` without `.tile` works from PTX ISA 8.0 / CUDA 12.0+.
-  // `.shared::cta` and the `.tile` qualifier both require PTX ISA 8.6 /
-  // CUDA 12.8+.  This matches the fallback form in CUTLASS
-  // cute/arch/copy_sm90_tma.hpp (which also omits `.tile` for SM90).
+  // Use `.shared::cluster` without `.tile` — the PTX ISA 8.0 / CUDA 12.0+ form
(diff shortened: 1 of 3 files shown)


bkryu (approved):

aleozlx:
passed both AOT and GPU test locally

nv-yunzheq (approved):

yongwww (approved):

aleozlx:
## Validation on CUDA 12.6

Reproduced the reported failure and validated this fix end-to-end, both compile-time and on hardware.

### 1. Compilation under ptxas 12.6

Container `flashinfer/flashinfer-ci-cu126:20260408-4cce866` (the tag pinned in `ci/docker-tags.yml`), `nvcc release 12.6, V12.6.20`. Compiled the real TU `csrc/fused_moe/monomoe/monomoe_binding.cu` with `-gencode arch=compute_90a,code=sm_90a`:

| `tma_load_2d()` form in `ptx_utils.h` | Result |
| --- | --- |
| `shared::cta.global.tile` (before this PR) | **fails** — `State space incorrect for instruction 'cp.async.bulk.tensor'`, 3 sites |
| `shared::cluster.global` (this PR) | **compiles**, 6.9 s |

The pre-fix control reproduces the CI error verbatim, so the check isn't vacuous.

### 2. Correctness on Hopper

H200 (`compute_cap 9.0`, SM90a) inside the *same* cu126 container, so the kernel is both JIT-compiled and executed under 12.6:

```
============================= test session starts ==============================
collected 26 items

tests/moe/test_monomoe.py::test_monomoe_accuracy[1-1] PASSED             [  3%]
...
tests/moe/test_monomoe.py::test_monomoe_small_scales[0.003] PASSED       [100%]

============================= 26 passed in 21.42s ==============================
```

Coverage: `test_monomoe_accuracy` (m ∈ {1,8} × top_k ∈ {1..8}), `test_monomoe_scratchpad_reuse_no_contamination`, `test_monomoe_cuda_graph_replay`, `test_monomoe_small_scales`.

These tests **skip** rather than fail when the module can't build, so I confirmed provenance rather than trusting the green: no `flashinfer` in site-packages, `flashinfer/data/csrc` → the tree under test, the `ptx_utils.h` the JIT actually compiled contains `shared::cluster` at line 348, and `has_monomoe()` returned `True`.

### 3. CI

`AOT Build Import (x64, cu126)` and `(arm64, cu126)` are both green on this PR — the jobs that were failing.

### Note on semantics / perf

No performance impact expected. The kernel is always launched with a plain `<<<grid, block>>>` triple-chevron (cluster size 1, no `cudaLaunchKernelEx` / `__cluster_dims__`), so `shared::cluster` and `shared::cta` address the same physical bytes. `.tile` was only ever a tile-vs-im2col disambiguator introduced alongside `.im2col` in PTX ISA 8.6, hence redundant on the older form. `shared::cluster` without `.tile` is what CUTLASS itself emits for SM90 (`cute/arch/copy_sm90_tma.hpp`).

### Related

Opened #4407 as a draft fallback (gate the monomoe AOT spec on CUDA >= 12.8) in case this fix needed longer to validate. Since cu126 CI is green here, that one can most likely just be closed.

<sub>Validation performed with AI assistance (Claude Code).</sub>

PR #4417, opened 2026-08-08
feat(kda): add packed-input CuTe decode kernel

ameynaik-hub:
# Packed-input CuTe-DSL KDA T=1 decode kernel for B200

Adds `flashinfer.kda_kernels.packed_kda_decode_cute.run_packed_kda_decode_cute`, a CuTe-DSL decode kernel for serving-native packed KDA (Kimi Delta Attention) T=1 inputs: packed bf16 QKV rows `[B, 3*12*128]`, raw gate/beta logits, fp32 internal math, bf16 state pool `[N, 12, 128, 128]` updated in place, `state_indices` rows outside the pool (negative or past the end) produce zero output and leave the pool untouched. CUDA-graph capture/replay safe; runs on the caller's current stream via TVM-FFI.

## Kernel design

A pipelined implementation tuned per batch size:

- **cp.async shared-memory ring** (B≥24): 128-thread CTAs stream the bf16 state through a 4–5-slot ring of 4KB chunks (`cp.async.cg`, L1 bypass) with a barrier-free per-thread pipeline and LDS double-buffering, so in-flight read volume is not bounded by the register file.
- **Fused output projection**: `o = (h·d)·q + vn·(k·q)` with `k·q` computed once at staging — removes one butterfly reduction tree per row.
- **bf16 register economy**: state stays packed bf16 in registers; unpack is a shift/mask bit trick and repack a single `cvt.rn.bf16x2.f32` (full-rate ALU instead of the conversion pipe), feeding packed f32x2 FMAs.
- **Per-batch policy** (`_select_config`): register-prefetch kernel with 128-thread CTAs (B≤11) or 32-thread CTAs (B∈[12,23]); cp.async kernel with half-head tiles (B∈[24,37]) or whole-head tiles (B≥38). `tile_v=` forcing maps onto tuned schedules for sweeps.

## Benchmark results

B200, driver 595.58.03, CUDA 13.0, `benchmarks/bench_packed_kda_decode.py` (CUPTI kernel timing, warmup 100, iterations 100, default clocks). "Previous" is the earlier single-warp variant of this PR at the same protocol; medians in µs:

| B | direct (prev → this) | cuda_graph (prev → this) | speedup (direct) |
|---|----------------------|--------------------------|------------------|
| 1 | 2.75 → **2.37** | 2.50 → **2.11** | 1.16× |
| 8 | 3.46 → **3.20** | 3.23 → **2.98** | 1.08× |
| 16 | 4.53 → **4.38** | 4.42 → **4.29** | 1.03× |
| 31 | 5.89 → **5.89** | 5.70 → 5.76 | 1.00× |
| 32 | 6.00 → 6.06 | 5.82 → 5.89 | 0.99× (tie) |
| 64 | 9.47 → **9.09** | 9.25 → **8.90** | 1.04× |
| 128 | 15.55 → **14.32** | 15.26 → **14.05** | 1.09× |
| 256 | 35.14 → **31.87** | 35.26 → **32.10** | 1.10× |
| 512 | 68.16 → **63.82** | 68.13 → **63.89** | 1.07× |

No regression at any batch size (B=31/32 within run-to-run noise). Correctness at every batch is checked against an fp32 torch reference before timing (`--refcheck` equivalent is built into the script); max output error is at the bf16 quantization level.

## How to run the benchmark

```bash
# full sweep, both launch modes
python benchmarks/bench_packed_kda_decode.py --warmup 100 --iterations 100

# specific batches / direct launches only / JSON report
python benchmarks/bench_packed_kda_decode.py \
    --batch-size 1 32 512 --mode direct \
    --warmup 100 --iterations 100 --json results.json

# force a tile schedule (benchmark override; default lets the policy pick)
python benchmarks/bench_packed_kda_decode.py --tile-v 64

# cold-L2 timing
python benchmarks/bench_packed_kda_decode.py --cold-l2
```

Requires a B200 (exact CC 10.0) and CUDA ≥12.8; `pip install -U cupti-python` for CUPTI timing (falls back to CUDA events).

Tests:

```bash
pytest tests/kda/test_packed_kda_decode_cute.py -m "" -v
```

36 tests: reference match B=1–512, all forced tiles, sanitizer (odd-stride) schedules, shifted/misaligned tensors for every argument, out-of-range slots, all-inactive bitwise no-op, CUDA-graph replay with changed inputs and indices, current-stream semantics, and a 512-step fp64 drift diagnostic.

## T=1 fast path inside `recurrent_kda` (no new public API)

The kernel takes q/k/v/g/beta as five independent strided tensors — the packed layout only ever existed in the launcher. A new `launch_unpacked_kda_decode_cute` entry point exposes this, and `run_recurrent_kda` now routes **eligible T=1 decode calls** to it: `T=1`, `H=HV=12`, `K=V=128`, bf16, SM100a, `use_gate_in_kernel=True` with `lower_bound=-5`, `beta_is_logit=True`, in-kernel QK L2 norm, 1-D `ssm_state_indices`, no spec/varlen/GQA/final-state. Ineligible calls use the existing kernels unchanged. Toggle: `FLASHINFER_KDA_T1_FAST_PATH=0` disables (default enabled).

The fast path indexes the state pool **in-kernel** (replacing the host-side slot gather/scatter), and additionally accepts padded state pools and strided q/k/v/g views (zero-copy from a fused-QKV projection GEMM). Measured through the public `recurrent_kda` API with a state cache (same protocol):

| B | fast path off | fast path on | speedup |
|---|---|---|---|
| 8 | 66.0µs | 4.2µs | 15.9× |
| 64 | 88.5µs | 11.1µs | 8.0× |
| 512 | 638.0µs | 63.7µs | 10.0× |

Unpacked dispatch is bitwise-identical to the packed entry point on equal inputs (same cubin, different base pointers).

Both `recurrent_kda` calling conventions are covered: the raw form (`use_gate_in_kernel=True`, `lower_bound=-5`, `beta_is_logit=True` — the kernel computes the decay and sigmoid in fp32) and the API-default pre-computed form (log-space `g`, pre-sigmoided `beta`), which compiles a kernel variant with `decay = exp(g)`. Pre-computed mode measures 3.9µs (B=8) / 63.9µs (B=512) on the fast path.

For attribution (per-kernel profile at B=512): the generic path's decode kernel itself takes 81.1µs (so kernel-vs-kernel the gain is 1.27×, matching its no-indices benchmark of 83.4µs); the remainder of its 638µs is the wrapper's state round-trip — a 69µs vectorized gather plus a 463µs elementwise `index_copy_` scatter running at 0.87 TB/s. The scatter is independently fixable in the wrapper, but even with a perfect scatter the gather/scatter design pays two extra passes over the state (~126µs at B=512); the fast path's in-kernel indexing removes those passes entirely.

## DRAM utilization

The kernel is DRAM-bound at large batch: per step it must read and rewrite the full `B×12×128×128` bf16 state (786KB/row round trip; ~402MB at B=512) plus ~15KB/row of activations. From the benchmark's `logical_TB/s` column (compulsory bytes ÷ median time), against the 8 TB/s HBM3e peak:

| B | effective DRAM (prev) | effective DRAM (this PR) |
|---|----------------------|--------------------------|
| 64 | 5.42 TB/s (68%) | 5.65 TB/s (71%) |
| 128 | 6.60 TB/s (82%)* | 7.17 TB/s (90%)* |
| 256 | 5.84 TB/s (73%) | 6.44 TB/s (81%) |
| 512 | 6.02 TB/s (75%) | **6.43 TB/s (80%)** |

\* B=128 exceeds the cold-memory ceiling because this benchmark's state pool (~50MB) partially survives in the 126MB L2 across iterations; B≥256 is the true DRAM-resident regime.

Nsight Compute confirms the gain is pure memory-pipeline efficiency, not traffic: at B=512 both kernels move identical bytes (`dram__bytes_read.sum` 207.7MB, in-kernel writes ~140MB with the remainder draining from L2 after kernel end), while this variant sustains ~8% higher `dram__throughput.avg.pct_of_peak_sustained_elapsed` (47.7% vs 44.1% under locked profiling clocks). At ~80% of peak on interleaved cold-L2 read/write, the kernel is at the observed machine ceiling for this access pattern — remaining headroom would require reducing bytes (e.g., fp8 state), not scheduling.


kahyunnam (changes requested):
Blocking API-design question: please clarify whether this is intended as a public operation or an internal experimental backend.

The existing KDA surface is already split between the phase-neutral flashinfer.recurrent_kda, the separate flashinfer.kda_decode.recurrent_kda, and flashinfer.fused_kda_decode. This PR currently introduces another pattern: the only callable is the backend-specific `flashinfer.kda_kernels.run_packed_kda_decode_cute`, exported from kda_kernels.__all__, with no `@flashinfer_api`, trace template, top-level facade, or documentation.

Because the packed serving contract differs materially from recurrent_kda, a separate operation may be justified. 
- If public, please expose a backend-neutral noun-form API (e.g., packed_recurrent_kda_decode) through the normal facade, decorator, trace, top-level export, and docs, while retaining run_packed_kda_decode_cute as the internal implementation. 
- If this is experimental/internal instead, please avoid exporting it as public API and state that explicitly.

(I don’t think this PR needs to solve the pre-existing KDA unification problem, but it should probably not add a third public-entrypoint convention)

flashinfer/kda_kernels/packed_kda_decode_cute.py
```diff
@@ -0,0 +1,2478 @@
...
+
+@functools.cache
+def _use_packed_fma() -> bool:
+    return torch.cuda.get_device_capability(0)[0] >= 10
+
+
+# tile_v thresholds on work_units = B * H: bigger tiles amortise the q/k/gate
+# staging and CTA overheads, smaller tiles add CTAs so small batches still
+# fill the SMs. Tuned on B200 (see benchmarks/bench_packed_kda_decode.py).
+_TILE_V_OVERRIDE = os.environ.get("FLASHINFER_PACKED_KDA_TILE_V")
```
saltyminty: This environment variable lookup will only run once during module import, and will not be updated if the variable is changed in the future. Verifying that this is intended, otherwise it should be re-evaluated/looked up during each call.

ameynaik-hub:
@kahyunnam 

  - The public way to use these kernels is the existing `flashinfer.recurrent_kda`. Its wrapper now routes eligible T=1 decode calls (12 heads, head dim 128, bf16, SM100a, 1-D `ssm_state_indices`; both raw and pre-computed gate/beta conventions) to the new kernel.
  All other calls (T>1, spec decode, GQA, varlen, other shapes) run the existing kernels unchanged. Kill switch: `FLASHINFER_KDA_T1_FAST_PATH=0`.
  - `run_packed_kda_decode_cute` is no longer exported from `kda_kernels.__all__`; its module docstring states it is internal (kept importable by module path for its tests and benchmark).
  - So no third public-entrypoint convention is added — eligible calls inherit `recurrent_kda`'s existing `@flashinfer_api`, trace template, top-level export, and docs.
  
  Benchmarks vs the existing recurrent kernel (B200, CUPTI median, warmup 100 / 100 iters, through the public `recurrent_kda` API): 
  
  | Batch | Existing kernel (kernel time only) | Existing full decode step (with state cache) | New kernel | Speedup (kernel / full step) |
  |------:|-----------------------------------:|---------------------------------------------:|-----------:|:-----------------------------|
  | 8     | ~4.0 µs                             | 63.3 µs                                       | **4.1 µs** | ~1.0× / 15× |
  | 64    | 13.5 µs                             | 87.8 µs                                       | **11.0 µs**| 1.2× / 8× |
  | 512   | 81.1 µs                             | 638.8 µs                                      | **63.7 µs**| 1.27× / 10× |
  
  The big full-step gap is not the existing kernel's fault: with `ssm_state_indices` its wrapper copies each request's state out of the pool and back (at B=512: 69 µs gather + 463 µs elementwise scatter). The new kernel looks up its slot inside the kernel, so
  those copies disappear. Kernel-vs-kernel the gain is a more modest 1.2–1.3×.
  
  Correctness: 42 tests pass, including fp32-reference parity for both paths and both gate conventions, toggle behavior, misaligned/strided inputs, out-of-range slots, and CUDA-graph replay.

flashinfer/kda_kernels/packed_kda_decode_cute.py
ameynaik-hub: Thanks @saltyminty , addressed it.

saltyminty (approved):

kahyunnam (approved):

PR #4420, opened 2026-08-08
feat: SM120 NVFP4 SVDQuant Gemm in CuteDSL

rosenrodt:
## 📌 Description

This PR adds SM120/SM121 CuTe DSL support for NVFP4 SVDQuant GEMM under the
existing FlashInfer API. It provides a true fused residual-plus-BF16 LoRA-up
kernel and retains the unfused composition as an explicit correctness oracle.

The complete path also:

- adds fused and unfused SM120 dispatch to `mm_nvfp4_svdquant` and
  `svdquant_linear` without changing API contracts;
- fuses BF16 smoothing into NVFP4 activation quantization and executes
  LoRA-down on BF16 tensor cores;
- aliases LoRA-up correction storage with released residual-mainloop A/B
  stages and uses the same producer/consumer pipeline;
- autotunes fused versus unfused implementation choice and SM120 M/N/K tile
  tactics;
- adds SM120/SM100 coverage plus SVDQuant trace templates, examples,
  and generated goldens; and
- instrument the sm120 gemm kernel with IKET regions. IKET is disabled by default and
  ordinary and instrumented specializations have distinct cache identities.

### Performance

Reportable performance is from RTX PRO 6000 only. The tables below use rank
32, CUDA Graph replay, and warm-L2 timing. They report complete
`svdquant_linear` latency, gains over hidden BF16 and per-tensor FP8 baselines,
and the fused/unfused correction comparison. `residual_fp4/fused` compares
latency; TFLOPS and microseconds share one cell as `TFLOPS / us`.

Reproduce the benchmark script's complete default Qwen3-image sweep from the
repository root on an idle, exclusive RTX PRO 6000. The defaults cover three
`(N,K)` pairs, four `M` values, and rank 32:

```bash
python -u benchmarks/bench_nvfp4_svdquant_gemm.py \
  --svdquant-backend fused \
  --cuda-graph \
  --no-cold-l2-cache
```

| M | N | K | R | `svdquant_linear` us | gain vs BF16 | gain vs FP8 | residual_fp4/fused | residual FP4 TF/s / us | fusion gain | fused TF/s / us | unfused TF/s / us |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 4096 | 3072 | 3072 | 32 | 97.44 | +144.1% | +23.0% | 0.958 | 1173.00 / 65.91 | +97.7% | 1135.86 / 68.77 | 574.56 / 135.96 |
| 6889 | 3072 | 3072 | 32 | 127.65 | +170.4% | +41.3% | 0.953 | 1335.73 / 97.34 | +100.6% | 1286.10 / 102.15 | 641.28 / 204.87 |
| 9216 | 3072 | 3072 | 32 | 182.14 | +152.6% | +35.6% | 0.950 | 1255.88 / 138.51 | +100.3% | 1205.55 / 145.79 | 601.89 / 292.01 |
| 16384 | 3072 | 3072 | 32 | 358.98 | +137.8% | +24.8% | 0.939 | 1262.19 / 245.00 | +167.9% | 1197.99 / 260.82 | 447.11 / 698.84 |
| 4096 | 12288 | 3072 | 32 | 308.59 | +171.5% | +45.1% | 0.950 | 1277.15 / 242.13 | +174.5% | 1226.15 / 254.83 | 446.72 / 699.46 |
| 6889 | 12288 | 3072 | 32 | 486.36 | +176.0% | +49.7% | 0.938 | 1244.37 / 417.96 | +197.4% | 1179.89 / 445.40 | 396.68 / 1324.81 |
| 9216 | 12288 | 3072 | 32 | 624.29 | +193.3% | +59.3% | 0.949 | 1271.11 / 547.38 | +208.3% | 1219.34 / 576.57 | 395.54 / 1777.42 |
| 16384 | 12288 | 3072 | 32 | 1121.85 | +191.3% | +54.5% | 0.943 | 1276.91 / 968.71 | +211.2% | 1216.82 / 1027.13 | 391.04 / 3196.17 |
| 4096 | 3072 | 12288 | 32 | 348.53 | +145.9% | +33.5% | 0.991 | 1243.46 / 248.69 | +24.2% | 1235.55 / 250.93 | 994.43 / 311.78 |
| 6889 | 3072 | 12288 | 32 | 633.92 | +115.2% | +13.8% | 0.986 | 1387.66 / 374.80 | +32.7% | 1372.01 / 380.07 | 1033.84 / 504.39 |
| 9216 | 3072 | 12288 | 32 | 883.67 | +112.4% | +9.6% | 0.986 | 1296.19 / 536.79 | +32.9% | 1281.84 / 544.21 | 964.17 / 723.52 |
| 16384 | 3072 | 12288 | 32 | 1552.04 | +116.1% | +14.6% | 0.980 | 1311.20 / 943.37 | +47.1% | 1287.78 / 963.03 | 875.48 / 1416.56 |

## 🔍 Related Issues

N/A.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

> If you are unsure about how to set up `pre-commit`, see the
> [pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

```bash
python -m pytest -q tests/gemm/test_nvfp4_svdquant_gemm.py -k "sm120"
```

## Summary by CodeRabbit

* **New Features**
  * Added NVFP4 SVDQuant support across SM100/SM103 and SM120/SM121 GPUs.
  * Added selectable fused, unfused, automatic, and CuTe DSL execution backends.
  * Added BF16 smooth quantization and FP8 residual benchmark baselines.
  * Added CUDA Graph, L2-cache, autotuning, and expanded performance measurement options.
  * Added trace definitions for NVFP4 SVDQuant GEMM, linear operations, smooth quantization, and BF16 GEMM.

* **Bug Fixes**
  * Updated SM120/SM121 backend selection to support CUDA 12.9 and newer.
  * Improved input, shape, dtype, and device validation.


rosenrodt:
@flashinfer-bot run

dhiraj113 (commented):
Reviewed the Python integration layer (dispatch, quantization wrapper, autotune/cache identity, tests, trace) on this branch, with verification run on a local B200. Kernel math in `dense_blockscaled_gemm_sm120_b12x.py` was read but not line-reviewed.

**This is a well-built PR** — the dispatch layer is careful (the nested-AutoTuner-under-capture note at line 430 shows real thought), and the trace work is clean: I regenerated all three SVDQuant definitions via `fi_trace(...)` with `example.py`'s exact shapes and byte-compared against the committed goldens — **all three match exactly**.

Things I checked that are **correct**, so nobody re-derives them:

- Backend selection cannot pick an unsupported backend. `is_backend_supported` -> 100/103: `['cutlass']`, 120/121: `['cute-dsl','cute-dsl-unfused']`, 90: `[]`. The CUDA-12.9 and DSL-availability raises correctly *exclude* backends via `suitable_auto_backends` rather than crashing.
- No `@backend_requirement` bypass despite the unfused runner calling `_b12x_gemm_fp4_runner` directly — I diffed the requirement sets and they're equivalent.
- The CUDA 12.9 gate is correct: `get_cuda_version()` returns a `packaging.version.Version` (so 12.10 > 12.9 orders properly) and is `functools.cache`d. Worth noting in the description though: on SM120 + CUDA 12.9-12.x, `mm_fp4(backend="auto")` now prefers b12x over cutlass/cudnn for **all** existing users.
- The autotune `TuningConfig` constrains every M-tracking tensor.
- Smooth-quantize is bit-exact vs `nvfp4_quantize((x*pqs).to(bf16))` at every shape I tried, with no regression to the existing path.
- IKET is fine. Line 203 omits the `iket` token from `options=`, but `CUTE_DSL_COMPILER_OPT=iket` — the DSL's documented switch — *composes* with an explicit options string (`apply_envar_settings` -> `_apply_opt_string`), which I verified yields a byte-identical instrumented object. Adding the token inline would make the `_iket1` cache entry self-contained, but it's ergonomics, not a bug.

**What I'd like addressed:** the cache-key comment on `gemm_svdquant.py:215` before merge (one line, with direct in-repo precedent, and I reproduced the stale-artifact behaviour). The `alpha.numel()` comment needs a decision either way since it changes behaviour on the untouched SM100 path. Everything else is test coverage and hygiene.

One CI note that isn't this PR's fault but bears on the test comments: `pr-test.yml` has only sm75/sm86/sm90 GPU runners, so *every* test in `test_nvfp4_svdquant_gemm.py` skips in CI (I measured `1 passed, 21 skipped` on a B200). The coverage gaps below will only ever be caught by a human on an SM120 part.

Each comment includes a reproduction command where one exists.

flashinfer/gemm/gemm_svdquant.py
```diff
@@ -62,6 +69,463 @@ def _swizzled_sf_size(rows: int, sf_cols: int) -> int:
...
+    kernel_name = (
+        f"r{rank}_bias{int(with_bias)}_t{mma_tiler_mn[0]}x{mma_tiler_mn[1]}"
+        f"x{tile_k}_swap{int(swap_ab)}_pf{int(use_prefetch)}_mac{max_active_clusters}"
+        f"_pdl{int(enable_pdl)}_iket{int(enable_iket)}"
+    )
+    compiled = build_and_load_cute_dsl_kernel(
+        "mm_nvfp4_svdquant_sm120",
+        kernel_name,
+        compile_kernel,
+        extra_key_files=(__file__, dense_blockscaled_gemm_sm120_b12x.__file__),
```
dhiraj113: **Bug — stale kernel artifact on any edit to `flashinfer/cute_dsl/utils.py`.**

`extra_key_files` is the *entire* source fingerprint for this disk-cached module — `_hash_source_files` opens only the listed paths, there is no transitive import walk. But `dense_blockscaled_gemm_sm120_b12x.py:79-81` imports `sm120_make_smem_layout_sfa` / `sm120_make_smem_layout_sfb` from `flashinfer/cute_dsl/utils.py`, and those are called from `_setup_attributes` (357/363) and `_make_smem_layouts` (2648/2654) to build the staged SFA/SFB **shared-memory layouts that are compiled into the artifact**.

So editing those helpers changes codegen but leaves `source_sha256` unchanged, and `try_load()` reloads the stale `.o`.

Reproduce (replicates `_hash_source_files` verbatim, restores the file afterwards):

```bash
cd $(git rev-parse --show-toplevel)
python - <<'PY'
import hashlib, pathlib
def key(paths):
    h = hashlib.sha256()
    for p in paths:
        h.update(pathlib.Path(p).read_bytes())
    return h.hexdigest()[:16]

svd = ["flashinfer/gemm/gemm_svdquant.py",
       "flashinfer/gemm/kernels/dense_blockscaled_gemm_sm120_b12x.py"]
utils = pathlib.Path("flashinfer/cute_dsl/utils.py")
orig = utils.read_bytes()
before = key(svd)
utils.write_bytes(orig + b"\n# probe\n")
after = key(svd)
utils.write_bytes(orig)          # restore byte-for-byte
print("svdquant key before/after:", before, "/", after,
      "->", "STALE REUSE" if before == after else "ok")
PY
```
Output: `e7df99df63144810 / e7df99df63144810 -> STALE REUSE`.

The correct pattern is already established for this exact module: `flashinfer/fused_moe/cute_dsl/blackwell_sm12x/moe_dispatch.py:640-670` — the only *other* disk-cached module compiling this same kernel — lists both `cute_dsl_utils.__file__` and `dense_blockscaled_gemm_sm120_b12x.__file__`, with a docstring naming the reason. See also `nvfp4_quantize.py:1461-1466`.

**Fix:** add a `_svdquant_kernel_source_files()` helper returning `(__file__, dense_blockscaled_gemm_sm120_b12x.__file__, flashinfer.cute_dsl.utils.__file__)`.

In fairness this bites editable dev installs rather than released wheels (fixed sources + version-scoped cache dir), but it's a one-line fix with in-repo precedent.

flashinfer/gemm/gemm_svdquant.py
```diff
@@ -190,12 +683,43 @@ def _check_mm_nvfp4_svdquant_problem(
...
+    expected_b_sf = _swizzled_sf_size(n, k // 16)
+    if a_sf.numel() < expected_a_sf or b_sf.numel() < expected_b_sf:
+        raise ValueError(
+            "128x4 scale buffers are too small: "
+            f"a_sf has {a_sf.numel()} elements (need {expected_a_sf}), "
+            f"b_sf has {b_sf.numel()} elements (need {expected_b_sf})"
+        )
+    if not a_sf.is_contiguous() or not b_sf.is_contiguous():
+        raise ValueError("a_sf and b_sf must be contiguous")
+    if alpha.dtype != torch.float32 or alpha.numel() != 1:
```
dhiraj113: **Behaviour change on the *unchanged* SM100 path — needs a decision.**

This check is new; the pre-PR `_check_mm_nvfp4_svdquant_problem` had no alpha validation, and the pre-PR docstring explicitly documented **`numel >= 1`**:

```bash
git show 5931d449:flashinfer/gemm/gemm_svdquant.py | sed -n '239p'
#     alpha: float32, device scalar (``numel >= 1``)
```

Because this lives in `common_check`, the tightening applies to **all** backends — including the pre-existing SM100/SM103 CUTLASS path this PR is otherwise not touching.

Reproduce on any SM100/SM103 box (worked before this PR, raises now):

```bash
python - <<'PY'
import torch, flashinfer
# alpha drawn from a pooled scalar buffer -- the documented numel>=1 contract
alpha = torch.empty(8, dtype=torch.float32, device="cuda")
alpha[0] = 1.0 / (448 * 6) ** 2
# ... build a/b/a_sf/b_sf/d/l1 as in tests/gemm/test_nvfp4_svdquant_gemm.py::_make_gemm_problem
# flashinfer.mm_nvfp4_svdquant(..., alpha=alpha, backend="cutlass")
# -> ValueError: alpha must be a float32 device scalar
PY
```

**Fix:** either keep `numel >= 1` and use `alpha.reshape(-1)[:1]` at the launch sites (note line 274 currently does `alpha.reshape(1)`, which needs the same change), or call this out as an intentional breaking change in the PR description.

To be clear, the sibling checks added in this same block — contiguity on `a`/`b`/`a_sf`/`b_sf`/`d`/`l1`/`out`, `bias.shape == (n,)`, same-device — are defensible hardening. It's specifically `numel != 1` that contradicts a documented allowance.

flashinfer/quantization/kernels/nvfp4_quantize.py
```diff
@@ -2001,6 +2055,68 @@ def nvfp4_quantize_cute_dsl(
...
+        "bfloat16",
+        k,
+        SF_LAYOUT_128x4,
+        enable_pdl,
+        _env_flag_enabled("FLASHINFER_DISABLE_FP4_QUANT_FAST_MATH"),
+        current_nvfp4_4over6_config(),
+        global_scale_is_tensor=True,
+        smooth_quant=True,
+    )
+    num_blocks = min(
```
dhiraj113, line 2126: **Crash on empty input — verified on a B200.**

`nvfp4_quantize_smooth_cute_dsl` has no empty-input early return. With `m == 0`: `padded_m = 0`, so `num_blocks = min(0, ...) = 0` and the launch at line 2107 fails. With `k == 0`, `_compute_optimal_threads(0)` divides by `threads_per_row == 0`.

```bash
python - <<'PY'
import torch
from flashinfer.quantization.kernels.nvfp4_quantize import nvfp4_quantize_smooth_cute_dsl
x   = torch.empty(0, 3072, dtype=torch.bfloat16, device="cuda")
pqs = torch.ones(3072, dtype=torch.bfloat16, device="cuda")
gs  = torch.tensor([1/(448*6)], dtype=torch.float32, device="cuda")
nvfp4_quantize_smooth_cute_dsl(x, pqs, gs)
PY
# RuntimeError: CUDA Error Code: 9   (cudaErrorInvalidConfiguration)
```

The sibling `silu_and_mul_nvfp4_quantize_cute_dsl` handles this correctly at line 2177 (`if m == 0 or k == 0:` -> return empty tensors).

Related: the dummy argument at line 2046 of `nvfp4_quantize_cute_dsl`, `input[0] if m > 0 else torch.empty(k, ...)`, suggests `m == 0` was considered reachable — but that `else` branch is dead: `nvfp4_quantize_cute_dsl` with `m=0` also dies with CUDA Error 9 at the launch immediately after. The guard only converts an `IndexError` into a CUDA launch error.

**Impact:** a serving step with an empty batch, or a CUDA-graph / dynamo warmup at batch 0.

**Fix:** mirror the silu path's early return after shape validation, and either drop the now-dead `else torch.empty(k, ...)` or give `nvfp4_quantize_cute_dsl` the same guard.

flashinfer/gemm/gemm_svdquant.py
```diff
@@ -62,6 +69,463 @@ def _swizzled_sf_size(rows: int, sf_cols: int) -> int:
...
+    correction = torch.mm(d, l1.T)
+    correction.mul_(alpha)
+    out.add_(correction)
+    if bias is not None:
+        out.add_(bias)
+    return out
+
+
+def _sm120_nvfp4_svdquant_runner(enable_pdl: bool):
+    class Sm120Nvfp4SvdquantRunner(TunableRunner):
```
dhiraj113, line 359: **`enable_pdl` is missing from the autotune profiling-cache key (perf, not correctness).**

Both this runner and `Sm120Nvfp4SvdquantUnfusedRunner` (line 444) capture `enable_pdl` in a closure. Neither class defines `__init__` or assigns any instance attribute, so `self.__dict__` is empty. `TunableRunner.__hash__` (`autotuner.py:766`) hashes the values of `self.__dict__` (skipping `*_cache` keys, `id()` fallback for unhashables), which for an empty dict is `hash(())` — identical for every instance. Neither class overrides `get_cache_key_extras()` either, so `_get_cache_key` yields the same `ProfilingCacheKey` for `enable_pdl=True` and `False`. The persisted `file_key` drops `runner_hash` entirely, so a saved cache is shared even more aggressively.

`enable_pdl` *is* a codegen parameter here — it's in the kernel cache key (line 110) and the disk name `_pdl{0|1}` (line 209).

```bash
python - <<'PY'
src = open("flashinfer/gemm/gemm_svdquant.py").read()
for cls in ("Sm120Nvfp4SvdquantRunner", "Sm120Nvfp4SvdquantUnfusedRunner"):
    i = src.find("class " + cls); j = src.find("class ", i + 10)
    print(cls, "defines __init__:", "def __init__" in src[i:j])
PY
# -> both False, so self.__dict__ is empty and hash(runner) collides
```

**Failure:** process A tunes with `enable_pdl=True` (the SM120 device default) and `save_configs()`; process B runs with explicit `enable_pdl=False`, hits the identical `file_key`, and replays a tactic ranked for the PDL variant. Silent perf regression — the tuned-range-exceeded warning doesn't fire because it *is* a hit.

**Fix:** override `get_cache_key_extras(self, inputs)` on both classes to return `(enable_pdl,)`. Precedent: `flashinfer/fused_moe/cute_dsl/tuner.py:849-857`.

Note the pre-existing `Nvfp4SvdquantGemmRunner` in this file has the same gap — inherited rather than invented, but worth fixing in the same pass.

tests/gemm/test_nvfp4_svdquant_gemm.py
```diff
@@ -357,6 +462,423 @@ def test_svdquant_linear_matches_reference(use_bias, rank):
     assert _sqnr_db(ref, out.float()) > 40.0
 
 
+def test_nvfp4_quantize_smooth_sm120_cute_dsl():
```
dhiraj113: **Test gap: the large-K column-loop branch this PR adds is executed by no test.**

This test uses one shape (m=129, k=256), and `test_svdquant_linear_sm120_fused` (line 807) uses the same k=256. At K=256, `threads_per_row = 16 <= num_threads`, so `needs_col_loop` is False and only the small-K multi-row branch (`nvfp4_quantize.py:612-681`) runs. `padded_sf_cols == num_sf_blocks_per_row == 16`, so the SF-column-padding zero-fill is never hit either.

The large-K branch at `nvfp4_quantize.py:554-611` — a **separate `_process_block(..., pre_quant_scale, ...)` call site added by this PR** — is therefore never executed. The SM100 cutlass sibling `test_nvfp4_quantize_smooth` (line 225) sweeps m in {44,129,256,1000} x n in {3072,12288,4096}.

I ran the missing shapes manually; all are bit-exact today, so this is coverage rather than a latent bug:

```bash
python - <<'PY'
import torch
from flashinfer.quantization.kernels.nvfp4_quantize import (
    nvfp4_quantize_smooth_cute_dsl, nvfp4_quantize_cute_dsl)
gs = torch.tensor([1/(448*6)], dtype=torch.float32, device="cuda")
for m, k in [(129, 256), (129, 12288), (70, 3088)]:   # small-K / col-loop / SF-pad
    x   = torch.randn(m, k, dtype=torch.bfloat16, device="cuda")
    pqs = torch.rand(k, dtype=torch.bfloat16, device="cuda") + 0.5
    a_q, a_sf = nvfp4_quantize_smooth_cute_dsl(x, pqs, gs)
    b_q, b_sf = nvfp4_quantize_cute_dsl((x * pqs).to(torch.bfloat16), gs)
    print(m, k, torch.equal(a_q, b_q), torch.equal(a_sf, b_sf))
PY
# 129 256 True True / 129 12288 True True / 70 3088 True True
```

**Fix:** parametrize this test over at minimum a column-loop K (12288), a K whose K/16 is not a multiple of 4 (3088), and an m that is not a multiple of 128 (70). Keep the bit-exact `torch.equal` assertions.

tests/gemm/test_nvfp4_svdquant_gemm.py
```diff
@@ -357,6 +462,423 @@ def test_svdquant_linear_matches_reference(use_bias, rank):
...
+    # single BF16 store, whereas the oracle rounds the residual and correction
+    # in separate launches. Compare numerically, not bitwise.
+    assert _sqnr_db(expected.float(), out.float()) > 35.0
+    assert _sqnr_db(expected.float(), out_auto.float()) > 35.0
+    fp32_ref = p["ref_bias"] if use_bias else p["ref"]
+    assert _sqnr_db(fp32_ref, out.float()) > 35.0
+
+
+@pytest.mark.parametrize("tile_k,rank", [(64, 32), (128, 32), (256, 64)])
+def test_mm_nvfp4_svdquant_sm120_large_m_tile(tile_k, rank):
```
dhiraj113: **Test gap: the pinned tactics here are ones the tuner can never select, while the one it *will* select is unchecked.**

`use_prefetch` is baked into the compiled specialization — it's a constructor arg of `Sm120B12xBlockScaledDenseGemmKernel`, part of the in-memory cache key (`gemm_svdquant.py:101-112`) and of the on-disk name `_pf{0|1}` (line 206-210). It is a **different kernel binary**, not a scheduling hint.

This test always passes `use_prefetch=False`, including for the `(256, 64)` tile. But `get_valid_tactics` only ever emits `(256,64)` with `prefetch_candidates=(True,)`, and only with `tile_k=128` (`gemm_svdquant.py:394-400`). So the three `(256,64)` tactics pinned here are ones the tuner will never choose, and the one it *will* choose for m>=256 — `((256,64), 128, False, True)` — has no deterministic correctness test. Under `autotune(True)` prefetch tactics get profiled, but only the winner's output is checked.

```bash
python -c "
import re,sys
src=open('flashinfer/gemm/gemm_svdquant.py').read()
i=src.find('prefetch_candidates'); print(src[i-400:i+300])"
```

**Failure:** if the prefetch stage ring has a bug (e.g. the prefetched BF16 rank chunk is consumed one stage early), the suite stays green and it surfaces in production the moment autotuning picks the intended fast path.

**Fix:** parametrize over `use_prefetch in (False, True)`, and change the `(256,64)` case to `tile_k=128, use_prefetch=True` so the pinned tactics are a subset of what `get_valid_tactics` can actually return.

tests/gemm/test_nvfp4_svdquant_gemm.py
```diff
@@ -357,6 +462,423 @@ def test_svdquant_linear_matches_reference(use_bias, rank):
...
+        wq,
+        w_sf.reshape(-1),
+        alpha,
+        pqs,
+        l2t_smoothed,
+        l1_scaled,
+        global_sf,
+        bias=bias,
+    )
+    assert torch_mm_calls == 1
```
dhiraj113, line 963: **Order-dependent assertion — passes in isolation, can fail in file order.**

`svdquant_linear` is called here with the default `backend="auto"`, so `mm_nvfp4_svdquant` takes the tuned path (`gemm_svdquant.py:795-818`) with `custom_op="nvfp4_svdquant_gemm_sm120_auto"` and `runners=[fused, unfused]`.

`AutoTuner` is a process-wide singleton and `profiling_cache` is never reset between tests. `_get_cache_key` is `(custom_op, runner_class_name, hash(runner), nearest_profile, extras)` — and because both runner classes have an empty `__dict__` (see the `enable_pdl` comment above), `hash(runner) == hash(())` for every instance, so entries survive across tests.

`test_mm_nvfp4_svdquant_sm120_autotuned_replay[auto]` (line 582) runs **earlier in the same file** under `autotune(True)` with identical shapes (m=129, n=256, k=256, rank=32). With `use_bias=True` the input-shape tuple matches exactly and `search_cache` hits. If the unfused runner won that profile, `Sm120Nvfp4SvdquantUnfusedRunner.forward` calls `torch.mm(d, l1.T)` (line 487) and this counter becomes 2.

```bash
# passes
pytest tests/gemm/test_nvfp4_svdquant_gemm.py -k test_svdquant_linear_sm120_fused
# can fail with `assert 2 == 1` on an SM120 part where unfused wins the earlier profile
pytest tests/gemm/test_nvfp4_svdquant_gemm.py
```

(`use_bias=False` doesn't collide — the bias slot is `(0,)` vs `(256,)`.)

**Fix:** pass `backend="cute-dsl"` explicitly (which pins the fused-only `custom_op`), use distinct shapes, or clear `AutoTuner.get().profiling_cache` in a fixture.

tests/gemm/test_nvfp4_svdquant_gemm.py
```diff
@@ -357,6 +462,423 @@ def test_svdquant_linear_matches_reference(use_bias, rank):
...
+    ref = residual + down.float() @ lora_b.float().T
+    if bias is not None:
+        ref.add_(bias.float())
+
+    assert out.shape == (m, n) and out.dtype == torch.bfloat16
+    assert _sqnr_db(ref, out.float()) > 35.0
+
+
 @pytest.mark.parametrize("rank", [32, 128])
 def test_mm_nvfp4_svdquant_cuda_graph(rank):
```
dhiraj113, line 990: **Test gap: no SM120 CUDA-graph coverage, and the SM120 path is structurally different.**

This test is `_skip_unless_sm100()` and calls the raw CUTLASS module op directly. The SM120 path differs in kind: every `mm_nvfp4_svdquant` call routes through `AutoTuner.choose_one` (`gemm_svdquant.py:812`), which explicitly raises when profiling begins under an active outer capture (`autotuner.py:1746-1749`), and the SM120 tuning configs set `use_cuda_graph` differently for the fused vs implementation tuners (lines 497-526, with a comment noting capture stays active after a failed tactic).

```bash
# on an SM120 box -- currently untested territory
python - <<'PY'
import torch, flashinfer
g = torch.cuda.CUDAGraph()
with torch.cuda.graph(g):
    ...  # mm_nvfp4_svdquant(...) on SM120
PY
# expect: RuntimeError: AutoTuner profiling cannot begin during an active outer CUDA Graph capture
```

A serving stack that graph-captures an SM120 SVDQuant layer either hits that error or silently bakes in a stale tactic. Worth an SM120 variant that warms up (tunes) outside capture, then captures.

flashinfer/quantization/kernels/nvfp4_quantize.py
```diff
@@ -1449,6 +1489,8 @@ def _nvfp4_kernel_name(
         name += "_host_sf"
     if silu_and_mul:
         name += "_silu"
+    if smooth_quant:
```
dhiraj113, line 1495: **This new codegen parameter walks past the cache-name tripwire test.**

`smooth_quant` was added to `_get_compiled_kernel_nvfp4` (line 1513) and encoded here, but **not** added to `NVFP4_NAME_BASELINE` / `NVFP4_NAME_PERTURBED` in `tests/jit/test_cute_dsl_cache.py:66-88`. `test_nvfp4_kernel_name_varies_with_every_argument` is parametrized over `sorted(NVFP4_NAME_BASELINE)`, so the new parameter is never exercised. The sibling `test_..._signature_covers_codegen_params` only compares parameter *names* between the two signatures, so it passes as soon as the parameter exists in both — it cannot detect one the name function accepts but ignores.

```bash
grep -c smooth tests/jit/test_cute_dsl_cache.py     # -> 0
pytest tests/jit/test_cute_dsl_cache.py -q          # -> 28 passed, none touching smooth_quant
```

The name function *does* encode it correctly today (distinct artifacts land on disk: `swizzled_bfloat16_k256_sf0_pdl1.o` vs `..._pdl1_smooth.o`), so there's no live bug — but that file exists precisely to be a tripwire for this class of mistake.

**Failure it would miss:** a future PR adds another codegen parameter, threads it into `_nvfp4_kernel_name`'s signature, but forgets the `name +=` line. Both tests pass, two specializations share one `.o`, and the second silently runs the first's kernel.

**Fix:** add `"smooth_quant": False` / `True` to the baseline/perturbed dicts. Companion: the new SM120 SVDQuant disk name is an inline f-string inside `_compile_sm120_nvfp4_svdquant` (`gemm_svdquant.py:206-210`) rather than an extracted `_..._kernel_name()` function — the two pre-existing CuTe-DSL disk names are extracted precisely so this file can assert per-argument variation and symbol safety. Extracting it would bring the new one under the same coverage.

tests/trace/fi_trace_out/gemm_bf16_N32_K3072.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "gemm_bf16_N32_K3072",
```
dhiraj113: **Stale golden — this file is not reproducible from `example.py`, and the code path it traces was reverted inside this same PR.**

Git archaeology: added by `5bc56d6a`, which had `down = mm_bf16(x, l2t_smoothed, backend="cudnn") if CUDNN_AVAILABLE else torch.mm(...)`. The next commit `5ddf8bb3` reverted that to plain `torch.mm` (`gemm_svdquant.py:1013`; `test_svdquant_linear_sm120_fused` now asserts exactly one `torch.mm` call), but neither the JSON nor its `tests/trace/example.py:26` docstring entry was removed.

```bash
git log --diff-filter=A --oneline -- tests/trace/fi_trace_out/gemm_bf16_N32_K3072.json
# -> 5bc56d6a Optimize complete SM120 SVDQuant execution

grep -n "mm_bf16" flashinfer/gemm/gemm_svdquant.py    # no hits -- the caller is gone
sed -n '1013p' flashinfer/gemm/gemm_svdquant.py       # down = torch.mm(x, l2t_smoothed)
# example.py's only mm_bf16 loop covers (N,K) in {(4096,4096), (256,7168)}

# the documented regeneration step drops it:
rm -rf tests/trace/fi_trace_out && python tests/trace/example.py
git status --short tests/trace/fi_trace_out/   # -> D gemm_bf16_N32_K3072.json
```

I confirmed this on a B200: regenerating the `mm_bf16` + `svdquant_linear` sections into a clean directory produced `gemm_bf16_N256_K7168.json`, `gemm_bf16_N4096_K4096.json` and the three SVDQuant goldens — and **no** `gemm_bf16_N32_K3072.json`.

Besides breaking the CLAUDE.md checklist step, it advertises to flashinfer-bench a 128x3072x32 `mm_bf16` LoRA-down GEMM the library never issues.

**Fix:** delete this JSON and its `example.py:26` docstring line. Same-commit residue worth reverting: the `svdquant_linear` docstring at `gemm_svdquant.py:965` still says "BF16 tensor-core GEMM" (pre-PR text was "plain BF16 GEMM") though the code is back to `torch.mm`.

For the record, the three *SVDQuant* goldens are clean — I regenerated all three via `fi_trace(...)` with example.py's exact shapes and byte-compared: all match.

tests/gemm/test_nvfp4_svdquant_gemm.py
```diff
@@ -33,6 +31,60 @@
...
+        assert not api.is_backend_supported("cutlass", 120)
+        assert api.is_backend_supported("cute-dsl", 120)
+        assert api.is_backend_supported("cute-dsl", 121)
+        assert not api.is_backend_supported("cute-dsl", 100)
+    assert mm_nvfp4_svdquant.is_backend_supported("cute-dsl-unfused", 120)
+    assert mm_nvfp4_svdquant.is_backend_supported("cute-dsl-unfused", 121)
+    assert not mm_nvfp4_svdquant.is_backend_supported("cute-dsl-unfused", 100)
+
+
+def test_sm120_svdquant_can_implement_rejects_ragged_rank():
```
dhiraj113: **Nit: this test needs `nvidia-cutlass-dsl`, not SM120 hardware — so it runs on no machine in CI.**

It only calls the static classmethod `Sm120B12xBlockScaledDenseGemmKernel.can_implement(...)` and reads `.enable_iket` off constructed (never compiled, never launched) objects. Gating it behind `_skip_unless_sm120()` means it never executes on this repo's PR runners (sm75/sm86/sm90 per `.github/workflows/pr-test.yml`) nor on B200.

```bash
pytest tests/gemm/test_nvfp4_svdquant_gemm.py -k "backend_arch_support or can_implement or sm120" -rs
# on B200: 1 passed, 21 skipped  -- only the arch-table test at line 34 runs anywhere
```

It exists to catch a regression in the rank-granularity guard (`dense_blockscaled_gemm_sm120_b12x.py:2787-2801`), which is exactly the kind of host-side invariant CI *can* check. Swap `_skip_unless_sm120()` for `pytest.importorskip("cutlass")`.

tests/gemm/test_nvfp4_svdquant_gemm.py
```diff
@@ -357,6 +462,423 @@ def test_svdquant_linear_matches_reference(use_bias, rank):
...
+        p["d"],
+        p["l1_scaled"],
+        bias=bias,
+    )
+
+    assert out.data_ptr() == out_buffer.data_ptr()
+    # The fused path accumulates the BF16 rank correction in FP32 before its
+    # single BF16 store, whereas the oracle rounds the residual and correction
+    # in separate launches. Compare numerically, not bitwise.
+    assert _sqnr_db(expected.float(), out.float()) > 35.0
```
dhiraj113: **Nit: the SQNR gates sit ~13-15 dB below the measured noise floor.**

Measured on a B200 with this file's own problem generator: SQNR(fused vs fp32 ref) = 53.4 / 53.4 / 53.6 dB and SQNR(fused vs the bf16 composition oracle) = 49.7 / 49.9 / 50.0 dB for (m,n,k,r) = (129,256,256,32), (129,3072,3072,32), (33,128,128,32). The gates are `> 40.0` (SM100) and `> 35.0` (SM120).

The gate does catch gross errors — I simulated "drop the last rank chunk" (10.3-10.9 dB) and "drop the whole LoRA correction" (4.6-4.8 dB) — so this is tightness, not soundness. But a regression that rounds the LoRA correction to BF16 before accumulating (instead of the FP32 accumulate the comment at 541-543 relies on) lands at ~40-45 dB and passes.

SQNR is also a whole-matrix mean: at m=6912, n=3072 (21.2M elements), four garbage elements still yield ~67 dB.

**Suggestion:** raise the SM120 bar to ~45 dB with a comment recording the measured floor, and add a bounded elementwise check (`torch.testing.assert_close` with an explicit atol tied to `out.abs().max()`) so localized corruption is caught too.

flashinfer/quantization/kernels/nvfp4_quantize.py
```diff
@@ -2001,6 +2055,68 @@ def nvfp4_quantize_cute_dsl(
...
+        raise ValueError(f"K ({k}) must be divisible by {NVFP4_SF_VEC_SIZE}")
+    if (
+        pre_quant_scale.dtype != torch.bfloat16
+        or not pre_quant_scale.is_cuda
+        or pre_quant_scale.numel() != k
+    ):
+        raise ValueError("pre_quant_scale must be a CUDA BF16 tensor with K elements")
+
+    input = input.contiguous()
+    pre_quant_scale = pre_quant_scale.reshape(k).contiguous()
```
dhiraj113, line 2094: **Nit: `.contiguous()` is a no-op on an already-contiguous *slice*, so a storage-offset `pre_quant_scale` keeps its original alignment.**

`process_nvfp4_block_bfloat_smooth` loads this with `ld_global_v4_u32` (a 16-byte vector load) and the compile-time fake at lines 1563-1565 declares `assumed_align=16`. A caller passing e.g. `big_bf16[1:1+k]` (contiguous, correct dtype, correct numel — passes every existing check) keeps a 2-byte-aligned pointer.

```bash
python - <<'PY'
import torch
from flashinfer.quantization.kernels.nvfp4_quantize import nvfp4_quantize_smooth_cute_dsl
k = 256
base = torch.randn(k + 8, dtype=torch.bfloat16, device="cuda")
pqs  = base[1:1+k]                      # contiguous, but 2-byte aligned
x    = torch.randn(129, k, dtype=torch.bfloat16, device="cuda")
gs   = torch.tensor([1/(448*6)], dtype=torch.float32, device="cuda")
nvfp4_quantize_smooth_cute_dsl(x, pqs, gs)
PY
# ValueError: Misaligned Tensor data on argument #7 ... expected data alignment=16 bytes
```

Two qualifications, so this is weighted correctly: it surfaces as a clean **pre-launch** Python exception from the TVM-FFI arg check (no silent corruption, no illegal access), and it is **not a regression** — the sibling `nvfp4_quantize_cute_dsl` has the same hole, and the pre-existing CUTLASS backend of the same public API fails *worse* (async `CUDA error: misaligned address`, which poisons the context).

So this is a pre-existing API-contract gap the new path inherits. The best home for a fix is probably `_check_nvfp4_quantize_smooth_problem` (`gemm_svdquant.py:852`) so both backends are covered — or mirror the one-line clone guard that `silu_and_mul_nvfp4_quantize_cute_dsl` already has at line 2190-2192.

flashinfer/gemm/gemm_svdquant.py
```diff
@@ -208,18 +732,19 @@ def mm_nvfp4_svdquant(
...
-    NVFP4 K-loop, plus an optional fused per-column bias. The LoRA rank ``r`` is inferred
-    from the ``d``/``l1`` shapes and must be a positive multiple of 32 (ranks 32-128 are
-    validated). ``1/alpha`` must be folded into ``l1`` by the caller
-    (``l1 = svdquant_lora_b / alpha``) so the epilogue ``out = alpha * acc + bias`` yields
-    the correction unscaled.
+    On SM100/SM103, CUTLASS fuses the block-scaled NVFP4 residual GEMM with the rank-r
+    BF16 LoRA-up correction and optional bias. On SM120/SM121, ``"cute-dsl"`` fuses the
+    correction and bias into the b12x CuTe DSL kernel's FP32 accumulator epilogue, while
+    ``"cute-dsl-unfused"`` retains the compositional implementation as a differential
+    oracle and fallback. The LoRA rank ``r`` is inferred from the ``d``/``l1`` shapes and must
```
dhiraj113: **Nit: `"cute-dsl-unfused"` is documented as a "fallback", but there is no fallback path.**

In the `auto` branch the unfused runner is only an autotune *candidate*. With autotuning off — the default — `choose_one` returns `runners[0]` (fused) with `tactic=-1`, and if the fused compile or launch raises, the exception propagates. Nothing catches it and retries unfused.

Similarly the `"auto"` bullet at line 780 ("autotunes across both...") is only true inside `autotune(True)`.

Suggest dropping "fallback" and qualifying the auto bullet, so callers don't rely on resilience that isn't there.

flashinfer/trace/templates/gemm.py
```diff
@@ -2042,16 +2046,21 @@ def _mm_nvfp4_svdquant_init(
...
     axes={
         "M": Var(),
         "N": Const(),
         "K_packed": Const(description="K / 2 (two e2m1 values per byte)."),
-        "SF_A": Const(description="128x4-swizzled activation scale buffer size."),
-        "SF_B": Const(description="128x4-swizzled weight scale buffer size."),
+        "SF_A": Var(
+            description="128x4-swizzled activation scale buffer size derived from M and K."
+        ),
+        "SF_B": Var(
```
dhiraj113, line 2210: **Nit: `SF_B` is now `Var` here but stays `Const` in the sibling `svdquant_linear_trace` (line 2234), so the two committed goldens disagree about the same buffer.**

The `linear_*` golden carries `SF_B589824` in its name; the `gemm_*` one doesn't.

To be clear about what is *not* wrong here, since it took some digging: promoting `SF_A` to `Var` is a genuine fix (it tracks the `Var` axis `M`). And "Var axis pinned by a `constraints` entry" is the dominant house convention — 19 pre-existing instances, including `nvfp4_quantize_smooth_trace` in this very file (`N: Const`, `N_half: Var`, `constraints=["N_half == N // 2"]`). Filenames can't collide either, since `N` remains `Const` and `definition_name` walks Const axes (verified: N=2048/3072/4096 -> distinct names), and the new `constraints=` entry is emitted into the JSON.

So this is purely a cross-template **consistency** nit: `SF_B` derives only from the two `Const` axes `N` and `K_packed`. Either representation is defensible — they should just match.

flashinfer/gemm/gemm_svdquant.py
rosenrodt: Thanks. Added `extra_key_files=_svdquant_kernel_source_files()` where it returns
```
   (
        __file__,
        dense_blockscaled_gemm_sm120_b12x.__file__,
        cute_dsl_utils.__file__,
    )
```

flashinfer/gemm/gemm_svdquant.py
rosenrodt: Kept `alpha.numel() > 1`. But SVD Quant linear would just use `alpha.reshape(-1)[:1]` as suggested

flashinfer/quantization/kernels/nvfp4_quantize.py
rosenrodt: Adopted early return to return empty tensor

flashinfer/gemm/gemm_svdquant.py
rosenrodt: Good catch. Added `XXXRunner.get_cache_key_extras()` to return `(enable_pdl,)`

tests/gemm/test_nvfp4_svdquant_gemm.py
rosenrodt: Added

```
@pytest.mark.parametrize("m,k", [(129, 256), (129, 12288), (70, 3088)])
def test_nvfp4_quantize_smooth_sm120_cute_dsl(m, k):
  ...
```

tests/gemm/test_nvfp4_svdquant_gemm.py
rosenrodt: I chose the other direction. I completely removed `use_prefetch` from list returned by `get_valid_tactics()`. the backing `Sm120B12xBlockScaledDenseGemmKernel` does not implement prefetch at all so it is always false.

tests/gemm/test_nvfp4_svdquant_gemm.py
rosenrodt: Pinned the test to the fused CuTe DSL backend.
```
    out = svdquant_linear(
        ...,
        backend="cute-dsl",
    )
    assert torch_mm_calls == 1
```

tests/gemm/test_nvfp4_svdquant_gemm.py
rosenrodt: Added the test
```
@pytest.mark.parametrize("backend", ["cute-dsl", "auto"])
def test_mm_nvfp4_svdquant_sm120_cuda_graph_replay(...)
  ...
```

tests/trace/fi_trace_out/gemm_bf16_N32_K3072.json
rosenrodt: Removed the obsolete definition through trace regeneration.

tests/gemm/test_nvfp4_svdquant_gemm.py
rosenrodt: Replaced `_skip_unless_sm120()` with `pytest.importorskip("cutlass")`

tests/gemm/test_nvfp4_svdquant_gemm.py
rosenrodt: Now `assert _sqnr_db(ref_f32, got_f32) > 45.0`

flashinfer/quantization/kernels/nvfp4_quantize.py
rosenrodt: Added test to cover the said `smooth_quant` case 

```
@pytest.mark.parametrize(
    "param,alternate",
    [
        ...
        pytest.param("smooth_quant", True, id="smooth_quant"),
    ],
)
def test_nvfp4_kernel_name_varies_with_every_argument(param, alternate):
  ...
```

flashinfer/quantization/kernels/nvfp4_quantize.py
rosenrodt: .clone() when data_ptr() is not 16-byte aligned.

flashinfer/gemm/gemm_svdquant.py
rosenrodt: Revised wording

flashinfer/trace/templates/gemm.py
rosenrodt: both use `Var` now

dhiraj113:
@flashinfer-bot run

dhiraj113 (approved):

rosenrodt:
@flashinfer-bot run

dhiraj113:
@flashinfer-bot run

PR #3522, opened 2026-06-06
perf: cache cudaGetDeviceProperties in CudaDevice (fmha_v2)

aws-jiadingg:
## 📌 Description

The `CudaDevice` struct (`csrc/fmha_v2/fused_multihead_attention_utils.h`) calls `cudaGetDeviceProperties` in its constructor, and a `CudaDevice` is constructed on every `fmha_v2_run` / `TRTLLMFMHAv2Run` call. `cudaGetDeviceProperties` is synchronous and costs ~1.7 ms/call on H200 (CUDA 12.8), so this was paid on every attention forward.

This PR caches device-0 properties once via a function-local `static const` (thread-safe initialization per C++11), dropping per-construction cost to a struct copy. Device-0 properties don't change at runtime and only scalar fields (`major`, `minor`, `multiProcessorCount`, `l2CacheSize`) are read downstream, so the change is behavior-preserving.

## 🔍 Related Issues

Follow-up in the spirit of #2509 (removed a per-call `cudaGetDeviceProperties` overhead in the GDN prefill launcher).

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.
> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [x] Tests have been added or updated as needed.
- [ ] All tests are passing (`unittest`, etc.).

## Reviewer Notes

Behavior-preserving: only device 0 is queried (hard-coded, as before), so there is no multi-GPU behavior change, and it is covered by existing fmha_v2 tests. Happy to switch to per-field `cudaDeviceGetAttribute` instead of caching the full `cudaDeviceProp` if preferred.

## Summary by CodeRabbit

* **Performance**
  * Optimized device initialization to reduce redundant operations and improve startup performance.

PR #2509, opened 2026-02-06
perf: cache cudaGetDeviceProperties in gdn_prefill to avoid per-call overhead

xutizhou:
## Summary
- Cache `device_properties.major` and `device_properties.multiProcessorCount` using
  `static thread_local` variables in `gdn_prefill_launcher()` and `gdn_prefill()`
  respectively, instead of calling `cudaGetDeviceProperties` on every invocation.
- Each `cudaGetDeviceProperties` call adds ~1ms of synchronous CPU overhead, which
  accumulates to ~72ms across 36 GDN layers during prefill. This change eliminates
  that repeated overhead.
## Changes
- **`csrc/gdn_prefill_launcher.cu`**: Replace per-call `cudaGetDeviceProperties` with
  `static thread_local` cached values for `device_major` (used for SM architecture
  dispatch) and `sm_count` (used for kernel launch configuration).

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [x] Tests have been added or updated as needed.
- [x] All tests are passing (`unittest`, etc.).

## Summary by CodeRabbit

* **Refactor**
  * Streamlined GPU device attribute retrieval mechanisms for accessing compute capability and processor information.

PR #4422, opened 2026-08-08
perf(attention): replace cudaGetDeviceProperties in nvfp4 sm120 forward

samuelkim7:
## 📌 Description

`fwd()` calls `cudaGetDeviceProperties` on every forward. It only needs `major` and `minor` for the SM120 check. That call costs about 1.7 ms on an RTX PRO 6000.

I replaced it with two `cudaDeviceGetAttribute` calls. The check is the same. The launcher this binding calls already uses `cudaDeviceGetAttribute` (`api/launcher.h:112`).

Measured on an RTX PRO 6000 Blackwell Server Edition (sm_120), CUDA 12.9.1, driver 580.95.05:

| shape (b, h, s, d) | host before | host after | wall before | wall after | speedup |
|---|---|---|---|---|---|
| 1, 8, 128, 128 | 1788 us | 14.6 us | 2340 us | 14.8 us | 158x |
| 1, 32, 1024, 128 | 1781 us | 14.4 us | 1804 us | 35.4 us | 51x |
| 4, 32, 4096, 128 | 1879 us | 14.2 us | 2137 us | 1287 us | 1.66x |

`host` is the time the Python call takes to return. `wall` is 50 back-to-back forwards with one sync at the end, per iteration.

The 14 us left over is Python argument checking. Outputs are bit-identical before and after.

The win depends on the kernel being short. On the 4096 shape the kernel itself runs about 1.3 ms, which is why the speedup collapses there.

## 🔍 Related Issues

N/A. Same class of fix as #3522 and #2509, which removed per-call `cudaGetDeviceProperties` from fmha_v2 and the GDN prefill launcher.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

I ran `tests/attention/test_nvfp4_attention_sm120.py` on an RTX PRO 6000 (sm_120). 13 passed, 0 skipped, both before and after the change. I did not add a test. The check is unchanged, so the existing suite already covers it.

## Reviewer Notes

I did not cache the result. The two attribute calls cost 29 ns here, so a cache would not save anything measurable before a kernel launch. The other compute-capability gates that take a device_id do the same two uncached calls: `CheckFlashKDATarget` (`csrc/kda/flashkda_binding_common.cuh:136`) and `CheckSm100Family` (`csrc/tinygemm2_sm100.cu:1409`). This makes the SM120 gate match them.

#2509 was opened as a `static thread_local` cache and switched to `cudaDeviceGetAttribute` before it merged. #3522 does cache, but it caches a whole `cudaDeviceProp` for four fields, on a path that already assumed device 0.

A cache here would also have to be per-device, because `fwd()` has `q_fp4.device().device_id`. That is about six lines of static state. I can add it if you prefer.

`props` was a local in `fwd()`, and was never passed to `set_params_fprop` or the launcher.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GPU capability detection during attention processing.
  * Added clearer, attribute-specific error handling when device capability information cannot be retrieved.
  * Preserved validation for supported compute capabilities 12.0 and 12.1.


samuelkim7:
@coderabbitai summary

PR #4431, opened 2026-08-10
fix(moe): use per-expert Humming residual scales in SM90 CUTLASS MXFP4xFP8 MoE backend

StudyingShao:
## 📌 Description

Humming weight preprocessing naturally produces one residual scale per expert.
However, the CUTLASS fused MoE Humming interface previously required
quant-scale slots 1 and 4 to contain one value per routed row (`M * top_k`),
already reordered into FlashInfer's internal expert-permuted order. This leaked
an internal runtime layout into the API and forced callers to duplicate the
token-to-expert permutation before every fused MoE invocation.

This PR changes the two Humming residual inputs to use rank-local expert order:

- Slot 1: FC1 residual scale with shape `[num_local_experts]`.
- Slot 4: FC2 residual scale with shape `[num_local_experts]`.

The fused MoE runtime reuses its existing expert-permuted token map to select
the correct local-expert residual for each routed row:

- FC1 performs the lookup while expanding and dynamically quantizing routed
  input rows.
- FC2 performs the lookup while applying the activation and dynamically
  quantizing the GEMM1 output.
- In both cases, the selected residual is folded into the row's dynamic FP8
  activation dequantization scale before the combined scale is consumed by the
  GEMM epilogue.

Routing inputs may continue to use global expert IDs. FlashInfer's existing MoE
permutation converts valid routes to rank-local expert indices before FC1 and
FC2 read the residual arrays.

This contract lets downstream integrations pass the preprocessing residuals
directly after applying Humming's existing fixed `2^6` compensation. They no
longer need to materialize two `[M * top_k]` FP32 residual tensors or run the
residual-specific pointwise, gather, stable `argsort`, and permutation
operations outside FlashInfer.

Additional changes include:

- Validate all five Humming quant-scale inputs as CUDA, contiguous tensors with
  the expected dtype.
- Validate slots 1 and 4 against `num_local_experts` instead of the routed-row
  count.
- Size the corresponding profiler buffers per local expert rather than per
  routed row.
- Document the per-local-expert contract and fixed `2^6` compensation in
  `cutlass_fused_moe` and the Humming preprocessing API.
- Update the Humming end-to-end test to use distinct per-expert residuals and
  different FC1/FC2 residual values.
- Add an EP rank-1 regression case that verifies global routed expert IDs map
  to the correct rank-local residual entries.

## 🔍 Related Issues

N/A

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used
  my preferred method).
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed
  any reported issues.

> If you are unsure about how to set up `pre-commit`, see the
> [pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [x] Full repository pre-commit checks:

  ```bash
  pre-commit run -a
  ```

- [x] A representative Humming runtime correctness case was run on NVIDIA H100
  PCIe during implementation:

  ```bash
  python -m pytest -q -x \
    'tests/moe/test_trtllm_cutlass_fused_moe.py::test_moe_fp8_mxfp4_humming_prescale_hopper_correctness[False-small]'
  ```

  Result: `1 passed`.

The updated regression coverage exercises distinct per-expert residuals,
different FC1 and FC2 residuals, and non-zero EP rank global-to-local expert
mapping.

## Reviewer Notes

- This changes the Humming-specific quant-scale contract. Callers must update
  slots 1 and 4 together with FlashInfer: each slot is now a contiguous FP32
  `[num_local_experts]` tensor in the same local order as the rank's weights.
- Callers must still multiply both preprocessing residual arrays by Humming's
  fixed `2^6` compensation before passing them to `cutlass_fused_moe`.
- Slot 2 remains required by the five-slot API and continues to accept a scalar
  or one value per local expert.
- The expert map is passed to FC1 and FC2 only for the Humming path because
  other fused MoE prologue paths do not always populate this buffer.
- The standalone GEMM1 profiler interface does not provide an expert map. Its
  post-GEMM1 activation output is not consumed by the independently profiled
  GEMM2 path, so profiling does not require a synthetic routed-row residual
  tensor.
- Generic non-Humming row operations retain the existing
  `expert_first_token_offset` lookup fallback when no expert map is available.
- The focused EP rank-1 case skips duplicate autotune coverage; the other
  Humming cases continue to exercise autotuning.

## Summary by CodeRabbit

- **New Features**
  - Added expert-specific residual scaling for FP8 and MXFP4 mixture-of-experts workloads.
  - Enhanced routed and distributed expert execution, including Humming mixed-precision paths.
  - Improved MXFP8 quantization scale allocation and initialization.

- **Bug Fixes**
  - Corrected activation dequantization and residual handling for expert-routed tokens.
  - Improved scale mapping across local and global expert IDs.

- **Documentation**
  - Clarified Humming quantization scale requirements and runtime routing behavior.

- **Tests**
  - Expanded coverage for expert routing, distributed execution, and per-expert quantization scales.


samuellees (approved):
LGTM

PR #4441, opened 2026-08-10
Per Token quantization for NVFP4 in mm_fp4

b8zhong:
## 📌 Description

Need this for online quant of dense layers of MoE and dense models.

Only add it to cute-dsl nvfp4 and cutedsl nvfp4 quantize

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * Added per-token scaling support for FP4 GEMM, including scalar, row-wise, and column-wise scaling where supported.
  * Added optional output-scale folding for per-token NVFP4 quantization.
  * Extended support across compatible hardware, backends, autotuning, kernel caching, and precompilation.

* **Bug Fixes**
  * Improved validation for unsupported hardware, backends, batching, and scaling combinations.
  * Kept cached kernel variants distinct across scaling modes.

* **Tests**
  * Added coverage across supported shapes, data types, backends, and autotuning modes.


aleozlx (commented):
looks good.

i'll let Dhiraj take another look

leejnau:
> looks good.
> 
> i'll let Dhiraj take another look

cc @dhiraj113

b8zhong:
Fixed a merge conflict and documentation issue.

flashinfer/gemm/gemm_base.py
```diff
@@ -4992,6 +4993,18 @@ def _get_sm100_block_scaled_tactics(
     return valid_tactics
 
 
+# FP4 backends with a per-row alpha epilogue, by the compute capability major
+# that builds it.
+_PER_TOKEN_ALPHA_BACKENDS_BY_SM_MAJOR = {
+    10: ("cute-dsl",),
```
dhiraj113, line 4999: Is this API now? Can this be be avoided

tests/gemm/test_mm_fp4.py
```diff
@@ -14,7 +14,10 @@
     version_at_least,
     LibraryError,
 )
-from flashinfer.gemm.gemm_base import CUDNN_FP4_MXFP4_SM120_CUDNN_VERSION_ERROR
+from flashinfer.gemm.gemm_base import (
+    CUDNN_FP4_MXFP4_SM120_CUDNN_VERSION_ERROR,
+    _PER_TOKEN_ALPHA_BACKENDS_BY_SM_MAJOR,
```
dhiraj113, line 19: can importing this be avoided?

tests/gemm/test_mm_fp4.py
```diff
@@ -275,5 +278,83 @@ def test_mm_fp4_cute_dsl_misaligned_n_raises():
...
+@pytest.mark.parametrize("res_dtype", [torch.bfloat16, torch.float16])
+@pytest.mark.parametrize("backend", ["cute-dsl", "auto"])
+@pytest.mark.parametrize("auto_tuning", [False, True])
+def test_mm_fp4_per_token_alpha(m, n, k, res_dtype, backend, auto_tuning):
+    """A per-token alpha must scale each output row by its own dequant scale."""
+    major, minor = get_compute_capability(torch.device("cuda"))
+    arch_backends = _PER_TOKEN_ALPHA_BACKENDS_BY_SM_MAJOR.get(major, ())
+    if backend == "auto":
+        if not arch_backends:
+            pytest.skip(f"No per-token alpha FP4 backend on SM{major}{minor}.")
```
dhiraj113, line 293: shouldn't this skip / error reporting come from the implementation.

tests/gemm/test_mm_fp4.py
```diff
@@ -275,5 +278,83 @@ def test_mm_fp4_cute_dsl_misaligned_n_raises():
...
+    elif backend not in arch_backends:
+        pytest.skip(f"{backend} has no per-token alpha epilogue on SM{major}{minor}.")
+
+    torch.manual_seed(0)
+    a = torch.randn([m, k], device="cuda", dtype=torch.bfloat16)
+    b = torch.randn([n, k], device="cuda", dtype=torch.bfloat16)
+    g_in = (448 * 6) / a.float().abs().nan_to_num().max()
+    g_w = (448 * 6) / b.float().abs().nan_to_num().max()
```
dhiraj113, line 301: Can you explan what is happening in this scale logic.

dhiraj113 (commented):
Nice piece of work -- the scalar-vs-per-token cross-check in the new test is the right way to validate a per-row epilogue, and the `out_scale` fold is a genuinely clean way to collapse the pipeline. I verified the fold is numerically exact on a B200 across all three `sf_layout` values at `m` in {1, 17, 130}, and `pre-commit` is clean on all 7 files.

One blocker and one test gap:

1. **`backend="auto"` + per-token alpha bypasses `_cute_dsl_gemm_fp4_requirement` entirely.** Reachable from the documented public API; crashes with a `TypeError` instead of the clear `ValueError` the requirement function already defines.
2. **`out_scale` / `fold_out_scale` has no numerical test** -- only a cache-key naming assertion. Half the feature surface is unguarded in CI.

Everything else is polish. Detail inline.

flashinfer/gemm/gemm_base.py
```diff
@@ -6886,10 +6947,30 @@ def mm_fp4(
...
+    # A backend without a per-row epilogue would silently consume alpha[0] for
+    # every row, so narrow the candidates instead of letting the tuner pick one.
+    per_token_alpha = _is_per_token_alpha(alpha)
+    if per_token_alpha:
+        arch_backends = _PER_TOKEN_ALPHA_BACKENDS_BY_SM_MAJOR.get(major, ())
+        supported = [bk for bk in backends if bk in arch_backends]
+        if not supported and backend == "auto":
+            # cute-dsl is never in the auto list, yet on SM10X it is the only
+            # backend with a per-row epilogue.
+            supported = list(arch_backends)
```
dhiraj113, line 6959: **Blocker.**

`_heuristic_func_mm_fp4` only ever returns backends from `("b12x", "cutlass", "cudnn")` -- `"cute-dsl"` is never in `mm_fp4.suitable_auto_backends`. So on the `backend="auto"` path the filter two lines up is **always** empty and this fallback fires unconditionally, handing execution to cute-dsl without `_cute_dsl_gemm_fp4_requirement` ever running.

That skips every guard it defines: `N % 8 == 0`, the `use_8x4_sf_layout` rejection, `_check_cute_dsl_availability()`, and its `@supported_compute_capability([100, 103])` gate.

Repro on a B200 (SM100) with `n = 252`:

```
requirement fn verdict for this shape:
  -> returns False  ('do not use cute-dsl')
actual mm_fp4(backend='auto') call:
  -> TypeError: cannot unpack non-iterable NoneType object
```

`get_valid_tactics` returns `[]` (its own `n_aligned` guard), the tuner falls back to `tactic=-1`, and `_compute_tactic_for_m` returns `None`, so the unpack at line 6322 crashes. Passing a *scalar* alpha for the same shape succeeds via cutlass, which isolates the failure to this narrowing.

The requirement function already produces exactly the right message here (`"CuTe-DSL FP4 GEMM requires N % 8 == 0, got n=252"`); it just never gets asked. Suggest gating the fallback:

```python
if not supported and backend == "auto":
    supported = [
        bk for bk in arch_backends
        if mm_fp4.is_backend_supported(bk, major * 10 + minor)
    ]
```

That fixes the capability gate; catching the shape guards (`N % 8`, 8x4) needs the per-backend requirement function called directly.

flashinfer/gemm/gemm_base.py
```diff
@@ -4992,6 +4993,18 @@ def _get_sm100_block_scaled_tactics(
     return valid_tactics
 
 
+# FP4 backends with a per-row alpha epilogue, by the compute capability major
+# that builds it.
+_PER_TOKEN_ALPHA_BACKENDS_BY_SM_MAJOR = {
```
dhiraj113, line 4998: Keying on **major** alone over-claims: `major == 10` also matches SM107, but `_cute_dsl_gemm_fp4_requirement` is `@supported_compute_capability([100, 103])`.

```
cc    table claims cute-dsl    requirement supports
100   True                     True
103   True                     True
107   True                     False   <-- mismatch
```

Combined with the unconditional fallback at line 6959, SM107 gets forced onto a backend that declares itself unsupported there. Suggest keying on the full capability (`major * 10 + minor`) so this table can't drift from the requirement declarations.

flashinfer/quantization/kernels/nvfp4_quantize.py
```diff
@@ -2276,17 +2323,19 @@ def nvfp4_quantize_per_token_cute_dsl(
     )
     nvfp4_4over6_config = current_nvfp4_4over6_config()
 
+    fold_out_scale = out_scale is not None
```
dhiraj113, line 2326: **Test gap.** `fold_out_scale` has no numerical test anywhere:

```
$ grep -rn "fold_out_scale" tests/ --include=*.py
tests/jit/test_cute_dsl_cache.py:76:    "fold_out_scale": False,
tests/jit/test_cute_dsl_cache.py:88:    "fold_out_scale": True,
```

Both are cache-key *naming* assertions. The fold math, and the `torch.zeros` -> `torch.empty` plus padding-row-fill change that goes with it, are entirely unguarded.

I verified both are correct today on a B200, so this is coverage, not a bug:

```
layout=128x4   m=  1  max|fold - 3*base| = 0.000e+00   scale buf all-finite=True
layout=128x4   m= 17  max|fold - 3*base| = 0.000e+00   scale buf all-finite=True
layout=128x4   m=130  max|fold - 3*base| = 0.000e+00   scale buf all-finite=True
layout=8x4     ... 0.000e+00 ...
layout=linear  ... 0.000e+00 ...
```

Worth pinning: `per_token_scale(out_scale=s) == s * per_token_scale()` across all three `sf_layout` values, plus an assertion that the padding rows of `scale_output` are actually zeroed now the buffer is uninitialised. `m = 1` with `128x4` matters most -- 127 of 128 rows are padding, so a regression in the new `row_idx >= M` branch would feed the GEMM pure garbage.

flashinfer/gemm/gemm_base.py
```diff
@@ -6320,9 +6346,17 @@ def forward(
...
-            cache_key = _mm_fp4_cache_key(sf_vec_size, tactic, enable_pdl, out_dtype)
+            alpha_mode = per_token_alpha_mode(per_token_alpha, swap_ab)
+            cache_key = _mm_fp4_cache_key(
+                sf_vec_size, tactic, enable_pdl, out_dtype, alpha_mode
+            )
 
             if kernel_type == "sm103" and Sm103Kernel is not None:
+                if alpha_mode is not None:
+                    raise ValueError(
+                        "The SM103 3xFP4 CuTe-DSL kernel has no per-token alpha "
```
dhiraj113, line 6357: Minor: on SM103 with a per-token alpha, `get_valid_tactics` still emits `"sm103"` tactics and every one lands on this raise. The autotuner catches per-tactic exceptions and scores them `inf`, and `_compute_tactic_for_m` only ever returns `"sm100"`, so the result is correct -- but each tuning pass burns profiling time on tactics guaranteed to fail.

Cheaper to filter them out in `get_valid_tactics` when `per_token_alpha` is set, keeping this raise as a backstop.

flashinfer/gemm/gemm_base.py
```diff
@@ -6886,10 +6947,30 @@ def mm_fp4(
...
+                f"Per-token alpha is implemented on SM{major}{minor} only by "
+                f"{sorted(arch_backends)}, got backends {backends}."
+            )
+        backends = supported
+
     tuner = AutoTuner.get()
-    tuning_config = (
-        _MM_FP4_TUNING_CONFIG_8x4 if use_8x4_sf_layout else _MM_FP4_TUNING_CONFIG_128x4
-    )
+    if use_8x4_sf_layout:
```
dhiraj113, line 6968: Minor / latent: `if use_8x4_sf_layout` / `elif per_token_alpha` means an 8x4 + per-token call selects `_MM_FP4_TUNING_CONFIG_8x4`, which has no alpha `ConstraintSpec` -- exactly the OOB the new config was added to prevent:

```
_MM_FP4_TUNING_CONFIG_128x4                  constrains alpha (idx 4)? False
_MM_FP4_TUNING_CONFIG_8x4                    constrains alpha (idx 4)? False
_MM_FP4_TUNING_CONFIG_128x4_PER_TOKEN_ALPHA  constrains alpha (idx 4)? True
```

Currently unreachable -- I confirmed the combination dies earlier with `BackendSupportedError: No suitable auto backends found for mm_fp4`, since no auto backend accepts 8x4 -- so latent rather than live. Still a trap if an 8x4-capable per-token backend ever lands. Checking `per_token_alpha` first, or asserting the two are mutually exclusive, would close it.

flashinfer/gemm/gemm_base.py
```diff
@@ -6726,6 +6766,23 @@ def _mxfp8_swizzled_scale_len(m: int, k: int, swizzle_layout: SfLayout) -> int:
 )
 
 
+# Alpha has to follow the M the tuner picks for its profiling shapes, or the
+# kernel reads past the caller's alpha while profiling a larger bucket.
+_MM_FP4_TUNING_CONFIG_128x4_PER_TOKEN_ALPHA = TuningConfig(
```
dhiraj113, line 6771: Nit: this re-lists 4 of `TuningConfig`'s 8 fields. The other four -- `tensor_initializers`, `value_aware_input_indices`, `profile_arena_input_indices`, `inputs_pre_hook` -- are all at their defaults on the base config today, so nothing diverges yet. But the moment someone sets one on `_MM_FP4_TUNING_CONFIG_128x4`, the per-token variant silently won't get it.

`dataclasses.replace` expresses the intent and can't drift:

```python
_MM_FP4_TUNING_CONFIG_128x4_PER_TOKEN_ALPHA = dataclasses.replace(
    _MM_FP4_TUNING_CONFIG_128x4,
    constraint_specs=(
        *_MM_FP4_TUNING_CONFIG_128x4.constraint_specs,
        ConstraintSpec(4, 0, lambda shapes: shapes[0][0]),
    ),
)
```

flashinfer/gemm/gemm_base.py
```diff
@@ -5881,7 +5892,19 @@ def _check_mm_fp4_problem_size(
     if alpha is not None and alpha.dtype != torch.float:
         raise ValueError(f"alpha must be a float tensor, got {alpha.dtype}")
     if alpha is not None and alpha.numel() != 1:
-        raise ValueError(f"alpha must be a scalar, got {alpha.numel()}")
+        if alpha.numel() != a.shape[0]:
+            raise ValueError(
+                "alpha must be a scalar, or one scale per row of a for the "
+                f"per-token path. Got {alpha.numel()} for m={a.shape[0]}."
+            )
+        per_token_backends = sorted(
```
dhiraj113, line 5900: Nit: this unions across **all** SM majors, so the message can name backends that don't exist on the caller's device. On SM100 it reads correctly:

```
per-token alpha is only supported by the ['cute-dsl'] backends, got backend='cutlass'.
```

but on e.g. SM120 the same string would still recommend `cute-dsl`, which isn't available there at all. `mm_fp4` already computes the arch-specific set a few hundred lines down -- reusing it would keep the two messages consistent.

flashinfer/gemm/kernels/dense_blockscaled_gemm_sm100.py
```diff
@@ -1326,6 +1376,27 @@ def kernel(
...
+                            token = (
+                                coord[0]
+                                if cutlass.const_expr(self.per_token_alpha == "m")
+                                else coord[1]
+                            )
+                            # Elements past the token extent are dropped by the
+                            # TMA store, so clamping the index just avoids
+                            # reading alpha out of bounds.
+                            token = cutlass.min(token, alpha_extent - 1)
+                            tTR_rAcc[i] = tTR_rAcc[i] * alpha[token].to(cutlass.Float32)
```
dhiraj113, line 1397: Question, not a blocker: this is one gmem load per accumulator element, and the comment above explains why there's no sub-mode to hoist it out of -- understood.

The PR reports 10.55us -> 7.75us at bs=1, which is dominated by the two eliminated launches rather than by this epilogue. Do you have per-token-vs-scalar-alpha numbers on the *same* shape at larger M, where the epilogue cost stops being amortised away? If the delta is material, staging `alpha` through smem once per tile is the obvious alternative; if it's negligible, a one-line note saying so would help, since the loop looks expensive at a glance.

flashinfer/quantization/fp4_quantization.py
```diff
@@ -1375,6 +1376,10 @@ def nvfp4_quantize(
         ``a_global_sf`` is the inverse base scale multiplier (typically
         ``1 / (448 * 6)``) and the function also returns per-token FP32
         scales.
+    out_scale : torch.Tensor, optional
```
dhiraj113, line 1379: Nit: docstring order doesn't match the signature.

```
signature tail : ['per_token_activation', 'expanded_idx_to_permuted_idx', 'out_scale']
docstring order: ['out_scale', 'expanded_idx_to_permuted_idx']
```

Moving this block below `expanded_idx_to_permuted_idx` keeps them aligned.

tests/gemm/test_mm_fp4.py
```diff
@@ -275,5 +278,83 @@ def test_mm_fp4_cute_dsl_misaligned_n_raises():
         )
 
 
+@pytest.mark.parametrize("m", [4, 17, 48, 128, 257])
```
dhiraj113, line 281: Nit on CI cost: 5 x 2 x 2 x 2 x 2 x 2 = 160 cases, each compiling and (for `auto_tuning=True`) profiling CuTe-DSL kernels. A 5-case subset took 107s on a B200, so the full matrix is roughly an hour of CI for this one function.

The cross-check assertion is the valuable part and doesn't need the full cross product -- collapsing `k` to a single value and `n` to one or two would cut this 2-4x with little loss of signal. Worth keeping the `m` sweep in full, since that's the axis the per-row indexing actually varies over.

(Separately: the scalar-vs-per-token comparison here is a really nice way to make the quantisation error common-mode so it cancels. Worth keeping.)

aleozlx (approved):

aleozlx:
Thanks @b8zhong for the feature and @dhiraj113 for the review. To get this landed I have carried the branch into #5504 (your commits, authorship preserved, plus one follow-up commit), since this PR does not allow maintainer edits. Follow-up mapping to the open threads:

- **Blocker (`gemm_base.py`, auto fallback bypassing the requirement):** `_heuristic_func_mm_fp4` now returns `cute-dsl` for a per-token alpha and `_cute_dsl_gemm_fp4_requirement` gates the per-token path on the full compute capability (SM100/SM103). The unconditional fallback in `mm_fp4` is gone; only a backstop for `skip_check=True` remains. Pinned by `test_mm_fp4_per_token_alpha_auto_misaligned_n_raises`.
- **SM-major keying (`_PER_TOKEN_ALPHA_BACKENDS_BY_SM_MAJOR`, "is this API now?"):** table removed; the requirement function is the single source of truth and the test no longer imports it.
- **`fold_out_scale` test gap:** `test_nvfp4_quantize_per_token_out_scale_fold` covers all three layouts at m in {1, 17, 130}, asserts `per_token_scale(out_scale=s) == s * per_token_scale()` exactly, and checks the padding rows of the `torch.empty` scale buffer against the zero-padded reference layout. `test_nvfp4_quantize_out_scale_rejected_off_the_fold_path` covers the rejection paths.
- **SM103 tactics profiled and raising:** filtered in `get_valid_tactics` when the alpha is per-token; the raise in `forward()` stays as a backstop.
- **8x4 + per-token tuning config:** per-token config is selected first.
- **`dataclasses.replace`:** done.
- **Error message unions all SM majors:** message now names the backend and SM100/SM103 explicitly, on every arch.
- **Docstring order:** fixed.
- **Test matrix cost:** 160 -> 80 cases (`k` fixed at 256, `m` sweep kept in full). Scale logic in the test is documented in comments.
- **Skip/error reporting from the implementation:** the new rejection tests assert the implementation refuses per-token alpha on other backends; the positive test keeps the file's existing arch-skip convention.
- **Epilogue per-element alpha load:** unchanged; the larger-M per-token-vs-scalar measurement is still owed and will be posted on #5504.

Review continues on #5504. I will close this one once that merges.

PR #4442, opened 2026-08-10
feat: CUB `DeviceBatchedTopK` top-k backend with variable-length support

NaderAlAwar:
## 📌 Description

This PR adds a new API, `top_k_cub`, that uses `cub::DeviceBatchedTopK` internally. It supports fp32/fp16/bf16 inputs, tie-break modes, per-row variable lengths, and CUDA graph capture — including the variable-lengths path, where lengths is read on the device, so a captured graph can be replayed with different per-row lengths at a fixed shape.

This improves performance for plain top_k across the board: CUB wins about 85% of benchmarked shapes in eager mode (geomean ~1.6x over the radix default) and 94-97% under CUDA graphs (geomean 1.7-2.0x), with the only weak spot being large batches (128+) at sequence lengths around 16k-64k. The biggest win is tie-break: it costs CUB only ~3-9% over its own baseline while the native tie-break path is much slower, leaving CUB ~1.5-2x faster, and CUB supports tie-break at k=4096, which the native path rejects. The fused transforms favor the native kernels in eager mode, but under CUDA graphs (how SGLang calls them) CUB wins from mid sequence lengths up and nearly everywhere with tie-break. Shape-based gates in the dispatcher route each call to the faster backend automatically. Full performance results here: https://gist.github.com/NaderAlAwar/4038e4c44365b93737a55add0e6ec1b5

As a follow up, we plan on making the following improvements in future PRs:
- Allow specifying exact dtypes in `gen_topk_module` to improve compilation times (currently we compile all dtypes)
- Allow specifying certain runtime parameters that we know won't change as compile time parameters by passing them to `gen_topk_module` to improve performance. For example, this applies to the max row length in the variable lengths path
- Support `sorted=True`
- Support `deterministic=True`
- Further optimize `cub::DeviceBatchedTopK` to improve performance

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Reviewer Notes

This PR depends on https://github.com/NVIDIA/cccl/pull/9224, and should therefore remain in draft until that PR is merged. We should also update the CCCL submodule to use a more recent commit that includes the changes added in that PR.

## Summary by CodeRabbit

* **New Features**
  * Added an optimized CUB backend for standard, ragged, and page-table top-k operations.
  * Top-k automatically selects the most suitable backend based on input characteristics and execution mode.
  * Added workspace queries, custom workspaces, tie-breaking, raw-index outputs, and CUDA Graph-compatible execution.
  * Short transformed rows now use `-1` padding where applicable.

* **Performance**
  * Expanded benchmark coverage and backend comparisons, including CUDA Graph-aware measurements.

* **Documentation**
  * Updated top-k API and configuration documentation with backend selection details.


NaderAlAwar:
@coderabbitai review

csrc/cub_topk.cu
```diff
@@ -0,0 +1,299 @@
...
+
+#include <cuda/__execution/determinism.h>
+#include <cuda/__execution/output_ordering.h>
+#include <cuda/__execution/require.h>
+#include <cuda/__execution/tie_break.h>
+#include <cuda/std/__execution/env.h>
+
+#include <cub/device/device_batched_topk.cuh>
+#include <cuda/argument>
+#include <cuda/iterator>
```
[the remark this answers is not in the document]
NaderAlAwar: This PR will not be merged until the CCCL PR gets merged and the submodule gets updated

flashinfer/topk.py
```diff
@@ -927,3 +976,148 @@ def top_k_ragged_transform(
...
+    # Host-side size query (launches nothing); the workspace is cached per device so
+    # repeated calls (including under CUDA graph capture) reuse a stable allocation.
+    workspace_bytes = topk_module.cub_topk_workspace_size(
+        input, lengths, k, int(tie_break)
+    )
+    workspace_buffer: torch.Tensor = _get_cache_buf(
+        f"cub_topk_workspace_{device}", workspace_bytes, device
+    )
+
+    return topk_module.cub_topk(input, k, int(tie_break), lengths, workspace_buffer)
```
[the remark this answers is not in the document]
NaderAlAwar: DeviceBatchedTopK temp storage query is guaranteed to return at least 1 byte for none empty batches. For empty batches, the launcher early returns

flashinfer/topk.py
kahyunnam: let's update this doctstring contract, since CUB is chosen without looking at dsa_graph_safe (@NaderAlAwar is this on purpose?)

flashinfer/topk.py
```diff
@@ -821,6 +901,34 @@ def top_k_page_table_transform(
         if not out_raw_indices.is_contiguous():
             raise ValueError("out_raw_indices must be contiguous")
 
+    if can_use_cub_topk(input, tie_break) and not deterministic:
```
kahyunnam: This `can_use_cub_topk` check is before `can_use_clusters_topk` check, which makes `topk_clusters_page_table_transform` a fallback for sm100 auto backend selection. Can we double check if this is justified for perf numbers (CUB dominates clusters topk across the DSA shapes clusters was meant for, like large page-table shapes)?

kahyunnam (commented):
Thanks for the contribution @NaderAlAwar ! Left some comments below. Please mark this PR as "ready" and not "draft" once you are ready to run the pre-merge CI. 

Also, the PR description seems outdated, can you please update?

flashinfer/topk.py
NaderAlAwar: Yeah the CUB backend is always graph safe by default, so this flag doesn't affect it. I updated the docstring

zianglih (commented):
Thank you @NaderAlAwar for this awesome work!

Could we fold `test_cub_page_table_transform_tie_break` into the existing tie-break parametrizations rather than maintain a CUB-only page-table contract test? `test_top_k_tie_break_modes` and `test_top_k_tie_break_modes_transform_apis` already validate exact `SMALL`/`LARGE` selection across plain, page-table, and ragged APIs. Adding compact, SM90-gated `cub` cases there would keep backend coverage in one matrix.

The partial boundary-tie construction here is stronger than the existing all-equal input, so could we preserve it as a shared input-pattern/helper rather than make it CUB-specific? Then the CUB-only section can focus on genuinely backend-specific behavior such as nonpositive lengths and workspace handling.

flashinfer/topk.py
NaderAlAwar: Originally I hadn't accounted for the clusters backend. I reran the benchmarks (including clusters) after my latest changes and in some cases it is justified, so I updated the performance gates

flashinfer/topk.py
```diff
@@ -510,14 +639,223 @@ def topk_clusters_ragged_transform(logits, seq_lens, offsets, top_k, pdl=False):
...
+
+        return (num_rows < 128 and d >= 65536) or d >= 524288
+
+    # When the clusters backend is available (SM100, eager, non-deterministic,
+    # no tie-break, clusters-compatible arguments), it beats CUB on nearly every
+    # measured cell; CUB keeps only very wide row counts at long d. Checked
+    # after the graphed branch because the clusters backend doesn't support
+    # graph capture.
+    if clusters_eligible:
+        return num_rows >= 512 and d >= 65536
```
kahyunnam, line 772: Can we return false here instead? 

The current condition on line 777 sends exactly the large DSA-style shapes to CUB. Measured on this B200, fp16, k=2048, eager, non-deterministic, no tie-break; wall-clock mean of 50 calls after 10 warmup calls, all three backends timed in one process. Auto tracks CUB above the threshold and clusters below it:

| rows | d | auto | CUB | clusters | auto / clusters |
|---|---|---|---|---|---|
| 256 | 131072 | 88.1 µs | 254.7 µs | 88.1 µs | 1.00× (gate false) |
| **512** | **65536** | **319.5 µs** | 319.6 µs | 104.6 µs | **3.05×** |
| **512** | **131072** | **446.7 µs** | 446.5 µs | 161.7 µs | **2.76×** |

flashinfer/topk.py
```diff
@@ -510,14 +639,223 @@ def topk_clusters_ragged_transform(logits, seq_lens, offsets, top_k, pdl=False):
...
+    if dsa_graph_safe or torch.cuda.is_current_stream_capturing():
+        if num_rows < 128 or d <= 8192:
+            return True
+        return d >= (32768 if dtype == torch.float32 else 65536)
+
+    # When the clusters backend is available (SM100, eager, non-deterministic,
+    # no tie-break), it beats CUB nearly everywhere; CUB only keeps the fp32
+    # d ~ 8192 band (below batch 256), fp32 small-batch mid-d, and single-row
+    # long-d calls.
+    if clusters_eligible:
```
kahyunnam, line 702: Ssame thing here as other comment: swept on B200, eager, non-deterministic, no tie-break, median of 5 × 300 iters. **Every cell where the gate returns True is slower than clusters**, and no cell where it returns False would have been a win:

| rows | d | k | dtype | gate | auto | clusters | auto / clusters |
| 1 | 8192 | 256 | fp32 | True | 22.5 µs | 15.9 µs | **1.42×** |
| 1 | 32768 | 2048 | fp32 | True | 39.6 µs | 19.2 µs | **2.06×** |
| 1 | 262144 | 2048 | fp32 | True | 46.5 µs | 22.1 µs | **2.11×** |
| 32 | 16384 | 256 | fp32 | True | 31.3 µs | 17.0 µs | **1.84×** |
| 32 | 65536 | 256 | fp32 | True | 46.7 µs | 19.2 µs | **2.44×** |
| 64 | 8192 | 2048 | fp32 | True | 24.5 µs | 19.1 µs | **1.28×** |
| 128 | 8192 | 2048 | fp32 | True | 24.5 µs | 22.7 µs | 1.08× |
| **1** | **262144** | **256** | **fp16** | True | 52.4 µs | 16.8 µs | **3.11×** |
| 32 | 65536 | 256 | fp16 | False | 17.5 µs | 16.8 µs | 1.04× (correctly declined) |

kahyunnam (commented):
Hi @NaderAlAwar could you check a few of these benchmark numbers?

flashinfer/topk.py
kahyunnam: Resolving after re-measuring with benchmarking harness!

flashinfer/topk.py
NaderAlAwar: made it return false if clusters are eligible

kahyunnam:
@flashinfer-bot run

kahyunnam (approved):

PR #4455, opened 2026-08-11
fix(gemm): validate cached CuTeDSL BF16 tactics

jiahanc:
## Summary

- validate a selected CuTeDSL BF16 direct tactic against the concrete runtime shape in the BF16 dispatch
- fall back through the actual-shape runner order when the cached direct tactic is incompatible
- add M=25 to the existing BF16 accuracy matrix

## Root cause

Bucketed M=25 can reuse a direct tactic tuned for M=32 with rows_per_block=8, which does not divide 25.

## Testing

- B300/SM103a: CuTeDSL BF16 M=25 accuracy with autotune disabled and enabled
- touched-file pre-commit checks, including mypy and Ruff

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved BF16 matrix multiplication reliability by validating selected performance configurations before execution.
  * Added fallback behavior when a selected configuration is incompatible with the input shape or operation settings.
  * Prevented unsupported bias-enabled inputs from using the direct BF16 execution path.

* **Tests**
  * Expanded BF16 matrix multiplication coverage to include an additional matrix size.

### DIFF
--- a/flashinfer/gemm/gemm_base.py
+++ b/flashinfer/gemm/gemm_base.py
@@ -1564,9 +1564,31 @@ def _cute_dsl_direct_bf16_gemm_runner(
         autotune_tactics,
         default_tactic,
         run_direct_dense,
+        validate_tactic,
     )
 
     class CuteDSLDirectBf16Runner(TunableRunner):
+        def is_tactic_compatible(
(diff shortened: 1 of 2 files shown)


StudyingShao:
LGTM

StudyingShao (approved):

PR #4461, opened 2026-08-11
[Bugfix]: handle empty query in SM120 sparse MLA decode for speculative decoding

XingSong-Sun:
## 📌 Description

What does this PR do?

This PR fixes a crash in the SM120 sparse MLA decode path when certain EP ranks receive zero tokens, which happens during speculative decoding.

Changes:

1. Early return on empty query in _trtllm_batch_decode_sparse_mla_sm120: when query_flat.shape[0] == 0, the function now returns a correctly-shaped empty output tensor (and LSE if requested) instead of passing zero tokens to the sparse MLA kernel, which would fail shape validation or crash.

2. Explicit reshape dimension in _normalize_sparse_mla_indices_and_lens: changed reshape(..., -1) to reshape(..., sparse_topk) for clarity and safety — the dimension is already known from the indices tensor, so explicit is better than implicit.

Why it's needed:

In speculative decoding with expert parallelism (EP), it's common for some EP ranks to have no tokens to process in a given step. Without the early return, the zero-token query would hit shape errors deep inside the SM120 sparse MLA kernel. With this fix, the function gracefully handles the empty case and returns immediately, matching the behavior that the TRTLLM-GEN path already has for zero-token batches upstream.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of sparse MLA indices when using configured top-k widths.
  * Sparse MLA speculative decoding now safely handles ranks with zero tokens by returning empty results.
  * Prevented unnecessary decoder setup for empty speculative-decoding inputs.

### DIFF
--- a/flashinfer/mla/_core.py
+++ b/flashinfer/mla/_core.py
@@ -209,7 +209,7 @@ def _normalize_sparse_mla_indices_and_lens(
             raise ValueError(
                 f"Expected {name}.shape == {expected_shape}, got {tuple(indices.shape)}"
             )
-        indices = indices.reshape(batch_size * q_len_per_request, -1)
+        indices = indices.reshape(batch_size * q_len_per_request, sparse_topk)
     elif indices.ndim == 2:
         sparse_topk = int(indices.shape[-1])
         expected_shape = (batch_size * q_len_per_request, sparse_topk)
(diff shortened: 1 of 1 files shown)


bkryu:
Thanks @XingSong-Sun for this PR. 

#4802 has been merged with preserved authorship so I will be closing. We look forward for more contributions from the community.

PR #4473, opened 2026-08-12
fix(comm): align MNNVL two-shot workspace stages

aoshen02:
## Summary

- align each MNNVL Lamport buffer to 32 bytes so both 16-byte two-shot stages remain aligned
- round requested and allocated workspace sizes to the same invariant
- add a large-workspace regression test matching the failing Kimi K3 configuration

## Root cause

The workspace previously aligned each Lamport buffer to 16 bytes. Two-shot all-reduce then divides that buffer into two stages and accesses each stage with 16-byte vector operations. A buffer whose size is 16 modulo 32 therefore places the second stage at an 8-byte-aligned address, causing `CUDA error: misaligned address`.

For the reproduced `max_num_tokens=16384`, BF16, hidden-size 7168 case, the old per-buffer size was 536,870,224 bytes (`size % 32 == 16`). The fixed size is 536,870,208 bytes (`size % 32 == 0`).

## Duplicate check

I searched open FlashInfer PRs for MNNVL stage/workspace alignment, Lamport two-shot alignment, and MNNVL misaligned-address fixes. No open PR fixes this issue. The closest historical changes (#2074, #2118, and #2955) add large-workspace support or change allocation, but retain the 16-byte per-buffer alignment that causes this failure.

## Validation

- `pre-commit run --files flashinfer/comm/trtllm_mnnvl_ar.py tests/comm/test_trtllm_mnnvl_allreduce.py`
- TP2 GB200 regression: `1 passed, 275 deselected`
- standalone TP4 GB200 large-workspace two-shot all-reduce
- four-node Kimi K3 startup with TP4 / DP4 / EP16 and FlashInfer all-reduce: all 51 piecewise and 35 full CUDA graphs captured; no misaligned-address error

OpenAI Codex was used to assist with diagnosis and implementation. The submitter reviewed the complete diff and validation results.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workspace allocation alignment for more reliable communication operations.
  * Fixed capacity calculations when using larger workspaces, including workspaces exceeding the tested sequence length.
  * Improved support for two-stage operations with large workspace configurations.
  * Increased reliability for operations using varied workspace sizes and aligned memory buffers.

* **Tests**
  * Added regression coverage for large-workspace alignment and two-stage operation scenarios.

### DIFF
--- a/flashinfer/comm/trtllm_mnnvl_ar.py
+++ b/flashinfer/comm/trtllm_mnnvl_ar.py
@@ -17,7 +17,7 @@ from flashinfer.comm.mapping import Mapping
 from flashinfer.api_logging import flashinfer_api
 
 from ..jit import gen_trtllm_mnnvl_comm_module
-from ..utils import register_custom_op
+from ..utils import register_custom_op, round_up
 from ..fp4_quantization import _compute_swizzled_layout_sf_size
 from .mnnvl import CommBackend, McastGPUBuffer, MPIBackend, SymmDeviceMemory
 from .trtllm_ar import QuantizationSFLayout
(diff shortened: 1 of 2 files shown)


aleozlx (approved):

aoshen02:
Thanks!

aoshen02:
should be able to merge?

Issue #3605, opened 2026-06-12
[RFC] FlashInfer Release-Quality Action Plan


YangXu1990uiuc:
# FlashInfer Release-Quality Action Plan

*Owner: Yang Xu · written for community review — feedback welcome.*

> **Status update 2026-09-06.** Both harnesses are merged and gate CI by default; the MoE fuzzer has
> since caught two more real bugs; Stream 0 landed a per-unit timeout runner. Sampling and Attention
> are still unstarted (contributors wanted — see §3); Norm/RoPE is being picked up by @0z5a. Changes
> from the June text are marked *(updated)*. Running tally of harness-found bugs added at the end of §3.

## TL;DR

Recent release-quality gaps cluster in low-precision and MoE, and reached users mainly because the
test signal was too weak to gate on and architecture coverage was thin. The plan: port one proven
test kit — 7 invariant oracles + an authoritative reference + adversarial inputs — op by op, run it
on a trustworthy, gating CI signal, and back it with complementary tiers (perf, compute-sanitizer,
framework-integration). Two harnesses (MoE, scaled GEMM) are already built and each found real bugs
soon after landing; the rest are net-new. Then re-audit what still escapes and repeat the loop.

## Why this doc exists

Downstream serving frameworks and users have flagged gaps in FlashInfer's recent release quality,
and the pattern is visible directly in our own issue tracker — e.g. attention quality regressions
(#1645), BF16 MoE stability (#3364), hangs (#3329), and MoE perf (#3521).

This plan lays out, grounded in that history: **what's broken, why it escapes today's tests, what
we're building to catch it before release, and — issue by issue — how much of the historical bug
stream each piece addresses.** The catch-rate figures are forward-looking estimates — from reviewing
past issues one at a time and judging what each harness will catch once fully built out, across
backends/dtypes/architectures — not a measurement of today's coverage, and are given as ranges.

## Precedent — a proven playbook, not a bet

The design here is borrowed from cuDNN, where it already worked. After cuDNN shipped
`test_mhas_v2` — a single consolidated attention fuzzer (one strong harness, an authoritative
reference, adversarial inputs), the direct model for the design below — attention-related bug
reports dropped substantially and the time from kernel-completion to end-to-end convergence
shortened (far fewer silent-correctness round-trips). We are porting that working system to
FlashInfer, op by op. The sharpest techniques carry over directly:

- **sparse + exact-grid inputs** so structural bugs aren't averaged away (a low-precision analog of
[...]

PR #4475, opened 2026-08-12
test: enable the unified MoE fuzzer by default and prune legacy UTs

feih-nv:
## 📌 Description

Takes over and supersedes #3958: rebase onto tot and make the accumulated fuzzer the default regression for the #3957 CUDA-context corruption (fixed by #4186).

- Enable `tests/moe/test_unified_moe_fuzz.py` by default. `FLASHINFER_UMOE_FUZZ=0` remains an emergency waiver. Randomized sweep default is 160 configs (was 80).
- Keep that accumulated sequence in one pytest process via `shard_group("unified-moe-accumulated")`, so node-level CI sharding cannot split the #3957 regression.
- Add a shared finding/quarantine ledger (`tests/test_helpers/fuzz_ledger.py`):
  - Wrong-answer findings still run, then report XFAIL.
  - Crash findings are quarantined before kernel launch.
  - All-backend quarantines report XFAIL rather than SKIP.
  - Unexpected passes fail strictly.
  - Every curated fuzzer seed is unique; duplicates are rejected at import.
- Add CUTLASS backends to the unified MoE fuzzer (`CutlassBf16Config`, `CutlassW4A16Config`) with a shared BF16-grid routing-weight contract and an SM90-safe Torch MXFP4 reference.
- Replace large GEMM/MoE Cartesian grids with curated smoke/regression cases. Randomized shape breadth moves to the default-on unified fuzzers; backend × quant × routing × layout matrices and error-path anchors stay in the original files.
- Fix the MxFP8 B-layout used by the cuDNN override-shape path (column-major `[b, k, n]` view).
- Document that #3547 and #3957 are fixed. The live ledger is empty; those cases remain as regression coverage, not active waivers.

## 🔍 Related Issues

- Supersedes #3958
- #3957 — cumulative CUDA-context corruption; fixed by #4186
- #3547 — expert-offset all-zeros; fixed
- #3605 — release-quality / fuzzing plan

### ✅ Pre-commit Checks

## 🧪 Tests

Focused local checks (SM100):

- [x] `FuzzLedger` unit tests
- [x] Targeted unified-MoE fuzzer cases (CUTLASS BF16, W4A16 reference, seed 99, historical MXFP4 config)
- [x] #4186 output-tail / tactic guards
- [x] Full post-rebase accumulated sweep: 191 passed, 2 skipped
- [ ] CUTLASS W4A16 fuzzer path on SM90/H100 (needs GPU CI)

A previous `tests/gemm` + `tests/moe` GitLab run passed 18/18 jobs, but that pipeline started before the duplicate-seed fix. Re-run after this description lands.

## Reviewer Notes

Legacy-test reductions are intentional: keep kernel-selection and error-path anchors in the original files, and put randomized shape breadth in the default-enabled unified fuzzer. Model-relevant 1024/768 routing sizes remain where the fuzzer does not reproduce the full implementation × weight-layout × activation matrix.

The sigmoid grid dropping `intermediate_size=512` matches that test’s compatible sizes (`384/768/1024`). Linear MxFP8 scale layout is not represented by the public 3D BMM API; that coverage stays in `tests/gemm/test_unified_gemm_fuzz.py`.

## Summary by CodeRabbit

- **New Features**
  - Added public access to the BF16 routed MoE runner.
  - Enabled unified MoE fuzz testing by default in CI, with failure tracking, quarantine handling, and unexpected-pass detection.

- **Bug Fixes**
  - Retained regression coverage for expert-offset handling and improved reference validation for quantized MoE cases.

- **Tests**
  - Streamlined GEMM and MoE tests into focused smoke suites.
  - Expanded randomized coverage through unified fuzz testing across backends, layouts, dtypes, routing, and autotuning scenarios.

- **Documentation**
  - Added contributor guidance explaining smoke-test scope and randomized coverage responsibilities.


feih-nv:
@coderabbitai review

aleozlx (approved):

aleozlx:
H100 job failed on public CI, will circle back later today

copilot:

The job failed because the PR test workflow ultimately reported failures in tests/moe/test_unified_moe_fuzz.py.

Key evidence from the logs:

The run completed all planned nodes, but the summary says:
Failed: 3
Failed test files: tests/moe/test_unified_moe_fuzz.py
The job itself was not failing due to infra or cancellation; it exited with python_exit_code=1 after test failures.
What to fix

The failure is in the unified MoE fuzzer test suite, not in the workflow definition.
The most likely issue is one or more configs in tests/moe/test_unified_moe_fuzz.py producing a mismatch, crash, or unsupported backend path.
Since the log excerpt you provided does not include the exact assertion traceback for the failing seed/config, the fastest path is to inspect the failing config via the repro command printed by that test and/or narrow down the specific seed from the CI output.
Important details from the test file

The test is fully seed-driven and prints a reproducible command:
FLASHINFER_UMOE_FUZZ_ONLY_SEED=<seed> pytest -s tests/moe/test_unified_moe_fuzz.py::test_unified_moe_fuzz
The test checks:
non-finite outputs
numeric agreement vs canonical reference
determinism for backends marked deterministic
output-buffer poisoning
tactic sweep correctness
autotune winner correctness
device-state probe
The file also contains a shared ledger for known bad configs; if a regression reappears, it should be added there only if it is a real tracked issue and not a transient failure.
Most likely fix areas in code

If the failure is numeric, look at the corresponding handler/reference pair in tests/moe/test_unified_moe_fuzz.py:
_nvfp4_reference
_bf16_reference
_mxfp4_reference
_block_fp8_reference
_fp8_per_tensor_reference
_mxint4_reference
If the failure is on a new or changed config, make sure the config is only generated when the backend actually supports the routing mode and architecture.
If it is a real backend regression, the fix likely belongs in the MoE backend implementation or its prepare_weights / pack_inputs path, not in the workflow.
Workflow note

.github/workflows/pr-test.yml is just the CI runner definition; it is not the root cause here. It already stops on the first failing matrix entry, which is expected.
Practical remediation

Reproduce the failing seed from the test’s printed repro command.
Compare out vs ref for that config and backend.
If the issue is a known backend bug, add/update the ledger entry.
If the issue is a genuine regression, fix the corresponding MoE backend path or reference math in tests/moe/test_unified_moe_fuzz.py.
If you want, I can help you turn this into a concise PR review comment or a fix suggestion targeted at the MoE fuzzer file.

aleozlx:
Pushed two commits to unblock this PR:

**99b80589 — merge main.** One conflict hunk in `tests/moe/test_trtllm_gen_fused_moe.py`: kept main's TopKSigmoid comment + this PR's pruned `num_tokens [8, 3072]`. Verified main's new routing configs (TopKSigmoid, DSv3_DA, TopK_16e) still intersect the pruned parametrize grids.

**1b525e7d — fix the 3 H100 fuzzer failures.** All three were expert-parallel configs where CUTLASS is the only wired backend on SM90; CUTLASS runners reject EP in `check_support()`, so `MoELayer` ended with zero usable runners and raised, which the fuzzer counted as FAIL instead of SKIP.
- New `MoERunner.supports_expert_parallelism` ClassVar (mirrors `supports_fused_shared_experts`), enforced in the base `_check_support`; drops the duplicated inline EP checks in the CUTLASS/b12x bases.
- Fuzzer derives `_EP_BACKENDS` from that declaration and restricts EP configs to EP-capable backends, so an EP config with no capable backend on the arch SKIPs like any other unsupported combo.
- `MoELayer`'s no-usable-backend error now includes an EP hint.

**Validated on H200 (SM90)**: seed 90 (the exact CI config) FAILs with the fix reverted, SKIPs with it applied; non-EP CUTLASS seeds 26/32 still PASS (no over-skip); full `test_unified_moe_fuzz.py`: 14 passed / 0 failed (all SM90-runnable configs); `test_unified_moe_cutlass.py` + `test_unified_moe_b12x.py`: 102 passed.

Note: merging main added TopKSigmoid to `_ROUTING_METHODS`, which reshuffles the randomized seed→config mapping — CI's s141/s11 configs no longer exist under those seeds (s90 was stable). Repro commands are per-revision; the EP filter covers the whole failure class.

aleozlx:
@flashinfer-bot run

Issue #4214, opened 2026-07-28
GDN CuTe-DSL cache architecture improvements


kahyunnam:
## Context

Tracking issue for cache correctness and lifecycle problems across the GDN CuTe-DSL kernels. **For open todos, see items in "Suggested implementation sequence" section below.**

Review comments on #4117 prompted a closer look at GDN caching. Most of the findings below pre-date that PR and remain open after it merged:

- Whether a cache uses a module dictionary or `@functools.cache` is not the central correctness question. Either mechanism is unsafe if its key omits a value that changes generated code or typed tensor descriptors.
- Auditing the keys surfaced a related but separate problem: some operands the keys omit are ones the kernels cannot specialize on at all, because the kernel bodies hard-code bf16. For those, completing the key produces a second, still-incorrect binary. Operand dtype has to be pinned or converted, and only the genuinely polymorphic operands keyed.
- #4117 fixed the SM121 target/decorator problem in the WY output-only kernel and added compute capability to that kernel's compile key. It did **not** add `_io_dtype` to the WY key and does not address the broader cache lifecycle described here.

> **Edit 2026-09-04** — status re-audited on `main` @ `8613260a`, verified on B200. Every file path below predates the move of these modules into `flashinfer/gdn_kernels/` (read `~~gdn_decode_bf16_state.py~~` as `gdn_kernels/gdn_decode_bf16_state.py`, and so on). Modules added since filing: `gdn_decode_bf16_wy_ucache.py`, `gdn_decode_bf16_wy_ucache_flush.py`, `dtype_compat.py`, `experimental/`. Strikethrough below marks what has changed since the issue was written; the findings themselves are unedited.

The common underlying problem is that several GDN paths combine three different lifecycles in process-global cache values:

1. compiled-kernel identity;
2. device/architecture policy; and
3. mutable execution state such as default outputs, staging buffers, or workspaces.

FlashInfer's more mature JIT and wrapper paths generally keep these concerns separate: static specialization parameters identify a compiled module, while runtime workspaces are caller- or wrapper-owned.

## Actionable findings


| ID     | Priority                         | Area                                       | Finding and evidence                                                                                                                                                                                                                                                                                                                                                                                         | User-visible impact                                                                                                                                                         | Recommended action                                                                                                                                                             | Regression coverage                                                                                                              |
[...]

PR #4476, opened 2026-08-12
fix(gdn): stream-scope WY staging buffers and prefill launch workspace

elwhyjay:
## 📌 Description

- `gdn_decode_bf16_wy_output_only.py`: the persistent short-T staging buffers (`_STAGE`) were keyed by device/shape/dtype only. Two same-shape calls on different streams shared one buffer set, so the second call's staging `copy_()` could overwrite inputs the first call's kernel was still reading. The stream is now part of the staging key; same-stream reuse keeps the original zero-copy-tail behavior.
- `blackwell/gdn_prefill.py`: the growable launch workspace lived inside the per-specialization compile-cache dict. The kernel writes TMA tensormap descriptors into that workspace on every launch, so concurrent calls on the same specialization could clobber each other's descriptors. The workspace is now allocated per call, and the compile cache holds only the compiled callable and `num_sm`, matching the lifecycle separation described in #4214.

Reproduction of the WY race on unmodified main, measured on RTX 5090 (SM120): two streams, same shape (B=512, T=2, H=HK=16, HV=32), interleaved without sync. Each result is compared against the single-stream result of the same inputs.

| run | corrupted iterations | max abs err |
| --- | --- | --- |
| two streams | stream1 2/100, stream2 41/100 | 0.34 / 0.37 |
| single-stream control (same jitter) | 0/100 | 0.0 |

One observation from reproducing this: the race only materializes when the enqueued GPU work outlasts the host launch gap (at B=64 the GPU drained every op before the next enqueue and 200 iterations showed no corruption), so the regression test deliberately uses a large batch.

New `tests/gdn/test_multistream_overlap.py`:

- Deterministic per-stream buffer-ownership checks for both WY staging branches (full staged T=2, "ab" staging pair at T=4 with non-contiguous a/b). These fail on the pre-fix code regardless of timing.
- The two-stream overlap stress above as a regression test.
- SM100 prefill: two-stream self-consistency plus a white-box check that the compile-cache value holds only `compiled`/`num_sm`.

Existing `tests/gdn/test_decode_delta_rule.py -k wy_output_only`: 15 passed on SM120 before and after the change. I attempted the full `test_decode_delta_rule.py` locally as well, but my dev box's network filesystem kept stalling multi-hour runs, so I'd like to lean on GPU CI for the full suite. The changed code paths are the WY module and the SM100 prefill wrapper, and both are covered by the targeted runs above.

## 🔍 Related Issues

Part 3 of the plan in #4214 (GDN-H4): mutable execution state was stored in process-global caches without stream ownership.
## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Reviewer Notes
- I don't have SM100 hardware locally, so the prefill change is verified only as import/skip-clean on SM120; I'd appreciate a GPU CI run (or a check on H100/B200) for `test_blackwell_prefill_workspace_not_in_compile_cache`.
- While auditing for other instances of the same pattern I noticed `delta_rule_dsl/delta_rule_cp_sm120.py` (`_get_cp_workspace`) also caches workspace tensors in a `functools.cache`. It's not listed under GDN-H4 in #4214, so I left it out to keep the scope as planned. Happy to address it here or in a follow-up if that seems reasonable.
- A destroyed stream leaves its staging entry in `_STAGE` (small, bounded by the number of streams actually used for GDN decode). If that's a concern I'm open to an explicit eviction hook or a capped pool instead.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability during concurrent operations across multiple CUDA streams.
  * Prevented shared temporary buffers from causing incorrect or corrupted results.
  * Ensured each invocation uses isolated workspace memory.
  * Improved safe reuse of temporary resources for short-sequence workloads.

* **Tests**
  * Added coverage for overlapping multi-stream execution and workload correctness.
  * Added validation for compilation-cache safety and numerical output consistency.


kahyunnam:
@flashinfer-bot run

kahyunnam (approved):
Thanks for the contribution @elwhyjay ! This looks solid; approving, and I will help merge once the CI results return.

kahyunnam (changes requested):
@elwhyjay it seems like there's some CI failures. Could you please take a look?

## What CI hit

`tests/gdn/test_multistream_overlap.py::test_blackwell_prefill_workspace_not_in_compile_cache` failed on B300, GB200 and GB300, all cu130:

```
AssertionError: Tensor-likes are not close!
Mismatched elements: 458752 / 458752 (100.0%)
Greatest absolute difference: nan at index (0, 0, 0) (up to 0.001 allowed)
```

The cu129 columns are green only because `_skip_if_not_sm100()` skips on CUDA < 13, and the 5090/H100/RTX 6000 runners skip on arch. SM100 + CUDA 13 is the only combination where this test body actually runs, so this isn't a cu130-specific regression — it's just the first place the test executed. That also answers your reviewer note: the prefill test did get real GPU coverage, and this is what it found.

## Root cause

`_make_prefill_inputs` builds inputs outside the kernel's valid domain, in two independent ways.

**1. `gate` is in the wrong space.** `chunk_gated_delta_rule_sm100` consumes `gate` as linear-space alpha in (0, 1] — `flashinfer/gdn_prefill.py` defaults it to `torch.ones(...)` — and the kernel takes the log itself:

```python
# flashinfer/gdn_kernels/blackwell/gated_delta_net_chunked.py:2074
tGrGate[i] = cute.math.log2(tGrGate[i] + 1e-10, fastmath=True)
```

The test passes `gate=rnd(total, HV, float32) * 0.1`, which is centred on zero, so ~50% of entries are negative. `log2(negative)` is NaN, and the warp-wide inclusive prefix sum right below spreads it across the whole chunk. Both `out` and `ref` end up all-NaN, and `assert_close` counts NaN-vs-NaN as a mismatch — hence the exactly-100% figure.

**2. `q`/`k` are not L2-normalized.** With raw `randn` the row norms are ≈ sqrt(128), and the delta-rule triangular inverse `(I + tril(diag(beta) K Kᵀ))⁻¹` overflows over a 512-token sequence.

Isolating the two on B200:

| inputs | NaN fraction in `out` |
| --- | --- |
| as written | 1.000 |
| linear-space gate only | 0.857 |
| linear-space gate + L2-normalized q/k | 0.000 |

`tests/gdn/test_prefill_state_indices.py` (lines 45-67) already uses the right convention for this kernel, so it's a good template.

elwhyjay:
Thanks for the detailed analysis @kahyunnam. Fixed in _make_prefill_inputs, following the convention in `tests/gdn/test_prefill_state_indices.py`. 
1. gate is now linear-space alpha in (0, 1] (exp(-softplus(...))) instead of the zero-centred randn * 0.1.
2. q/k are L2-normalized.

Only the test inputs changed. Could you re-run tests/gdn when you have a moment? Thank you.

kahyunnam:
Rebased to pick up some CI fixes from main

kahyunnam (approved):

kahyunnam:
@flashinfer-bot run

elwhyjay:
@kahyunnam Thank you for running CI, The test_multistream_overlap failure looks like a signature mismatch after merging with main. I pushed a one-line fix to the test call so the key matches what the wrapper computes. The kernel-side change is unaffected.

kahyunnam:
@flashinfer-bot run

PR #4480, opened 2026-08-12
feat(norm): fused Add+RMSNorm+1x128 fp8 block-quant producer

NVShreyas:
## Summary
Adds `FusedAddRMSNormFP8BlockQuant` — the **block-scale sibling** of the existing `FusedAddRMSNormQuant` (which uses a scalar per-tensor scale). It fuses the common 2-kernel producer before fp8 block-scaled GEMMs — `fused_add_rmsnorm` (bf16) followed by a separate per-token-group 1×128 fp8 quant — into a **single memory pass**.

In one kernel it emits:
- fp8 (e4m3) activation `[M, H]`
- a **dynamic per-1×128-block** fp32 scale in the column-major TMA-aligned `(H/128, round_up(M, 4))` layout that `flashinfer/deep_gemm.py` (`check_sf_layout`) already consumes — no scale rework downstream
- the pre-norm residual (bf16, in place)
- the bf16 normed output (for consumers that need the pre-quant activation, e.g. an MoE router)

The fp8 is quantized from the bf16-rounded normed, so `out` exactly matches re-quantizing `normed_out`.

## Kernel
One CTA per token row, single-wave (`num_threads = d/VEC_SIZE`, VEC8 for `d<=8192`, VEC16 to 16384). `128/VEC_SIZE` consecutive warp-local lanes own one 1×128 block, so its amax is a register-only warp-subgroup shfl (no smem). Mirrors `FusedAddRMSNormQuant`'s PDL + `cudaLaunchKernelEx`. Requires `d % 128 == 0` (and `% 256` / `% 512` for the vec width). e4m3 output (1×128 block-scale activations are e4m3).

## Validation (H200)
Correctness matches a torch reference — residual, `normed_out`, `block_scale`, and dequantized fp8 (see the benchmark's `check_correctness`).

## Benchmark (H200, `benchmarks/bench_fused_add_rmsnorm_fp8_block_quant.py`)
Fused kernel vs the 2-kernel producer (`fused_add_rmsnorm` + a per-token-group 1×128 fp8 quant), bf16 in, warmed, median of 100:

| M | H | fused (µs) | 2-kernel (µs) | speedup | fused GB/s |
|---|---|---|---|---|---|
| 256 | 4096 | 10.2 | 69.0 | 6.78× | 929 |
| 256 | 6144 | 10.2 | 68.0 | 6.66× | 1394 |
| 256 | 8192 | 10.4 | 69.6 | 6.71× | 1823 |
| 1024 | 4096 | 12.9 | 79.4 | 6.17× | 2937 |
| 1024 | 6144 | 19.9 | 112.7 | 5.67× | 2856 |
| 1024 | 8192 | 26.7 | 146.5 | 5.49× | 2838 |
| 4096 | 4096 | 45.3 | 280.5 | 6.19× | 3345 |
| 4096 | 6144 | 65.0 | 405.9 | 6.25× | 3497 |
| 4096 | 8192 | 84.2 | 529.0 | 6.28× | **3599** |

**geomean speedup: 6.23×.** At large M the fused kernel is memory-bound at **~3,600 GB/s (~75% of the H200's ~4.8 TB/s HBM peak)**.

Caveat on the speedup: this baseline's 2nd kernel is a **triton** quant (flashinfer has no public 1×128 CUDA quant), which runs well below peak, so it inflates the ratio. The always-true wins are structural — **one launch instead of two, and ~2× less memory traffic** (no intermediate bf16 round-trip); against an optimal CUDA 2-kernel path the kernel-compute win is ~1.7×.

## API
```python
flashinfer.fused_add_rmsnorm_fp8_block_quant(out, block_scale, normed_out, input, residual, weight, eps, enable_pdl)
```
Caller-allocates the outputs (matching `fused_add_rmsnorm_quant`): `out` fp8 `[M,H]`, `block_scale` fp32 `(H//128, round_up(M,4))`, `normed_out` bf16 `[M,H]`; `residual` updated in place.

## Summary by CodeRabbit

- **New Features**
  - Added a fused operation combining residual addition, RMS normalization, and dynamic FP8 block quantization.
  - Produces updated residuals, BF16 normalized output, FP8 output, and FP32 block scales.
  - Exposed the operation through the public FlashInfer API, with optional programmatic dependency launch support.
- **Benchmarks**
  - Added correctness checks and performance benchmarks across multiple matrix sizes, including comparisons with a two-step baseline.
- **Tests**
  - Added coverage for supported shapes, output accuracy, block scales, and optional execution modes.

### DIFF
--- a/csrc/norm.cu
+++ b/csrc/norm.cu
@@ -186,6 +186,57 @@ void fused_add_rmsnorm_quant(TensorView output, TensorView input, TensorView res
   });
 }
 
+// Add-residual + RMSNorm + 1x128 fp8 block-quant. Emits fp8 `output` [M,d], fp32 `block_scale`
+// (d/128, round_up(M,4)) column-major TMA layout, bf16 `normed_out` [M,d], and updates `residual`
+// in place with (input+residual). See norm::FusedAddRMSNormFP8BlockQuant.
+void fused_add_rmsnorm_fp8_block_quant(TensorView output, TensorView block_scale,
+                                       TensorView normed_out, TensorView input, TensorView residual,
+                                       TensorView weight, double eps, bool enable_pdl) {
+  CHECK_LAST_DIM_CONTIGUOUS_INPUT(input);
+  CHECK_LAST_DIM_CONTIGUOUS_INPUT(residual);
+  CHECK_LAST_DIM_CONTIGUOUS_INPUT(weight);
+  CHECK_LAST_DIM_CONTIGUOUS_INPUT(output);
+  CHECK_LAST_DIM_CONTIGUOUS_INPUT(normed_out);
+  CHECK_DEVICE(input, residual);
+  CHECK_DEVICE(input, weight);
+  CHECK_DEVICE(input, output);
+  CHECK_DEVICE(input, block_scale);
+  CHECK_DEVICE(input, normed_out);
+  CHECK_DIM(2, input);     // (batch_size, hidden_size)
+  CHECK_DIM(2, residual);  // (batch_size, hidden_size)
+  CHECK_DIM(1, weight);    // (hidden_size)
+  CHECK_DIM(2, output);    // fp8 (batch_size, hidden_size)
+  CHECK_DIM(2, normed_out);
+  CHECK_DIM(2, block_scale);  // (hidden_size/128, round_up(batch_size, 4))
+  unsigned int batch_size = input.size(0);
+  unsigned int hidden_size = input.size(1);
+  unsigned int m_pad = (batch_size + 3) & ~3u;
+  TVM_FFI_ICHECK_EQ(residual.size(0), batch_size);
+  TVM_FFI_ICHECK_EQ(residual.size(1), hidden_size);
+  TVM_FFI_ICHECK_EQ(weight.size(0), hidden_size);
+  TVM_FFI_ICHECK_EQ(normed_out.size(0), batch_size);
+  TVM_FFI_ICHECK_EQ(normed_out.size(1), hidden_size);
+  TVM_FFI_ICHECK_EQ(block_scale.size(0), hidden_size / 128);
+  TVM_FFI_ICHECK_EQ(block_scale.size(1), m_pad);
+  TVM_FFI_ICHECK_EQ(hidden_size % 128, 0);
+  ffi::CUDADeviceGuard device_guard(input.device().device_id);
+  const cudaStream_t stream = get_stream(input.device());
+
+  // 1x128 block-scale activations are e4m3 (the layout DeepGEMM/flashinfer fp8 block GEMMs
+  // consume).
+  DISPATCH_DLPACK_DTYPE_TO_CTYPE_FP16(input.dtype(), c_type, [&] {
+    cudaError_t status = norm::FusedAddRMSNormFP8BlockQuant<c_type, __nv_fp8_e4m3>(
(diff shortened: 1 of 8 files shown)


flashinfer/norm/__init__.py
```diff
@@ -389,6 +389,74 @@ def _fused_add_rmsnorm_quant_fake(
     pass
 
 
+@register_custom_op(
+    "flashinfer::fused_add_rmsnorm_fp8_block_quant",
+    mutates_args=("out", "block_scale", "normed_out", "residual"),
+)
+def fused_add_rmsnorm_fp8_block_quant(
```
kahyunnam, line 397: Is this the intended public interface? It is missing the `@flashinfer_api` decorator (please refer to flashinfer API conventions)

kahyunnam (commented):
Please address below comment + testing as discussed on slack before we run CI

flashinfer/norm/__init__.py
NVShreyas: thanks, added

csrc/norm.cu
```diff
@@ -186,6 +186,63 @@ void fused_add_rmsnorm_quant(TensorView output, TensorView input, TensorView res
...
+  // number of warps of vectors -> <= 16384 and a multiple of 32*vec_size (256 for <=8192, 512
+  // above).
+  unsigned int vec_size = hidden_size <= 8192 ? 8 : 16;
+  TVM_FFI_ICHECK(hidden_size <= 16384 && hidden_size % (32 * vec_size) == 0)
+      << "fused_add_rmsnorm_fp8_block_quant: hidden_size " << hidden_size
+      << " unsupported (must be <= 16384 and a multiple of " << (32 * vec_size) << ")";
+  ffi::CUDADeviceGuard device_guard(input.device().device_id);
+  const cudaStream_t stream = get_stream(input.device());
+
+  // 1x128 block-scale activations are e4m3 (the layout DeepGEMM/flashinfer fp8 block GEMMs
```
kahyunnam, line 242: Can we also add checks for validating out's shape and dtype, block_scale's dtype and contiguity, and normed_out's dtype?

flashinfer/norm/__init__.py
```diff
@@ -389,6 +389,75 @@ def _fused_add_rmsnorm_quant_fake(
     pass
 
 
+@flashinfer_api
+@register_custom_op(
+    "flashinfer::fused_add_rmsnorm_fp8_block_quant",
+    mutates_args=("out", "block_scale", "normed_out", "residual"),
+)
+def fused_add_rmsnorm_fp8_block_quant(
```
kahyunnam, line 397: can we add this new API to the docs? docs/api/norm.rst

kahyunnam (approved):
This LGTM overall, thanks for the contribution! A few nits below that might be good to look at quickly before merging.

csrc/norm.cu
NVShreyas: Done in 3268a087

New test `test_fused_add_rmsnorm_fp8_block_quant_rejects_bad_outputs` covers all five cases.

flashinfer/norm/__init__.py
NVShreyas: Done in 2b2e6cbd — added `fused_add_rmsnorm_fp8_block_quant` to the autosummary in `docs/api/norm.rst`, right after `fused_add_rmsnorm_quant`.

kahyunnam (approved):

kahyunnam:
@flashinfer-bot run

PR #4505, opened 2026-08-13
ci: upgrade sccache and retain stats

dierksen:
## Summary

- pin the JIT-cache build helper to [sccache v0.17.0](https://github.com/mozilla/sccache/releases/tag/v0.17.0) for both x86_64 and aarch64 builds
- collect text, JSON, and advanced JSON stats from an EXIT hook so failed builds retain evidence without changing their exit status
- add a dedicated `Display sccache stats` workflow step to each nightly, release, and PR AOT matrix job that enables sccache
- retain safe cache/build metadata, including cache mode, key prefix, source SHA, CUDA architecture list, container tag, and pulled image digest
- upload the stats and metadata with `if: always()` for later comparison

## Motivation

The nightly JIT-cache jobs are reporting unexpectedly low hit rates. The existing `::group::sccache stats` markers were valid and GitHub rendered them as a foldable section, but that section was buried near the end of a roughly 294,000-line container-build step. A separate workflow step makes the final stats immediately visible in the job step list, while machine-readable artifacts make runs easy to compare later.

The v0.17 client-side execution mode remains disabled because it is opt-in; this keeps the version upgrade separate from an execution-mode change. Full sccache debug logging is intentionally not enabled to avoid runtime, artifact-size, and credential-exposure risk.

## Artifact contents

- `sccache-stats.txt`
- `sccache-stats.json`
- `sccache-advanced-stats.json`
- `sccache-metadata.txt`
- `job-metadata.txt`

Artifacts are retained for 14 days and are uploaded on both success and failure.

## Validation

- verified official v0.17.0 Linux musl archives and checksum assets for x86_64 and aarch64
- exercised setup, metadata, text/JSON/advanced stats, secret exclusion, and EXIT collection with a mocked sccache lifecycle
- confirmed the stats-only path does not set `SCCACHE_LOG` or `SCCACHE_ERROR_LOG` and does not create a debug-log artifact
- `bash -n` on all changed shell scripts
- YAML parsing on all changed workflows
- actionlint v1.7.12, filtering only the repository baseline shellcheck and custom self-hosted runner-label diagnostics
- repository pre-commit hooks on all changed files
- `git diff --check`

A real CUDA JIT-cache build was not run locally; the workflow matrix will provide that behavioral validation.

## Summary by CodeRabbit

* **New Features**
  * Added build diagnostics capturing cache statistics and build environment metadata.
  * Build and test artifacts now include diagnostic reports retained for 14 days.

* **Bug Fixes**
  * Diagnostic collection now runs even when builds fail or exit early.
  * Collection failures no longer change the original build result.
  * Added warnings when cache statistics are unavailable.

### DIFF
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -167,7 +167,7 @@ jobs:
           export CUDA_MINOR
           export FLASHINFER_LOCAL_VERSION="cu${CUDA_MAJOR}${CUDA_MINOR}"
 
-          mkdir -p ${{ github.workspace }}/ci-cache
+          mkdir -p ${{ github.workspace }}/ci-cache ${{ github.workspace }}/sccache-diagnostics
 
           docker run --rm \
             -v ${{ github.workspace }}:/workspace \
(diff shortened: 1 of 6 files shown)


dierksen:
@flashinfer-bot run

yongwww (approved):
overall lgtm. wondering if there is any new hit rate number after this change

PR #4507, opened 2026-08-13
fix(gdn): compile and tune CuTe-DSL decode kernels for the operand device

kahyunnam:
## Summary

PR #2 of the GDN CuTe-DSL cache audit (#4214), covering **GDN-H2** and **GDN-H3**: GDN decode compiled and tuned for whatever GPU is CUDA device 0, not for the device the operands live on.

- **GDN-H2** — `gdn_decode_bf16_state.py` read SM count and compute capability from device 0 at import time. Those drive the `tile_v` wave heuristic and the `use_packed_fma` Constexpr, so operands elsewhere got the wrong specialization. On a mixed L40S + H100 NVL box, `_get_bf16_mtp_config(B=17, T=2, HV=32, V=128)` picks `tile_v=64` from the L40S's 142 SMs but `tile_v=128` from the H100's 132.
- **GDN-H3** — every GDN `cute.compile` omitted an explicit target, so the DSL resolved the arch from device 0, and the cubin was cached under a key containing neither arch nor device.

## What this does

New `flashinfer/gdn_kernels/device_target.py` resolves one target per device (arch, SM count, `use_packed_fma`, honoring `CUTE_DSL_ARCH`). The GDN decode and Blackwell prefill compile sites now drive both their compile options and their cache keys from that target, and launch streams follow `q.device`.

Two constraints shaped it:

- Compile options had to move from the string form to option objects. The DSL replaces subscripted options wholesale when a string `options=` is present, so `cute.compile[(GPUArch(...),)](..., options="...")` silently drops the arch.
- The target splits across two identities. The in-process key carries `(device_index, arch)` because those entries also hold device-resident default tensors that nothing else keys by device; the compiled artifact itself is device agnostic from DSL 4.6.2 on. The on-disk artifact name carries the arch alone — naming it by device would give every GPU its own copy of every kernel and defeat the disk cache.

Pinning `GPUArch` does not by itself let one process serve another device's architecture: the DSL builds a JIT engine only when its process-global arch (`CUTE_DSL_ARCH`, else device 0) can run the requested target. That case now raises an error naming `CUTE_DSL_ARCH=<arch>` instead of a DSL internal error or a launch-time `cudaErrorNoKernelImageForDevice`.

On homogeneous hardware this is a no-op: the arch suffix rule is the one the DSL's own `detect_gpu_arch` uses, so the compiled arch string is byte-identical to what it picked implicitly.

## Testing

B200 (`sm_100a`), `nvidia-cutlass-dsl==4.7.0`: full `tests/gdn/` — 3810 passed, 928 skipped, 13 failed, with all 13 reproducing on unmodified main in the same environment (12 isolated-subprocess tests that import `flashinfer` from the editable install; 1 disk-cache round trip that never triggers a compile). Earlier run on the mixed L40S + H100 NVL box: 3760 passed, 984 skipped, 0 failed.

`tests/gdn/test_gdn_device_target.py` adds resolver coverage (requested device, `CUTE_DSL_ARCH` override of arch *and* policy, rejection of a target the DSL would cross-compile), two AST guards — compile sites must pin the operand's target, and adopters must not read launch policy or streams off the ambient device — and a multi-GPU bit-exactness test that skips on a single-GPU box. The guards are mutation-checked: dropping the arch from the options, the device from the key's consumers, or the device from a stream or policy read each fails at least one.

Still to do before leaving draft: the multi-GPU test on a homogeneous two-GPU box. Its bf16-state MTP arm is the one that can fail there, since that cache value holds per-B index tensors the kernel dereferences; both arms have been run single-device.

## Out of scope

- GDN-H4 staging / workspace ownership → #4476
- `delta_rule_dsl/` and `blackwell/gdn_cp_prefill.py` compile through their own `cached_compile` shim (~20 sites). They do pin an arch, but from a module constant rather than the operand's device, and their cache is not keyed by device — the same GDN-H3 defect, left for a follow-up.
- `_get_compile_arch()` in `flashinfer/jit/cute_dsl_core.py` resolves the disk cache's arch from the *current* device, which need not be the operand's — the same defect one layer down. Harmless today only because this PR puts the arch in the kernel name, so a mislabeled directory cannot collide with a correctly labeled one.

## Summary by CodeRabbit

- **New Features**
  - Added device-aware compilation for GDN kernels, automatically selecting architecture-specific options and execution settings.
  - Added support for cross-compilation architecture overrides with validation.
  - Improved kernel caching to distinguish device targets and architectures while sharing compatible compiled artifacts.

- **Bug Fixes**
  - Ensured kernels use the correct device-associated CUDA stream and runtime properties.

- **Tests**
  - Expanded coverage for device targeting, compilation options, cache specialization, multi-device execution, and cross-device consistency.

### DIFF
--- a/flashinfer/gdn_kernels/blackwell/gdn_prefill.py
+++ b/flashinfer/gdn_kernels/blackwell/gdn_prefill.py
@@ -50,7 +50,7 @@ from .gated_delta_net_chunked import GatedDeltaNetChunkedKernel
 # the key because the tile scheduler and GQA reshape logic bake them in.
 @functools.cache
 def _get_compiled_cache(
-    arch: str,
+    target_key: tuple,
     num_sm: int,
     io_dtype_str: str,
     state_dtype_str: str,
@@ -194,7 +194,7 @@ def chunk_gated_delta_rule_sm100(
     # num_sm is baked into the kernel's tile scheduler, so it belongs in the key.
     target = gdn_device_target(q.device)
     cache = _get_compiled_cache(
-        target.arch,
+        target.compile_key,
         get_num_sm(q.device),
         str(q.dtype),
         str(state_torch_dtype),
--- a/flashinfer/gdn_kernels/blackwell/gdn_prefill.py
+++ b/flashinfer/gdn_kernels/blackwell/gdn_prefill.py
@@ -37,6 +37,7 @@ from cutlass.cute.runtime import from_dlpack
 
 from flashinfer.cute_dsl.utils import get_num_sm
 
+from ..device_target import gdn_compile_options, gdn_device_target
 from .gated_delta_net_chunked import GatedDeltaNetChunkedKernel
 
 
@@ -49,6 +50,8 @@ from .gated_delta_net_chunked import GatedDeltaNetChunkedKernel
 # the key because the tile scheduler and GQA reshape logic bake them in.
 @functools.cache
 def _get_compiled_cache(
+    arch: str,
+    num_sm: int,
     io_dtype_str: str,
     state_dtype_str: str,
     HQ: int,
@@ -188,7 +191,11 @@ def chunk_gated_delta_rule_sm100(
     use_state_indices = state_indices is not None
     _state_indices = state_indices if use_state_indices else None
 
+    # num_sm is baked into the kernel's tile scheduler, so it belongs in the key.
+    target = gdn_device_target(q.device)
     cache = _get_compiled_cache(
+        target.arch,
+        get_num_sm(q.device),
         str(q.dtype),
         str(state_torch_dtype),
         HQ,
@@ -290,7 +297,9 @@ def chunk_gated_delta_rule_sm100(
 
         stream = cuda.CUstream(torch.cuda.current_stream(device=q.device).cuda_stream)
 
-        compiled = cute.compile(
+        compiled = cute.compile[
+            gdn_compile_options(q.device, cute.EnableTVMFFI(True), cute.OptLevel(3))
+        ](
             gdn,
             q_cute,
             k_cute,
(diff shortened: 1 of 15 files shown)


hebo1221:
I reran current head `8054d4cd` on a GB10 (`sm_121a`) with `nvidia-cutlass-dsl==4.7.0`/cu13. `test_gdn_device_target.py` passed 13/14 (the multi-GPU case skipped), and a small BF16 matrix covering ILP4 cache on/off, the wide dispatch boundary, direct wide cache on/off, and fused recovery passed 7/7.

Two by-path WY tests still fail with the default SM121 target:

- `test_output_matches_fp32_reference[base0-empty_P0-bf16]`
- `test_verify_only_output_matches_fp32_reference`

Both fail in `CuteExperimentalDSL._generate_kernel_attrs`: the native 4.7 parser rejects `#core.compute_capability<arch = sm_121>` (its accepted enum stops at `sm_120`). The same two tests pass with `CUTE_DSL_ARCH=sm_120f`. Since these modules use `@cute.experimental.kernel` while the passing BF16-state kernels use `@cute.kernel`, this looks isolated to the experimental WY launch path. I would treat `sm120f` as a useful discriminator here, not necessarily the final fix.

One rebase wrinkle: replaying this head onto current main `38ebb151` makes current main's two `test_verify_kernel_mtp_reuses_compile_across_cache_modes` cases fail with `TypeError: 'function' object is not subscriptable`. Their mock replaces `cute.compile` with a plain function, while this PR calls `cute.compile[options]`. A small proxy implementing both `__call__` and `__getitem__` makes both cases pass.

kahyunnam:
@flashinfer-bot run

kahyunnam:
@flashinfer-bot run

kahyunnam:
@flashinfer-bot run tests/gdn

bkryu (approved):

PR #4444, opened 2026-08-10
perf(gdn): reuse pretranspose kernels across pool capacity and stride

leonardHONG:
## 📌 Description

Make the leading pool dimension of the GDN pretranspose path dynamic:

- Treat pool capacity and `stride[0]` as runtime values.
- Keep `HV/V/K` and inner strides compile-time static.
- Remove pool capacity and leading stride from the compilation cache key.
- Preserve 64-bit pool-offset arithmetic.

This allows one compiled kernel to be reused across non-contiguous state pools with different capacities and leading strides.

## 🔍 Related Issues

Partially addresses #4110.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit`.
- [x] I have installed the hooks.
- [x] I have run `pre-commit run --all-files`.

## 🧪 Tests

- [x] Tests have been added or updated as needed.
- [x] All relevant tests are passing.

Validated on H800/SM90a and RTX 5090/SM120, including non-contiguous pools, compile reuse, and large 64-bit offsets. No steady-state regression was observed; cold compilation time was reduced by approximately 62%.

## Reviewer Notes

Only the outer pool mode is dynamic. Inner dimensions and strides remain static for kernel specialization.

## Summary by CodeRabbit

* **Performance**
  * Improved pretranspose kernel reuse across different pool capacities and outer memory layouts, reducing unnecessary recompilation.
  * Preserved separate handling for layouts with differing inner strides.

* **Reliability**
  * Added validation to reject pool slot strides that are not properly aligned for efficient data transfer.

* **Tests**
  * Expanded coverage for cache reuse, layout variations, alignment validation, and environments where the optional kernel is unavailable.

PR #4128, opened 2026-07-24
perf(gdn): reuse MTP decode kernels across cache modes

hebo1221:
## 📌 Description

This partially addresses #4110 by removing one redundant specialization axis from the FP32-state MTP verify kernels.

`cache_intermediate_states=True` and `False` previously produced separate CuTe-DSL binaries even though the flag does not change launch geometry or shared-memory size. The flag is now a CTA-uniform runtime `cutlass.Boolean`, so both modes share one compiled kernel. The disabled path uses a cached `[1, V, K]` placeholder because TVM-FFI still validates static inner dimensions even when the runtime branch does not access the tensor.

Cold-compilation telemetry for the existing 30-case `test_verify_kernel_mtp` matrix on GB10 / SM121, PyTorch 2.13.0+cu130, and nvidia-cutlass-dsl 4.6.0.dev0:

| | `cute.compile` calls | compile time | total test time |
|---|---:|---:|---:|
| current `main` | 20 | 111.3 s | 112.7 s |
| this PR | 10 | 58.2 s | 59.5 s |

Steady-state GPU latency did not regress in the same environment for `B=8, T=4, H=16, HV=32, K=V=128`:

| mode | current `main` median | this PR median |
|---|---:|---:|
| intermediate cache on | 397.9 µs | 396.5 µs |
| intermediate cache off | 42.208 µs | 42.208 µs |

The remaining T/tile/kernel-selection specializations are intentionally unchanged in this PR.

## 🔍 Related Issues

Partially addresses #4110.

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [ ] I have installed the hooks with `pre-commit install`.
- [ ] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up pre-commit, see the [pre-commit documentation](https://pre-commit.com/).

The complete hook set passed for both changed files via:

```text
pre-commit run --files flashinfer/gdn_kernels/gdn_decode_mtp.py tests/gdn/test_decode_delta_rule.py
```

## 🧪 Tests

- [x] Tests have been added or updated as needed.
- [x] All relevant tests are passing.

Validation on GB10 / SM121:

- `72 passed` across MTP cache modes, state writeback representatives, contiguous and non-contiguous pools, pool indexing, and packed QKV.
[...]

Issue #4110, opened 2026-07-23
[Compilation] GDN decode cold kernel compilation time is too long on SM90/100/120/121


kahyunnam:
Noticed by @bkryu : "In our recent CI telemetry, we have been seeing that tests/gdn/test_decode_delta_rule.py runs very long (30-50 mins on every arch) and has been straining our CI. 

TL;DR of the cause: the reference and kernels are all fine. Most time is spent on cold kernel compilation. On a B200 I locally see ~800 cases that collapses to ~200 cute.compile calls based on the cache key logic.

For some kernels, B/tile_v/T/cache flags/pool-strides each key a new compile. Would be great if we can make the kernels batch and/or seqlen dynamic to systematically reduce unit testing time."

kahyunnam:
This is similar to #2 in https://github.com/flashinfer-ai/flashinfer/issues/3602 , resolved in https://github.com/flashinfer-ai/flashinfer/pull/3649

bkryu:
Issue present on SM90/100/120/121. Investigation is needed to determine whether all SM archs share the same cause.

kahyunnam:
^ SM90/100/120/121 use the same CuTeDSL decode source and cute.compile flow

leonardHONG:
I’d like to work on MTP sequence-length reuse and making pool capacity/leading stride dynamic. I have sm90 and sm120 available for validation and will send separate PRs.

kahyunnam:
@leonardHONG thanks for volunteering! I've assigned the issue to you

akurathiswaraj:
Hi @kahyunnam and @leonardHONG, would it be possible for me to help with a separate part of this issue? I have access to B200 and B300 GPUs for testing. I’m happy to coordinate with @leonardHONG and work on a clearly scoped subtask to avoid duplicating effort.

leonardHONG:
Hi,@akurathiswaraj .That would be really helpful! I’m currently working on the pool/stride specialization pieces, so a B200/B300 sweep of the remaining tile_v and kernel-selection configs would be a clean separate task.

akurathiswaraj:
Thanks @leonardHONG. I completed the B200/B300 `tile_v` and kernel-selection sweep we discussed.

### Test environment

- FlashInfer commit: `553c2280a88581c7ede39aa571a54d1232854b81`
- PyTorch: `2.8.0+cu129`
- CUDA runtime: `12.9`
[...]

PR #4513, opened 2026-08-13
test(gdn): trim GDN decode test matrix to distinct kernel specializations

kahyunnam:
## 📌 Description

`tests/gdn/test_decode_delta_rule.py` re-ran the same compiled cubins many times
over. Two independent reasons, both verified against the cache keys rather than
assumed:

**1. Batch size is not a compile key.** It has been dynamic since #3649, so it
reaches the cache only through coarse buckets:

| Path | How `B` enters the key |
|------|------------------------|
| pretranspose (`gdn_decode_pretranspose.py:964`) | not at all |
| bf16-state wide-vec (`gdn_decode_bf16_state.py:3440`) | not at all — the tests build a contiguous pool, so `pool_size_key = -1` and `pool_slot_stride = (-1,)` are B-independent sentinels |
| nontranspose (`gdn_decode_nontranspose.py:725`) | only via `use_small_batch = B < 32` |
| fp32 / bf16 MTP (`gdn_decode_mtp.py:2500`, `gdn_decode_bf16_state.py:3813`) | only via `get_mtp_config` / `_get_bf16_mtp_config` |

The clearest case was `test_gdn_decode_bf16_state_wide_vec_mtp_kernel`: 378 of the
file's 817 parametrized cases (46%) but only **42** compile keys, because `tile_v`
is an explicit monkeypatched axis and the 9 batch sizes contribute nothing.

The first commit keeps one batch size per bucket. I verified each kept set
reproduces the *full* key set, at `NUM_SMS` 108/132/148 — this caught a real
mistake, where `[1,8,16,32]` for `test_gdn_decode_bf16_state_t1_kernel` silently
dropped the `HV=64, tile_v=32` key (that test sweeps `HV` ∈ {32,64}).

**2. Intermediate `seq_len` values only re-specialize on `T`.** `get_mtp_config`
returns an identical `(tile_v, vec_size, ilp_rows, use_smem_v)` set for every
`T >= 3`, so T=3/5/6/7 compile fresh MTP cubins without covering a tile config
that T=4 or T=8 does not already cover. `T=2` is kept as the one structurally
distinct case — it alone reaches the `ilp=8` and `tile_v=16 / ilp=2` branches.

### Effect

| | cases | compile keys (retuned tests) |
|---|---|---|
| before | 817 | 102 |
| after batch collapse | 501 | 102 |
| after `seq_len` trim | 409 | 62 |

Collected tests go 838 → 416 (409 parametrized + 7 non-parametrized).

### Coverage cost

The first commit costs nothing in specialization coverage — the same cubins still
run, just at fewer runtime batch sizes. The second commit is a deliberate
reduction: T=3/5/6/7 still exercise distinct unrolled loop counts, so a
T-specific off-by-one would no longer be caught. It is a separate commit so it
can be dropped if reviewers would rather keep the full sweep.

## 🔍 Related Issues

Refs #4110 (GDN cold-compile CI time). Complements #4128 and #4444, which remove
key entries that provably do not reach codegen; this removes test cases that map
onto keys already covered.

### ✅ Pre-commit Checks

## 🧪 Tests

- [x] `pytest tests/gdn/test_decode_delta_rule.py -q` on H100: **416 passed in 32m25s**
- [ ] GPU CI for the timing comparison

## Reviewer Notes

- I did not measure a clean before/after wall clock: the ~38 min baseline I was
  working from comes from #4219's description rather than the same machine, so
  I'd rather let CI provide the comparison than quote a number I can't stand
  behind. The case and key counts above are exact and static.
- Worth noting for #4110 more broadly: cutting 50% of the cases bought
  substantially less than 50% of the wall clock, which suggests the remaining
  cost is dominated by compilation and fixed overhead rather than per-case
  execution. That points at persistent/AOT CuTe-DSL artifacts (GDN-P1 in #4214)
  as the larger lever.
- The batch sizes kept per test are load-bearing, not arbitrary — each set is one
  representative per config bucket. I left a one-line comment at each site so
  they don't get "restored" later.

## Summary by CodeRabbit

* **Tests**
  * Streamlined decode and MTP test coverage to use representative batch sizes and sequence lengths.
  * Preserved coverage for key thresholds, tile configurations, transposition modes, precision variants, and sequence-length scenarios.

### DIFF
--- a/tests/gdn/test_decode_delta_rule.py
+++ b/tests/gdn/test_decode_delta_rule.py
@@ -1271,7 +1271,8 @@ def test_verify_kernel_mtp(
 # ============================================================================
 
 
-@pytest.mark.parametrize("seq_len", [2, 3, 4, 5, 6, 7, 8])
+# T >= 3 all select the same tile config, so 3/5/6/7 only re-specialize on T.
+@pytest.mark.parametrize("seq_len", [2, 4, 8])
 # One B per get_mtp_config bucket.
 @pytest.mark.parametrize("batch_size", [1, 4, 8, 16, 64])
 @pytest.mark.parametrize("dtype", ["bfloat16"])
@@ -1288,8 +1289,8 @@ def test_mtp_fp32_state_with_cache_and_state_update(
     - FP32 h state (not bf16)
     - cache_intermediate_states=True
     - disable_state_update=False (h is updated)
-    - All batch sizes: 1, 2, 4, 8, 16, 32, 64, 128, 256, 512
-    - All sequence lengths: 2, 3, 4, 5, 6, 7, 8
+    - One batch size per get_mtp_config bucket: 1, 4, 8, 16, 64
+    - Sequence lengths 2, 4, 8 (T>=3 shares one tile config)
     """
     scale_val = 1.0 / math.sqrt(128)  # head_size=128
     _test_verify_kernel_mtp(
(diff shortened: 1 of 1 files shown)


kahyunnam:
@flashinfer-bot run

bkryu (approved):

PR #4386, opened 2026-08-07
feat(moe_ep): SM100 BF16 CuTeDSL MegaMoE kernel

mhoqueanik:
# feat(moe_ep): SM100 BF16 CuTeDSL MegaMoE kernel

> Note: This PR is a rebased version of https://github.com/flashinfer-ai/flashinfer/pull/4120 from @djns99

## Summary

Adds an **unquantized BF16 MegaMoE backend** to `flashinfer.moe_ep` — bf16 weights, bf16 activations, bf16 combine, no quantization anywhere in the pipeline — as a fused mega kernel (dispatch + grouped GEMM + combine in one launch) for Blackwell SM100. Port of draft PR #4120 (BF16 MegaMOE integration) onto the restructured taxonomy layout.

Two roles:

1. **A backend for serving bf16 MoE checkpoints** through the same `MoEEpLayer` mega path as the fp8/fp4 kernels.
2. **The accuracy baseline for the quantized backends**: it runs real bf16 model math (~0.29% rel-L2 vs the fp32 dense reference, i.e. bf16 rounding only), so quantized-kernel speedups and quality losses can both be quoted against it.

## What's included

- **Kernel drop**: `kernel_src/cutedsl_megamoe/src/moe_bf16_glu/` + `shim/bf16.py`, plus the bf16-enabling generalizations of the shared mxfp8/nvfp4 kernel sources (epilogue `fc1_output` width guards, `epi_flag_batch` as a `(fc1, fc2)` pair, `TopkReduce` `sm_arch` parameter, iket ranges).
- **Backend**: `backends/mega/kernel/sm100/bf16_bf16_bf16_cutedsl/` with taxonomy config `Sm100_Bf16_Bf16_Bf16_Cutedsl_MegaMoeConfig`, kernel name `sm100_bf16_bf16_bf16_cutedsl`, and deprecated alias `bf16_cutedsl`.
- **Tests**: bf16 oracle + config + mega multirank wired into `run_tests.sh`.
- **Benchmark**: `benchmarks/bench_bf16_cutedsl_megamoe.py`.

## Usage

```python
from flashinfer.moe_ep import (
    BootstrapConfig, FleetParams, MegaConfig, MoEEpLayer, MoEEpTensors, MoEWeightPack,
    Sm100_Bf16_Bf16_Bf16_Cutedsl_MegaMoeConfig,
)

layer = MoEEpLayer(
    BootstrapConfig(world_size=world_size, rank=rank),
[...]

PR #4531, opened 2026-08-15
fix(moe_ep): fix in_kernel_fc2_reduce livelock on zero-token launches (MXFP8 + NVFP4)

mhoqueanik:
## Summary

Fixes a livelock in the MXFP8 and NVFP4 CuTeDSL MegaMoE kernels when `in_kernel_fc2_reduce` (IKR) is enabled and a rank receives zero tokens for a launch: the reduce path spins waiting for FC2 tiles that will never be produced, hanging the fleet. Found while integrating moe_ep with SGLang, where empty-rank launches occur routinely under real routing distributions. Also guards the autotuner so a tuned `token_back_mode` can no longer conflict with the IKR setting. Standalone repro scripts and multirank regression tests (with routing jitter to reproduce the pytest-vs-script scheduling gap) are included for both dtypes.

## Directories affected

- `flashinfer/moe_ep/kernel_src/cutedsl_megamoe/shim/` — `mxfp8.py`, `nvfp4.py` (functional fixes in the shim layer; the vendored `src/` tree is untouched)
- `tests/moe_ep/` — MXFP8 and NVFP4 mega multirank regression tests
- `tests/` — standalone `repro_ikr_zero_token_idle{,_nvfp4}.py` artifacts

6 files changed, +824 / −7.

## Changes

- `shim/mxfp8.py`, `shim/nvfp4.py`: on zero-token launches the IKR path is bypassed so the kernel completes and the combine step sees an empty contribution instead of spinning; the tuned-knob resolution no longer lets a cached `token_back_mode` enable a reduce mode that conflicts with the active IKR configuration.
- `tests/moe_ep/test_moe_ep_mxfp8_cutedsl_mega_multirank.py`, `test_moe_ep_nvfp4_cutedsl_mega_multirank.py`: regression cases that drive a rank to zero tokens and assert completion; routing jitter added because the deterministic pytest distribution masked the livelock that the standalone scripts reproduced.
- `tests/repro_ikr_zero_token_idle.py`, `tests/repro_ikr_zero_token_idle_nvfp4.py`: self-contained repro artifacts documenting the failure mode outside pytest.

## Testing

Reproduced and verified fixed on 8x B200 via the standalone repros and the new multirank tests for both MXFP8 and NVFP4. The zero-token case livelocks deterministically before the fix and completes after.

## Notes for reviewers

- The fix lives entirely in the shim layer, not in the vendored kernel drop, so no provenance update is needed.
- Touches the same shim files area as the pending BF16 drop PR (#4386) only at the directory level; no file overlap — whichever lands second should still re-run the mega multirank suites.

AI-assisted (Claude Code).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved distributed Mixture-of-Experts processing when some devices receive zero tokens.
  - Prevented potential stalls or livelocks during mixed zero-token and real-token workloads.
  - Ensured all devices participate consistently in dispatch, reduction, cleanup, and synchronization.
  - Resolved conflicting token-back configuration when in-kernel reduction is enabled.

- **Tests**
  - Added multi-device regression coverage for MXFP8 and NVFP4 zero-token scenarios.
  - Added standalone diagnostic scripts with progress monitoring and stall detection.

### DIFF
--- a/flashinfer/moe_ep/kernel_src/cutedsl_megamoe/shim/mxfp8.py
+++ b/flashinfer/moe_ep/kernel_src/cutedsl_megamoe/shim/mxfp8.py
@@ -875,9 +875,22 @@ def get_symm_buffer_for_mxfp8_mega_moe(
     # override that conflicts with the (unrelated, untouched) in_kernel_fc2_reduce
     # field raises immediately inside with_knobs(), before any post-hoc fixup
     # here could run.
-    if in_kernel_fc2_reduce and knobs and knobs.get("token_back_mode") not in (
-        None,
-        "epi_warps",
+    # The knobs dict may itself carry in_kernel_fc2_reduce (e.g. a verbatim
+    # tuned-cache entry); the effective value after with_knobs() is the knob,
+    # not the argument, so sanitize against whichever will win.
+    effective_ikr = (
+        knobs.get("in_kernel_fc2_reduce", in_kernel_fc2_reduce)
+        if knobs
+        else in_kernel_fc2_reduce
+    )
+    if (
+        effective_ikr
+        and knobs
+        and knobs.get("token_back_mode")
+        not in (
+            None,
+            "epi_warps",
+        )
     ):
         knobs = {**knobs, "token_back_mode": "epi_warps"}
     cfg = with_knobs(cfg, knobs)
--- a/flashinfer/moe_ep/kernel_src/cutedsl_megamoe/shim/mxfp8.py
+++ b/flashinfer/moe_ep/kernel_src/cutedsl_megamoe/shim/mxfp8.py
@@ -867,11 +867,21 @@ def get_symm_buffer_for_mxfp8_mega_moe(
             topk=num_topk,
             max_tokens=num_max_tokens,
         )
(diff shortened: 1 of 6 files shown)


mhoqueanik:
**`shim/mxfp8.py` sanitizer bypass:**

> Good catch — fixed. The sanitizer now derives `effective_ikr = knobs.get("in_kernel_fc2_reduce", in_kernel_fc2_reduce)` and keys off whichever value `with_knobs()` will actually apply, so a caller-supplied knobs dict carrying `in_kernel_fc2_reduce=True` plus a non-`epi_warps` `token_back_mode` is sanitized before the frozen-config `dataclasses.replace` can raise.

**docstring "2000 iterations":**

> Clarified: the confirmed livelock repro ran with a `NUM_ITERS=2000` env override; the shipped script defaults to 60. The docstring now says so explicitly.

**ruff-format on repro scripts:**

> Reformatted (`ruff format` + end-of-file) in the same commit; `pre-commit run -a` is green on the branch.

mhoqueanik:
@flashinfer-bot run

Anerudhan (approved):

Issue #4534, opened 2026-08-15
trtllm-gen MLA: numCtasZ (= batch size) still unguarded against the gridDim.z limit (65535), and the failed cuLaunchKernelEx only fprintf's — output silently left unwritten (cf. #3125)


zkyue:
## Context

#3125 correctly diagnosed this in April and the maintainer direction there was "just raise an error" — but that was never implemented and the issue was closed. As of flashinfer 0.6.17:

- `include/flashinfer/trtllm/fmha/fmhaKernels.cuh:516` still computes `int numCtasZ = params.mBatchSize;` with no cap, and `:388` passes it to `cuLaunchKernelEx` as `gridDim.z`.
- `include/flashinfer/trtllm/common.h:44-50` — `cuErrCheck_` **only `fprintf`s and returns** when `cuLaunchKernelEx` fails. A launch-configuration error is not a sticky context error, so nothing downstream notices either.

CUDA's `gridDim.z` maximum is 65535, so any caller with a pseudo-batch of ≥ 65536 gets a failed launch whose only symptom is one stderr line — and an **unwritten output tensor**.

## Fresh evidence (end-to-end, from the sglang side)

sgl-project/sglang#34947 just hit this in production-shaped usage: sglang's DSA `prefill=trtllm` path reuses the decode kernel and flattens every query token into its own batch entry, so a single unchunked extend > 65535 tokens drives `mBatchSize` past the limit. Measured on SM100 / B200:

- Exact onset: **65535 works, 65536 does not** (one-token bisection).
- A sentinel-filled `out` tensor comes back **100 % untouched**; no Python exception is raised.
- torch-profiler shows **0 of `num_layers` attention kernels** — the model returns garbage and the benchmark reports a **falsely fast** number (the broken 131072-token prefill is "1.80x faster" than the correct 65535-token one because all attention work silently disappears).

## Minimal reproducer (flashinfer only, no sglang, ~30 s)

```python
import sys, torch, flashinfer

B = int(sys.argv[1])                       # pseudo-batch == number of query tokens
NH, TOPK, KVL, ROPE, HD = 16, 2048, 512, 64, 576
PAGE, PAGES, SENT = 64, 8192, -31337.0

kv  = torch.zeros(PAGES, 1, PAGE, HD, dtype=torch.float8_e4m3fn, device="cuda")
ws  = torch.empty(256 << 20, dtype=torch.uint8, device="cuda")
[...]

PR #3238, opened 2026-05-06
trtllm_batch_decode_with_kv_cache_mla trtllm-gen backend cum_seq_lens_q support

saltyminty:
## 📌 Description

Adds `cum_seq_lens_q` support to the `trtllm-gen` path in `trtllm_batch_decode_with_kv_cache_mla`.

## 🔍 Related Issues

https://github.com/flashinfer-ai/flashinfer/issues/3131

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

## 🧪 Tests

- [x] Tests have been added or updated as needed.
- [x] All tests are passing (`unittest`, etc.).

Validation run:

- `pre-commit run --files flashinfer/mla/_core.py flashinfer/trace/templates/attention.py tests/attention/test_trtllm_gen_mla.py`
- `python3 -m py_compile flashinfer/mla/_core.py flashinfer/trace/templates/attention.py tests/attention/test_trtllm_gen_mla.py`
- `git diff --check`
- Remote SM100 GPU pytest in FlashInfer CI container:
  - `tests/attention/test_trtllm_gen_mla.py::test_trtllm_batch_decode_mla[True-True-False-trtllm-gen-False-False-2-64-dtype1-1.0-4-layer_dimensions0]`
  - `tests/attention/test_trtllm_gen_mla.py::test_trtllm_batch_decode_mla[True-False-False-trtllm-gen-False-False-2-64-dtype1-1.0-4-layer_dimensions0]`

## Reviewer Notes

The new argument is query-side metadata only. `seq_lens` and `max_seq_len` continue to describe KV cache lengths; `max_q_len` is derived from `cum_seq_lens_q` and is not exposed as a new public API argument.

## Summary by CodeRabbit

* **New Features**
  - Support for variable-length per-request (ragged) MLA queries in batch decode, with optional cumulated-sequence-lengths and max-query-length handling and runtime dispatch between dense and ragged traces.
  - New public trace variants for dense and ragged per-token MLA decodes and a dispatching trace entrypoint.

* **Tests**
  - Added and extended tests covering dense/ragged decode paths, cumulated-sequence-lengths workflows, validation, and mismatch/error cases.

Issue #3125, opened 2026-04-20
trtllm-gen attention error on large batch size


nvjullin:
Batch size (number of tokens) maps directly to numCtasZ in https://github.com/flashinfer-ai/flashinfer/blob/main/include/flashinfer/trtllm/fmha/fmhaKernels.cuh#L380
CUDA's limit for gridDim.z is 65535, so cuLaunchKernelEx returns CUDA_ERROR_INVALID_VALUE on batch size >65535.
However, no python exception is thrown and if users don't notice the error in the logs, will run with incorrect results.
This affects all trtllm-gen attention kernels.

The following is a reproducer (`trtllm_batch_decode_with_kv_cache_mla` is used in prefill so hitting batch size >65535 is possible)
```py
import sys
import torch
import flashinfer

BATCH_SIZE = int(sys.argv[1]) if len(sys.argv) > 1 else 65536

# GLM-5 shapes
NUM_HEADS = 64
KV_LORA_RANK = 512
QK_ROPE_HEAD_DIM = 64
QK_NOPE_HEAD_DIM = 192
HEAD_DIM_QK = KV_LORA_RANK + QK_ROPE_HEAD_DIM  # 576

SPARSE_MLA_TOP_K = 2048
PAGE_SIZE = 32
MAX_SEQ_LEN = SPARSE_MLA_TOP_K
NUM_PAGES = MAX_SEQ_LEN // PAGE_SIZE

DEVICE = "cuda:0"

torch.manual_seed(0)

query = torch.randn(BATCH_SIZE, 1, NUM_HEADS, HEAD_DIM_QK, dtype=torch.bfloat16, device=DEVICE)

kv_cache = torch.randn(NUM_PAGES, 1, PAGE_SIZE, HEAD_DIM_QK, dtype=torch.bfloat16, device=DEVICE)

token_indices = torch.arange(SPARSE_MLA_TOP_K, dtype=torch.int32, device=DEVICE)
block_tables = token_indices.view(1, 1, SPARSE_MLA_TOP_K).expand(BATCH_SIZE, 1, SPARSE_MLA_TOP_K).contiguous()

seq_lens = torch.full((BATCH_SIZE,), MAX_SEQ_LEN, dtype=torch.int32, device=DEVICE)

workspace = torch.zeros(256 * 1024 * 1024, dtype=torch.int8, device=DEVICE)

bmm1_scale = 1.0 / (HEAD_DIM_QK**0.5)

[...]

PR #4536, opened 2026-08-15
fix(trtllm-gen): reject fmha launches whose batch dimension exceeds the CUDA gridDim.z limit

zkyue:
## Motivation

Implements the check proposed by the maintainer in #3125 ("just raise an error"), which never
landed. Filed as #4534.

trtllm-gen fmha maps the batch dimension straight onto `gridDim.z`
(`numCtasZ = params.mBatchSize`). CUDA caps `gridDim.z` at 65535; beyond that,
`cuLaunchKernelEx` returns `CUDA_ERROR_INVALID_VALUE`, which `cuErrCheck` only prints to stderr.
The call then returns normally with the output buffer untouched — a **silent no-op** visible to
callers only as wrong results. A sentinel-filled-output micro-reproducer shows the exact onset:
batch 65535 writes output; 65536 leaves 100 % of the output untouched with no Python exception
(one stderr line). This is reachable in practice from sglang's DSA path, which flattens query
tokens into the batch dimension (sgl-project/sglang#34947).

## Modifications

One `FLASHINFER_CHECK(launch_config.gridDimZ <= 65535, ...)` immediately before
`cuLaunchKernelEx` (+13 lines), surfacing as a Python `RuntimeError` through the FFI. The comment
documents why only Z is guarded: gridDim.x carries Q tiles (limit 2^31−1), and gridDim.y
(head × head-dim splits) would need more than 2048 query heads to reach 65535.

Deliberately NOT changed: `cuErrCheck`'s print-only semantics (wide blast radius across the
trtllm tree) and the callers doing the flattening — the caller-side guard is
sgl-project/sglang#34948, and the structural cure for MLA prefill is the already-merged
`cum_seq_lens_q` path (#3238), which callers can adopt.

## Verification (B200)

| batch (= gridDim.z) | before | after |
|---|---|---|
| 65535 | output written, no error | unchanged (output written, no error) |
| 65536 | output 100 % untouched, `exception=None`, 1 stderr line | `RuntimeError` naming the dimension, the value, and the remedy |
| 65536, unpatched control in the same env | silent | silent (proves the flip is this change) |

clang-format (pinned) clean.

## Related

- #3125 (this check), #3238 (structural cure for the MLA-prefill flattening)
- #4534 (issue tracking this guard; the print-only `cuErrCheck` half is deliberately
  left open there)
- sgl-project/sglang#34947 (end-to-end report), sgl-project/sglang#34948 (caller-side guard)

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for large batch dimensions before launching CUDA operations.
  * Prevents invalid launches when the grid’s depth exceeds the supported limit and reports a clear error instead.


zkyue:
**CI status: the single red job is a PyPI 502, not this change.**

`AOT Build Import (arm64, cu129)` (run 31895229042, job 95037467515) failed inside `pip install`,
while resolving a *torch* dependency:

```
Collecting nvidia-cudnn-cu12==9.20.0.48 (from torch->flashinfer-python==0.6.18)
  Obtaining dependency information for nvidia-cudnn-cu12==9.20.0.48 from
  https://files.pythonhosted.org/packages/0c/77/...aarch64.whl.metadata
ERROR: Could not install packages due to an OSError.
pip._vendor.urllib3.exceptions.ResponseError: too many 502 error responses
...
ERROR: Failed to install flashinfer package
```

The memory diagnostics printed by the same step rule out the OOM class this job instruments for
(`oom_kill 0`, peak cgroup 1.1 GiB). `fail-fast: true` then cancelled the other 14 matrix entries,
which is why `Test Results Summary` is red — no entry got far enough to compile the change.
Everything outside the gated matrix is green (pre-commit, docs, public-API check, Claude review,
CodeRabbit).

The diff is one `FLASHINFER_CHECK` in a header and cannot influence wheel downloads, so I don't
think there is anything to fix on my side. `main` has moved 3 commits since the base and none of
them touch `include/flashinfer/trtllm/`, so the branch is still current and a rebase would only
churn the SHA — and per `pr-test.yml` a push does not authorize CI on its own anyway (only applying
`run-ci` does). A rerun of the failed/cancelled jobs is all this needs; I don't have the repo role
to issue the bot command myself. Happy to force-push a rebase instead if that is the preferred way
to get a fresh trigger.

PR #4545, opened 2026-08-16
Add ReplaySSM autotuning

askliar:
## Summary

Adds framework-agnostic runtime autotuning for ReplaySSM checkpointing SSU. For eligible dense decode workloads, FlashInfer profiles supported monolithic and two-kernel launches and caches the fastest tactic for reuse.

## What changes

- Exposes main-kernel pipeline stages and CTAs per SM as runtime launch controls, alongside the existing precompute heads-per-CTA control.
- Adds a `CheckpointingSSURunner` that tunes the algorithm, `d_split`, main launch geometry, and precompute grouping through FlashInfer's generic autotuner.
- Buckets the active batch dimension while profiling compact private state/ring buffers, preserving the production `dt` broadcast layout and representative replay history.
- Uses cold-L2 and CUDA-graph profiling, and exposes a scratch allocator so inference frameworks can provide graph-stable two-kernel workspace.
- Keeps explicit algorithm/launch overrides and unsupported configurations on the normal fallback path.

## Scope

Generic autotuning is enabled only for dense calls (`cu_seqlens is None`) with the two-kernel scratch buffers present and launch controls left on automatic selection. The implementation contains no vLLM-specific policy or state.

## Validation

- Base-head H100 focused suite: 7 passed, covering scratch allocation, dynamic profiles, cache keys, tactic generation, `d_split`, autotune cache reuse, and monolithic/two-kernel parity.
- Review-followup H100 checks: 3 focused tests passed, and the newly admitted non-power-of-two launch (`heads_per_group=14`, `heads_per_cta=3`) compiled and executed successfully.
- H100 `autotune(True)` wall clock at `max_batch=1024`: **63.856 seconds** with a warm JIT module and empty autotuner cache. This used dense `T=4`, BF16 inputs, FP32 state, and Nemotron TP8 geometry (`nheads=16`, `ngroups=1`, `head_dim=64`, `d_state=128`); all 12 batch buckets were saved (3,574 distinct tactic profiles after launch deduplication).
- Pre-commit and public API/documentation checks pass. The full repository test matrix is pending maintainer CI authorization.


askliar:
@flashinfer-bot run

ishovkun:
None of these are blockers:

1. [Medium] The scratch-allocator migration stops halfway through a function the PR itself rewrites.
tests/mamba/test_checkpointing_ssu.py:563 — the author adds the canonical allocate_checkpointing_ssu_scratch, converts _two_kernel_scratch (L127) to it, then leaves five inline torch.empty scratch-triple copies (L563, 849, 928, 1031, 1147) — L563 sitting inside _run of test_two_kernel_matches_monolithic, edited extensively in this very diff (the two_kernel→algorithm rename is in the same hunk). Each copy re-derives k_old = ((max_window+7)//8)*8 — exactly the drift the helper exists to prevent. Remedy: route all five through the helper.

2. [Medium] _two_kernel_scratch hardcodes num_predicted_tokens=1.
tests/mamba/test_checkpointing_ssu.py:127-129. It works only because token_pad = ceil16(n) is 16 for every legal n≤16, so 1 and the real T give identical shapes today. If the MMA-M padding ever changes, every caller silently gets wrong cb_scaled/cumAdt_vec shapes with no failing assert at the allocation site. Pass the caller’s real T; it costs nothing.

3. [Low] Profiling pre-hook can write into user tensors via the contiguity filter.
_prepare_checkpointing_ssu_profile_inputs mutates ring_start/prev_num_accepted_tokens/state_batch_indices — safe when they’re cloned into the profile arena, but profile_arena_inputs filters on .is_contiguous(), so a non-contiguous ring_start (unusual but legal) is excluded from the arena and the hook then mutates the caller’s tensor during tuning. Tuning-time only; still, either force-include the three mutated tensors in the arena or fail loudly when they can’t be.


Nits
- Cached TuningConfig bakes the first call’s contiguity snapshot: get_tuning_config caches per-runner, but profile_arena_inputs depends on per-call .is_contiguous() which is not in the runner cache key. Their own test_autotune_runner_key_is_synthesis_invariant asserts stride-invariance of the key while the config content is stride-dependent — a latent inconsistency (tuning-time impact only).
- _make_tactics is functools.cached on raw batch (unbounded key space in principle; bounded by buckets in practice).
New tests reach into tuner privates (tuner._file_configs, profiling_cache) — brittle against autotuner refactors.
- cache_size = batch + 15 — undocumented magic; a comment saying it exercises cache-capacity > batch under tuning would do.

ishovkun (approved):

bkryu:
@flashinfer-bot run

flashinfer/mamba/checkpointing_ssu.py
```diff
@@ -66,6 +88,420 @@ def _get_module(
...
+    heads_per_cta_candidates = tuple(
+        heads_per_group >> shift
+        for shift in range(heads_per_group.bit_length())
+        if heads_per_group % (heads_per_group >> shift) == 0
+    )
+    tactics = [(0, 0, 0, monolith_d_split)]
+    total_work = two_kernel_d_split * batch * num_heads
+    seen_launches: set[tuple[int, int, int]] = set()
+    for stages in (1, 2):
+        for ctas_per_sm in (1, 2, 4, 8, 16):
```
hnover-nv: Nit:
There's no need for this to be a power of 2.  I'd suggest sweeping it exhaustively (1..32) locally and see where the winners are on h100, b200, b300 and then re-allocate your search.
Or you could do it in two stages.  Search coarsely, then search a few ctas_per_sm around each winner or something.  Probably overkill?
Also, if compilation has already happened by this point, you could use utilities to see how many blocks even CAN fit on an SM and bound it that way.  Although it might be more than you want to search over every time.  You could aim for 1 more than the actual bound, which might mitigate tail effects as these kernels don't do dynamic load balancing.

flashinfer/mamba/checkpointing_ssu.py
```diff
@@ -66,6 +88,420 @@ def _get_module(
...
+        state_batch_indices.copy_(
+            torch.arange(1, batch + 1, device=state.device).to(
+                state_batch_indices.dtype
+            )
+        )
+    slots = torch.arange(state.size(0), device=state.device)
+    ring_start.copy_((slots % x_cache.size(2)).to(ring_start.dtype))
+    max_window = x_cache.size(2) - x.size(1)
+    prev_num_accepted_tokens.copy_(
+        ((slots * max_window) % (max_window + 1)).to(prev_num_accepted_tokens.dtype)
```
hnover-nv, line 186: Nit / fyi:
This is making PNAT uniform?  Due to how the flushing works it isn't, but it's probably fine.  If acceptance length was markov and we knew the acceptance lengths you can get the PNAT distribution and draw from it randomly (my microbenchmark did this, using some super v3 acceptance lengths), but that's all overkill, plus we don't have those numbers in general.
This might explain some mismatch between your #'s and Igor's, though, he may have tuned for a different scenario, in terms of both T and the PNAT distro.

flashinfer/mamba/checkpointing_ssu.py
```diff
@@ -66,6 +88,420 @@ def _get_module(
...
+    ]
+    if inputs[13] is not None:  # z
+        batch_inputs.append(13)
+    if inputs[15] is not None:  # state_batch_indices
+        batch_inputs.append(15)
+    dynamic_specs = (
+        DynamicTensorSpec(
+            input_idx=tuple(batch_inputs),
+            dim_idx=(0,) * len(batch_inputs),
+            # Powers of two through 256, steps of 256 through 2048, steps of
```
hnover-nv, line 261: thorough!  Just to check, has tensor and other parallelisms properly flowed down to this stage?  Want to make sure we look up the right "batch 1" tuning result later.

flashinfer/mamba/checkpointing_ssu.py
```diff
@@ -323,14 +801,15 @@ def checkpointing_ssu(
...
-        ``batch * nheads >= sm_count``.  The crossover collapses in
-        ``batch * nheads`` across nheads (measured at TP 8/4/2) and state widths:
-        the monolith wins <= 128 work-units and the split wins from 256 on a
-        148-SM B200 (mixed-PNAT + conv1d/PDL bench; ties take the split).
-        ``"two-kernel"`` forces the split (scratch quartet required); ``"monolith"``
-        forces the monolith (scratch ignored).  Benches/tests that must pin the
-        path should force it.
+        With the scratch trio and no explicit tuning knobs, ``"auto"`` uses
+        FlashInfer's cached autotuner tactic.  Inside an ``autotune(True)``
+        context, it profiles monolithic against every supported combination of
```
hnover-nv: only monolithic?  I thought when scratch is provided it does both?  I think this comment is maybe off.

hnover-nv (commented):
Some nits

flashinfer/mamba/checkpointing_ssu.py
askliar: @hnover-nv Ran a few tests on GB200 with updated grid:

| Batch | Winner | SSU-only time |
|---:|---|---:|
| 1 | M(2) | 10.240 µs |
| 2 | M(2) | 12.288 µs |
| 4 | M(2) | 14.336 µs |
| 8 | M(2) | 16.384 µs |
| 16 | K(1,6,8,1) | 21.472 µs |
| 32 | K(1,14,8,1) | 29.696 µs |
| 64 | K(1,21,8,1) | 45.088 µs |
| 128 | K(1,29,8,1) | 74.752 µs |
| 256 | K(1,28,8,1) | 134.080 µs |
| 512 | K(2,2,8,1) | 253.952 µs |

Looks like indeed non-pow of 2 options work well!

flashinfer/mamba/checkpointing_ssu.py
askliar: It's not per-se super realistic and assumes that PNAT is static across a bunch of scenarios but I think it's a good enough proxy for tuning happening during an actual run.

flashinfer/mamba/checkpointing_ssu.py
askliar: Yeah, it does, tested that. Also, added `(compute_capability, SM count)` as a key just to make sure we're clear there.

flashinfer/mamba/checkpointing_ssu.py
askliar: fixed!

askliar:
FYI also ran all tests on H100 and GB200 - all pass fine so think we're good to merge.

bkryu:
@flashinfer-bot run

kahyunnam (approved):

flashinfer/mamba/checkpointing_ssu.py
hnover-nv: I take your comment as we should do non-power of 2 (is that the 6,14,21,... columm above?)  But it still looks like 1,2,4,8,16 is what is being searched?

flashinfer/mamba/checkpointing_ssu.py
hnover-nv: Sorry, I'm still confused.  The comment to me says you are only tuning monolithic in the auto-tuner, and so preusmably only picking monolithic in that path?  But my recollection of the code is otherwise.

flashinfer/mamba/checkpointing_ssu.py
hnover-nv: nm, shows it now.  probably my error.

kahyunnam:
@flashinfer-bot run

flashinfer/mamba/checkpointing_ssu.py
hnover-nv: Discussed elsewhere.  tuning it all sgtm!

hnover-nv (approved):

kahyunnam:
@flashinfer-bot run

qiching (approved):
LGTM.

Could you add a measured autotune(True) wall-clock time to the PR description (H100 or GB200, max_batch=1024)? At ~8.4k tactic-profiles users will otherwise assume it hung.

Also, in `_make_tactics` please drop the `heads_per_group % (heads_per_group >> shift) == 0` filter to match the C++ dispatch_heads_per_cta halving chain, for non-power-of-2 `heads_per_group` it excludes the value the launcher's own heuristic picks (e.g. 3 at hpg=14).

kahyunnam:
@flashinfer-bot run

askliar:
@flashinfer-bot run

askliar:
@qiching updated it! Please, take a look.

qiching (approved):
LGTM, approved

qiching:
@flashinfer-bot run

nvpohanh:
cc @YAMY1234 for vis

PR #4380, opened 2026-08-06
feat(sm120): consolidate DSV4 sparse MLA top-k 192/256 support

lucifer1004:
## 📌 Description

This PR consolidates the community work needed to run DeepSeek-V4-Flash-0731 with DSpark through the SM120/SM121 sparse MLA backend.

A downstream DSpark configuration has 133 active sparse entries per query: 128 sliding-window entries plus 5 draft entries. The index tensor is padded to 192 because the kernel consumes 64-entry tiles. Other integrations use the 256 bucket. Neither shape was fully represented in the existing SM120 decode and prefill dispatch matrices, so decode-sized requests could miss the standalone kernel and fall through to the prefill-only orchestrator, which aborts on `num_tokens <= 64`.

This change:

- adds DSV4 decode instantiations for `topk=192` and `topk=256` with 8, 16, 32, 64, and 128 query heads;
- adds DSV4 single-cache prefill dispatch for `topk=192` and `topk=256` using BF16 QK;
- supports TP8 prefill by padding H8 into the existing 16-head MG tile while gating all global Q, sink, output, and LSE accesses;
- enables the padded H8 MG path for both single-cache and dual-cache prefill, including both supported secondary-cache page sizes and runtime top-k lengths;
- keeps the Python and CUDA dispatch matrices and user-facing documentation synchronized;
- raises an actionable Python `ValueError` when an unsupported decode shape would otherwise enter the prefill-only kernel and abort the process;
- fixes the sparse MLA benchmark KV cache view to match its default HND layout; and
- expands correctness and benchmark coverage for the new dispatch shapes, active-length truncation, HND/NHD layouts, caller-owned workspaces, and output-buffer reuse.

## 🔍 Related Issues and Community PRs

After coordination with the maintainers, this PR is intended as the single consolidated replacement for the overlapping open implementations:

- #4309 — DSV4 `topk=192` decode/prefill and active-length coverage.
- #4372 — broad DSV4 `topk=256` decode/prefill dispatch and public API coverage.
- #3989 — combined DSpark `topk=256` fix and benchmark HND layout correction.
- #3817 — DSV4 `topk=256` decode instantiations and decode-to-prefill fail-fast behavior.
- #3834 — DSV4 `topk=256` BF16 prefill dispatch.

It also covers the implementation scope of the earlier closed alternatives #3896 and #3923. Related reports include #3828, #3988, and #4336.

The NVFP4 KV-cache extension in #3937 is a separate feature and is not superseded by this PR.

[...]

PR #4551, opened 2026-08-16
fix(sm120): explain sparse-MLA decode dispatch misses; add config query API

### LINKED ISSUE #4541: [sm120][dsv4] Sparse MLA decode-form calls with non-instantiated topk silently fall through to the prefill orchestrator
### Environment

FlashInfer 0.6.17, `_sparse_mla_sm120` JIT kernel family, GB10 (SM 12.1). Caller: vLLM `FLASHINFER_MLA_SPARSE_DSV4` backend, DeepSeek-V4-Flash-0731 with DSpark speculative decoding (2× DGX Spark, TP=2).

### Problem

The SM120 sparse-MLA decode kernels are instantiated for a fixed set of `(num_heads, topk)` pairs with `page_block_size=64` (in 0.6.17: topk ∈ {128, 512, 1024}). A call that is *semantically decode* (num_tokens ≤ 64) but carries a non-member topk does not fail dispatch — it **silently falls through to the varlen prefill orchestrator**, which then asserts:

```
num_tokens > 64 ... Decode ... must go through sparse_mla_sm120_decode_dsv3_2 or sparse_mla_sm120_decode_dsv4
```

For a 5-token speculative-decode call this message is maximally confusing: the caller *is* trying to reach the decode kernels; the library routed it away because of an unrelated shape parameter, and reports the failure in terms of token count. We lost most of a day to this exact message before finding the dispatch tables (`_decode_dsv4_dispatchable` in `flashinfer/mla/_sparse_mla_sm120.py`).

We're aware #4380 (merged) consolidates top-k 192/256 support, which removes the *specific* gap that bit us (DSpark non-causal width) on future releases — thank you! This issue is about the failure mode itself, which remains for any shape outside whatever the instantiated set is.

### Suggested fix

When a decode-form call (num_tokens ≤ 64) fails to match a decode instantiation, raise at dispatch time naming the actual mismatch, e.g.:

```
No SM120 sparse-MLA decode kernel for (num_q_heads=64, topk=256, page_block_size=64);
available topk: {128, 192, 256, 512, 1024}. Prefill fallback is invalid for q_len <= 64.
```

A `supported_sparse_mla_sm120_configs()` query API would additionally let callers validate at init instead of first-request time (vLLM now does an init-time check along these lines, see vllm-project/vllm#51538).

SamMausberg:
## Description

Follow-up to #4380 for #4541.

On SM120, sparse-MLA decode kernels are instantiated for a fixed set of
`(num_heads, topk)` pairs with `page_block_size=64`. Since #4380, a
decode-form call (`num_tokens <= 64`) that misses every decode instantiation
raises in Python instead of reaching the prefill orchestrator's C++
`num_tokens > 64` assert, but the error is a flat parameter dump that points
at private module internals (`_DECODE_DSV4_DISPATCH` /
`_DECODE_DSV3_2_DISPATCH`), so the caller still has to open the source to
learn what to change. #4541 asks for the two remaining pieces, added here:

1. **Dispatch-miss diagnosis.** The `ValueError` now names the actual
   mismatch and enumerates what is instantiated, e.g.:

   ```
   SM120 sparse-MLA has no decode kernel for this shape: num_tokens=5,
   num_heads=64, topk=384, d_qk=512, page_block_size=64, model_type=dsv4,
   extra_topk=0. Mismatch: topk=384 is not instantiated for num_heads=64;
   available topk: [128, 192, 256, 512, 1024]. The prefill orchestrator only
   serves num_tokens > 64, so a decode-form call must match a decode
   instantiation exactly. Query supported shapes at init time with
   flashinfer.mla.supported_sparse_mla_sm120_configs().
   ```

   The diagnosis distinguishes uninstantiated `topk` (for an instantiated
   head count), uninstantiated `num_heads`, both, `page_block_size != 64`,
   and `d_qk`/family mismatch. The pre-existing message prefix and shape
   summary (`no decode kernel ... num_tokens=..., num_heads=..., topk=...`)
   are kept so existing callers matching on it keep working.

2. **`flashinfer.mla.supported_sparse_mla_sm120_configs()`** — a public
   query API returning a `{"dsv4" | "dsv3_2" | "glm_nsa":
   SparseMLASm120DecodeConfig}` mapping (frozen dataclass with `d_qk`,
   `page_block_size`, `max_num_tokens`, `head_topk_pairs`, plus
   `supports_decode()` / `supported_topk()` / `supported_num_heads()`
   helpers). This lets callers such as vLLM validate a serving configuration
   at init time instead of poking at private dispatch tables
   (vllm-project/vllm#51538 currently has to do the latter). Exported
   lazily from `flashinfer.mla` following the existing prims-ts lazy-export
   pattern, and added to the `docs/api/attention.rst` autosummary.

No kernel or dispatch-behavior changes: every shape that dispatched to a
kernel before still dispatches identically; only the no-kernel error path
and the new query API changed.

## Related Issues

Fixes #4541. Follow-up to #4380. Motivating downstream check:
vllm-project/vllm#51538.

## Pull Request Checklist

### Pre-commit Checks

## Tests

Tested locally on an RTX 5070 Ti (SM120, CUDA 13.2, torch 2.8, JIT build from source):

- `pytest tests/attention/test_sparse_mla_sm120_dispatch.py` — **12 passed** (new file; pure Python, no GPU required)
- `pytest tests/attention/test_sparse_mla_sm120.py` — **310 passed** (full SM12x suite: all decode/prefill correctness tests plus the extended and new dispatch-failure tests)
- `pre-commit run --all-files` — all hooks pass (ruff check/format, mypy, whitespace/EOL)

## Reviewer Notes

- `tests/attention/test_sparse_mla_sm120_dispatch.py` is deliberately not
  gated on SM12x: it covers the config query and the message builder as pure
  Python so the diagnosis logic is exercised on any CI node. The actual raise
  inside the dispatcher stays covered by the (extended) SM12x-gated tests in
  `test_sparse_mla_sm120.py`.
- `supported_sparse_mla_sm120_configs()` exposes the instantiation sets by
  reference as `frozenset`s (immutable), so the public view can never drift
  from the dispatch tables.
- The `"glm_nsa"` entry aliases the `"dsv3_2"` config object since GLM-NSA
  decode shares the DSv3.2 instantiations; if the sets ever diverge, only the
  constructor needs to split.
- The new test file carries the same NVIDIA BSD-3 header as its sibling
  `test_sparse_mla_sm120.py` for consistency within the SM120 sparse-MLA
  family; happy to switch it to the Apache-2.0 header (or drop it) if you
  prefer for externally-authored tests.

Developed in combination with Claude Fable 5.

## Summary by CodeRabbit

- **New Features**
  - Added public configuration details for supported SM120 sparse MLA decode variants, including dimensions, page sizes, token limits, head counts, and top-k values.
  - Added validation helpers for checking supported decode configurations.
  - Improved access to these APIs through the MLA package.

- **Bug Fixes**
  - Decode errors now identify unsupported dimensions, page sizes, head counts, and top-k values.

- **Documentation**
  - Expanded API documentation for the configuration query and configuration type.

### DIFF
--- a/docs/api/attention.rst
+++ b/docs/api/attention.rst
@@ -217,6 +217,8 @@ PageAttention for MLA
     convert_compressed_page_aligned_sparse_indices_to_hca_metadata
     DSV4HCAMetadata
     xqa_batch_decode_with_kv_cache_mla
+    supported_sparse_mla_sm120_configs
+    SparseMLASm120DecodeConfig
 
 .. note::
 
--- a/flashinfer/mla/__init__.py
+++ b/flashinfer/mla/__init__.py
@@ -22,9 +22,16 @@ _PRIMS_TS_LAZY_EXPORTS = frozenset(
     }
 )
 
+_SPARSE_MLA_SM120_LAZY_EXPORTS = frozenset(
+    {
+        "SparseMLASm120DecodeConfig",
+        "supported_sparse_mla_sm120_configs",
+    }
+)
+
 
 def __getattr__(name: str):
-    """Resolve PrimTS MLA APIs without loading their runtime at import."""
+    """Resolve lazily-exported MLA APIs without loading their runtime at import."""
 
     if name in _PRIMS_TS_LAZY_EXPORTS:
         from ..attention.prims_ts import mla_decode as prims_ts_mla_decode
@@ -32,8 +39,17 @@ def __getattr__(name: str):
         value = getattr(prims_ts_mla_decode, name)
         globals()[name] = value
         return value
+    if name in _SPARSE_MLA_SM120_LAZY_EXPORTS:
+        from . import _sparse_mla_sm120
+
+        value = getattr(_sparse_mla_sm120, name)
+        globals()[name] = value
+        return value
     raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
 
 
 def __dir__():
-    return sorted(set(globals()) | _PRIMS_TS_LAZY_EXPORTS)
+    """Include lazily-exported names alongside the module globals."""
+    return sorted(
+        set(globals()) | _PRIMS_TS_LAZY_EXPORTS | _SPARSE_MLA_SM120_LAZY_EXPORTS
+    )
--- a/flashinfer/mla/_sparse_mla_sm120.py
+++ b/flashinfer/mla/_sparse_mla_sm120.py
@@ -38,17 +38,23 @@ The user-facing sparse MLA entry points are
 for the DSv3.2 / GLM sparse top-k path. This module keeps only the SM120
 implementation hooks used by those dispatchers and focused kernel
 tests/benchmarks.
+
+Decode kernels are instantiated for a fixed set of ``(num_heads, topk)``
+pairs; ``flashinfer.mla.supported_sparse_mla_sm120_configs`` enumerates them
+so callers can validate a configuration at init time.
 """
 
 from __future__ import annotations
 
 import functools
 import os
+from dataclasses import dataclass
 from types import SimpleNamespace
 from typing import List, Optional
 
 import torch
 
+from ..api_logging import flashinfer_api
 from ..autotuner import (
     AutoTuner,
     ConstraintSpec,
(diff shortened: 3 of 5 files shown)


SamMausberg:
Hello, @bkryu, can you please take a look at this follow-up to #4380 when you get a chance?

bkryu:
Thanks @SamMausberg  for this PR. 

#4802 has been merged with preserved authorship so I will be closing. We look forward for more contributions from the community.

PR #4552, opened 2026-08-16
Declutter output by collating all tests not run due to runner timeout

righthandabacus:
## 📌 Description

Currently, when a unit test is terminated due to a timeout, the list of test nodes could be marked as failed. Then, in the summary, all failures are printed with the failure message and other details. Since a runner timeout can cause a lot of test nodes to fail, the list of failures will be very long and cause the summary to be cluttered. This PR removes these (since the failures are not to be fixed one by one) but reports a summary count instead.

## 🔍 Related Issues

Nil. Internal discussion only.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Reviewer Notes

Unit test failures can be due to an actual failure because the test did not pass, or a "synthetic failure" that is assigned to a test because of a timeout. The former is still reported with the failure message. The latter is the focus of this PR. Synthetic failures can be caused by the following reasons:

- The node executing when the timeout occurred.
- Nodes that were never started.
- Nodes that completed earlier in the killed batch, since their temporary results are also discarded.
- Unit timeout versus overall runner deadline on a per-node basis.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved test result summaries for timed-out tests.
  * Timeout failures are now clearly distinguished from regular failures.
  * Timeout summaries use accurate singular or plural wording, while regular failures retain detailed diagnostics and links.

### DIFF
--- a/scripts/test_sharding/junit.py
+++ b/scripts/test_sharding/junit.py
@@ -11,6 +11,9 @@ from .io import atomic_write_xml
 from .models import base_function_for_nodeid, source_file_for_nodeid
 
 
+TIMEOUT_DIAGNOSTIC = "not executed due to timeout"
+
+
 @dataclass(frozen=True)
 class TestCaseResult:
     nodeid: str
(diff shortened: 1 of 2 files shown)


dierksen (approved):

righthandabacus:
Superceded by https://github.com/flashinfer-ai/flashinfer/pull/4895

PR #4572, opened 2026-08-18
fix(gdn): honor intermediate cache view strides

yyihuang:
## Description

Fix GDN MTP intermediate-state cache indexing when a caller reuses a backing cache whose physical token capacity is larger than the current `T`.

This keeps the public `gated_delta_rule_mtp` API and semantics unchanged. The caller allocates `[B, cache_steps, HV, V, K]` and passes the sliced view `cache[:, :T, ...]`, whose logical shape remains `[B, T, HV, V, K]`. The backend validates contiguous inner modes, derives the physical capacity from `stride(0) / (HV * V * K)`, and creates a zero-copy compact alias over the same storage for the existing kernels. There is no `.contiguous()` copy, temporary allocation, or public `cache_steps` argument.

The derived physical capacity is used by the FP32 inline/warp routes and the BF16 ILP4/wide-vector/T1 routes. Dense BF16 cache addressing remains widened to 64-bit. Output, state-update, and valid checkpoint semantics are unchanged.

Related to #4254.

## Tests

Exact source: `d705bbfbe4c9a266e41a780bcd045d37a1cb65e7`; source archive SHA256 `6e0547dcf0878b7f82febcb05f68434fb0431a0d2a5c0c917b500079b2a396b5`; CUTLASS DSL 4.7.0; Torch 2.13 nightly + CUDA 13.3.

- B200 / SM100, driver 595.58.03, Slurm job `3762590`: `6 passed`.
- B300 SXM6 AC / SM103, driver 580.126.09, Slurm job `276037`: `6 passed`.
- Covered routes: FP32 inline, FP32 warp, BF16 ILP4, BF16 wide-vector, BF16 T1 wide-vector, and the >2^31-element dense-cache boundary.
- Padded-cache cases compare the sliced-capacity view against an exact-`T` cache at the fixed BF16 tolerance `atol=rtol=1e-2` and require hidden trailing slots to remain bit-exact sentinels.

This is a correctness-only addressing repair; it does not make a performance claim.

### DIFF
--- a/flashinfer/gdn_decode.py
+++ b/flashinfer/gdn_decode.py
@@ -696,11 +696,11 @@ def gated_delta_rule_mtp(
         Pre-allocated output tensor of shape ``[B, T, HV, V]``.
     intermediate_states_buffer : torch.Tensor, optional
         Buffer for caching intermediate states, shape
-        ``[B, cache_steps, HV, V, K]`` with ``cache_steps >= T``
-        (first dim is indexed per-batch, not per-pool-slot — buffer must
-        be at least ``B`` rows and contiguous; must be float32 when
-        provided). When ``None``, intermediate states are not cached.
-        Mutually exclusive with ``ssm_state_indices``.
+        ``[B, T, HV, V, K]`` (first dim is indexed per-batch, not per-pool
+        slot; must be float32 when provided). A leading slice of a larger
+        cache, for example ``cache[:, :T]``, is supported; its batch stride
+        determines the physical cache capacity. When ``None``, intermediate
+        states are not cached. Mutually exclusive with ``ssm_state_indices``.
     ssm_state_indices : torch.Tensor, optional
         Per-token pool scatter indices of shape ``[B, T]`` and dtype
         ``torch.int32``.  When provided, the kernel writes each intermediate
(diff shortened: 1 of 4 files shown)


yyihuang:
Pushed `df8e4af1` to close the remaining padded-cache hole identified in review.

- propagates physical `cache_steps` through the BF16 T=1 wide-vector kernel, launcher, compile arguments, and cache key
- fixes the flat cache offset to use `batch * cache_steps * HV`
- aligns the MTP buffer contract with the existing `buffer_size == B` assertion
- adds a direct B=2, T=1, HV=64, `cache_steps=4` regression that checks requested cache rows and bit-exact trailing sentinels

Targeted dual-architecture GPU validation used the exact PR source at `df8e4af19b43d17bbc1964bdc8ea2e005420c8ab` and CUTLASS DSL 4.7.0. Both runners forced the BF16 backend import before pytest, so skipped tests cannot be counted as a pass:

- B200 / SM100, job `3756633`: BF16 T=2 wide-vector + BF16 T=1 wide-vector, `2 passed`
- GB300 / SM103, job `512688`: same two cells, `2 passed`

The earlier public-main FP32/BF16 padded-cache matrix remains covered by jobs `3727353` / `3727354`; this update specifically closes the T=1 wide-vector route that that matrix did not reach.

yyihuang:
@flashinfer-bot run

yyihuang:
Follow-up `4d38f631ef` addresses the review finding on dense BF16 cache address width. All four batch-scoped dense-cache formulas now widen `i_n` to `cutlass.Int64` before multiplying by `cache_steps * HV`; the dynamic wide-vector predeclaration is Int64 as well.

A direct CUDA regression crosses the first unsafe address at B=257, T=1, cache_steps=8, HV=64, K=V=128 (4.02 GiB cache), and compares the final batch output/cache against a B=1 reference at the fixed BF16 `atol=rtol=1e-2`. The state pool remains compact, so this tests the actual >2^31-element address without unrelated allocation.

Dual-architecture forced-import qualification:
- B200 job `3757174`, `umb-b200-249`, NVIDIA B200 CC10.0, driver 595.58.03: 3/3 passed, exit `0:0`.
- GB300 job `512799`, `nvl72d022-T17`, NVIDIA GB300 CC10.3, driver 580.167.08: 3/3 passed, exit `0:0`.
- CUTLASS DSL 4.7.0; source archive SHA256 `74148ef2a60a3bc5b68da10b9bffe5b5465e91144b32320d85d6c58b831d2bb8`.

@flashinfer-bot run

yyihuang:
Formatting-only follow-up a1aae91f applies the exact Ruff layout reported by pre-commit. `git diff --check` and `py_compile` pass; kernel semantics are unchanged from dual-architecture-qualified parent 4d38f631 (B200 3757174, GB300 512799, 3/3 each).

@flashinfer-bot run

yzh119 (commented):
I'm trying to understand the motivation of this PR.

Does it mean we allow user to reuse a cached state buffer for different T? It sounds reasonable to me.

I have no problem about updating the backend-specific APIs (e.g. cute-dsl/cake implementations). But I'll be conservative about changing public APIs, if there are more elegant solutions.

For this specific case, I suppose our user-facing API is `gated_delta_rule_mtp`, we actually don't need to change the API/semantics:

At user side, they can specify a sliced tensor:
```
cached_state = torch.empty((B, cache_steps, HV, V, K), ...)
sliced_cached_state = cached_state[:, :T, ...] # [B, T, HV, V, K]
```
sliced_cached_state would carry all the information we need for the underlying implementations (strides would encode size of cache_steps), and it should be compatible with original API.

yyihuang:
@yzh119 Agreed. I changed the implementation in `d705bbfbe4c9a266e41a780bcd045d37a1cb65e7` to preserve the original public API and exactly use the sliced-view contract you described:

```python
cache = torch.empty((B, cache_steps, HV, V, K), ...)
sliced_cache = cache[:, :T, ...]  # logical [B, T, HV, V, K]
```

The wrapper now derives physical capacity from the view's batch stride, validates contiguous inner modes, and makes a zero-copy compact alias over the same storage for the existing backend. No public `cache_steps` argument or semantic change remains, and there is no `.contiguous()` temporary.

I also updated the regressions to pass sliced views and require hidden trailing slots to remain bit-exact. Exact-commit validation passed on both targets:

- B200 / SM100 job `3762590`: `6 passed`
- B300 SXM6 AC / SM103 job `276037`: `6 passed`

The six cells cover FP32 inline/warp, BF16 ILP4/wide/T1, and the >2^31-element dense-cache boundary.

PR #4574, opened 2026-08-18
fix(moe): expose tactic-dependent MXFP8 activation scale layout

VertexC:
## Summary

- Preserve the selected TRT-LLM FP4 MoE FC1 scale layout in tensor rank: rank 4 for `SWIZZLED_8x4`, rank 5 for `SWIZZLED_128x4`.
- Add opt-in `return_gemm1_output_scale=True` to `trtllm_fp4_block_scale_routed_moe` while keeping the return-list length from #3987 unchanged by default.
- Validate MXFP4×MXFP8 LoRA intermediates with the kernel-returned scales, including a large-M case that selects the 128-token tile.

## Motivation

`gemm1_lora_delta` exposes the post-activation FC1 output so callers can compute a down-projection LoRA delta. MXFP8 consumers also need that output's block scales.

The physical scale layout is tactic-dependent:

- token tiles up to 64 write `SWIZZLED_8x4`;
- 128/256-token tiles write `SWIZZLED_128x4`.

The layouts have the same element count. An opaque flat scale tensor therefore cannot be decoded safely: applying the 8x4 decoder to a 128x4 buffer silently maps NaN padding scales onto routed rows rather than raising a shape error.

## API

When `gemm1_lora_delta` is set and `return_gemm1_output_scale=True`, the finalized result becomes:

```text
[output, expanded_idx_to_permuted_idx, gemm1_output, gemm1_output_scale]
```

The scale tensor's shape carries graph-safe host metadata without a device scalar or stream synchronization:

```text
(M_tiles, K_tiles, 8, 4)       -> SWIZZLED_8x4
(M_tiles, K_tiles, 32, 4, 4)   -> SWIZZLED_128x4
```

The underlying allocation size and data pointer are unchanged. With the new option left at its default `False`, the public result remains exactly the three tensors introduced by #3987.

## Validation

- [x] `pre-commit run --files ...` (clang-format, ruff, mypy, standard hooks)
- [x] Downstream semantic replay using unchanged kernel bytes: previously failing 5,725-token and unchunked 26,600-token MXFP4×MXFP8 LoRA prefills both become finite when the flat scale buffer is reshaped using this metadata contract
- [x] Decoder checks for both 8x4 and 128x4 physical layouts
- [ ] `pytest tests/moe/test_trtllm_gen_fused_moe.py::test_moe_lora_delta` on compiled extension (CI / downstream wheel build in progress)

## Relationship to existing work

#3987 added FP4 `BiasType::Mn` / `gemm1_lora_delta` and exposed the FC1 activation. This PR completes the auxiliary-output contract for MXFP8 consumers by optionally returning the corresponding scales and their layout.

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

- **New Features**
  - FP4 block-scale MoE execution can now optionally return GEMM1 activation scale data alongside activation outputs.
  - Added support for scale layouts across different tile sizes, including large-matrix transitions.
  - Existing result formats remain unchanged unless scale output is explicitly requested.

- **Bug Fixes**
  - Improved FP4 LoRA intermediate-output dequantization using kernel-provided scale metadata.

- **Tests**
  - Expanded coverage for large-token MoE LoRA scenarios and tactic-specific scale layouts.

### DIFF
--- a/csrc/trtllm_fused_moe_kernel_launcher.cu
+++ b/csrc/trtllm_fused_moe_kernel_launcher.cu
@@ -494,6 +494,10 @@ struct MoeRunResultBuffers {
   // Optional activation output.
   // Sync with flashinfer/fused_moe/core.py:_unpack_trtllm_moe_output.
   Optional<Tensor> activation_output;
+  // Optional block scales for a quantized activation output. Physical layout is
+  // preserved in tensor rank: rank 4 is SWIZZLED_8x4, rank 5 is SWIZZLED_128x4.
+  // Sync with flashinfer/fused_moe/core.py:_unpack_trtllm_moe_output.
+  Optional<Tensor> activation_output_scale;
   // Whether FFI[0] is finalized output rather than an unfinalized FC2 buffer.
   // Sync with flashinfer/fused_moe/core.py:_unpack_trtllm_moe_output.
   bool is_finalized{true};
@@ -515,6 +519,9 @@ struct MoeRunResultBuffers {
     }
     if (activation_output.has_value()) {
       tensors.push_back(activation_output.value());
+    }
+    if (activation_output_scale.has_value()) {
+      tensors.push_back(activation_output_scale.value());
     }
     return tensors;
   }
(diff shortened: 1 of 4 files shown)


VertexC:
@flashinfer-bot run

VertexC:
Downstream compiled validation update:

- Built and published matching x86/ARM wheel sets from the v0.6.13 backport.
- Installed the compiled Python/JIT-cache/cubin wheel trio directly into a B300 dev container.
- Formerly failing 5,725-token MXFP4×MXFP8 LoRA prefill: finite.
- Unchunked 26,600-token prefill (`[16384, 10215]`): finite.
- Exact benchmark completed with 20k/18k cached and 100k/98k cached prompts, 500 forced output tokens, with no NaNs.

Upstream CI still requires maintainer authorization; `@flashinfer-bot run` was requested above.

murphymatt:
@aleozlx can you help triage this?

PR #4599, opened 2026-08-19
Adding var-window FMHA context support for PrimsTS.

mingxu1067:
## 📌 Description

1. Adding variable attention window support which allows each Q token has its own window range.
2. Performance:
- Q/K/V/O shape: [64, 2560, 8, D], D=128 or 256.
- Dtype: [FP15, FP8]
- 4 Seqence in one batchL T1[0:256], I1[256:1280], T2[1280:1536], I2[1536:2560]
- Base mask is causal with `left_window=128`.
- `I1` and `I2` additionally allow `right_window=128`, capped at their own segment end.

|  D  | Dtype  | FmhaTs ms | TRT-LLM-GEN ms |
| --- | ------ | --------- | -------------- |
| 256 |  fp16  |   0.6066  |     0.6554     |
| 128 |  fp16  |   0.3538  |     0.4027     |
| 256 |  fp8   |   0.3898  |     0.4424     |
| 128 |  fp8   |   0.3361  |     0.3231     |

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

* **New Features**
  * Added variable-window attention support for packed sequences.
  * Planning and prefill APIs now accept per-query window start and end bounds.
  * Supports configurable attention ranges with CUDA int32 boundary tensors.
  * Improved window-boundary handling across query tiles for more accurate masking.
  * Added support for longer text and image sequences with variable attention windows.

* **Bug Fixes**
  * Added validation for window bounds, incompatible mask settings, and output aliasing.
  * Paged variable-window attention now reports a clear unsupported-configuration error.

* **Tests**
  * Added correctness coverage across multiple head dimensions, sequence lengths, and boundary cases.

### DIFF
--- a/flashinfer/attention/prims_ts/context.py
+++ b/flashinfer/attention/prims_ts/context.py
@@ -1701,10 +1701,15 @@ class BatchPrefillTSWrapper:
             Fixed or packed query, key, and value tensors.
         qo_indptr, kv_indptr : torch.Tensor, optional
             Cumulative query and K/V offsets for packed-ragged input.
-        mask_type : {"dense", "causal"}
-            Attention mask mode.
+        mask_type : {"dense", "causal", "variable_window"}
+            Attention mask mode. ``variable_window`` is supported only for
+            fixed-shape inputs.
         window_left : int
             Left sliding-window extent, or ``-1`` to disable the window.
+        variable_window_token_starts, variable_window_token_ends : torch.Tensor, optional
+            Inclusive per-query K bounds required for ``variable_window``.
+            Both must be CUDA int32 tensors shaped ``[B, Sq]`` and satisfy
+            ``0 <= starts[b, q] <= ends[b, q] < Sk``.
         sm_scale : float, optional
             Softmax scale; defaults to the inverse square root of head size.
         output_scale : float
@@ -2107,10 +2112,15 @@ def batch_prefill(
         Fixed or packed query, key, and value tensors.
     qo_indptr, kv_indptr : torch.Tensor, optional
         Cumulative query and K/V offsets for packed-ragged input.
-    mask_type : {"dense", "causal"}
-        Attention mask mode.
+    mask_type : {"dense", "causal", "variable_window"}
+        Attention mask mode. ``variable_window`` is supported only for
+        fixed-shape inputs.
     window_left : int
         Left sliding-window extent, or ``-1`` to disable the window.
+    variable_window_token_starts, variable_window_token_ends : torch.Tensor, optional
+        Inclusive per-query K bounds required for ``variable_window``. Both
+        must be CUDA int32 tensors shaped ``[B, Sq]`` and satisfy
+        ``0 <= starts[b, q] <= ends[b, q] < Sk``.
     sm_scale : float, optional
         Softmax scale; defaults to the inverse square root of head size.
     output_scale : float
--- a/flashinfer/attention/prims_ts/context.py
+++ b/flashinfer/attention/prims_ts/context.py
@@ -220,9 +220,10 @@ def _validate_device(device: torch.device) -> int:
 def _validate_mask(mask_type: str) -> None:
     if not isinstance(mask_type, str):
         raise TypeError("mask_type must be a string")
-    if mask_type not in ("dense", "causal"):
+    if mask_type not in ("dense", "causal", "variable_window"):
         raise ValueError(
-            f"mask_type must be exactly 'dense' or 'causal', got {mask_type!r}"
+            "mask_type must be exactly 'dense', 'causal', or "
+            f"'variable_window', got {mask_type!r}"
         )
 
 
@@ -238,6 +239,40 @@ def _validate_window_left(window_left: int, mask_type: str) -> int:
     if window_left > 0 and mask_type != "causal":
         raise ValueError("a positive window_left requires mask_type='causal'")
     return window_left
+
+
+def _validate_variable_window_bounds(
+    starts: Optional[torch.Tensor],
+    ends: Optional[torch.Tensor],
+    *,
+    geometry: _ContextGeometry,
+) -> tuple[torch.Tensor, torch.Tensor]:
+    """Validate inclusive per-query K bounds for contiguous attention."""
+    if starts is None or ends is None:
+        raise ValueError(
+            "variable_window_token_starts and variable_window_token_ends are "
+            "required for mask_type='variable_window'"
+        )
+    if geometry.packed:
+        raise NotImplementedError(
+            "variable_window currently requires fixed [B, S, H, D] tensors"
+        )
+    expected_shape = (geometry.batch_size, geometry.max_seq_len_q)
+    for tensor, name in (
+        (starts, "variable_window_token_starts"),
+        (ends, "variable_window_token_ends"),
+    ):
+        _validate_tensor(tensor, name)
+        if tensor.device != geometry.device:
+            raise ValueError(f"{name} must be on {geometry.device}, got {tensor.device}")
+        if tensor.dtype != torch.int32:
+            raise ValueError(f"{name} must have dtype torch.int32, got {tensor.dtype}")
+        if tuple(tensor.shape) != expected_shape:
+            raise ValueError(
+                f"{name} must have shape {expected_shape}, got {tuple(tensor.shape)}"
+            )
+        _validate_compact(tensor, name, "[B, Sq]")
+    return starts.flatten(), ends.flatten()
 
 
 def _validate_scale(value: object, name: str) -> float:
@@ -1010,6 +1045,7 @@ def _get_compiled_context(
     input_dtype = dtype_map[q_dtype_key]
     output_dtype = dtype_map[output_dtype_key]
     is_causal = mask_type == "causal"
+    has_variable_window = mask_type == "variable_window"
     # Construct the scheduler-independent topology first. Immutable
     # single-instance domains can use the lower-overhead static persistent
     # queue; paired or live-ragged domains are reconstructed with CLC.
@@ -1021,6 +1057,7 @@ def _get_compiled_context(
         d=head_dim,
         is_persistent=True,
         is_causal=is_causal,
+        has_variable_window=has_variable_window,
         balance_causal_workload=_uses_heavy_first_static_causal_raster(
             mask_type=mask_type,
             window_left=window_left,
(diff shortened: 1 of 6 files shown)


flashinfer/attention/prims_ts/context.py
```diff
@@ -1580,8 +1636,10 @@ def plan(
         *,
         qo_indptr: Optional[torch.Tensor] = None,
         kv_indptr: Optional[torch.Tensor] = None,
-        mask_type: Literal["dense", "causal"] = "dense",
+        mask_type: Literal["dense", "causal", "variable_window"] = "dense",
         window_left: int = -1,
+        variable_window_token_starts: Optional[torch.Tensor] = None,
+        variable_window_token_ends: Optional[torch.Tensor] = None,
```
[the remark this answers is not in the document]
rosenrodt: +1 on this.

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_kernel.py
```diff
@@ -515,6 +516,60 @@ def __init__(
...
+        if kwargs.get("schedule") is None:
+            raise ValueError("VariableWindow domain tasks require a captured schedule")
+        super().__init__(**kwargs)
+        self._variable_window_token_starts = variable_window_token_starts
+        self._variable_window_token_ends = variable_window_token_ends
+        self._variable_window_cta_starts = variable_window_cta_starts
+        self._num_kv_tiles = num_kv_tiles
+        self._q_stride = q_stride
+        self._cta_m = cta_m
+        self._kv_n = kv_n
```
PerkzZheng: I would rename it as tile_size_q and tile_size_kv. please help clean other codes if needed. Thanks!

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_kernel.py
```diff
@@ -515,6 +516,60 @@ def __init__(
...
+        variable_window_token_starts: cute.Tensor,
+        variable_window_token_ends: cute.Tensor,
+        variable_window_cta_starts: cute.Tensor,
+        num_kv_tiles: int | Int32,
+        q_stride: int | Int32,
+        cta_m: int,
+        kv_n: int,
+        seq_idx: int,
+        batch_idx: int,
+        offset: int = 0,
```
PerkzZheng, line 535: what offset is used for here ? we can remove it if not used.

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_kernel.py
```diff
@@ -515,6 +516,60 @@ def __init__(
         )
 
 
+class VariableWindowDomainTask(Task):
+    """Task whose K-loop domain comes from packed-Q window endpoints."""
+
+    def __init__(
```
PerkzZheng, line 524: it would be good if you can add more comments about what each parameter means here.

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_resources.py
```diff
@@ -106,6 +107,57 @@
...
+    valid_bits: Int32,
+    score0: Float32,
+    score1: Float32,
+    score2: Float32,
+    score3: Float32,
+) -> tuple[Float32, Float32, Float32, Float32]:
+    """Expand four bitmap bits with setp and replace invalid scores.
+
+    Use and/setp because CUDA 13.3 libNVVM fails to compile the equivalent
+    `r2p.b32 {valid0, valid1, valid2, valid3}.reverse, {$r0}.b0, 0x0f;`.
```
PerkzZheng: we cannot expose r2p publicly IIUC.

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_resources.py
```diff
@@ -2907,6 +3045,61 @@ def packed_dense_k_masked_row_max(
...
+                Int32(tmem_x),
+            )
+            valid_bits = _bmsk_clamp(chunk_left, chunk_right - chunk_left)
+            chunk = s_data[chunk_idx]
+            masked_scores = []
+            for quad_idx in cutlass.range_constexpr(tmem_x // 4):
+                quad_base = quad_idx * 4
+                # CUDA 13.3 NVVM rejects r2p in this kernel, so expand each
+                # four-bit group with ordinary setp instructions.
+                masked_scores.extend(
```
PerkzZheng, line 3087: we can skip masking for dense tiles.

PerkzZheng:
Also, do you know why prims_ts has better performance compared to trtllm-gen ? just want to know the optimization tricks. Thanks!

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_kernel.py
mingxu1067: Done.

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_kernel.py
mingxu1067: It is used to derive two task-domain lengths from the same variable-window K-tile span, like.
```
domain_n_kwargs = {**base_kwargs, "offset": 0}
domain_n_minus_1_kwargs = {**base_kwargs, "offset": 1}
...
last_k_tile - first_k_tile - self._offset
```
Refer to [fmha_kernel.py#L591](https://github.com/mingxu1067/flashinfer/blob/mingh/prim_ts_variable_fmha_context/flashinfer/attention/prims_ts/kernels/fmha_context/fmha_kernel.py#L591)

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_kernel.py
mingxu1067: Added.

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_resources.py
mingxu1067: I tried few ways to skip it, like conditional around `_mask_score_quad`, but got slightly perf regression. The `variable_window_token_starts/ends` are CUDA tensors, therefore we cannot decide the skip at compiling time.

mingxu1067:
> Also, do you know why prims_ts has better performance compared to trtllm-gen ? just want to know the optimization tricks. Thanks!

I found there is a benchmark mismatch. TRT-LLM-Gen was using pagedKV but Prims-TS using continuous KV. By fixing this, the performance reduce from 20% to around 5-10%. I updated the perf table of this PR as well.
The remaining difference should come from that Prims-TS loads `window-start/end` outside K-loop and but TRT-LLM-Gen loads them inside K-loop.

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_resources.py
mingxu1067: Removed.

PerkzZheng:
> > Also, do you know why prims_ts has better performance compared to trtllm-gen ? just want to know the optimization tricks. Thanks!
> 
> I found there is a benchmark mismatch. TRT-LLM-Gen was using pagedKV but Prims-TS using continuous KV. By fixing this, the performance reduce from 20% to around 5-10%. I updated the perf table of this PR as well. The remaining difference should come from that Prims-TS loads `window-start/end` outside K-loop and but TRT-LLM-Gen loads them inside K-loop.

thanks. LGTM

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_resources.py
PerkzZheng: okay. it looks good to me. Previously, we assume window_start/end should always increase as offset_q so the boundary can be determined by the first token or last token. Anyway, it is something we can revisit later.

PerkzZheng (approved):

saltyminty (approved):
LGTM (mainly checked plumbing and didn't look too closely at kernel changes since Perkz already approved)

qsang-nv:
**`cache_variable_window_bounds` reads the bound tensors out of bounds for padded Q rows.**

In `fmha_resources.py` it computes `packed_q = batch_coord * q_stride + local_q` and indexes the flattened `[B * Sq]` tensors without clamping `local_q` to `Sq - 1`. `VariableWindowDomainTask.get_domain` does clamp the same coordinate (`last_local_q = min(..., q_stride - 1)`), and `_build_variable_window_cta_starts` pads its reduction with `_INT32_MAX`, so the padded tail is handled everywhere except here.

Any `Sq` that is not a multiple of the Q CTA tile (256 for D128, 128 for D256) leaves padding rows in the last Q tile, and for the last batch `packed_q` runs past the allocation: with `B=2, Sq=33, head_dim=128` it reaches `1 * 33 + 255 = 288` while the tensors hold 66 elements. These rows do not normally affect valid outputs when the kernel completes, because their O stores are dropped by the TMA bounds. However, the accesses remain undefined and may be reported by compute-sanitizer or fault near an allocation boundary. Both new tests use tile-aligned `Sq` (2560, 256), so the path is uncovered.

Suggest clamping `local_q` as `get_domain` does, plus a regression case with unaligned `Sq` and `B > 1` (batch offsets are currently untested).

mingxu1067:
> **`cache_variable_window_bounds` reads the bound tensors out of bounds for padded Q rows.**
> 
> In `fmha_resources.py` it computes `packed_q = batch_coord * q_stride + local_q` and indexes the flattened `[B * Sq]` tensors without clamping `local_q` to `Sq - 1`. `VariableWindowDomainTask.get_domain` does clamp the same coordinate (`last_local_q = min(..., q_stride - 1)`), and `_build_variable_window_cta_starts` pads its reduction with `_INT32_MAX`, so the padded tail is handled everywhere except here.
> 
> Any `Sq` that is not a multiple of the Q CTA tile (256 for D128, 128 for D256) leaves padding rows in the last Q tile, and for the last batch `packed_q` runs past the allocation: with `B=2, Sq=33, head_dim=128` it reaches `1 * 33 + 255 = 288` while the tensors hold 66 elements. These rows do not normally affect valid outputs when the kernel completes, because their O stores are dropped by the TMA bounds. However, the accesses remain undefined and may be reported by compute-sanitizer or fault near an allocation boundary. Both new tests use tile-aligned `Sq` (2560, 256), so the path is uncovered.
> 
> Suggest clamping `local_q` as `get_domain` does, plus a regression case with unaligned `Sq` and `B > 1` (batch offsets are currently untested).

Added

qsang-nv (approved):

PR #4479, opened 2026-08-12
feat(b12x): support MXFP4 dense GEMM on SM120

ayrnb:
## 📌 Description

Enables **MXFP4** (block size 32, E8M0 scale factors) for the **b12x** (CuTe DSL) FP4 dense GEMM kernel on SM120, alongside the existing NVFP4 (block size 16, E4M3) path.

Ports the dense GEMM portion of #3541 onto current `main`.

### Changes

**Kernel** (`flashinfer/gemm/kernels/dense_blockscaled_gemm_sm120_b12x.py`)
- Select `MmaMXF4Op` when `sf_vec_size == 32`, keep `MmaMXF4NVF4Op` for 16 — the cutlass-dsl wrapper hard-codes `sf_vec_size`/`sf_dtype` per op, so the two modes need separate ops. Matches the pattern already used by the SM12x MoE kernels.
- Add `_collapse_to_vmk()` and `_align_sf_copy_ranks()` to normalize scale-factor view ranks. MXFP4 carries one extra non-degenerate K sub-mode (`blk_sf // mma_nsf` is 2 for MXFP4 vs 1 for NVFP4), which the rank-3 mainloop indexing and the SMEM→RMEM SF copies both require folded back. Both helpers are conditional, so NVFP4 behavior is unchanged.

**Backend dispatch** (`flashinfer/gemm/gemm_base.py`)
- Drop the `use_nvfp4` restriction in `_b12x_gemm_fp4_requirement`.
- Make `sf_vec_size`/`sf_dtype` dynamic in `B12xFp4GemmRunner` (`16`/`Float8E4M3FN` for NVFP4, `32`/`Float8E8M0FNU` for MXFP4).
- Let `_heuristic_func_mm_fp4` prefer b12x for MXFP4 as well on SM120 + CUDA 13.

**Benchmark harness** (`benchmarks/routines/gemm.py`)
- Extend the `alpha_for_cute_dsl_mxfp4` whitelist to include `b12x`. MXFP4 passes `alpha=None`, and b12x — being a CuTe DSL kernel — needs the same unit-alpha tensor as `cute-dsl`.

### Performance

Hardware: **NVIDIA RTX PRO 5000 Blackwell 72GB** (SM120), CUDA 13, `N=2048, K=2048`, bf16 output, `use_128x4_sf_layout`. Latency in ms, `speedup = cudnn / b12x`.

[...]

PR #4602, opened 2026-08-19
fix(moe): restore SM12x MoE kernels broken by self-resolved helper in borrowed dense methods

lucifer1004:
# fix(moe): restore SM12x MoE kernels broken by self-resolved helper in borrowed dense methods

## Summary

#4479 (MXFP4 dense GEMM on SM120) added the `_collapse_to_vmk` staticmethod to
`DenseGemmKernel` and called it as `self._collapse_to_vmk(...)` inside
`_partition_fragment_SFA` / `_partition_fragment_SFB`. Those two methods are
also invoked **unbound** by the SM12x MoE kernels
(`MoEStaticKernel`, `MoEMicroKernel`, `MoEDynamicKernel`,
`MoEGatedDynamicKernel`), which pass their own instance as `self`:

```python
tCrSFA = self._dense_cls._partition_fragment_SFA(self, ...)
```

Since none of the MoE kernel classes define `_collapse_to_vmk`, every NVFP4
SM12x MoE path now fails at kernel compile time:

```
AttributeError: 'MoEGatedDynamicKernel' object has no attribute '_collapse_to_vmk'
```

This makes `main` unusable for SM12x fused MoE (e.g. vLLM's
`flashinfer_b12x` NVFP4 MoE backend serving Qwen3.5-397B-A17B-NVFP4 on
4x RTX PRO 6000 Blackwell).

## Fix

Move `_collapse_to_vmk` to a module-level function in
`dense_blockscaled_gemm_sm120_b12x.py` (next to the existing module-level
helpers) and call it directly instead of via `self.`. The helper is pure —
it never touches instance state — so nothing needs to live on the class, and
borrowers become immune to this class of change for this helper.

Net diff: one method moved, two call sites updated. No behavior change.

## Regression test

Adds a CPU-only borrow-contract test section in `tests/moe/test_b12x_fused_moe.py`
(pure source scanning via AST; imports no kernel module, so it collects and
runs even where the CuTe DSL stack is unavailable) that enforces
the structural contract the borrowing pattern relies on:

- discovers borrower classes by scanning the `fused_moe` sources for
  `self._dense_cls.<method>(self, ...)` call sites — new kernels are covered
  without editing the test;
- computes the transitive set of `self.<attr>` reads inside each borrowed
  `DenseGemmKernel` method and asserts the borrowing class provides every one;
- fails with an actionable message, e.g.
  `MoEStaticKernel borrows DenseGemmKernel._partition_fragment_SFA() but does
  not provide ['_collapse_to_vmk']; make the helper a module-level function or
  add it to MoEStaticKernel.`

Discovery is guarded by a non-empty assertion, so renaming `_dense_cls` or
changing the borrow idiom fails loudly instead of silently skipping the
parametrized test. Verified in both directions: the test passes on this PR
and fails on all four borrowers when the `self._collapse_to_vmk` call is
temporarily reintroduced.

## Testing

- New contract test: 4/4 pass (MoEStaticKernel, MoEMicroKernel,
  MoEDynamicKernel, MoEGatedDynamicKernel); auto-discovery verified by adding
  a temporary borrowing class, which was collected and failed as expected.
- End-to-end: vLLM `flashinfer_b12x` serving Qwen3.5-397B-A17B-NVFP4 on
  4x RTX PRO 6000 (SM120, CUDA 13) now starts, captures CUDA graphs
  (FULL_AND_PIECEWISE), and completes a full C1/C4/C16/C64 serving benchmark
  (8K/1K, 850 requests total) with zero errors.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of partitioned scale-factor tensors for dense matrix operations, including MXF4 formats.
  * Updated scale-factor processing to provide more consistent fragment partitioning.

* **Tests**
  * Added validation to ensure fused mixture-of-experts components correctly inherit required dense matrix operation behavior and dependencies.


bkryu:
@flashinfer-bot run

lucifer1004:
Review addressed in 5c93bf8: the borrow-contract test moved into tests/moe/test_b12x_fused_moe.py, and discovery now has a non-empty assertion so renaming `_dense_cls` or changing the borrow idiom fails loudly instead of silently skipping. All 5 contract tests pass locally (CPU-only). This push will need a fresh `run-ci` trigger — thanks!

bkryu:
@flashinfer-bot run

lucifer1004:
Both new CodeRabbit findings addressed in 4fe4cfd:

1. **Module-scope kernel import** — removed entirely instead of just guarded: the contract test now works purely on parsed sources (AST via `Path(__file__)`), so it collects and runs even where the CuTe DSL stack is unavailable, rather than skipping there.
2. **`_dense_cls` target matching** — broadened to class-level assignments, annotated assignments, and qualified values (`kernels.DenseGemmKernel`); unresolvable forms now fail with a message that names the resolved target instead of `None`.

Verified locally: 5/5 contract tests pass; reintroducing the `self._collapse_to_vmk` call fails all borrowers; a fake borrower with a class-level qualified assignment is discovered and reported correctly. pre-commit (ruff format/check, mypy) clean.

bkryu:
@flashinfer-bot run

aleozlx:
tests look good

bkryu (approved):

aleozlx:
## PR Review Screening

**CI verdict:** ✅ auto-run ok
**Review category:** `human` (rule fired: C6 durable kernel-internal change — shared `DenseGemmKernel` borrowed by four SM12x MoE kernels — with C7 adequate; no base-class/dispatch redesign, so not `live`)
**Blocking checks:** S2 (PR template replaced with custom format), S3 (template checklist absent)
**Release blocker:** 🚨 candidate — fixes `AttributeError` breaking every NVFP4 SM12x MoE compile on main, regression from #4479 (C8)
**Early stop:** no

### Security
| Q | Answer | Evidence |
|---|--------|----------|
| S2 template overwritten | yes | Body uses custom Summary/Fix/Testing headers; all 5 template sections missing |
| S3 template obligations | unmet | Checklist absent (0 checkboxes); Tests section is present with results |
| S4 agent-directed text | no | — |

### Packaging
| Q | Answer | Evidence |
|---|--------|----------|
| C1 dependency bump | no | — |
| C2 public API changes | no (internal-only) | `_collapse_to_vmk` staticmethod → module-level fn in `dense_blockscaled_gemm_sm120_b12x.py`; no exports/decorators touched |
| C3 AOT/trace registration | n-a | No new `gen_*_module()`; no new API |

### Presentation
| Q | Answer | Evidence |
|---|--------|----------|
| C4 perf claim backed | n-a | Bugfix; no perf claim |

### Implementation
| Q | Answer | Evidence |
|---|--------|----------|
| C5 experimental-track | no | — |
| C6 shared internals | yes (durable) | `flashinfer/gemm/kernels/dense_blockscaled_gemm_sm120_b12x.py` shared by dense GEMM + 4 MoE borrower classes; pure helper relocation, no behavior change |
| C7 tests match change | yes | CPU-only AST borrow-contract test in `tests/moe/test_b12x_fused_moe.py`; verified fail-before/pass-after; e2e SM120 serving run reported |
| C8 critical fix | yes | Body: SM12x fused MoE "fails at kernel compile time" on main since #4479; latest bot pipeline 16/16 pass |

Notes for the maintainer:
- Earlier `tests/moe` pipeline #63425940 showed `test_unified_moe_fuzz` failures on GB200/GB300 marked "attribution uncertain" (trtllm_mxint4 path, untouched by this PR); latest pipeline #63511699 passed 16/16.
- Regression is on main only (post-#4479), so this gates the next release rather than requiring a patch release.

aleozlx (approved):

PR #4603, opened 2026-08-19
feat(moe): allow B12xMoEWrapper to share pre-allocated workspaces

lucifer1004:
## Summary

`B12xMoEWrapper` pre-allocates its static workspace, dynamic workspace, and
output buffer per instance for CUDA graph compatibility. Frameworks that hold
one wrapper per MoE layer — e.g. vLLM's `flashinfer_b12x` NVFP4 MoE backend —
pay this cost once **per layer**.

For Qwen3.5-397B-A17B-NVFP4 (60 MoE layers, 512 experts, top-10 routing) with
`max_num_tokens=16384`, the per-layer buffers total ~68 GiB:

- static workspace: `(num_experts, max_rows, k/2)` packed input ≈ 640 MiB/layer
- dynamic workspace: ≈ 506 MiB/layer (routed_rows = 16384 × 10)
- output buffer: 128 MiB/layer

With weights taking ~52.5 GiB per GPU (TP4 on 96 GB RTX PRO 6000), the model
cannot start: workspace allocation OOMs during vLLM's memory profiling.

## Change

MoE layers execute strictly sequentially, so identically-shaped wrappers can
safely share a single set of buffers. Add optional constructor parameters:

- `shared_static_workspace`
- `shared_dynamic_workspace`
- `shared_output`

When provided, `_allocate_buffers()` reuses them instead of allocating.
Default behavior (no sharing) is unchanged.

With sharing, the same model's total wrapper memory drops from ~68 GiB to
~1.2 GiB and the server starts within the default
`gpu_memory_utilization=0.80` budget.

## Safety

- CUDA graphs record fixed buffer addresses; sharing keeps every layer's
  captured pointers valid, and graph capture preserves the sequential
  layer-to-layer stream order, so there are no cross-layer hazards.
- Workspace barrier state is epoch-based and restored after each launch, the
  same sequential-reuse pattern a single wrapper already relies on across
  forward passes.
- vLLM copies the returned output slice out immediately, so sharing the
  output buffer is safe for this caller pattern.

## Testing

- New CPU tests in `tests/moe/test_b12x_fused_moe.py`:
  - `test_wrapper_defaults_allocate_own_buffers` — default behavior unchanged
  - `test_wrapper_shared_buffers_are_reused` — injected buffers reused, no new
    allocations
- End-to-end on 4x RTX PRO 6000 (SM120, CUDA 13): vLLM `flashinfer_b12x`
  serving Qwen3.5-397B-A17B-NVFP4 with layer-shared workspaces starts,
  captures FULL_AND_PIECEWISE CUDA graphs, and completes a full
  C1/C4/C16/C64 serving benchmark (8K/1K, 850 measured requests) with zero
  errors and zero preemptions. Runtime validation was done on top of #4602,
  which fixes a separate SM12x MoE compile-time regression.

## Consumer

A corresponding vLLM change makes `FlashInferB12xExperts` share one buffer
set across layers (keyed by shape/device); it depends on this PR.

## Summary by CodeRabbit

* **New Features**
  * Added support for reusing shared workspace and output buffers with the B12x MoE wrapper when CUDA graph mode is enabled.
  * Automatically allocates only missing buffers across supported quantization modes.

* **Bug Fixes**
  * Improved default allocation so separate wrapper instances receive independent resources.
  * Added validation for compatible output shape, data type, and exact CUDA device.
  * Reports clearer errors when shared buffers are used outside graph mode or on the wrong device.

* **Tests**
  * Added coverage for shared-buffer requirements and device validation.


lucifer1004:
Both CodeRabbit findings addressed in ac0940b:

1. **`shared_output` validation** — the constructor now rejects incompatible injected output buffers before any CUDA graph capture: requires 2-D shape `(>=max_num_tokens, hidden_size)`, the configured `output_dtype`, and a matching device type, with the actual values in the error message. (Workspaces are opaque dispatcher structs, so validation is scoped to the output tensor.)
2. **Allocation spy in the sharing test** — `test_wrapper_shared_buffers_are_reused` now wraps `torch.empty` during the second wrapper's construction and asserts zero calls, so a regression that allocates before adopting the shared output would fail. Added `test_wrapper_shared_output_is_validated` for the new check.

Local: 3/3 wrapper buffer tests pass, pre-commit clean.

samuellees (approved):
LGTM

PR #4619, opened 2026-08-19
fix(moe): resolve _collapse_to_vmk on the dense class for b12x MoE

kahyunnam:
## 📌 Description

Every b12x fused MoE test fails on RTX Pro 6000 Blackwell with CUDA 13 with:

```
AttributeError: 'MoEStaticKernel' object has no attribute '_collapse_to_vmk'
```

(same for `MoEDynamicKernel`, `MoEGatedDynamicKernel`, `MoEMicroKernel` — 120 failures in the `v0.6.18rc4` unit-test pipeline).

The b12x MoE kernels deliberately do **not** subclass `Sm120B12xBlockScaledDenseGemmKernel`; they borrow its unbound methods and pass their own instance as `self`, e.g.

```python
tCrSFA_full = self._dense_cls._partition_fragment_SFA(
    self,  # type: ignore[arg-type]
    sSFA[None, None, 0], thr_mma, tidx,
)
```

`_partition_fragment_SFA`/`_SFB` reach the `@staticmethod` `_collapse_to_vmk` through that borrowed `self`, which resolves against the MoE class — and none of the four borrowers mirrors it. All four define `_thrfrg_SFA`/`_thrfrg_SFB` delegating wrappers, which is why only `_collapse_to_vmk` breaks.

This fix resolves the static helper on the class instead, which fixes all four borrowers in one place. `DenseGemmKernel` has no subclasses and nothing overrides `_collapse_to_vmk`, so the dense path is behaviour identical.

The failures are CUDA-13-only because the b12x fused MoE tests are gated on `_cuda_13_or_newer()`, not because the bug is CUDA-version dependent.

## 🔍 Related Issues

Regression from #4479 (`support MXFP4 dense GEMM on SM120`), which introduced `_collapse_to_vmk`. Not previously reported upstream.

### ✅ Pre-commit Checks

## 🧪 Tests

`test_borrowed_dense_methods_resolve_on_moe_kernels` derives the borrowed method set from the MoE modules and asserts that every class-level attribute those methods look up is mirrored on each borrower. It needs no GPU, so the next helper added this way fails on CPU instead of only on SM120 hardware. Verified it fails on the unfixed tree with

```
AssertionError: MoEStaticKernel borrows DenseGemmKernel._partition_fragment_SFA but is missing ['_collapse_to_vmk']
```

and passes with the fix. `tests/moe/test_b12x_fused_moe.py` is otherwise unchanged: 34 passed, 151 skipped (SM120-gated) on the SM100 box used here.

## Reviewer Notes

I do not have SM120 hardware, so the end-to-end b12x MoE kernels are not exercised locally — the fix is a pure attribute-resolution change and the new test reproduces the exact failure without a GPU, but a CI run on RTX Pro 6000 + CUDA 13 is the real confirmation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility for shared dense matrix multiplication methods used by mixture-of-experts kernels.
  * Prevented failures when these methods are reused across different kernel types.

* **Tests**
  * Added coverage verifying that all supported mixture-of-experts kernels provide the required shared functionality.

### DIFF
--- a/flashinfer/gemm/kernels/dense_blockscaled_gemm_sm120_b12x.py
+++ b/flashinfer/gemm/kernels/dense_blockscaled_gemm_sm120_b12x.py
@@ -600,7 +600,9 @@ class DenseGemmKernel:
         thr_vmk = (thr_vmnk[0], (thr_vmnk[1], thr_vmnk[3]))
         partitioned_sfa = thr_tensor[thr_vmk, (None, None)]
         partitioned_sfa = cute.group_modes(cute.flatten(partitioned_sfa), 0, 2)
-        partitioned_sfa = self._collapse_to_vmk(partitioned_sfa)
+        # Class-resolved: the b12x MoE kernels borrow this method with a foreign
+        # ``self`` that does not define ``_collapse_to_vmk``.
+        partitioned_sfa = DenseGemmKernel._collapse_to_vmk(partitioned_sfa)
         return cute.make_fragment_like(partitioned_sfa)
 
     def _partition_fragment_SFB(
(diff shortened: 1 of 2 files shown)


kahyunnam:
Closing as a duplicate of #4602, which landed the same `_collapse_to_vmk` borrow-contract fix (module-level helper instead of `self._collapse_to_vmk`) and is already merged into `main`.

Sorry for the extra PR — I opened this from the `v0.6.18rc4` unit-test failures before noticing #4602.

kahyunnam:
Duplicate of already-merged #4602.

PR #4466, opened 2026-08-11
feat: sm100 cute_dsl w4a16 gemm

IwakuraRein:
## 📌 Description

- Add sm100 cute-dsl nvfp4 w4a16 GEMM
- Rename `BlackwellDenseGemmBf16Fp4Kernel` to `Sm12xDenseGemmBf16Fp4Kernel`

### Performance

**N=6656, K=19968**

| M | torch-bf16 | Marlin MXFP8 | Marlin NVFP4 | FlashInfer cute-dsl |
|---|---:|---:|---:|---:|
| 1 | 46.22 µs (0.63×) | 35.28 µs (0.83×) | **29.17 µs (1.00×)** | 36.11 µs (0.81×) |
| 8 | 44.36 µs (0.66×) | 35.90 µs (0.82×) | **29.28 µs (1.00×)** | 36.14 µs (0.81×) |
| 32 | 46.58 µs (0.97×) | 51.08 µs (0.88×) | 45.09 µs (1.00×) | **36.54 µs (1.23×)** |
| 128 | 59.76 µs (1.77×) | 120.26 µs (0.88×) | 106.01 µs (1.00×) | **43.90 µs (2.41×)** |
| 512 | 129.56 µs (2.95×) | 436.68 µs (0.87×) | 382.09 µs (1.00×) | **125.71 µs (3.04×)** |
| 1024 | **213.10 µs (3.50×)** | 856.78 µs (0.87×) | 745.84 µs (1.00×) | 217.25 µs (3.43×) |
| 2048 | 415.42 µs (3.59×) | 1713.92 µs (0.87×) | 1491.82 µs (1.00×) | **408.66 µs (3.65×)** |
| 4096 | **795.58 µs (3.75×)** | 3427.44 µs (0.87×) | 2983.51 µs (1.00×) | 821.58 µs (3.63×) |

**N=19968, K=6656**

| M | torch-bf16 | Marlin MXFP8 | Marlin NVFP4 | FlashInfer cute-dsl |
|---|---:|---:|---:|---:|
| 1 | 45.38 µs (0.65×) | 35.78 µs (0.82×) | 29.38 µs (1.00×) | **25.41 µs (1.16×)** |
| 8 | 45.77 µs (0.65×) | 36.26 µs (0.82×) | 29.59 µs (1.00×) | **25.77 µs (1.15×)** |
| 32 | 44.40 µs (0.86×) | 41.51 µs (0.92×) | 38.24 µs (1.00×) | **26.06 µs (1.47×)** |
[...]

PR #4620, opened 2026-08-19
fix(trace): model the SM100/103 cute-dsl mm_bf16_fp4 prepared layout

kahyunnam:
## 📌 Description

`tests/trace/test_mm_bf16_fp4_reference_correctness.py::test_mm_bf16_fp4_reference_correctness[*-cute-dsl]` fails on the whole SM100 family (12 failures on B300 / GB200 / GB300, both CUDA versions, in the `v0.6.18rc4` unit-test pipeline):

```
    device = b.device
>   k_sf, n = b_descale.shape
E   ValueError: too many values to unpack (expected 2)
flashinfer/trace/templates/gemm.py:550: ValueError
```

`prepare_bf16_fp4_weights` gained a **third** prepared layout when the SM100 cute-dsl w4a16 GEMM landed. The trace layer only knew two:

| path | `b` | `b_descale` |
|---|---|---|
| cuDNN | canonical `(N, K//2)` uint8 | linear `(N, K_sf)` fp8-e4m3 |
| cute-dsl SM12x | `(K//16, N*2)` int32 tile-packed | `(K_sf, N)` uint8 S0E5M3 |
| cute-dsl SM100/103 | canonical `(N, K//2)` uint8 | **6-D** `(32, 4, N//128, 4, K_sf//4, 1)` strided view over the 128x4-swizzled buffer |

Two consequences on SM100/103:

- `mm_bf16_fp4_cute_dsl_trace.reference` unpacks `b_descale.shape` as 2-D and raises the `ValueError` above.
- `mm_bf16_fp4_trace_dispatch` keys off the weight dtype (`int32` → cute-dsl), and the SM100 weight is `uint8`, so a cute-dsl call silently resolves to the **cuDNN** template. Shapes are not validated at trace time, so this would have quietly dumped a definition labelled `mm_bf16_fp4_cudnn` describing the wrong prepared layout.

This PR adds `mm_bf16_fp4_cute_dsl_sm100_trace` with its own init and reference, restricts the SM12x init to SM12x, and gives the dispatch a discriminator for all three layouts.

The new reference recovers the canonical scale buffer by permuting the strided view back to its documented physical order — `convert_sf_to_mma_layout` returns logical `(outer_m, inner_m, m_tile, inner_k, k_tile, group)` over physical `(group, m_tile, k_tile, 32, 4, 4)`, which is exactly what `_unswizzle_sf_128x4` expects — then decodes the canonical nvfp4 weight the same way the cuDNN reference does. I checked that reconstruction against `_unswizzle_sf_128x4` of the pre-`convert` buffer, and independently against the documented element-wise index mapping; both are bit-exact.

## 🔍 Related Issues

Regression from #4466 (`feat: sm100 cute_dsl w4a16 gemm`), which added the SM100 prepared layout without touching the trace layer. Not previously reported upstream.

### ✅ Pre-commit Checks

## 🧪 Tests

Verified on a B200 (SM100, CUDA 13.0, cutlass-dsl 4.7.0), where the test reproduces the reported `ValueError` before the change:

- `tests/trace/test_mm_bf16_fp4_reference_correctness.py` — 4 failed → 4 passed
- `tests/gemm/test_mm_bf16_fp4.py` — 242 passed, 3 skipped
- `tests/trace/test_fi_trace_template_consistency.py`, `test_template_registry.py`, `test_template_init.py`, `test_rendered_source_standalone.py`, `test_fi_trace.py` — 1050 passed, 180 skipped

The reference test now picks the cute-dsl template matching the device and asserts `mm_bf16_fp4_trace_dispatch` resolves a real prepared call back to that template, so the misrouting cannot come back unnoticed.

## Reviewer Notes

Worth a second opinion on two judgement calls:

1. **A separate template rather than one branching reference.** The declared `axes`/`inputs` are what land in the dumped definition, and the SM100 weight and scale shapes differ from SM12x, so a single template cannot describe both honestly. If you would rather not trace this layout at all, the alternative is to make the SM12x init raise on SM100/103 and stop there — that also turns the failure into a skip, but leaves the dispatch mislabelling SM100 cute-dsl calls as cuDNN.
2. **Declaring a non-contiguous 6-D view as a template input.** It is an odd thing to put in a bench definition since it is a view, not an allocation. I described it as-is rather than substituting the underlying canonical buffer, because that view is literally what callers pass to `mm_bf16_fp4`. Happy to change the dim naming if you prefer something else.

I have no SM103 hardware; the SM100 and SM103 paths share `_prepare_cute_dsl_sm100`, so the gating treats them together.

## Summary by CodeRabbit

* **New Features**
  * Added support for BF16×FP4 operations on SM100/103 GPUs with six-dimensional scale tensors.
  * Automatically selects the appropriate processing path based on tensor format and GPU capability.

* **Bug Fixes**
  * Improved scale reconstruction and dequantization accuracy for supported SM100/103 workloads.

* **Tests**
  * Expanded correctness coverage for backend and GPU-specific template selection.
  * Tests now handle environments without CUDA gracefully.

### DIFF
--- a/flashinfer/trace/templates/gemm.py
+++ b/flashinfer/trace/templates/gemm.py
@@ -593,6 +593,34 @@ def _mm_bf16_fp4_cute_dsl_reference(a, b, b_descale, alpha=None, block_size=16):
     return _bf16_fp4_matmul(a, lut[codes] * sf, alpha)
 
 
+def _mm_bf16_fp4_cute_dsl_sm100_reference(a, b, b_descale, alpha=None, block_size=16):
+    """Reference for the SM100/103 cute-DSL-prepared layout.
+
+    b: [N, K//2] uint8 -- the canonical nvfp4 weight, passed through unchanged.
+    b_descale: (32, 4, N//128, 4, K_sf//4, 1) uint8 -- the strided view
+    ``convert_sf_to_mma_layout`` builds over the canonical 128x4-swizzled
+    FP8-E4M3 buffer, whose physical order is (groups, n_tiles, k_tiles, 32, 4,
+    4); permuting back to that order recovers the buffer to unswizzle.
+    """
+    n, k_half = b.shape
+    k = k_half * 2
+    sf_buffer = b_descale.permute(5, 2, 4, 0, 1, 3).reshape(-1)
+    sf_bytes = _unswizzle_sf_128x4(sf_buffer, n, k // block_size).contiguous()
+    sf = sf_bytes.view(torch.float8_e4m3fn).to(torch.float32)
+    lut = torch.tensor(_E2M1_VALUES, dtype=torch.float32, device=b.device)
+    b_int = b.to(torch.int64)
+    codes = torch.stack([b_int & 0xF, (b_int >> 4) & 0xF], dim=-1).reshape(n, k)
+    sf = sf.repeat_interleave(block_size, dim=1)
+    return _bf16_fp4_matmul(a, (lut[codes] * sf).T, alpha)
+
+
+cast(Any, _mm_bf16_fp4_cute_dsl_sm100_reference)._trace_reference_dependencies = (
+    _bf16_fp4_matmul,
+    _unswizzle_batched_sf_128x4,
+    _unswizzle_sf_128x4,
+)
+
+
 def _mm_bf16_fp4_cudnn_init(
     *,
     M: int,
@@ -658,7 +686,7 @@ def _mm_bf16_fp4_cute_dsl_init(
     device: str = "cuda",
     seed: int = 0,
 ):
-    """Build inputs for ``flashinfer.mm_bf16_fp4`` (cute-DSL backend).
+    """Build inputs for ``flashinfer.mm_bf16_fp4`` (cute-DSL backend, SM12x).
 
     Sourced from ``tests/gemm/test_mm_bf16_fp4.py``: quantize a randn
     bf16 weight via ``flashinfer.nvfp4_quantize`` (layout_128x4), then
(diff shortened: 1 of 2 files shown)


kahyunnam:
@flashinfer-bot run

flashinfer/trace/templates/gemm.py
```diff
@@ -678,6 +706,66 @@ def _mm_bf16_fp4_cute_dsl_init(
...
+    from flashinfer import (  # noqa: PLC0415
+        nvfp4_quantize,
+        prepare_bf16_fp4_weights,
+    )
+    from flashinfer.quantization.fp4_quantization import SfLayout  # noqa: PLC0415
+
+    if not torch.cuda.is_available() or torch.device(device).type != "cuda":
+        raise NotImplementedError("mm_bf16_fp4 init requires a CUDA device")
+    major, minor = torch.cuda.get_device_capability(torch.device(device))
+    if (major, minor) not in ((10, 0), (10, 3)):
```
dhiraj113, line 764: `_mm_bf16_fp4_cudnn_init` and `_mm_bf16_fp4_cute_dsl_init` both gate on `mm_bf16_fp4.is_backend_supported(...)`; this one checks only the arch tuple. On SM100/103 without a usable cute-dsl backend it falls through to `nvfp4_quantize(..., backend="cute-dsl")` instead of raising `NotImplementedError`.

```python
if not mm_bf16_fp4.is_backend_supported("cute-dsl", major * 10 + minor):
    raise NotImplementedError(f"mm_bf16_fp4 is not supported on SM{major}{minor}")
```

Needs `mm_bf16_fp4` back in the local import block.

flashinfer/trace/templates/gemm.py
```diff
@@ -782,17 +870,69 @@ def _mm_bf16_fp4_cute_dsl_init(
...
+    },
+    inputs={
+        "A": Tensor(["M", "K"], param="a", description="Activation, bfloat16."),
+        "B": Tensor(
+            ["N", "K_div_2"],
+            param="b",
+            description="Weight, fp4 e2m1fn_x2 packed as uint8, [N, K//2].",
+        ),
+        "b_descale": Tensor(
+            ["32", "4", "N_div_128", "4", "K_sf_div_4", "num_groups"],
```
dhiraj113, line 897: Meant to hold when `N % 128 != 0` or `K_sf % 4 != 0`? Both tested shapes are multiples.

`_unswizzle_batched_sf_128x4` pads and slices, so the arithmetic probably survives, but `convert_sf_to_mma_layout` would return the padded extents and `N_div_128`/`K_sf_div_4` would then misdescribe the buffer in the dumped definition. If unpadded `N` is out of scope, worth asserting it in `_fp4_gemm_check`.

dhiraj113 (commented):
Two inline comments, neither blocking.

flashinfer/trace/templates/gemm.py
```diff
@@ -782,17 +870,69 @@ def _mm_bf16_fp4_cute_dsl_init(
     init=_mm_bf16_fp4_cute_dsl_init,
 )
 
+mm_bf16_fp4_cute_dsl_sm100_trace = TraceTemplate(
```
dhiraj113, line 873: New `TraceTemplate` also needs the `CLAUDE.md` checklist steps: an `example.py` call and regenerated `tests/trace/fi_trace_out/` JSON. Both sibling templates have one, so this would be the only layout with no committed definition.

No SM100 hardware needed — auto-dump runs before the wrapped call (`api_logging.py`: "Generate trace BEFORE the actual call") and dispatch keys off `b_descale.dim() == 6`, not the device, so synthetic zeros in the existing `try/except` pattern work:

```python
b_w4 = torch.zeros(N, K // 2, dtype=torch.uint8, device=device)
sf_w4 = torch.zeros(
    32, 4, N // 128, 4, (K // BSW) // 4, 1, dtype=torch.uint8, device=device
)
```

→ `mm_bf16_fp4_cute_dsl_sm100_N2048_K7168_block_size16.json`

dhiraj113 (commented):
One more inline.

dhiraj113 (commented):
_(duplicate submission — disregard.)_

IwakuraRein (approved):

bkryu (approved):

PR #4621, opened 2026-08-19
Enable top_k_varlen on Rubin (SM107) and add the DKG filtered-radix backend (`radix_filter`)

dhiraj113:
## 📌 Description

Two changes to `top_k_varlen`: make the existing backends work on Rubin (SM107), and add a fourth backend ported from DKG.

### 1. Enable `top_k_varlen` on Rubin (SM107)

`top_k_varlen` refused *every* backend on SM107 — including `radix_cutlass`, the plain CUTLASS fallback that has nothing Blackwell-specific about it. The compute-capability sets attached to the backend checkers simply had no `107` entry, so dispatch raised `BackendSupportedError` before any kernel ran.

No kernel changes were needed. The CuTe-DSL kernels use only family-portable operations, and the DSL already knows `sm_107a`, so they compile for Rubin as-is. This adds `107` to the capability lists.

It also adds a small probe distinguishing two questions that used to be conflated: *"does FlashInfer ship a kernel for this tier?"* (the capability lists) versus *"can the installed CuTe DSL emit code for this device?"*. They differ on new silicon — a DSL predating Rubin fails on `sm_107a` deep inside `cute.compile`. Consulting the second makes `backend="auto"` fall back cleanly instead of crashing, and makes `is_backend_supported` tell the truth.

Result on SM107: **5 passed / 43 failed / 67 skipped → 115 passed.**

### 2. Add the DKG filtered-radix kernel as a `radix_filter` backend

Vendors the filtered top-k kernel from the NVIDIA dynamic-kernel-generator repo (MR !25590, *"[CuTeDSL] Adapt radix top-k to Rubin arch"*).

All three CuTe-DSL backends solve the same problem — find the k-th largest value, then keep everything above it — and all three do it by histogramming the high bits rather than sorting. They differ in **what they hold in shared memory while narrowing**:

- **`radix`** stages a chunk of the *raw row* in SMEM; all CTAs merge histograms each round to agree on a global pivot.
- **`radix_filter`** (new) stages only the *survivors*. It streams the row twice — once to histogram, once to collect the elements landing in the threshold bin — then refines within that candidate set on-chip, never touching DRAM again.
- **`gvr`** skips the search when it can: it guesses a threshold from the previous decode step's indices, verifies, and refines. It needs a `pre_idx` hint; the other two don't.

So `radix_filter` is a sibling of `radix` — hint-free, same input contract, directly interchangeable — rather than an alternative to `gvr`.

The practical consequence of staging survivors instead of input is that `radix_filter`'s SMEM pressure scales with *how many elements fall in the threshold bin* rather than with a chunk size it can shrink at will. That is why it alone carries overflow policies, and why it is the backend most sensitive to a device's SMEM capacity.

Two notes on the port:

- The vendored files are byte-faithful to upstream apart from the provenance header, removal of DKG-internal release markers, and import rewrites, so upstream fixes can be re-applied by re-vendoring. They are excluded from mypy in the same way as the other vendored CuTe-DSL kernels.
- Upstream's wrapper has no `compress_ratio` and no `pre_idx` parameter, so the checker declares those unsupported rather than silently ignoring them — `backend="auto"` then falls through to a backend that implements them.

- The backend checker also probes for the `cutlass.memory` namespace the vendored kernel requires; that API postdates the repo's minimum nvidia-cutlass-dsl pin, and without the probe the kernel dies with an `AttributeError` inside `cute.compile` on the pinned floor. The probe fails closed, so older DSLs get a clean unsupported verdict and `backend="auto"` is unaffected.

## 🔍 Related Issues

DKG merge request !25590 (`dlarch-fastkernels/dynamic-kernel-generator`).

### ✅ Pre-commit Checks

## 🧪 Tests

Every suite exercising the `top_k_varlen` API passes on **both** SM100 and SM107, with identical counts: `tests/topk_varlen` 115 passed, its `-k gvr` subset 11 passed, `tests/trace/test_top_k_varlen_reference_correctness.py` 2 passed, and the trace-template suites (consistency 714, registry 4, init 373) passing on both.

Also checked against an independent per-row `torch.topk` reference that shares no code with the kernels — verifying index count, range, uniqueness, and that every selected value clears the true k-th largest — across all four backends, fp32/fp16/bf16, fixed and variable lengths, and N up to 262144: **0 failures on either arch**.

The SM107 enablement was additionally verified across seven GPUs (SM80, SM89, SM90, SM100 ×2, SM107, SM120) with zero failures; non-datacentre parts are unchanged.

## Reviewer Notes

Performance characterisation is deliberately left out of this description — the numbers need more validation before I'd want them relied on, and I'll follow up separately.

### Positioning vs the fused page-table path (from review)

Review measurement (B200, fp32, k=2048, page_size=1) compared `radix_filter` via `top_k_varlen` plus a gather-based page-table transform against the existing fused `top_k_page_table_transform(dsa_graph_safe=True)`. The fused path wins most of the decode regime; `radix_filter` only pulls ahead at large N with enough rows (up to 1.77x end-to-end at 256 rows x 262K). The raw radix/GVR comparisons in this PR therefore do **not** establish an end-to-end win over the fused path for ordinary decode shapes. The backend lands here as an opt-in selection engine; the justified follow-up is wiring it in **under** the fused dispatcher for the large-N regime rather than substituting it at the adapter level.

## Summary by CodeRabbit

* **New Features**
  * Added support for Rubin GPUs and expanded compatibility across Blackwell and Rubin datacentre platforms.
  * Added a filtered top-k backend for variable-length workloads, with automatic selection and validation.
  * Improved top-k performance through architecture-aware tuning, shared-memory optimization, and multi-CTA execution.
  * Added configurable overflow handling and efficient execution for larger workloads.
  * Added graceful fallback when the installed CuTe DSL version is unsupported.
* **Documentation**
  * Updated backend documentation with Rubin support and filtered top-k usage.

### DIFF
--- a/flashinfer/topk_varlen/kernels/filtered_topk_decode.py
+++ b/flashinfer/topk_varlen/kernels/filtered_topk_decode.py
@@ -1,0 +1,1921 @@
+# Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
+# SPDX-License-Identifier: BSD-3-Clause
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this
+# list of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+# this list of conditions and the following disclaimer in the documentation
+# and/or other materials provided with the distribution.
+#
+# 3. Neither the name of the copyright holder nor the names of its
+# contributors may be used to endorse or promote products derived from
+# this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
+# Vendored from the NVIDIA dynamic-kernel-generator (DKG) repository,
+# cutlass_ir/compiler/python/examples/CuTeDSL/cute/blackwell/kernel/top_k/filtered_top_k_decode_varlen.py
+# at DKG master b45e50a7336 (merge request !25590, "[CuTeDSL] Adapt radix top-k to
+# Rubin arch").
+#
+# Changes from upstream are limited to: this header, removal of DKG-internal
+# release markers, and import rewrites to flashinfer-relative paths. The kernel
+# algorithm is unmodified so upstream fixes can be re-applied by re-vendoring.
+
+import math
+
+import cuda.bindings.driver as cuda
+import cutlass
+import cutlass.cute as cute
+import torch
+from cutlass.cute.nvgpu import cpasync
+from cutlass import testing
+from cutlass.torch import dtype as torch_dtype
+
+from .filtered_topk_util import (
+    _RUBIN_TOPK_ARCHITECTURES,
+    FilteredTopKKernelVarlen,
+    auto_tma_load,
+    auto_unroll_factor,
+    compare_top_k_results,
+    create_random_logits,
+    get_topk_architecture_config,
+    run_reference_top_k,
+    tma_tuned_default,
+)
+
+
+def _get_num_sms() -> int:
+    """Return the number of SMs on the current device (cached)."""
+    if not hasattr(_get_num_sms, "_value"):
+        _get_num_sms._value = torch.cuda.get_device_properties().multi_processor_count
+    return _get_num_sms._value
+
+
+"""
+A high-performance topk kernel example based on radix-based filter algorithm for
+the NVIDIA Blackwell SM100 architecture based on CuTe DSL.
+
+The radix-based filter top-k algorithm mainly includes two phases: coarse filter and multi-round fine-grained filter.
+For each phase:
+1. histogram: Build a histogram of the input values using vectorized loads.
+2. prefix sum: Find the threshold bin using prefix sum.
+3. find target bin id: Find the target bin id using multiple rounds.
+Finally, write the top-k values and indices to the output tensor.
+
+Supported data types:
+- Float32
+- Float16
+- BFloat16
+
+To run this example:
+.. code-block:: bash
+    python examples/blackwell/sort/filter_top_k_decode_varlen.py  \
+      --dtype Float32 --batch_size 1 --max_num_cols 4096 --next_n 3 \
+      --top_k 2048 --wrapper-mode single-cta \
+      --do_ref_check --return_val --benchmark
+
+Constraints for this example:
+* The supported top_k range is [1, 16384].
+* The input tensor has data contiguous on the n dimension (row-major).
+* The supported input data types are Float32, Float16, or BFloat16.
+"""
+
+
+class FilteredTopKKernelVarlenDecode(FilteredTopKKernelVarlen):
+    def __init__(
+        self,
+        dtype: cutlass.Numeric,
+        max_num_cols: int,
+        top_k: int,
+        next_n: int = 1,
+        num_copy_bits: int = 256,
+        return_val: bool = True,
+        large_occupancy: bool = False,
+        # for multi-cta version.
+        enable_multi_cta: bool = False,
+        chunk_size_per_cta: int = 16384,
+        num_ctas_per_row: int = 1,
+        merge_blocks: bool = False,
+        enable_dynamic_multi_cta: bool = False,
+        varlen_merge_input: bool = False,
+        overflow_policy: str = "REREAD",
+        cache_smem_values: bool = False,
+        single_pass_multi_cta: bool = False,
+        target_blocks_per_sm: int = 4,
+        architecture: str = "sm_100",
+        unroll_factor: int = 4,
+        enable_tma_load: bool = False,
+        enable_tma_load_p3: bool = False,
+        tma_num_stages: int = 4,
+        tma_num_stages_p3: int = 2,
+    ):
+        self._large_occupancy = large_occupancy
+        self._target_blocks_per_sm = target_blocks_per_sm
+        # async-TMA load (p1 + p3), hard-restricted to the large-occupancy single-CTA
+        # path (_tma_ok) -- cluster / small-occupancy always keep LDG even if forced on.
+        # The tuned auto-default lives in _prepare_one_pass_topk, which passes a concrete
+        # flag; None here defaults OFF (constructing this class directly needs an opt-in).
+        _tma_ok = large_occupancy and not single_pass_multi_cta
+        _enable_tma_load = _tma_ok and enable_tma_load
+        _enable_tma_load_p3 = _tma_ok and enable_tma_load_p3
+        _tma_num_stages = tma_num_stages
+        _tma_num_stages_p3 = tma_num_stages_p3
+        super().__init__(
+            dtype,
+            max_num_cols,
+            top_k,
+            num_copy_bits,
+            return_val,
+            enable_multi_cta,
+            chunk_size_per_cta,
+            num_ctas_per_row,
+            merge_blocks,
+            overflow_policy=overflow_policy,
+            # large_occupancy always uses 512 threads; pass it early so that
+            # _compute_smem_input_size_for_occupancy() sees the correct num_warps.
+            num_threads_override=512 if large_occupancy else 0,
+            cache_smem_values=cache_smem_values,
+            single_pass_multi_cta=single_pass_multi_cta,
+            architecture=architecture,
+            unroll_factor=unroll_factor,
+            enable_tma_load=_enable_tma_load,
+            tma_num_stages=_tma_num_stages,
+            enable_tma_load_p3=_enable_tma_load_p3,
+            tma_num_stages_p3=_tma_num_stages_p3,
+        )
+        self.next_n = next_n
+        self.enable_multi_cta = enable_multi_cta
+        self.chunk_size_per_cta = chunk_size_per_cta
+        self.merge_blocks = merge_blocks
+        self.num_ctas_per_row = num_ctas_per_row
+        self.enable_dynamic_multi_cta = enable_dynamic_multi_cta
+        self.varlen_merge_input = varlen_merge_input
+
+        if cutlass.const_expr(self.merge_blocks):
+            # Cap vec_size so tile_width (num_threads_per_cta * vec_size) <= max_num_cols,
+            # preventing OOB s_indices from _fill_oob padding.
+            _vec_cap = max(
+                1,
+                2
+                ** int(
+                    math.log2(max(self.max_num_cols // self.num_threads_per_cta, 1))
+                ),
+            )
+            self.num_copy_bits = min(self.num_copy_bits, _vec_cap * self.dtype.width)
+            self.vec_size = self.num_copy_bits // self.dtype.width
+
+        # Resolve the "auto" (0) unroll sentinel now that threads/vec_size and
+        # the per-CTA scan extent are known.
+        if self.unroll_factor == 0:
+            blocks_per_sm = self._target_blocks_per_sm if self._large_occupancy else 1
+            per_cta_cols = (
+                self.chunk_size_per_cta
+                if (self.enable_multi_cta or self.single_pass_multi_cta)
+                else self.max_num_cols
+            )
+            self.unroll_factor = auto_unroll_factor(
+                per_cta_cols,
+                self.num_threads_per_cta,
+                blocks_per_sm,
+                self.vec_size,
+                self.dtype.width // 8,
+            )
+
+    def _compute_smem_input_size(self) -> int:
+        if cutlass.const_expr(self._large_occupancy):
+            return self._compute_smem_input_size_for_occupancy(
+                target_blocks_per_sm=self._target_blocks_per_sm
+            )
+        return self._compute_smem_input_size_for_occupancy(target_blocks_per_sm=1)
+
+    @cute.kernel
+    def filtered_topk_kernel(
+        self,
+        input: cute.Tensor,
+        indices: cute.Tensor,
+        extra_buffer: cute.Tensor,
+        seqlen: cute.Tensor,
+        output_indices: cute.Tensor,
+        output_values: cute.Tensor,
+        tma_atom=None,
+        tma_tensor=None,
+    ):
+        """CuTe DSL implementation of TopK kernel based on radix-based filter algorithm."""
+        cute.arch.griddepcontrol_wait()
+
+        smem = cutlass.memory.SmemAllocator()
+        # Keep control fields separate because each has a distinct indexing scheme.
+        s_histogram_buf_layout = cute.make_ordered_layout((self.radix + 1), order=(0))
+        s_histogram = smem.allocate_tensor(
+            element_type=cutlass.Int32,
+            layout=s_histogram_buf_layout,
+            byte_alignment=128,
+        )
+        s_counter = smem.allocate_tensor(
+            element_type=cutlass.Int32,
+            layout=cute.make_ordered_layout((1), order=(0)),
+            byte_alignment=128,
+        )
+        s_threshold_bin_id = smem.allocate_tensor(
+            element_type=cutlass.Int32,
+            layout=cute.make_ordered_layout((1), order=(0)),
+            byte_alignment=128,
+        )
+        s_num_input = smem.allocate_tensor(
+            element_type=cutlass.Int32,
+            layout=cute.make_ordered_layout((2,), order=(0)),
+            byte_alignment=128,
+        )
+        if cutlass.const_expr(self.enable_gmem_store or self.enable_bounded_spill):
+            g_num_input = smem.allocate_tensor(
+                element_type=cutlass.Int32,
+                layout=cute.make_ordered_layout((2), order=(0)),
+                byte_alignment=128,
+            )
+        else:
+            g_num_input = None
+        s_indices = smem.allocate_tensor(
+            element_type=self.index_type,
+            layout=cute.make_ordered_layout((self.top_k,), order=(0)),
+            byte_alignment=128,
+        )
+        # Candidate buffer + async-TMA staging. Each TMA ring (p1 histogram, p3 filter)
+        # needs SMEM; instead of a dedicated buffer (which would shrink the candidate
+        # capacity S), it ALIASES a candidate region that is idle while that phase runs
+        # -> zero extra SMEM. The ring must fit its aliased region
+        # (n_stages * tile_bytes <= region bytes); that is guaranteed upstream (__init__
+        # asserts capacity, adaptively caps n_stages, and reserves the mbarriers), so no
+        # runtime check is needed here. Three cases:
+        #   p1                : whole candidate buffer is unused during the histogram
+        #                       (no candidate written yet) -> aliased below.
+        #   p3 fp32 (nbuf==2) : buffer 1 is idle during the filter (candidates go to
+        #                       buffer 0) -> alias buffer 1 (this branch).
+        #   p3 fp16/bf16 (1)  : single buffer, nothing idle -> dedicated carve (else).
+        if cutlass.const_expr(
+            self.enable_tma_load_p3 and self.num_buffer_smem_input_idx == 2
+        ):
+            # Buffer-major layout: one contiguous Int32 pool split into `n_buffers`
+            # equal, 128B-padded slots (so buffer 1 is one contiguous region the p3
+            # staging can alias). idx/val are (n_buffers, capacity) strided views into
+            # the slots, so kernel access is unchanged.
+            n_buffers = self.num_buffer_smem_input_idx
+            capacity = self.filtered_topk_smem_input_size  # candidate slots / buffer
+            idx_bytes = self.index_type.width // 8
+            val_bytes = self.ordered_type.width // 8 if self.cache_smem_values else 0
+            slot_bytes = ((capacity * (idx_bytes + val_bytes) + 127) // 128) * 128
+            int32_bytes = 4  # pool element size; byte offsets below are / int32_bytes
+
+            pool = smem.allocate_tensor(
+                element_type=cutlass.Int32,
+                layout=cute.make_ordered_layout(
+                    ((n_buffers * slot_bytes) // int32_bytes,), order=(0,)
+                ),
+                byte_alignment=128,
+            )
+            base = pool.iterator
+            # idx double-buffer: buffer b at byte offset b * slot_bytes.
+            s_input_idx = cute.make_tensor(
+                cute.recast_ptr(base, dtype=self.index_type),
+                cute.make_layout(
+                    (n_buffers, capacity), stride=(slot_bytes // idx_bytes, 1)
+                ),
+            )
+            if cutlass.const_expr(self.cache_smem_values):
+                # val double-buffer, placed right after idx inside each slot.
+                s_input_val = cute.make_tensor(
+                    cute.recast_ptr(
+                        base + (capacity * idx_bytes) // int32_bytes,
+                        dtype=self.ordered_type,
+                    ),
+                    cute.make_layout(
+                        (n_buffers, capacity), stride=(slot_bytes // val_bytes, 1)
+                    ),
+                )
+            else:
+                s_input_val = None
+            # p3 staging ring aliases buffer 1 (byte offset slot_bytes).
+            tile_cols = self.num_threads_per_cta * self.vec_size
+            s_tma_stage_p3 = cute.make_tensor(
+                cute.recast_ptr(base + slot_bytes // int32_bytes, dtype=self.dtype),
+                cute.make_ordered_layout(
+                    (self.tma_num_stages_p3, 1, tile_cols), order=(2, 1, 0)
+                ),
+            )
+            s_tma_mbar_p3 = smem.allocate_array(
+                cutlass.Int64, 2 * self.tma_num_stages_p3
+            )
+        else:
+            # fp16/bf16 p3: dedicated carve (no idle buffer to alias); or p3 off.
+            # Either way the candidate buffer keeps its plain flat (nbuf,S) layout.
+            if cutlass.const_expr(self.enable_tma_load_p3):
+                tile_cols = self.num_threads_per_cta * self.vec_size
+                s_tma_stage_p3 = smem.allocate_tensor(
+                    element_type=self.dtype,
+                    layout=cute.make_ordered_layout(
+                        (self.tma_num_stages_p3, 1, tile_cols), order=(2, 1, 0)
+                    ),
+                    byte_alignment=128,
+                )
+                s_tma_mbar_p3 = smem.allocate_array(
+                    cutlass.Int64, 2 * self.tma_num_stages_p3
+                )
+            else:
+                s_tma_stage_p3 = None
+                s_tma_mbar_p3 = None
+            if cutlass.const_expr(not self.enable_reread_always):
+                s_input_idx = smem.allocate_tensor(
+                    element_type=self.index_type,
+                    layout=cute.make_ordered_layout(
+                        (
+                            self.num_buffer_smem_input_idx,
+                            self.filtered_topk_smem_input_size,
+                        ),
+                        order=(1, 0),
+                    ),
+                    byte_alignment=128,
+                )
+            else:
+                s_input_idx = None
+            if cutlass.const_expr(
+                self.cache_smem_values and not self.enable_reread_always
+            ):
+                s_input_val = smem.allocate_tensor(
+                    element_type=self.ordered_type,
+                    layout=cute.make_ordered_layout(
+                        (
+                            self.num_buffer_smem_input_idx,
+                            self.filtered_topk_smem_input_size,
+                        ),
+                        order=(1, 0),
+                    ),
+                    byte_alignment=128,
+                )
+            else:
+                s_input_val = None
+        # p1 staging ring aliases the (idle) candidate buffer s_input_idx -- see the
+        # alias note above. Zero extra SMEM; S unchanged.
+        if cutlass.const_expr(self.enable_tma_load):
+            tile_cols = self.num_threads_per_cta * self.vec_size
+            s_tma_stage = cute.make_tensor(
+                cute.recast_ptr(s_input_idx.iterator, dtype=self.dtype),
+                cute.make_ordered_layout(
+                    (self.tma_num_stages, 1, tile_cols), order=(2, 1, 0)
+                ),
+            )
+            s_tma_mbar = smem.allocate_array(cutlass.Int64, 2 * self.tma_num_stages)
+        else:
+            s_tma_stage = None
+            s_tma_mbar = None
+        if cutlass.const_expr(self.enable_reread or self.enable_bounded_spill):
+            s_overflow_flag = smem.allocate_tensor(
+                element_type=cutlass.Int32,
+                layout=cute.make_ordered_layout((1,), order=(0,)),
+                byte_alignment=128,
+            )
+        else:
+            s_overflow_flag = None
+        s_last_remain = smem.allocate_tensor(
+            element_type=cutlass.Int32,
+            layout=cute.make_ordered_layout((1,), order=(0,)),
+            byte_alignment=128,
+        )
+        num_warps = cutlass.const_expr(
+            min(self.radix, self.num_threads_per_cta) // cutlass.Int32(32)
+        )
+        s_warp_sums = smem.allocate_tensor(
+            element_type=cute.Int32,
+            layout=cute.make_ordered_layout((num_warps,), order=(0,)),
+            byte_alignment=128,
+        )
+        # SP multi-CTA (radix-filter cluster): separate DSMEM merge target so the
+        # local s_histogram is never written in-place while peers read it.
+        # (The collection prefix-scan scratch reuses s_histogram, not this buffer.)
+        if cutlass.const_expr(self.single_pass_multi_cta):
+            s_hist_merged = smem.allocate_tensor(
+                element_type=cutlass.Int32,
+                layout=cute.make_ordered_layout((self.radix + 1), order=(0)),
+                byte_alignment=128,
+            )
+        else:
+            s_hist_merged = None
+
+        # Thread and block indexing
+        bidx, bidy, _ = cute.arch.block_idx()
+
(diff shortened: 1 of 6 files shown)


dhiraj113:
@flashinfer-bot run

dhiraj113:
@flashinfer-bot run

dhiraj113:
@flashinfer-bot stop

dhiraj113:
@flashinfer-bot run

dhiraj113:
@flashinfer-bot run

dhiraj113:
@flashinfer-bot stop

dhiraj113:
@flashinfer-bot run

flashinfer/topk_varlen/kernels/filtered_topk_util.py
```diff
@@ -0,0 +1,3366 @@
...
+        # 2-pass codegen is unchanged (there length > top_k in this branch).
+        if cutlass.const_expr(self.single_pass_multi_cta):
+            _len_nonneg = length
+            if _len_nonneg < 0:
+                _len_nonneg = cutlass.Int32(0)
+            if prologue_elems > _len_nonneg:
+                prologue_elems = _len_nonneg
+            remaining = _len_nonneg - prologue_elems
+        else:
+            remaining = length - prologue_elems
```
leejnau: Please clamp prologue_elems to max(length, 0) in every mode. Otherwise a short misaligned row can have prologue_elems > length, and the scalar prologue reads past the valid row, potentially beyond the allocation.

flashinfer/topk_varlen/kernels/filtered_topk_decode.py
```diff
@@ -0,0 +1,1921 @@
...
+    get_topk_architecture_config,
+    run_reference_top_k,
+    tma_tuned_default,
+)
+
+
+def _get_num_sms() -> int:
+    """Return the number of SMs on the current device (cached)."""
+    if not hasattr(_get_num_sms, "_value"):
+        _get_num_sms._value = torch.cuda.get_device_properties().multi_processor_count
```
leejnau: _get_num_sms caches one ambient-device SM count process-wide. Please key this cache by device and query the properties of input_values.device since a device-context guard alone does not prevent later calls on another GPU from reusing stale device state.

flashinfer/topk_varlen/kernels/filtered_topk_decode.py
```diff
@@ -0,0 +1,1921 @@
...
+        tma_tuned_default(dtype, architecture, bucketed_num_cols)
+        and (num_cols % _tma_div == 0)
+        # TMA aliases the candidate buffer; REREAD_ALWAYS allocates none, so the
+        # auto default must skip it (env-forced TMA still hits the explicit guard).
+        and overflow_policy != "REREAD_ALWAYS"
+    )
+    _enable_tma_load = _tma_ok and auto_tma_load(enable_tma_load, _tuned_tma)
+    _enable_tma_load_p3 = _tma_ok and auto_tma_load(enable_tma_load_p3, _tuned_tma)
+
+    key = (
```
leejnau, line 920: Please include the actual CuTe compile architecture/CC in this cache key and the two-pass key below. Otherwise identical shapes compiled first on SM100 and then used on SM103 can reuse an architecture-specific callable and SMEM configuration from the wrong GPU.

flashinfer/topk_varlen/kernels/filtered_topk_decode.py
```diff
@@ -0,0 +1,1921 @@
...
+        unroll_factor,
+        _enable_tma_load,
+        _enable_tma_load_p3,
+    )
+    if key not in compiled_filter_topk_dict:
+        n_rows = cute.sym_int()
+        _tma_on = _enable_tma_load or _enable_tma_load_p3
+        n_cols = cute.sym_int(divisibility=_tma_div) if _tma_on else cute.sym_int()
+        n_batch = cute.sym_int()
+        input_fake = cute.runtime.make_fake_compact_tensor(
```
leejnau: This specializes a compact row-major ABI with 32-byte base alignment, but the public checker accepts padded, strided, and offset tensors. Please either support a runtime leading stride, which is preferred for SGLang’s padded score views, or reject/materialize unsupported layouts before dispatch; otherwise calls can fail late or read rows with the wrong stride.

flashinfer/topk_varlen/kernels/filtered_topk_decode.py
```diff
@@ -0,0 +1,1921 @@
...
+                cache_smem_values=cache_smem_values,
+                architecture=architecture,
+                unroll_factor=unroll_factor,
+                # num_cols-divisibility-gated resolution from above (explicit so the
+                # kernel does not re-resolve the tuned default from bucketed_num_cols).
+                enable_tma_load=_enable_tma_load,
+                enable_tma_load_p3=_enable_tma_load_p3,
+            )
+
+        compiled_kernel = cute.compile(
```
leejnau: Please route this and the two-pass compile below through FlashInfer’s build_and_load_cute_dsl_kernel, with filtered_topk_decode.py, filtered_topk_util.py, and block_scan.py in the source key. Direct cute.compile bypasses persistent caching, cross-process locking, architecture/DSL/source invalidation, and FLASHINFER_DISABLE_JIT, forcing every SGLang worker to compile on first use.

flashinfer/topk_varlen/topk_varlen.py
```diff
@@ -66,24 +72,68 @@
...
+@functools.cache
+def _cute_dsl_supports_arch(major: int, minor: int) -> bool:
+    """Whether the installed CuTe DSL can target this compute capability.
+
+    The compute-capability lists above say which tiers FlashInfer *has* kernels
+    for; this says whether the *installed* DSL can emit code for the device. The
+    two differ on new silicon: a DSL release predating Rubin resolves ``sm_107a``
+    to a ``KeyError`` deep inside ``cute.compile``. Consulting this here makes
+    ``backend="auto"`` fall back to ``radix_cutlass`` instead, and makes
+    ``is_backend_supported`` report the truth. Mirrors
```
leejnau: is_backend_supported does not call this runtime probe; the generic helper only checks backend registration and the static compute-capability list. It can therefore return True when the installed DSL or cutlass.memory support is missing and an actual call is rejected. Consider either using a runtime availability probe or revise this claim and document the helper’s static semantics.

leejnau:
The manual SM100/SM107 validation is helpful.  Consider preserving a focused subset as checked-in regression tests for backend="radix_filter"? Some especially useful coverage would be:
 

- numerical parity across dtypes, variable lengths, next_n, returned values, and preallocated outputs;
- boundary and checker behavior for top_k, pre_idx, and unsupported DSL versions
- short misaligned rows and the documented padded/non-contiguous layout behavior
- non-current-device handling and CUDA graph replay after warm-up
 
Architecture-specific cases could be guarded to skip cleanly elsewhere. If available, even one SM100 or SM107 CI smoke configuration would help ensure the new backend is exercised rather than only compiled.

flashinfer/topk_varlen/kernels/filtered_topk_util.py
```diff
@@ -0,0 +1,3366 @@
...
+            if self.enable_tma_load_p3 and self.num_buffer_smem_input_idx == 2:
+                _p3_fit = (_S * _slot) // _tile_bytes
+                self.tma_num_stages_p3 = min(self.tma_num_stages_p3, _p3_fit)
+                if self.tma_num_stages_p3 < 1:
+                    self.enable_tma_load_p3 = False
+
+        _needs_extra = self.max_num_cols > self.filtered_topk_smem_input_size
+        if (
+            overflow_policy == "TRUNCATE"
+            and self.top_k > self.filtered_topk_smem_input_size
```
leejnau: Unless there is an explicit exact-total path, TRUNCATE appears to require top_k < filtered_topk_smem_input_size, not <=. When they are equal, truncation can leave exactly K candidates, while the fine-threshold search requires the cumulative count to exceed K.  No threshold is then selected and refinement can use stale control state.

flashinfer/topk_varlen/kernels/filtered_topk_decode.py
```diff
@@ -0,0 +1,1921 @@
...
+            output_values_fake = cute.runtime.make_fake_compact_tensor(
+                dtype,
+                (n_rows, top_k),
+                stride_order=(1, 0),
+            )
+        else:
+            output_values_fake = None
+        filtered_topk_func_second = FilteredTopKKernelVarlenDecode(
+            dtype,
+            num_ctas_per_row * top_k,  # num_cols
```
leejnau, line 1322: This fixed merge width includes stage-one padding. Valid -inf values and padded (-1, -inf) entries have identical radix keys, so the merge can select -1 even when the original row contains at least K valid elements. Please carry valid candidate counts or mask negative indices during the merge.

flashinfer/topk_varlen/kernels/filtered_topk_util.py
```diff
@@ -0,0 +1,3366 @@
...
+            val_one = cutlass.Int32(1)
+
+            _ik_hist = self._iket_begin("p1_histogram")
+            _ik_h1_clear = self._iket_begin("h1_clear")
+            # Stage 1: Coarse histogram.
+            # Use a strided loop so every bin is cleared even when
+            # num_threads_per_cta < radix (e.g. 128 < 256).
+            for _hi in range(tidx, self.radix + 1, self.num_threads_per_cta):
+                s_histogram[_hi] = 0
+            if cutlass.const_expr(self.enable_reread):
```
leejnau: s_overflow_flag is allocated and consumed for BOUNDED_SPILL, but it is cleared only for REREAD. Please initialize it when enable_reread or enable_bounded_spill; otherwise bounded spill reads stale shared memory and can follow the non-overflow path with a candidate count larger than the bounded buffer.

flashinfer/topk_varlen/topk_varlen.py
```diff
@@ -956,6 +1010,108 @@ def _run_radix(
...
+    The vendored files target the ``cutlass.memory`` namespace (``SmemAllocator``,
+    ``get_smem_capacity_in_bytes``), which exists only in DSL releases newer than
+    this repo's minimum pin -- on the pinned floor the kernel raises
+    ``AttributeError`` from inside ``cute.compile``. Unlike the arch probe above,
+    this one fails CLOSED: without the module the kernel definitely cannot run,
+    so reporting unsupported (clean fallback / clean dispatch error) is strictly
+    better than the deferred crash.
+    """
+    try:
+        import cutlass.memory  # noqa: F401, PLC0415
```
leejnau, line 1053: Even if SGLang upgrades to nvidia-cutlass-dsl==4.7.1, this backend remains gated off: 4.7.1 exposes these APIs under cutlass.utils, not cutlass.memory, so this probe still returns false. It also lacks the sm_107 architecture/capacity metadata, and the TMA path uses the newer cutlass.block namespace. Could we either add and test a 4.7 compatibility path for these APIs and Rubin sizing, or explicitly document CUTLASS DSL 4.8+ as required for radix_filter?

flashinfer/topk_varlen/topk_varlen.py
```diff
@@ -956,6 +1010,108 @@ def _run_radix(
...
+    compress_ratio: int,
+    return_output_values: bool,
+    out_indices: torch.Tensor,
+    out_values: Optional[torch.Tensor],
+) -> Tuple[torch.Tensor, Optional[torch.Tensor]]:
+    """DKG filtered-radix (coarse histogram -> filter -> refine) decode top-k.
+
+    The upstream wrapper's contract already matches this API: it takes the
+    ``(batch * next_n, N)`` logits and the ``(batch,)`` int32 ``seq_lens``, and
+    returns row-relative indices padded with ``-1``. It allocates its own
```
leejnau: These row-relative indices are not sufficient for SGLang’s current FlashInfer DSA interface. Its PAGED/RAGGED paths fuse index transformation and support row_starts, deterministic tie handling, and dsa_graph_safe=True. Ordinary decode could add a separate transform launch, but extend/ragged cannot substitute this backend directly. Could we document this integration boundary and/or clarify how those transformation and graph-safety semantics would be provided?

flashinfer/topk_varlen/topk_varlen.py
```diff
@@ -956,6 +1010,108 @@ def _run_radix(
...
+    """
+    if not _cute_dsl_ready(logits.device):
+        return False
+    if not _radix_filter_kernel_dsl_ok():
+        return False
+    if compress_ratio != 1:
+        return False
+    if logits.dtype not in _RADIX_FILTER_DTYPES:
+        return False
+    return True
```
leejnau, line 1099: This checker does not validate the grouped next_n ABI. The kernel indexes seq_lens[row_id // next_n], so next_n == 0 divides by zero and logits.shape[0] != seq_lens.numel() * next_n can index beyond seq_lens or silently use the wrong grouping. This is particularly easy for an SGLang adapter to get wrong because SGLang supplies expanded per-row lengths. Please require next_n >= 1 and validate the exact row/group relationship before dispatch, or support per-row lengths explicitly.

flashinfer/topk_varlen/topk_varlen.py
```diff
@@ -165,7 +215,11 @@ def _top_k_varlen_heuristic(
     call this function with positional args on the skip_check=True path without
     raising TypeError.  Mirrors the pattern used by _heuristic_func_mm_fp4.
     """
-    return [b for b in ("gvr", "radix", "radix_cutlass") if b in suitable_backends]
+    return [
+        b
+        for b in ("gvr", "radix", "radix_filter", "radix_cutlass")
```
leejnau: radix_filter cannot be selected by backend="auto" here: its supported CCs are a subset of radix, and its checker is strictly more restrictive, so radix is always suitable first. If filtered is intentionally explicit-only, could we remove this dead entry and keep that behavior clear in the documentation? If automatic use is intended, this needs a shape/architecture-aware crossover rule since simply moving it ahead would regress the smaller-row and small-batch regions described in the performance results.

flashinfer/topk_varlen/topk_varlen.py
```diff
@@ -956,6 +1010,108 @@ def _run_radix(
...
+
+    idx, val = cute_dsl_radix_filter_topk_wrapper(
+        logits,
+        seq_lens,
+        top_k,
+        next_n,
+        return_val=return_output_values,
+    )
+
+    if out_indices is not None and out_indices.data_ptr() != idx.data_ptr():
```
leejnau: This branch is unconditional for radix_filter: top_k_varlen always allocates or receives out_indices, while the filtered wrapper always allocates a separate result tensor. Every public call therefore performs another output allocation and a full num_rows × top_k device copy, with the same duplication for returned values. Could we thread out_indices/out_values through the wrapper so the kernel writes directly into the caller’s buffers? That would also let SGLang reuse stable graph destinations.

flashinfer/topk_varlen/kernels/filtered_topk_decode.py
```diff
@@ -0,0 +1,1921 @@
...
+            )
+            if use_cold_l2
+            else 1
+        )
+        if print_verbose:
+            print(f"Workspace count: {workspace_count}")
+        torch_stream = torch.cuda.Stream()
+        benchmark_stream = cuda.CUstream(torch_stream.cuda_stream)
+        with torch.cuda.stream(torch_stream):
+            time = testing.benchmark(
```
leejnau, line 1828: For the SGLang use case, could we also benchmark the public path end-to-end against top_k_page_table_transform(..., dsa_graph_safe=True)? This timer invokes the prepared launcher directly, so it excludes the public wrapper’s output allocation/copy and the additional page-table transform launch that SGLang would require. The raw radix/GVR comparison therefore does not establish whether this backend wins over SGLang’s current fused graph-safe path.

flashinfer/topk_varlen/kernels/filtered_topk_util.py
dhiraj113: Fixed in b0004abe by generalizing the existing single-pass-multi-CTA clamp to every mode, so the total scanned is always max(length, 0). You were right, and it was worse than a stray read: with top_k < length < prologue span, out-of-range elements entered the coarse histogram and indices beyond the row's valid length could be emitted as results (the negative `remaining` also propagated into the tail-loop bounds). Added a regression test that plants a sentinel beyond each row's valid length across fp32/fp16/bf16 with misaligned row strides -- it fails on the unfixed kernel for all three dtypes. Marked as a deliberate divergence from the vendored snapshot; will report upstream.

flashinfer/topk_varlen/kernels/filtered_topk_decode.py
dhiraj113: Fixed in 9bbb3d1a: the cache is now a per-device-index dict, and the three consumers (the cluster-size chooser and both large-occupancy decisions) query it with the input tensor's device rather than relying on the ambient one. Added a host-only regression test that checks the cached count against every visible device's actual properties; verified on a 5-GPU heterogeneous host (SM counts 108-148) to fail before the fix and pass after.

flashinfer/topk_varlen/kernels/filtered_topk_decode.py
dhiraj113: Fixed in 6933aa59 together with the JIT-routing thread below: both in-process keys now lead with the input device's compute capability, and the on-disk artifacts are additionally arch-tagged by the persistent cache (radix_filter_topk_sm100a_cute_dsl/...), so a heterogeneous process cannot reuse an SM100 callable or SMEM config on SM103. The public wrapper also enters the input tensor's device for compile+launch so the tag and stream agree with the data's device.

flashinfer/topk_varlen/kernels/filtered_topk_decode.py
dhiraj113: Done in 6933aa59: all three compile sites (single-pass, and both stages of the two-pass path) route through build_and_load_cute_dsl_kernel under a radix_filter_topk module, with filtered_topk_decode.py, filtered_topk_util.py and block_scan.py in the source key -- so arch/DSL/source invalidation, cross-process locking, and FLASHINFER_DISABLE_JIT all apply. Measured the worker-startup effect directly: a fresh process's first radix_filter call went from a multi-second compile to a 0.04 s disk-cache load. A regression test asserts the exported artifact exists after one public-API call.

flashinfer/topk_varlen/kernels/filtered_topk_decode.py
dhiraj113: Took the runtime-leading-stride option in aad38cb1: both input fakes now declare a symbolic leading stride, so a padded score view is zero-copy and part of the declared ABI (measured first -- the runtime already honored the true stride, but only accidentally; row-0 masks the difference, rows 1+ were used to verify). The genuinely unsupported layouts -- non-unit inner stride or a base sliced off 32-byte alignment, which previously died late with an opaque FFI alignment error -- are materialized in the public wrapper before dispatch. Regression test covers both: padded-view correctness pinned per row, and the misaligned-base case end to end.

flashinfer/topk_varlen/topk_varlen.py
dhiraj113: You're right -- the helper is static (registration + CC lists; it never runs the checkers), so the docstring's claim was wrong. Took the documentation option in ad95d0f5: the claim is removed and the static semantics are stated explicitly, including that a True does not guarantee a call is accepted in this environment (installed-DSL arch support and radix_filter's cutlass.memory gate apply at dispatch through the checkers). Making the generic helper consult runtime probes would change every @backend_requirement API repo-wide, so I left that to a separate discussion.

flashinfer/topk_varlen/kernels/filtered_topk_util.py
dhiraj113: Confirmed against the fine-search predicate (the selected bin needs the inclusive cumulative count to STRICTLY exceed the remaining k), so equality leaves no selectable threshold exactly as you describe. Fixed in 5c8539ef: the constructor now rejects top_k >= filtered_topk_smem_input_size for TRUNCATE, and the docstring bound follows. For context, the public top_k_varlen path pins the policy to REREAD, so this hardens the kernel-level surface; a host-only constructor test covers both sides of the boundary.

flashinfer/topk_varlen/kernels/filtered_topk_util.py
dhiraj113: Fixed in 85edc1b6: the clear is now gated on (enable_reread or enable_bounded_spill), matching the reader's own gate. Added a kernel-level test that forces the spill-capacity overflow path under BOUNDED_SPILL against a torch reference -- being honest about its limits, the unfixed read depends on residual SMEM contents, so the test is functional coverage of the fixed path rather than a deterministic repro of the stale read. Also for context: the public top_k_varlen path pins the policy to REREAD, so this hardens the kernel-level surface.

flashinfer/topk_varlen/kernels/filtered_topk_decode.py
dhiraj113: Verified the mechanism you describe: the pad site (trivial-branch emit) writes (-1, -inf), and nothing in the flipped radix key distinguishes that from a valid -inf element, so a row whose top-k legitimately contains -inf can lose slots to padding in the merge. This is the one finding I'm documenting rather than fixing here, with reasoning I'd rather have challenged than hidden: the pad write is shared with the FINAL stage's trivial branch (changing the pad value would alter the public -inf output contract), index-masking lives in the hot scan loops shared with the one-pass mode, and carrying valid counts does not compose with interleaved per-chunk padding -- a cross-stage redesign that belongs upstream, where the two-pass harness lives. The path is not reachable from flashinfer's public top_k_varlen (one-pass only), and 7a530682 documents the limitation at the two-pass entry so nobody routes public traffic there until it is fixed upstream. Reporting it upstream together with the prologue-clamp, TRUNCATE-boundary and overflow-flag fixes from the other threads.

flashinfer/topk_varlen/topk_varlen.py
dhiraj113: Validated in ac9f6d36 -- and one level up from where the comment pointed: the grouped-row contract is shared by every backend, so the checks (next_n >= 1 and logits.shape[0] == seq_lens.numel() * next_n) live in the top_k_varlen body itself, which also still runs under skip_check=True where the per-backend checkers do not. The error message spells out the per-row-lengths alternative (next_n=1 with one seq_lens entry per row) for exactly the expanded-lengths adapter case you flagged. Regression test is message-matched across radix_filter/radix/radix_cutlass so the pre-fix behavior cannot satisfy it.

flashinfer/topk_varlen/topk_varlen.py
dhiraj113: Agreed on both points. Explicit-only is the intent for now, so 1a7facf0 removes the dead entry and documents why in the heuristic itself: after radix it can never be chosen, and ahead of radix it would regress the small-row/small-batch regions per the measured results. The auto crossover (large-N routing, arch-dependent boundary) is left as deliberate future work rather than a naive reorder. The module docstring already states the opt-in/never-auto behavior at the API surface.

flashinfer/topk_varlen/topk_varlen.py
dhiraj113: Threaded in b1d1acbf: the one-pass wrapper takes optional out_indices/out_values, validates them, and passes them straight into the kernel launch arguments -- safe because the kernel writes every slot including -1/-inf padding. The public path's allocate+copy fallback is gone (internal allocations also moved to the input tensor's device). Regression test asserts the returned tensors alias the caller's buffers and that a profiled call launches no aten::copy_ (the old path launches one), so graph users get stable destinations.

flashinfer/topk_varlen/topk_varlen.py
dhiraj113: Took the documentation option in d31266ac: the module docstring's backend table and the capability-probe docstring now state nvidia-cutlass-dsl >= 4.8 explicitly, with your three reasons recorded (cutlass.utils vs cutlass.memory in 4.7, missing sm_107 capacity metadata, and the cutlass.block TMA namespace). I ruled out the 4.7 compat path deliberately: aliasing the probe would make the checker accept environments where the kernels themselves still reference the 4.8 namespaces, and we have no 4.7 environment with those semantics to test a real back-port against -- so it would be an untested pseudo-compat layer. The probe already fails closed, so 4.7 users get a clean unsupported verdict and auto falls through.

flashinfer/topk_varlen/topk_varlen.py
dhiraj113: Documented in 2cf2b3f8 at the dispatch site: this backend is row-relative top-k selection only -- no fused page-table translation, no row_starts/row_to_batch, no deterministic ties, no dsa_graph_safe -- and a consumer of the fused interface can substitute it for ordinary decode only with an added transform launch, and not at all for the row_starts-based ragged/extend paths. On providing those semantics: the honest answer is that belongs in the fused top_k_page_table_transform dispatcher (where backend-shared transform/tie/graph-safety machinery already lives) rather than re-implemented per backend here; wiring radix_filter in as a selection engine under that dispatcher is the natural follow-up if the end-to-end numbers justify it -- which is exactly what the benchmark thread below measures.

flashinfer/topk_varlen/kernels/filtered_topk_decode.py
dhiraj113: Ran the end-to-end comparison you asked for on B200 (public APIs both sides; fp32, k=2048, page_size=1; radix_filter through top_k_varlen with threaded out-buffers plus a gather-based page-table transform launch, vs top_k_page_table_transform(dsa_graph_safe=True); ratio > 1 means radix_filter+transform is faster):

| rows | N | eager ratio | graph ratio |
|---|---|---|---|
| 16 | 8192 | 0.24 | 0.56 |
| 16 | 65536 | 0.52 | 0.71 |
| 16 | 262144 | 0.80 | 0.77 |
| 64 | 8192 | 0.26 | 0.55 |
| 64 | 65536 | 0.90 | 0.93 |
| 64 | 262144 | 1.29 | 1.16 |
| 256 | 8192 | 0.46 | 0.84 |
| 256 | 65536 | 1.19 | 1.07 |
| 256 | 262144 | **1.75** | **1.77** |

Your critique is confirmed: end-to-end, the fused graph-safe path wins most of the regime, and this backend only earns its keep at large N with enough rows (up to 1.77x at 256x262K). So the raw radix/GVR comparisons in the PR do NOT establish a win over the current fused path for ordinary decode shapes -- the justified follow-up is wiring radix_filter in as a selection engine under the fused dispatcher for the large-N regime, not substituting it at the adapter level. Adding this framing to the PR description.

dhiraj113:
Done incrementally through this review -- tests/topk_varlen/test_radix_filter.py now has focused regression coverage: short misaligned rows across fp32/fp16/bf16 (fails on the unfixed prologue), numerical parity on padded/strided views and misaligned bases, preallocated out_indices/out_values (aliasing asserted and a profiled call must launch no aten::copy_), grouped next_n ABI rejection across three backends, the TRUNCATE boundary (host-only constructor check), a BOUNDED_SPILL overflow-path correctness run, per-device SM-count caching (verified to fail unfixed on a 5-GPU heterogeneous host), and persistent JIT-cache artifacts after a public call. Two gaps stated openly: unsupported-DSL-version rejection was verified manually against a 4.6.1 environment but is not CI-testable without pinning an old DSL, and pre_idx boundary behavior is covered only implicitly (radix_filter ignores the hint; gvr's own tests cover its constraints).

dhiraj113:
@flashinfer-bot run

dhiraj113:
The internal pipeline failures on B300/GB200/GB300 were all one issue: those images resolve the DSL pin to the newest PyPI release (nvidia-cutlass-dsl 4.7.1), which predates the cutlass.memory APIs the radix_filter kernels need, so the backend's fail-closed version probe correctly rejects every call -- but the new test file only skipped by compute capability, not by DSL capability, so on supported-CC machines the tests failed instead of skipping. Fixed in c4500b4e: the skip helper now also consults the backend's dynamic DSL probe, and the two host-only tests guard their kernel-module imports the same way. Reproduced pre-fix against a genuine 4.7.1 install (same error, same failure count as GB300) and verified post-fix on 4.7.1 and 4.6.1 environments plus the full 7-GPU sweep. Note the flip side: CI will skip (not run) the radix_filter tests until its images carry a >= 4.8 DSL; they activate automatically at that point.

dhiraj113:
@flashinfer-bot run

dhiraj113:
@flashinfer-bot rerun failed

flashinfer/topk_varlen/kernels/filtered_topk_decode.py
leejnau: After making the leading stride symbolic, num_cols % _tma_div == 0 no longer proves that the TMA row stride is 16-byte aligned. For example, an fp32 padded view with width N=131072 and stride(0)=N+1 passes the public layout check and can auto-enable TMA on large-occupancy SM107, but its byte stride is invalid for cuTensorMapEncodeTiled. Could we gate TMA on input_values.stride(0) % _tma_div == 0, fall back or materialize otherwise, and give the fake leading stride matching divisibility when TMA is enabled? A padded, unaligned-stride regression test would also help.

flashinfer/topk_varlen/kernels/filtered_topk_decode.py
leejnau: Thanks for confirming and documenting this. Since the multi-pass path is not reachable through public top_k_varlen, I don’t think it needs to block this PR. Could you link the upstream issue once filed and add an xfail regression covering valid -inf versus padded candidates? That would preserve the failure case and prevent this path from quietly being treated as fully supported before the upstream fix lands. With that tracking in place, I’d be comfortable resolving this thread.

flashinfer/topk_varlen/topk_varlen.py
```diff
@@ -956,6 +1006,84 @@ def _run_radix(
...
+    returning False here makes ``backend="auto"`` fall through to a backend that
+    does implement them instead of silently ignoring the argument.
+    """
+    if not _cute_dsl_ready(logits.device):
+        return False
+    if compress_ratio != 1:
+        return False
+    if logits.dtype not in _FILTERED_DTYPES:
+        return False
+    return True
```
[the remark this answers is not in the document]
leejnau: This still looks valid at the current head. Although radix_filter is now explicit-only, so the original auto-routing example no longer applies, the checker still omits both advertised constraints. An explicit backend="radix_filter" call silently ignores a non-None pre_idx, while an unsupported top_k fails later instead of during backend validation.  Could we add pre_idx is None and 1 <= top_k <= 16384 checks here?

flashinfer/topk_varlen/topk_varlen.py
```diff
@@ -956,6 +1006,84 @@ def _run_radix(
...
```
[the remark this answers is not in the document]
leejnau: resolving as duplicate

flashinfer/topk_varlen/kernels/filtered_topk_decode.py
```diff
@@ -0,0 +1,2049 @@
...
+    # so no stale data survives); otherwise allocate on the INPUT device --
+    # not the ambient one -- so multi-GPU callers get outputs next to their
+    # data. DIVERGENCE FROM UPSTREAM (internal-only allocation), after
+    # review (flashinfer PR #4621): threading the buffers through avoids a
+    # full num_rows x top_k allocate+copy per public call and gives
+    # CUDA-graph users stable destinations.
+    def _as_out(buf, dtype_, what):
+        if buf is None:
+            return None
+        if not (buf.is_cuda and buf.dtype == dtype_):
```
leejnau: _as_out accepts a CUDA buffer on any device, but the kernel launches on input_values.device. For example, logits on cuda:1 with an output buffer on cuda:0 passes this validation and forwards a foreign-device pointer to the kernel. Could we require buf.device == input_values.device for both output buffers and add a multi-GPU rejection test?

flashinfer/topk_varlen/kernels/filtered_topk_decode.py
dhiraj113: Confirmed, and it was worse than the descriptor rejection: on SM107 the misaligned-stride TMA launch did not error at cuTensorMapEncodeTiled -- it returned silently wrong top-k results (caught by a torch.topk cross-check on the unfixed path). Fixed in cf562add exactly as you suggested: the tuned auto default now also requires stride(0) % (16 // elem_size) == 0 and falls back to LDG; explicitly forced TMA with a misaligned stride raises an actionable ValueError before compiling; and when TMA is enabled the fake tensor declares the leading stride's 16-byte divisibility so any residual mismatch fails at argument check. Two regression tests: a fail-fast one that runs on any radix_filter arch (forced TMA, proven to fail pre-fix), and an auto-path correctness test on the arch where the tuned default fires, with the padded unaligned-stride view you described (proven to return wrong results pre-fix on SM107).

flashinfer/topk_varlen/topk_varlen.py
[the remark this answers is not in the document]
dhiraj113: Both added in db364117: the checker now requires pre_idx is None and 1 <= top_k <= 16384 (the vendored constructor's documented bound), so an explicit radix_filter call fails at backend validation instead of silently dropping the hint or erroring later in the constructor. The regression test is message-matched to the backend-validation error, so neither pre-fix behavior (silent success for pre_idx, the deeper constructor error for top_k) can satisfy it.

flashinfer/topk_varlen/kernels/filtered_topk_decode.py
dhiraj113: Fixed in 279f33fa: _as_out now requires both output buffers to be on input_values.device, and the output validation/allocation moved ahead of the compile block so a rejected call fails fast without compiling anything. Multi-GPU rejection test added and run on a heterogeneous 5-GPU host (input on the SM100 part, buffer on another device). I deliberately did not run the unfixed path as a teeth check -- it performs a cross-device write that can poison the CUDA context -- so the test is functional coverage of the rejection.

Writing that test flushed out a second, deeper issue you'll want to know about: cute.compile resolves its target from the process-ambient device rather than the input's device, so importing cutlass while cuda:0 is current and then calling radix_filter with input on cuda:1 compiled PTX for the wrong architecture (reproduced: import on L40S, input on B100, ptxas -arch=sm_89) -- and where the wrong arch happens to compile, the artifact would be persisted under the correct device's module directory. Fixed in 723b1133 by passing an explicit --gpu-arch derived from the input device at all three compile sites; the same repro now compiles sm_100a and returns correct results.

flashinfer/topk_varlen/kernels/filtered_topk_decode.py
dhiraj113: Added in 52431988. Constructing it turned up a nuance worth recording: with a fully-valid row every chunk contributes k real candidates, so the merge width carries no padding at all and nothing can leak -- the case needs a short final chunk. With the second chunk holding a single valid element (row length 16385, k=1024, 8 finite values), roughly 1087 of 2048 output slots come back -1 on SM100. I marked it xfail with strict=False rather than strict: the leak rides on tie-collection order, and back-to-back runs of the same input measurably flip between leaking and not leaking, so a strict xfail would flake CI on architectures whose collect order differs. The test tries three seeds per run to keep the manifestation probability high. Upstream issue still to be filed -- I'll link it here once it exists.

dhiraj113:
@flashinfer-bot run

dhiraj113:
@flashinfer-bot rerun failed

dhiraj113:
@flashinfer-bot run

flashinfer/topk_varlen/kernels/filtered_topk_decode.py
dhiraj113: Upstream issue filed: https://gitlab-master.nvidia.com/dlarch-fastkernels/dynamic-kernel-generator/-/issues/56 (internal). It carries the full repro (row length 16385, k=1024, 8 finite values -> ~1087/2048 slots return -1 on SM100), the observation that the leak rides on tie-collection order and varies run to run, the invariant showing stage one always supplies >= k real candidates when the row has >= k valid elements, and both cross-stage fix directions. Together with the xfail regression in 52431988 that should complete the tracking you asked for.

flashinfer/topk_varlen/kernels/filtered_topk_util.py
dhiraj113: Upstream issue filed for this one: https://gitlab-master.nvidia.com/dlarch-fastkernels/dynamic-kernel-generator/-/issues/53 (internal). The other vendored-kernel findings from this review are reported alongside it: the TRUNCATE boundary (issue 54), the BOUNDED_SPILL overflow-flag init (issue 55), and the multi-GPU SM-count/compile-key/compile-target set (issue 57). Each links back to the corresponding fix commit and regression test in this PR.

leejnau (approved):

kahyunnam (approved):

Issue #3178, opened 2026-04-25
[Feat] FMHA prefill kernels support quantized output


carlyou:
[Tracking] In-kernel output quantization for prefill attention (CUTLASS SM100 + TRT-LLM Gen ragged + cuDNN)

## Background

Several downstream consumers (vLLM MLA `forward_mha`, see [vllm-project/vllm#35792](https://github.com/vllm-project/vllm/issues/35792)) want the prefill attention kernel to write its output directly in a quantized format, fused with the post-attention quant op. This avoids the BF16 temp + post-kernel quant (`scaled_fp8_quant`, `per_token_group_fp8_quant`, `scaled_fp4_quant`) that today hangs off `attention → o_proj` paths.

Three quantization flavors are in scope:

| Flavor | Output layout | Scale layout |
|---|---|---|
| **Static per-tensor FP8** | dense `e4m3` | single host scalar `o_scale` |
| **Dynamic per-group FP8** | dense `e4m3` | per-128 / per-64-element group scale tensor along the head_dim axis |
| **NVFP4** | uint8-packed `e2m1` | per-16-element block scale tensor (e4m3) + host scalar `o_sf_scale` |

Three prefill kernel families are in scope:

- **CUTLASS Blackwell FMHA** (`backend="cutlass"` in `BatchPrefillWithRaggedKVCacheWrapper`, dispatched into `csrc/fmha_cutlass_sm100.cu` → `include/flashinfer/attention/blackwell/`).
- **TRT-LLM Gen FMHA ragged DeepSeek** (`trtllm_ragged_attention_deepseek` in `flashinfer/prefill.py`, dispatched into `trtllm_ragged_attention_launcher` in `csrc/trtllm_fmha_kernel_launcher.cu`, kernels from the `TLLM_GEN_FMHA_METAINFO_HASH` artifact bundle).
- **cuDNN SDPA** (`backend="cudnn"` in `BatchPrefillWithRaggedKVCacheWrapper`, dispatched into  `cudnn_batch_prefill_with_kv_cache` in `flashinfer/cudnn/prefill.py`, which builds a cuDNN SDPA graph at runtime).   

## Status Details
**Static per-tensor FP8**
- **cuTLASS**: 🟡 PR: #3177 — scaled output is already supported in the mainloop collective `Sm100FmhaFwdMainloopTmaWarpspecialized`. This PR surfaces the param with added type handling
[...]

PR #4624, opened 2026-08-19
feat(cudnn): expose FP8 output scaling (o_scale) and amax outputs in cudnn prefill

### LINKED ISSUE #4224: [Feature]: Expose per-tensor amax outputs on the cuDNN FP8 prefill path
### Before submitting

- [x] I have searched existing issues and this request has not been filed yet.

### Problem

cuDNN's `sdpa_fp8` takes six scale tensors — `descale_q/k/v`, `scale_s`, `descale_s`, `scale_o` — and returns `amax_s` / `amax_o` ([docs](https://docs.nvidia.com/deeplearning/cudnn/latest/operations/Attention.html#id13)), which state: "Since scaling and descaling are critical for convergence with FP8 datatype, you are required to pass scaling and descaling input tensors, as well as amax output tensors."

`cudnn_batch_prefill_with_kv_cache` already builds the graph with all six and [passes `scale_o` into `sdpa_fp8`](https://github.com/flashinfer-ai/flashinfer/blob/6812ddb4e5beaa02157cc53855f637ca6d784827/flashinfer/cudnn/prefill.py#L500-L509). But the public API exposes only `q_scale` / `k_scale` / `v_scale` (the three descales), and at execute time `scale_s`, `descale_s`, and `scale_o` are all [bound to a shared constant `1.0`](https://github.com/flashinfer-ai/flashinfer/blob/6812ddb4e5beaa02157cc53855f637ca6d784827/flashinfer/cudnn/prefill.py#L668-L673).
`amax_s` / `amax_o` are additionally [marked `set_output(False)`](https://github.com/flashinfer-ai/flashinfer/blob/6812ddb4e5beaa02157cc53855f637ca6d784827/flashinfer/cudnn/prefill.py#L528-L533), so the calibration data cuDNN already computes is discarded.

Two consequences:

1. **No usable FP8 output.** FP8 output is quantized with scale `1.0`: magnitudes above the `e4m3` max saturate while small values lose most of their mantissa. This impact final output quality. Moreover is a silent error.
2. **No way to calibrate externally.** With `amax_o` suppressed, a caller cannot measure the output range to choose a scale outside the kernel.

This is FlashInfer-side wiring, not a cuDNN limitation: [#3178](https://github.com/flashinfer-ai/flashinfer/issues/3178) already classifies the cuDNN row for static per-tensor FP8 as "wiring needed", in contrast to its per-group and NVFP4 rows, which it marks "blocked on cuDNN API support". FlashInfer exposes `o_scale` on other backends already (`trtllm_batch_decode_with_kv_cache` takes `o_scale` and `o_sf_scale`), so this is a per-backend inconsistency.

### Requested outcome

- An `o_scale` argument on `cudnn_batch_prefill_with_kv_cache` and on `BatchPrefillWith{Paged,Ragged}KVCacheWrapper.run`, symmetric with the existing `q_scale` / `k_scale` / `v_scale`, applied when the requested output dtype is FP8. Defaulting to `1.0` preserves today's behavior exactly.
- Optional `amax_s` / `amax_o` output tensors so callers can drive delayed scaling.
- Until `o_scale` lands, reject or warn on an FP8 `o_data_type` rather than silently emitting unit-scaled FP8.

### Target hardware

SM90 (H100, H200, H20), SM100 (B200, GB200)

### Inference engine

SGLang, Custom / other, vLLM

### Affected model or model family

_No response_

### Workload and configuration

- Data type and quantization:
- Batch size or request concurrency:
- Sequence lengths or token counts:
- Parallelism: TP / EP / DP / PP / disaggregated
- Relevant shapes: heads, head dimension, hidden size, experts, top-k, page size, or other
- Environment: output of `python -m flashinfer.collect_env`

### Current workaround

_No response_

### Impact

_No response_

### Acceptance criteria

_No response_

### Related work, dependencies, or suggested scope

_No response_

### Timing or release need

_No response_

vedaanta:
## 📌 Description

Wires up the FP8 output-scaling support that the cuDNN `sdpa_fp8` graph already provides but the FlashInfer Python layer left disconnected: `scale_o` was bound to a constant 1.0 and the amax outputs were built with `set_output(False)`. As a result, requesting an FP8 `o_data_type` silently emitted unit-scaled FP8 that saturates at the e4m3 max (448) with no error and no calibration signal.

Changes in `flashinfer/cudnn/prefill.py`:
- New keyword-only args on `cudnn_batch_prefill_with_kv_cache`: `o_scale` (quantization scale applied to O before FP8 conversion, `(1,1,1,1)` fp32 GPU tensor) and `amax_o` (optional pre-allocated `(1,1,1,1)` fp32 output buffer written in place with the absolute maximum of O before `o_scale` quantization, for delayed-scaling recipes — CUDA-graph friendly; return signature unchanged). The post-softmax `amax_s` is intentionally **not** exposed (no consumer needs it); the graph-internal tensor stays a non-output.
- FP8 scale binding is now driven by `q.dtype` (all six scale UIDs always bound for fp8 q, user tensor or a cached dummy 1.0) instead of `q_scale is not None`.
- Graph-cache key hardening in `_sdpa_prefill_key_fn`: `o_data_type`, the amax output flag, `k_cache`/`v_cache` dtypes, `v_cache`-derived dims, block-table shape/dtype, seq-len/offset presence flags, and buffer strides are now part of the key. Each of these is baked into the built graph, so two same-shape calls differing only in one of them must not replay the same cached graph.
- `warnings.warn` when `o_data_type` is fp8 and no `o_scale` is given (the silent-saturation case); `ValueError` when the new fp8-only args are passed with non-fp8 q (pre-existing `q/k/v_scale` misuse keeps working, with a warning, for backward compatibility).

Changes in `flashinfer/prefill.py`:
- `BatchPrefillWithRaggedKVCacheWrapper.run` already accepted `o_scale` but the cudnn branch silently dropped it — now forwarded (floats converted to the `(1,1,1,1)` fp32 tensor cuDNN expects; the cudnn call also gets `o_data_type=out.dtype` so the graph's O dtype always matches the bound buffer).
- `BatchPrefillWithPagedKVCacheWrapper.run` gains keyword-only `o_scale` with the same handling; explicit `NotImplementedError` on backends that would silently ignore it.

Test updates:
- The unconditional "cuDNN FP8 prefill has known issues on Blackwell" `pytest.xfail` is now gated on `cudnn.backend_version() < 92600`: FP8 prefill works on SM100 with cuDNN ≥ 9.26, so the 288 previously-xfailed cases actually run (and pass) there.
- New tests: dequantized-output match vs a bf16-output reference across paged/non-paged × causal × `o_scale` ∈ {4.0, 1/16}; a saturation test proving unit scale clamps a 512-magnitude output to 448 while `o_scale=1/16` recovers it exactly with `amax_o = 512`; validation-error coverage; end-to-end wrapper coverage in new `tests/attention/test_cudnn_wrapper_oscale.py` (18 cases, both wrappers, float and tensor `o_scale` forms, cross-checked against an fa2 bf16 reference).

## 🔍 Related Issues

Closes #4224. Related context: #3178 (classifies the cuDNN backend as "wiring needed" for FP8 output scale).

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

(Hooks were run pinned to the repo config via `uvx pre-commit run --files <changed files>` — all hooks pass.)

## 🧪 Tests

Verified on SM100 (B200-class, cc 10.0) with cuDNN backend 9.26 + cudnn-frontend 1.27: `tests/attention/test_cudnn_prefill.py` full file 589 passed / 0 failed (includes the 288 formerly-xfailed FP8 cases); after the amax_s removal, the fp8 selection + `tests/attention/test_cudnn_wrapper_oscale.py` re-ran at 316 passed / 0 failed. Regression runs of `test_cudnn_graph_cache_key.py`, `test_cudnn_prefill_token_indptr.py`, and `test_cudnn_prefill_paged_cu_seqlens.py` (164 passed).

## Reviewer Notes

- The amax buffer is caller-allocated and written in place rather than returned, to stay CUDA-graph capturable and keep the `(out, lse)` return signature stable.
- FlashInfer's `q_scale`/`k_scale`/`v_scale` are semantically *descales* (fed to cuDNN `descale_q/k/v`); `o_scale` is a *quantization* scale on O — the docstrings spell this out.
- The graph-cache-key additions fix latent stale-graph-replay hazards that predate this PR (e.g. same-shape calls differing only in `o_data_type`); they are included here because the new options widen the space of same-shape-different-graph calls.

AI-assisted (Claude Code), reviewed and tested end-to-end on hardware.

## Summary by CodeRabbit

* **New Features**
  * Added FP8 output scaling support for paged and ragged prefill workflows.
  * Added optional output amax reporting for FP8 cuDNN execution.
  * Output scales can now be provided as numeric values or tensors.
  * Improved support for paged and non-paged sequence-length formats.

* **Bug Fixes**
  * Prevented incompatible cached execution graphs from being reused.
  * Added validation and clearer warnings for unsupported FP8 configurations and backends.
  * Improved output dtype handling and reduced FP8 saturation risks.

### DIFF
--- a/flashinfer/cudnn/prefill.py
+++ b/flashinfer/cudnn/prefill.py
@@ -1,4 +1,5 @@
 import functools
+import warnings
 from enum import Enum
 from typing import Optional
 
@@ -69,6 +70,75 @@ def _get_dummy_scale_tensor(device: torch.device):
     return t
 
 
+_FP8_DTYPES = (torch.float8_e4m3fn, torch.float8_e5m2)
+
+
+def _check_fp8_scale_args(
+    q_dtype: torch.dtype,
+    o_data_type: torch.dtype,
+    q_scale: Optional[torch.Tensor],
+    k_scale: Optional[torch.Tensor],
+    v_scale: Optional[torch.Tensor],
+    o_scale: Optional[torch.Tensor],
+    amax_s: Optional[torch.Tensor],
+    amax_o: Optional[torch.Tensor],
+) -> None:
+    """Validate the fp8-only scale/amax arguments against the query dtype.
+
+    With a non-fp8 query, raises ValueError for an fp8 ``o_data_type`` (a
+    non-fp8 graph cannot quantize its output: it produces NaN garbage) and for
+    the fp8-output-only arguments ``o_scale``/``amax_s``/``amax_o``.  The
+    long-standing ``q_scale``/``k_scale``/``v_scale`` arguments only warn:
+    they were always silently ignored on non-fp8 graphs and callers pass them
+    unconditionally.  With an fp8 query, warns when an fp8 output is requested
+    without ``o_scale``: the output is then emitted with unit scale and
+    saturates at the fp8 dtype maximum (448 for e4m3).
+    """
+    if q_dtype not in _FP8_DTYPES:
+        if o_data_type in _FP8_DTYPES:
+            raise ValueError(
+                f"o_data_type={o_data_type} requires an fp8 query dtype "
+                f"(torch.float8_e4m3fn or torch.float8_e5m2), got {q_dtype}: "
+                "non-fp8 graphs cannot quantize their output"
+            )
+        rejected = [
+            name
+            for name, t in (
+                ("o_scale", o_scale),
+                ("amax_s", amax_s),
+                ("amax_o", amax_o),
+            )
+            if t is not None
+        ]
+        if rejected:
+            raise ValueError(
+                f"{', '.join(rejected)} require(s) an fp8 query dtype "
+                f"(torch.float8_e4m3fn or torch.float8_e5m2), got {q_dtype}"
+            )
+        ignored = [
+            name
+            for name, t in (
+                ("q_scale", q_scale),
+                ("k_scale", k_scale),
+                ("v_scale", v_scale),
+            )
+            if t is not None
+        ]
+        if ignored:
+            warnings.warn(
+                f"{', '.join(ignored)} are only consumed by fp8 graphs and "
+                f"are ignored for query dtype {q_dtype}",
+                stacklevel=3,
+            )
+    elif o_data_type in _FP8_DTYPES and o_scale is None:
+        warnings.warn(
+            "o_data_type is fp8 but o_scale was not provided; the output is "
+            "emitted with unit scale and saturates at the fp8 dtype maximum "
+            "(448 for e4m3). Pass o_scale to quantize the output.",
+            stacklevel=3,
+        )
+
+
 def _create_cudnn_handle(stream: torch.cuda.Stream):
     global _cudnn_handle
 
@@ -137,6 +207,8 @@ def _sdpa_prefill_key_fn(
     out: Optional[torch.Tensor] = None,
     lse: Optional[torch.Tensor] = None,
     o_data_type: Optional[torch.dtype] = None,
+    return_amax_s: bool = False,
+    return_amax_o: bool = False,
 ):
     if actual_seq_lens_q is not None:
         graph_b = actual_seq_lens_q.shape[0]
@@ -150,10 +222,13 @@ def _sdpa_prefill_key_fn(
     elif q.dim() == 4:
         h_qo, d_qk = q.shape[1], q.shape[3]
 
+    # h_kv/d_vo mirror the builder, which reads them from v_cache: d_vo can
+    # differ from d_qk (e.g. DeepSeek's 192/128 split), so keying them off
+    # k_cache would let same-shape calls differing only in d_vo share a graph.
     if v_cache.dim() == 3:
-        h_kv, d_vo = k_cache.shape[1], k_cache.shape[2]
-    elif k_cache.dim() == 4:
-        h_kv, d_vo = k_cache.shape[1], k_cache.shape[3]
+        h_kv, d_vo = v_cache.shape[1], v_cache.shape[2]
+    elif v_cache.dim() == 4:
+        h_kv, d_vo = v_cache.shape[1], v_cache.shape[3]
 
     if block_tables is not None:
         page_size = k_cache.shape[2]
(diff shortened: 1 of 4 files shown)


YangXu1990uiuc (changes requested):
Reviewed commit `c1b83630cdeb48372cbfe6d475cecdb7450520a2`. Requesting changes for one reproduced buffer-validation issue.

**[P1] Validate `o_scale` and `amax_o` before binding them as FP32 device scalars.**

`_check_fp8_scale_args` validates query/output dtype combinations and argument presence, but does not check the new buffers' dtype, shape, or device. [Both buffers are then bound directly](https://github.com/flashinfer-ai/flashinfer/blob/c1b83630cdeb48372cbfe6d475cecdb7450520a2/flashinfer/cudnn/prefill.py#L850), while their graph descriptors declare FP32 scalars.

I independently reproduced both consequences on this head:

- A correctly shaped CUDA **float64** `o_scale=4.0` is accepted without an error. Its low four bytes are read as FP32 zero, so the output becomes entirely zero. The valid FP32 scale produces output absolute maximum 9.0 for the same inputs.
- A correctly shaped CUDA **float16** `amax_o` is accepted and receives a four-byte FP32 write, overwriting the adjacent half element. Using a padded allocation to keep the physical write in bounds, the neighboring sentinel changed from `-7` to `2.044921875`:

```python
backing = torch.full((16,), -7.0, dtype=torch.float16, device="cuda")
amax_o = backing[:1].reshape(1, 1, 1, 1)
# Run with otherwise-valid FP8 inputs and FP32 o_scale, then synchronize.
# backing[1] is modified even though it is outside the supplied amax_o view.
```

These are invalid caller buffers and should be rejected explicitly. Before graph build/execute, please require FP32 tensors of the documented `(1, 1, 1, 1)` shape on the same CUDA device as Q, and add negative dtype/shape/device tests. This agrees with the outstanding buffer-validation comment and adds hardware reproduction of the overwrite.

The selected FP8 tests, including both wrapper paths, passed **318 tests**, with **144 skipped and 433 deselected**. The valid FP32 paths pass; the missing negative cases are the blocker.

The PR also currently has merge conflicts. During the rebase, please adapt its wrapper tests from element-unit to current token-unit indptrs; the same integration issue is reproduced in #4626.

Validation: local SM100, PyTorch 2.13.0+cu130, cudnn-frontend 1.27.0, and a cuDNN 9.26 development build (`backend_version() == 92600`). Correctness only; no performance claims.

— Yang's codex bot

PR #4625, opened 2026-08-19
feat(decode): backend="cudnn" for BatchDecodeWithPagedKVCacheWrapper; cudnn decode honors q.dtype + base-2 LSE

vedaanta:
## 📌 Description

Two parts, rebased onto current `main`:

### 1. `backend="cudnn"` for `BatchDecodeWithPagedKVCacheWrapper` (new)

Until now cuDNN decode was reachable only through the standalone
`flashinfer.decode.cudnn_batch_decode_with_kv_cache`; the wrapper's `backend=` had no cuDNN option
(prefill got one in #5133). This adds it:

- `plan()` (`_plan_cudnn`): validates the supported envelope (fp16/bf16, one dtype for q/kv/o,
  `q_len_per_req == 1`, no RoPE / soft-cap / sliding window), stages the per-request KV lengths
  in a device buffer, and builds a dense `[batch, max_pages]` block table from the CSR
  `indptr/indices` (or uses a caller-passed `block_tables`). `max_seq_len_kv` and the table width
  are rounded up to a 1024-token bucket: cuDNN bakes both into the built graph (they are part of
  the graph-cache key), so without bucketing every decode step whose longest sequence crossed a
  page boundary would rebuild the graph (~0.5 s).
- `run()`: calls `cudnn_batch_decode_with_kv_cache` with the planned state; NHD caches are
  presented as transposed views (the graph is stride-based, no copy); `return_lse` supported;
  `sinks` / NVFP4 / skip-softmax raise `NotImplementedError`. Only device ops at run time, so it is
  CUDA-graph capturable (test included; pass a fixed `block_tables` to `plan()` in graph mode so
  the captured buffer is stable).
- Benchmark: `--backends cudnn` now goes through the wrapper; the previous direct call is
  `--backends cudnn-native` (mirrors prefill). fp16 output is no longer gated out for cuDNN.
- Docs: `docs/api/cudnn.rst`, wrapper docstring.

Measured on B200 (SM100), cuDNN 9.26.0.51 / cudnn-frontend 1.29, Qwen3-235B decode shape
(b=32, s_kv=4096, 64/4 heads, d128, page 16, CUDA graph): cudnn (wrapper) 48 µs vs fa2_tc 58 µs vs
trtllm-gen 44 µs; refcheck passes. A broader sweep (6 model shapes × batch × s_kv) shows the wrapper
path at parity with trtllm-gen for d128 GQA at b≥32 and ahead of fa2_tc; d256 shapes are slower
than fa2_tc today (cuDNN-side).

### 2. `cudnn_batch_decode_with_kv_cache`: honor `q.dtype`, add `return_lse` (original PR content)

- Q/O cuDNN data types derived from `q.dtype` (fp16/bf16) instead of hard-coded BF16 (fp16 callers
  got `cudnnGraphNotSupportedError` on 9.26).
- New keyword-only `return_lse` / `lse`: builds with `generate_stats`, returns `(out, lse)`.
  **LSE is base-2** — FlashInfer's contract (#4663 made the prefill path base-2 for the same
  reason: the cascade-merge kernels consume it). cuDNN's natural-log stats are multiplied by
  `log2(e)` after `graph.execute`. The wrapper test cross-checks the LSE against the fa2 backend.
- Graph-cache key includes q/k/v dtypes, `return_lse`, v_cache shape, block-table shape/dtype, k/v
  strides.

## 🔍 Related Issues

Decode-side counterpart of #5133 (backend="auto"/cudnn for prefill) and #4663 (base-2 LSE).

### ✅ Pre-commit Checks

- [x] `ruff format` + `ruff check` (v0.12.8, the pinned pre-commit version) clean on all touched
      files; whitespace/EOF hooks applied. (The pre-commit hook environments could not be installed
      on the test box — network — so mypy was not re-run locally; CI will.)

## 🧪 Tests

- [x] `tests/attention/test_cudnn_decode.py` (44) + `test_cudnn_graph_cache_key.py` pass on B200,
      cuDNN 9.26.0.51 / FE 1.29. New: `test_cudnn_wrapper_matches_fa2` (bf16/fp16 × HND/NHD ×
      page 16/64 × 32/8, 64/4 heads; output + LSE vs fa2), `test_cudnn_wrapper_replan_reuses_block_table_and_bucketed_max`,
      `test_cudnn_wrapper_cuda_graph_with_user_block_tables`, `test_cudnn_wrapper_rejects_unsupported`,
      `test_cudnn_wrapper_rejects_sinks_at_run`.
- [x] `benchmarks/flashinfer_benchmark.py --routine BatchDecodeWithPagedKVCacheWrapper --backends cudnn cudnn-native fa2_tc trtllm-gen --refcheck` passes.

## Reviewer Notes

- Not supported through the wrapper yet (cuDNN-side gaps measured on 9.26/9.27): `q_len_per_req > 1`
  (cuDNN builds it but lands on a ~17× slower non-split-KV engine), attention sinks at `s_q == 1`
  (rejected by cudnn-frontend), fp8 KV (NVRTC compile failure for some head configs, unsupported for
  d>128). These raise `NotImplementedError` rather than silently degrading.
- `backend="auto"` still never selects cuDNN for decode; that can follow once the d256 perf gap is
  closed on the cuDNN side.

AI-assisted (Claude Code), reviewed and tested end-to-end on hardware.

## Summary by CodeRabbit

* **New Features**
  * Added optional log-sum-exp (LSE) output for cuDNN decode operations, including caller-provided buffers.
  * Added FP16 and BF16 support for cuDNN decode queries and outputs.
  * Added cuDNN support to paged decode wrappers, including GQA, CUDA graphs, and broader GPU compatibility.

* **Bug Fixes**
  * Improved graph selection across shapes, strides, layouts, and data types.
  * Added validation and clear errors for unsupported configurations and unavailable cuDNN features.
  * Improved dtype consistency checks for query, key/value, and output tensors.
  * Improved validation for user-provided block tables.


vedaanta:
Addressed the CodeRabbit findings in the follow-up commit:

- graph-cache key now includes the seq-len tensors' shape/stride and the block-table stride (all `tensor_like`-bound)
- fallback (cubin) path rejects non-bf16 `k_cache`/`v_cache`, not just `q`
- `lse` buffer must be on `q.device`
- wrapper: `jit_args` rejected for `backend="cudnn"`, `workspace_size()` raises `NotImplementedError` for it (like cute-dsl/trtllm-gen), `run()` validates `v_cache.dtype` against the planned kv dtype
- benchmark skips fp16 `cudnn`/`cudnn-native` when cudnn-frontend is not importable
- test regex escaped

The older "add `trace=`" note is already satisfied (`@flashinfer_api(trace=cudnn_batch_decode_trace)`).

Re-verified on B200 (cuDNN 9.26.0.51 / FE 1.29): `tests/attention/test_cudnn_decode.py` + `test_cudnn_graph_cache_key.py` → 46 passed. The "Test Results Summary" check only reports "CI skipped — pending authorization"; could a maintainer trigger CI with `@flashinfer-bot run`?

YangXu1990uiuc (changes requested):
Reviewed commit `155dc58513f18c3a3798e86515cbf9ef331a454c`. Requesting changes for one reproduced correctness issue.

**[P1] Preserve noncontiguous Q when routing through the new wrapper backend.**

The new [wrapper dispatch](https://github.com/flashinfer-ai/flashinfer/blob/155dc58513f18c3a3798e86515cbf9ef331a454c/flashinfer/decode.py#L2533) forwards Q unchanged, but [the cuDNN graph descriptor](https://github.com/flashinfer-ai/flashinfer/blob/155dc58513f18c3a3798e86515cbf9ef331a454c/flashinfer/cudnn/decode.py#L166) hardcodes contiguous Q strides. The graph-cache key also omits Q strides.

For a query sliced from packed QKV with `B=4, Hq=32, Hkv=4, D=128`, Q has strides `(5120, 128, 1)`, while the graph uses batch stride 4096. Requests after the first read the wrong elements. This is a normal wrapper input pattern:

```python
packed = torch.randn(4, (32 + 2 * 4) * 128, device="cuda", dtype=dtype)
q = packed[:, :32 * 128].view(4, 32, 128)
# Compare wrapper.run(q, kv, return_lse=True) with the same call using q.contiguous().
```

I reproduced this for BF16/FP16 × HND/NHD, page size 16, maximum KV length 128. Against an independent FP32 attention reference:

- Contiguous-Q controls all passed; maximum absolute output error was 0.000281–0.002504.
- Packed-Q maximum absolute output error was 1.499–1.522; relative L2 was 1.021–1.024.
- Packed-Q LSE maximum absolute error was 0.772–0.845.

The standalone cuDNN function's contiguous-Q assumption predates this PR. The new issue is exposing that assumption through the general decode wrapper without normalizing or rejecting the input.

Please either materialize contiguous Q before dispatch, or use the actual Q strides in the descriptor and graph-cache key. Add a packed-Q wrapper regression with both `batch_size > 1` and `kv_len > 1`, so attention actually depends on Q. If supporting strides directly, exercise both cache population orders.

The existing `test_cudnn_decode.py` and `test_cudnn_graph_cache_key.py` suites passed **46 tests** on this head; they do not cover this case.

Validation: local SM100, PyTorch 2.13.0+cu130, cudnn-frontend 1.27.0, and a cuDNN 9.26 development build (`backend_version() == 92600`). Correctness only; no performance claims.

— Yang's codex bot

vedaanta:
@YangXu1990uiuc thanks — reproduced and fixed in the latest commit. The decode graph now binds Q with the caller's actual strides (batch/head strides from the tensor, `head_dim` innermost) and the graph-cache key includes `q.stride()`, so a packed-QKV slice and a contiguous query get distinct graphs in either population order; a non-unit innermost stride falls back to `.contiguous()`, and the cubin path gets a contiguous copy. Added `test_cudnn_wrapper_packed_q_strides` (bf16/fp16 × HND/NHD × packed-first / contiguous-first; `batch_size=4`, `kv_len` up to 128; packed vs contiguous vs fa2 reference). 54 tests pass on B200 (cuDNN 9.26.0.51 / FE 1.29).

Also folded in CodeRabbit's last finding: `plan()` now rejects a caller-supplied `block_tables` that cannot hold every planned sequence (`rows == batch_size`, `width >= ceil(max_kv_len / page_size)`).

vedaanta:
@flashinfer-bot run

vedaanta:
Follow-up for the last two CodeRabbit items: `cudnn_batch_decode_with_kv_cache` now checks that every tensor bound to the graph (k/v caches, workspace, lse, block tables, seq lens on the graph path) is on `q.device` and that a caller `out` has shape `(b, h, d_vo)`; the wrapper rejects a caller `block_tables` that is not on the wrapper device. 54 tests still pass on B200. GPU CI needs another `@flashinfer-bot run` from a maintainer after this push.

vedaanta:
@flashinfer-bot run

YangXu1990uiuc:
@flashinfer-bot run

YangXu1990uiuc (approved):
Re-reviewed commit `3f47f6d687da7f38ae613e55d1e352b5542fe591`. My previous packed-Q correctness blocker is resolved: the graph now uses the actual Q strides, and the cache key includes those strides. I found no remaining blocking issue in the updated change and am updating my review to approve.

Independent validation on this exact head:

- Decode and graph-cache test files: **54 passed**.
- Additional FP32-reference checks: **48 output/LSE comparisons passed**, covering BF16/FP16, HND/NHD, packed-Q batch strides, non-unit head strides, non-unit innermost strides, and both contiguous/strided call orders. Maximum absolute errors were 0.00352 for output and 1.44e-6 for base-2 LSE.
- **16 invalid-input cases** correctly rejected, covering output/LSE metadata, CPU graph inputs/workspace, and caller-owned block-table dimensions/device.

Validation used a local SM100 device, PyTorch 2.13.0+cu130, cuDNN Frontend 1.27.0, and cuDNN 9.26.0.39 development build. These are correctness results only; I made no performance measurements. GitHub's current failing test-summary annotation says `CI skipped — pending authorization`, so this approval does not claim that upstream GPU CI has passed.

— Yang's codex bot

vedaanta:
@flashinfer-bot run

flashinfer/decode.py
```diff
@@ -2361,7 +2510,52 @@ def run(
         if self._backend == "trtllm-gen":
             q = q.view(q.size(0) // q_len_per_req, q_len_per_req, q.size(1), q.size(2))
 
-        if self.use_tensor_cores:
+        if self._backend == "cudnn":
```
saltyminty, line 2513: My agent is flagging that we also need a rejection for K and V having different strides here, as the fallback assumes they have the same stride.

PR #4634, opened 2026-08-20
fix: Make GPU-direct RDMA capability configurable for local TRT-LLM IPC

shpgy-shpgy:
## 📌 Description

  This PR makes the `gpuDirectRDMACapable` allocation property configurable for symmetric device memory.

  Previously, `SymmDeviceMemory` unconditionally set `gpuDirectRDMACapable=1`. This assumption is invalid for local TRT-LLM IPC on consumer Blackwell GPUs, such as SM120, where GPUDirect RDMA may be unavailable while local CUDA IPC remains supported.

  This change:

  - Adds an optional `gpu_direct_rdma_capable` parameter to `SymmDeviceMemory`.
  - Preserves the existing behavior by defaulting the parameter to `True`.
  - Disables the GDR requirement only for the local TRT-LLM IPC workspace.
  - Keeps multicast configuration and MNNVL behavior unchanged.
  - Does not change FlashInfer backend selection or the `auto` fallback logic.
  - Keeps the API backward compatible for existing callers.

  ### Downstream vLLM Integration

  This change is used by a downstream vLLM integration that enables the existing FlashInfer TRT-LLM fused all-reduce path on consumer Blackwell GPUs.

  The vLLM integration uses the installed FlashInfer package and does not compile or patch FlashInfer. Without this allocator fix, local TRT-LLM IPC workspace initialization fails because `gpuDirectRDMACapable=1` is unsupported on these GPUs, causing vLLM to disable the FlashInfer all-reduce path and fall back to another all-reduce backend.

  ## 🔍 Related Issues

  N/A

  ## 🚀 Pull Request Checklist

  ### ✅ Pre-commit Checks

  ## 🧪 Tests

  - [x] Existing communication tests pass.
  - [x] Local TRT-LLM all-reduce workspace initialization was verified on SM120 consumer GPUs.
  - [x] Existing MNNVL/multicast paths were verified to retain their previous behavior.

  ## Reviewer Notes

  The default value preserves the original GDR allocation behavior, while the local TRT-LLM IPC path explicitly sets it to `False` because that path does not require GPUDirect RDMA.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved communication memory allocation for systems without GPU-direct RDMA support.
  * Updated all-reduce workspace allocation to use compatible settings, improving reliability across supported hardware configurations.

### DIFF
--- a/flashinfer/comm/mnnvl.py
+++ b/flashinfer/comm/mnnvl.py
@@ -738,6 +738,7 @@ class SymmDeviceMemory:
         comm_backend_for_handle_transfer: Optional[CommBackend] = None,
         enable_multicast: bool = True,
         allocate_signal_pads: bool = True,
+        gpu_direct_rdma_capable: bool = True,
     ):
         cu_device = checkCudaErrors(cuda.cuDeviceGet(device_idx))
 
@@ -765,6 +766,7 @@ class SymmDeviceMemory:
         self.allocation_size = 0
         self.comm_backend = comm_backend_for_handle_transfer or MPIBackend()
         self._enable_multicast = enable_multicast
+        self._gpu_direct_rdma_capable = gpu_direct_rdma_capable
 
         # CUDA memory handles and pointers
         self.mc_ptr = 0  # CUdeviceptr mMcPtr
@@ -1002,7 +1004,9 @@ class SymmDeviceMemory:
             cuda.CUmemLocationType.CU_MEM_LOCATION_TYPE_DEVICE
         )
         allocation_prop.location.id = self.device_idx
-        allocation_prop.allocFlags.gpuDirectRDMACapable = 1
+        allocation_prop.allocFlags.gpuDirectRDMACapable = int(
+            self._gpu_direct_rdma_capable
+        )
 
         # Get allocation granularity
         alloc_granularity = checkCudaErrors(
--- a/flashinfer/comm/trtllm_ar.py
+++ b/flashinfer/comm/trtllm_ar.py
@@ -693,6 +693,9 @@ def trtllm_create_ipc_workspace_for_all_reduce_fusion(
                 comm_backend_for_handle_transfer=comm_backend,
                 enable_multicast=False,
                 allocate_signal_pads=False,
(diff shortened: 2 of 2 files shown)


troycheng:
We tested this fix in our downstream serving stack.

We applied the change from `104d2ba` on top of FlashInfer main `4927c0e1` and
tested it with vLLM 0.25.1, TP2 on two RTX 5090 GPUs, CUDA 13.0. Before the
change, the MNNVL attempt failed because multicast was unavailable, then the
local TRT-LLM workspace failed with `CUDA_ERROR_INVALID_DEVICE`, so vLLM
disabled the FlashInfer all-reduce/norm fusion. With this change, the workspace
initialized with `backend=trtllm` and the existing fused
all-reduce + residual + RMSNorm path was selected.

In a four-lifecycle A-B-B-A service test at concurrency 16 (64 requests and
59,193 generated tokens per lifecycle), enabling the now-working fusion gave:

- output throughput: **+2.25%**
- mean TPOT: **-2.58%**
- p99 TPOT: **-3.43%**

The full decode CUDA Graph median also decreased by **3.98% at B8** and
**2.74% at B16**. The existing FlashInfer fused all-reduce/norm correctness
test passed for world size 2, BF16, B8, hidden size 2048.

The measured gain comes from making the existing fusion available; the
allocator change is not itself a new communication algorithm. Thanks for the
focused fix.

shpgy-shpgy:
Hi @aleozlx @yzh119 @jimmyzho @bkryu @nv-yunzheq  @Anerudhan — gentle ping for a review of this PR when you have a chance.

  This is a small, backward-compatible fix for local TRT-LLM IPC on consumer Blackwell/SM120: `gpuDirectRDMACapable` remains enabled by default and is set to `False` only for the local TRT-LLM workspace. MNNVL/multicast behavior and backend selection are unchanged.

  Could someone take a look? Happy to provide additional logs or run more checks. Thanks!

troycheng:
@nv-yunzheq Could you please review this pull request when you get the chance? Thanks!

PR #4649, opened 2026-08-20
fix(cute_dsl): consult the arch gate in the GEMM and GDN dispatchers

Vinnie6167:
## Problem

On SM107 (Rubin) with a CuTe DSL older than 4.8, FlashInfer fails with a bare `KeyError: 'sm_107a'` raised from `enum.py` inside `cute.compile` — no FlashInfer frame in the traceback, and no warning.

The root cause is that `supported_compute_capability` gates on **hardware capability alone**. #4122 widened the cute-dsl lists to `[100, 103, 107]`, which tells the dispatcher Rubin is supported regardless of which DSL is installed. Public CuTe DSL tops out at 4.7.0 on PyPI, and that release has no `sm_107a` in its `Arch` enum.

This is reachable without asking for cute-dsl explicitly — two `auto` heuristics route SM107 *toward* it:

- `_heuristic_func_mm_fp4`: `elif is_sm107: candidate_backends = ("cudnn", "cutlass", "cute-dsl")`
- `_heuristic_func_bmm_fp8`: appends `"cute-dsl_sm107"` when `is_sm107_supported`

## Changes

**1. Decline the cute-dsl backend when the installed DSL cannot emit for the device** (`fix(gemm)`)

The three cute-dsl requirement functions now call `_check_cute_dsl_arch(...)`, which sits beside the existing `_check_cute_dsl_availability()` and delegates to `require_cute_dsl_arch()` — the helper added in #4122, which owns both the predicate and the message (it derives the family arch and names the exact `CUTE_DSL_ARCH=sm_100f` to export).

Only the exception type is adapted, and that part is load bearing: `require_cute_dsl_arch` raises `NotImplementedError`, while `suitable_auto_backends` catches `ValueError` to mean "backend not suitable" and keeps searching. Left unadapted, an unsupported DSL would propagate out of the auto path and fail the call instead of falling back to cutlass/cudnn. Returning `False` instead of raising was also rejected: on the explicit-backend path that surfaces as `ValueError: Problem size is not supported`, which is misleading.

**No capability lists change.** This is deliberately an *availability* check, not a capability one. The kernels do exist for sm_107, so `@supported_compute_capability([100, 103, 107])` stays as-is and `is_backend_supported("cute-dsl", 107)` keeps answering `True` — it is a public method on the wrapper, called with no tensors by e.g. `flashinfer/trace/templates/gemm.py:707`, and making it vary with an installed pip package would also have made the skip reason in `tests/grouped_mm/conftest.py` environment-dependent. This mirrors how the codebase already separates the two axes: `_cudnn_mm_mxfp8_requirement` lists its capabilities statically while `CUDNN_AVAILABLE` handles presence, and `_is_cudnn_override_shape_available` handles a dependency that is present but too old.

**2. GDN CP delta rule resolves the arch instead of formatting it** (`fix(gdn)`)

`_blackwell_compile_options` guards on the major only, then builds `f"sm_{major}{minor}a"`. Rubin is 10.7, so it passes a check written when "compute 10.x" meant Blackwell 10.0/10.3. This is the only place FlashInfer names the arch for a compute-10 device; everywhere else the DSL derives it internally.

`cute_dsl_compile_arch()` returns the device's own arch when the DSL has it, the family arch when the DSL is targeting `sm_100f`, and otherwise raises `NotImplementedError` naming `CUTE_DSL_ARCH`. Same rule as the capability gate, so the two cannot disagree.

## Testing

Rubin CI, `TEST_PATH="tests/gemm tests/gdn"`, against `release-v0.6.18`, with `CUTE_DSL_ARCH=sm_100f` exported and public CuTe DSL 4.7.0:

| | before | after |
|---|---|---|
| passed | 8,141 | **12,341** |
| failed | 4,230 | **1** |
| `KeyError: 'sm_107a'` | 4,482 | **0** |

Identical results on **both** VR200 (`hecate`, 4 workers, 2,078s) and GR100 (8 workers, 3,424s); `suite_complete=true` on both, well inside the 13,500s deadline.

Per-file, verified independently on both boards:

| File | before | after |
|---|---|---|
| `tests/gdn/test_prefill_delta_rule.py` | 2,678 | **0** |
| `tests/gemm/test_mm_mxfp8.py` | 501 | **0** |
| `tests/gdn/test_decode_delta_rule.py` | 417 | **0** |
| `tests/gdn/test_prefill_cp_delta_rule.py` | 232 | **0** |

The remaining failures are `tests/gdn/test_decode_ucache.py` and `tests/gemm/test_bmm_fp8.py` — see below.

`BackendSupportedError` count is **0**, so the cute-dsl backends are being selected and compiling successfully against `sm_100f` — not silently skipped.

The node accounting reconciles exactly: the plan drops 44,275 → 44,229 nodes and 37 → 36 units, i.e. the 46 tests in the ucache module leave collection entirely (a module-level skip is taken during collection, so those nodes are not counted as `skipped`). `passed` moves +287 = +288 Triton tests now compiling, −1 ucache test that previously passed; `failed` moves −333 = −288 Triton −45 ucache.

Also unit-tested away from hardware: the new decorator resolves conditional 107 as False on DSL 4.7, True on 4.8+/`CUTE_DSL_ARCH`, False when the predicate raises, and yields a plain `set` when no conditional is given. `cute_dsl_compile_arch` was verified against a stubbed `Arch` enum for native / family / unsupported / Blackwell-unchanged, and the skip predicate for all four DSL-vs-arch combinations.

### Caveats

- **The numbers above do not reflect this branch.** They were measured at `93143db2`, which carried a skip guard for `tests/gdn/test_decode_ucache.py` that has since been reverted, so 45 of those tests now fail again rather than skipping.
- **The gemm mechanism changed after that measurement.** The two commits after it moved the check out of the decorator and into the requirement functions; that mechanism is unit-tested (adapter pass-through, `NotImplementedError` → `ValueError`, silent when the probe cannot be imported) but has not been re-run on hardware.
- The measurement runs also carry `CUTE_DSL_ARCH=sm_100f` from the CI side. With it set the DSL *can* target sm_107, so `_check_cute_dsl_arch` passes and the gemm change is a no-op; only a run without that variable exercises the deselect-and-fall-back path.
- `cute_dsl_compile_arch` changes `gdn_cp_prefill.py` for **all** compute-10 devices, not just Rubin. Blackwell resolution (`sm_100a` / `sm_103a`) is verified against a stubbed `Arch` enum, not on B200/GB200 hardware.

## Not addressed

- **45 `KeyError: 'sm_107a'`** in `tests/gdn/test_decode_ucache.py`. Not fixable from FlashInfer: those kernels compile through `@cute.experimental.jit` / `@cute.experimental.kernel`, passing no arch and no compile options, so the DSL resolves the device arch itself and looks up `sm_107a` in its own enum. There is no FlashInfer-side site to guard, the traceback bottoms out at `enum.py:813` with no FlashInfer frame, and `CUTE_DSL_ARCH=sm_100f` does not help because that path never consults it — which points at a genuine **CuTe DSL 4.8** requirement. Left visible rather than skipped; the kernel author (#4081) is better placed to say whether it is inherent.
- **1 `No valid cute-dsl SM107 bmm_fp8 config`** in `tests/gemm/test_bmm_fp8.py` — pre-existing, and present on the internal DSL 4.8 stack too (18 vs 20 occurrences across stacks), so it is independent of the DSL version question.

The 288 Triton `PTXASError` failures previously seen in `tests/gemm/test_group_gemm.py` were a CI-side issue, not a FlashInfer one: Triton resolves ptxas through its own knobs (`TRITON_PTXAS_PATH`, and `TRITON_PTXAS_BLACKWELL_PATH` for arch >= 100, which is the one Rubin selects) and otherwise falls back to `$CUDA_HOME/bin/ptxas`. Fixed in flashinfer-ci!354; this run confirms 0 remaining.

## Summary by CodeRabbit

* **New Features**
  * Added architecture detection for CuTe DSL compilation, including native and family-compatible GPU architectures.
  * Added clear guidance when the installed DSL cannot compile for a target GPU.

* **Bug Fixes**
  * Improved Blackwell architecture handling, including support for devices with nonstandard architecture identifiers.
  * Prevented unsuitable CuTe DSL backends from being selected automatically when architecture support is unavailable.


aleozlx:
## PR Review Screening

**CI verdict:** ✅ auto-run ok
**Review category:** `live` (rule fired: C3.2 durable — alters the shared `supported_compute_capability` gating abstraction with a new conditional-capability convention)
**Blocking checks:** S2 (template overwritten), S3 (unmet formally)
**Release blocker:** 🚨 candidate — fixes `KeyError: 'sm_107a'` crash inside `cute.compile` on SM107, regression from #4122 (C3.4)
**Early stop:** no

### Security
| Q | Answer | Evidence |
|---|--------|----------|
| S2 template overwritten | yes | Custom Problem/Changes/Testing format; all 5 template headers missing, 0 checkboxes |
| S3 template obligations | unmet | Unmet formally, met in substance — detailed description, before/after test tables (VR200 + GR100), but no checklist structure |

### Packaging
| Q | Answer | Evidence |
|---|--------|----------|
| C1.1 dependency bump | no | — |
| C1.2 public API changes | yes — extension | See list below |
| C1.3 AOT/trace registration | n/a | No `gen_*_module()` added; no new `@flashinfer_api` |

C1.2 signatures:
- `flashinfer/utils.py`: `supported_compute_capability(supported_ccs: Iterable[int])` → `supported_compute_capability(supported_ccs: Iterable[int], conditional_ccs: Optional[Dict[int, Callable[[], bool]]] = None)` — **extension of existing shape** (backward-compatible kwarg; `_supported_ccs` stays a plain `set` when unused)
- `flashinfer/cute_dsl/utils.py`: new `cute_dsl_compile_arch(major: int, minor: int) -> str` — **internal-only** (not exported, no `@flashinfer_api`)

### Presentation
| Q | Answer | Evidence |
|---|--------|----------|
| C2.1 perf claim backed | n/a | Fix PR; no perf claim |

### Implementation
| Q | Answer | Evidence |
|---|--------|----------|
| C3.1 experimental-track declared | no | — |
| C3.2 durable internals touched | yes — durable | `flashinfer/utils.py` `supported_compute_capability` (`@backend_requirement` gating infra): lazily-evaluated `_ConditionalComputeCapabilities` set-like, predicates re-run per membership test, raising predicate → unsupported |
| C3.3 tests match change | partial | Only `tests/gdn/test_decode_ucache.py` module-level skip in diff; the decorator/arch-resolver unit tests described in the body are not committed |
| C3.4 release-blocker candidate | yes | Library-side fix for `KeyError: 'sm_107a'` crash in `cute.compile`; regression from #4122; SM107 (Rubin, pre-release arch) only; no linked issue |

Notes for the maintainer:
- The body's "unit-tested away from hardware" claims (conditional-CC resolution, `cute_dsl_compile_arch` stub tests) have no corresponding test files in the diff — request they be committed.
- Author's own caveat: measurement runs also exported `CUTE_DSL_ARCH=sm_100f`, which makes the conditional gate a no-op in those numbers; the fix and the env var were not isolated.
- SM107 is a pre-release arch, so the blocker flag gates Rubin bring-up rather than shipped-release users; urgency is a maintainer call.

Vinnie6167:
@flashinfer-bot run

bkryu (approved):

bkryu:
@flashinfer-bot run

bkryu (approved):

bkryu (approved):

PR #4664, opened 2026-08-21
fix(attention): make PrimTS paged block-sparse metadata live

heyuhhh:
## 📌 Description

A follow-up to #4474 that makes paged PrimTS block-sparse attention consume live request metadata.

- Keep only static geometry and capacity in `BlockSparsePagedTSWrapper.plan()`.
- Pass live page indptr, page IDs, KV sequence lengths, BSR routes, and validity bits to `run()`.
- Reusable wrappers validate tensor structure on the host but trust device-side metadata values. Invalid values are unsupported and may access out of bounds; setting `CUTE_DSL_ENABLE_ASSERTIONS=1` before kernel compilation enables diagnostic device assertions.
- One-shot APIs synchronously inspect live paged-KV and BSR values before planning and execution.
- Add reusable wrapper trace schemas for contiguous, paged tuple-cache, and paged combined-cache forms, with CUDA Graph coverage.

This lets one static capacity plan be reused across decode steps and CUDA Graph replays.

### Intentional API migration

This intentionally changes the experimental paged block-sparse API introduced by #4474. The old API appeared in nightly builds but has not shipped in a stable or RC release.

- `plan()` now takes `batch_size`, `max_seq_len_kv`, and `device`; it no longer snapshots `paged_kv_indptr` or `seq_lens_kv`.
- `run()` now takes live `paged_kv_indptr`, `paged_kv_indices`, and `seq_lens_kv`.
- `block_sparse_attention_with_paged_kv_cache()` now uses `max_seq_len_kv` and requires `seq_lens_kv`.

Nightly users must move request metadata from `plan()` to each `run()` call and keep tensor addresses stable across CUDA Graph replay.

## 🔍 Related Issues

Dependency for the TensorRT-LLM PrimTS block-sparse / VisualGen VSA integration.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit`.
- [x] I have run the hooks on all changed files and fixed all reported issues.

## 🧪 Tests

- [x] All relevant focused tests are passing.

Validation:
- Focused block-sparse/trace checks after rebase: `21 passed`.
- Full trace template consistency: `750 passed`.
- Batch-aware trace cleanup: `7 passed, 743 deselected`.
- Pre-commit on changed files: passed.

## Reviewer Notes

The paged kernel ABI already accepted page metadata and KV lengths as runtime tensors. This change removes wrapper-side snapshots and always selects the dynamic-length specialization for paged storage; it does not change the FMHA math kernel.

## Summary by CodeRabbit

* **New Features**
  * Paged block-sparse attention now accepts live per-request sequence lengths and page mappings at execution time.
  * Planning uses static capacity limits, while reusable wrappers consume current runtime metadata.
  * Added reusable wrapper tracing for contiguous and paged cache formats.
  * One-shot APIs validate live metadata before execution; reusable APIs support optional assertion diagnostics.

* **Documentation**
  * Updated APIs, trace schemas, examples, and guidance to distinguish static capacities from live runtime metadata.
  * Documented requirements and limitations for CUDA Graph capture and invalid runtime values.

### DIFF
--- a/flashinfer/attention/prims_ts/_block_sparse/compiler.py
+++ b/flashinfer/attention/prims_ts/_block_sparse/compiler.py
@@ -53,7 +53,7 @@ def _compile_block_sparse(key: _BlockSparseCompileKey) -> Callable[..., object]:
         kv_route_size=key.kv_route_size,
         has_token_bits=key.use_kv_valid_bits,
         page_size=key.page_size,
-        use_variable_seqlens_kv=key.use_variable_seqlens_kv,
+        mask_type=key.mask_type,
     )
     Int32 = cutlass.Int32
     Int64 = cutlass.Int64
@@ -183,7 +183,7 @@ def _compile_block_sparse(key: _BlockSparseCompileKey) -> Callable[..., object]:
             static_config,
             static_seq_len_kv,
             seq_lens_kv.iterator,
-            key.use_variable_seqlens_kv,
+            True,
             num_physical_kv_pages,
             k_page_stride,
             v_page_stride,
--- a/flashinfer/attention/prims_ts/_block_sparse/compiler.py
+++ b/flashinfer/attention/prims_ts/_block_sparse/compiler.py
@@ -249,7 +249,7 @@ def _compile_block_sparse(key: _BlockSparseCompileKey) -> Callable[..., object]:
             valid_bits_fake,
             row_route_offsets_fake,
             route_workspace_fake,
-            Int32(-1),
+            Int32(0),
             Float32(1.0),
         )
     else:
@@ -307,7 +307,7 @@ def _compile_block_sparse(key: _BlockSparseCompileKey) -> Callable[..., object]:
             seq_lens_kv_fake,
             row_route_offsets_fake,
             route_workspace_fake,
-            Int32(-1),
+            Int32(0),
             Int64(1),
             Int64(1),
             Int64(1),
--- a/flashinfer/attention/prims_ts/block_sparse.py
+++ b/flashinfer/attention/prims_ts/block_sparse.py
@@ -36,7 +36,9 @@ import torch
 from flashinfer.api_logging import flashinfer_api
 from flashinfer.trace.templates.attention import (
     prims_ts_block_sparse_trace,
+    prims_ts_block_sparse_wrapper_trace_dispatch,
     prims_ts_paged_block_sparse_trace_dispatch,
+    prims_ts_paged_block_sparse_wrapper_trace_dispatch,
 )
 
 from ._block_sparse.config import (
@@ -218,7 +220,7 @@ class BlockSparseTSWrapper(_BlockSparseWrapperBase):
         # previously published revision intact and runnable.
         self._plan_state = candidate
 
-    @flashinfer_api
+    @flashinfer_api(trace=prims_ts_block_sparse_wrapper_trace_dispatch)
     def run(
         self,
         q: torch.Tensor,
@@ -556,7 +558,7 @@ class BlockSparsePagedTSWrapper(_BlockSparseWrapperBase):
             )
         self._plan_state = candidate
 
-    @flashinfer_api
+    @flashinfer_api(trace=prims_ts_paged_block_sparse_wrapper_trace_dispatch)
     def run(
         self,
         q: torch.Tensor,
--- a/flashinfer/attention/prims_ts/block_sparse.py
+++ b/flashinfer/attention/prims_ts/block_sparse.py
@@ -22,8 +22,10 @@ compiled adapter on the caller's CUDA stream. The one-shot entry point retains
 synchronous canonical-BSR inspection to derive its temporary plan capacity.
 
 ``BlockSparsePagedTSWrapper`` shares the same scheduling policy and decode
-kernel, but freezes request-level page spans and sequence lengths at plan time
-while accepting physical page IDs and sparse routes at run time.
+kernel, but plans only fixed Q geometry and maximum K/V capacity. Page spans,
+physical page IDs, sequence lengths, sparse routes, and token bits are all live
+run inputs. The caller owns the live sequence-length value contract; route
+preparation validates page and sparse-route structure on device.
 """
 
 import threading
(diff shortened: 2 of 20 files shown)


flashinfer/attention/prims_ts/_block_sparse/compiler.py
```diff
@@ -53,7 +53,7 @@ def _compile_block_sparse(key: _BlockSparseCompileKey) -> Callable[..., object]:
         kv_route_size=key.kv_route_size,
         has_token_bits=key.use_kv_valid_bits,
         page_size=key.page_size,
-        use_variable_seqlens_kv=key.use_variable_seqlens_kv,
```
PerkzZheng, line 56: why we removed the use_variable_seqlens_kv parameter ? does it mean we already enable variable seqlens.

flashinfer/attention/prims_ts/_block_sparse/compiler.py
```diff
@@ -183,7 +183,7 @@ def paged_tensor_adapter(
             static_config,
             static_seq_len_kv,
             seq_lens_kv.iterator,
-            key.use_variable_seqlens_kv,
+            True,
```
PerkzZheng, line 186: same here.

PerkzZheng (approved):

flashinfer/attention/prims_ts/_block_sparse/compiler.py
heyuhhh: Actually we can still pass it here but there is no need to do that, so for code clean we removed it

flashinfer/attention/prims_ts/_block_sparse/compiler.py
heyuhhh: Right, for paged kv we enable `use_variable_seqlens_kv` by default

flashinfer/attention/prims_ts/_block_sparse/compiler.py
PerkzZheng: thanks.

flashinfer/attention/prims_ts/_block_sparse/compiler.py
PerkzZheng: thanks.

saltyminty (commented):
+1 on coderabbit comment https://github.com/flashinfer-ai/flashinfer/pull/4664#pullrequestreview-5027168548 (though the lack of trace template was a pre-existing state and can be treated as non-blocking).

Otherwise LGTM

qsang-nv:
Non-blocking but needs to be addressed.
The public documentation, the PR description, and the implemented handling of out-of-range `seq_lens_kv` are not aligned.

The README and `BlockSparsePagedTSWrapper.run()` say out-of-range live lengths are unsupported and not guaranteed to fail closed. But the prepare kernel does check them and makes every affected row noncanonical:

```python
live_seq_len_is_valid = cutlass.Int32(
    raw_seq_len_kv >= cutlass.Int32(self.minimum_seq_len_kv)
    and raw_seq_len_kv <= cutlass.Int32(self.cfg.seq_len_kv)
)
```

That yields a negative row header, which `_load_prepared_sparse_row_warp()` clamps to a zero route count; the mandatory HEAD slots then resolve dummy routes that the score paths mask out. I also could not find a reachable configuration where split-KV invalidates this: `_make_block_sparse_config()` selects `split_kv_mode="disabled"`, and `FmhaDecodeConfig.validate_block_sparse_profile()` independently rejects block-sparse with split-KV reduction.

So this looks like deterministic fail-closed behavior on every currently supported paged block-sparse profile — matching the PR description's "validate live lengths on device and fail closed" — while the public docs decline to promise it. The newly added invalid-length path also has no regression coverage: the CUDA Graph fail-closed test mutates `paged_kv_indptr` and `paged_kv_indices`, but never supplies an invalid `seq_lens_kv`.

Could you clarify which contract is intended?

- If out-of-range lengths should fail the affected request closed, please document that guarantee and cover both bounds in the existing graph test (batch 0 set to `0`, `seq_len_q - 1`, and `max_seq_len_kv + 1`, reusing `assert_batch0_fails_closed()`).
- If they intentionally remain unsupported, please clarify that the range check is defense-in-depth rather than an API guarantee, and document what limitation or anticipated configuration motivates the disclaimer.

flashinfer/attention/prims_ts/block_sparse.py
```diff
@@ -552,60 +497,59 @@ def plan(
...
                 device=device,
                 device_index=device_index,
                 plan_stream=plan_stream,
-                paged_kv=paged_kv,
             )
         self._plan_state = candidate
 
-    @flashinfer_api
+    @flashinfer_api(trace=prims_ts_paged_block_sparse_wrapper_trace_dispatch)
     def run(
```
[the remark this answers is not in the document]
heyuhhh: Resolve this due to it's an intentional change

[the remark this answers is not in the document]
heyuhhh: This is an intentional migration of the experimental paged block-sparse API introduced by #4474. It appeared in nightly builds but not a stable or RC release. The PR description now documents the `plan()`, `run()`, and one-shot migrations; a compatibility shim would conflict with the new live-metadata ownership model.

heyuhhh:
> +1 on coderabbit comment [#4664 (review)](https://github.com/flashinfer-ai/flashinfer/pull/4664#pullrequestreview-5027168548) (though the lack of trace template was a pre-existing state and can be treated as non-blocking).
> 
> Otherwise LGTM

Hi @saltyminty , thanks for the figure-out! I've added the trace template in commit 95b3ccf05beb235759bfc44f7ff3b01e63a27d53

heyuhhh:
Hi @qsang-nv , thanks for your comments! I refactored this part of code in commit c23bf6a2de402c508dc93510d50e88bb26d0e810:
* For one-shot API, we inspect live paged-KV and BSR values before planning and execution
* In the wrapper plan we check the tensor safety on host and in wrapper run we check live metadatas from device side. By setting `CUTE_DSL_ENABLE_ASSERTIONS=1` before kernel compilation enables diagnostic device assertions.

saltyminty (approved):
Approved pending CI

heyuhhh:
Hi @saltyminty @qsang-nv , the CI pipeline has finished and seems there is no issues related to my changes, could you please take a look again and help merge this one, thanks!

qsang-nv (approved):

PR #4605, opened 2026-08-19
feat(kda): add CuTe DSL recurrent prefill backend

Observer007:
## Description

Add a CuTe DSL BT=16 backend for recurrent KDA prefill on B200 and B300.

- Route supported fixed and packed prefill through CuTe DSL with `backend="auto"`.
- Keep Cake as the fallback and as an explicitly selectable backend.
- Support eager implicit LPT scheduling and explicit planning for packed CUDA Graph capture.
- Support zero-length packed sequences, padded/indexed state pools, and state checkpoints.
- Derive decomp chunk ownership by binary-searching the compact cumulative chunk prefix instead of storing a dense `chunk_to_seq` tensor.
- Keep the existing public API unchanged except for the optional backend selector and planner wrapper.

## Performance

Measured on one NVIDIA B200 through the public `recurrent_kda` API with CUPTI device time, cold L2, no CUDA Graph, 20 ms warmup, and a 100 ms measurement target. Cake and CUDA 13.3 CuTe DSL use the interleaved order Cake A, CuTe A, CuTe B, Cake B; each reported value is the midpoint of the two run medians. Cake was freshly compiled from the rebased source tree. CUDA 12.9 is one complete 16-case CuTe DSL pass. Times are microseconds; speedup is Cake / CUDA-13.3 CuTe DSL.

Each backend is selected in a separate benchmark invocation:

```bash
python benchmarks/bench_recurrent_kda_prefill.py --case-set all --candidate-route dispatcher --backend cake --warmup-ms 20 --bench-ms 100
python benchmarks/bench_recurrent_kda_prefill.py --case-set all --candidate-route dispatcher --backend cute-dsl --warmup-ms 20 --bench-ms 100
```

| Case | Route | CuTe CUDA 12.9 | CuTe CUDA 13.3 | Cake | Speedup |
|---|---:|---:|---:|---:|---:|
| H96 fixed 1x8192 | engine | 486.620 | 482.555 | 509.434 | 1.06x |
| H96 mixed [1300,547,2048,963,271,3063] | engine | 358.685 | 355.501 | 385.364 | 1.08x |
| H96 packed 8x1024 | engine | 402.237 | 398.397 | 415.852 | 1.04x |
| H64 fixed 1x8192 | decomp | 425.884 | 422.868 | 471.364 | 1.11x |
| H64 mixed [1300,547,2048,963,271,3063] | engine | 254.301 | 251.342 | 292.501 | 1.16x |
[...]

PR #4667, opened 2026-08-21
fix(kda): fall back to Cake when CuTe DSL predates cutlass.experimental

kahyunnam:
## 📌 Description

The BT=16 recurrent KDA prefill kernel added in #4605 is built on `cutlass.experimental`, a namespace that only exists from CuTe DSL 4.7 onwards. `_is_cute_dsl_kda_prefill_eligible` checked only the tensor contract and the compute capability, so on SM100/SM103 with the 4.6.2 floor that `requirements.txt` still permits, an ordinary `recurrent_kda` prefill using the default `backend="auto"` dispatched into the kernel and failed with a bare `ModuleNotFoundError` — even though the caller never asked for the CuTe DSL backend.

The graceful path already existed (`test_public_prefill_auto_falls_back_to_cake` covers the ineligible case); the probe simply had no way to know that a missing `cutlass.experimental` is a reason to decline. This PR adds that probe:

- `is_cute_dsl_experimental_available()` in `flashinfer/cute_dsl/utils.py`, modelled on the neighbouring `is_rubin_cute_dsl_available()`. It probes the `cutlass.experimental` package rather than a leaf module because the whole namespace (`experimental.cuda`, `experimental.primitives`, `experimental.task_scheduling`) is the 4.6.2/4.7.0 boundary, so the same helper can gate other 4.7-only kernels.
- `_is_cute_dsl_kda_prefill_eligible` consults it, which activates the existing Cake fallback for `backend="auto"`.
- `recurrent_kda` reports the version requirement directly when `backend="cute-dsl"` is requested explicitly, instead of claiming the prefill *contract* is unsupported. That report is scoped to the compute capabilities this kernel serves, so the SM120 backend added in #4633 — which does not use `cutlass.experimental` and runs fine on 4.6.2 — keeps its own rejection message.

Reproduced and verified on a B200 (SM100) against a genuine cutlass-dsl 4.6.2 install.

Before, on `main`:

```
cutlass-dsl : 4.6.2
--- backend="auto"      ModuleNotFoundError: No module named 'cutlass.experimental'
--- backend="cute-dsl"  ModuleNotFoundError: No module named 'cutlass.experimental'
```

After:

```
cutlass-dsl : 4.6.2
--- backend="auto"      OK   shape=(1, 32, 2, 128) torch.bfloat16 finite=True
--- backend="cute-dsl"  ImportError: backend='cute-dsl' requires nvidia-cutlass-dsl>=4.7.0
                        (cutlass.experimental); backend='auto' falls back to Cake
```

On 4.7.0 both backends behave exactly as before.

## 🔬 Why only KDA: the 4.6.2 vs 4.7 audit

This fix came out of a wider audit of `main` against the 4.6.2 floor, done by diffing the extracted 4.6.2 and 4.7.0 wheels and AST-parsing every `cutlass` import in the package. Summarising it here so reviewers can see why the change is scoped to KDA:

- **45 files import 4.7-only `cutlass.experimental` modules** across 248 import sites. 44 of them are `flashinfer/attention/prims_ts/**`; the 45th is `flashinfer/kda_kernels/kda_chunked_bt16.py`, the kernel this PR guards.
- **`import flashinfer` is unaffected on 4.6.2.** No 4.7-only module is reachable through module-scope imports, which was confirmed by actually importing the package under a real 4.6.2 install.
- **PrimTS decode and MLA need no change.** Their entry points in `flashinfer/decode.py` and `flashinfer/mla/__init__.py` are already lazy `__getattr__` hooks, and the APIs are opt-in: nothing dispatches into them implicitly, so on 4.6.2 they can only fail for a caller who explicitly asked for a PrimTS kernel by name. There is no alternative backend to fall back to, so the only possible improvement there is a clearer message — worth doing, but it is a separate cosmetic change rather than a correctness fix.
- **The test suite already copes.** The five PrimTS attention tests carry `pytest.importorskip("cutlass", minversion="4.7.0")` and skip cleanly, and `tests/trace/template_registry.py` filters unimportable modules with `except ImportError`.

KDA was the only place where an *automatic* dispatch on a default code path turned a missing optional dependency into a crash, which is why it is the only behaviour changed here.

For completeness, the SM107/Rubin kernels depend on `cutlass.utils.rubin_helpers` and `tcgen05.mma.CollectorOp`, which are absent from both 4.6.2 and 4.7.0 (they arrive in 4.8). Those are already gated by `is_rubin_cute_dsl_available()` and are untouched by this PR; that existing helper is the pattern the new probe follows.

## 🔍 Related Issues

Follow-up to #4605, which introduced the CuTe DSL recurrent prefill backend.

### ✅ Pre-commit Checks

## 🧪 Tests

`tests/kda/` on a B200, after merging `main`: **577 passed, 112 skipped**, including the real CuTe DSL kernel tests (`test_cute_dsl_checkpoints_match_cake`, `test_cute_dsl_padded_indexed_state_matches_cake`). 88 of the skips are the new SM120 suite asking for a CC 12.0 device and the rest are `fla` not being installed; none are new here.

Also run against a genuine cutlass-dsl 4.6.2 install on the same B200, where the fix takes `tests/kda/` from **23 failed / 410 passed** to **9 failed / 424 passed**. The 9 remaining are pre-existing tests that drive the CuTe DSL API directly and have no version guard; they fail identically before this PR and are left alone here.

Added `test_prefill_without_cute_dsl_experimental_falls_back_to_cake`, which runs on ordinary 4.7.0 CI by simulating the older DSL through the probe. It computes a reference through the real CuTe DSL kernel, then forces the probe false and patches `_run_cute_dsl_kda_prefill` to `pytest.fail`, so it asserts the routing actually changed rather than only that the numbers match; it then checks the fallback output against the reference and that the explicit backend raises.

## Reviewer Notes

The probe lives inside `_is_cute_dsl_kda_prefill_eligible` rather than as a separate gate in `recurrent_kda`. A standalone gate reads more cleanly, but the existing routing tests monkeypatch the eligibility function to exercise dispatch on CPU tensors, and an ungated check in `kda.py` would have made those pure-Python tests require cutlass >= 4.7 to run at all.

Within that function the probe runs *after* the contract checks rather than before them. That ordering is deliberate: it confines the probe to calls that would have imported the kernel anyway, so anything rejected on tensor shape, dtype or compute capability reaches Cake by exactly the path it did before this PR.

One consequence worth naming: on an old DSL the version error takes precedence, so an explicit `backend="cute-dsl"` call on an SM100-family device that *also* violates the kernel contract is told to upgrade and only sees the contract error afterwards. Reporting both would mean evaluating the contract and the runtime separately at the dispatch site, which would make the existing CPU-only routing tests depend on cutlass >= 4.7. The upgrade is a genuine prerequisite either way, so the message is incomplete rather than wrong.

That precedence is why `_is_cute_dsl_kda_prefill_dsl_too_old` takes the compute capability into account instead of reusing the bare runtime probe at the dispatch site. `main` now tries the SM120 backend first and records its rejection reason for the shared explicit-backend error, so an unscoped version check would have replaced that reason with an irrelevant upgrade instruction on CC 12.0. `test_dsl_version_guard_is_scoped_to_the_sm100_family` pins the scoping for CC 10.0, 10.3 and 12.0 without needing any of those devices.

`_is_cute_dsl_kda_runtime_available` imports the helper lazily inside a `try`: `flashinfer/cute_dsl/utils.py` imports `cutlass` at module scope, and `kda_prefill_cute.py` deliberately keeps the DSL stack off the import path.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling when the CuTe DSL backend is unavailable.
  - Added clear installation guidance for the required CuTe DSL version.
  - Preserved contract-specific validation errors when the runtime is available.
  - Improved automatic fallback behavior for eligible recurrent KDA prefill operations.

- **Documentation**
  - Clarified CuTe DSL runtime requirements and fallback behavior in the API documentation.


kahyunnam:
@flashinfer-bot run

kahyunnam:
@flashinfer-bot  run

saltyminty (approved):

kahyunnam:
@flashinfer-bot run

kahyunnam:
@flashinfer-bot run

yyihuang (approved):

bkryu (approved):

Issue #4644, opened 2026-08-20
[Feature]:  Add NVFP4 W4A4 MLA decode for SM100/SM103


mispa-ms:
### Before submitting

- [x] I have searched existing issues and this request has not been filed yet.

### Problem

MHA/GQA accepts an NVFP4 KV cache through the public `kv_cache_sf` argument (#2702). MLA does not: every trtllm-gen MLA invocation in `flashinfer/mla/_core.py` passes `None` for `key_block_scales` and `value_block_scales` (`_core.py:3291-3292`, `4167-4168` on `main`).

vLLM's guard is per-model, not per-layer — `nvfp4 KV cache is not supported with MLA (Multi-head Latent Attention) backends` — so a hybrid model keeps its whole cache at FP8. Kimi-K3-NVFP4 has 24 MLA layers out of 93.

The MLA latent is `kv_lora_rank=512` + `qk_rope_head_dim=64`. Per token per layer: 1152 B in BF16, 576 B in FP8, 324 B in NVFP4 (288 B data + 36 B block scales).

#4568 scopes MLA prefill/decode to BF16 I/O with an FP8 KV cache, so this is not tracked there.

### Requested outcome

MLA decode on SM100/SM103 with an NVFP4 paged KV cache and NVFP4 compute, matching the BF16 and FP8 MLA decode kernels in semantics.

- `kv_cache_sf` accepted on `trtllm_batch_decode_with_kv_cache_mla` and `BatchMLAPagedAttentionWrapper`, replacing the hard-coded `None` block scales.
- NVFP4 query, with FP4 global scales folded into `bmm1_scale` / `bmm2_scale`.
- Both variants of the trailing `qk_rope_head_dim` channels: `mla_use_nope=true` (Kimi-K3) and RoPE applied (DeepSeek-style).
- An NVFP4 cache write path, rather than materializing BF16/FP8 and requantizing.

Dense MLA only; sparse MLA and DSA top-k paths are out of scope.

### Target hardware

SM100 (B200, GB200), SM103 (B300, GB300)

### Inference engine

vLLM

### Affected model or model family

nvidia/Kimi-K3-NVFP4 — https://huggingface.co/nvidia/Kimi-K3-NVFP4

### Workload and configuration

- Data type and quantization: NVFP4 KV (packed `uint8` + FP8 E4M3 block scales, `sf_vec_size=16`,
[...]

PR #4676, opened 2026-08-22
feat: add NVFP4 quantize-append path for the MLA paged KV cache

elwhyjay:
## 📌 Description

Add `nvfp4_quantize_append_paged_mla_kv_cache`, the cache-write path for 512+64 MLA latent rows, as an MLA analog of `nvfp4_quantize_append_paged_kv_cache`.

The API quantizes `ckv` directly from FP16/BF16 to packed E2M1 with one FP8 E4M3 scale per 16 channels. The `kpe` part is stored as FP8 E4M3.

Cache layout:

- `ckv_cache`: `uint8 [num_pages, page_size, 256]`
- `ckv_sf_cache`: `float8_e4m3fn [num_pages, page_size, 32]`
- `kpe_cache`: `float8_e4m3fn [num_pages, page_size, 64]`

This uses 352 bytes per token, compared with 576 bytes for an all-FP8 layout, reducing the footprint by 39%.

We keep `kpe` in FP8 because DeepSeek-V2-Lite latent measurements show approximately 6x higher RMS energy in the RoPE channels than in the compressed-KV channels. In the same experiment, the selected `ckv-FP4/kpe-FP8` layout had relative L2 error against a BF16 attention reference of 0.092-0.102, compared with 0.074-0.078 for the FP8 baseline. Quantizing all 576 channels to FP4 increased the error to 0.174-0.193 and was rejected.

The shared `nvfp4_append_quantize_block` helper gains an opt-in `precise_rounding` parameter (default off, GQA callers unchanged) so the new path is bit-reproducible under `-use_fast_math`.

This PR only adds the cache-write path. Dense MLA decode does not yet consume this 352-byte layout and now raises an explicit error for packed `uint8` caches. Decode support and slot-mapping/tensor-scale support are follow-ups.

RoPE is applied by the caller. NoPE models should pass unrotated `kpe` values.
## 🔍 Related Issues

#4644
## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

`tests/attention/test_mla_page.py` (96 tests pass on B200):

- Bit-exact comparison of packed data, scale bytes, and FP8 kpe bytes against a torch reference over ragged lengths, page sizes 1/16/64, fp16/bf16, and permuted page tables, plus untouched-slot checks, bad-scale rejection, and the dense decode fail-fast.
- Format interop: the written cache dequantizes through the shared `e2m1_and_ufp8sf_scale_to_float` utility (linear scale layout), agrees with an independent E2M1 lookup-table dequantization, and stays within the expected round-trip error of the original input.
## Reviewer Notes

The `kpe`-in-FP8 layout is the main design decision here. I am happy to share the full numerics report in the issue or adjust the layout if a different tradeoff is preferred.

## Summary by CodeRabbit

* **New Features**
  * Added NVFP4 quantization and append support for paged MLA KV caches.
  * Supports packed NVFP4 compressed-KV storage and FP8 KPE data.
  * Added a public Python API and tracing support.
  * Added validation for inputs, scales, cache layouts, and unsupported dense decoding scenarios.

* **Documentation**
  * Added the operation to the paged KV-cache API documentation.

* **Tests**
  * Added coverage for accuracy, layouts, invalid scales, page permutations, and unsupported decode configurations.


aditya4d:
@elwhyjay, are you actively working on this? If so, what is the ETA?

elwhyjay:
Hi @aditya4d, thanks for checking. Yes, this is still active. From my side the branch is ready for review.
@qsang-nv gentle ping, could you take a look when you have time?

qsang-nv:
Found two issues that need fixing before merge:

## 1. [P2] `ckv_scale`/`kpe_scale` validated in double but used as float — can silently write zero/Inf/garbage

`csrc/page.cu:536` validates the scale contract (`isfinite && > 0.0`) on the `double` parameters:

```cpp
TVM_FFI_ICHECK(std::isfinite(ckv_scale) && std::isfinite(kpe_scale) && ckv_scale > 0.0 &&
               kpe_scale > 0.0)
    << "ckv_scale and kpe_scale must be positive finite global decode scales";
```

But the kernel launch (`csrc/page.cu:590`) truncates both to `float`:

```cpp
static_cast<float>(ckv_scale), static_cast<float>(kpe_scale), stream);
```

A value that's a legitimate positive finite `double` can silently become `0.0f` or `Inf` after this cast (e.g. `1e-50` underflows to `0.0f`, `1e300` overflows to `Inf`). Downstream in `NVFP4QuantizeAppendPagedKVMlaCacheKernel`:
- `global_scale == 0.0f` makes `nvfp4_append_quantize_block`'s `amax > 0.0f && global_scale > 0.0f` guard false, so `sf_value`/`output_scale` are forced to 0 — the ckv block is silently quantized to all-zero instead of erroring.
- `kpe_inv_scale = __frcp_rn(kpe_scale)`: if `kpe_scale` underflows to 0, `kpe_inv_scale` becomes `Inf`, and multiplying by any nonzero `kpe` value produces `Inf`/`NaN` that gets silently cast into the FP8 output.

Both the Python wrapper's `math.isfinite`/`> 0.0` check and the C++ `TVM_FFI_ICHECK` operate on double precision, so neither catches this. Please re-validate after the cast to `float` (or clamp/reject in the Python layer to the representable positive-finite FP32 range) so out-of-float32-range scales fail loudly instead of silently corrupting the cache.

## 2. [P2] The largest parametrized correctness test is missing the pre-SM80 skip

`tests/attention/test_mla_page.py:228`, `test_nvfp4_quantize_append_paged_mla_kv_cache`, creates `float8_e4m3fn` cache tensors and actually runs the new kernel, but its body never calls `_skip_if_fp8_e4m3_unsupported()`. The other three new tests in this file (`..._dequant_close` at line 304, `..._rejects_bad_scale` at line 384, and the dense-decode test) all call it. This test is parametrized over `kv_len_configs`(6) × `page_size`(3) × `dtype`(2) × `permute_pages`(2) = **72** combinations — the widest-coverage test in the new suite — and it will fail (rather than skip) on pre-SM80 test environments. Please add the missing `_skip_if_fp8_e4m3_unsupported()` call at the top of the test body.

elwhyjay:
@qsang-nv Thanks for the careful review, both points were valid. I fix both points. PTAL

elwhyjay:
Hi @qsang-nv, Thanks for triggering the CI run. The JIT Unittest 2 failure looks like my test code: the interop test passes the scales flattened to 1D, which the pre-SM90 fallback of e2m1_and_ufp8sf_scale_to_float does not accept. It is a one-line fix, is it okay if I fix and push?

qsang-nv:
> Hi @qsang-nv, Thanks for triggering the CI run. The JIT Unittest 2 failure looks like my test code: the interop test passes the scales flattened to 1D, which the pre-SM90 fallback of e2m1_and_ufp8sf_scale_to_float does not accept. It is a one-line fix, is it okay if I fix and push?

Sure

elwhyjay:
@qsang-nv I fixed it. PTAL.

qsang-nv (approved):

PR #3141, opened 2026-04-21
Add TGV NVFP4 GEMM tactic to mm_fp4 cute-dsl backend (SM100/SM103)

Sinestro38:
## Summary
- Ports the TGV NVFP4 GEMM CuTe-DSL kernel (low-latency Blackwell GEMM tuned for small-N / decode shapes) to `flashinfer/cute_dsl/tgv_gemm_nvfp4.py`.
- Exposes it as an **additional tactic inside the existing `"cute-dsl"` backend** of `mm_fp4` (not a new top-level backend). The autotuner picks between the standard SM100/SM103 tactics and TGV.
- Tactic is only added when the problem satisfies the TGV alignment constraints: NVFP4, `M%128==0`, `N%8==0`, `K%256==0`.

## Notes
- Uses the `tcgen05.cp` (non-`tcgen05.st`) SFB path, which matches the layout produced by `nvfp4_quantize(sfLayout=layout_128x4, do_shuffle=False)` so scale factors pass through zero-copy.
- `alpha` is applied post-hoc (`out.mul_`) since the kernel doesn't fuse it.
- Compiled kernels cached by `(out_dtype, enable_pdl)`; problem M/N/K are runtime-dynamic.
- The tactic falls back to an M-contiguous scratch + copy-back when `out` is row-major (TGV writes M-contiguous). Allocating `out` column-major upstream avoids this extra copy.

## Test plan
- [ ] Run on SM100 hardware: `mm_fp4(..., backend="cute-dsl")` with autotuning enabled on decode-shape problems (e.g., `M=128, N=8, K=4096`) and confirm the TGV tactic wins.
- [ ] Verify numerical match against a cutlass/cudnn reference.
- [ ] Benchmark latency vs the existing SM100/SM103 cute-dsl tactics on small-N shapes.

## Summary by CodeRabbit

* **New Features**
  * Added a TGV NVFP4 GEMM backend for FP4 matrix multiplication when shape constraints and NVFP4 are enabled.
  * DSL interface now exposes the TGV NVFP4 kernel and a compile helper when the DSL is available.
  * Runtime handles differing output layouts, caches compiled kernels, and applies scalar scaling after kernel execution when needed.

* **Documentation**
  * Updated mm_fp4 docs to describe the new tactic, selection rules, and execution behavior.

PR #4685, opened 2026-08-22
Add CuteDSL Low Latency GEMM tactic to mm_fp4, mm_fp8, mm_mxfp8, tgv_gemm_sm100 backends (SM100/SM103)

PetersonGuo:
## 📌 Description

  Adds the CuTe-DSL low-latency block-scaled GEMM kernel and integrates it behind existing FlashInfer GEMM APIs.

  - Adds the low-latency Blackwell block-scaled GEMM kernel with FP4, FP8, and mixed FP8 × FP4 support.
  - Exposes the kernel as the "cutedsl_low_latency" backend for mm_fp4, mm_mxfp8, and mm_fp8.
  - Extends tgv_gemm_sm100 with automatic dispatch between dense BF16/FP16 GEMM and block-scaled FP4/FP8 GEMM.
  - Supports optional bias and an FP32 output scale applied in accumulator precision before bias and output conversion.
  - Supports scale and bias correctly for both regular and split-K execution.
  - Generates all supported CTA-K, pipeline-stage, and split-K tactics, filtered by layout, TMEM, and SMEM constraints.
  - Adds static FP8 support using unit E8M0 block scales, allowing the existing alpha argument to be applied by the block-scaled kernel.
  - Preserves existing backend heuristics and adds "cutedsl_low_latency" as the final eligible automatic-dispatch candidate.
  - Keeps existing dense TGV behavior and weight preparation unchanged.
  - Documents backend layout, dtype, architecture, shape, and alignment requirements through the public API docstrings.
  - Adds correctness coverage for every generated tactic, GPT-OSS-120B and DeepSeek-V3 shapes, edge cases, alpha, bias, PDL, and mixed operand formats.

  The low-latency backend requires SM100 or SM103 and currently supports public GEMM shapes with M <= 8. NVFP4 requires K divisible by 64; MXFP4, MXFP8, and static FP8 require K divisible by 128.

  Overall diff versus main: approximately 3,470 net lines added, primarily from the new kernel and its correctness coverage.

  ## 🔍 Related Issues

  N/A

  ## 🚀 Pull Request Checklist

  ✅ Pre-commit Checks

  - Pre-commit checks were run against the changed files.
  - Mypy, Ruff checks, Ruff formatting, Python compilation, and repository hygiene checks pass.

  ## 🧪 Tests

  - Low-latency kernel tactic correctness: 17 passed.
  - Public FP4 and MXFP8 low-latency dispatch coverage: 19 passed.
  - FP8 backend coverage: 54 passed, 6 expected skips for unsupported shapes.
  - Quantized tgv_gemm_sm100 coverage: 4 passed.
  - Every generated tactic is checked with alpha and bias.
  - Coverage includes NVFP4, MXFP4, MXFP8, FP8 × FP8, mixed FP8 × FP4, split-K, PDL, GPT-OSS-120B shapes, DeepSeek-V3 shapes, and edge-case dimensions.

## Reviewer Notes
Closes #3141 as this builds on top.

## Summary by CodeRabbit

- **New Features**
  - Added low-latency block-scaled GEMM support for FP8, MXFP8, and FP4 workloads on supported Blackwell GPUs.
  - Added CuTe-DSL access, automatic tactic tuning, optional bias and output scaling, and split-K execution.
  - Added block-scaled input support to TGV GEMM, including scaling factors and bias.
  - Added support for model-shaped workloads.

- **Bug Fixes**
  - Improved validation for unsupported layouts and misaligned matrix dimensions.

- **Tests**
  - Expanded coverage across formats, architectures, backends, workloads, edge cases, and correctness checks.


flashinfer/gemm/gemm_base.py
```diff
@@ -4633,27 +4694,68 @@ def _expand_block_scale_tensor_shape(block_scale_tensor, batch_size):
...
+
+
+@backend_requirement(
+    {
+        "trtllm_low_latency": _trtllm_low_latency_gemm_fp8_requirement,
+        "low_latency": _low_latency_blockscaled_gemm_fp8_requirement,
+    }
+)
 @flashinfer_api(trace=mm_fp8_trace)
 def mm_fp8(
```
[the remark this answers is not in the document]
saltyminty: Looks like a false positive

flashinfer/gemm/gemm_base.py
```diff
@@ -2075,41 +2079,74 @@ def tgv_gemm_sm100(
...
-        )
-
-    if a.dtype != b.dtype:
-        raise ValueError(
-            f"Input tensors must have the same dtype. Got {a.dtype} and {b.dtype}."
-        )
+    fp4_dtype = get_native_fp4_dtype()
+    quantized_dtypes = (fp4_dtype, torch.float8_e4m3fn, torch.float8_e5m2)
+    is_blockscaled = a.dtype in quantized_dtypes
+    if is_blockscaled:
```
saltyminty, line 2205: I believe this section is missing a check for scale tensors' dimensionality and shape

flashinfer/gemm/gemm_base.py
```diff
@@ -4633,27 +4694,68 @@ def _expand_block_scale_tensor_shape(block_scale_tensor, batch_size):
...
+    ):
+        raise ValueError("low_latency mm_fp8 requires FP8 operands")
+    _check_cute_dsl_availability()
+    return True
+
+
+@backend_requirement(
+    {
+        "trtllm_low_latency": _trtllm_low_latency_gemm_fp8_requirement,
+        "low_latency": _low_latency_blockscaled_gemm_fp8_requirement,
```
saltyminty: I feel "cutedsl_low_latency" would be a cleaner/more clear than just naming the backend "low_latency," I'll let @aleozlx chime in

saltyminty (commented):
Left two minor comments

PetersonGuo:
+@Sinestro38 for viz

PetersonGuo:
Reviewer to close #3141 after merge

flashinfer/gemm/gemm_base.py
PetersonGuo: renamed to cutedsl_low_latency

Sinestro38:
does this version optimize away unnecessary ULOP3 instructions materialized in tcgen05.st pathway for sfb? @PetersonGuo 

Also didn't read through all line diffs, but is this generalizing same TGV nvfp4 kernel in #3141? Been a bit busy so that PR fell stray.

PetersonGuo:
> does this version optimize away unnecessary ULOP3 instructions materialized in tcgen05.st pathway for sfb? @PetersonGuo
> 
> Also didn't read through all line diffs, but is this generalizing same TGV nvfp4 kernel in #3141? Been a bit busy so that PR fell stray.

Yes just generalization

tests/gemm/test_low_latency_blockscaled_gemm.py
```diff
@@ -0,0 +1,325 @@
+"""Kernel-level tests for the low-latency block-scaled GEMM."""
+
+import pytest
+import torch
+
+cutlass = pytest.importorskip("cutlass")
+
+from flashinfer.gemm.gemm_base import _low_latency_blockscaled_gemm_runner
```
dhiraj113: **[CRASH]** This entire new kernel test file fails at pytest collection: it imports `_low_latency_blockscaled_gemm_runner`, which no longer exists.

Commit 5289b19f renamed the backend to `cutedsl_low_latency` and the factory in `gemm_base.py` to `_cutedsl_low_latency_blockscaled_gemm_runner` (`gemm_base.py:6522`), but this import (and the call in `_make_runner()` at line ~89) was not updated, and no alias remains. The `pytest.importorskip("cutlass")` on line 6 does not guard this, so the module raises ImportError at collection on every machine (including non-SM100 CI), meaning none of the can_implement cases or the all-tactics correctness sweep can run — the "17 passed" in the PR description is not reproducible at this head. Fix: rename both usages to `_cutedsl_low_latency_blockscaled_gemm_runner`.

<details>
<summary>Repro — verified at PR head 259eddaf (static, no GPU needed)</summary>

```bash
python -m pytest tests/gemm/test_low_latency_blockscaled_gemm.py --collect-only -q
```

Observed:
```
_______ ERROR collecting tests/gemm/test_low_latency_blockscaled_gemm.py _______
ImportError while importing test module '.../tests/gemm/test_low_latency_blockscaled_gemm.py'.
tests/gemm/test_low_latency_blockscaled_gemm.py:8: in <module>
    from flashinfer.gemm.gemm_base import _low_latency_blockscaled_gemm_runner
E   ImportError: cannot import name '_low_latency_blockscaled_gemm_runner' from 'flashinfer.gemm.gemm_base'
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
no tests collected, 1 error in 0.60s
```
</details>

tests/gemm/test_mm_fp8.py
```diff
@@ -17,17 +17,21 @@
 @pytest.mark.parametrize("input_dtype", [torch.float8_e4m3fn])
 @pytest.mark.parametrize("mat2_dtype", [torch.float8_e4m3fn])
 @pytest.mark.parametrize("res_dtype", [torch.bfloat16])
+@pytest.mark.parametrize("backend", ["trtllm_low_latency", "low_latency"])
```
dhiraj113: **[CRASH]** `"low_latency"` is not a registered `mm_fp8` backend, so every new small-m parametrization hard-fails with `BackendSupportedError` on the SM100/SM103 hardware this PR targets.

The backend was renamed to `cutedsl_low_latency` in 5289b19f (per review feedback) and the registration in `gemm_base.py` (`@backend_requirement` keys and the `supported_backends` tuple around `gemm_base.py:4809`/`4880`) was updated, but this parametrize list still says `"low_latency"`. On SM100 the skip guard at line 33 only fires for `m > 8`, so all 24 `m <= 8` cases (of the 30 selected) reach `mm_fp8(..., backend="low_latency")` and raise — the new fp8 backend gets zero passing integration coverage and the suite goes red. Fix: rename to `"cutedsl_low_latency"` here and in the skip condition on line 33.

<details>
<summary>Repro — verified on B200 (sm100), PR head 259eddaf</summary>

```bash
python -c "from flashinfer import mm_fp8; print({b: mm_fp8.is_backend_supported(b) for b in ('trtllm_low_latency','cutedsl_low_latency','low_latency')})"
pytest -x -q tests/gemm/test_mm_fp8.py -k "low_latency and not trtllm"
```

Observed:
```
{'trtllm_low_latency': True, 'cutedsl_low_latency': True, 'low_latency': False}
>                       raise BackendSupportedError(
E                       flashinfer.utils.BackendSupportedError: mm_fp8 does not support backend 'low_latency' with capability 100
flashinfer/utils.py:1432: BackendSupportedError
FAILED tests/gemm/test_mm_fp8.py::test_mm_fp8[low_latency-res_dtype0-mat2_dtype0-input_dtype0-8192-2560-1]
1 failed, 30 deselected, 13 warnings in 0.77s
```
</details>

tests/gemm/test_mm_fp4.py
```diff
@@ -146,8 +152,25 @@ def _test_mm_fp4(
 ]
 
 
+_LOW_LATENCY_MODEL_CASES = [
```
dhiraj113: **[BUG]** All 10 `_LOW_LATENCY_MODEL_CASES` silently skip on every GPU — `"low_latency"` is not the registered backend name.

The backend was renamed to `cutedsl_low_latency` in 5289b19f (registration at `gemm_base.py:7678`), but this case list was not updated. `_test_mm_fp4` first calls `mm_fp4.is_backend_supported(backend, cc)` (`test_mm_fp4.py:32`), which returns `False` for any unregistered name, so every case hits `pytest.skip` at line 33 — including on SM100/SM103, the exact hardware these cases target. CI stays green while the new backend gets zero mm_fp4 coverage; with the corrected name the same case passes the guard and dispatches into the backend (in my environment it then fails on the separate `cute.dsmem` DSL-version issue flagged in another comment — that is orthogonal to the name mismatch). Fix: rename `"low_latency"` to `"cutedsl_low_latency"` in this case list and in the skip conditions at lines 42 and 56–64 (same stale name also appears in `test_mm_fp8.py` and `test_mm_mxfp8.py`).

<details>
<summary>Repro — verified on B200 (sm100), PR head 259eddaf</summary>

```bash
python -m pytest tests/gemm/test_mm_fp4.py -k low_latency -rs -q --no-header
python -c 'from flashinfer import mm_fp4
print("low_latency:        ", mm_fp4.is_backend_supported("low_latency", 100))
print("cutedsl_low_latency:", mm_fp4.is_backend_supported("cutedsl_low_latency", 100))'
```

Observed:
```
SKIPPED [10] tests/gemm/test_mm_fp4.py:33: Skipping test for low_latency because it is not supported on compute capability 100.
10 skipped, 23 deselected, 13 warnings in 0.24s
low_latency:         False
cutedsl_low_latency: True
```

Calling the file's own `_test_mm_fp4(3, 129, 320, torch.bfloat16, "cutedsl_low_latency", True, False, "nvfp4")` with the corrected name passes the skip guard and dispatches into the backend (it then raised `AttributeError: module 'cutlass.cute' has no attribute 'dsmem'` here — nvidia-cutlass-dsl 4.8.0a0 predates the kernel's `cute.dsmem` usage, covered separately).
</details>

tests/gemm/test_mm_mxfp8.py
```diff
@@ -195,25 +195,36 @@ def test_mm_mxfp8_large_dimensions(m, n, k, input_dtype, out_dtype, backend):
...
-        (32, 5376, 4096),
+        (4, 6144, 4096, "cutlass"),
+        (8, 6144, 4096, "cutlass"),
+        (16, 6144, 4096, "cutlass"),
+        (32, 2688, 1856, "cutlass"),
+        (32, 1856, 2688, "cutlass"),
+        (32, 2688, 4096, "cutlass"),
+        (32, 5376, 4096, "cutlass"),
+        # GPT-OSS-120B (K padded to 128)
+        (1, 1280, 2944, "low_latency"),
```
dhiraj113: **[BUG]** All 9 new `low_latency` mxfp8 small-m cases silently skip on every GPU because the parametrize name doesn't match the registered backend name.

These cases pass `backend="low_latency"` into `_run_mm_mxfp8`, whose first step `_skip_if_unsupported` (`test_mm_mxfp8.py:51`) calls `mm_mxfp8.is_backend_supported(...)` — but the backend was renamed to `cutedsl_low_latency` in 5289b19f and is registered under that name (`gemm_base.py:5799`), so the check returns `False` unconditionally and every new GPT-OSS/DeepSeek shape skips even on SM100 where the backend is supported. The result is zero mm_mxfp8 integration coverage for the new backend, masked as skips rather than failures. Fix: rename these 9 parametrize entries to `"cutedsl_low_latency"` (same stale name also appears in the other test files touched by this PR).

<details>
<summary>Repro — verified on B200 (sm100), PR head 259eddaf</summary>

```bash
python - <<"PY"
import subprocess, sys, torch
from flashinfer.gemm.gemm_base import mm_mxfp8
from flashinfer.utils import get_compute_capability
cc = get_compute_capability(torch.device("cuda"))
ccn = cc[0]*10 + cc[1]
print("cc:", ccn)
print('is_backend_supported("low_latency"):        ', mm_mxfp8.is_backend_supported("low_latency", ccn))
print('is_backend_supported("cutedsl_low_latency"):', mm_mxfp8.is_backend_supported("cutedsl_low_latency", ccn))
r = subprocess.run([sys.executable, "-m", "pytest",
    "tests/gemm/test_mm_mxfp8.py::test_mm_mxfp8_small_m[1-1280-2944-low_latency]",
    "-rs", "-q", "--no-header"], capture_output=True, text=True)
print(r.stdout[-600:])
PY
```

Observed:
```
cc: 100
is_backend_supported("low_latency"):         False
is_backend_supported("cutedsl_low_latency"): True
=========================== short test summary info ============================
SKIPPED [1] tests/gemm/test_mm_mxfp8.py:51: Skipping test because mm_mxfp8 backend is not supported on compute capability 100.
1 skipped, 13 warnings in 0.24s
```
</details>

dhiraj113 (commented):
Review of the CuTe-DSL low-latency block-scaled GEMM integration, scoped to **crashes and correctness bugs only** (design/docs notes withheld for now).

Every finding below was verified by **executing a repro against this PR's head (`259eddaf`) on a B200 (sm100)** — each inline comment carries the exact command and its observed output. This was an AI-assisted review (per `docs/code_review_guidance.md`); all findings survived an adversarial verification pass, and none are speculative.

The dominant theme: the `low_latency` → `cutedsl_low_latency` rename (commit `5289b19f`) updated the backend registration but not the tests, so **at this head the new backend has effectively zero live test coverage**, and the pass counts in the PR description cannot be reproduced:

1. `test_low_latency_blockscaled_gemm.py` dies at collection — it imports the pre-rename symbol `_low_latency_blockscaled_gemm_runner` (line 8). All 17 claimed tests are uncollectable on any machine.
2. `test_mm_fp8.py` parametrizes the stale name `"low_latency"` — on SM100 the m≤8 cases hard-FAIL with `BackendSupportedError` (24 fail / 6 skip).
3. `test_mm_fp4.py`'s 10 `_LOW_LATENCY_MODEL_CASES` silently skip everywhere (unknown backend name reads as "unsupported"); the identical case under the registered name dispatches into the backend.
4. `test_mm_mxfp8.py`'s 9 small-m cases silently skip for the same reason.
5. The kernel uses `cute.dsmem`, which is absent from the `nvidia-cutlass-dsl` builds we could obtain (incl. a 4.8.0a0 nightly) while `pyproject` pins only `>=4.7.0a0` — dispatching into the backend raises an `AttributeError` from inside the kernel. Please state/pin the minimum DSL version and feature-gate the requirement check.
6. `backend="auto"` on `mm_fp4`/`mm_mxfp8` raises `RuntimeError` instead of falling back to cudnn/cutlass when the optional DSL package is missing (the new requirement gates raise; `suitable_auto_backends` only catches `ValueError`).
7. Under PDL (hardcoded on for the fp8 path), `griddepcontrol_wait` only orders the B/SFB loads — the A/SFA TMA loads and the epilogue's output-scale read are unordered w.r.t. the preceding kernel (code-reading-confirmed; bias is correctly gated).
8. `cute.assume(n, 32)` is violated on every dispatched call (kernel-n = user-M ≤ 8) and the assumes are never validated by `can_implement` — a real, exploited compiler contract (traces to `Int32 ?{div=32}`), i.e. latent-miscompile UB even though results are currently correct.
9. The pure-FP8 test operands decode to denormal magnitudes, so the `atol=1e-1` correctness checks are near-vacuous — an all-zeros GEMM passes `assert_close` for every pure-FP8 format (the FP4 control correctly rejects it).

Items 1–4 are mechanical renames; 5–6 are gating/dependency fixes; 7–9 need kernel/test attention. Because of item 5 we could not execute the kernel end-to-end, so mainloop/epilogue/split-K correctness assessment rests on static review — re-running the full suite after the renames (with meaningful FP8 magnitudes per item 9) and updating the description's test results would put the PR on solid ground.

flashinfer/gemm/kernels/cute_dsl/low_latency_blockscaled_gemm.py
```diff
@@ -0,0 +1,2545 @@
...
+            # Arm the local mailbox before peer stores can arrive
+            if epi_tid == 0:
+                cute.arch.mbarrier_arrive_and_expect_tx(
+                    mailbox_mbar, self._mailbox_tx_total
+                )
+
+            # Complete tensor memory loads before the distributed scatter
+            cute.arch.barrier(barrier_id=15, number_of_threads=128)
+
+            dsmem_payload = cute.dsmem.B32PayloadCodec([("value", cutlass.Float32)])
```
dhiraj113: **[BUG]** The kernel requires a `cutlass.cute.dsmem` API that does not exist in the DSL versions the package allows, so dispatching into `cutedsl_low_latency` raises `AttributeError` from inside the kernel trace instead of a clean gate rejection.

`cute.dsmem.B32PayloadCodec` / `cute.dsmem.store_to_peer` (`low_latency_blockscaled_gemm.py:1819,1832`, reached whenever the chosen tactic has `split_k != 1`) are absent from the current nvidia-cutlass-dsl 4.8.0a0 nightly (2026-08-18) and from all published builds, yet `pyproject.toml:28-29` only pins `>=4.7.0a0` and the requirement gates in `gemm_base.py` (`_check_cute_dsl_availability()`, e.g. `gemm_base.py:4805`) only check that the DSL imports, not that it is new enough. On such an install the backend passes the gate — including in auto mode — and then explodes mid-compile. Please state/pin the minimum DSL version that ships `cute.dsmem`, and add a feature check (e.g. `hasattr(cutlass.cute, "dsmem")`) to the requirement gate so unsupported installs are skipped in auto mode and rejected with a clear message when the backend is requested explicitly.

<details>
<summary>Repro — verified on B200 (sm100), PR head 259eddaf</summary>

```bash
python - <<"PY"
import importlib.metadata, torch, cutlass.cute as cute
from flashinfer import mm_fp8

print("nvidia-cutlass-dsl:", importlib.metadata.version("nvidia-cutlass-dsl"))
print("hasattr(cute, 'dsmem'):", hasattr(cute, "dsmem"))

a = torch.randn(8, 128, device="cuda").to(torch.float8_e4m3fn)
b = torch.randn(128, 128, device="cuda").to(torch.float8_e4m3fn)
alpha = torch.ones((), device="cuda", dtype=torch.float32)
mm_fp8(a, b, alpha, backend="cutedsl_low_latency")
PY
```

Observed:
```
nvidia-cutlass-dsl: 4.8.0a0+20260818210715.9e27bc7
hasattr(cute, 'dsmem'): False
  File ".../flashinfer/gemm/kernels/cute_dsl/low_latency_blockscaled_gemm.py", line 1796, in epilog_warp
    if self.split_k == 1:
  File ".../flashinfer/gemm/kernels/cute_dsl/low_latency_blockscaled_gemm.py", line 1819, in else_block_4
    dsmem_payload = cute.dsmem.B32PayloadCodec([("value", cutlass.Float32)])
                    ^^^^^^^^^^
AttributeError: module 'cutlass.cute' has no attribute 'dsmem'
```
</details>

flashinfer/gemm/gemm_base.py
```diff
@@ -6177,6 +6438,44 @@ def _cute_dsl_gemm_fp4_requirement(
...
+        )
+    real_k = a.shape[1] * 2
+    k_alignment = 64 if use_nvfp4 else 128
+    if a.shape[0] > 8 or real_k % k_alignment != 0:
+        if backend != "cutedsl_low_latency":
+            return False
+        raise ValueError(
+            f"cutedsl_low_latency FP4 GEMM requires M <= 8 and K divisible by {k_alignment}"
+        )
+    _check_cute_dsl_availability()
```
dhiraj113: **[BUG]** `backend="auto"` fails with `RuntimeError('CuTe DSL is not available.')` on small-M FP4 GEMMs when `nvidia-cutlass-dsl` is not installed, instead of falling back to cudnn/cutlass.

`_cutedsl_low_latency_gemm_fp4_requirement` ends with `_check_cute_dsl_availability()` (`gemm_base.py:6475`), which raises `RuntimeError` (`gemm_base.py:5402-5409`), but `suitable_auto_backends` (`utils.py:1362-1368`) probes every registered backend's requirement and only catches `ValueError` — so on an SM100/SM103 box without the optional DSL package, any `mm_fp4(..., backend="auto")` with M<=8, K%64==0 and matching layouts dies even though cudnn/cutlass are suitable. The `CUTE_DSL_AVAILABLE` guard in the heuristic never helps because the exception fires in the suitability loop, before the heuristic runs. The same pattern exists in the new mxfp8 gate (~`gemm_base.py:5287`), and while the pre-existing `_cute_dsl_gemm_fp4_requirement` shares the flaw, this PR extends it to the common small-M decode shapes.

Fix direction: in the gates, `return False` when the DSL is unavailable in auto mode (raising only for an explicit `backend="cutedsl_low_latency"`), mirroring the shape checks just above — or make `_check_cute_dsl_availability` raise `ValueError`.

<details>
<summary>Repro — verified on B200 (sm100), PR head 259eddaf</summary>

```bash
python - <<"PY"
import torch
from flashinfer import SfLayout, mm_fp4, nvfp4_quantize
import flashinfer.cute_dsl.utils as cute_dsl_utils

torch.manual_seed(0)
m, n, k = 4, 256, 256  # M<=8, K%64==0 -> passes the cutedsl_low_latency shape gate
inp = torch.randn([m, k], device="cuda", dtype=torch.bfloat16)
mat2 = torch.randn([n, k], device="cuda", dtype=torch.bfloat16)
gs_in = (448 * 6) / inp.float().abs().max()
gs_w = (448 * 6) / mat2.float().abs().max()
a_fp4, a_sf = nvfp4_quantize(inp, gs_in, sfLayout=SfLayout.layout_128x4, do_shuffle=False)
b_fp4, b_sf = nvfp4_quantize(mat2, gs_w, sfLayout=SfLayout.layout_128x4, do_shuffle=False)
out = torch.empty([m, n], device="cuda", dtype=torch.bfloat16)

def call_auto():
    mm_fp4(a_fp4, b_fp4.T, a_sf, b_sf.T, 1.0 / (gs_in * gs_w), torch.bfloat16,
           out, block_size=16, use_8x4_sf_layout=False, backend="auto", use_nvfp4=True)

call_auto()  # control: DSL present -> auto works
torch.cuda.synchronize()
print("CONTROL OK: backend='auto' succeeded with CuTe DSL available")

# Simulate nvidia-cutlass-dsl not installed (it is an optional extra):
cute_dsl_utils.is_cute_dsl_available = lambda: False
try:
    call_auto()
    print("no bug: auto fell back to another backend")
except RuntimeError as e:
    print(f"BUG: RuntimeError escaped backend='auto': {e!r}")
PY
```

Observed:
```
CONTROL OK: backend='auto' succeeded with CuTe DSL available
BUG: RuntimeError escaped backend='auto': RuntimeError('CuTe DSL is not available.')
```
</details>

flashinfer/gemm/kernels/cute_dsl/low_latency_blockscaled_gemm.py
```diff
@@ -0,0 +1,2545 @@
...
+                cute.group_modes(sSFB, 0, 3),
+                cute.group_modes(tCgSFB, 0, 3),
+            )
+            tBsSFB = cute.filter_zeros(tBsSFB)
+            tBgSFB = cute.filter_zeros(tBgSFB)
+        tBgSFB = tBgSFB[(None, work_tile_info.N_idx, None, work_tile_info.L_idx)]
+
+        # PDL: wait on dependent grids only for B (the activation tensor)
+        if self.use_pdl:
+            cute.arch.griddepcontrol_wait()
```
dhiraj113, line 1222: **[BUG]** Under PDL, `griddepcontrol_wait()` only orders the B/SFB loads: the DMA_A warp's A/SFA TMA loads and the epilogue's `mScale[0]` read can observe stale GMEM from the predecessor kernel.

PDL wait semantics are per-thread, and the single `griddepcontrol_wait()` lives in `dma_b_warp` (`low_latency_blockscaled_gemm.py:1226`, comment says "only for B"). But `dma_a_warp` issues the A and SFA TMA loads (`:1114`, `:1120`) with no wait at all, and the epilogue reads the alpha scale at `:1746` before either barrier that chains back to DMA_B's wait (`tma_epilog_full_bar` at `:1771` — which correctly gates the bias load — and the mma_epilog wait at `:1778`). This is reachable in production: the mm_fp8 path fills the neutral E8M0 SFA/SFB workspace on-stream (`gemm_base.py:4935`) and then hardcodes `enable_pdl=True` (`gemm_base.py:4941`), so with PDL overlap the un-waited SFA load can race the fill, and an alpha produced by a preceding kernel can be read stale — silently wrong output. This is confirmed by code reading (the ordering-facts checker below); no runtime race repro was attempted since PDL overlap is timing-dependent.

Fix direction: execute `griddepcontrol_wait()` in `dma_a_warp` as well (or drop the A-is-preloaded assumption), and move the `mScale[0]` read after a barrier that is ordered after DMA_B's wait, mirroring the bias gating.

<details>
<summary>Repro — verified on B200 (sm100), PR head 259eddaf</summary>

Run from the repo root (static source-fact checker; exits 1 when the unordered structure is present):

```bash
python - <<"PY"
import re, sys
src = open("flashinfer/gemm/kernels/cute_dsl/low_latency_blockscaled_gemm.py").read()
L = src.splitlines()
ln = lambda pos: src.count("\n", 0, pos) + 1
def span(name):
    m = re.search(rf"^    def {name}\(", src, re.M)
    n = re.search(r"^    (?:@|def )", src[m.end():], re.M)
    return m.start(), m.end() + (n.start() if n else 0)
ok = []
# Fact 1: exactly one griddepcontrol_wait, inside dma_b_warp
waits = [ln(m.start()) for m in re.finditer(r"griddepcontrol_wait", src)]
b0, b1 = span("dma_b_warp")
print(f"F1 wait sites: {waits}; dma_b_warp=L{ln(b0)}-{ln(b1)}")
ok.append(len(waits) == 1 and ln(b0) <= waits[0] <= ln(b1))
# Fact 2: dma_a_warp issues A+SFA TMA loads with no wait
a0, a1 = span("dma_a_warp")
body = src[a0:a1]
copies = [ln(a0 + m.start()) for m in re.finditer(r"cute\.copy\(", body)]
print(f"F2 dma_a_warp=L{ln(a0)}-{ln(a1)}; cute.copy at {copies}; "
      f"griddepcontrol_wait in body: {'griddepcontrol_wait' in body}")
ok.append(len(copies) >= 2 and "griddepcontrol_wait" not in body)
# Fact 3: epilog reads mScale[0] before both wait-ordered barriers (bias IS gated)
s = ln(re.search(r"scale_value = mScale\[0\]", src).start())
bw = ln(re.search(r"mbarrier_wait\(tma_epilog_full_bar, 0\)", src).start())
mw = ln(re.search(r"mma_epilog_full_bar \+ mma_epilog_state\.index", src).start())
print(f"F3 mScale[0] read L{s} < bias barrier wait L{bw} < mma_epilog wait L{mw}")
ok.append(s < bw < mw)
# Fact 4: mm_fp8 host path fills SFA/SFB on-stream then launches with enable_pdl=True
gb = open("flashinfer/gemm/gemm_base.py").read().splitlines()
f4 = ["fill_(127)" in l for l in gb[4930:4940]], [l.strip() == "True," for l in gb[4938:4943]]
print(f"F4 fill_(127) near L4935: {any(f4[0])}; enable_pdl=True near L4941: {any(f4[1])}")
ok.append(any(f4[0]) and any(f4[1]))
print(("ALL 4 FACTS HOLD -> A/SFA loads and alpha read are not PDL-wait-ordered"
       if all(ok) else f"facts: {ok}"))
sys.exit(1 if all(ok) else 0)
PY
```

Observed:
```
F1 wait sites: [1226]; dma_b_warp=L1143-1260
F2 dma_a_warp=L1030-1142; cute.copy at [1114, 1120]; griddepcontrol_wait in body: False
F3 mScale[0] read L1746 < bias barrier wait L1771 < mma_epilog wait L1778
F4 fill_(127) near L4935: True; enable_pdl=True near L4941: True
ALL 4 FACTS HOLD -> A/SFA loads and alpha read are not PDL-wait-ordered
```
</details>

flashinfer/gemm/kernels/cute_dsl/low_latency_blockscaled_gemm.py
```diff
@@ -0,0 +1,2545 @@
...
+        :type problem_mnkl: Tuple
+        :param stream: CUDA stream for asynchronous execution.
+        :type stream: cuda.CUstream
+        """
+        m, n, k, l = problem_mnkl
+        a = cute.make_tensor(
+            a_ptr, cute.make_ordered_layout((cute.assume(m, 32), k, l), order=(1, 0, 2))
+        )
+        b = cute.make_tensor(
+            b_ptr, cute.make_ordered_layout((cute.assume(n, 32), k, l), order=(1, 0, 2))
```
dhiraj113: **[BUG]** `cute.assume(n, 32)` / `cute.assume(m, 32)` assert divisibility contracts that every dispatched call violates — latent-miscompile UB, even though results are currently correct.

The tensor layouts here (lines 440/443/446, also the bias layout at 461) promise the DSL compiler that both problem `m` and `n` are divisible by 32. But `mm_fp8` swaps operands and passes `problem_mnkl=(n, m, k, 1)` (`gemm_base.py:4950`), so kernel-`n` is the token count, which the requirement gate caps at `a.shape[0] <= 8` (`gemm_base.py:4796`) — `assume(n, 32)` is false on every supported call — and kernel-`m` is the user's weight output dim, whose divisibility is never validated by `can_implement` (its only `% 32` check is on `cta_m`, a tile shape). This is not a doc nit: `assume` yields a constrained type (`?{div=32}`) that the installed DSL actively exploits for folding (see repro, where `assume(m,32) % 32` compiles to the constant 0 and returns 0 for `m=7`), so any current or future optimization relying on it can silently miscompile. Today the kernel happens to produce correct output, but that is unspecified behavior, not a guarantee.

Fix direction: drop the `assume(..., 32)` wrappers (or reduce to a divisor that actually holds), or validate `m % 32 == 0` / `n % 32 == 0` in `can_implement` — the latter would reject all currently supported shapes, so dropping the assumes looks right.

<details>
<summary>Repro — verified on B200 (sm100), PR head 259eddaf</summary>

```bash
cat > /tmp/assume_repro.py <<"PY" && python /tmp/assume_repro.py
import re, inspect
import cutlass, cutlass.cute as cute
import flashinfer.gemm.kernels.cute_dsl.low_latency_blockscaled_gemm as llk

src = open(llk.__file__).read().splitlines()
for ln in (440, 443, 446):
    print(f"low_latency_blockscaled_gemm.py:{ln}:", src[ln - 1].strip())
print("mm_fp8 swaps operands, problem_mnkl=(n, m, k, 1):",
      "(n, m, k, 1)" in open(gb.__file__).read())
ci = inspect.getsource(llk.LowLatencyBlockscaledGemmKernel.can_implement)
print("can_implement validates problem m%32 / n%32:",
      bool(re.search(r"(?<![\w.])m\s*%\s*32", ci)), "/",
      bool(re.search(r"(?<![\w.])n\s*%\s*32", ci)))

@cute.jit
def probe(m: cutlass.Int32) -> cutlass.Int32:
    ma = cute.assume(m, divby=32)
    print("traced type of cute.assume(m, 32):", ma)
    return ma % 32

got = cute.compile(probe, cutlass.Int32(64))(cutlass.Int32(7))
print("compiled (assume(m,32) % 32) invoked with m=7 returns:", got,
      " [true 7 % 32 = 7]")
PY
```

Observed:
```
low_latency_blockscaled_gemm.py:440: a_ptr, cute.make_ordered_layout((cute.assume(m, 32), k, l), order=(1, 0, 2))
low_latency_blockscaled_gemm.py:443: b_ptr, cute.make_ordered_layout((cute.assume(n, 32), k, l), order=(1, 0, 2))
low_latency_blockscaled_gemm.py:446: c_ptr, cute.make_ordered_layout((cute.assume(m, 32), n, l), order=(0, 1, 2))
mm_fp8 swaps operands, problem_mnkl=(n, m, k, 1): True
can_implement validates problem m%32 / n%32: False / False
traced type of cute.assume(m, 32): ?{div=32}
compiled (assume(m,32) % 32) invoked with m=7 returns: 0  [true 7 % 32 = 7]
```
</details>

flashinfer/gemm/kernels/cute_dsl/low_latency_blockscaled_gemm.py
```diff
@@ -0,0 +1,2545 @@
...
+            # FP4 is packed two elements per byte along K
+            operand = torch.empty(
+                (l, mn, k // 2), dtype=torch.int8, device="cuda"
+            ).permute(1, 2, 0)
+            operand.copy_(
+                torch.randint(-2, 2, operand.shape, dtype=torch.int8, device="cuda")
+            )
+        else:
+            # Small bit patterns avoid NaNs in E4M3FN
+            operand = torch.randint(
```
dhiraj113: **[BUG]** FP8 test operands are denormal-magnitude bit patterns, so the `atol=1e-1` correctness checks pass even for an all-zeros GEMM — the pure-FP8 test results are vacuous.

`make_operand` in `low_latency_blockscaled_gemm.py:2010` draws FP8 operands as raw bit patterns 0–3, which decode to 0 or denormals (~2^-9 for E4M3FN, ~2^-16 for E5M2). At the test shape `(129, 7, 256, 2)` with `alpha=0.125`, max `|alpha*GEMM|` ranges from ~1e-3 down to ~5.8e-8 across the four pure-FP8 combos — orders of magnitude below the `atol=1e-1` used by `compare()` and `tests/gemm/test_low_latency_blockscaled_gemm.py:112/287` — so a kernel that dropped the GEMM entirely and only applied the alpha/bias epilogue would still pass every pure-FP8 case (the FP4 path is fine: its data decodes to unit-scale values, max ~6e2). Fix direction: generate FP8 operands with normal-range magnitudes (e.g. pick bit patterns whose decoded values are O(1), as the FP4 branch effectively does) or switch the check to a relative-error-dominant tolerance sized to the actual data magnitude.

<details>
<summary>Repro — verified on B200 (sm100), PR head 259eddaf</summary>

```bash
python - <<"PY"
import torch, cutlass
from flashinfer.gemm.kernels.cute_dsl.low_latency_blockscaled_gemm import (
    _decode_ab_to_f32, make_blockscaled_tensors)
torch.manual_seed(0)
M, N, K, L = 129, 7, 256, 2          # the all-tactics correctness-test shape
ALPHA, ATOL, RTOL = 0.125, 1e-1, 1e-3  # the test's own alpha / tolerances

def rep(s, v):  # expand per-block scales to per-element, as compare() does
    return torch.repeat_interleave(s.cpu().float(), v, dim=1)[:, :K]

def check(ad, bd, sfd, vec):
    a, b, _, _, _, sfa, sfb = make_blockscaled_tensors(
        (M, N, K, L), ad, bd, sfd, vec, cutlass.BFloat16)
    bias = torch.linspace(-0.5, 0.5, M, dtype=torch.float32)
    gemm = torch.einsum("mkl,nkl->mnl",
        _decode_ab_to_f32(a) * rep(sfa, vec), _decode_ab_to_f32(b) * rep(sfb, vec))
    ref = (gemm * ALPHA + bias[:, None, None]).to(torch.bfloat16)
    zeros_gemm = (torch.zeros_like(gemm) + bias[:, None, None]).to(torch.bfloat16)
    try:
        torch.testing.assert_close(zeros_gemm, ref, atol=ATOL, rtol=RTOL)
        verdict = "PASSES assert_close  <-- BUG"
    except AssertionError:
        verdict = "rejected (ok)"
    print(f"{ad.__name__:>13} x {bd.__name__:<13} "
          f"max|alpha*GEMM|={(gemm*ALPHA).abs().max():.2e}  all-zeros GEMM {verdict}")

for ad in (cutlass.Float8E4M3FN, cutlass.Float8E5M2):
    for bd in (cutlass.Float8E4M3FN, cutlass.Float8E5M2):
        check(ad, bd, cutlass.Float8E8M0FNU, 32)
check(cutlass.Float4E2M1FN, cutlass.Float4E2M1FN, cutlass.Float8E4M3FN, 16)  # control
PY
```

Observed:
```
 Float8E4M3FN x Float8E4M3FN  max|alpha*GEMM|=1.01e-03  all-zeros GEMM PASSES assert_close  <-- BUG
 Float8E4M3FN x Float8E5M2    max|alpha*GEMM|=7.65e-06  all-zeros GEMM PASSES assert_close  <-- BUG
   Float8E5M2 x Float8E4M3FN  max|alpha*GEMM|=7.98e-06  all-zeros GEMM PASSES assert_close  <-- BUG
   Float8E5M2 x Float8E5M2    max|alpha*GEMM|=5.85e-08  all-zeros GEMM PASSES assert_close  <-- BUG
 Float4E2M1FN x Float4E2M1FN  max|alpha*GEMM|=6.04e+02  all-zeros GEMM rejected (ok)
```
</details>

flashinfer/gemm/kernels/cute_dsl/low_latency_blockscaled_gemm.py
PetersonGuo: This is a likely a false positive. `griddepcontrol_wait()` is only needed when the kernel is waiting on dependent data which is activations on B. A for this kernel is the weights which can be preloaded during the previous kernel. Unless there is a case where the A matrix is not known however this pattern is already done in flashinfer for other TGV kernels see
https://github.com/flashinfer-ai/flashinfer/blob/main/include/flashinfer/gemm/tgv_gemm.cuh#L328
https://github.com/flashinfer-ai/flashinfer/blob/main/flashinfer/gemm/kernels/tgv_gemm_cute_ext.py#L663

The alpha scale is gated because epilogue is gated by B warp

flashinfer/gemm/kernels/cute_dsl/low_latency_blockscaled_gemm.py
PetersonGuo: This is to be release to `nvidia-cutlass-dsl==4.9.0` will update to use `nvidia-cutlass-dsl==4.8.0`

flashinfer/gemm/kernels/cute_dsl/low_latency_blockscaled_gemm.py
PetersonGuo: Not a miscompile but wrong assumption for this kernel. Will update

tests/gemm/test_low_latency_blockscaled_gemm.py
dhiraj113: Verified at 409c9850: the file now collects 17 tests; all 7 `can_implement` cases pass and the fmt0 all-tactics case runs end-to-end on a B200 (JIT included, ~92s). Resolved, thanks.

tests/gemm/test_mm_fp8.py
dhiraj113: Verified at 409c9850: the m=1 `cutedsl_low_latency` case dispatches into the kernel and passes on a B200. No stale names left in the file. Resolved, thanks.

tests/gemm/test_mm_fp4.py
dhiraj113: The case-list rename is in, and the 5 nvfp4 model cases now run and pass on a B200 — thanks. One remnant from the original comment: the mxfp4 allow-list at lines 56–64 still says `"low_latency"`, so the 5 mxfp4/mxfp4_alpha model cases still skip silently (`mx_fp4 is only supported for …`). A one-line rename there fixes it (the stale reference at line 42 is now dead code, worth cleaning at the same time). Non-blocking.

tests/gemm/test_mm_mxfp8.py
dhiraj113: Verified at 409c9850: the (3,129,384) case runs and passes on a B200 (~74s with JIT), and no stale `"low_latency"` strings remain in the file. Resolved, thanks.

flashinfer/gemm/kernels/cute_dsl/low_latency_blockscaled_gemm.py
dhiraj113: Verified the `cute.dsmem` → `cute.arch.store_async_dsmem` swap at 409c9850: zero `cute.dsmem` references remain, and a split_k=2 tactic compiled and matched the reference end-to-end on `nvidia-cutlass-dsl` 4.8.0a0 (B200). Remaining: the pins weren't bumped yet — `pyproject.toml` still has `>=4.7.0a0` (both extras) and `requirements.txt` `>=4.6.2a0`, so a 4.7.x user would hit an `AttributeError` on any split-k tactic. Bumping to `>=4.8.0a0` as you mentioned (or `hasattr`-gating split-k tactics) closes it. Non-blocking.

flashinfer/gemm/gemm_base.py
dhiraj113: Verified at 409c9850: with the DSL unavailable, `backend="auto"` now falls back cleanly to cudnn/cutlass, and the `ValueError` change fixes the same pre-existing hazard in `_cute_dsl_gemm_fp4_requirement` too — nice. Tiny nit: an *explicit* `backend="cutedsl_low_latency"` call without the DSL now surfaces as a raw `ModuleNotFoundError('cutlass')` from the runner factory instead of the old clear message; since `_check_cute_dsl_availability` now raises `ValueError` (auto-safe), calling it in the gate only when `backend == "cutedsl_low_latency"` would restore the clean error without re-breaking auto. Non-blocking.

flashinfer/gemm/kernels/cute_dsl/low_latency_blockscaled_gemm.py
dhiraj113: Agreed on the A-matrix half — kernel-A is the prepared-weights operand and `tgv_gemm.cuh:365` has exactly this pattern; conceding that part. Two narrower points the weights argument doesn't cover, noting as non-blocking: (1) on the mm_fp8 path kernel-SFA isn't weights — it's the descale region carved from the workspace that `gemm_base.py` `fill_(127)`s on the same stream immediately before the `enable_pdl=True` launch, and the DMA_A warp TMA-loads it with no `griddepcontrol_wait`, so under PDL that read can overlap the fill (SFB is safe — loaded after DMA_B's wait); (2) the epilogue's `mScale[0]` read precedes the wait-ordered barriers, and alpha can be a device tensor written by a preceding kernel (bias, by contrast, is gated). If A-side SF staleness is considered acceptable by the same preloadable-weights reasoning, a short comment in the kernel saying so would settle it; the `fill_(127)` case may deserve either ordering or moving the fill ahead of the PDL region.

flashinfer/gemm/kernels/cute_dsl/low_latency_blockscaled_gemm.py
dhiraj113: Verified at 409c9850: all `assume(m/n, 32)` sites removed (the surviving `assume(k // sf_vec_size, 16)` is guarded by `can_implement`'s K checks), and the previously contract-violating M=7, N=1000, K=256 case is now bit-exact vs an fp32 reference on a B200. Resolved, thanks.

flashinfer/gemm/kernels/cute_dsl/low_latency_blockscaled_gemm.py
dhiraj113: Verified at 409c9850: operands now land at |A|,|B| = 2.0 with |alpha·GEMM| ≈ 35–40, and an all-zeros GEMM is rejected by `assert_close` in all four pure-FP8 combos (previously all four passed vacuously); the new `gemm_only.abs().max() > 1.0` assertion guards against regressions. Resolved, thanks.

dhiraj113 (approved):
Re-reviewed at head `409c9850` — every finding from the previous review was re-verified by execution on a B200 (sm100), including end-to-end kernel runs now that the `store_async_dsmem` swap compiles on nvidia-cutlass-dsl 4.8.0a0: the test suite collects and passes (fmt0 all-tactics, mm_fp8/mm_fp4-nvfp4/mxfp8 small-m cases), auto-dispatch falls back cleanly without the DSL, the `assume` contracts are gone, and the FP8 tolerance checks are now meaningful (an all-zeros GEMM is rejected).

Three small non-blocking items remain, detailed in the comment threads: the mxfp4 allow-list in `test_mm_fp4.py:56-64` still uses the old backend name (5 model cases silently skip), the `nvidia-cutlass-dsl` pin should be bumped to `>=4.8.0a0` per your reply, and the PDL discussion has two narrow residual points (SFA workspace fill ordering, epilogue scale read) — conceded on the A-matrix itself.

Approving — the integration-layer issues are all addressed. (AI-assisted review per `docs/code_review_guidance.md`.)

jimmyzho:
@flashinfer-bot run

dhiraj113:
@flashinfer-bot run

b8zhong:
Hi @PetersonGuo, I'm wondering if you have some performance comparison with the existing cute-dsl (with both of these backends under autotune)? (I'm wondering this for MXFP8 and NVFP4). Thanks a lot!

PR #4687, opened 2026-08-23
Fix W4A16 large weight bank addressing

earlvanze:
## 📌 Description

Fix the Blackwell W4A16 fused-MoE path for pooled expert weight banks whose
flattened shape exceeds the signed 32-bit range.

The current custom-op ABI passes `w13` and `w2` as dynamic CuTe tensors. A
large serving cache can exceed `INT32_MAX` elements (the reproduced W13 bank
had 10,151,264,256 int32 elements), which overflows the runtime memref shape
descriptor before launch. After bypassing that descriptor, expert offsets also
wrapped in 32-bit arithmetic and produced an illegal global-memory address.

This change:

- passes packed weight banks across the compile boundary as aligned raw CuTe
  pointers and rebuilds their compile-time flat layouts inside the kernel;
- keeps expert-bank, packed, and NF3 global-read offsets in 64-bit arithmetic;
- adds a source-contract regression test for the pointer ABI and Int64 address
  path.

Validation on one NVIDIA RTX PRO 6000 Blackwell 96 GB GPU with a 753B GLM-5.2
NVFP4 checkpoint and a 419 GB system-RAM expert bank:

- the original path failed with the memref overflow;
- a pointer-only build reached launch but hit an illegal address;
- this pointer + Int64 build completed 16-, 64-, and 256-token generations;
- four simultaneous 128-token requests also completed through one shared model
  server (508 completion tokens total).

This is a correctness fix, not a performance claim. On that host the corrected
b12x path reached 7.343 steady decode tokens/s for the 256-token case and 9.671
aggregate tokens/s for four simultaneous requests.

## 🔍 Related Issues

No existing matching issue found.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

Focused validation completed: `ruff check`, `ruff format --check`,
`py_compile`, the AST source-contract assertions, and live Blackwell GLM-5.2
generation/concurrency. The full FlashInfer GPU matrix was not available
locally.

## Reviewer Notes

Please focus on whether raw pointer arguments are the preferred CuTe DSL ABI
for compile-time-known flat weight-bank layouts and whether any remaining
weight-bank offset can be narrowed before global address formation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed fused MoE processing for weight banks larger than 2³¹ elements.
  * Prevented address and offset overflow during weight loading, staging, and GEMM execution.
  * Improved handling across packed, ModelOpt, and NF3 weight layouts.

* **Tests**
  * Added regression coverage for large weight-bank addressing and 64-bit offset handling.

### DIFF
--- a/flashinfer/fused_moe/cute_dsl/blackwell_sm12x/moe_w4a16_kernel.py
+++ b/flashinfer/fused_moe/cute_dsl/blackwell_sm12x/moe_w4a16_kernel.py
@@ -1490,8 +1490,12 @@ class W4A16GemmKernel:
                 scales_expert_stride = Int32((self.size_n * self.size_k) // (32 * 16))
         else:
             scales_expert_stride = Int32((self.size_n * self.size_k) // (16 * 16))
-        b_expert_off = (
-            Int32((self.size_n * self.size_k) // (_PACK_FACTOR * 4)) * expert_idx
+        # A pooled expert cache may make the flat packed bank larger than 2 Gi
+        # int32 elements. Keep weight-bank addressing 64-bit end-to-end; an
+        # Int32 expert offset silently wraps and makes cp.async read an invalid
+        # address even when the bank itself crossed the ABI as a raw pointer.
+        b_expert_off = Int64((self.size_n * self.size_k) // (_PACK_FACTOR * 4)) * Int64(
+            expert_idx
         )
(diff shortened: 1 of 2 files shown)


earlvanze:
@flashinfer-bot run

PR #3908, opened 2026-07-10
Ameyn/gdn wy perf followup

ameynaik-hub:
## 📌 Stacked on #3720

This branch builds on #3720's head (`00a655e`); the diff will show its commits until it merges. **Review only the 5 commits listed below.** Once #3720 lands, the diff collapses to just these changes (one rebase needed only if #3720 is squash-merged).

## What

Performance and correctness follow-ups to the WY output-only GDN decode kernel, driven by Nsight Compute analysis on B200. Two files: `flashinfer/gdn_kernels/gdn_decode_bf16_wy_output_only.py`, `benchmarks/bench_gdn_decode.py`.

### Performance
- **`49a7ccf` Raise launch-bounds cap 4→8.** With `min_blocks_per_mp` capped at 4, the compiler used 73 regs/thread, limiting occupancy to 6 CTAs/SM (37.5%). Requesting 8 fits 64 regs/thread with no spills → 7 CTAs/SM (43.75%). Adds `GDN_WY_MBP` env override for tuning experiments.
- **`39693d1` SMEM-staged coalesced output epilogue.** The MMA-fragment-direct 4-byte stores wasted half of every 32-byte sector (NCU: 23% of all L2 sector traffic). The output tile now stages through `sH`'s SMEM (dead at that point — zero extra usage) and flushes with consecutive-lane 16-byte stores at 100% sector utilization. Excessive sectors: 23% → 5%.
- **`6d34642` Enable native-short-T + native-a/b by default for T∈{4,8}.** The kernel reads the real `[B,T,…]` tensors (loads only the T valid rows), removing 4 of 5 host staging copies per call. Previously gated off due to an accuracy signal that traced to the bf16-cache bug fixed in `856f21f`, not this path. `SGLANG_GDN_WY_NATIVE_T=0` / `SGLANG_GDN_WY_NATIVE_AB=0` restore staging.

### Benchmarking
- **`45e6956` Time the wy_output_only head-to-head with CUPTI + cold L2** — the same `bench_gpu_time(enable_cupti=True, cold_l2_cache=True)` methodology as every other version in the file

## Benchmark results

```bash
python benchmarks/bench_gdn_decode.py --version bf16_wy_output_only \
  --num-q-heads 16 --num-k-heads 16 --num-v-heads 64 --head-size 128 \
  --seq-len 4 8 16 --batch-size 1 2 4 8 16 32 64 128 256 \
  --warmup 10 --iters 1000
```

[...]

PR #4709, opened 2026-08-24
feat(kda): output-only KDA decode + drop-in for vLLM Kimi K3 recurren…

ameynaik-hub:
# Output-only KDA decode + drop-in for vLLM Kimi K3 recurrent verify

An output-only (frozen-state) Kimi Delta Attention decode kernel — the KDA
analog of the GDN WY output-only kernel (flashinfer #3908) — and a verified
drop-in replacement for the recurrent speculative-verify kernel from
[vLLM PR #52993](https://github.com/vllm-project/vllm/pull/52993).

## Files

| File | Role |
|---|---|
| `flashinfer/kda_kernels/kda_decode_wy_output_only.py` | kernel (self-contained; reuses PTX helpers from `gdn_decode_bf16_wy_output_only` and gate helpers from `kda_kernels/recurrent_kda`) |
| `flashinfer/kda_decode.py` | public API + docstrings |
| `flashinfer/__init__.py` | exports |
| `flashinfer/trace/templates/kda.py` | trace template |
| `docs/api/kda_decode.rst` | docs |
| `tests/kda/test_kda_output_only.py` | 39 tests |
| `benchmarks/bench_kda_output_only.py` | benchmark |

## API

1. `flashinfer.kda_output_only_decode(q, k, v, g, beta, state_pool,
   state_indices, ..., emit_corrections=False, backend="auto")`

   Computes outputs for 1..16 tokens/seq from a read-only state pool; never
   writes state. Backends: `"wy"` (WY-parallel tensor-core kernel:
   per-K-channel decay folded into `k*exp(±cumsum g)` / `q*exp(cumsum g)`
   operand tiles, T×T triangular inverse via register-resident block
   forward-substitution, TMA-streamed state GEMMs) and `"recurrent"` (a
   grouped register recurrence for tiny batches); `"auto"` dispatches by
   measured crossover. `emit_corrections=True` additionally returns
   corrections `U_t = sigmoid(beta_t) * (v_t - u_t)` and a kg cache.

2. `flashinfer.kda_recoverssm_verify(q, k, v, raw_g, raw_beta, A_log,
   dt_bias, lower_bound, checkpoint_state, correction_cache, kg_cache,
   query_start_loc, state_indices, spec_query_len, out=None)`

   Exact signature and contract of
   `vllm/models/kimi_k3/nvidia/ops/recoverssm.py::kda_recoverssm_verify`:
   varlen-packed inputs with runtime per-sequence lengths and token strides,
   null-slot semantics (`state_indices <= 0` → zero outputs, no cache
   writes), slot-indexed caches with rows past `query_len` left untouched.

## Precision

| | |
|---|---|
| inputs | `q/k/v/raw_g/raw_beta` bf16; `A_log/dt_bias` fp32; `checkpoint_state` pool bf16 (the vLLM default: `mamba_cache_dtype="auto"` → model dtype) |
| compute | bf16 tensor-core MMAs with fp32 accumulation; in-kernel L2 norm (eps 1e-6), gate transforms and beta sigmoid in fp32 |
| outputs | attention out bf16 `[1, total_tokens, H, 128]`; `correction_cache` fp32 `[blocks, H, spec_len, 128]` (stored from fp32 MMA fragments, no bf16 rounding on the store); `kg_cache` bf16 `[blocks, H, spec_len, 256]` = raw k \| raw g, bit-exact vs the vLLM kernel |
| accuracy | vs fp32 reference: outputs ~2e-4; vs vLLM's Triton kernel on identical inputs: outputs ≤ 4.9e-4 (bf16 rounding, summation order), corrections ≤ 1.6e-2 (their all-fp32 math vs bf16 MMA), kg bit-exact, untouched cache bytes identical |

## Benchmark

**Baseline:** the recurrent verify kernel (`_kda_recoverssm_verify_kernel`, a
serial token-by-token Triton recurrence) from vLLM PR #52993 — the optimized
"New" version fetched from the PR head (`pull/52993/head`; vLLM main
predates the PR's verify tuning: batch-sensitive BV, `num_warps=1`,
eviction hints).

**Config:** 1× B200 (SM100); Kimi K3 TP=8 shard = 96 total KDA heads / 8 →
H = HV = 12 heads/GPU, K = V = 128; T = 8 tokens/seq (uniform);
`lower_bound = -5` gate with `dt_bias`; beta logits; bf16 state pool;
batch B = 1..512; both kernels fed byte-identical tensors and producing the
full verify contract (out + fp32 corrections + kg).

**Method:** `flashinfer.testing.bench_gpu_time(fn, enable_cupti=True,
dry_run_iters=10, repeat_iters=50)` — per-iteration CUPTI GPU kernel time
with L2 flushed before every iteration (cold cache); median of 50. All
three output artifacts refchecked between the two kernels at every point
before timing. Harness anchored externally: it reproduces the GDN WY
PR #3908 endpoint (132.2 vs 129.7 µs published) and, under hot L2, the
vLLM PR #52993 published B300 verify table within ~4–8%.

| B | flashinfer (µs) | vLLM PR recurrent (µs) | speedup (cold L2) |
|---|---|---|---|
| 1 | 6.94 | 10.46 | 1.51× |
| 2 | 6.96 | 11.07 | 1.59× |
| 4 | 7.25 | 12.05 | 1.66× |
| 8 | 7.39 | 15.52 | 2.10× |
| 16 | 8.42 | 20.03 | 2.38× |
| 32 | 9.63 | 37.12 | 3.85× |
| 64 | 17.63 | 69.30 | 3.93× |
| 128 | 27.14 | 123.07 | 4.54× |
| 256 | 48.11 | 257.86 | 5.36× |
| 512 | 91.74 | 537.95 | 5.86× |

Hot-L2 speedups match (1.40×–5.96×).

### DP / lower-TP configurations

Same setup at larger per-GPU head counts (DP = no TP sharding = all 96 KDA
heads on one GPU), refchecked per point:

| H (TP) | B=1 | B=8 | B=32 | B=128 | B=256 |
|---|---|---|---|---|---|
| 24 (TP=4) | 1.58× | 3.36× | 3.90× | 5.38× | 5.91× |
| 48 (TP=2) | 1.59× | 4.45× | 4.65× | 5.96× | 6.10× |
| 96 (DP) | 2.19× | 4.45× | 5.51× | 6.23× | 6.15× |

(e.g. H=96, B=128: 178.9 µs vs 1114.9 µs.)

## Correctness checking

`pytest tests/kda/test_kda_output_only.py` (39 tests): fp32 recurrent
reference across gate modes / GQA / ragged T / odd batches; RecoverSSM
contract tests (ragged lengths, null and negative slots, fp32 corrections,
raw-k|raw-g kg, checkpoint-pool immutability); cross-check vs
`flashinfer.recurrent_kda`; preallocated-output paths.

For the direct diff against vLLM's Triton kernel, fetch
`vllm/models/kimi_k3/nvidia/ops/recoverssm.py` from `pull/52993/head` and
run it standalone next to `kda_recoverssm_verify` on identical inputs (a
local harness exists but is intentionally not committed, as it depends on
the vLLM source file).

**Limits** (clear errors, no silent fallback): K = V = 128,
`spec_query_len <= 16`, bf16 checkpoint pool, SM90+ (validated on SM100a / B200).
Checkpoint blocks must be dense `[H, V, K]`, but the pool's block stride may
be padded (16 B-aligned) — the stride is a runtime kernel argument, so one
compiled kernel serves any padding (validated vs the vLLM Triton kernel with
128/256/1024-element paddings, padding bytes verified untouched).

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

- **New Features**
  - Added output-only KDA decoding for speculative verification using read-only committed state.
  - Added automatic, WY, and recurrent backend selection.
  - Added optional correction and key/gate cache outputs.
  - Added a RecoverSSM verification API with top-level package access.
  - Added trace support for sequences of up to 16 tokens on supported hardware.

- **Documentation**
  - Documented supported inputs, constraints, backend behavior, usage, and tuning options.

- **Tests**
  - Added coverage for accuracy, backends, ragged sequences, output buffers, optional outputs, and state immutability.

- **Benchmarks**
  - Added configurable GPU latency and dispatch speedup benchmarks.


hnover-nv (commented):
A few high level questions from the PR description before I dive into code:
- The state folding, do you need to do the "anchoring" thing?  Or are we pretty sure the direction that can get unbounded won't?  Because the decay gates are capped?  Probably important to run a comparison here with realistic-but-extreme decay gates, vs random data.
- You say use the recurrent for small batch.  That surprises me.  Small batch you care less so I could see it not mattering there, but I'd expect recurrent to win for small T, not small batch.

ameynaik-hub:
@hnover-nv 
`You say use the recurrent for small batch. That surprises me. Small batch you care less so I could see it not mattering there, but I'd expect recurrent to win for small T, not small batch.`

here is the table but for T=8 which I believe what is used for kimi K3, WY kernel is better. 

# Performance Table

Each cell is **rec µs / wy µs (ratio)**
- ratio = recurrent + WY

| T \ B | 1 | 2 | 4 | 8 | 16 | 32 | 64 | 128 | 256 | 512 |
|---|---|---|---|---|---|---|---|---|---|---|
| 1 | 2.8/4.8 (0.58) | 2.8/4.8 (0.59) | 3.0/4.9 (0.62) | 3.6/5.0 (0.71) | 4.7/5.8 (0.81) | 7.3/6.6 (1.10) | 11.5/8.8 (1.30) | 19.8/15.2 (1.31) | 35.9/26.8 (1.34) | 67.9/47.2 (1.44) |
| 2 | 3.0/4.8 (0.62) | 3.1/4.8 (0.65) | 3.4/4.9 (0.69) | 5.5/5.1 (1.08) | 7.6/5.9 (1.28) | 13.8/6.8 (2.01) | 20.4/9.1 (2.24) | 35.8/15.6 (2.30) | 64.9/27.3 (2.38) | 122.8/48.2 (2.55) |
| 3 | 3.3/4.8 (0.69) | 3.6/4.8 (0.74) | 4.0/5.0 (0.79) | 6.6/5.3 (1.24) | 8.9/6.1 (1.47) | 16.1/7.0 (2.28) | 24.5/9.4 (2.59) | 43.7/16.1 (2.72) | 80.6/28.2 (2.85) | 153.6/49.3 (3.12) |
| 4 | 3.8/4.9 (0.79) | 4.0/4.9 (0.81) | 4.5/5.0 (0.90) | 7.5/5.3 (1.41) | 10.0/6.1 (1.63) | 18.4/7.2 (2.55) | 28.0/9.6 (2.93) | 50.0/16.4 (3.06) | 92.2/28.5 (3.23) | 175.4/50.2 (3.50) |
| 5 | 4.1/4.9 (0.85) | 4.5/4.9 (0.92) | 5.2/5.0 (1.04) | 8.5/5.4 (1.59) | 11.8/6.2 (1.89) | 22.2/7.3 (3.05) | 32.9/9.9 (3.34) | 61.0/16.9 (3.61) | 110.6/29.1 (3.80) | 210.6/51.1 (4.12) |
| 6 | 4.5/4.9 (0.91) | 4.9/4.9 (1.00) | 5.7/5.1 (1.12) | 9.3/5.5 (1.69) | 12.4/6.2 (1.98) | 24.4/7.4 (3.30) | 36.2/10.0 (3.64) | 68.1/17.3 (3.93) | 123.7/29.6 (4.19) | 236.5/51.9 (4.55) |
| 7 | 4.9/4.9 (0.99) | 5.4/4.9 (1.09) | 6.3/5.1 (1.24) | 10.3/5.5 (1.87) | 13.5/6.4 (2.12) | 24.6/7.5 (3.27) | 39.2/10.2 (3.84) | 72.5/17.6 (4.13) | 134.6/30.1 (4.46) | 258.6/53.1 (4.87) |
| 8 | 4.9/4.9 (1.01) | 5.5/5.0 (1.10) | 6.8/5.2 (1.29) | 11.2/5.7 (1.97) | 14.8/6.6 (2.23) | 27.6/7.7 (3.57) | 43.9/10.6 (4.16) | 83.4/17.9 (4.66) | 151.0/31.3 (4.82) | 290.8/54.7 (5.32) |

hnover-nv:
https://github.com/flashinfer-ai/flashinfer/pull/4709#issuecomment-5433584925

Ah, ok, it wins at small batch or small T, a whole upper left triangle.  I guess the explanation is: The WY formulation is literally less serial compute time even for T=1 as it's output only (skip the outer products completely, tensor cores are fast).  So at large batch, WY wins even for small T.  But when BS is also small, you aren't saturating the GPU, and the recurrence has value slicing so even though it's more work you win on parallelism.  The bigger T, the more extra work and so the smaller the batch you need to have for the parallelism to win.
Theoretically you could value slice for WY too?  You'd end up duplicating the Householder parts (K dot K, triangle solve)  but you'd slice everything you dot against S.  _Not_ worth doing, these numbers are great as-is, just mentioning.

kahyunnam (commented):
There’s an ongoing effort to keep FlashInfer’s public surfaces unified rather than adding a new entrypoint per kernel or per caller (vLLM). This looks like the same KDA decode as recurrent_kda (already supports packed cu_seqlens, initial_state_source, and a speculative window) with a frozen/verify contract, not a separate op. I’d rather explore a mode on recurrent_kda (e.g. no state write, WY dispatch, optional correction/kg outs) than ship kda_output_only_decode and kda_recoverssm_verify as new top-level APIs. (vLLM can adapt to that; we shouldn’t add a drop-in just to match their current signature)

flashinfer/kda_decode.py
```diff
@@ -363,3 +389,108 @@ def fused_kda_decode(
...
+            ``use_gate_in_kernel=True``.
+        dt_bias: ``[H*128]`` float32 per-channel decay bias (optional).
+        scale: Query scale; defaults to ``128**-0.5``.
+        use_gate_in_kernel: Compute the gate in-kernel from raw ``g``.
+        lower_bound: If set (negative, e.g. ``-5.0`` for Kimi K3), uses the
+            ``lower_bound * sigmoid(exp(A_log) * (g + dt_bias))`` gate;
+            otherwise the ``-exp(A_log) * softplus(g + dt_bias)`` gate. Only
+            meaningful with ``use_gate_in_kernel=True``.
+        beta_is_logit: Apply sigmoid to ``beta`` inside the kernel.
+        output: Optional preallocated contiguous ``[B, T, HV, 128]`` bf16
```
hnover-nv: Otherwise what?

flashinfer/kda_decode.py
```diff
@@ -363,3 +389,108 @@ def fused_kda_decode(
...
+            output (required for allocation-free CUDA-graph replay).
+        backend: ``"auto"`` (default), ``"wy"``, or ``"recurrent"``.
+        emit_corrections: Also emit the speculative-verify caches consumed by
+            a RecoverSSM-style commit kernel: per-token corrections
+            ``U_t = beta_t * (v_t - u_t)`` (with ``beta_t`` post-sigmoid when
+            ``beta_is_logit=True``) of shape ``[B, T, HV, 128]`` and a
+            kg cache ``[B, T, HV, 256]`` holding the L2-normalized key in
+            ``[..., :128]`` and the raw gate in ``[..., 128:]``.
+        corrections_out: Optional preallocated contiguous bf16 corrections
```
hnover-nv: otherwise what, if emit corrections = true?

benchmarks/bench_kda_output_only.py
```diff
@@ -0,0 +1,186 @@
+# Copyright (c) 2026 by FlashInfer team.
+"""Benchmark the output-only KDA decode kernel vs the recurrent baseline.
+
+Compares at Kimi K3 per-GPU shapes (H = HV = 12, K = V = 128):
```
hnover-nv: Clarify TP=8 maybe, it took me a second to figure out why only 12.

benchmarks/bench_kda_output_only.py
```diff
@@ -0,0 +1,186 @@
...
+    )
+    h0 = torch.randn(max(B, 2), HV, V, K, dtype=torch.bfloat16, device=dev) * 0.1
+    idx = torch.arange(B, dtype=torch.int32, device=dev)
+    outb = torch.empty(B, T, HV, V, dtype=torch.bfloat16, device=dev)
+    scale = K**-0.5
+
+    corrb = torch.empty(B, T, HV, V, dtype=torch.bfloat16, device=dev)
+    kgb = torch.empty(B, T, HV, 2 * K, dtype=torch.bfloat16, device=dev)
+
+    def run_oo(be):
```
hnover-nv, line 119: I notice this is the normal kda recurrence stuff, not the vllm replacement.  So that differs in production usage for tthe constexpr's:  lower gate limit and within kernel activations for decay and beta.  Also, the correction factor is half the size (b16 vs fp32).
So I think the reported benchmark figures will be _slightly_ optimisic, givne taht plus the annoying ragged tensor bookkeeping.

flashinfer/kda_kernels/kda_decode_wy_output_only.py
hnover-nv, line 1: File name is no longer accurate, as it has the output-only recurrent kernel as well.

flashinfer/kda_kernels/kda_decode_wy_output_only.py
```diff
@@ -0,0 +1,2476 @@
+"""CuTeDSL KDA (Kimi Delta Attention) MTP decode kernel — WY-parallel,
+OUTPUT-ONLY (frozen state).
+
+Computes the decode output for 1..16 tokens per sequence from a read-only
```
hnover-nv, line 4: mention recurrent fallback here as it's in the file?

flashinfer/kda_kernels/kda_decode_wy_output_only.py
```diff
@@ -0,0 +1,2476 @@
...
+               o_t = S_t q_t
+
+  WY form:     cum_t[k] = sum_{s<=t} g_s[k]        (per-channel log cumsum)
+               khat_t = k_t * exp(cum_t)           qhat_t = q_t * exp(cum_t)
+               ktil_t = k_t * exp(-cum_t)
+               M      = strict_lower(khat @ ktil^T)      # decay folds in
+               Tmat   = (I + diag(beta) M)^{-1} diag(beta)
+               QT     = tril(qhat @ ktil^T) @ Tmat       # incl. diagonal
+               A_full = qhat - QT @ khat
+               O      = A_full @ S0^T + QT @ V
```
hnover-nv, line 43: This is a nice formulation, folding the QT into the Q to minimize matmuls with state!

flashinfer/kda_kernels/kda_decode_wy_output_only.py
```diff
@@ -0,0 +1,2476 @@
...
+double-issue into a SW128-swizzled tile), the QT@V epilogue and the
+SMEM-staged coalesced output flush are inherited unchanged.
+
+Gate modes (compile-time):
+  GATE_PRECOMPUTED (0): g is the log-space gate, used as-is.
+  GATE_LOWER_BOUND (1): g_log = lower_bound * sigmoid(exp(A_log) * (g + dt_bias))
+                        (the Kimi K3 contract; lower_bound < 0, e.g. -5).
+  GATE_SOFTPLUS    (2): g_log = -exp(A_log) * softplus(g + dt_bias)
+                        (the original Kimi-Linear gate).
+
```
hnover-nv, line 70: mention the correction stuff up here?  Maybe not necessary, I'm not sure which descriptions are more canonical to come to.

flashinfer/kda_kernels/kda_decode_wy_output_only.py
```diff
@@ -0,0 +1,2476 @@
...
+        # serving pools (e.g. vLLM's KDA cache) pad between blocks. The TMA
+        # tile modes must be static, so the incoming (possibly fully
+        # dynamic) descriptor layout is not used directly. Reorder to
+        # (V, K, HV, pool) so the per-CTA tile is (V_DIM_C, K_HALF); the
+        # trailing (HV, pool) modes survive tma_partition as outer coords.
+        _h0_blk = cute.assume(s_h0_blk, divby=8)  # 16 B alignment (bf16)
+        gH0_dense = cute.make_tensor(
+            gH0.iterator,
+            cute.make_layout(
+                (gH0.layout.shape[0], HV, V_DIM_C, K_DIM),
```
hnover-nv, line 281: you extracted gH0.layout.shape[0] above as B_val, so could we use that?  do we lose some constexpr or something?  I'm not sure what tensor geometry is baked in here.

flashinfer/kda_kernels/kda_decode_wy_output_only.py
```diff
@@ -0,0 +1,2476 @@
...
+        so_hv = V_DIM
+        so_t = HV * V_DIM
+        so_b = _nv * so_t
+        skg_hv = 2 * K_DIM
+        skg_t = HV * 2 * K_DIM
+        skg_b = _nv * skg_t
+
+        tidx, _, _ = cute.arch.thread_idx()
+        _pid_vt, pid_hv, pid_b = cute.arch.block_idx()
+        lane_id = tidx & 31
```
hnover-nv, line 404: Total nit:  apparently there exist cutlass.cute.arch.lane_idx and cutlass.cute.arch.warp_idx to get these, and they may even be faster (lane_idx is a dedicated register)

flashinfer/kda_kernels/kda_decode_wy_output_only.py
```diff
@@ -0,0 +1,2476 @@
...
+        _exp_A = f32(0.0)
+        _dtb = f32(0.0)
+        if const_expr(self._gate_mode != GATE_PRECOMPUTED):
+            _exp_A = _exp2_approx_f32(gAlog.iterator[i_h].to(f32) * f32(LOG2_E))
+            if const_expr(self._has_dt_bias):
+                _dtb = gDtbias.iterator[i_h * K_DIM + tidx].to(f32)
+
+        smem = utils.SmemAllocator()
+
+        _WSTAGE_N = (T * V_PADDED) if self._dropin else 8
```
hnover-nv: If non-dropin in fsater than dropin, is it partly due to occupancy due to shared memory?  If so, coudl we re-use h_buf (possibly allocated larger if necessary) vs this new memory allocation, when emitting corrections?

flashinfer/kda_kernels/kda_decode_wy_output_only.py
```diff
@@ -0,0 +1,2476 @@
...
+                    _flush_rows_w = 16 if self._t_input > 8 else 8
+                    if const_expr(self._n_valid >= _flush_rows_w):
+                        _st_global_v4_b32(_gCorr_base, _fl_off, _v0, _v1, _v2, _v3)
+                    else:
+                        if _fl_row < Int32(self._n_valid):
+                            _st_global_v4_b32(_gCorr_base, _fl_off, _v0, _v1, _v2, _v3)
+
+
+# ============================================================================
+# Small-batch grouped OUTPUT-ONLY kernel.
```
hnover-nv: small batch and small t, right?

flashinfer/kda_kernels/kda_decode_wy_output_only.py
```diff
@@ -0,0 +1,2476 @@
...
+        output: optional preallocated ``[B, T, HV, 128]`` bf16 output.
+
+    Returns:
+        ``[B, T, HV, 128]`` bf16 output.
+
+    Notes:
+        Q/K L2 normalization is always applied in-kernel (eps 1e-6), matching
+        ``recurrent_kda(use_qk_l2norm_in_kernel=True)``.
+    """
+    assert q.dtype == torch.bfloat16, "q must be bf16"
```
hnover-nv: Is there more shape validation to do, that everythng that should match across tensors does match?

hnover-nv (commented):
Very nice!  I saw you pushed a little during my review, but I need to log off now and wanted to get you this asap.  So sorry if some is obsolete.

flashinfer/kda_decode.py
ameynaik-hub: Thanks @hnover-nv for detailed reviews. Appreciate it a lot. Addressing them 1 by 1.

ameynaik-hub:
> There’s an ongoing effort to keep FlashInfer’s public surfaces unified rather than adding a new entrypoint per kernel or per caller (vLLM). This looks like the same KDA decode as recurrent_kda (already supports packed cu_seqlens, initial_state_source, and a speculative window) with a frozen/verify contract, not a separate op. I’d rather explore a mode on recurrent_kda (e.g. no state write, WY dispatch, optional correction/kg outs) than ship kda_output_only_decode and kda_recoverssm_verify as new top-level APIs. (vLLM can adapt to that; we shouldn’t add a drop-in just to match their current signature)

@kahyunnam 

Done.
  - No new public APIs. kda_output_only_decode and kda_recoverssm_verify are removed from the top level.
  - The frozen/verify behavior is now a mode on recurrent_kda, following the GDN precedent (gated_delta_rule_mtp's disable_state_update flag and its slot-indexed intermediate_states_buffer):
  
```
  out, _ = flashinfer.recurrent_kda( 
      q, k, v, g, beta, A_log=A_log, dt_bias=dt_bias,
      lower_bound=-5.0, use_gate_in_kernel=True,
      cu_seqlens=seq_starts, initial_state_source=state,
      initial_state_indices=slot_ids, 
      disable_state_update=True,      # frozen/verify: state never written
      correction_cache=corr_cache,    # optional spec-decode commit caches
      kg_cache=kg_cache,
      output=out)
```
  
  - When the flag is set, the op dispatches internally to the WY / grouped output-only kernels. backend="cake" raises in this mode (no frozen-state cake kernels, no silent fallback). Both the batched [B, T, ...]
  form and packed cu_seqlens form (ragged lengths, null slots) work.
  
  On the cache layout question: I went with slot-indexed [slot, head, token, dim] + fp32 corrections, since that's how any commit/recovery consumer addresses these (and it matches GDN's buffer). The kernel computes cache addresses from runtime strides internally, so if you'd rather have a different layout, it's a small wrapper change happy to adjust.

benchmarks/bench_kda_output_only.py
ameynaik-hub: fixed

benchmarks/bench_kda_output_only.py
ameynaik-hub: You're right on both : this mode was measuring an easier problem than production: precomputed gate/beta instead of in-kernel, bf16 batch-major corrections instead of fp32 slot-indexed, and rectangular batches with none of the ragged bookkeeping.

Fixed in 89d53bae.

flashinfer/kda_kernels/kda_decode_wy_output_only.py
ameynaik-hub: fixed.

flashinfer/kda_kernels/kda_decode_wy_output_only.py
ameynaik-hub: fixed.

flashinfer/kda_kernels/kda_decode_wy_output_only.py
ameynaik-hub: Thanks! Its the same formulation as the GDN WY output-only kernel.

flashinfer/kda_kernels/kda_decode_wy_output_only.py
ameynaik-hub: great find! Thanks you! 
I reused h_buf see gains. −19% at B=64, −15/−16% at B=256/512

flashinfer/kda_kernels/kda_decode_wy_output_only.py
ameynaik-hub: Done in f8284afe (+ 25678625 for a clearer GQA error): cross-tensor shape/dtype/device checks at all three entry layers, with negative tests. The wrappers compute raw cp.async offsets from assumed shapes, so anything mismatched now raises with the tensor name and both shapes instead of reading out of bounds.

flashinfer/kda_kernels/kda_decode_wy_output_only.py
ameynaik-hub: They're different quantities, so B_val can't be reused: B_val = gH0idx.shape[0] is the batch size (grid extent — one CTA column per sequence), while gH0.layout.shape[0] is the state-pool slot count (the TMA descriptor's outer extent). Slot indices >= B are legal, so substituting B_val would shrink the TMA box and corrupt loads for high slots. And no constexpr is at stake either way: the pool dim is marked shape-dynamic (mode 0), so both are runtime scalars — the baked-static geometry is the inner (HV, V, K) dims/strides. Added comments at both sites in f8284afe so this doesn't get 'simplified' later.

kahyunnam:
@flashinfer-bot run

kahyunnam (approved):

flashinfer/kda_kernels/kda_decode_wy_output_only.py
```diff
@@ -379,7 +445,12 @@ def kernel(
 
         smem = utils.SmemAllocator()
 
-        _WSTAGE_N = (T * V_PADDED) if self._dropin else 8
+        # The fp32 W (corrections) stage lives inside h_buf (see the emit
+        # epilogue): out-stage region [0, 8704) is dead after the output
+        # flush and the D tile sits at [8704, 13056). Reusing it reclaims
+        # ~8.7 KB/CTA of SMEM (one occupancy step at large batch); this
+        # placeholder stays only to avoid struct churn.
```
hnover-nv, line 452: It's new code, so better to remove?

flashinfer/kda_kernels/kda_decode_wy_output_only.py
hnover-nv, line 479: comment now out of date?  And dead code we can remove anyway?

hnover-nv (approved):

kahyunnam:
@flashinfer-bot run

PR #4714, opened 2026-08-24
feat: Add cute-dsl-prims backend for SM120 FP8 GQA prefill

Tom-Zheng:
## 📌 Description

This PR adds SM120 CuTe DSL PRIMS FMHA prefill support.

Key changes:

- Add FP8 FMHA kernels for SM120.
- Support ragged and paged KV-cache prefill.
- Integrate the `cute-dsl-prims` backend with FlashInfer batch prefill wrappers.
- Support FP16/BF16 outputs, causal attention, LSE output, and CUDA Graph execution.
- Add benchmark integration, documentation, and correctness tests.
- Require `nvidia-cutlass-dsl>=4.7.0` at runtime only when the SM120 backend is used.

## 🔍 Related Issues

None.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Summary by CodeRabbit

- **New Features**
  - Added SM120 FP8 attention prefill for ragged and paged key/value data.
  - Added `cute-dsl-prims` backend support on compatible SM120 hardware.
  - Added support for combined or separate HND-layout key/value caches without copying.
  - Added optional LSE outputs, causal masking, scaling, variable sequence lengths, and programmatic dependent launch support.
  - Added runtime capability detection for SM120 CuTe DSL features.

- **Documentation**
  - Updated benchmark guidance, supported-backend matrices, cache-layout information, and examples.

- **Tests**
  - Added comprehensive correctness and validation coverage for SM120 FP8 ragged and paged prefill.


flashinfer/prefill.py
```diff
@@ -2499,6 +2548,56 @@ def plan(
...
+                window_left=window_left,
+                logits_soft_cap=logits_soft_cap,
+                prefix_len_ptr=prefix_len_ptr,
+                token_pos_in_items_ptr=token_pos_in_items_ptr,
+                max_item_len_ptr=max_item_len_ptr,
+                max_sequence_kv=max_sequence_kv,
+                fixed_split_size=fixed_split_size,
+            )
+            if block_tables is None:
+                block_tables = _build_block_tables_from_paged_kv_indices(
```
rosenrodt, line 2607: [P2] Make the CUDA Graph block-table ownership contract explicit

The realistic serving path is graph-safe when the caller supplies one CUDA `int32` contiguous `block_tables` tensor and updates its contents in place. The [complete hardware-validated test](https://github.com/flashinfer-ai/flashinfer/pull/4714#discussion_r3853182012) supplies the setup for this contrast.

Pseudocode:

```python
wrapper.plan(..., block_tables=block_tables)
block_tables.copy_(next_block_tables)
graph.replay()  # works: stable pointer, updated page mapping

wrapper.plan(..., block_tables=None)
# contract gap: no stable block-table buffer is exposed for replay updates
```

The caller-owned case passes on an RTX 5080 (SM120, CC 12.0) at `004d1aea43208be901a51c37d52805b142358d67`: eager mean `-1.0`, replay mean `-1.0`.

When `block_tables` is omitted, this branch derives a new tensor during every `plan()`, so that convenience path is not suitable for updating an already captured graph. Please either document that CUDA Graph users of `cute-dsl-prims` must retain and update a caller-owned `block_tables` tensor, or add wrapper-owned stable block-table storage. This is an ownership/API-contract gap, not a demonstrated production correctness failure.

_AI-assisted review feedback._

tests/attention/test_sm120_prims_prefill_backend.py
```diff
@@ -0,0 +1,211 @@
...
+    hq, hkv, d = 4, 2, 32
+    qo = torch.tensor([0, 3, 5], dtype=torch.int32, device="cuda")
+    page_indptr = torch.tensor([0, 1, 3], dtype=torch.int32, device="cuda")
+    page_indices = torch.tensor([3, 1, 4], dtype=torch.int32, device="cuda")
+    last_page_len = torch.tensor([5, 4], dtype=torch.int32, device="cuda")
+    q = _fp8((5, hq, d))
+    combined_cache = _fp8((5, 2, hkv, page_size, d))
+    k_pool, v_pool = combined_cache.unbind(dim=1)
+    workspace = torch.empty(16 << 20, dtype=torch.uint8, device="cuda")
+    wrapper = flashinfer.BatchPrefillWithPagedKVCacheWrapper(
```
rosenrodt, line 110: [P2] Test the production CUDA Graph metadata-update path

The relevant serving sequence updates a caller-owned `block_tables` tensor in place rather than calling `plan()` after capture. The existing test covers eager execution but not this graph-stable metadata update. With the capture and tensors defined in the complete test below, the missing contract is:

```python
block_tables.fill_(0)
graph.replay()
torch.testing.assert_close(captured, torch.ones_like(captured))   # works: page 0
block_tables.fill_(1)
graph.replay()
torch.testing.assert_close(captured, -torch.ones_like(captured))  # must work: page 1
```

<details>
<summary>Full minimal validation (pytest)</summary>

```python
import torch

import flashinfer


def test_cuda_graph_reads_updated_caller_block_table():
    device, i32 = "cuda", torch.int32
    workspace = torch.empty(16 << 20, dtype=torch.uint8, device=device)
    wrapper = flashinfer.BatchPrefillWithPagedKVCacheWrapper(
        workspace,
        "HND",
        backend="cute-dsl-prims",
        use_cuda_graph=True,
        qo_indptr_buf=torch.empty(2, dtype=i32, device=device),
        paged_kv_indptr_buf=torch.empty(2, dtype=i32, device=device),
        paged_kv_indices_buf=torch.empty(1, dtype=i32, device=device),
        paged_kv_last_page_len_buf=torch.empty(1, dtype=i32, device=device),
    )

    hq, hkv, head_dim, page_size = 2, 1, 32, 16
    q = torch.ones((1, hq, head_dim), device=device).to(torch.float8_e4m3fn)
    cache = torch.zeros(
        (2, 2, hkv, page_size, head_dim),
        dtype=torch.float8_e4m3fn,
        device=device,
    )
    cache[0, 1].fill_(1)
    cache[1, 1].fill_(-1)
    indptr = torch.tensor([0, 1], dtype=i32, device=device)
    page_indices = torch.tensor([0], dtype=i32, device=device)
    last_page_len = torch.tensor([page_size], dtype=i32, device=device)
    block_tables = torch.tensor([[0]], dtype=i32, device=device)
    wrapper.plan(
        indptr,
        indptr,
        page_indices,
        last_page_len,
        hq,
        hkv,
        head_dim,
        page_size,
        block_tables=block_tables,
        q_data_type=q.dtype,
        kv_data_type=cache.dtype,
        o_data_type=torch.float16,
    )

    captured = torch.empty((1, hq, head_dim), dtype=torch.float16, device=device)
    for _ in range(3):
        wrapper.run(q, cache, out=captured, enable_pdl=False)
    torch.cuda.synchronize()
    graph = torch.cuda.CUDAGraph()
    with torch.cuda.graph(graph):
        wrapper.run(q, cache, out=captured, enable_pdl=False)

    block_tables.fill_(1)
    eager = torch.empty_like(captured)
    wrapper.run(q, cache, out=eager, enable_pdl=False)
    graph.replay()
    torch.cuda.synchronize()

    torch.testing.assert_close(captured, eager)
    torch.testing.assert_close(eager, torch.full_like(eager, -1))
```

</details>

Reproduced on an RTX 5080 (SM120, CC 12.0) at `004d1aea43208be901a51c37d52805b142358d67`: eager mean `-1.0`, replay mean `-1.0`, pytest `1 passed`.

Please add this case so the supported CUDA Graph ownership/update contract remains covered.

_AI-assisted review feedback._

flashinfer/prefill.py
```diff
@@ -4145,7 +4346,21 @@ def run(
                 q.device,
                 "out",
             )
-        if self._backend == "fmha_v2":
+        if self._backend == "cute-dsl-prims":
```
rosenrodt, line 4407: [P1] Reject stashed attention sinks before the prims return

The serving integration stores attention-sink logits on the wrapper as `_sinks`, as acknowledged by the `fmha_v2` and `cute-dsl` branches later in this method. This new branch returns before consulting that state, so `cute-dsl-prims` silently computes ordinary attention instead of applying or rejecting the requested sinks.

With the tensors and wrapper defined in the complete reproducer below, the failure is:

```python
wrapper._sinks = None
no_sink = wrapper.run(q, k, v)
torch.testing.assert_close(no_sink, torch.ones_like(no_sink))  # works: mean 1
wrapper._sinks = torch.full((hq,), 10.0, device="cuda")
with_sink = wrapper.run(q, k, v)
expected = torch.full_like(with_sink, 0.0007257)
torch.testing.assert_close(with_sink, expected, atol=1e-3, rtol=1e-3)  # fails: mean remains 1
```

Reproduced on an RTX 5080 (SM120, CC 12.0) at `004d1aea43208be901a51c37d52805b142358d67`. Zero Q/K logits and sixteen unit V tokens should produce `16 / (16 + exp(10)) = 0.0007257` after adding the sink to the softmax denominator. Hardware instead returned the no-sink mean `1.0` for both calls; `sink_changed_output` was `False`, all 64 elements mismatched the sink-aware reference, and the process exited 1.

Please either consume `_sinks` in this backend or raise `NotImplementedError` before launching, matching the explicit behavior already used by the other unsupported routes, and add this case as a regression.

<details>
<summary>Lean self-contained SM120 counterexample</summary>

```python
import math
import torch
import flashinfer

device, hq, hkv, dim, n = "cuda", 2, 1, 32, 16
fp8 = torch.float8_e4m3fn
q = torch.zeros((1, hq, dim), device=device).to(fp8)
k = torch.zeros((n, hkv, dim), device=device).to(fp8)
v = torch.ones((n, hkv, dim), device=device).to(fp8)
qo = torch.tensor([0, 1], dtype=torch.int32, device=device)
kv = torch.tensor([0, n], dtype=torch.int32, device=device)
workspace = torch.empty(16 << 20, dtype=torch.uint8, device=device)
wrapper = flashinfer.BatchPrefillWithRaggedKVCacheWrapper(
    workspace, "NHD", backend="cute-dsl-prims"
)
wrapper.plan(
    qo, kv, hq, hkv, dim,
    q_data_type=fp8, kv_data_type=fp8, o_data_type=torch.float16,
)
wrapper._sinks = None
no_sink = wrapper.run(q, k, v)
wrapper._sinks = torch.full((hq,), 10.0, device=device)
with_sink = wrapper.run(q, k, v)
expected = torch.full_like(with_sink, n / (n + math.exp(10.0)))
print(no_sink.float().mean().item(), with_sink.float().mean().item())
print(expected.float().mean().item(), torch.equal(no_sink, with_sink))
torch.testing.assert_close(with_sink, expected, atol=1e-3, rtol=1e-3)
```

</details>

_AI-assisted review feedback._

qsang-nv:
Reviewed at `004d1aea`. Three things I'd like to see addressed before this lands.

**`head_dim=256` is in the supported contract but has zero test coverage.** `_validate_config` accepts `{32, 64, 128, 256}` and `plan_*` pins `q_tile=kv_tile=128`, so `head_dim=256` is the only configuration that selects `kv_pipeline_stages=3` / `uses_distributed_row_state`. `mma_qk`, `online_softmax`, `mma_pv` and `compute_one_kv_tile` are reached exclusively from that branch — roughly 650 lines, including the 3-slot ring in the load warp and the shuffle-based epilogue. The three new test files only cover D=32/64/128. The config is reachable (98352 B vs `get_smem_capacity_in_bytes("sm_120")` = 101376), and the module docstring still states the SMEM bound as `max(2 * kv_tile * head_dim * input_dtype_size, ...)`, which is wrong for that path.

**CUDA Graph is listed as a supported feature but nothing tests it.** No `torch.cuda.CUDAGraph` usage in `test_sm120_fmha.py`, `test_sm120_fmha_paged.py` or `test_sm120_prims_prefill_backend.py`, and the new `is_current_stream_capturing()` guard in `_ensure_kernel` is never exercised.

**`plan()` compiles a PDL variant that the default `run()` never uses.** `plan_ragged`/`plan_paged` hardcode `use_pdl=False`, while `run()`'s `enable_pdl=None` resolves through `device_support_pdl()` to `True` on SM120 (`prefill.py:2894`, `prefill.py:4250`). Every default call therefore discards the plan-time compilation and JITs a second kernel on the first `run()`, which also interacts with the CUDA Graph warm-up requirement above. `griddepcontrol_wait()` and `griddepcontrol_launch_dependents()` are emitted unconditionally, so the two artifacts appear to differ only in the launch attribute.

Also +1 to @rosenrodt's `_sinks` finding.

saltyminty:
No new comments beyond the existing findings. Will await further changes before running CI.

flashinfer/prefill.py
Tom-Zheng: I added path to directly consume caller-provided `block_tables`, same as `trtllm-gen` backend.

flashinfer/prefill.py
Tom-Zheng: Added guard for `_sinks` for ragged run, thanks.

tests/attention/test_sm120_prims_prefill_backend.py
Tom-Zheng: Added CUDA graph tests. Thanks.

Tom-Zheng:
> Reviewed at `004d1aea`. Three things I'd like to see addressed before this lands.
> 
> **`head_dim=256` is in the supported contract but has zero test coverage.** `_validate_config` accepts `{32, 64, 128, 256}` and `plan_*` pins `q_tile=kv_tile=128`, so `head_dim=256` is the only configuration that selects `kv_pipeline_stages=3` / `uses_distributed_row_state`. `mma_qk`, `online_softmax`, `mma_pv` and `compute_one_kv_tile` are reached exclusively from that branch — roughly 650 lines, including the 3-slot ring in the load warp and the shuffle-based epilogue. The three new test files only cover D=32/64/128. The config is reachable (98352 B vs `get_smem_capacity_in_bytes("sm_120")` = 101376), and the module docstring still states the SMEM bound as `max(2 * kv_tile * head_dim * input_dtype_size, ...)`, which is wrong for that path.
> 
> **CUDA Graph is listed as a supported feature but nothing tests it.** No `torch.cuda.CUDAGraph` usage in `test_sm120_fmha.py`, `test_sm120_fmha_paged.py` or `test_sm120_prims_prefill_backend.py`, and the new `is_current_stream_capturing()` guard in `_ensure_kernel` is never exercised.
> 
> **`plan()` compiles a PDL variant that the default `run()` never uses.** `plan_ragged`/`plan_paged` hardcode `use_pdl=False`, while `run()`'s `enable_pdl=None` resolves through `device_support_pdl()` to `True` on SM120 (`prefill.py:2894`, `prefill.py:4250`). Every default call therefore discards the plan-time compilation and JITs a second kernel on the first `run()`, which also interacts with the CUDA Graph warm-up requirement above. `griddepcontrol_wait()` and `griddepcontrol_launch_dependents()` are emitted unconditionally, so the two artifacts appear to differ only in the launch attribute.
> 
> Also +1 to @rosenrodt's `_sinks` finding.

Thanks for the review:

1. `head_dim=256` tests are added.
2. Tests are added for the `is_current_stream_capturing` guard.
3. Now `use_pdl` does not go into compilation, since the kernel binary should be the same.

saltyminty (approved):

PR #4725, opened 2026-08-25
Fix sm_constraint_gemm epilogue reading an uninitialized output buffer

### LINKED ISSUE #4724: sm_constraint_gemm: the epilogue reads the output buffer even when beta == 0, so gemm(a, b) can return NaN
## Before submitting

- [x] I have searched existing issues and did not find a solution or an existing report.

## Bug description

The three Triton GEMMs in `flashinfer/triton/` compute `C = alpha * (A @ B) + beta * C`, and they read `C` on every call. When the caller does not pass `c`, the wrapper allocates it with `torch.empty`, which does not write the memory, so the kernel reads whatever the allocator hands back. `beta = 0.0` does not make that safe — `0.0 * NaN` is `NaN` — so the plain call `gemm(a, b)`, every argument at its documented default, can return an all-NaN matrix with no error and no warning. BLAS treats `C` as write-only when `beta == 0` and never reads its prior contents, which is why callers are entitled to leave it uninitialized.

The load is unconditional here:

https://github.com/flashinfer-ai/flashinfer/blob/083012d6819cf97128e559616b12acb666f2fffe/flashinfer/triton/kernels/sm_constraint_gemm.py#L289-L290

`gemm_kernel_persistent` (`kernels/sm_constraint_gemm.py:128`) and `gemm_kernel_descriptor_persistent` (`:208`, `:209`, `:217`) have the same epilogue, and all three host wrappers allocate the same way (`triton/sm_constraint_gemm.py:64`, `:147`, `:242`), each declaring `c=None, beta=0.0` — so this is the default path, not an edge case. `alpha` and `beta` are runtime arguments declared before the `tl.constexpr` block, so nothing folds the load away at compile time. The kernels were added in #982.

## Minimal reproduction

```python
import torch
from flashinfer.triton.sm_constraint_gemm import gemm

M = N = K = 256
a = torch.randn((M, K), device="cuda", dtype=torch.float16)
b = torch.randn((K, N), device="cuda", dtype=torch.float16)
ref = torch.matmul(a.float(), b.float())
zeros = lambda: torch.zeros((M, N), device="cuda", dtype=torch.float16)

gemm(a, b, c=zeros())  # warm up (compile), so only the buffer differs below

# Leave NaN in a freed block of exactly the output's size and dtype; the caching
# allocator hands that block to the torch.empty inside gemm().
stale = torch.full((M, N), float("nan"), device="cuda", dtype=torch.float16)
del stale

out = gemm(a, b)          # c=None, alpha=1.0, beta=0.0 -- all documented defaults
good = gemm(a, b, c=zeros())
torch.cuda.synchronize()

print(f"gemm(a, b)           NaN cells: {int(torch.isnan(out).sum()):>5}/{out.numel()}")
print(f"gemm(a, b, c=zeros)  NaN cells: {int(torch.isnan(good).sum()):>5}/{good.numel()}"
      f"   max|out-ref|: {(good.float() - ref).abs().max().item():.4g}")
```

```
gemm(a, b)           NaN cells: 65536/65536
gemm(a, b, c=zeros)  NaN cells:     0/65536   max|out-ref|: 0.03097
```

Only the output buffer differs between the two calls. `gemm_persistent` and `gemm_descriptor_persistent` behave identically (65536/65536 NaN for fp16, bf16 and, where supported, fp32). The script poisons the block on purpose to make the failure deterministic; in normal use the same read happens on every call and the result is whatever the previous owner of that memory left behind, which is why this can hide for a long time and then surface as an unexplained NaN.

## Error messages / logs

The read itself is visible without the poisoning trick — `compute-sanitizer --tool initcheck` on a single `gemm(a, b)` call (with `PYTORCH_NO_CUDA_MEMORY_CACHING=1`):

```
========= Uninitialized __global__ memory read of size 16 bytes
=========     at gemm_kernel+0x30a0 in sm_constraint_gemm.py:289
=========     by thread (0,0,0) in block (0,0,0)
=========         Host Frame: gemm in sm_constraint_gemm.py:154
========= ERROR SUMMARY: 2048 errors
```

## Environment

<details><summary><code>python -m flashinfer.collect_env</code> (trimmed to the relevant sections)</summary>

```
==== FlashInfer ====
flashinfer                   : 0.6.18
flashinfer commit            : 083012d6819cf97128e559616b12acb666f2fffe
Target CUDA archs (resolved) : {(10, '0a')}

==== Python / Platform ====
Python            : 3.12.14
Platform          : Linux-5.14.0-570.60.1+21.1.el9_6_ciq.x86_64-x86_64-with-glibc2.34
OS                : Rocky Linux 9.6 (Blue Onyx)

==== GPU / Driver ====
Driver version       : 595.71.05
GPU 0 (CUDA order)   : NVIDIA B200 | SM100 | 148 SMs | 178.3 GiB

==== CUDA Toolkit ====
nvcc version                            : release 13.1, V13.1.115

==== PyTorch ====
torch                          : 2.13.0+cu130
torch.version.cuda             : 13.0

==== Relevant Packages ====
triton                       : 3.7.1
```

</details>

truong-v:
## 📌 Description

The epilogue of all three Triton GEMMs reads `C` unconditionally, but the wrappers allocate `C` with `torch.empty` when the caller passes none. With the documented defaults (`c=None, beta=0.0`) the kernel therefore folds uninitialized memory into the result, and `0.0 * NaN` is `NaN`, so `gemm(a, b)` can return an all-NaN matrix.

- Kernels: read `C` only when `beta != 0`, at all four epilogue sites (`gemm_kernel:289`, `gemm_kernel_persistent:128`, and both branches of `gemm_kernel_descriptor_persistent`). `beta` is uniform across the grid, so this is a scalar branch. When `beta == 0` the result is `alpha * (A @ B)`, which is what the epilogue already computed.
- Wrappers: `assert c is not None or beta == 0`. Asking for `beta != 0` without supplying `C` has no defined meaning, and the internal `torch.empty` would otherwise be read as if it were `C`.

Results for `beta != 0` are unchanged — bit-identical on all three wrappers before and after (fp16, 512³, `alpha=1.5, beta=0.5`).

Performance, 4096³ fp16 on a B200, `do_bench(warmup=100, rep=500)`, caller-provided `c` on both sides so only the kernel differs (two independent rounds agreed to within 0.1 µs):

| kernel | beta | before | after |
|---|---|---|---|
| `gemm_kernel` (the `# only for testing` path) | 0.0 | 154.5 µs | 149.3 µs |
| `gemm_kernel` | 0.5 | 154.8 µs | 158.4 µs |
| `gemm_kernel_persistent` | 0.0 | 531.7 µs | 463.9 µs |
| `gemm_kernel_persistent` | 0.5 | 532.2 µs | 473.0 µs |
| `gemm_kernel_descriptor_persistent` | 0.0 | 218.7 µs | 210.5 µs |
| `gemm_kernel_descriptor_persistent` | 0.5 | 218.7 µs | 219.3 µs |

Skipping the load pays for itself on the `beta == 0` path everywhere. The persistent kernel also gets faster on the `beta != 0` path; there is a single autotune config, so that is codegen, not config selection. The one cost is `gemm_kernel` with `beta != 0` (+2.5% here, +5.7% at 1024³). A branch-free variant that folds the condition into the load mask (`mask=c_mask & (beta != 0), other=0.0`) removes that cost but also gives up the persistent kernel's gain (529.3 µs / 533.5 µs at 4096³); the branch is used because it applies unchanged to the descriptor path, whose TMA loads take no mask, and because the kernel it costs is the one the file marks as test-only. Happy to switch if you would rather keep the naive path untouched.

## 🔍 Related Issues

Fixes #4724

### ✅ Pre-commit Checks

## 🧪 Tests

`test_sm_constraint_gemm` always passes an explicit `c`, so the default `c=None` path — the one that breaks — was never exercised.

Two cases added to `tests/gemm/test_sm_constraint_gemm.py`, over all three wrappers:

- `test_sm_constraint_gemm_beta_zero_does_not_read_c` — frees a NaN-filled block of the output's exact size and dtype so the caching allocator returns it to the internal `torch.empty`, then requires the result to be finite and bit-identical to the same call with a caller-provided zeroed `c`.
- `test_sm_constraint_gemm_beta_nonzero_reads_c` — `alpha=0.0, beta=0.5` must return exactly `0.5 * c`, so the `beta != 0` branch cannot be dropped.

Both assert exact equality rather than tolerances, so they carry no architecture assumption. On this B200: 24 passed, 3 skipped (fp32 is not supported by the descriptor kernel). Reverting the two kernel/wrapper files while keeping the tests fails all 16 `beta_zero` cases and leaves the 8 `beta_nonzero` ones passing.

`compute-sanitizer --tool initcheck` on one `gemm(a, b)` call reports 2048 uninitialized-read errors at `sm_constraint_gemm.py:289` before the fix and 0 after.

The rest of the file is unaffected but cannot run here: `test_sm_constraint_gemm` skips unless the GPU is Hopper (`compute_capability[0] != 9`), so all 27648 of its cases skip on a B200.

## Reviewer Notes

The `beta != 0` branch needs a separate result variable in the pointer epilogues (`res`): `c` holds the accumulator cast to the output dtype, both branches produce fp32, and Triton rejects a branch that changes a live variable's type.

## Environment

- flashinfer `083012d6819cf97128e559616b12acb666f2fffe` (main), editable install
- 1× NVIDIA B200 (SM100), driver 595.71.05
- torch 2.13.0+cu130, triton 3.7.1, CUDA toolkit 13.1, Python 3.12

## Summary by CodeRabbit

- **Bug Fixes**
  - GEMM operations with `beta = 0` no longer read uninitialized output data, producing reliable results.
  - GEMM operations with nonzero `beta` continue incorporating existing output values correctly.
  - Added validation to require output data when it is needed for nonzero-beta calculations.

- **Tests**
  - Added coverage for beta-zero safety and nonzero-beta accumulation across supported GEMM variants.

### DIFF
--- a/flashinfer/triton/kernels/sm_constraint_gemm.py
+++ b/flashinfer/triton/kernels/sm_constraint_gemm.py
@@ -125,8 +125,12 @@ def gemm_kernel_persistent(
         c_mask = (offs_cm[:, None] < M) & (offs_cn[None, :] < N)
         c = accumulator.to(c_ptr.dtype.element_ty)
 
-        c = tl.fma(c, alpha, beta * tl.load(c_ptrs, mask=c_mask))
-        tl.store(c_ptrs, c, mask=c_mask)
+        # C is write-only when beta == 0, so it must not be read there.
+        if beta != 0:
+            res = tl.fma(c, alpha, beta * tl.load(c_ptrs, mask=c_mask))
+        else:
+            res = c * alpha
+        tl.store(c_ptrs, res, mask=c_mask)
 
 
 @triton.jit(launch_metadata=_matmul_launch_metadata)
@@ -205,18 +209,29 @@ def gemm_kernel_descriptor_persistent(
             acc = tl.reshape(accumulator, (BLOCK_SIZE_M, 2, BLOCK_SIZE_N // 2))
             acc = tl.permute(acc, (0, 2, 1))
             acc0, acc1 = tl.split(acc)
-            acc0 = tl.fma(acc0, alpha, beta * c_desc.load([offs_cm, offs_cn]))
-            acc1 = tl.fma(
-                acc1, alpha, beta * c_desc.load([offs_cm, offs_cn + BLOCK_SIZE_N // 2])
-            )
+            # C is write-only when beta == 0, so it must not be read there.
+            if beta != 0:
+                acc0 = tl.fma(acc0, alpha, beta * c_desc.load([offs_cm, offs_cn]))
+                acc1 = tl.fma(
+                    acc1,
+                    alpha,
+                    beta * c_desc.load([offs_cm, offs_cn + BLOCK_SIZE_N // 2]),
+                )
+            else:
+                acc0 = acc0 * alpha
+                acc1 = acc1 * alpha
             c0 = acc0.to(dtype)
             c_desc.store([offs_cm, offs_cn], c0)
             c1 = acc1.to(dtype)
             c_desc.store([offs_cm, offs_cn + BLOCK_SIZE_N // 2], c1)
         else:
-            accumulator = tl.fma(
-                accumulator, alpha, beta * c_desc.load([offs_cm, offs_cn])
-            )
+            # C is write-only when beta == 0, so it must not be read there.
+            if beta != 0:
+                accumulator = tl.fma(
+                    accumulator, alpha, beta * c_desc.load([offs_cm, offs_cn])
+                )
+            else:
+                accumulator = accumulator * alpha
             c = accumulator.to(dtype)
             c_desc.store([offs_cm, offs_cn], c)
 
@@ -286,5 +301,9 @@ def gemm_kernel(
     offs_cn = pid_n * BLOCK_SIZE_N + tl.arange(0, BLOCK_SIZE_N)
     c_ptrs = c_ptr + stride_cm * offs_cm[:, None] + stride_cn * offs_cn[None, :]
     c_mask = (offs_cm[:, None] < M) & (offs_cn[None, :] < N)
-    c = tl.fma(c, alpha, beta * tl.load(c_ptrs, mask=c_mask))
-    tl.store(c_ptrs, c, mask=c_mask)
+    # C is write-only when beta == 0, so it must not be read there.
+    if beta != 0:
+        res = tl.fma(c, alpha, beta * tl.load(c_ptrs, mask=c_mask))
+    else:
+        res = c * alpha
+    tl.store(c_ptrs, res, mask=c_mask)
--- a/flashinfer/triton/sm_constraint_gemm.py
+++ b/flashinfer/triton/sm_constraint_gemm.py
@@ -60,6 +60,9 @@ def gemm_persistent(a, b, c=None, alpha=1.0, beta=0.0, out_dtype=None, num_sms=N
         "Incompatible dtypes between c and out_dtype"
     )
 
+    # The kernel reads c only when beta != 0, and torch.empty does not initialize it.
+    assert c is not None or beta == 0, "c must be provided when beta != 0"
+
     # Allocates output.
     c = torch.empty((M, N), device=a.device, dtype=out_dtype) if c is None else c
 
@@ -142,6 +145,9 @@ def gemm(a, b, c=None, alpha=1.0, beta=0.0, out_dtype=None):
     assert c is None or c.dtype == out_dtype, (
         "Incompatible dtypes between c and out_dtype"
     )
+
+    # The kernel reads c only when beta != 0, and torch.empty does not initialize it.
+    assert c is not None or beta == 0, "c must be provided when beta != 0"
 
     # Allocates output.
     c = torch.empty((M, N), device=a.device, dtype=out_dtype) if c is None else c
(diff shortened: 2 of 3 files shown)


bkryu:
@flashinfer-bot run

bkryu (commented):
Hi @truong-v, thanks for contributing this PR. I am finding that the change here is causing a failure on RTX PRO 6000 
```
  - tests/gemm/test_sm_constraint_gemm.py::test_sm_constraint_gemm_beta_nonzero_reads_c[dtype0-descriptor_persistent]
    E   triton.runtime.errors.OutOfResources: out of resource: shared memory, Required: 131096, Hardware limit: 101376. Reducing block sizes or `num_stages` may help.
    log: /tmp/junit/shards/shard-0000/units/unit-8219533420cb809a/batches/batch-d7a2a09570815fbd.log
  - tests/gemm/test_sm_constraint_gemm.py::test_sm_constraint_gemm_beta_nonzero_reads_c[dtype1-descriptor_persistent]
    E   triton.runtime.errors.OutOfResources: out of resource: shared memory, Required: 131096, Hardware limit: 101376. Reducing block sizes or `num_stages` may help.
    log: /tmp/junit/shards/shard-0000/units/unit-8219533420cb809a/batches/batch-d7a2a09570815fbd.log
  - tests/gemm/test_sm_constraint_gemm.py::test_sm_constraint_gemm_beta_nonzero_reads_c[dtype2-naive]
    E   triton.runtime.errors.OutOfResources: out of resource: shared memory, Required: 131072, Hardware limit: 101376. Reducing block sizes or `num_stages` may help.
    log: /tmp/junit/shards/shard-0000/units/unit-8219533420cb809a/batches/batch-d7a2a09570815fbd.log
```
i.e. on SM120 with ~100KB shared memory,  trying to allocate 128KB is causing issues. Is this something you can fix in the PR?

PR #4734, opened 2026-08-25
feat(gemm): add SM10x contiguous grouped FP8 CuTe backend

### LINKED ISSUE #3712: [Feature] Performant Blockwise 128x128 FP8 MoE Group Gemms for SM10x GPUs
[Feature] Performant Blockwise 128x128 FP8 MoE Group Gemms for SM10x GPUs

Currently, SGLang is using DeepGemm kernels for Blockwise 128x128 FP8 MoE Group Gemms for SM10x GPUs, but the simulation shows that the kernels only achieve ~50% of the projection, and even just 30-50% in the Prefill phase (in 8k1k TP mapping).

The request is to have new Blockwise 128x128 FP8 MoE Group Gemms (or MoE) implementations that can push the performance to 80% of the projection.

Expected perf gain is: +14% for Ctx iter perf and +5% e2e perf for Qwen 3.5 on SM10x GPUs for TP setting.

My current guess is that DeepGemm kernels are more optimized for EP (and wideEP) settings but less optimized for TP settings. But ideally we want an implementation that is not slower than DeepGemm kernels for EP/wideEP setting, while being faster than DeepGemm kernels for TP setting.

The desired backend is cute-dsl, but it is also okay if this must be implemented with trtllm-gen backend.

samodi-nv:
## 📌 Description

Adds `flashinfer.gemm.group_gemm_fp8_nt_groupwise_contiguous`, a standalone CuTe-DSL function for grouped FP8 GEMM on SM100 and SM103.

The existing `group_deepgemm_fp8_nt_groupwise` API is unchanged. The new function does not take a `backend` argument.

### Behavior

- Accepts FP8 E4M3 inputs with per-row 128-element scales for A and 128×128 block scales for B.
- Returns bfloat16 output and supports a caller-provided `out` tensor.
- Requires positive N and K divisible by 128. M can have a partial final tile or be zero.
- Requires sorted, in-range expert indices with 128-row-aligned internal expert boundaries. `-1` padding is unsupported.
- Provides optional `validate_indices=True` checks. These synchronize with the CPU and must run outside CUDA graph capture. With validation disabled, callers must satisfy the documented index requirements.
- Uses the input tensor’s CUDA device and current PyTorch stream.
- Reuses compiled kernels across token counts, keeping at most two variants per device and weight shape. Concurrent calls share one compilation.

Includes API documentation, trace support, generated trace examples, regression tests, and a benchmark script.

### B200 performance

The original comparison below uses a caller-preallocated output for both implementations, so output allocation is excluded. Measurements are the arithmetic mean of two complete runs with 10 warmups and 50 timed iterations per workload. Device-active timing gives a 1.659x weighted geomean, close to the 1.676x end-to-end kernel-call result.

Original benchmark environment: B200 at a stable 1500 MHz SM clock, CUDA 13.2, PyTorch 2.13.0+cu132, CUTLASS/CuTe DSL 4.6.1, and `sgl-deep-gemm==0.1.5.post3`. Each expert has 256 consecutive rows.

| Workload | Shape `(M, N, K, experts)` | DeepGEMM | CuTe DSL | Speedup |
|---|---:|---:|---:|---:|
| TP8 gate/up | `(131072, 256, 4096, 512)` | 268.184 us | 195.192 us | 1.374x |
| TP8 down | `(131072, 4096, 128, 512)` | 564.415 us | 229.384 us | 2.461x |
| EP8 gate/up | `(16384, 2048, 4096, 64)` | 194.672 us | 159.224 us | 1.223x |
| EP8 down | `(16384, 4096, 1024, 64)` | 125.016 us | 96.312 us | 1.298x |
| WideEP32 gate/up | `(4096, 2048, 4096, 16)` | 74.440 us | 52.744 us | 1.411x |
| WideEP32 down | `(4096, 4096, 1024, 16)` | 52.632 us | 31.288 us | 1.682x |
| **Weighted geomean** | | **246.389 us** | **146.987 us** | **1.676x** |

### Current validation

Additional validation used production B200 through CudaGym 2.4.32 with CuTe DSL 4.7.0 and PyTorch 2.13.0+cu132.

The standalone API passed all six production workloads. Compared with the previous CuTe API, geometric-mean GPU latency changed by approximately +0.22%. The device kernel and compile-cache implementation were unchanged by the API split.

The cache fix reduced calls with new token counts from 309–315 ms to 0.28–0.42 ms once the corresponding aligned/tail variant was compiled. The first call for each variant still incurs compilation.

The original DeepGEMM comparison above has not been remeasured on this newer software stack.

### Reproduction

    python benchmarks/bench_grouped_fp8.py --cache-probe
    python benchmarks/bench_grouped_fp8.py --production-shapes --backends cute_dsl

## 🔍 Related Issues

Closes #3712.

### ✅ Pre-commit Checks

- [x] Pre-commit checks run through `uvx`.
- [ ] Pre-commit hooks installed locally.
- [x] All changed files pass pre-commit, including mypy and Ruff.

## 🧪 Tests

- [x] 79 focused tests passed on a two-B200 allocation.
- [x] All six production workloads passed correctness checks.
- [x] Tests cover invalid inputs, empty batches, supplied outputs, streams, device selection, CUDA graphs, cache reuse, concurrent compilation, and trace correctness.
- [x] Confirmed the existing DeepGEMM function and its checker match the PR base.
- [ ] Full repository test suite.

Remote validation used the exact changed sources over the production image’s FlashInfer 0.6.18 supporting dependencies. It was focused source-overlay validation, not a full clean repository installation.


samodi-nv:
@nv-yunzheq, could you please review this SM10x grouped FP8 CuTe backend? The main areas to look at are the public backend contract and the role-scoped barrier handling.

flashinfer/gemm/gemm_base.py
```diff
@@ -8897,19 +8969,29 @@ def group_deepgemm_fp8_nt_groupwise(
     - This function requires NVIDIA Blackwell (SM100) architecture
     - The scaling factors should be generated using appropriate quantization functions
       like ``per_token_cast_to_fp8`` for `a` and ``per_block_cast_to_fp8`` for `b`
-    - The function internally uses the DeepGEMM backend for optimized FP8 computation
+    - The ``cute_dsl`` backend requires M, N, and K to be multiples of 128
```
nvpohanh: Isn't this constraint too strict? gemmM is num_tokens, right? If so, it is not usually multiples of 128

samodi-nv: Ah the constraint was too strict, updated to support any M > 0. Are the other constraints alright?

nvpohanh: yes, N and K can be multiples of 128

nv-yunzheq:
@flashinfer-bot run

flashinfer/gemm/gemm_base.py
```diff
@@ -8915,8 +8916,74 @@ def _check_group_deepgemm_fp8_nt_groupwise_problem_size(
...
+            f"a_scale.shape must be {expected_a_scale_shape}, but got {a_scale.shape}"
+        )
+    if b_scale.shape != expected_b_scale_shape:
+        raise ValueError(
+            f"b_scale.shape must be {expected_b_scale_shape}, but got {b_scale.shape}"
+        )
+    if a_scale.dtype != torch.float32 or b_scale.dtype != torch.float32:
+        raise ValueError("a_scale and b_scale must use torch.float32")
+
+    tensors = [a, b, a_scale, b_scale, m_indices]
```
dhiraj113: **[BUG]** The documented `m_indices` contract for `backend="cute_dsl"` (sorted rows, 128-row-aligned internal expert boundaries, in-range group ids) is validated nowhere; a violating input is accepted without error and rows are silently computed against the wrong expert.

The docstring (`gemm_base.py:9106-9108`) makes sortedness and 128-aligned internal boundaries a hard requirement, and the PR description says the backend "validates the CuTe DSL contract: ... sorted expert rows, and 128-row-aligned internal expert boundaries" — but `_check_group_deepgemm_fp8_nt_groupwise_cute_dsl` only passes `m_indices` through the device/contiguity/16B-alignment loop here (`gemm_base.py:8969-8977`), the common check pins only dtype/numel/contiguity, and the host entry `grouped_gemm_fp8_nt_groupwise_contiguous_sm100` (`grouped_gemm_contiguous_blackwell.py:3413-3455`) just wraps it with `from_dlpack`. The kernel then resolves one group id per 128-row tile, `gidx = gidx_mapping[cur_tile_coord[0] * 128]` (lines 1221/1275/1392), and uses it unchecked as the `l` index into `b`/`b_scale`, so a boundary at row 64 applies expert 0's weights to rows 64-127. Range is equally unchecked: I also probed DeepGEMM's `-1` padding convention — at k=128 the TMA load zero-fills the out-of-range tile so the result does not diverge from deepgemm there, but the `-1` still flows unguarded into the raw-pointer deep-K scale-load path, which TMA clamping does not protect.

To be fair, the default deepgemm backend mis-assigns rows 64-127 identically on this input (it likewise reads the group id once per 128-row block), so this is inherent to the contiguous-tile layout rather than a regression. The defect is the validation claim with no rejection and no negative tests — every new case in `test_groupwise_scaled_gemm_fp8.py:379-398` is happy-path, and the only non-128 count (`[128, 64]`) is deliberately placed last.

Suggested direction: either really validate (sortedness, internal-boundary alignment, `0 <= idx < num_groups` — a fused device-side reduction, optionally behind a debug knob if the D2H sync is unacceptable on the hot path), or correct the PR description and docstring to state that violations are undefined behaviour and call out the `-1` incompatibility with deepgemm's skip convention. Either way, add a rejection (or documented-UB) test so the contract that ships is the one that is asserted.

<details>
<summary>Repro — verified on B200 (sm100), PR head 037a4f5d</summary>

```bash
python - <<"PY"
import torch
from flashinfer.gemm import group_deepgemm_fp8_nt_groupwise
from flashinfer.testing.utils import dequantize_fp8, quantize_fp8

m, n, k, G = 256, 128, 128, 2
torch.manual_seed(0)
a = torch.randn(m, k, device="cuda")
b = torch.randn(G, n, k, device="cuda")
a_fp8, a_scale = quantize_fp8(a, (m, k // 128), (1, 128), "K")
b_fp8, b_scale = quantize_fp8(b, (G, n // 128, k // 128), (1, 128, 128), "K")
a_dq, b_dq = dequantize_fp8(a_fp8, a_scale, "K"), dequantize_fp8(b_fp8, b_scale, "K")

def run(label, idx):
    mi = torch.tensor(idx, device="cuda", dtype=torch.int32)
    ref = torch.einsum("mk,mnk->mn", a_dq, b_dq[mi]).to(torch.bfloat16)  # per-row expert
    out = group_deepgemm_fp8_nt_groupwise(a_fp8, b_fp8, a_scale, b_scale, mi, backend="cute_dsl")
    torch.cuda.synchronize()
    err = (out.float() - ref.float()).abs()
    print(f"{label}: accepted, no error")
    print(f"  max|out-ref|  rows[0:64)={err[:64].max():.3f}  rows[64:128)={err[64:128].max():.3f}"
          f"  rows[128:256)={err[128:].max():.3f}")
    return out

run("control  m_indices=[0]*128+[1]*128 (boundary at row 128)", [0] * 128 + [1] * 128)
out = run("violates m_indices=[0]*64 +[1]*192 (boundary at row 64) ", [0] * 64 + [1] * 192)
expert0 = (a_dq @ b_dq[0].t()).to(torch.bfloat16)  # what "whole tile -> group 0" produces
print(f"  rows[64:128) vs expert 0's output: max|diff| = {(out[64:128].float()-expert0[64:128].float()).abs().max():.3f}")
PY
```

Observed:
```
control  m_indices=[0]*128+[1]*128 (boundary at row 128): accepted, no error
  max|out-ref|  rows[0:64)=0.000  rows[64:128)=0.000  rows[128:256)=0.000
violates m_indices=[0]*64 +[1]*192 (boundary at row 64) : accepted, no error
  max|out-ref|  rows[0:64)=0.000  rows[64:128)=65.125  rows[128:256)=0.000
  rows[64:128) vs expert 0's output: max|diff| = 0.000
```
</details>

flashinfer/gemm/kernels/grouped_gemm_contiguous_blackwell.py
```diff
@@ -0,0 +1,3455 @@
...
+    a_c = _t_fp8(a.unsqueeze(-1))
+    b_c = _t_fp8(b.permute(1, 2, 0))
+    c_c = _t(out.unsqueeze(-1))
+    sfa_c = _t(a_scale.unsqueeze(-1))
+    sfb_c = _t(b_scale.permute(1, 2, 0))
+    gidx_c = from_dlpack(m_indices).mark_layout_dynamic()
+
+    stream = cuda.CUstream(torch.cuda.current_stream(a.device).cuda_stream)
+
+    key = (a.device.index, m, n, k, num_groups)
```
dhiraj113: **[DESIGN]** `_COMPILED` is keyed on the exact `m`, so every distinct routed-row count pays a fresh `cute.compile` (~0.5 s measured) and is retained forever, even though the artifact already compiled for one `m` runs a different `m` correctly.

`key = (a.device.index, m, n, k, num_groups)` at grouped_gemm_contiguous_blackwell.py:3435, but nothing downstream specializes on `m`: `_pick_config` (3324-3373) reads only `n` and `k`, `_build` consumes `m` solely through the `(m % 128 == 0) and (n % 128 == 0)` flag at 3408, and every operand goes through `mark_layout_dynamic` (3314-3321, 3431), so shapes are runtime values inside the compiled function. For the contiguous grouped layout `m` is the total (padded) routed-token count, which changes from step to step in MoE serving, so in practice this is one host-side DSL compile per new batch shape on the hot path plus unbounded growth of a module-level dict (no eviction, no lock around get/compile/set, and no on-disk persistence, so a new process recompiles even repeated shapes). For fairness: the default `deepgemm` backend does not share this behavior, since with its default `compiled_dims="nk"` (deep_gemm.py:1442) `M` is a runtime argument (deep_gemm.py:866) and `m` reaches the cubin-hash key only via the bounded tile-config table; the masked CuTe-DSL sibling (grouped_gemm_masked_blackwell.py:3159) does key on `m`, but there `m` is the fixed padded max per group, not a per-step count.

Suggested direction: key on the inputs that actually reach the compile, i.e. `(a.device.index, n, k, m % 128 == 0)` (plus `num_groups` only if the kernel ever specializes on it), which caps the cache at two artifacts per weight shape; a lock around the compile and, longer term, the repo's `JitSpecCuteDsl` / `build_and_load_cute_dsl_kernel` disk cache would round it out.

<details>
<summary>Repro — verified on B200 (sm100), PR head 037a4f5d</summary>

```bash
python - <<"PY"
import time, torch
from flashinfer.gemm import group_deepgemm_fp8_nt_groupwise
import flashinfer.gemm.kernels.grouped_gemm_contiguous_blackwell as mod

torch.manual_seed(0)
n = k = 128; G = 2  # two experts, boundary at row 128, unit scales -> ref is a plain fp8 matmul

def inputs(m):
    a = torch.randn(m, k, device="cuda").to(torch.float8_e4m3fn)
    b = torch.randn(G, n, k, device="cuda").to(torch.float8_e4m3fn)
    sa, sb = torch.ones(m, 1, device="cuda"), torch.ones(G, 1, 1, device="cuda")
    mi = torch.repeat_interleave(torch.arange(G, device="cuda", dtype=torch.int32),
                                 torch.tensor([128, m - 128], device="cuda"))
    ref = torch.cat([a[:128].float() @ b[0].float().t(), a[128:].float() @ b[1].float().t()]).to(torch.bfloat16)
    return a, b, sa, sb, mi, ref

def call(tag, x):
    a, b, sa, sb, mi, ref = x
    torch.cuda.synchronize(); t0 = time.perf_counter()
    out = group_deepgemm_fp8_nt_groupwise(a, b, sa, sb, mi, backend="cute_dsl")
    torch.cuda.synchronize()
    print(f"{tag}: {1e3 * (time.perf_counter() - t0):7.1f} ms  max|err|={(out.float() - ref.float()).abs().max().item():.4f}  len(_COMPILED)={len(mod._COMPILED)}")

x256, x384, x512 = inputs(256), inputs(384), inputs(512)  # all m % 128 == 0 -> identical compile flag
call("m=256 cold  ", x256); call("m=256 repeat", x256)
call("m=384 new m ", x384); call("m=384 repeat", x384)
call("m=512 new m ", x512)
print("_COMPILED keys:", sorted(mod._COMPILED))

# Is the recompile needed? Run the m=256 artifact on m=384 tensors (all layouts are mark_layout_dynamic).
fn = next(v for key, v in mod._COMPILED.items() if key[1] == 256)
a, b, sa, sb, mi, ref = x384
out = torch.empty(384, n, device="cuda", dtype=torch.bfloat16)
fn(mod._t_fp8(a.unsqueeze(-1)), mod._t_fp8(b.permute(1, 2, 0)), mod._t(out.unsqueeze(-1)),
   mod._t(sa.unsqueeze(-1)), mod._t(sb.permute(1, 2, 0)), mod.from_dlpack(mi).mark_layout_dynamic(),
   mod.cuda.CUstream(torch.cuda.current_stream().cuda_stream))
torch.cuda.synchronize()
print(f"m=256 artifact on m=384 inputs: max|err|={(out.float() - ref.float()).abs().max().item():.4f}")
PY
```

Observed (the first call also includes one-time DSL initialization; the `m=384` / `m=512` lines are the steady-state cost of each new `m`):
```
m=256 cold  :   701.9 ms  max|err|=0.0000  len(_COMPILED)=1
m=256 repeat:     0.5 ms  max|err|=0.0000  len(_COMPILED)=1
m=384 new m :   522.8 ms  max|err|=0.0000  len(_COMPILED)=2
m=384 repeat:     0.3 ms  max|err|=0.0000  len(_COMPILED)=2
m=512 new m :   510.7 ms  max|err|=0.0000  len(_COMPILED)=3
_COMPILED keys: [(0, 256, 128, 128, 2), (0, 384, 128, 128, 2), (0, 512, 128, 128, 2)]
m=256 artifact on m=384 inputs: max|err|=0.0000
```
</details>

flashinfer/gemm/gemm_base.py
```diff
@@ -8915,8 +8916,74 @@ def _check_group_deepgemm_fp8_nt_groupwise_problem_size(
...
+    a_scale: torch.Tensor,
+    b_scale: torch.Tensor,
+    m_indices: torch.Tensor,
+    scale_granularity_mnk: Tuple[int, int, int] = (1, 128, 128),
+    out: Optional[torch.Tensor] = None,
+    out_dtype: Optional[torch.dtype] = None,
+    backend: Literal["deepgemm", "cute_dsl"] = "cute_dsl",
+) -> bool:
+    if not CUTE_DSL_AVAILABLE:
+        raise RuntimeError("The cute_dsl backend requires nvidia-cutlass-dsl")
```
dhiraj113: **[DESIGN]** The DSL-availability gate raises `RuntimeError` where this file's backend gates raise `ValueError`, and `backend="auto"` is reachable but has no working path for this API.

`gemm_base.py:8936-8937` raises `RuntimeError("The cute_dsl backend requires nvidia-cutlass-dsl")`, while every other rejection in this same gate, and the sibling `_cute_dsl_bmm_fp8_requirement` at `gemm_base.py:7410-7413` under the identical `CUTE_DSL_AVAILABLE` condition, raise `ValueError`. The convention is documented one call below at `_check_cute_dsl_arch` (`gemm_base.py:5263-5267`): `suitable_auto_backends` (`utils.py:1366-1372`) catches only `ValueError` while probing, so any other type escapes the probe loop and fails the whole call instead of deselecting the backend (the same mismatch that was fixed on #4685). It is reachable today because the `Literal` at `gemm_base.py:8999` is not enforced: `backend="auto"` goes straight to `suitable_auto_backends` (`utils.py:1424`), where with the DSL absent the `RuntimeError` escapes even though `deepgemm` had already been accepted as suitable, and with the DSL present the call dies on `assert heuristic_func is not None` (`utils.py:1374`) since no heuristic is registered here. To be fair, this is not a regression (`backend=` did not exist before this PR) and an explicit `backend="cute_dsl"` still gets a clear message; the issue is the convention break and that any future `auto` support inherits it.

Suggested direction: make the raise a `ValueError` (one word), and decide what `"auto"` means for this API, either register a `heuristic_func` (e.g. prefer `cute_dsl` when suitable on SM100/SM103, else `deepgemm`) or reject `"auto"` up front with a clear `BackendSupportedError` rather than an `AssertionError`.

<details>
<summary>Repro — verified on B200 (sm100), PR head 037a4f5d</summary>

Run from the checkout root. No kernel is compiled or launched; every case raises inside a backend gate. The DSL-absent cases are simulated by patching the module flag the gate reads.

```bash
python - <<"PY"
import torch
from flashinfer.gemm import group_deepgemm_fp8_nt_groupwise

# Minimal inputs that pass the common check; every case below raises inside a
# backend gate, before any kernel is compiled or launched.
m = n = k = 128
a = torch.randn(m, k, device="cuda").to(torch.float8_e4m3fn)
b = torch.randn(1, n, k, device="cuda").to(torch.float8_e4m3fn)
a_scale = torch.rand(m, k // 128, device="cuda")
b_scale = torch.rand(1, n // 128, k // 128, device="cuda")
m_indices = torch.zeros(m, dtype=torch.int32, device="cuda")

def call(backend):
    try:
        group_deepgemm_fp8_nt_groupwise(a, b, a_scale, b_scale, m_indices, backend=backend)
        return "ok"
    except BaseException as e:
        return f"{type(e).__name__}: {e}"

def sibling_gate():
    try:
        gb._cute_dsl_bmm_fp8_requirement(None, None, None, None, torch.bfloat16)
        return "ok"
    except BaseException as e:
        return f"{type(e).__name__}: {e}"

orig = gb.CUTE_DSL_AVAILABLE
gb.CUTE_DSL_AVAILABLE = False  # simulate nvidia-cutlass-dsl not installed
print("[DSL absent ] backend='cute_dsl' ->", call("cute_dsl"))
print("[DSL absent ] backend='auto'     ->", call("auto"))
print("[DSL absent ] sibling _cute_dsl_bmm_fp8_requirement ->", sibling_gate())
gb.CUTE_DSL_AVAILABLE = orig
print(f"[DSL installed={orig}] backend='auto' ->", call("auto"))
PY
```

Observed:
```
[DSL absent ] backend='cute_dsl' -> RuntimeError: The cute_dsl backend requires nvidia-cutlass-dsl
[DSL absent ] backend='auto'     -> RuntimeError: The cute_dsl backend requires nvidia-cutlass-dsl
[DSL absent ] sibling _cute_dsl_bmm_fp8_requirement -> ValueError: CuTe-DSL is not available. Please install cutlass with cute support.
[DSL installed=True] backend='auto' -> AssertionError: Heuristic function must be provided
```
</details>

flashinfer/gemm/gemm_base.py
```diff
@@ -8915,8 +8916,74 @@ def _check_group_deepgemm_fp8_nt_groupwise_problem_size(
...
+    if scale_granularity_mnk != (1, 128, 128):
+        raise ValueError(
+            "The cute_dsl backend requires scale_granularity_mnk=(1, 128, 128), "
+            f"but got {scale_granularity_mnk}"
+        )
+
+    m, k = a.shape
+    num_groups, n, _ = b.shape
+    if min(m, n, k, num_groups) <= 0:
+        raise ValueError("m, n, k, and the number of groups must be positive")
```
dhiraj113: **[DESIGN]** `m == 0` (empty routed batch) is a silent no-op on the default `deepgemm` backend but raises `ValueError` on `cute_dsl`, so the new backend is not a drop-in swap for a routine serving case.

`_check_group_deepgemm_fp8_nt_groupwise_cute_dsl` rejects `min(m, n, k, num_groups) <= 0` at `gemm_base.py:8947-8948`, whereas the deepgemm path (`flashinfer/deep_gemm.py:1456-1457`, `m_grouped_fp8_gemm_nt_contiguous`) explicitly early-returns on `m == 0` and the API hands back the pre-allocated `(0, n)` output. Zero-token steps happen in MoE serving (no tokens routed to this rank's experts in an iteration), so a caller that works today with the default backend has to add an `m == 0` special case before it can switch to `backend="cute_dsl"`. The `cute_dsl` notes in the docstring (`gemm_base.py:9104-9108`) cover the N/K multiple-of-128 and sorted/128-aligned-boundary requirements but not `m > 0`, so nothing signals the divergence. To be fair, the error is loud rather than a silent-corruption issue; the point is the behavioral mismatch between two backends behind one API.

Suggested direction: match deepgemm by treating `m == 0` as a no-op (accept it in the gate, keep the `n`/`k`/`num_groups` positivity checks, and return `out` before dispatching to the kernel), or state the `m > 0` requirement in the backend notes.

<details>
<summary>Repro — verified on B200 (sm100), PR head 037a4f5d</summary>

```bash
python - <<"PY"
import torch
from flashinfer.gemm import group_deepgemm_fp8_nt_groupwise

# Empty routed batch (m == 0): n = k = 128, one expert, unit block scales.
n = k = 128
a = torch.empty(0, k, device="cuda", dtype=torch.float8_e4m3fn)
b = torch.randn(1, n, k, device="cuda").to(torch.float8_e4m3fn)
a_scale = torch.empty(0, k // 128, device="cuda", dtype=torch.float32)
b_scale = torch.ones(1, n // 128, k // 128, device="cuda", dtype=torch.float32)
m_indices = torch.empty(0, device="cuda", dtype=torch.int32)

for backend in ("deepgemm", "cute_dsl"):
    try:
        out = group_deepgemm_fp8_nt_groupwise(
            a, b, a_scale, b_scale, m_indices, backend=backend
        )
        torch.cuda.synchronize()
        print(f"backend={backend!r:10} -> returned out.shape={tuple(out.shape)}")
    except Exception as e:
        print(f"backend={backend!r:10} -> {type(e).__name__}: {e}")
PY
```

Observed:
```
backend='deepgemm' -> returned out.shape=(0, 128)
backend='cute_dsl' -> ValueError: m, n, k, and the number of groups must be positive
```
</details>

dhiraj113 (commented):
Review of the `cute_dsl` backend integration for `group_deepgemm_fp8_nt_groupwise`, scoped to the **FlashInfer integration layer** (gate, dispatch, host-side launch/caching, tests) — the device kernel body itself was deliberately not reviewed here.

Every finding below was verified by **executing a repro against this PR's head (`037a4f5d`) on a B200 (sm100)**; each inline comment carries the exact command and its observed output. This was an AI-assisted review (per `docs/code_review_guidance.md`); all findings survived an adversarial verification pass before posting. Happy-path numerics were correct in every repro we ran — the issues are at the API boundary and in the compile cache, not in the GEMM itself.

1. **[BUG]** The `m_indices` contract (sorted rows, 128-aligned internal expert boundaries) is documented in the docstring and described as validated in this PR's description, but no layer validates it — a mid-tile expert boundary is accepted silently and part of the tile is computed against the wrong expert's weights (`gemm_base.py:8969`). The default `deepgemm` backend mis-assigns identically on that input, so this is not a regression; the defect is the validation claim with no enforcement and no rejection tests.
2. **[DESIGN]** The compile cache is keyed on the exact `m`, so every distinct token count pays a fresh `cute.compile` (~0.5–0.7 s measured, vs ~2 ms on a hit) into an unbounded, lock-free module dict — and the recompile is unnecessary, since an artifact compiled for one `m` runs a different `m` correctly (`grouped_gemm_contiguous_blackwell.py:3435`). On an MoE token path where `m` changes per step this is a real per-step latency and memory hazard.
3. **[DESIGN]** The DSL-availability check raises `RuntimeError`, against this file's post-#4685 `ValueError` convention for backend unsuitability; it is reachable via `backend="auto"` (not runtime-rejected), which escapes `suitable_auto_backends`' ValueError-only probe — and with the DSL present, `"auto"` dies with an `AssertionError` because no heuristic is registered (`gemm_base.py:8937`).
4. **[DESIGN]** `m == 0` silently no-ops on `deepgemm` but raises `ValueError` on `cute_dsl`, so an empty routed batch works on the default backend and crashes on the new one behind the same API (`gemm_base.py:8947`).

Smaller items held back from inline comments to keep this focused (happy to expand if useful): no negative-path tests for the gate's rejection paths, the `out=` path is untested for `cute_dsl`, the side-stream test has a cross-stream read-after-write race on its inputs, `can_implement`/`is_valid_*` are dead on the integration path with a stale CUTLASS-example module docstring, compile/`get_max_active_clusters` run on the current context rather than `a.device`, and the perf table isn't reproducible from in-repo benchmarks (no bench exercises `backend="cute_dsl"`) while the stated bench environment (CuTe DSL 4.6.1) is below the repo's `nvidia-cutlass-dsl` pin.

Nice work on the kernel and the write-up overall — the perf table and partial-M sweep are exactly the kind of evidence reviewers want.

flashinfer/gemm/gemm_base.py
samodi-nv: Added validate_indices=True to check index range, ordering, and 128-row expert boundaries, including rejecting -1. It is optional to avoid a GPU-to-CPU synchronization on every call. The docstring explains that callers must provide valid indices when it is off. Added tests for invalid inputs

flashinfer/gemm/kernels/grouped_gemm_contiguous_blackwell.py
samodi-nv: The cache now reuses kernels across token counts, keeping at most two versions per device and weight shape: aligned and partial-tail. A lock prevents duplicate compilation.

 On production B200, new token counts took 0.28–0.42 ms instead of 309–315 ms once those versions were compiled. GPU timing stayed similar. Added reuse and concurrency tests. The first call still compiles; disk caching is not included

flashinfer/gemm/gemm_base.py
samodi-nv: Missing CuTe DSL now raises ValueError. The CuTe implementation is now a separate function, group_gemm_fp8_nt_groupwise_contiguous, and the existing DeepGEMM API is restored unchanged. Neither accepts a backend argument, so backend="auto" gets a TypeError instead of an internal assertion. Added tests for both cases.

flashinfer/gemm/gemm_base.py
samodi-nv: M=0 now returns an empty (0, N) output without compiling or launching a kernel. A supplied out tensor is returned unchanged. Added tests with and without a supplied output.

nv-yunzheq:
@flashinfer-bot run tests/gemm

nv-yunzheq (approved):

flashinfer/gemm/gemm_base.py
dhiraj113: Verified at ec6ebb7a on a B200: with `validate_indices=True`, a misaligned boundary (`[0]*64+[1]*192`), unsorted rows, `-1` padding, and an out-of-range id are each rejected with a clear `ValueError` (re-ran our original repro adapted to the new API), while the 128-aligned control still matches the reference exactly. The default-off path is now honestly documented as undefined behavior in both the docstring and the PR description, the CUDA-graph-capture refusal is a nice touch, and `test_grouped_cute_dsl_rejects_indices` covers it. Making the sync opt-in is the right call for the hot path. Resolved, thanks.

flashinfer/gemm/kernels/grouped_gemm_contiguous_blackwell.py
dhiraj113: Verified at ec6ebb7a on a B200: after one ~690 ms compile at m=256, new aligned token counts (384/512/640) run in ~0.2 ms with no recompile; the partial-tail class compiles once (m=200) and then hits (m=328); `_COMPILED` holds exactly the two `m % 128` variants, and the double-checked `_COMPILE_LOCK` keeps hits and launches outside the lock. `test_grouped_cute_dsl_reuses_m_specializations` / `_concurrent_cache_miss` pin the behavior, and the `--cache-probe` mode in the new bench makes it reproducible. Resolved, thanks.

flashinfer/gemm/gemm_base.py
dhiraj113: Verified at ec6ebb7a: missing DSL now raises `ValueError`, and splitting the CuTe path into the standalone `group_gemm_fp8_nt_groupwise_contiguous` (no `backend` argument, so `backend="auto"` is a plain `TypeError`) while leaving `group_deepgemm_fp8_nt_groupwise` byte-identical to `main` is a cleaner resolution than the one-word fix I suggested. `test_grouped_cute_dsl_requires_package` / `_has_no_backend_selector` cover both. Resolved, thanks.

flashinfer/gemm/gemm_base.py
dhiraj113: Verified at ec6ebb7a: `M == 0` returns an empty `(0, N)` output — or the supplied `out` unchanged — before any import, compile, or launch, with the gate's positivity check correctly scoped to `n`, `k`, and the group count and the `validate_indices` path guarded for the empty case. `test_grouped_cute_dsl_empty` covers both variants. Resolved, thanks.

dhiraj113 (approved):
Re-reviewed at head `ec6ebb7a` — all four findings from the previous review were re-verified by re-running our original repros (adapted to the new standalone API) on a B200 (sm100), and the PR's 70 new/updated cute tests pass there as well.

All four are resolved: opt-in `validate_indices` rejection of invalid routing with the default path honestly documented as undefined behavior; the compile cache keyed on the 128-row alignment class with a proper lock (new token counts now hit in ~0.2 ms); the CuTe path split into `group_gemm_fp8_nt_groupwise_contiguous` with `group_deepgemm_fp8_nt_groupwise` restored unchanged; and `M == 0` handled before any launch. The smaller items I had held back were addressed too — negative-path tests, `out=` coverage, cross-stream synchronization in the side-stream tests, `can_implement` wired into the host entry, device-pinned compilation, and an in-repo benchmark (`bench_grouped_fp8.py`) that makes the perf table reproducible.

Approving. (AI-assisted review per `docs/code_review_guidance.md`; every verdict above is backed by an executed repro at this head.)

nv-yunzheq:
@flashinfer-bot run tests/gemm

nv-yunzheq:
@flashinfer-bot run

nv-yunzheq:
@flashinfer-bot run

PR #4735, opened 2026-08-25
Fix MoE Tile Mapping Zeroing

Vinnie6167:
## 📌 Description

Removes an unnecessary zeroing of MoE tile mapping.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Bug Fixes

* Improved validation for unsupported multi-CTA MoE execution configurations, with clearer errors before launch.
* Corrected tile-count handling during Rubin MoE execution.
* Prevented invalid or padded tiles from causing out-of-bounds metadata reads.
* Preserved sorting buffer contents until execution begins.
* Improved bounds handling for odd tile counts, including CUDA graph replay scenarios.

## Performance

* Reduced unnecessary buffer initialization during MoE sorting, improving execution efficiency.


Vinnie6167:
@flashinfer-bot run

yufeiwu-nv:
Conclusion: PR [#4735](https://github.com/flashinfer-ai/flashinfer/pull/4735) fixes the performance regression reported in Bug 6661924.
[Pipeline 64626498] succeeded. The raw GB300 CSV results were:

Run 0:0.0548165 ms
Run 1:0.0551365 ms
Average:0.0549765 ms
Run-to-run difference: approximately0.58%
Both measurements are within the expected 0.0546–0.0552 ms range. The average is approximately 5.38% faster than the regression-range midpoint of 0.0581 ms, significantly exceeding normal benchmark variance.

flashinfer/fused_moe/cute_dsl/fused_moe.py
```diff
@@ -277,6 +277,32 @@ def _moe_core_impl(
...
+    # already exited. Handling that needs two coordinated changes: round the
+    # count the kernel sees up to a multiple of cluster_shape_m, and initialize
+    # the tile-metadata buffers to match, because the kernel's bounds check on
+    # them is that same rounded value.
+    #
+    # No reachable Rubin tactic uses a multi-CTA cluster today -- tile_size is
+    # restricted to 128, which forces mma_tiler_m == 128 and hence
+    # cluster_shape_m == 1 -- so rather than carry a pair of coupled code paths
+    # nothing exercises, refuse the case loudly. Whoever re-enables
+    # tile_size=256 gets a clear pointer instead of a silent uninitialized read.
```
bkryu: Can we try to trim this one?

Vinnie6167:
@flashinfer-bot run

bkryu (approved):

flashinfer/fused_moe/cute_dsl/fused_moe.py
```diff
@@ -277,6 +277,22 @@ def _moe_core_impl(
...
             memset_event = memset_event or resources["memset_event"]
 
+    is_rubin = gemm1_mma_tiler is not None and gemm1_mma_inst_shape is not None
+
+    # No reachable Rubin tactic uses a multi-CTA cluster (tile_size is capped at
+    # 128, which forces cluster_shape_m == 1), so refuse rather than carry a
+    # workaround nothing exercises.
+    cluster_m = max(gemm1_cluster_shape_mn[0], gemm2_cluster_shape_mn[0])
+    if is_rubin and cluster_m > 1:
+        raise NotImplementedError(
```
nv-yunzheq, line 360: Is it possible to move the logic to support check instead a runtime error?

flashinfer/fused_moe/cute_dsl/moe_utils.py
```diff
@@ -652,23 +652,15 @@ def moe_sort(
     # Pre-allocation is required for CUDA graph compatibility
     if out_tile_idx_to_expert_idx is not None:
         tile_idx_to_expert_idx = out_tile_idx_to_expert_idx
-        # Zero-fill to ensure safe defaults for entries beyond num_non_exiting_tiles.
-        # This prevents out-of-bounds weight accesses when Rubin kernels round up
-        # the tile count to an even number for cluster synchronization.
-        tile_idx_to_expert_idx.zero_()
     else:
-        tile_idx_to_expert_idx = torch.zeros(
+        tile_idx_to_expert_idx = torch.empty(
```
nv-yunzheq, line 661: @leejnau Could you take a look to see if you expect any potential issue with this?

leejnau: > ###
> **[nv-yunzheq](/nv-yunzheq) ** [Aug 28, 2026](#discussion_r3884048047)

After tracing the final diff, I don’t think this introduces an output-correctness regression. The zero-fill was needed because the previous Rubin path rounded an odd `num_non_exiting_tiles` up to an even count, exposing one unwritten padding entry. This PR removes that rounding, passes the exact active count to both GEMMs, and rejects Rubin `cluster_shape_m > 1` before `moe_sort`; current Rubin tactics are all `cluster_shape_m == 1`.

So removing only the zero-fill would be a bug, but the coupled final change is safe. One cleanup is still needed: GEMM2 currently loads `tile_idx_to_expert_idx[tile_idx]` before checking `tile_idx < num_valid_tiles`. Although the value is discarded, with `torch.empty` this becomes an uninitialized read. Could we move that load inside the existing bounds check in both the Blackwell and Rubin schedulers?

leejnau:
Could we add focused coverage for the new initialization contract?

* On SM107, force an odd active tile count, poison both tile-map buffers, and compare eager and CUDA-graph replay results against the reference. Change routing between replays so stale tail values remain present and would expose any regression in the active-count bounds contract.
* Assert that every currently enumerated Rubin tactic has `cluster_shape_m == 1` for both GEMMs.
* Construct a synthetic Rubin `cluster_shape_m == 2` tactic and verify it is rejected before routing or GEMM execution.

flashinfer/fused_moe/cute_dsl/moe_utils.py
leejnau, line 661: Could we add a comment something like:

```
# Entries outside the active prefix are intentionally uninitialized.
# Routing writes [0, num_non_exiting_tiles), and both GEMMs consume exactly
# that count. Rubin multi-CTA tactics, which would require rounded/padded
# metadata entries, are rejected before moe_sort.
```

Note this is true only after the GEMM2 expert-index load moves inside the bounds check. Your first comment already makes that dependency clear.

Vinnie6167: I moved the load to after the bounds check. See `blockscaled_contiguous_grouped_gemm_finalize_fusion.py‎`

Vinnie6167: Done

Vinnie6167:
> On SM107, force an odd active tile count, poison both tile-map buffers, and compare eager and CUDA-graph replay results against the reference. Change routing between replays so stale tail values remain present and would expose any regression in the active-count bounds contract.
Assert that every currently enumerated Rubin tactic has cluster_shape_m == 1 for both GEMMs.
Construct a synthetic Rubin cluster_shape_m == 2 tactic and verify it is rejected before routing or GEMM execution.

@leejnau I have added the recommended test. Can you give them a review?

Vinnie6167:
@flashinfer-bot run

Vinnie6167:
@flashinfer-bot run

aleozlx:
## PR Review Screening

**CI verdict:** ✅ auto-run ok
**Review category:** `human` (rule fired: C3.2 durable kernel-internal changes with adequate tests; C1.2 also narrows accepted input by rejecting multi-CTA Rubin)
**Blocking checks:** none
**Release blocker:** 🚨 candidate — the substantive fix is an **out-of-bounds metadata read**: `tile_idx_to_expert_idx` / `tile_idx_to_group_idx` / `tile_idx_to_mn_limit` were indexed *before* the `if tile_idx < num_valid_tiles` guard, so padded and invalid tiles read past the written prefix (5 sites across the Blackwell and Rubin finalize-fusion kernels). The removed even-rounding workaround also existed to avoid a **cluster-synchronization deadlock**. Library-side fix with tests, and the PR carries the `v0.6.19` label (C3.4)
**Early stop:** no

### Security
| Q | Answer | Evidence |
|---|--------|----------|
| S3 template obligations | met | 5/5 checked and substantiated — 445 lines of new tests are genuinely in the diff |

### Packaging
| Q | Answer | Evidence |
|---|--------|----------|
| C1.2 public API changes | yes — no signature change; **narrowing** of accepted input | `_moe_core_impl` now raises `NotImplementedError` when a Rubin tactic has `cluster_shape_m > 1` (either GEMM), and `tuner.py` returns `False` for those tactics during enumeration. The in-code comment states the tuner filter is primary and the raise is "the backstop for callers that pass tactic parameters directly" — so autotuned callers were never selecting these, but a caller supplying tactics directly loses a path that previously worked via the even-rounding workaround |

### Presentation
| Q | Answer | Evidence |
|---|--------|----------|
| C2.1 perf claim backed by data | n-a | No speedup is claimed and no measurement is offered. `torch.zeros` → `torch.empty` on two metadata buffers plus the removal of two `zero_()` calls is a real allocation-path saving, but the PR presents it as removing something unnecessary rather than as a perf result |

### Implementation
| Q | Answer | Evidence |
|---|--------|----------|
| C3.2 shared/durable areas touched | yes — **durable** | Five library files in the shared CuteDSL MoE layer: the Blackwell finalize-fusion kernel (3 index sites), the Rubin one (2 sites), `fused_moe.py` `_moe_core_impl` (validation + removal of the rounding), `moe_utils.py` (buffer allocation), and `tuner.py` (tactic enumeration, i.e. kernel selection). Kernel-internal and dispatch-adjacent, but the tactic filter narrows the space rather than reshaping the convention — which is why this lands at `human` rather than `live` |
| C3.3 tests match behavior change | yes | 445 lines across two files, targeting exactly the new behaviour: the rejection is parametrized over `(2,1) / (1,2) / (2,2)` with `pytest.raises(NotImplementedError, match="cluster_shape_m")`; `test_every_rubin_tactic_is_single_cta` pins the enumeration invariant so a future tactic cannot silently reintroduce a multi-CTA cluster; and `target_tiles` `[9, 11]` exercises the odd tile counts the removed rounding used to mask. These are tactic-enumeration and validation tests rather than arch-gated GPU tests, so they execute broadly |

### Experimental track
| Q | Answer | Evidence |
|---|--------|----------|

**Notes for the maintainer**
- **The description materially understates the change.** "Removes an unnecessary zeroing of MoE tile mapping" describes the *last* of three steps. The diff also fixes an out-of-bounds metadata read at five sites and removes a capability (multi-CTA Rubin) by raising. A reviewer working from the description alone would not know either had happened, and the release notes for `v0.6.19` will be drawn from it. Worth expanding before merge; Related Issues is also empty.
- **The three parts are one argument and should be reviewed together.** The zeroing existed *because* the even-rounding workaround made the kernels read metadata slots routing never wrote — the deleted comments say so explicitly ("prevents out-of-bounds weight accesses when Rubin kernels round up the tile count", "a stale value there would corrupt row stores"). Fixing the reads and rejecting the configuration that needed rounding is what makes `torch.empty` safe. If a reviewer rejects the multi-CTA removal, the `zeros` → `empty` change must come back with it; they cannot be taken separately.
- **The capability removal is the one call needing confirmation.** Multi-CTA Rubin worked before via rounding and now raises. The code asserts the tuner already filtered these tactics, and `test_every_rubin_tactic_is_single_cta` now enforces that — so the exposure is limited to callers passing tactic parameters directly. Worth a check that no downstream integration does.

tests/moe/test_cute_dsl_fused_moe.py
```diff
@@ -3632,3 +3632,421 @@ def test_w4a8_fused_moe_tactics_and_apis(
...
+        for _ in range(3):
+            _run()
+        torch.cuda.synchronize()
+
+        g = torch.cuda.CUDAGraph()
+        with torch.cuda.graph(g):
+            output = _run()
+        torch.cuda.synchronize()
+
+        for routing, label in ((routing_a, "A"), (routing_b, "B")):
```
leejnau: This currently replays A (9 active tiles) and then B (11 active tiles). Because the active prefix grows, the second replay overwrites entries 9–10 instead of leaving stale tail values from the previous replay. Could we reverse the order to B → A (11 → 9), and assert the realized active count after each replay? That would exercise the stale-tail condition.

leejnau (approved):

aleozlx (approved):

aleozlx:
@flashinfer-bot run tests/moe

Vinnie6167:
@flashinfer-bot run tests/moe

bkryu:
@flashinfer-bot run tests/moe

bkryu:
@flashinfer-bot run

PR #4736, opened 2026-08-26
perf(attention): specialize paged FA2 equal-stride KV

saltyminty:
## 📌 Description

Paged FA2 currently computes and stores independent V offsets even when K and V have identical strides. That extra address-generation work regresses common equal-stride paged decode and prefill workloads.

This change:

- checks the relevant K/V strides once during host dispatch and selects a compile-time `SAME_KV_STRIDES` specialization;
- reuses the K offsets for V in the equal-stride specialization;
- preserves independent V offsets for unequal-stride layouts, including asymmetric NVFP4; and
- adds focused coverage to the existing tensor-core decode, batch prefill, and JIT generator suites.

The stride comparison is host-side dispatch work. It does not add a per-element branch to the GPU kernel.

## 🔍 Related Issues

https://nvbugspro.nvidia.com/bug/6634590

https://nvbugspro.nvidia.com/bug/6634592

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

- [ ] All tests are passing (`unittest`, etc.). Full repository CI is pending; the focused validation below passed.

Focused validation on B300/SM103:

- JIT specialization test: 1 passed.
- Equal-stride tensor-core decode and paged prefill correctness tests: 2 passed.
- Reference-checked acceptance matrix:

| Workload | Target backend | Median | Gate |
| --- | --- | ---: | ---: |
| Llama-4 Scout | `fa2_tc` | 0.334 ms | ≤ 0.3675 ms |
| Llama-3.1 70B | `fa2_tc` | 0.336 ms | ≤ 0.3675 ms |
| GPT-OSS | `fa2` | 1.355 ms | ≤ 1.446 ms |

All matrix cases returned zero with reference checking enabled and no mismatch diagnostics. CUPTI was unavailable, so benchmark timing used CUDA events. Unsupported control backends were skipped.

## Reviewer Notes

The extra compile-time specialization increases generated build artifacts. Qualification measured approximately +26.9% aggregate object bytes and +33.2% shared-library bytes, with unchanged generated source/module counts. This is the main tradeoff for recovering the equal-stride runtime performance.

POD and batch-POD callers retain the conservative independent-stride default unless they explicitly select the new specialization.

## Summary by CodeRabbit

- **New Features**
  - Added support for paged key/value caches with independent memory layouts and strides.
  - Added automatic routing between matching- and independent-stride execution paths.
  - Added lazy loading and optional prewarming of stride-specific variants before CUDA graph capture.
  - Extended support across batch prefill, decode, cascade, shared-prefix, POD, and block-sparse workflows.

- **Bug Fixes**
  - Improved correctness for bfloat16 paged prefill and tensor-core decoding, including grouped query attention.

- **Tests**
  - Added coverage for routing, reuse, CUDA graphs, compilation, and reference-output correctness.


saltyminty:
@flashinfer-bot run

qsang-nv:
**Local A/B on the build-size tradeoff**

I reproduced the artifact cost locally: took `prefill.cuh` + `batch_prefill_paged_kernel_inst.jinja` from `main` and from this PR's head, rendered the paged kernel TUs from the same config (bf16 q/kv/o, `head_dim_qk = head_dim_vo = 128`, no sliding window / soft cap, `DefaultAttention`) and compiled both with the JIT's flags (`-O3 -DNDEBUG -use_fast_math -std=c++17 --expt-relaxed-constexpr -static-global-template-stub=false`), CUDA 13.2, `sm_90a`. Object bytes over all four mask modes:

| | main | this PR | delta |
|---|---:|---:|---:|
| 4 paged kernel TUs | 3,130,640 | 5,579,424 | **+78.2%** |
| + the 4 unchanged ragged TUs | 5,742,992 | 8,191,776 | **+42.6%** |
| nvcc wall time, 4 paged TUs | 64 s | 85 s | +33% |

The relative increase was similar in my four-architecture build (2,788,992 → 5,006,696 B, +79.5%), suggesting that the cost compounds across the release wheels' multi-architecture fatbins (currently 6–7 architecture targets depending on CUDA/platform). Linking and wheel compression may change the final ratio. This is consistent with the +26.9% / +33.2% in the reviewer notes—the growth is concentrated in the paged TUs (~+78%), while the aggregate figures are diluted by unchanged TUs.

Two observations from `cuobjdump -res-usage` on the same build (`MaskMode::kNone`, `CTA_TILE_Q=16`):

- The `SAME_KV_STRIDES=false` specialization has the same register and spill usage as `main` at every inspected `NUM_MMA_KV`, so I found no resource-usage regression for the conservative path in this build.
- The `true` specialization's register/spill benefit is config-dependent: `NUM_MMA_KV=1` goes 163 → 128 registers and `NUM_MMA_KV=8` goes 456 → 288 spill bytes/thread, while `NUM_MMA_KV=2` goes 178 → 182 registers and `NUM_MMA_KV=4` remains at 255 registers / 104 spill bytes. It still removes V-offset address-generation work in every configuration, so these resource figures alone do not establish the runtime benefit. This is `sm_90a`, not the `sm_103a` on which the bugs were filed, and I did not run the benchmark—the PR's B300 measurements remain the relevant performance evidence.

Since the specialization duplication applies to every paged TU that ships in the wheel—and was ~78% for the configuration measured here—could you run an A/B build of `flashinfer-jit-cache` and post the resulting wheel-size delta? Object bytes don't translate cleanly to the shipped artifact, and the published v0.6.9 wheels are already large (cu128 1314 MiB, cu129 1898 MiB, cu130 2033 MiB, x86_64), so the linked-and-packaged number is the one worth deciding on.

saltyminty:
Updated the implementation to ship equal-K/V-stride FA2 primaries in the default cache and lazily load a separate paged-only module for unequal strides. The complete cu130 x86_64 wheel is now 1.20% smaller than the PR base, replacing the previous 15.10% increase; the B300 correctness and performance acceptance checks passed.

The tradeoff is first-use latency: a cold unequal-stride configuration took 18.35–21.15 seconds to JIT compile/load in testing. Prewarm during initialization (and before CUDA-graph capture) to keep this out of request latency; subsequent calls reuse the module. Unequal layouts require JIT when no compatible cached artifact is available, and fail clearly with JIT disabled. The optional precompiled supplement is deferred.

saltyminty:
@flashinfer-bot run

aleozlx:
I arrived at the same root cause independently in #5071 and am closing that in favour of this PR — your approach keeps the unequal-stride capability that #3684 shipped in 0.6.18, and the module split answers the JIT-cache concern you raised earlier. Three things from my side that may be worth folding in.

**1. `persistent.cuh` has the same defect and is not covered here.**

#3684 doubled the offset array in two places. This PR fixes `prefill.cuh`; `include/flashinfer/attention/persistent.cuh` still carries it on `main`:

```
269:  thr_local_kv_offset_k[NUM_MMA_KV * KTraits::KV_THR_LAYOUT_COL / 2 / KTraits::NUM_WARPS_Q];
271:  thr_local_kv_offset_v[NUM_MMA_KV * KTraits::KV_THR_LAYOUT_COL / 2 / KTraits::NUM_WARPS_Q];
```

plus a second `prefetch_offest` pass at :326/:329 to fill it. So `BatchAttention` keeps the register cost. Entirely reasonable as scope — just flagging it so it is a decision rather than an oversight.

**If you do extend to it, there is a trap.** `csrc/batch_attention.cu:90-101` reads K and V strides from two independent tensors (`k_cache`, `v_cache`) into `params.k_stride_*` / `params.v_stride_*` and **never compares them** — there is no `ICHECK` on that path at all. So the host-side stride comparison this PR does in `batch_prefill.cu` has no equivalent there and would need to be added first; otherwise a `SAME_KV_STRIDES=true` decision silently addresses V through the K offsets. I hit exactly this: my first revision gated `persistent.cuh` without adding the host check, and it was a silent-wrong-output bug until I dropped that half.

**2. Offer: B300 perf numbers.** This PR has correctness coverage but no before/after. I measured on B300 (SM103, CUDA 13.0) using NVBug 6634590's own repro command — `BatchDecodeWithPagedKVCacheWrapper --backends fa2_tc --page_size 16 --batch_size 64 --s_qo 1 --s_kv 8192 --num_qo_heads 64 --num_kv_heads 8 --head_dim_qk 128 --head_dim_vo 128 --q_dtype bfloat16 --kv_dtype bfloat16 --num_iters 100`:

| build | median |
| --- | --- |
| `8f9ad200^` (pre-regression) | 0.801 / 0.804 ms |
| `8f9ad200` (first bad) | 0.859 ms |
| `main` | 0.860 / 0.860 ms |
| with the register fix | 0.787 / 0.797 / 0.803 ms |

i.e. the bisect is confirmed and the fix returns main to the pre-#3684 baseline. Since our kernel-side edits are equivalent, these should carry over — happy to re-run on your head if that is more useful than quoting mine.

**3. NVBug 6634592 (prefill) is unmeasured by both PRs.** Both are marked as fixing 6634590 *and* 6634592, but the only numbers either of us has are for `fa2_tc` decode. `head_dim 64/64` takes the same path, so a single prefill row from 6634592's own repro command would close that out. qsang-nv raised the same point on #5071.

Note the board is not urgent: both NVBugs' first-bad commit `8f9ad200` shipped in v0.6.18, so this is not a new regression for 0.7.0.

aleozlx:
## PR Review Screening

**CI verdict:** ✅ auto-run ok
**Review category:** `live` (rules fired: C1.2 net-new public interface; C3.2 durable change to a dispatch/build convention)
**Blocking checks:** S3 (partial), C2.1 (no)
**Release blocker:** 🚨 candidate — restores a perf regression introduced by #3684 (first-bad `8f9ad200`) that shipped in v0.6.18; NVBug 6634590 / 6634592 (C3.4)
**Early stop:** no

### Security
| Q | Answer | Evidence |
|---|--------|----------|
| S2 template overwritten | no | — |
| S3 template obligations | ❗ partial | "All tests are passing" unchecked — accurate: pipeline #67320202 is 6/19 |

### Packaging
| Q | Answer | Evidence |
|---|--------|----------|
| C1.1 dependency bump | no | no supply-bucket or submodule changes |
| C1.2 public API changes | yes — **net-new interface** | `prewarm_paged_kv_stride_variant(self, variant: str = "independent") -> None`, `@flashinfer_api`, added to four public wrappers: `BatchPrefillWithPagedKVCacheWrapper` (`prefill.py:2848`), `BatchDecodeWithPagedKVCacheWrapper` (`decode.py:2049`), `MultiLevelCascadeAttentionWrapper` (`cascade.py:520`), `BatchPrefillWithSharedPrefixPagedKVCacheWrapper` (`cascade.py:1021`). `gen_batch_prefill_module()` keeps its signature; the new modes are private `_gen_*` entry points |
| C1.3 AOT/trace registration | ❗ gap | deliberate: `aot.py` ships only the equal-stride primary (`_gen_batch_prefill_primary_module`); the independent-stride paged module is JIT-only, so prebuilt-cache users on unequal K/V strides get `MissingJITCacheError` — a capability #3684 shipped in 0.6.18. Reviewer decision, not an oversight |

### Presentation
| Q | Answer | Evidence |
|---|--------|----------|
| C2.1 perf claim backed | ❗ no | acceptance matrix reports post-change medians against thresholds (B300/SM103) — no baseline for this head; the before/after in the thread was measured on an equivalent branch, not this one |

### Implementation
| Q | Answer | Evidence |
|---|--------|----------|
| C3.2 durable areas touched | yes — durable | `SAME_KV_STRIDES` template parameter on `BatchPrefillWithPagedKVCacheDevice` (`include/flashinfer/attention/prefill.cuh:3799`) plus a new module-surface / URI-suffix convention in `flashinfer/jit/attention/modules.py` |
| C3.3 tests match change | yes | new `tests/jit/test_batch_prefill_lazy_loader.py` (309 lines) and `tests/jit/test_attention_utils.py` stride-mode cases, plus stride variants in 4 attention suites — all in default lanes |
| C3.4 release-blocker candidate | 🚨 yes | fixes a regression in a shipped release (v0.6.18); library and test sides both change |

### Experimental track
| Q | Answer | Evidence |
|---|--------|----------|

**Notes for the maintainer**

- The 6680 `tests.attention.test_attention_sink` failures the pipeline filed under *"New relative to nightly (attribution uncertain)"* are attributable to this PR: `flashinfer/attention/_core.py` now adds `paged_kv_stride_mode` and `module_surface` to the customize jit kwargs, but `get_customize_batch_prefill_module` (`flashinfer/prefill.py:179`) was not extended to accept them — matching the reported `TypeError`. The bot's attribution is path-touch based and that test path is untouched, so the label understates it.
- The one job the bot *did* label **PR-related** — `test_shared_prefix_kernels` on H100 — is this PR's own new test meeting its own guard: prewarm raises on `backend='fa3'`, which is what the shared-prefix wrapper selects on SM90.
- Both linked NVBugs are claimed fixed, but every number posted so far is `fa2_tc` decode; 6634592 (prefill) is still unmeasured by either this PR or the closed #5071.

saltyminty:
/bot run tests/attention/test_attention_sink.py tests/attention/test_shared_prefix_kernels.py tests/jit/test_attention_utils.py tests/jit/test_batch_prefill_lazy_loader.py

saltyminty:
/bot run tests/attention/test_attention_sink.py tests/attention/test_shared_prefix_kernels.py tests/jit/test_attention_utils.py tests/jit/test_batch_prefill_lazy_loader.py

saltyminty:
Performance measurements show that performance has recovered to pre-regression numbers:

Case | Backend | Before (ms) | After (ms) 
-- | -- | -- | -- 
Llama-4-Scout decode | fa2_tc | 0.411974 | 0.334672 
Llama-3.1-70B decode | fa2_tc | 0.413104 | 0.335893 
GPT-OSS prefill | fa2 | 1.445904 | 1.359776

qsang-nv (approved):

PR #4737, opened 2026-08-26
Paged-MQA logits: Rubin (SM107) enablement, FP4 next_n=4, and DKG epilogue/scheduler optimizations

dhiraj113:
## 📌 Description

Three changes to the paged-MQA logits kernels (`flashinfer.attn_scores`): enable them on Rubin (SM107), support FP4 `next_n=4` on both architectures — natively on Rubin, via an in-kernel atom split on Blackwell — and port the epilogue/scheduler optimizations from DKG MR !26202.

### 1. Enable paged-MQA logits on Rubin (SM107)

`fp8_paged_mqa_logits` / `fp4_paged_mqa_logits` refused SM107 outright: their capability list was `[100, 103]`, so dispatch raised before any kernel ran. No kernel changes were needed — both CuTe-DSL kernels use the family-portable tcgen05 surface and compile for `sm_107a` as-is.

The change also separates two questions that were conflated: *does FlashInfer ship a kernel for this tier* (the capability list) versus *can the installed CuTe DSL emit code for this device*. A DSL release predating Rubin fails on `sm_107a` deep inside `cute.compile`; the new probe rejects that up front, using `native_only=True` because these kernels need block-scaled `tcgen05` MMA, for which the family-conditional (`sm_100f`) fallback is not sound.

Multi-GPU correctness note: the kernels' construction-time arch decisions (the FP4 atom cap, the per-arch optimization gates, the TMEM column count) are derived from the compile-target arch threaded from the input tensor's device — not from the DSL's ambient device-0 probe — so heterogeneous Blackwell+Rubin nodes get the right specialization in both directions, and nothing arch-dependent is cached under a key that does not carry it.

### 2. FP4 `next_n=4` (port of DKG MR !25506)

The kernel packs all `next_n` speculative positions into one MMA tile so the KV cache is streamed once for all of them. The tile's accumulator lives in TMEM, and `next_n=4` needs 544 columns: over SM100's 512-column cap, within Rubin's 576 via an exclusive (non-power-of-two) allocation. So the same shape that is physically impossible on Blackwell runs as a single direct kernel on Rubin.

Two pieces, both from the upstream MR:

- **Arch-aware TMEM sizing** — below 512 columns the SM100 rule stands (pow2, min 32); above it, 32-column-aligned exclusive allocation, capped by `get_max_tmem_alloc_cols(arch)`.
- **In-kernel atom split** — `next_n` decomposes into atoms; the MMA/TMEM tile is sized by the *atom*, so the atom is what the arch cap gates. The kernel maps each atom task back onto the native `block_tables`/`seq_lens` (per-atom context is staggered to reproduce the unsplit causal masks), which is what lets Blackwell run `next_n=4` as two atoms of 2 inside one launch, with no page-table duplication and no context-length rebuild.

This is the follow-up that commit 29ca0629 anticipated when it removed the old caller-side `next_n=4` split ("the kernel needs to be smarter to avoid unnecessary tensor creation at runtime"). The in-kernel split eliminates that removal's per-call tensor costs; the one hazard that cannot be engineered away — a caller-supplied `schedule_meta` under a split describes B rows while the kernel iterates B×atoms, which **hangs** the persistent kernel — is handled by construction: `compute_paged_mqa_logits_schedule` now takes the call's `next_n`/`variant` and applies the same internal decomposition, so a correctly-built caller schedule exists for every configuration (a mismatched one is caught by the opt-in freshness check); that exact combination (untested when 29ca0629 found it) now has a test.

The decomposition is a **fixed internal rule, not an API knob**: always direct, splitting into the minimal legal atoms only where TMEM forces it (`next_n=4` on SM100/SM103 → two atoms of 2). A forced-decomposition sweep (batch 1–64, ctx {4K, 16K}, `next_n` {2, 4}, CUDA-graph-timed with `benchmarks/bench_paged_mqa_logits.py`) on both B100 (SM100) and Rubin measured direct winning or tying every cell: upstream's wave-count heuristic would pick small-batch splits that run up to 1.26× slower than direct here, and finer-than-minimal splits of the forced case lose up to 1.37× at large batch. The small-batch occupancy win that motivates a shape heuristic upstream does not materialise on this repo's kernels, so callers state the problem and the library picks the strategy — nothing shape- or user-dependent survives in the decomposition.

### 3. Epilogue/scheduler optimizations (port of DKG MR !26202)

Five SASS-campaign findings expressed at source level, applied to both kernels while keeping this repo's epilogue text: paired pipeline polls (fp8; the two mbarrier round-trips overlap), a loop-invariant TMEM tiled-copy (the per-task rebuild re-derived invariant copy descriptors in the hot loop), a flat logits view with a carried row offset (the store's row term only changes on q-change), a two-level task loop (the rare q-change/rollover blocks move off the hot path; the task sequence is replayed exactly), and per-role scheduler state (deriving the loop-carried scheduler words before the warp-role branches spilled them to local memory).

Also from that MR, one **correctness** fix: an SFB WAR-closure mbarrier in the FP4 kernel. At a q transition, warp 0's s2t `tcgen05.cp` overwrites the SFB TMEM region both UMMA warps' previous-q MMAs read, and `mma`→`cp` is not a pipelined tcgen05 pair — MMA completion is only observable via `tcgen05.commit`→mbarrier. Relying on issue-order completion is unspecified on both architectures, so the closure is unconditional.

Upstream measured several of these levers as **regressions on Rubin**, so they carry per-arch auto gates (kernel-internal, deterministic; the public API is unchanged). One deliberate divergence from upstream's policy: with this repo's epilogue, both gated levers regress at a next_n atom of 3 on B200 — the register-pressure maximum (fp32 w-cache cap 56) — confirmed by ablation (both-on −5%, both-off = baseline), so atom=3 is gated off too.

### 4. API alignment (breaking, pre-release)

The functions inherited DeepGEMM's vocabulary; they now use flashinfer's: `seq_lens` / `block_tables` / `max_seq_len` / `q_sf` / `padded_seq_len()`, with `block_tables` preceding `seq_lens` positionally as in `trtllm_batch_decode_with_kv_cache`. Old positional callers fail loudly on the rank checks; old keywords raise TypeError. Trace templates and committed artifacts were regenerated against the new signature. `num_epi_subtiles` was removed from both signatures for the same reason `next_n_atom` never shipped: measured never-better-than-default at every reachable geometry and numerics-neutral; the compile layer keeps the parameter and its cache tag so a future per-shape policy can route a winning value through.

A follow-up consistency pass (a 47-finding audit across every user-facing surface) finished the job: kernel implementation constants (`SPLIT_KV=256`, the 128-token compute tile) no longer appear in user-facing text -- the contracts are expressed through two opaque sizing helpers, `padded_seq_len()` (output columns) and the new `min_block_table_width()` (block-table width; this also fixed the docstring Examples, which under-allocated the table). `compute_paged_mqa_logits_schedule(next_n=, variant=)` builds the right schedule on every architecture, replacing the former fp4 `next_n=4` rejection, so the caller-buffer CUDA-graph pattern works on every configuration. Vocabulary is uniformly block/seq_lens (the benchmark flag is now `--seq-len`), and both docstrings carry dtype-annotated formulas, fused-KV byte-map diagrams with nibble/endianness zoom-ins, and validated `out=` examples (run verbatim under FLASHINFER_VALIDATE_INPUTS=1).

### 5. Review follow-ups (Lee Nau's review, 2026-09-11)

Ten threads, each fixed at the layer that owns the constraint, one commit per root cause:

- **Natural-width block tables** (`ea926adf`): both kernels tail-predicate the block-table load per column on the row's own `ceil(seq_len / block_size)` instead of per 128-token compute tile, so the conventional serving-stack table (SGLang/vLLM `ceil(seq / page_size)`) is the contract and `min_block_table_width()` collapses to that width. compute-sanitizer showed the previous kernels reading 4 bytes past a natural-width table's allocation (`seq_lens=[129, 257, 385]`, block 64); 0 errors after. Graph-replay perf on B200 and Rubin is at parity (interleaved A/B on the noisiest cells). New SGLang-shaped regression, eager under `FLASHINFER_VALIDATE_INPUTS=1` and under CUDA-graph replay, plus an opt-in memcheck test.
- **Internal fp8 epilogue policy** (`b2111922`): at `num_heads=32` the epilogue's weight register cache spilled (Nsight Compute: ~12k local loads / ~8k local stores per launch at `next_n=6`); a 96-register budget removes every spill and runs 8-35% faster at `next_n=6` and up to 37% faster at `next_n=8` on B200/Rubin, bit-identical to the shipped kernel. Encoded as an internal per-shape rule (`_fp8_epilogue_policy`: fp32 epilogue at `num_heads=32` -> 96-register budget; its fp16 epilogue at `next_n=6` -> two subtiles; default elsewhere), keyed into the compile cache and artifact tag; no public knob. `benchmarks/bench_paged_mqa_logits.py --num-heads 32 --next-n 6` reproduces the numbers.
- **Contracts enforced only where the kernels have them** (`e1883993`, `35beee0e`, `37966de2`, `5a1ea510`): `block_tables[::next_n]` and `seq_lens[::next_n]` views are accepted zero-copy (symbolic strides in all three compiled kernels; only `block_tables.stride(1) == 1` is checked); `out=` rows must be distinct storage (`stride(0) >= padded_seq_len(max_seq_len)`; a pitched `as_strided` or `expand()`ed view used to return silently wrong logits); `compute_paged_mqa_logits_schedule` validates rank / dtype / device up front on every internal path (`[1, B]` used to be accepted silently by two of them); fp8 `num_heads` must be a multiple of 4, checked at the API boundary instead of leaking the kernel constructor's `num_epi_subtiles` message.
- **Split-schedule consistency** (`d57b50aa`, `c197056f`): the stale-`schedule_meta` error's recipe is the exact call that repairs it (a test parses and executes it); precompile warms the schedule buckets the fp4 atom split actually requests (`sched_b64` for `B=32` on SM100/SM103); a caller-supplied schedule no longer triggers an unconditional length expansion under capture.
- **Benchmark inputs** (`fdbeb497`): `kv_fused` is built in the API's fused `[all values][all scales]` layout (the old per-token slice left 66% of fp8 and 11% of fp4 logits non-finite) with a finiteness self-check over the API-defined output region only (`293768a8`: the causal tail is unspecified and the fp4 next_n=4 split leaves it unwritten at seq_len=257; out= is NaN-prefilled so an unwritten defined cell cannot hide, 257 joins the default grid, and a test pins the contract) that aborts the sweep on drift.

## 🔍 Related Issues

- DKG MR !25506 (*"Support FP4 paged-MQA next_n=4 on Rubin"*) and MR !26202 (*"fp8_paged_mqa_logits: port sass-gen (epilogue) optimization back to cute dsl kernel"*) — the upstream sources.
- Commit 29ca0629 — the deliberate removal of the previous caller-side split, whose recorded hazards this PR re-verifies one by one.

### ✅ Pre-commit Checks

## 🧪 Tests

`tests/attn_scores`: **599 passed / 14 skipped on B200 (SM100)**, **598 passed / 15 skipped on Rubin (SM107)**, and **602 passed / 11 skipped on a two-GPU SM100 pair** (which exercises the cross-device rejection tests); SM80/89/90/120 skip cleanly (the host-side validation tests pass everywhere). A fresh-eyes API-usability review round additionally landed: runnable docstring examples (the fp8 one verified byte-identical to the test helper's fused-KV layout), the output-validity contract in the Returns sections, literal-valued Restrictions, a consolidated CUDA-graph recipe per function, and curated errors for every reachable failure path (wrong-rank/dtype/device/layout inputs, schedule-helper input validation, precompile toolchain guards, and a block-table pool-bounds check under FLASHINFER_VALIDATE_INPUTS). New coverage: `next_n=4` numerics against the torch reference on both architectures (direct on Rubin, the in-kernel two-atom split on Blackwell — the only decompositions that ship); an adversarial zero-work-atom test (a short-but-nonzero row's leading atoms are zero work under a split, which must not desync the producer prefetch); the helper-built caller `schedule_meta` under the split (accepted and equal to the internal schedule; a default-args schedule is caught by the opt-in freshness check); and a construction-time arch-gate regression test pinning kernel specialization to the compile target rather than the ambient device. The trace suite (template consistency, registry, paged-MQA reference/checker) passes with the templates and committed artifacts updated to the multi-arch (`sm100/sm103/sm107`) tags.

Performance — methodology and reproduction are checked in as `benchmarks/bench_paged_mqa_logits.py` (CUDA-graph replay isolates device work, which is the timing behind the numbers below; the script also reports eager wall-clock per call, which on these microsecond kernels is host-dispatch-bound and several times larger). CUDA-graph timing vs `main`: the `next_n<=3` Blackwell paths are net faster — the !26202 port measures **+4.7% geomean on B200 FP4 (up to +11.7%)**, FP8 neutral-to-positive in this harness. The marginal cost of the 4th speculative position over the 3rd is ~1.20× on Rubin (direct) versus ~1.55× on Blackwell (split, KV read twice); Rubin runs these kernels ~1.7× faster than B200 at serving shapes (SM ratio 1.405×), rising to 2.3–2.4× for FP4 `next_n=4`. On Rubin the !26202 levers measure neutral (back-to-back ablations; the gates keep upstream's measured regressions off) — its unconditional benefit there is the SFB WAR fix.

## Summary by CodeRabbit

- **New Features**
  - Added FP4 and FP8 paged attention support for Rubin (SM107).
  - Enabled FP4 attention with `next_n=4` through automatic device-specific execution.
- **Performance**
  - Improved FP4 and FP8 attention with architecture-aware scheduling and memory optimizations.
- **Bug Fixes**
  - Improved handling of short, empty, and partially processed sequences.
  - Added validation for incompatible schedules during split execution.
- **Documentation**
  - Updated attention documentation and architecture metadata to include Rubin support.


dhiraj113:
@flashinfer-bot run

dhiraj113:
@flashinfer-bot run

tests/attn_scores/test_attn_scores_adversarial.py
```diff
@@ -718,3 +718,102 @@ def test_adv_interspersed_zero_row_uses_correct_q(variant):
...
+
+    The shape combines every skip trigger: a native zero row (all atoms
+    skipped), a ctx=2 row (leading atoms skipped, trailing atom live), and a
+    num_sms*256 row so one CTA's range spans the gaps. Trailing zeros cannot
+    expose the bug. Weights are distinct per (b, t) so any row mixup is
+    unmissable. The decomposition is the fixed internal rule, so on Blackwell
+    this exercises the 2-atom split (the only split that can execute) and on
+    Rubin the direct path, where zero-work atoms coincide with zero rows.
+    """
+    _skip_if_not_sm100()
```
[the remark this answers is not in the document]
dhiraj113: The gate already admits Rubin: the helper calls is_sm100a_supported, which returns True for any compute-capability major==10 device (SM100/SM103/SM107) -- and this suite runs on Rubin, where 582 tests pass on the current head. The misleading part was the name, so 0d15bacd renames it to _skip_if_no_paged_mqa_support with a skip message naming the class.

dhiraj113:
@flashinfer-bot run

dhiraj113:
@flashinfer-bot run

dhiraj113:
@flashinfer-bot run

dhiraj113:
@flashinfer-bot run

dhiraj113:
@flashinfer-bot run

dhiraj113:
@flashinfer-bot stop

dhiraj113:
@flashinfer-bot run

dhiraj113:
@flashinfer-bot stop

dhiraj113:
@flashinfer-bot run

dhiraj113:
@flashinfer-bot rerun failed

flashinfer/attn_scores/attn_scores.py
```diff
@@ -847,95 +1141,172 @@ def _gpu_schedule(
...
+    index (0 works).
+
+    Allocate ``block_tables`` as::
+
+        [batch_size, min_block_table_width(int(seq_lens.max()), block_size)]
+
+    (using ``max_seq_len`` as the bound also works; see the Example in
+    :func:`fp8_paged_mqa_logits`).
+    """
+    return -(-seq_len // _COMPUTE_BLOCK_KV) * (_COMPUTE_BLOCK_KV // block_size)
```
leejnau, line 1185: SGLang’s DSV4 metadata emits the conventional ceil(seq/page_size) page-table width. For raw length 1028, its compressed path has seq_len=257, block size 64, and 5 table entries, while this kernel reads 6. Because the width check is opt-in and skipped during CUDA graph capture, a direct integration risks a device-side OOB read rather than a clean rejection. Could the final physical-block load be tail-predicated, or could an always-on shape check fail safely? Please add an SGLang-shaped odd-width regression.

flashinfer/attn_scores/attn_scores.py
```diff
@@ -306,69 +354,114 @@ def _validate_schedule_meta_fresh(
...
+    expected = compute_paged_mqa_logits_schedule(seq_lens, device=schedule_meta.device)
     if not torch.equal(expected, schedule_meta):
         raise ValueError(
-            f"{fn_name}: schedule_meta does not match context_lens. It is a "
-            f"function of the whole ceil(context_lens / {_SPLIT_KV}) vector and "
+            f"{fn_name}: schedule_meta does not match seq_lens. It is a "
+            f"function of the whole ceil(seq_lens / {_SPLIT_KV}) vector and "
             f"the device SM count, so a single sequence crossing a "
             f"{_SPLIT_KV}-token boundary invalidates it even when every shape is "
             f"unchanged. Recompute it with compute_paged_mqa_logits_schedule("
```
leejnau, line 364: This recovery recipe is unsafe for SM100/103 FP4 next_n=4. Freshness is checked against the expanded 2B problem, but the suggested default helper call on native [B] lengths creates a different schedule. Once validation is skipped during CUDA graph capture, the resulting persistent-kernel endpoint mismatch can hang. Consider recommending the exact call compute_paged_mqa_logits_schedule(seq_lens, next_n=4, variant="fp4", out=schedule_meta) and add a test that verifies the remediation text.

flashinfer/attn_scores/attn_scores.py
leejnau, line 2515: batch_sizes is documented in caller units, but SM100/103 FP4 next_n=4 schedules 2B atom rows. For B=32, precompile warms sched_b32 here, while the first runtime or capture call requests sched_b64. Please derive the reachable scheduler row counts from variant/next_n using _fp4_atom_decomposition before aligning and deduplicating them. A cache-hit regression should expect bucket 64 on SM100/103 and bucket 32 on SM107.

flashinfer/attn_scores/attn_scores.py
```diff
@@ -306,69 +354,114 @@ def _validate_schedule_meta_fresh(
...
-            f"padded_context_len(max_context_len) for the column count. "
+            f"{fn_name}: out must be at least ({rows}, {padded_max_seq_len}); the "
+            f"kernel writes into the padded trailing region. Use "
+            f"padded_seq_len(max_seq_len) for the column count. "
             f"Got shape {tuple(out.shape)}."
         )
+    # The kernel is compiled against a row-major output with unit inner stride;
+    # a column-strided view would reach the FFI binding with a bare layout
+    # error (or worse), so reject it here with the allocation recipe.
+    if out.stride(1) != 1:
```
leejnau, line 459: stride(1) == 1 does not ensure output rows are disjoint. An out tensor with shape (2, 256) and strides (128, 1) passes this check, but adjacent rows overlap by 128 elements and are written concurrently; an expanded (0, 1) view is another example. Please require rows <= 1 or out.stride(0) >= padded_max_seq_len and add overlapping-row and zero-stride tests.

flashinfer/attn_scores/attn_scores.py
```diff
@@ -847,95 +1141,172 @@ def _gpu_schedule(
...
-                        changes -- including on every graph replay where the
-                        lengths may have moved across a 256-token boundary.
+                        recomputed into it whenever the contents of seq_lens
+                        change -- including on every graph replay where the
+                        lengths may have moved.
 
     Returns:
         schedule_meta: [num_sms+1, 2] int32 on CUDA (``out`` if provided).
     """
+    if seq_lens.dtype != torch.int32:
```
leejnau, line 1234: Please validate the helper’s complete [B] contract before dispatch: rank 1, contiguous layout for the GPU path, and exact device equality when seq_lens is already CUDA.  Currently [B, 1], a stride-2 [B] view, or seq_lens on cuda:0 with target cuda:1 pass these checks and fail only at the compiled binding or launch layer. [B, 1] is a real SGLang metadata shape, so an adapter could easily select the wrong tensor.

flashinfer/attn_scores/attn_scores.py
```diff
@@ -1003,17 +1401,19 @@ def _check_fp8_paged_mqa_logits_supported(
     # 8 <= N <= 256 and N % 8 == 0, but got 320").
     if N < _MMA_N_MIN or N > _MMA_N_MAX or N % _MMA_N_MULTIPLE != 0:
```
leejnau, line 1402: The wrapper now fixes num_epi_subtiles=1, but the public validator no longer checks the kernel’s num_heads % 4 == 0 requirement. For example, next_n=4, num_heads=10, and head_dim=128 give N=40 and pass the public checks, but then fail during JIT construction. Consider either restoring the fixed-policy divisibility check and/or adding a regression test.

flashinfer/attn_scores/attn_scores.py
```diff
@@ -235,57 +253,87 @@ def _cached_num_sms(device_index: int) -> int:
...
+    # strided view reaches the FFI binding with a bare layout error otherwise.
+    if not seq_lens.is_contiguous():
         raise ValueError(
-            f"context_lens must be 1-D with shape[0] == batch_size ({batch_size}) "
-            f"inferred from q.shape[0]; got shape {tuple(context_lens.shape)}"
+            f"{fn_name}: seq_lens must be contiguous; got stride "
+            f"{tuple(seq_lens.stride())}. Call .contiguous() first."
         )
-    if block_table.dim() != 2 or block_table.shape[0] != batch_size:
+    if not block_tables.is_contiguous():
```
leejnau, line 319: Could block_tables accept stride(1) == 1 with a symbolic outer stride? SGLang target verification intentionally passes block_tables[::next_n] to de-expand its repeated table without copying. That view retains unit inner stride but has a larger outer stride, so requiring full contiguity forces another serving-path buffer and copy. KV and output already support symbolic outer strides.

flashinfer/attn_scores/attn_scores.py
```diff
@@ -611,6 +873,13 @@ def _cached_compile_fp8_kernel(
     epi_dtype,  # cutlass dtype object
     acc_dtype,  # cutlass dtype object
     output_dtype,  # cutlass dtype object
+    # num_epi_subtiles is deliberately NOT public: measured on sm_100a at
+    # num_heads=64, subtile counts 1/2/4 tie and 8 is 5-11% slower, so the API
```
leejnau, line 877: The fixed-policy rationale here covers num_heads=64, but FP8 also admits num_heads=32, next_n=6, which is SGLang’s GLM-5.2 geometry. SGLang measured register spilling with the full 24-weight cache and uses max_w_in_reg=8; its split path also uses two epilogue subtiles and pipeline tuning. Consider benchmarking H32/N6 and adding an internal per-shape policy, or retaining sufficient tuning surface.

flashinfer/attn_scores/attn_scores.py
```diff
@@ -1448,44 +2238,96 @@ def fp4_paged_mqa_logits(
...
     # whatever the batch size, and its min(start_q, batch_size - 1) clamp is -1
-    # here. After the out= branch so a bad buffer still raises.
+    # here. After the buffer validation so a bad out/schedule still raises.
     if B == 0:
         return logits
 
+    # The scheduler enumerates batch*num_atoms q_atom tasks, so it must be
+    # built from the per-atom lengths (identity when num_atoms == 1).  Caller
+    # schedules are freshness-checked against the same expanded lengths.
+    sched_ctx = _expand_seq_lens(seq_lens, num_atoms, atom)
```
leejnau, line 2326: For SM100/103 FP4 next_n=4, a supplied schedule still unconditionally materializes the expanded 2B lengths. Freshness validation returns immediately by default and always returns during CUDA graph capture, so this otherwise-unused subtraction is captured and replayed. Please expand only when building an internal schedule, or lazily when synchronous freshness validation is enabled. Separately, the earlier docstring still incorrectly says caller schedules are rejected.

benchmarks/bench_paged_mqa_logits.py
```diff
@@ -0,0 +1,186 @@
...
+        (batch, next_n, _HEADS), 0x7F7F7F7F, dtype=torch.int32, device=device
+    )
+    kv = torch.randint(
+        0,
+        256,
+        (num_blocks, block_size, 1, _HEAD_DIM // 2 + 4),
+        dtype=torch.uint8,
+        device=device,
+    )
+    kv[:, :, :, _HEAD_DIM // 2 :] = 0x7F
```
leejnau, line 99: FP4’s fused block layout is flat [all values][all scale bytes], but this slice treats it as a per-token [value|scale] layout. At the default shape, it overwrites 240 value bytes and initializes only 16 of the 256 actual scale bytes. FP8 likewise uses arbitrary bytes that can encode NaNs and nonfinite float32 scales. Please construct the value and scale regions separately and concatenate them, as the API tests do.

flashinfer/attn_scores/attn_scores.py
dhiraj113: Agreed -- and the recipe was worse than unsafe, it was ineffective: the recommended default call reproduced exactly the schedule the check had just rejected, so following it could never make the check pass. Fixed in d57b50aa: the freshness check now runs the very public call it recommends, with the failing call's own next_n and variant, and the message's recipe is `compute_paged_mqa_logits_schedule(seq_lens, next_n=<n>, variant='<v>', out=schedule_meta)`. `test_stale_schedule_message_recipe_is_actionable` parses the recipe out of the error text, executes it verbatim into the caller's buffer and checks the retried call is accepted -- for fp8 next_n=2 and for fp4 next_n=4 (the two-atom split on SM100/SM103).

flashinfer/attn_scores/attn_scores.py
dhiraj113: Both fixed in d57b50aa. The expansion now happens only on the internal-schedule branch; a caller-supplied schedule goes straight to the freshness check, which recomputes the expected schedule through the public helper -- so the expansion runs inside it, only when FLASHINFER_VALIDATE_INPUTS=1 and never under capture. `test_fp4_caller_schedule_does_not_expand_seq_lens` patches the expansion and asserts it is never called on the caller-schedule path. The fp4 docstring and its CUDA-graph section no longer say caller schedules are rejected.

flashinfer/attn_scores/attn_scores.py
dhiraj113: Right: precompile warmed sched_b32 and the first fp4 next_n=4 call on SM100/SM103 then compiled sched_b64 at runtime. Fixed in c197056f: the bucket rule lives in one place (`_schedule_bucket`, shared by the launch path and precompile), and precompile derives the reachable row counts from `_fp4_atom_decomposition` for every fp4 next_n it warms, so B=32 warms bucket 32 and, where the split is active, bucket 64 as well. `test_precompile_warms_fp4_split_schedule_bucket` precompiles for B=32, runs fp4 next_n=4 and asserts the schedule-kernel cache took no miss -- bucket 64 on SM100/SM103, 32 on SM107 where the direct path runs. The batch_sizes doc now states the row-count rule.

flashinfer/attn_scores/attn_scores.py
dhiraj113: Yes -- both of your examples passed every check and ran; with the (2,256)/(128,1) pitch, row 1 overwrote the second half of row 0 (reproduced before fixing). 37966de2 adds exactly the rule you state: rows <= 1 or out.stride(0) >= padded_seq_len(max_seq_len). A larger stride (a row slice of a wider buffer) stays accepted. Tests cover the pitched view, the expand()ed zero stride, the wider-buffer slice and the single-row case, and the out= docs on both APIs spell out the rule.

flashinfer/attn_scores/attn_scores.py
dhiraj113: Done in 35beee0e, with one deliberate difference from your list. Rank 1, int32 and exact device equality (when seq_lens is already on CUDA) are checked up front, before the fp4 expansion that used to reshape a [1, B] into a plausible flat vector; [B, 1], the CPU fallback and the split path now all raise the same message, and a two-GPU test pins the cuda:0/cuda:1 rejection on each path. I did not add a contiguity requirement: e1883993 (the block_tables thread below) declares seq_lens with a symbolic stride in all three compiled kernels, so a stride-2 view is read correctly through its layout instead of failing at the binding -- the test runs a `[::2]` view through every path and checks it yields the same schedule as the contiguous tensor. The API only enforces constraints the kernels actually have.

flashinfer/attn_scores/attn_scores.py
dhiraj113: Restored in 5a1ea510 as `_EPI_SUBTILE_UNROLL = 4` next to the other fp8 shape constants, checked in the public checker right after the N-mode rule and listed in the restrictions table. Your (next_n=4, num_heads=10) shape plus (8, 9) are the regression: rejected with "num_heads must be a multiple of 4", without the internal parameter name and without entering the compile layer (the test asserts the compile cache took no miss). The fp8 kernel's comment already referred to this mirror; it exists again.

flashinfer/attn_scores/attn_scores.py
dhiraj113: Yes -- and it turned out neither index tensor needs contiguity at all. e1883993 declares block_tables with a symbolic row stride (inner stride pinned to 1, the one real constraint: a row is loaded as a contiguous run) and seq_lens with a fully symbolic stride, in both kernels and the schedule kernel; the validator now checks only block_tables.stride(1) == 1. `block_tables[::next_n]` and `seq_lens[::next_n]` are accepted zero-copy: the test runs those views (offsets 0 and 1, fp8 and fp4) through the compiled kernels and gets bit-identical logits to the contiguous inputs, and a column-strided table is still rejected with the stride(1) message.

benchmarks/bench_paged_mqa_logits.py
dhiraj113: Confirmed and fixed in fdbeb497; your byte counts match what I measured (240 value bytes overwritten, 16 of 256 scale bytes set), and 66% of the fp8 and 11% of the fp4 logits were non-finite. The regions are now built separately and concatenated, as in the API docstring example, and bench_one asserts the logits in [0, seq_len) are finite after the warm call -- AssertionError is kept out of the sweep's skip set so a future drift aborts the run instead of printing timings. The old construction trips the check for both kinds.

flashinfer/attn_scores/attn_scores.py
dhiraj113: Agreed on all counts, and I took the first option: both kernels now tail-predicate the block-table load per column on the row's own ceil(seq_len / block_size) instead of per compute tile, so the conventional ceil(seq / page_size) table -- your seq_len=257 / block 64 / 5-entry case included -- is the contract. min_block_table_width collapses to that natural width (kept as a helper so callers name the rule in one place), the debug validator checks it and looks only at the entries the kernel reads, and the always-on shape check stays out: the bound needs the per-row lengths from device memory, so it would cost a sync and be impossible under capture anyway -- the kernel is now safe without it.

Evidence: compute-sanitizer on seq_lens=[129, 257, 385] (natural [3, 7] table, longest row last) reported "Invalid __global__ read of size 4 bytes ... 1 bytes after the nearest allocation of size 84 bytes" for fp8 and fp4 before the change and 0 errors after. Cost is one warp-uniform, L2-hot Int32 load per re-prefetch plus a compare per column; the graph-replay ledger on B200 and Rubin (batch {16, 64} x seq {4K, 16K} x next_n {1, 4}, both kinds) is at parity, and an interleaved A/B on the noisiest B200 cells shows no regression (two cells came out faster).

Tests: `test_natural_width_block_table_serving_layout` is the SGLang-shaped regression -- odd lengths, natural width, out-of-pool sentinels in the trailing entries, eager with FLASHINFER_VALIDATE_INPUTS=1 and under CUDA-graph replay, fp8 and fp4; `test_natural_width_block_table_memcheck` runs the same case under compute-sanitizer (opt-in via FLASHINFER_MEMCHECK_TESTS=1, about ten seconds); and every block table the suite builds is now natural width.

flashinfer/attn_scores/attn_scores.py
dhiraj113: You were right, and the spill is real on this kernel too: at (32, 6) with the fp32 epilogue the formula caches 24 weights per slot (144 registers) and Nsight Compute shows ~12k local loads and ~8k local stores per launch; a 96-register budget (16 per slot) removes every spill. I measured rather than transplanting the number: the sweep (one config per process, graph replay, same node per arch) at (batch, seq) = (1, 4K) (16, 4K) (64, 16K) gives, as a ratio to the shipped config, B200 / Rubin:

- (32, 6) fp32, budget 96: 0.92 0.85 0.89 / 0.81 0.89 0.65
- (32, 8) fp32, budget 96: 1.00 0.85 0.71 / 0.79 0.78 0.63
- (32, 4) fp32, budget 96: 1.00 1.00 1.01 / 0.90 1.00 0.95 (neutral)
- (32, 6) fp32, budgets 64 and 48 (SGLang's 8 per slot): also spill-free but 2-4% behind 96 (more SMEM weight reads); two subtiles on top of budget 96: no further gain
- (32, 6) fp16 epilogue: the budget does not bind; two subtiles give 1.00 1.00 0.95 / 1.00 0.97 0.87
- (64, 1..4): the shipped config stays best

Every configuration's logits are bit-identical to the shipped kernel's, so the choice is purely speed. b2111922 adds it as an internal per-shape policy, as you suggested: the kernel gains a `w_cache_regs` budget (default unchanged), the compile layer keys and tags on (num_epi_subtiles, w_cache_regs), and `_fp8_epilogue_policy` encodes the table -- a 96-register budget for the fp32 epilogue at num_heads=32, two subtiles for its fp16 epilogue at next_n=6, the default elsewhere. No public knob or env var, per the repo's rule that execution strategy is decided from measurements on these kernels. Tests pin bit-identity against the default kernel and the distinct cache entries; the bench gained `--num-heads` so `--kind fp8 --num-heads 32 --next-n 6` reproduces the numbers. The stale `_EPI_SUBTILE_UNROLL` cross-reference you would have hit next is restored by the num_heads thread above.

dhiraj113:
@flashinfer-bot run

benchmarks/bench_paged_mqa_logits.py
```diff
@@ -0,0 +1,250 @@
...
+    out = torch.empty(
+        (batch * next_n, padded_seq_len(seq_len)),
+        dtype=torch.float32 if kind == "fp8" else torch.bfloat16,
+        device=device,
+    )
+
+    # Warm: JIT compile + schedule-bucket compile happen here, outside timing.
+    _call(kind, args, out)
+    torch.cuda.synchronize()
+    # The kernels compute every column in [0, seq_len) (the caller masks the
```
leejnau: The fused-layout construction is fixed, but this assertion checks cells outside the API’s defined output region. Row b*next_n+t is valid only through seq_lens[b]-next_n+t; the causal tail is explicitly unspecified. On SM100/103 FP4 with next_n=4, seq_len=257, the first two-row atom schedules context 255, so column 256 in those rows is never written and retains arbitrary torch.empty data, yet this assertion reads it. Could this check only a causal-valid mask and add a 257 boundary case?

flashinfer/attn_scores/attn_scores.py
```diff
@@ -704,6 +1063,47 @@ def _compile_fn():
...
+# The rule below encodes exactly that: a 96-register budget for the fp32
+# epilogue at num_heads=32, two subtiles for its fp16 epilogue at next_n=6,
+# the default everywhere else.  Unmeasured shapes get the default.
+_FP8_EPILOGUE_DEFAULT = (1, None)
+
+
+def _fp8_epilogue_policy(num_heads: int, next_n: int, epi_dtype) -> "tuple":
+    """Return ``(num_epi_subtiles, w_cache_regs)`` for the fp8 kernel; see the
+    measured table above."""
+    if num_heads == 32:
```
leejnau, line 1111: The measurements cover H32 next_n={4,6,8}, but this branch applies the 96-register budget to every FP32 H32 shape. next_n=5 and 7 are both public-valid (N=160/224), and this materially changes their cache: from 32 to 16 weights per slot at N5 and from 20 to 12 at N7. That contradicts the comment that unmeasured shapes use the default.  Could you either include N5/N7 measurements or restrict the policy to the measured shapes?

benchmarks/bench_paged_mqa_logits.py
dhiraj113: Correct, and reproduced exactly as you describe: with out= pre-filled with NaN, fp4 next_n=4 at seq_len=257 on B200 leaves column 256 unwritten in the two leading rows of each request (4 cells, all in the causal tail, none in the defined region); fp8 and the direct fp4 paths write the whole window, so the old check only passed by the luck of torch.empty. 293768a8 masks the self-check to the API-defined region (row b*next_n+t through seq_len-next_n+t), NaN-fills out= once outside the timed loops so an unwritten defined cell can no longer hide behind stale data, and adds 257 to the default --seq-len grid so every default run exercises the boundary. `test_defined_output_region_fully_written_at_tile_boundary` pins the contract the bench relies on: at seq_len=257, next_n 3 and 4, fp8 and fp4, every defined cell of a NaN-prefilled out= is finite after the call.

dhiraj113:
@flashinfer-bot run

flashinfer/attn_scores/attn_scores.py
dhiraj113: Fair -- the rule was broader than the table, and the sentence "unmeasured shapes get the default" was wrong for 5 and 7. I measured both rather than narrowing the rule, same methodology (one config per process, graph replay, (batch, seq) = (1, 4K) (16, 4K) (64, 16K)), B200:

- next_n=5, budget 96 (16 weights/slot instead of 32): 0.99 0.84 0.76 vs the default. Nsight Compute on the default: ~33k local loads + ~22k local stores per launch; 0 at 96.
- next_n=7, budget 96 (12/slot instead of 20): 0.91 0.99 0.91. Default: ~14k local loads + ~11k stores; 0 at 96.
- Both bit-identical to the default kernel, as at 6 and 8.

So at num_heads=32 with the fp32 epilogue the default budget spills at every next_n from 5 to 8 and 96 removes it each time; at 1..3 the formula already caches all 32 heads within 96 registers (identical kernel), and 4 measured neutral. The rule stays as written and the comment now lists every one of those shapes with its numbers (3e813603), and states explicitly what is *not* covered: other num_heads and the other fp16 shapes keep the default. One honesty note on Rubin: its 5/7 ratios (0.88 0.87 0.73 and 0.96 0.85 0.74) point the same way, but that board was stuck at idle clocks during the run (absolutes 8x off its ledger), so the comment marks them as direction only; I will re-measure when the board is healthy.

dhiraj113:
@flashinfer-bot run

leejnau (approved):

kahyunnam (approved):

PR #4739, opened 2026-08-26
feat(decode): add prims-ts backend and is_causal to paged decode

elwhyjay:
## 📌 Description

`BatchDecodeWithPagedKVCacheWrapper` derives the attention mask from `q_len_per_req`, so a multi-token request is always causal:

```python
is_causal = q_len_per_req > 1
```

Non-causal multi-token decode is therefore not reachable from the public API, even though the task-scheduled decode kernel in `flashinfer.attention.prims_ts` takes `mask_type` as a plan argument and supports `"dense"` for GQA, FP8 KV, and fixed `SQ > 1`. This PR connects the two.

Two changes:

1. `backend="prims-ts"` on the paged decode wrapper, following the `cute-dsl` backend structure: a delegate wrapper built in `__init__`, a `plan()` branch, and a `run()` branch that returns before the tensor-core argument assembly.
2. `is_causal: Optional[bool] = None` on `plan()`. `None` keeps the existing derivation, so the other backends behave exactly as before. An explicit value is honored only by `prims-ts`; the other backends raise, because their mask mode is baked into the planned `qo_indptr` and JIT key.

The `kv_len >= q_len_per_req` plan check now applies only under causal masking. Dense attention places no such constraint, and the kernel accepts `kv_len < q_len_per_req` in that mode.

The `prims-ts` kernel exposes a narrow surface, so unsupported wrapper options are rejected rather than silently ignored:

| Option | Reason |
| --- | --- |
| `kv_layout="NHD"` | The kernel takes HND pages with compact inner strides. |
| `use_cuda_graph=True` | The delegate re-allocates `seq_lens` and scratch and re-specializes the kernel on every `plan()`, so a captured `run()` does not follow a re-plan. Graph support needs a graph mode on the prims-ts side and is left to a follow-up. |
| `jit_args` | No JIT customization hook. |
| `logits_soft_cap`, `pos_encoding_mode` | Not in the kernel contract. |
| `fixed_split_size`, `disable_split_kv` | Split-kv topology is chosen internally. |
| `q_data_type != kv_data_type` | The kernel requires them equal. |
| `kv_cache_sf` | No block-scale path for NVFP4 KV. |
| `sinks` | Not in the kernel contract. |
| `return_lse` | LSE and split-kv partials are internal scratch. |
| `skip_softmax_threshold_scale_factor` | Not in the kernel contract. |

A sliding window under a dense mask is rejected by the kernel itself, since `window_left >= 0` requires `mask_type="causal"`.

Automatic backend selection is untouched. `prims-ts` stays opt-in until it is benchmarked against `trtllm-gen` on the causal path.

The `gqa_paged_decode` trace template is unaffected: it is scoped to one query per sequence, where dense and causal masking coincide.

## 🔍 Related Issues

#3570, #3335

### ✅ Pre-commit Checks

## 🧪 Tests

New file `tests/attention/test_prims_ts_decode_backend.py`. The rejection tests run on any CUDA device because they fire before the kernel is planned. The numerical tests are gated on SM100a, where the kernel is qualified.

Numerical coverage: dense and causal against a paged reference at `q_len_per_req` 1 and 4, dense output differing from causal on the same inputs, dense with `kv_len < q_len_per_req`, and graph capture of a fixed plan. The graph test only covers capture and replay of one plan; the wrapper-level `use_cuda_graph=True` flow (re-plan then replay) is rejected for this backend, see above.

Run on B200 (SM100a) with `nvidia-cutlass-dsl` 4.7.0:

```
tests/attention/test_prims_ts_decode_backend.py    24 passed
tests/attention/test_batch_decode_kernels.py       16 passed   (-k uniform_multi_token)
tests/attention/test_workspace_size.py              6 passed
```

## Reviewer Notes

`is_causal` is the spelling at the wrapper boundary because it replaces the variable of that name; the branch translates it to the kernel's `mask_type`.

NVFP4 KV under a dense mask is deliberately out of scope. It needs a block-scale data path inside the kernel rather than wrapper plumbing, so it belongs in a separate PR.

cc:@PerkzZheng

## Summary by CodeRabbit

* **New Features**
  * Added tracing support for paged key-value cache decode planning, capturing configuration and request metadata for improved analysis.
  * Expanded paged decode coverage across causal, dense, sliding-window, and split-KV scenarios.

* **Bug Fixes**
  * Added clearer validation for unsupported backend and runtime-option combinations.
  * Prevented incompatible CUDA Graph planning configurations from being accepted.
  * Improved validation for sequence metadata, grouped-query attention dimensions, tensor contiguity, and sliding-window settings.

### DIFF
--- a/flashinfer/decode.py
+++ b/flashinfer/decode.py
@@ -876,7 +876,8 @@ class BatchDecodeWithPagedKVCacheWrapper:
             no RoPE/ALiBi/soft-cap).
             The ``prims-ts`` backend uses the task-scheduled decode kernel on SM100a/SM103a.
             It is the only backend that accepts ``is_causal=False`` with
-            ``q_len_per_req > 1``, and it requires ``kv_layout="HND"``.
+            ``q_len_per_req > 1``. It requires ``kv_layout="HND"`` and does not
+            support ``use_cuda_graph=True``.
 
         jit_args : Optional[List[Any]]
             If provided, the wrapper will use the provided arguments to create the JIT module,
@@ -984,6 +985,12 @@ class BatchDecodeWithPagedKVCacheWrapper:
             if kv_layout != "HND":
                 raise NotImplementedError(
                     "prims-ts decode backend requires kv_layout='HND'"
+                )
+            # The delegate snapshots seq_lens, scratch, and the compiled kernel
+            # at plan time, so a captured run() does not follow a later plan().
+            if use_cuda_graph:
+                raise NotImplementedError(
+                    "prims-ts decode backend does not support use_cuda_graph=True"
                 )
             from .attention.prims_ts import BatchDecodePagedTSWrapper
 
--- a/flashinfer/decode.py
+++ b/flashinfer/decode.py
@@ -1667,6 +1667,17 @@ class BatchDecodeWithPagedKVCacheWrapper:
                     "prims-ts decode backend requires q_data_type == kv_data_type, "
                     f"got {q_data_type} and {kv_data_type}"
                 )
+            # The delegate derives kv lengths from the page table and has no
+            # seq_lens input, so a divergent seq_lens would be silently ignored.
+            if seq_lens is not None:
+                derived_kv_lens = get_seq_lens(
+                    indptr_host, last_page_len_host, page_size
+                ).to(torch.int64)
+                if not torch.equal(kv_lens_arr_host.to(torch.int64), derived_kv_lens):
+                    raise ValueError(
+                        "prims-ts decode backend derives kv lengths from "
+                        "(indptr, last_page_len, page_size); seq_lens must match them"
+                    )
             self._max_kv_len = int(max(kv_lens_arr_host).item())
             self._prims_ts_wrapper.plan(
                 self._paged_kv_indptr_buf,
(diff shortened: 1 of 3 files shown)


flashinfer/decode.py
saltyminty, line 1514: I don't believe the prim-ts backend is cuda graph compatible (it allocates and it's own `_seq_lens`, workspace, etc. buffers during it's planning stage) – we'd either need additional plumbing or reject prim-ts with `use_cuda_graph=True`

elwhyjay: @saltyminty Thanks for catching this, and I agree. I think rejecting use_cuda_graph=True for prims-ts is the right call for the scope of this PR. I didn't account for the CUDA graph axis when wiring the backend, and looking again, the prims-ts delegate allocates its own seq_lens and workspace on every plan(), so a captured graph would keep pointing at the old tensors after a re-plan. The replay test here only covers the "plan once, replay unchanged" case, which is why it didn't surface.

The underlying gap is that the wrapper's use_cuda_graph=True assumes re-plan then replay, while prims-ts specializes the kernel at plan time and expects graph users to mutate device metadata in place under a static bound. Closing that needs a proper graph mode on the prims-ts side rather than wrapper plumbing, and I think it's worth doing since multi-token decode is exactly where graphs matter in serving.

So for this PR I'll raise on use_cuda_graph=True for prims-ts, narrow the replay test and description accordingly, and open a follow-up for the graph mode. @PerkzZheng, flagging this since it touches prims-ts's own contract. Happy to adjust the follow-up shape if you see a better path.

saltyminty (approved):

PR #4752, opened 2026-08-26
perf(sparse mla): Update sparse MLA half-Q-tile heuristic to apply to short queries only

emascarenhas-nv:
### Summary

This PR restricts the sparse-MLA half-Q-tile heuristic to short queries
(`mMaxSeqLenQ <= 16`). Short decode/MTP queries retain the existing Q8 choice,
while long-query DeepSeek-V4 prefill uses Q16.

The existing heuristic considers the number of KV-split CTAs and batch size,
but not the amount of Q-axis parallelism. For the DeepSeek-V4 prefill case on
B200:

- `numHeadsQPerKv = 16`, so the full and half Q tiles are Q16 and Q8.
- `maxNumCtasPerSeqKv = ceil(min(maxSeqLenKv, sparseMlaTopK) / tileSizeKv)`.
- With `sparseMlaTopK = 2048` and `tileSizeKv = 128`,
  `maxNumCtasPerSeqKv = 16`.
- At batch 1 on a 148-SM B200, `1 * 16 <= 148 / 8`, so the current condition
  selects Q8 even when `maxSeqLenQ = 16384`.

Q8 is useful for short batch-1 queries because it creates more head-splitting
CTAs. A 16K query already supplies 16K-way Q-axis parallelism. In that case Q8
changes the main-kernel grid from `(16384, 1, 1)` to `(16384, 2, 1)` and makes
two CTAs load the same sparse KV set for the two eight-head groups. Q16 avoids
that duplicated KV traffic.

No new cubin is required: both Q8 and Q16 variants are already shipped. The
change only narrows the host-side selector condition.

### vLLM DeepSeek-V4 prefill call site

vLLM invokes this generation-form FlashInfer API from its DeepSeek-V4
`_forward_prefill` path: [`flashinfer_sparse.py:L781-L886`](https://github.com/vllm-project/vllm/blob/728d3ad09/vllm/models/deepseek_v4/nvidia/flashinfer_sparse.py#L781-L886).
Specifically, `_forward_prefill` calls
`flashinfer_trtllm_batch_decode_sparse_mla_dsv4(query=q_chunk, ...)`, which is
why prefill reaches `selectSparseMlaGenerationKernel()`.

### Native FlashInfer B200 benchmark

This PR adds `trtllm_batch_decode_sparse_mla_dsv4` as a first-class attention
routine in the repository's canonical `benchmarks/flashinfer_benchmark.py`
driver. It uses the public FlashInfer API and the driver's existing backend
filtering, common CLI flags, `bench_gpu_time`, cold-L2 handling, CSV schema,
test-list execution, and generated reproduction commands. Correctness is
checked before timing against an independent sampled FP32 attention oracle.

Native benchmark contract:

- 1x NVIDIA B200; PyTorch `2.13.0+cu130`; FlashInfer `0.6.16.post3`.
- Batch 1, 16 Q heads, head dim 512, FP8 E4M3 Q/KV, BF16 output.
- KV sequence length 16,384; 128 SWA + 1,920 compressed entries per query
  (`sparseMlaTopK=2048`); HND cache layout; PDL disabled.
- Query lengths `1,8,16,17,32,256,4096,16384` cover both sides of the guard
  and the target long-prefill case.
- Canonical `bench_gpu_time`: cold L2, 20 untimed warmups and 100 measured
  iterations per case. `cupti-python` was absent in the application image, so
  the driver used its documented CUDA-event fallback.
- Sampled FP32 reference checking ran before timing for every case. All BF16
  outputs were finite and the maximum absolute error was `1.81e-4`.

From the FlashInfer repository root, the target 16K case can be reproduced
with the canonical benchmark driver as follows:

```bash
python3 benchmarks/flashinfer_benchmark.py \
  --routine trtllm_batch_decode_sparse_mla_dsv4 \
  --backends trtllm-gen \
  --batch_size 1 \
  --s_qo 16384 \
  --s_kv 16384 \
  --num_qo_heads 16 \
  --num_kv_heads 1 \
  --head_dim_qk 512 \
  --head_dim_vo 512 \
  --page_size 256 \
  --swa_topk 128 \
  --compressed_kv_len 4096 \
  --compressed_page_size 64 \
  --compressed_topk 1920 \
  --kv_layout HND \
  --q_dtype fp8_e4m3 \
  --kv_dtype fp8_e4m3 \
  --out_dtype bfloat16 \
  --no_cuda_graph \
  --refcheck \
  --dry_run_iters 20 \
  --num_iters 100 \
  --generate_repro_command \
  --case_tag dsv4-q16384 \
  -vv
```

For the full boundary sweep, repeat the same command with
`--s_qo 1, 8, 16, 17, 32, 256, 4096, 16384` individually. Parent and PR
measurements must be run from their respective source checkouts with a freshly
built `fmha_gen` host launcher; an installed AOT launcher can otherwise hide
the selector change.

| Q length | Parent tile | PR tile | Parent (ms) | PR (ms) | Speedup |
|---:|:---:|:---:|---:|---:|---:|
| 1 | Q8 | Q8 | 0.027616 | 0.027616 | 1.000x |
| 8 | Q8 | Q8 | 0.027648 | 0.027584 | 1.002x |
| 16 | Q8 | Q8 | 0.029696 | 0.029680 | 1.001x |
| 17 | Q8 | Q16 | 0.029696 | 0.025568 | 1.162x |
| 32 | Q8 | Q16 | 0.037920 | 0.029696 | 1.277x |
| 256 | Q8 | Q16 | 0.134144 | 0.080736 | 1.662x |
| 4,096 | Q8 | Q16 | 1.565744 | 0.825200 | **1.897x** |
| 16,384 | Q8 | Q16 | 4.931104 | 2.614256 | **1.886x** |

The canonical-run logs show that the kernel name stays Q8 through `Q=16` and
changes to Q16 at `Q=17`, exactly matching the new boundary. At the target
`Q=16384`, the native API result is a 46.99% latency reduction. The `Q<=16`
arms use the same Q8 selector path and agree within 0.23%.

### vLLM DeepSeek-V4 B200 benchmark

Hardware and software:

- 8x NVIDIA B200, TP=8
- DeepSeek-V4-Pro, FP8 KV cache, block size 256
- vLLM `0.1.dev19908+g728d3ad09`
- FlashInfer `0.6.16.post3`
- Image `vllm/vllm-openai:nightly-dev-x86_64-cu13.0.1-728d3ad`
  (SHA-256 `9d812de47e34568de2a945635dd29cf2d6784fee62e3a602283d2665a9459cac`)
- `FLASHINFER_MLA_SPARSE_DSV4`, prefill-query quantization enabled, FP4
  indexer cache enabled, MTP=3
- Concurrency 16, ISL=16,384, OSL=1, seed 42,
  `max_num_batched_tokens=16384`, prefix caching disabled

The server used the DeepSeek-V4 configuration with the workload changes above.
The relevant launch arguments were:

```bash
vllm serve /models/DeepSeek-V4-Pro \
  --served-model-name deepseek-ai/DeepSeek-V4-Pro \
  --trust-remote-code --tensor-parallel-size 8 \
  --kv-cache-dtype fp8 --block-size 256 --max-model-len 1048576 \
  --max-num-seqs 32 --max-num-batched-tokens 16384 \
  --attention-config '{"backend":"FLASHINFER_MLA_SPARSE_DSV4","use_prefill_query_quantization":true,"use_fp4_indexer_cache":true}' \
  --speculative-config '{"method":"mtp","num_speculative_tokens":3,"rejection_sample_method":"synthetic","synthetic_acceptance_length":2.49}' \
  --no-enable-flashinfer-autotune --no-enable-prefix-caching --enforce-eager
```

Each arm ran four unprofiled repeats of 16 requests. The first repeat used 16
warmup requests; the server remained warm for the other repeats. The client
command was:

```bash
python utils/bench_serving/benchmark_serving.py \
  --model deepseek-ai/DeepSeek-V4-Pro \
  --served-model-name deepseek-ai/DeepSeek-V4-Pro \
  --backend vllm --base-url http://127.0.0.1:8888 \
  --dataset-name random --random-input-len 16384 --random-output-len 1 \
  --random-range-ratio 1.0 --max-concurrency 16 --request-rate inf \
  --num-prompts 16 --ignore-eos --seed 42 --trust-remote-code \
  --tokenizer-mode deepseek_v4 --use-chat-template --dsv4
```

Q8 is the parent selector. Q16 is the same image and workload with only this
query-length guard applied and the AOT host launcher rebuilt.

| Metric | Q8 (parent) | Q16 (this PR) | Change |
|---|---:|---:|---:|
| Four throughput repeats (tok/s) | 25,292.38; 25,286.59; 25,308.48; 25,294.95 | 27,892.23; 27,939.74; 27,923.39; 27,938.74 | -- |
| Mean total throughput | 25,295.60 tok/s | 27,923.53 tok/s | **+10.39%** |
| Mean throughput per GPU | 3,161.95 tok/s | 3,490.44 tok/s | **+10.39%** |
| Exact main sparse-MLA kernel | 142.38 ms/step | 75.89 ms/step | **1.88x faster; -46.70%** |
| Overlap-accounted attention | 198.72 ms/step | 132.66 ms/step | **-33.24%** |
| Iteration wall time | 648.73 ms/step | 589.02 ms/step | **-9.20%** |
| Main-kernel grid | `(16384, 2, 1)` | `(16384, 1, 1)` | **2x fewer CTAs** |

The throughput coefficient of variation was 0.037% for Q8 and 0.079% for Q16.
The wall-time ratio implies +10.14% throughput, consistent with the independent
unprofiled +10.39% result.

A 64-step profile validated the selector on all eight ranks. Per rank, all
`62 layers * 64 steps = 3968` main sparse-MLA calls used the Q16 kernel, with
zero main-path Q8 calls. The two short MTP calls per step remained Q8
(`128` calls per rank), as intended.

The Q8 and Q16 throughput arms used different B200 nodes. The image, framework,
model, and workload were identical; non-attention categories agreed within
about 2.4%, and the exact main FMHA reduction (66.48 ms/step) accounts for the
full raw-attention reduction (66.10 ms/step). A same-node rerun would provide
the strictest hardware-controlled A/B.

## 🔍 Related Issues

N/A

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

Validation performed:

- Canonical `benchmarks/flashinfer_benchmark.py` public-API benchmark on one
  B200: 20 warmups + 100 measured iterations for each of eight Q lengths and
  both selector variants; all sampled FP32 correctness checks passed.
  Long-prefill Q16 was 1.886x faster at `Q=16384` with cold L2.
- 4x unprofiled B200 throughput repeats per arm: passed, 16/16 requests each.
- 64-step, eight-rank B200 selector profile: Q16 on all 3968 main-layer calls
  per rank; short MTP remained Q8.

## Reviewer Notes

- This deliberately changes only the host-side tile selector. It does not
  modify numerical code or cubin artifacts.
- The `<= 16` boundary matches the existing short-query/DSv3 min-latency
  boundary in this runner and covers decode plus the configured MTP length.
- The benchmarked package used an AOT host launcher, so the validation rebuilt
  that launcher after editing the header. Normal source/CI builds will compile
  the updated selector into the launcher.

## Summary by CodeRabbit

* **New Features**
  * Added benchmarking support for DeepSeek-V4 sparse MLA attention with configurable sparse-window, compressed-cache, paging, and cache-length settings.
  * Added performance metrics, support-matrix coverage, documentation, and sample configurations for supported hardware and the TensorRT-LLM generation backend.
* **Bug Fixes**
  * Improved sparse MLA tile selection for long-query prefill workloads.
* **Tests**
  * Added validation for availability, configuration, tensor layouts, execution behavior, and reported metrics.

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -36,6 +36,7 @@ Currently supports testing attention, gemm, fused MOE, normalization, quantizati
         - Also supports computationally similar `cudnn_batch_prefill_with_kv_cache` (cudnn-native) and  `trtllm_ragged_attention_deepseek`.
     - `BatchMLAPagedAttentionWrapper` - MLA attention proposed in DeepSeek series of models.
         - Also supports computationally similar `trtllm_batch_decode_with_kv_cache_mla` (trtllm-native) and CuTe DSL MLA decode kernel (cute-dsl, SM100+).
+    - `trtllm_batch_decode_sparse_mla_dsv4` - DeepSeek-V4 sparse MLA using the public TRTLLM-GEN API on SM100/SM103. Supports varlen prefill-style query lengths, causal SWA and compressed-cache sparse tables, FP8/BF16 inputs, sampled FP32 reference checking, and hot-path Q-tile selector benchmarks.
     - All four attention routines accept `--backends prims-ts` on SM100/SM103 to benchmark the experimental task-scheduled attention implementation. `prims_ts` is accepted as an alias.
 - GEMM:
     - `gemm_fp8_nt_groupwise` - GEMM with FP8 data types using groupwise scaling.
@@ -239,6 +240,11 @@ The output CSV will contain detailed metrics including:
 | `--out_dtype`            | Data type for the output tensor. Default: same as q_dtype. Backend-dependent; PrimTS context accepts bfloat16, float16, or fp8_e4m3, while PrimTS FP8 decode accepts float16 or fp8_e4m3. FP8 ragged comparisons with non-PrimTS backends require bfloat16 or float16. |
 | `--causal`               | Use causal attention masking for context/prefill. Multi-query FMHA and MLA decode use bottom-right causal masking automatically. |
 | `--random_actual_seq_len`| Use random sequence lengths up to max length. If False, use max length.                                    |
+| `--swa_topk`             | DSV4 sparse MLA only: sliding-window segment width. Must be 128.                                           |
+| `--compressed_topk`      | DSV4 sparse MLA only: maximum compressed-cache rows selected per query. Default: 1920.                    |
+| `--compressed_kv_len`    | DSV4 sparse MLA only: compressed-cache rows per request. Default: `ceil(s_kv / 4)`.                       |
+| `--compressed_page_size` | DSV4 sparse MLA only: compressed-cache page size. Default: 64.                                             |
+| `--kv_layout`            | DSV4 sparse MLA only: `HND` (default) or `NHD` for both KV-cache pools.                                    |
 
 ### GEMM Flags
 | Flag                     | Description                                                                                                 |
@@ -526,6 +532,7 @@ Legend:
 | **BatchPrefillWithPagedKVCacheWrapper** |  | fa2, cudnn, cudnn-native | fa2, cudnn, cudnn-native | fa2, cudnn, cudnn-native | fa2, fa3, cudnn, cudnn-native | fa2, cudnn, cudnn-native, trtllm-gen, trtllm-native, prims-ts | fa2, cudnn, cudnn-native, trtllm-gen, trtllm-native, prims-ts | fa2, cudnn, cudnn-native |
 | **BatchPrefillWithRaggedKVCacheWrapper** |  | fa2, cudnn, cudnn-native | fa2, cudnn, cudnn-native | fa2, cudnn, cudnn-native | fa2, fa3, cudnn, cudnn-native | fa2, cudnn, cudnn-native, cutlass, trtllm-native, prims-ts | fa2, cudnn, cudnn-native, cutlass, trtllm-native, prims-ts | fa2, cudnn, cudnn-native |
 | **BatchMLAPagedAttentionWrapper** |  | fa2 | fa2 | fa2 | fa2, fa3 | fa2, cutlass, trtllm-native, cute-dsl, prims-ts | fa2, cutlass, trtllm-native, prims-ts | fa2 |
+| **trtllm_batch_decode_sparse_mla_dsv4** |  |  |  |  |  | trtllm-gen | trtllm-gen |  |
 | **gemm_fp8_nt_groupwise** |  |  |  |  |  | cutlass | cutlass |  |
 | **group_gemm_fp8_nt_groupwise** |  |  |  |  |  | cutlass | cutlass |  |
 | **bmm_fp8** |  |  |  | cudnn, cublas | cudnn, cublas | cudnn, cublas, cutlass | cudnn, cublas, cutlass | cudnn, cublas |
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -37,7 +37,7 @@ Currently supports testing attention, gemm, fused MOE, normalization, quantizati
     - `BatchMLAPagedAttentionWrapper` - MLA attention proposed in DeepSeek series of models.
         - Also supports computationally similar `trtllm_batch_decode_with_kv_cache_mla` (trtllm-native) and CuTe DSL MLA decode kernel (cute-dsl, SM100+).
     - `trtllm_batch_decode_sparse_mla_dsv4` - DeepSeek-V4 sparse MLA using the public TRTLLM-GEN API on SM100/SM103. Supports varlen prefill-style query lengths, causal SWA and compressed-cache sparse tables, FP8/BF16 inputs, sampled FP32 reference checking, and hot-path Q-tile selector benchmarks.
-    - All four attention routines accept `--backends prims-ts` on SM100/SM103 to benchmark the experimental task-scheduled attention implementation. `prims_ts` is accepted as an alias.
+    - All four wrapper attention routines above accept `--backends prims-ts` on SM100/SM103. The standalone `trtllm_batch_decode_sparse_mla_dsv4` routine supports only `trtllm-gen`.
 - GEMM:
     - `gemm_fp8_nt_groupwise` - GEMM with FP8 data types using groupwise scaling.
     - `group_gemm_fp8_nt_groupwise` - Group GEMM with FP8 data types using groupwise scaling.
--- a/benchmarks/routines/attention.py
+++ b/benchmarks/routines/attention.py
@@ -1,4 +1,5 @@
 from collections import defaultdict
+import math
 
 import numpy as np
 import torch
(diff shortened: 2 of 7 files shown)


benchmarks/routines/attention.py
```diff
@@ -3773,3 +3823,429 @@ def run_timed_backend(
...
+            ),
+            dim=0,
+        ).float()
+        scores = torch.matmul(query[query_row].float(), selected_kv.T) * scale_qk
+        expected = torch.matmul(torch.softmax(scores, dim=-1), selected_kv) * scale_out
+        actual = output[query_row].float()
+        max_abs_error = max(
+            max_abs_error, float((actual - expected).abs().max().item())
+        )
+        torch.testing.assert_close(actual, expected, rtol=1e-1, atol=1e-1)
```
saltyminty: Can be treated as non-blocking, but I believe `1e-1` rtol/atol is abnormal/too tolerant for bf16.

benchmarks/README.md
```diff
@@ -36,6 +36,7 @@ Currently supports testing attention, gemm, fused MOE, normalization, quantizati
         - Also supports computationally similar `cudnn_batch_prefill_with_kv_cache` (cudnn-native) and  `trtllm_ragged_attention_deepseek`.
     - `BatchMLAPagedAttentionWrapper` - MLA attention proposed in DeepSeek series of models.
         - Also supports computationally similar `trtllm_batch_decode_with_kv_cache_mla` (trtllm-native) and CuTe DSL MLA decode kernel (cute-dsl, SM100+).
+    - `trtllm_batch_decode_sparse_mla_dsv4` - DeepSeek-V4 sparse MLA using the public TRTLLM-GEN API on SM100/SM103. Supports varlen prefill-style query lengths, causal SWA and compressed-cache sparse tables, FP8/BF16 inputs, sampled FP32 reference checking, and hot-path Q-tile selector benchmarks.
     - All four attention routines accept `--backends prims-ts` on SM100/SM103 to benchmark the experimental task-scheduled attention implementation. `prims_ts` is accepted as an alias.
```
saltyminty: NIT: should be updated "four" -> "five"

saltyminty (approved):
Left two minor comments

saltyminty:
/bot run tests/attention benchmarks/test_flashinfer_benchmark.py

benchmarks/routines/attention.py
emascarenhas-nv: Makes sense, updated it to `1e-2`.

benchmarks/README.md
emascarenhas-nv: trtllm_batch_decode_sparse_mla_dsv4 only supports trtllm-gen, not prims-ts. I’ll clarify that “four” refers to the four wrapper routines.

PR #4755, opened 2026-08-26
feat(prims_ts): accept SM107 (Rubin) in the attention-ts device gate

Vinnie6167:
## 📌 Description

`attention-ts` (the task-scheduled attention path under `flashinfer/attention/prims_ts/`) rejects compute capability **10.7 (Rubin / SM107)** outright, in both the context and decode entry points:

```
NotImplementedError: attention-ts context requires an SM100a/B200 or SM103a/B300 GPU;
    device cuda:0 has compute capability (10, 7)
```

Because `_validate_device` runs early in `plan()`, this refusal short-circuits **all** downstream validation. The most visible effect is on the negative-path tests: `test_attention_ts_context_plan_rejects_critical_public_contracts` feeds deliberately-invalid inputs (bad dtype, bad `head_dim`, bad Q/KV head ratio, bad packed offset, `window_left` without a causal mask) and asserts each is rejected with its specific message. On Rubin those assertions never see their expected error — they get the arch rejection instead.

### Why the exclusion looks conservative rather than technical

Three signals, all pointing the same way:

1. **No stated rationale.** The tuple carries no explaining comment, unlike the comparable `_TINYGEMM2_SM100_SUPPORTED_COMPUTE_CAPABILITIES` in `gemm/routergemm.py`, which documents exactly why other 10.x parts must use the reference kernel.
2. **No `native_only` declaration.** `is_cute_dsl_arch_supported(..., native_only=True)` exists precisely to mark kernels that need architecture-specific `tcgen05` instructions (the block-scaled MMA kinds the DSL only accepts for `sm_100a`/`sm_103a`). Only two call sites use it, and `prims_ts` is not one of them.
3. **Only family-portable `tcgen05` usage.** The kernels call `tcgen05_alloc` / `tcgen05_dealloc` / `relinquish_alloc_permit` — tensor-memory allocation primitives that are valid across the SM100 family, including 10.7.

### What this changes

- Adds `(10, 7)` to `_SUPPORTED_COMPUTE_CAPABILITIES` in `context.py` and `decode.py`, and updates the diagnostic text to name SM107/Rubin.
- Gates the new capability on the installed CuTe DSL. Rubin runs these kernels through the `sm_100f` family target, and a DSL older than 4.8 has no `sm_107a` member in `cutlass.base_dsl.arch.Arch`. Without the gate, widening the tuple would swap one failure for a worse one — `KeyError: 'sm_107a'` raised from deep inside kernel compilation. `require_cute_dsl_arch` instead produces an actionable message naming `CUTE_DSL_ARCH=sm_100f`. The import is deferred so the module keeps its current import-time behaviour.
- Updates two test regexes that pinned the old wording. Both tests monkeypatch the capability to `(9, 0)`, so they continue to exercise the rejection path; only the message pattern changes.

No behaviour change on any non-107 device: every edit either adds 10.7 to a tuple or widens a message.

## 🔍 Related Issues

None filed publicly.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

Measured on SM107 hardware, `tests/attention/test_attention_ts_context.py`:

| | `..._plan_rejects_critical_public_contracts` |
|---|---|
| before | 3 failed, 2 passed |
| after | **5 passed** |

(Two of the five passed even with the gate closed, because their validation happens to run *before* the device check — the ordering is not uniform, which is why only three were affected.)

`pre-commit run` passes on all four changed files, including `mypy`, `ruff check` and `ruff format`.

## Reviewer Notes

**This fixes the device gate, not attention-ts on Rubin — please read before merging.**

`plan()` only validates inputs and builds scheduling metadata; it never compiles a kernel, which is why the context-path contract tests go green. The **decode** path is a different story: `tests/attention/test_attention_ts_decode.py` is unchanged at **4 failed / 45 passed / 76 skipped both before and after** this change. Those four fail earlier, at kernel construction, in the exhaustive deadlock/race checker:

```
ValueError: Exhaustive checker found 1 aliasing race(s) after exploring 121713 states:
    Softmax0Task writes tmemSoftmaxLocal0 vs MmaTask prod tmemS0 (tmem[256:288])
```

That failure is **not Rubin-specific** — it reproduces on a real B200 (compute capability 10.0) and on SM107 with entirely unmodified code. `exhaustive_deadlock_race_check` defaults to `True` and runs host-side at kernel construction; it never queries the device. It also looks like a known false-positive class: `fmha_decode/fmha_decode_kernel.py` already carries a `TODO` noting that the checker "currently sees their aggregate allocation and reports false races" for D256 configurations with two physical S/stats/P stages, and the reported offset `tmem[256:288]` points at D256.

So the honest trade-off for the decode path is that this replaces a clear `NotImplementedError` with an opaque checker `ValueError`. Reviewers may reasonably prefer to land this only once that checker issue is resolved. The context path benefits either way.

## Summary by CodeRabbit

* **New Features**
  * Added support for Rubin (SM107a) GPUs in attention context and decode operations.
  * Added compatibility checks requiring CuTe DSL 4.8 or newer for Rubin devices.

* **Bug Fixes**
  * Updated unsupported GPU error messages to accurately reflect supported architectures.
  * Updated validation tests for the revised architecture guidance.


Vinnie6167:
@coderabbitai review

Vinnie6167:
@flashinfer-bot run

Vinnie6167:
@flashinfer-bot run

Vinnie6167:
@flashinfer-bot run

saltyminty (approved):
> Reviewers may reasonably prefer to land this only once that checker issue is resolved. The context path benefits either way.

Should be fine to merge, but is there a follow-up issue/bug tracking this?

Vinnie6167:
> > Reviewers may reasonably prefer to land this only once that checker issue is resolved. The context path benefits either way.
> 
> Should be fine to merge, but is there a follow-up issue/bug tracking this?

It was found to be a false alarm

Vinnie6167:
@flashinfer-bot run

Vinnie6167:
@flashinfer-bot run

Vinnie6167:
@flashinfer-bot run

PR #4769, opened 2026-08-27
perf(attention): repack NVFP4 KV tiles to 16-bit on targets without a native E2M1 convert

lesj0610:
## 📌 Description

FP8 KV already dequantizes a whole tile into the 16-bit staging buffer and lets the QK/PV MMAs read it back with the native `ldmatrix` path (`repack_fp8_tile_to_bf16`, `KernelTraits::USE_KV_REPACK`). NVFP4 was excluded from that path and kept dequantizing inside the fully unrolled `(mma_d, mma_kv)` nest, so its dequant sequence — magnitude table, sign spread, scale-factor gather, scale multiply, plus two `__shfl_sync` per fragment register for the 4-bit fragment swizzle — was replicated once per pair.

On targets without a native E2M1 conversion that turns the kernel instruction-fetch bound rather than math bound. `ncu` on FA2 paged prefill, batch 1, 2048x2048, head_dim 128, sm_80:

| | NVFP4 before | FP8 | NVFP4 after |
|---|---|---|---|
| `no_instruction` stall, % of warp-active | 24.9 | 1.4 | 1.3 |
| `sm__pipe_tensor_cycles_active`, % of peak | 18.7 | 51.5 | 48.6 |
| kernel time | 1.403 ms | 0.539 ms | 0.552 ms |

`repack_fp4_tile_to_16b` dequantizes one K or V tile, scale factors included, into the same staging layout. One thing makes it cheaper than the in-loop path it replaces, beyond amortizing the work over the tile: one padded FP4 b128 spans exactly `HEAD_DIM` 16 == `NVFP4_SF_VEC_SIZE` elements, i.e. one scale-factor group, so the per-fragment gather of two scale factors plus a four-wide E4M3 convert collapses to one byte load and one convert per b128. The cross-lane fragment swizzle disappears with it.

### Architecture scope

The repack only makes sense where the E2M1 conversion is a software sequence. From SM100 it is a single instruction, the premise above does not hold, and the staging buffer would cost shared memory -- and a smaller `NUM_MMA_KV`, since the occupancy budget counts it -- for nothing.

That decision has to be visible to the host, which sizes the storage and picks `NUM_MMA_KV`, not just to the device. `__CUDA_ARCH__` is not, but `__CUDA_ARCH_LIST__` is: it names every target of the module in both passes. So the policy is a template parameter, `ENABLE_FP4_REPACK`, threaded through one predicate:

```cpp
template <typename DTypeKV, uint32_t CTA_TILE_Q, uint32_t HEAD_DIM_QK, uint32_t HEAD_DIM_VO>
constexpr bool use_kv_repack(bool enable_fp4_repack);
```

`KVRepackSmem`, `KernelTraits::USE_KV_REPACK` and the `kUseRepack` in all three launcher budgets now go through it, so they cannot disagree. Each launcher body became `...DispatchedImpl<..., ENABLE_FP4_REPACK>` and the public entry point is a thin wrapper that picks the variant before any budget arithmetic runs.

- a module built only for pre-SM100 targets folds to one variant, with the repack;
- a module built only for SM100+ folds to one variant, with no staging buffer and no budget for it;
- a module spanning both (the release wheel bundles SM7.5 through SM12.x) builds both and selects on the device's compute capability. The device query happens only on that path, so FP16, FP8 and single-regime FP4 launches gain no CUDA API call.

Measured on the FP4 paged prefill translation unit, `head_dim` 128:

| built for | kernel entries | software dequant `prmt` | native e2m1 cvt |
|---|---|---|---|
| `sm_80` | 13 | 340 | 0 |
| `sm_100a` | 13 | 0 | 2240 |
| `sm_80` + `sm_100a` | 36 | 740 | 2240 |

and the storage and budget follow, at `head_dim` 128 with `NUM_WARPS_KV` 1 and a 2-byte query dtype:

| variant | `use_kv_repack` | `kKVSmemPerMmaKV` | `sizeof(SmemStorage)` |
|---|---|---|---|
| repack | 1 | 8448 | 41216 |
| native | 0 | 4352 | 37120 |

The 4096-byte difference is the staging buffer: on a native-only build it is neither allocated nor charged.

The repack condition also gained a `HEAD_DIM_QK` bound for FP4. The staging buffer is sized `max(HEAD_DIM_QK, HEAD_DIM_VO)`, so an asymmetric shape such as `HEAD_DIM_QK` 480 with `HEAD_DIM_VO` 128 would need more shared memory than the in-loop path it replaces and leave no launchable `NUM_MMA_KV`.

Single prefill dispatches `CTA_TILE_Q` inside the Impl, so its eligibility probe asks whether any CTA tile in the specialization could be repack-eligible rather than whether a particular one is. A mixed-architecture build therefore emits both policy variants for that whole specialization, including its `CTA_TILE_Q` 16 kernels.

### Stacking

Both prerequisites have landed. #4746 brought in the `e2m1x8_to_f16x8` / `e2m1x8_to_bf16x8` helpers and `nvfp4_sf4_to_packed2x2` that the repack calls, and #4767 the scale-factor budget the corrected repack condition depends on. Rebasing onto `main` dropped both by patch id with nothing to resolve by hand, so this PR is now the repack and its policy default: two commits touching only `include/flashinfer/attention/prefill.cuh`.

Earlier revisions of this section described those prerequisites while they were still open, and one of them claimed that reviewing the last commit alone was enough. Neither applies now.

## 🔍 Related Issues

N/A

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

- [ ] Tests have been added or updated as needed. — the repack is an alternative lowering of an existing computation, covered by the existing NVFP4 prefill/decode tests; the checks below are what I ran to establish it is lossless.
- [ ] All tests are passing (`unittest`, etc.) — I ran the NVFP4 subset on sm_80, not the full suite.

```bash
pytest -q tests/attention/test_batch_prefill_kernels.py \
          tests/attention/test_batch_decode_kernels.py \
          tests/attention/test_single_prefill.py -k "nvfp4 or fp4"
# 355 passed, 9 skipped
```

Re-run after the per-architecture variant work, same result: 355 passed, 9 skipped.

Output equivalence against FA2 attention run on the dequantized KV in the query dtype, so kernel and reference consume the same values. This covers the FP4 payload path end to end; the scale-factor conversion is a separate path and is scoped below:

```
head_dim 128  f16  maxabs 0        head_dim 128  bf16  maxabs 0
head_dim 256  f16  maxabs 0        head_dim 256  bf16  maxabs 0
head_dim 512  f16  maxabs 0        head_dim 512  bf16  maxabs 0
```

Against the previous NVFP4 build, head_dim 128 and 512 are bit-identical and head_dim 256 prefill is not: the corrected budget makes the chooser pick a `NUM_MMA_KV` that actually fits two blocks per SM, which changes the flash-attention accumulation order. The reference comparison above is what settles which side is closer — at head_dim 256 the previous build differs from the reference by up to 1.95e-3 (f16) and this one by 0.

Latency on sm_80, median of three, one process per dtype, FA2 paged:

| case | NVFP4 before | NVFP4 after | FP8 | FP16 |
|---|---|---|---|---|
| prefill b1 4096x4096 hd128 | 4.763 | 1.900 | 1.858 | 1.516 |
| prefill b4 2048x2048 hd128 | 4.427 | 1.840 | 1.857 | 1.546 |
| prefill b8 1024x1024 hd128 | 2.407 | 0.994 | 0.983 | 0.884 |
| prefill b2 2048 hd256 | 3.790 | 1.254 | 1.209 | 0.910 |
| decode b32 kv8192 hd128 | 1.154 | 1.036 | 0.718 | 0.848 |
| decode b16 kv4096 hd512 | 1.114 | 0.612 | n/a | 0.398 |

head_dim 512 keeps the in-loop path (`HEAD_DIM_VO <= 256`); its gain comes from the sign-spread commit underneath.

Scale-factor contract. `nvfp4_sf4_to_packed2x2` arrives with #4746 and is unchanged by the repack commits, but the repack calls it, so its contract applies here too and is now written down at the declaration. Enumerating all 256 E4M3 encodings, the software and hardware conversions agree on 254 and differ only on the NaN encodings `0x7F` and `0xFF`, which the software converter maps to +-480 instead of NaN:

| byte | software | hardware |
|---|---|---|
| `0x7F` | +480 | NaN |
| `0xFF` | -480 | NaN |

Scale factors are expected finite. Quantization does not produce a NaN scale, and the only way one reaches the kernel is an out-of-range KV row, whose scores `logits_mask` replaces with `-inf` before they can reach the accumulator. The repack does not widen that exposure: it converts the same scale factors the in-loop path converted, with the same helper. Callers supplying an externally built `kv_cache_sf` that can contain those encodings should not expect NaN to propagate.

Review follow-ups in this branch:

- the E2M1 nibble helpers moved out of the `FLASHINFER_ENABLE_FP4_E2M1` guard. They have no FP4 type in their signatures, and `repack_fp4_tile_to_16b` names them from a template whose FP4 branch is discarded rather than removed, so a build without FP4 enabled failed to compile. Checked `sm_80` and `sm_100a`, with and without the macro.
- `clang-format` applied (the pinned 19.1.1), covering both spots the automated check flagged.
- the repack policy became a template parameter selected per architecture regime, so a module built only for SM100+ neither allocates the staging buffer nor charges `NUM_MMA_KV` for it; see Architecture scope.
- the single-prefill kernel now takes the repack for NVFP4 as well. It shares `SharedStorage` with the ragged kernel, so whenever the variant allocates the staging buffer it is already counted in that kernel's `NUM_MMA_KV` budget; NVFP4 was paying for it without using it. On a native-only build there is no staging buffer to pay for. FP8 keeps the in-loop path it has always taken in that kernel: switching it over is a behavior change this PR does not measure, and `tests/attention/test_single_prefill.py` has no FP8 KV case that would catch a regression.

The `sm_80` code path is unchanged by the variant work in the sense that matters -- it is the `ENABLE_FP4_REPACK = true` instantiation, and its dequant/repack instruction sequences are the ones measured above. Whole-cubin byte identity is not the right oracle here: the storage type and its size differ per variant, so the comparison is per-path, plus the functional checks below.

Additional runs for the single-prefill change and for the shared-memory budget arithmetic:

```bash
pytest -q tests/attention/test_single_prefill.py \
       tests/attention/test_batch_prefill_kernels.py::test_batch_prefill_paged_cta_tile_q_smem_probe_qk448_vo256
# 207 passed, 13 skipped
```

The second one pins the `FA2DetermineCtaTileQ` probe at `(head_dim_qk, head_dim_vo) = (448, 256)` for both a 2-byte and a 1-byte KV dtype, which is the shape class the corrected budget and the new `HEAD_DIM_QK` bound act on.

Not verified: the mixed-module runtime selection on an SM100+ device, and latency on sm_89 and sm_90 -- no hardware for either. The numbers above are local measurements; the fork's CI test matrix is permission-gated and did not run them. Both compile, and `ptxas -v` shows register spilling on the FP4 prefill kernels drops there too — 424 B to 28 B of spill stores on sm_89, 376 B to 28 B on sm_90a, against 392 B to 8 B on sm_80.

## Summary by CodeRabbit

* **New Features**
  * Added optimized NVFP4 support for single, ragged, and paged prefill attention workflows.
  * Added automatic GPU architecture-aware selection of FP4 data repacking.
  * Added FP4 scale conversion and tile dequantization for attention processing.
  * Added faster FP4 conversion to FP16 and BF16 for supported vector sizes.
* **Compatibility**
  * Preserved FP8 behavior and fallback conversion paths across supported configurations.
  * Added support for consistent operation across older and newer GPU architectures.

### DIFF
--- a/include/flashinfer/attention/prefill.cuh
+++ b/include/flashinfer/attention/prefill.cuh
@@ -69,6 +69,46 @@ constexpr uint32_t WARP_SIZE = 32;
 // Number of NVFP4 elements sharing one scale factor (UE4M3 byte).
 constexpr uint32_t NVFP4_SF_VEC_SIZE = 16;
 
+// Which E2M1-conversion regimes this translation unit is being built for.
+//
+// The NVFP4 tile repack exists to keep the software E2M1 -> 16-bit conversion out of the fully
+// unrolled MMA loops, where its instruction footprint starves the fetch pipe. From SM100 that
+// conversion is a single instruction, the premise does not hold, and the staging buffer would cost
+// shared memory -- and a smaller NUM_MMA_KV, since the occupancy budget counts it -- for nothing.
+//
+// The decision therefore has to be visible to the host, which sizes the storage and picks
+// NUM_MMA_KV, not just to the device. `__CUDA_ARCH__` is not, but `__CUDA_ARCH_LIST__` is: it holds
+// every target of this module in both passes. A module built for one regime hard-codes the answer
+// and instantiates a single variant; a module built for both (the release wheel bundles SM7.5
+// through SM12.x) instantiates both and the launcher picks by the device's compute capability.
+#if defined(__CUDA_ARCH_LIST__)
+namespace fp4_repack_targets {
+constexpr int kArchs[] = {__CUDA_ARCH_LIST__};
+constexpr bool any_below(int bound) {
+  for (int arch : kArchs) {
+    if (arch < bound) return true;
+  }
+  return false;
+}
+constexpr bool any_at_least(int bound) {
+  for (int arch : kArchs) {
+    if (arch >= bound) return true;
+  }
+  return false;
+}
+}  // namespace fp4_repack_targets
+// Targets that need the software conversion, i.e. that the repack is for.
+constexpr bool kTargetsSoftwareE2M1 = fp4_repack_targets::any_below(1000);
+// Targets whose E2M1 conversion is a single instruction, i.e. that must not pay for the repack.
+constexpr bool kTargetsNativeE2M1 = fp4_repack_targets::any_at_least(1000);
+#else
+// No arch list (host-only or non-nvcc translation): assume both, which forces the runtime choice.
+constexpr bool kTargetsSoftwareE2M1 = true;
+constexpr bool kTargetsNativeE2M1 = true;
+#endif
+// True when one instantiation cannot serve every target of this module.
+constexpr bool kFp4RepackNeedsRuntimeChoice = kTargetsSoftwareE2M1 && kTargetsNativeE2M1;
+
 /*!
  * \brief Convert four NVFP4 scale-factor bytes into two packed 16-bit pairs.
  *
@@ -83,6 +123,14 @@ constexpr uint32_t NVFP4_SF_VEC_SIZE = 16;
  * Where the hardware instruction exists we keep `static_cast` so it still lowers to it. The gate is
  * FLASHINFER_HARDWARE_FP8_CONVERSION_ENABLED -- the same one vec_cast<*, __nv_fp8_e4m3> uses -- so
  * this does not introduce a second, divergent notion of "has FP8 conversion".
+ *
+ * Scale factors are expected to be finite. Enumerating all 256 E4M3 encodings shows the two paths
+ * agree on 254 of them and differ only on the NaN encodings 0x7F and 0xFF, which the software
+ * converter maps to +-480 instead of NaN (a long-standing property of fast_dequant_f8f16x4, shared
+ * with the FP8 KV path). A NaN scale factor is not something quantization produces, and the only
+ * way one reaches here is an out-of-range KV row, whose scores logits_mask replaces with -inf
+ * before they can reach the accumulator. Callers passing an externally built kv_cache_sf that can
+ * contain those encodings should not expect NaN to propagate.
  */
 template <typename DTypeQ>
 __device__ __forceinline__ void nvfp4_sf4_to_packed2x2(uint8_t b0, uint8_t b1, uint8_t b2,
@@ -144,16 +192,31 @@ struct KVScaleFactorSmem<DTypeKV, CTA_TILE_KV, HEAD_DIM_QK, HEAD_DIM_VO, true> {
   alignas(16) uint8_t v_sf_smem[CTA_TILE_KV * HEAD_DIM_VO / NVFP4_SF_VEC_SIZE];
 };
 
-// BF16/FP16 staging buffer for the FP8 "repack" path (compute_qk REPACK_BF16). Empty base
-// (0 bytes via EBO) when unused; same smem rationale as KVScaleFactorSmem above.
+// BF16/FP16 staging buffer for the one-byte-KV "repack" path (compute_qk REPACK_BF16), used by
+// both FP8 and NVFP4. Empty base (0 bytes via EBO) when unused; same smem rationale as
+// KVScaleFactorSmem above.
+// The single repack policy. The storage layout, the kernel traits and the launchers' occupancy
+// budget all go through this, so they cannot drift apart -- which is what would silently reserve
+// (and charge NUM_MMA_KV for) a staging buffer the device never reads.
+//
+// ENABLE_FP4_REPACK is the per-variant answer to "does this instantiation serve targets whose E2M1
+// conversion is a software sequence"; see kTargetsSoftwareE2M1 above. It only gates FP4: FP8 keeps
+// the policy it has always had.
+template <typename DTypeKV, uint32_t CTA_TILE_Q, uint32_t HEAD_DIM_QK, uint32_t HEAD_DIM_VO>
+constexpr bool use_kv_repack(bool enable_fp4_repack) {
+  return (sizeof(DTypeKV) == 1) && (HEAD_DIM_VO != 64) && (HEAD_DIM_VO <= 256) &&
+         (CTA_TILE_Q > 16) &&
+         (!is_fp4_type_v<DTypeKV> || (HEAD_DIM_QK <= 256 && enable_fp4_repack));
+}
+
 template <typename DTypeQ, typename DTypeKV, uint32_t CTA_TILE_Q, uint32_t CTA_TILE_KV,
-          uint32_t HEAD_DIM_QK, uint32_t HEAD_DIM_VO,
-          bool = ((sizeof(DTypeKV) == 1) && !is_fp4_type_v<DTypeKV> && (HEAD_DIM_VO != 64) &&
-                  (HEAD_DIM_VO <= 256) && (CTA_TILE_Q > 16))>
+          uint32_t HEAD_DIM_QK, uint32_t HEAD_DIM_VO, bool ENABLE_FP4_REPACK,
+          bool = use_kv_repack<DTypeKV, CTA_TILE_Q, HEAD_DIM_QK, HEAD_DIM_VO>(ENABLE_FP4_REPACK)>
 struct KVRepackSmem {};
 template <typename DTypeQ, typename DTypeKV, uint32_t CTA_TILE_Q, uint32_t CTA_TILE_KV,
-          uint32_t HEAD_DIM_QK, uint32_t HEAD_DIM_VO>
-struct KVRepackSmem<DTypeQ, DTypeKV, CTA_TILE_Q, CTA_TILE_KV, HEAD_DIM_QK, HEAD_DIM_VO, true> {
+          uint32_t HEAD_DIM_QK, uint32_t HEAD_DIM_VO, bool ENABLE_FP4_REPACK>
+struct KVRepackSmem<DTypeQ, DTypeKV, CTA_TILE_Q, CTA_TILE_KV, HEAD_DIM_QK, HEAD_DIM_VO,
+                    ENABLE_FP4_REPACK, true> {
   // Sized to max(HEAD_DIM_QK, HEAD_DIM_VO); time-shared between K and V within a tile.
   alignas(16) DTypeQ
       kv_smem_repack[CTA_TILE_KV * (HEAD_DIM_QK > HEAD_DIM_VO ? HEAD_DIM_QK : HEAD_DIM_VO)];
@@ -176,10 +239,11 @@ struct VOSplitSmem<NUM_WARPS_KV, CTA_TILE_Q, CTA_TILE_KV, HEAD_DIM_VO, DTypeQ, k
 
 template <uint32_t NUM_WARPS_KV, uint32_t CTA_TILE_Q, uint32_t CTA_TILE_KV, uint32_t HEAD_DIM_QK,
           uint32_t HEAD_DIM_VO, typename DTypeQ, typename DTypeKV, typename DTypeO,
-          bool kEnableVOSplitOpt = false>
+          bool ENABLE_FP4_REPACK = true, bool kEnableVOSplitOpt = false>
 struct SharedStorageQKVO
     : KVScaleFactorSmem<DTypeKV, CTA_TILE_KV, HEAD_DIM_QK, HEAD_DIM_VO>,
-      KVRepackSmem<DTypeQ, DTypeKV, CTA_TILE_Q, CTA_TILE_KV, HEAD_DIM_QK, HEAD_DIM_VO>,
+      KVRepackSmem<DTypeQ, DTypeKV, CTA_TILE_Q, CTA_TILE_KV, HEAD_DIM_QK, HEAD_DIM_VO,
+                   ENABLE_FP4_REPACK>,
       VOSplitSmem<NUM_WARPS_KV, CTA_TILE_Q, CTA_TILE_KV, HEAD_DIM_VO, DTypeQ, kEnableVOSplitOpt> {
   static constexpr bool kKVShareShape =
       (HEAD_DIM_VO / 16 > 16) && ((HEAD_DIM_VO / 16) % NUM_WARPS_KV == 0);
@@ -208,9 +272,8 @@ struct SharedStorageQKVO
     };
     alignas(16) DTypeO smem_o[CTA_TILE_Q * HEAD_DIM_VO];
   };
-  static constexpr bool USE_KV_REPACK = (sizeof(DTypeKV) == 1) && !is_fp4_type_v<DTypeKV> &&
-                                        (HEAD_DIM_VO != 64) && (HEAD_DIM_VO <= 256) &&
-                                        (CTA_TILE_Q > 16);
+  static constexpr bool USE_KV_REPACK =
+      use_kv_repack<DTypeKV, CTA_TILE_Q, HEAD_DIM_QK, HEAD_DIM_VO>(ENABLE_FP4_REPACK);
   static constexpr bool VO_SPLIT_SMEM = kVOSplit;
 
   // Accessors for the empty-base buffers above: return the buffer when present, else nullptr.
@@ -237,8 +300,8 @@ struct SharedStorageQKVO
   }
   __device__ __forceinline__ DTypeQ* kv_smem_repack_ptr() {
     if constexpr (USE_KV_REPACK) {
-      using Base =
-          KVRepackSmem<DTypeQ, DTypeKV, CTA_TILE_Q, CTA_TILE_KV, HEAD_DIM_QK, HEAD_DIM_VO, true>;
+      using Base = KVRepackSmem<DTypeQ, DTypeKV, CTA_TILE_Q, CTA_TILE_KV, HEAD_DIM_QK, HEAD_DIM_VO,
+                                ENABLE_FP4_REPACK, true>;
       return static_cast<Base*>(this)->kv_smem_repack;
     } else {
       return nullptr;
@@ -273,7 +336,7 @@ template <MaskMode MASK_MODE_, uint32_t CTA_TILE_Q_, uint32_t NUM_MMA_Q_, uint32
           uint32_t NUM_MMA_D_QK_, uint32_t NUM_MMA_D_VO_, uint32_t NUM_WARPS_Q_,
           uint32_t NUM_WARPS_KV_, PosEncodingMode POS_ENCODING_MODE_, typename DTypeQ_,
           typename DTypeKV_, typename DTypeO_, typename DTypeQKAccum_, typename IdType_,
-          typename AttentionVariant_>
+          typename AttentionVariant_, bool ENABLE_FP4_REPACK_ = true>
 struct KernelTraits {
   static constexpr uint32_t NUM_STAGES = 1;  // used for BatchAttention Template
   static constexpr MaskMode MASK_MODE = MASK_MODE_;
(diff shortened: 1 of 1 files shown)


lesj0610:
@qsang-nv, could you take a look at this one too?

It moves the pre-SM100 NVFP4 dequant work out of the unrolled MMA loops and repacks each KV tile once into the 16-bit staging layout. On the profiled SM80 shape, the kernel went from 1.403 ms to 0.552 ms; FP8 was 0.539 ms.

The dispatch and shared-memory sizing are the parts I am least comfortable leaving without another set of eyes. Single-regime builds instantiate one policy, while mixed-architecture builds instantiate both and choose at runtime. The launcher budget, KernelTraits, and staging allocation all use the same policy now. A native-E2M1-only build should neither allocate nor budget the repack buffer.

The SM80 NVFP4 run finished with 355 passed and 9 skipped. The architecture builds, SASS checks, and pre-commit also pass. I did not have hardware to exercise mixed-build selection on SM100+ or measure SM89/SM90.

qsang-nv:
Reviewed the full diff. I did not find a correctness problem in the repack, the dispatch, or the budget math.

One thing on scope: the Stacking section says reviewing the last commit alone is enough, but `main` currently has none of `use_kv_repack`, `nvfp4_sf4_to_packed2x2`, `repack_fp4_tile_to_16b`, `ENABLE_FP4_REPACK` or `e2m1x8_to_f16x8`, and there is no open PR for either prerequisite branch — so all seven commits land together here. Is that intended, or should the prerequisite commits be split into preceding PRs?

lesj0610:
You're right, and the Stacking note was simply wrong. It described the branch back when the two prerequisites were still open PRs. I closed them on Aug 27 after folding them in here and never fixed the note. Same for the line calling `nvfp4_sf4_to_packed2x2` unchanged by this PR: the third commit is what adds it.

So I've split it. #4746 and #4767 are open again, both rebased onto current `main`, and this branch now sits on top of them.

#4746 is the four dequant commits, +170/-24 in two files. It stands on its own for targets that lower E2M1 in software, and it is where `e2m1x8_to_f16x8`, `e2m1x8_to_bf16x8` and `nvfp4_sf4_to_packed2x2` come from.

#4767 is the budget fix, ten lines. It is not stacked on #4746, just independent of it: both batch launchers size their budget without the `KVScaleFactorSmem` planes, and those grow with `NUM_MMA_KV` the same way the K/V tiles do. It reproduces on `main` with no repack in the picture, so the two can go in either order.

What is left here is `repack_fp4_tile_to_16b` and the `ENABLE_FP4_REPACK` policy, the last two commits. The diff will keep showing all seven until the prerequisites land, and the description now says which two are mine.

Each prerequisite is carried here as exactly one commit, the shape a squash merge will give it, so rebasing this branch after either lands drops it by patch id instead of conflicting against the squashed version.

No code changed in any of this. All three branches are rebased onto 012cfdb9, every commit is patch-id identical to what you reviewed, and none of the rebases hit a conflict.

lesj0610:
@qsang-nv Sorry — I force-pushed here at 09:45, after your /bot run. MR !1363's pipeline is therefore running against a3703479, which is no longer the branch head, and the push also cancelled the GitHub Actions run for that commit. I should have waited for both.

No code changed: the new head has an identical tree and the same +529/-101 in two files, so your read of it stands. What changed is the commit shaping. The branch was carrying #4746 as its original four commits, and once #4746 lands as a squash a rebase here would have reapplied them and conflicted in vec_dtypes.cuh; I confirmed that against a simulated squash merge before pushing. Both prerequisites are now one commit each, matching the shape they will merge in, so a rebase after either one lands drops just that one, and once both are in only the repack and its policy default are left. Seven commits became four.

Could you re-issue /bot run tests/attention when convenient?

lesj0610:
@qsang-nv Rebased onto main now that #4746 has landed. The prerequisite dropped out on its own with nothing to resolve by hand, leaving three commits and +359/-77 in one file: the budget fix from #4767, the repack, and its policy default. vec_dtypes.cuh is no longer part of the diff, since #4746 brought it in.

The last pipeline here was red only on H100 / CUDA 12.9, with CUDA unknown error and FlashInfer requires GPUs with sm75 or higher across 36 test groups this branch does not touch. #4747 failed the same way in the same window, with matching failure counts in all but one group, while #4746 and #4767 passed 16/16.

qsang-nv:
Please resolve conflicts.

lesj0610:
@qsang-nv Rebased onto main, conflicts resolved. Two commits left: the repack and its policy default.

qsang-nv (approved):

lesj0610:
@qsang-nv Thanks for the approval. The merge is still blocked by the required Test Results Summary, which is red from an earlier run (33605098821). On the current run the AOT builds and the T4 and A10G jobs have all passed, with JIT Unittest (H100) still going.

TensorRaya:
@lesj0610 @qsang-nv — one data point from the architecture this PR switches the repack *off* for. Short version: on sm_120 the repack still buys about a third of the NVFP4 prefill time, and I think the reason is that the native `cvt` was never the expensive part on this shape.

**Hardware / build.** RTX PRO 4000 Blackwell (sm_120, 24 GB, 672 GB/s), driver 580.173.02, CUDA 13.0, torch 2.13.0+cu130. FlashInfer `main` at `83ce2dab`, built from source in two otherwise identical containers. The only difference is two lines in `prefill.cuh`:

```cpp
// upstream
constexpr bool kTargetsSoftwareE2M1 = fp4_repack_targets::any_below(1000);
constexpr bool kTargetsNativeE2M1   = fp4_repack_targets::any_at_least(1000);
// "force": treat sm_120 as a software-E2M1 target
constexpr bool kTargetsSoftwareE2M1 = true;
constexpr bool kTargetsNativeE2M1   = false;
```

So `use_kv_repack<>()` returns false on upstream (this PR's policy) and true on the forced build. JIT modules for `sm_120f`, same GPU, same process pattern, runs three minutes apart.

**Benchmark.** `BatchPrefillWithPagedKVCacheWrapper`, `backend="fa2"`, HND, causal, 6 QO heads / 1 KV head, `head_dim` 256, page 16, one request: `qo_len` new tokens appended to `kv_len` cached tokens (the prefix-cache-extend shape a serving stack produces). bf16 KV as reference. CUDA events, 2 warmup, mean of 5.

| kv_len | qo_len | bf16 ms | nvfp4 ms, repack off | nvfp4 ms, repack on | nvfp4/bf16 off | nvfp4/bf16 on | nvfp4 time saved |
| 32,768 | 8,192 | 16.1 | 30.8 | 20.8 | 1.92x | 1.27x | 32% |
| 131,072 | 8,192 | 76.5 | 141.3 | 94.0 | 1.85x | 1.21x | 33% |
| 262,144 | 8,192 | 157.2 | 288.3 | 189.3 | 1.83x | 1.25x | 34% |
| 400,000 | 8,192 | 242.8 | 431.8 | 292.5 | 1.78x | 1.25x | 32% |
| 400,000 | 32,768 | 904.2 | 1664.9 | 1155.2 | 1.84x | 1.27x | 31% |
| 131,072 | 131,072 | 687.1 | 1142.1 | 798.0 | 1.66x | 1.16x | 30% |

Same picture on 0.6.16.post3 with the repack backported (that is what we run in production): 1.87x → 1.33x at 400k. End to end in vLLM, TP4 on four of these cards, Qwen3.8-27B NVFP4 KV: cold 400k prefill 249 s → 196 s, 131k 47.6 s → 39.6 s, a 9k-token extend on a 160k prefix 5.5 s → 4.4 s.

**Why the native convert does not settle it here.** I compiled FlashInfer's own `vec_cast<nv_bfloat16, __nv_fp4x2_e2m1>` to PTX with this toolchain:

| target | E2M1 → 16-bit lowering |
|---|---|
| `sm_80`, `sm_90` | 40× `prmt.b32` (the table path) |
| `sm_100a`, `sm_120a` | 16× `cvt.rn.f16x2.e2m1x2`, then 24× f16 → bf16 |
| `sm_120` (no `a`) | ptxas: `Instruction 'cvt with .e2m1x2' not supported on .target 'sm_120'` |

Two things follow. First, the bf16 path on CUDA < 13.2 goes through f16 and converts twice — `cvt.rn.bf16x2.e2m1x2` only exists from 13.2, and the fallback in `vec_dtypes.cuh` is exactly that detour. Second, and I think more important: the repack does not mainly remove the convert. It removes the per-fragment work around it — SF gather, the two `__shfl_sync` for the 4-bit fragment swizzle, `__hmul2` — from the fully unrolled `(mma_d, mma_kv)` nest, and amortizes it once per tile. At `head_dim` 256 that nest is twice the size of the `head_dim` 128 shape in the PR description. Shared memory is also a different budget on this part: 99 KB opt-in per block and 100 KB per SM on sm_120, versus 227 KB on sm_100, so the staging buffer's cost in `NUM_MMA_KV` is not the sm_100 number either. `ncu` on the 131k/8k nvfp4 launch, same two builds:

| `ncu`, kv 131,072 / qo 8,192, one launch | nvfp4, repack off (upstream) | nvfp4, repack on (forced) | bf16 KV |
|---|---|---|---|
| duration | 198.5 ms | 161.6 ms | 102.1 ms |
| instructions executed | 42.7 G | 13.5 G | 10.4 G |
| of which ALU / FMA / tensor pipe | 20.2 / 14.0 / 1.58 G | 6.3 / 2.8 / 1.59 G | 2.3 / 3.9 / 1.58 G |
| tensor pipe active | 37.6 % | 46.3 % | 73.1 % |
| `no_instruction` stall / issue | 0.03 | 0.02 | 0.05 |
| `math_pipe_throttle` stall / issue | 0.35 | 0.01 | 1.47 |
| dynamic smem / block | 50,176 B | 100,352 B | 49,152 B |
| occupancy limit | 2 blocks (regs) | **1 block (smem)** | 2 blocks (regs) |
| achieved occupancy | 16.1 % | 8.3 % | 16.1 % |

What I read from this: the in-loop path on sm_120 is not instruction-fetch bound the way the sm_80 profile in the description was (`no_instruction` 0.03 vs 24.9 %); it is simply executing **4.1× the instructions of bf16** for the same 1.58 G tensor ops — the dequant ALU/FMA work around each fragment, replicated per `(mma_d, mma_kv)` pair. The repack cuts that to 1.3× bf16. What it costs on this part is occupancy: the staging buffer takes `NUM_MMA_KV` from 2 to 4 and doubles dynamic smem to 98 KB, which on a 99 KB-per-block / 100 KB-per-SM part is one block per SM instead of two. So the forced build wins 3.2× fewer instructions but loses half its occupancy, and nets 19 % kernel time (31–34 % wall time on the wrapper, where the planner and reduce also shrink). A repack variant that keeps `NUM_MMA_KV` = 2 and fits two blocks should do better than either; I have not tried that.

**What I am asking.** Not to flip the default for all of SM100+ — I have no sm_100 card and the PR's reasoning may well hold there. But for sm_120 the `any_at_least(1000)` policy costs a third of NVFP4 prefill wall time on the workstation/consumer Blackwell parts (RTX PRO 4000/4500/5000/6000, RTX 5090 — all sm_120), which is where the FA2 NVFP4-KV path from vllm#46329 is aimed. Options I can see: (a) treat 12.x as a software-E2M1 target in `fp4_repack_targets`, (b) a build-time override so downstreams can opt in without patching the header. Happy to test either on this hardware, and to run the same table on an RTX PRO 4500 (also sm_120) if a second card helps.

Correctness, against a real reference: same seeded inputs in both builds, and the reference is the *same kernel* fed with bf16 KV dequantised from the identical nvfp4 bytes and E4M3 scales — so the diff isolates the dequant arithmetic of the nvfp4 path, nothing else.

| kv_len / qo_len | max \|Δ\| repack off | max \|Δ\| repack on | mean \|Δ\| off / on | max \|ref\| |
|---|---|---|---|---|
| 8,192 / 1,024 | 0.00195 | 0.00195 | 4.0e-6 / 7.5e-6 | 1.44 |
| 32,768 / 4,096 | 0.00195 | 0.00195 | 1.7e-6 / 2.8e-6 | 1.22 |
| 131,072 / 8,192 | 0.00195 | 0.00195 | 2.1e-6 / 3.2e-6 | 0.43 |
| 131,072 / 131,072 | 0.0156 | 0.0156 | 2.5e-6 / 4.2e-6 | 4.13 |

Both variants sit at one bf16 ULP of the dequantised reference at every shape (the 0.0156 row is one ULP at magnitude 4). Outputs of the two builds are not bit-identical to each other — the repack multiplies the scale once per b128, the in-loop path once per fragment, so the rounding order differs — but neither is closer to the reference than the other.

Build recipe (the two-line patch as a `--build-arg`), both benchmark scripts, the PTX probe, the `ncu` scripts and the raw logs: https://github.com/TensorRaya/nvfp4-kv-sm120-flashinfer-evidence

qsang-nv:
@TensorRaya I reproduced on two Blackwell parts — your two-line patch against current `main`, separate JIT caches, CUDA 13.2, both bf16 and fp16 Q.

- **sm_120 (RTX PRO 6000 Blackwell)**: repack on is 12–21% faster across `head_dim` 128/256 and kv/qo from 32k/8k to 128k/128k. bf16 and fp16 gain about the same, so this is not the `e2m1 -> f16 -> bf16` detour.
- **sm_100 (B200)**: 1.61x at `head_dim` 128 and **3.04x** at `head_dim` 256 (kv 131072 / qo 8192, bf16, causal).

`ncu` explains the B200 case: the in-loop path there is far more instruction-fetch starved than sm_120 (`no_instruction` stall 3.03 vs 0.02), because the extra shared memory lets the launcher pick `NUM_MMA_KV` 8 instead of 2, so the unrolled `(mma_d, mma_kv)` nest replicates the dequant 4x more. Instructions 1.805 G -> 0.733 G, tensor pipe 11.2% -> 38.1%, occupancy unchanged.

So the gate looks wrong for SM100 too, not just too coarse. Feel free to open a PR removing it.

lesj0610:
Opened #5239 for this. It drops the arch target list and the runtime policy dispatch, so the repack is on wherever the shape already allows it. POD and persistent keep the in-loop path, since they have no repack call site.

The Blackwell numbers in there are yours and TensorRaya's — I have no such card, so CI has to carry the correctness side. Could you run `/bot run tests/attention` on it when you get a chance? The PR also adds symmetric head_dim 256 cases; the suite had none, and that is the shape your B200 run was on.

TensorRaya:
Gate removal measured separately in flashinfer-ai/flashinfer#5239 on sm_120 (RTX PRO 4500, CUDA 13.0): NVFP4 prefill ~27-29% faster on the removed-gate branch vs main across head_dim 128 and 256, kv 32k-400k, bf16-Q and fp16-Q consistent. The gain holding on native-cvt parts matches your hoisted-per-fragment-work reading rather than the convert-paying theory. Numbers, JIT-cold runs and the dequant-reference check are in the #5239 comment. The NUM_MMA_KV=2 two-block variant stays open on our side.

TensorRaya:
Confirmed the reproduction on a third card class: RTX PRO 4500 (sm_120), same pairing method — PR head vs main as the only variable, bf16 reference in each round. Removing the gate makes the NVFP4-KV repack path active on sm_120 and speeds up nvfp4 prefill ~27–30% vs the gated build (ratio vs bf16 drops from ~1.57→1.15 at head_dim 128 and ~1.8→1.32 at head_dim 256). Correctness holds (maxabs <= 0.002 vs bf16). Details in #5239.

Issue #4770, opened 2026-08-27
[Feature]: CUDA graph mode for the prims-ts paged decode backend


elwhyjay:
### Before submitting

- [x] I have searched existing issues and this request has not been filed yet.

### Problem

`BatchDecodeWithPagedKVCacheWrapper(backend="prims-ts", use_cuda_graph=True)` is rejected since #4739, so the only backend that supports dense (non-causal) multi-token decode cannot be used under CUDA graphs. Multi-token decode is exactly the serving path where graphs matter.

 The wrapper's graph contract is re-plan then replay: `plan()` runs outside the graph each step and writes into fixed buffers, and `replay()` picks up the new metadata. `BatchDecodePagedTSWrapper.plan()` does not fit that contract. It snapshots `seq_lens`, allocates scratch, and specializes the compiled kernel on plan-time values, so a captured `run()` would keep the tensors and kernel of the plan it was captured under.

 One correction to my note on #4739: I wrote that this needs a graph mode inside prims-ts. Looking closer, the functional `prims_ts_batch_decode_with_kv_cache` path already has the required properties. It compiles the dynamic variant only, reads kv lengths from a caller-owned device tensor, takes caller scratch, and does no device-to-host reads. The remaining work is on the wrapper side.

### Requested outcome

`use_cuda_graph=True` works with `backend="prims-ts"` with the same re-plan then replay semantics as the other tensor-core backends: plan once per step with changing kv lengths within the fixed batch, replay the captured `run()`, and get the same output as the eager path.

### Target hardware

SM100 (B200, GB200)

### Inference engine

vLLM

### Workload and configuration

 - Data type and quantization: bf16 and fp8 e4m3 KV, matching the prims-ts decode contract
 - Batch size or request concurrency: fixed graph batch, measured at 8, 64, 256
 - Sequence lengths or token counts: kv 2K to 32K, `q_len_per_req` 1 and 4
 - Parallelism: not applicable
 - Relevant shapes: GQA Hq 32 / Hkv 8, head_dim 128, page size 16, HND layout
 - Environment: B200, CUDA 13.0, torch 2.9.0+cu130, nvidia-cutlass-dsl 4.7.0

### Related work, dependencies, or suggested scope

Follow-up to #4739

PR #4771, opened 2026-08-27
feat(prims-ts): CUDA graph mode for the prims-ts decode backend

elwhyjay:
## 📌 Description

`BatchDecodeWithPagedKVCacheWrapper(backend="prims-ts", use_cuda_graph=True)` is rejected since #4739. This PR enables it.

The wrapper's CUDA graph contract is re-plan then replay: `plan()` runs outside the graph each step and writes into fixed buffers, and `replay()` picks up the new metadata. `BatchDecodePagedTSWrapper.plan()` does not fit that contract. It snapshots `seq_lens`, allocates scratch, and specializes the compiled kernel on plan-time values, so a captured `run()` would keep the tensors and kernel of the plan it was captured under.

The functional `prims_ts_batch_decode_with_kv_cache` path already fits. It compiles the dynamic variant only, reads kv lengths from a caller-owned device tensor, takes caller scratch, and does no device-to-host reads. Under `use_cuda_graph` the wrapper now routes to it:

- kv lengths are staged into the existing `_kv_lens_buffer`, the same pattern trtllm-gen and cute-dsl use
- scratch is carved from the front of `float_workspace_buffer`, sized by `get_prims_ts_batch_decode_workspace_size`, and re-zeroed when the compile key changes
- the static kv-length bound is the kv-token capacity of `paged_kv_indices_buffer`. Any request whose pages fit the index buffer stays under it, so it never changes between plans
- `plan()` compiles the key through the new `warm_prims_ts_batch_decode` so capture does not JIT

Eager mode is unchanged and keeps the plan-specialized delegate.

### Bound choice

The functional path needs a static `max_seq_len` for its compile key. Measured on B200 (bf16, Hq 32, Hkv 8, D 128, page 16), the automatic policy (`splits_kv`, persistent scheduling, kv tile) and the scratch size are identical for bounds from the exact length up to the 2,147,483,392 cap, at batch 8, 64, and 256. Latency is identical within 0.1%. So the bound costs nothing and an explicit `plan(max_kv_len=...)` knob is not needed.

### Cost of the dynamic variant

The eager delegate specializes on plan-time metadata (`planned_uniform_max`, `planned_full`). The graph path gives that up. Same shapes, median of CUDA event timing:

| batch | kv_len | eager ms | graph ms | ratio | policy |
|---|---|---|---|---|---|
| 64 | 2048 | 0.1012 | 0.1024 | 1.011 | 1 split, persistent |
| 64 | 8192 | 0.3237 | 0.3237 | 1.000 | 1 split, persistent |
| 64 | 32768 | 1.2298 | 1.2309 | 1.001 | 1 split, persistent |
| 8 | 8192 | 0.0595 | 0.0615 | 1.034 | 2 splits |
| 256 | 4096 | 0.6113 | 0.6116 | 1.000 | 1 split, persistent |

## 🔍 Related Issues

#4770 

### ✅ Pre-commit Checks

## 🧪 Tests

`tests/attention/test_prims_ts_decode_backend.py`: re-plan then replay across three kv-length sets for dense SQ=1, dense SQ=4, and causal SQ=4 against the paged reference; graph output equals the eager delegate; scratch is rebound after `reset_workspace_buffer()`; rejection of an undersized `float_workspace_buffer`. The rejection and rebind cases are gated on SM100a because the prims-ts sizing helper is architecture-gated and raises `NotImplementedError` first on other GPUs. The previous `use_cuda_graph` rejection test is removed.

Run on B200 (SM100a), `nvidia-cutlass-dsl` 4.7.0:

```
tests/attention/test_prims_ts_decode_backend.py    33 passed
tests/attention/test_batch_decode_kernels.py      742 passed   (-k "uniform_multi_token or fast_plan or workspace", with test_workspace_size.py)
tests/attention/test_attention_ts_decode.py        10 passed   (-k "workspace or functional or graph")
```
## Reviewer Notes

`warm_prims_ts_batch_decode` is the only prims-ts side change. It mirrors `get_prims_ts_batch_decode_workspace_size` and compiles the dynamic kernels for the key. Happy to fold it into the sizing helper or keep it private if that is preferred.

`_plan_prims_ts_graph` imports `_DECODE_MAX_KV_LEN` to clamp the capacity bound. A public constant would avoid the private import if reviewers prefer.

cc:@PerkzZheng @saltyminty

## Summary by CodeRabbit

- **New Features**
  - Added CUDA Graph support for prims-ts batch decoding, including replay after re-planning and workspace reset.
  - Added a helper to warm and precompile batch decode kernels without running inference.
  - Added support for caller-provided sequence-length and workspace buffers during planning.
  - Added validation for decode configuration, data types, layouts, masking, and device compatibility.

- **Bug Fixes**
  - Improved workspace sizing, reuse, validation, and rebinding during planning and reset.

### DIFF
--- a/flashinfer/decode.py
+++ b/flashinfer/decode.py
@@ -1035,6 +1035,9 @@ class BatchDecodeWithPagedKVCacheWrapper:
             device="cpu",
             pin_memory=True,
         )
+        if self._backend == "prims-ts":
+            self._prims_ts_graph_key = None
+            self._prims_ts_workspace = None
 
     @flashinfer_api
     def workspace_size(
--- a/flashinfer/decode.py
+++ b/flashinfer/decode.py
@@ -876,8 +876,10 @@ class BatchDecodeWithPagedKVCacheWrapper:
             no RoPE/ALiBi/soft-cap).
             The ``prims-ts`` backend uses the task-scheduled decode kernel on SM100a/SM103a.
             It is the only backend that accepts ``is_causal=False`` with
-            ``q_len_per_req > 1``. It requires ``kv_layout="HND"`` and does not
-            support ``use_cuda_graph=True``.
+            ``q_len_per_req > 1`` and requires ``kv_layout="HND"``. Under
+            ``use_cuda_graph`` it runs the dynamic kernel variant, bounded by the
+            kv-token capacity of ``paged_kv_indices_buffer``, with scratch carved
+            from ``float_workspace_buffer``.
 
         jit_args : Optional[List[Any]]
             If provided, the wrapper will use the provided arguments to create the JIT module,
@@ -924,7 +926,7 @@ class BatchDecodeWithPagedKVCacheWrapper:
             device="cpu",
         )
         self._kv_lens_buffer: Optional[torch.Tensor] = None
-        if backend in ("trtllm-gen", "cute-dsl"):
+        if backend in ("trtllm-gen", "cute-dsl", "prims-ts"):
             self._kv_lens_buffer = torch.empty(
                 (32768,), dtype=torch.int32, device=self.device
             )
@@ -986,15 +988,17 @@ class BatchDecodeWithPagedKVCacheWrapper:
                 raise NotImplementedError(
                     "prims-ts decode backend requires kv_layout='HND'"
                 )
-            # The delegate snapshots seq_lens, scratch, and the compiled kernel
-            # at plan time, so a captured run() does not follow a later plan().
-            if use_cuda_graph:
-                raise NotImplementedError(
-                    "prims-ts decode backend does not support use_cuda_graph=True"
-                )
-            from .attention.prims_ts import BatchDecodePagedTSWrapper
+            # The delegate specializes its kernel on plan-time metadata, so it
+            # cannot back a captured run() that must follow later plan() calls.
+            # Under CUDA graph the wrapper drives the functional entry point
+            # with a static kv-length bound and wrapper-owned scratch instead.
+            self._prims_ts_graph_key = None
+            self._prims_ts_workspace: Optional[torch.Tensor] = None
+            self._prims_ts_max_kv_len: Optional[int] = None
+            if not use_cuda_graph:
+                from .attention.prims_ts import BatchDecodePagedTSWrapper
 
-            self._prims_ts_wrapper = BatchDecodePagedTSWrapper()
+                self._prims_ts_wrapper = BatchDecodePagedTSWrapper()
 
     @property
     def use_tensor_cores(self) -> bool:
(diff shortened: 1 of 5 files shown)


tests/attention/test_prims_ts_decode_backend.py
```diff
@@ -472,12 +484,75 @@ def test_rejects_noncontiguous_multi_q_tensors():
 
 
 @requires_prims_ts_gpu
-def test_graph_capture_of_fixed_plan_replays():
-    """Capturing run() after a plan replays that plan; re-planning is not covered.
+@pytest.mark.parametrize("q_len_per_req,is_causal", [(1, False), (4, False), (4, True)])
```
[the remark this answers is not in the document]
elwhyjay: the marker is on line 486 above `@pytest.mark.parametrize`

elwhyjay:
Gentle ping @PerkzZheng. Could you review this PR if you have a moment? Thank you

flashinfer/decode.py
```diff
@@ -1679,22 +1686,39 @@ def _plan_impl(
...
+                    q_len_per_req=q_len_per_req,
+                    q_data_type=q_data_type,
+                    kv_data_type=kv_data_type,
+                    o_data_type=o_data_type,
+                    mask_type=mask_type,
+                    window_left=window_left,
+                    non_blocking=non_blocking,
+                )
+            else:
+                self._prims_ts_wrapper.plan(
```
PerkzZheng, line 1715: why self._prims_ts_wrapper.plan( cannot work with cuda graph ? or why we didn't change self._prims_ts_wrapper.plan( to work with cuda graph ?

flashinfer/decode.py
```diff
@@ -1860,6 +1884,95 @@ def _plan_impl(
...
+        o_data_type: torch.dtype,
+        mask_type: Literal["dense", "causal"],
+        window_left: int,
+        non_blocking: bool,
+    ) -> None:
+        from .attention.prims_ts import (
+            get_prims_ts_batch_decode_workspace_size,
+            warm_prims_ts_batch_decode,
+        )
+        from .attention.prims_ts.decode import _DECODE_MAX_KV_LEN
```
PerkzZheng, line 1918: we probably want to remove the _DECODE_MAX_KV_LEN limitation (not quite sure why it was added).

flashinfer/decode.py
elwhyjay: Thank you for review!

As i understand, `plan()` specializes the compiled kernel on plan-time values (kv_prefix_mode / kv_lengths_mode and the planned max_kv_len are part of the JIT key) and allocates fresh seq_lens each plan, so a captured run() keeps replaying the old plan. It also requires `indptr[-1] == paged_kv_indices.numel()`, which rejects the oversized graph buffers. Making it graph-safe would mean always compiling the dynamic variant with stable buffers, which is exactly what the functional entry already provides, so I routed through that to keep the eager-path specialization.

 If you think the delegate to own a graph mode is a right way, I will address it.

flashinfer/decode.py
elwhyjay: This is the existing kernel contract rather than a new limit: the K/V mask computes tile_offset_k + tile_size_kv in signed Int32, so _validate_max_kv_len caps max_seq_len at 2^31 - 255 (see the constant's comment in attention/prims_ts/decode.py). The min() only clamps the capacity-derived bound, since indices.numel() * page_size can exceed that ceiling for large page pools and would otherwise fail sizing validation. 

Removing the underlying limitation would need the kernel tile-endpoint arithmetic widened beyond Int32, which felt out of scope here, but I am happy to file it as a follow-up issue, or to expose the constant publicly instead of the private import if that is the concern.

PerkzZheng: okay, then let us keep it for now (even though it seems to be a bit verbose). Thanks

flashinfer/decode.py
PerkzZheng: I think we should modify plan() to not allocate seqlens and workspace inside it. Instead, we should construct them once and pass it to plan() like other attention backends. Can you help modify that ? Thanks.

elwhyjay: Sure! I will modify it

elwhyjay: I addressed this in like this.

`plan()` now takes caller-owned seq_lens and workspace_buffer, the wrapper integration stages kv lengths in the reusable device buffer and binds scratch from float_workspace_buffer, and the standalone path keeps one owned buffer instead of reallocating per plan. 

Happy to adjust the shape of the API if you had something different in mind.

PerkzZheng: that makes a lot sense to me. Thanks!

elwhyjay:
@coderabbitai all fixed

elwhyjay:
Hi @PerkzZheng. Could you run the CI if you have a moment? Thank you

PR #4787, opened 2026-08-27
Rubin open issue fixes

Vinnie6167:
## 📌 Description

Four independent SM107 (Rubin) fixes found during bring-up. They are unrelated to each other and can be reviewed separately.

### `fix(trtllm-gen)` — accept SM107 for the `Sm100f` family-conditional target

`isArchCompatible` mapped `CudaArch::Sm100f` to `smVersion` 100 or 103, omitting 107. `Sm100f` is the family-conditional target and is valid across the whole SM100 line, so on sm107 the filter discarded every family cubin. When the pinned pack also lacks native `Sm107a` entries for the requested option sets, nothing matches at all.

### `fix(moe)` ×2 — decline the CuTe DSL NVFP4 backend when the DSL predates 4.8

The SM107 gather/activation-fusion and finalize-fusion kernels are built on `cutlass.utils.rubin_helpers`, which only exists from CuTe DSL 4.8. Two entry points needed guarding, and they are genuinely distinct:

* `cute_dsl/tuner.py` covers the **autotuning** path — without it the tactic import raises `ModuleNotFoundError` instead of simply being declined, aborting autotuning rather than falling back.
* `fused_moe/runners.py` covers **backend selection**. `_check_support()` previously checked only the activation type and the W4A16 per-token scale, so on an older DSL the runner passed the support check, survived `build()`, and entered `MoELayer.runners`; the failure then surfaced from inside `forward()`. A direct `forward(tactic=-1)` call bypasses tactic filtering entirely, so the tuner guard alone does not cover it.

The `runners.py` probe is **arch-conditional on purpose**: only the SM107 kernels need `rubin_helpers`, so an older DSL remains fully usable on SM100/SM103 and gating unconditionally would drop a working backend on Blackwell.

### `test(moe)` — skip SM107 parameterizations the CuTe DSL kernels do not implement

`fused_moe/cute_dsl/rubin/` holds a narrower specialisation of the Blackwell kernels rather than a port of them: the gather kernel hardcodes SwiGLU and exposes no `activation_type`, its wrapper has no `a_per_token_scale_ptr`, and the finalize kernel has no unfused path. The resulting `NotImplementedError`s for `use_a_per_token_scale`, `use_fused_finalize=False` and `GegluTanh` are accurate statements about kernel code that does not exist — product gaps, not defects — but they report as failures on every SM107 run.

No dispatch-level fix is possible, and that was measured rather than assumed: the affected tests call `cute_dsl_fused_moe_nvfp4()` directly and contain zero `MoELayer` references, so there is no backend selection to influence.

The skip is decided **from the parameterization, before the test body runs**, so it cannot absorb a genuine regression — anything failing for a different reason still fails. The three parameters are parametrized only in this file, so no other MoE test is affected.

## 🔍 Related Issues

None filed upstream.

**Split out of this PR:** a JIT cache fix (atomic publish of the linked `.so` plus discard-and-rebuild recovery) was originally included here. It is **not** SM107-specific — nothing in it is arch-conditional, and it can strand any GPU whose node has an interrupted link left behind. It is being carried separately so it is not reviewed as a Rubin change.

### ✅ Pre-commit Checks

## 🧪 Tests

Validated on SM107 hardware:

| Change | Evidence |
|---|---|
| `fix(trtllm-gen)` | Controlled A/B, identical tree, only these lines differing: unpatched 158 × "no kernels runnable on sm107", 79 failed / 2 passed → patched 0 × that error, 26 failed / **56 passed**. |
| `test(moe)` | Affected subset on SM107 hardware — before: **3 failed / 11 passed / 0 skipped**; after: **0 failed / 11 passed / 3 skipped**. All three outcome counts checked against baseline, so the skips are exactly the three known gaps and no passing test was lost. |

Additionally, the `fix(moe)` lazy-import paths were **executed** on SM107 hardware (`_assert_rubin_cute_dsl_available()` runs, the probe resolves and returns `True`, the guarded `.rubin` import succeeds). Its *declining* branch remains unexercised — that needs a public CuTe DSL < 4.8 stack.

## Reviewer Notes

* **`fix(moe)`'s decline branch is still unexercised.** Its import paths were executed on SM107 hardware, but the container ships CuTe DSL 4.8, so the probe returns `True` and the decline never fires. Only a public CuTe DSL 4.7.0 stack exercises it. An earlier revision of this commit imported `is_rubin_cute_dsl_available` from `cute_dsl.availability`, which exists on `release-v0.6.18` but **not** on `main`; because the import is function-local, `py_compile` and `ruff` both passed and only a runtime call would have caught it. It now imports from `cute_dsl.utils`, which provides the symbol on both branches, verified by execution.

Four things I would rather flag than have a reviewer discover:

1. **`Sm100f` is not universally family-valid.** Twelve lines below the changed filter there is already a carve-out — *"Sm100f cubins miss the f2fp patch, so sm103 must fall back to Sm103a for it."* The sm107 case is empirically fine, but a similar per-feature gap would need the same treatment.

2. **One of two patched runs of the trtllm-gen A/B segfaulted** with zero arch errors. The repeat completed normally and it did not reproduce — plausibly a known IMA class for that cubin pin rather than this change — but it is 1-in-2 and worth watching in CI rather than dismissing.

3. * One case is matched by function identity, not by parameter.** `test_geglu_tanh_accuracy` sets its activation in the test body rather than via a parameter, so it is matched with `request.node.function.__name__ == "test_geglu_tanh_accuracy"`. An earlier revision used a substring match on `"geglu_tanh"`, which also caught `test_geglu_tanh_activation_is_supported` — a pure-Python assertion about `normalize_cute_dsl_moe_activation_type` that touches no kernel and passes on SM107. That silently cost one passing test; the exact match restores it, confirmed by the pass count above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fused MoE autotuning on Rubin (SM107) hardware by gracefully falling back when required CuTe DSL kernel support is unavailable.
  * Added compatibility checks to prevent unsupported Rubin configurations from causing runtime failures.
  * Improved FP8 attention decode validation to accommodate expected numerical differences.
* **Tests**
  * Expanded coverage for Rubin capabilities and unsupported feature combinations.
  * Enabled Rubin paths in relevant fused MoE autotuning tests.


Vinnie6167:
@flashinfer-bot run

Vinnie6167:
@flashinfer-bot run

aleozlx (approved):
approving released changes to main

Vinnie6167:
@flashinfer-bot run

aleozlx (approved):

saltyminty (approved):

PR #4795, opened 2026-08-28
feat(moe_ep): make the nccl_ep split path CUDA-graph capturable

Anerudhan:
## 📌 Description

A `moe_ep` split-path `Handle` is created per forward today, which makes the path impossible to capture into a CUDA graph. A graph records the device pointers it sees at capture time, so a handle created *and destroyed* inside the captured forward leaves the replay dereferencing freed memory.

Measured on 4×B200 (dp=4, Qwen3-30B-A3B, driven through vLLM's `flashinfer_ep_low_latency` backend):

| Configuration | Result |
|---|---|
| `--enforce-eager` | rc=0, ~616 tok/s per rank across 4 ranks |
| CUDA graphs enabled | capture completes (PIECEWISE 35/35, FULL 35/35), then the **first replay** raises `CUDA error: an illegal memory access was encountered` |

Silent at capture, crash at replay — invisible to any test that doesn't actually run inference under graphs.

**NCCL-EP itself supports capture.** `contrib/nccl_ep/ep_test.cu` has a `--use_cuda_graph` mode, and it is *not* gated on the algorithm, so LL and HT are both captured there. Its recipe is a split:

> Non-graph mode tests CreateHandle (combined Init+Update). Graph mode tests the Init+Update split, where **InitHandle stays outside the capture** (host-side allocation only) **and UpdateHandle is recorded inside** the captured region.
> — `ep_test.cu:478-481`

and `nccl_ep.h:422` states the rule directly:

> to avoid CUDA graph invalidation, all Handles must be created before the beginning of the CUDA graph capture.

Two changes are needed, and **the second is the one that actually makes capture work**.

### 1. `Handle.update()` — the missing per-step half

`moe_ep` only ever called the combined `ncclEpCreateHandle`, so the Init/Update split could not be expressed through this API.

```python
# Before: one handle per forward -- uncapturable
for step in ...:
    handle = fleet.create_handle(HandleParams(topk_ids=topk_ids))   # Init + Update
    handle.dispatch(...); handle.combine(...); handle.complete()

# After: one durable handle, per-step rebind -- capturable
handle = fleet.create_handle(HandleParams(topk_ids=topk_buf))       # outside capture
for step in ...:
    handle.update(HandleParams(topk_ids=topk_buf))                  # inside capture
    handle.dispatch(...); handle.combine(...); handle.complete()
```

Optional capability with a raising default on the ABC, matching `dispatch_send_only` / `dispatch_recv_only`, so `NixlEpHandle` and any out-of-tree `Handle` are unaffected.

### 2. `NcclEpHandle._op_stream()` — issue transport work on the capture stream

Every dispatch/combine/complete previously issued on `self._stream`, the handle's **creation-time** stream (the `HandleAlgoKnobUserStream` value, else the fleet's). That is correct for a per-forward handle, which is created on the same stream it runs on. It is wrong for a persistent one: it is created *before* the capture begins, so its stream is not the stream being captured, and the transport work lands outside the graph entirely — **the capture records nothing and the replay is a silent no-op.**

`_op_stream()` returns the capture stream while capturing and `self._stream` otherwise, so non-graph behaviour — including an explicit `UserStream` — is byte-identical. Handle *creation* deliberately still uses `self._stream`: `nccl_ep.h:422` requires it to happen outside any capture.

This one is easy to miss because it fails silently rather than loudly; `update()` alone is not sufficient.

## 🧪 Verification on 4×B200

`tests/moe_ep/test_moe_ep_cudagraph_multirank.py`, 4 ranks, **both algorithms, all ranks passing**:

```
rank 0..3: low_latency     capture + replay OK across changed routing
rank 0..3: high_throughput capture + replay OK across changed routing
2 passed
```

The test asserts three properties in increasing order of what they would catch:

1. capture completes;
2. replay does not fault and reproduces the eager result — the regression above;
3. **the transport's reported routing changes when `topk_ids` is rewritten in place between replays.** This is the one that matters: an identity round trip is routing-invariant, so comparing outputs cannot distinguish "`update()` replayed" from "`update()` skipped". The test interrogates the transport (`expert_counts`, falling back to `recv_topk_idx`) instead.

Assertions are algorithm-aware. **HT's identity round trip is deliberately not asserted, because HT does not have that property:** `_dispatch_ht` sizes its recv buffer to `max_tokens_per_rank * world` and dispatch only writes the slots that actually received tokens, so an identity pass-through hands `combine` the unwritten remainder. Real HT consumers compute over the whole static buffer or trim to `recv_total_counter`. Capture correctness is still fully covered for HT (replay must match eager, routing must track across replays); only the numerical check is weaker than LL's.

The test also completes all collective work and tears the fleet down *before* asserting — a bare assert mid-test aborts one rank inside a collective and strands the rest at the next barrier, turning a one-line failure into a wedged multi-hour job.

## 🔍 Related Issues

Follow-up to #4183 (EP fault tolerance). Consumer: vllm-project/vllm#47948.

**No caller in this repo uses `update()` yet** — consuming it is the vLLM follow-up, which pins `flashinfer-python` and so needs a release first.

### ✅ Pre-commit Checks

## 🧪 Tests

`tests/moe_ep/nccl_ep/test_handle_mock.py` gains eight cases:

- rebinding **reuses the native handle** rather than creating a second one — the property that makes capture safe, and the one that would regress silently
- `top_k` changes are rejected
- growing past the creating token count is rejected; **shrinking is allowed** (decode steps are smaller than the capture shape)
- the ABC default raises `NotImplementedError`
- `update()` is capturable — i.e. contains no host sync, the failure mode that makes a prepare path uncapturable
- the caller's buffer is **bound, not copied** (a copy would make every replay re-run stale routing)
- outside capture, ops stay on the knob stream (pins that `_op_stream()` did not change non-graph behaviour)
- under capture, ops move to the capture stream

`FakeHandle` in `tests/moe_ep/nccl_ep/conftest.py` gains an `update()` mirroring `ncclEpUpdateHandle`'s contract (rebinds routing, never reallocates).

**72 passed** in the `nccl_ep` mock suite, locally and on the 4×B200 rig; `ruff check` / `ruff format` clean; `pre-commit` clean.

AI-assisted (Claude Code); every change reviewed and the failure reproduced end-to-end by the submitter.

## Summary by CodeRabbit

* **New Features**
  * Added support for updating existing communication handles with new routing information without reallocating buffers.
  * Enables handle reuse across forward passes and CUDA graph replays.
  * Ensures communication operations use the appropriate stream during CUDA graph capture.

* **Bug Fixes**
  * Added validation for unsupported updates, routing changes, token limits, and low-latency hidden sizes.
  * Prevents dispatch operations from exceeding per-rank capacity limits.

* **Tests**
  * Expanded coverage for handle reuse, CUDA graph replay, capacity limits, routing validation, and stream behavior.

### DIFF
--- a/flashinfer/moe_ep/backends/split/comm/nccl_ep/fleet.py
+++ b/flashinfer/moe_ep/backends/split/comm/nccl_ep/fleet.py
@@ -17,6 +17,7 @@ from ..... import _require_built
 from .....errors import MoEEpFaultToleranceUnsupportedError, MoEEpNotBuiltError
 from .....core.validation.common import (
     validate_arch_for_backend,
+    validate_ll_hidden_size,
     validate_bootstrap_world_size,
     validate_fleet_params,
 )
@@ -160,6 +161,7 @@ class NcclEpFleet(FaultToleranceMixin, Fleet):
     ) -> None:
         _require_built("nccl_ep")
         validate_arch_for_backend("nccl_ep")
+        validate_ll_hidden_size(params, "nccl_ep")
         validate_bootstrap_world_size(bootstrap)
 
         # HT: clamp the per-rank dispatch budget to the library's build-time cap so
--- a/flashinfer/moe_ep/backends/split/comm/nccl_ep/handle.py
+++ b/flashinfer/moe_ep/backends/split/comm/nccl_ep/handle.py
@@ -115,6 +115,9 @@ class NcclEpHandle(Handle):
             topk_idx = topk_idx.to(torch.int64)
         self._topk_idx = topk_idx
         self._num_tokens_in = topk_idx.shape[0]
+        # Buffers are sized for the creating shape; update() may only
+        # rebind a token count at or below it.
+        self._max_num_tokens_in = topk_idx.shape[0]
         self._top_k = topk_idx.shape[1]
         self._topk_idx_t = self._wrap(topk_idx)
 
@@ -175,6 +178,8 @@ class NcclEpHandle(Handle):
             self._topk_idx_t,
             layout_info=create_layout_info,  # HT recv-count opt-in; None otherwise
             config=None,
+            # Creation is host-side allocation and must happen OUTSIDE any
+            # capture (nccl_ep.h:422), so it stays on the handle's own stream.
             stream=self._stream,
         )
         _t = _hp("hinit.create_handle_c", _t)
@@ -182,6 +187,25 @@ class NcclEpHandle(Handle):
     def _knob_stream(self) -> int:
         k = self._handle_knobs.get(HandleAlgoKnobUserStream)
         return int(k.stream) if k is not None else self._fleet.stream  # type: ignore[attr-defined]
+
+    def _op_stream(self) -> int:
+        """Stream to issue transport work on.
+
+        Normally the handle's own stream: the ``HandleAlgoKnobUserStream``
+        value, else the fleet's. Under CUDA-graph capture that is the wrong
+        one. A handle that outlives a capture is created *before* it begins
+        (see ``update``), so its creation-time stream is not the stream being
+        captured, and work issued there lands outside the graph entirely --
+        the capture records nothing and the replay is a no-op.
+
+        Outside capture this returns exactly what it always did, so non-graph
+        behaviour (including an explicit UserStream) is unchanged.
+        """
+        import torch
+
+        if torch.cuda.is_current_stream_capturing():
+            return torch.cuda.current_stream().cuda_stream
+        return self._stream
 
     # Only memoize wrappers of SMALL tensors: the wrapper keeps the torch tensor
     # alive, so caching wraps of large activations (e.g. 8k-token prefill inputs,
@@ -214,6 +238,49 @@ class NcclEpHandle(Handle):
             w = self._ep.Tensor(t)
             hot[key] = w
         return w
+
+    def update(self, params) -> None:
+        """Rebind to a new step's routing via ``ncclEpUpdateHandle``.
+
+        This is the per-step half of the split that makes CUDA-graph capture
+        possible: ``ncclEpInitHandle`` (done in ``__init__``, via
+        ``create_handle``) allocates and must stay outside the capture, while
+        this call only recomputes routing metadata and is safe to record
+        inside it. Without it a handle is created and destroyed per forward,
+        so a captured graph replays against freed device memory.
+
+        ``top_k`` is bound at handle creation (LL passes ``num_topk`` to
+        InitHandle), so only the token count may vary, and only downward --
+        the recv buffers were sized for the creating shape.
+        """
+        import torch
+
+        topk_idx = params.topk_ids
+        if topk_idx.dtype != torch.int64:
+            topk_idx = topk_idx.to(torch.int64)
+        if topk_idx.shape[1] != self._top_k:
+            raise ValueError(
+                f"Handle.update cannot change top_k: handle was created with "
+                f"top_k={self._top_k}, got {topk_idx.shape[1]}. Create a new "
+                "handle instead."
+            )
+        if topk_idx.shape[0] > self._max_num_tokens_in:
+            raise ValueError(
+                f"Handle.update: {topk_idx.shape[0]} tokens exceeds the "
+                f"{self._max_num_tokens_in} this handle's buffers were sized "
+                "for. Create the handle at the largest shape you will use."
+            )
+        self._topk_idx = topk_idx
+        self._num_tokens_in = topk_idx.shape[0]
+        self._topk_idx_t = self._wrap(topk_idx)
+        # layout_info is the HT recv-count opt-in and None for LL, which is
+        # exactly what ncclEpUpdateHandle requires of each mode. See
+        # _op_stream() for why this is not simply self._stream.
+        self._handle.update(
+            self._topk_idx_t,
+            layout_info=self._create_layout_info,
+            stream=self._op_stream(),
+        )
 
     def dispatch(self, params: DispatchInputParams) -> DispatchOutput:
         x = params.x[0]
@@ -285,10 +352,10 @@ class NcclEpHandle(Handle):
             outputs,
             layout_info=layout_info,
             config=config,
-            stream=self._stream,
+            stream=self._op_stream(),
         )
         _t = _hp("ll_disp.ffi_dispatch", _t)
-        self._handle.complete(stream=self._stream)
+        self._handle.complete(stream=self._op_stream())
         _t = _hp("ll_disp.ffi_complete", _t)
 
         self._dispatch_inputs = inputs
@@ -354,9 +421,9 @@ class NcclEpHandle(Handle):
             outputs,
             layout_info=layout_info,
             config=config,
-            stream=self._stream,
+            stream=self._op_stream(),
         )
-        self._handle.complete(stream=self._stream)
+        self._handle.complete(stream=self._op_stream())
 
         self._dispatch_inputs = inputs
         self._dispatch_outputs = outputs
@@ -449,10 +516,14 @@ class NcclEpHandle(Handle):
         _t = _hp("ht_disp.build_ffi_objs", _t)
 
         self._handle.dispatch(
-            inputs, outputs, layout_info=None, config=config, stream=self._stream
+            inputs,
+            outputs,
+            layout_info=None,
+            config=config,
+            stream=self._op_stream(),
         )
         _t = _hp("ht_disp.ffi_dispatch", _t)
-        self._handle.complete(stream=self._stream)
+        self._handle.complete(stream=self._op_stream())
         _t = _hp("ht_disp.ffi_complete", _t)
 
         self._dispatch_inputs = inputs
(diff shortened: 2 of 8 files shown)


feih-nv:
Review comments by the agent:
1. High — shrinking a handle leaves stale routing weights.
`update()` permits fewer `topk_ids`, but `HandleAlgoKnobTopKWeights` remains frozen at its creation shape. Subsequent LL/HT dispatch or combine can hit native shape assertions. Either rebind weights during update or require a fixed token count.

2. Medium — dispatch does not require `x.shape[0] == self._num_tokens_in`.
After an update, a mismatched activation count passes the capacity guard and reaches NCCL-EP. Add a common equality check before dispatch, plus device/contiguity validation for updated routing tensors.

3. Medium — cross-stream initialization ordering is missing.
Creation runs on `self._stream`; capture immediately switches to another stream without an event/wait. Existing tests synchronize and warm up first, masking this race.

4. Test blocker — the four-rank graph test is not wired into `run_tests.sh`’s multirank path, so normal execution skips it. Add it to the torchrun suite.

5. Test gap — use a non-identity expert operation.
The routing signature proves update/dispatch replay, but identity output can still hide a missing combine replay.

Anerudhan:
> Review comments by the agent:
> 
> 1. High — shrinking a handle leaves stale routing weights.
>    `update()` permits fewer `topk_ids`, but `HandleAlgoKnobTopKWeights` remains frozen at its creation shape. Subsequent LL/HT dispatch or combine can hit native shape assertions. Either rebind weights during update or require a fixed token count.
> 2. Medium — dispatch does not require `x.shape[0] == self._num_tokens_in`.
>    After an update, a mismatched activation count passes the capacity guard and reaches NCCL-EP. Add a common equality check before dispatch, plus device/contiguity validation for updated routing tensors.
> 3. Medium — cross-stream initialization ordering is missing.
>    Creation runs on `self._stream`; capture immediately switches to another stream without an event/wait. Existing tests synchronize and warm up first, masking this race.
> 4. Test blocker — the four-rank graph test is not wired into `run_tests.sh`’s multirank path, so normal execution skips it. Add it to the torchrun suite.
> 5. Test gap — use a non-identity expert operation.
>    The routing signature proves update/dispatch replay, but identity output can still hide a missing combine replay.

Thanks — all five land, and all five are fixed in a112cb9b, pushed after this review. Point by point.

**1. Shrinking leaves stale routing weights — High. Correct; fixed by removing the allowance rather than rebinding.**

`HandleAlgoKnobTopKWeights` is bound at `InitHandle` and `update()` has no path to rebind it, so a shorter `topk_ids` leaves combine reading weights for rows that no longer exist. Of your two options I took the second: the token count is now fixed for the handle's lifetime, exactly like `top_k`. Only routing *values* may change.

That loses nothing the feature exists for — a CUDA graph bakes shapes at capture, so a varying token count was never expressible under capture anyway, and the shrink allowance had no caller; it was speculative. `update()` now raises on any `topk_idx.shape[0] != self._num_tokens_in`. `test_update_rejects_a_different_token_count` is parametrized over both grow and shrink.

**2. `dispatch` does not require `x.shape[0] == self._num_tokens_in` — Medium. Correct; fixed.**

The equality check is now in the common `dispatch()` before the LL/HT split (raising `MoEEpConfigError`), so it fires on every path rather than depending on each per-path capacity guard to catch it incidentally. `update()` additionally validates that the new `topk_ids` are on the GPU, on the same device as the routing it replaces, and contiguous. Tests: `test_dispatch_rejects_activation_count_mismatch`, `test_update_rejects_non_contiguous_routing`.

Adding the guard immediately caught a second thing: the existing LL over-dispatch test had been tripping the new equality check rather than the capacity guard it targets, so it now sizes the handle past the fleet and tests what it claims again.

**3. Cross-stream initialization ordering — Medium. Guard added; let me be precise about what it does and does not do.**

`update()` now refuses to be the *first* handle operation when that operation is itself inside a capture:

> `RuntimeError: Handle.update: first cross-stream update cannot be the captured one -- the dependency on InitHandle would not be recorded. Run one update outside the capture first (the standard warmup does this).`

Confirmed on device: capturing a fresh handle's first `update()` raises exactly this.

Three caveats I would rather state than paper over:

- **The event half of that change is currently unreachable.** `_op_stream()` returns `self._stream` whenever `is_current_stream_capturing()` is false, so `op_stream != self._stream` can only hold *under* capture — which is the branch that raises. The `wait_event` line therefore never executes; I verified this on device. The effective protection is the raise, not the event. I will either drop the event or correct the comment so it stops claiming a dependency it never records.
- **Under `torch.cuda.graph()` the race was already closed upstream.** Its `__enter__` does a full `torch.cuda.synchronize()` before `capture_begin`, so `InitHandle` is retired before any capture-stream work regardless. The guard earns its keep for a caller driving `cudaStreamBeginCapture` directly. So: a real gap in the API contract, not a live bug in the path the tests exercise — which is also why, as you say, the existing tests could not have caught it.
- **No test covers this guard yet.** I will add one.

**4. Not wired into `run_tests.sh` — correct; fixed.**

`tests/moe_ep/test_moe_ep_cudagraph_multirank.py` is now in `run_multirank()`'s torchrun suite (`-m "nvep and gpu_4"`, `nccl_ep` only — `nixl_ep` has no Init/Update split to capture) and `--ignore`d in `run_unit()`, matching the other multirank files.

**5. Identity expert operation — correct; fixed.**

Two changes. The expert step is now a constant scale (`GAIN = 2.0`) rather than the identity, and a fourth phase rewrites the **activations** in place (`rig.x.mul_(-3.0)`) and replays. The second is what closes the hole you named: the routing signature witnesses `update` + `dispatch`, but a combine that never replayed would leave the previous contents in the output buffer and still satisfy it. Only a replayed combine tracks a changed `x`.

LL asserts `replay_newx == rig.x * GAIN` exactly. HT asserts only `replay_newx != replay_new`, since HT's recv buffer keeps unwritten slots and so has no exact round-trip property to assert against.

---

**Status.** 104 mock tests pass (`tests/moe_ep/nccl_ep` + `tests/moe_ep/nixl_ep`); `ruff` and `pre-commit` clean. The 4-rank B200 run has **not** been repeated since a112cb9b — the check on that commit reads `CI skipped — pending authorization`, so `@flashinfer-bot run tests/moe_ep` still needs to fire. I will land the point-3 cleanup plus its test, re-run the rig and CI, and report back before asking for another look.

aleozlx:
## PR Review Screening

**CI verdict:** ✅ auto-run ok
**Review category:** `live` (rule fired: C3.2 durable — the publicly exported `Handle` ABC gains a lifecycle method that defines a create-once/update-per-step contract for every EP comm backend)
**Blocking checks:** none
**Release blocker:** no — the IMA is only reachable on a path this PR newly enables; no shipped behavior regresses (C3.4)
**Early stop:** no

### Security
| Q | Answer | Evidence |
|---|--------|----------|
| S2 template overwritten | no | Description / Related Issues / Checklist / Tests present; optional "Reviewer Notes" omitted |
| S3 template obligations | met | 5/5 boxes checked; tests bucket non-empty (4 files, +708) |

### Packaging
| Q | Answer | Evidence |
|---|--------|----------|
| C1.1 external dependency bump | no | no `supply` bucket files |
| C1.2 public API changes | yes — **extension of existing shape** | `flashinfer/moe_ep/core/comm/handle.py`: `Handle.update(self, params: "HandleParams") -> None` (new optional ABC method, `NotImplementedError` default, mirroring `dispatch_send_only`); `flashinfer/moe_ep/backends/split/comm/nccl_ep/handle.py`: `NcclEpHandle.update(self, params) -> None`; internal-only: `core/validation/common.py::validate_ll_hidden_size(params: FleetParams, backend: str) -> None`. `Handle` is in `flashinfer.moe_ep.__all__` |
| C1.3 AOT/trace registration | no gap | no `gen_*_module()` added; moe_ep has no `TraceTemplate` family |

### Presentation
| Q | Answer | Evidence |
|---|--------|----------|
| C2.1 perf claim backed | n-a | no speedup claimed; the 4×B200 / Qwen3-30B-A3B numbers are repro context for the capture failure |

### Implementation
| Q | Answer | Evidence |
|---|--------|----------|
| C3.1 experimental track declared | no | no `experimental` label, no `@flashinfer_experimental_api`, nothing under `flashinfer/experimental/` |
| C3.2 shared/durable areas touched | yes — durable | `core/comm/handle.py` (backend-facing ABC), `core/validation/common.py` (shared validation), and `NcclEpHandle._op_stream()` changing the issue stream for all dispatch/combine/complete |
| C3.3 tests match behavior change | yes | 8 mock cases in `tests/moe_ep/nccl_ep/test_handle_mock.py` execute in CI (`scripts/task_jit_run_tests_part1.sh` → `run_tests.sh unit`); the 4-GPU `test_moe_ep_cudagraph_multirank.py` is `--ignore`d there and runs only on the manual multirank rig |
| C3.4 release-blocker candidate | no | IMA/`no-op replay` occur only under graph capture, which the split path did not support before this PR |

### Experimental track
| Q | Answer | Evidence |
|---|--------|----------|

**Notes for the maintainer**
- The design question for the live pass: `update()` has no in-repo caller (consumer is vllm-project/vllm#47948, which pins a release), so the handle-reuse contract — fixed `top_k`/token count, `_ran_outside_capture` ordering guard, capture-stream selection — is being frozen into a public ABC before any first-party use exercises it.
- Downstream vLLM work is release-pinned, so a maintainer may want this in the nearest release for scheduling reasons even though it is not a C3.4 blocker.
- Body says shrinking the token count is allowed; `NcclEpHandle.update()` rejects any count `!= self._num_tokens_in` — a doc/code mismatch for the reviewer to confirm (not a code-review finding).

aleozlx (approved):

PR #4798, opened 2026-08-28
fix(jit): avoid invalidating bgmv_moe cache on source staging

### LINKED ISSUE #4782: [Bug] JIT: bgmv_moe rebuilds from scratch in every process because shutil.copy resets the copied sources' mtime
## Summary

`gen_bgmv_moe_module()` copies its CUDA sources and headers into the JIT `generated/` directory with `shutil.copy`, which does not preserve mtime. The copies therefore receive the current time on every process start, ninja sees them as newer than the previously built objects, and the module is fully recompiled — even when `FLASHINFER_WORKSPACE_BASE` is shared across runs and nothing has changed.

Measured on one machine with the default (release) JIT configuration: every fresh process pays ~217 s of rebuild. With `shutil.copy2` the second and later processes load from cache in ~4.3 s, i.e. ~213 s saved per process.

This is easy to miss because the cache *looks* correct — the generated directory, the cache key and `build.ninja` are all stable across runs; only the mtimes move.

## Root cause

https://github.com/flashinfer-ai/flashinfer/blob/main/flashinfer/jit/bgmv_moe.py#L100
https://github.com/flashinfer-ai/flashinfer/blob/main/flashinfer/jit/bgmv_moe.py#L108

`JitSpecNvcc.try_load()` returns `None` for JIT-path modules (its comment states that it delegates JIT-path freshness to ninja), so `build_and_load()` always invokes ninja and the freshness decision rests entirely on mtime comparison. Resetting the copied sources' mtime therefore guarantees a full rebuild every time.

### 1. mtime of the copied sources

The source file mtime is constant. For `moe_bgmv_binding.cu` in the generated directory, across three separate processes:

With `shutil.copy` (current behaviour) — a different mtime each time, equal to the moment of copying:

```
run 1: 2026-08-27 16:08:17.288354236
run 2: 2026-08-27 16:11:53.781513130
run 3: 2026-08-27 16:15:32.347126118
```

With `shutil.copy2` — identical to the source, to the nanosecond:

```
run 1: 2026-08-27 15:51:52.591185647
run 2: 2026-08-27 15:51:52.591185647
run 3: 2026-08-27 15:51:52.591185647
```

### 2. ninja dry run, compiling nothing

Starting from an already-built cache and only changing *how* the sources are copied, then asking `ninja -n` for its decision:

```
step 0, untouched cache
  generated source mtime 16:15:32 / object mtime 16:15:37
  ninja: no work to do.

step 1, replay shutil.copy()
  generated source mtime 16:20:32 / object mtime 16:15:37   <-- sources now look newer
  ninja: 8 build actions pending
    bgmv_moe_moe_bgmv_binding.cuda.o
    bgmv_moe_moe_bgmv_bf16_bf16_bf16.cuda.o
    ... all 7 objects ...
    bgmv_moe.so

step 2, replay shutil.copy2()
  generated source mtime 15:51:52.591185647 / object mtime 16:15:37
  ninja: no work to do.
```

The decision flips back and forth with nothing but the copy call changing, which isolates mtime as the sole cause.

### 3. Wall time, three consecutive independent processes

Default release configuration, cache directory wiped before each arm. "rebuilt" was determined by comparing the mtimes of the 7 `.o` files plus `bgmv_moe.so`, not by parsing log text.

| | run 1 | run 2 | run 3 |
|---|---|---|---|
| `shutil.copy` (current) | 216.3 s, built | 218.5 s, all 8 artifacts rebuilt | 217.2 s, all 8 artifacts rebuilt |
| `shutil.copy2` | 217.3 s, built (cold cache, expected) | **4.3 s, nothing rebuilt** | **4.3 s, nothing rebuilt** |

## Reproduce

```bash
export FLASHINFER_WORKSPACE_BASE=/tmp/fi-cache
rm -rf "$FLASHINFER_WORKSPACE_BASE"

for i in 1 2 3; do
  /usr/bin/time -f "run $i: %e s" \
    python3 -c 'from flashinfer.jit.bgmv_moe import load_bgmv_moe_module; load_bgmv_moe_module()'
done
```

Every run takes the full compile time. Applying the patch below makes runs 2 and 3 return almost immediately.

## Suggested fix

```diff
--- a/flashinfer/jit/bgmv_moe.py
+++ b/flashinfer/jit/bgmv_moe.py
@@ -97,7 +97,7 @@
         dest_path = gen_directory / fname
-        shutil.copy(src_path, dest_path)
+        shutil.copy2(src_path, dest_path)
         sources.append(dest_path)
@@ -105,4 +105,4 @@
-        shutil.copy(src_path, gen_directory / fname)
+        shutil.copy2(src_path, gen_directory / fname)
```

The same pattern appears in two other JIT modules, which presumably have the same behaviour:

- `flashinfer/jit/rmsnorm_silu.py:420`
- `flashinfer/jit/monomoe.py:110` (structurally almost identical to `bgmv_moe.py`)

For what it is worth, `flashinfer/aot.py:953` already uses `shutil.copy2`, so the codebase is internally inconsistent on this point.

## Impact

Any setup that invokes `bgmv_moe` repeatedly in separate processes pays the full compile cost every time. We hit this in a benchmarking harness that runs the routine once per commit per configuration, where it added tens of minutes of pure recompilation per job and caused time-limited jobs to be killed.

## Environment

- flashinfer `0.6.18`, commit `9caa448`, built from source
- Blackwell, `sm_120`, `TORCH_CUDA_ARCH_LIST=12.0a`
- CUDA 13.2 driver, PyTorch 2.12 nightly
- Behaviour verified against current `main` by inspection; the two `shutil.copy` call sites are unchanged there

200lz:
## 📌 Description

Every new process that JIT-loads `bgmv_moe` rebuilt all kernels from scratch, even with a fully warm cache. `gen_bgmv_moe_module()` stages its sources and headers into `FLASHINFER_GEN_SRC_DIR` with `shutil.copy`, which stamps a fresh mtime on every copy. Since the JIT path delegates `.so` freshness entirely to ninja's dependency scan (`JitSpecNvcc.try_load()` returns `None` for non-AOT modules), the re-staged inputs always looked newer than the cached objects, so ninja recompiled the whole module on every warm start. #4782 reports 216–218 s per process on Blackwell versus 4.3 s once mtimes are preserved (as measured by the reporter).

**Fix:** stage with `shutil.copy2`, which preserves source mtimes, so unchanged inputs stay older than the built artifact and ninja no-ops. `shutil.copy` already propagated permission bits, so the only staging behavior that changes is timestamp preservation.

**Regression test:** `tests/jit/test_bgmv_moe_jit_cache.py` is deterministic and needs no GPU, nvcc, or sleeps. It runs `gen_bgmv_moe_module()` against a fake csrc tree with fixed old mtimes in a redirected workspace, simulates an already-built `.so` newer than the sources, clears the `functools.cache` to model a fresh process, re-stages, and asserts the JIT-cache freshness invariant: no staged source/header may become newer than the existing build artifact. It fails on current `main` (all 12 staged files get fresh mtimes) and passes with this change.

**Validation**
- Regression test: fails on `main`, passes with the fix.
- `tests/jit/test_jit_cpp_ext.py`: 18 passed; the 3 failures reproduce identically on unmodified `main` in my CPU-only environment (no CUDA toolkit installed) and are unrelated to this change.
- Warm-cache ninja check over the actual staged output of `gen_bgmv_moe_module()` (using a `cp` stand-in build graph, since nvcc is unavailable locally): after re-staging unchanged sources, `ninja -n` reports "no work to do" with this change, while on `main` it rebuilds all targets.
- `pre-commit run --files <changed files>`: all hooks pass (mypy, ruff check, ruff format, whitespace hooks).

## 🔍 Related Issues

Fixes #4782

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

- [ ] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues. *(ran `pre-commit run --files <changed files>` instead — all hooks pass; see Reviewer Notes)*

## 🧪 Tests

- [ ] All tests are passing (`unittest`, etc.). *(Targeted regression and changed-file checks pass; 3 CUDA-toolkit-dependent failures reproduce identically on unmodified main.)*

## Reviewer Notes

Hooks were run as `pre-commit run --files flashinfer/jit/bgmv_moe.py tests/jit/test_bgmv_moe_jit_cache.py` (the changed files) rather than `--all-files`, to avoid churning unrelated files. GPU kernel tests for bgmv_moe were not run locally (no CUDA device available); the added regression test is GPU-free by design.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved JIT compilation caching by preserving CUDA source and header file timestamps.
  * Prevented unnecessary full rebuilds when generating BGMV MoE modules from unchanged files.

* **Tests**
  * Added regression coverage to verify cached builds do not trigger avoidable recompilation.

### DIFF
--- a/flashinfer/jit/bgmv_moe.py
+++ b/flashinfer/jit/bgmv_moe.py
@@ -97,7 +97,8 @@ def gen_bgmv_moe_module():
         if not src_path.exists():
             raise FileNotFoundError(f"BGMV MoE source file not found: {src_path}")
         dest_path = gen_directory / fname
-        shutil.copy(src_path, dest_path)
+        # copy2 keeps source mtimes; a fresh mtime forces a full ninja rebuild (#4782)
+        shutil.copy2(src_path, dest_path)
         sources.append(dest_path)
 
     # Copy headers to gen directory
@@ -105,7 +106,7 @@ def gen_bgmv_moe_module():
         src_path = csrc_dir / fname
         if not src_path.exists():
             raise FileNotFoundError(f"BGMV MoE header file not found: {src_path}")
-        shutil.copy(src_path, gen_directory / fname)
+        shutil.copy2(src_path, gen_directory / fname)
 
     # Get nvcc flags for supported architectures (SM70+)
     nvcc_flags = current_compilation_context.get_nvcc_flags_list(
--- a/tests/jit/test_bgmv_moe_jit_cache.py
+++ b/tests/jit/test_bgmv_moe_jit_cache.py
@@ -1,0 +1,121 @@
+"""
+Copyright (c) 2026 by FlashInfer team.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+Regression test for https://github.com/flashinfer-ai/flashinfer/issues/4782
+
+The JIT path delegates .so freshness entirely to ninja's dependency scan
+(JitSpecNvcc.try_load returns None for non-AOT modules), so staging must not
+touch the mtime of unchanged sources: if gen_bgmv_moe_module() re-stages an
+unchanged source with a fresh timestamp, every new process makes the staged
+sources newer than the already-built objects and ninja rebuilds from scratch.
+"""
+
+import os
+
+import pytest
+
+import flashinfer.jit.bgmv_moe as bgmv_moe_jit
+from flashinfer.jit import core as jit_core
+from flashinfer.jit import env as jit_env
+
+# Filenames gen_bgmv_moe_module() expects in the bgmv_moe csrc directory.
+_SOURCE_FILES = [
+    "moe_bgmv_binding.cu",
+    "moe_bgmv_bf16_bf16_bf16.cu",
+    "moe_bgmv_bf16_fp32_bf16.cu",
+    "moe_bgmv_fp16_fp16_fp16.cu",
+    "moe_bgmv_fp16_fp32_fp16.cu",
+    "moe_bgmv_fp32_bf16_bf16.cu",
+    "moe_bgmv_fp32_fp16_fp16.cu",
+]
+_HEADER_FILES = [
+    "moe_bgmv_impl.cuh",
+    "moe_bgmv_config.h",
+    "moe_bgmv_ops.h",
+    "moe_bgmv_ops.cu",
+    "kernel_config.h",
+]
+
+# A fixed timestamp far in the past (2020-09-13T12:26:40Z), so that any staging
+# step that stamps "now" onto a copy is unambiguously detectable without sleeps.
+_OLD_MTIME_NS = 1_600_000_000_000_000_000
+
+
+class _StubCompilationContext:
+    """No-GPU stand-in so the generation pass needs no CUDA device."""
+
+    TARGET_CUDA_ARCHS = [(9, "0a")]
+
+    def get_nvcc_flags_list(self, supported_major_versions=None):
+        return []
+
+
+@pytest.fixture
+def bgmv_moe_workspace(tmp_path, monkeypatch):
+    """Fake csrc dir with old mtimes plus a redirected JIT workspace."""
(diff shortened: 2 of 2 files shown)


aeichler-ac:
Ran on GB10 (sm_121a) at 790e8283. test_bgmv_moe_jit_cache passes.

Three fresh processes, mtime on moe_bgmv_binding.cu did not budge.

Had the same copy2 patch locally for #4782. Your test is better than mine.

200lz:
Thanks for validating this independently on GB10/sm_121a — especially the fresh-process mtime check. That’s very helpful confirmation that the regression captures the actual cache-freshness behavior on another architecture.

Also good to know you arrived at the same `copy2` fix independently. Appreciate the cross-check.

bkryu:
@flashinfer-bot run

bkryu (approved):

PR #4751, opened 2026-08-26
perf(mla): optimize SM120 sparse MLA with swapped MMA operands

Lemon7-UP:
### Performance summary
Compared with the existing MG prefill kernel, the swapAB(swapped MMA operands) kernel delivers a **1.31–2.16× speedup** across the tested configurations (64/128 heads and 128–8192 query tokens). It achieves up to **1.95×** speedup with the `auto`
scale format and up to **2.16×** with `arbitrary_fp32`.

### Performance
GPU: NVidia RTX PRO5000, SM120
CUDA: 13.0

`python benchmarks/bench_sparse_mla_sm120.py
`(2048/8192 query tokens are not included in benchmark script)

| KV scale format | Heads | Tokens | Before (µs) | After (µs) | Speedup |
|---|---:|---:|---:|---:|---:|
| auto | 64 | 128 | 424.6 | 324.2 | 1.31× |
| auto | 128 | 128 | 682.7 | 420.4 | 1.62× |
| auto | 64 | 512 | 1377.3 | 802.8 | 1.72× |
| auto | 128 | 512 | 2522.8 | 1321.0 | 1.91× |
| auto | 64 | 2048 | 5169.7 | 2945.0 | 1.76× |
| auto | 128 | 2048 | 9784.6 | 5023.6 | 1.95× |
| auto | 64 | 8192 | 20320.3 | 11403.0 | 1.78× |
| auto | 128 | 8192 | 38686.8 | 19896.3 | 1.94× |
| arbitrary_fp32 | 64 | 128 | 553.6 | 356.9 | 1.55× |
| arbitrary_fp32 | 128 | 128 | 912.0 | 494.1 | 1.85× |
| arbitrary_fp32 | 64 | 512 | 1814.1 | 864.8 | 2.10× |
| arbitrary_fp32 | 128 | 512 | 3422.9 | 1601.5 | 2.14× |
| arbitrary_fp32 | 64 | 2048 | 6835.8 | 3264.5 | 2.09× |
| arbitrary_fp32 | 128 | 2048 | 13437.5 | 6220.8 | 2.16× |
| arbitrary_fp32 | 64 | 8192 | 26861.1 | 13077.9 | 2.05× |
| arbitrary_fp32 | 128 | 8192 | 53355.0 | 24661.7 | 2.16× |

### Motivation
[...]

PR #4732, opened 2026-08-25
[bug fix] Issue 3700 sparse mla sm121 hang

kahyunnam:
## 📌 Description

Sparse-MLA SM120 prefill uses CTA barrier 1 as an asymmetric handshake (count 384): 256 math threads `barrier.cta.arrive` (non-blocking), 128 IO threads `barrier.cta.sync`. Math can arrive for tile `ti`, run `ti+1`, and arrive **again** before IO syncs `ti`. The PTX ISA warns against exactly that (`arrive` then another `barrier.cta` on the same id before reset). The surplus arrival completes the phase without IO; a later phase starves. Captured on GB10: IO warps parked on `BAR.SYNC 1, 384` after math had exited.

Fix: alternate the handshake between ids **1 and 5** by tile parity (`bar_arrive_alt` / `bar_sync_alt`) at all five call sites. A run-ahead tile lands on the other id. Two ids suffice because math cannot start tile `ti` until IO has returned from the sync for `ti-2`, so at most one arrival is outstanding per id.

Ungated: this is an invalid arrival pattern, not an sm_121 quirk. The hang is only demonstrated on GB10; the kernel is SM12x-only.

## 🔍 Related Issues

Closes #3700

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

## 🧪 Tests

- [x] Tests have been added or updated as needed.
- [x] All tests are passing (`unittest`, etc.).

GB10 / sm_121: `pytest tests/attention/test_sparse_mla_sm120.py` → **309 passed**. Single-process soak on `heads=128, topk=1024, tokens=256`: baseline **5/5 hung** (usually tens of iterations, sometimes a few hundred); fix **5/5 × 30k, 0 hangs**. Two concurrent pytest (`-k "prefill_glm_nsa_arbitrary_fp32 or prefill_dsv4"`): baseline wedged in round 3; fix **12/12 clean**. All four instantiated prefill variants soaked 30k each (with `assert_close`); 30k bitwise-identical iterations vs a golden result (fix does not turn the hang into silent corruption). No new unit test — the race is a soak, not a unit.

[...]

PR #4802, opened 2026-08-28
Refactor Sparse MLA SM120

lucifer1004:
## Summary

Consolidated SM120 sparse-MLA rework, decode + prefill. All numbers on RTX PRO 6000 (SM120).

- **Faster decode, small T**: T=1 14.9 → **11.3µs** (−24%, graph replay, dual-cache 18K context); bitwise-identical outputs.
- **Calibrated dispatch**: an analytical `chunks_per_block` model + measured decode/prefill crossover replace the per-shape autotune sweep and the hard `T ≤ 64` cutoff (up to −61% on rerouted configs; tables below).
- **Continuous envelopes**: decode serves any `num_heads ∈ [1,128]` and any `topk ≥ min_topk`; prefill serves any `T ≥ 1` and any `topk % 64 == 0` width. Runtime-topk prefill drops instantiations **75 → 55**.
- **swapAB prefill** carried from #4751 behind a per-call `prefill_impl` override; independently re-benched at **1.12–2.37×** over MG.
- **Two new model types**: `GLM53_NOPE` (carried from #4791) and `DOTS3_SWA` (sliding-window MLA, d_qk=1088, d_v=1024, 1160 B/token footer-scale, padded-row KV support) — the latter also fixing five latent bugs along the way (rope writeback overrun at D_V==D_NOPE, flat-vs-paged addressing keyed on the wrong trait, a Python chunk-width hardcode, an undersized amax scratch at 4 math warps, a vestigial SG register array).
- **Public runner**: `flashinfer.mla.SparseMLASm120Wrapper` — one persistent instance, memoized dispatch, CUDA-graph-safe (decode scratch is routing-aware and instance-owned).
- Merged current main, incl. the #4732 SM121 prefill-hang fix.
- Also: row-strided `indices` (unblocks vllm-project/vllm#53574's persistent-buffer narrowing) and row-strided `out_lse`; T=0 decode returns empty instead of aborting; decode bindings now validate `out_lse`/index dtypes/dim0.

Carries (authorship preserved): #4461 zero-token decode (rewritten; XingSong), #4551 dispatch diagnostics + `supported_sparse_mla_sm120_configs()` (Sam Mausberg), #4751 swapAB (Lemon7-UP), #4791 GLM53_NOPE (lucamotz; extended with H=64/TP1 decode, swapAB@2176, calibration coverage). Supersedes #4683: its per-shape sweep profiles L2-resident synthetic indices, which distorts cpb when production caches are DRAM-resident (observed on 5070 Ti) — this PR removes the sweep instead (thanks Sam for the original analysis).

## Performance vs main (adc49a85)

Same GPU, fixed-seed identical inputs, CUDA-graph replay GPU-only, both sides out-of-box (no tactic cache / no calibrated constants). Only surfaces present on both sides listed.

| shape | main | PR | speedup |
|---|---|---|---|
| dsv4-dual-h64 (topk 128+512), T=1 | 14.80µs | 11.40µs | 1.30x |
| dsv4-dual-h64 (topk 128+512), T=8 | 19.80µs | 16.14µs | 1.23x |
| dsv4-dual-h64 (topk 128+512), T=16 | 36.66µs | 29.46µs | 1.24x |
| dsv4-dual-h64 (topk 128+512), T=64 | 97.19µs | 89.79µs | 1.08x |
| dsv4-h128 (topk 1024), T=1 | 14.70µs | 11.44µs | 1.29x |
| dsv4-h128 (topk 1024), T=64 | 231.60µs | 219.79µs | 1.05x |
| dsv3_2-h64 (topk 2048), T=1 | 14.08µs | 10.68µs | 1.32x |
| dsv3_2-h64 (topk 2048), T=64 | 216.78µs | 220.23µs | 0.98x |
| dsv3_2-h128 (topk 2048), T=1 | 16.57µs | 14.08µs | 1.18x |
| dsv3_2-h128 (topk 2048), T=64 | 324.19µs | 323.53µs | 1.00x |
| dsv4-prefill-h128 (topk 1024), T=128 | 293.82µs | 266.49µs | 1.10x |
| dsv4-prefill-h128 (topk 1024), T=2048 | 4486.89µs | 3928.68µs | 1.14x |
| dsv4-prefill-dual-h64 (topk 128+512), T=128 | 124.66µs | 124.64µs | 1.00x |
| dsv4-prefill-dual-h64 (topk 128+512), T=2048 | 1559.56µs | 1559.35µs | 1.00x |

Decode gains concentrate at small T (launch-bound); the two decode commits behind them: `quantize_q_to_smem` rewritten as a vectorized single pass (3 `bar.sync` → 1), and the decode-dsv4 IO gather reads each candidate's index once instead of twice. T=64 decode and dual-cache prefill are unchanged within noise.

## swapAB prefill (#4751)

Re-benched on the PRO 6000 (#4751's table was measured on a PRO 5000), same grid, MG↔swapAB cross-checked at 5e-2 on identical inputs, `auto` bitwise-identical to forced swapAB:

| shape | MG | swapAB | speedup |
|---|---|---|---|
| H=64, T=128 | 250.8µs | 159.7µs | 1.57× |
| H=64, T=512 | 798.7µs | 565.6µs | 1.41× |
| H=64, T=2048 | 2948.1µs | 2158.6µs | 1.37× |
| H=64, T=8192 | 11673.6µs | 8607.7µs | 1.36× |
| H=128, T=128 | 349.6µs | 267.9µs | 1.30× |
| H=128, T=512 | 1348.2µs | 840.0µs | 1.60× |
| H=128, T=2048 | 5330.9µs | 3011.6µs | 1.77× |
| H=128, T=8192 | 21156.9µs | 11847.7µs | 1.79× |

Wins everywhere; the H=64 large-T plateau (~1.4×, one CTA per token saturates ~1280 GB/s vs ~1860 at H=128) is a flat asymptote out to T=32768, so no dispatch range limit. KV layout and all parameters unchanged; both scale formats, sinks, and variable `topk_length` supported. `prefill_impl`: `"auto"` (default) / `"swapab"` / `"mg"`; forcing swapab at an ineligible shape raises.

## Dispatch: cpb model + crossover

**cpb model** — analytical pick over gather bandwidth/latency, per-block overhead, and the exact list-scheduling makespan of the split grid, with an L2-footprint guard rail (at topk=1024+2176 dual the heuristic picks a single 50-chunk block at 2.7× L2 — ncu: L2 hit 69.7% vs 86.8%, costing 33%; the guard recovers it to 1.02×). Calibrated once per device inside `autotune()` tuning mode (6 fixed measurements over a ~2 GiB pool, timed as queued batches over rotating fresh index sets — launch latency overlaps execution, and the batch length keeps each set's reuse distance past an L2 turnover; small numpy LM fit; any failure = silent fallback to the C++ heuristic, so the new path can't be worse than status quo). Offline pick error vs exhaustive sweep (DRAM-cold protocol): **mean 1.011× / max 1.061×**; beats the heuristic by up to **1.37×** at mid shapes. A GPU accuracy-guard test fails loudly if a future kernel change breaks the model's assumptions, measured with the same protocol the calibration runs. Host cost ~8µs/call, memoized; zero per-replay under CUDA graphs.

**Per-shape refinement** — the model's residual pick error concentrates at mid-T wave-quantization shapes (measured up to **1.35×**, e.g. DOTS3_SWA T=32: 78.0µs → 57.8µs). tuning-mode decode-form calls time the model pick ±6 candidates with the calibration protocol and persist the measured best as a per-shape override in the same tuning cache; `_resolve_cpb` consults overrides first, then the model. Across 12 production bucket shapes (T=16..64, three families, two-pass re-timing): **never worse than the model (12/12), closes every pocket to ≤1.03×**. Shapes never warmed (off-graph calls, arbitrary T, dual-cache) stay on the model. Capture-time calls only read the table/model and freeze — no measurement ever runs under graph capture or in serving.

**Crossover** — per-config `decode_max_tokens` measured during the same tuning pass (probe T ∈ {4..64}, both paths, DRAM-faithful fresh indices; decode wins iff ≤ 0.95× prefill). Uncalibrated behavior is unchanged. Measured examples:

| config | `decode_max_tokens` | Σ T∈{24,32,48,64}: old policy → calibrated |
|---|---|---|
| DSv3.2 H=128 topk=2048 (swapAB side) | 8 | 1271.4 → 494.0 µs (−61%) |
| DSv4 H=64 topk=512 | 24 | 292.3 → 216.7 µs (−26%) |
| DSv4 H=64 topk=128 | 16 | 132.3 → 96.7 µs (−27%) |
| DSv4 H=8 topk=1024 | 64 (decode dominates) | no rerouting |

Full per-probe data for all 71 calibrated configs: kernel-bench `crossover-v5` baseline. A public `calibrate_sparse_mla_sm120(device, heads=, topks=, families=, force=)` makes any envelope shape tunable outside tuning mode (idempotent skip-existing; `force=True` re-measures).

## Runtime envelopes (head counts and topk widths)

- **Decode**: any H ∈ [1,128] — dedicated instantiations on the production grid (0.9–2.5% faster), one runtime-H instance otherwise, **40/40 bitwise-identical** between the two. Any `topk ≥ min_topk` (1; 513 for DOTS3_SWA so the window fits). The `_DECODE_*_DISPATCH` objects vLLM probes are membership predicates with exactly this meaning; `supported_sparse_mla_sm120_configs()` exposes the envelopes for init-time validation. Off-grid example: H=80 T=16 is 1.14× faster than the pad-to-128 workaround callers needed before.
- **Prefill**: same topk rule across SG / MG / dual / swapAB. One deliberate residual asymmetry: **decode serves ragged widths (partial tail chunk, tested at topk=500); prefill requires whole 64-wide index tiles** — all production topk widths qualify, tail support needs predicated gathers + tail masking across the IO and math paths, and is deferred until a model needs it. This is safe at the routing layer: a ragged decode-form call has no prefill envelope and simply stays on decode (no crossover), and a ragged T>64 call fails loudly at the binding. 50-config parity vs the pinned build: worst **+0.94%**. One variant needed kernel-side help: DOTS3_SWA SG's BI=32 tiles are too short to cover the index→rope address-chain latency once the compile-time trip count disappeared (+24% `long_scoreboard` in NCU). The SG loop now stages the three per-tile index reads one tile ahead in registers, `if constexpr`-scoped to short tiles (unconditional staging taxed BI=64 SG +2.3%). Net: **374.6µs vs the pinned build's 380.7µs** at H=64/T=256, registers flat, `long_scoreboard` back to parity.

## Plan layer

All dispatch policy lives in one memoized Python planner (`_sparse_mla_sm120_plan.py`): each variant declares its envelope once, `plan()` picks by envelope + crossover + `prefill_impl`. The C++ side is a policy-free launcher registry (the old `dispatch_v32` chain is deleted). Single-sourcing surfaced two latent upstream bugs, fixed here: prefill launchers never checked `page_block_size` against the compiled 64 (silent wrong-stride launch), and dual-cache decode-form DSv3.2-family calls silently ignored the secondary cache.

## Runner and CUDA graphs

`SparseMLASm120Wrapper` holds buffers persistently: LSE pre-sized at construction, decode split-K scratch allocated only when the call actually routes to decode and cached for the instance's lifetime (a per-call temporary's freed block can be recycled into a later capture while an older graph replays into it). Capture contract: construct and warm up every captured shape before capture (or pass `out_lse`/scratch explicitly); replay is pure graph replay with zero Python. Both routing variants are correct for any T, so a crossover inside a padding bucket is at worst suboptimal, never wrong. GPU tests pin capture/replay for crossover dispatch and for runner-internal scratch.

## Compatibility

Public Python API: unchanged except additive kwargs; `flashinfer.mla` exports purely additive; no-constants path behaves exactly as today. Deliberate behavior changes:

- Per-shape tactic caches (`sparse_mla_sm120_decode_dsv{4,3_2}.json`) are ignored; the new calibration file is schema-versioned (v1), unrecognized versions treated as absent and recalibrated.
- `autotune(True)` runs a one-time-per-device calibration (~2 GiB transient pool) instead of profiling each new shape; honors `skip_ops={"sparse_mla_sm120"}`; refuses to run under CUDA graph capture; cache writes serialized with a FileLock.
- With calibration present, decode-form calls beyond the measured crossover route to prefill (the point of the feature).
- T ≤ 64 shapes outside the old fixed grid now take the runtime decode instantiation instead of raising.
- Prefill serves any `topk % 64 == 0` (≥ 513 for DOTS3_SWA); ragged widths fail at the binding.
- Inline-scale (DSv3.2/GLM) KV caches must be contiguous through the paged entry (prefill flat-addresses the cache and crossover makes routing dynamic); contiguous padded-row caches remain decode-served and fail loudly only if prefill-routed.
- `indices`/`out_lse` may be row-strided views (widening); the decode binding previously corrupted a strided `out_lse` silently.
- C++ launcher entries gained row-stride parameters — internal to the JIT module, no stable ABI consumers.

Out of scope (tracked follow-ups): H=64 swapAB bandwidth at large T; a pinned-topk fast path à la decode-H for DOTS3_SWA SG (locked clocks show ~2% there, boost clocks show nothing — not worth the instantiation axis on current evidence).

## Test plan

All on RTX PRO 6000: **658 passed** across `test_sparse_mla_sm120{,_dispatch,_cpb_model}.py` and `test_autotuner_core.py`, pre-commit clean — including the 68-config small-T prefill matrix vs the reference (T ∈ {1..64} × SG/MG/swapAB/dual × sink/truncation), 27 C++⟺Python envelope-consistency probes, runtime-H/topk parity gates (bitwise where required), crossover routing + CUDA-graph capture/replay tests, runner scratch routing/lifetime tests, and the review-round regression tests (row-strided `out_lse`, cpb save/publish/FileLock, grid-completeness gating, padded-cache rejection, skip_ops/capture guards).

This PR was prepared with AI assistance; all changes reviewed and tested locally by the submitter.


include/flashinfer/attention/sparse_mla_sm120/prefill_swapab_kernel.cuh
```diff
@@ -0,0 +1,452 @@
...
+    const int mwarp = warp_rank;
+    const int gid = lane >> 2, tid = lane & 3;
+    const float sm_scale_log2e = cold.sm_scale * LOG2E;
+    // B fragment column n == gid, so this lane feeds head h_base + gid; the C
+    // fragment hands back heads h_base + 2*tid and h_base + 2*tid + 1.
+    const int h_base = h_start + mwarp * HPW;
+    const bf16* q_base = Q + ((size_t)s_i * NUM_HEADS + h_base + gid) * KV::D_QK;
+
+    QSwapABRegs<MT> q = quantize_q_to_regs_swapab<MT>(q_base, lane);
+    KVRopePrefetch<MT> q_rope = prefetch_kv_rope<MT>(q_base + KV::D_NOPE, lane);
```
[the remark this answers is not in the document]
lucifer1004: Thanks for the careful look — this one is a false positive. All four rope helpers bound their loops by N_CHUNKS = KVCacheTraits<MT>::D_ROPE / 16, which is 0 for GLM53_NOPE, so the loads and MMAs are compile-time zero-trip: no gmem/smem access happens (the address arithmetic in the call arguments is never dereferenced), and the struct arrays use the N_CHUNKS > 0 guard so they compile. q_rope.cuh documents this 0-trip NoPE behavior. GLM53_NOPE prefill is covered by reference-checked tests through both the MG and swapAB paths, which would fail if rope bytes leaked into the scores.

csrc/sparse_mla_sm120_jit_binding.cu
```diff
@@ -54,28 +60,35 @@ inline PagedKVLayout parse_paged_kv_layout(const TensorView& kv, int bpt, const
...
   }
   if (kv.size(2) == 1) {
-    return {static_cast<int>(kv.size(1)), static_cast<size_t>(kv.stride(0)) * elem_bytes};
+    return {static_cast<int>(kv.size(1)), static_cast<size_t>(kv.stride(0)) * elem_bytes,
+            row_bytes(kv.size(-1))};
   }
   TVM_FFI_ICHECK(false) << name << " 4D form must have singleton KV-head axis at dim 1 "
                         << "(HND) or dim 2 (NHD)";
-  return {0, 0};
+  return {0, 0, 0};
```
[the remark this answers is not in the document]
lucifer1004: Good catch — fixed. The binding now rejects padded KV rows (stride_kv_row != bytes_per_token) on the footer-layout paths (DSV4 / DOTS3_SWA / extra cache) with a loud error; row-strided pages remain a decode-v32-only capability, and the prefill binding already required exact widths. Negative test added (test_sparse_mla_sm120_decode_dsv4_padded_row_rejected).

flashinfer/mla/_sparse_mla_sm120_plan.py
```diff
@@ -0,0 +1,572 @@
...
+}
+
+# Kernel-family names used in the public config query and error messages.
+_MODEL_TYPE_TO_FAMILY = {
+    _MODEL_TYPE_DSV3_2: "dsv3_2",
+    _MODEL_TYPE_DSV4: "dsv4",
+    _MODEL_TYPE_GLM_NSA: "glm_nsa",
+    _MODEL_TYPE_GLM53_NOPE: "glm53_nope",
+    _MODEL_TYPE_DOTS3_SWA: "dots3_swa",
+}
```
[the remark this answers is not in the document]
lucifer1004: Right — the refactor dropped the GLM_NSA->dsv3_2 alias. Fixed via _CPB_FAMILY_ALIAS: GLM_NSA shares the dsv3_2 calibration family (same kernel and ABI) for constants and calibration, while crossover keys stay per-model (scale format changes prefill speed). Regression test covers both the tuning path (no more ValueError) and the non-tuning constants pickup.

flashinfer/mla/_sparse_mla_sm120_cpb.py
```diff
@@ -0,0 +1,937 @@
...
+                    num_tokens, num_heads, d_qk, device=device, dtype=torch.float32
+                )
+                / 10.0
+            )
+            .clamp(-1, 1)
+            .to(torch.bfloat16)
+        )
+        indices = torch.randint(
+            0, num_slots, (num_tokens, topk), dtype=torch.int32, device=device
+        )
```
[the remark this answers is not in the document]
lucifer1004: Correct — the fresh-indices protocol only covered crossover calibration. calibrate() now redraws full-pool indices per rep. Recalibration confirms the L2 taint was real: inv_rsm moved +66%/+82% on dsv4/dsv3_2 (the latency measurement's working set is only ~0.6 MB, fully L2-resident under fixed indices), and the corrected constants produce 5-12% faster decode on part of the grid. Accuracy guards re-pass on the new constants.

csrc/sparse_mla_sm120_decode_dsv4.cu
```diff
@@ -71,11 +81,10 @@ static bool launch_decode_dsv4_impl(const bf16* Q, const uint8_t* KV_cache, cons
   // (ceil(waves) - waves). On ties, prefer the largest cpb so fewer
   // launched blocks contend on L2. The ceil_w cap rules out cpb values
   // whose fractional gap looks small but require many integer waves.
-  // AutoTuner can override per-shape via chunks_per_block_override.
+  // chunks_per_block_override: calibrated-model or explicit caller choice;
+  // the heuristic is the fallback when no override is given.
   int chunks_per_block;
   if (chunks_per_block_override >= 1 && chunks_per_block_override <= num_splits) {
```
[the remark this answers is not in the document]
lucifer1004: Fixed — the dsv4 launcher now mirrors the dsv3_2 guard: sm_count <= 0 after a successful query fails the launch instead of dividing by zero.

benchmarks/bench_sparse_mla_sm120.py
```diff
@@ -513,3 +517,18 @@ def fn():
         print(
             f"{h:>10}  {2048:>6}  {t:>11}  {lat_us:>10.1f}  {kvbw:>13.1f}  {tfl:>12.2f}"
         )
+
+    # DSv3.2 prefill: topk fixed at 2048, num_tokens > 64. Sweep is num_heads ×
+    # num_tokens × kv_scale_format; 64/128 heads run the swapAB kernel.
```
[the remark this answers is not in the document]
lucifer1004: Fixed (x instead of the multiplication sign).

include/flashinfer/attention/sparse_mla_sm120/prefill_common.cuh
```diff
@@ -0,0 +1,200 @@
...
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+// DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+// FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+// OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+
+#pragma once
+
```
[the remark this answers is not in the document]
lucifer1004: Fixed — completed the disclaimer on all three split prefill headers (prefill_common/prefill_mg/prefill_swapab all carried the truncated block).

csrc/sparse_mla_sm120_jit_binding.cu
```diff
@@ -54,28 +60,35 @@ inline PagedKVLayout parse_paged_kv_layout(const TensorView& kv, int bpt, const
...
   if (kv.size(1) == 1) {
-    return {static_cast<int>(kv.size(2)), static_cast<size_t>(kv.stride(0)) * elem_bytes};
+    return {static_cast<int>(kv.size(2)), static_cast<size_t>(kv.stride(0)) * elem_bytes,
+            row_bytes(kv.size(-1))};
   }
   if (kv.size(2) == 1) {
-    return {static_cast<int>(kv.size(1)), static_cast<size_t>(kv.stride(0)) * elem_bytes};
+    return {static_cast<int>(kv.size(1)), static_cast<size_t>(kv.stride(0)) * elem_bytes,
+            row_bytes(kv.size(-1))};
   }
```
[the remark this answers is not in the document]
lucifer1004: Fixed: the row advance now comes from the token axis's actual stride (dim 1 for the 3D/NHD forms, dim 2 for HND) with a last-dim contiguity check, instead of size(-1) * elem_bytes. A wider-row slice can no longer make the kernel step by the wrong stride; padded rows are still rejected loudly on the footer-layout decode path.

flashinfer/mla/_sparse_mla_sm120_cpb.py
```diff
@@ -0,0 +1,937 @@
...
+    device = torch.device(device)
+    key = _device_key_cache.get(device)
+    if key is None:
+        index = (
+            device.index if device.index is not None else torch.cuda.current_device()
+        )
+        name = torch.cuda.get_device_properties(index).name
+        key = f"{index}:{name}"
+        _device_key_cache[device] = key
+    return key
```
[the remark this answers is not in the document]
lucifer1004: Fixed: an unindexed torch.device("cuda") now resolves its index against the current device on every call and is cached under the resolved indexed device, so a set_device switch no longer serves the first-resolved device's constants.

flashinfer/mla/_sparse_mla_sm120_cpb.py
```diff
@@ -0,0 +1,937 @@
...
+                if family == _DECODE_MAX_TOKENS_KEY:
+                    if isinstance(raw, dict):
+                        _crossover[dev_key] = {str(k): int(v) for k, v in raw.items()}
+                    continue
+                _constants[(dev_key, family)] = CpbConstants(**raw)
+        _cache_mtime = mtime
+        _constants_version += 1
+    except (OSError, ValueError, TypeError, KeyError):
+        # Keep mtime unchanged so the next cold call retries.
+        pass
```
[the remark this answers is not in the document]
lucifer1004: Fixed: _maybe_load_disk now parses the whole document into locals and publishes to the process caches only after the full parse succeeds; a mid-document failure leaves nothing half-applied and keeps mtime/version untouched so the next cold call retries.

flashinfer/mla/_sparse_mla_sm120_cpb.py
```diff
@@ -0,0 +1,937 @@
...
+    except OSError:
+        return
+    if mtime <= _cache_mtime:
+        return
+    try:
+        payload = json.loads(path.read_text())
+        if payload.get("schema_version") not in _LOADABLE_SCHEMA_VERSIONS:
+            return
+        devices = payload["devices"]
+        if not isinstance(devices, dict):
```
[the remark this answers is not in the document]
lucifer1004: Fixed: all three adoption sites (_maybe_load_disk, save_constants, save_crossover) isinstance-guard the parsed JSON before .get, and the save paths normalize non-dict devices/device entries, so a structurally corrupt file counts as absent instead of escaping with AttributeError.

flashinfer/mla/_sparse_mla_sm120_plan.py
```diff
@@ -0,0 +1,579 @@
...
+        except (CalibrationError, torch.cuda.OutOfMemoryError) as e:
+            logger.warning(
+                "SM120 sparse-MLA %s crossover calibration failed (%s); "
+                "keeping the decode-first routing default for this process.",
+                family,
+                e,
+            )
+            _cpb.mark_crossover_failed(device, family)
+        else:
+            _cpb.save_crossover(device, table)
```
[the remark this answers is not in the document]
lucifer1004: Fixed: both _resolve_cpb calibration call sites now also catch RuntimeError, so an FFI launch failure during calibration falls back to the documented heuristic path instead of failing the attention call. Log wording unchanged.

tests/attention/test_sparse_mla_sm120_cpb_model.py
```diff
@@ -0,0 +1,628 @@
...
+    CpbConstants,
+    predict_time_s,
+    select_cpb,
+)
+from flashinfer.utils import is_sm12x_supported
+
+requires_sm12x = pytest.mark.skipif(
+    not is_sm12x_supported(torch.device("cuda")),
+    reason="Sparse-MLA SM120 requires SM12x.",
+)
```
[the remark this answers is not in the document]
lucifer1004: Fixed: the import-time guard now short-circuits on torch.cuda.is_available() before calling is_sm12x_supported, so CPU-only hosts can collect the module. Applied the same guard to test_sparse_mla_sm120.py's pytestmark, which had the identical pattern.

flashinfer/mla/_sparse_mla_sm120.py
```diff
@@ -1231,30 +1106,33 @@ def sparse_mla_sm120_decode_dsv4(
...
+      (calibrated once per device in ``autotune()`` tuning mode and cached on
+      disk) → use the model's choice.
     - Otherwise → fall back to the C++ closed-form heuristic.
 
     Parameters
     ----------
     q : torch.Tensor
-        ``[T, num_heads, d_qk]`` bf16. ``d_qk == 512`` (DSV4 only).
+        ``[T, num_heads, d_qk]`` bf16. ``d_qk == 512`` (DSV4) or
+        ``d_qk == 1088`` (DOTS3_SWA; d_v is then 1024).
```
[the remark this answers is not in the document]
lucifer1004: Fixed: the indices doc now lists topk=576 for DOTS3_SWA and notes its 32-wide chunk width for the num_splits computation.

lucifer1004:
Addressed both outside-diff comments from the 2026-08-28 review in c2eec117:

1. **`SparseMLASm120DecodeConfig.d_qk` family coverage** — fixed: the attribute now names all five families (512 for DSv4/GLM53_NOPE, 576 for DSv3.2/GLM-NSA, 1088 for DOTS3_SWA, noting its d_v=1024). The "also applies to" `d_v` spot was already updated in an earlier cleanup round and now reads "512 for DSV3_2 / DSV4 / GLM variants, 1024 for DOTS3_SWA".

2. **Secondary-cache argument pairing** — confirmed reachable: `extra_indices` without `extra_kv_cache` reached the planner as `has_extra=False` with `extra_topk>0`, so a single-cache variant could be selected while the extra indices were forwarded with a null cache. `_sparse_mla_sm120_paged_attention` now rejects mismatched `extra_kv_cache`/`extra_indices` and a lone `extra_topk_length` with a `ValueError` before planning/module load (same pairing style as the `_core.py` sparse-backend checks). Covered by `test_extra_cache_args_pairing`, which exercises the runner's call path on CPU tensors (no GPU/JIT needed).

lucamotz:
Given that GLM-5.3-Flash support is actively being upstreamed in vLLM (#53906), would it help review velocity if I extracted the GLM53_NOPE portion carried from #4791 into a small standalone PR?

I originally implemented and validated that path on GB10/SM121. I could base the standalone version on the current #4802 implementation so it includes the subsequent H=64/TP1 and prefill improvements, while leaving the planner/calibration/general sparse-MLA refactor here.

Happy to do that if maintainers/the author prefer; otherwise I’ll leave it consolidated here to avoid duplicate review work.

ima-helikoptaaa:
Hey @lucifer1004, let me know if any help is required in terms of testing this on sm_120.
In case we are moving forward with the extracted the GLM53_NOPE in a separate PR direction then I can also validate just that bit with vLLM (#53906) on sm_120 @lucamotz.

Would be helpful to understand the direction this is headed, as this is a blocker for running GLM5.3-Flash on vLLM on sm_120

lucamotz:
> Hey @lucifer1004, let me know if any help is required in terms of testing this on sm_120. In case we are moving forward with the extracted the GLM53_NOPE in a separate PR direction then I can also validate just that bit with vLLM (#53906) on sm_120 @lucamotz.
> 
> Would be helpful to understand the direction this is headed, as this is a blocker for running GLM5.3-Flash on vLLM on sm_120

Hey! I have already validated this whole PR on SM121 and I’ve already prepared a draft of only the bits needed for GLM 5.3-Flash. That’s also fully tested as well. I think it might be worth moving forward with that and leaving the big changes in this PR separate. But even those big changes are confirmed working

Edit: @ima-helikoptaaa PR #4842 is now marked for review if you want to test on sm_120. Would be cool if we could move forward a bit quicker and make it a bit easier for the reviewers haha

ima-helikoptaaa:
> Edit: @ima-helikoptaaa PR #4842 is now marked for review if you want to test on sm_120. Would be cool if we could move forward a bit quicker and make it a bit easier for the reviewers haha

Yes, will test on sm_120 and let you know if everything is working fine.

bkryu:
@flashinfer-bot run

flashinfer/mla/_sparse_mla_sm120_plan.py
```diff
@@ -0,0 +1,603 @@
...
+
+def _decode_scratch_heads(num_heads: int) -> int:
+    """Head rows the split-K decode scratch must carry for ``num_heads``.
+
+    The dedicated num_heads=8 instantiation strides mid_out/mid_lse by the
+    true head count; the runtime-H instantiation (all other eligible counts)
+    HPB-aligns the scratch so both halves of the 16-head tile always exist.
+    """
+    if num_heads == 8:
+        return 8
```
bkryu, line 205: `GLM53_NOPE` has no dedicated H=8 instantiation. I'd think H=8 (e.g. a 64-head layer at TP8, which decode_splitk_eligible accepts) falls to the runtime-H kernel. Can you double check this?

flashinfer/mla/_sparse_mla_sm120.py
bkryu, line 935: Comment from an agent. Can you check?
When the caller pre-sizes out_lse at (max_num_tokens, max_num_heads) — the documented wrapper usage — this returns a column-sliced, non-contiguous view, but the decode FFI takes data_ptr() with no stride check and the merge kernel writes LSE densely (decode_dsv4_kernel.cuh:1036: out_lse[t_idx*q_heads + h]). With max_num_heads=128 and num_heads=64, the kernel writes at flat t*64+h while the returned view reads t*128+h: every row t>0 is garbage. (The prefill route trips CHECK_CONTIGUOUS on the same view and hard-errors instead.) Either require/copy to a contiguous view here, or pass the row stride through to the kernels. Same issue on the auto-allocated slice at line 891 when a prior larger call cached self._out_lse.

include/flashinfer/attention/sparse_mla_sm120/prefill_swapab_kernel.cuh
```diff
@@ -0,0 +1,453 @@
...
+                                                           uint64_t* mbar, int io_tid,
+                                                           uint64_t cache_policy) {
+  constexpr int STRIDE = SmemLayoutSwapAB<MT>::KV_STRIDE;
+
+  if (io_tid == 0) mbarrier_arrive_expect_tx(mbar, BI * STRIDE);
+
+#pragma unroll 1
+  for (int bi = io_tid; bi < BI; bi += IO_THREADS) {
+    const int idx = indices[bi];
+    const uint8_t* src = kv_ptr + (size_t)(idx >= 0 ? idx : 0) * STRIDE;
```
bkryu: This gather addresses the KV cache flat, ignoring the block stride — but the orchestrator explicitly accepts padded block strides (csrc/sparse_mla_sm120.cu:161: "padded block stride is OK") and decode handles them correctly via block_idx*stride_kv_block + local_idx*stride_kv_row.

Can you check this?

flashinfer/mla/_sparse_mla_sm120_cpb.py
```diff
@@ -0,0 +1,1261 @@
...
+        logger.warning(
+            "SM120 sparse-MLA cpb constants not persisted to %s (%s); "
+            "using them in-process only.",
+            path,
+            e,
+        )
+    _constants[key] = c
+    _constants_version += 1
+    with contextlib.suppress(OSError):
+        _cache_mtime = path.stat().st_mtime
```
bkryu, line 1041: The save merges pre-existing on-disk entries into the rewritten JSON but never publishes them into the in-process _constants/_crossover dicts, then consumes the new  file's mtime — so _maybe_load_disk (line 835: if mtime <= _cache_mtime: return) permanently refuses to load them for the rest of the process. E.g.  calibrate_sparse_mla_sm120(families=('dsv4',), force=True) as the first cpb activity makes get_constants(device, 'dsv3_2') return None for the process lifetime — silent heuristic fallback or a pointless recalibration. Fix: publish the merged payload into the in-process dicts before updating _cache_mtime.

flashinfer/mla/_sparse_mla_sm120_cpb.py
```diff
@@ -0,0 +1,1261 @@
...
+    path = default_cache_path()
+    payload = _read_payload_for_merge(path)
+    if not isinstance(payload.get("devices"), dict):
+        payload["devices"] = {}
+    if not isinstance(payload["devices"].get(key[0]), dict):
+        payload["devices"][key[0]] = {}
+    payload["devices"][key[0]][family] = asdict(c)
+    try:
+        path.parent.mkdir(parents=True, exist_ok=True)
+        tmp = path.with_name(path.name + ".tmp")
```
bkryu: Do we want to use FileLock like we do in `flashinfer/jit/core.py`? I wonder whether this will be a concern in multi-GPU scenarios

csrc/sparse_mla_sm120.cu
bkryu: Python's _packed_kv_page_block_size accepts padded KV rows (≥ bytes_per_token) for all model types, but prefill requires the last dim to equal bytes_per_token exactly, and plan() carries no row-stride info. A padded-row cache (this PR's headline decode feature) serves all decode-form batches, then the first prefill-routed call raises a generic "3D form must be […]" error that never mentions padded rows are decode-only. Suggest surfacing the restriction at plan time (and in the wrapper docstring) with a targeted message.

flashinfer/mla/_sparse_mla_sm120_plan.py
```diff
@@ -0,0 +1,603 @@
...
+    c = _cpb.get_constants(device, cpb_family)
+    if (
+        c is None
+        and AutoTuner.get().is_tuning_mode
+        and not _cpb.is_calibration_failed(device, cpb_family)
+    ):
+        from ._sparse_mla_sm120 import _get_sparse_mla_sm120_decode_module
+
+        try:
+            c = _cpb.calibrate(_get_sparse_mla_sm120_decode_module, cpb_family, device)
```
bkryu, line 459: This calibration trigger piggybacks on tuning mode but bypasses the AutoTuner contract: skip_ops only gates choose_one so it can't exclude this multi-second  calibration; there's no is_current_stream_capturing guard while calibrate() does torch.cuda.synchronize/empty_cache and ~2 GiB allocations (capture-fatal if a framework captures graphs during tuning mode); and results go to a private JSON that autotune(cache=...) neither persists nor replays — fused_moe/da_runtime.py's  namespaced-records mechanism handles exactly this. Suggest a capture guard at minimum, and consider routing persistence through the AutoTuner cache.

flashinfer/mla/_sparse_mla_sm120_plan.py
```diff
@@ -0,0 +1,603 @@
...
+            )
+            _cpb.mark_calibration_failed(device, cpb_family)
+        else:
+            _cpb.save_constants(device, cpb_family, c)
+    if (
+        c is not None
+        and AutoTuner.get().is_tuning_mode
+        and not _cpb.is_crossover_failed(device, family)
+        and not _cpb.has_crossover(device, family)
+    ):
```
bkryu, line 476: has_crossover (cpb.py:997-999) returns True if any key with the family prefix exists, so a single previously-persisted targeted entry (e.g. a one-off calibrate_sparse_mla_sm120(heads=(48,), topks=(256,))) silently suppresses calibration of the family's entire production grid on later full autotune passes — every production shape keeps uncalibrated decode-first routing with no warning. Since calibrate_crossover writes whole grids, consider gating on grid completeness (or per-entry presence) instead of prefix existence.

flashinfer/mla/_sparse_mla_sm120_plan.py
```diff
@@ -0,0 +1,603 @@
...
+
+
+def prefill_sg_eligible(
+    model_type: int, num_heads: int, topk: int, page_block_size: int, has_extra: bool
+) -> bool:
+    if model_type == _MODEL_TYPE_DOTS3_SWA:
+        return (
+            not has_extra
+            and page_block_size == _PAGE_BLOCK_SIZE
+            and topk == _DOTS3_SWA_TOPK
```
bkryu: Decode accepts any topk >= 513 for DOTS3_SWA (line 264 via the envelope at line 166), but prefill — SG-only for this family, no MG/swapAB fallback — requires topk == 576 exactly. A deployment running topk in [513, 575] works while batches stay in the decode envelope, then gets a dispatch-miss ValueError the moment num_tokens crossesinto the prefill regime. Either accept the same topk >= 513 range here or reject topk ≠ 576 in the decode envelope so the failure is immediate and consistent.

csrc/sparse_mla_sm120_decode_dsv4.cu
```diff
@@ -99,88 +116,105 @@ static bool launch_decode_dsv4_impl(const bf16* Q, const uint8_t* KV_cache, cons
...
+// tile with zero-Q rows and HPB-aligns the mid scratch.
+bool launch_sparse_mla_decode_dsv4(
+    ModelType mt, int num_heads, int topk, int page_block_size, int num_tokens, int num_splits,
+    const bf16* Q, const uint8_t* KV_cache, const int32_t* indices, bf16* mid_out, float* mid_lse,
+    bf16* output, float* out_lse, const int* topk_length, const float* attn_sink,
+    const uint8_t* extra_KV_cache, const int32_t* extra_indices, const int* extra_topk_length,
+    int extra_topk, int pbs_extra, size_t stride_extra_kv_block, int chunks_per_block_override,
+    float sm_scale, size_t stride_kv_block, size_t stride_indices_token,
+    size_t stride_extra_indices_token, cudaStream_t stream) {
+  if (mt != ModelType::DSV4 && mt != ModelType::DOTS3_SWA) return false;
```
bkryu, line 168: The planner comments that DOTS3_SWA has "no dual-cache form for this family," but that's enforced only in Python — this launcher passes extra_KV_cache/extra_indices straight into the DOTS3_SWA instantiations with no rejection (contrast the prefill MG_DUAL path, which requires mt == DSV4 for dual-cache). A direct FFI caller with d_qk=1088 and a non-null extra cache silently runs an untested path. Suggest if (mt == ModelType::DOTS3_SWA && extra_KV_cache != nullptr) return false;.

bkryu (commented):
Thanks @lucifer1004 I gave an agent-assisted pass and left some comments.

csrc/sparse_mla_sm120_jit_binding.cu
```diff
@@ -135,11 +177,21 @@ void SparseMlaSm120DecodeDsv4(TensorView q, TensorView kv_cache, TensorView indi
...
   size_t stride_extra_kv_block = 0;
+  size_t stride_extra_indices_token = 0;
   if (extra_kv_cache.has_value()) {
     TVM_FFI_ICHECK(extra_indices.has_value()) << "extra_kv_cache requires extra_indices";
     const auto& ekv = extra_kv_cache.value();
-    extra_topk_arg = static_cast<int>(extra_indices.value().size(-1));
-    const PagedKVLayout extra_layout = parse_paged_kv_layout(ekv, BPT_DSV4, "extra_kv_cache");
+    const auto& eidx = extra_indices.value();
+    CHECK_INPUT_TYPE(eidx, dl_int32);
+    TVM_FFI_ICHECK_EQ(eidx.stride(-1), 1) << "extra_indices last dimension must be contiguous";
```
saltyminty, line 200: It seems we're missing a check here to determine that `indices` has dimension 0 size T (same as `q`) – prefill has this guard at sparse_mla_sm120.cu:133:
`TVM_FFI_ICHECK_EQ(idx.size(0), num_tokens) << name << " leading dimension must match num_tokens";`

csrc/sparse_mla_sm120.cu
saltyminty: +1

flashinfer/mla/_sparse_mla_sm120.py
```diff
@@ -1231,30 +1144,34 @@ def sparse_mla_sm120_decode_dsv4(
...
-        {128, 192, 256, 512, 1024}; ``-1`` marks invalid slots.
+        ``[T, topk]`` int32. Any ``topk >= 1`` for DSV4 (any ``topk >= 513``
+        for DOTS3_SWA: the 513-token sliding-window floor; tiled 32-wide, so
+        ``num_splits`` uses the 32-candidate chunk width); ``-1`` marks
+        invalid slots. Row-strided views into a wider persistent buffer
+        are accepted (the last dim must stay contiguous).
     mid_out : torch.Tensor
         Scratch, ``[T, num_heads, num_splits, d_v]`` bf16. ``num_splits =
-        ceil(topk / 64) + ceil(extra_topk / 64)``.
+        ceil(topk / 64) + ceil(extra_topk / 64)`` (64-wide candidate tiles;
```
saltyminty, line 1271: Non-blocking, but the head dimension of `mid_out` and `mid_lse` is expected to either be 8, or padded to the nearest multiple of 16 (enforced in `_decode_scratch_heads`).

lucifer1004:
@lucamotz Thanks for offering — and for validating on SM121. We'd prefer to keep GLM53_NOPE consolidated here: the path is fully validated in this PR's matrix (incl. the H=64/TP1 and swapAB extensions), CI is green on the current tip, and we're turning review around quickly, so the split shouldn't be load-bearing for vLLM#53906 for long. If maintainers would rather land #4842 first to unblock vLLM, that's fine by us — it's based on this implementation, so nothing conflicts semantically; we'll rebase the remnant out.

lucifer1004:
Round 2 is up. Merged current upstream main (incl. the #4732 SM121 prefill-hang fix, ported onto the split prefill headers) and addressed all of @bkryu's and @saltyminty's threads:

- out_lse is row-strided end to end (decode previously corrupted a wide caller buffer silently);
- padded inline-scale caches fail up front at the entry, with binding-level defense;
- cpb cache save/publish/FileLock/grid-completeness fixed; calibration honors skip_ops and refuses capture;
- prefill is runtime-topk, so the DOTS3_SWA decode/prefill envelope asymmetry is gone (75 -> 55 instantiations);
- runner decode scratch is plan-aware (allocated only for decode-routed calls) and instance-owned (CUDA-graph-pool safe);
- DOTS3_SWA SG prefill's off-pin regression was root-caused (lost compile-time trip-count pipelining) and fixed with register-staged index reads, scoped to short tiles — now ahead of the pinned build.

656 tests pass, pre-commit clean; prefill parity within +0.94% on the 50-config grid. Could someone please `/bot run` when convenient?

flashinfer/mla/_sparse_mla_sm120_plan.py
lucifer1004: Confirmed — GLM53_NOPE has dedicated 32/64 instantiations only; H=8 (like any other count in [1, 128]) is served by the runtime-H instantiation (`DSV3_2_DISPATCH_RT_MT(GLM53_NOPE)` in the decode-v32 launcher). Verified numerically on GPU against the reference, and H=8 is now in the parametrized decode test so the path is locked in (61267415).

flashinfer/mla/_sparse_mla_sm120.py
lucifer1004: Real bug — reproduced: with a (max_tokens, 128) buffer and a num_heads=64 call, decode wrote LSE densely at t*64+h into the strided view, so every row past the first read back garbage (output unaffected); prefill raised on the same shape. Fixed end to end: out_lse now carries a row stride through the prefill cold params and the shared decode merge kernel (same pattern as the indices row stride), the bindings check stride(-1) == 1, and the decode bindings validate out_lse at all now. Regression tests cover both the constructor-pre-allocated wide buffer and a caller-passed one (31bb01a0).

include/flashinfer/attention/sparse_mla_sm120/prefill_swapab_kernel.cuh
lucifer1004: Checked — not swapAB-specific: every inline-scale (DSv3.2/GLM) prefill kernel addresses the cache as a flat token array by design (prefill_kv_entry_base; the idiom predates this PR — upstream #3395). Only footer-scale models (DSv4/DOTS3_SWA) are block-stride-aware, because their per-block scale footer forces the decomposition anyway. The real hole was silent acceptance: with crossover, any decode-form call can route to prefill, so a padded-stride cache would misread at an unpredictable batch size. Now rejected up front at the Python entry, with the prefill binding ICHECK as ABI-level defense (837812c6, 097a3398). Contiguous padded-row caches stay decode-served (decode-v32 honors stride_kv_row); a prefill-routed call fails with a message naming the decode-only restriction.

flashinfer/mla/_sparse_mla_sm120_cpb.py
lucifer1004: Real — reproduced with an uncontaminated family pair. save() now parses the merged document it just wrote and publishes all of it in-process (shared parse helper with _maybe_load_disk), so _cache_mtime truthfully reflects the held content (e9285a22). Regression test added.

flashinfer/mla/_sparse_mla_sm120_cpb.py
lucifer1004: Agreed — concurrent multi-GPU tuning processes save sibling device keys into the same file via read-modify-write, so a lost update was a real (if benign-ish: recalibration) race. The merge+replace now runs under a FileLock next to the cache, following the jit/core.py precedent; reads stay lock-free thanks to the atomic replace (53472f6f).

csrc/sparse_mla_sm120.cu
lucifer1004: Done as far as the routing layer allows: the prefill binding's row-width error now names the decode-only restriction explicitly, and the paged_attention docstring states the contract. We deliberately did NOT reject padded-row caches at plan time: decode-v32 serves them (the shared-cache-group use case is real and tested), and plan() has no stride metadata — see the swapAB thread for the contiguity rule we do enforce up front (097a3398).

flashinfer/mla/_sparse_mla_sm120_plan.py
lucifer1004: Two of the three points are now handled (53472f6f):
- `autotune(skip_ops={"sparse_mla_sm120"})` opts out of the calibration passes too (read from the effective skip stack at the trigger), documented in the public calibration API.
- calibrate()/calibrate_crossover() raise CalibrationError under CUDA graph capture (they synchronize, empty the allocator cache, and allocate GiB-scale pools — capture-fatal), which the trigger already catches and falls back from.

On the private JSON: we kept it on purpose. The calibration store is device-keyed, shared across processes and runs, and mtime-gated — its lifecycle is "this machine", not "this tuning session", which is what the autotuner record cache models. Happy to revisit unification separately if you feel strongly.

flashinfer/mla/_sparse_mla_sm120_plan.py
lucifer1004: Fixed: the tuning-mode trigger now requires every (num_heads, topk) pair of the family's calibration grid (both key spaces for dsv3_2) via crossover_grid_complete(); a one-off targeted calibration no longer suppresses the full sweep. has_crossover keeps its loose semantics for introspection (553a2685).

flashinfer/mla/_sparse_mla_sm120_plan.py
lucifer1004: Resolved by removing the asymmetry rather than pinning: prefill is now runtime-topk — the TOPK template parameter only ever drove the indices row stride, the topk_length default, and the tile count, so all prefill variants serve any topk % 64 == 0 width (>= 513 for DOTS3_SWA). The decode and prefill envelopes can no longer disagree mid-stream, and the DSv4 MG topk instantiation axis collapsed (75 -> 55 module instantiations). Off-pin correctness tests + a C++<->Python envelope-consistency probe added (f56f7f3c, f3d9cfb3).

One kernel note: DOTS3_SWA SG initially lost ~5-7% off-pin — the compile-time trip count had let nvcc pipeline the math warps' per-tile index/rope-address loads across the loop barriers, and its BI=32 tiles are too short to cover that latency. The SG loop now stages those index reads one tile ahead in registers, scoped by `if constexpr` to short-tile configs (BI=64 tiles don't need it and were measurably taxed by it): 374.6µs vs the pinned build's 380.7µs at H=64/topk=576/T=256, long_scoreboard back at parity, registers/thread unchanged.

csrc/sparse_mla_sm120_decode_dsv4.cu
lucifer1004: Done: the binding rejects it with a clear message and the launcher returns false as the ABI-level guard (d69af14d). Negative test added.

csrc/sparse_mla_sm120_jit_binding.cu
lucifer1004: Added to both decode bindings (and extra_indices), mirroring the prefill guard (d69af14d). Regression test via the routed entry.

flashinfer/mla/_sparse_mla_sm120.py
lucifer1004: Documented at the run() parameters: exactly 8 for num_heads==8, else padded to a multiple of 16 (f59f1ac8). Two related fixes landed on top: the runner now plans the call first and allocates internal scratch only when it routes to a decode kernel (e11b88f3), and the internal scratch is cached on the instance (grow-only) instead of freed per call, closing the CUDA-graph pool hazard where a freed scratch block could be recycled into a later capture while an older graph replays into it (613002f8). New tests cover both, incl. capture/replay with runner-internal scratch.

Defilan:
Tested #4802 at `5e4980c2` on the same two DGX Spark (GB10) nodes, same model (`deepseek-ai/DeepSeek-V4-Flash-Vision-Exp`), same vLLM (`0.28.1rc1.dev137+g5ab628dd1`, the `dsv4-vision` branch image). Overlay method: the PR head's whole `flashinfer/` Python tree plus `csrc/` and `include/flashinfer/` copied over the wheel's FlashInfer 0.6.18 (wheel cutlass and the other AOT modules kept), the prebuilt `sparse_mla_sm120.so` removed so the module JIT-builds. A partial overlay of only `flashinfer/mla/` fails at import (`No module named 'flashinfer.mla._batch_mla'`), for anyone else trying this against a 0.6.18 wheel.

**Serves with no dispatch arms at all.** TP2 + EP over RoCE, `--kv-cache-dtype fp8 --block-size 256 --max-model-len 262144`, no speculative decoding:

| | #4850 @7a55473b9 | #4802 @5e4980c2 |
|---|---|---|
| `sparse_mla_sm120` JIT time | ~120 s | **18 s** |
| KV cache (tokens) | 562,864 | 538,813 |
| prefill tok/s, 3.7k prompt | 2,089 | 2,103 |
| decode tok/s, single stream | 24.0 | 25.4 |
| prefill tok/s, 28k prompt | 2,151 | 2,133 |
| decode tok/s at 28k | 23.4 | 25.2 |

Image prompts and tool calls unchanged. vLLM's `has_flashinfer_sparse_mla_sm120_config(32, 512)` still returns true against the new `_DecodeDispatchEnvelope` (it only needs `in`), so the probe surface holds. Three runs per cell, greedy, median; the two configs are within noise of each other on speed, and #4802 wins on build time and generality. Thanks @lucifer1004.

saltyminty (approved):

Hefulalala:
**Data point: NH=16 (TP4) production serving on SM120, PCIe — the envelope works**

Thanks for the refactor — the runtime-topk design removing the enumerated dispatch arms is exactly what we needed. Sharing a production validation at a head count none of the prior #4850 reporters covered (TP1 → H=64, TP2 → H=32; we run TP4 → H=16 per rank).

**Setup:** 4× RTX PRO 6000 Blackwell Server (SM120, cc 12.0), PCIe only, no NVLink. `deepseek-ai/DeepSeek-V4-Flash-Vision-Exp`, TP=4 + EP, `--kv-cache-dtype fp8 --block-size 256 --max-model-len 524288` (also validated at 1048576), DSpark k=3, vLLM main (`0.28.1rc1.dev349+gee0a4c46a`, post-#54566).

Since this PR isn't merged, we ran stock `flashinfer-python 0.6.18` with a local backport of the same envelope onto the current dispatcher (dual-cache CM arms at `topk=512`, `extra_page_block_size ∈ {64, 2}`, `num_heads=16`). Two notes that may be useful:

1. We instantiated `ComputeMode::FP8` arms at `<DSV4, FP8, 16, 512, 64, {64|2}, 1>` — consistent with the stock single-cache dispatcher in this file, which already uses FP8 for its 512 arms (verified in the 0.6.18 module symbols). #4850's dual arms used BF16 at H=32/64; both compile and run at 512 in the respective reports, in case that informs the sizing rule you land on.
2. The first real image prefill is the only trigger — warmup with no image never widens the window, so `topk` stays 128 and serving looks healthy until the first multimodal request. Worth a regression test covering a vision-widened prefill at each head count.

**Validated in production:** 96-conc text 180s and 32-conc with 30% image traffic 120s (both at `--max-model-len 131072`) — zero `Unsupported sparse-MLA` / zero errors; 1M-context serving validated separately incl. 60K-token+image prefills (~21 s). DSpark k=3 mean accepted length ≈ 2.87 tok/step (62% acceptance); TPOT p50 110 ms at conc 32. Image-size matrix from 100×80 (upscale path) to 12 MP (token-capped) including 25:1 aspect-crop paths all pass; 128 GiB `SimpleCPUOffloadConnector` stable.

Looking forward to this landing so we can drop the local patch — happy to run this branch's csrc through the same production workload as an independent check if useful.

bkryu:
@flashinfer-bot run

bkryu (approved):

madalin-dogaru:
Tested against my own config based on the previous PR: 
```
   Test                                                 Baseline      Candidate                Difference
  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━  ━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━
   Controlled 256-token prompt, median decode         47.1 tok/s     48.2 tok/s                     +2.3%
  ────────────────────────────────────────────────  ─────────────  ─────────────  ────────────────────────
   Natural high-reasoning workload, median decode    31.88 tok/s    31.34 tok/s                     -1.7%
  ────────────────────────────────────────────────  ─────────────  ─────────────  ────────────────────────
   Fresh 131K prompt TTFT                               71.880 s       69.611 s                     -3.2%
  ────────────────────────────────────────────────  ─────────────  ─────────────  ────────────────────────
   Fresh 131K prompt decode                           42.7 tok/s     53.9 tok/s                    +26.2%
  ────────────────────────────────────────────────  ─────────────  ─────────────  ────────────────────────
   MTP acceptance, 20 trials                               78.4%          76.2%    -2.2 percentage points
  ────────────────────────────────────────────────  ─────────────  ─────────────  ────────────────────────
   Warm vision regression suite                      8/8, 4.15 s    8/8, 4.07 s     Effectively unchanged
  ────────────────────────────────────────────────  ─────────────  ─────────────  ────────────────────────
   Tool-calling suite                                        7/7            7/7             No regression
```

  **### MTP acceptance by position**

  Baseline, 20 trials:
  - Overall: 78.4% (1790/2283)
  - Position 0: 87.6%
  - Position 1: 76.1%
  - Position 2: 71.5%

  Candidate, 20 trials:
  - Overall: 76.2% (1769/2322)
  - Position 0: 87.9%
  - Position 1: 73.1%
  - Position 2: 67.6%
  
# Testing Configuration 

  ### Hardware

  - 2x ASUS Ascent GX10
  - NVIDIA GB10 GPU on each node
  - 128 GB unified memory per node (approximately 121 GiB visible to Linux)
  - 20-core ARM CPU per node
  - ConnectX-7 200 Gb/s RoCEv2 interconnect
  - Tensor parallel across both nodes
  - Concurrency: one sequence
  - Architecture: aarch64
  - Kernel: Linux 6.17.0-1029-nvidia
  - NVIDIA driver: 580.173.02

  ### Software

   Component                      Version
  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
   vLLM                           0.28.1rc1.dev317+g1356635d8
  ─────────────────────────────  ──────────────────────────────────────────
   vLLM commit                    1356635d837c4ef002ec98c1a0296e7ff60be3c1
  ─────────────────────────────  ──────────────────────────────────────────
   PyTorch                        2.13.0+cu130
  ─────────────────────────────  ──────────────────────────────────────────
   CUDA                           13.0
  ─────────────────────────────  ──────────────────────────────────────────
   NCCL                           2.29.7
  ─────────────────────────────  ──────────────────────────────────────────
   FlashInfer package             0.6.18
  ─────────────────────────────  ──────────────────────────────────────────
   Baseline FlashInfer commit     26fabfe93ab7e866b1a3b581ca6ba2b984d49706
  ─────────────────────────────  ──────────────────────────────────────────
   Candidate FlashInfer commit    453aa7c7296e9ec711fd4c1f3aa6ee061a6b69dc

  ### Model

  - Model: deepseek-ai/DeepSeek-V4-Flash-Vision-Exp
  - Pinned revision: e46e16bf6035c6f317eb2ac7458eb0362926d402
  - Architecture: DeepseekV4ForCausalLM
  - Weight quantization: FP8, E4M3, dynamic activations
  - Weight block size: [128, 128]
  - KV cache: FP8
  - Maximum context: 1,048,576
  - Maximum sequences: 1
  - Vision limit: 8 images per prompt
  - Video disabled
  - MTP: 3 speculative tokens
  - MTP sampling: probabilistic
  - Adaptive verification: disabled

  ### Relevant vLLM launch arguments

  vllm serve deepseek-ai/DeepSeek-V4-Flash-Vision-Exp \
    --revision e46e16bf6035c6f317eb2ac7458eb0362926d402 \
    --served-model-name deepseek-v4-flash-vision-exp \
    --tensor-parallel-size 2 \
    --pipeline-parallel-size 1 \
    --enable-expert-parallel \
    --distributed-executor-backend mp \
    --nnodes 2 \
    --kv-cache-dtype fp8 \
    --block-size 256 \
    --max-model-len 1048576 \
    --max-num-seqs 1 \
    --max-num-batched-tokens 8192 \
    --long-prefill-token-threshold 1024 \
    --gpu-memory-utilization 0.80 \
    --max-cudagraph-capture-size 4 \
    --enable-prefix-caching \
    --enable-chunked-prefill \
    --async-scheduling \
    --speculative-config '{
      "method": "dspark",
      "model": "deepseek-ai/DeepSeek-V4-Flash-Vision-Exp",
      "revision": "e46e16bf6035c6f317eb2ac7458eb0362926d402",
      "num_speculative_tokens": 3,
      "draft_sample_method": "probabilistic",
      "enable_adaptive_verification": false
    }' \
    --tokenizer-mode deepseek_v4 \
    --limit-mm-per-prompt '{"image":8,"video":0}' \
    --no-enable-flashinfer-autotune

  ### Relevant environment

  ```bash
  TORCH_CUDA_ARCH_LIST=12.1a
  FLASHINFER_CUDA_ARCH_LIST=12.1a
  VLLM_USE_BREAKABLE_CUDAGRAPH=1
  VLLM_PREFIX_CACHE_RETENTION_INTERVAL=4096
  PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True

  NCCL_IB_DISABLE=0
  NCCL_NET=IB
  NCCL_IB_GID_INDEX=3
  NCCL_IB_ROCE_VERSION_NUM=2
  NCCL_CROSS_NIC=1
  NCCL_IB_MERGE_NICS=1
  NCCL_CUMEM_ENABLE=0
  NCCL_NVLS_ENABLE=0

The candidate was explicitly calibrated on both nodes for DeepSeek V4 with 64 local heads per TP rank and top-k 128. The generated calibration caches were isolated from the baseline caches.

JaviAFKzX:
Thanks for landing this. Reporting a follow-up from the same SM120 envelope, in case
it is useful to whoever touched this code most recently.

With a build carrying the runtime `topk_extra` dispatch (803c466), DSV4 Vision on
2x RTX PRO 6000 (SM120, TP=2, PHB) now starts cleanly and serves **images
correctly** — a synthetic PNG is described accurately and usage reports
`multimodal_tokens`. So the dispatch side looks right.

What still fails is a long **text-only** prompt (~13.6k tokens) on the same
vision checkpoint: generation starts and then both TP workers abort with
`CUDA error: an illegal memory access was encountered`.

The part that made me think it belongs here rather than in an integration issue:
it reproduces identically on two independent implementations of the dual-cache
path — the 803c466 runtime-dispatch build, and a locally patched 0.6.18 using the
`topk == 512` template arm from #4850. Same request, same abort, both workers.

Full environment, config and traces: <https://github.com/flashinfer-ai/flashinfer/issues/4973>

The text-only DeepSeek-V4-Flash-0731 checkpoint runs for days on the same box with
the same TP/KV/spec-decode settings without faults, so it looks specific to the
vision variant's wider prefill SWA index (128 + 384 = 512).

Happy to re-run with compute-sanitizer or `CUDA_LAUNCH_BLOCKING=1` if that helps.

lucamotz:
> Thanks for landing this. Reporting a follow-up from the same SM120 envelope, in case it is useful to whoever touched this code most recently.
> 
> With a build carrying the runtime `topk_extra` dispatch ([803c466](https://github.com/flashinfer-ai/flashinfer/commit/803c4664f4771ddc418f20a57f752469a237a825)), DSV4 Vision on 2x RTX PRO 6000 (SM120, TP=2, PHB) now starts cleanly and serves **images correctly** — a synthetic PNG is described accurately and usage reports `multimodal_tokens`. So the dispatch side looks right.
> 
> What still fails is a long **text-only** prompt (~13.6k tokens) on the same vision checkpoint: generation starts and then both TP workers abort with `CUDA error: an illegal memory access was encountered`.
> 
> The part that made me think it belongs here rather than in an integration issue: it reproduces identically on two independent implementations of the dual-cache path — the [803c466](https://github.com/flashinfer-ai/flashinfer/commit/803c4664f4771ddc418f20a57f752469a237a825) runtime-dispatch build, and a locally patched 0.6.18 using the `topk == 512` template arm from #4850. Same request, same abort, both workers.
> 
> Full environment, config and traces: #4973
> 
> The text-only DeepSeek-V4-Flash-0731 checkpoint runs for days on the same box with the same TP/KV/spec-decode settings without faults, so it looks specific to the vision variant's wider prefill SWA index (128 + 384 = 512).
> 
> Happy to re-run with compute-sanitizer or `CUDA_LAUNCH_BLOCKING=1` if that helps.

Hey! Have you tried https://github.com/vllm-project/vllm/pull/54110 + https://github.com/vllm-project/vllm/pull/54631 on the current main vLLM? I haven’t had the time yet to identify which of my changes would explicitly fix your problems, but I don’t see the same issues with long text prompts as you, but I did before

lucifer1004:
@JaviAFKzX It seems you are not using this PR (or FlashInfer main since it has been merged now)?

JaviAFKzX:
> @JaviAFKzX It seems you are not using this PR (or FlashInfer main since it has been merged now)?

You are right, and thanks for catching it — we were not on your PR.

The image we tested pins FlashInfer at `803c466`, which is a diverged branch, not
main. Checking now:

- `803c466` vs `main`: **diverged**, main is **236 commits ahead**
- `if (topk != 128) return false;` in `sparse_mla_sm120_prefill.cu`:
  **still present on `803c466`, gone on `main`**

So everything we reported was measured against a tree that still had the original
limitation. That also explains a result we could not account for: after backporting
vllm#54110 on top of that old FlashInfer, the illegal access disappeared but we got
`torch_call_dispatcher("aten::new_empty") API call failed` instead, and then on a
later run it worked but at 8.3 tok/s (vs ~190 tok/s for the text-only checkpoint on
the same box) and we could not reproduce the difference deterministically. We were
patching around something main had already solved differently.

We will rebuild against FlashInfer main and re-test, and report back with numbers.

One thing from our side that may still be worth a look regardless, since it is
independent of which FlashInfer we use: `top_k_per_row_decode` in `sampler.cu`
branches on `kSortingAlgorithmThreshold = 12288`. Below that it uses insertion sort
with no allocation; above it, it calls `torch::stable::empty` (four sites). Our
failing row was ~13.6k tokens, i.e. just above the threshold — which is why short
prompts were always fine. If vllm#54110's fallback is meant to cover long rows on
<128 KiB-smem devices, that allocating branch is the one it will land on.

Environment for reference: 2x RTX PRO 6000 Blackwell (SM120, 99 KiB opt-in smem per
block), TP=2, PHB, driver 595.84, torch 2.13.0, CUDA 13.3.

PR #4809, opened 2026-08-28
feat(kda): add fused speculative decode kernel

djmmoss:
## 📌 Description

Add `flashinfer.fused_kda_decode_packed`, an SM10x CuTe DSL kernel that fuses
Kimi-K3 short convolution, recurrent KDA, and gated RMS normalization for
packed speculative verification (`T >= 1`). The API updates convolution and
recurrent caches according to accepted-token metadata, supports ragged and
empty rows, accepts caller-owned output storage, and is safe to capture with
dynamic CUDA Graph metadata.

The change also adds public API documentation, FI trace coverage and a trace
fixture, focused correctness/validation tests, and a benchmark driver.

## 🔍 Related Issues

None.

### ✅ Pre-commit Checks

## 🧪 Tests

- [x] All focused tests are passing.

On one NVIDIA B300 (SM103):

```text
pytest -q tests/kda/test_fused_kda_decode.py \
  tests/trace/test_fi_trace_template_consistency.py \
  tests/trace/test_fi_trace.py
829 passed, 2 warnings in 21.47s
```

On eight NVIDIA B300s with synthetic Kimi-K3 weights, DSpark with six draft
tokens, 8,192-token inputs, and 1,024-token outputs, the fused vLLM integration
matched the baseline checksum and improved total output throughput by 2.63%,
2.54%, 3.29%, 1.76%, and 1.75% at concurrency 1, 4, 8, 16, and 32.

## Reviewer Notes

The implementation is intentionally limited to D=128, W=4, BF16 activation
and convolution state, FP32 recurrent state, bounded gates, supported Kimi-K3
head counts, and the SD convolution-cache layout. The cluster synchronization
around split-head state updates and the accepted-token cache commit semantics
are the highest-value review areas.

## Summary by CodeRabbit

* **New Features**
  * Added packed, multi-token fused KDA decoding for ragged sequences with recurrent state checkpoints.
  * Added fused convolution, recurrence, and normalization during packed decoding.
  * Added automatic backend selection for single- and multi-token KDA decoding.
  * Exposed new KDA decoding, MXFP8 quantization, and DCP utility APIs at the package’s top level.

* **Documentation**
  * Documented the packed and fused KDA decoding APIs.

* **Tests**
  * Added coverage for correctness, compatibility, dynamic metadata, determinism, CUDA graph capture, and trace generation.


flashinfer/kda_decode.py
```diff
@@ -476,6 +487,151 @@ def fused_kda_decode(
...
+        if x.shape[0] != legacy_state_indices.shape[0]:
+            packed_rows = torch.arange(
+                x.shape[0], dtype=query_start_loc.dtype, device=x.device
+            )
+            sequence_indices = torch.searchsorted(
+                query_start_loc[1:], packed_rows, right=True
+            )
+            legacy_state_indices = legacy_state_indices.index_select(
+                0, sequence_indices
+            )
```
kahyunnam, line 596: The guard at `:587` tests a *shape* (`x.shape[0]` vs the sequence count), but whether row *i* belongs to sequence *i* is a property of `query_start_loc`'s *values* — it holds only when `query_start_loc == arange(N+1)`. With `num_rows == N` and unbalanced sequences (e.g. live rows followed by trailing padding rows whose sequences are zero-length) the remap is skipped, and on a B200 I see conv and state slots `{3,4}` mutated where only `{1,2}` should be, plus non-zero values written into the padding output rows — which contradicts `:531`, "non-positive **or zero-length** rows are padding and do not mutate either cache". With `num_rows != N` the unclamped `searchsorted` at `:591-593` returns `N`, and the `index_select` at `:594` trips `indexSelectSmallIndex: Assertion 'srcIndex < srcSelectDimSize' failed`, which poisons the CUDA context rather than raising a catchable error. The sharpest case is CUDA-graph capture: this branch is evaluated once on the host, so capturing with an all-active `query_start_loc` compiles the remap *out* of the graph, and replaying with a zero-length sequence returns wrong live-row output (0.41 against the suite's `atol=2e-2`) with a corrupted state pool — the `T>1` path replays correctly because the kernel reads `query_start_loc` on device.

Suggest dropping the guard and always remapping, which is a no-op when the mapping really is the identity, with one null slot appended so out-of-range rows land on the backend's existing null path instead of faulting:

```python
legacy_state_indices = torch.cat(
    [state_indices[:, 0], state_indices.new_zeros(1)]
).index_select(0, torch.searchsorted(query_start_loc[1:], torch.arange(
    x.shape[0], dtype=query_start_loc.dtype, device=x.device), right=True))
```

Both shapes below are ones the `T>1` path is explicitly tested for — `test_fused_kda_decode_packed_ragged_rows_and_padding` and `test_fused_kda_decode_packed_cuda_graph_dynamic_metadata` — so this is an inconsistency between the two halves of one facade rather than an arguable reading of the contract.

Output on a B200 (SM100, cc 10.0):

```
case A: trailing padding rows, num_rows == N == 4, query_start_loc=[0,1,2,2,2]
  conv slots mutated : [1, 2, 3, 4]   expected [1, 2]
  state slots mutated: [1, 2, 3, 4]   expected [1, 2]
  padding rows 2:4 written: True  max|value|=0.8164

case B: captured with query_start_loc=[0,1,2,3], replayed with [0,1,1,2]
  live-row output correct : False
  max|replay - truth|     : 0.4072   (test tolerance atol=2e-2)
  per-row max abs diff    : [0.0, 0.4072]
  state pool correct      : False
```

In case B, row 0 is correct and row 1 is wrong, which is what pins the diagnosis to the mapping rather than to the kernel.

<details>
<summary>Repro script (save at the repo root, <code>PYTHONPATH=$PWD python repro_4809_t1.py</code>)</summary>

```python
import sys
import torch

sys.path.insert(0, "tests")
from kda.test_fused_kda_decode import _clone_strided, _make_inputs

import flashinfer
from flashinfer.kda_decode import fused_kda_decode_packed

DEV, H = torch.device("cuda"), 12


def build(num_rows, qsl, slots):
    n = len(qsl) - 1
    inp = _make_inputs(H, max(num_rows, n), state_dtype=torch.float32, conv_history=3)
    for key, dim in (("x", 0), ("raw_gate", 1), ("raw_beta", 1), ("output_gate", 0)):
        inp[key] = inp[key].narrow(dim, 0, num_rows)
    inp["state_indices"] = torch.tensor([[s] for s in slots], dtype=torch.int32, device=DEV)
    inp["query_start_loc"] = torch.tensor(qsl, dtype=torch.int32, device=DEV)
    inp["num_accepted_tokens"] = torch.ones(n, dtype=torch.int32, device=DEV)
    return inp


def mutated(after, before):
    d = (after.float() - before.float()).abs()
    return d.flatten(1).amax(1).nonzero().flatten().tolist()


# Case A: 2 live rows (sequences 0,1) + 2 trailing padding rows (sequences 2,3
# zero-length). num_rows == N == 4, so the guard at :587 is False and the remap
# never runs. Per :531, slots 3 and 4 must be untouched.
inp = build(num_rows=4, qsl=[0, 1, 2, 2, 2], slots=[1, 2, 3, 4])
conv0, state0 = _clone_strided(inp["conv_state"]), _clone_strided(inp["state"])
conv, state = _clone_strided(inp["conv_state"]), _clone_strided(inp["state"])
out = torch.full((1, 4, H, 128), torch.nan, dtype=torch.bfloat16, device=DEV)
fused_kda_decode_packed(**{**inp, "conv_state": conv, "state": state, "output": out})
torch.cuda.synchronize()

print("case A: trailing padding rows, num_rows == N == 4, query_start_loc=[0,1,2,2,2]")
print(f"  conv slots mutated : {mutated(conv, conv0)}   expected [1, 2]")
print(f"  state slots mutated: {mutated(state, state0)}   expected [1, 2]")
print(f"  padding rows 2:4 written: {not bool(torch.isnan(out[0, 2:]).all())}"
      f"  max|value|={float(out[0, 2:].float().nan_to_num().abs().max()):.4f}")

# Case B: capture all-active (num_rows == N, guard False, remap compiled OUT),
# then replay with a zero-length sequence.
inp = build(num_rows=3, qsl=[0, 1, 2, 3], slots=[1, 2, 3])
conv, state = _clone_strided(inp["conv_state"]), _clone_strided(inp["state"])
out = torch.zeros((1, 3, H, 128), dtype=torch.bfloat16, device=DEV)
call = {**inp, "conv_state": conv, "state": state, "output": out}

s = torch.cuda.Stream()
s.wait_stream(torch.cuda.current_stream())
with torch.cuda.stream(s):
    for _ in range(3):
        fused_kda_decode_packed(**call)
torch.cuda.current_stream().wait_stream(s)

g = torch.cuda.CUDAGraph()
conv.copy_(inp["conv_state"]); state.copy_(inp["state"])
with torch.cuda.graph(g):
    fused_kda_decode_packed(**call)

conv.copy_(inp["conv_state"]); state.copy_(inp["state"])
call["query_start_loc"].copy_(torch.tensor([0, 1, 1, 2], dtype=torch.int32, device=DEV))
g.replay()
torch.cuda.synchronize()

# Truth: rows 0 and 1 belong to sequences 0 and 2, i.e. slots 1 and 3.
truth_state = _clone_strided(inp["state"])
truth = flashinfer.fused_kda_decode(
    x=inp["x"].narrow(0, 0, 2),
    weight=inp["weight"],
    conv_state=_clone_strided(inp["conv_state"]),
    raw_gate=inp["raw_gate"].narrow(1, 0, 2),
    raw_beta=inp["raw_beta"].narrow(1, 0, 2),
    A_log=inp["A_log"],
    dt_bias=inp["dt_bias"],
    state_indices=torch.tensor([1, 3], dtype=torch.int32, device=DEV),
    state=truth_state,
    output_gate=inp["output_gate"].narrow(0, 0, 2),
    norm_weight=inp["norm_weight"],
)
torch.cuda.synchronize()

d = (out[0, :2].float() - truth.reshape(2, H, 128).float()).abs()
print("\ncase B: captured with query_start_loc=[0,1,2,3], replayed with [0,1,1,2]")
print(f"  live-row output correct : {bool(d.max() < 2e-2)}")
print(f"  max|replay - truth|     : {float(d.max()):.4f}   (test tolerance atol=2e-2)")
print(f"  per-row max abs diff    : {[round(float(v), 4) for v in d.flatten(1).amax(1)]}")
print(f"  state pool correct      : {bool((state - truth_state).abs().max() < 1e-3)}")
```

</details>

flashinfer/kda_decode.py
```diff
@@ -476,6 +487,151 @@ def fused_kda_decode(
...
+    state_indices: torch.Tensor,
+    query_start_loc: torch.Tensor,
+    num_accepted_tokens: torch.Tensor,
+    state: torch.Tensor,
+    output_gate: torch.Tensor,
+    norm_weight: torch.Tensor,
+    lower_bound: Optional[float] = -5.0,
+    norm_eps: float = 1e-5,
+    output: Optional[torch.Tensor] = None,
+) -> torch.Tensor:
```
kahyunnam, line 508: API nits: could be fixed before merge here, otherwise could become expensive to clean up after release (we can't change anything about this API after unless its bwd compatible or go through the long process for deprecation).

**(a) Two parameters are inserted mid-signature, and nothing is keyword-only.**
`inspect.signature` on both ops gives 16 parameters each, with **14 shared —
identical names, defaults and relative order** (`x, weight, conv_state, raw_gate,
raw_beta, A_log, dt_bias, state_indices, state, output_gate, norm_weight,
lower_bound, norm_eps, output`). But `query_start_loc` and `num_accepted_tokens`
land at positions 8 and 9, so everything from `state` onward shifts by two
**[GPU]**:

| pos | `fused_kda_decode` | `fused_kda_decode_packed` |
|---|---|---|
| 7 | `state_indices` | `state_indices` |
| 8 | `state` | **`query_start_loc`** |
| 9 | `output_gate` | **`num_accepted_tokens`** |
| 10 | `norm_weight` | `state` |
| 11 | `lower_bound` | `output_gate` |
| 12 | `norm_eps` | `norm_weight` |
| 13 | `output` | `lower_bound` |
| 14 | `backend` *(keyword-only)* | `norm_eps` |
| 15 | `state_indices_mode` *(keyword-only)* | `output` |

All 16 parameters on the new op are `POSITIONAL_OR_KEYWORD`, whereas the sibling
makes its two newest (`backend`, `state_indices_mode`) `KEYWORD_ONLY`. So a caller
moving a positional call between two functions that otherwise share fourteen
parameters in the same order passes `state` where `query_start_loc` belongs. In
fairness, the validators would catch most tensor mix-ups with a confusing shape
error rather than corrupting anything — the hazard is the confusion, plus the fact
that the `lower_bound` / `norm_eps` / `output` tail is less certainly caught. The
reason to fix it now is that it is free today and a breaking change once 0.7 cuts,
which is #4936's own central thesis applied to this PR.

**Ask:** make `query_start_loc` and `num_accepted_tokens` keyword-only, following
the precedent the sibling already set in this file.

**(b) The name inverts the modifier order of the op 190 lines above it.**
`packed_kda_decode` already exists in this module; the new op spells the same
modifier the other way round. Cheapest item here, and the only one that is purely
cosmetic once released.

**(c) The packed facade has no `backend=`, unlike the sibling.**

`fused_kda_decode` exposes `backend: Literal["auto", "cute-dsl", "cake"]` and
`state_indices_mode` as keyword-only parameters
(`flashinfer/kda_decode.py:386-389`) and forwards them (`:485-486`).
`fused_kda_decode_packed` has neither. That is two sibling facades in one module,
sharing fourteen parameters, one with a backend selector and one without —
structurally the same divergence as #4936's problem 6 between `kda.py` and
`kda_decode.py`, on a new axis.

The concrete consequence lands on the `T=1` branch, which calls
`_run_fused_kda_decode` at `:597-612` without either argument and therefore always
takes the default `"cute-dsl"` route. A `T=1` call that would have selected the
exported Cake kernel on SM100a/SM103a through `fused_kda_decode` cannot reach it
through this entrypoint, and nothing in the docstring says so. The claim at `:518`
that "T=1 dispatches to :func:`fused_kda_decode`" is also not quite right — it
dispatches to that op's *backend*, with a narrower argument set.

- **Ask:** say whether `backend=` is planned for the packed path or  forward both arguments on the `T=1` branch

kahyunnam (commented):
LGTM overall, but please take a look at one bug concern raised by claude review below + one general API design concern.

flashinfer/kda_decode.py
djmmoss: fixed in 0e280ebca: T=1 now resolves the packed row to its sequence inside the CuTe kernel from query_start_loc on every replay, and rows past query_start_loc[-1] go through the null slot. I added the trailing-padding and dynamic CUDA Graph cases from the repro; the fused/API/trace suite passes 1046/1046 on B300.

flashinfer/kda_decode.py
djmmoss: I renamed it to packed_fused_kda_decode and moved query_start_loc and num_accepted_tokens behind *, so the shared positional arguments stay aligned with fused_kda_decode. I kept this facade CuTe-only rather than exposing backend=: T>1 exists only in CuTe and Cake doesn't consume packed metadata; the docstring now says that explicitly. I'd add backend selection if Cake gains packed semantics.

kahyunnam:
@flashinfer-bot run tests/kda

kahyunnam (approved):

kahyunnam:
@flashinfer-bot run tests/kda

djmmoss:
@kahyunnam any other blockers?

djmmoss:
@flashinfer-bot run tests/kda

djmmoss:
@flashinfer-bot run tests/kda

PR #4811, opened 2026-08-28
feat(topk): self-sampling GVR V2 backend (SM100/103/107), oracle-tracking auto, and the V1 threshold repair

dhiraj113:
# feat(topk): self-sampling GVR V2 backend, oracle-tracking `auto`, and the V1 threshold repair

Three commits, one arc: port TRT-LLM's self-sampling GVR V2 top-K decode as a new `top_k_varlen` backend, make `backend="auto"` track the measured per-config winner, and fix the correctness bug in the existing V1 `gvr` backend that the new benchmark sweep uncovered.

## 1. `gvr_2` — self-sampling GVR V2 port (NVIDIA/TensorRT-LLM#17821, commit `ed94d4cfbf`)

**Algorithm** — V1 GVR guesses its selection threshold from the previous step's `pre_idx` hint and pays full-row refinement rescans when the hint is stale. V2 instead derives a *bracketed ladder* of candidate thresholds from an in-kernel sample of the row itself (Floyd–Rivest-style) and resolves exact, tie-interchangeable top-K in a **single streaming pass**; exactness is guaranteed by count-crossing invariants, never by the estimate, and the hint survives only as a degenerate-case anchor. Four kernel families (streaming `main` with multi-CTA SPLIT, register-resident `reg`/`regimg`, clustered `clus`/`reg_clus`) are chosen by a pure host dispatch `route(b, n, npad, k)`; per-request lengths are read **on device**, so one launch serves the whole ragged batch — no prepare kernel, no LJF sort, no host reads (sync-free and CUDA-graph safe; the length envelope comes from the logits row width).

**Port notes**
- `flashinfer/topk_varlen/kernels/gvr2_topk_decode.py` (device) and `gvr2_topk_host.py` (dispatch/workspace/entry) are near-verbatim upstream drops, excluded from ruff/mypy like the repo's other verbatim kernel ports so future syncs stay mechanical. Local changes: module rename, provenance notes, a `_persist()` hook routing the four `get_compiled*` builders through the persistent CuTe-DSL kernel cache (the upstream compile closures already use `--enable-tvm-ffi` + fully symbolic shapes, matching the cache's TVM-FFI reload convention), and a fixed missing kv-arg in the `regclus_topk` debug entry.
- Backend contract: fp32 logits only (bf16/fp16 are an upstream follow-up), `top_k ∈ {512, 1024, 2048}`, `compress_ratio ∈ {1, 4}`, `pre_idx` required with width == `top_k`, datacenter Blackwell (sm_100/103). The ~21 MB per-device workspace slab is zero-initialized once and self-restoring; multi-stream callers can pass `workspace={"gvr2_workspace": ...}`.
- Upstream caveat found by this PR's adversarial tests and reproduced **bit-identically by the TRT-LLM implementation**: literal `+inf` logits may not be selected (at least in the clustered-register family); all finite values — including 3.1e38, above the kernel's 3e38 pad sentinel — and `-inf` are tie-aware exact. Documented in the API docs + a non-strict `xfail` (`test_gvr2_plus_inf_upstream_caveat`) that flips visible if a future upstream sync fixes it. Same class as upstream's implementation-specific NaN ordering.

**Performance** (B200, fp32, CUDA-graph timing, tie-aware output validation before every timing; 213 configs across uniform/mixed/short lengths, K ∈ {512, 1024, 2048}, B ∈ [1, 256], N ∈ [1K, 128K], cr ∈ {1, 4}, next_n ∈ {1, 2}):

| vs backend | geomean | gvr_2 faster | range |
|---|---|---|---|
| `gvr` (existing, LB) | **2.58×** | 210/211 | 0.47–6.23× |
| `gvr` (non-LB) | 3.80× | 210/211 | 0.76–10.94× |
| `radix` (CuTe DSL) | 2.13× | 211/213 | 0.30–10.05× |
| `radix_cutlass` | 3.29× | 211/213 | 0.56–5.92× |
| TRT-LLM upstream twin (port parity) | **1.00×** | — | 0.95–1.05× |

The only losing regime is batches whose rows are mostly barely longer than K (upstream's own documented short-row trade-off).

## 2. Shape/dtype-aware `auto` (oracle-tracking)

The previous static gvr-first `auto` never reached `gvr_2`, always picked `gvr` for hinted bf16/fp16 where `radix` wins by up to 2.8×, and never reached `radix_cutlass` in its fp32 big corner (up to 2.8×). The new ranking uses only capture-stable host facts (dtype, N, B — never `seq_lens` contents, which would cost a D2H sync), derived from a 500+-cell sweep:

1. hinted fp32 → **`gvr_2`**;
2. `gvr` outranks `radix` only when `B·N ≥ 2²²` (fp32) / `2²³` (bf16/fp16);
3. `radix_cutlass` outranks `radix` only in the fp32 corner `N ≥ 65536` and `B·N ≥ 2²³`; in half precision `radix` always leads.

**Validated on 100 off-grid configs** (shapes never used to fit the thresholds), regret = t(auto's pick) / t(fastest backend):

| policy | geomean | max | oracle hits |
|---|---|---|---|
| new | **1.004×** | 1.16× | 97/100 |
| old | 1.317× (hinted 1.694×) | 3.05× | 55/100 |

Documented static blind spot: mostly-short rows inside a wide N favor `radix` but are indistinguishable without reading `seq_lens`; such callers should pass `backend="radix"` explicitly. `benchmarks/bench_topk_varlen_gvr2.py` (new comparison benchmark, with an upstream-twin mode) gained `--dtype`; the bf16/fp16 sweeps drove rules 2–3.

## 3. V1 `gvr` threshold-search repair (folded from #4813)

This PR's sweep found the existing `gvr` backend shipping silently wrong top-K when its Phase-2 threshold search terminates without converging: identity indices `row[0:K]` on degenerate hint brackets, or underfilled rows (stale/−1 output slots) on hostile hints, tie plateaus wider than the candidate buffer, and `N_eff = K+1` batches. Fixed by porting the correctness-relevant subset of three upstream commits FlashInfer's V1 snapshot (~Jul 22) predates:

- NVIDIA/TensorRT-LLM#16457 + #16877: the tie-plateau layer — adjacent-float bracket terminal (`done=3`) in both `phase2_secant_search` copies, `s_iscalars (6,)→(8,)`, plateau flag/ticket, gated Branch-C pad, and a post-Phase-4 fill completing the row from the bitwise-equal tie class;
- NVIDIA/TensorRT-LLM#18094 (the upstream tip for this kernel, verified via live GitHub): the two-sided repair — Phase 3's overflow-only retry becomes an anchored bisection on the signed fp32 order-key image (provable collapse in ≤32 steps), handling undershoot, restoring to `val_lo` when a collapse ends under K, and handing collapsed plateaus to the `done=3` machinery; the degenerate-hint identity emit becomes a synthetic-bracket fall-through, so correctness no longer depends on the hint at all.

Both LB paths are covered automatically (`GvrTopKLBKernel` reuses `GvrTopKKernel.run_one_row`). Before/after: the fix is ~4% *faster* geomean in the converging common case (the repair shrinks oversized candidate sets) and turns the previously-wrong regimes correct at 1.1–7× cost (their old timings were the price of not computing the answer). Full A/B table and root-cause discussion preserved in the #4813 thread. Ported with a 3-lens adversarial review against the upstream reference (port fidelity, DSL/barrier execution, FlashInfer-divergence interactions) — all clean.

## Testing

- `tests/topk_varlen/test_topk_varlen_gvr2.py` (79 tests): tie-aware exactness with poisoned pads, short-row identity+pad contract, degenerate hints, varlen/MTP/compress-ratio grids, zero-kv-slot rows, per-family admission parity, CUDA-graph capture/replay with in-place growing `seq_lens`, warmup-then-capture, non-contiguous arena views, workspace override, adversarial tie/huge-value/denormal patterns, a >10k-point `route_split == route` dispatch fuzz, and the `+inf` caveat xfail.
- `tests/topk_varlen/test_gvr_threshold_repair.py` (31 tests): upstream #18094's regression patterns on the FlashInfer API (hostile/degenerate hints × K × LB modes, ReLU-sparse plateaus, MTP hostile hints) plus the FlashInfer-found `N_eff = K+1` case — all fail pre-fix.
- Cross-backend value-multiset consistency and the shape-aware heuristic-priority unit test (meta tensors, runs off-GPU) extended in `test_topk_varlen.py`.
- Results: see section 4 for the post-rebase sweep on SM80/89/90/100 (B200 and B100)/107 (Rubin)/120.

## 4. Rubin (SM107) enablement and rebase onto #4621

Rebased onto current `main`, which now includes #4621 (Rubin support for `top_k_varlen` plus the `radix_filter` backend). Conflicts were confined to `topk_varlen.py` (docstrings, the backend `Literal`/registration, the `auto` heuristic body, and the `next_n` validation), `pyproject.toml`, and `.pre-commit-config.yaml` (both sides added vendored-kernel exclusions; union kept). Resolutions of note:

- The `next_n`/`seq_lens` grouped-row validation lands as #4621's `ValueError` raises (they hold under `python -O`); this branch's equivalent asserts were dropped and its up-front-validation test now expects `ValueError`.
- `auto` keeps this branch's shape-aware ranking; `radix_filter` stays explicit-only, with #4621's rationale comment retained.

`gvr_2` on Rubin (commit `feat(topk): enable gvr_2 on Rubin (SM107)`):

- Admission comes through the shared `_GVR_CCS = [100, 103, 107]` list from #4621; the gvr_2 checker now uses #4621's per-device `_cute_dsl_ready()` probe instead of the process-wide DSL flag, so a DSL predating the device degrades `auto` cleanly.
- No kernel changes: the device module uses only family-portable ops (no `tcgen05` / block-scaled MMA), so the same source compiles for `sm_107a`; the compile target follows the current device and the persistent CuTe-DSL cache is arch-namespaced.
- The host router's `148` (B200 SM count, mirrored from the upstream CUDA dispatch) is documented as an occupancy heuristic only; on Rubin's 208 SMs the same constants are conservative, never incorrect. Per-arch retuning is a perf follow-up.

Verification after the rebase (full `tests/topk_varlen/`, which now includes #4621's `test_radix_filter.py`):

| Arch | GPU | `tests/topk_varlen/` at `9f518813` (identical to the swept tree except one unused loop index in a test) |
|---|---|---|
| SM80 | A100-PCIE-40GB | 54 passed, 203 skipped |
| SM89 | L40S | 54 passed, 203 skipped |
| SM90 | H100 NVL | 54 passed, 203 skipped |
| SM100 | B200 | 252 passed, 3 skipped, 2 xfailed |
| SM100 | B100 | 252 passed, 3 skipped, 2 xfailed |
| **SM107** | **Rubin GR100** | **253 passed, 2 skipped, 2 xfailed** (7 min 19 s cold, including the first `sm_107a` compiles of every gvr_2 family) |
| SM120 | RTX 5080 | 54 passed, 203 skipped |

The skips on SM100/107 are all in `test_radix_filter.py` (two need a second visible GPU; one wants an async-TMA default that is active on SM107 but not SM100, hence Rubin's extra pass). The two xfails are the known upstream caveats (radix_filter's padded-merge tie, gvr_2's `+inf`). Every gvr_2 test ran on SM100 and SM107.

Basic perf after the rebase (`benchmarks/bench_topk_varlen_gvr2.py`, fp32, K = 1024, B ∈ {1, 16, 64, 256}, N ∈ {8K, 32K, 128K}, uniform and mixed lengths, CUDA-graph timing, tie-aware validation before every timing):

| GPU | vs `gvr` | vs `radix` | vs `radix_cutlass` | gvr_2 wins |
|---|---|---|---|---|
| B100 (SM100) | 2.40× | 2.49× | 3.40× | 24/24 vs each |
| Rubin GR100 (SM107) | 2.31× | 2.55× | 3.37× | 24/24 vs each |

Consistent with the B200 sweep in section 1, and Rubin lands there with the router's B200 occupancy constants untouched.

## Follow-ups (out of scope)

- Consider an `N_eff`-aware escape for gvr_2's short-row regime if it is ever promoted for length-skewed workloads.
- bf16/fp16 gvr_2 tracks the upstream roadmap; once ported, the V1 `gvr` backend's last winning corner (hinted bf16/fp16 at large B·N) disappears and it can be deprecated.
- Report the `+inf` selection caveat upstream (reproducer in the xfail test).

AI-assisted (Claude Code).

## Summary by CodeRabbit

- **New Features**
  - Added the GVR V2 backend for variable-length top-k operations on supported Blackwell hardware.
  - Added support for variable sequence lengths, returned values, workspace controls, warmup, and CUDA Graph workflows.
  - Added automatic backend selection based on input shape and data type.
  - Added a configurable benchmark for comparing top-k implementations.

- **Bug Fixes**
  - Improved handling of ties, plateaus, sparse results, short rows, and challenging threshold-search cases.
  - Added validation and clearer handling for unsupported configurations.

- **Tests**
  - Expanded correctness, consistency, validation, and CUDA Graph coverage.

## 5. Review follow-up (`ea069804`)

Addresses the open review threads:

* **Row-length clamp on arena views.** `gvr_main`'s varlen prologue clamped each row's kv-derived length to the row stride; the other three families clamp to the envelope. It now clamps to the envelope too (carried in the previously dead `n` launch slot), so an oversized `seq_lens` value on a wider-stride view can never classify arena columns in `[width, stride)`. Contiguous inputs are unchanged. New test `test_gvr2_arena_view_oversized_seq_lens` covers all families; it fails on the previous kernel for both `gvr_main` parametrisations.
* **Heterogeneous multi-GPU processes.** `run` / `run_ws` / `run_varlen` re-enter under `torch.cuda.device(logits.device)` when the logits are not on the current device, so the launcher cache key, the DSL compile target and the launch agree by construction. Every gvr_2 `cute.compile` also passes an explicit `--gpu-arch` for the current device: the DSL otherwise detects its target once per process from device 0, so a kernel compiled while another device is current was built for device 0's architecture and failed to load (`cudaErrorNoKernelImageForDevice`). New test `test_gvr2_logits_on_non_current_device` (exercised with device 0 = L40S, device 1 = B100; failed before both fixes).
* Benchmark: `mixed`/`short` rows drawn from the documented domain, GVR-family backends report `n/a` outside `top_k ∈ {512, 1024, 2048}`, summary no longer requires `fi_gvr2`; `benchmarks/routines/topk_varlen.py` drops `gvr_2` when `max_seq_len % 4 != 0`; raw-string regex patterns in tests.
* The `+inf` caveat stays an `xfail` with the upstream tracking issue; see the thread for the rationale.

**Public API note.** The "potential breaking change" advisory is the `backend` `Literal` gaining the `"gvr_2"` member. That is additive: every previously valid call, including positional ones, is unchanged.

## 6. Known kernel defect: DKG issue #58 (`+inf` / NaN dropped) and the TRT-LLM #18501 port (`8e76f056`)

While validating this port I filed DKG issue #58 (internal GitLab, `dlarch-fastkernels/dynamic-kernel-generator`) against the upstream self-sampling GVR V2 kernel: a row containing a literal `+inf` returns a top-k without it, and NaN is not ranked on top the way `torch.topk` does. The result does not depend on hint quality (oracle and `-1` hints fail identically). The `+inf` case is inside the kernel's documented "finite + inf exact" contract; the NaN case is a torch-parity gap (upstream documents NaN ordering as implementation-specific).

The kernel owner pointed at TensorRT-LLM PR #18501 as the latest fix. **Reproduced on this branch and ported here** (register family `GvrTopkRegKernel.kern`, verbatim apart from our module rename): count-crossing enforcement when the hint-derived bracket makes the histogram total fall short of k, a radix-descent escape replacing the 32-step key-space bisection, and the `-inf` fill-lane bound. Before the port, the two upstream regressions were severe on this branch: a cold-start hint buffer (all zeros plus an argmax anchor) left 130,304 of 131,072 output slots unwritten across 256 rows, and an in-window `-inf` in a row's tail column left every row entirely unwritten. After the port both are exact with every slot written, on B100, B200 and Rubin (`test_gvr2_high_anchor_hint_completeness`, `test_gvr2_neginf_tail_completeness`).

**`+inf` closed (`48572ef2` + `274fad3e`):** the kernel owner's follow-up TensorRT-LLM PR #18625 is ported in full. Its first commit fixes `GvrTopkRegKernel.kern` (an infinite bracket width fails the collapse guard and the row takes the key-space escape); its second commit, written after the repro posted on DKG issue #58, fixes `GvrRegClusKernel.kern` the same way (bounded guard, and an infinite-width bracket forces the whole-row `degen` fallback instead of the collapsed histogram, which used to turn the `+inf` into NaN in the trash bin and return the true top-K with the `+inf` replaced by the (K+1)-th value). Verified on B200, B100 and Rubin against upstream at `9f920331` and through `top_k_varlen(backend="gvr_2")`: `+inf` inside or outside the bracket sample, K in {512, 1024, 2048}, more `+inf` than K, the 4 x 32K and 4 x 64K `reg_clus` shapes, with oracle, `-1` and random hints; `main` and `clus` were already exact. `test_gvr2_posinf_completeness` (register family) and `test_gvr2_posinf_reg_clus` (clustered-register family, replaces the former strict `xfail`) pin it; the API docstring caveat is gone. NaN ordering stays implementation-specific upstream, although NaN now ranks on top in every shape tested. Perf of the reg_clus hunks (B100, graph replay, 5 alternating passes, deterministic to the nanosecond): 4 x 32K +2.5%, 8 x 32K +2.8% (about 0.15 us each), 16 x 32K and the register family unchanged.

## 7. `auto` heuristic refinement (`993f6c9e`): radix_filter admission and the half-precision gvr rule

Measured `auto` against an oracle (fastest explicit backend per cell, CUDA-graph replay, tie-aware validation) over 4 batch sizes x 5 lengths (8K..1M) x uniform/mixed/short lengths x hinted/hint-free, fp32 and bf16, K=1024, on B100/B200 (SM100) and Rubin (SM107). Efficiency = best-of-all time / auto time (1.00 = auto matched the oracle). Findings and the resulting rules (one rule set for every arch; the SM100 crossovers are conservative on SM107, where radix_filter's margins are larger):

* Hinted fp32 (the DSA decode path) was already at 0.95-1.00: gvr_2 first stays. The remaining sub-0.95 cells are rows whose valid length is within 1-4 of K, a register-family fallback path (perf follow-up).
* Hint-free fp32 sat at 0.83 (SM100) / 0.72 (SM107) and bf16 at 0.86 / 0.82 purely because `radix_filter` was excluded from `auto` while being the fastest hint-free backend in most cells from 32K columns up. It is now admitted where it measured fastest: fp32 for N >= 32K (except the single-row case at N >= 512K) and at every N once B >= 256; half precision for 32K <= N <= 128K with B <= 16, N >= 128K with B >= 64, and N <= 8K with B >= 256. Its checker no longer rejects `pre_idx`: the hint is optional steering for the GVR family, so a hinted caller may use any hint-free backend, explicitly or via `auto`, exactly as radix and radix_cutlass already allowed; the hint is accepted and ignored.
* One genuine ranking error: for half precision the old `B*N >= 2^23` rule chose gvr (V1) at B <= 64 for N >= 512K, where it loses 1.5-7x to radix (16 rows x 2M: 1518 us vs 210 us). gvr is now chosen for half precision only when B >= 256 and 32K <= N <= 512K, where it measured fastest. The fp32 gvr rule is unchanged (it only matters when gvr_2 is unsuitable; radix_filter ranks ahead of it there).

Re-measured with the refined heuristic (cells >= 0.95 in parentheses; worst remaining cell listed):

| dtype, K | hint | GPU | before | after | worst cell after |
|---|---|---|---|---|---|
| fp32, K=1024 | yes | SM100 (B100) | 0.95 (46/48) | **0.99** (57/60) | short B=64 N=8K |
| fp32, K=1024 | yes | SM107 (Rubin) | 0.95 (45/48) | not re-measured (GPU shared with another job at the time; same rule set, radix_filter's margins are larger on SM107 so the SM100 crossovers are conservative there) | |
| fp32, K=1024 | no | SM100 (B100) | 0.83 (19/48) | **0.99** (58/60) | short B=256 N=128K |
| bf16, K=1024 | yes | SM100 (B100) | 0.92 (29/48) | **0.97** (53/60) | short B=64 N=1024K |
| bf16, K=1024 | no | SM100 (B100) | 0.86 (26/48) | **1.00** (59/60) | mixed B=16 N=128K |

`test_backend_heuristic_priority` pins the new boundaries off-GPU; the public docstring's `backend` section now documents `radix_filter` and the refined `auto` order.

## 8. SGLang comparison -> two dispatch rungs and hint-free `gvr_2` (`117415eb`)

An SGLang report ("gvr_v2 slower than `topk_transform_512_v2` at 32K with c4") was reproduced with their input recipe (fp32 randn logits, compressed widths padded to 64, uniform and variable lengths, perfect previous-step hint, identity page table, CUDA-graph replay medians, K = 512). Every comparison below ran in ONE process on ONE isolated computelab node (B200: 1000 W part, driver 610.57; B300: SM103 600 W part, driver 595.58), three columns per cell: sglang `topk_v2`, `gvr_2` with the upstream dispatch, `gvr_2` with this commit. Both `gvr_2` variants were value-multiset exact in every cell.

**Finding.** The claim held in one band only: compressed width 6144-8192, where `gvr_2` was 3-7% slower at bs <= 128 and 13-21% slower at bs 256; it was already 8-45% faster at every other width from 4096 to 131072. The cause is the dispatch, not the kernels: for `1024 < n4 <= 4096` the upstream `route()` runs every `wide` row on the VPT=4 register kernel (two empty float4 slots per thread up to n = 8192) and sends b > 148 to the streaming slab.

**Change (FlashInfer-local rungs in `route()`, reported upstream as DKG issue #61):** `n4 <= 2048` and `b <= 148` -> `_reg(1024, 2, 1, 2*NB)`; `n4 <= 2048` and `148 < b <= 296` (one wave at MINB=2) -> `_reg(512, 4, 2, NB)` with `NBSEL` following so `IMGOFF == NBH`; b > 296 keeps the slab (at b = 512 the two trade places by shape). Same-node A/B on B100, B200, B300 and Rubin: 18-20% faster in the wide band for K in {512, 1024, 2048}, 1.3-1.7x at b = 160-256.

sglang `topk_v2` time / `gvr_2` time (> 1 = `gvr_2` faster), uniform lengths, bs 1 / 16 / 64 / 128 / 256:

| compressed width | B200 upstream dispatch | B200 this commit | B300 upstream dispatch | B300 this commit |
|---|---|---|---|---|
| 4096 | 1.08 / 1.08 / 1.08 / 1.09 / 1.24 | 1.01 / 1.08 / 1.08 / 1.09 / 1.24 | 1.08 / 1.06 / 1.04 / 1.06 / 1.24 | 1.01 / 1.06 / 1.04 / 1.06 / 1.24 |
| 6144 | 0.95 / 0.95 / 0.95 / 0.95 / 0.80 | 1.14 / 1.14 / 1.14 / 1.14 / 1.30 | 0.92 / 0.93 / 0.93 / 0.93 / 0.79 | 1.12 / 1.12 / 1.12 / 1.12 / 1.31 |
| 8192 | 0.96 / 0.97 / 0.99 / 0.98 / 0.84 | 1.17 / 1.19 / 1.20 / 1.19 / 1.39 | 0.93 / 0.97 / 0.97 / 0.96 / 0.87 | 1.16 / 1.21 / 1.20 / 1.17 / 1.38 |
| 12288 | 1.10 / 1.15 / 1.15 / 1.16 / 1.01 | 1.16 / 1.15 / 1.15 / 1.17 / 1.01 | 1.11 / 1.11 / 1.13 / 1.15 / 0.96 | 1.17 / 1.13 / 1.14 / 1.15 / 0.96 |
| 16384 | 1.23 / 1.26 / 1.28 / 1.28 / 1.17 | 1.23 / 1.26 / 1.28 / 1.28 / 1.17 | 1.25 / 1.25 / 1.32 / 1.63 / 1.19 | 1.25 / 1.24 / 1.32 / 1.61 / 1.19 |
| 32768 | 1.59 / 1.70 / 1.45 / 1.27 / 1.41 | 1.65 / 1.70 / 1.45 / 1.27 / 1.41 | 1.67 / 1.77 / 1.51 / 1.29 / 1.47 | 1.61 / 1.75 / 1.49 / 1.29 / 1.47 |
| 131072 | 1.47 / 1.44 / 2.33 / 2.00 / 1.80 | 1.47 / 1.44 / 2.33 / 2.00 / 1.80 | 1.52 / 1.13 / 2.39 / 2.04 / 2.08 | 1.52 / 1.13 / 2.38 / 2.04 / 2.09 |

Faster-or-equal cells vs sglang, 49 uniform + 28 variable-length per GPU: B200 35 -> 49 uniform, 19 -> 26 variable-length; B300 34 -> 48, 16 -> 23; Rubin (SM107, 208 SMs, same job layout in the cu135 container) 35 -> 49, 19 -> 26. Rubin ratios with this commit, uniform, bs 1 / 16 / 64 / 128 / 256: width 4096 1.05 / 1.10 / 1.11 / 1.12 / 1.05; 6144 1.13 / 1.13 / 1.12 / 1.11 / 1.55; 8192 1.17 / 1.19 / 1.18 / 1.18 / 1.49; 12288 1.15 / 1.11 / 1.12 / 1.12 / 1.37; 16384 1.23 / 1.23 / 1.27 / 1.26 / 1.48; 32768 1.92 / 1.91 / 1.66 / 1.39 / 1.69; 131072 1.63 / 1.12 / 2.87 / 2.31 / 2.89. The remaining misses are bs 256 at width 12288 (0.96-1.01, the second CTA wave of the new rung) and few-row variable-length cells at width 4096 (about 1 us total; sglang's short-row early exit is 0.3-0.5 us cheaper). Absolute at the reported cell (width 8192, B200): sglang 4.27-4.48 us vs `gvr_2` 3.62-3.76 us for bs 1-128; 6.77 vs 4.89 us at bs 256. Note that sglang's kernel fuses the page-table transform, so it does slightly more work per call.

**SM-count-aware register band + second wave (`f15221e5`).** Tuning the same band on B300 (sm_103, 160 SMs) and Rubin (sm_107, 208 SMs) showed the remaining losses sat exactly where `route()`'s hard-coded 148 (the B200 SM count) mis-sizes a wave on other parts. `route(b, n, npad, k, sms=148)` now takes the device SM count (`_sm_count()`, part of both launcher cache keys) for the register-band tests only: `wide = b <= sms`, the `QC`/`CURE` flags, and the BLK=512 rung's cutoff (`b <= 2*sms`, plus a second wave `b <= 4*sms` for n >= 6144, where it still beats the slab: B200 b=512 +11-12%, and the ragged-row 6144-8192 x 320-400 cells the slab lost by up to 22%). The streaming constants stay at 148 on every part: scaling them made the 131072 x 192-256 slab cells 3-28% slower on B300 and Rubin. Tried and rejected: swapping the `regimg` flavour for plain `reg` at n <= 4K (+5%..-12% by cell and hint quality), clustered-register configs for 8K < n <= 16K at b > sms (2-4x slower), `_reg(1024, 4, 1)` for the second wave of 8K-16K (mixed). Same-node vs sglang after this commit, 84 uniform (bs 1-512, 7 widths) + 48 variable-length cells per GPU, gvr_2 faster-or-equal: B200 83/84 and 45/48 (upstream routing: 65 / 35); B300 82/84 and 43/48 (62 / 31); Rubin 84/84 and 46/48 (58 / 31). What sglang still wins: 1-4 randomly short rows at width 4096 (~1 us calls, its short-row early exit is 0.3-0.5 us cheaper); the second CTA wave at width 12288 (B200 bs160 0.99; B300 bs256 0.97, bs320 0.91); B300 ragged rows at width 4096 for bs 16-64 (0.96-0.99) and width 5120 at bs320 (0.93-0.97 on B200/B300, the lower half of the band where the second wave loses on uniform rows).

**Hint-free `gvr_2`.** `pre_idx` is now optional for `gvr_2`: `run_varlen(pre_idx=None, top_k=K)` runs on a cached `arange(k)` table (`_hint_free_pre_idx`, per (device, k), grown eagerly, refused under CUDA-graph capture, pre-sized by `warmup_varlen`). The kernels use the hint only as a sampling anchor, so the result is exact; measured hint-free vs perfect-hint on isolated B200/B300 and on B100/Rubin: within ~10% on most cells, and 1.5-5x ahead of `radix` / `radix_filter` everywhere except K >= 2048, N <= 4096, B = 1 (radix_filter 1.3x faster). `auto` therefore ranks `gvr_2` first for fp32 with or without a hint, with that one cell carved out to `radix_filter`; a malformed hint on `backend="gvr_2"` is discarded with the existing RuntimeWarning and the call still runs on `gvr_2`; the "requires pre_idx" refusal now applies to `gvr` only.

Tests added: hint-free exactness on every kernel family incl. short/MTP/cr=4 rows, hint-free CUDA-graph replay with table-growth refusal, host `top_k` contract, the two rungs (tuple, NBH == IMGOFF, one-wave cutoff, `route_split` parity), auto order with and without a hint, per-backend malformed-hint behaviour. Gate before push: full `tests/topk_varlen` on A100, L40S, H100, B100, DRIVE P2021, RTX 5080, Rubin and isolated computelab B200 and B300 (322-323 passed on SM100-class, 74-77 elsewhere, 2 known xfails).

### DIFF
--- a/benchmarks/bench_topk_varlen_gvr2.py
+++ b/benchmarks/bench_topk_varlen_gvr2.py
@@ -59,7 +59,7 @@ def parse_args():
     p = argparse.ArgumentParser(description=__doc__)
     p.add_argument("--batch-sizes", type=str, default="1,4,16,32,64,128,256")
     p.add_argument("--n-vals", type=str, default="4096,8192,16384,32768,65536,131072")
-    p.add_argument("--top-k", type=int, default=1024, choices=[512, 1024, 2048])
+    p.add_argument("--top-k", type=int, default=1024)
     p.add_argument(
         "--scenarios",
         type=str,
@@ -69,6 +69,14 @@ def parse_args():
     )
     p.add_argument("--next-n", type=int, default=1)
     p.add_argument("--compress-ratio", type=int, default=1, choices=[1, 4])
+    p.add_argument(
+        "--dtype",
+        type=str,
+        default="fp32",
+        choices=["fp32", "bf16", "fp16"],
+        help="logits dtype; gvr_2/trtllm_gvr2 are fp32-only and are dropped "
+        "for bf16/fp16",
+    )
     p.add_argument("--hint-quality", type=float, default=0.6)
     p.add_argument("--oracle", action="store_true", help="perfect pre_idx hints")
     p.add_argument(
@@ -84,11 +92,13 @@ def parse_args():
     return p.parse_args()
 
 
-def make_inputs(scenario, batch, N, K, nn, cr, hint_quality, oracle, seed):
-    """fp32 logits [batch*nn, N] (compressed space), uncompressed seq_lens."""
+def make_inputs(scenario, batch, N, K, nn, cr, hint_quality, oracle, seed, dtype):
+    """Logits [batch*nn, N] (compressed space), uncompressed seq_lens."""
     gen = torch.Generator(device="cuda").manual_seed(seed)
     rows = batch * nn
-    logits = torch.randn(rows, N, generator=gen, device="cuda", dtype=torch.float32)
+    logits = torch.randn(rows, N, generator=gen, device="cuda", dtype=torch.float32).to(
+        dtype
+    )
     if scenario == "uniform":
         seq_lens = torch.full((batch,), N * cr, dtype=torch.int32, device="cuda")
     elif scenario == "mixed":
@@ -265,6 +275,16 @@ def main():
     scenarios = args.scenarios.split(",")
     backends = args.backends.split(",")
     K, nn, cr = args.top_k, args.next_n, args.compress_ratio
+    dtype = {
+        "fp32": torch.float32,
+        "bf16": torch.bfloat16,
+        "fp16": torch.float16,
+    }[args.dtype]
+    if args.dtype != "fp32":
+        dropped = [b for b in backends if b in ("fi_gvr2", "trtllm_gvr2")]
+        if dropped:
+            print(f"[WARN] {dropped} are fp32-only; dropping for {args.dtype}")
+            backends = [b for b in backends if b not in dropped]
 
     trt_host = None
     if "trtllm_gvr2" in backends:
@@ -281,7 +301,7 @@ def main():
 
     print(f"device: {torch.cuda.get_device_name(0)}")
     print(
-        f"top_k={K} next_n={nn} cr={cr} hint_quality="
+        f"dtype={args.dtype} top_k={K} next_n={nn} cr={cr} hint_quality="
         f"{'oracle' if args.oracle else args.hint_quality} "
         f"timing=cuda-graph median, repeat_ms={args.repeat_ms}"
     )
@@ -307,6 +327,7 @@ def main():
                     args.hint_quality,
                     args.oracle,
                     seed=args.seed + B * 7 + N // 64,
+                    dtype=dtype,
                 )
                 row = {}
                 for backend in backends:
--- a/benchmarks/bench_topk_varlen_gvr2.py
+++ b/benchmarks/bench_topk_varlen_gvr2.py
@@ -101,20 +101,27 @@ def make_inputs(scenario, batch, N, K, nn, cr, hint_quality, oracle, seed, dtype
     )
     if scenario == "uniform":
         seq_lens = torch.full((batch,), N * cr, dtype=torch.int32, device="cuda")
-    elif scenario == "mixed":
-        seq_lens = torch.randint(
-            K * cr + nn, N * cr + 1, (batch,), generator=gen, device="cuda"
-        ).int()
-    elif scenario == "short":
-        seq_lens = torch.randint(
-            K * cr + nn,
-            max(N * cr // 8, K * cr + nn + 1),
-            (batch,),
-            generator=gen,
-            device="cuda",
-        ).int()
-        n_long = max(batch // 4, 1)
-        seq_lens[:n_long] = N * cr
+    elif scenario in ("mixed", "short"):
+        # Smallest uncompressed kv whose EVERY next_n row is a long row
+        # (n_r = (kv - nn + j + 1) // cr >= K + 1 for all j): keeps the
+        # documented domains ("mixed" = rows in [K+1, N]) so no row silently
+        # bypasses refcheck's n <= K skip.
+        lo = (K + 1) * cr + nn - 1
+        if scenario == "mixed":
+            hi = N * cr + 1
+        else:
+            hi = N * cr // 8
+            if hi <= lo:
+                print(
+                    f"[NOTE] short scenario is degenerate for N={N}, K={K}: "
+                    "N/8 <= K+1, so the 'short' rows are not short",
+                    flush=True,
+                )
+        hi = max(hi, lo + 1)
+        seq_lens = torch.randint(lo, hi, (batch,), generator=gen, device="cuda").int()
+        if scenario == "short":
+            n_long = max(batch // 4, 1)
+            seq_lens[:n_long] = N * cr
     else:
         raise ValueError(scenario)
 
@@ -142,6 +149,13 @@ def make_backend_fns(backend, logits, seq_lens, pre_idx, K, nn, cr, trt_host):
     or None if the backend cannot run this config."""
     rows = logits.shape[0]
     out_i = torch.empty(rows, K, dtype=torch.int32, device="cuda")
+
+    if backend in ("fi_gvr2", "fi_gvr", "fi_gvr_nolb", "trtllm_gvr2") and K not in (
+        512,
+        1024,
+        2048,
+    ):
+        return None, None  # outside the GVR-family top_k domain: n/a, not an error
 
     if backend == "fi_gvr2":
 
@@ -369,16 +383,17 @@ def main():
                     if msg and msg != "n/a":
                         print(f"    [{b}] {msg}", flush=True)
 
-    # ---- summary ----------------------------------------------------------
-    if "fi_gvr2" not in backends:
+    # ---- summary (relative to gvr_2 when present, else the first backend) ---
+    if not backends:
         return
-    print("\n================ SUMMARY (ratios are other/gvr_2; >1 = gvr_2 faster)")
+    base = "fi_gvr2" if "fi_gvr2" in backends else backends[0]
+    print(f"\n================ SUMMARY (ratios are other/{base}; >1 = {base} faster)")
     for other in backends:
-        if other == "fi_gvr2":
+        if other == base:
             continue
         ratios, wins, total = [], 0, 0
         for _, _, _, row in results:
-            g2, g2_ok, _ = row["fi_gvr2"]
+            g2, g2_ok, _ = row[base]
             ot, ot_ok, _ = row.get(other, (float("nan"), False, ""))
             if math.isnan(g2) or math.isnan(ot) or not (g2_ok and ot_ok):
                 continue
@@ -388,7 +403,7 @@ def main():
         if total:
             print(
                 f"  vs {other:<17} geomean {geomean(ratios):5.2f}x   "
-                f"gvr_2 wins {wins}/{total}"
+                f"{base} wins {wins}/{total}"
             )
     print("\nper-config winner:")
     for scenario, B, N, row in results:
--- a/benchmarks/bench_topk_varlen_gvr2.py
+++ b/benchmarks/bench_topk_varlen_gvr2.py
@@ -82,7 +82,7 @@ def parse_args():
     p.add_argument(
         "--backends",
         type=str,
-        default="fi_gvr2,fi_gvr,fi_gvr_nolb,fi_radix,fi_radix_cutlass",
+        default="fi_gvr2,fi_gvr,fi_gvr_nolb,fi_radix,fi_radix_cutlass,fi_radix_filter",
         help="comma list; add trtllm_gvr2 for the upstream port-parity twin",
     )
     p.add_argument("--trtllm-path", type=str, default=_DEFAULT_TRTLLM_TOPK)
@@ -226,6 +226,22 @@ def make_backend_fns(backend, logits, seq_lens, pre_idx, K, nn, cr, trt_host):
                 compress_ratio=cr,
                 out_indices=out_i,
                 backend="radix_cutlass",
+            )
+
+        return fn, out_i
+    if backend == "fi_radix_filter":
+        if cr != 1:
+            return None, None  # the vendored DKG kernel has no compress_ratio
+
+        def fn():
+            flashinfer.top_k_varlen(
+                logits,
+                seq_lens,
+                K,
+                next_n=nn,
+                compress_ratio=cr,
+                out_indices=out_i,
+                backend="radix_filter",
             )
 
         return fn, out_i
--- a/benchmarks/bench_topk_varlen_gvr2.py
+++ b/benchmarks/bench_topk_varlen_gvr2.py
@@ -1,0 +1,381 @@
+# Copyright (c) 2026 by FlashInfer team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Benchmark the gvr_2 (self-sampling GVR V2) top_k_varlen backend.
+
+Compares, on identical fp32 inputs (gvr_2 is fp32-only):
+
+* ``fi_gvr2``          — the new self-sampling GVR V2 backend (this PR)
+* ``fi_gvr``           — existing GVR, load-balance path (production default)
+* ``fi_gvr_nolb``      — existing GVR, single-CTA path
+* ``fi_radix``         — CuTe DSL single-pass multi-CTA radix
+* ``fi_radix_cutlass`` — masked CUTLASS radix fallback
+* ``trtllm_gvr2``      — upstream TRT-LLM run_varlen (port-parity check;
+                         needs a local TRT-LLM checkout, see --trtllm-path)
+
+Timing: ``bench_gpu_time(..., use_cuda_graph=True)`` — decode top-k kernels
+are microsecond-scale, so graph replay removes launch-overhead noise (same
+method as bench_topk_varlen_vs_trtllm.py). Every backend's output is
+validated (tie-aware value multiset per row) before timing; a wrong backend
+is reported as WRONG and excluded from win statistics.
+
+Hints: ``pre_idx`` mixes a ``--hint-quality`` fraction of the true top-K with
+random valid indices (hint quality shifts GVR-family perf but never
+exactness). ``--oracle`` uses perfect hints (GVR-favorable upper bound).
+
+Example:
+    python benchmarks/bench_topk_varlen_gvr2.py \
+        --batch-sizes 1,16,64,256 --n-vals 8192,32768,131072 --top-k 1024
+"""
+
+import argparse
+import math
+import os
+import sys
+
+import torch
+
+import flashinfer
+from flashinfer.testing import bench_gpu_time
+
+# Directory holding TRT-LLM's top_k CuTe-DSL kernels (a local checkout's
+# tensorrt_llm/_torch/cute_dsl_kernels/blackwell/top_k), used only by the
+# optional trtllm_gvr2 port-parity twin. No path ships as a default — pass
+# --trtllm-path or set TRTLLM_TOPK_DIR.
+_DEFAULT_TRTLLM_TOPK = os.environ.get("TRTLLM_TOPK_DIR", "")
+
+
+def parse_args():
+    p = argparse.ArgumentParser(description=__doc__)
+    p.add_argument("--batch-sizes", type=str, default="1,4,16,32,64,128,256")
+    p.add_argument("--n-vals", type=str, default="4096,8192,16384,32768,65536,131072")
+    p.add_argument("--top-k", type=int, default=1024, choices=[512, 1024, 2048])
+    p.add_argument(
+        "--scenarios",
+        type=str,
+        default="uniform,mixed",
+        help="comma list of: uniform (all rows = N), mixed (ragged in "
+        "[K+1, N]), short (75%% of rows <= N/8)",
+    )
+    p.add_argument("--next-n", type=int, default=1)
+    p.add_argument("--compress-ratio", type=int, default=1, choices=[1, 4])
+    p.add_argument("--hint-quality", type=float, default=0.6)
+    p.add_argument("--oracle", action="store_true", help="perfect pre_idx hints")
+    p.add_argument(
+        "--backends",
+        type=str,
+        default="fi_gvr2,fi_gvr,fi_gvr_nolb,fi_radix,fi_radix_cutlass",
+        help="comma list; add trtllm_gvr2 for the upstream port-parity twin",
+    )
+    p.add_argument("--trtllm-path", type=str, default=_DEFAULT_TRTLLM_TOPK)
+    p.add_argument("--repeat-ms", type=int, default=200)
+    p.add_argument("--seed", type=int, default=0)
+    p.add_argument("--no-refcheck", action="store_true")
+    return p.parse_args()
+
+
+def make_inputs(scenario, batch, N, K, nn, cr, hint_quality, oracle, seed):
+    """fp32 logits [batch*nn, N] (compressed space), uncompressed seq_lens."""
+    gen = torch.Generator(device="cuda").manual_seed(seed)
+    rows = batch * nn
+    logits = torch.randn(rows, N, generator=gen, device="cuda", dtype=torch.float32)
+    if scenario == "uniform":
+        seq_lens = torch.full((batch,), N * cr, dtype=torch.int32, device="cuda")
+    elif scenario == "mixed":
+        seq_lens = torch.randint(
+            K * cr + nn, N * cr + 1, (batch,), generator=gen, device="cuda"
+        ).int()
+    elif scenario == "short":
+        seq_lens = torch.randint(
+            K * cr + nn,
+            max(N * cr // 8, K * cr + nn + 1),
+            (batch,),
+            generator=gen,
+            device="cuda",
+        ).int()
+        n_long = max(batch // 4, 1)
+        seq_lens[:n_long] = N * cr
+    else:
+        raise ValueError(scenario)
+
+    # request-level hints: hint_quality fraction of the true top-K of the
+    # request's first row, rest random valid indices
+    pre_idx = torch.zeros(batch, K, dtype=torch.int32, device="cuda")
+    sl = seq_lens.cpu().tolist()
+    for b in range(batch):
+        n0 = min(max((sl[b] - nn + 1) // cr, 1), N)
+        k_eff = min(K, n0)
+        true_top = torch.topk(logits[b * nn, :n0], k_eff).indices.int()
+        if oracle:
+            pre_idx[b, :k_eff] = true_top
+        else:
+            n_hit = int(k_eff * hint_quality)
+            pre_idx[b] = torch.randint(
+                0, n0, (K,), generator=gen, device="cuda", dtype=torch.int32
+            )
+            pre_idx[b, :n_hit] = true_top[:n_hit]
+    return logits, seq_lens, pre_idx
+
+
+def make_backend_fns(backend, logits, seq_lens, pre_idx, K, nn, cr, trt_host):
+    """Return a zero-arg closure running the backend into preallocated outputs,
+    or None if the backend cannot run this config."""
+    rows = logits.shape[0]
+    out_i = torch.empty(rows, K, dtype=torch.int32, device="cuda")
+
+    if backend == "fi_gvr2":
+
+        def fn():
+            flashinfer.top_k_varlen(
+                logits,
+                seq_lens,
+                K,
+                pre_idx=pre_idx,
+                next_n=nn,
+                compress_ratio=cr,
+                out_indices=out_i,
+                backend="gvr_2",
+            )
+
+        return fn, out_i
+    if backend in ("fi_gvr", "fi_gvr_nolb"):
+        lb = backend == "fi_gvr"
+        ws = None
+        if lb:
+            if seq_lens.shape[0] > 1024:
+                return None, None  # LB prepare kernel cap
+            mbs = 64
+            while mbs < seq_lens.shape[0]:
+                mbs *= 2
+            ws = {
+                "gvr_order_row": torch.empty(mbs, dtype=torch.int32, device="cuda"),
+                "gvr_counters": torch.empty(2, dtype=torch.int32, device="cuda"),
+            }
+
+        def fn():
+            flashinfer.top_k_varlen(
+                logits,
+                seq_lens,
+                K,
+                pre_idx=pre_idx,
+                next_n=nn,
+                compress_ratio=cr,
+                out_indices=out_i,
+                backend="gvr",
+                load_balance=lb,
+                workspace=ws,
+            )
+
+        return fn, out_i
+    if backend == "fi_radix":
+
+        def fn():
+            flashinfer.top_k_varlen(
+                logits,
+                seq_lens,
+                K,
+                next_n=nn,
+                compress_ratio=cr,
+                out_indices=out_i,
+                backend="radix",
+            )
+
+        return fn, out_i
+    if backend == "fi_radix_cutlass":
+
+        def fn():
+            flashinfer.top_k_varlen(
+                logits,
+                seq_lens,
+                K,
+                next_n=nn,
+                compress_ratio=cr,
+                out_indices=out_i,
+                backend="radix_cutlass",
+            )
+
+        return fn, out_i
+    if backend == "trtllm_gvr2":
+        if trt_host is None:
+            return None, None
+
+        msl = logits.shape[1] * cr
+
+        def fn():
+            trt_host.run_varlen(
+                logits,
+                pre_idx,
+                seq_lens,
+                out_i,
+                next_n=nn,
+                compress_ratio=cr,
+                max_seq_len=msl,
+            )
+
+        return fn, out_i
+    raise ValueError(backend)
+
+
+def refcheck(logits, seq_lens, out_i, K, nn, cr):
+    """Tie-aware per-row check; rows with n<=K only checked for -1 padding
+    conventions loosely (backends differ there). Returns (ok, msg)."""
+    rows = logits.shape[0]
+    sl = seq_lens.cpu().tolist()
+    for r in range(rows):
+        n = min(max((sl[r // nn] - nn + (r % nn) + 1) // cr, 0), logits.shape[1])
+        if n <= K:
+            continue  # short-row pad conventions differ across backends
+        idx = out_i[r].to(torch.int64)
+        if int(idx.min()) < 0 or int(idx.max()) >= n:
+            return False, f"row {r}: index out of range"
+        if int(torch.unique(idx).numel()) != K:
+            return False, f"row {r}: duplicate indices"
+        got = torch.sort(logits[r].gather(0, idx) + 0.0, descending=True).values
+        ref = torch.sort(
+            torch.topk(logits[r, :n], K).values + 0.0, descending=True
+        ).values
+        if not torch.equal(got, ref):
+            nbad = int((got != ref).sum())
+            return False, f"row {r}: value multiset mismatch ({nbad}/{K} slots)"
+    return True, ""
+
+
+def geomean(xs):
+    xs = [x for x in xs if x > 0]
+    if not xs:
+        return float("nan")
+    return math.exp(sum(math.log(x) for x in xs) / len(xs))
+
+
+def main():
+    args = parse_args()
+    torch.manual_seed(args.seed)
+    batch_sizes = [int(x) for x in args.batch_sizes.split(",")]
+    n_vals = [int(x) for x in args.n_vals.split(",")]
+    scenarios = args.scenarios.split(",")
+    backends = args.backends.split(",")
+    K, nn, cr = args.top_k, args.next_n, args.compress_ratio
+
+    trt_host = None
+    if "trtllm_gvr2" in backends:
+        if not args.trtllm_path:
+            print("[WARN] trtllm_gvr2 needs --trtllm-path or TRTLLM_TOPK_DIR; dropping")
+            backends = [b for b in backends if b != "trtllm_gvr2"]
+        else:
+            sys.path.insert(0, args.trtllm_path)
+            try:
+                import gvr_topk_decode_self_sampling_host as trt_host  # noqa: N813
+            except ImportError as e:
+                print(f"[WARN] trtllm_gvr2 unavailable ({e}); dropping")
+                backends = [b for b in backends if b != "trtllm_gvr2"]
+
+    print(f"device: {torch.cuda.get_device_name(0)}")
+    print(
+        f"top_k={K} next_n={nn} cr={cr} hint_quality="
+        f"{'oracle' if args.oracle else args.hint_quality} "
+        f"timing=cuda-graph median, repeat_ms={args.repeat_ms}"
+    )
+    header = f"{'scenario':<8} {'B':>4} {'N':>7} " + "".join(
+        f"{b:>17}" for b in backends
+    )
+    print(header)
+    print("-" * len(header))
+
+    results = []  # (scenario, B, N, {backend: (us, ok)})
+    for scenario in scenarios:
+        for B in batch_sizes:
+            for N in n_vals:
+                if N <= K:
+                    continue
+                logits, seq_lens, pre_idx = make_inputs(
+                    scenario,
+                    B,
+                    N,
+                    K,
+                    nn,
+                    cr,
+                    args.hint_quality,
+                    args.oracle,
+                    seed=args.seed + B * 7 + N // 64,
+                )
+                row = {}
+                for backend in backends:
+                    fn, out_i = make_backend_fns(
+                        backend, logits, seq_lens, pre_idx, K, nn, cr, trt_host
+                    )
+                    if fn is None:
+                        row[backend] = (float("nan"), False, "n/a")
+                        continue
+                    try:
+                        fn()  # warmup / JIT outside timing
+                        torch.cuda.synchronize()
+                        ok, msg = (True, "")
+                        if not args.no_refcheck:
+                            ok, msg = refcheck(logits, seq_lens, out_i, K, nn, cr)
+                        timings = bench_gpu_time(
+                            fn,
+                            dry_run_time_ms=50,
+                            repeat_time_ms=args.repeat_ms,
+                            use_cuda_graph=True,
+                            num_iters_within_graph=10,
+                        )
+                        med_us = float(torch.tensor(timings).median()) * 1e3
+                        row[backend] = (med_us, ok, msg)
+                    except Exception as e:  # noqa: BLE001 — report and move on
+                        row[backend] = (float("nan"), False, f"{type(e).__name__}: {e}")
+                results.append((scenario, B, N, row))
+                cells = ""
+                for b in backends:
+                    us, ok, msg = row[b]
+                    if math.isnan(us):
+                        cells += f"{'—':>17}"
+                    else:
+                        tag = "" if ok else "!WRONG"
+                        cells += f"{us:>11.1f}us{tag:<5}"
+                print(f"{scenario:<8} {B:>4} {N:>7} {cells}", flush=True)
+                for b in backends:
+                    us, ok, msg = row[b]
+                    if msg and msg != "n/a":
+                        print(f"    [{b}] {msg}", flush=True)
+
+    # ---- summary ----------------------------------------------------------
+    if "fi_gvr2" not in backends:
+        return
+    print("\n================ SUMMARY (ratios are other/gvr_2; >1 = gvr_2 faster)")
+    for other in backends:
+        if other == "fi_gvr2":
+            continue
+        ratios, wins, total = [], 0, 0
+        for _, _, _, row in results:
+            g2, g2_ok, _ = row["fi_gvr2"]
+            ot, ot_ok, _ = row.get(other, (float("nan"), False, ""))
+            if math.isnan(g2) or math.isnan(ot) or not (g2_ok and ot_ok):
+                continue
+            ratios.append(ot / g2)
+            wins += ot > g2
+            total += 1
+        if total:
+            print(
+                f"  vs {other:<17} geomean {geomean(ratios):5.2f}x   "
+                f"gvr_2 wins {wins}/{total}"
+            )
+    print("\nper-config winner:")
+    for scenario, B, N, row in results:
+        valid = {b: us for b, (us, ok, _) in row.items() if ok and not math.isnan(us)}
+        if valid:
+            w = min(valid, key=valid.get)
+            print(f"  {scenario:<8} B={B:<4} N={N:<7} -> {w} ({valid[w]:.1f}us)")
+
+
+if __name__ == "__main__":
+    main()
--- a/benchmarks/routines/topk_varlen.py
+++ b/benchmarks/routines/topk_varlen.py
@@ -29,12 +29,12 @@ from .flashinfer_benchmark_utils import (
     print_perf_metrics,
 )
 
-# top_k_varlen's three runners. Per-CC support is NOT hard-coded here; it is
+# top_k_varlen's runners. Per-CC support is NOT hard-coded here; it is
 # resolved at runtime from top_k_varlen's @backend_requirement decorator via
 # ``flashinfer.top_k_varlen.is_backend_supported(backend, cc)`` (the single
 # source of truth), mirroring how the GEMM routines rely on their support
 # checkers (e.g. mm_fp4 / bmm_fp8) instead of a compute-capability table.
-_TOP_K_VARLEN_BACKENDS = ("radix", "gvr", "radix_cutlass")
+_TOP_K_VARLEN_BACKENDS = ("radix", "gvr", "gvr_2", "radix_cutlass")
 
 
 def parse_topk_varlen_args(line, parser):
@@ -96,7 +96,7 @@ def run_topk_varlen_test(args):
 
 
 def testTopKVarlen(args):
-    """Benchmark top_k_varlen with three runners: 'radix', 'radix_cutlass', 'gvr'.
+    """Benchmark top_k_varlen with its runners: 'radix', 'radix_cutlass', 'gvr', 'gvr_2'.
 
     Runners
     -------
@@ -107,6 +107,9 @@ def testTopKVarlen(args):
                      (sm_100/103) only. Per-CC support is resolved via
                      ``top_k_varlen.is_backend_supported`` (its
                      ``@backend_requirement`` decorator).
+    gvr_2          — self-sampling GVR V2 (TRT-LLM PR #17821 port); passes the
+                     same ``pre_idx``. Blackwell (sm_100/103) + fp32 logits +
+                     top_k in {512, 1024, 2048} only; skipped otherwise.
 
     Reference check compares the *set* of selected indices against ``torch.topk``
     applied to logits masked to ``seq_lens``.
@@ -143,6 +146,16 @@ def testTopKVarlen(args):
                 f"[WARNING] {backend} for routine {args.routine} is not supported "
                 f"on compute capability {major}.{minor}. Skipping."
             )
+    # gvr_2's CC check passes on sm_100/103, but the backend is fp32-only with
+    # a fixed top_k domain — drop it up front instead of failing the call.
+    if "gvr_2" in backends and (
+        args.input_dtype != "float32" or top_k not in (512, 1024, 2048)
+    ):
+        backends.remove("gvr_2")
+        print(
+            "[WARNING] gvr_2 requires float32 logits and top_k in {512, 1024, "
+            "2048}. Skipping."
+        )
     if len(backends) == 0:
         print("[ERROR] No backends to test. Exiting.")
         return res
@@ -176,6 +189,10 @@ def testTopKVarlen(args):
         elif backend == "gvr":
             return flashinfer.top_k_varlen(
                 logits, seq_lens, top_k, pre_idx=pre_idx, backend="gvr"
+            )
+        elif backend == "gvr_2":
+            return flashinfer.top_k_varlen(
+                logits, seq_lens, top_k, pre_idx=pre_idx, backend="gvr_2"
             )
         else:
             raise ValueError(f"Unsupported backend: {backend}")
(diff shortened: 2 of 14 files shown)


dhiraj113:
@flashinfer-bot run

dhiraj113:
@flashinfer-bot rerun

dhiraj113:
## Full backend comparison (213 configs, B200, fp32, CUDA-graph timing, tie-aware output validation)

Sweeps: uniform + mixed (ragged) at K=1024; K=512 including the N≤2K small-N band; K=2048; a `short` scenario (75% of rows near K); compress_ratio=4; next_n=2. Every backend's output is validated per config (tie-aware value-multiset vs `torch.topk`) before timing; invalid outputs are excluded from win/loss stats. GPU was held exclusively during the window.

### Pooled results (`ratio = other / gvr_2`, >1 ⇒ gvr_2 faster)

| vs backend | configs | geomean | gvr_2 faster | min | max |
|---|---|---|---|---|---|
| `gvr` (existing, LB) | 211 | **2.58×** | 210/211 | 0.47× | 6.23× |
| `gvr` (non-LB) | 211 | 3.80× | 210/211 | 0.76× | 10.94× |
| `radix` (CuTe DSL) | 213 | 2.13× | 211/213 | 0.30× | 10.05× |
| `radix_cutlass` | 213 | 3.29× | 211/213 | 0.56× | 5.92× |
| TRT-LLM upstream twin | 84 | **1.00×** | — | 0.95× | 1.05× |

Per-sweep geomeans vs `gvr` / `radix`: uniform 2.52×/2.29×, mixed 2.49×/2.31×, K=512 3.08×/2.11×, K=2048 2.49×/1.81×, small-N band (N∈{2k,4k}, K=1024) 2.93×/1.65×, short scenario 1.82×/1.77×, cr=4 2.21×/2.67×, next_n=2 2.16×/2.49×.

### Does gvr_2 always win? Almost — the losing regime is narrow and known

All 6 losing cells live in one regime: the `short` scenario at N=8192, B∈{16,64}, where ~75% of rows have `N_eff ≈ K+1` (barely above the short path). Worst cell: gvr_2 28 µs vs radix 8.3 µs (0.30×). This matches the upstream PR's own documented trade-off (short rows at high row counts); the sampling-ladder machinery is pure overhead when most rows barely exceed K. Since `auto` ordering keeps `gvr_2` behind `gvr`, this regime regresses nothing; worth revisiting if/when gvr_2 is promoted (e.g. an N_eff-aware fallback).

### Data-quality flag on the existing `gvr` (V1) backend (pre-existing, not from this PR)

In 4 cells of the `short` scenario (B∈{64,256}, N=8192, rows with `N_eff = K+1 = 1025`), the **existing `gvr` backend (both LB and non-LB) produced incorrect top-K value multisets** (excluded from stats above). gvr_2, radix, and radix_cutlass are correct on the same inputs. Plausibly related to the V1 threshold-search behavior fixed upstream in NVIDIA/TensorRT-LLM#18094, which postdates the V1 port — will file separately with a reproducer.

dhiraj113:
@flashinfer-bot rerun

dhiraj113:
## `auto` now tracks the measured per-config winner (+ one upstream caveat found)

### Shape/dtype-aware auto selection

The static gvr-first `auto` policy left large factors on the table: it never reached `gvr_2`, always picked `gvr` for hinted bf16/fp16 (where `radix` wins by up to 2.8×), and never reached `radix_cutlass` in its fp32 big corner (up to 2.8×). Replaced with a ranking derived from a 500+-cell B200 sweep (fp32/bf16/fp16, K ∈ {512, 1024, 2048}, B ∈ [1, 256], N ∈ [1K, 128K], uniform/mixed/short lengths, cr ∈ {1,4}, next_n ∈ {1,2}), using only capture-stable host facts (dtype, N, B — never `seq_lens` contents, which would cost a D2H sync):

1. hinted fp32 → **`gvr_2`** (won 211/213 cells);
2. `gvr` outranks `radix` only when `B·N ≥ 2²²` (fp32) / `2²³` (bf16/fp16);
3. `radix_cutlass` outranks `radix` only in the fp32 corner `N ≥ 65536` and `B·N ≥ 2²³`; in half precision `radix` always leads.

**Validation on 100 off-grid configs** (B ∈ {2,8,48,96,192}, N ∈ {6K,12K,24K,48K,96K} — none used to fit the thresholds); regret = t(auto's pick) / t(fastest backend):

| policy | geomean | max | oracle hits |
|---|---|---|---|
| new | **1.004×** | 1.16× | 97/100 |
| old | 1.317× (hinted 1.694×) | 3.05× | 55/100 |

Documented static blind spot: batches whose rows are mostly far shorter than N favor `radix` at large N but are indistinguishable without reading `seq_lens`; such callers should pass `backend="radix"` explicitly.

### Upstream caveat found while hardening the tests: `gvr_2` drops literal `+inf` logits

Deterministically reseeding the adversarial tests exposed that `gvr_2` selects **zero** `+inf` logits at (N=32768, B=4) (`reg_clus` family), 40/40 seeds — and the **upstream TRT-LLM implementation fails bit-identically on the same inputs** (gvr/radix return all infs correctly), so this is inherited upstream kernel behavior, not a port defect. Probing bounds it exactly: 2.9e38, **3.1e38 (above the kernel's 3e38 pad sentinel)**, and 1e30 floods, plus `-inf`, are all tie-aware exact — only literal `+inf` is special-cased away, the same class as upstream's documented implementation-specific NaN ordering. Encoded as an API-docs caveat + a non-strict `xfail` test that flips visible if a future upstream sync fixes it; worth reporting to TRT-LLM (reproducer in the xfail test).

dhiraj113:
@flashinfer-bot rerun

dhiraj113:
@flashinfer-bot rerun

dhiraj113:
@flashinfer-bot rerun

dhiraj113:
@flashinfer-bot rerun

dhiraj113:
@flashinfer-bot rerun

dhiraj113:
@flashinfer-bot run

dhiraj113:
@flashinfer-bot run

flashinfer/topk_varlen/topk_varlen.py
```diff
@@ -783,6 +841,49 @@ def _run_gvr(
...
+        logits,
+        pre_idx,
+        seq_lens,
+        out_indices,
+        next_n=next_n,
+        compress_ratio=compress_ratio,
+        values=out_values if return_output_values else None,
+        max_seq_len=logits.shape[1] * compress_ratio,
+        workspace=workspace.get("gvr2_workspace") if workspace else None,
+    )
```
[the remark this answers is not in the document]
dhiraj113: Fixed in ea069804. Confirmed against the kernel source: the `reg`, `reg_clus` and `clus` varlen prologues already clamp the kv-derived row length to the envelope argument `n`, but `gvr_main` clamped to the row stride `npad`, so on a wider-stride arena view an oversized `seq_lens` value could classify the arena columns in `[width, stride)`. `gvr_main` now clamps to the envelope too: the host passes the envelope in the previously dead `n` launch slot and the prologue clamps to it (radix parity — `seq_lens` is effectively clamped to `logits.shape[1]`). Contiguous inputs are unaffected (`npad == width`). Regression test `test_gvr2_arena_view_oversized_seq_lens` covers all four families on a 256-rounded arena view with the arena tail poisoned to `+3e38` and half the rows oversized; it fails on the previous kernel for both `gvr_main` parametrisations and passes now.

tests/topk_varlen/test_topk_varlen_gvr2.py
```diff
@@ -0,0 +1,743 @@
...
+# backend registration / guards / cross-backend
+# ---------------------------------------------------------------------------
+
+
+@requires_gvr2
+def test_gvr2_rejects_non_fp32():
+    logits = torch.randn(4, 8192, dtype=torch.bfloat16, device=_DEV)
+    seq_lens = torch.full((4,), 8192, dtype=torch.int32, device=_DEV)
+    pre_idx = torch.zeros(4, 512, dtype=torch.int32, device=_DEV)
+    with pytest.raises(Exception, match="not supported|Problem size"):
```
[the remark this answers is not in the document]
dhiraj113: Done in ea069804: the three regex `match=` patterns are raw strings now.

benchmarks/bench_topk_varlen_gvr2.py
```diff
@@ -0,0 +1,381 @@
...
+            K * cr + nn, N * cr + 1, (batch,), generator=gen, device="cuda"
+        ).int()
+    elif scenario == "short":
+        seq_lens = torch.randint(
+            K * cr + nn,
+            max(N * cr // 8, K * cr + nn + 1),
+            (batch,),
+            generator=gen,
+            device="cuda",
+        ).int()
```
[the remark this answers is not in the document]
dhiraj113: Fixed in ea069804. `mixed` and `short` now draw from `lo = (K + 1) * cr + nn - 1`, the smallest uncompressed kv for which every `next_n` row has `n_r >= K + 1`, so no generated row can bypass refcheck's `n <= K` skip. When `N / 8 <= K + 1` the `short` scenario is labelled degenerate with a `[NOTE]` line instead of silently producing non-short rows.

benchmarks/bench_topk_varlen_gvr2.py
```diff
@@ -0,0 +1,402 @@
...
+# optional trtllm_gvr2 port-parity twin. No path ships as a default — pass
+# --trtllm-path or set TRTLLM_TOPK_DIR.
+_DEFAULT_TRTLLM_TOPK = os.environ.get("TRTLLM_TOPK_DIR", "")
+
+
+def parse_args():
+    p = argparse.ArgumentParser(description=__doc__)
+    p.add_argument("--batch-sizes", type=str, default="1,4,16,32,64,128,256")
+    p.add_argument("--n-vals", type=str, default="4096,8192,16384,32768,65536,131072")
+    p.add_argument("--top-k", type=int, default=1024)
```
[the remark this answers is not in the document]
dhiraj113: Fixed in ea069804. GVR-family backends (`fi_gvr2`, `fi_gvr`, `fi_gvr_nolb`, `trtllm_gvr2`) now return `(None, None)` for `top_k` outside {512, 1024, 2048}, so the cell reads `n/a` instead of an exception, and the summary / per-config winner sections are computed relative to `fi_gvr2` when present and otherwise to the first listed backend, so half-precision runs (which drop the fp32-only backends) still get their summary.

tests/topk_varlen/test_topk_varlen_gvr2.py
```diff
@@ -0,0 +1,785 @@
...
+    for r in range(batch):
+        pos = torch.randperm(n_valid, generator=gen, device=_DEV)[: top_k // 4]
+        logits[r, pos] = float("inf")
+    seq_lens = torch.full((batch,), n_valid, dtype=torch.int32, device=_DEV)
+    pre_idx = torch.randint(
+        0, n_valid, (batch, top_k), generator=gen, device=_DEV, dtype=torch.int32
+    )
+    indices, _ = _run_gvr2(logits, seq_lens, top_k, pre_idx)
+    got_inf = torch.isinf(logits.gather(1, indices.long().clamp_min(0)))
+    assert int(got_inf.sum()) == batch * (top_k // 4), "+inf logits were dropped"
```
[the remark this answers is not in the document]
dhiraj113: Not changed in this PR, deliberately, and tracked upstream. The `+inf` behaviour is reproduced bit-identically by the TRT-LLM kernels this backend ports (issue filed with the kernel owners on 2026-09-02); a fix has to land in the shared kernel source rather than in this port, otherwise the two implementations diverge. Routing around it in `auto` would need a scan of the logits for `+inf` on every call, which is a device round trip / extra kernel on a 4-5 µs path, and `auto` only prefers `gvr_2` for the hinted fp32 decode case (indexer scores are finite; masking uses `-inf`, which is exact). The caveat is stated in the public docstring and the `xfail` keeps it visible in CI rather than hidden. I will follow up with the kernel fix when it lands upstream.

benchmarks/routines/topk_varlen.py
```diff
@@ -143,6 +146,16 @@ def testTopKVarlen(args):
                 f"[WARNING] {backend} for routine {args.routine} is not supported "
                 f"on compute capability {major}.{minor}. Skipping."
             )
+    # gvr_2's CC check passes on sm_100/103, but the backend is fp32-only with
+    # a fixed top_k domain — drop it up front instead of failing the call.
+    if "gvr_2" in backends and (
+        args.input_dtype != "float32" or top_k not in (512, 1024, 2048)
+    ):
```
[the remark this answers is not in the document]
dhiraj113: Done in ea069804: the filter also drops `gvr_2` when `max_seq_len % 4 != 0` and the warning names all three conditions.

flashinfer/topk_varlen/kernels/gvr2_topk_host.py
```diff
@@ -0,0 +1,1620 @@
...
+    import os
+
+    env = os.environ.get("CUTE_DSL_ARCH")
+    if env:
+        return env
+    try:
+        major, minor = torch.cuda.get_device_capability()
+        return f"sm{major}{minor}"
+    except Exception:  # noqa: BLE001 — no GPU context: single-arch fallback
+        return "unknown"
```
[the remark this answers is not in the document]
dhiraj113: Fixed in ea069804. Agreed: `_arch_token()`, the DSL compile target (`_compile_arch_token`) and the launch stream all follow the *current* device. Rather than thread a device through the two caches only (the compile target and launch would still be wrong), `run`, `run_ws` and `run_varlen` now re-enter themselves under `torch.cuda.device(logits.device)` whenever the logits are not on the current device, so cache key, compile target and launch agree by construction. Cost on the common path is one device-index compare.

The regression test then exposed a second layer: the DSL detects its compile target once per process from **device 0** (`detect_gpu_arch` → `get_compute_capability_major_minor()` with the default `device_id=0`) unless `CUTE_DSL_ARCH` is set or the compile options carry `--gpu-arch`. On a host whose device 0 is an L40S, every gvr_2 compile was therefore built for sm_89 while the cache labelled it sm_100a, and the launch on the B100 failed with `cudaErrorNoKernelImageForDevice`. Every gvr_2 `cute.compile` now passes `--gpu-arch` for the current (pinned) device; `CUTE_DSL_ARCH` still wins when set. Regression test `test_gvr2_logits_on_non_current_device` runs the public API with logits on a gvr_2-capable device that is not the current one (device 0 = L40S, device 1 = B100) and asserts exactness plus that the current device is restored; it failed before both fixes and passes now.

dhiraj113:
@flashinfer-bot run

dhiraj113:
CI status for `ea069804`:

* **Internal pipeline** [#65822179](https://nv/flashinfer-ci/-/pipelines/65822179) on `tests/topk_varlen`: 16/16 executed test jobs passed.
* **Public CI** (bot-triggered run [33616621108](https://github.com/flashinfer-ai/flashinfer/actions/runs/33616621108)): every job green except **JIT Unittest (H100)**, which fails on exactly one node out of 489,285: `tests/moe/test_unified_moe_activation_matrix.py::test_documented_activation_matrix_matches_runner_registry` ("docs/design_docs/flashinfer_moe_api.md is stale"). That is pre-existing on `main`: today's main run [33615254406](https://github.com/flashinfer-ai/flashinfer/actions/runs/33615254406) fails on the identical node and nothing else. All four `tests/topk_varlen/*` files pass on the H100 (441 source files passed, 1 failed = that MoE doc check).
* The **"Test Results Summary: failure"** on the push-triggered PR Test run is the usual artefact of the push run skipping Setup/matrix (nothing to summarize); the authorized bot-triggered run above is the real one.

dhiraj113:
@flashinfer-bot run

dhiraj113:
@flashinfer-bot run

flashinfer/topk_varlen/kernels/gvr2_topk_host.py
```diff
@@ -0,0 +1,1661 @@
...
+    d = logits.get_device()
+    if not 0 <= d < _GVR_MAX_DEV:
+        raise RuntimeError(f"device index out of range: {d}")
+    if workspace is not None:
+        # multi-stream escape hatch (run_ws parity): concurrent varlen
+        # launches on one device must not share the SPLIT publish slab
+        validate_run_ws(workspace, logits)
+        ws = kernel_view(workspace)
+    else:
+        ws = _ws_hot.get(d)
```
kahyunnam, line 1404: This cache only keys per device, but might run into races across CUDA streams. Shoul dwe cache per (device, CDUA stream)

flashinfer/topk_varlen/topk_varlen.py
```diff
@@ -214,22 +274,100 @@ def _top_k_varlen_heuristic(
...
+    else:
+        rf_first = (
+            (batch <= 16 and 32768 <= n_cols <= 131072)
+            or (batch >= 64 and n_cols >= 131072)
+            or (batch >= 256 and n_cols <= 8192)
+        )
+        gvr_first = batch >= 256 and 32768 <= n_cols <= 524288
+    cutlass_first = fp32 and n_cols >= 65536 and elems >= (1 << 23)
+
+    order = ["gvr_2"]
```
kahyunnam, line 396: Can we either keep `gvr_2` out of the `auto` ranking (explicit-only) until the upstream kernel selects `+inf` exactly, or make that precondition explicit in the API contract rather than a footnote?

It seems `gvr_2` returns a top-K containing none of the row's `+inf` values, and the new `auto` makes that the default. If you call `flashinfer.top_k_varlen` on fp32 logits with a `pre_idx` hint and any row contains a literal `+inf`, the default `backend="auto"` now returns a top-K that omits *every one* of those `+inf` values (they would be the largest elements in the row, and they are silently replaced by finite ones). On `main` this cannot happen, because `auto` had no way to reach this kernel. The `gvr` and `radix` backends return the correct answer on the identical input.

flashinfer/topk_varlen/topk_varlen.py
```diff
@@ -199,6 +211,54 @@ def _gvr_top_k_varlen_check(
...
+    if not (_cute_dsl_ready(logits.device) and pre_idx is not None):
+        return False
+    # fp32 only: the upstream self-sampling kernels declare bf16/fp16 a
+    # follow-up (run_varlen raises on any other dtype).
+    if logits.dtype != torch.float32:
+        return False
+    # Validated K domain (matches the upstream dispatch tuning + dsa.py gate).
+    if top_k not in (512, 1024, 2048):
+        return False
+    if pre_idx.shape[1] != top_k:
```
kahyunnam: Should we also add the contiguity and `pre_idx.shape[0]` checks to the checker so `auto` demotes these configs to `gvr`/`radix` instead of raising, and reject an out-of-contract `out_indices` width rather than flat-packing it?

kahyunnam (commented):
Overall LGTM, few comments below.

dhiraj113:
@flashinfer-bot run

dhiraj113:
@flashinfer-bot run

flashinfer/topk_varlen/kernels/gvr2_topk_host.py
```diff
@@ -0,0 +1,1661 @@
...
+def _varlen_launcher(num_rows, npad, k, n_env, next_n, cr):
+    """Capture-time varlen plan + compiled launcher.  The gvr_main port is
+    the universally correct fallback; specialist family tiers below.  Every
+    choice here is a function of capture-stable quantities only — mirroring
+    the in-tree runner's pick_tuning(graph_capture=...) discipline."""
+    key = (num_rows, npad, k, n_env, next_n, cr, _arch_token())
+    hit = _VARLEN_CACHE.get(key)
+    if hit is not None:
+        return hit
+    n_eff = max(min(n_env, npad), k + 1)
```
leejnau: n_eff can exceed the physical row width here. With logits.shape == (2, 256), k == 512, and oversized kv_lens, it becomes 513; the register prologue then reads 513 elements using a 256-element row stride, crossing into the next row and beyond storage. Please keep the device read bound ≤ the logical/physical width, using a separate inflated value for routing if necessary.  Possibly add this case as a regression test if needed.

flashinfer/topk_varlen/kernels/gvr_topk_decode.py
```diff
@@ -1289,29 +1503,68 @@ def phase3_collect_candidates(
...
             if tidx == 0:
-                if s_iscalars[0] > cutlass.Int32(kCC):
-                    s_thr[1] = s_thr[0]  # val_lo = threshold
+                c0 = s_iscalars[0]
+                if c0 > cutlass.Int32(kCC):
+                    s_thr[1] = s_thr[0]
+                    s_thr[2] = cutlass.Float32(self.FLT_MAX)
+                elif c0 < cutlass.Int32(kK):
+                    s_thr[2] = s_thr[0]
+                    s_thr[1] = cutlass.Float32(self.NEG_FLT_MAX)
```
leejnau: -FLT_MAX is not a valid lower search bound when logits may contain literal -inf. For an all--inf row with N > K, count(>= -FLT_MAX) == 0, contradicting the invariant below that val_lo admits at least K candidates and candidate collection remains empty and the plateau path can leave output slots unwritten. This reproduces with both GVR load-balancing modes. Please use actual -inf or explicitly handle this boundary, possibly with an all -inf regression test for both modes.

flashinfer/topk_varlen/topk_varlen.py
leejnau: workspace=None is not safe for arbitrary concurrency: GVR2 resolves it to one mutable slab per device, so concurrent SPLIT launches or graph replays share the counters, offsets, and candidate buffer while asynchronously restoring them. Please either make the default workspace stream/graph scoped or state that every concurrently active stream or graph must provide a distinct gvr2_workspace. This matters for SGLang-style multi-stream serving.

flashinfer/topk_varlen/kernels/gvr2_topk_host.py
```diff
@@ -0,0 +1,1661 @@
...
+    n_env = max(1, int(max_seq_len) // int(compress_ratio))
+    if row_stride is None:
+        npad = (n_env + 63) // 64 * 64
+    else:
+        npad = int(row_stride)
+        if npad < n_env or npad % 4:
+            raise RuntimeError(
+                f"row_stride must be a float4-multiple >= n_env={n_env}, got {row_stride}"
+            )
+    key = (dev, int(top_k), int(max_seq_len), int(compress_ratio), nn, tuple(rows_list), npad)
```
leejnau, line 1660: This completion key omits the exact req_rows. For (K=512, N=8192), warming (64,) and then (63, 64) produces the same representative rows_list == [1], so the second call returns here before _varlen_launcher(63, ...) is populated below. Capturing row count 63 then raises the launcher-cache-miss error. Please include req_rows in the done key or populate exact requested launchers before this early return, maybe with a sequential-warmup regression test.

flashinfer/topk_varlen/kernels/gvr2_topk_host.py
```diff
@@ -0,0 +1,1661 @@
...
+        if not (pre_idx.is_contiguous() and indices.is_contiguous() and kv_lens.is_contiguous()):
+            raise RuntimeError("pre_idx/indices/kv_lens must be contiguous")
+        # logits: accept row-major views with a wider row stride (the DSL
+        # paged-MQA logits arena is 256-aligned and column-sliced — a legal
+        # NON-contiguous view). The kernel only needs (base, row stride):
+        # widen back to a compact [rows, stride] view over the same storage;
+        # the tail columns are never classified (per-row n gates all reads).
+        if logits.stride(1) != 1:
+            raise RuntimeError("logits inner stride must be 1")
+        npad = logits.stride(0) if num_rows > 1 else logits.shape[1]
```
leejnau, line 1450: This assumes stride(0) is at least the logical row width, but overlapping layouts are currently accepted. For example, base.expand(2, -1) has stride (0, 1), passes the checker, and makes npad=0, changing the top-k search domain and potentially driving invalid reads. Please reject stride(0) < logits.shape[1] for multi-row inputs here and in _gvr2_top_k_varlen_check, or explicitly support such layouts.

flashinfer/topk_varlen/kernels/gvr2_topk_host.py
```diff
@@ -0,0 +1,1661 @@
...
+
+def validate_run_ws(workspace: torch.Tensor, logits: torch.Tensor) -> None:
+    """run_ws() workspace hardening, in a fixed predicate order:
+    CUDA + same device as logits; numel*element_size >= workspace_bytes();
+    base 8-byte aligned."""
+    if not (workspace.is_cuda and workspace.get_device() == logits.get_device()):
+        raise RuntimeError("workspace must be a CUDA tensor on the same device")
+    if workspace.numel() * workspace.element_size() < WS_BYTES:
+        raise RuntimeError(f"workspace too small: need {WS_BYTES} bytes")
+    if workspace.data_ptr() & 7:
```
leejnau: This accepts an 8-byte-aligned workspace, but the compiled ws_fake ABI assumes 16-byte alignment; kernel_view even notes that an 8-but-not-16-byte pointer will fail at launch.  Please require 16-byte alignment here and update the public documentation. pre_idx and indices likewise have 16-byte alignment assumptions in every compiled family but are checked only for contiguity, so shifted contiguous views need equivalent validation or a weaker compiled assumption.

flashinfer/topk_varlen/topk_varlen.py
dhiraj113: the `inf` issue has been addressed here - https://gitlab-master.nvidia.com/dlarch-fastkernels/dynamic-kernel-generator/-/issues/58. No unkown functional issues remain for gvr_v2.

dhiraj113: Fixed on this branch rather than routed around: 48572ef2 ports TensorRT-LLM #18625 for the register family and 274fad3e its second commit for the clustered-register family. `+inf` inside or outside the bracket sample, K in {512, 1024, 2048}, more `+inf` than K and the 4 x 32K / 4 x 64K reg_clus shapes are exact on B200, B100 and Rubin (`test_gvr2_posinf_completeness`, `test_gvr2_posinf_reg_clus`); the docstring caveat is removed. NaN ordering remains implementation-specific upstream, although NaN ranks on top in every shape tested.

tests/topk_varlen/test_topk_varlen_gvr2.py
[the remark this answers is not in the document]
dhiraj113: Fixed on this branch rather than routed around: 48572ef2 ports TensorRT-LLM #18625 for the register family and 274fad3e its second commit for the clustered-register family. `+inf` inside or outside the bracket sample, K in {512, 1024, 2048}, more `+inf` than K and the 4 x 32K / 4 x 64K reg_clus shapes are exact on B200, B100 and Rubin (`test_gvr2_posinf_completeness`, `test_gvr2_posinf_reg_clus`); the docstring caveat is removed. NaN ordering remains implementation-specific upstream, although NaN ranks on top in every shape tested.

flashinfer/topk_varlen/kernels/gvr2_topk_host.py
dhiraj113: Documented rather than re-keyed for now: only the multi-CTA streaming path touches the slab, and a per-stream default has two traps I did not want to ship blind (a slab first touched under CUDA-graph capture would come from the graph pool with its memset only recorded, and graphs captured on the default capture stream would still share one slab). The docstring in 89128c57 states the contract: concurrently active streams and concurrently replayed graphs pass their own `gvr2_workspace`. The new serving harness (3f208c45) pins it: four graphs of eight SPLIT launches replayed concurrently are exact with private workspaces, and the shared-default variant is a non-strict xfail that reproduces the race (17/5120 rows on B100). A per-stream default keyed on the current stream can follow if eager multi-stream serving needs it.

flashinfer/topk_varlen/topk_varlen.py
dhiraj113: Agreed; the sentence described the gvr buffers. Rewritten in 89128c57: one slab per device for gvr_2, every concurrently active stream or concurrently replayed graph must pass its own workspace, stream-ordered use needs nothing, and the condition is stated structurally (the streaming path is a routing decision) rather than as a column threshold. The concurrent-graph-replay test in 3f208c45 exercises both sides of that contract.

flashinfer/topk_varlen/topk_varlen.py
dhiraj113: Done in 89128c57, with one design choice: a malformed hint (shape, dtype, device, contiguity or alignment) is dropped with a RuntimeWarning and the call runs hint-free, rather than being demoted silently. gvr fails the same way as gvr_2 on those hints (measured: DSL stride / dtype / alignment / batch errors), so demotion would only ever have reached the radix family, and the warning keeps the perf cliff visible. Explicit gvr / gvr_2 with a malformed hint are refused up front, also under skip_check. Output buffers must be [num_rows, top_k] or a 1-D buffer with exactly that many elements (viewed in place, which the radix_filter in-place test relies on); wider, taller or transposed-shape buffers raise instead of being flat-packed. One test case per variant.

flashinfer/topk_varlen/kernels/gvr2_topk_host.py
dhiraj113: Fixed in 16d51848 as you describe: the kernels get min(n_env, npad) and only route() sees max(., k+1). Your (2, 256) / k=512 / oversized kv case returns the identity + -1 tail now; before the fix compute-sanitizer showed 37 invalid 16-byte global reads from the second CTA. Regression test with oversized, equal and in-range seq_lens and a poisoned neighbour row. Reported upstream (DKG #60) since the line is shared.

flashinfer/topk_varlen/kernels/gvr_topk_decode.py
dhiraj113: Confirmed for both load-balance modes, and wider than the all -inf row: any row with fewer than K values above -FLT_MAX lost slots (10 finite + -inf tail left 1014 unwritten). Anchor is -inf in 85521aa0; the +FLT_MAX side was checked with > K and > kC +inf values and is fine. Tests for both modes: three all -inf shapes, a mixed batch, the partial row. The same commit also clamps the per-row length to the logits width, which a new serving-pattern harness (3f208c45) found gvr and radix_filter (51c85242) did not do (reads past the row when a request's length outgrows the score buffer).

flashinfer/topk_varlen/kernels/gvr2_topk_host.py
dhiraj113: Fixed in 16d51848: the exact-row launcher population also runs on the early-return path, so the done key keeps gating only the band launches and the upstream block stays byte-identical. Your [64] then [63, 64] sequence followed by a capture is the regression test (with 61 rows so the precondition is checkable). Reported upstream (DKG #60).

flashinfer/topk_varlen/kernels/gvr2_topk_host.py
dhiraj113: Rejected in both places in 16d51848 (run_varlen raises) and 89128c57 (checker returns False so auto skips gvr_2). Measured on the other backends first: gvr and radix_filter handle an expand view correctly, radix and radix_cutlass raise their own errors, so their checkers now gate it too and auto lands on a backend that runs (the test asserts an exact auto result on the view).

flashinfer/topk_varlen/kernels/gvr2_topk_host.py
dhiraj113: validate_run_ws now requires 16 bytes and the docs say so; run_varlen checks pre_idx / indices for the same alignment with a FlashInfer message (16d51848). At the API a shifted pre_idx is treated as a malformed hint (dropped with a warning) and a shifted out buffer is a ValueError (89128c57), so neither reaches the DSL error. The kernels access pre_idx / indices with scalar loads and stores (only logits use 128-bit loads), so the 16-byte declaration is not load-bearing; I asked upstream (DKG #60) whether the fakes can declare 4, which would make shifted views simply work.

dhiraj113:
@flashinfer-bot run

leejnau (approved):

kahyunnam (approved):

dhiraj113:
@flashinfer-bot run

dhiraj113:
@flashinfer-bot run

dhiraj113:
Heads-up: the two commits pushed to this branch after the squash-merge (117415eb, f15221e5: the 4K-8K dispatch rungs, hint-free `gvr_2`, SM-count-aware register band) never landed; they are re-based as #4986. Section 8 of this description belongs to that PR. The CI comments above from today (after the merge) can be ignored.

PR #4813, opened 2026-08-28
fix(topk): repair GVR's non-converged threshold search (port of TRT-LLM #18094)

dhiraj113:
# fix(topk): repair GVR's non-converged threshold search (port of TRT-LLM #18094)

## The bug

FlashInfer's `gvr` backend of `top_k_varlen` was ported from TRT-LLM's V1 kernel at its ~Jul 22, 2026 state and missed two upstream hardening layers. The Phase-2 secant threshold search can terminate **without** a threshold whose above-count lands in the `[K, kC]` acceptance window; the pre-fix code then shipped a silently wrong top-K:

- **degenerate hint bracket** (all `pre_idx` values identical / out of range) → emitted identity indices `row[0:K]` verbatim;
- **non-convergence** (hostile hints whose bracket excludes any valid threshold; tie plateaus wider than the candidate buffer; rows with `N_eff = K + 1` where the acceptance window is a knife-edge) → an **underfilled row** whose untouched output slots keep stale/−1 garbage.

Found while benchmarking the new `gvr_2` backend (#4811): the short-row sweep configs (B ≥ 64, N=8192, K=1024, most rows at `N_eff = 1025`) returned out-of-range indices, identically for the LB and non-LB paths. Deterministic reproducer included as a regression test.

## The fix (correctness-relevant subset of three upstream commits, current upstream tip for this kernel)

- **NVIDIA/TensorRT-LLM#16457 (Jul 28) + #16877 (Aug 13) — the tie-plateau layer**: adjacent-float bracket terminal (`done=3`, threshold = sure-winner side) and a budget-exhausted plateau-collapse bisection in *both* `phase2_secant_search` copies (redundant-warp and leader); `s_iscalars` grows `(6,) → (8,)` with `[6]` = plateau flag captured before Phase 4 and `[7]` = fill ticket; Branch-C's −1 pad is gated off under the plateau terminal, and a post-Phase-4 fill completes the row from the bitwise-equal tie class (any `(K − count)`-subset is a valid tie-aware completion).
- **NVIDIA/TensorRT-LLM#18094 (Aug 25) — the two-sided repair**: Phase 3's retry loop (previously overflow-only, with an arithmetic-midpoint step that stalls on adjacent floats) becomes an anchored bisection on the signed fp32 **order-key** image (provable collapse in ≤ 32 steps, budget 48): undershoot anchors the untested bracket end at a float extreme, a collapse ending under K restores to `val_lo` (which admits ≥ K by construction), and collapsed tie plateaus hand off to the `done=3` machinery. The degenerate-hint identity emit is replaced by a synthetic-bracket fall-through — correctness no longer depends on the hint at all.

Both LB paths are covered automatically (`GvrTopKLBKernel` reuses `GvrTopKKernel.run_one_row`). Converging rows — the overwhelmingly common case — pay only a flag capture plus one barrier before Phase 4.

Verified against live upstream: `496a002efe` (#18094) is the newest commit touching this kernel on TRT-LLM `main`, so this port matches the current upstream tip. Intentionally not ported: upstream's R0 histogram-ladder admission and tiered dispatch (perf-only machinery FlashInfer's copy never had).

## Testing

- New `tests/topk_varlen/test_gvr_threshold_repair.py` (31 tests) ports upstream #18094's regression patterns onto the FlashInfer API — hostile `bottom_k`/`uniform`/`random` hints × K ∈ {512, 1024, 2048} × LB modes, ReLU-sparse tie plateaus (n_pos ∈ {3, 100, 1000} over an exact-0.0 plateau), MTP hostile hints with a mod-cr boundary — plus the FlashInfer-found `N_eff = K+1` batch case. **31/31 pass** (all fail on the pre-fix kernel).
- Full `topk_varlen` suite: **146/146** on B200 (SM100). SM80/SM120 sweeps unaffected (48 passed / 98 skipped each; `gvr` is sm_100/103-gated).
- Port reviewed with a 3-lens adversarial pass against the upstream reference (hunk-for-hunk port fidelity; DSL/barrier-execution correctness incl. the redundant-warp parity double-buffering and the new atomicAdd ticket; FlashInfer-divergence interactions incl. cluster lockstep, LB's dual traced instances, and the SMEM layout growth) — all lenses clean. Two inherited upstream-parity notes for reviewers: rows violating the finite-logits contract (> N−K entries at −inf) combined with a degenerate hint can still leave trailing slots unwritten (identical upstream), and genuine `done=3` rows pay a redundant Phase-3 re-bisection (identical upstream; a `done != 3` gate is a possible upstream-worthy follow-up).

Perf before/after comparison of the gvr backend is being measured and will be posted as a comment.

AI-assisted (Claude Code).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved top-K result accuracy for rows with duplicate values, sparse plateaus, short inputs, and challenging threshold-search conditions.
  * Prevented incomplete results and stale or invalid entries when processing edge-case inputs.
  * Improved recovery from degenerate search hints and non-converged threshold searches.

* **Tests**
  * Added regression coverage for tie-heavy inputs, short rows, varying sequence lengths, and near-minimum top-K sizes.

### DIFF
--- a/flashinfer/topk_varlen/kernels/gvr_topk_decode.py
+++ b/flashinfer/topk_varlen/kernels/gvr_topk_decode.py
@@ -127,6 +127,49 @@ def float_as_uint32(float_val):
     return llvm.bitcast(cutlass.Uint32.mlir_type, float_val.ir_value())
 
 
+def float_as_int32(float_val):
+    """Interpret FP32 value as int32 bit pattern (cuTe DSL bit-cast)."""
+    return cutlass.Int32(llvm.bitcast(cutlass.Int32.mlir_type, float_val.ir_value()))
+
+
+def f32_order_key(float_val):
+    """Order-preserving fp32 -> int32 key (unsigned-monotonic bit pattern).
+
+    ``s ^ ((s >> 31) | 0x80000000)``: positive floats map to
+    ``bits | 0x80000000``, negative floats to ``~bits`` — the standard radix
+    transform whose UNSIGNED order equals fp32 order (NaN-free inputs). The
+    returned Int32 must only be consumed digit-wise (``(k >> s) & 0xFF``) or
+    via equality / prefix-equality; for a full ordered compare, flip the top
+    bit first (``k ^ 0x80000000`` is signed-monotonic).
+    """
+    s = float_as_int32(float_val)
+    return s ^ ((s >> cutlass.Int32(31)) | cutlass.Int32(-2147483648))
+
+
+def f32_order_key_signed(float_val):
+    """Signed-monotonic Int32 key (fp32 order == signed Int32 order), so the
+    Phase-3 repair can bisect with provable collapse in <= 32 steps."""
+    return f32_order_key(float_val) ^ cutlass.Int32(-2147483648)
+
+
+def order_key_signed_to_f32(m):
+    """Branchless inverse of :func:`f32_order_key_signed` (non-NaN keys)."""
+    k = m ^ cutlass.Int32(-2147483648)
+    top = k >> cutlass.Int32(31)  # -1 if top bit set, else 0
+    mask = cutlass.Int32(-2147483648) | (~top & cutlass.Int32(2147483647))
+    s = k ^ mask
+    return cutlass.Float32(llvm.bitcast(cutlass.Float32.mlir_type, s.ir_value()))
+
+
+def order_key_mid_f32(v_lo, v_hi):
+    """Ordered-key midpoint; returns (mid_float, is_adjacent) where adjacency
+    means no float strictly between v_lo and v_hi exists."""
+    m_lo = f32_order_key_signed(v_lo)
+    m_hi = f32_order_key_signed(v_hi)
+    m_mid = (m_lo & m_hi) + ((m_lo ^ m_hi) >> cutlass.Int32(1))
+    return order_key_signed_to_f32(m_mid), m_mid == m_lo
+
+
 def _fmin_f32_inline(a, b):
     """Single PTX ``min.f32`` → one SASS FMNMX.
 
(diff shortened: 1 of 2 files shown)


dhiraj113:
@flashinfer-bot run

dhiraj113:
## Perf before/after (B200, fp32, CUDA-graph timing, identical seeds, both LB modes)

Two config classes, 58 timed cells total (`ratio = fixed / pre-fix`; < 1 ⇒ fix is faster):

**Converging class** — uniform + mixed lengths, K=1024, B ∈ {1,16,64,256} × N ∈ {8k,32k,128k} (48 cells; the production-representative case):

| | geomean | range |
|---|---|---|
| fixed vs pre-fix | **0.957× (≈4% faster)** | 0.83× – 1.02× |

The fix is a mild *speedup* here, not a cost: on ragged rows the old search frequently ended non-converged with an oversized candidate set (its give-up fallback accepted counts up to 2·kC), which Phase 4 then processed; the repair lands the count inside [K, kC], shrinking Phase-4 work. Uniform-length non-LB cells are exactly neutral (1.00–1.02×), confirming the added flag-capture + barrier is noise-level.

**Repair class** — the configs the old kernel got wrong (short-row batches with N_eff = K+1, degenerate/hostile hints, ReLU tie plateaus; 10 cells):

| config | pre-fix | fixed | note |
|---|---|---|---|
| short B∈{64,256} (4 cells) | 13–35 µs **WRONG** | 14–36 µs OK | 0.94–1.18× |
| degenerate hint N=64k (2 cells) | 9–14 µs **WRONG** (identity emit) | 63–67 µs OK | the old speed was the price of not computing the answer |
| ReLU plateau K=2048 (2 cells) | 33–45 µs **WRONG** (−1-padded) | 77–101 µs OK | plateau fill + inherited re-bisection |
| short B=16 (2 cells, was already correct) | 13–21 µs OK | 13–19 µs OK | 0.90–0.97× |

8 of 10 pre-fix cells produced incorrect output, so their timings weren't a valid baseline. For workloads living in these regimes, `gvr_2` (#4811) remains the perf answer; this PR makes `gvr` *correct* there.

dhiraj113:
@flashinfer-bot rerun

dhiraj113:
Folding this into #4811 per maintainer-of-one preference — the fix commit is now cherry-picked onto that branch (same content, clean pick; regression suite re-verified there). Closing; the review artifacts in this thread (root-cause analysis, before/after perf table) remain the reference.

PR #4815, opened 2026-08-29
Replayssm prefix materialize

hnover-nv:
## 📌 Description

For ReplaySSM, instead of keeping an up-to-date SSM state, we instead keep a historical state and store the inputs since in a buffer.  When the buffer fills up, we "flush" it into an updated state. See checkpointing_ssu.cu and related files in this repo.

Because prefill is not aware of the ReplaySSM buffer, for prefix caching to work we need to materialize the state at arbitrary tokens.  This shouldn't be done as part of checkpointing_ssu.cu because that can't output an updated state until the next step, and we want to do it at the end of the current step, after we know how many tokens were accepted (for the MTP case).

Note we don't apply all the accepted tokens. For example, we may save states that are multiples of 512 tokens, so if our last state is materialized state is at 508 and our buffer represents another 7, then we only want to apply 4 of the tokens in our state update.  And that is independent of whether the 7 came from accepting 7 this step, or accepting fewer over the course of a few steps since our last checkpointing.

This PR provides the necessary kernel and wrapping to allow that state materialization.  The main differences from checkpointing_ssu are:
 - No new input or output tokens, we just use the old state, the ring buffer and its metadata.
 - Does not update any of those inputs, rather it writes out new state to a provided location.
 - Applies a user-given # of tokens, rather than all tokens in the buffer.
 - Runs on all layers at once, to minimize launches.  This means most of our inputs are actually per-layer tables of pointers, slots, etc.
 - Persistent kernels (see next).

We do make extensive use of the helper function's in checkpointing_ssu's nicely modular code, so not much needed writing.  This probably comes with some inefficiencies (helper functions naturally want to process input or produce output tokens), but we hope dead code elimination mostly covers it.  And as this runs infrequently a few % inefficient is not a big deal.

One unexpected wrinkle was that to avoid host/device sync, we expect to normally be invoked for all requests in the latest batch, with -1 as the flag for not needing to materialize.  For large batches and large models like Nemotron v3 Ultra, a naive kernel approach would then spend over a millisecond every forward pass just to launch nearly empty CTAs that do nothing, which would be a material loss.  Instead we adopt persistent kernels and require the user to pass in an additional batch-sized tensor that is the indices in the batch of requests that need flushing, and -1 thereafter.  This lets us minimize the no-write case to 2 microseconds on a B300, even for nemo ultra v3 TP=1 (48 layers, 256 heads/layer, mamba-2 dim 64x128) at batch size 256.  With that setup, each actual request whose state is written costs about 70-80 additional micros.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Reviewer Notes

This is my first FlashInfer PR, so I may have messed up some plumbing, missed some typically updated file.  For example I missed the rst the first time around.  I attempted to have my agent audit for gaps though.

AI (codex) was used, but I have reviewed every line.

## Summary by CodeRabbit

- **New Features**
  - Added `replayssm_materialize` to the public Mamba API for efficiently materializing ReplaySSM prefix states.
  - Supports active-request selection, multilayer and grouped-head configurations, zero-length prefixes, and persistent GPU execution.
  - Supports BF16, FP16, INT8, and FP8 state workflows, including optional stochastic replay and state scaling.
  - Added optional dependency anchors for improved `torch.compile` integration.

- **Documentation**
  - Added `replayssm_materialize` to the Mamba API documentation.

- **Benchmarks**
  - Added configurable performance benchmarking with CUDA event, CUPTI, and CUDA Graph timing modes.


csrc/replayssm_materialize_customize_config.jinja
```diff
@@ -0,0 +1,16 @@
+#pragma once
+#include <cuda_bf16.h>
```
ishovkun, line 2: Are you sure about these includes here? I don't think that's the right place

csrc/replayssm_materialize.cu
```diff
@@ -0,0 +1,445 @@
...
+      static_cast<const int64_t*>(scale_slot_strides.data_ptr()),
+      static_cast<const int32_t*>(src_slots.data_ptr()),
+      static_cast<const int32_t*>(dst_slots.data_ptr()),
+      static_cast<const int32_t*>(ring_start.data_ptr()),
+      static_cast<const int32_t*>(replay_prefix_len.data_ptr()),
+      static_cast<const int32_t*>(active_request_indices.data_ptr()),
+      rand_seed.has_value() ? static_cast<const int64_t*>(rand_seed.value().data_ptr()) : nullptr,
+      int(replay_prefix_len.size(0)),
+      int(num_layers),
+      int(num_heads),
```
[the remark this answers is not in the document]
hnover-nv: num_heads is a model parameter that would require the caller to allocate layers * num_heads * 64*128 * state.dtype size amount of memory to be used correclty, so it cannot be anywhere near int max.  even int32 is  incredibly unlikely.  tyipcal values are well below 512.

csrc/replayssm_materialize.cu
```diff
@@ -0,0 +1,445 @@
...
+  constexpr int kThreadsPerCta = warpSize * 4;
+  int device = 0, sm_count = 0, blocks_per_sm = 0;
+  FLASHINFER_CUDA_CHECK(cudaGetDevice(&device));
+  FLASHINFER_CUDA_CHECK(cudaDeviceGetAttribute(&sm_count, cudaDevAttrMultiProcessorCount, device));
+  FLASHINFER_CUDA_CHECK(cudaOccupancyMaxActiveBlocksPerMultiprocessor(
+      &blocks_per_sm, materialize_replay_kernel<state_t>, kThreadsPerCta, smem));
+  int64_t const persistent_ctas = int64_t(sm_count) * blocks_per_sm;
+  FLASHINFER_CHECK(persistent_ctas > 0, "ReplaySSM materialize has no resident CTA configuration");
+  dim3 grid(work_items < persistent_ctas ? int(work_items) : int(persistent_ctas));
+  if (std::getenv("FLASHINFER_REPLAYSSM_MATERIALIZE_VERBOSE")) {
```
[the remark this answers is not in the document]
hnover-nv: removed

flashinfer/mamba/replayssm_materialize.py
```diff
@@ -0,0 +1,217 @@
...
+        dim,
+        dstate,
+        heads_per_group,
+        max_window,
+        philox_rounds,
+    ).build_and_load()
+
+
+@flashinfer_api
+def replayssm_materialize(
```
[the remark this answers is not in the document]
hnover-nv: fixed

flashinfer/jit/mamba/replayssm_materialize.py
```diff
@@ -0,0 +1,81 @@
...
+import os
+import jinja2
+import torch
+from ...compilation_context import CompilationContext
+from .. import env as jit_env
+from ..core import JitSpec, gen_jit_spec
+from ..utils import write_if_different
+
+_DTYPE = {
+    torch.float16: "half",
```
ishovkun, line 27: **[Medium, high confidence] FP16 input caches are silently consumed as BF16**

The JIT dtype map accepts `torch.float16` for `input_dtype`, and the public wrapper does not narrow that contract. The reused replay helper only checks that `input_t` is two bytes, while the MMA operand type is fixed to `__nv_bfloat16`; its implementation explicitly treats `old_x` as already being that operand type. An FP16 input cache therefore compiles but its bit patterns are interpreted as BF16 during replay. For example, FP16 `1.0` (`0x3c00`) becomes BF16 `0.0078125`, causing silent numerical corruption.

Requested resolution: either restrict `input_dtype` to `torch.bfloat16` in the public wrapper and JIT generator, document the restriction, and add rejection tests, or implement a real FP16 conversion/MMA path. The dtype map should not advertise float32, int8, or FP8 input caches either unless those paths are made valid; the current two-byte assertion rejects them at compile time.

csrc/replayssm_materialize.cu
```diff
@@ -0,0 +1,433 @@
...
+  while (work < work_items) {
+    if (virtual_request != last_virtual_request) {
+      physical_request = p.active_request_indices[virtual_request];
+      if (physical_request < 0) break;
+      count = p.replay_prefix_len[physical_request];
+      last_virtual_request = virtual_request;
+    }
+    int64_t const table = int64_t(layer) * p.batch + physical_request;
+    int const src_slot = p.src_slots[table], dst_slot = p.dst_slots[table];
+    if (src_slot == p.pad_slot_id || dst_slot == p.pad_slot_id || count > MAX_WINDOW) {
```
ishovkun: **[Medium] Define and enforce the replay-prefix bounds**

The kernel silently skips a work item when `replay_prefix_len > MAX_WINDOW`, leaving the destination slot unchanged. The [design document](https://github.com/flashinfer-ai/flashinfer/blob/a73fe28a14469b12c54e49927ca4fab065799efe/docs/design_docs/replayssm_prefix_materialization.md#L22-L28) defines negative, zero, and positive prefix lengths, while the [public API docstring](https://github.com/flashinfer-ai/flashinfer/blob/a73fe28a14469b12c54e49927ca4fab065799efe/flashinfer/mamba/replayssm_materialize.py#L128-L140) separately states a caller precondition tied to `prev_num_accepted_tokens`; neither defines the oversized-prefix outcome.

Requested resolution: decide which contract is intended and encode it in both the docstring and design document:

- If an oversized prefix is an expected integration state, document that the destination remains unchanged and explain when the caller must first run a real checkpoint flush.
- If it is a caller error, surface it through validation or an error/status mechanism instead of silently retaining stale destination state.

Also make the relationship between `replay_prefix_len` and `ring_buffer_len` explicit. A count that is within `MAX_WINDOW` but exceeds the actual ring length is not rejected, while the reused ring-loading helpers rely on valid ring indexing. As defense in depth, an active entry with a negative count should take the documented no-write path even if `active_request_indices` and `replay_prefix_len` disagree.

flashinfer/mamba/replayssm_materialize.py
```diff
@@ -0,0 +1,256 @@
...
+    _module(
+        state_dtype,
+        input_dtype,
+        matrixA_dtype,
+        dim,
+        dstate,
+        heads_per_group,
+        max_window,
+        philox_rounds,
+    ).replayssm_materialize(
```
ishovkun: **[Medium] Register the operation through `torch.library`**

The public wrapper calls the JIT FFI module directly. Peer Mamba operations, including [`checkpointing_ssu`](https://github.com/flashinfer-ai/flashinfer/blob/a73fe28a14469b12c54e49927ca4fab065799efe/flashinfer/mamba/checkpointing_ssu.py#L554-L707) and [`selective_state_update`](https://github.com/flashinfer-ai/flashinfer/blob/a73fe28a14469b12c54e49927ca4fab065799efe/flashinfer/mamba/selective_state_update.py#L414-L511), use `register_custom_op` plus `register_fake_op`. The new every-forward-pass API will otherwise cause a graph break under `torch.compile(fullgraph=True)`.

Requested resolution: add an internal custom-op wrapper and fake implementation following the peer pattern. Because the true destinations are reached through device pointer tables, document which table tensor is used as the mutation/dependency token and why; `mutates_args` cannot describe the indirect destination tensors themselves.

csrc/replayssm_materialize.cu
```diff
@@ -0,0 +1,433 @@
...
+  const int64_t* a_ptrs;
+  const int64_t* scale_ptrs;
+  const int64_t* scale_slot_strides;
+  const int32_t* src_slots;
+  const int32_t* dst_slots;
+  const int32_t* ring_start;
+  const int32_t* replay_prefix_len;
+  const int32_t* active_request_indices;
+  const int64_t* rand_seed;
+  int batch, layers, heads, ring_buffer_len, pad_slot_id;
```
ishovkun: could you make it int32_t for consistency ? :-)

csrc/replayssm_materialize.cu
```diff
@@ -0,0 +1,433 @@
...
+    layer -= layers;
+    ++virtual_request;
+  }
+  virtual_request += virtual_request_delta;
+}
+
+// Derive a per-layer Philox key with the SplitMix64 finalizer (Stafford mix13).
+// The golden-ratio increment and avalanche multipliers are SplitMix64 constants.
+__device__ __forceinline__ int64_t layer_philox_seed(int64_t base_seed, int layer) {
+  uint64_t z = static_cast<uint64_t>(base_seed) ^ (0x9E3779B97F4A7C15ULL * (uint64_t(layer) + 1));
```
ishovkun: move to include/flashinfer/mamba/conversion.cuh

csrc/replayssm_materialize.cu
```diff
@@ -0,0 +1,433 @@
...
+            int64_t(dst_slot) * p.scale_slot_strides[layer] + int64_t(head) * DIM;
+        dst_scales[scale_dst + tid] = scales[scale_src + tid];
+      }
+      advance_persistent_coordinates(work, virtual_request, layer, head, work_stride,
+                                     virtual_request_delta, layer_delta, head_delta, p.layers,
+                                     p.heads);
+      continue;
+    }
+    load_state_per_warp<T, DIM, DSTATE, NUM_WARPS>(smem, state, state_src_base, warp, lane);
+    auto const* x = reinterpret_cast<input_t const*>(p.x_ptrs[layer]);
```
ishovkun: check if you might need __restrict__ here?

csrc/replayssm_materialize.cu
```diff
@@ -0,0 +1,433 @@
...
+      // view's state and scale are used below only for the
+      // destination.  state is already in smem, source_scale is
+      // loaded directly into registers below.
+      view.state = state_dst;
+      view.state_stride_seq = p.state_slot_strides[layer];
+      view.state_scale = reinterpret_cast<void*>(p.scale_ptrs[layer]);
+      view.state_scale_stride_seq = p.scale_slot_strides[layer];
+      auto tiled_mma_chain =
+          cute::make_tiled_mma(cute::MMA_Atom<cute::MMA_Traits<checkpointing::MMA_prop::AtomK16>>{},
+                               cute::Layout<cute::Shape<cute::_4, cute::_1>>{});
```
ishovkun: (4, 1) is the warp layout, right? Might bite you if you leave it hard-coded.

docs/design_docs/replayssm_prefix_materialization.md
```diff
@@ -0,0 +1,89 @@
...
+
+Use a dedicated replay-only driver with the existing inline-template recurrence
+and quantization helpers; do not duplicate the numerical recurrence. Zero
+count uses an explicit raw-copy branch so it cannot consume Philox randomness
+or re-quantize state. Preserve the existing forward specializations: any
+required 8-bit source/destination I/O split is compile-time specialized and
+must be checked for unchanged forward resource use and SASS.
+
+The initial launch is one four-warp CTA per `(request, layer, head)` with
+`D_PER_CTA == DIM`; it has no persistent work queue, D split, or autotuning
```
ishovkun: **[Medium] Bring the design document in sync with the implementation**

The document still uses [`flush_count`](https://github.com/flashinfer-ai/flashinfer/blob/a73fe28a14469b12c54e49927ca4fab065799efe/docs/design_docs/replayssm_prefix_materialization.md#L22-L34), while the public API names it [`replay_prefix_len`](https://github.com/flashinfer-ai/flashinfer/blob/a73fe28a14469b12c54e49927ca4fab065799efe/flashinfer/mamba/replayssm_materialize.py#L60-L64). This section also says the launch uses one four-warp CTA per `(request, layer, head)` and has no persistent work queue, but the [implemented kernel](https://github.com/flashinfer-ai/flashinfer/blob/a73fe28a14469b12c54e49927ca4fab065799efe/csrc/replayssm_materialize.cu#L97-L125) is a persistent grid-stride launch over the compacted `active_request_indices` list.

Requested resolution: use the public name consistently and update the launch, active-list, and prefix-bound descriptions to match the implementation.

tests/mamba/test_replayssm_materialize.py
```diff
@@ -0,0 +1,546 @@
...
+        zero_table,
+        src_slots,
+        dst_slots,
+        ring_start,
+        replay_prefix_len,
+        active_request_indices
+        if active_request_indices is not None
+        else _active_request_indices(replay_prefix_len),
+        state_dtype=state_dtype,
+        input_dtype=torch.bfloat16,
```
ishovkun: **[Low] Add coverage for the advertised FP16 input-cache path**

The `state_dtype=torch.float16` parametrization later in this file only exercises FP16 state storage: this helper still hardcodes `input_dtype=torch.bfloat16`, and all `x`/`B` cache fixtures are BF16. Therefore no test instantiates `input_dtype=torch.float16`, which is the path whose dtype semantics differ.

Please add a test with FP16 `x`/`B` cache tensors and `input_dtype=torch.float16`, compared against the independent PyTorch recurrence, if that dtype is intended to be supported. If input caches are BF16-only, restrict the public/JIT dtype contract and add a rejection test instead.

tests/mamba/test_replayssm_materialize.py
```diff
@@ -0,0 +1,546 @@
...
+        matrixA_dtype=torch.float32,
+        dim=64,
+        dstate=64,
+        num_heads=state[0].size(1),
+        heads_per_group=heads_per_group,
+        max_window=8,
```
ishovkun, line 85: **[Low] Cover `replay_prefix_len > MAX_WINDOW`**

I do not see a test whose replay prefix exceeds `max_window`; this helper fixes `max_window=8`, while every positive test count is at most 5. The kernel’s explicit oversized-count branch is therefore untested.

Once the intended contract is resolved, please add a regression test for it: if skipping is intentional, prefill the destination with a sentinel and assert that it remains unchanged; if an oversized count is invalid input, assert the chosen error/status behavior. This makes the current silent stale-destination outcome observable rather than accidental.

tests/mamba/test_replayssm_materialize.py
```diff
@@ -0,0 +1,546 @@
...
+        ring_buffer_len=1,
+    )
+    torch.cuda.synchronize()
+    assert torch.equal(state[0], state_before[0])
+    assert torch.equal(state[1], state_before[0])
+    assert torch.equal(scales[0], scales_before[0])
+    assert torch.equal(scales[1], scales_before[0])
+
+
+def _materialize_int8(
```
ishovkun: Could `_materialize` accept optional per-layer state scales and derive `state_dtype`, `input_dtype`, `matrixA_dtype`, `dim`, and `dstate` from its tensors? Then `_materialize_int8` could delegate to it—or disappear—instead of duplicating the pointer/stride construction. Deriving `input_dtype` would also make the missing FP16 input-cache test expressible through the shared helper.

ishovkun (changes requested):

csrc/replayssm_materialize_customize_config.jinja
hnover-nv: Could you expand on why?  does this not match checkpointing_ssu_customize_config.jinja?

flashinfer/jit/mamba/replayssm_materialize.py
hnover-nv: Oops!  Yes, should be just bf16 for input cache.  Should be fixed at both wrapper and JIT.

csrc/replayssm_materialize.cu
hnover-nv: All these are device-side tensors and so we can't actually do meaningful error checking here.  I removed the defensive replay_prefix_len > max_window, although as that was in the undefiend part of the API I think it also was arguably fine the way it was.  It's just a question of, given we can't raise errors, do we want to try to be defensive in terms of causing "corruption" or not.  But I was not coming down on that uniformly.

new version: remove that check.  Like checkpointing_ssu, any out-of-bounds ring_start or prev_num_accepted_tokens is just used (or in our case, replay_prefix_len> prev_num_accepted, whether more than max window or not), and the chips will fall where they may in terms of reading and computing w/ invalid data.

active_request_indices indexing into a negative replay_prefix_len will just skip, but I'm not documenting that as it's not guaranteed in future.  That's just what happens on that particular malformed input.  But it's a great meta-point, so:  I removed any documentation for replay_prefix_len entries NOT selected by active_request_indices: the requirement now is that if active_request_indices picks you, replay_prefix_len shoudl be >= 0, <= prev_num_accepted_tokens (not passed) <= MAX_WINDOW.  It can be anything it wants if it isn't picked.

Also, deleted the design doc, that was only intended as implementation scratch, not a permanent record.

flashinfer/mamba/replayssm_materialize.py
hnover-nv: I believe the torch compilation wrappers are currently a no-op to avoid extra launch overhead, but we should definitely support it.

I'm not sure what the suggested fix is about the tokens.  What's described does not seem to enable capturing the dependencies for torch compile to use.  I implemented what I think is the right fix: there are two optinal args which if not None should be list[Tensor], one shoudl contain the backing tensors for all the inputs, one for all the outputs. For something like TRT-LLM this woudl just be the base state x,B,C, etc. buffers that are [max_requests, layer, ...].  For vLLM it would be the backing tensors for the slot storage that is subdivided into the individual SSM,ring X, ring B, etc. fields, or perhaps that specialized to all the ring types, etc.

csrc/replayssm_materialize.cu
hnover-nv: Done!

csrc/replayssm_materialize.cu
hnover-nv: Done, w/ updated comment so readers don't think it's used in single layer kernels.

csrc/replayssm_materialize.cu
hnover-nv: Thanks added here and audited restrict usage against checkpointing_ssu.

csrc/replayssm_materialize.cu
hnover-nv: Good point, tied it to a constant name.

docs/design_docs/replayssm_prefix_materialization.md
hnover-nv: Sorry, this doc should not have been committed.  Removed.

tests/mamba/test_replayssm_materialize.py
hnover-nv: fp16 input was a mistake, so no need to cover it.

tests/mamba/test_replayssm_materialize.py
hnover-nv: Now undefined behavior, so no test needed.

tests/mamba/test_replayssm_materialize.py
hnover-nv: Thanks, good cleanup.  done.

csrc/replayssm_materialize.cu
```diff
@@ -0,0 +1,442 @@
...
+  ffi::CUDADeviceGuard device_guard(state_ptrs.device().device_id);
+  const cudaStream_t stream = get_stream(state_ptrs.device());
+  int64_t const work_items = int64_t(p.batch) * p.layers * p.heads;
+  constexpr size_t smem =
+      sizeof(std::conditional_t < sizeof(state_t) == 1,
+             CheckpointingSsuStorage8bit<input_t, state_t, NPREDICTED, MAX_WINDOW, DIM, DSTATE>,
+             CheckpointingSsuStorage < input_t, state_t, NPREDICTED, MAX_WINDOW, DIM, DSTATE >>);
+  FLASHINFER_CUDA_CHECK(cudaFuncSetAttribute(materialize_replay_kernel<state_t>,
+                                             cudaFuncAttributeMaxDynamicSharedMemorySize, smem));
+  constexpr int kThreadsPerCta = warpSize * kNumWarps;
```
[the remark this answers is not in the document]
hnover-nv: No, it's a macro in sm_30_intrinsics.h, hence constexpr.

csrc/replayssm_materialize_customize_config.jinja
ishovkun: It does, my bad :-) Just seemed out of place here....

csrc/replayssm_materialize.cu
ishovkun: if (!valid) {
    printf(
        "invalid input: n=%d value=%f block=%d thread=%d\n",
        n, value, blockIdx.x, threadIdx.x
    );
    __trap();
}

hnover-nv: Ah, right.  I'd rather not have it on the hot path in normal compilation, but like checkpiotning_ssu (that uses assert, so implicitly  requires not NDEBUG) I put it under a compile flag.  I added a printf but still fail w/ assert vs trap so that we get even more printing.  PTAL

ishovkun (approved):

kahyunnam:
@flashinfer-bot run

kahyunnam (approved):

PR #4718, opened 2026-08-25
perf(sm12x): optimize and unify static MoE

EricChen02:
Optimize the SM12x static MoE path while keeping NVFP4 and MXFP4 in one
`MoEStaticKernel` implementation.

## What changed

- Fold the retained NVFP4 schedule into the existing `MoEStaticKernel`; no
  separate retained-kernel class or source file remains.
- Split oversized routed experts into 32-row virtual tasks, so skewed experts
  remain within the tile's computed M extent.
- Retain two FC1 N-slices per scheduled work group and use the compact
  O(1)-claim scheduler.
- Retune the static tile/MAC ladder and extend the default NVFP4 static cutover
  from 640 to 1024 routed rows (`M <= 128` for top-k 8).
- Keep MXFP4 on the same unified class with its original 640-row cutover.
- Update workspace sizing, compile/launch ABI, source tracking, and dispatch
  tests for the unified kernel.

The branch is rebased on current `origin/main`. Upstream #4603 now provides the
shared B12x workspace support that was previously a separate commit in this PR,
so this update deliberately drops that redundant commit. The PR is now one
commit touching four files.

## Why

The former static schedule could assign more rows from a skewed expert than a
tile64/tile128 launch computed, and it repeated scheduling/staging work across
the two FC1 N-slices. Virtual 32-row tasks bound each physical work item while
the retained schedule reuses pipeline state across both slices.

## Correctness

Validated source-exact against the upstream measurement base. The branch is
now rebased on the current upstream head; the intervening upstream commits do
not modify any of this PR's four changed files.

- Measurement upstream base: `083012d6819cf97128e559616b12acb666f2fffe`
- Current upstream base: `3bbfeba6218b6de32d1e894243c010c8d3aacb21`
- PR head: `e184523e35f59144132d750e085243d409a16cf4`
- Local SM120 GPU: `GPU-1c189c11-e797-a795-cefd-495b190afebc`
- Shape: Qwen3.5-35B TP1, `E=256`, `H=2048`, `I=512`, `topk=8`
- Routes: three exact-marginal Zipf-0.75 samples
- M: 32, 64, 96, 128, 256, 512
- Candidate repeats: three per `(M, route)`

Result: **18/18 cases passed**.

| Metric | Result |
|---|---:|
| Maximum relative L2 | 0.00791765 |
| Minimum cosine similarity | 0.99996866 |
| Maximum zero rows | 0 |
| Static candidate repeats | Bitwise equal |

[...]

PR #4827, opened 2026-08-31
Prevent SM12x MoE graph replays from corrupting memory after workspace growth

ormandj:
## 📌 Description

On the SM12x MoE path, a larger prefill after decode-graph capture can free the workspace that those graphs still use. Their next replay writes through stale pointers, causing an illegal memory access or corrupting unrelated tensors if the allocator has reused that storage. Clearing the workspace cache can cause the same failure.

This PR keeps graph-referenced workspaces alive when cache entries grow or are cleared. Workspaces not used by a captured graph retain their existing lifetime.

- Mark workspace allocations and cache hits made during CUDA-graph capture.
- Retain marked workspaces when replaced or cleared, under the same lock as lookup and marking.
- Export `release_graph_referenced_workspaces()` for callers that have destroyed every graph using those workspaces. Calling it earlier is unsafe. Retained memory grows with replacement or clearing of graph-referenced entries.

Two graphs sharing a workspace still must not replay concurrently. This PR does not change that requirement.

## 🔍 Related Issues

No separate issue was filed for this defect.

### ✅ Pre-commit Checks

- [x] Run formatting and pre-commit checks.

## 🧪 Tests

Author validation at `7cc5b29a18` on main `9c10d2e8f2` with CUDA hidden: 8 tests passed. Targets: `tests/moe/test_b12x_workspace_graph_pinning.py`.

Author CPU run at `88444f3af2` on FlashInfer main `91bda04c66`: 8 tests, with CUDA hidden.

```bash
python -m pytest -q tests/moe/test_b12x_workspace_graph_pinning.py
```

The CPU regression uses a mocked allocator and capture flag to reproduce cache replacement and clearing. It checks that graph-marked entries remain referenced, ordinary entries are released, capture-time cache hits are marked, and explicit release clears retained entries. A two-thread test holds a capture-time lookup in progress and verifies replacement cannot race ahead of its marking. Fixtures restore all module caches afterward.

Historical author-reported serving reproduction, not rerun on the updated branch head: GLM-5.3-Flash W4A16, TP2, 2x RTX PRO 6000 Blackwell, decode graphs at batch sizes 1-4. A later prefill grew the workspace; replay then corrupted a persistent tensor with expandable segments enabled or raised an illegal memory access without them. A four-stream workload failed within 5-20 waves in 12 attempts before the change; two 48-wave runs, totaling 384 requests, and five context ladders completed without corruption or restarts afterward.

## Reviewer Notes

The serving observations are author reports from an integration deployment, not a standalone full-model reproducer in this repository. The committed test covers the workspace lifetime and locking contract with mocks. No isolated performance benchmark is claimed.

Developed with AI assistance.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CUDA Graph safety for Mixture-of-Experts workspace management.
  * Prevented workspaces referenced by captured graphs from being freed or overwritten during cache updates.
  * Added synchronized cache handling for safer concurrent operations.

* **New Features**
  * Added a public cleanup function to release retained workspaces after captured graphs are destroyed.

* **Tests**
  * Added coverage for workspace reuse, replacement, clearing, release, and concurrent graph-capture scenarios.


ormandj:
Rebased onto main b41735dc; clean rebase, no functional change (#4718 did not touch the workspace cache functions).

PR #4829, opened 2026-08-31
refactor(attention): unify PrimTS plan/run and page-table contracts

yuxianq:
## 📌 Description

Refactor the experimental PrimTS context, decode, and MLA APIs around one consistent `plan()`/`run()` ownership rule. The goal is to let an LLM inference framework cache a wrapper by its static execution profile and reuse it across requests without rebuilding the plan whenever request metadata changes. This is the FlashInfer-side prerequisite for [NVIDIA/TensorRT-LLM#17399](https://github.com/NVIDIA/TensorRT-LLM/pull/17399).

### Static planning and request-varying runs

- `plan()` defines the static profile used to select or compile a kernel and size its resources: device, exact batch profile, maximum Q/KV capacities, head and page geometry, dtypes, packed/fixed query storage, and mask/window specialization policy.
- `run()` receives all data that may change from request to request: query/KV tensors, page tables, sequence lengths, packed-query offsets, variable-window bounds, scale overrides, and optional output tensors.

Plans may retain defaults or resources whose addresses must remain stable rather than use them as compilation discriminators. Context plans retain default scale tensors and allocation-free variable-window scratch; decode and MLA plans bind caller-owned or private workspace. A failed re-plan does not replace the last complete plan.

`BatchPrefillTSWrapper` now follows the same contract as `BatchPrefillPagedTSWrapper`: planning is keyword-only and based on maximum capacities, while fixed or packed request tensors and metadata are supplied to `run()`. Packed offsets, request totals, and variable-window bounds can change between launches within the planned profile. Planning and warmup occur outside CUDA Graph capture, and a successful re-plan invalidates captures made from the previous plan. Before using a CUDA stream that is not already ordered after the planning stream, the caller must establish that dependency.

Decode's optional CPU `seq_lens` plan argument is specialization evidence only. When supplied, validated runs must preserve the proven length mode; omit it for fully dynamic sequence-length reuse. Every launch still receives the current device sequence lengths.

Validation remains enabled by default. Trusted framework integrations may use `validate=False` only after guaranteeing the complete dtype, device, shape, stride, alignment, value, aliasing, and lifetime contract, avoiding host synchronization in steady-state execution and CUDA Graph replay.

The benchmark adapters use packed-query storage for PrimTS decode and MLA (`packed_query=True`) with preallocated output and `validate=False`. Existing CUDA Graph, CUPTI, cold-L2, workspace, scale, mask/window, and automatic scheduling settings remain enabled.

### Unified fixed-row page metadata

Reusable paged context, decode, and MLA paths use the same request-varying metadata:

- `block_tables`: an `int32` CUDA tensor with shape `[B, C]`, unit inner stride, and a non-overlapping row stride;
- `seq_lens` / `seq_lens_kv`: contiguous `int32` CUDA sequence lengths with shape `[B]`.

For request `b`, logical page `p` resolves as `block_tables[b, p]`. Only the prefix required by the current sequence length is active, so inactive capacity columns are ignored. Compact `(C, 1)` tables and row-padded views such as `[B, 2, C][:, 0, :]` with stride `(2*C, 1)` are accepted directly.

The PrimTS one-shot APIs now use that same contract instead of a CSR triplet:

- `batch_prefill_with_paged_kv_cache(..., qo_indptr, block_tables, seq_lens_kv, ...)`
- `batch_decode_with_paged_kv_cache(..., block_tables, seq_lens_kv, ...)`

The decode trace schema, examples, and generated fixture are updated accordingly. Decode and MLA wrapper trace constraints also describe the runtime bounds enforced by the APIs. The separate generic FlashInfer decode adapter retains its existing CSR-to-fixed-table compatibility path.

### MLA public API naming

The MLA helpers now use a consistent `batch_mla_decode` name:

| Previous name | New name |
|---|---|
| `batch_decode_mla_with_paged_kv_cache` | `batch_mla_decode_with_paged_kv_cache` |
| `get_prims_ts_batch_decode_mla_workspace_size` | `get_prims_ts_batch_mla_decode_workspace_size` |
| `prims_ts_batch_decode_with_kv_cache_mla` | `prims_ts_batch_mla_decode_with_kv_cache` |

These are hard renames; the old aliases are not retained. Public exports, lazy MLA exports, documentation, tests, trace examples, and `fi_api` tags use the new names.

## 🔍 Related Issues

Follow-up to #4357 and prerequisite for [NVIDIA/TensorRT-LLM#17399](https://github.com/NVIDIA/TensorRT-LLM/pull/17399).

### ✅ Pre-commit Checks

- [x] Applicable pre-commit hooks, mypy, Ruff check/format, Python compilation, and diff checks passed.

## 🧪 Tests

- [x] Context suite: `100 passed`, `81 skipped` on the local non-Blackwell host.
- [x] Decode suite: `97 passed`, `88 skipped`.
- [x] MLA suite: `15 passed`, `84 skipped`.
- [x] Trace-template consistency: `759 passed`.
- [x] PrimTS benchmark contract tests: `15 passed`.
- [x] B200/SM100 focused matrix on the preceding implementation snapshot: `9 passed`, `0 skipped`. Two cases specifically covered the plan-stream readiness machinery removed by this follow-up; the remaining cases covered dynamic packed offsets, variable-window and fixed-table CUDA Graph replay, fixed-table one-shot context, and decode/MLA public parity. No current-head performance result is claimed.
- [x] Independent final staged review found no P0-P2 issues.

## Reviewer Notes

Please focus on whether each argument belongs to the static compilation/resource profile in `plan()` or the request-varying launch state in `run()`, and whether the resulting wrappers are safe to cache and reuse across requests and CUDA Graph replays under the documented caller-managed stream and lifetime contract.


flashinfer/trace/templates/attention.py
```diff
@@ -883,22 +800,39 @@ def _make_prims_ts_decode_wrapper_trace(
...
     inputs: dict[str, Tensor | Scalar] = {"q": Tensor(q_shape)}
     _add_fmha_cache_schema(
         inputs, axes, cache_param="paged_kv_cache", combined=combined
     )
     inputs.update(
         {
-            # Wrapper metadata is supplied to plan(), not run(). It remains in
-            # the trace as optional context, matching the existing paged GQA
-            # wrapper templates elsewhere in this file.
+            "paged_kv_indptr": Tensor(
```
PerkzZheng: Does it mean those tensors are maintained by wrapper previously, and now we are exposing it for framework reuse and maintaining ?

yuxianq: We provide it to the wrapper during `run` instead of `plan`, the wrapper just need to validate it during `run`. For every plan arguments that can change per forward, we defined it as live metadata like `paged_kv_indptr`. Live metadata makes the run a little slower due to extra validation, but it make the wrapper's plan cacheable, avoiding repeatedly plan during forward, which saves e2e time.

PerkzZheng: that makes sense, so we plan to add a fast path to skip checks later to reduce host overhead ? I see many changes related to check tensors.

yuxianq: Yes, skip some check, optimize tensor preparation, enable PDL and so on. There will be multiple PRs to improve perf, but this PR is just for correct implementation of cacheable wrappers.

PerkzZheng: thanks. LGTM now.

PerkzZheng (approved):

yuxianq:
@flashinfer-bot run

benchmarks/routines/attention.py
```diff
@@ -1021,31 +1021,33 @@ def testBatchDecodeWithPagedKVCacheWrapper(args):
         prims_ts_mask_type = (
             "causal" if not speculative_decode or effective_causal else "dense"
         )
+        # The common fixture intentionally exposes nonstandard outer strides;
+        # PrimTS accepts compact HND pages, so preserve the logical values in a
+        # backend-specific compact cache.
+        prims_ts_kv_cache = kv_cache.contiguous()
```
PerkzZheng, line 1076: this might introduce memory copies and allocation. we probably want to pass the original kv cache and relies TMA strides to deal with that ? sorry this might have been introduced in my previous changes. I can help resolve it later.

benchmarks/test_flashinfer_benchmark.py
PerkzZheng: this seems over-engineering, we might want to remove that.

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_resources.py
```diff
@@ -1893,20 +1993,128 @@ def k_desc(self, stage_info: StageInfo) -> prims.Tcgen05SmemDesc:
...
+    def _zero_paged_v_tail(
+        self,
+        stage_info: StageInfo,
+        *,
+        section: cutlass.Constexpr[FmhaStage],
+        tile_offset: cutlass.Constexpr[int],
+        seqlen_k: Int32,
+        kv_tile_start: Int32,
+    ) -> None:
+        """Overwrite request-invalid V rows after TMA completion."""
```
PerkzZheng, line 2016: this is not needed as we will mask it properly in softmax. Or this is meant to avoid NANs happening in V paddings ? if so, we should zero it in previous updating kv cache kernel instead of fmha kernels which should have obvious overheads.

flashinfer/attention/prims_ts/context.py
```diff
@@ -1736,131 +2210,114 @@ def _prepare_out(
...
+        out_dtype : torch.dtype, optional
+            Output dtype; defaults to ``q_dtype``.
+        packed : bool
+            Whether run tensors use packed ``[total_tokens, H, D]`` storage.
         mask_type : {"dense", "causal", "variable_window"}
             Attention mask mode. ``variable_window`` is supported only for
             fixed-shape inputs.
         window_left : int
             Left sliding-window extent, or ``-1`` to disable the window.
-        variable_window_token_starts, variable_window_token_ends : torch.Tensor, optional
```
PerkzZheng, line 1834: why it has been removed ?

flashinfer/attention/prims_ts/context.py
```diff
@@ -1736,131 +2210,114 @@ def _prepare_out(
...
-            Inclusive per-query K bounds required for ``variable_window``.
-            Both must be CUDA int32 tensors shaped ``[B, Sq]`` and satisfy
-            ``0 <= starts[b, q] <= ends[b, q] < Sk``.
         sm_scale : float, optional
             Softmax scale; defaults to the inverse square root of head size.
         output_scale : float
             Scale applied to the attention output.
-        out_dtype : torch.dtype, optional
```
PerkzZheng, line 1842: same here

flashinfer/attention/prims_ts/context.py
```diff
@@ -422,20 +538,19 @@ def _build_variable_window_cta_starts(
...
-        padded = torch.full(
-            (geometry.batch_size, padded_rows),
-            _INT32_MAX,
-            dtype=torch.int32,
-            device=geometry.device,
-        )
-        padded[:, : geometry.max_seq_len_q] = starts_2d
+    padded = state.variable_window_padded_starts
+    if padded is not None:
+        padded.fill_(_INT32_MAX)
```
PerkzZheng, line 579: why it has been changed to fill + copy ?

flashinfer/attention/prims_ts/decode.py
```diff
@@ -2715,27 +3033,71 @@ def batch_decode_with_paged_kv_cache(
...
+        kv_data_type=k_cache.dtype,
         o_data_type=output_dtype,
         mask_type=mask_type,
         window_left=window_left,
+        seq_lens=seq_lens_host,
     )
     return wrapper.run(
         q,
         paged_kv_cache,
+        seq_lens_kv,
```
PerkzZheng: this will need @saltyminty and @qsang-nv 's opinions where we move the three tensors to run.

qsang-nv:
Potential silent-correctness gap when manually capturing the prims-ts adapter with mutable stable metadata.

In `flashinfer/decode.py`, the `prims-ts` branch always forwards the plan-time KV lengths as optional specialization evidence:

```python
self._prims_ts_wrapper.plan(
    ...,
    seq_lens=kv_lens_arr_host,
    workspace_buffer=self._float_workspace_buffer,
)
```

The low-level `BatchDecodePagedTSWrapper.plan()` contract says that omitting `seq_lens` keeps both KV-prefix and KV-length handling dynamic. The generic adapter does not expose that choice.

For configurations where `requires_runtime_kv_lengths` is false, a uniform plan-time batch whose lengths all equal `max_kv_len` selects `kv_lengths_mode="planned_uniform_max"`. This becomes `static_native_uniform_kv=True`; the kernel then uses the compiled maximum for task-domain and masking decisions instead of loading each request's runtime length.

Normally, a validated run rechecks that the runtime lengths still satisfy this proof. During manual CUDA graph capture, however, the adapter calls:

```python
validate=not torch.cuda.is_current_stream_capturing()
```

and graph replay does not execute the Python validation again.

Consequently, if the contents of the wrapper-owned stable `_kv_lens_buffer` are refreshed in place — for example, by integration code updating metadata or replanning into the same storage — without re-recording the graph, a graph captured from a uniform warm-up batch can continue using the `planned_uniform_max` kernel for later ragged lengths. That can include KV positions beyond the logical per-request length without raising an error.

The current graph test explicitly covers only replaying an unchanged fixed plan and says that replanning is not covered. Could we clarify the intended contract here?

- If stable metadata contents may change between replays, the adapter should omit the evidence (`seq_lens=None`) by default, or expose specialization as an explicit opt-in.
- If they must remain immutable, that restriction should be documented/enforced; calling them "live device metadata" is otherwise misleading.

A regression test that captures from uniform-max lengths, updates the same stable length buffer to ragged values, and replays would distinguish these two intended contracts.

flashinfer/attention/prims_ts/context.py
yuxianq: It is still a plan argument, just move it to L2290.

flashinfer/attention/prims_ts/context.py
yuxianq: We move them to wrapper's run arguments, so that we can use different variable_window_token_starts/variable_window_token_ends for different batches.

flashinfer/attention/prims_ts/decode.py
yuxianq: We move them to run so that we can cache wrapper's plan and pass different `seq_lens_kv` for different batches without re-planning.

flashinfer/attention/prims_ts/context.py
yuxianq: We allow to change variable_window_token_starts for different batches. If CUDA graph enabled, we need a stable buffer to store variable_window_token_starts and update its content every forward, so we choose to use state.variable_window_padded_starts as stable buffer, fill + copy updated values into it. The original create new tensor method does not work.

yuxianq:
@qsang-nv Instead of passing seq_lens to both plan and run and validate the consistency, how about we just allow to pass it in either place? If we pass it in plan, we copy it to GPU and use its GPU peer in run, assert that seq_lens is None for kv_lengths_mode=planned_uniform_max. No further validation is need with this solution.

yuxianq:
@flashinfer-bot run

benchmarks/test_flashinfer_benchmark.py
yuxianq: Removed

benchmarks/routines/attention.py
yuxianq: This is benchmark code and it executes outside benchmark timing, so I think it is fine to have some copies.

qsang-nv:
> @qsang-nv Instead of passing seq_lens to both plan and run and validate the consistency, how about we just allow to pass it in either place? If we pass it in plan, we copy it to GPU and use its GPU peer in run, assert that seq_lens is None for kv_lengths_mode=planned_uniform_max. No further validation is need with this solution.

That design is a good fix — it turns the plan/run consistency requirement into a structural invariant (mutually exclusive sources) instead of a runtime check, so `validate=False` no longer matters for this specific field. One thing worth confirming: the `seq_lens is None` assert for `kv_lengths_mode=="planned_uniform_max"` needs to be unconditional (not gated behind `validate`), since it's a cheap host-side Python check with no device sync — if it's skipped during capture the same gap reopens.

Also: does `BatchMLADecodePagedTSWrapper` (`mla_decode.py`) have the same plan-time `seq_lens` specialization path? If so, the same fix should probably apply there too, otherwise MLA decode keeps the gap this PR is fixing for GQA decode.

yuxianq:
@qsang-nv 
> the seq_lens is None assert for kv_lengths_mode=="planned_uniform_max" needs to be unconditional 

Sure

>  does BatchMLADecodePagedTSWrapper (mla_decode.py) have the same plan-time seq_lens specialization path?

Not, this specialization is only for BatchDecodePagedTSWrapper

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_resources.py
yuxianq: This solves a real accuracy bug for e2e test to zero out NaNs in last V page's padding part. Its overhead is low because it clears only a small region already in shared memory. It does not launch another kernel or touch the full KV cache. It reuse the TMA wrap that loads V and may overlap with other wraps. The micro benchmark shows a 0.181% slow down in kernel, which is negligible. I am also exploring other solutions in trtllm side (mainly for flashinfer backend and flashinfer trtllm-gen path, their path has this accuracy bug), but I don't think we need to drop this fix from this PR.

benchmarks/routines/attention.py
PerkzZheng: okay. thanks.

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_resources.py
PerkzZheng: have we benchmarked various cases (head dim, dtypes) to validate the regression ? if all good, I am okay to keep that. thanks.

flashinfer/attention/prims_ts/context.py
PerkzZheng: okay. thanks.

flashinfer/attention/prims_ts/context.py
PerkzZheng: thanks.

flashinfer/attention/prims_ts/context.py
PerkzZheng: shouldn't the update be handled outside the run() by previous kernel ? in other words, the previous kernel should update the starts/ends buffer directly ?

PerkzZheng (approved):

qsang-nv:
@yuxianq Thanks for confirming. I don't see this landed yet — `flashinfer/attention/prims_ts/decode.py` is unchanged since `430c20dda2`. Is the either-plan-or-run exclusivity + unconditional `seq_lens is None` assert still planned for this PR, or should this become a tracked follow-up?

yuxianq:
@qsang-nv Pushed now.

flashinfer/attention/prims_ts/context.py
yuxianq: I re-investigate this issue, it is unrelated to CUDA graph. variable_window_padded_starts is used to calculate variable_window_cta_starts, we use padding here because fill+copy+one_reduce is 6–15% faster than two_reduce. The real issue is should we calculate `variable_window_cta_starts` outside of run. I will try to add `variable_window_cta_starts` as an optional argument to reduce overhead.

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_resources.py
yuxianq: I bench it in various cases, it may cause >10% perf drop in some cases. But we cannot assume that the KV cache update kernel outside of prims_ts will zero out the final V page. I add an `paged_v_tail_is_zero` plan argument, it default to False to guarantee the accuracy. If the framework to call it ensure that it has zero out V page, it can set paged_v_tail_is_zero=True to avoid overhead. I will create a trtllm PR later to zero out V page in context's pre-processing instead.

flashinfer/attention/prims_ts/context.py
yuxianq: I have added `variable_window_cta_starts` argument.

flashinfer/decode.py
```diff
@@ -1586,13 +1600,44 @@ def _plan_impl(
         self._batch_size = batch_size
         self._num_qo_heads = num_qo_heads
         self._num_kv_heads = num_kv_heads
-        self._block_tables: Optional[torch.Tensor] = block_tables
+        next_block_tables: Optional[torch.Tensor] = block_tables
+        if self._backend != "prims-ts":
```
saltyminty: We defer setting `self._block_tables` until after a successful plan – I assume to avoid a failed re-plan from updating plan state? Why does this not also apply to the other fields?

saltyminty (commented):
Left one minor comment, will rerun CI. The flagged API breakage can be ignored as prim-ts is experimental

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_resources.py
PerkzZheng: thanks. it makes sense to me now by marking it as optional.

flashinfer/attention/prims_ts/context.py
PerkzZheng: thanks.

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_resources.py
yuxianq: The trtllm PR is https://github.com/NVIDIA/TensorRT-LLM/pull/18925

flashinfer/decode.py
yuxianq: This change is unnecessary for the latest implementation, I have reverted it. Thanks for reminding me.

qsang-nv (approved):

PR #4831, opened 2026-08-31
feat: add is_deterministic mode to top_k_renorm_probs

zpeng-xai:
## Problem

`top_k_renorm_probs` returns bitwise-different results on repeated calls with bitwise-identical input. On GB300 (192x131072 fp32, k=50): 19/19 repeated invocations differed, max_abs_diff ~3e-8.

This matters for replicated execution: tensor-parallel LLM serving runs sampling/spec-decode verify redundantly on every TP rank with identical inputs, and per-invocation nondeterminism makes ranks compute rank-divergent `target_probs`. With shared rejection-sampling coins, committed tokens then flip exactly at inverse-CDF boundaries (measured ~1e-7..1e-5 per sampling decision, roughly linear in top_k), which diverges rank-local radix caches and eventually surfaces as collective-shape mismatch hangs — the exact chain independently root-caused in sgl-project/sglang#22458.

## Root cause

The radix **pivot selection is already deterministic** (integer histogram accumulation) — we verified the kept support is stable 20/20 across calls. The only nondeterministic step is the multi-CTA **renormalization sum**: per-CTA partial sums are combined with float `atomicAdd(&state->sum_topk, block_sum)`, so CTA arrival order changes the rounding of the normalizer, wobbling every output value in its last bits.

## Fix

Mirror `top_p_renorm_probs`' existing deterministic option: an `is_deterministic` flag (default `False`, existing behavior and performance unchanged) that switches the multi-CTA accumulation to order-independent fixed-point integer atomics:

- `RadixRowState` gains a `sum_topk_fixed` (u64) field; contributions are quantized at 2^44 scale and combined with integer `atomicAdd` (associative, order-independent).
- Overflow bound is a kept mass of 2^20 — far above probability-input magnitudes (docs note the assumption); quantization error (≤2^-44 per CTA) is below fp32 ulp.
- Single-CTA path already sums in a fixed block-reduction order and is unaffected; the flag only changes the multi-CTA path.
- Plumbed through `csrc/renorm.cu`, the sampling binding, and the Python API with the same signature/doc shape as `top_p_renorm_probs`.

## Validation (GB300, JIT build, 192x131072 fp32)

```
[det=True ] bitwise-different: 0/19            (was 19/19)
[det=False] bitwise-different: 19/19           (default unchanged)
[correct  ] same_support_vs_default=True  max_val_diff=1.5e-08  row_sum_err=2.4e-07
[k=2000   ] bitwise-different: 0/9             (per-row top_k tensor)
[perf     ] default=0.148ms  deterministic=0.151ms  overhead=+2.6%
```

Added `test_top_k_renorm_probs_deterministic` (bitwise stability across 10 calls, support equality vs default, value closeness, row-sum normalization) across batch/vocab/k/dtype parametrizations.

## Summary by CodeRabbit

* **New Features**
  * Added an optional deterministic execution mode for top-k probability renormalization.
  * Deterministic mode supports scalar and per-row top-k values.
  * Supports float32, float16, and bfloat16 inputs while preserving existing default behavior.
  * The option is available through the top-k renormalization API and produces consistent results across repeated calls.

* **Tests**
  * Added coverage for deterministic and non-deterministic modes across supported data types and top-k configurations.


zpeng-xai:
Extended validation matrix on GB300 (3 dtypes x 4 shapes x 3 k-regimes + mixed per-row k, JIT build):

**Determinism** — `is_deterministic=True` was bitwise-stable in **0/29 differing calls across all 37 configs**, including the `k >= vocab_size` sum-only branch and a mixed per-row-k batch exercising both branches in one launch. Two fresh processes produced **37/37 identical output checksums** (cross-process reproducibility, which replicated TP execution actually requires).

**Default path unchanged** — fp32 multi-CTA remains nondeterministic (9/9), single-CTA small-vocab remains deterministic (dispatch keeps the non-det kernel, flag is a no-op there). Notably, fp16/bf16 outputs are stable even on the default path: the fp32 normalizer wobble is rounded away by the half-precision output cast, so this bug specifically bites fp32 consumers.

**Performance** (default → deterministic):

| regime | typical | overhead |
|---|---|---|
| k < vocab, multi-CTA (e.g. 192x131072, k=50) | 0.147 → 0.151 ms | **+2.0% ~ +3.6%** |
| k >= vocab (sum-only branch) | 0.056 → 0.060 ms | +4% ~ +9% (tiny absolute) |
| single-CTA (vocab 32000) | 0.021 → 0.021 ms | ~0% (no-op) |

Absolute overhead ≤ 0.02 ms in every configuration.

include/flashinfer/topk.cuh
```diff
@@ -143,8 +143,18 @@ struct RadixRowState {
...
   int output_counter;          // For collecting top-k indices (RadixTopK)
   float sum_topk;              // For RenormProb: sum of top-k elements
+  // For deterministic RenormProb: fixed-point sum of top-k elements. Integer
+  // atomics are associative, so the accumulated value is independent of CTA
+  // arrival order (float atomicAdd on sum_topk is not).
+  unsigned long long sum_topk_fixed;
 };
 
+// 2^44 fixed-point scale for the deterministic renorm sum. Probability inputs
+// keep row sums near 1, far below the 2^20 overflow bound of a u64
```
zpeng-xai: **[suggestion]** The new field and scale comments (146-155) plus the two kernel write-ups (1863-1866, 1977-1978) restate the `if constexpr` and paste the PR design rationale (associativity, CTA arrival order, 2^20 overflow bound, fp32 ulp). That belongs in the PR body; the scale comment is also slightly sloppy ("<= 2^-44" is the truncation bound; nearest rounding is 2^-45, and "below fp32 ulp at probability magnitudes" is about the *group sum*, not individual tiny probs).

**Suggestion:** Keep a one-liner on the field (`// order-independent fixed-point stand-in for sum_topk`) and `constexpr double kRenormDeterministicScale = 0x1.0p44; // u64 overflows at sum >= 2^20`. Drop the kernel restatements.

tests/utils/test_sampling.py
```diff
@@ -543,6 +543,42 @@ def test_top_k_renorm_probs(batch_size, vocab_size, k, distribution, dtype):
...
+    """is_deterministic=True must be bitwise reproducible across invocations.
+
+    The default multi-CTA path accumulates the renormalization sum with float
+    atomicAdd, so CTA arrival order changes the last-bit rounding and repeated
+    calls on identical input differ bitwise. Replicated execution (e.g.
+    tensor-parallel ranks running the sampler redundantly on identical inputs)
+    then diverges, which desynchronizes downstream state. The deterministic
+    mode accumulates in fixed-point integer atomics instead.
+    """
+    if k > vocab_size:
```
zpeng-xai: **[suggestion]** `test_top_k_renorm_probs_deterministic` skips whenever `k > vocab_size`, so it never executes the sum-only branch (`k >= vocab_size`) that has its own copy of the new init/atomicAdd/reconvert sequence. It also never passes a per-row / mixed-k tensor, which is the other persistent-loop combination of those two copies. vocab=128256 does hit multi-CTA fp32 on current GPUs; the untested path is the duplicated one.

**Suggestion:** Add cases with `k >= vocab_size` (scalar and mixed per-row `top_k`) under `is_deterministic=True`, and assert bitwise stability there too.

include/flashinfer/topk.cuh
```diff
@@ -2000,7 +2046,7 @@ template <typename DType, typename IdType>
 cudaError_t RadixTopKRenormProbMultiCTA(DType* probs, DType* renormed_prob, IdType* top_k_arr,
                                         uint32_t batch_size, uint32_t top_k_val,
                                         uint32_t vocab_size, RadixRowState* row_states_buffer,
-                                        cudaStream_t stream = 0) {
+                                        bool deterministic = false, cudaStream_t stream = 0) {
```
zpeng-xai: **[suggestion]** `bool deterministic = false` was inserted *before* the existing `cudaStream_t stream = 0`. Any out-of-tree C++ caller of `RadixTopKRenormProbMultiCTA(..., buffer, stream)` now silently treats a non-null stream as `deterministic=true` and launches on the default stream. In-tree `renorm.cu` was updated, but this is a positional ABI/source footgun.

**Suggestion:** Put the new flag last (`cudaStream_t stream = 0, bool deterministic = false`) so existing positional stream arguments keep compiling with the old meaning.

kahyunnam:
@flashinfer-bot run

flashinfer/sampling.py
```diff
@@ -520,6 +520,7 @@ def top_k_renorm_probs(
         probs: torch.Tensor,
         maybe_top_k_arr: Optional[torch.Tensor],
         top_k_val: int,
+        is_deterministic: bool,
```
kahyunnam: Let's either: 1) move the new `is_deterministic` arg to the end of this arg list and give it a default value, or 2) double check to update every caller for the new custom_op schema.  

Eg for 2), the unchanged `ProbsTopKOp` direct caller is incompatible: https://github.com/flashinfer-ai/flashinfer/blob/main/flashinfer/logits_processor/operators.py#L139-L140

zpeng-xai: Make sense to give it a `is_deterministic=false` for backward compatibility.

zpeng-xai:
@kahyunnam Thanks for the review! Commit 1e7fad2 implements your suggested backward-compatible signature: is_deterministic is now trailing with a default of False, with regression coverage for the existing four-argument caller.

kahyunnam:
@flashinfer-bot run tests/utils

jazzikp:
@kahyunnam The A10G failure (`tests/utils/test_logging_replay.py::test_bmm_fp8_replay`) is pre-existing and unrelated to this PR: the test hard-codes `backend="cudnn"` without a capability skip while `bmm_fp8`'s cuDNN backend is gated to SM89+, so it fails on SM86 in any scoped `tests/utils` run; this PR doesn't touch it or the GEMM path, and H100 plus the internal pipeline (25/25) passed on this head.

kahyunnam:
@flashinfer-bot run tests/utils

kahyunnam:
@flashinfer-bot run tests/utils

kahyunnam:
@flashinfer-bot run tests/utils

kahyunnam:
@flashinfer-bot run tests/utils/test_sampling.py tests/utils/test_logits_processor.py tests/trace/test_top_k_renorm_probs_reference_correctness.py

kahyunnam (approved):

jazzikp:
@kahyunnam Thank you for testing and review!

Issue #4833, opened 2026-08-31
[Bug] BatchMLAPagedAttentionWrapper fails when kv_len_arr metadata is not rank-1


yufeiwu-nv:
## Summary

`BatchMLAPagedAttentionWrapper` consistently fails for the DeepSeek-R1 BF16 benchmark case because the generated `kv_len_arr` metadata is not rank-1. The failing tensor has shape `(batch_size, 1, 1, 1)`.

## Environment

- GPU: GB300
- Branch: `main`
- FlashInfer version reported by the benchmark: `0.6.18`
- First observed failing commit: `2af72b0`
- Confirmed failing commits include `4442800` and `231f708`

`2af72b0` is the first observed failing commit in the available CI history; it is not claimed to be the introducing commit.

## Reproduction

```bash
python3 flashinfer_benchmark.py --routine BatchMLAPagedAttentionWrapper --backends trtllm-native fa2 fa3 cutlass cute-dsl --page_size 16 --batch_size 1 --s_qo 1 --s_kv 8192 --num_qo_heads 128 --num_kv_heads 128 --head_dim_ckv 512 --head_dim_kpe 64 --random_actual_seq_len -vv --refcheck --q_dtype bfloat16 --kv_dtype bfloat16 --generate_repro_command --case_tag DeepSeek-R1
```

## Actual behavior

```text
Error: kv_len_arr metadata must be rank-1, got shape (1, 1, 1, 1).
```

The same signature also reproduces for batch sizes 16 and 64.

## Expected behavior

The wrapper should accept or normalize the generated sequence-length metadata and complete the benchmark reference check.

## CI evidence

- Commit `4442800`: https://nv/flashinfer-ci/-/jobs/417141338
- Commit `231f708`: https://nv/flashinfer-ci/-/jobs/418037682

yufeiwu-nv:
also reproduces on B300 https:/nv/flashinfer-ci/-/pipelines/65546656

yufeiwu-nv:
verified fixed: https:/nv/flashinfer-ci/-/pipelines/65915316

PR #4835, opened 2026-08-31
fix(benchmark): align MLA metadata with backend contracts

lunarz-dev:
## 📌 Description

Flatten sampled KV lengths before planning MLA backends and pass page tables only to CUTLASS.

This allows the benchmark harness to use the metadata representation required by each planned MLA backend.

Objective: Restore execution of MLA benchmark cases with randomly sampled sequence lengths.

Scope: Updates only the MLA benchmark routine and its regression test; production MLA backend implementations are unchanged.

## 🔍 Related Issues

https://github.com/flashinfer-ai/flashinfer/issues/4833

### ✅ Pre-commit Checks

- [x] I have run `pre-commit run --all-files` and fixed any reported issues.

## 🧪 Tests

- [x] B300 MLA benchmark cases completed with non-empty results for batch sizes 1 and 16.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MLA paged-attention benchmark compatibility by passing key/value sequence lengths in the expected format.
  * Updated backend-specific argument handling for more reliable FlashAttention and CUTLASS benchmark execution.

* **Tests**
  * Added regression coverage for MLA benchmarks with a single query token.
  * Verified that each backend receives correctly shaped attention metadata.


saltyminty (approved):
LGTM – cutlass actually prefers an explicit plan call before run now, though we can defer updating the benchmark to that setup to a future change.

PR #4839, opened 2026-08-31
feat(attention): opt-in LDTM.STAT for PrimTS context row_max

kris1025:
## Summary

Adds the `LDTM.STAT` (`tcgen05.ld.red.max`) row-max primitive to the PrimTS
context FMHA softmax, exposed as the `FmhaConfig.uses_ldtm_stat` flag and
**enabled by default on B300 (SM103) and Rubin (SM107)** via the context runner,
plus the CI DSL version bump needed to build and test it. On B200 (SM100) the
instruction is unavailable, so the default stays off and the shipped kernel is
**byte-identical to `main`** there.

## What's in this PR

**Feature — LDTM.STAT row_max** (`uses_ldtm_stat`)

- `fmha_resources.py`: adds the `FmhaConfig.uses_ldtm_stat` flag (kernel-level
  default `False`) and, in `TmemSPResource`, switches the non-masked
  `compute_row_max` path to `tcgen05.ld.red.max` — fusing the per-chunk row_max
  into the TMEM S load instead of a register-side lane max. Masked variants
  (causal tail, sliding-window, dense-k) keep the register-reduce path; they
  must observe raw S before masking.
- `fmha_kernel.py`: exposes `uses_ldtm_stat` as an `FmhaTs` kwarg.
- `context.py`: the context runner derives the flag per device via
  `_default_uses_ldtm_stat()` — **on by default on B300 (SM103) and Rubin
  (SM107)**, off on B200 (SM100) where `tcgen05.ld.red` is unavailable — and
  threads it through both the ragged and paged build paths.
- `tests/attention/test_attention_ts_context.py`: adds wiring + invariant unit
  tests (kernel default-off, schedule builds, and the per-device default tracks
  `(10, 3)` / `(10, 7)`).

**Dependency — nvidia-cutlass-dsl 4.7.0 → 4.8.0.dev0** (required for the
`tcgen05.ld.red.max` codegen)

- Pins the CI build/test DSL in `ci/cuda-versions.json`; raises the install
  floor in `requirements.txt` and the `cu12`/`cu13` pyproject extras.
- `ci/validate_cuda_versions.py`: accepts and correctly orders a PEP 440
  `.devN` suffix (dev sorts before `a`/`b`/`rc`).
- Updates the DSL version strings referenced in the PrimTS/GDN test skip-guards,
  the `sm100_blk64` DSL helpers, and the GDN experimental README.

## Safety

`uses_ldtm_stat` is a trace-time constant, so the CuTe-DSL tracer emits only the
selected branch. On B200 (SM100) the runner leaves it `False` and the shipped
kernel is **byte-identical to `main`** there. On B300 (SM103) and Rubin (SM107)
the runner turns it on by default — the intended behavior change — and only the
non-masked `compute_row_max` path differs: `TmemOResource` and the net
`fmha_tasks.py` diff are unchanged from `main`, and all masked paths are
untouched.

A couple of scheduling tweaks prototyped alongside the feature measured no
benefit on B300 and were removed within the branch, so they net to zero in the
diff.

## Validation (B300 / SM103)

Benchmarked on an **exclusive** B300 (SM103) node (whole node reserved, no
co-tenant), **batch size 16**, GQA H=64 / Hkv=8, D=128, with `uses_ldtm_stat`
forced OFF vs ON in separate processes (the context runner's per-device default
is pinned so the OFF process never silently runs the ON path). The sweep covers
**both storage layouts** — packed THD and paged HND with a non-identity page
table — × dense/causal × S ∈ {2048, 4096, 8192, 16384}: 16 cases per dtype, 32
total. 100 timed iters/case with cold L2, medians reported. Timing is via CUDA
events (the CI container ships no CUPTI, so `bench_gpu_time` falls back
automatically — identically for OFF and ON). The OFF baseline was reproduced
against a separate, non-exclusive run on the same node and matched to within
run-to-run noise (14 of 16 cases within ±0.6%, worst 1.1%), so the deltas below
are not an artifact of node sharing.

- **Correctness**: validated with the repo's own context-attention checker
  (`tests/attention/test_attention_ts_context.py::_assert_context_correct` — a
  dtype-branched `torch.testing.assert_close` plus a global relative-L2 gate,
  the same bar the unit tests enforce). All 32 cases pass in **both** bf16 and
  fp8 (`float8_e4m3fn` in → bf16 out), for both layouts. LDTM.STAT-on is
  **bit-identical** to LDTM.STAT-off — the per-case relative-L2 matches to every
  reported digit in all 32 cases — because `tcgen05.ld.red.max` returns the same
  row-max as the manual reduction it replaces, so the fusion changes throughput,
  not results. Relative-L2 ≈ 0.021–0.024 for fp8 (gate 0.1; the E4M3
  probability-quantization floor) and ≈ 0.0019–0.0021 for bf16 (gate 0.02).
- **Perf**: LDTM.STAT-on is faster in every case measured. Mean speedup by
  layout and dtype:

  | dtype | packed | paged |
  |-------|-------:|------:|
  | bf16  | +2.72% | +2.20% |
  | fp8   | +5.52% | +4.93% |

  Fusing the per-chunk row_max into the TMEM S-load pays off more in fp8, where
  the matmuls are cheaper and softmax is a larger share of the kernel, and more
  on dense than causal (the fused row-max lives in the non-masked path, so
  causal — which masks part of it — sees less).

**bf16** (H=64 / Hkv=8, D=128, batch=16), median GPU ms:

| case | off_ms | on_ms | speedup |
|------|-------:|------:|--------:|
| packed dense 2048 | 2.8252 | 2.7218 | +3.66% |
| packed dense 4096 | 10.8150 | 10.4725 | +3.17% |
| packed dense 8192 | 43.3981 | 41.7961 | +3.69% |
| packed dense 16384 | 172.3961 | 166.3278 | +3.52% |
| packed causal 2048 | 1.7254 | 1.6948 | +1.77% |
| packed causal 4096 | 6.0836 | 5.9710 | +1.85% |
| packed causal 8192 | 23.1711 | 22.7205 | +1.94% |
| packed causal 16384 | 90.5610 | 88.5955 | +2.17% |
| paged dense 2048 | 2.8683 | 2.7822 | +3.00% |
| paged dense 4096 | 10.9589 | 10.6609 | +2.72% |
| paged dense 8192 | 43.7944 | 42.3752 | +3.24% |
| paged dense 16384 | 174.7451 | 168.7577 | +3.43% |
| paged causal 2048 | 1.8893 | 1.8708 | +0.98% |
| paged causal 4096 | 6.5137 | 6.4256 | +1.35% |
| paged causal 8192 | 24.3318 | 23.9939 | +1.39% |
| paged causal 16384 | 94.0002 | 92.6024 | +1.49% |

**fp8** (`float8_e4m3fn` in → bf16 out, H=64 / Hkv=8, D=128, batch=16), median GPU ms:

| case | off_ms | on_ms | speedup |
|------|-------:|------:|--------:|
| packed dense 2048 | 2.3045 | 2.2006 | +4.51% |
| packed dense 4096 | 8.9436 | 8.4695 | +5.30% |
| packed dense 8192 | 35.7765 | 34.3255 | +4.06% |
| packed dense 16384 | 142.3252 | 133.9633 | +5.88% |
| packed causal 2048 | 1.4551 | 1.3650 | +6.19% |
| packed causal 4096 | 5.1071 | 4.8046 | +5.92% |
| packed causal 8192 | 19.3607 | 18.1105 | +6.46% |
| packed causal 16384 | 75.7499 | 71.3134 | +5.86% |
| paged dense 2048 | 2.3317 | 2.2272 | +4.48% |
| paged dense 4096 | 9.0434 | 8.5688 | +5.25% |
| paged dense 8192 | 36.1381 | 34.4387 | +4.70% |
| paged dense 16384 | 143.1347 | 135.6068 | +5.26% |
| paged causal 2048 | 1.4449 | 1.3752 | +4.82% |
| paged causal 4096 | 5.0580 | 4.8189 | +4.73% |
| paged causal 8192 | 19.2594 | 18.3464 | +4.74% |
| paged causal 16384 | 75.5303 | 71.4147 | +5.45% |

Rubin (SM107) enables the same path because it exposes `tcgen05.ld.red.max` and
compiles through the `sm_100f` family target (handled by `_validate_device`); it
has not yet been perf/correctness-validated on Rubin silicon.

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in attention execution mode using fused row-maximum reduction for supported unmasked workloads.
  * Added configuration support for selecting this mode.

* **Compatibility**
  * Updated CuTe-DSL requirements and validation to support the 4.8.0 development release and development-version ordering.

* **Tests**
  * Added coverage for the new attention option and updated version compatibility checks.

* **Documentation**
  * Updated CuTe-DSL version references, requirements, and related guidance.

### DIFF
--- a/flashinfer/attention/prims_ts/kernels/fmha_context/fmha_kernel.py
+++ b/flashinfer/attention/prims_ts/kernels/fmha_context/fmha_kernel.py
@@ -2366,6 +2366,10 @@ class FmhaTs:
         grouped-query attention with an even repeat count.
     enable_skip_correction : bool, optional
         Enable skip-correction for softmax rescaling (default: True).
+    uses_ldtm_stat : bool, optional
+        Use ``tcgen05.ld.red.max`` (LDTM.STAT) to fuse the per-chunk row_max
+        into the TMEM S load on the non-masked path (default: False). Requires
+        SM103+/SM110+.
     causal_single_kv_tile : bool, optional
         Use the fixed causal one-K/V-tile task domains. The context runner
         enables this only for query-paired, fixed-length inputs whose K/V
@@ -2389,6 +2393,7 @@ class FmhaTs:
         has_variable_window: bool = False,
         h_r: int = 1,
         enable_skip_correction: bool = True,
+        uses_ldtm_stat: bool = False,
         use_paged_kv: bool = False,
         num_tokens_per_page: int = 32,
         max_num_pages_per_seq_kv: int = 1,
@@ -2470,6 +2475,7 @@ class FmhaTs:
             cfg.num_regs_correction = 88
             cfg.num_regs_other = 56
         cfg.enable_skip_correction = enable_skip_correction
+        cfg.uses_ldtm_stat = uses_ldtm_stat
         cfg.qk_acc_dtype = qk_acc_dtype or cutlass.Float32
         cfg.pv_acc_dtype = pv_acc_dtype or cutlass.Float32
 
--- a/flashinfer/attention/prims_ts/kernels/fmha_context/fmha_kernel.py
+++ b/flashinfer/attention/prims_ts/kernels/fmha_context/fmha_kernel.py
@@ -2368,8 +2368,9 @@ class FmhaTs:
         Enable skip-correction for softmax rescaling (default: True).
     uses_ldtm_stat : bool, optional
         Use ``tcgen05.ld.red.max`` (LDTM.STAT) to fuse the per-chunk row_max
-        into the TMEM S load on the non-masked path (default: False). Requires
-        SM103+/SM110+.
+        into the TMEM S load on the non-masked path (default: False). The
+        context runner enables this by default on SM103 (B300) and SM107
+        (Rubin).
     causal_single_kv_tile : bool, optional
         Use the fixed causal one-K/V-tile task domains. The context runner
         enables this only for query-paired, fixed-length inputs whose K/V
--- a/flashinfer/attention/prims_ts/kernels/fmha_context/fmha_resources.py
+++ b/flashinfer/attention/prims_ts/kernels/fmha_context/fmha_resources.py
@@ -91,6 +91,7 @@ from .helpers import (
     variable_window_cta_min_start,
 )
 from cutlass.experimental import primitives as prims
+from cutlass._mlir.dialects import arith as arith_dialect
 
 SmemDescOffsets: TypeAlias = tuple[int, int]
 TmemAddr: TypeAlias = int | Int32
@@ -331,6 +332,11 @@ class FmhaConfig:
 
     seq_tile_n: int = 128
     tmem_x_load_s: int = 32
+    # Opt-in to `tcgen05.ld.red.max` (LDTM.STAT) for the non-masked
+    # `compute_row_max` path: fuses the per-chunk max into the TMEM load.
+    # Requires SM103+/SM110+ with tcgen05.ld.red support; the primitive is
+    # emitted with the 32dp x 32bit x 32rep shape only.
+    uses_ldtm_stat: bool = False
     # Causal S_q < S_kv shifts Q rows right by q_offset = S_kv - S_q.
     # This flag selects the shifted causal mask; there is no second causal mode.
     has_q_offset: bool = False
@@ -2530,6 +2536,53 @@ class TmemSPResource(MemoryResource):
         return old_row_max, row_max_safe
 
     @cute.jit
+    def _load_s_chunks_and_reduce_row_max(
+        self,
+        stage_info: StageInfo,
+        row_max: SoftmaxScalar,
+    ) -> tuple[SoftmaxScalar, SoftmaxScalar]:
+        """LDTM.STAT: fuse S load and per-chunk row_max via tcgen05.ld.red.max.
+
+        Instead of loading each S chunk with ``tcgen05.ld`` and then folding
+        a per-lane max in registers, use ``tcgen05.ld.red.max`` so the
+        reduction happens as part of the TMEM load.  Only valid on the
+        non-masked path (masked variants must observe raw S values before
+        applying the mask).
+        """
+        tmem_s_addr = self.tmem_s_addr_cached + self._stage_col_offset(stage_info)
+        tmem_x = self.cfg.tmem_x_load_s
+        num_chunks = self.cfg.qk_mma_tiler[1] // tmem_x
+        old_row_max = row_max
+        s_data: SoftmaxChunks = [None] * num_chunks
+        for chunk_idx in cutlass.range_constexpr(num_chunks):
+            loaded_words, red_word = prims.tcgen05_ld_red(
+                prims.Tcgen05LdStShape.SHAPE_32X32B,
+                prims.make_tmem_ptr(
+                    tmem_s_addr + chunk_idx * tmem_x, self.cfg.qk_acc_dtype
+                ),
+                prims.ReductionKind.MAX,
+                num=tmem_x,
+            )
+            s_data[chunk_idx] = cutlass.Vector.from_elements(
+                tuple(
+                    loaded_words[i].bitcast(self.cfg.qk_acc_dtype)
+                    for i in range(tmem_x)
+                ),
+                dtype=self.cfg.qk_acc_dtype,
+            )
+            chunk_max = self.cfg.qk_acc_dtype(
+                arith_dialect.bitcast(
+                    self.cfg.qk_acc_dtype.mlir_type, red_word.ir_value()
+                )
+            )
+            row_max = cute.math.max(row_max, chunk_max)
+        _tmem_sp_sdata[id(self)] = s_data
+        row_max_safe = row_max
+        if row_max == -Float32.inf:
+            row_max_safe = Float32(0.0)
+        return old_row_max, row_max_safe
+
+    @cute.jit
     def _exp2_p_store(
         self,
         stage_col_offset: TmemAddr,
@@ -2970,6 +3023,8 @@ class TmemSPResource(MemoryResource):
         row_max: SoftmaxScalar,
     ) -> tuple[SoftmaxScalar, SoftmaxScalar]:
         """Main K-loop stage: load S from TMEM and compute unmasked row_max."""
+        if cutlass.const_expr(self.cfg.uses_ldtm_stat):
+            return self._load_s_chunks_and_reduce_row_max(stage_info, row_max)
         s_data = self._load_s_chunks(stage_info)
         return self._reduce_row_max(s_data, row_max)
 
--- a/flashinfer/attention/prims_ts/kernels/fmha_context/fmha_resources.py
+++ b/flashinfer/attention/prims_ts/kernels/fmha_context/fmha_resources.py
@@ -2576,6 +2576,7 @@ class TmemSPResource(MemoryResource):
                 )
             )
             row_max = cute.math.max(row_max, chunk_max)
+        cute.arch.fence_view_async_tmem_load()
         _tmem_sp_sdata[id(self)] = s_data
         row_max_safe = row_max
         if row_max == -Float32.inf:
--- a/flashinfer/attention/prims_ts/context.py
+++ b/flashinfer/attention/prims_ts/context.py
@@ -35,6 +35,7 @@ register trace templates.
 
 from dataclasses import dataclass
 import functools
+import importlib.metadata
 import itertools
 import math
 import numbers
@@ -62,6 +63,8 @@ _SUPPORTED_DTYPES = (
     torch.float8_e4m3fn,
 )
 _SUPPORTED_COMPUTE_CAPABILITIES = ((10, 0), (10, 3), (10, 7))
+# tcgen05.ld.red.max (LDTM.STAT) is available on B300 (SM103) and Rubin
+# (SM107), not B200 (SM100).
 _INT32_MAX = 2**31 - 1
 _CUDA_GRID_YZ_MAX = 65_535
 _CONTEXT_KV_TILE_N = 128
@@ -215,6 +218,7 @@ def _make_context_kernel(
     has_q_offset: bool,
     causal_single_kv_tile: bool,
     scheduler: _ContextScheduler,
+    uses_ldtm_stat: bool,
     page_size: int | None = None,
     max_num_pages_per_seq_kv: int | None = None,
 ):
@@ -260,6 +264,7 @@ def _make_context_kernel(
         window_size_left=window_left if window_left > 0 else 0,
         h_r=num_qo_heads // num_kv_heads,
         enable_skip_correction=True,
+        uses_ldtm_stat=uses_ldtm_stat,
         causal_single_kv_tile=(causal_single_kv_tile and not use_paged_kv),
         **paged_kwargs,
     )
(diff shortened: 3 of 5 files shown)


kris1025:
@flashinfer-bot run

kahyunnam:
@flashinfer-bot run

aleozlx:
## PR Review Screening

**CI verdict:** ✅ auto-run ok
**Review category:** `human` (rule fired: C1.1 — `nvidia-cutlass-dsl` floor raised to a dev pre-release in three places; C3.2 durable kernel-internal change also maps to `human`)
**Blocking checks:** S2 (default PR template discarded), S3 (unmet — 0 checkboxes, no template sections)
**Release blocker:** no — opt-in feature plus a dependency floor raise; no hang/crash/corruption or prior-release regression claim (C3.4)
**Early stop:** no

### Security
| Q | Answer | Evidence |
|---|--------|----------|
| S2 template overwritten | ❗ yes | Custom `## Summary` / `## What's in this PR` / `## Safety` / `## Validation` structure; none of the five template headers present |
| S3 template obligations | ❗ unmet | 0 checkboxes; unmet formally, met in substance — the body documents tests, correctness gates, and a measured sweep |

### Packaging
| Q | Answer | Evidence |
|---|--------|----------|
| C1.1 external dependency bump | yes | `nvidia-cutlass-dsl` floor raised in three places: `requirements.txt` `>=4.6.2a0` → `>=4.8.0.dev0`, `pyproject.toml` `cu12`/`cu13` extras `>=4.7.0a0` → `>=4.8.0.dev0`, and `ci/cuda-versions.json` where `ci_image_specifier` moves from the final release `==4.7.0` to `==4.8.0.dev0`. Same package, same source — only the version floor moves. `ci/validate_cuda_versions.py` extends the grammar to accept `.devN` with documented PEP 440 ordering (`X.devN < XaN < XbN < XrcN < X`) |
| C1.2 public API changes | no — internal-only | New `FmhaConfig.uses_ldtm_stat` field and matching `FmhaTs` kwarg live under `flashinfer/attention/prims_ts/kernels/`, which is exported from neither `flashinfer/__init__.py` nor `flashinfer/attention/__init__.py` |
| C1.3 AOT registration | n-a | No `gen_*_module()` added; `aot.py` untouched; no new API requiring a `TraceTemplate` |

### Presentation
| Q | Answer | Evidence |
|---|--------|----------|
| C2.1 perf claim backed by data | yes | Exclusive B300 (SM103) node named; dense/causal × S ∈ {2048, 4096, 8192, 16384} at H=64/Hkv=8/D=128, batch 1; CUPTI medians over 100 timed iters with cold L2, OFF vs ON in separate processes; per-case tables plus means (+3.5% bf16, +6.0% fp8) and correctness deltas against the repo's own `_assert_context_correct` gate |

### Implementation
| Q | Answer | Evidence |
|---|--------|----------|
| C3.1 experimental-track declared | no | No experimental label or decorator; `flashinfer/attention/prims_ts/` is a shipped subsystem |
| C3.2 shared/durable areas touched | yes — durable | `fmha_resources.py` switches `TmemSPResource`'s non-masked `compute_row_max` to `tcgen05.ld.red.max`, fusing the per-chunk row max into the TMEM S load for every non-masked context attention on this backend; `context.py` adds `_default_uses_ldtm_stat()`, which turns the new path **on by default** for CC `(10, 3)` and `(10, 7)`. Masked paths (causal tail, sliding-window, dense-k) keep the register reduce |
| C3.3 tests match behavior change | yes | `tests/attention/test_attention_ts_context.py` (53+/3−) adds three wiring/invariant tests: kernel default off, schedule builds, and the per-device default tracking `(10, 3)` / `(10, 7)`. Numerical coverage of the ON path comes from the existing correctness tests in the same file, which flip to it automatically on SM103 runners (B300 / GB300 are in the CI matrix) — see the maintainer note on SM107 |
| C3.4 critical fix | no | Feature plus dependency bump; no hang/crash/IMA/corruption or prior-release regression claim |

### Experimental track
| Q | Answer | Evidence |
|---|--------|----------|

**Notes for the maintainer**
- The blast radius of the feature and of the dependency change are different sizes, and that asymmetry is the decision here. `uses_ldtm_stat` is opt-in and benefits two architectures; the floor raise to `>=4.8.0.dev0` applies to **every** install, and CI's image pin moves from a final release to a dev build. Worth deciding explicitly whether the floor should wait for 4.8.0 final, or whether the feature should degrade to the register-reduce path on an older DSL rather than requiring one.
- The default is enabled for **Rubin (SM107) as well as B300 (SM103)**, but validation was B300-only and the CI matrix has no SM107 runner — so on that architecture the new instruction path would ship on by default with neither measurements nor automated coverage. SM103 is genuinely covered, since the existing context correctness tests exercise the ON path there by default.
- The write-up is unusually good — per-case tables, the byte-identical claim on SM100, and an explicit note that abandoned scheduling experiments net to zero in the diff. S2/S3 fire purely on structure: the repo template was replaced rather than filled in. Cheapest fix is to keep this content under the template's own headings.

aleozlx:
all tests so far passed

aleozlx:
bumping nvidia-cutlass-dsl is probably not ok but should gate new features with a minimum version in the library

aleozlx:
assigning @qsang-nv to review and help merge PR

kris1025:
> bumping nvidia-cutlass-dsl is probably not ok but should gate new features with a minimum version in the library

But the ldtm.stat requires nvidia-cutlass-dsl 4.8.

Is there any potential issue with bumping nvidia-cutlass-dsl ?

aleozlx:
> Is there any potential issue with bumping nvidia-cutlass-dsl ?

yes there is. we should not bump it until vllm and sglang bump it because if they don't our bumping version will immediately block their adoption. and they may be blocked by other dependencies reliance on an older dsl version. 

this is why we need runtime feature guards. it is a flaw of python packaging ecosystem and bumping version here is the wrong answer

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_kernel.py
```diff
@@ -2366,6 +2366,11 @@ class FmhaTs:
         grouped-query attention with an even repeat count.
     enable_skip_correction : bool, optional
         Enable skip-correction for softmax rescaling (default: True).
+    uses_ldtm_stat : bool, optional
```
PerkzZheng, line 2369: we should automatically enable it on blackwell ultra and next.

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_resources.py
```diff
@@ -2970,6 +3023,8 @@ def compute_row_max(
         row_max: SoftmaxScalar,
     ) -> tuple[SoftmaxScalar, SoftmaxScalar]:
         """Main K-loop stage: load S from TMEM and compute unmasked row_max."""
+        if cutlass.const_expr(self.cfg.uses_ldtm_stat):
```
PerkzZheng, line 3027: this only works for dense tiles. for tiles that need masking, we should use normal LDTM + Masking + MAX reduction.  Please add corresponding tests.

requirements.txt
```diff
@@ -29,7 +29,7 @@ nccl4py>=0.3.1
 ninja
 numpy
 nvidia-cudnn-frontend>=1.25.0
-nvidia-cutlass-dsl>=4.6.2a0
+nvidia-cutlass-dsl>=4.8.0.dev0
```
PerkzZheng: we shouldn't modify package requirement directly for compatibility reasons as many repos still adopt 4.6.2.  we should only update the package version for testing. see https://github.com/Vinnie6167/flashinfer/blob/main/scripts/test_utils.sh#L297. @saltyminty @bkryu can correct me if I am wrong about the script.

qsang-nv:
**One P1 correctness issue and one blocking test issue:**

**1. Missing TMEM load fence** (`fmha_resources.py:2557-2578`)

`_load_s_chunks_and_reduce_row_max` never calls `cute.arch.fence_view_async_tmem_load()` after `tcgen05_ld_red`. True register dependencies ensure dependent instructions wait for `loaded_words`/`red_word`, but PTX does not order the async TMEM read against subsequent memory ops — and this path later writes P back into the same overlapping TMEM stage. Needs the fence before that storage is reused, like `_load_s_chunks` above it does.

**2. New schedule test cannot run** (`test_attention_ts_context.py:961`)

```python
build_fmha_task_manager(kernel.cfg)
```
`build_fmha_task_manager` needs 8 more required args (`tile_sched_params`, `tma_*_desc`, `cum_seqlen_q/k`, `num_kv_tiles` — see line 1119). This call will raise `TypeError`, so the test never actually builds the task graph.

requirements.txt
kris1025: The cutedsl version bumping is deleted now.

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_kernel.py
kris1025: It is auto enabled in context.py.

kris1025:
> > Is there any potential issue with bumping nvidia-cutlass-dsl ?
> 
> yes there is. we should not bump it until vllm and sglang bump it because if they don't our bumping version will immediately block their adoption. and they may be blocked by other dependencies reliance on an older dsl version.
> 
> this is why we need runtime feature guards. it is a flaw of python packaging ecosystem and bumping version here is the wrong answer

Thanks for the reply. The cute dsl verison bumping is deleted now. The feature will only be enabled when cute dsl verison is >= 4.8.0.

kris1025:
> **One P1 correctness issue and one blocking test issue:**
> 
> **1. Missing TMEM load fence** (`fmha_resources.py:2557-2578`)
> 
> `_load_s_chunks_and_reduce_row_max` never calls `cute.arch.fence_view_async_tmem_load()` after `tcgen05_ld_red`. True register dependencies ensure dependent instructions wait for `loaded_words`/`red_word`, but PTX does not order the async TMEM read against subsequent memory ops — and this path later writes P back into the same overlapping TMEM stage. Needs the fence before that storage is reused, like `_load_s_chunks` above it does.
> 
> **2. New schedule test cannot run** (`test_attention_ts_context.py:961`)
> 
> ```python
> build_fmha_task_manager(kernel.cfg)
> ```
> 
> `build_fmha_task_manager` needs 8 more required args (`tile_sched_params`, `tma_*_desc`, `cum_seqlen_q/k`, `num_kv_tiles` — see line 1119). This call will raise `TypeError`, so the test never actually builds the task graph.

Thanks for the review, the  two issue are solved now. Please refer to the latest commit.

kris1025:
@flashinfer-bot run

kris1025:
@flashinfer-bot run

qsang-nv:
Latest CI run (pipeline #66359573) reports 18 new failures on GB300 across both CUDA 12.9 and CUDA 13.0, all in `tests.attention.test_attention_ts_context`. The representative failure is:

```
AssertionError: assert False is True
+ where False = _default_uses_ldtm_stat(0)
```

`test_attention_ts_context_uses_ldtm_stat_default_follows_gpu` computes `expected` from device capability alone, but `_default_uses_ldtm_stat()` now additionally requires `_dsl_supports_ldtm_stat()` to succeed. On these jobs that runtime support check is evaluating to `False`, while the test still expects `True` solely because the device is SM103/SM107.

Please make the expected value reflect both parts of the documented contract. Since the CUDA 13 test setup is intended to install DSL >= 4.8.0.dev0, please also verify why the support probe is returning `False` there, and also ensure that at least one GB300 CI lane actually enables and exercises the LDTM.STAT path rather than skipping or falling back to the old implementation.

kris1025:
@flashinfer-bot run

kris1025:
@flashinfer-bot run

qsang-nv:
Three points on the latest two commits.

**1. The feature probe turns LDTM.STAT off on the CI lanes — what is the reasoning behind that choice?**

Pipeline #66559662 proved GB300 / CUDA 13.0 has no `prims.tcgen05_ld_red`, so `_dsl_supports_ldtm_stat()` now returns `False` there and `_default_uses_ldtm_stat()` short-circuits to `False`. On that lane every context attention compiles the register-reduce path, `_REQUIRES_LDTM_STAT` skips `test_..._schedule_builds`, and `test_..._default_is_off` only asserts host-side `cfg` fields without tracing. `test_..._default_follows_gpu` cannot catch it either, because `expected` is now rebuilt from the same `_dsl_supports_ldtm_stat()` the implementation calls, so it passes whether the probe is right or wrong.

The suite therefore goes green without ever executing `tcgen05.ld.red.max`, while the feature ships on by default for SM103/SM107. Why should a missing primitive silently disable the path rather than surface, and which lane is expected to run with a DSL that exposes it?

**2. The `scripts/test_utils.sh` floor no longer matches its stated purpose**

The comment justifies the cu13 bump as a "Test-only floor for LDTM.STAT (tcgen05.ld.red.max, DSL >= 4.8.0)", but after the last commit nothing gates on the version, and #66559662 shows the wheel this floor selects satisfies `>=4.8.0.dev0` yet lacks `tcgen05_ld_red`. The hunk moves the DSL for every cu13 test lane without producing the coverage it claims.

**3. Orphan comment in `context.py`**

```python
_SUPPORTED_COMPUTE_CAPABILITIES = ((10, 0), (10, 3), (10, 7))
# tcgen05.ld.red.max (LDTM.STAT) is available on B300 (SM103) and Rubin
# (SM107), not B200 (SM100).
_INT32_MAX = 2**31 - 1
```

It sits between two unrelated constants and documents neither; the capability list it describes now lives inline in `_default_uses_ldtm_stat`.

kris1025:
@flashinfer-bot run

qsang-nv:
Pls take a look at new CI failures.

kris1025:
@flashinfer-bot run

qsang-nv (approved):

PR #4842, opened 2026-08-31
feat(mla): support standalone GLM-5.3 NoPE sparse MLA on SM120/121

lucamotz:
## Summary

Extract the `GLM53_NOPE` sparse-MLA implementation currently carried in #4802 into a focused standalone change based on current `main`.

This preserves the original implementation and authorship from #4791, incorporates the useful GLM-specific H=64 / TP1 decode improvement from #4802, and avoids the unrelated parts of the broader SM120 sparse-MLA refactor.

Downstream consumer: vllm-project/vllm#53906.

## Included functionality

- Add the `GLM53_NOPE` model type and KV-cache traits.
- Model the native GLM-5.3-Flash geometry as `d_qk=512`, `D_NOPE=512`, and true `D_ROPE=0`.
- Preserve the existing 656-byte packed KV-row ABI:
  - bytes `[0:512)` contain FP8 latent values;
  - bytes `[512:528)` contain four arbitrary FP32 inline scales;
  - bytes `[528:656)` are reserved padding and are never interpreted as RoPE data.
- Route `d_qk=512` plus `kv_scale_format="arbitrary_fp32"` to `GLM53_NOPE`, while leaving the existing DSv4 route unchanged.
- Instantiate only the native top-k 2176 decode configurations:
  - H=32 / TP2;
  - H=64 / TP1.
- Support H=32 and H=64 through the existing standalone SM120 prefill path.
- Keep routing deliberately limited to the reviewed GLM shape envelope rather than adding general runtime arbitrary-head or arbitrary-top-k support.

## Deliberately excluded from #4802

- `DOTS3_SWA`;
- general runtime arbitrary-head support;
- general runtime arbitrary-top-k support;
- the cpb calibration framework;
- decode/prefill crossover calibration;
- the broad planner rewrite;
- unrelated zero-token, row-stride, and other model-family refactors.

The swapAB prefill kernel is also intentionally excluded. Porting it cleanly would require roughly 796 lines across 11 files, including new architecture and shared-memory helpers. The existing prefill path covers the required H=32 and H=64 GLM configurations and passes correctness testing, so importing that larger kernel addition is not necessary for a standalone implementation.

No technical dependency on the broader #4802 refactor remains. The H=64 specialization is the only narrowly extracted behavior derived from it.

## Testing

Validated with source JIT on GB10 / SM121:

```text
python3 -m pytest -vv -s tests/attention/test_sparse_mla_sm120.py -k glm53_nope
5 passed, 309 deselected
```

This covers:

- H=32 and H=64 decode;
- H=32 and H=64 prefill;
- arbitrary FP32 inline scales;
- top-k 2176;
- true NoPE behavior;
- exact output/LSE invariance when bytes `[528:656)` are overwritten.

Focused CPU-only routing and ABI tests:

```text
python3 -m pytest -q tests/attention/test_sparse_mla_sm120_glm53_nope.py
4 passed
```

Additional checks:

- Python byte-compilation passed.
- `git diff --check` passed.
- Source JIT explicitly instantiated `GLM53_NOPE<32, 2176>` and `GLM53_NOPE<64, 2176>` for SM121a.

Review follow-up `c5608462` narrows public routing and changes only the
GLM53_NOPE prefill page-address calculation to honor padded page strides. It
adds a focused public-API regression test, but the SM121 source-JIT suite above
predates that follow-up and was not repeated; the follow-up was checked with
Python byte-compilation, `git diff --check`, and CodeRabbit's completed
re-review.

## Provenance and credit

- The two original #4791 commits retain Luca Motz as author.
- The H=64 / TP1 decode commit credits Zihua Wu, preserving its provenance from #4802.
- The focused standalone test expansion is kept in a separate commit.

## Diffstat

```text
16 files changed, 603 insertions(+), 130 deletions(-)
```

### DIFF
--- a/csrc/sparse_mla_sm120.cu
+++ b/csrc/sparse_mla_sm120.cu
@@ -68,8 +68,8 @@ inline ModelType resolve_model_type(int d_qk, int64_t model_type) {
   if (d_qk == 512) {
     const auto mt = static_cast<ModelType>(
         model_type == kAuto ? static_cast<int64_t>(ModelType::DSV4) : model_type);
-    TVM_FFI_ICHECK(mt == ModelType::DSV4)
-        << "d_qk=512 supports only model_type auto or DSV4; got " << model_type;
+    TVM_FFI_ICHECK(mt == ModelType::DSV4 || mt == ModelType::GLM53_NOPE)
+        << "d_qk=512 supports model_type auto, DSV4, or GLM53_NOPE; got " << model_type;
     return mt;
   }
   TVM_FFI_ICHECK(false) << "Unsupported d_qk=" << d_qk
@@ -81,6 +81,7 @@ inline int bytes_per_token(ModelType mt) {
   switch (mt) {
     case ModelType::DSV3_2:
     case ModelType::GLM_NSA:
+    case ModelType::GLM53_NOPE:
       return 656;
     case ModelType::DSV4:
       return 584;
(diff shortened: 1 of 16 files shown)


lucamotz:
#4802 which carries this with authorship preserved is now close to merging, so I’m closing this as superseded.

Issue #4841, opened 2026-08-31
[SM120] trtllm cutlass fused-MoE: autotuned (non-default) tactics intermittently corrupt generation under CUDA-graph serving; 14/60 tactics presented on NVFP4 cannot run at all


SSHdotCodes:
## Summary

On SM120 (RTX PRO 6000 Blackwell Workstation), serving `RadixArk/Qwen3.8-Flash-Next-NVFP4` through SGLang with FlashInfer **0.6.18** and autotuning enabled, generations intermittently collapse into a run of `!` (token id 0) at a rate of ~4–19% of 1024-token generations, varying per boot. With `trtllm::fused_moe::gemm1/gemm2` excluded from autotuning (all other ops still tuned), the rate is 0/36. I ran a ~15-arm bisection trying to isolate a mergeable fix and could not — but the elimination matrix is now strong enough that I believe this needs someone with kernel-level visibility into the trtllm-gen/cutlass grouped-GEMM path under CUDA-graph replay. Everything below is reproducible; harnesses: https://gist.github.com/SSHdotCodes/8abd7f719194c55b841b83c8c819666a

Related: my earlier report of the same class on 0.6.17 in sgl-project/sglang#34629 (0.6.17 was 36/36 corrupt; 0.6.18 improved to ~5/36 — the occupancy pre-filter now removes some unrunnable tactics from profiling — but did not fix it).

## Environment

- RTX PRO 6000 Blackwell Workstation Edition (SM120, 96 GB), driver 595.84
- flashinfer-python/cubin 0.6.18, jit-cache 0.6.18+cu130, torch 2.13.0+cu130
- SGLang nightly `d91c3682` + the Qwen3.8-Flash-Next overlay, `--moe-runner-backend flashinfer_cutlass`, NVFP4 (`modelopt_fp4`), NEXTN MTP spec decoding, decode inside CUDA graphs
- Two fused-MoE instantiations in serving, both 512 experts / top-10 / hidden 2560 / intermediate 640:
  - target: packed int64 NVFP4 weights, BF16 activations in (in-kernel quant), `g1c=20 g2c=40`
  - MTP draft: unpacked weights, `g1c=9 g2c=9`

## Corruption phenomenology (in serving)

[...]

PR #4847, opened 2026-08-31
fix(moe): validate TMA warp-specialized configs in queryOccupancyForConfig

SSHdotCodes:
## 📌 Description

`MoeGemmRunner::queryOccupancyForConfig` returns 1 unconditionally for every TMA warp-specialized config. A TMA WS config can still be rejected at dispatch time — on SM120 NVFP4 grouped GEMM (bf16 activations × nvfp4 weights, the Qwen3.8-Flash-Next serving shape), 12 of the 60 offered tactics throw `Unsupported tile shape config 128128256 / 256128128 for MoE gemm` from the tile-shape switch and can never launch. Because occupancy reported them as runnable, `get_valid_tactics`' occupancy pre-filter passed them through, and every autotuning pass ran them into a host-side exception: profiling noise, wasted tuning time, and polluted failed-tactic statistics (they were part of what made #4841 so hard to diagnose).

This PR makes the TMA WS branch of `queryOccupancyForConfig` probe the real dispatch in workspace-query mode — the same graceful-skip idiom `calcMaxWorkspaceSize` already uses (`calcMaxWorkspaceSizeTmaWarpSpecialized` walks `dispatchMoeGemmSelectTileShapeTmaWarpSpecialized`, which throws for unsupported tiles, without launching anything) — and reports occupancy 0 when the dispatch rejects the config. Supported configs keep reporting 1, so behavior is unchanged wherever every tile shape is implemented.

Notes:
- The probe uses `EpilogueFusion::NONE` for all configs, matching `calcMaxWorkspaceSize`'s own loop (tile-shape validity is fusion-independent).
- Two further SM120 tactics fail later, at cutlass `initialize` with real problem shapes (`Failed to initialize cutlass TMA WS grouped gemm. Error: Error Internal`); those need problem-shape context and are intentionally not covered by this query-time probe — they continue to be skipped gracefully during profiling as before.
- The probe is guarded to the specializations `calcMaxWorkspaceSize` itself supports (`isValidTmaWarpSpecializedMOESpecialisation`, not w4afp8 / wfp4a16 / sm90-mixed-input); everything else keeps the previous conservative `return 1`.

## 🔍 Related Issues

Fixes the tactic-presentation defect in #4841. (The generation-corruption half of that issue was root-caused to sgl-project/sglang#36537, fixed by sgl-project/sglang#36806 — full analysis in the issue thread — and is not a FlashInfer defect.)

### ✅ Pre-commit Checks

## 🧪 Tests

Verified on SM120 hardware (RTX PRO 6000, driver 595.84) via a JIT rebuild of `fused_moe_120` with this change, against the NVFP4 config above (512 experts / top-10 / hidden 2560 / inter 640, bf16 activations, packed int64 weights):

- `get_tactic_occupancy` now returns 0 for exactly the 12 dispatch-rejected tactics — gemm1 {4, 5, 14, 15} and gemm2 {24, 25, 34, 35, 44, 45, 54, 55} — and they no longer appear in profiling. Before the change all 60 reported occupancy > 0 and the 12 threw `Unsupported tile shape config` on every tuning pass.
- The two cutlass-init failures (gemm2 {32, 33}) still surface at profiling time and are skipped there, as documented above.
- A per-tactic numeric sweep (every surviving tactic × M ∈ {1, 4} × routing, vs the default tactic) is unchanged: zero numeric findings.

## Reviewer Notes

The edit is hand-formatted to the file's existing style (Google, 100 cols); local clang-format 16 reflows unrelated lines in this file, so I left formatting to the CI pre-commit for confirmation. Happy to run any additional SM100/SM90 checks reviewers want — I only have SM120 hardware to test on, but the probe path is compile-time guarded to the same specializations `calcMaxWorkspaceSize` handles.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dispatch validation for TMA warp-specialized configurations.
  * Unsupported configurations are now rejected during occupancy checks instead of being incorrectly reported as available.
  * Preserved support for valid configurations while accounting for the configured expert count and scaling mode.
  * Improved configuration selection by preventing unavailable options from being chosen during execution.


aleozlx:
@flashinfer-bot run

aleozlx (approved):

aleozlx:
@flashinfer-bot run

PR #4849, opened 2026-08-31
feat(gemm): enable tinygemm2_sm100 on SM107 (Rubin)

Vinnie6167:
## 📌 Description

The generated SM100 `tinygemm2` variants run correctly on SM107, but three separate gates kept them off that architecture:

1. `CheckSm100Family` (`csrc/tinygemm2_sm100.cu`) hard-rejected any minor other than 0 or 3, so the kernel aborted at dispatch on `sm_107`.
2. `gen_tinygemm2_sm100_module` emitted only `compute_100a`/`compute_103a` gencode, so no `sm_107a` cubin was produced even once the check passed.
3. `_TINYGEMM2_SM100_SUPPORTED_COMPUTE_CAPABILITIES` excluded `(10, 7)`, so the router GEMM kept falling back to the reference kernel regardless.

All three are needed. Fixing only the first two leaves the kernel unreachable in production — the dispatch gate silently sends every call to the reference path.

The existing comment reserved SM107 for the reference kernel *until the generated variants were validated there*, so that is what this change rests on (see Tests).

The allowlist stays explicit rather than becoming a `major == 10` check: other 10.x devices remain unvalidated. The gencode is gated on CUDA >= 13.4 because earlier toolkits reject `compute_107a`; on those toolkits the module builds exactly as before.

## 🔍 Related Issues

None.

### ✅ Pre-commit Checks

## 🧪 Tests

`tests/model_optimizations/test_tinygemm2_sm100.py` on an SM107 device:

| | Result |
|---|---|
| Before this change | **29 failed** |
| With this change | **29 passed** |

That includes all 20 `test_tinygemm2_sm100_bitwise_parity` cases, which assert `torch.equal` against the reference kernel rather than a tolerance — the bit-identical contract the generated variants are held to on SM100/SM103.

No new test file is added: the existing suite already targets SM107 (its skip helper uses `is_sm100a_supported`, a `major == 10` predicate), so it was running and failing there rather than skipping.

Behaviour on SM100/SM103 is unchanged, and on toolkits older than CUDA 13.4 the built module is byte-for-byte what it was.

## Reviewer Notes

Point 3 is the easy one to miss — `_TINYGEMM2_SM100_SUPPORTED_COMPUTE_CAPABILITIES` lives in `flashinfer/gemm/routergemm.py`, some distance from the other two, and without it the first two changes are dead code. `test_tinygemm2_sm100_dispatch_and_escape_hatch` is the test that catches it.

## Summary by CodeRabbit

* **New Features**
  * Added TinyGEMM support for SM107 (Rubin-class) GPUs when using CUDA 13.4 or newer.
  * Enabled SM107 kernel generation with CUDA 13.4 or newer.

* **Compatibility**
  * Existing SM100 and SM103 support remains unchanged.
  * On older CUDA toolkits, SM107 operations automatically fall back to the reference kernel.

### DIFF
--- a/csrc/tinygemm2_sm100.cu
+++ b/csrc/tinygemm2_sm100.cu
@@ -2038,9 +2038,9 @@ inline void CheckSm100Family(int device_id) {
             "cudaDeviceGetAttribute(compute capability major)");
   CheckCuda(cudaDeviceGetAttribute(&minor, cudaDevAttrComputeCapabilityMinor, device_id),
             "cudaDeviceGetAttribute(compute capability minor)");
-  TVM_FFI_ICHECK(major == 10 && (minor == 0 || minor == 3))
-      << "tinygemm2_sm100 requires an SM100/SM103 (B200/B300 class) device, got sm_" << major
-      << minor;
+  TVM_FFI_ICHECK(major == 10 && (minor == 0 || minor == 3 || minor == 7))
+      << "tinygemm2_sm100 requires an SM100/SM103/SM107 (B200/B300/Rubin class) device, got sm_"
+      << major << minor;
   {
     std::lock_guard<std::mutex> lock(fam_mu);
     fam_ok.push_back(device_id);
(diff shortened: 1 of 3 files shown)


Vinnie6167:
@flashinfer-bot run

bkryu (approved):

Issue #3067, opened 2026-04-14
[CuteDSL MoE] Re-enable tile size 256(use_2cta_instrs=True)


hlu1:
>     # Only tile_size=128 is enabled. tile_size=256 (use_2cta_instrs=True)
>     # produces incorrect results in the GEMM1 gather+SwiGLU kernel and is
>     # disabled until the kernel bug is fixed.

https://github.com/nv-yunzheq/flashinfer/blob/main/flashinfer/fused_moe/cute_dsl/tuner.py#L166-L168

tile_size=256 (use_2cta_instrs=True) is now enabled in trtllm. We need to enable it in flashinfer as well and see how much perf improvements it brings.

nvpohanh:
@hlu1 could you close this given that @leejnau 's https://github.com/flashinfer-ai/flashinfer/pull/3171 has been merged? thanks

PR #4851, opened 2026-08-31
feat(moe): enable tile_size=256 for Rubin MoE autotuning

Vinnie6167:
## 📌 Description

`get_rubin_moe_valid_tactics` restricted the autotuner to `tile_size=128`, citing illegal memory accesses at `tile_size=256` with B-reuse and pointing at the equivalent restriction in `get_blackwell_moe_valid_tactics`.

**That Blackwell restriction no longer exists.** It was the gemm1(2CTA)/gemm2(1CTA) layout mismatch of #3067, fixed in #3171 by parameterizing the gemm2 candidate list on `tile_size`. `get_rubin_gemm2_valid_tactics` is already parameterized the same way, so the Rubin path carries that fix too — the restriction was mirroring a constraint that had since been lifted.

This iterates over `VALID_TILE_SIZES`, matching the Blackwell path.

The candidate lists are **unchanged**. The existing `mma_tiler_m == tile_size` and `cluster_shape_m * mma_tiler_m <= tile_size` constraints already decide what is reachable at each tile size, and the 2CTA / B-reuse entries were deliberately left in place for this — the note on them read *"kept because they become reachable again once tile_size=256 is re-enabled."* Those notes, which described the entries as unreachable, are corrected.

Net effect is a deletion: 16 insertions, 20 deletions, no new data tables or special cases.

## 🔍 Related Issues

Follows #3067 / #3171, which fixed and then re-enabled the same restriction on the Blackwell path.

### ✅ Pre-commit Checks

## 🧪 Tests

Verified on an SM107 device by autotuning `cute_dsl_fused_moe_nvfp4`, with the autotuner's per-tactic profiling traced at debug level:

- The tactic list grows from **4 to 8** entries, split `{128: 4, 256: 4}`.
- All 8 candidates profile on device and **none are skipped**, so every newly reachable tactic compiles, launches and completes.
- A sweep of **10 shapes** — 1 to 4096 tokens, 8 to 128 experts, `top_k` 2 to 8, hidden/intermediate 1024 and 2048 — completes with **no illegal memory access** (the process exits cleanly; an IMA is not catchable and would abort it).
- Output matches the non-autotuned reference in every case, cosine similarity **>= 0.999993**.

The routing-varying dimensions were chosen deliberately, since the restriction described the IMAs as routing-dependent.

No new test file: `get_rubin_moe_valid_tactics` is only reachable on `cc == (10, 7)` via `_get_arch_tactics()`, so a tile-256 assertion would skip everywhere in CI today. The existing `TestTacticEnumeration` invariants continue to pass.

## Reviewer Notes

**This is not a performance claim.** It removes a restriction that no longer applies and lets the autotuner consider tactics it could not previously reach; it asserts nothing about their speed. The autotuner continues to select whichever candidate it measures as best.

An earlier revision of this PR added four tile-256 tactics as a hand-written allowlist. That approach is dropped: those entries had `cluster_shape_m * mma_tiler_m = 512 > tile_size = 256` and so were **rejected by the generator's own validity rule** — they sat outside the constraints the module asserts. Removing the stale gate reaches the intended configurations through the existing rules instead.

## Summary by CodeRabbit

* **New Features**
  * Added support for W4A8 quantization alongside W4A4 in the fused MoE tuner.
  * Added expanded Blackwell and Rubin tactic selection, including 256-sized tile options.
  * Added validation and default tactic handling for W4A8 configurations.

* **Compatibility**
  * Preserved access to the previous runner name with a deprecation warning.

### DIFF
--- a/flashinfer/fused_moe/cute_dsl/tuner.py
+++ b/flashinfer/fused_moe/cute_dsl/tuner.py
@@ -188,20 +188,16 @@ def get_rubin_gemm1_valid_tactics(tile_size: int) -> List[Tuple]:
 
     # (mma_tiler_m, mma_inst_m) candidates — B-reuse when tiler_m = 2 * inst_m
     #
-    # NOTE: while tile_size is restricted to 128 (see get_rubin_moe_valid_tactics),
-    # the mma_tiler_m == tile_size and cluster_shape_m * mma_tiler_m <= tile_size
-    # constraints below force mma_tiler_m = 128 and cluster_shape_m = 1.  Every
-    # 2CTA and B-reuse entry here is therefore currently unreachable, and only
-    # the two 1CTA (128, 128) tactics are actually tuned.  The entries are kept
-    # because they become reachable again once tile_size=256 is re-enabled.
+    # The mma_tiler_m == tile_size and cluster_shape_m * mma_tiler_m <= tile_size
+    # constraints below decide which of these are reachable at a given tile_size;
+    # entries unreachable at one tile_size are reachable at another.
     mma_m_candidates = [
         (128, 128),  # no B-reuse, 1CTA
-        (256, 256),  # no B-reuse, 2CTA   (unreachable at tile_size=128)
-        (256, 128),  # B-reuse, 1CTA      (unreachable at tile_size=128)
-        (512, 256),  # B-reuse, 2CTA      (unreachable at tile_size=128)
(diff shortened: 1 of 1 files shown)


Vinnie6167:
@flashinfer-bot run

Vinnie6167:
@flashinfer-bot run

aleozlx (approved):

PR #4856, opened 2026-08-31
fix(fp4_gemm autotune) SM107 kernel family excluded from autotune candidates and heuristics update

### LINKED ISSUE #5065: [Bug] SM107 dense NVFP4 mm_fp4 kernel family is silently disabled — parent import from `nvidia_cutlass_dsl.examples` can never resolve
## 📌 Summary

On SM107 (Rubin), the dense NVFP4 `mm_fp4` **SM107 kernel family is silently disabled in every environment we have tested**. The SM107 kernel module fails to import, the failure is swallowed by a bare `except ImportError: pass`, `Sm107Kernel` stays `None`, and every SM107-gated code path in `gemm_base.py` quietly falls through to the non-SM107 branch. There is no warning, no log line, and no error — SM107 simply never gets its native tactics.

@Victor49152 (Mingyuan Ma) has proposed a fix in #4856. This issue tracks the underlying defect so it is visible independently of that PR, and so the packaging question below gets an explicit decision.

## 🔍 Root cause

`flashinfer/gemm/kernels/dense_blockscaled_gemm_sm107.py` imports its parent class from the CUTLASS DSL wheel's *examples* tree:

```python
# flashinfer/gemm/kernels/dense_blockscaled_gemm_sm107.py
from nvidia_cutlass_dsl.examples.CuTeDSL.cute.blackwell.kernel.blockscaled_gemm import (
    dense_blockscaled_gemm_persistent as _sm100_blockscaled_gemm,
)
```

`nvidia-cutlass-dsl` does not expose an importable `examples` namespace at all. The wheel is essentially a `.pth` redirector:

```
nvidia_cutlass_dsl_packages.pth:
  import sys, os, nvidia_cutlass_dsl
  sys.path.insert(0, os.path.join(nvidia_cutlass_dsl.__path__[0], 'dsl_packages'))
```

Its `top_level.txt` is empty, `examples` appears zero times in its `RECORD`, and the installed package contains only `cu12`, `cu13`, and `dsl_packages`. We checked the published 4.7.0 and the internal 4.8.0a0 builds — neither ships it. **This import path could never have resolved.**

## 🙈 Why it went unnoticed

Both call sites swallow the failure:

```python
# flashinfer/gemm/gemm_base.py:6399
Sm107Kernel = None
if sm_version == 107:
    try:
        from .kernels.dense_blockscaled_gemm_sm107 import (
            Sm107BlockScaledPersistentDenseGemmKernel,
        )
        Sm107Kernel = Sm107BlockScaledPersistentDenseGemmKernel
    except ImportError:
        pass
```

(`flashinfer/gemm/kernels/utils.py` does the same.) With `Sm107Kernel = None`, the guards at `gemm_base.py:6535`, `:6683` and `:6722` all take the non-SM107 path.

## 💥 Impact

- SM107 native dense NVFP4 GEMM tactics never enter the autotune shortlist — SM107 silently runs a non-native configuration.
- SM107-specific heuristic code is unreachable: with `Sm107Kernel is None`, `_compute_sm107_tactic_for_m` never executes.
- Regressed since #4526 landed (2026-08-19); no test or CI signal fires, because the fallback is a successful no-op.

## ⚠️ Packaging constraint (needs a decision)

The fix in #4856 re-points the import at FlashInfer's bundled CUTLASS sources:

```python
from flashinfer.data.cutlass.examples.python.CuTeDSL.blackwell import (...)
```

This **resolves in an editable/source install but not in a released wheel**. `flashinfer/data/` is a build-time artifact (gitignored, populated by `build_backend._create_data_dir()`), and packaging does not ship the `examples` tree:

```toml
# pyproject.toml
include-package-data = false                                       # :54
"flashinfer.data.cutlass" = ["include/**", "tools/util/include/**"] # :90
```

`examples/**` is not in package-data, and `[tool.setuptools.packages.find] include = ["flashinfer*"]` will not discover `examples/python/CuTeDSL/blackwell` as a package. Since both call sites catch `ImportError`, **wheel behaviour would be unchanged from today** — still silently disabled.

## 🛠️ Options

1. **Vendor** the base class + `scaled_mm` into `flashinfer/gemm/kernels/`, as `dense_blockscaled_gemm_sm100.py` and `_sm103.py` already do (they depend on nothing outside the wheel).
2. **Ship the single file** as package-data. It stands alone (no sibling example imports) — ~117 KB, not the full 5.4 MB tree. Still needs the intermediate packages declared, and it makes a CUTLASS *example* a runtime dependency with no API-stability guarantee.
3. **Reconcile the two `Sm100BlockScaledPersistentDenseGemmKernel` classes.** FlashInfer already defines one in `dense_blockscaled_gemm_sm100.py:50`; it is currently a strict subset of the CUTLASS example's (10 methods vs 17 — missing `_compute_stages` and the `*_copy_and_partition` helpers), which is why it cannot be swapped in as-is.

We lean toward **(3)**. cc @nv-yunzheq, who owns this file.

## ✅ Acceptance criteria

Whichever option lands, the acceptance check must run against an **installed wheel**, not a source checkout — all three options behave differently between the two, and that gap is exactly what let this sit unnoticed. Suggested assertions:

- [ ] `Sm107Kernel is not None` on an SM107 device from a pip-installed wheel.
- [ ] Native SM107 tactics appear in the dense NVFP4 autotune shortlist.
- [ ] The import failure is no longer silent — log or raise rather than `except ImportError: pass`.

## 🔗 Related

- PR #4856 — proposed fix by @Victor49152
- PR #4526 — introduced the regression

Victor49152:
## 📌 Description

1. SM107 kernel family is excluded from autotune candidates due to silenced import error, fixes added 

-   Load the SM107 dense FP4 GEMM parent from FlashInfer's bundled CUTLASS
 sources. The published `nvidia-cutlass-dsl` wheels do not package the
`nvidia_cutlass_dsl.examples` tree used by the original import.
-   Add the scheduler adapter required because the bundled Blackwell parent has
 the older `_compute_grid` helper signature, while the SM107 kernel passes
`swizzle_size` and `raster_order`.

2. Rank and cap actual FP4 tactics rather than assuming every ranked structural
  group expands to two prefetch variants. This makes the 32-tactic limit a
  real measurement budget for SM100, SM103, and SM107.
3. Retain the existing operand-orientation penalty through M=8192, but leave
  `swap_ab=False` and `swap_ab=True` neutral above M=8192 so measured
  autotuning can choose between them.
4. Remove the untuned fallback's M=4096 bucket cap. Larger power-of-two buckets
  are populated lazily so large prefill shapes do not inherit the M=4096
  analytical decision.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Reviewer Notes

 - The CUTLASS-DSL 4.8 compatibility patch was exercised on SM107: the bundled
  parent imported, the scheduler path compiled and launched, and native SM107
  tactics entered the FP4 shortlist.
 - The corrected shortlist contains 32 actual tactics. At M=16384 and M=32768,
  both matched SM107 operand orientations survive to measured autotuning.
 - The current heuristic penalizes operand swapping once \(M\) is outside the decode range. Although this bias is appropriate for small-\(M\) decode shapes, where orientation primarily affects tile utilization, it can incorrectly exclude the better orientation for large prefill shapes. At large \(M\), both orientations generally have comparable tile efficiency, and cache locality—particularly operand-tile reuse distance—can become the dominant performance factor. This bias can therefore cause regressions in prefill-heavy workloads.

### Swap microbenchmarks across M, N, and K (Justify no swap penalty should be added universally)

We first varied M across five `(N,K)` pairs: `(4096,8192)`, `(9216,4096)`,
`(8192,8192)`, `(7168,4096)`, and `(3072,8192)`. Within each pair, the SM107
tile, cluster, instruction shape, and prefetch distance were fixed; only
`swap_ab` changed. Results below report the mean advantage of the faster
orientation and the range across the five `(N,K)` pairs.

| M | Faster orientation | Hot-cache mean (range) | Cold-L2 mean (range) |
| ---: | --- | ---: | ---: |
| 4096 | `swap0` | +2.64% (+1.89% to +3.73%) | +4.37% (+2.74% to +5.68%) |
| 8192 | `swap0` | +1.46% (+0.60% to +2.58%) | +1.28% (+0.65% to +1.71%) |
| 16384 | `swap1` | +0.36% (-0.90% to +1.87%) | +3.45% (-0.98% to +7.40%) |
| 32768 | `swap1` | +39.67% (+6.84% to +66.21%) | +38.62% (+10.36% to +66.71%) |

The preference changes with M, and the cost of pruning `swap1` can become
large at M=32768. We therefore retain the established `swap1` penalty through
M=8192 but remove it above M=8192, allowing measured autotuning to decide.
The change is neutral scoring, not a hard preference for `swap1`.

We also swept N at `K=8192` while holding the native SM107 tactic fixed at
tile `256x256`, cluster `2x1`, and instruction parameters
`(256,256,128,256,0)`. Each point used 120 alternating samples per orientation
for both hot-cache and cold/HBM inputs.

| M | Cache regime | Smallest tested N favoring `swap1` | Boundary result |
| ---: | --- | ---: | ---: |
| 16384 | Hot | 3072 | +0.54%; N=3328 reverses to -1.22% |
| 16384 | Cold/HBM | 576 | +6.08%; N=512 is -5.29% |
| 32768 | Hot | 320 | +14.52%; N=256 is -3.20% |
| 32768 | Cold/HBM | 320 | +23.10%; N=256 is -3.32% |

The N dependence is non-monotonic and cannot be explained by padding alone.
For example, at M=32768, N=256 produces 128 output tiles, below one wave on a
212-SM Rubin, while padded N=320 produces 256 tiles and crosses a full-device
wave. This rules out a safe N-only threshold for scoring the swap tactics and further supports removing
the large-M penalty while leaving both orientations eligible for timing.

### Swap × raster experiment: tile reuse root cause

A follow-up controlled experiment
tested whether `swap1` is intrinsically faster or whether it changes locality
under the SM107 kernel's fixed M-major persistent scheduler. At
`N=4096, K=8192`, it held the tile (`256x256`), cluster (`2x1`), prefetch,
instruction shape, swizzle, inputs, and timing procedure fixed, and varied
only `swap_ab` and M- versus N-major raster order.

| M | M-raster / swap0 | M-raster / swap1 | N-raster / swap0 | N-raster / swap1
-- | -- | -- | -- | -- |
| 8192 | 0.00% | +3.92% | +3.46% | −0.23% |
| 16384 | 0.00% | +8.18% | +10.13% | −3.35% |
| 32768 | 0.00% | +40.60% | +40.52% | +0.12% |

Reversing raster reverses the winning swap: `swap1` wins with M-major raster,
while `swap0` wins with N-major raster. The same
reversal appeared with hot inputs and in an independent repeat on a second
Rubin node.

### Conclusion
For the production M-major scheduler, swapping rotates the grid from
`large x 16` to `16 x large`. This makes uses of the large physical activation
operand consecutive instead of revisiting them after 32, 64, or 128 tile
positions. The symmetry under raster reversal isolates shorter activation-tile
reuse distance and resulting memory locality as the cause of the large-M
benefit; it is not an intrinsic advantage of the swapped math layout.
## Commit structure

1. `fix(gemm): restore SM107 CuTe DSL kernel loading`
2. `perf(gemm): improve large-M FP4 tactic heuristics`

## Summary by CodeRabbit

- **Performance**
  - Improved FP4 GEMM tactic selection for more efficient autotuning across matrix shapes, including larger prefill workloads.
  - Refined orientation scoring to improve kernel selection for narrow and large matrix dimensions.

- **Compatibility**
  - Improved support for newer Blackwell GPU execution paths and scheduling behavior.
  - Enhanced synchronization support for integrations using the latest CuTe DSL interfaces.


Vinnie6167:
Thanks for chasing this down — the diagnosis is right and worth calling out: SM107 has been silently excluded since #4526 landed on 2026-08-19, and nobody had noticed because the failure is a swallowed `ImportError`.

One blocking problem with the fix as written, though: **it resolves in an editable/source install but not in a released wheel**, so the module stays silently disabled for anyone installing from a wheel.

The new parent import is:

```python
# flashinfer/gemm/kernels/dense_blockscaled_gemm_sm107.py:59
from flashinfer.data.cutlass.examples.python.CuTeDSL.blackwell import (...)
```

but `flashinfer/data/` is a build-time artifact (it is gitignored, and `build_backend._create_data_dir()` populates it), and the packaging config does not ship the `examples` tree:

```toml
# pyproject.toml
include-package-data = false                                       # :54
"flashinfer.data.cutlass" = ["include/**", "tools/util/include/**"] # :90
```

`examples/**` is not in package-data, and `[tool.setuptools.packages.find] include = ["flashinfer*"]` will not discover `examples/python/CuTeDSL/blackwell` as a package. Since both call sites catch `ImportError` and set `Sm107Kernel = None` (`gemm_base.py:6404`, `kernels/utils.py`), the wheel behaviour is unchanged from today.

For what it is worth, the underlying reason the *original* import never worked is that `nvidia-cutlass-dsl` does not expose an importable `examples` namespace at all. The wheel is essentially a `.pth` redirector:

```
nvidia_cutlass_dsl_packages.pth:
  import sys, os, nvidia_cutlass_dsl
  sys.path.insert(0, os.path.join(nvidia_cutlass_dsl.__path__[0], 'dsl_packages'))
```

Its `top_level.txt` is empty and `examples` appears zero times in its `RECORD`; the installed package contains only `cu12`, `cu13`, `dsl_packages`. I checked the published 4.7.0 and the internal 4.8.0a0 builds and none of them ship it — so that import path could never have resolved.

**Suggested next step:** could you hold this particular hunk rather than re-fixing it in place? There is a third option we would prefer, and it needs sign-off before anyone commits to an approach:

1. Vendor the base class + `scaled_mm` into `flashinfer/gemm/kernels/`, as `dense_blockscaled_gemm_sm100.py` and `_sm103.py` already do (they depend on nothing outside the wheel).
2. Add package-data for the single file. It stands alone — I checked its imports and it pulls in no sibling example modules — so this is ~117 KB, not the whole 5.4 MB tree. Still needs the intermediate packages declared, and it makes a CUTLASS *example* a runtime dependency, which carries no API-stability guarantee.
3. Reconcile the two `Sm100BlockScaledPersistentDenseGemmKernel` classes. flashinfer already defines one in `dense_blockscaled_gemm_sm100.py:50`; it is currently a strict subset of the CUTLASS example's (10 methods vs 17 — it is missing `_compute_stages` and the `*_copy_and_partition` helpers), which is why it cannot be swapped in as-is today.

We are leaning toward (3), but the comment asserting the in-tree class is the wrong parent is mine, from the TRT-LLM sync, so I owe an explanation of that constraint before we ask anyone to design around it. I will follow up with @nv-yunzheq, who owns this file.

Separately: whichever way this lands, the acceptance check needs to be an **installed wheel** rather than a source checkout — all three options behave differently between the two, and that gap is exactly what let this sit unnoticed.

I will leave the heuristic-side comments separately. One question that affects those: with `Sm107Kernel = None` in every environment I tested, `gemm_base.py:6683` always takes the non-SM107 branch, so `_compute_sm107_tactic_for_m` never executes. How were the SM107 heuristic changes validated?

Victor49152:
Thanks for the comments. 
Yes, I'm happy to hold on the import failure part before a path is decided.

For the heuristics change, I noticed this part because we had these sm107 kernel families available when we submit the rubin preview for mlperf, and this upstream environment has a perf gap with the mlperf environment. 
Then I basically patched this environment to get sm107 kernel families imported, re-run autotune but still find the kernel selection different. After investigation, I found with sm107 kernels, the supposed winning tactic excluded from 32 autotune candidates due to the //2 estimation, which makes less than 32 candidates entered autotune in this case (that's one fix I made in this PR). 
The mlperf env selected swap1 tactic because that version just autotune through all candidates without the 32 Cap. However, with that fix above, the winning swap1 tactic still not in the 32 autotune list.  So I digged a bit into why swap1 is ranking too low to even be an autotune candidate. That is the microbenchmark results I paste above. That is also done in a patched environment rather than what the wheel delivers. Therefore, they validate the heuristic changes and their performance motivation, but not the clean-wheel import path.

Hope that makes sense.

flashinfer/gemm/kernels/utils.py
```diff
@@ -191,7 +235,11 @@ def _select_sm100_mm_fp4_cute_dsl_tactic(m, n, real_k, sm_count, sf_vec_size):
             )
         _SM100_MM_FP4_TACTIC_CACHE[cache_key] = bucket_tactics
 
-    bucket = min(next_positive_power_of_2(m), _M_BUCKETS[-1])
+    bucket = next_positive_power_of_2(m)
+    if bucket not in bucket_tactics:
+        bucket_tactics[bucket] = _compute_tactic_for_m(
+            bucket, n, real_k, sm_count, sf_vec_size
+        )
```
Vinnie6167: My concern here is that a bucket which the tactic has not been pre-computed for (a bucket larger than 4096) will encounter a larger penalty at inference time if the user is deploying without CUDA graphs. Instead of uncapping the bucket size, can we increase the largest bucket size?

flashinfer/gemm/kernels/utils.py
```diff
@@ -156,13 +164,49 @@ def _score_sm100_mm_fp4_tactic(
...
+    # the preferred orientation, so neither orientation receives a score
+    # offset.
     rule_swap = (n % 8 == 0) and (1 <= m <= 32) and n > m
-    if swap_ab != rule_swap:
+    if m <= _MM_FP4_SWAP_PENALTY_MAX_M and swap_ab != rule_swap:
         score *= 0.95
     return score
 
 
+def _score_mm_fp4_autotune_tactic(
```
Vinnie6167: Can we drop this pure wrapper? Maybe rename _score_sm100_mm_fp4_tactic to _score_mm_fp4_tactic ?

flashinfer/gemm/kernels/utils.py
```diff
@@ -156,13 +164,49 @@ def _score_sm100_mm_fp4_tactic(
...
         score *= 1e-6
 
-    # 6. Penalize deviations from the swap rule mildly, so both swap variants stay in the top-N.
+    # 6. Penalize deviations from the established narrow-M orientation rule
+    # mildly. At large M, N-dependent launch geometry and cache state can flip
+    # the preferred orientation, so neither orientation receives a score
+    # offset.
     rule_swap = (n % 8 == 0) and (1 <= m <= 32) and n > m
-    if swap_ab != rule_swap:
+    if m <= _MM_FP4_SWAP_PENALTY_MAX_M and swap_ab != rule_swap:
```
Vinnie6167, line 189: Do you have data to support this?

Victor49152: Yes. The M-sweep in the PR description is intended to support this boundary. It fixes the SM107 tile, cluster, instruction shape, and prefetch distance and changes only swap_ab. Across five (N,K) shapes, swap0 consistently wins at M=4096 and M=8192 under both hot-cache and cold-L2 measurements. At M=16384 the preference becomes mixed and shifts toward swap1 on average, while at M=32768 swap1 wins across all tested shapes. Therefore, we retain the existing penalty through M=8192, but remove it above that point so measured autotuning can decide. This does not add a preference for swap1.

I do notice that the microbenchmarks did here are also conducted at larger N/K values, which may amplified this effect. But since we are not adding preference just removing penalty, it should be fair. Or if you have other MNK shape want to test, please let me know.

flashinfer/gemm/kernels/utils.py
Victor49152: Yes, dropped the wrapper and renamed the score function

flashinfer/gemm/kernels/utils.py
Victor49152: That's a good point. I increased bucket sizes up with three more elements. Uncapped changes are removed.

flashinfer/gemm/kernels/utils.py
Victor49152: And this change also restored and even passed the E2E perf we had for mlperf submission.

Vinnie6167:
I think we can remove the tests added in `tests/gemm/test_mm_fp4_tactic_heuristic.py`. Otherwise LGTM.

@bkryu can you review as well (I am not a codeowner).

tests/gemm/test_mm_fp4_tactic_heuristic.py
bkryu, line 1: Agree with @Vinnie6167's comment that this can probably be deleted

bkryu (commented):
Thanks @Victor49152  left one comment that is redundant with @Vinnie6167's comment.

bkryu:
@flashinfer-bot run

bkryu (approved):

PR #4859, opened 2026-09-01
feat(attention): add skip-softmax to SM120 PRIMS

lishunyang12:
## 📌 Description

Add opt-in skip-softmax to the SM120 CuTe DSL PRIMS FP8 ragged and paged prefill APIs, building on the merged #4714.

- The direct APIs accept `skip_softmax_threshold: Optional[Union[float, torch.Tensor]]`. A float supplies the final e-based threshold for every request; a contiguous CUDA float32 tensor of shape `[batch_size]` supplies one final threshold per request. There is no sequence-length normalization.
- A compute warp skips a K/V tile only when every valid row satisfies `exp((tile_max - running_max) * sm_scale) < threshold`. Skipping omits both online-softmax updates and PV MMA. The first visited tile always runs the initialization path.
- `None` selects the dense specialization. Any supplied threshold, including zero, selects the skip-enabled specialization; zero preserves dense behavior. Dynamic threshold loads follow the PDL dependency wait. Caller-owned threshold tensors can be updated in place between CUDA Graph replays.
- The D128 Q128/KV128 path uses a three-stage unified K/V TMA ring. V tiles are still transported and their pipeline barriers consumed when PV is skipped.
- Shared batch-prefill wrappers retain their existing APIs; skip-softmax through those wrappers remains unsupported for `cute-dsl-prims`.

The implementation is scoped to FP8 Q/K/V, with FP16/BF16 output. Tests cover API validation, JIT specialization, CUDA Graph/PDL, ragged/paged storage, and real skipping with output and log2 LSE checks for D64/D128/D256.

### D128 TMA pipeline optimization

One dedicated load warp prefetches the K/V tensor-map descriptors, initializes the arrived/consumed mbarriers, reduces its register allocation to 24, and issues asynchronous TMA copies into shared memory. TMA hardware transfers the tile payload; the load warp controls the copies and pipeline state. The compute warps receive 240 registers and consume the shared-memory stages.

Before this optimization, D128 used two type-fixed slots (`sK`, `sV`). After QK consumed `K0`, the producer could reload `K1` into the K slot while `V0` was still live, but startup held only `K0,V0`, `V1` remained blocked until `V0` was fully consumed, and every iteration serialized the producer through separate K-then-V consumed waits.

The optimized path treats K and V as one ordered three-slot ring. Its initial contents are logically `K0, V0, K1`; afterwards the producer waits only when it wraps around and reuses a consumed slot. For FP8 D128, one 128x128 operand tile is 16 KiB, so the ring uses 3x16 KiB = 48 KiB of shared memory. This adds K lookahead and overlaps TMA transport with QK/softmax/PV work without changing attention math.

The optimization does not yet suppress V transport for skipped tiles: tensor work and memory traffic measurements below remain effectively unchanged. The gain comes from overlap and reduced producer/consumer serialization.

### RTX PRO 5000 performance

Synthetic controlled-skip microbenchmark, not an end-to-end model claim:

| Realized tile skip fraction | Dense | Skip specialization | Speedup vs dense | max abs | relative L2 |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 0.000000 | 235.425 ms | 245.553 ms | 0.9588x | 0 | 0 |
| 0.248996 | 237.767 ms | 214.058 ms | 1.1108x | 0 | 0 |
| 0.500000 | 237.781 ms | 180.761 ms | 1.3154x | 0.00012207 | 0.000340933 |
| 0.748996 | 237.200 ms | 147.742 ms | 1.6055x | 0.000244141 | 0.00058125 |
| 0.899598 | 236.882 ms | 128.137 ms | 1.8487x | 0.000244141 | 0.000673415 |
| 0.989960 | 236.825 ms | 116.889 ms | 2.0261x | 0.000976562 | 0.00265553 |
| 0.997992 | 236.632 ms | 115.881 ms | 2.0420x | 0.00390625 | 0.00593319 |

At zero skipped tiles the specialization costs 4.1%; the measured crossover is below 25% skipped tiles. The feature remains explicitly opt-in through its threshold argument.

Configuration: SM120 RTX PRO 5000, GPU 6, driver 580.95.05, CUDA 13.0, torch 2.11.0+cu130, nvidia-cutlass-dsl 4.7.1, batch 1, ragged non-causal attention, q_len=kv_len=63744, 56 Q/KV heads, head_dim=128, Q/KV tile 128x128, FP8 E4M3 Q/K/V, BF16 output, legacy threshold scale factor 1.0 (equivalent to final `skip_softmax_threshold=1/63744` with the current API), 5 warmups, 30 repeats, warm L2, CUDA-event timing because CUPTI Python timing was unavailable.

#### Same-GPU schedule A/B

| Mode | Two-stage baseline | Three-stage | Three-stage improvement |
| --- | ---: | ---: | ---: |
| Dense | 261.852 ms | 232.138 ms | 1.128x |
| Skip specialization, no tiles skipped | 269.336 ms | 239.996 ms | 1.122x |
| Skip specialization, 99.8% tiles skipped | 176.888 ms | 115.325 ms | 1.534x |
| Active-skip speedup vs its own dense | 1.480x | 2.013x | +0.533x |

The three-stage schedule reduces active-skip latency by 34.8% relative to the two-stage implementation. The following Nsight Compute metrics are for active skip on the same GPU and shape:

| NCU metric | Two-stage | Three-stage |
| --- | ---: | ---: |
| Duration | 176.888 ms | 115.325 ms |
| SM throughput | 56.13% | 86.34% |
| Issue active | 12.99% | 22.94% |
| Eligible warps | 0.18 | 0.29 |
| Long-scoreboard stall | 5.71% | 3.87% |
| Wait stall | 22.77% | 34.71% |
| Math-pipe throttle | 28.89% | 32.91% |
| MIO throttle | 6.33% | 0.04% |
| Barrier stall | 0.01% | 0.05% |
| Tensor work ratio vs dense | 0.501 | 0.501 |
| L2 traffic | 425.148 GiB | 425.150 GiB |
| DRAM read | 1.291 GiB | 1.292 GiB |

The key changes are higher SM utilization and issue activity, more eligible warps, lower long-scoreboard stalls, and the near-elimination of MIO throttling. The higher wait-stall percentage is a share of a much shorter kernel and is not an increase in absolute runtime. Stable tensor work and traffic show that the result is a scheduling improvement, not less input data or a changed skip decision.

Measured kernel commit: `9965a38b483847f0d770bbbe72ff50bc8fc9e3b6`; fraction-sweep runner commit: `b68dd3b45d8b23ca5213579c6fd94e54e123217f`. These are historical pipeline measurements taken before the direct-threshold API and PDL updates. The current revision has been correctness-tested as described below; these performance measurements were not rerun for the test-only follow-up.

Result roots:

- three-stage benchmark: `/lustre/raplab/client/sylarl/sm120-d128-three-stage-results/20260901T043127Z`;
- three-stage NCU: `/lustre/raplab/client/sylarl/sm120-d128-three-stage-ncu/20260901T043326Z`;
- two-stage baseline NCU: `/lustre/raplab/client/sylarl/sm120-d128-two-stage-baseline-ncu/20260901T043500Z`;
- skip-fraction sweep: `/lustre/raplab/client/sylarl/sm120-d128-three-stage-fraction-sweep`.

## 🔍 Related Issues

- Builds on #4714 (merged).

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

- [x] I have run the hooks manually on changed files and fixed any reported issues.

## 🧪 Tests

- [ ] All tests are passing (`unittest`, etc.). Full upstream GPU CI still requires authorization.

Local validation on 2026-09-17: SM120 GPU (reported as NVIDIA Graphics Device, compute capability 12.0), driver 580.95.05, CUDA 13.0, PyTorch 2.13.0+cu130, nvidia-cutlass-dsl 4.7.1.

```bash
python -m pytest tests/attention/test_sm120_fmha.py \
  tests/attention/test_sm120_fmha_api.py \
  tests/attention/test_sm120_prims_prefill_backend.py -v
```

- **56 passed** across the three files; the targeted paged/causal skip selection passed all 9 cases.
- Paged real-skip tests cover D64/D128/D256, non-identity page tables, and both output and log2 LSE.
- Causal real-skip tests cover ragged and paged storage at D64/D128/D256, GQA, Q128/KV128, rectangular Q/K lengths (129/449), partial Q/K tails, and a non-initial masked-frontier skip.
- `pre-commit run --files tests/attention/test_sm120_fmha.py` and `git diff --check`: passed.

The previous upstream Test Results Summary failed because CI was skipped pending authorization, before any GPU test ran. An authorized maintainer must start public CI for the new head.

## Reviewer Notes

Please focus on the warp-uniform skip predicate, D128/D256 ring stage and phase progression, and V-barrier consumption on skipped tiles.

The new causal fixture has a bottom-right offset of 320. In its first Q CTA, rows 0–63 see an entirely masked first K tile and must initialize on the second frontier tile; rows 64–127 initialize on the first tile and skip the second while `is_masked_frontier_tile=True` and `uses_softmax_init_path=False`. Tile-specific V values and analytic retained-token LSE make real skipping observable. The extra Q row and partially populated last page cover tail masking.

### DIFF
--- a/flashinfer/attention/cute_dsl/sm120_fmha.py
+++ b/flashinfer/attention/cute_dsl/sm120_fmha.py
@@ -41,7 +41,7 @@ import math
 import os
 from functools import lru_cache
 from importlib.metadata import PackageNotFoundError, version
-from typing import Optional
+from typing import Optional, Union
 
 import torch
 from cutlass.cute.typing import Float32, Int32
@@ -167,6 +167,47 @@ def _validate_lse(q: torch.Tensor, lse: Optional[torch.Tensor]) -> None:
         raise ValueError("lse must be contiguous")
 
 
+def _prepare_skip_softmax_threshold(
+    skip_softmax_threshold: Optional[Union[float, torch.Tensor]],
+    q: torch.Tensor,
+    batch_size: int,
+) -> Optional[torch.Tensor]:
+    """Validate or expand the per-request e-based skip threshold."""
+    if skip_softmax_threshold is None:
+        return None
+    if isinstance(skip_softmax_threshold, torch.Tensor):
+        if skip_softmax_threshold.dtype != torch.float32:
+            raise ValueError(
+                "skip_softmax_threshold must have dtype torch.float32, got "
+                f"{skip_softmax_threshold.dtype}"
+            )
+        if not skip_softmax_threshold.is_cuda:
+            raise ValueError("skip_softmax_threshold must be a CUDA tensor")
+        if skip_softmax_threshold.device != q.device:
+            raise ValueError("skip_softmax_threshold and q must be on the same device")
+        expected_shape = (batch_size,)
+        if tuple(skip_softmax_threshold.shape) != expected_shape:
+            raise ValueError(
+                f"skip_softmax_threshold must have shape {expected_shape}, got "
+                f"{tuple(skip_softmax_threshold.shape)}"
+            )
+        if not skip_softmax_threshold.is_contiguous():
+            raise ValueError("skip_softmax_threshold must be contiguous")
+        return skip_softmax_threshold
+
+    try:
+        value = float(skip_softmax_threshold)
+    except (TypeError, ValueError) as exc:
+        raise TypeError(
+            "skip_softmax_threshold must be None, a Python float, or a torch.Tensor"
+        ) from exc
+    if not math.isfinite(value) or value < 0:
+        raise ValueError(
+            f"skip_softmax_threshold must be finite and >= 0; got {value!r}"
+        )
+    return torch.full((batch_size,), value, dtype=torch.float32, device=q.device)
+
+
 def _validate_hnd_paged_pool(name: str, pool: torch.Tensor) -> None:
     """Validate the HND inner layout while allowing combined-cache views."""
     if pool.ndim != 4:
@@ -203,6 +244,7 @@ def sm120_fmha_fp8_ragged_prefill(
     lse: Optional[torch.Tensor] = None,
     v_scale: Optional[float] = None,
     enable_pdl: bool = False,
+    skip_softmax_threshold: Optional[Union[float, torch.Tensor]] = None,
 ) -> None:
     """Run SM120 FP8 FMHA on packed contiguous ragged Q/K/V.
 
@@ -237,6 +279,19 @@ def sm120_fmha_fp8_ragged_prefill(
         Scalar multiplier folded into the normalized output. Defaults to 1.
     enable_pdl : bool
         Whether to enable Programmatic Dependent Launch.
+    skip_softmax_threshold : float or torch.Tensor, optional
+        Skip a K/V tile when all rows owned by a compute warp satisfy
+        ``exp((tile_max - running_max) * sm_scale) < threshold``. A Python
+        float applies one static threshold to the whole batch. During CUDA
+        Graph capture its value is fixed in the captured graph. A tensor is
+        the dynamic form: it must be a contiguous CUDA float32 tensor on the
+        same device as ``q`` with shape ``(batch_size,)``. The caller owns its
+        allocation and lifetime, must keep its address stable, and may update
+        values in place between graph replays. Tensor values are required to
+        be finite and non-negative; this content contract is not checked to
+        avoid a device-to-host synchronization. ``None`` selects the dense
+        specialization; every supplied value, including zero, selects the
+        skip-enabled specialization.
     Raises
     ------
     RuntimeError
@@ -254,6 +309,9 @@ def sm120_fmha_fp8_ragged_prefill(
 
     _check_sm120(q.device)
     _validate_lse(q, lse)
+
+    batch_size = cu_seqlens_q.numel() - 1
+    threshold = _prepare_skip_softmax_threshold(skip_softmax_threshold, q, batch_size)
 
     _, Hq, D = q.shape
     _, Hkv, D_k = k.shape
@@ -302,6 +360,7 @@ def sm120_fmha_fp8_ragged_prefill(
         device=q.device,
         with_lse=lse is not None,
         balanced_scheduler=_use_balanced_scheduler(is_causal),
+        enable_skip_softmax=threshold is not None,
     )
 
     if sm_scale is None:
@@ -317,6 +376,7 @@ def sm120_fmha_fp8_ragged_prefill(
         lse,
         scale_log2,
         output_scale,
+        threshold,
         None,
         cu_seqlens_q_i32,
         None,
@@ -347,6 +407,7 @@ def sm120_fmha_fp8_paged_prefill(
     lse: Optional[torch.Tensor] = None,
     v_scale: Optional[float] = None,
     enable_pdl: bool = False,
+    skip_softmax_threshold: Optional[Union[float, torch.Tensor]] = None,
 ) -> None:
     """Run SM120 FP8 FMHA prefill with paged K/V cache.
 
@@ -396,6 +457,19 @@ def sm120_fmha_fp8_paged_prefill(
         Scalar multiplier folded into the normalized output. Defaults to 1.
     enable_pdl : bool
         Whether to enable Programmatic Dependent Launch.
+    skip_softmax_threshold : float or torch.Tensor, optional
+        Skip a K/V tile when all rows owned by a compute warp satisfy
+        ``exp((tile_max - running_max) * sm_scale) < threshold``. A Python
+        float applies one static threshold to the whole batch. During CUDA
+        Graph capture its value is fixed in the captured graph. A tensor is
+        the dynamic form: it must be a contiguous CUDA float32 tensor on the
+        same device as ``q`` with shape ``(batch_size,)``. The caller owns its
+        allocation and lifetime, must keep its address stable, and may update
+        values in place between graph replays. Tensor values are required to
+        be finite and non-negative; this content contract is not checked to
+        avoid a device-to-host synchronization. ``None`` selects the dense
+        specialization; every supplied value, including zero, selects the
+        skip-enabled specialization.
     Raises
     ------
     RuntimeError
@@ -434,6 +508,7 @@ def sm120_fmha_fp8_paged_prefill(
         f"block_tables must be (B, max_pages), got {block_tables.shape}"
     )
     B = block_tables.shape[0]
+    threshold = _prepare_skip_softmax_threshold(skip_softmax_threshold, q, B)
 
     in_ct = _cutlass_dtype(q.dtype)
     out_ct = _cutlass_dtype(o.dtype)
@@ -475,6 +550,7 @@ def sm120_fmha_fp8_paged_prefill(
         device=q.device,
         with_lse=lse is not None,
         balanced_scheduler=_use_balanced_scheduler(is_causal),
+        enable_skip_softmax=threshold is not None,
     )
 
     if sm_scale is None:
@@ -495,6 +571,7 @@ def sm120_fmha_fp8_paged_prefill(
         lse,
         scale_log2,
         output_scale,
+        threshold,
         seqlens_kv_i32,
         cu_seqlens_q_i32,
         block_tables,
--- a/flashinfer/attention/cute_dsl/sm120_fmha.py
+++ b/flashinfer/attention/cute_dsl/sm120_fmha.py
@@ -710,6 +710,7 @@ class SM120PrimsBatchPrefillBackend:
             device=self.device,
             with_lse=False,
             balanced_scheduler=_use_balanced_scheduler(causal),
+            enable_skip_softmax=False,
         )
         self._compiled_lse_variants = {False}
 
@@ -783,6 +784,7 @@ class SM120PrimsBatchPrefillBackend:
             device=self.device,
             with_lse=False,
             balanced_scheduler=_use_balanced_scheduler(causal),
+            enable_skip_softmax=False,
         )
         self._compiled_lse_variants = {False}
 
@@ -834,30 +836,36 @@ class SM120PrimsBatchPrefillBackend:
             compile_sm120_fmha_fp8_ragged_kernel,
         )
 
-        common = (
-            self._q_dtype,
-            self._o_dtype,
-            self._num_qo_heads,
-            self._num_kv_heads,
-            self._head_dim,
-            self._causal,
-            128,
-            128,
-        )
         if self._mode == "ragged":
             compile_sm120_fmha_fp8_ragged_kernel(
-                *common,
-                self.device,
-                with_lse,
-                _use_balanced_scheduler(self._causal),
+                in_dtype=self._q_dtype,
+                out_dtype=self._o_dtype,
+                num_qo_heads=self._num_qo_heads,
+                num_kv_heads=self._num_kv_heads,
+                head_dim=self._head_dim,
+                is_causal=self._causal,
+                kv_tile=128,
+                q_tile=128,
+                device=self.device,
+                with_lse=with_lse,
+                balanced_scheduler=_use_balanced_scheduler(self._causal),
+                enable_skip_softmax=False,
             )
         else:
             compile_sm120_fmha_fp8_paged_kernel(
-                *common,
-                self._page_size,
-                self.device,
-                with_lse,
-                _use_balanced_scheduler(self._causal),
+                in_dtype=self._q_dtype,
+                out_dtype=self._o_dtype,
+                num_qo_heads=self._num_qo_heads,
+                num_kv_heads=self._num_kv_heads,
+                head_dim=self._head_dim,
+                is_causal=self._causal,
+                kv_tile=128,
+                q_tile=128,
+                num_tokens_per_page=self._page_size,
+                device=self.device,
+                with_lse=with_lse,
+                balanced_scheduler=_use_balanced_scheduler(self._causal),
+                enable_skip_softmax=False,
             )
         self._compiled_lse_variants.add(with_lse)
 
--- a/flashinfer/cute_dsl/attention/fmha/sm120/fmha_prefill_fp8_tma.py
+++ b/flashinfer/cute_dsl/attention/fmha/sm120/fmha_prefill_fp8_tma.py
@@ -2245,12 +2245,6 @@ class SM120FusedMultiHeadAttentionFP8ForwardTMA:
             q_tile_idx = block_x
             q_head_idx = block_y
         batch_idx = block_z
-        skip_softmax_threshold_log2 = None
-        if cutlass.const_expr(skip_softmax_threshold is not None):
-            skip_softmax_threshold_log2 = cute.math.log2(
-                cutlass.Float32(skip_softmax_threshold[batch_idx]),
-                fastmath=True,
-            )
         q_seq_idx = q_tile_idx * self.q_tile
 
         warp = cute.arch.make_warp_uniform(cute.arch.warp_idx())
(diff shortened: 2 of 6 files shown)


lishunyang12:
@Tom-Zheng Can you please take a look? I already rebase after your PR and retested. Thanks

flashinfer/attention/cute_dsl/sm120_fmha.py
```diff
@@ -167,6 +167,23 @@ def _validate_lse(q: torch.Tensor, lse: Optional[torch.Tensor]) -> None:
...
+        return None
+    scale_factor = float(scale_factor)
+    if not math.isfinite(scale_factor) or scale_factor < 0:
+        raise ValueError(
+            "skip_softmax_threshold_scale_factor must be finite and >= 0; "
+            f"got {scale_factor!r}"
+        )
+    if max_kv_len <= 0:
+        return None
+    return Float32(math.log2(scale_factor / max_kv_len))
```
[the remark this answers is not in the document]
lishunyang12: Fixed in 00a7df55 by computing the two logarithms separately. Added regression coverage using the smallest positive Python float with max_kv_len=256.

qsang-nv:
Deterministic skip tests only use `head_dim=64`. `uses_direct_state_ring` requires `head_tile == 128`, `uses_distributed_row_state` requires `head_tile == 256`, both also require `q_tile == kv_tile == 128` — the new/modified ring-barrier paths this PR touches. Neither is verified under a real skip:

- `test_sm120_ragged_skip_softmax_omits_negligible_tiles` / `test_ragged_public_wrapper_skip_softmax`: real skip, but `D=64`, and only check `output`, never `lse`.
- `test_ragged_public_wrapper` / `test_paged_public_wrapper`: cover causal + LSE + dtype, but use `1e-30` specifically to preserve the dense expected result — they don't demonstrate any tile was actually skipped.
- `test_skip_softmax_has_a_distinct_jit_specialization`: `D=128, q_tile=kv_tile=128`, but monkeypatches `build_and_load` — compile-only, never runs the kernel.
- `test_sm120_ragged_return_lse[d128]` / `[d256-three-stage]`: exercise the D128/D256 pipelines with LSE, but never pass a skip threshold.

No test combines `D ∈ {128, 256}, q_tile=kv_tile=128` with an actual skip and a numeric check. Since skip changes the result, verify against a skip-aware analytical value, not dense. E.g. `D=128, Sq=128, Skv=256`, `Q=K=0`, right V-tile=+1/left=-1, threshold `512/256=2`: dense → `output=0, lse=8`; skip → `output=1, lse=7`. Please add such a case (ideally also `D=256`, paged/causal) — current `D=64` coverage doesn't protect the PR's highest-risk change.

lishunyang12:
@qsang-nv Thanks for identifying the gap. Added deterministic real-skip coverage in fd7092cd. The test now explicitly uses q_tile=kv_tile=128 and covers D=64 (fixed buffer), D=128 (direct-state ring), and D=256 (distributed-state ring). Each case checks both paths analytically: dense output=0 / log2-LSE=8 and skipped output=1 / log2-LSE=7. Static checks and pre-commit pass locally; the kernel cases require the SM120 CI run.

lishunyang12:
SM120 validation completed on PR head `5b25dea1c716f37873920133b24e794e8330bd50`.

Environment:
- SM120 GPU (compute capability 12.0)
- Python 3.12.13
- `nvidia-cutlass-dsl==4.8.0.dev0` from a temporary CUDA 13 overlay

Focused regression run after the cache-key fix:

```bash
python -m pytest -vv \
  tests/attention/test_sm120_prims_prefill_backend.py::test_cuda_graph_rejects_uncompiled_specialization \
  tests/attention/test_sm120_prims_prefill_backend.py::test_prims_accepts_combined_hnd_cache_and_pdl
```

Result: `2 passed, 24 warnings in 1.04s`.

Full related test run:

```bash
python -m pytest -vv \
  tests/attention/test_sm120_fmha.py \
  tests/attention/test_sm120_prims_prefill_backend.py
```

Result: `49 passed, 44 warnings in 1.17s`.

The full run includes successful real skip-softmax coverage for all three state-storage paths requested in review:
- D64 fixed buffer
- D128 direct state ring
- D256 distributed state ring

It also passes the ragged/paged public-wrapper coverage, LSE coverage, CUDA Graph behavior, combined HND cache handling, and runtime-dynamic PDL checks.

lishunyang12:
Follow-up after history cleanup: the PR is now a single signed-off commit, `183699362b074b1f9b3f4f5ee009028e7147cfa4`.

This was a history-only squash. Both the validated pre-squash head `5b25dea1c716f37873920133b24e794e8330bd50` and the current head have the identical Git tree `161a84813eb784ee0b43af5a3cee35c29be85bb7`, so the source and test contents covered by the SM120 results above did not change.

bobboli (changes requested):

bobboli:
`skip_softmax_threshold_scale_factor` comes from the existing shared FlashInfer APIs and should be treated as a legacy convention that we will clean up separately. New kernel-specific APIs should accept the final skip-softmax threshold directly instead of propagating this convention.

This is especially relevant for DiT, where the sequence length is fixed: multiplying the desired threshold by the sequence length in the framework and then dividing it again inside FlashInfer is redundant. For variable-length batches, the framework may also need to provide a different threshold for each sequence.

Please change the new SM120 PRIMS API to accept:

```python
skip_softmax_threshold: Optional[Union[float, torch.Tensor]] = None
```

A `float` applies the same final threshold to the whole batch, while a tensor contains one final threshold per sequence with shape `[batch_size]`. The kernel should select the tensor value using `batch_idx`. How the scalar and tensor forms are normalized internally is an implementation detail.

Please also remove the sequence-length normalization from this SM120 path, along with `max_seqlen_k` / `max_seqlen_kv` where they are used only for that normalization. Converting the final threshold to log2 internally is fine, but at interface level we should make sure it's e-based rather than 2-based.

The existing `BatchPrefillWithRaggedKVCacheWrapper` and `BatchPrefillWithPagedKVCacheWrapper` APIs do not need to be migrated in this PR. These wrappers are not mandatory: for example, vLLM-Omni's [`trtllm_attn.py`](https://github.com/vllm-project/vllm-omni/blob/main/vllm_omni/diffusion/attention/backends/trtllm_attn.py) directly invokes `trtllm_ragged_attention_deepseek` instead of using the batch-prefill wrappers. Although that is currently a different kernel path, a future SM120 integration can follow the same pattern and call the new SM120-specific API directly.

To keep this PR scoped, it is fine for skip-softmax through the shared batch-prefill wrappers to remain unsupported for the SM120 PRIMS backend for now. We can migrate those shared APIs and remove the legacy `threshold_scale_factor` convention consistently in a follow-up.

lishunyang12:
Thanks for the detailed guidance. Before I update the implementation, could you please confirm a few contract details so I do not encode the wrong semantics?

1. Is `skip_softmax_threshold` intended to be the final per-tile bound in the ordinary softmax contribution domain, i.e. the kernel condition is conceptually

   `exp(tile_max - running_max) < skip_softmax_threshold`?

   In that case, should the old example `scale_factor=512, max_seqlen_k=256` map to the new API as `skip_softmax_threshold=2.0`? This changes the meaning from a sequence-normalized scale factor to an absolute per-tile threshold, so I want to confirm that this is the intended numerical contract.

2. For the tensor form, should the exact contract be a CUDA tensor of shape `[batch_size]`, with one final e-based threshold per request and selection by `batch_idx` inside the kernel? Are there requirements for dtype, device, finite/positive validation, or the semantics of `None` versus `0`?

3. For CUDA Graph replay, should the threshold tensor's address remain stable while its values may be updated between replays, or should changing threshold values require a new capture? I want the direct API and tests to reflect the expected pointer/lifetime contract.

4. When removing `max_seqlen_k` / `max_seqlen_kv`, should I remove them only from the new SM120-specific API where they serve threshold normalization, while retaining any occurrence that is still needed for launch configuration or other validation?

5. To confirm the scope: the shared `BatchPrefillWithRaggedKVCacheWrapper` and `BatchPrefillWithPagedKVCacheWrapper` should remain without skip-softmax support in this PR, and the new parameter should not be added to their trace/public signatures. The new direct SM120 APIs are the only supported path for this change, correct?

bobboli:
Yes. The intended contract is:

1. `skip_softmax_threshold` is the final threshold in the ordinary e-based softmax domain. Since the current kernel keeps unscaled QK maxima, the conceptual condition is:

   ```text
   exp((tile_max - running_max) * softmax_scale)
       < skip_softmax_threshold
   ```

   If `tile_max` and `running_max` refer to already-scaled logits, the softmax scale can be omitted. Therefore, the old `scale_factor=512, max_seqlen_k=256` case maps to:

   ```text
   skip_softmax_threshold = 512 / 256 = 2.0
   ```

2. The Python API should accept:

   ```python
   skip_softmax_threshold: Optional[Union[float, torch.Tensor]] = None
   ```

   A float applies the same final threshold to every sequence. A tensor must be a contiguous CUDA `float32` tensor on the same device as Q, with shape `[batch_size]`, and element `i` applies to request `i`.

   `None` is the only value that selects the dense kernel without skip-softmax support. Any provided float or tensor, including zero, selects the skip-enabled kernel. A zero threshold needs no special handling in that kernel: the skip condition is naturally false for every valid tile. If the threshold is converted to log2, zero corresponds to `-inf`; it must not be treated as `None`.

   I would prefer only two kernel specializations:

   - the dense kernel selected by `None`;
   - one skip-enabled kernel that consumes a per-sequence threshold tensor.

   The Python float form can be expanded to a CUDA `float32` tensor of shape `[batch_size]` before launch, so we do not need separate scalar-threshold and tensor-threshold kernel variants. Shape, dtype, device, and contiguity should be validated. Tensor values must be finite and non-negative, but content validation may remain a caller contract to avoid device-to-host synchronization.

3. Please document the CUDA Graph semantics of the two provided forms:

   - A Python float is a static batch-wide threshold. If it is used during CUDA Graph capture, its value is fixed in the captured graph and cannot be changed between replays.
   - A tensor is the dynamic form. The framework owns its allocation, lifetime, and updates, keeps its address stable, and may update its threshold values in place between replays. FlashInfer and the kernel only consume the tensor; they do not manage or update it.

   `None` versus a provided threshold is sufficient to select the dense or skip-enabled specialization at API dispatch time.

4. Please remove `max_seqlen_k` / `max_seqlen_kv` only from the new SM120-specific APIs where they exist solely for threshold normalization. Retain any sequence-length argument still required for launch configuration, indexing, masking, or validation. In particular, `max_seqlen_q` remains necessary as the launch-grid bound, and unrelated existing APIs such as `trtllm_ragged_attention_deepseek` should not be changed.

5. In this PR, SM120 PRIMS skip-softmax should be exposed only through the direct:

   ```text
   sm120_fmha_fp8_ragged_prefill
   sm120_fmha_fp8_paged_prefill
   ```

   APIs. Please do not add or forward the new threshold parameter through the shared batch-prefill wrappers or their trace signatures.

   Existing wrapper APIs should otherwise remain unchanged. In particular, `BatchPrefillWithPagedKVCacheWrapper` already has the legacy `skip_softmax_threshold_scale_factor` parameter for other backends, so it should not be removed. It can remain unsupported when that wrapper dispatches to `backend="cute-dsl-prims"`. The wrapper APIs can be migrated separately in a follow-up.

   For an initial vLLM-Omni integration, it can call `sm120_fmha_fp8_ragged_prefill` directly, bypassing the shared batch-prefill wrappers. This follows the same integration pattern as the existing [`trtllm_attn.py`](https://github.com/vllm-project/vllm-omni/blob/main/vllm_omni/diffusion/attention/backends/trtllm_attn.py), which directly calls `trtllm_ragged_attention_deepseek`. That is currently a different kernel path, but it demonstrates that the shared wrappers are not required for a kernel-specific integration.

david6666666:
Thanks for the review. I verified the threshold path already computes `log2(scale_factor) - log2(max_kv_len)` separately, avoiding underflow from dividing the raw values. I will address the remaining public-wrapper signature compatibility warning by preserving the existing argument ordering and add a focused regression check before pushing the fix.

lishunyang12:
@bobboli Thanks for the detailed contract clarification. This is addressed in the current head, `e5188fee` (the API rewrite landed in `a30c5f1e`, followed by PDL/cache hardening in `e5188fee`):

- The two direct SM120 APIs now accept the final e-based `skip_softmax_threshold: Optional[Union[float, torch.Tensor]]`. The sequence-length normalization and direct-API `max_seqlen_k` / `max_seqlen_kv` parameters are gone; `max_seqlen_q` remains only as the launch-grid bound.
- A Python float is expanded to a CUDA `float32` tensor of shape `[batch_size]`. A tensor is validated for CUDA device, `float32`, shape, same-device placement, and contiguity, then selected in the kernel by `batch_idx`. Value validation remains a caller contract to avoid synchronization.
- `None` is the only dense-kernel selector. Any supplied threshold, including zero, selects the single skip-enabled specialization; the kernel converts each request's threshold to log2 internally.
- The docstrings describe the CUDA Graph contract: floats are capture-static, while caller-owned tensors may be updated in place between replays with stable address/lifetime. The dynamic threshold load is also placed after `griddepcontrol_wait()` so PDL launches cannot observe a stale producer value.
- Skip-softmax is exposed only through `sm120_fmha_fp8_ragged_prefill` and `sm120_fmha_fp8_paged_prefill`. It is not forwarded through the shared wrapper or trace signatures. The paged wrapper keeps its legacy `skip_softmax_threshold_scale_factor` parameter and rejects it for `backend="cute-dsl-prims"`.

Validation on SM120 GPU 6: all three related test files pass (`48 passed`), including D64/D128/D256 skip paths, ragged/paged coverage, per-request threshold selection, zero-threshold behavior, CUDA Graph tensor updates, and CUDA Graph + PDL. Ruff check/format and `git diff --check` also pass.

Could you please take another look when you have a chance?

flashinfer/cute_dsl/attention/fmha/sm120/fmha_prefill_fp8_tma.py
```diff
@@ -1782,6 +1832,131 @@ def wait_pipeline_barrier(
...
+                previous_row_max = row_max[row_half]
+
+            row_wants_skip = cutlass.Boolean(False)
+            if not q_row_is_valid or tile_row_max == -cutlass.Float32.inf:
+                row_wants_skip = cutlass.Boolean(True)
+            elif previous_row_max != -cutlass.Float32.inf:
+                row_wants_skip = (
+                    (tile_row_max - previous_row_max)
+                    * softmax_params.softmax_scale_log2
+                    < softmax_params.skip_softmax_threshold_log2
```
Tom-Zheng, line 1922: Do we not divide the skip softmax threshold by seqlen now?

tests/attention/test_sm120_fmha.py
```diff
@@ -275,11 +275,161 @@ def test_sm120_ragged_empty_kv_return_lse():
     assert torch.isneginf(lse).all()
 
 
+@pytest.mark.parametrize(
+    "D",
+    [
+        pytest.param(64, id="d64-fixed-buffer"),
+        pytest.param(128, id="d128-direct-state-ring"),
+        pytest.param(256, id="d256-distributed-state-ring"),
```
Tom-Zheng, line 283: Currently, only ragged path tested D64, 128, 256. I recommend add D64, 128, 256 tests to paged kv as well.

flashinfer/cute_dsl/attention/fmha/sm120/fmha_prefill_fp8_tma.py
```diff
@@ -1782,6 +1832,131 @@ def wait_pipeline_barrier(
...
+            )
+            q_row_is_valid = (
+                basic_params.q_seq_idx + q_row_in_cta < basic_params.seqlen_q
+            )
+
+            valid_cols = basic_params.seqlen_k
+            has_valid_cols = True
+            if cutlass.const_expr(is_masked_frontier_tile):
+                if cutlass.const_expr(self.is_causal):
+                    valid_cols = cute.math.min(
```
Tom-Zheng, line 1872: Please add a real-skip causal test. Existing tests cover causal masking only in the dense path, while every test passing `skip_softmax_threshold` is non-causal. We therefore do not currently exercise the causal masking logic inside `should_skip_softmax()`. The test should preferably create multiple masked-frontier iterations so a tile reaches the predicate with
`is_masked_frontier_tile=True` and `uses_softmax_init_path=False`; the first visited tile alone is insufficient because it is always processed for softmax initialization.

flashinfer/cute_dsl/attention/fmha/sm120/fmha_prefill_fp8_tma.py
bobboli: No, please see https://github.com/flashinfer-ai/flashinfer/pull/4859#issuecomment-5530078378

bobboli (approved):

tests/attention/test_sm120_fmha.py
lishunyang12: Thanks for your suggestion. I have added tests for paged kv path.

flashinfer/cute_dsl/attention/fmha/sm120/fmha_prefill_fp8_tma.py
lishunyang12: Added. Thanks.

qsang-nv (approved):

PR #4870, opened 2026-09-01
feat(comm): copy-engine based kernel for the PCIe IPC all-reduce

qsang-nv:
## 📌 Description

Adds a copy-engine data plane to the PCIe IPC all-reduce, alongside the SM
kernels from #4393, and extends the autotuner to select between the two.

The new plane moves the payload with `cudaMemcpyAsync` on side streams and
synchronises through monotonic flags in the workspace rather than through
sentinels in the payload. It is a separate kernel family: the seven SM kernels
are unchanged, and a workspace that has not been tuned selects exactly what it
selected before.

Two schedules are added:

| variant | shape | where |
|---|---|---|
| `COPY_ENGINE_RING` | flat neighbour ring, reduce-scatter then all-gather | 4 and 8 ranks |
| `COPY_ENGINE_ISLAND` | 4+4 decomposition, one cross-socket exchange | 8 ranks, and only where the topology probe reports that grouping |

Both are reached through the existing `IpcVariant` enum and the existing tactic
encoding: `blocks` carries the ring's sub-chunk depth on these variants, and the
add kernel's thread count is fixed rather than searched.

Workspace layout gains three regions appended at the tail — flag slots, rank-local
counters, and ring staging — so every existing offset is unchanged and tuned
entries from #4393 remain valid.

### Autotuner changes

- **The bucket ladder is derived from `max_numel`** when `tune_batches` is not
  given, instead of a fixed list that stopped far below the payload a
  prefill-sized workspace admits.
- **A ranking pass before profiling.** Every candidate is already launched once
  for a zero-tolerance check against a reference; that launch is now timed, and
  only the top candidates go on to full sampling.
- **A thread-count screen above a payload threshold**, so prefill-sized tuning
  does not spend full warmup and repeat on candidates orders of magnitude off
  the winner.
- `PCIE_IPC_TUNE_VERSION` is bumped, so caches written before this change miss
  cleanly rather than resolving to a configuration that is no longer legal.

## 🔍 Related Issues

Follows #4393, which added the SM data plane this sits beside.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

- [x] I have run the hooks manually and fixed any reported issues.

## 🧪 Tests

- [x] All tests are passing.

- `tests/comm/test_pcie_ipc_ce_ring.py` — zero-tolerance correctness against
  NCCL on exact small integers, both schedules, bf16 and fp16; and survival
  across interleaved eager calls and CUDA-graph replays.
- `tests/comm/test_pcie_ipc_workspace_layout.py` — restates the layout
  arithmetic independently and pins the copy-engine region at `2*(N-1)/N` of the
  payload.
- Updated: policy, tuning and dispatch tests for the two new variants.

On 8x L40S: 81 CPU-only passed / 1 xfailed, 9 passed / 3 skipped for the
copy-engine suite, 25 passed for the SM regressions. The same suites were run
independently on an 8x sm_120 box.

## Reviewer Notes

**Performance.** bf16, `benchmarks/comm/bench_pcie_ipc_all_reduce.py`, against
the tuned SM path on the same machine. Clocks are not locked on the L40S box, so
those numbers carry a few percent of noise; NCCL reproduced its pre-existing
figures to within 1.3% across all eight buckets.

8x L40S, PCIe Gen4, `rootcplx-noswitch`:

| | payload | SM tuned | this PR | | selected |
|---|---|---|---|---|---|
| TP4, hidden 4096 | 12 MB | 1163 us | 992.7 | 1.17x | flat ring |
| | 24 MB | 2198 | 1906.9 | 1.15x | flat ring |
| | 48 MB | 4381 | 3863.4 | 1.13x | flat ring |
| | 96 MB | 8973 | 7636.0 | **1.18x** | flat ring |
| TP8, hidden 6144 | 12 MB | 1650 | 1337.2 | 1.23x | island |
| | 24 MB | 3255 | 2663.9 | 1.22x | island |
| | 48 MB | 6467 | 5387.5 | 1.20x | island |
| | 96 MB | 12819 | 10415.6 | **1.23x** | island |

8x sm_120, PCIe Gen5, `pcieswitch-pairs`, at 96 MiB: **1.71x** at four ranks and
**1.73x** at eight, measured there by that machine's owners. At two ranks the
tuner selects the SM path at every bucket.

### DIFF
--- a/benchmarks/comm/bench_pcie_ipc_all_reduce.py
+++ b/benchmarks/comm/bench_pcie_ipc_all_reduce.py
@@ -166,6 +166,10 @@ def _parse_args() -> argparse.Namespace:
     p.add_argument("--hidden", type=int, default=None)
     p.add_argument("--batches", type=str, default=None)
     p.add_argument("--dtype", choices=["bfloat16", "float16"], default="bfloat16")
+    # Lower these to look around at prefill sizes, where the default sample
+    # counts make a sweep drag; tune() then refuses to persist the result.
+    p.add_argument("--tune-warmup", type=int, default=None)
+    p.add_argument("--tune-repeat", type=int, default=None)
     p.add_argument("--json", type=str, default=None)
     p.add_argument(
         "--tune",
@@ -589,6 +593,11 @@ def main() -> None:
         group=group,
         max_numel=hidden * max(batches),
         dtype=dtype,
+        # Tune the batches this run actually sweeps. The default TUNE_BATCHES
+        # stops at 128; the bucket mapper floors, so without this a prefill
+        # sweep would replay a config measured at 128 rows on a payload three
+        # orders of magnitude larger and report it as "tuned".
+        tune_batches=batches,
         tune_cache=args.tune_cache,
     )
     # Recorded before tuning overwrites it: under --tune this shows what the
@@ -604,7 +613,12 @@ def main() -> None:
             print(f"tuning {len(batches)} batch buckets at hidden {hidden} ...")
         torch.cuda.synchronize()
         workspace.rebind_stream()
-        workspace.tune([hidden], dtype=dtype)
+        tune_kwargs = {}
+        if args.tune_warmup is not None:
+            tune_kwargs["warmup"] = args.tune_warmup
+        if args.tune_repeat is not None:
+            tune_kwargs["repeat"] = args.tune_repeat
+        workspace.tune([hidden], dtype=dtype, **tune_kwargs)
     if rank == 0:
         print(f"world_size={world_size} hidden={hidden} dtype={args.dtype}")
         print(f"profile={workspace.profile} ({workspace.profile_reason})")
--- a/csrc/pcie_ipc_all_reduce.cu
+++ b/csrc/pcie_ipc_all_reduce.cu
@@ -18,6 +18,7 @@
 #include <cstdint>
 
 #include "flashinfer/comm/pcie_ipc_all_reduce.cuh"
+#include "flashinfer/comm/pcie_ipc_ce_ring.cuh"
 #include "tvm_ffi_utils.h"
 
 namespace fi = flashinfer::comm::pcie_ipc;
@@ -36,12 +37,65 @@ namespace {
 struct PcieIpcHandle {
   fi::PeerViews views;
   fi::WorkspaceLayout layout;
+  fi::CeResources ce;
+  bool ce_ready;
   int rank;
   int world_size;
   int max_blocks;
   int64_t max_numel;
   int elem_size;
 };
+
+// One dtype arm for both data planes, so the switch below does not have to be
+// written twice as the second one grows.
+template <typename T>
+cudaError_t dispatch_one(const PcieIpcHandle* h, const T* in, T* out, int64_t numel, int blocks,
+                         int threads, fi::Variant algo, bool use_pdl, cudaStream_t stream) {
+  if (algo == fi::Variant::kCopyEngineRing) {
+    return fi::ce_ring_all_reduce_flat<T>(in, out, numel, h->views, h->rank, h->world_size,
+                                          h->layout, h->ce, blocks, threads, stream);
+  }
+  if (algo == fi::Variant::kCopyEngineIsland) {
+    return fi::ce_ring_all_reduce_island<T>(in, out, numel, h->views, h->rank, h->world_size,
+                                            h->layout, h->ce, blocks, threads, stream);
+  }
+  return fi::all_reduce<T>(in, out, numel, h->views, h->rank, h->world_size, h->max_blocks,
+                           h->max_numel, blocks, threads, algo, use_pdl, stream);
+}
+
+// Streams and events for the copy-engine ring. Non-blocking so the side streams
+// never implicitly serialise against the legacy default stream, and events with
+// timing disabled because the ring records tens of them per collective and the
+// device-side timestamp write on a timing event is not free.
+cudaError_t ce_resources_create(fi::CeResources* ce) {
+  auto stream = [](cudaStream_t* s) { return cudaStreamCreateWithFlags(s, cudaStreamNonBlocking); };
+  auto event = [](cudaEvent_t* e) { return cudaEventCreateWithFlags(e, cudaEventDisableTiming); };
+  cudaError_t err = stream(&ce->copy_stream);
+  if (err == cudaSuccess) err = stream(&ce->flag_stream);
+  if (err == cudaSuccess) err = event(&ce->input_ready);
+  if (err == cudaSuccess) err = event(&ce->copy_done);
+  if (err == cudaSuccess) err = event(&ce->flag_done);
+  for (int i = 0; err == cudaSuccess && i < fi::kCeMaxPieces; ++i) err = event(&ce->add_done[i]);
+  const int n = 2 * (fi::kMaxWorldSize - 1) * fi::kCeMaxPieces;
+  for (int i = 0; err == cudaSuccess && i < n; ++i) err = event(&ce->copied[i]);
+  return err;
+}
+
+// Events before streams: destroying a stream with queued work is legal and
+// deferred, destroying an event a queued node still references is not.
+void ce_resources_destroy(fi::CeResources* ce) {
+  // Null on the init failure path, where nullptr means the legacy default stream.
+  if (ce->copy_stream != nullptr) cudaStreamSynchronize(ce->copy_stream);
+  if (ce->flag_stream != nullptr) cudaStreamSynchronize(ce->flag_stream);
+  const int n = 2 * (fi::kMaxWorldSize - 1) * fi::kCeMaxPieces;
+  for (int i = 0; i < n; ++i) cudaEventDestroy(ce->copied[i]);
+  for (int i = 0; i < fi::kCeMaxPieces; ++i) cudaEventDestroy(ce->add_done[i]);
+  cudaEventDestroy(ce->flag_done);
+  cudaEventDestroy(ce->copy_done);
+  cudaEventDestroy(ce->input_ready);
+  cudaStreamDestroy(ce->flag_stream);
+  cudaStreamDestroy(ce->copy_stream);
+}
 
 }  // namespace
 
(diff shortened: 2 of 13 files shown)


samuellees (approved):
LGTM

qsang-nv:
# PCIe IPC copy-engine all-reduce — SOL analysis

- **CX7** — 8× L40S, PCIe Gen4, 4+4 NUMA islands (`rootcplx-noswitch`). Tuner selects the island schedule at TP8.
- **CX8** — 8× sm_120, PCIe Gen5, switch pairs (0,1)(2,3)(4,5)(6,7), halves {0-3}/{4-7} (`pcieswitch-pairs`). Tuner selects the flat ring at both world sizes.

## What the denominator is

**The time it takes to move the same bytes at the rate a bare `cudaMemcpyAsync` achieves under the same concurrency.**

- Bytes = `2(N−1)/N × S`, what every rank must send in a ring all-reduce.
- Rate = measured by a separate program that shares no code with the kernel, driving the schedule's own hop pattern at the schedule's own concurrency.

| measured rate | CX7 | CX8 |
|---|---|---|
| single hop, alone | 26.26 | 53.2–55.0 |
| intra-island ring / intra-switch pair | 25.46 (4) / 25.43 (8) | 52.33 (4) / 52.37 (8) |
| flat ring, all ranks | 20.75 | 52.32 (4) / 49.64 (8) |
| cross-boundary, all flows | 9.70 | 23.75 |

```
CX7  TP4 floor = 144 MiB / 25.46                            = 5932 us
     TP8 floor = 72 MiB/25.43 + 24 MiB/9.70 + 72 MiB/25.43  = 8532 us   (island: 3 phases)
CX8  TP4 floor = 144 MiB / 52.32                            = 2886 us
     TP8 floor = 168 MiB / 49.64                            = 3549 us   (flat ring: 1 segment)
```

## Measured

| | measured | busbw | of floor |
|---|---|---|---|
| **CX7** TP4 h4096 / h6144 | 6072 / 6062 us | 24.87 / 24.91 | **97.7% / 97.9%** |
| **CX7** TP8 h4096 / h6144 | 8643 / 8641 | 20.38 / 20.39 | **98.7% / 98.7%** |
| **CX8** TP4  | 2959 | 51.0 | **97.5%** |
| **CX8** TP8  | 4475 | 39.4 | **79.3%** |
| CX7 SM data plane | TP4 ~8995 / TP8 13067–13567 | 16.8 / 13.0–13.5 | 66% / 63–65% |
| CX8 SM data plane | TP4 6113 / TP8 12007–13776 | 24.7 / 12.8–14.7 | 47% / 26–30% |

CE over SM: **CX7 1.48× / 1.51×, CX8 2.07× / 3.08×.**

## Why the CX7 numbers are transfer-bound (nsys, rank 0 — the gating rank)

Copy-engine duty cycle **99.17%**; non-transfer work on the critical path **12.35 us**.
The add kernels have 333 us of GPU time per collective and **7.0 us of it exposed** —
at `p=2` the sibling piece's 497 us transfer covers the whole 41 us add, so they are
hidden by construction, not merely fast. Rank 3 runs at 95.49% duty with 340.8 us of
boundary idle: slack from waiting on its predecessor, not a cost of its own work.

CX8 TP4 behaves the same way: 73 us recoverable inside the kernel, hardware gap ≈ 0.

## Why CX8 TP8 falls short — and why it is not the kernel

The copy engine idles a quarter of the time (duty cycle **74.70%**, inter-hop idle averaging **81.5 us** against CX7's 3.63), but its transfers are not slow: **53.2 GB/s** per hop, *above* the 49.83 GB/s eight-rank denominator.

Device timestamps name the laggards. Six ranks take 237–242 us per outbound copy; **rank 3 takes 330.6 and rank 7 takes 516.7** — the only two whose hop crosses the half-region boundary. Solo, those same links run 53.19–53.70, no slower than any other, so this is contention, not a slow path. Every rank waits on its predecessor, so those two hops pace the ring.

The contended resource is the **PCIe switch uplink, ~48–50 GB/s shared by the two GPUs behind it**: one crossing pair gets 46.58 GB/s, a second from the same switch halves both to ~24.7, and a third from a *different* switch runs at 47.73 beside them. Four pairs aggregate to 98.5 GB/s, so the socket path is not the limit.

Against what is actually reachable — a bare ring carrying only the hop-to-hop dependency (3778 us), or the uplink capacity, 168 MiB / 48 GB/s (3670 us) — the kernel sits at **84.4%** and **82%**. The bottleneck belongs to the schedule, not the kernel: three kernel-side changes and two alternative schedules were measured and none moved the number; the island schedule, the obvious alternative, is 11% slower (4924 us).


## Conclusion

**CX7 TP4/TP8 and CX8 TP4 are at speed-of-light** — at most 1.3–2.5% of the wall clock is attributable to the kernel's own flag, add and copy-scheduling logic, against 34–37% (CX7) and 53% (CX8 TP4) for the SM data plane on the identical yardstick.

**CX8 TP8 is at 79–84%, and the shortfall is not in the kernel**: its transfers run faster than the concurrent-ring denominator while the copy engine idles a quarter of the time, waiting on the two hops per step that cross the half-region boundary, where each switch's ~48 GB/s uplink is shared by the two GPUs behind it.

aleozlx:
## PR Review Screening

**CI verdict:** ✅ auto-run ok
**Review category:** `live` (rule fired: C3.2 — a second data plane with its own synchronisation protocol, plus autotuner changes that alter how *all* candidates are ranked, not only the new ones. The PR argues containment explicitly and a maintainer may reasonably downgrade to `human` on that basis — see the notes)
**Blocking checks:** none
**Release blocker:** no — new capability plus a performance result; no hang/crash/corruption or prior-release regression claim (C3.4)
**Early stop:** no

### Security
| Q | Answer | Evidence |
|---|--------|----------|
| S3 template obligations | met | 5/5 checked and substantiated with per-suite counts on named hardware rather than asserted |

### Packaging
| Q | Answer | Evidence |
|---|--------|----------|
| C1.2 public API changes | yes — **extension of an existing shape**, additive | `PcieIpcVariant` (public, documented in `docs/api/comm.rst`) gains `COPY_ENGINE_RING = 4` and `COPY_ENGINE_ISLAND = 5`, appended so existing member values are untouched. No signature changes; the new planes are reached through the existing `IpcVariant` enum and existing tactic encoding, with `blocks` reinterpreted as the ring's sub-chunk depth on these variants only. `PCIE_IPC_TUNE_VERSION` moves 2 → 3 |
| C1.3 AOT registration | n-a | No `gen_*_module()` added; `aot.py` untouched. No new API entry point — the variants extend an interface that #4393 established, so no `TraceTemplate` obligation arises |

### Presentation
| Q | Answer | Evidence |
|---|--------|----------|
| C2.1 perf claim backed by data | yes | bf16 via `benchmarks/comm/bench_pcie_ipc_all_reduce.py`, compared against the tuned SM path on the same machine. Hardware and topology both named (8× L40S, PCIe Gen4, `rootcplx-noswitch`), eight buckets with absolute µs for both arms, the ratio, and which variant the tuner selected: 1.13–1.18× at TP4, 1.20–1.23× at TP8. A second machine (8× sm_120, PCIe Gen5, `pcieswitch-pairs`) reports 1.71×/1.73× at 96 MiB. The noise floor is stated rather than glossed — clocks unlocked on the L40S box, with NCCL reproducing its pre-existing figures to within 1.3% across all eight buckets as the control |

### Implementation
| Q | Answer | Evidence |
|---|--------|----------|
| C3.2 shared/durable areas touched | yes — **durable** | Two axes. (1) A second protocol: `include/flashinfer/comm/pcie_ipc_ce_ring.cuh` (408 new) plus changes to the existing shared header (99+/4−) and launcher (109+/10−). The SM path encodes readiness in the payload (`+0.0` means "not yet written"), which a copy engine can neither produce nor observe, so the new plane synchronises through monotonic flags on side streams — a different correctness argument living beside the old one. (2) `pcie_ipc_tuning.py` (223+/15−) changes candidate selection itself: the bucket ladder is now derived from `max_numel` rather than a fixed list, a ranking pass times the existing zero-tolerance verification launch so only top candidates reach full sampling, and a thread-count screen applies above a payload threshold |
| C3.3 tests match behavior change | yes | Five files, 546 lines. `test_pcie_ipc_ce_ring.py` checks zero-tolerance correctness against NCCL on exact small integers for both schedules in bf16 and fp16, and survival across interleaved eager calls and CUDA-graph replays; `test_pcie_ipc_workspace_layout.py` restates the layout arithmetic independently and pins the copy-engine region at `2*(N-1)/N` of the payload; `test_pinning_a_searched_dimension_moved_the_tune_version` pins the version-bump discipline itself. Guarded by `_skip_unless_runnable(world_size, variant)` and parametrised over world sizes {2,4,8}. Reported: 81 CPU-only passed / 1 xfailed, 9 passed / 3 skipped for the copy-engine suite, 25 passed for the SM regressions, on 8× L40S and independently on an 8× sm_120 box |
| C3.4 critical fix | no | New capability and a perf result; no hang/crash/IMA/corruption or prior-release regression claim |

### Experimental track
| Q | Answer | Evidence |
|---|--------|----------|

**Notes for the maintainer**
- **The containment claim is the thing to verify, and it is falsifiable.** The PR states that the seven SM kernels are unchanged and that "a workspace that has not been tuned selects exactly what it selected before". That property is what separates this from a behaviour change for existing users, and it is the reason a downgrade from `live` to `human` would be defensible. It is worth confirming directly, because the autotuner edits — the `max_numel`-derived ladder and the pre-profiling ranking pass — apply to *every* candidate including the existing SM ones, not only to the two new variants.
- **Workspace compatibility and cache invalidation point in opposite directions, deliberately.** The layout appends its three new regions at the tail so every existing offset is unchanged and #4393's tuned entries remain structurally valid — yet `PCIE_IPC_TUNE_VERSION` is bumped so those caches miss and re-tune. Both are correct together: the *workspace* is compatible, the *tuning cache* is intentionally discarded because the candidate space changed. The practical question is the first-use re-tune cost for deployments that tuned at prefill sizes, which is exactly where the ladder now reaches higher than before.
- **The strongest correctness evidence needs 4–8 ranks, so most lanes will not see it.** Zero-tolerance agreement with NCCL on exact small integers is a better bar than a tolerance comparison, and graph-replay survival is tested — but those run only where `_skip_unless_runnable` passes, leaving 81 CPU-only tests elsewhere. `COPY_ENGINE_ISLAND` is offered at world size 8 only, so confirming a lane actually exercises 8 ranks is what keeps that schedule from being validated solely by the author's two machines.

aleozlx (approved):

PR #4872, opened 2026-09-01
feat(prims-ts): support proxy-compensated block-sparse attention

heyuhhh:
## 📌 Description

This PR extends PrimTS contiguous block-sparse attention with **proxy compensation** and a packed **bitmask** sparse format.

The proxy path keeps selected blocks exact while approximating omitted KV blocks with one arithmetic-mean K summary and one summed-V summary per semantic KV block. Route preparation emits score-keep metadata so a block already covered by an exact route is not counted again by its proxy. The softmax denominator is corrected by the represented token mass, including the true length of the final partial KV block.

### Sol-Attn motivation and scope

This PR is intended to provide the PrimTS **core-attention** substrate for [Sol-Attn's sparse-attention flow](https://github.com/NVlabs/Sana/tree/sol-engine/techniques/sparse_backends), described in the [Sol-Attn paper](https://arxiv.org/abs/2607.24027). Sol-Attn identifies critical KV blocks for exact computation and reuses block-level proxy information from omitted blocks to approximate their contribution during online softmax.

At the API boundary, the intended flow is:

`prediction / routing (out of scope) → exact-block BSR or bitmask + K/V summaries → PrimTS route preparation → proxy-compensated core attention`

This PR implements only the PrimTS route-preparation and core-attention stages. It does **not** include the Sol-Attn predictor, thresholding and exact-block selection, summary generation, or the end-to-end Sana integration; callers provide the selected-block metadata and K/V summaries.

### Bitmask format

In addition to canonical BSR `block_indptr` / `block_indices`, contiguous block-sparse attention can now consume packed `UInt32` exact-block bitmaps owned per `(batch, KV head, Q block)`. Each bit selects one semantic KV block and out-of-range padding bits in the final word are ignored. BSR and bitmask inputs are prepared into the same internal route stream, so the attention core remains format-independent.

Bitmask and proxy modes can be combined: set bits use raw K/V exact routes, while unset blocks are represented by proxy summaries. Both reusable plan/run and one-shot contiguous APIs are supported. Proxy execution currently requires a dense mask; paged-KV proxy execution is outside this PR.

The PR also teaches Trace Apply to honor a runtime-selected trace template, which keeps BSR/bitmask and exact/proxy schemas isolated instead of binding every call to template zero.

## 📊 Proxy overhead

Controlled A/B on one NVIDIA B200 with identical inputs, exact-block pattern, capacity, and static scheduler; only `use_proxy_routes` and the required summaries change.

| Path | Proxy disabled | Proxy enabled | Additional overhead |
| --- | ---: | ---: | ---: |
| Public `BlockSparseTSWrapper.run` | 1.3172 ms | 1.4298 ms | **+8.55%** |
| Prevalidated adapter | 1.3162 ms | 1.4318 ms | **+8.78%** |

Workload: BF16 Q/K/V `[1, 32760, 12, 128]`, Q-block 64, KV-block 64, physical KV route 256, dense bitmask, and 84 of 512 exact blocks per row (16.4%). Proxy adds two prepared routes per row (21 → 23). Results use five balanced rounds with 80 CUDA-event samples per arm (400 samples/arm). Timings include route preparation and attention; summary construction is excluded. Each mode was checked against its own independent reference because proxy compensation intentionally changes the output semantics.

Kernel split diagnostic: attention p50 increased by 7.64% (1.1261 → 1.2121 ms); prepare increased by 0.0030 ms (0.0145 → 0.0176 ms).

## 🧪 Tests

- [x] `tests/trace_apply/test_trace_apply.py`: 26 passed
- [x] `tests/trace/test_fi_trace_template_consistency.py`: 774 passed
- [x] `tests/attention/test_attention_ts_block_sparse.py`: 140 passed, 88 skipped
- [x] BK8 SWAPS and BK64 Keeps GPU proxy validation passed for both BSR and bitmask
- [x] BSR/bitmask results are bitwise identical in the focused GPU cases; final-word padding bits are ignored
- [x] Repository pre-commit hooks passed, including mypy, Ruff check, and Ruff format

### ✅ Pre-commit Checks

- [x] I have run the hooks manually and fixed all reported issues.

### ✅ Tests

- [x] All targeted tests are passing.

## Reviewer Notes

The commits are separated into generic runtime trace dispatch, block-sparse functionality, focused correctness/trace tests, and review-driven cleanup. `experimental/sol_attention` is not part of this PR.

## Summary by CodeRabbit

* **New Features**
  * Added BSR and bitmask routing options for block-sparse attention.
  * Added proxy-route support using caller-provided K/V summaries.
  * Added route-specific validation and metadata handling for wrapper and one-shot APIs.
  * Expanded tracing to support all sparse format and route-mode combinations.
  * Improved trace application to select and execute only the runtime-matched template.

* **Documentation**
  * Documented contiguous bitmask routing, proxy routes, capacity semantics, and summary behavior.

* **Tests**
  * Added GPU coverage for BSR and bitmask proxy routes and multi-template trace dispatch.

## Current integration stack (2026-09-04)

- This branch is rebased onto FlashInfer `main` `60b49158ab4fb81718aef486c2d3c89aec4c1901`; current PR head: `6b5b2d31f3314be77ccf6655c899b7b971863a1a`.
- The TRT-LLM composite pin is [heyuhhh/flashinfer@`71bf7842`](https://github.com/heyuhhh/flashinfer/tree/yuhangh/tmp-sol-attn-trtllm-pr17399), which reapplies all eight commits from Yuxian's newest `trtllm-prims-ts` head (`edddf6f5`) on top of this PR.
- TensorRT-LLM stack: [#17399 PrimTS base](https://github.com/NVIDIA/TensorRT-LLM/pull/17399) → [general block-sparse FMHA](https://github.com/yuxianq/TensorRT-LLM/pull/2) → [VisualGen VSA/shared workflow](https://github.com/heyuhhh/TensorRT-LLM/pull/7) → [VisualGen SOL integration](https://github.com/heyuhhh/TensorRT-LLM/pull/9).
- Final B200 integration verification includes raw/proxy/paged generic routes, VSA CUDA Graph live routes, SOL all-exact parity, and mixed proxy parity at sequence lengths 256 and 257.

### DIFF
--- a/flashinfer/attention/prims_ts/_block_sparse/common.py
+++ b/flashinfer/attention/prims_ts/_block_sparse/common.py
@@ -22,6 +22,20 @@ _BLOCK_SPARSE_Q_TILE_SIZES = (8, 16, 32, 64, 128)
 _BLOCK_SPARSE_MAX_HEADS_Q_PER_KV = 32
 
 
+def _validate_contiguous_route_mode(
+    sparse_format: object,
+    use_proxy_routes: object,
+) -> None:
+    """Validate the two public continuous-route axes before device work."""
+
+    if not isinstance(sparse_format, str):
+        raise TypeError("sparse_format must be 'bsr' or 'bitmask'")
+    if sparse_format not in ("bsr", "bitmask"):
+        raise ValueError("sparse_format must be 'bsr' or 'bitmask'")
+    if type(use_proxy_routes) is not bool:
+        raise TypeError("use_proxy_routes must be a bool")
+
+
 def _validate_sparse_q_block_size(value: object) -> int:
     """Return a positive semantic Q block size representable by the ABI."""
 
@@ -110,6 +124,17 @@ def _block_sparse_kv_atom_size(kv_block_size: int) -> int:
     )
 
 
+def _block_sparse_proxy_summary_geometry(
+    seq_len_kv: int,
+    kv_block_size: int,
+) -> tuple[int, int]:
+    """Return the summary count and final summary's represented-token mass."""
+
+    num_summaries = (seq_len_kv + kv_block_size - 1) // kv_block_size
+    tail_mass = seq_len_kv - (num_summaries - 1) * kv_block_size
+    return num_summaries, tail_mass
+
+
 def _prepared_kv_routes_are_block_aligned(
     kv_block_size: int,
     kv_route_size: int,
@@ -117,3 +142,24 @@ def _prepared_kv_routes_are_block_aligned(
     """Return whether each prepared route stays within one semantic BSR block."""
 
     return _validate_sparse_kv_block_size(kv_block_size) % kv_route_size == 0
+
+
+def _block_sparse_contiguous_kv_copy_geometry(
+    *,
+    kv_block_size: int,
+    kv_route_size: int,
+) -> tuple[int, int, bool]:
+    """Return source-independent primary/atom TensorMap geometry.
+
+    Exact and proxy routes address different logical matrices, but a route's
+    physical copy shape depends only on its semantic block and physical route
+    sizes. Coarse routes prefer KV128 copies and keep a KV64 descriptor only
+    when KV256 staging or runtime adjacency requires it.
+    """
+
+    atom_size = _block_sparse_kv_atom_size(kv_block_size)
+    primary_box_size = 2 * atom_size if atom_size == 64 else atom_size
+    needs_aux_atom = atom_size == 64 and (
+        kv_route_size == 256 or kv_block_size % kv_route_size != 0
+    )
+    return primary_box_size, atom_size, needs_aux_atom
--- a/flashinfer/attention/prims_ts/_block_sparse/compiler.py
+++ b/flashinfer/attention/prims_ts/_block_sparse/compiler.py
@@ -36,31 +36,45 @@ def _compile_block_sparse(key: _BlockSparseCompileKey) -> Callable[..., object]:
 
     from ..kernels.fmha_decode.fmha_decode_config import FmhaDecodeConfig
     from ..kernels.fmha_decode.block_sparse_prepare import (
-        _PrepareBlockSparseRoutes,
+        _PrepareBitmaskRoutes,
+        _PrepareBsrRoutes,
     )
     from ..kernels.fmha_decode.fmha_decode_kernel import (
         fmha_block_sparse_launch,
     )
 
     config = _make_block_sparse_config(key)
-    prepare_routes = _PrepareBlockSparseRoutes(
-        batch_size=key.batch_size,
-        num_kv_heads=key.num_kv_heads,
-        seq_len_q=key.seq_len_q,
-        seq_len_kv=key.seq_len_kv,
-        q_block_size=key.q_block_size,
-        kv_block_size=key.kv_block_size,
-        kv_route_size=key.kv_route_size,
-        has_token_bits=key.use_kv_valid_bits,
-        page_size=key.page_size,
-        mask_type=key.mask_type,
-    )
+    prepare_kwargs = {
+        "batch_size": key.batch_size,
+        "num_kv_heads": key.num_kv_heads,
+        "seq_len_q": key.seq_len_q,
+        "seq_len_kv": key.seq_len_kv,
+        "q_block_size": key.q_block_size,
+        "kv_block_size": key.kv_block_size,
+        "kv_route_size": key.kv_route_size,
+        "use_proxy_routes": key.use_proxy_routes,
+        "use_causal_mask": key.mask_type == "causal",
+        "apply_token_mask": key.use_kv_valid_bits,
+    }
+    if key.page_size is not None:
+        if key.sparse_format != "bsr" or key.use_proxy_routes:
+            raise AssertionError("paged block-sparse supports exact BSR routes only")
+        prepare_kwargs["page_size"] = key.page_size
+
+    if key.sparse_format == "bsr":
+        prepare_routes = _PrepareBsrRoutes(**prepare_kwargs)
+    elif key.sparse_format == "bitmask":
+        prepare_routes = _PrepareBitmaskRoutes(**prepare_kwargs)
+    else:
+        raise AssertionError("sparse_format must be 'bsr' or 'bitmask'")
+
+    route_metadata_base = prepare_routes.route_metadata_base_word_offset
     Int32 = cutlass.Int32
     Int64 = cutlass.Int64
     Float32 = cutlass.Float32
 
     @cute.jit
-    def contiguous_tensor_adapter(
+    def exact_bsr_adapter(
         q: cute.Tensor,
         k: cute.Tensor,
         v: cute.Tensor,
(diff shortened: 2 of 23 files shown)


flashinfer/attention/prims_ts/_block_sparse/common.py
```diff
@@ -110,10 +124,42 @@ def _block_sparse_kv_atom_size(kv_block_size: int) -> int:
...
 
 
+def _block_sparse_proxy_summary_geometry(
+    seq_len_kv: int,
+    kv_block_size: int,
+) -> tuple[int, int]:
+    """Return the summary count and final summary's represented-token mass."""
+
+    num_summaries = (seq_len_kv + kv_block_size - 1) // kv_block_size
+    tail_mass = seq_len_kv - (num_summaries - 1) * kv_block_size
```
PerkzZheng, line 134: we always assume there is a tail even if seq_len_kv is divisble by kv block size, right ?

flashinfer/attention/prims_ts/_block_sparse/common.py
```diff
@@ -110,10 +124,42 @@ def _block_sparse_kv_atom_size(kv_block_size: int) -> int:
...
+    """Return source-independent primary/atom TensorMap geometry.
+
+    Exact and proxy routes address different logical matrices, but a route's
+    physical copy shape depends only on its semantic block and physical route
+    sizes. Coarse routes prefer KV128 copies and keep a KV64 descriptor only
+    when KV256 staging or runtime adjacency requires it.
+    """
+
+    atom_size = _block_sparse_kv_atom_size(kv_block_size)
+    primary_box_size = 2 * atom_size if atom_size == 64 else atom_size
```
PerkzZheng, line 161: can you explain a bit more about this ?

flashinfer/attention/prims_ts/_block_sparse/common.py
```diff
@@ -22,6 +22,20 @@
...
+def _validate_contiguous_route_mode(
+    sparse_format: object,
+    use_proxy_routes: object,
+) -> None:
+    """Validate the two public continuous-route axes before device work."""
+
+    if not isinstance(sparse_format, str):
+        raise TypeError("sparse_format must be 'bsr' or 'bitmask'")
+    if sparse_format not in ("bsr", "bitmask"):
+        raise ValueError("sparse_format must be 'bsr' or 'bitmask'")
```
PerkzZheng, line 34: what "bsr" means here ? boolean flags ? it is possible to only keep bitmask ?

flashinfer/attention/prims_ts/_block_sparse/compiler.py
```diff
@@ -119,6 +133,169 @@ def contiguous_tensor_adapter(
...
+        exact_block_bits: cute.Tensor,
+        kv_valid_bits: cute.Tensor,
+        row_route_offsets: cute.Tensor,
+        route_workspace: cute.Tensor,
+        max_blocks_per_row: cutlass.Int32,
+        sm_scale: cutlass.Float32,
+        stream: cuda_drv.CUstream,
+        static_config: cutlass.Constexpr[FmhaDecodeConfig],
+        static_batch_size: cutlass.Constexpr[int],
+        static_seq_len_kv: cutlass.Constexpr[int],
```
PerkzZheng, line 151: it seems that we use static seqlen_kv and batch size here ?

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/tmem_s.py
```diff
@@ -1906,19 +1924,20 @@ def _compute_softmax_loop_swaps(
...
-                    if cutlass.const_expr(not token_word_covers_kv_tail):
-                        score_is_valid = cutlass.Boolean(
-                            score_is_valid and logical_k < seq_len_kv
-                        )
-                    if cutlass.const_expr(cfg.uses_uniform_causal_mask):
-                        score_is_valid = cutlass.Boolean(
-                            score_is_valid and logical_k < element_mask_end_idx
-                        )
-                    if cutlass.const_expr(cfg.use_kv_valid_bits):
+                    if not route_is_proxy:
```
PerkzZheng, line 1903: we always assume proxy ones are valid ? could there be paddings ?

flashinfer/attention/prims_ts/_block_sparse/common.py
heyuhhh: Right, it is an optimization point but will introduce one flag, i'll look at here further to see the number of perf gains. Thanks!

flashinfer/attention/prims_ts/_block_sparse/common.py
heyuhhh: We keep an optimization which merges two adjacent blocks of size 64 to one 128 TMA load. So firstly we should get atom_size of TMA load (8, 16, 32, 64) and determine the TMA load granularity. Here is also a flag `needs_aux_atom` means whether we should prepare auxilary TMA descriptor.

For example, if block_size=64 we will have one primary TMA descriptor with size 128 and also an auxilary TMA descriptor with size 64. We will choose to use which one at runtime. 

P.S. This is only used for block_size=64 and tile_size_kv=128. For tile_size_kv=256 we always use TMA64

PerkzZheng: > We keep an optimization which merges two adjacent blocks of size 64 to one 128 TMA load. So firstly we should get atom_size of TMA load (8, 16, 32, 64) and determine the TMA load granularity. Here is also a flag `needs_aux_atom` means whether we should prepare auxilary TMA descriptor.
> 
> For example, if block_size=64 we will have one primary TMA descriptor with size 128 and also an auxilary TMA descriptor with size 64. We will choose to use which one at runtime.
> 
> P.S. This is only used for block_size=64 and tile_size_kv=128. For tile_size_kv=256 we always use TMA64

how can we merge two sparse blocks into one TMA load ? or you find adjacent ones from the perspective of software.

flashinfer/attention/prims_ts/_block_sparse/common.py
heyuhhh: BSR means block sparse format, which has `block_sparse_indptr` and `block_sparse_indices`. It can be derived from topk results easily.
BITMASK means naive bitmask map, shape of it is like [num_q_blocks, num_kv_heads, num_kv_blocks // 32], which can be derived from topp results easily.

BITMASK seems more intuitive but will consume more memory, for the shape determined operations like topk we should choose BSR i think.

flashinfer/attention/prims_ts/_block_sparse/compiler.py
heyuhhh: Right, we don't support paged kv for now. In the AIGV domain there is always a static seq_lens

flashinfer/attention/prims_ts/_block_sparse/common.py
PerkzZheng: > BITMASK seems more intuitive but will consume more memory'

curious why it would consume more memory or you mean it needs additional memory except for block_sparse_indices.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/tmem_s.py
heyuhhh: It is almost valid because if all proxy blocks are skipped, we should still compute this item for compensation. However, if all proxy blocks are choosed we can skip it but in real sparse patterns i don't think it happens frequently: we have 64 block size and 256 kv tile size, we can skip which means we choose continous 256 kv blocks, about 16k tokens. So we don't need to add an extra conditional branch here

flashinfer/attention/prims_ts/_block_sparse/common.py
heyuhhh: Yes, from perspective of software

flashinfer/attention/prims_ts/_block_sparse/common.py
heyuhhh: BSR only stores the information of valid blocks: how many valid blocks in each row and what's the column of them
For BITMASK we don't need BSR format, it's just a 2D mask map

Before attention there is a prepare kernel to transform these two formats to one standard route level format consumed by kernel

Maybe we can unify this later? When observed some convergence trend. However, in the community most of kernels use BSR as far as i know.

flashinfer/attention/prims_ts/_block_sparse/common.py
PerkzZheng: thanks. that makes sense.

flashinfer/attention/prims_ts/_block_sparse/common.py
PerkzZheng: thanks.

flashinfer/attention/prims_ts/_block_sparse/compiler.py
PerkzZheng: thanks.

flashinfer/attention/prims_ts/_block_sparse/common.py
heyuhhh: I let my agent do a quick try about this and the result is there is no perf gains: if tail_mass == kv_block_size so tail_delta == 0 and can skip the following compensated code block. Also the compiler can help to reduce some overheads when seq_len_kv%kv_block_size==0

BTW, we don't assume there is a divisible relation. If tail block has padding tokens we will substract them from results

qsang-nv (approved):

heyuhhh:
Hi @saltyminty @qsang-nv , the CI has passed and could you help merge this PR? Thanks!

PR #4877, opened 2026-09-01
docs(tests): use SM120/SM121 in XQA MLA skip strings

aeichler-ac:
## Summary
- Align XQA MLA trace-test skip/doc strings from `SM120/121` to `SM120/SM121`.

The original Spark-audit PR also staged `copy2` for monomoe/rmsnorm_silu; that
landed on main via #5063, so those hunks were dropped on rebase.

Refs #3170

## Test plan
- [x] Diff vs main is the two trace-test string files only
- [ ] CI after authorized `@flashinfer-bot run`

### DIFF
--- a/tests/trace/test_xqa_batch_decode_mla_reference_correctness.py
+++ b/tests/trace/test_xqa_batch_decode_mla_reference_correctness.py
@@ -24,12 +24,12 @@ from tests.trace.reference_utils import (
     ],
 )
 def test_xqa_batch_decode_mla_reference_correctness(shape_kwargs):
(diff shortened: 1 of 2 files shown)


yongwww:
@flashinfer-bot run

aeichler-ac:
Most of this PR landed elsewhere: the monomoe/rmsnorm_silu copy2 staging
fix is already on main via #5063, which is why GitHub reported conflicts.

What remained unique was the XQA test string cleanup
(`SM120/121` -> `SM120/SM121`) in the two trace tests. Rebased onto main,
dropped the copy2 hunks, and left only that string fix (`be633eaf`).

PR #4879, opened 2026-09-01
feat(prims-ts): support QK-BF16/PV-FP8 in FMHA context

harrisonzhy:
## 📌 Description

Adds support for a differing PV dtype (e.g. QK-BF16/PV-FP8) in primsTS context/prefill attention. Splits `qk_dtype`/`pv_dtype` through the geometry, compile-spec, and kernel config, fixes TMA copy granularity/SMEM descriptor/swizzle assumptions that previously derived V's layout from Q/K's dtype, and gives K and V independent SMEM pipeline configs since a shared TMA pipeline can't satisfy both dtypes' byte-count wait conditions.

## 🔍 Related Issues

N/A

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## Reviewer Notes

Scoped to `fmha_context` kernels only. Decode has not been modified in this PR (but coming soon).

Tested on B300: 

**No performance regression for QKV-BF16 context case.**
[qkvbf16-perf-b300.md](https://github.com/user-attachments/files/32069254/qkvbf16-perf-b300.md)

**Average of 10-26% performance gain with QK-BF16/PV-FP8 quantization over QKV-BF16 baseline.**
Certain configs lead to smaller gains: head_dim_256 and page_size_128/page_size_64 (causal=False) at +3-8%
Largest gains cluster around chunked-prefill and large-batch configs at +25-31%. Moreover, max abs error against FA2 is what we expect for PV-FP8 quantization.
[qkbf16-pvfp8-perf-b300.md](https://github.com/user-attachments/files/32069275/qkbf16-pvfp8-perf-b300.md)

## New Features
- Context attention now supports separate data types for query/key and value inputs.
- Added support for mixed BF16 query/key with FP8 value configurations.
- Added independent key and value processing for mixed-precision attention workloads.
- Benchmarks now support selecting a separate value data type.

## Bug Fixes
- Improved validation for supported query/key and value-type combinations.

## Documentation
- Updated usage guidance for mixed-dtype context attention and value scaling.

## Tests
- Added coverage for packed and paged context attention with mixed BF16/FP8 inputs.


harrisonzhy:
@flashinfer-bot run

3rdparty/cccl
PerkzZheng, line 1: why we would need to update cccl ?

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_kernel.py
```diff
@@ -1685,11 +1725,15 @@ def _infer_single_instance_kv_stages(
...
+    # K and V share one combined SMEM data buffer (disjoint sub-ranges, sized
+    # to the wider of the two dtypes) but keep independent pipelines/barriers. 
+    combined_stage_bytes = (
+        cfg.qk_mma_tiler[1]
+        * cfg.head_dim_per_stage_kv
+        * max(cfg.k_dtype.width, cfg.v_dtype.width)
+        // 8
     )
-    kv_stage_footprint_bytes = kv_stage_bytes + _PIPELINE_BARRIER_BYTES_PER_STAGE
+    kv_stage_footprint_bytes = combined_stage_bytes + _PIPELINE_BARRIER_BYTES_PER_STAGE
```
PerkzZheng: I think we also want to separate smem buffers for K and V to support different dtypes of K and V, this is more structured and can save memory allocation. For other cases, we can still use one SmemKv resource and pipeline.

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_kernel.py
```diff
@@ -789,19 +790,30 @@ def build_context_task_manager(
...
+    kv_stage_k, kv_stage_v = _split_kv_stage(cfg)
+    smem_k_pipeline_cfg = PipelineConfig.create_tma_umma_pipeline_cfg(
+        num_stages=kv_stage_k,
         num_bytes=cfg.tma_copy_kv_bytes,
         producer_group=tma_producer_group,
         consumer_group=pipeline.CooperativeGroup(Agent.Thread),
         cta_layout_vmnk=cluster_shape_vmnk,
         advance_on_wait=True,
     )
+    smem_v_pipeline_cfg = PipelineConfig.create_tma_umma_pipeline_cfg(
```
PerkzZheng: Do you see perf changes after using two pipelines ? I assume not as we just move from stages to pipelines.

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_kernel.py
```diff
@@ -2555,6 +2618,20 @@ def __init__(
         )
         cfg.tma_copy_kv_bytes = cfg.tma_copy_kv_elements * k_dtype.width // 8
 
+        # TMA copy granularity for V
```
PerkzZheng, line 2682: I think those won't be needed if we use standalone buffer for V.

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_resources.py
```diff
@@ -1797,6 +1809,207 @@ def k_load_stage(
...
+            sK_curr,
+            leading_byte_offset=leading_byte_offset,
+            stride_byte_offset=stride_byte_offset,
+            layout=_qk_smem_layout(self.cfg),
+        )
+        return desc_k_base
+
+
+@dataclass(kw_only=True)
+class SmemVResource(MemoryResource):
```
PerkzZheng: we can still use one SmemKv buffer, and handle differently based on K or V inside the class.

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_kernel.py
harrisonzhy: i didn't profile yet but I assume not. for now, this is just draft PR to check my approach first

3rdparty/cccl
harrisonzhy: I think this is just the updates from cccl repo https://github.com/nvidia/cccl
im not sure why it shows here

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_kernel.py
harrisonzhy: so you are saying if K and V are the same dtype, then we can use one instance of SmemKVResource? that makes sense to me

3rdparty/cccl
farazkh80: This rolls cccl back from the Aug-13 commit on main (`16bd510c`) to Mar-18 (`87686768`), looks like a stale local submodule from the rebase rather than an upstream update. Could you reset it with `git checkout origin/main -- 3rdparty/cccl && git submodule update --init 3rdparty/cccl` and drop it from the PR?

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_kernel.py
farazkh80: I would expect a perf change. For the paired D128 topology `_split_kv_stage` gives `kv_stage_k = kv_stage_v = 3*1//2 = 1`, so we go from 3 shared in-flight K/V slots to 2 and `V_{i+1}` cannot be issued until PV releases `V_i`, plus one budgeted SMEM stage goes unused.

Should we try sizing the K/V rings independently from the SMEM budget using each dtype width, and measure bf16-d128 / fp8-d128 dense+causal before/after numbers?

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_kernel.py
```diff
@@ -987,15 +999,21 @@ def make_page_offsets_pipeline_cfg(num_stages: int) -> PipelineConfig:
...
         cfg=cfg,
         page_offsets_kv=smem_page_offsets_kv,
+        page_idx_kv=g_page_idx_kv,
+        name="smem_k",
+    )
+    smem_v = SmemVResource(
+        tma_v_desc=tma_v_desc,
+        pipeline_config=smem_v_pipeline_cfg,
+        cfg=cfg,
         page_offsets_v=smem_page_offsets_v,
```
farazkh80, line 1007: The old SmemKVResource fell back to `page_offsets_kv` when `page_offsets_v` was None. That fallback is gone, so causal paged D256 now takes the direct-gmem page-ID branch without the `_runtime_last_valid_page_idx` clamp while spo still cycles around the V loads.


Is that intentional ?

tests/attention/test_attention_ts_context.py
```diff
@@ -2200,6 +2205,83 @@ def test_attention_ts_context_uniform_packed_window_offsets_accuracy():
...
+    Q/K stays BF16 and V is quantized to FP8 before the PV GEMM. 
+    This tests the k_dtype != v_dtype SMEM staging path across 
+    SmemKResource and SmemVResource.
+    """
+    case = _make_context_case(
+        q_lengths=(64, 96),
+        k_lengths=(192, 160),
+        num_qo_heads=4,
+        num_kv_heads=4,
+        head_dim=256,
```
farazkh80: This d128 test passes `head_dim=256` and is identical to the d256 test below, so the paired-D128 mixed path (FP8 P-cast under the bf16 early-tile-sum cadence) is untested. Also both are named `dense` but use `mask_type="causal"`.

tests/attention/test_attention_ts_context.py
```diff
@@ -2200,6 +2205,83 @@ def test_attention_ts_context_uniform_packed_window_offsets_accuracy():
     _assert_context_correct(wrapper.run(case.q, case.k, case.v), case)
 
 
+
+@pytest.mark.arch_blackwell
+@_REQUIRES_CONTEXT_GPU
+def test_attention_ts_context_d128_qkbf16_pvfp8_dense_accuracy():
```
farazkh80: Could we add a paged-KV mixed case (`v_cache` FP8) and a non-causal one? Extending `_assert_context_correct` to select the FP8 tolerance on `case.v.dtype` would let both reuse it instead of the hand-rolled assert.

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_kernel.py
harrisonzhy: Regarding sizing K/V rings independently: V_i can only be usefully prefetched after K_i QK MMA completes because you need attention score before doing the PV computation. so I don't think asymmetric depth (e.g. where V would get more slots since its FP8) would help here in terms of performance. But let me know if I am wrong.

I will work on benchmarking for context phase.

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_kernel.py
harrisonzhy: Good point, in the updated diff addressing comments and consolidating `Smem{K,V}Resource` back into one class `SmemKVResource`, I put it back. Should be in `SmemKVResource.__init__` now.

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_kernel.py
harrisonzhy: I think it is still needed. The K and V resources share a FmhaConfig, which is also what the original code does. Thus we need these params to differentiate V TMA box dimensions and swizzle from K when their dtypes are different

flashinfer/attention/prims_ts/kernels/fmha_context/fmha_kernel.py
PerkzZheng: > For the paired D128 topology _split_kv_stage gives kv_stage_k = kv_stage_v = 3*1//2 = 1, so we go from 3 shared in-flight K/V slots to 2 and V_{i+1} cannot be issued until PV releases V_i, plus one budgeted SMEM stage goes unused.

right, that would make a difference with separate smem buffers.

PerkzZheng (approved):

saltyminty:
prims-ts is experimental so API breakages warnings can be ignored

benchmarks/routines/attention.py
```diff
@@ -1440,6 +1450,11 @@ def testBatchPrefillWithPagedKVCacheWrapper(args):
         return res
     kv_init_dtype = torch.float16 if kv_dtype == torch.float16 else torch.bfloat16
 
+    v_dtype = dtype_str_to_torch_dtype(args.v_dtype) if args.v_dtype else kv_dtype
+    if v_dtype not in [torch.float16, torch.bfloat16, torch.float8_e4m3fn]:
```
saltyminty: This `Unsupported v_dtype` guard should only be executed for prims-ts – it could be supported by other backends, i.e. nvfp4 trtllm-gen

saltyminty (approved):
One minor benchmark comment but otherwise LGTM

benchmarks/routines/attention.py
harrisonzhy: I will fix, thanks!

PR #4906, opened 2026-09-02
perf(mla): swap-AB Q tiles for small-M decode and a parallel split-KV merge on SM90

### LINKED ISSUE #2995: [FA3-MLA]Reducing WGMMA padding overhead in MLA decode for small M
Hi everyone,

I've been analyzing the **FA3 BF16 MLA Attention** on H200 and noticed significant compute underutilization during the decode phase.

When the number of active query rows ($M$) is small (e.g., $BS=1, num\\_heads=8, M < 16$), the current QK, QV, and PV kernels pad $M$ up to the fixed WGMMA-M dimension of 64. This wastes a large portion of compute on zero-padded rows.

By swapping the A and B operands, we can move the small query dimension from **WGMMA-M (fixed at 64)** to **WGMMA-N (granularity of 8 for BF16)**:

```text
                  queries(M=64)  seq_block(64, from KV cache block size)
                               ↓  ↓
Original PV:  O[64, 512]  = P[64, 64]  × kv_c[64, 512]    → mostly padding in M
Swapped PV:   O^T[512, 8] = kv_c^T[512,64] × P^T[64, 8]   → queries in N=8, ~8x less compute
                                                     ↑
                                              queries(N=8)
```

The same logic applies to the QK and QV GEMMs. 
Would it be possible to implement a specialized path for small $M$ using this layout to mitigate the padding overhead?

SamMausberg:
## Description

MLA decode on Hopper (`BatchMLAPagedAttentionWrapper`, `fa3` backend) runs `qo_len * num_heads` packed query rows per work item. Under tensor parallelism a decode step has 8, 16 or 32 rows, which the kernel padded to its fixed 64-row WGMMA-M tile (issue #2995). This PR fixes that and the merge bottleneck it exposed. Two commits.

### Swap-AB Q tiles

`hopper::HopperKernelTraits` takes the Q tile as a parameter. The 16- and 32-row tiles swap the GEMM operands so the query rows land on WGMMA-N (granularity 8):

```
S^T[kv, q]  = K[kv, d]   * Q^T[d, q]     A = K   (K-major),  B = Q   (K-major)
O^T[d, q]  += V^T[d, kv] * P^T[kv, q]    A = V^T (MN-major), B = P^T (K-major)
```

The products read the same shared-memory tiles and descriptors as before; only the operand roles and the accumulator layout change, with each warpgroup's half of `HEAD_DIM_CKV` split into 64-row WGMMA-M chunks. Softmax statistics are per column, reduced over the eight lanes of a column and once across the four warps through shared memory. P and O are staged with `stmatrix.trans`. The 64-row tile keeps the original layout, and both layouts share the same producer and consumer bodies.

`MLAPlan` picks the tile from the longest packed query in the batch (`MLAPlanInfo::cta_tile_q`), one kernel instantiation per tile, only when a work item runs on a single CTA. FA2 keeps the 64-row tile. A CUDA graph replays the kernel and grid it was captured with, so a replan under a graph now keeps the captured cluster size and tile; any tile is correct (a narrower one only means more Q tiles per request), and this also covers the pre-existing cluster-size hazard.

### Parallel split-KV merge

With the tiles in place the tensor pipe went from 29% to 9% busy but decode time barely moved: long-context decode was bound by the split-KV merge, which walked up to 128 partial states per row one at a time at L2 latency with 64 of 256 threads active (29 us of a 77 us kernel at 128K tokens). `DevicePersistentMergeStates` (shared with FA2) now splits a row's partials over the four 64-thread groups of a CTA, keeps eight loads in flight per group, and combines the groups through shared memory. A CTA that owns several rows still gives each group its own row.

### Results

Kernel time (CUPTI median) on a GH200, BF16, `head_dim_ckv=512`, `head_dim_kpe=64`; `qo_len=2` is causal. Before is `main` at 61f9b29d.

| batch | qo_len | heads | kv_len | page_size | before (us) | after (us) | speedup |
|---:|---:|---:|---:|---:|---:|---:|---:|
| 1 | 1 | 8 | 4096 | 64 | 16.2 | 9.4 | 1.72x |
| 1 | 1 | 8 | 16384 | 64 | 27.2 | 13.6 | 2.00x |
| 1 | 1 | 8 | 65536 | 64 | 57.0 | 32.7 | 1.74x |
| 1 | 1 | 8 | 131072 | 64 | 77.5 | 52.7 | 1.47x |
| 1 | 1 | 16 | 16384 | 64 | 27.7 | 13.7 | 2.02x |
| 1 | 1 | 16 | 131072 | 64 | 79.2 | 53.6 | 1.48x |
| 1 | 1 | 32 | 16384 | 64 | 27.8 | 15.7 | 1.77x |
| 1 | 1 | 32 | 131072 | 64 | 80.2 | 55.7 | 1.44x |
| 1 | 1 | 64 | 16384 | 64 | 29.5 | 18.2 | 1.62x |
| 1 | 1 | 64 | 131072 | 64 | 83.7 | 58.6 | 1.43x |
| 1 | 1 | 128 | 131072 | 64 | 85.7 | 72.5 | 1.18x |
| 1 | 2 | 16 | 65536 | 64 | 59.9 | 34.8 | 1.72x |
| 1 | 1 | 16 | 65536 | 1 | 60.0 | 33.7 | 1.78x |
| 8 | 1 | 16 | 8192 | 64 | 33.8 | 29.8 | 1.13x |
| 32 | 1 | 16 | 32768 | 64 | 332.9 | 329.5 | 1.01x |

At batch 1 with 8 or 16 heads the main loop now runs at the HBM streaming rate of this device (3.56 TB/s in a `cp.async` microbenchmark); at 128K tokens the kernel spends 42 us streaming 151 MB and about 10 us on launch, grid sync and merge. Larger batches were already bandwidth-bound. The 64- and 128-head rows keep the original tile and gain from the merge only.

## Related Issues

Closes #2995.

### Pre-commit Checks

## Tests

New tests in `tests/attention/test_deepseek_mla.py`: narrow tiles (8, 16, 32 heads; `qo_len` 1, 2, 4; FP16 and BF16; page sizes 1 and 64; causal), long-context split-KV decode, mixed-`qo_len` batches, graph replans that keep the captured configuration, and FP8 KV with 8 and 32 heads. Existing FA2 and FA3 MLA tests pass, including out-of-bounds KV, no-KPE and FP8. Outputs were also checked against an fp32 PyTorch reference over 416 configurations with FA3 errors at or below FA2's.

## Reviewer Notes

- The 64-row kernel keeps its register budget (234 registers, no spills); the 16- and 32-row kernels use 102 and 174.
- `MLAPlan` gained `min_cta_tile_q` (FA2 passes 64) and the captured `graph_num_blks_x` / `graph_cta_tile_q` of a graph replan; `MLAPlanInfo` gained `cta_tile_q` (last entry), so the JIT URI tag moves to `planabi3`.
- Named barriers 6 and 7 are the per-warpgroup barriers of the swap-AB tiles; CUTLASS reserves hardware barriers 8 to 15.
- The merge commit stands on its own and also speeds up FA2 split-KV decode.

Designed in combination with Claude Fable 5.1.

## Summary by CodeRabbit

* **New Features**
  * Added adaptive query-tile selection for Hopper-based MLA attention, supporting tile sizes of 16, 32, and 64.
  * Improved CUDA Graph planning and replanning by preserving captured execution settings.
  * Enhanced long-context and split-KV attention merging for more flexible workloads.
* **Performance**
  * Optimized SM90 MLA execution for narrow query batches and mixed query lengths.
* **Bug Fixes**
  * Improved numerical handling and output validation across supported data types.
  * Corrected transposed matrix shared-memory operations on supported CUDA architectures.

### DIFF
--- a/csrc/batch_mla_binding.cu
+++ b/csrc/batch_mla_binding.cu
@@ -25,7 +25,8 @@ using tvm::ffi::Tuple;
 Tuple<Array<int64_t>, int64_t> BatchMLAPagedAttentionPlan(
     TensorView float_workspace_buffer, TensorView int_workspace_buffer,
     TensorView page_locked_int_workspace_buffer, TensorView qo_indptr, TensorView kv_indptr,
-    TensorView kv_len, int64_t num_heads, int64_t head_dim_o, bool causal);
+    TensorView kv_len, int64_t num_heads, int64_t head_dim_o, bool causal, int64_t graph_num_blks_x,
+    int64_t graph_cta_tile_q);
 
 void BatchMLAPagedAttentionRun(TensorView float_workspace_buffer, TensorView int_workspace_buffer,
                                Array<int64_t> plan_info_vec, TensorView q_nope, TensorView q_pe,
--- a/csrc/batch_mla_plan.cu
+++ b/csrc/batch_mla_plan.cu
@@ -28,7 +28,8 @@ using tvm::ffi::Tuple;
 Tuple<Array<int64_t>, int64_t> BatchMLAPagedAttentionPlan(
     TensorView float_workspace_buffer, TensorView int_workspace_buffer,
     TensorView page_locked_int_workspace_buffer, TensorView qo_indptr, TensorView kv_indptr,
-    TensorView kv_len, int64_t num_heads, int64_t head_dim_o, bool causal) {
+    TensorView kv_len, int64_t num_heads, int64_t head_dim_o, bool causal, int64_t graph_num_blks_x,
+    int64_t graph_cta_tile_q) {
   CHECK_INPUT_TYPE(qo_indptr, dl_int32);
   CHECK_INPUT_TYPE(kv_indptr, dl_int32);
   CHECK_INPUT_TYPE(kv_len, dl_int32);
@@ -51,7 +52,8 @@ Tuple<Array<int64_t>, int64_t> BatchMLAPagedAttentionPlan(
       int_workspace_buffer.data_ptr(), page_locked_int_workspace_buffer.data_ptr(),
       int_workspace_size_in_bytes, plan_info, staged_int_workspace_bytes,
       static_cast<IdType*>(qo_indptr.data_ptr()), static_cast<IdType*>(kv_indptr.data_ptr()),
-      static_cast<IdType*>(kv_len.data_ptr()), batch_size, num_heads, head_dim_o, causal, stream);
+      static_cast<IdType*>(kv_len.data_ptr()), batch_size, num_heads, head_dim_o, causal,
+      /*min_cta_tile_q=*/64, graph_num_blks_x, graph_cta_tile_q, stream);
   TVM_FFI_ICHECK(status == cudaSuccess)
       << "Failed to plan MLA, error: " << cudaGetErrorString(status);
 
--- a/csrc/batch_mla_sm90_binding.cu
+++ b/csrc/batch_mla_sm90_binding.cu
@@ -25,7 +25,8 @@ using tvm::ffi::Tuple;
 Tuple<Array<int64_t>, int64_t> BatchMLAPagedAttentionSM90Plan(
     TensorView float_workspace_buffer, TensorView int_workspace_buffer,
     TensorView page_locked_int_workspace_buffer, TensorView qo_indptr, TensorView kv_indptr,
-    TensorView kv_len, int64_t num_heads, int64_t head_dim_o, bool causal);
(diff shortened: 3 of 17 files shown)


SamMausberg:
Addressed the review in 43c13437: stmatrix.trans now uses the documented mnemonic order and is gated on PTX ISA 7.8 (CUDA 11.8); the merge scratch has a compile-time size check against the kernel's shared memory; the graph replan test captures a CUDA graph with static buffers, replans for batches that would pick a 64-row tile or two-CTA clusters, and replays it (passes on FA2 and FA3).

PR #4910, opened 2026-09-03
feat(moe): persistent disk cache for the SM12x W4A16 CuTe-DSL MoE kernels

### LINKED ISSUE #4892: [Feature] Persistent disk cache for the SM12x W4A16 MoE kernels (moe_w4a16_compiler._COMPILE_CACHE; extend the #4331/#4701 CuTe-DSL disk cache)
## Summary

The SM12x W4A16 CuTe-DSL MoE kernels (`moe_w4a16_kernel.py`, memoized through `moe_w4a16_compiler._COMPILE_CACHE`) are the remaining member of the b12x kernel family compiled in-process only: the b12x MMA kernels already go through the shared on-disk CuTe-DSL cache (#4331, `build_and_load_cute_dsl_kernel`), and #4701 is extending it to the direct-micro kernel. Every fresh process — each engine start, each TP/EP rank, each test process — re-pays the full JIT compile for every W4A16 specialization it touches. This issue tracks rolling the same persistence mechanism out to the W4A16 path.

## Measured cost (RTX 5090, sm_120)

From a serving-like sweep (CUDA-graph-style m buckets 1→512, two MoE configs), timed after process-wide DSL init so the numbers are the kernels' own compile cost:

- one fused-MoE specialization: **0.6–2.0 s** cold compile (grows with `block_size_m`; 2.0 s at block 64 / prefill scale)
- one model config across the full m range: 5 fused specializations (m buckets collapse into `block_size_m` ∈ {8,16,32,48}), **≈ 4.7 s**
- two model configs: 8 specializations, **7.3 s** total
- cross-process: fully re-paid every time — verified with back-to-back fresh processes (identical timings) and no W4A16 artifacts ever appearing under `FLASHINFER_JIT_DIR`, while the sibling adopters' module directories persist alongside.

To set expectations honestly: this is a seconds-scale per-process tax, not the ~minute-scale compiles of the direct-micro megakernel. The tax scales linearly with model configs × batch buckets × ranks × restarts (and is re-paid by every test process during development), and fixing it closes the one remaining gap in the b12x family's disk-cache coverage.

## Scope

- Infra-only, following the adopter pattern established in #4701: `--enable-tvm-ffi` compiles against a fake TVM-FFI env stream, per-entry launch-ABI conversion for the four compile entries in `moe_w4a16_kernel.py` (gemm, fused_moe, activation, topk_sum), and a dedicated on-disk module directory (not shared with the MMA or direct-micro modules, whose per-module source hash would otherwise cause mutual invalidation).
- Cache keys derive from the existing `KernelCompileSpec.json_key` (which already covers the specialization axes), plus source-file hashing over the W4A16 device-code modules so stale-artifact reuse is impossible. The in-process caches stay as the first lookup layer.
- No kernel numerics changes. Tests mirroring the direct-micro battery: cold/warm round-trip, cache-key invalidation, and CUDA-graph capture on warm-loaded artifacts.

## Prior context

- I ran the SM120 verification for the direct-micro cache work on #4317 (81 naming-contract tests, w4a16 route-packing incl. CUDA graph, disk-cache cold/warm timings) and contributed the direct-micro barrier launch-ABI fix now carried in #4701 (commit c61816c).
- The disk-cache primitive is already on main (#4331), so this work does not depend on #4701 merging; it follows #4701's adopter pattern and rebases trivially over it.
- Distinct from #4561: that issue tracks the SM100 CuTe DSL W4A16 MoE implementation and its follow-ups. This one is compiler-cache infrastructure on the existing SM12x W4A16 path.

## Plan

I'd like to author this PR (claiming below). Verification hardware on hand: RTX 5090 (sm_120), with same-day full-battery re-runs on revisions, as done for #4317.

gavinkvx:
Closes #4892

## Summary

Routes the SM12x W4A16 CuTe-DSL MoE kernels -- the last member of the b12x kernel family compiled in-process only -- through the shared on-disk kernel cache (#4331), following the adopter pattern of the direct-micro conversion (#4701). All four compile entries in `moe_w4a16_kernel.py` are converted (topk_sum, activation, fused_moe, gemm); the production-launched fused entry additionally converts its launch site to the TVM-FFI ABI.

Measured on RTX 5090 (sm_120): the W4A16 test batteries drop from **4.73 s → 0.61 s** (route-pack) and **8.27 s → 3.71 s** (unified, w4a16 selection) on warm processes; per-specialization cold compiles of 0.6–2.0 s (≈7.3 s across a serving-like two-config sweep) are paid once per machine instead of once per process. This is a seconds-scale per-process tax, not the minute-scale direct-micro compiles; it scales with model configs × batch buckets × ranks × restarts and is re-paid by every test process during development.

## Extent policy (the substantive design point)

TVM-FFI exact-checks every DLTensor's baked extent at call time, yet extents never bound a device-side access: the kernels index through stride math bounded by their scalar arguments, and the pre-FFI launch passed buffers whose sizes bore no relation to the fake shapes. The kernel cache keys also do not track the m bucket, so one artifact serves all m for a given specialization. The policy that follows, applied to all four entries:

- m-independent tensors with real size contracts (weights, scales, global scales, activation_amax) bake true extents and pass full flat views;
- every m-, route- or scratch-capacity tensor bakes `(1,)` barrier-style (#4701) and passes a length-1 view -- an extent that varies with an unkeyed fact would poison cross-m reuse of the cached artifact.

This surfaced three pre-existing fictions in the historical fake shapes, latent only because nothing checked them before TVM-FFI: the FC2 fake ignored TC-decode mode (which writes the m×hidden output buffer, not routed rows), the c_tmp fakes carried a `max(..., 4*256*mbs*256)` floor unrelated to the real `packed_gemm_scratch_elements` allocation, and the locks fake assumed a ≤256-SM part while the real workspace follows `sms*4+2`.

## Other notes for review

- Opt levels are preserved: the fused entry's explicit `OptLevel(3)` becomes `--opt-level 3 --enable-tvm-ffi`; entries that used the default compile at `--opt-level 2 --enable-tvm-ffi`, matching the direct-micro default.
- The fused launch drops its explicit stream argument for the TVM-FFI env stream and now raises on a non-current stream (the env-stream ABI cannot target another stream); `pack_topk_routes_by_expert` already enforces the same contract in this file.
- When amax collection is off, the const-expr-dead `activation_amax` param gets an F32-scratch stand-in cut to its baked 2E extent.
- On-disk names are readable facts plus a sha256 digest of the entry's full `cache_key`; injectivity over the codegen parameters is inherited from the key that already gates the in-process caches. `tests/moe/test_b12x_w4a16_kernel_cache.py` pins that contract (digest sensitivity incl. float-sign/None-vs-0, determinism, symbol safety) plus GPU disk round-trip and CUDA-graph capture on a warm-loaded artifact.
- The standalone gemm and activation entries have no in-repo launchers; their verification is the compile/persist/reload round-trip, with the family's device codegen exercised by the fused battery.
- A dedicated `b12x_moe_w4a16` module directory avoids the mutual-invalidation hazard of sharing a per-module source hash with the MMA or direct-micro adopters. Independent of #4701: the cache primitive is on main, and this rebases trivially over that PR when it lands.

## Verification (RTX 5090, sm_120)

Full W4A16 battery green on cold and warm processes: `tests/moe/test_b12x_w4a16_route_pack.py` (64), `tests/moe/test_unified_moe_b12x.py -k w4a16` (23, incl. CUDA-graph capture/replay and dispatch-accuracy conformance), and the new cache-contract suite (6).

## Summary by CodeRabbit

* **New Features**
  * Added persistent disk caching for W4A16 kernel compilation, enabling reuse across runs.
  * Added source-change tracking to invalidate outdated cached kernels.
  * Added support for CUDA graph-capturable warm-loaded kernels.

* **Bug Fixes**
  * Improved validation of output buffer capacity and CUDA stream usage.
  * Corrected handling of scalar-sized route, activation, output, and scratch buffers.
  * Updated top-k sum output handling for tensor-based launches.
  * Added validation for required activation calibration data.

* **Tests**
  * Added coverage for deterministic, collision-resistant cache keys and cached-kernel numerical behavior.

### DIFF
--- a/flashinfer/fused_moe/cute_dsl/blackwell_sm12x/moe_w4a16_kernel.py
+++ b/flashinfer/fused_moe/cute_dsl/blackwell_sm12x/moe_w4a16_kernel.py
@@ -5939,8 +5939,9 @@ def compile_w4a16_activation(
 
     # The on-disk artifact name must be injective over every compile fact.
     # ``cache_key`` also carries the swiglu scalars, fast-math and gating
-    # flags, and the row-specialization bucket, so the readable facts are
-    # suffixed with a digest of the full key.
+    # flags; the row bucket is deliberately excluded so one artifact can
+    # be reused across rows. The readable facts are suffixed with a digest
+    # of the full key.
     compiled = build_and_load_cute_dsl_kernel(
         _CUTE_DSL_MODULE,
         _w4a16_disk_kernel_name(
--- a/flashinfer/fused_moe/cute_dsl/blackwell_sm12x/moe_w4a16_kernel.py
+++ b/flashinfer/fused_moe/cute_dsl/blackwell_sm12x/moe_w4a16_kernel.py
@@ -10,7 +10,6 @@ import cuda.bindings.driver as cuda
 import cutlass
 import cutlass.cute as cute
 import torch
-from cutlass.base_dsl.compiler import OptLevel
 from cutlass._mlir.dialects import llvm
 from cutlass.cutlass_dsl import Int32, Int64, T, Uint32, dsl_user_op
 
--- a/flashinfer/fused_moe/cute_dsl/blackwell_sm12x/moe_w4a16_kernel.py
+++ b/flashinfer/fused_moe/cute_dsl/blackwell_sm12x/moe_w4a16_kernel.py
@@ -5605,13 +5605,16 @@ def compile_w4a16_fused_moe(
             blocks_per_sm=kernel.blocks_per_sm,
         )
 
-    compile_size_m = _fake_m_for_specialization(size_m)
-    compile_routed_rows = int(compile_size_m) * int(top_k)
-    compile_route_blocks = compile_routed_rows if direct_topk_routes else 1
-    compile_route_slots = compile_route_blocks * int(moe_block_size)
-    packed_route_fake_elements = (
-        compile_routed_rows if direct_topk_routes else compile_route_slots
-    )
+    # Extent policy: TVM-FFI exact-checks every DLTensor's baked extent at
+    # call time, extents never bound a device-side access (the kernels
+    # index through stride math bounded by their scalar arguments), and
+    # the kernel cache key does not track the m bucket -- one artifact is
+    # shared across m for a given specialization. So only m-independent
+    # tensors with real size contracts (weights, scales, global scales,
+    # activation_amax) bake true extents; every m-, route- or
+    # scratch-capacity tensor bakes (1,) barrier-style (#4701), since any
+    # extent that varies with an unkeyed fact would poison cross-m reuse
+    # of the cached artifact.
     a_fake = make_ptr(cutlass_dtype, 16, cute.AddressSpace.gmem, assumed_align=16)
     if weight_layout == "modelopt":
         w13_fake = cute.runtime.make_fake_compact_tensor(
@@ -5649,17 +5652,17 @@ def compile_w4a16_fused_moe(
         )
     fc1_fake = cute.runtime.make_fake_compact_tensor(
         cutlass_dtype,
-        (compile_routed_rows * fc1_cols,),
+        (1,),
         assumed_align=16,
     )
     activated_fake = cute.runtime.make_fake_compact_tensor(
         cutlass_dtype,
-        (compile_routed_rows * intermediate_size,),
+        (1,),
         assumed_align=16,
     )
     fc2_fake = cute.runtime.make_fake_compact_tensor(
         cutlass_dtype,
-        (compile_routed_rows * hidden_size,),
+        (1,),
         assumed_align=16,
     )
     w13_scales_fake = cute.runtime.make_fake_compact_tensor(
@@ -5700,12 +5703,12 @@ def compile_w4a16_fused_moe(
     )
     packed_routes_fake = cute.runtime.make_fake_compact_tensor(
         cutlass.Int32,
-        (packed_route_fake_elements,),
+        (1,),
         assumed_align=16,
     )
     block_experts_fake = cute.runtime.make_fake_compact_tensor(
         cutlass.Int32,
-        (compile_route_blocks,),
+        (1,),
         assumed_align=16,
     )
     route_count_fake = cute.runtime.make_fake_compact_tensor(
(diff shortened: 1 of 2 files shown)


gavinkvx:
@coderabbitai The activation finding is correct -- that entry was converted before the extent policy the later fused/gemm commits established, and was never brought under it. Fixed in 878c8d0a, though by the policy-consistent resolution rather than keying the bucket: the row bucket is deliberately absent from the cache key (the in-process cache reuses one compiled kernel across rows via `replace(cached, rows=rows)`), so the m-varying fake extents now bake `(1,)` like every other unkeyed-capacity tensor and a single artifact genuinely serves all row counts. Keying the bucket would instead split the artifacts and contradict that existing cross-row reuse. Added a regression test pinning the sharing semantics: both buckets resolve to one artifact, warm-launchable at either row count.

gavinkvx:
Hi team! Hope this helps! 

@Aneureka @bkryu Since the PR CI matrix has no SM120 runner, here is a quick summary of how this was verified locally (RTX 5090, sm_120), to make authorization/review easier:

- **Correctness**: the full existing W4A16 GPU batteries pass on both cold and warm processes -- `tests/moe/test_b12x_w4a16_route_pack.py` (64) and `tests/moe/test_unified_moe_b12x.py -k w4a16` (23, incl. dispatch-accuracy conformance and CUDA-graph capture/replay). The dispatch-accuracy suite is also what established the extent policy empirically: each of the three historical fake-shape fictions in the description was caught by it during development, fixed, and re-run to green.
- **Cache behavior**: the new cache-contract suite (7 tests) pins artifact-name injectivity (digest sensitivity incl. float-sign / None-vs-0 collisions), the disk round-trip surviving an in-process cache clear, single-artifact sharing across activation row buckets (in an isolated test-local JIT directory), and CUDA-graph capture on a warm-loaded artifact.
- **Effect**: warm-process batteries drop 4.73 s → 0.61 s (route-pack) and 8.27 s → 3.71 s (unified); artifacts persist under a dedicated `b12x_moe_w4a16` module directory, invalidated on source change.

Happy to run any additional command or config locally and paste full output same-day. When convenient, could one of you authorize CI (`@flashinfer-bot run` -- it needs a member; the bot politely declined mine)?

@bkryu tagging you for continuity from #4317/#4701 -- this is the W4A16 rollout of that same mechanism.

Please let me know, thank you!

PR #4914, opened 2026-09-03
fix(moe): address review findings for PR #4610

cindyzxq:
## 📌 Description

Follow-up hardening for PR #4610.

  This PR:

  - rejects combining `TrtllmFp4Config` and `CutlassNvfp4Config` (and `CuteDslConfig` + `CutlassNvfp4Config`) as NVFP4 candidates because they require incompatible `MoEActivationPack` layouts;
  - updates the unified MoE documentation to use backend-native NVFP4 examples;
  - documents that TRT-LLM and CUTLASS FP8 per-tensor packs cannot be shared;
  - strengthens CUTLASS FP8 per-tensor coverage at the runner input-packing boundary;
  - updates the MXFP8 numerical test to build its reference from the prepared quantized weights and scales instead of the original BF16 weights.

### ✅ Pre-commit Checks

## 🧪 Tests

## Reviewer Notes

Please focus on:

  1. Whether rejecting mixed TRT-LLM/CUTLASS NVFP4 candidates is the appropriate fail-fast behavior.
  2. Whether the documented activation contracts match the backend implementations.
  3. Whether the MXFP8 prepared-weight reference is sufficiently independent of the execution path.

## Summary by CodeRabbit

- **New Features**
  - Added an opt-in NVFP4 inference backend for supported SM103 hardware.
  - Added backend discovery and configuration support for the new option.

- **Documentation**
  - Clarified NVFP4 backend selection and activation-pack compatibility requirements.
  - Updated examples to use valid single-backend configurations and current activation-pack fields.
  - Added the new backend to the supported MoE configuration matrix.

- **Tests**
  - Strengthened FP8 and MXFP8 validation against quantized reference results.


flashinfer/fused_moe/api.py
```diff
@@ -1475,6 +1475,27 @@ def __post_init__(self) -> None:
...
+        while autotuning.  TRT-LLM NVFP4 consumes packed FP4 values and block
+        scales, whereas CUTLASS NVFP4 consumes raw BF16 values without an
+        activation scale.  Letting both reach dispatch therefore fails only
+        after the layer has begun constructing and profiling runners.
+        """
+        if self.quant.variant is not QuantVariant.NVFP4:
+            return
+
+        candidate_types = {type(candidate) for candidate in self.backend}
+        if {TrtllmFp4Config, CutlassNvfp4Config} <= candidate_types:
```
feih-nv: This only rejects `{TrtllmFp4Config, CutlassNvfp4Config}`. CuteDSL NVFP4 uses the same packed-FP4 + scale layout as TRT-LLM, so `CuteDslConfig` + `CutlassNvfp4Config` is the same incompatible pair.

`tests/moe/test_unified_moe.py` (`test_cutedsl_nvfp4`) still constructs:

```python
BackendOptions(candidates=(CuteDslConfig(), CutlassNvfp4Config()))
```

That CPU expressiveness test still passes because this guard does not fire, but a real `MoELayer` autotune cannot share one `MoEActivationPack` between them.

Please extend this check to cover CuteDSL + CUTLASS NVFP4, and switch that test to `TrtllmFp4Config()` (compatible packed-FP4 pair) or a `CutlassNvfp4Config()` singleton.

docs/design_docs/flashinfer_moe_api.md
```diff
@@ -42,24 +42,24 @@ config = MoEConfig(
     ),
     quant=QuantConfig(QuantDtype.FP4, QuantGranularity.BlockScale),
     experts=ExpertConfig(intermediate_size=2048, local_num_experts=32),
-    backends=[TrtllmFp4Config(extra_backend_params...), CutlassNvfp4Config()],
+    # NVFP4 activation packs are backend-native; choose one candidate set.
+    backends=[TrtllmFp4Config(extra_backend_params...)],
```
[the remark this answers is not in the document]
feih-nv: @coderabbitai These snippets are in the aspirational §2–§4 design, not the shipped API. The doc already says that at "MVP As-Built Reference": `moe_layer(...)`, `find_backends`, `backends=`, and `selected_experts`/`final_scales` are the long-range surface.

The current public API (`backend=BackendOptions`, `topk_ids`/`topk_weights`) is already shown in that as-built section. This PR only removes the incompatible `TrtllmFp4Config` + `CutlassNvfp4Config` mix from the aspirational examples; rewriting §2–§4 to the MVP API is out of scope.

cindyzxq:
@flashinfer-bot run

flashinfer/fused_moe/api.py
feih-nv: this has been addressed.

feih-nv:
@flashinfer-bot run

flashinfer/fused_moe/api.py
```diff
@@ -1572,6 +1572,32 @@ def __post_init__(self) -> None:
         # Not in check_support(): MoELayer swallows its exceptions to filter
         # backends, so errors raised there surface as "no backend available".
         self._validate_fused_shared_experts()
+        self._validate_activation_pack_contract()
+
+    def _validate_activation_pack_contract(self) -> None:
```
feih-nv: I don't think this fail-fast belongs to `MoEConfig`. The mismatch is `(pack, runner)`, not a pair of config classes, and `MoEConfig` never sees the pack.

`pack_inputs` already rejects the wrong layout. `check_support` cannot hold this rule: there is no pack yet, and `MoELayer` would swallow it to skip backends. A hardcoded class matrix also misses B12x / cuTile.

Let's drop the constructor check and the mixed-reject constructor tests. Keep the doc fix. Leave the error to `pack_inputs`.

Filtering mixed candidates from a runner-owned layout tag can be a follow-up.

cindyzxq: Thanks,Fei. Have updated  codes.

feih-nv:
@flashinfer-bot run

feih-nv (approved):
Approving as this is purely tests and doc side change.

PR #4924, opened 2026-09-03
feat(comm): add SageAttention2 low-precision QKV exchange to UlyssesCommunicator

DwenGu:
## 📌 Description

Extend `flashinfer.comm.UlyssesCommunicator` with `prepare_qkv()` and `scatter_qkv()` for low-precision input communication. They accept FP16/BF16 Q/K/V sequence shards, quantize and pack them before all-to-all, and return the received SageAttention2 operands plus scales and metadata. Internally, Q/K are encoded as INT8 and V as FP8 E4M3. Quantization groups remain on the global sequence grid across rank boundaries.

The caller supplies the process group and attention invocation. The communicator prepares reusable storage and owns one statistics AllGather and one combined payload AllToAll on that group's NCCL backend. Existing `scatter_heads()` / `gather_heads()` behavior and backend selection are preserved. Lowp numeric helpers live in the private `_ulysses_lowp.py` module; the former public module and duplicate `flashinfer.comm` exports are removed.

[Design and numerical contracts](https://github.com/DwenGu/flashinfer/blob/59e4fbe8aa2e6b524a4e03ae6d037427a0fa0378/docs/design_docs/ulysses_lowp.md) · [Distributed acceptance example](https://github.com/DwenGu/flashinfer/blob/59e4fbe8aa2e6b524a4e03ae6d037427a0fa0378/benchmarks/comm/validate_ulysses_lowp.py) · [Communicator benchmark](https://github.com/DwenGu/flashinfer/blob/59e4fbe8aa2e6b524a4e03ae6d037427a0fa0378/benchmarks/comm/bench_ulysses_qkv.py).

### Execution and API

```python
from flashinfer.comm import UlyssesCommunicator

with UlyssesCommunicator(group, max_elems=q.numel(), dtype=q.dtype,
                         device=q.device, backend="nccl") as comm:
    workspace = comm.prepare_qkv(q.shape, quantization="sage2", used_sequence=used)
    encoded = comm.scatter_qkv(q, k, v, workspace=workspace)
    # Call the framework's Sage2 adapter with encoded operands and scales.
    # For subsequent serialized calls, reuse output storage with out=encoded.
```

Preparation is collective and validates configuration, layout/JIT availability and allocation outcomes across ranks. Each execution checks actual input/output metadata and storage overlap, then invokes shared private stage implementations without repeating the standalone wrappers' checks. QKV projection views are accepted without three full contiguous copies. `UlyssesQKV` contains six tensors plus layout, logical/live sequence lengths and input dtype; `out=None` owns independent storage and `out=encoded` reuses it. Each Q/K/V must fit `max_elems`; combined wire and statistics capacity are derived internally.

One `BOUNDARY_MERGE` computation path handles aligned and nonaligned shards. Boundary statistics are merged for every rank contributing to a global quantization group, including groups spanning more than two ranks. The single FP32 statistics AllGather carries K sums, V channel maxima, Q boundary maxima and K boundary extrema. Quantize-and-pack consumes the finalized scales; the receiver applies Sage's V permutation and reconstructs global Q/K scale arrays.

Both architecture modules compile head dimensions 64 and 128. `prepare_qkv()` selects the device layout and dimension from the supplied shape; callers do not construct layout objects. Host bindings and kernel templates are shared across architecture specializations.

### Supported configurations and contracts

| Property | Contract |
|---|---|
| Input | Q/K/V with identical `[B, L, H, D]` shape, CUDA device and FP16/BF16 dtype; positive B/L/H; D=64 or 128 |
| Distribution | Equal L on all ranks; H divisible by P; the prepared Lowp API accepts P=2/4/8 and rejects P=1/6. Ordinary communicator behavior is unchanged |
| Sequence | Arbitrary positive local L; global live prefix `0 < U <= P*L`; caller zero-fills the global tail before sharding |
| SM90 | Q groups of 16 tokens; K groups of 128; V storage padded to 128 rows |
| SM120 | Q groups of 32 tokens; K groups of 64; V storage padded to 64 rows |
| Memory layout | Dense D axis with 16-byte-aligned row addresses, including interleaved projection views; exact contiguous output buffers |
| Numeric behavior | FP32 K sums, mean divided by U and rounded to input dtype before centering; Q/K grouped-amax floor `1e-7`; V channel scale `amax / 2.25`, with zero-amax channels encoded as positive zero |
| Consumer buffers | Initialize the entire required scale allocation, unused payload slots and storage tails. Extra Q scales for invalid rows may be finite nonzero values |

SM89 compilation is retained but is outside the current hardware acceptance matrix. GQA/MQA with different Q/KV head counts, unequal rank shard lengths, per-example ragged lengths, training, multi-node execution, and CUDA graph/`torch.compile` integration are outside this submission's validated scope.

#### SM90: prepared communicator versus the previous standalone API

8× NVIDIA H20-3e (SM90), PyTorch 2.13.0+cu130, CUDA 13.0; P=8, B=1, H=56, D=128, BF16, U=P×L. Both paths use preallocated output/payload/statistics-gather buffers and copy V scales into independently owned output storage. Each run uses 5 warmups, 5 rounds × 30 calls, alternating the compared paths; the reported synchronized wall time is the median of the per-round slowest-rank means. The frozen old implementation was run before (A1) and after (A2) the new implementation (B).

| Local sequence L | Previous standalone A1 (ms) | Previous standalone A2 (ms) | Prepared communicator B (ms) | Reduction versus previous standalone |
|---|---:|---:|---:|---:|
| 4736 | 0.9788 | 0.9867 | 0.9652 | 1.39%–2.18% |
| 4737 | 1.0891 | 1.0808 | 1.0255 | 5.12%–5.84% |

The timed interval includes local/final statistics, quantization/packing, statistics AllGather, payload AllToAll, unpacking and the V-scale copy. It excludes preparation/JIT, attention, output gather and full diffusion. Redundant Python validation between stages was removed while retaining input/output checks at each communicator call. CUDA kernels and communication volume are unchanged. These figures do not establish an attention-kernel or full-video speedup.

These September 17 measurements preceded the final Python module rename and public-export cleanup. That cleanup preserves executable numerical code, but its timings were not remeasured. No new SM120 timing or video/timeline collection is claimed.

<details>
<summary>Measured source provenance and reproduction</summary>

Both measurements used local snapshots based on `56c3b4c3fe94674f92e163f288c2e6cfabf126c2`; a branch HEAD alone does not identify the measured sources. File names below are those at measurement time, before `ulysses_lowp.py` became `_ulysses_lowp.py`.

| Source | Frozen A SHA-256 | Updated B SHA-256 |
|---|---|---|
| `flashinfer/comm/ulysses.py` | `54044b2d612a70e0d5cd2c902d3554a269bfadab99ee54a6eee7605a7f0b8c7a` | `90f88ec09ee3442fe8a88b131ce9e50934f9e756c4f86844464a5790a820ef84` |
| `flashinfer/comm/ulysses_lowp.py` | `1ca6d748ad3b75c32174aecf76c6f6c600f11aae066dc190453374271eddde05` | `f3267d536604569b7dfb48ca593e61f8d12d7a8914c62f7ebb49dc3a952597be` |

Both used benchmark SHA-256 `58c5d18f46c246f2534b56bbb2c1952e048f7044ddd02b464e5f4d1942547cef`. The checked-in benchmark now imports the private module and compares the current standalone implementation with the current communicator. To reproduce a historical cross-revision comparison, run it against the corresponding frozen sources above.

```bash
FLASHINFER_LOGLEVEL=0 FLASHINFER_TRACE_DUMP=0 \
  torchrun --standalone --nproc-per-node=8 benchmarks/comm/bench_ulysses_qkv.py \
  --batch 1 --heads 56 --head-dim 128 --dtype bfloat16 \
  --lengths 4736 4737 --warmup 5 --rounds 5 --iterations 30 \
  --output ulysses-qkv-microbench.json
```

</details>

The integration measurements below cover **historical D=128 runs on SM90 and SM120, before the D=64 extension**. Each architecture has its own measured revision and environment; neither integration table is a measurement of the current PR head or D=64.

#### SM90: 2026-09-11 working-tree snapshot

Environment: 8×H20-3e, UP8/TP1, FSDP off, AdaLN offload on, compile off; MiniMax-H3, 1248×704, 124 output frames at 24 fps, seed 2101. Baseline: BF16 input AllToAll followed by **CUDA per-warp Sage2** quantization and attention. Lowp: statistics AllGather plus low-precision input AllToAll and prequantized Sage2 attention.

| Measurement | CUDA Sage2 baseline | Lowp-Sage2 | Reduction |
|---|---:|---:|---:|
| GPU0 attention-input preparation, cumulative over 100 modules | 167.017 ms | 94.942 ms | 43.15% |
| 50-timestep pipeline, hummingbird case | 97.73 s | 96.28 s | 1.48% |
| 50-timestep pipeline, musician case | 97.75 s | 96.26 s | 1.52% |
| 50-timestep pipeline, waves case | 97.70 s | 96.39 s | 1.34% |

The preparation window starts at the first input preparation GPU operation and ends at the attention kernel, excluding QKV projection, attention itself and output AllToAll. The trace requested 3 timesteps and captured 2 updates × 50 layers; the video runs requested 50 timesteps and performed 49 updates. Each video case has one measured run, without repeat-run confidence intervals.

The measured benefit combines communication and layout handling; it does **not** establish that each quantization kernel is faster. For example, in the inspected single module, Q amax plus quantization took 38.368 µs in Sage2 and 68.544 µs in Lowp. Lowp uses V scale_max=2.25 while the measured Sage2 baseline uses 448, so this is not an isolated transport-only comparison.

The three historical Lowp videos were byte-identical after decoding to the preceding Lowp implementation. Similarity against BF16 varied by prompt; this does not establish universal quality parity with BF16 or stock Sage2.

[Integration environment, reproduction scripts and historical tables](https://github.com/DwenGu/sglang-minimax/blob/545c135c788e88365b05274484cb10f2538d705f/ULYSSES_LOWP_SM120_HANDOFF.md). That handoff pins an earlier FlashInfer revision; explicitly use this PR's revision when producing new acceptance evidence. The original local report records FlashInfer HEAD `481cb83a` with tracked-diff SHA-256 `e3757e774aede56e561a1d833a4d0ce91525cc803cc32638f6d70d1064952094`. A performance claim for the final PR head still requires a fresh measurement and reviewer-accessible artifacts.

#### SM120: 2026-09-13 UP8 integration measurements

Source: author-supplied `SM12-Report.md`, batch `results-sm120-up8-20260913-a01` (report SHA-256: `069a1b843782c0cda14ea38365c029d73c1e23a15a89ec1e30d3ca1b5128338e`). These are reported measurements from that batch, not a new run during this PR update.

Measured revisions (all three source trees reported clean): [FlashInfer `1c828322`](https://github.com/DwenGu/flashinfer/tree/1c8283228b97c1ce575cf3815d822f640eb40e4f), [SGLang `545c135c`](https://github.com/DwenGu/sglang-minimax/tree/545c135c788e88365b05274484cb10f2538d705f), and [stock SageAttention `d1a57a54`](https://github.com/thu-ml/SageAttention/tree/d1a57a546c3d395b1ffcbeecc66d81db76f3b4b5). This FlashInfer revision predates D=64 support and the current rebase. The later D=64/128 validation matrix in the Tests section is a separate run.

| Setting | Measured configuration |
|---|---|
| GPU / interconnect | 8× NVIDIA Graphics Device, CC 12.0, 73,415 MiB per GPU; PCIe/NUMA, no NVLink; no commercial model inferred |
| Software | PyTorch 2.13.0+cu130, CUDA 13.0, driver 580.95.05; Nsight Systems 2026.4.1 |
| Docker image | `lmsysorg/sglang@sha256:9e148f5ac788e856a06166bd6347a831831eb9fcfab4d1770874823a7c29a1a1` |
| Workload | MiniMax-H3 / FL2VA / speed / t2va; D=128; UP8 / TP1 / ring=1 |
| Memory / compilation | FSDP OFF, AdaLN online/offload ON, torch.compile OFF |
| Generation | 1248×704, 124 frames at 24 fps, seed 2101; flow_shift=12, audio_flow_shift=3 |
| Sampling | Warmup per backend, one formal run per prompt; 50 diffusion timesteps / 49 denoise updates |
| Profiling | Separate case1 run per backend; CUDA/NVTX, 3 timesteps / 2 updates / 100 DiT attention calls per GPU |

The BF16 reference actually used **Torch SDPA**, despite the requested `fa` setting: the fixed SGLang revision routes SM12.x to this backend. Sage2 used **CUDA per-warp Q32/K64**; Lowp used the same SM120 Sage attention core and V `scale_max=2.25`. Thus the main comparison is Lowp versus CUDA Sage2. BF16 comparisons also change the attention implementation and precision.

**50-timestep video runs.** Reductions below mean `(baseline - Lowp) / baseline`; they are elapsed-time reductions, not throughput multipliers. Pipeline and denoise are separate timers and must not be added.

| Prompt | BF16 SDPA pipeline (s) | CUDA Sage2 pipeline (s) | Lowp-Sage2 pipeline (s) | Reduction vs Sage2 | Reduction vs BF16 |
|---|---:|---:|---:|---:|---:|
| case1: hummingbird / hibiscus | 170.17 | 151.80 | 119.31 | 21.40% | 29.89% |
| case2: rainy sidewalk / musician | 170.28 | 152.24 | 119.91 | 21.24% | 29.58% |
| case3: waves / volcanic rocks | 170.33 | 152.48 | 120.05 | 21.27% | 29.52% |

| Prompt | BF16 SDPA denoise (s) | CUDA Sage2 denoise (s) | Lowp-Sage2 denoise (s) | Reduction vs Sage2 |
|---|---:|---:|---:|---:|
| case1 | 167.0945 | 148.7115 | 116.2086 | 21.86% |
| case2 | 167.1682 | 149.1258 | 116.7757 | 21.69% |
| case3 | 167.2079 | 149.3525 | 116.8975 | 21.73% |

**Separate 3-timestep CUDA/NVTX traces: complete input-preparation windows.** Each window starts at the first input-pack operation (BF16/Sage2) or local-stats operation (Lowp), and ends when the actual attention core starts. It includes communication, layout handling, statistics, and launch/wait gaps; it excludes QKV projection, QK norm/RoPE, attention computation, and output AllToAll.

| Backend | GPU0 sum of 100 windows (ms) | Per-call mean (us) | Range of per-GPU sums across 8 GPUs (ms) |
|---|---:|---:|---:|
| BF16 Torch SDPA | 2659.113 | 26591.126 | 2659.113–2782.819 |
| CUDA Sage2 | 2749.947 | 27499.469 | 2729.988–2787.158 |
| Lowp-Sage2 | 1474.711 | 14747.106 | 1474.711–1505.731 |

GPU0 preparation time decreased **46.37%** versus Sage2. This is a preparation-window reduction, not a full-pipeline reduction. The eight GPU totals must not be summed into an end-to-end latency.

**One attention module: GPU0, `denoising_step_1`, `transformer.blocks.25.attn`.** Stage rows are CUDA kernel-duration sums; zero means no independent kernel for that stage. Shared statistics and fused packing are charged to their actual rows. The final preparation window also includes gaps and may differ from the sum of kernel durations.

| Stage | CUDA Sage2 (us) | Lowp-Sage2 (us) |
|---|---:|---:|
| Q: amax + INT8 quantization | 53.984 | 94.079 |
| K: separate statistics + INT8 quantization | 139.520 | 135.999 |
| V: separate FP8 quantization | 27.680 | 56.768 |
| Shared K/V statistics | 0.000 | 93.088 |
| Statistics helpers / finalization | 0.000 | 67.264 |
| Statistics AllGather | 0.000 | 564.384 |
| Input AllToAll | 27905.831 | 13312.528 |
| Separate sending QKV pack | 301.088 | 0.000 |
| Receiver BF16 contiguous copies | 328.928 | 0.000 |
| V: separate padding / transpose | 153.088 | 0.000 |
| Receiver Lowp unpack / tail clearing | 0.000 | 147.392 |
| **Complete preparation window** | **28934.247** | **14541.806** |
| Attention core, excluded above | 9439.255 | 9437.141 |

The benefit comes mainly from less input communication and fewer separate BF16 layout copies. Q quantization and the V quantization stage are slower in this inspected Lowp module; the Sage core time is essentially unchanged. NCCL kernel durations include peer synchronization/waiting and do not measure pure link bandwidth. These PCIe/NUMA results should not be extrapolated to NVLink systems or SM90.

Quality scope: all nine videos decoded successfully. Full-frame SSIM versus this batch's local BF16 reference was 0.879573 / 0.822718 / 0.960857 for Lowp and 0.905763 / 0.802927 / 0.971350 for Sage2. The report records visible pose/composition differences and incomplete human full-playback acceptance; these similarity scores do not establish perceptual quality parity. Each prompt has one formal run, without repeated-batch statistics or confidence intervals.

Reproduce the fixed workload using the [pinned SGLang harness and environment instructions](https://github.com/DwenGu/sglang-minimax/blob/545c135c788e88365b05274484cb10f2538d705f/ULYSSES_LOWP_SM120_HANDOFF.md), with the measured revisions above. From the project root, after environment setup, choose fresh output and scratch directories:

```bash
H3_BENCH=./sglang-lowp-fi/python/sglang/multimodal_gen/benchmarks/ulysses_lowp
python "$H3_BENCH/run.py" --model-path ./models/MiniMax-H3 --up 8 --port 30041 \
  --mode videos --output ./results-sm120-up8-new --scratch ./work-sm120-up8-new
python "$H3_BENCH/run.py" --model-path ./models/MiniMax-H3 --up 8 --port 30041 \
  --mode timelines --output ./results-sm120-up8-new --scratch ./work-sm120-up8-new
```

## 🔍 Related Issues

None.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

Pre-commit passed for all 32 files changed in the unified-API update, including the communicator, private-module migration, tests, trace fixtures and documentation. The public API checker reported zero findings against PR base `2694e1e025e8e5cb8d9117299027ebe816229a33`. The static documentation checker exited successfully with 19 warnings for private trace/helper entries in `flashinfer.comm._ulysses_lowp` that are intentionally absent from the public API index; the public communicator API is documented. Hook installation and a clean full-repository `--all-files` run have not been verified for this submission, so those items remain unchecked.

## 🧪 Tests

### September 17 unified-API validation

The author confirmed that the unified communicator passed SM120 validation before the final internal-module rename/export cleanup. No new tested SHA, test counts or timings were supplied with that confirmation. The cleanup was then validated locally on SM90; it is not presented as a new SM120 hardware run.

Tested unified-API sources are committed as [`59e4fbe8`](https://github.com/DwenGu/flashinfer/commit/59e4fbe8aa2e6b524a4e03ae6d037427a0fa0378). Local environment: 8× NVIDIA H20-3e, PyTorch 2.13.0+cu130, CUDA 13.0, with the stock CUDA SageAttention2 extension. The following checks cover the unified API and the final private-module layout:

| Check | Result |
|---|---|
| Unified QKV API, Lowp primitive and boundary suites | 459 passed, 239 skipped, 0 failed; includes 14 unified-API cases |
| Real NCCL + stock CUDA Sage2, D64/P8/B2/H16/L65/U511/BF16 | All 8 ranks passed; six outputs and send/receive payloads matched the standalone path byte-for-byte, and actual Sage attention passed numerical acceptance |
| Ulysses trace checks | 58 passed, 3 skipped |
| Full trace-template consistency and registry checks | 1070 passed |
| Import/API boundary | Public communicator/types import; old public module/aliases absent; ordinary communication import does not eagerly load Lowp |
| Static checks and benchmark CLIs | Targeted pre-commit, Python syntax, diff checks and both benchmark `--help` commands passed |

The Lowp module has the same executable AST after removing docstrings as the snapshot before its rename. This cleanup changes Python import paths and public exports; CUDA/JIT/AOT sources and custom-op names are unchanged.

The ordinary communicator suite was also exercised during the preceding refactor. Its combined regression run reported 527 passed, 239 skipped and one TCP rendezvous `EADDRINUSE` startup failure; the affected existing close test passed when rerun with a fresh port. The final 459-test pass above is a separate, narrower selection and does not relabel that first attempt as failure-free.

### Historical D64/D128 hardware matrix

The numerical and performance history below retains its original revision and is not relabeled as a new SM120 run.

The historical validation matrix below was executed on commit `31fda23045612136c1ee64f2adcae24b9ec1204f` (branch `feat/ulysses-lowp-sage2`). SM90 runs were performed locally; SM120 results were reported by the author after executing the same validation plan. Both used PyTorch 2.13.0+cu130 and CUDA 13.0. These results retain their original tested-revision provenance. The September 15 review-fix validation is recorded separately below.

| Environment | SM90 | SM120 |
|---|---|---|
| GPUs | 8× NVIDIA H20-3e | 8× NVIDIA Graphics Device |
| Compute capability | 9.0 | 12.0 |
| Visible memory per GPU | Not recorded here | 73,415 MiB |

The SM120 device name is reported exactly as exposed by the system; no commercial model is inferred. The tables below describe the explicitly pinned historical run.

| Validation | SM90 | SM120 |
|---|---|---|
| Lowp unit tests, including D=64/128 | 432 passed, 239 skipped, 0 failed | 475 passed, 196 skipped, 0 failed |
| Ulysses trace tests | 26 passed, 3 skipped, 0 failed | 26 passed, 3 skipped, 0 failed |
| Full trace-template consistency suite | 774 passed | Full-suite result not reported |
| D=64 real NCCL collective + attention acceptance | 3 configurations passed | 3 configurations passed |
| D=128 real NCCL collective + attention acceptance | Not rerun in this validation batch | Same 3 configurations passed |
| D=64 Compute Sanitizer memcheck | 3 configurations, 0 errors | 3 configurations, 0 errors |
| D=128 regression against `1c828322` on the same architecture | 6/6 byte-fingerprint matches | 6/6 byte-fingerprint matches |
| Rank-local benchmark CLI | D=64 completed | D=64 and D=128 completed; stage timings recorded by the author |

The D=128 fingerprint comparison covers all sender payloads and rank-0 unpacked outputs in six FP16/BF16 cases at P=2/4/8. It compares old and new implementations on each machine; it does not require identical bytes across SM90 and SM120. Hardware/dependency skips are retained in the unit-test totals. SM89 compilation passed, but SM89 hardware validation is outside this matrix.

Distributed acceptance used real NCCL statistics AllGather and payload AllToAll, followed by Sage2 attention and an FP32 SDPA reference:

| P | B | H | L | U | Dtype | V case | SM90 D=64 | SM120 D=64 | SM120 D=128 |
|---|---|---|---|---|---|---|---|---|---|
| 8 | 2 | 16 | 65 | 511 | BF16 | Random with two zero channels | Passed | Passed | Passed |
| 4 | 1 | 28 | 129 | 257 | FP16 | Random with two zero channels | Passed | Passed | Passed |
| 2 | 1 | 8 | 257 | 513 | BF16 | All zero | Passed | Passed | Passed |

No failures were reported in these targeted runs. The full-repository checks and latest-head upstream CI remain outstanding, so the all-tests checkbox is left unchecked. The D=64/128 rank-local benchmark completion establishes that both dimensions execute; it does not establish a speedup. The separate historical D=128 SM120 integration measurements and their measured revision are reported in the Performance evidence section.

Tests cover independent global-grid references, groups crossing rank boundaries, padded tails, multiple statistics chunks, long scale arrays, strided projection inputs, output reuse, mismatched layout dimensions, zero V, and partition invariance with fixed statistics. Nonzero-channel attention acceptance requires cosine ≥0.999 and relative L1 ≤1.10× stock CUDA Sage2; complete Lowp output must be finite, and all-zero V must produce exactly zero output. Historical video similarity is separate from these numerical gates.

Example commands from the FlashInfer checkout, with the matching CUDA SageAttention2 extension installed:

```bash
python -m pytest tests/comm/test_ulysses_qkv.py tests/comm/test_ulysses_lowp.py tests/comm/test_ulysses_lowp_boundary.py -q -rs
python -m pytest tests/trace/ -k ulysses -q -rs
python -m pytest tests/trace/test_fi_trace_template_consistency.py -q

torchrun --standalone --nproc-per-node=8 benchmarks/comm/validate_ulysses_lowp.py \
  --compare-communicator --head-dim 64 --local-sequence 65 --used 511 --heads 16 --batch 2 \
  --output lowp_d64_up8.json

python benchmarks/comm/bench_ulysses_lowp.py \
  --head-dim 64 --local-sequence 257 --num-heads 8 --world-size 8 \
  --cuda-events --hot-l2 --dry-run-iters 3 --repeat-iters 10
```

The rank-local benchmark excludes collective communication. Use the distributed validator for communication correctness, `bench_ulysses_qkv.py` for statistics/quantization/input communication latency, and the integration harness for application performance.

### September 15 review-fix validation

Head tested on September 15: `56c3b4c3fe94674f92e163f288c2e6cfabf126c2`, one feature commit based on upstream main `2694e1e025e8e5cb8d9117299027ebe816229a33`. That conflict was resolved by keeping both upstream head-chunk exports and the then-public Lowp exports; this revision removes the latter in favor of the communicator API. The earlier MoE JIT dispatch is preserved.

The update adds shared C++ FFI row-alignment validation, fixes standalone trace initialization and unpack fixture sizing, registers the shared SM89/SM120 AOT module for either target, and completes API documentation. The scale-slot formula remains `ceil((L+G-1)/G)`: a documented `L=50, G=32, rank=1` counterexample requires three slots. CUDA kernel arithmetic and payload layout are unchanged.

Validation environment: 8 NVIDIA H20-3e GPUs (SM90), PyTorch 2.13.0+cu130, CUDA 13.0, using `lmsysorg/sglang@sha256:9e148f5ac788e856a06166bd6347a831831eb9fcfab4d1770874823a7c29a1a1`. Runtime tests cover the updated feature sources; subsequent edits changed documentation and the commit message only. The final rebase picked up upstream ragged prefill changes; all 28 PR-modified files remained byte-identical to the locally checked `8cd63bbe` snapshot.

| Check | Result |
|---|---|
| Lowp primitive and boundary suites, D64/D128 | 445 passed, 239 skipped, 0 failed |
| `test_fi_trace.py`, including six standalone JSON init cases | 58 passed, 0 failed |
| Trace init, registry, template consistency and standalone source suites | 1519 passed, 231 skipped, 0 failed |
| AOT inventory, including SM89-only, SM90, SM120 and mixed targets | 16 passed, 0 failed |
| Real NCCL + stock CUDA SageAttention acceptance | D64 and D128 each passed P8 BF16, P4 FP16 and P2 BF16 zero-V; 6/6 configurations |
| Compute Sanitizer memcheck on direct FFI alignment tests | 12 passed, 0 errors |
| SM90 module and shared SM89/SM120 fat binary | Both compiled; SM90 executed |
| Public API and static documentation checks against main | 0 findings in each |
| Sphinx HTML documentation | Build passed; Lowp documentation rendering errors fixed |
| Applicable pre-commit hooks across 28 changed files | All passed |
| Conflict preservation smoke checks | Both Ulysses export blocks and legacy/SM100/SM103 MoE JIT generators passed |

Commands below run from the FlashInfer checkout in the configured environment:

```bash
python -m pytest tests/comm/test_ulysses_lowp.py tests/comm/test_ulysses_lowp_boundary.py -q
python -m pytest tests/trace/test_fi_trace.py tests/jit/test_ulysses_lowp_aot.py -q
compute-sanitizer --tool memcheck --error-exitcode 99 \
  python -m pytest tests/comm/test_ulysses_lowp.py -k ffi_shard_alignment -q
for d in 64 128; do
  torchrun --standalone --nproc-per-node=8 benchmarks/comm/validate_ulysses_lowp.py \
    --head-dim "$d" --dtype bfloat16 --output "result-d$d-p8.json"
  torchrun --standalone --nproc-per-node=4 benchmarks/comm/validate_ulysses_lowp.py \
    --head-dim "$d" --dtype float16 --output "result-d$d-p4.json"
  torchrun --standalone --nproc-per-node=2 benchmarks/comm/validate_ulysses_lowp.py \
    --head-dim "$d" --dtype bfloat16 --zero-v --output "result-d$d-p2-zero-v.json"
done
```

An additional all-operations `tests/trace/` run was interrupted during unrelated attention JIT compilation; it is not counted as a pass. The completed trace checks are listed above. Hardware-specific skips are not passes. SM89 and SM120 hardware execution was outside that September 15 SM90 run. Upstream GPU CI on the final head requires maintainer authorization; local validation does not replace that check.

## 🔬 Experimental Track

```experimental-tests
#
#
```

## Reviewer Notes

Please review collective preparation and failure handling, workspace ownership and output reuse, per-call validation, global group ownership and live-tail handling, the K mean rounding point, the zero-amax V contract, and consumer scale/buffer sizes for both head dimensions. The design document records these invariants.

Kernel definitions remain framework-independent under `include/`; tensor bindings live under `csrc/`. Both architecture modules participate in JIT/AOT. Trace templates infer D from inputs, with generated D=64 and D=128 examples included.

The submission is intended for the normal feature track. API documentation/index coverage has been checked locally. Reviewer-accessible performance artifacts and upstream full-repository CI on the final PR head remain required for final review.

## Summary by CodeRabbit

* **New Features**
  * Added prepared Ulysses QKV exchange for SageAttention, including reusable workspaces and output buffers.
  * Added low-precision INT8 Q/K and FP8 V quantization with Sage-compatible packing and unpacking for supported GPUs and head dimensions 64 and 128.
  * Added architecture-specific builds, tracing, benchmarking, and distributed validation tools.

* **Documentation**
  * Updated design and API documentation for prepared QKV exchange, supported layouts, and usage constraints.

* **Tests**
  * Added comprehensive coverage for numerical correctness, boundaries, layouts, tracing, workspace behavior, and architecture-specific support.


flashinfer/comm/ulysses_lowp.py
```diff
@@ -0,0 +1,2144 @@
...
+        return module.ulysses_lowp_abi_version()
+
+    return SimpleNamespace(
+        ulysses_lowp_k_sum_v_amax=ulysses_lowp_k_sum_v_amax,
+        ulysses_lowp_q_grouped_amax=ulysses_lowp_q_grouped_amax,
+        ulysses_lowp_k_grouped_amax=ulysses_lowp_k_grouped_amax,
+        ulysses_lowp_quant_q_int8_pack=ulysses_lowp_quant_q_int8_pack,
+        ulysses_lowp_quant_kv_int8_fp8_pack=ulysses_lowp_quant_kv_int8_fp8_pack,
+        ulysses_lowp_quant_q_int8_pack_fused=ulysses_lowp_quant_q_int8_pack_fused,
+        ulysses_lowp_quant_kv_int8_fp8_pack_fused=ulysses_lowp_quant_kv_int8_fp8_pack_fused,
```
Anerudhan, line 326: ulysses_lowp_compiled_q_group / _k_group / _head_dim seems to be missing here? 

Would it not error out with code here [https://github.com/flashinfer-ai/flashinfer/pull/4924/changes#diff-c980d80a627cd4080202ca593bb099674dbc799268124dab3cd0c228ba2aa24fR446]( https://github.com/flashinfer-ai/flashinfer/pull/4924/changes#diff-c980d80a627cd4080202ca593bb099674dbc799268124dab3cd0c228ba2aa24fR446)

flashinfer/comm/ulysses_lowp.py
```diff
@@ -0,0 +1,2144 @@
...
+    )
+    def ulysses_lowp_quant_v_fp8_with_scale(
+        input: torch.Tensor,
+        scale: torch.Tensor,
+        output: torch.Tensor,
+        enable_pdl: bool,
+    ) -> None:
+        module.ulysses_lowp_quant_v_fp8_with_scale(input, scale, output, enable_pdl)
+
+    @register_custom_op("flashinfer::ulysses_lowp_abi_version", mutates_args=[])
```
Anerudhan: I dont see definition for this?

csrc/ulysses_lowp.cu
```diff
@@ -0,0 +1,778 @@
...
+  CHECK_DIM(2, output);
+  CHECK_INPUT_TYPE(output, dl_uint8);
+  CHECK_DEVICE(q, q_amax_final);
+  CHECK_DEVICE(q, output);
+
+  const int64_t batch_size = q.size(0);
+  const int64_t local_sequence = q.size(1);
+  const int64_t num_heads = q.size(2);
+  const int64_t head_dim = q.size(3);
+  const int64_t local_heads = num_heads / world_size;
```
Anerudhan, line 326: Do we need to assert if local_heads is a multiple of world_size?

flashinfer/comm/__init__.py
```diff
@@ -62,6 +62,47 @@
...
+# NOTE: never export a function named exactly 'ulysses_lowp' here — it would
+# shadow the flashinfer.comm.ulysses_lowp submodule (see ulysses.py's
+# ulysses_a2a merge note). Generic names (capability) are exported
+# with the ulysses_lowp_ prefix, following the vllm_meta_size precedent.
+from .ulysses_lowp import StatsContext as StatsContext
+from .ulysses_lowp import UlyssesLowpSageLayout as UlyssesLowpSageLayout
+from .ulysses_lowp import V2GStats as V2GStats
+from .ulysses_lowp import aligned_length as aligned_length
+from .ulysses_lowp import capability as ulysses_lowp_capability
+from .ulysses_lowp import finalize_stats as finalize_stats
```
Anerudhan: Can we name this as ulysses_finalieze_stats similar to ulysses_lowp_capability. Ditto for all the other re-exports

Anerudhan (changes requested):

aleozlx:
## PR Review Screening

**CI verdict:** ✅ auto-run ok
**Review category:** `live` (rule fired: C1.2 — a net-new public interface: a 2144-line `flashinfer.comm.ulysses_lowp` module, ~17 `@flashinfer_api` entry points, and a versioned payload ABI exported from `flashinfer/comm/__init__.py`)
**Blocking checks:** none
**Release blocker:** no — net-new feature; no hang/crash/corruption or prior-release regression claim (C3.4)
**Early stop:** no

### Security
| Q | Answer | Evidence |
|---|--------|----------|
| S3 template obligations | met | 5/5 checked; Description carries a file-by-file table, an explicit scope/platform note, and the test inventory; Reviewer Notes flag the load-bearing kernel-math constraint |

### Packaging
| Q | Answer | Evidence |
|---|--------|----------|
| C1.1 external dependency bump | no | No requirements/pyproject/submodule/action changes. The `sageattention` fork import in the tests is `try/except ImportError`-guarded and gates one test only — it is not a declared dependency |
| C1.2 public API changes | yes — **net-new interface** | `flashinfer/comm/ulysses_lowp.py` (2144 lines) exported via `flashinfer/comm/__init__.py` (+41): `capability`, `payload_spec`, `k_sum_v_amax`, `q_grouped_amax`, `k_grouped_amax`, `quant_and_pack` / `unpack_for_sage`, the split `quant_q_into_payload_fused` / `quant_kv_into_payload_fused` overlap path, `stats_protocol_for`, `required_alignment` / `aligned_length`, `StatsContext` / `V2GStats`, and `ulysses_lowp_abi_version()`. The **payload ABI (v3)** is the durable part: it is a wire contract an out-of-repo consumer (SageAttention2) decodes, so its layout is harder to change later than the Python signatures are. Narrow platform scope (`sm_120a` only, `head_dim == 128` required, `world_size ∈ {2,4,6,8}` with P=6 untested) — but the surface is large enough that the guidance doc's "disposable model-specific op" downgrade does not obviously apply |
| C1.3 AOT registration | yes | `gen_ulysses_lowp_module()` added in `flashinfer/jit/comm.py` **and** registered in `flashinfer/aot.py` behind `if has_sm120`, matching the generator's `sm_120a` pin; `TraceTemplate`s added in `flashinfer/trace/templates/comm.py` (+471) with the module listed in `tests/trace/template_registry.py`. No `fi_trace_out/` goldens — consistent with existing convention rather than a gap: `tests/trace/example.py` never exercises `comm`, so no comm template has one |

### Presentation
| Q | Answer | Evidence |
|---|--------|----------|
| C2.1 perf claim backed by data | yes | GPU named (RTX PRO 6000 Blackwell / SM120); end-to-end ~9–11% lower step latency vs the BF16-A2A + SageAttention2 prequant path and ~30% vs BF16 baseline, on an identified workload (sglang MiniMax-H3, Ulysses-8, 8×SM120); `benchmarks/comm/bench_ulysses_lowp.py` (328 lines, `bench_gpu_time`, cold-L2 default, JSON output) ships the reproducible per-rank path. Caveat rather than a violation: the e2e figures come from an out-of-repo pipeline, so only the per-rank benchmark is reproducible from here |

### Implementation
| Q | Answer | Evidence |
|---|--------|----------|
| C3.1 experimental-track declared | no | No experimental label or decorator; `flashinfer/comm/` is a shipped subsystem |
| C3.2 shared/durable areas touched | no | The whole PR is **6453+/0−** — nothing is modified, only added. `include/flashinfer/comm/ulysses_lowp.cuh` and `csrc/ulysses_lowp.cu` are new files; no shared header, base class, or dispatch path is reshaped. The edits to `flashinfer/comm/__init__.py`, `jit/comm.py`, `aot.py` and `template_registry.py` are registrations and exports |
| C3.3 tests match behavior change | yes | `tests/comm/test_ulysses_lowp.py` (1461 lines, 36 test functions / 83 cases) behind a `requires_sm120` skipif, and SM120 runners exist in the CI matrix, so it executes. Covers protocol-2 ≡ protocol-3 on 128-aligned shards, fused ≡ split pack, interleaved-view ≡ contiguous, PDL on/off, unaligned lengths (65/193/1180), unpack round-trips and scale widths, and the ABI layout/tail invariants. One named caveat: `test_bit_parity_vs_fork` — the only test comparing against the actual upstream fork — is gated on `import sageattention.ulysses_v2g_ops` and skips when that is absent, so the "bit-exact port" claim rests in CI on the in-repo anchors, not on the fork itself |
| C3.4 critical fix | no | New feature; no hang/crash/IMA/corruption or prior-release regression claim |

### Experimental track
| Q | Answer | Evidence |
|---|--------|----------|

**Notes for the maintainer**
- Still a **draft**, and the author offers to split it into core primitives → protocol 2 → optimizations → trace/aot/benchmark, with the commit history already organized that way. For 6.5k net-new lines routed `live`, taking that offer is probably the cheapest route to a real review — worth answering explicitly rather than leaving open.
- **Provenance is a deliberate design constraint here, not incidental resemblance.** The kernels are described as a bit-exact port of the thu-ml/SageAttention fork's fused quantization kernels, mirroring multiply/divide order and specific PTX conversions verbatim because that *is* the payload-ABI contract. Attribution is present (Apache-2.0 headers plus a source link in `ulysses_lowp.cuh`); worth a maintainer confirming that form meets the project's bar for adapted third-party code, since the copying is intentional and load-bearing.
- **Everything about this lands on one architecture.** The JIT spec pins `-gencode sm_120a`, AOT prebuilds only when SM120 is in the arch list, and tests skip elsewhere — because the byte-anchors hold under `sm_120a` but not necessarily under the default `compute_120f` normalization, where FMA semantics can differ by ULPs. That makes durable SM120 CI coverage a precondition for this code being testable at all, which is worth confirming before merge rather than after.

flashinfer/comm/ulysses_lowp.py
DwenGu: ulysses_lowp_abi_version was dead code: no callers existed anywhere in the codebase and it was never exported from __init__.py. The "stats protocol 3 / ABI v3" terminology it represented has been fully superseded by the ALIGNED / BOUNDARY_MERGE protocol names introduced in a prior commit. Rather than returning a meaningless literal 3, we removed the function and its TVM_FFI_DLL_EXPORT_TYPED_FUNC registration entirely. The module docstring was updated to match the current naming.

flashinfer/comm/ulysses_lowp.py
DwenGu: These three functions are defined and exported from csrc/ulysses_lowp_sm90.cu via TVM_FFI_DLL_EXPORT_TYPED_FUNC (added in 484c26ec). is_supported() calls them only after the SM90 JIT module is successfully loaded via get_ulysses_lowp_sm90_module(), so they are resolved at module-load time rather than at import time — no error occurs. A follow-up (95fe1c23) also added compiled_head_dim() to the check, making it a full three-field verification.

csrc/ulysses_lowp.cu
DwenGu: Yes, done in 95fe1c23. TVM_FFI_ICHECK_EQ(num_heads % world_size, 0) << "num_heads must be divisible by world_size" is now placed before all four occurrences of local_heads = num_heads / world_size in csrc/ulysses_lowp.cu, converting silent integer truncation into an early readable assertion failure.

flashinfer/comm/__init__.py
DwenGu: Done in 6709474c. All 28 function re-exports now carry the ulysses_lowp_ prefix (e.g. finalize_stats → ulysses_lowp_finalize_stats, local_stats → ulysses_lowp_local_stats, unpack_for_sage → ulysses_lowp_unpack_for_sage etc.), following the ulysses_lowp_capability precedent already in the file. Class re-exports (StatsContext, V2GStats, UlyssesLowpSageLayout*) and JIT helpers (gen_ulysses_lowp_module, get_ulysses_lowp_module) are unchanged as they already carry sufficient context.

DwenGu:
Thanks for the detailed screening report. A few updates since this was generated:

### Note 3 — Architecture coverage

Since the initial screening, the scope has expanded beyond SM120:

- **SM89 (Ada L40S / RTX 4090)** — `484c26ec` adds SM89 support via a fat binary
  (`-gencode=arch=compute_89,code=sm_89` + `sm_120a`). Same Q_GROUP=32 / K_GROUP=64
  tile sizes as SM120 (both are HMMA warp-level per_warp_int8), so no new kernel code
  is needed.
- **SM90 (H100 / H200)** — `484c26ec` introduces `UlyssesLowpSageLayoutSM90` with a
  dedicated `csrc/ulysses_lowp_sm90.cu` compiled with `sm_90a` (Q_GROUP=16 /
  K_GROUP=128 to match SageAttention2's WGMMA tile sizes).
- **Runtime guard fix** — `4487ff62` fixes an inconsistency where `_require_sm120`
  was rejecting SM89 devices at runtime despite `is_supported()` returning `True` for
  them.


### Note 1 — PR split

Happy to split along the existing commit structure (core primitives → protocol 2 →
optimizations → trace/aot/benchmark) if that is the maintainers' preference — the
history is already organized that way. Please advise and we will proceed accordingly.

### Note 2 — Provenance

The bit-exact port is intentional and load-bearing: preserving the multiply/divide
order and specific PTX conversions verbatim *is* the payload ABI contract between this
side and the SageAttention2 decoder. Apache-2.0 headers and a source link to the
`thu-ml/SageAttention` fork are present in `ulysses_lowp.cuh`. Happy to add a more
explicit attribution block if the project requires a different form.

csrc/ulysses_lowp.cu
```diff
@@ -0,0 +1,579 @@
...
+void check_shape_4d(const TensorView& t, const char* name, int64_t s0, int64_t s1, int64_t s2,
+                    int64_t s3) {
+  TVM_FFI_ICHECK(t.size(0) == s0 && t.size(1) == s1 && t.size(2) == s2 && t.size(3) == s3)
+      << name << " has shape (" << t.size(0) << ", " << t.size(1) << ", " << t.size(2) << ", "
+      << t.size(3) << "), expected (" << s0 << ", " << s1 << ", " << s2 << ", " << s3 << ")";
+}
+
+void check_v2g_shard(const TensorView& tensor, const char* name) {
+  TVM_FFI_ICHECK_EQ(tensor.device().device_type, kDLCUDA) << name << " must be a CUDA tensor";
+  TVM_FFI_ICHECK_EQ(tensor.stride(-1), 1) << name << " must be contiguous at the last dimension";
```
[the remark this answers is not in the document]
DwenGu: Fixed in `56c3b4c3`. The shared C++ checker now validates dimensionality before indexing strides, a 16-byte-aligned data pointer, and all three outer strides modulo eight FP16/BF16 elements. Both architecture modules use this checker. Direct-FFI tests cover misaligned offsets/strides and valid strided projection views for D64/D128 and FP16/BF16. On H20-3e: 12 passed; Compute Sanitizer memcheck: 0 errors.

docs/design_docs/ulysses_lowp.md
```diff
@@ -0,0 +1,189 @@
...
+| Section | Storage |
+|---|---|
+| Q | B*L*h*D INT8 bytes |
+| K | B*L*h*D INT8 bytes |
+| V | B*L*h*D FP8 bytes |
+| Q scales | B*h*slots(L,Q_GROUP) FP32 values |
+| K scales | B*h*slots(L,K_GROUP) FP32 values |
+| Alignment tail | round up chunk size to 128 bytes; zero |
+
+slots(L,G)=ceil((L+G-1)/G), an upper bound for all source offsets.
```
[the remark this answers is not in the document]
DwenGu: Checked against both implementations: they use `(L + 2*G - 2) // G`, which equals the documented `ceil((L+G-1)/G)`. This is an upper bound on the groups a shard can touch at an arbitrary global offset. With L=50, G=32, rank=1 owns tokens 50..99 and touches three groups; `ceil(L/G)` would allocate only two slots. I retained the formula and added this counterexample to the design doc and structural tests in `56c3b4c3`.

flashinfer/aot.py
```diff
@@ -816,10 +813,22 @@ def gen_all_modules(
...
+            gen_ulysses_lowp_sm90_module,
             gen_vllm_comm_module,
         )
 
         jit_specs.append(gen_comm_alltoall_module())
+        if has_sm120:
+            # ulysses_lowp: byte-anchored against its SageAttention golden under
+            # sm_120a only (the generator pins -gencode sm_120a; see
+            # gen_ulysses_lowp_module), so it is prebuilt for that target alone.
+            jit_specs.append(gen_ulysses_lowp_module())
```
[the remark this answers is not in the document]
DwenGu: Fixed in `56c3b4c3`. AOT detects SM89 explicitly and registers the shared module when either SM89 or SM120 is present, once for mixed targets. Its existing SM89/SM120 fat binary requires CUDA >=12.8 because it emits sm_120a; this requirement is documented and version-gated. The 16-case inventory test covers SM80/SM86 exclusion, SM89-only, SM90, SM120, mixed targets, CUDA version gating, and add_comm=False. The shared fat binary compiled with CUDA 13.0; SM89 hardware was not available for runtime validation.

tests/trace/fi_trace_out/ulysses_lowp_q_grouped_amax_h8_d64_p8.json
```diff
@@ -0,0 +1,72 @@
...
...eeze(-1)).to(torch.float8_e4m3fn)\n    inv_permute = [0] * (prefix_ndim + 4)\n    for i, p in enumerate(permute_dims):\n        inv_permute[p] = i\n    w_fp8 = q_blocks.permute(*inv_permute).reshape(*prefix, R, C).contiguous()\n    return w_fp8, scales\n\n\n__all__ = [\n    \"make_paged_kv_indices\",\n    \"make_ragged_indptr\",\n    \"make_uniform_qo_indptr\",\n    \"make_pos_ids\",\n    \"make_rope_cos_sin_cache\",\n    \"make_probs\",\n    \"make_logits\",\n    \"fp8_safe_randn\",\n    \"per_tensor_fp8_quantize\",\n    \"fp8_block_quant_1d\",\n    \"fp8_block_quant_2d\",\n]\n\n# ----- init -----\ndef _ulysses_lowp_q_grouped_amax_init(\n    *,\n    batch: int = 1,\n    local_sequence: int = 4736,\n    q_slots: int = 0,  # derived\n    num_heads: int = 56,\n    head_dim: int = _ULYSSES_LOWP_HEAD_DIM,\n    rank: int = 0,\n    world_size: int = 8,\n    device: str = \"cuda\",\n    seed: int = 0,\n):\n    del q_slots  # derived from local_sequence: slots(L, 32)\n    torch.manual_seed(seed)\n    q = torch.randn(\n        batch, local_sequence, num_heads, head_dim, dtype=torch.bfloat16, device=device\n    )\n    return {\"q\": q, \"rank\": int(rank), \"world_size\": int(world_size)}\n"
```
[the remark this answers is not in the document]
DwenGu: Fixed in `56c3b4c3`. The three init templates now have self-contained literal defaults; D64/D128 are still supplied through the trace axes. Regenerated all six affected JSON definitions. Six regression cases execute the embedded init in an empty namespace, instantiate CPU inputs with the requested D, and compare the embedded source with regenerated output. All passed.

tests/trace/test_fi_trace.py
```diff
@@ -1435,3 +1387,46 @@ def test_nvfp4_append_trace_json_init_is_self_contained():
...
+                k_amax_final=torch.empty(1, 8, 2),
+                v_scale_global=mean.float(),
+                rank=0,
+                world_size=8,
+            ),
+        ),
+        (
+            lowp.unpack_for_sage,
+            dict(
+                recv_u8=torch.empty(8, 16384, dtype=torch.uint8),
```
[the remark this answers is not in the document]
DwenGu: Fixed in `56c3b4c3`. The unpack fixture now obtains chunk_bytes from payload_spec(), giving 12544 bytes for D64 and 25088 for D128 at B=1, L=65, H=8, P=8. The full test_fi_trace.py suite passed (58 tests).

csrc/ulysses_lowp.cu
DwenGu: @Anerudhan The PR is updated to `56c3b4c3`, rebased onto current main with both Ulysses export blocks retained. The previously added head-divisibility guard remains in place; the shared FFI checker now also rejects misaligned pointers/strides. Trace init, AOT registration and API documentation fixes are included. The PR Tests section records the new SM90 validation and preserves the separate historical SM120 results. Could you take another look? Upstream GPU CI still needs authorization from a maintainer; local validation does not replace that gate. The formal re-request-review endpoint returned 404, so I am following up in this thread.

docs/design_docs/ulysses_lowp.md
```diff
@@ -0,0 +1,275 @@
...
+Fused amax/pack and aligned-only receiver implementations are deferred.
+
+## Prepared communicator API
+
+Use the existing `UlyssesCommunicator` with two additional methods:
+
+```python
+with UlyssesCommunicator(group, max_bytes=3 * q.nbytes, dtype=q.dtype,
+                         device=q.device, backend="auto") as comm:
+    workspace = comm.prepare_qkv(q.shape, used_sequence=U)
```
forrestl111: What's the definition of `U`?

docs/design_docs/ulysses_lowp.md
```diff
@@ -0,0 +1,275 @@
...
+    encoded = comm.scatter_qkv(q, k, v, workspace=workspace)
```
forrestl111: Could we specify the quantization recipe of q/k/v? If we are not using `Sage2`, e.g., [vc attention,](https://arxiv.org/html/2609.15810v1) , we will need a method to control the quantization recipe.

PR #4686, opened 2026-08-22
perf(gemm): optimize CuTe DSL W4A16 dense GEMM

zianglih:
## 📌 Description

@humansand

This optimizes the SM100/SM103 CuTe DSL dense W4A16 `mm_bf16_fp4` path added by #4466.

- Compile the SM100/SM103 dense kernel at CuTe optimizer level 3 instead of its explicit level-2 override. SM12x compilation is unchanged.
- Make raster direction a full autotune axis: every one of the 15 structural tile/cluster tactics now has both M-major and N-major variants, for a 30-tactic Cartesian product.

The final review diff is limited to `flashinfer/gemm/gemm_bf16_fp4_cute_dsl.py`. The fail-closed CUPTI evidence harness is retained in commit [`100e9527`](https://github.com/flashinfer-ai/flashinfer/commit/100e95275d55280c110c66e9a4693b07b86ff4d4), with its [orchestrator](https://github.com/flashinfer-ai/flashinfer/blob/100e95275d55280c110c66e9a4693b07b86ff4d4/benchmarks/bench_dense_w4a16_sm100.py) and [worker](https://github.com/flashinfer-ai/flashinfer/blob/100e95275d55280c110c66e9a4693b07b86ff4d4/benchmarks/bench_dense_w4a16_sm100_worker.py) intentionally absent from the final tree.

The dense kernel already retains the intended W4A16 architecture from the shared MoE design: the tensor-wide FP32 weight scale is applied to the FP32 accumulator in the epilogue, and the CTA uses two four-warp transform groups with the full 65,536-register allocation. This PR does not change those contracts, the public API, numerical ordering, warp specialization, or pipeline stages.

## 🔍 Related Issues

- Tracked in #4561
- Follow-up to #4466

### Environment and workload

- Image: `nvcr.io/nvidia/pytorch:26.05-py3`
- Devbox: `c2`, namespace `infra`, host `hu-pdx-117`; 8 x NVIDIA B300 SXM6 AC (SM103), measurements pinned to GPU 0
- GPU 0: UUID `GPU-ee0843de-7ab2-7b46-8af4-1344b209180a`, 1100 W power limit, 2032 MHz maximum SM clock
- Driver: `590.48.01`
- Python: `3.12.3`
- PyTorch: `2.12.0a0+5aff3928d8.nv26.05`
- PyTorch CUDA / system nvcc: `13.2` / `13.2.78`
- FlashInfer Python: `0.6.18` (editable checkout)
[...]

PR #4941, opened 2026-09-03
perf(gemm): cute-dsl bf16-nvfp4 dense gemm fix to include no-autotune fallback tactic

jimmyzho:
## 📌 Description

Two fixes to the SM100/SM103 CuTe-DSL dense W4A16 (`mm_bf16_fp4`) path. Both affect only the no-autotune path; autotuned callers are untouched by change 2 and lose ~2.7% from change 1 (see below).

### 1. Restore `--opt-level 2` (regression from #4686)

#4686 dropped this path's explicit `--opt-level 2` and let the CuTe DSL default (3) apply. On B300 that is a 1.5-1.8x regression on the shapes the perf CI runs:

| case | m | before (O3) | after (O2) |
|---|---:|---:|---:|
| 4K-hidden-prefill | 256 | 0.0952 ms | 0.0523 ms |
| 4K-hidden-decode | 1 | 0.0183 ms | 0.0120 ms |

Measured per tactic, **O3 is ~2-3% faster on 26 of the 30 tactics and 1.44-1.81x slower on exactly 4** — cluster `(1,1)` + `raster_along_m=True` + a narrow N tile, the same 4 at m=1 and m=256. The no-autotune path always selects `((128, 8, 256), (1, 1), True)`, the worst cell in that space, which is why #4686's autotuned benchmarks reported a real ~1.03x win while CI regressed. Reverting the level costs autotuned callers that ~2.7% and buys back 1.8x for everyone who does not autotune.

Not register pressure: `ncu` shows 128 registers/thread and zero local-memory traffic under both levels with an identical launch config; issued warps/scheduler drops 0.67 -> 0.39 at unchanged occupancy. The kernel goes latency-bound, so there is no register budget to retune. Worth reporting the cliff itself to the CuTe DSL team separately.

The 15 -> 30 raster-direction tactic expansion from #4686 is kept — it is inert on the no-autotune path and only widens the autotuner's search.

### 2. Shape-aware no-autotune default tactic

`forward()`'s `tactic == -1` path took `valid_tactics[0]`, which is always the smallest row tile (`route_tile=8`) regardless of `m`, because `get_valid_tactics()` filters on legality only. Each CTA streams all of K for its tile, so weights are re-read once per row tile — `ceil(m/8)` of them. That is harmless at m <= 8 (one tile) and degrades linearly above it: measured 2.1-5.1x off the best tactic at m=128, 6.1-9.1x at m=512, 8.7-10.2x at m=2048. Every caller that does not autotune pays this, including the perf CI suite.

`_select_sm100_bf16_fp4_default_tactic()` now ranks by `waves x per-CTA bytes` (`tile_n` FP4 weights at 1/2 byte, `tile_m` BF16 activations at 2; K is common to all tactics and cancels). The two terms oppose each other, so the optimum is an interior tile that widens with `m`. Ties go to `raster_along_m=False`, which also keeps the default clear of the level-3 cliff above.

Rejected on measurement, so they are not retried: ranking on wave count alone picks a 3-wave tactic over a faster 4-wave one at m=2048, n=k=4096 (it ignores per-CTA cost), and breaking ties toward the 2-CTA cluster helps m=512 but is worse overall (geomean 1.12 vs 1.07).

---

## ⏱️ Performance

End-to-end through `mm_bf16_fp4`, no `--autotune`, B200 (148 SMs), CUPTI, `--num_iters 30 --dry_run_iters 5`. Speedup of change 2 over the previous default, at `--opt-level 2`:

| m | (n,k)=(4096,4096) | (6656,19968) | (19968,6656) |
|---:|---:|---:|---:|
| 1 / 8 | 1.00x | 1.00x | 1.00x |
| 32 | 1.00x | 1.84x | 2.14x |
| 128 | 2.65x | 4.87x | 4.64x |
| 512 | 5.50x | 7.28x | 7.71x |
| 2048 | 7.94x | 7.26x | 9.27x |

Geomean 2.64x, max 9.27x, no shape slower than before. Against per-tactic ground truth (30 tactics x 18 shapes) the heuristic lands 1.069 geomean from the per-shape optimum, worst 1.336x, versus 2.823 / 10.223 for the rule it replaces. It is a heuristic to stop the untuned path being catastrophic, not a replacement for autotuning.

---

## 🧪 Tests

*   **18/18 shapes:** the tactic actually launched matches the selector's independently computed prediction (verified by intercepting `_launch_cute_dsl_sm100`); all outputs finite.
*   **9/9 shapes:** pass `flashinfer_benchmark.py --refcheck` (fp32 reference).
*   **Linting:** `ruff check` and `ruff format` clean.

> **Note:** Measured on B300 SXM6 (cc 10.3) for change 1 and B200 (cc 10.0) for change 2; both report 148 SMs and share the code path. **Not measured on GB300.**

### DIFF
--- a/flashinfer/gemm/gemm_bf16_fp4_cute_dsl.py
+++ b/flashinfer/gemm/gemm_bf16_fp4_cute_dsl.py
@@ -639,6 +639,45 @@ def _sm100_bf16_fp4_tactic_configs() -> List[Tuple]:
 
 _SM100_BF16_FP4_TACTICS = tuple(_sm100_bf16_fp4_tactic_configs())
 
+
+def _sm100_bf16_fp4_tactic_ctas(tactic: Tuple, m: int, n: int) -> int:
+    """CTA count this tactic launches for a public ``(m, n)``.
+
+    The kernel's M mode carries public output channels ``n`` and its N mode
+    carries public rows ``m`` (see ``_sm100_bf16_fp4_tactic_configs``).  A
+    256-wide M tile is produced by a two-CTA cluster, so it costs two CTAs per
+    tile; every other shape is one.
+    """
+    (tile_n, tile_m, _), _, _ = tactic
+    ctas_per_tile = 2 if tile_n == 256 else 1
+    return -(-n // tile_n) * ctas_per_tile * -(-m // tile_m)
+
+
+def _select_sm100_bf16_fp4_default_tactic(
+    valid_tactics: List[Tuple], m: int, n: int, sm_count: int
+) -> Tuple:
+    """Pick the no-autotune fallback tactic for the SM100/103 dense W4A16 GEMM.
+
+    Ranks by ``waves x per-CTA bytes`` (``tile_n`` FP4 weights at half a byte,
+    ``tile_m`` BF16 activations at two): a wider row tile cuts waves but makes
+    each CTA dearer, so the best tile is an interior one that grows with ``m``.
+    A heuristic -- 1.07x off the per-shape optimum in geomean over 18 measured
+    shapes, worst 1.34x, against 2.8x/10.2x for the ``valid_tactics[0]`` rule
+    it replaces.  Ties go to ``raster_along_m=False``.
+    """
+
+    def rank(tactic: Tuple) -> Tuple:
+        tile_n, tile_m, _ = tactic[0]
+        waves = -(-_sm100_bf16_fp4_tactic_ctas(tactic, m, n) // sm_count)
+        return (
+            waves * (tile_n + 4 * tile_m),
+            bool(tactic[2]),
+            _SM100_BF16_FP4_TACTICS.index(tactic),
+        )
+
+    return min(valid_tactics, key=rank)
+
+
 _SM100_BF16_FP4_CUTE_DSL_TUNING_CONFIG = TuningConfig(
     dynamic_tensor_specs=(
         DynamicTensorSpec(
@@ -717,7 +756,7 @@ def _get_sm100_bf16_fp4_kernel(
         k,
         max_active_clusters=max_active_clusters,
         stream=stream,
-        options="--opt-level 3 --enable-tvm-ffi",
+        options="--opt-level 2 --enable-tvm-ffi",
     )
     _SM100_BF16_FP4_KERNEL_CACHE[cache_key] = compiled
     return compiled
@@ -863,10 +902,12 @@ def _cute_dsl_sm100_bf16_fp4_runner(enable_pdl: bool = True) -> TunableRunner:
             a, b, b_descale, alpha_for_launch, _, out, _ = inputs
             if tactic == -1:
                 valid = self.get_valid_tactics(inputs, None)
+                m, k = map(int, a.shape)
                 if valid:
-                    tactic = valid[0]
+                    tactic = _select_sm100_bf16_fp4_default_tactic(
+                        valid, m, int(b.shape[0]), get_device_sm_count(a.device)
+                    )
                 else:
-                    m, k = map(int, a.shape)
                     raise ValueError(
                         "no SM100 cute-dsl w4a16 tactic supports "
                         f"m={m}, n={int(b.shape[0])}, k={k}"


jimmyzho:
@flashinfer-bot run

flashinfer/gemm/gemm_bf16_fp4_cute_dsl.py
```diff
@@ -639,6 +639,45 @@ def _sm100_bf16_fp4_tactic_configs() -> List[Tuple]:
...
+    it replaces.  Ties go to ``raster_along_m=False``.
+    """
+
+    def rank(tactic: Tuple) -> Tuple:
+        tile_n, tile_m, _ = tactic[0]
+        waves = -(-_sm100_bf16_fp4_tactic_ctas(tactic, m, n) // sm_count)
+        return (
+            waves * (tile_n + 4 * tile_m),
+            bool(tactic[2]),
+            _SM100_BF16_FP4_TACTICS.index(tactic),
```
dhiraj113, line 675: **[non-blocking]** This adds ~17 us of host time to every no-autotune call. `rank()` runs `_SM100_BF16_FP4_TACTICS.index(tactic)` — an O(30) scan of nested tuples — inside a `min()` over the 30 tactics, so roughly 900 tuple comparisons per call; I measured 17.1–18.2 us regardless of shape, versus O(1) for the `valid[0]` it replaces. For an m=1 decode whose kernel is ~26 us on B200 that is a real addition to the launch path when the caller is not CUDA-graphed — and the untuned path is exactly the one such callers hit.

Two cheap fixes, either is enough: memoize the result per `(m, n, sm_count)` (`valid` is itself a pure function of shape, so an `lru_cache` on a small helper is safe), or precompute a `{tactic: index}` dict so the tie-break is O(1). `get_device_sm_count` is already cached, so that part is fine.

<details>
<summary>Measurement — B200, PR head 5af0b757</summary>

```bash
python - <<"PY"
import time
import flashinfer.gemm.gemm_bf16_fp4_cute_dsl as mod
T = list(mod._SM100_BF16_FP4_TACTICS)
for m, n in ((1, 4096), (128, 6656), (2048, 19968)):
    t0 = time.perf_counter(); N = 200
    for _ in range(N):
        mod._select_sm100_bf16_fp4_default_tactic(T, m, n, 148)
    print(f"m={m:4d} n={n:5d}: {(time.perf_counter()-t0)/N*1e6:6.1f} us/call")
PY
```

Observed:
```
m=   1 n= 4096:   17.1 us/call
m= 128 n= 6656:   17.7 us/call
m=2048 n=19968:   18.2 us/call
```
</details>

flashinfer/gemm/gemm_bf16_fp4_cute_dsl.py
```diff
@@ -863,10 +902,12 @@ def forward(
             a, b, b_descale, alpha_for_launch, _, out, _ = inputs
             if tactic == -1:
                 valid = self.get_valid_tactics(inputs, None)
+                m, k = map(int, a.shape)
                 if valid:
-                    tactic = valid[0]
+                    tactic = _select_sm100_bf16_fp4_default_tactic(
```
dhiraj113, line 907: **[non-blocking]** Nothing in `tests/` pins the behavior this PR exists to fix. The selector is a pure, deterministic function, but the "18/18 shapes verified by intercepting `_launch_cute_dsl_sm100`" check lives only in your ad-hoc script; the existing tests (`test_backend_matches_handwritten_dequant_matmul` with `auto_tuning=False`, `test_cute_dsl_every_tactic_matches_reference`) assert numerics only, which *any* valid tactic passes. #4686 regressed precisely because nothing asserted the opt-level or the untuned tactic.

A ~20-line unit test would close that class for good: expected picks for a few `(m, n, sm_count)` (e.g. `(1,4096)` -> `(128,8,256),(1,1),False`; `(128,4096)` -> `(128,32,...)`; `(2048,4096)` -> `(128,128,...)`), tile growing monotonically with `m`, ties resolving to `raster_along_m=False`, and an assertion that the dense path's compile options carry `--opt-level 2`.

```bash
grep -n '_select_sm100_bf16_fp4_default_tactic\|opt-level' tests/gemm/test_mm_bf16_fp4.py || echo "no test references the selector or the opt-level"
```

flashinfer/gemm/gemm_bf16_fp4_cute_dsl.py
```diff
@@ -639,6 +639,45 @@ def _sm100_bf16_fp4_tactic_configs() -> List[Tuple]:
...
+def _sm100_bf16_fp4_tactic_ctas(tactic: Tuple, m: int, n: int) -> int:
+    """CTA count this tactic launches for a public ``(m, n)``.
+
+    The kernel's M mode carries public output channels ``n`` and its N mode
+    carries public rows ``m`` (see ``_sm100_bf16_fp4_tactic_configs``).  A
+    256-wide M tile is produced by a two-CTA cluster, so it costs two CTAs per
+    tile; every other shape is one.
+    """
+    (tile_n, tile_m, _), _, _ = tactic
+    ctas_per_tile = 2 if tile_n == 256 else 1
```
dhiraj113, line 652: **[counterpoint to the CTA-count bot comment above — I think this line is correct as written.]** `use_2cta_instrs = mma_tiler_mnk[0] == 256` (lines 724 / 890) shows the 256-wide tile is the cooperative two-CTA tile, so it costs 2 CTAs per tile; a `(128, ..., (2, 1))` tactic is a cluster of two *independent* 128-tiles at one CTA each, so its per-tile cost is 1 and the total CTA count is unchanged. Multiplying by `cluster_shape_mn` would double-count those 10 tactics.

Empirically it also would not matter: re-ranking with cluster-based counting changes 0 of 18 selections on the description's shape grid (both rules pick cluster `(1,1)` everywhere). A one-sentence note in the docstring on why cluster shape is deliberately not used would pre-empt the question for the next reader.

<details>
<summary>Check — CPU only, PR head 5af0b757</summary>

```bash
python - <<"PY"
import flashinfer.gemm.gemm_bf16_fp4_cute_dsl as mod
T = mod._SM100_BF16_FP4_TACTICS
def ctas_cluster(t, m, n):
    (tn, tm, _), (cx, cy), _ = t
    return -(-n // tn) * (cx * cy) * -(-m // tm)
def pick(count, m, n, sm=148):
    return min(T, key=lambda t: (-(-count(t, m, n) // sm) * (t[0][0] + 4 * t[0][1]), bool(t[2]), T.index(t)))
shapes = [(m, n) for m in (1, 8, 32, 128, 512, 2048) for n in (4096, 6656, 19968)]
diff = sum(pick(mod._sm100_bf16_fp4_tactic_ctas, m, n) != pick(ctas_cluster, m, n) for m, n in shapes)
print(f"selections that change with cluster-based counting: {diff}/{len(shapes)}")
PY
```

Observed:
```
selections that change with cluster-based counting: 0/18
```
</details>

dhiraj113 (commented):
Reviewed with an independent before/after measurement on a B200 (148 SMs, CUPTI median kernel time, no autotune, `n=k=4096`), comparing `main` (O3 + `valid[0]`) against this head (`5af0b757`):

| m | before | after | speedup |
|---:|---:|---:|---:|
| 1 | 0.0283 ms | 0.0265 ms | 1.07x |
| 32 | 0.0290 | 0.0272 | 1.07x |
| 128 | 0.0914 | 0.0274 | 3.3x |
| 512 | 0.2988 | 0.0308 | 9.7x |
| 2048 | 1.1695 | 0.0886 | 13.2x |

No shape slower; the m<=32 gain is the O2 restore, the rest is the selector. These exceed the description's table because they include both changes (the description isolates change 2 at O2) — direction and magnitude are consistent. Clocks were unlocked, both runs back-to-back.

No correctness issues found — any valid tactic is numerically correct and the existing untuned-path and every-tactic tests cover that. Three non-blocking notes inline: the selector's per-call host cost (~17 us measured), the absence of a behavioral test pinning what this PR fixes, and a counterpoint to the earlier CTA-count bot comment (I believe the current `tile_n == 256` rule is right). One nit not worth a thread: the commit message says "Re reset o-level to 3" while the diff sets `--opt-level 2` — the PR title is correct, but the commit is what bisect shows.

The write-up (ncu root-cause, rejected alternatives, per-tactic ground truth, honest GB300 caveat) is the standard I'd point other perf PRs at. (AI-assisted review per `docs/code_review_guidance.md`.)

jimmyzho:
@flashinfer-bot run tests/gemm/test_mm_bf16_fp4.py

PR #4944, opened 2026-09-04
perf(sm120): optimize BF16 VSA attention

lishunyang12:
## Summary

Optimize the SM120 BF16 VSA CuTe-DSL path for MiniMax-H3 style block-sparse attention.

The patch keeps the existing SM120 VSA interface and focuses on the measured positive changes:

- skip unused V tile transfers;
- schedule sparse Q tiles across heads;
- skip warp-uniform VSA rescaling;
- elide the unused LSE epilogue;
- elide unchanged row-scale exponentials;
- alias Q/V shared storage when lifetimes do not overlap;
- retain SM120 VSA correctness coverage.

## PRO5000 microbenchmark

Hardware: NVIDIA RTX PRO 5000, CC 12.0, CUDA 13.0, PyTorch 2.13.0+cu130.

Workload: batch=1, seq_len=109632, heads=7, head_dim=128, logical_block=64, topk=64, logical_nnz=136464.

| Backend | p50 | Effective TFLOPS | Relative speed |
| --- | ---: | ---: | ---: |
| FastVideo Triton | 15.701 ms | 127.589 | 1.00x |
| FlashInfer SM120 VSA baseline | ~10 ms | — | 1.00x |
| Optimized FlashInfer SM120 VSA | 8.985 ms | 222.967 | 1.748x vs Triton |

Correctness for the reported run: max_abs=0.00097656, relative_l2=0.00302094.

NCU (optimized CuTe-DSL path): 8.97 ms kernel duration, 85.8% compute throughput, 54.6% memory throughput, zero local-memory spills, 246 registers/thread, 16.7% theoretical occupancy, 16.6% achieved occupancy.

## Tests

On the same PRO5000 environment:

```text
CUDA_VISIBLE_DEVICES=4 pytest -q tests/attention/test_vsa_block_sparse_sm120.py -k 'not perf_sweep'
18 passed, 1 deselected
```

The deterministic local precheck passes against `upstream/main`.

## Excluded experiments

Native CUDA load-pipeline, 3-CTA residency, register-paired softmax, and staggered CTA variants were measured but excluded because they were slower or unstable for this workload.

## Reproduction

The MiniMax-H3 benchmark command and shape configuration are preserved in the existing SM120 benchmark scripts. The result above uses the same input, warmup, repeat count, and GPU isolation for both backends.

Signed-off-by: lishunyang12 <lishunyang12@163.com>

### DIFF
--- a/flashinfer/cute_dsl/sparse/bsa_attn_sm120.py
+++ b/flashinfer/cute_dsl/sparse/bsa_attn_sm120.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import math
 from typing import Optional, Tuple
 
 import torch
@@ -112,6 +113,7 @@ def bsa_attn_sm120_blk64_fwd(
     block_sizes: Optional[torch.Tensor] = None,
     q2k_block_nums: Optional[torch.Tensor] = None,
     softmax_scale: Optional[float] = None,
+    skip_softmax_threshold_scale_factor: Optional[float] = None,
     return_lse: bool = False,
     out: Optional[torch.Tensor] = None,
     lse: Optional[torch.Tensor] = None,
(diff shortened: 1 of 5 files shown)


LeslieWylie:
The matching vLLM-Omni integration now has additional historical served E2E evidence on 8× RTX PRO 5000: client latency 43.142730 → 39.006554 s (-9.59%) for the same 15-second FastH3/Mystic0.2 workload, changing only the VSA compute provider. The deployed FlashInfer snapshot was `a2be065f4a93ffd6701c90be05aa547419dc2099`.

Configuration, source provenance, numerical/visual limitations, and the discarded compilation experiment are recorded in https://github.com/vllm-project/vllm-omni/pull/7034#issuecomment-5647006162 . This supports that deployed snapshot, not the current head of this PR; it is a single-case A/B rather than a broad speed guarantee.

PR #4945, opened 2026-09-04
fix(docs): deploy stable documentation from release tags

cindyzxq:
## 📌 Description

  The documentation site currently deploys on every push to `main`, while its displayed version comes from `version.txt`. As a result, the site can be labeled as a released version (currently `0.6.18`) while documenting APIs that only exist  on `main`.

  This PR separates documentation validation from production deployment:

  - Continue building documentation for pull requests and pushes to `main`.
  - Build and validate documentation for `v*` tag pushes without deploying it.
  - Deploy production documentation only:
    - after the release workflow completes successfully; or
    - through an explicit manual dispatch for an existing release tag.
  - Validate that the requested release tag:
    - has a supported version format;
    - matches `version.txt`; and
    - resolves to the checked-out commit.
  - Split documentation build and deployment into separate jobs.
  - Scope the `pages` concurrency group to the deployment job so ordinary `main`/PR builds cannot replace a pending release deployment.
  - Restrict GitHub Pages write and OIDC permissions to the deployment path.
  - Make the documentation workflow reusable from `release.yml`.
  - Run documentation deployment only after the GitHub Release, PyPI publication,and wheel-index update have all succeeded.

  After this PR is merged, the current site can be repaired by manually running the `Build FlashInfer Docs` workflow from `main` with:
  ```text
  release_tag: v0.6.18
```
  This rebuilds the site from the immutable v0.6.18 tag rather than from main.

## 🔍 Related Issues
https://github.com/flashinfer-ai/flashinfer/issues/4934

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests
- [x] Tests have been added or updated as needed. No unit tests are applicable
    because this is a GitHub Actions workflow-only change.

- [x] Static validation completed:
      - both modified YAML files parse successfully;
      - the release dependency chain and deploy-job isolation were checked;
      - the release-tag validation script passes bash -n;
      - git diff --check passes;
      - the existing v0.6.18 tag matches its version.txt and checked-out commit.

## Reviewer Notes
Please focus on:

  1. Whether documentation should wait for all of create-release, publish-to-pypi, and update-wheel-index before deployment.
  2. The reusable-workflow permission chain for pages: write and id-token: write.
  3. The intentional behavior that a v* tag push builds and validates documentation but does not deploy it.
  4. The manual workflow_dispatch path used to restore the current site from v0.6.18.

## Summary by CodeRabbit

* **New Features**
  * Documentation builds now run automatically for pushes, pull requests, and version tags, or can be triggered manually and reused by other workflows.
  * Release workflows can deploy documentation for published releases.
  * Documentation deployment uses dedicated permissions and serialized runs to prevent overlapping deployments.

* **Bug Fixes**
  * Release-tag deployments validate the tag format, project version, and checked-out commit.
  * Prevented deployments from using a branch with the same name as a release tag.
  * Limited documentation artifacts and deployment to release-tag runs.

### CODE REVIEW

File: .github/workflows/build-doc.yml [hunk 1]
```diff
@@ -4,33 +4,69 @@ on:
   push:
     branches:
       - main
+    tags:
+      - 'v*'
```
aleozlx: can or should this be more specific? .. `very_good_counterexample`

aleozlx (approved):

.github/workflows/build-doc.yml
cindyzxq: Good point — updated. The workflow now explicitly matches stable, rc, and post-release version tags instead of the broad `v*`. The in-workflow tag/version/commit validation remains as a second guard.

aleozlx:
@flashinfer-bot run

aleozlx:
what copilot says about the jit-cache errors 

> The self-hosted JIT-cache wheel build is compiling CUDA extensions with a host compiler configuration that is not using C++20, but the installed PyTorch headers require C++20 or newer.

aleozlx:
the unrelated issue seem to be https://github.com/flashinfer-ai/flashinfer/issues/4995?

PR #4917, opened 2026-09-03
perf(gdn): optimize SM100 CP kernels

guangyunh-nv:
## Description

This PR ports the useful SM100 CP kernel optimizations identified while comparing the CuTeDSL implementation with the CAKE-generated kernels in #4539:

- vectorize state loads in the final prefill kernel;
- interleave fixup accumulators to expose instruction-level parallelism;
- reuse gate factors and express their coordinates directly in MN semantics;
- replace runtime integer division in CP work mapping with fast divmod.

The launch-grid heuristic and optional max_seqlen interface change were split into independent PR #4946 so they do not block these kernel-only improvements.

## Tests

- Pre-commit hooks passed for the optimization stack.
- Before the history split, the combined stack passed the GDN prefill selection: 3087 passed, 864 skipped. The split only removes the independent host launch/API commit.
- CUDA-graph capture and replay passed.

## B200 CUDA-graph benchmark

Built-in benchmarks/bench_gdn_prefill.py with CUDA graphs, 10 warmups, and 100 measured iterations. GPU: NVIDIA B200 at a 1000 W power limit.

Only single-sequence cases are retained here because they are independent of the launch-grid change moved to #4946. For one sequence, max_seqlen equals total_seq_len under both launch policies.

```
Heads            Seqlens          h_qk  h_v   Opt graph  TFLOPS  Main graph  Speedup
-----------------------------------------------------------------------------------
397B/122B TP8    1x65536             2    8      0.330ms   104.0      0.353ms    1.07x
397B/122B TP8    1x32768             2    8      0.195ms    88.3      0.218ms    1.12x
397B/122B TP8    1x16384             2    8      0.127ms    67.5      0.150ms    1.18x
397B/122B TP8    1x8192              2    8      0.093ms    46.4      0.114ms    1.23x
397B/122B TP8    1x4096              2    8      0.070ms    30.7      0.087ms    1.24x
397B/122B TP8    1x2048              2    8      0.057ms    19.0      0.072ms    1.26x

397B/122B TP4    1x65536             4   16      0.620ms   110.8      0.635ms    1.02x
397B/122B TP4    1x32768             4   16      0.325ms   105.6      0.337ms    1.04x
397B/122B TP4    1x16384             4   16      0.188ms    91.6      0.201ms    1.07x
397B/122B TP4    1x8192              4   16      0.120ms    71.7      0.133ms    1.11x
397B/122B TP4    1x4096              4   16      0.085ms    50.4      0.097ms    1.14x
[...]

PR #4946, opened 2026-09-04
perf(gdn): bound CP launches by longest sequence

guangyunh-nv:
## Description

CP launch grids currently use the packed token count as the per-sequence bound. For batched inputs this can launch substantially more T, MN, and final-prefill CTAs than any individual sequence requires.

This PR adds an optional max_seqlen hint to chunk_gated_delta_rule and uses it to bound CP launch grids without reading cu_seqlens back to the host.

When max_seqlen is omitted, the CP path falls back to ceil(total_seq_len / num_seqs). That fallback assumes a balanced batch and can under-launch an imbalanced batch, so callers allowing unequal sequence lengths must provide the exact maximum whenever use_cp=True or use_cp=auto may select CP.

The in-repo GDN prefill benchmark now supplies the exact maximum from its host-side sequence specification, and CP tests pass the known maximum for variable-length cases.

## Related PR

Split from #4917 so the SM100 kernel optimizations can be reviewed and merged independently of this public interface change.

## Tests

- Focused public CP wrapper coverage: 6 passed.
- The identical patch passed the complete GDN prefill test selection before the history split: 3087 passed, 864 skipped.
- CUDA-graph capture and replay were previously verified with the explicit hint.

## Pull Request Checklist

- [x] Focused tests pass.

## Summary by CodeRabbit

* **New Features**
  * Added an optional maximum-sequence-length hint for gated delta-rule prefill operations.
  * Improved context-parallel processing for batches containing sequences of different lengths.
  * When no hint is provided, processing uses a safe launch bound while optimizing based on balanced sequence-length estimates.

* **Bug Fixes**
  * Improved sequence-length handling for more accurate processing across variable-length and empty-sequence batches.


jiahanc (approved):
lgtm

flashinfer/gdn_prefill.py
```diff
@@ -923,7 +924,23 @@ def chunk_gated_delta_rule(
                 if _arch_major in (9, 10, 12)
                 else {}
             )
-            cp_max_seqlen = get_cp_max_seqlen(total_seq_len, num_seqs, max_seqlen)
+            cp_tuning_max_seqlen = get_cp_max_seqlen(
+                total_seq_len, num_seqs, max_seqlen
+            )
+            cp_launch_max_seqlen = (
+                max_seqlen if max_seqlen is not None else total_seq_len
+            )
```
guangyunh-nv: Make `get_cp_max_seqlen` return a tuple, then document the reason:

1. tuning for maximizing parallelism
2. launch to avoiding under-launch.

PR #4952, opened 2026-09-04
feat(moe): split QuantConfig into weight/activation/output format axes

feih-nv:
### Why

`QuantVariant` mixed three kinds of names (formats like `NVFP4`, WxAy labels like `W4A16`, kernel names like `Humming`), so it could not express new MMA pairs:

- `W4A16` already means two weight encodings: MXFP4 on TRTLLM / CUTLASS SM90, NVFP4 on CuTe-DSL / b12x. With several candidates, the autotuner's pick decided the weight format.
- `MXFP4` means MXFP4×MXFP8 (W4A8), and MXFP4×MXFP4 was inexpressible.
- `MxInt4` is MXINT4 weights × BF16 activations, a third W4A16-style encoding the name did not show.

### Design

```python
class QuantFormat(Enum):
    BF16, FP16, FP8PerTensor, DeepSeekFp8, MXFP8, NVFP4, MXFP4, MXINT4, INT4

@dataclass(frozen=True)
class QuantConfig:
    weight: QuantFormat                          # MMA weight operand
    activation: QuantFormat                      # MMA activation operand
    output: QuantFormat = QuantFormat.BF16       # layer output
    variant: Optional[QuantVariant]              # derived from the pair (init=False); constructor accepts variant= as a deprecated preset
    swizzled_scale_factors / per_token_scale     # unchanged

    # an omitted axis is BF16 (unquantized); knobs after the three axes are keyword-only
```

Axes are the MMA numeric formats, not the dtype crossing the Python API: CUTLASS NVFP4 takes BF16 activations and quantizes in-kernel, and is still `(NVFP4, NVFP4)`.

Runners declare `supported_quant_variants` and `supported_output_formats` separately because output does not vary by pair today (every runner is BF16-only; `FP16` is in the enum for later). `MoELayer` filters candidates on both before build.

### API

- New `QuantFormat` enum, exported from `flashinfer.fused_moe`. `FP16` and `INT4` are new names; `MXFP8` / `MXINT4` replace the `MxFp8` / `MxInt4` spellings.
- `QuantConfig(weight, activation, output)`: an omitted axis is BF16, i.e. unquantized (`QuantConfig()` is BF16×BF16; `QuantConfig(weight=MXFP4)` is MXFP4 weights with BF16 activations, W4A16; MXFP4×MXFP8 needs both axes). The three axes are positional, everything after them keyword-only. `QuantConfig` is structural only; legal combinations are runner capabilities.
- `QuantVariant` stays as a deprecated preset: `QuantConfig(variant=NVFP4)` expands to the pair and emits a `DeprecationWarning`. The stored `variant` is derived from the pair and excluded from `__init__` fields, so `dataclasses.replace` re-derives it instead of replaying it.
- `QuantVariant.W4A16` is rejected as `variant=`; spell the pair (`QuantConfig(MXFP4, BF16)` / `QuantConfig(NVFP4, BF16)`) or use `QuantConfig.from_variant(W4A16, w4a16_weight=...)`, which also warns.

### Runners and layer

- `supported_quant_variants` now holds `(weight, activation)` pairs (a variant is one such combination) plus `supported_output_formats`; `supported_activation_classes_by_quant` is keyed by pair.
- Pairs without a `QuantVariant` mapping are rejected by `check_support()` for now: runner build paths and weight preparation still dispatch on `QuantVariant`, so such a pair would pass capability checks and fail later. Lifted when dispatch keys on `quant.pair` (config-merge follow-up).
- `_cache_key_extras` uses the three format names instead of `variant.name`; persisted autotune cache keys invalidate once.

### Consumers

- moe_ep split kernel: `build_activation_pack` and siblings take `quant: QuantConfig` instead of `quant_variant: QuantVariant`; `mxfp8_dispatch` validates `(MXFP4, MXFP8)`; weight materialization dispatches on the pair.
- `benchmarks/bench_cute_dsl_moe_distributed.py` builds the pair explicitly.
- `scripts/generate_moe_activation_matrix.py` labels rows `weight×activation` (matrix regenerated). Design doc §2 / §3 updated and a "QuantConfig three-axis formats" section added. `docs/api/fused_moe.rst` gains a "Unified MoE API" autosummary (`MoELayer`, config types incl. `QuantFormat`, packs), which were not listed before.

### ✅ Pre-commit Checks

## 🧪 Tests

New CPU tests:

- `QuantConfig` construction: both-or-neither axes, `variant=` expansion and conflicts, `dataclasses.replace`, repr round-trip.
- Runner and `MoELayer` gating on `(weight, activation)` pairs and `output`, including the error text.
- Pair-keyed `supported_activation_classes_by_quant` and the activation-matrix generator.

Updated: unified-MoE, EP bridge, and activation-matrix tests construct configs from pairs. No GPU tests added; kernel paths are unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added explicit weight, activation, and output quantization formats to the unified MoE API.
  * Added backend support checks for format pairs and output formats.
  * Exposed `QuantFormat` through the public fused MoE API.
  * Defaulted unspecified quantization axes to BF16.
  * Retained legacy presets with deprecation warnings and conversion support.

* **Documentation**
  * Updated API and design documentation with the new configuration model and examples.

* **Bug Fixes**
  * Improved unsupported-configuration errors to identify configured quantization formats.


feih-nv:
@flashinfer-bot run

docs/design_docs/flashinfer_moe_api.md
```diff
@@ -1020,6 +1031,23 @@ kernel); (5) tighten the quantized-numeric net via the QuantSpec scale policy
...
+granularity into one enum. `W4A16` then grew two weight encodings (MXFP4 on
+TRTLLM/CUTLASS SM90, NVFP4 on CuTe-DSL/b12x), and `MXFP4` already meant
+MXFP4×MXFP8 rather than MXFP4×MXFP4.
+
+`QuantConfig` now has three `QuantFormat` axes: `weight`, `activation`, and
+`output`. `QuantConfig` itself is structural only (members must be
+`QuantFormat`; no `torch.dtype` coercion). Legal combinations are runner
+capabilities: `supported_quant_variants` plus `supported_output_formats`.
+`QuantVariant` remains a deprecated preset that expands every member except
+`W4A16`, which must be spelled as `weight=QuantFormat.MXFP4` or
```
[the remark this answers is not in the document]
feih-nv: @coderabbitai we have explicitly said `with activation=QuantFormat.BF16` in next line, so this should not be a concern.

flashinfer/moe_ep/backends/split/kernel/fused_moe/weights.py
```diff
@@ -100,7 +108,9 @@ def materialize_fused_moe_weights(
             )
             return pack
 
-        if variant == QuantVariant.NVFP4 and isinstance(backend_cfg, TrtllmFp4Config):
+        if pair == (QuantFormat.NVFP4, QuantFormat.NVFP4) and isinstance(
+            backend_cfg, TrtllmFp4Config
+        ):
```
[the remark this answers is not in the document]
feih-nv: @coderabbitai this matches tot, expanding that support is not part of this pr

bkryu (approved):

flashinfer/fused_moe/api.py
```diff
@@ -60,13 +62,42 @@
...
+
+    Axes describe the format consumed by the MMA, not the dtype of the tensor
+    that crosses the Python API. A CUTLASS NVFP4 runner that takes BF16
+    activations and quantizes in-kernel is still ``(NVFP4, NVFP4)``.
+    """
+
+    BF16 = 0
+    FP16 = 1
+    FP8PerTensor = 2
+    DeepSeekFp8 = 3  # e4m3 + fp32 block scales (128x128 weight, 1x128 activation)
```
PetersonGuo, line 82: Rather than specifying DeepSeekFp8 tensor it might make sense to create a new axis with `vec_size` similar to what is done in our CuteDSL kernels. This will help when we have vec sizes of 16 for MXFPx format and vec size of 32 for NVFP4 format.

flashinfer/fused_moe/api.py
```diff
@@ -148,9 +234,137 @@ class QuantConfig:
...
+                raise ValueError(
+                    f"QuantConfig(variant={self.variant!r}) expands to "
+                    f"weight={pair[0]!r}, activation={pair[1]!r}, which conflicts "
+                    f"with the explicit weight={self.weight!r}, "
+                    f"activation={self.activation!r}."
+                )
+            object.__setattr__(self, "weight", pair[0])
+            object.__setattr__(self, "activation", pair[1])
+            return
+        if self.weight is None and self.activation is None:
```
PetersonGuo: I think this should be moved into default args in the `__init__`

flashinfer/fused_moe/runners.py
```diff
@@ -71,6 +71,8 @@
     MoEActivationPack,
     MoEConfig,
     MoEWeightPack,
+    QuantConfig,
+    QuantFormat,
     QuantVariant,
```
PetersonGuo, line 76: Is this being kept for backwards compatibility? I feel like some of the uses within this file can start to be migrated/may be silently wrong and doesn't properly error catch the new config. Didn't look too much into this, please verify first

feih-nv: Yes, transitional. `QuantVariant` is just deprecated in this PR,  because `runners.py` still has ~24 dispatch sites on `config.quant.variant` (CuteDslRunner._build, the TRTLLM FP4 SM107 / per-token gates, MxFp8 shuffled-weight selection, ...), which is why `QuantConfig` keeps a derived `variant` field.

Migrating those sites to `QuantFormat` and deleting `QuantVariant` is a follow-up task, kept out of this PR to keep the diff reviewable.

flashinfer/fused_moe/api.py
feih-nv: Agree the kernels need the scale vector size, but I'd expose it as a property of the format rather than a config axis. `vec_size` is determined by the format `(NVFP4 = 16 / e4m3 + global scale, MXFP4/MXFP8 = 32 / ue8m0, MXINT4 = 32 / bf16, DeepSeekFp8 = 128 / fp32 with 128×128 weights and 1×128 activations)`, so a separate axis only adds combinations we'd have to reject `(NVFP4 + 32)`, and a single `int` can't carry scale `dtype` or the 2-D DeepSeek block. That's the `QuantDtype + QuantGranularity` shape the MVP collapsed for the same reason.

I'd do that as a follow-up rather than in this PR: it's a new public surface and the mapping deserves its own review.

Follow-up I'd take: read-only properties on `QuantFormat (scale_vec_size, scale_dtype, element_dtype)` so runners read `quant.weight.scale_vec_size` / `quant.activation.scale_vec_size` instead of keeping their own tables. Missing scales are `None` (BF16 activations in W4A16). `QuantConfig` stays three axes.

PetersonGuo (approved):

flashinfer/fused_moe/api.py
PetersonGuo: How would you plan to make this work with dtypes with multiple `scale_vec_size` if it's read only? If a user wanted nvfp4 at a 32 block granularity they would need to select between the two variants. Though this likely won't be an issue in the near future so it would be good to keep in mind for if there needs to be a follow up.

feih-nv: Good question. A different scale granularity would need a distinct `QuantFormat` value: runners
declare support per (weight, activation) pair, and `NVFP4@16` and `NVFP4@32` are not kernel- or
checkpoint-compatible, so MoELayer has to tell them apart — like NVFP4 vs MXFP4 today. The enum
lists formats that exist rather than the cross product of dtype × vec_size, most of which have
no kernel or checkpoint.

Whether that value is a new member (e.g. `NVFP4_K32`) or `QuantFormat` members carrying a small
scale spec `(element dtype, scale dtype, block size, global scale)` is a spelling question we can
settle when the format actually shows up. Either way `QuantConfig` keeps its three axes (weight,
activation, output) and the read-only properties still hold.

Does it make sense?

flashinfer/fused_moe/api.py
```diff
@@ -128,14 +195,33 @@ def __repr__(self) -> str:
         return f"RoutingConfig({', '.join(parts)})"
 
 
-@dataclass(frozen=True)
+@dataclass(frozen=True, init=False)
```
aleozlx, line 198: could this init=False cause surprises? not super familiar. shall we double check and summarize findings?

flashinfer/fused_moe/api.py
```diff
@@ -148,9 +234,137 @@ class QuantConfig:
...
+    variant: Optional[QuantVariant] = field(init=False)
+    swizzled_scale_factors: Optional[bool]
+    per_token_scale: Optional[bool]
+
+    def __init__(
+        self,
+        weight: Optional[QuantFormat] = None,
+        activation: Optional[QuantFormat] = None,
+        output: QuantFormat = QuantFormat.BF16,
+        variant: Optional[QuantVariant] = None,
```
aleozlx, line 252: should we leave a keyword-only marker so the positional args don't grow out of control?

feih-nv: This is intentional: `init=False` here only suppresses the generated `__init__` so we can hand-write one.
The hand-written `__init__` takes `weight/activation` as `Optional` (None = "not given") so it can
enforce "both or neither" (`QuantConfig()` is `BF16×BF16,` `QuantConfig(weight=MXFP4)` alone raises),
while the stored fields stay `QuantFormat`.

Think we can simplify `QuantConfig` to a plain dataclass like this where an omitted axis is by default BF16:
```
    @dataclass(frozen=True)
    class QuantConfig:
        weight: QuantFormat = QuantFormat.BF16
        activation: QuantFormat = QuantFormat.BF16
        output: QuantFormat = QuantFormat.BF16
        _: KW_ONLY
        swizzled_scale_factors: Optional[bool] = None
        per_token_scale: Optional[bool] = None
```

flashinfer/fused_moe/api.py
feih-nv: This is intentional: `init=False` here only suppresses the generated `__init__` so we can hand-write one.
The hand-written `__init__` takes `weight/activation` as `Optional` (None = "not given") so it can
enforce "both or neither" (`QuantConfig()` is `BF16×BF16,` `QuantConfig(weight=MXFP4)` alone raises),
while the stored fields stay `QuantFormat`.

Think we can simplify `QuantConfig` to a plain dataclass like this where an omitted axis is by default BF16:
```
    @dataclass(frozen=True)
    class QuantConfig:
        weight: QuantFormat = QuantFormat.BF16
        activation: QuantFormat = QuantFormat.BF16
        output: QuantFormat = QuantFormat.BF16
        _: KW_ONLY
        swizzled_scale_factors: Optional[bool] = None
        per_token_scale: Optional[bool] = None
```

aleozlx (approved):

flashinfer/fused_moe/api.py
feih-nv: Added in b2411b3a: `*` after `output`, so the three axes stay positional
(QuantConfig(MXFP4, MXFP8, BF16)) and variant / swizzled_scale_factors / per_token_scale
must be named. Same commit also drops the both-or-neither rule: an omitted axis is BF16
(unquantized), so QuantConfig(weight=MXFP4) is MXFP4 weights with BF16 activations.

flashinfer/fused_moe/api.py
```diff
@@ -148,9 +234,129 @@ class QuantConfig:
...
+                )
+        if self.variant is not None:
+            try:
+                pair = QUANT_VARIANT_TO_PAIR[self.variant]
+            except KeyError as exc:
+                raise ValueError(
+                    f"QuantConfig cannot expand {self.variant!r}."
+                ) from exc
+            given = (self.weight, self.activation)
+            if given != (QuantFormat.BF16, QuantFormat.BF16) and given != pair:
```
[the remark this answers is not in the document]
feih-nv: @coderabbitai Known and accepted. An explicit BF16×BF16 alongside `variant=` is indistinguishable from the default since b2411b3a made omitted axes default to BF16 (replacing the sentinel-based both-or-neither rule at reviewers' request). `variant=` is a deprecated preset that already emits a `DeprecationWarning` and is removed entirely in the follow-up PR, so I'd rather not reintroduce a sentinel for it; any non-BF16 explicit pair that disagrees with the variant still raises (tested). Docstring now states the BF16×BF16 exception (8fcc9ec6a).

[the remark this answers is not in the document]
feih-nv: @coderabbitai Known and accepted. An explicit BF16×BF16 alongside `variant=` is indistinguishable from the
default since b2411b3a made omitted axes default to BF16 (replacing the sentinel-based
both-or-neither rule at reviewers' request). variant= is a deprecated preset that already
emits a DeprecationWarning and is removed entirely in the follow-up PR, so I'd rather not
reintroduce a sentinel for it; any non-BF16 explicit pair that disagrees with the variant
still raises (tested). Clarifying the docstring to state the BF16×BF16 exception.

flashinfer/fused_moe/api.py
feih-nv: Changed the default to bf16. The class-level `init=False` is kept to not explode this PR -- it will be removed in the follow up PR when removing the `QuantVariant`.

feih-nv:
@flashinfer-bot run

docs/design_docs/flashinfer_moe_api.md
```diff
@@ -1020,6 +1031,25 @@ kernel); (5) tighten the quantized-numeric net via the QuantSpec scale policy
...
+`weight=QuantFormat.NVFP4` with `activation=QuantFormat.BF16`, or expanded via
+`QuantConfig.from_variant(W4A16, w4a16_weight=...)`. An omitted axis is BF16
+(unquantized): `QuantConfig()` is BF16×BF16 and `QuantConfig(weight=MXFP4)` is
```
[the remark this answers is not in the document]
feih-nv: Fixed in 4a1e9b2b: `QuantConfig.from_variant(QuantVariant.W4A16, ...)` and `QuantConfig(weight=QuantFormat.MXFP4)`.

feih-nv:
@flashinfer-bot run

PR #4959, opened 2026-09-04
[DO NOT MERGE] ci verification B: malformed rejection, then targeted scope

aleozlx:
## 📌 Description

**Throwaway PR — do not merge. Delete once CI verification is done.**

Filed as a real experimental-track PR to exercise the handler merged in #4880, now that it is
live on `main`. `issue_comment` workflows load from the default branch, so nothing on #4880
itself could ever execute that code; this is the first opportunity to run it.

Adds a mock `@flashinfer_experimental_api` under `flashinfer/experimental/` plus three
CPU-only tests under `tests/experimental/`, so a declared scope has something real to target.

**What this PR is verifying:** the **rejection** path (a malformed scope must not start CI) and then the **targeted** path (a valid scope must produce `Targeted Unittest` lanes carrying it).

## 🔍 Related Issues

Verification of #4880. No tracking issue — this PR is disposable and will be closed.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

## 🧪 Tests

- [x] Tests have been added or updated as needed.
- [x] All tests are passing (`unittest`, etc.).

## 🔬 Experimental Track

- [x] This PR is **experimental**: it adds or changes code under `flashinfer/experimental/` and/or an `@flashinfer_experimental_api`. Tracking issue: n/a — throwaway verification PR
  - [ ] The tracking issue names an owner, the reason for the experimental path, and a graduation plan with a target release. — n/a, disposable
  - [x] Core changes are limited to a thin entry point (signature, shared validation, feature-gate check, backend selection, handoff).
  - [x] Tests live in `tests/experimental/` and were validated on the intended hardware; a runnable example is included.
  - [x] Nothing is registered in `flashinfer/aot.py`, and no experimental backend is reachable from `backend="auto"` without `FLASHINFER_ALLOW_EXPERIMENTAL_AUTO_BACKENDS=1`.
  - [x] **Test scope declared below.**

```experimental-tests
tests/experimental/test_mock_verify_b.py
```

## Reviewer Notes

Nothing to review — this exists to be driven by bot commands and then closed.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

## Summary by CodeRabbit

* **New Features**
  * Added an experimental utility that scales a value by a configurable factor.
[...]

PR #4962, opened 2026-09-04
fix(ci): publish the test scope with GITHUB_TOKEN, not the bot PAT

aleozlx:
## 📌 Description

A valid scoped command fails in production:

```
gh: Resource not accessible by personal access token (HTTP 403)
##[error]Process completed with exit code 1
```

The `Handle` step runs with `GH_TOKEN` set to the bot PAT. The `permissions:` block at the top
of the file grants `statuses: write` to **`GITHUB_TOKEN`** — it does not apply to a PAT — and the
bot PAT carries no commit-status permission. So `POST /repos/.../statuses/{sha}`, which is how a
scope reaches the test job, is refused.

**Effect: the targeted-testing path does not work at all.** A valid scoped command fails the
handler, applies no label, and starts no CI. Bare and malformed commands are unaffected, because
neither touches the statuses API.

**Fix:** publish the scope with `${{ github.token }}`. The PAT is still required for the label,
because a `labeled` event raised by `GITHUB_TOKEN` does not trigger workflows. A status needs to
trigger nothing, and `pr-test.yml` already reads it with `GITHUB_TOKEN`. Splitting by purpose also
keeps the privileged token's blast radius smaller. Alternative would be granting the PAT
`repo:status`, which needs token administration and widens it instead.

## 🔍 Related Issues

Fixes the scoped path introduced in #4880. Found by driving the merged handler on throwaway PR
#4959.

### ✅ Pre-commit Checks

## 🧪 Tests

**This file cannot be tested by CI**, on this PR or any other: `issue_comment` workflows load from
the default branch, so it takes effect only once merged. That is also why the bug reached `main`.

Verification therefore has to be post-merge: after this lands, re-issue a scoped command on #4959
and confirm the scope publishes and `Targeted Unittest` lanes appear.

**Why the pre-merge sandbox missed it.** The scoped path *was* exercised end to end before merge,
in a scratch repository running byte-identical logic — publish, chunk, reassemble, matrix, all the
way into the runner's parsed argv. But that mirror ran the step under `${{ github.token }}`, while
production runs it under the PAT. Faithful in behaviour, wrong in identity — and authorization is a
property of identity, so no amount of logic fidelity could have caught it. The check that would
have: diffing the two workflows' `GH_TOKEN:` lines against their `permissions:` blocks.

## Reviewer Notes

Three lines of substance: one added `env:` entry and two `GH_TOKEN="$STATUS_TOKEN"` prefixes. No
control flow changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed commit status publishing for the “run” bot command by using the appropriate authorization token.
  * Preserved workflow-triggering capabilities while ensuring status updates are successfully posted.

### DIFF
--- a/.github/workflows/ci-bot-commands.yml
+++ b/.github/workflows/ci-bot-commands.yml
@@ -116,6 +116,16 @@ jobs:
         if: steps.check-permission.outputs.authorized == 'true' && steps.parse.outputs.command == 'run'
         env:
           GH_TOKEN: ${{ secrets.FLASHINFER_BOT_TOKEN }}
+          # Commit statuses are published with GITHUB_TOKEN, not the bot PAT. The
+          # `permissions:` block at the top of this file grants `statuses: write` to
+          # GITHUB_TOKEN only -- it does not apply to a PAT -- and the bot PAT has no
+          # commit-status permission, so publishing a scope with it fails:
+          #     gh: Resource not accessible by personal access token (HTTP 403)
+          # The PAT is still required for the label, because a `labeled` event raised by
+          # GITHUB_TOKEN does not trigger workflows. A status needs to trigger nothing;
+          # pr-test.yml reads it with GITHUB_TOKEN too. So splitting by purpose is safe
+          # and keeps the privileged token's blast radius smaller.
+          STATUS_TOKEN: ${{ github.token }}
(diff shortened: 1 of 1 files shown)


aleozlx:
@claude review this

PR #4974, opened 2026-09-04
Ameyn/gdn stp replayssm kernel

ameynaik-hub:
## 📌 Description

feat(gdn): STP (T=1) ucache decode kernel
Adds flashinfer/gdn_kernels/gdn_decode_bf16_wy_ucache_stp.py: a
self-contained GDN (gated delta rule) decode kernel for single-token
prediction with a u-cache. Instead of rewriting the [HV, V, K] SSM state
on every decoded token, each step appends the token's update ingredients
(L2-normalized k, delta-rule correction u, log-decay g) to a small
16-slot per-request cache and reads the state WITHOUT writing it; the
state is rewritten only on a periodic flush. This removes the per-step
state write from steady-state decode while keeping outputs equivalent to
the standard per-token state update.

Entry point: gated_delta_rule_stp_ucache_flush. SM90+; K == V == 128.

Inputs (IO dtype = bf16 default, fp16 via GDN_UCACHE_IO_DTYPE; STATE and
RING dtypes default to IO, fp16 selectable via GDN_UCACHE_STATE_DTYPE /
GDN_UCACHE_RING_DTYPE; all GEMM accumulation and gate math in fp32):
```
  q, k                  [B, 1, H, K]       IO
  v                     [B, 1, HV, V]      IO
  a, b                  [B, 1, HV]         IO
  A_log, dt_bias        [HV]               cast + cached to IO
  initial_state_source  [pool, HV, V, K]   STATE  (read-only except flush)
  initial_state_indices [B]                int32
  k_cache               [pool, H,  16, K]  RING
  u_cache               [pool, HV, 16, V]  RING
  g_cache               [pool, HV, 16]     fp32
  hist_len              [B]                int32  (cache fill; committed
                                                   in place every call)
  cache_base            [B]                int32  (must be zeros)
  scale                 float
```
q/k/v/a/b may alternatively be passed prepadded as persistent [B, 4, ...]
buffers (row 0 real, rows 1..3 zero) -- a zero-copy serving fast path
that removes all per-step staging copies.

Outputs: the decoded token's output [B, 1, HV, V] in the IO dtype. Each
verify step also appends exactly one (k, u, g) entry at cache slot
hist_len and commits hist_len += 1 (single fused kernel, CUDA-graph
safe).

Flush: when a request arrives with hist_len == flush_min (default 15,
legal range [1, 15]), the kernel folds the cached window AND the current
token into the state pool -- S <- e^G * S + sum_j w_j u_j k_j^T, one
fp32-accumulated rewrite with a single rounding to the STATE dtype --
and the cache restarts empty (hist_len <- 0). Steady state: 15 read-only
steps + 1 flush per 16 decoded tokens.

Correctness -- 27 STP tests added to the existing ucache suite (its fp32
oracle, dtype arms, and case builder are generalized to T=1 and reused;
no new test file). Run:

  pytest tests/gdn/test_decode_ucache.py -k stp

Coverage: all 5 dtype arms, exhaustive cache depths P in [0, 15],
shuffled non-identity state indices, single-append audits, multistep
fold cycles, a 200-step drift run against a never-resynced fp32
reference (max output error 6.5e-4, flat across 12 flush cycles), and
the prepadded path bitwise-matching the default path.

Benchmark -- extends the existing harness with --kernel stp (default
--kernel mtp behavior unchanged). Run:

`  python benchmarks/bench_gdn_ucache_flush.py --kernel stp --rates 0 100`

B200, bf16, CUDA-graph replay, CUPTI cold-L2 timing, median of 300:
```
     B | verify step (us) | flush step (us)
     8 |            11.8  |           17.8
    16 |            15.4  |           24.7
    32 |            25.3  |           42.4
    64 |            43.2  |           76.1
   128 |            74.4  |          143.1
   256 |           134.9  |          274.8
```   
## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## 🔬 Experimental Track

```experimental-tests
#
#
```

## Summary by CodeRabbit

* **New Features**
  * Added benchmarking support for both multi-token and single-token GDN ucache processing.
  * Added configurable token counts and ring-buffer depths for benchmark scenarios.
  * Added command-line selection between single-token and multi-token benchmark modes.

* **Bug Fixes**
  * Improved validation of state updates, history replay, ring-window handling, data layouts, and precision across supported configurations.
  * Added coverage for single-token appends, fold/absorb operations, long-running stability, and state-index variations.

* **Tests**
  * Added comprehensive correctness tests against a reference implementation, including compatibility checks with the vLLM Triton implementation.


kahyunnam (commented):
Some high level comments: 

- adding a PR description would be helpful for review 
- gdn_decode_bf16_wy_ucache_stp.py is ~90% byte-identical to gdn_decode_bf16_wy_ucache_flush.py. For future maintenance, should we consider consolidating these? 
- conventions/organization nit: this new op is not reachable from flashinfer.gdn_kernels, unlike every sibling in the family: gated_delta_rule_stp_ucache_flush is in the module's __all__ but flashinfer/gdn_kernels/__init__.py never imports it, so from flashinfer.gdn_kernels import gated_delta_rule_stp_ucache_flush fails

kahyunnam:
@flashinfer-bot run

kahyunnam:
@flashinfer-bot run tests/gdn

PR #4719, opened 2026-08-25
feat(mla): support variable-Q decode with DCP in CuTeDSL

Observer007:
## 📌 Description

Enable `cum_seq_lens_q` / `max_q_len` for CuTeDSL MLA decode with DCP.

  - Route variable-Q DCP to monolithic CuTeDSL
  - Fix compact O/LSE writes for empty DCP ranks
  - Use request-local Q length in the FP8 DCP mask
  - Add BF16/FP8 correctness tests

## 🔍 Related Issues

https://github.com/flashinfer-ai/flashinfer/issues/4658

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [x] Tests have been added or updated as needed.
- [x] All tests are passing (`unittest`, etc.).

## Summary by CodeRabbit

* **New Features**
  * Added support for compact variable-length queries with distributed context parallelism (DCP).
  * Enabled this support across monolithic MLA attention paths, including BF16 and FP8 execution.
  * Improved handling of varying query lengths, empty workloads, split-KV processing, and causal visibility.
  * Updated compatibility checks and documentation to reflect support for both fixed-length and compact variable-length inputs.

* **Bug Fixes**
  * Corrected output and log-sum-exp handling for empty DCP work with compact queries.
  * Improved request-specific query addressing for variable-length batches.

Issue #3047, opened 2026-04-13
Correctness issue in the FlashInfer-backed MLA chunked-prefill path on Blackwell


yhyang201:
Correctness issue in the FlashInfer-backed MLA chunked-prefill path on Blackwell

## Problem

We started from a user-visible accuracy problem in SGLang on Blackwell: when chunked prefill was enabled, the default MLA path produced much larger logprob errors than expected on some documents.

The reproduction script `chunked_prefill_repro.py` comes from [sglang#22291](https://github.com/sgl-project/sglang/issues/22291).

Repro command:

```bash
python chunked_prefill_repro.py --model=/workspace/deepseek-ai/DeepSeek-V3-Base --tp=8
```

Observed on SGLang's default `trtllm_mla` backend:

```text
=== /workspace/deepseek-ai/DeepSeek-V3-Base | chunked_prefill_size=8192 ===
  Doc 0 ( 10752 tokens): MSE = 0.021289
  Doc 1 (  5909 tokens): MSE = 3.708292
  Doc 2 (  2471 tokens): MSE = 3.269478
  Doc 3 ( 18993 tokens): MSE = 0.006363
  Doc 4 ( 12526 tokens): MSE = 1.603370
  Doc 5 ( 20001 tokens): MSE = 0.000000
  Doc 6 ( 15570 tokens): MSE = 0.016094
  Doc 7 (  2781 tokens): MSE = 1.214392
```

This was first reported in [sglang#22291](https://github.com/sgl-project/sglang/issues/22291).

## First issue we fixed

The first bug we found was in `trtllm_ragged_attention_deepseek`.

### Why kv_len=0 with q_len>0 is valid

DeepSeek MLA stores KV as compressed latents that need decompression before attention. To bound memory, we chunk the prefix KV and attend one chunk at a time, merging via LSE. With different prefix lengths in a batch (e.g. `[256, 512, 768, 1024]`, `chunk_len=256`):

```
chunk 0:  kv_len = [256, 256, 256, 256]
chunk 1:  kv_len = [  0, 256, 256, 256]   ← seq 0 done
chunk 2:  kv_len = [  0,   0, 256, 256]
chunk 3:  kv_len = [  0,   0,   0, 256]
```

Queries are shared across chunks in the ragged batch, so shorter-prefix sequences naturally hit `kv_len=0` in later chunks. Expected: `out=0, lse=-inf`, so the merge ignores them.

### The bug

[...]

PR #4975, opened 2026-09-04
fix(trtllm-gen): write the empty-KV identity for MLA decode rows with seq_lens == 0

YAMY1234:
## 📌 Description

`trtllm_batch_decode_with_kv_cache_mla(backend="trtllm-gen")` leaves the output rows of a request with `seq_lens == 0` untouched. The generation cubin exports `lse = -inf` for such rows but never writes `out`, so the caller sees whatever was in the buffer (`torch.empty` when FlashInfer allocates it). Rows like this are legal and common in decode context parallelism (a rank that owns no KV slice of a request) and in chunked-KV merging: the caller combines rank/chunk partial states with `merge_state`, and a stale `NaN`/`Inf` in `out` poisons the merged result even though its weight is `exp(-inf) = 0`. sglang currently works around it with its own post-launch kernel (`fixup_zero_kv_rows`, sgl-project/sglang#22688, #33881).

This PR materializes the empty-attention identity (`out = 0`, `lse = -inf`) inside the trtllm-gen launcher, right after the generation kernel:

- `csrc/trtllm_fmha_kernel_launcher.cu`: add `FixupEmptyKvRowsKernel`, a tiny kernel with one block per `(request, query token)` that returns immediately unless `seq_lens[b] == 0`, and zeroes that token's `out` row (16-byte stores when aligned) and writes `-inf` to its LSE. It reads `seq_lens` / `cum_seq_lens_q` on the device only, so it is sync-free and CUDA-graph safe (the decode path is graph-replayed by every serving engine).
- Scope: dense MLA decode (`is_mla_decode`) without attention sinks, and only when the caller requests the LSE (`lse != nullptr`), because only a merge can observe an empty row. Plain decode (`return_lse=False`) launches nothing extra, so its cost is unchanged. Generic MHA/GQA decode is not touched: with attention sinks an empty-KV row has a finite LSE (the sink is in the denominator), so the `-inf` identity would be wrong there. Block-sparse attention (per-head `seq_lens`) and sparse MLA are left untouched as well.
- `flashinfer/mla/_core.py`: document the contract on `seq_lens` and `return_lse`, limited to dense MLA decode without sinks on `trtllm-gen` / `cute-dsl`. `cute-dsl` already satisfies it (#4719); this PR brings `trtllm-gen` in line.

The ragged prefill wrapper (`trtllm_ragged_attention_deepseek`) handles the same class of rows by host-side compaction (#3779), which cannot run under CUDA-graph capture (#4703 now skips it there). Extending this device-side fixup to the context kernel so that captured ragged prefill is also well defined is a possible follow-up; not done here to keep the change scoped.

## 🔍 Related Issues

- Follow-up to #3047 / #3779 (the TRTLLM-GEN empty-KV case, decode side).
- Motivated by sgl-project/sglang#33881 (discussion on whether sglang's `fixup_zero_kv` workaround can be removed).

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

New test `tests/attention/test_trtllm_gen_mla.py::test_trtllm_batch_decode_mla_empty_kv_rows_are_neutral` (bf16 and fp8 KV, 16/128 heads, q_len 1/2): pre-fills `out`/`lse` with sentinels, runs a batch with `seq_lens = [130, 0, 5, 0, 0, 300, 64, 0]`, asserts the empty rows are `0` / `-inf` and the active rows match the same batch with the empty rows removed.

Validation on GB300 (SM103, CUDA 13.0, torch 2.13):

- New test: 8/8 passed (`-k empty_kv_rows`), re-run after narrowing the scope to `is_mla_decode && attention_sinks == nullptr` (8/8, MLA bf16 regression 3168 passed again).
- Existing `test_trtllm_batch_decode_mla` (bf16, trtllm-gen, all parametrizations; the LSE-checking subset asserts finite LSE and an untouched workspace guard): 3168 passed. Other trtllm-gen MLA decode entry points (`q1_mla`, `grouped_mla*`, `non_power_of_two_heads*`, `fixed_q_batch_stride`, `variable_q_lengths`, `multi_ctas*`): 104 passed, 1 skipped.
- Generic (non-MLA) trtllm-gen decode, which the fixup must not touch: full `tests/attention/test_trtllm_gen_attention_decode.py` decode matrix including `enable_sink=True`, `test_trtllm_batch_decode_lse_contract` and the gpt-oss counter test: 62841 passed, 29136 skipped.
- Before/after (this branch with and without the launcher change, same cubins), 128 heads, q_len 1, seq_len 4096, page 64, bf16, CUDA-graph replay of 20 calls, median per call:

| batch | `return_lse=False` before → after | `return_lse=True` before → after |
|---|---|---|
| 8 | 17.8 → 17.7 µs | 19.0 → 22.0 µs |
| 64 | 66.1 → 65.6 µs | 67.8 → 70.2 µs |
| 256 | 261.0 → 259.7 µs | 262.9 → 266.0 µs |

The plain decode path is unchanged; the LSE path pays one extra ~3 µs launch per call, which is what a caller would otherwise spend on its own post-launch fixup (sglang's `fixup_zero_kv_rows` measures about the same).

## Reviewer Notes

- The extra launch is gated on `lse != nullptr`. If you would rather always run it (one ~2 µs launch per attention call on the plain decode path), or gate it on an explicit flag instead, happy to change.
- `out` is assumed contiguous `[sum_seq_q, num_heads, head_dim_o]`, matching what the Python wrapper allocates/validates today; the row byte count is derived from `num_qo_heads * head_dim_vo * bits(o_dtype)`.
- Ideally the cubin itself would write the identity; this is a wrapper-level fix because the kernel is closed source.

## Summary by CodeRabbit

- **Bug Fixes**
  - Empty key-value rows during MLA decoding now produce neutral attention results (`out = 0`, `LSE = -inf`) when log-sum-exp output is requested.
  - Prevents empty rows from affecting downstream attention result merges.

- **Documentation**
  - Clarified behavior for empty attention rows, including backend-specific guarantees and behavior when LSE output is not requested.

- **Tests**
  - Added coverage for empty-row handling in decode-context-parallel and chunked-KV merge scenarios.

### DIFF
--- a/flashinfer/mla/_core.py
+++ b/flashinfer/mla/_core.py
@@ -3046,7 +3046,11 @@ def trtllm_batch_decode_with_kv_cache_mla(
     seq_lens : Optional[torch.Tensor]
         Per-request physical KV sequence lengths for dense and TRTLLM-GEN
         paths. With DCP these are rank-local lengths and continue to control
-        paging, memory bounds, and split-KV. For
+        paging, memory bounds, and split-KV. A request with ``seq_lens == 0``
+        (for example a DCP rank that owns no KV slice of it) is the attention
+        identity: when ``return_lse=True`` the ``trtllm-gen`` and ``cute-dsl``
+        backends write ``out = 0`` and ``lse = -inf`` for its query rows so a
+        downstream merge ignores them. For
         SM120/SM121 sparse v32/GLM, pass ``[batch_size, q_len_per_request]`` or
         flattened ``[batch_size * q_len_per_request]`` active top-k lengths; if
         ``None``, every column in ``block_tables`` is active.
@@ -3131,7 +3135,10 @@ def trtllm_batch_decode_with_kv_cache_mla(
     return_lse : bool = False
         Whether to return LSE values. Supported by ``trtllm-gen``,
         ``cute-dsl``, and ``sparse`` backends. When True, the function
-        returns ``(out, lse)``. With compact variable Q, LSE is currently
+        returns ``(out, lse)``, and rows with ``seq_lens == 0`` are
+        guaranteed to hold the empty-attention identity (``out = 0``,
+        ``lse = -inf``); without the LSE their ``out`` rows are left
+        undefined. With compact variable Q, LSE is currently
         supported only by monolithic CuTeDSL.
     cute_dsl_impl : str = "auto"
         Which cute-dsl implementation to use. Honored when
--- a/flashinfer/mla/_core.py
+++ b/flashinfer/mla/_core.py
@@ -3048,8 +3048,9 @@ def trtllm_batch_decode_with_kv_cache_mla(
         paths. With DCP these are rank-local lengths and continue to control
         paging, memory bounds, and split-KV. A request with ``seq_lens == 0``
         (for example a DCP rank that owns no KV slice of it) is the attention
-        identity: when ``return_lse=True`` the ``trtllm-gen`` and ``cute-dsl``
-        backends write ``out = 0`` and ``lse = -inf`` for its query rows so a
+        identity: for dense MLA decode without attention sinks, the
+        ``trtllm-gen`` and ``cute-dsl`` backends write ``out = 0`` and
+        ``lse = -inf`` for its query rows when ``return_lse=True`` so a
         downstream merge ignores them. For
         SM120/SM121 sparse v32/GLM, pass ``[batch_size, q_len_per_request]`` or
         flattened ``[batch_size * q_len_per_request]`` active top-k lengths; if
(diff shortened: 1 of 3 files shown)


flashinfer/mla/_core.py
```diff
@@ -3131,7 +3136,12 @@ def trtllm_batch_decode_with_kv_cache_mla(
     return_lse : bool = False
         Whether to return LSE values. Supported by ``trtllm-gen``,
         ``cute-dsl``, and ``sparse`` backends. When True, the function
-        returns ``(out, lse)``. With compact variable Q, LSE is currently
+        returns ``(out, lse)``. For dense MLA decode without attention
```
PerkzZheng: Please expose an option for users to determine whether they want to set those padded rows to zero or not.

PerkzZheng (approved):

PR #4981, opened 2026-09-04
bench: sm107 whitelist and updates

jimmyzho:
## 1. Norm — fp8 refcheck compared in the wrong units

`benchmarks/routines/norm.py`

The three fp8 refcheck sites cast an fp8 result and a float reference to float
and called `torch.isclose(rtol=1e-1, atol=1e-1)`. Both bounds were wrong, in
opposite directions.

`rtol=1e-1` is **stricter than the dtype can represent**: adjacent fp8 codes are
12.5% apart (e4m3) and 25% apart (e5m2), so a single element where the kernel and
torch round a tie to neighbouring codes exceeded rtol and failed the case.

`atol=1e-1` is **far too loose**, and dominates at small magnitudes. Sweeping the
old predicate over every finite e4m3 code:

| ref | furthest `out` that passed | codes apart |
|---|---|---|
| 4.0 | 3.5 | 2 |
| 1.0 | 0.8125 | 3 |
| 0.25 | 0.140625 | 7 |
| 0.1 | 0.00195 | 28 |

Replaced with `_fp8_code_stats`, an exact comparison in fp8 **code space**. Both
tensors are already the same fp8 dtype, so their codes are mapped to a monotonic
sign-magnitude index and subtracted: distance 0 is bit-equal, 1 is adjacent codes
(a rounding tie), and >1 is a real mismatch. No `rtol`, no `atol`.

Verified across all 256 codes of both dtypes: consecutive finite values differ by
exactly 1, `+0`/`-0` compare equal, subnormals order correctly. NaN is excluded
from the distance and handled separately. Ties are reported at `verbose>=1` so a
systematic one-code bias stays visible rather than silently tolerated.

## 2. Norm — timing closures could not rotate their inputs

`benchmarks/routines/norm.py`

`bench_gpu_time` achieves a cold L2 by cloning the tensors in `input_args` and
rotating between the copies; its docs require `fn` to accept them. The DiT
layernorm and QK-rmsnorm-rope benchmarks passed **zero-argument closures** that
captured their inputs, so every iteration re-read one address, L2 stayed warm and
the reported times were optimistic.

`fused_fn`/`eager_fn` now take `(input_t, residual)` and `run_fused` takes `qkv`,
with the matching `input_args`.

Separately, `testFusedDitLayernorm` hardcoded
`enable_cupti=True, dry_run_iters=10, repeat_iters=100` and never graph-captured,
so `--use_cuda_events`, `--dry_run_iters` and `--num_iters` were inert and this
one routine was timed on a different path from every sibling in the file. It now
honours the flags and CUDA-graphs like the others.

## 3. MoE — fp4_quantize scale-factor padding

`benchmarks/routines/moe.py`

`fp4_quantize` zero-pads the **scale** tensor on both dims — outer to 128
(swizzled SF layout), inner to 4 — while the **data** tensor is unpadded. The
nvfp4 path reshaped the scales with the raw extents.

Two consequences, the second more serious:

1. Where the dims were not already aligned the reshape raised — e.g. gpt-oss at
   `hidden_size=2880`.
2. Where it did not raise, the benchmark **silently substituted `torch.ones()`
   for the scale factors**. 51 of 82 runnable `trtllm_fp4_block_scale_moe` lines
   (~62% of benchmarked *and refchecked* cases) were measured and "validated"
   against fabricated scales.

Now uses the padded extents via `_sf_outer`/`_sf_inner`. The cutlass path in the
same file already did this via `round_up(...)`; this was an inconsistency within
one file. **Any nvfp4 MoE number produced by an unpatched harness at
`num_tokens % 128 != 0` should be treated as void.**

## 4. RoPE — `cos_sin_cache` must be fp32

`benchmarks/routines/rope.py`

`apply_rope_with_cos_sin_cache` asserts `cos_sin_cache.dtype == torch.float32`
(`flashinfer/rope.py:1209`), but the benchmark built it with `input_dtype`, so
every case failed with `cos_sin_cache should be float32` on **every**
architecture. That is why this routine had no testlist on any branch. The three
sibling call sites in the same file already used fp32.

The bandwidth term is corrected to match — the cache read is 4 bytes per element.

## 5. Compilation-flag fixes

`flashinfer/compilation_context.py`, `flashinfer/jit/tinygemm2.py`

`cutlass_supports_sm107()` decided whether to emit native `compute_107a` by
grepping a CUTLASS header for `struct Sm107`. That describes the *template
library*, not the *toolkit* — `compute_107a` only lands in CUDA 13.4. A tree
pairing a Sm107-aware CUTLASS with CUDA 13.2/13.3 emitted a `-gencode` its own
nvcc rejects:

```
nvcc fatal : Unsupported gpu architecture 'compute_107a'
```

which the JIT swallowed into a generic "Ninja build failed" at `rc=0`. All ten
attention testlists produced zero rows while the job looked green.
`FLASHINFER_CUDA_ARCH_LIST=10.0f` does not work around it.

Now asks nvcc directly via `--list-gpu-arch`, resolved the same way the JIT
resolves nvcc, failing closed to the sm100f family target. The fallback is
unconditional rather than gated on `map_sm107_to_100f`, because ~15
`gen_*_module` callers (norm, sampling, topk) never opt in and were emitting the
bad target regardless.

`gen_tinygemm2_module()` additionally needs `map_sm107_to_100f=True`: the no-bias
path (`routergemm.py:509`) reaches this generic module on SM107, bypassing
`gen_tinygemm2_sm100_module()`.

## 6. Attention — GQA group_size 16 dispatch

`include/flashinfer/utils.cuh`

`DISPATCH_GQA_GROUP_SIZE` handled 1/2/3/4/6/8 and sent everything else to
`FLASHINFER_ERROR`. A GQA ratio of 16 (Nemotron 16/1 and 32/2) therefore raised
at dispatch, and because a raise aborts the whole benchmark line it also took
down every other backend sharing that line. Adding the arm took
`attention_decode_fp8` from 15 to 21 `[PERF]` rows.

## 7. cc 10.7 backend rows for 46 routines

`benchmarks/routines/flashinfer_benchmark_utils.py`

`filter_backends_by_compute_capability` does
`cc_to_supported_backends.get(compute_capability, [])`, so a missing key drops
every backend and the case prints `[ERROR] No backends to test` and exits 0.
Only 10 of 59 routines had a `"10.7"` key; this adds 46, taking it to 56.

Every backend listed is backed by an observed `[PERF]` row on GR100 (cc10.7) or
B200 (cc10.0). Nothing is inferred from the sibling `10.3` row — a
plausible-looking row that was never run is the exact failure mode being fixed.

**Five rows deliberately differ from `10.3` and should not be "fixed" to match:**

| Routine | Difference | Reason |
|---|---|---|
| `mm_mxfp8` | drops `cudnn` | library *raises* at cc10.7, killing the whole line. `bmm_mxfp8` keeps cudnn — the guard is per-routine |
| `mxfp4_quantize` | drops `cute-dsl` | no `[PERF]` observed at cc10.7 |
| `chunk_gated_delta_rule` | drops `fla` | **environment, not capability** — `flash-linear-attention` is absent from the image, so it could not emit `[PERF]`. Re-add and verify if CI installs it |
| `recurrent_kda_prefill` | `10.7` empty | observed: the library rejects with `requires compute capability 12.0, got sm_107` |
| `recurrent_kda_prefill` | `10.0` populated | see below |

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## 🔬 Experimental Track

```experimental-tests
#
#
```

## Summary by CodeRabbit

* **New Features**
  * Added support for grouped-query attention with a group size of 16.
  * Improved compatibility with NVIDIA SM107 hardware and CUDA toolchains, including automatic fallback handling when needed.

* **Bug Fixes**
  * Corrected FP4 mixture-of-experts benchmark handling for padded scale tensors.
  * Improved FP8 output validation to detect code-level mismatches more accurately.

* **Benchmarking**
  * Updated normalization and rotary-position-embedding benchmarks for more consistent configurable execution.


bkryu (approved):

jimmyzho:
@flashinfer-bot run

PR #4912, opened 2026-09-03
feat(gdn): adopt the CuTe-DSL disk cache for TVM-FFI GDN kernels

elwhyjay:
## 📌 Description

Migrate the GDN CuTe-DSL call sites that already compile with `--enable-tvm-ffi` to `build_and_load_cute_dsl_kernel`, so each specialization is exported to the on-disk kernel cache and later processes JITLink it instead of recompiling.

Covered call sites (8 compile sites in 5 files):

| File | Sites | Module |
| --- | --- | --- |
| `gdn_decode_nontranspose.py` | small/big batch | `gdn_decode_nontranspose` |
| `gdn_decode_pretranspose.py` | 1 | `gdn_decode_pretranspose` |
| `gdn_decode_mtp.py` | inline, warp | `gdn_decode_mtp` |
| `gdn_decode_bf16_state.py` | wide_vec, wide_vec_t1, mtp_ilp4 | `gdn_decode_bf16_state` |
| `blackwell/gdn_prefill.py` | 1 | `gdn_blackwell_prefill` |

How it works:

- Only the `cute.compile(...)` call is wrapped in a `compile_fn` closure. Every in-process cache structure (getter dicts, `defaults_by_B`, aux buffers) is unchanged, and on a disk hit the closure never runs.
- Specialization names mirror the in-process cache keys, which #4219 and #4436 already audited to cover every codegen parameter.
- Where several entry points share one module, the name adds a variant tag. The Blackwell prefill name also adds `num_sm`, which the compile bakes in as `max_active_clusters`.
- A small shared formatter (`gdn_kernels/cute_dsl_cache_naming.py`) keeps names symbol-safe and caps their length.

Measurements with fresh processes and an isolated cache dir:

| GPU | Selection | Cold (compile + export) | Warm (disk hit) |
| --- | --- | --- | --- |
| RTX PRO 6000 (SM120) | bf16_state + fp32 MTP verify subsets (56 tests) | 604s | **2.3s** |
| RTX PRO 6000 (SM120) | pretranspose + nontranspose basics (7 tests) | 2.9s | **1.4s** |
| B200 (SM100) | prefill suites (2835 tests, includes GPU test execution) | 217s | **110s** |

## 🔍 Related Issues

#4214 (GDN-P1, PR 4 of the suggested implementation sequence)

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

[...]

PR #4982, opened 2026-09-04
feat(MSA): Add NVFP4 paged-KV MSA sparse decode and prefill for SM100/SM103

nv-yunzheq:
## 📌 Description

Compute capability 10.0/10.3 (B200/GB300) had no NVFP4 route for MiniMax Sparse Attention: msa_sparse_attention and msa_sparse_decode_attention rejected packed uint8 K/V and tensor block scales on these parts. This adds that capability behind the existing public API. The only signature change is a new optional out= argument on the decode entry point.

Supported inputs (both routes)
- A packed NVFP4 paged KV cache: one planar 73,728 B page per block laid out as [K data | K scale | V data | V scale], passed as four strided views of the same allocation -- uint8 (num_pages, 4, 128, 64) data and (num_pages, 4, 128, 8) block scales. The kernels read the page in place and never materialize a dense or dequantized copy in global memory.
- e2m1 values with one e4m3 block scale per 16 elements. K scales are stored linearly, V scales in the cache writer's (4, 4) swizzle. Scale views may be uint8 or float8_e4m3fn. k_global_scale and v_global_scale are required.
- BF16 queries and BF16 output. The kernel bodies are built for 64 query heads, 4 KV heads, head_dim 128 and page_size 128; the guard enforces that geometry and the page layout, including the byte offsets between the four views.
- Batch size, per-request KV length (seqused_k), block-table width and the outer strides of q2k_indices are runtime arguments, so the head-major view of a token-major selection buffer is read without a copy.
- A call outside this surface falls through to the pre-existing NotImplementedError, whose message now names the axis that declined.

Decode (msa_sparse_decode_attention)
- top-k in [1, 32], seqlen_q >= 1 for speculative decode, causal or not.
- Two implementations sit behind one route and are chosen internally from the call geometry:
  * csrc/msa_decode_nvfp4_specialized.cu, a warp-specialized CUDA kernel that dequantizes e2m1 to e4m3 with the block scales folded in and runs QK and PV on tcgen05 kind::f8f6f4 with FP32 accumulation. It holds a parametric instantiation family plus a family pinned to the geometry above.
  * flashinfer/msa_ops/cute_dsl/sparse_decode_nvfp4_sm100.py, a CuTe-DSL kernel that dequantizes to FP16, runs mma.sync with FP32 accumulation, splits the selected blocks across a thread-block cluster and combines the partials through distributed shared memory. Both keep an online row maximum and neither issues a native FP4 MMA. FLASHINFER_MSA_DECODE_NVFP4_ROUTE=auto|pingpong|specialised overrides the choice.
- CUDA-graph capturable without an MSASparseAttentionWorkspace: everything before the launch is host-side shape arithmetic and no call shape can trigger a compile. One eager launch per device is needed before capture; msa_decode_nvfp4_specialized_warmup(device) provides it.
- out= lets the caller supply the destination (contiguous BF16 of q.shape), so no output temporary is allocated or copied.

Prefill (msa_sparse_attention)
- Only supports CUDA 13.X. No CUDA 12.x support
- Causal only, no LSE, top-k 16, paged layout with seqused_k as the sole source of per-request KV length (cu_seqlens_k is refused). Per-request query length and total query count are arbitrary.
- csrc/msa_prefill_nvfp4_specialized.cu: each CTA dequantizes the K and V sides of one selected page into its own dynamic shared memory and runs QK and PV on tcgen05 kind::f16 with BF16 operands and FP32 accumulation. The per-tile block union is a 128-slot hash table keyed on block id, so the block-table width is free up to the 24-bit block id (about 2^31 tokens at page_size 128) and the route allocates nothing beyond the output.
- The softmax fast path exponentiates against a fixed origin, paired with a range check on the row denominator and an exact replay, against a data-derived running maximum, of any tile that fails it. Output is bitwise reproducible run to run up to 16,384 tokens of context per request; above that the accumulation order may vary while staying correct.

Packaging and introspection
- Both kernels are registered in flashinfer/aot.py for sm100a and sm103a and build on CUDA 12.8+ toolkits: the BF16/E4M3/E2M1 conversions use native cvt instructions on CUDA 13.1+/13.2+ and bit-exact fallbacks below.
- msa_{decode,prefill}_nvfp4_specialized_stats() and the capability manifests in flashinfer/msa_ops/*_workloads.json expose the supported surface and the CUDA-graph contract to callers.
- An FP32 composable reference for each route is exported for tests.

Tests
- Host-only guard tests for both routes, and device tests against the FP32 reference across batch sizes, KV lengths, top-k values, block-table widths, multi-token decode, empty selections, strided selection views, both decode implementations, CUDA-graph capture and out=.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## 🔬 Experimental Track

```experimental-tests
#
#
```

## Summary by CodeRabbit

- **New Features**
  - Added specialized NVFP4 paged-KV sparse attention prefill and decode support for Blackwell SM100/SM103 GPUs.
  - Added support for packed KV-cache layouts, tensor parallel head configurations, causal and variable-length workloads, CUDA Graph capture, warmup, and runtime capability reporting.
  - Decode now supports writing directly into a caller-provided `out` tensor.
  - Added cached JIT loading for specialized prefill and decode kernels.

- **Documentation**
  - Documented supported layouts, limits, reproducibility behavior, CUDA requirements, and route selection.

- **Bug Fixes**
  - Unsupported configurations now report the specific reason they were declined.


nv-yunzheq:
@flashinfer-bot run test/msa_ops

nv-yunzheq:
@flashinfer-bot run tests/msa_ops

flashinfer/msa_ops/_nvfp4_decode_sm100.py
```diff
@@ -0,0 +1,1784 @@
...
+    return None
+
+
+# ---------------------------------------------------------------------------
+# the second implementation, and the predicate that selects it
+# ---------------------------------------------------------------------------
+def _route_choice() -> str:
+    """``auto`` unless overridden. Read at call time, never at import."""
+
+    value = (os.environ.get(_ROUTE_ENV) or "auto").strip().lower()
```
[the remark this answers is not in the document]
bkryu: This should be addressed.

csrc/msa_decode_nvfp4_specialized.cu
```diff
@@ -0,0 +1,2528 @@
...
+//
+//   cluster  chunks  rescales  rate          the same code without the rescale
+//      2        2        1     3.37 / 3.40%  0.07% / 0.00%
+//      2        4        3    19.67 / 18.73% --
+//      4        2        1     7.87%         --
+//      1        4        3     0.00%  (0 of 3,000, twice, with and without
+//      1        8        7     0.00%   the split-cluster PV ordering)
+//
+// so the rescale is the site and a cluster is its precondition.  WHY a cluster
+// opens the window is not explained: nothing in the rescale is cluster-scoped
```
bkryu, line 265: Could maybe trim comments a bit? Performance might be fine but a comment saying "I don't know why" seems a bit odd to have in the codebase.

flashinfer/msa_ops/cute_dsl/sparse_decode_nvfp4_sm100.py
```diff
@@ -0,0 +1,2822 @@
...
+        # The L2 prefetch pays only where the grid, not the register file, is
+        # what limits the warps in flight.  A four-resident-CTA binary already
+        # has 13.8 warps/SM hiding its DRAM, sits exactly on the 128-register
+        # cliff, and streams a working set far larger than L2 -- there the
+        # prefetch's stream-coordinate arithmetic and extra live registers cost
+        # more than the latency it removes (measured b64 44 -> 48 us, b128
+        # 75 -> 84 us).  Two-resident-CTA binaries are grid-limited with 128
+        # spare registers and gain (b32 27 -> 24 us).
+        pf=pf, hoist=hoist, mbpm=mbpm, qreg=qreg,
+    )
```
bkryu, line 2333: I don't think we are compiling with tvm-ffi compatibility here. Can you take a look?

flashinfer/msa_ops/_nvfp4_decode_sm100.py
```diff
@@ -0,0 +1,1784 @@
...
+    Two separate preconditions, and both are this function's job:
+
+    * the implementation's call path never compiles, so its variant table has
+      to be built here or ``run`` on that side declines;
+    * the first LAUNCH of an instantiation is what makes the driver resolve it
+      (``CUDA_MODULE_LOADING=LAZY``), and that may not happen inside a
+      CUDA-graph capture -- so one eager launch per compiled instantiation is
+      taken here, on the smallest batch that selects it.
+
+    Compilation is minutes of ptxas on a cold cache. It happens once per
```
bkryu, line 881: Can you look at examples like https://github.com/flashinfer-ai/flashinfer/pull/4912 and persist the cute-dsl kernel on disk? Not sure if the frameworks will like spending a few minutes to compile a kernel that gets purged at the end of the process

flashinfer/msa_ops/_nvfp4_decode_sm100.py
```diff
@@ -0,0 +1,1784 @@
...
+number is the graph column. But an eager decode step -- capture disabled, or a
+batch size outside the captured rungs -- pays that difference ONCE PER LAYER,
+and it is three to fifteen times the kernel's own device time.
+
+Nothing here acts on that. Routing on ``is_current_stream_capturing()`` would
+put a different kernel in the captured graph than in the eager warm-up that
+precedes it, which is a behaviour change with its own numerics to measure, and
+the two implementations already differ by more than either differs from the
+FP32 reference. It is stated so the next serving measurement starts from it
+rather than rediscovering it.
```
bkryu, line 108: Overall there seems to be too much perf numbers in the comments. Let's try to clean up a bit

docs/api/sparse.rst
bkryu, line 1: Do we need to state that the additions here (NVFP4 paged route and out=) are SM100/103 only

flashinfer/msa_ops/_blackwell_sm100.py
```diff
@@ -1985,6 +2000,181 @@ def _run_decode_module(
...
+    return_softmax_lse: bool,
+    return_temperature_lse: bool,
+    lse_temperature_scale: float,
+    k_scale,
+    v_scale,
+    k_global_scale,
+    v_global_scale,
+    q_offset,
+    workspace: Optional[MSASparseAttentionWorkspace],
+) -> Optional[torch.Tensor]:
```
bkryu, line 2043: Both helpers are annotated `-> Optional[torch.Tensor]` but return `(tensor | None, reason | None)` tuples, and the docstrings still describe the old "return None to fall through" contract. Should be `Tuple[Optional[torch.Tensor], Optional[str]]` I think?

flashinfer/msa_ops/sparse_decode.py
```diff
@@ -242,11 +247,17 @@ def msa_sparse_decode_attention(
         16 elements, rows padded to a multiple of 128). Scale rows follow the
         cache layout: ``(token, head)`` order for flat K/V, ``(page, head,
         token)`` for paged. SM120/SM121-only.
+        On compute capability 10.0/10.3 the paged NVFP4 decode route instead
```
saltyminty: This reads a little strange – the previous lines are SM120/121 only, this line is 10.0/10.3 only?

csrc/msa_decode_nvfp4_specialized.cu
```diff
@@ -0,0 +1,2528 @@
...
+                          : pick(std::false_type{}, split, fast, odd_tail);
+
+  cudaError_t status =
+      cudaFuncSetAttribute(kernel, cudaFuncAttributeMaxDynamicSharedMemorySize, kSmemBytes);
+  if (status != cudaSuccess) return status;
+  status = cudaFuncSetAttribute(kernel, cudaFuncAttributePreferredSharedMemoryCarveout, 85);
+  if (status != cudaSuccess) return status;
+
+  cudaLaunchConfig_t config{};
+  config.gridDim = dim3(num_kv_heads * split, total_q);
```
saltyminty, line 1329: My agent flagged that `total_q` requires a guard to avoid exceeding the max 65535 limit

flashinfer/msa_ops/_nvfp4_decode_sm100.py
```diff
@@ -0,0 +1,1784 @@
...
+        topk=int(q2k_indices.shape[2]),
+        page_size=int(k.shape[2]),
+        seqlen_q=int(seqlen_q),
+        causal=int(bool(causal)),
+        softmax_scale=float(softmax_scale),
+        k_global_scale=float(k_global_scale),
+    )
+    if reason is not None:
+        return reason
+    if device_warm and not module.is_warm(q.device):
```
saltyminty: `is_warm` only checks whether warmup occurred, but that compilation/run actually succeeded. We should be checking `_specialised_warm_failed`/`_specialised_warm_devices` here instead – otherwise, auto mode could select cute, fail, and still dispatch to cute at runtime.

flashinfer/msa_ops/cute_dsl/sparse_decode_nvfp4_sm100.py
```diff
@@ -0,0 +1,2822 @@
...
+            sPT[j] = ld_global_s32(pt_row + Int64(4) * Int64(j))
+
+    # ---- Q global read ------------------------------------------------------
+    # Issued AFTER the selection list and the block-table row, and consumed
+    # only once the first KV chunk is already in flight.  All three are in the
+    # same shadow, but the CTA-serial chain that gates every later load is
+    # q2k / page_table -> validity scan -> KV, and Q is 4 KiB per CTA against
+    # the block-table row's 512 B.  Letting Q enter the queue first put the
+    # BIGGEST request ahead of the one the whole prologue waits on.
+    q_row_base = (qi * num_qo_heads + h * grp) * Int32(HEAD_DIM)
```
saltyminty: My agent is flagging that `q_row_base` (and the later `o_base` on line 2017) can overflow and should be Int64. It is later cast to Int64 on line 929, but it could already have overflowed here.

saltyminty (commented):
Seems like there's CUDA 12.9 failures in CI, can you check?

flashinfer/msa_ops/_nvfp4_decode_sm100.py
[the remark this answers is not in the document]
nv-yunzheq: Fixed

csrc/msa_decode_nvfp4_specialized.cu
nv-yunzheq: Addressed

flashinfer/msa_ops/_nvfp4_decode_sm100.py
nv-yunzheq: addressed

flashinfer/msa_ops/_nvfp4_decode_sm100.py
nv-yunzheq: Cleaned up in [0f015e8](https://github.com/flashinfer-ai/flashinfer/pull/4982/commits/0f015e8050c8b5b7d3c177eb41d22313dd5d623b)

docs/api/sparse.rst
nv-yunzheq: [0f015e8](https://github.com/flashinfer-ai/flashinfer/pull/4982/commits/0f015e8050c8b5b7d3c177eb41d22313dd5d623b)

flashinfer/msa_ops/_blackwell_sm100.py
nv-yunzheq: Fixed in [0f015e8](https://github.com/flashinfer-ai/flashinfer/pull/4982/commits/0f015e8050c8b5b7d3c177eb41d22313dd5d623b)

flashinfer/msa_ops/sparse_decode.py
nv-yunzheq: Fixed in [0f015e8](https://github.com/flashinfer-ai/flashinfer/pull/4982/commits/0f015e8050c8b5b7d3c177eb41d22313dd5d623b)

csrc/msa_decode_nvfp4_specialized.cu
nv-yunzheq: Addressed in [f95c05d](https://github.com/flashinfer-ai/flashinfer/pull/4982/commits/f95c05daa35caf57261c9aee63ff362c318cb687)

flashinfer/msa_ops/_nvfp4_decode_sm100.py
nv-yunzheq: Addressed in [f95c05d](https://github.com/flashinfer-ai/flashinfer/pull/4982/commits/f95c05daa35caf57261c9aee63ff362c318cb687)

flashinfer/msa_ops/cute_dsl/sparse_decode_nvfp4_sm100.py
nv-yunzheq: Addressed in [f95c05d](https://github.com/flashinfer-ai/flashinfer/pull/4982/commits/f95c05daa35caf57261c9aee63ff362c318cb687)

nv-yunzheq:
@flashinfer-bot run tests/msa_ops

nv-yunzheq:
@flashinfer-bot run tests/msa_ops

nv-yunzheq:
@flashinfer-bot run

nv-yunzheq:
@flashinfer-bot run

flashinfer/msa_ops/_blackwell_sm100.py
```diff
@@ -2341,10 +2753,42 @@ def blackwell_msa_sparse_decode_attention(
     partial_dtype: Optional[torch.dtype] = None,
     force_fused: Optional[bool] = None,
     workspace: Optional[MSASparseAttentionWorkspace] = None,
+    out: Optional[torch.Tensor] = None,
 ):
     """Run sparse decode on compute capability 10.0 or 10.3."""
 
     del partial_dtype
+    specialized, nvfp4_decline_reason = _try_nvfp4_decode(
```
saltyminty, line 2761: Doesn't need to be a blocker now, but I worry that this structure of the branch basically be "try nvfp4, proceed with orgiginal if not nvfp4 or not supported" will quickly become difficult to maintain if the number of variants grows.

saltyminty (approved):
Left one non-blocking comment for thought but approved.

flashinfer/msa_ops/_blackwell_sm100.py
nv-yunzheq: Agree, let's list it as a follow-up task. The code structure right now basically doing support check alongside the actual execution. We should clean up the logic later and force to do support check first and then doing execution through different module

nv-yunzheq:
@flashinfer-bot run

nv-yunzheq:
@flashinfer-bot run

Issue #65, opened 2024-01-10
[Tracking Issue] Setting up CI and Performance Regression Testing


yzh119:
We need to set up CI to guarantee the robustness of FlashInfer, following is the list of unit tests:
1. C++ API tests (kernel correctness).
2. PyTorch binding Tests (kernel correctness & end-to-end correctness).
3. TVM Binding Tests (kernel correctness & end-to-end correctness).

Test environments: cuda 12.3, on RTX 4090 (sm89), A100 (sm80).

Performance Regression Testing
1. End-to-end tests (through MLC-Serving and Punica), criterions: throughput and End-to-end Latency.
2. Kernel performance (RTX 4090, RTX Ada 6000, A100 PCIe 80GB).

sricketts:
@yzh119 this looks pretty stale. Marking as `wontfix` and closing, as this is superseded by more updated NV-internal CI testing. Of course please re-open if you disagree.

PR #4986, opened 2026-09-05
perf(topk_varlen): gvr_2 dispatch rungs for 4K-8K rows, SM-count-aware register band, hint-free gvr_2 in auto

dhiraj113:
## 📌 Description

Follow-up to #4811 (these two commits were pushed to its branch after the squash-merge, so they never landed). Both are host-side changes to the `gvr_2` self-sampling top-k backend; no kernel code changes.

**1. `gvr_2` dispatch rungs for compressed widths 4K < n <= 8K (`c7f9075f`).** An SGLang report ("gvr_v2 slower than `topk_transform_512_v2` at 32K with c4") reproduced with their input recipe held in exactly one band: compressed width 6144-8192, where `gvr_2` was 3-7% slower at bs <= 128 and 13-21% slower at bs 256, while already 8-45% faster at every other width from 4096 to 131072. The cause is the dispatch, not the kernels: for `1024 < n4 <= 4096` the upstream `route()` runs every `wide` row on the VPT=4 register kernel (two empty float4 slots per thread up to n = 8192) and sends b > 148 to the streaming slab. Two FlashInfer-local rungs (marked inline, reported to the kernel owner as internal DKG issue #61): `n4 <= 2048` and `wide` -> `_reg(1024, 2, 1, 2*NB)` (18-20% faster for K in {512, 1024, 2048}); `n4 <= 2048` and one wave of `MINB=2` CTAs -> `_reg(512, 4, 2, NB)` instead of the slab (1.3-1.7x), with `NBSEL` following so `IMGOFF == NBH`.

**Hint-free `gvr_2` (same commit).** `pre_idx` is now optional for `gvr_2`: `run_varlen(pre_idx=None, top_k=K)` runs on a cached `arange(k)` table (`_hint_free_pre_idx`, per (device, k), grown eagerly, refused under CUDA-graph capture, pre-sized by `warmup_varlen`). The kernels use the hint only as a sampling anchor, so the result is exact; hint-free is within ~10% of the hinted time on most cells and 1.5-5x ahead of `radix` / `radix_filter` everywhere except K >= 2048, N <= 4096, B = 1 (radix_filter 1.3x faster). `auto` ranks `gvr_2` first for fp32 with or without a hint, with that one cell carved out to `radix_filter`; a malformed hint on `backend="gvr_2"` is discarded with the existing RuntimeWarning and the call still runs on `gvr_2`; the "requires pre_idx" refusal now applies to `gvr` only.

**2. SM-count-aware register band (`4cb98eab`).** `route()` hard-codes 148 (the B200 SM count). Tuning the same band on B300 (160 SMs) and Rubin (208 SMs) showed the remaining losses sat exactly where that constant mis-sizes a wave: rows in (148, sms] are one wave of 1024-thread CTAs there and the register kernels beat the slab by 1.4-1.8x (B300 12288 x 160: 7.2 -> 4.6 us; Rubin 12288 x 160-192: 5.6 -> 4.0 us). `route(b, n, npad, k, sms=148)` takes the device SM count (`_sm_count()`, part of both launcher cache keys) for the register-band tests only: `wide`, the `QC`/`CURE` flags, and the BLK=512 rung's wave cutoffs (`b <= 2*sms`, plus a second wave `b <= 4*sms` from n = 6144 up). The streaming constants stay at 148 on every part: scaling them made the 131072 x 192-256 slab cells 3-28% slower on B300 and Rubin.

**Measured** (same-node, one process per node, isolated computelab B200 / B300 nodes and Rubin; sglang `topk_v2` vs `gvr_2` with the upstream dispatch vs with this PR; K = 512, perfect hint, CUDA-graph replay medians; 84 uniform cells = 7 widths 4096-131072 x 12 batch sizes 1-512, 48 ragged cells = 4 widths x 12 batch sizes, per-row lengths uniform in [1, width]). `gvr_2` faster-or-equal than sglang:

| GPU | uniform, upstream dispatch | uniform, this PR | ragged, upstream | ragged, this PR |
|---|---|---|---|---|
| B200 (148 SMs) | 65 / 84 | 83 / 84 | 35 / 48 | 45 / 48 |
| B300 (160 SMs) | 62 / 84 | 82 / 84 | 31 / 48 | 43 / 48 |
| Rubin (208 SMs) | 58 / 84 | 84 / 84 | 31 / 48 | 46 / 48 |

At the reported cell (width 8192, B200): sglang 4.27-4.48 us vs `gvr_2` 3.62-3.76 us for bs 1-128 (was 4.46-4.58); 6.77 vs 4.89 us at bs 256 (was 8.04). Every `gvr_2` cell in every run was value-multiset exact. What sglang still wins: 1-4 randomly short rows at width 4096 (~1 us calls; its short-row early exit is 0.3-0.5 us cheaper), the second CTA wave at width 12288 (B200 bs 160 0.99; B300 bs 256-320 0.91-0.97), and a few B300 ragged cells within 7%. Note sglang's kernel fuses the page-table transform, so it does slightly more work per call.

Tried and rejected (measured, same-node): swapping the `regimg` flavour for plain `reg` at n <= 4K (+5%..-12% by cell and hint quality); clustered-register configs for 8K < n <= 16K at b > sms (2-4x slower); `_reg(1024, 4, 1)` for the second wave of 8K-16K (mixed); scaling the streaming constants (see above).

## 🔍 Related Issues

Follow-up to #4811. Dispatch findings reported to the TRT-LLM kernel owner as internal DKG issue #61 (the DSL host in TensorRT-LLM has the same rungs).

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

New tests: hint-free exactness on every kernel family incl. short/MTP/cr=4 rows; hint-free CUDA-graph replay with table-growth refusal; host `top_k` contract; the rungs (tuple, `NBH == IMGOFF`, wave cutoffs, `route_split` parity); SM-count-aware dispatch (rungs at sms / 2*sms / 4*sms for sms in {160, 208}, streaming half identical for every sms); `auto` order with and without a hint; per-backend malformed-hint behaviour; family-parity key carries the SM count. `tests/topk_varlen` on the final code: A100, L40S, H100, RTX 5080 (74-78 passed, `gvr_2` skips), B100, DRIVE P2021, isolated computelab B200 and B300, Rubin (323-324 passed, 2 known xfails).

## 🔬 Experimental Track

```experimental-tests
#
#
```

## Reviewer Notes

Both commits are FlashInfer-local deviations from the vendored TRT-LLM host dispatch (`route()` is otherwise a pure mirror); each deviation is marked inline with its measurement. The `logs/`-style benchmark scripts are not part of the PR.

## Summary by CodeRabbit

* **New Features**
  * GVR2 varlen top-k execution supports workloads without a precomputed index hint.
  * Backend selection adapts to hint availability, top-k values, workload size, and GPU resources.
  * Workspace management supports separate allocations per device and CUDA stream.

* **Bug Fixes**
  * Improved validation and warnings for malformed hints and missing or invalid top-k values.
  * Improved CUDA Graph reliability through stream-specific warmup and workspace requirements.
  * Improved correctness for concurrent multi-stream and multi-device execution.

* **Performance**
  * Expanded SM-aware optimization coverage across GPU configurations and larger top-k ranges.

### DIFF
--- a/flashinfer/topk_varlen/kernels/gvr2_topk_host.py
+++ b/flashinfer/topk_varlen/kernels/gvr2_topk_host.py
@@ -23,7 +23,11 @@ bound at the physical width and inflates only the routing value (DKG #60);
 ``validate_run_ws`` enforces the 16-byte alignment the compiled workspace
 declares; ``run_varlen`` rejects overlapping row layouts; ``run``/``run_ws``/
 ``run_varlen`` re-enter under the logits device and every compile passes an
-explicit ``--gpu-arch``. FlashInfer's ``top_k_varlen(backend="gvr_2")``
+explicit ``--gpu-arch``; ``route()`` adds two register-kernel rungs in the
+4K < n <= 8K band (VPT=2 for b <= 148, BLK=512 for 148 < b <= 296 — see the inline
+note; DKG #61); ``run_varlen`` accepts ``pre_idx=None`` and runs
+hint-free on a cached ``arange(k)`` anchor (``_hint_free_pre_idx``).
+FlashInfer's ``top_k_varlen(backend="gvr_2")``
 calls ``run_varlen`` below; the batch-uniform ``run``/``run_ws`` entries are
 kept for parity tests and benchmarking. Keep future diffs against upstream
 mechanical.
@@ -153,7 +157,11 @@ BLKC = 1024  # CTA size of the clustered register path
 
 
 def route(b: int, n: int, npad: int, k: int) -> dict[str, object]:
-    """Mirror of the CUDA gvr_topk_launch dispatch. Pure. See module doc."""
+    """Mirror of the CUDA gvr_topk_launch dispatch. Pure. See module doc.
+
+    FlashInfer-local deviations (both in the 4K < n <= 8K band, marked
+    inline): the wide VPT=2 rung and the 148 < b <= 296 BLK=512 register
+    rung."""
     if b < 1:
         raise RuntimeError(f"route requires b >= 1, got {b}")
     # 148 is the B200 SM count, baked in by the upstream CUDA dispatch (route()
@@ -171,7 +179,11 @@ def route(b: int, n: int, npad: int, k: int) -> dict[str, object]:
     DEGE = (n <= 3 * k) or (n <= 4 * k + 64)
     if DEGE and CMP < n:
         CMP = n
-    NBSEL = (2 * NB) if (n4 > 512 and not (n4 <= 1024 and not wide)) else NB
+    # FlashInfer-local: the `n4 <= 2048 and not wide` window runs the BLK=512
+    # register kernel with NBH = NB for b <= 296 (upstream: main slab, which
+    # ignores NBSEL), so the NB selector must follow (IMGOFF == NBSEL == NBH
+    # is asserted at launch).
+    NBSEL = (2 * NB) if (n4 > 512 and not (n4 <= 2048 and not wide)) else NB
     IMGOFF = NBSEL
     smem_reg = (NBSEL + 2 * CMP) * 4
 
@@ -266,6 +278,23 @@ def route(b: int, n: int, npad: int, k: int) -> dict[str, object]:
                 "ws": False,
             }
 
+    # FlashInfer-local rungs for the 4K < n <= 8K band (the upstream dispatch
+    # runs `wide` rows here on the VPT=4 kernel below and everything else on
+    # the streaming slab). Measured on B100/B200, K in {512, 1024, 2048},
+    # exact in every cell (logs/sglang_perf/force_plan*.py):
+    #   * wide (b <= 148): VPT=2 covers n4 <= 2048 exactly; the VPT=4 kernel
+    #     iterates two empty float4 slots per thread in every pass and is
+    #     ~18-20% slower (4.2 -> 3.4 us at n=8192, K=512).
+    #   * 148 < b <= 296 (one wave at MINB=2, two CTAs per SM): the
+    #     BLK=512/VPT=4 register kernel beats the main slab by 1.3-1.7x
+    #     (8.0 -> 4.9 us at b=256, n=8192, K=512). Beyond one wave the two
+    #     trade places by shape (b=512: -13%..+12%), so the slab keeps it.
+    # Reported upstream for adoption as DKG issue #61.
+    if n4 <= 2048:
+        if wide:
+            return _reg(1024, 2, 1, 2 * NB)
+        if b <= 296:
+            return _reg(512, 4, 2, NB)
     if n4 <= 4096 and wide:
         return _reg(1024, 4, 1, 2 * NB)
 
@@ -1024,6 +1053,42 @@ builder in _build_launcher; only the main family takes the workspace.
 # shape key (b, n, npad, k) -> (fn, args tuple of python ints, needs_ws)
 _LAUNCH_CACHE = {}
 _DUMMY_KV = {}
+# hint-free entry: (device index, k) -> int32[cap, k] table whose every row
+# is arange(k). Superseded tables are kept alive (captured graphs may still
+# address them).
+_HINT_FREE = {}
+_HINT_FREE_KEEP = []
+
+
+def _hint_free_pre_idx(batch, k, device):
+    """FlashInfer-local: the ``pre_idx`` stand-in for hint-free ``run_varlen``.
+
+    The kernels consume the hint only as a sampling anchor — the block-wide
+    (min, max) of ``logits[pre_idx[j]]`` — and never for exactness, so ANY k
+    distinct in-range indices are a valid hint; ``arange(k)`` is the cheapest
+    (rows shorter than k take the in-kernel identity path before the anchor
+    matters, and columns < k always lie inside the row's storage). One table
+    per (device, k), grown by doubling and reused as a contiguous row-prefix
+    view (16-byte aligned: k is a multiple of 512), so the hot path performs
+    no allocation and CUDA-graph replays see a stable address. Growth is
+    refused under capture — run one eager call (or ``warmup_varlen``) at the
+    largest batch first, like every other launcher resource.
+    """
+    key = (device.index if device.index is not None else torch.cuda.current_device(), k)
+    t = _HINT_FREE.get(key)
+    if t is None or t.shape[0] < batch:
+        if _is_capturing():
+            raise RuntimeError(
+                f"hint-free gvr_2: no pre_idx table for batch={batch}, k={k} on "
+                f"{device}; run one eager call (or warmup_varlen) at this batch "
+                "before CUDA-graph capture"
+            )
+        cap = max(batch, 64 if t is None else 2 * t.shape[0])
+        new = torch.arange(k, dtype=_I32, device=device).unsqueeze(0).expand(cap, k).contiguous()
+        if t is not None:
+            _HINT_FREE_KEEP.append(t)
+        _HINT_FREE[key] = t = new
+    return t[:batch]
 
 
 def _dummy_kv(dev_index, device):
@@ -1302,7 +1367,7 @@ def run_ws(
 
 def run_varlen(
     logits: torch.Tensor,
-    pre_idx: torch.Tensor,
+    pre_idx: torch.Tensor | None,
     kv_lens: torch.Tensor,
     indices: torch.Tensor,
     next_n: int = 1,
@@ -1311,8 +1376,15 @@ def run_varlen(
     max_seq_len: int | None = None,
     engine: str = "auto",
     workspace: torch.Tensor | None = None,
+    top_k: int | None = None,
 ) -> None:
     """Production-contract varlen entry (per-row device kv_lens).
+
+    HINT-FREE (FlashInfer-local): ``pre_idx=None`` runs with the cached
+    ``arange(k)`` stand-in of ``_hint_free_pre_idx`` and then REQUIRES
+    ``top_k`` (normally ``k = pre_idx.shape[1]``). Exact like the hinted
+    call; only the sampling anchor is weaker. When both are given they must
+    agree.
 
     Row semantics (mirror of ``heuristicTopKDecode.cu`` and the in-tree
     ``cute_dsl_gvr_topk_decode`` runner):
@@ -1360,6 +1432,7 @@ def run_varlen(
                 max_seq_len=max_seq_len,
                 engine=engine,
                 workspace=workspace,
+                top_k=top_k,
             )
     if logits.dtype is not torch.float32:
         raise RuntimeError(
@@ -1388,6 +1461,12 @@ def run_varlen(
     batch = num_rows // nn
     if kv_lens.shape[0] != batch:
         raise RuntimeError(f"kv_lens length {kv_lens.shape[0]} != num_rows/next_n = {batch}")
+    if pre_idx is None:
+        if top_k is None:
+            raise RuntimeError("run_varlen: top_k is required when pre_idx is None (hint-free)")
+        pre_idx = _hint_free_pre_idx(batch, _index(top_k), logits.device)
+    elif top_k is not None and len(pre_idx.shape) == 2 and pre_idx.shape[1] != _index(top_k):
+        raise RuntimeError(f"top_k={top_k} != pre_idx.shape[1]={pre_idx.shape[1]}")
     if len(pre_idx.shape) != 2 or pre_idx.shape[0] != batch:
         raise RuntimeError(
             f"pre_idx must be [batch={batch}, k] REQUEST-level, got {tuple(pre_idx.shape)}"
@@ -1623,6 +1702,9 @@ def warmup_varlen(
     # rows) even when CUDA-graph batch lists reach thousands of rows.
     n_env_c = max(1, int(max_seq_len) // int(compress_ratio))
     npad_c = (n_env_c + 63) // 64 * 64 if row_stride is None else int(row_stride)
+    # hint-free callers capture with the arange table: size it for the
+    # largest requested batch now (growth is refused under capture)
+    _hint_free_pre_idx(req_rows[-1] // nn, int(top_k), torch.device("cuda", dev))
     seen_keys = set()
     rows_list = []
     r = nn
--- a/flashinfer/topk_varlen/kernels/gvr2_topk_host.py
+++ b/flashinfer/topk_varlen/kernels/gvr2_topk_host.py
@@ -24,8 +24,9 @@ bound at the physical width and inflates only the routing value (DKG #60);
 declares; ``run_varlen`` rejects overlapping row layouts; ``run``/``run_ws``/
 ``run_varlen`` re-enter under the logits device and every compile passes an
 explicit ``--gpu-arch``; ``route()`` adds two register-kernel rungs in the
-4K < n <= 8K band (VPT=2 for b <= 148, BLK=512 for 148 < b <= 296 — see the inline
-note; DKG #61); ``run_varlen`` accepts ``pre_idx=None`` and runs
+4K < n <= 8K band (VPT=2 for b <= sms, BLK=512 for sms < b <= 2*sms) and sizes its
+register band by the device SM count ``sms`` instead of the hard-coded 148
+(see the inline note; DKG #61); ``run_varlen`` accepts ``pre_idx=None`` and runs
 hint-free on a cached ``arange(k)`` anchor (``_hint_free_pre_idx``).
 FlashInfer's ``top_k_varlen(backend="gvr_2")``
 calls ``run_varlen`` below; the batch-uniform ``run``/``run_ws`` entries are
@@ -83,6 +84,20 @@ def _device():
             import gvr2_topk_decode as _m
         _dev_mod = _m
     return _dev_mod
+
+
+_SM_COUNT = {}
+
+
+def _sm_count() -> int:
+    """SM count of the CURRENT device (per-device cache). FlashInfer-local
+    knob of ``route()``'s register-resident band (``wide`` / one-wave tests);
+    the streaming-path constants stay at upstream's 148 — see route()."""
+    d = torch.cuda.current_device()
+    v = _SM_COUNT.get(d)
+    if v is None:
+        v = _SM_COUNT[d] = int(torch.cuda.get_device_properties(d).multi_processor_count)
+    return v
 
 
 def _arch_token() -> str:
@@ -156,26 +171,33 @@ CMPC = 4096  # crossing-bin slots per CTA, clustered register path
 BLKC = 1024  # CTA size of the clustered register path
 
 
-def route(b: int, n: int, npad: int, k: int) -> dict[str, object]:
+def route(b: int, n: int, npad: int, k: int, sms: int = 148) -> dict[str, object]:
     """Mirror of the CUDA gvr_topk_launch dispatch. Pure. See module doc.
 
-    FlashInfer-local deviations (both in the 4K < n <= 8K band, marked
-    inline): the wide VPT=2 rung and the 148 < b <= 296 BLK=512 register
-    rung."""
+    FlashInfer-local deviations, marked inline: the register-resident band
+    is sized by ``sms`` (the device SM count; upstream hard-codes 148) — the
+    ``wide`` one-wave test, QC/CURE and the one-wave cutoff of the BLK=512
+    rung — and the 4K < n <= 8K band has two extra rungs (wide VPT=2, and
+    sms < b <= 2*sms BLK=512). The streaming-path constants keep upstream's
+    148 on every part."""
     if b < 1:
         raise RuntimeError(f"route requires b >= 1, got {b}")
     # 148 is the B200 SM count, baked in by the upstream CUDA dispatch (route()
     # is a pure mirror of it). It only steers occupancy heuristics (one-wave
-    # tests, split factors, cluster sizing), never correctness: on Rubin
-    # (SM107, 208 SMs) the same constants are merely conservative. Retuning
-    # per-arch is a perf follow-up, not a functional requirement.
-    wide = b <= 148
+    # tests, split factors, cluster sizing), never correctness. FlashInfer-
+    # local: the REGISTER-band tests use the real SM count `sms` (B300 160,
+    # Rubin 208): rows in (148, sms] are one wave of 1024-thread CTAs there
+    # and the register kernels beat the slab by 1.4-1.8x (B300 12288 x 160:
+    # 7.2 -> 4.6 us; Rubin 12288 x 160-192: 5.6 -> 4.0 us; measured same-node
+    # vs sglang, DKG #61). The STREAMING constants below stay at 148: scaling
+    # them too made the 131072 x 192-256 slab cells 3-28% slower on both parts.
+    wide = b <= sms
 
     # ======================= register-resident block ========================
     n4 = n >> 2
     CMP = n if n < 2560 else 2560
-    QC = 1024 if b > 148 else QUADC
-    CURE = not (n < 2 * k and b > 148)
+    QC = 1024 if b > sms else QUADC
+    CURE = not (n < 2 * k and b > sms)
     DEGE = (n <= 3 * k) or (n <= 4 * k + 64)
     if DEGE and CMP < n:
         CMP = n
@@ -285,15 +307,19 @@ def route(b: int, n: int, npad: int, k: int) -> dict[str, object]:
     #   * wide (b <= 148): VPT=2 covers n4 <= 2048 exactly; the VPT=4 kernel
     #     iterates two empty float4 slots per thread in every pass and is
     #     ~18-20% slower (4.2 -> 3.4 us at n=8192, K=512).
-    #   * 148 < b <= 296 (one wave at MINB=2, two CTAs per SM): the
+    #   * sms < b <= 2*sms (one wave at MINB=2, two CTAs per SM): the
     #     BLK=512/VPT=4 register kernel beats the main slab by 1.3-1.7x
-    #     (8.0 -> 4.9 us at b=256, n=8192, K=512). Beyond one wave the two
-    #     trade places by shape (b=512: -13%..+12%), so the slab keeps it.
+    #     (8.0 -> 4.9 us at b=256, n=8192, K=512; Rubin 8192 x 400: 6.3 ->
+    #     4.2 us). A second wave (b <= 4*sms) still wins for the upper half
+    #     of the band (n >= 6144: B200 b=512 +11..12%, and the ragged-row
+    #     6144-8192 x 320-400 cells where the slab lost to sglang by up to
+    #     22%) but loses for the lower half (n=4160, b=512: -13%), so the
+    #     slab keeps that.
     # Reported upstream for adoption as DKG issue #61.
     if n4 <= 2048:
         if wide:
             return _reg(1024, 2, 1, 2 * NB)
-        if b <= 296:
+        if b <= 2 * sms or (b <= 4 * sms and n4 >= 1536):
             return _reg(512, 4, 2, NB)
     if n4 <= 4096 and wide:
         return _reg(1024, 4, 1, 2 * NB)
@@ -524,11 +550,11 @@ _DYN_RT = {
 _DYN_SMEM = ("reg", "regimg")  # smem depends on CMP/IMGW -> recomputed per n
 
 
-def route_static(b: int, n: int, npad: int, k: int) -> dict[str, object]:
+def route_static(b: int, n: int, npad: int, k: int, sms: int = 148) -> dict[str, object]:
     """route() with the n-continuous fields redacted (see _DYN_RT/_DYN_SMEM).
     Constant on maximal n-intervals ("bands"); every redacted field is
     reconstructible from (static, n) by route_dynamic."""
-    plan = route(b, n, npad, k)
+    plan = route(b, n, npad, k, sms=sms)
     st = {key: (dict(val) if isinstance(val, dict) else val) for key, val in plan.items()}
     for f in _DYN_RT[st["kernel"]]:
         st["rt"].pop(f)
@@ -614,10 +640,10 @@ def route_dynamic(static: dict[str, object], n: int) -> tuple[dict[str, object],
     )
 
 
-def route_split(b: int, n: int, npad: int, k: int) -> dict[str, object]:
+def route_split(b: int, n: int, npad: int, k: int, sms: int = 148) -> dict[str, object]:
     """route_static + route_dynamic recombined — must equal route() exactly
     (the factorization fuzz in the unit tests asserts this)."""
-    st = route_static(b, n, npad, k)
+    st = route_static(b, n, npad, k, sms=sms)
     dyn, smem = route_dynamic(st, n)
     plan = {key: (dict(val) if isinstance(val, dict) else val) for key, val in st.items()}
     plan["rt"].update(dyn)
@@ -763,7 +789,8 @@ def _varlen_launcher(num_rows, npad, k, n_env, next_n, cr):
     the universally correct fallback; specialist family tiers below.  Every
     choice here is a function of capture-stable quantities only — mirroring
     the in-tree runner's pick_tuning(graph_capture=...) discipline."""
-    key = (num_rows, npad, k, n_env, next_n, cr, _arch_token())
+    sms = _sm_count()
+    key = (num_rows, npad, k, n_env, next_n, cr, _arch_token(), sms)
     hit = _VARLEN_CACHE.get(key)
     if hit is not None:
         return hit
@@ -784,7 +811,7 @@ def _varlen_launcher(num_rows, npad, k, n_env, next_n, cr):
     # admission window (n4 <= 32768) fits capture-frozen envelopes. The
     # choice is a pure function of this cache key, so CUDA-graph replay
     # safety is unchanged; per-row n / short-row handling lives in-kernel.
-    plan_free = route(num_rows, n_route, npad, k)
+    plan_free = route(num_rows, n_route, npad, k, sms=sms)
     if plan_free["kernel"] == "reg_clus":
         fn = dev.get_compiled__regclus(
             tuple(plan_free["tpl"]), varlen=True, next_n=next_n, cr_shift=cr_shift
@@ -1116,7 +1143,7 @@ _GVR_MAX_DEV = GVR_MAX_DEV
 # per-family launcher builders (cold path: once per distinct shape key)
 # ---------------------------------------------------------------------------
 def _build_launcher(b, n, npad, k):
-    rd = route(b, n, npad, k)
+    rd = route(b, n, npad, k, sms=_sm_count())
     fam = rd["kernel"]
     tpl = tuple(rd["tpl"])
     rt = rd["rt"]
@@ -1294,7 +1321,7 @@ def _run_impl(logits, pre_idx, n_valid, indices, ws, values=None):
                 values[:, n:] = torch.finfo(_F32).min  # -FLT_MAX pad
         return
 
-    key = (b, n, npad, k, _arch_token())
+    key = (b, n, npad, k, _arch_token(), _sm_count())
     lc = _LAUNCH_CACHE.get(key)
     if lc is None:
         lc = _build_launcher(b, n, npad, k)
@@ -1566,7 +1593,7 @@ def run_varlen(
             # R increment (bounded plans, bounded _VARLEN_CACHE)
             n_env = 1 << max(n_env - 1, 1).bit_length()
         n_env = min(max(n_env, 1), npad)
-        key = (num_rows, npad, k, n_env, nn, cr, _arch_token())
+        key = (num_rows, npad, k, n_env, nn, cr, _arch_token(), _sm_count())
         lc = _VARLEN_CACHE.get(key)
         if lc is None:
             if _is_capturing():
@@ -1710,7 +1737,9 @@ def warmup_varlen(
     r = nn
     r_max = req_rows[-1]
     while r <= r_max:
-        plan_free = route(r, max(min(n_env_c, npad_c), int(top_k) + 1), npad_c, int(top_k))
+        plan_free = route(
+            r, max(min(n_env_c, npad_c), int(top_k) + 1), npad_c, int(top_k), sms=_sm_count()
+        )
         if plan_free["kernel"] == "reg_clus":
             ekey = ("reg_clus", tuple(plan_free["tpl"]))
         elif plan_free["kernel"] in ("reg", "regimg"):
--- a/flashinfer/topk_varlen/kernels/gvr2_topk_host.py
+++ b/flashinfer/topk_varlen/kernels/gvr2_topk_host.py
@@ -790,6 +790,44 @@ def route_streaming(
 
 _VARLEN_CACHE = {}
 
+# ---------------------------------------------------------------------------
+# first-call gate for compiled DSL kernel objects
+# ---------------------------------------------------------------------------
+# libcute_dsl_runtime's ``cuda_dialect_init_library_once`` (the per-kernel
+# CUDA init that runs inside the FIRST invocation of a compiled kernel) is a
+# double-checked once-init behind ONE process-wide spinlock. In
+# nvidia-cutlass-dsl 4.6.3, 4.7.0 and 4.7.1 the thread that acquires the lock
+# after another thread already initialised the same kernel returns WITHOUT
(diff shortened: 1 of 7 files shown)


dhiraj113:
@flashinfer-bot run

flashinfer/topk_varlen/topk_varlen.py
```diff
@@ -392,8 +401,14 @@ def _top_k_varlen_heuristic(
...
     cutlass_first = fp32 and n_cols >= 65536 and elems >= (1 << 23)
+    # the one measured cell where hint-free radix_filter beats gvr_2 (rule 1);
+    # hinted gvr_2 is still ahead there, so the carve-out is hint-free only
+    hinted = (
+        pre_idx is not None and _hint_problem(pre_idx, logits, seq_lens, top_k) is None
+    )
+    gvr2_tiny_loss = (not hinted) and top_k >= 2048 and n_cols <= 4096 and batch == 1
 
-    order = ["gvr_2"]
+    order = [] if gvr2_tiny_loss else ["gvr_2"]
```
leejnau, line 415: This newly routes hint-free auto calls to GVR2, including SPLIT plans, while workspace=None still shares one mutable slab per device. The existing xfail in test_topk_varlen_serving.py:225-234 records 17/5120 corrupted rows under concurrent graph replay. Previously a missing hint excluded GVR2, so this expands the known race to ordinary default calls and is unsafe for multi-stream serving integrations. Please either make the default GVR2 workspace stream/graph-scoped or keep workspace-using GVR2 plans out of auto unless a private workspace["gvr2_workspace"] is supplied.

flashinfer/topk_varlen/kernels/gvr2_topk_host.py
```diff
@@ -1024,6 +1080,42 @@ def _reset_for_tests() -> None:
...
+            raise RuntimeError(
+                f"hint-free gvr_2: no pre_idx table for batch={batch}, k={k} on "
+                f"{device}; run one eager call (or warmup_varlen) at this batch "
+                "before CUDA-graph capture"
+            )
+        cap = max(batch, 64 if t is None else 2 * t.shape[0])
+        new = torch.arange(k, dtype=_I32, device=device).unsqueeze(0).expand(cap, k).contiguous()
+        if t is not None:
+            _HINT_FREE_KEEP.append(t)
+        _HINT_FREE[key] = t = new
```
leejnau: This cache growth is neither host-thread-safe nor CUDA-stream-ordered. Concurrent growers can both read the old entry and publish out of order, allowing a smaller table to overwrite a larger one; a later capture can then fail despite an eager warmup at that batch size. The new table is also published immediately after its asynchronous arange/contiguous initialization, so another stream can consume it without waiting for the producer stream. Please serialize growth with a locked recheck and monotonic capacity, retain the displaced table, and establish producer-to-consumer CUDA event ordering before reuse. A concurrent cold-start/growth regression would cover both cases.

flashinfer/topk_varlen/topk_varlen.py
```diff
@@ -1511,9 +1538,10 @@ def top_k_varlen(
...
-                              Rubin sm_107); requires
-                              ``pre_idx`` (hints steer sampling, never
-                              exactness) and fp32 logits;
+                              Rubin sm_107); fp32 logits; ``pre_idx``
+                              optional (hints steer sampling, never
+                              exactness — without one the host supplies a
+                              cached ``arange`` anchor);
                               ``top_k`` in {512, 1024, 2048}. ``load_balance``
                               is ignored (the kernel families load-balance
                               internally). CUDA graphs: warm up each
```
leejnau, line 1601: This warmup contract is incomplete now that hint-free mode has its own lazily sized table. An eager call at this exact geometry with pre_idx does not populate _HINT_FREE, so capturing the same geometry with pre_idx=None still raises; the new test at test_topk_varlen_gvr2.py:1258-1277 demonstrates that sequence. Please either document hint mode as an additional warmup dimension—specifically, that hint-free capture requires a hint-free eager warmup—or expose a supported public warmup path that sizes the synthetic table. As written, “one eager call” per listed geometry is not sufficient.

flashinfer/topk_varlen/kernels/gvr2_topk_host.py
```diff
@@ -152,26 +171,41 @@ def _on_other_device(logits: torch.Tensor) -> bool:
...
-    # per-arch is a perf follow-up, not a functional requirement.
-    wide = b <= 148
+    # tests, split factors, cluster sizing), never correctness. FlashInfer-
+    # local: the REGISTER-band tests use the real SM count `sms` (B300 160,
+    # Rubin 208): rows in (148, sms] are one wave of 1024-thread CTAs there
+    # and the register kernels beat the slab by 1.4-1.8x (B300 12288 x 160:
+    # 7.2 -> 4.6 us; Rubin 12288 x 160-192: 5.6 -> 4.0 us; measured same-node
+    # vs sglang, DKG #61). The STREAMING constants below stay at 148: scaling
+    # them too made the 131072 x 192-256 slab cells 3-28% slower on both parts.
+    wide = b <= sms
```
leejnau, line 195: This makes only part of the register-resident routing SM-aware. The reg_clus branch below still computes availability from the fixed B200 value 148 at line 271, so higher-SM parts can unnecessarily fall through to main, while lower-SM parts can overestimate residency. Please either propagate sms into that decision and add non-148 routing coverage, or document and benchmark why the clustered-register path intentionally remains B200-tuned.

flashinfer/topk_varlen/kernels/gvr2_topk_host.py
```diff
@@ -495,11 +550,11 @@ def _main(BLK, MINB, U, SPLIT):
 _DYN_SMEM = ("reg", "regimg")  # smem depends on CMP/IMGW -> recomputed per n
 
 
-def route_static(b: int, n: int, npad: int, k: int) -> dict[str, object]:
+def route_static(b: int, n: int, npad: int, k: int, sms: int = 148) -> dict[str, object]:
```
leejnau, line 560: route_static() now accepts sms, but the exported route_bands() helper still calls it without that argument, silently enumerating bands using the 148-SM default. On B300/Rubin this can report bands that differ from the runtime launch plan. Please propagate an sms argument through route_bands() and add a non-148 test, or explicitly document the helper as B200-only.

tests/topk_varlen/test_topk_varlen_gvr2.py
```diff
@@ -1145,3 +1184,221 @@ def test_gvr2_route_pure_and_total():
...
+
+@requires_gvr2
+@pytest.mark.parametrize(
+    "kv,next_n,cr,top_k",
+    [
+        ([8192, 5000, 300, 1], 1, 1, 512),  # reg VPT=2 rung + short rows
+        ([8192 * 4, 6000 * 4, 2048 * 4], 2, 4, 512),  # MTP rows, cr=4
+        ([65536, 40000, 1000], 1, 1, 1024),  # reg_clus band
+        ([262144, 100000, 2047], 1, 1, 2048),  # main/clus band, K=2048
+        ([4096] * 200 + [100] * 56, 1, 1, 512),  # b > 148 BLK=512 rung
```
leejnau: This case does not exercise the claimed new BLK512/VPT4 rung: N=4096 gives n4=1024 and returns through the pre-existing n4 <= 1024 branch. Please use N=8192 or another width in (4096, 8192] so the hint-free exactness test actually covers the new b > sms rung.

flashinfer/topk_varlen/topk_varlen.py
```diff
@@ -45,8 +45,9 @@
                        accepted and ignored (the kernel takes no hint).
 ``"auto"``           — shape/dtype-aware ranking that tracks the measured
                        per-config winner (see ``_top_k_varlen_heuristic``):
-                       gvr_2 for hinted fp32; radix_filter for hint-free
-                       fp32 from 32K columns up and for the mid/large bf16 and
+                       gvr_2 for fp32, hinted or hint-free (one tiny
```
leejnau, line 51: This correctly describes hint-free GVR2 routing, but the backend table above still says that GVR2 requires pre_idx at line 36. Please update that entry to say the hint is optional and that a cached synthetic anchor is used when absent, consistent with the detailed public docstring.

flashinfer/topk_varlen/topk_varlen.py
dhiraj113: Agreed, and the exposure was wider than hint-free: hinted `auto` already reached the same slab-using plans since #4811. Fixed at the host: the default slab is now one per (device, CUDA stream), created only by an eager launch (a capture on a stream that never ran one raises instead of allocating from the graph pool), and it is resolved lazily so only the streaming `main` plans ever touch it — register/cluster plans capture on any stream as before. Graphs captured on different streams therefore never share scratch; the xfail in test_topk_varlen_serving.py is now the positive test `test_gvr2_default_workspace_per_stream_across_concurrent_graphs` (4 streams x 8 SPLIT launches, 0 corrupted rows), plus a test that a slab-using capture on a fresh stream raises and then succeeds after one eager launch there. The remaining rule, stated in the docstring: graphs captured on the SAME stream share that stream's slab and must not be replayed concurrently with each other (pass a private `gvr2_workspace` for that). I preferred this over keeping slab-using plans out of hint-free auto, which would have dropped gvr_2 from auto for every width above ~16K where it wins 1.5-2.9x.

(Fixed in `a11d8a74`.)

flashinfer/topk_varlen/kernels/gvr2_topk_host.py
dhiraj113: Right on both counts. Growth now takes a module lock with a re-check, capacity is `max(batch, 2*old)` so it never shrinks, the displaced table stays alive, and the producing stream is synchronized before the new table is published (eager-only and rare, so a host sync is the simplest correct ordering). New test: eight threads on eight streams growing the same (device, k) table from cold behind a barrier; asserts the published capacity is the maximum requested, every row reads arange(k) from every stream, and hint-free calls on each stream are exact afterwards. It fails on the previous tree.

(Fixed in `a11d8a74`.)

flashinfer/topk_varlen/topk_varlen.py
dhiraj113: Agreed. Rather than documenting hint mode as a sixth warm-up dimension I removed it: a hinted eager call at a geometry now also sizes the anchor table for that batch and k (dict hit once sized, allocation only on growth, never under capture), so "one eager call per geometry" holds again whichever way the capture runs. The graph test now does exactly the sequence you described — hinted eager call at the larger batch, then hint-free capture — and it succeeds; growth is still refused under capture for a batch no eager call has seen. The docstring's CUDA-graph bullet now lists the three things the eager call creates (launcher, anchor table, this stream's slab) and that it must run on the capturing stream.

(Fixed in `a11d8a74`.)

flashinfer/topk_varlen/kernels/gvr2_topk_host.py
dhiraj113: You were right and I had it wrong: I assumed the cs=8 veto made `av` inert above 148 SMs, but enumerating the plan space showed 1200 differing plans on Rubin and 240 on B300. Measured same-node, exact everywhere: Rubin gets 4-CTA instead of 2-CTA clusters at b 38-48 (8-12%) and reg_clus instead of main/clus at b 80-104 and at 48K-64K (up to 35%); B300 12-34%. On DRIVE P2021 (68 SMs) the SM-aware rule wins 34/44 cells but loses 10 by up to 40% (it falls to clus/main at 48K-128K), so the admission is `max(sms, 148) // b`: SM-aware at or above 148, upstream's value below. Inline comment carries the numbers; `test_gvr2_route_reg_clus_sm_count_aware` pins cluster size and admission for 158/160/208 and the unchanged low-SM plans.

(Fixed in `a11d8a74`.)

flashinfer/topk_varlen/kernels/gvr2_topk_host.py
dhiraj113: Done: `route_bands(..., sms=148)` passes it through to route_static, with a docstring note that the caller should pass the target part's SM count. Test: at b=160 the bands for 148 vs 208 differ exactly where route() does (the 148-SM slab band starts at n=8193, the 208-SM bands stay on the register kernels), and every band's plan equals route_static at both endpoints.

(Fixed in `a11d8a74`.)

tests/topk_varlen/test_topk_varlen_gvr2.py
dhiraj113: Correct, N=4096 lands on the pre-existing (512, 2, 4) tuple on every SM count. The case now uses width 8192 for the 200 long rows and asserts through route() that the exercised plan is the (512, 4, 2) rung.

(Fixed in `a11d8a74`.)

flashinfer/topk_varlen/topk_varlen.py
dhiraj113: Updated: the gvr_2 entry now says `pre_idx` is optional and a cached synthetic anchor is used when it is absent, exact either way.

(Fixed in `a11d8a74`.)

dhiraj113:
@flashinfer-bot run

leejnau (approved):

dhiraj113:
Added `tests/topk_varlen/test_topk_varlen_concurrency.py` (`4317ef5f`): an adversarial concurrency suite for the defect class this review found — eager launches overlapping across threads/streams, four graphs replayed back to back against eager work, concurrent captures from threads, same-stream graph sharing, anchor-table growth/keep-alive/cross-stream visibility, and the warm-up-on-capturing-stream rule. Every test runs under a calibrated deadline that dumps thread stacks and fails on a hang. Against the pre-review host the four-graph stress test corrupts rows in 4/5 runs and the fresh-stream capture test fails deterministically. 10 passed on B100, DRIVE P2021, Rubin, computelab B200/B300; skips elsewhere.

dhiraj113:
@flashinfer-bot run

tests/topk_varlen/test_topk_varlen_concurrency.py
```diff
@@ -0,0 +1,579 @@
...
+    def capture(i):
+        try:
+            g = torch.cuda.CUDAGraph()
+            barrier.wait()
+            # thread-local capture mode: other threads' CUDA calls must not
+            # invalidate this thread's capture (the default "global" mode does)
+            with (
+                torch.cuda.stream(streams[i]),
+                torch.cuda.graph(
+                    g, stream=streams[i], capture_error_mode="thread_local"
```
leejnau: capture_error_mode="thread_local" only controls which threads are checked for capture-unsafe API calls; it does not permit multiple captures in one process. PyTorch explicitly requires that only one capture be underway per process and warns that violations may cause silent numerical errors or undefined behavior. This test starts four captures concurrently, so its result is not valid. Please serialize capture—as SGLang does—and retain concurrent replay for workspace-race coverage, or isolate captures in separate processes/contexts.

https://docs.pytorch.org/docs/main/notes/cuda.html#constraints

tests/topk_varlen/test_topk_varlen_concurrency.py
```diff
@@ -0,0 +1,579 @@
...
+    other = torch.cuda.Stream()
+    for b in (100, 300, 1024):
+        with torch.cuda.stream(other):
+            _launch((big_logits[:b], big_seq[:b], None, big_out[:b]), k, hint=False)
+    torch.cuda.synchronize()
+    assert _host._HINT_FREE[key] is not old and _host._HINT_FREE[key].shape[0] >= 1024
+    assert old in _host._HINT_FREE_KEEP
+    assert bool((old == torch.arange(k, dtype=torch.int32, device=_DEV)).all())
+    for _ in range(3):
+        out.fill_(-7)
```
leejnau: out.fill_(-7) is enqueued on the default stream, while g.replay() writes the same buffer on s without any dependency between them. The fill can therefore execute after the replay and overwrite a correct result, causing a false failure. Please enqueue the fill on s immediately before g.replay(), or make s wait for the default stream before replaying.

tests/topk_varlen/test_topk_varlen_concurrency.py
```diff
@@ -0,0 +1,579 @@
...
+            def body():
+                try:
+                    fn(*args, **kwargs)
+                except BaseException as e:  # noqa: BLE001
+                    result["error"] = e
+
+            t = threading.Thread(target=body, name=f"{fn.__name__}-body", daemon=True)
+            t.start()
+            t.join(budget)
+            if t.is_alive():
```
leejnau: This does not actually enforce the promised deadline. If t remains alive, pytest.fail() exits only the wrapper; the daemon body and its nested CUDA threads continue running, potentially holding locks and contaminating subsequent tests. _per_launch_seconds() also runs before the monitored thread starts, so calibration can still hang indefinitely, and moving the body into a fresh host thread loses the caller’s thread-local CUDA device state. Please use the repository’s timeout support in a mode that terminates the test process, or isolate the body in a killable subprocess rather than continuing pytest with live CUDA work.

tests/topk_varlen/test_topk_varlen_concurrency.py
```diff
@@ -0,0 +1,579 @@
...
+                grown.set()
+                _launch((logits, seq, None, outs[0]), k, hint=False)
+        except Exception as e:  # noqa: BLE001
+            errors.append(e)
+
+    def consumer():
+        try:
+            grown.wait()
+            with torch.cuda.stream(b):
+                _launch((logits, seq, None, outs[1]), k, hint=False)  # no sync with A
```
leejnau: This does not validate anchor-table publication ordering. GVR2 remains exact for arbitrary hint contents, and invalid indices are bounds-filtered, so stream B can consume a partially initialized table while both final top-k checks still pass. Please snapshot and compare the table contents to arange(k) from stream B. Ideally, add a producer-stream delay or gate before table initialization so the pre-fix publish-before-completion behavior fails deterministically.

tests/topk_varlen/test_topk_varlen_concurrency.py
```diff
@@ -0,0 +1,579 @@
...
+@_deadline(launches=8)
+def test_gvr2_warmup_varlen_on_stream_enables_capture_on_that_stream_only():
+    """warmup_varlen run on stream A creates A's slab and sizes the anchor
+    table; a hint-free slab-using capture on A then succeeds, while the same
+    capture on a stream that never saw an eager launch raises (never allocates
+    from the graph pool)."""
+    k, work = _split_inputs(1, seed=71)
+    logits, seq, _, out = work[0]
+    rows, n = logits.shape
+    a, b = torch.cuda.Stream(), torch.cuda.Stream()
```
leejnau, line 521: torch.cuda.Stream() does not guarantee a fresh raw stream: PyTorch returns one of 32 pooled streams round-robin. Because _ws_keep persists by raw handle, b may already have a cached slab from an earlier test, making the expected capture failure order-dependent. The following register-family test has the converse problem: a cached slab can let capture succeed even if that path incorrectly consults the workspace. Please synchronize and reset or explicitly clear/check the relevant cache keys before establishing these preconditions, and assert that register capture does not populate a slab.

https://github.com/pytorch/pytorch/blob/main/c10/cuda/CUDAStream.h#L22-L31

leejnau (changes requested):

tests/topk_varlen/test_topk_varlen_concurrency.py
dhiraj113: Right — I had read the error that "global" mode raised as the thing to silence rather than the constraint it enforces. The graphs are now captured one at a time, each on its own stream after its own eager warm-up, and only the replays run concurrently (renamed `test_gvr2_graphs_captured_sequentially_replay_concurrently_exact`). The file's docstring lists this as a ground rule so it does not creep back.

(Fixed in `aaa9f18e`.)

tests/topk_varlen/test_topk_varlen_concurrency.py
dhiraj113: Fixed: the fill is issued on `s` right before `g.replay()`. I audited the other tests for the same pattern — every other fill is either on the replaying stream or followed by a device synchronize before the replay — and added the rule to the file docstring.

(Fixed in `aaa9f18e`.)

tests/topk_varlen/test_topk_varlen_concurrency.py
dhiraj113: Agreed on all three points; the decorator is gone. The file now uses the repository's convention, `pytest.mark.timeout(300, method="thread")` on every test (pytest-timeout is in requirements-test.txt and tests/attention already carries the mark), which dumps all thread stacks and terminates the process — verified with an injected non-returning body, killed after 6 s. 300 s is a ~50x margin over the slowest test on the slowest part we have. Barrier timeouts remain as the finer guard inside the thread tests.

(Fixed in `aaa9f18e`.)

tests/topk_varlen/test_topk_varlen_concurrency.py
dhiraj113: Correct, and it explains why the test passed against the pre-review host in my teeth check. It now measures the table itself: the consumer on stream B snapshots the published table without any synchronization with A and compares every row to `arange(k)`. Making the pre-fix behaviour fail deterministically took three details, all measured: (1) the ~2 s busy-wait has to be queued on A between the `arange` and the `expand().contiguous()` fill (queuing it before the growth does not work — `torch.arange` itself blocked the host until the stream drained); (2) the growth's exact allocation path has to be walked once on stream A beforehand, because a `cudaMalloc` inside the growth blocks the host behind the busy-wait and closes the window even on a publish-early host; (3) that warm block is left holding -1 so recycled memory cannot hold the right answer by accident. With that, the pre-review host fails 512/512 rows on every run and the current host passes.

(Fixed in `aaa9f18e`.)

tests/topk_varlen/test_topk_varlen_concurrency.py
dhiraj113: Right. Added `_forget_slab(stream)` (synchronize, then drop the entry for the stream's raw handle under the workspace lock) and use it before the must-raise captures here and in the round-1 serving test — which, as you predicted, was already order-dependent: it failed with DID NOT RAISE when the two files ran in the other order. The register-family test now clears the slab for its stream before capture and asserts afterwards that the capture did not create one. The API docstring also states the consequence for production: default slabs are bounded at 32 x 21 MB per device, and two Stream objects that share a pooled raw stream share a slab and are stream-ordered.

(Fixed in `aaa9f18e`.)

dhiraj113:
@flashinfer-bot run

flashinfer/topk_varlen/kernels/gvr2_topk_host.py
```diff
@@ -904,7 +974,18 @@ def route_bands(
...
+# buffer, restored by the kernel), so two launches that can overlap in time —
+# concurrent streams, or graphs captured on different streams and replayed
+# together — must not share it (measured 17/5120 corrupted rows through a
+# device-wide slab; see test_topk_varlen_serving). Keyed by the raw stream
+# handle; the tensor refcount is the keep-alive.
+_ws_keep = {}  # (device index, cuda stream handle) -> keep-alive int32 view
+
+
+def _ws_key(d: int) -> tuple[int, int]:
+    return (d, torch.cuda.current_stream(d).cuda_stream)
```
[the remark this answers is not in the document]
dhiraj113: Documented rather than adding a cleanup API: the API docstring now states the bound (slabs are keyed by the raw stream handle and PyTorch pools 32 streams per priority, so at most 32 x 21 MB per device), and `_reset_for_tests()` remains the test-only way to drop them. A public cleanup seam is an execution-strategy knob we keep internal.

tests/topk_varlen/test_topk_varlen_serving.py
```diff
@@ -167,8 +167,11 @@ def launch(w, out):
             workspace={"gvr2_workspace": ws} if private_workspace else None,
         )
 
-    for w in work:  # compile eagerly before capture
-        launch(w, w[3][0])
+    # compile eagerly before capture, on the stream that will capture: the
+    # eager call also creates that stream's default workspace slab
+    for s, w in enumerate(work):
+        with torch.cuda.stream(cuda_streams[s]):
+            launch(w, w[3][0])
```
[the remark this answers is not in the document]
dhiraj113: Applied in `718974c9`: each side stream waits on the default stream before its eager warm-up.

tests/topk_varlen/test_topk_varlen_concurrency.py
```diff
@@ -0,0 +1,579 @@
...
+    work = []
+    for _ in range(copies):
+        logits = torch.randn(rows, n, generator=gen, device=_DEV)
+        seq = torch.randint(
+            40000, n, (rows,), generator=gen, device=_DEV, dtype=torch.int32
+        )
+        pre = torch.full((rows, k), -1, dtype=torch.int32, device=_DEV)
+        out = torch.full((rows, k), -7, dtype=torch.int32, device=_DEV)
+        work.append((logits, seq, pre, out))
+    return k, work
```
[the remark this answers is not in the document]
dhiraj113: Applied in `718974c9`: `_split_inputs` synchronizes before returning, and every site that creates inputs or -7 output buffers on the default stream synchronizes before the first side-stream consumer.

dhiraj113:
@flashinfer-bot run

flashinfer/topk_varlen/kernels/gvr2_topk_host.py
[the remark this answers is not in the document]
leejnau: agree with coderabbitai we should probably address this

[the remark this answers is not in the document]
dhiraj113: Done in `f5fe8836`: `flashinfer.topk_varlen.release_gvr2_resources(device=None)` synchronizes the device, drops every default slab (all streams) and every hint-free anchor table (current and superseded) for it, and returns the bytes released to the caching allocator. The docstrings state the invalidation rule — a graph captured against a released slab/table must be re-captured after one eager launch on its stream, the same warm-up rule as the first capture — and that caller-provided `workspace=` buffers are untouched. Test covers: entries dropped, `memory_allocated` falls by at least the released bytes, a slab-using capture right after release raises the documented error instead of allocating, the next eager launch recreates both and is exact, a second release returns 0. No weak map keyed by `torch.cuda.Stream` and no automatic eviction (either could invalidate a live graph behind the caller's back).

dhiraj113:
@flashinfer-bot run

flashinfer/topk_varlen/kernels/gvr2_topk_host.py
```diff
@@ -982,6 +1074,37 @@ def _reset_for_tests() -> None:
...
+    hands it back to the driver). Returns the number of bytes released.
+
+    INVALIDATION: a CUDA graph captured against a released slab or table
+    replays on freed memory. Callers release only when no such graph will be
+    replayed again, and re-capture after a fresh eager warm-up on the
+    capturing stream (the same rule as the first capture)."""
+    d = torch.cuda.current_device() if device is None else torch.device(device).index
+    if d is None:
+        d = torch.cuda.current_device()
+    torch.cuda.synchronize(d)
```
leejnau: torch.cuda.synchronize(d) only waits for work submitted before the synchronization; it does not exclude concurrent host callers. The workspace and hint-table hot paths read their dictionaries without these locks, and the two caches are subsequently cleared under separate locks. Another thread can therefore obtain or recreate a resource after the synchronization, allowing release to race an asynchronous launch or return with a repopulated cache. Please make release mutually exclusive with GVR2 lookup/launch enqueue, or explicitly require and document complete caller quiescence—including no concurrent eager calls, not only no graph replays. Device synchronization alone is insufficient.

flashinfer/topk_varlen/kernels/gvr2_topk_host.py
```diff
@@ -982,6 +1074,37 @@ def _reset_for_tests() -> None:
...
+    with _mu:
+        for key in [k for k in _ws_keep if k[0] == d]:
+            freed += _ws_keep.pop(key).numel() * 4
+    with _HINT_FREE_LOCK:
+        for key in [k for k in _HINT_FREE if k[0] == d]:
+            freed += _HINT_FREE.pop(key).numel() * 4
+        keep = [t for t in _HINT_FREE_KEEP if t.device.index == d]
+        for t in keep:
+            freed += t.numel() * 4
+            _HINT_FREE_KEEP.remove(t)
```
leejnau: list.remove(t) uses tensor equality, not identity. If another retained tensor precedes t—especially an entry from another device—this can raise a cross-device or ambiguous-boolean error and leave cleanup partially completed. This is the same tensor-membership bug fixed earlier in this PR. Please partition/rebuild _HINT_FREE_KEEP by device.index without invoking tensor equality, and add an interleaved multi-device regression.

flashinfer/topk_varlen/kernels/gvr2_topk_host.py
```diff
@@ -982,6 +1074,37 @@ def _reset_for_tests() -> None:
...
+    (default: the current device). Synchronizes the device first so no
+    in-flight launch still uses them, then drops the references; the memory
+    returns to the torch caching allocator (``torch.cuda.empty_cache()``
+    hands it back to the driver). Returns the number of bytes released.
+
+    INVALIDATION: a CUDA graph captured against a released slab or table
+    replays on freed memory. Callers release only when no such graph will be
+    replayed again, and re-capture after a fresh eager warm-up on the
+    capturing stream (the same rule as the first capture)."""
+    d = torch.cuda.current_device() if device is None else torch.device(device).index
```
leejnau: This discards the device type and retains only .index. Consequently, "cpu" silently releases the current CUDA device, while "cpu:1" can release CUDA device 1. An invalid device argument should never redirect cleanup to unrelated CUDA state. Please reject non-CUDA devices before deriving the index—while preserving the intended None/integer inputs—and add coverage for both indexed and unindexed non-CUDA devices.

tests/topk_varlen/test_topk_varlen_concurrency.py
```diff
@@ -0,0 +1,579 @@
...
+    _FLASHINFER_AVAILABLE = False
+
+pytestmark = [
+    pytest.mark.skipif(
+        not (_FLASHINFER_AVAILABLE and torch.cuda.is_available()),
+        reason="flashinfer + CUDA required",
+    ),
+    # > cold-cache JIT compile + the longest test by 50x; a real hang is
+    # immediate. "thread" mode dumps all stacks and terminates the process.
+    pytest.mark.timeout(300, method="thread"),
```
leejnau, line 67: This 300-second timeout is being hit repeatedly inside the concurrency suite. The latest targeted pipeline has non-allowed failures on GB200/cu129 and GB300/cu129/cu130, and the preceding pipeline failed at the same kernel-launch locations. This appears repeatable rather than ordinary timeout noise. Please diagnose the underlying kernel or suite-state hang and rerun the targeted matrix cleanly before merge.

tests/topk_varlen/test_topk_varlen_gvr2.py
```diff
@@ -1145,3 +1184,424 @@ def test_gvr2_route_pure_and_total():
...
+    assert not any(t.device.index == dev for t in _host._HINT_FREE_KEEP)
+    assert before - torch.cuda.memory_allocated(dev) >= freed - 1024, (
+        before,
+        torch.cuda.memory_allocated(dev),
+        freed,
+    )
+    # capture without a fresh warm-up on s: loud, not a silent allocation
+    g = torch.cuda.CUDAGraph()
+    with (
+        pytest.raises(RuntimeError, match="no slab for this stream|hint-free gvr_2"),
```
leejnau: This assertion can pass without exercising the workspace guard. Because the launch is hint-free, the missing anchor table is checked before default-workspace resolution, and this regex explicitly accepts that earlier error. The test therefore does not prove that a released slab cannot be allocated during capture. Please provide a valid pre_idx (or restore only the anchor table), require the specific "no slab for this stream" error, and assert that the workspace cache remains empty.

flashinfer/topk_varlen/kernels/gvr2_topk_host.py
leejnau: This workspace documentation is stale. The default cache now keeps one slab per (device, raw stream handle), not one slab per device, and concurrent launches on distinct streams no longer require caller-provided workspaces. Please update this section accordingly. Please also avoid describing 32 slabs as a hard per-device bound elsewhere: the pool is per priority, and external stream handles are not bounded by that pool.

flashinfer/topk_varlen/topk_varlen.py
```diff
@@ -1070,10 +1096,33 @@ def _run_gvr2(
...
+    returns the number of bytes released to the torch caching allocator
+    (``torch.cuda.empty_cache()`` returns that memory to the driver).
+
+    Any CUDA graph captured against a released slab or table would replay on
+    freed memory: release only when no such graph will be replayed again, and
+    re-capture after one eager ``gvr_2`` launch on the capturing stream, the
+    same warm-up rule as the first capture. Explicit ``workspace=`` buffers
+    passed by the caller are never touched.
+    """
+    from .kernels import gvr2_topk_host
```
leejnau: flashinfer/topk_varlen/topk_varlen.py:1121 unconditionally imports .kernels, whose __init__.py eagerly imports optional CUTLASS DSL modules. On installations without nvidia-cutlass-dsl, release_gvr2_resources() raises instead of returning 0, even though no GVR2 cache can exist.

flashinfer/topk_varlen/kernels/gvr2_topk_host.py
dhiraj113: Agreed that the device sync alone cannot exclude a concurrent caller, and a lock on the launch hot path is not something I want to pay on every call. Done in `68827fce` as the contract option you offered, plus the part that can be made atomic: `release_cached_resources` now holds both cache locks (fixed order `_mu` -> `_HINT_FREE_LOCK`, never nested elsewhere) around the device sync and both clears, so a slow-path creation either completes before the clear or starts after it — nothing half-published, no repopulated cache behind the caller's back from a creator that raced in. The docstrings (host + public wrapper) state caller quiescence explicitly: no gvr_2 call in flight or being issued on the device from any thread, eager or graph replay, for the duration of the call — the same rule as `torch.cuda.empty_cache()` versus live tensors. Test `test_gvr2_release_between_quiescent_multithreaded_phases_is_exact` exercises the contract in a threaded program (eight streams park at a barrier, release, resume): exact results, resources recreated, second release frees the same slabs.

flashinfer/topk_varlen/kernels/gvr2_topk_host.py
dhiraj113: Right — the same class of bug as the `in` check fixed earlier, I should have caught it. `68827fce` partitions `_HINT_FREE_KEEP` by `t.device.index` into keep/drop lists and reassigns in place; no tensor comparison anywhere. Regression `test_gvr2_release_partitions_kept_tables_without_tensor_equality` interleaves superseded tables of two shapes per k (and of two devices when the host has two GPUs) and releases one device at a time: the released device's entries are gone, the other device's count and current tables are untouched, no error.

flashinfer/topk_varlen/kernels/gvr2_topk_host.py
dhiraj113: Fixed in `68827fce`: ints map to `cuda:<i>`, everything else goes through `torch.device` and must be `type == "cuda"`, otherwise `ValueError("expected a CUDA device ...")` before anything is touched; `None` stays "current device". `test_gvr2_release_rejects_non_cuda_device` covers `"cpu"`, `"cpu:1"`, `torch.device("cpu")`, `torch.device("cpu", 1)` and asserts the caches are intact afterwards; the existing release test also exercises int / device / string spellings of the same CUDA device.

tests/topk_varlen/test_topk_varlen_concurrency.py
dhiraj113: Diagnosed; it is a CuTe DSL runtime bug, and the suite is the first thing in the tree that trips it.

**What the traces show.** In all five failed jobs the concurrency batch stops at the FIRST call of a compiled DSL kernel object — a worker thread of test 1 (`main` family) or the main thread of test 5 (`reg` family, tests 1–4 passed in ~15 s) — right after "Loaded cached CuTe-DSL kernel". No kernel is on the GPU; the other three pytest workers on the same GPU keep passing.

**Native evidence.** Reproduced locally in the same four-process layout and attached gdb: the thread is on-CPU at `cuda_dialect_init_library_once+55` in `libcute_dsl_runtime.so`, the `jne` after a `lock cmpxchg` on the runtime's single process-wide once-init spinlock, `rax=1` (held), no other thread anywhere near the runtime, `rip` identical 10 s apart.

**The bug.** Disassembly of the function in five DSL builds: in 4.6.3, 4.7.0 (the CI image), 4.7.1 and nightly 20260810, the path "lock acquired, but another thread already set this kernel's once-flag" returns without the `movq $0, <lock>` release. 4.8.0.dev0 and nightlies >= 20260818 have the release on that path. So the loser of a first-call race leaks the global lock, nothing fails at that moment, and the next first call of ANY other kernel in the process spins forever. Standalone reproducer (8 threads race the first call of kernel A, then first-call kernel B): 8/8 hang; control with a single-thread warm-up of A: 3/3 pass. Filed as DKG #65 with a request for a 4.7.x backport; no public changelog entry mentions it.

**Why here.** Test 1 races eight threads on a cold kernel by design (host thread-safety of the slab path); under sibling-process load the race window is wide; test 5 is the next cold kernel. Matrix: exact CI image on x86 B100 hung 6/6 four-process iterations, Rubin (nightly 20260810) 2–3/6, parley venv (nightly 20260818) 0/6; test 5 alone and the file alone never hang. Host code did not change between the last passing pipeline and the first failing one.

**Fix in `68827fce`.** `_gate_first_call` wraps every compiled kernel object the host obtains from the device module: concurrent first calls of one object are serialized until the first has returned (the once-init is synchronous inside it); afterwards the wrapper is one list-element load per launch. This protects real multi-threaded callers on the current stable DSL releases, not just the tests, and is removable once the floor passes the fix. Teeth: the same CI image with the gate, 3/3 clean four-process iterations. The suite is unchanged — test 1 followed by test 5 is exactly the trigger and is now the regression test. Targeted matrix rerun requested below.

tests/topk_varlen/test_topk_varlen_gvr2.py
dhiraj113: Correct, the hint-free launch tripped the anchor-table guard first and the alternation hid it. `68827fce` captures with a valid `pre_idx` (an arange table built by the test) so only the workspace guard can fire, requires `match="no slab for this stream"` exactly, and asserts afterwards that no slab was allocated for the device; the hint-free case is asserted separately against `"hint-free gvr_2: no pre_idx table"` with the table cache still empty.

dhiraj113:
@flashinfer-bot run

flashinfer/topk_varlen/kernels/gvr2_topk_host.py
dhiraj113: Updated in `20b5ce8e`: the workspace section now describes one slab per (device, raw stream handle) created by the first eager slab-using launch on that stream, states that concurrent streams each own a slab so a caller-provided workspace is optional rather than required for concurrency, keeps the same-stream-graphs rule, and describes the count as the number of distinct handles that ran an eager launch — `torch.cuda.Stream()` recycles a fixed pool per device and priority, but external / other-priority handles add slabs beyond it, so it is not a hard per-device bound. The `top_k_varlen` and `release_gvr2_resources` docstrings say the same instead of "32 x 21 MB per device".

flashinfer/topk_varlen/topk_varlen.py
dhiraj113: Good catch. `20b5ce8e`: `release_gvr2_resources` looks the host module up in `sys.modules` and returns 0 if it was never imported — no gvr_2 launch can have happened, so nothing can be cached — instead of importing `.kernels`; the device-argument check runs before the lookup so `"cpu"` raises identically whether or not the host is loaded. `test_gvr2_release_without_loaded_host_is_a_no_op` runs a fresh interpreter, asserts the host is not imported before or after the call, that it returns 0 for `None` and `0`, and that `"cpu"` raises `ValueError`.

dhiraj113:
@flashinfer-bot run

leejnau (approved):

kahyunnam (approved):
LGTM. One nit would be that some changes (e.g., changes in flashinfer/topk_varlen/kernels/gvr2_topk_host.py) are leaning comment heavy. but its not a blocker.

dhiraj113:
yes, changes in `gvr_topk_host.py` are mostly internal heuristics routing to different kernels, and all this was done by the agent, so leaving the comments as is.

PR #4993, opened 2026-09-06
fix(attention): scope the NVFP4 split-KV gate on sm120/121 to the VO-split plan shape

jethac:
## Summary

Scope the NVFP4 split-KV gate on sm120/sm121 to the plan shape it was introduced for, instead of disabling split-KV for every NVFP4 KV plan on those archs.

`_nvfp4_kv_requires_disabled_split_kv` (mine, #3684, later arch-scoped in #4747) forces `disable_split_kv=True` for any NVFP4 KV `BatchPrefillWithPagedKVCacheWrapper.plan()` on sm120/121. The corruption it works around was only ever observed on the asymmetric VO-split plan (`head_dim_qk=512`, `head_dim_vo=256`, Gemma-4 global layers). For symmetric plans it costs a lot: with split-KV off a decode / spec-decode-verify step streams the whole KV range from one CTA per (batch, head), so decode throughput falls roughly linearly with context.

This PR keeps the gate for `head_dim_qk != head_dim_vo` or `head_dim_qk > 256`, and for plans that do not pass head dims, and lets symmetric plans up to 256 keep split-KV. Both `plan()` call sites (paged and ragged) pass `head_dim_qk`/`head_dim_vo` through.

## Measurements

vLLM (jethac/vllm `nvfp4-fa2-consumer-blackwell`, FA2 NVFP4 KV path), `--kv-cache-dtype nvfp4`, MTP k=3, Qwen3.8-27B-MTP-NVFP4 (head_dim 256), decode tok/s at increasing context; "gate on" is stock 0.6.18, "narrowed" is the same wheel with this `prefill.py`:

| context | RTX 5090 gate on | RTX 5090 narrowed | GB10 gate on | GB10 narrowed |
|---|---|---|---|---|
| 1k | 153.6 | 135.1 | 21.9 | 24.1 |
| 8k | 105.5 | 158.4 | 22.1 | 23.7 |
| 32k | 46.8 | 145.2 | 17.3 | 23.6 |
| 64k | 26.6 | 138.7 | 12.3 | 25.3 |
| 128k | 14.2 | 117.2 | 9.3 | 18.7 |

Needle retrieval 10/10 on every run; MTP acceptance identical on both sides (mean accepted length 3.0-3.7 of 4). The 1k point on the 5090 is ~12% slower with split-KV on; I have not chased it.

Correctness of the newly un-gated shape: Gemma-3-1B with a calibrated 4-bit KV scheme (head 256, split-KV on) retrieves 15/16 needles, identical to bf16 KV on the same prompts. Independently, @TechPrototyper's kernel ladder (qo 1..9 x kv 2k-32k, `head_dim_qk=512` / `head_dim_vo=256`, both V chunks, split-KV forced on) is 72/72 clean at `max_rel` ~0.005 (vllm-project/vllm#46329, comment 5554547616), which suggests the remaining gated shape may also be safe; I have left it gated here since that is the shape the original report came from.

## Test

`tests/attention/test_nvfp4_attention_sm120.py::test_nvfp4_split_kv_gate_shape_scope` fakes the compute capability: symmetric 64/128/256 not gated; 512/256, 512/512, 256/128 gated; no shape -> gated; bf16 never gated; (9,0) never gated. Passes on CPU-only import of the module and on a 5090 against the real device (128/128 and 256/256 allowed, 512/256 and 512/512 disabled).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved NVFP4 attention planning on supported GPU architectures by keeping split-KV enabled for symmetric head dimensions up to 256.
  * Continued disabling split-KV for asymmetric configurations, head dimensions wider than 256, or when shape details are unavailable.

* **Tests**
  * Added coverage for shape-specific split-KV behavior, including architecture and data type safeguards.

### DIFF
--- a/flashinfer/prefill.py
+++ b/flashinfer/prefill.py
@@ -1518,9 +1518,12 @@ _NVFP4_SPLIT_KV_BROKEN_ARCHS = ((12, 0), (12, 1))
 
 
 def _nvfp4_kv_requires_disabled_split_kv(
-    kv_data_type: torch.dtype, device: torch.device
+    kv_data_type: torch.dtype,
+    device: torch.device,
+    head_dim_qk: Optional[int] = None,
+    head_dim_vo: Optional[int] = None,
 ) -> bool:
-    """Whether split-KV must be disabled because the KV cache is NVFP4 on an affected arch.
+    """Whether split-KV must be disabled: NVFP4 KV, on an affected arch, on the VO-split plan.
 
     This gate is an *empirical workaround*: with split-KV (flash-decoding)
     enabled, NVFP4 paged KV was observed to produce corrupted outputs whenever
@@ -1545,10 +1548,39 @@ def _nvfp4_kv_requires_disabled_split_kv(
     boundaries, with the gate-on/gate-off difference staying at output-dtype
     rounding scale. FP8 and 16-bit KV caches were never gated and keep split-KV
     everywhere.
+
+    The gate is further scoped to the plan shape the corruption was actually
+    observed on: the asymmetric VO-split plan (``head_dim_qk != head_dim_vo``,
+    i.e. Gemma-4's 512-wide QK against a 256-wide VO slice) or a head wider
+    than 256. Symmetric plans at ``head_dim <= 256`` on SM120/121 were tested
+    clean with split-KV enabled from several directions (vllm-project/vllm
+    #46329): a kernel-level A/B ladder at head_dim 128 (qo 1..9, kv to 262k,
+    fixed and auto split, ragged batches), a dense-reference comparison and a
+    needle battery served from cached prefix at head_dim 256, paired
+    gate-on/gate-off end-to-end runs on both RTX 5090 (sm120) and GB10 (sm121)
+    with byte-identical greedy retrieval to 262k, and a week of production
+    traffic on a build that predates the gate with the planner confirmed to be
+    splitting for every prefix-cache-extend request. On those shapes the gate
+    only costs: speculative-decode verify batches (``qo_len`` 2..9) are planned
+    through this wrapper and, with split-KV off, stream the whole KV from one
+    CTA row, so decode falls off linearly with context (about 9x at 262k on a
+    5090). The VO-split shape has not been re-tested with split-KV on since the
+    original report and stays gated until it is.
+
+    ``head_dim_qk``/``head_dim_vo`` default to ``None`` for callers that only
+    know the dtype and device; that conservative path keeps the arch-level
+    gate unchanged.
     """
     if not _is_nvfp4_kv_dtype(kv_data_type):
         return False
-    return get_compute_capability(device) in _NVFP4_SPLIT_KV_BROKEN_ARCHS
+    if get_compute_capability(device) not in _NVFP4_SPLIT_KV_BROKEN_ARCHS:
+        return False
+    if head_dim_qk is None:
+        # Shape unknown: keep the arch-level gate.
+        return True
+    if head_dim_vo is None:
+        head_dim_vo = head_dim_qk
+    return head_dim_qk != head_dim_vo or head_dim_qk > 256
 
 
 def _build_block_tables_from_paged_kv_indices(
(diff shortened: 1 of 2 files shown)


TensorRaya:
Tested the shape this PR un-gates on an sm_120 card, kernel level, since #4747 left that as the open item ("needs an affected GPU to be meaningful").

**Setup.** RTX PRO 4000 Blackwell (sm_120, 24 GB), driver 580.173.02, CUDA 13.0, torch 2.13.0+cu130, FlashInfer `main` `83ce2dab` built from source. `BatchPrefillWithPagedKVCacheWrapper`, `backend="fa2"`, HND, causal, 6 QO / 1 KV heads, `head_dim_qk = head_dim_vo = 256`, page 16, NVFP4 KV with E4M3 block scales. Decode-like shapes: `qo_len` 1 / 4 / 9 (plain decode and MTP verify) against `kv_len` 8k–262k. Same seed for both arms; the only change between arms is `_nvfp4_kv_requires_disabled_split_kv` returning `True` (current release behaviour on sm_120) or `False` (what this PR does for this shape). 3 warmup, mean of 10, CUDA events.

| kv_len | qo_len | gate on, ms | gate off, ms | ratio | max abs diff | max rel diff |
|---|---|---|---|---|---|---|
| 8,192 | 1 | 0.235 | 0.023 | 10x | 0.031 | 0.38% |
| 8,192 | 4 | 0.692 | 0.021 | 33x | 0.031 | 0.38% |
| 8,192 | 9 | 0.691 | 0.023 | 31x | 0.063 | 0.72% |
| 32,768 | 1 | 0.905 | 0.032 | 28x | 0.016 | 0.57% |
| 32,768 | 4 | 2.753 | 0.054 | 51x | 0.016 | 0.49% |
| 32,768 | 9 | 2.757 | 0.056 | 49x | 0.016 | 0.41% |
| 131,072 | 1 | 3.615 | 0.075 | 48x | 0.008 | 0.47% |
| 131,072 | 4 | 10.994 | 0.175 | 63x | 0.016 | 0.50% |
| 131,072 | 9 | 10.994 | 0.177 | 62x | 0.016 | 0.50% |
| 262,144 | 1 | 7.139 | 0.160 | 45x | 0.004 | 0.41% |
| 262,144 | 4 | 21.792 | 0.371 | 59x | 0.008 | 0.56% |
| 262,144 | 9 | 21.812 | 0.375 | 58x | 0.008 | 0.39% |

The diff between gate on and off is one bf16 ULP of the output — same order as the residual @lesj0610 reported on sm_80 in #4747. Nothing that looks like the corruption pattern from the original report. The kernel-time ratio is what one CTA per (request, head) over the whole KV range costs; at 128k with `qo_len` 4 that is 11 ms per layer per step.

Serving side, same conclusion, different card: RTX PRO 4500 Blackwell (sm_120, 32 GB), TP1, vLLM FA2 NVFP4-KV path, Qwen3.8-27B NVFP4 KV, MTP k=4, exclusive pod. FlashInfer 0.6.18 stock vs the same wheel with the gate predicate returning `False`:

| prompt tokens | gate on, tok/s | gate off, tok/s |
|---|---|---|
| 1,210 | 56.9 | 51.8 |
| 32,954 | 26.6 | 53.5 |
| 131,578 | 10.5 | 43.7 |

Gate off restores the 0.6.16.post3 numbers (51.9 / 51.1 / 46.5) with the version that introduced the gate. Retrieval with gate off: needles 3/3 at 33k and 132k prefix; a 300-token greedy continuation on a cached 129k prefix was byte-identical to the uncached run, Δlogprob 0.0. The 1k point is slower with split-KV on, same as your 5090 row; I did not chase it either.

Not covered here: Gemma-4's 512/256 plan, head_dim 128, and anything on sm_121. `test_nvfp4_split_kv_gate_shape_scope` reads right to me for the scope it claims. If a second sm_120 data point in the PR body helps, feel free to fold the tables in.

Scripts and raw logs: https://github.com/TensorRaya/nvfp4-kv-sm120-flashinfer-evidence (`bench_nvfp4_splitkv_gate_sm120.py`, `results/nz-fimain-upstream.log`).

PR #4996, opened 2026-09-07
feat(prims-ts): add QToken-KvBlock-Sparse-Attention

PerkzZheng:
## 📌 Description

Adds PrimTS QToken-KvBlock-Sparse-Attention. Each query independently selects logical K/V blocks. One prepared API supports packed prefill and fixed-group decode without expanding block IDs into token indices.

### Interface

Public APIs are exported from `flashinfer.decode`. The [prefill/decode example](https://github.com/PerkzZheng/prims-ts-examples/blob/main/q_token_kv_block_sparse_attention.py) uses the optional group-size suggestion with a caller-cached SM count.

| Input | Layout / meaning |
| --- | --- |
| Packed Q | `[total_q, Hq, D]`; `qo_indptr` defines request-safe groups |
| Fixed Q | `[B, num_query_groups, G, Hq, D]`; no query offsets |
| K and V | HND views `[num_pages, Hkv, page_size, D]` |
| `block_table` | Dense Int32 `[num_requests, max_storage_pages]`; no CSR |
| `indexer_block_ids` | Logical selected-block IDs `[total_q, block_topk]` |
| `token_to_request`, `query_positions` | Per-query request ID and causal position |

Allocate one byte buffer with `get_q_token_kv_block_sparse_workspace_size`. Call `QTokenKvBlockSparsePagedTSWrapper.plan` outside graph capture, warm `run` once, then capture prepared runs. The plan's `batch_size` is the route count; `seq_len_q` is the maximum group size G.

`run(q, (k, v), block_table, indexer_block_ids, token_to_request, query_positions, ...)` accepts live scales and a caller-owned output. Inputs remain separate from workspace. Metadata outputs and attention/split-KV scratch occupy disjoint workspace regions. Retain the wrapper, inputs and workspace for graph lifetime. The eager convenience API is `q_token_kv_block_sparse_attention_with_paged_kv_cache`.

Callers choose G1/G2/G4/G5 with `G * (Hq/Hkv) <= 64`. `suggest_q_token_kv_block_sparse_group_size` optionally selects a smaller group to expose more CTAs. Partial groups and variable request lengths do not require SQ divisibility. The indexer supplies a distinct causal prefix of `min(block_topk, (position+1)//kv_block_size)` completed-block IDs; metadata adds the causal partial block. Q1 does not compact holes inside that prefix.

Currently supported: `kv_block_size=4`, causal non-windowed D256, BF16 Q/K/V/output or FP8-E4M3 Q/K/V with BF16/FP16 output. `page_size` is physical storage-page size; `max_seq_len_kv` bounds one request, not aggregate cache capacity. Source supports SM100/SM103; runtime qualification here is on GB300/SM103.

The general CUTLASS DSL dependency/provider minimum remains `>=4.6.2a0`. Upstream's CUDA-extra minimum `>=4.7.0a0` and CI image pin `==4.7.0` are unchanged; the general dependency floor is not a claim of PrimTS runtime qualification on 4.6.2.

### Implementation

Q1 directly maps selections in CUDA C++. Grouped routes CUB-sort at most `G*(block_topk+1)` candidates, then unique-reduce membership bits. There is no full-context bitmap. Physical locators and four-byte-packed membership words are separate; the final word's unused bytes are zeroed. Membership and causal masks preserve each query's selections.

Production sparse attention uses KV128. Packed prefill is nonsplit; fixed decode uses occupancy-based split-KV. PDL acquire/release is uniform across threads and respects barrier, SMEM and TMEM lifetime.

The rebase preserves main's static dense `BatchDecodePagedTSWrapper.plan/run` interface and row-strided `block_tables`. Encoded subpages support sparse block size four within larger physical cache pages. The generic CSR-facing wrapper rejects `backend="prims-ts"`; use the native dense PrimTS interface.

## 🔍 Related Issues

Preserves the upstream PrimTS static-plan/dense-table work in #4829. Naming follows the related block-sparse interface in #5002.

### ✅ Pre-commit Checks

Changed-file hooks pass. [Pipeline 67122954](https://github.com/flashinfer-ai/flashinfer/pull/4996#issuecomment-5623416935) reported sparse-test architecture-eligibility failures and unrelated failures. The test-only eligibility fix below is validated locally; a hosted CI rerun is still required. This is not a full-project CI pass.

## 🧪 Tests

Test-only follow-up `5720483a` gives sparse FMHA/workspace cases an exact SM100/SM103 guard; CPU contracts and metadata-only CUDA tests remain enabled. Fresh GB300/SM103 validation passes 141 sparse tests and 232 decode/backend tests, with 96 existing skips. Reported-capability probes for 9.0/12.0/10.7 retain 101 passes and skip the 40 attention-dependent cases; these probes run on SM103 and are not numerical qualification on those other GPUs. No kernel, dispatch policy or tolerance changes, and no Rubin numerical fix is claimed.

Review follow-up `3438e258` adds complete prepared-run documentation, validates encoded trace extents and restricts the FP8/BF16 reduction exception to structurally valid sparse routes. Fresh affected-suite validation passes on GB300/SM103 with CUTLASS DSL 4.7.1. Loader/resolver edits only clarify comments; the TMA schedule is unchanged. These are not full-project CI or SM100-runtime qualification claims.

- Metadata/wrapper: 141 passed.
- Decode/backend: 232 passed, 96 existing hardware/profile skips.
- Shared block-sparse: 229 passed, one skip.
- Trace schemas: 972 passed, including new rejection tests for incompatible encoded storage extents.
- Prior qualification also includes 96 dense passes in a local SM103 opt-in probe without changing production architecture guards.
- Eight focused cases pass both memcheck and initcheck with zero device-memory errors, including inert and poisoned page-table padding. Memcheck retains the previously documented CUDA-Python API-probe reporting workaround; all device-memory checks remain enabled.
- Prior example qualification covers packed/fixed, request-boundary, reference and poisoned graph-replay checks; the recorded vLLM ownership/reference gate passes 119/119.

Coverage includes Q1/Q2/Q4/Q5, partial groups, variable lengths, causal tails, BF16/FP8, storage-page mapping, split-KV, workspace ownership and PDL.

## Standalone performance

Timing qualification caveat: the preserved tables below used a uniform-byte L2 scrub, which Blackwell may compress. They are historical results, not signoff for non-compressible cold-L2 eviction. The [two published benchmark suites](https://github.com/PerkzZheng/flashinfer-prims-ts-validation/blob/main/docs/sparse_attention.md) now use separately seeded random bytes and record a distinct timing protocol; corrected-protocol GPU timing must be rerun. The recorded numerical checks and separately collected warmed Nsight stage results are unaffected by this timing caveat.

Real recorded top-k routes; CUDA graphs; 20 warmups and 300 balanced samples. Each sample starts with a same-stream 258.5 MiB L2 eviction before the complete graph, not between its constituent kernels. PrimTS includes metadata, attention and reduction; Triton includes index expansion, sparse attention and reduction.

TP2: Hq/Hkv=12/1, D256, 128K logical model bound. All 62 cases pass: Q5 2/2, automatic-G matrix 54/54 and G4 prefill controls 6/6. The comparator is current-main vLLM `a841edb6` on `c55e15a4`, whose Triton kernels are faster than the earlier PR53896-based comparator. Speedup is `Triton / PrimTS`.

| TP2 workload | Cases | BF16 speedup | FP8 speedup |
| --- | --- | --- | --- |
| Packed prefill, automatic G5 | BS1, 8K–32K | 2.542–2.715x | 2.242–2.461x |
| Decode SQ1, G1 | BS8, 8K–32K | 0.959–0.990x | 0.951–0.982x |
| Decode SQ1, G1 | BS64/256, 8K–32K | 1.147–1.220x | 1.182–1.233x |
| Decode SQ4, G4 | BS8, 8K–32K | 0.835–0.875x | 0.859–0.905x |
| Decode SQ4, G4 | BS64/256, 8K–32K | 1.570–2.059x | 1.278–4.436x |
| Q5/MTP4 grouped-route proxy | BF16 16/32 groups, 8K | 1.274–1.438x | — |

Prefill complete times span 544.12–2511.94 µs PrimTS versus 1477.24–6384.75 µs Triton for BF16, and 539.34–2487.74 versus 1327.45–5578.20 µs for FP8. G5 reduces latency by 5.12–7.45% versus matched G4. Q5 takes 34.85/40.88 µs at 16/32 groups versus Triton's 44.39/58.78 µs. The Q5 proxy shares one request/cache across route groups; it is not a disjoint-cache production BS16/32 measurement.

Low-route decode remains slower: BS8/SQ1 by 1.1–5.2%, BS8/SQ4 by 10.5–19.8%, and BS1 as listed below. Other BS64/256 rows are faster. These are not new rebase regressions.

| BS1 configuration, 8K/16K/32K | PrimTS complete | Triton complete | PrimTS regression |
| --- | --- | --- | --- |
| BF16 SQ1/MTP0, G1 | 19.79–20.34 µs | 14.42–15.89 µs | 24.7–41.1% |
| FP8 SQ1/MTP0, G1 | 20.58–20.68 µs | 14.29–14.98 µs | 37.8–44.8% |
| BF16 SQ4/MTP3, automatic G1 | 22.30–22.48 µs | 16.48–17.14 µs | 30.6–35.3% |
| FP8 SQ4/MTP3, automatic G1 | 21.12–22.04 µs | 16.28–16.39 µs | 28.9–35.1% |

For BS1/SQ4, automatic G1 keeps four independent routes, avoiding grouped sort/union and exposing more CTAs. No shape-specific policy was added for these outliers.

All 62 same-node old-head controls pass. Versus `96d9f727`, automatic-G latency changes by −0.04% geometrically, with a largest increase of 3.03%; Q5 changes by at most +0.09%, and G4 controls by at most +0.55%. No tile/split configuration changed. Raw samples, component timings and source/trace fingerprints are retained in `qsa_bench/rebase_main_20260909/`.

### Framework validation

The [vLLM integration guide](https://github.com/PerkzZheng/vllm/blob/7e8090112e8ea2c53da0bed5a4faf1102abfdb97/vllm/models/qwen4_exp/nvidia/Q_TOKEN_KV_BLOCK_SPARSE_TS.md) describes packed prefill, fixed decode and shared workspace ownership. The recorded model-validation results and limits are summarized below; this review follow-up does not rerun model accuracy or performance.

TP2 FP8/MTP3 accuracy uses the fixed temperature-0.6, seed-42, xhigh sampler and 131072-token output budget. LongBench is the existing untruncated 48-question cohort, not the full benchmark.

| Task | Triton | PrimTS |
| --- | --- | --- |
| GSM8K | 1290/1319 (97.80%) | 1286/1319 (97.50%) |
| GPQA-Diamond, two reps | 365/396 (92.17%) | 361/396 (91.16%) |
| LongBench v2 cohort | 32/48 (66.67%) | 33/48 (68.75%) |

All 3,526 requests complete without request errors. GPQA has one capped Triton answer and two capped PrimTS answers; both PrimTS caps have no final choice. All caps are incorrect, so strict scores are unchanged. PrimTS is 1.01 percentage points lower on GPQA; both backends also change nine choices between repetitions. The unfinished-answer caveat remains, and these stochastic scores do not prove numerical equivalence.

All eight paired pure-stage Nsight comparisons pass the both-rank work audit. These are warmed CUDA-graph node traces, separate from cold-L2 standalone results. Prefill measures the fourth request after three warmups; decode includes MTP3 draft work and uses 64 resident requests (256 target query tokens). Resident BS256 exceeds this TP2 hybrid-cache capacity and is not claimed.

| Pure stage, 8K/16K | KV | Sparse speedup | All-layer speedup |
| --- | --- | --- | --- |
| Prefill, BS1 | BF16 | 2.172–2.473x | 1.083–1.088x |
| Prefill, BS1 | FP8 | 1.915–2.211x | 1.067x |
| Decode MTP3, BS64 | BF16 | 1.565–1.657x | 1.020–1.022x |
| Decode MTP3, BS64 | FP8 | 1.178–1.284x | 1.014x |

Prefill sparse and all-layer absolute savings agree within 0.29 ms. The pinned image needs main's BF16 MoE activation-padding fix for prefill; both backends use the same local backport. This is an image compatibility fix, not a sparse-attention change. Remaining low-route standalone regressions and SM100 runtime qualification stay explicit.

## 🔬 Experimental Track

```experimental-tests
#
#
```

## Reviewer Notes

Please focus on dense-table/workspace ownership, sort/union membership semantics, causal masking, packed/fixed query layouts and PDL resource lifetime. SM100 runtime qualification, unfinished GPQA answers and low-route latency remain follow-ups. AI assistance was used.

## Summary by CodeRabbit

* **New Features**
  * Added QToken-KV block-sparse attention with packed and fixed-query support.
  * Added prepared batch decode planning and execution APIs.
  * Added four-token pages, separate semantic and storage page sizes, encoded page locators, and query-to-KV head ratios up to 128.
  * Expanded trace coverage for new decode configurations.

* **Documentation**
  * Documented sparse-attention APIs, planning workflows, metadata requirements, and supported configurations.

* **Breaking Changes**
  * The legacy paged-KV wrapper no longer supports the `prims-ts` backend or explicit causal-mode overrides.
  * PrimTS decode now uses dense block tables instead of CSR page metadata.
  * Storage-overlap checks are no longer performed; callers must keep buffers disjoint.


flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/helpers_softmax.py
```diff
@@ -95,10 +96,13 @@ def _compute_fp8_p_regs_and_local_sums(
         safe_new_max_0 = Float32(0.0)
     if safe_new_max_1 == _neg_max_f32():
         safe_new_max_1 = Float32(0.0)
+    probability_scale_log2 = Float32(0.0)
+    if cutlass.const_expr(cfg.use_fp8_p448):
```
PerkzZheng: we should by default use that, please remove it.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/smem_resources.py
```diff
@@ -435,23 +496,40 @@ def tma_load(self, stage_info: StageInfo) -> None:
...
-                        self._tma_coords(
-                            logical_q_group_idx,
-                            logical_h_k_idx,
-                            logical_b_idx,
-                            Int32(head_dim_offset),
-                        ),
-                        stage_info.barrier,
-                    )
-                self._complete_grouped_q_padding(stage_info)
+                # Page-4 profiles divide their D64 Q chunks across at most one
```
PerkzZheng: we should make the comment general. For example, if there are multiple warps for load task, we should split the work.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/smem_resources.py
```diff
@@ -986,17 +1078,96 @@ def _producer_load(
                 num_chunks = head_dim_stage // chunk_hd
                 tile_chunk_elems = chunk_hd * cfg.tile_size_kv
                 page_chunk_elems = chunk_hd * cfg.num_tokens_per_page
+                if cutlass.const_expr(cfg.uses_encoded_subpage_locators):
```
PerkzZheng: what does uses_encoded_subpage_locators mean ?

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/smem_resources.py
```diff
@@ -1298,6 +1552,98 @@ def page_ids(self, tile_idx: Int32) -> cutlass.Array:
...
+            ) * Int32(QSA_PAGE_MEMBERSHIP_BITS)
+            return (membership_word >> membership_shift) & Uint32(
+                QSA_PAGE_MEMBERSHIP_MASK
+            )
+        offset = local_tile_idx * pages_per_tile + page_frag
+        packed_locator = Int32(self._smem_page_offsets[offset])
+        return Uint32(packed_locator) & Uint32(QSA_PAGE_MEMBERSHIP_MASK)
+
+    @cute.jit
+    def qsa_page_memberships4(
```
PerkzZheng: where this is been called ?

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/smem_resources.py
```diff
@@ -1298,6 +1552,98 @@ def page_ids(self, tile_idx: Int32) -> cutlass.Array:
...
+        self,
+        stage_info: StageInfo,
+        local_tile_idx: Int32,
+        page_frag: Int32,
+    ) -> cutlass.Array:
+        """Load four aligned grouped-Q membership bytes from one vector."""
```
PerkzZheng: why we keep padded member ship bytes in SMEM (not using packed format like in GMEM where we only need 32 bytes for one tile kv) ? if just loading 32 bytes per tile kv, we might just do LDG instead staging through smem.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/smem_resources.py
```diff
@@ -1383,7 +1728,114 @@ def _producer_load_page_offsets(
             page_idx_kv = self.page_idx_kv
         smem_page_offsets = self._smem_page_offsets
         lane_idx = cute.arch.thread_idx()[0] & Int32(0x1F)
-        if cutlass.const_expr(self.stage_page_ids_per_tile):
+        if cutlass.const_expr(
+            self.stages_encoded_locators_per_tile
+            and cfg.uses_qsa_page_membership
+            and section == FmhaStage.Head
+        ):
+            # Retain only the low membership byte for Softmax. The full
```
PerkzZheng: what this is meant to do ?

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/tmem_corr.py
```diff
@@ -1723,7 +1724,45 @@ def _stage_fp16_o_regs_to_smem(
         fragments into the swizzled SMEM layout before a second operation
         reloads contiguous vectors.
         """
-        if cutlass.const_expr(output_pair_regs < 4):
+        if cutlass.const_expr(
+            not cfg.use_keeps_mma_ab
+            and cfg.head_dim_per_stage_kv != 0
+            and cfg.head_dim_kv_stage == 64
+            and cfg.tile_size_q in (8, 16, 32)
```
PerkzZheng: does this mean we haven't supported this configuration before "?

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/smem_resources.py
PerkzZheng: This block is the **long-route membership side-channel**. It only runs when we do **not** hold the full encoded-locator window (`stages_encoded_locators_per_tile`: graph-stable QSA with more than `QSA_HELD_LOCATOR_MAX_TILES` local tiles).

On that path the page-offset ring keeps only **one KV128 tile of locators** (32 `int32`s) for TMA, then overwrites them. Softmax still needs every page’s 8-bit query-membership mask later, so HEAD copies just the low byte of each packed locator for the CTA-local tile window into `_smem_qsa_memberships`.

Layout: 8 bits × 32 pages/tile, packed 4 bytes per `uint32` (`QSA_PAGE_MEMBERSHIPS_PER_WORD`). That is 32 bytes of membership per tile, not a second 128-byte locator buffer. `qsa_page_membership` / `qsa_page_memberships4` read this compact buffer; the held-locator path never enters here and peels `& 0xFF` from the locators already in SMEM.

Invalid / OOB pages store membership `0`, so Softmax can drop the whole page from that query without a live locator.

This is **not** a 32-byte LDG of a standalone membership tensor: GMEM still stores fused `(locator << 8) | membership`. We strip after the locator load so Softmax does not depend on the one-tile locator ring remaining live.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/tmem_s.py
```diff
@@ -2005,6 +2134,56 @@ def _compute_softmax_loop_swaps(
                             s_vals[s_base + 0] = _neg_max_f32()
                             s_vals[s_base + 1] = _neg_max_f32()
 
+        if cutlass.const_expr(cfg.uses_qsa_page_membership):
```
PerkzZheng: can you use a member function to wrap this.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_kernel.py
```diff
@@ -2847,7 +3043,11 @@ def fmha_decode_launch(
         cluster=cluster_shape,
         stream=stream,
         min_blocks_per_mp=_decode_min_blocks_per_mp(cfg, effective_seq_len_kv),
-        use_pdl=cfg.use_parallel_separate_reduction_pdl,
+        # QSA initializes all local task resources before acquiring its
+        # immediately preceding metadata producer. Split-QSA additionally
+        # defers runtime prefix pruning until after that acquire. Ordinary
+        # dense/paged attention retains stream ordering.
+        use_pdl=cfg.uses_qsa_metadata_pdl,
```
PerkzZheng: we shouldn't use uses_qsa_metadata_pdl option. PDL should be a general feature that doesn't rely on metadata kernels. it is just use_pdl or not.

flashinfer/attention/prims_ts/decode.py
```diff
@@ -1221,55 +1538,168 @@ def fits_one_service_wave(config: "FmhaDecodeConfig", num_q_ctas: int) -> bool:
     # Device capacity participates in automatic selection. Resolve it in the
     # target device context without introducing caller-visible policy knobs.
     with torch.cuda.device(device_index):
-        cfg = make_config({"use_variable_seqlens_q": True} if use_packed_q else None)
         max_active_clusters = get_max_active_clusters_for_cluster_size(1)
+        if use_qsa_route:
```
PerkzZheng: can we wrap qsa route into one function ?

tests/attention/test_attention_ts_pdl_contract.py
PerkzZheng, line 1: I don't think this is necessary.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/helpers_softmax.py
PerkzZheng: Done. I removed the optional branch and made the safe FP8 softmax behavior unconditional. `_compute_fp8_p_regs_and_local_sums` now always substitutes zero for an all-masked row's `NEG_FLT_MAX` exponent base, so the row produces zero P values instead of NaNs.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/smem_resources.py
PerkzZheng: Done. The comment and implementation are now general: when the load task has multiple warps, D64 Q chunks are partitioned by load-warp rank. `q_issuer_warps = min(num_chunks, load_num_warps)` ensures each chunk is issued exactly once, and any excess K/V loader warps remain idle for the Q load.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/smem_resources.py
PerkzZheng: I removed that ambiguous name. The current config separates the concepts as `has_storage_subpages` (one physical storage page contains multiple logical pages), `uses_scattered_page_route` (the loader consumes a locator per page fragment), and `uses_qsa_sparse_page_route` (the QSA specialization). This avoids calling plain page locators “encoded.”

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/smem_resources.py
PerkzZheng: It is called by `TmemSResource._qsa_keeps_kv128_membership_word` in `tmem_s.py`. That helper invokes `qsa_page_memberships4` four times to assemble the lane-local 16-page keep word used by the Q64/KV128 Keeps softmax path. I added this call-site information to the function docstring.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/smem_resources.py
PerkzZheng: Agreed, and the current implementation keeps membership packed in both GMEM and SMEM. `qsa_page_memberships` stores four 8-bit membership masks per Int32 word; the producer cooperatively stages those packed words, and the consumer extracts bytes into registers. There is no longer a four-times-larger padded-byte SMEM representation. SMEM is retained so the packed row is loaded once and reused by the softmax consumers across the K/V pipeline instead of issuing duplicate GMEM loads from each consumer.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/smem_resources.py
PerkzZheng: This old branch preserved membership while a long-route locator window was being refilled. I removed that side channel along with fused locator/membership encoding. The current design has plain Int32 locators plus one separate packed GMEM membership row, which the Head page-offset producer stages once. There is no distinct long-route membership format now.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/tmem_corr.py
PerkzZheng: No. That branch was a new experimental D64-staged Swaps output-reorganization workaround, not evidence of a previously supported configuration. I removed it and restored the original generic STSM path. Production QSA now stays on the qualified KV128 configurations rather than adding a special output layout here.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/tmem_s.py
PerkzZheng: Done. The QSA Swaps membership-mask block is now wrapped in the `TmemSResource._apply_qsa_swaps_page_membership_mask` member function, and the softmax loop only dispatches to that helper.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_kernel.py
PerkzZheng: Done. The attention launch now uses the general `cfg.use_pdl`; `uses_qsa_metadata_pdl` was removed. The acquire is uniform across all CTA threads and occurs after SMEM/TMEM/barrier initialization. Reducer release remains separately guarded by `use_parallel_separate_reduction_pdl` because that flag describes the optional attention-to-reducer dependency.

flashinfer/attention/prims_ts/decode.py
PerkzZheng: Done. QSA configuration and split selection are isolated in `_resolve_qsa_decode_config`; the common launch-spec resolver now delegates to it and keeps the ordinary dense/paged selection path separate.

tests/attention/test_attention_ts_pdl_contract.py
PerkzZheng: Agreed. I removed this source-inspection test file. PDL behavior is covered through the runtime QSA metadata/attention CUDA-graph tests instead of testing source text and implementation structure.

flashinfer/attention/prims_ts/decode.py
```diff
@@ -2191,9 +3016,8 @@ def __init__(self, kv_layout: Literal["HND"] = "HND") -> None:
     @flashinfer_api
     def plan(
```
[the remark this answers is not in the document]
PerkzZheng: This signature change is intentional: the author requested dense page tables only for PrimTS. `plan(block_table, seq_lens, ...)` replaces the three CSR tensors; the updated description documents the migration. The generic CSR-facing wrapper rejects `backend="prims-ts"` and points callers to the standalone API. See [the current API](https://github.com/PerkzZheng/flashinfer/blob/c7f73ca6f8cfcf961908d49d55ddda82e8fe6441/flashinfer/attention/prims_ts/decode.py).

flashinfer/attention/prims_ts/decode.py
```diff
@@ -2578,9 +3411,8 @@ def run(
 def batch_decode_with_paged_kv_cache(
```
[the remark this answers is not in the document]
PerkzZheng: Intentional dense-only migration requested by the author. The convenience API now takes `q, paged_kv_cache, block_table, seq_lens`; it no longer accepts CSR inputs. This compatibility change and the standalone migration path are stated explicitly in the PR description and [current docstrings](https://github.com/PerkzZheng/flashinfer/blob/c7f73ca6f8cfcf961908d49d55ddda82e8fe6441/flashinfer/attention/prims_ts/decode.py).

flashinfer/attention/prims_ts/decode.py
```diff
@@ -2160,20 +2766,239 @@ def prims_ts_batch_decode_with_kv_cache(
...
+        page_size=page_size,
+        use_qsa_route=False,
+    )
+    if prepared_out is not out:
+        raise RuntimeError("prepared PrimTS output storage changed unexpectedly")
+    return plan
+
+
+@flashinfer_api(trace=prims_ts_decode_trace_dispatch)
+def prims_ts_batch_decode_with_kv_cache(
```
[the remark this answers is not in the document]
PerkzZheng: Intentional dense-only migration requested by the author. The standalone PrimTS API keeps caller-owned workspace and now takes dense `block_table`, `seq_lens`, and `max_seq_len`, without CSR indptr/indices. The updated PR description discloses the signature change; [the current API](https://github.com/PerkzZheng/flashinfer/blob/c7f73ca6f8cfcf961908d49d55ddda82e8fe6441/flashinfer/attention/prims_ts/decode.py) documents the inputs.

flashinfer/attention/prims_ts/decode.py
```diff
@@ -2160,20 +2766,239 @@ def prims_ts_batch_decode_with_kv_cache(
...
+        _block_table=block_table,
+        _workspace=workspace,
+        _compiled_main=compiled_main,
+        _compiled_reducer=compiled_reducer,
+    )
+    return plan, runtime.out
+
+
+@flashinfer_api
+def prepare_prims_ts_batch_decode_with_kv_cache(
```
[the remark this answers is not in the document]
PerkzZheng: Addressed in the current tree: `prepare_prims_ts_batch_decode_with_kv_cache` now has full Parameters and Returns sections, including dense table/sequence inputs and prepared-plan lifetime. See [the docstring](https://github.com/PerkzZheng/flashinfer/blob/c7f73ca6f8cfcf961908d49d55ddda82e8fe6441/flashinfer/attention/prims_ts/decode.py).

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/smem_resources.py
```diff
@@ -1826,6 +2312,99 @@ def _producer_load_kv(
                 num_chunks = head_dim_stage // chunk_hd
                 tile_chunk_elems = chunk_hd * cfg.tile_size_kv
                 page_chunk_elems = chunk_hd * cfg.num_tokens_per_page
+                stage_base = self._stage_base(stage_info)
+                if cutlass.const_expr(cfg.uses_encoded_subpage_locators):
+                    assert page_fragments in (16, 32)
```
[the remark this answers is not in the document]
PerkzZheng: The 16-fragment case is not supported by this branch's scattered-storage load path. `validate_storage_page_size` rejects encoded subpages unless the semantic page size is 4, and scattered routes require KV128. Hence this path has exactly 32 fragments; the single-warp load asserts that invariant. The multiwarp path also bounds transaction indices. See [the geometry validation](https://github.com/PerkzZheng/flashinfer/blob/c7f73ca6f8cfcf961908d49d55ddda82e8fe6441/flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_config.py#L4052) and [load implementation](https://github.com/PerkzZheng/flashinfer/blob/c7f73ca6f8cfcf961908d49d55ddda82e8fe6441/flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/smem_resources.py#L2298).

flashinfer/attention/prims_ts/qsa_metadata.py
```diff
@@ -0,0 +1,2211 @@
...
+        )
+
+    tail_tokens = (last_position + 1) % SEMANTIC_PAGE_SIZE
+    tail_padding = tl.where(
+        tail_tokens == 0,
+        0,
+        SEMANTIC_PAGE_SIZE - tail_tokens,
+    )
+    seq_len = union_pages * SEMANTIC_PAGE_SIZE - tail_padding
+    tl.store(seq_lens_ptr + group, tl.where(group_valid, seq_len, 1))
```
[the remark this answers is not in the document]
PerkzZheng: Addressed by the replacement CUDA C++ producer. The old Triton bitmap packers no longer exist. An empty/inert route publishes locator -1, a zero membership word, and sequence length 1. The membership-padding follow-up also zeros unused bytes in the final live word. See [the producer](https://github.com/PerkzZheng/flashinfer/blob/c7f73ca6f8cfcf961908d49d55ddda82e8fe6441/include/flashinfer/attention/prims_ts/q_token_kv_block_sparse_metadata.cuh); metadata and poisoned-output/replay tests pass.

tests/trace/example.py
```diff
@@ -1752,9 +1752,11 @@
     )
 
     _pts_B, _pts_SQ, _pts_SK = 4, 4, 2048
-    _pts_Hq, _pts_Hkv, _pts_D, _pts_PS = 32, 4, 128, 32
+    _pts_Hq, _pts_Hkv, _pts_D = 32, 4, 128
```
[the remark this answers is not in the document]
PerkzZheng: These fixtures exercise generic dense decode with encoded storage subpages, not the explicit query-token sparse membership route. Storage page encoding alone does not select the D256-only sparse policy. D128 therefore remains valid here; changing it would remove generic decode coverage. The current storage-subpage policy tests verify that this route does not implicitly enable sparse-query attention.

tests/trace/fi_trace_out/attention_ts_decode_tuple_encoded_page4_multi_q_sq4_h32_kv4_d128_ps32_sps4.json
```diff
@@ -84,24 +88,24 @@
...
     },
-    "paged_kv_indptr": {
-      "shape": [
-        "len_indptr"
-      ],
-      "dtype": "int32"
+    "semantic_page_size": {
+      "shape": null,
+      "dtype": "int32",
+      "description": "Semantic block-table page size passed to the API; fixed to 4 when locators address subpages of larger physical cache pages."
```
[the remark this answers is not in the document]
PerkzZheng: Addressed: both current encoded-page fixtures include `semantic_page_size == page_size`, with the fixed page-size axis set to 4. This ties the API scalar to the schema axis. See [the attention fixture](https://github.com/PerkzZheng/flashinfer/blob/c7f73ca6f8cfcf961908d49d55ddda82e8fe6441/tests/trace/fi_trace_out/attention_ts_decode_tuple_encoded_page4_multi_q_sq4_h32_kv4_d128_ps32_sps4.json) and the corresponding standalone fixture.

tests/trace/fi_trace_out/prims_ts_decode_wrapper_tuple_encoded_page4_multi_q_sq4_h32_kv4_d128_ps32_sps4.json
```diff
@@ -34,18 +34,25 @@
...
+      "type": "const",
+      "value": 4,
+      "description": "Semantic block-table page size for encoded subpage locators."
     }
   },
   "constraints": [
     "head_dim in (64, 128, 256)",
-    "page_size in (16, 32, 64, 128)",
+    "storage_page_size > page_size",
+    "storage_page_size % page_size == 0",
```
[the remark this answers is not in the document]
PerkzZheng: Addressed in all three current encoded-page decode fixtures: the constraint is now `1 <= num_qo_heads // num_kv_heads <= 128`, matching the templates. The separate block-sparse fixtures retain their intended bounds.

flashinfer/attention/prims_ts/q_token_kv_block_sparse_metadata.py
```diff
@@ -0,0 +1,2068 @@
...
+            self._default_out = None
+
+    def _require_config(self) -> _QTokenKvBlockSparsePlanConfig:
+        config = self._config
+        if config is None:
+            raise RuntimeError("plan() must be called before run()")
+        return config
+
+    @flashinfer_api
+    def run(
```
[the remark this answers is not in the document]
PerkzZheng: Fixed in [3438e258](https://github.com/flashinfer-ai/flashinfer/commit/3438e2588216dda21182105e5e2f75b45cb612f2). `QTokenKvBlockSparsePagedTSWrapper.run` now has NumPy-style Parameters and Returns sections covering all arguments, packed/fixed layouts, scales and output ownership. The repository's docstring checker reproduces the warning on the parent and reports no finding for this method after the change.

flashinfer/attention/prims_ts/q_token_kv_block_sparse_metadata.py
```diff
@@ -0,0 +1,2068 @@
...
+        Q1 does not compact missing entries or arbitrary ``-1`` holes inside
+        this required prefix. The incomplete causal tail is derived from
+        ``query_positions`` and must not be inserted into the selected prefix.
+
+        Run once eagerly after :meth:`plan`; capture only subsequent calls with
+        stable tensor storage. A changed packed extent or retained K/V/offset
+        storage triggers a new preparation outside capture. ``sm_scale`` and
+        ``v_scale`` are live launch scalars, so frameworks may update Q/K and V
+        dequantization scales without rebuilding the capacity plan.
+        """
```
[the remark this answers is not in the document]
PerkzZheng: Fixed in [3438e258](https://github.com/flashinfer-ai/flashinfer/commit/3438e2588216dda21182105e5e2f75b45cb612f2). Added Parameters/Returns for all ten run arguments, including packed-route offsets, live dequantization scales and caller-owned/reused output. The existing selected-block-prefix and graph-lifetime contracts remain documented. The repository docstring checker passes for this method; metadata tests pass 141/141.

flashinfer/attention/prims_ts/README.md
```diff
@@ -39,6 +42,75 @@ default; callers that have already validated their inputs may use
...
+`suggest_q_token_kv_block_sparse_group_size` helper takes a caller-cached SM
+count and prefers the largest legal group that can fill one service wave using
+independent routes plus useful split-KV work. It falls back toward Q1 when a
+large union would leave SMs idle. The helper never queries device properties
+or reads tensors.
+
+`q_token_kv_block_sparse_attention_with_paged_kv_cache` is the eager
+plan-plus-run convenience API and is not graph-capturable. Shared argument
+spellings intentionally match paged block-sparse attention: `q`,
+`paged_kv_cache`, `page_size`, `kv_block_size`, `mask_type`,
```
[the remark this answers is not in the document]
PerkzZheng: Fixed in [3438e258](https://github.com/flashinfer-ai/flashinfer/commit/3438e2588216dda21182105e5e2f75b45cb612f2). Removed `page_size` from the convenience API's documented argument list and explicitly stated that its physical page extent is inferred from `paged_kv_cache`. The prepared wrapper's `plan` still takes `page_size` explicitly; no API signature changed.

flashinfer/trace/templates/attention.py
```diff
@@ -923,23 +933,59 @@ def _add_fmha_cache_schema(inputs, axes, *, cache_param: str, combined: bool):
...
+            "whose physical storage extent can be inspected."
+        )
+    if semantic_page_size == storage_page_size:
+        return False
+    if semantic_page_size != 4:
+        raise ValueError(
+            "A PrimTS page_size that differs from the physical cache extent "
+            "must be 4 (encoded subpage locators)."
+        )
+    return True
```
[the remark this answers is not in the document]
PerkzZheng: Fixed in [3438e258](https://github.com/flashinfer-ai/flashinfer/commit/3438e2588216dda21182105e5e2f75b45cb612f2). When semantic and physical extents differ, trace selection now requires physical storage to be larger than and divisible by the semantic page size. The existing semantic-page-four restriction remains. New public `.fi_trace()` regressions cover physical extents 2 and 6 for both one-shot and standalone entry points: both fail before the fix and pass afterward. The complete trace consistency suite passes 972 tests.

flashinfer/jit/q_token_kv_block_sparse_metadata.py
```diff
@@ -0,0 +1,52 @@
...
+        checkout / "csrc" / "prims_ts_q_token_kv_block_sparse_metadata_jit_binding.cu",
+    ]
+    checkout_include = checkout / "include"
+    if all(path.is_file() for path in checkout_sources) and checkout_include.is_dir():
+        return checkout_sources, [checkout_include]
+    return [
+        jit_env.FLASHINFER_CSRC_DIR / "prims_ts_q_token_kv_block_sparse_metadata.cu",
+        jit_env.FLASHINFER_CSRC_DIR
+        / "prims_ts_q_token_kv_block_sparse_metadata_jit_binding.cu",
+    ], []
```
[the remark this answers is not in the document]
PerkzZheng: I traced this through the actual compiler flags; the installed fallback is already covered. `gen_jit_spec` forwards *extra* includes, while `cpp_ext.build_common_cflags()` always calls `get_system_includes()`, which adds `jit_env.FLASHINFER_INCLUDE_DIR.resolve()` independently.

On [3438e258](https://github.com/flashinfer-ai/flashinfer/commit/3438e2588216dda21182105e5e2f75b45cb612f2), I forced the installed source-path fallback, verified that the extra include list was empty and the package include remained in the compiler flags, then compiled and loaded those fallback sources under a fresh module name. This avoids a cached binary hiding an include failure. No redundant include-path change is needed. This was an installed-layout JIT probe, not a separate wheel-build qualification.

tests/attention/test_attention_ts_decode.py
```diff
@@ -2252,6 +2652,713 @@ def test_attention_ts_decode_page_offsets_cross_window_schedule_is_safe(
...
+        q,
+        (k_cache, v_cache),
+        workspace,
+        block_table,
+        seq_lens,
+        max_seq_len,
+        out=output,
+        mask_type="causal",
+        page_size=semantic_page_size,
+    )
```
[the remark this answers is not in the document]
PerkzZheng: The cited test uses a different plan class from the value-checking wrapper. `prims_ts_batch_decode_with_kv_cache` prepares a `PrimsTSBatchDecodePlan`; its `run` does not call `_validate_decode_run_metadata_values`. `_validate_block_table_metadata` checks tensor structure, not locator values. The synchronizing value validator belongs to `BatchDecodePagedTSWrapper.run`.

The existing `test_attention_ts_decode_page4_inert_block_table_row_is_zero` passes on [3438e258](https://github.com/flashinfer-ai/flashinfer/commit/3438e2588216dda21182105e5e2f75b45cb612f2) with exact-zero padded output, both normally and under memcheck/initcheck. I retained the standalone inert-row contract and did not relax the other wrapper's negative-locator validation.

tests/attention/test_attention_ts_decode.py
```diff
@@ -4026,6 +5132,384 @@ def _record_kv_promoter(*args, q_candidate, **kwargs):
...
+
+    from flashinfer.attention.prims_ts import decode as decode_module
+    from flashinfer.attention.prims_ts.kernels.fmha_decode import fmha_decode_config
+
+    monkeypatch.setattr(
+        fmha_decode_config,
+        "get_max_active_clusters_for_cluster_size",
+        lambda cluster_size: 148 // cluster_size,
+    )
+    monkeypatch.setattr(torch.cuda, "device", lambda *_args, **_kwargs: nullcontext())
```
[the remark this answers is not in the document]
PerkzZheng: This call path is no longer reachable in the current config builder. The sparse resolver supplies an explicit positive `splits_kv` (including 1) and explicit launch fields. `make_decode_config` only falls through to `_apply_auto_launch_mode` when the launch fields are implicit and `splits_kv <= 0`.

I ran all nine existing structural sparse-policy cases on [3438e258](https://github.com/flashinfer-ai/flashinfer/commit/3438e2588216dda21182105e5e2f75b45cb612f2) with `CUDA_VISIBLE_DEVICES=''` and replaced `HardwareInfo` with a function that raises on any call. All nine pass, including packed Q and direct FP8 Q2. I therefore left the helper unchanged rather than adding an unused hardware mock.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/smem_resources.py
```diff
@@ -985,17 +1076,97 @@ def _producer_load(
...
+                                partition_stage_base.subview(lane_smem_page_offset),
+                                tma_desc,
+                                (
+                                    Int32(lane_global_head_dim_offset),
+                                    lane_token_offset,
+                                    logical_h_k_idx,
+                                    lane_physical_page,
+                                ),
+                                stage_info.barrier,
+                            )
```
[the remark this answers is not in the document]
PerkzZheng: These lanes issue different TMA transactions, not duplicate copies of one tile. The [PTX cp.async.bulk.tensor definition](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#data-movement-and-conversion-instructions-cp-async-bulk-tensor) is a per-thread instruction; complete_tx accounts for bytes copied. The one-thread/election idiom applies when the participating lanes describe the same copy.

For production BF16 KV128/D128, there are 32 page fragments times two D64 chunks: 64 independent copies of `4*64*2 = 512` bytes. Eight loader warps own eight distinct copies each. Destinations do not overlap, total bytes are 32768, and the pipeline assigns 4096 expected bytes per loader warp. Adding a warp election here would omit seven of each warp's eight required copies.

[3438e258](https://github.com/flashinfer-ai/flashinfer/commit/3438e2588216dda21182105e5e2f75b45cb612f2) clarifies this ownership and corrects a stale elected-lane comment; the loader's executable AST is unchanged. Partition accounting covers every transaction exactly once. Fresh GB300/SM103 metadata/decode tests pass, and eight focused cases covering split/graph/membership/padding paths pass both memcheck and initcheck with zero device-memory errors. No blanket issuer serialization is warranted by this finding.

PerkzZheng:
Addressed the additional reduction-policy nit in [3438e258](https://github.com/flashinfer-ai/flashinfer/commit/3438e2588216dda21182105e5e2f75b45cb612f2): `supports_reduction_dtypes` now uses `uses_q_token_kv_block_sparse_page_route`, not the raw sparse flag, for the FP8/BF16 exception. The page32 negative regression fails on the parent and passes after the fix; page4 remains supported.

This incremental commit also fixes API documentation and encoded trace validation. Fresh GB300/SM103 + CUTLASS DSL 4.7.1 results: 141 metadata, 232 decode/backend, 972 trace and 229 shared block-sparse tests pass (1574 total; 97 existing hardware/profile skips). Eight focused cases pass each of memcheck and initcheck with zero device-memory errors; memcheck retains the previously documented CUDA-Python API-probe reporting workaround. Changed-file pre-commit hooks pass. The installed-layout, inert-row and CPU-policy findings were checked and answered individually; TMA ownership is clarified without changing the schedule. No new performance/e2e-equivalence or SM100-runtime qualification is claimed. Thread resolution is left to the reviewer.

tests/attention/test_attention_ts_decode.py
```diff
@@ -1464,6 +1833,38 @@ def _apply_speculative_tail_markers(case):
...
+        out_dtype=BFloat16,
+        use_paged_kv=True,
+        num_tokens_per_page=page_size,
+        groups_tokens_heads_q=True,
+        use_q_token_kv_block_sparse_route=True,
+        heads_q_per_kv=12,
+        max_seq_len_q=4,
+        tile_size_q=64,
+    )
+    assert cfg.supports_reduction_dtypes is (page_size == 4)
```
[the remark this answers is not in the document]
PerkzZheng: The semantic page-size check is already part of [`uses_q_token_kv_block_sparse_page_route`](https://github.com/flashinfer-ai/flashinfer/blob/5720483a03821d37c213d10d53275d102d132108/flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_config.py#L705): it includes `num_tokens_per_page == 4`. `supports_reduction_dtypes` consumes that predicate, so ordinary page32 configurations do not receive the FP8/BF16 exception.

I reran the existing page4/page32 regression on the current head: both cases pass (page4 enabled, page32 disabled), within 232 passing decode/backend tests on GB300/SM103 + CUTLASS DSL 4.7.1. No additional production check is needed.

PerkzZheng:
Fixed the relevant test-eligibility issues in [5720483a](https://github.com/flashinfer-ai/flashinfer/commit/5720483a03821d37c213d10d53275d102d132108).

The sparse FMHA/workspace tests now use exact SM100/SM103 eligibility. CUDA availability alone admitted the 17 failing attention-dependent cases on H100/SM120; the common `arch_blackwell` marker also admitted Rubin because it checks only major capability 10. CPU contracts, metadata-only CUDA tests and early validation cases remain enabled. No production kernels, policies, test bodies or numerical tolerances changed.

| Validation | Result |
| --- | --- |
| Actual GB300/SM103 sparse suite | 141 passed |
| Actual GB300 decode/backend suite | 232 passed, 96 existing skips |
| Reported SM90/SM120, before fix | 17 failed, 101 passed, 23 skipped each |
| Reported SM90/SM120/SM107, after fix | 101 passed, 40 skipped each |
| CPU-only sparse suite | 36 passed, 105 GPU skips |

Capability probes execute on a real SM103 GPU with the reported capability overridden; they test guards, not numerical correctness on H100/SM120/Rubin. Before/after JUnit comparison preserves the same 141 case IDs and every existing passing case. Changed-file pre-commit hooks pass. Environment: GB300, CUDA 13.0, Torch 2.13.0+cu130, CUTLASS DSL 4.7.1.

I left the MoE autotuner and FP8 GEMM findings untouched: the report does not establish PR attribution, and this change does not modify their implementations or the dependency floor. Other pre-existing failures remain out of scope. I could not access the raw Rubin tracebacks, so this fixes eligibility for the declared SM100/SM103 sparse test scope, not an alleged Rubin numerical/compiler bug. A hosted rerun is still needed; no full-CI pass is claimed.

The other open findings were rechecked: the prepared-run docstring is complete, the old membership side channel is gone, and both page4/page32 reduction-predicate tests pass. The page-size thread has a separate evidence-backed reply; manual thread resolution remains with the reviewer.

qsang-nv:
### [P2] `out` is never validated against the route metadata inputs

`flashinfer/attention/prims_ts/q_token_kv_block_sparse_metadata.py` imports only `_validate_tensor_does_not_overlap_inputs`; `_validate_out_does_not_overlap_inputs`, in the same `_tensor_aliasing` module, is not imported or called on this path. The one guard, `_validate_q_token_kv_block_sparse_workspace_aliasing()`, takes `workspace_buffer` as its checked subject, so passing `out=out` into it only establishes `out ∩ workspace_buffer = ∅`.

The decode layer's `_validate_decode_output_aliasing()` does check `out`, but this path hands it `views.q_token_kv_block_sparse_page_indices`, `views.seq_lens` and `views.q_token_kv_block_sparse_page_memberships` — all workspace slices. So `out` is compared against `q`, `k_cache`, `v_cache`, `qo_indptr` and the workspace, but never against `indexer_block_ids`, the caller's `block_table`, `token_to_request` or `query_positions`.

A caller can therefore pass an `out` that clears every shape and dtype check while sharing storage with one of those four Int32/Int64 route tensors (e.g. via views over shared storage). The metadata kernel reads them, attention writes `out` over them, and the next prepared run or CUDA-graph replay rebuilds routes from corrupted metadata — silently wrong results rather than an error. The private `_PrimsTSQTokenKvBlockSparsePlan` contract says that replacement tensors must remain disjoint, but neither public entry point states or enforces that restriction.

PerkzZheng:
> ### [P2] `out` is never validated against the route metadata inputs

@qsang-nv I think _validate_tensor_does_not_overlap_inputs a bit over-engineering (conservative) created by codex. I would like to remove the validation entirely instead to avoid potential host overhead.

qsang-nv (approved):

PR #5002, opened 2026-09-07
perf(prims-ts): Optimize&refine PrimsTS block sparse attention

heyuhhh:
## Summary

Performance work on the PrimTS block-sparse decode attention kernel (`flashinfer/attention/prims_ts`) that follows #4872. The changes target the exact block-sparse path first (Q64/KV256 and Q128/KV128 Keeps profiles) and also carry over to the proxy-compensated routes and to the dense KV256 decode profile that shares the same kernel. Eleven commits; each performance commit was measured with paired ABBA runs on B200 against the commit before it.

### What changed

1. **Stream the KV256 P fragments from one rolled loop** (`5d8d79a2`). The per-fragment P path was fully unrolled and made the kernel instruction-cache bound. One rolled loop over the four K32 fragments, with the FMA exp2 emulation limited to 4 of 16 pairs, shrinks the code and keeps MUFU and FMA balanced. Also drops the redundant same-instance PV completion wait before QK (tcgen05 orders the A-operand read before the accumulator write). Dense KV256 shares the masked write-back path.
2. **Proxy routes use the shared scheduler selection** (`f5b4a6f1`). Proxy plans no longer force the static grid, so the CLC persistent scheduler is selected for large shapes as it is for exact plans.
3. **Trim the KV256 per-tile fixed costs** (`b1c737f7`). Full-sector output stores in the epilogue, static-grid row-header prefetch in the CTA prologue, and releasing the Q stage before the tail PV MMAs.
4. **Rebalance KV256 registers and free the K/V ring from the tail** (`02f83c00`). Softmax/correction register split 152/152 and a dedicated 11 KB SMEM exchange for the tail merge instead of a rotating credit on the K/V ring.
5. **Consolidate the KV256 fragment and tail paths and the shared dense/sparse load cadence** (`58107d1b`). Codegen-neutral (SASS byte-identical); removes an unreachable unrolled path and several duplicated helpers.
6. **Stream Q128/KV128 P as K32 fragments** (`c7b42e41`). The 16-bit two-instance Q128 profile now uses the same fragment streaming as KV256 (TMEM layout O-first with P aliased at S column 0, per-fragment barriers, PV k-slices per fragment). Includes a fix for fragment-to-origin mapping with 128-token KV blocks and a proxy test for that layout.
7. **One streamed Keeps max pass for dense and block-sparse tiles** (`37f351cc`). Deletes the unreachable sparse Q64/KV128 complete-row arm and unifies the dense and sparse fragment max pass; dense tiles derive one 32-bit keep word per fragment instead of per-element boundary compares.
8. **Prepare structural score words for dense Keeps plans** (`c2557b69`). Exact plans with a dense mask type now prepare the per-fragment keep words too, so the streamed max pass can trust them.
9. **Prefetch block-sparse route records one iteration ahead** (`aea0b1e8`). PerfSim warp timelines showed the load warp saturated per route by TMA issue plus a fully exposed global load of the next route record. The shared-KV load cadence now issues that load one resolution ahead, threading the prefetch state like the cached page IDs; the split-ring load variants keep the immediate load.
10. **Keep dense Q128/KV128 on the complete-row P path** (`ee119fe4`). Streaming the Q128 P row only pays off where the route loop waits on the K/V loads; the dense GQA-128 decode profile measured slower with it. `streams_tmem_p_fragments` is now the single policy (16-bit two-instance profiles with a KV256 tile or a block-sparse plan), and the 16-bit x16-slice publication of the complete-row path is restored. Dense Q128 is back at main's timing with identical output.
11. **Review fixes** (`679bf00f`, `f2f7cf52`). The KV256 split-KV partial store advances its column offset by the two-byte partial width instead of the final output width. No reachable configuration changes today since KV256 admits only 16-bit output, and the FP16/BF16 split-KV SASS is unchanged. The `proxy_bk128_keeps` test case now selects only in-range KV blocks and keeps the ragged tail block in its set.

### Correctness

- `tests/attention/test_attention_ts_block_sparse.py`: 228 passed (one pre-existing failure, `test_runtime_routes_cuda_graph_replays_routes_and_token_mask`, deselected; it fails on main as well).
- `tests/attention/test_attention_ts_decode.py`: 212 passed (rebased on main with #4915).
- New tests: the config truth table for streamed profiles, the alias layout of the streamed TMEM P, the kv_block 128 proxy tail case, the structural score-word preparation for dense Keeps plans, and the rejection of an unstreamed Keeps profile.

### Performance

Paired ABBA on one B200 (5 runs per leg, 3 legs, min of graph-replay time per call), block-sparse decode attention, S=10800, H=40, D=128, bf16, block 64, exact mask density 0.175 ("hot" pattern), automatic scheduler selection. Baseline is upstream `main` at 1989b509.

| Profile (auto scheduler) | main 1989b509 | this PR | paired B/A (95% CI) |
|---|---|---|---|
| Q64/KV256 exact (block-sparse, SOL shape) | 664.5 us | 521.8 us | 0.786 [0.785, 0.787] (-21.4%) |
| Q64/KV256 proxy-compensated | 705.0 us | 594.3 us | 0.842 [0.839, 0.844] (-15.8%) |
| Q128/KV128 exact (q_block 128) | 468.9 us | 426.3 us | 0.909 [0.909, 0.910] (-9.1%) |
| Q128/KV128 proxy-compensated | 861.2 us | 592.0 us | 0.688 [0.685, 0.691] (-31.2%) |

Dense decode (same kernel, `prims_ts_batch_decode_with_kv_cache`, sq=1024, kv=10800, page 128, bf16; four interleaved main/branch runs, min per call):

| Dense profile | main 1989b509 | this PR | change |
|---|---|---|---|
| Q64/KV256 Keeps, H=40 (grouped) | 323.7-324.9 us | 240.6-240.9 us | -25.7% |
| Q128/KV128 Keeps, GQA ratio 128 (H=128, 1 KV head, ungrouped) | 599.9-602.9 us | 600.6-603.0 us | neutral (kept on the complete-row P path, see item 10) |

The block-sparse SASS is byte-identical across the two codegen-neutral refactor commits and across the dense-Q128 gating commit.

## Summary by CodeRabbit

* **New Features**
  * Expanded streamed attention support across KV256 and Q128/KV128 configurations.
  * Added score-validity handling for dense block-sparse Keeps plans.
  * Added route metadata prefetching for sparse attention.
  * Enabled shared scheduling behavior for proxy and exact routes.
  * Added support for two-stage KV256 scheduling.

* **Bug Fixes**
  * Improved validation for block-sparse Keeps and proxy-route configurations.
  * Improved softmax and output handling across streamed attention paths.

### DIFF
--- a/flashinfer/attention/prims_ts/_block_sparse/compiler.py
+++ b/flashinfer/attention/prims_ts/_block_sparse/compiler.py
@@ -55,6 +55,7 @@ def _compile_block_sparse(key: _BlockSparseCompileKey) -> Callable[..., object]:
         "use_proxy_routes": key.use_proxy_routes,
         "use_causal_mask": key.mask_type == "causal",
         "apply_token_mask": key.use_kv_valid_bits,
+        "store_score_words": config.uses_prepared_score_keep_words,
     }
     if key.page_size is not None:
         if key.sparse_format != "bsr" or key.use_proxy_routes:
--- a/flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/smem_block_sparse_metadata.py
+++ b/flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/smem_block_sparse_metadata.py
@@ -249,6 +249,18 @@ class SmemBlockSparseKvMetadataResource(DecodeGenResourceBase):
             Int32(-1),
             "Metadata-relative record offset, or -1 for a dummy route.",
         ),
+        (
+            "prefetched_record_word_slot",
+            Int32,
+            Int32(0),
+            "Lane-owned record word loaded one resolution ahead.",
+        ),
+        (
+            "prefetched_record_offset_slot",
+            Int32,
+            Int32(-1),
+            "Record offset of the prefetched route, or -1 for a dummy route.",
+        ),
     )
     cfg: Constexpr[FmhaDecodeConfig] = None
     inst_id: Constexpr[int] = 0
@@ -268,6 +280,12 @@ class SmemBlockSparseKvMetadataResource(DecodeGenResourceBase):
         TaskLocalVariable.uninitialized()
     )
     route_record_word_offset_slot: Constexpr[TaskLocalVariable] = (
+        TaskLocalVariable.uninitialized()
+    )
+    prefetched_record_word_slot: Constexpr[TaskLocalVariable] = (
+        TaskLocalVariable.uninitialized()
+    )
+    prefetched_record_offset_slot: Constexpr[TaskLocalVariable] = (
         TaskLocalVariable.uninitialized()
     )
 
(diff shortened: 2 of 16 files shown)


flashinfer/attention/prims_ts/_block_sparse/compiler.py
```diff
@@ -55,6 +55,7 @@ def _compile_block_sparse(key: _BlockSparseCompileKey) -> Callable[..., object]:
         "use_proxy_routes": key.use_proxy_routes,
         "use_causal_mask": key.mask_type == "causal",
         "apply_token_mask": key.use_kv_valid_bits,
+        "store_score_words": config.uses_prepared_score_keep_words,
```
PerkzZheng, line 58: what does this mean ?

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/tmem_corr.py
```diff
@@ -2329,6 +2396,97 @@ def _kv_tile_256_temporal_fragment(
...
+                    )
+                scaled_vector = cutlass.Vector.from_elements(scaled_values, Float32)
+                if cutlass.const_expr(partial_o_uses_bf16):
+                    packed = scaled_vector.to(cutlass.BFloat16).bitcast(Int32)
+                else:
+                    packed = scaled_vector.to(cutlass.Float16).bitcast(Int32)
+                partial_o_dst = cutlass.inttoptr(
+                    self.partial_o_ptr.toint()
+                    + partial_row_base
+                    + Int64(output_col * cfg.o_dtype_bytes),
```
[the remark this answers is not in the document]
PerkzZheng: This makes sense.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/smem_block_sparse_metadata.py
```diff
@@ -359,6 +378,78 @@ def _prepared_route_physical_page_id_if_valid(
...
+                self.inst_id
+            )
+        else:
+            route_idx = (stage_info.loop_offset + Int32(2)) * num_insts + Int32(
+                self.inst_id
+            )
+        route_record_word_offset = self._route_record_word_offset(stage_info, route_idx)
+        record_word = Int32(0)
+        if cutlass.const_expr(self.route_layout.uses_one_warp_transport):
+            assert self.route_layout.token_words_word_offset is not None
```
PerkzZheng, line 437: what does token_words_word_offset mean here ? does it mean the offset will bring misaligned offset (in terms of kv tile size) ?

qsang-nv:
**`proxy_bk128_keeps` selects an out-of-range block index, invalidating its intended coverage.**

`tests/attention/test_attention_ts_block_sparse.py` (~L570-584, with the index generated at ~L639-642): this case sets `seq_len_kv=269` and `kv_block_size=128`, so `num_kv_blocks == 3` and the valid block indices are 0..2. The shared `proxy_tail` pattern takes its `num_kv_blocks <= 32` branch and produces `(1, 3)`.

On the BSR path, `_PrepareBsrRoutes` validates each row against `cfg.num_kv_blocks` through `_validate_bsr_row_lane` and rejects index 3 with `runtime_assert(..., "block_indices row must be canonical and in range")`, so this case cannot successfully complete its BSR run. On the bitmask path, bit 3 lies in the padding region and is masked during route preparation, effectively reducing the exact selection to `{1}`.

Consequently, the test cannot compare the same block set through both frontends or validate the intended KV128 atom-to-fragment mapping end to end.

flashinfer/attention/prims_ts/_block_sparse/compiler.py
heyuhhh: It means we have prepared mask words before attention kernel so kernel can just read accroding mask word to do masking with no need to resolve this on the fly.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/smem_block_sparse_metadata.py
heyuhhh: It means the offset of token mask word as mentioned above that a word with 32 bits can represent the mask of 32 tokens. Actually, for block sparse attention we have a prepare kernel can prepare kernel level informations before core attention, so we should the offsets to predict the right slots.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/tmem_corr.py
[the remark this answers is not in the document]
heyuhhh: Thanks, i'll refine here. But it's a fake case because we can't support fp8 outputs for now in block sparse.

heyuhhh:
Thanks @qsang-nv , please see commit f2f7cf528e1427cfc1c29798cff95f5cb0bfc50d, now for 3 kv blocks there will be {1, 2} blocks to compute

qsang-nv (approved):

PR #5013, opened 2026-09-07
fix(activation): pick a vector width that divides hidden_size in act_and_mul (fixes d=3420)

### LINKED ISSUE #2526: silu_and_mul: qwen2.5-vl VIT intermediate_size=3420, but vec_size in flashinfer is allways 16
reproduce by replace dim with 3420
https://github.com/flashinfer-ai/flashinfer/blob/f84ac1c97e2e1a4391150ca73971b973ee569e5a/tests/utils/test_activation.py#L40

Also you can refer to:
https://github.com/alibaba/rtp-llm/blob/main/3rdparty/flashinfer/0010-silu-mul-vec-size.patch

Wint3rNight:
## 📌 Description

`act_and_mul_kernel` hardcodes `vec_size = 16 / sizeof(T)` (8 for fp16/bf16) and reads the second half of each row at element offset `d` and writes the output row at `token_idx * d`. When `d` is not a multiple of 8 those addresses are not 16-byte aligned and the vectorized loads fault. Qwen2.5-VL's vision MLP has `intermediate_size = 3420`: the y half starts at byte 6840, which is only 8-byte aligned, and `silu_and_mul` dies with `cudaErrorMisalignedAddress` in both fp16 and bf16.

The Python guard did not catch it because it checks the full row: `input.shape[-1] * itemsize % 16`, i.e. `4d % 16`, which passes for any `d % 4 == 0`. So `d = 1234` was rejected cleanly while `d = 3420` reached the kernel and crashed.

This PR:

- makes `vec_size` a template parameter of `act_and_mul_kernel`, and has the launcher pick the largest power of two that divides `d` (8, 4, 2 or 1). With `vec_size | d` every access (x half, y half at `+d`, output row at `token_idx * d`) is aligned to `vec_size * sizeof(T)`. Power-of-two and multiple-of-8 dims keep the full width, so common models are untouched.
- relaxes the guard in `silu_and_mul`, `gelu_and_mul` and `gelu_tanh_and_mul` to "last dimension must be even".
- removes the kernel's tail loop. With `vec_size | d` the grid-stride main loop already covers every element, so the tail can only re-process elements. Since #4733 capped `blockDim` at 256, its bound `d % (blockDim * vec_size)` no longer equals `d % vec_size`, so on main it re-computes up to 2047 already-written elements per row with scalar accesses (768 for `d = 11008`). Output was correct, the work was wasted.

Same design as #2613 (approved by @yzh119 in February, never rebased and now conflicting), redone on current main.

## 🔍 Related Issues

Fixes #2526. Supersedes #2613.

### ✅ Pre-commit Checks

## 🧪 Tests

`tests/utils/test_activation.py` gains `855` (odd, `vec_size = 1`), `1710` (`vec_size = 2`) and `3420` (`vec_size = 4`) in all three dim lists.

On an RTX 3050 Laptop (sm_86, CUDA 13.3, torch 2.14.0+cu130):

```
pytest tests/utils/test_activation.py -k "855 or 1710 or 3420 or 11008"
540 passed, 540 skipped (enable_pdl=True needs sm_90), 1620 deselected
```

Before the patch, `silu_and_mul` on `(4, 2*3420)` raises `CUDA error: misaligned address` in fp16 and bf16; after it matches the torch reference.

Throughput (fp16, 4096 tokens, median of 5x200 iterations, kernel is memory-bound at ~179 GB/s on this GPU):

| d | before (us) | after (us) |
| --- | ---: | ---: |
| 3416 | 482.2 | 468.8 |
| 4096 | 561.0 | 560.3 |
| 8192 | 1118.9 | 1119.8 |
| 11008 | 1517.1 | 1504.0 |
| 12288 | 1680.2 | 1679.4 |
| 16384 | 2240.4 | 2239.9 |

The small gains at 3416 and 11008 are the removed duplicate tail; the rest is noise.

## Reviewer Notes

- Only fp16/bf16 go through `DISPATCH_DLPACK_DTYPE_TO_CTYPE_FP16`, so the maximum `vec_size` is 8 and the dispatch has four cases.
- Base pointers are still assumed 16-byte aligned, as before. Offset views (e.g. `x[:, 1:]`) are out of scope here.

## Summary by CodeRabbit

- **New Features**
  - Activation-and-mul operations now support any input shape with an even final dimension, including dimensions that were previously rejected.
  - Improved handling of varying activation dimensions enables broader compatibility across supported inputs.

- **Tests**
  - Expanded coverage for fused SiLU, GELU-tanh, and GELU activation-and-mul operations with additional nonstandard dimensions.

### DIFF
--- a/flashinfer/activation.py
+++ b/flashinfer/activation.py
@@ -63,6 +63,14 @@ def get_act_and_mul_module(act_func_name: str):
     return SimpleNamespace(**{fname: _act_and_mul})
 
 
+def _check_act_and_mul_input(input: torch.Tensor) -> None:
+    # The kernel picks a vector width that divides hidden_size, so any even
+    # last dimension (2 * hidden_size) is supported; it no longer has to be a
+    # multiple of 16 bytes.
+    if input.shape[-1] % 2 != 0:
+        raise ValueError("The last dimension must be even (2 * hidden_size).")
+
+
 def _check_shape(input: torch.Tensor, output: torch.Tensor) -> None:
     assert input.ndim == output.ndim, f"{input.ndim} != {output.ndim}"
     assert input.shape[:-1] == output.shape[:-1], (
@@ -100,8 +108,7 @@ def silu_and_mul(
     """
     if enable_pdl is None:
         enable_pdl = device_support_pdl(input.device)
-    if input.shape[-1] * input.dtype.itemsize % 16 != 0:
-        raise ValueError("The pointers must be multiple of 16 bytes.")
+    _check_act_and_mul_input(input)
     if out is not None:
         _check_shape(input, out)
     else:
@@ -145,8 +152,7 @@ def gelu_tanh_and_mul(
     """
     if enable_pdl is None:
         enable_pdl = device_support_pdl(input.device)
-    if input.shape[-1] * input.dtype.itemsize % 16 != 0:
-        raise ValueError("The pointers must be multiple of 16 bytes.")
+    _check_act_and_mul_input(input)
     if out is not None:
         _check_shape(input, out)
     else:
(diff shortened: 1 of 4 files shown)


bkryu:
@flashinfer-bot run

bkryu (approved):
Thanks @Wint3rNight  for the fix!

PR #5016, opened 2026-09-07
Nccl extensions dependency update

Anerudhan:
## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## 🔬 Experimental Track

```experimental-tests
#
#
```

## Summary by CodeRabbit

- **New Features**
  - Added support for the `nccl-extensions` package as the provider of NCCL-EP functionality.
  - Improved runtime library selection across CUDA environments and ensured the correct NCCL library version is loaded.
  - Added NCCL-EP package version reporting.

- **Bug Fixes**
  - NCCL version checks now prioritize the actively loaded library, with metadata fallback when unavailable.
  - Updated container builds and installation scripts to use compatible NCCL-EP packages and runtime libraries.

- **Documentation**
  - Updated setup guidance, dependency references, architecture notes, and troubleshooting messages to reflect the new package split.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,9 +1,25 @@
 apache-tvm-ffi>=0.1.6,!=0.1.8,!=0.1.8.post0,<0.2
 click
-# cuda-python + nccl4py: runtime deps for the moe_ep EP transport backends
-# (default since the EP install became opt-out; CUDA-13 wheels only — see
-# build_backend.py). NCCL-EP is the released nccl4py wheel (`nccl.ep` API +
-# bundled libnccl_ep.so); no in-tree NCCL build.
+# cuda-python + nccl-extensions: runtime deps for the moe_ep EP transport
+# backends (default since the EP install became opt-out; CUDA-13 wheels only
+# — see build_backend.py). NCCL-EP is the released nccl-extensions wheel
+# (`nccl.ep` API + bundled libnccl_ep.so); no in-tree NCCL build.
+#
+# nccl.ep MOVED: it used to ship inside nccl4py, but the EP/M2N libraries were
+# split out to https://github.com/NVIDIA/nccl-extensions and nccl4py 0.4.1
+# dropped `nccl/ep` entirely. Depending on nccl4py alone therefore resolves to
+# a wheel with no nccl.ep at all. Both distributions install into the shared
+# `nccl` namespace package: nccl4py owns nccl/core + nccl/bindings (we still
+# need `nccl.core.Communicator`, and nccl-extensions pulls it in transitively),
+# nccl-extensions owns nccl/ep + nccl/m2n + nccl/_extensions. The explicit
+# nccl4py>=0.4.1 floor is a *conflict* guard, not a feature floor: nccl4py
+# 0.3.1 still shipped its own nccl/ep into that same namespace directory, so
+# co-installing it with nccl-extensions overwrites files and yields an
+# ambiguous nccl.ep.
+#
+# The nccl-extensions cu12/cu13 extras are deliberately NOT used here: they
+# pin nvidia-nccl-cuXX==2.30.7 exactly, which trips the torch-eviction problem
+# described below. Plain `nccl-extensions` pulls no nvidia-nccl wheel.
 #
 # nvidia-nccl-cu13 is deliberately NOT a base dep: torch's cu13 wheels pin it
 # EXACTLY (e.g. ==2.29.7), so any floor here (>=2.30.7 for B200 EP) makes the
@@ -25,7 +41,8 @@ click
 cuda-python>=12.0
 cuda-tile>=1.4.0
 einops
-nccl4py>=0.3.1
+nccl-extensions>=0.1.0
+nccl4py>=0.4.1
 ninja
 numpy
 nvidia-cudnn-frontend>=1.25.0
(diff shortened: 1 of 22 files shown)


Anerudhan:
## Multi-GPU verification on 4× B200 (sm100)

Ran the full `tests/moe_ep/run_tests.sh` matrix against this branch (`cfad91d7`) on a Pre-Nyx B200 node, in `nvcr.io/nvidia/pytorch:26.05-py3` (py3.12, torch 2.12, CUDA 13.2), built via `docker/install/build_flashinfer_ep_pytorch.sh`.

**All 9 sections pass, 0 failures:**

| Section | Result |
| --- | --- |
| unit + mock | PASS (433 passed, 72 skipped) |
| torch-oracle correctness (1 GPU) | PASS |
| split-path multirank (NCCL-EP) | PASS |
| split_path_correctness_bf16 (4 GPU) | PASS |
| split_path_correctness_nvfp4 (4 GPU) | PASS |
| split_path_correctness_ht (4 GPU) | PASS |
| mega multirank (Blackwell) | PASS |
| smoke scripts | PASS |
| fault tolerance (4 GPU) | PASS |

`BUILD_RC=0 ALL_RC=0 FT_RC=0`. Note `run_all` does not include `ft`, so it was run separately.

CUDA-graph coverage specifically (all PASS):

- `test_moe_ep_cudagraph_multirank.py::test_round_trip_is_capturable_with_a_persistent_handle[low_latency]` and `[high_throughput]`
- `test_mega_cuda_graph.py::test_mega_layer_graph_capture_replay_matches_eager[{1600-832,2048-1024}-{mxfp8,nvfp4}]`
- `test_mega_cuda_graph.py::test_mega_layer_multi_size_graphs_and_eager_interleave`, `test_mega_layer_capture_without_warmup_raises`

Fault tolerance genuinely exercised the mask path — the run logs `nccl_ep FT OK (survivor rank 0, mask=[1, 1, 0, 1])`.

## Why the dependency move alone was not sufficient

Two separate problems; the first motivates the PR, the second was found only by running on real multi-GPU hardware.

**1. `nccl4py>=0.3.1` no longer provides `nccl.ep` at all.** nccl4py **0.4.1 dropped `nccl/ep`** when EP/M2N split out to [NVIDIA/nccl-extensions](https://github.com/NVIDIA/nccl-extensions). `uv pip compile` on today's `requirements.txt` resolves nccl4py **0.5.0**, which ships no `nccl.ep` — so a fresh install silently loses the NCCL-EP backend. Both distributions share the `nccl` namespace package: nccl4py keeps `nccl/core` (still needed for `nccl.core.Communicator`, pulled in transitively), nccl-extensions owns `nccl/ep`. The added `nccl4py>=0.4.1` floor is a *conflict* guard — 0.3.1 shipped its own `nccl/ep` into the same directory.

**2. nccl-extensions bundles a newer `libnccl_ep` (0.1.0 → 0.2.0) that requires NCCL 2.30.7.** With only the dependency swap, every NCCL-EP collective aborted:

```
NCCL WARN NCCL library is too old. This application was compiled with
NCCL version 2.30.7, but is running with NCCL library version 2.30.4.
NCCL error 5 at nccl_ep.cc:1962          # ncclInvalidUsage
```

Installing `nvidia-nccl-cu13==2.30.7` is not enough: NGC images ship a system `/usr/lib/x86_64-linux-gnu/libnccl.so.2.30.4`, the linker finds it first, and torch loads it before flashinfer is imported — so the backend's existing `RTLD_GLOBAL` preload runs far too late to displace it. The old `libnccl_ep` 0.1.0 tolerated 2.30.4, which is why this only surfaced with the package move. Confirmed by A/B in the same container: `main` + nccl4py 0.3.1 passes, this branch without the fix fails.

Fixed by registering the wheel's lib dir with **ldconfig** (`/etc/ld.so.conf.d` + `ldconfig`) rather than `LD_LIBRARY_PATH` — an `export` in the build script dies with that subprocess and never reaches the test phase, and `/etc/profile.d` is login-shells-only. The build now asserts the loaded version twice, once under `env -u LD_LIBRARY_PATH`, to prove the pin holds for processes that don't inherit the build environment.

Also inverted `_installed_nccl_version()`: it preferred the **pip metadata** over `ncclGetVersion`, so the existing Blackwell ≥2.30.7 guard read "2.30.7" while 2.30.4 was actually loaded, and `libnccl_ep` then aborted the process before any Python exception could be raised (which is why the original failure had no traceback). It now probes the loaded library first and falls back to metadata. `test_installed_nccl_version_prefers_wheel_metadata` asserted the old precedence and is replaced by two tests covering the new contract.

## Not covered by this run

- **aarch64 / Grace (GB200).** x86_64 B200 only. `nccl-extensions` 0.1.0 does publish `manylinux_2_28_aarch64` wheels for cp310–cp314, but that path is unverified here.
- **Hopper / consumer-Blackwell sections** — `oracle_sm90`, `mega_sm90`, `sm90_push`, `mega_sm120` do not apply on sm100 and were not run.
- The ldconfig pin is applied in `build_flashinfer_ep_pytorch.sh` and `Dockerfile.flashinfer-ep-pytorch`. Other entry points that build their own environment (`scripts/build_in_container.sh`, `Dockerfile.flashinfer-nvep`) may need the same treatment — I could not test those paths.

Worth noting for reviewers: upstream CI cannot catch the class of bug fixed here — the runners are T4/A10G/H100 and the NCCL-EP suites need ≥4 GPUs, so a green CI on this PR does not exercise the EP transport on Blackwell.

requirements.txt
```diff
@@ -25,7 +41,8 @@ click
 cuda-python>=12.0
 cuda-tile>=1.4.0
 einops
-nccl4py>=0.3.1
+nccl-extensions>=0.1.0
+nccl4py>=0.4.1
```
aleozlx, line 45: is this intentional?

Anerudhan: Yes. That is the goal of the change.
nccl-ep is now shipped in nccl-extensions instead of nccl4py.

aleozlx (approved):

Issue #4749, opened 2026-08-26
[Bug] NVFP4 split-KV corruption root cause is unconfirmed; does not reproduce on SM80


lesj0610:
## Summary

`_nvfp4_kv_requires_disabled_split_kv()` disables split-KV (flash-decoding) for NVFP4 KV to work around output corruption whose root cause is, by the gate's own docstring, unconfirmed. This issue is to keep that defect tracked, and to contribute a negative result that narrows where it can live.

The gate was added in 8f9ad200 while bringing up asymmetric VO-split NVFP4 paged prefill for Gemma-4 on SM120/121, and the corruption was reported there: a short query attending a long KV range (`qo_len << kv_len`, i.e. decode and prefix-cache extend) produced wrong output, while dense full-prefill was fine.

## Evidence: it does not reproduce on sm_80

Same generic 1-byte-KV FA2 path, sm_80, flashinfer `3bbfeba6` (0.6.18). For each shape the paged prefill wrapper was run twice — gate on (split-KV forced off) and gate off (split-KV allowed) — and both scored against a dense reference from the dequantized KV:

| qo_len | kv_len | gate on, rel err | gate off, rel err |
|---|---|---|---|
| 1 | 8192 | 7.06e-04 | 7.06e-04 |
| 16 | 8192 | 8.40e-04 | 8.40e-04 |
| 16 | 4096 | 4.07e-04 | 8.14e-04 |
| 64 | 8192 | 7.32e-04 | 7.32e-04 |
| 256 | 8192 | 6.69e-04 | 6.69e-04 |
| 2048 | 2048 | 0.0 | 0.0 |

`qo_len=1` against `kv_len=8192` is the exact regime the gate exists for, and enabling split-KV there changes nothing. The residual ~7e-4 is NVFP4 quantization granularity — identical with and without the gate — not corruption. `2048x2048` is dense full-prefill, exact either way, consistent with the original report.

## What that rules out, and what it does not

The docstring's leading hypothesis is that small per-split KV chunks interact badly with the `NUM_MMA_KV` tile floor of the 1-byte-KV FA2 path. sm_80 runs that same path with the same tile floor and does not fail, so if the hypothesis holds, the mechanism has to be something SM120/121-specific layered on it rather than the tile floor alone. The obvious candidate is the asymmetric VO-split path introduced in the same commit, which pre-SM100 does not take.

[...]

PR #5018, opened 2026-09-07
test(attention): exercise the NVFP4 split-KV gated path on hardware

Champollion9012:
## 📌 Description

`_nvfp4_kv_requires_disabled_split_kv` disables split-KV for NVFP4 KV as an empirical
workaround whose root cause is, by its own docstring, unconfirmed. #4747 scoped it to
`_NVFP4_SPLIT_KV_BROKEN_ARCHS = ((12, 0), (12, 1))`, so SM120/121 are now the only architectures
carrying it.

Nothing currently runs the configuration the gate disables. The two tests that cover it,
`test_nvfp4_split_kv_gate_dtype_logic` and `test_nvfp4_split_kv_gate_arch_scope`, are a dtype
classification check and an architecture-scoping check with a faked compute capability; both run
without a GPU, and `disable_split_kv` does not appear in that file. So on the parts that pay for the
workaround its premise is untestable today: the gate could stop being a no-op, or stop being
necessary, and CI would not notice either way.

This adds one test to `tests/attention/test_batch_prefill_kernels.py`, next to the NVFP4 paged
prefill tests whose helpers it reuses, plus `_PLAN_INFO_SPLIT_KV_IDX` alongside the existing
`_PLAN_INFO_CTA_TILE_Q_IDX`. It runs the same inputs twice through the FA2 NVFP4 paged prefill entry
point -- once with split-KV force-disabled as the gate does today, once with the gate bypassed via
`monkeypatch` exactly as `test_nvfp4_split_kv_gate_arch_scope` already does -- and pins:

- the gated run does not select split-KV and the bypassed run does, both read from `PrefillPlanInfo`
  rather than assumed;
- both runs are within the tolerance the neighbouring NVFP4 tests use of a float32 dequantization
  oracle built with `nvfp4_to_float` from the exact quantized bytes the kernel reads, so it is a
  dequantization reference and not a requantized approximation;
- allowing split-KV does not move the result more than 2x further from that oracle than the gated
  run already is. Split-KV changes the softmax reduction order, so some drift is expected; the
  largest split/non-split ratio reported on an affected part in #4749 is about 1.4x, so 2x leaves
  room for reduction-order drift without admitting the corruption the gate exists for.

Parametrized over symmetric `head_dim` 128 and the asymmetric (512, 256) shape the gate was added
alongside in 8f9ad200, at `qo_len` 1 and 16 against long KV, which is the `qo_len << kv_len` regime
the docstring names. It self-skips where the gate is not in force, so it is a no-op on every
architecture #4747 exempted.

The gate is not left unexercised for lack of hardware in CI: the internal matrix already has
`unit_test_rtx_pro_6000` and `unit_test_5090` at cc 12.0 and `unit_test_spark` at cc 12.1.

## 🔍 Related Issues

- #4749 -- the open issue tracking the unconfirmed root cause. This PR does not change the gate;
  whether the workaround should be narrowed further or removed belongs there.
- #4747 -- scoped the gate to `((12, 0), (12, 1))` and added the two no-GPU gate tests this one
  complements.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## 🔬 Experimental Track

```experimental-tests
#
#
```

## Reviewer Notes

Validated on hardware the gate is still in force on, so the test is exercised rather than skipped:

```
GPU     NVIDIA RTX PRO 6000 Blackwell Server Edition, cc 12.0, driver 590.48.01
CUDA    13.1, torch 2.13.0+cu130, flashinfer main @ 91bda04

tests/attention/test_batch_prefill_kernels.py -k nvfp4_paged_split_kv_matches_gated_result
    4 passed in 615s
tests/attention/test_batch_prefill_kernels.py -k "nvfp4_asymmetric or cta_tile_q_smem_probe or unequal_kv_strides"
    14 passed in 129s
tests/attention/test_nvfp4_attention_sm120.py -k split_kv_gate
    2 passed in 0.07s
pre-commit run --all-files
    all hooks passed, no file rewritten
```

Most of the 615 s is JIT compilation: bf16 query with NVFP4 KV at these head dims is outside the
module set the file's pre-generation fixture builds, the same as the neighbouring
`test_batch_prefill_with_paged_kv_cache_nvfp4_asymmetric`. Happy to trim the parametrization if that
is too much for the cc 12.0 lanes.

Two choices I would rather have questioned than left implicit:

1. **The split-enabled plan is asserted, not skipped.** `batch_size` 1 with few heads against a long
   KV range is the only regime where split-KV is what supplies the parallelism, so if the planner
   stops splitting there the gate has become a no-op. That seemed worth a failure rather than a
   silently vacuous pass, but it is a judgement call and is easy to soften to a `pytest.skip` if you
   would rather the test never fail for a planner change.
2. **The 2x factor in the oracle-distance check** is the one tuned constant. It is scale-free and
   grounded in the ~1.4x split/non-split ratio reported in #4749, but if you would prefer a
   different bound, or a comparison against the gated run's own drift instead of the oracle's, say
   which and I will change it.

## Summary by CodeRabbit

- **Tests**
  - Added coverage for NVFP4 paged prefill with split-KV enabled and disabled.
  - Validated both execution paths against a float32 reference and against each other.
  - Added checks to confirm split-KV planning occurs and remains within the expected accuracy tolerance.
  - Included coverage across supported device architectures and multiple attention head-dimension configurations.
  - Added test setup to prebuild the required kernels, improving consistency and reliability during validation.

### DIFF
--- a/tests/attention/test_batch_prefill_kernels.py
+++ b/tests/attention/test_batch_prefill_kernels.py
@@ -1748,6 +1748,7 @@ def test_batch_prefill_with_paged_kv_cache_nvfp4_asymmetric(
 
 
 _PLAN_INFO_CTA_TILE_Q_IDX = 3  # PrefillPlanInfo::ToVector layout (scheduler.cuh)
+_PLAN_INFO_SPLIT_KV_IDX = 14  # PrefillPlanInfo::ToVector layout (scheduler.cuh)
 
 
 @pytest.mark.parametrize("kv_dtype", [torch.float16, torch.float8_e4m3fn])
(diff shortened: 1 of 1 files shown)


Champollion9012:
I verified this test on real Blackwell hardware, since the public CI here can't run the GPU path.

**Environment:** NVIDIA RTX PRO 6000 Blackwell (compute capability 12.0), torch 2.13.0+cu130 / CUDA 13.0. `_NVFP4_SPLIT_KV_BROKEN_ARCHS = ((12, 0), (12, 1))` is in force on this arch, so the test executes rather than skipping. The `dtype_q_bf16 / dtype_kv_fp4x2_e2m1 / dtype_o_bf16` prefill kernels JIT-build and run for head dims (128, 128) and (512, 256).

```
pytest tests/attention/test_batch_prefill_kernels.py::test_nvfp4_paged_split_kv_matches_gated_result -v
  [1-8192-128-128]   PASSED
  [1-8192-512-256]   PASSED
  [16-4096-128-128]  PASSED
  [16-4096-512-256]  PASSED
  4 passed in 349.05s
```

The gated (split-KV disabled) and split-KV-selected paths produce matching outputs across all four shapes.

One note: a `[jit-prebuild]` warning fires because the two new NVFP4 modules aren't registered in `tests/attention/conftest.py`'s prebuild collector, so they compile at test time. I can register them there if that's preferred.

Could a maintainer kick off the internal GPU CI (`/bot run`) when convenient? The matrix already covers `unit_test_rtx_pro_6000` and `unit_test_b200`, which is where this path is exercised.

PR #5022, opened 2026-09-08
Support compact GLM NoPE FP8 rows and fix masked reads

ormandj:
## 📌 Description

GLM NoPE cache rows contain 512 FP8 values and four FP32 scales. The 656-byte layout also reserves 128 bytes for RoPE values that GLM NoPE does not use. Support 528-byte rows and advertise that capability to callers without changing the stored values or scales.

Masked sparse candidates currently read mutable cache slot zero. If that slot contains NaNs, zero attention weights do not prevent `0 * NaN` from contaminating valid outputs. Make masked candidates read a dedicated zero row and apply explicit length bounds before prefill gathers. Add an eight-head decoder that matches the existing eight-row scratch layout.

Compact rows require kernels that understand this format and explicit three- or four-dimensional row geometry. The flat two-dimensional API continues to use the 656-byte layout, and existing 656-byte callers remain supported. This format applies to GLM NoPE; models that use the reserved RoPE values retain their existing layout.

## 🔍 Related Issues

This extends the native GLM NoPE backend merged in #4802. The companion [sgl-project/sglang#38430](https://github.com/sgl-project/sglang/pull/38430) selects compact rows only when this capability is advertised and requires the masked-read correction for padded sparse candidates.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

Configured pre-commit hooks, including clang-format and Ruff, pass on the changed files. The full-repository hook run is not claimed.

## 🧪 Tests

Author CPU validation at `a47c44ec32`: 35 dispatch tests passed with CUDA hidden. This update removes the model-specific contract-version field; the CUDA/C++ sources and GPU tests are unchanged from `3b3ef40730`.

Earlier author validation at `3b3ef40730` on main `9c10d2e8f2` on one RTX PRO 6000 Blackwell Max-Q (SM120): 605 tests passed. Targets: `tests/attention/test_sparse_mla_sm120_dispatch.py`, `tests/attention/test_sparse_mla_sm120.py`. The dispatch cases run on CPU; sparse-attention cases run on the GPU. The sanitizer and adapter results below retain their earlier source scope.

Historical author-run results on RTX PRO 6000 Blackwell SM120, using FlashInfer source `c35c4d12a6` and SGLang adapter source `6ccd5e37e4`:

| Test scope | Result |
|---|---|
| `tests/attention/test_sparse_mla_sm120.py` | 570 cases passed |
| Compact versus 656-byte output and LSE comparisons, included above | 18 cases passed bitwise |
| Masked-input tests, included above | 64 cases passed |
| Compute Sanitizer on the 64 masked-input cases plus 6 SGLang adapter cases | 70 cases passed; 0 errors |

Before the masked-read correction, six targeted tests with NaNs in unused cache slot zero produced NaNs in valid outputs. The masked-input tests cover partial tiles, holes, explicit length bounds, empty rows and CUDA graph replay with changed inputs. The adapter cases exercise both row layouts with 1/4/65 tokens and 8/32 heads.

```bash
python -m pytest -q tests/attention/test_sparse_mla_sm120.py
compute-sanitizer --tool memcheck --error-exitcode 99 --target-processes all python -m pytest -q tests/attention/test_sparse_mla_sm120.py::test_glm53_masked_cache_rows_ignore_poisoned_slot_zero
```

The sanitizer command above runs the 64 in-repository masked-input cases. The six adapter cases are additional cross-library validation. These kernel and adapter tests do not establish whole-model accuracy or a serving speedup.

Per-row storage decreases from 656 to 528 bytes, a 128-byte reduction (19.5%). Total storage saved is `128 bytes × cached token slots × affected layers`, before allocator padding. The FP8 values and FP32 scales are unchanged; reduced storage alone does not establish faster attention.

## 🔬 Experimental Track

```experimental-tests
#
#
```

## Reviewer Notes

Developed with AI assistance.

## Summary by CodeRabbit

- **New Features**
  - Added support for compact 528-byte-per-token GLM53 NoPE KV-cache layouts across 3D and 4D cache formats.
  - Added compatibility with compact and legacy cache layouts during sparse MLA decoding and prefill.
  - Added dedicated handling for the GLM53 NoPE eight-head configuration.

- **Bug Fixes**
  - Prevented invalid or out-of-range cache indices from reading unintended data.
  - Improved masked-cache handling for stable, finite outputs.

- **Tests**
  - Added coverage for compact caches, masked entries, CUDA graph replay, and cache-layout compatibility.

### DIFF
--- a/csrc/sparse_mla_sm120.cu
+++ b/csrc/sparse_mla_sm120.cu
@@ -184,15 +184,15 @@ void SparseMlaSm120PagedAttention(
   const int d_qk = static_cast<int>(q.size(2));
   const int topk = check_dense_indices_2d_or_s_q_3d(indices, "indices", num_tokens);
   const ModelType mt = resolve_model_type(d_qk, model_type);
-  const PagedKVLayout kv_layout = parse_paged_kv_layout(kv_cache, bytes_per_token(mt), "kv_cache");
+  const int bpt = bytes_per_token(mt, kv_cache.ndim() >= 3 ? kv_cache.size(-1) : 0);
+  const PagedKVLayout kv_layout = parse_paged_kv_layout(kv_cache, bpt, "kv_cache");
   const int page_block_size = kv_layout.page_block_size;
   // Inline-scale models (DSV3_2 / GLM_NSA / GLM53_NOPE) are addressed by the
   // prefill kernels as a flat token array (prefill_kv_entry_base), so a
   // padded block stride would be silently misread; only footer-scale models
   // honor stride_kv_block. Padded strides remain a decode-path capability.
   if (mt == ModelType::DSV3_2 || mt == ModelType::GLM_NSA || mt == ModelType::GLM53_NOPE) {
-    TVM_FFI_ICHECK_EQ(kv_layout.stride_kv_block,
-                      static_cast<size_t>(page_block_size) * bytes_per_token(mt))
+    TVM_FFI_ICHECK_EQ(kv_layout.stride_kv_block, static_cast<size_t>(page_block_size) * bpt)
         << "prefill for inline-scale KV caches (DSv3.2/GLM) requires densely packed blocks "
            "(stride_kv_block == page_block_size * bytes_per_token); padded block strides are "
            "decode-only";
--- a/csrc/sparse_mla_sm120_decode_dsv3_2.cu
+++ b/csrc/sparse_mla_sm120_decode_dsv3_2.cu
@@ -189,6 +189,9 @@ bool launch_sparse_mla_decode_dsv3_2(ModelType mt, int num_heads, int topk, int 
   // indexer window. The TP1 (64-head) and TP2 (32-head) shapes keep
   // dedicated instantiations; any other shard rides the runtime-H fallback.
   if (mt == ModelType::GLM53_NOPE) {
+    // The public scratch allocator uses eight rows for H=8. A dedicated
+    // instantiation preserves that ABI instead of the runtime-H padded stride.
+    DSV3_2_DISPATCH_MT(ModelType::GLM53_NOPE, 8)
     DSV3_2_DISPATCH_MT(ModelType::GLM53_NOPE, 32)
     DSV3_2_DISPATCH_MT(ModelType::GLM53_NOPE, 64)
     DSV3_2_DISPATCH_RT_MT(ModelType::GLM53_NOPE)
--- a/csrc/sparse_mla_sm120_jit_binding.cu
+++ b/csrc/sparse_mla_sm120_jit_binding.cu
@@ -263,7 +263,8 @@ void SparseMlaSm120DecodeDsv3_2(TensorView q, TensorView kv_cache, TensorView in
       << "decode-v32 expects DSV3_2/GLM_NSA d_qk=576 or GLM53_NOPE d_qk=512; got d_qk=" << d_qk
       << " model_type=" << model_type;
 
-  const PagedKVLayout kv_layout = parse_paged_kv_layout(kv_cache, bytes_per_token(mt), "kv_cache");
+  const int bpt = bytes_per_token(mt, kv_cache.ndim() >= 3 ? kv_cache.size(-1) : 0);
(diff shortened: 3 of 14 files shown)


lucifer1004:
It is not good to add arguments like `glm53_nope_contract_version` here.

ormandj:
Thanks for the feedback, Gabriel.

Removed `glm53_nope_contract_version` from the shared configuration and updated the companion SGLang PR to use the existing compact-row capability. The complete FlashInfer patch remains required, including the masked-read and eight-head fixes.

lucifer1004:
Hi @ormandj — thank you for this PR. The compact-row motivation and the masked-read analysis here (the slot-0 poison tests in particular) are what prompted us to re-examine the SM120 sparse-MLA cache contract, and we've now opened flashinfer-ai/flashinfer#5075 as an alternative design for the same two goals. A quick map of how the two relate:

Same goals: 528-byte GLM NoPE rows, and masked sparse candidates no longer reading mutable cache slot 0.

Different mechanism:

- Layout: #5075 makes the 528B payload canonical (`bytes_per_token = 528`) and takes the gmem row advance as a runtime stride everywhere, instead of keeping two layouts selected by shape and advertised through capability fields. A legacy 656B vLLM pool and a packed 528B pool are then the same kernel with a different stride, so existing 3D/4D callers keep working unchanged; only flat 2D caches must be packed at 528B (a flat block carries no row stride to infer). GLM53_NOPE has no released downstream, so this is the cheapest moment to make the flip.
- Masked-read / bounds fixes: #5075 applies them to every model family rather than gating on GLM NoPE. The reason: the hazard is not glm53-specific. #5075's poison tests fail with NaN outputs on a main build for glm53_nope, dsv4 decode, dsv4 prefill, and dsv3_2 prefill alike, and its stale-padding cases crash a main build with an illegal memory access on both prefill routes. Your sglang companion hits the same fix on the glm53 path; we found it live everywhere.

Not included in #5075: the eight-head decoder from this PR — it is orthogonal to the layout and masked-read changes and can proceed independently, whichever way the layout question resolves.

If #5075 lands, it would supersede the compact-row and masked-read parts of this PR, and the sgl-project/sglang#38430 companion would no longer need capability negotiation for the row layout: a 656B pool keeps working, and moving to packed 528B rows is transparent to the kernels. We would very much appreciate your review on #5075, and we're happy to coordinate the sequencing however the maintainers prefer.

ormandj:
Thanks, Gabriel, for investigating the other model families and acknowledging the analysis and tests here. Making the payload size canonical while keeping row spacing independent, and extending the masking fixes across the affected families, looks like a useful direction.

I’m closing #5022 in favor of #5075 and have submitted the remaining work against your branch in https://github.com/lucifer1004/flashinfer/pull/3. It carries over the eight-head decode fix and additional regression coverage, including compact and padded layouts, fully masked rows, and CUDA graph replay with changed cache contents. It also tightens cache-view validation, invalidates stale calibration data, and checks additional padded-row and block-stride cases and all sixteen DeepSeek-V4.1 scale groups.

That should give maintainers one proposal to review and let us combine the strongest parts of both. I’m happy to help with the integration and validation.

For future work, I’m always open to other approaches, including substantial changes to the design, and happy to revise or split my PRs based on feedback. Please feel free to bring those ideas into the existing discussion. I’d welcome the chance to work through them together early, share the implementation and testing effort, and give maintainers one combined proposal to review.

Two details in #5075’s comparison could use an update:

- `glm53_nope_contract_version` was already removed in [a47c44ec](https://github.com/flashinfer-ai/flashinfer/commit/a47c44ec32c6763e368311cfd1b7613ea29c875f) following your feedback, along with the corresponding SGLang change.
- #5022 already uses runtime row advances for GLM NoPE prefill and swapAB, with decode honoring `stride_kv_row`. The layout difference is making 528 bytes canonical versus retaining the 656-byte default and flat-2D interpretation while advertising compact support.

Please preserve my commits’ authorship when incorporating them, including co-author credit in the final commit if the PR is squash-merged.

Thanks again for helping move this forward.

ormandj:
@lucifer1004, the follow-up PR against your branch is ready: https://github.com/lucifer1004/flashinfer/pull/3. Thanks again, Gabriel.

PR #4845, opened 2026-08-31
feat(cake_kda): add fp32 state & long context kda prefill kernels

yyihuang:
## Summary

This PR exports and integrates the complete recurrent-KDA portfolio for Blackwell GPUs.

- 104 variants, 106 physical selectors, and 59 unique generated CUDA bodies.
- BF16 compact/indexed state, FP32 indexed state, sparse `int32` P=257 pools, checkpoint support, and contexts through 1,048,576 tokens.
- Final validation covers all 531 paired rows: B200 260, B300 268, and NVIDIA GB300 3.

## Validated revisions and evidence

| Item | Revision or digest |
|---|---|
| Final public commit | `9f1f3ea7807799a4b01face909b64ddd416ebe18` |
| Final public tree | `854ae41eb6f40e07a32a798fc085bdb750b51648` |
| Matrix manifest SHA-256 | `824255af5843c9121c9cf7bee4f4fad83c30dcfdb81e580c0b0b33aa1e3a117c` |
| Matrix row-content SHA-256 | `8ee84e64d4a7f609b8cd7a7e572ef57d854bb406a0d7b879c2bf3fc66af22af6` |
| Public variant metadata SHA-256 | `ac22b0581e65ec4d41ca9ad02a5fe11b6b1da011bacdee77ac4a2c9ee17c95cd` |
| Generation receipt input SHA-256 | `d03e7d7bba5eddc5be5035cc8a76f0b5c04e6df705ca699de9f32f86831d627c` |
| Source-closure table SHA-256 | `51e98b3467f79e6f6cd4568e15ec0176ba5b7536a549949d2837675e3f63e3ba` |
| Physical-selector index SHA-256 | `d13fe2c6ea524e8758472be368ebfb45bb6c044a6fb7e43be3d1436ee71b48af` |
| Aggregate bridge SHA-256 | `28b1b5ded4e28ee7ce540873925a73accec7822b78543b0060ad433d575cd019` |
| Execution-equivalence receipt SHA-256 | `491e4da0a429a6fd3b364ecfe1e6c2d88cc5bc539ee50aa0015b53d1d51ec3e8` |
| Derived external-speedup rows SHA-256 | `b43a4efd5865f93e2bdbb3ba133444c8860cf601597db27c6c39c682a6a9a7ad` |
| All-row export-fidelity manifest SHA-256 | `9f85077dc93283dfeb295ec6176fdcf69b5f4247f1123e7f270bf082d4ab1c1b` |
[...]

PR #5035, opened 2026-09-08
test(kda): restore the frozen module-ident consistency check

kahyunnam:
## 📌 Description

`_FLASH_KDA_MODULE_IDENTS` holds a hand-copied 10-hex digest of each KDA variant's sources, and `get_flash_kda_uri()` embeds it in the module name. On the prebuilt path that name is all that is checked — `JitSpecNvcc.try_load()` returns the AOT `.so` whenever one exists under that name, with no source comparison — so the ident is the only part of the cache key carrying source identity. If it stops describing the sources, a module compiled from different sources can be loaded against the current caller.

The check that enforced this for the 11 common-header prefill variants was `tests/jit/test_flash_kda_prefill_jit.py::test_common_header_prefill_variant_cache_key`. That file was deleted in #4845 and the check was not carried over, so the invariant is currently unguarded. This PR restores just that check, in its own file.

Two deliberate differences from the deleted version, which is why this one should not rot the same way. First, nothing is pinned: each ident is recomputed from the sources on disk and compared to the table, so legitimate kernel edits never require touching this test. The deleted file also pinned literal idents, literal per-file SHA-256s, and literal constants like `#define SMEM_TOTAL 112256` in its `@parametrize` data, and 8 of its cases fail against today's `main` purely because of that churn. Second, `test_every_frozen_ident_is_accounted_for` asserts the table's key set equals the covered variants plus an explicit exclusion set, so adding a variant forces a conscious choice about coverage instead of silently escaping it.

The failure message names both resolved paths — where the table was imported from and where the sources were read from — because the two realistic causes need different responses. Same tree means an ident needs regenerating. Different trees means the imported `flashinfer` and its `data/csrc` symlink disagree and the package needs reinstalling; `get_kda_csrc_dir()` prefers `<package>/data/csrc/kda`, which `build_backend._create_data_dir` writes as an absolute build-time symlink, so a stale editable install silently pairs one tree's table with another tree's sources.

Scope: this adds coverage only. No library code changes, and the underlying source-resolution behaviour is left alone.

## 🔍 Related Issues

Internal: IKL-404. No linked GitHub issue.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

```
pytest tests/jit/test_flash_kda_frozen_idents_jit.py -q   # 12 passed
pytest tests/jit/ -k kda -q                               # 149 passed
```

Both negative cases were checked, since a tripwire that cannot fire is worthless. Appending a comment to `csrc/kda/flashkda_bf16_fused_m64.cu` fails the `m64` case with `assert 'f7f6a7e1f4' == 'd3c6b5f6fb'` and prints both resolved trees. Adding an uncovered entry to the ident table fails `test_every_frozen_ident_is_accounted_for` and names the new key.

Run on a B200 (CC 10.0), but these cases only hash files and build `JitSpec` objects, so they compile nothing and are arch-independent.

## Reviewer Notes

The 6 `bt16_*` idents are excluded rather than covered. They seal a different source closure, and I could not reproduce their digests from the common-header formula or from any of ~30 plausible body/binding/header permutations I tried. Someone who knows how those idents are generated should extend `_COMMON_HEADER_VARIANT_BODIES` or add a second formula; the exclusion set makes the gap explicit in the meantime.

`_COMMON_HEADER_VARIANT_BODIES` is the one piece of structural data the test still carries, because the variant-to-body mapping is not derivable: bindings shared by several variants (`cake_flashkda_bf16_fused_m128_n16_binding.cu`) `#include` more than one body and select between them with a `-D` flag. It could reasonably live next to `_FLASH_KDA_BINDING_STEMS` in `flashinfer/jit/flash_kda.py` instead; I kept it in the test to avoid adding a library constant with no library consumer.

Worth noting for whoever owns the KDA cache: nothing in the tree regenerates these idents, so they are maintained by hand. A small `scripts/` regenerator that this test could then verify against would be a better end state than either.

## Summary by CodeRabbit

* **Tests**
  * Added coverage to verify FlashInfer’s `flash_kda` JIT module identifiers remain consistent with their associated CUDA sources.
  * Added checks ensuring all supported module variants are represented and correctly validated.

### DIFF
--- a/tests/jit/test_flash_kda_frozen_idents_jit.py
+++ b/tests/jit/test_flash_kda_frozen_idents_jit.py
@@ -1,0 +1,83 @@
+# Copyright (c) 2026 by FlashInfer team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+
+import hashlib
+
+import pytest
+
+from flashinfer.jit import flash_kda
+
+# Body compiled into each common-header variant. Bindings come from
+# _FLASH_KDA_BINDING_STEMS; the ones shared by several variants
+# (m128_n16 / m128_n16_short) pick their body with a -D flag.
+_COMMON_HEADER_VARIANT_BODIES = {
+    "m64": "flashkda_bf16_fused_m64.cu",
+    "m128": "flashkda_bf16_fused_m128.cu",
+    "m128_tensor_state_decay": "cake_flashkda_bf16_fused_m128_tensor_state_decay.cu",
+    "m128_h12_short": "cake_flashkda_bf16_fused_m128_h12_short.cu",
+    "m128_h12_long": "cake_flashkda_bf16_fused_m128_h12_long.cu",
+    "m128_n16": "cake_flashkda_bf16_fused_m128_n16.cu",
+    "m128_n16_checkpoint": "flashkda_bf16_fused_m128_n16_checkpoint.cu",
+    "m128_n16_short": "cake_flashkda_bf16_fused_m128_n16_short.cu",
+    "persistent_m128": "cake_flashkda_bf16_persistent_m128.cu",
+    "piece_persistent_m128": "cake_flashkda_bf16_piece_persistent_m128.cu",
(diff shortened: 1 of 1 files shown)


kahyunnam:
@flashinfer-bot run

jimmyzho (approved):

Issue #5030, opened 2026-09-08
[BUG] GQA decode on sm75 crash with fp8 kv cache


ir1ka:
### Before submitting

- [x] I have searched existing issues and did not find a solution or an existing report.

### Environment

```text
### FlashInfer environment report
<!-- generated by `python -m flashinfer.collect_env`; paste into your issue -->

==== FlashInfer ====
flashinfer                   : 0.6.18
flashinfer commit            : 24db19e3b45d071055583049d9ed30fa5d8d4eb8
flashinfer file              : /home/mm/builder/src/flashinfer/flashinfer/__init__.py
flashinfer-cubin             : not installed
flashinfer-jit-cache         : not installed
JIT cache                    : /home/mm/.cache/flashinfer (54 compiled .so)
Cubin store (local)          : /home/mm/.cache/flashinfer/cubins (0 cubins, 0 MiB)
Target CUDA archs (resolved) : {(7, '5')}

==== Python / Platform ====
Python            : 3.12.3 (main, Jun 19 2026, 12:46:00) [GCC 13.3.0]
Python executable : /home/mm/.venv/bin/python
Virtual env       : /home/mm/.venv
Platform          : Linux-6.6.87.2-microsoft-standard-WSL2-x86_64-with-glibc2.39
libc              : glibc 2.39
OS                : Ubuntu 24.04.4 LTS
Container         : no / not detected

==== GPU / Driver ====
Driver version       : 595.79
CUDA_VISIBLE_DEVICES : <unset>
GPU 0 (CUDA order)   : NVIDIA T600 | SM75 | 10 SMs | 4.0 GiB

==== CUDA Toolkit ====
nvcc on PATH                            : /usr/local/cuda/bin/nvcc
CUDA_HOME (env)                         : <unset>
CUDA toolkit resolved by flashinfer JIT : /usr/local/cuda (version 13.2)
nvcc version                            : release 13.2, V13.2.78

==== PyTorch ====
torch                          : 2.11.0+cu130
torch.version.cuda             : 13.0
torch CXX11 ABI                : True
torch compiled archs           : sm_75 sm_80 sm_86 sm_90 sm_100 sm_120
torch.backends.cudnn.version() : 91900
torch file                     : /home/mm/.venv/lib/python3.12/site-packages/torch/__init__.py

==== cuDNN Frontend ====
cudnn-frontend : <not importable: ModuleNotFoundError: No module named 'cudnn'>

[...]

PR #5038, opened 2026-09-08
fix(decode): reserve max(K+V, st.o) smem for FA2 FP8+GQA decode

ir1ka:
The FA2 decode kernels (single + batch) placed smem_md (and the batch kv_offset_smem) at a fixed offset assuming the K+V shared-memory buffer is 2*num_stages_smem*bdy*tile_size_per_bdx*bdz*head_dim*sizeof(DTypeKV) bytes. But sync_state (bdz > 1, i.e. GQA) reuses that buffer as float storage for st.o, which needs bdz*bdy*head_dim*sizeof(float) bytes. When FP8 (1-byte KV)
+ GQA (tile_size_per_bdx=1) + SM75 (num_stages_smem=1) all hold, the float st.o is twice as large as the K+V buffer, so st.o overwrites smem_md and runs past the allocation (illegal smem write -> crash).

Reserve max(K+V data, st.o float) for the K+V region in both the kernel smem layout and the launcher smem_size. For every non-crashing config the max() is a no-op (K+V >= st.o), so there is no behavior change.

Add tests/attention/test_decode_fp8_gqa.py covering FP8+GQA single/batch decode against an fp16 reference.

#5030

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

## 🔬 Experimental Track

```experimental-tests
#
#
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed potential shared-memory overflow during FP8 grouped-query attention decode operations.
  - Improved reliability for single and paged batch decoding with FP8 KV caches and grouped-query attention.

- **Tests**
  - Added regression coverage for FP8 grouped-query attention in single- and batch-decode workflows.
  - Added validation across supported FP8 formats and paged KV-cache configurations.


bkryu (commented):
Thanks @ir1ka. The fix looks good to me and I can approve it.

Request though:
- Please drop the newly added `tests/attention/test_decode_fp8_gqa.py`. `test_decode_fp8_calibration_scale.py` already exercises FP8+GQA decode through the same API (its batch test is a superset of yours: both fp8 dtypes, head_dim 128/256, GQA 32/4). The kernel comments you added already document the root cause, and since the crash only fires on SM75 (our CI does not cover it), the new file can't act as a crash regression test anyway.

ir1ka:
The new test is already removed.
@bkryu Thanks for your comment.

bkryu:
@flashinfer-bot run

bkryu (approved):

PR #4965, opened 2026-09-04
refactor(kda): remove the unreleased recurrent KDA training/backward family

YangXu1990uiuc:
## Summary

KDA training (forward + backward) ships from cuDNN Frontend, where `cudnn.linear_attention.ops.kimi_delta_attention` already provides KDA fwd+bwd with autograd and where the CAKE training kernels are being added as an engine. FlashInfer keeps the inference surface. This removes the paired training API and the legacy backward that shares its JIT module, before either reaches a release.

Context: https://github.com/flashinfer-ai/flashinfer/issues/4936#issuecomment-5538954882 (step 3b of #4936 becomes removal instead of relocation to `flashinfer/experimental/`).

### Removed

- Public symbols: `recurrent_kda_training_forward`, `recurrent_kda_training_backward`, `RecurrentKDATrainingContext`, `recurrent_kda_backward`, `RecurrentKDABackwardWorkspace` (re-exports in `__init__.py` and `kda.py`).
- Modules: `kda_training.py`, `_kda_training_impl.py`, `_kda_training_dispatch.py`, `kda_backward.py`, `jit/flash_kda_training.py`, `jit/flash_kda_backward.py`; the `flash_kda_backward_sm10{0,3}a` capability flags and specs in `aot.py`.
- Frozen CUDA sources: `csrc/kda/flashkda_training_*.cu`, `flashkda_backward*.cu`, `training_fallback_pointer_sm_10{0,3}a.cu`, `training_grouped_row_wg8_pointer_sm_10{0,3}a.cu`, `cake_aligned_training_export/` (and their `.pre-commit-config.yaml` exclude).
- Docs: `docs/api/kda_training.rst`, `docs/api/kda_backward.rst`, two toctree lines.
- Tests / benchmarks: `tests/kda/test_recurrent_kda_{training,backward}.py`, `tests/jit/test_flash_kda_{training,backward}_jit.py`, `benchmarks/bench_recurrent_kda_{training,backward}.py`.

### Kept (released inference surface)

- The `recurrent_kda` prefill path including the `m128_n16_checkpoint` variant and the `state_checkpoints` / `checkpoint_cu_starts` / `checkpoint_every_n_tokens` kwargs (SGLang prefix caching).
- The `flashkda_binding_common.cuh` changes from #4636 (SM103a target, checkpoint granularity 16).

### Compatibility

[...]

Issue #4936, opened 2026-09-03
[KDA] Unify the public `recurrent_kda` API, and settle the unreleased KDA surface before FI v0.7


kahyunnam:
**TL;DR.** FlashInfer exports **two different public functions named `recurrent_kda`** — `flashinfer.recurrent_kda` and `flashinfer.kda_decode.recurrent_kda` — which register the same trace label, default to different backends, and disagree about whether to classify a call's phase, so **the same call can succeed through one import path and raise through the other, in both directions**. This issue proposes consolidating them onto one canonical entry point behind a shared decode dispatcher, keeping the older facade as a deprecated shim.

- Writing or reviewing KDA kernel or backend code right now? → [Rule for new kernels (immediate guidance)](#rule-for-new-kernels-immediate-guidance)
- Want to pick up part of this? → [Appendix: implementation sequence](#appendix-implementation-sequence)

## Context

FlashInfer grew its recurrent KDA surface in two places at once. `flashinfer.kda.recurrent_kda` arrived as a phase-neutral facade that classifies prefill from decode by shape; `flashinfer.kda_decode.recurrent_kda`, the older decode-only entry point, kept its own signature and its own default. Both are public, both are decorated, and both register the same trace label. A `backend=` parameter has since landed on the top-level facade ([#4605](https://github.com/flashinfer-ai/flashinfer/pull/4605)) without landing identically on the other, and open PRs are actively widening one enum and not the other (problem 6), so the gap between the two is still growing.

The problem is that **the two entry points disagree about what a defaulted call does.** They differ in their default backend — `"auto"` against `"cute-dsl"` — and in whether they classify a call's phase at all. The import path a caller happens to use therefore changes both which kernel runs and which inputs are accepted, in both directions. Problem 3 is the core of it; problems 1 through 8 enumerate the rest.

[...]

PR #5040, opened 2026-09-08
refactor(kda): stop 0.7 advertising unreleased KDA surface (mark prefill wrapper experimental, trim kda_kernels.__all__)

kahyunnam:
## 📌 Description

Stops `0.7` from advertising KDA surface that has never shipped and that the unification plan expects to change. No kernel logic, no behaviour change on any released path.

- `RecurrentKDAPrefillWrapper.plan` / `.run` get `@flashinfer_experimental_api`, and the planning implementation moves to `flashinfer/experimental/kda_prefill_wrapper/`. The class keeps its signatures and docstrings in `flashinfer/kda.py` as the thin entry point, reaching the planner through a function-local import, so core never imports `flashinfer.experimental` at module level. Removal is a one-directory deletion plus the thin class. Same shape as `sm110_gqa_decode`.
- The three `*_kda_prefill_sm120` names leave `kda_kernels.__all__`, which is now byte-identical to `v0.6.18`'s. They are dispatch plumbing for `flashinfer.kda_prefill`, and they are bound to `None` on any install where the SM120a module fails its optional import, so they were never fit to be public exports. Removing a name from `__all__` leaves the attribute in place and every consumer already reaches them by attribute or through `kda_kernels.sm120_prefill`, so no call site changes.
- The wrapper's tests move to `tests/experimental/`, including its CUDA-graph coverage. The stable lane no longer constructs the wrapper, so deleting the experimental directory leaves it green.

None of this is in `v0.6.18` — that tag has no `kda_kernels/sm120_prefill/` at all. Both changes are free today and cost a deprecation cycle once `0.7` cuts.

## 🔍 Related Issues

- #5069 — graduation plan for the wrapper (owner, criteria, target).
- #4936 — KDA API unification audit; this is steps 3b-i and 3c.
- #4965 has landed, removing the training/backward family, which is why only `RecurrentKDAPrefillWrapper` is marked here. With it merged and this PR, no unreleased KDA symbol is advertised at the top level for 0.7.

## 🧪 Tests

All on a B200 (CC 10.0).

- `tests/experimental/test_kda_prefill_wrapper.py` — 10 passed. These pin the experimental
  marking itself, the top-level export, the plan metadata and buffer handoff, the planner's
  validation, the planned path against the eager packed path it wraps (plain and checkpointed,
  bit-exact), and CUDA-graph capture/replay.
- `tests/kda/test_recurrent_kda_prefill.py` — 262 passed, 1 skipped, run with
  `flashinfer/experimental/kda_prefill_wrapper/` deleted. The stable lane has no remaining
  reference to the wrapper, so removal touches no stable test.
- `tests/kda/test_recurrent_kda_prefill_sm120.py` — 50 passed, 88 skipped (the skips are CC 12.0 cells).
- `examples/experimental/kda_prefill_wrapper.py` runs.
- Verified `flashinfer.experimental` is not in `sys.modules` after `import flashinfer`, that
  `pytest tests/` does not collect the relocated file, and that all three SM120a names remain
  reachable as attributes.

Three mutations were used to check the tests actually hold the change down, each of which used to
pass unnoticed: deleting both `@flashinfer_experimental_api` decorators, and dropping `scale` or
`use_qk_l2norm_in_kernel` from the handoff. All three now fail. The handoff assertion is derived
from `run`'s signature, so it covers parameters that do not change numerics.

The two cake-vs-CuTe comparisons that used to drive the wrapper now call the eager packed path.
That is a different dispatch configuration inside the kernel wrapper, not the same call, so the
reference test covers the planned path at both the plain and the checkpointed/indexed
configurations those tests exercise. Shared input builders and reference implementations moved to
`tests/test_helpers/` so both lanes can use them; that accounts for most of the diff in the stable
file.

## 📝 Notes for reviewers

- The class is deliberately **not** listed in `docs/api/kda_prefill.rst`, per
  `flashinfer/experimental/README.md` ("Experimental APIs are not listed in `docs/api/*.rst`
  before graduation"). The prose note still states the experimental status, the opt-in and the
  limitations, and the package README carries the detail.
- `RecurrentKDAPrefillWrapper.__init__` is not decorated, so construction imports the experimental
  package and allocates a workspace without warning. `plan` and `run` both warn, once each per
  process, so any actual use is warned; construction alone is not.
- `_planned`, `_num_sequences`, `_total_tokens` and `_lock` moved to the planner and are no longer
  attributes of the wrapper. `device`, `plan` and `run` are the whole public surface.

## 📋 Not included (deliberately)

Step 3c in #4936 also proposes giving SM120a an explicit `backend=` literal and gating its automatic branch. Both change `"auto"` prefill routing on CC 12.0, mostly by converting working calls into raises, so they belong in their own PR validated on a CC 12.0 device.

## 🚀 Pull Request Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/flashinfer-ai/flashinfer/blob/main/CONTRIBUTING.md).
- [x] I have installed and run `pre-commit`.

## 🔬 Experimental Track

- [x] This PR is **experimental**: it adds or changes code under `flashinfer/experimental/` and/or an `@flashinfer_experimental_api`. Tracking issue: #5069
  - [x] The tracking issue names an owner, the reason for the experimental path, and a graduation plan with a target release.
  - [x] Core changes are limited to a thin entry point (signature, shared validation, feature-gate check, backend selection, handoff).
  - [x] Tests live in `tests/experimental/` and were validated on the intended hardware, including a correctness test against a reference. `examples/experimental/kda_prefill_wrapper.py` is runnable, and `docs/api/kda_prefill.rst` states the experimental status, the opt-in and the limitations.
  - [x] Nothing is registered in `flashinfer/aot.py`, and no experimental backend is reachable from `backend="auto"` without `FLASHINFER_ALLOW_EXPERIMENTAL_AUTO_BACKENDS=1`. (Calling an `@flashinfer_experimental_api` or naming a backend explicitly is itself the opt-in and needs no environment variable.) The stable kernels the wrapper hands off to stay AOT-registered, since `run` calls the stable `recurrent_kda` facade.
  - [x] **Test scope declared below.** The experimental CI lane runs exactly these targets, so keep them as narrow as the change allows.

```experimental-tests
tests/experimental/test_kda_prefill_wrapper.py
```

## Summary by CodeRabbit

- **New Features**
  - Packed prefill planning now generates ordering and chunk metadata on the device, reducing host-side preparation.
  - CUDA Graph replays can use shape-based capacity while adapting to the actual chunk count.

- **Documentation**
  - Documented `RecurrentKDAPrefillWrapper` as experimental, including opt-in behavior, supported compute capabilities, backend limitations, and warmup requirements.
  - Updated guidance for accessing SM120 prefill functionality.

- **API**
  - Removed SM120 prefill utilities from package-level exports.

## Summary by CodeRabbit

- **New Features**
  - Added an experimental recurrent KDA packed prefill planner for supported CUDA architectures using the CuTe DSL backend.
  - Added a runnable example with configurable sequence lengths and attention heads.

- **Documentation**
  - Clarified opt-in behavior, supported hardware, warmup constraints, kernel dispatch, and experimental status.

- **Breaking Changes**
  - Removed SM120 KDA prefill functions from package-level exports; use the dedicated prefill modules instead.


kahyunnam:
@flashinfer-bot run

kahyunnam:
Thanks @bkryu — both addressed in 5a5c667.

**Containment.** `flashinfer/experimental/kda_prefill_wrapper/` now holds `RecurrentKDAPrefillPlanner`: the fixed-address buffers, the `plan` validation, the graph-replay invariants and the `backend="cute-dsl"` routing. `RecurrentKDAPrefillWrapper` stays in `flashinfer/kda.py` as the thin entry point — signature, docstring, gate, handoff — reaching the planner through a function-local import, so core never imports `flashinfer.experimental` at module level. That follows the `sm110_gqa_decode` shape and what `README.md` and `CLAUDE.md` ask for: the public experimental API lives in core, its implementation does not. Removal is now a one-directory deletion plus the thin class.

I had argued in the original description that the wrapper was exempt because it is host-side planning with no backend to contain. That was wrong — the README is explicit that when the *API* is experimental its implementation moves regardless of the implementation's maturity, and the point is the deletion property rather than the maturity of what is being contained.

**Graduation issue.** Filed as #5069, modelled on #5051: owner, use case, why experimental, graduation plan. It is deliberately short and wrapper-scoped — #4936 stays as the wider unification audit and is now linked as context rather than as the tracking issue. The one deviation from the four-week default is that graduation is conditional on the #4936 steps that can still invalidate the plan-and-run shape rather than on a timer, with the first release after `v0.7` as the target; happy to make it a hard date if you would rather.

The class docstring and `docs/api/kda_prefill.rst` now point at #5069 and state the opt-in and the limitations. Both lanes still pass on B200 — `tests/kda/test_recurrent_kda_prefill.py` 265 passed / 1 skipped, `tests/experimental/test_kda_prefill_wrapper.py` 2 passed — and the four buffer attributes the tests assert on are forwarded as read-only properties, so no test changed.

kahyunnam:
@flashinfer-bot run tests/experimental/test_kda_prefill_wrapper.py

kahyunnam:
@flashinfer-bot run

kahyunnam:
@flashinfer-bot run

kahyunnam:
@flashinfer-bot run tests/experimental/test_kda_prefill_wrapper.py

bkryu (approved):

PR #4697, opened 2026-08-24
refactor(mla): isolate planned FA2, FA3, and CUTLASS backends

saltyminty:
## 📌 Description

This is the first standalone section extracted from [the original Batch MLA redesign PR](https://github.com/flashinfer-ai/flashinfer/pull/4031). That PR covered several independent backend and API changes; this series breaks the work into smaller PRs that can be reviewed, validated, and landed individually.

### Current scope

This PR establishes the planned Batch MLA foundation for the existing FA2, FA3, and CUTLASS backends:

- moves concrete planning and execution state into backend-owned implementations;
- introduces canonical plan metadata and explicit packed/split query and KV-cache layout contracts;
- keeps the planned `run()` path backend-native, with no runtime backend discovery or replanning, and reuses plan-owned workspaces and empty-LSE storage;
- makes CUDA Graph replanning transactional while preserving the wrapper attributes used by SGLang's fast replay path;
- preserves legacy public imports, positional/flat-CSR planning, split tensor calls, dynamic LSE behavior, and caller-owned output/LSE identity;
- retains deprecated planless CUTLASS execution as a validated per-call path without publishing or mutating planned state;
- preserves historical trace identities while allowing trace replay to recognize the reorganized implementation; and
- documents the backend ownership and lifecycle conventions introduced by this slice.

The structural input resolver also preserves the existing zero-width PE compatibility case: a packed plan can still accept `(left, empty_right)` without requiring a copy.

### Incoming follow-up sections

Subsequent PRs are expected to cover, independently:

1. a public sync-free CUDA Graph plan-update API and removal of the temporary private SGLang bridge;
2. TRTLLM-GEN, XQA, and CuTe DSL backend verticals;
3. deterministic automatic selection with typed unsupported-backend fallback;
4. the unified tensor-first functional API; and
5. optional autotuning and benchmark follow-ups.

## 🔍 Related Issues

- Original umbrella PR: https://github.com/flashinfer-ai/flashinfer/pull/4031
- Related issue: https://github.com/flashinfer-ai/flashinfer/issues/4037

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [ ] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

[...]

PR #5041, opened 2026-09-08
feat(mla): add CUDA graph plan update API

saltyminty:
(DON'T MERGE YET)

## 📌 Description

Add a public, opt-in CUDA-graph plan-update lifecycle for the generated FA2 and
FA3 MLA backends:

```python
wrapper = BatchMLAPagedAttentionWrapper(
    workspace,
    use_cuda_graph=True,
    enable_cuda_graph_plan_update=True,
    qo_indptr=qo_indptr_buf,
    kv_indptr=kv_indptr_buf,
    kv_indices=kv_indices_buf,
    kv_len_arr=kv_len_arr_buf,
)
wrapper.plan(metadata=initial_metadata, ...)
# capture wrapper.run(...)
wrapper.update_cuda_graph_plan(metadata=next_metadata)
# replay on the same stream
```

The initial plan freezes the captured backend and buffer contract. Each update
uses bounded preallocated host staging, prepares a candidate schedule without
modifying live graph buffers, and publishes the schedule plus CSR metadata with
one native CUDA launch. The steady-state path performs no device/pinned-host
allocation, device-to-host copy, or blocking host synchronization.

The update state is allocated only when explicitly enabled. It reuses unused
regions of the existing device and pinned planner workspaces when capacity
allows, retains no committed schedule or page-index shadow, and preserves the
unused tail of the caller's reserved index buffer.

This keeps the private wrapper mirrors and existing planner entry points
behavior-compatible for older SGLang releases. Their deprecation is
documentation-only here because some existing callers promote deprecation
warnings to errors.

## 🔍 Related Issues

Follow-up to #4697. This enables the public-API migration demonstrated in
sgl-project/sglang#35555.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

Final validation after rebasing onto `f32f740a`:

- H100 (SM90), CUDA 13.0: `tests/attention/test_mla_cuda_graph_plan_update.py`
  and `tests/attention/test_mla_wrapper.py` — 121 passed.
- B200 (SM100), CUDA 13.0, downstream SGLang public-API integration:
  MLA CUDA-graph suite — 6 passed, 1 pre-registered skip, 15 subtests passed;
  dense block-table suite — 15 passed; repeated public update probe — 20/20.
- `pre-commit run --all-files` — all applicable hooks passed.

Paired H100 update-latency probe against the previous private direct-replan
path (100 warmups, 30 repetitions, 200 samples per arm):

| Backend | Workload | Median CPU delta | One-sided 95% regression UCB |
| --- | --- | ---: | ---: |
| FA2 | decode | -0.73 us | -0.39% |
| FA2 | target verify | +0.49 us | +0.35% |
| FA2 | speculative | +0.27 us | +0.24% |
| FA3 | decode | +0.07 us | +0.67% |
| FA3 | target verify | +0.56 us | +0.43% |
| FA3 | speculative | +0.77 us | +0.59% |

An Nsight Systems trace over 48 warmed public updates observed zero device or
pinned-host allocation/free calls, zero blocking CUDA synchronization calls,
and zero device-to-host copies. The private comparator issued 96
`cudaStreamSynchronize` calls.

## 🔬 Experimental Track

```experimental-tests
#
#
```

## Reviewer Notes

- The first eligible update binds its CUDA stream. The corresponding graph
  replay must use that same stream; cross-stream replay and concurrent updates
  on one wrapper are unsupported.
- `kv_indices` is a non-overlapping device source that must remain alive until
  publication completes. Host controls must be contiguous CPU `int32` tensors.
- Validation, planning, and staging failures preserve the prior runnable plan.
  Failures discovered after the native publication launch has been submitted
  are intentionally outside the synchronization-free rollback guarantee.
- Only FA2 and FA3 opt in through backend capabilities. CUTLASS, cuTile, and
  future backends default to unsupported.

## Summary by CodeRabbit

- **New Features**
  - Added optional CUDA graph plan updates for MLA workloads using FA2 and FA3 backends.
  - Added support for refreshing scheduling metadata while preserving captured buffer locations.
  - Added preallocated staging to reduce update-time allocations.

- **Bug Fixes**
  - Added validation for metadata, tensors, capacities, devices, streams, and update lifecycle conditions.
  - Failed replans now preserve the active plan and resources with correct cross-stream rollback ordering.

- **Documentation**
  - Documented configuration, lifecycle rules, supported backends, limitations, and failure behavior.

- **Tests**
  - Added coverage for validation, staging, stream handling, capture/replay, and transactional failures.


qsang-nv:
Four points on the transactional contract this API is built around.

**1. Update state is initialized after the live metadata buffers have already been overwritten.**

`plan_from_wrapper` builds the update state *after* `backend.plan()`, by which point `_stage_metadata` has copied the new CSR in place into the capture-reserved `qo_indptr` / `kv_indptr` / `kv_indices` / `kv_len_arr`. Its rollback covers only its own copy loop, and the pre-existing outer guard in `_wrapper.plan()` snapshots and restores only `graph_plan_int_workspace_buffer`.

So a failure in between — control allocations, event setup, or the `.to("cpu")` conversion when `qo_indptr` is on device — leaves the integer workspace holding the **old** schedule and the four metadata buffers the **new** CSR. With a shorter replacement CSR the kernel can follow old `kv_start` / `kv_end` offsets into `kv_indices` past the new logical prefix at `kv_indptr[-1]`, reading stale or never-initialized entries. A wrong result also requires replaying the captured graph after `plan()` raised.

**2. The two-slot state machine lacks critical automated coverage.**

`test_real_generated_capture_update_replay_uses_lean_state` is the only test running the generated kernels, and it does a single update. Missing:

- slot 0 → slot 1 rotation and busy rejection on a third update (both should be exercised with real CUDA events);
- event-record failure and both-slots-poisoned recovery;
- the `plan_info` / `staged_int_workspace_bytes` mismatch rejections;
- initialization failure after `backend.plan()` succeeded, which also covers point 1.

The H100/B200 results are manual runs and the GPU matrix jobs here are skipped.

**3. `_cuda_graph_plan_update_stream` is never cleared by a successful `plan()`.**

CodeRabbit flagged this; I verified it. The docs name `plan()` as the recovery path for poisoned slots, so re-planning and re-capturing on a different stream leaves every update rejected with `must use the initially bound CUDA stream`.

**4. May the public update and the private fast-plan bridge share one wrapper?**

Slot 0 is backed by `[0:staged]` of `_pin_memory_int_workspace_buffer`, which the legacy raw planner rewrites synchronously from the CPU — not ordered against an in-flight H2D still sourcing from it. The standing compatibility promise is not broken on its own, since `enable_cuda_graph_plan_update` defaults to False. But the docs do not say the two lifecycles cannot be mixed: if unsupported, stating that next to the compatibility promise would settle it; if supported, the legacy planner must not rewrite the prefix until slot 0's H2D has completed.

flashinfer/mla/_batch_mla/_wrapper.py
```diff
@@ -648,13 +669,29 @@ def plan(
...
+                graph_metadata_snapshots = tuple(
+                    (tensor, tensor.clone())
+                    for tensor in (
+                        self._qo_indptr_buf,
+                        self._kv_indptr_buf,
+                        self._kv_indices_buf,
+                        self._kv_len_arr_buf,
+                    )
+                    if tensor is not None
+                )
```
bkryu, line 711: The transactional snapshot/rollback here handles replan failure well (and the five-failure-mode test is convincing), but replanning  never queries or drains the pending update slots first — safety rests on same-stream ordering plus the pinned-workspace layout  argument. A replan issued on a different stream while an old slot's H2D is still in flight is currently a documented-only caller  obligation. A `slot.completion_event.query()` loop here that warns (or debug-asserts) when slots are still pending would convert that  silent hazard into a visible one at near-zero cost — no wait, just a query.

Related cost note worth a docs sentence: this snapshot clones the entire kv_indices reservation on every replan; fine off steady state, but SGLang-scale reservations make replans measurably heavier than before.

flashinfer/mla/_batch_mla/_wrapper.py
```diff
@@ -686,6 +723,122 @@ def plan(
...
+            if not hasattr(planned_backend, update_hook_name):
+                raise RuntimeError(
+                    "the planned backend does not implement CUDA graph plan updates."
+                )
+            getattr(planned_backend, update_hook_name)(metadata=metadata)
+        finally:
+            self._cuda_graph_plan_update_in_progress = False
+
+    @flashinfer_api
+    def update_cuda_graph_plan(self, *, metadata: MLAPlanMetadata) -> None:
```
bkryu, line 792: Small robustness items, none blocking: (a) initialize `_cuda_graph_plan_update_stream` in `__init__` rather than relying on `getattr(...,None)`; (b) hold the `torch.cuda.Stream` object instead of the raw `cuda_stream` pointer — a destroyed-then-recreated stream at the same address currently passes the binding check falsely; (c) an opted-in wrapper whose backend was planned on a non-CUDA device skips  state creation and a later update dies with `AttributeError` instead of a clean error; (d) the `_cuda_graph_plan_update_in_progress` check-then-set is not thread-safe — the docstring's "rejected when detectable" wording covers it, but say that explicitly here.

qsang-nv:
Thanks — the rollback, the stream-lifecycle reset, and the real-event slot tests (the `cudaLaunchHostFunc` gate, the `torch.Tensor.clone` patch proving the cross-stream rejection short-circuits before snapshotting) address the earlier points well.

Point 1's fix does not cover the cross-stream case.

The replan guard queries the bound stream once at entry. After that, `plan()` stages the new CSR and the `except` block issues its rollback `copy_()` calls, both on the *current* (new) stream and both async with respect to the host. A failed plan keeps the old binding — `test_real_generated_capture_update_replay_uses_lean_state` asserts exactly that.

The entry-time query only proves the old stream was idle before replanning began; it creates no ordering edge for the staging and rollback that follow, and the failure path adds none. So after a cross-stream replan fails past staging, a caller that catches the exception and replays the captured graph on the old stream — which both the preservation contract and the retained binding invite — can race with those writes until they complete, observing the newly staged metadata, a partially restored state, or the restored state.

Coverage does not reach this: the late-initialization-failure injection captures, fails, and replays on one stream, so same-stream ordering hides it; the cross-stream tests cover only a successful replan and the busy rejection, which short-circuits before staging.

Same transactional gap as point 1: the last round fixed *what* state a failure restores, but not *when* that restoration is ordered before reuse on the bound stream. It is confined to the failure path of a cross-stream replan.

qsang-nv (approved):

saltyminty:
/bot run tests/attention/test_mla_cuda_graph_plan_update.py tests/attention/test_mla_wrapper.py

bkryu (approved):

PR #5043, opened 2026-09-09
perf(attention): use Tensor.max() for the longest lengths in plan()

gf239:
## 📌 Description

`plan()` finds the longest query and KV length with Python's built-in `max()`.
That is a Python loop over the batch.
At large batches it is most of the host time of `plan()`.

This PR uses `Tensor.max()` instead. Same result, one call.
The ragged prefill wrapper already does this.

Five lines change: two in paged prefill, three in paged decode.

`seq_lens` is documented as uint32. Torch cannot compare or reduce uint32 on CPU, so `plan()` already failed on uint32 input with two or more requests.
This PR converts `seq_lens` to int32 once on entry. For int32 input this is a no-op.

Host time of one `plan()` call, RTX 4090:

| batch size | prefill, before | prefill, after | decode, before | decode, after |
| --- | --- | --- | --- | --- |
| 64 | 281 µs | 90 µs | 173 µs | 76 µs |
| 256 | 850 µs | 99 µs | 461 µs | 87 µs |
| 1,024 | 3,118 µs | 130 µs | 1,605 µs | 112 µs |

In vLLM with MTP at 64 to 256 concurrent requests: ITL is 1.2% to 2.4% lower, output is 1.3% to 2.5% higher.

<details>
<summary>Benchmark details</summary>

`plan()` timing: Ryzen 9 5950X, torch 2.14. KV length 512 to 4,096, page size 16, 8 heads, head dim 128. Median of 6 runs. Decode uses tensor cores.

| batch size | prefill, before | prefill, after | decode, before | decode, after |
| --- | --- | --- | --- | --- |
| 1 | 89 µs | 87 µs | 75 µs | 75 µs |
| 8 | 115 µs | 87 µs | 90 µs | 75 µs |

End to end: vLLM 0.29.0, Qwen3.5-0.8B, MTP with 3 draft tokens, FlashInfer backend, RTX 4090. Random prompts, 6 rounds per arm, arms alternated. Only the five lines differ. 95% confidence intervals:

| prompt / output tokens | concurrency | median ITL | output tokens/s |
| --- | --- | --- | --- |
| 128 / 128 | 64 | -2.2% [-2.6, -1.8] | +2.4% [+1.6, +3.1] |
| 128 / 128 | 256 | -2.4% [-3.3, -1.4] | +2.5% [+1.3, +3.6] |
| 1,024 / 128 | 64 | -1.9% [-2.6, -1.1] | +1.5% [-0.1, +3.1] |
| 1,024 / 128 | 256 | -1.2% [-1.9, -0.5] | +1.3% [+0.6, +1.9] |

Why it shows: without TRTLLM attention kernels, vLLM sends speculative verify rows through the prefill wrapper. `plan()` then runs every step, one row per request.

No measurable change: concurrency 8 to 16, 8,192-token prompts, or no speculative decoding.

Script for the `plan()` timing:

```python
"""Host time of BatchPrefillWithPagedKVCacheWrapper.plan() and
BatchDecodeWithPagedKVCacheWrapper(use_tensor_cores=True).plan() when the caller
does not pass the longest lengths. Prints one CSV row per configuration:
wrapper,batch_size,median_us,p10_us,p90_us"""

import statistics
import sys
import time

import torch

import flashinfer

PAGE, HEADS, HEAD_DIM, CALLS, WARMUP = 16, 8, 128, 2000, 200


def inputs(bs):
    g = torch.Generator().manual_seed(bs)
    kv = torch.randint(512, 4097, (bs,), dtype=torch.int32, generator=g)
    q = torch.randint(1, 5, (bs,), dtype=torch.int32, generator=g)
    pages = (kv + PAGE - 1) // PAGE
    kv_indptr = torch.nn.functional.pad(torch.cumsum(pages, 0, dtype=torch.int32), (1, 0))
    qo_indptr = torch.nn.functional.pad(torch.cumsum(q, 0, dtype=torch.int32), (1, 0))
    last = kv - (pages - 1) * PAGE
    indices = torch.arange(int(kv_indptr[-1]), dtype=torch.int32, device="cuda")
    return qo_indptr, kv_indptr, indices, last


def timeit(fn):
    for _ in range(WARMUP):
        fn()
    torch.cuda.synchronize()
    ts = []
    for _ in range(CALLS):
        t0 = time.perf_counter()
        fn()
        ts.append((time.perf_counter() - t0) * 1e6)
        torch.cuda.synchronize()
    ts.sort()
    return statistics.median(ts), ts[len(ts) // 10], ts[len(ts) * 9 // 10]


def main():
    ws = torch.empty(256 * 1024 * 1024, dtype=torch.uint8, device="cuda")
    for bs in (1, 8, 64, 256, 1024):
        qo, kvp, idx, last = inputs(bs)
        pre = flashinfer.BatchPrefillWithPagedKVCacheWrapper(ws, "NHD")
        dec = flashinfer.BatchDecodeWithPagedKVCacheWrapper(ws, "NHD", use_tensor_cores=True)
        # indptr and last-page lengths on the CPU, as serving engines pass them.
        p = lambda: pre.plan(qo, kvp, idx, last, HEADS, HEADS, HEAD_DIM, PAGE, causal=True, q_data_type=torch.float16)
        d = lambda: dec.plan(kvp, idx, last, HEADS, HEADS, HEAD_DIM, PAGE, q_data_type=torch.float16)
        for name, fn in (("prefill", p), ("decode_tensor_cores", d)):
            med, p10, p90 = timeit(fn)
            print(f"{name},{bs},{med:.1f},{p10:.1f},{p90:.1f}", flush=True)


if __name__ == "__main__":
    print(f"# flashinfer={flashinfer.__file__} torch={torch.__version__} gpu={torch.cuda.get_device_name(0)}", file=sys.stderr)
    main()
```

</details>

## 🔍 Related Issues

None.

### ✅ Pre-commit Checks

## 🧪 Tests

New tests pin the lengths `plan()` computes:

- `tests/attention/test_batch_prefill_kernels.py::test_batch_prefill_plan_max_lens`
- `tests/attention/test_tensor_cores_decode.py::test_batch_decode_tensor_cores_plan_max_kv_len`

They run with `seq_lens` omitted, int32 and uint32. The uint32 cases fail without the int32 conversion.

Existing prefill and decode tests: same results before and after, 0 failures (RTX 4090).

<details>
<summary>Test counts</summary>

| file | passed | skipped | xfailed |
| --- | --- | --- | --- |
| `test_tensor_cores_decode.py` | 2,448 | 0 | 0 |
| `test_batch_decode_kernels.py` | 3,180 | 82 | 0 |
| `test_batch_prefill_kernels.py` | 8,791 | 1,329 | 5,760 |

</details>

## 🔬 Experimental Track

```experimental-tests
#
#
```

## Reviewer Notes

- The cute-dsl and trtllm-gen decode lines were not run here. They need SM100 or CuTe DSL hardware.
- In vLLM only the prefill lines are on the per-step path.

Generated with Claude Code and Human In The Loop 🙈


gf239:
@saltyminty could you take a look? It replaces the built-in `max()` in `plan()` with `Tensor.max()` in all five places. Timings and test results are in the description. Could you also run `@flashinfer-bot run`?

saltyminty:
There is _technically_ a potential regression where a singleton uint32 tensor will fail now fail during the `max` call whereas it would have passed in the past. This one edge case that likely won't actually happen, so can be treated as acceptable risks/non-blocking.

saltyminty (approved):

saltyminty:
/bot run tests/attention/test_batch_prefill_kernels.py tests/attention/test_tensor_cores_decode.py tests/attention/test_cute_dsl_decode.py tests/attention/test_trtllm_gen_attention_decode.py tests/attention/test_trtllm_gen_attention_prefill.py

gf239:
@saltyminty thanks for the review. I checked the uint32 case.

`seq_lens` is documented as uint32, and torch has no CPU comparison or max for uint32, so on main the built-in `max()` already fails for two or more requests.
Commit fca030b converts `seq_lens` to int32 once on entry (a no-op for int32).
The plan-length tests now run with int32 and uint32; the uint32 cases fail without that commit and pass with it.

Could you please re-run CI?

saltyminty:
/bot run tests/attention/test_batch_prefill_kernels.py tests/attention/test_tensor_cores_decode.py tests/attention/test_cute_dsl_decode.py tests/attention/test_trtllm_gen_attention_decode.py tests/attention/test_trtllm_gen_attention_prefill.py

gf239:
@saltyminty the pipeline is red on one job only: `tests.attention.test_trtllm_gen_attention_prefill` on B200 / CUDA 13.4, with `RuntimeError: Failed to load cubin for fmhaSm100aKernel_...`.

That looks like the runner rather than this change:
- the failure is a cubin load, before any kernel runs;
- the same B200 passes on CUDA 12.9 and 13.0, and every other GPU passes on 13.4;
- this PR touches `plan()` in `flashinfer/decode.py` and `flashinfer/prefill.py` plus two test files, and nothing in the trtllm-gen path;
- #5177 had the same kind of B200 load error and it was unrelated there too.

Could you rerun that job please?

saltyminty:
/bot run tests/attention/test_batch_prefill_kernels.py tests/attention/test_tensor_cores_decode.py tests/attention/test_trtllm_gen_attention_prefill.py tests/attention/test_trtllm_gen_attention_decode.py

Issue #4432, opened 2026-08-10
download_artifacts() silently skips every DSL_FMHA (cute-dsl) kernel — get_available_cubin_files matches only *.cubin, but those artifacts are *.so


RuiyangSi:
## Summary

`flashinfer.artifacts.download_artifacts()` intends to pre-fetch the cute-dsl FMHA
kernels — `get_subdir_file_list()` explicitly adds one directory per arch to
`cubin_dirs`:

```python
# DSL FMHA: per cpu-arch and sm-arch subdirectories
*(
    safe_urljoin(ArtifactPath.DSL_FMHA, f"{cpu_arch}/{arch}/")
    for arch in ArtifactPath.DSL_FMHA_ARCHS
),
```

but it downloads none of them, because the directory enumerator keeps only
`.cubin` names:

```python
# flashinfer/artifacts.py, get_available_cubin_files
hrefs = re.findall(r'\<a href=".*\.cubin">', response.text)
```

The cute-dsl artifacts are shared objects, not cubins
(`cute_dsl_fmha_bf16_h128_causal_nonpersistent_varlen_lse_pdl_tvmffi.so`), so
they never enter the download list.

Two things make this hard to notice:

1. Each directory's `checksums.txt` **is** downloaded (it is yielded separately,
   from `CheckSumHash.map_checksums`), so on disk the manifest looks complete
   while every kernel it lists is absent.
2. `download_artifacts()` finishes with a full progress bar and no error. Its
   post-download checksum loop only iterates the files it enumerated, so the
   files it never enumerated cannot fail it.

On our machine this left **960 files missing**, including **all 320 for
`sm_103a`, the arch we actually run on**, while the run reported success.

## Why it matters

Anything missing from the cache is fetched over HTTP by `get_cubin_callback` →
`get_artifact` → `download_file` on first use — which happens **inside the model
forward pass**. Under a serving workload that blocks the scheduler until the
network answers. On an 8-way tensor-parallel deployment the other ranks then
stall behind the collective, the HTTP front end keeps answering health checks in
~1 ms, and the framework's watchdog eventually aborts the process. Ours took
~2 h to reload afterwards.

The per-rank native stack at one such hang:

```
trtllm_ragged_attention_deepseek        (flashinfer/prefill.py:4419)
  __tvm_ffi_trtllm_ragged_attention     (fmha_gen.so)
    get_cubin_callback                  (flashinfer/jit/cubin_loader.py:316)
[...]

PR #5050, opened 2026-09-09
fix(cute-dsl): verify the DSL FMHA checksum manifest against its pin

Dev-next-gen:
## 📌 Description

`_get_checksums()` in `flashinfer/attention/cute_dsl/fmha.py` fetches `checksums.txt` with `download_file()` and parses whatever lands on disk. Nothing checks it, even though `CheckSumHash.DSL_FMHA_CHECKSUMS` in `flashinfer/artifacts.py` exists to pin exactly that manifest, per cpu-arch and sm-arch, and is commented as such:

```python
    # SHA256 of the checksums.txt manifest file per cpu-arch/sm-arch,
    # NOT hashes of individual kernel .so files.
    DSL_FMHA_CHECKSUMS: dict[str, dict[str, str]] = {...}
```

The hashes parsed out of that manifest are the ones `get_artifact()` then verifies the `.so` kernels against, so the kernel check was circular — whoever served the manifest also chose the hashes it was judged by. Every other manifest consumer already passes its pin (`jit/fused_moe.py`, `jit/moe_utils.py`, `jit/attention/modules.py`, and `artifacts.py` via `CheckSumHash.map_checksums`); the cute-dsl path is the one that doesn't.

Two smaller things fall out of the same line:

- `download_file()` returns `False` on failure rather than raising, so an unfetchable manifest surfaced as a bare `FileNotFoundError` on the local cache path. That is the same failure mode `get_checksums()` was fixed for in #4280, whose test docstring says it "reads like a corrupt cache rather than an unreachable pin". The cute-dsl copy never got that fix.
- The `if not local_path.exists()` guard means a cached manifest is never revalidated. Once a bad file is on disk it is trusted for the life of the cache. `get_artifact()` re-fetches when the local hash doesn't match.

The fix routes the fetch through `get_artifact()` with the pin, which handles all three. It removes code rather than adding a new path, and adds no API.

## 🔍 Related Issues

None open for this. Adjacent but different: #4432 / #4443 / #4548 are about *enumerating* cute-dsl `.so` artifacts for pre-fetch. Those touch `flashinfer/artifacts.py` and `tests/test_artifacts.py`; this touches `flashinfer/attention/cute_dsl/fmha.py` plus appended tests, so it should sit alongside them.

### ✅ Pre-commit Checks

To be exact: I ran `pre-commit run --files` on the two changed files rather than `--all-files`. `mypy`, `ruff check` and `ruff format` all pass; the rest of the hooks report no files to check or pass.

## 🧪 Tests

Four tests appended to `tests/test_artifacts.py`, next to the existing `get_checksums` ones. They stub `download_file` and redirect the cubin dir through both `jit.env` and `jit.cubin_loader`, so they never touch a real `~/.cache/flashinfer` and never hit the network.

Against `main`, three of the four fail:

```
test_dsl_fmha_manifest_matching_its_pin_is_parsed PASSED
test_dsl_fmha_manifest_is_verified_against_its_pin FAILED   Failed: DID NOT RAISE RuntimeError
test_dsl_fmha_unreachable_manifest_raises          FAILED   FileNotFoundError: [Errno 2] No such file
                                                            or directory: '.../x86_64/sm_100a/checksums.txt'
test_dsl_fmha_unpinned_arch_raises                 FAILED   Failed: DID NOT RAISE RuntimeError
```

The one that passes on `main` is the honest path, which is the point — the change doesn't alter it.

With the fix, the whole file passes:

```
$ pytest tests/test_artifacts.py -q
...........                                                              [100%]
11 passed in 30.26s
```

Run on Linux / Python 3.12, CPU only. This path is artifact plumbing, so it needs no GPU; I did not run the cute-dsl kernel tests.

## Reviewer Notes

Worth a look at the error text I chose for a pin miss — I named the arch and told the user to clear the cache, following `verify_symlinked_headers`, but you may prefer it to point at the publishing script instead.

I left `os.path.join` for `checksums_rel` alone rather than switching it to `safe_urljoin`, to keep the diff to the actual bug. It is the pre-existing spelling and behaves the same on the platforms FlashInfer supports.

AI-assisted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of DSL FMHA checksum manifests using pinned checksums.
  * Added clearer runtime errors for missing checksum pins, unavailable manifests, and tampered manifest data.
  * Prevented confusing file-not-found failures when checksum manifests cannot be retrieved.
  * Improved reliability when loading DSL FMHA artifacts across supported architectures.

* **Tests**
  * Added coverage for valid, tampered, unreachable, and unpinned checksum manifest scenarios.


Dev-next-gen:
@flashinfer-bot run

Dev-next-gen:
The one red check here is the "Test Results Summary / Report unauthorized PR status" step, not a test failure. The actual test jobs (Setup, matrix.name) are skipped, and that summary step exits 1 by design when a fork PR has not been authorized yet — its log says "CI skipped — pending authorization" and asks a ci-users member to comment @flashinfer-bot run to start CI. So the diff itself has nothing red against it; the checks just need someone to trigger the run. Happy to rebase or adjust if you would like anything changed first.

Dev-next-gen:
The only red check here is Test Results Summary, and it isn't a test failure — the job ends on `CI skipped — pending authorization`, because the workflow no longer starts CI on its own for an outside contributor.

@bkryu @aleozlx — could one of you kick it off with `@flashinfer-bot run`? The diff is Python only: a verification step in `flashinfer/attention/cute_dsl/fmha.py` and a test in `tests/test_artifacts.py`, no kernel or build change. Happy to re-ask after any commit I push.

PR #5066, opened 2026-09-09
fix(bench): use float32 cos_sin_cache in apply_rope_with_cos_sin_cache

### LINKED ISSUE #5025: [Bug] Unified benchmark creates invalid cache dtype for apply_rope_with_cos_sin_cache
### Before submitting

- [x] I have searched existing issues and did not find a solution or an existing report.

### Environment

- Repository: `flashinfer-ai/flashinfer`
- Branch: `main`
- Commit: `866acb62d31c8dfa9c1e8b1aebc8fea5957279b5`
- Hardware recorded by the committed sample output: NVIDIA B200

This is a deterministic benchmark-source issue rather than an environment-specific kernel failure. The repository's committed sample output reproduces it on both supported CLI input dtypes.

### Bug description

`testApplyRopeWithCosSinCache` creates `cos_sin_cache` with `dtype=input_dtype`. The routine parser only permits `float16` and `bfloat16` for `--input_dtype`, while `flashinfer.rope.apply_rope_with_cos_sin_cache` requires the cache to be `torch.float32` and raises otherwise.

As a result, every valid invocation of this unified benchmark routine fails before timing and produces no benchmark result. The cache should always be created with `dtype=torch.float32`, independently of the Q/K input dtype. The cache contribution in `problem_bytes` should consequently use 4 bytes per element rather than `input_dtype.itemsize`.

### Minimal reproduction

```bash
python benchmarks/flashinfer_benchmark.py \
  --routine apply_rope_with_cos_sin_cache \
  --batch_size 16 \
  --seq_len 1024 \
  --num_qo_heads 32 \
  --num_kv_heads 8 \
  --head_dim 128 \
  --input_dtype float16 \
  -vv \
  --generate_repro_command \
  --case_tag apply_rope_cos_sin_cache
```

The same failure occurs with `--input_dtype bfloat16`.

### Error messages / logs

```text
[ERROR] Error running test: --routine apply_rope_with_cos_sin_cache --batch_size 16 --seq_len 1024 --num_qo_heads 32 --num_kv_heads 8 --head_dim 128 --input_dtype float16 -vv --generate_repro_command --case_tag "apply_rope_cos_sin_cache"
[ERROR] Error: cos_sin_cache should be float32
```

### Additional context

- [Benchmark creates the cache using `input_dtype`](https://github.com/flashinfer-ai/flashinfer/blob/866acb62d31c8dfa9c1e8b1aebc8fea5957279b5/benchmarks/routines/rope.py#L878-L883)
- [Public API validates that the cache is FP32](https://github.com/flashinfer-ai/flashinfer/blob/866acb62d31c8dfa9c1e8b1aebc8fea5957279b5/flashinfer/rope.py#L1205-L1210)
- [Committed sample output contains the FP16 and BF16 failures](https://github.com/flashinfer-ai/flashinfer/blob/866acb62d31c8dfa9c1e8b1aebc8fea5957279b5/benchmarks/samples/sample_testlist_output.txt#L1125-L1138)
- The routine was introduced in commit [`1748eb59`](https://github.com/flashinfer-ai/flashinfer/commit/1748eb59a76a563c64d77aebd2547a7a0d7329e9).

aeichler-ac:
## Summary
- `testApplyRopeWithCosSinCache` allocated `cos_sin_cache` with `input_dtype` (fp16/bf16), but `flashinfer.rope.apply_rope_with_cos_sin_cache` requires float32 and raised before any timing.
- Allocate the cache as `torch.float32` and count it as 4 bytes/elem in `problem_bytes`.

Fixes #5025

## Test plan
- [x] DGX Spark (GB10): issue repro with `--input_dtype float16` -> `[PERF] median 1.571 ms` (no `cos_sin_cache should be float32`)
- [x] Same with `--input_dtype bfloat16` -> `[PERF] median 1.572 ms`

## Summary by CodeRabbit

- **Benchmarks**
  - Updated rotary-position-embedding benchmark calculations to use the required 32-bit floating-point cache format.
  - Adjusted memory-bandwidth measurements to accurately reflect the cache’s fixed 4-byte element size.

### DIFF
--- a/benchmarks/routines/rope.py
+++ b/benchmarks/routines/rope.py
@@ -878,8 +878,9 @@ def testApplyRopeWithCosSinCache(args):
     # Precomputed cos_sin_cache: (max_seq_len, rotary_dim)
     # First half is cos, second half is sin
     max_seq_len = seq_len
+    # API requires FP32 cache regardless of Q/K dtype (#5025).
     cos_sin_cache = torch.randn(
-        max_seq_len, rotary_dim, dtype=input_dtype, device=device
+        max_seq_len, rotary_dim, dtype=torch.float32, device=device
     )
 
     # positions: (total_tokens,)
@@ -936,7 +937,7 @@ def testApplyRopeWithCosSinCache(args):
                 * num_kv_heads
                 * head_dim
                 * input_dtype.itemsize  # k read
-                + max_seq_len * rotary_dim * input_dtype.itemsize  # cos_sin_cache read
+                + max_seq_len * rotary_dim * 4  # cos_sin_cache read (float32)
                 + total_tokens * 8  # positions read (int64)
                 + total_tokens
                 * num_qo_heads


bkryu (approved):
Thanks @aeichler-ac for the quick fix!

PR #5067, opened 2026-09-09
fix(jit): resolve CUDA lib dirs and versioned cudart for pip toolkits

### LINKED ISSUE #5064: [Bug] JIT link fails with `cannot find -lcudart` against a pip-installed CUDA toolkit: cpp_ext.py hardcodes $cuda_home/lib64, cu13 wheels ship lib/
### Summary

With a **pip-installed CUDA toolkit** (`nvidia-cuda-*` cu13 wheels) and no
system CUDA under `/usr/local/cuda`, every JIT link step fails:

```
/usr/bin/x86_64-linux-gnu-ld.bfd: cannot find -lcudart: No such file or directory
collect2: error: ld returned 1 exit status
ninja: build stopped: subcommand failed.
```

Two independent causes, both in `flashinfer/jit/cpp_ext.py`:

1. `generate_ninja_build_for_op()` hardcodes the link search paths as
   [`-L$cuda_home/lib64` and `-L$cuda_home/lib64/stubs`](https://github.com/flashinfer-ai/flashinfer/blob/main/flashinfer/jit/cpp_ext.py#L256-L257).
   The cu13 wheels lay libraries out in **`lib/`**; there is no `lib64/` at all,
   so both `-L` entries point at nonexistent directories.
2. A follow-on that bites as soon as cause 1 is fixed: the wheels ship only
   **versioned** sonames (`libcudart.so.13`, `libnvrtc.so.13`), with no
   unversioned `libcudart.so` / `libnvrtc.so` symlink for `ld` to resolve
   `-lcudart` / `-lnvrtc` against. (In our own run cause 2 was already masked --
   we had created the unversioned symlinks earlier for CMake's `FindCUDA` while
   building vLLM -- so what we actually *observed* is cause 1. Cause 2 is read
   off the wheel manifests below, and would hit anyone who applies only the
   `lib64` fix.)

Straight from the wheel's own `RECORD` (`nvidia_cuda_runtime-13.2.86.dist-info`),
the only two `libcudart` entries are:

```
nvidia/cu13/lib/libcudart.so.13
nvidia/cu13/lib/libcudart_static.a
```

and `nvidia_cuda_nvrtc-13.2.86.dist-info/RECORD`:

```
nvidia/cu13/lib/libnvrtc-builtins.so.13.2
nvidia/cu13/lib/libnvrtc.so.13
```

`get_cuda_path()` lands on that wheel directory in the normal case: either
`CUDA_HOME` is pointed at `<venv>/lib/pythonX.Y/site-packages/nvidia/cu13`, or
it is unset and `which nvcc` resolves to `nvidia/cu13/bin/nvcc` from the
`nvidia-cuda-nvcc` wheel, whose `dirname(dirname(...))` is the same directory.

This is still present on `main` at the time of writing.

### Why it is hard to diagnose

When this happens under vLLM the visible error is a generic `Ninja build failed`
plus ~97 compile lines, and it surfaces **minutes after a successful weight
load**, during memory profiling — so it reads like a runtime fault rather than a
link error. The actual cause is one `ld` line buried in the dump. It took a
`grep -nE "FAILED:|cannot find -l"` over a 2 MB log to find it.

### Reproduction

Environment (RTX PRO 6000 Blackwell Workstation, SM120, Ubuntu 26.04, driver 595.84):

| | |
|---|---|
| `flashinfer-python` | `0.6.16.post3` and `0.6.18` (both reproduce) |
| `torch` | `2.13.0+cu130` |
| CUDA toolkit | pip wheels, `nvidia-cuda-{runtime,nvcc,crt,nvrtc}==13.2.86`, `nvidia-nvvm==13.2.86` |
| system CUDA | none (`/usr/local/cuda` does not exist) |

```bash
python -m venv .venv && . .venv/bin/activate
pip install torch --index-url https://download.pytorch.org/whl/cu130
pip install flashinfer-python==0.6.18
pip install "nvidia-cuda-nvcc==13.2.86" "nvidia-cuda-crt==13.2.86" \
            "nvidia-nvvm==13.2.86" "nvidia-cuda-runtime==13.2.86" \
            "nvidia-cuda-nvrtc==13.2.86"
export CUDA_HOME=$PWD/.venv/lib/python3.13/site-packages/nvidia/cu13
ls "$CUDA_HOME/lib64"               # -> No such file or directory
ls "$CUDA_HOME/lib/libcudart.so"    # -> No such file or directory (only .so.13)
ls "$CUDA_HOME/lib/stubs"           # -> No such file or directory
# then trigger any JIT op; the link step fails as above
```

The failing link line, verbatim from our log (trimmed to the flags — the full
line lists 97 objects):

```
[97/97] c++ ... ~/.cache/flashinfer/0.6.16.post3/120f/cached_ops/fused_moe_120/*.cuda.o \
  -shared \
  -L<venv>/lib/python3.13/site-packages/nvidia/cu13/lib64 \
  -L<venv>/lib/python3.13/site-packages/nvidia/cu13/lib64/stubs \
  -lcudart -lcuda -lnvrtc \
  -o ~/.cache/flashinfer/0.6.16.post3/120f/cached_ops/fused_moe_120/fused_moe_120.so
FAILED: [code=1] .../fused_moe_120/fused_moe_120.so
/usr/bin/x86_64-linux-gnu-ld.bfd: cannot find -lcudart: No such file or directory
/usr/bin/x86_64-linux-gnu-ld.bfd: cannot find -lnvrtc: No such file or directory
collect2: error: ld returned 1 exit status
ninja: build stopped: subcommand failed.
```

`-lcuda` is the only one that resolves, and only because the driver's
`/usr/lib/x86_64-linux-gnu/libcuda.so` is on `ld`'s default search path -- the
`$cuda_home/lib64/stubs` directory the flags point at does not exist under
either name (`lib/stubs` is absent from the wheels too).

### Workaround we are using

A `lib64` symlink plus unversioned-soname symlinks, per venv, recreated after
every CUDA-version change:

```bash
CU=$CUDA_HOME
ln -sfn lib "$CU/lib64"
cd "$CU/lib" && for f in *.so.*; do
  base=${f%%.so.*}.so; [ -e "$base" ] || ln -s "$f" "$base"
done
```

(Our launcher recreates the `lib64` link unconditionally at boot, because a
`pip install --upgrade` of the CUDA wheels silently removes it.)

### Suggested fix

In `generate_ninja_build_for_op()`, build the `-L` list from directories that
actually exist rather than assuming `lib64`:

```python
libdirs = [d for d in ("lib64", "lib") if (Path(cuda_home) / d).is_dir()]
ldflags = ["-shared"]
for d in libdirs:
    ldflags += [f"-L$cuda_home/{d}", f"-L$cuda_home/{d}/stubs"]
```

That alone fixes cause 1. Cause 2 needs one of:

- link the runtime by soname — `-l:libcudart.so.13` (or discover the versioned
  filename by globbing `libcudart.so.*`) when no unversioned `libcudart.so` is
  present; or
- link `libcudart_static.a`, which the wheel does ship; or
- at minimum, detect the situation and raise a clear error naming the missing
  `libcudart.so` instead of letting `ld` fail 97 objects deep.

I am happy to open a PR for the `-L` probing plus the `-l:libcudart.so.<major>`
fallback if that is the direction you would prefer.

---

Filed with AI assistance; the environment details, the wheel `RECORD` contents,
and the link line above were all read off this machine rather than reconstructed.

aeichler-ac:
## Summary
- Probe `$cuda_home/lib64` and `$cuda_home/lib` instead of hardcoding `lib64` in `generate_ninja_build_for_op` (pip `nvidia-cuda-*` cu13 wheels only ship `lib/`).
- Add stubs `-L` entries only when the stubs directory exists.
- When `libcudart.so` / `libnvrtc.so` are missing, link by versioned soname (`-l:libcudart.so.13`) so `ld` can resolve the wheel layout.
- Avoid shadowing the JIT op `name` when iterating lib dir names (would rewrite the module to `lib`).

Fixes #5064

## Test plan
- [x] DGX Spark (GB10): `CUDA_HOME=<venv>/.../nvidia/cu13` full JIT `apply_rope_with_cos_sin_cache` succeeds; `build.ninja` has `name = rope`, `-L$cuda_home/lib`, `-l:libcudart.so.13`
- [x] Same box: `CUDA_HOME=/usr/local/cuda` full JIT still succeeds; `build.ninja` has `-L$cuda_home/lib64`, stubs, `-lcudart`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved CUDA library discovery across installations using either `lib` or `lib64` directory layouts.
  - JIT compilation now supports versioned CUDA libraries when unversioned links are unavailable.
  - CUDA runtime and NVRTC linker settings are resolved more reliably from supplied configuration.
  - Added a clear error when the expected CUDA library directories cannot be found.

### DIFF
--- a/flashinfer/jit/cpp_ext.py
+++ b/flashinfer/jit/cpp_ext.py
@@ -44,6 +44,39 @@ def _get_glibcxx_abi_build_flags() -> List[str]:
     return glibcxx_abi_cflags
 
 
+def _cuda_lib_dirs(cuda_home: str) -> List[str]:
+    """Return existing CUDA library directory names under ``cuda_home``.
+
+    System toolkits ship ``lib64/``; pip ``nvidia-cuda-*`` cu13 wheels ship
+    ``lib/`` only. Hardcoding ``lib64`` makes every JIT link fail with
+    ``cannot find -lcudart`` against a wheel-only toolkit (#5064).
+    """
+    return [name for name in ("lib64", "lib") if (Path(cuda_home) / name).is_dir()]
+
+
+def _cuda_shared_lib_flag(cuda_home: str, libdirs: List[str], stem: str) -> str:
+    """Return an ``ld`` flag that resolves ``lib{stem}`` under ``cuda_home``.
+
+    Pip CUDA wheels often ship only a versioned soname (``libcudart.so.13``)
+    without an unversioned ``libcudart.so`` symlink. Plain ``-lcudart`` then
+    fails even when ``-L`` points at the right directory; ``-l:libcudart.so.13``
+    links by exact filename instead.
+    """
+    candidates: List[Path] = []
+    for name in libdirs:
+        lib_dir = Path(cuda_home) / name
+        unversioned = lib_dir / f"lib{stem}.so"
+        if unversioned.exists():
+            return f"-l{stem}"
+        candidates.extend(sorted(lib_dir.glob(f"lib{stem}.so.*")))
+    if not candidates:
+        # Keep the conventional flag so the linker error still names the library.
+        return f"-l{stem}"
+    # Prefer the shortest versioned soname (libcudart.so.13 over .so.13.0.96).
+    chosen = min(candidates, key=lambda path: (len(path.name), path.name))
+    return f"-l:{chosen.name}"
+
+
 @functools.cache
 def get_cuda_path() -> str:
     cuda_home = os.environ.get("CUDA_HOME") or os.environ.get("CUDA_PATH")
@@ -252,11 +285,22 @@ def generate_ninja_build_for_op(
     cflags = build_cflags(common_cflags, extra_cflags)
     cuda_cflags = build_cuda_cflags(common_cflags, extra_cuda_cflags)
 
-    ldflags = [
-        "-shared",
-        "-L$cuda_home/lib64",
-        "-L$cuda_home/lib64/stubs",
-        "-lcudart",
+    libdirs = _cuda_lib_dirs(cuda_home)
+    if not libdirs:
+        raise RuntimeError(
+            f"No CUDA library directory found under {cuda_home!r} "
+            f"(expected lib64/ or lib/). pip nvidia-cuda-* cu13 wheels use lib/; "
+            f"system toolkits typically use lib64/ (#5064)."
+        )
+
+    ldflags = ["-shared"]
+    for libdir in libdirs:
+        ldflags.append(f"-L$cuda_home/{libdir}")
+        stubs = Path(cuda_home) / libdir / "stubs"
+        if stubs.is_dir():
+            ldflags.append(f"-L$cuda_home/{libdir}/stubs")
+    ldflags += [
+        _cuda_shared_lib_flag(cuda_home, libdirs, "cudart"),
         "-lcuda",
     ]
 
@@ -266,6 +310,16 @@ def generate_ninja_build_for_op(
 
     if extra_ldflags is not None:
         ldflags += extra_ldflags
+
+    # Rewrite -lcudart / -lnvrtc from callers/env so pip wheels without
+    # unversioned .so symlinks still link (#5064).
+    resolved = []
+    for flag in ldflags:
+        if flag in ("-lcudart", "-lnvrtc"):
+            resolved.append(_cuda_shared_lib_flag(cuda_home, libdirs, flag[2:]))
+        else:
+            resolved.append(flag)
+    ldflags = resolved
 
     cxx = os.environ.get("CXX", "c++")
     nvcc = os.environ.get("FLASHINFER_NVCC", "$cuda_home/bin/nvcc")
--- a/tests/jit/test_jit_cpp_ext.py
+++ b/tests/jit/test_jit_cpp_ext.py
@@ -58,6 +58,110 @@ def test_generate_ninja_uses_sccache_compatible_nvcc_depfile_flag(
     assert "--dependency-output" not in ninja
 
 
+def _ldflags_block(ninja: str) -> str:
+    """Return the multiline ``ldflags = ...`` assignment from a ninja file."""
+    lines = ninja.splitlines()
+    start = next(i for i, line in enumerate(lines) if line.startswith("ldflags ="))
+    end = start + 1
+    while end < len(lines) and lines[end].startswith("    "):
+        end += 1
+    return "\n".join(lines[start:end])
+
+
+def test_generate_ninja_supports_pip_cuda_wheel_lib_layout(monkeypatch, tmp_path):
+    # pip nvidia-cuda-* cu13 wheels ship lib/ only with versioned sonames (#5064).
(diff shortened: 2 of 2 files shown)


Cyxuan0311:
Verified this approach independently on my box **(RTX 4050 Laptop, SM89 / CUDA 12.8, WSL2).** Full disclosure: I don't have a physical pip-cu13 toolkit here, so this is logic-level verification of generate_ninja_build_for_op() output against simulated $cuda_home layouts, not an end-to-end JIT link against real cu13 wheels:

- pip cu13 layout (lib/ only, libcudart.so.13/libnvrtc.so.13, no lib64/): now emits -L$cuda_home/lib + -l:libcudart.so.13 — previously pointed at nonexistent lib64/ with bare -lcudart. Fixed.
- system toolkit (lib64/ + unversioned .so + stubs/): unchanged (-L.../lib64, -L.../lib64/stubs, plain -lcudart). No regression.
- both lib64/+lib/ present: both -L emitted, lib64 first. Good.
- caller/env-supplied -lnvrtc with only versioned soname present: rewritten to -l:libnvrtc.so.13. Good.
- neither dir present: clear RuntimeError naming the expected dirs instead of a linker failure 97 objects deep. Good.

Core case as a regression test (same style as the existing tmp_path tests in **tests/jit/test_jit_cpp_ext.py)**:

```python
def test_generate_ninja_supports_pip_cuda_wheel_lib_layout(monkeypatch, tmp_path):
    # pip nvidia-cuda-* cu13 wheels ship lib/ only with versioned sonames (#5064).
    cuda_home = tmp_path / "pip_cuda"
    (cuda_home / "lib").mkdir(parents=True)
    (cuda_home / "lib" / "libcudart.so.13").touch()
    (cuda_home / "lib" / "libnvrtc.so.13").touch()
    monkeypatch.setattr(cpp_ext, "get_cuda_path", lambda: str(cuda_home))
    ...
    assert "-L$cuda_home/lib $" in ldflags
    assert "-L$cuda_home/lib64" not in ldflags
    assert "-l:libcudart.so.13" in ldflags
 ```

Actual ldflags block produced on my box for that layout:

```
ldflags = -shared $
    -L$cuda_home/lib $
    -l:libcudart.so.13 $
    -lcuda
```

Implementation-wise this matches the PR line-for-line on all six points I checked (_cuda_lib_dirs order, -l: shortest-soname pick, RuntimeError, conditional stubs/, -lcudart/-lnvrtc rewrite scope, -lcuda untouched).

**One gap:** this PR has no regression tests. I have 4 ready (pip layout / system layout / -lnvrtc rewrite / missing-dir error). Happy to contribute them here — let me know if you'd like me to push them to this branch or open a companion PR against your fork. Also happy to help with rebase if needed.

aeichler-ac:
Thanks for the independent checks and the test sketch. I added the four
regression cases to tests/jit/test_jit_cpp_ext.py on this branch (pip lib/
layout, system lib64+stubs, -lnvrtc rewrite, missing-dir RuntimeError) and
ran them on DGX Spark. No need for a companion PR unless you spot a gap.

Issue #4037, opened 2026-07-17
Refactor Batch MLA Wrapper and Backends


saltyminty:
Refactor Batch MLA into an operation-owned private package while preserving the existing public `flashinfer.mla` API and compatibility bindings.
- Move Batch MLA orchestration into `flashinfer/mla/_batch_mla/`, with `_core.py` and `_planning.py` owning the public adapters and planning lifecycle.
- Split concrete implementations into explicit backend modules under `_batch_mla/_backends/`, with narrowly scoped `_fa_common.py` and `_layout.py` helpers.
- Connect the CuTe DSL, TRTLLM-GEN, and XQA backends to `BatchMLAPagedAttentionWrapper` alongside FA2, FA3, and CUTLASS, so all six follow the same backend selection, `plan()`, and `run()` lifecycle.
- Preserve the existing functional MLA APIs as compatibility adapters over the same backend-owned planning and execution paths used by the wrapper, rather than maintaining separate backend implementations.
- Keep sparse MLA in the top-level MLA core for now, while extracting shared backend selection and preserving legacy imports, signatures, and fallback behavior.

Draft PR: https://github.com/flashinfer-ai/flashinfer/pull/4031

Companion design document: https://docs.google.com/document/d/1zE9FsvjcRwXFV99kqAylh16JLAAc8cZRC0ojAI02bLg/edit

PR #4031, opened 2026-07-17
refactor(mla): redesign batch MLA backend internals

saltyminty:
## 📌 Description

This PR reorganizes Batch MLA around isolated backend implementations and
provides consistent wrapper and functional APIs across the supported dense MLA
backends.

### Highlights

- **Isolated backend implementations:** FA2, FA3, CUTLASS, TRTLLM-GEN, XQA,
  CuTe DSL monolithic, and CuTe DSL modular now live in backend-specific modules
  that own their support checks, planning/preparation, state, and launch logic.

- **Expanded wrapper API:** `BatchMLAPagedAttentionWrapper` now supports
  `fa2`, `fa3`, `cutlass`, `trtllm-gen`, `xqa`,
  `cute-dsl-monolithic`, and `cute-dsl-modular`.
  `cute-dsl` remains a family selector between its two implementations.

- **General wrapper auto fallback:** `backend="auto"` ranks every wrapper
  backend and tries candidates until one reports that it supports the request.
  Fallback occurs only for the typed unsupported-backend result; invalid input
  and unexpected failures remain visible. The current ordering is primarily
  architecture-based. Workload- and performance-informed ranking based on the
  complete problem configuration is deferred, but the new structure makes it
  possible without modifying every backend.

- **Unified functional API:** `batch_mla_paged_attention` is the supported
  one-shot functional entrypoint for explicit FA2, FA3, CUTLASS, TRTLLM-GEN,
  XQA, and CuTe DSL execution. Functional `auto` retains its existing
  TRTLLM-GEN/CuTe autotuning policy and remains separate from wrapper `auto`.

- **Canonical planning and layouts:** `MLAPlanMetadata` provides canonical
  CSR/dense planning metadata, while `query_layout` and `kv_cache_layout`
  declare packed or split runtime representations. Every wrapper backend accepts CSR, dense, or equivalent
  dual metadata; the selected backend's native representation is derived
  lazily. Packed query and KV inputs are lowered into zero-copy views when
  possible.

- **Compatibility and deprecations:** Existing public imports remain available,
  while the following legacy forms are deprecated:
  - direct TRTLLM-GEN and XQA functional entrypoints, in favor of
    `batch_mla_paged_attention`;
  - positional wrapper arguments and flat planning metadata, in favor of
    keyword arguments and `MLAPlanMetadata`;
  - legacy split `q_nope`/`q_pe` and `ckv_cache`/`kpe_cache` arguments,
    whether positional or keyword, in favor of structural `query` and
[...]

PR #5070, opened 2026-09-09
feat(mla): add planned TRTLLM-GEN, XQA, and CuTe DSL backends

saltyminty:
## 📌 Description

Add explicit planned `BatchMLAPagedAttentionWrapper` support for the
TRTLLM-GEN, XQA, and CuTe DSL MLA backends.

- register `trtllm-gen`, `xqa`, `cute-dsl-monolithic`, and
  `cute-dsl-modular` as explicit-only planned backends
- retain `cute-dsl` as a thin compatibility selector that lowers to a
  concrete CuTe DSL backend and falls back only on typed unsupported results
- acquire backend executables during `plan()` and reuse backend-owned state
  during `run()`
- extend the shared planned-wrapper contracts for PDL, sinks,
  skip-softmax-threshold scaling, and BMM scales
- preserve existing automatic backend selection and direct backend APIs
- document the resulting backend and CUDA Graph lifecycle

## 🔍 Related Issues

This is part of a series of PRs to break up #4031.

Follow up to https://github.com/flashinfer-ai/flashinfer/pull/4697, independent of https://github.com/flashinfer-ai/flashinfer/pull/5041

Tracking Issue: https://github.com/flashinfer-ai/flashinfer/issues/4037

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

The complete changed-file hook set passed, including mypy, Ruff check, Ruff
format, file-integrity checks, and whitespace checks. The repository-wide
`--all-files` target was not run.

## 🧪 Tests

Targeted validation:

- SM100: `tests/attention/test_mla_wrapper.py` — 103 passed, 3 expected
  architecture skips
- SM120: `tests/attention/test_mla_wrapper.py` — 97 passed, 9 expected
  architecture skips
- changed-file pre-commit hooks and `git diff --check` passed

## 🔬 Experimental Track

```experimental-tests
#
#
```

## Reviewer Notes

Please focus on the planned-wrapper/backend ownership boundary, the narrow
typed-error fallback used by the `cute-dsl` compatibility selector, and the
fixed-pointer CUDA Graph lifecycle. Proper cross-backend test unification is
intentionally deferred.

## New Features
- Added MLA backend options for TRTLLM-GEN, XQA, and CuTe DSL (monolithic and modular).
- Added support for attention sinks, softmax skipping, and scalar or tensor BMM scaling where supported.
- Added backend capability reporting and CUDA Graph replanning support for additional backends.
- Added support for dense device metadata requirements in applicable CUDA Graph workflows.

## Bug Fixes
- Improved validation and handling of unsupported MLA configurations.

## Documentation
- Expanded backend architecture documentation, including requirements, metadata, lifecycle, and CUDA Graph behavior.


saltyminty:
On the detected API change: we added more options for an existing parameter and added new keyword-only optional parameters. None are API breaking.

qsang-nv:
**Scope note:** I only reviewed the XQA backend (`flashinfer/mla/_batch_mla/_backends/xqa_backend.py`), not the TRTLLM-GEN or CuTe DSL slices. The launch mapping itself checks out against the raw `xqa_mla` op.

**1. `max_q_len != 1` rejects legal upper-bound metadata.** `_validate_dense_metadata` treats `max_q_len` as a caller-declared upper bound and only rejects values *below* the actual maximum; the TRTLLM-GEN and CuTe DSL backends added here follow that. The check is also redundant — `plan()` already verifies `torch.equal(cum_seq_lens_q, arange(batch_size + 1))`, and `max_q_len` has no other use here (the kernel's `max_seq_len` comes from `block_tables.shape[1] * page_size`). Trigger is specific to **canonical dense metadata with an explicit `max_q_len`** (CSR derives it exactly and cannot overestimate): `MLAPlanMetadata.dense(arange(B+1), block_tables, seq_lens, max_q_len=2)` with all actual lengths 1 is accepted by the metadata layer and the other backends, refused here. Since `!= 1` runs before the `arange` check, the caller is told `requires exactly one query token per request, got 2` while every request length *is* one — "got 2" is the bound, not a length. No test passes an overestimated `max_q_len`.

**2. The architecture gate does not reuse the repository's authoritative support condition.** `_validate_xqa_device_capability` checks only `(major, minor) in ((12, 0), (12, 1))`, while the direct `xqa_batch_decode_with_kv_cache_mla` gates on `is_sm12x_supported`, which also requires CUDA >= 12.8 (SM120a) / >= 12.9 (SM121a) and is written as `major == 12` so future SM12x minors stay covered. A configuration the direct API refuses at entry can therefore proceed past the typed preflight into module compilation and fail outside it, and a minor variant the helper intentionally supports is refused. The new tests skip on `is_sm12x_supported`, i.e. the test guard is stricter than what the production preflight enforces.

**3. `cast(float, ...)` is a runtime no-op, so a non-`float` scalar reaches the tensor slot.** `_validate_scalar_bmm_scales` and `MLAInputContract.validate_run_options` accept any `numbers.Real` except `bool`, and the value is forwarded as `cast(float, resolved_bmm1_scale)`, which converts nothing. The registered op then separates scalar and tensor scales using `isinstance(..., float)` without normalizing accepted numeric scalars first:

```python
1.0 if isinstance(q_scale, torch.Tensor) else q_scale,   # double qScale
None if isinstance(q_scale, float) else q_scale,          # Optional<TensorView> qScaleTensor
```

So `run(..., bmm1_scale=1, bmm2_scale=1.0)` — an `int`, which passes both validation layers — reaches the `Optional<TensorView>` parameter of `xqa_wrapper_mla`, surfacing as a binding-level type error unrelated to the cause. The CuTe DSL path converts via `state.Float32(float(...))` and is unaffected. Reasoned statically, not run.

flashinfer/mla/_batch_mla/_backends/trtllm_gen_backend.py
```diff
@@ -0,0 +1,673 @@
...
+    *,
+    device: torch.device,
+    block_tables: torch.Tensor,
+    seq_lens: torch.Tensor,
+    cum_seq_lens_q: torch.Tensor,
+) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+    return (
+        block_tables.to(
+            device=device, dtype=torch.int32, non_blocking=True
+        ).contiguous(),
```
bkryu, line 192: Under use_cuda_graph=True, whether a captured plan aliases the caller's dense metadata or pins a hidden snapshot is data-dependent: device/int32/contiguous input passes through, anything else — including CPU metadata, which the wrapper explicitly permits — gets copied here (and in the generic _dense_on_device path the other backends use). 

The architecture doc promises the opposite workflow: "TRTLLM-GEN preserves the supplied dense tensors" and "later in-place length updates stay within the compiled contract." A user who plans a graph with CPU metadata will make in-place updates that every replay silently ignores.

flashinfer/mla/_batch_mla/_backends/_cute_dsl_common.py
```diff
@@ -0,0 +1,604 @@
...
+        raise _BackendPlanUnsupportedError(
+            "cute-dsl backend expects max_q_len to be at least the maximum "
+            f"query length {actual_max_q_len}, got {max_q_len}."
+        )
+    if block_tables.ndim != 2 or block_tables.shape[0] != batch_size:
+        raise _BackendPlanUnsupportedError(
+            "cute-dsl backend expects rank-2 block_tables with batch dimension "
+            f"{batch_size}, got {tuple(block_tables.shape)}."
+        )
+    if block_tables.shape[1] == 0:
```
bkryu, line 177: The direct cute-dsl path requires block_table_width % (128 // page_size) == 0 (_check_trtllm_gen_mla_shape with require_aligned_block_table=True); this planned path checks only rank/batch and positive width, and resolves the dense table at its native width. The kernel's 128-row KV tile spans 128 // page_size table entries per step, so an unaligned width can read past the end of a table row. Either resolve the dense metadata with table_width_alignment=128 // page_size (as the CUTLASS backend does) or add the modulus check here (as the XQA backend does at its equivalent site). The architecture doc's claim that table width "is aligned to 128 / page_size for backends whose native kernel requires that ownership convention" becomes true with this fix.

flashinfer/mla/_batch_mla/_backends/trtllm_gen_backend.py
```diff
@@ -0,0 +1,673 @@
...
+            None,
+            skip_softmax_threshold_scale_factor,
+            True,
+            lse,
+            lse_stride_tokens,
+            lse_stride_heads,
+            False,
+            None,
+            0,
+            None,
```
bkryu: The huge positional input here seems a bit error-prone. Maybe make an agent call this with keyword inputs?

bkryu (commented):
Thanks @saltyminty. Left a few comments.

General comments: Instead of `functools.cache`, can we use `functools.lru_cache` with a finite cache size? Some will have only a size of 1 even with the unbound cache, but would be just good practice to do so throughout. Today we get questions about unbound `functools.cache` in our library.

flashinfer/mla/_batch_mla/_backends/trtllm_gen_backend.py
saltyminty: This binding doesn't allow us to use keyword inputs, but added each comments labeling each input.

flashinfer/mla/_batch_mla/_backends/_cute_dsl_common.py
saltyminty: Fixed.

flashinfer/mla/_batch_mla/_backends/trtllm_gen_backend.py
saltyminty: Fixed.

qsang-nv:
The three earlier findings are fixed with regression tests — thanks. One remaining point, not blocking.

**CSR-derived table width is not aligned for XQA, and the doc states the opposite.** XQA plans through `args.native_device_dense()`, so `_derive_dense_from_csr` runs with `table_width_alignment=None` and yields `max(1, max_pages)`, while `plan()` requires `block_tables.shape[1] % (128 // page_size) == 0`. Survival therefore depends on batch shape: with `page_size=64` a largest-request span of 2 pages passes, but a single 64-token request gives width 1 and is rejected with `XQA MLA block_tables width must be a positive multiple of 2`. In graph mode CSR is rejected outright by the new `require_cuda_graph_dense_metadata` guard.

The doc updated in this commit says the resolver "pads CSR-derived tables" for backends requiring `128 / page_size` alignment, and that "non-graph plans continue accepting CPU and CSR metadata". XQA does require that alignment — it enforces it in its own `plan()` — but requests the unaligned resolver. CuTe DSL moved to `device_dense(table_width_alignment=128 // args.page_size)` in the same commit and has `test_cute_dsl_planned_pads_csr_table`; XQA has neither.

saltyminty:
@flashinfer-bot run tests/attention/test_mla_wrapper.py

saltyminty:
@flashinfer-bot run

qsang-nv (approved):

bkryu (approved):

PR #5071, opened 2026-09-09
perf(attention): stop doubling the paged K/V offset registers when K/V strides match

aleozlx:
## 📌 Description

#3684 gave the FA2 paged kernels independent K/V page strides so the asymmetric NVFP4 VO-split
cache (`head_dim_qk=512`, `head_dim_vo=256`) could address its two pools separately. It did so
unconditionally: the per-thread offset array `thr_local_kv_offset` was split into `_k` and `_v`
copies of `NUM_MMA_KV * KV_THR_LAYOUT_COL / 2 / NUM_WARPS_Q` 64-bit entries each, and the
persistent kernel gained a second `prefetch_offest` pass to fill it.

Every symmetric configuration pays for that. At `head_dim` 128 the array is 16 entries, so the
split costs ~32 extra registers per thread on the hot producer path, and the occupancy loss shows
up directly in decode and prefill latency.

This gates the second array on `KernelTraits::SEPARATE_KV_OFFSETS`, true for the two families that
can actually see different K/V strides:

- `head_dim_qk != head_dim_vo` — the asymmetric NVFP4 cache #3684 added, and
- `head_dim_vo >= 512` — large heads, which reach the shared-KV-smem on-the-fly producer and route
  V through the V strides there (`test_batch_prefill_paged_shared_kv_smem_unequal_kv_strides` covers
  exactly this with symmetric 512/512 fp16 pools).

Everywhere else the V rows sit at the K offsets, so the V producer reads the K array and the V
offsets never reach a register. `csrc/batch_prefill.cu` regains the K/V stride-equality check for
that remaining set — the host-side contract that makes the compile-time choice sound. It was dropped
wholesale in #3684; the decode launcher never dropped its equivalent.

`persistent.cuh` is **deliberately left alone**, though the same transformation applies there. The
holistic `BatchAttention` launcher (`csrc/batch_attention.cu:90-101`) reads K and V strides from two
independent tensors and never compares them, so gating the persistent kernel's V offsets without
also adding that host check would make it address V through the K offsets *silently* rather than
fail loudly. That is a second contract change on a path this PR does not measure, so it belongs in
its own change. An earlier revision of this PR did include it; that was a mistake.

## 🔍 Related Issues

Fixes NVBug 6634590 (`BatchDecodeWithPagedKVCacheWrapper` fa2_tc batch-64 latency regression) and
NVBug 6634592 (`BatchPrefillWithPagedKVCacheWrapper` fa2 GPT-OSS latency regression). Both were
bisected by the perf suite to `8f9ad200` (#3684). No GitHub issue exists for either.

### ✅ Pre-commit Checks

## 🧪 Tests

No new test is added: `test_batch_prefill_paged_shared_kv_smem_unequal_kv_strides` (added by #3684)
already pins the large-head unequal-stride behaviour this change has to preserve, and it is what
caught an earlier, too-broad version of this patch.

Validated on B300 (SM103), CUDA 13.0:

| build | fa2_tc decode median |
| --- | --- |
| `8f9ad200^` (pre-regression) | 0.801 / 0.804 ms |
| `8f9ad200` (first bad) | 0.859 ms |
| `main` (`dbb52c8a`) | 0.860 / 0.860 ms |
| `main` + this patch | 0.787 / 0.797 / 0.803 ms |

Benchmark: `BatchDecodeWithPagedKVCacheWrapper`, `--backends fa2_tc --page_size 16 --batch_size 64
--s_qo 1 --s_kv 8192 --num_qo_heads 64 --num_kv_heads 8 --head_dim_qk 128 --head_dim_vo 128
--q_dtype bfloat16 --kv_dtype bfloat16 --num_iters 100`, i.e. NVBug 6634590's own repro command.
The patch returns `main` to the pre-#3684 baseline.

Correctness on the same board:

```
tests/attention/test_batch_prefill_kernels.py -k 'paged and not fp8 and not nvfp4'
    5442 passed, 1140 skipped, 5760 xfailed
tests/attention/test_batch_decode_kernels.py -k 'not fp8 and not nvfp4'
    2972 passed
tests/attention/test_batch_prefill_kernels.py -k unequal_kv_strides
    4 passed
```

Note this fix is deliberately narrower than the regression: symmetric configurations at
`head_dim >= 512` still carry #3684's register cost, because that is where the shared-KV-smem
producer legitimately needs the V strides. Removing it there too would require dispatching on
runtime stride equality — a second kernel instantiation — which seemed worse than the benefit.

The CI B300 reports a larger step for these NVBugs (25.8% on decode) than this engineering-sample
board does; the bisected commit and the full recovery are the same.

AI-assisted (Claude Opus 5): root-cause, patch, and B300 validation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to ensure matching K/V cache strides when using matching, smaller attention head dimensions.
  * Provides a clearer error when incompatible cache strides are supplied.

* **Performance**
  * Reduced memory overhead and improved paged attention efficiency for configurations that share K/V cache strides.
  * Preserved independent K/V stride support for asymmetric layouts and larger head dimensions.

### DIFF
--- a/csrc/batch_prefill.cu
+++ b/csrc/batch_prefill.cu
@@ -289,6 +289,20 @@ void BatchPrefillWithPagedKVCacheRun(TensorView float_workspace_buffer,
       RaggedParams, PagedParams, [&] {
         PagedParams params;
 
+        if constexpr (HEAD_DIM_QK == HEAD_DIM_VO && HEAD_DIM_VO < 512) {
+          // These caches address K and V through a single set of page offsets in the
+          // kernel (KernelTraits::SEPARATE_KV_OFFSETS), so the two pools must agree on
+          // strides. Independent K/V strides stay supported for the asymmetric (NVFP4
+          // VO-split) layout and for the large-head configurations that reach the
+          // shared-KV-smem on-the-fly producer.
+          for (int i = 0; i < paged_k_cache.ndim(); ++i) {
+            TVM_FFI_ICHECK_EQ(paged_k_cache.stride(i), paged_v_cache.stride(i))
+                << "k/v strides differ at dim " << i
+                << "; independent K/V strides require head_dim_qk != head_dim_vo or head_dim >= "
+                   "512";
+          }
+        }
+
         params.q = static_cast<DTypeQ*>(q.data_ptr());
         paged_kv_t<DTypeKV, IdType> paged_kv(
             num_kv_heads, page_size, HEAD_DIM_VO, batch_size, kv_layout,
--- a/include/flashinfer/attention/prefill.cuh
+++ b/include/flashinfer/attention/prefill.cuh
@@ -406,6 +406,16 @@ struct KernelTraits {
   static constexpr uint32_t REPACK_STRIDE_QK = HEAD_DIM_QK / upcast_size<DTypeQ_>();
   static constexpr uint32_t REPACK_STRIDE_VO = HEAD_DIM_VO / upcast_size<DTypeQ_>();
 
+  // Whether the K and V pools can carry different page strides. Two families can:
+  // the asymmetric NVFP4 VO-split cache (head_dim_qk != head_dim_vo), and the
+  // large-head configurations, which reach the shared-KV-smem on-the-fly producer
+  // (USE_VO_SPLIT requires head_dim >= 512) and route V through the V strides
+  // there. Everywhere else the launcher requires the two pools to share strides,
+  // so a V page offset is bit-identical to the K one; keeping a second in-register
+  // offset array for that case costs NUM_PAGED_KV_OFFSETS extra 64-bit registers
+  // per thread and drops occupancy (~26% on B300 fa2_tc decode, NVBug 6634590).
+  static constexpr bool SEPARATE_KV_OFFSETS = (HEAD_DIM_QK != HEAD_DIM_VO) || (HEAD_DIM_VO >= 512);
+
   static constexpr bool IsInvalid() {
     // The first clause prunes (CTA_TILE_Q, head_dim) pairs FA2DetermineCtaTileQ
     // never selects: it picks from {16, 32} when head_dim_vo >= 512 (VO-split),
@@ -4008,10 +4018,15 @@ __device__ __forceinline__ void BatchPrefillWithPagedKVCacheDevice(
       // Separate K/V page offsets: the K and V pools may carry different strides
       // (e.g. NVFP4 VO-split caches where head_dim_qk != head_dim_vo). The K/V-shared
       // smem path computes offsets on the fly, so the arrays collapse to [1] stubs there.
+      // With shared strides (KTraits::SEPARATE_KV_OFFSETS == false) the V offsets equal
+      // the K offsets, so the V array collapses to a [1] stub as well rather than
+      // doubling this thread's 64-bit offset registers.
+      constexpr uint32_t NUM_PAGED_V_OFFSETS =
+          KTraits::SEPARATE_KV_OFFSETS ? NUM_PAGED_KV_OFFSETS : 1;
       [[maybe_unused]] size_t
           thr_local_kv_offset_k[KTraits::USE_KV_SHARED_SMEM ? 1 : NUM_PAGED_KV_OFFSETS];
       [[maybe_unused]] size_t
-          thr_local_kv_offset_v[KTraits::USE_KV_SHARED_SMEM ? 1 : NUM_PAGED_KV_OFFSETS];
+          thr_local_kv_offset_v[KTraits::USE_KV_SHARED_SMEM ? 1 : NUM_PAGED_V_OFFSETS];
 
       uint32_t k_smem_offset_r = k_smem.template get_permuted_offset<UPCAST_STRIDE_K>(
                    get_warp_idx_kv<KTraits>(tid.z) * NUM_MMA_KV * 16 + 8 * (lane_idx / 16) +
@@ -4070,8 +4085,10 @@ __device__ __forceinline__ void BatchPrefillWithPagedKVCacheDevice(
               (lane_idx % KV_THR_LAYOUT_COL) * upcast_size<DTypeKV>() / fp4_pack_factor;
           thr_local_kv_offset_k[i] = paged_kv.protective_get_k_offset(
               page_iter, kv_head_idx, entry_idx, feat_idx, last_indptr);
-          thr_local_kv_offset_v[i] = paged_kv.protective_get_v_offset(
-              page_iter, kv_head_idx, entry_idx, feat_idx, last_indptr);
+          if constexpr (KTraits::SEPARATE_KV_OFFSETS) {
+            thr_local_kv_offset_v[i] = paged_kv.protective_get_v_offset(
+                page_iter, kv_head_idx, entry_idx, feat_idx, last_indptr);
+          }
         }
         page_produce_kv<false, KTraits>(&smem_storage, &k_smem_offset_w, paged_kv.k_data, 0,
                                         thr_local_kv_offset_k, chunk_size, warp_idx, lane_idx);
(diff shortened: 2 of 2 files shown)


aleozlx:
@flashinfer-bot run tests/attention/test_batch_prefill_kernels.py tests/attention/test_batch_decode_kernels.py tests/attention/test_batch_attention.py

aleozlx:
/bot run tests/attention/test_batch_prefill_kernels.py tests/attention/test_batch_decode_kernels.py tests/attention/test_batch_attention.py

aleozlx:
## PR Review Screening

**CI verdict:** ✅ auto-run ok
**Review category:** `live` (rule fired: C3.2 — a new `KernelTraits::SEPARATE_KV_OFFSETS` member gates the producer-path register layout for every paged instantiation in two shared attention headers, paired with a host-side contract change. The change is well-bounded and derived from existing template parameters rather than a new knob, so a maintainer may reasonably read it as a contained perf fix and downgrade to `human`)
**Blocking checks:** none
**Release blocker:** 🚨 candidate — fixes NVBug 6634590 (`BatchDecodeWithPagedKVCacheWrapper` fa2_tc batch-64) and NVBug 6634592 (`BatchPrefillWithPagedKVCacheWrapper` fa2 GPT-OSS) latency regressions, bisected to `8f9ad200` (#3684), which shipped in **v0.6.18** and is in the current release v0.6.18.post1. Library-side fix (C3.4)
**Early stop:** no

### Security
| Q | Answer | Evidence |
|---|--------|----------|
| S2 template overwritten | no | Template structure present; only the optional Reviewer Notes heading is absent |
| S3 template obligations | met | 5/5 checked, and the Tests section substantiates them rather than asserting them |

### Packaging
| Q | Answer | Evidence |
|---|--------|----------|
| C1.2 public API changes | yes — no signature change; **narrowing** of accepted input | `csrc/batch_prefill.cu` restores a `TVM_FFI_ICHECK_EQ(paged_k_cache.stride(i), paged_v_cache.stride(i))` loop, gated on `HEAD_DIM_QK == HEAD_DIM_VO && HEAD_DIM_VO < 512`. Callers who have been passing differing K/V strides on symmetric sub-512 head dims since #3684 will now hard-error. The description frames this as restoring a check #3684 dropped and that decode never lost — but for those configurations it is a capability removal, traded for the register saving |

### Presentation
| Q | Answer | Evidence |
|---|--------|----------|
| C2.1 perf claim backed by data | yes | B300 (SM103), CUDA 13.0 named; a four-row table walking the bisect (`8f9ad200^` 0.801/0.804 ms → `8f9ad200` 0.859 → `main` 0.860/0.860 → patched 0.787/0.797/0.803), multiple runs per build, and NVBug 6634590's own repro command reproduced in full. Two caveats are volunteered rather than buried: `persistent.cuh` received the identical transformation but was **not separately benchmarked**, and CI's B300 reports a larger step (25.8% decode) than this engineering-sample board |

### Implementation
| Q | Answer | Evidence |
|---|--------|----------|
| C3.2 shared/durable areas touched | yes — **durable** | `include/flashinfer/attention/prefill.cuh` (42+/10−) and `persistent.cuh` (38+/16−) add `SEPARATE_KV_OFFSETS = (HEAD_DIM_QK != HEAD_DIM_VO) \|\| (HEAD_DIM_VO >= 512)` — a local `constexpr` in the first, a `static constexpr` KTraits member in the second — and size `thr_local_kv_offset_v` as `SEPARATE_KV_OFFSETS ? NUM_PAGED_KV_OFFSETS : 1` with every use `if constexpr`-gated. It is a compile-time decision about which configurations get independent K/V addressing, in the shared attention path |
| C3.3 tests match behavior change | yes | No new test, and the reasoning is given rather than asserted: `test_batch_prefill_paged_shared_kv_smem_unequal_kv_strides` (from #3684) pins the large-head unequal-stride behaviour the change must preserve, and reportedly **caught an earlier, too-broad version of this patch** — which is direct evidence the guard is load-bearing. Validation on B300: 5442 passed / 1140 skipped / 5760 xfailed on paged prefill, 2972 passed on decode, 4 passed on `unequal_kv_strides`. Named gap: the **new host-side rejection path** has no test of its own, so nothing pins the error a caller now receives for differing strides at symmetric sub-512 head dims |

### Experimental track
| Q | Answer | Evidence |
|---|--------|----------|

**Notes for the maintainer**
- **The capability removal is the decision to ratify, not the register saving.** Since #3684, symmetric sub-512 configurations have been able to pass independent K/V strides and have them work, because the kernel carried separate offset arrays. This PR makes that a hard error. The register win is well-evidenced; what a reviewer should confirm is that no downstream caller depends on the gap, since the failure mode there is a hard rejection at call time rather than a slow path.
- **The release-blocker flag is qualified.** #3684 is in the current release, and the fix restores the pre-regression baseline on the bug's own repro. Weighing against: this is a *latency* regression rather than a crash, it is tracked in NVBugs rather than by the community, and the description states no GitHub issue exists for either bug — so there is no public signal of user impact to weigh alongside the internal one.
- **`persistent.cuh` is the untested half.** The description is explicit that it received the identical transformation "for consistency" without separate benchmarking. Two headers changed the same way, one measured — worth deciding whether that is acceptable or whether the persistent path deserves its own number before merge.

aleozlx:
@flashinfer-bot run tests/attention/test_batch_prefill_kernels.py tests/attention/test_batch_decode_kernels.py

aleozlx:
@flashinfer-bot run

qsang-nv:
I went through the correctness side first: the host check's condition is the exact complement of `SEPARATE_KV_OFFSETS`, all four V-offset sites are converted, and the other two consumers of `BatchPrefillWithPagedKVCacheDevice` (`pod.cuh:167`, `batch_pod.cuh:138/156`) already enforce per-dim K/V stride equality in `csrc/pod.cu` / `csrc/batch_pod.cu` and build `paged_kv_t` with the shared-stride constructor, so the compile-time gate holds for every caller. Two requests before merge:

**1. The narrowed stride contract deserves an explicit note and a rejection test.** Symmetric `head_dim < 512` with independent K/V strides has been accepted on `main` since #3684, and this restores a hard `ICHECK` for it. No test in the repo covers that combination, and `BatchDecodeWithPagedKVCacheRun` has always rejected it, so this aligns prefill with decode rather than inventing a new restriction — but it is still a user-visible behavior change and worth stating in the PR description / release notes.

More importantly, the kernel's V addressing is now correct *because* that host check is there: with `SEPARATE_KV_OFFSETS == false` the V producer reads `thr_local_kv_offset_k`, so if the check is later removed or bypassed the failure mode is silently misaddressed V rather than a loud error. A `head_dim=128` negative test — two fp16 pools with different padding, `pytest.raises` on the new ICHECK — would pin that, mirroring `test_batch_decode_rejects_unequal_kv_strides_nvfp4_contract` on the decode side.

**2. NVBug 6634592 has no before/after number.** The PR is marked as fixing both bugs, and `head_dim 64/64` takes the same `SEPARATE_KV_OFFSETS == false` path, but the validation table only covers `fa2_tc` decode. A prefill row measured with 6634592's own repro command would close that out.

saltyminty:
This PR is effectively dropping support for unequal K/V strides for almost all cases (`head_dim <=512`). Since https://github.com/flashinfer-ai/flashinfer/pull/3684 went out in 0.6.18, this would be a support surface change.

I have a prior PR https://github.com/flashinfer-ai/flashinfer/pull/4736 which picks between the unequal or identical strides as needed, but lazily compiles the unequal strides module (otherwise the jit-cache size would be too large).
We can discuss which approach is better here.

aleozlx:
Closing in favour of #4736, which predates this by two weeks, targets the same two NVBugs with the same root cause, and is authored by the attention code owner.

The substantive difference: #4736 selects a `SAME_KV_STRIDES` specialization from a host-side stride comparison, so it keeps unequal K/V strides working for symmetric sub-512 head dims — a capability #3684 shipped in v0.6.18. This PR instead narrowed the contract and restored a host-side `ICHECK` for that case, which @saltyminty correctly identified as a support-surface change. Our kernel-side edits turned out to be equivalent; theirs has the better host-side story, and the JIT-cache concern that was open when they first posted is now handled inside that PR by splitting paged prefill into its own module.

I have carried the useful parts over to #4736: the B300 before/after table keyed to NVBug 6634590's repro command, the observation that `persistent.cuh` still has the same doubled array and is out of scope there, and the warning that `csrc/batch_attention.cu` never compares K/V strides — so extending the specialization to the persistent kernel needs a host-side check added first, which is a silent-wrong-output trap I hit here.

Thanks @qsang-nv and @saltyminty for the reviews — both pointed at real things.

Issue #5009, opened 2026-09-07
[Bug] routing_replay_out out-of-bounds writes: fused-shared stride and missing dim0 lower bound


feih-nv:
Two out-of-bounds writes on `routing_replay_out` in the flat `trtllm_*_moe` APIs. Both only touch that optional buffer (`if (mPtrRoutingReplayOut != nullptr)`). Default `None` / unified `MoELayer` and MoE numerics are unaffected.

`routing_replay_out` is `[num_tokens, top_k]` `int16`: the routed top-k expert ids from FromLogits routing. Fused shared slots are not stored there.

## 1. Fused shared experts: replay indexed as if it were packed `[T, K+S]`

**Where:** `routingMainKernel` in `csrc/fused_moe/trtllm_backend/trtllm_fused_moe_routing_deepseek.cu`.

DeepSeek packed ids/weights are `[T, K+S]`: each token has `K` routed slots then `S` fused-shared slots (`mTotalExpertsPerToken = top_k + num_fused_shared_experts`). Replay is a different tensor, `[T, K]` — routed ids only.

The replay store reused the packed index:

```cpp
auto idxTopK = blockIdx.x * params.mTotalExpertsPerToken + laneIdx;  // token * (K+S) + k
params.mPtrRoutingReplayOut[idxTopK] = static_cast<int16_t>(expertIdx);
```

That is correct for packed ids. For replay it is `S` extra elements per token. Token 0 lands in the right place; token 1 starts at element `K+S` instead of `K`, overlapping token 0's row; later tokens walk off the end of the `[T, K]` allocation.

`S == 0` → `K+S == K`, so the bug is invisible. Only DeepSeek sets `mNumFusedSharedExperts`; custom/llama4 already index replay with `* mTopK`.

Host and C++ rejected `routing_replay_out` with `S > 0` (#4239), so this OOB was unreachable and the combination was unavailable.

**Fix:** index replay with `blockIdx.x * mTopK + laneIdx`. Leave packed/weights on `K+S`. Drop the `S > 0` rejects. Replay stays `[T, K]`; passing `[T, K+S]` should still fail `dim1 == top_k`.

**PR:** #4894

## 2. Undersized dim0 accepted, written past the end

**Where:** `validate_routing_replay_out` / `_validate_routing_replay_out`.

[...]

PR #4894, opened 2026-09-02
fix(moe): write fused-shared routing replay as routed-only [T, K]

feih-nv:
## 📌 Description

Fixes #5009 part 1. DeepSeek `routingMainKernel` wrote `routing_replay_out` with the packed index `token * (K+S) + k`. Replay is `[T, K]` (routed ids only), so token 1 starts at `K+S` instead of `K`. `S == 0` hid it; host/C++ rejected `S > 0` (#4239), so the combo was unavailable.

This PR indexes replay with `token * K + k` and drops the `S > 0` rejects. Packed ids/weights stay `[T, K+S]`.

## 🔍 Related Issues

- https://github.com/flashinfer-ai/flashinfer/issues/5009
- https://github.com/flashinfer-ai/flashinfer/pull/5072
- https://github.com/flashinfer-ai/flashinfer/pull/4239

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

## 🧪 Tests

- [x] Tests have been added or updated as needed.
- [x] All tests are passing (`unittest`, etc.).

### CPU reject
- `dim1 == K+S`
- non-DeepSeek + `S > 0`

### GPU (SM100) accept
- FP8 / FP4 fused-shared replay, `T ∈ {8,32}`, `S ∈ {1,2}`
- one FP8 CUDA-graph capture on oversized `[32, K]`

## Reviewer Notes

With `S > 0`, `routing_replay_out` is still `[T, K]` `int16`: the same `K` routed expert ids as `S = 0` on the same logits, ids in `[0, E)`. The `S` fused-shared slots live only in the internal packed ids/weights (`[T, K+S]`, ids `E … E+S-1`). They are not appended to replay. A `[T, K+S]` replay buffer is rejected (`dim1` must equal `top_k`).

## Summary by CodeRabbit

* **New Features**
  * Routing replay is now supported with fused shared experts for FP8 and FP4 MoE operations.
  * Replay buffers retain a routed-only layout with width equal to `top_k`; shared-expert slots are not recorded.
  * Support includes oversized buffers and CUDA graph execution.

* **Documentation**
  * Updated integration and testing guidance for routed-only replay behavior.

* **Tests**
  * Added coverage for replay contents, layouts, sentinel rows, and FP8/FP4 execution paths.

PR #5072, opened 2026-09-09
fix(moe): bound routing_replay_out dim0 from below in both validators

aleozlx:
## 📌 Description

The trtllm routing kernels write one replay row per token unconditionally — `routingDeepSeek`
launches `numBlocks == num_tokens` and writes row `blockIdx.x`, and the custom/llama4 kernels write
row `tokenIdx` — so the kernel touches rows `[0, num_tokens)` regardless of the buffer's actual
`dim0`.

Neither validator checked that lower bound. The C++ one receives `hidden_states` but only compared
`device_id`; the Python one never saw `num_tokens` at all. A caller that passes a shorter buffer
(e.g. `num_tokens=1024` against a `[8, 2]` replay tensor) is accepted by both, and the kernel then
writes past the end of the allocation. Under the caching allocator that lands silently in a
neighbouring tensor; `compute-sanitizer` only sees it with `PYTORCH_NO_CUDA_MEMORY_CACHING=1`.

This adds `dim0 >= num_tokens` to both validators. Oversized buffers stay legal, which is what
CUDA-graph capture at a fixed maximum batch size actually needs — the original "dim0 is
intentionally NOT checked" comment conflated the two directions.

## 🔍 Related Issues

Addresses part 2 of #5009.

Part 1 of that issue — the fused-shared-experts replay stride — is deliberately **not** in this PR.
It removes host-side rejections in order to enable a currently unreachable feature combination
(`routing_replay_out` together with `num_fused_shared_experts > 0`), which is a larger change with a
different review surface. The OOB it describes is unreachable today precisely because those
rejections exist.

### ✅ Pre-commit Checks

## 🧪 Tests

Validated on B300 (SM103), CUDA 13.0, against this branch:

```
tests/moe/test_trtllm_gen_fused_moe.py tests/moe/test_trtllm_gen_routed_fused_moe.py
  -k 'replay or shared_expert'          46 passed, 4459 deselected
tests/model_optimizations/test_dsv3_fused_routing.py -k replay
                                        84 passed, 60 skipped, 4681 deselected
```

Existing callers were checked against the new bound before it was added, since a new rejection is
exactly the kind of change that breaks a test quietly:

- `test_dsv3_fused_routing.py` and `test_trtllm_gen_fused_moe.py` allocate `(num_tokens, top_k)`,
  so the bound holds exactly.
- `test_trtllm_gen_routed_fused_moe.py` uses `replay_capacity = num_tokens + 5`, deliberately
  oversized — which this change keeps legal.
- `test_trtllm_gen_fused_moe.py:2461` passes `torch.empty((1, 1))` and asserts
  `match="routing_replay_out is not supported"`. A `(1, 1)` buffer would also fail the new bound,
  so the raised message could have changed; it does not, because the
  `num_fused_shared_experts > 0` rejection sits first in `_validate_routing_replay_out` and the new
  check is second-to-last.

**No new test is added.** #5009 suggests FP8/FP4 host-side rejection tests for the new bound; happy
to add them here if a reviewer prefers that over a follow-up.

AI-assisted (Claude Opus 5).

## Summary by CodeRabbit

* **Bug Fixes**
  * Routing replay buffers are now validated to ensure they contain at least one row per input token.
  * Undersized buffers now produce a clear error showing the required minimum and received size, preventing potential out-of-bounds writes.
  * Oversized buffers remain supported for CUDA graph pre-allocation.
  * Validation is consistently applied across supported Mixture-of-Experts operations.

### DIFF
--- a/csrc/trtllm_fused_moe_kernel_launcher.cu
+++ b/csrc/trtllm_fused_moe_kernel_launcher.cu
@@ -949,8 +949,12 @@ void cast_fp32_to_bf16(void* output, void const* input, int64_t num_elements, cu
 }  // namespace
 
 // Validate routing_replay_out tensor properties.
-// NOTE: dim0 >= num_tokens is intentionally NOT checked — with CUDA graphs the buffer
-// is pre-allocated at maximum batch size and reused across steps with varying num_tokens.
+// NOTE: dim0 is only bounded from below. The routing kernels write one replay row per
+// token unconditionally (DeepSeek launches numBlocks == num_tokens and writes row
+// blockIdx.x; the custom and llama4 kernels write row tokenIdx), so a buffer with fewer
+// rows than tokens is written past its end. Oversized buffers stay legal: with CUDA
+// graphs the buffer is pre-allocated at maximum batch size and reused across steps with
+// varying num_tokens.
 static void validate_routing_replay_out(TensorView const& replay, TensorView const& hidden_states,
                                         int64_t top_k) {
   TVM_FFI_ICHECK(replay.device().device_type == kDLCUDA)
@@ -959,6 +963,9 @@ static void validate_routing_replay_out(TensorView const& replay, TensorView con
       << "routing_replay_out must be on the same device as hidden_states";
   TVM_FFI_ICHECK(replay.ndim() == 2) << "routing_replay_out must be 2D [num_tokens, top_k]";
   TVM_FFI_ICHECK(replay.size(1) == top_k) << "routing_replay_out dim1 must equal top_k";
+  TVM_FFI_ICHECK(replay.size(0) >= hidden_states.size(0))
+      << "routing_replay_out dim0 must be >= num_tokens (" << hidden_states.size(0) << "), got "
+      << replay.size(0) << "; the routing kernel writes one replay row per token";
   TVM_FFI_ICHECK((replay.dtype() == DLDataType{kDLInt, 16, 1}))
       << "routing_replay_out must be int16 dtype";
   TVM_FFI_ICHECK(replay.IsContiguous())
--- a/flashinfer/fused_moe/core.py
+++ b/flashinfer/fused_moe/core.py
@@ -5447,8 +5447,14 @@ def _validate_routing_replay_out(
     routing_replay_out: Optional[torch.Tensor],
     top_k: int,
     num_fused_shared_experts: int = 0,
+    num_tokens: Optional[int] = None,
 ) -> None:
-    """Validate routing_replay_out tensor properties before passing to C++ kernels."""
+    """Validate routing_replay_out tensor properties before passing to C++ kernels.
+
+    ``num_tokens`` bounds dim0 from below: the routing kernels write one replay row per
+    token unconditionally, so a shorter buffer is written past its end. Oversized buffers
+    stay legal for CUDA-graph capture at a fixed maximum batch size.
+    """
     if routing_replay_out is None:
         return
     if num_fused_shared_experts > 0:
@@ -5467,6 +5473,12 @@ def _validate_routing_replay_out(
     if routing_replay_out.shape[1] != top_k:
         raise ValueError(
             f"routing_replay_out dim1 must equal top_k={top_k}, got {routing_replay_out.shape[1]}"
+        )
+    if num_tokens is not None and routing_replay_out.shape[0] < num_tokens:
+        raise ValueError(
+            f"routing_replay_out dim0 must be >= num_tokens={num_tokens}, "
+            f"got {routing_replay_out.shape[0]}; the routing kernel writes one replay "
+            "row per token"
         )
     if not routing_replay_out.is_contiguous():
         raise ValueError("routing_replay_out must be contiguous (packed row-major)")
@@ -5650,7 +5662,9 @@ def trtllm_bf16_moe(
         scalar return; will become ``[output]`` in v0.8.0).  Otherwise returns
         ``[gemm2_output, expert_weights, expanded_idx_to_permuted_idx]``.
     """
-    _validate_routing_replay_out(routing_replay_out, top_k)
+    _validate_routing_replay_out(
+        routing_replay_out, top_k, num_tokens=hidden_states.shape[0]
+    )
     _validate_bf16_gemm1_activation_params(
         activation_type,
         gemm1_alpha,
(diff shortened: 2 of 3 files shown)


aleozlx:
@flashinfer-bot run tests/moe/test_trtllm_gen_fused_moe.py tests/moe/test_trtllm_gen_routed_fused_moe.py tests/model_optimizations/test_dsv3_fused_routing.py

aleozlx:
/bot run tests/moe/test_trtllm_gen_fused_moe.py tests/moe/test_trtllm_gen_routed_fused_moe.py tests/model_optimizations/test_dsv3_fused_routing.py

aleozlx:
## PR Review Screening

**CI verdict:** ✅ auto-run ok
**Review category:** `human` (rule fired: C1.2 — a new lower bound on `routing_replay_out` dim0 narrows accepted input on seven public `trtllm_*_moe` entry points)
**Blocking checks:** S3 (partial — the two Tests checkboxes are unchecked)
**Release blocker:** 🚨 candidate — fixes an out-of-bounds device write: the routing kernels write one replay row per token unconditionally, so a `routing_replay_out` shorter than `num_tokens` is written past the end of its allocation. Under the caching allocator that lands silently in a neighbouring tensor. Library-side fix in both validators (C3.4)
**Early stop:** no

### Security
| Q | Answer | Evidence |
|---|--------|----------|
| S2 template overwritten | no | Template structure present; only the optional Reviewer Notes heading is absent |
| S3 template obligations | ❗ partial | 3/5 checked. "Tests have been added or updated" and "All tests are passing" are both left **unchecked**, matching the body's statement that this is not yet exercised on GPU — an accurate checklist rather than a false one, but the obligation is genuinely open |

### Packaging
| Q | Answer | Evidence |
|---|--------|----------|
| C1.2 public API changes | yes — no signature change on the public entry points; **narrowing** of accepted input | `csrc/trtllm_fused_moe_kernel_launcher.cu` adds `TVM_FFI_ICHECK(replay.size(0) >= hidden_states.size(0))`; `_validate_routing_replay_out` gains `num_tokens: Optional[int] = None` and raises when `shape[0] < num_tokens`. Buffers that were accepted and silently corrupted memory are now rejected at the call. Oversized buffers stay legal, which preserves the CUDA-graph fixed-max-batch pattern the old comment was protecting |

### Presentation
| Q | Answer | Evidence |
|---|--------|----------|
| C2.1 perf claim backed by data | n-a | Correctness fix; no perf claim |

### Implementation
| Q | Answer | Evidence |
|---|--------|----------|
| C3.2 shared/durable areas touched | no | `csrc/trtllm_fused_moe_kernel_launcher.cu` is a shared launcher, but the change is argument validation only — no kernel logic, dispatch, or abstraction is altered |
| C3.3 tests match behavior change | ❗ no — *executes but does not cover* | No test is added, and the description states the change is **not yet exercised on GPU**. The reasoning offered is that both changes are host-side validation, so a mistake false-rejects rather than corrupts — which is fair as far as it goes, but it means nothing pins either new rejection, and nothing demonstrates the OOB that motivated the fix. #5009 reportedly suggests FP8/FP4 host-side rejection tests, which this PR does not add |

### Experimental track
| Q | Answer | Evidence |
|---|--------|----------|

**Notes for the maintainer**
- **All seven call sites were updated — I checked, because the default makes a miss silent.** `num_tokens` is `Optional[int] = None`, so a call site that forgot to pass it would skip the new bound with no error. At the PR head there are 8 occurrences of `_validate_routing_replay_out`: one definition and seven call sites, and every one passes `num_tokens=hidden_states.shape[0]`. That is complete today; the parameter's default means any future entry point silently opts out, which is worth a thought about making it required.
- **The old comment was not merely absent, it was wrong**, and the PR corrects it in place: `dim0 >= num_tokens is intentionally NOT checked` conflated the two bound directions. That is the kind of confidently-stated invariant that suppresses future scrutiny, so replacing the text matters roughly as much as adding the check.
- **The scoping decision is stated and looks right.** Part 1 of #5009 — the fused-shared-experts replay stride — is deliberately excluded because it would remove host-side rejections to enable a currently unreachable combination, and the OOB it describes is unreachable precisely because those rejections exist. Splitting a reachable memory bug from an unreachable one is the correct order; the open question is only whether part 1 gets its own tracked follow-up.

aleozlx:
@flashinfer-bot run

feih-nv:
**Test:** I have a host-side rejection test for exactly this bound in `b81f0946` on #4894 — `test_routing_replay_out_rejects_undersized_dim0`, FP8 and FP4, CPU tensors, no GPU needed:
https://github.com/feih-nv/flashinfer/blob/b81f0946/tests/moe/test_trtllm_gen_fused_moe.py#L2495

Feel free to lift it. Since you're the one landing the bound, it makes sense for the test to land here rather than in #4894; I'll take `b81f0946` out of #4894 once this is in so the two don't collide on `_validate_routing_replay_out` and the seven call sites.

**Signature:** one small thing to consider — `num_tokens: Optional[int] = None` with the `is not None` guard means a future call site that forgets to pass it silently loses the check. Making it a required `num_tokens: int` turns that into a `TypeError` at the call site. All seven callers already pass it, so this is only about the next one.

Validation looks thorough — thanks for running `test_dsv3_fused_routing.py`; I hadn't.

aleozlx:
@flashinfer-bot run

aleozlx:
Thanks — took both.

**Signature:** `num_tokens` is now a required `int` rather than `Optional[int] = None`, so forgetting it is a `TypeError` at the call site instead of a silently skipped bound. It moved ahead of the defaulted `num_fused_shared_experts`, which means the two call sites that pass both (`trtllm_fp8_block_scale_moe`, `trtllm_fp4_block_scale_moe`) now name each keyword explicitly — otherwise `nfse`/`nsfe` would have bound positionally to `num_tokens`. All seven callers checked.

**Test:** lifted `test_routing_replay_out_rejects_undersized_dim0` from `b81f0946` verbatim, credited to you via `Co-Authored-By`. Its `match=r"dim0 must be >= num_tokens"` lines up with the message the Python validator raises. Please do drop that commit from #4894 once this lands.

Both in `5a26fb64`. CI is running now; the earlier B300 numbers in the description were from the previous revision, and I'll refresh them once this run reports.

One thing worth your eye since you filed the issue: this PR is still only part 2. Part 1 — the fused-shared-experts replay stride — is deliberately out, because it removes host-side rejections to enable a combination that is currently unreachable, and the OOB you describe there is unreachable precisely because those rejections exist. If you would rather that land as part of this change than as a follow-up, say so and I will fold it in.

feih-nv:
Keep Part 1 in #4894 — it's the todo item that PR exists for, and the two land cleanly in sequence: this one first, then #4894 rebases and drops `b81f0946`. Thanks for taking the test and the signature change.

feih-nv (approved):
LGTM

PR #4908, opened 2026-09-03
[feat] Add warp level split k bf16 gemm

jiahanc:
## 📌 Description

Three changes on the `cute-dsl` BF16 GEMM path for decode-sized M on SM100/SM103, targeting the low-M shapes vLLM routes to `dsv3_fused_a_gemm` (vllm-project/vllm#54524), where the existing direct / cluster split-K kernels were behind fused-A on most cells.

### 1. Warp split-K kernel (`flashinfer/gemm/kernels/dense_bf16_gemm_warp_splitk.py`)

- The four compute warps of a CTA each own one K quarter of every tile and reduce through a shared-memory mailbox; the kernel is the first `cute-dsl` runner whenever it is eligible (M <= 32, N % 16 == 0, K % 128 == 0, <= 64 K tiles).
- 2 weight-loader warps stream the [N, K] weight through a `cp.async` ring with an `L2::evict_first` hint (the activation rows every CTA re-reads stay L2-resident); 1-2 activation-loader warps; 4 compute warps (`mma.sync` m16n8k16 BF16 -> FP32), each on its own K quarter of every stage; the first compute warp reduces the four partials and stores BF16 (plus an optional bias, added in FP32).
- The first weight tile is issued before the mbarrier init/CTA barrier, and the consumer K loop is fully unrolled (the K-tile count is static) so every `ldmatrix` is `[R_lane + imm]` — a rolled loop kept the stage index in a vector register and cost 3-16% on K=7168.
- Tactic space: `(output_tile 16|32, token_tile 8|16|32, k_tile 128|256, stages <= 16, b_loader_warps 1|2)`, bounded to 64 K tiles; `default_tactic` picks by wave count and CTA residency (counting the 1 KB per-CTA reservation).
- Two ptxas 13.0 workarounds are documented in the source: the pre-barrier tile stays unhinted and single-stage rings are excluded when K spans several tiles, because the hinted `LDGSTS` is otherwise mis-encoded (`[R+UR], desc[UR<odd>]`, illegal instruction).

### 2. `cute-dsl` runner selection refactor (`flashinfer/gemm/gemm_base.py`)

- One base class, `_CuteDSLBf16Runner`, for the backend's runners: shared cache-key extras plus `supports_inputs` / `is_tactic_compatible`, so no runner is special-cased by name.
[...]

PR #5089, opened 2026-09-10
fix: guard BF16 warp Split-K availability and prune tactics

jiahanc:
## 📌 Description

Guard BF16 warp Split-K runner creation with the existing `is_cute_dsl_experimental_available()` probe. CuTe DSL 4.6.2 lacks `cutlass.experimental`; skip only this runner before importing its kernel, keeping Direct, cluster Split-K, and the cuBLASLt fallback available.

Also include the retained BF16 warp Split-K tactic-pruning heuristic:
- For multi-tile K, prune larger token tiles when a smaller supported tile produces the same CTA grid. Keep single-K-tile alternatives, which can still win, and preserve the default tactic.
- Exclude `k_tile=256, k_tile_count=2` from candidate generation and explicit validation: this schedule can produce an invalid hinted `LDGSTS`.
- Bump the BF16 CuTe DSL autotune cache version from 11 to 12 so earlier selections are not reused.

The GPU mainloop, JIT/launch implementation, default-selection heuristic, dependencies, and test files are unchanged. Document the warp runner's CuTe DSL >= 4.7 requirement without raising the package dependency floor.

## 🔍 Related Issues

Follow-up to #4908.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

All hooks passed for both changed files: `pre-commit run --files flashinfer/gemm/gemm_base.py flashinfer/gemm/kernels/dense_bf16_gemm_warp_splitk.py`. Repository-wide hooks were not run.

## 🧪 Tests

No test files added or modified. Reused the existing accuracy tests on NVIDIA B200 (SM100), CUDA 13.0, PyTorch 2.13.0+cu130:

```bash
python -m pytest -q tests/gemm/test_mm_bf16.py -k 'cute and not cutile'
```

- CuTe DSL 4.6.2: 10 passed; verified the warp kernel module was never imported.
- CuTe DSL 4.7.0: 10 passed; verified the warp kernel remained available.

Additionally verified the public K=512, N=256 path: autotune at M=64 with bias and PDL, then run M=8/25/33 and compare with an FP32 reference. No test files were added for these validation runs.

Validation is scoped to these existing BF16 CuTe DSL cases and the stated runtime checks, not the full repository suite.

### Autotuning cost

Before = `31a6926b` (availability guard only); after = this heuristic update.

| Metric | Before | After |
|---|---:|---:|
| Warp tactic evaluations across the full-table harness's 77 tuning profiles | 9,714 | 5,834 |
| Existing B200 accuracy/autotune suite wall time, one run per version | 247.26 s | 124.38 s |

Candidate counts follow the 210-cell, 16-family workload from vllm-project/vllm#54524: tune once at M=16 per family, or M=8 when the family's largest M is 8. The 39.9% reduction measures the search space, not kernel latency.

The wall-time rows use the exact pytest command and B200/CUDA/PyTorch/CuTe DSL 4.7.0 environment above, including JIT compilation, autotuning, setup, and correctness checks. They are single-run observations, not a steady-state kernel benchmark or a general speedup guarantee.

## 🔬 Experimental Track

```experimental-tests
#
#
```

## Reviewer Notes

Missing experimental support silently excludes the warp runner from both default selection and autotuning; it does not disable the whole `cute-dsl` backend or suppress unrelated import failures.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compatibility with older CuTe DSL versions by omitting unsupported experimental warp Split-K runners.
  - Prevented unavailable runners from being included in BF16 matrix multiplication execution paths.
  - Prevented selection of an invalid BF16 GEMM schedule that could produce incorrect GPU instructions.
  - Improved autotuning by excluding redundant token-tile options when they provide no additional grid coverage.
- **Documentation**
  - Clarified CuTe DSL version requirements and available BF16 runner behavior.

### DIFF
--- a/flashinfer/gemm/gemm_base.py
+++ b/flashinfer/gemm/gemm_base.py
@@ -735,11 +735,11 @@ def mm_bf16(
         explicitly. Without autotuning, M > 32 runs cuBLASLt; below, the
         direct kernel runs where its shape heuristic applies, otherwise the
         warp Split-K kernel whenever it is eligible (N % 16 == 0, K % 128 == 0
-        with at most 64 K tiles), and cluster Split-K otherwise. With
-        autotuning, one call profiles the low-M kernels on the M <= 32
-        buckets and the cuBLASLt fallback on the larger ones, so a single
-        large-M warm-up tunes both ranges; with bias the direct kernel is
-        excluded.
+        with at most 64 K tiles; requires CuTe DSL >= 4.7), and cluster Split-K
+        otherwise. With autotuning, one call profiles the available low-M
+        kernels on the M <= 32 buckets and the cuBLASLt fallback on the larger
+        ones, so a single large-M warm-up tunes both ranges; with bias the
+        direct kernel is excluded.
         ``"auto"`` allows selecting the best tactic from all available backends when autotune is enabled.
 
     Returns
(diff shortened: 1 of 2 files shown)


StudyingShao (approved):
LGTM.

PR #4514, opened 2026-08-14
feat(jit-cache): split cache into architecture provider wheels

dierksen:
## 📌 Description

This PR turns `flashinfer-jit-cache` into a small top-level shim backed by independently installable, architecture-specific provider wheels. It retains legacy monolithic cache discovery as a fallback while adding:

- provider package metadata and entry-point discovery
- exact CUDA architecture matching, including suffixed targets such as `sm90a`, `sm120f`, and `sm121a`
- a default installation mode whose shim dependencies name the complete published provider set
- a minimal installation mode that selects only an explicitly requested or locally detected provider, with no implicit SM80 baseline
- provider wheelhouse build, binary inventory, installation, and JIT-disabled GPU smoke tooling
- AOT capability gating so unsupported modules, currently BGMV MoE on SM80, are omitted from that provider rather than failing its build

The provider matrix is intentionally independent of the size-pruned monolithic wheel matrix. Architecture-specific wheels may therefore preserve useful native targets such as SM121a without adding them back to every monolithic wheel. Native providers are SASS-only; the shim does not rely on PTX or an SM80 compatibility baseline.

## 🔍 Related Issues

- Follow-up design to #3265
- Incorporates current-main behavior from #4469, #4527, #4682, #4711, #4757, and #4760

### Provider canaries

Two end-to-end CUDA 13.0 canaries have exercised provider build, strict binary inspection, shim/provider installation, provider discovery, top-level import without CUTLASS DSL, and a JIT-disabled `silu_and_mul` numerical smoke:

| System | Target | Provider result |
| --- | --- | --- |
| x86_64 A100 test system | A100 / SM80 | 113.8 MiB; every CUDA-bearing module SM80-only; zero PTX |
| DGX Spark | aarch64 GB10 / SM121a | 173.0 MiB; 558 modules; every CUDA-bearing module SM121a-only; zero PTX |

The x86_64 A100 artifact was built before the final BGMV capability correction and contained 207 modules. Final-branch source-level SM80 AOT enumeration produces 206 modules with `bgmv_moe` absent, and the associated focused suite passed 30 tests. A final-head SM80 provider wheel still needs to be rebuilt to confirm that exact packaged inventory.

The Spark provider could load its packaged BGMV module, but a BGMV numerical invocation exceeded the device's dynamic shared-memory limit (approximately 216 KiB requested versus approximately 101 KiB available). That is a pre-existing kernel/runtime limitation and is outside this packaging change; the independent `silu_and_mul` provider smoke passed.

[...]

PR #5096, opened 2026-09-10
ci(jit-cache): enable provider-only wheel releases

dierksen:
## 📌 Description

This PR enables the architecture-specific JIT-cache provider wheels introduced
by #4514 and removes the monolithic JIT-cache wheel mode from release, build,
package, and runtime paths.

- Adds explicit provider inventories per CUDA and CPU platform in
  `ci/cuda-versions.json`.
- Derives the individual provider build matrix and the CUDA/CPU shim matrix
  from that single inventory.
- Builds each binary provider, builds one platform-specific
  `flashinfer-jit-cache` shim per CUDA/CPU pair, and validates the assembled set
  before publication.
- Builds the patched CUDA 13.4 `sccache` binary once per CPU architecture and
  shares it across the corresponding provider builds.
- Validates provider SASS using CUDA binary-compatibility rules while continuing
  to reject PTX and targets that cannot execute on the declared architecture.
- Propagates the selected architecture flags through NVCC device linking so
  linked modules do not acquire the toolkit's default `sm52` or `sm75` cubin.
- Selects the closest compatible installed provider at runtime, with exact
  targets preferred over family targets and family targets preferred over
  unsuffixed baselines.
- Requires each provider's full CUDA-specific version to match the installed
  shim, preventing a provider such as `+cu134` from being loaded by a `+cu130`
  shim.
- Supports heterogeneous processes by resolving the best provider independently
  for every visible CUDA architecture, loading all selected copies of a module,
  and dispatching each call according to its CUDA tensor devices.
- Keeps partial provider inventories usable: a module present only in the SM103
  provider can run on SM103 while an uncovered SM120 call falls back to the
  normal multi-architecture JIT module.
- Registers the cubin callback for every selected provider library instead of
  only the primary path.
- Makes minimal installation discover the provider inventory from the exact
  published shim wheel instead of maintaining a second architecture table.
- Drops separate SM86 provider builds because the unsuffixed SM80 provider is
  forward compatible within compute-capability major 8.
- Teaches wheel-index generation to recognize provider distribution names.
- Removes monolithic wheel compilation, artifact publication, package metadata,
  runtime discovery, and workflow selection paths.
- Keeps reusable provider validation independent from the disposable one-off
  wheelhouse tooling.

Pull request runs exercise the complete split provider build, shim assembly,
installation, and artifact-validation path without publishing artifacts. Once
merged, scheduled nightlies and release dispatches build and publish only the
split provider wheel set. Rollback requires reverting to a workflow revision
that still builds monolithic wheels; there is no runtime format switch.

## 🔍 Related Issues

- Follow-up to #4514
- Design follow-up to #3265

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

> If you are unsure about how to set up pre-commit, see the [pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

Completed on the current revision:

- commit-time pre-commit checks for all changed files, including Ruff and mypy
- `python3 ci/validate_cuda_versions.py`
- `pytest tests/test_env.py -q --noconftest` (`6 passed`), including
  heterogeneous provider resolution and nested tensor-device discovery
- Python byte-compilation and `git diff --check`

Completed on an earlier revision of this PR:

- `pytest -q --confcutdir=tests/jit tests/jit/test_jit_cache_release.py`
  (`6 passed` in a lightweight local environment)
- provider, release, and CLI suites in a CUDA container on an ARM64 SM121 test
  system (`86 passed`)
- focused NVCC device-link regression test and a direct `cuobjdump --list-elf`
  check of an affected `sm103a` module (only `sm103a` cubins present)

The complete provider build matrix and assembled-wheel installation checks are
exercised by this PR's release dry run. Actual same-process dispatch across an
SM103 and an SM120 GPU remains a hardware validation gap. The updated canary
supports that test as one invocation with
`--provider sm103a --provider sm120f` and JIT disabled.

## 🔬 Experimental Track

```experimental-tests
#
#
```

## Reviewer Notes

Please focus on the provider inventory policy, CUDA binary-compatibility rules,
provider-only workflow fan-in, platform-specific shim tags and dependencies,
artifact assembly, removal of monolithic paths, and the heterogeneous AOT
dispatcher's tensor-device selection and JIT fallback behavior.

## Summary by CodeRabbit

- **New Features**
  - Releases and nightly builds now provide architecture-specific JIT-cache provider wheels alongside a shim wheel.
  - Provider wheels include validation reports and SHA-256 checksums.
  - Installed providers are discovered and validated automatically.
  - CUDA provider architecture coverage has expanded.

- **Breaking Changes**
  - Legacy monolithic JIT-cache wheel builds and format selection are no longer supported.

- **Documentation**
  - Updated JIT-cache documentation to describe provider-based releases and validation.


dierksen:
@flashinfer-bot run

dierksen:
@flashinfer-bot run

dierksen:
@flashinfer-bot run

dierksen:
@flashinfer-bot run

dierksen:
@flashinfer-bot run

dierksen:
@flashinfer-bot run

aleozlx:
## PR Review Screening

**CI verdict:** ⛔ hold
**Review category:** `live`
**Early stop:** yes — screening halted after the Security section; Packaging, Presentation, Implementation and Experimental track were not evaluated.

A maintainer has authorized disclosing the reason on this PR.

### What fired

Check **S1 (injection / supply-chain)**, on new `subprocess` call sites introduced by two files this PR adds:

- `scripts/verify_jit_cache_provider_release.py`
- `scripts/verify_jit_cache_provider_artifact.py`

Before flagging, the screen checked whether these were relocated rather than new: the PR deletes no files, and both paths are added by this diff, so these are new call sites rather than code moved from elsewhere. Other `subprocess` occurrences in the diff were checked against the base commit and cleared as pre-existing or modified-in-place.

### Why it fired, and what it does not mean

S1 adjudication is deliberately **mechanical**. When one of the primitives it names appears on an added line, the check fires on presence alone — regardless of the caller, the surrounding file, or whether it plausibly matters in that instance. The screen is explicitly not permitted to decide that a given occurrence is benign; narrowing what a primitive means is a rule change that a maintainer makes once, for every future PR, rather than something a screen decides on the PR that happens to raise it.

So this is **not** a finding that the code is unsafe, incorrect, or written in bad faith. It is a routing outcome: CI is held and a human adjudicates instead of the screen. Nothing here needs to be "fixed" to clear it — a maintainer decides whether the rule should reach this class of change at all, and that question is open rather than settled against this PR.

The practical consequence is the early stop: once S1 fires the screen stops reading, so the remaining four sections carry no verdict. If a maintainer clears this, re-running the screen produces the full report.

<sub>Generated by flashinfer-pr-screen · AI screening can make mistakes — a maintainer's judgment supersedes this report.</sub>

aleozlx (approved):

aleozlx (approved):

aleozlx:
@flashinfer-bot run

PR #5102, opened 2026-09-10
fix(pod): stop silently ignoring decode-side run() arguments

### LINKED ISSUE #3509: [Code Quality] BatchPrefillWithPagedKVCacheWrapper.run() silently overrides 6 caller-provided decode arguments
## Summary

`BatchPrefillWithPagedKVCacheWrapper.run()` in `flashinfer/pod.py` advertises several per-call decode-side arguments in its signature, but the first thing the function body does on the decode path is overwrite all of them with plan-time state stored on `self`. Caller-provided values are silently discarded, with no warning and no error.

There is also a pre-existing `TODO_AK` comment at the override site, which suggests even the original author was uncertain about the contract.

## Location

`flashinfer/pod.py` — `BatchPrefillWithPagedKVCacheWrapper.run()`

**Signature** (lines ~443–479) declares these decode-side runtime args:

- `kv_layout_d`
- `pos_encoding_mode_d`
- `window_left_d`
- `sm_scale_d`
- `rope_scale_d`
- `rope_theta_d`
- (`logits_soft_cap_d` is also affected — see below)

**Override site** (lines 526–532):

```python
pos_encoding_mode_d = self._pos_encoding_mode
window_left_d       = self._window_left
logits_soft_cap_d   = self._logits_soft_cap
sm_scale_d          = self._sm_scale
rope_scale_d        = self._rope_scale
rope_theta_d        = self._rope_theta
```

(`kv_layout_d` is accepted in the signature but is not consumed at all on the decode path.)

## Impact

- The signature is the API contract. Right now it lies.
- Callers tuning these values per step (e.g. `sm_scale` in speculative decoding or multi-LoRA scenarios, or `window_left` for streaming) see **no effect and no error**.
- Debugging this requires reading the function body and noticing the silent shadowing of locals by `self._x`. Unfriendly.

## Proposed fix — needs maintainer input

Two reasonable options; I'd like maintainers to pick before I send a PR:

1. **Recommended: shrink the signature.** Remove the dead decode-side parameters from `run()` (they belong only to `plan()`). Update the docstring. Smallest, safest change; makes the API honest.

2. **Honor the runtime args.** Actually thread the `run()`-time values into the kernel launch and remove the `self._x` shadowing. Larger change and may collide with assumptions baked into `plan()` / `_cached_module` (e.g. `use_sliding_window=window_left != -1`, `use_logits_soft_cap=logits_soft_cap > 0` are compiled into the cached module at plan time, so some of these parameters genuinely *cannot* be changed per-call without invalidating the plan).

cc: @aleozlx

SamMausberg:
## Description

`PODWithPagedKVCacheWrapper.run()` accepts decode-side arguments (`kv_layout_d`, `pos_encoding_mode_d`, `sm_scale_d`, `window_left_d`, `rope_scale_d`, `rope_theta_d`) and then overwrites them with the values cached by `plan()` (the `TODO_AK` block), so a caller tuning them sees no effect and no error (#3509); `kv_layout_d` is never read at all. Three more decode-side arguments, `custom_mask_d`, `packed_custom_mask_d` and `causal_d`, are documented as functional, but the decode launch hardcodes a non-causal mask mode and passes no custom mask, so they never reach the kernel either.

The decode kernel is selected at plan time (`pos_encoding_mode`, `window_left` and the soft-cap flag pick the compiled module), so these settings cannot change per call. This PR keeps the signature compatible (every parameter stays in its position) and makes the contract explicit instead of silent:

- The six settings become optional (`None` by default). A value passed to `run()` emits a `DeprecationWarning` and must equal the planned value, defaults resolved; otherwise `run()` raises `ValueError`. This is the pattern `BatchDecodeWithPagedKVCacheWrapper.run` already uses for `window_left` and `q_len_per_req`.
- `custom_mask_d`, `packed_custom_mask_d` and `causal_d` raise `NotImplementedError` when set; their defaults pass silently, as before. Wiring a decode-side custom mask through the binding parameters that already exist in `csrc/pod.cu` would be a separate feature.
- The override block and its two `TODO_AK` comments are gone; `run()` resolves the plan-time values once, and `q_scale`/`k_scale` still fold into the decode `sm_scale`.
- Docstrings: `run()` marks the nine arguments as deprecated and states what each does now; `plan()` says its `pos_encoding_mode`, `window_left`, `sm_scale`, `rope_scale` and `rope_theta` are also the decode-side settings of `run()`, and that logits soft-capping is not supported by this wrapper (it is hardcoded off in `plan()`).
- `benchmarks/bench_mixed_attention.py` no longer passes `causal_d=True`, which was silently ignored and would now raise.

Migration: pass the decode-side settings to `plan()` and drop the `*_d` arguments from `run()`. Removing the parameters, the option the issue recommends, is left for a later release so that existing callers get a warning first. `BatchPODWithPagedKVCacheWrapper.run()` never had these arguments and is unchanged.

## Related Issues

Closes #3509.

## Pull Request Checklist

### Pre-commit Checks

## Tests

- New `tests/utils/test_pod_kernels.py::test_pod_with_paged_kv_cache_decode_plan_settings`: plans the POD wrapper with a non-default decode `window_left` (32 over 128-token histories) and `sm_scale` (0.5 / sqrt(head_dim)) and checks the decode output against `BatchDecodeWithPagedKVCacheWrapper` planned with the same values; then checks that passing the same values to `run()` warns and gives the same output, that a different `window_left_d` raises `ValueError`, and that `causal_d=True` raises `NotImplementedError`. The JIT warmup fixture gains the sliding-window decode, prefill and POD variants the test needs.
- Mutation check: making `run()` fall back to `window_left = -1` makes the new test fail (`Decode mismatch`); restored, it passes.

Runs on an NVIDIA GH200 480GB (SM90), CUDA 12.8, torch 2.7.0+cu128:

```
pytest tests/utils/test_pod_kernels.py                                                                   169 passed, 24 skipped, 21 warnings in 476.81s
pytest tests/trace/test_pod_with_paged_kv_cache_run_reference_correctness.py tests/trace/test_batch_pod_run_reference_correctness.py tests/trace/test_fi_trace_template_consistency.py tests/trace/test_template_init.py tests/trace/test_template_registry.py   1229 passed, 192 skipped, 22 warnings in 26.12s
```

## Reviewer Notes

- The first version of this PR removed the nine arguments; the public API check asked for a deprecation path, so this version keeps them with a warning and validation. Removing them later is a small follow-up.
- Pre-existing and unchanged: `return_lse_p` and `return_lse_d` allocate an LSE buffer that `run()` does not return (already documented in the docstring).

Made in combination with Claude Fable 5.1.

## Summary by CodeRabbit

- **Bug Fixes**
  - Decode attention now consistently uses the settings configured during planning, including scaling, positional encoding, layout, and sliding-window behavior.
  - Added validation to prevent conflicting decode-time settings and provide clear warnings for deprecated overrides.
  - Decode operations now explicitly reject custom masks and causal mode, which are unsupported for this path.

- **Documentation**
  - Updated attention API documentation to clarify planning requirements, decode behavior, and unsupported logits soft-capping.

### DIFF
--- a/benchmarks/bench_mixed_attention.py
+++ b/benchmarks/bench_mixed_attention.py
@@ -208,7 +208,6 @@ def run_bench(
                 q_d,
                 kv_d,
                 causal_p=causal,
-                causal_d=causal,
             )
         )
         ms_pod = np.median(measurements)
--- a/flashinfer/pod.py
+++ b/flashinfer/pod.py
@@ -16,6 +16,7 @@ limitations under the License.
 
 import functools
 import math
+import warnings
 from types import SimpleNamespace
 from typing import Any, List, Optional, Tuple, Union
 
@@ -56,6 +57,41 @@ def get_pod_module(*args):
 def get_batch_pod_module(*args):
     module = gen_batch_pod_module(*args).build_and_load()
     return SimpleNamespace(run_tensor=module.batch_pod_with_kv_cache_tensor)
+
+
+def _check_deprecated_decode_run_args(passed: dict, planned: dict) -> None:
+    """Warn for decode-side settings passed to run() and reject values that
+    differ from plan(); the decode kernel is selected at plan() time, so they
+    can only be validated here."""
+    given = {name: value for name, value in passed.items() if value is not None}
+    if not given:
+        return
+    warnings.warn(
+        f"Passing {sorted(given)} to PODWithPagedKVCacheWrapper.run() is "
+        "deprecated; set the decode-side settings in plan() instead. Scheduled "
+        "for removal in a future release.",
+        DeprecationWarning,
+        stacklevel=3,
+    )
+    for name, value in given.items():
+        if value != planned[name]:
+            raise ValueError(
+                f"{name}={value!r} differs from the value planned for the decode "
+                f"side ({planned[name]!r}); decode-side settings are fixed by plan()"
+            )
+
+
+def _check_decode_mask_args(
+    custom_mask_d: Optional[torch.Tensor],
+    packed_custom_mask_d: Optional[torch.Tensor],
+    causal_d: bool,
+) -> None:
+    if custom_mask_d is not None or packed_custom_mask_d is not None or causal_d:
+        raise NotImplementedError(
+            "the decode side of PODWithPagedKVCacheWrapper runs non-causal "
+            "attention without a custom mask; custom_mask_d, packed_custom_mask_d "
+            "and causal_d are not supported"
+        )
 
 
 class PODWithPagedKVCacheWrapper:
@@ -320,8 +356,7 @@ class PODWithPagedKVCacheWrapper:
             The data type of both the query and key/value tensors. Defaults to torch.float16.
             data_type is deprecated, please use q_data_type and kv_data_type instead.
         sm_scale : Optional[float]
-            Softmax scale.  If ``None``, defaults to ``1 / sqrt(head_dim)``.  Cached on
-            the wrapper and reused at :meth:`run` time.
+            Softmax scale.  If ``None``, defaults to ``1 / sqrt(head_dim)``.
         rope_scale : Optional[float]
             Scale factor applied during RoPE interpolation.  Only consulted when
             ``pos_encoding_mode != "NONE"``.  Defaults to ``1.0`` when ``None``.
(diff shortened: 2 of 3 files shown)


SamMausberg:
The positional-binding concern from the first revision is addressed in the updated revision: every parameter stays at its original position. The six settings now default to None, emit a DeprecationWarning when passed and must equal the planned value (ValueError otherwise); custom_mask_d, packed_custom_mask_d and causal_d raise NotImplementedError when set. A legacy positional call therefore binds exactly as before and nothing can shift into q_scale, k_scale or v_scale.

PR #5105, opened 2026-09-10
fix(attention): validate index tensors in the batch attention wrappers

### LINKED ISSUE #1654: Feature Request: `last_page_len` dtype validation
I ran into a somewhat similar issue to the one described in #1634 with the `BatchDecodeWithPagedKVCacheWrapper` attention kernel.
I mistakenly passed an int64 tensor to the plan function's `last_page_len` parameter, when the expected dtype is int32. When I ran the kernel with two sequences, this led to the second sequence returning all 0s, without any error being reported.

In this case, I believe the documentation is missing the dtype requirement, if it is one, altogether, though the example usage does have this (and other tensors) with int32 dtype. An assertion raised would have saved me a few hours of debugging in the wrong direction...

SamMausberg:
## Description

The batch attention wrappers hand their index tensors (`qo_indptr`, `paged_kv_indptr`, `paged_kv_indices`, `paged_kv_last_page_len`, `kv_indptr`) to kernels that read them as the compiled `IdType`, int32 for every module the wrappers build, whatever dtype the tensor actually has. An int64 `paged_kv_last_page_len` with otherwise int32 inputs therefore gave wrong output for every request after the first, with no error, on paged prefill (fa2 and fa3) and on tensor-core decode (#1654); only the CUDA-core decode path rejected it, from C++. A non-contiguous index tensor that already lived on the device was passed through `x.to(device)` unchanged and read with the wrong stride.

This PR validates the index tensors before anything is planned or launched:

- `_check_index_tensor` (`flashinfer/utils.py`): 1-D, `torch.int32`, optional length cross-check, optional device and contiguity checks; every violation is a `ValueError` naming the tensor, the expectation and the actual value.
- `plan()` of `BatchPrefillWithPagedKVCacheWrapper`, `BatchPrefillWithRaggedKVCacheWrapper` and `BatchDecodeWithPagedKVCacheWrapper`, and the copy-free `fast_decode_plan` used by SGLang, validate dtype, rank and matching lengths (`len(indptr) == batch_size + 1`, `len(last_page_len) == batch_size`). CPU inputs stay allowed, as serving frameworks pass them. In the non-graph path the tensors go through `.contiguous()` before the device copy, which is a no-op for contiguous inputs.
- CUDA-graph buffers passed to `__init__` (`qo_indptr_buf`, `paged_kv_indptr_buf`, `paged_kv_indices_buf`, `paged_kv_last_page_len_buf`, `mask_indptr_buf` when given, and the decode `*_buffer` arguments) must be int32, 1-D, contiguous and on the workspace device, checked once at construction because captured kernels keep reading them by address.
- Defense in depth: the plan, ragged run and paged run bindings in `csrc/batch_prefill.cu` and `csrc/batch_prefill_sm90.cu` check their index tensors against the compiled `IdType` (the SM90 plan checks only the width of `kv_len_arr`, whose documented `seq_lens` source may be uint32) through a new `dl_dtype_for<IdType>()` helper in `csrc/tvm_ffi_utils.h` (a module can be built for int32 or int64, so no hard-coded `dl_int32`); the existing checks in `csrc/batch_decode.cu` switch to the same helper. These checks also cover callers that reach the bindings without the wrappers' Python validation (the block-sparse wrappers and `fast_decode_plan`): an index tensor of the wrong dtype there now raises a `RuntimeError` from `run()` instead of being misread.
- Docstrings state `torch.int32` for each index tensor and the buffer requirements.

int32 is the rule because the CUDA-core decode binding, `docs/tutorials/kv_layout.rst`, every docs example and the AOT-built modules use it; the JIT module is keyed by the dtype of the leading `indptr`, which is how a mixed dtype ended up misread. Consistently int64 index tensors, which used to build an int64 module through the JIT, are now rejected in `plan()` as well.

## Related Issues

Closes #1654.

## Pull Request Checklist

### Pre-commit Checks

## Tests

New file `tests/attention/test_index_tensor_validation.py`:

- The issue's scenario (batch 3, int64 `last_page_len`, other tensors int32) raises `ValueError` from `plan()` on paged prefill (fa2, fa3) and decode (CUDA core, tensor core).
- int64 `qo_indptr`, `paged_kv_indptr`, `paged_kv_indices` and ragged `kv_indptr`, wrong lengths, and a 2-D tensor raise.
- A strided device `last_page_len` (paged prefill and tensor-core decode) or ragged `kv_indptr` gives output identical to the contiguous run.
- CUDA-graph buffers with int64 dtype or a strided layout raise in `__init__` (paged prefill, ragged prefill, decode).
- The C++ binding check: after a valid `plan()`, an int64 buffer swapped in behind the wrapper is rejected by `run()` on fa2 and fa3 (the fa3 case skips without SM90a).
- `fast_decode_plan` rejects an int64 `last_page_len` and gives identical output for a strided one.

Mutation check: with the Python `plan()` validation disabled, the regression test fails; restored, it passes.

Runs on an NVIDIA GH200 480GB (SM90), CUDA 12.8, torch 2.7.0+cu128, from an empty JIT cache (no prebuilt or staged kernels, so every binding was compiled from this branch):

```
pytest tests/attention/test_index_tensor_validation.py    27 passed, 21 warnings (107.08s including the kernel builds)
pytest tests/attention/test_tensor_cores_decode.py         2448 passed, 21 warnings in 389.32s
pytest tests/utils/test_jit_example.py                     14 passed, 21 warnings in 143.13s
pytest tests/attention/test_batch_prefill_kernels.py       8788 passed, 1329 skipped, 5760 xfailed, 21 warnings in 561.73s
pytest tests/attention/test_batch_decode_kernels.py        3081 passed, 181 skipped, 21 warnings in 421.58s
```

## Reviewer Notes

- The same index tensors in the MLA, POD, sparse and cascade wrappers, the `workspace_size()` helpers that mirror `plan()` and the cudnn-only `v_indptr`/`o_indptr` of the ragged wrapper are not covered here (follow-ups).
- `seq_lens` and `block_tables` (documented as uint32) are not validated by this PR.
- Two pre-existing ad-hoc length checks (the paged wrapper's buffer lengths in `__init__` and the ragged wrapper's `kv_indptr` length in `plan()`) are replaced by the helper, so their messages change and now name the actual values.

Made in combination with Claude Fable 5.1.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added consistent validation for attention index tensors, including dtype, shape, length, device, and contiguity checks.
  * Prevented mismatched index buffers from reaching decode and prefill operations.
  * Automatically handles non-contiguous index tensors where supported, preserving expected results.
  * Improved validation for CUDA graph buffers and custom mask-indirection inputs.

* **Documentation**
  * Clarified requirements for index tensor types, dimensions, lengths, devices, and contiguity across decode and prefill APIs.

* **Tests**
  * Added coverage for invalid inputs, non-contiguous tensors, CUDA graph buffers, and multiple attention backends.

### DIFF
--- a/csrc/batch_decode.cu
+++ b/csrc/batch_decode.cu
@@ -42,7 +42,7 @@ Array<int64_t> BatchDecodeWithPagedKVCachePlan(
     int64_t num_qo_heads, int64_t num_kv_heads, int64_t page_size, bool enable_cuda_graph,
     int64_t window_left, double logits_soft_cap, int64_t head_dim_qk, int64_t head_dim_vo,
     TensorView empty_q_data, TensorView empty_kv_data) {
-  CHECK_INPUT_TYPE(indptr, dl_int32);
+  CHECK_INPUT_TYPE(indptr, dl_dtype_for<IdType>());
 
   size_t float_workspace_size_in_bytes =
       float_workspace_buffer.size(0) * get_element_size(float_workspace_buffer);
@@ -89,7 +89,7 @@ Array<int64_t> BatchDecodeWithPagedKVCacheWorkspaceSize(
   (void)logits_soft_cap;
   (void)empty_q_data;
   (void)empty_kv_data;
-  CHECK_INPUT_TYPE(indptr, dl_int32);
+  CHECK_INPUT_TYPE(indptr, dl_dtype_for<IdType>());
 
   TVM_FFI_ICHECK_EQ(head_dim_qk, head_dim_vo)
       << "CUDA cores template only supports equal head dim for QK and VO, please use tensor "
@@ -134,9 +134,9 @@ void BatchDecodeWithPagedKVCacheRun(TensorView float_workspace_buffer,
                                     TensorView o, Optional<TensorView> maybe_lse,
                                     int64_t kv_layout_code, int64_t window_left,
                                     bool enable_pdl ADDITIONAL_FUNC_PARAMS) {
-  CHECK_INPUT_TYPE(paged_kv_indptr, dl_int32);
-  CHECK_INPUT_TYPE(paged_kv_indices, dl_int32);
-  CHECK_INPUT_TYPE(paged_kv_last_page_len, dl_int32);
+  CHECK_INPUT_TYPE(paged_kv_indptr, dl_dtype_for<IdType>());
+  CHECK_INPUT_TYPE(paged_kv_indices, dl_dtype_for<IdType>());
+  CHECK_INPUT_TYPE(paged_kv_last_page_len, dl_dtype_for<IdType>());
 
   DecodePlanInfo plan_info;
   plan_info.FromVector(std::vector<int64_t>(plan_info_vec.begin(), plan_info_vec.end()));
--- a/csrc/batch_prefill.cu
+++ b/csrc/batch_prefill.cu
@@ -51,6 +51,8 @@ Array<int64_t> BatchPrefillWithKVCachePlan(
     int64_t num_kv_heads, int64_t page_size, bool enable_cuda_graph, int64_t head_dim_qk,
     int64_t head_dim_vo, bool causal, int64_t window_left, int64_t fixed_split_size,
     bool disable_split_kv, int64_t num_colocated_ctas = 0, int64_t uniform_q_len = 0) {
+  CHECK_INPUT_TYPE(qo_indptr, dl_dtype_for<IdType>());
+  CHECK_INPUT_TYPE(kv_indptr, dl_dtype_for<IdType>());
   size_t float_workspace_size_in_bytes =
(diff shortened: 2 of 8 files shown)


SamMausberg:
Both points are addressed in the updated revision.

- `kv_len_arr` in the SM90 plan: it now checks that the tensor is a signed or unsigned integer of the compiled `IdType` width (one lane), so an int64 tensor from a direct FFI caller is rejected while the documented uint32 `seq_lens` still passes; a strict `IdType` check would have rejected that documented input.
- `fast_decode_plan`: it now runs the same `_check_index_tensor` checks as `_plan_impl` and stores `.contiguous()` views in the non-graph path, which is a no-op for the contiguous tensors this copy-free path expects; two tests cover the int64 rejection and the strided input.

PR #5121, opened 2026-09-11
feat(sm120): instantiate 128/256-token extra-cache pages in sparse-MLA prefill

qqtang-code:
## 📌 Description

`dispatch_dsv4_dual` only instantiated `extra_page_block_size == 64` (plus the 2-token WFP8 variant), so a dual-cache prefill whose candidate pool pages its extra KV cache at **128 or 256 tokens** returned `false` from the dispatcher and aborted the engine:

```
tvm.error.InternalError: Check failed: (ok) is false: Unsupported sparse-MLA prefill
configuration: model=DSV4 num_heads=8 topk=128 page_block_size=64 topk_extra=512
extra_page_block_size=128
```

DeepSeek-V4.1-Flash's hierarchical candidate pool is exactly that shape, so on SM120 the model could only be served by falling back to a non-FlashInfer sparse-MLA implementation.

The page size is purely an **addressing** parameter for these kernels — `prefill_kv_entry_base` derives `(idx / PAGE_BLOCK_SIZE, idx % PAGE_BLOCK_SIZE)` per token, and the only page-size-dependent special case is `USE_WFP8_ROW_XOR = DUAL_CACHE && (PAGE_BLOCK_SIZE_EXTRA == 2)`, which stays `false` for 128/256 exactly as for 64. The two sizes therefore need instantiations, not new kernel code.

This PR adds 128/256 branches to both the full-tile and the general dual-cache dispatchers (support guards + `DISPATCH_*_BY_NH_PBSX`), i.e. ten additional kernel instantiations across `num_heads ∈ {8, 16, 32, 64, 128}`.

### Performance (the reason this matters)

Serving `deepseek-ai/DeepSeek-V4.1-Flash` on **8× RTX 6000D (SM120)**. Protocol for every row: 16k input / 512 output, concurrency 16, N=100, warmup 16 — the DeepSeek-V4 bring-up scenario.

**Like-for-like A/B, prefix caching disabled on both sides** (`--disable-radix-cache`), 3 runs each:

| prefill path | total tok/s | input tok/s | output tok/s | TTFT median | TTFT p99 | ITL median |
| --- | --- | --- | --- | --- | --- | --- |
| Triton sparse-MLA fallback (the only way to serve these shapes before this PR) | 1,978 | 1,923 | 55 | 61,539 ms | — | 64.4 ms |
| **FlashInfer with this PR** | **5,850** | 5,685 | 164 | **19,148 ms** | 35,408 ms | 26.5 ms |

* **3.0× throughput, 3.2× TTFT, 2.4× ITL** — attributable to the prefill path itself, with caching held constant.
* FlashInfer side reproduced at 5,843.69 / 5,850.38 / 5,852.39 tok/s (median 5,850.38); Triton side at 1,977.19 / 1,977.86.
* The server reaches ready with zero `Unsupported sparse-MLA prefill configuration` errors.

**In the configuration we actually serve** (prefix caching enabled, which this model's agent traffic wants), the same protocol reaches 17,950 tok/s at 963 ms median TTFT — the extra gain there is prefix reuse, not the kernel, so it is listed separately rather than folded into the number above.

## 🔍 Related Issues

No issue filed. Found while bringing DeepSeek-V4.1-Flash up on RTX 6000D (SM120); the model's candidate pool issues `extra_page_block_size` 128 and 256 at different phases of the same run, so both are added.

### ✅ Pre-commit Checks

> Not ticked: this environment has no `pre-commit`, so the hooks were not run end to end. The format-affecting hooks were run individually at the versions pinned in `.pre-commit-config.yaml`: **clang-format 19.1.1** on `csrc/sparse_mla_sm120_prefill.cu`, and **ruff 0.12.8** (`ruff-check` + `ruff-format`) on `tests/attention/test_sparse_mla_sm120.py` — both clean. The remaining hooks (tabs, trailing whitespace, EOF) are satisfied by inspecting the diff. Please re-run `pre-commit run --all-files` on your side before merging.

## 🧪 Tests

`tests/attention/test_sparse_mla_sm120.py` already parametrizes the dual-cache prefill over `extra_pbs ∈ {64, 2}` and checks numerically against `_ref_sparse_attn`. This PR extends that parametrization with `(extra_topk, extra_pbs) = (512, 128)` and `(512, 256)` across all five head counts, so the new instantiations are exercised for dispatch acceptance **and** numerical correctness. The parametrized cases fail on `main` with the dispatcher error above and pass with the patch (verified on RTX 6000D, SM120).

## Reviewer Notes

* Both dispatchers are widened for symmetry: the full-tile path (used when no `topk_length` is supplied and `topk_extra % BI == 0`) and the general dual path.
* `extra_topk` stays a runtime argument; only the page size is template-bound because it changes the KV stride, so no new launch parameters are introduced.
* If you would rather keep the instantiation set tighter, dropping `256` is safe for DeepSeek-V4.1 prefill but the model does hit it in the same end-to-end run, so both are kept.

### DIFF
--- a/csrc/sparse_mla_sm120_prefill.cu
+++ b/csrc/sparse_mla_sm120_prefill.cu
@@ -436,6 +436,12 @@ inline bool dispatch_dsv4_single(int num_heads, int topk, int page_block_size, c
   return false;  // unreachable
 }
 
+// Dispatch a DSV4 dual-cache (hierarchical candidate pool) sparse-MLA prefill
+// to a pre-built kernel instantiation. The main cache must use 64-token pages.
+// Uniform candidate counts take the full-tile path; per-token topk lengths take
+// the length-aware path. Both select by num_heads (8-128, folded into 1-2 head
+// groups) and extra-cache page size (2, 64, 128 or 256 tokens). Returns false
+// when the configuration has no instantiation so the caller can fall back.
 inline bool dispatch_dsv4_dual(int num_heads, int topk, int topk_extra, int page_block_size,
                                int extra_page_block_size, const bf16* Q, const uint8_t* KV,
                                const int32_t* indices, const uint8_t* KV_extra,
--- a/csrc/sparse_mla_sm120_prefill.cu
+++ b/csrc/sparse_mla_sm120_prefill.cu
@@ -388,6 +388,11 @@ inline bool dispatch_dots3_swa_sg(int num_heads, int topk, int page_block_size, 
 #undef DISPATCH_DOTS3_SWA_SG
 }
 
+// Dispatch a DSV4 single-cache sparse-MLA prefill to a pre-built kernel
+// instantiation. The cache must use 64-token pages; num_heads (8-128) folds
+// into 1-2 head groups and the compute mode follows the candidate count.
+// Returns false when the configuration has no instantiation so the caller
+// can fall back.
 inline bool dispatch_dsv4_single(int num_heads, int topk, int page_block_size, const bf16* Q,
                                  const uint8_t* KV, const int32_t* indices, const float* attn_sink,
                                  bf16* output, float* out_lse, float sm_scale, int num_tokens,
--- a/csrc/sparse_mla_sm120_prefill.cu
+++ b/csrc/sparse_mla_sm120_prefill.cu
@@ -446,7 +446,8 @@ inline bool dispatch_dsv4_dual(int num_heads, int topk, int topk_extra, int page
                                const int* topk_length_extra_ptr, cudaStream_t stream) {
   if (page_block_size != 64) return false;
   if (topk_length_ptr == nullptr && topk_length_extra_ptr == nullptr && topk_extra % BI == 0 &&
-      (extra_page_block_size == 64 || extra_page_block_size == 2)) {
+      (extra_page_block_size == 64 || extra_page_block_size == 2 || extra_page_block_size == 128 ||
+       extra_page_block_size == 256)) {
 #define DISPATCH_DUAL_MG_FULLTILE(NH, PBSX, NHG)                                                   \
   launch_prefill_mg_dual_fulltile<ModelType::DSV4, NH, 64, PBSX, NHG>(                             \
       Q, KV, indices, KV_extra, idx_extra, attn_sink, output, out_lse, sm_scale, num_tokens, topk, \
@@ -477,6 +478,10 @@ inline bool dispatch_dsv4_dual(int num_heads, int topk, int topk_extra, int page
 
     if (extra_page_block_size == 64) {
       DISPATCH_FULLTILE_BY_NH_PBSX(64);
+    } else if (extra_page_block_size == 128) {
+      DISPATCH_FULLTILE_BY_NH_PBSX(128);
+    } else if (extra_page_block_size == 256) {
+      DISPATCH_FULLTILE_BY_NH_PBSX(256);
     } else {
       DISPATCH_FULLTILE_BY_NH_PBSX(2);
     }
@@ -520,6 +525,15 @@ inline bool dispatch_dsv4_dual(int num_heads, int topk, int topk_extra, int page
     DISPATCH_BY_NH_PBSX(64);
   } else if (extra_page_block_size == 2) {
     DISPATCH_BY_NH_PBSX(2);
+  } else if (extra_page_block_size == 128) {
+    // DeepSeek-V4.1's hierarchical candidate pool pages its extra KV cache
+    // at 128 or 256 tokens. The page size is purely an addressing parameter
+    // (prefill_kv_entry_base splits a token index into page / in-page offset)
+    // and the WFP8 row-XOR swizzle is gated on the 2-token variant, so these
+    // sizes need instantiations rather than new kernel code.
+    DISPATCH_BY_NH_PBSX(128);
+  } else if (extra_page_block_size == 256) {
+    DISPATCH_BY_NH_PBSX(256);
   }
   return false;
 #undef DISPATCH_BY_NH_PBSX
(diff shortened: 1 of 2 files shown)


qqtang-code:
Cross-references for reviewers:

* sgl-project/sglang#38969 — makes the SM120 sparse-MLA wrapper fall back to Triton **per call** when the flashinfer backend has no instantiation for the shape (this is what makes the DeepSeek-V4.1 candidate-pool prefill survivable on stock FlashInfer).
* sgl-project/sglang#38970 — builds the DeepGEMM schedule metadata the V4.1 low-ratio indexer needs on SM120 (base branch `dsv4.1`).

With #5121 present, the shapes added here stay on FlashInfer; sglang#38969 only kicks in for whatever the envelope still does not cover.

qqtang-code:
Requesting CI authorization when convenient — the summary job reported:

```
CI skipped (pending authorization)
A contributor who can label the PR, or a `ci-users` member, can comment ...
```

The change is a pure dispatcher widening plus two parametrized cases in `tests/attention/test_sparse_mla_sm120.py`; it needs the SM12x lane (RTX 6000D) to actually exercise the new instantiations. Thanks!

qqtang-code:
Thanks — good catch. Fixed in a20cb45: `test_sparse_mla_sm120_prefill_dsv4_dual_extra_topk_length_truncation` is now parametrized over `extra_pbs` 64/128/256 (2 heads × 3 lengths × 3 page sizes = 18 cases), so the length-aware path that bypasses the full-tile dispatcher covers the new page sizes too, each still compared against the masked `_ref_sparse_attn` reference.

`ruff 0.12.8` (pinned in `.pre-commit-config.yaml`) reports no findings and no formatting diff on the file.

yiakwy-xpu-ml-framework-team:
Hi @qqtang-code I am working with sm121a (DGX spark). 

Checking the implementation of the kernel is a simple multi stage gemm : https://github.com/qqtang-code/flashinfer/blob/a20cb459ff295fdf85aa1b8b6a8f7864e0683a31/csrc/sparse_mla_sm120_nvfp4_tile.cu#L65

The kernel didn't use TMA for data loading.

We have demonstrate that using TMA in DGX Spark sm121a is handy for nvfp4 gemm : https://github.com/yiakwy-xpu-ml-framework-team/flash-float-jit-kernels/blob/main/jit_kernel/csrc/velox_voice/dgx/block/nv_block_wasp_mxfp4.h and up 2x accleration copmared to torch fp16.

This is for your consideration.

qqtang-code:
Thanks @yiakwy-xpu-ml-framework-team — we took the suggestion for a spin on sm_120 (RTX 6000D, 156 SMs).

Scope first: this PR only widens the dispatch envelope in `sparse_mla_sm120_prefill.cu`. The file you linked (`sparse_mla_sm120_nvfp4_tile.cu`) is a standalone validation tile for the QK/PV operand and scale-factor mappings, and you're right that its operand loads are plain synchronous word loops.

We built three load variants on that same M16N32K64 tile — (a) the sync word-loop as shipped, (b) `cp.async.bulk` + mbarrier, (c) `cp.async.bulk.tensor.2d` with a host-encoded `CUtensorMap` (swizzle NONE, so the shared layout and the downstream `ldmatrix`/MMA path stay untouched) — and compared on hardware:

* **Correctness:** the tensor-map path reproduces the word-loop path's MMA outputs bit-for-bit, for a single tile and across a 131k-tile sweep. So TMA operand delivery for these layouts works on sm_120 and is layout-compatible with the existing mapping — that part of the wiring is de-risked.
* **Throughput:** with a DRAM-sized working set and multi-CTA pressure, all three paths land within ±9% of each other — 1117 / 1214 / 1184 GB/s for loop / bulk / TMA (load-only), and the end-to-end tile loop (load + `ldmatrix` + 4×MMA per tile) is unchanged. At these shapes the pipeline is bandwidth-bound, so offloading the address arithmetic buys no throughput.

Framing context: the production SM120 sparse-MLA kernels already issue `cp.async.bulk` (the TMA unit's non-tensor bulk copy) with mbarrier `expect_tx` tracking, L2 evict-first/last policy hints, and ahead-of-gather L2 prefetch — see `io_bulk_gather_tile` in `include/flashinfer/attention/sparse_mla_sm120/common/kv_cache_io.cuh` and the FP4 decode kernel. What's missing is the tensor-map flavor specifically. Where it could still pay is exactly the issue chain your comment points at: `io_bulk_gather_tile` computes `idx / page_block_size`, `idx % page_block_size` and a byte address per candidate on the issue path (the code even stages `idx` a tile ahead to keep that LDG off the copy chain), and a tensor map over the block-structured KV layout would replace that with a coordinate. One caveat we hit: the candidate-major PV operand has 4-byte rows, below the 16-byte inner-box granularity `cuTensorMapEncodeTiled` requires, so that operand can't move to tensor maps without repacking.

On the 2× number for sm121a: since the baseline there is torch fp16, we read that as the fp4 tensor-core + pipelining win rather than TMA-vs-bulk per se — consistent with what we see here, where the copy flavor stops mattering once the pipeline is bandwidth-bound.

We've noted the tensor-map gather as a follow-up direction for the sparse orchestrator (revisit if the gather issue chain shows up in profiles), and happy to compare notes if you prototype the sm121a side.

qqtang-code:
@coderabbitai docstring

qqtang-code:
Correction to the performance section (now updated in the description): the two rows I first posted were **not** measured under the same prefix-caching configuration — the pre-patch row ran with `--disable-radix-cache` (the protocol used by the earlier DeepSeek-V4 bring-up) while the post-patch row ran with prefix caching enabled, which is the configuration we serve. Re-stated properly:

* like-for-like, no prefix cache: **1,978 → 5,857 tok/s (3.0×), TTFT 61.5 s → 19.2 s (3.2×)** — the part attributable to the prefill path;
* with a warm prefix cache (steady state as deployed): 17,950 tok/s, TTFT 0.96 s — the additional gain is prefix reuse, which the pre-patch configuration cannot exploit because the engine aborts on these shapes.

Sorry for the sloppy first pass; the numbers are now labelled by their actual conditions.

PR #5123, opened 2026-09-11
feat(prims-ts): Sage attention (8-bit Q/K/V with dequantization scales) for the PrimTS decode kernels

heyuhhh:
## 📌 Description

Adds Sage attention (per-block dequantization scales, trtllm-gen semantics) to the PrimTS `fmha_decode` Q64/KV256 and Q128/KV128 profiles for contiguous K/V, in dense mode and in block-sparse mode (exact and proxy routes), with E4M3 or INT8 Q/K and E4M3 P/V. The byte-wide kernels run at 0.81 to 0.91 of the BF16 kernel time of the same plan on the block-sparse and large dense cases and match it on the small dense case.

**API**

- `BlockSparseTSWrapper.plan(...)` gains `use_block_sparse` (default `True`; `False` plans dense attention over the whole contiguous K/V sequence with the same Q-tile/KV-route selection and profile matrix as block-sparse plans), `sage_config=SageAttentionConfig(...)` and `v_data_type`. `SageAttentionConfig` is the compile-time recipe (`q_block_size`, `k_block_size`, `v_mean`); it joins the kernel cache key.
- `BlockSparseTSWrapper.run(...)` gains `sage=SageAttentionParams(...)` with the scale tensors of that call: `q_scale`, `k_scale`, `v_scale`, optional `v_mean`, and `k_summary_scale` for proxy plans. The plan resolves their expected shapes once (`sage_scale_shapes`) and every run is validated against them (shape, dtype, device, contiguity). The scale layout is TensorRT-LLM's `sageQuant` flat layout, so the quantizer output plugs in without a copy.
- `block_sparse_attention(..., sage=SageAttentionParams(...), sage_config=SageAttentionConfig(...))` runs Sage in one shot in the dense and block-sparse modes; without `sage_config` the recipe is the default `SageAttentionConfig` with a V mean exactly when the scales carry one, and the INT8 recipe is inferred from the Q/K/V dtypes. `use_block_sparse=False` runs the dense mode in one shot. The paged wrappers accept neither `sage` nor INT8. The trace templates gain the dense variants `prims_ts_block_sparse_dense` and `prims_ts_block_sparse_wrapper_dense`; trace capture of the Sage scale tensors is a follow-up.
- The decode kernel configuration owns the Sage recipe rules, each next to the rule it extends: the dtype recipe (8-bit Q/K with E4M3 V and a 16-bit output, INT8 needs E4M3 V) in `validate_dtypes`, the block sizes and the streamed two-instance Keeps requirement in `validate_sage_profile`, the direct-output-grid exclusions shared with block-sparse routes in one helper both validators call, and the qualified recipes in a Sage table beside the contiguous grouped-Keeps recipes. The wrapper adds only the Q == K check its compile key needs and surfaces the kernel's error from `plan()`. The proxy summary geometry (summary count, block mass, ragged tail, summary groups and the tail's compile-time fragment) is a set of configuration properties that the prepare kernel, the softmax and the tests read.
- Every contiguous plan (dense or block-sparse, exact or proxy, BSR or bitmask, with or without Sage) compiles one adapter. Its routing tensors, summaries, route storage and scale tensors are optional slots that are `None` at compile time and at every call when the plan does not use them, and the wrapper issues one launch call. The paged adapter keeps its own signature.

**Kernel**

- Scales enter the softmax before the row maximum and the exponential (`S = sfQ * sfK * S8`); P keeps the static 448 E4M3 scale; V scales are applied per channel in the epilogue.
- Where a tile's `sfK` words live during the two softmax passes is a strategy object built per softmax instance and shared by its S and P resources: a rotating register array for K blocks of 16 tokens and more, a two-tile SMEM ring for the 4-token and one-token blocks. Both strategies take a tile's words through one loader (from `k_scale` on a dense tile, from the staged route words on a block-sparse route) and hand the passes one fragment's scales at a time through the same calls, with one algebra: raw `sfK` per scale group, `sfQ` applied once per tile.
- INT8 scores are accumulated onto an FP32 bias seeded by one BF16 MMA step, so the softmax reads them as FP32 without per-element conversion.
- The SMEM the byte-wide tiles free is reinvested: four-stage K/V ring (the depth at which every K load takes the slot the previous QK freed instead of waiting for a V slot held until its tile's PV), and output columns split between the two lane groups of the tail.
- The softmax and correction mechanisms that do not depend on the element width are written once and used by every dtype: one exponent addend formed per row, the rolled masked max pass with a compile-time tail location (proxy routes are fixed summary groups, so the ragged final summary sits in a known half, fragment and lane for every sequence length), seeded probability sum chains, the CLC response slots in the unified SMEM block, one correction store wait per O stage and one KV256 tail denominator block. The 16-bit kernels pick them up as well, so their SASS differs from `main`; the second table below records the effect.
- Dense plans of every dtype follow the decode kernel's launch heuristic between the static grid and the persistent scheduler (persistent above one resident wave). The kernel config qualifies the contiguous grouped-Keeps recipes from one set for block-sparse and dense launches with either scheduler, so dense Q128/KV128 in float16, bfloat16 and the Sage dtypes may run persistent as block-sparse plans already did.

**Breaking change**

Block-sparse proxy routes take `v_summary` as the per-block **mean** of V (was the sum), for BF16/FP16 as well as Sage. A sum is not detectable by the kernel and scales the output by the block mass. Producers of `v_summary` (the TensorRT-LLM SOL predictor: `v_sum / valid_tokens`) must change. The proxy block's mass enters the softmax logit (`x_b + log2(mass)`) instead of weighting the denominator.

**Performance** (B200, `benchmarks/bench_primts_proxy_block_sparse.py`, CUDA Graph replay minimum of the kernel, D=128, FP8 recipe `(fp8, fp8, (1, 16, 1))` versus the BF16 kernel of the same plan, both scheduled by the same launch heuristic, three interleaved legs per case; measured at the tip of this branch on `01045366e`)

| Case | BF16 | FP8 Sage | ratio |
| --- | ---: | ---: | ---: |
| dense contiguous S=10800 H=40 | 2142.6 us | 1778.2 us | 0.83 |
| dense contiguous S=4096 H=8 | 66.7 us | 66.7 us | 1.00 |
| block-sparse exact S=4096 H=8, density 0.25 | 37.8 us | 34.4 us | 0.91 |
| block-sparse exact S=15360 H=40, density 0.125 (VSA shape) | 689.0 us | 556.4 us | 0.81 |
| block-sparse exact S=10800 H=40, density 0.175, hot pattern (SOL shape) | 471.7 us | 386.6 us | 0.82 |
| block-sparse proxy, same SOL shape | 542.2 us | 467.5 us | 0.86 |

The INT8 recipe `(int8, fp8, (1, 16, 1))` is within 1-4% of the FP8 recipe (SOL exact 399.9 us, proxy 472.5 us). The ratios do not depend on the logit distribution (checked at 0.17, 1 and 3 nats spread). Random 8-bit inputs and scales; the harness is not part of this PR.

Small K blocks (same harness, CUDA Graph replay minimum over three legs at the tip on `main`, so the columns are comparable with each other):

| Case | `k_block_size` 16 | 4 | 1 |
| --- | ---: | ---: | ---: |
| block-sparse exact SOL shape, FP8 | 387.5 us | 428.3 us | 501.0 us |
| block-sparse proxy SOL shape, FP8 | 469.3 us | 517.4 us | 613.6 us |
| block-sparse exact SOL shape, INT8 | 400.7 us | | 537.9 us |
| block-sparse proxy SOL shape, INT8 | 474.3 us | | 650.6 us |
| dense contiguous S=10800 H=40, FP8 | 1782.1 us | 1904.9 us | 2167.2 us |
| block-sparse exact S=4096 H=8 GQA 2 (Q128/KV128), FP8 | 38.1 us | | 42.3 us |

The one-token block reads 32 scale words per K32 fragment from SMEM and scales every score pair with its own multipliers, which is where its cost comes from; the 16-token and larger blocks are unchanged by the small-block support (their exact, INT8, dense and 16-bit kernels are byte-identical; the FP8 SOL proxy kernel differs only in register allocation at an equal instruction count).

The byte-wide K/V ring runs at four stages. Against a five-stage ring (same harness, three paired legs): dense S=10800 H=40 0.982, INT8 proxy SOL shape 0.967, FP8 proxy 0.996, FP8 and INT8 exact within 0.3%, Q128 exact 0.995. The four-stage build keeps one pair of persistent scheduler words on an 8-byte stack per warp role (16 bytes on INT8), reloaded once per tile off the softmax critical path.

16-bit kernels of this branch relative to `main`'s (same plan, same launch policy, paired ABBA legs, geometric mean of the per-leg ratios):

| 16-bit case | time ratio |
| --- | ---: |
| block-sparse exact SOL shape (bf16 Q64/KV256) | 0.95 |
| block-sparse proxy SOL shape (bf16 Q64/KV256, mean contract on both sides) | 0.93 |
| block-sparse exact S=4096 H=8 (bf16 Q64/KV256) | 0.98 |
| block-sparse exact S=4096 H=8 GQA 2 (bf16 Q128/KV128) | 0.98 |
| block-sparse exact VSA shape | 0.96 |

The dense contiguous mode has no counterpart on `main`. The paged fp16 Q8/Q16/Q32 Swaps kernels (D=128, page 32) publish bitwise-equal outputs in the decode suite.

**Not supported / assumptions**

- `k_block_size`: 1, 4, 16, 32, 64, 128, 256 (the TensorRT-LLM recipes `(1,1,1)` and `(1,4,1)` included). Blocks below 16 tokens keep a tile's `sfK` words in SMEM instead of a per-lane register array; they cost more than the larger blocks (table below). `q_block_size`: a power of two up to the Q tile.
- Scale positivity and finiteness are assumed and not checked at run time (the quantizer floor guarantees them).
- Sage requires the streamed Keeps profiles (Q64/KV256, Q128/KV128), D=128, contiguous K/V; no split-KV, sliding window or attention sinks.

**Numerics**

- E4M3 P with the 448 anchor; deferred anchor updates stay off for 8-bit P.
- INT8: two per-group roundings versus converting every element (exponent addend at most 0.75 quantized score units; proxy ragged-tail shift at most 0.5 unit), below INT8 quantization noise; pinned by an extreme-score test.
- Row sums in the P pass are accumulated as seeded chains (FP32 reassociation; max abs 1e-3 on bf16 output versus the previous kernel).
- Test tolerances: dequantized-reference fidelity 2e-3 dense / 2e-2 block-sparse; recipe-level tests against BF16 attention with TensorRT-LLM tolerances plus per-head relative L2 bounds (INT8 0.075, E4M3 0.10 / 0.20).
- INT8 proxy summaries: one `k_summary_scale` per 16 summaries covers 1024 tokens of K. With the torch quantizer on synthetic summary sequences whose block means spread in magnitude (log-normal, sigma 1.0), the worst summary of a 16-summary block carries a relative error of 0.59 (rounds to a few INT8 steps), while per-summary scales (`k_block_size = 1`) keep every summary near 0.01; E4M3 summaries do not depend on the block size. INT8 recipes with proxy routes should quantize the summary sequence with `k_block_size = 1`.

**Also in this PR**

- The contiguous K/V TensorMap in `fmha_decode_launch` changed stride semantics (BHSD to BSHD); it had no caller before this branch.
- The dense branch of the contiguous adapter passes typed zeros for the paged-KV and Q-token launcher arguments, which #4829 and #4996 forward into typed kernel parameters.
- `kv_dtype` on the kernel config is the K dtype and `v_dtype` the V dtype; both are always concrete. `make_decode_config` sets V from K unless `v_dtype` / `v_data_type` is given (INT8 K requires E4M3 V), and the block-sparse compile key carries the resolved output and V dtype keys.

## 🔍 Related Issues

Follows #5002 (PrimTS block-sparse attention optimizations), which this branch was developed on. Follow-ups tracked outside this PR: the TensorRT-LLM SOL predictor must emit V means; trace capture of the Sage scale tensors.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

On B200 at the tip of this branch on `main` (`01045366e`):

- `tests/attention/test_attention_ts_sage.py`: 208 passed (new; dequantized-reference fidelity for dense and block-sparse including proxy sequences that span two summary groups, the one-shot form against the planned wrapper, recipe-level tests against BF16 attention, INT8 extreme-score and floor-scale tests, and the 4-token and one-token K blocks on dense, exact and proxy plans)
- `tests/attention/test_attention_ts_block_sparse.py`: 290 passed, 1 skipped
- `tests/attention/test_attention_ts_decode.py`: 397 passed; the 11 QToken-KvBlock-Sparse tests of `main` fail in this container on `main` as well (their JIT C++ op needs a CCCL that provides `cuda::fast_mod_div`)
- `tests/trace/test_fi_trace_template_consistency.py`: 778 passed, 1 skipped

The Sage test file also passes at every commit of the series from the dense Sage commit on (106, 106, 135, 157, 161, 161, 161, 161, 161 and 208 tests passed along the ten commits), so the series bisects.

Shared helpers in `tests/attention/prims_ts_test_utils.py`; the torch quantizer reference in `tests/attention/sage_quant_reference.py`.

## 🔬 Experimental Track

```experimental-tests
#
#
```

## Reviewer Notes

- The fourteen commits are meant to be read in order. Two preparatory commits touch only pre-existing code: a `refactor` that shares the softmax schedule body between the two softmax instances and forms the exponent addend once for every P path, and a `test` commit that moves the block-sparse test input builders into a shared module. Then six `feat` commits: FP8 streamed profiles, dense mode of the wrapper (which also unifies the contiguous adapters into one with optional slots), Sage scales on dense (the Sage slots join that adapter, the one-shot form, the recipe table and validators), the proxy `v_summary` mean / mass-in-logit contract change on its own (the only commit that changes results for existing 16-bit callers), Sage scales on block-sparse routes, INT8 recipe. Then three `perf` commits (SMEM reinvestment with the four-stage ring and the persistent scheduler config built once, softmax passes, KV256 tail), one `docs`, a `perf` commit that issues one correction store wait per O stage, and one `feat` commit that accepts the K block sizes 4 and 1 through the SMEM scale strategy (the kernels of the larger blocks keep their SASS apart from a register re-allocation of the FP8 SOL proxy kernel). Each piece of code appears once, in its final shape; no Sage-specialized adapter exists at any point of the series.
- The 16-bit kernels are not byte-identical to `main`: the mechanisms listed under Kernel are shared by every dtype. The second performance table gives their effect on the 16-bit kernels; the paged fp16 Q8/Q16/Q32 kernels produce bitwise-equal outputs at the measured shapes.
- Dense plans follow the same launch heuristic as block-sparse plans, so 16-bit dense goes persistent above one wave as well.
- The `v_summary` mean contract is the one result change for existing callers; please weigh in if a sum-compatible mode is wanted instead.
- Draft while the TensorRT-LLM side (V means from the SOL predictor, `sageQuant` recipe wiring) is prepared.


flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/helpers_common.py
```diff
@@ -154,18 +154,18 @@ def _swaps_routed_coordinate(
...
 
 
 def _mma_kind_for_qk(cfg: FmhaDecodeConfig) -> prims.Tcgen05MMAKind:
-    """Select the tcgen05 MMA opcode family for the QK GEMM."""
+    """Select the tcgen05 MMA kind of the QK GEMM from the Q/K dtype.
+
+    Int8 accumulates INT32 scores; E4M3 and the 16-bit types accumulate FP32.
+    """
+    if cfg.uses_int32_scores:
+        return prims.Tcgen05MMAKind.INT8
```
PerkzZheng, line 165: I think INT8 MMAs are only supported with sm100, while removed with sm103 and next. we should guard it.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/helpers_common.py
```diff
@@ -400,17 +437,25 @@ def _major_k_stride_bytes(dtype_bytes: int, headdim: int) -> int:
...
+def _masked_weighted_sum(terms: tuple) -> Float32:
+    """Return the sum of ``value * weight`` over ``(uses, value, weight)`` terms.
+
+    The terms are added in order and a term whose ``uses`` is false is skipped.
+    """
+    total = Float32(0.0)
+    for term_idx in cutlass.range_constexpr(len(terms)):
+        uses, value, weight = terms[term_idx]
+        if uses:
+            total += value * weight
```
PerkzZheng: should we use fma2 explicitly ?

tests/attention/test_attention_ts_decode.py
```diff
@@ -3560,8 +3721,235 @@ def test_attention_ts_decode_streamed_p_fragments_follow_kv_tile(
...
+    from cutlass.experimental.primitives import Tcgen05MMAKind
+
+    cfg = make_sage_decode_config(
+        tile_size_q=tile_size_q,
+        tile_size_kv=256 if tile_size_q == 64 else 128,
+        qkv_dtype=Int8,
+        v_dtype=Float8E4M3FN,
+        o_dtype=o_dtype,
+    )
+    assert cfg.use_sage_attention and cfg.uses_int32_scores
```
PerkzZheng: we might want to clean up those over-engineering tests. Similarly for above ones, please help clean them. Thanks.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/tmem_s.py
```diff
@@ -2182,35 +2551,57 @@ def _compute_softmax_loop_keeps_fragments(
...
+        group maxima fold with ``sfK`` alone and ``sfQ`` applies once to the
+        tile maximum.
         """
         cfg = self.cfg
         assert cfg.streams_tmem_p_fragments
         num_fragments = cfg.num_softmax_score_fragments
         fragment_regs = cfg.softmax_score_fragment_regs
         # The seven-slot softmax metadata ABI carries exactly four token words.
         assert num_fragments == 4 and fragment_regs == 32
+        if cutlass.const_expr(not use_sparse):
```
PerkzZheng: curious why we need to initialise it if it is not used IIUC ?

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/tmem_s.py
```diff
@@ -2326,41 +2717,178 @@ def _compute_softmax_loop_keeps_fragments(
...
+        # fold and the TMEM write-back, so the P pass reloads a score whose
+        # exponent already carries the right mass. The route kind is uniform
+        # across the CTA, so the branch below is uniform as well.
+        proxy_max_shift = Float32(0.0)
+        tail_shift = Float32(0.0)
+        tail_fragment_mask = Int32(0)
+        tail_lane: Constexpr[int] = 0
+        shifts_tail: Constexpr[bool] = False
+        route_is_proxy = cutlass.Boolean(False)
+        mixed_scales: Constexpr[bool] = use_sparse and cfg.sage_mixed_k_geometry
```
PerkzZheng, line 2537: you might want to explain what does mixed_scales mean and why it is needed.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_config.py
```diff
@@ -186,7 +211,22 @@
...
+
+def _dtype_bytes(dtype: type) -> int:
+    """Return the storage bytes of one element of a CUTLASS numeric type."""
+    return dtype.width // BITS_PER_BYTE
+
+
+# ``(Q, K/V, O)`` dtype recipes of the Q64/KV256 profile without Sage
+# attention: 16-bit profiles keep one dtype throughout, and FP8 Q/K/V
+# publishes FP8 or FP16 output, matching the FP8 D128 Keeps profiles.
+_KV256_DTYPE_RECIPES = {
```
PerkzZheng, line 213: why we need to specify recipes for KV256 only ?

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/tmem_corr.py
```diff
@@ -571,6 +609,24 @@ def _separate_partial_norm_scale(self, sum_val: Float32) -> Float32:
...
+        scale is the row's public output-domain normalization. Fused GMEM and
+        cluster reductions receive unnormalized partials: 16-bit ones as they
+        are, byte-wide ones with the 448 P scale removed before they are
+        narrowed to 16 bits (the reducer restores it).
+        """
+        cfg = self.cfg
+        partial_scale = Float32(1.0)
+        if cutlass.const_expr(cfg.use_separate_reduction_kernel):
+            partial_scale = self._separate_partial_norm_scale(denominator)
+        elif cutlass.const_expr(cfg.use_8bit_qkv):
```
PerkzZheng: would that also apply to int8 ? we probably want to introduce mma_qk_dtype and mma_pv_dtype if possible.

qsang-nv:
`fmha_decode_kernel.py`'s generic SMEM-total check (`smem_bytes(smem_allocator) > budget`) only runs when `cfg.uses_q_token_kv_block_sparse_page_membership`. The other static check, `_validate_kv256_static_config` in `fmha_decode_config.py` (`pipeline_smem_bytes = q_stages * smem_q_tile_bytes + kv_stages * smem_kv_tile_bytes`), covers only Q/KV tile bytes and excludes the Sage `sfK` scale ring (`SmemSageKScales`: 4 KiB total for a KV256 `sage_k_block_size=1` config — 2 instances × 2 slots × 2 halves × 4 fragments × 32 scales × 4 bytes). For an explicit `kv_stages=6`, INT8, KV256, `sage_k_block_size=1` config, the static resource lower bound reaches at least 232 KiB against `TOTAL_SMEM_BUDGET_KIB = 218` (itself kept below the 227 KiB SM100 hardware carveout per `fmha_decode_constants.py`'s comment), and neither check catches it. The JIT-time `TaskManager(..., skip_validation=True, ...)` construction here only demotes such a failure to a warning per its own comment; the stricter `build_decode_task_manager()` path (`skip_validation=False`) would catch it, but only if a test or caller exercises this exact config through that offline path.

`flashinfer/attention/prims_ts/README.md`'s new "### Performance" section states specific B200 kernel-time ratios inline as prose (e.g. "0.83 on dense S=10800 H=40 ... 0.81 on a VSA-shaped block-sparse case, ... 0.86 on its proxy variant"). These are measurements for a particular revision, not an API contract, and nothing pins them to stay accurate as the kernel is tuned further.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/sage_scales.py
```diff
@@ -0,0 +1,607 @@
...
+        )
+        buffer = self.words(stage_info.context)
+        tile_base = self.tile_base(stage_info)
+        for round_idx in cutlass.range_constexpr((num_words + threads - 1) // threads):
+            word_idx = warp_grp_thread_idx + Int32(round_idx * threads)
+            if word_idx < Int32(num_words):
+                buffer[tile_base + word_idx] = word_value(
+                    word_idx, word_idx // Int32(arr_size), word_idx % Int32(arr_size)
+                )
+        prims.barrier_cta_sync(self.sync_barrier_id, thread_count=threads)
```
PerkzZheng: seems that you are loading from gmem to smem and from smem to regiters in softmax task at the same time, without explicit warp-specialization pipelines. This is confusing for TS, is it possible to use extra warps to implement this otherwise it could easily cause data races/dead locks without TS validation support.

PerkzZheng (commented):
It seems that this sage attention implementation is coupled with block sparse attention as context kernels (2 num instances Q) haven't been modified. Do you expect to add sage attention support for context kernels ? I am also thinking that we probably want to fuse context and decode kernels, but it might need too many changes and validations. Feel free to give it a try if you think it is worthy.

heyuhhh:
> It seems that this sage attention implementation is coupled with block sparse attention as context kernels (2 num instances Q) haven't been modified. Do you expect to add sage attention support for context kernels ? I am also thinking that we probably want to fuse context and decode kernels, but it might need too many changes and validations. Feel free to give it a try if you think it is worthy.

Thanks for the comment, Perkz. Actually, the sage attention is based on the decode kernels so the context kernel can't support it for now. Because this PR mainly focuses on block sparse + sage attention.

Regarding to the divergency of context and decode kernel, i'm totally agree with you. Now for decode kernels it can only support paged kv and small tile size q. The block sparse attention is based on it and adds a continous kv wrapper. However, if we need larger tile size q we have to use context kernels or some optimizations in them, but it can't be directly supported.

However, i think it's not easy to unify these two type of kernels for now. Maybe i can have a try later when there is more bw released

heyuhhh:
Thanks @qsang-nv , both points are right.

1. SMEM check: confirmed. With an explicit `kv_stages=6` on the INT8 KV256 recipe the static pipeline check passes (16 + 6 x 32 = 208 KiB <= 218) while the built schedule needs 232 KiB (k16) / 240 KiB (k1) against the 227 KiB SM100 capacity; the default four-stage rings sit at 168 / 176 KiB. The public wrappers do not expose `kv_stages`, so only direct kernel-config callers could hit it, but the gap was real. Fix: the schedule builder compares the complete layout (metadata, Sage scale rings, tail exchange, barriers) against the SM capacity for every configuration and raises with the stage counts; the static KV256 check is documented as a pipeline lower bound; a test drives the INT8 recipe with a six-stage ring through the new check for the 16-token and the one-token K block. The 218 KiB budget stays a planning bound (the 16-bit default itself is 221.7 KiB), so the hard check uses the hardware capacity.

2. README figures: rewritten in the style of the QToken interface section, describing where the Sage kernels' time goes and what the 8-bit tiles buy, with the magnitude of the gain stated roughly and the benchmark harness named for current figures; the mixed-geometry note follows the same style.

Commits `cd89f6676` and `56231255b` on the branch.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/tmem_s.py
heyuhhh: This is needed in SOL attention, which composes of block sparse + compensated part. The later usually comes from the meaning pooling over the block size. So the K matrix in sage attention is by token group, but the compensated matrix should have their own flexibility. For example, k_block_size=16 but k_summary_block_size=1, so it's a mixed version we should resolve it.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_config.py
heyuhhh: It's used in [here](https://github.com/flashinfer-ai/flashinfer/pull/5123/changes/BASE..e9ff7813e45e83c9eb8ce197befb45b874b8ec61#diff-0617b614a08a71d1c438f21d299c9d2b05514c09845d65ea684bbd90d4557558R2634) which is in kv_tile_size=256 branch, and we have a guard for dtypes

PerkzZheng:
> > It seems that this sage attention implementation is coupled with block sparse attention as context kernels (2 num instances Q) haven't been modified. Do you expect to add sage attention support for context kernels ? I am also thinking that we probably want to fuse context and decode kernels, but it might need too many changes and validations. Feel free to give it a try if you think it is worthy.
> 
> Thanks for the comment, Perkz. Actually, the sage attention is based on the decode kernels so the context kernel can't support it for now. Because this PR mainly focuses on block sparse + sage attention.
> 
> Regarding to the divergency of context and decode kernel, i'm totally agree with you. Now for decode kernels it can only support paged kv and small tile size q. The block sparse attention is based on it and adds a continous kv wrapper. However, if we need larger tile size q we have to use context kernels or some optimizations in them, but it can't be directly supported.
> 
> However, i think it's not easy to unify these two type of kernels for now. Maybe i can have a try later when there is more bw released

No worries. we can support sage attention for context kernels later. It might need fewer changes than merging context into decode, and it is more flexible.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/helpers_common.py
heyuhhh: Got it. Thanks for the reminder. I'll guard this

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/tmem_corr.py
heyuhhh: Agreed. Actually, this comes from the refactor of `cfg.use_fp8_qkv`. However, we can have better ways for cleaner code

tests/attention/test_attention_ts_decode.py
heyuhhh: Right. I don't spend much time on tests' clean-up. I'll give a round of clean-up especially for tests in later commits. Thanks!

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/tmem_s.py
heyuhhh: Seems that we can remove this, which introduced by some historic changes.

heyuhhh:
> No worries. we can support sage attention for context kernels later. It might need fewer changes than merging context into decode, and it is more flexible.

Yes, and it's much simpler because there is no block sparse variants.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/sage_scales.py
heyuhhh: Right, thanks for the reminder. Let me think about it and see how to refactor this part

qsang-nv:
Thanks, both points are addressed. The capacity check runs ungated on the complete layout, ahead of `smem_allocator.allocate()` and the `skip_validation=True` TaskManager construction, and compares against `get_smem_capacity_in_bytes("sm_100")` rather than the 218 KiB planning budget; the parametrized rejection test pins exactly the gap. The README rewrite drops the revision-bound ratios.

One non-blocking note: the new README text ends with "`benchmarks/bench_primts_proxy_block_sparse.py` measures the current tree", but that file is not in the repo, and the PR description states "the harness is not part of this PR".

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/tmem_s.py
PerkzZheng: thanks. Let use add comments then.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/helpers_common.py
heyuhhh: Guard this in commit 7d7dd96e243b724d7a954b70de4bb2a6bfefc539. Thanks!

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/helpers_common.py
heyuhhh: Updated in commit b3bad4844c6c7756ff72a07313c51a4f98384ae5. Thanks!

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_kernel.py
```diff
@@ -2925,6 +3068,14 @@ def decode_gen_kernel(
                 tma_desc_v_summary=tma_desc_v_summary,
                 tma_desc_k_summary_atom=tma_desc_k_summary_atom,
                 tma_desc_v_summary_atom=tma_desc_v_summary_atom,
+                g_q_scale=g_q_scale,
```
PerkzZheng: should we add sage_attention as prefix to make it easier to read. Similar for other parameters.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/tmem_s.py
heyuhhh: Removed in commit 74858b6bede35386ba96b0715558f5e61867f71e. Thanks!

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/tmem_s.py
heyuhhh: It's indeed easy to confuse people. I have added some comments in commit 420cf095a19c0c23a99ca6ba29389e52b849e66b. You can double check this to see whether it's clear. Thanks

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/tmem_corr.py
heyuhhh: I added `qk_mma_dtype` and `pv_mma_dtype` in fmha_decode_config.py in commit 554a3195c7e7eaa24e951ffd11dea95211e300fc.  Also replace `use_8bit_qkv` with `k_dtype_bytes==1 and v_dtype_bytes==1`. I think it's cleaner now.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_resources/sage_scales.py
heyuhhh: commit 4af9a4ae8eb2e3a7d7bc19d45f4619742ecf7547 added SageKScales resource and commit 15e7960412d3549e42bd4a40ded171e927c23ea6 added SageVScales resource (both in the file `.../fmha_decode_resources/sage_scales.py`. Now the sage scales live in the warp-specilization pipelines explicitly.

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_tasks.py
```diff
@@ -2875,6 +2893,257 @@ def mma_schedule(
...
+                sparse_token_word3,
+            ) = sparse_softmax_metadata.load_route()
+            if cutlass.const_expr(cfg.use_sage_attention):
+                # The route's ``sfK`` words leave the held metadata stage for
+                # the task's own scale resources: ProdAcquire/ProdWork/
+                # ProdCommit fill and publish an SMEM ring slot, ConsWait
+                # makes it visible to both passes, ConsWork routes the lane's
+                # register words. A mixed-geometry plan fills the resource of
+                # the route's kind; the other one routes ones.
+                for scales in sage_ring_scales:
```
PerkzZheng: what does ring scales mean here ? it means we will prefetch multiple tiles's scale ?

heyuhhh: There is no prefetch. It means the sage scales which need smem buffers. Different sage block size will use different strategies some in registers and some in smem. The name will confuse people and i will refine this later, thanks

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_kernel.py
heyuhhh: Thanks, refined in commit cd09a7cec72c1a8826f9c833197ff9b8a1820c97

flashinfer/attention/prims_ts/kernels/fmha_decode/fmha_decode_tasks.py
heyuhhh: Renamed to `sage_smem_scales` in commit cd09a7cec72c1a8826f9c833197ff9b8a1820c97

heyuhhh:
Thanks @PerkzZheng @qsang-nv , i have resolved all of the comments and refined this PR (by removing some redundant tests, validations and docs). Could you please take a look again? Thanks!

PerkzZheng (approved):

qsang-nv:
Two gaps from the recent condensation, plus one question.

**1. Sage no longer rejects split-KV on KV256, and the split paths drop the V scales.**

`validate_sage_profile()` lost the direct-output-grid requirement (`_require_direct_output_grid` is gone), and the generic check does not replace it: the KV256 branch of `supports_grouped_keeps` computes `direct = not (self.use_split_kv or self.use_separate_reduction_kernel)` but still returns `True` for a split launch with `splits_kv > 1` and `mask_type == DENSE`.

No split flavor restores the V scales. `_apply_sage_channel_scales` is reached only from `_store_kv_tile_256_direct_fragment` (`tmem_corr.py:2522`) and `tmem_corr.py:3566`; the partial writer `_store_kv_tile_256_partial_fragment` (`:2538`) applies `_split_partial_scale` alone, the fused reduction's final store `_store_softmax_normalized_o_vec8` (`:1089`) does plain normalization with no Sage handling, and `reduction.py` never mentions `v_scale` or `v_mean`.

For a low-level caller configuring FP8 Q/K/V, FP16 out, Q64/KV256, D128, Sage k16 and a two-way GMEM split, the accepted path would omit `v_scale` and the `v_mean` restore — statically, V ≡ 1 with `v_scale = 2` and no mean would lack the expected ×2. Public wrappers cannot reach this (`_block_sparse/config.py` pins `split_kv=False`).

**2. The Sage scale resources are wired into tasks but absent from the dependency graph.**

`fmha_decode_tasks.py:3214/3218` and `:3320` put `sage_scales` into the Softmax tasks' `src_resources`/`dst_resources` (Correction takes `sageVScales`), but no Sage resource appears in `resource_dependency_graph` in `fmha_decode_kernel.py`. In nvidia-cutlass-dsl 4.7.x, `TaskManager._verify_resource_coverage()` raises `ValueError` for exactly this, and `print_and_verify()` documents that `skip_validation=True` turns each check into a warning.

This is a structural-validation gap, not a demonstrated runtime race: the JIT path downgrades it to a warning, and the three strict `skip_validation=False` sites in `tests/attention/test_attention_ts_decode.py` (2722, 3588, 3931) all use non-Sage configs, so no strict test catches it as a failure.

**3.** `main` now carries #5442 (`cake_sage`, SM100/SM103 Sage-FP8 block-sparse attention with a fused quantizer), reachable as `bsa_attn_sm100_blk64_fwd(..., backend="cake")`, and this branch has rebased onto it. The two differ in dtype coverage, exact/proxy support, hardware range, and scale layout / quantizer compatibility. How do the two paths divide, and for overlapping workloads what should guide a user's choice?

heyuhhh:
Thanks @qsang-nv for catching these. Both gaps are fixed in bcd02cf63. 

For cake sage attention, there is some difference between them. For example, cake sage only supports fp8 qk and primsts can support int8/fp8 qk. The scale layout is also different: primsts layout follows trtllm-gen which will serve trtllm attention firstly. One other important difference is that primsts sage attention can support different sparse block size or sage block size and it can be easier extend to many other applications, such as SOL attention.

Apart from the functional difference, I measured both on B200 with the same quantized inputs (outputs agree to 0.033 relative L2; each is about 0.022 from an FP32 reference)  under some overlapped cases (FP8, 64-token blocks, exact routes, per-token Q and 16-token K scales). The results show that primsts sage can have a better performance:

| Shape | Selected KV blocks | PrimTS / cake |
  | --- | --- | ---: |
  | B8 H32 S4096 | 10% / 50% / 90% | 0.78 / 0.68 / 0.81 |
  | B1 H8 S4096 | 10% / 50% / 90% | 0.84 / 0.77 / 0.84 |
  | B1 H40 S15360 | 12.5% | 0.70 |
  | B1 H40 S75600 | 10% | 0.80 |

**Guidance for users**: cake is the natural choice when FlashInfer should quantize (its fused quantizer) or when V scales are per batch ([B, Hkv, D], which this path does not take). This path fits callers that already hold sageQuant scales (TensorRT-LLM), need INT8 Q/K, proxy routes, dense attention, other block sizes or v_mean. On the shared FP8 exact workload the measurements above favor this path.

Issue #261, opened 2024-05-27
build raise  "cub::BlockAdjacentDifference<__nv_bool, 1024, 1, 1, 860>" has no member "SubtractLeft"


WanBenLe:
[1/1] /usr/local/cuda/bin/nvcc --generate-dependencies-with-compile --dependency-output /root/autodl-tmp/wsl/flashinfer/python/build/temp.linux-x86_64-cpython-310/csrc/sampling.o.d -I/root/autodl-tmp/wsl/flashinfer/python/include -I/root/autodl-tmp/conda/envs/wslconda/lib/python3.10/site-packages/torch/include -I/root/autodl-tmp/conda/envs/wslconda/lib/python3.10/site-packages/torch/include/torch/csrc/api/include -I/root/autodl-tmp/conda/envs/wslconda/lib/python3.10/site-packages/torch/include/TH -I/root/autodl-tmp/conda/envs/wslconda/lib/python3.10/site-packages/torch/include/THC -I/usr/local/cuda/include -I/root/autodl-tmp/conda/envs/wslconda/include/python3.10 -c -c /root/autodl-tmp/wsl/flashinfer/python/csrc/sampling.cu -o /root/autodl-tmp/wsl/flashinfer/python/build/temp.linux-x86_64-cpython-310/csrc/sampling.o --expt-relaxed-constexpr -DFLASHINFER_ENABLE_BF16 -DFLASHINFER_ENABLE_FP8 --compiler-options ''"'"'-fPIC'"'"'' -O3 -std=c++17 --threads 8 -Xfatbin -compress-all -DTORCH_API_INCLUDE_EXTENSION_H '-DPYBIND11_COMPILER_TYPE="_gcc"' '-DPYBIND11_STDLIB="_libstdcpp"' '-DPYBIND11_BUILD_ABI="_cxxabi1011"' -DTORCH_EXTENSION_NAME=_kernels -D_GLIBCXX_USE_CXX11_ABI=0 -gencode=arch=compute_86,code=compute_86 -gencode=arch=compute_86,code=sm_86
        FAILED: /root/autodl-tmp/wsl/flashinfer/python/build/temp.linux-x86_64-cpython-310/csrc/sampling.o
[...]

PR #5124, opened 2026-09-11
fix(sampling): remove CUDA 11.8 FlagHeads fallback

yonglai:
## 📌 Description

The FlagHeads branch in DeviceSamplingFromProb was added in #265 to support CUDA 11.8, whose bundled CUB predates SubtractLeft. FlashInfer's minimum supported CUDA is now 12.9 and CI covers cu129/cu130 only, so the condition the fallback was written for can no longer occur.

The guard selecting it is also keyed on the wrong thing: __CUDACC_VER_MAJOR__/MINOR describes the compiler, but the availability of SubtractLeft vs FlagHeads is a property of the CUB headers on the include path. FlashInfer bundles its own CCCL and adds it with -I while the CUDA toolkit's own headers go in with -isystem (jit/cpp_ext.py:build_common_cflags), so the vendored copy deliberately takes precedence over the CTK-bundled one — and the compiler version therefore tells you nothing about which CUB you are compiling against. On nvcc < 12.1 the macro goes undefined and FlagHeads is emitted against bundled CCCL, where the public overloads were removed in CCCL 3.0, producing a compile failure that names neither the real cause nor the unsupported toolkit.

SubtractLeft is present in every CUB tested here, down to 2.0.1 -- well below the CCCL version shipped with CUDA 12.9, the minimum supported toolkit. Removing the fallback deletes code unreachable on every supported configuration and additionally lets nvcc 12.0 compile rather than fail.

### Verification

One source file (`probe.cu`, below), API branch selected explicitly via `-D`, only the CUB headers on the include path varied:

| nvcc | CUB / CCCL headers | Source | `FlagHeads` | `SubtractLeft` |
|---|---|---|---|---|
| 12.0.140 | CUB 2.0.1 | Ubuntu `libcub-dev`, `/usr/include` | pass (deprecated) | pass |
| 12.0.140 | CCCL 3.0.0 | `NVIDIA/cccl` tag `v3.0.0` | **fail** | pass |
| 12.0.140 | CCCL 3.0.3 | `NVIDIA/cccl` tag `v3.0.3` | **fail** | pass |
| 12.0.140 | CCCL 3.3.2 | wheel, `flashinfer-python` 0.6.18.post1 | **fail** | pass |
| 13.3.73 | CCCL 3.3.2 | wheel, `flashinfer-python` 0.6.18.post1 | **fail** | pass |
| 12.0.140 | CCCL 3.6.0 | `3rdparty/cccl` submodule at `main` | **fail** | pass |

`FlagHeads` is deprecated in CUB 2.0.1 and absent from CCCL 3.0.0 onward; `SubtractLeft` compiles across the entire range, including CUB 2.0.1 — older than anything reachable under the CUDA 12.9 floor. In CCCL 3.6.0 a `FlagHeads` symbol survives at `block_adjacent_difference.cuh:155`, but it is a private helper taking `int linear_tid` and is not the public overload this code calls, so the fallback cannot be repaired — only removed.

Error signature (nvcc 12.0.140, bundled CCCL 3.6.0, before this change):

```
include/flashinfer/sampling.cuh(623): error:
class "cub::_V_300600_SM_860::BlockAdjacentDifference<__nv_bool, 128, 1, 1>"
has no member "FlagHeads"
    detected during instantiation of
    "void flashinfer::sampling::DeviceSamplingFromProb<...>"
```

After this change, `sampling.cuh` compiles under both nvcc 12.0.140 and 13.3.73 against bundled CCCL 3.6.0.

<details>
<summary><code>probe.cu</code> — minimal reproduction</summary>

```cuda
#include <cub/cub.cuh>

struct BoolDiffOp {
  __device__ __forceinline__ bool operator()(const bool &a, const bool &b) const {
    return a != b;
  }
};

constexpr int BLOCK_THREADS = 128;
constexpr int VEC_SIZE = 4;

__global__ void probe(bool *out) {
  using BAD = cub::BlockAdjacentDifference<bool, BLOCK_THREADS>;
  __shared__ typename BAD::TempStorage tmp;
  bool in[VEC_SIZE] = {true, false, true, false};
  bool diff[VEC_SIZE];
#ifdef USE_SUBTRACTLEFT
  BAD(tmp).SubtractLeft<VEC_SIZE>(in, diff, BoolDiffOp());
#else
  BAD(tmp).template FlagHeads<VEC_SIZE>(diff, in, BoolDiffOp(), 0);
#endif
  out[threadIdx.x] = diff[0];
}
```

Compile with and without `-DUSE_SUBTRACTLEFT`, varying `-I` to select the header set:

```bash
nvcc -ccbin g++-12 -arch=sm_86 $INC -c probe.cu -o /dev/null
nvcc -ccbin g++-12 -arch=sm_86 $INC -c probe.cu -o /dev/null -DUSE_SUBTRACTLEFT
```

</details>

## 🔍 Related Issues

- #265 introduced the `FlagHeads` fallback in May 2024 to support CUDA 11.8,   closing #261 (the same `has no member` compile failure, in the opposite   direction: `SubtractLeft` missing on CUB 1.x). #410 was a follow-up cu118 CUB fix.
- #4763 reports the same failure and fixes it by preferring system CUB/Thrust   over the bundled CCCL in `env.py`. This PR fixes the version predicate instead,   so the two are alternatives rather than complements — deferring to maintainers   on which is preferable. I raised the root cause there before opening this.

One clarification for anyone comparing the two reports: #4763's description  says "bundled CCCL 3.0.3", but the pasted error names `cub::_V_300302_SM_890`.  Per the `MMMmmmpp` encoding documented in `cub/version.cuh`, `300302` decodes  as 3.3.2 — the same version reported here, not 3.0.3. (For contrast,  `_V_300003` in the table above really is 3.0.3.)

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

This is a compile-time change; no new unit test would be meaningful. Existing `tests/test_sampling.py` coverage exercises the affected code path unchanged. CI runs cu129/cu130 only, so no tested configuration takes the removed branch — a green run confirms no regression on supported versions.

## 🔬 Experimental Track

```experimental-tests
#
#
```

## Reviewer Notes

If a version guard is preferred over removal, the correct predicate is `CUB_VERSION` / `CCCL_VERSION` rather than `__CUDACC_VER_*` — but since every CUB reachable under the 12.9 floor has `SubtractLeft`, any such guard would be unconditionally true. Happy to switch approach if maintainers prefer.

## Summary by CodeRabbit

* **Refactor**
  * Standardized sampling behavior by consistently using the same adjacent-difference operation across supported CUDA environments.

### DIFF
--- a/include/flashinfer/sampling.cuh
+++ b/include/flashinfer/sampling.cuh
@@ -109,10 +109,6 @@ constexpr BlockReduceAlgorithm REDUCE_ALGO = BLOCK_REDUCE_WARP_REDUCTIONS;
 #define FLASHINFER_SAMPLING_LAUNCH_BOUNDS(block_threads)
 #endif
 
-#if (__CUDACC_VER_MAJOR__ * 10000 + __CUDACC_VER_MINOR__ * 100 >= 120100)
-#define FLASHINFER_CUB_SUBTRACTLEFT_DEFINED
-#endif
-
 template <typename T>
 struct ValueCount {
   T value;
@@ -615,13 +611,8 @@ __device__ __forceinline__ void DeviceSamplingFromProb(
     }
 
     bool greater_than_u_diff[VEC_SIZE];
-#ifdef FLASHINFER_CUB_SUBTRACTLEFT_DEFINED
     BlockAdjacentDifference<bool, BLOCK_THREADS>(temp_storage->block_prim.adj_diff)
         .SubtractLeft<VEC_SIZE>(greater_than_u, greater_than_u_diff, BoolDiffOp());
-#else
-    BlockAdjacentDifference<bool, BLOCK_THREADS>(temp_storage->block_prim.adj_diff)
-        .template FlagHeads<VEC_SIZE>(greater_than_u_diff, greater_than_u, BoolDiffOp(), 0);
-#endif
     __syncthreads();
 
 #pragma unroll


yonglai:
First PR here, so I don't have permission to start CI. Could a maintainer run @flashinfer-bot run when convenient? Happy to rebase first if preferred.

yyihuang:
@flashinfer-bot run

yonglai:
Thanks @yyihuang for the CI run. All checks passed (20 successful, 2 skipped). Ready for review whenever someone has time.

PR #5125, opened 2026-09-11
fix(cute-dsl): use has_side_effects to prevent wrong llvm optimisation

namonakimono:
PLEASE ANALYSE THE ATTACHED ZIP FILES, WHICH IS SELF-CONTAINED. (You can ask an agent to do it.)

[flashinfer-inline-asm-evidence.zip](https://github.com/user-attachments/files/32085057/flashinfer-inline-asm-evidence.zip)

# [CuTeDSL] Shared-load inline-asm helpers omit effects, allowing stale-load reuse

`ld_shared_v2_u32` and `ld_shared_v4_u32` read shared memory, but both emit
`llvm.inline_asm` with `has_side_effects=False`, register-only inputs, and no
memory clobber. I can demonstrate elimination of a required second read using
the actual FlashInfer helpers with NVIDIA's public CuTeDSL bindings.

This report concerns the helpers' compiler-visible contract. It does not claim
that NVIDIA's normal CuTeDSL JIT currently fails the full FlashInfer test, or
that LLVM is violating the contract it receives.

## Affected source and connection to FlashInfer

Verified at FlashInfer v0.6.18, commit
[`69ff11fc4954396d98326656dc85debd2223f637`](https://github.com/flashinfer-ai/flashinfer/blob/69ff11fc4954396d98326656dc85debd2223f637/flashinfer/cute_dsl/fp4_common.py#L305-L359).
A read-only check of public `main` at
[`40d7876acaf33f3895088cff92698fe2d5c094fe`](https://github.com/flashinfer-ai/flashinfer/blob/40d7876acaf33f3895088cff92698fe2d5c094fe/flashinfer/cute_dsl/fp4_common.py#L305-L359)
found the identical helper file. Runtime results below use the pinned v0.6.18.

For example, `ld_shared_v4_u32` emits:

```python
"ld.shared.v4.u32 {$0, $1, $2, $3}, [$4];",
"=r,=r,=r,=r,r",
has_side_effects=False,
```

This is used by the [NVFP4 TMA quantization consumer](https://github.com/flashinfer-ai/flashinfer/blob/69ff11fc4954396d98326656dc85debd2223f637/flashinfer/quantization/kernels/nvfp4_quantize.py#L1300-L1343).
When a pipeline stage is reused, the address may repeat while its contents change.

The attached evidence repository's `emit_from_flashinfer.py` **calls the imported
FlashInfer helpers**, using NVIDIA CuTeDSL 4.5.2. It constructs only the small
surrounding LLVM function and records source hashes. The helper file SHA-256 is
`a430b3171c7c972a2b98a176e5a47ddcaf36ac71e6231420e961e269d0d045d1`, matching the
pinned source. No downstream compiler or workaround pass is involved in emission.
The `.ll` files are obtained by translating this LLVM-dialect MLIR; they are not
claimed to be NVIDIA full-JIT LLVM IR dumps.

## Minimal reproduction

The test performs:

```text
write A to shared address p
first = FlashInfer.ld_shared_v4_u32(p)
write B to p
second = FlashInfer.ld_shared_v4_u32(p)
return both snapshots
```

Both stores explicitly declare side effects and a memory clobber. The numerical
check uses one GPU thread and 16-byte-aligned shared storage, removing TMA and
inter-thread synchronization as confounders.

With the attached repository and the pinned Python packages:

```bash
python -m pytest tests -q            # 2 failures on unpatched v0.6.18
python -m pytest tests -q --patched  # 2 passes with the source flag changed
```

These tests need no GPU or separate LLVM build: they use NVIDIA's own bundled
`builtin.module(cse)` pass. The baseline emits two loads and CSE leaves one.
Changing only `has_side_effects` in the two helper definitions preserves both.
`--patched` applies this source change only in the test process.

A second route translates the same original MLIR **without first running CSE**:

```bash
mlir-translate --mlir-to-llvmir v4.original.mlir -o before.ll
opt -O2 -S before.ll -o after.ll
llc -O3 -mcpu=sm_80 after.ll -o after.ptx
```

In the recorded LLVM 22.1.6 run, the two calls become one in `GVNPass`, and the
PTX contains one `ld.shared.v4.u32`. Direct `llc -O3` on `before.ll` also emits
one load. Version output, binary hashes, pass traces, and both input/output IR
are included; these CLI binaries are an existing source build, not a certified
official binary distribution. The NVIDIA-only CSE test is independent of them.

On B200, the numerical v4 result is:

```text
Expected: [11, 12, 13, 14, 101, 102, 103, 104]
Original: [11, 12, 13, 14,  11,  12,  13,  14]
Patched:  [11, 12, 13, 14, 101, 102, 103, 104]
```

The v2 helper exhibits the same issue. All returned components are observed, so
this is not deletion of an unused result.

## Proposed change

Set `has_side_effects=True` in both shared-load helpers. The attached patch changes
only those two lines, and both NVIDIA CSE regression tests pass with that change.

We also checked adding `~{memory}` alone: it preserves both loads on the direct
LLVM route, but the tested MLIR CSE still merges them. Thus a memory-clobber-only
change does not cover both layers. This proposal does not replace barriers or
claim to audit every ordering requirement; it ensures these hidden memory reads
are not treated as effect-free operations in the tested paths.

See [LLVM's inline-asm contract](https://llvm.org/docs/LangRef.html#inline-assembler-expressions)
and [the MLIR operation](https://mlir.llvm.org/docs/Dialects/LLVM/#llvminline_asm-llvminlineasmop).
The address is passed in an ordinary register, so the PTX spelling of a load
cannot by itself communicate the memory dependency to generic optimizations.

## Why existing NVIDIA runs can pass

A passing normal JIT run is useful baseline evidence, but it does not test every
optimization shape. The recovered original TMA case retained a consumer loop in
NVIDIA PTX; its four static load sites execute again on subsequent iterations.
In the alternate pipeline, unrolling exposed repeated stage addresses. Archived
artifacts show 4 input LLVM load sites, 32 after `opt`, and 16 in final PTX. A
memory-aware control has 4 → 32 → 32.

That historical elimination occurred during code generation, whereas this new,
straight-line reproducer also triggers MLIR CSE and LLVM GVN independently. The
repository labels these as separate experiments and includes sanitized historical
artifacts as supplementary evidence. The minimized test and source patch are the
primary basis for this report.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of shared-memory data loading by ensuring required load operations are preserved and executed in the correct order.


namonakimono:
@bkryu Hello, please review the two-line changes.

ChinoMars:
@bkryu @Aneureka would you please help review this PR?

Aneureka:
/bot run tests/utils/test_fp4_quantize.py tests/norm/test_add_rmsnorm_fp4_quant_cute_dsl.py tests/gemm/test_mm_bf16_fp4.py

Aneureka (approved):
LGTM.

PR #4951, opened 2026-09-04
feat(cake_sage): add SM120 block-sparse attention backend

yyihuang:
## Summary

- add generated SM120 kernels for prequantized Sage block-sparse attention
- add a manifest-driven JIT loader for six exported specializations
- expose the implementation through `bsa_attn_sm120_blk64_sage_fwd(..., backend="cake")`
- keep output storage and TMA descriptor workspace caller-owned

## Performance

Baseline: [SM120 Sage CuTe DSL implementation at `a14241f2`](https://github.com/flashinfer-ai/flashinfer/blob/a14241f2a82712ff58f571d6698202512e2661ef/flashinfer/cute_dsl/sparse/bsa_attn_sm120.py), introduced in [#4691](https://github.com/flashinfer-ai/flashinfer/pull/4691).

All latency values are milliseconds. `Cake / Export` above 1 means the export is faster than its production source; `Baseline / Export` above 1 means this PR is faster than the existing implementation.

The 13 `ut_*` rows are correctness-only timing sanity checks. They exercise ragged lengths, empty rows, block-size metadata modes, custom scale, and V permutation, but they are not part of the #4691 performance portfolio. Their source/export parity is shown without claiming a #4691 speedup.

| Correctness shape | Cake source | Export | Cake / Export |
|---|---:|---:|---:|
| `ut_aligned_full_b1_h2_s128` | 0.004992 | 0.004928 | 1.012987x |
| `ut_aligned_half_b1_h8_s512` | 0.006881 | 0.006784 | 1.014298x |
| `ut_aligned_sparse_b2_h4_s256` | 0.004097 | 0.004000 | 1.024250x |
| `ut_aligned_dense_b2_h8_s1024` | 0.019713 | 0.019616 | 1.004945x |
| `ut_custom_softmax_scale` | 0.005024 | 0.004545 | 1.105391x |
| `ut_ragged_q` | 0.004992 | 0.004704 | 1.061224x |
| `ut_ragged_k` | 0.005120 | 0.005088 | 1.006289x |
| `ut_ragged_qk` | 0.005152 | 0.005088 | 1.012579x |
| `ut_empty_first_row` | 0.006880 | 0.006912 | 0.995370x |
| `ut_block_sizes_global` | 0.005120 | 0.005184 | 0.987654x |
| `ut_block_sizes_batch` | 0.005152 | 0.005216 | 0.987730x |
| `ut_block_sizes_head` | 0.004992 | 0.005056 | 0.987342x |
[...]

PR #5127, opened 2026-09-11
fix(cake_sage_attn): cover Sage combinations and pass tensor maps by value

yyihuang:
Sage attention could fail to find an exported module for legal parameter combinations. The binding also assumed caller-owned TMA workspace contents remained valid: overwriting or recycling that storage could cause a CUDA fault or silently access the wrong tensors.

Expand the SM120 export from 7 to 25 specializations, covering omitted and uniform counts, every block-size layout, partial KV tiles, and empty selections. Pass all four tensor maps by value through `const __grid_constant__` kernel parameters. Each eager call and captured graph owns its descriptor values, so attention needs no descriptor workspace cache or upload kernel. Keep the existing Python `tma_descriptor_workspace` keyword accepted but unused; an empty tensor is sufficient.

Add 73 attention cases and six isolated descriptor regressions: overwritten workspace, stale valid descriptors, recycled views, interleaved graph replay and eager bindings, ordered streams, and empty workspace. Tests also verify that workspace contents remain unchanged. Document count semantics and captured tensor lifetimes.

Validation on RTX PRO 6000 Blackwell Server Edition (SM120):
- All 79 public GPU tests pass in 158.97s, covering all 25 specializations.
- The affected source GPU regression and fresh-binding graph capture test pass; all 40 contract correctness rows and eight unchanged performance rows pass.
- Separate Compute Sanitizer runs pass: synccheck 0 errors in 6.127s; racecheck 0 hazards in 6.575s. Each process has a 20s hard limit.

Three-arm, interleaved cold-L2 CUPTI measurements include the complete exported FFI GPU execution span. All three versions were measured on the same GPU in the same run; JIT, allocation, and reference computation are excluded.

| Case | Cached workspace baseline (µs) | Previous upload fix (µs) | Current grid-constant ABI (µs) |
| --- | ---: | ---: | ---: |
| B1 H2 S128 | 5.087 | 7.457 | 4.800 |
| Non-contiguous KV4000 | 25.984 | 28.001 | 25.312 |
| Non-contiguous KV4033 | 19.328 | 21.649 | 19.008 |
| B8 H32 S1024, density90 | 221.137 | 223.905 | 218.114 |

CUPTI records one, two, and one kernel activities per call respectively. The current ABI removes the extra upload launch and has lower median latency than both prior versions in these four measured cases. These results replace the earlier table for the upload-based implementation.

Follow-up to #5083.

## Summary by CodeRabbit

* **New Features**
  * Improved SM120 Sage block-sparse attention execution with more reliable tensor-map handling and CUDA device management.
  * Added support for per-row, uniform, omitted, empty, and contiguous block-count modes, including ragged sequence lengths.
  * CUDA Graph capture and replay now work more reliably; each captured launch preserves its own tensor descriptors.
  * The compatibility workspace is accepted but no longer required for descriptor storage.

* **Bug Fixes**
  * Strengthened validation for tensor layouts, strides, alignment, scalar ranges, and empty selections.

* **Tests**
  * Added coverage for workspace reuse, stale or empty workspace contents, stream ordering, and CUDA Graph replay.


yyihuang:
@triple-mu Updates requested by https://github.com/flashinfer-ai/flashinfer/pull/5083#issuecomment-5628415175 in https://github.com/flashinfer-ai/flashinfer/pull/5083

triple-mu:
Verified at `523920a5`. **This closes everything I raised on #5083** — all four call shapes from the original #4951 repro now resolve, on both aligned and unaligned lengths:

| call shape | `S=4096` | `S=4000` |
|---|---|---|
| `block_sizes=None` + nums | `(1,0,1,0,0)` | `(1,0,0,0,0)` |
| nums omitted, `uniform=True` | `(0,0,1,1,0)` | `(0,0,0,1,0)` |
| `block_sizes` + nums | `(1,1,0,0,0)` | `(1,1,0,0,0)` |
| `block_sizes` + `uniform=True` | `(0,1,0,1,0)` | `(0,1,0,1,0)` |

Only the top-left cell worked before. I also re-derived the reachable key set independently and it comes out to exactly the 25 you exported — no gaps, and no test case that resolves to an unexported key.

**The workspace fix holds.** Both repros from the #5083 thread are now clean:

| repro | before | now |
|---|---|---|
| clobber the workspace in place, call again | `illegal instruction` | output bit-identical to the first call |
| stale-but-valid descriptor, two live tensor sets | silently wrote the other set, `cosine=0.0000` | `cosine=1.0000`, `relative_L1=0.0000` |

Nice touch running each scenario in a child process — a faulted context would otherwise poison the rest of the session. `_zeroed_workspace` deliberately skipping the host sync is stricter than my version was.

**Numerics unchanged.** Independent FP32 reference, not rounded to BF16. Errors are bit-identical to what I measured on #5083, so the refresh didn't perturb anything:

| shape | max abs | cosine |
|---|---:|---:|
| `Sk=4000` ragged counts | 3.03e-3 | 0.999811 |
| `Sk=4033` ragged + empty rows | 2.84e-3 | 0.999811 |
| `Sk=65` (1-token tail) | 1.46e-3 | 0.999999 |
| workload A `S=37874` | 2.34e-4 | 0.999803 |
| workload C `S=82337` | 1.78e-4 | 0.999814 |

**Sanitizer**: I ran `memcheck` and `initcheck` alongside the synccheck/racecheck you reported, over a driver exercising six shapes through the new specialization — 0 errors each. Worth adding memcheck to the routine for this kernel: K isn't padded, so a tail over-read leaves no numerical signal at `atol=1e-2`.

**On the cost of the refresh** — your table and mine disagree, and I think it's a units question rather than a contradiction. Same inputs, same specialization, `#5083` tree vs `#5127` tree, CUDA event, median of 50 iters, 3 repeats:

| shape | #5083 | #5127 | delta |
|---|---:|---:|---:|
| B1 H2 Sq128 Sk128 | 8.192 us | 8.384 us | +0.19 us (+2.3%) |
| `Sk=4000` | 34.496 | 34.816 | +0.32 (+0.9%) |
| `Sk=4033` | 22.528 | 22.528 | 0 |
| workload A `S=37874` | 1981.4 | 2016.3 | +34.8 (within run-to-run spread) |
| workload C `S=82337` | 9406.5 | 9464.3 | +57.9 (+0.6%) |

You measure the GPU span with CUPTI, where the upload launch is a large fraction of a 5 us kernel; I measure the whole call with CUDA events, where it mostly disappears into the pipeline. Both are right — but from a caller's point of view the fixed cost is 0.2-0.3 us, not 43%, so I don't think the smallest-case number should worry anyone. (Different silicon too: RTX PRO 5000 Blackwell, 110 SMs, vs your PRO 6000.)

LGTM from my side.

Two things left over that are **not** this PR's to fix — both predate it, and I'll file one issue covering them:

- per-row `q2k_block_nums` values are still unrange-checked; a count several times `q2k_capacity` is accepted and indexes past the row
- `contiguous_block_indices` is still trusted rather than verified — lying about it returns successfully with `cosine=0.3202`

Both would be cheap behind the existing `FLASHINFER_VALIDATE_INPUTS` switch, zero cost on the normal path.

<details><summary>Environment</summary>

NVIDIA RTX PRO 5000 Blackwell, SM120 / cc 12.0, 110 SMs, 73415 MiB, driver 580.95.05, PCI device `0x2BB310DE` (nvidia-smi reports the generic "NVIDIA Graphics Device" for this id). torch 2.13.0+cu132, kernels JIT-compiled with nvcc 13.3.73. 78/78 public GPU tests pass, 2m50s including a cold build of all 25 modules.

</details>

yyihuang:
@flashinfer-bot run

yzh119 (approved):

yyihuang:
/bot run tests/attention/test_cake_sage_block_sparse_attention.py tests/attention/test_cake_sage_tma_workspace.py

yyihuang:
Thanks @triple-mu for the thorough review and validation! Please feel free to assign me the follow-up issue for the two input-validation checks.

triple-mu:
Re-verified at `1af7b764`. My earlier LGTM was against `523920a5` and the upload-kernel approach, so treating this as a fresh review — **the grid-constant rewrite is better than what it replaced, and it's now faster than the pre-fix baseline.**

| check | result |
|---|---|
| test suite | 79 collected / 79 passed, 179 s including a cold build of all 25 modules |
| specialization coverage | 25 exported = 25 reachable, no gaps |
| independent FP32 numerics | 6 cases + my two production shapes pass; errors bit-identical to both earlier revisions |
| `memcheck` / `initcheck` / `synccheck` / `racecheck` | 0 errors each |

**The workspace really is inert now.** Six degenerate inputs produce bit-identical output — conventional aligned uint8, an empty tensor, a deliberately misaligned single byte, a float32 tensor, a CPU tensor, and `None`:

```
all six:  max_abs=6.6323e-04  rel_L1=1.9433e-02  cos=0.99980700
```

So the "accepted and unused, an empty tensor is sufficient" wording holds.

**Latency, caller's view.** CUDA event, median of 50 iters, 3 repeats, merged `#5083` tree vs this one:

| shape | #5083 | this PR | delta |
|---|---:|---:|---:|
| B1 H2 Sq128 Sk128 | 8.192 us | 8.192 us | 0 (below the event quantum here) |
| `Sk=4000` | 34.288 | 32.768 | **-1.52 us (-4.4%)** |
| `Sk=4033` | 22.528 | 21.200 | **-1.33 us (-5.9%)** |
| workload A `S=37874` | 1969.0 | 1985.5 | +16.5 (within spread; mins are 1864.9 vs 1862.9) |
| workload C `S=82337` | 9360.4 | 9410.9 | +50.5 (+0.5%) |

Same direction as your CUPTI table. Against the upload-kernel revision I measured earlier (`Sk=4000` 34.816, `Sk=4033` 22.528), this one is ~2 us and ~1.3 us faster. Dropping the `acquire` fences is right — with by-value parameters there's no generic-proxy write to synchronize against — and `&Q_map` for `prefetch.tensormap` and the TMA operands matches what the programming guide shows for grid-constant tensor maps.

Still LGTM, more confidently than before.

One small thing on the compatibility parameter: `tma_descriptor_workspace` is still a **required** keyword-only argument annotated `torch.Tensor`, but passing `None` now succeeds — the annotation no longer describes the behaviour, and there's no deprecation path for eventually removing it. Defaulting it to `None` with a `DeprecationWarning` when a caller does pass something would let new code stop threading a dummy tensor through, and give you a clean removal later.

And a process suggestion rather than a review comment: this is the third revision validated with synccheck and racecheck only. For this kernel specifically memcheck is the one that earns its keep — K isn't padded, so a tail over-read produces no numerical signal at `atol=1e-2` and neither of the other two tools would see it. Might be worth adding to the standard sanitizer set for `cake_sage_*`.

The two input-validation gaps are unchanged and still out of scope here; I'll carry them into the follow-up issue.

<details><summary>Environment</summary>

NVIDIA RTX PRO 5000 Blackwell, SM120 / cc 12.0, 110 SMs, driver 580.95.05. torch 2.13.0+cu132, nvcc 13.3.73.

</details>

yyihuang:
@flashinfer-bot run

PR #5133, opened 2026-09-11
feat(prefill): let backend="auto" reach cuDNN/CUTLASS on Blackwell ragged prefill (up to 3.3x)

Anerudhan:
## 📌 Description

On Blackwell, ragged prefill with `backend="auto"` resolves to **FA2 unconditionally**. `determine_attention_backend` can only answer `"fa3"` or `"fa2"`, and FA3 is gated on `is_sm90a_supported()`, which is `False` on SM100:

```python
>>> determine_attention_backend(cuda, NONE, False, False, bf16, bf16, head_dim_qk=128, head_dim_vo=128)
'fa2'
>>> is_sm90a_supported(cuda), is_sm100a_supported(cuda)
(False, True)
```

So `auto` misses not just cuDNN but *every* Blackwell-capable backend. FA2 has no tcgen05 path and measures at ~320–390 TFLOPS, roughly 17% of B200 bf16 peak.

This adds a post-resolution upgrade in `BatchPrefillWithRaggedKVCacheWrapper.plan()`, modelled on the `fmha_v2` SM120 precedent directly above it: `auto` tries cuDNN, then CUTLASS, then stays on FA2. It is **not** implemented by teaching `determine_attention_backend` a new answer — that function is shared with decode, paged prefill and sparse, none of which accept `"cutlass"`, so a new return value would have to be filtered back out at four other call sites.

cuDNN is chosen only when the installed cuDNN can consume the caller's **token-unit** indptrs directly, which keeps the caller contract byte-identical to every other backend.

The final commit goes further and **normalizes the explicit `backend="cudnn"` path to token-unit `qo_indptr` as well**, so every backend in this wrapper now takes the same thing. That is a breaking change for existing explicit-cuDNN callers, who previously passed element-unit offsets (`cumsum(seq_lens) * num_qo_heads * head_dim_qk`); the wrapper detects that shape and raises a message naming the new convention rather than computing something wrong. `benchmarks/routines/attention.py` and the three `test_cudnn_prefill*` files are updated to match.

**The exclusions are load-bearing.** Both candidate run paths receive only `causal` and the scales — never `window_left`, `logits_soft_cap`, a custom mask, or the multi-item-scoring pointers. Upgrading while any of those were requested would silently drop them and return quietly wrong numbers, so each is an explicit disqualifier.

### Performance

All numbers below were re-taken on one standardised stack so the two parts are comparable:
**CUDA 13.4**, **cuDNN frontend 1.29.0 / backend 9.26.0 (92600)**, bf16, causal, ragged prefill,
median of 20 iterations after 5 warmup, via `benchmarks/flashinfer_benchmark.py`.

- **B200** — cc 10.0, torch 2.14.0+cu130, nvidia-cutlass-dsl 4.7.1
- **B300 (GB300)** — cc 10.3, torch 2.13.0a0 (NGC container), same CUDA 13.4 / cuDNN 1.29 + 92600 overlay

Every backend is measured at every shape, so CUTLASS and cuDNN can be compared directly rather
than only through `auto`'s pick. **Correctness:** all 16 shapes below were re-run with
`--refcheck` and without `--allow_output_mismatch`; FA2 is the reference and every cuDNN and
CUTLASS output matched it elementwise within tolerance (any difference raises). `auto` was FA2 in **every** row before this PR, so the
"cudnn vs fa2" column is exactly what `auto` gains.

#### By head dimension

The `d192/128` and `d256/256` rows use GLM-5.3's attention shape: MLA with
`qk_nope 192` + `qk_rope 64` → `qk_head_dim 256`, `v_head_dim 256`, **64 heads**.

| shape | fa2 | cutlass | cudnn | cudnn vs fa2 | `auto` picks |
|---|---|---|---|---|---|
| **B200 (cc 10.0)** | | | | | |
| b16 s4096 64×8 d128 | 12.277 ms | 5.752 | **4.006** | **3.06×** | cudnn |
| b4 s16384 64×8 d128 | 46.763 | 19.647 | **17.170** | **2.72×** | cudnn |
| b64 s1024 32×8 d128 | 1.776 | 1.440 | **0.733** | **2.42×** | cudnn |
| b16 s4096 64×64 d192/128 | 14.808 | 7.239 | **6.005** | **2.47×** | cudnn |
| b64 s1024 64×64 d192/128 | 4.471 | 3.139 | **1.970** | **2.27×** | cudnn |
| b16 s2048 64×64 **d256/256** | 7.016 | *declined* | **2.287** | **3.07×** | cudnn |
| b8 s4096 64×64 **d256/256** | 13.346 | *declined* | **3.919** | **3.41×** | cudnn |
| **B300 / GB300 (cc 10.3)** | | | | | |
| b16 s4096 64×8 d128 | 10.128 | 4.588 | **2.499** | **4.05×** | cudnn |
| b4 s16384 64×8 d128 | 38.942 | 12.892 | **9.443** | **4.12×** | cudnn |
| b64 s1024 32×8 d128 | 1.477 | 1.251 | **0.480** | **3.08×** | cudnn |
| b16 s4096 64×64 d192/128 | 12.119 | 5.501 | **4.002** | **3.03×** | cudnn |
| b64 s1024 64×64 d192/128 | 3.616 | 2.809 | **1.401** | **2.58×** | cudnn |
| b16 s2048 64×64 **d256/256** | 5.950 | *declined* | **1.492** | **3.99×** | cudnn |
| b8 s4096 64×64 **d256/256** | 11.357 | *declined* | **2.536** | **4.48×** | cudnn |

`d256/256` is new coverage: CUTLASS rejects it outright (*"CUTLASS backend requires head dimension
to be 128 or 192"*), so before this PR that shape had no Blackwell path at all and ran on FA2.
The gains are consistently larger on B300 than on B200 — cuDNN scales onto sm103 better than FA2 does.

#### GLM-5.3 attention across input / output sequence lengths

Same GLM-5.3 shape throughout (`d_qk 256`, `d_vo 256`, 64 heads, batch 4), sweeping the query
length (`s_qo`, the chunk being computed) against the context length (`s_kv`). The first five rows
are square full prefill; the last four are chunked prefill, where a short query chunk attends a
long cached context — the shape that actually dominates long-context serving. CUTLASS declines
every row here on head dimension, so `auto`'s alternative is FA2 in all nine.

| s_qo → s_kv | B200 fa2 | B200 cudnn | B200 speedup | B300 fa2 | B300 cudnn | B300 speedup |
|---|---|---|---|---|---|---|
| 1k → 1k | 0.669 ms | **0.192** | **3.49×** | 0.442 ms | **0.145** | **3.05×** |
| 2k → 2k | 1.952 | **0.579** | **3.37×** | 1.540 | **0.396** | **3.89×** |
| 4k → 4k | 6.827 | **1.954** | **3.49×** | 5.763 | **1.289** | **4.47×** |
| 8k → 8k | 25.944 | **7.722** | **3.36×** | 22.202 | **4.661** | **4.76×** |
| 16k → 16k | 101.536 | **33.264** | **3.05×** | 87.285 | **19.228** | **4.54×** |
| 512 → 4k | 1.738 | **0.494** | **3.52×** | 1.401 | **0.323** | **4.34×** |
| 1k → 8k | 6.257 | **1.755** | **3.56×** | 5.243 | **1.111** | **4.72×** |
| 2k → 16k | 24.049 | **6.933** | **3.47×** | 20.534 | **4.127** | **4.98×** |
| 512 → 32k | 12.882 | **3.885** | **3.32×** | 11.221 | **2.437** | **4.60×** |

The speedup is flat in sequence length — **3.05–3.56× on B200 and 3.05–4.98× on B300** across a
64× span of context length — and does not decay in the chunked-prefill rows. `auto` picks cuDNN
for all of them.

Repro:

```bash
python benchmarks/flashinfer_benchmark.py --routine BatchPrefillWithRaggedKVCacheWrapper \
  --backends fa2 cutlass cudnn --batch_size 4 --s_qo 2048 --s_kv 16384 \
  --num_qo_heads 64 --num_kv_heads 64 --head_dim_qk 256 --head_dim_vo 256 \
  --causal --q_dtype bfloat16 --kv_dtype bfloat16 --refcheck -vv
```

### Two caveats reviewers should weigh

**1. At d192/128 the cuDNN-first order is right on average but not universally.** An earlier
revision of this PR ordered CUTLASS ahead of cuDNN at `(192, 128)` on the strength of a single
configuration — b16/s4096 with **128×128** heads — where CUTLASS measured ~10% faster over four
independent runs (14.11/14.88/14.71/14.16 ms vs 15.65/16.16/16.43/15.62). Sweeping the rest of
that cell on B200 showed the winner flips with sequence length and head config, so a
`head_dim`-keyed override generalises one point to a space where it is wrong more often than
right (medians, 30 iters after 10 warmup):

| config (d192/128) | cutlass | cudnn | winner |
|---|---|---|---|
| b4 s1024 128×128 | 0.334 ms | **0.265** | cudnn 1.26× |
| b16 s1024 128×128 | 1.257 | **1.030** | cudnn 1.22× |
| b64 s1024 128×128 | 5.764 | **4.178** | cudnn 1.38× |
| b8 s2048 32×8 | 0.509 | **0.383** | cudnn 1.33× |
| b8 s2048 64×8 | 0.969 | **0.716** | cudnn 1.35× |
| b1 s4096 128×128 | **0.738** | 0.816 | cutlass 1.11× |
| b16 s4096 128×128 | **14.061** | 15.818 | cutlass 1.12× |
| b4 s16384 128×128 | **49.555** | 68.782 | cutlass 1.39× |

cuDNN takes 5 of 8 — every short-sequence case and every GQA case; CUTLASS leads only
long-sequence MHA. On **GLM-5.3's own 64×64 head config** cuDNN wins d192/128 outright on both
parts (tables above). The override was therefore removed and the measurements kept in a source
comment. Callers who *do* know their shape can pin the order with
`FLASHINFER_RAGGED_AUTO_BACKEND_ORDER=cutlass,cudnn`. A principled fix needs a heuristic keyed on
the axes that actually discriminate (sequence length, MHA vs GQA), measured per architecture;
that is deliberately not attempted here.

**2. The cuDNN path is silently inert on older cuDNN.** `_cudnn_supports_direct_seqlens` requires
cuDNN **library** ≥ 9.24 for fp16/bf16. On a 9.19 runtime — the version torch's
`nvidia-cudnn-cu12` pulled here by default — `auto` resolved to `cutlass` at d128/d192 and **all
the way back to FA2 at d256**, and 8 of the new tests skipped. The same measurement on 9.19 would
report CUTLASS numbers under a cuDNN label. The headline gains need a 9.24+ runtime.

### End-to-end impact

**Not measured, and deliberately not claimed.** Everything above is per-op prefill attention.
Converting it to TTFT is Amdahl arithmetic over attention's share of prefill, which varies widely
with model (dense vs MoE), sequence length and TP degree — at a 30% attention share a 3.4× kernel
gives ~21% lower prefill time; at 10% it gives ~7%. No end-to-end serving benchmark was run, so no
single number is quoted. The per-op measurements and the repro command are what this PR stands on.

## 🔍 Related Issues

None.

### ✅ Pre-commit Checks

## 🧪 Tests

All measurements above were re-taken on the final commit (after the `qo_indptr` normalization, which also changed how `benchmarks/routines/attention.py` feeds cuDNN) and then re-taken again on the standardised CUDA 13.4 / cuDNN frontend 1.29 + backend 92600 stack described in Performance; they land within run-to-run noise of the pre-normalization figures, confirming the change is a calling convention and not a kernel change.

New `tests/attention/test_auto_backend_upgrade.py` — **22 passed, 0 skipped** on B200 with cuDNN 9.26. A new `tests/attention/test_cudnn_prefill_token_indptr.py` covers the normalized explicit-cuDNN contract, including the element-unit-offset detection error. Covers the upgrade firing, every negative case (sliding window, logits soft cap, out-of-domain head dims, HND layout), numerics against FA2 over seeded byte-identical inputs, indptr preservation on the auto-resolved cuDNN path, packed-LSE output, and that an explicit `backend=` is never rewritten.

On a 9.19 runtime the same file is **13 passed, 8 skipped** — the skips are the cuDNN cases, see caveat 2.

Existing ragged suite (`test_batch_prefill_with_ragged_kv_cache`, `_custom_mask`, `test_ragged_prefill_one_valid_key`): **3278 passed, 180 skipped, 0 failed**.

A green regression run only means something if the changed code ran, so the suite's own parametrization was replayed through `plan()` and the resolutions counted: **192/1152 configs (16.7%)** leave FA2 — exactly the `d=128 + pos_encoding=NONE` cell, with `d=64`, `d=256` and all `ROPE_LLAMA` correctly staying on FA2 (measured on the CUTLASS-only first commit; the second commit widens this to d256).

The selection tests deliberately stop at `plan()`: an HND `run()` would fail on tensor shapes, so a naive test would "pass" while proving nothing about which backend was picked.

## Reviewer Notes

- Worth a look at the `d192/128` ordering (caveat 1). An earlier revision pinned CUTLASS first there; a wider sweep showed the winner flips with sequence length and head config, so the override was removed and `FLASHINFER_RAGGED_AUTO_BACKEND_ORDER` is the supported escape hatch. The ordering is the judgement call most worth a second opinion.
- Environment note that cost me time: building from a bare worktree needs `git submodule update --init` **and** the `flashinfer/data/{csrc,include,cccl,cutlass,spdlog}` symlinks an editable install creates. Without `data/cccl` specifically, nvcc silently falls back to the CUDA toolkit's older libcu++ and `fastdiv.cuh` fails with `"impl_" is not a nonstatic data member` — a confusing error with an unrelated-looking cause.

AI-assisted (Claude Code); all measurements taken in-tree on the B200 described above.

## Summary by CodeRabbit

* **New Features**
  * Added shape-aware automatic backend selection for Blackwell ragged prefill.
  * Added an environment variable override for ragged prefill backend order.
  * Added support for packed LSE shapes when batch offset statistics are provided.
  * Added validation and guidance for token-based offsets in cuDNN prefill.

* **Bug Fixes**
  * Improved cuDNN prefill offset handling and backend selection across supported shapes.

* **Documentation**
  * Documented the new backend-order environment variable.


YangXu1990uiuc:
Stacked commit 5cb5801b on top of this PR (with @Anerudhan's OK): let `auto` reach **cuDNN** ahead of CUTLASS on SM100a, via the token-unit indptr path from #3921.

**Why the cuDNN exclusion in the description no longer holds.** `cudnn_batch_prefill_with_kv_cache` has accepted `batch_offsets_units="tokens"` since #3921: the caller's ordinary token indptrs go to the engine as-is (they double as `cu_seq_len_q/kv` for the mask and as ragged offsets scaled in-engine). `auto` callers already pass token indptrs, so handing them to cuDNN changes nothing about the contract. Only an explicit `backend="cudnn"` keeps the legacy element-unit indptrs, and that is untouched.

**What the commit does**
- Replaces the inline CUTLASS `elif` with `_blackwell_ragged_auto_upgrade`, which walks an explicit preference list `("cudnn", "cutlass")` and returns the first eligible backend (none -> stay on fa2). Semantic exclusions (RoPE, custom mask, sliding window, soft cap, multi-item scoring) are shared and checked first; each backend has its own kernel-domain predicate.
- cuDNN is eligible on SM100a when `_cudnn_supports_direct_seqlens(dtype)` (cuDNN 9.24+/FE 1.25+ bf16/fp16), for `(d_qk, d_vo) in {(128,128), (192,128), (256,256)}`. Older cuDNN falls through to CUTLASS rather than to the host-side element conversion. SM110a stays on CUTLASS.
- `run()` passes the caller's indptrs with `batch_offsets_units="tokens"`; `max_token_per_sequence`/`max_sequence_kv` default to the maxima `plan()` already computes. Per-call GPU work is exactly cuDNN's own launches (Memset + `qkv_tma_setup` + fprop kernel), no FlashInfer glue kernels.
- Fixes explicit `backend="cudnn"` + `return_lse=True`, which today raises `lse must have shape (num_sequences, max_token_per_sequence, h_qo)`: the wrapper now passes a stats ragged offset and the functional API accepts the packed `[tokens, heads]` LSE with it (prerequisite for vLLM chunked prefill, which merges on LSE).
- Tests: the positive cases assert `cudnn` where the direct path is available, else `cutlass`; the semantics-dropped cases assert `fa2`; new cases cover auto->cudnn incl. d256, VARLEN output + base-2 LSE vs FA2, indptr reaching cuDNN untouched, cudnn-unavailable fallback to cutlass, explicit-cudnn `return_lse`. 21 passed; `test_cudnn_prefill*.py` regression green.

**Numbers** (SM100, 148 SMs, cuDNN 9.27 / FE 1.28, bf16 causal, ms; same token indptrs to all three)

| shape | fa2 | cutlass | cudnn | cudnn vs cutlass |
|---|---|---|---|---|
| b16 s4096 64x8 d128 | 13.65 | 8.43 | 6.65 | 1.27x |
| same, random lengths | 5.99 | 3.88 | 2.86 | 1.36x |
| b4 s16384 64x8 d128 | 53.25 | 26.92 | 23.57 | 1.14x |
| b64 s1024 32x8 d128 | 2.01 | 1.42 | 0.66 | 2.15x |
| same, random lengths | 0.88 | 0.97 | 0.52 | 1.87x |
| b16 s4096 128x128 d192/128 | 39.03 | 26.59 | 26.59 | 1.00x |
| b16 s4096 32x8 d256 | 16.39 | declined | 5.74 | only cuDNN |

Output vs FA2 rel-max 3-4e-3, base-2 LSE max-abs ~2.5e-3. Host time per call: fa2/cutlass ~17.5 us, cuDNN ~50 us (graph-cache lookup), irrelevant under CUDA graphs.

Not in this commit: fp8 `auto` (left off), and the one remaining glue kernel `lse *= log2e` (vLLM multiplies by ln2 right after; an `lse_base` knob would remove both).

<sub>note to self: claude::61d24ed2-7c90-4a97-9cbb-b91ae18136fd — "PR #5133 auto backend: why not cuDNN" · cwd /home/scratch.yanxu_libs/flashinfer · workspace /home/scratch.yanxu_libs/flashinfer-wt-5133</sub>

egilliam-nv:
Stacked commit 1e73b84a (from @egilliam-nv, taking over the units part per the Slack thread): **normalize the cudnn wrapper `qo_indptr` contract to token units**, superseding my standalone #5147.

- Token units are now the single cudnn contract for both the Paged and Ragged wrappers, whether `backend="cudnn"` is explicit or `auto` resolved to it. The `auto` token-unit plan/run path from 5cb5801b becomes *the* cudnn path, so `_cudnn_token_offsets` and the element-unit stats-offset branch are gone.
- The wrappers pass `batch_offsets_units="tokens"`; cuDNN consumes the indptrs as `cu_seq_len` plus ragged offsets with per-tensor multipliers. No conversion kernel; the paged `d_qk != d_vo` O-offset rescale is no longer needed. The low-level `cudnn_batch_prefill_with_kv_cache` still defaults to element units, so direct callers (vLLM, SGLang) are unaffected.
- Ragged: `seq_lens`/`seq_lens_q`/`max_token_per_sequence`/`max_sequence_kv` become optional for cudnn (derived from the indptrs); `return_lse` works via the stats ragged offset from 5cb5801b.
- Old element-unit indptrs to a cudnn wrapper raise `ValueError` with a migration hint.
- Addresses both CodeRabbit findings: `auto` is re-resolved on every `plan()` (requested vs resolved backend kept separately, with a re-plan test), and the cudnn run passes the planned `o_data_type`.
- Tests/harness converted to token units; `test_explicit_cudnn_element_offsets_return_lse` becomes `test_explicit_cudnn_token_indptr_return_lse`.

Verified: H100 `test_cudnn_prefill` / `_deepseek` / `_token_indptr` (519 passed); sm100 `test_auto_backend_upgrade.py` 22 passed (auto→cudnn incl. LSE vs fa2, re-plan fallback); benchmark harness `--refcheck` Paged + Ragged with `fa2 cudnn cudnn-native`.

flashinfer/prefill.py
```diff
@@ -4194,6 +4280,36 @@ def plan(
...
+                            prefix_len_ptr is not None
+                            or token_pos_in_items_ptr is not None
+                            or max_item_len_ptr is not None
+                        ),
+                    )
+                    if upgraded is not None:
+                        self._backend = upgraded
+                        # cuDNN via `auto` takes the caller's token-unit
+                        # indptrs as-is (see _blackwell_ragged_auto_upgrade).
+                        self._cudnn_token_offsets = upgraded == "cudnn"
```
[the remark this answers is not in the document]
egilliam-nv: Fixed in 1e73b84a: the requested backend is kept in `_requested_backend` and `auto` is re-resolved on every `plan()`; `test_auto_is_reresolved_on_replan` covers the sliding-window re-plan. The `_cudnn_token_offsets` flag no longer exists (token units are the only cudnn contract now).

flashinfer/prefill.py
```diff
@@ -4674,6 +4830,39 @@ def run(
...
+                    v_scale=v_scale,
+                    batch_offsets_q=self._qo_indptr_buf,
+                    batch_offsets_k=self._kv_indptr_buf,
+                    batch_offsets_stats=(
+                        self._cudnn_stats_offsets if return_lse else None
+                    ),
+                    batch_offsets_units="tokens",
+                    is_cuda_graph_compatible=self._use_cuda_graph,
+                    out=out,
+                    lse=lse,
```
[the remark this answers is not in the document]
egilliam-nv: Fixed in 1e73b84a: the ragged cudnn call passes `o_data_type=out_dtype` (there is only one cudnn call site now).

flashinfer/prefill.py
```diff
@@ -4227,6 +4342,27 @@ def plan(
...
+            if self._max_token_per_sequence is None:
+                self._max_token_per_sequence = max_qo_len
+            if self._max_sequence_kv is None:
+                self._max_sequence_kv = max_seq_in_batch
+            if (
+                self._qo_indptr_buf.dtype != torch.int32
+                or self._kv_indptr_buf.dtype != torch.int32
+            ):
+                if self.is_cuda_graph_enabled:
+                    raise ValueError(
```
saltyminty, line 4655: Could we move this check into `_blackwell_ragged_auto_upgrade`? Otherwise, auto would resolve to cudnn and then fail, instead of routing to a supported backend like fa2.

flashinfer/prefill.py
```diff
@@ -1579,6 +1580,75 @@ def _build_block_tables_from_paged_kv_indices(
...
+# Measured on B200 (bf16, causal): cuDNN 1.1-2.2x faster than CUTLASS at d128
+# (tie at 128-head d192/128 MHA) and the only backend serving d256; CUTLASS
+# 1.7-2.5x faster than FA2 where cuDNN is unavailable.
+_BLACKWELL_RAGGED_AUTO_PREFERENCE = ("cudnn", "cutlass")
+
+# (head_dim_qk, head_dim_vo) pairs validated against FA2 on the cuDNN path.
+_CUDNN_RAGGED_AUTO_HEAD_DIMS = frozenset({(128, 128), (192, 128), (256, 256)})
+
+
+def _blackwell_ragged_auto_upgrade(
```
saltyminty, line 1785: Per my agent, 

- `o_data_type` should be added to this check as cutlass doesn't support certain combinations, i.e. bf16 qkv with fp16 output.
- cutlass needs a capacity check as it later allocates buffers
-

saltyminty (commented):
I need to take a closer look at the cuda graph flow, but otherwise looks ok

Anerudhan:
Rebased onto current `main` — was 43 commits behind (branched at `63b24de5`, now on `5d0c89ea`). No conflicts, and the rebase is semantically clean rather than merely textually clean; details below.

### Why the rebase mattered here specifically

`main` moved twice in ways that land inside this PR's blast radius:

- **#5182** raised the `nvidia-cudnn-frontend` floor to **1.29.0**, re-landing #5131 after #5159 reverted it for breaking `JIT Unittest (H100)` (#5157).
- **#5161** landed the ragged-`Stats` plumbing in `flashinfer/cudnn/prefill.py` (`batch_offsets_stats`, `Stats.set_ragged_offset(...)`).

This PR widens `backend="auto"` to reach cuDNN ragged prefill — it *increases* traffic on precisely the path #5157 broke. It should not land on a `main` older than both of those. Now it doesn't.

The two are complementary rather than overlapping. #5161 added the plumbing; this PR's only edit to `cudnn/prefill.py` relaxes the caller-side LSE shape check to accept the packed `[num_tokens, h_qo]` buffer that plumbing produces:

```python
packed_ok = batch_offsets_stats is not None and lse.shape == (num_tokens, h_qo)
```

### Re-verified at the new floor

Re-ran on a **B200 (cc 10.0)** against **cuDNN FE 1.29.0 / backend 9.26.0** — i.e. at the floor this PR now inherits, not the 1.28 the branch was originally written against:

| Test file | Result |
|---|---|
| `tests/attention/test_cudnn_prefill_token_indptr.py` — the 133 nodes that failed in #5157 | **135 passed** |
| `tests/attention/test_cudnn_graph_cache_key.py` — the 1 node that failed in #5157 | **2 passed** |
| `tests/attention/test_auto_backend_upgrade.py` — this PR's own coverage | **22 passed** |

One caveat worth stating rather than burying: this environment resolves backend **9.26.0**, which is *above* the `< 9.26.0` cutoff that produced the original #5157 failure, so it cannot reproduce that specific condition. It does not need to — cudnn-frontend#1023 lowered the forward-`Stats` threshold to 9.12.0, so the guard no longer fires at any supported backend. The H100 job (backend 9.24.0.43) remains the real check; `run-ci` is applied.

### Unrelated observation, offered for a follow-up

While confirming the above I hit 14 failures in `test_auto_backend_upgrade.py` that turned out to be environmental, not code: `FileNotFoundError` on `flashinfer/data/csrc/*.jinja`. That directory is the symlink tree created by `pip install -e .`, so any fa2/CUTLASS test JITting from a source tree without an editable install (or without `3rdparty/cutlass` initialised) fails this way, while the cuDNN and CuTe-DSL paths pass unaffected — which makes the failure look backend-specific when it isn't. Not something this PR should fix, but a clearer error than a raw `FileNotFoundError` on a Jinja path would save the next person the detour.

flashinfer/prefill.py
Anerudhan: Good catch — moved it into `_blackwell_ragged_auto_upgrade` as `cudnn_indptr_is_int32` in 4263872.

Outside graph mode `plan()` can just re-dtype the buffer in place, so cuDNN stays eligible there; it's only under capture that it can't, and that's now an eligibility test rather than a late raise, so `auto` routes on to cutlass/FA2. An explicit `backend="cudnn"` still raises, which seems right for someone who named the backend.

Added `test_auto_declines_cudnn_on_non_int32_indptr_under_cuda_graph`.

flashinfer/prefill.py
Anerudhan: Both confirmed and fixed in 4263872.

**Output dtype** — your agent is right. `DISPATCH_DTYPE_IN_OUT` only handles `out == in` for fp16/bf16 and sends everything else to its FP8 branch, which a bf16 input doesn't match, so the call dies on a check reporting unsupported *head dimensions* — misleading enough that it'd cost someone an afternoon. `o_data_type == q_data_type` is now part of the CUTLASS domain test.

**Capacity** — also real, and a bit worse than I expected: `fmha_varlen_plan` allocates the work-index buffers at a fixed 131072 entries, and `plan_kernel` writes one entry per (qo_tile, head, batch) with no bounds check. So e.g. b64/s8192/128 heads is 262144 work items and overruns them. `auto` now counts the work items first — from the host indptr copy `plan()` already materialises, so no extra device sync — and declines CUTLASS past capacity.

Worth flagging that the overflow itself is pre-existing on the explicit `backend="cutlass"` path; this PR only stops `auto` steering people into it. Happy to file a separate issue for the kernel-side bounds check if you'd like — it felt out of scope here.

Tests: 3 added (dtype mismatch, capacity straddled on head count alone so the two cases differ in nothing else, non-int32 indptr under capture). 25 passed on B200 / cuDNN 9.26.

Anerudhan:
@saltyminty addressed both comments in 4263872 — thanks, they were good ones.

The common thread was that `auto` could resolve to a backend that then refused the call, which is worse for the caller than staying on FA2. Three conditions were being enforced downstream of selection instead of inside it (cuDNN's int32 indptr under graph capture, CUTLASS's output dtype, CUTLASS's plan-buffer capacity); all three are now eligibility tests in `_blackwell_ragged_auto_upgrade`, and I documented that invariant on the function so the next person adding a backend doesn't reintroduce it.

Also re-measured everything on one standardised stack (CUDA 13.4, cuDNN frontend 1.29 / backend 92600) and refreshed the PR description with B200 + B300 tables, plus a new GLM-5.3 sequence-length sweep. Two things there worth a look:

- **Caveat 1 changed.** It previously claimed CUTLASS beats cuDNN at d192/128 by 9–12%. That only held at 128×128 heads — a wider sweep has cuDNN taking 5 of 8 configs, and on GLM-5.3's own 64×64 heads cuDNN wins outright on both parts. The caveat now says that and keeps the CUTLASS-favouring numbers rather than dropping them.
- **d256/256 across context lengths** is the new table: 3.05–3.56× over FA2 on B200 and 3.05–4.98× on B300, flat across a 64× span of sequence length and holding up in the chunked-prefill rows.

Still on your list: the CUDA graph flow you wanted a closer look at. Re-requesting review.

flashinfer/prefill.py
```diff
@@ -1678,6 +1680,182 @@ def _build_block_tables_from_paged_kv_indices(
...
+            if (
+                (is_sm100a_supported(device) or is_sm110a_supported(device))
+                and q_data_type in (torch.float16, torch.bfloat16)
+                # DISPATCH_DTYPE_IN_OUT only handles out == in for fp16/bf16;
+                # any other pairing falls into its FP8 branch, which a bf16 or
+                # fp16 input does not match, and the call dies on a check that
+                # reports unsupported *head dimensions*. Mismatched output
+                # dtype is therefore out of domain, not a slow path.
+                and o_data_type == q_data_type
+                and ((head_dim_qk == 128 and head_dim_vo == 128) or head_dim_qk == 192)
```
saltyminty: cutlass doesn't support head_dim 192/192, this check allows any `head_dim_vo` for `head_dim_qk==192`

flashinfer/prefill.py
saltyminty: This predates this PR, but now that cutlass is exposed through auto mode, `qo_indptr.device` needs to be updated to `self.device`

flashinfer/prefill.py
```diff
@@ -4358,6 +4596,27 @@ def plan(
...
+                        "the cudnn backend reads qo_indptr/kv_indptr as int32; "
+                        "pass int32 indptr buffers in CUDA-graph mode"
+                    )
+                self._qo_indptr_buf = self._qo_indptr_buf.to(torch.int32)
+                self._kv_indptr_buf = self._kv_indptr_buf.to(torch.int32)
+            self._cudnn_stats_offsets = self._qo_indptr_buf
+
         if self._backend == "cutlass":
             self._plan_info = fmha_varlen_plan(
                 self._cached_module, qo_indptr, kv_indptr, num_qo_heads, causal
```
saltyminty: This predates this PR, but now that cutlass is exposed through auto mode, we should be passing in the device mirrors `self._qo_indptr_buf` and `self._kv_indptr_buf` instead of `qo_indptr` and `kv_indptr`.

flashinfer/prefill.py
saltyminty, line 5335: This function isn't cuda graph re-plan safe (due to the buffer allocation), long term we should probably have these buffers updated in place when in cuda graph mode, but for now we should probably have a guard in the wrapper to explicitly error on replan or reject cutlass.

flashinfer/prefill.py
Anerudhan: Good catch — now an explicit `(128,128)/(192,128)` frozenset instead of `head_dim_qk == 192`. 8248c88

flashinfer/prefill.py
Anerudhan: Addressed — uses `self.device` now. 8248c88

flashinfer/prefill.py
Anerudhan: Addressed — passes `self._qo_indptr_buf`/`self._kv_indptr_buf`, and `_max_qo_len` follows them. 8248c88

flashinfer/prefill.py
Anerudhan: Addressed — `auto` treats graph mode as CUTLASS-ineligible and explicit `backend="cutlass"` now raises at plan(); in-place buffers left as the follow-up. 8248c88

Anerudhan:
Known issue, if the q,k,v is not contiguous cudnn will not work correctly.

saltyminty (approved):

saltyminty:
Per offline discussion:
Both cudnn and cutlass backends assume fully contiguous QKV tensors. For callers passing in non-contiguous QKV tensors, such as in T3HD, and using auto mode, this PR may then select the unsupported cudnn and cutlass and fail at runtime. Because tensor strides are not known at plan time, backend selection cannot account for this during plan time. Fa2 accesses tensor strides at run time and thus doesn't have this issue, while cudnn requires tensor strides/offsets at compile time.

It is believed that T3HD is not widely used, so we will be merging this and seeing if an issue gets raised. We can evaluate adding tensor strides or contiguity boolean flags to the plan API if that is the case.

YangXu1990uiuc:
@Anerudhan heads-up: I am going to stack one more commit on this branch (same arrangement as 5cb5801b), covering the review threads still open at 8248c884 plus the contiguity point from the last comment. Feel free to drop or amend it.

1. **Attention sinks** (CodeRabbit, unresolved): `_blackwell_ragged_auto_upgrade` does not look at `self._sinks`, and the cuDNN run branch does not check it either, so `auto -> cudnn` with vLLM's wrapper-stashed sinks would run without them (wrong output, no error; the cute-dsl branch raises). Plan: sinks set at `plan()` time disqualify the upgrade; the cuDNN and CUTLASS run branches raise the same way cute-dsl does if sinks appear later.
2. **fp8 via auto** (CodeRabbit, unresolved): `_cudnn_supports_direct_seqlens()` is True for fp8 on recent cuDNN, so the cuDNN eligibility test admits fp8 q/k/v, and the cuDNN run branch returns before the fp8 handling further down. Plan: cuDNN eligibility limited to fp16/bf16, like CUTLASS's.
3. **int64 indptr** (CodeRabbit, unresolved): `plan()` converts `_qo/_kv_indptr_buf` to int32 but `_o/_v_indptr_buf` keep aliasing the pre-conversion tensors. Plan: convert all four, rebinding the defaulted ones.
4. **Non-contiguous q/k/v (T3HD)**: cuDNN itself follows strides and the graph already takes `q.stride()` / `k.stride()`; what assumes contiguity is the token-unit ragged-offset *multiplier*, hard-coded to `h*d` in `cudnn/prefill.py`. With the tensor's actual token stride as the multiplier the cuDNN route handles T3HD; CUTLASS's assumption stays and gets a run-time check.

Each gets a test in `test_auto_backend_upgrade.py` / the cuDNN prefill tests; I will verify on B200 and post numbers here.

<sub>note to self: claude::61d24ed2-7c90-4a97-9cbb-b91ae18136fd — "FlashInfer frost integration: PR chain 1035-1039" · cwd /home/scratch.yanxu_libs/flashinfer · workspace /home/scratch.yanxu_libs/flashinfer-wt-5133</sub>

YangXu1990uiuc:
Stacked **16272f8f** on the branch, per the heads-up above (@Anerudhan drop or amend as you like). One commit, four items, each with a test:

- **Sinks**: `_sinks` set on the wrapper at `plan()` disqualifies the upgrade (`has_sinks` in `_blackwell_ragged_auto_upgrade`); set after `plan()`, the cuDNN and CUTLASS run branches raise the way the cute-dsl branch already did. `test_auto_declines_with_wrapper_sinks`, `test_upgraded_backends_refuse_sinks_set_after_plan`.
- **fp8**: cuDNN eligibility is fp16/bf16 only (CUTLASS already was). `test_auto_keeps_fp8_off_cudnn` checks the eligibility function directly — a full fp8 `plan()` has no Blackwell ragged backend to land on, fa2 rejects fp8 q at JIT time.
- **int64 indptrs**: `plan()` re-dtypes all four indptr buffers (o/v default to aliases of q/kv taken before the conversion), and graph-mode eligibility checks all four. `test_explicit_cudnn_int64_indptr_converts_all_four_buffers` (explicit cudnn, int64, matches fa2 with LSE).
- **Non-contiguous q/k/v (T3HD)**: the cuDNN graph already took `q.stride()`/`k.stride()`; the piece that assumed contiguity was the token-unit ragged-offset multiplier hard-coded to `h*d`. It is now each tensor's own token stride, and `_sdpa_prefill_key_fn` includes the q/k/v strides so a contiguous same-shape call does not replay its graph for the packed views. `test_cudnn_ragged_handles_packed_t3hd_views` (views into one `[tokens, 3, h, d]` buffer, contiguous graph built first, both match fa2). CUTLASS keeps its contiguity assumption, so the "auto may pick an unsupported backend for T3HD" caveat now only concerns CUTLASS; cuDNN handles the layout.

Verified on B200 (cc 10.0; cuDNN backend 9.25.1 with a 1.30-dev frontend): `test_auto_backend_upgrade.py` 33 passed; `test_cudnn_prefill_token_indptr.py` + `test_cudnn_prefill.py` 423 passed / 288 skipped / 288 xfailed (pre-existing). pre-commit clean.

<sub>note to self: claude::61d24ed2-7c90-4a97-9cbb-b91ae18136fd — "FlashInfer frost integration: PR chain 1035-1039" · cwd /home/scratch.yanxu_libs/flashinfer · workspace /home/scratch.yanxu_libs/flashinfer-wt-5133</sub>

saltyminty:
Stacked another commit to
• prevent auto cutlass selection with non-int32 offsets 
• add o_dtype to cudnn cache key

YangXu1990uiuc:
Follow-up for the cuDNN branch of `auto`: #5245 builds the ragged cuDNN graph once per length class and passes the real (batch, max_seq_len) at execute time through cuDNN's shape override. Without it every serving step with a new shape paid a 55-70 ms cuDNN plan build (283 of 300 steps in a chunked-prefill stream on SM100); with it 2 of 300, bit-identical output. @Anerudhan @saltyminty FYI since you reviewed the auto path.

PR #5144, opened 2026-09-11
refactor(attention): use fixed page tables for paged block-sparse attention

heyuhhh:
## 📌 Description

Make both paged PrimTS block-sparse entry points, `BlockSparsePagedTSWrapper.run()` and the one-shot
`block_sparse_attention_with_paged_kv_cache()`, consume a fixed 2D `block_tables` tensor with an explicit row
stride instead of CSR page indices (`paged_kv_indptr` / `paged_kv_indices`). This matches the PrimTS decode
contract from #4829, so callers that already hold page tables (TensorRT-LLM's paged KV cache manager, for example)
can pass a zero-copy view instead of converting on the host. The reusable wrapper reads the table directly during
`run()`; the one-shot API validates the live prefix of every page-table row on the device before planning. Like the
dense PrimTS wrappers, `run()` accepts `validate=False` to treat every argument as a trusted binding and skip the
explicit structural, plan-geometry, and alias checks.

The PrimTS block-sparse trace templates are rewritten as literal schemas in two factories, one per K/V storage
form, instead of deriving the paged and wrapper templates from another template by deleting or rewriting entries.
Template names, dispatchers, and the axis, input, constraint, and tag sets are unchanged apart from the page-table
and `validate` inputs; the six block-sparse goldens are regenerated and a test now pins every golden to its
template.

Follow-up to #4872 and #5002; the contiguous entry points are untouched.

## 🔍 Related Issues

Follow-up to #4829 (PrimTS decode page-table contract), #4872 and #5002 (PrimTS block-sparse attention).

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

`tests/attention/test_attention_ts_block_sparse.py` (paged one-shot and wrapper cases now build page tables,
including padded row strides and live-prefix validation) and `tests/trace/test_fi_trace_template_consistency.py`
(golden-to-template pinning): 1056 passed, 1 skipped on B200 (SM100) with CUTLASS DSL 4.7.0; pristine `main` gives 1058 passed, 1 skipped in the same environment (the difference is the reshaped paged test cases).

## 🔬 Experimental Track

```experimental-tests
#
#
```

## Reviewer Notes

The paged API signature changes: `paged_kv_indptr` / `paged_kv_indices` are replaced by `block_tables` (Int32
`[B, C]`, padded outer stride allowed) plus `seq_lens_kv`, and `run()` gains `validate=True`. The README documents
the new value contract. TensorRT-LLM already consumes this form through its vendored PrimTS copy.

## Summary by CodeRabbit

- **Changed**
  - Paged block-sparse attention now uses fixed 2D `block_tables` with optional row padding for KV-cache page addressing, replacing the previous indirection arrays.
  - Run methods support optional structural and aliasing validation, including a faster trusted-input path when disabled.
  - One-shot APIs now accept sequence lengths positionally.

- **Bug Fixes**
  - Validation now checks page capacity, tensor structure, device compatibility, aliasing, and physical page IDs with clearer errors.

- **Tests**
  - Expanded coverage for block-table validation, tracing schemas, and paged block-sparse correctness.


qsang-nv:
**Two loose ends after the CSR → `block_tables` migration.**

`BlockSparsePagedTSWrapper.plan()`'s docstring (`flashinfer/attention/prims_ts/block_sparse.py:564-565`) still says "Every run supplies live page-table offsets, page IDs, sequence lengths, ...", describing the removed `paged_kv_indptr`/`paged_kv_indices` CSR contract. `run()`'s own docstring and the README were updated to describe `block_tables`, but `plan()`'s was not.

`test_prims_ts_block_sparse_goldens_match_templates` (`tests/trace/test_fi_trace_template_consistency.py:803-824`) only compares each golden's input `name`, `optional`, and `description` against the template. It never compares input `shape`, and never touches `outputs` (name/shape/dtype/param) at all. A golden with a mis-bound tensor shape or wrong output ABI would still pass this test.

tests/attention/test_attention_ts_block_sparse.py
```diff
@@ -5517,16 +5626,18 @@ def varlen_reference(
 
     block_indptr = indptr_a.clone()
     block_indices = indices_a.clone()
-    paged_kv_indices = page_ids_a.clone()
-    paged_kv_indptr = paged_kv_indptr_a.clone()
+    block_table_storage = torch.full(
+        (case.batch_size, 3), -1, device="cuda", dtype=torch.int32
+    )
+    block_tables = block_table_storage[:, :2]
+    block_tables.copy_(torch.stack((page_ids_a[:2], page_ids_a[1:3])))
```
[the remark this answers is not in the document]
heyuhhh: The rows are intentional. This test uses page_size=64; in round A request 0 holds 48 live tokens and in round B request 1 holds 64, so each owns a single live page. block_tables documents that entries beyond the live length are ignored, so the trailing entries [0, 2] and [2, 2] exercise exactly that rule while the live prefixes [0], [2, 3], [1, 4], [2] match the references. Replacing them with the reference page IDs would drop that coverage. The rows are now spelled out as literals with a comment explaining this.

heyuhhh:
@qsang-nv thanks for the review! The docs and validations have made up in commit 61f8848864434c175ccbed63fb997bcc21e2d2d0

flashinfer/attention/prims_ts/block_sparse.py
```diff
@@ -698,28 +731,26 @@ def run(
 def block_sparse_attention_with_paged_kv_cache(
```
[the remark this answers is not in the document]
heyuhhh: It's an intentional change

flashinfer/attention/prims_ts/block_sparse.py
```diff
@@ -815,6 +849,12 @@ def block_sparse_attention_with_paged_kv_cache(
         kv_dtype=k_cache.dtype,
         output_dtype=q.dtype if out is None else out.dtype,
     )
+    if table_capacity * page_size < static.seq_len_kv:
```
PerkzZheng: those checks seem duplicated as shown in flashinfer/attention/prims_ts/_block_sparse/runtime.py

PerkzZheng (approved):

qsang-nv (approved):

flashinfer/attention/prims_ts/block_sparse.py
heyuhhh: Thanks for the findings! We can not directly remove these checks but can hoist it as a common helper. if `wrapper.run()` with `validate=True` it actually happens twice. However, they are in two different scopes: one is in one-shot API for the following inspect kernel and another is in `wrapper.run(..., validate=True)`

heyuhhh: Updated in bded04a57885005e6c5a4b37dea7e01cfa3310dc

bkryu:
@flashinfer-bot run

heyuhhh:
@qsang-nv @bkryu There are some conflicts and i have resolved. Could you please help trigger the CI again? Thanks!

yuxianq (approved):

PR #5147, opened 2026-09-11
refactor(prefill): normalize cudnn wrapper qo_indptr to token units

egilliam-nv:
Prerequisite for letting `backend="auto"` pick cuDNN. #5133 excludes cuDNN from `auto` because the `backend="cudnn"` prefill wrappers were the only backend taking **element-unit** `qo_indptr` (`cumsum(seq_lens) * num_heads * head_dim`); every other backend takes token units. That deviation made a units mismatch an `auto` footgun, so this PR removes it.

## Change
- `BatchPrefillWithPagedKVCacheWrapper` / `BatchPrefillWithRaggedKVCacheWrapper` with `backend="cudnn"` now take **token-unit** `qo_indptr` (and `kv_indptr`, `v_indptr`, `o_indptr`) like every other backend.
- The wrappers pass `batch_offsets_units="tokens"` to `cudnn_batch_prefill_with_kv_cache`, which consumes the indptrs directly as `cu_seq_len` plus ragged offsets with per-tensor `num_heads * head_dim` multipliers (the direct path from #4222). No conversion kernel is added.
- The paged `head_dim_qk != head_dim_vo` O-offset rescale from #4663 is gone: the O multiplier handles it.
- Passing the old element-unit `qo_indptr` to a cudnn wrapper raises `ValueError` with a migration hint (detected when `q.numel() == qo_indptr[-1]`).
- `plan()`'s `total_num_rows` / `_max_q_len` bookkeeping, previously inflated by the element factor for cudnn, is now correct.

## Compatibility
- The **low-level** `cudnn_batch_prefill_with_kv_cache` is unchanged and still defaults to `batch_offsets_units="elements"`. vLLM and SGLang call the low-level function directly with element offsets and are unaffected; neither uses the `backend="cudnn"` wrappers.
- `benchmarks/routines/attention.py` now feeds token-unit indptrs to the cudnn wrappers and keeps element-unit offsets only for the low-level `cudnn-native` path.

## Testing
On H100 (sm90): `test_cudnn_prefill` **288**, `test_cudnn_prefill_deepseek` **96**, `test_cudnn_prefill_token_indptr` **135** (new `test_cudnn_wrapper_token_indptr`: token-unit cudnn wrapper matches the fa2 wrapper for `d_qk == d_vo` and `192/128`, and element-unit indptrs raise the migration error). Benchmark harness `--refcheck` passes for both Paged and Ragged routines with `fa2 cudnn cudnn-native`.

Follow-up (not in this PR): wrapper `return_lse` for cudnn allocates a 2D `lse` while the low level needs `(batch, max_q_len, heads)`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Standardized cuDNN prefill offset handling to use token-based indexing.
  - Added validation to reject legacy element-based offsets with a clear migration error.
  - Corrected paged and ragged prefill execution and validation across supported configurations.

- **Tests**
  - Added coverage for token-based offsets and reference-backend consistency.
  - Updated existing attention tests to use token-based sequence offsets.

### DIFF
--- a/benchmarks/routines/attention.py
+++ b/benchmarks/routines/attention.py
@@ -1674,7 +1674,7 @@ def testBatchPrefillWithPagedKVCacheWrapper(args):
         )
         .long()
         .to(device)
-    )  # For cuDNN
+    )  # Element-unit offsets, for the low-level cudnn-native call only
     qo_indptr = (
         torch.cat(
             [
@@ -1875,7 +1875,7 @@ def testBatchPrefillWithPagedKVCacheWrapper(args):
                 )
             )
             backend_wrappers["cudnn"].plan(
-                q_indptr,
+                qo_indptr,
                 kv_indptr,
                 kv_indices,
                 kv_last_page_len,
@@ -2612,7 +2612,7 @@ def testBatchPrefillWithRaggedKVCacheWrapper(args):
         )
         .long()
         .to(device)
-    )  # For cuDNN
+    )  # Element-unit offsets, for the low-level cudnn-native call only
 
     k_indptr = torch.cat(
         [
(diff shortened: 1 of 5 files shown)


egilliam-nv:
Superseded: stacked onto #5133 (commit 1e73b84a) so the units normalization and the cudnn `auto` upgrade land together.

PR #5161, opened 2026-09-12
cudnn prefill: return a packed LSE, declared as ragged Stats

egilliam-nv:
## 📌 Description

`cudnn_batch_prefill_with_kv_cache(..., return_lse=True)` returns the packed LSE that FlashInfer's API documents, `(total_qo_tokens, num_qo_heads)`, and declares it to cuDNN as a **ragged Stats tensor** over the query token indptr, the same packing as `q`. The historical padded `(batch, max_token_per_sequence, num_qo_heads)` buffer is still accepted when passed in explicitly.

**Why.** FlashInfer's LSE contract is a packed `[qo_len, num_qo_heads]` tensor: the wrappers allocate that shape, the cascade / merge kernels consume it, and every other backend returns it. The cuDNN low level wrote a padded per-request block instead, so a direct caller got a shape nothing downstream accepts, and the `backend="cudnn"` wrappers could not use the low level's LSE at all. #5133 fixed the ragged wrapper by computing a Stats offset on the wrapper side; this PR moves the fix into the low level, so the paged wrapper and direct callers get the packed LSE too, and the wrapper-side plumbing goes away. A ragged Stats tensor is cuDNN's sanctioned packed form: cuDNN writes each request's rows straight into the packed buffer, with no padding rows to fill with `-inf` and no gather afterwards. It is also the form the FROST engines' FlashInfer-shaped conformance suite exercises (NVIDIA/cudnn-frontend#1028).

**Change.** `flashinfer/cudnn/prefill.py`
- `lse=None` allocates the packed `(total_qo_tokens, h_qo)` buffer on the cuDNN graph path. The `"cubin"` backend keeps the padded form it writes.
- A passed-in `lse` may be packed or padded; any other shape is a `ValueError` naming both. It must be contiguous float32 on `q`'s device, and a passed-in `out` contiguous on `q`'s device, checked up front rather than at graph execution.
- For a packed LSE the Stats ragged offsets are derived when not given: the token-unit q indptr on the `"tokens"` path (the graph applies the `h_qo` multiplier), `batch_offsets_q // q.stride(0) * h_qo` on the element-unit path (element offsets count elements of each tensor's own storage, so a non-contiguous `q` such as a T3HD view is handled), and a device-built `[0, total]` for a single request without offsets (CUDA-graph capturable). `batch_offsets_stats` with a padded `lse` is rejected: the offsets address packed rows.
- The legacy token-to-element conversion (cuDNN too old for the direct path) uses each tensor's real token stride, as the direct path's multipliers already did.
- **NVBug 6783545 workaround (cuDNN < 9.28).** cuDNN's single-token (`s_q == 1`) GQA kernel mis-stores a ragged Stats tensor: most heads of the packed LSE are never written. Fixed in cuDNN 9.28 (MR !4496). Below that, with `max_token_per_sequence == 1` and one token per request, the padded `(b, 1, h)` block is byte-identical to the packed buffer, so the low level binds the packed buffer as the padded form (no ragged Stats offset) and returns it as packed. Ragged and paged KV. A batch that mixes zero-length and single-token requests cannot be served packed on affected versions and raises, asking for a padded `lse`.
- The graph cache key includes whether Stats carries a ragged offset. The two are different graphs; previously they shared a key.

`flashinfer/prefill.py`
- The ragged wrapper no longer computes or passes `_cudnn_stats_offsets`; the low level derives the offset from the token-unit `qo_indptr` it already receives.
- The #5245 single-token GQA guards (`auto` eligibility, explicit-cudnn `NotImplementedError` on `return_lse`) narrow to the one shape the low level cannot serve below cuDNN 9.28: a batch mixing zero-length and one-token requests. `auto` now keeps cuDNN for ordinary single-token GQA rows, LSE included.

**Behaviour change to flag:** low-level callers passing `return_lse=True, lse=None` receive the packed 2-D LSE instead of the padded 3-D one. The trace template already documented `["num_tokens", "num_heads_qo"]`. Callers that pass their own padded buffer are unchanged. vLLM and SGLang call the low level and discard the LSE.

## 🔍 Related Issues

Deferred from #4663; builds on #5133 and #5245; NVBug 6783545 (cuDNN fix MR !4496, 9.28); NVIDIA/cudnn-frontend#1023, #1036.

### ✅ Pre-commit Checks

## 🧪 Tests

`tests/attention/test_cudnn_prefill_lse_layout.py` (new): packed by default on both offset units × batch 1/4 × causal, checked per token against a float reference; an explicit padded buffer still works and equals the packed one request by request; non-contiguous `q` (T3HD view) on both unit paths; single-token rows, GQA and MHA, ragged and paged, with a NaN-poisoned buffer; a single request with no offsets captured into a CUDA graph; wrong shapes, wrong dtype, non-contiguous storage and padded-plus-offsets each rejected; `BatchPrefillWithRaggedKVCacheWrapper(backend="cudnn").run(return_lse=True)` returns the documented 2-D LSE matching the default backend. `test_cudnn_prefill_lse_is_base2` updated for the packed shape; the strict xfail `test_cudnn_single_token_gqa_lse_is_correct` from #5245 becomes a passing test; `test_auto_takes_cudnn_for_single_token_gqa_rows` and `test_single_token_gqa_with_empty_rows_stays_off_cudnn_lse` replace the two guard tests.

Local, 2026-09-16, cuDNN 9.27 dev (bug present) + cudnn-frontend develop (1.30), on top of main with #5133 and #5245:

| GPU | suites | result |
|---|---|---|
| H100 (sm90) | lse_layout, token_indptr, test_cudnn_prefill, deepseek, ragged_shape_override | 552 passed, 693 skipped (pre-existing shape skips) |
| H100 (sm90) | paged_cu_seqlens | 32 passed |
| B200-class (sm100) | test_auto_backend_upgrade | 36 passed |
| B200-class (sm100), `CUDNN_FRONTEND_ENABLE_FROST_ENGINES=1`, `FLASHINFER_CUDNN_PREFILL_SHAPE_OVERRIDE=0` | lse_layout, token_indptr | 158 passed |

## Reviewer Notes

FROST engines × #5245 shape override: with `CUDNN_FRONTEND_ENABLE_FROST_ENGINES=1` and the override on, the frontend's FROST path rejects the override graph (`cu_seq_len_q must have B + 1 = 4097 elements; got 2`). This is pre-existing on main (the two `test_cudnn_wrapper_token_indptr` cases fail there the same way); this PR exposes it to the `return_lse=True` paths as well, because they now carry a Stats offset and so qualify for the override. It is a frontend-side interaction (FROST should decline override graphs and fall back), not something this PR changes; noting it here so it is tracked.


flashinfer/cudnn/prefill.py
```diff
@@ -913,19 +923,31 @@ def cudnn_batch_prefill_with_kv_cache(
...
+    packed_shape = (num_tokens, h_qo)
+    padded_shape = (num_sequences, max_token_per_sequence, h_qo)
+    if return_lse and lse is None:
+        lse = torch.empty(
+            packed_shape if use_cudnn_graph else padded_shape,
+            device=q.device,
+            dtype=torch.float32,
+        )
+    lse_packed = lse is not None and tuple(lse.shape) == packed_shape
+    if lse is not None and not lse_packed and tuple(lse.shape) != padded_shape:
```
egilliam-nv, line 1206: Fixed in bc0e6a3a: a caller-provided padded `lse` together with `batch_offsets_stats` now raises `ValueError` (`batch_offsets_stats addresses a packed LSE ...`), with `test_cudnn_prefill_lse_rejects_padded_with_stats_offsets` as the regression test. The default (`lse=None`) path is unaffected since it allocates the packed buffer.

flashinfer/cudnn/prefill.py
```diff
@@ -922,31 +933,40 @@ def cudnn_batch_prefill_with_kv_cache(
...
```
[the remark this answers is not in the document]
egilliam-nv: Done in 656b67e5: `lse` is checked for float32, q's device and contiguity before the shape checks, with `test_cudnn_prefill_lse_rejects_unbindable_buffers` covering the dtype and non-contiguous cases.

flashinfer/cudnn/prefill.py
```diff
@@ -998,6 +1028,31 @@ def cudnn_batch_prefill_with_kv_cache(
...
+                        else num_tokens * h_qo,
+                    ],
+                    dtype=torch.int32,
+                    device=q.device,
+                )
+            elif batch_offsets_units == "tokens":
+                batch_offsets_stats = batch_offsets_q
+            else:
+                batch_offsets_stats = torch.div(
+                    batch_offsets_q, d_qk, rounding_mode="floor"
```
saltyminty: We should be using real q strides here, otherwise we'd get failures when deriving from noncontiguous Q. Specifically, this should read:
```
batch_offsets_stats = (
    torch.div(batch_offsets_q, q.stride(0), rounding_mode="floor")
    * h_qo
)
```

egilliam-nv: Good catch, thanks. Done in d54ada9c exactly as you wrote it: `torch.div(batch_offsets_q, q.stride(0), rounding_mode="floor") * h_qo`. The legacy token-to-element conversion a few lines below had the same contiguity assumption for q/k/v (`h * d` multipliers), while the direct path already uses the real token strides, so it now uses `q.stride(0)` / `k_cache.stride(0)` / `v_cache.stride(0)` for the 3-D case too. New test `test_cudnn_prefill_lse_packed_with_noncontiguous_q` runs a T3HD-style q view (token stride 3·h·d) on both unit paths against the contiguous result.

flashinfer/cudnn/prefill.py
```diff
@@ -922,31 +933,50 @@ def cudnn_batch_prefill_with_kv_cache(
...
+    # h_qo) buffer. The cubin backend writes the padded form only. A packed LSE
+    # is declared to cuDNN as a ragged Stats tensor over the query token
+    # indptr, exactly like the packed q it accompanies.
+    packed_shape = (num_tokens, h_qo)
+    padded_shape = (num_sequences, max_token_per_sequence, h_qo)
+    if return_lse and lse is None:
+        lse = torch.empty(
+            packed_shape if use_cudnn_graph else padded_shape,
+            device=q.device,
+            dtype=torch.float32,
```
YangXu1990uiuc, line 1193: **[P2] Preserve compatibility with affected cuDNN versions at low-level/paged LSE entry points**

Severity updated from P1 to P2. This is the known cuDNN single-token GQA ragged-Stats issue already tracked in [#5245](https://github.com/flashinfer-ai/flashinfer/pull/5245). That PR protects the ragged wrapper through auto-dispatch fallback and rejection of explicit cuDNN LSE calls for this shape. Its low-level API is deliberately left unguarded, and the paged wrapper is outside the protection. The finding here is the narrower compatibility gap when #5161 changes the default Stats layout for those callers on affected cuDNN builds. #5245 is still open as of this update.

For `max_token_per_sequence=1` with `Hq > Hkv`, this changes a working default LSE call into silent numerical corruption. I reproduced it on H100 and SM100 with BF16, `B=3, Hq=8, Hkv=2, D=128`, query lengths `[1,1,1]`, KV lengths `[16,31,65]`, `causal=False`, and `return_lse=True`. It affects both ragged and paged KV (page size 16), using either token or element offsets.

At this head, passing a NaN-filled packed `[3,8]` LSE leaves **18/24 values NaN**; the default `lse=None` also disagrees with an FP32 logsumexp reference. Attention output remains correct. Explicit padded `[3,1,8]` Stats writes all 24 values correctly. The identical `lse=None` call on base `24c30bdd` returns correct padded LSE on SM100, so this is a regression in the default API even though the cuDNN ragged-Stats limitation already exists. The `max_token_per_sequence=2` control passes.

Please use a safe Stats path for this case (e.g. padded Stats followed by packing), or reject/reroute it explicitly on affected configurations. A wrapper-only guard would leave this public low-level API exposed. Add a single-token GQA test with a poisoned output buffer so unwritten heads cannot pass accidentally.

Yang's codex bot

flashinfer/cudnn/prefill.py
```diff
@@ -998,6 +1028,33 @@ def cudnn_batch_prefill_with_kv_cache(
...
+        if return_lse and lse_packed and batch_offsets_stats is None:
+            # Each request's rows of the packed LSE start where its query
+            # tokens start. Token-unit offsets are the q indptr itself (the
+            # graph applies the per-tensor multiplier h_qo); element-unit q
+            # offsets are token_offset * q.stride(0) (h_qo * d_qk only for a
+            # contiguous q), so the token offset is recovered with q's real
+            # token stride and rescaled by h_qo. A single request without
+            # offsets starts at row 0 and spans the buffer.
+            if batch_offsets_q is None:
+                batch_offsets_stats = torch.tensor(
```
YangXu1990uiuc: **[P2] Create the implicit single-request Stats offsets without a host copy during capture**

When a single-request caller omits `batch_offsets_q` (explicitly supported by this API), `return_lse=True` now reaches this `torch.tensor([...], device=q.device)` on every call. During CUDA Graph capture it copies an unpinned CPU tensor to CUDA and raises `RuntimeError: Cannot copy between CPU and CUDA tensors during CUDA graph capture unless the CPU tensor is pinned`.

Reproduced on H100 after three warmup calls on a side stream, with `Q=[37,8,128]`, `K/V=[64,2,128]`, BF16, GPU int32 `actual_seq_lens_q=[37]` / `actual_seq_lens_kv=[64]`, `causal=False`, `return_lse=True`, `is_cuda_graph_compatible=True`, and all offsets omitted. The same script at base `24c30bdd` captures and successfully replays after changing Q; this head fails before executing the captured attention.

Please construct these offsets through capture-safe device operations or prepare/reuse them before capture, and add a regression test for the omitted-offset path. Warmup does not fix it because this allocation/copy runs again inside capture.

Yang's codex bot

YangXu1990uiuc (changes requested):
Reviewed `d54ada9c80af40ac5f29ced7c2955233fa6c55f3` against base `24c30bddbe1628e225458380b476de756b506f69`. Requesting changes for two scoped P2 findings, detailed inline. The LSE finding is downgraded from P1 after accounting for the existing workaround in [#5245](https://github.com/flashinfer-ai/flashinfer/pull/5245): that PR protects the ragged wrapper, while the remaining #5161 concern is compatibility with affected cuDNN versions at low-level/paged entry points outside that guard.

1. **P2: preserve compatibility with affected cuDNN versions when changing the default LSE layout.** With `B=3`, `Hq=8`, `Hkv=2`, `D=128`, BF16, query lengths `[1,1,1]` and KV lengths `[16,31,65]`, a NaN-initialized packed LSE has only **6/24** entries written. The default allocation returns incorrect values in the remaining entries. Reproduced on H100 and SM100, with both ragged/paged KV and token/element offsets. Attention output is correct. Explicit padded Stats is correct at this head; the identical default-buffer call is correct on the original base on SM100. The underlying cuDNN ragged-Stats limitation predates this PR, but switching previously working low-level default calls to that path introduces this regression.
2. **P2: single-request CUDA Graph capture now fails when offsets are omitted.** The synthesized `torch.tensor([...], device=q.device)` performs a CPU-to-CUDA copy during capture. On H100, the base captures and replays correctly after input changes; this head raises a RuntimeError during capture, despite warmup and `is_cuda_graph_compatible=True`.

Validation on this head:

- H100: `test_cudnn_prefill_lse_layout.py`, `test_cudnn_prefill_token_indptr.py`, and `test_cudnn_prefill.py`: **441 passed, 576 skipped**.
- SM100: `test_cudnn_prefill_lse_layout.py` and `test_auto_backend_upgrade.py`: **51 passed**.
- Additional probes covered the failures above, padded/packed switching and a two-query-token control, which passes for both KV layouts and offset units.
- Environment: PyTorch `2.13.0+cu130`, official cudnn-frontend `1.29.0`, cuDNN backend `92600` (local `9.26.0.39.dev64544457` build). These are correctness checks; no performance claim.
- The current failing GitHub summary says **“CI skipped — pending authorization”**; it is not evidence of a test assertion failure.

Please add regression coverage for single-token GQA Stats and the single-request/no-offset capture path. The supplied tests pass but miss both cases.

Yang's codex bot

YangXu1990uiuc:
Follow-up on my review of `d54ada9c`: **downgrading the single-token GQA LSE finding from P1 to P2**, with the existing workaround made explicit.

The broader context is the same known cuDNN single-token GQA ragged-Stats issue already tracked in [#5245](https://github.com/flashinfer-ai/flashinfer/pull/5245). That PR provides these protections in `BatchPrefillWithRaggedKVCacheWrapper`:

- `backend="auto"` routes `max_qo_len == 1` GQA steps away from cuDNN.
- Explicit `backend="cudnn"` rejects `return_lse=True` for that combination; output-only calls remain supported.
- It deliberately leaves the low-level `cudnn_batch_prefill_with_kv_cache` API unguarded, with a strict xfail monitoring the backend behavior. The paged wrapper is outside that guard's scope.

At the time of this update, #5245 is still open and its guard is not in this PR's head. Once that workaround lands, the ragged wrapper's affected serving path is protected. My #5161 finding is narrower: changing the **low-level default from padded to ragged Stats**, and enabling packed LSE for the paged path, exposes affected older cuDNN builds at entry points outside that protection. The numerical reproducer used cuDNN `9.26.0.39.dev64544457`; it establishes the behavior of that build.

The remaining ask is therefore **P2 compatibility handling for supported affected cuDNN versions** at those entry points: a safe Stats fallback/guard, or version-gated enablement once a fixed runtime can be required, plus regression coverage. The kernel fix, the #5245 wrapper workaround, and this default-layout compatibility issue should be tracked separately.

The other **P2 CUDA Graph capture regression** remains independent: constructing the implicit single-request Stats offsets copies an unpinned CPU tensor to CUDA during capture. #5245's GQA guard does not address that path. I am retaining the request-changes review while these scoped issues are addressed, and have updated the original LSE inline comment and review summary to P2.

Yang's codex bot

YangXu1990uiuc (approved):

aleozlx:
## PR Review Screening

**CI verdict:** ✅ auto-run ok
**Review category:** `human` (rule fired: 2 — C1.2 changes an existing public API's documented return shape)
**Blocking checks:** S2 (default template overwritten), S3 (template obligations unmet)
**Release blocker:** no
**Early stop:** no

### Security
| Q | Answer | Evidence |
|---|--------|----------|
| S2 template overwritten | ❗ yes | Custom `## Summary` / `## Why` / `## Change` / `## Tests`; all five template headers absent |
| S3 template obligations | ❗ unmet | 0 checkboxes, no Related Issues / Reviewer Notes — unmet formally, met in substance (tests listed, per-GPU results given) |

### Packaging
| Q | Answer | Evidence |
|---|--------|----------|
| C1.1 dependency bump | no | No `supply`-bucket files; no trust-changing lines |
| C1.2 public API changes | yes — extension of existing shape | see below |
| C1.3 AOT / trace registration | n-a | No new `gen_*_module()`; `cudnn_batch_prefill_trace` already declares `lse` as `["num_tokens", "num_heads_qo"]` |

- `flashinfer/cudnn/prefill.py::cudnn_batch_prefill_with_kv_cache` (`@flashinfer_api`, exported from `flashinfer.cudnn`) — **signature unchanged**; `return_lse=True, lse=None` now returns packed `(total_qo_tokens, num_heads_qo)` instead of padded `(batch_size, max_token_per_sequence, num_heads_qo)`; a passed-in `lse` may be packed or padded (new dtype/device/contiguity `ValueError`s); `batch_offsets_stats` derived when omitted, rejected with a padded `lse` → **extension of existing shape**
- `flashinfer/prefill.py::BatchPrefillWithRaggedKVCacheWrapper` — private `_cudnn_stats_offsets` dropped from `__init__` / `plan()` / `run()`; public signatures unchanged → **internal-only**

### Presentation
| Q | Answer | Evidence |
|---|--------|----------|
| C2.1 perf claim backed by data | n-a | No perf claim in title or body |

### Implementation
| Q | Answer | Evidence |
|---|--------|----------|
| C3.1 experimental track declared | no | No `experimental` label, no `@flashinfer_experimental_api`, nothing under `flashinfer/experimental/` |
| C3.2 shared / durable areas | yes | cuDNN graph cache key (`_sdpa_prefill_key_fn`) + ragged-wrapper plumbing; no `include/`, `csrc/` or base-class change |
| C3.3 tests match behavior change | yes | `tests/attention/test_cudnn_prefill_lse_layout.py` (packed/padded × both offset units × causal, numeric reference, rejection cases, wrapper vs default backend); collected by the default `TEST_PATH=tests/` lane |
| C3.4 critical fix / release blocker | no | Layout-contract fix; no hang/crash/corruption claim, no linked regression issue |

### Experimental track
| Q | Answer | Evidence |
|---|--------|----------|

**Notes for the maintainer**
- Behavior change on a public surface: low-level callers using `return_lse=True, lse=None` get a 2-D LSE where they got 3-D; the body states vLLM/SGLang discard it. Worth a release note.
- Validation was on cuDNN 9.27 dev + cudnn-frontend `develop` (1.29 with two unreleased patches) while `requirements.txt` floors `nvidia-cudnn-frontend>=1.28.0`; the new tests skip wholesale when cudnn-frontend is absent, so confirm the CI image actually exercises the ragged-Stats path.
- Removes the wrapper-side `_cudnn_stats_offsets` plumbing introduced by #5133 — the two need to land in order.

<sub>Generated by flashinfer-pr-screen · rubric: docs/code_review_guidance.md · not a code
review · AI screening can make mistakes — a maintainer's judgment supersedes this
report.</sub>

aleozlx (approved):

aleozlx:
@flashinfer-bot run tests/attention

flashinfer/cudnn/prefill.py
egilliam-nv: Reproduced on H100 with cuDNN 9.26.0 and our 9.27 dev build (18/24 NaN, ragged and paged), and confirmed the cause on the cuDNN side: the s_q == 1 decode-class kernel used the sequence stride for the per-head tile rows of a ragged Stats store. Your fix MR !4496 is in cuDNN dev at version 9.28.0, so the low level now gates on `backend_version() < 92800`.

Fixed in 94c5755d: for `max_token_per_sequence == 1` with `h_qo != h_kv` on affected versions, the packed `(tokens, h)` buffer is bound as the padded `(b, 1, h)` Stats form (no ragged offset). The two are byte-identical when every request has exactly one token, so the fixed-stride store writes it correctly and the caller still gets the packed tensor back; no extra pass. Covers ragged and paged KV. The one corner it cannot serve is a batch mixing zero-length and single-token requests; that raises and asks for a padded `lse`. Test: `test_cudnn_prefill_lse_packed_single_token_rows` (GQA + MHA × ragged + paged) with a NaN-poisoned buffer against an fp32 reference; your strict xfail `test_cudnn_single_token_gqa_lse_is_correct` now passes, so I turned it into a plain correctness test. I left the #5245 wrapper guards in place; with the low level safe they could be relaxed in a follow-up if you want explicit `backend="cudnn"` + `return_lse` allowed for these rows.

flashinfer/cudnn/prefill.py
egilliam-nv: Reproduced (same error at capture). Fixed in 94c5755d: the implicit single-request Stats offsets are built on the device as `arange(2) * n`, no host tensor involved. Regression test `test_cudnn_prefill_lse_single_request_no_offsets_cuda_graph` warms up on a side stream, captures with all offsets omitted, replays, and checks the LSE against an fp32 reference.

flashinfer/cudnn/prefill.py
```diff
@@ -998,6 +1028,33 @@ def cudnn_batch_prefill_with_kv_cache(
...
+                    ],
+                    dtype=torch.int32,
+                    device=q.device,
+                )
+            elif batch_offsets_units == "tokens":
+                batch_offsets_stats = batch_offsets_q
+            else:
+                batch_offsets_stats = (
+                    torch.div(batch_offsets_q, q.stride(0), rounding_mode="floor")
+                    * h_qo
```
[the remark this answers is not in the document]
egilliam-nv: Done in 94c5755d: the `batch_offsets_q` / `batch_offsets_o` docstrings now state element offsets in each tensor's own storage (`token_offset * q.stride(0)`, resp. `* out.stride(0)`, which reduce to `num_heads * head_dim` only for contiguous tensors), and the missing-offset error gives the same formula.

flashinfer/cudnn/prefill.py
```diff
@@ -1050,16 +1063,18 @@ def cudnn_batch_prefill_with_kv_cache(
...
+        omitted only for ``batch_size == 1``.
     batch_offsets_o : Optional[torch.Tensor]
         Cumulative per-request start offsets into the packed output tensor,
         shape ``(batch_size + 1,)``, int32, on GPU, in the units given by
-        ``batch_offsets_units`` (element offsets are
-        ``token_offset * num_heads_qo * head_dim_vo``).  Required when
-        ``batch_size > 1`` on the cuDNN graph path; with
+        ``batch_offsets_units`` (element offsets count elements of the output's
+        own storage: ``token_offset * out.stride(0)``, i.e.
+        ``token_offset * num_heads_qo * head_dim_vo`` for a contiguous output).
```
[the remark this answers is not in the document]
egilliam-nv: Done in e86f4c44: a caller-provided `out` must be contiguous and on `q`'s device, checked at the public entry before the graph binds it (O is declared with contiguous `(tokens, h_qo, d_vo)` strides).

egilliam-nv:
e86f4c44 also takes CodeRabbit's outside-the-diff suggestion, which matches what I flagged on the P2 thread: with the low level now serving the packed LSE for single-token GQA on every supported cuDNN, the #5245 wrapper guards narrow to the one shape it cannot serve below 9.28, a batch that mixes zero-length and one-token requests (there the padded `(b, 1, h)` form is no longer the packed layout). Concretely:

- `auto` keeps cuDNN for ordinary single-token GQA rows (`single_token_gqa` → `single_token_gqa_with_empty_rows`, also gated on the cuDNN version), and the LSE matches fa2.
- explicit `backend="cudnn"` + `return_lse=True` is refused only for the empty-row batch, with a message that says why; output-only calls are unchanged.
- `test_auto_declines_single_token_gqa_rows` / `test_explicit_cudnn_refuses_single_token_gqa_lse` become `test_auto_takes_cudnn_for_single_token_gqa_rows` / `test_single_token_gqa_with_empty_rows_stays_off_cudnn_lse` (the latter skips on cuDNN ≥ 9.28).

@YangXu1990uiuc since these are your guards from #5245: shout if you would rather keep `auto` off cuDNN for single-token GQA regardless (e.g. for reasons beyond the LSE bug) and I will revert that part. sm100 `test_auto_backend_upgrade`: 36 passed; H100 cuDNN suites: 552 + 32 passed. The earlier `/bot run` was against 94c5755d's predecessor, so it will need another trigger.

saltyminty (approved):
LGTM but seems like there is still pending discussion/decision

flashinfer/cudnn/prefill.py
egilliam-nv: Leaving this thread open on purpose: the one decision still pending is the version cutoff. The gate (and the narrowed #5245 wrapper guards in e86f4c44) currently say cuDNN 9.28 (`backend_version() < 92800`), where MR !4496 landed. We may cherry-pick that fix into a 9.27 point release; if so the cutoff becomes that build and the 9.28 mentions get swept. Merge should wait for that call; everything else on the PR is settled.

PR #5172, opened 2026-09-12
fix(comm): fence DCP receiver unpack before async output copy

klshuster:
## 📌 Description

Order the LL128 receiver's shared-memory unpack stores before its asynchronous shared-to-global output copy.

In `helixAllToAllKernel`, `LL128Proto::protoUnpack` restores payload words over protocol flags using generic shared-memory stores. Its `__syncwarp()` orders the warp's generic accesses, but does not establish visibility to the async proxy used by `cp_async_bulk_s2g` in `s2gAllFields`.

Add `fence.proxy.async.shared::cta` on each receiver lane, then synchronize the warp before lane 0 issues the bulk copy. This follows the [PTX generic/async proxy ordering requirement](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#asynchronous-data-movement-async-proxy). No block-wide barrier is introduced into the divergent sender/receiver paths.

## 🔍 Related Issues

The affected kernel was introduced in #2951. No separate issue filed.

### ✅ Pre-commit Checks

Ran `pre-commit run --files csrc/nv_internal/tensorrt_llm/kernels/helixAllToAll.cu`: all applicable checks passed. The full-repository hook sweep was not run.

## 🧪 Tests

No standalone failing reproducer or new unit test is included. No full upstream-suite pass is claimed. A compact reproducer that reliably exposes the missing proxy ordering is still needed for a red/green regression test.

## Reviewer Notes

This is a four-line memory-ordering correctness change, not a performance claim. Please focus on generic-to-async shared-memory ordering and the all-lane fence followed by the warp rendezvous. The rationale above is based on the public kernel implementation and PTX specification; no private workload data or logs are attached.

AI-assisted patch preparation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved data synchronization during communication operations on newer NVIDIA GPU architectures, helping ensure received data is read correctly.

### DIFF
--- a/csrc/nv_internal/tensorrt_llm/kernels/helixAllToAll.cu
+++ b/csrc/nv_internal/tensorrt_llm/kernels/helixAllToAll.cu
@@ -454,6 +454,10 @@ __global__ void helixAllToAllKernel(HelixAllToAllParams params) {
       LL128Proto::protoUnpack(shmem, tail, singlePacked128ByteCount, fifoEntry128ByteIndexBase,
                               loaded128ByteCount, laneId);
 
+#if defined(__CUDA_ARCH__) && __CUDA_ARCH__ >= 900
+      asm volatile("fence.proxy.async.shared::cta;" : : : "memory");
+      __syncwarp();
+#endif
       // note: fields are already unpacked in shared memory
       s2gAllFields<ALLOW_VARIABLE_FIELD1>(params.recvFields, dataIndex, shmem, laneId);
       // wait for data to be read from shared memory


bkryu:
@flashinfer-bot run

bkryu:
/bot run tests/comm/test_dcp_alltoall.py tests/comm/test_mnnvl_dcp_alltoall.py

aleozlx:
## PR Review Screening

**CI verdict:** ✅ auto-run ok
**Review category:** `human` (rule fired: C3.2 — shared kernel internals; C3.3 inadequate keeps it at `human`, with the reviewer deciding what verification to require)
**Blocking checks:** S3 (partial)
**Release blocker:** 🚨 candidate — missing generic→async-proxy ordering between the LL128 shared-memory unpack and the bulk S2G copy is a wrong-results/corruption class defect in the DCP all-to-all receiver; kernel introduced in #2951 (C3.4)
**Early stop:** no

### Security
| Q | Answer | Evidence |
|---|--------|----------|
| S2 template overwritten | no | — |
| S3 template obligations | ❗ partial | four of five boxes unticked; pre-commit was run scoped to the changed file rather than `--all-files`, and both Tests items are unmet — accurately, since no reproducer or regression test is included and the body says so |

### Packaging
| Q | Answer | Evidence |
|---|--------|----------|
| C1.1 dependency bump | no | no supply-bucket changes |
| C1.2 public API changes | no | kernel-internal only; no signature, export or binding change |
| C1.3 AOT/trace registration | no | `aot.py` untouched; no module generators |

### Presentation
| Q | Answer | Evidence |
|---|--------|----------|
| C2.1 perf claim backed | n-a | no performance claim — stated explicitly as a correctness change |

### Implementation
| Q | Answer | Evidence |
|---|--------|----------|
| C3.2 durable areas touched | yes — durable | `csrc/nv_internal/tensorrt_llm/kernels/helixAllToAll.cu`, inside `helixAllToAllKernel`'s receiver path. No abstraction or dispatch convention is reshaped, so this routes `human` rather than `live` |
| C3.3 tests match change | ❗ no — *executes but does not cover* | `tests/comm/test_dcp_alltoall.py` and `tests/comm/test_mnnvl_dcp_alltoall.py` exercise this kernel on the multi-GPU lanes, so the path is not uncovered — but neither asserts the ordering property, and a race of this shape can pass indefinitely while still being wrong. The body acknowledges no reproducer exists |
| C3.4 release-blocker candidate | 🚨 yes | changes shipped library code; no accompanying regression test, so this is the fix alone rather than the fix-plus-test combination |

### Experimental track
| Q | Answer | Evidence |
|---|--------|----------|

**Notes for the maintainer**

- **The exposure is wider than the usual blocker.** #2951 merged 2026-04-23 and is contained in `v0.6.18`, `v0.6.18.post1`, `v0.7.0rc1` and `v0.7.0rc2` — so this is not a main-only defect gating the next release; it is live in the current release and may warrant a patch alongside the 0.7.0 fix.
- **This is vendored code.** The file sits under `csrc/nv_internal/tensorrt_llm/`, so a local-only fix diverges from upstream and the next vendor sync can silently revert it. Worth confirming whether the same ordering fix is going to TRT-LLM, and recording the answer here either way.
- **Green CI will not settle this one.** The GitHub lanes currently passing are single-GPU (A10G / T4) and do not exercise an all-to-all; GitLab pipeline #67832418 was created minutes ago and has not reported. When it does, a pass demonstrates no regression — it does not demonstrate the race is fixed, because nothing asserts the ordering. The decision in front of a reviewer is whether the PTX-specification argument is sufficient on its own, which is a judgement call rather than something CI can answer.

saltyminty (approved):

saltyminty:
Looks good to me, but would recommend running benchmarks prior to merging

bkryu:
@saltyminty, will merge the PR because it fixes a functional bug, not a performance issue. Seems like a minimal fix that will be the optimal local fix.

PR #5178, opened 2026-09-13
feat(comm): add H5120 MNNVL CuTe DSL profiles (TP4/TP8) and an apply_rms_norm switch

zyongye:
## 📌 Description

This PR extends the Blackwell MNNVL CuTe DSL all-reduce fusion backend (added in #4358) to a second static shape — `hidden_size=5120`, bf16, `top_k` 6 and 3 — for **TP4 and TP8**, and adds a compile-time switch that removes the fused RMSNorm from the epilogue.

No kernel algorithm changes: the existing LL/BT/HT protocols are reused as-is. The new shape is new tuning presets plus measured routing tables, which is what the tuning dataclasses are designed for — they carry launch knobs only, while `hidden_size` and `top_k` reach the kernels through `StaticProfile`.

### Profiles

New LL, BT and HT presets for `H5120` with routing tables mapping `M` ranges onto them. Two knobs are shape-derived rather than copied:

- **LL `cluster_size=5`.** A 5120-wide token is 640 bf16x8 fragments; `fragment_stride = cluster_size * threads / rank_lanes = 5 * 128 = 640`, so the Lamport RMSNorm covers one token in a single fully utilised trip. The H8192 `cluster_size` of 8 would leave 37% of the lanes idle here.
- **`prefetch_group` sized to `top_k`**, so the staged rmem tensor holds no unused slots.

**HT is structurally unreachable at `hidden=5120` for `tp >= 8`.** A token is 640 bf16x8 packs, so `packs_per_reduction_shard = 640/tp`: 160 at tp=4 (a multiple of one 32-thread reduction warp), but 80 at tp=8 and 40 at tp=16, neither of which divides evenly across any legal reduction-warp count. An exhaustive sweep of the 1080 legal knob combinations finds none. The TP8 routes therefore stay on BT all the way up, and only TP4 carries HT presets. Lifting this would need a predicated tail in the HT reduction loop — a kernel change, not a tuning change.

### Routing

Every boundary below is measured with the benchmark added in this PR, on NVIDIA GB200 — TP4 on one node, TP8 across two nodes over multi-node NVLink. Each bound is the largest swept `M` at which the lower-`M` protocol was still ahead.

| Path | TP | LL | BT | HT |
|---|---:|---:|---:|---:|
| finalize + AR + residual + RMSNorm, `top_k=6` | 4 | `M <= 42` | `43-56` (P0), `57-1024` (P1) | `M >= 1025` |
| finalize + AR + residual + RMSNorm, `top_k=3` | 4 | `M <= 38` | `39-40` (P0), `41-2048` (P1) | `M >= 2049` |
| AR + residual + RMSNorm | 4 | `M <= 36` | `37-192` (P0), `193-2048` (P1) | `M >= 2049` |
| finalize + AR + residual + RMSNorm, `top_k=6` | 8 | `M <= 24` | `25-64` (P0), `M >= 65` (P1) | — |
| finalize + AR + residual + RMSNorm, `top_k=3` | 8 | `M <= 24` | `25-32` (P0), `M >= 33` (P1) | — |
| AR + residual + RMSNorm | 8 | `M <= 20` | `21-512` (P0), `M >= 513` (P1) | — |

Two boundaries needed more key structure than the H8192 profiles use, and the measurements are why:

- **The finalize BT→HT crossover is keyed by `(tp, top_k)`.** HT hides the `top_k` gather in its producer warps, so its edge over BT arrives at `M=1024` for `top_k=6` but not until `M=2048` for `top_k=3`. A single tp-keyed bound would route `top_k=3` to HT ~1000 tokens too early (13% slower at `M=1536`).
- **The BT preset split is keyed per pattern**, as in the H8192 profiles. The finalize path leaves the narrow 2-element/256-thread tiling as soon as the gather dominates, while the plain all-reduce stays on it about four times longer (TP4: 56 vs 192). Sharing one number costs up to 8% on finalize.

### `apply_rms_norm`

A new compile-time workspace argument that drops the normalisation from the epilogue:

```python
MNNVLCuteDSLAllReduceFusionWorkspace(..., apply_rms_norm=False)
```

The chain then stops after finalize, shared-expert add, all-reduce and the optional residual add, and materialises that value into `residual_out`. This is for callers that need the un-normalised reduced tensor and would otherwise pay for a norm they discard.

The guards are `const_expr`, so the sum-of-squares, `rsqrt`, gamma multiply, norm store and LL's cluster-wide sum barrier are not emitted at all rather than branched around. Contract, mirroring the existing `add_residual`/`residual_in` rule:

- requires `write_residual_output=True` — otherwise the fusion would produce no output at all;
- `rms_gamma` and `norm_out` must be `None`;
- HT raises `NotImplementedError`: its norm lives in a dedicated warp group with its own named barriers and shared staging, so it is not a guardable block the way the LL and BT tails are.

Default is `True`, so existing callers are unchanged.

## 🔍 Related Issues

Follow-up to #4358, which added the backend and the H8192 TP8/TP16 profiles.

## Performance

NVIDIA GB200, bf16, `hidden=5120`. TP4 is one node; TP8 spans two nodes over multi-node NVLink. Exclusive nodes. Timed with `bench_gpu_time(use_cuda_graph=True, num_iters_within_graph=20)`, reporting the median over replays of the max across ranks — these are collective kernels, so per-iteration CUPTI or event timing measures rank launch skew, not kernel time (a 4 us LL all-reduce measures at ~235 us that way). "fused" is the protocol the shipped `DEFAULT_CONFIG` selects; "unfused" is a torch MoE finalize + NCCL all-reduce + FlashInfer `fused_add_rmsnorm`. All times in microseconds.

#### MoE finalize + shared-expert add + AR + residual + RMSNorm, `top_k=6`

| `M` | TP4 fused | TP4 unfused | TP4 speedup | TP8 fused | TP8 unfused | TP8 speedup |
|---:|---:|---:|---:|---:|---:|---:|
| 1 | 4.09 (LL) | 75.78 | 18.51x | 4.81 (LL) | 88.57 | 18.42x |
| 2 | 4.30 (LL) | 86.21 | 20.07x | 4.90 (LL) | 99.30 | 20.27x |
| 4 | 4.30 (LL) | 95.29 | 22.17x | 4.91 (LL) | 110.33 | 22.47x |
| 8 | 4.30 (LL) | 113.68 | 26.46x | 5.22 (LL) | 130.20 | 24.95x |
| 16 | 4.81 (LL) | 152.30 | 31.68x | 6.04 (LL) | 166.14 | 27.52x |
| 32 | 5.83 (LL) | 96.37 | 16.53x | 7.57 (BT) | 111.46 | 14.73x |
| 64 | 7.66 (BT) | 99.78 | 13.03x | 8.29 (BT) | 113.57 | 13.70x |
| 128 | 9.31 (BT) | 108.28 | 11.63x | 9.93 (BT) | 121.70 | 12.26x |
| 256 | 13.00 (BT) | 151.19 | 11.63x | 13.72 (BT) | 154.52 | 11.27x |
| 512 | 20.32 (BT) | 242.68 | 11.94x | 21.50 (BT) | 271.72 | 12.64x |
| 1024 | 42.03 (BT) | 368.19 | 8.76x | 41.52 (BT) | 426.40 | 10.27x |
| 2048 | 72.10 (HT) | 693.88 | 9.62x | 83.86 (BT) | 719.97 | 8.59x |
| 4096 | 128.35 (HT) | 1305.18 | 10.17x | 147.66 (BT) | 1388.71 | 9.41x |
| 8192 | 240.37 (HT) | 2534.64 | 10.54x | 304.85 (BT) | 2715.85 | 8.91x |

#### MoE finalize + shared-expert add + AR + residual + RMSNorm, `top_k=3`

| `M` | TP4 fused | TP4 unfused | TP4 speedup | TP8 fused | TP8 unfused | TP8 speedup |
|---:|---:|---:|---:|---:|---:|---:|
| 1 | 3.99 (LL) | 51.05 | 12.80x | 4.81 (LL) | 64.46 | 13.41x |
| 2 | 4.09 (LL) | 57.25 | 13.99x | 4.91 (LL) | 70.10 | 14.28x |
| 4 | 4.19 (LL) | 62.04 | 14.80x | 5.01 (LL) | 76.02 | 15.16x |
| 8 | 4.40 (LL) | 70.03 | 15.92x | 5.08 (LL) | 85.74 | 16.86x |
| 16 | 4.91 (LL) | 88.64 | 18.05x | 5.93 (LL) | 102.24 | 17.23x |
| 32 | 6.02 (LL) | 62.42 | 10.37x | 7.37 (BT) | 76.76 | 10.42x |
| 64 | 7.06 (BT) | 64.97 | 9.20x | 8.19 (BT) | 77.96 | 9.52x |
| 128 | 8.80 (BT) | 71.01 | 8.07x | 9.42 (BT) | 83.60 | 8.88x |
| 256 | 12.49 (BT) | 97.49 | 7.81x | 13.21 (BT) | 103.18 | 7.81x |
| 512 | 19.76 (BT) | 152.21 | 7.70x | 20.88 (BT) | 173.92 | 8.33x |
| 1024 | 36.09 (BT) | 234.22 | 6.49x | 38.14 (BT) | 291.61 | 7.65x |
| 2048 | 68.26 (BT) | 435.39 | 6.38x | 75.32 (BT) | 462.07 | 6.13x |
| 4096 | 127.94 (HT) | 793.26 | 6.20x | 136.82 (BT) | 875.81 | 6.40x |
| 8192 | 238.59 (HT) | 1522.96 | 6.38x | 292.99 (BT) | 1692.72 | 5.78x |

#### AR + residual + RMSNorm

| `M` | TP4 fused | TP4 unfused | TP4 speedup | TP8 fused | TP8 unfused | TP8 speedup |
|---:|---:|---:|---:|---:|---:|---:|
| 1 | 3.89 (LL) | 19.11 | 4.92x | 4.60 (LL) | 32.62 | 7.09x |
| 2 | 4.06 (LL) | 20.73 | 5.10x | 4.71 (LL) | 33.26 | 7.07x |
| 4 | 4.19 (LL) | 20.81 | 4.96x | 4.81 (LL) | 34.76 | 7.23x |
| 8 | 4.20 (LL) | 19.92 | 4.75x | 5.01 (LL) | 35.14 | 7.01x |
| 16 | 4.71 (LL) | 20.55 | 4.37x | 5.93 (LL) | 33.99 | 5.73x |
| 32 | 5.55 (LL) | 20.30 | 3.66x | 6.96 (BT) | 34.10 | 4.90x |
| 64 | 6.45 (BT) | 21.46 | 3.33x | 7.37 (BT) | 34.04 | 4.62x |
| 128 | 8.08 (BT) | 23.98 | 2.97x | 8.75 (BT) | 36.19 | 4.14x |
| 256 | 11.82 (BT) | 29.77 | 2.52x | 12.45 (BT) | 40.21 | 3.23x |
| 512 | 18.89 (BT) | 47.34 | 2.51x | 20.48 (BT) | 69.19 | 3.38x |
| 1024 | 34.56 (BT) | 61.82 | 1.79x | 35.68 (BT) | 110.38 | 3.09x |
| 2048 | 64.15 (BT) | 116.42 | 1.81x | 68.65 (BT) | 139.56 | 2.03x |
| 4096 | 121.29 (HT) | 158.19 | 1.30x | 131.35 (BT) | 238.90 | 1.82x |
| 8192 | 225.93 (HT) | 273.44 | 1.21x | 287.38 (BT) | 441.14 | 1.54x |

Summary over the swept range:

| Semantics | TP | vs. unfused baseline |
|---|---:|---:|
| With finalize, `top_k=6` | 4 | `8.76-31.68x` faster |
| With finalize, `top_k=6` | 8 | `8.59-27.52x` faster |
| With finalize, `top_k=3` | 4 | `6.20-18.05x` faster |
| With finalize, `top_k=3` | 8 | `5.78-17.23x` faster |
| Without finalize | 4 | `1.21-5.10x` faster |
| Without finalize | 8 | `1.54-7.23x` faster |

LL leads decode, BT covers the middle, and at TP4 persistent HT takes over into prefill. At TP8 LL's per-rank mailbox traffic grows with `tp`, which is why its crossovers sit at much smaller `M` than TP4 (24 vs 42 for finalize) and why it degrades sharply at large `M` (955 us vs BT's 305 us at `M=8192`).

### Cost of the fused RMSNorm

Measured by compiling the same chain with and without the norm (GB200, TP4, `top_k=6`), which is what `apply_rms_norm` now exposes:

| | `M <= 128` | `M >= 2048` |
|---|---|---|
| LL | `0.82 us` — 12.8% of the chain | `4.63 us` — 1.6% |
| BT | `0.47 us` — 7.0% of the chain | `6.04 us` — 3.3% |

Below `M ~ 128` the norm is a near-constant fixed cost — the block reduction and `rsqrt`, latency rather than bandwidth — so it is proportionally largest exactly in the decode regime where LL is selected. Note that reading a profile is misleading here: BT's `MaterializeRMSNorm` kernel is 37-41% of the chain's GPU time, but most of that is the remote prenorm-mailbox read and the `residual_out` write that a norm-free variant would still have to do.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

`tests/comm/test_mnnvl_cutedsl_no_rms_norm.py` is new. It runs at **four or eight** ranks against `hidden=5120`, unlike `test_mnnvl_cutedsl_numerical_contract.py`, which is pinned to `hidden=8192` and eight or sixteen ranks — so this shape gets coverage in a 4-GPU lane. It asserts that with the norm disabled `residual_out` is bitwise equal to finalize + shared-expert add + all-reduce, and that each rejected operand combination raises. The HT-rejection case skips where no HT profile exists, since the workspace then fails earlier for a different reason.

Validated on GB200 at TP4 (one node) and TP8 (two nodes, multi-node NVLink):

| Check | TP4 | TP8 |
|---|---|---|
| `test_mnnvl_cutedsl_no_rms_norm.py` | 5 passed | 4 passed, 1 skipped (no HT profile at tp8) |
| Protocol-pinned numerics, both patterns, LL/BT/HT | ALL PASS | ALL PASS |
| `DEFAULT_CONFIG` routing, `M` straddling every boundary | ALL PASS | ALL PASS |
| `tests/comm/test_mnnvl_cutedsl_config.py` | 18 passed | — |

The numerics checks apply the same per-protocol contracts as #4358: LL and HT prenorm bitwise equal, BT within 1 BF16 ULP, norms within 1-2 ULP, with HT compared against `multimem_one_shot_all_reduce_out` to match its multicast reduction order.

## Reviewer Notes

The routing tables and their measured crossovers are the substance here; the kernels are unchanged apart from the `apply_rms_norm` guards. Worth a close look:

- the `(tp, top_k)` key on the finalize BT→HT bound and the per-pattern BT split — both are departures from the H8192 tables, justified by the measurements above;
- the HT divisibility argument for `tp >= 8`, which is why TP8 has no HT route;
- the `apply_rms_norm` operand contract, and the placeholder gamma / `residual_out`-as-`norm_output` substitution used to satisfy the compiled kernel signature when the norm is guarded out.

One consequence worth naming: a process that uses both `apply_rms_norm` settings compiles two specialisations per protocol.

## Summary by CodeRabbit

- **New Features**
  - Added optimized GB300 support for hidden size 5120 across tensor-parallel configurations and top-k 3/6 workloads.
  - Added automatic routing across low-, mid-, and high-token-count execution paths for improved performance.
  - Added an option to skip RMS normalization while still producing residual outputs.
  - Added tuning presets for all-reduce and MoE finalization operations.

- **Benchmarks**
  - Added coverage to compare communication protocols, fused operations, and reference implementations across token counts.

- **Tests**
  - Added validation for unnormalized outputs and invalid option combinations.


qiangyicheng:
What about adding a dedicated NO_NORM_CONFIG? Some recent models using iHC put Norm in another kernel, so they might prefer a kernel without Norm.
For H6144, top-k8, BF16, TP8 on B200, our no-norm finalize benchmarks suggest LL → BT around M24–28, and BT P0 → P1 around M64. With our experimental HT no-norm implementation, BT → HT is around M1152–1280.

jiahanc (approved):
LGTM

flashinfer/comm/mnnvl_cutedsl_ar.py
```diff
@@ -463,12 +495,16 @@ def _mnnvl_cutedsl_allreduce_fusion(
                 device=device,
                 alignment=16,
             )
+        # With the norm compiled out the kernel writes only
+        # residual_output; passing it as norm_output too satisfies the
+        # signature and is never written through.
+        norm_target = norm_out if workspace.apply_rms_norm else residual_out
```
jiahanc, line 520: The wrapper passes None for both output slots. Each protocol's `_outputs()` allocates separate norm and residual tensors. The no-RMSNorm kernel writes only the residual tensor, but the wrapper returns the norm tensor and discards the residual result. This silently returns uninitialized data rather than the reduced value.

zyongye: Good catch — confirmed and fixed in 169a2950.

You traced it exactly: when the caller omits `residual_out`, the wrapper passed `None` for both slots, `_outputs()` allocated two independent `torch.empty` tensors, the kernel filled only the residual one, and the wrapper returned the other. Uninitialised memory, silently.

The fix allocates the residual slot in the wrapper when it is missing, hands the same tensor to `norm_output`, and returns whichever slot the compiled kernel actually writes.

Worth noting why the existing tests missed it: they all passed `residual_out` explicitly, which aliased the two slots onto one buffer and masked the bug. Added `test_disabled_norm_returns_reduce_when_output_not_preallocated` which omits it — it fails on the old code for both LL and BT and passes now.

Re-validated on GB200, TP4 (one node) and TP8 (two nodes over MNNVL):

| Check | TP4 | TP8 |
|---|---|---|
| `test_mnnvl_cutedsl_no_rms_norm.py` | 7 passed | 6 passed, 1 skipped (no HT profile at tp8) |
| Protocol-pinned numerics, norm ON | ALL PASS | ALL PASS |
| `DEFAULT_CONFIG` routing, norm ON | ALL PASS | ALL PASS |

zyongye:
Added `NO_NORM_CONFIG` in 53aa2405, plus a fix for the uninitialised-output bug @jiahanc found.

@qiangyicheng you were right that the norm-free path needs its own table. Mine reused the norm-on boundaries, and they do move — dropping the norm frees LL of a cluster-wide barrier while BT only loses an output store, so LL stays ahead further. Measured on GB200 (tp=4 one node, tp=8 two nodes over MNNVL):

| route | norm-on | norm-free |
|---|---:|---:|
| tp4 finalize k=6, LL→BT | 42 | 64 |
| tp4 finalize k=3, LL→BT | 38 | 56 |
| tp4 all-reduce, LL→BT | 36 | 40 |
| tp4 all-reduce, BT split | 192 | 128 |
| tp8 finalize k=6, LL→BT | 24 | 28 |
| tp8 finalize k=6, BT split | 64 | 96 |
| tp8 finalize k=3, BT split | 32 | 48 |

Same direction as your H6144/top-k8/TP8 numbers, so the effect looks shape-independent.

`MNNVLCuteDSLConfig` gains `applies_rms_norm`, and a workspace rejects a mismatched pairing in **either** direction — a norm-on config with `apply_rms_norm=False` was exactly the silent mis-routing you were pointing at. The protocol-pinned `*_ONLY` configs carry `None` and suit either, since their ranges force a protocol rather than describe a crossover.

**HT is deliberately out of scope here, and needs a separate change.** These routes are LL → BT only. At tp ≥ 8 HT is structurally unreachable at hidden 5120 (640/8 packs per reduction shard divides no legal reduction-warp count). At tp = 4 it is reachable and `DEFAULT_CONFIG` uses it above M=1024, but there is no norm-free HT kernel — its RMS warp group owns the sum-of-squares and its barriers, so removing the norm means restructuring the warp specialisation rather than adding a `const_expr` guard. That is unimplemented, not impossible, and it costs a tp=4 norm-free caller ~11% at M=2048 and ~25% at M=8192 against the norm-on HT route.

Since you mentioned an experimental HT no-norm implementation — would you rather land that yourself as a follow-up? I would sooner not duplicate it, and the tp=4 BT→HT crossover for the norm-free path still needs measuring either way.

Validated on GB200: 11 passed at tp=4, 10 passed / 1 skipped at tp=8, with norm-on numerics and `DEFAULT_CONFIG` routing unchanged at both widths.

qiangyicheng:
@zyongye Thanks a lot! I will do a no-norm HT PR as a follow up.

bkryu (approved):

PR #4861, opened 2026-09-01
fix: honor per-request seed/offset in sampling kernels

zhaoyangwang-nvidia:
## 📌 Description

The sampling kernels accept `seed`/`offset` as per-request tensors, and the Python layer validates them as length 1 or `batch_size`, but every kernel read only `seed_arr[0]` / `offset_arr[0]` and applied that single value to the entire batch. Rows were separated solely by the Philox subsequence (`blockIdx.x`).

So rows did draw from distinct streams, but no row except the first used the seed its caller asked for. A user-specified per-request seed was unreproducible whenever `batch_size > 1`, and a request's tokens depended on where it happened to land in the batch. Concretely, seeds `[1111, 2222, 3333, 4444]` produced the same output as `[1111, 1111, 1111, 1111]`.

This PR indexes the arrays by `blockIdx.x`, matching the convention already used by the other per-request arrays in the same kernels (e.g. `top_k_arr[bx]`). Seven kernels are affected: `SamplingFromLogitsKernel`, `SamplingFromProbKernel`, `TopKSamplingFromProbKernel`, `TopPSamplingFromProbKernel`, `MinPSamplingFromProbKernel`, `TopKTopPSamplingFromProbKernel` and `ChainSpeculativeSampling`.

A length-1 tensor still broadcasts to every row, so existing callers that pass the tensor spelling of a scalar seed are unaffected. The broadcast is resolved in-kernel via a flag rather than by materializing an expanded tensor, since expanding would allocate on every call and break CUDA graph capture — which is precisely the case per-request seeds are needed for.

The length that makes the indexing safe was previously only checked in Python, so a direct FFI caller could read out of bounds. `validate_seed_offset_tensors` now checks it too (and derives the broadcast flag), and seed/offset tensors of differing lengths are rejected in both layers.

**Scope:** this is deliberately limited to making per-request seed/offset take effect. It does not include the in-kernel atomic offset update / round-based offset tracking from #2345; those are separable and can follow if wanted.

## 🔍 Related Issues

- Closes #1104
- Supersedes #2345 — opened as a new PR at @yzh119's suggestion, since he does not currently have bandwidth to carry that one forward. Credit for the original design and the C++/Python plumbing goes to him; this PR reuses that surface and fixes the kernel-side indexing. Downstream, TensorRT-LLM already plumbs per-row Philox `(seed, offset)` end-to-end and is blocked on this behavior.

## 🚀 Pull Request Checklist

[...]

PR #5196, opened 2026-09-13
perf(sampling): defer logits block reduction

tomalmog:
## Description

`SamplingFromLogitsKernel` currently performs a block reduction and barrier for
every vocabulary tile. This patch keeps a running maximum per vector lane and
reduces once after the loop. For an FP32 vocabulary of 151,936, that replaces
38 block reductions with one.

The RNG, tie ordering, vector-lane reduction layout and padded-output behavior
stay the same. There are no public API changes.

Adds 62 regression cases for masked tile boundaries, int32/int64 indexed rows,
cross-tile ties and CUDA-graph input updates, plus performance and seeded-output
comparison scripts for the public API.

### Performance

Compared with unmodified `93e9eef092c11bc93a6f432a16404054e85d19c5` on A100-SXM4
80GB and H100 80GB, using Torch 2.8.0 / CUDA 12.8 / TVM-FFI 0.1.11 and source JIT
builds targeting SM80/SM90a. FP32 logits, temperature 1, no filtering, scalar
seed/offset. CUDA-graph event timing excludes compilation, allocation and Python
dispatch. Each arm runs sequentially on the same GPU, with 11 timing rounds.

Across B=1/8/32/128/512, V=32000/128256/151936 and hot/128MiB-streamed inputs,
all 60 configurations improve: 8.6-35.3% lower latency on A100 and 21.2-32.0% on H100.
Representative hot-input results:

| GPU | Batch | Vocab | Cache | Before (us) | After (us) | Speedup |
|---|---:|---:|---|---:|---:|---:|
| A100 | 1 | 151936 | hot | 91.989 | 59.972 | 1.534x |
| A100 | 128 | 151936 | hot | 184.047 | 119.569 | 1.539x |
| A100 | 512 | 151936 | hot | 458.445 | 296.482 | 1.546x |
| H100 | 1 | 151936 | hot | 62.293 | 42.775 | 1.456x |
| H100 | 128 | 151936 | hot | 63.399 | 43.780 | 1.448x |
| H100 | 512 | 151936 | hot | 247.218 | 168.175 | 1.470x |

Copy the two added benchmark/check scripts to the baseline checkout as well.
Use an editable/source installation on each revision and separate JIT caches:

```bash
FLASHINFER_WORKSPACE_BASE=/tmp/fi-before \
  python benchmarks/bench_sampling_logits.py --output before.json
FLASHINFER_WORKSPACE_BASE=/tmp/fi-after \
  python benchmarks/bench_sampling_logits.py --output after.json
```

These results apply to the existing logits sampling API. Whole-model throughput
was not measured.

## Related Issues

No tracked issue. Nearby #4861 changes seed resolution; this change leaves that logic untouched.

## Pull Request Checklist

### Pre-commit Checks

## Tests

Affected tests below passed; the full repository suite was not run.

```bash
python -m pytest -q tests/utils/test_sampling_logits_reduction.py \
  tests/utils/test_sampling.py \
  tests/utils/test_logits_processor.py::TestLogitsPipeVsSamplingOps::test_logits_sample \
  -k 'logits and not top_k and not top_p'
```

110 passed, 1878 deselected on **each** GPU. The 62 new tests also pass on the
unmodified revision. Separately, 576 original-versus-patched seeded cases per GPU
(12,672 token comparisons total) agree exactly, covering random/tied/masked logits,
both deterministic modes, indexed rows and scalar/tensor seed inputs.

For the seeded comparison, run `python benchmarks/check_sampling_logits.py before-seeded.json`
on the baseline and the same script with `after-seeded.json` on the patched source,
then compare the JSON files. Compare within the same GPU, not across architectures.

Applicable pre-commit hooks passed on all four changed files. Pre-commit and its
hook environments ran in a container; no host Git hook was installed and full-tree hooks
were not run (the corresponding checklist boxes remain unchecked).

A broader H100 `-k logits` attempt was interrupted while compiling unrelated
top-k routines, after 86 tests passed. That incomplete attempt is not counted as
a passing broad suite. Earlier native-dispatch memcheck/racecheck runs passed on
both GPUs; no full-model benchmark was performed.

## Experimental Track

```experimental-tests
#
#
```

## Reviewer Notes

Please review the tie comparator and preserved vector-lane reduction layout.
GPU CI on the other supported architectures remains pending maintainer triggering.
AI assistance was used for implementation, investigation and validation.


tomalmog:
First PRs here, so I don't have permission to start CI. Could a maintainer run `@flashinfer-bot run` when convenient? This one and #5201 are both small sampling-kernel changes with tests and before/after benchmarks in the descriptions. Happy to rebase either if preferred.

Issue #5192, opened 2026-09-13
SM120: no DSV4 dual-cache prefill template for extra_page_block_size == 32 (blocks DeepSeek-V4.1-Flash)


steven-umt:
### Environment

- **GPU**: NVIDIA RTX PRO 6000 Blackwell, compute capability **12.0** (sm_120), TP4
- **flashinfer-python**: `0.6.18` (as shipped in `vllm/vllm-openai:deepseekv41-flash-0909`)
- **Model**: `deepseek-ai/DeepSeek-V4.1-Flash` via vLLM (`0.1.dev20904+g179dd0fa9`)
- Related vLLM issue: https://github.com/vllm-project/vllm/issues/56702

### Problem

`sparse_mla_sm120_prefill.cu` instantiates DSV4 dual-cache prefill templates for
`extra_page_block_size` **64 or 2 only**. Both the full-tile fast path and the general
fallback end the same way:

```c
if (topk != 128) return false;
if (extra_page_block_size == 64)      DISPATCH_BY_NH_PBSX(64);
else if (extra_page_block_size == 2)  DISPATCH_BY_NH_PBSX(2);
return false;
```

so anything else surfaces as:

```
Unsupported sparse-MLA prefill configuration: model=DSV4 num_heads=16 topk=128
  page_block_size=64 topk_extra=512 extra_page_block_size=32
```

### Why 32 is reachable in practice

For DeepSeek-V4.1-Flash, `extra_page_block_size == cache_block_size // compress_ratio`, and
that checkpoint **mixes ratio-1 and ratio-2 layers** (`compress_ratios` in its `config.json`).
The two available block sizes are each blocked by a different constraint:

| cache block | ratio-1 layers | ratio-2 layers | outcome |
|---|---|---|---|
| **64** | indexer `num_states = 64` ✅ | `extra_page_block_size = 32` | **no compiled template** |
| **128** | indexer `num_states = 128` → DeepGEMM `block_kv ∈ {32,64}` assert ❌ | `extra_page_block_size = 64` ✅ | rejected earlier |

On sm_120 the decode path additionally pins the SWA page size to 64
(`_DECODE_DSV4_PAGE_BLOCK_SIZE = 64` in `flashinfer/mla/_sparse_mla_sm120.py`), so 64 is the
only viable cache block size — which forces `extra_page_block_size = 32` on every ratio-2
layer. There is currently no configuration of this model that sm_120 can serve.

[...]

Issue #5142, opened 2026-09-11
[Bug] SM120 sparse-MLA cpb calibration fails when it runs during engine startup (DGX Spark, TP=2) — plus a dead OOM fallback in `_allocate_kv_pool`


de1tydev:
## Summary

On 2× DGX Spark (GB10 / SM121, TP=2), `sparse_mla_sm120` cpb calibration fails on the **head rank** during vLLM engine startup and falls back to the C++ heuristic, losing #4802's calibrated dispatch. The same calibration **succeeds reliably when run on an idle GPU**. The failure is a measurement-contamination artifact of running calibration concurrently with engine startup, not a modelling bug.

```
WARNING:flashinfer.mla._sparse_mla_sm120_plan:SM120 sparse-MLA dsv4 cpb calibration failed
(implausible cpb calibration constants for dsv4: inv_bw=4.999999999999996e-13,
inv_rsm=1.9613565783730287e-10, c0=6.546037797171649e-06 (relative rms residual 0.374));
falling back to the C++ heuristic for this process.
```

Threshold is `rel_rms > 0.25` in `_sparse_mla_sm120_cpb.py::calibrate`.

## Evidence that it is timing contamination

1. **`c0` is ~10% higher when measured during startup.** Startup fit: `c0=6.546e-06`. Four clean runs on the same machine with the GPU idle: `c0 = 5.958 / 5.972 / 5.977 / 5.987e-06`.
2. **Head fails, worker succeeds.** In the same TP=2 boot, rank 1 produced a valid cached calibration while rank 0 failed. The head rank concurrently runs the API server and vLLM memory profiling.
3. **Idle-GPU calibration succeeds 4/4** on the identical machine, kernel build and device.

## Ruled out: the `is_integrated` L2 halving

We first suspected the recent change that halves `l2_cache_bytes` on integrated-memory devices, since that constant is the regime switch inside `predict_time_s`
(`min(g, c.sm_count) * cpb * c.bytes_per_chunk > c.l2_cache_bytes`).

Single-variable test — same process, same compiled kernel module, same GPU, only `props.is_integrated` swapped via a proxy object, 2 reps each, GPU idle:

| arm | `l2_cache_bytes` | result |
|---|---|---|
| `is_integrated=1` (halved) | 12,582,912 | **SUCCESS 2/2** — `inv_rsm` 1.97e-10 / 1.92e-10, `c0` 5.973e-06 / 5.972e-06 |
[...]

Issue #5135, opened 2026-09-11
[SM120] Sparse-MLA DSV4 prefill kernel missing macro instantiation for num_heads=16, topk=512


karn-shivam:
#### Summary
When serving `DeepSeek-V4-Flash-Vision-Exp` using **vLLM** on 4x NVIDIA RTX 6000 PRO Blackwell Server Edition (SM120) with Tensor Parallelism set to `TP=4` (which evaluates to `num_heads=16` per worker), the server crashes during startup within the `flashinfer_autotune` dummy warmup pass.

#### Error
```text
Unsupported sparse-MLA prefill configuration: model=DSV4 num_heads=16 topk=512 page_block_size=64 topk_extra=512 extra_page_block_size=64', please check the stack trace above for the root cause
```

#### Stack Trace & Log
```text

(Worker_TP0_EP0 pid=3998) INFO 09-11 09:13:44 [kernel_warmup.py:361] Running FlashInfer autotune with 16384 tokens and token buckets (1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 768, 1024, 1280, 1536, 1792, 2048, 2560, 3072, 3584, 4096, 8192, 16384).
	2026-09-11 14:43:44.983	
(Worker_TP0_EP0 pid=3998) 2026-09-11 09:13:44,983 - INFO - autotuner.py:995 - flashinfer.jit: [Autotuner]: Autotuning process ends
	2026-09-11 14:43:44.988	
(Worker_TP0_EP0 pid=3998) ERROR 09-11 09:13:44 [multiproc_executor.py:1055] WorkerProc hit an exception.
	2026-09-11 14:43:44.988	
(Worker_TP0_EP0 pid=3998) ERROR 09-11 09:13:44 [multiproc_executor.py:1055] Traceback (most recent call last):
	2026-09-11 14:43:44.988	
(Worker_TP0_EP0 pid=3998) ERROR 09-11 09:13:44 [multiproc_executor.py:1055]   File "/usr/local/lib/python3.12/dist-packages/vllm/v1/executor/multiproc_executor.py", line 1047, in _execute_worker_rpc
	2026-09-11 14:43:44.988	
(Worker_TP0_EP0 pid=3998) ERROR 09-11 09:13:44 [multiproc_executor.py:1055]     output = func(*args, **kwargs)
	2026-09-11 14:43:44.988	
[...]

PR #5048, opened 2026-09-09
fix(mla): prevent intermittent hang in SM120 sparse-MLA swapAB prefill on DGX Spark

bkryu:
## 📌 Description

Fixes the intermittent `test_sparse_mla_sm120` hang on Spark CI (#5001).

The swapAB prefill kernel (#4802) runs 4 IO warps, but only `BI` IO threads gather (`BI=64` for DSv3.2, 32 for DOTS3_SWA). IO threads with `io_tid >= BI` do no work and no barrier counts them, yet they still spin on `mbarrier_wait_parity` each tile. Since the pipeline's phases advance without them, a spectator warp starved past a full phase window near kernel drain waits on a parity that never completes again and spins  forever — the CTA never retires. A cuda-gdb autopsy of a live hang confirms it: the whole grid drained except one spectator IO warp spinning at the wait.

Matches the symptoms: intermittent (~2% per launch of the worst shape), worst at `num_heads=128` (max occupancy pressure), and absent from the mg impl (every thread participates in its barriers).

**Fix:** retire spectator IO threads before the pipeline loop (`if (io_tid >= BI) return;`) — they have no work after the block-wide sync.

## 🔍 Related Issues

#5001

## 🚀 Pull Request Checklist

Thank you for contributing to FlashInfer! Before we review your pull request, please make sure the following items are complete.

### ✅ Pre-commit Checks

- [x] I have installed `pre-commit` by running `pip install pre-commit` (or used your preferred method).
- [x] I have installed the hooks with `pre-commit install`.
- [x] I have run the hooks manually with `pre-commit run --all-files` and fixed any reported issues.

> If you are unsure about how to set up `pre-commit`, see [the pre-commit documentation](https://pre-commit.com/).

## 🧪 Tests

- [ ] Tests have been added or updated as needed.
- [ ] All tests are passing (`unittest`, etc.).

## 🔬 Experimental Track

- [ ] This PR is **experimental**: it adds or changes code under `flashinfer/experimental/` and/or an `@flashinfer_experimental_api`. Tracking issue: #
  - [ ] The tracking issue names an owner, the reason for the experimental path, and a graduation plan with a target release.
  - [ ] Core changes are limited to a thin entry point (signature, shared validation, feature-gate check, backend selection, handoff).
  - [ ] Tests live in `tests/experimental/` and were validated on the intended hardware; a runnable example is included.
[...]

PR #5197, opened 2026-09-13
refactor(sparse-mla): unify SM120 execution, calibration and DSv4.1 support

lucifer1004:
### TLDR

Reorganize SM120 sparse MLA into a package with clear ownership and add **DSv4.1 dual-cache attention** (FP8/BF16 compute, FP4/FP8 extra cache), **runtime page/head/topk dispatch** for the six ordinary families, and **per-configuration L2-cold calibration** with unified prepared execution. Fixes empty-effective-KV semantics (#4484), startup calibration recovery (#5142), and the finite-instantiation dispatch gaps reported in #4541/#5135/#5192. 947 tests pass on SM120 (CUDA 12.9 and 13.2) and SM121 (GB10); kernel performance is within 1% of the pre-refactor baseline on a 95-point matrix, with DOTS3 SG 6.9–11.9% faster.

### Features

- **DSv4.1 dual cache**: primary FP8 cache plus optional FP4/FP8 extra cache, independent topk/page sizes, FP8 or BF16 compute. New public APIs `dsv41_fp4_quantize_pack_sparse_mla_cache` / `dsv41_fp4_quantize_append_sparse_mla_cache` with docs, trace templates, and generated examples. The DSv4 compatibility layout accepts page 32, covering vLLM's page-32 DSv4.1 allocation.
- **Runtime dispatch**: all six ordinary families (`dsv4`, `dsv3_2`, `glm_nsa`, `glm53_nope`, `dots3_swa`, `dsv4_1`) accept **positive integer runtime page sizes, including non-powers of two**, in decode and their supported prefill routes, with independently sized main/extra pages for dual caches. Existing constraints remain: inline prefill requires gap-free pages (row padding is supported); footer caches require aligned page strides. **DSv4 NVFP4 is not broadened** (main page 64, extra pages 2/64, existing finite head/topk instantiations).
- **Calibration**: every ordinary family gets per-configuration, per-device token-bucket profiles alongside NVFP4 in the shared schema-v3 store. The cold protocol measures **100 samples and their mean** with a **4x L2** eviction before each timed call (eviction outside the interval); a protocol identifier discards stale measurements. Off-grid token counts are refined outside the store lock and merged atomically; `force` replaces complete profiles. Profiles apply to canonical packed layouts only. `calibrate_sparse_mla_sm120` gains keyword-only precision, page-size, extra-cache, topk-length, and sink options; defaults preserve single-cache behavior.
- **Prepared execution**: downstream entry points share cached plans, modules, and workspaces. Eager calls execute from growable per-wrapper scratch arenas (sweeping tokens 1→64 grew device memory by ~103 MiB before, ~33 MiB now ≈ the largest signature); a warmed signature's first capture pins dedicated buffers before recording, so captured graphs survive arena growth. Caller-supplied and functional-path scratch are unchanged; eager calls sharing wrapper-owned scratch must not race across streams (documented).

### Bug fixes

- **#4484 empty effective KV**: rows with zero length or only `-1` padding indices now return **O = 0 and LSE = -inf** (base 2) from every SM120 sparse store path; a finite attention sink still contributes (`LSE = sink·log2(e)`, O = 0). Producers previously scaled the `-1e30` sentinel so all-padding rows got spurious unit probability; masked probabilities are now zeroed by validity, internal partials keep the finite sentinel, and only the final store exports `-inf`. Covered for zero-length and all-negative rows, single/dual caches with either side empty, finite and extreme sinks, real CUDA `merge_state` neutrality, and graph replay with in-place metadata updates.
- **#5142 startup calibration**: `torch.AcceleratorError` OOM is recognized, restoring pool-size fallback (initializer references are released before retrying); an implausible analytical fit is remeasured once without relaxing the residual threshold; failed fits do not roll back successful profiles. The recommended workflow — idle pre-calibration in a separate process, then cache-only warmup via `skip_ops` (which also suppresses off-grid refinement) — is documented on the API. Concurrent load can still contaminate measurements; SM121 TP2 startup has not been revalidated.
- **Finite-instantiation dispatch gaps** (#4541 non-instantiated decode topk falling through to prefill, #5135 H16/K512 dual prefill, #5192 extra page 32): removed by the runtime dispatch above. These shapes are reachable in the resolver and generic runtime-page tests; their exact serving configurations have not been rerun end to end.
- **#5001 hang guard**: swapAB spectator IO warps retire after CTA sync and register release, as upstream #5048 established — they gather nothing and no barrier counts them, so waiting on ring phases could spin forever at kernel drain. Verified with 24,000 swapAB executions/replays on both SM120 and SM121.
- **CUDA 12.9 mixed-cache BF16 conversion**: uses the `mov.b16` low-byte extraction and `h` constraint matching CUDA's FP4 header; the five original CI failures were reproduced first. The remaining `mov.b32` form in `e2m1x2_code_to_float2` is safe because its `uint8_t` input guarantees zero upper bytes (documented in code); checked on 13.2 with a maximum 0.68% slowdown at three points.
- **Review hardening**: capability probing is JIT-free again (static coarse pre-filter; C++ dispatch stays authoritative); the custom op's new `extra_fp4` parameter has a default; functional entries accept padded-capacity LSE buffers like the wrapper; the calibration store self-heals a corrupt same-schema cache during locked merge and keeps failure marks process-lifetime; ordinary lazy calibration has the NVFP4-style re-entry guard and failed refinements are suppressed in-process; NVFP4 pool floors account for `2x L2` (large-L2 parts) and its profiles are gated to canonical packed layouts; NVFP4 merge2 handles partial head blocks; MG dual prefill no longer reads extra indices past the row end (compute-sanitizer: 133 invalid reads before, 0 after); dead exports/fields removed; standalone bindings range-check narrowing overrides and require `num_splits` to cover the resolved chunking.

### Organization and compatibility

- `flashinfer/mla/_sparse_mla_sm120.py` becomes a same-named package with explicit root re-exports and separate API, execution, prepared-plan, policy, calibration, and NVFP4 owners. Root entry points used by vLLM, SGLang, TRT-LLM, and b12x remain available, including dispatch iteration for capability probes. Old auxiliary module paths, root monkeypatch reassignment, and pickles referencing them are not compatibility guarantees.
- C++ bindings/dispatch are split by format and headers by responsibility under one `sparse_mla_sm120` JIT/AOT module, including NVFP4 exports and the test-only MMA layout probe; the removed `gen_sparse_mla_nvfp4_sm120_module` / `_tile_module` names remain as compatibility aliases.
- Kernel work: SG pointer arrays replaced by base-plus-offset addressing (256 B less local storage); DOTS3 keeps its four page-64 SG performance specializations; DSv4 tile addressing simplified; H32 FP8 single-cache MG uses a 240/24 register split.

## 🔍 Related Issues

Follow-up to #4802. Defines the empty-effective-KV contract requested by #4484 for the SM120 sparse MLA paths (other backends are out of scope), addresses the calibration recovery and OOM fallback from #5142, and removes the dispatch gaps reported in #4541, #5135, and #5192.

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

`prek` checks passed for changed files. Hook installation and an all-files run were not performed.

## 🧪 Tests

- **CUDA 13.2.86, RTX PRO 6000 Blackwell (SM120): 949 passed, no skips** on the final tree across the seven files below plus `test_sparse_mla_sm120_binding_guards.py`. CUDA 12.9.86 ran 882 passed on the pre-#4484 tree; the conversion fix itself is re-covered by the current suite.
- **GB10 (SM121), torch 2.13 / nvcc 13.3: 923 passed, 0 skipped** on the #4484 tree (same scope minus binding guards); later commits touch Python policy/docs and two comment-level C++ cleanups only. A separate 24,000-replay swapAB soak passed on both SM120 and SM121.
- **#5142 targeted CPU/mock regression: 14 passed, no skips** (AcceleratorError OOM fallback, initializer release before retry, bounded fit remeasurement, non-OOM propagation, cached-profile startup behavior).
- **Seven-file CPU-only run: 112 passed / 770 skipped.** Earlier architecture-skip checks with no GPU and simulated major architectures 9/10 each yielded 99 passed / 116 skipped on the four host-policy files. These are not GPU execution results for SM90/SM100.
- Real `prek` checks passed for changed files. Neither a full-repository test run nor hook installation/all-files hook execution is claimed.

From the repository root in an SM120 CUDA development environment, run the same scope:

```bash
python -m pytest -q \
  tests/attention/test_sparse_mla_sm120.py \
  tests/attention/test_sparse_mla_sm120_policy.py \
  tests/attention/test_sparse_mla_sm120_calibration.py \
  tests/attention/test_sparse_mla_sm120_execution.py \
  tests/attention/test_sparse_mla_sm120_precision.py \
  tests/attention/test_sparse_mla_sm120_cache_contracts.py \
  tests/attention/test_sparse_mla_sm120_dsv4_nvfp4.py \
  tests/attention/test_sparse_mla_sm120_binding_guards.py
```

## 🔬 Experimental Track

```experimental-tests
#
#
```

## Reviewer Notes

Performance evidence on **RTX PRO 6000 Blackwell (SM120)**:

- The SG base-plus-offset rewrite reduced measured DOTS3 kernel latency by **6.9–11.9%**.
- A **95-point kernel regression matrix**, compared with the SG-pointer-optimized version before the subsequent SSOT cleanup, has **no confirmed stable regression above 1%** after follow-up measurements. This is not a claim that every raw sample is within 1% or a comparison against upstream main.
- Across a **196-point profile-transfer regret matrix**, refinement yields **0.1013% mean regret, 3.825% maximum, and 7/196 points above 1%**, versus a pre-refinement maximum of **36.324%**. Calibration rotates index sets; the reference candidate sweep uses fixed indices. These numbers measure the selected candidate against the best measured legal candidate on that reference workload, not before/after kernel speedup.
- The CUDA 12.9 BF16 conversion fix was also checked on CUDA 13.2 at three performance points: maximum measured slowdown **0.68%**.

Review focus: runtime-page/stride eligibility versus canonical profile eligibility, per-configuration cache invalidation, and the unchanged finite NVFP4 boundary.

## Summary by CodeRabbit

* **New Features**
  * Added DeepSeek-V4.1 sparse MLA support with FP8 and mixed FP8/FP4 cache formats.
  * Added APIs for quantizing, packing, and appending DSV4.1 FP4 KV-cache data.
  * Added FP8, BF16, and NVFP4 execution options for supported SM120 hardware.
  * Improved CUDA Graph compatibility, workspace handling, validation, and execution-plan reuse.

* **Bug Fixes**
  * Strengthened cache layout, alignment, stride, capacity, masking, and optional-input validation.
  * Improved calibration and execution planning for decode, prefill, and dual-cache workloads.


aleozlx:
## PR Review Screening

**CI verdict:** ✅ auto-run ok
**Review category:** `live` (rules fired: C1.2 net-new public interface; C3.2 durable change reorganizing shared abstractions and dispatch conventions)
**Blocking checks:** S2, S3
**Release blocker:** no
**Early stop:** no

<sub>The diff exceeds the API cap (109 files, 17,933+/10,919−), so the evidence below is derived from a local diff against merge-base `93e9eef0`.</sub>

### Security
| Q | Answer | Evidence |
|---|--------|----------|
| S2 template overwritten | ❗ yes | body uses a custom Summary / Motivation / Changes / Testing format; none of the five required headers present, zero checklist items |
| S3 template obligations | ❗ partial | unmet formally, largely met in substance — no checklist exists at all, but the body documents test results (823 passed, 8 skipped on SM120), A/B measurements and design rationale |

### Packaging
| Q | Answer | Evidence |
|---|--------|----------|
| C1.1 dependency bump | no | no supply- or CI-bucket paths in the diff |
| C1.2 public API changes | yes — **net-new interface** + extension | net-new, both added to `flashinfer/mla/__init__.py` lazy exports: `dsv41_fp4_quantize_pack_sparse_mla_cache(latent_kv: torch.Tensor, *, kv_layout: str = "HND") -> torch.Tensor` and `dsv41_fp4_quantize_append_sparse_mla_cache(latent_kv: torch.Tensor, slot_mapping: torch.Tensor, cache: torch.Tensor) -> None`. Extension: `calibrate_sparse_mla_sm120` gains 8 keyword-only parameters, defaults preserving current behaviour. The `_`-prefixed module renames are private and change no surface |
| C1.3 AOT/trace registration | ❗ gap | AOT half is clean — the NVFP4 route folds into `gen_sparse_mla_sm120_module()` and `aot.py` registers the single module. Trace half is not: both new public `@flashinfer_api` helpers are tensor-in/tensor-out and carry no `trace=`, and no `tests/trace/fi_trace_out/` JSONs are regenerated |

### Presentation
| Q | Answer | Evidence |
|---|--------|----------|
| C2.1 perf claim backed | yes | RTX PRO 6000 Blackwell (SM120) named; page-64 A/B against the pre-dispatch tree at worst ratio 1.002 across the production shape grid, page 32 within 1.06x on MG prefill, and the rejected runtime-page variant quantified (+30% generic division, +4–7% pow2 rewrite); four `benchmarks/bench_sparse_mla_nvfp4_*.py` scripts ship with the change |

### Implementation
| Q | Answer | Evidence |
|---|--------|----------|
| C3.2 durable areas touched | yes — durable | `include/flashinfer/attention/sparse_mla_sm120/` reorganized into `arch/ compute/ kernels/ execution/ pipeline/ model/`; `csrc/sparse_mla_sm120/` split into per-format dispatch/binding units; JIT module generation consolidated; calibration store schema bumped to v3 |
| C3.3 tests match change | yes | 10 test files, +6,212/−2,603, restructured by responsibility (policy / calibration / execution / precision / cache contracts); SM120 is a lane in the fleet, so this suite does execute in CI |
| C3.4 release-blocker candidate | no | no hang/crash/corruption or prior-release regression claimed; see the first maintainer note for a compatibility item that is adjacent but does not meet the bar |

### Experimental track
| Q | Answer | Evidence |
|---|--------|----------|

**Notes for the maintainer**

- **The page-32 work is downstream unblocking, not refactoring.** vLLM main allocates 32-token pages for DeepSeek V4.1 sparse attention, and the fixed-page-64 DSv4 compat cache rejected that allocation at warmup. A `refactor(...)` title and a 109-file diff hide a change that some consumers are currently blocked on — worth surfacing in the release notes, and worth asking whether it should land ahead of the reorganization rather than inside it.
- **The trace gap is wider than this PR.** `sparse_mla_sm120_paged_trace` is defined in `flashinfer/trace/templates/attention.py:2971` and wired to nothing — that is already true at the merge base, so the family has no trace coverage today and the two new helpers extend the gap rather than open it. Fixing it properly is a family-level task; deciding that here is reasonable, deciding it silently is not.
- `gen_sparse_mla_nvfp4_sm120_module` and `gen_sparse_mla_nvfp4_sm120_tile_module` remain defined in `flashinfer/jit/mla.py` with no callers anywhere after the AOT consolidation.

include/flashinfer/attention/sparse_mla_sm120/kernels/fp8_prefill/prefill_swapab.cuh
```diff
@@ -158,10 +157,10 @@ __global__ void __launch_bounds__(BLOCK_THREADS, 1)
     for (int ti = 0; ti < actual_ni; ti++) {
       const int buf = ti & 1;
       const int next = ld_idx(ti + 2);
-      mbarrier_wait_parity(sm.mbar_wr + buf, wr_phase);
+      Ring::Free::wait(sm.mbar_wr + buf, wr_phase);
```
bkryu, line 162: The rewrite deleted the #5075-era starvation-hazard documentation along with the `if (io_tid >= BI) return;` retirement guard  ("Threads with io_tid >= BI gather nothing and no barrier counts them… one starved past a full phase window near kernel drain  would spin on a parity that never completes again"). On inspection the new structure is safe — `ld_idx` returns -1 for those lanes,  the gather helpers early-return, and surplus threads now execute `Ring::Free::wait` every tile so they stay paced with the phase  window — i.e., this is an alternative fix, not a regression. But the hazard rationale vanished with the old fix, and this exact  spin-on-dead-parity failure mode was a shipped GB10 hang once (#5001/#5048). Please restore the comment in its new form: state  that surplus lanes are deliberately kept in the wait loop so they can never fall a full phase window behind.

flashinfer/mla/_sparse_mla_sm120/_api.py
```diff
@@ -1613,6 +1324,119 @@ def sparse_mla_sm120_decode_dsv4(
...
+        model_type,
         cpb_override,
+        extra_fp4,
     )
     return output
+
+
+@supported_compute_capability([120, 121])
+@flashinfer_api
+def dsv41_fp4_quantize_pack_sparse_mla_cache(
```
bkryu, line 1334: These two new public APIs are exported, @flashinfer_api-decorated, capability-gated, and AOT-covered  but don't have documentation in `docs/api/attention.rst`. Can we try adding them?

flashinfer/mla/_sparse_mla_sm120/_calibration.py
```diff
@@ -1224,26 +1498,434 @@ def save_cpb_override(
...
+        "pool_bytes": [
+            ctx.main.numel(),
+            0 if ctx.extra is None else ctx.extra.numel(),
+        ],
+        "buckets": buckets,
+    }
+
+
+@_target_calibration
+def refine_dsv41(
```
bkryu, line 1896: Both refine paths run their full GPU timing sweep while holding the module-global _store_lock — unlike _calibrate_dsv41, which deliberately measures outside the lock and merges under it. Every other thread's store read blocks for the sweep duration. Hoist the measurement out with an absent-recheck-before-merge, matching the calibrate path's pattern.

csrc/sparse_mla_sm120/dsv4_nvfp4_resolve.cu
```diff
@@ -0,0 +1,169 @@
...
+  TVM_FFI_ICHECK(
+      extra_topk >= 0 &&
+      (extra_topk == 0 ||
+       (visit_extra_page(extra_page_size, [](auto) { return true; }).supported &&
+        extra_page_stride_bytes >= size_t(extra_page_size) * Dsv4Nvfp4Layout::BYTES_PER_TOKEN &&
+        extra_page_stride_bytes % 16 == 0)))
+      << "unsupported NVFP4 extra page layout";
+  TVM_FFI_ICHECK_GT(sm_count, 0) << "SM count must be positive";
+  TVM_FFI_ICHECK(!prefill || !stage1_only) << "stage1_only is decode-only";
+  const int chunks = (topk + DECODE_CAND_WINDOW - 1) / DECODE_CAND_WINDOW +
```
bkryu, line 35: The unified resolver guards `topk <= INT_MAX - 64 - extra_topk` before this class of arithmetic; this NVFP4 resolver lacks the  equivalent guard, so chunks can overflow int for adversarial `topk/extra_topk`. Practically unreachable, but add the same ICHECK for parity — the two resolvers should reject the same garbage.

bkryu (commented):
Thanks @lucifer1004, gave a first pass.

I'm noticing that two GLM tests are dropped: `test_glm53_compact_rows_match_padded_rows` and `test_glm53_canonical_payload_is_scoped_to_model`. Maybe we should add them back?

flashinfer/mla/_sparse_mla_sm120_dsv4_nvfp4_policy.py
```diff
@@ -296,17 +291,60 @@ def plan_nvfp4_sparse_mla_sm120(
...
+
+    # Each bucket stores its own selection.  The shared selector consumes the
+    # decision directly, so NVFP4 does not inherit FP8's monotonic crossover
+    # assumption; a missing profile keeps the decode-first fallback.
+    bucket = (
+        None
+        if profile is None or num_tokens > _DECODE_MAX_TOKENS
+        else _cpb._profile_bucket(profile, num_tokens)
+    )
+    t_bucket = _token_bucket(num_tokens) if num_tokens <= _DECODE_MAX_TOKENS else -1
```
saltyminty: `bucket` resolves to the exact bucket returned by `_profile_bucket`, while the cache entry uses `t_bucket` which uses the standard bucket sizes of `(1, 4, 8, 16, 24, 32, 48, 64)`.  Is there a reason for this discrepancy, and wouldn't this break caching (multiple bucket sizes map to the same cache key)

flashinfer/mla/_sparse_mla_sm120_calibration.py
```diff
@@ -965,61 +1108,178 @@ def _parse_payload_devices(devices: dict) -> tuple[dict, dict, dict]:
...
+                if not k.startswith(prefixes)
+            }
+    dev.update(constants)
+    for section, entries in (
+        (_DECODE_MAX_TOKENS_KEY, crossover),
+        (_CPB_OVERRIDES_KEY, overrides),
+    ):
+        dev.setdefault(section, {}).update(entries)
+    if profiles:
+        dev.setdefault("profiles", {}).update(profiles)
```
saltyminty: This issue was flagged by my agent and I don't have a super great understanding of this so take with a grain of salt:
This update step updates the outer dictionary: for an existing profile key, it replaces the whole profile, including its nested bucket dictionary. It does not merge the individual bucket entries. This can cause a lost-update race where two processes refining the same calibration profile can overwrite each other’s results, even though writes use a file lock.

flashinfer/mla/_sparse_mla_sm120/_api.py
```diff
@@ -26,47 +26,27 @@
 # OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-"""Internal Sparse-MLA paged attention implementation for SM120.
```
saltyminty, line 29: As a general statement and not specific to this PR:
Given that we have a multitude of `_sparse_mla_sm120*` files in the `mla` folder, would it be better organizational structure to have an explicit `_sparse_mla` or `_sparse_mla_sm120` folder (matching the `_batch_mla` folder?

flashinfer/mla/_core.py
```diff
@@ -625,63 +481,28 @@ def _trtllm_batch_decode_sparse_mla_sm120(
...
         extra_indices=extra_segment.indices if extra_segment is not None else None,
-        extra_topk_length=extra_segment.lengths if extra_segment is not None else None,
-        out_lse=out_lse_arg,
-        mid_out=mid_out,
-        mid_lse=mid_lse,
-        return_lse=return_lse,
+        extra_lengths=extra_segment.lengths if extra_segment is not None else None,
+        lse=out_lse_arg,
+        kv_scale_format=kv_scale_format,
+        is_dsv4_nvfp4=kv_cache_format == "nvfp4",
```
saltyminty: NIT, could we do `is_dsv4_nvfp4=(kv_cache_format == "nvfp4"),` for better readability?

flashinfer/mla/_sparse_mla_sm120/_api.py
lucifer1004: I agree that there should be a separate folder, but need to do a bit investigation since many downstream packages (like vLLM, SGLang) currently depend on these private namespaces.

flashinfer/mla/_sparse_mla_sm120_dsv4_nvfp4_policy.py
lucifer1004: Good catch. Fixed in c86f61fc: the variant memo now keys on the exact token count, matching refined selections. Added a regression covering both lookup orders.

flashinfer/mla/_sparse_mla_sm120_calibration.py
lucifer1004: Confirmed and fixed in c86f61fc. Refinement now merges only new buckets under the file lock, including failed-write overlay replay. Full recalibration still replaces the complete profile.

flashinfer/mla/_sparse_mla_sm120/_calibration.py
lucifer1004: Fixed both paths in c86f61fc: GPU allocation and measurement run outside the store lock, followed by a short recheck and incremental publication.

csrc/sparse_mla_sm120/dsv4_nvfp4_resolve.cu
lucifer1004: Fixed in c86f61fc. The descriptor path already bounded extra_topk, but the legacy tensor binding bypassed that check and performed split arithmetic earlier. Both paths now enforce the shared bound before arithmetic; tensor widths are checked before narrowing.

include/flashinfer/attention/sparse_mla_sm120/kernels/fp8_prefill/prefill_swapab.cuh
lucifer1004: Added a short comment explaining why surplus IO lanes retain the per-tile wait in c86f61fc. No kernel behavior change.

flashinfer/mla/_sparse_mla_sm120/_api.py
lucifer1004: Added both APIs to docs/api/attention.rst in c86f61fc.

flashinfer/mla/_core.py
lucifer1004: Done in c86f61fc.

lucifer1004:
@bkryu Thanks for flagging the dropped GLM coverage. In c86f61fc, I strengthened the existing compact-row test with poisoned padding, final-slot access, variable lengths, and changed-input graph replay rather than restoring a duplicate test. The layout rejection checks are also restored, accounting for DSv4.1 legitimately sharing the 528-byte row size.

Validation: 836 tests passed across the seven SM120 test files; 101 execution/NVFP4 tests passed again after the final bound-constant cleanup. Pre-commit passed. The directory reorganization remains deferred pending downstream import checks.

flashinfer/mla/_sparse_mla_sm120/_api.py
lucifer1004: Implemented in d9aed261 after checking current vLLM, SGLang, TensorRT-LLM, and b12x callers. The implementation now lives in `_sparse_mla_sm120/`; its package root retains the functions, runner, loader, and capability members those callers import. Also restored dispatch iteration required by SGLang. All 838 SM120 tests and pre-commit passed; full downstream services were not launched.

lucifer1004:
Integrated the validated spike through 379e2e58, retaining the DOTS3 page-64 specializations and cold-L2 calibration.

The CI failures had separate causes:
- CPU policy tests accidentally invoked SM120-only compiled metadata. They now isolate that dependency; real compiled tests have scoped prerequisites, without blanket exception-to-skip handling.
- CUDA 12.9 miscompiled the FP4 low-byte extraction used by BF16 attention. Reproduced all five numerical failures with an isolated NVIDIA 12.9 toolchain and fixed the expression using the mov.b16 / halfword constraint pattern from its CUDA FP4 header. No tolerance changes or version skips.
- The allocation assertion now detects allocation events instead of comparing process-wide live bytes, which can decrease when unrelated objects are released.

The complete seven-file suite passes with both CUDA 12.9.86 and 13.2.86: 870 tests each. GPU-hidden tests with no target, SM90 and SM100 targets each give 99 passed / 116 appropriately skipped. Pre-commit passed. Python 3.12 / Torch cu130 was used locally, rather than reproducing every CI environment detail.

flashinfer/mla/_sparse_mla_sm120/_api.py
saltyminty: Can we also add an explicit trace template to both new APIs? Otherwise PR looks good to me

saltyminty:
Left one comment about trace templates: https://github.com/flashinfer-ai/flashinfer/pull/5197#discussion_r4018783538

Will rerun CI in the meantime.

flashinfer/mla/_sparse_mla_sm120/_api.py
lucifer1004: Added explicit trace templates for both APIs in 67304b67, including supported layouts and in-place append semantics, with generated examples and trace tests. Verified autodump on SM120 and CPU tracing with JIT disabled.

qsang-nv:
Two findings from the SM120 execution and FP4 conversion paths.

### 1. [P1] Internally-owned scratch is retained per prepared signature, including on the eager path

`wrapper_run` keys the prepared-call cache on `tensor_signature` for all twelve tensors, so any shape or stride change produces a new key. For each new key, `prepare(owned=True)` allocates fresh `mid`/`mlse`/`lse` whenever the corresponding caller-owned buffers (`mid_out`, `mid_lse`, `out_lse`) are absent, and the `PreparedCall` holding them goes into `wrapper._prepared_calls`. That dict has no `clear()`, `pop()`, `del` or size bound anywhere in `_prepared.py`, and the `current.workspace == workspace` reuse branch only applies to a re-`prepare()` of the *same* key. `test_sparse_mla_sm120.py` explicitly codifies this: `assert grown.mid is not small.mid`, `assert len(runner._prepared_calls) == 3`.

Retention is deliberate for capture — `_api.py:831` says "live graph buffers are retained across shape changes" — but `wrapper_run` applies it unconditionally, so peak memory goes from `O(max observed workspace)`, which `main` provided via a single per-dimension-grown `self._mid_out`, to `O(sum of all observed signatures)`.

For DSv4 decode with `H=128`, `topk=1024` (`chunk_capacity=16`), `d_v=512`, `partial_bytes` is 2 MiB per token for `mid` alone and `_DECODE_MAX_TOKENS` is 64: `main` stays at 128 MiB for any number of shapes, while this PR reaches 590 MiB for a capture set like `[1, 2, 4, 8, 16, 24, 32, 40, 48, 56, 64]` and 4.1 GiB if every `T` in `1..64` is seen. A dual cache with `extra_topk=1024` doubles `chunk_capacity` and therefore the dominant `mid`/`mlse` storage. This is reachable through legitimate dynamic-batch use with no CUDA graph involved.

I am raising this as blocking. The documented graph-capture retention is not what I am contesting; the unbounded accumulation for signatures only ever executed eagerly is, and it needs to be resolved before merge. Is that accumulation an intentional trade-off, or a side effect of the eager path sharing one mechanism with capture?

### 2. Non-blocking: the scope of the ptxas 12.9 workaround is not recorded

`379e2e58` rewrites the low-byte extraction in `decode_e2m1_or_e4m3_pair_bf16`:

```cpp
-    asm("{ .reg .b8 x; mov.b32 {x,_,_,_}, %1; cvt.rn.f16x2.e2m1x2 %0,x; }"
-        : "=r"(half_pair) : "r"(raw));
+    asm("{ .reg .b8 x, unused; mov.b16 {x,unused}, %1; cvt.rn.f16x2.e2m1x2 %0,x; }"
+        : "=r"(half_pair) : "h"(uint16_t(raw)));
```

`e2m1x2_code_to_float2` in `compute/nvfp4_quantization.cuh` still uses the `mov.b32 {b8, _, _, _}` form, and this PR's new `kernels/dsv41_fp8/convert.cuh:56` calls it on the `fp8_dsv41_fp4_ca` path. That site passed in the reported CUDA 12.9 full-suite run, and the two may genuinely differ: the fixed helper takes `uint32_t raw` and its caller in `dsv41_bf16/decode.cuh:91` passes `uint32_t(raw >> (j * 8))` from a `uint64_t` payload, so bytes above byte 0 hold live data for `j < 3`, whereas `e2m1x2_code_to_float2` takes `uint8_t codes` and both call sites truncate explicitly (`& 0xFF`, `static_cast<uint8_t>`).

If that is the trigger, the CUDA-12.9-specific reason the remaining form is safe is not documented, even though the zero-extension invariant is enforced today by the `uint8_t` interface and both call sites. Could you confirm whether live upper bytes are the condition you reproduced, so the boundary is on record?

lucifer1004:
@qsang-nv Both points addressed in 1f23a45c.

**1. Eager scratch retention** — side effect, not an intentional trade-off; fixed. Eager wrapper entries no longer retain per-signature owned mid/mlse: execution slices views from two shared, growable per-wrapper arenas at call time, so eager memory tracks the largest signature instead of the sum. Measured on DSv4 decode sweeping tokens 1→64: allocated growth went from ~103 MiB (accumulated) to ~33 MiB (arena capacity for the largest signature), with zero mid tensors retained by eager entries. Capture semantics are preserved by pinning: the first capture of a warmed signature allocates dedicated buffers before recording, and those stay live for the wrapper lifetime, so later arena growth cannot invalidate a captured graph. The existing capture-guard test now permits exactly that one pin allocation and still rejects any other capture-time allocation. Two new regressions cover eager-sweep memory growth and capture A → eager-grow → replay A exact equality. 924 passed on the seven SM120 files.

Two behavior notes, also added to the PR body and wrapper docs: the first capture of a shape may perform this pin allocation (not fully allocation-free), and eager calls sharing wrapper-owned scratch must not race on different streams; callers needing that can pass their own mid/mlse.

**2. ptxas 12.9 scope** — confirmed: live upper bytes are the trigger. The rewritten helper takes uint32_t and its caller passes raw >> (j*8) with live data above byte 0 for j < 3. The remaining mov.b32 {b8,...} site in e2m1x2_code_to_float2 takes uint8_t and both call sites truncate explicitly, so the upper bytes are zero and the miscompile condition does not apply. That invariant is now recorded in a comment at the helper.

saltyminty:
Tests largely failed to start, rerunning now

saltyminty (approved):

qsang-nv (approved):

saltyminty:
/bot run tests/attention/test_sparse_mla_sm120_calibration.py

PR #5198, opened 2026-09-14
fix(comm): match the library filename when scanning /proc/self/maps

JacobHelwig:
## 📌 Description

`find_loaded_library` accepts any mapping whose path merely **contains** the requested name, and
stops at the first hit. `/proc/self/maps` is walked in address order, so an unrelated library that happens to sort earlier
silently wins. A concrete case: [tilelang](https://github.com/tile-ai/tilelang) ships `libcudart_stub.so`, which
exports only a subset of the CUDA runtime. When it is mapped before the real `libcudart`,
`CudaRTLibrary()` binds the stub:

```python
```

Fix: Compare the mapping's own filename rather than a substring of the line. This accepts the forms the loader actually produces:

| | |
|---|---|
| `libcudart.so` | ✅ |
| `libcudart.so.12` | ✅ |
| `libcudart.so.11.0` | ✅ |
| `libcudart-d0da41ae.so.11.0` (torch's hash-suffixed copy) | ✅ |
| `libcudart_stub.so` | ❌ |
| `libcudart_mock.so` | ❌ |
| `libcudartfoo.so.1` | ❌ |

The hash-suffixed form is kept deliberately bc the function's own comment calls it out as a case it
must handle.

Behavior is otherwise unchanged: a genuine match returns the same path as before, and a process
with no CUDA runtime mapped still returns `None`. Returning `None` when only a stub is loaded is
the intended outcome — a missing runtime is a different, and much clearer, failure than binding
the wrong library.

## 🔍 Related Issues

https://github.com/tile-ai/tilelang/issues/2314

https://github.com/sgl-project/sglang/issues/30875 

## 🚀 Pull Request Checklist

### ✅ Pre-commit Checks

## 🧪 Tests

Error goes away after patching.

## 🔬 Experimental Track

NA

## Reviewer Notes

NA

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CUDA library detection to match complete, valid library filenames.
  * Prevents similarly named files, trailing suffixes, stubs, and directory names from being mistaken for the requested library.
  * Supports valid build-suffixed library names and versioned `.so` filenames.
  * Improves reliability when multiple library variants are present or when only an unsupported stub is available.


yyihuang (commented):
The reported TileLang case is fixed by this change. I checked the functions from d2f2e66 with mocked `/proc/self/maps` input: the seven filename examples in the description and six lookup scenarios passed. This was a CPU-only check of the extracted lookup functions, not a CUDA integration test.

Two follow-ups:

1. Could we add regression tests for `find_loaded_library` with mocked maps? In particular: the underscore stub before the real runtime, the reverse order, only the stub (expect `None`), a hash-suffixed runtime, and `libcudart` appearing only in a directory name. The existing IPC test does not construct these mappings, so it does not deterministically exercise the bug this PR fixes.

2. One additional boundary case alongside CodeRabbit's existing hyphen/hash-suffix finding: `rpartition(".so")[0]` also accepts `libcudart.something` and `libcudart.so.backup`. With either path before `/cuda/lib64/libcudart.so.12` in mocked maps, the lookup still returns the first path. When tightening the matcher, could we validate the whole filename, including the `.so` boundary and supported version suffix, and add negative cases for these names? These are synthetic examples of a remaining matching gap that also exists in the base revision, rather than a demonstrated new runtime regression.

yyihuang (approved):

JacobHelwig:
> The reported TileLang case is fixed by this change. I checked the functions from [d2f2e66](https://github.com/flashinfer-ai/flashinfer/commit/d2f2e664edb66b7dda57cf458788f80aedcf79d1) with mocked `/proc/self/maps` input: the seven filename examples in the description and six lookup scenarios passed. This was a CPU-only check of the extracted lookup functions, not a CUDA integration test.
> 
> Two follow-ups:
> 
> 1. Could we add regression tests for `find_loaded_library` with mocked maps? In particular: the underscore stub before the real runtime, the reverse order, only the stub (expect `None`), a hash-suffixed runtime, and `libcudart` appearing only in a directory name. The existing IPC test does not construct these mappings, so it does not deterministically exercise the bug this PR fixes.
> 2. One additional boundary case alongside CodeRabbit's existing hyphen/hash-suffix finding: `rpartition(".so")[0]` also accepts `libcudart.something` and `libcudart.so.backup`. With either path before `/cuda/lib64/libcudart.so.12` in mocked maps, the lookup still returns the first path. When tightening the matcher, could we validate the whole filename, including the `.so` boundary and supported version suffix, and add negative cases for these names? These are synthetic examples of a remaining matching gap that also exists in the base revision, rather than a demonstrated new runtime regression.

Thanks @yyihuang, here are the changes: https://github.com/flashinfer-ai/flashinfer/compare/d2f2e664edb66b7dda57cf458788f80aedcf79d1...JacobHelwig:flashinfer:jhelwig/fix-libcudart-lookup-matches-stub?expand=1

yyihuang (commented):
Re-reviewed 819d774. Both follow-ups from my previous review are addressed: the matcher now rejects the `.something` / `.so.backup` cases, and the new tests cover the requested maps-order, stub-only, hashed-runtime, and directory-name scenarios.

I ran the unmodified new test file against the two production lookup functions extracted from this commit: **19 passed**. This bypassed package imports and repository conftest on macOS; it does not validate the normal package import path or CUDA integration.

The remaining hyphen-suffix gap is already tracked in [the existing review thread](https://github.com/flashinfer-ai/flashinfer/pull/5198#discussion_r4002585093). I confirmed that `libcudart-stub.so` still wins when placed before the real runtime in mocked maps; the new negative cases do not cover that suffix yet.

One detail from the failing pre-commit log: Ruff reformats **both** `flashinfer/comm/cuda_ipc.py` and `tests/comm/test_cuda_ipc_library_lookup.py`, so please format both files. No additional correctness findings from this pass.

JacobHelwig:
> Re-reviewed [819d774](https://github.com/flashinfer-ai/flashinfer/commit/819d7744839c89e7144dc35219762e8cd66fca50). Both follow-ups from my previous review are addressed: the matcher now rejects the `.something` / `.so.backup` cases, and the new tests cover the requested maps-order, stub-only, hashed-runtime, and directory-name scenarios.
> 
> I ran the unmodified new test file against the two production lookup functions extracted from this commit: **19 passed**. This bypassed package imports and repository conftest on macOS; it does not validate the normal package import path or CUDA integration.
> 
> The remaining hyphen-suffix gap is already tracked in [the existing review thread](https://github.com/flashinfer-ai/flashinfer/pull/5198#discussion_r4002585093). I confirmed that `libcudart-stub.so` still wins when placed before the real runtime in mocked maps; the new negative cases do not cover that suffix yet.
> 
> One detail from the failing pre-commit log: Ruff reformats **both** `flashinfer/comm/cuda_ipc.py` and `tests/comm/test_cuda_ipc_library_lookup.py`, so please format both files. No additional correctness findings from this pass.

Thanks, @yyihuang , here's the diff since your last review:  https://github.com/flashinfer-ai/flashinfer/compare/819d7744839c89e7144dc35219762e8cd66fca50...JacobHelwig:flashinfer:jhelwig/fix-libcudart-lookup-matches-stub?expand=1

yyihuang (approved):
Re-reviewed c4da14f. All previously raised concerns are addressed: the build suffix is restricted to hexadecimal characters, negative cases cover hyphenated stubs and mocks, and pre-commit now passes. No remaining correctness findings from this review.

The unmodified lookup test file passes all **23 cases** against the production functions extracted from this commit. This was a macOS CPU-only check that bypassed package imports and repository conftest, not a CUDA integration run.

Approving the code change. The latest commit's Test Results Summary still reports `CI skipped — pending authorization`; the earlier pipeline result predates these commits, so CI for the current head remains outstanding.

yyihuang:
@flashinfer-bot run

yzh119 (approved):

PR #5200, opened 2026-09-14
feat(moe): tag the unified MoE API as official with @flashinfer_api

aleozlx:
## 📌 Description

Promotes the unified MoE API to an official FlashInfer API by tagging it with `@flashinfer_api`.

The unified MoE API was deliberately left untagged while it matured. FlashInfer intends lower-level component APIs and higher-level unified APIs to **co-exist** — this is a maturity milestone, not the correction of an oversight, and neither layer supersedes the other. `docs/design_docs/flashinfer_moe_api.md` recorded adding `@flashinfer_api` as an explicit release-time gate; this closes it.

**Exactly two entry points are tagged:**

| entry point | signature |
|---|---|
| `MoELayer.__init__` | `(self, config: MoEConfig, device: Optional[torch.device] = None)` |
| `MoELayer.__call__` | `(self, act_pack: MoEActivationPack, weight_pack: MoEWeightPack) -> Union[torch.Tensor, List[torch.Tensor]]` |

Everything else in the surface is deliberately **not** tagged: the config dataclasses (`MoEConfig`, `QuantConfig`, `RoutingConfig`, `ExpertConfig`, `ActivationConfig` and subclasses, `MoEFinalizeConfig`, `ExecutionConfig`, the per-backend `*Config` types) are frozen data rather than callables, and the `TunableRunner` methods are driven by the autotuner in a tight per-tactic loop where per-call logging would be actively harmful.

### Two supporting changes — required, not cosmetic

**`flashinfer/api_logging.py`** — `MoELayer` joins the class-name prefix list. Its entry point is `__call__`, so without the prefix every unified-MoE call logs as a bare `__call__`: unreadable, and useless as a `FLASHINFER_DUMP_INCLUDE` / `FLASHINFER_DUMP_EXCLUDE` pattern since it would also match any other decorated `__call__`.

**`flashinfer/fused_moe/api.py`** — metadata-only `__repr__` on `MoEActivationPack` and `MoEWeightPack`. These are the arguments of the newly decorated `__call__`, and they are **dataclasses holding CUDA tensors**. The logging guards against `str()`/`repr()` on tensor-bearing containers are keyed on the concrete `list`/`tuple`/`dict` types, so a dataclass falls through to the generic `repr` fallback in both `_format_value` (level 3+ logging) and `_serialize_value` (level-10 dump metadata). Tensor `repr` reads device memory and **can invalidate CUDA graph capture**. Fixing it at the source covers every consumer rather than special-casing two pack types inside the shared logger. Non-tensor values fall back to `repr` instead of assuming `.shape`, so an unrelated debugger inspection cannot raise.

### Documentation

`MoELayer` appeared only in an `autosummary` stub, so `__call__` — a dunder — never rendered. This adds an explicit `.. autoclass::` with `automethod` for both entry points, matching the pattern already used for the `CuteDsl*` wrappers in the same file, plus a short note that the two API layers co-exist by design.

`scripts/pr_checks/check_api_docs` is unchanged at **MISSING 0**. (STALE 48 is pre-existing debt on `main`, predominantly `flashinfer.comm`; untouched here.)

## 🔍 Related Issues

Closes the `@flashinfer_api` release gate recorded in `docs/design_docs/flashinfer_moe_api.md`.

### ✅ Pre-commit Checks

## 🧪 Tests

Verified: `py_compile`, `mypy`, `ruff check`, `ruff format`, and the full pre-commit hook set, plus `scripts/pr_checks/check_api_docs`. No behavioural change — `@flashinfer_api` is zero-overhead at the default `FLASHINFER_LOGLEVEL=0`.

## Reviewer Notes

**Deliberately not included: the `trace=` argument** from CLAUDE.md's Trace Template Checklist. Two reasons, both worth a reviewer's opinion:

1. `TraceTemplate` models **flat tensor argument lists** (`axes` / `inputs` / `outputs` as `Tensor`/`Scalar`/`Var` specs, with `init` building flat kwargs). `MoELayer.__call__` takes **pack dataclasses**. Wiring `trace=` therefore means either extending the trace machinery to model packs, or writing a flattening adapter — a design change rather than wiring, and better reviewed on its own.
2. The checklist's mandatory steps — regenerating `tests/trace/fi_trace_out/` via `python tests/trace/example.py` and running `pytest tests/trace/` — need a torch environment that was not available while preparing this change. Committing a `TraceTemplate` whose generated JSON had not actually been regenerated would be worse than deferring it.

Raising it explicitly so the deviation is deliberate and visible rather than silent. Happy to land it as a follow-up, or to fold it in here if a reviewer would rather not split them.

**Keyword-only guard (included — this section previously said otherwise).** `RoutingConfig` and `ExpertConfig` each gain a trailing `_: KW_ONLY` sentinel. It is placed **at the end**, with no fields after it, so it is a **no-op today**: every existing field keeps its current binding and `RoutingConfig(8, 2)` still binds positionally. Only fields *appended below* the sentinel become keyword-only, which is the growth this is meant to make safe — inserting or appending a field can otherwise silently rebind an existing positional argument at a call site that still type-checks. `QuantConfig` already carries the same guard from #5061.

Corrections to an earlier revision of this description, both caught in review: it claimed the keyword-only change was "not in this PR" (it is), and it quoted field counts of 12 / 8 for `RoutingConfig` / `ExpertConfig` — the real counts on this branch are **6** and **4**. The inflated numbers came from a script that counted docstring `Parameters` entries alongside real fields.

## Summary by CodeRabbit

- **Documentation**
  - Added API documentation for `MoELayer`, including initialization, invocation, instrumentation, and lower-level interfaces.
  - Updated design documentation covering logging, tracing, and tensor-dump behavior.

- **Improvements**
  - Improved operation logging and diagnostics for `MoELayer`.
  - Added concise metadata displays for MoE activation and weight data.
  - Updated configuration constructors so selected optional settings are keyword-only.

- **Tests**
  - Added coverage verifying class-specific operation names in logging.


aleozlx:
@flashinfer-bot run tests/moe tests/utils

flashinfer/fused_moe/layer.py
```diff
@@ -143,6 +144,7 @@ class MoELayer:
     >>> out = layer(act_pack, weight_pack)
     """
 
+    @flashinfer_api
     def __init__(self, config: MoEConfig, device: Optional[torch.device] = None):
```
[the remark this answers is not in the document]
aleozlx: Added a full docstring (Parameters / Returns / Raises). Verified the claims against the implementation rather than assuming: `__init__` raises `RuntimeError` when no configured backend is usable on the device arch, and `__call__` returns a list only when `config.finalize.do_finalize=False`. `check_pr_document.py` now reports 0 new findings.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

flashinfer/fused_moe/layer.py
```diff
@@ -230,6 +232,7 @@ def __init__(self, config: MoEConfig, device: Optional[torch.device] = None):
         # Backend key selected on the most recent call (introspection hook).
         self._last_winner_backend: Optional[str] = None
 
+    @flashinfer_api
     def __call__(
```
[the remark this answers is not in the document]
aleozlx: Added a full docstring (Parameters / Returns / Raises). Verified the claims against the implementation rather than assuming: `__init__` raises `RuntimeError` when no configured backend is usable on the device arch, and `__call__` returns a list only when `config.finalize.do_finalize=False`. `check_pr_document.py` now reports 0 new findings.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

aleozlx:
@flashinfer-bot run tests/moe tests/moe_ep

aleozlx:
### GPU smoke test — B200, all checks pass

Ran on a B200 (`2u2g-spr-0143`) against this PR at `9394c898`, covering the three properties CI cannot reach:

```
PASS  LOGLEVEL=0 returns the original function
PASS  LOGLEVEL=3 logs 'MoELayer.__call__'
PASS  non-allow-listed class is NOT prefixed
PASS  pack repr mentions shape/dtype/device
      MoEActivationPack(hidden_states_q=Tensor(shape=(4, 8), dtype=torch.bfloat16, device=...), ...)
PASS  pack repr does NOT dump tensor values
PASS  repr is safe under CUDA graph capture

RESULT: ALL PASS
```

**Why these needed a GPU / why CI would not have caught them.**

- The existing `tests/utils/test_logging.py::test_class_method_logging` asserts `"TestWrapper.run" in log_contents`, which passes via the `"Wrapper" in class_name` branch — pure name matching. The line this PR changes is the *other* branch, the explicit allow-list. Nothing covered it, so CI would have gone green whether or not the `MoELayer` prefix worked. The negative control (`NotOnTheAllowList.__call__` absent from the log) confirms the pass comes from the allow-list branch. A unit test for this is added in `c7ee1b1a`.
- `LOGLEVEL=0` is asserted to return the *original function object*, i.e. genuinely zero overhead on the default path rather than a cheap wrapper.

**One caveat, stated rather than buried:** PyTorch warns `The CUDA Graph is empty` on the capture check. That is expected — `repr(pack)` enqueues no CUDA work, which is the property under test — but it does mean capture succeeded partly *because nothing was recorded*. Treat that check as supporting evidence, not proof; the two direct assertions above it (metadata present, no tensor values) are unambiguous.

_🤖 Verified by [Claude Code](https://claude.com/claude-code)_

aleozlx:
@flashinfer-bot run

docs/api/fused_moe.rst
```diff
@@ -49,6 +49,19 @@ weight / activation formats and the layer output format as ``QuantFormat`` axes.
     MoEActivationPack
     MoEWeightPack
 
+``MoELayer`` is the official entry point of this API: both its constructor and
+its call operator are decorated with ``@flashinfer_api``, so they participate in
+``FLASHINFER_LOGLEVEL`` logging and ``FLASHINFER_DUMP_*`` capture. The lower-level
+per-backend functions above remain official in their own right — the two layers
+are designed to co-exist, and neither supersedes the other.
+
+.. autoclass:: MoELayer
```
feih-nv, line 57: `MoELayer` is already in the `autosummary` list above (line 40) with `:toctree: ../generated`, and `autosummary_generate = True` emits a stub page that runs `autoclass` on it. Adding a second `.. autoclass:: MoELayer :members:` here documents the same object twice, so Sphinx will warn `duplicate object description of flashinfer.fused_moe.MoELayer ... use :no-index: for one of them` and cross-references become ambiguous. The docs build has no `-W`, so it is only a warning.

Note the `CuteDsl*MoEWrapper` / `B12xMoEWrapper` blocks below are `autoclass`-only; none of them is also in an autosummary list, so to match that pattern drop `MoELayer` from the autosummary list (or add `:no-index:` to this block).

flashinfer/fused_moe/layer.py
```diff
@@ -230,11 +253,41 @@ def __init__(self, config: MoEConfig, device: Optional[torch.device] = None):
...
         self,
         act_pack: MoEActivationPack,
         weight_pack: MoEWeightPack,
     ) -> Union[torch.Tensor, List[torch.Tensor]]:
+        """Run the MoE layer, selecting and caching the fastest usable backend.
+
+        Only runners that support this pack's ``routing_input_mode`` compete —
+        not every backend has an in-kernel router — and the winner is cached
+        per ``(num_tokens, mode)`` so repeated calls skip reselection.
```
feih-nv: Nit: the winner is cached per `(map_to_hybrid_bucket(num_tokens, ceiling), mode)`, not per `num_tokens` — several token counts share one bucket. Suggest "per (token-count bucket, mode)" so the docstring matches the cache key below.

flashinfer/fused_moe/api.py
```diff
@@ -114,6 +114,13 @@ class RoutingConfig:
     n_group: Optional[int] = None
     topk_group: Optional[int] = None
     routed_scaling_factor: Optional[float] = None
+    # Append new fields BELOW this KW_ONLY sentinel. ``_: KW_ONLY`` is a
+    # ``dataclasses`` marker, not a real field -- every field declared after it
+    # becomes keyword-only. That way a field added later can never silently
+    # rebind an existing positional argument at a call site that still looks
+    # correct. The fields above keep their current binding, so adding this
+    # changes nothing for existing callers.
+    _: KW_ONLY
```
feih-nv, line 124: This (and the same sentinel on `ExpertConfig`) contradicts the PR description, whose Reviewer Notes say the keyword-only change is "not in this PR"; CodeRabbit's summary already reads it as "selected optional settings are keyword-only". As placed — trailing, with no fields after it — it is a no-op today (`RoutingConfig(8, 2)` still binds positionally; verified), so the guard itself is fine, but please describe it accurately. Also, the description's field counts (RoutingConfig 12 / ExpertConfig 8) are 6 and 4 on this branch.

docs/api/fused_moe.rst
aleozlx: Right — `autosummary_generate = True` (docs/conf.py:46) already emits a stub that runs `autoclass`, so this was a duplicate object description. Dropped `MoELayer` from the autosummary list and kept the `autoclass` block, matching the `CuteDsl*`/`B12x` pattern you pointed at (none of those are in an autosummary list). Keeping `autoclass` is what renders `__call__`, which the stub would omit. `0af06795`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

flashinfer/fused_moe/layer.py
aleozlx: Correct — `self._winners[(bucket, mode)]` with `bucket = map_to_hybrid_bucket(act_pack.num_tokens, ceiling)` (layer.py:313,317), so several token counts share one entry. Docstring now says "per (token-count bucket, mode)". `0af06795`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

flashinfer/fused_moe/api.py
```diff
@@ -1716,6 +1730,39 @@ def __getitem__(self, key: str):
 # backend-specific layout logic out of the dispatch hot-path.
 
 
+def _tensor_summary(value: Any) -> str:
```
feih-nv, line 1734: Nit: this docstring restates the whole CUDA-graph / `_serialize_value` rationale that the design doc (Release Gates) and the PR body already carry, and it quotes an `api_logging.py` comment verbatim. That copy will silently go stale the next time the logger changes, and nobody will think to update `fused_moe/api.py`. Two or three lines are enough here: tensors are summarized by shape/dtype/device to avoid a device-to-host copy (which also breaks CUDA-graph capture); see the design doc for why the logger cannot do this itself.

flashinfer/fused_moe/api.py
aleozlx: Both right, and both were my errors. The description said the keyword-only change was "not in this PR" while the diff contained it, and the 12/8 counts came from a script that counted docstring `Parameters` entries as fields — the real counts are 6 and 4. PR description rewritten to describe the sentinel accurately (trailing, no-op today, only future appended fields become keyword-only) and with the corrected counts.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

flashinfer/fused_moe/api.py
```diff
@@ -1930,3 +1996,12 @@ def get_view(self, backend_key: str) -> Dict[str, Tensor]:
                 f"Available: {list(self.native_views)}"
             )
         return self.native_views[backend_key]
+
+    def __repr__(self) -> str:
+        views = ", ".join(
+            f"{key!r}: {{"
+            + ", ".join(f"{n}: {_tensor_summary(t)}" for n, t in view.items())
```
feih-nv: Nit: this assumes every view is a `dict`. `prepare_for` stores whatever it is given without checking, and `native_views` is a public field, so a non-mapping view makes `repr(pack)` raise `AttributeError` on `.items()`. That is the same failure mode `_tensor_summary` guards against for values ("a `__repr__` that raises turns an unrelated pytest assertion or debugger inspection into a confusing AttributeError"). The logger itself is safe (`_format_value` catches and prints `<MoEWeightPack object>`), so this only bites interactive use. Suggest `repr(view)` when `not isinstance(view, dict)`, or validate the type in `prepare_for`.

tests/utils/test_logging.py
```diff
@@ -314,6 +314,50 @@ def run(self, x):
...
+                def __call__(self, x):
+                    return x * 3
+
+            assert MoELayer()(5) == 15
+            assert NotOnTheAllowList()(5) == 15
+
+            with open(log_file, "r") as f:
+                log_contents = f.read()
+
+            assert "Wrapper" not in "MoELayer"  # guards the premise of this test
```
feih-nv: Nit: this compares two literals, so it can never fail and does not guard anything. The premise (the class name lacks "Wrapper") is established by how the class is named two lines up; a comment says that just as well. Suggest dropping the assert.

flashinfer/fused_moe/api.py
aleozlx: Agreed — a verbatim copy of a comment in another file is a stale-doc generator. Trimmed to the contract plus pointers: it now references `_serialize_value` in `api_logging.py` and the design doc Release Gates section rather than restating either. `01328731`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

flashinfer/fused_moe/api.py
aleozlx: Good catch, and this one was a real bug rather than a nit — I applied the defensive fallback to values via `_tensor_summary` but not one level up to the views themselves, so a non-mapping view made `repr(pack)` raise the exact `AttributeError` the helper exists to prevent. `__repr__` now falls back to `_tensor_summary` for anything that is not a `Mapping`. `01328731`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

tests/utils/test_logging.py
aleozlx: Right — two literals, so it can never fail. Dropped; the class name two lines above establishes the premise on its own. `01328731`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

aleozlx:
@flashinfer-bot run

feih-nv (approved):
LGTM.

Aneureka (approved):

